PR #1189, opened 2026-06-05
fix: wrap ELBv2 void query responses in their Result envelope

mvanhorn:
## Summary

ELBv2 void query actions (`DeleteTargetGroup`, `DeleteListener`, `DeleteLoadBalancer`, `SetSubnets`) return their response without the `<XResult>` wrapper element. The AWS query protocol always nests an empty Result element inside the Response, and SDK/Terraform deserializers expect it - against floci, `terraform destroy` and the AWS CLI fail to parse these responses. This adds the `<{Action}Result></{Action}Result>` element in `ElbV2QueryHandler.voidResponse()`, deriving the name from the response name. Closes #1096.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: real ELBv2 returns e.g. `<DeleteTargetGroupResponse xmlns=...><DeleteTargetGroupResult/><ResponseMetadata>...</ResponseMetadata></DeleteTargetGroupResponse>`; floci omitted the `<DeleteTargetGroupResult/>` element, so query-protocol deserializers (AWS CLI, Terraform AWS provider) rejected the response. The wrapper shape matches the ELBv2 2015-12-01 API reference response examples for all four void actions.

## Checklist

- [x] `./mvnw test` passes locally (`ElbV2IntegrationTest`: 35/35)
- [x] New or updated integration test added (Result-envelope assertions for all four void actions, plus a SetSubnets round-trip)
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

PR #1206, opened 2026-06-06
feat(elbv2): align DescribeLoadBalancers availability zones with AWS behavior

### LINKED ISSUE #1076: [FEAT] Replace static mock Maps with real AvailabilityZone entities to prevent breaking dependent plugins and SDKs
## Service
EC2 / ELBv2

## API Action / Feature
Implementation of AvailabilityZone as a real class instead of static mock Maps.

## AWS Documentation
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeAvailabilityZones.html
https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_AvailabilityZone.html
https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_DescribeLoadBalancers.html

## Why is this needed?
I discovered this issue while using the Ansible plugin `community.aws.elb_application_lb`. The plugin crashes during idempotency checks because the ELBv2 describe response omits required relation data like `SubnetId`.

Digging deeper, the root cause is that `AvailabilityZone` does not exist as an actual entity/class in Floci. It is entirely mocked using static Maps. For example, in `Ec2Service.java`:
```java
    public List<Map<String, String>> describeAvailabilityZones(String region) {
        List<Map<String, String>> zones = new ArrayList<>();
        String[] azSuffixes = {"a", "b", "c"};
        for (String suffix : azSuffixes) {
            Map<String, String> az = new LinkedHashMap<>();
            az.put("zoneName", region + suffix);
            az.put("state", "available");
            az.put("regionName", region);
            az.put("zoneId", region + "-az" + (suffix.charAt(0) - 'a' + 1));
            az.put("zoneType", "availability-zone");
            zones.add(az);
        }
        return zones;
    }
```
Because the class does not exist, there is no real data or state mapped to it. Consequently, any specific verification, resource association, or strict lookup from official AWS clients fails.

- [ ] Yes
- [x] No

shrimptails-f:
## Summary

Align `CreateLoadBalancer` / `DescribeLoadBalancers` availability zone output with AWS-compatible ELBv2 behavior.

This PR:
- adds an ELBv2 `AvailabilityZone` model with `SubnetId` and `ZoneName`
- changes load balancers to store structured availability zone data instead of raw subnet ID strings
- resolves `subnetId -> subnet -> availabilityZone` via EC2 when creating/updating load balancers
- returns `AvailabilityZones.member.SubnetId` and `AvailabilityZones.member.ZoneName` in the ELBv2 XML response shape
- adds an integration test covering subnet-aware load balancer creation and describe behavior

Closes #1076

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Bug fix.

Previous behavior:
- ELBv2 `LoadBalancer.availabilityZones` stored subnet IDs as `List<String>`
- `DescribeLoadBalancers` returned subnet IDs in `AvailabilityZones.member.ZoneName`
- `AvailabilityZones.member.SubnetId` was missing entirely

Fixed behavior:
- ELBv2 now stores structured availability zone data
- `CreateLoadBalancer` / `DescribeLoadBalancers` return AWS-compatible `SubnetId` and `ZoneName` fields for each availability zone entry

Verified with:
- integration test for subnet-based load balancer create/describe flow in `ElbV2IntegrationTest`

## Checklist

`./mvnw -Dtest=ElbV2IntegrationTest test` passed; full `./mvnw test` currently hits unrelated existing `ElastiCacheMemcachedIntegrationTest` connection-refused errors.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -96,7 +96,7 @@ public class Ec2Service {
 
     // ─── Default resource seeding ──────────────────────────────────────────────
 
-    void ensureDefaultResources(String region) {
+    public void ensureDefaultResources(String region) {
         if (!seededRegions.add(region)) {
             return;
         }
@@ -204,7 +204,7 @@ public class Ec2Service {
         // Resolve subnet
         Subnet subnet = null;
         if (subnetId != null && !subnetId.isEmpty()) {
-            subnet = getRequiredSubnet(region, subnetId);
+            subnet = requireSubnet(region, subnetId);
         } else {
             // Pick first default subnet
             subnet = subnets.values().stream()
@@ -319,7 +319,8 @@ public class Ec2Service {
         return reservation;
     }
     
-    private Subnet getRequiredSubnet(String region, String subnetId) {
+    public Subnet requireSubnet(String region, String subnetId) {
+        ensureDefaultResources(region);
         Subnet subnet = subnets.get(key(region, subnetId));
         if (subnet == null) 
             throw new AwsException("InvalidSubnetID.NotFound", "The subnet ID '" + subnetId + "' does not exist", 400);
@@ -649,7 +650,7 @@ public class Ec2Service {
 
     public void modifySubnetAttribute(String region, String subnetId, String attribute, String value) {
         ensureDefaultResources(region);
-        Subnet subnet = getRequiredSubnet(region, subnetId);
+        Subnet subnet = requireSubnet(region, subnetId);
         switch (attribute) {
             case "mapPublicIpOnLaunch"           -> subnet.setMapPublicIpOnLaunch(Boolean.parseBoolean(value));
             case "assignIpv6AddressOnCreation"   -> subnet.setAssignIpv6AddressOnCreation(Boolean.parseBoolean(value));
--- a/src/main/java/io/github/hectorvent/floci/services/elbv2/ElbV2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/elbv2/ElbV2QueryHandler.java
@@ -829,8 +829,11 @@ public class ElbV2QueryHandler {
         xml.start("State").elem("Code", safe(lb.getState())).end("State");
         xml.elem("Type", safe(lb.getType()));
         xml.start("AvailabilityZones");
-        for (String az : lb.getAvailabilityZones()) {
-            xml.start("member").elem("ZoneName", az).end("member");
+        for (AvailabilityZone az : lb.getAvailabilityZones()) {
+            xml.start("member")
+                    .elem("SubnetId", safe(az.getSubnetId()))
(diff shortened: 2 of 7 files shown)


hectorvent:
The structured availability zone fix is correct and matches AWS. Verified against the ELBv2 API reference: `AvailabilityZone.SubnetId` is "The ID of the subnet" and `ZoneName` is "The name of the Availability Zone", and the `CreateLoadBalancer` sample response emits each member as `<SubnetId>` then `<ZoneName>`, exactly what this PR produces. Resolving subnet to AZ name for `ZoneName` and storing the subnet id in `SubnetId` fixes #1076.

One change before merge: the missing subnet path returns the wrong error code. `resolveAvailabilityZones` calls `ec2Service.requireSubnet`, which throws EC2's `InvalidSubnetID.NotFound`. This now fires inside the ELBv2 `CreateLoadBalancer` and `SetSubnets` operations. The ELBv2 `CreateLoadBalancer` Errors section documents:

```
SubnetNotFound  - The specified subnet does not exist.  (HTTP 400)
InvalidSubnet   - The specified subnet is out of available addresses.  (HTTP 400)
```

So the correct code for a non existent subnet is `SubnetNotFound` (not `InvalidSubnet`, which is address exhaustion). An SDK client modeling ELBv2 will not map an EC2 `InvalidSubnetID.NotFound`. Suggest translating it in `ElbV2Service.resolveAvailabilityZones`:

```java
try {
    subnet = ec2Service.requireSubnet(region, subnetId);
} catch (AwsException e) {
    throw new AwsException("SubnetNotFound",
            "Subnet '" + subnetId + "' not found", 400);
}
```

This path is newly reachable because the PR makes a non existent subnet a hard failure, which is the more AWS accurate behavior (AWS requires real subnets), but it is a behavior change worth calling out: callers passing subnets that are not present in EC2 now fail, which is why the CloudFormation test moved from `subnet-aaa`/`subnet-bbb` to `subnet-default-a`/`subnet-default-b`.

Heads up on a cross PR conflict with #1189 (SetSubnets). That PR returns a populated `SetSubnetsResult` and its test asserts `ZoneName` equals the subnet id using unseeded subnets (`subnet-00000001`). Both contradict this PR (`ZoneName` is now the real AZ, and unseeded subnets now throw). Whichever merges second will need a rebase and a test update.

shrimptails-f:
@hectorvent 
For the cross-PR conflict with #1189, I think it is fine to let whichever PR merges second handle the rebase and test updates. The desired final behavior should be that `ZoneName` contains the actual AZ name and unseeded subnets fail with `SubnetNotFound`, so the later PR can adjust its `SetSubnets` test expectations accordingly.

I also addressed the error-code issue called out in this review. The ELBv2 path now translates a missing subnet into `SubnetNotFound` instead of leaking EC2's `InvalidSubnetID.NotFound`, and I added integration coverage for both  `CreateLoadBalancer` and `SetSubnets`.

hectorvent (approved):
Thanks @shrimptails-f

PR #1213, opened 2026-06-06
feat(athena): support DeleteWorkGroup action

mansiverma897993:
## Type of change

- [x] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [x] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/athena/AthenaJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/athena/AthenaJsonHandler.java
@@ -91,7 +91,7 @@ public class AthenaJsonHandler {
             }
             case "DeleteWorkGroup" -> {
                 String wg = request.path("WorkGroup").asText(null);
-                if (wg == null || wg.isBlank()) {
+                if (wg == null || !wg.matches("[a-zA-Z0-9._-]{1,128}")) {
                     throw new AwsException("InvalidRequestException", "WorkGroup is required.", 400);
                 }
                 if ("primary".equals(wg)) {
--- a/src/main/java/io/github/hectorvent/floci/services/athena/AthenaJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/athena/AthenaJsonHandler.java
@@ -90,6 +90,13 @@ public class AthenaJsonHandler {
                 yield Response.ok(Map.of("TableMetadata", athenaService.getTableMetadata(catalog, database, tableName))).build();
             }
             case "DeleteWorkGroup" -> {
+                String wg = request.path("WorkGroup").asText(null);
+                if (wg == null || wg.isBlank()) {
+                    throw new AwsException("InvalidRequestException", "WorkGroup is required.", 400);
+                }
+                if ("primary".equals(wg)) {
+                    throw new AwsException("InvalidRequestException", "The primary workgroup cannot be deleted.", 400);
+                }
                 yield Response.ok(Map.of()).build();
             }
             default -> throw new AwsException("InvalidAction", "Action " + action + " is not supported", 400);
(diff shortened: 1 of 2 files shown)


hectorvent:
Reviewed against the athena `service-2.json` model. LGTM, no blocking wire issues.

The shape is correct: Athena is awsjson1.1 and `DeleteWorkGroupOutput` has no members, so an empty body serializes as `{}`. Returning `Response.ok(Map.of())` matches, and no undeclared error codes are thrown (the op only declares `InternalServerException` and `InvalidRequestException`). It is also consistent with the existing `CreateWorkGroup -> {}` no op, and since no workgroup state is persisted there is nothing to actually delete, so the stub is internally consistent and unblocks `aws_athena_workgroup` destroy.

Two minor parity notes, both optional given the workgroup subsystem is mocked:

1. The required `WorkGroup` field is ignored. The model marks it REQUIRED (`pattern [a-zA-Z0-9._-]{1,128}`), so a request missing it returns `InvalidRequestException` on AWS while this returns 200.
2. `primary` is deletable here. The op doc states the primary workgroup cannot be deleted (AWS returns `InvalidRequestException`).

A cheap way to cover both:

```java
case "DeleteWorkGroup" -> {
    String wg = request.path("WorkGroup").asText(null);
    if (wg == null || wg.isBlank())
        throw new AwsException("InvalidRequestException", "WorkGroup is required.", 400);
    if ("primary".equals(wg))
        throw new AwsException("InvalidRequestException", "The primary workgroup cannot be deleted.", 400);
    yield Response.ok(Map.of()).build();
}
```

Process nit: the PR description is the untouched template with all four Type of change boxes checked (including Breaking change, which it is not) and an empty AWS Compatibility section. Worth tidying so the squash merge metadata is accurate. The change is a minor feat.

mansiverma897993:
@hectorvent I have sucessfully update this as per your accordance , plz cheack at once and lemme what changes is required ?

hectorvent:
Thanks for addressing the earlier feedback. The error code is now correct (`InvalidRequestException` is one of the two declared errors for this action), the primary workgroup block matches the docs, and the empty `{}` output is right. I verified this against the Athena SDK model and the API reference.

One suggestion before merge, about the name validation. The commit headline mentions validating the WorkGroup name, but the handler currently only checks for null or blank. AWS constrains the name to the pattern `[a-zA-Z0-9._-]{1,128}`, and this is enforced on the service side. The AWS SDK for Java v2 does not validate the pattern on the client, so a caller can send a name with spaces or one longer than 128 characters, and real Athena rejects it with `InvalidRequestException`. Today that name would pass here.

You can cover both the blank case and the pattern in one check:

```java
String wg = request.path("WorkGroup").asText(null);
if (wg == null || !wg.matches("[a-zA-Z0-9._-]{1,128}")) {
    throw new AwsException("InvalidRequestException", "WorkGroup is required.", 400);
}
if ("primary".equals(wg)) {
    throw new AwsException("InvalidRequestException", "The primary workgroup cannot be deleted.", 400);
}
```

A small test would round it out, for example a name with a space:

```java
given()
    .header("X-Amz-Target", "AmazonAthena.DeleteWorkGroup")
    .contentType(CONTENT_TYPE)
    .body("{ \"WorkGroup\": \"bad name\" }")
.when()
    .post("/")
.then()
    .statusCode(400)
    .body("__type", equalTo("InvalidRequestException"));
```

Everything else looks good to me. Verified against `athena/service-2.json` and the DeleteWorkGroup API reference.

mansiverma897993:
@hectorvent Done I have successfully update the PR !!

hectorvent (approved):
Thanks @mansiverma897993

mansiverma897993:
Thnxx @hectorvent I am moving to solve next issue !!

Issue #1198, opened 2026-06-05
[BUG] Lambda ZIP created with PowerShell Compress-Archive is extracted with incorrect path separators


Tobotobo:
Note: This issue was translated and drafted with the assistance of AI.

## Service

Lambda

## AWS API Action

CreateFunction  
UpdateFunctionCode

## Expected behavior

When deploying a Lambda function from a ZIP file created on Windows, files under directories should be extracted as normal Linux paths.

For example:

```text
wwwroot/app.css
wwwroot/_framework/blazor.web.js
wwwroot/lib/bootstrap/dist/css/bootstrap.css
```

## Actual behavior

When the ZIP file is created with PowerShell `Compress-Archive`, files are extracted with backslashes as literal characters in the filename.

Inside `/var/task`:

```text
...
SnapshotRestore.Registry.dll
'wwwroot\app.css'
'wwwroot\app.css.br'
'wwwroot\DotnetStudy.styles.css'
'wwwroot\_framework\blazor.web.js'
'wwwroot\lib\bootstrap\dist\css\bootstrap.css'
...
```

Because of this, the application cannot find files such as:

```text
/var/task/wwwroot/app.css
```

This caused static file requests in a Blazor app to fail.

Creating the ZIP with `tar.exe` works correctly.

## Reproduction

```ps1
dotnet new blazor -n Sample -int None --no-https
dotnet add Sample package Amazon.Lambda.AspNetCoreServer.Hosting
dotnet publish -r linux-x64 Sample
Compress-Archive -Path "./Sample/bin/Release/net10.0/linux-x64/publish/*" -DestinationPath "./publish.zip" -Force
```

```bash
aws lambda create-function \
  --function-name dotnet-blazor-lambda \
  --runtime dotnet10 \
  --role "$ROLE_ARN" \
  --handler Sample \
  --zip-file fileb://publish.zip \
  --architectures x86_64
```

```bash
docker exec -it floci-dotnet-blazor-lambda-b2fdf52b bash
cd /var/task
ls
```

Actual output includes:

```text
...
SnapshotRestore.Registry.dll
'wwwroot\app.css'
'wwwroot\app.css.br'
...
```

## Workaround

Creating the ZIP with Windows `tar.exe` works.

```ps1
tar.exe -a -c -f publish.zip -C "./Sample/bin/Release/net10.0/linux-x64/publish" .
```

## Environment

* floci/floci:1.5.22
* Host OS: Windows
* Runtime: dotnet10

[...]

PR #1215, opened 2026-06-06
fix(lambda): normalize backslash ZIP path separators on extraction

ousamabenyounes:
## Summary

Lambda ZIPs created on Windows with PowerShell `Compress-Archive` store entry names with backslash (`\`) path separators. The ZIP spec (APPNOTE.TXT 4.4.17) mandates `/`, and on Linux `\` is a literal filename character — so `ZipExtractor` wrote a single flat file named `wwwroot\app.css` instead of creating the nested `wwwroot/app.css` path.

**However**, real AWS Lambda does NOT normalize `\` to `/` — a PowerShell 5 `Compress-Archive` zip extracts as a literal-named file on Lambda, identical to what Floci does today. AWS recommends building on a POSIX OS (or WSL), and the C# guide steers you to `dotnet lambda` / SAM / CDK, all of which emit forward-slash zips. The fix lives on the packaging side, not the consumer side.

So this PR now:
- Keeps backslash separators literal (matching AWS Lambda exactly)
- Emits a clear `LOG.warnv` when backslashes are detected, telling the developer how to repackage correctly (`tar`, `pwsh`, or `dotnet lambda` CLI)
- Keeps the traversal guard and containment check (defense-in-depth)

Floci behaves exactly like AWS, and the developer gets an actionable signal instead of a silent local-only success that breaks on deploy.

## Type of change

- [x] Bug fix (AWS compatibility — Floci now matches real Lambda behavior)

## RED → GREEN verification (real `./mvnw test` run, not simulated)

**RED** — old normalization code vs new test:

```
Tests run: 3, Failures: 1, Errors: 0, Skipped: 0 <<< FAILURE!

ZipExtractorTest.extractsBackslashEntriesAsLiteralFilename FAILED
  backslashed entry must extract as a literal filename (matching AWS Lambda)
```

The normalization code creates nested paths, but the new test asserts literal filenames (AWS-congruent). The test catches the fidelity mismatch.

**GREEN** — warning code:

```
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS

WARN ZIP entry "wwwroot\_framework\blazor.web.js" uses backslash separators (PowerShell Compress-Archive).
  It extracts as a literal filename and will also fail on real AWS Lambda.
  Repackage with tar, PowerShell Core (pwsh), or the dotnet lambda CLI.

WARN ZIP entry "..\..\evil.sh" uses backslash separators ...
WARN Skipping suspicious ZIP entry: ..\..\evil.sh
```

New tests:
- `extractsBackslashEntriesAsLiteralFilename` — backslashed entry creates a literal flat file, not nested
- `stillExtractsStandardForwardSlashEntries` — forward-slash entries unchanged
- `rejectsBackslashTraversalEntries` — `..\..\evil.sh` still caught by traversal guard

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/zip/ZipExtractor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/zip/ZipExtractor.java
@@ -20,13 +20,7 @@ public class ZipExtractor {
 
     private static final Logger LOG = Logger.getLogger(ZipExtractor.class);
 
-    // The ZIP spec (APPNOTE.TXT 4.4.17) mandates '/' as the entry path separator.
-    // PowerShell's Compress-Archive instead writes '\', which on Linux is a literal
-    // filename character — so "wwwroot\app.css" would become a single flat file
-    // instead of a nested path. Normalize '\' -> '/' so such archives extract
-    // correctly (matching how real AWS Lambda unpacks them).
-    private static final char WINDOWS_SEPARATOR = '\\';
-    private static final char UNIX_SEPARATOR = '/';
+    private static final char BACKSLASH = '\\';
 
     public void extractTo(byte[] zipBytes, Path targetDir) throws IOException {
         // Resolve to absolute path so that normalize() on entry paths stays comparable
@@ -36,9 +30,17 @@ public class ZipExtractor {
         try (ZipInputStream zis = new ZipInputStream(new ByteArrayInputStream(zipBytes))) {
             ZipEntry entry;
             while ((entry = zis.getNextEntry()) != null) {
-                // Normalize Windows-style separators before any path handling so the
-                // traversal guard and resolution below operate on canonical names.
-                String entryName = entry.getName().replace(WINDOWS_SEPARATOR, UNIX_SEPARATOR);
+                String entryName = entry.getName();
+
+                // PowerShell 5 Compress-Archive writes '\' as a literal filename byte on
+                // Linux. Real AWS Lambda does NOT normalize this (the archive extracts to
+                // a flat "wwwroot\app.css" file), so neither do we; masking it would let
+                // a broken package pass locally and then fail on deploy.
+                if (entryName.indexOf(BACKSLASH) >= 0) {
+                    LOG.warnv("ZIP entry \"{0}\" uses backslash separators (PowerShell Compress-Archive). "
+                            + "It extracts as a literal filename and will also fail on real AWS Lambda. "
+                            + "Repackage with tar, PowerShell Core (pwsh), or the dotnet lambda CLI.", entryName);
+                }
 
                 // Security: prevent path traversal
                 if (entryName.contains("..") || entryName.startsWith("/")) {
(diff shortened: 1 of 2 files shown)


hectorvent:
Thanks for the thorough writeup and the RED/GREEN test discipline. One part of the security work is a genuine improvement and should stay: running the traversal guard on the resolved name is good defense-in-depth, and the `targetPath.startsWith(absTarget)` containment check is correct.

I'd like to request a change to the core behavior, though, because the premise needs correcting.

**Real AWS Lambda does not normalize `\` to `/`.** The "actual behavior" in #1198 is Floci's output, not AWS's. On Lambda (Linux) a PowerShell 5 `Compress-Archive` zip extracts `wwwroot\app.css` as a single literal-named file, identical to what Floci does today. AWS documents this only indirectly: it never claims Lambda repairs separators, it recommends building on a POSIX OS (or WSL), and the C# guide steers you to `dotnet lambda` / SAM / CDK, all of which emit forward-slash zips. The fix lives on the packaging side, not the consumer side.

So normalizing here makes Floci **more forgiving than AWS**: the package works locally, then breaks on deploy. For an emulator whose value is fidelity, that hides the bug instead of surfacing it.

Proposed direction: keep the separators literal (match AWS), but emit a clear warning so the developer knows why and how to fix it:

```java
String entryName = entry.getName();

// PowerShell 5 Compress-Archive writes '\' as a literal filename byte on Linux.
// Real AWS Lambda does NOT normalize this (the archive extracts to a flat
// "wwwroot\app.css" file), so neither do we; masking it would let a broken
// package pass locally and then fail on deploy.
if (entryName.indexOf('\\') >= 0) {
    LOG.warnv("ZIP entry \"{0}\" uses backslash separators (PowerShell Compress-Archive). "
        + "It extracts as a literal filename and will also fail on real AWS Lambda. "
        + "Repackage with tar, PowerShell Core (pwsh), or the dotnet lambda CLI.", entryName);
}
```

The existing traversal guard and containment check stay as-is. The `extractsBackslashEntriesAsNestedPaths` test would invert to assert the entry stays literal (and ideally that the warning fires); the forward-slash and traversal tests are unaffected.

Net: Floci behaves exactly like AWS, and the developer gets an actionable signal instead of a silent local-only success.

(Extraction isn't in the Lambda `service-2.json` model, so this is validated against AWS docs and observed Lambda behavior, not the wire model.)

ousamabenyounes:
Thanks for the thorough review @hectorvent — you're right, I got the premise wrong.

I assumed AWS Lambda normalized `\` to `/` on extraction, but that was Floci's output, not AWS's. On a real Linux Lambda, `wwwroot\app.css` extracts as a single literal filename. For an emulator whose value is fidelity, silently fixing a broken package is worse than surfacing the bug — "works locally, breaks on deploy" is the exact failure mode we should prevent.

Pivoted to your suggested direction:
- Removed the `\` → `/` normalization — backslash entries now extract as literal filenames, matching real AWS Lambda.
- Added a `LOG.warnv` with actionable repackaging instructions when backslash separators are detected.
- Kept the traversal guard and containment check as-is.

Test pivoted: `extractsBackslashEntriesAsNestedPaths` → `extractsBackslashEntriesAsLiteralFilename`, asserting the flat file IS created (AWS-congruent) and the nested path is NOT.

**RED → GREEN** (real `./mvnw test` run, not simulated):

```
RED (normalization code):  Tests run: 3, Failures: 1
  extractsBackslashEntriesAsLiteralFilename FAILED
  → expected: <true> but was: <false>

GREEN (warning code):     Tests run: 3, Failures: 0, BUILD SUCCESS
  WARN visible for both backslash entries
```

hectorvent (approved):
Thanks @ousamabenyounes

PR #1224, opened 2026-06-08
feat(ec2): add VPC Flow Logs and Interface VPC Endpoints

skwolvie:
## What

Adds three EC2/VPC networking-observability capabilities to the emulator, plus the `SecurityAudit` AWS-managed policy:

- **VPC Flow Logs** — `CreateFlowLogs`, `DescribeFlowLogs`, `DeleteFlowLogs`. A flow log attached to a VPC/subnet/ENI delivers synthetic VPC flow-log records to the configured S3 bucket, using the AWS **default (version 5) record format** under the real AWS key layout (`AWSLogs/{account}/vpcflowlogs/{region}/{yyyy}/{MM}/{dd}/{HH}/...log.gz`, gzipped). Records are correlated to the EC2 resources that actually exist in the emulator (instance ids, ENI ids, subnet/VPC ids, private IPs), and same-VPC instances converse with each other plus any interface VPC endpoints, so the generated traffic is realistic rather than purely random.
- **Interface VPC Endpoints (PrivateLink)** — `CreateVpcEndpoint`, `DescribeVpcEndpoints`, `DescribeVpcEndpointConnections`, `DeleteVpcEndpoints`. An interface endpoint creates one ENI per subnet with a private IP inside the VPC, mirroring AWS PrivateLink. `DescribeVpcEndpointConnections` returns a well-formed empty set so clients that enumerate connections per endpoint don't error.
- **`SecurityAudit` managed policy** — added to the IAM managed-policy seed list so `AttachRolePolicy arn:aws:iam::aws:policy/SecurityAudit` succeeds. This is a standard AWS managed policy that was missing.
- **Optional `FLOCI_EC2_REGIONS`** — an opt-in env var that restricts the regions reported by `DescribeRegions` to a comma-separated subset. Useful when a client enumerates regions and sweeps each one and you want to scope that sweep. When unset, the full default region list is returned (no behavior change).

## Why

The emulator already models EC2 inventory (VPCs, subnets, instances, ENIs, security groups) well, but had no way to produce **VPC flow logs** or model **PrivateLink interface endpoints** — both of which are needed to exercise tools that consume network-traffic data (log pipelines, SIEMs, traffic-analysis and cloud-visibility software) entirely locally, with no real cloud spend. These additions let such tools be tested against realistic, resource-correlated traffic and PaaS-access topologies.

The `SecurityAudit` policy and the `FLOCI_EC2_REGIONS` filter are small supporting changes that come up in the same workflow (audit roles, and scoping a multi-region enumeration).

## Implementation notes

- All actions use the real AWS **EC2 Query Protocol** (form-encoded request, XML response) and **IAM Query Protocol** — no custom endpoints; the AWS SDK/CLI and Terraform work unmodified.
- The flow-log generator runs on a background scheduler (60s) and skips delivery gracefully if the destination bucket doesn't exist yet, retrying on the next tick.
- New EC2 actions are registered in `AwsQueryController` and dispatched in `Ec2QueryHandler`; state lives in `Ec2Service` / `FlowLogService` with new `FlowLog` and `VpcEndpoint` models.

## Tests

- `Ec2FlowLogAndVpcEndpointIntegrationTest` — create/describe/delete flow logs (asserts the resource correlation and record-format fields), create/describe an interface endpoint (asserts it gets an ENI), and the empty `DescribeVpcEndpointConnections` set.
- `SecurityAuditManagedPolicyIntegrationTest` — `GetPolicy` resolves `SecurityAudit` and `AttachRolePolicy` succeeds against a role.

Both follow the existing `@QuarkusTest` + RestAssured Query-Protocol style. The existing suite continues to pass (the `DescribeRegions` default behavior is unchanged).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/AwsQueryController.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/AwsQueryController.java
@@ -161,7 +161,10 @@ public class AwsQueryController {
             "DescribeIamInstanceProfileAssociations",
             "DescribeAvailabilityZones", "DescribeRegions", "DescribeAccountAttributes",
             "DescribeInstanceTypes",
-            "DescribeNetworkInterfaces"
+            "DescribeNetworkInterfaces",
+            "CreateFlowLogs", "DescribeFlowLogs", "DeleteFlowLogs",
+            "CreateVpcEndpoint", "DescribeVpcEndpoints", "DescribeVpcEndpointConnections",
+            "DeleteVpcEndpoints"
     );
 
     private final CloudFormationQueryHandler cloudFormationQueryHandler;
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
@@ -24,10 +24,12 @@ public class Ec2QueryHandler {
             .withZone(ZoneOffset.UTC);
 
     private final Ec2Service service;
+    private final FlowLogService flowLogService;
 
     @Inject
-    public Ec2QueryHandler(Ec2Service service) {
+    public Ec2QueryHandler(Ec2Service service, FlowLogService flowLogService) {
         this.service = service;
+        this.flowLogService = flowLogService;
     }
 
     public Response handle(String action, MultivaluedMap<String, String> params, String region) {
@@ -115,6 +117,15 @@ public class Ec2QueryHandler {
                 case "DescribeInstanceTypes" -> handleDescribeInstanceTypes(params, region);
                 // Network Interfaces
                 case "DescribeNetworkInterfaces" -> handleDescribeNetworkInterfaces(params, region);
+                // VPC Flow Logs
+                case "CreateFlowLogs" -> handleCreateFlowLogs(params, region);
+                case "DescribeFlowLogs" -> handleDescribeFlowLogs(params, region);
+                case "DeleteFlowLogs" -> handleDeleteFlowLogs(params, region);
+
+                case "CreateVpcEndpoint" -> handleCreateVpcEndpoint(params, region);
+                case "DescribeVpcEndpoints" -> handleDescribeVpcEndpoints(params, region);
+                case "DescribeVpcEndpointConnections" -> handleDescribeVpcEndpointConnections(params, region);
+                case "DeleteVpcEndpoints" -> handleDeleteVpcEndpoints(params, region);
                 // Volumes
                 case "CreateVolume" -> handleCreateVolume(params, region);
                 case "DescribeVolumes" -> handleDescribeVolumes(params, region);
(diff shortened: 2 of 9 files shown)


hectorvent:
Thank you for this. It is a substantial and genuinely useful contribution, and the management plane is in great shape. I checked everything against `ec2/service-2.json` and the VPC Flow Logs documentation. Below is what already matches AWS, one item that needs attention, and a few small notes.

## What already matches AWS

1. Every response wire shape lines up with the model `locationName` values. I verified `flowLogIdSet`, the `flowLogSet` item fields, `vpcEndpointSet`, `vpcEndpoint`, `subnetIdSet`, `networkInterfaceIdSet`, `routeTableIdSet`, `groupSet`, `dnsEntrySet`, and the empty `unsuccessful` sets.
2. The S3 key layout is exactly right, including the gzip and the file name shape:
   ```
   AWSLogs/{account}/vpcflowlogs/{region}/{yyyy}/{MM}/{dd}/{HH}/{account}_vpcflowlogs_{region}_{flowLogId}_{YYYYMMDDTHHmmZ}_{hash}.log.gz
   ```
3. The interface endpoint model (one private IP ENI per subnet, surfaced in `DescribeNetworkInterfaces` with `interfaceType=vpc_endpoint`) is a clean AWS shaped model.
4. The `SecurityAudit` policy resolves to `arn:aws:iam::aws:policy/SecurityAudit`, which is correct.

## Main item: the flow log record content is not the AWS default format

The code and description say the records use the "AWS default format (version 5, full 29 field layout, with a header line)." The documentation says the opposite on each point:

> With the default format, the flow log records include the version 2 fields, in the order shown in the available fields table. You cannot customize or change the default format.

> version: If you use the default format, the version is 2.

So the real default is 14 fields, in the documented table order, with the version field equal to 2:

```
version account-id interface-id srcaddr dstaddr srcport dstport protocol packets bytes start end action log-status
```

A real default record looks like this (one line per flow, no header row in the file):

```
2 123456789012 eni-0abc12 10.0.1.20 10.0.1.31 41234 443 6 24 5860 1718000000 1718000060 ACCEPT OK
```

The current output differs in four ways:

1. It writes 29 fields instead of 14. The v3, v4, and v5 fields only appear when the user requests a custom format, never in the default.
2. The field order is alphabetical (`account-id action az-id bytes ...`). AWS always uses the table order in both default and custom formats.
3. The `version` column is set to `5`. For the default it must be `2`.
4. A header row is written as the first line. The objects AWS writes to S3 contain records only, with no header line.

There is also a desync in the custom path. When `LogFormat` is provided, `customHeader()` reflects the requested fields, but `record()` still emits the same fixed 29 alphabetical fields, so the header no longer describes the rows. Example of the mismatch today:

```
# header written from the user format
srcaddr dstaddr srcport dstport protocol
# but every record line still has all 29 fields in alphabetical order
123456789012 ACCEPT use1-az1 5860 10.0.1.31 443 1718000060 egress i-0a1b ...
```

Why it matters: the goal of this feature is that a tool which parses VPC flow logs can ingest the emulated traffic. A consumer built for the AWS default (14 version 2 fields, no header) or for a specific custom format will not parse a 29 field alphabetical record. The good news is that the field values you compute are fine (protocol 6, log-status OK, type IPv4, the direction and traffic path values, the `S3` service tag). Only the field set, order, version value, and header need to change.

Suggested approach:

1. When no `LogFormat` is supplied, emit the 14 version 2 fields in table order, set the version column to `2`, and write no header.
2. When `LogFormat` is supplied, emit exactly those fields in that order, and set the version column to the highest field version among them.

## Small notes

1. `FLOCI_EC2_REGIONS` is read with `System.getenv` directly inside `Ec2Service.describeRegions()`. Per the project config conventions this is better as a `floci.*` property in `EmulatorConfig` plus `application.yml`, which Quarkus maps from `FLOCI_EC2_REGIONS` automatically. Same behavior, easier to test.
2. `DescribeVpcEndpoints` accepts `filters` but applies only region and ids, and `DescribeFlowLogs` applies no filters. AWS supports filters such as `vpc-id`, `service-name`, `vpc-endpoint-state`, `resource-id`, and `traffic-type`. This is incomplete rather than incorrect.
3. `awsServiceFromEndpoint` always returns `S3`, so a DynamoDB or other endpoint is still tagged `S3`. Deriving it from the endpoint `serviceName` would make `pkt-dst-aws-service` accurate.
4. `CreateFlowLogs` does not echo `clientToken` in the response when one is supplied. Optional, since AWS only returns it when present.
5. `vpcEndpoints`, `endpointEnis`, and `flowLogs` are kept in memory rather than through `StorageFactory`. This is consistent with the rest of `Ec2Service`, so it is not a regression. It does mean endpoints and flow logs do not survive a restart, the same as the rest of EC2.

## Summary

The EC2 and IAM management surface is wire accurate and the endpoint modeling is solid. The one change worth making before merge is the flow log record format, plus the custom format desync. Once the records emit the true default (14 version 2 fields, table order, version 2, no header) and honor a supplied `LogFormat`, this will deliver on the realistic flow log goal. Thanks again for the thorough work.

Verified against `ec2/service-2.json` and the VPC Flow Logs records documentation.

hectorvent:
closing in favor or #1611

PR #1238, opened 2026-06-08
fix(msk): populate currentBrokerSoftwareInfo in cluster describe response

silfabio:
## Summary

`DescribeCluster` / `DescribeClusterV2` never populated `currentBrokerSoftwareInfo`
in the response, so SDK / CLI / Terraform reads of
`ClusterInfo.CurrentBrokerSoftwareInfo.KafkaVersion` always saw `null`, even
though it's a real, commonly-inspected field on MSK clusters.

- Accept an optional `kafkaVersion` on `CreateCluster` / `CreateClusterV2`,
  defaulting to `3.6.0` when omitted or blank
- Add a `BrokerSoftwareInfo` model and populate
  `MskCluster.currentBrokerSoftwareInfo` at creation time so it round-trips
  through `DescribeCluster`
- Add `MskServiceTest` coverage for both the default-version and
  explicit-version paths

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Matches the real `DescribeClusterResponse` shape, where
`ClusterInfo.CurrentBrokerSoftwareInfo.KafkaVersion` reflects the version the
cluster was created with — `CreateCluster` callers that pass `kafkaVersion`
now see it echoed back on describe, and those that omit it get a sensible
default instead of `null`.

## Checklist

- [x] New or updated integration test added (`MskServiceTest`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/msk/MskController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/msk/MskController.java
@@ -23,7 +23,8 @@ public class MskController {
     @Path("/v1/clusters")
     public Response createCluster(Map<String, Object> request) {
         String clusterName = (String) request.get("clusterName");
-        MskCluster cluster = mskService.createCluster(clusterName);
+        String kafkaVersion = (String) request.get("kafkaVersion");
+        MskCluster cluster = mskService.createCluster(clusterName, kafkaVersion);
         return Response.ok(Map.of("clusterArn", cluster.getClusterArn(), "clusterName", cluster.getClusterName(), "state", cluster.getState())).build();
     }
 
@@ -32,7 +33,8 @@ public class MskController {
     public Response createClusterV2(Map<String, Object> request) {
         // Simple mapping to V1 for now
         String clusterName = (String) request.get("clusterName");
-        MskCluster cluster = mskService.createCluster(clusterName);
+        String kafkaVersion = (String) request.get("kafkaVersion");
+        MskCluster cluster = mskService.createCluster(clusterName, kafkaVersion);
         return Response.ok(Map.of("clusterArn", cluster.getClusterArn(), "clusterName", cluster.getClusterName(), "state", cluster.getState())).build();
     }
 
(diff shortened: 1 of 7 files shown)


hectorvent (approved):

hectorvent:
Nice fix, thanks for adding this. The `kafkaVersion` round trip through `currentBrokerSoftwareInfo` looks correct, and the casing matches the model exactly (the wire `locationName` values are `currentBrokerSoftwareInfo` and `kafkaVersion`, both lowercase first), so v1 `DescribeCluster` reads will line up with real MSK.

One thing worth tightening on the V2 path. Verified against the `kafka` `service-2.json`, `CreateClusterV2` does not carry `kafkaVersion` at the top level. Its input is `{clusterName, tags, provisioned, serverless}`, and the version lives at `provisioned.kafkaVersion` (`ProvisionedRequest.KafkaVersion`). So `createClusterV2`'s `request.get("kafkaVersion")` is always null today, which means a V2 caller that explicitly sets a version still gets the `3.6.0` default echoed back on describe.

Something like this would pick it up:

```java
// createClusterV2
Map<String, Object> provisioned = (Map<String, Object>) request.get("provisioned");
String kafkaVersion = provisioned != null ? (String) provisioned.get("kafkaVersion") : null;
MskCluster cluster = mskService.createCluster(clusterName, kafkaVersion);
```

That keeps the existing "map V2 to V1" approach while sourcing the version from the right place. Everything else looks good to me, the default fallback and the v1 path are both correct.

(Strict V2 parity also nests `currentBrokerSoftwareInfo` under `provisioned` in the `DescribeClusterV2` response, but that is a pre existing flatness in `MskCluster` and out of scope for this PR.)

albert-einshutoin (changes requested):
Blocking this for now because the V2 path still reads kafkaVersion from the top-level request. CreateClusterV2 carries it under provisioned.kafkaVersion, so explicit V2 versions will still describe as the default 3.6.0. This matches the earlier review comment and should be fixed before merge.

silfabio:
thanks @hectorvent and @albert-einshutoin for flagging that and for your comments. I agree with you and I'm working on the suggested corrections, will commit the fix soon.

hectorvent:
Thanks @silfabio, 
I see all addressed... merging

PR #1237, opened 2026-06-08
fix(msk): poll correct Redpanda admin readiness endpoint

silfabio:
## Summary

`RedpandaManager.isReady()` polls Redpanda's admin API to decide whether the
MSK-emulation container is ready, but it polled `/ready` — an endpoint that
always returns 404 on Redpanda's actual admin API. The cluster therefore
never reported ready, and anything waiting on it (e.g. Terraform's
`create_msk = true` provisioning) timed out.

- Poll `/v1/status/ready` (the endpoint Redpanda's admin API actually serves
  readiness on) instead of `/ready`
- Add `RedpandaManagerTest.isReadyPollsTheCorrectAdminReadinessPathInNativeMode`:
  a local HTTP server stands in for the admin API (200 on `/v1/status/ready`,
  404 on `/ready`), the Docker container-inspect chain is mocked to resolve
  the admin port binding to it, and `isReady()` is asserted to report ready —
  reverting `ADMIN_READY_PATH` back to `/ready` makes the test fail, pinning
  the fix as a regression guard

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore

## AWS Compatibility

Not an AWS wire-protocol change — `/v1/status/ready` is Redpanda's own admin
API (the container floci runs to emulate MSK locally), not an AWS API
surface. Verified locally end-to-end: with the fix, `isReady()` correctly
reports the Redpanda container ready and Terraform's `create_msk = true`
provisioning completes against it.

## Checklist

- [x] `./mvnw test` passes locally
- [x] New or updated integration test added (`RedpandaManagerTest`)
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

PR #1239, opened 2026-06-08
fix(msk): advertise externally reachable broker address

silfabio:
## Summary

Redpanda embeds broker addresses directly in Kafka protocol responses
(`Metadata`, `FindCoordinator`, ...), and clients reconnect to whatever
address it advertises. Without `--advertise-kafka-addr`, Redpanda advertised
its self-perceived loopback (`127.0.0.1:9092`) — unreachable from any client
outside the container, host-side or sibling container alike. Producers and
consumers would connect to the published bootstrap address, receive metadata
pointing at `127.0.0.1:9092`, and then hang/timeout trying to reconnect there.

- Resolve and pass `--advertise-kafka-addr` explicitly at container startup:
  a pre-allocated `localhost:<port>` in native mode (the host port must be
  known *before* the container starts, since the flag is a startup arg — so
  it's allocated via `PortAllocator` instead of letting Docker assign one
  dynamically), or `<containerName>:9092` in container mode, resolved by
  sibling containers through Docker's embedded DNS
- Add `RedpandaManagerTest` covering both address-resolution paths
  (`nativeModeAdvertisesPreAllocatedLocalhostAddress`,
  `containerModeAdvertisesContainerNameAddress`)
  
Update: TOCTOU-safe port allocation + release on delete

allocateAny() found a free port via an ephemeral socket but didn't reserve it — a race if multiple containers start concurrently. Now uses portAllocator.allocate(kafkaHostPortBase, kafkaHostPortMax) (new 9300-9399 range, configurable via floci.services.msk.kafka-host-port-base/-max), which reserves the port until released — matching RDS/ElastiCache/ECR.

stopContainer() releases the port back to the pool (parsed from bootstrapBrokers); a failed createAndStart also releases it, so the range doesn't leak across start/stop cycles.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Not an AWS wire-protocol change — `--advertise-kafka-addr` is a Redpanda
startup flag (the container floci runs to emulate MSK locally), not an AWS
API surface. Verified locally end-to-end: with the fix, producers/consumers
connect to the advertised address, discover the group coordinator, and
complete the full produce → consume → process round trip without
`Disconnecting from node -1` / reconnect-timeout symptoms.

## Checklist

- [x] New or updated integration test added (`RedpandaManagerTest`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
+++ b/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
@@ -488,6 +488,12 @@ public interface EmulatorConfig {
 
         @WithDefault("redpandadata/redpanda:latest")
         String defaultImage();
+
+        @WithDefault("9300")
+        int kafkaHostPortBase();
+
+        @WithDefault("9399")
+        int kafkaHostPortMax();
     }
 
     interface ElastiCacheServiceConfig {
--- a/src/main/java/io/github/hectorvent/floci/services/msk/RedpandaManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/msk/RedpandaManager.java
@@ -10,6 +10,7 @@ import io.github.hectorvent.floci.core.common.docker.ContainerLifecycleManager.E
 import io.github.hectorvent.floci.core.common.docker.ContainerLogStreamer;
 import io.github.hectorvent.floci.core.common.docker.ContainerSpec;
 import io.github.hectorvent.floci.core.common.docker.ContainerStorageHelper;
+import io.github.hectorvent.floci.core.common.docker.PortAllocator;
 import io.github.hectorvent.floci.services.msk.model.MskCluster;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
@@ -40,6 +41,7 @@ public class RedpandaManager {
     private final ContainerDetector containerDetector;
     private final EmulatorConfig config;
     private final RegionResolver regionResolver;
+    private final PortAllocator portAllocator;
     private final Map<String, Closeable> logStreams = new ConcurrentHashMap<>();
 
     @Inject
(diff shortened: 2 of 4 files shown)


hectorvent:
Thanks for chasing this down, the `--advertise-kafka-addr` diagnosis is spot on. Redpanda baking `127.0.0.1:9092` into the `Metadata` response is exactly the kind of thing that makes producers hang on reconnect, and advertising `localhost:<hostPort>` in native mode and `<containerName>:9092` for sibling containers is the right call. Switching native mode to `withPortBinding` so the bootstrap address and the advertised address agree is a nice touch too.

One thing I would tighten. The pre-allocation uses `portAllocator.allocateAny()`, which opens an ephemeral `ServerSocket(0)`, reads the port, and closes it right away without recording it:

```java
public int allocateAny() {
    try (ServerSocket socket = new ServerSocket(0)) {   // closed on return
        return socket.getLocalPort();                   // never added to `reserved`
    } ...
}
```

That reopens the TOCTOU window the allocator was built to close. Its own docs call this out:

> `reserved` — Ports reserved by this process but not yet bound by Docker. Prevents TOCTOU races when multiple containers are launched concurrently.

Since we bake the port into `--advertise-kafka-addr` before the container binds it, a concurrent cluster start (or any other container) can grab that port in the gap, and Redpanda then comes up advertising a port it does not actually own. The reserving variant avoids that, and is what RDS and ElastiCache already use:

```java
kafkaHostPort = portAllocator.allocate(MSK_KAFKA_BASE_PORT, MSK_KAFKA_MAX_PORT); // held in `reserved`
kafkaAdvertiseAddr = "localhost:" + kafkaHostPort;

// and on container removal / cluster delete, so reservations do not leak:
portAllocator.release(kafkaHostPort);
```

Heads up that this is also conflicting now since #1237 merged and already added `RedpandaManagerTest` plus touched `RedpandaManager`. A rebase that folds these advertised address tests into the existing test class (rather than re adding the file) should sort it out, keeping both the readiness constant and this startup logic.

Really solid fix overall, just the port reservation and the rebase to square away.

hectorvent:
@silfabio also please resolve conflicts.

silfabio:
thanks @hectorvent for the comments, very good catch. I've switched allocateAny() to the reserving portAllocator.allocate(kafkaHostPortBase, kafkaHostPortMax) (new 9300-9399 range, same *BasePort/*MaxPort pattern as RDS/ElastiCache/ECR), and added portAllocator.release(...) in stopContainer() (plus on a failed createAndStart) so the port returns to the pool on cluster deletion. Also added stopContainerReleasesAllocatedKafkaHostPort covering the release path, and verified manually end-to-end: cluster create → binds host port 9300 → localhost:9300 reachable from the host via nc → delete → port 9300 correctly reused by the next cluster. Could you please review those changes and let me know if you are ok with them?

hectorvent:
@silfabio one more time, I will try to get it merged as soon it is resolved. 
Thanks!

silfabio:
@hectorvent thanks again for looking into this. I've solved the conflicts and merged as requested. Would you mind taking a look?

hectorvent (approved):

PR #1252, opened 2026-06-09
Add EC2 provisioning primitives

jvanzyl:
## Summary
- Load EC2 AMI metadata from a resource-backed catalog
- Add Ubuntu 24.04 AMI metadata
- Add launch template, launch template version, VPC endpoint, NAT gateway, and instance offering support

## Tests
- ./mvnw test -Dtest=Ec2ImageCatalogTest,Ec2IntegrationTest


hectorvent (approved):
It looks good to me.

PR #1253, opened 2026-06-09
Add RDS provisioning support

jvanzyl:
## Summary
- Add DB subnet group APIs and orderable instance options
- Preserve instance tags and requested engine versions
- Return managed master user secret metadata and routable proxy endpoints
- Improve PostgreSQL proxy compatibility for SSL negotiation and IAM role setup

## Tests
- ./mvnw test -Dtest=RdsQueryHandlerTest,RdsServiceTest,PostgresProtocolHandlerTest,RdsContainerManagerTest

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/RdsControlPlaneTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/RdsControlPlaneTest.java
@@ -1,0 +1,55 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.rds.RdsClient;
+import software.amazon.awssdk.services.rds.model.CreateDbSubnetGroupResponse;
+import software.amazon.awssdk.services.rds.model.DescribeDbSubnetGroupsResponse;
+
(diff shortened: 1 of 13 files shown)


hectorvent:
Really clean work on the subnet group and orderable options support, the field names and list elements line up with the RDS model nicely.

One wire bug to fix though. In `dbSubnetGroupInnerXml` the `Subnets` list wraps each entry in `<member>`, but the model's `SubnetList` member has `locationName = "Subnet"`, so real RDS emits `<Subnet>`:

```xml
<Subnets>
  <Subnet>
    <SubnetIdentifier>subnet-…</SubnetIdentifier>
    <SubnetAvailabilityZone><Name>…</Name></SubnetAvailabilityZone>
    <SubnetStatus>Active</SubnetStatus>
  </Subnet>
</Subnets>
```

So the one line change is `.start("Subnet")` instead of `.start("member")`. With `<member>`, a real AWS SDK unmarshaller parses an empty subnets list on `DescribeDBSubnetGroups`/`CreateDBSubnetGroup`, since it keys off `<Subnet>`. Every other list in the PR already uses the right element (`DBSubnetGroup`, `OrderableDBInstanceOption`, `AvailabilityZone`, `Tag`), so it is just this one.

Heads up that the handler test asserts `Subnets.member.…`, which matches the current output and hides this, so a quick `DescribeDBSubnetGroups` round trip through the Java SDK would be the surest check. Everything else (the `MasterUserSecret` block, subnet group fields, orderable options) matches the model.

hectorvent (approved):

PR #1254, opened 2026-06-09
Improve ELBv2 target resolution and persistence

jvanzyl:
## Summary
- Resolve instance target addresses for ELBv2 data plane and health checks
- Persist load balancer state
- Preserve listener sockets during updates
- Restore target health after persisted state reload

## Tests
- ./mvnw test -Dtest=ElbV2IntegrationTest,ElbV2TargetResolverTest,ElbV2ServiceTest

### DIFF
--- a/docs/services/elb.md
+++ b/docs/services/elb.md
@@ -2,12 +2,12 @@
 
 **Protocol:** Query (XML) — `POST http://localhost:4566/` with `Action=` parameter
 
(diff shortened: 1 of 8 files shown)


hectorvent (approved):
All good, merging

hectorvent:
see conflicts that need to be addressed

hectorvent (approved):
Thanks @jvanzyl

PR #1260, opened 2026-06-09
feat(pipes): support Kafka pipe sources and polling

LiamMacP:
## Summary

Existing [documentation](https://github.com/floci-io/floci/blob/f669116908bd23e4034b56a0a35bf8269e8f5ffe/docs/services/pipes.md?plain=1#L84) specifies that MSK can be used as a source for the EventBridge Pipes. Within the current implementation, the [PipePoller](https://github.com/floci-io/floci/blob/f669116908bd23e4034b56a0a35bf8269e8f5ffe/src/main/java/io/github/hectorvent/floci/services/pipes/PipesPoller.java#L122-L130) does not support integrating with either Kafka or MSK. As such this PR adds support for both Amazon MSK and Kafka (via `smk://`) sources to EventBridge Pipes. 

As part of this the change introduces Kafka consumer management for Pipes so Floci can:
- resolve bootstrap brokers for managed and self-managed Kafka
- validate required Kafka source parameters at pipe creation/update time
- poll Kafka topics safely across poll cycles
- apply filter criteria against decoded Kafka key/value payloads
- commit offsets only after successful delivery

It also expands test coverage with:
- unit tests for Kafka consumer manager behaviour
- unit tests for Kafka poller delivery and commit semantics
- integration coverage for Kafka pipe creation and validation through the Pipes API

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

This is a new EventBridge Pipes source capability rather than a wire-protocol bug fix.

Verified with:
- AWS CLI `aws pipes` request/response compatibility already exercised by the existing Pipes REST integration suite
- added integration coverage for Kafka pipe creation and validation through the Floci Pipes API

Implementation notes:
- MSK sources use the existing MSK bootstrap broker lookup
- self-managed Kafka sources use the `smk://` source form plus `SelfManagedKafkaParameters`
- Kafka filter matching is applied to decoded key/value payloads, while delivery preserves the expected Kafka event shape
- offsets are committed only after successful target delivery

## Checklist


hectorvent:
Really nice addition, thanks for wiring up both MSK and self managed Kafka here. The `smk://` source form, the `:kafka:` ARN detection, the required `TopicName` validation, the base64 `key`/`value` encoding, and committing offsets only after a successful delivery all line up with AWS. The test coverage is great too.

One thing to flag for parity: the **Kafka header values** are base64 encoded, but AWS encodes them as a JSON array of byte numbers.

I checked this against the AWS Lambda event type source (the same per record encoding Pipes uses for Kafka):

- `aws-lambda-java-events` → `private List<Map<String, byte[]>> headers;` (the value is a `byte[]`)
- `aws-lambda-go/events` → `Headers []map[string]JSONNumberBytes`, where `JSONNumberBytes` is a `[]byte` type whose custom marshaller exists specifically to serialize bytes as numbers, not base64.

So a header value of `hello` should look like this on the wire:

```json
"headers": [
  { "myHeader": [104, 101, 108, 108, 111] }
]
```

but the current `buildKafkaRecordNodes` produces:

```json
"headers": [
  { "myHeader": "aGVsbG8=" }
]
```

`key` and `value` are correctly base64, only `headers` differs. A small change covers it:

```java
// current:
// headerNode.put(header.key(), base64(headerValue));

// AWS shape (array of unsigned byte numbers):
var values = headerNode.putArray(header.key());
for (byte b : headerValue) {
    values.add(b & 0xFF);
}
```

Minor, non blocking note while you are in there: each record also carries an `eventSourceKey` field (`topic-partition`). In the AWS event types that `topic-partition` string is only ever the key of the `records` map, never a field inside the record, so it is not part of the AWS shape. It works fine as your internal dedup key, but it might be worth keeping it out of the payload delivered to the target.

Thanks again for the contribution!

LiamMacP:
Sorry for the delay @hectorvent – I've applied that change requested.

hectorvent:
Thanks for this, really nice addition. MSK backed by Redpanda plus the commit-after-delivery semantics look great, and the source parameter shapes match the model exactly (verified `TopicName`, `StartingPosition`, `BatchSize`, `ConsumerGroupID`, `AdditionalBootstrapServers` against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/pipes/2015-10-07/service-2.json), and `smk://host:port` is allowed by the `Source` pattern). A few small things before merge:


**1. Kafka record shape.** Comparing `buildKafkaRecordNodes` in [`PipesPoller.java`](src/main/java/io/github/hectorvent/floci/services/pipes/PipesPoller.java) to the [AWS MSK source docs](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-msk.html), two fields differ from the documented event:

```json
{
  "eventSource": "aws:kafka",
  "eventSourceArn": "arn:aws:kafka:...",
  "eventSourceKey": "mytopic-0",
  "topic": "mytopic",
  "partition": "0",
  "offset": 15
}
```

`eventSourceKey` (the `"topic-partition"` value) is not emitted, and `partition` is documented as a string but the code writes a number. Both are easy adds if you want byte exact fidelity for input transformers and targets.

**2. Header byte values.** `putKafkaBinaryField` writes header bytes as `b & 0xFF` (0 to 255). AWS uses signed Java bytes, so a `0xFF` byte should serialize as `-1`. The doc example is all ASCII so it does not show, but binary header values above 127 will diverge. Dropping the `& 0xFF` would match.

Nothing blocking on the wire contract itself, the param names and error codes are all correct. Mostly just want to be sure the native build survives. Thanks again!

hectorvent (approved):

PR #1266, opened 2026-06-10
fix(sts): persist session secret key so RDS/ElastiCache IAM tokens validate

simlau:
## Problem

STS session credentials (ASIA... access keys) issued by `AssumeRole`, `GetSessionToken`, `AssumeRoleWithWebIdentity`, `AssumeRoleWithSAML`, and `GetFederationToken` are always rejected when used as RDS (or ElastiCache) IAM auth tokens. The connection fails with `28P01 password authentication failed`.

## Root cause

Four interlocking gaps in three files:

**1. `StsQueryHandler` discards the temporary secret key**

Every STS handler generates `String secretKey = randomSecret(40)` and returns it to the caller in the XML response, but calls:
```java
iamService.registerSession(accessKeyId, roleArn, expiration, sessionPolicy);
// ↑ secretKey is NOT passed — it is silently discarded
```

**2. `SessionCredential` has no `secretAccessKey` field**

The model stores `accessKeyId`, `roleArn`, `expiration`, and `sessionPolicyDocument` — but never the secret.

**3. `IamService.findSecretKey` only checks `accessKeys`**

```java
public Optional<String> findSecretKey(String accessKeyId) {
    return accessKeys.get(accessKeyId).map(AccessKey::getSecretAccessKey);
    // sessions store is never consulted
}
```

**4. `RdsSigV4Validator` falls back to using the access key ID as the secret**

```java
String secretKey = iamService.findSecretKey(accessKeyId).orElse(accessKeyId);
//                                                        ↑ always reached for ASIA... keys
//                                                          uses the key ID itself as the signing secret → always wrong
```

Result: every HMAC comparison fails → `"RDS IAM token signature mismatch for accessKey=ASIA…"` → `28P01`.

Note: `GetSessionToken` also did not call `registerSession` at all, so its tokens had no entry anywhere.

## Fix

**`SessionCredential.java`** — add `secretAccessKey` field + constructor + getter/setter.

**`IamService.java`** — add a `registerSession` overload that accepts the secret, and make `findSecretKey` check `sessions` as a fallback:

```java
public Optional<String> findSecretKey(String accessKeyId) {
    Optional<String> fromAccessKey = accessKeys.get(accessKeyId).map(AccessKey::getSecretAccessKey);
    if (fromAccessKey.isPresent()) {
        return fromAccessKey;
    }
    return sessions.get(accessKeyId).map(SessionCredential::getSecretAccessKey);
}
```

**`StsQueryHandler.java`** — update all five handlers to pass `secretKey` to `registerSession`. Add `registerSession` call to `GetSessionToken` (which previously had none).

**Tests** — add `IamServiceTestHelper.iamServiceWithSessionCredential` helper and two new `RdsSigV4ValidatorTest` cases covering the ASIA key path (`validateAcceptsTokenSignedWithStsSessionCredentials` and `validateRejectsStsTokenWithWrongSecret`).

## Scope

The same `findSecretKey` call is made by the ElastiCache `SigV4Validator`, so ElastiCache IAM auth with STS credentials is fixed by the same change.

## Testing

All 5640 existing tests continue to pass. The two new test cases specifically exercise the previously-broken ASIA key path.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
@@ -866,6 +866,10 @@ public class IamService {
                 sessions.delete(accessKeyId);
                 return null; // expired — unknown key → bypass
             }
+
+            if (session.getRoleArn() == null) {
+                return null; // identity session without mapped caller context — preserve historical bypass
+            }
             List<String> identityPolicies = collectRolePolicies(session.getRoleArn());
             String boundaryDoc = resolveRoleBoundaryDocument(session.getRoleArn());
             return new CallerContext(identityPolicies, session.getSessionPolicyDocument(), boundaryDoc);
@@ -898,6 +902,9 @@ public class IamService {
     }
 
     private String resolveRoleBoundaryDocument(String roleArn) {
+        if (roleArn == null) {
+            return null;
+        }
         String roleName = roleArn.contains("/") ? roleArn.substring(roleArn.lastIndexOf('/') + 1) : roleArn;
         return roles.get(roleName)
                 .map(IamRole::getPermissionsBoundaryArn)
@@ -984,6 +991,9 @@ public class IamService {
     }
 
     private List<String> collectRolePolicies(String roleArn) {
+        if (roleArn == null) {
+            return null;
+        }
         String roleName = roleArn.contains("/") ? roleArn.substring(roleArn.lastIndexOf('/') + 1) : roleArn;
         Optional<IamRole> roleOpt = roles.get(roleName);
         if (roleOpt.isEmpty()) {
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
@@ -805,7 +805,11 @@ public class IamService {
     // =========================================================================
 
     public Optional<String> findSecretKey(String accessKeyId) {
-        return accessKeys.get(accessKeyId).map(AccessKey::getSecretAccessKey);
+        Optional<String> fromAccessKey = accessKeys.get(accessKeyId).map(AccessKey::getSecretAccessKey);
+        if (fromAccessKey.isPresent()) {
+            return fromAccessKey;
+        }
+        return sessions.get(accessKeyId).map(SessionCredential::getSecretAccessKey);
     }
 
     // =========================================================================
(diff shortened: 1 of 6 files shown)


hectorvent:
**Core fix looks correct.** Verified against the STS `service-2.json` and the IAM identifiers doc: STS temporary keys use the `ASIA` prefix (long term keys use `AKIA`), so the new `findSecretKey` fallback that checks `sessions` after `accessKeys` is collision safe, and persisting the session secret is exactly what `RdsSigV4Validator` needs to recompute the HMAC. The accept/reject tests on the ASIA path cover it well, and ElastiCache rides the same `findSecretKey` path as you noted.

**One side effect to address: `GetSessionToken` now stores a session with a null `roleArn`.** That value is the first one of its kind in the `sessions` store, and it reaches IAM enforcement:

```
IamEnforcementFilter.filter
  -> resolveCallerPolicies(akid)
    -> resolveCallerContext(akid)        // finds the GetSessionToken session
      -> collectRolePolicies(session.getRoleArn())   // roleArn == null
```

Both `collectRolePolicies` and `resolveRoleBoundaryDocument` start with `roleArn.contains("/")`, so a null throws an NPE:

```java
private List<String> collectRolePolicies(String roleArn) {
    String roleName = roleArn.contains("/") ? ...   // NPE when roleArn == null
```

Before this PR, `GetSessionToken` keys hit the "unknown access key, bypass" branch and were allowed. With `iam.enforcement-enabled=true` they now 500 instead. It is off by default (so the suite stays green and most users are unaffected), but it is a real regression for the enforcement path.

Minimal fix that keeps RDS/ElastiCache validation working (the secret is still read by `findSecretKey`) while restoring the bypass:

```java
private List<String> collectRolePolicies(String roleArn) {
    if (roleArn == null) return null;   // identity session, bypass
    ...
}
private String resolveRoleBoundaryDocument(String roleArn) {
    if (roleArn == null) return null;
    ...
}
```

Semantically GetSessionToken credentials carry the calling IAM user's permissions (per the STS docs), so mapping them to the user's identity policies would be the fully correct behavior, but that is larger scope; the null guard is enough to make this PR safe. A small enforcement-on test would lock it in.

simlau:
Thanks for the detailed review and for tracing the enforcement path.

I added the minimal fix for this PR: `GetSessionToken`-style identity sessions with a null `roleArn` now bypass IAM enforcement resolution instead of reaching the role-policy path and throwing an NPE. I also added a regression test covering that path.

I agree the fully semantic behavior would be to resolve `GetSessionToken` credentials back to the issuing IAM user and evaluate that user policy context. I am leaving that as a separate follow-up so this PR stays focused on fixing STS session secret persistence for RDS/ElastiCache IAM token validation plus the enforcement regression.

hectorvent (approved):

PR #1271, opened 2026-06-10
fix(cloudmap): initialize-at-run-time for GraalVM native image

ousamabenyounes:
## What

`CloudMapService` (added in #1217) declares `private static final SecureRandom RANDOM = new SecureRandom()` (used by `randomId()` / `generateHostedZoneId()`) but was never registered for GraalVM run-time initialization.

A `static` `SecureRandom` field is initialized at **build** time, so native-image captures the instance — with a frozen seed — into the image heap and aborts:

```
Detected an instance of Random/SplittableRandom class in the image heap.
scanning root java.security.SecureRandom@5702ecf4: NativePRNG embedded in
    io.github.hectorvent.floci.services.cloudmap.CloudMapService.randomId(CloudMapService.java:543)
```

The last 3 nightly native builds on `main` failed for this reason (e.g. [run 27255364817](https://github.com/floci-io/floci/actions/runs/27255364817), arm64 + amd64).

## Fix

Add `CloudMapService` to `quarkus.native.additional-build-args → --initialize-at-run-time`, matching the existing pattern for `AcmService`, `SesService`, `Route53Service`, `KmsService`, `CloudFrontService`, `CognitoSrpHelper` and `CertificateGenerator` — every other class holding a static `SecureRandom`.

## Why CI was green but nightly was red

The failure only manifests in the **native** build, which runs in the nightly workflow — never in the JVM-based CI jobs. So this PR also adds a plain JUnit guard (`NativeImageRuntimeInitializationTest`) that runs in ordinary CI: it scans every main-source class for a static `Random`/`SecureRandom`/`SplittableRandom` field and asserts each is registered under `--initialize-at-run-time`. Any future service that reintroduces this mistake now fails in CI within seconds instead of hours later in the nightly.

## Test verification (RED → GREEN)

`./mvnw test -Dtest=NativeImageRuntimeInitializationTest`

**RED** — fix line removed from `application.yml`:

```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0 <<< FAILURE!
org.opentest4j.AssertionFailedError:
These classes declare a static Random/SecureRandom field but are NOT registered under
--initialize-at-run-time= in src/main/resources/application.yml. GraalVM native builds will fail
(instance captured in the image heap). Add an --initialize-at-run-time entry for each:
  io.github.hectorvent.floci.services.cloudmap.CloudMapService ==> expected: <true> but was: <false>
```

**GREEN** — fix in place (all 8 classes with a static SecureRandom field covered):

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -44,6 +44,7 @@ quarkus:
       - --initialize-at-run-time=io.github.hectorvent.floci.services.route53.Route53Service
       - --initialize-at-run-time=io.github.hectorvent.floci.services.kms.KmsService
       - --initialize-at-run-time=io.github.hectorvent.floci.services.cloudfront.CloudFrontService
+      - --initialize-at-run-time=io.github.hectorvent.floci.services.cloudmap.CloudMapService
 
 
 floci:
--- a/src/test/java/io/github/hectorvent/floci/config/NativeImageRuntimeInitializationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/config/NativeImageRuntimeInitializationTest.java
@@ -1,0 +1,102 @@
+package io.github.hectorvent.floci.config;
+
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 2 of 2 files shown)


hectorvent (approved):
Thanks @ousamabenyounes. 

Good Catch.

PR #1263, opened 2026-06-09
fix(ses): omit MailFromDomain/Status fields in GetEmailIdentity when no MAIL FROM is configured

kapoorp99:
## Summary

Closes #1245 

## Type of change

- [✅] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore

Root cause:
`buildFullIdentityResponse` in `SesController` was previously using an invalid sentinel value `"NOT_STARTED"` for identities with no MAIL FROM domain configured. This value does not exist in the official AWS `MailFromDomainStatus` enum. 

Changes made:
* **SesController.java (`buildFullIdentityResponse`)**: Updated the logic to keep the `MailFromAttributes` block intact but omit the two inner fields (`MailFromDomain` and `MailFromStatus`) when no MAIL FROM domain is configured. This aligns perfectly with verified, real-world AWS SES behavior.
* **SesIdentityAttributesV2IntegrationTest.java**: Updated the integration tests to assert that the `MailFromAttributes` block is present (non-null) in the response, but that its inner fields are properly absent when unconfigured.

## Checklist

- [✅] `./mvnw test` passes locally
- [✅] New or updated integration test added
- [✅] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

Issue #1245, opened 2026-06-08
[BUG] SES GetEmailIdentity returns "NOT_STARTED" — not in MailFromDomainStatus enum


okinaka:
## Service

SES (v2 REST JSON)

## AWS API Action

`GetEmailIdentity` — specifically the `MailFromAttributes.MailFromDomainStatus` field of the response.

## Expected behavior

> [!IMPORTANT]
> **Correction (2026-06-11):** the two expectations below turned out to be inaccurate — verified actual AWS behavior is different. See the **Correction: verified actual AWS behavior** section at the bottom; the original text is kept as-is for context.

The AWS Smithy IDL constrains `MailFromDomainStatus` to this enum:

```
PENDING | SUCCESS | FAILED | TEMPORARY_FAILURE
```

Reference: [`com.amazonaws.sesv2#MailFromDomainStatus`](https://github.com/aws/api-models-aws/blob/main/models/sesv2/service/2019-09-27/sesv2-2019-09-27.json) (search for the shape).

The `MailFromAttributes` member itself is optional on `GetEmailIdentityResponse`, so for an identity with no MAIL FROM domain configured AWS is expected to either:

- omit the block entirely, or
- include the block with `MailFromDomainStatus` set to `PENDING` (the implicit "not configured yet" value).

## Actual behavior

For an identity with no MAIL FROM domain configured, Floci returns a value **not in the enum**, `"NOT_STARTED"`:

```json
{
  "MailFromDomain": "",
  "MailFromDomainStatus": "NOT_STARTED",
  "BehaviorOnMxFailure": "USE_DEFAULT_VALUE"
}
```

Site: `src/main/java/io/github/hectorvent/floci/services/ses/SesController.java:1086-1087`:

```java
mailFromAttributes.put("MailFromDomainStatus",
        mailFromDomain == null ? "NOT_STARTED" : toV2Status(identity.getMailFromDomainStatus()));
```

The AWS Java SDK v2 marshaller falls back to `UNKNOWN_TO_SDK_VERSION` for unknown enum values, so no exception is raised — but every downstream `equals(MailFromDomainStatus.X)` check silently misses, which is the kind of drift that's hard to spot from compat tests alone.

### Suggested resolution

In order of cleanliness:

1. **Omit the `MailFromAttributes` block entirely when no MAIL FROM is configured** (matches the Smithy optionality, removes the need for a placeholder status):
   ```java
   if (identity.getMailFromDomain() != null) {
[...]

PR #1273, opened 2026-06-10
fix(ses): omit MailFromAttributes if MAIL FROM domain is not configured

iam-ssrivastav:
For an identity with no MAIL FROM domain configured, GetEmailIdentity response should omit the MailFromAttributes block entirely rather than returning a custom 'NOT_STARTED' status which causes AWS SDK unmarshalling errors.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesController.java
@@ -1080,13 +1080,13 @@ public class SesController {
 
         result.set("DkimAttributes", buildDkimAttributes(identity));
 
-        ObjectNode mailFromAttributes = result.putObject("MailFromAttributes");
         String mailFromDomain = identity.getMailFromDomain();
-        mailFromAttributes.put("MailFromDomain", mailFromDomain == null ? "" : mailFromDomain);
-        mailFromAttributes.put("MailFromDomainStatus",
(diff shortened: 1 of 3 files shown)


okinaka:
#1263

hectorvent (approved):

okinaka:
@hectorvent @iam-ssrivastav @kapoorp99 — heads-up that this merged change diverges from verified real AWS behavior, and the root cause is an inaccuracy in my original issue description, so I'd like to follow up with a corrective fix.

When I filed #1245, my "Expected behavior" section offered two options, the first being *"omit the `MailFromAttributes` block entirely."* That first suggestion was wrong. I later verified against real AWS (2026-06-10) and added a **Correction** section to #1245 (2026-06-11): for an identity with no MAIL FROM domain configured, AWS **keeps** the `MailFromAttributes` block and omits only the inner `MailFromDomain` / `MailFromDomainStatus` fields, leaving `BehaviorOnMxFailure`:

```json
{ "BehaviorOnMxFailure": "USE_DEFAULT_VALUE" }
```

This PR followed the original (incorrect) suggestion and now drops the whole block, so an SDK caller sees `mailFromAttributes() == null` against the emulator but non-null against real AWS.

#1263 (closed) had actually implemented the verified-correct shape — block kept, only the two inner fields omitted.

Since the inaccuracy originated in my issue, I'll take care of the corrective fix myself and open a PR to restore the AWS-faithful shape. Apologies for the confusion.

PR #1275, opened 2026-06-10
feat(secretsmanager): Support Filters and Pagination in BatchGetSecretValue

### LINKED ISSUE #1267: [FEAT] Support Filters for Secret Manager BatchGetSecretValue
## Service

Secret Manager

## API Action / Feature

BatchGetSecretValue

## AWS Documentation

https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_BatchGetSecretValue.html#SecretsManager-BatchGetSecretValue-request-Filters

## Why is this needed?

Secret Manager's `BatchGetSecretValue` currently [supports only `SecretIdList`](https://github.com/floci-io/floci/blob/1.5.23/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java#L70), and `Filters` are ignored . We'd like to make this work.

- [ ] Yes
- [x] No

iam-ssrivastav:
## Summary

Implement filter support and pagination for Secrets Manager `BatchGetSecretValue` and `ListSecrets` APIs.
Closes #1267

- Added a robust filter matching engine supporting keys: `name`, `description`, `tag-key`, `tag-value`, `primary-region`, and `all`.
- Handled case-sensitivity, prefix-matching, and negation rules (using the `!` prefix) per AWS Secrets Manager specifications.
- Implemented mutually exclusive parameter validation (`SecretIdList` vs `Filters`).
- Added pagination support using `MaxResults` and `NextToken` for filtered batch queries.
- Shared the filter engine with `ListSecrets` to add query filter capabilities.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

We verified this implementation against the AWS Secrets Manager API specification. Previously, local integration tests or clients using AWS Secrets Manager SDK to retrieve batch secrets with `Filters` instead of an explicit `SecretIdList` would fail or return unfiltered results.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java
@@ -98,7 +98,7 @@ public class SecretsManagerJsonHandler {
             int maxResults = request.has("MaxResults") ? request.path("MaxResults").asInt() : 20;
             if (maxResults < 1 || maxResults > 20) {
                 return Response.status(400)
-                        .entity(new AwsErrorResponse("ValidationException", "MaxResults must be between 1 and 20."))
+                        .entity(new AwsErrorResponse("InvalidParameterException", "MaxResults must be between 1 and 20."))
                         .build();
             }
 
--- a/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java
@@ -1,6 +1,7 @@
 package io.github.hectorvent.floci.services.secretsmanager;
 
 import io.github.hectorvent.floci.core.common.AwsErrorResponse;
+import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.services.secretsmanager.model.Secret;
 import io.github.hectorvent.floci.services.secretsmanager.model.SecretVersion;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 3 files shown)


iam-ssrivastav:
@hectorvent  can you please take a look.

hectorvent:
Nice work on this. The filter engine matches AWS closely: `name`/`tag-key`/`tag-value`/`primary-region` are case sensitive prefix matches, `description` is case insensitive, `!` negation and the value OR / filter AND logic all line up, and the `SecretIdList` vs `Filters` mutual exclusivity correctly returns `InvalidParameterException`.

One wire contract fix. The MaxResults range check throws `ValidationException`, but Secrets Manager has no such error code.

Verified against the SDK source in `local/aws`: `ValidationException` appears 0 times in the secretsmanager `service-2.json` (both aws-sdk-java-v2 and botocore), and aws-sdk-go-v2 `types/errors.go` has no `ValidationException` type. `BatchGetSecretValue`'s declared errors are `ResourceNotFoundException, InvalidParameterException, InvalidRequestException, DecryptionFailure, InternalServiceError, InvalidNextTokenException`. A client cannot map `ValidationException` to a typed exception.

```java
// current
new AwsErrorResponse("ValidationException", "MaxResults must be between 1 and 20.")

// AWS code (also the one this PR already uses for the SecretIdList/Filters check)
new AwsErrorResponse("InvalidParameterException", "MaxResults must be between 1 and 20.")
```

Two small non blocking notes:

1. `owning-service` is a valid `Filter.Key` enum value but the engine returns `false` for it. Fine to defer, just worth a comment so it is a known gap.
2. `NextToken` is a plain integer index rather than an opaque token. Works fine since clients echo it back, only noting it differs from real AWS.

Thanks for the contribution!

cia-rana:
Hi, @hectorvent !

Would it be possible for you to review this again?

hectorvent:
@cia-rana thanks for your patience. Please fix conflicts and tag me, I will try to get it merged.

hectorvent:
Thanks again @iam-ssrivastav, the implementation was solid and your follow up commit already addressed the review. To unblock the merge I rebased both of your commits onto current main (authorship preserved) and resolved the ListSecrets conflict, since main gained CreatedDate sorted pagination after this PR was written. The filters now apply first and the pagination runs on the filtered list.

Continuing in #1737, which supersedes this PR and closes #1267. Closing this one.

PR #1283, opened 2026-06-10
fix(docker): fall back to the default command when entrypoint argv is empty

### LINKED ISSUE #1279: docker-entrypoint.sh exits 0 when invoked with no arguments (breaks Testcontainers' LocalStackContainer)
## Problem

`docker/entrypoint.sh` ends with `exec "$@"`. When the entrypoint is invoked with an
**empty argv**, there is nothing to exec, the script reaches EOF and the container
exits with code 0 — no log output at all.

This is not a theoretical invocation: **testcontainers-java's `LocalStackContainer`
does exactly this**. To inject environment that depends on the mapped port (known only
after start), it replaces the entrypoint with a sleep-loop, then copies in a starter
script that exports those variables and finally re-execs
`/usr/local/bin/docker-entrypoint.sh` **with no arguments** (and the created container
has `Cmd=[]`). LocalStack's entrypoint ignores its argv and starts supervisord
regardless, so this contract has been implicitly relied on by the whole Testcontainers
ecosystem. On Floci the container silently dies, and every

```java
new LocalStackContainer(DockerImageName.parse("floci/floci:1.5.23-compat")
        .asCompatibleSubstituteFor("localstack/localstack"))
```

fails with `Wait strategy failed. Container exited with code 0`.

## Reproduction

```bash
# the image runs fine normally...
docker run -d -P floci/floci:1.5.23-compat            # serves :4566

# ...but with the entrypoint invoked argv-less (what the Testcontainers starter
# script does; --entrypoint also clears the image CMD):
docker run -d -P --entrypoint /usr/local/bin/docker-entrypoint.sh floci/floci:1.5.23-compat
# → container exits 0 within a second, zero log lines
```

Or run the Java snippet above with testcontainers `1.21.x`.

## Proposal

In `docker/entrypoint.sh`, immediately before the final `exec "$@"`, fall back to the
image's default command when argv is empty:

```sh
if [ $# -eq 0 ]; then
    if [ -x /app/application ]; then
        set -- /app/application -Dquarkus.http.host=0.0.0.0
    else
        set -- java -jar /app/quarkus-app/quarkus-run.jar -Dquarkus.http.host=0.0.0.0
    fi
fi
```

The two branches cover the two image families that share this entrypoint (native
images ship `/app/application`, JVM images ship the Quarkus runner jar), matching each
variant's `CMD`. Placing the fallback at the bottom of the script also covers the
root→gosu path, which re-enters the script via `exec gosu floci "$0" "$@"` with the
same empty argv. Explicit arguments keep working unchanged, and LocalStack drop-in
parity is preserved (LocalStack ignores argv entirely).

Fixing this in Floci helps all Testcontainers versions at once; the alternative
(making testcontainers-java's starter script append the image CMD) would only help
future Testcontainers releases.

## Aside (separate decision)

The Testcontainers starter script itself has a `#!/bin/bash` shebang. The published
native images (ubi9-minimal) ship bash, so they are fine once this fix lands — but the
`-jvm` images (alpine) have no bash, so stock Testcontainers still cannot start them.
If `-jvm` testcontainers support is desired, `apk add bash` in the JVM Dockerfiles
would complete the picture. Happy to file separately if you want it.

## Context

Found while migrating Cato Networks' service test suites (~36 Java modules) from
LocalStack to Floci — same investigation pipeline as #784
(`ApproximateFirstReceiveTimestamp`, fixed in 1.5.16). We currently work around this
with a `withCreateContainerCmdModifier` restoring the image entrypoint+cmd in every
fixture, and will delete those on the next release.

TomDav28:
Fixes #1279.

`exec "$@"` with an empty argv made the container exit 0 silently. Testcontainers'
`LocalStackContainer` invokes the entrypoint exactly that way (starter script,
`Cmd=[]`), so Floci could not be used as a LocalStack substitute from
testcontainers-java at all — the most common way JVM projects consume a LocalStack
image in CI.

**Change:** before the final `exec "$@"`, an empty argv now falls back to the image's
default command — `/app/application -Dquarkus.http.host=0.0.0.0` when the native
binary is present, otherwise the Quarkus runner jar (the JVM images' CMD). The
fallback sits at the bottom of the script so the gosu re-exec path (which re-enters
with `"$0" "$@"`) is covered by the same lines. Explicit argv (e.g. `docker run image
sh`) is untouched.

**Tests:** `docker/test-entrypoint.sh` (same style as `docker/test-localstack-parity.sh`,
run with `sh docker/test-entrypoint.sh`): explicit args pass through unchanged;
empty argv execs the JVM default when `/app/application` is absent; empty argv
prefers the native binary when present (skipped where `/app` is not writable —
covered by the image validation below).

**Validation (local images built from this branch):**
- `docker run --entrypoint /usr/local/bin/docker-entrypoint.sh <image>` (no args):
  before — exits 0 silently; after — serves :4566 and passes `/_floci/health`.
- Normal `docker run` / explicit-command runs unchanged.
- A stock testcontainers-java 1.21.4 `LocalStackContainer` with only
  `.asCompatibleSubstituteFor("localstack/localstack")` (no custom wait, no
  entrypoint modifier) starts in <2 s and passes an S3 put/get + SNS→SQS
  publish/receive round trip (together with the parity `Ready.` line PR).
- Full `mvn test` suite: unchanged vs. main on the same machine.

**Downstream:** Cato Networks (server-services, ~36 Testcontainers modules) pins
`floci/floci:1.5.23-compat` and carries a tagged
`withCreateContainerCmdModifier(...)`/"restore image entrypoint+cmd" workaround in
every SQS/S3/SNS fixture; those get deleted as soon as a release ships this. Same
contribution pipeline as #784.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -41,4 +41,20 @@ if [ "${LOCALSTACK_PARITY:-true}" != "false" ]; then
     . /usr/local/bin/localstack-parity.sh
 fi
 
+# Fall back to the image's default command when invoked with no arguments.
+# Some tooling re-executes this entrypoint directly with an empty argv —
+# Testcontainers' LocalStackContainer does exactly that via its starter
+# script — and `exec "$@"` with no argv would make the script reach EOF
+# and exit 0 without ever starting the emulator. LocalStack's entrypoint
+# ignores its argv entirely, so this keeps drop-in parity.
+# The default matches the CMD of the image variant: native images ship
+# /app/application, JVM images ship /app/quarkus-app/quarkus-run.jar.
+if [ $# -eq 0 ]; then
+    if [ -x /app/application ]; then
+        set -- /app/application -Dquarkus.http.host=0.0.0.0
+    else
+        set -- java -jar /app/quarkus-app/quarkus-run.jar -Dquarkus.http.host=0.0.0.0
+    fi
+fi
+
 exec "$@"
--- a/docker/test-entrypoint.sh
+++ b/docker/test-entrypoint.sh
@@ -1,0 +1,88 @@
+#!/bin/sh
+# Unit tests for entrypoint.sh argument handling.
+# Run directly: sh docker/test-entrypoint.sh
+# Exit 0 on success, non-zero on first failure summary.
+#
+# These tests run the entrypoint as an unprivileged user with
+# LOCALSTACK_PARITY=false, so the root-only gosu block and the parity
+# script (installed at an absolute path inside the image) stay out of
+# the way. The root/gosu path is covered by the Docker image tests.
+
+set -eu
+
+SCRIPT="$(cd "$(dirname "$0")" && pwd)/entrypoint.sh"
+PASS=0
+FAIL=0
+
+assert_eq() {
+    desc="$1"; expected="$2"; actual="$3"
+    if [ "${actual}" = "${expected}" ]; then
+        printf '[PASS] %s\n' "${desc}"
+        PASS=$((PASS + 1))
+    else
+        printf '[FAIL] %s\n  expected: %s\n  actual:   %s\n' "${desc}" "${expected}" "${actual}"
+        FAIL=$((FAIL + 1))
+    fi
+}
+
+if [ "$(id -u)" = '0' ]; then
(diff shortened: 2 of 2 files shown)


hectorvent (approved):
Thanks! Welcome to Floci.

PR #1284, opened 2026-06-10
feat(parity): emit LocalStack-compatible "Ready." log line on startup

### LINKED ISSUE #1280: Emit a LocalStack-parity "Ready." log line so default log-based wait strategies match
## Problem

LocalStack prints a line ending in `Ready.` when its gateway is up. A lot of ecosystem
tooling keys on that exact string — most prominently testcontainers-java's
`LocalStackContainer`, whose **default wait strategy** is the log regex `.*Ready\.\n`
(verified against `localstack-1.21.4.jar`). Floci's readiness banner is
`=== AWS Local Emulator Ready ===` — no `Ready.` substring — so the default wait can
never match and the container start times out after 60 s even when the emulator was
actually ready in milliseconds.

Together with the empty-argv entrypoint issue (#1279) this is what currently
makes `floci/floci` unusable as a drop-in `localstack/localstack` substitute under
testcontainers-java.

## Reproduction

```bash
docker run -d floci/floci:1.5.23-compat
sleep 2
docker logs <cid> 2>&1 | grep -c 'Ready\.$'     # → 0   (localstack/localstack → 1)
```

## Proposal

When LocalStack parity is enabled (`LOCALSTACK_PARITY` unset or != `false` — the same
gate `docker/entrypoint.sh` already uses), log one additional line consisting of
exactly `Ready.` right after the existing banner, from `EmulatorLifecycle`'s two
ready paths (the no-hooks fast path and the deferred init-hooks path). The Floci
banner stays untouched; with `LOCALSTACK_PARITY=false` nothing changes.

With the configured console format (`… %s%e%n`) the line ends with `Ready.`, which is
what the Testcontainers regex needs.

## Context

Found while migrating Cato Networks' Testcontainers suites from LocalStack to Floci
(same pipeline as #784). We work around it today with an explicit
`.waitingFor(Wait.forHttp("/"))` on every container — which we'll keep as good
practice, but out-of-the-box compatibility needs the parity log line.

TomDav28:
Fixes #1280.

testcontainers-java's `LocalStackContainer` default wait strategy polls the container
log for `.*Ready\.\n` — LocalStack's readiness line. Floci's banner never matches, so
every stock usage times out after 60 s.

**Change:** `EmulatorLifecycle` now logs a single extra line, exactly `Ready.`, right
after the existing `=== AWS Local Emulator Ready ===` banner, on both ready paths
(no-hooks fast path in `onStart`, deferred path in `onHttpStart`). It is gated on the
existing parity switch: a new `localstack.parity` config property (bound to the
`LOCALSTACK_PARITY` env var through the standard MicroProfile env mapping) defaulting
to enabled, with the same "anything but the literal `false`" semantics as
`docker/entrypoint.sh`. With `LOCALSTACK_PARITY=false` the line is not printed.

**Tests:** three new `EmulatorLifecycleTest` cases capture the lifecycle logger's
records: exactly one `Ready.` after the banner by default; none when parity is
disabled (banner still present); exactly one on the deferred init-hooks path.

**Validation (local image built from this branch):**
- `docker logs <cid> | grep -c 'Ready\.$'` → `1` with parity on, `0` with
  `LOCALSTACK_PARITY=false`; `localstack/localstack:4.14.0` → `1` (parity).
- Stock testcontainers-java 1.21.4 `LocalStackContainer` (only
  `.asCompatibleSubstituteFor`, default wait) becomes ready in <2 s and passes an
  S3 + SNS→SQS round trip (together with the entrypoint PR #1283).
- Full `mvn test` suite: failure set identical to main on the same machine.

**Downstream:** Cato Networks' suites currently carry explicit
`.waitingFor(Wait.forHttp("/"))` workarounds on ~36 fixtures; an HTTP wait stays
reasonable practice, but stock fixtures shouldn't need it.

Docs: one sentence added to the README parity section and the migration guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -692,7 +692,7 @@ LocalStack environment variables are translated automatically:
 | `LAMBDA_REMOVE_CONTAINERS=1` | `FLOCI_SERVICES_LAMBDA_EPHEMERAL=true` |
 | `DEBUG=1` | `QUARKUS_LOG_LEVEL=DEBUG` |
 
-Init scripts mounted under `/etc/localstack/init/` run unchanged. The `/_localstack/init` and `/_localstack/health` endpoints are still served. Set `LOCALSTACK_PARITY=false` to opt out of automatic translation.
+Init scripts mounted under `/etc/localstack/init/` run unchanged. The `/_localstack/init` and `/_localstack/health` endpoints are still served. Once the emulator is up, the log also ends with a LocalStack-style `Ready.` line, so tooling that watches the log for it — such as the default wait strategy of Testcontainers' `LocalStackContainer` — works unchanged. Set `LOCALSTACK_PARITY=false` to opt out of automatic translation.
 
 See the [full migration guide](https://floci.io/floci/getting-started/migrate-from-localstack/).
 
--- a/docs/getting-started/migrate-from-localstack.md
+++ b/docs/getting-started/migrate-from-localstack.md
@@ -23,6 +23,8 @@ services:
 
 Explicitly set Floci variables always win — the translation only fills in values that haven't been set. To disable the translation entirely, set `LOCALSTACK_PARITY=false`.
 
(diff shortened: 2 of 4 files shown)


hectorvent (approved):
Thanks @TomDav28, 

Solid feature

PR #1285, opened 2026-06-10
fix(sqs): report ApproximateNumberOfMessagesDelayed in GetQueueAttributes

### LINKED ISSUE #1281: SQS GetQueueAttributes omits ApproximateNumberOfMessagesDelayed
## Problem

`GetQueueAttributes` returns `ApproximateNumberOfMessages` and
`ApproximateNumberOfMessagesNotVisible` but never returns
`ApproximateNumberOfMessagesDelayed` — neither when requested explicitly nor under
`AttributeNames=All`. Real AWS and LocalStack always return all three counters.

Clients commonly read the trio with
`Integer.valueOf(attributes.get("ApproximateNumberOfMessagesDelayed"))` (queue-drain
helpers, test assertions, monitoring) and blow up with NPE/NumberFormatException on
the absent key.

A second-order effect of the same internal accounting: messages waiting out their
`DelaySeconds` are currently counted into `ApproximateNumberOfMessagesNotVisible`,
while AWS counts delayed messages **only** in the Delayed counter (NotVisible is
received-and-in-flight only).

## Reproduction

```bash
awslocal sqs create-queue --queue-name probe
awslocal sqs get-queue-attributes --queue-url <url> --attribute-names All
# → response has ApproximateNumberOfMessages + …NotVisible, no …Delayed key
awslocal sqs send-message --queue-url <url> --message-body x --delay-seconds 30
awslocal sqs get-queue-attributes --queue-url <url> --attribute-names All
# → still no …Delayed key (and the delayed message shows up in …NotVisible)
```

LocalStack (4.14.0) returns `"ApproximateNumberOfMessagesDelayed": "0"` on a fresh
queue, `"1"` while the message is delayed, `"0"` once it becomes visible.

## Proposal

`SqsService` already schedules `DelaySeconds` deliveries (a delayed message is one
with a future `visibleAt` that was never claimed, i.e. has no receipt handle).
Split the existing per-queue counters three-way — visible / in-flight / delayed —
and report `ApproximateNumberOfMessagesDelayed` ("0" when none) from
`GetQueueAttributes`, both for explicit attribute names and under `All`, on both the
Query and JSON protocol paths.

## Context

Found while migrating Cato Networks' cato-events Testcontainers suites to Floci
(same pipeline as #784): the shared queue-count helper inherited by ~36 consumer
test classes failed with `NumberFormatException: Cannot parse null string`. We
work around it with an absent-count=0 guard for now.

TomDav28:
Fixes #1281.

**Change:** `GuardedMessageQueue.messageCounts()` now classifies messages the way AWS
reports them: visible / in flight / delayed. A message that is not visible and was
never claimed (no receipt handle) can only be waiting out its `DelaySeconds`, so it
counts as delayed; claimed messages stay in flight. `GetQueueAttributes` returns
`ApproximateNumberOfMessagesDelayed` (always present, `"0"` when none) alongside the
two existing counters — automatically on both the Query and JSON 1.0 paths, since
both handlers serialize the service-level map. This also corrects
`ApproximateNumberOfMessagesNotVisible`, which previously absorbed delayed messages
(AWS counts them only in Delayed). `StartMessageMoveTask`'s to-move estimate keeps
its previous total (visible + in flight + delayed, i.e. everything the mover drains).

**Tests:**
- `GuardedMessageQueueTest`: three-way classification incl. delayed→visible
  transition and claimed-vs-delayed separation.
- `SqsServiceTest`: fresh queue reports `"0"`; `DelaySeconds=1` message reports
  Delayed=1 / NotVisible=0 / Messages=0 and flips to Messages=1 / Delayed=0 after the
  delay; an in-flight (received) message reports NotVisible=1 / Delayed=0; explicit
  `AttributeNames=[ApproximateNumberOfMessagesDelayed]` returns exactly that key.
- Wire-level: Query (`SqsIntegrationTest`) and JSON (`SqsJsonProtocolTest`) responses
  contain all three counters.

**Validation:** behavior matrix against `localstack/localstack:4.14.0` — identical
observable lifecycle (`absent-key` bug before; present/0 → 1 while delayed → 0 after,
on both emulators after the fix). Full `mvn test` suite: failure set identical to
main on the same machine. (Checked against 4.14.0 because `localstack:latest` now
exits with a license error without an auth token.)

**Downstream:** Cato Networks' cato-events test helper (inherited by ~36 consumer
test classes) NFE'd on the absent key; it now carries a tagged absent-count=0 guard
that becomes pure defense-in-depth once this ships. Same pipeline as #784.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueue.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueue.java
@@ -203,18 +203,32 @@ class GuardedMessageQueue {
         }
     }
 
-    record MessageCounts(long visible, long inFlight) {
+    record MessageCounts(long visible, long inFlight, long delayed) {
     }
 
+    /**
+     * Splits the queue contents the way AWS SQS reports them in
+     * GetQueueAttributes: visible (ApproximateNumberOfMessages), in flight
+     * (ApproximateNumberOfMessagesNotVisible) and delayed
+     * (ApproximateNumberOfMessagesDelayed). A message that is not visible and
+     * was never claimed — no receipt handle — can only be waiting out its
+     * DelaySeconds, so it counts as delayed rather than in flight.
+     */
     MessageCounts messageCounts() {
         try (var _ = hold()) {
             long visible = 0;
             long inFlight = 0;
+            long delayed = 0;
             for (Message m : messages) {
-                if (m.isVisible()) visible++;
-                else inFlight++;
+                if (m.isVisible()) {
+                    visible++;
+                } else if (m.getReceiptHandle() == null) {
+                    delayed++;
+                } else {
+                    inFlight++;
+                }
             }
-            return new MessageCounts(visible, inFlight);
+            return new MessageCounts(visible, inFlight, delayed);
         }
     }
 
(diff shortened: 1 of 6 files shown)


hectorvent (approved):
Thanks @TomDav28

PR #1291, opened 2026-06-11
feat(ec2): support Spot Instance request actions

### LINKED ISSUE #1287: [FEAT] EC2 Spot Instance Support
## Service

EC2

## API Action / Feature

Creating/Managing Spot Instances

## AWS Documentation

For creating the spot instance request https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SpotInstanceRequest.html

For monitoring the status of the spot instance request https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSpotInstanceRequests.html

## Why is this needed?

Currently you support launching EC2 instances, but you don't offer support for spot instances. Since floci doesn't actually connect to AWS this could theoretically use the same logic as creating an instance behind the scenes.

- [ ] Yes
- [x] No

mansiverma897993:
- Implement requestSpotInstances, describeSpotInstanceRequests, and cancelSpotInstanceRequests in Ec2Service

- Implement RequestSpotInstances, DescribeSpotInstanceRequests, and CancelSpotInstanceRequests XML serialization and request routing in Ec2QueryHandler

- Add LaunchSpecification and SpotInstanceRequest models

- Add spotInstanceLifecycle integration test to cover the full lifecycle

Fixes #1287 

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist


RomanJobRunr:
This does look like what it should work for what I requested, I don't see where the definition for the method in the Ec2Client itself is created for for example `ec2Client.requestSpotInstances` beyond `case "RequestSpotInstances" -> handleRequestSpotInstances(params, region);` in the query handler and the Ec2Service getting `requestSpotInstances` but I guess this is done automatically?

mansiverma897993:
@RomanJobRunr Yes, this is handled automatically! 

Floci is a local emulator for AWS services, so it doesn't define or bundle client SDKs itself. Instead, it relies directly on the official AWS SDK clients (like `software.amazon.awssdk.services.ec2.Ec2Client` in Java, `boto3` in Python, or `@aws-sdk/client-ec2` in Node.js) with their endpoints overridden to the emulator.

Since these official clients already have the standard methods like `requestSpotInstances`, `describeSpotInstanceRequests`, and `cancelSpotInstanceRequests` defined by AWS, they work out of the box on the client side once they are routed and handled on the server side in Floci's `Ec2QueryHandler` and `Ec2Service`.

mansiverma897993:
Let me know if still any change required ??

RomanJobRunr:
It looks ok to me

mansiverma897993:
@hectorvent what about yours views?

hectorvent:
Really nice work, the lifecycle and the XML wire shapes look great. Element casing (`spotInstanceRequestId`, `spotPrice`, `launchSpecification`, `groupSet`, `userData`, `tagSet`, etc.), the `spotInstanceRequestSet`/`item` list pattern, the base64 user data round trip, and the `active`/`fulfilled` and `cancelled` states all match the EC2 model.

One parity fix on the tag `ResourceType` value. The handler (and the test) use `spot-instance-request`, but AWS uses the plural `spot-instances-request`.

Verified against the SDK source in `local/aws`:

- aws-sdk-java-v2 ec2 `service-2.json` `ResourceType` enum contains `spot-instances-request` (the singular form is not present)
- botocore ec2 model: same, only `spot-instances-request`
- aws-sdk-go-v2 `types/enums.go`: `ResourceTypeSpotInstancesRequest = "spot-instances-request"`

The `RequestSpotInstances` doc is explicit too: "The value for ResourceType must be `spot-instances-request`, otherwise the Spot Instance request fails."

So a real SDK or CLI client sends the plural form, the `if` below is false, and the spot request tags are silently dropped (they never show up in the describe `tagSet`):

```java
// current
if ("spot-instance-request".equals(resType)) {

// AWS value
if ("spot-instances-request".equals(resType)) {
```

The integration test passes only because it sends the same singular value, so it is worth updating both:

```java
.formParam("TagSpecification.1.ResourceType", "spot-instances-request")
```

The `instance` branch for instance level tags is already correct.

Small non blocking notes: `DescribeSpotInstanceRequests` does not implement `NextToken`/`MaxResults` pagination yet (fine to defer), and it is worth defaulting `productDescription` to something like `Linux/UNIX` rather than empty since AWS always returns a value.

Thanks for the contribution!

mansiverma897993:
@hectorvent now check at once !!

hectorvent:
@mansiverma897993 please resolve conflicts when you have a chance?

mansiverma897993:
@hectorvent now check at once ?

RomanJobRunr:
I will also confirm that I have now tested this branch with my implementation of spot instance requests/cancellations and it works as expected.

hectorvent (approved):

hectorvent:
@mansiverma897993 will need another conflict solve. I will the to get it merged as soon as conflicts get resolved. 

Thanks!

mansiverma897993:
@hectorvent now I have resloved all merge conflicts !!check now

PR #1295, opened 2026-06-11
fix(rds): align subnet group placement handling with AWS behavior

### LINKED ISSUE #1276: [BUG] rds: CreateDBInstance/CreateDBCluster do not validate DB subnet group or availability zone
## Service

RDS

## AWS API Action

CreateDBInstance, CreateDBCluster, DescribeDBInstances, DescribeDBClusters

## Expected behavior

When creating RDS instances or clusters, Floci should validate network-placement inputs in a way that is materially compatible with AWS.

This includes:

- accepting and resolving `DBSubnetGroupName`
- validating that referenced subnets exist
- validating that subnet/VPC/AZ relationships are coherent
- validating `AvailabilityZone` against the selected placement
- validating `MultiAZ` constraints where applicable

`DescribeDBInstances` and `DescribeDBClusters` should return subnet-group and availability-zone data derived from the created resource, rather than fixed placeholder values.

## Actual behavior

Floci currently does not implement subnet or availability-zone validation for RDS creation.

Observed implementation behavior:

- `CreateDBInstance` does not read or validate `DBSubnetGroupName`, `AvailabilityZone`, or `MultiAZ`
- RDS creation does not consult EC2 subnet state
- `DescribeDBInstances` returns a hard-coded `DBSubnetGroup` structure with placeholder values such as `default`, `vpc-00000000`, and `subnet-00000000`
- `DescribeDBInstances` returns `AvailabilityZone` from `config.defaultAvailabilityZone()`
- `DescribeDBClusters` returns hard-coded `DBSubnetGroup=default`
- `DescribeDBClusters` returns `AvailabilityZone` from `config.defaultAvailabilityZone()`
- `MultiAZ` is always returned as `false`

This means IaC or SDK flows that expect subnet/AZ validation or accurate describe responses can succeed incorrectly or observe non-AWS-faithful metadata.

## Reproduction

```bash
cd /home/dev/backend/floci

# Start Floci
docker compose up -d

# Attempt to create an RDS instance with explicit placement inputs
aws --endpoint-url http://localhost:4566 rds create-db-instance \
  --db-instance-identifier mydb \
  --db-instance-class db.t3.micro \
  --engine postgres \
  --master-username admin \
  --master-user-password secret123 \
  --allocated-storage 20 \
  --availability-zone ap-northeast-1a \
  --db-subnet-group-name my-subnet-group \
  --multi-az

# Inspect returned placement metadata
aws --endpoint-url http://localhost:4566 rds describe-db-instances \
  --db-instance-identifier mydb
```

## Environment

- Floci version / image tag: `1.5.20`
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (devcontainer)

shrimptails-f:
## Summary

Align RDS placement handling more closely with AWS for subnet-group-based creation flows.
Closes #1276

- read and resolve `DBSubnetGroupName` for `CreateDBInstance` and `CreateDBCluster`
- validate referenced subnets through EC2 subnet state
- validate subnet/VPC/AZ coherence and `AvailabilityZone` / `MultiAZ` constraints
- return placement-derived subnet-group and availability-zone data from `DescribeDBInstances` and `DescribeDBClusters`
- add minimal `CreateDBSubnetGroup` / `DescribeDBSubnetGroups` support needed for custom subnet-group flows

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, RDS create flows ignored `DBSubnetGroupName`, `AvailabilityZone`, and `MultiAZ`, and describe flows returned hard-coded placeholder placement metadata. This change validates placement inputs against EC2 subnet state and returns placement data derived from the created resource instead of fixed defaults.

## Checklist


hectorvent:
## AWS wire protocol review

Thanks for this, the placement modeling and test coverage look great. The default VPC dependency is fine too since `describeSubnets` lazily seeds the default VPC and subnets, so standalone creates do not regress. Two items worth fixing before merge since both sit right in this PR's scope, plus a couple of optional notes.

### 1. `DBSubnetGroupAlreadyExistsFault` is the wrong wire code

`createDbSubnetGroup` throws `DBSubnetGroupAlreadyExistsFault`, but the model's error `code` for that shape is `DBSubnetGroupAlreadyExists` (no `Fault` suffix). SDK clients map typed exceptions off the `code` trait, so the current value will not resolve to `DbSubnetGroupAlreadyExistsException`. Your other new codes match exactly, this is the only one that drifts.

```java
throw new AwsException("DBSubnetGroupAlreadyExists",
        "DB subnet group " + name + " already exists.", 400);
```

### 2. `DescribeDBInstances` renders `Subnets` with `<member>` instead of `<Subnet>`

The model's `SubnetList` member carries `locationName: "Subnet"`, so the wire shape is `<Subnets><Subnet>...</Subnet></Subnets>`. The new `dbSubnetGroupInnerXml` already does this correctly, but the `dbInstanceInnerXml` block still emits `<member>`. An SDK deserializing `DescribeDBInstances` looks for `Subnet`, so the per subnet AZ data this PR adds will come back empty, which works against the goal of surfacing placement data from describe.

```java
xml.start("Subnet")   // not "member"
   .elem("SubnetIdentifier", entry.getKey())
   .start("SubnetAvailabilityZone").elem("Name", entry.getValue()).end("SubnetAvailabilityZone")
   .elem("SubnetStatus", "Active")
 .end("Subnet");
```

### Optional

- `DBSubnetGroupArn` is part of the `DBSubnetGroup` shape and AWS always returns it. Currently omitted from the create, describe, and instance responses.
- Missing required parameters (`DBSubnetGroupName`, `SubnetIds`) return `InvalidParameterValue`; AWS query services use `MissingParameter` for absent params. Cosmetic.

Everything else lines up with the model. `DBSubnetGroupNotFoundFault` (404), `InvalidVPCNetworkStateFault` (400), and `InvalidSubnet` (400) all match code and status, and rendering `DBCluster.DBSubnetGroup` as a plain string is correct.

Verified against `service-2.json`.

shrimptails-f:
@hectorvent 
Thanks, fixed both required wire mismatches and picked up the optional subnet-group follow-ups in this PR.

createDbSubnetGroup now returns DBSubnetGroupAlreadyExists without the Fault suffix, and DescribeDBInstances now renders DBSubnetGroup.Subnets with <Subnet> entries so SDKs can deserialize the AZ data correctly. I also added DBSubnetGroupArn to the subnet-group shape responses and switched missing CreateDBSubnetGroup params to MissingParameter.

Verified with targeted RDS tests.

hectorvent:
Nice work, the placement modeling looks solid. Checked the error codes and shapes against the RDS [`service-2.json`](https://github.com/aws/aws-sdk-java-v2/blob/master/services/rds/src/main/resources/codegen-resources/service-2.json) model and they hold up: `DBSubnetGroupAlreadyExists` (the wire code does drop the `Fault` suffix), `DBSubnetGroupNotFoundFault` at 404, `InvalidSubnet`, `InvalidVPCNetworkStateFault`, plus the `DBSubnetGroup`/`Subnet` response shapes. Switching to `MissingParameter` is the right call too.

One thing I'd fix before merge. In `dbInstanceInnerXml` this local is never read:

```java
DbSubnetGroup subnetGroup = service.resolveDbSubnetGroupView(i.getDbSubnetGroupName());
```

The XML is still rendered from `dbSubnetGroupForInstance(i)`, so it's dead. It's not harmless though: `resolveDbSubnetGroupView` throws `DBSubnetGroupNotFoundFault` for an unknown group, so a `DescribeDBInstances` on an instance whose subnet group was deleted would 404, which is exactly the case `dbSubnetGroupForInstance` swallows on purpose. Tests stay green only because `setUp` stubs the call. Dropping the line fixes it.

Two smaller things, fine to leave for later:

- The MultiAZ coverage check throws `InvalidVPCNetworkStateFault`. AWS has a dedicated code for this, `DBSubnetGroupDoesNotCoverEnoughAZs` (also declared on both create ops). Both are valid, the specific one is just more faithful.
- `AvailabilityZone` + `MultiAZ=true` together is accepted, but real AWS rejects that combo with `InvalidParameterCombination`.

Verified against the RDS [`service-2.json`](https://github.com/aws/aws-sdk-java-v2/blob/master/services/rds/src/main/resources/codegen-resources/service-2.json). Happy to approve once the dead local is gone.

shrimptails-f:
@hectorvent 
Thanks, fixed.

- removed the dead `resolveDbSubnetGroupView(...)` call from `dbInstanceInnerXml`, so `DescribeDBInstances` no longer fails with `DBSubnetGroupNotFoundFault` when the instance points at a deleted subnet group
- switched the Multi-AZ subnet-group coverage error to `DBSubnetGroupDoesNotCoverEnoughAZs`
- rejected `AvailabilityZone` + `MultiAZ=true` with `InvalidParameterCombination`

I also added coverage for the two validation cases in `RdsServiceTest`.

Validation:
- `./mvnw -Dtest=RdsServiceTest,RdsQueryHandlerTest test`
- `./mvnw test`

PR #1299, opened 2026-06-11
fix(ecs): translate portMappings into Docker PortBindings for bridge mode

### LINKED ISSUE #1270: [BUG] ECS task portMappings are ignored when creating Docker containers
## Summary

When running an ECS service in Floci, containers are launched successfully and the application starts correctly, but ECS `portMappings` are not translated into Docker port bindings.

As a result, the container is running but is not accessible from the host machine.

## Environment

* Floci: latest
* Docker Desktop
* macOS
* ECS network mode: `bridge`

## ECS Task Definition

```json
{
  "family": "employee-app",
  "networkMode": "bridge",
  "containerDefinitions": [
    {
      "name": "employee-app",
      "image": "000000000000.dkr.ecr.us-east-1.localhost:5000/employee-app:latest",
      "cpu": 256,
      "memory": 512,
      "essential": true,
      "portMappings": [
        {
          "containerPort": 5000,
          "hostPort": 5000,
          "protocol": "tcp"
        }
      ]
    }
  ]
}
```

## Steps to Reproduce

1. Create ECR repository.
2. Push Docker image.
3. Create ECS cluster.
4. Register task definition with:

```json
{
  "containerPort": 5000,
  "hostPort": 5000
}
```

5. Create ECS service.
6. Verify task reaches RUNNING state.

## Expected Behavior

Docker container should be created with host port bindings:

```bash
docker ps
```

Expected:

```text
0.0.0.0:5000->5000/tcp
```

Docker inspect should show:

```json
"Ports": {
  "5000/tcp": [
    {
      "HostIp": "0.0.0.0",
      "HostPort": "5000"
    }
  ]
}
```

Application should be reachable at:

```text
http://localhost:5000
```

## Actual Behavior

Container starts successfully:

```bash
docker ps
```

Shows:

```text
5000/tcp
```

instead of:

```text
0.0.0.0:5000->5000/tcp
```

Docker inspect shows:

```json
"Ports": {
  "5000/tcp": null
}
```

The Flask/Gunicorn application is running and reachable inside the container, but not from the host.

Container logs:

```text
[INFO] Listening at: http://0.0.0.0:5000
```

## Additional Verification

Running the same image manually works:

```bash
docker run -p 5000:5000 employee-app
```

The application becomes accessible at:

```text
http://localhost:5000
```

This suggests ECS task `portMappings` are not being translated into Docker `PortBindings` when Floci launches the container.

Aafiyanawab:
## Problem
Fixes #1270

ECS task portMappings were not being translated into Docker PortBindings
when Floci launches containers in native mode. The container started 
successfully but was unreachable from the host.

## Root Cause
In EcsContainerManager.java, the port mapping loop used withDynamicPort()
which assigns a random host port and ignores the hostPort value defined
in the task definition. The pm.hostPort() value was never passed to Docker.

## Fix
Replaced withDynamicPort() with withPortBinding(containerPort, hostPort),
using the hostPort from the task definition. Falls back to containerPort
if hostPort is 0 (not set).

## Before
docker ps showed: 5000/tcp (no host binding)

## After  
docker ps shows: 0.0.0.0:5000->5000/tcp ✅

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/container/EcsContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/container/EcsContainerManager.java
@@ -117,7 +117,8 @@ public class EcsContainerManager {
                 boolean publishToHost = !containerDetector.isRunningInContainer();
                 for (PortMapping pm : def.getPortMappings()) {
                     if (publishToHost) {
-                        specBuilder.withDynamicPort(pm.containerPort());
+                        int hostPort = pm.hostPort() > 0 ? pm.hostPort() : pm.containerPort();
+                        specBuilder.withPortBinding(hostPort, pm.containerPort());
                     } else {
                         specBuilder.withExposedPort(pm.containerPort());
                     }


hectorvent:
## Review

Thanks for tracking this down, the diagnosis is right and the direction is correct. One blocking issue though: the arguments to `withPortBinding` are reversed.

The builder signature is container port first:

```java
public Builder withPortBinding(int containerPort, int hostPort) {
    this.portBindings.put(containerPort, hostPort);
    this.exposedPorts.add(containerPort);
    return this;
}
```

The fix passes them host port first:

```java
int hostPort = pm.hostPort() > 0 ? pm.hostPort() : pm.containerPort();
specBuilder.withPortBinding(hostPort, pm.containerPort());
```

This only works when `hostPort == containerPort`, which is the `5000->5000` case in the description and the awsvpc default, so the manual test passed by coincidence. For an asymmetric bridge mapping (container `8080`, host `80`) it produces the reversed binding `0.0.0.0:8080->80/tcp` and exposes port `80`, leaving the app on `8080` unreachable, which is the same symptom #1270 is fixing.

Container port first fixes it:

```java
int hostPort = pm.hostPort() > 0 ? pm.hostPort() : pm.containerPort();
specBuilder.withPortBinding(pm.containerPort(), hostPort);
```

### Minor

`hostPort == 0` now binds `containerPort -> containerPort` statically. That is predictable for local use, but in real ECS bridge mode `hostPort: 0` means a dynamic host port, so two tasks sharing a `containerPort` with `hostPort: 0` will collide on the host where dynamic allocation would not. Reasonable tradeoff, just worth a short comment.

### Test

A small test with `hostPort != containerPort` asserting the resulting binding would catch the swap and guard against regressions, since the symmetric case hides it.

One line to fix, happy to re-review once it is in.

hectorvent:
Closing PR because inactivity. A new PR will be open.

Issue #1127, opened 2026-06-02
[BUG] Athena.DeleteWorkGroup - Action DeleteWorkGroup is not supported


MicDuff1:
## Service

Athena

## AWS API Action

[Athena.DeleteWorkGroup](https://docs.aws.amazon.com/athena/latest/APIReference/API_DeleteWorkGroup.html)

## Expected behavior

HTTP 200 response with an empty HTTP body

## Actual behavior

`An error occurred (InvalidAction) when calling the DeleteWorkGroup operation: Action DeleteWorkGroup is not supported`

## Reproduction

```bash
aws athena delete-work-group --work-group "athena-workgroup-sample" --endpoint-url http://localhost:4566 --region eu-central-1
```

## Environment

- Floci version / image tag: 1.5.21
- Java SDK version (if applicable): 17.0.11
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker

alhasan-mridha:
@hectorvent I would like to work on this. Could you please assign this to me?

mansiverma897993:
@MicDuff1 I have raised PR on this issue please review it and let me any chnages required!!

alhasan-mridha:
@hectorvent Sorry I was late. Since there is already an open PR for this issue, please feel free to reassign it. I’ll find another issue to contribute to.

mansiverma897993:
@hectorvent I have update the code can you plz check at once ? and Leeme know if any chnage required?

MicDuff1:
Thanks guys and thanks to @mansiverma897993 and sorry for the late reply, I was on holiday. 
I see, the fix is noted in release notes v1.5.24 and it works executing

1) ```aws athena create-work-group --name "athena-workgroup-sample" --endpoint-url http://localhost:4566 --region eu-central-1```
2) ```aws athena get-work-group --work-group "athena-workgroup-sample" --endpoint-url http://localhost:4566 --region eu-central-1``` --> response
3) ```aws athena delete-work-group --work-group "athena-workgroup-sample" --endpoint-url http://localhost:4566 --region eu-central-1```
4) ```aws athena get-work-group --work-group "athena-workgroup-sample" --endpoint-url http://localhost:4566 --region eu-central-1``` --> InvalidRequestException - WorkGroup athena-workgroup-sample is not found

So the issue can be closed.

hectorvent:
Thanks @MicDuff1 @mansiverma897993, 
Closing issue

PR #1301, opened 2026-06-12
fix(athena): 🐛 align GetWorkGroup with AWS behavior

### LINKED ISSUE #1128: [BUG] Athena.GetWorkGroup - Action GetWorkGroup is not supported
## Service

Athena

## AWS API Action

[Athena.GetWorkGroup](https://docs.aws.amazon.com/athena/latest/APIReference/API_GetWorkGroup.html)

## Expected behavior

HTTP 200 response and Workgroup information as JSON, see AWS documentation link above.

## Actual behavior

`An error occurred (InvalidAction) when calling the GetWorkGroup operation: Action GetWorkGroup is not supported`

## Reproduction

```bash
aws athena get-work-group --work-group "athena-workgroup-sample" --endpoint-url http://localhost:4566 --region eu-central-1 
```

## Environment

- Floci version / image tag: 1.5.21
- Java SDK version (if applicable): 17.0.11

LorenzoGalassi:
## Summary

Closes #1128

`GetWorkGroup` previously returned a hardcoded default response for any workgroup name: it ignored everything stored by `CreateWorkGroup` (description, configuration, creation time) and returned `ENABLED` plus default configuration even for workgroups that do not exist.

This PR makes `GetWorkGroup` read real state, scoped by region:

- created workgroups return their stored `Name`, `State`, `Description`, `CreationTime` and `Configuration`
- unknown workgroups return `InvalidRequestException` (400) with the AWS message `WorkGroup <name> is not found.`
- the `primary` workgroup keeps its default view

It also completes `DeleteWorkGroup` (#1127 / PR #1213), which validated input but never removed the entry from the store. This was invisible while `GetWorkGroup` was hardcoded, but it already broke the create -> delete -> recreate cycle (`WorkGroup already exists`). Deletion now actually removes the stored entry while keeping the existing lenient 200 for non-existent names, so all tests added in #1213 still pass unchanged. `DeleteWorkGroup` is also added to the supported actions table in the docs.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `aws athena get-work-group --work-group <created-wg>` returned default configuration instead of the stored one, and `--work-group <missing-wg>` returned 200 instead of `InvalidRequestException`. Real AWS returns the stored workgroup or fails with `WorkGroup <name> is not found.`

## Checklist

- [x] `./mvnw test` passes locally (all Athena suites pass; the only failing classes are 16 Docker-dependent suites that fail identically on an unmodified `main` checkout on this machine, which has no Docker available)
- [x] New or updated integration test added (`AthenaGetWorkGroupIntegrationTest`, 4 tests)

### DIFF
--- a/docs/services/athena.md
+++ b/docs/services/athena.md
@@ -16,6 +16,7 @@ Floci emulates Amazon Athena with **real SQL execution** powered by a [floci-duc
 | `StopQueryExecution` | Cancels a running query |
 | `CreateWorkGroup` | Creates a new workgroup |
 | `GetWorkGroup` | Returns information about a workgroup |
+| `DeleteWorkGroup` | Deletes a workgroup |
 | `ListWorkGroups` | Lists all workgroups |
 
 ## How it works
--- a/src/main/java/io/github/hectorvent/floci/services/athena/AthenaJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/athena/AthenaJsonHandler.java
@@ -66,7 +66,7 @@ public class AthenaJsonHandler {
             }
             case "GetWorkGroup" -> {
                 String name = request.has("WorkGroup") ? request.get("WorkGroup").asText() : "primary";
-                yield Response.ok(Map.of("WorkGroup", athenaService.getWorkGroup(name))).build();
+                yield Response.ok(Map.of("WorkGroup", athenaService.getWorkGroup(name, region))).build();
             }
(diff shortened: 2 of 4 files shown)


hectorvent (approved):
Thanks @LorenzoGalassi 
I appreciate your contribution.

PR #1302, opened 2026-06-12
fix(sqs): 🐛 report ApproximateNumberOfMessagesDelayed in queue attributes

### LINKED ISSUE #1281: SQS GetQueueAttributes omits ApproximateNumberOfMessagesDelayed

LorenzoGalassi:
## Summary

Closes #1281

`GetQueueAttributes` never returned `ApproximateNumberOfMessagesDelayed`, neither when requested explicitly nor under `AttributeNames=All`. On top of that, messages waiting out their `DelaySeconds` were counted into `ApproximateNumberOfMessagesNotVisible`, while AWS counts delayed messages only in the Delayed counter (NotVisible is received-and-in-flight only).

Changes:

- `GuardedMessageQueue.messageCounts()` now distinguishes three states: visible, in flight (invisible with a receipt handle, i.e. claimed) and delayed (invisible and never claimed, i.e. waiting out `DelaySeconds`)
- `GetQueueAttributes` always returns `ApproximateNumberOfMessagesDelayed` alongside the other two counters, matching AWS and LocalStack ("0" on a fresh queue, "1" while a message is delayed, back to "0" once visible)
- delayed messages no longer inflate `ApproximateNumberOfMessagesNotVisible`
- the message move task total still includes delayed messages, so redrive counts are unchanged

Both the Query and JSON protocol paths share `SqsService.getQueueAttributes`, so the attribute is returned on both.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: clients reading the standard counter trio (queue drain helpers, test assertions, monitoring) hit NPE / NumberFormatException on the absent `ApproximateNumberOfMessagesDelayed` key, and delayed messages showed up under NotVisible. Real AWS and LocalStack always return all three counters.

## Checklist

- [x] `./mvnw test` passes locally (all SQS suites pass, 141 tests; the only failing classes are 16 Docker-dependent suites that fail identically on an unmodified `main` checkout on this machine, which has no Docker available)
- [x] New or updated integration test added (unit coverage in `GuardedMessageQueueTest` and `SqsServiceTest` for the delayed counter, including the explicit attribute-name filter path)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueue.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueue.java
@@ -203,18 +203,25 @@ class GuardedMessageQueue {
         }
     }
 
-    record MessageCounts(long visible, long inFlight) {
+    record MessageCounts(long visible, long inFlight, long delayed) {
     }
 
     MessageCounts messageCounts() {
         try (var _ = hold()) {
             long visible = 0;
             long inFlight = 0;
+            long delayed = 0;
             for (Message m : messages) {
-                if (m.isVisible()) visible++;
-                else inFlight++;
+                if (m.isVisible()) {
+                    visible++;
+                } else if (m.getReceiptHandle() == null) {
+                    // Invisible but never claimed: waiting out its DelaySeconds
+                    delayed++;
+                } else {
+                    inFlight++;
+                }
             }
-            return new MessageCounts(visible, inFlight);
+            return new MessageCounts(visible, inFlight, delayed);
         }
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/SqsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/SqsService.java
@@ -321,6 +321,7 @@ public class SqsService {
         var counts = getOrCreateQueue(storageKey).messageCounts();
         attrs.put("ApproximateNumberOfMessages", String.valueOf(counts.visible()));
         attrs.put("ApproximateNumberOfMessagesNotVisible", String.valueOf(counts.inFlight()));
+        attrs.put("ApproximateNumberOfMessagesDelayed", String.valueOf(counts.delayed()));
 
         if (attributeNames == null || attributeNames.contains("All")) {
             return attrs;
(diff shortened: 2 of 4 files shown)


hectorvent:
Closing in favor of https://github.com/floci-io/floci/pull/1285

hectorvent:
Duplicate of #1285, which carries identical production code plus broader protocol-path tests. 

Thanks @LorenzoGalassi!

Issue #996, opened 2026-05-22
[BUG]  AppConfig: Call for listHostedConfigurationVersions fall into S3 bucket


hsales-intl:
## Service

AppConfig > listHostedConfigurationVersions

## AWS API Action

Using boto3, call for listHostedConfigurationVersions raises a 404

The GET without a version number at the end (LIST route for listHostedConfigurationVersions) has no matching JAX-RS route. Unmatched paths fall through to S3Controller's catch-all @Path("/{bucket}/{key:.+}")

## Expected behavior

Output expected:
{
    'Items': [
        {
            'ApplicationId': '339ohji',
            'ConfigurationProfileId': 'ur8hx2f',
            'ContentType': 'application/json',
            'VersionNumber': 1,
        },
    ],
    'ResponseMetadata': {
        '...': '...',
    },
}

## Actual behavior

{ClientError}ClientError('An error occurred (404) when calling the ListHostedConfigurationVersions operation: <?xml version="1.0" encoding=...de><Message>The specified bucket does not exist.</Message><RequestId>7ae39b41...</RequestId></Error>')

## Reproduction

```
#boto3
history = config_client.list_hosted_configuration_versions(
      ApplicationId=app_id,
      ConfigurationProfileId=profile_id,
      MaxResults=1,
  )
```

## Environment

- Floci latest docker image

PR #1317, opened 2026-06-13
fix(appconfig): 🐛 stop AppConfig route hijacking S3 bucket named "configuration"

### LINKED ISSUE #1294: S3 bucket named "configuration" is hijacked by the AppConfig route → NullPointerException / HTTP 500
**Summary**

Any S3 request against a bucket named exactly `configuration` is mis-dispatched to the AppConfig controller (which binds the path `/configuration`) instead of the S3 service. The AppConfig handler then throws a `NullPointerException` on a key it never stored, returning **HTTP 500**. This happens even for a freshly created, empty bucket — it is name-specific, not state-dependent.

**Version:** `floci/floci:1.5.24-compat`

**Minimal repro**

```bash
docker run -d -p 4566:4566 floci/floci:1.5.24-compat

aws --endpoint-url=http://localhost:4566 s3api create-bucket   --bucket configuration   # 200 OK
aws --endpoint-url=http://localhost:4566 s3api list-objects-v2 --bucket configuration   # HTTP 500

# control — any other name is fine:
aws --endpoint-url=http://localhost:4566 s3api create-bucket   --bucket cato-configuration
aws --endpoint-url=http://localhost:4566 s3api list-objects-v2 --bucket cato-configuration  # 200, empty
```

Confirmed working names: `cato-configuration`, `appconfig`, `secretsmanager`, `kms`, `lambda`. Only the literal `configuration` collides.

**Stack trace (the 500)**

```
java.lang.NullPointerException
  at java.util.concurrent.ConcurrentHashMap.get
  at io.github.hectorvent.floci.core.storage.InMemoryStorage.get(InMemoryStorage.java:26)
  at io.github.hectorvent.floci.core.storage.AccountAwareStorageBackend.get
  at io.github.hectorvent.floci.services.appconfig.AppConfigDataService.getLatestConfiguration(AppConfigDataService.java:52)
  at io.github.hectorvent.floci.services.appconfig.AppConfigDataController.getLatestConfiguration(AppConfigDataController.java:41)
```

**Root cause**

`GET /configuration?list-type=2` (S3 ListObjectsV2, path-style) matches the AppConfig controller's `/configuration` mapping, so the request never reaches the S3 service.

**Impact**

S3 is a core service and `configuration` is a very plausible real bucket name; clients that use it get a confusing AppConfig NPE/500 on routine list/read calls.

**Suggested direction**

Let S3 path-style requests (`?list-type=2`, `?prefix=`, etc.) claim `/{bucket}` before the bare `/configuration` AppConfig mapping matches (or namespace AppConfig under a distinct prefix); and have `AppConfigDataController` / `AppConfigDataService` null-guard instead of throwing an NPE.

**Related:** #996 (the inverse collision — an AppConfig `listHostedConfigurationVersions` LIST falling through to S3's `@Path("/{bucket}/{key:.+}")` catch-all; fixed). This report is the reverse direction: an S3 path-style request being claimed by the AppConfig `/configuration` route.

LorenzoGalassi:
## Summary

Closes #1294

A path-style S3 request to a bucket named exactly `configuration` (e.g. `GET /configuration?list-type=2`, ListObjectsV2) matched AppConfigData's literal `GET /configuration` route ahead of S3's `/{bucket}` template. With no `configuration_token`, the AppConfig handler called `sessionStore.get(null)` and returned **HTTP 500** (`NullPointerException`) for routine list/read calls against a perfectly valid bucket.

Fix via path rewrite (consistent with `SqsQueueUrlRouterFilter` / `LambdaUrlRoutingFilter`):

- AppConfigData `GetLatestConfiguration` now binds an internal path instead of the public `/configuration`.
- A new `@PreMatching` `AppConfigConfigurationRouteFilter` rewrites genuine requests (those carrying a `configuration_token`) onto that internal path; requests without a token keep the bare `/configuration` path, which now falls through to S3's `/{bucket}` route.
- The internal path contains an underscore (forbidden in S3 bucket names), so it can never collide with a real bucket.

This is the reverse-direction sibling of #996 (an AppConfig LIST falling into S3's `/{bucket}/{key:.+}` catch-all).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `aws s3api list-objects-v2 --bucket configuration` (against a created, empty bucket) returned HTTP 500 instead of an empty `ListBucketResult`. Only the literal name `configuration` collided; other names (`cato-configuration`, `appconfig`, etc.) worked. Observed stack trace:

```
java.lang.NullPointerException
  at java.util.concurrent.ConcurrentHashMap.get
  at ...AppConfigDataService.getLatestConfiguration(AppConfigDataService.java:52)
  at ...AppConfigDataController.getLatestConfiguration(AppConfigDataController.java:41)
```

Disambiguation matches AWS: AppConfigData `GetLatestConfiguration` is `GET /configuration` with a required `configuration_token`; S3 path-style ListObjectsV2 has no such token, so it is routed to S3.

## Checklist

- [x] `./mvnw test` passes locally for the affected suites: `S3ConfigurationBucketIntegrationTest` + `AppConfigIntegrationTest` (26 passed, incl. genuine AppConfig `GetLatestConfiguration` with a token, an invalid token, and an empty configuration) and `S3IntegrationTest` (113 passed, no routing regression). The full suite's only failures on this machine are the 16 Docker-dependent suites that fail identically on an unmodified `main` (no Docker available locally).
- [x] New or updated integration test added: `S3ConfigurationBucketIntegrationTest` (RestAssured) creates bucket `configuration`, asserts ListObjectsV2 returns `ListBucketResult` (not an AppConfig error), then puts/gets/lists an object.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/appconfig/AppConfigDataController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/appconfig/AppConfigDataController.java
@@ -1,6 +1,7 @@
 package io.github.hectorvent.floci.services.appconfig;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.services.appconfig.AppConfigDataService.ConfigurationData;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.*;
@@ -35,9 +36,17 @@ public class AppConfigDataController {
         return Response.status(201).entity(Map.of("InitialConfigurationToken", token)).build();
     }
 
+    // Bound to an internal path rather than the public "/configuration" so that
+    // an S3 bucket named "configuration" is not shadowed by this route (issue
+    // #1294). AppConfigConfigurationRouteFilter rewrites genuine GetLatestConfiguration
+    // requests (those carrying a configuration_token) onto this path; requests
+    // without a token fall through to S3's /{bucket} handler.
     @GET
-    @Path("/configuration")
+    @Path(AppConfigConfigurationRouteFilter.INTERNAL_PATH)
     public Response getLatestConfiguration(@QueryParam("configuration_token") String token) {
+        if (token == null || token.isBlank()) {
+            throw new AwsException("BadRequestException", "configuration_token is required", 400);
+        }
         ConfigurationData data = service.getLatestConfiguration(token);
         return Response.ok(data.content())
                 .header("Content-Type", data.contentType())
--- a/src/main/java/io/github/hectorvent/floci/services/appconfig/AppConfigDataController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/appconfig/AppConfigDataController.java
@@ -1,7 +1,6 @@
 package io.github.hectorvent.floci.services.appconfig;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.services.appconfig.AppConfigDataService.ConfigurationData;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.*;
@@ -44,9 +43,6 @@ public class AppConfigDataController {
     @GET
     @Path(AppConfigConfigurationRouteFilter.INTERNAL_PATH)
     public Response getLatestConfiguration(@QueryParam("configuration_token") String token) {
-        if (token == null || token.isBlank()) {
-            throw new AwsException("BadRequestException", "configuration_token is required", 400);
-        }
         ConfigurationData data = service.getLatestConfiguration(token);
         return Response.ok(data.content())
                 .header("Content-Type", data.contentType())
--- a/src/main/java/io/github/hectorvent/floci/services/appconfig/AppConfigConfigurationRouteFilter.java
+++ b/src/main/java/io/github/hectorvent/floci/services/appconfig/AppConfigConfigurationRouteFilter.java
@@ -1,0 +1,57 @@
+package io.github.hectorvent.floci.services.appconfig;
+
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerRequestFilter;
+import jakarta.ws.rs.container.PreMatching;
+import jakarta.ws.rs.core.UriInfo;
+import jakarta.ws.rs.ext.Provider;
+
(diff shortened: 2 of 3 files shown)


hectorvent:
@LorenzoGalassi Thanks for the PR. 

I am looking if we can fix this is another way. We are injecting a controller inside another one. We fixed issues like this in the past using PATH rewrite, just need to find where.

LorenzoGalassi:
Thanks @hectorvent, you're right. I've pushed an update using exactly that approach: a `@PreMatching` `AppConfigConfigurationRouteFilter` that rewrites genuine `GetLatestConfiguration` requests (the ones carrying a `configuration_token`) onto an internal path, leaving the bare `/configuration` to fall through to S3's `/{bucket}` route. It follows the same pattern as `SqsQueueUrlRouterFilter` / `LambdaUrlRoutingFilter`.

This was actually the second option I had in mind as a resolution, and in hindsight I underestimated the code smell of injecting one controller into another, your call was the right one. All AppConfig and S3 suites pass.

src/main/java/io/github/hectorvent/floci/services/appconfig/AppConfigDataController.java
```diff
@@ -35,9 +36,17 @@ public Response startConfigurationSession(String body) throws IOException {
...
+    // #1294). AppConfigConfigurationRouteFilter rewrites genuine GetLatestConfiguration
+    // requests (those carrying a configuration_token) onto this path; requests
+    // without a token fall through to S3's /{bucket} handler.
     @GET
-    @Path("/configuration")
+    @Path(AppConfigConfigurationRouteFilter.INTERNAL_PATH)
     public Response getLatestConfiguration(@QueryParam("configuration_token") String token) {
+        if (token == null || token.isBlank()) {
+            throw new AwsException("BadRequestException", "configuration_token is required", 400);
+        }
```
hectorvent: Will this be reachable? I think not required.

LorenzoGalassi: Good catch, agreed. That branch can't actually be reached: the method only binds the internal path, and the filter only rewrites onto it when a configuration_token is present, so the guard was dead code. Dropped it (and the now-unused import) in 272a61e8.

hectorvent (approved):

hectorvent:
Thanks @LorenzoGalassi

Issue #964, opened 2026-05-20
[BUG] S3 website hosting does not serve custom error document


SrEloyF:
## Service

S3

## AWS API Action

PutBucketWebsite / GetObject (website hosting behavior)

## Expected behavior

When a bucket is configured with a custom error document:

```bash
aws s3 website s3://my-bucket \
  --index-document index.html \
  --error-document error.html
```

requests to missing objects through the website endpoint should return the configured `error.html` document, matching AWS S3 static website hosting behavior.

For example:

```text
http://my-bucket.s3.localhost.floci.io:4566/missing-page
```

should render the contents of `error.html`.

## Actual behavior

The website endpoint correctly serves `index.html` at the root path:

```text
http://my-bucket.s3.localhost.floci.io:4566
```

However, requests to non-existent objects do not return the configured custom error document.

Instead, Floci responds with the default S3 XML error response:

```xml
<Error>
  <NoSuchKey/>
  <Message>The specified key does not exist.</Message>
  <RequestId>fd86feff-edba-46dd-8707-e5ae9c886bef</RequestId>
</Error>
```

Example failing request:

```text
http://my-bucket.s3.localhost.floci.io:4566/missing-page
```

Expected behavior is to return the contents of `error.html` instead.

## Reproduction

```bash
# Create bucket
aws --endpoint-url http://localhost:4566 s3 mb s3://my-bucket

# Configure static website hosting
aws --endpoint-url http://localhost:4566 s3 website s3://my-bucket \
  --index-document index.html \
  --error-document error.html

# Upload website files
echo '<h1>Home</h1>' > index.html
echo '<h1>Custom Error</h1>' > error.html

aws --endpoint-url http://localhost:4566 s3 cp index.html s3://my-bucket/index.html
aws --endpoint-url http://localhost:4566 s3 cp error.html s3://my-bucket/error.html

# Works correctly
curl http://my-bucket.s3.localhost.floci.io:4566

# Expected to return error.html, but does not
curl http://my-bucket.s3.localhost.floci.io:4566/missing-page
```

## Environment

* Floci version / image tag: `floci/floci:latest`
* Java version: OpenJDK 17.0.18
* How you're running Floci: Docker Compose

```yaml
services:
  floci:
    image: floci/floci:latest
    container_name: floci
    ports:
      - "4566:4566"
    environment:
      FLOCI_STORAGE_MODE: hybrid
      FLOCI_STORAGE_PERSISTENT_PATH: /app/data
[...]

PR #1334, opened 2026-06-13
fix(s3): serve custom error document for missing keys in static websi…

JESUSLUG:
## Summary #964 

 serve custom error document for missing keys in static website hosting

When a bucket has a website configuration with an ErrorDocument, requests
to missing keys now return the error document with a 404 status instead
of the default XML NoSuchKey error.

Also fixes a flaky test in `StepFunctionsDynamoDbIntegrationTest` where
`waitForExecution` returned null output on slow CI runners, causing the
test to fail.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -305,8 +305,8 @@ public class S3Controller {
                                 @QueryParam("marker") String marker,
                                 @Context UriInfo uriInfo,
                                 @Context HttpHeaders httpHeaders) {
-        validateRawUri();
         try {
+            validateRawUri();
             if (hasQueryParam(uriInfo, "uploads")) {
                 return handleListMultipartUploads(bucket);
             }
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -372,7 +372,17 @@ public class S3Controller {
                                     .header("x-amz-website-redirect-location", "index")
                                     .build();
                         } catch (AwsException e) {
-                            // If index.html is missing, we could serve ErrorDocument, but for now we fall back to listObjects
+                            if (webConfig.getErrorDocument() != null) {
+                                try {
+                                    S3Object errorObj = s3Service.getObject(bucket, webConfig.getErrorDocument());
+                                    return Response.status(404)
+                                            .entity(errorObj.getData())
+                                            .type(errorObj.getContentType())
+                                            .header("Content-Length", errorObj.getSize())
+                                            .build();
+                                } catch (AwsException ignored) {
+                                }
+                            }
                         }
                     }
                 } catch (AwsException e) {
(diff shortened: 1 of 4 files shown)


JESUSLUG:
Hi @hectorvent ! I see the test (Build and test) failed due to StepFunctionsDynamoDbIntegrationTest - I'm curious, I don't think the error is from my changes since I only touched services/s3. And even when running the tests locally, they pass 20/20.

JESUSLUG:
Hello again! I pushed a commit that fixes the flaky test in CI (Build and Test). It wasn't related to this issue, but I took the opportunity to fix it.

hectorvent:
Nice fix, thanks for this. A few small things that would be good to tidy up, none of them blocking:

**1. Split out the StepFunctions change.** The `StepFunctionsDynamoDbIntegrationTest` tweak is unrelated to S3 website hosting, so it would be cleaner as its own PR, or at least a mention in the description.

**2. When the error document itself is missing, render HTML rather than XML.** Right now it falls back to the XML `<Error>` body, but a real website endpoint returns an HTML 404 page:

```html
<html><head><title>404 Not Found</title></head>
<body><h1>404 Not Found</h1>
<ul><li>Code: NoSuchKey</li><li>Message: The specified key does not exist.</li></ul></body></html>
```

**3. Add the website error headers.** AWS website error responses also include `x-amz-error-code` and `x-amz-error-message`:

```java
Response.status(404)
    .entity(errorObj.getData())
    .type(errorObj.getContentType())
    .header("Content-Length", errorObj.getSize())
    .header("x-amz-error-code", "NoSuchKey")
    .header("x-amz-error-message", "The specified key does not exist.")
    .build();
```

**4. De-duplicate the error document block.** The same lookup plus 404 response now lives in both `listObjects` and `getObject`, so a small shared helper would avoid the drift:

```java
private Response serveErrorDocument(String bucket, WebsiteConfiguration cfg) {
    if (cfg.getErrorDocument() == null) {
        return null;
    }
    try {
        S3Object err = s3Service.getObject(bucket, cfg.getErrorDocument());
        return Response.status(404).entity(err.getData())
            .type(err.getContentType())
            .header("Content-Length", err.getSize()).build();
    } catch (AwsException ignored) {
        return null;
    }
}
```

Thanks again for the contribution.

JESUSLUG:
Hi again @hectorvent ! Sorry for the delay on this. 
Regarding your comments, I’ve made the corrections and updated the PR description to include the fix for StepFunctionsDynamoDbIntegrationTest.

hectorvent:
Thanks for the updates @JESUSLUG, the helper, HTML fallback, and error headers all look good now.

One thing I missed in my first pass, sorry for raising it late: this serves the error document on the normal REST `GetObject` path, gated only on the bucket having a website config. In AWS the error document is a website endpoint behavior (`bucket.s3-website-<region>.amazonaws.com`); the REST API endpoint that SDKs use always returns the XML `<Error><Code>NoSuchKey</Code></Error>` for a missing key. So as written, an `s3.getObject()` for a missing key on a website enabled bucket now gets HTML instead of the modeled XML error, and the SDK can no longer map it to `NoSuchKeyException`.

Could we condition the error document (and ideally the existing index document logic too) on the website host, and fall through to the XML error for normal API requests? That keeps SDK behavior intact while still fixing #964 for the website endpoint. Happy to talk through the detection if useful.

JESUSLUG:
Hi @hectorvent ! Good catch. I've updated the logic to isolate this behavior based on the request host.

### 1. Website endpoint:
```bash
curl -v -H "Host: test-website.s3-website-us-east-1.localhost:4566" \
  http://localhost:4566/missing-page

# Output:
# HTTP/1.1 404 Not Found
# Content-Type: text/html
# <html><body>Custom 404</body></html>

```

### 2. REST API:

```bash
curl -v http://localhost:4566/test-website/missing-page

# Output:
# HTTP/1.1 404 Not Found
# Content-Type: application/xml
# <?xml version="1.0" encoding="UTF-8"?><Error><Code>NoSuchKey</Code>...

```
This restores the expected SDK behavior for REST API requests while preserving custom index/error document handling on the website endpoint.

Please let me know if you'd like me to make any further adjustments.

hectorvent (approved):

Issue #1333, opened 2026-06-13
[BUG] Unable to parse response (not well-formed (invalid token)


ashraf-minhaj:
## Service

AWS EC2

## AWS API Action

Tried creating VPC using terraform or trying ec2 api from CLI - 

## Expected behavior

Create VPC, subnets etc. Or list ec2 resources.

## Actual behavior

With terraform when we try ec2 API, keeps on trying `aws_vpc.main: Still creating... [07m10s elapsed]` , also the output from cli -
Command -

```bash
 AWS-IaCs-with-Terraform % aws ec2 describe-regions \
  --endpoint-url http://localhost:4567
```

Output -
```
# this works
> aws sts get-caller-identity \
  --endpoint-url http://localhost:4567
{
    "UserId": "000000000000",
    "Account": "000000000000",
    "Arn": "arn:aws:iam::000000000000:root"
}

# this does not
>  aws ec2 describe-regions \
  --endpoint-url http://localhost:4567

Unable to parse response (not well-formed (invalid token): line 1, column 0), invalid XML received. Further retries may succeed:
```

## Reproduction

```terraform
provider "aws" {
  region     = var.aws_region
  access_key = "test"
  secret_key = "test"

  skip_credentials_validation = true
  skip_metadata_api_check     = true
  skip_requesting_account_id  = true
  s3_use_path_style           = true

  endpoints {
    iam     = var.endpoint
    # vpc     = var.endpoint
    sts     = var.endpoint
    ec2     = var.endpoint
    eks     = var.endpoint
    route53 = var.endpoint
  }
}

terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 5.17.0"
    }
  }
}

variable "aws_region" {
  type    = string
  default = "us-east-1"
}

variable "endpoint" {
  type    = string
  default = "http://localhost:4567"
}

variable "component_prefix" {
  type    = string
  default = "minhaj-infra"
}

variable "component_postfix" {
  type    = string
  default = "dev"
}

resource "aws_vpc" "main" {
  cidr_block           = "10.0.0.0/16"
  enable_dns_hostnames = false #true
  enable_dns_support   = false #true

  tags = {
    Name = "floci-vpc"
  }
}
```

## Environment

- Floci version / image tag: FLOCI_VERSION=1.5.24, docker container, on mac m1
- Java SDK version (if applicable): 
- How you're running Floci (Docker / native / `mvn quarkus:dev`): docker

[...]

PR #1347, opened 2026-06-15
fix(eventbridge): sync EventBridge tags to ResourceGroupsTaggingService

### LINKED ISSUE #1343: [BUG] Resource Groups Tagging API ignores EventBus rules
## Service

Resource Groups Tagging API

## AWS API Action

GetResources

## Expected behavior

Rule is found by tag filter.

```
[
  {
    "ResourceARN": "arn:aws:events:eu-west-1:000000000000:rule/eventBus-bug-tags/get-tags-bug-225803",
    "Tags": [
      {
        "Key": "tag-key",
        "Value": "com.example.test.v1.0.0"
      }
    ]
  }
]
```

## Actual behavior

```
[]
```

## Reproduction

```bash
#!/usr/bin/env bash
# Usage:
#   ./bin/floci-get-tags-tag-bug.sh <image>   # starts container
#   ./bin/floci-get-tags-tag-bug.sh            # uses ambient AWS credentials
set -euo pipefail

IMAGE="${1:-}"
REGION="${2:-eu-west-1}"
CONTAINER_NAME="eb-get-tags-bug-$$"
BUS_NAME="eventBus-bug-tags"
RULE_NAME="get-tags-bug-$$"
TAG_KEY="tag-key"
TAG_VAL="com.example.test.v1.0.0"

export AWS_DEFAULT_REGION="${REGION}"
export AWS_PAGER=""

if [[ -n "${IMAGE}" ]]; then
    ENDPOINT="http://localhost:4566"
    export AWS_ACCESS_KEY_ID=test
    export AWS_ACCOUNT_ID="000000000000"
    AWS="aws --output json --endpoint-url=${ENDPOINT} --region ${REGION}"
else
    ENDPOINT=""
    AWS="aws --output json --region ${REGION}"
    AWS_ACCOUNT_ID="$(${AWS} sts get-caller-identity | jq -r '.Account')"
    export AWS_ACCOUNT_ID
fi

wait_for_ready() {
    local I=0
    while ! curl -sf "${ENDPOINT}/_localstack/health" > /dev/null 2>&1; do
        I=$(( I + 1 ))
        [[ "${I}" -ge 30 ]] && { echo "ERROR: container not ready" >&2; exit 1; }
        sleep 1
    done
}

cleanup() {
    local BUS_ARN="arn:aws:events:${REGION}:${AWS_ACCOUNT_ID}:event-bus/${BUS_NAME}"
    if [[ -n "${IMAGE}" ]]; then
        docker stop "${CONTAINER_NAME}" > /dev/null 2>&1 || true
    else
        ${AWS} events delete-rule --name "${RULE_NAME}" --event-bus-name "${BUS_ARN}" 2>/dev/null || true
        ${AWS} events delete-event-bus --name "${BUS_NAME}" 2>/dev/null || true
    fi
}

trap cleanup EXIT

if [[ -n "${IMAGE}" ]]; then
    docker run -d --rm --name "${CONTAINER_NAME}" -p 4566:4566 "${IMAGE}"
    wait_for_ready
fi

BUS_ARN="$(${AWS} events create-event-bus --name "${BUS_NAME}" | jq -r '.EventBusArn')"
echo "Bus ARN: ${BUS_ARN}"

RULE_ARN="$(
    ${AWS} events put-rule \
        --name "${RULE_NAME}" \
        --event-bus-name "${BUS_ARN}" \
        --event-pattern "{\"source\":[\"test\"]}" \
        --state DISABLED \
        --tags "Key=${TAG_KEY},Value=${TAG_VAL}" \
    | jq -r '.RuleArn'
)"
echo "Rule ARN: ${RULE_ARN}"

echo ""
echo "=== Check: resourcegroupstaggingapi:GetResources ==="
RESULT="$(
    ${AWS} resourcegroupstaggingapi get-resources \
        --resource-type-filters "events:rule" \
        --tag-filters "Key=${TAG_KEY},Values=${TAG_VAL}"
)"
echo "${RESULT}" | jq '.ResourceTagMappingList'
COUNT="$(echo "${RESULT}" | jq '.ResourceTagMappingList|length')"
[[ "${COUNT}" -gt 0 ]] && echo "PASS" || echo "FAIL" >&2

```

## Environment

- Floci version / image tag: 1.5.25, nightly-06152026
- How you're running Floci (Docker / native / `mvn quarkus:dev`): docker

9046balaji:
## Summary

Fixes #1343

esourcegroupstaggingapi:GetResources with --resource-type-filters events:rule returned an empty list even when EventBridge rules had tags.

## Root Cause

EventBridgeService stored tags on rules, buses, and archives in its own stores (
uleStore, usStore, rchiveStore) but never propagated them to the central ResourceGroupsTaggingService. The tagging API's getResources() method queries only its own internal store map, which was never populated by EventBridge operations.

## Fix

Inject ResourceGroupsTaggingService into EventBridgeService (following the existing pattern established by GlueService) and sync tags to the central tagging store during:

- putRule — when tags are provided at rule creation time
- 	agResource — for rules, buses, and archives
- untagResource — for rules, buses, and archives
- deleteRule — cleanup the tagging entry

## Files Changed

- EventBridgeService.java — added ResourceGroupsTaggingService injection and tag sync calls
- EventBridgeServiceTest.java — updated constructor to pass new parameter
- EventBridgeSchedulerIntegrationTest.java — updated constructor to pass new parameter

## Testing

All existing tests pass (98 tests, 0 failures):
- EventBridgeTagResourceIntegrationTest — 8/8 ✅
- ResourceGroupsTaggingIntegrationTest — 14/14 ✅
- EventBridgeSchedulerIntegrationTest — 14/14 ✅
- EventBridgeServiceTest — 62/62 ✅

No new tests were added because the fix is covered by the existing EventBridgeTagResourceIntegrationTest and ResourceGroupsTaggingIntegrationTest suites. The end-to-end flow (put-rule → tag → GetResources) can be verified with the reproduction script from the issue.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeService.java
@@ -150,6 +150,7 @@ public class EventBridgeService {
         );
         if (tags != null) {
             bus.getTags().putAll(tags);
+            resourceGroupsTaggingService.tagResources(List.of(bus.getArn()), tags, region);
         }
         busStore.put(key, bus);
         LOG.infov("Created event bus: {0} in region {1}", name, region);
@@ -166,7 +167,7 @@ public class EventBridgeService {
             throw new AwsException("ValidationException", "Cannot delete the default event bus.", 400);
         }
         String key = busKey(region, effectiveName);
-        busStore.get(key)
+        EventBus bus = busStore.get(key)
                 .orElseThrow(() -> new AwsException("ResourceNotFoundException",
                         "EventBus not found: " + effectiveName, 404));
         String rulePrefix = ruleKeyPrefix(region, effectiveName);
@@ -176,6 +177,7 @@ public class EventBridgeService {
                     "Cannot delete event bus with existing rules: " + name, 400);
         }
         busStore.delete(key);
+        resourceGroupsTaggingService.deleteResources(List.of(bus.getArn()), region);
         LOG.infov("Deleted event bus: {0}", name);
     }
 
@@ -1092,11 +1094,12 @@ public class EventBridgeService {
 
     public void deleteArchive(String archiveName, String region) {
         String key = archiveKey(region, archiveName);
-        archiveStore.get(key)
+        Archive archive = archiveStore.get(key)
                 .orElseThrow(() -> new AwsException("ResourceNotFoundException",
                         "Archive not found: " + archiveName, 404));
         archiveStore.delete(key);
         archivedEventStore.delete(archivedEventKey(region, archiveName));
+        resourceGroupsTaggingService.deleteResources(List.of(archive.getArchiveArn()), region);
         LOG.infov("Deleted archive: {0}", archiveName);
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeService.java
@@ -17,6 +17,7 @@ import io.github.hectorvent.floci.services.eventbridge.model.ReplayState;
 import io.github.hectorvent.floci.services.eventbridge.model.Rule;
 import io.github.hectorvent.floci.services.eventbridge.model.RuleState;
 import io.github.hectorvent.floci.services.eventbridge.model.Target;
+import io.github.hectorvent.floci.services.resourcegroupstagging.ResourceGroupsTaggingService;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -48,6 +49,7 @@ public class EventBridgeService {
     private final ObjectMapper objectMapper;
     private final RuleScheduler ruleScheduler;
     private final EventBridgeInvoker invoker;
+    private final ResourceGroupsTaggingService resourceGroupsTaggingService;
     private final ReplayDispatcher replayDispatcher;
 
     @Inject
@@ -57,7 +59,8 @@ public class EventBridgeService {
                               ObjectMapper objectMapper,
                               RuleScheduler ruleScheduler,
                               EventBridgeInvoker invoker,
-                              ReplayDispatcher replayDispatcher) {
+                              ReplayDispatcher replayDispatcher,
+                              ResourceGroupsTaggingService resourceGroupsTaggingService) {
         this(
                 storageFactory.create("eventbridge", "eventbridge-buses.json",
                         new TypeReference<Map<String, EventBus>>() {}),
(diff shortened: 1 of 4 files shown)


hectorvent:
Nice fix on the tagging sync, this is the right call. Verified the rule ARNs resolve to `events`/`rule`, so `GetResources --resource-type-filters events:rule` picks them up now. `PutRule` does carry `Tags` in the model, so syncing at create time is correct too.

One thing before this merges though: this PR also changes `AwsQueryController` (the `MissingAction` error plus the EC2 volume fallbacks), and that's the exact same diff as #1348:

```
+  "CreateVolume", "DescribeVolumes", "DeleteVolume"
-  return null;
+  return xmlErrorResponse("MissingAction", "The request must contain the parameter Action", 400);
```

Both also add the same `AwsQueryControllerIntegrationTest`. Since #1348 is the dedicated PR for that (and closes #1333), could you drop those bits here and keep them there? Otherwise whichever lands second hits a conflict, and it keeps this PR focused on #1343.

Small follow up, not blocking: `deleteRule` cleans up the central store, but bus and archive deletes don't, so a tagged-then-deleted bus would linger in `GetResources`. Out of scope for this issue, just flagging since you wired up bus/archive tag sync here.

Rest looks good. 👍

9046balaji:
Hi @hectorvent,

Thanks for the feedback!

1). I've rebased the branch onto the latest main and dropped the duplicate AwsQueryController and AwsQueryControllerIntegrationTest changes. This PR now focuses solely on the EventBridge tagging sync.

2). Thanks for flagging the lingering tagged buses/archives on delete. I'll leave that clean-up out of scope for this PR since we wanted to keep this focused, but I'll make sure it's addressed in a subsequent task.
Let me know if everything looks good now!

hectorvent:
Nice work, the rule path checks out against [service-2.json](https://github.com/aws/aws-sdk-java-v2/blob/master/services/eventbridge/src/main/resources/codegen-resources/service-2.json) and this resolves #1343.

One gap in the same spirit: `CreateEventBus` also takes `Tags`, and `createEventBus` stores them but never pushes them to the central store, so a bus created with `--tags` stays invisible to `GetResources` until someone calls `tag-resource`. The same one liner you used elsewhere covers it:

```suggestion
        if (tags != null) {
            bus.getTags().putAll(tags);
            resourceGroupsTaggingService.tagResources(List.of(bus.getArn()), tags, region);
        }
```

Small follow up too: only `deleteRule` clears the central entry, so `deleteEventBus` and `deleteArchive` leave stale mappings and a deleted bus still shows up in `GetResources`. Worth mirroring the same `deleteResources` cleanup there.

9046balaji:
@hectorvent  Thanks for the feedback! I've updated the PR with both suggestions:

1. **Tagging on EventBus Creation**: Added tag registration using the suggested one-liner inside `createEventBus`.
2. **Deletion Cleanup**: Added deletion cleanup using `resourceGroupsTaggingService.deleteResources` in both `deleteEventBus` and `deleteArchive`.
3. **Integration Tests**: Added `createEventBusWithTagsAndVerifyCentralTagging` and `deleteArchiveClearsCentralTagging` tests to verify that `GetResources` is updated correctly during creation and deletion.

All tests ran and passed successfully.

9046balaji:
@hectorvent 

can you review this one and can you merge it's been a longs time since i opened this request

hectorvent (approved):
Thanks @9046balaji

PR #1352, opened 2026-06-15
fix(rds-data): add native MySQL JDBC support

### LINKED ISSUE #1351: RDS Data API should include native MySQL JDBC support
## Service
RDS Data API

## AWS API Action
`rds-data:ExecuteStatement`

## Expected behavior
`execute-statement` against a local MySQL/Aurora-compatible RDS Data API resource should run the SQL and return results from the published Floci image.

## Actual behavior
On `floci/floci:1.5.25`, the same call can fail in the native image because the MySQL JDBC driver is not available to the native binary:

```text
DatabaseErrorException: No suitable driver found for jdbc:mysql://<host>:<port>/mydb?useSSL=false&allowPublicKeyRetrieval=true
```

The JVM compatibility image can pass this path because a plain JDBC driver jar is available on the JVM classpath. The published native image needs the same capability at native-image build time.

## Reproduction
```bash
# Run against the published native image, then create a local Aurora MySQL/MySQL-compatible
# RDS resource and matching secret through Floci before executing SQL.
aws --endpoint-url http://localhost:4566 rds-data execute-statement \
  --resource-arn arn:aws:rds:us-east-1:000000000000:cluster:mydb \
  --secret-arn arn:aws:secretsmanager:us-east-1:000000000000:secret:mydb \
  --database mydb \
  --sql "SELECT 1"
```

## Environment
- Floci version / image tag: `floci/floci:1.5.25`
- Java SDK version (if applicable): n/a; reproduced with AWS CLI-compatible request
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker published native image

devansh-jain-18:
## Summary
Fixes #1351.

This updates the native-image packaging path used by RDS Data API MySQL execution. The bug report has the repro and native-vs-JVM context; follow-up CI coverage is tracked in #1353.

## Type of change
- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility
Verified `rds-data execute-statement` against a local MySQL/Aurora-compatible RDS resource on a `docker/Dockerfile.native-package` build using create/insert/select SQL.

Additional checks:

- `./mvnw -q -Dtest=RdsDataServiceTest,RdsDataControllerIntegrationTest test`
- `./mvnw -q -DskipTests package`
- `./mvnw -q clean package -Dnative -DskipTests`

## Checklist
- [ ] New or updated integration test added (native-image CI coverage tracked in #1353)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -230,9 +230,8 @@
             <artifactId>jackson-annotations</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.mysql</groupId>
-            <artifactId>mysql-connector-j</artifactId>
-            <version>8.3.0</version>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-mysql</artifactId>
         </dependency>
         <dependency>
             <groupId>com.h2database</groupId>
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -11,6 +11,8 @@ quarkus:
     host: 0.0.0.0
     limits:
       max-body-size: ${floci.max-request-size}M
+  datasource:
+    active: false
   security:
      security-providers: BC
   shutdown:
--- a/src/test/resources/application.yml
+++ b/src/test/resources/application.yml
@@ -4,6 +4,8 @@ quarkus:
     test-port: 0
     limits:
       max-body-size: ${floci.max-request-size}M
+  datasource:
+    active: false
 
 floci:
   base-url: "http://localhost:4566"


hectorvent:
Picking this one to merge over #1355 (same `quarkus-jdbc-mysql` fix, but this version is more complete).

The part that makes it the right call: Floci uses raw `DriverManager.getConnection(...)` rather than an Agroal datasource, so adding the extension means Quarkus would otherwise try to manage a default datasource we never configure. The `quarkus.datasource.active: false` here keeps that inert, and replacing the `8.3.0` pin with the extension keeps the driver version and native metadata in one place.

Two follow-ups before/at merge:

- Worth pulling in the root-cause explanation from #1355 (build-time `DriverManager` registration) into the commit, credit to @hughesjs.
- No automated regression test yet, native CI coverage is tracked in #1353, fine to defer but let's note it.

Thanks!

PR #1355, opened 2026-06-15
fix(rds-data): register MySQL JDBC driver in native image

### LINKED ISSUE #1354: RDS Data API ExecuteStatement fails in native image: "No suitable driver found for jdbc:mysql"
`ExecuteStatement` always fails on `floci/floci:1.5.25` against a provisioned, `available` Aurora MySQL cluster — the JDBC driver isn't registered in the native image.

## Repro

```bash
docker run --rm -p 4566:4566 floci/floci:1.5.25
```

Then (any SDK/CLI): create an `aurora-mysql` cluster via `rds CreateDBCluster`, wait until `DescribeDBClusters` shows `available`, create a Secrets Manager secret with `{"username":...,"password":...}`, and call `rds-data ExecuteStatement` with that cluster's ARN, the secret ARN, and `SELECT 1`. Response:

```json
{"__type":"DatabaseErrorException","message":"No suitable driver found for jdbc:mysql://172.17.0.x:3306/..."}
```

The MySQL container is up and healthy when this happens — it's an in-process `DriverManager.getConnection()` failure inside Floci, not a connectivity problem. The JVM build is unaffected, which is why the compatibility tests don't catch it.

## Cause

`RdsDataConnectionFactory` opens connections with `DriverManager.getConnection("jdbc:mysql://…")`, and `com.mysql:mysql-connector-j` is on the classpath — but in the GraalVM native image the MySQL driver never registers itself with `DriverManager`. `DriverManager` is initialised at *build* time, so its registered-driver list is computed during the native build (when no driver has self-registered) and frozen into the image; `Connector/J`'s self-registration static initialiser never runs.

Worth noting for anyone who finds this via the EC2 image-catalog issue (#1298): this is *not* simply a missing resource registration. Adding `META-INF/services/java.sql.Driver` to `resource-config.json` (and reflection on `com.mysql.cj.jdbc.Driver`) is **not** sufficient on its own — I tried it and `ExecuteStatement` still fails identically, because those only affect runtime visibility, not the build-time `DriverManager` initialisation.

## Fix

Add the Quarkus MySQL JDBC extension, which carries the native-image metadata + build-time driver registration for Connector/J:

```xml
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-jdbc-mysql</artifactId>
</dependency>
```

## Validation

Built a native image off `main` with that one-line change via `docker/Dockerfile.native` and re-ran the repro: `ExecuteStatement` now returns the expected result set. Also confirmed end-to-end against a downstream Testcontainers RDS Data round-trip (provision cluster + secret → `ExecuteStatement SELECT 1`), which goes from failing to green on the patched image. PR to follow.

Seen on `linux/arm64`. `BeginTransaction` / `BatchExecuteStatement` use the same connection path.

hughesjs:
## Problem

RDS Data API `ExecuteStatement` always fails on the published native image (`floci/floci:1.5.25`) against a provisioned, `available` Aurora MySQL cluster:

```json
{"__type":"DatabaseErrorException","message":"No suitable driver found for jdbc:mysql://..."}
```

`RdsDataConnectionFactory` opens `jdbc:mysql://…` via `DriverManager`, and `com.mysql:mysql-connector-j` is on the classpath — but in the native image Connector/J never self-registers. `DriverManager` is initialised at **build** time, so its registered-driver list is frozen before any driver runs its self-registration static initialiser. The JVM build is unaffected, which is why the compatibility tests don't catch it.

## Fix

Add the `quarkus-jdbc-mysql` extension, which provides the build-time driver registration and native-image metadata for Connector/J.

Note: registering `META-INF/services/java.sql.Driver` in `resource-config.json` (+ reflection on `com.mysql.cj.jdbc.Driver`) is **not** sufficient on its own — that only affects runtime resource/reflection visibility, not the build-time `DriverManager` initialisation. I tried that first and `ExecuteStatement` failed identically.

## Validation

Built a native image off `main` with this change via `docker/Dockerfile.native` and confirmed `ExecuteStatement` returns the expected result set. Also verified end-to-end against a downstream Testcontainers RDS Data round-trip (provision cluster + Secrets Manager secret → `ExecuteStatement SELECT 1`), which goes from failing to green on the patched image.

Closes #1354

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -235,6 +235,10 @@
             <version>8.3.0</version>
         </dependency>
         <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-mysql</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
             <version>2.3.232</version>


hughesjs:
Accidentally closed this with my PR from the other repo, whoops

hectorvent:
Thanks for this, your root-cause writeup is the clearest explanation of the bug we have. The build-time `DriverManager` freeze (and why the JVM build hides it, and why `resource-config.json` alone isn't enough) is exactly right.

Going to close this in favor of #1352, which lands the same `quarkus-jdbc-mysql` fix but also handles two things this one leaves open:

- Floci opens connections with raw `DriverManager.getConnection(...)`, not an Agroal datasource, so pulling in the extension means Quarkus now sees a JDBC datasource it shouldn't manage. #1352 adds `quarkus.datasource.active: false` to keep that inert.
- It replaces the explicit `mysql-connector-j:8.3.0` pin with the extension instead of keeping both, so the driver version and the extension's native metadata stay in sync.

Not a knock on this PR at all, the analysis here is what made the call easy. We'll carry your explanation over so it isn't lost. 🙏

Issue #1038, opened 2026-05-25
[FEAT] Support use of AWS IoT core


jtaySWE:
## Service

AWS IoT Core

## API Action / Feature

Support using of the service.

## AWS Documentation

[](https://docs.aws.amazon.com/pdfs/iot/latest/apireference/iot-api.pdf)(https://docs.aws.amazon.com/iot/latest/developerguide/what-is-aws-iot.html)

## Are you willing to contribute a PR?

- [ ] Yes
- [X] No

cachebag:
Would be great for us at work.

nblomquist:
@hectorvent I am looking into this. Hardest part will be the MQTT broker integration. No guarantees on getting this working.
Do we have ways to mark things in the code/documentation/settings as experimental?

hectorvent:
@nblomquist I will take a look and we can chat about it. Are you on the floci slack? Ping me tomorrow and we can brainstorm about it. 

Thanks!

nblomquist:
> [@nblomquist](https://github.com/nblomquist) I will take a look and we can chat about it. Are you on the floci slack? Ping me tomorrow and we can brainstorm about it.
> 
> Thanks!

@hectorvent I am not on the floci slack.

hectorvent:
https://join.slack.com/t/floci/shared_invite/zt-3tjn02s3q-A00kEjJ1cZxsg_imTfy6Cw

hectorvent:
Implemented by https://github.com/floci-io/floci/pull/1359

PR #1359, opened 2026-06-15
feat(iot) Feature/issue 1038 iot core

nblomquist:
## Summary

EXPERIMENTAL: IoT Api.

Would like to see if this enough for people to get started. It uses a patched version of Moquette MQTT broker.
The patched fork is vendored into this PR.
The code can be found here:
https://github.com/nblomquist/moquette-floci/tree/floci/0.18.0-reserved-topics

related to #1038 

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility
AWS SDK for Java v2 2.42.24
Python: boto3==1.37.1, botocore==1.37.1

## Checklist

- [x] `./mvnw test` passes locally (some unrelated fail)


hectorvent:
@nblomquist Please resolve conflicts.

nblomquist:
Finally got the tests to pass. It was having trouble with raw sockets. Shouldn't have used those in the first place. I added a MQTT client library for testing and that seems to have cleared it.

hectorvent:
Really happy to see IoT Core getting off the ground, and honestly the control plane plus the test coverage here is a lot of solid work. I do have one architectural worry on the data plane that I think is worth sorting out before this lands.

The MQTT broker is a vendored, patched fork of Moquette shipped as a binary jar (`vendor/maven/.../moquette-broker-0.18.0-floci-1.jar`), pulled in through a new `file://` repo plus `jitpack.io`, force tracked in git, and copied into both Docker images, on top of dragging in `h2-mvstore`, `jaxb-api`, and two `servlet-api`s. That bumps into Floci's first principles in a few places. The jar can't be rebuilt from source in CI or CVE patched by maintainers, the fork only exists to flip one ACL flag (`ALLOW_RESERVED_PUBLISH_PREFIXES = "$aws/"`), and there's no GraalVM config for it, so the `Dockerfile.native` path is a quiet break waiting to happen. CLAUDE.md actually calls this exact case out: data plane protocols should use raw TCP, with ElastiCache and RDS as the precedent.

The encouraging part is that the integration is already cleanly seamed. `IotMqttBrokerService` is the only class that touches Moquette, and everything else talks to it through about seven methods plus the publish and subscribe callbacks. The service even keeps its own `subscriptionsByClient` registry already.

So the idea would be to keep that seam and just swap the engine for `io.vertx:vertx-mqtt`. It's first party Vert.x on the exact Netty version Quarkus already ships (4.5.27), it's on Maven Central, and it's native friendly. Its handlers line up almost one to one with the interceptors you already wrote:

```java
MqttServer.create(vertx).endpointHandler(endpoint -> {
    endpoint.publishHandler(msg -> route(endpoint, msg.topicName(), msg.payload()))
            .subscribeHandler(sub -> registry.subscribe(endpoint, sub))
            .unsubscribeHandler(unsub -> registry.unsubscribe(endpoint, unsub))
            .accept(false);
}).listen(port);
```

You'd implement the subscriber routing, the `+` and `#` topic matcher, and a retained message map (call it 250 to 400 lines), and the `$aws/` reserved topic issue just goes away because you own delivery and never apply a `$` prefix ACL. No fork, no jitpack, no vendored binary, none of the h2/jaxb/servlet weight. The integration and Paho tests you already have become the regression net for the swap.

Glad to help sketch the target interface if that's useful. Could also land the control plane first with the data plane behind the existing `FLOCI_SERVICES_IOT_MQTT_ENABLED` flag to keep the change easy to review.

hectorvent (changes requested):
Happy to chat and draft the re-implementation

nblomquist:
I wasn't happy with Moquette but it seemed the most expedient. I saw Vert.x wasn't a full broker and decided against it. Hive definitely wasn't going to be native friendly.

I have begun looking at Vert.x in a separate branch just to see what it will take.
Thanks for the input!

nblomquist:
@hectorvent I updated this with Vert.x MQTT.

hectorvent (approved):
Hi @nblomquist, 

Solid implementation. Thanks for the implementation. 
We will need to need to do implement some lazy load for services to starts on demand, but we are good for this round. 

Thank you again.

PR #1375, opened 2026-06-16
ci: run focused native compatibility tests

### LINKED ISSUE #1353: Compatibility tests should exercise the native release image
## Service
Compatibility tests / release image coverage

## API Action / Feature
Run at least one compatibility check against the same native Docker image shape that gets published in release and nightly builds.

## AWS Documentation
N/A — this is about CI coverage for Floci's published image, not a single AWS API action.

## Why is this needed?
Right now the compatibility workflow gives us confidence in the JVM-packaged image, but the images users actually pull are native images:

- `compatibility.yml` builds `docker/Dockerfile.jvm-package` as `floci:test`.
- `release.yml` publishes `docker/Dockerfile.native-package` as `floci/floci:<version>` and `latest`.
- `nightly.yml` also publishes `docker/Dockerfile.native-package` for nightly tags.

Those two image shapes are close, but they are not identical. A feature can work in the JVM image and still fail in the native image because native builds need some things to be included up front at build time.

RDS Data API is a recent example: the SDK compatibility path passed on the JVM image, but the published native image still needed native JDBC driver support before `rds-data execute-statement` worked there.

It would be useful to add a small native-image smoke check so CI proves that the published image shape can start and handle at least a representative compatibility-sensitive workflow. This does not need to duplicate the full compatibility matrix at first; the goal is simply to catch native-only packaging/runtime gaps before release.

- [ ] Yes
- [ ] No

Looking for maintainer guidance on the preferred scope before committing to an implementation.

electrum:
Compatibility CI now runs the Java SDK and AWS CLI suites against the native image as well as the JVM image, making native-only runtime issues visible during pull request testing.

Fixes: #1353

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - 'src/**'
       - 'pom.xml'
+      - 'docker/Dockerfile.native'
       - 'docker/Dockerfile.jvm-package'
       - 'compatibility-tests/**'
       - '.github/workflows/compatibility.yml'
@@ -18,7 +19,7 @@ concurrency:
 
 jobs:
   build:
-    name: Build floci image
+    name: Build JVM floci image
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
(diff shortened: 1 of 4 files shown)


electrum:
This includes the fixes from #1374, #1377, #1379 to show that this CI change works. Two of those will conflict with each other as they both modify the `application.yml`. You could merge them separately or together here.

hectorvent:
@electrum thank you.

PR #1383, opened 2026-06-16
fix(secretsmanager): 🐛 honor ListSecrets MaxResults and NextToken pagination

### LINKED ISSUE #1358: [BUG] Secrets Manager ListSecrets ignores MaxResults pagination parameter
Creating this a bug but arguably it could also be a feature request to support max results for listing secrets.

## Service

Secrets Manager

## AWS API Action

ListSecrets

## Expected behavior

When calling `ListSecrets` with `MaxResults=N`, the response should return at most 'N' secrets (or whatever the standard AWS pagination pattern defines).

## Actual behavior

Floci ignores the `MaxResults` parameter entirely and returns all secrets.

## Reproduction

A complete reproducer is available here: [floci-secretsmanager-maxresults-reproducer.zip](https://github.com/user-attachments/files/28958275/floci-secretsmanager-maxresults-reproducer.zip)

To run:
```bash
cd floci-secretsmanager-maxresults-reproducer
mvn clean test
```

The test:
1. Creates 5 secrets
2. Calls ListSecrets with MaxResults=2
3. Expects 2 secrets
4. Actually receives all 5 secrets

## Environment

- Floci version / image tag: floci/floci:1.5.25
- Java SDK version: AWS SDK for Java v2 2.46.10
- How you're running Floci: Docker / Podman via testcontainers

LorenzoGalassi:
## Summary

Closes #1358

`ListSecrets` ignored the `MaxResults` parameter and always returned every secret, so SDK callers that paginate received the full set in one response.

Changes:

- Page the result set by `MaxResults` and emit a `NextToken` (an opaque offset into the stably-sorted secret list) when more results remain; subsequent calls resume from it.
- Validate `MaxResults` (1-100, `ValidationException`) and `NextToken` (`InvalidNextTokenException`).
- When `MaxResults` is omitted, behavior is unchanged (all secrets returned, no `NextToken`).

Secrets are sorted by name before paging so pages are stable across calls (the underlying store scan order is not guaranteed).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `ListSecrets` with `MaxResults=2` against 5 stored secrets returned all 5 instead of 2 plus a `NextToken`. Real AWS Secrets Manager returns at most `MaxResults` items and a `NextToken` to fetch the next page.

Verified against the AWS SDK for Java v2 pagination contract described in the issue (create 5 secrets, `MaxResults=2` returns 2 with a `NextToken`, walking the token yields each secret exactly once).

## Checklist

- [x] `./mvnw test` passes locally for the affected suites: `SecretsManagerJsonHandlerTest` + `SecretsManagerServiceTest` (64 passed). The full suite's only failures on this machine are the 16 Docker-dependent suites that fail identically on an unmodified `main` (no Docker available locally).
- [x] New or updated integration test added: `SecretsManagerJsonHandlerTest` gains coverage for `MaxResults` paging + `NextToken` walk (every secret seen once), all-results-when-omitted, out-of-range `MaxResults` (400), and invalid `NextToken` (400).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java
@@ -244,8 +244,10 @@ public class SecretsManagerJsonHandler {
 
     private Response handleListSecrets(JsonNode request, String region) {
         List<Secret> secrets = new ArrayList<>(service.listSecrets(region));
-        // Sort for stable pagination across calls (the store scan order is not guaranteed).
-        secrets.sort(Comparator.comparing(Secret::getName));
+        // AWS lists secrets by CreatedDate when SortBy is absent; sort on it (name as a
+        // tiebreaker) for AWS-matching, stable pagination across calls.
+        secrets.sort(Comparator.comparing(Secret::getCreatedDate, Comparator.nullsLast(Comparator.naturalOrder()))
+                .thenComparing(Secret::getName));
 
         // MaxResults is constrained to 1-100; when absent the full result set is returned.
         int maxResults = secrets.size();
@@ -253,10 +255,8 @@ public class SecretsManagerJsonHandler {
             maxResults = request.path("MaxResults").asInt();
             if (maxResults < 1 || maxResults > 100) {
                 return Response.status(400)
-                        .entity(new AwsErrorResponse("ValidationException",
-                                "1 validation error detected: Value '" + maxResults + "' at 'maxResults' failed to "
-                                        + "satisfy constraint: Member must have value less than or equal to 100 and "
-                                        + "greater than or equal to 1"))
+                        .entity(new AwsErrorResponse("InvalidParameterException",
+                                "Invalid MaxResults value, must be between 1 and 100"))
                         .build();
             }
         }
--- a/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java
@@ -12,6 +12,7 @@ import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


hectorvent:
Nice fix on the pagination, the `NextToken` / `MaxResults` / `SecretList` shape matches the paginator spec and the 1..100 bound is right.

One blocking thing: `ValidationException` isn't a valid error for `ListSecrets`. It doesn't exist anywhere in the Secrets Manager model (verified in both the Java v2 and botocore `service-2.json`, and there's no sdk-extras override). The declared errors are `InvalidParameterException`, `InvalidRequestException`, `InvalidNextTokenException`, `InternalServiceError`. Because the SDK doesn't range check `MaxResults` client side, the request actually reaches the server, so the code you return is what the caller sees, and an unmodeled `ValidationException` won't map to a typed exception. Real AWS returns `InvalidParameterException` here, which this handler already uses elsewhere.

```java
return Response.status(400)
        .entity(new AwsErrorResponse("InvalidParameterException",
                "Invalid MaxResults value, must be between 1 and 100"))
        .build();
```

(the Coral style "1 validation error detected..." string belongs to services that actually emit `ValidationException`, so it can go too.)

Two minor notes, non blocking:

Default ordering. AWS lists secrets by `CreatedDate` when `SortBy` is absent (`SortBy` default is `created-date`). The `Secret` model already exposes `getCreatedDate()`, so sorting on that keeps pages stable and matches AWS order:

```java
secrets.sort(Comparator.comparing(Secret::getCreatedDate)
        .thenComparing(Secret::getName));
```

With `MaxResults` omitted you return every secret and no token. Real AWS still paginates at a server default. Fine to keep for the emulator, just flagging the divergence.

Verified against `secretsmanager` [`service-2.json`](https://github.com/aws/aws-sdk-java-v2/blob/master/services/secretsmanager/src/main/resources/codegen-resources/service-2.json) (Java v2 + botocore) and the ListSecrets API reference.

LorenzoGalassi:
Thanks for the thorough review, and for checking it against the model. Pushed 5f9f4b6:

- Switched the out-of-range `MaxResults` error to `InvalidParameterException` (and dropped the Coral-style message) since `ValidationException` isn't in the `ListSecrets` model and wouldn't map to a typed exception. Good catch, agreed.
- Took the ordering suggestion too: now sorting by `CreatedDate` (with `Name` as a tiebreaker) to match the default `SortBy=created-date` and keep pages stable.
- On the omitted-`MaxResults` case, I left it returning all results without a token for the emulator, as you suggested.

Also strengthened the test to assert the returned error code is `InvalidParameterException`.

hectorvent (approved):

Issue #1378, opened 2026-06-16
[BUG] API Gateway VTL templates fail in native image


electrum:
## Service

API Gateway

## AWS API Action

[CreateRestApi](https://docs.aws.amazon.com/apigateway/latest/api/API_CreateRestApi.html)

[PutMethod](https://docs.aws.amazon.com/apigateway/latest/api/API_PutMethod.html)

[PutIntegration](https://docs.aws.amazon.com/apigateway/latest/api/API_PutIntegration.html)

[CreateDeployment](https://docs.aws.amazon.com/apigateway/latest/api/API_CreateDeployment.html)

[Invoking REST APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-call-api.html)

## Expected behavior

The native Docker image should evaluate API Gateway VTL request and response mapping templates the same way as AWS behavior.

## Actual behavior

The native Docker image returns HTTP 500 when invoking an API Gateway REST API route that uses VTL mapping templates because Velocity cannot load bundled resources and classes that it resolves by name.

Error excerpts:

```text
Cannot get Velocity Runtime default properties!
ClassNotFoundException: org.apache.velocity.runtime.resource.ResourceManagerImpl
Error loading directive.properties!
ClassNotFoundException: org.apache.velocity.runtime.parser.StandardParser
```

## Reproduction

```bash
docker run --rm -p 4566:4566 floci/floci:1.5.25

export AWS_ACCESS_KEY_ID=test
export AWS_SECRET_ACCESS_KEY=test
export AWS_DEFAULT_REGION=us-east-1
export AWS_EC2_METADATA_DISABLED=true
export FLOCI_ENDPOINT=http://localhost:4566

cd compatibility-tests/sdk-test-java
../../mvnw -q -Dtest='ApigwSfnJsonataCrudlTests' test
```

The SDK test creates a REST API, configures AWS integrations with VTL request and response templates, deploys a stage, and invokes the REST API through the execute-api path.

## Environment

- Floci version / image tag: `floci/floci:1.5.25`
- Java SDK version (if applicable): AWS SDK for Java v2 from `compatibility-tests/sdk-test-java`
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker native image

devansh-jain-18:
Thanks for the linked PR.

[...]

PR #1385, opened 2026-06-16
fix(apigateway): support SQS query-protocol integrations (path-style URI)

### LINKED ISSUE #1384: [BUG] API Gateway → SQS integration fails for path-style URI + query-protocol (form-urlencoded) request templates
## Service

API Gateway (REST/v1) → SQS AWS integration

## AWS API Action

`PUT /restapis/{id}/resources/{id}/methods/{m}/integration` with an SQS service integration, invoked via `execute-api`.

## Expected behavior

A standard API Gateway → SQS proxy (`AWS` integration type) should work. The canonical AWS recipe — also supported by LocalStack — is:

- Integration URI in **path style**: `arn:aws:apigateway:{region}:sqs:path/{account}/{queueName}`
- `integration.request.header.Content-Type` set to `'application/x-www-form-urlencoded'`
- A VTL request template that renders the SQS **query protocol** body, e.g.
  `Action=SendMessage&QueueUrl=$util.urlEncode("...")&MessageBody=$util.urlEncode($input.body)`

Floci should parse the path-style URI, render the VTL template (this part already works), and dispatch the resulting form-urlencoded `Action=SendMessage&...` body to SQS, returning the SQS query-protocol XML response.

## Actual behavior

The request fails. Two distinct gaps:

1. **URI parsing** — `AwsServiceRouter.parseIntegrationUri` only accepts the `action/{ActionName}` form and returns `null` for the `path/{account}/{queue}` form, so the call fails with *"Cannot parse AWS integration URI"* (HTTP 500).

2. **Body dispatch** — even past (1), `ApiGatewayExecuteController.invokeAwsIntegration` does `objectMapper.readTree(transformedBody)` and routes the result to `SqsJsonHandler`. The rendered template body is **form-urlencoded query protocol** (`Action=SendMessage&...`), not JSON, so parsing fails.

Note: Floci already has a `SqsQueryHandler` that consumes exactly this form-encoded `MultivaluedMap` (it backs the normal SQS query-protocol endpoint). The execute controller just never routes to it.

## Reproduction

Configure a REST API with an `AWS` integration:
- URI: `arn:aws:apigateway:eu-west-1:sqs:path/000000000000/my-queue`
- `--integration-http-method POST`
- `--request-parameters '{"integration.request.header.Content-Type":"'\''application/x-www-form-urlencoded'\''"}'`
- request template (`application/json`): `Action=SendMessage&QueueUrl=$util.urlEncode("http://localhost:4566/000000000000/my-queue")&MessageBody=$util.urlEncode($input.body)`

`POST` to the deployed stage → 500 instead of the message landing on the queue.

This is the standard pattern real services use; it surfaced migrating a service from LocalStack to Floci.

## Environment

- Floci version / image tag: `floci/floci:latest` (reproduces on `main` @ current HEAD)
- Related: native-image VTL fix is tracked separately in #1378 / #1379 — this is a distinct functional gap, not the native-image issue.

stashev:
## Context

API Gateway → SQS is a common pattern: expose an SQS queue directly through a REST API so producers can `POST` JSON and have it land on a queue, with no Lambda in between. The canonical AWS recipe (also supported by LocalStack) configures the integration as:

- a **path-style** integration URI: `arn:aws:apigateway:{region}:sqs:path/{account}/{queue}`
- `integration.request.header.Content-Type` forced to `application/x-www-form-urlencoded`
- a VTL request template that renders the SQS **query protocol**: `Action=SendMessage&QueueUrl=$util.urlEncode("...")&MessageBody=$util.urlEncode($input.body)`

Floci couldn't service this, which blocked migrating a service from LocalStack. See #1384.

## What changed

Floci already had a working `VtlTemplateEngine`, an `AWS` (non-proxy) integration path, and an `SqsQueryHandler` for the form-encoded query protocol — they just weren't wired together for this case. Three fixes connect them:

1. **Path-style URI parsing** — `AwsServiceRouter.parseIntegrationUri` only accepted the `action/{Action}` form and returned `null` for `path/{resourcePath}`, so the request failed immediately with *"Cannot parse AWS integration URI"* (500). It now parses the `path/` form too, leaving the action `null` (the action is carried in the rendered body).

2. **Query-protocol dispatch** — `invokeAwsIntegration` always parsed the rendered body as JSON and routed to the JSON handlers. For a path-style target it now parses the body as form-urlencoded and dispatches through a new `AwsServiceRouter.invokeQuery`, which routes to the existing `SqsQueryHandler`. The SQS query-protocol XML response is passed straight back.

3. **Request-template selection by incoming Content-Type** — AWS selects the request template using the *incoming request* Content-Type, but Floci applied the `integration.request.header.Content-Type` override (`application/x-www-form-urlencoded`) into the header map *before* template selection, so a JSON client request was misread and rejected with a spurious 415. The incoming Content-Type is now captured before parameter mapping mutates the map. This also makes single-key templates (`application/json` only) behave like AWS, instead of requiring the template to be registered under both content types as a workaround.

Scope is intentionally SQS-only — the dispatch in `invokeQuery` has a single `sqs` case today and is structured to extend to other query-protocol services (e.g. SNS) later if needed.

## Test coverage

New `ApiGatewaySqsIntegrationTest` exercises the full path-style + form-urlencoded `SendMessage` flow end to end (configure API → deploy → POST through `execute-api` → assert the SQS XML response → `ReceiveMessage` to confirm the body landed on the queue). Full suite (606 tests) passes.

Gaps: only SQS `SendMessage` is covered for the query-protocol path; other SQS actions reachable this way (e.g. `SendMessageBatch`) reuse the same `SqsQueryHandler` but aren't separately exercised through API Gateway.

## Known follow-up items

- Extending `invokeQuery` to other query-protocol services (SNS, etc.) is out of scope here.
- Relationship to #1378 / #1379: that work fixes the Velocity engine in the **native image**; this PR is a distinct functional gap and is orthogonal to it.

Closes #1384

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -813,6 +813,13 @@ public class ApiGatewayExecuteController {
                 bodyStr, headerMap, queryMap, pathMap, stageName, httpMethod,
                 resource.getPath(), requestId, regionResolver.getAccountId(), null);
 
+        // AWS selects the request template by the *incoming* request Content-Type. Capture it
+        // before parameter mapping runs, since an integration.request.header.Content-Type
+        // override (common for SQS query-protocol integrations) would otherwise clobber it and
+        // misdirect template selection.
+        String incomingContentType = headerMap.getOrDefault("Content-Type",
+                headerMap.getOrDefault("content-type", "application/json"));
+
         // Apply request parameter mapping (method.request.* → integration.request.*)
         Map<String, String> integrationReqParams = integration.getRequestParameters();
         if (integrationReqParams != null && !integrationReqParams.isEmpty()) {
@@ -835,8 +842,6 @@ public class ApiGatewayExecuteController {
         // Content-Type negotiation and passthrough behavior
         String transformedBody;
         Map<String, String> requestTemplates = integration.getRequestTemplates();
-        String incomingContentType = headerMap.getOrDefault("Content-Type",
-                headerMap.getOrDefault("content-type", "application/json"));
 
         if (requestTemplates != null && !requestTemplates.isEmpty()) {
             // Try exact match first, then wildcard fallback
@@ -879,13 +884,23 @@ public class ApiGatewayExecuteController {
             transformedBody = bodyStr != null ? bodyStr : "";
         }
 
-        // Dispatch to service
+        // Dispatch to service.
+        //
+        // A path-style integration URI (arn:...:{service}:path/...) carries no action: the
+        // rendered template body is the AWS query protocol (form-urlencoded,
+        // "Action=SendMessage&..."). Action-style URIs (arn:...:{service}:action/{Action})
+        // carry the action in the URI and render a JSON body.
         Response serviceResponse;
         String errorType = null;
         String errorMessage = null;
         try {
-            JsonNode requestJson = objectMapper.readTree(transformedBody);
-            serviceResponse = serviceRouter.invoke(target.service(), target.action(), requestJson, region);
+            if (target.action() == null) {
+                MultivaluedMap<String, String> formParams = parseFormUrlEncoded(transformedBody);
+                serviceResponse = serviceRouter.invokeQuery(target.service(), formParams, region);
+            } else {
+                JsonNode requestJson = objectMapper.readTree(transformedBody);
+                serviceResponse = serviceRouter.invoke(target.service(), target.action(), requestJson, region);
+            }
         } catch (AwsException e) {
             errorType = e.getErrorCode();
             errorMessage = e.getMessage();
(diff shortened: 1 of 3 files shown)


hectorvent:
Solid fix. The wiring is clean and it reuses the existing query-protocol path correctly. `MissingAction` (400) matches what `AwsQueryController` already returns and is the real AWS query code, the `path/`/`action` URI split is right, and capturing the incoming Content-Type before parameter mapping matches how API Gateway actually selects the request template.

One small thing to fix now: the `MissingAction` message in `invokeQuery` drifts from the rest of Floci and from AWS. `AwsQueryController` and real AWS use "The request must contain the parameter Action", so let's match that exactly:

```java
throw new AwsException("MissingAction",
        "The request must contain the parameter Action", 400);
```

Two follow ups worth tracking separately, not blocking here:

Path style queue routing. The `path/{account}/{queue}` components are parsed and then dropped, so the queue is only resolvable when the template re-supplies `QueueUrl` in the body (which this test does). In real AWS the path is the routing: the request lands on the SQS endpoint `POST /{account}/{queue}` and the queue comes from the path, so the common recipe body is just `Action=SendMessage&MessageBody=$input.body` with no `QueueUrl`. Deriving the queue from the parsed path when `QueueUrl` is absent would make that recipe work too. Good as a follow up issue.

Extending `invokeQuery` to other query-protocol services (SNS, etc.), already noted as out of scope.

Nice work, this unblocks the canonical APIGW to SQS pattern.

stashev:
Thanks for the thorough review. Fixed the message in b1dc3e6b: `invokeQuery` now throws `"The request must contain the parameter Action"`, matching `AwsQueryController` and AWS.

Agreed on the path-style queue routing follow-up. Deriving the queue from the parsed `path/{account}/{queue}` when `QueueUrl` is absent would make the canonical `Action=SendMessage&MessageBody=$input.body` recipe work without the template re-supplying `QueueUrl`. Filed it as #1408. The SNS / other query-protocol services one I've left out of scope; `invokeQuery` is structured to extend to them later.

hectorvent (approved):

PR #1389, opened 2026-06-16
fix(config): make runtime defaults explicit

jvanzyl:
## Summary
- make the default request body size explicit in config and docs
- gate the seeded IAM deployer principal behind an opt-in config flag
- keep IAM managed policy seeding unchanged while avoiding default admin credentials

## Tests
- ./mvnw -Dtest='ApplicationDefaultsTest,IamServiceTest,IamIntegrationTest' test

### DIFF
--- a/docs/configuration/advanced/application-yml.md
+++ b/docs/configuration/advanced/application-yml.md
@@ -36,7 +36,7 @@ The block below mirrors `src/main/resources/application.yml`, it's the effective
 
 ```yaml
 floci:
-  max-request-size: 512              # Max HTTP request body size in MB
+  max-request-size: 2048             # Max HTTP request body size in MB
   base-url: "http://localhost:4566"  # Used to build response URLs (SQS QueueUrl, SNS endpoints, etc.)
   # hostname: ""                     # When set, overrides the host in base-url for multi-container Docker
   default-region: us-east-1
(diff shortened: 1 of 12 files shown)


hectorvent:
Looks good. Gating the seeded `floci-deployer` admin behind an opt-in flag is the right call, and it is actually closer to real AWS which has no preseeded principal. The `GetCallerIdentity` fallback stays valid, returning the account root identity when the `floci` key is not seeded:

```xml
<Arn>arn:aws:iam::000000000000:root</Arn>
```

Config wiring is complete and nothing else in the repo depends on the seeded key, so the change is well contained.

A few notes, none blocking:

The `max-request-size` 512 to 2048 bump is really a separate concern from the IAM gate, and it is a real 4x increase rather than just making the default explicit. Fine to keep, just worth calling both out in the changelog.

Dropping the default `floci` / `floci` key changes behavior for anyone who adopted it. The docs cover the opt-in, so just worth a changelog note since this is a `fix(config)`.

`listUsers` moved from `hasItem("test-user")` to `equalTo("test-user")`, which now asserts exactly one user and leans on the single member collapsing to a scalar. A bit brittle if another user ever lands in that run.

hectorvent (approved):

PR #1392, opened 2026-06-16
fix(rds): report DB parameter group attachments

jvanzyl:
## Summary
- include DBParameterGroups in DescribeDBInstances responses
- report the stored DB parameter group name with an in-sync apply status
- preserve existing DB instance response shape around subnet groups, tags, and metadata

## Tests
- ./mvnw -Dtest='RdsQueryHandlerTest' test

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
@@ -621,6 +621,7 @@ public class RdsQueryHandler {
                .elem("Status", "active")
              .end("VpcSecurityGroupMembership")
            .end("VpcSecurityGroups")
+           .raw(dbParameterGroupsXml(i))
            .raw(dbSubnetGroupXml(dbSubnetGroupForInstance(i)))
            .elem("DbiResourceId", i.getDbiResourceId())
            .elem("DBInstanceArn", i.getDbInstanceArn());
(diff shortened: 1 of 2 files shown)


hectorvent:
Shape looks right, the `DBParameterGroup` member element, `DBParameterGroupName` and `ParameterApplyStatus` all match the model, and `in-sync` is a valid status.

One thing worth handling: when the caller does not pass `DBParameterGroupName` at create time, `getParameterGroupName()` stays null and we emit an empty `<DBParameterGroups></DBParameterGroups>`. Real AWS never returns that empty. If you do not specify a group, RDS assigns and reports the default, something like `default.postgres16`, with `in-sync`.

So this surfaces the explicit attachment case but still under-reports the common default case. Falling back to `default.{engine}{majorVersion}` when nothing is stored would make `DescribeDBInstances` always report at least one group like AWS does, and would be worth a test alongside the explicit one.

jvanzyl:
Updated this to match the default attachment behavior you described: when no DBParameterGroupName is stored, DescribeDBInstances now reports an AWS-style default group name derived from engine + major version, for example default.postgres16, with ParameterApplyStatus=in-sync. Added the default-case test alongside the explicit attachment test. Focused local check: ./mvnw -Dtest='RdsQueryHandlerTest' test.

hectorvent (approved):
Thanks @jvanzyl

PR #1394, opened 2026-06-16
fix(cognito): add missing UserPoolClient validation

shrimptails-f:
## Summary

Add missing `UserPoolClient` validation and tests to better align Cognito behavior with AWS.

The previous `UserPoolClient` implementation accepted invalid OAuth and token validity settings that should have been rejected.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

This bug fix adds validation for `UserPoolClient` settings that were previously accepted incorrectly.

Aligned behavior includes:
- rejecting invalid `TokenValidityUnits` keys and unit values
- rejecting non-positive token validity values
- rejecting OAuth settings when `AllowedOAuthFlowsUserPoolClient` is `false`
- rejecting `DefaultRedirectURI` values that aren't included in `CallbackURLs`

Error messages were adjusted to follow AWS documentation wording for `CreateUserPoolClient` and `UpdateUserPoolClient`.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -304,30 +304,52 @@ public class CognitoService {
             Boolean enableTokenRevocation) {
         describeUserPool(userPoolId);
         String clientId = UUID.randomUUID().toString().replace("-", "").substring(0, 26);
+        List<String> normalizedAllowedOAuthFlows = normalizeStringList(allowedOAuthFlows);
+        List<String> normalizedAllowedOAuthScopes = normalizeStringList(allowedOAuthScopes);
+        List<String> normalizedCallbackUrls = normalizeStringList(callbackURLs);
+        String normalizedDefaultRedirectUri = normalizeOptionalString(defaultRedirectURI);
+        List<String> normalizedExplicitAuthFlows = normalizeStringList(explicitAuthFlows);
+        List<String> normalizedLogoutUrls = normalizeStringList(logoutURLs);
+        List<String> normalizedReadAttributes = normalizeStringList(readAttributes);
+        List<String> normalizedSupportedIdentityProviders = normalizeStringList(supportedIdentityProviders);
+        Map<String, String> copiedTokenValidityUnits = copyStringMap(tokenValidityUnits);
+        List<String> normalizedWriteAttributes = normalizeStringList(writeAttributes);
+
+        validateUserPoolClientConfiguration(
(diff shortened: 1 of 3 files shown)


hectorvent:
Thanks for tightening this up. The error code choice is right, `InvalidParameterException` is declared for both `CreateUserPoolClient` and `UpdateUserPoolClient`, and the DefaultRedirectURI and OAuth flow rules match the model.

One behavior worth fixing before merge: `RefreshTokenValidity: 0` should be accepted, not rejected. AWS coerces it to the default instead of erroring. From the CreateUserPoolClient API reference:

> You can't set RefreshTokenValidity to 0. If you do, Amazon Cognito overrides the value with the default value of 30 days.

The valid range for `RefreshTokenValidity` is min 0, while `AccessTokenValidity` and `IdTokenValidity` are min 1. So the `<= 0` check is correct for access and id tokens but too strict for refresh:

```java
validateTokenValidityValue("AccessTokenValidity", accessTokenValidity); // min 1, reject <= 0  ok
validateTokenValidityValue("IdTokenValidity", idTokenValidity);         // min 1, reject <= 0  ok
// RefreshTokenValidity min is 0; 0 means "use default", so it should pass (ideally coerce to 30 days)
```

This request returns 200 on AWS today, with `RefreshTokenValidity` set to the 30 day default:

```json
{ "UserPoolId": "...", "ClientName": "c", "RefreshTokenValidity": 0 }
```

Two smaller things, optional:

* `LogoutURLs` should also require `AllowedOAuthFlowsUserPoolClient` true. The docs list it next to CallbackURLs, AllowedOAuthScopes and AllowedOAuthFlows, so right now it slips through when the flag is false.
* The unit check lowercases the value, so `"Hours"` passes here but AWS only accepts the lowercase enum `seconds|minutes|hours|days`.

Verified against `service-2.json` (Cognito Identity Provider) and the CreateUserPoolClient API reference.

shrimptails-f:
@hectorvent 
Thankyou!

Addressed.

- `RefreshTokenValidity=0` is now accepted and coerced to the 30-day default for both create and update paths.
- `LogoutURLs` now also requires `AllowedOAuthFlowsUserPoolClient=true`.
- `TokenValidityUnits` now only accepts the lowercase enum values `seconds|minutes|hours|days`, so mixed-case values like `"Hours"` are rejected.

I also added unit coverage in `CognitoServiceTest` for these cases.

hectorvent (approved):
Thanks @shrimptails-f

PR #1406, opened 2026-06-17
fix(sqs): message processing from non-default account

alexjplant:
## Summary

The Lambda SQS Event Source Mapping poller runs on a Vert.x periodic timer thread with no CDI request scope. When polling a queue owned by a non-default account (e.g. `000000000001`), `doReceiveMessage()` called `queueStore.get(storageKey)`, which went through `AccountAwareStorageBackend.prefix()`. That method caught `ContextNotActiveException` and fell back to `defaultAccountId` (`000000000000`), causing a key mismatch — the queue was stored under one account prefix but looked up under another. Result: queue not found, empty receive, Lambda never fires.

Additionally, `GuardedMessageQueue.persist()` called `messageStore.put(storageKey, ...)` which had the same account-prefix problem, corrupting message state persistence (visibility changes, deletions) when they originated from the poller's async thread.

Three targeted fixes:

1. **`SqsService.doReceiveMessage()`** — now accepts `queueUrl` and uses the existing `getQueueByUrl(storageKey, queueUrl)` helper (extracts account from URL path, bypasses request-context dependency) instead of `queueStore.get(storageKey)`. Mirrors how `sendMessage` and `deleteMessage` already resolve accounts.

2. **`GuardedMessageQueue.persist()`** — when the message store is `AccountAwareStorageBackend`, extracts the account ID from the storage key (`region::/accountId/queueName`) via a new `extractAccountFromStorageKey()` helper and calls `putForAccount(accountId, storageKey, messages)` instead of `put()`. Falls back to the existing `put()` path when the store is not account-aware or the account cannot be extracted.

3. **`EsmIntegrationTest`** — new `esmPollerReceivesMessagesFromNonDefaultAccount` test that creates a queue, Lambda function, and ESM under account `000000000001` (using SigV4 `Credential=000000000001/...` per the LocalStack 12-digit AKID convention), sends a message, and asserts `ApproximateNumberOfMessagesNotVisible > 0` within a timeout — proving the poller successfully receives from a non-default account's queue.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

SQS `ReceiveMessage` called from the Lambda ESM poller now correctly resolves the queue's account from the queue URL rather than from the CDI request context (which is absent on background threads). This matches AWS behaviour where a queue URL like `https://sqs.us-east-1.amazonaws.com/000000000001/my-queue` unambiguously identifies the account. No changes to request or response wire formats. The `GuardedMessageQueue` persistence fix ensures message state transitions (visibility timeout, deletion) are stored under the correct account, matching how they were originally created.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueue.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueue.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.services.sqs;
 
+import io.github.hectorvent.floci.core.storage.AccountAwareStorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageBackend;
 import io.github.hectorvent.floci.services.sqs.model.Message;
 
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Thanks @alexjplant

PR #1409, opened 2026-06-17
fix(ecs): reject negative desiredCount in CreateService and UpdateService

### LINKED ISSUE #1382: [BUG] ECS accepts negative Desired count when creating a service
## Service

ECS

## AWS API Action
`CreateService` and `UpdateService`
## Expected behavior
When negative desired count is provided, AWS rejects it with an InvalidParameterException or ClientException with 400

## Actual behavior

Floci accepts and creates a service with with negative desiredCount

## Reproduction

1. Start floci
2. register task definition
```bash
aws ecs register-task-definition \
    --family "test-negative-count" \
    --container-definitions "[{\"name\":\"app\",\"image\":\"nginx:latest\"}]" \
    --endpoint-url http://localhost:4566 \
    --region us-east-1
```

3. create a service with negative desired count (here -5)
```bash
aws ecs create-service \
    --cluster default \
    --service-name "buggy-service" \
    --task-definition "test-negative-count" \
    --desired-count -5 \
    --endpoint-url http://localhost:4566 \
    --region us-east-1
```

4. output
```bash
{
    "service": {
        "serviceArn": "arn:aws:ecs:us-east-1:000000000000:service/default/buggy-service",
        "serviceName": "buggy-service",
        "clusterArn": "arn:aws:ecs:us-east-1:000000000000:cluster/default",
        "status": "ACTIVE",
        "desiredCount": -5,
        "runningCount": 0,
        "pendingCount": 0,
        "launchType": "FARGATE",
        "taskDefinition": "test-negative-count",
        "createdAt": "2026-06-16T14:31:35.426000+05:30"
    }
}

```

## Environment

- Floci version / image tag: 1.5.24
- Java SDK version (if applicable): Java 25
- How you're running Floci (Docker / native / `mvn quarkus:dev`): mvn quarkus:dev

9046balaji:
## Problem

Closes #1382

`CreateService` and `UpdateService` accepted any integer for `desiredCount`, including negative values. Floci stored and returned negative counts without error. AWS rejects negative values with a `ClientException` / HTTP 400.

## Root Cause

`EcsService.createService()` and `EcsService.updateService()` both assigned `desiredCount` directly to the model with no bounds check. The JSON handler (`EcsJsonHandler`) only parsed the integer from the request and forwarded it — no validation in either layer.

## Fix

Added a single guard in `EcsService.java` in both methods before the value is set on the model:

```java
if (desiredCount < 0) {
    throw new AwsException("ClientException", "Invalid desiredCount", 400);
}
```

Zero is a valid value (it stops all tasks for the service) and continues to be accepted.

## Testing

4 new test cases added to `EcsIntegrationTest`:

- `createServiceRejectsNegativeDesiredCount` — `-5` returns 400 + `ClientException`
- `createServiceRejectsNegativeOneDesiredCount` — `-1` returns 400 + `ClientException`
- `updateServiceRejectsNegativeDesiredCount` — `-3` on update returns 400 + `ClientException`
- `createServiceAcceptsZeroDesiredCount` — `0` returns 200 (valid boundary value)

All 43 `EcsIntegrationTest` tests pass.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java
@@ -590,6 +590,9 @@ public class EcsService {
         svc.setClusterArn(cluster.getClusterArn());
         svc.setTaskDefinition(taskDefinition);
         svc.setLaunchType(launchType != null ? launchType : LaunchType.FARGATE);
+        if (desiredCount < 0) {
+            throw new AwsException("InvalidParameterException", "desiredCount cannot be a negative number.", 400);
+        }
         svc.setDesiredCount(desiredCount);
         svc.setLoadBalancers(loadBalancers);
         svc.setNetworkConfiguration(networkConfiguration);
@@ -623,6 +626,9 @@ public class EcsService {
                     "Service " + serviceName + " is not active.", 400);
         }
         if (desiredCount != null) {
+            if (desiredCount < 0) {
+                throw new AwsException("InvalidParameterException", "desiredCount cannot be a negative number.", 400);
+            }
             svc.setDesiredCount(desiredCount);
         }
         if (networkConfiguration != null) {
--- a/src/test/java/io/github/hectorvent/floci/services/ecs/EcsIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ecs/EcsIntegrationTest.java
@@ -646,6 +646,70 @@ class EcsIntegrationTest {
                 .body("message", containsString("Creation of service was not idempotent."));
     }
 
+    // ── Negative desiredCount validation (issue #1382) ───────────────────────
+
+    @Test
+    @Order(36)
+    void createServiceRejectsNegativeDesiredCount() {
+        ecs("CreateService")
+                .body("""
+                {
+                    "cluster": "%s",
+                    "serviceName": "negative-count-svc",
+                    "taskDefinition": "%s",
+                    "desiredCount": -5,
+                    "launchType": "FARGATE"
+                }
+                """.formatted(CLUSTER_NAME, TASK_DEF_FAMILY))
+                .when()
+                .post("/")
+                .then()
+                .statusCode(400)
+                .body("__type", containsString("InvalidParameterException"))
+                .body("message", containsString("desiredCount cannot be a negative number."));
+    }
+
+
+
+    @Test
+    @Order(38)
+    void updateServiceRejectsNegativeDesiredCount() {
+        ecs("UpdateService")
+                .body("""
+                {
+                    "cluster": "%s",
+                    "service": "%s",
+                    "desiredCount": -3
+                }
+                """.formatted(CLUSTER_NAME, SERVICE_NAME))
+                .when()
+                .post("/")
+                .then()
+                .statusCode(400)
+                .body("__type", containsString("InvalidParameterException"))
+                .body("message", containsString("desiredCount cannot be a negative number."));
+    }
+
+    @Test
+    @Order(39)
+    void createServiceAcceptsZeroDesiredCount() {
+        ecs("CreateService")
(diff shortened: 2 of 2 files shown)


src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java
```diff
@@ -590,6 +590,9 @@ public EcsServiceModel createService(String clusterRef, String serviceName, Stri
         svc.setClusterArn(cluster.getClusterArn());
         svc.setTaskDefinition(taskDefinition);
         svc.setLaunchType(launchType != null ? launchType : LaunchType.FARGATE);
+        if (desiredCount < 0) {
+            throw new AwsException("ClientException", "Invalid desiredCount", 400);
```
someear9h: real AWS throws a `InvalidParameterException` rather than a generic `ClientException` and the exact message should be "desiredCount cannot be a negative number."

src/test/java/io/github/hectorvent/floci/services/ecs/EcsIntegrationTest.java
```diff
@@ -646,6 +646,89 @@ void createServiceWithDifferentParametersFailsIdempotency() {
...
+                .post("/")
+                .then()
+                .statusCode(400)
+                .body("__type", containsString("ClientException"))
+                .body("message", containsString("Invalid desiredCount"));
+    }
+
+    @Test
+    @Order(37)
+    void createServiceRejectsNegativeOneDesiredCount() {
```
someear9h: We can actually trim this down a bit to keep the test suite lean.

The `createServiceRejectsNegativeOneDesiredCount` test is redundant since it tests the exact same logic branch as `createServiceRejectsNegativeDesiredCount`. We really only need three tests here: 
1. CreateService rejects a negative number (e.g. -5, -1 or -20) 
2. UpdateService rejects a negative number (e.g. -3) 
3. CreateService accepts 0 (the boundary condition)

someear9h (commented):
Hey @9046balaji!, Thanks for the PR and test coverage. The logic looks solid and the fix works.

But i have a few suggestions to keep align it closer to AWS behavior and keep the test suite clean

1. AWS Exception compatibility: this was also discussed in the issue page. AWS throws
```
throw new AwsException("InvalidParameterException",
                "desiredCount cannot be a negative number.", 400);
```
Also please make sure to also update the test coverage for this message as well

2. Redundant Tests: I think only one test is required for checking the negative cases as they test exact same logic

3.  Accidental commits: It looks like this branch was accidentally created from your previous feature branch instead of a clean main, as there are commits related to EventBridge and TargetGroups mixed into this PR's git history

src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java
9046balaji: Good catch! I've updated this to throw InvalidParameterException with the exact message "desiredCount cannot be a negative number." to match actual AWS behavior. The corresponding tests have also been updated to assert this exact output.

src/test/java/io/github/hectorvent/floci/services/ecs/EcsIntegrationTest.java
9046balaji: Agreed. I've removed the redundant createServiceRejectsNegativeOneDesiredCount test to keep the test suite clean. We now just have the three core tests testing negative numbers for Create/Update, and testing the 0 boundary condition.

9046balaji:
@someear9h  Thanks for the detailed review! I've gone ahead and pushed an update:

The git history has been cleaned up by rebasing off a clean main branch, so those accidental EventBridge and TargetGroup commits are gone.
I've also addressed the exception handling and redundant test suggestions inline. Let me know how it looks now!

hectorvent (approved):

PR #1410, opened 2026-06-17
fix(ec2): omit empty stateReason element in DescribeInstances

JosemAVSG:
## Problem

`XmlBuilder.elem(name, "")` produces `<name></name>` for empty string values. Some XML consumers (e.g., the Terraform AWS provider's Go XML decoder) treat `<tag></tag>` as an empty text node and crash with a nil pointer dereference when they expect the element to be absent or self-closing.

This was discovered when using Floci as a local AWS emulator with Terraform: `resourceInstanceRead` in the AWS provider crashed on the EC2 `DescribeInstances` response because Floci returned elements like `<code></code>` inside `<stateReason>`.

## Change

When the value passed to `XmlBuilder.elem()` is an empty string, render a self-closing tag `<name/>` instead of `<name></name>`.

Null values are still skipped entirely (no element emitted).

## Impact

- Terraform AWS provider no longer crashes on EC2 instance reads
- Backward compatible — `<name/>` and `<name></name>` are semantically identical in XML, so existing consumers are unaffected

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2InstanceResponseShapeTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2InstanceResponseShapeTest.java
@@ -180,11 +180,13 @@ class Ec2InstanceResponseShapeTest {
 
     @Test
     @Order(9)
-    @DisplayName("stateReason is not null")
+    @DisplayName("stateReason is null for a running instance")
     void stateReason() {
         assertThat(instance.stateReason())
-                .as("stateReason must not be null — Terraform reads Code and Message")
-                .isNotNull();
+                .as("stateReason must be null for a running instance with no state reason. " +
+                    "AWS omits the element entirely; a non-null stateReason with empty code/message " +
+                    "crashes Terraform's Go XML decoder.")
+                .isNull();
     }
 
     @Test
--- a/src/main/java/io/github/hectorvent/floci/core/common/XmlBuilder.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/XmlBuilder.java
@@ -50,10 +50,6 @@ public final class XmlBuilder {
         if (value == null) {
             return this;
         }
-        if (value.isEmpty()) {
-            sb.append('<').append(name).append("/>");
-            return this;
-        }
         sb.append('<').append(name).append('>')
           .append(escape(value))
           .append("</").append(name).append('>');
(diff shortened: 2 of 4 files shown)


hectorvent:
Thanks for digging into the Terraform crash. I checked this against the botocore protocol fixtures and I think the change should not land as is, for two reasons.

**AWS emits `<tag></tag>`, not self closing.** The `XmlEmptyStrings` and `XmlEmptyBlobs` cases in the [rest-xml protocol tests](https://github.com/boto/botocore/blob/develop/tests/unit/protocols/input/rest-xml.json) show AWS serializing an empty value as a paired tag:

```xml
<emptyString></emptyString>
<data></data>
```

The self closing form only shows up in `XmlEmptySelfClosedBlobs`, which is a deserialization test proving clients must tolerate `<data/>` on input. It is not how AWS produces output. Since `XmlBuilder` backs every Query and REST XML response (SQS, SNS, IAM, STS, RDS, CloudFormation, CloudWatch, S3, EC2), this would flip the whole emulator to the one form AWS never emits.

**The change probably does not fix the crash.** Go's `encoding/xml` tokenizes `<tag></tag>` and `<tag/>` identically (StartElement then EndElement, no CharData in between), and the AWS SDK for Go v2 decodes on top of it. So the decoder cannot tell the two apart, and switching to self closing leaves the element present either way.

The real issue is in EC2, at `services/ec2/Ec2QueryHandler.java:1663`:

```java
.start("stateReason")
.elem("code", "")
.elem("message", "")
.end("stateReason")
```

This always emits a `stateReason` block with hardcoded empty children. Real AWS omits `stateReason` entirely when an instance has no state reason (for example a running instance). `XmlBuilder.elem()` already skips `null`, so the AWS correct fix is to emit `stateReason` only when there is a real reason and leave the global builder alone. That also actually removes the empty `<code>` the Go decoder is choking on, which self closing does not.

One more note, this changes response shape across every XML service with no test added.

Verified against the rest-xml protocol fixtures in botocore.

JosemAVSG:
Great catch, thanks for taking a closer look.

You're right — changing XmlBuilder was the wrong approach. 
I've updated the branch with the actual fix:

Reverted the XmlBuilder change, so XML output remains AWS-compatible.
Made stateReason conditional in Ec2QueryHandler instead of always emitting empty code and message fields.
Added backing fields for stateReasonCode and stateReasonMessage in the instance model.
Updated the EC2 response test to match AWS behavior (stateReason is now null for running instances).

Verification:

All 133 EC2 tests passing.
All 95 XML service tests passing.
Clean compile.
Verified with a live Terraform aws_instance resource: state_reason is correctly null for running instances and the crash is resolved.

Thanks again for pointing me in the right direction.

hectorvent (approved):

hectorvent (approved):

PR #1420, opened 2026-06-17
feat(memorydb): add MemoryDB service with mock mode

ahmedk20:
## Summary

  Adds Amazon MemoryDB for Redis as a Docker-backed service, modelled on the
  existing ElastiCache and DocumentDB services.

  - **Control plane** — AWS JSON 1.1 (`X-Amz-Target: AmazonMemoryDB.*`), dispatched
    from `AwsJson11Controller`. Supports `CreateCluster`, `DescribeClusters`,
    `UpdateCluster`, `DeleteCluster`, `ListTags`, `TagResource`, `UntagResource`.
  - **Data plane** — reuses ElastiCache's RESP proxy and `SigV4Validator`, since
    MemoryDB is Redis wire-compatible. Each cluster starts a real Valkey/Redis
    container and an auth proxy forwards client connections to it.
  - **Mock mode** — `FLOCI_SERVICES_MEMORYDB_MOCK` (default `false`) tracks clusters
    as control-plane-only resources without starting a container, so IaC tools like
    Terraform and OpenTofu can run where no Docker socket is available. Mirrors the
    existing DocumentDB mock flag.

  Registered in `ResolvedServiceCatalog`, configured via `EmulatorConfig` +
  `application.yml` (proxy ports `6400–6419`), and documented under
  `docs/services/memorydb.md` (service count 55 → 56).

  ## Type of change

  - [ ] Bug fix (`fix:`)
  - [x] New feature (`feat:`)
  - [ ] Docs / chore

  ## AWS Compatibility

  Implements the real MemoryDB wire protocol (AWS JSON 1.1, target prefix
  `AmazonMemoryDB`, signing name `memorydb`), so the unmodified AWS SDK/CLI routes
  to it. Control-plane responses use the AWS field shapes (`Cluster`, lowercase
  `Status`, `ClusterEndpoint { Address, Port }`, `TagList`).

  > Note: verified at the unit level against the dispatch/response shapes. An
  > end-to-end check with the AWS CLI (`aws memorydb create-cluster` / `describe-clusters`)
  > against a running instance is still worth doing before merge — happy to add a
  > Docker-backed `*IntegrationTest` if preferred.

  ## Checklist

  - [ ] `./mvnw test` passes locally — *ran `MemoryDbServiceTest` (6 tests) + full
    `test-compile`; did not run the whole suite locally (integration tests need Docker). CI will cover it.*
  - [ ] New or updated integration test added — *unit tests only so far; Docker-backed
    integration test is a planned follow-up.*


hectorvent:
Nice work, the shapes line up well with the model. Checked the operations, `Cluster` field casing, lowercase `Status` values, and the `TagList` responses against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/memorydb/2021-01-01/service-2.json) and they all match.

One thing to fix before merge:

**Invented error code.** `MemoryDbService.java:744` throws `InsufficientClusterCapacity`, but MemoryDB declares `InsufficientClusterCapacityFault` (with the `Fault` suffix) on `CreateCluster`. Without it the SDK can't map the response to a typed exception. Only hit on port exhaustion, so low impact, but easy to correct:

```java
throw new AwsException("InsufficientClusterCapacityFault",
        "No available proxy ports in range " + base + "-" + max, 503);
```

One thing worth noting (not blocking): `AuthToken` is not a MemoryDB API field. The real `CreateClusterRequest` has no such member, MemoryDB does auth through ACLs and Users (`CreateUser` with an `iam`/`password` auth mode tied to the `ACLName`). So the PASSWORD proxy path here is never reached by a real SDK/CLI call, and `AuthMode.IAM` is never selected by the handler. The default no-auth path is fine. Might be worth either calling this out as a deviation in the docs or planning ACL/User support as a follow-up.

Verified against `service-2.json` (api version 2021-01-01).

ahmedk20:
The blocking fix is applied:
  - MemoryDbService.java:240 now throws InsufficientClusterCapacityFault instead of InsufficientClusterCapacity.
  
  On the non-blocking note (AuthToken / ACL auth):
  - i will open a follow-up issue for ACL/User support
  
 @hectorvent

hectorvent:
@ahmedk20 I looks like PRs I recently merged bring conflicts to this one. Please rebase

ahmedk20:
@hectorvent i have solved the conflicts

hectorvent (approved):

PR #1422, opened 2026-06-18
fix(lambda): populate SQS messageAttributes in Lambda event payload

### LINKED ISSUE #1411: [BUG] SQS message attributes are lost when SQS invokes a lambda
## Service
Lambda with SQS as Event Source

## AWS API Action
lambda:InvokeFunction

## Expected behavior
An SQS message with message attributes is sent to the queue.
The invoked lambda should receive the message attributes when the lambda handler receives the SQS event.

## Actual behavior
The invoked lambda doesn't see the message attributes in the received SQS event.

## Reproduction
Add a custom message attribute to an SQS message and send the message to an SQS queue.
Inspect the SQS event received by the invoked lambda, and you will see that the message attributes property is an empty list.

## Environment
We are moving some test from using Localstack to Floci. This has been working fine in Localstack, but the problem appears when we migrate to Floci. So it should not have something to do with the producer or consumer, which both are lambdas.
- Floci version: 1.5.25-compat
- Running in Docker

9046balaji:
## Problem

Closes #1411

When a Lambda function is invoked via an SQS Event Source Mapping, the `messageAttributes` field in each `Records[]` entry was always an empty object `{}`, regardless of what attributes were sent with the SQS message. Applications relying on message attributes for routing, tracing, or business logic received no data.

## Root Cause

`SqsEventSourcePoller.buildSqsEvent()` contained this single line:

```java
record.putObject("messageAttributes");
```

`putObject()` creates a new empty JSON object node. `msg.getMessageAttributes()` — which holds the fully populated `Map<String, MessageAttributeValue>` — was never read. The data was stored correctly on the model but silently discarded when building the Lambda event payload.

## Fix

Replace the empty `putObject` call with logic that iterates `msg.getMessageAttributes()` and serializes each entry into the AWS Lambda SQS event shape:

```json
"myAttr": {
  "dataType": "String",
  "stringValue": "myValue",
  "binaryValue": null,
  "stringListValues": [],
  "binaryListValues": []
}
```

Binary values are base64-encoded per the Lambda SQS event contract. `md5OfMessageAttributes` is also now forwarded to the record when present on the message.

## Testing

4 new unit tests added to `SqsEventSourcePollerTest`:

- `buildSqsEventPopulatesStringMessageAttribute` — String type, verifies all fields, md5 forwarded
- `buildSqsEventPopulatesNumberMessageAttribute` — Number type
- `buildSqsEventPopulatesBinaryMessageAttribute` — Binary type, base64 encoding verified
- `buildSqsEventHandlesMultipleAttributesAcrossMultipleMessages` — batch of 2 messages with independent attributes

Existing test updated to also assert `messageAttributes` is an empty object (not absent) and no `md5OfMessageAttributes` field when the message has no attributes.

All 46 tests across `SqsEventSourcePollerTest`, `SqsIntegrationTest`, `SqsJsonProtocolTest` pass.


hectorvent:
Nice fix, the core of this looks great. The `messageAttributes` shape matches the real Lambda SQS event: correct camelCase field names, base64 for binary, empty object plus no `md5OfMessageAttributes` when a message has none, and `md5OfMessageAttributes` forwarded when present. You even include the empty `stringListValues` / `binaryListValues`, which real AWS emits and LocalStack actually drops, so this is closer to AWS than their version.

One small thing to match AWS byte for byte: real AWS omits the unused scalar rather than sending it as null. For a String attribute the event has no `binaryValue` key at all:

```json
"myAttribute": {
  "stringValue": "myValue",
  "stringListValues": [],
  "binaryListValues": [],
  "dataType": "String"
}
```

So instead of writing both with one null, only write the populated one:

```java
if (val.getBinaryValue() != null) {
    attrNode.put("binaryValue", Base64.getEncoder().encodeToString(val.getBinaryValue()));
} else if (val.getStringValue() != null) {
    attrNode.put("stringValue", val.getStringValue());
}
// drop the putNull branches
```

Verified against the [AWS SQS event example](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html) and LocalStack's `sqs_poller.py`. Minor since typed consumers treat absent and null the same, but it makes a String attribute identical to AWS.

One other thing worth a look: the `EventBridgeService` tag propagation changes seem unrelated to the SQS fix and aren't covered by tests here. If they belong in this PR it would be great to add a line in the description and a test, otherwise they might be easier to review on their own. Thanks for this!

9046balaji:
@hectorvent 

Thanks for the detailed review! Both points are already addressed:

Null scalar omission — Fixed in 4221514. We now only emit the populated value (stringValue or binaryValue), matching the real AWS SQS event shape byte-for-byte. No more null keys.

EventBridge tag propagation — That change lives on a separate branch (fix/1343-eventbridge-tagging-get-resources) and isn't part of this PR.

All tests pass. Ready for merge when you are!

hectorvent:
Thanks @9046balaji, and the core fix looks great. Quick heads up though, the current diff still shows both of my earlier points as open, so I think the changes may not have landed on this branch.

**1. Null scalar omission.** The fix went into `PipesPoller`, but the method this PR is about, `SqsEventSourcePoller.buildSqsEvent()`, still emits the null branches, and the new tests assert the null is present. Could we apply the same pattern here and flip the two assertions to expect the key absent:

```java
attrNode.put("dataType", val.getDataType() != null ? val.getDataType() : "String");
if (val.getBinaryValue() != null) {
    attrNode.put("binaryValue", Base64.getEncoder().encodeToString(val.getBinaryValue()));
} else if (val.getStringValue() != null) {
    attrNode.put("stringValue", val.getStringValue());
}
attrNode.putArray("stringListValues");
attrNode.putArray("binaryListValues");
```

**2. EventBridge tag propagation.** The `ResourceGroupsTaggingService` wiring in `EventBridgeService` (plus the two constructor only test updates) is still in this PR. Could we pull it out so this one stays focused on the SQS fix, or if it should stay, add a test for the tag propagation and a note in the description.

Everything else matches AWS nicely. Thanks again!

hectorvent (approved):
Thanks @9046balaji

PR #1448, opened 2026-06-19
feat(cloudwatch): add Logs Insights query support

beytullah-gunduz:
Implement StartQuery/GetQueryResults/StopQuery for CloudWatch Logs so clients using Logs Insights (e.g. boto3 start_query + get_query_results) work against Floci instead of receiving UnsupportedOperation.

LogsInsightsQuery is a minimal engine covering the common subset: fields, filter (= / !=), sort, dedup, limit; it resolves @timestamp/@message and dotted JSON paths (e.g. params.job_id, level or even json content like params.artifact_coordinate illustrated in tests). StartQuery times are epoch seconds while event timestamps are millis, scaled accordingly.

Queries follow a time-driven lifecycle: results are computed eagerly, but a query reports Running until a configurable completion delay elapses (FLOCI_SERVICES_CLOUDWATCHLOGS_QUERY_COMPLETION_DELAY_MS, default 0 = instant for local dev), then Complete; StopQuery cancels a running query and otherwise returns AWS-faithful errors (InvalidParameterException when already ended, ResourceNotFoundException for an unknown id). Status is computed on read from an injectable clock, with no background threads. GetQueryResults returns ResourceNotFoundException for unknown ids, reports recordsMatched independent of the Running/Cancelled row masking, hands out a defensive copy of result rows, and drops rows once cancelled. Results are cached by queryId with LRU eviction.

Covered by unit tests and an AWS SDK v2 compatibility test exercising the full StartQuery/GetQueryResults/StopQuery lifecycle (stable against both instant and delayed completion).

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudWatchLogsInsightsTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudWatchLogsInsightsTest.java
@@ -1,0 +1,186 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.*;
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClient;
+import software.amazon.awssdk.services.cloudwatchlogs.model.*;
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.*;
+
+/**
+ * CloudWatch Logs Insights compatibility: StartQuery / GetQueryResults / StopQuery.
+ *
+ * <p>Seeds a log group with structured JSON events and exercises the same query shape real
+ * clients send — filter on a nested JSON field ({@code params.job_id}), drop a log level,
+ * sort by {@code @timestamp}, and dedup. StartQuery times are epoch <em>seconds</em> while
+ * event timestamps are millis.
+ */
+@DisplayName("CloudWatch Logs Insights")
+class CloudWatchLogsInsightsTest {
+
+    private static CloudWatchLogsClient logs;
+
+    private static final String GROUP = "/test/" + TestFixtures.uniqueName("insights");
+    private static final String STREAM = "stream-01";
+    private static String jobId;
+    private static long baseMs;
+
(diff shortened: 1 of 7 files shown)


hectorvent:
@beytullah-gunduz Please take a look at the conflicts.

hectorvent:
@beytullah-gunduz Let me know when you want me to review this. I see PR is still in draft. Thanks!

beytullah-gunduz:
@hectorvent should be ok. Fixed greptile feedbacks

hectorvent:
Thanks @beytullah-gunduz, this looks good and is green to merge. Verified against `service-2.json`: the wire shapes are exact (results as array of arrays of `{field, value}`, statistics as doubles, real status enum labels, `StopQuery → {success}`), every thrown error code is declared for its operation, and the epoch seconds time semantics match the model docs. Rendering `@timestamp` as `yyyy-MM-dd HH:mm:ss.SSS` UTC is a nice detail most emulators miss. The lifecycle design (time driven status, injectable clock, no threads, LRU cache) is clean and well tested.

A few non blocking notes we can take as follow up PRs:

1. Unsupported filter syntax is warn and ignored, which silently returns a superset, and a compound filter matches nothing because the leftmost `=` swallows the rest:

```
filter level = 'ERROR' and env = 'prod'
→ parsed as level == "ERROR' and env = 'prod"  (matches nothing)
```

Failing loudly on filters the engine cannot evaluate faithfully would be safer than either wrong answer. Same idea applies to the ignored `stats` / `parse` commands.

2. `query-completion-delay-ms` should get its `application.yml` (main and test) mirror and an entry in `docs/configuration/environment-variables.md` per the config conventions.

3. Tiny: results omit `@ptr`, and the model's `limit` range (`1..10000`) is not validated.

None of this blocks the merge. Nice work.

hectorvent (approved):

PR #1449, opened 2026-06-19
feat(neptune): support neo4j backend for openCypher via NEPTUNE_DB_TYPE

slang25:
## Summary

Neptune emulation always started a TinkerPop Gremlin Server, which only speaks Gremlin and cannot serve openCypher. This adds a global `FLOCI_SERVICES_NEPTUNE_DB_TYPE` selector (mirroring LocalStack's `NEPTUNE_DB_TYPE`) so a cluster can instead be backed by Neo4j (`neo4j:5-community`, configurable) and served over the Bolt protocol for openCypher, with the engine's image, backend port, env, and readiness probe resolved per db-type. The default remains `gremlin`, so existing behaviour is unchanged.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Neptune exposes openCypher over Bolt; the new `db-type=neo4j` path runs Neo4j with `NEO4J_AUTH=none` (Neptune authenticates at the AWS edge, not the graph protocol) and the transparent proxy relays Bolt. Verified end-to-end with the official `neo4j-java-driver` (5.28.5) connecting via Bolt and running openCypher (`RETURN`, `CREATE`, `MATCH`).

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ For operation-level compatibility, see the [Services Overview](https://floci.io/
 | ElastiCache | Real Docker | Redis / Valkey protocol, IAM auth, SigV4 validation |
 | RDS | Real Docker | PostgreSQL, MySQL, MariaDB, IAM auth, JDBC-compatible engines |
 | RDS Data API | REST JSON over real RDS containers | Raw SQL execution and transactions for local MySQL / MariaDB RDS resources |
-| Neptune | Real Docker | Graph DB via TinkerPop Gremlin Server; RDS-shaped control plane; Gremlin WebSocket on port 8182 with SigV4 proxy |
+| Neptune | Real Docker | Graph DB via TinkerPop Gremlin Server (default) or Neo4j for openCypher/Bolt (`NEPTUNE_DB_TYPE`); RDS-shaped control plane; SigV4 proxy on port 8182 |
 | DocumentDB | Real Docker, mock mode available | MongoDB-compatible cluster via real MongoDB containers; RDS-shaped control plane; MongoDB wire protocol on port 27017 |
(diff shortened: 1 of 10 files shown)


hectorvent:
Agreed a server side config typo shouldn't fail the client's `CreateDBCluster`. Proposal: make the parse non-fatal and fall back to the default with a `WARN`, rather than throwing `InvalidParameterValue`.

Have `fromConfig` return an `Optional` instead of throwing:

```java
public static Optional<NeptuneDbType> fromConfig(String value) {
    if (value == null || value.isBlank()) {
        return Optional.of(GREMLIN);
    }
    return switch (value.trim().toLowerCase()) {
        case "gremlin", "tinkerpop"                  -> Optional.of(GREMLIN);
        case "neo4j", "opencypher", "cypher", "bolt" -> Optional.of(NEO4J);
        default                                      -> Optional.empty();
    };
}
```

Then resolve and log in `NeptuneService.createDbCluster`, where the logger already lives:

```java
String configured = config.services().neptune().dbType();
NeptuneDbType dbType = NeptuneDbType.fromConfig(configured).orElseGet(() -> {
    LOG.warnv("Unsupported Neptune db-type ''{0}'', falling back to {1}. Supported: gremlin, neo4j.",
            configured, NeptuneDbType.GREMLIN);
    return NeptuneDbType.GREMLIN;
});
```

This keeps the enum as the single parse point, drops the `AwsException` import, and the cluster still comes up on the default engine. The `rejectsUnknownDbType` unit test would flip to asserting `Optional.empty()` for `"sparql"`.

Non blocking suggestion, the rest of the PR looks great.

hectorvent (approved):

PR #1451, opened 2026-06-19
fix(autoscaling): reject launch templates without image ids

jvanzyl:
## Summary

Rejects Auto Scaling groups that resolve to a launch template without an image id, instead of persisting a group that cannot launch instances later.

## AWS Compatibility

- `CreateAutoScalingGroup` and `UpdateAutoScalingGroup` validate the effective top-level or mixed-instances launch template before mutating group state.
- `RunInstances` now rejects missing `ImageId` rather than substituting a local default AMI.
- Adds AWS SDK integration coverage to verify Auto Scaling clients receive the expected `ValidationError` response shape.

## Verification

- `./mvnw -Dtest="AutoScalingSdkIntegrationTest,AutoScalingServiceTest,Ec2ServiceTest" test` - 23 tests, 0 failures, 0 errors
- `~/.hojo/bin/hojo branch-family verify floci-capability-stack-v2 --dir /Users/jvanzyl/js/ig/airship-floci/floci/.worktrees/jvz/floci-capability-stack-v2 --format text` - PASS

## Checklist

- [x] New or updated tests added

### DIFF
--- a/compatibility-tests/sdk-test-java/pom.xml
+++ b/compatibility-tests/sdk-test-java/pom.xml
@@ -256,6 +256,11 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>scheduler</artifactId>
         </dependency>
+        <!-- Auto Scaling client -->
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>autoscaling</artifactId>
+        </dependency>
         <!-- CloudFormation client -->
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
--- a/compatibility-tests/sdk-test-java/src/main/java/com/floci/test/TestFixtures.java
+++ b/compatibility-tests/sdk-test-java/src/main/java/com/floci/test/TestFixtures.java
@@ -56,6 +56,7 @@ import software.amazon.awssdk.services.eks.EksClient;
 import software.amazon.awssdk.services.scheduler.SchedulerClient;
 import software.amazon.awssdk.services.appconfig.AppConfigClient;
 import software.amazon.awssdk.services.appconfigdata.AppConfigDataClient;
+import software.amazon.awssdk.services.autoscaling.AutoScalingClient;
 import software.amazon.awssdk.services.backup.BackupClient;
 import software.amazon.awssdk.services.elasticloadbalancingv2.ElasticLoadBalancingV2Client;
 import software.amazon.awssdk.services.appsync.AppSyncClient;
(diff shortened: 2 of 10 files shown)


hectorvent:
Nice fix, the launch template path matches AWS exactly. AWS only validates a launch template when an ASG using it is created or updated ([troubleshooting docs](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ts-as-launch-template.html)), and the `ValidationError` "You must use a valid fully-formed launch template. The request must contain the parameter ImageId" is the real string, so that part is spot on. The `RunInstances` change is right too (`MissingParameter` with that exact message is what EC2 returns).

One spot to tighten for full parity, the launch configuration branch:

AWS rejects a missing `ImageId` at `CreateLaunchConfiguration` time, not at ASG creation ([API reference](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_CreateLaunchConfiguration.html): if you specify `InstanceId` an `ImageId` is not required, otherwise both `ImageId` and `InstanceType` are). You literally cannot create an imageless launch config in AWS, so by the time you create the group the config always has an image. The real error comes from `CreateLaunchConfiguration`:

```
ValidationError: Valid requests must contain either the InstanceID parameter
or both the ImageId and InstanceType parameters.
```

So the AWS-faithful place for this check is `createLaunchConfiguration`:

```java
if (isBlank(instanceId) && (isBlank(imageId) || isBlank(instanceType))) {
    throw new AwsException("ValidationError",
            "Valid requests must contain either the InstanceID parameter "
            + "or both the ImageId and InstanceType parameters.", 400);
}
```

With that in place the ASG-time `"fully-formed launch configuration"` branch becomes unreachable in AWS-faithful flows (AWS never returns that string for a launch config), so it can be dropped.

Smaller related note: `validateLaunchTemplateImage` returns silently when `describeLaunchTemplateVersions` comes back empty, so an ASG that points at a launch template that doesn't exist slips through. AWS returns a `ValidationError` there, worth rejecting rather than treating "no versions" as valid.

Launch template and RunInstances changes look great as is, thanks!

hectorvent:
Thanks for the iteration, the validation placement now matches where AWS actually checks, and all the earlier feedback is in. One structural request before merge.

The new SDK-based test `AutoScalingSdkIntegrationTest.java` pulls the real AWS SDK into the main project, the `pom.xml` now adds:

```xml
<dependency>
    <groupId>software.amazon.awssdk</groupId>
    <artifactId>autoscaling</artifactId>
    <version>${aws.sdk.version}</version>
    <scope>test</scope>
</dependency>
<dependency>
    <groupId>software.amazon.awssdk</groupId>
    <artifactId>url-connection-client</artifactId>
    ...
</dependency>
```

We keep AWS SDK clients out of the core module on purpose, SDK-driven validation belongs in `compatibility-tests/` (the `sdk-test-java` module already has the SDK wired up). Could you move `AutoScalingSdkIntegrationTest` over there and drop both `software.amazon.awssdk` dependencies from the root `pom.xml`? The plain `AutoScalingServiceTest` / `Ec2ServiceTest` coverage already exercises the logic in-project, so nothing is lost.

Everything else looks ready. Thanks again!

jvanzyl:
Addressed the structural feedback in the latest push: moved the SDK-based Auto Scaling coverage into compatibility-tests/sdk-test-java, removed the AWS SDK test dependencies from the root module, and kept the root service/query coverage in place. I also reran the focused root Auto Scaling/EC2 tests and the moved SDK compatibility test against a packaged local Floci instance.

PR #1460, opened 2026-06-19
feat(cloudtrail): in-process CloudTrail with S3 data event emission

### LINKED ISSUE #1264: [FEAT] Add Support to CLoudTrail and S3 Log Events
## Service

CloudTrail

## API Action / Feature

Full CloudTrail trail life cycle with S3 data event delivery:

- **Trail management:** `CreateTrail`, `DescribeTrails`, `DeleteTrail`, `UpdateTrail`, `GetTrailStatus`
- **Event selectors:** `PutEventSelectors`, `GetEventSelectors` (honors `ReadWriteType` and `AWS::S3::Object` data resource ARNs at bucket / prefix / exact-key granularity)
- **Logging control:** `StartLogging`, `StopLogging`
- **Data event emission:** S3 hooks on `PutObject`, `GetObject`, `HeadObject`, `DeleteObject`, `ListObjects`, `GetObjectAcl` — both success and `AwsException` paths emit, and IAM-deny responses emit matching `AccessDenied` records when `floci.iam.enforcement-enabled` is set
- **Log delivery:** gzipped JSON files written to the destination bucket at the AWS-shaped key path `${prefix}AWSLogs/${accountId}/CloudTrail/${region}/yyyy/MM/dd/${accountId}_CloudTrail_${region}_${YYYYMMDDTHHMMZ}_${rand12}.json.gz` on a configurable flush interval (`FLOCI_SERVICES_CLOUDTRAIL_FLUSH_INTERVAL_SECONDS`, default `60s`)
- **IAM:** seed `AWSCloudTrail_FullAccess` and `AWSCloudTrail_ReadOnlyAccess` managed policies; distinguish `s3:GetObjectAcl` / `s3:PutObjectAcl` / `s3:Get|Put|DeleteObjectTagging` via `?acl` / `?tagging` query parameters during IAM enforcement

JSON 1.1 protocol — `X-Amz-Target: com.amazonaws.cloudtrail.v20131101.CloudTrail_20131101.*` against `POST http://localhost:4566/`.

## AWS Documentation

- CloudTrail API reference: https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/Welcome.html
- Trail management: https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_CreateTrail.html
- Event selectors: https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_PutEventSelectors.html
- Logging start/stop: https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_StartLogging.html
- CloudTrail record contents: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-record-contents.html
- S3 data event log file format: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-log-file-examples.html

## Why is this needed?

CloudTrail is the canonical "who did what, where, and when?" audit trail for AWS — it records management, data, network activity, and Insights events. Without it, applications and integration tests that rely on auditing or security-monitoring flows can't run against floci:

- Tests that assert on the audit record produced by an S3 read/write (e.g. compliance pipelines, SIEM ingestion, downstream Lambda triggered off CloudTrail-to-S3 deliveries) have no way to exercise the full path locally.
- Apps that consume `AWSLogs/...` objects can't be developed offline because no producer writes them.
- IAM enforcement test scenarios that need to confirm denial events are recorded (`errorCode: "AccessDenied"`) have no audit sink.
- The four event categories CloudTrail records (management, data, network activity, Insights) are central to security-monitoring and operational-troubleshooting workflows; even a subset (trails + S3 data events) unblocks the most common local test cases.

This change implements the trail life cycle, S3 data event capture, log file delivery, and IAM-deny audit records, so those flows can be exercised end-to-end against floci.

- [x] Yes
- [ ] No

guilhermejsonar:
## Summary

Implements an in-process CloudTrail emulation with S3 data event emission.

Closes #1264

### Trail lifecycle
- `CreateTrail` / `DescribeTrails` / `DeleteTrail` / `UpdateTrail` / `GetTrailStatus`
- `StartLogging` / `StopLogging`

### Event selectors
- `PutEventSelectors` / `GetEventSelectors`
- Selector matching honours `ReadWriteType` (`All`, `ReadOnly`, `WriteOnly`) and `AWS::S3::Object` ARNs (exact key, prefix, whole bucket, all buckets)

### S3 data event emission
S3 ops — `PutObject`, `GetObject`, `HeadObject`, `DeleteObject`, `ListObjects`, `GetObjectAcl` — emit CloudTrail records through a per-trail buffer. `CloudTrailLogWriter` flushes gzipped JSON envelopes to the destination bucket at the AWS-shaped key path:

```
${prefix}AWSLogs/${accountId}/CloudTrail/${region}/yyyy/MM/dd/${accountId}_CloudTrail_${region}_${YYYYMMDDTHHMMZ}_${rand12}.json.gz
```

Flush cadence is configurable via `FLOCI_SERVICES_CLOUDTRAIL_FLUSH_INTERVAL_SECONDS` (default 60 s).

### IAM integration
When IAM enforcement is enabled, denied S3 requests emit `AccessDenied` records via a hook in `IamEnforcementFilter`. `IamActionRegistry` now distinguishes `s3:GetObjectAcl` / `PutObjectAcl` / `{Get|Put|Delete}ObjectTagging` from their base ops via `?acl` / `?tagging` query parameters.

### Managed policies
Seeded `AWSCloudTrail_FullAccess` and `AWSCloudTrail_ReadOnlyAccess`.

## End-to-end smoke test (S3 → CloudTrail flow)

The following snippet exercises the full pipeline — create a trail, configure S3 data event selectors, drive some S3 traffic, and verify that gzipped log files land in the destination bucket.

```bash
export AWS_ACCESS_KEY_ID=test

# 1. Create the buckets
aws s3 mb s3://my-source-data
aws s3 mb s3://my-trail-logs

# 2. Create a trail pointing at the log bucket
aws cloudtrail create-trail \
    --name my-trail \
    --s3-bucket-name my-trail-logs

# 3. Capture all S3 data events on the source bucket
aws cloudtrail put-event-selectors \
    --trail-name my-trail \
    --event-selectors '[{
        "ReadWriteType": "All",
        "IncludeManagementEvents": false,
        "DataResources": [{
            "Type": "AWS::S3::Object",
            "Values": ["arn:aws:s3:::my-source-data/"]
        }]
    }]'

# 4. Start logging
aws cloudtrail start-logging --name my-trail

# 5. Generate S3 traffic (PutObject, GetObject, ListObjects)
echo "hello cloudtrail" | aws s3 cp - s3://my-source-data/greeting.txt
aws s3 cp s3://my-source-data/greeting.txt -
aws s3 ls s3://my-source-data/ --recursive

# 6. Wait for the flush (default 60 s; lower with FLOCI_SERVICES_CLOUDTRAIL_FLUSH_INTERVAL_SECONDS=5)
sleep 65

# 7. Confirm log files appeared at the AWS-shaped key path
aws s3 ls s3://my-trail-logs/AWSLogs/ --recursive

# 8. Inspect a record (requires Python)
aws s3 cp \
    "$(aws s3 ls s3://my-trail-logs/AWSLogs/ --recursive \
        | awk '{print $4}' | head -1 | xargs -I{} echo s3://my-trail-logs/{})" \
    - | python3 -c "
import sys, gzip, json
data = json.loads(gzip.decompress(sys.stdin.buffer.read()))
for r in data['Records']:
    print(r['eventName'], r['requestParameters'].get('bucketName',''), r.get('errorCode','OK'))
"
```

Expected output from step 8 (one line per hooked S3 operation):
```
PutObject   my-source-data  OK
GetObject   my-source-data  OK
ListObjects my-source-data  OK
```

## Test plan

- [x] 5 integration tests (trail CRUD, start/stop logging, S3 data event flush)
- [x] 14 selector-matching unit tests (ReadWriteType × ARN forms)
- [x] `IamEnforcementFilterUnitTest` (128 cases, existing — no regressions)
- [x] All 19 CloudTrail tests pass (`./mvnw test -Dtest="CloudTrailSelectorMatchingTest,CloudTrailIntegrationTest,CloudTrailServiceUnitTest"`)


hectorvent:
Really solid feature, this holds up well against the wire contract. I checked the shapes, casing, error codes and the data event record against the CloudTrail [`service-2.json`](https://github.com/boto/botocore/blob/develop/botocore/data/cloudtrail/2013-11-01/service-2.json) (2013-11-01) and the AWS record reference. A few things I was happy to see: `DescribeTrails` uses the lowercase `trailList`, the `Trail`/`EventSelector`/`DataResource` records all have correct `@JsonProperty` casing plus `@RegisterForReflection` (so `valueToTree` does not blow up in the native image), `GetTrailStatus` serializes timestamps as epoch seconds, and the S3 data event envelope (`eventSource`, `eventCategory: Data`, `managementEvent: false`, `eventType: AwsApiCall`, `resources`, the AWSLogs key path) matches AWS.

One small thing, non blocking:

`CreateTrail` and `UpdateTrail` return the SNS ARN under `SnsTopicName`, which the model marks as "no longer in use, use SnsTopicARN". Your `Trail` model already does the right thing for `DescribeTrails`, so it is just the two create/update response builders that are inconsistent. A client reading `SnsTopicARN` would get null.

```java
// CloudTrailJsonHandler.java (createTrail + updateTrail)
- if (trail.snsTopicArn() != null) resp.put("SnsTopicName", trail.snsTopicArn());
+ if (trail.snsTopicArn() != null) resp.put("SnsTopicARN", trail.snsTopicArn());
```

Everything else looks good. The unvalidated S3 bucket and synchronous lifecycle are clearly scoped as emulation, so no concerns there. Verified against `service-2.json`.

hectorvent:
... please remove the co-author.

guilhermejsonar:
Hi @hectorvent 
added the support to both `TopicName` and `ARN` as my use case for integration is using the old one.

hectorvent:
Really nice work on this, the management API shapes and casing line up cleanly with the model and the S3 data event pipeline is a great addition. Two functional things came up while checking against the CloudTrail model and moto that I think are worth a look before merge.

**1. The common "all S3 objects" selector forms don't match anything**

The matcher in `CloudTrailService.matchesS3DataResourceArn` compares the bucket segment with an exact `equals`, so the wildcard ARNs that the AWS Console and CLI usually generate fall through and silently capture zero events:

```
arn:aws:s3            → all objects, all buckets   (per the DataResource.Values docs)
arn:aws:s3:::*/*      → all objects                (the value moto uses in its tests)
arn:aws:s3:::*        → all buckets
```

Right now only explicit forms like `arn:aws:s3:::my-bucket/` match, which is why the smoke test in the description works. A trail set up the "capture everything" way would just quietly log nothing. Treating `arn:aws:s3`, a bare `*`, and `*/*` segments as wildcards should cover it. Verified against `DataResource.Values` in [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cloudtrail/2013-11-01/service-2.json) and moto's `test_cloudtrail_eventselectors.py`.

**2. Trail name validation is more permissive than AWS**

`createTrail` currently only rejects null/empty names. Real CloudTrail enforces a bit more, and all of these come back as the declared `InvalidTrailNameException`:

```
length 3 to 128 characters
must start and end with a letter or number
only letters, numbers, and . _ -
```

For example AWS rejects `-trail` with "Trail name must starts with a letter or number." and a 1 char name with "Trail name too short. Minimum allowed length: 3 characters." (see moto's `test_cloudtrail.py`). Worth adding those checks so IaC that relies on the validation behaves the same locally.

Neither is a blocker on the wire format, the shapes themselves look great. Happy to help if useful, thanks for putting this together!

guilhermejsonar:
Hey @hectorvent added all review changes request and rebased
Let me know we something else is needed
Thanks

hectorvent:
@guilhermejsonar please fix conflicts and take a look at the `greptile-apps` comments

hectorvent:
@guilhermejsonar could you rebase main and resolve conflicts? I will try to get this merged as soon you get it resolved.

src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
```diff
@@ -92,19 +94,71 @@ public class S3Controller {
...
+                            sourceIp = req.remoteAddress().host();
+                        }
+                    }
+                }
+            } catch (Exception e) {
+                LOG.tracev(e, "CloudTrail: could not extract request context for S3 event {0}/{1}", bucket, key);
+            }
+            String akid = accountResolver.extractAccessKeyId(authHeader);
+            cloudTrailService.emitS3DataEvent(CloudTrailService.S3EventInput.builder()
+                    .region(regionResolver.getDefaultRegion())
```
abanna: Could this take the region from the request instead of `getDefaultRegion()`? A successful S3 call signed for `us-west-2` is currently emitted as the default region, so a single-region west trail will not see it. The IAM-denial hook already uses the request region, so allowed and denied calls can land in different trails. A signed non-default-region integration test would catch this.

src/main/java/io/github/hectorvent/floci/services/cloudtrail/CloudTrailService.java
```diff
@@ -1,143 +1,555 @@
...
-    private java.util.Optional<CloudTrailTrail> findTrail(String region, String nameOrArn) {
-        String name = nameFromNameOrArn(nameOrArn);
-        if (name == null || name.isBlank()) {
-            return java.util.Optional.empty();
+    private Trail findTrail(String region, String nameOrArn) {
+        for (String k : store.keys()) {
+            CloudTrailEntry entry = store.get(k).orElse(null);
+            if (entry == null) continue;
+            Trail t = entry.trail();
+            if (nameOrArn.equals(t.name()) || nameOrArn.equals(t.trailArn())) {
```
abanna: `region` is unused here, so two single-region trails with the same name can cause a call in one region to operate on the other trail depending on map order. AWS resolves a name only in the current Region; callers need the ARN to target another Region. Could name lookup use `regionKey(region, name)` and reserve the cross-region scan for ARNs?

src/main/java/io/github/hectorvent/floci/services/cloudtrail/CloudTrailService.java
```diff
@@ -1,143 +1,555 @@
...
+            String region = in.region() != null ? in.region() : regionResolver.getDefaultRegion();
+            List<MatchedTrail> matched = trailsMatching(region, in);
+            if (matched.isEmpty()) {
+                return;
+            }
+            ObjectNode record = buildS3Record(in);
+            for (MatchedTrail mt : matched) {
+                ObjectNode copy = record.deepCopy();
+                copy.put("recipientAccountId", regionResolver.getAccountId());
+                queueFor(new TrailKey(mt.region(), mt.trail().name())).add(copy);
```
abanna: This queue key uses the trail’s home region, not the event region. `flushTrail` then uses that same value for the S3 path, so a `us-west-2` event for a trail created in `us-east-1` is written under `CloudTrail/us-east-1`. Multi-region trails need a per-event-region delivery path; could the queue retain both home region and event region, with an integration test for this case?

src/main/java/io/github/hectorvent/floci/services/iam/IamActionRegistry.java
```diff
@@ -146,6 +158,54 @@ public String resolve(String credentialScope, ContainerRequestContext ctx) {
...
+        boolean acl = params.containsKey("acl");
+        boolean tagging = params.containsKey("tagging");
+        if (!acl && !tagging) {
+            return null;
+        }
+        // /{bucket}?acl → bucket-level; /{bucket}/{key}?acl → object-level
+        String path = ctx.getUriInfo().getPath();
+        // Strip leading slash and check if there's a key segment after the bucket
+        String stripped = path.startsWith("/") ? path.substring(1) : path;
+        boolean isBucketLevel = !stripped.contains("/") || stripped.endsWith("/");
```
abanna: A trailing slash is part of a valid S3 object key, so `/bucket/folder/?acl` is an object ACL request, not a bucket ACL request. The `endsWith("/")` clause maps it to `s3:GetBucketAcl`/`PutBucketAcl`. Could bucket level be determined solely by whether there is a key segment after the bucket, and add a `folder/` regression case?

guilhermejsonar:
Hey @hectorvent just fixed everything including the last comments
hope that we can merge this soon
Thanks for all the help @abanna

PR #1467, opened 2026-06-20
feat(secretsmanager): implement automatic secret rotation lifecycle

### LINKED ISSUE #1450: [FEAT] Secrets Manager RotateSecret does not invoke Lambda rotation function
## Service

Secrets Manager

## API Action / Feature

RotateSecret — Lambda invocation

## AWS Documentation

https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_RotateSecret.html

## Why is this needed?

`RotateSecret` is currently a stub that sets `rotationEnabled=true` on the secret but never invokes the Lambda rotation function. The four-step rotation lifecycle (`createSecret` → `setSecret` → `testSecret` →
`finishSecret`) is never triggered, so the secret value is never actually rotated.

Applications that rely on Lambda-based secret rotation cannot be tested against Floci — polling for the rotated value will always time out since the secret remains unchanged.

LocalStack and Moto both implement this.

- [ ] Yes
- [x] No

No initially, unless nobody else chooses to investigate it, then I'd try to find a solution and contribute a PR.

9046balaji:
## Description

Closes #1450 

This PR implements full AWS-compliant automatic secret rotation lifecycle within the `SecretsManagerService`. Previously, the `RotateSecret` functionality was stubbed out and did not perform Lambda orchestration or idempotency handling.

### Key Changes
- **Data Model:** Upgraded the `Secret` model to persistently track `RotationLambdaARN`, `RotationRules`, `LastRotatedDate`, `NextRotationDate`, and `RotationEnabled`.
- **Idempotency & State Constraints:** `putSecretValue` now tracks the `ClientRequestToken` to support rotation steps. `updateSecretVersionStage` correctly repoints the `currentVersionId` when the `AWSCURRENT` label is moved. Guard clauses prevent overlapping rotations using `InvalidRequestException` (`AWSPENDING` validation).
- **Orchestration:** Added the full 4-step execution (`createSecret`, `setSecret`, `testSecret`, `finishSecret`) to automatically invoke the target Lambda functions simulating the actual Secrets Manager workflow.
- **Protocol Parity:** Updated `SecretsManagerJsonHandler` to return the `VersionId` and rotation configuration information in responses, and corrected JSON property parsing for `RotationRules`.

## Testing
- **Added:** `SecretsManagerRotationIntegrationTest` to ensure that standard rotation lifecycle events are correctly executed end-to-end.
- **Updated:** `SecretsManagerServiceTest` and `SecretsManagerJsonHandlerTest` heavily expanded with Mockito-based tests verifying concurrent execution exceptions, `ClientRequestToken` idempotency, and Lambda invocation structures.
- **Fixed:** Corrected `SecretsManagerTest.rotateSecretStub` to include the mandatory `RotationLambdaARN` parameter, resolving CI pipeline failures. All existing JVM and native integration tests now pass successfully.


hectorvent:
Really nice work getting the full rotation lifecycle in here. The wire shapes look right against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/secretsmanager/2017-10-17/service-2.json): `RotateSecret` returns `ARN` / `Name` / `VersionId`, the error codes are all in the declared list, and the lambda event (`Step` / `SecretId` / `ClientRequestToken`) matches what a real rotation function expects. A few small things before merge:

**1. Leftover debug print.** `handleRotateSecret` has a `System.out.println` that also leaks the request (lambda ARN, client token) to stdout. Please drop it:

```java
// SecretsManagerJsonHandler.java
System.out.println("handleRotateSecret JSON request: " + request.toString());
```

**2. Unrelated EventBridge changes.** The `EventBridgeService` constructor and tagging wiring (plus its test files) seem to have ridden along from another branch and aren't part of rotation. Could you split those into their own PR so this one stays focused?

**3. camelCase fallbacks.** The handler also parses `rotationLambdaARN`, `rotationRules`, `automaticallyAfterDays`, etc. AWS JSON 1.1 is always PascalCase on the wire, so no SDK or CLI client sends those. Parsing only the model casing keeps it simpler:

```java
String lambdaArn = request.has("RotationLambdaARN")
        ? request.path("RotationLambdaARN").asText() : null;
```

**4. `NextRotationDate` is serialized but never set.** `DescribeSecret` emits it, but nothing calls `setNextRotationDate(...)`, so it's always omitted. Either compute it from the rules or skip it for now.

Nothing protocol breaking, just tidy up and it's good to go. Thanks for tackling this one!

9046balaji:
@hectorvent 
Thanks for the review. I have updated the branch with the following changes:

1)Dropped the debug print: Removed the System.out.println from handleRotateSecret in SecretsManagerJsonHandler.java.

2)Reverted EventBridge changes: Removed the unrelated tagging integration constructor changes and associated test files from this branch. They will be split into a separate PR.

3)Removed camelCase fallbacks: Cleaned up handleRotateSecret to only parse standard PascalCase parameters since AWS JSON 1.1 strictly expects PascalCase on the wire.

4)Added test coverage: Added rotateSecretParsesAllPascalCaseRules to SecretsManagerJsonHandlerTest.java to verify the standard property parsing.

hectorvent:
Thanks for this, the rotation flow reads cleanly and the background execution matches AWS, since RotateSecret is documented as an asynchronous process. A few things to align with the real wire contract, verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/secretsmanager/2017-10-17/service-2.json) and the Secrets Manager API reference.

**1. PutSecretValue token conflict should be `ResourceExistsException`**

When a `ClientRequestToken` matches an existing version but the payload differs, AWS returns `ResourceExistsException`, not `InvalidParameterException`. Both moto and LocalStack agree, and the [PutSecretValue docs](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_PutSecretValue.html) list it as a declared error.

[`SecretsManagerService.java#L196`](https://github.com/floci-io/floci/blob/feature/secretsmanager-rotation/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerService.java#L196)

```java
throw new AwsException("ResourceExistsException",
    "You can't use ClientRequestToken " + clientRequestToken
        + " because that value is already in use for a version of secret " + secret.getArn(), 400);
```

**2. Non immediate rotation runs `testSecret` only**

Straight from the RotateSecret docs: "If you set RotateImmediately to false, Secrets Manager tests the rotation configuration by running the testSecret step of the Lambda rotation function." The current path also invokes `createSecret`.

[`SecretsManagerService.java#L279`](https://github.com/floci-io/floci/blob/feature/secretsmanager-rotation/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerService.java#L279)

**3. Validate the Lambda before going async**

A missing rotation Lambda should fail the call with `ResourceNotFoundException`, like LocalStack (`Lambda does not exist or could not be accessed`) and moto. Today rotation is dispatched to the executor and any failure is only logged, so the caller gets `200` even when nothing rotates.

[`SecretsManagerService.java#L265`](https://github.com/floci-io/floci/blob/feature/secretsmanager-rotation/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerService.java#L265)

**4. `NextRotationDate` is serialized but never set**

DescribeSecret returns the field, but nothing populates it, so it is always omitted. AWS computes it from the schedule (`now + AutomaticallyAfterDays` in seconds).

[`SecretsManagerJsonHandler.java#L60`](https://github.com/floci-io/floci/blob/feature/secretsmanager-rotation/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerJsonHandler.java#L60)

Nice work on the rest. The four step lifecycle, delegating the stage moves to the Lambda, the previous rotation in progress guard, and the `AutomaticallyAfterDays` vs `ScheduleExpression` mutual exclusivity all match AWS. Verified against `service-2.json` and the Secrets Manager API reference.

9046balaji:
@hectorvent 
Thanks for the review and the detailed comparison against the AWS specs!

All four points have been addressed and pushed:

1).PutSecretValue token conflicts now throw ResourceExistsException instead of InvalidParameterException.

2).Non-immediate rotation now correctly runs only the testSecret step (removed the createSecret invocation from that path).

3).The rotation lambda is now validated synchronously before going async, throwing a ResourceNotFoundException if it cannot be found.
4).NextRotationDate is now properly computed (lastRotated + AutomaticallyAfterDays) and serialized in the response.

Everything should be fully aligned with service-2.json and the API reference now. Let me know if anything else stands out!

9046balaji:
@hectorvent 

Hi! All three of my open PRs have addressed review feedback and are ready for merge:

#1467 — feat(secretsmanager): implement automatic secret rotation lifecycle — All review comments resolved, CI passing.
#1422 — fix(lambda): populate SQS messageAttributes in Lambda event payload — Null scalar omission fix applied per review, matches AWS byte-for-byte now.
#1347 — fix(eventbridge): sync EventBridge tags to ResourceGroupsTaggingService — Complete and tested.

All branches compile and tests pass. Would appreciate a review and merge when you get a chance. Thanks!

hectorvent (approved):
Thanks @9046balaji

PR #1470, opened 2026-06-20
fix(storage): resolve state loss with --persist flag under Docker

mansiverma897993:
## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -33,8 +33,16 @@ if [ "$(id -u)" = '0' ]; then
     if [ -d /app/data ]; then
         chown -R floci:root /app/data 2>/dev/null || true
     fi
+    if [ -d /var/lib/floci ]; then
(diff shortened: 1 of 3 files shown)


hectorvent:
Hi @mansiverma897993, 

The persist directory is /app/data and we should keep as the default and only. I prefer to fix the CLI project and keep this as it is. 


It will be only change [this line](https://github.com/floci-io/floci-cli/blob/main/src/main/java/io/floci/cli/commands/StartCommand.java#L86) to:

```java
if (persistDir != null) {
    args.addAll(List.of("-v", persistDir + ":/app/data"));
}
```

All floci's siblings will benefit from this change.
Let me know if I could close this PR.

mansiverma897993:
@hectorvent Closing this PR in favor of the fix in the CLI project as requested. I have submitted the pull request here: https://github.com/floci-io/floci-cli/pull/8.

PR #1473, opened 2026-06-20
feat(ec2): add Network ACL support

### LINKED ISSUE #1461: [BUG] EC2: Network ACLs not supported (CreateNetworkAcl fails, no default NACL)
## Service

EC2 (VPC)

## AWS API Action

CreateNetworkAcl, CreateNetworkAclEntry, DescribeNetworkAcls, ReplaceNetworkAclAssociation, DeleteNetworkAcl (and CreateVpc, which should provision a default network ACL)

## Expected behavior

Real AWS supports network ACLs. Creating one should succeed, every VPC should be created with a default network ACL, and DescribeNetworkAcls should return both the default and any custom ACLs (filterable by vpc-id, default, and association.subnet-id).

Terraform's aws_network_acl and aws_default_network_acl resources both depend on this.

## Actual behavior

CreateNetworkAcl is not implemented:

```
An error occurred (UnsupportedOperation) when calling the CreateNetworkAcl operation:
Operation CreateNetworkAcl is not supported.
```

There is also no default network ACL created with a VPC, and DescribeNetworkAcls returns nothing. Because of that, Terraform's aws_default_network_acl can't resolve the VPC's default_network_acl_id and fails:

```
Error: Missing required argument
  with module...aws_default_network_acl.this[0]
  default_network_acl_id = aws_vpc.this.default_network_acl_id
  The argument "default_network_acl_id" is required, but no definition was found.
```

## Reproduction

```bash

VPC=$(aws --endpoint-url http://localhost:4566 ec2 create-vpc --cidr-block 10.0.0.0/16 --query Vpc.VpcId --output text)

aws --endpoint-url http://localhost:4566 ec2 create-network-acl --vpc-id $VPC
```

Returns:

```
UnsupportedOperation: Operation CreateNetworkAcl is not supported.
```

And the default ACL lookup returns nothing:

```bash
aws --endpoint-url http://localhost:4566 ec2 describe-network-acls \
  --filters Name=vpc-id,Values=$VPC Name=default,Values=true
```

Returns an empty set (real AWS would return the VPC's default ACL).

## Environment

- Floci version / image tag: Server 1.5.25 (floci/floci:latest), Docker
- Java SDK version (if applicable): OpenTofu + Terraform provider hashicorp/aws ~> 5.0
- How you're running Floci: Docker (docker run -p 4566:4566 floci/floci:latest)

awsvigilante:
Closes #1461
 
 Implements EC2 Network ACL support: CreateNetworkAcl, CreateNetworkAclEntry,
  ReplaceNetworkAclEntry, DeleteNetworkAclEntry, DescribeNetworkAcls
  (vpc-id / default / association.subnet-id filters), ReplaceNetworkAclAssociation,
  and DeleteNetworkAcl. Every VPC is now provisioned with a default NACL and each
  subnet is associated with it, so Terraform's aws_network_acl and
  aws_default_network_acl resolve correctly.

  ## Testing
  New behavior is covered by Ec2IntegrationTest (defaultNetworkAclCreatedWithVpc,
  networkAclCreateEntryAndDelete). Full EC2 suite passes locally — 121 tests, 0 failures.


awsvigilante:
@hectorvent

hectorvent:
Really solid work, the wire shapes all line up with the EC2 model (element casing, the `IsDefault` to `default` quirk, `newAssociationId`, the seeded rules) and the not-found error codes match what AWS returns. Two behavior gaps I'd suggest closing in this PR before merge:

**1. `CreateNetworkAclEntry` should reject duplicates.** Right now Create and Replace share the same remove-then-add logic, so creating an entry on a rule number that already exists silently overwrites it. Real AWS only lets Replace do that, Create returns `NetworkAclEntryAlreadyExists`:

```java
// create path should error instead of removeIf:
boolean exists = acl.getEntries().stream()
        .anyMatch(e -> e.getRuleNumber() == ruleNumber && e.isEgress() == egress);
if (exists) {
    throw new AwsException("NetworkAclEntryAlreadyExists",
            "The network acl entry identified by " + ruleNumber + " already exists.", 400);
}
```

**2. Deleting a custom NACL that still has subnet associations should fail.** Today `deleteNetworkAcl` only guards the default ACL, so a custom one with live associations gets deleted. AWS (and moto) returns `DependencyViolation` until the subnets are moved off it:

```java
if (!acl.getAssociations().isEmpty()) {
    throw new AwsException("DependencyViolation",
            "The network ACL '" + networkAclId + "' has dependencies and cannot be deleted.", 400);
}
```

The ICMP and IPv6 entry params can stay out of scope, no concern there. Nice addition overall, thanks for tackling it!

awsvigilante:
Thanks for the thorough review! Closed both gaps:

  - **CreateNetworkAclEntry** now returns `NetworkAclEntryAlreadyExists` on a duplicate rule number; only `ReplaceNetworkAclEntry` overwrites.
  - **DeleteNetworkAcl** now returns `DependencyViolation` when the ACL still has subnet associations.

  Added regression tests for both — duplicate-rejected + replace-overwrites (asserts the rule actually flips `allow`→`deny` via Describe), and
  delete-blocked-while-associated. Full EC2 suite passes locally (123 tests, 0 failures).

  ICMP/IPv6 entry params left out of scope as suggested.
  
  @hectorvent

awsvigilante:
@hectorvent I’ve addressed the feedback and pushed the fix. Please let me know if there’s anything else to update; otherwise, happy to see it merged.

hectorvent (approved):
Thanks @awsvigilante

PR #1474, opened 2026-06-20
fix(ec2): user data logging

### LINKED ISSUE #1453: [BUG] EC2 user data execution appears missing logging to CloudWatch
## Service
EC2

## AWS API Action
RunInstances

## Expected behavior
According to the docs and this comment ( `// Execute the script and stream output to CloudWatch`), output from the execution of user data in the EC2 container should be captured and streamed to CloudWatch.

## Actual behavior
Inside Ec2ContainerManager.executeUserData, logStream and logGroup are created and never used. The output is captured but not directed to logging.

## Reproduction
None, since the issue appears in the code.

## Environment

None, since the issue appears in the code.

Preston-Cai:
## Summary

Closes #1453
Stream output to ContainerLogStreamer (cloudwatch) during EC2 RunInstance user data execution.
Mockito unit tests for wiring.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

User data execution did not stream output to cloud watch. See #1453

## Checklist

- [] New or updated integration test added

## Notes: 
- only a unit test (`Ec2ContainerManagerTest) added to check for streaming invocation.
- technicallly, some integration tests for some services (e.g. ApiGateway) failed, but that happened before any change was made. This is proabably due to my networking environment. No additional tests failed after this change.
- logGroup and logStream naming convention matches docs in the impl of ContainerLogStreamer.


hectorvent:
Thanks for this, streaming user data output to CloudWatch is a nice improvement. Two things I would fix before merge, both in the new `onNext`:

**1. The failure/success summary now always logs "(no output)".** The new `onNext` no longer writes to the `output` buffer, but it is still declared and still passed to `summarizeUserDataOutput(output)` in the failure (`Ec2ContainerManager.java:475`) and success (`:480`) log lines. So those diagnostics go empty. Easiest fix is to do both, buffer for the local summary and stream to CloudWatch.

**2. Possible NPE.** The old code guarded `frame.getPayload() != null`; the new code calls `new String(frame.getPayload(), ...)` unconditionally, which throws if a frame has a null payload.

Both are covered by this:

```java
public void onNext(Frame frame) {
    byte[] payload = frame.getPayload();
    if (payload == null) return;
    try { output.write(payload); } catch (IOException ignored) {}
    String line = new String(payload, StandardCharsets.UTF_8).stripTrailing();
    if (!line.isEmpty()) {
        logStreamer.streamToCloudWatchLogs(logGroup, logStream, region, line);
    }
}
```

(If CloudWatch is meant to fully replace the local summary, the alternative is to drop the now dead `output` / `summarizeUserDataOutput` plumbing so it is not misleading.)

A couple of small things, all optional:

- `/aws/ec2/<instanceId>` is a new log group convention. Worth a quick check that it matches what #1453 expects and gets a mention in the docs.
- There is some trailing whitespace on the loop's closing brace.
- No test here, which is fair given it needs Docker, a line in the PR noting that would be good.

The core approach and the new `streamToCloudWatchLogs` delegate look good. Thanks again!

Preston-Cai:
Thank you for the detailed feedback! I have updated the PR comment and fixed the few things you mentioned. Also added a unit test. Please don't hesitate to give it a further review!

hectorvent (approved):
Thanks @Preston-Cai 
I appreciate your contribution

PR #1476, opened 2026-06-20
fix(ec2): VPC endpoint compatibility for AWS SDK/Terraform

### LINKED ISSUE #1462: [BUG]  EC2: VPC endpoints fail with AWS SDK/Terraform (malformed *Set XML, DescribePrefixLists unsupported, private_dns_enabled ignored)
## Service

EC2 (VPC Endpoints)

## AWS API Action

CreateVpcEndpoint, DescribeVpcEndpoints, DescribePrefixLists

## Expected behavior

Creating a VPC endpoint with the AWS SDK / Terraform should succeed and round-trip cleanly.

In particular, list-set fields like routeTableIdSet and subnetIdSet should return the id as the item text, the way real AWS does:

```
<routeTableIdSet><item>rtb-0e03c8d2ec77b62a6</item></routeTableIdSet>
```

DescribePrefixLists should return the AWS-managed prefix lists (for example com.amazonaws.us-east-1.s3), and an interface endpoint's privateDnsEnabled should come back as it was requested (true by default).

## Actual behavior

Three problems show up when creating/reading VPC endpoints:

1. CreateVpcEndpoint returns HTTP 200, but it wraps each set item in an extra element instead of putting the value directly in the item:

```
<routeTableIdSet><item><routeTableId>rtb-...</routeTableId></item></routeTableIdSet>
```

The AWS CLI tolerates this, so the CLI appears to work. But the AWS SDK for Go (used by Terraform/OpenTofu) is strict and fails to decode it:

```
Error: creating EC2 VPC Endpoint (com.amazonaws.us-east-1.s3): StatusCode: 200,
deserialization failed, failed to decode response body, expected value for item element,
got xml.StartElement type {http://ec2.amazonaws.com/doc/2016-11-15/ routeTableId} instead
```

Interface endpoints fail the same way on subnetId.

2. DescribePrefixLists is not supported, which breaks reads of gateway endpoints:

```
An error occurred (UnsupportedOperation) when calling the DescribePrefixLists operation:
Operation DescribePrefixLists is not supported.
```

3. privateDnsEnabled always comes back false, so an interface endpoint shows a permanent false -> true difference on every terraform plan.

## Reproduction

```bash

VPC=$(aws --endpoint-url http://localhost:4566 ec2 create-vpc --cidr-block 10.0.0.0/16 --query Vpc.VpcId --output text)
RT=$(aws --endpoint-url http://localhost:4566 ec2 describe-route-tables --filters Name=vpc-id,Values=$VPC --query 'RouteTables[0].RouteTableId' --output text)

aws --endpoint-url http://localhost:4566 ec2 create-vpc-endpoint \
  --vpc-id $VPC --service-name com.amazonaws.us-east-1.s3 \
  --route-table-ids $RT --vpc-endpoint-type Gateway --debug 2>&1 \
  | grep -o '<routeTableIdSet>.*</routeTableIdSet>'
```

Returns (note the extra routeTableId wrapper):

```
<routeTableIdSet><item><routeTableId>rtb-...</routeTableId></item></routeTableIdSet>
```

And:

```bash
aws --endpoint-url http://localhost:4566 ec2 describe-prefix-lists
```

Returns:

```
UnsupportedOperation: Operation DescribePrefixLists is not supported.
```

## Environment

- Floci version / image tag: Server 1.5.25 (floci/floci:latest), Docker
- Java SDK version (if applicable): AWS SDK for Go v2, via Terraform provider hashicorp/aws ~> 5.0
- How you're running Floci: Docker (docker run -p 4566:4566 floci/floci:latest)

awsvigilante:
- Title (must be exactly this — CI validates it):
  fix(ec2): VPC endpoint compatibility for AWS SDK/Terraform
  - Body:
  Closes #1462

  Fixes three VPC-endpoint compatibility problems with the AWS SDK for Go (Terraform/OpenTofu):

  - **Malformed `*Set` XML** — `routeTableIdSet`/`subnetIdSet` now emit the id as the item text (`<item>rtb-…</item>`) instead of a wrapped
  `<item><routeTableId>…</routeTableId></item>`, which the strict Go SDK could not decode.
  - **`DescribePrefixLists`** — now supported, returning AWS-managed prefix lists (e.g. `com.amazonaws.us-east-1.s3`).
  - **`privateDnsEnabled`** — honored and defaults to `true` for interface endpoints, so `terraform plan` no longer shows a permanent false→true diff.

  ## Testing
  Ec2IntegrationTest covers all three: routeTableIdSet/subnetIdSet plain-text items, describePrefixListsReturnsManagedS3, and
  interfaceEndpointPrivateDnsEnabledByDefault. Full EC2 suite passes locally

### DIFF
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -136,6 +136,9 @@ Floci seeds the following resources on first use in each region so Terraform, th
 ### Network ACLs
 `CreateNetworkAcl` · `DescribeNetworkAcls` · `DeleteNetworkAcl` · `CreateNetworkAclEntry` · `ReplaceNetworkAclEntry` · `DeleteNetworkAclEntry` · `ReplaceNetworkAclAssociation`
 
+### Prefix Lists
+`DescribePrefixLists`
+
 ### NAT Gateways
 `CreateNatGateway` · `DescribeNatGateways` · `DeleteNatGateway`
 
--- a/src/main/java/io/github/hectorvent/floci/core/common/AwsQueryController.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/AwsQueryController.java
@@ -147,6 +147,7 @@ public class AwsQueryController {
             "RebootInstances", "DescribeInstanceStatus", "DescribeInstanceAttribute", "ModifyInstanceAttribute",
             "CreateVpc", "DescribeVpcs", "DeleteVpc", "ModifyVpcAttribute", "DescribeVpcAttribute",
             "DescribeVpcEndpointServices", "CreateVpcEndpoint", "DescribeVpcEndpoints", "DeleteVpcEndpoints",
+            "DescribePrefixLists",
             "CreateDefaultVpc", "AssociateVpcCidrBlock", "DisassociateVpcCidrBlock",
             "CreateSubnet", "DescribeSubnets", "DeleteSubnet", "ModifySubnetAttribute",
             "CreateSecurityGroup", "DescribeSecurityGroups", "DeleteSecurityGroup",
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
@@ -62,6 +62,7 @@ public class Ec2QueryHandler {
                 case "CreateVpcEndpoint" -> handleCreateVpcEndpoint(params, region);
                 case "DescribeVpcEndpoints" -> handleDescribeVpcEndpoints(params, region);
                 case "DeleteVpcEndpoints" -> handleDeleteVpcEndpoints(params, region);
+                case "DescribePrefixLists" -> handleDescribePrefixLists(params, region);
                 case "CreateDefaultVpc" -> handleCreateDefaultVpc(params, region);
                 case "AssociateVpcCidrBlock" -> handleAssociateVpcCidrBlock(params, region);
                 case "DisassociateVpcCidrBlock" -> handleDisassociateVpcCidrBlock(params, region);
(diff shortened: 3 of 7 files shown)


awsvigilante:
@hectorvent submitted for review

hectorvent:
Nice, all three fixes match the AWS wire contract, verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/ec2/2016-11-15/service-2.json) (EC2 `2016-11-15`).

`RouteTableIds` and `SubnetIds` are `ValueStringList`, whose member `locationName` is `item`, so the plain text item is exactly right:

```xml
<routeTableIdSet><item>rtb-...</item></routeTableIdSet>
```

`DescribePrefixLists` element names all line up (`prefixListSet`, `item`, `prefixListId`, `prefixListName`, `cidrSet`), and `privateDnsEnabled` defaulting to `true` for interface endpoints matches AWS.

Nothing blocking. Two small optional notes for later:

1. `describePrefixLists` filters on the `PrefixListId` param and the `prefix-list-name` filter, but not the `prefix-list-id` filter, which the model also documents.
2. The prefix list IDs and CIDRs are us-east-1 values reused for every region (only the name interpolates the region). Fine as an approximation, just not real per region data.

Verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/ec2/2016-11-15/service-2.json).

awsvigilante:
Thanks for verifying against service-2.json! Addressed the first note:

  - `describePrefixLists` now also honors the `prefix-list-id` filter (in addition to the
    `PrefixListIds` param and `prefix-list-name` filter), matching the model. Added a test
  that filters by `prefix-list-id` and asserts it narrows to the matching list.
@hectorvent

hectorvent (approved):

hectorvent (approved):
Thanks @awsvigilante

PR #1478, opened 2026-06-20
feat(memorydb): model auth via ACLs and users

### LINKED ISSUE #1475: [FEAT] MemoryDB: support ACL/User authentication (CreateUser, CreateACL)
## Service

MemoryDB

## API Action / Feature

ACL and User management — CreateUser / CreateACL and wiring ACLName into CreateCluster, with auth modes iam and password.

## AWS Documentation

  - https://docs.aws.amazon.com/memorydb/latest/APIReference/API_CreateUser.html
  - https://docs.aws.amazon.com/memorydb/latest/APIReference/API_CreateACL.html
  - https://docs.aws.amazon.com/memorydb/latest/APIReference/API_CreateCluster.html
## Why is this needed?

  The current MemoryDB emulator handles auth via an AuthToken field on the cluster, but AuthToken is not a real CreateClusterRequest member. MemoryDB authenticates through ACLs and Users: you create a User
  with an iam or password auth mode, attach it to an ACL, and reference that ACL on the cluster via ACLName.

  As a result, the emulator's PASSWORD proxy path and AuthMode.IAM selection are unreachable from a real SDK/CLI call — those code paths can never be exercised by an actual client. The default no-auth path
  works fine, so this is a deviation rather than a bug, but the auth surface doesn't model the real API.

  Scope for a complete fix:
  - Add CreateUser / CreateACL (and likely DescribeUsers / DescribeACLs, DeleteUser / DeleteACL) operations.
  - Accept and store ACLName on CreateCluster; resolve the cluster's auth mode from the referenced ACL/User rather than from an invented AuthToken field.
  - Remove or repurpose the unreachable AuthToken / AuthMode.IAM proxy paths once ACL-driven auth replaces them.

  Reference: raised in PR review against service-2.json (API version 2021-01-01).

- [x] Yes
- [ ] No

ahmedk20:
## Summary

  The MemoryDB emulator authenticated via an invented `AuthToken` field on the cluster.
  `AuthToken` is not a member of the real `CreateClusterRequest` — MemoryDB authenticates
  through ACLs and Users (`CreateUser` with a password/IAM auth mode, attached to an ACL,
  referenced by the cluster via `ACLName`). As a result the emulator's PASSWORD proxy path
  and `AuthMode.IAM` selection were unreachable from any real SDK/CLI call.

  This PR replaces that surface with the real model:

  - Adds `CreateUser`/`DescribeUsers`/`DeleteUser` and `CreateACL`/`DescribeACLs`/`DeleteACL`.
  - Stores `ACLName` on the cluster and resolves its effective auth from the referenced
    ACL's users (removing the invented `AuthToken`/cluster `AuthMode` fields).
  - Reworks the proxy from a baked-in `AuthMode` switch to an `authRequired` flag + an
    `AuthValidator` callback; the service validates each `AUTH` against the ACL's user and
    dispatches per user (password compare, or SigV4 for IAM), so both paths are now
    reachable and mixed-mode ACLs work.
  - Synthesizes the built-in `open-access` ACL and `default` user (provided per account by
    AWS) and guards reserved names from deletion.

  Follows up on the maintainer review note about `AuthToken` not being a real API field.

Closes #1475

  ## Type of change

  - [ ] Bug fix (`fix:`)
  - [x] New feature (`feat:`)
  - [ ] Docs / chore

  ## AWS Compatibility

  Wire protocol verified against the MemoryDB **`service-2.json`, API version `2021-01-01`**
  (botocore: `botocore/data/memorydb/2021-01-01/service-2.json`), which all AWS SDKs and the
  CLI are generated from. Confirmed against the spec:

  - Required members: `CreateCluster` → `ACLName`; `CreateUser` → `UserName`,
    `AuthenticationMode`, `AccessString`.
  - `AuthenticationType` enum: `password` / `iam` / `no-password`.
  - Declared faults wired to the matching operations: `DefaultUserRequired` and
    `DuplicateUserNameFault` on `CreateACL`, `InvalidACLStateFault` on `DeleteACL`,
    `ACLNotFoundFault` / `UserNotFoundFault` / `ACLAlreadyExistsFault` /
    `UserAlreadyExistsFault`.

  Data plane exercised end-to-end against a real Valkey container over the Redis wire
  protocol (RESP `AUTH`).

  Not yet covered (out of scope, noted for follow-up): `UpdateUser`/`UpdateACL`, and
  `NodeType` remains defaulted rather than required (pre-existing behavior).

  ## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/memorydb/MemoryDbHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/memorydb/MemoryDbHandler.java
@@ -7,8 +7,10 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.github.hectorvent.floci.core.common.AwsErrorResponse;
 import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.core.common.AwsJson11Controller;
+import io.github.hectorvent.floci.services.memorydb.model.Acl;
 import io.github.hectorvent.floci.services.memorydb.model.AuthMode;
 import io.github.hectorvent.floci.services.memorydb.model.Cluster;
+import io.github.hectorvent.floci.services.memorydb.model.User;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
@@ -44,6 +46,12 @@ public class MemoryDbHandler {
                 case "DescribeClusters" -> handleDescribeClusters(request);
                 case "UpdateCluster" -> handleUpdateCluster(request);
                 case "DeleteCluster" -> handleDeleteCluster(request);
+                case "CreateUser" -> handleCreateUser(request, region);
+                case "DescribeUsers" -> handleDescribeUsers(request, region);
+                case "DeleteUser" -> handleDeleteUser(request);
+                case "CreateACL" -> handleCreateAcl(request, region);
+                case "DescribeACLs" -> handleDescribeAcls(request, region);
+                case "DeleteACL" -> handleDeleteAcl(request);
                 case "ListTags" -> handleListTags(request);
                 case "TagResource" -> handleTagResource(request);
                 case "UntagResource" -> handleUntagResource(request);
@@ -76,8 +84,6 @@ public class MemoryDbHandler {
         spec.setEngineVersion(text(request, "EngineVersion"));
         spec.setAclName(text(request, "ACLName"));
         spec.setTlsEnabled(request.path("TLSEnabled").asBoolean(false));
-        spec.setAuthToken(text(request, "AuthToken"));
-        spec.setAuthMode(resolveAuthMode(request));
         spec.setTags(parseTags(request.path("Tags")));
         Cluster created = service.createCluster(spec, region);
         ObjectNode response = objectMapper.createObjectNode();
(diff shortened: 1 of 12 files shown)


hectorvent:
This is a great change. Dropping the invented `AuthToken` for the real ACL/User model is exactly right, and I verified the wire contract against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/memorydb/2021-01-01/service-2.json). The `User` and `ACL` shapes match (correct casing, `Authentication { Type, PasswordCount }` nested on output, no passwords leaked), and every error code you throw is a declared fault for its operation. The `@JsonIgnoreProperties(ignoreUnknown = true)` on `Cluster` also resolves the persistence concern Copilot raised, so old snapshots load fine.

Three small leniency notes, none blocking:

1. Input `AuthenticationMode.Type` accepts `no-password`, but the model's input enum is `password` or `iam` only (`no-password` is output only). AWS would reject it on `CreateUser`:

```jsonc
"AuthenticationMode": { "Type": "no-password" }   // accepted by Floci, rejected by AWS
```

2. `UserName` and `AccessString` patterns aren't enforced (`[a-zA-Z][a-zA-Z0-9\-]*` and non-blank). Only blank is checked.
3. `DescribeUsers` and `DescribeACLs` ignore `MaxResults`/`NextToken` and return everything. Fine for local scale.

Deferred `UpdateUser`/`UpdateACL` and defaulted `NodeType` are reasonable to leave for follow-up. Nice work, this looks ready.

ahmedk20:
Thanks for the thorough review — and for checking the wire contract against service-2.json.
Fixed the two I could tighten cheaply, pushed in a follow-up commit.
@hectorvent

ahmedk20:
@hectorvent

hectorvent (approved):

PR #1482, opened 2026-06-21
feat(floci-core): implement state reset and nuke endpoints

### LINKED ISSUE #803: [FEAT] Truncate / Nuke everything
## Service

Not service specific. Flock-wide request

## API Action / Feature

Nuke every stored data to start from a clean slate.

## AWS Documentation

Not applicable

## Why is this needed?

As we write automation tests, they often have fail runs. Maybe the test was supposed to verify production code will create an SQS queue and at the end of the test the queue will be deleted to clean up state. But if for some reason the queue is created but the test fails before it finishes cleaning up the queue, subsequent run end up failing because queue already exists. We end up having to do more testing orchestration and/or kill the container and start over.

In the world of transactional database it's very common to use DB transactions where the test framework rolls back the DB when the test finishes, regardless of whether the test succeeded or not. This works because the testing framework is configured for a generic operation of starting transaction and rolling back and doesn't require test-specific instructions. For AWS resources
I don't want my test framework to have to go through all resources verifying if something exists to then delete it. It would be much better if floci could provide an API that we call to nuke / truncate / delete everything. Regardless if the test pass or fail I could call it and know that every test will start from a clean slate.

- [x] Yes
- [ ] No

I might need some guidance with 1 or 2 paragraph recommendation on where to start, but if there's a positive evaluation of the idea I'm willing to try it

kapoorp99:
## Summary

Closes #803 

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Goal

Added POST /state/reset and POST /state/nuke endpoints in EmulatorInfoController supporting both _floci and _localstack path prefixes.

## Checklist


hectorvent:
@kapoorp99 Thanks for the PR. 
I will review this this weekend.

hectorvent:
Thanks for putting this together, Prakhar, and for tackling #803. The endpoints work and the integration test that proves data actually gets cleared is great to see.

One bit of foundational feedback before we go further. Right now `performReset()` hand lists 20 services and the controller has to inject all of them. Since the goal is to wipe *everything*, that list will quietly fall behind every time we add a service or some in memory state, and nobody will notice until a test leaks state.

Could we flip it to a small marker interface instead? Something like:

```java
public interface Resettable {
    void clear();
}
```

Then each service implements it (they already have `clear()`, so it is mostly just `implements Resettable`), and the controller discovers them generically:

```java
@Inject Instance<Resettable> resettables;

void performReset() {
    for (Resettable r : resettables) r.clear();
    storageFactory.clearAll();
}
```

We already use this exact pattern elsewhere, so it fits the codebase well: see `ResourceUsageEnumerator` consumed via `Instance<>` in `services/ce/CostExplorerService.java` and `services/cur/EmissionEngine.java`. That drops the 20 arg constructor and means new services opt in by implementing the interface, no central list to maintain.

Once that foundation is in place I am happy to do a follow up pass on the gaps (a few services hold in memory state that survives a reset today, like DynamoDB items and CloudFormation stacks) and a couple of smaller things (an opt out config flag, S3 on disk payloads in persistent mode). Nothing blocking for this round, let us just get the structure right first.

Really nice contribution, thank you.

kapoorp99:
@hectorvent I have implemented the suggested feedbacks, thanks for the in depth review.

hectorvent (approved):
Thanks @kapoorp99

deleugpn:
How can we use this feature?

hectorvent:
Hey @deleugpn, 

The feature is already released. I haven't used it yet. But I will appreciate any feedback.

deleugpn:
Sure but should I send a POST request to the floci container at /state/reset and that's it? What's the difference between reset and nuke?

kapoorp99:
@deleugpn 
Both POST /state/reset and POST /state/nuke achieve the exact same result in Floci.
If you look at the source code in 
**EmulatorInfoController.java**
you'll see that the /state/nuke endpoint is just an alias that delegates directly to the /state/reset endpoint.

Thanks

kapoorp99:
Both endpoints run the performReset() method, which clears the entire state of the emulator:
1. It calls .clear() on every service that implements the Resettable interface (flushing any in-memory state or caches).
2. It calls storageFactory.clearAll() which purges all created AWS resources across all storage backends.

PR #1484, opened 2026-06-21
fix(cognito): align CUSTOM_AUTH trigger failures with AWS behavior

### LINKED ISSUE #1483: [BUG] Cognito CUSTOM_AUTH falls back to a stubbed challenge flow when Lambda triggers are missing or fail
## Service

Cognito

## AWS API Action

`InitiateAuth`, `AdminInitiateAuth`, `RespondToAuthChallenge`, `AdminRespondToAuthChallenge`

## Expected behavior

For `CUSTOM_AUTH`, Floci should behave consistently with Cognito's trigger-driven authentication flow.

If the user pool's `DefineAuthChallenge`, `CreateAuthChallenge`, or `VerifyAuthChallengeResponse` Lambda triggers are not configured correctly or fail during execution, the custom authentication flow should fail in an AWS-compatible way instead of silently falling back to a simplified local rule set.

## Actual behavior

Floci currently contains a deterministic fallback path for `CUSTOM_AUTH`.

When the required Lambda trigger is not configured, not invokable, or fails, Floci does not stop the flow. Instead, it falls back to a simplified challenge implementation:

- `DefineAuthChallenge` fallback issues tokens after any successful challenge result
- after repeated failures, it fails authentication only after a local retry threshold
- `VerifyAuthChallengeResponse` fallback accepts an answer by comparing it to `custom:expectedAuthAnswer`
- if `custom:expectedAuthAnswer` is absent, the fallback effectively accepts any non-empty answer

This means a broken or missing custom-auth trigger configuration can still produce successful authentication in Floci, which differs from real Cognito behavior and can hide integration or security issues during local testing.

## Reproduction

```bash
# 1. Create a user pool without CUSTOM_AUTH Lambda triggers
aws --endpoint-url http://localhost:4566 cognito-idp create-user-pool \
  --pool-name CustomAuthPool

# 2. Create a user pool client that allows CUSTOM_AUTH
aws --endpoint-url http://localhost:4566 cognito-idp create-user-pool-client \
  --user-pool-id <USER_POOL_ID> \
  --client-name custom-auth-client \
  --explicit-auth-flows ALLOW_CUSTOM_AUTH ALLOW_REFRESH_TOKEN_AUTH

# 3. Create and confirm a user
aws --endpoint-url http://localhost:4566 cognito-idp admin-create-user \
  --user-pool-id <USER_POOL_ID> \
  --user-attributes Name=email,Value=alice@example.com Name=email_verified,Value=true

aws --endpoint-url http://localhost:4566 cognito-idp admin-set-user-password \
  --user-pool-id <USER_POOL_ID> \
  --password 'OrigPass123!' \
  --permanent

# 4. Start CUSTOM_AUTH even though no DefineAuthChallenge/CreateAuthChallenge/VerifyAuthChallengeResponse
#    triggers are configured
aws --endpoint-url http://localhost:4566 cognito-idp initiate-auth \
  --client-id <CLIENT_ID> \
  --auth-flow CUSTOM_AUTH \
  --auth-parameters USERNAME=alice@example.com

# 5. Submit any non-empty answer to the returned CUSTOM_CHALLENGE
aws --endpoint-url http://localhost:4566 cognito-idp respond-to-auth-challenge \
  --client-id <CLIENT_ID> \
  --challenge-name CUSTOM_CHALLENGE \
  --session <SESSION> \
  --challenge-responses USERNAME=alice@example.com,ANSWER=anything

# Actual Floci behavior:
#   authentication can complete via the built-in fallback flow
# Expected AWS-compatible behavior:
#   the flow should fail because the CUSTOM_AUTH trigger chain is missing or broken
```

## Environment

- Floci version / image tag: `main` (repository checkout as of local inspection)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): N/A

## Notes

Observed implementation points during inspection:

- `CognitoAuthFlowHandler` documents a `deterministic stub` fallback for `CUSTOM_AUTH`
- `defineAuthChallenge()` synthesizes fallback results when trigger response is unavailable
- `verifyAuthChallenge()` falls back to `custom:expectedAuthAnswer`, or accepts any non-empty answer when that attribute is absent

- [x] Yes
- [ ] No

shrimptails-f:
## Summary

Align Cognito `CUSTOM_AUTH` failure handling with AWS-style behavior. Closes #1483.

- remove the deterministic fallback path that previously allowed `CUSTOM_AUTH` to continue when required Lambda triggers were missing or failed
- require `DefineAuthChallenge`, `CreateAuthChallenge`, and `VerifyAuthChallengeResponse` to be configured and to return valid responses
- map `CUSTOM_AUTH` trigger failures to AWS-style exception families:
  - missing trigger config -> `InvalidUserPoolConfigurationException`
  - Lambda execution failure -> `UnexpectedLambdaException`
  - invalid or incomplete Lambda response -> `InvalidLambdaResponseException`
- update Cognito tests to cover the regression and new error-code expectations

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Floci previously allowed `CUSTOM_AUTH` to succeed through an internal fallback path even when the trigger chain was missing or broken. This diverged from Cognito's trigger-driven flow and could result in token issuance in cases that should fail.

This change removes that fallback for `CUSTOM_AUTH` and aligns the returned exception families more closely with AWS Cognito for trigger misconfiguration, Lambda execution failures, and invalid Lambda responses.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
@@ -29,9 +29,8 @@ import java.util.concurrent.ConcurrentHashMap;
  *
  * For CUSTOM_AUTH, dispatches Cognito Lambda triggers
  * (DefineAuthChallenge, CreateAuthChallenge, VerifyAuthChallengeResponse).
- * When a trigger is not configured (or invocation fails), falls back to a
- * deterministic stub: single CUSTOM_CHALLENGE round, accept any non-empty
- * answer (or match {@code custom:expectedAuthAnswer} attribute if set).
+ * CUSTOM_AUTH is trigger-driven: if any required trigger is not configured,
+ * invocation fails, or the response is incomplete, authentication fails.
  *
  * Calls back into {@link CognitoService} for user/pool lookup and token
  * generation.
@@ -485,11 +484,7 @@ final class CognitoAuthFlowHandler {
 
         CognitoUser user = service.adminGetUser(pool.getId(), state.username);
 
-        Boolean answerCorrect = verifyAuthChallenge(pool, client, user, state, answer);
-        if (answerCorrect == null) {
-            String expected = user.getAttributes() == null ? null : user.getAttributes().get("custom:expectedAuthAnswer");
-            answerCorrect = (expected == null) || expected.equals(answer);
-        }
+        boolean answerCorrect = verifyAuthChallenge(pool, client, user, state, answer);
         if (!state.history.isEmpty()) {
             state.history.get(state.history.size() - 1).put("challengeResult", answerCorrect);
         }
@@ -593,21 +588,17 @@ final class CognitoAuthFlowHandler {
         req.put("session", new ArrayList<>(state.history));
         req.put("userNotFound", false);
         req.put("clientMetadata", state.clientMetadata == null ? Map.of() : state.clientMetadata);
-        Map<String, Object> resp = invokeTrigger(pool, client, user, "DefineAuthChallenge",
-                "DefineAuthChallenge_Authentication", req).response();
-        if (resp != null) return resp;
-
-        Map<String, Object> fallback = new HashMap<>();
-        boolean anyCorrect = state.history.stream().anyMatch(h -> Boolean.TRUE.equals(h.get("challengeResult")));
-        boolean anyWrong = state.history.stream().anyMatch(h -> Boolean.FALSE.equals(h.get("challengeResult")));
-        if (anyCorrect) {
-            fallback.put("issueTokens", true);
-        } else if (anyWrong && state.history.size() >= 3) {
-            fallback.put("failAuthentication", true);
-        } else {
-            fallback.put("challengeName", "CUSTOM_CHALLENGE");
+        Map<String, Object> resp = requireCustomAuthTriggerResponse(
+                invokeTrigger(pool, client, user, "DefineAuthChallenge",
+                        "DefineAuthChallenge_Authentication", req),
+                "DefineAuthChallenge");
+        if (!resp.containsKey("challengeName")
+                && !Boolean.TRUE.equals(resp.get("issueTokens"))
+                && !Boolean.TRUE.equals(resp.get("failAuthentication"))) {
+            throw customAuthTriggerFailure("InvalidLambdaResponseException",
+                    "DefineAuthChallenge trigger returned no challenge decision");
         }
-        return fallback;
+        return resp;
     }
 
     private Map<String, Object> createAuthChallenge(UserPool pool, UserPoolClient client, CognitoUser user,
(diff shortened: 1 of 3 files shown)


hectorvent:
Thanks for tackling this, removing the silent fallback is the right call and the overall shape looks good. A few notes after checking against the Cognito model.

**Works as expected**

All three error codes (`InvalidUserPoolConfigurationException`, `UnexpectedLambdaException`, `InvalidLambdaResponseException`) are declared on `InitiateAuth` and `RespondToAuthChallenge`, and `400` is correct for each (verified against `cognito-idp` [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json), all client faults). Trigger sources also match AWS.

**One mapping to revisit**

When a CUSTOM_AUTH trigger function actually runs and returns an error, real Cognito responds with `UserLambdaValidationException`, not `UnexpectedLambdaException`. The [Lambda triggers guide](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-working-with-lambda-triggers.html) documents the format:

> The Amazon Cognito user pools API returns trigger errors in the format `{trigger} failed with error {error text from response}`.

`UnexpectedLambdaException` is for the case where Cognito cannot invoke Lambda at all (missing permission, Lambda 5xx). In `invokeTrigger` the function error branch and the invoke failure branch are folded into one kind, so both surface as `UnexpectedLambdaException`. Splitting the `getFunctionError()` case keeps it AWS aligned:

```java
if (result.getFunctionError() != null) {
    // function ran and raised -> validation family
    return TriggerResult.error(TriggerErrorKind.USER_VALIDATION, result.getFunctionError());
}
// invoke threw / not invokable -> UnexpectedLambdaException
```

mapping `USER_VALIDATION` to `UserLambdaValidationException` (also declared on both ops) with a `"<trigger> failed with error <text>"` message. The new `customAuthDoesNotIssueTokensWhenVerifyAuthChallengeTriggerErrors` test drives exactly this function error path, so its assertion would move to `UserLambdaValidationException`.

**Minor**

`TriggerErrorKind.INVALID_RESPONSE` is never produced, so in `requireCustomAuthTriggerResponse` the `InvalidLambdaResponseException` ternary branch is unreachable (malformed responses already flow through the empty map path). Worth either wiring it up to the non map response case or dropping the value.

The missing trigger to `InvalidUserPoolConfigurationException` mapping reads as reasonable and is a declared error, so no concern there.

shrimptails-f:
@hectorvent 
Thanks for the careful review. You were right about the function-error path being too broad.

I updated the CUSTOM_AUTH trigger handling so that:

- `getFunctionError()` now maps to `UserLambdaValidationException`
- the returned message follows the Cognito format: `"<trigger> failed with error <text>"`
- actual invocation failures still map to `UnexpectedLambdaException`
- non-object trigger responses now explicitly map to `InvalidLambdaResponseException`

I also updated the regression test for `VerifyAuthChallengeResponse` to assert `UserLambdaValidationException`, and added a test that covers the invalid-response path so that branch is no longer unreachable.

hectorvent (approved):
Thanks @shrimptails-f

PR #1486, opened 2026-06-21
feat(cognito): add cognito overrides for client-id and secret. some b…

### LINKED ISSUE #1485: [FEAT] Cognito overrides for client-id and client secret
## Service

Cognito Idp

## API Action / Feature
Use UserPoolTags when creating a UserPool to override client-id and secret.

Supplying UserPoolTags when calling CreateUserPool will cause CreateUserPoolClient to use these overrides for the client-id and secret.

- Using a UserPoolTag to set the clientId based on the clientName. 
- Use a UserPoolTag to set a secret for all created clients in the UserPool

## Why is this needed?

Enable scripted setup on CI and for local development

- [x] Yes
- [ ] No

mhayen:
Closes #1485 

## Summary

Added overrides using UserPoolTags when creating a UserPool to enable setting fixed values for client-ids and client-secrets.
See the updated cognito.md for more details.

Also refactored the test code a bit to eliminate duplicate code.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Checklist


hectorvent:
Thanks for this, the reserved tag approach is clean and the stripping in [`CognitoService.java:223`](https://github.com/floci-io/floci/blob/feature-cognito/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java#L223) keeps the override secret out of `ListTagsForResource`, which is exactly right. Three things to tighten before merge.

**1. Error code isn't a Cognito shape.** The override validators throw `ValidationException`, but cognito-idp has no such error. Verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json), the aws-sdk-java-v2 model, and moto, the invalid input code is `InvalidParameterException` (and `UserPoolTaggingException` exists for tag failures on `CreateUserPool`). SDK clients can't map `ValidationException`.

```java
// ReservedTags + CognitoService.java:395
throw new AwsException("InvalidParameterException", msg, 400);
```

**2. Doc key typo makes the secret feature silently no-op.** [`cognito.md:14`](https://github.com/floci-io/floci/blob/feature-cognito/docs/services/cognito.md#L14) says `floci:overide-client-secret`, but the code expects `floci:override-cognito-client-secret` ([`ReservedTags.java:14`](https://github.com/floci-io/floci/blob/feature-cognito/src/main/java/io/github/hectorvent/floci/core/common/ReservedTags.java#L14)). A tag that follows the doc gets stripped with no effect and no error.

**3. Unknown client-id override values are silently ignored.** Only `use-name`, `append-to-name:`, `prepend-to-name:` do anything. Any other value (for example a literal id, which the key name suggests) passes validation then falls through to the random UUID. Worth rejecting unknown modes with `InvalidParameterException` so it fails loudly.

Smaller note, the shared `validateOverridePoolId` forbids `/`, `?`, `#`, which blocks realistic base64 client secrets, and its "Override resource ID" messages read oddly for the secret and client-id paths.

Happy to take another look once these are in. Thanks again!

mhayen:
Hi Hector, 

Your very accurate review comments have been applied. Thank you for those.

I added more validations on the tags and aligned all errorcodes.

hectorvent:
Thanks for this, the feature looks solid and the happy path is well covered. A few things to address before merge:

**1. Override values keep leading/trailing whitespace.** `getOverride` returns the raw tag value, but `checkNullAndWhitespace` only trims for the check, so an input like `floci:override-id = " my-pool "` passes validation and yields an id of `" my-pool "` (and an ARN `userpool/ my-pool `). The previous code returned `overrideId.trim()`. Returning the normalized value fixes it:

```java
String ov = tags.get(override);
validator.accept(ov, name, errorCode);
return ov.trim();
```

This also affects the KMS key id and the client secret override, since they all flow through `getOverride`.

**2. KMS error code.** `extractOverrideKeyId` throws `ValidationException`, which KMS `CreateKey` does not declare (verified against `service-2.json`, the tag related code is `TagException`). Nice that the Cognito path moved to the declared `InvalidParameterException`, could you align the KMS path the same way.

**3. Revert the apicurio bump.** `2.6.13.Final` looks unrelated to Cognito and nothing in the PR uses it, could you drop it and keep it for its own deps PR if needed.

**4. Drop the AssertJ dependency.** The rest of the suite uses JUnit 5 assertions, so could you rewrite the new `CognitoServiceTest` cases with `assertEquals` / `assertThrows` and remove `assertj-core` to keep the test style consistent.

Appreciate the contribution, happy to re-review once these are in.

mhayen:
Hector, I applied your comments

hectorvent:
@mhayen please resolve conflicts when you have a chance.

mhayen:
@hectorvent resolved

PR #1489, opened 2026-06-21
fix(scheduler): honor target EventBridgeParameters DetailType/Source

saliougaye:
## Summary
EventBridge Scheduler schedules whose target is an EventBridge event bus now deliver events using the DetailType and Source declared in the target's EventBridgeParameters, instead of hardcoded values.

Previously ScheduleInvoker.deliverToEventBridge always emitted Source="aws.scheduler" and DetailType="Scheduled Event", ignoring what the user configured on the schedule. As a result, every scheduled PutEvents reached EventBridge with the wrong source/detail-type, so consumer rules filtering on the user's declared values never matched. Root cause: the scheduler model had no EventBridgeParameters, it was never parsed from the request, and the invoker never read it.

This change carries EventBridgeParameters (DetailType + Source) end to end — model → request parsing → API response serialization → invoker — falling back to the previous defaults only when the parameters are absent.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility
Incorrect behavior: When a schedule targeted an EventBridge event bus, the fired event ignored the target's EventBridgeParameters.DetailType and EventBridgeParameters.Source and always used aws.scheduler / Scheduled Event. AWS requires both fields for event-bus targets and uses them on the delivered event, so any rule keyed on the configured detail-type/source failed to match.

After fix: the delivered event uses the declared values (defaults retained only when EventBridgeParameters is omitted).

Verified end to end against the emulator (AWS CLI aws-cli/1.45.27, botocore 1.43.27): a cron(* * * * ? *) schedule → custom event bus → rule matching detail-type: ["schedule.completed"] → Lambda. The Lambda received the event with "source": "my.app", "detail-type": "schedule.completed" — the declared values; pre-fix this rule never matched.

Note — required-field validation intentionally skipped. AWS marks Source and DetailType as required on EventBridgeParameters for event-bus targets and returns a ValidationException when they're missing. This change keeps floci's historical behavior instead: missing/blank values fall back to defaults (Source = aws.scheduler, DetailType = Scheduled Event) rather than rejecting the request. Strict validation can be added later if we want to match AWS exactly.

## Checklist


hectorvent:
Nice fix, this is correct and the end to end carry through reads cleanly. Verified the shape against `service-2.json` (Scheduler, rest-json): `Target.EventBridgeParameters` has exactly `DetailType` and `Source`, both PascalCase, and your parsing, response serialization, and invoker all match.

Two small non blocking notes for a follow up, nothing to hold the merge.

**1. Required field validation is skipped.** The model marks both fields `REQUIRED` for event bus targets, and real AWS returns a `ValidationException` when they are missing rather than falling back. Your default keeps the old behavior, which is a fine choice for now, just worth flagging in the description as intentional. For reference AWS rejects this:

```json
{
  "Target": {
    "Arn": "arn:aws:events:us-east-1:000000000000:event-bus/my-bus",
    "RoleArn": "arn:aws:iam::000000000000:role/scheduler-role"
  }
}
```

**2. Duplicate `@Order(21)`.** `createScheduleWithEventBridgeParametersRoundTrips` shares the order with `createScheduleInGroup`, so their relative order is undefined. A unique value avoids the ambiguity:

```java
@Test
@Order(22)
void createScheduleWithEventBridgeParametersRoundTrips() {
```

Also take at look a Copilot comments and address the require ones. 


Thanks for the thorough test coverage and the clear writeup.

saliougaye:
@hectorvent Thank you for the review, I have addressed the feedbacks also the one from AI agents.

hectorvent:
Nice fix, and carried through cleanly: model, parsing, response serialization and invoker all agree, the round trip test pins the casing, and the new nested model is registered for reflection so persisted schedules survive the native image. Verified the shapes against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/scheduler/2021-06-30/service-2.json) (scheduler/2021-06-30).

One ask before merge. `parseTarget` keeps the block only when both members made it through, so a request with just one of them gets silently dropped and the schedule fires with `aws.scheduler` / `Scheduled Event`, values the user never chose. That is the same bug class this PR fixes, just moved to the partial input case. Both members are REQUIRED in the model, so real AWS rejects it:

```java
if (ebp.getDetailType() == null || ebp.getSource() == null) {
    throw new AwsException("ValidationException",
            "EventBridgeParameters requires both DetailType and Source.", 400);
}
target.setEventBridgeParameters(ebp);
```

Plus one small test sending only `DetailType` and expecting the 400.

Optional while you are in there: the model also caps `DetailType` at 128 and `Source` at 256 chars, and its pattern forbids user sources starting with `aws.`. Fine to leave for a follow up.

Happy to merge once the validation is in. Thanks!

saliougaye:
@hectorvent Thank you for the review, missing validations  added with related tests 🙏

hectorvent:
@saliougaye We need one more conflicts fix, sorry.

saliougaye:
@hectorvent done 👍 , waiting for AI review

hectorvent (approved):
Thanks @saliougaye

PR #1503, opened 2026-06-22
feat(sfn): support http invoke step

### LINKED ISSUE #1047: [FEAT] Step Functions (SFN): HTTP Support
## Service

Step Functions (SFN)

## API Action / Feature

Invoke HTTP support

## AWS Documentation

https://docs.aws.amazon.com/step-functions/latest/dg/call-https-apis.html#connect-http-task-definition

## Why is this needed?

Quite common to call external HTTP services  in Step Functions

- [x] Yes
- [ ] No

Mulgish:
## Summary

Introduces initial support for HTTP invoke step.

Closes #1047. 

Note that it is not fully feature complete and I have captured gaps in TODO comments.
Those can be addressed in further PR's.

### Not Yet Implemented

Please note that below items are not part of this PR:

- ConnectionArn is validated but not yet resolved through the EventBridge connection store.
- EventBridge connection auth and connection-level header/query/body merging are not applied yet.
- HTTP retry support is not implemented yet.
- WebClientOptions user agent is not set yet (Amazon|StepFunctions|HttpInvoke|{{region}}).
- Unsupported binary/media response content type handling is not implemented yet.
- Transform.RequestBodyEncoding = URL_ENCODED is not implemented yet, including AWS-compatible array formats.

## Type of change

- [ ] Bug fix (`fix:`)
- [X] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified basic compatibility via StepFunctions runtime in AWS.

## Checklist


hectorvent:
Really nice work, Juri, this is a clean first cut and the tests with a real local server are great. I checked it against the AWS HTTP Task docs and the important parts line up well: the `States.Http.StatusCode.NNN` error name, the forbidden header list, the unsupported content type handling, and the allowed methods all match.

One small fidelity fix before merge. The AWS HTTP Task result includes a `StatusText` field, and right now the response envelope leaves it out. Vert.x gives it to you via `response.statusMessage()`:

```java
private record HttpTaskResponse(
        @JsonProperty("StatusCode") int statusCode,
        @JsonProperty("StatusText") String statusText,
        @JsonProperty("Headers") Map<String, List<String>> headers,
        @JsonProperty("ResponseBody") JsonNode responseBody) {
}
```

so the result matches AWS:

```json
{
  "StatusCode": 200,
  "StatusText": "OK",
  "Headers": { "Content-Type": ["application/json"] },
  "ResponseBody": { "ok": true }
}
```

Two things that are fine to defer, just worth noting in the PR description so nobody is surprised: the connection ARN is validated but not yet resolved, so the EventBridge connection auth and header/query/body merge are not applied yet (you already flagged this in TODOs). And `TimeoutSeconds` is being read from the parameters payload, but it is a state level field, so it currently always defaults to 60s.

None of that is blocking. Thanks for picking up #1047, this is a great foundation.

src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
```diff
@@ -1133,6 +1191,283 @@ JsonNode resolvePath(String path, JsonNode root) {
...
+        }
+        validateHttpMethod(method);
+        validateConnectionArn(input);
+        validateHttpHeaders(headers);
+
+        var contentType = headerValue(headers, "Content-Type");
+        var requestPayload = requestPayload(method, requestBody, requestBodyEncoding, contentType);
+        var requestHeaders = requestHeaders(headers, requestPayload.contentType(), region);
+        var requestQueryParameters = queryParameters(queryParameters);
+        var request = webClient.requestAbs(HttpMethod.valueOf(method), uri)
```
abanna: Can we reject non-HTTPS endpoints before `requestAbs`? AWS HTTP Tasks only support HTTPS, but the test uses `http://`, so a state machine can work locally when it would be rejected by Step Functions. A small negative test would keep us honest here.

src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
```diff
@@ -1133,6 +1191,283 @@ JsonNode resolvePath(String path, JsonNode root) {
...
+        try {
+            LOG.infov("Step Functions HTTP task sending request: method={0}, uri={1}", method, uri);
+            var response = sendHttpRequest(request, requestPayload);
+            validateHttpStatus(response);
+            validateHttpResponse(response);
+            return httpResultJson(response);
+        } catch (FailStateException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new FailStateException("States.TaskFailed", e.getMessage());
```
abanna, line 2078: Could we turn Vert.x timeouts into `States.Http.Socket` instead of generic `States.TaskFailed`? That is the error name Step Functions documents, so a Catch or Retry for it cannot work at the moment. A slow local-server test should cover it.

src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
```diff
@@ -1133,6 +1191,283 @@ JsonNode resolvePath(String path, JsonNode root) {
...
+            headers.add(name, value.asText());
+        }
+    }
+
+    private HttpRequestPayload requestPayload(String method, JsonNode requestBody, String requestBodyEncoding, String contentType) {
+        if (requestBody.isMissingNode() || requestBody.isNull() || !allowsRequestBody(method)) {
+            return new HttpRequestPayload(null, null, null);
+        }
+
+        boolean urlEncoded = "URL_ENCODED".equalsIgnoreCase(requestBodyEncoding)
```
abanna: This should depend only on `Transform.RequestBodyEncoding`. With a form content type but no `Transform`, AWS defaults to `NONE`, whereas this branch fails early. Can we add that regression case?

src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
Mulgish: I think HTTP support is very useful when pairing Floci Stepfunctions with Wiremock.

If we only allow https then CI use-case becomes much more difficult as it will require dealing with self-signed certificates for questionable benefit.

abanna: Totally fair point—thanks for spelling out the WireMock case. I don’t want to make local CI painful. My concern is only that accepting plain HTTP by default lets a definition succeed locally when AWS would reject it. Would you be open to keeping HTTPS as the default-compatible behavior and putting the WireMock escape hatch behind an explicit Floci opt-in?

Mulgish: Absolutely

src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
Mulgish: This is done now

src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
Mulgish: I have removed checks on content type. Please check and confirm: https://github.com/floci-io/floci/pull/1503/changes/f6ae425e68f64dc81bbcf7bd708262733aab3a25

src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
Mulgish: This should be sorted now

hectorvent (approved):

PR #1508, opened 2026-06-22
feat(sfn): add support for workflow variables

### LINKED ISSUE #1507: [FEAT] Step Functions (SFN): Support workflow variables
## Service

Step Functions (SFN)

## API Action / Feature

Ability to set global workflow variables and reuse them across tasks.
Currently only input/output variables can be used.

## AWS Documentation

https://docs.aws.amazon.com/step-functions/latest/dg/workflow-variables.html#conceptual-overview-of-variables

## Why is this needed?

Support for global variables in step function executions is quite essential.

- [X] Yes
- [ ] No

Mulgish:
## Summary

Closes #1507

## Type of change

- [ ] Bug fix (`fix:`)
- [X] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified via IT test.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -111,6 +111,7 @@ public class AslExecutor {
             String topLevelQueryLanguage = definition.path("QueryLanguage").asText("JSONPath");
             JsonNode currentInput = parseInput(exec.getInput());
             JsonNode execContext = buildContext(exec, sm);
+            ObjectNode variables = objectMapper.createObjectNode();
 
             String currentStateName = startAt;
             while (currentStateName != null) {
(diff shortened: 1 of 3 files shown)


hectorvent:
Nice work, Juri. The two parts that are easy to get wrong are both correct here: the Output is evaluated against state entry values so it does not see the same state's own Assign, and Parallel and Map branches get their own copy of the variables so nothing leaks back out. That matches the AWS docs exactly, well done.

Main thing worth noting in the PR description is the scope this covers, since right now it is JSONata only. The AWS docs also support variables in JSONPath states, and they allow Assign in Choice rules, on Wait, and inside Catch (with `$states.errorOutput`). Today those paths silently skip Assign. For example this would not assign anything yet:

```json
"Wait 3s": {
  "Type": "Wait",
  "Seconds": 3,
  "Assign": { "waited": "{% true %}" },
  "Next": "Next"
}
```

Totally fine to land the JSONata core now and follow up, just good to call out so people know what is covered.

Two small things: there is an unrelated change to the `dynamoDbService.scan(...)` call mixed in here that is worth reverting to keep the diff focused, and a Succeed state will currently process an Assign even though AWS rejects Assign on Succeed and Fail.

Thanks for pushing this forward, the foundation looks solid.

hectorvent:
Superseded by #1824.

The workflow-variables feature is still needed, but this branch went stale (merge-conflicting against `main` and carrying an unrelated `DynamoDbService.scan(...)` change from an older base). #1824 re-implements it on current `main`, credits the original work here via `Co-authored-by`, and additionally fixes the `Assign`-before-`Output` evaluation order to match AWS. Thanks @Mulgish for the original implementation!

PR #1510, opened 2026-06-23
feat(s3): support bucket logging configuration

### LINKED ISSUE #1500: [FEAT] S3 bucket logging
## Service

S3

## API Action / Feature

S3 Bucket Logging — `PutBucketLogging` (and the read side `GetBucketLogging`). These are the `PUT`/`GET` `/{bucket}?logging` subresource operations.

Just mocking the calls so that there is basic CRUD for the config even if no actual logging is done would be very helpful.

## AWS Documentation

- PutBucketLogging: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLogging.html
- GetBucketLogging: https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLogging.html

## Why is this needed?

This is a gap when migrating from  Localstack.

- [ ] Yes
- [X] No

codingkiddo:
## Summary

Adds basic S3 bucket logging configuration support.

This implements mock-style support for:

* `PutBucketLogging`
* `GetBucketLogging`
* storing logging configuration per bucket
* returning an empty `BucketLoggingStatus` when logging is not configured
* disabling logging by sending an empty `BucketLoggingStatus`

This does not implement actual access-log delivery.

Closes #1500

## Type of change

* [ ] Bug fix (`fix:`)
* [x] New feature (`feat:`)
* [ ] Docs / chore

## AWS Compatibility

Implemented basic S3 bucket logging configuration behavior:

* `GET /{bucket}?logging` returns the stored bucket logging configuration.
* If no logging configuration exists, it returns an empty `BucketLoggingStatus`.
* `PUT /{bucket}?logging` stores the provided `LoggingEnabled` configuration.
* `PUT /{bucket}?logging` with an empty `BucketLoggingStatus` disables logging.

Actual server access-log delivery is intentionally out of scope for this initial implementation.

## Testing

* [x] `./mvnw test -Dtest=S3IntegrationTest#bucketLogging_roundTripAndDisable`
* [x] `./mvnw test -Dtest=S3IntegrationTest`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -179,6 +179,10 @@ public class S3Controller {
             if (hasQueryParam(uriInfo, "website")) {
                 return handlePutBucketWebsite(bucket, body);
             }
+            if (hasQueryParam(uriInfo, "logging")) {
+                s3Service.putBucketLogging(bucket, new String(body, StandardCharsets.UTF_8));
+                return Response.ok().build();
+            }
             if (hasQueryParam(uriInfo, "policy")) {
                 s3Service.putBucketPolicy(bucket, new String(body, StandardCharsets.UTF_8));
                 return Response.ok().build();
@@ -329,6 +333,11 @@ public class S3Controller {
             }
             if (hasQueryParam(uriInfo, "website")) {
                 return handleGetBucketWebsite(bucket);
+            }
+            if (hasQueryParam(uriInfo, "logging")) {
+                return Response.ok(s3Service.getBucketLogging(bucket))
+                        .type("application/xml")
+                        .build();
             }
             if (hasQueryParam(uriInfo, "policy")) {
                 return Response.ok(s3Service.getBucketPolicy(bucket)).build();
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
@@ -206,6 +206,39 @@ public class S3Service {
         return bucketStore.scan(key -> true);
     }
 
+    public void putBucketLogging(String bucketName, String loggingConfigurationXml) {
+        Bucket bucket = bucketStore.get(bucketName)
+                .orElseThrow(() -> new AwsException("NoSuchBucket", "The specified bucket does not exist.", 404));
+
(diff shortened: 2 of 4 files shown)


hectorvent:
Thanks for adding this @codingkiddo, the round trip works and the wire shape is correct (verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/s3/2006-03-01/service-2.json): routing, the `BucketLoggingStatus` root, the namespace, and `NoSuchBucket` on a missing bucket all match). One thing to align with the project conventions before merge.

The repo's XML rules ask us to avoid raw string matching and to reuse the shared helpers (`XmlParser`, `XmlBuilder`, `AwsNamespaces`). Two small changes cover it:

**1. Detect enable vs disable with `XmlParser` instead of `contains`.** A `<LoggingEnabled>` carrying attributes or a namespace prefix would slip past `contains("<LoggingEnabled>")` and get stored as disabled:

```java
String targetBucket = XmlParser.extractFirst(loggingConfigurationXml, "TargetBucket", null);
if (targetBucket == null) {
    bucket.setLoggingConfiguration(null); // empty BucketLoggingStatus disables logging
} else {
    bucket.setLoggingConfiguration(loggingConfigurationXml);
}
```

**2. Build the empty response with `XmlBuilder` and the `AwsNamespaces.S3` constant** rather than a hardcoded literal:

```java
return new XmlBuilder()
        .raw("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
        .start("BucketLoggingStatus", AwsNamespaces.S3)
        .end("BucketLoggingStatus")
        .build();
```

Optional and fine to leave for later: AWS validates that `TargetBucket` exists before accepting the config (`InvalidTargetBucketForLogging`). Since this PR scopes out log delivery, accepting any target is reasonable for a first pass, just noting it.

Nice first cut on this, thanks again.

codingkiddo:
Thanks @hectorvent, that makes sense.

I’ve updated the implementation to use `XmlParser.extractFirst(...)` for detecting enabled logging instead of raw string matching, and changed the empty `GetBucketLogging` response to use `XmlBuilder` with `AwsNamespaces.S3`.

I also adjusted the regression test to cover a `LoggingEnabled` element with attributes so this behavior is protected.

hectorvent (approved):
Thanks @codingkiddo

codingkiddo:
> Thanks @codingkiddo


Thanks, @hectorvent ! Glad to contribute.

PR #1522, opened 2026-06-23
fix(elbv2): omit terminal SSL policy marker

Abhi011999:
## Summary

Fixes `DescribeSSLPolicies` pagination behavior by omitting the terminal `NextMarker` element when there are no more SSL policies to return. Previously Floci returned an empty `<NextMarker></NextMarker>`, which caused clients that check marker presence to continue paginating indefinitely.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Bug fix: `DescribeSSLPolicies` incorrectly returned an empty terminal `NextMarker`. AWS-compatible behavior is to omit `NextMarker` when the result is complete.

Verified with:
- Floci ELBv2 integration test for `DescribeSSLPolicies`
- Java AWS SDK compatibility test asserting `nextMarker()` is `null`

## Checklist


Abhi011999:
@hectorvent 
There is bug when paginating for ELBv2...

hectorvent (approved):
Thanks @Abhi011999, 
Solid fix

hectorvent (approved):
Thanks @Abhi011999

Issue #1137, opened 2026-06-02
[FEAT] EKS create-nodegroup support


ismail-bradai:
Description:
aws eks create-nodegroup returns 400 InvalidArgument.
The request is being routed to S3 instead of EKS.
This makes it impossible to create managed node groups.

Expected: create-nodegroup creates worker nodes
Actual: 400 error — POST requires ?uploads parameter
Version: 1.5.21

PR #1523, opened 2026-06-23
Nodegroups eks fargate eks

IoannisLafiotis:
## Summary

- [ ] New feature (`feat:`)
Branch: nodegroupsEks_FargateEks

Commits:
- c50f3881 Testintg on node groups and fargate profiles
  - Added service-level EKS tests for nodegroup support.
  - Added service-level EKS tests for Fargate profile support.
  - Covered create, describe, list, delete, duplicate, missing cluster, missing name, and not-found behavior.

- 7b27f4ec Add EKS SDK nodegroup and Fargate tests
  - Added AWS SDK compatibility lifecycle tests for EKS nodegroups.
  - Added AWS SDK compatibility lifecycle tests for EKS Fargate profiles.
  - Added cleanup for partially created SDK test resources.
  - Removed Mockito from EksServiceTest setup to avoid JDK 25 inline mock attachment failures.

Verification:
- JAVA_HOME=/Users/ioannislafiotis/.sdkman/candidates/java/25.0.1-amzn ./mvnw -Dtest=EksServiceTest test passed.
- ../../mvnw -DskipTests test-compile passed in compatibility-tests/sdk-test-java.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/EksTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/EksTest.java
@@ -16,6 +16,8 @@ class EksTest {
     private static EksClient eks;
     private static String clusterName;
     private static String clusterArn;
+    private static String nodegroupName;
+    private static String fargateProfileName;
 
     @BeforeAll
     static void setup() {
@@ -26,6 +28,24 @@ class EksTest {
     @AfterAll
     static void cleanup() {
         if (eks != null) {
+            if (clusterName != null) {
+                if (nodegroupName != null) {
+                    try {
+                        eks.deleteNodegroup(DeleteNodegroupRequest.builder()
+                                .clusterName(clusterName)
+                                .nodegroupName(nodegroupName)
+                                .build());
+                    } catch (Exception ignored) {}
+                }
+                if (fargateProfileName != null) {
+                    try {
+                        eks.deleteFargateProfile(DeleteFargateProfileRequest.builder()
+                                .clusterName(clusterName)
+                                .fargateProfileName(fargateProfileName)
+                                .build());
+                    } catch (Exception ignored) {}
+                }
+            }
             try {
                 eks.deleteCluster(DeleteClusterRequest.builder()
                         .name(clusterName)
@@ -145,7 +165,7 @@ class EksTest {
 
     @Test
     @Order(10)
-    void createNodegroup() {
+    void createNodeGroup() {
         CreateNodegroupResponse response = eks.createNodegroup(CreateNodegroupRequest.builder()
                 .clusterName(clusterName)
                 .nodegroupName(NODEGROUP)
@@ -159,14 +179,13 @@ class EksTest {
         assertThat(response.nodegroup().clusterName()).isEqualTo(clusterName);
         assertThat(response.nodegroup().nodegroupArn())
                 .contains("nodegroup/" + clusterName + "/" + NODEGROUP);
-        assertThat(response.nodegroup().status())
-                .isIn(NodegroupStatus.CREATING, NodegroupStatus.ACTIVE);
+        assertThat(response.nodegroup().status()).isEqualTo(NodegroupStatus.ACTIVE);
         assertThat(response.nodegroup().scalingConfig().desiredSize()).isEqualTo(2);
     }
 
     @Test
     @Order(11)
-    void listNodegroups() {
+    void listNodeGroups() {
         ListNodegroupsResponse response = eks.listNodegroups(ListNodegroupsRequest.builder()
                 .clusterName(clusterName).build());
         assertThat(response.nodegroups()).contains(NODEGROUP);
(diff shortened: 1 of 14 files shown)


hectorvent:
Thanks for putting this together, the node group and Fargate profile support is a great addition and the wire shapes look spot on. I checked the paths, response wrappers, field casing, the `FargateProfileStatus` enum, and the error codes against [`service-2.json`](https://github.com/aws/aws-sdk-java-v2/blob/master/services/eks/src/main/resources/codegen-resources/service-2.json) and they all match AWS. Nice work.

A couple of things worth a look before this is ready.

**Node groups and Fargate profiles stay in `CREATING` forever.** In [EksService.java](https://github.com/floci-io/floci/blob/95addbefc5b340a8d945137df7ec2d9ec58ec893/src/main/java/io/github/hectorvent/floci/services/eks/EksService.java) create sets the status to `CREATING`, but the readiness poller only advances clusters, so describe always returns `CREATING` and never `ACTIVE`. The previous code set node groups to `ACTIVE` right away, and moto does the same for both resources, so this is a small regression. The practical impact is that SDK waiters like `NodegroupActive` and tools like eksctl or terraform will hang waiting for a state that never arrives. Either set `ACTIVE` immediately in mock mode like before, or teach the poller to advance these too.

**The diff got reindented from 4 spaces to 2.** Most of the line count in [EksController.java](https://github.com/floci-io/floci/blob/95addbefc5b340a8d945137df7ec2d9ec58ec893/src/main/java/io/github/hectorvent/floci/services/eks/EksController.java) and [EksService.java](https://github.com/floci-io/floci/blob/95addbefc5b340a8d945137df7ec2d9ec58ec893/src/main/java/io/github/hectorvent/floci/services/eks/EksService.java) is whitespace, which makes the real changes hard to spot and drops the comment explaining why the node group routes are declared explicitly (they have to outrank the S3 catch all, issue #1137). Reverting the formatting so only the new endpoints show would help a lot.

Smaller notes:
- The new `createNodeGroup` path in [EksService.java](https://github.com/floci-io/floci/blob/95addbefc5b340a8d945137df7ec2d9ec58ec893/src/main/java/io/github/hectorvent/floci/services/eks/EksService.java) no longer validates `nodeRole` and `subnets`, which AWS marks as required. The old code rejected those.
- There are duplicate method pairs now (`createNodegroup` and `createNodeGroup`, etc.), probably worth collapsing to one.

Also the CI is red on a few checks, could you take a look? `Build and Test`, the native image build, `sdk-test-go`, and the Conventional Commits check are all failing. The commit titles like `Testintg on node groups` will need the `feat:` prefix to pass that last one.

Example of the stuck status:

```
aws eks create-nodegroup ...   # status: CREATING
aws eks describe-nodegroup ... # status: CREATING  (never becomes ACTIVE)
```

Thanks again, this is close and the protocol side is solid.

hectorvent:
@IoannisLafiotis check greptile comments.

hectorvent:
Nice work on the Fargate profile side. I checked the shapes against botocore's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/eks/2017-11-01/service-2.json) and `FargateProfile`, `FargateProfileStatus` and the `fargateProfileNames` list response all match. Adding the cluster existence check to the describe and list paths is a real fix too.

One thing to sort before merge. `CreateNodegroup` now drops `remoteAccess`, `taints`, `launchTemplate` and `nodeRepairConfig`. The controller used to bind straight into `Nodegroup` and the service mutated that object in place, so whatever the client sent came back. Now `CreateNodeGroupRequest` only declares 14 of the 18 input members and [`EksService.createNodeGroup`](https://github.com/floci-io/floci/blob/350c50e3faf5e1e0cf235858a0e7dd1785ab24b9/src/main/java/io/github/hectorvent/floci/services/eks/EksService.java#L206) builds a fresh `Nodegroup` from an explicit field list, so those four disappear from the create response and from every later `DescribeNodegroup`.

```json
// sent
{"nodegroupName":"ng1","nodeRole":"...","subnets":["subnet-abc"],
 "taints":[{"key":"dedicated","value":"gpu","effect":"NO_SCHEDULE"}]}

// returned
{"nodegroup":{"nodegroupName":"ng1","nodeRole":"...","subnets":["subnet-abc"]}}
```

All four are on the `Nodegroup` output shape, and Terraform reads every one of them back on refresh (`remote_access`, `taint`, `launch_template`, `node_repair_config` in [node_group.go](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/eks/node_group.go#L492-L536)), so an `aws_eks_node_group` with a taint or a launch template will show permanent drift. Adding the four properties to the request DTO and copying them alongside the rest should cover it.

Two smaller notes, neither blocking:

`defaultScalingConfig()` moved from `min=1, max=2, desired=2` to `1/1/1`. botocore doesn't document a default, but moto uses `2/2/2` and the AWS docs say `maxSize` defaults to 2, so the new value matches neither and nothing in the PR explains the change.

The Mockito removal in [`EksServiceTest`](https://github.com/floci-io/floci/blob/350c50e3faf5e1e0cf235858a0e7dd1785ab24b9/src/test/java/io/github/hectorvent/floci/services/eks/EksServiceTest.java) probably isn't needed. 95 other test files use Mockito and they pass on the project JDK 25.0.3, so the inline mock failure looks like a local `25.0.1-amzn` toolchain thing rather than a repo one. Keeping Mockito would shrink the diff.

hectorvent (approved):
Thanks @IoannisLafiotis, 
Solid implementation

PR #1527, opened 2026-06-23
fix(lambda): propagate SQS FIFO system attributes to ESM event

### LINKED ISSUE #1526: [BUG] SQS event source mapping omits FIFO system attributes (MessageGroupId, SequenceNumber, MessageDeduplicationId) from the Lambda event
## Summary

When a Lambda **SQS event source mapping (ESM)** invokes a function with messages from a **FIFO** queue, the SQS-specific **system attributes** — `MessageGroupId`, `SequenceNumber`, and `MessageDeduplicationId` — are **missing** from each record's `attributes` object in the Lambda event.

The values exist in floci's SQS service (a direct `ReceiveMessage` with `--attribute-names All` returns them correctly), but they are not propagated into the event payload that the ESM delivers to the function.

> Note: this is about the system `attributes` map, not the user-defined `messageAttributes` map.

## Expected behaviour

For a FIFO queue, each record in the SQS event delivered to Lambda should include the FIFO system attributes, e.g.:

```json
"attributes": {
  "ApproximateReceiveCount": "1",
  "SentTimestamp": "...",
  "SenderId": "...",
  "ApproximateFirstReceiveTimestamp": "...",
  "SequenceNumber": "...",
  "MessageGroupId": "groupA",
  "MessageDeduplicationId": "..."
}
```

`MessageGroupId` in particular is required for any consumer that processes FIFO messages in group order.

## Actual behaviour

The event delivered to the function contains only the standard attributes; the FIFO ones are absent:

```json
"attributes": {
  "ApproximateReceiveCount": "1",
  "SentTimestamp": "1782246969194",
  "SenderId": "000000000000",
  "ApproximateFirstReceiveTimestamp": "1782246969910"
}
```

## Environment

- Image: `floci/floci:latest`
- Docker socket mounted so the ESM can invoke the function.

## Reproduction

`docker-compose.yml`:

```yaml
services:
  floci:
    image: floci/floci:latest
    ports:
      - "4566:4566"
    volumes:
    environment:
      FLOCI_HOSTNAME: floci
      FLOCI_BASE_URL: http://floci:4566
      FLOCI_SERVICES_DOCKER_NETWORK: floci_default
networks:
  default:
    name: floci_default
```

```bash
docker compose up -d

# Node function that dumps each record's attributes
mkdir -p fn && cat > fn/index.js <<'EOF'
  for (const r of (event.Records || [])) {
    console.log("RECORD-ATTRS " + JSON.stringify(r.attributes));
  }
  return {};
};
EOF
(cd fn && zip -q ../dump.zip index.js)

aws() { command aws --endpoint-url http://localhost:4566 "$@"; }

aws iam create-role --role-name repro-role \
  --assume-role-policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Service":"lambda.amazonaws.com"},"Action":"sts:AssumeRole"}]}'

FIFO=$(aws sqs create-queue --queue-name attr-fifo.fifo \
  --attributes '{"FifoQueue":"true","ContentBasedDeduplication":"true"}' --query QueueUrl --output text)
FIFO_ARN=$(aws sqs get-queue-attributes --queue-url "$FIFO" --attribute-names QueueArn --query 'Attributes.QueueArn' --output text)

aws lambda create-function --function-name dumpfn --runtime nodejs20.x \
  --role arn:aws:iam::000000000000:role/repro-role --handler index.handler \
  --zip-file fileb://dump.zip --timeout 10
sleep 3
aws lambda create-event-source-mapping --function-name dumpfn --event-source-arn "$FIFO_ARN" --batch-size 1
sleep 2

aws sqs send-message --queue-url "$FIFO" --message-body "hello-fifo" --message-group-id "groupA"
sleep 8

echo "=== attributes the Lambda received (FIFO fields missing) ==="
docker compose logs floci | grep "RECORD-ATTRS" | tail -1

echo "=== for comparison, a direct ReceiveMessage DOES return them ==="
aws sqs send-message --queue-url "$FIFO" --message-body "hello2" --message-group-id "groupB"
aws sqs receive-message --queue-url "$FIFO" --attribute-names All --query 'Messages[0].Attributes'
```

### Observed output

Lambda event (FIFO attributes absent):

```
RECORD-ATTRS {"ApproximateReceiveCount":"1","SentTimestamp":"1782246969194","SenderId":"000000000000","ApproximateFirstReceiveTimestamp":"1782246969910"}
```

Direct `ReceiveMessage` on the same queue (FIFO attributes present):

```json
{
  "SenderId": "000000000000",
  "SentTimestamp": "1782246977527",
  "ApproximateReceiveCount": "1",
  "ApproximateFirstReceiveTimestamp": "1782246977804",
  "MessageGroupId": "groupB",
  "SequenceNumber": "27",
  "MessageDeduplicationId": "6e809cbda0732ac4845916a59016f954"
}
```

## Impact

Consumers that rely on `MessageGroupId` to process FIFO messages in group order cannot do so when triggered via an event source mapping, because the attribute never reaches the handler. On real AWS the SQS→Lambda event includes these FIFO system attributes.

## Suggested fix

When the ESM builds the SQS event for an invocation, copy the message's system attributes through to each record's `attributes` map — including the FIFO fields `MessageGroupId`, `SequenceNumber`, and `MessageDeduplicationId` — using the same values the SQS service already returns from `ReceiveMessage`.

slang25:
## Summary

Closes #1526. When an SQS event source mapping invokes a Lambda from a FIFO queue, `SqsEventSourcePoller.buildSqsEvent()` now copies the FIFO system attributes (`MessageGroupId`, `SequenceNumber`, `MessageDeduplicationId`) into each record's `attributes` map, matching the values `ReceiveMessage` already returns. Standard-queue messages are unaffected since each attribute is only added when present.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously the ESM event only carried the four standard system attributes, so consumers relying on `MessageGroupId` to process FIFO messages in group order never received it. On real AWS the SQS→Lambda event includes these FIFO system attributes; the fix restores that behaviour.

## Checklist

- [x] `./mvnw test` passes locally (SqsEventSourcePollerTest, 7 tests)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/SqsEventSourcePoller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/SqsEventSourcePoller.java
@@ -285,6 +285,15 @@ public class SqsEventSourcePoller {
                 attrs.put("SentTimestamp", String.valueOf(msg.getSentTimestamp().toEpochMilli()));
                 attrs.put("SenderId", AwsArnUtils.accountOrDefault(esm.getEventSourceArn(), "000000000000"));
                 attrs.put("ApproximateFirstReceiveTimestamp", String.valueOf(System.currentTimeMillis()));
+                if (msg.getSequenceNumber() > 0) {
+                    attrs.put("SequenceNumber", String.valueOf(msg.getSequenceNumber()));
+                }
+                if (msg.getMessageGroupId() != null) {
+                    attrs.put("MessageGroupId", msg.getMessageGroupId());
+                }
+                if (msg.getMessageDeduplicationId() != null) {
+                    attrs.put("MessageDeduplicationId", msg.getMessageDeduplicationId());
+                }
                 // Populate messageAttributes from the message model
                 ObjectNode msgAttrs = record.putObject("messageAttributes");
                 if (msg.getMessageAttributes() != null) {
--- a/src/test/java/io/github/hectorvent/floci/services/lambda/SqsEventSourcePollerTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/lambda/SqsEventSourcePollerTest.java
@@ -206,6 +206,46 @@ class SqsEventSourcePollerTest {
     }
 
     @Test
+    void buildSqsEventIncludesFifoSystemAttributes() throws Exception {
+        Message msg = new Message();
+        msg.setBody("hello-fifo");
+        msg.setSentTimestamp(Instant.parse("2026-01-15T10:30:00Z"));
+        msg.setMessageGroupId("groupA");
+        msg.setSequenceNumber(27);
+        msg.setMessageDeduplicationId("6e809cbda0732ac4845916a59016f954");
+
+        EventSourceMapping esm = new EventSourceMapping();
+        esm.setEventSourceArn("arn:aws:sqs:us-east-1:123456789012:my-queue.fifo");
+        esm.setRegion("us-east-1");
+
+        String event = poller.buildSqsEvent(List.of(msg), esm);
+        JsonNode attrs = OBJECT_MAPPER.readTree(event).get("Records").get(0).get("attributes");
+
+        assertEquals("groupA", attrs.get("MessageGroupId").asText());
+        assertEquals("27", attrs.get("SequenceNumber").asText());
+        assertEquals("6e809cbda0732ac4845916a59016f954",
+                attrs.get("MessageDeduplicationId").asText());
+    }
+
+    @Test
+    void buildSqsEventOmitsFifoSystemAttributesForStandardQueueMessages() throws Exception {
+        Message msg = new Message();
+        msg.setBody("hello");
+        msg.setSentTimestamp(Instant.parse("2026-01-15T10:30:00Z"));
+
+        EventSourceMapping esm = new EventSourceMapping();
+        esm.setEventSourceArn("arn:aws:sqs:us-east-1:123456789012:my-queue");
+        esm.setRegion("us-east-1");
+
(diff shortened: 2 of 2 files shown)


hectorvent:
Thanks @slang25, this looks correct. The three FIFO attributes (`MessageGroupId`, `SequenceNumber`, `MessageDeduplicationId`) match what AWS puts in the SQS to Lambda event, the PascalCase names are right, and only adding them when present keeps standard queue messages clean. Nice that it now lines up with what `ReceiveMessage` already returns, and the tests cover both the FIFO and standard cases.

Two things before merge:

1. The PR is currently showing conflicts, could you rebase on main.
2. Heads up that this is one of three open PRs touching the same `buildSqsEvent()` attributes block (#1422 for messageAttributes and #1548 for firstReceiveTimestamp), so they will conflict with each other. Once you rebase it should be fine, just flagging so the merge order is coordinated.

Thanks again!

slang25:
Thanks @hectorvent! Rebased on main — #1422 (messageAttributes) had already landed, so I resolved the overlap in `buildSqsEvent()` by keeping both: the FIFO system attributes go into the record's `attributes` map and the messageAttributes population stays untouched. They're independent additions, so no behavioural interaction. All 11 tests in `SqsEventSourcePollerTest` pass locally and the PR now shows mergeable.

On the coordination point: this and #1548 (firstReceiveTimestamp) still both touch the same block, so whichever merges second will need a trivial rebase — happy to take that one if mine goes first.

hectorvent (approved):
Thanks!

PR #1529, opened 2026-06-23
fix(ssm): retain local service diagnostics on command failure

jvanzyl:
## Summary
- wrap local SSM command execution with failure-only diagnostics
- include compact listening port, process, and log-tail evidence when a local command fails
- redact common authorization, token, password, and secret values from diagnostic output

## Tests
- ./mvnw test -Dtest=SsmDirectCommandExecutorTest

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ssm/SsmDirectCommandExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ssm/SsmDirectCommandExecutor.java
@@ -140,22 +140,106 @@ public class SsmDirectCommandExecutor {
         String standardOutput = stdout.toString(StandardCharsets.UTF_8);
         String standardError = stderr.toString(StandardCharsets.UTF_8);
         if (isTimeoutExitCode(responseCode)) {
+            logFailureDiagnostics(containerId);
             return ExecutionResult.timedOut(standardOutput, standardError, start);
         }
         if (responseCode == 0) {
             return ExecutionResult.success(standardOutput, standardError, responseCode, start);
         }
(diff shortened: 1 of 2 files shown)


hectorvent:
Nice debugging aid, and the exit code plus timeout handling look correct. Two things before merge.

The diagnostics are written to stderr, which flows into `StandardErrorContent` via `services/ssm/SsmCommandService.java:494`. AWS documents that field as exactly what the script wrote ("The first 8,000 characters written by the plugin to stderr", verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/ssm/2014-11-06/service-2.json), `GetCommandInvocation`), so a failed invocation now returns lines AWS never emits:

```
[floci:ssm] command failed; compact service diagnostics
[floci:ssm] listening ports
```

Could we gate this behind a `floci.*` flag (off by default) or send it to the Floci server log instead, so the command output stays AWS faithful?

Also the process grep hardcodes specific app names (`airship`, `iceguard`, `trino`, `coordinator`) into shared Floci source. Could we drop those or make the pattern configurable? Thanks!

jvanzyl:
Addressed the review feedback in the latest push. The direct SSM command path no longer injects Floci diagnostics into the command stderr/stdout returned by GetCommandInvocation; diagnostics are collected with a separate Docker exec after failure/timeout and written to the Floci server log. I also removed the project-specific process-name grep and replaced it with a generic, redacted process listing.\n\nFocused verification: `./mvnw -q -Dtest=SsmDirectCommandExecutorTest test`

hectorvent (approved):
Thanks @jvanzyl

PR #1530, opened 2026-06-24
feat(core): add opt-in Private Network Access for CORS preflights

Jongsic:
## Summary

Adds an opt-in `FLOCI_SECURITY_CORS_ALLOW_PRIVATE_NETWORK` flag (default `false`). When enabled, `GlobalCorsFilter` answers a browser's Private Network Access preflight with `Access-Control-Allow-Private-Network: true` — but only after the request's `Origin` already passes the existing CORS allow-list.

This unblocks a real case: a UI served from a public/secure origin (e.g. an HTTPS-hosted console) calling a Floci on `localhost`. Chromium gates that public→loopback request behind a [Private Network Access](https://developer.chrome.com/blog/private-network-access-preflight/) preflight, and without the response header the browser blocks the follow-up request even when CORS is otherwise correct.

**Motivation / parity:** Floci is used as a LocalStack drop-in, and LocalStack already returns `Access-Control-Allow-Private-Network: true` for this preflight — so a browser console that works against LocalStack on localhost currently breaks when pointed at Floci. This change closes that gap, while keeping it default-off so existing setups see no behavior change and no new exposure.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Not an AWS wire-protocol action — this is a browser/CORS concern on the global filter, so there is no SDK action to verify. The trigger and response follow the Private Network Access spec (`Access-Control-Request-Private-Network` → `Access-Control-Allow-Private-Network`). Verified by probing the preflight:

```
curl -i -X OPTIONS http://localhost:4566/ \
  -H 'Origin: https://example.test' \
  -H 'Access-Control-Request-Method: GET' \
  -H 'Access-Control-Request-Private-Network: true'
```

With the flag on (and the origin allow-listed) the response carries `Access-Control-Allow-Private-Network: true`; with it off the header is absent. This matches the header LocalStack returns for the same probe.

## Checklist


hectorvent (approved):
Thanks @Jongsic

PR #1533, opened 2026-06-24
fix(cloudfront): honor configured domain-suffix for generated distribution domains

sai-gillingham:
## Summary

`FLOCI_SERVICES_CLOUDFRONT_DOMAIN_SUFFIX` (config: `floci.services.cloudfront.domain-suffix`) was defined and documented but never actually consumed — `CloudFrontService` hardcoded the `.cloudfront.net` suffix when generating distribution domain names, so the env var had no effect.

This wires the existing config value through:

- Capture `config.services().cloudfront().domainSuffix()` into a `domainSuffix` field in the `CloudFrontService` constructor (which already receives `EmulatorConfig`).
- Use it in `createDistribution()` and `createStreamingDistribution()` in place of the hardcoded literal: `id + "." + domainSuffix`.

Default behaviour is unchanged — `domainSuffix()` defaults to `cloudfront.net` via `@WithDefault`, so existing setups still generate `<id>.cloudfront.net`. Users who set the env var (e.g. `cloudfront.local` to route distribution domains to a local emulator host) now get it honoured.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behaviour (bug fix):** The `domain-suffix` config property existed (`EmulatorConfig.CloudFrontServiceConfig.domainSuffix()`, default `cloudfront.net`) and was documented in `docs/services/cloudfront.md`, but `CloudFrontService` ignored it and always emitted `<id>.cloudfront.net` from both `createDistribution` and `createStreamingDistribution`. Setting `FLOCI_SERVICES_CLOUDFRONT_DOMAIN_SUFFIX` had no observable effect.

The default value remains `cloudfront.net`, matching real AWS CloudFront distribution domain names, so the AWS-compatible default is preserved; the suffix is only overridden when a user explicitly configures it (a local-emulation convenience, not an AWS wire-protocol change).

## Checklist

- [x] `./mvnw test` passes locally — ran the new test class: `./mvnw test -Dtest=CloudFrontServiceTest` → `Tests run: 3, Failures: 0, Errors: 0` (full suite delegated to CI)
- [x] New or updated integration test added — `CloudFrontServiceTest` with 3 cases: default suffix (`.cloudfront.net`), configured suffix for a distribution (`.cloudfront.local`), and configured suffix for a streaming distribution

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontService.java
@@ -57,6 +57,7 @@ public class CloudFrontService {
     private final StorageBackend<String, FieldLevelEncryptionProfile> fleProfileStore;
     private final StorageBackend<String, MonitoringSubscription> monitoringStore;
     private final String accountId;
+    private final String domainSuffix;
 
     @Inject
     public CloudFrontService(StorageFactory factory, EmulatorConfig config) {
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thanks @sai-gillingham

PR #1538, opened 2026-06-24
fix: implement read APIs required by Steampipe resource collection

Abhi011999:
## Summary
Implements three read-only AWS APIs that Floci previously rejected with `UnsupportedOperation` (HTTP 400) because they fell through to each service handler's `default` branch:
- IAM `ListMFADevices`
- ECR `BatchGetRepositoryScanningConfiguration`
- CloudWatch Logs `GetDataProtectionPolicy`

Each now returns a minimal, wire-accurate response matching the AWS API reference, so read-only clients (SDKs, IaC tools, resource collectors) that call these actions during read-back no longer fail.

## Type of change
- [x] Bug fix (`fix:`)

## AWS behavior implemented
- **IAM `ListMFADevices`** → empty `MFADevices` list, `IsTruncated=false`. *(Real AWS returns `NoSuchEntity` / HTTP 404 for an unknown user; Floci returns empty regardless, since MFA state isn't modeled.)*
- **ECR `BatchGetRepositoryScanningConfiguration`** → one `scanningConfigurations` entry per repo (`repositoryArn`, `repositoryName`, `scanOnPush`, `appliedScanFilters: []`) with `scanFrequency` derived from `scanOnPush` — `SCAN_ON_PUSH` when true, else `MANUAL` (both valid members of the AWS enum `SCAN_ON_PUSH | CONTINUOUS_SCAN | MANUAL`); empty `failures`. *(Deviation: real AWS reports unknown repos in the `failures[]` array; this reuses `describeRepositories`, which throws `RepositoryNotFoundException` and fails the whole batch.)*
- **CloudWatch Logs `GetDataProtectionPolicy`** → HTTP 200 with `logGroupIdentifier`, no `policyDocument`. *(Real AWS returns `ResourceNotFoundException` / HTTP 400 when the resource doesn't exist; 200-empty is a deliberate simplification for read-only callers.)*

## Checklist
- [ ] `./mvnw test` passes locally (focused tests pass; full suite optional)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsHandler.java
@@ -52,6 +52,7 @@ public class CloudWatchLogsHandler {
             case "PutSubscriptionFilter" -> handlePutSubscriptionFilter(request, region);
             case "DescribeSubscriptionFilters" -> handleDescribeSubscriptionFilters(request, region);
             case "DeleteSubscriptionFilter" -> handleDeleteSubscriptionFilter(request, region);
+            case "GetDataProtectionPolicy" -> handleGetDataProtectionPolicy(request, region);
             default -> Response.status(400)
                     .entity(new AwsErrorResponse("UnsupportedOperation", "Operation " + action + " is not supported."))
                     .build();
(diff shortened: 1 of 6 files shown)


hectorvent (approved):
Thanks @Abhi011999 

Could you please update the commit history to remove the Co-authored-by: Claude <...> line from the commit message? We prefer to keep the attribution focused on human contributors for this project.

The rest looks good to me.

Abhi011999:
Done bro, and thanks for merging these...much appreciated

Abhi011999:
wait dont merge yet, let me correctly update the comment, to be not steampipe specific...

Abhi011999:
@hectorvent 
I have removed steampipe specifc comments and updated them generically in AWS terms...let me know if it makes sense...

hectorvent (approved):

PR #1540, opened 2026-06-24
feat(ses): publish SNS notifications to identity feedback topics

okinaka:
## Summary

`SetIdentityNotificationTopic` previously only stored the SNS topic ARN per identity and echoed it back via `GetIdentityNotificationAttributes` — no notification was ever published. This PR wires those topics into the send path: a verified identity (or its parent domain) with a Bounce/Complaint/Delivery topic now receives an Amazon SNS notification when a matching event occurs, independently of any configuration set.

- `SesEventPayload`: added `buildIdentityNotification(...)`, sharing the `mail`/`bounce`/`complaint`/`delivery` blocks with event publishing via a common `buildRoot(...)`, but emitting the legacy notification format.
- `SesEventPublisher`: added `publishIdentityNotification(...)`, independent of configuration-set event destinations.
- `SesService`: `publishSendEvents` now resolves the sending identity's topics (email-address identity taking precedence over its parent domain) and publishes alongside any configuration-set destinations, even when no configuration set is present. The sending identity is resolved from the bare email address (so a display-name `Source` yields a valid `identity/<email>` ARN), and `SendRawEmail` falls back to the MIME `From` header when `Source` is omitted.
- `docs/services/ses.md`: documented the behavior.

The legacy notification format differs from event publishing in three spec-mandated ways, all honored here: the top-level discriminator is `notificationType` (not `eventType`); the `mail` object omits the `tags` field; and the original headers (`headers`/`commonHeaders`/`headersTruncated`) are included only when the identity has headers-in-notifications enabled for that notification type (`SetIdentityHeadersInNotificationsEnabled`).

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The payload shape was verified against the AWS documentation for [Amazon SNS notification contents for Amazon SES](https://docs.aws.amazon.com/ses/latest/dg/notification-contents.html) and cross-checked against the [event publishing SNS contents](https://docs.aws.amazon.com/ses/latest/dg/event-publishing-retrieving-sns-contents.html) to isolate the three legacy-vs-event-publishing differences listed above. The integration test drives the classic SES (`2010-12-01`) Query API — `VerifyEmailIdentity`, `VerifyDomainIdentity`, `SetIdentityNotificationTopic`, `SetIdentityHeadersInNotificationsEnabled`, `SendEmail`, `SendRawEmail` — and asserts the legacy `notificationType` form is delivered, that `mail.tags` is never present, and that headers appear only once headers-in-notifications is enabled. Verified end-to-end against a running instance (SES SendEmail -> SNS topic -> SQS) for Bounce, Complaint, and Delivery, with no regression to configuration-set event publishing.

## Checklist

### DIFF
--- a/docs/services/ses.md
+++ b/docs/services/ses.md
@@ -31,7 +31,7 @@ Floci exposes the classic Amazon SES Query API used by `aws ses ...` commands an
 | `UpdateAccountSendingEnabled`       | Enable or disable account-wide sending                    |
 | `ListVerifiedEmailAddresses`        | List verified email identities                            |
 | `DeleteVerifiedEmailAddress`        | Delete a verified email identity                          |
-| `SetIdentityNotificationTopic`      | Store SNS notification topic ARNs for an identity         |
+| `SetIdentityNotificationTopic`      | Set the SNS topic for an identity's bounce/complaint/delivery notifications |
 | `GetIdentityNotificationAttributes` | Read stored notification topic settings                   |
 | `SetIdentityFeedbackForwardingEnabled`     | Toggle feedback forwarding for an identity        |
 | `SetIdentityHeadersInNotificationsEnabled` | Toggle headers-in-notifications per notification type |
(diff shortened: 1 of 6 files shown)


hectorvent (approved):
Thanks @okinaka

PR #1544, opened 2026-06-24
fix(floci-ui): report real sidecar-start failures instead of always blaming the image

### LINKED ISSUE #1543: Floci UI sidecar fails to launch under Podman, and all failures are misreported as a missing image
## Summary

The "Open Floci UI" sidecar launch fails under Podman (Docker-socket compat / docker-out-of-docker), and the error surfaced to the user **misattributes every failure to a missing image**, which makes it very hard to diagnose. Reproduced end-to-end against the real `floci/floci` and `floci/floci-ui` images running under Podman.

There are two distinct problems.

## 1. Misleading error message (code defect — highest confidence)

`FlociUiManager.ensureStarted()` wraps **every** exception as a missing-image error:

`src/main/java/io/github/hectorvent/floci/services/floci/ui/FlociUiManager.java:143-148`
```java
} catch (Exception e) {
    this.lastError = "Could not start the Floci UI from image '" + image + "': " + e.getMessage()
            + ". Ensure the image is available (docker pull " + image
            + ", or build it from the floci-ui repo).";
    LOG.errorv(e, "Failed to start floci-ui sidecar from image {0}", image);
}
```

During reproduction this exact message appeared for **three completely different root causes**, in every case with the `floci/floci-ui:latest` image already pulled and present locally:

- `java.net.BindException: Permission denied` (could not connect to the Docker socket)
- `Status 500: ... listen tcp :4500: bind: address already in use` (host port conflict)
- `Status 500: ... listen tcp4 :4577: bind: address already in use` (host port conflict)

In all three the user is told to `docker pull floci/floci-ui`, which is never the actual problem. This is almost certainly why Podman users report "it just doesn't work" — the message points them at the wrong cause.

The sidecar launch is also the **first action that actually uses the Docker socket** (S3/SQS/DynamoDB/etc. are in-process), so Floci starts up healthy and looks fine until the button is pressed — at which point a pre-existing socket/permission problem surfaces here for the first time.

### Suggested fix
Inspect the caught exception instead of assuming a missing image. Only suggest `docker pull` on a `NotFoundException` / HTTP 404; surface socket-permission and port-conflict errors as themselves (and ideally log/return the underlying cause distinctly).

## 2. SELinux + the bind-mounted Docker socket (functional blocker on SELinux-enforcing Podman hosts)

On a SELinux-enforcing Podman host the floci container cannot connect to the bind-mounted socket, even following the documented `:Z` mount:

```
ERROR [io.git.hec.flo.ser.flo.ui.FlociUiManager] Failed to start floci-ui sidecar ...
java.lang.RuntimeException: java.net.BindException: Permission denied
    at com.github.dockerjava.httpclient5.ApacheDockerHttpClientImpl.execute(...)
Caused by: java.net.BindException: Permission denied
    at java.base/sun.nio.ch.UnixDomainSockets.connect0(Native Method)
    at com.github.dockerjava.transport.UnixSocket.<init>(UnixSocket.java:45)
```

Findings:
- With `-v /run/podman/podman.sock:/var/run/docker.sock:Z`, the connect failed — even when the container ran as **root**. A plain `alpine` container with the same `:Z` mount could reach the socket in the same session, but the floci image could not.
- `:Z` (private relabel) applied to the **shared** Podman API socket churns its MCS label and does not reliably grant `connectto`. The shared socket wants `:z` (lowercase, shared) or `--security-opt label=disable`.
- It worked only after bypassing SELinux labeling entirely (`--security-opt label=disable`). After that the sidecar created, started, became `ready:true`, and `http://localhost:<port>/` served the **Floci UI** page (HTTP 200). So the feature **can** work under Podman.

### Suggested fix (docs)
`docs/configuration/docker.md` recommends `:Z` on the socket mount; for the shared Podman API socket that is the wrong relabel mode and can block access. Recommend `:z` or `--security-opt label=disable`, and note this applies to the UI feature too (the Podman section currently frames the socket/network setup only around Lambda).

## Reproduction

Environment: macOS, Podman 5.8.3 (`podman machine`, rootful, SELinux enforcing in the VM), `docker` CLI pointed at the Podman socket (`DOCKER_HOST` → podman API socket).

```bash
docker pull floci/floci:latest
docker pull floci/floci-ui:latest         # image IS present

docker network create floci-net
docker run -d --name floci \
  --network floci-net -p 4566:4566 \
  -v /run/podman/podman.sock:/var/run/docker.sock:Z \
  -e FLOCI_SERVICES_LAMBDA_DOCKER_NETWORK=floci-net \
  -e FLOCI_HOSTNAME=floci \
  floci/floci:latest

# App is healthy:
curl http://localhost:4566/_floci/health        # 200

# Press the button (what the browser does):
curl http://localhost:4566/_floci/ui
curl http://localhost:4566/_floci/ui/status
# -> {"ready":false,"url":null,
#     "error":"Could not start the Floci UI from image 'floci/floci-ui:latest':
#              java.net.BindException: Permission denied.
#              Ensure the image is available (docker pull floci/floci-ui:latest, ...)"}
```

With `--security-opt label=disable` (and a free host port), the same flow returns `{"ready":true,"url":"http://localhost:<port>/"}` and the UI loads.

## Expected vs actual

- **Expected:** pressing "Open Floci UI" launches the sidecar under Podman; on failure the error names the real cause (socket permission, port conflict, etc.).
- **Actual:** sidecar fails to launch on SELinux-enforcing Podman hosts even with the documented `:Z` mount, and **all** failures are reported as a missing image, telling users to `docker pull` an image that is already present.

slang25:
## What & why

Fixes #1543.

Pressing **Open Floci UI** under Podman (Docker-socket compat / docker-out-of-docker) reports **every** sidecar-start failure as a missing image and tells the user to `docker pull floci/floci-ui` — even when the image is already present. The real cause is hidden.

I reproduced this end-to-end against the real `floci/floci` + `floci/floci-ui` images under Podman and saw the **same** misleading message for three different root causes, all with the image present:

- `java.net.BindException: Permission denied` — the floci container couldn't reach the bind-mounted Podman socket (SELinux)
- `Status 500: ... listen tcp :4500: bind: address already in use` — host port clash
- `Status 500: ... listen tcp4 :4577: bind: address already in use` — host port clash

The sidecar launch is also the first thing that actually *uses* the Docker socket (other services are in-process), so a pre-existing socket problem surfaces here for the first time — making the wrong "pull the image" guidance especially costly.

## Changes

- **`FlociUiManager.describeStartFailure(...)`** classifies the failure instead of assuming a missing image:
  - only a genuinely unavailable image (`NotFoundException`, or docker-java's `Could not pull image: ...` wrapper) keeps the `docker pull` guidance;
  - an unreachable runtime (`SocketException`/`BindException`/`ConnectException`, `UnknownHostException`) points at the socket mount + SELinux relabeling;
  - any other daemon error is surfaced as-is (e.g. `address already in use`).
- **Podman docs** (`docs/configuration/docker.md`): use `:z` (shared relabel) rather than `:Z` on the *shared* Podman API socket — `:Z` applies a container-private SELinux label that can break access; note it applies to the UI sidecar too; mention the `--security-opt label=disable` fallback.
- **Tests**: 4 new unit tests covering the missing-image (2 forms), unreachable-socket, and port-conflict classifications.

## Testing

`./mvnw -o -Dtest=FlociUiManagerTest test` → `Tests run: 14, Failures: 0, Errors: 0`.

Manually verified under Podman: once the socket is reachable and the port is free, the sidecar launches, `/_floci/ui/status` returns `{"ready":true,...}`, and the UI page loads (HTTP 200). When it fails, the message now names the actual cause.

## Notes / out of scope

This PR fixes diagnosability (issue #1543, part 1) and the docs. The underlying SELinux socket-access behavior on some Podman hosts (part 2) is addressed via the docs change here; if maintainers want deeper handling (e.g. auto-detecting an unreachable socket at startup) that can be a follow-up.

### DIFF
--- a/docs/configuration/docker.md
+++ b/docs/configuration/docker.md
@@ -132,7 +132,7 @@ podman network create floci-net
 podman run -d --name floci \
   --network floci-net \
   -p 4566:4566 \
-  -v /run/user/$(id -u)/podman/podman.sock:/var/run/docker.sock:Z \
+  -v /run/user/$(id -u)/podman/podman.sock:/var/run/docker.sock:z \
   -e FLOCI_SERVICES_LAMBDA_DOCKER_NETWORK=floci-net \
   -e FLOCI_HOSTNAME=floci \
   floci/floci
@@ -147,9 +147,14 @@ What each setting does and why it is needed:
   Lambda containers it spawns to that same named network.
 - **`FLOCI_HOSTNAME=floci`** — gives Floci a stable name that Lambda containers
   resolve when calling back to the Runtime API.
-- **`:Z` on the socket mount** — relabels the Podman socket for SELinux. Without
-  it, Floci fails to talk to the Podman socket and Lambda container creation
-  errors with `java.io.IOException: Broken pipe`.
+- **`:z` on the socket mount** — relabels the Podman socket for SELinux. Without
+  it, Floci fails to talk to the Podman socket: Lambda/ECR container creation
+  errors with `java.io.IOException: Broken pipe`, and the **Floci UI** sidecar
+  fails to launch with `java.net.BindException: Permission denied`. Use the
+  lowercase `:z` (shared relabel) rather than `:Z` — the Podman API socket is
+  shared with the Podman service, and `:Z` applies a container-private SELinux
+  label that can break access. If `:z` is still not enough on your host, fall
+  back to `--security-opt label=disable`.
 
 !!! tip "When the Runtime API address is still unreachable"
     On some Podman network topologies the auto-detected Runtime API address
--- a/src/main/java/io/github/hectorvent/floci/services/floci/ui/FlociUiManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/floci/ui/FlociUiManager.java
@@ -21,6 +21,8 @@ import io.github.hectorvent.floci.core.common.docker.ContainerLogStreamer;
 import io.github.hectorvent.floci.core.common.docker.ContainerSpec;
 import io.github.hectorvent.floci.core.common.docker.CurrentContainerNetworkResolver;
 import io.github.hectorvent.floci.core.common.docker.DockerHostResolver;
+import com.github.dockerjava.api.exception.DockerClientException;
+import com.github.dockerjava.api.exception.NotFoundException;
 import com.github.dockerjava.api.model.Container;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
@@ -141,9 +143,7 @@ public class FlociUiManager {
             LOG.infov("Started floci-ui sidecar {0} on host port {1}", name, String.valueOf(hostPort));
             attachLogStream();
         } catch (Exception e) {
-            this.lastError = "Could not start the Floci UI from image '" + image + "': " + e.getMessage()
-                    + ". Ensure the image is available (docker pull " + image
-                    + ", or build it from the floci-ui repo).";
+            this.lastError = describeStartFailure(image, e);
             LOG.errorv(e, "Failed to start floci-ui sidecar from image {0}", image);
         }
     }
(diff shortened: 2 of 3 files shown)


hectorvent (approved):
Thanks @slang25

PR #1545, opened 2026-06-24
fix(dynamodb): invalid expressions throw 400 with exact error messages

### LINKED ISSUE #1512: [BUG] DynamoDB: invalid ConditionExpression/FilterExpression/KeyConditionExpression returns 500 instead of 400
## Service

DynamoDB

## AWS API Action

DeleteItem / PutItem / UpdateItem (`ConditionExpression`), Query (`KeyConditionExpression` / `FilterExpression`), Scan (`FilterExpression`)

## Expected behavior

For a lexically-invalid `ConditionExpression` / `FilterExpression` / `KeyConditionExpression` (e.g. a bare hyphen, `a-b`), AWS returns **HTTP 400 `ValidationException`** — `Invalid <ExprType>: Syntax error; token: "-", near: "a-b"`.

Floci itself already does the right thing on the `UpdateExpression` path, returning 400 for the same bad token, so the condition/filter/key paths should be made consistent with it.

**Confirmed against real AWS** (DynamoDB, ap-northeast-1):

```
DeleteItem  --condition-expression 'a-b'      -> ValidationException: Invalid ConditionExpression:  Syntax error; token: "-", near: "a-b"
PutItem     --condition-expression 'a-b'      -> ValidationException: Invalid ConditionExpression:  Syntax error; token: "-", near: "a-b"
Scan        --filter-expression 'a-b'         -> ValidationException: Invalid FilterExpression:     Syntax error; token: "-", near: "a-b"
Query       --key-condition-expression 'a-b'  -> ValidationException: Invalid KeyConditionExpression: Syntax error; token: "-", near: "a-b"
UpdateItem  --update-expression 'a-b'         -> ValidationException: Invalid UpdateExpression:     Syntax error; token: "a", near: "a-"
```

All five return HTTP 400 `ValidationException`. Floci matches AWS only on `UpdateExpression`; the other four return 500 (see below).

## Actual behavior

Floci returns **HTTP 500 `InternalFailure`** — an uncaught exception leaks out of the expression tokenizer:

```json
{"__type":"InternalFailure","message":"Unexpected character '-' at position 1 in expression: a-b"}   // HTTP 500
```

Because 500 `InternalFailure` is a retryable error, the AWS SDK/CLI also retries the (deterministically invalid) request — e.g. the CLI reports `reached max retries: 2`, so one bad input becomes three server calls. A 400 `ValidationException` would not be retried.

### Root cause

In `services/dynamodb/ExpressionEvaluator.java`, `validateExpression` (line ~389) calls `tokenize()` **outside** the try/catch:

```java
static void validateExpression(String expression, String exprType, ...) {
    if (expression == null || expression.isBlank()) return;
    List<Token> tokens = tokenize(expression.trim());   // <-- OUTSIDE try/catch
    checkRedundantParentheses(tokens, exprType);
    Expr expr;
    try {
        expr = parse(expression);                        // <-- only parse() is guarded
    } catch (IllegalArgumentException e) {
        return;                                           // syntax errors swallowed
    }
    validateSemantics(expr, exprType, exprAttrNames, exprAttrValues);
}
```

`tokenize()` throws `IllegalArgumentException("Unexpected character ...")` (same file, line ~151) for an invalid character. Because the call is outside the try/catch, the exception is never converted to an `AwsException(400)` and reaches the exception mapper as a 500. (Note the asymmetry: `parse()` errors are *swallowed* with a bare `return`, i.e. malformed-but-tokenizable expressions are silently accepted, while a tokenizer error crashes.)

Call sites of `validateExpression` in `DynamoDbJsonHandler.java`:

| Op | Expression | Line |
|---|---|---|
| DeleteItem | ConditionExpression | 331 |
| PutItem | ConditionExpression | 437 |
| UpdateItem | ConditionExpression | 504 |
| Query | KeyConditionExpression / FilterExpression | 757 / 758 |
| Scan | FilterExpression | 894 |

Trigger: any character the tokenizer does not recognize (a hyphen is the obvious one). Real clients usually reference such names via `ExpressionAttributeNames` placeholders (`#name`), so production exposure is limited — but returning a 500 (uncaught exception) on bad input is a robustness defect.

### Regression

`git log -L` shows the current `validateExpression` shape (with `tokenize()` before the try) was introduced in **#1442 "fix(dynamodb): reject redundant parens, contains() dup operand, begins_with non-string" (2026-06-19)**. It does not reproduce on the released 1.5.26; it is a regression on `main` after that release.

### Suggested fix (no code change in this issue)

Move `tokenize()` / `checkRedundantParentheses()` inside the guarded block and convert the tokenizer's `IllegalArgumentException` into `AwsException("ValidationException", "Invalid <exprType>: Syntax error; ...", 400)`, matching the `UpdateExpression` path. The silently-swallowed `parse()` errors are worth revisiting too (they currently let malformed expressions through).

## Reproduction

```bash
ENDPOINT=http://localhost:4566   # Floci. For real-AWS check: drop --endpoint-url

aws --endpoint-url $ENDPOINT dynamodb create-table --table-name repro \
  --attribute-definitions AttributeName=id,AttributeType=S \
  --key-schema AttributeName=id,KeyType=HASH --billing-mode PAY_PER_REQUEST

# BUG: invalid ConditionExpression -> Floci 500 InternalFailure (AWS: 400 ValidationException)
aws --endpoint-url $ENDPOINT dynamodb delete-item --table-name repro \
  --key '{"id":{"S":"a"}}' --condition-expression 'a-b'

# Consistency control: same bad token in UpdateExpression -> correctly 400
aws --endpoint-url $ENDPOINT dynamodb update-item --table-name repro \
  --key '{"id":{"S":"a"}}' --update-expression 'a-b'
```

## Environment

- Floci version / image tag: `main` `7b3c5b23` (local `mvn clean package`, reports version `dev`). Not reproducible on released `1.5.26` — regression after #1442.
- Java SDK version (if applicable): reproduces via AWS CLI and via raw `application/x-amz-json-1.0` POST.
- How you're running Floci: native (`java -jar target/quarkus-app/quarkus-run.jar`).

someear9h:
## Summary

This PR makes Floci return a 400 Validation Exception on invalid condition, filter, KeyCondition expressions, before it was returning 500 which is a retryable error which resulted into 3 api calls for one bad input

Closes #1512

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Floci returned HTTP 500 InternalFailure (instead of 400 validation exception) resulting into 3 api calls because its a retryable error

## Checklist

Local test note: All DynamoDB tests pass, full test suite run experienced unrelated local environmental timeouts and container initialization failures for ELB and cloudformation

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/ExpressionEvaluator.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/ExpressionEvaluator.java
@@ -148,8 +148,13 @@ final class ExpressionEvaluator {
                 continue;
             }
 
+            int nearStart = Math.max(0, i - 15);
+            int nearEnd = Math.min(expression.length(), i + 15);
+            
+            String near = expression.substring(nearStart, nearEnd);
+
             throw new IllegalArgumentException(
-                    "Unexpected character '%c' at position %d in expression: %s".formatted(c, i, expression));
+                    "token: \"%c\", near: \"%s\"".formatted(c, near));
         }
 
         tokens.add(new Token(TokenType.EOF, "", len));
@@ -389,14 +394,22 @@ final class ExpressionEvaluator {
     static void validateExpression(String expression, String exprType,
                                    JsonNode exprAttrNames, JsonNode exprAttrValues) {
         if (expression == null || expression.isBlank()) return;
-        List<Token> tokens = tokenize(expression.trim());
-        checkRedundantParentheses(tokens, exprType);
+        List<Token> tokens;
         Expr expr;
         try {
+            tokens = tokenize(expression.trim());
+            checkRedundantParentheses(tokens, exprType);
             expr = parse(expression);
         } catch (IllegalArgumentException e) {
-            // Other syntax errors are surfaced by the existing parse/evaluate paths.
-            return;
+            String detail = e.getMessage();
+            
+            if (detail.startsWith("token:")) {
+                throw new AwsException("ValidationException", 
+                    "Invalid " + exprType + ": Syntax error; " + detail, 400);
+            } else {
+                throw new AwsException("ValidationException", 
+                    "Invalid " + exprType + ": Syntax error", 400);
+            }
         }
         validateSemantics(expr, exprType, exprAttrNames, exprAttrValues);
     }
--- a/src/test/java/io/github/hectorvent/floci/services/dynamodb/ExpressionEvaluatorTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/dynamodb/ExpressionEvaluatorTest.java
@@ -8,6 +8,8 @@ import org.junit.jupiter.api.Test;
 
 import java.util.List;
 
+import io.github.hectorvent.floci.core.common.AwsException;
+
 import static org.junit.jupiter.api.Assertions.*;
 
 class ExpressionEvaluatorTest {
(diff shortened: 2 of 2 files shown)


hectorvent (approved):
Thanks @someear9h

PR #1548, opened 2026-06-25
fix(lambda): use persisted firstReceiveTimestamp in SQS ESM

### LINKED ISSUE #1497: [BUG] SQS→Lambda ESM resets ApproximateFirstReceiveTimestamp to current time on every poll cycle
## Service

* **Component:** Lambda (SQS Event Source Mapping poller)

## AWS API Action

* **Action:** Lambda invocation via SQS ESM
* **Field:** `Records[*].attributes.ApproximateFirstReceiveTimestamp`

## Expected Behavior

The `ApproximateFirstReceiveTimestamp` should reflect the epoch time (in milliseconds) when the message was **first** received from the queue. On retries (e.g., when the Lambda function fails and the message returns to the queue), this value must remain constant—it tracks the initial receipt, not the latest poll.

> See the [[AWS SQS Message Metadata Documentation](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html)](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html) for reference.

## Actual Behavior

The timestamp is recalculated using `System.currentTimeMillis()` on every single poll cycle. As a result, it always reflects the current wall-clock time rather than the original receipt time.

### Root Cause

The bug originates in [`SqsEventSourcePoller.java` (Line 287)](https://www.google.com/search?q=%5Bhttps://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/lambda/SqsEventSourcePoller.java%5D(https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/lambda/SqsEventSourcePoller.java)):

```java
// Current code — always sets current time, not the first-receive time
attrs.put("ApproximateFirstReceiveTimestamp", String.valueOf(System.currentTimeMillis()));

```

---

## Steps to Reproduce

1. Create an SQS queue, a Lambda function, and an Event Source Mapping (ESM).
2. Deploy a Lambda function configured to intentionally throw an error (to trigger a retry).
3. Send a message to the queue:
```bash
aws --endpoint-url http://localhost:4566 sqs send-message \
  --queue-url http://localhost:4566/000000000000/test-queue \
  --message-body '{"test":"retry"}'

```

4. Observe the Lambda invocation logs across retries:
* **First invocation:** `ApproximateFirstReceiveTimestamp = T1`
* **Second invocation:** `ApproximateFirstReceiveTimestamp = T2` *(where T2 > T1)*

**Expected Result:** Both invocations should show `T1`.

---

## Environment

* **Floci version / image tag:** `latest` (`main` branch)
* **Java SDK version:** AWS SDK v2
* **Execution Method:** `mvn quarkus:dev`

9046balaji:
Both SqsEventSourcePoller.buildSqsEvent() and PipesPoller.buildSqsRecordNodes() were using System.currentTimeMillis() for ApproximateFirstReceiveTimestamp, causing the value to change on every poll cycle. AWS SQS defines this as the epoch time when the message was first received from the queue, and it must remain constant across retries.

The SQS service layer already tracks this correctly via Message.firstReceiveTimestamp (set once in GuardedMessageQueue.tryClaim()), but both pollers bypassed the stored value.

Changes:
- SqsEventSourcePoller: read msg.getFirstReceiveTimestamp() with null fallback
- PipesPoller: same fix for the identical code path
- SqsEventSourcePollerTest: set firstReceiveTimestamp on test messages and assert the output matches the stored value; add a regression test proving the timestamp remains stable across simulated retries

Closes #1497


hectorvent:
Thanks @9046balaji, this looks good. Reading the persisted `firstReceiveTimestamp` matches AWS, the value stays constant across retries, both the Lambda ESM and Pipes pollers are covered, and the regression test locks it in.

Could you resolve the merge conflicts with main when you get a chance, then it should be ready to go. Thanks again!

9046balaji:
@hectorvent 

Thanks for checking in. I’ve fixed the issue now, everything is working fine.

hectorvent (approved):

PR #1549, opened 2026-06-25
feat(iam): route assumed-role credentials to the target account

abanna:
## Summary

Assumed-role temporary credentials (`ASIA…` keys) collapsed to the default account because account resolution only recognised 12-digit access key IDs and had no link to the STS session store. The cross-account *"assume a role in account B, then provision"* pattern (e.g. CloudFormation deploying into a target account) therefore landed resources in `000000000000` instead of B.

This routes STS temporary credentials to the correct account so that pattern works locally exactly as it does in AWS.

### What changed

- **`SessionAccountLookup` port** (`core.common`), implemented by `IamService` — maps a temporary access key to its account: the account in the session's role (or federated-user) ARN, or the caller account captured at mint time for role-less `GetSessionToken` credentials. The port keeps the dependency direction `services → core` (no cycle).
- **`AccountContextFilter`** now resolves accounts with the precedence **12-digit AKID → session lookup → default**, for both the `Authorization` header and the presigned `X-Amz-Credential` parameter.
- **STS handlers** record the caller account on every minted session; **`SessionCredential`** gains an `originAccountId` field. The session lookup is account-independent because sessions are keyed by a globally unique access key.
- **`IamService` is now `@Startup`** so AWS-managed policies seed under the default account at boot. Seeding is account-namespaced, and `AccountContextFilter` now instantiates `IamService` to resolve the request account — without eager init, lazy seeding would bind managed policies to whichever account happened to make the first request.
- Docs: a "Temporary Credentials (AssumeRole)" section in the multi-account guide + a README note.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

No wire-protocol changes. STS still returns standard `AssumeRole`/`GetSessionToken`/… responses with `ASIA…` keys; only the internal account that those credentials resolve to is corrected. Resolution precedence matches the existing documented 12-digit-AKID convention, extended to temporary sessions. Verified against the AWS SigV4 `Authorization` header and presigned `X-Amz-Credential` formats.

## Checklist

- [x] `./mvnw test` passes locally (targeted: resolver/filter/IAM/STS unit + integration suites — 135 tests green; build run in a JDK 25 container)
- [x] New or updated integration test added — `AssumedRoleAccountRoutingIntegrationTest` proves a table created with role-B credentials materialises in account B and is invisible to account A

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,8 @@ AWS_ACCESS_KEY_ID=222222222222 aws sqs create-queue --queue-name orders
 
 Any other key format, such as `test` or `AKIA...`, causes Floci to fall back to `FLOCI_DEFAULT_ACCOUNT_ID`, which defaults to `000000000000`.
 
+STS temporary credentials are routed too: credentials from `AssumeRole` resolve to the assumed role's account, so the cross-account assume-role-then-provision pattern works locally. Resolution precedence is 12-digit AKID → temporary-session lookup → `FLOCI_DEFAULT_ACCOUNT_ID`.
+
 See the [Multi-Account Isolation docs](https://floci.io/floci/configuration/multi-account/).
 
 ## SDK Integration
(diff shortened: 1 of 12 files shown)


hectorvent (approved):
Thanks @abanna

PR #1552, opened 2026-06-25
feat(iam): enforce AssumeRole trust policies when enforcement is enabled

abanna:
## Summary

When `FLOCI_SERVICES_IAM_ENFORCEMENT_ENABLED=true`, STS `AssumeRole` now evaluates the target role's trust policy (`AssumeRolePolicyDocument`) against the caller and returns `AccessDenied` if it isn't permitted. **Enforcement is off by default, so existing behavior is unchanged.**

Trust policies are principal-centric and carry no `Resource` element, so the identity/resource-oriented `IamPolicyEvaluator` can't evaluate them. This adds a focused `AssumeRolePolicyEvaluator` that:

- matches each statement's `Action` against `sts:AssumeRole` (supports `sts:*` / `*`),
- matches the AWS `Principal` — `"*"`, a bare account id, an account-root ARN (`arn:aws:iam::<acct>:root`), or an exact principal ARN,
- applies AWS precedence: an explicit `Deny` wins, otherwise a matching `Allow` grants.

The role is resolved in **its own account** (parsed from the role ARN) via a new `IamService.findRole(account, roleName)` overload, since IAM roles are account-namespaced. Roles that Floci has no record of stay permissive, so this only affects roles created through IAM with a real trust policy.

> Note: the plan that motivated this work suggested reusing `IamPolicyEvaluator`, but its statements require a `Resource` match and ignore `Principal`, so a dedicated evaluator is used instead (it reuses the shared glob matcher).

## Type of change

- [x] New feature (`feat:`) — gated behind an existing opt-in flag

## AWS Compatibility

Query protocol, unchanged wire format. Only adds an authorization decision on `AssumeRole` when enforcement is enabled. `Service` / `Federated` principals are intentionally not matched against SigV4 callers (function-before-fidelity).

## Checklist

- [x] `./mvnw test` passes locally (targeted: `AssumeRolePolicyEvaluatorTest` 12 + `AssumeRoleTrustPolicyIntegrationTest` 3 + `IamIntegrationTest` 37 green; built in a JDK 25 container)
- [x] New unit + integration tests added (permitted caller allowed, unauthorized denied, unknown role permissive; evaluator principal/action/deny/malformed coverage)

Independent of #1549 and #1551, though it composes with them for full cross-account assume-role fidelity.


hectorvent:
This is really clean, Alex, and the evaluator is nicely tested. The wire side checks out: AccessDenied at 403 is the right denial, and the action, principal, and explicit deny precedence all match how AWS evaluates a trust policy.

Two gaps worth noting, both fine to defer as documented limitations since they are about evaluating more than the trust policy itself:

1. Condition blocks are not evaluated, so `sts:ExternalId` is not enforced. A role whose trust policy requires an external id is currently assumable without one, and the `ExternalId` request param is ignored too. For example this statement matches purely on the principal today, with the condition skipped:

```json
{
  "Effect": "Allow",
  "Principal": { "AWS": "arn:aws:iam::111111111111:root" },
  "Action": "sts:AssumeRole",
  "Condition": { "StringEquals": { "sts:ExternalId": "secret-123" } }
}
```

So the caller gets in without passing `ExternalId=secret-123`. For what it is worth, moto and localstack do not evaluate this either, so you are already ahead here. Just worth a line in the docs so people know the confused deputy guard is not active yet.

2. Only the trust policy side is checked. Real cross account AssumeRole also needs the caller's identity policy to allow `sts:AssumeRole`. Reasonable to leave for later, just good to state as a known limitation.

One tiny thing: AWS prefixes the denial message with the caller, like:

```
User: arn:aws:iam::111111111111:user/alice is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::222222222222:role/foo
```

so adding the `User: ` prefix would match it exactly.

Thanks for gating this behind the opt in flag and keeping the default behavior untouched, nice and safe.

abanna:
Thanks @hectorvent! Addressed in 5f4a3e83 (also merged latest `main`, which had diverged):

- **`User:` prefix** — the AccessDenied message is now `User: <callerArn> is not authorized to perform: sts:AssumeRole on resource: <roleArn>`, matching AWS exactly. Strengthened the integration test to assert the prefix, action, and resource.
- **ExternalId / Condition blocks** — documented as a known limitation in `docs/services/sts.md`: condition blocks aren't evaluated, so `sts:ExternalId` isn't enforced and the request param is ignored (matching moto/LocalStack). Left the implementation as a follow-up per your note.
- **Identity-policy side** — also documented as a known limitation: only the trust policy is checked, not the caller's own `sts:AssumeRole` permission.

The merge against main was non-trivial — it reworked `IamService`/`StsQueryHandler` for account-aware sessions — so I resolved those and re-ran the IAM/STS suite (151 tests green) in a temurin-25 container against the updated main.

abanna:
Thanks @greptile-apps — went through all three:

- **`Configuration` section dropped from sts.md** (P2): Valid, fixed in 5ffda29c — restored the `FLOCI_SERVICES_STS_ENABLED` table.
- **`NotAction` silently skipped** (P2): Valid in principle — a `Deny` with `NotAction` won't block the assume. Trust policies using `NotAction` are very rare, and like `Condition` evaluation it's an advanced policy feature this PR intentionally defers, so I documented it as a known limitation rather than implementing partial `NotAction` semantics now.
- **`findRole` flat-backend fallback ignores accountId** (P2): In practice `roles` is always an `AccountAwareStorageBackend` (main made IAM account-namespaced), so the fallback is only a safety net. A genuinely flat backend has no account dimension to key on, so `roles.get(roleName)` is the only thing it *can* do — passing accountId wouldn't change the result. Left as-is.

hectorvent (approved):
Thanks @abanna

PR #1553, opened 2026-06-25
feat(ses): implement PutConfigurationSetDeliveryOptions for SES v1

okinaka:
## Summary

`PutConfigurationSetDeliveryOptions` was only reachable through the SES v2 REST path — the classic SES Query (v1) API returned `UnsupportedOperation`, even though the `setConfigurationSetDeliveryOptions` service logic and validation already existed. This wires the v1 action:

- `SesQueryHandler`: added `PutConfigurationSetDeliveryOptions`, parsing `ConfigurationSetName` and `DeliveryOptions.TlsPolicy` and delegating to the shared service.
- `AwsQueryController`: registered the action in the SES action set so it routes to the SES handler.
- `docs/services/ses.md`: listed the action.

The v1 `DeliveryOptions` structure carries only `TlsPolicy` (`Require`/`Optional`), whereas Floci stores the v2 canonical `REQUIRE`/`OPTIONAL`. The handler normalizes the value before delegating, so a v1 SDK client sending `Require` is accepted rather than rejected by the shared validator. An all-null payload passes `null` to clear the block, mirroring the v2 path rather than persisting an empty object.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Parameter and enum shapes were verified against the AWS API reference for [PutConfigurationSetDeliveryOptions](https://docs.aws.amazon.com/ses/latest/APIReference/API_PutConfigurationSetDeliveryOptions.html) and [DeliveryOptions](https://docs.aws.amazon.com/ses/latest/APIReference/API_DeliveryOptions.html): request params `ConfigurationSetName` and `DeliveryOptions.TlsPolicy` with valid values `Require | Optional` (the v1 structure has no `SendingPoolName`). Coverage: v1 Query integration tests (happy path, unknown configuration set, invalid TLS policy) and an AWS Java SDK v1 compatibility test that writes `Require` via the v1 `SesClient`, reads it back as the canonical `REQUIRE` via v2 `GetConfigurationSet`, asserts an empty put clears the block, and asserts the modeled `ConfigurationSetDoesNotExistException` for an unknown set.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SesConfigurationSetDeliveryOptionsV1Test.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SesConfigurationSetDeliveryOptionsV1Test.java
@@ -1,0 +1,109 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+import software.amazon.awssdk.services.ses.SesClient;
+import software.amazon.awssdk.services.ses.model.ConfigurationSet;
+import software.amazon.awssdk.services.ses.model.ConfigurationSetDoesNotExistException;
+import software.amazon.awssdk.services.ses.model.CreateConfigurationSetRequest;
+import software.amazon.awssdk.services.ses.model.DeleteConfigurationSetRequest;
+import software.amazon.awssdk.services.ses.model.DeliveryOptions;
+import software.amazon.awssdk.services.ses.model.PutConfigurationSetDeliveryOptionsRequest;
+
+import software.amazon.awssdk.services.sesv2.SesV2Client;
+import software.amazon.awssdk.services.sesv2.model.GetConfigurationSetRequest;
+import software.amazon.awssdk.services.sesv2.model.GetConfigurationSetResponse;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * SDK compatibility test for the SES V1 (Query) {@code PutConfigurationSetDeliveryOptions}
+ * action. The V1 API names the TLS policy {@code Require}/{@code Optional} (PascalCase),
+ * while Floci stores the V2 canonical {@code REQUIRE}/{@code OPTIONAL}. A value written
+ * through the V1 {@link SesClient} must therefore read back as {@code REQUIRE} through the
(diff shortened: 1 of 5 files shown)


hectorvent:
Really nice clean wiring here, and the v1 to v2 round trip test is a lovely touch.

One small thing on the error path. When `TlsPolicy` is invalid, the v1 action returns `<Code>BadRequestException</Code>`, but that's a SES v2 shape and isn't one of the errors v1 declares for this action. The v1 model only lists `ConfigurationSetDoesNotExist` and `InvalidDeliveryOptions` (both have an explicit `error.code` in [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/ses/2010-12-01/service-2.json)), so real AWS returns `InvalidDeliveryOptions` here and a v1 SDK can map it to the typed exception. With `BadRequestException` it just falls back to a generic `SesException`.

It's coming from the shared `validateDeliveryOptions` (which is correct for the v2 REST path), so the v1 handler just needs a tiny translation:

```java
if (tlsPolicy != null && !VALID.contains(tlsPolicy.toUpperCase(Locale.ROOT))) {
    throw new AwsException("InvalidDeliveryOptions",
            "Invalid delivery options: TlsPolicy must be Require or Optional", 400);
}
```

That also keeps it consistent with how the unknown-set case already throws `ConfigurationSetDoesNotExist` without the suffix. The new test assertion would just flip to `<Code>InvalidDeliveryOptions</Code>`.

Everything else matches the model cleanly, the params, the `Require`/`Optional` enum, the empty response, and the unknown-set error. Thanks for this!

okinaka:
Thanks for the careful review and the kind words!

You're right that the v2-style `BadRequestException` is wrong on the v1 surface. I probed real AWS to pin the exact code/message before fixing, and it turned out a little different from `InvalidDeliveryOptions` — for an invalid `TlsPolicy` *value*, v1 returns `ValidationError`:

```
$ aws ses put-configuration-set-delivery-options \
    --configuration-set-name <cs> --delivery-options TlsPolicy=Bogus
An error occurred (ValidationError) when calling the PutConfigurationSetDeliveryOptions operation:
1 validation error detected: Value at 'deliveryOptions.tlsPolicy' failed to satisfy constraint:
Member must satisfy enum value set: [Optional, Require]
```

The enum-value rejection happens at the protocol/Smithy validation layer (`ValidationError`) before the service logic runs, so `InvalidDeliveryOptions` looks reserved for semantic delivery-option errors (e.g. a non-existent `SendingPoolName`) rather than a bad enum value. `TlsPolicy` is also case-sensitive — lowercase `require` is rejected with the same `ValidationError`; only `Require`/`Optional` are accepted.

So I'll add the v1-handler translation you suggested, but throwing `ValidationError` with that exact message (validating `Require`/`Optional` case-sensitively before normalizing), and flip the test assertion to `<Code>ValidationError</Code>`. Thanks again!

okinaka:
@hectorvent done in `02b88a00` — now returns `ValidationError` with AWS's exact message. Thanks!

hectorvent (approved):
Thanks @okinaka

PR #1556, opened 2026-06-25
feat(stepfunctions): add aws-sdk CloudFormation/EC2 and S3 PutObject integrations

abanna:
## Summary

Step Functions could only invoke a handful of optimized/`aws-sdk` integrations (Lambda, DynamoDB, SQS, nested state machines); every other resource failed with `Unsupported resource`. That blocks a common pattern: provisioning state machines that drive **CloudFormation StackSets**, look up regions, or stage templates in S3.

This adds:
- a **generic Query-protocol bridge** for `arn:aws:states:::aws-sdk:cloudformation:*` — flattens the task input to Query parameters, dispatches to the CloudFormation handler, and converts the XML response into the JSON shape the `aws-sdk` integration returns (`<member>` lists unwrapped to arrays);
- `arn:aws:states:::aws-sdk:ec2:describeRegions`;
- `arn:aws:states:::s3:putObject` (optimized and `aws-sdk` forms).

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

`aws-sdk` service integrations follow the standard `arn:aws:states:::aws-sdk:<service>:<action>` form and return the SDK-style JSON response. The CloudFormation bridge reuses the existing query handler, so it tracks CloudFormation behavior automatically.

## Checklist

- [x] `./mvnw test` passes locally (targeted: `QueryXmlToJsonTest`, `StepFunctionsAwsSdkQueryIntegrationTest`, plus existing SFN integration suites for regression — 36 green; built in a JDK 25 container)
- [x] New unit test (XML→JSON converter) + integration test (ec2:describeRegions, cloudformation:validateTemplate, s3:putObject through Step Functions)
- [x] Commit messages follow Conventional Commits

## Note

The CloudFormation **StackSet** actions this lights up (CreateStackSet/CreateStackInstances/DescribeStackInstance/…) are implemented in #1551 — with both merged, a Step Functions workflow can drive cross-account StackSet provisioning end-to-end. Verified locally: an SFN chain `createStackSet → createStackInstances → describeStackInstance` provisions resources into a target account's namespace.

PR #1558, opened 2026-06-25
feat(stepfunctions): ResultSelector, Pass Parameters, ArrayContains, wildcard projection; fix IsPresent for absent paths

abanna:
## Summary

Three JSONPath / intrinsic features the Step Functions executor was missing — each independently useful, and required together by provisioning state machines that list results and validate an input against them:

- **`ResultSelector` is now applied** to a Task's result (before `ResultPath`), reusing the same template resolution as `Parameters`. It was previously accepted as a field but silently ignored, so the raw result flowed through unchanged.
- **`$.X.*.Y` wildcard projection** in path resolution: a `*` segment projects the remainder of the path over each element of the current array and collects the results (e.g. `$.Regions.*.RegionName`).
- **`States.ArrayContains(array, value)`** intrinsic.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

`ResultSelector`, the `*` projection, and `States.ArrayContains` all follow the AWS States Language spec. No change to existing fields/behavior — `ResultSelector` was a no-op before, and the `*` projection only affects paths that contain a `*` segment.

## Checklist

- [x] `./mvnw test` passes locally (clean run: new `AslExecutorPathIntrinsicsTest` + existing Step Functions suites — 43 green; built in a JDK 25 container)
- [x] Unit tests for wildcard projection, `States.ArrayContains`, and `ResultSelector` evaluation

These features are exercised end-to-end by a real-world CloudFormation StackSet provisioning state machine (validate template → list regions via `ResultSelector` projection → `ArrayContains` region check → create stack set/instances), composed with #1551 and #1556.


hectorvent:
Really nice set of additions, Alex. The ResultSelector ordering before ResultPath looks right, and so do Pass Parameters and ArrayContains.

A few things to line up with AWS parity:

1. The wildcard projection works for the dot form, but AWS uses the bracket form, like `$.Regions[*].RegionName`. Right now that bracket form actually throws, because the segment parser tries to read `*` as an array index. Worth supporting `[*]` since that is the form almost every state machine uses:

```
$.Regions.*.RegionName     works today
$.Regions[*].RegionName    the AWS form, currently throws
```

2. ResultSelector also applies to Map and Parallel in AWS, not just Task. The headline use is flattening an array of arrays:

```json
"ResultSelector": { "flattenArray.$": "$[*][*]" }
```

so it would be good to apply it in those two states too. That one also needs the `[*]` wildcard from above, so the two compose.

3. Good catch on absent fields for IsPresent. One AWS nuance worth keeping: a field that exists with a null value still counts as present, so `{"x": null}` with `$.x` and IsPresent should be true. Today it returns false because an absent path and an explicit null both collapse to the same null node, so distinguishing missing from null in resolvePath would close that last gap.

None of this is blocking, the core behavior is solid. Thanks for pushing these forward.

abanna:
Thanks for the thorough review, @hectorvent! Addressed all three in 356b5d10:

1. **Bracket wildcard** — path resolution now normalizes bracket notation into segments, so `$.Regions[*].RegionName` works identically to the dot form, and root-level `$[*][*]` flattens an array of arrays. Multi-level wildcards flatten as JSONPath does.
2. **ResultSelector on Map/Parallel** — now applied to both states before `ResultPath` merges, mirroring the Task path. The flatten use case (`{"flattenArray.$": "$[*][*]"}`) composes with #1.
3. **IsPresent + explicit null** — added `resolvePathNode` which preserves the missing-vs-null distinction that `resolvePath` collapses; `{"x": null}` with `$.x` + IsPresent now returns true.

Added unit coverage (bracket form, `$[*][*]` flatten, numeric index, missing-vs-null) plus integration tests for Map/Parallel ResultSelector and the explicit-null Choice case. Also merged latest `main`.

abanna:
Thanks @greptile-apps — reviewed all three, two were spot-on and are fixed in 830c1cbd:

- **P1 — nulls dropped from wildcard projection:** Valid. Fixed — the projection now skips only `isMissingNode()` matches and keeps explicit nulls, so `$[*].field` over `[{"field":null},{"field":"x"}]` yields `[null,"x"]`. This is also consistent with the missing-vs-null distinction this PR added for `IsPresent`. Added a unit test.
- **P2 — `States.ArrayContains` silent false on non-array:** Valid. Fixed — it now throws `States.Runtime`, matching AWS and the existing `States.ArrayLength` behavior in this same file. Added a unit test.
- **P2 — `split("\\.")` and dotted field keys:** Correct that bracket-quoted keys (`$.a['b.c']`) aren't supported, but this is pre-existing behavior (the prior resolver split the same way) and bracket-quoted keys would have thrown before too. ASL reference paths rarely use dotted keys, so I documented it as a known limitation in `splitPathSegments` rather than adding quoted-key parsing here. Happy to open a follow-up if it comes up in practice.

abanna:
Thanks, @hectorvent — all three parity items are in the branch (they landed after your comment).

**1. Bracket-form wildcard** `$.Regions[*].RegionName` — `splitPathSegments` normalizes `[*]`→`.*` and `[n]`→`.n`, so both the AWS bracket form and the dot form resolve, and `$[*][*]` flattens an array of arrays.

**2. `ResultSelector` on Map & Parallel** — applied at all three sites (Task, Map, Parallel) via `resolveParameters`, so `"flattenArray.$": "$[*][*]"` composes with the wildcard.

**3. `IsPresent` treats an explicit null as present** — the choice path uses `resolvePathNode` (which preserves `MissingNode`) instead of `resolvePath` (which collapses missing→null), so `{"x":null}` + `$.x` is present while an absent `$.x` is not.

Covered by `AslExecutorPathIntrinsicsTest` plus the Map/Parallel, Pass-Parameters, and IsPresent integration tests. Branch is up to date with `main` (merge) and green in the build.

hectorvent (approved):
Thanks @abanna

PR #1559, opened 2026-06-25
feat(iam): add AmazonRDSEnhancedMonitoringRole managed policy

dnlopes:
## Summary

Seeds `AmazonRDSEnhancedMonitoringRole` into the catalog of AWS-managed policies that floci creates at startup.

- Adds the policy definition to `AwsManagedPolicies` under `/service-role/`, matching the real AWS ARN `arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole`
- Without this entry callers that attach this policy to an RDS monitoring role (e.g. via Terraform's `aws_iam_role_policy_attachment`) get a `NoSuchEntityException` on startup

## Tests

- Extended `seedAwsManagedPolicies()` in `IamServiceTest` to assert the new policy is reachable by ARN with the correct name and path
- Added `attachRdsEnhancedMonitoringRolePolicyToRole()` unit test that attaches the policy to a role and verifies the attachment
- Added `getRdsEnhancedMonitoringPolicy()` integration test (`@Order(17)`) that retrieves the policy via the `GetPolicy` wire action and checks both `PolicyName` and `Arn`


hectorvent (approved):
Thanks @dnlopes 
Solid implmentation

Issue #797, opened 2026-05-11
SES SendRawEmail rejects requests when Source param is omitted but MIME From header is present


pditommaso:
## Summary

Floci's `SendRawEmail` operation returns `400 "Source email is required"` when the `Source` request parameter is omitted, even when the raw MIME message contains a valid `From:` header. This diverges from real AWS SES and from LocalStack, both of which fall back to the MIME `From:` header when `Source` is absent.

## Expected behavior

Per the [AWS SES SendRawEmail API docs](https://docs.aws.amazon.com/ses/latest/APIReference/API_SendRawEmail.html):

> **Source** – The identity's email address. [...] If you do not specify a value for this parameter, you must specify a `"From:"` header in the raw message.

A `SendRawEmail` request with no `Source` parameter but a valid `From:` header in the MIME body should be accepted, using the From header as the source identity. Real AWS SES and LocalStack both behave this way.

## Observed behavior in floci

Floci returns:

```
HTTP/1.1 400 Bad Request
SesException: Source email is required.
```

regardless of whether the MIME body contains a `From:` header.

## Reproduction

```bash
# Start floci
docker run --rm -d --name floci \
  -p 4566:4566 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  floci/floci:latest
sleep 5

# Verify a sender identity
aws --endpoint-url http://localhost:4566 --region us-east-1 \
    ses verify-email-identity --email-address sender@example.com

# Build a minimal raw email with From, To, Subject, body (base64-encoded)
RAW=$(printf 'From: sender@example.com\r\nTo: recipient@example.com\r\nSubject: hello\r\n\r\nbody\r\n' | base64)

# (1) SendRawEmail WITHOUT --source — fails in floci, succeeds against real AWS / LocalStack
aws --endpoint-url http://localhost:4566 --region us-east-1 \
    ses send-raw-email --raw-message "Data=$RAW"
# → An error occurred (400) when calling the SendRawEmail operation: Source email is required.

# (2) Same request WITH explicit --source — succeeds in floci
aws --endpoint-url http://localhost:4566 --region us-east-1 \
    ses send-raw-email \
    --source sender@example.com \
    --raw-message "Data=$RAW"
```

Request (1) succeeds against real AWS SES and against LocalStack (community edition); fails against floci.

## Environment

- floci `1.5.14` (native build)
[...]

PR #1561, opened 2026-06-25
fix(ses): make FromEmailAddress optional for v2 SendEmail

okinaka:
## Summary

Floci's v2 `SendEmail` handler rejected any request without `FromEmailAddress`, diverging from AWS where the field is marked `Required: No`. This aligns the behavior per content type, matching what real AWS returns (confirmed via probes against a live account, including the raw HTTP error bodies):

- **Raw** content takes its sender from the MIME `From` header, so `FromEmailAddress` may be omitted. When neither is present, the send fails with `BadRequestException` `Missing required header 'From'.`. (This also corrects the v1 `SendRawEmail` path, which previously accepted a From-less raw message.)
- **Simple** content with no `FromEmailAddress` returns `BadRequestException` with a `null` message body (`{"message":null}`), exactly as AWS does.
- **Templated** content with no `FromEmailAddress` returns `BadRequestException` `Source cannot be empty`.

Related to #797 (the v2 sender-resolution counterpart to the v1 `SendRawEmail` MIME-`From` handling).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the AWS API reference for [SendEmail](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_SendEmail.html) (`FromEmailAddress` Required: No) and against real AWS via probes that captured the exact wire responses for each content type with the sender omitted: Raw-with-MIME-From succeeds (200), Raw-with-no-From → 400 `Missing required header 'From'.`, Simple → 400 `{"message":null}`, Templated → 400 `Source cannot be empty`. A new v2 integration test pins all four cases.

**Reproduce** the Raw behavior with the AWS CLI against a running Floci (`./mvnw quarkus:dev`). Before this PR both calls failed with `400 FromEmailAddress is required.`:

```bash

# Raw, no FromEmailAddress but the MIME has a From header -> 200, returns a MessageId
aws sesv2 send-email --cli-input-json '{"Destination":{"ToAddresses":["r@example.com"]},
  "Content":{"Raw":{"Data":"'"$(printf 'From: s@floci.test\r\nSubject: x\r\n\r\nb' | base64 | tr -d '\n')"'"}}}'

# Raw, no From anywhere -> BadRequestException: Missing required header 'From'.
aws sesv2 send-email --cli-input-json '{"Destination":{"ToAddresses":["r@example.com"]},
  "Content":{"Raw":{"Data":"'"$(printf 'Subject: x\r\n\r\nb' | base64 | tr -d '\n')"'"}}}'
```

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesController.java
@@ -270,11 +270,10 @@ public class SesController {
             JsonNode request = objectMapper.readTree(body);
             requireJsonObject(request);
 
+            // FromEmailAddress is optional per the AWS v2 contract. Each content type below
+            // enforces the sender requirement the way AWS does: Raw can take its From from the
+            // MIME message, while Simple and Templated require FromEmailAddress.
             String fromEmailAddress = request.path("FromEmailAddress").asText(null);
-            if (fromEmailAddress == null || fromEmailAddress.isBlank()) {
-                throw new AwsException("BadRequestException",
-                        "FromEmailAddress is required.", 400);
-            }
 
             JsonNode destination = requireObjectOrAbsent(request, "Destination");
             List<String> toAddresses = jsonArrayToList(destination.path("ToAddresses"));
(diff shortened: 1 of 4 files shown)


hectorvent (approved):
Thanks @okinaka, 
Always appreciate your contribution.

PR #1562, opened 2026-06-25
feat(stepfunctions): add state machine version APIs (Publish/List/DeleteStateMachineVersion)

abanna:
## Summary

Adds the Step Functions state-machine **version** APIs: `PublishStateMachineVersion`, `ListStateMachineVersions`, `DeleteStateMachineVersion`, plus publish-on-create (`publish=true` on `CreateStateMachine`).

**Why:** the Terraform AWS provider calls `ListStateMachineVersions` immediately after creating an `aws_sfn_state_machine` (to read `state_machine_version_arn`). Floci returned `UnsupportedOperation`, so **both `terraform apply` and `terraform destroy` failed** for any state machine. Versions are tracked per state machine as `<stateMachineArn>:<n>`.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

Standard JSON 1.0 actions under `AWSStepFunctions.*`. `ListStateMachineVersions` returns an empty list when nothing has been published (the default, `publish=false`), matching AWS.

## Checklist

- [x] `./mvnw test` passes locally (clean: `StepFunctionsVersionsIntegrationTest` + existing SFN suites — 39 green; JDK 25 container)
- [x] Integration test for publish/list/delete + create-with-publish
- [x] Conventional Commits

## Verified end-to-end

`terraform apply` then `terraform destroy` of a module covering DPS's resource set (Lambda, DynamoDB, SQS, S3, API Gateway v2, EventBridge, **Step Functions**, ECR, CloudWatch Logs, IAM) against a live Floci: **8 added, 8 destroyed**, clean. Before this PR, the run failed at `aws_sfn_state_machine` with `UnsupportedOperation: ListStateMachineVersions`.


abanna:
Thanks @greptile-apps:

- **`listStateMachineVersions` returned the internal list** (P2): Valid, fixed in 3e913ee7 — now returns `List.copyOf(...)`.
- **`deleteStateMachineVersion` silent success on missing** (P1): Here I'll respectfully push back — AWS's `DeleteStateMachineVersion` is idempotent and does **not** return `ResourceNotFound` for a missing version (its only documented errors are `ConflictException`, `ValidationException`, and `InvalidArn`), mirroring `DeleteStateMachine`'s eventually-consistent delete. So returning 200 for an already-absent version is the AWS-correct behavior, and throwing `ResourceNotFound` would actually be a deviation. I kept it idempotent.

abanna:
Re: greptile's P1 on `deleteStateMachineVersion` ("silent success when the version doesn't exist") — I verified this against the AWS API reference and it's a **false positive**. `DeleteStateMachineVersion` documents only `ConflictException`, `InvalidArn`, and `ValidationException` — there is no `ResourceNotFound`, and AWS returns HTTP 200 with an empty body for a missing/already-deleted version (the call is idempotent). Adding a throw would break AWS protocol compatibility (and the Terraform `aws_sfn_state_machine` flow this PR supports). The sibling `deleteStateMachine` is intentionally idempotent the same way, so the current `ifPresent` + `removeIf` no-op is correct. The P2 on `listStateMachineVersions` is already addressed (now returns `List.copyOf(...)`).

hectorvent:
Really nice, Alex, and great that it unblocks the Terraform apply/destroy flow. The shapes all line up with AWS: the `stateMachineVersionArn` / `creationDate` outputs, the `<stateMachineArn>:<n>` ARN format, and the publish-on-create field are all correct.

Two things on parity:

1. `ListStateMachineVersions` should return versions newest first (AWS sorts descending by creation time). Right now it returns them in insertion order, so once a state machine has more than one version the order is reversed, and the Terraform provider reads the version ARN off this list so it can latch onto the wrong one:

```
publish v1, v2, v3  ->  ListStateMachineVersions
  AWS:   [ ...:3, ...:2, ...:1 ]
  Floci: [ ...:1, ...:2, ...:3 ]
```

A sort on the snapshot fixes it (descending by `creationDate`, tie-breaking on the version number since `creationDate` is second resolution).

2. Listing versions for a state machine that does not exist currently throws `StateMachineDoesNotExist` (via `describeStateMachine`), but AWS returns `InvalidArn` for `ListStateMachineVersions`, and `StateMachineDoesNotExist` is not one of that operation's declared errors. Publish is fine since it does declare `StateMachineDoesNotExist`.

Neither is blocking. The description/revisionId fields being ignored and no pagination are fine to defer. The lifecycle test is nice, maybe add a two version case to lock the ordering once fixed.

hectorvent (approved):

abanna:
Thanks, @hectorvent — both parity items done.

**1. `ListStateMachineVersions` now returns newest-first.** Sorted descending by `creationDate`, tie-broken on the version number (since `creationDate` is second-resolution), so the Terraform provider reads the latest version ARN off `[0]`. Added an ordering test that publishes three versions and asserts `:3, :2, :1`.

**2. Non-existent state machine → `InvalidArn`.** `ListStateMachineVersions` was delegating to `describeStateMachine`, which raises `StateMachineDoesNotExist` — not a declared error for this operation. It now throws `InvalidArn` for a missing state machine. `PublishStateMachineVersion` is left using `describeStateMachine`, since it *does* declare `StateMachineDoesNotExist`. Added a missing-state-machine test asserting `InvalidArn`.

(For the bot thread: Greptile's earlier P1 on `deleteStateMachineVersion` returning silent-success was withdrawn by Greptile itself as AWS-correct — `DeleteStateMachineVersion` declares only `ConflictException`/`ValidationException`, so the idempotent behavior stands.)

Branch is up to date with `main` (merge) and green in the build.

PR #1564, opened 2026-06-25
feat(stepfunctions): add ecs:runTask service integration

abanna:
## Summary

Adds the Step Functions **`ecs:runTask`** optimized service integration so state machines can launch ECS tasks and (optionally) wait for them to finish:

- `arn:aws:states:::ecs:runTask` — request-response; returns the `{ Tasks, Failures }` envelope
- `arn:aws:states:::ecs:runTask.sync` — blocks until the task reaches `STOPPED`

Previously `AslExecutor` had no `ecs:runTask` branch, so a state machine using these resources failed with *"Unsupported resource"* even though Floci already has a docker-backed ECS executor. This wires the two together, unblocking the common "launch a task, wait for it to complete" orchestration pattern.

## Details

- **Routing:** `invokeResource` now recognizes `arn:aws:states:::ecs:runTask[.sync]`. The `.waitForTaskToken` suffix is already stripped upstream, so that variant launches the task while the token future blocks for `SendTaskSuccess`.
- **Casing bridge:** Step Functions optimized integrations use PascalCase member names (`TaskDefinition`, `Tasks[].TaskArn`, `Containers[].ExitCode`), whereas Floci's ECS handlers use the lowerCamelCase of the ECS data-plane API. A small recursive key-recasing helper maps parameters in (Pascal→camel) and results out (camel→Pascal), reusing `EcsJsonHandler.taskNode` / `parseContainerOverrides` so the output matches the ECS task shape exactly.
- **`.sync` semantics:** polls `DescribeTasks` until `STOPPED`, then fails the state with `States.TaskFailed` when an essential container exits non-zero or the task never ran a container — matching real Step Functions.

## Tests

- `AslExecutorRecaseKeysTest` — unit coverage of the casing bridge (nested objects/arrays, inverse mapping, scalar round-trip).
- `StepFunctionsEcsRunTaskIntegrationTest` — PascalCase `Tasks`/`Failures` envelope, `TaskDefinition` validation, and `.sync` error propagation. Runs under an ECS **mock profile** so the wiring is asserted independently of a container runtime/image registry.

Full `stepfunctions` + `ecs` package run: **211 tests, 0 failures** (the 42 existing ECS integration tests are unaffected by the `taskNode`/`parseContainerOverrides` visibility widening).


abanna:
Thanks @greptile-apps — both valid, fixed in 1f58c1f2:

- **`.sync` orphans tasks beyond the first when Count > 1** (P1): Fixed — `.sync` now polls **all** launched task ARNs, waits for every one to reach `STOPPED`, and fails the state if **any** task's essential container exits non-zero or never ran. A single task still returns its description; multiple tasks return the array. (The STOPPED happy-path can't be exercised under the `floci.services.ecs.mock` profile — mock tasks never stop — so this is verified by the existing wiring/validation tests plus review, consistent with how the original .sync happy-path was left to live/empirical testing.)
- **Unknown mode suffix silently falls into `.sync`** (P2): Fixed — only `""` (request-response) and `.sync` are accepted; anything else now fails with `States.TaskFailed`. (`.waitForTaskToken` is stripped upstream in `executeTaskState`, so it arrives here as `""` and is unaffected.)

abanna:
Thanks @greptile-apps — valid P1, fixed in the latest commit. `ecsTaskFailureCause` now tracks `null` exit codes separately: a STOPPED container with no recorded exit code (OOM-killed, failed to start, force-stopped) is treated as a failure ("Essential container stopped without an exit code") rather than a clean exit, matching real Step Functions.

abanna:
Thanks @greptile-apps — valid P1, fixed in the latest commit. `ecsTaskFailureCause` now resolves the task definition (`describeTaskDefinition`) and ignores containers explicitly marked `essential: false`, so a non-essential sidecar exiting non-zero no longer fails the state — matching real Step Functions. Containers that can't be cross-referenced (or when the task def can't be resolved) default to essential, preserving the conservative behavior.

hectorvent:
This is a really nice integration, Alex. The PascalCase recasing bridge is the right call given ECS is uniformly camelCase, and having `.sync` return the single task's DescribeTasks description matches what AWS does (it returns the task description, not the runTask output).

Two parity things and a couple of small notes:

1. Request-response should not fail the state on a failure. Per the ECS integration docs, `ecs:runTask` (request-response) returns the `{Tasks, Failures}` envelope even when `Failures` is non-empty and does not fail the task. Only the `.sync` and `.waitForTaskToken` patterns fail on a non-empty `Failures` (and AWS uses the `AmazonECS.Unknown` error name for that). Right now the `launched.isEmpty()` check throws `States.TaskFailed` before the mode branch, so it fires for request-response too:

```
ecs:runTask (request-response), task cannot be placed
  AWS:   200, result = { "Tasks": [], "Failures": [ ... ] }, execution continues
  Floci: States.TaskFailed
```

Gating that empty/failure handling on `.sync`/`.waitForTaskToken` (and using `AmazonECS.Unknown` there) would line it up.

2. `NetworkConfiguration` is dropped. AWS supports it (plus `PlacementConstraints`, `PlacementStrategy`, `PropagateTags`, `EnableExecuteCommand`) as runTask parameters, and awsvpc/FARGATE tasks need it. It looks bounded by `ecsService.runTask(...)`'s signature more than anything, so mostly a note for real Docker runs rather than the mock profile.

Smaller stuff, all fine to defer: with `Count > 1` the `.sync` result returns an array where AWS's shape is less clear (it tracks the first task), and the 60s internal poll cap means a longer task would hit `States.Timeout` where real `.sync` waits for the state's `TimeoutSeconds`.

The recasing unit test and the mock-profile integration test are a nice touch. Thanks for wiring this up.

abanna:
Thanks for the review, @hectorvent — both parity items are addressed.

**1. Request-response no longer fails on a placement failure.** The `launched.isEmpty()` check that threw `States.TaskFailed` before the mode branch is gone. Request-response now returns the `{Tasks, Failures}` envelope (possibly with empty `Tasks`) and lets the execution continue. Only `.sync` and `.waitForTaskToken` fail the state on a placement failure, and they now do so with the `AmazonECS.Unknown` error name. `.waitForTaskToken` arrives as the bare `runTask` resource (suffix stripped upstream), so it's distinguished by its non-null task token to get the fail-on-failure semantics.

**2. `NetworkConfiguration` is no longer dropped.** It's parsed (reusing the ECS handler's `parseNetworkConfiguration` after a PascalCase→camelCase recasing) and threaded through `runTask` onto the launched `EcsTask` — symmetrically across all three runTask entry points (the real RunTask API, the service reconciler, and this integration). I deliberately stopped short of emulating the awsvpc ENI `attachments` in the task description and of the `PlacementConstraints` / `PlacementStrategy` / `PropagateTags` / `EnableExecuteCommand` request params: those have no observable effect in the mock profile and would need real Docker networking — the "note for real Docker runs" you flagged. Happy to open a follow-up for the ENI `attachments` echo if you'd like it tracked.

New unit tests cover request-response-doesn't-fail, `.sync` → `AmazonECS.Unknown`, and the `NetworkConfiguration` threading. Branch is up to date with `main` (merge) and green in the build.

PR #1565, opened 2026-06-25
fix(codebuild): stabilize compatibility tests

shrimptails-f:
## Summary

Stabilize CodeBuild compatibility tests by replacing unstable image/command assumptions in test fixtures.

- replace `public.ecr.aws/docker/library/alpine:latest` with a digest-pinned BusyBox image shared by both CodeBuild compatibility test scenarios
- switch the OS demo buildspec from `/etc/os-release` parsing to `uname -a` so the test no longer assumes distro-specific files
- keep the directory traversal bounded with `find / -maxdepth 3 -type d` to avoid the heavier full-tree listing approach

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The previous compatibility tests depended on `public.ecr.aws/docker/library/alpine:latest`, which is a mutable external image tag and can fail independently of Floci changes due to image drift or registry pull limits. They also assumed distro-specific OS metadata files. This change pins the image by digest and narrows the command assumptions to utilities available in the selected image, while keeping the test focus on CodeBuild execution and artifact upload behavior.

On PR #1295, CI failed for commit `cd5d289` but passed on the empty follow-up commit `04349b3`. That suggests the failure was not deterministically tied to the source diff and is consistent with test or external dependency instability rather than a stable product behavior regression.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CodeBuildTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CodeBuildTest.java
@@ -45,6 +45,9 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 class CodeBuildTest {
 
+    private static final String CODEBUILD_TEST_IMAGE =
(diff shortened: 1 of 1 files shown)


hectorvent (approved):
Thanks @shrimptails-f 
I appreciate your contribution

PR #1566, opened 2026-06-25
feat(stepfunctions): run executions under the execution's account

abanna:
## Summary

`AslExecutor` runs each execution on a worker-thread pool that has **no active CDI request scope**. As a result, service integrations (Lambda, DynamoDB, SQS, …) and the execution-store writes fell back to the configured **default account** instead of the account the execution belongs to.

So an execution started under account *B* had its integrations executed against account *A*'s resources — e.g. a `dynamodb:putItem` targeting a table that exists only in account *B* fails with `ResourceNotFound`, and execution-record updates land in the wrong namespace.

## Fix

Wrap `doExecute` in a synthetic request scope whose account is the one encoded in the **state machine ARN**, so integrations and store writes resolve to the execution's own account. This mirrors the existing `CurEmissionScheduler#runUnderAccount` pattern (Arc `ManagedContext` activate → `RequestContext.setAccountId` → terminate).

It falls back to running the body directly when Arc is not running, so plain unit tests that construct `AslExecutor` without a CDI container are unaffected.

## Test

`StepFunctionsAccountContextIntegrationTest` — an execution started under a non-default account (`000000000111`) writes via `dynamodb:putItem` to a table that exists **only** in that account. The execution **succeeds**, the item is readable in that account, and a different account (`000000000222`) has no such table. Without the fix the integration resolves to the default account and the execution fails.

Regression: the existing `StepFunctionsDynamoDbIntegrationTest` (20) and `StepFunctionsSqsIntegrationTest` (7), which run under the default account, are unchanged; the plain-unit `AslExecutorCatchTest` exercises the Arc-not-running fallback. Focused run: **30 tests, 0 failures**.

## Notes

This is independent of, and complementary to, the `ecs:runTask` integration (#1564): together they make Step Functions–driven cross-account provisioning (a state machine in one account launching/provisioning into its own account) resolve correctly.


abanna:
Hi @hectorvent 👋

A batch of my open PRs have their CI sitting at `action_required` — the fork-PR "approve and run workflows" gate — so no checks have actually run on them yet. When you have a moment, could you hit **Approve and run workflows** for these?

- #1566 — feat(stepfunctions): run executions under the execution's account
- #1564 — feat(stepfunctions): add ecs:runTask service integration
- #1562 — feat(stepfunctions): add state machine version APIs
- #1558 — feat(stepfunctions): ResultSelector, Pass Parameters, ArrayContains, wildcard projection
- #1556 — feat(stepfunctions): add aws-sdk CloudFormation/EC2 and S3 PutObject integrations
- #1552 — feat(iam): enforce AssumeRole trust policies when enforcement is enabled
- #1551 — feat(cloudformation): add StackSets with account-aware provisioning

Happy to rebase any of them on `main` if you'd prefer a fresh run. Thanks!

hectorvent:
Triggering them all. Thanks!

abanna:
> Triggering them all. Thanks!

Thank you for the quick response! ❤️

abanna:
Thanks @greptile-apps — valid, fixed in the latest commit: `runUnderExecutionAccount` now saves the previous account ID and restores it in `finally` when it runs inside an already-active scope, so it can't leave the execution's account behind on a reused thread. (Executions run on background workers that normally have no active scope, so this hardens the synchronous-caller path.)

hectorvent (approved):
Thanks @abanna, 
Solid fix

PR #1567, opened 2026-06-25
Honor queue-level DelaySeconds for standard SQS queues

Croway:
## Summary
- Standard SQS queues with a `DelaySeconds` attribute were not applying the delay to messages sent without an explicit per-message `DelaySeconds`
- Root cause: the handler→service API used primitive `int` (default 0), making it impossible to distinguish "omitted" from "explicit 0"
- Changed `int delaySeconds` to `Integer delaySeconds` across the handler/service chain so `null` falls back to the queue attribute while `Integer(0)` overrides it
- Added tests for both queue-level delay application and explicit zero override

## Test plan
- [x] `mvn verify` passes all tests including new standard queue delay tests
- [x] Existing FIFO delay tests still pass
- [x] Camel Spring Boot `SqsDelayedQueueTest` passes with the fixed floci image

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/SqsJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/SqsJsonHandler.java
@@ -195,7 +195,8 @@ public class SqsJsonHandler {
     private Response handleSendMessage(JsonNode request, String region) {
         String queueUrl = request.path("QueueUrl").asText(null);
         String messageBody = request.path("MessageBody").asText(null);
-        int delaySeconds = request.path("DelaySeconds").asInt(0);
+        JsonNode delayNode = request.path("DelaySeconds");
+        Integer delaySeconds = parseOptionalInteger(delayNode, "DelaySeconds");
         String messageGroupId = request.path("MessageGroupId").asText(null);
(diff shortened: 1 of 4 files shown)


hectorvent:
Nice fix, Federico, and it lines up exactly with how AWS models this. `DelaySeconds` is a `NullableInteger` on both `SendMessage` and the batch entry precisely so the service can tell an omitted value from an explicit `0`, so moving from `int` to `Integer` is the right root cause fix. The null to queue default, explicit `0` to override behavior matches the docs ("if you don't specify a value, the default value for the queue applies"), and I like that the tests cover both directions and you dropped the stale follow-up comment.

One small thing: in the query handler `getIntegerParam` swallows a parse error and returns null, so a malformed `DelaySeconds` like `"abc"` gets treated as omitted and quietly picks up the queue default, whereas AWS returns a validation error for a non-integer value. Low impact since clients send valid integers, just noting it.

Thanks for tracking this down.

Croway:
good catch, I'll throw an exception in case of `NumberFormatException`

hectorvent:
Thanks @Croway, could you please take a look at the `greptile-apps` comments?

hectorvent (approved):
Thanks @Croway

PR #1573, opened 2026-06-25
fix(iam): resolve AWS-managed policies from any account context

### LINKED ISSUE #1572: [BUG] IAM: AWS-managed policies (data.aws_iam_policy) only resolve for the default account
AWS behavior
------------
AWS-managed policies have account-agnostic ARNs (`arn:aws:iam::aws:policy/...`) and are visible to every account. `aws iam get-policy --policy-arn arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole` succeeds regardless of the caller account.

Current Floci behavior
----------------------
Floci seeds a catalog of managed policies at startup (`AwsManagedPolicies`), but the IAM policy store is account-partitioned (`AccountAwareStorageBackend`). Seeding runs at `@PostConstruct` with no request context, so the policies land under the **default** account only. A caller in any other account (e.g. an assumed-role or a 12-digit access-key credential routed to that account) gets `NoSuchEntity` from `GetPolicy`.

Reproduction (against a running Floci):

```
# default account — works
AWS_ACCESS_KEY_ID=test aws --endpoint-url=http://localhost:4566 iam get-policy \
  --policy-arn arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole   # OK

# 12-digit account credential — fails
AWS_ACCESS_KEY_ID=111111111111 aws --endpoint-url=http://localhost:4566 iam get-policy \
  --policy-arn arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
# NoSuchEntity: Policy ... does not exist.
```

Impact
------
Terraform `data "aws_iam_policy"` lookups of managed policies fail for any non-default account. This breaks common role modules (lambda/ECS execution roles attaching `AWSLambdaBasicExecutionRole`, `AmazonECSTaskExecutionRolePolicy`, `AWSXRayDaemonWriteAccess`, etc.) whenever the apply runs under a non-default account.

Fix
---
Managed policies are global, so serve them from an account-independent catalog keyed by ARN; `GetPolicy`/`GetPolicyVersion` resolve them for any caller. (Also seeds a few commonly-attached managed policies that were missing: `AWSCloudFormationFullAccess`, `AWSXRayDaemonWriteAccess`, `AmazonElasticFileSystemClientFullAccess`.)

abanna:
Closes #1572.

## Summary

AWS-managed policies (`arn:aws:iam::aws:policy/...`) are global — not owned by any account — but Floci seeded its `AwsManagedPolicies` catalog into the account-partitioned policy store at startup under the **default** account only. A caller in any other account (e.g. an assumed-role / 12-digit access-key credential) got `NoSuchEntity` from `GetPolicy`, which breaks Terraform `data "aws_iam_policy"` lookups of managed policies (lambda/ECS execution-role modules attaching `AWSLambdaBasicExecutionRole`, `AmazonECSTaskExecutionRolePolicy`, `AWSXRayDaemonWriteAccess`, …).

## What changed

- **`IamService`** now holds a global, account-independent catalog of managed policies (built once from `AwsManagedPolicies`, keyed by ARN). `getPolicy` serves managed ARNs from the catalog, so they resolve for any caller; `GetPolicyVersion`/tag reads route through `getPolicy` and are fixed transitively. The catalog is still mirrored into the default-account store so `ListPolicies(Scope=AWS)` keeps listing them.
- **`AwsManagedPolicies`** gains three commonly-attached entries: `AWSCloudFormationFullAccess`, `AWSXRayDaemonWriteAccess`, `AmazonElasticFileSystemClientFullAccess`.

## Testing

- `mvn test -Dtest="Iam*,RegionResolver*"` green (149 tests). New `IamManagedPolicyAccountScopeTest`: a managed policy resolves under a non-default request account while a customer (account-scoped) policy stays isolated; and the catalog includes the new entries.
- **Verified live**: `get-policy` for managed policies (incl. the 3 new ones) now succeeds under both the default account and a 12-digit account credential — previously `NoSuchEntity` for the latter.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
@@ -53,6 +53,12 @@ final class AwsManagedPolicies {
                 "Provides permissions required to use the CloudWatch agent on servers."),
         new ManagedPolicyDef("AWSLambdaFullAccess", "/",
                 "Provides full access to Lambda, S3, DynamoDB, CloudWatch Metrics and Logs."),
+        new ManagedPolicyDef("AWSCloudFormationFullAccess", "/",
+                "Provides full access to AWS CloudFormation."),
+        new ManagedPolicyDef("AWSXRayDaemonWriteAccess", "/",
+                "Allows write permissions to the AWS X-Ray daemon."),
+        new ManagedPolicyDef("AmazonElasticFileSystemClientFullAccess", "/",
+                "Provides root client access to an Amazon EFS file system."),
 
         // Lambda execution role policies
         new ManagedPolicyDef("AWSLambdaBasicExecutionRole", "/service-role/",
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
@@ -25,6 +25,7 @@ import java.time.Instant;
 import java.util.Optional;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -50,6 +51,13 @@ public class IamService {
     private final StorageBackend<String, SessionCredential> sessions;
     private final RegionResolver regionResolver;
     private final boolean seedDeployerPrincipal;
+
+    /**
+     * AWS-managed policies (arn:aws:iam::aws:policy/...), keyed by ARN. These are global —
+     * not owned by any account — so they live here rather than in the account-partitioned
+     * {@link #policies} store, and {@link #getPolicy} resolves them for any caller.
+     */
+    private final Map<String, IamPolicy> awsManagedPolicies = buildAwsManagedPolicies();
 
     @Inject
     public IamService(StorageFactory storageFactory, EmulatorConfig config, RegionResolver regionResolver) {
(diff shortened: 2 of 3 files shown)


hectorvent:
Really nice fix, serving managed policies from the global catalog for GetPolicy is exactly right.

One thing on the same kind of bug though. `ListPolicies(Scope=AWS)` still only returns them under the default account, since the catalog is mirrored into the default account store and `listPolicies` scans the request account's partition ([`IamService.java:434`](https://github.com/floci-io/floci/blob/feat/iam-managed-policies-global/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java#L434)). In real AWS these are visible to every account (moto seeds the full managed set into each account's backend, and the ARN itself has no account, `arn:aws:iam::aws:policy/...`), so a caller in another account gets an empty `ListPolicies(Scope=AWS)`.

That leaves the lookup by name still broken across accounts. `data "aws_iam_policy" { name = "AWSLambdaBasicExecutionRole" }` resolves through ListPolicies plus a name filter, so it keeps missing managed policies for any caller outside the default account, the same issue this PR already fixes for the ARN path.

Probably fine to do as a follow up rather than block on it, but worth serving `Scope=AWS` and `All` from the same catalog so both read paths line up. Thanks for this!

abanna:
Thanks for the close read, @hectorvent — you're right. Pushed a follow-up commit (`05bfc619`) to this PR.

**What changed:** `listPolicies` now serves AWS-managed policies from the same global catalog `GetPolicy` uses, for both `Scope=AWS` and `Scope=All`, so a caller in **any** account gets the full managed set — not just the default account. That also fixes the `data "aws_iam_policy" { name = ... }` path, since it resolves through `ListPolicies` + a name filter. To keep the default account from listing them twice (it still has the seed-time mirror), managed ARNs are excluded from the account-partition scan and served only from the catalog — so `GetPolicy` and `ListPolicies` line up.

**Tests:** added regression coverage in `IamManagedPolicyAccountScopeTest` — `Scope=AWS`/`All`/name-lookup return the full catalog from a non-default account, `Scope=Local` stays caller-scoped, and the default-account path doesn't double-list the mirrored managed policy.

One related thing I noticed: a few other read paths that resolve *attached* policy ARNs via the account-scoped `policies.get(...)` (e.g. authorization-details / policy simulation) have the same cross-account gap for managed policies. Happy to take that as a separate follow-up if you think it's worth it.

hectorvent (approved):
Thanks @abanna

PR #1575, opened 2026-06-25
fix(lambda): resolve handler paths with a leading "./" against the package

### LINKED ISSUE #1574: [BUG] Lambda: CreateFunction rejects handlers with a leading "./" (Handler file not found)
AWS behavior
------------
Node.js (and other file-based) Lambda handlers may be specified with a leading `./`, e.g. `./v1/lambda-handlers/entry.handler`. AWS resolves this against the deployment package the same as `v1/lambda-handlers/entry.handler`. The terraform-aws-modules/lambda module emits handlers in this `./`-prefixed form.

Current Floci behavior
----------------------
`CreateFunction` validates that the handler file exists in the zip. `resolveHandlerFilePath` strips the `.handler` export suffix, yielding `./v1/lambda-handlers/entry`, then compares it against package entries normalized **without** a `./` prefix (`v1/lambda-handlers/entry`). They never match, so CreateFunction fails:

```
InvalidParameterValueException: Handler file ./v1/lambda-handlers/entry not found in deployment package
```

A handler without the prefix (e.g. `index.handler` or `bdd-lambda-handler.handler`) works; only the `./`-prefixed nested form fails.

Reproduction
------------
Create a function with `Handler = "./v1/lambda-handlers/entry.handler"` and a zip containing `v1/lambda-handlers/entry.js` → 400 "Handler file not found".

Fix
---
Normalize a leading `./` away from the resolved handler module path before matching it against deployment-package entries.

abanna:
Closes #1574.

## Summary

A Node/file-based Lambda handler can be specified with a leading `./` — e.g. `./v1/lambda-handlers/entry.handler`, as the terraform-aws-modules/lambda module (and GE DPS) emit. `CreateFunction` validates the handler file exists in the deployment package; `resolveHandlerFilePath` stripped only the `.handler` export suffix, leaving `./v1/lambda-handlers/entry`, which never matched the package entry `v1/lambda-handlers/entry.js` (stored without the `./`). The function was rejected with `InvalidParameterValueException: Handler file ... not found in deployment package`.

Handlers without the prefix (`index.handler`, `bdd-lambda-handler.handler`) already worked; only the `./`-prefixed nested form failed.

## What changed

- `LambdaService.resolveHandlerFilePath` normalizes a leading `./` away from the resolved module path before the package-entry match (non-Python/.NET file-based runtimes).

## Testing

- `mvn test -Dtest=LambdaIntegrationTest` green (23 tests). New `createFunctionWithDotSlashNestedHandler`: a zip with `v1/lambda-handlers/entry.js` + `Handler="./v1/lambda-handlers/entry.handler"` now creates successfully (was 400 before).
- Found while applying GE DPS ship-builder-infra against Floci: 23 of 24 Lambdas used this handler form and failed CreateFunction; with the fix they resolve.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -1201,6 +1201,12 @@ public class LambdaService {
         if (fn.getRuntime().startsWith("python")) {
             return modulePath.replace('.', '/');
         }
+        // A file-based handler may be given with a leading "./" (e.g.
+        // "./v1/lambda-handlers/entry.handler"); deployment-package entries are stored without
+        // it, so normalize the prefix away before matching.
+        if (modulePath.startsWith("./")) {
+            modulePath = modulePath.substring(2);
+        }
         return modulePath;
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaIntegrationTest.java
@@ -279,6 +279,43 @@ class LambdaIntegrationTest {
         given().delete(BASE_PATH + "/functions/large-zip-fn");
     }
 
+    @Test
+    @Order(15)
+    void createFunctionWithDotSlashNestedHandler() throws Exception {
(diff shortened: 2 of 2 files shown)


hectorvent (approved):
Hi @abanna, 
Thank you!

It looks good to me. Action error not related to this PR and already resolved on main branch.

PR #1590, opened 2026-06-26
feat(appsync): Phase 5 — Management API Coverage + AWS Behavior Alignment + Async Schema Creation

AgustinBertagna:
## Summary

Implements Phase 5 of the AppSync implementation plan in #1173.

This phase combines two related scopes that close gaps with AWS AppSync behavior:

1. **Management API test coverage + AWS behavior alignment** — fills test gaps for already-implemented features and aligns wire behavior with the AWS AppSync API Reference and AWS SDK for Java v2 Javadoc.

2. **Async schema creation semantics** — `StartSchemaCreation` is asynchronous in AWS (returns `200 + PROCESSING`, transitions to `ACTIVE` or `FAILED` via polling). 43 operations return `ConcurrentModificationException 409` while a schema creation is `PROCESSING`.

## Type of change

- [x] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

### Management API coverage + AWS alignment

- **9 conflict / limit / 404 tests** for Create/Update/Delete on DataSource, Resolver, Function, Type, ChannelNamespace, DomainName, API keys.
- **Error code corrections** validated against AWS AppSync API Reference, AWS SDK for Java v2 Javadoc, and `service-2.json`:
  - `BadRequestException` 400 with `reason: "CODE_ERROR"` and `detail.codeErrors[]` for SDL errors (PARSER_ERROR / VALIDATION_ERROR)
  - `ConcurrentModificationException` 409 for duplicate Source API association (replaces `BadRequestException` 400)
  - `span: -1` default in `CodeErrorLocation`
  - DELETION_SCHEDULED associations can be replaced (no 409)
- **Removed Floci-only endpoints**: `getApiKey`, `listAllResolvers`.
- **Extended error format** added to `AwsException` (optional `extendedData` field) and `AwsExceptionMapper`.

### Async schema creation semantics

- **`StartSchemaCreation` is now async**: the request returns `200 + PROCESSING`. A new `SchemaCreationWorker` (singleton `ExecutorService` with configurable thread count) compiles the schema in a background thread, transitioning to `ACTIVE` or `FAILED` (with serialized extended error data in the polled `GetSchemaCreationStatus.details`).
- **43 operations return `ConcurrentModificationException 409`** while any schema creation is `PROCESSING`:
  - 27 per-API mutations (`createDataSource`, `getDataSource`, `updateDataSource`, `deleteDataSource`, `createResolver`, `getResolver`, `updateResolver`, `deleteResolver`, `createFunction`, `getFunction`, `updateFunction`, `deleteFunction`, `createType`, `getType`, `updateType`, `deleteType`, `listTypes`, `putEnvironmentVariables`, `createChannelNamespace`, `updateChannelNamespace`, `deleteChannelNamespace`, `updateGraphqlApi`, `deleteGraphqlApi`, `createSourceApiAssociation`, `createMergedApiAssociation`, `updateSourceApiAssociation`, `deleteSourceApiAssociation`, `deleteMergedApiAssociation`)
  - 2 account-wide operations (`createGraphqlApi`, `createApi`)
  - 1 self-blocking on second `StartSchemaCreation` on the same API
- **Orphan PROCESSING recovery**: `EmulatorLifecycle.onStart` calls `schemaCreationWorker.recoverOrphans()` after `storageFactory.loadAll()`, marking any leftover `PROCESSING` as `FAILED` (e.g., after an emulator crash mid-compilation).

### New configuration

| Env var | Default | Purpose |
|---|---|---|
| `FLOCI_SERVICES_APPSYNC_SCHEMA_WORKER_THREADS` | 4 | Worker pool size for schema compilation |
| `FLOCI_SERVICES_APPSYNC_SCHEMA_WORKER_SHUTDOWN_TIMEOUT_SECONDS` | 30 | Graceful shutdown wait for in-flight workers |

## Test Results

- `AppSyncIntegrationTest`: 154/154 pass
- `AppSyncTest` (SDK): 78/78 pass
- `EmulatorLifecycleTest`: 16/16 pass

## Validation sources

- [AppSync API Reference (Welcome)](https://docs.aws.amazon.com/appsync/latest/APIReference/Welcome.html)
- [`StartSchemaCreation`](https://docs.aws.amazon.com/appsync/latest/APIReference/API_StartSchemaCreation.html)
- [`AssociateSourceGraphqlApi`](https://docs.aws.amazon.com/appsync/latest/APIReference/API_AssociateSourceGraphqlApi.html)
- [`ConcurrentModificationException` Javadoc](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/appsync/model/ConcurrentModificationException.html)
- [`BadRequestException` Javadoc](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/appsync/model/BadRequestException.html)
- [`CodeError` / `CodeErrorLocation` Javadoc](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/appsync/model/CodeError.html)
- [`AppSyncClient` Javadoc (operation list)](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/appsync/AppSyncClient.html)
- [`service-2.json` (raw)](https://raw.githubusercontent.com/aws/aws-sdk-java-v2/master/services/appsync/src/main/resources/codegen-resources/service-2.json)

## Checklist

- [x] New or updated SDK compatibility test added
- [x] Protocol behavior validated with AWS SDK clients
- [x] Error codes validated against AWS AppSync API Reference, AWS SDK for Java v2 Javadoc, `service-2.json`
- [x] No custom endpoint shapes introduced

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
@@ -107,11 +107,11 @@ class AppSyncTest {
     @Test
     @Order(11)
     void getSchemaCreationStatus() {
-        // Schema creation is async in real AWS: poll until terminal (ACTIVE
+        // Schema creation is async in real AWS: poll until terminal (SUCCESS
         // or FAILED). All subsequent tests in this class depend on the
         // schema being ACTIVE because the 409 gate fires while PROCESSING.
         SchemaStatus status = pollSchemaStatusToTerminal();
-        assertThat(status).isEqualTo(SchemaStatus.ACTIVE);
+        assertThat(status).isEqualTo(SchemaStatus.SUCCESS);
     }
 
     // ── API Keys ────────────────────────────────────────────────────────
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
@@ -544,7 +544,7 @@ class AppSyncTest {
                         .apiId(apiId)
                         .definition(SdkBytes.fromUtf8String("type Query { invalid !!! }"))
                         .build()))
-                .isInstanceOf(AppSyncException.class)
+                .isInstanceOf(BadRequestException.class)
                 .hasMessageContaining("Invalid schema");
     }
 
@@ -600,9 +600,11 @@ class AppSyncTest {
                 UpdateChannelNamespaceRequest.builder()
                         .apiId(apiId)
                         .name(channelNsName)
+                        .codeHandlers("export function handler(ctx) { return ctx; }")
                         .build());
 
         assertThat(resp.channelNamespace()).isNotNull();
+        assertThat(resp.channelNamespace().codeHandlers()).contains("handler");
     }
 
     @Test
@@ -672,6 +674,7 @@ class AppSyncTest {
 
         assertThat(resp.domainNameConfig()).isNotNull();
         assertThat(resp.domainNameConfig().domainName()).isEqualTo(tempDomainName);
+        assertThat(resp.domainNameConfig().description()).isEqualTo("updated-domain");
     }
 
     @Test
@@ -949,31 +952,34 @@ class AppSyncTest {
 
     @Test
     @Order(80)
-    void getNonExistentApiThrowsException() {
+    void getNonExistentApiThrowsNotFoundException() {
         assertThatThrownBy(() -> client.getGraphqlApi(GetGraphqlApiRequest.builder()
                         .apiId("nonexistent12345678901234")
                         .build()))
-                .isInstanceOf(AppSyncException.class);
+                .isInstanceOf(NotFoundException.class)
+                .hasMessageContaining("GraphQL API not found:");
     }
 
     @Test
     @Order(81)
-    void getNonExistentDataSourceThrowsException() {
+    void getNonExistentDataSourceThrowsNotFoundException() {
         assertThatThrownBy(() -> client.getDataSource(GetDataSourceRequest.builder()
                         .apiId(apiId)
                         .name("nonexistent-ds")
                         .build()))
-                .isInstanceOf(AppSyncException.class);
+                .isInstanceOf(NotFoundException.class)
+                .hasMessageContaining("Data source not found:");
     }
 
     @Test
     @Order(82)
-    void getNonExistentTypeThrowsException() {
+    void getNonExistentTypeThrowsNotFoundException() {
         assertThatThrownBy(() -> client.getType(GetTypeRequest.builder()
                         .apiId(apiId)
                         .typeName("NonExistentType")
                         .build()))
-                .isInstanceOf(AppSyncException.class);
+                .isInstanceOf(NotFoundException.class)
+                .hasMessageContaining("Type not found:");
     }
 
     @Test
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
@@ -107,12 +107,11 @@ class AppSyncTest {
     @Test
     @Order(11)
     void getSchemaCreationStatus() {
-        GetSchemaCreationStatusResponse resp = client.getSchemaCreationStatus(
-                GetSchemaCreationStatusRequest.builder()
-                        .apiId(apiId)
-                        .build());
-
-        assertThat(resp.statusAsString()).isEqualTo("ACTIVE");
+        // Schema creation is async in real AWS: poll until terminal (ACTIVE
+        // or FAILED). All subsequent tests in this class depend on the
+        // schema being ACTIVE because the 409 gate fires while PROCESSING.
+        SchemaStatus status = pollSchemaStatusToTerminal();
+        assertThat(status).isEqualTo(SchemaStatus.ACTIVE);
     }
 
     // ── API Keys ────────────────────────────────────────────────────────
@@ -541,12 +540,28 @@ class AppSyncTest {
     @Test
     @Order(93)
     void startSchemaCreation_invalidSchema() {
-        assertThatThrownBy(() -> client.startSchemaCreation(StartSchemaCreationRequest.builder()
-                        .apiId(apiId)
-                        .definition(SdkBytes.fromUtf8String("type Query { invalid !!! }"))
-                        .build()))
-                .isInstanceOf(BadRequestException.class)
-                .hasMessageContaining("Invalid schema");
+        StartSchemaCreationResponse resp = client.startSchemaCreation(StartSchemaCreationRequest.builder()
+                .apiId(apiId)
+                .definition(SdkBytes.fromUtf8String("type Query { invalid !!! }"))
+                .build());
+        assertThat(resp.status()).isEqualTo(SchemaStatus.PROCESSING);
+        SchemaStatus terminal = pollSchemaStatusToTerminal();
+        assertThat(terminal).isEqualTo(SchemaStatus.FAILED);
+    }
+
+    private SchemaStatus pollSchemaStatusToTerminal() {
+        Instant deadline = Instant.now().plus(Duration.ofSeconds(10));
+        SchemaStatus s = client.getSchemaCreationStatus(r -> r.apiId(apiId)).status();
+        while (s == SchemaStatus.PROCESSING && Instant.now().isBefore(deadline)) {
+            try {
+                Thread.sleep(50);
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(ie);
+            }
+            s = client.getSchemaCreationStatus(r -> r.apiId(apiId)).status();
+        }
+        return s;
     }
 
     @Test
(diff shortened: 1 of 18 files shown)


AgustinBertagna:
## Fixed

1. **P1: Duplicate `awaitility` in `pom.xml`** — Removed the pre-existing version-less declaration (BOM-managed). The pinned `4.2.2` declaration remains.
2. **P1: `deleteChannelNamespace` missing PROCESSING gate** — Added `assertSchemaNotBusy(apiId)` to `getChannelNamespace`, which protects both `deleteChannelNamespace` and `getChannelNamespace` (same pattern as `getResolver`/`getFunction`/`getType`).
3. **P2: Config keys missing from `application.yml`** — Added the `appsync` section under `storage.services` in both main and test YAML with `flush-interval-ms`, `schema-worker-threads`, and `schema-worker-shutdown-timeout-seconds`.
4. **P2: TOCTOU race in `startSchemaCreation`** — Wrapped the check-and-set in a `synchronized(this)` block (the service is `@ApplicationScoped` singleton). The async `submitSchemaCreation` call stays outside the lock to avoid blocking the worker thread.

## Additional gates found during validation

While validating all 43 `ConcurrentModificationException` operations from `service-2.json` against the code, I found 4 more operations that were missing the gate (besides the one Greptile caught). All are now fixed and covered by tests (Orders 712-716):

- `DeleteDomainName` → now resolves the associated `apiId` and gates on it
- `DisassociateApi` → same pattern
- `DisassociateMergedGraphqlApi` → gate on `sourceApiIdentifier`
- `UpdateDomainName` → resolves the associated `apiId` and gates on it

## Not a bug (Greptile was incorrect)

> `createChannelNamespace` duplicate check still throws `ConflictException`

This is **intentional and correct**. `CreateChannelNamespace` is the **only** AppSync operation that declares `ConflictException` 409 in `service-2.json`. No other `Create*` operation has it — `CreateDataSource`, `CreateResolver`, `CreateFunction`, `CreateType` all have only `BadRequestException` + `ConcurrentModificationException`.

The AWS AppSync API Reference confirms this:
- [`CreateChannelNamespace` errors](https://docs.aws.amazon.com/appsync/latest/APIReference/API_CreateChannelNamespace.html#API_CreateChannelNamespace_Errors): `AccessDeniedException`, `BadRequestException`, `ConcurrentModificationException`, **`ConflictException`** (409), `InternalFailureException`, `NotFoundException`, `ServiceQuotaExceededException`, `UnauthorizedException`
- [`CreateDataSource` errors](https://docs.aws.amazon.com/appsync/latest/APIReference/API_CreateDataSource.html#API_CreateDataSource_Errors): `AccessDeniedException`, `BadRequestException`, `ConcurrentModificationException`, `InternalFailureException`, `NotFoundException` (no `ConflictException`)

The `ConflictException` 409 for duplicate channel namespace names is the AWS contract, not an oversight.

## Test results

438/438 tests pass (159 integration + 93 VTL engine + 91 util + 17 lifecycle + 78 SDK).

hectorvent:
Really solid piece of work, and the error corrections are well grounded. I checked them against the AppSync [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/appsync/2017-07-25/service-2.json) and they all hold up: `ConflictException` is not declared on those CRUD ops so moving to `BadRequestException` 400 is right, `ApiKeyLimitExceededException` really is a 400 in the model, the duplicate source association `ConcurrentModificationException` 409 is a declared error, and the extended `BadRequestException` shape matches exactly down to `span: -1` being the documented unknown default. The `environmentVariables` path casing fix and dropping the non AWS `getApiKey` / bare `listResolvers` endpoints are correct too. I even double checked the read ops you gate with 409 and, pleasantly, AWS does declare `ConcurrentModificationException` on `GetDataSource` / `GetType` / `ListTypes`, so that gating is faithful.

One small inconsistency: `getChannelNamespace` now gets `assertSchemaNotBusy`, so it can return `ConcurrentModificationException` 409 during a PROCESSING window, but that operation does not declare CME (only `BadRequestException`, `NotFoundException`, `UnauthorizedException`, `InternalFailureException`, `AccessDeniedException`). So an SDK client would get a code it can't map for that op. The create/update/delete channel namespace ops all declare CME, so it is just the getter. Simplest fix is to not gate the read:

```java
public ChannelNamespace getChannelNamespace(String apiId, String name) {
    // no assertSchemaNotBusy: GetChannelNamespace does not declare ConcurrentModificationException
    return channelNamespaceStore.get(apiKey(apiId, name))
            .orElseThrow(() -> new AwsException("NotFoundException",
                    "Channel namespace not found: " + name, 404));
}
```

Tiny impact in practice since the PROCESSING window is short, just the one spot where a gated op can return an undeclared code. Everything else looks great. Verified against `service-2.json`.

AgustinBertagna:
**@hectorvent** Thanks for the review. You're right — `GetChannelNamespace` doesn't declare `ConcurrentModificationException`. I validated it against the `AppSyncClient` in SDK v2 (`appsync-2.42.24.jar`) and it only throws `BadRequestException`, `NotFoundException`, `UnauthorizedException`, `InternalFailureException`, and `AccessDeniedException`.

The fix applied:
- Removed `assertSchemaNotBusy` from `getChannelNamespace`
- Added `assertSchemaNotBusy` to `deleteChannelNamespace`, which does declare CME

All 159 integration tests and 78 SDK tests pass.

hectorvent:
Solid phase. The gating list is well researched: exactly 43 of the 74 AppSync operations declare `ConcurrentModificationException` in botocore's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/appsync/2017-07-25/service-2.json), matching your number, and `GetSchemaCreationStatus` correctly stays ungated so the polling loop can't deadlock. `AppSyncIntegrationTest` is green on the branch, 159/159. That green is why the first item is easy to miss.

**Schema creation never reaches a terminal state for a non-default account.** [`startSchemaCreation`](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/AppSyncService.java#L207) writes `PROCESSING` on the request thread, [`markStatus`](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/graphql/SchemaCreationWorker.java#L98) writes `ACTIVE` on a pool thread. Both go through `AccountAwareStorageBackend`, which reads the key prefix from a `@RequestScoped` bean:

```java
try {
    String accountId = requestContextInstance.get().getAccountId();
    if (accountId != null) return accountId;
} catch (ContextNotActiveException ignored) { }   // worker thread lands here
return defaultAccountId;
```

The worker's pool uses raw `Thread`s, so nothing propagates the context. A caller on account `111122223333` stores `PROCESSING` at `111122223333/<apiId>`, the worker stores `ACTIVE` at `000000000000/<apiId>`, and every guarded operation on that API 409s forever. [`recoverOrphans`](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/graphql/SchemaCreationWorker.java#L121) has the same problem. The decorator's Javadoc prescribes the fix, and EventBridge, OpenSearch, Scheduler and BcmDataExports already follow it: store the account ID on `SchemaCreationStatus` and use the `*ForAccount` overloads. All tests run under the default account, so the suite can't see this.

**`CreateGraphqlApi` and `CreateApi` block account wide.** [`assertNoSchemaBusyAnywhere`](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/AppSyncService.java#L249) walks every key in the status store, so compiling one API 409s the creation of an unrelated one. `CreateApi` is an Event API operation with no GraphQL schema at all. The model declaring the error on `CreateGraphqlApi` makes the code legal, but says nothing about scope.

**Three ops in the description aren't guarded.** [`deleteResolver`](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/AppSyncService.java#L432), [`deleteFunction`](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/AppSyncService.java#L488), [`deleteType`](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/AppSyncService.java#L542) have no `assertSchemaNotBusy` call.

Smaller ones:

`startSchemaCreation` writes the SDL into `schemaStore` [before the worker validates it](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/AppSyncService.java#L211), so a `FAILED` compile leaves `GetIntrospectionSchema` serving broken SDL while the registry keeps the old schema. Move the put into the success path.

`GetSchemaCreationStatus` returns `ACTIVE` where AWS returns `SUCCESS`. The model doc is explicit: "The current state of the schema (PROCESSING, FAILED, SUCCESS, or NOT_APPLICABLE)." A client polling for `SUCCESS` hangs. Predates the PR, but `AppSyncTest.java:114` now pins it.

The new knobs sit in `AppSyncStorageConfig`, so the real env vars are `FLOCI_STORAGE_SERVICES_APPSYNC_*`, not the `FLOCI_SERVICES_APPSYNC_*` in the [Javadoc](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java#L329) and PR table. `AppSyncServiceConfig` already exists and a worker pool size isn't a storage concern. Also `pom.xml` pins `awaitility` `4.2.2` over the Quarkus BOM, and [`SchemaRegistry`](https://github.com/floci-io/floci/blob/58a01fe6164296c467337959f6cc337e077b4b52/src/main/java/io/github/hectorvent/floci/services/appsync/graphql/SchemaRegistry.java#L18) and `SchemaCreationWorker` inject each other just to forward one call.

Verified correct: the `ConflictException` to `BadRequestException` 400 conversions (none of those four ops declare `ConflictException`, and you correctly left `CreateChannelNamespace` alone), `ApiKeyLimitExceededException` at 400, the extended `BadRequestException` body against `BadRequestDetail`, `span: -1`, `200 + PROCESSING` on `StartSchemaCreation`, and dropping `getApiKey` and `listAllResolvers`.

And sorry for not catching the `ACTIVE` vs `SUCCESS` status in an earlier phase. It has been there since the schema work started and I should have flagged it before it reached this PR.

AgustinBertagna:
Thanks for the thorough review. All 7 findings addressed:

1. **getChannelNamespace CME gate** — Removed. getChannelNamespace is a read operation, no busy gate needed. The 409 gate belongs on deleteChannelNamespace (already added in the prior push).

2. **Account-aware schema status** — Added `accountId` field to `SchemaCreationStatus`. Worker uses `AccountAwareStorageBackend` (putForAccount/getForAccount). Status scoped to the requesting account so concurrent accounts don't interfere.

3. **Worker writes schemaStore directly** — Removed `schemaStore.put` from `startSchemaCreation`. The worker now puts the compiled SDL into the store on success only.

4. **409 gates on delete operations** — Added `assertSchemaNotBusy` calls to `deleteResolver`, `deleteFunction`, and `deleteType`.

5. **assertNoSchemaBusyAnywhere scoped** — Now iterates `keysForAccount(accountId)` instead of all keys.

6. **Terminal status ACTIVE→SUCCESS** — Changed to SUCCESS to match AWS behavior. Updated worker, integration tests, and SDK compatibility test.

7. **Circular dependency removed** — `SchemaRegistry` no longer depends on `SchemaCreationWorker`. Removed `submitSchemaCreation` method. `AppSyncService` calls the worker directly.

8. **Worker config relocated** — Moved `schemaWorkerThreads` and `schemaWorkerShutdownTimeoutSeconds` from `AppSyncStorageConfig` to `AppSyncServiceConfig` (env vars unchanged).

Tests: 159 integration + 78 SDK compatibility = 237 passing.

AgustinBertagna:
Good catch on the delete gates — `deleteResolver`, `deleteFunction`, and `deleteType` were missing `assertSchemaNotBusy`. Fixed in the latest push.

hectorvent (approved):
Thanks @AgustinBertagna

PR #1593, opened 2026-06-26
feat(apigateway): support floci:override-id for api gateways

Marcel2603:
## Summary

This PR adds support for specifying a custom resource ID for API Gateway V1 (REST API) and API Gateway V2 (HTTP/WebSocket API) using the standard `floci:override-id` tag. 
This unifies the custom ID configuration behavior across services like KMS and Cognito, enabling consistent API Gateway endpoints/URLs across different local environments and machines. 
Additionally, the old/undocumented `_custom_id_` tag key has been marked as deprecated for API Gateway and is now treated as a reserved tag (meaning it is stripped on creation and rejected on update), though it remains supported for backwards-compatible ID overrides on resource creation.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- **Previous behavior:** API Gateway V1 used a non-standard `_custom_id_` tag (which remained in the response payload tags), and API Gateway V2 did not support custom IDs. Other services (Cognito, KMS) used `floci:override-id`.
- **New behavior:** 
  - API Gateway V1 and V2 both accept `floci:override-id` during creation to override the auto-generated API ID.
  - The standard `floci:override-id` tag is correctly stripped from the returned resource tags.
  - The deprecated `_custom_id_` tag key remains supported for backwards compatibility, but is now treated as a reserved tag (meaning it is stripped from returned resource tags and rejected on resource updates).
     
## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
@@ -14,6 +14,7 @@ import com.fasterxml.jackson.core.type.TypeReference;
 
 import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.core.common.ReservedTags;
 import io.github.hectorvent.floci.core.storage.StorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageFactory;
 import io.github.hectorvent.floci.services.apigateway.model.Account;
@@ -172,7 +173,7 @@ public class ApiGatewayService {
         Map<String, String> tags = request.get("tags") instanceof Map<?, ?> m
                 ? (Map<String, String>) m : new HashMap<>();
 
-        String customId = tags.get("_custom_id_");
+        String customId = ReservedTags.extractOverrideId(tags);
         String apiId = (customId != null && !customId.isBlank()) ? customId : shortId(10);
 
         RestApi api = new RestApi();
@@ -180,7 +181,7 @@ public class ApiGatewayService {
         api.setName(name);
         api.setDescription(description);
         api.setCreatedDate(System.currentTimeMillis() / 1000L);
-        api.setTags(tags);
+        api.setTags(ReservedTags.stripReservedTags(tags));
 
         EndpointConfiguration endpointConfiguration = new EndpointConfiguration();
         if (request.get(EPC_KEY) instanceof Map<?, ?> epMap) {
--- a/docs/services/api-gateway.md
+++ b/docs/services/api-gateway.md
@@ -2,6 +2,13 @@
 
 Floci supports both API Gateway v1 (REST APIs) and API Gateway v2 (HTTP APIs).
 
+## Custom Resource IDs
+
+When creating APIs in Floci, you can optionally pin the generated `ApiId` (or `id`) by passing a reserved tag `floci:override-id`. This is particularly useful in testing environments where you want predictable endpoint URLs. Floci uses the tag value as the created API ID and automatically strips it from the stored and returned resource tags so it is never persisted as user-visible metadata.
+
+> [!WARNING]
+> The older `_custom_id_` tag is also supported for backwards compatibility, but is considered deprecated. Unlike `floci:override-id`, the `_custom_id_` tag is not stripped and will remain attached to the API resource.
+
 ## API Gateway v1 (REST APIs) {#v1}
 
 **Protocol:** REST JSON
--- a/src/main/java/io/github/hectorvent/floci/core/common/ReservedTags.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/ReservedTags.java
@@ -10,6 +10,7 @@ public final class ReservedTags {
 
     public static final String RESERVED_PREFIX = "floci:";
     public static final String OVERRIDE_ID_KEY = RESERVED_PREFIX + "override-id";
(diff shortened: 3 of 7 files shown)


hectorvent:
Thanks for this, unifying the API Gateway custom IDs behind `floci:override-id` is exactly the direction we want, and the test coverage is nice.

Heads up before you invest more here: the branch conflicts with main. #1486 reworked [ReservedTags.java](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/core/common/ReservedTags.java) and `extractOverrideId` no longer exists. Each service now gets its own extractor that runs the shared validation (blank, whitespace, `/?#`, control chars) with a service appropriate error code. For the rebase I'd suggest following that pattern instead of restoring the old raw lookup, something like:

```java
public static String extractOverrideApiId(Map<String, String> tags) {
    String id = getOverride(tags, OVERRIDE_ID_KEY, ReservedTags::validateOverrideId,
            "API ID", "BadRequestException");
    return id != null ? id : getOverride(tags, DEPRECATED_CUSTOM_ID_KEY,
            ReservedTags::validateOverrideId, "API ID", "BadRequestException");
}
```

`BadRequestException` because that's what API Gateway declares (verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/apigateway/2015-07-09/service-2.json), it has no `ValidationException`). Validation matters here since an unvalidated override with a `/` would land in the `/restapis/{apiId}` route space.

Two smaller things while you're in there. Keeping the `_custom_id_` fallback inside the API Gateway extractor (not in the global `isReserved`) avoids a surprise where a KMS or Cognito user with a legitimate tag literally named `_custom_id_` suddenly gets a 400 from `rejectUnknownReservedTags`, since that key is a legal tag name on real AWS. And the PR body says `_custom_id_` is rejected on updates, but neither V1 nor V2 tag endpoints call `rejectReservedTagsOnUpdate`, so either wire that in or trim the claim.

The doc warning Greptile flagged still says `_custom_id_` is not stripped while the code strips it, worth fixing in the same pass. Happy to take another look after the rebase!

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
```diff
@@ -172,15 +173,15 @@ public RestApi createRestApi(String region, Map<String, Object> request) {
         Map<String, String> tags = request.get("tags") instanceof Map<?, ?> m
                 ? (Map<String, String>) m : new HashMap<>();
 
-        String customId = tags.get("_custom_id_");
+        String customId = ReservedTags.extractOverrideId(tags);
         String apiId = (customId != null && !customId.isBlank()) ? customId : shortId(10);
```
abanna, line 177: Thanks for unifying the override tag. Could we check for an existing API before accepting this ID? A second create with the same ID currently replaces the API record (and leaves the old resources behind for v1). A collision test for both v1 and v2 would make this safer.

hectorvent:
Thank you for this, and sorry it sat for so long. Unifying API Gateway custom IDs behind
`floci:override-id` was the right call and the change is now on `main`'s doorstep in #2045, with your
commit carried as a co-author.

I picked it up rather than asking you to rebase because the conflict was not textual. #1486 landed
after you opened this and replaced the central `extractOverrideId` with per-service extractors, so the
main hunk needed rewriting against a helper that did not exist when you wrote it. That is our churn,
not yours.

What carried over unchanged: the v1 and v2 override support, the `_custom_id_` deprecation, and the
shape of your tests.

What changed, in case you are curious:

- Routing through `ReservedTags.getOverride` means the value is now validated rather than used
  verbatim, so an override containing `/`, `?` or `#` is rejected instead of producing an API at an
  unreachable URL.
- The error code is `BadRequestException`, which `CreateRestApi`, `CreateApi` and `TagResource` all
  declare. The shared reject helper uses `ValidationException`, which neither service declares, so API
  Gateway got its own guard.
- `_custom_id_` is reserved for API Gateway only. Extending `isReserved` globally would also have
  stripped it from KMS keys and Cognito pools, which was wider than intended.

Closing in favour of #2045. Thanks again, and please do send more.

Marcel2603:
> Thank you for this, and sorry it sat for so long. Unifying API Gateway custom IDs behind `floci:override-id` was the right call and the change is now on `main`'s doorstep in #2045, with your commit carried as a co-author.
> 
> I picked it up rather than asking you to rebase because the conflict was not textual. #1486 landed after you opened this and replaced the central `extractOverrideId` with per-service extractors, so the main hunk needed rewriting against a helper that did not exist when you wrote it. That is our churn, not yours.
> 
> What carried over unchanged: the v1 and v2 override support, the `_custom_id_` deprecation, and the shape of your tests.
> 
> What changed, in case you are curious:
> 
> * Routing through `ReservedTags.getOverride` means the value is now validated rather than used
>   verbatim, so an override containing `/`, `?` or `#` is rejected instead of producing an API at an
>   unreachable URL.
> * The error code is `BadRequestException`, which `CreateRestApi`, `CreateApi` and `TagResource` all
>   declare. The shared reject helper uses `ValidationException`, which neither service declares, so API
>   Gateway got its own guard.
> * `_custom_id_` is reserved for API Gateway only. Extending `isReserved` globally would also have
>   stripped it from KMS keys and Cognito pools, which was wider than intended.
> 
> Closing in favour of #2045. Thanks again, and please do send more.

thanks, sorry are was very busy :( Will try to be more active in the future

PR #1594, opened 2026-06-26
fix(ses): treat missing SendingEnabled as false in v2 PutConfigurationSetSendingOptions

okinaka:
## Summary

The v2 `PutConfigurationSetSendingOptions` required `SendingEnabled` and returned `400` when it was absent, which is stricter than AWS. Real AWS accepts an empty body, treating an absent `SendingEnabled` as `false` and returning `200` — the same way the already-AWS-aligned `PutConfigurationSetReputationOptions` handles a missing `ReputationMetricsEnabled`.

- `SesController.putConfigurationSetSendingOptions`: parse the body via the shared `readOptionBody` helper and reuse the AWS-aligned `parseSendingEnabled` deserialization already used by `CreateConfigurationSet`, instead of requiring the field.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Each body shape was verified against real AWS (raw signed requests, us-east-1); the "after" column matches AWS exactly by reusing the same `parseSendingEnabled` deserialization as `CreateConfigurationSet`:

| Body | Before this PR | After this PR |
| --- | --- | --- |
| empty body | `400 BadRequestException` | `200` (SendingEnabled→false) |
| `{}` | `400 BadRequestException` | `200` (false) |
| `{"SendingEnabled": "yes"}` (string) | `400 BadRequestException` | `200`, stored as `true` |
| `{"SendingEnabled": null}` | `400 BadRequestException` | `400 SerializationException` |
| `{"SendingEnabled": true/false}` | `200` | `200` (unchanged) |

This is a v2-only change — the v1 Query `UpdateConfigurationSetSendingEnabled` keeps `Enabled` required, matching the v1 contract.

**Reproduce** with the AWS CLI against a running Floci (`./mvnw quarkus:dev`). Before this PR this failed with `400 SendingEnabled must be present and must be a boolean.`:

```bash

aws sesv2 create-configuration-set --configuration-set-name cs

# Omit SendingEnabled -> 200 (empty body accepted)
aws sesv2 put-configuration-set-sending-options --configuration-set-name cs

# Reads back disabled
aws sesv2 get-configuration-set --configuration-set-name cs --query SendingOptions
# -> { "SendingEnabled": false }
```

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesController.java
@@ -768,23 +768,15 @@ public class SesController {
                                                        String body) {
         String region = regionResolver.resolveRegion(headers);
         try {
-            JsonNode request = (body == null || body.isBlank())
-                    ? objectMapper.createObjectNode()
-                    : objectMapper.readTree(body);
-            requireJsonObject(request);
-            JsonNode enabledNode = request.path("SendingEnabled");
-            if (enabledNode.isMissingNode() || enabledNode.isNull() || !enabledNode.isBoolean()) {
-                throw new AwsException("BadRequestException",
-                        "SendingEnabled must be present and must be a boolean.", 400);
-            }
-            sesService.setConfigurationSetSendingEnabled(name, enabledNode.booleanValue(), region);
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thanks @okinaka

PR #1595, opened 2026-06-26
feat(tls): serve HTTPS on 443 and trust Floci's cert in Lambda containers

abanna:
## Summary

CDK / CloudFormation **custom resources** (e.g. `Custom::LogRetention`, the CDK provider framework, `AwsCustomResource`) send their result to the `cfn-response` ResponseURL using `aws-cdk-lib`'s bundled response code. That bundled code parses only the URL's **hostname and path**, then calls `require("https").request(...)` — it ignores the scheme and port. So even though Floci hands the resource an `http://<host>:4566` ResponseURL, the callback is always sent to `https://<host>:443`. Against Floci that fails (nothing on 443, or an untrusted cert once 443 is served), so **every custom resource rolls back** — even though the resource's own AWS SDK calls (on 4566) succeed. In real AWS this works because the ResponseURL is a pre-signed S3 HTTPS URL on 443.

This makes Floci reachable — and trusted — where the callback actually connects, without touching the CDK-owned handler. Everything is gated on `floci.tls.enabled` (default `false`), so non-TLS setups are unchanged.

1. **`TlsProxyServer` also binds 443.** The same first-byte protocol-detecting handler now listens on a configurable `floci.tls.aws-https-port` (default `443`, `0` to disable, de-duped against the public port) in addition to the public Floci port. The first byte still selects the HTTP (4510) or HTTPS (4511) backend, so 443 serves both like the main port.
2. **`CertificateGenerator.generateSelfSignedCertificate()`** — a genuinely self-signed cert (issuer == subject, `CA:true`, `keyCertSign`) usable as a **trust anchor**. The existing `generateCertificate()` is unchanged (it still stamps the cosmetic Amazon issuer DN for ACM realism, which cannot be verified as an anchor). `TlsConfigSource` uses the self-signed variant for Floci's HTTPS cert and regenerates a legacy non-self-signed cert on upgrade.
3. **`ContainerLauncher`** copies Floci's CA cert into each Lambda container at `/etc/floci-ca.crt` and sets `NODE_EXTRA_CA_CERTS`, `AWS_CA_BUNDLE`, `SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE` so the container trusts Floci's HTTPS endpoint. `NODE_EXTRA_CA_CERTS` *adds* to Node's built-in CAs, so public TLS still works.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior before this change: with `floci.tls.enabled=true`, a CloudFormation/CDK custom resource's `cfn-response` callback (hardcoded `https://`, no port) could not reach Floci, so custom resources failed and their stacks rolled back. In real AWS the callback is a pre-signed S3 HTTPS URL on 443.

Verified with `aws-cdk` 2.155 (`cdklocal`) and AWS CLI 2.x:
- Proxy binds both `4566` and `443`; a direct HTTPS request to `:443` returns `200`.
- A Lambda issuing `https://<host>/...` with no explicit port (exactly the `cfn-response` pattern) now connects and the cert verifies — previously `ECONNREFUSED` / `UNABLE_TO_VERIFY_LEAF_SIGNATURE`.
- A real CDK app deployed via `cdklocal` reaches `CREATE_COMPLETE` with **all** resources, including the `Custom::LogRetention` that previously rolled back.

## Checklist

- [x] Tests pass locally — ran the affected/related classes in a JDK 25 container (`TlsProxyServerTest`, `CertificateGeneratorSelfSignedTest`, `ContainerLauncherCaTrustTest`, plus existing `TlsConfigSource*` and `CertificateGenerator*`): 93 passed, 0 failures. Full suite runs in CI.
- [x] New tests added — `TlsProxyServerTest` (dual-port bind + first-byte routing + `aws-https-port=0`/dedupe), `CertificateGeneratorSelfSignedTest` (self-signed cert is its own issuer/CA and self-verifies; ACM-style cert keeps the cosmetic issuer and is not a CA), `ContainerLauncherCaTrustTest` (CA path resolution + injected env vars). Behavior also verified end-to-end live (above).


hectorvent (approved):
Thanks @abanna, 

This is a great addition to Floci.

PR #1596, opened 2026-06-26
feat(ec2): add instance metadata catalog and architecture parity

jvanzyl:
## Summary
- add an EC2 instance type metadata catalog resource for exact DescribeInstanceTypes/Offerings behavior
- expose processor architecture details and enforce AMI/instance architecture compatibility
- cover representative Graviton large instance types with exact arm64 metadata

## Verification
- ./mvnw -q -Dtest='Ec2InstanceTypeCatalogTest,Ec2ServiceTest,Ec2IntegrationTest' test

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2Tests.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2Tests.java
@@ -159,6 +159,19 @@ class Ec2Tests {
     }
 
     @Test
+    @Order(7)
+    @DisplayName("DescribeInstanceTypes - processor architectures")
+    void describeInstanceTypeProcessorArchitectures() {
+        DescribeInstanceTypesResponse resp = ec2.describeInstanceTypes(DescribeInstanceTypesRequest.builder()
+                .instanceTypes(InstanceType.fromValue("m6gd.2xlarge"))
+                .build());
+
+        assertThat(resp.instanceTypes()).hasSize(1);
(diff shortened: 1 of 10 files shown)


hectorvent:
Reviewed against botocore's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/ec2/2016-11-15/service-2.json) and real `DescribeInstanceTypes` data. This quietly fixes a real wire bug: on main the architectures are double wrapped and not under `processorInfo`, so a real SDK parses `processorInfo()` as null. The new shape matches the model exactly, the catalog numbers (59/118/474 GB, vCPU, memory) all check out, and the SDK compat test is exactly the right proof. Worth mentioning the shape fix in the PR description, it's the most valuable part.

One small suggestion: the mismatch message is inverted vs real AWS, which phrases it instance type first ([source](https://repost.aws/knowledge-center/ec2-instance-types-disabled-at-launch)):

```
The architecture 'x86_64' of the specified instance type does not match
the architecture 'arm64' of the specified AMI. Specify an instance type
and an AMI that have matching architectures, and try again.
```

The `InvalidParameterValue` code is right, so this is cosmetic, but matching the text is cheap while you're here. Also tiny: the YAML field `localStorageGiB` holds GB values (model member is `TotalSizeInGB`), renaming it to `localStorageGB` avoids someone "fixing" the numbers with a GiB conversion later.

hectorvent (approved):
Thanks @jvanzyl

PR #1597, opened 2026-06-26
fix(ec2): align AMI guest and launch template parity

jvanzyl:
## Summary
- preserve launch template instance profile data used by ASG-launched instances
- support registry override for container-backed guest images
- align launch template AMI resolution and skip networkd wait-online in generated AMI guests

## Verification
- ./mvnw -q -Dtest='Ec2ServiceTest,Ec2IntegrationTest,AmiImageToolTest,ContainerBuilderTest' test


hectorvent (approved):
Thanks @jvanzyl, 
solid implementation

PR #1598, opened 2026-06-26
fix(autoscaling): align active instance refresh behavior

jvanzyl:
## Summary
- reject Auto Scaling group config updates while instance refresh is active
- align active refresh failure shape with AWS SDK expectations
- keep the change focused on active refresh control-plane parity

## Verification
- ./mvnw -q -Dtest='AutoScalingServiceTest,AutoScalingQueryHandlerTest' test

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AutoScalingTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AutoScalingTest.java
@@ -30,7 +30,7 @@ class AutoScalingTest {
             "Valid requests must contain either the InstanceID parameter "
                     + "or both the ImageId and InstanceType parameters.";
     private static final String ACTIVE_INSTANCE_REFRESH_DESIRED_CONFIGURATION_MESSAGE =
-            "You cannot update the launch template, launch configuration, or mixed instances policy while an instance refresh with a desired configuration is active.";
+            "An active instance refresh with a desired configuration exists. All configuration options derived from the desired configuration are not available for update while the instance refresh is active.";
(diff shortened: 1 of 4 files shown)


hectorvent:
Thanks for this, the guard itself matches real AWS behavior and the scoping (only launch source changes, only refreshes with a desired configuration) is exactly right.

One thing to fix before merging: the error message. AWS documents the exact string for this case in the [instance refresh limitations](https://docs.aws.amazon.com/autoscaling/ec2/userguide/instance-refresh-overview.html):

```
An active instance refresh with a desired configuration exists. All configuration options derived from the desired configuration are not available for update while the instance refresh is active.
```

The PR uses a different wording, and since the new compat test asserts it with `isEqualTo`, we'd be codifying a message that diverges from the real service. Could you swap the constant in `AutoScalingService.java` and the assertions to the documented string?

Everything else checks out against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/autoscaling/2011-01-01/service-2.json) (autoscaling/2011-01-01): `ValidationError` with 400 is the right code here (AWS calls this a validation error, and it follows the same pattern we already use), and the Query param nesting in the tests matches the model.

Small non blocker: the "LaunchTemplateVersion requires a LaunchTemplateId or LaunchTemplateName." message is an approximation since AWS doesn't document that one. Fine as is, just flagging it. Also worth double checking that `isActiveRefreshStatus` covers `Cancelling` and `RollbackInProgress`, since AWS keeps the lock during those phases too.

hectorvent (approved):
Thanks @jvanzyl

PR #1599, opened 2026-06-26
fix(rds): improve postgres proxy startup parity

jvanzyl:
## Summary
- close postgres proxy bridge relays cleanly on exit
- decline standalone Postgres SSL negotiation with the expected protocol response
- support Postgres SSL startup before forwarding startup/auth messages to the backend

## Verification
- ./mvnw -Dtest=PostgresProtocolHandlerTest,RdsQueryHandlerTest,RdsServiceTest test

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/proxy/PostgresProtocolHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/proxy/PostgresProtocolHandler.java
@@ -1,22 +1,44 @@
 package io.github.hectorvent.floci.services.rds.proxy;
 
 import org.jboss.logging.Logger;
+import org.bouncycastle.asn1.x500.X500Name;
+import org.bouncycastle.asn1.x509.Extension;
+import org.bouncycastle.asn1.x509.GeneralName;
+import org.bouncycastle.asn1.x509.GeneralNames;
+import org.bouncycastle.cert.X509v3CertificateBuilder;
+import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
+import org.bouncycastle.cert.jcajce.JcaX509v3CertificateBuilder;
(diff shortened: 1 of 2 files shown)


hectorvent:
Thanks for this one, the region scoping and the bridge close fix are both solid steps toward parity. I verified the touched error codes and the lowercase `default` subnet group handling against `service-2.json` and they all match AWS. Two things worth addressing before merge:

**1. Pre-upgrade persisted data gets stranded under the old keys.** Records persisted before this change live under bare `id` keys. `findInstance` finds them via the fallback, but delete and modify write only to the new `region|id` key, so a delete leaves the instance visible forever and a modify duplicates it:

```java
// stored as "legacy-db" (pre-upgrade), delete removes "us-east-1|legacy-db"
rdsService.deleteDbInstance("legacy-db");
rdsService.listDbInstances(null).size(); // still 1
```

Re-keying legacy entries in `restorePersistedRuntime`, or deleting/putting with the key the record was found under, would cover it. Same pattern in `deleteDbCluster` and `deleteDbSubnetGroup`.

**2. CloudFormation and the Data API still use the default region overloads.** A stack created in `us-west-2` provisions its DB under the default region, and now that describes are region filtered, `describe-db-instances --region us-west-2` no longer returns it. Likewise `RdsDataResourceResolver` lists only default region resources, so a cluster created with `--region us-west-2` stops resolving by ARN. The resolver already has the region inside the ARN it receives, so parsing it from there should be enough.

Two small non blocking notes: other services (SSM, Kinesis, DynamoDB, SQS) share a `regionKey(region, name)` convention using `region + "::" + name`, so reusing that would keep the key format consistent, and since `AwsQueryController` always resolves a non null region, the `region == null` ternaries in `RdsQueryHandler` could be dropped by passing the region unconditionally.

hectorvent (approved):
Thanks @jvanzyl

PR #1600, opened 2026-06-26
fix(autoscaling): fail SSM commands for stale ASG instances

jvanzyl:
## Summary
- mark active SSM command invocations failed when ASG cleanup removes a stale instance
- prevent pending command invocations from being orphaned on replaced ASG instances
- keep the fix generic to Auto Scaling and SSM behavior

## Verification
- ./mvnw -q -Dtest='AutoScalingReconcilerTest,SsmCommandServiceDirectExecutionTest' test


hectorvent (approved):
Thanks @jvanzyl

PR #1601, opened 2026-06-26
feat(ec2): support registered images and snapshots

jvanzyl:
## Summary
- add EC2 RegisterImage support for describable local AMIs
- persist block device mappings and linked snapshot metadata
- add DescribeSnapshots support for the registered image/snapshot path

## Verification
- ./mvnw -q -Dtest=Ec2ServiceTest,Ec2ServicePersistenceTest,Ec2IntegrationTest test

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/AwsQueryController.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/AwsQueryController.java
@@ -157,7 +157,7 @@ public class AwsQueryController {
             "DescribeSecurityGroupRules", "ModifySecurityGroupRules",
             "UpdateSecurityGroupRuleDescriptionsIngress", "UpdateSecurityGroupRuleDescriptionsEgress",
             "CreateKeyPair", "DescribeKeyPairs", "DeleteKeyPair", "ImportKeyPair",
-            "DescribeImages",
+            "DescribeImages", "RegisterImage", "DescribeSnapshots",
             "CreateTags", "DeleteTags", "DescribeTags",
             "CreateInternetGateway", "DescribeInternetGateways", "DeleteInternetGateway",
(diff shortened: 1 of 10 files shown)


hectorvent:
Nice addition, the wire shapes all check out against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/ec2/2016-11-15/service-2.json) (ec2/2016-11-15), including the easy-to-miss `State` → `status` mapping in the snapshot XML. A few minor things worth tightening up:

**DescribeSnapshots owner parameter.** The model serializes `OwnerIds` as `Owner.N` on the wire, which is what every SDK and the CLI send. The handler accepts it, but the new tests only exercise the `OwnerId.1` and `OwnerIds.1` aliases, which AWS never sends. Could you switch the tests to the real name and consider dropping the aliases?

```
.formParam("Owner.1", "self")
```

**Architecture default.** The model doc says the default for EBS backed AMIs is `i386`, while `registerImage` defaults to `x86_64`. Probably the right call for an emulator, just worth a one line note so it reads as intentional.

**Snapshot auto creation.** Real AWS requires the snapshot referenced in a block device mapping to already exist, while here it gets synthesized since CreateSnapshot isn't emulated. Sensible shortcut, would be good to mention it in `docs/services/ec2.md`. Small parity win while you're there: AWS uses the sentinel `vol-ffffffff` as the `volumeId` for snapshots not backed by a volume, so the synthesized ones could set that instead of omitting the element.

**Docs table.** The supported actions table in `docs/services/ec2.md` could gain `RegisterImage` and `DescribeSnapshots`.

Also worth listing the deferred scope in the description (pagination, `RestorableBy.N`, the remaining BDM fields like `VirtualName` and `Iops`) so it's clearly intentional. An SDK based compat test would be a nice follow up too, it would have caught the `Owner.N` thing automatically.

PR #1602, opened 2026-06-26
feat(cloudcontrol): list resources from local services

jvanzyl:
## Summary
- add CloudControl ListResources handling through the existing JSON protocol path
- return resources created by Floci-backed services for supported CloudControl type names
- cover S3, EC2, IAM, and related local resource discovery paths

## Verification
- ./mvnw -q -Dtest='CloudControlIntegrationTest,Ec2IntegrationTest' test


jvanzyl:
Addressed the CI failure and Copilot IAM identifier feedback in the latest push. CloudControl now returns IAM RoleName/UserName as the ListResources Identifier to match Floci CloudFormation physical IDs, and the IAM list integration assertions now select the expected member instead of assuming singleton list state.\n\nFocused verification: `./mvnw -q -Dtest='CloudControlIntegrationTest,IamIntegrationTest,Ec2IntegrationTest' test`

hectorvent:
Nice one, the plumbing checks out against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cloudcontrol/2021-09-30/service-2.json) (cloudcontrol/2021-09-30): json 1.0 with the `CloudApiService` target prefix, the response shape is right, and `Properties` is correctly a serialized JSON string. Ignoring `MaxResults` is even correct, the model marks it as reserved. A few suggestions:

**Unknown type names.** Right now any unrecognized `TypeName` returns an empty 200, so a caller can't tell "no resources" from a typo. Real AWS throws the declared `TypeNotFoundException` for types that don't exist in the registry, and the input has a pattern constraint too. Something like this would make it much clearer:

```java
if (!typeName.matches("[A-Za-z0-9]{2,64}::[A-Za-z0-9]{2,64}::[A-Za-z0-9]{2,64}")) {
    throw new AwsException("ValidationException", ...);
}
// well formed but not in the registry
throw new AwsException("TypeNotFoundException",
        "The type 'AWS::Does::NotExist' cannot be found.", 404);
```

Returning an empty list for real AWS types that Floci doesn't back yet is a fair emulator choice, just worth documenting as intentional.

**Test isolation.** The loosened assertions in `Ec2IntegrationTest` and `IamIntegrationTest` are working around the CloudControl test creating resources in the shared `us-east-1` state. If `CloudControlIntegrationTest` used its own region in the auth headers, those two files wouldn't need touching and their assertions could stay strict.

**Docs.** A small `docs/services/cloudcontrol.md` page listing the supported type names would round it off, every service has one.

Non blocking: only `ListResources` for now with no pagination, `RoleArn`, `TypeVersionId` or `ResourceModel`, fine as deferred scope, just worth listing in the description. An SDK based compat test would be a nice follow up too.

hectorvent (approved):

PR #1622, opened 2026-06-28
feat(lakeformation): add broader AWS Lake Formation API support

brcordeiro:
## Summary

Adds broader AWS Lake Formation support to Floci by implementing additional JSON 1.1 actions and response shapes for tagging, data cells filters, permissions, Identity Center configuration, LF tags, hybrid opt-ins, and resource registration.

This brings the emulator closer to AWS behavior for the Lake Formation surface that was previously only partially implemented.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, Floci handled only a narrow subset of Lake Formation APIs, and several operations returned incomplete or mismatched shapes for AWS-compatible JSON requests and responses.

This change adds support for additional Lake Formation actions and aligns the emulator behavior with the expected AWS wire-format handling for:
- tagging and untagging resources
- data cells filters
- permissions and grant/list flows
- Identity Center configuration
- LF tags and related metadata
- resource registration and listing

Local verification covered:
- `./mvnw -Dtest=LakeFormationIntegrationTest test`
  - Result: 3 tests run, 0 failures, 0 errors, 0 skipped

## Checklist


hectorvent:
Thanks for taking on Lake Formation, the storage layer and service wiring follow the house patterns nicely. Before this can work with real clients there is one structural issue to address, plus some shape fixes.

**Lake Formation is rest json, not awsjson1.1.** Verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/lakeformation/2017-03-31/service-2.json) (lakeformation/2017-03-31) and the [API reference](https://docs.aws.amazon.com/lake-formation/latest/APIReference/API_GetDataLakeSettings.html): every operation is routed by URL path with no `X-Amz-Target` header at all.

```
POST /GetDataLakeSettings HTTP/1.1
Content-type: application/json
```

The Go SDK serializer and moto's `urls.py` confirm the same. Since this PR routes through `AwsJson11Controller` on `X-Amz-Target: AWSLakeFormation.<Action>`, no AWS SDK, CLI or Terraform request will ever reach these handlers, only hand crafted HTTP like the current tests. The fix is a JAX-RS controller with per operation paths, same as Lambda and API Gateway in this repo. moto's `moto/lakeformation/urls.py` doubles as a handy path checklist.

**Four operations don't exist in the API.** `TagResource`, `UntagResource` and `ListTagsForResource` are not Lake Formation actions (the [actions list](https://docs.aws.amazon.com/lake-formation/latest/APIReference/API_Operations.html) has 61, none of them these), and `PutDataCellsFilter` should be `CreateDataCellsFilter` plus `UpdateDataCellsFilter`.

**Several response shapes differ from the model.** The most impactful ones: `GetLFTag` and `GetLFTagExpression` return members at the top level, not wrapped; `SearchTablesByLFTags` returns `TableList`; `GetResourceLFTags` returns `LFTagOnDatabase`/`LFTagsOnTable`/`LFTagsOnColumns`; `RegisterResource`, `CreateLFTag` and `CreateLakeFormationOptIn` return empty `{}`; opt ins take `Principal`/`Resource`/`Condition`, not `PrincipalIdentifier`; `GetDataLakePrincipal` returns `{"Identity": "..."}`; and `GetTemporaryGlueTableCredentials.Expiration` is a timestamp, not an ISO string. `DataLakeSettings` also needs `DataLakeAdmins` and `CreateDatabaseDefaultPermissions` (10 members total in the model), otherwise `PutDataLakeSettings` silently drops the fields Terraform manages most.

Two smaller notes: empty successes should return `{}` (`Map.of()`) rather than a bodyless 200, and the new model classes need `@RegisterForReflection` for the native image since they are persisted and wire returned.

A compatibility test in any of the supported languages (`compatibility-tests/sdk-test-java` is the usual choice, Python/Go/TypeScript work too) would have caught the routing issue on the very first SDK call, so please include one with the rework. Happy to help review the next iteration, the persistence layer you built can carry over as is.

hectorvent:
One clarifying note on the protocol point, because the model metadata makes this genuinely easy to misread. Lake Formation's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/lakeformation/2017-03-31/service-2.json) metadata contains `"jsonVersion": "1.1"`, which looks like awsjson1.1 at first glance:

```json
{
  "jsonVersion": "1.1",
  "protocol": "rest-json",
  "protocols": ["rest-json"]
}
```

The `protocol` field is the one that decides serialization, and for rest-json the `jsonVersion` entry is vestigial. Two tells confirm it: the metadata has no `targetPrefix` (mandatory for the json protocol, it is what `X-Amz-Target` is built from), and botocore's own serializer only writes `X-Amz-Target` in `JSONSerializer` (`botocore/serialize.py:419`), while `RestJSONSerializer` builds requests purely from each operation's `requestUri`. So SDK clients will always arrive as `POST /<OperationName>` with plain `application/json`.

src/main/java/io/github/hectorvent/floci/services/lakeformation/LakeFormationJsonHandler.java
```diff
@@ -0,0 +1,279 @@
...
+        service.deleteLFTagExpression(region, request.path("Name").asText(null));
+        return Response.ok().build();
+    }
+
+    private Response listLFTagExpressions(JsonNode request, String region) {
+        return Response.ok(Map.of("LFTagExpressions", service.listLFTagExpressions(region))).build();
+    }
+
+    private Response addLFTagsToResource(JsonNode request, String region) {
+        String resourceArn = request.path("ResourceArn").asText(null);
```
abanna, line 254: Thanks for tackling a broad Lake Formation slice. One wire-shape catch: these APIs receive `Resource`, not a top-level `ResourceArn`; remove also takes `LFTags`, not `TagKeys`. An SDK request will currently put everything under a null resource and removal will be a no-op. Could we derive the storage key from `Resource` and add an add/get/remove test using the SDK payload?

src/main/java/io/github/hectorvent/floci/services/lakeformation/LakeFormationService.java
```diff
@@ -0,0 +1,487 @@
...
+        entry.put("PermissionsWithGrantOption", request.get("PermissionsWithGrantOption"));
+        entry.put("CatalogId", request.get("CatalogId"));
+        entries.add(entry);
+        permissionsStore.put(permissionsKey(region), entries);
+        return entries;
+    }
+
+    public List<Map<String, Object>> revokePermissions(String region, Map<String, Object> request) {
+        List<Map<String, Object>> entries = readList(permissionsStore, permissionsKey(region));
+        entries.removeIf(existing -> samePermission(existing, request));
```
abanna, line 171: Nice start on the grant/revoke flow. Could we remove only the requested permissions here rather than the entire grant? A grant with `SELECT` and `ALTER`, followed by a revoke of `SELECT`, is removed completely because `samePermission` only compares the principal and resource. A small regression test would make the intended behavior clear.

src/main/java/io/github/hectorvent/floci/services/lakeformation/LakeFormationService.java
```diff
@@ -0,0 +1,487 @@
...
+
+    public Map<String, Object> getLFTag(String region, String tagKey) {
+        return readList(lfTagsStore, lfTagKey(region)).stream()
+                .filter(existing -> Objects.equals(existing.get("TagKey"), tagKey))
+                .findFirst()
+                .orElseGet(LinkedHashMap::new);
+    }
+
+    public Map<String, Object> updateLFTag(String region, Map<String, Object> request) {
+        return createLFTag(region, request);
```
abanna, line 307: Thanks for including the update path. Could this handle `UpdateLFTag` separately? That API sends `TagValuesToAdd` and `TagValuesToDelete`, while `CreateLFTag` expects `TagValues`, so a normal update currently overwrites the tag with a null value and creates a missing tag instead of returning `EntityNotFoundException`. A couple of add/delete cases would cover it nicely.

src/main/java/io/github/hectorvent/floci/services/lakeformation/LakeFormationService.java
```diff
@@ -0,0 +1,487 @@
...
+        Map<String, Object> principal = new LinkedHashMap<>();
+        principal.put("DataLakePrincipalIdentifier", dataLakePrincipalIdentifier);
+        principal.put("DataLakePrincipalType", "IAM");
+        return principal;
+    }
+
+    public Map<String, Object> createIdentityCenterConfiguration(String region, Map<String, Object> request) {
+        Map<String, Object> config = new LinkedHashMap<>();
+        config.put("CatalogId", request.get("CatalogId"));
+        config.put("ExternalFilteringConfiguration", request.get("ExternalFilteringConfiguration"));
```
abanna, line 196: Thanks for carrying this configuration through to storage. One naming mismatch: the API sends `ExternalFiltering`, so SDK requests are currently dropped here. The create/describe response also needs the other documented fields, such as `InstanceArn` and `ApplicationArn`, to round-trip. Could we model the request/response shape and add a create-to-describe SDK test?

hectorvent:
Thank you for this, and sorry for the slow close. Lake Formation is a big surface to take on, and the storage layer and service wiring here follow the house patterns nicely.

The blocker is still the protocol rather than the shapes. Lake Formation is rest-json: 61 operations, each with its own request URI (`POST /GetDataLakeSettings`) and no target prefix at all, per [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/lakeformation/2017-03-31/service-2.json). The branch still registers the service inside `AwsJson11Controller`'s `X-Amz-Target` switch, so a real SDK client would not reach the handler. The `fix: Lake Formation AWS protocol compliance issues` commit corrected response shapes, which was the smaller half of the earlier note, but the routing is unchanged.

Worth saying plainly that the integration tests pass because they speak the same dialect as the implementation, posting to `/` with an `X-Amz-Target` header, so green CI was never going to catch this. And it is an easy thing to get wrong: the model carries `jsonVersion: 1.1`, and Glue and Athena, which sit either side of Lake Formation in that same switch, genuinely are awsjson1.1. Lake Formation is the odd one out.

So what is left is a routing rewrite plus a rewrite of the tests, and there is no open issue asking for Lake Formation yet. I would rather not have you spend more evenings on it without knowing it is wanted. Closing for now with thanks. If Lake Formation does come up from users I will reopen this and ping you first, and if you would rather carry on regardless, say so and I will point you at `ApiGatewayController` as the rest-json pattern to copy.

PR #1623, opened 2026-06-28
fix(dynamodb): return a null Message for non-failed item cancellation reason

### LINKED ISSUE #1604: [BUG] Floci returns slightly different `CancellationReason` when compared to AWS/dynamodb-local
## Service

Dynamodb

## AWS API Action

`TransactWriteItemsRequest`

## Expected behavior

What the real AWS SDK/CLI returns after a cancelled transaction:

`CancellationReason code=None, message=null`

## Actual behavior

What floci returns:

`CancellationReason code=None, message=""`

It's an `""` (empty string) instead of `null`.

This blew up in my test suite, because my exception handling was:

```java
  public RuntimeException wrap(TransactionCanceledException e) {
    Iterator<? extends WriteOp> iterator = all.iterator();
    for (CancellationReason reason : e.cancellationReasons()) {
      WriteOp op = iterator.next();
      if (reason.message() != null) {
        return op.handler().onError(e);
      }
    }
    return e;
  }
```

It was easily fixed by changing 

`if (reason.message() != null) {`

to

`if (!"None".equals(reason.code())) {`

But this is indeed different behavior from the real thing.

## Reproduction

This compares against the `dynamodb-local` docker image.

I also compared against the real AWS dynamodb, and it was the same as `dynamodb-local`.

```java
import java.net.URI;
import java.util.List;
import java.util.Map;

import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
import software.amazon.awssdk.regions.Region;
import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
import software.amazon.awssdk.services.dynamodb.model.AttributeDefinition;
import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
import software.amazon.awssdk.services.dynamodb.model.BillingMode;
import software.amazon.awssdk.services.dynamodb.model.CancellationReason;
import software.amazon.awssdk.services.dynamodb.model.ConditionCheck;
import software.amazon.awssdk.services.dynamodb.model.KeySchemaElement;
import software.amazon.awssdk.services.dynamodb.model.KeyType;
import software.amazon.awssdk.services.dynamodb.model.ScalarAttributeType;
import software.amazon.awssdk.services.dynamodb.model.TransactWriteItem;
import software.amazon.awssdk.services.dynamodb.model.TransactWriteItemsRequest;
import software.amazon.awssdk.services.dynamodb.model.TransactionCanceledException;

/// Before running, choose one:
///
/// - `docker run --rm -it -p 4566:4566 floci/floci:latest`
/// - `docker run --rm -it -p 4566:8000 amazon/dynamodb-local:latest`
public final class FlociCancellationReasonRepro {
  public static void main(String[] args) {
    String endpoint = "http://localhost:4566";
    String table = "repro_txn_reason";

    DynamoDbClient db = DynamoDbClient.builder()
      .endpointOverride(URI.create(endpoint))
      .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create("x", "x")))
      .region(Region.US_EAST_1)
      .build();

    try {
      db.deleteTable(b -> b.tableName(table));
    } catch (RuntimeException ignored) {
      // table didn't exist
    }
    db.createTable(
      (b) ->
        b.tableName(table)
          .billingMode(BillingMode.PAY_PER_REQUEST)
          .attributeDefinitions(
            AttributeDefinition
              .builder()
              .attributeName("pk")
              .attributeType(ScalarAttributeType.S)
              .build()
          )
          .keySchema(
            KeySchemaElement.builder().attributeName("pk").keyType(KeyType.HASH).build()
          )
    );
    db.waiter().waitUntilTableExists(b -> b.tableName(table));
    db.putItem(b -> b.tableName(table).item(Map.of("pk", AttributeValue.fromS("A"))));

    // 2-item transaction: item 0 PASSES (A exists), item 1 FAILS (B absent) -> whole txn cancels.
    try {
      db.transactWriteItems(
        TransactWriteItemsRequest.builder()
          .transactItems(
            TransactWriteItem
              .builder()
              .conditionCheck(check(table, "A", "attribute_exists(pk)"))
              .build(),
            TransactWriteItem
              .builder()
              .conditionCheck(check(table, "B", "attribute_exists(pk)"))
              .build()
          )
          .build()
      );
      System.out.println("UNEXPECTED: transaction was not cancelled");
    } catch (TransactionCanceledException e) {
      List<CancellationReason> reasons = e.cancellationReasons();
      CancellationReason notFailed = reasons.get(0); // item 0 did NOT cause the cancellation
      System.out.println("endpoint = " + endpoint);
      System.out.println("reason[0] (non-failed item): code=" + notFailed.code()
        + ", message=" + (
        notFailed.message() == null
          ? "null"
          : "\"" + notFailed.message() + "\""
      ));
      System.out.println("reason[1] (failed item):     code=" + reasons.get(1).code());
      System.out.println();
      if (notFailed.message() == null) {
        System.out.println("OK (spec-conformant): non-failed item's reason has a null message.");
      } else {
        throw new RuntimeException(
          "BUG: non-failed item's reason.message() is "
            + "\"" + notFailed.message() + "\" but should be null. "
            + "Code is correctly \"None\"; only the message deviates."
        );
      }
    }
  }

  private static ConditionCheck check(String table, String pk, String expr) {
    return ConditionCheck
      .builder()
      .tableName(table)
      .key(Map.of("pk", AttributeValue.fromS(pk)))
      .conditionExpression(expr)
      .build();
  }
}
```

## Environment

- Floci version / image tag: `floci/floci:1.5.28`
- Java SDK version (if applicable): Tested both `2.40.6` and 
- How you're running Floci (Docker / native / `mvn quarkus:dev`): docker

b6k-dev:
## Summary

Closes #1604 

Summary:
- Add regression coverage for DynamoDB TransactWriteItems cancellation reasons
- Return a null Message field for non-failed items

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behaviour**: when a DynamoDB transaction is cancelled, non-failed transaction items should have CancellationReason code None and message == null. Instead floci returned an empty string `""` in the message field.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbConformanceChangesTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbConformanceChangesTest.java
@@ -7,14 +7,13 @@ import software.amazon.awssdk.services.dynamodb.model.AttributeDefinition;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValueUpdate;
 import software.amazon.awssdk.services.dynamodb.model.BatchGetItemResponse;
-import software.amazon.awssdk.services.dynamodb.model.BatchWriteItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.BillingMode;
 import software.amazon.awssdk.services.dynamodb.model.ComparisonOperator;
 import software.amazon.awssdk.services.dynamodb.model.Condition;
+import software.amazon.awssdk.services.dynamodb.model.ConditionCheck;
 import software.amazon.awssdk.services.dynamodb.model.ConditionalCheckFailedException;
-import software.amazon.awssdk.services.dynamodb.model.ExpectedAttributeValue;
 import software.amazon.awssdk.services.dynamodb.model.DynamoDbException;
-import software.amazon.awssdk.services.dynamodb.model.GetItemRequest;
+import software.amazon.awssdk.services.dynamodb.model.ExpectedAttributeValue;
 import software.amazon.awssdk.services.dynamodb.model.GetItemResponse;
 import software.amazon.awssdk.services.dynamodb.model.KeySchemaElement;
 import software.amazon.awssdk.services.dynamodb.model.KeyType;
@@ -37,6 +36,8 @@ import software.amazon.awssdk.services.dynamodb.model.ScanResponse;
 import software.amazon.awssdk.services.dynamodb.model.Select;
 import software.amazon.awssdk.services.dynamodb.model.Tag;
 import software.amazon.awssdk.services.dynamodb.model.TransactWriteItem;
+import software.amazon.awssdk.services.dynamodb.model.TransactWriteItemsRequest;
+import software.amazon.awssdk.services.dynamodb.model.TransactionCanceledException;
 import software.amazon.awssdk.services.dynamodb.model.UpdateItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.WriteRequest;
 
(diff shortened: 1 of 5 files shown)


b6k-dev:
@hectorvent 

I've run into a CI test job failure in Ec2ContainerManagerTest.launchInstanceUserDataStreamToCloudWatch - this was likely a pre-existing flaky test unrelated to this PR's DynamoDB changes.

**Root cause**: awaitUntil returns the instant the instance state becomes "running", but executeUserData (which calls streamToCloudWatchLogs) runs after the state transition in the same executor thread. The verify() then fires before the executor reaches that code path.

**Applied fix**: Replace the immediate verify() with verify(mock, timeout(2000)) so Mockito polls for the invocation.

hectorvent (approved):
Thanks @b6k-dev, 
Solid fix

Issue #1081, opened 2026-05-29
[BUG] Properly-confirmed HTTP subscriptions (200 flow) never receive SNS delivery — only failed-confirmation endpoints get delivery


alexhayhoetebex:
## Summary

Follow-on to #841 / PR #865 (implemented in 1.5.18). HTTP delivery is now implemented, but there is a logical inversion in the delivery queue: only subscriptions whose `SubscriptionConfirmation` HTTP handshake **failed** (endpoint returned non-2xx) receive delivery. Subscriptions that correctly implement the full AWS confirmation flow (endpoint returns 200 **and** calls back `ConfirmSubscription`) are confirmed in Floci's internal state but **never receive delivery**.

## Steps to reproduce

1. Start Floci 1.5.20 with `SERVICES=sns`
2. Create a topic
3. Subscribe an HTTP endpoint that:
   - Returns **HTTP 200** to Floci's `SubscriptionConfirmation` POST
   - Immediately POSTs back to Floci's SNS endpoint to confirm:
     ```
     POST http://localhost:4566/ HTTP/1.1
     Content-Type: application/x-www-form-urlencoded

     Action=ConfirmSubscription&TopicArn=<arn>&Token=<token-from-payload>&Version=2010-03-31
     ```
4. Verify: Floci logs `Confirmed subscription <arn>` and `sns-subscriptions.json` shows `PendingConfirmation: false`
5. Publish a message to the topic
6. **Observe**: message is **not delivered** to the endpoint

## Contrast — what does get delivery

Subscribe an HTTP endpoint that returns **HTTP 500** to Floci's `SubscriptionConfirmation` POST. Then API-confirm it via `aws sns confirm-subscription --token <token>`. After confirmation, publish a message — the endpoint **does** receive delivery.

## Expected behaviour (matches real AWS SNS)

Any subscription with `PendingConfirmation=false` should receive delivery, regardless of whether the confirmation was completed via the HTTP handshake or the `ConfirmSubscription` API directly.

## Actual behaviour

Floci's delivery queue appears to be populated only when it logs:
`Sent SubscriptionConfirmation to <url> returned non-success status <code>`

Subscriptions where the HTTP POST returned 200 are never entered into this set, so `Publish` never dispatches to them — even though `PendingConfirmation=false` in storage.

## Root cause hypothesis

The skip guard in `SnsService.java`:
```java
if ("true".equals(sub.getAttributes().get("PendingConfirmation"))) {
    continue; // skip delivery
}
```

[...]

PR #1625, opened 2026-06-28
fix(s3,sns): preserve root service-host routing

### LINKED ISSUE #1494: [BUG] S3 ListBuckets returns 404 NoSuchBucket when the endpoint host is the virtual-hosted S3 domain (e.g. s3.localhost.localstack.cloud)
## Service

S3

## AWS API Action

ListBuckets (and any other service-level S3 call that hits `GET /` with no bucket)

## Expected behavior

When the request host is the **S3 service endpoint** (a host whose first label is `s3.`, with **no bucket subdomain**), `ListBuckets` should be treated as a service-level operation and return the list of buckets (HTTP 200 + `ListAllMyBucketsResult`).

This mirrors real AWS: `ListBuckets` is sent to the bare service host `s3.amazonaws.com` (`GET https://s3.amazonaws.com/`) and returns the bucket list. The host starts with `s3.` but is **not** a bucket named `s3`.

## Actual behavior

Floci routes **any** host that starts with `s3.` through virtual-hosted-style parsing, including the bare service host. On a bucketless `GET /` it extracts an empty/invalid bucket and returns:

```
HTTP 404
api error NoSuchBucket: The specified bucket does not exist.
```

This happens for both `s3.localhost.localstack.cloud` and Floci's own `s3.localhost.floci.io`, which shows it is not specific to the localstack.cloud domain — it's that there is no service-level handler for an `s3.`-prefixed host.

## Reproduction

```bash
# Floci listening on :4566. Vary ONLY the Host header on a bucketless GET /.

# 1) Plain host -> works (200, returns ListAllMyBucketsResult)
curl -s -o /dev/null -w "localhost                       -> %{http_code}\n" \
  -H "Host: localhost" http://localhost:4566/

# 2) S3 service host (no bucket subdomain) -> 404 NoSuchBucket  <-- BUG
curl -s -o /dev/null -w "s3.localhost.localstack.cloud   -> %{http_code}\n" \
  -H "Host: s3.localhost.localstack.cloud" http://localhost:4566/

# 3) Floci's own S3 service host -> also 404  <-- BUG
curl -s -o /dev/null -w "s3.localhost.floci.io           -> %{http_code}\n" \
  -H "Host: s3.localhost.floci.io" http://localhost:4566/

# Equivalent via AWS CLI (endpoint = the S3 service host) also fails:
aws --endpoint-url http://s3.localhost.localstack.cloud:4566 s3api list-buckets
# -> An error occurred (NoSuchBucket) when calling the ListBuckets operation
```

Observed:

```
localhost                       -> 200
s3.localhost.localstack.cloud   -> 404
s3.localhost.floci.io           -> 404
```

### Where this endpoint comes from (so it's not just a contrived curl)

This is hit in practice through HashiCorp's `terraform-local` (`tflocal`) wrapper, which many LocalStack users run. `tflocal` sets the S3 provider endpoint to the **virtual-hosted service domain by default**:

```python
# terraform-local/bin/tflocal
LOCALHOST_HOSTNAME = "localhost.localstack.cloud"
S3_HOSTNAME = os.environ.get("S3_HOSTNAME") or f"s3.{LOCALHOST_HOSTNAME}"   # -> s3.localhost.localstack.cloud
# get_service_endpoint("s3") -> http://s3.localhost.localstack.cloud:<port>
```

So the AWS Terraform provider sends `ListBuckets` to `Host: s3.localhost.localstack.cloud` (e.g. triggered by the `aws_canonical_user_id` data source used for S3 bucket ACL/ownership), and Floci returns 404. Note that `s3_use_path_style = true` does **not** help, because `ListBuckets` has no bucket — path-style vs virtual-hosted only changes addressing for bucket-scoped calls; the bucketless request still goes to the endpoint host.

Workaround today: set `S3_HOSTNAME=localhost` so the endpoint becomes a plain host. The underlying request handling for the bare `s3.*` service host should still be fixed.

## Environment

- Floci version / image tag: `floci/floci:latest`
- Java SDK version (if applicable): n/a (reproduced via `curl` / AWS CLI; also via AWS Terraform provider `~> 6.0`)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (`floci/floci:latest`)

hampsterx:
## Summary

Closes #1494.
Refs #1081.

- Treat bare `s3.*` service endpoint hosts as bucketless S3 requests.
- Route SNS `ConfirmSubscription` GET requests on `/` through SNS before normal S3 bucket listing.
- Add parser and integration coverage for S3 service-host routing plus SNS HTTP confirmation delivery flows.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Floci incorrectly parsed bare S3 service endpoint hosts such as `s3.localhost.localstack.cloud` and `s3.localhost.floci.io` as virtual-hosted bucket requests for a bucket named `s3`. Bucketless `GET /` now remains a service-level ListBuckets request, while bucket virtual-host forms still rewrite to bucket paths.

SNS `SubscribeURL` confirmation uses a GET request like `/?Action=ConfirmSubscription&TopicArn=...&Token=...`. That root-path GET now reaches the SNS query handler before the S3 ListBuckets fallback, matching the AWS confirmation flow without changing normal S3 root behavior.

## Changes

- `S3VirtualHostFilter`: skip virtual-host rewriting for bare S3 service endpoint hosts.
- `S3Controller`: route root `ConfirmSubscription` query-string requests to SNS before ListBuckets.
- `S3VirtualHostFilterTest` and `S3VirtualHostIntegrationTest`: add service-host parser and ListBuckets regression coverage.
- `SnsHttpDeliveryIntegrationTest`: add confirmation regressions for inline 200 confirmation and SubscribeURL GET confirmation.

## Test plan

- [x] `./mvnw test -Dtest=S3VirtualHostFilterTest,S3VirtualHostIntegrationTest,SnsHttpDeliveryIntegrationTest,AwsQueryControllerIntegrationTest`

## Review

- Internal fallback review: clean, no regressions detected.
- Claude bridge review: unavailable locally because the bridge is not logged in.
- Gemini bridge review: unavailable locally because the Gemini CLI is not installed.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -7,6 +7,7 @@ import io.github.hectorvent.floci.core.common.AwsNamespaces;
 import io.github.hectorvent.floci.core.common.XmlBuilder;
 import io.github.hectorvent.floci.core.common.XmlParser;
 import io.github.hectorvent.floci.core.common.RegionResolver;
+import io.github.hectorvent.floci.services.sns.SnsQueryHandler;
 import io.github.hectorvent.floci.services.s3.model.Bucket;
 import io.github.hectorvent.floci.services.s3.model.GetObjectAttributesParts;
 import io.github.hectorvent.floci.services.s3.model.GetObjectAttributesResult;
@@ -29,6 +30,7 @@ import jakarta.ws.rs.*;
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.StreamingOutput;
 import jakarta.ws.rs.core.UriInfo;
@@ -86,17 +88,20 @@ public class S3Controller {
     private final S3Service s3Service;
     private final S3SelectService s3SelectService;
     private final RegionResolver regionResolver;
+    private final SnsQueryHandler snsQueryHandler;
     private final io.quarkus.vertx.http.runtime.CurrentVertxRequest currentVertxRequest;
     private final io.github.hectorvent.floci.services.floci.ui.UiPages uiPages;
 
     @Inject
     public S3Controller(S3Service s3Service, S3SelectService s3SelectService,
                         RegionResolver regionResolver,
+                        SnsQueryHandler snsQueryHandler,
                         io.quarkus.vertx.http.runtime.CurrentVertxRequest currentVertxRequest,
                         io.github.hectorvent.floci.services.floci.ui.UiPages uiPages) {
         this.s3Service = s3Service;
         this.s3SelectService = s3SelectService;
         this.regionResolver = regionResolver;
+        this.snsQueryHandler = snsQueryHandler;
         this.currentVertxRequest = currentVertxRequest;
         this.uiPages = uiPages;
     }
@@ -106,9 +111,17 @@ public class S3Controller {
     @GET
     @Produces({MediaType.APPLICATION_XML, MediaType.TEXT_HTML})
     public Response listBuckets(@HeaderParam("X-Amz-Target") String target,
-                                @HeaderParam("Accept") String accept) {
+                                @HeaderParam("Accept") String accept,
+                                @Context HttpHeaders httpHeaders,
+                                @Context UriInfo uriInfo) {
         if (target != null) {
             return null;
+        }
+        MultivaluedMap<String, String> queryParams = uriInfo.getQueryParameters();
+        String action = queryParams.getFirst("Action");
+        if ("ConfirmSubscription".equals(action)) {
+            String region = regionResolver.resolveRegion(httpHeaders);
+            return snsQueryHandler.handle(action, queryParams, region);
         }
         // A browser hitting the root endpoint (Accept: text/html) gets the Floci
         // landing page; SDK/CLI callers (no Accept, */*, or an XML/JSON Accept) fall
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -2,6 +2,7 @@ package io.github.hectorvent.floci.services.s3;
 
 import static io.github.hectorvent.floci.services.s3.S3RequestParser.hasQueryParam;
 
+import io.github.hectorvent.floci.core.common.AwsArnUtils;
 import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.core.common.AwsNamespaces;
 import io.github.hectorvent.floci.core.common.XmlBuilder;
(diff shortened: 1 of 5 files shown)


hampsterx:
Rebased onto main and fixed the red CI. The failure was a race in the new `SnsHttpDeliveryIntegrationTest`: the test's confirming-webhook recorded the inbound `SubscriptionConfirmation` before it actually confirmed, so `awaitRequests()` unblocked and the test could publish ahead of confirmation and observe 0 deliveries. It now records the request only after `confirmSubscription()` completes (making `awaitRequests()` a real barrier) and logs a confirmation failure instead of swallowing it (addresses the Greptile note). Product code unchanged. All checks green.

hectorvent:
Nice fix, verified against the Query protocol and it holds up: `SubscribeURL` confirmation is a plain GET with query params, and I confirmed your new SubscribeURL test fails on main and passes here, so it targets the real gap. The bare `s3.*` host check is also provably safe since bucket names need 3+ characters, so a bucket named `s3` cannot exist.

One finding while you are in this code: `UnsubscribeURL` has the identical routing gap. [SnsService.java:1411](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java#L1411) emits

```java
node.put("UnsubscribeURL", baseUrl + "/?Action=Unsubscribe&SubscriptionArn=" + subscriptionArn);
```

so following it today returns ListBuckets XML instead of unsubscribing. Extending the delegation would close the whole class:

```java
if ("ConfirmSubscription".equals(action) || "Unsubscribe".equals(action)) {
    return snsQueryHandler.handle(action, queryParams, region);
}
```

plus a small test mirroring the SubscribeURL one.

Two minors: the hardcoded host list in `isS3ServiceEndpointHost` duplicates what `floci.dns.extra-suffixes` already knows, fine now but worth deriving from config in a follow up. And once this merges, #1081 can be closed as well, the inline 200 confirmation scenario it reports already works on main and your new test locks it in.

hampsterx:
@hectorvent extended the root-GET delegation to `Unsubscribe` as suggested, plus a test.

One addition while in there: SubscribeURL/UnsubscribeURL are unsigned GETs, so the region can't come from the auth header. It fell back to the default region and would 404 a subscription created in any other region (`unsubscribe` keys by `sub::<region>::<SubscriptionArn>`). Now I pull the region off the ARN the link carries via `AwsArnUtils.regionOrDefault`, which also covers `ConfirmSubscription`. The new test drives the full confirm+unsubscribe-via-GET flow in us-west-2 to lock it in (fails on header-only region, passes here).

Minors: agree the hardcoded host list in `isS3ServiceEndpointHost` should derive from `floci.dns.extra-suffixes`, can take that as a follow-up unless you'd rather it land here. And I'll flag #1081 to close once this merges.

hampsterx:
Re the host list: keeping it as a separate follow-up like you suggested. It's a bit more than dedup, `localhost.localstack.cloud`/`localhost.floci.io` are always-on builtins (`DEFAULT_SUFFIX`/`LOCALSTACK_SUFFIX`) while `extra-suffixes` is opt-in, so it derives to `builtins ∪ extra-suffixes` and starts recognising `s3.<extra-suffix>` as a service host, which wants its own test. Separate PR once this lands.

hectorvent (approved):
Thanks @hampsterx

PR #1626, opened 2026-06-28
feat(kms): add ListKeyPolicies Support for KMS

### LINKED ISSUE #1528: [FEAT] KMS: implement ListKeyPolicies
## Service

KMS

## API Action / Feature

`ListKeyPolicies`

## AWS Documentation

https://docs.aws.amazon.com/kms/latest/APIReference/API_ListKeyPolicies.html

Related (already working in Floci): `GetKeyPolicy`, `PutKeyPolicy`

## Why is this needed?

**What happens:** `amazon.aws.kms_key` (Ansible) fails against Floci with:

```
UnsupportedOperation: Operation ListKeyPolicies is not supported.
```

**Why:** The module calls `ListKeyPolicies` for idempotency. The rest of the KMS key policy API works in Floci — we tested all 3 AWS operations:

| API | Floci |
|---|---|
| `ListKeyPolicies` | FAIL |
| `PutKeyPolicy` | OK |
| `GetKeyPolicy` | OK |

**How we found it:** Ansible playbook with `-vvv` (`env=test`, `AWS_ENDPOINT_URL=http://floci:4566`). Confirmed with direct HTTP calls via `floci.sh`.

**Repro:**

```bash
bash floci.sh
# or manually:
KEY_ID=$(curl -sS -X POST http://localhost:4566/ \
  -H "Content-Type: application/x-amz-json-1.1" \
  -H "X-Amz-Target: TrentService.CreateKey" \
  -d '{"Description":"test"}' | python3 -c "import sys,json; print(json.load(sys.stdin)['KeyMetadata']['KeyId'])")

curl -sS -X POST http://localhost:4566/ \
  -H "Content-Type: application/x-amz-json-1.1" \
  -H "X-Amz-Target: TrentService.ListKeyPolicies" \
  -d "{\"KeyId\":\"$KEY_ID\"}"
# → UnsupportedOperation
```

- [ ] Yes
- [x] No

Marcel2603:
## Summary

This PR implements the KMS `ListKeyPolicies` action, resolving an `UnsupportedOperation` error when clients attempt to list key policies. It dynamically fetches the default policy using `getKeyPolicy`.

Closes #1528 

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

This implementation adheres to the AWS KMS expected response shape by:
- Enforcing limit boundaries (1-1000).
- Throwing `InvalidMarkerException` on invalid markers.
- Fetching the actual policy applied to the key rather than hardcoding.
- Always returning a valid JSON shape including `"NextMarker": null` when `Truncated` is `false`.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsJsonHandler.java
@@ -67,6 +67,7 @@ public class KmsJsonHandler {
             case "ListResourceTags" -> handleListResourceTags(request, region);
             case "GetKeyPolicy" -> handleGetKeyPolicy(request, region);
             case "PutKeyPolicy" -> handlePutKeyPolicy(request, region);
+            case "ListKeyPolicies" -> handleListKeyPolicies(request, region);
             case "UpdateKeyDescription" -> handleUpdateKeyDescription(request, region);
             case "GetKeyRotationStatus" -> handleGetKeyRotationStatus(request, region);
             case "EnableKeyRotation" -> handleEnableKeyRotation(request, region);
@@ -459,6 +460,15 @@ public class KmsJsonHandler {
                 request.path("Policy").asText(),
                 region);
         return Response.ok(objectMapper.createObjectNode()).build();
+    }
+
+    private Response handleListKeyPolicies(JsonNode request, String region) {
+        Map<String, Object> result = service.listKeyPolicies(
+                requiredText(request, "KeyId"),
+                request.path("Limit").asInt(100),
+                request.path("Marker").asText(null),
+                region);
+        return Response.ok(objectMapper.valueToTree(result)).build();
     }
 
     private Response handleUpdateKeyDescription(JsonNode request, String region) {
--- a/docs/services/kms.md
+++ b/docs/services/kms.md
@@ -33,6 +33,7 @@
 | `ListResourceTags` | List tags |
 | `GetKeyPolicy` | Get a key's resource policy |
 | `PutKeyPolicy` | Update a key's resource policy |
+| `ListKeyPolicies` | List a key's resource policies |
 | `UpdateKeyDescription` | Update a key's description |
 | `GetKeyRotationStatus` | Check if automatic key rotation is enabled |
 | `EnableKeyRotation` | Enable automatic key rotation (symmetric keys only) |
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
@@ -489,6 +489,26 @@ public class KmsService {
         LOG.infov("Updated key policy for KMS key: {0} in {1}", key.getKeyId(), region);
     }
 
+    public Map<String, Object> listKeyPolicies(String keyId, int limit, String marker, String region) {
+        if (limit < 1 || limit > 1000) {
+            throw new AwsException("ValidationException",
+                    "1 validation error detected: Value '" + limit + "' at 'limit' failed to satisfy constraint: Member must have value greater than or equal to 1 and less than or equal to 1000", 400);
+        }
+        if (marker != null && !marker.isBlank()) {
+            throw new AwsException("InvalidMarkerException",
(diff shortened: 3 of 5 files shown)


hectorvent:
Thanks for this, it closes #1528 nicely. Verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/kms/2014-11-01/service-2.json) (kms/2014-11-01) and moto: the core response is right, `PolicyNames: ["default"]` matches the model doc ("The only valid value is default") and moto's reference response exactly, and the limit validation matches the `LimitType` bounds. The `NextMarker: null` point is already well covered by the existing review thread, +1 to fixing it there.

Two things not raised yet:

**InvalidMarkerException is not declared for this operation.** ListKeyPolicies declares only `NotFoundException`, `InvalidArnException`, `DependencyTimeoutException`, `KMSInternalException` and `KMSInvalidStateException`. moto simply ignores the Marker here, which is also the simpler and model faithful option, so I would drop the marker check rather than throw an undeclared code.

**The 404 assertion cements a pre-existing deviation.** `NotFoundException` carries no `httpStatusCode` in the model, so the JSON protocol default applies and real KMS returns 400. The existing key resolution path in Floci uses 404, which is not this PR's fault, but the new integration test now asserts `statusCode(404)`. Worth a follow up issue for the KMS wide status rather than more tests locking it in.

Nice touch updating the docs table alongside the code.

src/main/java/io/github/hectorvent/floci/services/kms/KmsJsonHandler.java
```diff
@@ -461,6 +462,15 @@ private Response handlePutKeyPolicy(JsonNode request, String region) {
         return Response.ok(objectMapper.createObjectNode()).build();
     }
 
+    private Response handleListKeyPolicies(JsonNode request, String region) {
+        Map<String, Object> result = service.listKeyPolicies(
+                requiredText(request, "KeyId"),
```
abanna, line 467: @Marcel2603 thanks for filling in this KMS operation. Could we make the input validation a little stricter? `KeyId: ""` passes `requiredText` and ends up as a not-found lookup, while `Limit: 1.5` is coerced to `1` by `asInt`. KMS treats both as validation errors. A couple of handler tests would cover it.

src/main/java/io/github/hectorvent/floci/services/kms/KmsJsonHandler.java
```diff
@@ -461,6 +462,15 @@ private Response handlePutKeyPolicy(JsonNode request, String region) {
         return Response.ok(objectMapper.createObjectNode()).build();
     }
 
+    private Response handleListKeyPolicies(JsonNode request, String region) {
+        Map<String, Object> result = service.listKeyPolicies(
```
abanna, line 466: One other small edge case: `ListKeyPolicies` is not cross-account, but this call strips an ARN down to its ID before lookup. If two accounts use the same override ID, an ARN for account B resolves the caller account’s key. Could we reject foreign account/Region ARNs first? A two-account test would pin it down.

hectorvent:
Thank you for this, and sorry it waited. The implementation was right and it is now in #2046 with your commit carried as a co-author, closing #1528 on merge.

I picked it up rather than sending you back for a third round, since only two small things were left and you had already been waiting a while.

Both were about not inventing error codes. `InvalidMarkerException` turned out to be a nice trap: it is a real KMS shape, but the model declares it only on the genuinely paginated calls like `ListKeys` and `ListGrants`, and pointedly not on `ListKeyPolicies`. That makes sense once you see why, since there is only ever one policy name, so the result cannot truncate and no marker is ever valid to reject. The same applied to the `Limit` check, where `ValidationException` is not even a shape in the KMS model, so I dropped that too and let both parameters be accepted and ignored. `listGrants` in the same class already clamps rather than throws, so that matches the house pattern.

The other change was dropping the `statusCode(404)` assertion from the new test. Floci returns 404 for KMS `NotFoundException` where real KMS uses 400, and that is a pre-existing service-wide thing worth its own issue rather than something a new test should lock in.

Everything else came over as you wrote it: the `default` policy name, the key resolution through `getKeyPolicy`, and the shape of your tests. Closing in favour of #2046, with thanks.

PR #1627, opened 2026-06-29
chore(ci): skip nightly workflow on forks

martincostello:
## Summary

Do not run the nightly workflow on a schedule from forks as its a waste of resources and they fail due to a lack of Docker credentials.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A

## Checklist

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -28,6 +28,7 @@ jobs:
   prepare:
     name: Prepare nightly version
     runs-on: ubuntu-latest
+    if: github.event_name != 'schedule' || !github.event.repository.fork
     outputs:
       version: ${{ steps.date.outputs.version }}
 


.github/workflows/nightly.yml
```diff
@@ -28,6 +28,7 @@ jobs:
   prepare:
     name: Prepare nightly version
     runs-on: ubuntu-latest
+    if: github.event_name != 'schedule' || !github.event.repository.fork
```
hectorvent, line 31: @martincostello thanks for the PR. 

I ran manually sometimes this action. 
```yml
 workflow_dispatch:        # Allow maintainers to trigger manually
```

Could we just keep? 
```yml
if: !github.event.repository.fork
```

hectorvent (changes requested):

.github/workflows/nightly.yml
martincostello: The condition is:

NOT a schedule (so allows manual trigger on any repo)
OR not fork (allows any trigger)

So it already does what you want?

Removing that condition disallows manual triggers of a fork.

hectorvent: you are correct. All good. Merging

hectorvent (approved):

.github/workflows/nightly.yml
hectorvent: Thanks!

PR #1630, opened 2026-06-29
fix(apigateway): fall through to less specific resources on method mi…

### LINKED ISSUE #1547: [BUG] API Gateway REST: method-less concrete/{id} resource shadows /{proxy+} ANY → 405 instead of falling through
## Service

API Gateway (REST v1) / Lambda

## AWS API Action

`CreateRestApi`, `CreateResource`, `PutMethod`, `PutIntegration`, `CreateDeployment`, then `_user_request_` invocation.

## Expected behavior

In a REST API, a request is routed to the **most specific resource that has a method matching the request** (an explicit method, or `ANY`). When the most specific resource that matches the *path* has **no method** for the request, real AWS API Gateway continues matching and falls through to a less specific resource — in particular the greedy `/{proxy+}` catch-all with `ANY`.

This is the standard multi-Lambda SAM pattern: one function owns `/{proxy+}` `ANY`, while other functions declare specific nested paths (e.g. `/items/{id}/attachments`). SAM materialises the intermediate parent resources (`/items`, `/items/{id}`) **without any method**. On real AWS:

```
GET /items/123              -> /items/{id} has no method  -> falls through to /{proxy+} ANY -> Lambda  ✅
GET /items/123/attachments  -> /items/{id}/attachments has no GET -> falls through to /{proxy+} ANY -> Lambda ✅
POST /items/123/attachments -> matches /items/{id}/attachments POST -> AttachmentsLambda ✅
```

## Actual behavior

Floci `1.5.27` stops at the most specific path match and returns `405 Method Not Allowed` instead of falling through to the greedy `/{proxy+}` `ANY` resource. The Lambda is never invoked.

```
GET /items/123  -> 405 Method Not Allowed   ({"message":"Method Not Allowed"})
GET /items      -> 405 Method Not Allowed
GET /status     -> reaches /{proxy+} ANY -> Lambda   (no method-less ancestor shadows it)
```

Because the method-less intermediate resource shadows `/{proxy+}`, **every** path whose most specific match is a method-less resource is unreachable, even though a valid `/{proxy+}` `ANY` proxy integration exists. The 405 is emitted by the gateway before any Lambda invocation (no invocation appears in the floci logs).

## Minimal reproduction (AWS SDK / CLI against floci)

```bash
EP="--endpoint-url http://localhost:4566"

RID=$(aws $EP apigateway create-rest-api --name fallthrough-demo --query id --output text)
ROOT=$(aws $EP apigateway get-resources --rest-api-id "$RID" --query "items[0].id" --output text)

# 1) Greedy catch-all: /{proxy+} ANY -> Lambda proxy integration
PROXY=$(aws $EP apigateway create-resource --rest-api-id "$RID" --parent-id "$ROOT" --path-part '{proxy+}' --query id --output text)
aws $EP apigateway put-method      --rest-api-id "$RID" --resource-id "$PROXY" --http-method ANY --authorization-type NONE
aws $EP apigateway put-integration --rest-api-id "$RID" --resource-id "$PROXY" --http-method ANY \
  --type AWS_PROXY --integration-http-method POST \
  --uri "arn:aws:apigateway:eu-west-1:lambda:path/2015-03-31/functions/arn:aws:lambda:eu-west-1:000000000000:function:MyFn/invocations"

# 2) A method-less concrete resource (as SAM creates for intermediate parents): /widgets  (NO method)
aws $EP apigateway create-resource --rest-api-id "$RID" --parent-id "$ROOT" --path-part widgets >/dev/null

# Deploy + create a reachable stage (on floci, create-deployment alone does not create
# a stage reachable via _user_request_, so create the stage explicitly).
DID=$(aws $EP apigateway create-deployment --rest-api-id "$RID" --stage-name bootstrap --query id --output text)
aws $EP apigateway create-stage --rest-api-id "$RID" --stage-name prod --deployment-id "$DID" >/dev/null

BASE="http://localhost:4566/restapis/$RID/prod/_user_request_"

# Shadowed by method-less /widgets.   Expected: fall through to /{proxy+} ANY -> MyFn.
# Actual (floci 1.5.27): 405 Method Not Allowed.
curl -s "$BASE/widgets"   # -> {"message":"Method Not Allowed"}   (HTTP 405)

# No method-less ancestor: correctly falls through to /{proxy+} ANY and invokes the integration.
curl -s "$BASE/other"     # -> {"message":"Function not found: MyFn"}   (reached the proxy integration)
```

`GET /widgets` should fall through to `/{proxy+}` `ANY` (→ `MyFn`); floci returns `405`. A request to a path with **no** method-less ancestor (`/other`) correctly reaches `/{proxy+}` (here a `404 Function not found`, because the demo `MyFn` isn't deployed — but the request was routed to the proxy integration rather than rejected with `405`).

## Likely source of the bug

The tiered resource matcher in
`src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java`
(the same area referenced by #792).

The matcher selects the most specific resource by **path** and then resolves the method on that single resource, returning `405` when the resource has no `ANY`/concrete method for the request. It does not continue to the next-best matching resource (the greedy `{proxy+}` tier) when the specific match lacks a matching method. Real AWS routing is *method-aware*: a path match with no usable method is skipped in favour of a less specific resource that does have one.

This is distinct from:
- #710 (`ANY` not matching concrete verbs **on the same resource** — fixed): here the shadowing resource has **no method at all**.
- #792 (sibling `{proxy+}` resources colliding — fixed): here there is a single `{proxy+}`, shadowed by a method-less **concrete/`{id}`** ancestor.

## Environment

- floci `1.5.27` (JVM/native image), `FLOCI_DEFAULT_REGION=eu-west-1`
- Deployed via AWS SAM (`samlocal`), REST API with implicit `/{proxy+}` `ANY` + nested specific paths owned by other Lambdas.

mansiverma897993:
…smatch (#1547)

Closes #1547

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -271,13 +271,17 @@ public class ApiGatewayExecuteController {
         ApiGatewayResource matched = null;
         MethodConfig method = null;
         for (ApiGatewayResource r : matchedResources) {
-            MethodConfig m = r.getResourceMethods().get(httpMethod.toUpperCase());
-            if (m == null) {
-                m = r.getResourceMethods().get("ANY");
-            }
-            if (m != null) {
-                matched = r;
-                method = m;
+            if (r.getResourceMethods() != null && !r.getResourceMethods().isEmpty()) {
+                MethodConfig m = r.getResourceMethods().get(httpMethod.toUpperCase());
+                if (m == null) {
+                    m = r.getResourceMethods().get("ANY");
+                }
+                if (m != null) {
+                    matched = r;
+                    method = m;
+                }
+                // Once we match a path that has methods configured, we must not fall back
+                // to less specific sibling resources (e.g. /{proxy+}), even on method mismatch.
                 break;
             }
         }
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -261,18 +261,28 @@ public class ApiGatewayExecuteController {
 
         // Find matching resource and method
         List<ApiGatewayResource> resources = apiGatewayService.getResources(region, apiId);
-        ApiGatewayResource matched = matchResource(resources, path);
-        if (matched == null) {
+        List<ApiGatewayResource> matchedResources = matchResources(resources, path);
+        if (matchedResources.isEmpty()) {
             return Response.status(404)
                     .entity(jsonMessage("Not Found"))
                     .type(MediaType.APPLICATION_JSON).build();
         }
 
-        MethodConfig method = matched.getResourceMethods().get(httpMethod.toUpperCase());
-        if (method == null) {
-            method = matched.getResourceMethods().get("ANY");
+        ApiGatewayResource matched = null;
+        MethodConfig method = null;
+        for (ApiGatewayResource r : matchedResources) {
+            MethodConfig m = r.getResourceMethods().get(httpMethod.toUpperCase());
+            if (m == null) {
+                m = r.getResourceMethods().get("ANY");
+            }
+            if (m != null) {
+                matched = r;
+                method = m;
+                break;
+            }
         }
-        if (method == null) {
+
+        if (matched == null) {
             return Response.status(405)
                     .entity(jsonMessage("Method Not Allowed"))
                     .type(MediaType.APPLICATION_JSON).build();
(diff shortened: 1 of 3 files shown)


hectorvent:
@mansiverma897993 Could you please take a look to the automated review comments?

mansiverma897993:
@hectorvent okk give me some time i will fix it and will raise PR  for other issues !!

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
```diff
@@ -261,18 +261,28 @@ Response dispatch(String httpMethod, String apiId, String stageName,
...
                     .entity(jsonMessage("Not Found"))
                     .type(MediaType.APPLICATION_JSON).build();
         }
 
-        MethodConfig method = matched.getResourceMethods().get(httpMethod.toUpperCase());
-        if (method == null) {
-            method = matched.getResourceMethods().get("ANY");
+        ApiGatewayResource matched = null;
+        MethodConfig method = null;
+        for (ApiGatewayResource r : matchedResources) {
```
abanna, line 274: @mansiverma897993 thanks for digging into the routing rules — this makes the matching logic much easier to follow. One behavior to double-check: once `/widgets` wins matching, this loop can fall back to a sibling `/{proxy+}` if `/widgets` does not have the method. For REST APIs that should be a 405 on the selected resource, not a different integration. Could we add a no-method case to keep that boundary clear?

mansiverma897993:
Hi! I have updated the codebase to import java.util.ArrayList and used the short form for both instantiations in matchResources to match the project's coding style conventions.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
abanna: I re-read the routing test and had conflated a resource with no methods with one that has methods but not this verb. The `/devices` case already pins down the 405 boundary I was looking for. Please ignore my earlier note—sorry for the noise.

mansiverma897993: No problem at all! Thanks for double-checking the boundary cases :)

hectorvent:

PR #1631, opened 2026-06-29
fix: S3 PutObjectRetention returns 400 instead of 500 on malformed RetainUntilDate

### LINKED ISSUE #1614: [BUG] S3: PutObjectRetention with a malformed RetainUntilDate returns 500 instead of 400
## Service

S3

## AWS API Action

PutObjectRetention

## Expected behavior

A `PutObjectRetention` whose `RetainUntilDate` can't be parsed as a date should
return a 4xx client error, never a 500.

**Confirmed against real AWS** (S3, ap-northeast-1). The AWS CLI / SDK reject a
non-date value *client-side* (`ParamValidation`), so this was verified by
sending the malformed value on the wire with a raw SigV4-signed request — the
AWS S3 *server* returns **HTTP 400 `MalformedXML`**:

```
HTTP 400
<Error><Code>MalformedXML</Code><Message>The XML you provided was not well-formed
or did not validate against our published schema</Message>...</Error>
```

Floci returns 500 for the same on-the-wire request.

## Actual behavior

Floci returns **HTTP 500** (an HTML "Internal Server Error" page, not an
AWS-shaped error) whenever `RetainUntilDate` is not a valid date-time. The
container log shows an uncaught `DateTimeParseException`:

```
ERROR ... HTTP Request to /<bucket>/<key>?retention&versionId=... failed ...:
java.time.format.DateTimeParseException: Text '1577836800' could not be parsed at index 0
    at io.github.hectorvent.floci.services.s3.S3Controller.handlePutObjectRetention(S3Controller.java:1566)
```

It reproduces with both an epoch-seconds value (`1577836800`) and an obviously
malformed value (`notadate`); a valid ISO-8601 value (`2030-01-01T00:00:00Z`)
returns 200. So any unparseable `RetainUntilDate` crashes the handler.

### Root cause

`services/s3/S3Controller.java:1566` parses `RetainUntilDate` with a strict
date-time parser and does not catch `DateTimeParseException`, so it surfaces as
a 500 `InternalFailure` instead of being mapped to a 400 client error. (Same
class of defect as the DynamoDB invalid-expression 500, `floci-io/floci#1512`,
and CloudWatch SetAlarmState, `floci-io/floci#1582` — an uncaught exception
leaking as 500.)

### Suggested fix (no code change in this issue)

Catch the parse failure and return an `AwsException` mapped to 400
(`MalformedXML` / `InvalidArgument`, matching what S3 returns) instead of letting
`DateTimeParseException` escape.

## Reproduction

Note: the `aws` CLI / SDKs validate `RetainUntilDate` *client-side* and won't
send a non-date, so reproduce with a raw HTTP request (Floci accepts a stub
SigV4 header):

```bash
B=http://localhost:4566
A='Authorization: AWS4-HMAC-SHA256 Credential=test/20260627/us-east-1/s3/aws4_request'

curl -s -o /dev/null -X PUT "$B/cov-probe-bkt" -H "$A"                 # bucket
curl -s -o /dev/null -X PUT "$B/cov-probe-bkt/obj1" -H "$A" --data hi  # object

# Malformed RetainUntilDate -> Floci 500 (AWS: 400 MalformedXML)
curl -s -w '\n%{http_code}\n' -X PUT "$B/cov-probe-bkt/obj1?retention" -H "$A" \
  -H 'Content-Type: application/xml' \
  --data '<Retention xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Mode>GOVERNANCE</Mode><RetainUntilDate>notadate</RetainUntilDate></Retention>'
```

(The same 500 also occurs with an epoch-seconds value like `1577836800`; a valid
ISO-8601 value like `2030-01-01T00:00:00Z` returns 200.)

## Environment

- Floci version / image tag: `floci/floci:1.5.28` (also on `main`,
  `S3Controller.java:1566`).
- Java SDK version (if applicable): reproduces via raw REST XML
  (`PUT /{bucket}/{key}?retention` with a `<Retention>` body).
- How you're running Floci: Docker (`floci/floci:1.5.28`).

mvanhorn:
## Summary

`PutObjectRetention` returned an HTTP 500 HTML "Internal Server Error" page when the `RetainUntilDate` could not be parsed, because `S3Controller.handlePutObjectRetention` called `Instant.parse` without catching the parse failure. Real AWS S3 returns HTTP 400 `MalformedXML` for the same request. This wraps the parse in a try/catch and throws an `AwsException` mapped to 400 `MalformedXML`, matching the existing S3 controller error pattern.

Closes #1614

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Before this change, a `RetainUntilDate` value that is not a valid ISO-8601 instant (for example epoch seconds `1577836800` or arbitrary text `notadate`) leaked an uncaught `DateTimeParseException` as a 500 HTML page. Real S3 returns 400 `MalformedXML`. The handler at `S3Controller.java:1565` now throws `AwsException("MalformedXML", ..., 400)`, which the existing `AwsExceptionMapper` renders as an AWS-shaped XML error. Valid ISO-8601 values keep returning 200.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -40,6 +40,7 @@ import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
@@ -1563,7 +1564,13 @@ public class S3Controller {
         String xml = new String(body, StandardCharsets.UTF_8);
         String mode = XmlParser.extractFirst(xml, "Mode", null);
         String dateStr = XmlParser.extractFirst(xml, "RetainUntilDate", null);
-        Instant retainUntil = dateStr != null ? Instant.parse(dateStr) : null;
+        Instant retainUntil;
+        try {
+            retainUntil = dateStr != null ? Instant.parse(dateStr) : null;
+        } catch (DateTimeParseException | IllegalArgumentException e) {
+            throw new AwsException("MalformedXML",
+                    "The XML you provided was not well-formed.", 400);
+        }
         boolean bypass = "true".equalsIgnoreCase(
                 httpHeaders.getHeaderString("x-amz-bypass-governance-retention"));
         s3Service.putObjectRetention(bucket, key, versionId, mode, retainUntil, bypass);
--- a/src/test/java/io/github/hectorvent/floci/services/s3/S3ObjectLockIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/s3/S3ObjectLockIntegrationTest.java
@@ -122,10 +122,83 @@ class S3ObjectLockIntegrationTest {
             .body(containsString("<Days>30</Days>"));
     }
 
+    // --- PutObjectRetention RetainUntilDate parsing ---
+
+    private static final String RETENTION_KEY = "retained-object.txt";
+
+    @Test
+    @Order(10)
+    void putObjectForRetention() {
+        given()
+            .body("hello")
+        .when()
+            .put("/" + LOCK_BUCKET + "/" + RETENTION_KEY)
+        .then()
+            .statusCode(200);
+    }
+
+    @Test
+    @Order(11)
+    void putObjectRetentionWithValidIsoDateReturns200() {
+        String body = """
+                <?xml version="1.0" encoding="UTF-8"?>
+                <Retention xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+                  <Mode>GOVERNANCE</Mode>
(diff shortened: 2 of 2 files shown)


hectorvent:
Nice fix, `MalformedXML` at 400 is exactly right for a timestamp that fails schema validation, and the test coverage (epoch seconds, garbage value, valid control) is solid. Two small things:

**The message is truncated.** The canonical S3 text, confirmed in moto (`moto/s3/exceptions.py:178`) and the S3 error code docs, is:

```
The XML you provided was not well-formed or did not validate against our published schema
```

The PR ships "The XML you provided was not well-formed." so clients comparing messages will notice. One string fix.

**Optional while you are in there:** an invalid `Mode` value (anything other than `GOVERNANCE` or `COMPLIANCE`) a few lines up still flows through silently, and real S3 returns the same `MalformedXML` for it. Same one line pattern if you want to include it, otherwise fine as a follow up.

mvanhorn:
Thanks for the kind review, @hectorvent - agreed a 400 MalformedXML is exactly what a bad RetainUntilDate deserves.

PR #1635, opened 2026-06-29
fix(cloudformation): register !Cidr and !GetAZs shorthand YAML tags

TAKEDA-Takashi:
## Summary

`Fn::Cidr` and `Fn::GetAZs` were added in #1365, but only in their long form. The matching YAML shorthand tags `!Cidr` and `!GetAZs` were never registered in `CloudFormationYamlParser`, so any template using the shorthand fails to parse:

```
could not determine a constructor for the tag !Cidr
```

The long form (`Fn::Cidr` / `Fn::GetAZs`) works fine — only the shorthand is affected. This registers the two missing tags alongside the other intrinsic shorthands (`!Ref`, `!Select`, `!Split`, …), so both forms behave identically.

```yaml
# Failed before this change, works after:
CidrBlock: !Select [0, !Cidr [!GetAtt Vpc.CidrBlock, 4, 8]]
```

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Both shorthand and long form are valid CloudFormation; the AWS CLI/SDK accept either. Added `CloudFormationYamlParserTest` verifying that `!Cidr` / `!GetAZs` expand to the same long-form map the engine already resolves (including the nested `!Select` / `!GetAtt` case above).

## Checklist

- [x] `./mvnw test` passes locally (`-Dtest=CloudFormationYamlParserTest`)
- [x] New or updated test added

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationYamlParser.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationYamlParser.java
@@ -56,6 +56,8 @@ public class CloudFormationYamlParser {
             register("!Base64",     n -> fnMap("Fn::Base64", n));
             register("!FindInMap",  n -> fnMap("Fn::FindInMap", n));
             register("!Split",      n -> fnMap("Fn::Split", n));
+            register("!Cidr",       n -> fnMap("Fn::Cidr", n));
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thanks @TAKEDA-Takashi, 
Solid fix

PR #1636, opened 2026-06-29
fix(iam): enforce STS session policies

snazy:
## Summary

STS session policies were recorded on temporary credentials but IAM enforcement evaluated only the caller's identity policies. As a result, an assumed role with broad role permissions could still perform an action even when the inline `AssumeRole` session policy should have denied or narrowed it.

This changes enforcement to evaluate the full caller context, including identity policies, session policies, and permission boundaries. It also keeps temporary-credential lookup account-independent so assumed-role sessions can be resolved after account routing moves the request into the role account, and it uses the resolved request account when building resource ARNs for policy evaluation.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

Incorrect behavior: STS `AssumeRole` accepted the standard inline `Policy` parameter, but later requests signed with those temporary credentials did not have that session policy applied by IAM enforcement.

The fix matches AWS policy evaluation semantics: explicit denies in session policies participate in the deny pass, and a present session policy must also explicitly allow the request for the role's permissions to be effective. This preserves the existing Floci bypass behavior for unknown local credentials and the `test` root credential.

## What changed

- `IamEnforcementFilter` resolves a `CallerContext` and evaluates the complete IAM context instead of identity-policy documents only.
- `IamService` resolves STS sessions across account namespaces for caller context, caller ARN, account routing, and temporary secret lookup.
- Enforcement resource ARN construction uses the request account already resolved by `AccountContextFilter`, so assumed-role credentials evaluate account-scoped resources against the role account.
- IAM and multi-account docs now describe session-policy intersection and temporary-credential resolution accurately.
- Java SDK compatibility coverage now checks that an assumed-role session policy deny overrides a role allow.

## Checklist

- [x] `./mvnw test` passes locally - attempted; only failures were unrelated LambdaReactiveSyncIntegrationTest container startup Broken pipe errors
- [x] Selected Java SDK compatibility tests pass locally - in `compatibility-tests/sdk-test-java`: `mvn -q -Dtest=IamEnforcementTest,IamTest,StsTest,S3Test test`

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/IamEnforcementTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/IamEnforcementTest.java
@@ -66,6 +66,7 @@ class IamEnforcementTest {
 
     private static final String DENY_S3_LIST_SESSION_POLICY = """
             {"Version":"2012-10-17","Statement":[
+              {"Effect":"Allow","Action":"s3:ListAllMyBuckets","Resource":"*"},
               {"Effect":"Deny","Action":"s3:ListAllMyBuckets","Resource":"*"}
             ]}""";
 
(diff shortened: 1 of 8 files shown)


snazy:
I've came across this one while migrating our use of Minio (and other test containers) in Nessie.
A local Floci image build with this change passed the existing Nessie tests (with minor error message assertion tweaks, which is fine).

snazy:
Thanks for the review!
Addressed the finding in a separate commit.

hectorvent (approved):
Thanks @snazy, 
Solid fix.

PR #738, opened 2026-05-07
Feat/automate docs gen

ericbsantana:
## Summary

The Supported Actions tables in `docs/services/*.md` were hand-maintained and drifted. This PR generates them from handler source instead. Based on this issue: https://github.com/floci-io/floci/issues/708

A Python script (as we are using `mkdocs` already that is built on Python) reads Java switch arms (`case "ActionName" ->`) or JAX-RS endpoint methods and rebuilds each table on every run. CI fails if the committed docs table don't match.

Covers all services but S3 is excluded. It also removes the description column on each `.md` documentation file.

## Why no description column

The original design included a two-column table with hand-written descriptions. The major reason I chose to drop is that AWS action names are self-documenting. When we search for `CreateSecret`, `ListFunctions`, `GetSecretValue` we either check AWS SDK docs or already know what these do.

The table is now action names only that are real handlers on the code. Floci-specific notes (partial implementations, unsupported variants, behavioral caveats) go in prose sections below the table - so any specific comment still done manually.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [x] Docs / chore

### New tooling
- `tools/docs/regen_action_docs.py` - single-file script (~165 lines). Two extraction modes: `switch` (Java switch arms `case "ActionName" ->`) and `rest` (JAX-RS `@GET/@POST/@PUT/@DELETE/@PATCH` methods inside `@Path` classes, PascalCased).
- `tools/docs/services.yaml` - registry of 38 services (37 with sources, S3 with `sources: []`).
- `tools/docs/test_regen_action_docs.py` + `tools/docs/fixtures/` - tests for `switch_basic`, `rest_basic`, `switch_multi_source`. Fixtures use synthetic Java files so unrelated handler PRs can't break them.
- `Makefile` targets: `docs-sync` (regen), `docs-test` (pytest), `docs-check` (CI gate).
- `.github/workflows/ci.yml` - new parallel `docs-sync` job: runs `docs-test` then `docs-check --strict` + `git diff --quiet docs/`.
- `CONTRIBUTING.md` - new "Documentation: Action Tables" section explaining how automated docs work.

### Doc migrations
- All 37 services migrated to marker-block format (`<!-- floci:actions:start/end -->`).
[...]

Issue #708, opened 2026-05-05
[FEAT] Auto-generate "Supported Actions" tables in MkDocs from handler source


ericbsantana:
## Service

Docs (all services).

## API Action / Feature

The "Supported Actions" tables in `docs/services/*.md` could be generated from the handler `switch` blocks instead of hand-written.

## AWS Documentation

n/a

## Why is this needed?

Right now whenever someone adds a handler, the doc table for that service has to be updated by hand. It's easy to forget. I just did it myself for Secrets Manager #698 and it made me wonder if there's a better way.

The action names already exist in code as `case "X" ->` in each `*Handler.java`. Could we just read them from there at build time and inject the table into the markdown? Something like a marker block in each service page that gets rewritten by a small script, plus a CI check so PRs fail if the docs don't match.

Just an idea - curious if this is something the project would want, or if it's been considered before.

## Are you willing to contribute a PR?

- [x] Yes
- [ ] No

hectorvent:
Hi @ericbsantana, 

If we do this will be so helpful. 
Assigning this to you. 

Thanks!

hectorvent:
@ericbsantana Closing PR because being inactive for long. Also no having the methods description as another column will be an issue. I am thinking to make an action that run a model and regenerate missing documentation and open a PR for review.

Thank you for you effort and time.

PR #1641, opened 2026-06-29
feat(docs): generate Supported Actions tables from handler source

hampsterx:
## What

Generates the **Supported Actions** tables in `docs/services/*.md` from handler source so they can't drift, while keeping the hand-written **Description** column.

Builds on #738 by @ericbsantana (closed) and resolves the open concern there: the description column is preserved, not dropped. The action **list** comes from code; the **descriptions** stay hand-written and are carried across regenerations, keyed by action name. Relates to #708.

## Why

The tables were maintained by hand and had drifted from the handlers. Examples this PR reconciles:

- `secrets-manager.md` was missing `RestoreSecret` (dispatched by `SecretsManagerJsonHandler`).
- `acm.md` listed `RenewCertificate`, which is not implemented anywhere in `src/` (removed).
- Several services were missing actions that exist in code (e.g. `ImportCertificate` for ACM).

## How

- `tools/docs/regen_action_docs.py` extracts actions from `case "X" ->` switch arms (including multi-label arms like `case "A", "B" ->`) and JAX-RS REST controllers. It rewrites only the rows between `<!-- floci:actions:start -->` and `<!-- floci:actions:end -->`; every byte outside the markers is untouched. Existing descriptions are preserved by action name; new actions render with a `-` placeholder. Matching is PascalCase-only, so `SCREAMING_CASE` labels from non-action switches (state enums, etc.) are not picked up.
- `tools/docs/services.yaml` registers the services covered so far, and lists every other switch handler under `deferred_handlers`, so the "new unregistered handler" check fires only on a genuinely new service.
- `make docs-sync` (regenerate), `make docs-check` (CI gate: regenerate, then fail on any diff or unregistered handler), `make docs-test` (unit tests).
- `.github/workflows/docs-actions.yml`: a Python-only job that runs on handler, doc, and tooling changes.
- A `CONTRIBUTING.md` note documenting the generate-and-commit workflow.

## Scope

This first cut covers single-table, switch-mode services (18). The following are intentionally deferred and tracked in `deferred_handlers` with reasons, to keep the PR reviewable:

- REST-mode services (need per-controller name overrides)
- mixed-mode (SES)
- multi-table docs (DynamoDB, SNS, SSM, CloudWatch, ...)
- non-tabular docs (IAM, EC2, ECS, Cognito, Glue, ...)
- S3 (query-string dispatch)
- ECR (doc packs several actions per row)

Each can be added incrementally: one entry in `services.yaml` plus marker insertion, no script changes.

## Test plan

- [x] `make docs-test`, 18 unit tests (switch single + multi-label, REST + rename/exclude, description preservation, orphan drop, escaped-pipe round-trip, malformed-table rejection, idempotency)
- [x] `make docs-check` clean on a fresh checkout; regeneration is idempotent (second run is a no-op)
- [x] Verified each registered doc's row count matches its handler's action count
- [x] Confirmed no bytes change outside the marker blocks


ideas/auto-generated-action-tables.spec.md
hectorvent, line 1: @hampsterx is this file required?

hampsterx: No, it was a planning artifact, not needed. Removed in 3a3e70c6 along with the four references to it (CONTRIBUTING.md, Makefile, the generator docstring, and services.yaml). The generate-and-commit workflow is documented in CONTRIBUTING.md, and the #738/@ericbsantana credit stays in the PR description.

hectorvent:
Really nice piece of tooling, and the drift it fixes is real. I verified the claims independently: `RestoreSecret` is indeed dispatched but was missing from the docs, `RenewCertificate` appears nowhere in `src/main`, and as a spot check I extracted the case labels from both SQS handlers and diffed them against the regenerated table, exact 23 for 23 match in both directions. The marker discipline, description preservation and the `deferred_handlers` sentinel are all thoughtful touches, and the CI gate with `make docs-sync` messaging is exactly the right shape.

Two tiny notes:

The new workflow pins `actions/checkout@v6.0.2`, but the repo moved every workflow to `v7.0.0` this week (#1690), so starting at v7 avoids dependabot immediately flagging the new file.

Worth one sentence in the generator docstring: a case label is treated as supported, so an arm that exists but explicitly throws `UnsupportedOperation` in its body would still be listed. Rare pattern here since handlers throw from `default`, but naming the assumption saves a future head scratch.

Phase 1 scope is sensible and honestly declared. Looks ready once the checkout pin is bumped.

hampsterx:
@hectorvent both addressed in 19fbb617: checkout pinned to v7.0.0, and the generator docstring now spells out the case-label-means-supported assumption (a per-action `UnsupportedOperation` stub would be a false positive; handlers here reject from `default` so it holds). Ready for another look.

hectorvent (approved):
Thanks @hampsterx

PR #1651, opened 2026-06-30
feat(kms): support KeyId filter on ListAliases

### LINKED ISSUE #1650: [FEAT] KMS: ListAliases should support filtering by KeyId
## Description

The AWS KMS `ListAliases` API accepts an optional `KeyId` parameter to return only the aliases associated with a specific key (`aws kms list-aliases --key-id <key-id>`).

Currently Floci ignores the `KeyId` parameter and always returns all aliases for the region.

## Expected behavior

When `KeyId` is provided, only aliases whose `TargetKeyId` matches the resolved key should be returned. The parameter should accept a plain key UUID, a full key ARN, or an alias ARN — matching AWS SDK behavior.

## Actual behavior

The `KeyId` parameter is silently ignored and all aliases in the region are returned.

mikelamutxastegi:
## Summary

The AWS KMS `ListAliases` API accepts an optional `KeyId` parameter to return only the aliases associated with a specific key — used by `aws kms list-aliases --key-id <key-id>`. Floci was ignoring the parameter and always returning all aliases for the region.

This change wires `KeyId` through the handler into a new `listAliases(String keyId, String region)` overload on `KmsService`. The parameter accepts a plain key UUID or a full key ARN, matching AWS behavior. Alias name and alias ARN forms are also accepted as intentional leniency beyond what AWS supports, consistent with how Floci resolves KeyId in other KMS operations. When `KeyId` is provided the service filters the result to only aliases whose `TargetKeyId` matches the resolved key. When `KeyId` is omitted the existing behavior is preserved.

Three new unit tests cover: filtering by plain key UUID, filtering by full key ARN, and returning an empty list for a key with no aliases.

Closes #1650

## Type of change

* [ ] Bug fix (`fix:`)
* [x] New feature (`feat:`)
* [ ] Docs / chore

## AWS Compatibility

Previous behavior:

`ListAliases` silently ignored the `KeyId` parameter and returned all aliases in the region, regardless of which key they were associated with.

Updated behavior:

When `KeyId` is provided, `ListAliases` returns only aliases whose `TargetKeyId` matches the resolved key. The parameter accepts a plain key UUID, a full key ARN, or an alias ARN — consistent with how other KMS operations resolve `KeyId`. When `KeyId` is omitted, all aliases for the region are returned as before.

## Checklist

* [x] `./mvnw test` passes locally
* [x] New unit tests added (`listAliasesFilteredByKeyId`, `listAliasesFilteredByKeyIdWithArn`, `listAliasesFilteredByKeyIdReturnsEmptyWhenNoAliases` in `KmsServiceTest`)
* [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsJsonHandler.java
@@ -384,7 +384,8 @@ public class KmsJsonHandler {
     }
 
     private Response handleListAliases(JsonNode request, String region) {
-        String keyId = request.path("KeyId").isMissingNode() ? null : request.path("KeyId").asText(null);
+        JsonNode keyIdNode = request.path("KeyId");
+        String keyId = (keyIdNode.isMissingNode() || keyIdNode.isNull()) ? null : keyIdNode.asText();
         List<KmsAlias> aliases = service.listAliases(keyId, region);
         ObjectNode response = objectMapper.createObjectNode();
         ArrayNode array = response.putArray("Aliases");
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsJsonHandler.java
@@ -384,7 +384,8 @@ public class KmsJsonHandler {
     }
 
     private Response handleListAliases(JsonNode request, String region) {
-        List<KmsAlias> aliases = service.listAliases(region);
+        String keyId = request.path("KeyId").isMissingNode() ? null : request.path("KeyId").asText(null);
+        List<KmsAlias> aliases = service.listAliases(keyId, region);
         ObjectNode response = objectMapper.createObjectNode();
         ArrayNode array = response.putArray("Aliases");
         for (KmsAlias a : aliases) {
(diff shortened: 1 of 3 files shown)


hectorvent:
Verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/kms/2014-11-01/service-2.json) (kms/2014-11-01) and the semantics are right: `KeyId` optional, omit returns everything, key ID or ARN accepted, and `NotFoundException` for an unknown key is a declared error for this operation. Nice focused feature.

Two things now that #1648 merged this morning:

The branch needs a rebase, it currently conflicts with main since both changes touch adjacent code in `KmsService`. Good news is the merge order worked in your favor: with #1648's normalization in place, your filter comparison is correct by construction for aliases created with ARNs too.

While rebasing, one small test would lock the two changes together permanently:

```java
kmsService.createAlias("alias/by-arn", key.getArn(), REGION);
List<KmsAlias> filtered = kmsService.listAliases(key.getKeyId(), REGION);
assertEquals(1, filtered.size());
```

That guards the normalize on write plus compare on filter invariant so neither side can regress independently.

One wording note: the PR body describes alias ARN acceptance for `KeyId` as compatibility, but the model restricts the parameter to key ID or key ARN, real AWS rejects alias forms here. Fine to keep the leniency, just worth describing it as intentional.

mikelamutxastegi:
@hectorvent Thanks for the detailed review!

I've rebased the branch and added the test you suggested.

You're right that AWS only supports the key ID or key ARN, so I've updated the PR description accordingly.

hectorvent (approved):
Thanks @mikelamutxastegi

PR #1659, opened 2026-06-30
fix(stepfunctions): resolve qualified Lambda function ARNs in lambda:invoke

abanna:
## Summary

The Step Functions `lambda:invoke` integration resolved the function name by taking the last
`:`-segment of the reference. For a qualified ARN like
`arn:aws:lambda:<region>:<acct>:function:NAME:$LATEST` that yields the qualifier (`$LATEST`)
instead of `NAME`, so the invoke failed with `Lambda function not found: $LATEST`.

This adds `extractLambdaFunctionName()`, which strips the `:function:` prefix and an optional
trailing version/alias qualifier, and routes both the direct-ARN and optimized-integration call
sites through it. The function store is keyed by name, so the qualifier must be dropped.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

Step Functions' optimized `lambda:invoke` accepts a `FunctionName` that is a bare name, a name with
a `:version`/`:alias` qualifier, or a full/partial function ARN (optionally qualified). All of these
must resolve to the bare function name. Verified against bare names, qualified names, partial ARNs,
and full qualified ARNs (incl. `$LATEST`, numeric versions, and named aliases).

## Checklist

- [x] `./mvnw test` passes locally — `AslExecutorLambdaArnTest` (run in `eclipse-temurin:25-jdk`)
- [x] New test added — `AslExecutorLambdaArnTest` (9 cases)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -302,19 +302,43 @@ public class AslExecutor {
         }
     }
 
+    /**
+     * Extracts the Lambda function name from a reference that may be a bare name, a name with a
+     * version/alias qualifier (e.g. "name:$LATEST"), or a full/partial function ARN
+     * (e.g. "arn:aws:lambda:region:acct:function:name[:qualifier]"). The qualifier is dropped
+     * because the function store is keyed by name. Taking the last ':'-segment is wrong for a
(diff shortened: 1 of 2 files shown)


abanna:
Closing as a duplicate of #1660 (identical change — same `extractLambdaFunctionName` fix and `AslExecutorLambdaArnTest`). Two PRs for this fix were opened in parallel by mistake; consolidating on #1660.

PR #1662, opened 2026-06-30
feat(stepfunctions): implement States.JsonMerge intrinsic

abanna:
## Summary

Adds the `States.JsonMerge` intrinsic function to the Step Functions executor. Previously any state using it failed with `Unsupported intrinsic function: States.JsonMerge`.

- Implements the **shallow merge** form `States.JsonMerge($.a, $.b, false)`: the result is `a`'s top-level fields with `b`'s top-level fields merged in, and **`b` wins on a key conflict**.
- **Rejects** the deep-merge form (third argument `true`) and non-object arguments, matching AWS (which only supports `false`).
- Also resolves `true`/`false`/`null` **literal** intrinsic arguments in `resolveIntrinsicArg`, so the third argument evaluates to a real boolean rather than falling through to path resolution.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

`States.JsonMerge(json1, json2, isDeepMerge)` performs a shallow merge of two JSON objects; the only supported `isDeepMerge` value is `false`, and conflicting top-level keys take the value from the second object. Deep merge (`true`) is rejected. Verified against the documented AWS intrinsic behavior and covered by `AslExecutorJsonMergeTest`.

## Checklist

- [x] `./mvnw test` passes locally (`AslExecutorJsonMergeTest`: 4 tests; sibling `AslExecutorCatchTest` regression green — run in an `eclipse-temurin:25-jdk` container)
- [x] New or updated integration test added (`AslExecutorJsonMergeTest`)


abanna:
Applied the Greptile P2 on error precedence: the object-type check now runs before the deep-merge-flag rejection, so `States.JsonMerge` with two non-object arguments and `true` reports `requires two JSON objects` (type validation first) rather than `shallow merge only`, matching AWS's ordering. Added a precedence test. Branch is up to date with `main` (merge) and green.

abanna:
Merged current `main` in and verified green.

Both greptile points are already handled in the current implementation:
- **Non-boolean 3rd arg** is rejected (`if (!deepArg.isBoolean()) throw …`) rather than coerced to `false`.
- **Argument-type check precedes the deep-merge flag** (`!a.isObject() || !b.isObject()` throws "requires two JSON objects" before the `deep` rejection), matching AWS error ordering — pinned by `nonObjectArgumentsAreRejectedBeforeTheDeepMergeFlag`.

The merge brought `AslExecutor`'s constructor to 13 args, so the test now passes `EcsService`/`EcsJsonHandler` mocks. `AslExecutorJsonMergeTest` — 6 tests, green (maven, temurin-25).

hectorvent:
Thanks for this, the `States.JsonMerge` logic looks good and the shallow-merge semantics match the reference behavior nicely.

One thing before it can go in: could you rebase on latest `main`? The branch predates the change that added a `context` argument to the intrinsic pipeline, so `resolveIntrinsicArg`/`evaluateIntrinsic` now take three parameters. GitHub still shows the PR as mergeable (it merges textually), but after merging `main` it no longer compiles:

```
AslExecutor.java: method resolveIntrinsicArg cannot be applied to given types;
  required: String, JsonNode, JsonNode
  found:    String, JsonNode
```

The fix is just threading `context` through the new `States.JsonMerge` block (the same `root, context` the surrounding cases already pass). Happy to help if anything is unclear. Thanks again!

abanna:
Done — thanks for the clear pointer. I brought the branch up to date with `main` and threaded `context` through the three `States.JsonMerge` argument resolutions (the same `root, context` the surrounding cases pass), so it compiles against the new three-parameter intrinsic pipeline. `AslExecutorJsonMergeTest` is green, and the branch is current with `main`.

I merged `main` in rather than rebasing, to avoid rewriting the already-pushed branch history.

hectorvent (approved):
Thanks @abanna

PR #1663, opened 2026-06-30
fix(iam): resolve attached AWS-managed policies for any account

abanna:
## Summary

Follow-up to #1573 (AWS-managed policies resolvable from any account context).

AWS-managed policies (`arn:aws:iam::aws:policy/...`) are global and live in the in-memory catalog, not the account-partitioned policy store. `getPolicy` already serves them from the catalog, but several IAM **read paths** resolved attached-policy ARNs straight from the account-scoped store (`policies.get(arn)` / the `flatMap` variants). As a result, a managed policy **attached to a principal owned by a non-default account** was silently dropped from the response.

This adds a catalog-aware `resolvePolicy(arn)` helper (modeled on `getPolicy`) and routes the affected reads through it:

- `listAttachedUserPolicies` / `listAttachedGroupPolicies` / `listAttachedRolePolicies`
- the user/role permissions-boundary document resolvers
- `collectUserPolicies` / `collectRolePolicies` (the `SimulatePrincipalPolicy` / caller-context paths)

Customer-policy-only logic and attachment-count bookkeeping on detach are left untouched.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

For bug fixes: an AWS-managed policy attached to a user/group/role in a non-default account was not returned by the attached-policy read APIs (and not evaluated by `SimulatePrincipalPolicy`), because those paths only consulted the account-scoped customer-policy store. Managed policies are global, so they are now resolved from the catalog regardless of the caller's account.

## Checklist

- [x] `./mvnw test` passes locally (`IamManagedPolicyAccountScopeTest`: 6 tests; `IamServiceTest`: 57 tests — run in an `eclipse-temurin:25-jdk` container)
- [x] New or updated integration test added (`IamManagedPolicyAccountScopeTest`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
@@ -439,6 +439,20 @@ public class IamService implements SessionAccountLookup {
                         "Policy " + policyArn + " does not exist.", 404));
     }
 
+    /**
+     * Resolves a policy by ARN without throwing, mirroring {@link #getPolicy} so that
+     * AWS-managed policies (arn:aws:iam::aws:policy/...) are served from the global catalog
+     * rather than the account-partitioned store. Attached-policy read paths must use this:
+     * a managed policy attached to a principal owned by a non-default account is absent from
+     * that account's {@link #policies} partition and would otherwise be silently dropped.
+     */
+    private Optional<IamPolicy> resolvePolicy(String arn) {
+        if (arn != null && arn.startsWith(AwsManagedPolicies.ARN_PREFIX)) {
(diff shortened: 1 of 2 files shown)


abanna:
Applied the Greptile P2 — added the group analog of the existing user/role tests: `attachedManagedPolicyResolvesForGroupInNonDefaultAccount` exercises `listAttachedGroupPolicies` (the `resolvePolicy` catalog lookup) for a group owned by a non-default account, with a customer policy attached to the same group as the account-scoping control. Branch is up to date with `main` (merge) and green.

hectorvent (approved):
Thanks @abanna

PR #1666, opened 2026-06-30
perf(lambda): drop the base64 encode-decode round-trip on the S3 code path

abanna:
## Summary

For S3-sourced function code, `extractZipCodeFromS3` (and the S3 code-update path) called
`extractZipCode(fn, Base64.getEncoder().encodeToString(obj.getData()))` — base64-encoding the fetched
bytes purely so `extractZipCode` could immediately base64-**decode** them again. That is a wasted
CPU + allocation of a String ~1.33x the (often tens-of-MB) zip on every create/update. This splits
the extraction into a `extractZipCodeBytes(LambdaFunction, byte[])` core, with `extractZipCode(fn, base64)`
becoming a thin decode-and-delegate shim; the S3 paths now pass the bytes straight through.

## Type of change

- [x] Performance (`perf:`)

## AWS Compatibility

Behavior-identical — `decode(encode(x)) == x`, so the extracted code, size, and SHA-256 are unchanged;
only the redundant round-trip is removed. The direct (request-supplied base64) upload path is unchanged.

## Checklist

- [x] `./mvnw test` compiles + Lambda code tests pass locally (eclipse-temurin:25-jdk)
- [x] No new test (behavior-neutral refactor; existing Lambda code tests cover extraction)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -1135,7 +1135,10 @@ public class LambdaService {
     }
 
     private void extractZipCode(LambdaFunction fn, String zipFileBase64) {
-        byte[] zipBytes = Base64.getDecoder().decode(zipFileBase64);
+        extractZipCodeBytes(fn, Base64.getDecoder().decode(zipFileBase64));
+    }
+
+    private void extractZipCodeBytes(LambdaFunction fn, byte[] zipBytes) {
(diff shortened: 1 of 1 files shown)


abanna:
Brought up to date with `main` (merge). No review changes needed — Greptile rated it 5/5 with no findings, and the behavior-neutral refactor is confirmed by `LambdaServiceTest` (40 tests green in the build).

hectorvent (approved):
Solid fix.

PR #1677, opened 2026-07-01
feat(apigateway): persist API key tags and add GetApiKey route

### LINKED ISSUE #1676: [BUG] apigateway: CreateApiKey drops "tags" and GetApiKey is not implemented (404)
## Service
API Gateway (REST)

## AWS API Action
`CreateApiKey`, `GetApiKey`, `GetApiKeys`

## Expected behavior
- `CreateApiKey` persists the `tags` passed in the request, and they are returned by `CreateApiKey`, `GetApiKey` and `GetApiKeys`.
- `GetApiKey` (`GET /apikeys/{apiKeyId}`) returns the API key as JSON, or a JSON `NotFoundException` when it doesn't exist.

## Actual behavior
- `CreateApiKey` silently drops `tags`: they are not stored and never appear in any response.
- `GetApiKey` is not routed --> HTTP 404 with an **XML** body, so the AWS SDK v3 (REST-JSON) fails to parse it: `SyntaxError: Unexpected token '<', "<?xml vers"... is not valid JSON`.
- `GetApiKeys` returns items with only `id, name, value, enabled`. No `tags`.

## Reproduction

AWS CLI:
```bash
EP="--endpoint-url http://localhost:4566 --region us-east-1"

# 1) Create a key with tags
aws $EP apigateway create-api-key --name k --enabled --tags Team=platform,Project=demo
# -> response contains id/name/value/enabled but NO "tags"

# 2) Read it back by id (use the id returned above)
aws $EP apigateway get-api-key --api-key <API_KEY_ID>
# -> HTTP 404 with an XML body (expected: the key, including its tags)

# 3) List keys
aws $EP apigateway get-api-keys
# -> items contain no "tags"
```

TypeScript (`@aws-sdk/client-api-gateway` v3):
```ts
import {
  APIGatewayClient, CreateApiKeyCommand, GetApiKeyCommand, GetApiKeysCommand
} from '@aws-sdk/client-api-gateway';

const client = new APIGatewayClient({
  endpoint: 'http://localhost:4566',
  region: 'us-east-1',
  credentials: { accessKeyId: 'test', secretAccessKey: 'test' }
});

// 1) Create a key with tags
const created = await client.send(new CreateApiKeyCommand({
  name: 'k', enabled: true, tags: { Team: 'platform', Project: 'demo' }
}));
console.log(created.tags); // undefined,  expected { Team: 'platform', Project: 'demo' }

// 2) Read it back by id
await client.send(new GetApiKeyCommand({ apiKey: created.id!, includeValue: false }));
// throws: HTTP 404 + "<?xml ...>" body → SyntaxError: Unexpected token '<'

// 3) List keys
const list = await client.send(new GetApiKeysCommand({}));
console.log(list.items?.[0]); // { id, name, value, enabled } , no tags
```

## Environment
- Floci version / image tag: `floci/floci:1.5.28`
- How you're running Floci: Docker (`docker compose`, image `floci/floci:latest`)

awuzi:
## Summary

Fixes the two API Gateway API-key gaps reported in #1676:

- `CreateApiKey` now persists the request `tags` and returns them from `CreateApiKey`, `GetApiKey` and `GetApiKeys`.
- Adds the missing `GetApiKey` route (`GET /apikeys/{apiKeyId}`), delegating to the existing `service.getApiKey(...)`. It omits `value` unless `includeValue=true`, and returns a JSON `NotFoundException` (404) for unknown keys.

## Changes

- `model/ApiKey.java` — add a `tags` map (mirrors `RestApi`).
- `ApiGatewayService#createApiKey` — read `tags` from the request (same pattern as `createRestApi`).
- `ApiGatewayController` — serialize `tags` in `toApiKeyNode`; add the `GET /apikeys/{apiKeyId}` route.
- `ApiGatewayApiKeyIntegrationTest` — covers tag persistence, GetApiKey (with/without `value`), listing, and 404.

## Notes

Backward-compatible: keys persisted before this change deserialize with an empty `tags` map (`@JsonIgnoreProperties(ignoreUnknown = true)` + default field value).

Closes #1676

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
@@ -613,12 +613,19 @@ public class ApiGatewayController {
 
     @GET
     @Path("/apikeys")
-    public Response getApiKeys(@Context HttpHeaders headers) {
+    public Response getApiKeys(@Context HttpHeaders headers,
+                               @QueryParam("includeValues") boolean includeValues) {
         String region = regionResolver.resolveRegion(headers);
         List<ApiKey> keys = service.getApiKeys(region);
         ObjectNode root = objectMapper.createObjectNode();
         ArrayNode items = root.putArray("item");
-        keys.forEach(k -> items.add(toApiKeyNode(k)));
+        keys.forEach(k -> {
+            ObjectNode node = toApiKeyNode(k);
+            if (!includeValues) {
+                node.remove("value");
+            }
+            items.add(node);
+        });
         return Response.ok(root.toString()).type(MediaType.APPLICATION_JSON).build();
     }
 
@@ -1655,8 +1662,8 @@ public class ApiGatewayController {
         node.put("name", k.getName());
         node.put("value", k.getValue());
         node.put("enabled", k.isEnabled());
-        ObjectNode tags = node.putObject("tags");
-        if (k.getTags() != null) {
+        if (k.getTags() != null && !k.getTags().isEmpty()) {
+            ObjectNode tags = node.putObject("tags");
             k.getTags().forEach(tags::put);
         }
         return node;
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
@@ -1655,8 +1655,8 @@ public class ApiGatewayController {
         node.put("name", k.getName());
         node.put("value", k.getValue());
         node.put("enabled", k.isEnabled());
-        if (k.getTags() != null && !k.getTags().isEmpty()) {
-            ObjectNode tags = node.putObject("tags");
+        ObjectNode tags = node.putObject("tags");
+        if (k.getTags() != null) {
             k.getTags().forEach(tags::put);
         }
         return node;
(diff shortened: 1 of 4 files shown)


hectorvent:
Thanks @awuzi, nice focused contribution. Verified against `service-2.json` (both the [botocore](https://github.com/boto/botocore/blob/develop/botocore/data/apigateway/2015-07-09/service-2.json) and [aws-sdk-java-v2](https://github.com/aws/aws-sdk-java-v2/blob/master/services/apigateway/src/main/resources/codegen-resources/service-2.json) copies): the route shape, the `includeValue` query param and its omit by default behavior, the `tags` map casing, and the declared `NotFoundException` are all exact. The backward compatibility note checks out too.

Two small things I would like fixed in this PR before merge, both cheap:

1. The 404 message. Real AWS returns `Invalid API Key identifier specified` (confirmed in moto's `get_api_key`, which LocalStack also uses), while our pre existing service message says `API Key not found`. That part is not your bug, but the new test pins the wrong string, which makes it painful to correct later. Since you are already here, aligning both is two lines:

```java
// ApiGatewayService.getApiKey
new AwsException("NotFoundException", "Invalid API Key identifier specified", 404)
// test
.body("message", equalTo("Invalid API Key identifier specified"))
```

2. The list route. `GetApiKeys` has the plural `includeValues` query param in the model, and real AWS omits every key's `value` from the list unless it is true. Our existing list route serializes values unconditionally, so secrets leak by default. You already implemented exactly the right stripping for the singular route ten lines below it, so this is the same pattern once more:

```java
@QueryParam("includeValues") boolean includeValues
// per key: if (!includeValues) node.remove("value");
```

Optional, take it or leave it: emit `tags` only when non empty, since rest json omits absent members (SDKs tolerate the empty map fine).

With those two in, this is good to merge. Thanks again!

awuzi:
Hi! I’ll address both points and update the PR 👍🏼

awuzi:
@hectorvent all addressed, let me know if they meet your requirements or if you need any more adjustments

hectorvent (approved):
Thanks @awuzi

PR #1569, opened 2026-06-25
feat(ecs): mount efsVolumeConfiguration task volumes as shared local volumes

abanna:
Closes #1568.

## Summary

ECS task definitions can declare EFS-backed task volumes via `efsVolumeConfiguration`, but Floci modelled only the EC2 `host` volume shape. An EFS volume’s `mountPoint` resolved to no source path and was skipped, so the container launched with **no filesystem at the mount path** — and the `efsVolumeConfiguration` block was dropped from `RegisterTaskDefinition`/`DescribeTaskDefinition`, causing Terraform drift.

This teaches the ECS layer to honor EFS volumes. Since Docker cannot mount a real Amazon EFS file system, an EFS-backed volume is materialised as a shared local Docker **named volume** `floci-efs-<fileSystemId>`, mounted at the container path (read-only/read-write per the `mountPoint`). Containers that mount the same file system share persistent storage that survives task restarts — the local stand-in for an EFS mount, mirroring the documented LocalStack EFS-simulation pattern but applied automatically at the ECS layer.

## What changed

- **`Volume` model** — carries an optional `EfsVolumeConfiguration` alongside the existing host source path (back-compatible 2-arg constructor retained).
- **New `EfsVolumeConfiguration`** — `fileSystemId`, `rootDirectory`, `transitEncryption`, `transitEncryptionPort`, `authorizationConfig` (`accessPointId`, `iam`).
- **`EcsJsonHandler`** — parses `efsVolumeConfiguration` in `RegisterTaskDefinition` and serializes it back in `DescribeTaskDefinition` (full round-trip, no drift).
- **`EcsContainerManager`** — resolves an EFS volume’s `mountPoint` to a named volume `floci-efs-<fileSystemId>` via `withNamedVolume(...)`; host volumes keep their existing bind-mount behavior; truly-unresolved mountPoints still warn and skip.
- **`ContainerBuilder`** — adds a read-only overload `withNamedVolume(name, path, readOnly)`.

`rootDirectory`/`transitEncryption`/`transitEncryptionPort`/`authorizationConfig` are modelled for API round-trip but have no local effect.

## Testing

- `mvn test -Dtest="Ecs*,ContainerBuilder*"` green (65 tests).
- New unit tests: EFS volume → shared named volume by access mode (`EcsContainerManagerVolumesTest`); `efsVolumeConfiguration` JSON round-trip (`EcsJsonHandlerVolumesTest`).
[...]

PR #1686, opened 2026-07-02
feat(ecs): apply EFS access point POSIX ownership to shared local volumes

abanna:
## What

Emulate an Amazon EFS **access point's POSIX ownership** for the shared local Docker volumes that stand in for EFS file systems (introduced in #1569).

## Why

#1569 materialises an ECS task's `efsVolumeConfiguration` volume as a shared local Docker named volume (`floci-efs-<fileSystemId>`). Docker creates a named volume `root:root 0755`, so a task whose **image runs as a non-root `USER`** — as workloads mounting an EFS access point typically do — cannot create files on it and fails with `EACCES: permission denied, mkdir ...`.

Real EFS initialises an access point's root directory per `RootDirectory.CreationInfo {OwnerUid, OwnerGid, Permissions}` and applies the access point's `PosixUser` to all I/O regardless of the container process uid. This change emulates that mechanism for the local stand-in volume.

## How

All behaviour is **opt-in and driven by generic config** — defaults are empty/false, so a shared volume behaves exactly as before unless configured (zero change for existing users):

- **`floci.storage.efs.*`** — `owner-uid`, `owner-gid`, `root-permissions`, `setgid`, `init-image`, `mount-user`, `mount-group-add`.
- **`ContainerLifecycleManager.ensureSharedVolume(...)`** — ensures the volume, then (if ownership is configured) `chown`/`chmod`s the mount root once via a short-lived helper container (run-once guard; optional setgid so subdirectories inherit the owner gid). No-op — a plain `ensureVolume` — when unconfigured.
- **`ContainerSpec`/`ContainerBuilder`** gain `withUser`/`withGroupAdd`, wired into `createContainerCmd.withUser(...)` and `HostConfig.withGroupAdd(...)`.
- **`EcsContainerManager`** initialises the EFS volume root via `ensureSharedVolume` before mounting it.

## Testing

`ContainerBuilderTest`, `ContainerLifecycleManagerVolumeTest`, `EcsContainerManagerVolumesTest` cover:
- the **no-op default** (no helper container spun up, no image pulled) — guards against any behaviour change for existing users;
- the one-off `chown`/`chmod`/`setgid` helper and its run-once guard;
- the builder round-trip of `withUser`/`withGroupAdd`;
- `EcsContainerManager` invoking `ensureSharedVolume` with the configured ownership.

Follow-on to #1569.


abanna:
Applied the Greptile review points:

- **P1 — `mount-user`/`mount-group-add` were inert.** Wired them into the ECS container spec: a container mounting an EFS-backed shared volume now runs under the configured `uid[:gid]` (`withUser`) and adds the supplementary group (`withGroupAdd`), emulating the access point's `PosixUser` squashing. Without this a non-root image `USER` couldn't write a volume owned by `owner-uid`/`owner-gid`. Added `EcsContainerManagerVolumesTest#efsMountAppliesConfiguredPosixUserAndGroupToContainer`.
- **P2 — `application.yml`.** Added the `floci.storage.efs.*` section (with the `@WithDefault` values and the opt-in keys documented) so the effective defaults are discoverable.
- **P2 — `root-permissions` Javadoc.** Corrected: leaving it empty skips only the permission `chmod`; the init helper still runs if `owner-uid`/`owner-gid`/`setgid` are set. Volume-root init is skipped entirely only when all four are at their defaults.

Branch is up to date with `main` (merge); green in the build (66 tests incl. the ECS integration suite, which boots the app with the new YAML).

hectorvent:
@abanna Thanks for the PR. Please review the comments by the greptile-apps bot and resolve conflicts.

Reviewed the EFS emulation against `service-2.json` ([efs](https://github.com/aws/aws-sdk-java-v2/blob/master/services/efs/src/main/resources/codegen-resources/service-2.json) and [ecs](https://github.com/aws/aws-sdk-java-v2/blob/master/services/ecs/src/main/resources/codegen-resources/service-2.json) models in the AWS SDK source). No wire contract issues since this adds no API surface, but a few things to address:

**1. Partial ownership config produces a broken `chown` that is then never retried.** In the model `CreationInfo.OwnerUid`, `OwnerGid` and `Permissions` are all REQUIRED together; a partial CreationInfo does not exist in AWS. With only `owner-uid` set the helper runs:

```sh
chown 1001: /floci-shared-volume
```

The trailing colon makes chown resolve the login group from `/etc/passwd`, which fails in busybox for a uid not present there. The helper exits nonzero, which only logs a warning, and `initializedSharedVolumes` is rolled back only in the `catch (RuntimeException)` path, so the volume stays marked initialized and ownership is silently never applied for the process lifetime. Suggest requiring uid and gid together (matching the model) and rolling back the guard on `status != 0` as well.

**2. Validate `root-permissions` before splicing it into `sh -c`.** The model constrains it to `^[0-7]{3,4}$`. Validating both matches AWS and keeps a typo from producing a mangled script that soft fails into the state above. Same applies to the `mount-user` `uid[:gid]` format.

**3. The `setgid` flag duplicates the AWS surface.** The 4 digit octal form already encodes it, e.g. `root-permissions: "2775"`. Consider dropping the extra flag to keep the config 1:1 with `CreationInfo.Permissions`.

**4. Run once guard races with concurrent launches.** The volume is marked initialized before the helper finishes, so a second task launched concurrently can mount a still `root:root 0755` root. A per volume lock or future would close it.

**5. Docs.** `floci.storage.efs.*` is user facing but appears nowhere under `docs/configuration/` or `docs/services/ecs.md`.

Non blocking observations, fine to defer but worth stating in the PR body: ownership is one global config block while `EfsVolumeConfiguration` already parses `authorizationConfig.accessPointId`, so per `fileSystemId` overrides are the natural follow up; `mount-group-add` is a single gid while `PosixUser.SecondaryGids` allows up to 16.

What already matches AWS: config names map 1:1 to `RootDirectory.CreationInfo{OwnerUid,OwnerGid,Permissions}` and `PosixUser{Uid,Gid}`, the chown/chmod once semantics match the documented CreationInfo behavior, and the default is a proven no op so existing users are unaffected.

abanna:
Thanks for the detailed review — worked through all five points, addressed the greptile bot, and merged current `main` in.

**1. Partial ownership → broken `chown`, + guard not rolled back on failure.** `owner-uid` and `owner-gid` are now required together (a partial `CreationInfo` isn't valid on AWS), so the helper always emits `chown uid:gid` — the malformed `chown <uid>:` (which resolved the login group and failed in busybox) is gone. `initSharedVolumeRoot` now **throws on a non-zero helper exit**, and the guard is only recorded on success, so a failed init is retried on the next launch instead of leaving the root `root:root 0755` forever. *(greptile P1 @241 and @254)*

**2. Validation before splicing into `sh -c`.** `root-permissions` is validated against `^[0-7]{3,4}$` and `mount-user` against `uid[:gid]`; a typo is rejected up front rather than producing a mangled script.

**3. Dropped the `setgid` flag.** It duplicated `CreationInfo.Permissions`, which already encodes setuid/setgid/sticky via a 4-digit octal — so setgid is now expressed as `root-permissions: "2775"`, keeping the config 1:1 with the AWS surface.

**4. Run-once guard race.** The one-off init now runs inside `computeIfAbsent`, so a concurrent launch for the same volume blocks under the per-volume lock until initialisation finishes rather than mounting a still-`root:root 0755` root; returning `null` on failure leaves it unmemoised (ties into the retry in #1).

**5. Docs.** Added a `floci.storage.efs.*` reference table to `docs/services/ecs.md` and the block to `docs/configuration/advanced/application-yml.md`.

The remaining greptile items were already handled in the current revision: `mount-user`/`mount-group-add` are wired through to the container in `EcsContainerManager`, the `application.yml` `efs:` section is present, and the `root-permissions` Javadoc was corrected. Non-blocking follow-ups (per-`fileSystemId` overrides, `PosixUser.SecondaryGids`) noted for later.

Verified (maven, temurin-25): `ContainerLifecycleManagerVolumeTest` (14, incl. new partial-ownership/invalid-permissions/retry tests), `EcsContainerManagerVolumesTest` (4), `EcsContainerManagerOverridesTest` (1) — 19 tests, green.

hectorvent (approved):
Thanks @abanna

PR #1687, opened 2026-07-02
feat(ecs): resolve containerDefinitions[].secrets from SSM / Secrets Manager

### LINKED ISSUE #1624: [FEAT] ECS: inject containerDefinitions[].secrets from Secrets Manager / SSM
## Service

ECS

## API Action / Feature

`RegisterTaskDefinition` / `RunTask`: support `containerDefinitions[].secrets` (the `Secret` object: `name` + `valueFrom`), and inject the resolved values as environment variables when launching task containers.

Today the `secrets` field is dropped on register (not parsed, not stored, not returned by describe) and never resolved at launch, so containers started against Floci come up without the env vars their task definition declares via `secrets`.

## AWS Documentation

- ContainerDefinition: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html
- Secret: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Secret.html
- Secrets Manager env vars: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/secrets-envvar-secrets-manager.html
- SSM Parameter Store env vars: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/secrets-envvar-ssm-paramstore.html

`valueFrom` accepts a Secrets Manager secret **full ARN**, or an SSM parameter (name when same-region, or full ARN). Secrets Manager also supports a selector suffix on the ARN: `...:secret:name-AbCdEf:json-key:version-stage:version-id`.

## Why is this needed?

Real task definitions pull config/credentials through `secrets` rather than plain `environment`. When emulating against Floci, those containers launch with the env vars missing, so the app under test can't start or misbehaves. Plain `environment` already works end to end; `secrets` is the remaining gap for running realistic task definitions locally.

There's an in-tree precedent for the resolution itself: `CodeBuildRunner.buildEnvList(...)` resolves both sources locally via `SsmService.getParameter(name, region)` and `SecretsManagerService.getSecretValue(secretId, versionId, versionStage, region)`. ECS launch (`EcsContainerManager.buildEnvVars(...)`) can resolve through the same local services rather than adding HTTP hops.

Rough scope:

- Parse / store / round-trip `secrets` on the container definition (describe returns the references, not resolved values, matching AWS).
- Resolve to env vars at launch, with the existing RunTask `containerOverrides` environment still winning on a name collision.
- Initial resolution: SSM parameter name or full ARN, and Secrets Manager full ARN resolving to `SecretString`. The Secrets Manager `:json-key:version-stage:version-id` selector suffix needs parsing beyond what CodeBuild does today (it passes the whole `valueFrom` as the secret id). Happy to include it or defer it, your call.

## One design question

Behavior when a referenced secret/parameter can't be resolved. AWS doesn't silently start the container with the var omitted; the unresolved reference surfaces as a task launch / resource-initialization failure. Floci's existing `CodeBuildRunner` instead swallows it (debug-log and continue). @hectorvent, for ECS do you want to fail the launch (and if so, as a `RunTask` error vs a started-then-stopped task in the current local launch model), or keep CodeBuild's lenient skip-and-continue for local-dev convenience?

- [x] Yes
- [ ] No

hampsterx:
Closes #1624.

## What

Support `containerDefinitions[].secrets` end to end: parse and round-trip the `Secret` references on register/describe, and inject the resolved values as environment variables when a task launches. Previously `secrets` was dropped on register and never resolved, so containers came up without the env vars their task definition declares via `secrets`.

## How

- **Register / describe**: `EcsJsonHandler` parses `secrets[]` (`name` + `valueFrom`) into the container definition and emits them back on `RegisterTaskDefinition` / `DescribeTaskDefinition` as references. Resolved values are never returned, matching AWS.
- **Launch**: `EcsContainerManager` resolves each secret through the in-process `SsmService` / `SecretsManagerService`, the same local resolution `CodeBuildRunner` already uses (no extra HTTP hops):
  - SSM parameter by name or full ARN → `Parameter.value`
  - Secrets Manager full ARN → `SecretString`
  - a full ARN's own region is used, so cross-region references resolve against the right store instead of the task's region
- **Precedence**: task-def `environment` < resolved `secrets` < RunTask container-override `environment`. Overrides still win on a name collision.

## Failure behavior

Per the design question in #1624, this matches AWS rather than CodeBuild's lenient skip: if a referenced secret/parameter can't be resolved, no container is started and the task goes `STOPPED` with a `ResourceInitializationError` stopped reason (naming the offending `valueFrom`), instead of silently starting without the variable.

## Deferred

The Secrets Manager `valueFrom` selector suffix (`:json-key:version-stage:version-id`) is not yet parsed; such a `valueFrom` is resolved as the plain secret. Called out as deferrable in #1624; happy to follow up.

## Test plan

- [x] `EcsIntegrationTest` (register/describe round-trip of `secrets`)
- [x] `EcsContainerManagerSecretsTest` (SSM name + SM ARN resolution, override-wins, SSM-ARN name parsing, cross-region ARN region, unresolved → propagates with no container created)
- [x] `EcsServiceContainerSecretsTest` (unresolved secret → task STOPPED with `ResourceInitializationError` reason)
- [x] `EcsTests` Java SDK compatibility: register + describe round-trips the `Secret`
- [x] Full ECS suite green locally (`./mvnw test -Dtest='Ecs*'`)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/EcsTests.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/EcsTests.java
@@ -17,6 +17,7 @@ class EcsTests {
     private static String clusterName;
     private static String family;
     private static String serviceName;
+    private static String secretArn;
     private static Cluster cluster;
     private static TaskDefinition taskDef;
     private static TaskDefinition taskDefRev2;
@@ -30,6 +31,7 @@ class EcsTests {
         clusterName = "sdk-test-cluster-" + suffix;
         family = "sdk-test-task-" + suffix;
         serviceName = "sdk-test-svc-" + suffix;
+        secretArn = "arn:aws:secretsmanager:us-east-1:000000000000:secret:sdk-db-password-AbCdEf";
     }
 
     @AfterAll
@@ -1022,5 +1024,36 @@ class EcsTests {
     void listClustersAfterDelete() {
         List<String> arns = ecs.listClusters(ListClustersRequest.builder().build()).clusterArns();
         assertThat(arns).doesNotContain(cluster.clusterArn());
+    }
+
+    @Test
+    @Order(57)
+    @DisplayName("RegisterTaskDefinition - container secrets round trip")
+    void registerTaskDefinitionWithSecrets() {
+        String secretFamily = family + "-secrets";
+        TaskDefinition secretTaskDef = ecs.registerTaskDefinition(RegisterTaskDefinitionRequest.builder()
+                .family(secretFamily)
+                .containerDefinitions(ContainerDefinition.builder()
+                        .name("app")
+                        .image("alpine:latest")
+                        .essential(true)
+                        .secrets(Secret.builder()
+                                .name("DB_PASSWORD")
+                                .valueFrom(secretArn)
+                                .build())
+                        .build())
+                .build()).taskDefinition();
+
+        assertThat(secretTaskDef.containerDefinitions().get(0).secrets()).hasSize(1);
+        assertThat(secretTaskDef.containerDefinitions().get(0).secrets().get(0).name()).isEqualTo("DB_PASSWORD");
+        assertThat(secretTaskDef.containerDefinitions().get(0).secrets().get(0).valueFrom()).isEqualTo(secretArn);
+
+        TaskDefinition described = ecs.describeTaskDefinition(DescribeTaskDefinitionRequest.builder()
+                .taskDefinition(secretTaskDef.taskDefinitionArn())
+                .build()).taskDefinition();
+
+        assertThat(described.containerDefinitions().get(0).secrets()).hasSize(1);
+        assertThat(described.containerDefinitions().get(0).secrets().get(0).name()).isEqualTo("DB_PASSWORD");
+        assertThat(described.containerDefinitions().get(0).secrets().get(0).valueFrom()).isEqualTo(secretArn);
     }
 }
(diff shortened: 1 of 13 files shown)


hectorvent:
Thanks for this, really clean work. Verified the wire contract against botocore's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/ecs/2014-11-13/service-2.json) and found no blocking issues. The `Secret` shape casing (`name`/`valueFrom`) round-trips correctly, `secrets` is omitted when absent as awsjson1.1 expects, ARN-region-wins matches the developer guide, and the resolution failure correctly surfaces as a stopped task instead of a `RunTask` error. The deferred selector suffix is fine as a follow-up.

Two small notes, neither blocking:

1. The stopped reason picks up a prefix from [`EcsService.java:484`](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java#L484) (pre-existing code), so clients see:

```
Failed to start: ResourceInitializationError: unable to pull secrets or registry auth: /missing: ...
```

while real AWS starts the reason directly with `ResourceInitializationError: ...`. Worth dropping the prefix here since the PR otherwise matches the AWS failure shape exactly.

2. A secret with only `SecretBinary` makes `resolveSecretValue` return null and the env join emits a literal `NAME=null`. Real AWS fails the launch for binary secrets, so a null check throwing the same `ResourceInitializationError` would be more faithful.

Happy to merge with or without those addressed.

hampsterx:
@hectorvent both addressed, and rebased onto current main (upstream ECS awsvpc/port + the awsEnv baseline work had moved out from under the branch):

1. The resolve path now throws an `AwsException` coded `ResourceInitializationError`, and `runTask` keys off that code to pass the reason through verbatim, no `Failed to start:` prefix, so an unresolved secret reads exactly like real AWS. Other start failures keep the prefix.

2. A `SecretBinary`-only secret (null `SecretString`) now fails the launch with the same `ResourceInitializationError` instead of injecting `NAME=null`; the null-guard also covers a null service response, so the resolve path can't NPE past the wrapper.

Also corrected the Secrets Manager comment: a `:json-key:version-stage:version-id` selector suffix is passed through unparsed and fails to resolve rather than silently returning the whole secret (still deferred, #1624).

Ready to merge. Only gap I see is an end-to-end test resolving a real SSM parameter into a launched container's env, the wiring is covered by the integration tests, the resolved value only by mocks. Happy to add it here or leave as a follow-up, your call.

hectorvent (approved):
Thanks @hampsterx

PR #1691, opened 2026-07-02
feat(stepfunctions): resolve payload-template .$ inside arrays and $$ context in intrinsic args

abanna:
Two AWS States Language parity fixes in the Step Functions payload-template / intrinsic path.

### 1. `resolveParameters` now recurses into arrays
A `Parameters` / `ItemSelector` / `ResultSelector` payload template must resolve `.$` references at **any** nesting depth. Previously `resolveParameters` recursed into nested objects but passed array elements through verbatim, so a `.$` reference inside an array was never substituted.

Real-world shape this fixes — an ECS `Overrides.ContainerOverrides[].Environment[].Value.$`:

```json
"Parameters": {
  "Overrides": {
    "ContainerOverrides": [
      { "Name": "app", "Environment": [ { "Name": "ID", "Value.$": "$.systemId" } ] }
    ]
  }
}
```

Without array recursion, `Value.$` stayed literal and the launched task got an empty environment.

### 2. Intrinsic arguments can reference the Context Object (`$$`)
`resolvePath` / `evaluateIntrinsic` / `resolveIntrinsicArg` now thread the Context Object through, so an intrinsic argument can read `$$`:

```
States.Format('id={}', $$.Map.Item.Value.systemId)
$$.Execution.Id
```

Ordinary `$.`/`States.*` input-only resolution is unchanged — `context` is `null` on that path, so existing behavior is preserved.

### Tests
- `AslExecutorArrayParamsTest` — `.$` resolution inside arrays.
- `AslExecutorIntrinsicContextTest` — `$$` references in intrinsic args (and that input-only resolution still ignores context).

Built + tested in the temurin container (targeted SFN suite green).


hectorvent:
@abanna Thanks for the PR. Reviewed both semantics against the LocalStack ASL interpreter as reference (`service-2.json` does not model States Language semantics) and both changes match AWS behavior:

* Array recursion matches `PayloadArr`, which evaluates every array element as a payload value, so `.$` bindings resolve at any depth including the ECS `ContainerOverrides[].Environment[].Value.$` shape.
* `$$` intrinsic args match `ArgumentContextPath`, which strips one `$` and resolves the rest against the Context Object, exactly what `resolvePath("$." + arg.substring(3), context)` does, with bare `$$` returning the whole context.

Two things to address:

**1. The branch conflicts with main and the conflict is semantic.** Main has since split `resolvePath` into `resolvePath` (collapses `MissingNode` to `NullNode`) plus `resolvePathNode` (preserves missing vs null for `IsPresent`), and the `States.` intrinsic dispatch now lives in `resolvePathNode`. When rebasing, thread the context through `resolvePathNode` rather than restoring the old single method shape, and keep the 3 arg `resolvePath` collapsing missing to null:

```java
JsonNode resolvePath(String path, JsonNode root, JsonNode context) {
    JsonNode node = resolvePathNode(path, root, context);
    return node.isMissingNode() ? NullNode.getInstance() : node;
}
```

Otherwise either the `IsPresent` semantics or the context threading silently regresses in the conflict resolution.

**2. Minor: the null context fallback pins behavior AWS does not have.** With `context == null` a `$$.` intrinsic arg formats as the string `"null"`, and `noContext_intrinsicResolutionUnchanged` asserts exactly that. On real AWS the Context Object always exists. Fine today since every payload template call site passes context, but a comment marking it as Floci internal transitional behavior would keep that test from blocking a future change that threads context into the remaining `resolvePath` callers.

Non blocking: no SDK level end to end test (e.g. a Map state `ItemSelector` with `$$.Map.Item.Value` via `StartExecution`/`DescribeExecution`); the unit tests on package private methods follow the repo convention and the targeted suite is reported green, so fine as a follow up.

Also checked: input only resolution is genuinely unchanged (2 arg overload passes null context), the nested intrinsic fall through is consistent, JSONata mode is untouched, and no wire protocol surface is affected.

abanna:
Thanks for the detailed review — addressed both points and merged current \`main\` in.

**1. Semantic conflict (resolved exactly as suggested).** Threaded the Context Object through a 3-arg \`resolvePathNode\` (where \`States.\` dispatch now lives) and kept the 3-arg \`resolvePath\` collapsing missing→null:

```java
JsonNode resolvePath(String path, JsonNode root, JsonNode context) {
    JsonNode node = resolvePathNode(path, root, context);
    return node.isMissingNode() ? NullNode.getInstance() : node;
}
```

So neither `IsPresent` (uses `resolvePathNode`, preserves `MissingNode`) nor `$$` context threading regresses. `main` also added `States.ArrayContains` after this branch forked; its arg resolution now passes `context` through the new 3-arg `resolveIntrinsicArg`. The two new tests were adapted to `main`’s 13-arg `AslExecutor` constructor.

**2. Null-context fallback marked transitional.** Added a comment in `resolveIntrinsicArg` and on `noContext_intrinsicResolutionUnchanged` noting the `"null"` string is Floci-internal transitional behavior (not AWS semantics — the Context Object always exists on AWS), so it should be updated rather than treated as a contract once context is threaded into the remaining `resolvePath` callers.

**Greptile naming suggestion (`*ServiceTest`):** respectfully declining — the repo already has sibling `AslExecutor*Test` files (`AslExecutorPathIntrinsicsTest`, `AslExecutorCatchTest`, `AslExecutorRecaseKeysTest`, …), so the new tests follow the established local convention for this class.

**Verified** (maven, temurin-25): `AslExecutorArrayParamsTest`, `AslExecutorIntrinsicContextTest`, `AslExecutorPathIntrinsicsTest`, `AslExecutorCatchTest`, `AslExecutorEcsRunTaskModeTest`, `AslExecutorRecaseKeysTest`, `StepFunctionsChoiceIsPresentIntegrationTest`, `StepFunctionsResultSelectorMapParallelIntegrationTest`, `StepFunctionsAccountContextIntegrationTest` — 34 tests, all green.

hectorvent (approved):

PR #1692, opened 2026-07-02
feat(cloudformation): support UpdateTerminationProtection

abanna:
CDK `cdk bootstrap` calls `UpdateTerminationProtection` on the CDKToolkit stack; Floci returned `UnknownAction`, so bootstrap couldn't complete.

### Change
- **`Stack`**: adds a `enableTerminationProtection` flag (default false) + accessors.
- **`CloudFormationService.updateTerminationProtection(stackName, enabled, region)`**: resolves the stack (`ValidationError` if it doesn't exist), sets the flag, and persists.
- **`CloudFormationQueryHandler`**: dispatches `UpdateTerminationProtection` → returns the `UpdateTerminationProtectionResult` envelope with the `StackId`; and echoes `<EnableTerminationProtection>` in the `DescribeStacks` stack XML.
- **`AwsQueryController`**: registers the action in the CloudFormation action set so the Query-protocol POST routes to CloudFormation (without this it fell through to another service).

### Test
`updateTerminationProtection_togglesFlagAndReflectsInDescribeStacks` — DescribeStacks reports `false` by default, `UpdateTerminationProtection` returns the StackId, and DescribeStacks then reflects `true`.

Built + tested in the temurin container; the full `CloudFormationIntegrationTest` passes (the one unrelated failure is an `AWS::ECR::Repository` test that needs a Docker socket the build container doesn't have).


hectorvent:
@abanna Thanks for the PR. Verified against the CloudFormation `service-2.json`: the wire contract is correct. Input `StackName` and `EnableTerminationProtection` are both read, the output envelope matches `UpdateTerminationProtectionOutput { StackId }`, and `EnableTerminationProtection` is a real `Stack` member with exactly that casing so the `DescribeStacks` echo is right. The operation declares no errors in the model and the undeclared `ValidationError` for a missing stack matches real AWS. LocalStack's provider implements the operation the same way.

A few things:

**1. `DeleteStack` does not enforce the protection.** The model doc for this operation says "If a user attempts to delete a stack with termination protection enabled, the operation fails and the stack remains unchanged", but `deleteStack` never checks the new flag, so a protected stack deletes normally. Real AWS returns a 400 `ValidationError`:

```
Stack [arn:aws:cloudformation:...] cannot be deleted while TerminationProtection is enabled
```

Neither moto nor LocalStack enforces it either, so this is refimpl parity, but since the flag is now surfaced in `DescribeStacks` a client can rely on it and the check is a few lines in `deleteStack`. Please add it (with a test: delete protected stack returns 400 and the stack remains) or note the deferral explicitly in the PR body.

**2. Minor: `CreateStack` still ignores its own `EnableTerminationProtection` input.** The model declares the same member on `CreateStackInput`. Not needed for the CDK bootstrap flow, fine as a follow up.

**3. Minor: `Boolean.parseBoolean(params.getFirst(...))` turns an absent required `EnableTerminationProtection` into `false`, and a missing `StackName` flows through as null. AWS rejects missing required params with `ValidationError`. Matches the handler's existing style, so non blocking.

**4. The branch conflicts with main**, needs a rebase (the `CLOUDFORMATION_ACTIONS` set and the handler switch are hot spots).

Also checked and fine: response envelope and casing, default false semantics (LocalStack defaults the same), persistence through the existing `persistStack` path, action registered in the existing Query routing with no custom endpoints. The test is raw form encoded HTTP in the existing file's style; an SDK based case via the compatibility suite would be the preferred validation if you get a chance.

abanna:
Thanks — addressed all of it and merged current `main` in.

**1. DeleteStack now enforces protection.** `deleteStack` rejects a stack with termination protection enabled with a 400 `ValidationError` (`Stack [<id>] cannot be deleted while TerminationProtection is enabled`) and leaves the stack unchanged. New test `deleteStack_protectedStackIsRejectedAndStackRemains` asserts the 400 + XML error and that `DescribeStacks` still returns the stack (not `DELETE_*`).

**2. CreateStack honors `EnableTerminationProtection`.** The `CreateStackInput` member is now applied, so a stack can be created already-protected. New test `createStack_withTerminationProtectionEnabled` verifies it reflects in `DescribeStacks`.

**greptile P1 — XML error contract.** Both `updateTerminationProtection` and `deleteStack` handlers now wrap the service call in `catch (AwsException e) { return xmlError(...); }`, so a missing/protected stack returns the Query-protocol XML error rather than the JSON body from `AwsExceptionMapper` — matching `describeStacks`/`describeChangeSet`/etc.

**3. Param validation (`Boolean.parseBoolean` / missing `StackName`)** — left as-is per your "non-blocking, matches the handler's existing style" note; happy to tighten it repo-wide separately.

**4. Conflict** — resolved (kept both the termination-protection test and main's new CFN integration tests).

Verified (maven, temurin-25): `updateTerminationProtection_togglesFlagAndReflectsInDescribeStacks`, `deleteStack_protectedStackIsRejectedAndStackRemains`, `createStack_withTerminationProtectionEnabled` — green. (Re the SDK-based compatibility case: I kept the raw form-encoded style of the existing file; glad to add an SDK case in `compatibility-tests` if you'd prefer that as the canonical validation.)

hectorvent (approved):
Thanks @abanna

PR #1696, opened 2026-07-02
test(stepfunctions): fix AslExecutorPathIntrinsicsTest compile against ecs:runTask ctor

abanna:
## Problem

The test sources on `main` no longer compile. `AslExecutor` gained `EcsService` and `EcsJsonHandler` constructor parameters with the ecs:runTask service integration (#1564), but `AslExecutorPathIntrinsicsTest` (added in #1558) was never updated to pass them. Because Maven `testCompile` is all-or-nothing, this breaks the whole test compilation — and therefore CI — on `main` and on every open PR.

```
constructor AslExecutor in class ...AslExecutor cannot be applied to given types;
  required: ...,S3Service,EcsService,EcsJsonHandler,ObjectMapper,...
  found:    ...,S3Service,ObjectMapper,...
```

## Fix

Pass the two missing mocks (`EcsService`, `EcsJsonHandler`) in the constructor call, in the same position and style as the sibling `AslExecutor` tests (e.g. `AslExecutorCatchTest`, `AslExecutorEcsRunTaskModeTest`). No production code changes.

## Verification

Built and ran the affected test in a JDK 25 container: `AslExecutorPathIntrinsicsTest` — 13 tests, all pass.

### DIFF
--- a/compatibility-tests/compat-opentofu/main.tf
+++ b/compatibility-tests/compat-opentofu/main.tf
@@ -257,3 +257,50 @@ output "zone_id" {
 output "health_check_id" {
   value = aws_route53_health_check.app.id
 }
+
+# ── Cognito User Pool ─────────────────────────────────────────────────────
+resource "aws_cognito_user_pool" "pool" {
+  name = "floci-compat-pool"
+
+  password_policy {
+    minimum_length    = 12
+    require_lowercase = true
+    require_numbers   = true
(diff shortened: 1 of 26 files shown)


hectorvent:
@abanna it looks like changes from this PR were already merge by another PR. This PR doesn't contain any file changed.

PR #1697, opened 2026-07-02
feat(ecs): inject the AWS SDK baseline env into launched task containers

abanna:
> **Two commits.** The first commit is the same one-line test-compile fix as #1696 (`AslExecutorPathIntrinsicsTest` was not updated for the `ecs:runTask` constructor change, which breaks `testCompile` on `main`). It's included here so this PR's CI is green on its own; once #1696 merges it resolves cleanly (identical change) and can be dropped when this branch is updated off `main`. The second commit is the feature described below. Happy to rebase this onto #1696 if you'd prefer to review the feature commit in isolation.

## What

Give containers launched for **ECS tasks** the same baseline AWS SDK environment that Lambda containers already receive — region, credentials, and the Floci endpoint — so AWS SDK calls made *inside* a task container reach the emulator instead of the container's own loopback.

Today a launched ECS task container gets only its task-definition/override environment. An AWS SDK running in that container therefore has no `AWS_ENDPOINT_URL` (so it targets real AWS, not Floci), no region, and an empty credential-provider chain (`Unable to load credentials from any providers`). Lambda containers already avoid this because `ContainerLauncher` injects those variables. This change extracts that baseline into a small shared helper and applies it to ECS tasks too.

## How

- **New `LaunchedContainerAwsEnv`** (`core/common/docker`) — an `@ApplicationScoped` helper that builds the `KEY=value` baseline for any launched container: `AWS_DEFAULT_REGION`/`AWS_REGION`, credentials (a mounted `~/.aws` dir when configured, otherwise placeholder credentials from Floci's own env, falling back to `test`), and `FLOCI_HOSTNAME`/`FLOCI_ENDPOINT`/`AWS_ENDPOINT_URL` resolved via `ContainerReachableEndpoint`.
- **`ContainerLauncher` (Lambda)** — replaced the inline region/credentials/endpoint block with a call to the shared helper. The resulting environment is byte-for-byte identical to before (same keys, same values, same order); only the source of those entries moved. `ContainerReachableEndpoint` is no longer injected here directly — the helper owns it.
- **`EcsContainerManager` (ECS)** — injects the helper and prepends the baseline to each task container's environment as **overridable defaults**: task-definition and RunTask `containerOverride` values still win on a key conflict, so a task's own configuration is never clobbered. Also enables `withHostDockerInternalOnLinux()` + `withEmbeddedDns()` so the reachable endpoint hostname resolves to Floci from inside the task container, exactly as Lambda containers do.

## Note on credentials (pragmatic shortcut)

Credentials are injected as environment placeholders (`test`/`test`/`test`, or the values in Floci's own environment), matching what `ContainerLauncher` already does for Lambda — Floci does not verify request signatures. This is **not** the real ECS task-role credential mechanism (the `169.254.170.2` container-credentials metadata endpoint referenced by `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`). Emulating that endpoint would be a reasonable follow-up for workloads that specifically exercise task-role credential fetching; the env-placeholder approach is sufficient for SDK calls to reach the emulator with usable credentials.

## Tests

- New `LaunchedContainerAwsEnvTest` — unit-covers the helper: region, placeholder-credential and mounted-`~/.aws` branches, and endpoint/hostname derivation.
- New `EcsContainerManagerAwsBaselineTest` — asserts the baseline is injected into task containers and is overridden by both task-definition env and RunTask `containerOverride` env on key conflict.
- Updated the existing `ContainerLauncher`/`EcsContainerManager`/`LambdaImageConfig` unit tests for the new constructor wiring. Existing Lambda env assertions are unchanged (the env is byte-identical).
- Verified via the `EcsIntegrationTest` `@QuarkusTest` that the new bean wires correctly through CDI.

## Environment precedence

Intended precedence for a launched ECS task container's environment, lowest to highest (stated for the mechanical merge with #1687, which restructures the same `buildEnvVars` env map):

```
baseline (this PR)  <  task-definition environment  <  resolved secrets (#1687)  <  RunTask containerOverride
```

The baseline is prepended as overridable defaults; task-definition env, secrets, and per-invocation `containerOverride` each win on a key conflict in that order. This PR and #1687 change `buildEnvVars` to the same new signature and restructure the same map, so whichever merges second is a manual merge that should preserve the above order.

_Rebased onto current `main` (merge, not rebase — the branch is public); the only conflict was an import ordering in `EcsContainerManager`. Both greptile items are already addressed in the current revision: the unused single-arg constructor was removed (no `reachableEndpoint` NPE path), and the `EcsContainerManager` null guard was dropped in favor of stubbing `sdkBaselineEnv` in the affected tests._


abanna:
Thanks @greptile — both addressed in `66d97f8`:

- **P1 (single-arg ctor NPE):** removed the unused single-argument constructor. It was the only thing that stored a `null` `ContainerReachableEndpoint`. While there I also dropped the vestigial `EmulatorConfig` field/param, since `sdkBaselineEnv` never reads it — the helper now depends only on `ContainerReachableEndpoint`.
- **P2 (dead null guard):** removed the `if (awsBaseline != null)` guard in `EcsContainerManager.buildEnvVars` (the real `sdkBaselineEnv` always returns a non-null list) and instead stubbed the mocked helper with `List.of()` in the three ECS unit tests, so the contract is explicit.

Verified locally (JDK 25 container): all affected unit tests + `EcsIntegrationTest` (@QuarkusTest, CDI wiring with the simplified constructor) pass.

hectorvent:
Thanks, this is a well reasoned change and the disclosure of the credentials shortcut is appreciated. Nothing wire facing to flag: the injection happens only at Docker launch, so `RegisterTaskDefinition` and `DescribeTaskDefinition` responses are untouched, which matches AWS (runtime injected env is never echoed through the task definition). I also verified the Lambda refactor is byte identical as claimed (same keys, same order, `/opt/aws-config` preserved), the override precedence is pinned by the new tests, and `withHostDockerInternalOnLinux()` plus `withEmbeddedDns()` mirror exactly what `ContainerLauncher` does for Lambda. Real ECS supplies credentials via the `169.254.170.2` metadata endpoint rather than static env vars, but the env placeholder approach is the same tradeoff Floci already makes for Lambda and everything stays overridable, so that is fine as documented.

Two things before merging:

1. The branch is conflicting with current `main`, so it can be dropped on rebase.

2. This PR and #1687 (ECS secrets) change `buildEnvVars` to the identical new signature and restructure the same env map, so whichever merges second has a manual merge. The combined precedence should come out as:

```
baseline (this PR)  <  task-def environment  <  resolved secrets (#1687)  <  RunTask containerOverride
```

Stating that intended order in the PR description would make the second merge mechanical.

One small inherited quirk, fine as a follow up: if the Floci host process itself has real `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` set, those values are copied into every launched container. That behavior comes from the existing Lambda block and now extends to ECS tasks, worth a hard look someday.

abanna:
Thanks for the thorough review. Addressed:

1. **Conflict with `main`** — merged current `main` in (merge, not rebase, since the branch is public). The only conflict was an import ordering in `EcsContainerManager`.
2. **Precedence / #1687 merge order** — documented in the PR description: `baseline < task-definition env < resolved secrets (#1687) < RunTask containerOverride`, so the second of the two to merge is a mechanical merge.
3. **Host-credential passthrough quirk** — noted as a follow-up; it's inherited from the existing Lambda block, agreed it's worth a separate look.

Both greptile items are already handled in the current revision (from the earlier review-feedback commit): the unused single-arg constructor was removed, so there's no `reachableEndpoint` NPE path, and the `EcsContainerManager` null guard was dropped in favor of stubbing `sdkBaselineEnv(...)` in `EcsContainerManagerOverridesTest`/`PortMappingsTest`/`VolumesTest`.

Verified (maven, temurin-25): `LaunchedContainerAwsEnvTest`, `EcsContainerManagerAwsBaselineTest`, `EcsContainerManager{Overrides,PortMappings,Volumes}Test`, `ContainerLauncherTest` — 26 tests, green.

hectorvent (approved):
Thanks @abanna

PR #1700, opened 2026-07-02
fix(elbv2): dispatch same-port listeners by host

jvanzyl:
## Summary
- route same-port listener traffic by request host
- return a clear 502 when no listener matches the host
- add focused ELBv2 coverage for host-dispatched listeners

## Verification
- ./mvnw -q -Dtest=ElbV2SamePortHostDispatchIntegrationTest test

### DIFF
--- a/docs/services/docdb.md
+++ b/docs/services/docdb.md
@@ -21,6 +21,7 @@ The management API shares the RDS Query endpoint (`POST /` with an `Action=` par
 | --- | --- |
 | `CreateDBCluster` | Create a DocumentDB cluster and start a MongoDB container |
 | `DescribeDBClusters` | List clusters and their connection details |
+| `DescribeDBClusterSnapshots` | - |
 | `DeleteDBCluster` | Stop and remove a cluster (must have no instances) |
 | `ModifyDBCluster` | Update engine version or IAM auth setting |
 | `CreateDBInstance` | Add an instance to a cluster |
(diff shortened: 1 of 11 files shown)


hectorvent:
Really nice architecture change. The per port server with host dispatch is the right model, and the silent bind failure for a second same port listener was a real bug worth killing. The lifecycle handling (closing the port server only when the last listener leaves, the value based remove, the bind failure cleanup) is careful work.

Two asks before merge, both small.

First, real ALBs do not require the Host header to match their DNS name. Any request reaching the listener is processed by its rules, and host routing is opt in through rule conditions. Strict matching here breaks every workflow that curls `localhost:<port>` against a single ALB, which is visible in the PR itself since two existing tests had to move off `127.0.0.1`. A fallback keeps your dispatch and the old behavior:

```java
String listenerArn = listenersByHost.get(host);
if (listenerArn == null && listenersByHost.size() == 1) {
    listenerArn = listenersByHost.values().iterator().next(); // sole listener wins, like a real ALB
}
```

With that, the 502 only fires on genuine ambiguity (two or more listeners, no host match) and the edits to [ElbV2LambdaTargetDataPlaneIntegrationTest.java](https://github.com/floci-io/floci/blob/main/src/test/java/io/github/hectorvent/floci/services/elbv2/ElbV2LambdaTargetDataPlaneIntegrationTest.java) and the ApiGatewayV2 test can revert.

Second, connecting to `http://<dnsName>:port` in [ApiGatewayExecuteController.java](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java) now depends on `*.localhost.floci.io` resolving through public DNS, which fails offline. The comment you removed existed for a cousin of this pitfall. Connecting to `127.0.0.1` and setting the Host header to the DNS name gives exact dispatch with no DNS dependency.

Happy to merge once those are in. Thanks for the thorough test coverage, the unknown host cases included.

hectorvent (approved):
Thanks @jvanzyl

PR #1701, opened 2026-07-02
feat: Implement Cognito GlobalSignOut

AbhigyaKrishna:
## Summary

Implements the Cognito **`GlobalSignOut`** action, which was previously an
unsupported operation (it fell through to the `UnsupportedOperation` default in
the JSON dispatcher).

`GlobalSignOut` is the self-service counterpart to the already-supported
`AdminUserGlobalSignOut`: instead of admin credentials, it authenticates with the
caller's own **access token** and invalidates all tokens (access, ID, refresh)
that Cognito issued to that user.

- **`CognitoService.globalSignOut(accessToken)`** — extracts the username/pool/jti
  from the access token, rejects invalid or already-revoked/signed-out tokens with
  `NotAuthorizedException`, then revokes every issued token for the user via the
  existing `revokeAllUserTokens(...)`.
- **`CognitoJsonHandler`** — adds `handleGlobalSignOut(...)` and wires
  `case "GlobalSignOut"` into the action dispatch (returns HTTP 200 with an empty
  body, matching AWS).

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

New action `GlobalSignOut` on the `AWSCognitoIdentityProviderService` target
(AWS JSON 1.1). Wire contract:

- **Request:** `{ "AccessToken": "<jwt>" }`
- **Success:** `200` with empty JSON body `{}`
- **Failure:** `NotAuthorizedException` for an invalid, revoked, or
  already-signed-out access token.

Behavior mirrors the existing, already-verified `AdminUserGlobalSignOut` path and
was validated end-to-end through the project's integration-test harness
(`GlobalSignOutIntegrationTest`), which exercises the real request/response
envelope: sign-in → `GlobalSignOut` → old access token rejected → old refresh
token rejected → a fresh login still succeeds.

## Checklist

- [x] New or updated integration test added (`GlobalSignOutIntegrationTest`, 9 cases)


hectorvent:
@AbhigyaKrishna Thanks for the PR. Verified against `service-2.json` (both the aws-sdk-java-v2 and botocore copies). One wire contract issue, the rest looks good.

**Blocking: `UserNotFoundException` is not a declared error for `GlobalSignOut`.** The declared list is `ResourceNotFoundException`, `InvalidParameterException`, `NotAuthorizedException`, `TooManyRequestsException`, `PasswordResetRequiredException`, `UserNotConfirmedException`, `InternalErrorException`, `ForbiddenException`. `UserNotFoundException` is only declared for `AdminUserGlobalSignOut`, where the caller names the user. The self service op authenticates by token, and on real AWS a token for a nonexistent user cannot pass signature validation, so the failure is `NotAuthorizedException`. moto does the same: any access token that does not resolve raises `NotAuthorizedError`. Please remap the user missing case in `globalSignOut` and flip the Order 10 assertion:

```java
try {
    adminGetUser(poolId, username);
} catch (AwsException e) {
    if ("UserNotFoundException".equals(e.getErrorCode())) {
        throw new AwsException("NotAuthorizedException", "Invalid access token", 400);
    }
    throw e;
}
```

Pool missing can stay `ResourceNotFoundException` since that one is declared for this operation.

**Minor: missing `AccessToken` maps to the wrong error.** `request.path("AccessToken").asText()` turns an absent parameter into `""` which flows into the `NotAuthorizedException` path. The parameter is REQUIRED with pattern `[A-Za-z0-9-_=.]+` and real AWS rejects a missing required param with `InvalidParameterException`, which is declared.

Non blocking: the model's member doc says the token must include the `aws.cognito.signin.user.admin` scope; Floci's other token authenticated paths do not validate scope either, so fine to leave as consistent existing behavior.

Also checked and fine: awsjson1.1 `200` with `{}` for the empty `GlobalSignOutResponse`, `NotAuthorizedException` for invalid, revoked, or already signed out tokens (declared, matches moto), and the token lifecycle semantics match the operation doc (all issued tokens invalidated including the presenting one, fresh logins unaffected, old refresh token stays dead after re login). The lifecycle test coverage is genuinely thorough.

hectorvent:
@AbhigyaKrishna please rebase main to resolve documentation action failure.

hectorvent (approved):
Thanks @AbhigyaKrishna

PR #1703, opened 2026-07-02
fix(docker): isolate managed resource containers

jvanzyl:
## Summary
- add a per-run discriminator to managed container and volume names
- avoid creating host filesystem paths from inside containerized runs
- keep RDS child containers isolated when separate emulator instances create the same logical DB identifier

## Verification
- ./mvnw -q -Dtest=ContainerStorageHelperTest,RdsContainerManagerTest test


hectorvent:
Thanks @jvanzyl, this is a well built change. The namespace is default off, every `resourceName`/`applyStorage` call site passes config so create and remove stay symmetric, the sanitization is safe for both docker names and path segments, and skipping `ensureHostDir` inside containerized runs is the right call. Nice test coverage too.

One thing to fix before merge: OpenSearch containers get renamed even with no namespace configured. The container name switched from the domain name to `resourceName(...)`, which prefers `volumeId`, and every domain gets a random one at creation (`OpenSearchService:126`):

```java
// before:  floci-opensearch-{domainName}
// after:   floci-opensearch-{volumeId}   (volumeId is always set)
```

So after upgrading, `isReady()` probes the wrong container DNS name for any persisted domain and stop/remove misses the pre upgrade container. Passing `null` as the volumeId for the container name (keeping it only for the volume) restores the old name when the namespace is blank.

Two small non blocking notes: `floci-amazonmq-{brokerId}`, `floci-ec2-fwd-...` and `floci-efs-...` stay un namespaced (safe in practice since the ids are random, but a `dockerName()` wrap each would make isolation uniform), and the new `FLOCI_DOCKER_RESOURCE_NAMESPACE` deserves a row in `docs/configuration/environment-variables.md`.

Happy to merge once the OpenSearch name is settled.

hectorvent (approved):
Thanks @jvanzyl

PR #1705, opened 2026-07-02
feat: emit jti/origin_jti and support RevokeToken

AbhigyaKrishna:
## Summary

Brings Floci's Cognito token generation to parity with real AWS Cognito by emitting the two native token-revocation claims and honoring the RevokeToken API. Implements the Sapphire Terminal team's patch request (floci-patch-origin-jti-request.md) that unblocks end-to-end dev testing of "remote device logout" (instant session revocation).

- jti — a unique UUID stamped on every access and ID token, regenerated on each issuance/refresh.
- origin_jti — a per-refresh-token-family UUID: minted at initial authentication, embedded in the refresh token, and kept identical for every access/ID token later minted from that refresh token (stable across REFRESH_TOKEN_AUTH / GetTokensFromRefreshToken), while differing between independent logins.
- `jti` is always emitted (unconditional). `origin_jti` is gated on the app client's `EnableTokenRevocation` flag (default true) and omitted when revocation is disabled, matching AWS behavior.
- RevokeToken API wired: revoking a refresh token invalidates its whole `origin_jti` family, so subsequent refreshes **and access-token-gated operations** (GetUser, ChangePassword, UpdateUserAttributes, DeleteUserAttributes) fail with NotAuthorizedException. Error codes on RevokeToken itself use `UnauthorizedException` per the declared API model.

## Type of change

- [ ] Bug fix (fix:)
- [x] New feature (feat:)
- [ ] Docs / chore

## AWS Compatibility

- Protocol: Cognito AWS JSON 1.1 (X-Amz-Target: AWSCognitoIdentityProviderService.*). No wire-format or endpoint-shape changes.
- Claims: jti / origin_jti match AWS's documented token revocation semantics — origin_jti is the stable revocation identifier for the refresh-token family; jti is unique per JWT.
- RevokeToken errors modeled after AWS: UnsupportedOperationException when the client has EnableTokenRevocation=false, UnsupportedTokenTypeException for a non-refresh token, UnauthorizedException (403) for a client mismatch / bad client secret, InvalidParameterException for missing Token/ClientId; empty 200 on success.
- DescribeUserPoolClient already reports EnableTokenRevocation (defaults to true on client creation) — verified by the new test.
- Verification: RestAssured integration tests exercise the real JSON 1.1 wire path (InitiateAuth USER_PASSWORD_AUTH / REFRESH_TOKEN_AUTH, GetTokensFromRefreshToken, RevokeToken, DescribeUserPoolClient) and decode the issued JWTs to assert claim presence, family stability, per-token jti uniqueness, and revocation behavior.

## Checklist

- [x] ./mvnw test passes locally — full Cognito suite 234 + 10 new = green (run under JDK 26 with -Denforcer.skip=true, as no JDK 25 is installed locally; CI on JDK 25 is authoritative)
- [x] New or updated integration test added — OriginJtiIntegrationTest (12 tests)
- [x] Commit messages follow Conventional Commits — feat: emit jti/origin_jti and support RevokeToken

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
@@ -91,6 +91,7 @@ public class CognitoJsonHandler {
             case "AdminRemoveUserFromGroup" -> handleAdminRemoveUserFromGroup(request);
             case "AdminListGroupsForUser" -> handleAdminListGroupsForUser(request);
             case "GetTokensFromRefreshToken" -> handleGetTokensFromRefreshToken(request);
+            case "RevokeToken" -> handleRevokeToken(request);
             case "ListUserPoolClientSecrets" -> handleListUserPoolClientSecrets(request);
             case "AddUserPoolClientSecret" -> handleAddUserPoolClientSecret(request);
             case "DeleteUserPoolClientSecret" -> handleDeleteUserPoolClientSecret(request);
@@ -448,6 +449,15 @@ public class CognitoJsonHandler {
                 request.path("RefreshToken").asText()
         );
         return Response.ok(objectMapper.valueToTree(result)).build();
+    }
+
+    private Response handleRevokeToken(JsonNode request) {
+        service.revokeToken(
+                request.path("ClientId").asText(null),
+                request.path("Token").asText(null),
+                request.path("ClientSecret").asText(null)
+        );
+        return Response.ok(objectMapper.createObjectNode()).build();
     }
 
     private Response handleInitiateAuth(JsonNode request) {
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -1476,14 +1476,17 @@ public class CognitoService {
         }
 
         UserPoolClient client = findClientById(clientId);
-        if (!isTokenRevocationEnabled(client)) {
-            throw new AwsException("UnsupportedOperationException",
-                    "Please enable token revocation before revoking tokens for this client", 400);
-        }
 
+        // Authenticate the caller before disclosing configuration state: a confidential client must
+        // present a valid secret first, matching AWS which validates identity ahead of feature checks.
         String secret = client.getClientSecret();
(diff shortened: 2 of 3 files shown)


hectorvent:
Nice work on the `origin_jti` lifecycle, it matches moto exactly: minted once at auth, embedded in the refresh token, reused for every token minted from it, fresh on an independent login. `OriginJtiIntegrationTest` is green on the branch, 11/11.

**`RevokeToken` doesn't revoke access tokens yet**, which is the remote logout case this is for. The revocation is stored under `origin_jti`, but the access path looks up a different key:

```java
String jti = extractJtiFromToken(accessToken);   // fresh UUID on every issuance
validateTokenNotRevoked(jti, poolId, "access");
```

That `jti` was never written to the store, so the lookup misses and an already issued access token keeps authorizing `GetUser`, `GlobalSignOut`, `ChangePassword` and `UpdateUserAttributes` until it expires. The [model](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json) says AWS kills those too: "Revokes all of the access tokens generated by, and at the same time as, the specified refresh token." Which is why `origin_jti` is on the access token at all. The access path wants to read `origin_jti` and check it alongside `jti`, plus a test, since `revokeTokenInvalidatesRefreshFamily` only asserts the refresh flows fail.

**`NotAuthorizedException` isn't declared on `RevokeToken`.** You throw it for a bad client secret and a client mismatch, but the operation only declares `UnauthorizedException`, `InvalidParameterException`, `UnsupportedOperationException`, `UnsupportedTokenTypeException`, `OperationNotEnabledException`, `Forbidden`, `TooManyRequests` and `InternalError`. The SDK can't map an undeclared code, so callers get a generic `CognitoIdentityProviderException`. `UnauthorizedException` looks intended. Being upfront, neither moto nor localstack implements `RevokeToken`, so the declared list is all I have to go on.

Two small ones: the description says both claims are gated on `EnableTokenRevocation`, but `claims.put("jti", jti)` is unconditional and only `origin_jti` is gated. And the three arg `buildRefreshToken` now has no callers outside its own delegation.

The rest checks out. Empty `{}` response is right for awsJson1.1, `Token` and `ClientId` are both REQUIRED and validated, and the other three error codes are declared and used correctly.

hectorvent (approved):
Thanks @AbhigyaKrishna

PR #1706, opened 2026-07-02
fix(scheduler): forward MessageAttributes in universal sns:publish target

### LINKED ISSUE #1669: [BUG] EventBridge Scheduler: universal aws-sdk:sns:publish target drops MessageAttributes
## Summary

An EventBridge Scheduler schedule whose target is the universal `arn:aws:scheduler:::aws-sdk:sns:publish` action **drops `MessageAttributes`**. The schedule fires and the message is delivered with the correct `Subject` and `Message`, but the `MessageAttributes` supplied in the target `Input` are not published — subscribers receive an empty attribute map.

This breaks any consumer that relies on SNS message attributes — e.g. SQS subscriptions with a `FilterPolicy`, which match on `MessageAttributes`. Such subscribers silently receive nothing.

A **direct** `sns:Publish` API call (not via the scheduler) preserves `MessageAttributes` correctly, so the loss is specific to the scheduler's universal-target invocation.

Verified on `floci/floci:1.5.26`.

## Reproduction

Requirements: `docker`, `aws` CLI.

```bash
#!/usr/bin/env bash
set -euo pipefail
export AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy AWS_DEFAULT_REGION=us-east-1
EP=http://localhost:4566
ROLE=arn:aws:iam::000000000000:role/x

docker run -d --rm --name floci-attr -p 4566:4566 floci/floci:1.5.26 >/dev/null
for _ in $(seq 1 30); do curl -sf "$EP/" >/dev/null 2>&1 && break; sleep 1; done

TOPIC=$(aws --endpoint-url $EP sns create-topic --name t --query TopicArn --output text)
SUBQ=$(aws --endpoint-url $EP sqs create-queue --queue-name sub --query QueueUrl --output text)
SUBARN=$(aws --endpoint-url $EP sqs get-queue-attributes --queue-url "$SUBQ" \
           --attribute-names QueueArn --query 'Attributes.QueueArn' --output text)
# plain SQS subscription (no raw delivery) so the SNS envelope, incl. MessageAttributes, is visible
aws --endpoint-url $EP sns subscribe --topic-arn "$TOPIC" --protocol sqs \
    --notification-endpoint "$SUBARN" >/dev/null

# universal target Input carries Subject + MessageAttributes
cat > target.json <<JSON
{
  "Arn": "arn:aws:scheduler:::aws-sdk:sns:publish",
  "RoleArn": "$ROLE",
  "Input": "{\"TopicArn\":\"$TOPIC\",\"Subject\":\"my-subject\",\"Message\":\"{}\",\"MessageAttributes\":{\"EventName\":{\"DataType\":\"String\",\"StringValue\":\"my-subject\"}}}"
}
JSON

WHEN=$(date -u -d '+8 seconds' +%Y-%m-%dT%H:%M:%S 2>/dev/null || date -u -v+8S +%Y-%m-%dT%H:%M:%S)
aws --endpoint-url $EP scheduler create-schedule --name s \
  --schedule-expression "at($WHEN)" --schedule-expression-timezone UTC \
  --flexible-time-window Mode=OFF --action-after-completion DELETE \
  --target file://target.json >/dev/null

sleep 14
echo "Delivered SNS envelope:"
aws --endpoint-url $EP sqs receive-message --queue-url "$SUBQ" --wait-time-seconds 3 \
  --query 'Messages[0].Body' --output text

docker stop floci-attr >/dev/null
```

## Expected

The delivered SNS envelope contains the supplied attributes, e.g.:

```json
{ "Subject": "my-subject", "Message": "{}",
  "MessageAttributes": { "EventName": { "Type": "String", "Value": "my-subject" } } }
```

## Actual

`Subject` and `Message` are preserved, but `MessageAttributes` is empty:

```json
{ "Subject": "my-subject", "Message": "{}", "MessageAttributes": {} }
```

As a result, an SQS subscription with a `FilterPolicy` on `EventName` would not receive the message.

## Notes

- Related universal-target issue (now fixed in 1.5.26): #1265 — this is a remaining gap in the same `aws-sdk:sns:publish` path: the target `Input`'s `MessageAttributes` aren't forwarded to the underlying `Publish` call.
- A direct `aws sns publish ... --message-attributes ...` preserves attributes, confirming the loss is in the scheduler universal-target mapping, not SNS itself.

## Environment

- Image: `floci/floci:1.5.26`
- Client: AWS CLI v2 against `http://localhost:4566`

ctnnguyen:
## Summary

Fixes #1669.

When a schedule fires with a universal `arn:aws:scheduler:::aws-sdk:sns:publish` target, `MessageAttributes` supplied in the target `Input` were silently dropped. The message was delivered with the correct `Subject` and `Message`, but subscribers received an empty attribute map — breaking any SQS subscription with a `FilterPolicy` that matches on `MessageAttributes`.

The root cause: `invokeUniversalTarget` parsed `TopicArn`, `Message`, `Subject`, and FIFO parameters from `Input` but never read `MessageAttributes`, hardcoding `null` when calling `SnsService.publish`. A direct `sns:Publish` API call preserved attributes correctly, confirming the loss was specific to the scheduler's universal-target path.

The fix reads `MessageAttributes` from `Input` using the same `DataType`/`StringValue` per-entry format that `SnsJsonHandler` uses for direct publish calls, then forwards them through to `SnsService.publish`.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

AWS EventBridge Scheduler universal targets pass `MessageAttributes` through to SNS on publish. Subscribers with `FilterPolicy` scoped to `MessageAttributes` receive messages as expected. This change aligns the emulator with that behavior.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/scheduler/ScheduleInvoker.java
+++ b/src/main/java/io/github/hectorvent/floci/services/scheduler/ScheduleInvoker.java
@@ -14,6 +14,7 @@ import io.github.hectorvent.floci.services.scheduler.model.EcsParameters;
 import io.github.hectorvent.floci.services.scheduler.model.Target;
 import io.github.hectorvent.floci.services.sns.SnsService;
 import io.github.hectorvent.floci.services.sqs.SqsService;
+import io.github.hectorvent.floci.services.sqs.model.MessageAttributeValue;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
@@ -168,8 +169,9 @@ public class ScheduleInvoker {
                 String subject = text(params, "Subject");
                 String messageGroupId = text(params, "MessageGroupId");
                 String messageDeduplicationId = text(params, "MessageDeduplicationId");
+                Map<String, MessageAttributeValue> messageAttributes = parseMessageAttributes(params);
                 String snsRegion = extractRegion(topicArn != null ? topicArn : targetArn, region);
-                snsService.publish(topicArn, targetArn, null, message, subject, null,
+                snsService.publish(topicArn, targetArn, null, message, subject, messageAttributes,
                         messageGroupId, messageDeduplicationId, snsRegion);
                 LOG.debugv("Scheduler delivered to SNS (universal target): {0}", topicArn);
             }
@@ -183,6 +185,20 @@ public class ScheduleInvoker {
             }
             default -> LOG.warnv("Scheduler: unsupported universal target action: {0}", serviceAction);
         }
+    }
+
+    private Map<String, MessageAttributeValue> parseMessageAttributes(JsonNode params) {
+        JsonNode attrsNode = params.path("MessageAttributes");
+        if (!attrsNode.isObject()) {
+            return null;
+        }
+        Map<String, MessageAttributeValue> attributes = new HashMap<>();
+        attrsNode.fields().forEachRemaining(entry -> {
+            String dataType = entry.getValue().path("DataType").asText("String");
+            String stringValue = entry.getValue().path("StringValue").asText(null);
+            attributes.put(entry.getKey(), new MessageAttributeValue(stringValue, dataType));
+        });
+        return attributes.isEmpty() ? null : attributes;
     }
 
     private static String text(JsonNode node, String field) {
--- a/src/main/java/io/github/hectorvent/floci/services/scheduler/ScheduleInvoker.java
+++ b/src/main/java/io/github/hectorvent/floci/services/scheduler/ScheduleInvoker.java
@@ -12,6 +12,7 @@ import io.github.hectorvent.floci.services.lambda.model.InvocationType;
 import io.github.hectorvent.floci.services.scheduler.model.EventBridgeParameters;
 import io.github.hectorvent.floci.services.scheduler.model.EcsParameters;
 import io.github.hectorvent.floci.services.scheduler.model.Target;
+import io.github.hectorvent.floci.services.sns.SnsMessageAttributes;
 import io.github.hectorvent.floci.services.sns.SnsService;
 import io.github.hectorvent.floci.services.sqs.SqsService;
 import io.github.hectorvent.floci.services.sqs.model.MessageAttributeValue;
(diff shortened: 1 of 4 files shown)


hectorvent:
Clean fix, and the parsing matches the model. SNS's [`MessageAttributeValue`](https://github.com/boto/botocore/blob/develop/botocore/data/sns/2010-03-31/service-2.json) declares exactly `DataType` (required), `StringValue` and `BinaryValue` as a base64 blob, which is what you read, the constructor arg order is right on both branches, and defaulting `DataType` to `Binary` or `String` mirrors `SnsJsonHandler`. Passing `null` for no attributes is fine since `SnsService.publish` already treats that as none. Tests green on the branch, 8/8, and they cover string, binary and the empty case.

One suggestion. [`ScheduleInvoker.parseMessageAttributes`](https://github.com/floci-io/floci/blob/a78103dfe488040ccf7b8689baa4a6a54adb972d/src/main/java/io/github/hectorvent/floci/services/scheduler/ScheduleInvoker.java#L137) is a near verbatim copy of the private method in `SnsJsonHandler`, and the two have already drifted a little: yours returns `null` where the other returns an empty map, reads `StringValue` with `asText(null)` instead of `asText()`, and lets the base64 `IllegalArgumentException` escape where the other raises `InvalidParameterValue`. Not a crash, `ScheduleDispatcher` catches around `invoker.invoke`, so a bad `BinaryValue` just drops that one delivery. But it is the kind of drift that grows.

Pulling the existing one into somewhere both can reach would make this a one liner:

```java
// SnsMessageAttributes.java (or a static on SnsService)
public static Map<String, MessageAttributeValue> parse(JsonNode attrsNode) { ... }

// ScheduleInvoker
Map<String, MessageAttributeValue> messageAttributes = SnsMessageAttributes.parse(params.path("MessageAttributes"));
```

Also, not for this PR, but the `sqs:sendMessage` universal target one case below has the same gap. It drops `MessageAttributes` and hardcodes `DelaySeconds` to `0`, both of which are declared members of `SendMessage`. Worth a follow up issue.

hectorvent:
@ctnnguyen I got another PR that touched the same file and make a conflict with this one, please resolve it and I will try to get this merged as soon as possible.

ctnnguyen:
> @ctnnguyen I got another PR that touched the same file and make a conflict with this one, please resolve it and I will try to get this merged as soon as possible.

@hectorvent Ready for review!

hectorvent (approved):
Thanks @ctnnguyen

PR #1713, opened 2026-07-03
fix(s3): preserve question marks in copy source keys

Tomoya0k:
## Summary
This PR fixes S3 `CopyObject` and `UploadPartCopy` when `x-amz-copy-source` references a source key that contains a literal question mark.

The reported failure was that Floci returned `NoSuchKey` for a source object that already existed and could be resolved with `HeadObject`, as long as the copied key contained `?` and the header used either the AWS-style encoded form (`%3F`) or the raw literal character.

## Root cause
The copy-source parsing logic decoded the entire header value before splitting bucket, key, and optional query parameters.

That meant a header like:

`/bucket/folder/file%20with%20question%20%3F.txt`

was decoded first into:

`/bucket/folder/file with question ?.txt`

After that, the parser treated the first `?` as the beginning of a query string and truncated the source key to `folder/file with question `. The copy path then looked up the wrong key and returned `NoSuchKey`.

This also made raw `x-amz-copy-source` values containing a literal `?` behave the same way.

## What changed
The parsing flow now:

- splits the raw `x-amz-copy-source` into bucket and path before URL decoding
- decodes bucket, object-key, and query components separately
- only interprets a trailing `?` section as metadata when it actually contains `versionId=`
- preserves literal question marks in the object key when no `versionId` query is present
- keeps existing support for copy-source version selection via `?versionId=...`

The same parsing fix was applied to both `CopyObject` and `UploadPartCopy` so the two operations stay aligned.

## Tests
Added an integration regression test that verifies both of these succeed:

- encoded copy source: `file%20with%20question%20%3F.txt`
- raw copy source: `file with question ?.txt`

The test also verifies the copied object content after each operation.

## Validation
I could not execute the Maven test suite in this environment because `java` is not installed and `JAVA_HOME` is not configured, so local runtime validation is still pending.

Recommended follow-up commands:

```powershell
.\mvnw.cmd -Dtest=S3IntegrationTest#copyObjectWithQuestionMarkInSourceKeySucceeds test
.\mvnw.cmd -Dtest=S3CopyObjectVersionedIntegrationTest test
```

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -1738,21 +1738,13 @@ public class S3Controller {
                                       String contentType, HttpHeaders httpHeaders) {
         // copySource format: /bucket/key or bucket/key, where key is URL-encoded
         String source = copySource.startsWith("/") ? copySource.substring(1) : copySource;
-        
-        // URL decode the entire source first, then split
-        String decodedSource;
-        try {
-            decodedSource = URLDecoder.decode(source, StandardCharsets.UTF_8);
-        } catch (IllegalArgumentException e) {
-            throw new AwsException("InvalidArgument", "Invalid copy source: " + copySource, 400);
-        }
-        int slashIndex = decodedSource.indexOf('/');
+
+        int slashIndex = source.indexOf('/');
         if (slashIndex <= 0) {
             throw new AwsException("InvalidArgument", "Invalid copy source: " + copySource, 400);
         }
-        String sourceBucket = decodedSource.substring(0, slashIndex);
-        String pathAfterBucket = decodedSource.substring(slashIndex + 1);
-        ParsedCopySource sourceObject = parseCopySourceObject(pathAfterBucket);
+        String sourceBucket = decodeCopySourceComponent(source.substring(0, slashIndex), copySource);
+        ParsedCopySource sourceObject = parseCopySourceObject(source.substring(slashIndex + 1), copySource);
         String copyContentEncoding = toPersistedContentEncoding(httpHeaders.getHeaderString("Content-Encoding"));
         String copyContentDisposition = httpHeaders.getHeaderString("Content-Disposition");
         String copyCacheControl = httpHeaders.getHeaderString("Cache-Control");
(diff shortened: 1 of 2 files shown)


hectorvent:
Good news first: the parsing change is correct. Splitting the raw header before decoding is exactly what moto does (`urlparse(copy_source)` then `unquote(path)`), and I exercised it against the running emulator, both the encoded and the raw form copy fine once the source object really exists under the literal key.

You mentioned you couldn't run the suite, so, the new test fails. Everything else in S3 is green, 115/116, plus `S3CopyObjectVersionedIntegrationTest` 6/6.

It's the test, not `S3Controller`. RestAssured URL encodes by default, so the setup `PUT` stores the object under the literal string `folder/file%20with%20question%20%3F.txt` (it escapes the `%` to `%25`). I listed the bucket to confirm. The encoded copy source then correctly resolves to `folder/file with question ?.txt`, which genuinely isn't there, so `NoSuchKey` is the right answer to the wrong question. Turning encoding off makes it pass:

```java
given().urlEncodingEnabled(false)
    .contentType("text/plain")
    .body("copy test")
.when()
    .put("/" + sourceBucket + "/" + encodedSrcKey)
.then()
    .statusCode(200);
```

One gap in the new leniency. [`parseCopySourceObject`](https://github.com/floci-io/floci/blob/71eca4a63ac67f8e8b4e1f18872ac75c4b5b6cdc/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java#L2656) uses `indexOf('?')`, so a raw literal `?` in the key combined with a real `?versionId=` breaks. The first `?` becomes the query start, the remainder fails the `"versionId".equals(name)` check, and the whole string ends up as the key. Confirmed on a versioned bucket:

```
COPY q%3F.txt?versionId=…  ->  CopyObjectResult   (works)
COPY q?.txt?versionId=…    ->  NoSuchKey          (fails)
```

`lastIndexOf('?')` fixes it and leaves everything else intact, a raw `f?.txt` with no version still finds `.txt` as the query, fails the `versionId=` check, and falls back to the full path. Worth noting the encoded form is what every SDK actually sends and that one works, including with `versionId`, which was broken before this PR. Nice catch on that.

Two small ones. Since `extractVersionId` returning null now keeps the query glued to the key, `/bucket/key?foo=bar` resolves to `key?foo=bar` where it used to resolve to `key`, and moto truncates at the first `?` regardless. No SDK sends that, just flagging the divergence. And `decodeCopySourceComponent` uses `URLDecoder`, which is form encoding, so it turns `+` into a space. Pre-existing and latent since SDKs send `%2B`, but same family of bug.

hectorvent:
Thanks @Tomoya0k,

Closing in favor of #1924

PR #1718, opened 2026-07-03
fix(lambda): return a downloadable GetFunction Code.Location

Jongsic:
## Summary

`GetFunction` (`GET /2015-03-31/functions/{name}`) returned a hardcoded, fabricated real-AWS URL for `Code.Location`:

```java
code.put("Location", "https://awslambda-" + region + "-tasks.s3." + region
        + ".amazonaws.com/" + fn.getFunctionName());
```

That URL points at real AWS S3, is not presigned, and 404s. Any client that follows `Code.Location` to download the deployment package — CLIs, tooling, and browser-only UIs — cannot fetch the code. Real AWS and LocalStack both return a working, downloadable URL.

**Root cause:** Floci extracts the uploaded zip to a directory and discards the original bytes, so there was nothing to serve.

**Fix:** on every *successful* code deploy, persist the exact original zip bytes into Floci's own S3 under a per-region tasks bucket (`awslambda-<region>-tasks`), keyed by `snapshots/<account>/<functionName>`. `GetFunction` now builds a **path-style URL on the local endpoint** from the incoming request via `UriInfo`, so it targets the same Floci the client is already talking to (mirrors LocalStack). `RepositoryType: "S3"` stays truthful.

- **Byte-exact:** the served package hashes to the already-stored `CodeSha256` (`sha256(originalZipBytes)`). Re-zipping the extracted dir would break that.
- **Reuses existing infra:** internal `S3Service` + `S3Controller`'s path-style `GET /{bucket}/{key:.+}` + global CORS. No new endpoint, no new serving code.
- **Centralized** in `extractZipCode`, so it covers `ZipFile`, `S3Bucket`/`S3Key` source, and the reactive S3-update listener with one change. Stored only *after* handler-file verification passes, so a rejected deploy leaves no orphan object.

**Lifecycle:** the stable key means redeploys overwrite in place (no stale accumulation), and `DeleteFunction` best-effort deletes the object.

**Scope / tradeoffs:**
- Hot-reload functions (`S3Bucket == "hot-reload"`) have no zip, so no object is stored — `Code.Location` 404s for them, matching the "no downloadable package" reality.
- Functions deployed **before** this change 404 until their next deploy (both `NoSuchBucket`/`NoSuchKey` → clean 404, same "can't download" state as before, just against the local endpoint).
- The `awslambda-<region>-tasks` bucket becomes visible in `ListBuckets` (same as LocalStack). Hiding service buckets is a separate follow-up, out of scope here.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- **Incorrect behavior:** `GetFunction` returned a fabricated real-AWS S3 URL for `Code.Location` that 404s and is not presigned, so the deployment package could not be downloaded by any client following the URL.
- **Verified against:** AWS CLI `aws-cli/2.31.38` (`create-function` with `--zip-file`, then `get-function`), following the returned `Code.Location` with `curl`. Response is `200 application/zip` and the downloaded bytes' SHA-256 matches the function's `CodeSha256` (byte-exact). Behavior mirrors LocalStack's path-style `Code.Location`.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -24,6 +24,8 @@ import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.UriBuilder;
+import jakarta.ws.rs.core.UriInfo;
 import org.jboss.logging.Logger;
 
 import java.time.Instant;
@@ -79,6 +81,7 @@ public class LambdaController {
     @GET
     @Path("/functions/{functionName}")
     public Response getFunction(@Context HttpHeaders headers,
+                                @Context UriInfo uriInfo,
                                 @PathParam("functionName") String functionName) {
         String region = regionResolver.resolveRegion(headers);
         LambdaFunction fn = lambdaService.getFunction(region, functionName);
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Thanks @Jongsic, 
Solid fix.

Issue #1637, opened 2026-06-29
[FEAT] SES v2 contact list & subscription management — Contacts CRUD (staged rollout)


okinaka:
## Service

SES (v2 / REST JSON)

## API Action / Feature

Built-in contact-list & subscription management, delivered in stages to keep each
PR small and reviewable. The container half (**ContactList CRUD**) is already
implemented; this issue tracks the remaining work, with **Contact CRUD** as the
minimum target.

**Staged plan**

- [x] **Stage 1 — ContactList CRUD** (`CreateContactList` / `GetContactList` /
  `ListContactLists` / `UpdateContactList` / `DeleteContactList`), incl. topics,
  the one-list-per-account limit, and AWS input constraints. *(implemented; being upstreamed)*
- [x] **Stage 2 — Contact CRUD** (`CreateContact` / `GetContact` / `ListContacts` /
  `UpdateContact` / `DeleteContact`), incl. per-topic `TopicPreferences`,
  `TopicDefaultPreferences`, `UnsubscribeAll`, and `AttributesData`. ← minimum target
- [x] **Stage 3 — `SendEmail` `ListManagementOptions`** (`ContactListName` /
  `TopicName`): suppress `OPT_OUT` recipients and add the unsubscribe handling.

Each stage ships as its own PR.

## AWS Documentation

- ContactList: [CreateContactList](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_CreateContactList.html) · [GetContactList](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_GetContactList.html) · [ListContactLists](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_ListContactLists.html) · [UpdateContactList](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_UpdateContactList.html) · [DeleteContactList](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_DeleteContactList.html)
[...]

PR #1723, opened 2026-07-04
feat(ses): implement Contact CRUD for SES v2

okinaka:
## Summary

Stage 2 of #1637.

Implements the SES v2 Contact management APIs (the per-subscriber half of contact-list/subscription management), under `/v2/email/contact-lists/{name}/contacts`:

| Method | Path | Operation |
| --- | --- | --- |
| `POST` | `/v2/email/contact-lists/{name}/contacts` | `CreateContact` |
| `POST` | `/v2/email/contact-lists/{name}/contacts/list` | `ListContacts` |
| `GET` | `/v2/email/contact-lists/{name}/contacts/{email}` | `GetContact` |
| `PUT` | `/v2/email/contact-lists/{name}/contacts/{email}` | `UpdateContact` |
| `DELETE` | `/v2/email/contact-lists/{name}/contacts/{email}` | `DeleteContact` |

- New `Contact` / `TopicPreference` models; a `ses-contacts.json` store wired through `StorageFactory` (region- and list-scoped key).
- A contact carries per-topic subscription preferences, an unsubscribe-all flag, and free-form `AttributesData`. `TopicDefaultPreferences` is not stored — it is derived at read time from the list's topics for topics the contact hasn't set explicitly.
- `DeleteContactList` cascade-deletes its contacts, and a shared lock serializes contact create/update/delete against contact-list deletion so a concurrent delete can't leave an orphaned contact.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Behaviour and every error string were verified against real AWS (SES v2, boto3 / raw signed requests, us-east-1).

**CRUD error shapes.** A duplicate address is `AlreadyExistsException` / `<email> already exists in List.`; an unknown contact is `NotFoundException` / `<email> doesn't exist in List.`; a preference for a topic the list doesn't define is `BadRequestException` / `List: <list> doesn't contain Topic: <topic>`.

**EmailAddress validation** (probe-confirmed, and evaluated *before* contact-list existence):

| Input | Result |
| --- | --- |
| missing / `null` | Smithy validation error — `1 validation error detected: Value at 'emailAddress' failed to satisfy constraint: Member must not be null` |
| `""` (blank) | `EmailAddress can't be blank.` |
| non-blank malformed | `EmailAddress <x> is invalid` |

**Combined-violation order** (probe-confirmed): invalid `SubscriptionStatus` enum → EmailAddress format → contact-list existence → unknown topic.

**Boolean / String coercion** (Jackson-backed on AWS, consistent with the existing `SendingEnabled` handling): a JSON string coerces to `true` for the Boolean `UnsubscribeAll`, while a number/null/array/object is a `SerializationException` (`NUMBER_VALUE can not be converted to a Boolean`). `AttributesData` is a String; a non-string is rejected the same way (`NUMBER_VALUE`/`TRUE_VALUE can not be converted to a String`).

**UpdateContact semantics** (probe-confirmed): `TopicPreferences` merge by topic name (omitting keeps existing entries), while `AttributesData` and `UnsubscribeAll` are replaced (omitting clears / resets them).

`ListContacts` uses `POST .../contacts/list` with the filter/pagination request body; Floci returns all of a list's contacts (server-side `Filter`/`PageSize` are not yet implemented) but still rejects a malformed or non-object body like the other v2 endpoints. Timestamps are emitted as epoch seconds. This is a v2-only change.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SesContactTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SesContactTest.java
@@ -1,0 +1,162 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+import software.amazon.awssdk.services.sesv2.SesV2Client;
+import software.amazon.awssdk.services.sesv2.model.AlreadyExistsException;
+import software.amazon.awssdk.services.sesv2.model.BadRequestException;
+import software.amazon.awssdk.services.sesv2.model.CreateContactListRequest;
+import software.amazon.awssdk.services.sesv2.model.CreateContactRequest;
+import software.amazon.awssdk.services.sesv2.model.DeleteContactListRequest;
+import software.amazon.awssdk.services.sesv2.model.DeleteContactRequest;
+import software.amazon.awssdk.services.sesv2.model.GetContactRequest;
+import software.amazon.awssdk.services.sesv2.model.GetContactResponse;
+import software.amazon.awssdk.services.sesv2.model.NotFoundException;
+import software.amazon.awssdk.services.sesv2.model.SubscriptionStatus;
+import software.amazon.awssdk.services.sesv2.model.Topic;
+import software.amazon.awssdk.services.sesv2.model.TopicPreference;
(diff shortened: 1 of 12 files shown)


okinaka:
@greptile-apps thanks — both addressed in a follow-up commit.

- `getContact`/`listContacts` now return the resolved `ContactList` alongside the contact(s) (`ContactWithList`/`ContactsWithList`), so the controller no longer makes a second `getContactList` call and the read is a single consistent lookup — closing the TOCTOU window where a concurrent delete could turn a successful read into "List not found".
- `deleteContact` now re-checks list existence inside `contactMutationLock`, matching `createContact`/`updateContact`, so a concurrent `deleteContactList` surfaces "list not found" instead of "contact not found".

hectorvent (approved):
Thanks @okinaka

PR #1733, opened 2026-07-05
feat(scheduler): dispatch ECS targets to RunTask

jvanzyl:
Summary:
- Add EventBridge Scheduler ECS target parameter parsing and response round-tripping.
- Dispatch ECS scheduler targets through ECS RunTask with awsvpc network configuration.
- Add unit, integration, and Java SDK coverage for ECS target parameters.

Verification:
- ./mvnw -q -Dtest=ScheduleInvokerTest,SchedulerIntegrationTest test

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SchedulerTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SchedulerTest.java
@@ -6,6 +6,7 @@ import software.amazon.awssdk.services.scheduler.model.*;
 import software.amazon.awssdk.services.scheduler.model.Tag;
 
 import java.time.Instant;
+import java.util.Map;
 
 import static org.assertj.core.api.Assertions.*;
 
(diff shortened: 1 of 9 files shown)


hectorvent:
Thanks @jvanzyl, nice work. Verified against the scheduler `service-2.json` and the wire casing is exactly right, including the trap most implementations miss: `awsvpcConfiguration` is lowercase in the model while the members inside it are PascalCase. The RunTask dispatch through the real ECS service is clean, TaskCount defaults to 1 per the model, and the three level test coverage (unit, integration, SDK v2 round trip) is appreciated.

One suggestion before merge: the model declares 14 `EcsParameters` members and this round trips 5. The other 9 (`Group`, `ReferenceId`, `Tags`, `PropagateTags`, `EnableECSManagedTags`, `EnableExecuteCommand`, `CapacityProviderStrategy`, `PlacementConstraints`, `PlacementStrategy`) are silently dropped, so a client that sets them gets them missing from `GetSchedule`:

```
Terraform sets Group = "my-group"  →  GetSchedule omits it  →  perpetual diff on every plan
```

Even without dispatching them, storing and echoing them back is cheap and removes the IaC drift risk. If you do, wiring `Group` through to RunTask (it is hardcoded to "scheduler" right now) comes almost for free.

Non blocking: the model's required and range constraints (`TaskDefinitionArn` required, `Subnets` 1 to 16, `SecurityGroups` max 5) are not validated, but that matches this controller's existing leniency, fine as scope.

Happy to merge once the round trip question is settled either way.

hectorvent (approved):
Thanks @jvanzyl

PR #1735, opened 2026-07-05
fix(rds): preserve network placement metadata

jvanzyl:
Summary:
- Use the signed request region when resolving RDS subnet placement.
- Persist DB instance VPC security group IDs from create and modify calls.
- Return VPC security group membership from DescribeDBInstances.

Verification:
- ./mvnw -q -Dtest=RdsServiceTest,RdsQueryHandlerTest test

### DIFF
--- a/docs/services/docdb.md
+++ b/docs/services/docdb.md
@@ -21,6 +21,7 @@ The management API shares the RDS Query endpoint (`POST /` with an `Action=` par
 | --- | --- |
 | `CreateDBCluster` | Create a DocumentDB cluster and start a MongoDB container |
 | `DescribeDBClusters` | List clusters and their connection details |
+| `DescribeDBClusterSnapshots` | - |
 | `DeleteDBCluster` | Stop and remove a cluster (must have no instances) |
 | `ModifyDBCluster` | Update engine version or IAM auth setting |
 | `CreateDBInstance` | Add an instance to a cluster |
--- a/docs/services/elasticache.md
+++ b/docs/services/elasticache.md
@@ -23,6 +23,8 @@ Floci manages real Valkey/Redis Docker containers and proxies TCP connections to
 | `CreateCacheCluster` | - |
 | `DescribeCacheClusters` | - |
 | `DeleteCacheCluster` | - |
+| `DescribeCacheSubnetGroups` | - |
+| `DescribeCacheParameterGroups` | - |
 <!-- floci:actions:end -->
 
 ## Configuration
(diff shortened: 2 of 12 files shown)


hectorvent:
Core change checks out. `VpcSecurityGroupIdList` declares its member with `"locationName": "VpcSecurityGroupId"` in botocore's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/rds/2014-10-31/service-2.json), so the new regex matches what the SDK actually sends, the `VpcSecurityGroupMembership` response shape and its `active` status match, and `DBSubnetGroupArn` is a real member. Routing the signed region through is safe too, since `Ec2Service.describeSubnets` calls `ensureDefaultResources(region)` and seeds that region's default VPC. Ran the RDS tests on the branch, 107/107 green.

Three things I'd tidy before merge.

[`dbSubnetGroupForInstance`](https://github.com/floci-io/floci/blob/ec0446450a3c964f535d14b232079387d288d43a/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java#L847) picked up two dead branches. The new early return already catches null, blank and `default`, so the next guard is always true:

```java
if (groupName == null || groupName.isBlank() || "default".equalsIgnoreCase(groupName)) {
    return fallbackSubnetGroup(instance, "default", "default");
}
if (groupName != null && !groupName.isBlank()) {   // always true now
```

The [`resolveDbSubnetGroupView`](https://github.com/floci-io/floci/blob/ec0446450a3c964f535d14b232079387d288d43a/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java#L862) fallback below it is dead for the same reason. For a non default name it runs the same `subnetGroups.get(name).orElseThrow(...)` as the `getDbSubnetGroup` call above, so if the first threw the second throws identically. Both blocks can go.

The default subnet group description now disagrees with itself. `DescribeDBInstances` renders `"default"` through the new fallback, `DescribeDBSubnetGroups` still renders `"default subnet group"` from [`RdsService.java:1293`](https://github.com/floci-io/floci/blob/ec0446450a3c964f535d14b232079387d288d43a/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java#L1293). I couldn't pin AWS's exact string from a local source, moto doesn't model a default subnet group, so I'm not saying which is right, only that the two responses contradict each other for the same resource.

The `region == null` ternaries are dead in production. `resolveRegion` falls back to `defaultRegion` and never returns null, and `AwsQueryController` is the only production caller, so the two arg `handle` and everything behind it exists to keep the old tests compiling. Same for the [`vpcSecurityGroupIds.isEmpty()`](https://github.com/floci-io/floci/blob/ec0446450a3c964f535d14b232079387d288d43a/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java#L184) split, since the five arg `modifyDbInstance` already guards with `if (vpcSecurityGroupIds != null && !isEmpty())`. This takes `createDbInstance` from five overloads to eight. Updating the test call sites instead would keep it at one.

hectorvent (approved):

PR #1736, opened 2026-07-05
fix(s3): preserve literal '+' in object keys and support encoding-type=url in list-objects

### LINKED ISSUE #1589: [BUG] S3 ListObjectsV2 decodes '+' to space in returned object keys
## Service
    S3
    ## AWS API Action
    ListObjectsV2

    ## Expected behavior
    Real S3 returns object keys verbatim. A key containing a literal '+' must come
    back with the '+' intact when EncodingType is not specified.

    ## Actual behavior
    floci returns the key with '+' decoded to a space, so a follow-up GetObject/
    HeadObject on the listed key fails with NoSuchKey (404). The object itself is
    stored correctly (HEAD/GET on the literal '+' key succeed).

    ## Reproduction
    docker run -d -p 4566:4566 floci/floci:latest   # floci 1.6.0 native
    E=http://localhost:4566
    aws --endpoint-url $E s3 mb s3://b
    aws --endpoint-url $E s3api put-object --bucket b --key "run_id=2026-06-26T00:00:00+00:00/f.log" --body /etc/hostname
    aws --endpoint-url $E s3api head-object       --bucket b --key "run_id=2026-06-26T00:00:00+00:00/f.log"   # OK
    aws --endpoint-url $E s3api list-objects-v2    --bucket b --prefix run_id=                                  # Key => "run_id=2026-06-26T00:00:00 00:00/f.log"  (+ -> space)
    aws --endpoint-url $E s3api list-objects-v2    --bucket b --prefix run_id= --encoding-type url              # Key => "...%2B00:00..."  (correct)

    ## Environment
    - Floci image: floci/floci:latest  (observed: floci 1.6.0 native, Quarkus 3.36.0)
    - Client: AWS CLI / boto3 1.38; reproduces with default and path-style addressing
    - Storage mode: memory (default)
    - Real-world impact: Apache Airflow S3 remote task logs (run_id contains '+00:00')
      list-then-read, so completed task logs are unreadable.

mansiverma897993:
## Summary

Closes #1589

`ListObjectsV2` returned object keys with a literal `+` decoded to a space (e.g. Airflow task-log keys like `run_id=2026-06-26T00:00:00+00:00/f.log`), so a list-then-get round trip failed with `NoSuchKey` even though the object itself was stored correctly.

- Preserve literal `+` when decoding object keys, copy sources and raw paths: `+` only means space in form-encoded query strings, never in URL paths, so `URLDecoder.decode` is guarded with `replace("+", "%2B")`.
- Support `encoding-type=url` in `ListObjects`, `ListObjectsV2` and `ListObjectVersions`: `Key`, `Prefix`, `Delimiter`, `StartAfter` (v2), `Marker`/`NextMarker` (v1) and `KeyMarker`/`NextKeyMarker` (versions) are percent-encoded via a new `maybeEncode` helper. `ContinuationToken`/`NextContinuationToken` and version ids stay verbatim per the S3 spec, since they are opaque tokens a client feeds straight back (review feedback).
- Do not double-decode query parameters that the JAX-RS layer has already decoded once, so keys containing literal percent-escapes (e.g. `%20`) survive a list round trip (review feedback from @abanna).
- `ListObjectVersions` honours `delimiter`, returning `CommonPrefixes` and the encoded `<Delimiter>` (review feedback from @abanna).
- `max-keys` on `ListObjectVersions` now counts each `Version`, each `DeleteMarker` and each `CommonPrefixes` group as one entry, instead of collapsing all versions of a key into a single slot. A page can therefore stop part-way through a key, so a truncated response carries `NextVersionIdMarker` next to `NextKeyMarker`, and the request accepts `version-id-marker` to resume inside a key. `version-id-marker` without `key-marker` is rejected with `InvalidArgument`, as on AWS (review feedback from @abanna and Greptile).
- Rebased onto current `main` (copy-source separator refactor, auth enforcement, CloudTrail data events, website hosting) so the branch is mergeable and the diff stays scoped to this fix.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior being fixed:

1. Keys containing `+` came back from `ListObjectsV2` with the `+` turned into a space, and `encoding-type=url` responses double-encoded already-escaped prefixes (`%2B` → `%252B`). Real S3 returns keys verbatim when no `EncodingType` is specified and percent-encodes only `Delimiter`/`Prefix`/`Key`/`StartAfter` when `encoding-type=url` is set.
2. `ListObjectVersions` ignored `delimiter`, so `?versions&delimiter=/` returned nested keys as `Version` entries instead of grouping them into `CommonPrefixes`.
3. `?versions&max-keys=1` against a key with two versions returned both `Version` entries with `IsTruncated=false`. AWS counts every `Version`/`DeleteMarker`/`CommonPrefixes` entry against `max-keys` and returns `NextKeyMarker` plus `NextVersionIdMarker` so the client can resume mid-key.

Verified against the AWS CLI reproduction from #1589 (`put-object` / `head-object` / `list-objects-v2 [--encoding-type url]`) and covered by the integration tests `listObjectsV2PreservesPlusCharacter`, `listObjectsV2DoesNotDoubleDecodePercentEscapes`, `listObjectVersionsSupportsDelimiterAndUrlEncoding`, `listObjectVersionsPagesCombinedVersionsAndPrefixes` and `listObjectVersionsPagesEachVersionEntry`.

## Checklist

- [ ] `./mvnw test` passes locally (no local JDK 25 available — relying on CI for the full suite)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -443,7 +443,7 @@ public class S3Controller {
             }
             if (hasQueryParam(uriInfo, "versions")) {
                 s3Service.authorizeBucketRead(bucket, "s3:ListBucketVersions", authorization);
-                return handleListObjectVersions(bucket, prefix, maxKeys, keyMarker);
+                return handleListObjectVersions(bucket, prefix, delimiter, maxKeys, keyMarker, encodingType);
             }
             if (hasQueryParam(uriInfo, "location")) {
                 s3Service.authorizeBucketRead(bucket, "s3:GetBucketLocation", authorization);
@@ -531,8 +531,8 @@ public class S3Controller {
                     .raw("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
                     .start("ListBucketResult", AwsNamespaces.S3)
                     .elem("Name", bucket)
-                    .elem("Prefix", prefix != null ? prefix : "")
-                    .elem("Delimiter", delimiter)
+                    .elem("Prefix", maybeEncode(prefix != null ? prefix : "", encodingType))
+                    .elem("Delimiter", maybeEncode(delimiter, encodingType))
                     .elem("MaxKeys", max);
             if (v2) {
                 xml.elem("KeyCount", objects.size() + commonPrefixes.size());
@@ -540,7 +540,7 @@ public class S3Controller {
             xml.elem("IsTruncated", result.isTruncated());
             for (S3Object obj : objects) {
                 xml.start("Contents")
-                   .elem("Key", obj.getKey())
+                   .elem("Key", maybeEncode(obj.getKey(), encodingType))
                    .elem("LastModified", ISO_FORMAT.format(obj.getLastModified()))
                    .elem("ETag", obj.getETag())
                    .elem("Size", obj.getSize())
@@ -549,7 +549,7 @@ public class S3Controller {
             }
             for (String cp : commonPrefixes) {
                 xml.start("CommonPrefixes")
-                   .elem("Prefix", cp)
+                   .elem("Prefix", maybeEncode(cp, encodingType))
                    .end("CommonPrefixes");
             }
             if (encodingType != null) {
@@ -563,12 +563,12 @@ public class S3Controller {
                     xml.elem("NextContinuationToken", result.nextContinuationToken());
                 }
                 if (startAfter != null) {
-                    xml.elem("StartAfter", startAfter);
+                    xml.elem("StartAfter", maybeEncode(startAfter, encodingType));
                 }
             } else {
-                xml.elem("Marker", marker != null ? marker : "");
+                xml.elem("Marker", maybeEncode(marker != null ? marker : "", encodingType));
                 if (result.isTruncated() && result.nextContinuationToken() != null) {
-                    xml.elem("NextMarker", result.nextContinuationToken());
+                    xml.elem("NextMarker", maybeEncode(result.nextContinuationToken(), encodingType));
                 }
             }
             xml.end("ListBucketResult");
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
```diff
@@ -394,16 +394,16 @@ public Response listObjects(@PathParam("bucket") String bucket,
                     .raw("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
                     .start("ListBucketResult", AwsNamespaces.S3)
                     .elem("Name", bucket)
-                    .elem("Prefix", prefix != null ? prefix : "")
-                    .elem("Delimiter", delimiter)
+                    .elem("Prefix", maybeEncode(prefix != null ? prefix : "", encodingType))
```
abanna, line 536: Thanks for adding coverage around a fiddly S3 edge. One thing that still looks off: `prefix` and `delimiter` are passed into the service without decoding `%2B`, so filtering happens on the literal escaped value and `maybeEncode` emits `%252B`. Could we decode percent escapes once while preserving a literal `+` before filtering and returning these fields? The test should cover the actual object match too.

mansiverma897993: Hi! I have updated the PR to decode query parameters (like prefix and delimiter) while preserving literal + characters before they are passed to the filtering service.
I also updated the integration test (listObjectsV2PreservesPlusCharacter) to disable REST-Assured's automatic URL encoding and verified that the matched object is correctly returned in the XML response.

abanna: Thanks for turning this around so quickly—the `%2B` object-match coverage is exactly what I was hoping to see. I think there’s one small wrinkle left: `@QueryParam` values are already decoded, so `decodeQueryParam` can decode a literal percent escape a second time (`%2520` → `%20` → space). Could you add that regression case and avoid the second decode? Then this looks good to me.

mansiverma897993: Thanks for the feedback! I have removed the redundant manual decoding in S3Controller to avoid double-decoding and added the requested regression test case (listObjectsV2DoesNotDoubleDecodePercentEscapes) to S3IntegrationTest.java. :) now take a look

hectorvent:

mansiverma897993:
@hectorvent can you check now ? I have updated this

src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
abanna: Thanks for making that change. I checked the `%2520` case and it covers the double-decoding issue I was worried about. The code side looks good to me. It looks like the branch still needs a rebase and the failing sdk-test-go check addressed before it is ready to merge.

mansiverma897993:
@abanna @hectorvent I've rebased the branch onto main, resolved all merge conflicts, and updated the PR. Ready for review!

src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
```diff
@@ -428,7 +428,7 @@ public Response listObjects(@PathParam("bucket") String bucket,
             }
             if (hasQueryParam(uriInfo, "versions")) {
                 s3Service.authorizeBucketRead(bucket, "s3:ListBucketVersions", authorization);
-                return handleListObjectVersions(bucket, prefix, maxKeys, keyMarker);
+                return handleListObjectVersions(bucket, prefix, maxKeys, keyMarker, encodingType);
```
abanna: Thanks for rebasing this — I rechecked the current head and the original `%2B` / `%2520` cases look good now. One small compatibility gap in the new `ListObjectVersions` encoding path: `?versions&delimiter=/&encoding-type=url` still drops the delimiter, so it returns `dir%2Fone.txt` as a `Version` and omits `<Delimiter>%2F</Delimiter>` / `CommonPrefixes`. AWS groups version results by delimiter and includes the encoded delimiter here. Could we wire that through and add a regression test?

mansiverma897993:
@abanna Thanks for catching that! I've wired delimiter through listObjectVersions, added support for grouping CommonPrefixes and URL-encoding <Delimiter>, and added a regression test (listObjectVersionsSupportsDelimiterAndUrlEncoding). The branch has been updated!

src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
abanna: Thanks for wiring this through — the delimiter and `encoding-type=url` behavior now looks right. One pagination edge case still looks open: `max-keys` is applied only to direct versions, while `CommonPrefixes` are returned unbounded. With `?versions&delimiter=/&max-keys=1` and keys under `a/` and `b/`, Floci returns both prefixes with `IsTruncated=false`. AWS counts each `CommonPrefixes` group as one result against `max-keys`. Could we page the combined version/prefix result set and add that regression case?

mansiverma897993:
@abanna Good catch! Sorry for that I've updated listObjectVersions so max-keys pages across the combined set of versions and CommonPrefixes together (with each prefix counting as 1 entry against max-keys). I also added the regression test (listObjectVersionsPagesCombinedVersionsAndPrefixes). The branch has been updated!

mansiverma897993:
@hectorvent is there still any need to change something ?

abanna:
Thanks for the updates — the delimiter and combined-prefix cases look good now. I found one remaining pagination issue with real versioned objects: after two versions of the same key, `?versions&max-keys=1` returns both `<Version>` entries with `IsTruncated=false`. I verified that locally. AWS paginates version entries here, and resuming within a key needs both `NextKeyMarker` and `NextVersionIdMarker`. Could we make `max-keys` count each `Version`/`DeleteMarker`, wire through `version-id-marker` / `NextVersionIdMarker`, and add a regression test? Once that is in, I think we are good.

hectorvent:
Thank you for this, and sorry it sat waiting on us. The only thing outstanding was a check on our side, which I have now done, and it passes.

Against the model, `encoding-type=url` covers exactly the right members: `Delimiter`, `Prefix`, `Key` and `StartAfter` for V2, and `KeyMarker`, `NextKeyMarker`, `Prefix`, `Key`, `Delimiter` for ListObjectVersions ([s3 service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/s3/2006-03-01/service-2.json)). Leaving `VersionIdMarker` unencoded is right too, and easy to get wrong.

The encoder is the part I liked. `URLEncoder` is form encoding, so a space becomes `+`, and the correction to `%20` runs after a literal `+` has already become `%2B`. That ordering is what makes both cases work.

1. **(follow-up, separate PR)** The model describes `NextKeyMarker` as the first key **not** returned, while this sets it to the last entry returned. Paginators round trip either way, so nothing breaks, but worth a real AWS capture sometime.

No blockers from my side.

hectorvent (approved):

PR #1738, opened 2026-07-05
feat(firehose): add proper ResourceInUseException and ResourceNotFoundException error handling

### LINKED ISSUE #1674: [FEAT] Error handling correctly for Firehose
## Service

Firehose

## API Action / Feature

[CreateDeliveryStream](https://docs.aws.amazon.com/firehose/latest/APIReference/API_CreateDeliveryStream.html), etc

## AWS Documentation

https://docs.aws.amazon.com/firehose/latest/APIReference/API_CreateDeliveryStream.html#API_CreateDeliveryStream_Errors

## Why is this needed?

Firehose returns a `ResourceInUseException` when delivery streams are registered with duplicate key names, but in the [current implementation](https://github.com/floci-io/floci/blob/1.5.29/src/main/java/io/github/hectorvent/floci/services/firehose/FirehoseService.java#L52), the registration succeeds without throwing any exception and the stream is overwritten. we would like to align the behavior with the actual one.

- [ ] Yes
- [x] No

iam-ssrivastav:
## Summary

Closes #1674
## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/firehose/FirehoseService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/firehose/FirehoseService.java
@@ -50,6 +50,16 @@ public class FirehoseService {
 
     public String createDeliveryStream(String name, S3Destination s3Config, List<DeliveryStreamDescription.Tag> tags,
                                        String deliveryStreamType) {
+        if (name == null || name.isEmpty() || name.length() > 64 || !name.matches("[a-zA-Z0-9_.-]+")) {
+            throw new AwsException("InvalidArgumentException",
+                    "Delivery stream name must be between 1 and 64 characters and contain only letters, numbers, underscores, hyphens, or periods.", 400);
+        }
+
+        if (streamStore.get(name).isPresent()) {
+            throw new AwsException("ResourceInUseException",
+                    "Delivery stream " + name + " already exists.", 409);
+        }
+
         validateBufferingHints(s3Config);
         String arn = AwsArnUtils.Arn.of("firehose", regionResolver.getDefaultRegion(), regionResolver.getAccountId(), "deliverystream/" + name).toString();
         DeliveryStreamDescription description = new DeliveryStreamDescription(name, arn, s3Config);
--- a/src/main/java/io/github/hectorvent/floci/services/firehose/FirehoseJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/firehose/FirehoseJsonHandler.java
@@ -26,10 +26,22 @@ public class FirehoseJsonHandler {
         this.mapper = mapper;
     }
 
+    private String getDeliveryStreamName(JsonNode request) {
+        if (request == null || !request.has("DeliveryStreamName") || request.get("DeliveryStreamName").isNull()) {
+            throw new AwsException("InvalidArgumentException", "Delivery stream name must not be null or empty.", 400);
+        }
+        String name = request.get("DeliveryStreamName").asText();
+        if (name.isEmpty() || name.length() > 64 || !name.matches("[a-zA-Z0-9_.-]+")) {
+            throw new AwsException("InvalidArgumentException",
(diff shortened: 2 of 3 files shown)


hectorvent:
Thanks for this! The core fix is right and I verified it against the model: `ResourceInUseException` and `InvalidArgumentException` are both declared errors for `CreateDeliveryStream` in [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/firehose/2015-08-04/service-2.json), and the 400 status plus `__type`/`message` wire shape all match. I ran the suite locally too, 10/10 green.

Three small things before merge:

1. The title mentions `ResourceNotFoundException` handling, but that part already existed. `deleteDeliveryStream`, `putRecord`, `tagDeliveryStream` and `listTagsForDeliveryStream` all delegate to `describeDeliveryStream()`, which already throws it (`FirehoseService.java:188`). I confirmed the four new not found tests still pass with `FirehoseService.java` reverted to `main`. They are useful regression coverage, so please keep them, but could you retitle to something like `fix(firehose): return ResourceInUseException on duplicate CreateDeliveryStream`?

2. Name validation is partial. The model constrains `DeliveryStreamName` to `pattern=[a-zA-Z0-9_.-]+` and `len=[1..64]`, so these still slip through today:

```jsonc
{"DeliveryStreamName": "my stream"}   // space, invalid on AWS
{"DeliveryStreamName": "<65+ chars>"} // too long
```

Since you are already adding `InvalidArgumentException` here, it is a cheap addition, though a follow up is fine too.

3. Could you fill in the PR template (type of change and the checklist)? It is currently empty.

Nice, focused fix otherwise. Thanks again!

src/main/java/io/github/hectorvent/floci/services/firehose/FirehoseService.java
```diff
@@ -50,6 +50,16 @@ public String createDeliveryStream(String name, S3Destination s3Config, List<Del
 
     public String createDeliveryStream(String name, S3Destination s3Config, List<DeliveryStreamDescription.Tag> tags,
                                        String deliveryStreamType) {
+        if (name == null || name.isBlank()) {
```
abanna: @iam-ssrivastav thanks for picking up the validation work here. This guard is helpful, but an omitted `DeliveryStreamName` never reaches it: the handler calls `request.get("DeliveryStreamName").asText()` first and returns a 500. Could we update that path too so `{}` gets this documented 400 `InvalidArgumentException`? A `{}` test would nail it.

hectorvent:
@iam-ssrivastav this file should be reverted `.github/pull_request_template.md`. This is just the template to be used a reference when creating the PR

iam-ssrivastav:
done

Mide69:
Nice work on this. The duplicate-name check and name validation look good, and I like how the overload chain (createDeliveryStream(name, s3) → ...(name, s3, tags) → ...(name, s3, tags, type)) funnels everything through the same validation logic. The test coverage for invalid and duplicate names is solid too.

Just a couple of minor things I noticed:

For UpdateDestination, CurrentDeliveryStreamVersionId and DestinationId now default to an empty string if they're missing instead of throwing an error. That avoids the old NPE, but since AWS requires both fields, would it make more sense to throw an InvalidArgumentException here as well, similar to how getDeliveryStreamName is handled, rather than allowing "" through?

Also, the name regex and length validation are duplicated in both FirehoseJsonHandler and FirehoseService. It's not a big issue, but it might be worth pulling that into a shared helper if we end up needing the same validation elsewhere.

Nothing blocking from my side. Happy to help test after you've had a chance to look at the first point.

hectorvent:
@iam-ssrivastav thanks for the PR. For future PR it is required to have context what is the issue and what is being resolved by the PR. This will give a quick summary for reviewers before jumping into the code. This is also helpful for the IA review bot. 

Thanks!

PR #1740, opened 2026-07-05
fix(rds): fault with DBInstanceNotFound/DBClusterNotFoundFault for missing identifiers in Describe*

Ollstar:
## Problem

`DescribeDBInstances` / `DescribeDBClusters` return **HTTP 200 with an empty list** when the identifier *parameter* matches nothing. Real AWS **faults** in that case, and only returns empty lists for the `Filters` form:

```console
# real AWS (verified live):
$ aws rds describe-db-instances --db-instance-identifier does-not-exist
An error occurred (DBInstanceNotFound) when calling the DescribeDBInstances
operation: DBInstance does-not-exist not found.

# floci 1.5.30:
$ aws --endpoint-url http://localhost:4566 rds describe-db-instances \
    --db-instance-identifier does-not-exist
{ "DBInstances": [] }        # exit 0
```

### Why it bites

Provision-then-create scripts use the standard idiom `describe || create`. Against floci the describe "succeeds", the script concludes the resource exists, skips creation, and then waits forever on a resource that was never created. We hit exactly this bringing a docker-compose stack up from zero state.

### Root cause

`RdsQueryHandler.handleDescribeDbInstances` merges the `DBInstanceIdentifier` **parameter** and the `db-instance-id` **filter** into a single filter path, losing the semantic difference — `RdsService.listDbInstances` is a pure filter with no fault path:

https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java#L132-L148

Notably, `ElbV2Service.describeLoadBalancers` already implements the correct pattern (filter, then throw `LoadBalancerNotFound` on empty-with-criteria), so this brings RDS in line with the codebase's own precedent — and with the AWS-parity direction of fixes like #1431.

## Fix

In both RDS handlers: track whether the identifier came from the *parameter*; after listing, fault when the parameter form matched nothing. The `Filters` form keeps returning empty lists (matching AWS). Error codes, messages, and 404 status match live AWS (`DBInstanceNotFound` / `DBClusterNotFoundFault`, `"DBInstance <id> not found."`).

- No change for describes of existing resources.
- No change for the unfiltered list form.
- No change for the `Filters` form.

## Follow-ups (addressed in this PR)

Both raised in review; folded in here rather than deferred.

**1. Identifier parameter and id filters now also accept an ARN (RDS, DocDB, Neptune).** Per the AWS model `DBInstanceIdentifier` / `DBClusterIdentifier` and the `db-instance-id` / `db-cluster-id` filters accept an ARN, but the list methods matched on the identifier key only — so an ARN for an instance that *exists* would now (post-fix) fault `DBInstanceNotFound`. All three services' list methods match an ARN against each resource's **stored** ARN (rather than reducing it to the bare identifier), so a cross-account or cross-region ARN does not resolve a same-named local resource.

**2. DocDB / Neptune now honour the `Filters` vs parameter split too.** `DocDbQueryHandler` and `NeptuneQueryHandler` guarded on the *merged* id, so they faulted on the `Filters` form as well — the mirror image of the RDS bug. They now fault only for the identifier **parameter**; the `db-cluster-id` / `db-instance-id` `Filters` form returns an empty list. (DocDB/Neptune already faulted correctly on the parameter form — this only fixes the over-eager `Filters`-form fault.)

## Bonus: deletion waiters

[waiters-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/rds/2014-10-31/waiters-2.json) treats `DBInstanceNotFound` as success for `DBInstanceDeleted` and `DBClusterNotFoundFault` as success for `DBClusterDeleted`, so `aws rds wait db-instance-deleted` never completed against floci before this fix.

## Tests

- `RdsQueryHandlerTest`: parameter-form faults (404 + error code + message) and `Filters`-form still returns empty 200, for both instances and clusters.
- `RdsServiceTest` / `DocDbServiceTest` / `NeptuneServiceTest`: `listDbInstances` / `listDbClusters` match by ARN; cross-account and cross-region ARNs do **not** match.
- `DocDbIntegrationTest` / `NeptuneIntegrationTest`: `Filters` form returns an empty list (200, no fault) for a missing id.

`./mvnw test -Dtest=RdsServiceTest,RdsQueryHandlerTest,DocDbServiceTest,NeptuneServiceTest,DocDbIntegrationTest,NeptuneIntegrationTest` green locally (JDK 25), rebased on current `main`.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/docdb/DocDbQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/docdb/DocDbQueryHandler.java
@@ -75,14 +75,17 @@ public class DocDbQueryHandler {
     }
 
     private Response handleDescribeDbClusters(MultivaluedMap<String, String> params) {
-        String filterId = params.getFirst("DBClusterIdentifier");
+        String identifier = params.getFirst("DBClusterIdentifier");
+        String filterId = identifier;
         if (filterId == null || filterId.isBlank()) {
             filterId = extractFilterValue(params, "db-cluster-id");
         }
 
-        // When a specific cluster ID is requested, AWS returns 404 if not found
-        if (filterId != null && !filterId.isBlank()) {
-            service.getDbCluster(filterId); // throws DBClusterNotFoundFault if absent
+        // AWS parity: the DBClusterIdentifier parameter faults with
+        // DBClusterNotFoundFault when no cluster matches, while the
+        // db-cluster-id Filters form returns an empty list.
+        if (identifier != null && !identifier.isBlank()) {
+            service.getDbCluster(identifier); // throws DBClusterNotFoundFault if absent
         }
 
         Collection<DocDbCluster> result = service.listDbClusters(filterId);
@@ -154,13 +157,17 @@ public class DocDbQueryHandler {
     }
 
     private Response handleDescribeDbInstances(MultivaluedMap<String, String> params) {
-        String filterId = params.getFirst("DBInstanceIdentifier");
+        String identifier = params.getFirst("DBInstanceIdentifier");
+        String filterId = identifier;
         if (filterId == null || filterId.isBlank()) {
             filterId = extractFilterValue(params, "db-instance-id");
         }
 
-        if (filterId != null && !filterId.isBlank()) {
-            service.getDbInstance(filterId); // throws DBInstanceNotFound if absent
+        // AWS parity: the DBInstanceIdentifier parameter faults with
+        // DBInstanceNotFound when no instance matches, while the
+        // db-instance-id Filters form returns an empty list.
+        if (identifier != null && !identifier.isBlank()) {
+            service.getDbInstance(identifier); // throws DBInstanceNotFound if absent
         }
 
         Collection<DocDbInstance> result = service.listDbInstances(filterId);
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
@@ -502,6 +502,15 @@ public class RdsService implements Resettable {
 
     public Collection<DbInstance> listDbInstances(String filterId) {
         if (filterId != null && !filterId.isBlank()) {
+            // DBInstanceIdentifier also accepts an ARN per the AWS model. Match the
+            // full ARN against each instance's stored ARN rather than reducing it to
+            // the bare identifier, so a cross-account or cross-region ARN does not
+            // resolve a same-named local instance.
+            if (filterId.startsWith("arn:")) {
+                return instances.scan(k -> true).stream()
+                        .filter(i -> filterId.equalsIgnoreCase(i.getDbInstanceArn()))
+                        .toList();
+            }
             return instances.scan(k -> k.equalsIgnoreCase(filterId));
         }
         return instances.scan(k -> true);
@@ -726,6 +735,15 @@ public class RdsService implements Resettable {
 
     public Collection<DbCluster> listDbClusters(String filterId) {
         if (filterId != null && !filterId.isBlank()) {
+            // DBClusterIdentifier also accepts an ARN per the AWS model. Match the
+            // full ARN against each cluster's stored ARN rather than reducing it to
+            // the bare identifier, so a cross-account or cross-region ARN does not
+            // resolve a same-named local cluster.
+            if (filterId.startsWith("arn:")) {
+                return clusters.scan(k -> true).stream()
+                        .filter(c -> filterId.equalsIgnoreCase(c.getDbClusterArn()))
+                        .toList();
+            }
             return clusters.scan(k -> k.equalsIgnoreCase(filterId));
         }
         return clusters.scan(k -> true);
--- a/compatibility-tests/sdk-test-go/tests/rds_test.go
+++ b/compatibility-tests/sdk-test-go/tests/rds_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/rds"
+	"github.com/aws/aws-sdk-go-v2/service/rds/types"
 	_ "github.com/lib/pq"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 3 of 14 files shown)


Ollstar:
Addressed in ac5740cd — moved the new test block above the `Helpers` separator so all `@Test` methods precede the utilities. `RdsQueryHandlerTest` still 41/41 green.

hectorvent:
Nice fix, and the subtle part is exactly right.

Verified against botocore's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/rds/2014-10-31/service-2.json): the two faults do not serialize symmetrically, and this PR gets both right.

```json
"DBInstanceNotFoundFault": { "error": { "code": "DBInstanceNotFound",     "httpStatusCode": 404, "senderFault": true } }
"DBClusterNotFoundFault":  { "error": { "code": "DBClusterNotFoundFault", "httpStatusCode": 404, "senderFault": true } }
```

Each is the only declared error on its operation, both are 404, and the messages match [moto](https://github.com/getmoto/moto/blob/master/moto/rds/exceptions.py#L8-L12) exactly (`DBInstance <id> not found.` / `DBCluster <id> not found.`). The parameter vs `Filters` split matches AWS too.

One thing worth adding to the description: this also unblocks the deletion waiters. [waiters-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/rds/2014-10-31/waiters-2.json) treats `DBInstanceNotFound` as success for `DBInstanceDeleted`, and `DBClusterNotFoundFault` as success for `DBClusterDeleted`, so `aws rds wait db-instance-deleted` currently never completes against floci. Nice bonus.

Two follow ups, neither blocking:

1. `DBInstanceIdentifier` also accepts an ARN per the model, but [`RdsService.listDbInstances`](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java#L455) matches on the identifier key only (same for [`listDbClusters`](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java#L661)), so an ARN now returns `DBInstanceNotFound` for an instance that does exist. Pre existing, just newly reachable.

    ```java
    return instances.scan(k -> k.equalsIgnoreCase(filterId));  // never matches an ARN
    ```

2. DocDB and Neptune guard on the merged id, so they fault on the `Filters` form as well: [`DocDbQueryHandler.java:85`](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/docdb/DocDbQueryHandler.java#L85) and [`:163`](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/docdb/DocDbQueryHandler.java#L163), [`NeptuneQueryHandler.java:79`](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneQueryHandler.java#L79) and [`:149`](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneQueryHandler.java#L149). Mirror image of what this PR fixes for RDS.

Also, the `docdb describe-db-clusters` line in the description reads like floci's DocDB is broken the same way, but DocDB already faults here, so a reviewer may go looking for a DocDB change that is not in the diff. Might be worth a quick rephrase.

LGTM 👍

src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
```diff
@@ -399,7 +410,10 @@ public DbInstance getDbInstance(String id) {
 
     public Collection<DbInstance> listDbInstances(String filterId) {
         if (filterId != null && !filterId.isBlank()) {
-            return instances.scan(k -> k.equalsIgnoreCase(filterId));
+            // DBInstanceIdentifier also accepts an ARN per the AWS model, so match
+            // on the bare identifier extracted from a db ARN as well as a plain id.
+            String id = dbInstanceIdentifierFromResourceName(filterId);
```
abanna: @Ollstar thanks for adding ARN support to this filter. Could we validate the account and Region before reducing an ARN to its bare ID? A foreign ARN currently resolves a same-named local DB, which makes the identity ambiguous. A small cross-account or cross-Region case would keep this safe.

hectorvent:
@Ollstar thanks for addressing the comments. Two more thing before merging the PR: 
1. remove the Co-Author, you will need to for a push to do it. 
2. rebase main to resolve conflicts

hectorvent:
Thank you for working through the earlier comments; the remaining items are all quick.

**(blocking)** Commits 4285e84, ac5740c, 9fce126, 490c615 carry an AI co-author trailer; this repo keeps commit attribution with human contributors (see AGENTS.md). Would you amend or squash those commits and force push so the attribution is yours alone?

**(blocking)** A rebase on main is needed to clear the conflicts, and one scope question first: the issue this targeted, #1728, was closed by #1870, so part of what this PR fixes may already be on main. Could you check which `Describe*` paths still return the wrong fault today and narrow the PR to those? That keeps the diff small and makes it obvious what is still needed.

Both blockers are mechanical; happy to help if the rebase turns out messy.

src/main/java/io/github/hectorvent/floci/services/docdb/DocDbQueryHandler.java
```diff
@@ -74,14 +74,17 @@ private Response handleCreateDbCluster(MultivaluedMap<String, String> params) {
...
-        if (filterId != null && !filterId.isBlank()) {
-            service.getDbCluster(filterId); // throws DBClusterNotFoundFault if absent
+        // AWS parity: the DBClusterIdentifier parameter faults with
+        // DBClusterNotFoundFault when no cluster matches, while the
+        // db-cluster-id Filters form returns an empty list.
+        if (identifier != null && !identifier.isBlank()) {
+            service.getDbCluster(identifier); // throws DBClusterNotFoundFault if absent
         }
 
         Collection<DocDbCluster> result = service.listDbClusters(filterId);
```
abanna, line 91: Thanks for splitting the parameter and Filters forms apart — the RDS half reads exactly right, and `listDbInstancesDoesNotMatchForeignArn` is a nice answer to the cross-account note from earlier.

One gap on the DocDB/Neptune half. The `db-cluster-id` filter accepts cluster ARNs as well as bare identifiers (same for `db-instance-id`), but `DocDbService.listDbClusters` and `NeptuneService.listDbClusters` still match with `k.equalsIgnoreCase(filterId)` on the bare id only. This PR adds precisely that `arn:` branch to `RdsService`, so after this change:

```
--filters Name=db-cluster-id,Values=arn:aws:rds:us-east-1:123456789012:cluster:mycluster
```

returns the cluster on RDS but an empty list on DocDB and Neptune. Before this PR that path at least faulted; now it silently returns nothing, which is a good deal harder to notice from a caller.

Could the same `arn:` branch move into `DocDbService`/`NeptuneService`, with a foreign-ARN case mirroring the new RDS tests?

Unrelated to the code: GitHub currently reports the branch as conflicting with `main`, so it will need a rebase before it can go in.

hectorvent:
@Ollstar Please take a look a the actions failures.

hectorvent (approved):
Thanks @Ollstar

PR #1618, opened 2026-06-27
 fix(elasticache): roll back proxy port and container on failed provisioning

ahmedk20:
## Summary

`ElastiCacheService.createReplicationGroup` reserved a proxy port via
`allocateProxyPort()` and then started the container and auth proxy with **no
rollback**. If `containerManager.start(...)` or `proxyManager.startProxy(...)`
threw, the port stayed in `usedPorts` (it is only released on the delete path)
and any started container was orphaned. Repeated failures slowly exhausted the
proxy-port range and surfaced as a bogus `InsufficientReplicationGroupCapacity`
(503) even with no groups present.

This wraps startup in `try/catch`: on failure it stops the proxy, stops the
container if it was started, **always** releases the proxy port in `finally`,
then rethrows so the caller still receives the original error. This mirrors the
already-merged `MemoryDbService.startBackend` / `rollbackBackend` reference
implementation.

Closes #1605

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore
## AWS Compatibility

Bug fix — no wire-protocol change. The incorrect behavior was an internal
resource leak: a failed `CreateReplicationGroup` leaked the reserved proxy port
and orphaned its container, eventually producing a spurious
`InsufficientReplicationGroupCapacity` on later (otherwise valid) creates.

## Checklist

- [x] `./mvnw test` passes locally (`ElastiCacheServiceTest`: 4 run, 0 failures)
- [x] New or updated test added — regression test
      `failedProvisioningRollsBackContainerAndReleasesProxyPort` proves the port
      is released (next create reuses the base port) and the container + proxy
      are stopped
- [x] Commit messages follow Conventional Commits

|

PR #1742, opened 2026-07-05
fix(neptune): roll back proxy port and container on failed provisioning

### LINKED ISSUE #1741: [BUG] Neptune CreateDBCluster leaks the proxy port and orphans the container when provisioning failsNeptune CreateDBCluster leaks the proxy port and orphans the container when provisioning fails
## Service

Neptune
## AWS API Action

CreateDBCluster (failure path / resource cleanup)
## Expected behavior

When CreateDBCluster fails partway through provisioning, the emulator should clean up any resources it already reserved for that attempt — release the proxy port back to the pool and stop any container it started — so a failed create leaves no trace. Real AWS never leaks capacity from a failed create; a subsequent CreateDBCluster succeeds normally.
## Actual behavior

NeptuneService.createDbCluster reserves a proxy port (allocateProxyPort()) and then starts the container and proxy with no rollback. Any exception thrown after the port is reserved — a container start failure or a proxy bind failure — leaves the reserved port stuck in usedPorts (it's only ever released on the delete path) and orphans any container that already started.

Because the port is never returned, repeated failed creates exhaust the configured proxyBasePort..proxyMaxPort range. Once the range is used up, further creates fail with:

InsufficientNeptuneCapacity (503): No available proxy ports in range 18182-18199

…even though no clusters actually exist. Orphaned Docker containers accumulate across the failed attempts as well.

This is the same class of bug fixed for ElastiCache in #1618 and MemoryDB earlier — Neptune has the identical unguarded pattern.
## Reproduction

Trigger provisioning failures (e.g. Docker unavailable, an unreachable/invalid image, or the proxy port already bound by another process), then repeat until the port range is exhausted:

// With the Neptune backend unable to provision (e.g. Docker socket down,
// or neptune.default-image pointing at a non-pullable tag), repeatedly:
for i in $(seq 1 25); do
  aws --endpoint-url http://localhost:4566 neptune create-db-cluster \
    --db-cluster-identifier "leak-$i" --engine neptune || true
done

// Each attempt fails, but the reserved proxy port is never released.
// After ~18 attempts (range 18182-18199), even a create that *would*
// otherwise succeed fails with:
//   InsufficientNeptuneCapacity: No available proxy ports in range 18182-18199
// and `docker ps -a` shows orphaned Neptune containers from the failed creates.

## Environment

- Floci version / image tag: main (reproduced against upstream/main @ 1.5.29)
- Java SDK version (if applicable): AWS SDK for Java v2 (software.amazon.awssdk:neptune); also reproducible via AWS CLI
- How you're running Floci (Docker / native / mvn quarkus:dev): mvn quarkus:dev with a Docker daemon available (real-mode Neptune provisioning)

ahmedk20:
createDbCluster reserved a proxy port and started the container/proxy with no rollback: any throw after allocateProxyPort() leaked the port (only freed on delete) and orphaned the container. Repeated failures could exhaust the proxy-port range and surface as a spurious InsufficientNeptuneCapacity error.

Wrap startup in try/catch, add rollbackDbCluster() that stops the proxy and container in isolated try blocks (guarded by handle != null) and always releases the port in finally, then rethrow. Mirrors the merged ElastiCache fix (#1618). Adds unit tests for both rollback branches (proxy-start throws after container is up; container-start throws before any handle).

## Summary

NeptuneService.createDbCluster reserved a proxy port and started the container/proxy with no rollback: any throw after allocateProxyPort() leaked the port (only freed on the delete path) and orphaned the container. Repeated failures could exhaust the proxy-port range and surface as a spurious InsufficientNeptuneCapacity (503) even with no clusters present.

This wraps startup in try/catch and adds rollbackDbCluster() that stops the proxy and container in isolated try blocks (each guarded by handle != null) and always releases the port in finally, then rethrows the original exception. A failed create now persists nothing, so there's no orphaned storage entry to reconcile.

closes #1741

Mirrors the already-merged ElastiCache fix (#1618) — Neptune had the identical pattern.

## Type of change

- [x] Bug fix (fix:)
- [ ] New feature (feat:)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior fixed: on a mid-provision failure (container start or proxy bind throwing after the port was reserved), the reserved proxy port stayed in usedPorts and any started container was orphaned. Enough failed creates would exhaust the configured proxyBasePort..proxyMaxPort range and return a bogus InsufficientNeptuneCapacity error.
## Checklist

- [ ] New or updated integration test added — not applicable: this is failure-path hardening (forcing a mid-provision throw), covered by Mockito unit tests in NeptuneServiceTest for both rollback branches, same approach as the reference fix #1618. The existing NeptuneIntegrationTest still covers the happy path.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java
@@ -78,6 +78,7 @@ public class NeptuneService {
                 id, String.valueOf(proxyPort), dbType, image);
 
         NeptuneContainerHandle handle = null;
+        boolean provisioned = false;
         try {
             handle = containerManager.start(id, image, dbType);
 
@@ -105,13 +106,20 @@ public class NeptuneService {
             proxyManager.startProxy(id, proxyPort, handle.getHost(), handle.getPort());
 
             clusters.put(id, cluster);
+            provisioned = true;
             LOG.infov("Neptune cluster {0} created ({1}), endpoint={2}:{3}",
                     id, dbType, endpointHost, String.valueOf(proxyPort));
             return cluster;
         } catch (RuntimeException e) {
             LOG.warnv("Neptune cluster {0} provisioning failed, rolling back: {1}", id, e.getMessage());
-            rollbackDbCluster(id, handle, proxyPort);
             throw e;
+        } finally {
+            // Roll back on ANY non-success exit — including a JVM Error, which a
+            // catch (RuntimeException) would miss — so a failed create never leaks the
+            // reserved port or leaves a container behind. Idempotent and a no-op on success.
+            if (!provisioned) {
+                rollbackDbCluster(id, handle, proxyPort);
+            }
         }
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java
@@ -117,6 +117,8 @@ public class NeptuneService {
 
     private void rollbackDbCluster(String id, NeptuneContainerHandle handle, int proxyPort) {
         try {
+            // The proxy only starts after the container is ready, so a null handle means it
+            // never started — nothing to stop.
             if (handle != null) {
                 proxyManager.stopProxy(id);
             }
@@ -124,9 +126,11 @@ public class NeptuneService {
             LOG.warnv("Error stopping proxy for Neptune cluster {0}: {1}", id, e.getMessage());
         }
         try {
-            if (handle != null) {
-                containerManager.stop(handle);
-            }
+            // Stop by id, not handle: a readiness timeout in containerManager.start() throws
+            // after the container was created and registered but before the handle is returned,
+            // so cleaning up by handle here would miss (and orphan) it. stopByClusterId is
+            // idempotent, so it's safe when the container never started.
+            containerManager.stopByClusterId(id);
         } catch (RuntimeException e) {
             LOG.warnv("Error stopping container for Neptune cluster {0}: {1}", id, e.getMessage());
         } finally {
--- a/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java
@@ -77,35 +77,61 @@ public class NeptuneService {
         LOG.infov("Creating Neptune cluster {0} on proxy port {1}, dbType={2}, image={3}",
                 id, String.valueOf(proxyPort), dbType, image);
 
-        NeptuneContainerHandle handle = containerManager.start(id, image, dbType);
+        NeptuneContainerHandle handle = null;
+        try {
+            handle = containerManager.start(id, image, dbType);
 
-        String region = regionResolver.getDefaultRegion();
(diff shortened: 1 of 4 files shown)


ahmedk20:
@hectorvent

hectorvent:
@ahmedk20 please resolve conflicts. Thanks

hectorvent:
This is a careful fix and the comments hold up. I checked all three claims they rest on. `start()` does `activeContainers.put` at line 91 and `waitForBackendReady` at 104, returning the handle only at 106, so a readiness timeout really does leave a registered container with no handle at the call site, which is exactly why cleaning up by id instead of by handle is the right move. `stop()` removes from `activeContainers` before `stopAndRemove` and `stopProxy` no-ops on an unknown id, so both steps are genuinely idempotent. And rolling back in `finally` on `!provisioned` rather than in `catch (RuntimeException)` does cover the `Error` case, with `clusters.put` sitting immediately before `provisioned = true` so a failed create persists nothing. Tests green, 4/4, and they hit both branches.

One small thing in the same class. [`allocateProxyPort()`](https://github.com/floci-io/floci/blob/998c66450ec31f4e61d6731ac8848cad51dce304/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java#L65) runs at line 65 but the `try` doesn't open until [line 80](https://github.com/floci-io/floci/blob/998c66450ec31f4e61d6731ac8848cad51dce304/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java#L80). The config reads in between are unlikely to throw, but if one ever did the port would leak in exactly the way this PR prevents. Opening the try right after the allocation closes it for free:

```java
int proxyPort = allocateProxyPort();
NeptuneContainerHandle handle = null;
boolean provisioned = false;
try {
    NeptuneDbType dbType = ...;   // config reads move inside
    String image = ...;
    handle = containerManager.start(id, image, dbType);
    ...
} finally {
    if (!provisioned) {
        rollbackDbCluster(id, handle, proxyPort);
    }
}
```

Separately, and definitely not for this PR: the summary names `InsufficientNeptuneCapacity` as the symptom, and that code doesn't exist in AWS. Neptune's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/neptune/2014-10-31/service-2.json) has `InsufficientDBClusterCapacityFault` (403), `InsufficientDBInstanceCapacityFault` and `InsufficientStorageClusterCapacityFault` (wire code `InsufficientStorageClusterCapacity`, 400), and `CreateDBCluster` declares only the last of those. So [`NeptuneService.java:310`](https://github.com/floci-io/floci/blob/998c66450ec31f4e61d6731ac8848cad51dce304/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java#L310) throws something the SDK can't map, callers just get a generic `NeptuneException`. Your fix makes it rarer rather than impossible, so it might be worth a follow up issue.

ahmedk20:
Thanks for the careful review,
Port reserved outside the try fixed in ddabd562. Moved the try to open right after allocateProxyPort() so a throw from the config reads can't leak the port.
InsufficientNeptuneCapacity — agreed, that code isn't in Neptune's service-2.json so the SDK can't map it. It's orthogonal to the rollback, so I'll keep this PR focused and open a follow-up.
Also merged latest main to clear the conflict.
@hectorvent

hectorvent (approved):

PR #1743, opened 2026-07-05
fix(elbv2): expose target health reason parity

jvanzyl:
## Summary
- return AWS-shaped TargetHealth reason and description fields for registered, initial, timeout, failed, response-code-mismatch, and explicitly unregistered targets
- trigger a prompt health probe after target registration so DescribeTargetHealth can progress without waiting for the first periodic interval
- cover ASG target-group registration with reconciler coverage and SDK enum assertions for DescribeTargetHealth

## Verification
- ./mvnw -Dtest=ElbV2HealthCheckerTest,ElbV2ServiceTest,AutoScalingReconcilerTest test
- ./mvnw -Dtest=RdsCfnProvisionerTest,ElbV2HealthCheckerTest,ElbV2ServiceTest,AutoScalingReconcilerTest,ElbV2IntegrationTest,AutoScalingIntegrationTest,Ec2IntegrationTest,RdsServiceTest,RdsQueryHandlerTest test
- ./mvnw -f compatibility-tests/sdk-test-java/pom.xml -DskipTests test-compile
- hojo branch-family verify --dir . --format=text
- git diff --check

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/ElbV2Test.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/ElbV2Test.java
@@ -70,6 +70,7 @@ import software.amazon.awssdk.services.elasticloadbalancingv2.model.TargetDescri
 import software.amazon.awssdk.services.elasticloadbalancingv2.model.TargetGroup;
 import software.amazon.awssdk.services.elasticloadbalancingv2.model.TargetGroupAttribute;
 import software.amazon.awssdk.services.elasticloadbalancingv2.model.TargetHealthDescription;
+import software.amazon.awssdk.services.elasticloadbalancingv2.model.TargetHealthReasonEnum;
 import software.amazon.awssdk.services.elasticloadbalancingv2.model.TargetHealthStateEnum;
 import software.amazon.awssdk.services.elasticloadbalancingv2.model.TargetTypeEnum;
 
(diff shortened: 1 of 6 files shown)


hectorvent:
Thanks for this! The target health work is solid. I checked every reason code against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/elbv2/2015-12-01/service-2.json) and the state to reason pairings all match AWS exactly (`initial` with `Elb.RegistrationInProgress`/`Elb.InitialHealthChecking`, `unhealthy` with `ResponseCodeMismatch`/`Timeout`/`FailedHealthChecks`, `unused` with `Target.NotRegistered`, and no reason on `healthy`). That half looks ready to go.

Two asks before merge.

**Could you split out the same port host dispatch?** The PR summary only mentions target health, but it also carries `ElbV2DataPlane` host based listener dispatch, its integration test, and the `HttpProxyInvoker`/`ApiGatewayExecuteController` changes, so roughly 530 lines of a separate feature. The target health change is clean enough to merge on its own, and the dispatch work deserves its own review.

**The raw socket client in `HttpProxyInvoker` has a few issues.** I know it exists to work around `HttpClient` refusing to set `Host`, but as written:

1. Duplicate response headers are lost. The raw path does `headers.put(name, value)` into a `Map<String, String>`, so a backend sending two `Set-Cookie` headers keeps only the last, while the existing `HttpClient` path in the same class preserves them via `String.join(",", e.getValue())`.
2. Header values are written straight to the wire with no validation, so a value containing `\r\n` injects headers. `HttpRequest.Builder.header()` validates and rejects these, so this is less safe than the call it replaces.
3. The override is gated on `startsWith("http://")`, so for an `https://` backend the `Host` header is silently dropped by `RESTRICTED` and the feature quietly does not apply.

The good news is there is a one line alternative. Java allows unlocking the restricted headers:

```
-Djdk.httpclient.allowRestrictedHeaders=host
```

With that set, `HttpRequest.Builder.header("Host", ...)` just works over both HTTP and HTTPS, you keep HttpClient's header validation and multi value handling, and the whole socket path can go away.

One last small thing, `Target.HealthCheckDisabled` (state `unavailable`) is still unmodeled, so a target in a group with health checks disabled stays `initial`/`Elb.RegistrationInProgress` forever. Happy to leave that as a follow up.

Nice work on the health reasons, thanks again!

jvanzyl:
Thanks for the review. I split the same-port host dispatch and raw-socket HTTP work out of this PR. The current diff is now limited to ELBv2 target-health reason/description parity and its focused unit, integration, and SDK coverage.

The outdated scheduler review thread has also been resolved because that code is no longer present here. `Target.HealthCheckDisabled` remains intentionally out of scope for this PR as the suggested follow-up.

hectorvent (approved):
Thanks @jvanzyl

PR #1750, opened 2026-07-06
feat(ses): implement identity (sending authorization) policies for v1 and v2

okinaka:
## Summary

Adds the SES identity (sending authorization) policy APIs across **both** protocols, backed by one shared `StorageFactory` store.

**v1 (Query/XML)**
| Action | Behaviour |
| --- | --- |
| `PutIdentityPolicy` | upsert (create or overwrite) |
| `GetIdentityPolicies` | requested names only; missing omitted |
| `ListIdentityPolicies` | policy names |
| `DeleteIdentityPolicy` | idempotent (no error if absent) |

**v2 (REST JSON, `/v2/email/identities/{id}/policies/{name}`)**
| Method | Operation | Behaviour |
| --- | --- | --- |
| `POST` | `CreateEmailIdentityPolicy` | `AlreadyExistsException` on duplicate |
| `GET` | `GetEmailIdentityPolicies` | all policies as a `{name: doc}` map |
| `PUT` | `UpdateEmailIdentityPolicy` | `NotFoundException` when missing |
| `DELETE` | `DeleteEmailIdentityPolicy` | `NotFoundException` when missing |

Policies are shared between the two protocols (a v1-written policy is returned by the v2 Get, and vice versa), and deleting an identity drops its policies.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Behaviour and every error string were verified against real AWS (SES v1 + v2, boto3 / raw signed requests, us-east-1):

- Per-identity limit of **20** (`LimitExceededException` on v2, `InvalidParameterValue` on v1).
- v1/v2 strictness asymmetry: v2 requires the identity to exist and errors on duplicate-create / missing-update / missing-delete; v1 `Put` upserts and v1 `List`/`Get`/`Delete` are lenient (no identity check, idempotent delete).
- `PolicyName` rules (`[A-Za-z0-9_-]`, at most 64 characters). The wire error **codes differ by protocol** — invalid chars: `InvalidParameterValue` (v1) / `BadRequestException` (v2); over length: `ValidationError` (v1) / `BadRequestException` (v2) — with identical messages; the v2 controller remaps the v1 codes.
- `Policy` is a required string: a missing/null value returns the Smithy `Member must not be null` validation error, and a non-string value is rejected (not coerced).
- The policy document is stored with insignificant whitespace normalized, as AWS returns it.

**Deviations (documented in the SES docs):** Floci stores and returns policies as metadata but does not enforce the authorization — it does not validate that the `Principal` account exists or that the policy `Resource` matches the identity, and it does not gate `SendEmail` on policies. This matches Floci's existing stance of not enforcing identity verification, and these checks require an account registry Floci does not have.

## Checklist

- [x] `./mvnw test` passes locally (SES suite + SDK compatibility test)
- [x] New or updated integration test added (v1 + v2 integration and an SDK compatibility test)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
@@ -3,7 +3,6 @@ package io.github.hectorvent.floci.services.ses;
 import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsArnUtils;
 import io.github.hectorvent.floci.core.common.AwsException;
-import io.github.hectorvent.floci.core.storage.InMemoryStorage;
 import io.github.hectorvent.floci.core.storage.StorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageFactory;
 import io.github.hectorvent.floci.services.route53.Route53Service;
@@ -151,11 +150,12 @@ public class SesService {
                StorageBackend<String, DedicatedIpPool> dedicatedIpPoolStore,
                StorageBackend<String, ContactList> contactListStore,
                StorageBackend<String, Contact> contactStore,
+               StorageBackend<String, String> policyStore,
                SmtpRelay smtpRelay,
                ObjectMapper objectMapper,
                Clock clock) {
         this(identityStore, emailStore, accountSettingsStore, templateStore, configSetStore, suppressionStore,
-                accountSuppressionStore, dedicatedIpPoolStore, contactListStore, contactStore, smtpRelay,
+                accountSuppressionStore, dedicatedIpPoolStore, contactListStore, contactStore, policyStore, smtpRelay,
                 objectMapper, null, clock);
     }
 
@@ -169,6 +169,7 @@ public class SesService {
                StorageBackend<String, DedicatedIpPool> dedicatedIpPoolStore,
                StorageBackend<String, ContactList> contactListStore,
                StorageBackend<String, Contact> contactStore,
+               StorageBackend<String, String> policyStore,
                SmtpRelay smtpRelay,
                ObjectMapper objectMapper,
                Route53Service route53Service,
@@ -183,7 +184,7 @@ public class SesService {
         this.dedicatedIpPoolStore = dedicatedIpPoolStore;
         this.contactListStore = contactListStore;
         this.contactStore = contactStore;
-        this.policyStore = new InMemoryStorage<>();
+        this.policyStore = policyStore;
         this.smtpRelay = smtpRelay;
         this.objectMapper = objectMapper;
         this.eventPublisher = null;
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
@@ -3,6 +3,7 @@ package io.github.hectorvent.floci.services.ses;
 import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsArnUtils;
 import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.core.storage.InMemoryStorage;
 import io.github.hectorvent.floci.core.storage.StorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageFactory;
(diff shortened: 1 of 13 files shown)


src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
```diff
@@ -162,6 +171,7 @@ public SesService(StorageFactory storageFactory, SmtpRelay smtpRelay, ObjectMapp
         this.accountSuppressionStore = accountSuppressionStore;
         this.dedicatedIpPoolStore = dedicatedIpPoolStore;
         this.contactListStore = contactListStore;
+        this.policyStore = new InMemoryStorage<>();
```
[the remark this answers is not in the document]
okinaka: Fixed. `policyStore` is now a constructor parameter like the other stores (the package-private test constructors take it and the 5 callers pass an `InMemoryStorage`), so it is injectable/inspectable and consistent with the production wiring.

src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
```diff
@@ -162,6 +171,7 @@ public SesService(StorageFactory storageFactory, SmtpRelay smtpRelay, ObjectMapp
...
         this.smtpRelay = smtpRelay;
```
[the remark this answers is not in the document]
okinaka: Fixed. `policyStore` is now a constructor parameter like the other stores (the package-private test constructors take it and the 5 callers pass an `InMemoryStorage`), so it is injectable/inspectable and consistent with the production wiring.

src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
```diff
@@ -1434,6 +1456,171 @@ private static void validateDescription(String description) {
...
+    }
+
+    // v2 UpdateEmailIdentityPolicy: fails if the name is missing; requires the identity.
+    public void updateEmailIdentityPolicy(String identity, String policyName, String policy, String region) {
+        requireIdentityExists(identity, region);
+        validatePolicyName(policyName);
+        String normalized = normalizePolicy(policy);
+        String key = policyKey(region, identity, policyName);
+        synchronized (policyMutationLock) {
+            if (policyStore.get(key).isEmpty()) {
```
abanna, line 1505: @okinaka thanks for adding the policy operations and for cleaning up the storage wiring. One AWS behavior to preserve: both update and delete succeed when the policy name is absent. These guards turn that into `NotFoundException`. Could we keep the identity check, but make update an upsert and delete idempotent? A missing-policy regression case for both endpoints would be great.

okinaka: Thanks @abanna! I went back to real AWS (us-east-1) to double-check, using a verified identity and a fully valid policy document, and both v2 operations return `NotFoundException` when the policy name is absent:

```
[UpdateEmailIdentityPolicy(missing)] NotFoundException: Policy <floci-probe-does-not-exist> does not exist
[DeleteEmailIdentityPolicy(missing)] NotFoundException: Policy <floci-probe-does-not-exist> does not exist
```

(An earlier attempt returned `Invalid ARN` on update only because my test policy used `Resource:"*"` — AWS validates the policy document before the existence check; with a valid `Resource` ARN, update returns `NotFoundException`.)

This matches the v1/v2 asymmetry this PR models: v1 `PutIdentityPolicy` is an upsert and `DeleteIdentityPolicy` is idempotent, but the v2 `UpdateEmailIdentityPolicy` / `DeleteEmailIdentityPolicy` operations are strict and require the policy to exist. The current guards reflect that, and the missing-policy cases are already pinned by `updateMissing_throwsNotFound` / `deleteMissing_throwsNotFound` in the v2 integration test.

Happy to reconsider if you're seeing different behavior in another region or setup — could you share what you observed? Otherwise I'd lean toward keeping the strict v2 behavior to stay faithful to real AWS.

abanna: Thanks for checking that against real AWS—I had the v1 behavior in mind and conflated it with v2 here. The strict `NotFoundException` behavior and the regression tests are the right call. Please keep it as-is; I appreciate the careful verification.

hectorvent (approved):

PR #796, opened 2026-05-11
fix(cloudformation): provision AWS::ApiGateway::Authorizer and wire Method.AuthorizerId

ousamabenyounes:
## Summary

Fixes #788.

`CloudFormationResourceProvisioner` had no `case` for `AWS::ApiGateway::Authorizer`, so the resource fell through the default stub branch and never reached `ApiGatewayService`. Stacks reported `CREATE_COMPLETE`, but `apigateway get-authorizers --rest-api-id <api-id>` came back with an empty list. On top of that, `provisionApiGatewayMethod` did not read `AuthorizerId` from the template, so any `CUSTOM` / `COGNITO_USER_POOLS` method created via CFN was registered with `authorizerId: null`. Net effect: a CloudFormation-managed TOKEN authorizer flow silently degraded to `AuthorizationType: NONE`.

**Reproduces with** (from issue #788):

```text
authorizers:           []
authorizerExists:      false
method.authorizationType: CUSTOM
method.authorizerId:   null
methodHasAuthorizerId: false
broken:                true
```

After the fix, the authorizer is created server-side, `Ref` on the authorizer resource resolves to its real id, the method carries `authorizerId` pointing at that id, and the OOTB AWS CLI / SDK flow works against Floci.

## What changed

- New `case "AWS::ApiGateway::Authorizer"` in `CloudFormationResourceProvisioner` plus a dedicated `provisionApiGatewayAuthorizer` method that maps the CFN PascalCase properties (`Name`, `Type`, `RestApiId`, `AuthorizerUri`, `IdentitySource`, `AuthorizerResultTtlInSeconds`) onto the existing `ApiGatewayService.createAuthorizer` camelCase request shape. The created authorizer's id is stored as the resource's `PhysicalResourceId`, so `!Ref MyAuthorizer` resolves correctly.
- `provisionApiGatewayMethod` now reads `AuthorizerId` (optional) and forwards it to `ApiGatewayService.putMethod`, which already supports the field.
- No service-layer changes — `ApiGatewayService` already exposes `createAuthorizer` and `putMethod(authorizerId)` for the direct REST path. This PR closes the gap for CloudFormation-driven flows only.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

[...]

Issue #788, opened 2026-05-11
[BUG] CloudFormation drops API Gateway TOKEN authorizer and CUSTOM method AuthorizerId


danielkita:
## Service

CloudFormation / API Gateway

## AWS API Action

`cloudformation create-stack` with:

- `AWS::ApiGateway::RestApi`
- `AWS::ApiGateway::Authorizer`
- `AWS::ApiGateway::Method` using `AuthorizationType: CUSTOM`
- `AWS::ApiGateway::Method.AuthorizerId: !Ref Authorizer`

## Expected behavior

CloudFormation should create an API Gateway TOKEN authorizer and wire the `CUSTOM` method to that authorizer.

After stack creation:

```bash
aws apigateway get-authorizers --rest-api-id <api-id>
```

should include the created authorizer.

And:

```bash
aws apigateway get-method \
  --rest-api-id <api-id> \
  --resource-id <resource-id> \
  --http-method GET
```

should include:

```json
{
  "authorizationType": "CUSTOM",
  "authorizerId": "<authorizer-id>"
}
```

## Actual behavior

Floci reports the CloudFormation stack as `CREATE_COMPLETE` and returns an `AuthorizerId` output, but API Gateway has no authorizer resources.

The method is created with `authorizationType: CUSTOM`, but `authorizerId` is missing/null.

Observed output:

```json
{
  "authorizers": [],
  "authorizerExists": false,
  "method": {
    "httpMethod": "GET",
    "authorizationType": "CUSTOM",
    "authorizerId": null
  },
  "methodHasAuthorizerId": false,
  "methodAuthorizerMatches": false,
  "broken": true
}
```

## Reproduction

Start fresh Floci:

```bash
docker rm -f floci-apigw-authorizer-poc 2>/dev/null || true

docker run --name floci-apigw-authorizer-poc \
  -p 4566:4566 \
  -e AWS_ACCESS_KEY_ID=test \
  -e AWS_SECRET_ACCESS_KEY=test \
  -e AWS_DEFAULT_REGION=us-east-1 \
  -e FLOCI_STORAGE_MODE=memory \
  -d floci/floci:latest-compat

until curl -sf http://localhost:4566/_floci/health >/dev/null; do
  sleep 1
done
```

Create and run the POC:

```bash
cat > /tmp/poc-floci-cloudformation-apigateway-authorizer.mjs <<'EOF'
import { execFileSync } from 'node:child_process';
[...]

PR #1760, opened 2026-07-07
fix(cloudformation): provision AWS::ApiGatewayV2::Authorizer and wire Route.AuthorizerId

### LINKED ISSUE #1758: [BUG] CloudFormation drops `AWS::ApiGatewayV2::Authorizer`
## Service

API Gateway v2 (HTTP API) / CloudFormation

## AWS API Action

`CreateAuthorizer` (indirectly, via `cloudformation create-stack`/`deploy` on a template containing `AWS::ApiGatewayV2::Authorizer`)

## Expected behavior

A stack that declares an `AWS::ApiGatewayV2::Authorizer` and references it from a route's `AuthorizerId` creates a real authorizer:

```bash
aws apigatewayv2 get-authorizers --api-id <api-id>
```

should include the created authorizer, and the route should enforce it — a request with no `Authorization` header should be rejected before reaching the integration.

## Actual behavior

Floci reports the stack as `CREATE_COMPLETE`, but `apigatewayv2 get-authorizers` comes back empty. The route is created with `AuthorizationType: JWT`, but with no real authorizer behind it — a request with no `Authorization` header still reaches the integration and gets a normal response instead of a `401`.

Root cause (source inspection): `CloudFormationResourceProvisioner`'s resource-type switch has a `case` for `AWS::ApiGateway::Authorizer` (v1/REST) but none for `AWS::ApiGatewayV2::Authorizer`. The v2 resource falls through the default branch, so `apiGatewayV2Service.createAuthorizer` is never called.

## Reproduction

```bash
docker rm -f floci-apigwv2-authorizer-poc 2>/dev/null || true

docker run --name floci-apigwv2-authorizer-poc \
  -p 4566:4566 \
  -e AWS_ACCESS_KEY_ID=test \
  -d floci/floci:latest

```

```bash
cat > /tmp/floci-apigwv2-authorizer-template.yaml <<'EOF'
Resources:
  MyApi:
    Type: AWS::ApiGatewayV2::Api
    Properties:
      Name: my-api
      ProtocolType: HTTP

  MyAuthorizer:
    Type: AWS::ApiGatewayV2::Authorizer
    Properties:
      ApiId: !Ref MyApi
      Name: my-jwt-authorizer
      AuthorizerType: JWT
      IdentitySource:
        - $request.header.Authorization
      JwtConfiguration:
        Audience: [my-client-id]
        Issuer: https://cognito-idp.us-east-1.amazonaws.com/us-east-1_xxxxx

  MyIntegration:
    Type: AWS::ApiGatewayV2::Integration
    Properties:
      ApiId: !Ref MyApi
      IntegrationType: HTTP_PROXY
      IntegrationMethod: GET
      IntegrationUri: http://example.com
      PayloadFormatVersion: '1.0'

  MyRoute:
    Type: AWS::ApiGatewayV2::Route
    Properties:
      ApiId: !Ref MyApi
      RouteKey: "GET /hello"
      AuthorizationType: JWT
      AuthorizerId: !Ref MyAuthorizer
      Target: !Sub "integrations/${MyIntegration}"

  MyStage:
    Type: AWS::ApiGatewayV2::Stage
    Properties:
      ApiId: !Ref MyApi
      StageName: '$default'
      AutoDeploy: true
EOF

  --template-file /tmp/floci-apigwv2-authorizer-template.yaml \
  --stack-name test-apigwv2-authorizer-stack

API_ID=$(aws --endpoint-url http://localhost:4566 apigatewayv2 get-apis \
  --query "Items[?Name=='my-api'].ApiId" --output text)
echo "ApiId: $API_ID"

echo "--- get-authorizers (expected: one JWT authorizer, actual: empty) ---"
aws --endpoint-url http://localhost:4566 apigatewayv2 get-authorizers --api-id "$API_ID"

echo "--- get-routes (note: no AuthorizerId in the response) ---"
aws --endpoint-url http://localhost:4566 apigatewayv2 get-routes --api-id "$API_ID"

echo "--- invoke with no Authorization header (expected: 401, actual: 200, unauthenticated pass-through) ---"
curl -s -o /dev/null -w "%{http_code}\n" \
  "http://localhost:4566/restapis/$API_ID/\$default/_user_request_/hello"
```

Observed output:

```
ApiId: de443b31d0
--- get-authorizers (expected: one JWT authorizer, actual: empty) ---
{
    "Items": []
}
--- get-routes (note: no AuthorizerId in the response) ---
{
    "Items": [
        {
            "AuthorizationType": "JWT",
            "RouteId": "192ba2d2",
            "RouteKey": "GET /hello",
            "Target": "integrations/ef0e5b9b"
        }
    ]
}
--- invoke with no Authorization header (expected: 401, actual: 200, unauthenticated pass-through) ---
200
```

```bash
docker rm -f floci-apigwv2-authorizer-poc
```

## Additional notes

- #788 is the same failure mode already fixed for API Gateway v1 (`AWS::ApiGateway::Authorizer` + `Method.AuthorizerId`) — this is the same gap in the v2/HTTP API resource type.
- The authorizer runtime itself isn't the problem: `ApiGatewayV2Service.createAuthorizer`/`getAuthorizer` and `ApiGatewayExecuteController.enforceJwtAuthorizer` already work when an authorizer is created via a direct `CreateAuthorizer` call. This is purely a CloudFormation-provisioning gap for the v2 resource type.
- The invoke path above uses floci's REST-style `_user_request_` form; a `Host: <apiId>.execute-api.<region>.amazonaws.com` request to `/hello` 404s in this build — possibly worth a separate report if that's not intentional.

- [x] Yes
- [ ] No

## Environment

rogueserenity:
## Summary

Fixes #1758.

`CloudFormationResourceProvisioner` had no `case` for `AWS::ApiGatewayV2::Authorizer` (only `AWS::ApiGateway::Authorizer`, the v1/REST type, added in #796 for #788). The v2 resource fell through the default branch and was never provisioned — `apiGatewayV2Service.createAuthorizer` was never called, so `apigatewayv2 get-authorizers` came back empty regardless of what the template declared. `provisionApiGatewayV2Route` also never read `AuthorizerId` off the template at all, even though `ApiGatewayV2Service.createRoute`/`updateRoute` already handle it. Net effect: a CloudFormation-managed HTTP API v2 JWT authorizer silently degraded to `AuthorizationType: NONE` in practice — a route configured with `AuthorizationType: JWT` and a real `AuthorizerId` still let every request through unauthenticated, because neither the authorizer nor the route's link to it were ever persisted.

## What changed

- New `case "AWS::ApiGatewayV2::Authorizer"` in `CloudFormationResourceProvisioner`, plus `provisionApiGatewayV2Authorizer`, modeled on the neighboring `provisionApiGatewayV2Route`/`Integration`/`Stage` methods. Maps the CFN PascalCase properties (`Name`, `AuthorizerType`, `IdentitySource`, `JwtConfiguration.Audience`/`Issuer`, `AuthorizerUri`, `AuthorizerPayloadFormatVersion`, `AuthorizerResultTtlInSeconds`, `EnableSimpleResponses`) onto the existing `ApiGatewayV2Service.createAuthorizer`/`updateAuthorizer` camelCase request shape.
- `provisionApiGatewayV2Route` now reads `AuthorizerId` from the template and passes it through, so a route's `Ref` to an authorizer resource actually resolves to something and gets persisted.
- No delete case was added for `AWS::ApiGatewayV2::Authorizer` — consistent with the existing v2 `Route`/`Integration`/`Stage` types, none of which have a case in `delete(...)` either, since that method's signature (`resourceType`, `physicalId`, `region`) has no `apiId` to scope the lookup.

## Test plan

- New integration test `createStack_apiGatewayV2AuthorizerIsProvisionedAndWiredToRoute` in `CloudFormationIntegrationTest`: deploys a template with an `AWS::ApiGatewayV2::Api` + `Authorizer` + `Integration` + `Route` (`AuthorizerId: !Ref Authorizer`), then asserts `GetAuthorizers` returns the authorizer with the right `JwtConfiguration`, and `GetRoutes` shows the route's `AuthorizerId` resolved to that authorizer's real id.
- Ran the full `CloudFormationIntegrationTest` suite (90 tests) plus `RouteAuthorizerIdResponseTest` and `IntegrationConnectionTypeAndAuthorizerSimpleResponsesTest` (the existing v2 authorizer/route regression coverage) — 100/100 pass, no regressions.
- Also manually reproduced the bug against `floci/floci:latest` (1.5.30) via `docker run` + the AWS CLI before writing the fix, confirming `get-authorizers` returned `{"Items":[]}` and an unauthenticated request to the route returned `200` instead of `401`.


rogueserenity:
Good catch — fixed in f252ba5e.

`resolveStringListOrEmpty` silently returned an empty list for a scalar-string `IdentitySource`, which is stricter than `ApiGatewayV2Service.createAuthorizer`/`updateAuthorizer` themselves (they already accept `identitySourceRaw instanceof String`). Added a local `resolveIdentitySource` that accepts both forms, matching the service layer, plus a regression test for the scalar case and a stronger assertion on the existing test (it wasn't checking `IdentitySource` at all before).

One correction on the framing: the [AWS docs for `AWS::ApiGatewayV2::Authorizer.IdentitySource`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-authorizer.html) list the type as "Array of String" and the only example uses array syntax — so the array form is the documented one, not the scalar. The bug is real regardless (the service layer's own leniency should carry through), just wanted to flag that for the record.

rogueserenity:
@hectorvent can you approve this one for CI? I have a follow up to this one once it merges to fix the SAM side of this.

rogueserenity:
@hectorvent ready for you to review/merge as you have time. Thanks!

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -2770,12 +2771,74 @@ private Map<String, Object> parseApiGatewayV2Cors(JsonNode corsNode, CloudFormat
...
+            req.put("jwtConfiguration", jwtConfig);
+        }
+
+        Authorizer authorizer;
+        if (r.getPhysicalId() == null) {
+            authorizer = apiGatewayV2Service.createAuthorizer(region, apiId, req);
+        } else {
+            authorizer = apiGatewayV2Service.updateAuthorizer(region, apiId, r.getPhysicalId(), req);
+        }
+        r.setPhysicalId(authorizer.getAuthorizerId());
```
abanna, line 3284: @rogueserenity thanks for building out the CloudFormation support here. Could we put `AuthorizerId` in the resource attributes as well as the physical ID? `Ref` works, but `Fn::GetAtt Authorizer.AuthorizerId` does not, so route templates receive the literal string and fail later. A small GetAtt route case would cover it.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -2770,12 +2771,74 @@ private Map<String, Object> parseApiGatewayV2Cors(JsonNode corsNode, CloudFormat
...
+        return values;
+    }
+
+    private void provisionApiGatewayV2Authorizer(StackResource r, JsonNode props, CloudFormationTemplateEngine engine,
+                                                 String region) {
+        String apiId = resolveOptional(props, "ApiId", engine);
+        Map<String, Object> req = new HashMap<>();
+        req.put("name", resolveOptional(props, "Name", engine));
+        req.put("authorizerType", resolveOptional(props, "AuthorizerType", engine));
+        req.put("identitySource", resolveIdentitySource(props, "IdentitySource", engine));
```
abanna, line 3257: One runtime follow-up: HTTP API JWT authorizers support `$request.querystring.*` sources too, but the runtime only passes header information into token extraction. A template using a query-string token will 401 even with a valid JWT. Could we pass query parameters through and add a quick execution-level test?

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -277,6 +277,7 @@ public StackResource provision(String logicalId, String resourceType, JsonNode p
                 case "AWS::ApiGateway::Deployment" -> provisionApiGatewayDeployment(resource, properties, engine, region);
                 case "AWS::ApiGateway::Stage" -> provisionApiGatewayStage(resource, properties, engine, region);
                 case "AWS::ApiGatewayV2::Api" -> provisionApiGatewayV2Api(resource, properties, engine, region, accountId, stackName);
+                case "AWS::ApiGatewayV2::Authorizer" -> provisionApiGatewayV2Authorizer(resource, properties, engine, region);
```
abanna, line 306: One cleanup detail: the generic delete path does not know about this resource. If `ApiId` comes from a parameter rather than a stack-owned API, deleting the stack reports success but leaves the authorizer behind. Could we retain the API ID on the resource and add a scoped delete path for it?

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
rogueserenity: Fixed in def4f3b6 — provisionApiGatewayV2Authorizer now also stores r.getAttributes().put("AuthorizerId", authorizer.getAuthorizerId()) alongside the physical id, following the same pattern every other resource type already uses (S3, SQS, VPC, etc.) to make GetAtt resolvable. Fn::GetAtt Authorizer.AuthorizerId reads that attributes map, which was never populated before — only Ref (which reads the physical id) worked.

Added createStack_apiGatewayV2RouteResolvesAuthorizerIdViaGetAtt, a variant of the existing wired-route test that uses Fn::GetAtt instead of Ref for the route's AuthorizerId, asserting it resolves to the real id rather than the literal placeholder string. Verified the test actually catches the regression by temporarily removing the attribute and confirming it fails.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
rogueserenity: Fixed in 34d238b0 — extractToken only ever received HttpHeaders, with no query-parameter access at all, so any IdentitySource like $request.querystring.token was silently unusable and a request using it always 401'd. Threaded UriInfo through enforceJwtAuthorizer/extractToken (it was already available one call site up, in dispatchV2) and added a $request.querystring. branch mirroring the one enforceRequestAuthorizerV2 already has for Lambda REQUEST authorizers.

Added HttpApiJwtAuthorizerQuerystringTest — an execution-level test hitting a real JWT-protected route with the token only in the query string (asserting it gets past authorization), plus checks that a request with no token anywhere 401s, and that a token in the Authorization header alone doesn't satisfy an authorizer configured for a querystring-only identity source. Verified the primary case fails without the fix.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
rogueserenity: Fixed in def4f3b6 — added a scoped case for AWS::ApiGatewayV2::Authorizer in delete(StackResource, region), following the same shape already used for AWS::EKS::Nodegroup (a resource needing more context than type+physicalId to delete). ApiId is now stored as a resource attribute during provisioning and read back here to call apiGatewayV2Service.deleteAuthorizer(region, apiId, physicalId) directly — previously the generic type/physicalId switch had no case for this type at all and silently no-op'd.

Added deleteStack_apiGatewayV2AuthorizerOnNonStackOwnedApiIsRemoved, which deliberately creates the API out-of-band (not as a stack resource) so AWS::ApiGatewayV2::Api's own cascading delete can't be the thing removing the authorizer — that's exactly the scenario you flagged, and it isolates the new scoped delete path as the only thing that could make the test pass. Verified it fails without the fix.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
abanna: Thanks for adding that and for the regression coverage. `Fn::GetAtt` should work properly now.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
abanna: Nice, thanks for wiring that through the execution path and adding the test. This handles the query-string JWT case I was concerned about.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
abanna: Thanks for tracking the API ID and covering the external API case. This looks right now.

hectorvent:
Hello @rogueserenity,

Thanks for the PR!

Could you please update the commit history to remove the Co-authored-by: Claude <...> line from the commit message? We prefer to keep the attribution focused on human contributors for this project. Once that's cleaned up, I'm happy to take another look!

rogueserenity:
Done — history rewritten in e9b35533 to drop the `Co-Authored-By: Claude ...` trailers. Sorry about that.

Only the two commits that carried it were touched (`34d238b0` → `5939b6f1` and `def4f3b6` → `fda5cfce`); the rest were unchanged apart from new SHAs from the rewrite. Merge topology is preserved and the tree is byte-identical to what you last reviewed — `git diff` against the old head is empty, so nothing about the change itself moved.

Audit for the record:

    git log --pretty='%(trailers:key=Co-Authored-By,valueonly)' upstream/main..HEAD

comes back empty across all 7 commits. CI is re-running now.

rogueserenity:
@hectorvent this is ready for you to merge

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -411,6 +412,21 @@ public void delete(StackResource resource, String region) {
             }
             return;
         }
+        // Authorizer deletion needs the api id (a stored attribute, not the physical id, which is
+        // the authorizer id) — same shape as the Nodegroup case above. Without this, the generic
+        // type/physicalId delete path has no case for this type at all and silently no-ops,
+        // leaving the authorizer behind in AWS after the stack reports deleted.
+        if ("AWS::ApiGatewayV2::Authorizer".equals(resourceType)) {
```
abanna, line 429: Heads up before this gets merged — GitHub reports the branch as conflicting with `main` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`), so it can't go in as-is despite the green checks.

Exactly one file collides, and it's right here. `main` has since gained an `AWS::Events::Rule` case in `delete(StackResource, region)` in the same spot this PR adds its `AWS::ApiGatewayV2::Authorizer` case — both immediately after the `AWS::EKS::Nodegroup` block, and both ending on the same shared `return; }` lines, which is what makes git treat it as a conflict rather than two independent insertions.

The resolution is to keep both blocks, each closing with its own `return;`:

```java
if ("AWS::Events::Rule".equals(resourceType)) {
    deleteEventBridgeRuleSafe(resource.getPhysicalId(),
            resource.getAttributes().get("EventBusName"), region);
    return;
}
if ("AWS::ApiGatewayV2::Authorizer".equals(resourceType)) {
    String apiId = resource.getAttributes().get("ApiId");
    ...
    return;
}
```

Worth resolving deliberately rather than by reflex: because that trailing `return; }` is shared context, taking either side wholesale still compiles and CI still goes green — it just silently drops the other side's delete path, either `main`'s custom-bus rule cleanup or the authorizer cleanup this PR exists to add. That failure is invisible until a stack delete leaves something behind.

I checked the rest of the merge and nothing else collides: `resolveOptional`, `resolveStringListOrEmpty`, and the `createAuthorizer` / `updateAuthorizer` / `deleteAuthorizer` signatures are all unchanged on `main`, so no other fixups should be needed after resolving this hunk.

The three things I raised earlier — the `AuthorizerId` attribute, the `$request.querystring.*` identity source, and this scoped delete — all look right at `e9b35533`.

rogueserenity: Conflict resolved. Ready for merge.

hectorvent (approved):
Thanks @rogueserenity @abanna

PR #1773, opened 2026-07-07
feat(lambda): implement the Lambda Extensions API

### LINKED ISSUE #1771: [FEAT] Lambda Extensions API — no `/2020-01-01/extension/*` endpoints, and `/opt/extensions/*` binaries are never launched
## Service

Lambda

## API Action / Feature

Lambda Extensions API (`POST /extension/register`, `GET /extension/event/next`, `POST /extension/init/error`, `POST /extension/exit/error`), plus the runtime-side behavior of discovering and launching every binary under `/opt/extensions/` alongside the main process.

## AWS Documentation

- [Using the Lambda Extensions API to create extensions](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html)

## Why is this needed?

Any real Lambda Extension — most notably [`aws-lambda-web-adapter`](https://github.com/awslabs/aws-lambda-web-adapter), which is the standard way to run a plain HTTP server (Go/Rust/anything with `net/http`-style framework) as a container-image Lambda without SDK-specific handler code — is packaged under `/opt/extensions/<name>` in the image and registers itself against the Extensions API during the `Extension init` phase. Real AWS's runtime interface client discovers and launches every extension binary, and the platform holds `INVOKE` events until all registered extensions have acknowledged readiness.

Floci's `RuntimeApiServer` currently implements only the runtime invocation API (`2018-06-01/runtime/*`) — there is no `2020-01-01/extension/*` route at all, and `ContainerLauncher` only ever runs the image's own `ENTRYPOINT`/`CMD`; nothing discovers or starts binaries under `/opt/extensions/`.

Net effect: a container-image function that uses `aws-lambda-web-adapter` (or any other real Lambda Extension) deploys and starts successfully — the main process's own logs show it running — but every invocation times out. The extension process either exits immediately on a failed/404'd registration call, or never starts at all, so its proxy loop (in the adapter's case, translating Lambda invoke events into real HTTP requests against the app's own port) never runs.

### Confirmed failure mode

Deployed a real Go binary (a plain `net/http` server) packaged per `aws-lambda-web-adapter`'s documented pattern (`FROM public.ecr.aws/lambda/provided:al2023`, extension copied to `/opt/extensions/lambda-adapter`). The container launched correctly and the app's own process logged that it started listening on its configured port. Invoking the function:

```
WARN Function <name> timed out after 10s
```

Inspecting the running container's `/proc` during an in-flight invocation showed exactly one long-lived process: the app's own `bootstrap` binary (PID 1). The extension binary at `/opt/extensions/lambda-adapter` was never running as a persistent process — confirming its registration attempt against `/2020-01-01/extension/register` failed (the route doesn't exist) and the process exited rather than entering its normal register → poll-for-events → proxy loop.

## Suggested scope

Two pieces, both in `services/lambda`:

1. **`RuntimeApiServer`**: add the Extensions API routes, reusing the existing park/dispatch pattern already used for `GET /runtime/invocation/next`:
   - `POST /2020-01-01/extension/register` — reads the `Lambda-Extension-Name` header and an optional `events` array in the body, returns a generated `Lambda-Extension-Identifier` header.
   - `GET /2020-01-01/extension/event/next` — blocks per-extension (each registered extension needs its own event queue, since extensions can subscribe to different event sets) until an `INVOKE` or `SHUTDOWN` event is available.
   - `POST /2020-01-01/extension/init/error`, `POST /2020-01-01/extension/exit/error` — acknowledge and (at minimum) stop delivering further events to that extension.
   - `enqueue()` (on a runtime invocation) and `stop()` (on container shutdown) need to also fan the corresponding `INVOKE`/`SHUTDOWN` event out to every extension subscribed to it.

2. **`ContainerLauncher`**: after starting the container, discover binaries under `/opt/extensions/` (e.g. via `docker exec ls -1 /opt/extensions`) and launch each one as a detached `docker exec` process. This is the piece with no existing equivalent at all today — everything else in `ContainerLauncher` assumes the image's own `ENTRYPOINT`/`CMD` is the only process that needs to run.

A reasonable simplification for a local-dev-focused emulator: invoke completion (`resultFuture.complete(...)`) can continue to depend only on the runtime's own `/response`/`/error` call, rather than implementing the exact real-AWS semantics where the `Invoke` phase only ends once *every* registered extension has also called `/next` again. That precise timing mainly matters for extensions doing work after the response is already returned to the client (e.g. flushing telemetry) — not something a local emulator needs to get byte-perfect to unblock the common case of an extension like `aws-lambda-web-adapter` whose only real gate is successful registration.

- [x] Yes
- [ ] No

### LINKED ISSUE #1882: RuntimeApiServer: make stopped/queue/park state consistently synchronized across all access points
### Summary

`RuntimeApiServer` has several places that read and write shared state (`stopped`, `pendingQueue`, `waitingContexts`, `extensions`, and each `RegisteredExtension`'s pending events) using a lock-free discipline (`volatile` fields, `ConcurrentLinkedQueue`/`ConcurrentHashMap`, poll-then-recheck patterns). This has at least one known race: a request to `GET /2020-01-01/extension/event/next` can observe `stopped == true` in the gap before `stop()` has finished offering that extension's SHUTDOWN event, resulting in a bare 204 with the SHUTDOWN event silently orphaned. See PR #1773, [this review thread](https://github.com/floci-io/floci/pull/1773#discussion_r3590263803) for the detailed characterization of the window and why it's real but narrow in practice.

The same poll-then-check-stopped shape exists in `GET /2018-06-01/runtime/invocation/next` (`NEXT_PATH`), which predates #1773 and was intentionally left untouched there. `/extension/register` and the extension error handlers also mutate `extensions` without going through any lock that `stop()` respects.

### Why this needs a holistic fix, not a per-call-site patch

A version of this was fixed narrowly in #1773 (making just `enqueue()` and `stop()` mutually synchronized), then reverted, because it only closed the window for those two methods while leaving the identical race open in `NEXT_PATH`, `/extension/register`, and `handleExtensionFatalError` — all of which touch the same fields through the original discipline. Patching one call site at a time risks trading one race for another (e.g. synchronizing `enqueue()` against `stop()` without also covering the invoke-notification path can let an extension observe an INVOKE event after it already received SHUTDOWN).

### Suggested approach

Some options worth weighing (not prescriptive — whoever picks this up should choose based on what fits the codebase best):
- Extend a single, consistent locking discipline (e.g. `synchronized(this)` or a dedicated lock object) across every method that reads/writes `stopped`, `pendingQueue`, `waitingContexts`, and `extensions`.
- Alternatively, funnel all state mutation for a given `RuntimeApiServer` instance through a single-threaded executor, so no two operations on that state can interleave at all — removes the need to reason about lock ordering per method, at the cost of making every operation go through an async round-trip.

### Practical impact if left as-is

Low-severity: the window is sub-millisecond, requires two independent operations (an extension's first `/event/next` poll and `stop()`'s notify loop) to land in the same narrow gap, and the only consequence is one extension's final poll getting a 204 instead of a SHUTDOWN body at the moment its container is being killed anyway. Real extensions (e.g. `aws-lambda-web-adapter`) don't gate meaningful cleanup on receiving that specific event.

rogueserenity:
## Summary

Fixes #1771. Fixes #1882.

Floci's `RuntimeApiServer` implemented only the runtime invocation API (`2018-06-01/runtime/*`). Real AWS's init system also discovers and starts every binary under `/opt/extensions/` as a sibling process to the main entrypoint, and those extensions register against a separate Extensions API (`2020-01-01/extension/*`) before the platform delivers events to them. Genuine Lambda Extensions (`aws-lambda-web-adapter`, most observability agents) require both pieces to exist — without them, an extension's registration call 404s, the extension exits, and its own proxy/telemetry loop never starts, even though the main process runs fine.

Confirmed against a real container-image function using `aws-lambda-web-adapter`: the extension process registration was failing silently (no `/2020-01-01/extension/register` endpoint existed at all), so the adapter never started forwarding invocations to the app's HTTP server — every invoke timed out after the function's configured timeout even though the app itself started and logged successfully. Inspecting the running container's `/proc` during an in-flight invocation confirmed only the main process (PID 1) was running; the extension binary never became a persistent process.

## What changed

**`RuntimeApiServer`** — adds:
- `POST /2020-01-01/extension/register` (reads `Lambda-Extension-Name` + an optional `events` body array, returns a generated `Lambda-Extension-Identifier` header)
- `GET /2020-01-01/extension/event/next` (blocking, per-extension queue/park pattern mirroring the existing `GET /runtime/invocation/next` handler)
- `POST /2020-01-01/extension/init/error`, `POST /2020-01-01/extension/exit/error`

`enqueue()` and `stop()` now also fan `INVOKE`/`SHUTDOWN` events out to every extension subscribed to that event type.

Deliberately simplified relative to the real spec: invoke completion depends only on the runtime's own `/response` or `/error` call, not on every registered extension also signaling done via `/next`. Getting the precise post-invoke completion timing right mostly matters for extensions doing work after the response is already returned to the client (e.g. flushing telemetry) — not a fidelity this emulator's local dev/test use case needs, and it would have added real complexity (fan-out barrier synchronization) for marginal benefit.

**`ContainerLauncher`** — adds `launchExtensions`, which `docker exec`s `ls /opt/extensions` after the container starts and, for every binary found, execs it detached as a sibling process. This is the piece Floci had no equivalent for at all — it only ever ran the image's own `ENTRYPOINT`/`CMD`. Best-effort throughout: a function with no `/opt/extensions` directory (the common case) is a silent no-op, and a single extension's launch failure doesn't fail the container launch or block other extensions.

## Test plan

- New `RuntimeApiServerTest` cases exercise the full Extensions API contract over real HTTP: register (success, missing-name-400), `/event/next` (unknown-identifier-403, `INVOKE` delivery gated by subscription — an extension registered for `SHUTDOWN` only is confirmed *not* woken by an invocation — and `SHUTDOWN` delivery on server stop), and `init/error` unregistering the extension.
- New `ContainerLauncherTest` cases verify extension discovery and launch via a mocked `DockerClient` (asserts the exec happens after `startCreated`), plus that a function with no `/opt/extensions` directory launches normally with no extra exec calls beyond the discovery probe.
- Ran the full Lambda/Container test suite (392 tests) in isolation on this branch (based directly on current `main`, no dependency on other open PRs) — all pass, no regressions.
- Manually validated end-to-end against a real Go application (a plain `net/http` server behind `aws-lambda-web-adapter`, deployed via a real `AWS::Serverless::Function` + `AWS::Serverless::HttpApi` template with a Cognito JWT authorizer): confirmed both an unauthenticated `401` rejection and a fully authenticated request reaching the app and returning its real response, including an MCP tool-call round trip.

## Follow-up review changes

Added after the initial review rounds:

- **Extension fatal errors retire the execution environment.** `init/error` and `exit/error` now mark the server faulted; `enqueue()`/`NEXT_PATH` refuse new work and `WarmPool` retires the container instead of pooling or reusing it. The in-flight invocation still completes normally through `/response`.
- **Init-readiness barrier.** Extension binaries are launched as detached execs, so the first invoke could beat `/extension/register` and miss the adapter. `launch()` now waits for the discovered extensions to register (best-effort: no-extension functions don't wait, a timeout proceeds without them).
- **Extension output reaches CloudWatch.** `docker exec` output is a separate stream from PID 1, so extension stdout/stderr never appeared in the container log. `ContainerLogStreamer` gained `execLogCallback()`, sharing frame handling with `attach()`; the log group/stream is also created before extensions launch so their earliest output isn't dropped.
- **`Lambda-Extension-Accept-Feature: accountId`** is honored in the register response.
- **Includes #1882's locking pass** (originally planned as a separate follow-up). The two fixes above touch `stopped`/`pendingQueue`/`waitingContexts`/`extensions`, so they needed the single-lock discipline rather than the previous lock-free one.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
@@ -17,6 +17,9 @@ import io.github.hectorvent.floci.services.lambda.model.LambdaLayerVersion;
 import io.github.hectorvent.floci.services.lambda.runtime.RuntimeApiServer;
 import io.github.hectorvent.floci.services.lambda.runtime.RuntimeApiServerFactory;
 import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.api.async.ResultCallback;
+import com.github.dockerjava.api.model.Frame;
+import com.github.dockerjava.api.model.StreamType;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
@@ -328,6 +331,14 @@ public class ContainerLauncher {
         // Now start the container with code in place
         lifecycleManager.startCreated(containerId, spec);
 
+        // Real AWS's runtime interface client discovers and launches every binary under
+        // /opt/extensions/ as a sibling process to the main entrypoint before the runtime is
+        // considered ready; Floci only runs the image's own ENTRYPOINT/CMD, so extensions
+        // (e.g. aws-lambda-web-adapter) never start without this. Best-effort: an extension
+        // launch failure shouldn't fail the whole container launch, since a function with no
+        // extensions is the common case and this must be a no-op for it.
+        launchExtensions(dockerClient, containerId, fn.getFunctionName());
+
         ContainerHandle handle = new ContainerHandle(containerId, fn.getFunctionName(), runtimeApiServer, ContainerState.WARM, fn.isHotReload());
 
         // Attach log streaming
@@ -598,6 +609,99 @@ public class ContainerLauncher {
 
     private static boolean isProvidedRuntime(String runtime) {
         return runtime != null && runtime.startsWith("provided");
+    }
+
+    /** Directory Lambda's real init system scans for extension binaries. */
+    private static final String EXTENSIONS_DIR = "/opt/extensions";
+
+    /**
+     * Discovers binaries under {@link #EXTENSIONS_DIR} inside the (already started) container and
+     * launches each as a detached process — the piece of real AWS's init system (which starts every
+     * registered extension alongside the runtime) that Floci otherwise has no equivalent for. Uses
+     * {@code docker exec} rather than baking a wrapper entrypoint into the image, since the image is
+     * user-supplied and unmodified.
+     *
+     * <p>Extensions inherit the container's env (already set on the container itself, including
+     * {@code AWS_LAMBDA_RUNTIME_API}) automatically — {@code docker exec} does not need it re-supplied.
+     * Best-effort throughout: a function with no {@code /opt/extensions} directory (the common case)
+     * must be a silent no-op, and a failure launching one extension must not prevent the container
+     * (or other extensions) from running.
+     */
+    private void launchExtensions(DockerClient dockerClient, String containerId, String functionName) {
+        List<String> extensionNames = listExtensionBinaries(dockerClient, containerId, functionName);
+        for (String name : extensionNames) {
+            try {
+                String path = EXTENSIONS_DIR + "/" + name;
+                var create = dockerClient.execCreateCmd(containerId)
+                        .withCmd(path)
+                        .withAttachStdout(true)
+                        .withAttachStderr(true);
+                String execId = create.exec().getId();
+                // Detached: an extension runs for the container's whole lifetime, so this must not
+                // block the launch waiting for it to exit. Its stdout/stderr piggybacks onto the same
+                // ContainerLogStreamer output as the main process via Vert.x's own frame draining below,
+                // rather than left completely undrained (which can backpressure/stall the exec).
+                dockerClient.execStartCmd(execId).exec(new ResultCallback.Adapter<Frame>() {
+                    @Override
+                    public void onNext(Frame frame) {
+                        // Drained, not forwarded to CloudWatch: floci's ContainerLogStreamer already
+                        // tails the container's own PID-1 log stream for that; this only exists so the
+                        // exec's output pipe doesn't fill up and block the extension process.
+                    }
+                });
+                LOG.infov("Launched extension {0} for function {1} (container {2})",
+                        name, functionName, containerId);
+            } catch (Exception e) {
+                LOG.warnv(e, "Failed to launch extension {0} for function {1}; continuing without it",
+                        name, functionName);
+            }
+        }
+    }
+
+    /**
+     * Lists file names directly under {@link #EXTENSIONS_DIR} inside the container, or an empty
+     * list if the directory doesn't exist or the probe fails (most functions have no extensions).
+     */
+    private List<String> listExtensionBinaries(DockerClient dockerClient, String containerId, String functionName) {
+        try {
+            var create = dockerClient.execCreateCmd(containerId)
+                    .withCmd("/bin/sh", "-c", "ls -1 " + EXTENSIONS_DIR + " 2>/dev/null")
+                    .withAttachStdout(true)
+                    .withAttachStderr(false);
+            String execId = create.exec().getId();
+            var stdout = new java.io.ByteArrayOutputStream();
+            var latch = new java.util.concurrent.CountDownLatch(1);
+            dockerClient.execStartCmd(execId).exec(new ResultCallback.Adapter<Frame>() {
+                @Override
+                public void onNext(Frame frame) {
+                    if (frame.getStreamType() == StreamType.STDOUT && frame.getPayload() != null) {
+                        try { stdout.write(frame.getPayload()); } catch (IOException ignored) { /* in-memory */ }
+                    }
+                }
+
+                @Override
+                public void onComplete() {
+                    latch.countDown();
+                }
+            });
+            if (!latch.await(5, TimeUnit.SECONDS)) {
+                LOG.debugv("Timed out listing {0} for function {1}; assuming no extensions",
+                        EXTENSIONS_DIR, functionName);
+                return List.of();
+            }
+            String listing = stdout.toString(java.nio.charset.StandardCharsets.UTF_8).trim();
+            if (listing.isEmpty()) {
+                return List.of();
+            }
+            return java.util.Arrays.stream(listing.split("\\R"))
+                    .map(String::trim)
+                    .filter(s -> !s.isEmpty())
+                    .toList();
+        } catch (Exception e) {
+            LOG.debugv("Could not list {0} for function {1} ({2}); assuming no extensions",
+                    EXTENSIONS_DIR, functionName, e.getMessage());
+            return List.of();
+        }
     }
 
     /**
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
@@ -18,14 +18,13 @@ import io.github.hectorvent.floci.services.lambda.runtime.RuntimeApiServer;
 import io.github.hectorvent.floci.services.lambda.runtime.RuntimeApiServerFactory;
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.async.ResultCallback;
-import com.github.dockerjava.api.exception.NotFoundException;
 import com.github.dockerjava.api.model.Frame;
+import com.github.dockerjava.api.model.StreamType;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
-import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
 
 import java.io.Closeable;
@@ -663,38 +662,59 @@ public class ContainerLauncher {
     /**
      * Lists file names directly under {@link #EXTENSIONS_DIR} inside the container, or an empty
      * list if the directory doesn't exist or the probe fails (most functions have no extensions).
-     *
-     * <p>Uses Docker's archive API ({@code copyArchiveFromContainerCmd}) rather than {@code exec}ing
-     * a shell/{@code find}/{@code basename} pipeline inside the container: a minimal or distroless
-     * function image can have a valid extension binary under {@code /opt/extensions} without any of
-     * those utilities present, which would silently look like "no extensions" to a shell-based probe.
-     * The archive API only talks to the Docker daemon, so it works regardless of the image contents.
      */
     private List<String> listExtensionBinaries(DockerClient dockerClient, String containerId, String functionName) {
-        try (InputStream tarStream = dockerClient.copyArchiveFromContainerCmd(containerId, EXTENSIONS_DIR).exec();
-             TarArchiveInputStream tar = new TarArchiveInputStream(tarStream)) {
+        try {
+            // -maxdepth 1 -type f -perm -u+x: only regular, executable files directly under the
+            // directory (real AWS's documented layout — extension binaries, not subdirectories),
+            // so a stray non-executable file (e.g. a README dropped in by a layer) isn't fed into
+            // the exec loop below as if it were an extension.
+            var create = dockerClient.execCreateCmd(containerId)
+                    .withCmd("/bin/sh", "-c",
+                            "find " + EXTENSIONS_DIR + " -maxdepth 1 -type f -perm -u+x "
+                                    + "-exec basename {} \\; 2>/dev/null")
+                    .withAttachStdout(true)
+                    .withAttachStderr(false);
+            String execId = create.exec().getId();
+            var stdout = new java.io.ByteArrayOutputStream();
+            var latch = new java.util.concurrent.CountDownLatch(1);
+            dockerClient.execStartCmd(execId).exec(new ResultCallback.Adapter<Frame>() {
+                @Override
+                public void onNext(Frame frame) {
+                    if (frame.getStreamType() == StreamType.STDOUT && frame.getPayload() != null) {
+                        try {
+                            stdout.write(frame.getPayload());
+                        } catch (IOException e) {
+                            // ByteArrayOutputStream.write never actually throws IOException (its
+                            // Javadoc documents this), but the checked signature must be handled.
+                            LOG.debugv(e, "Unexpected write failure buffering /opt/extensions listing "
+                                    + "for function {0}", functionName);
+                        }
+                    }
+                }
 
-            List<String> names = new ArrayList<>();
-            TarArchiveEntry entry;
-            while ((entry = tar.getNextEntry()) != null) {
-                if (!tar.canReadEntryData(entry) || entry.isDirectory()) {
-                    continue;
+                @Override
+                public void onComplete() {
+                    latch.countDown();
                 }
-                // Docker wraps the requested directory itself as a leading path component (e.g.
-                // "extensions/lambda-adapter"); only direct children count as extension binaries,
-                // matching the real init system's non-recursive scan of the directory.
-                String name = entry.getName();
-                int slash = name.indexOf('/');
-                if (slash < 0 || name.indexOf('/', slash + 1) >= 0) {
-                    continue;
-                }
-                if ((entry.getMode() & 0111) == 0) {
-                    continue;
-                }
-                names.add(name.substring(slash + 1));
+            });
+            if (!latch.await(5, TimeUnit.SECONDS)) {
+                LOG.debugv("Timed out listing {0} for function {1}; assuming no extensions",
+                        EXTENSIONS_DIR, functionName);
+                return List.of();
             }
-            return names;
-        } catch (NotFoundException e) {
+            String listing = stdout.toString(java.nio.charset.StandardCharsets.UTF_8).trim();
+            if (listing.isEmpty()) {
+                return List.of();
+            }
+            return java.util.Arrays.stream(listing.split("\\R"))
+                    .map(String::trim)
+                    .filter(s -> !s.isEmpty())
+                    .toList();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            LOG.debugv("Interrupted while listing {0} for function {1}; assuming no extensions",
+                    EXTENSIONS_DIR, functionName);
             return List.of();
         } catch (Exception e) {
             LOG.debugv("Could not list {0} for function {1} ({2}); assuming no extensions",
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
@@ -150,6 +150,7 @@ public class ContainerLauncher {
 
         // Start Runtime API server first so container can connect on boot
         RuntimeApiServer runtimeApiServer = runtimeApiServerFactory.create();
+        runtimeApiServer.setFunctionMetadata(fn.getFunctionName(), fn.getVersion(), fn.getHandler());
 
         // Everything after the runtime-api server is allocated runs inside one try/catch: a failure
         // ANYWHERE below — image/host resolve, the code-volume populate (ensureCodeVolume), the spec
@@ -664,8 +665,14 @@ public class ContainerLauncher {
      */
     private List<String> listExtensionBinaries(DockerClient dockerClient, String containerId, String functionName) {
         try {
+            // -maxdepth 1 -type f -perm -u+x: only regular, executable files directly under the
+            // directory (real AWS's documented layout — extension binaries, not subdirectories),
+            // so a stray non-executable file (e.g. a README dropped in by a layer) isn't fed into
+            // the exec loop below as if it were an extension.
             var create = dockerClient.execCreateCmd(containerId)
-                    .withCmd("/bin/sh", "-c", "ls -1 " + EXTENSIONS_DIR + " 2>/dev/null")
+                    .withCmd("/bin/sh", "-c",
+                            "find " + EXTENSIONS_DIR + " -maxdepth 1 -type f -perm -u+x "
+                                    + "-exec basename {} \\; 2>/dev/null")
                     .withAttachStdout(true)
                     .withAttachStderr(false);
             String execId = create.exec().getId();
@@ -675,7 +682,14 @@ public class ContainerLauncher {
                 @Override
                 public void onNext(Frame frame) {
                     if (frame.getStreamType() == StreamType.STDOUT && frame.getPayload() != null) {
-                        try { stdout.write(frame.getPayload()); } catch (IOException ignored) { /* in-memory */ }
+                        try {
+                            stdout.write(frame.getPayload());
+                        } catch (IOException e) {
+                            // ByteArrayOutputStream.write never actually throws IOException (its
+                            // Javadoc documents this), but the checked signature must be handled.
+                            LOG.debugv(e, "Unexpected write failure buffering /opt/extensions listing "
+                                    + "for function {0}", functionName);
+                        }
                     }
                 }
 
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
@@ -18,13 +18,14 @@ import io.github.hectorvent.floci.services.lambda.runtime.RuntimeApiServer;
 import io.github.hectorvent.floci.services.lambda.runtime.RuntimeApiServerFactory;
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.async.ResultCallback;
+import com.github.dockerjava.api.exception.NotFoundException;
 import com.github.dockerjava.api.model.Frame;
-import com.github.dockerjava.api.model.StreamType;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
 
 import java.io.Closeable;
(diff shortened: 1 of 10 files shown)


rogueserenity:
All three addressed in af07eb48.

1. **Hardcoded functionName/functionVersion/handler** — `RuntimeApiServer` now has `setFunctionMetadata(...)`, called from `ContainerLauncher` right after `runtimeApiServerFactory.create()` (the `LambdaFunction` is already in scope there). The server has to stay generically constructed by the factory before that's known, so this couldn't be a constructor param. Added a regression test asserting the register response reflects the real values.

2. **`ls -1` picking up non-executables** — switched to `find /opt/extensions -maxdepth 1 -type f -perm -u+x -exec basename {} \;`, matching real AWS's documented layout (extension binaries live directly under the directory) and only discovering files that are actually executable.

3. **Empty catch block** — the `stdout.write` catch in the frame-draining callback now logs at debug level with the exception, per the project's "never leave a catch block empty" rule (it's true `ByteArrayOutputStream.write` never actually throws, noted in the log message, but the rule applies regardless of the practical risk).

rogueserenity:
@hectorvent can you approve this one please. I merged in upstream/main to fix the doc issue

rogueserenity:
Both fixed in 614eaf16.

1. **SHUTDOWN event orphaned after stop()** — the early `if (stopped) return 204` ran before polling `pendingEvents`, so a request arriving after `stop()` had already offered a SHUTDOWN there (because the extension wasn't parked yet) got a bare 204 instead. Reordered to poll before checking `stopped`, matching the fix direction suggested — an already-queued event is now always delivered.

2. **Dropped event on a concurrent takeWaitingContext race** — confirmed: `raced` could be dequeued and then silently lost if a concurrent `notifyExtensionsOfInvoke`/`stop()` had already consumed the waiting context (dispatching directly rather than via the queue) in between. Split the condition so a dequeued `raced` that can't be delivered gets re-offered to `pendingEvents` instead of dropped, exactly as suggested.

No new automated test for #1 — reproducing it requires a request to reach the handler after `stopped` is set but before the socket actually closes, and `stop()` tears the listener down as part of the same call, so a black-box HTTP test can't observe that window (by the time a client could connect, the server's already gone). The fix is a strict improvement either way (poll-then-check can't be worse than check-then-poll), and the existing `extensionEventNext_receivesShutdownEventWhenServerStops` test continues to cover the still-parked-at-stop-time path.

rogueserenity:
Fixed in 1feb7397.

Confirmed — the outer `catch (Exception e)` around `latch.await(5, TimeUnit.SECONDS)` was catching `InterruptedException` without restoring the thread's interrupt flag. Added a dedicated `catch (InterruptedException e)` that calls `Thread.currentThread().interrupt()` before falling back to the same empty-list result, matching the suggested fix exactly.

rogueserenity:
@hectorvent all green. need anything else before you merge?

hectorvent:
Thanks @rogueserenity,

I will need to go over the implementation. I will try to review it soon.

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
```diff
@@ -138,6 +185,90 @@ public CompletableFuture<Void> start() {
...
+        router.get(EXTENSION_NEXT_PATH).handler(ctx -> {
+            String identifier = ctx.request().getHeader(EXTENSION_ID_HEADER);
+            RegisteredExtension extension = identifier != null ? extensions.get(identifier) : null;
+            if (extension == null) {
+                ctx.response().setStatusCode(403)
+                        .putHeader("Content-Type", "application/json")
+                        .end("{\"errorMessage\":\"Unknown or missing Lambda-Extension-Identifier\"}");
+                return;
+            }
+            // Poll before checking stopped: stop() offers a SHUTDOWN event to pendingEvents for any
```
abanna: The poll-before-stopped ordering is a good fix. There is still a small window: `stop()` sets `stopped` before it queues `SHUTDOWN`. A `/event/next` request in that gap sees no event and returns 204, then the shutdown is orphaned after the server closes. Could we make the shutdown fan-out atomic with the stopped transition and add a small race test?

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
```diff
@@ -619,6 +631,117 @@ private static boolean isProvidedRuntime(String runtime) {
...
+     * list if the directory doesn't exist or the probe fails (most functions have no extensions).
+     */
+    private List<String> listExtensionBinaries(DockerClient dockerClient, String containerId, String functionName) {
+        try {
+            // -maxdepth 1 -type f -perm -u+x: only regular, executable files directly under the
+            // directory (real AWS's documented layout — extension binaries, not subdirectories),
+            // so a stray non-executable file (e.g. a README dropped in by a layer) isn't fed into
+            // the exec loop below as if it were an extension.
+            var create = dockerClient.execCreateCmd(containerId)
+                    .withCmd("/bin/sh", "-c",
```
abanna: Nice cleanup on the extension filtering. One deployment edge case: this assumes `/bin/sh`, `find`, and `basename` exist in every function image. A minimal image can have a valid extension binary but none of those utilities, so discovery quietly returns no extensions. Could we list `/opt/extensions` through Docker’s archive API instead?

rogueserenity: Fixed in 6416b786 — `listExtensionBinaries` now uses `copyArchiveFromContainerCmd` (the same archive API `CodeBuildRunner` already uses elsewhere in this codebase) instead of `exec`ing `/bin/sh -c "find ... -exec basename"`. It reads the returned tar directly, filtering to entries that are direct children of `/opt/extensions` (not nested) and have the executable bit set, so it no longer depends on any utility existing inside the target image at all.

Added coverage for the filtering logic specifically (`launchFunction_extensionDiscovery_filtersNonExecutableAndNestedEntries` — a non-executable file and a nested subdirectory entry are both present in the fake tar and confirmed skipped) and for multiple discovered binaries in one directory. Verified both by temporarily breaking the corresponding filter and confirming the tests fail.

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
rogueserenity: After digging further, I reverted the fix from 6416b786 rather than keeping it.

That fix only made `enqueue()` and `stop()` atomic with each other. But `stopped`, `pendingQueue`, `waitingContexts`, and `extensions` are also read/written directly by `NEXT_PATH`, `/extension/register`, and `handleExtensionFatalError` — all through the original lock-free `volatile`/`ConcurrentLinkedQueue`/`ConcurrentHashMap` discipline, not the new lock. So it closed the window for two call sites while leaving the same race open everywhere else touching that state, including `NEXT_PATH` itself (same poll-then-check-stopped ordering, unmodified, predating this PR).

On the window itself: between `stopped = true` and `stop()` reaching a given extension in its notify loop, the only work in between is a non-blocking `httpServer.close()` call, draining the in-memory `waitingContexts` queue, and iterating `extensions.values()` up to that extension — no I/O, no blocking. For the race to land, that extension's `/event/next` call also has to hit its first poll+stopped-check inside that same window, so two independent things have to coincide, not just "stop() is running." When it does, the only consequence is that extension's *final* poll gets a 204 instead of a SHUTDOWN body, at the moment the container is being killed anyway — invocation handling (`enqueue`/`inFlight`) is unaffected, and extensions like `aws-lambda-web-adapter` don't gate cleanup on receiving that specific event.

A correct fix means giving every touchpoint on this shared state — `NEXT_PATH` included — a consistent discipline, not just the two methods this PR happens to touch. That's cross-cutting enough that I don't think it belongs bundled into this PR; I'll open a follow-up issue covering `RuntimeApiServer`'s state handling as a whole instead of patching one call site at a time. For now, `EXTENSION_NEXT_PATH` is back to the same risk profile as the pre-existing `NEXT_PATH`.

abanna: Thanks for digging into this. I agree a consistent state-discipline pass is better than adding one narrow lock, and keeping it aligned with the existing runtime path makes sense. Please capture both the extension and runtime paths in the follow-up; I’m okay leaving it out of this PR.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
abanna: Nice, thanks—the archive API is exactly the portable route I had in mind. The filtering coverage looks right.

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
```diff
@@ -138,6 +185,90 @@ public CompletableFuture<Void> start() {
...
+                ctx.response().setStatusCode(204).end();
+                return;
+            }
+            extension.setWaitingContext(ctx);
+            // Re-check races the same way NEXT_PATH does: an event or stop() may have landed
+            // between our poll() and setWaitingContext().
+            if (stopped && extension.takeWaitingContext() != null) {
+                ctx.response().setStatusCode(204).end();
+                return;
+            }
```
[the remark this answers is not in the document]
rogueserenity: This is fixed in the follow-up (floci-io/floci#1882) rather than here — the old poll-then-recheck double-check pattern this comment flags is deleted entirely, not patched. The fix introduces a single lock per `RuntimeApiServer` instance guarding `stopped`/`pendingQueue`/`waitingContexts`/`extensions` (and each `RegisteredExtension`'s state), so every handler makes one atomic dispatch decision instead of poll → check-stopped → recheck → poll-again — closing this race and the equivalent one in `NEXT_PATH`, `/extension/register`, and `handleExtensionFatalError` consistently, per the discussion above.

It's ready on my fork (`rogueserenity:fix/runtime-api-server-sync` → `rogueserenity:feat/lambda-extensions-api`) but I'm holding off opening it against `floci-io/main` until this PR merges, since it depends on the Extensions API endpoints introduced here.

rogueserenity:
@abanna do I need to do anything else to get this ready for merge?

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
```diff
@@ -264,4 +417,63 @@ private void sendInvocation(RoutingContext ctx, PendingInvocation invocation) {
...
+                    .put("deadlineMs", event.getDeadlineMs());
+        }
+        ctx.response()
+                .setStatusCode(200)
+                .putHeader("Content-Type", "application/json")
+                .putHeader(EXTENSION_EVENT_ID_HEADER, UUID.randomUUID().toString())
+                .end(body.encode());
+    }
+
+    private void handleExtensionFatalError(RoutingContext ctx, String phase) {
```
abanna, line 673: Thanks for wiring these error endpoints up. Could we retire the execution environment here rather than only unregistering the extension? AWS treats both `init/error` and `exit/error` as fatal to the environment, so the old runtime should not continue taking work. I reproduced that a later invoke can still be served after `init/error`, which could hide a failed adapter or security extension. A regression test proving the old environment is no longer usable would make this solid.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
```diff
@@ -328,6 +333,14 @@ public ContainerHandle launch(LambdaFunction fn) {
         // Now start the container with code in place
         lifecycleManager.startCreated(containerId, spec);
 
+        // Real AWS's runtime interface client discovers and launches every binary under
+        // /opt/extensions/ as a sibling process to the main entrypoint before the runtime is
+        // considered ready; Floci only runs the image's own ENTRYPOINT/CMD, so extensions
+        // (e.g. aws-lambda-web-adapter) never start without this. Best-effort: an extension
+        // launch failure shouldn't fail the whole container launch, since a function with no
+        // extensions is the common case and this must be a no-op for it.
+        launchExtensions(dockerClient, containerId, fn.getFunctionName());
```
abanna: Thanks for the archive-based discovery work — it makes this much more portable. Could we add an init-readiness barrier before the container can receive its first invoke? The extension exec starts asynchronously while the warm-pool path can enqueue right away, so the first invoke can beat `/extension/register` and miss the adapter entirely. A delayed-registration regression test would make this deterministic.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
```diff
@@ -600,6 +613,96 @@ private static boolean isProvidedRuntime(String runtime) {
...
+                        .withAttachStdout(true)
+                        .withAttachStderr(true);
+                String execId = create.exec().getId();
+                // Detached: an extension runs for the container's whole lifetime, so this must not
+                // block the launch waiting for it to exit. Its stdout/stderr piggybacks onto the same
+                // ContainerLogStreamer output as the main process via Vert.x's own frame draining below,
+                // rather than left completely undrained (which can backpressure/stall the exec).
+                dockerClient.execStartCmd(execId).exec(new ResultCallback.Adapter<Frame>() {
+                    @Override
+                    public void onNext(Frame frame) {
```
abanna: Thanks for making sure the exec output is drained. Could we forward these frames through the normal container log path as well? Docker exec output is separate from the container’s PID 1 log stream; I verified locally that it does not appear in `docker logs`. As written, this callback drops extension stdout/stderr instead of sending it to CloudWatch, which is pretty important for observability extensions.

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
```diff
@@ -138,6 +185,90 @@ public CompletableFuture<Void> start() {
...
+            List<String> events = List.of("INVOKE", "SHUTDOWN");
+            var body = ctx.body().asJsonObject();
+            if (body != null && body.getJsonArray("events") != null) {
+                events = body.getJsonArray("events").stream().map(String::valueOf).toList();
+            }
+            String identifier = UUID.randomUUID().toString();
+            extensions.put(identifier, new RegisteredExtension(identifier, name, events));
+            LOG.infov("Extension registered: {0} ({1}), events={2}", name, identifier, events);
+
+            JsonObject responseBody = new JsonObject()
```
abanna, line 334: Small compatibility one: could we honor `Lambda-Extension-Accept-Feature: accountId` here? AWS includes `accountId` in the registration response when that feature is requested. Right now a telemetry extension gets a successful registration but silently loses account attribution. The function already has that value, so a header-driven regression test should keep this straightforward.

rogueserenity:
Pushed fixes for all four comments.

Heads up on scope: this branch now includes the RuntimeApiServer locking follow-up (#1882) we'd agreed to defer. Comments 1 and 2 both touch the state it refactors, and building on the old lock-free discipline would have meant another partial one-call-site fix like the one I reverted earlier here. Bigger diff than you last reviewed — flagging rather than letting you find it.

Every fix verified by breaking the code and confirming the test fails.

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
rogueserenity: Fixed in 35040f7d and 8c4f8a3d.

It was worse than a later invoke being served: `enqueue()` and `NEXT_PATH` only checked `stopped`, so an invocation arriving after `init/error` was queued to a runtime that would never receive it — the resultFuture never completed and the caller hung until the function timeout.

`RuntimeApiServer` now marks itself faulted on both `init/error` and `exit/error`. `enqueue()` and `NEXT_PATH` read `faulted` alongside `stopped` inside the existing atomic decision, and `handleExtensionFatalError` drains anything already queued or parked. `WarmPool` retires a faulted container in `release()` and skips it in `acquire()` — a faulted container is still running, so the liveness probe alone would hand it back out.

The in-flight invocation still completes normally through `/response`; `extensionFatalError_doesNotBreakInFlightInvocation` asserts that.

Regression tests: `invokeEnqueuedAfterExtensionFatalError_isNotServed` plus `WarmPoolTest` coverage of the post-invoke and pooled-fault paths. Each verified by reverting the guard and confirming failure.

Not covered: a container that faults while WARM in the pool is only reaped on the next `acquire()`, not by the idle evictor.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
rogueserenity: Added in 2978f3ce.

`RuntimeApiServer` gains `expectExtensions(n)` / `awaitExtensionsRegistered(timeout)` backed by a `CountDownLatch`. `ContainerLauncher` arms the barrier with the discovered binary count before starting any exec, so a fast-registering extension can't check in before there's a latch to count it down, then waits before handing back the container.

The latch being level-triggered matters: an await arriving after the last registration returns immediately rather than missing the signal. `awaitExtensionsRegistered_returnsImmediatelyWhenRegistrationAlreadyCompleted` pins that.

Best-effort, matching the surrounding code: no-extension functions don't wait, and a timeout logs and proceeds rather than failing the launch. A broken extension reports `init/error` instead, which condemns the environment via the fault path.

`launchFunction_waitsForExtensionsToRegisterBeforeReturning` delays registration and asserts `launch()` hasn't returned, with `InOrder` verification of arm → exec → await. Verified by removing the barrier and confirming it fails.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
rogueserenity: Fixed in 84009753. You're right — `attach()` uses `logContainerCmd`, which only covers PID 1, so an exec's stream never reaches the container log. The comment claiming otherwise is deleted, not reworded; it's what made the gap look intentional.

Put it in `ContainerLogStreamer` as `execLogCallback(...)`, sharing one frame-handling implementation with `attach()` so the two paths can't drift. Extension output lands in the function's own log group/stream. Console prefix carries the extension name for debugging; the CloudWatch message is unprefixed so the stream stays AWS-faithful.

Related bug found doing this: the log group/stream was created by `attach()`, which ran after extensions launched. Extensions can log immediately, and `putLogEvents` against a missing stream is swallowed at debug level — so the earliest output was silently dropped. `ensureLogGroupAndStream` now runs before the extension launch; it's idempotent.

Test uses a real `ContainerLogStreamer` over a mocked `CloudWatchLogsService` and pushes an actual `Frame` through, asserting the message reaches the function's log group. Verified by restoring the no-op callback. Also ran the other nine `ContainerLogStreamer` consumers since that class is shared.

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
rogueserenity: Added in 09847168.

`setFunctionMetadata` now also carries the function's account id (`ContainerLauncher` already had the `LambdaFunction` in scope), falling back to the same `000000000000` placeholder `LambdaService` uses when a function carries none. The register handler adds `accountId` only when the header requests it, since AWS omits the field otherwise.

The header is a comma-separated list, so matching tolerates other features alongside it, whitespace, and casing.

Tests cover with the feature, without it (asserts absence), among other features with odd casing, and the no-account fallback. Verified by disabling the header check and confirming the positive tests fail.

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
rogueserenity: Reversing this — the locking pass is now in this PR (39f0906d, merged as 8fbd26f2) rather than the follow-up. The fault-path guards for `init/error`/`exit/error` and the extension init-readiness barrier both touch this state, and building them on the lock-free discipline would have meant another partial fix like the one I reverted. Both the extension and runtime paths are covered, as you asked.

abanna:
Thanks for the thorough updates, and sorry for one more bit of cleanup. Could you please rewrite the branch to remove the `Co-Authored-By: Claude ...` trailers? Both `CONTRIBUTING.md` and `AGENTS.md` require commit attribution to remain human-only, and I still see them on ten commits in this PR. Thanks!

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
abanna: Thanks for digging into this — the faulted state and WarmPool coverage address the parked/queued invocation issue well. One AWS contract detail still looks open though: after an extension reports init/exit error, AWS says no further API calls succeed. We reject new runtime work, but `/extension/register` and `/extension/event/next` are not gated by `faulted`. Could we make the fault terminal for those endpoints too?

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
abanna: Thanks for adding the barrier — it fixes the registration race I was calling out. I checked the AWS lifecycle docs and there is one remaining distinction: Lambda treats an extension as init-ready after its first `/extension/event/next`, not `register`. Could we wait for that first `Next` and add a delayed-Next test?

rogueserenity:
Both review points fixed in 71f61dcb (now `f87ded55` after the trailer rewrite below), plus one regression I found while in there.

**1. Fault terminal for the Extensions API** — confirmed: `/extension/register` and `/extension/event/next` weren't gated on `faulted`, so a late registration still got a 200 and an identifier that could never receive an event, and a surviving extension's poll still parked on a container already on its way to being retired. Both now answer `500 Extension.SandboxFaulted`.

I went with a hard error rather than the 204 used for orderly shutdown deliberately — an extension reads 204 as "nothing right now" and polls again, so a 204 here would spin it against a dead environment instead of telling it to stop. Happy to change it if you'd rather match a specific status you've seen from real AWS.

**2. Readiness gated on first `Next`, not `register`** — you're right, and the gap was exactly the one the barrier exists to close: an extension that registered fast and then spent time initialising was counted ready during that whole window, letting the first invoke race it. The countdown moved to the first `/event/next`, tracked per-extension via `RegisteredExtension.markFirstNextReceived()` so repeated polling can't open another extension's slot. `awaitExtensionsRegistered` is renamed `awaitExtensionsReady` to match what it now waits for.

Added the delayed-Next test you asked for (`awaitExtensionsReady_waitsForDelayedFirstNextAfterFastRegistration`: registers immediately, asserts the barrier stays closed across a 700ms gap, then polls and asserts it opens), plus `awaitExtensionsReady_countsEachExtensionOnceAcrossRepeatedPolls` for the chatty-extension case.

**3. Unflagged regression — the poll-before-`stopped` ordering came back**

Not something you raised this round, but worth surfacing: the fix from 614eaf16 (reordering `/event/next` to poll `pendingEvents` before checking `stopped`, so a SHUTDOWN queued by `stop()` isn't orphaned) was silently reverted to check-then-poll by the locking pass in 39f0906d. The lock refactor rewrote that block and reintroduced the original ordering.

Restored it here, with the `faulted` check kept ahead of the poll since a fault is terminal even for already-queued events. Flagging it because it means a previously-accepted fix regressed during a later merge on this branch — worth a look if you want to check nothing else from that round got carried away the same way.

**Verification** — each fix confirmed by breaking it and watching the covering test fail: 4 assertion failures plus one hang, the hang being the ungated `/event/next` parking forever, which is the bug itself. Full lambda suite green at 454 tests.

**Trailer rewrite** — branch history rewritten to drop the `Co-Authored-By: Claude` trailers from all 10 commits, per `CONTRIBUTING.md` and `AGENTS.md`. Used `filter-branch --msg-filter` rather than an interactive rebase to preserve the merge commit topology; verified the tree is byte-identical to the pre-rewrite tip, only messages changed. Force-pushed, so you'll want a fresh pull.

src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
abanna: Thanks for the thorough follow-up here — the terminal guards look good now. One last API-contract detail: AWS marks `Lambda-Extension-Function-Error-Type` as required for both `init/error` and `exit/error`. `handleExtensionFatalError` never reads it, so a request with a valid identifier but no error type still returns 202 and faults the environment; I confirmed that locally. Could we validate the required headers before setting `faulted` (and reject missing or unknown identifiers without retiring the container), with focused coverage for those cases?

Separately, I also exercised the case where another extension is already parked in `/extension/event/next` when an error is reported. AWS defines the terminal state but not the exact response for that already-open long poll, so I am happy to leave that as a follow-up rather than expand this PR. Once the header validation is in, I am happy with this.

rogueserenity:
Fixed in e52ded5e.

**Header validation** — confirmed, and it was worse than the 202 suggested: `handleExtensionFatalError` read only the identifier, so a report missing `Lambda-Extension-Function-Error-Type` condemned the environment *and* told the caller it succeeded. Validation now happens before any state changes, since retiring the environment is destructive and unrecoverable. Missing/blank error type → 400; missing/unknown identifier → 403. Neither sets `faulted`, drains the pending queue, or unregisters anything.

**On the identifier check** — worth flagging that this reverses a deliberate choice I made earlier on this branch. I had it condemn regardless of whether the reporting extension was known, reasoning that an unrecognised identifier still meant *some* extension in the container had failed and swallowing that would hide a real failure. You're right and I was wrong: AWS validates first, and my version let anything that could reach the port retire a healthy container. The old comment arguing for the other behaviour is deleted rather than reworded — it's what made the gap look intentional.

Also logging the error-type value now, since it's the reason the environment is being retired and it was being discarded.

**Tests** — five: missing error-type on each endpoint, unknown identifier, missing identifier, and a positive-path test asserting a well-formed report still faults exactly as before (guarding against the validation over-rejecting). Verified by reverting the guards and confirming all four negative tests fail with precisely the behaviour you described — 202 and a faulted environment. Full lambda suite green at 459 tests.

**On the parked-poller case** — agreed, and thanks for scoping it out. Since AWS defines the terminal state but not the response for an already-open long poll, anything I picked would be a guess at unspecified behaviour. Current behaviour for the record: a poller parked before the error gets 204 from the drain in `handleExtensionFatalError`, while one arriving after gets the 500 `Extension.SandboxFaulted`. Happy to revisit in the follow-up if a real extension turns out to care about the difference.

rogueserenity:
Fixed in 01d3b2ab — Greptile's P1 on sibling extension pollers is real, and confirmed.

@abanna, this lands inside the parked-poller case you scoped out, so flagging why I fixed it rather than deferring: there are two separate questions in there, and only one of them is the undefined one.

- **What response** an already-open long-poll should get — genuinely unspecified by AWS, which is what you set aside.
- **Whether it gets one at all** — the poll was never completed. `handleExtensionFatalError` drained `waitingContexts` (the runtime's `/next` connections) and `pendingQueue`, but not the waiting context inside each *other* registered extension. With the faulted guard in `/event/next` meaning nothing will ever wake them, a parked sibling stayed open until `stop()` eventually ran — after `WarmPool.release()` on the next invocation or acquire, up to a full function timeout away.

The hang is a defect under any choice of status code, so I fixed that and deliberately did *not* invent semantics for the first question: it reuses the same `500 Extension.SandboxFaulted` that `/event/next` already returns to anything polling after a fault. Matching shipped behaviour rather than adding a second convention. If you'd rather it be a 204, or a synthetic SHUTDOWN so the extension can run its own cleanup, say the word — that's a one-line change and I have no strong attachment to the 500 here.

Mirrors the `extensions.values()` sweep `stop()` already does, inside the same lock that sets `faulted`, dispatching after release per the locking discipline.

Verified by removing the sweep and confirming the new test fails with a `TimeoutException` rather than an assertion mismatch — i.e. reproducing the hang itself. Full lambda suite green at 460 tests.

rogueserenity:
Fixed in 4e9da023. Greptile's TOCTOU call is correct — the `faulted` read in `/extension/register` was the one guard on that flag taken outside the lock, while `handleExtensionFatalError` sets it inside. A registration could pass the check, block on the monitor, and land in `extensions` after the environment was condemned. Now checked inside the same block as the `put()` it guards, with the response sent after release per the class's locking discipline.

One ordering change falls out: a register call missing `Lambda-Extension-Name` now gets 400 rather than 500 when the environment is already faulted, since the header check runs first. Both are rejections and the 400 is the better diagnostic, but flagging it since it wasn't asked for.

Rather than keep going one round at a time, I audited the whole Extensions API surface for this class of bug before pushing:

- **Every `faulted`/`stopped` read that gates a mutation** — all now inside the lock block containing that mutation. `isFaulted()` stays a deliberate lock-free volatile read for `WarmPool`.
- **Every park site vs every drain site** — both park sites (`waitingContexts.add`, `setWaitingContext`) sit behind guards in the same block, so nothing can park after a drain and hang. That was the sibling-poller bug's shape; it does not recur elsewhere.
- **Double-response** — every dispatch checks `!ended()`, and `takeWaitingContext()` nulls atomically under the lock, so two paths can't both act on one context.
- **Lock held across a Vert.x call** — none; every block collects a decision and dispatches after release.

Two things I found and deliberately did **not** change, since both are judgement calls rather than defects:

1. `enqueue()`'s deferred re-offer can drop an INVOKE into a `RegisteredExtension` that a concurrent fault has already removed from the map. The event is genuinely unreachable afterwards — but it's an INVOKE for an extension in a condemned environment, which must never receive it, so dropping is the correct outcome reached by an accidental route. Fixing it would add locking for no behavioural change.

2. `/extension/event/next` answers 204 on orderly shutdown. AWS documents only 200/403/500 there and never ends the poll empty, so this is a real divergence — but it predates this PR's fault work, and it's unreachable for any extension subscribed to SHUTDOWN (those get the real event from `stop()`'s fan-out). I've documented it in the code as an intentional deviation per AGENTS.md rather than changing shutdown semantics at this stage. Happy to change it if you'd rather.

Full lambda suite green at 460 tests.

Issue #1774, opened 2026-07-07
feat(cloudhsmv2): implement AWS CloudHSM v2 service (phased)


hectorvent:
## Summary

Implement the AWS CloudHSM v2 management API (`cloudhsmv2`, 2017-04-28) in Floci. This is a management-plane-only emulation (clusters, HSMs, backups, tags, resource policies) — no PKCS#11/data-plane. LocalStack Community does not implement CloudHSM at all, and moto covers only 11 of 19 operations with several model divergences, so full-fidelity support here is a differentiator.

## Protocol facts (from botocore [`cloudhsmv2/2017-04-28/service-2.json`](https://github.com/boto/botocore/blob/develop/botocore/data/cloudhsmv2/2017-04-28/service-2.json))

- Protocol: **JSON 1.1**, `X-Amz-Target` prefix **`BaldrApiService.`** (internal codename, like KMS's `TrentService.`)
- Endpoint prefix: `cloudhsmv2` — **SigV4 signing name: `cloudhsm`** (they differ; catalog `credentialScopes` must use `cloudhsm`)
- IDs: `cluster-<base32ish>`, `backup-<...>`, `hsm-<...>`; Backup ARN format `arn:aws:cloudhsm:<region>:<account>:backup/<backupId>`
- Errors: `CloudHsmAccessDeniedException`, `CloudHsmInternalFailureException` (5xx/fault), `CloudHsmInvalidRequestException`, `CloudHsmResourceLimitExceededException`, `CloudHsmResourceNotFoundException`, `CloudHsmServiceException`, `CloudHsmTagException` — all carry only `Message`
- Paginated: `DescribeClusters` (max 25), `DescribeBackups` (max 50), `ListTags` (max 100) — `NextToken`/`MaxResults`
- Cluster states: CREATE_IN_PROGRESS, UNINITIALIZED, INITIALIZE_IN_PROGRESS, INITIALIZED, ACTIVE, UPDATE_IN_PROGRESS, MODIFY_IN_PROGRESS, ROLLBACK_IN_PROGRESS, DELETE_IN_PROGRESS, DELETED, DEGRADED
- HSM states: CREATE_IN_PROGRESS, ACTIVE, DEGRADED, DELETE_IN_PROGRESS, DELETED; Backup states: CREATE_IN_PROGRESS, READY, DELETED, PENDING_DELETION
[...]

PR #1776, opened 2026-07-08
feat(cloudhsmv2): implement CloudHSM v2 service

namanONcode:
## Summary
This pull request introduces support for the AWS CloudHSM v2 service to the emulator. It adds configuration options, integrates CloudHSM v2 into the service catalog, and implements a new JSON handler to process CloudHSM v2 API actions. The changes ensure CloudHSM v2 requests are properly routed and handled in the emulator.

**CloudHSM v2 Service Support**

- Added a new configuration interface `CloudHsmV2ServiceConfig` to `EmulatorConfig.java` and included it in the `ServicesConfig` interface, allowing CloudHSM v2 to be enabled or disabled via configuration. [[1]](diffhunk://#diff-27a905c8526449c5e9a7b7553e391bc16b72b42ba0b4eac1ca9873fcd1d5215dR954-R958) [[2]](diffhunk://#diff-27a905c8526449c5e9a7b7553e391bc16b72b42ba0b4eac1ca9873fcd1d5215dR504)
- Registered CloudHSM v2 in the `ResolvedServiceCatalog`, enabling the emulator to recognize and route requests for this service.

**Controller and Routing Integration**

- Updated `AwsJson11Controller` to inject and route requests to the new `CloudHsmV2JsonHandler`, ensuring CloudHSM v2 API actions are processed. [[1]](diffhunk://#diff-ea7f71ada9c7f879949081385d73db911da48b953c07401bef3ce25937c18bdbR7) [[2]](diffhunk://#diff-ea7f71ada9c7f879949081385d73db911da48b953c07401bef3ce25937c18bdbR97) [[3]](diffhunk://#diff-ea7f71ada9c7f879949081385d73db911da48b953c07401bef3ce25937c18bdbL128-R131) [[4]](diffhunk://#diff-ea7f71ada9c7f879949081385d73db911da48b953c07401bef3ce25937c18bdbR168) [[5]](diffhunk://#diff-ea7f71ada9c7f879949081385d73db911da48b953c07401bef3ce25937c18bdbR231)

**CloudHSM v2 API Handling**

- Implemented `CloudHsmV2JsonHandler`, providing handlers for CloudHSM v2 API actions such as `CreateCluster`, `DescribeClusters`, `DeleteCluster`, `InitializeCluster`, `CreateHsm`, `DeleteHsm`, and tag management. This includes request parsing, response formatting, and error handling for the CloudHSM v2 service.

Refs #1774

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist


src/main/java/io/github/hectorvent/floci/services/cloudhsmv2/CloudHsmV2JsonHandler.java
```diff
@@ -0,0 +1,346 @@
...
+                    .entity(new AwsErrorResponse("CloudHsmInternalFailureException", e.getMessage()))
+                    .build();
+        }
+    }
+
+    // ──────────────────────────── Action Handlers ────────────────────────────
+
+    private Response handleCreateCluster(JsonNode request, String region) {
+        String hsmType = text(request, "HsmType");
+        Map<String, String> subnetMapping = parseStringMap(request.path("SubnetMapping"));
```
abanna: @namanONcode thanks for taking on CloudHSM. One request-shape snag: `CreateCluster` uses `SubnetIds`, while `SubnetMapping` is response-only. An SDK request leaves this null and fails before creating the cluster. Could we parse `SubnetIds` and derive the response mapping instead, with a normal SDK-shaped test?

src/main/java/io/github/hectorvent/floci/services/cloudhsmv2/CloudHsmV2Service.java
```diff
@@ -0,0 +1,400 @@
...
+ *
+ * @see <a href="https://docs.aws.amazon.com/cloudhsm/latest/APIReference/Welcome.html">AWS CloudHSM v2 API Reference</a>
+ */
+@ApplicationScoped
+public class CloudHsmV2Service {
+
+    private static final Logger LOG = Logger.getLogger(CloudHsmV2Service.class);
+    private static final String DEFAULT_HSM_TYPE = "hsm1.medium";
+    private static final String DEFAULT_BACKUP_POLICY = "DEFAULT";
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
```
abanna: One native-build concern: this static `SecureRandom` needs runtime initialization (or an instance-level alternative). The native-image gate captures it in the image heap and currently fails on this class. Could we add it to the existing runtime-init configuration?

src/main/java/io/github/hectorvent/floci/services/cloudhsmv2/CloudHsmV2Service.java
```diff
@@ -0,0 +1,400 @@
...
+        return new LinkedHashMap<>(cluster.getTagList());
+    }
+
+    // ──────────────────────────── Helpers ────────────────────────────
+
+    Cluster getCluster(String clusterId, String region) {
+        if (clusterId == null || clusterId.isBlank()) {
+            throw new AwsException("CloudHsmInvalidRequestException", "ClusterId is required.", 400);
+        }
+        return clusters.get(regionKey(region, clusterId)).orElseThrow(() ->
```
abanna, line 453: One small wire-contract detail: CloudHSM returns `CloudHsmResourceNotFoundException` with HTTP 400, not 404. Could we use the documented status here and add a response-status check so SDK behavior stays aligned?

hectorvent:
@namanONcode please check bot review comments and actions failures.

abanna:
@namanONcode — checking back on this one. The branch head is still `c8aa8976` (2026-07-08), which predates the review, so the three things I raised are all still open at the current head. Re-verified just now rather than assuming:

- `CloudHsmV2JsonHandler.java:77` still reads `parseStringMap(request.path("SubnetMapping"))`. `CreateCluster` takes `SubnetIds` (a list); `SubnetMapping` is response-only, so a real SDK request still leaves this null and fails in `CloudHsmV2Service.java:78` before the cluster is created.
- `CloudHsmV2Service.java:55` still has `private static final SecureRandom SECURE_RANDOM = new SecureRandom();`, and I don't see a matching runtime-initialization entry in `application.yml`. That's what the native image gate is failing on.
- `CloudHsmV2Service.java:247` and `:290` still throw `CloudHsmResourceNotFoundException` with `404`; CloudHSM documents `400` for that shape.

Separately, two things are blocking this independent of the code, and they're both mechanical:

**The Conventional Commits check isn't failing on your CloudHSM work.** The branch carries seven commits that aren't in `main`, and the failure is on these two:

```
c8aa8976 Revert "Merge branch 'main' of https://github.com/namanONcode/floci"
0cc37638 Merge branch 'main' of https://github.com/namanONcode/floci
```

It also still carries four unrelated commits from earlier work:

```
b653cad6 fix: move -march=armv8-a into application.yml native-image build args
2526c2b9 fix: correct -march=armv8-a flag passing to native-image
527a24f1 fix: build native image for baseline armv8-a architecture
ce3d03d9 feat: add GraalVM native image support for Raspberry Pi 3B+ with TUI console
```

Their content isn't in the net diff, but they're in the history the check walks.

**The branch conflicts with `main` in five files**, all of it the usual new-service registration plumbing that drifts as other services land:

```
src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
src/main/java/io/github/hectorvent/floci/core/common/AwsJson11Controller.java
src/main/java/io/github/hectorvent/floci/core/common/ResolvedServiceCatalog.java
src/main/resources/application.yml
src/test/resources/application.yml
```

A rebase onto current `main` keeping only `2558219f` should clear the merge/revert commits, the stale native-image commits, and the conflicts in one go — and would give you a green run to work the three code items against. Happy to re-review once it's back to green.

namanONcode:
@abanna @hectorvent i had implemented suggested fixes , please approve workflows again

abanna (changes requested):
Thank you for taking on the substantial work of implementing CloudHSM v2. I reviewed the current head and ran `CloudHsmV2IntegrationTest`—all 27 tests passed. A few items need to be addressed before this is merge-ready:

1. AWS `DeleteHsm` accepts exactly one of `HsmId`, `EniId`, or `EniIp`, but the handler currently reads only `HsmId`. Please support all three selectors and add SDK-oriented regression coverage.
2. `CloudHsmResourceNotFoundException` is documented as HTTP 400, while the current test and implementation use 404.
3. As a new service, CloudHSM v2 needs a service documentation page and generated Supported Actions table wired through the docs tooling.
4. Please rename the PR to a conventional title such as `feat(cloudhsmv2): implement CloudHSM v2 service`.

AWS documents the selectors and status code in [`DeleteHsm`](https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_DeleteHsm.html). Separately, the fork workflows still need maintainer approval; no action is required from you for that CI gate. Thank you again—once these items are addressed, this should be much closer.

namanONcode:
> Thank you for taking on the substantial work of implementing CloudHSM v2. I reviewed the current head and ran `CloudHsmV2IntegrationTest`—all 27 tests passed. A few items need to be addressed before this is merge-ready:
> 
>     1. AWS `DeleteHsm` accepts exactly one of `HsmId`, `EniId`, or `EniIp`, but the handler currently reads only `HsmId`. Please support all three selectors and add SDK-oriented regression coverage.
> 
>     2. `CloudHsmResourceNotFoundException` is documented as HTTP 400, while the current test and implementation use 404.
> 
>     3. As a new service, CloudHSM v2 needs a service documentation page and generated Supported Actions table wired through the docs tooling.
> 
>     4. Please rename the PR to a conventional title such as `feat(cloudhsmv2): implement CloudHSM v2 service`.
> 
> 
> AWS documents the selectors and status code in [`DeleteHsm`](https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_DeleteHsm.html). Separately, the fork workflows still need maintainer approval; no action is required from you for that CI gate. Thank you again—once these items are addressed, this should be much closer.

Resolved please review it again

abanna:
@namanONcode thanks for the update. I re-reviewed the full PR at `ed1a9ce3`, including the earlier inline threads, and ran the CloudHSM integration test plus AWS CLI compatibility probes.

The `DeleteHsm` selectors, HTTP 400 status, PR title, and generated action table are now addressed. There are still several merge-blocking compatibility gaps:

1. The requested SDK regression coverage is still missing. The new selector tests use handcrafted RestAssured payloads; there is no `CloudHsmV2Client` fixture or CloudHSM test in `compatibility-tests/sdk-test-java` as required by #1774.
2. `DescribeClusters` ignores `MaxResults`, `NextToken`, and `vpcIds`; `ListTags` also ignores pagination. Through the AWS CLI, `--max-results 1` returned two results with no token, and a nonexistent VPC filter returned every cluster.
3. `CreateCluster` accepts invalid/missing `HsmType` and silently drops `Mode`, `NetworkType`, and `BackupRetentionPolicy`. A valid CLI request using FIPS, DUALSTACK, and a 30-day policy returned all three fields as null; `HsmType=bogus` was accepted. Generated cluster/HSM ID suffixes are also 17 UUID-hex characters, outside AWS's documented 11-16-character format.
4. `CreateHsm` does not read `IpAddress` and accepts availability zones outside the cluster's `SubnetMapping`, falling back to the first subnet instead of returning an AWS-style error.
5. The hardware certificates are three independent self-signed certificates, not the `Manufacturer -> AWS -> HSM` chain specified in #1774. OpenSSL showed issuer equal to subject for all three.
6. `docs/services/cloudhsmv2.md` is registered with the action-table generator, but it is absent from both the MkDocs navigation and the canonical services matrix, so the page is currently orphaned.
7. The new commit message `refractor : delete HSM parameters and docs` will fail the Conventional Commits workflow when the fork workflows are approved. It needs to use a valid form such as `refactor(cloudhsmv2): support DeleteHsm selectors and docs`.

Local verification on this exact head: `CloudHsmV2IntegrationTest` passed 31/31 and `make docs-check` passed. `git diff --check` still reports trailing whitespace in `CloudHsmV2Service.java:246`. The GitHub Actions runs remain `action_required`, so CI has not validated this head yet.

Please address these before requesting another review.

namanONcode:
> @namanONcode thanks for the update. I re-reviewed the full PR at `ed1a9ce3`, including the earlier inline threads, and ran the CloudHSM integration test plus AWS CLI compatibility probes.
> 
> The `DeleteHsm` selectors, HTTP 400 status, PR title, and generated action table are now addressed. There are still several merge-blocking compatibility gaps:
> 
>     1. The requested SDK regression coverage is still missing. The new selector tests use handcrafted RestAssured payloads; there is no `CloudHsmV2Client` fixture or CloudHSM test in `compatibility-tests/sdk-test-java` as required by [feat(cloudhsmv2): implement AWS CloudHSM v2 service (phased) #1774](https://github.com/floci-io/floci/issues/1774).
> 
>     2. `DescribeClusters` ignores `MaxResults`, `NextToken`, and `vpcIds`; `ListTags` also ignores pagination. Through the AWS CLI, `--max-results 1` returned two results with no token, and a nonexistent VPC filter returned every cluster.
> 
>     3. `CreateCluster` accepts invalid/missing `HsmType` and silently drops `Mode`, `NetworkType`, and `BackupRetentionPolicy`. A valid CLI request using FIPS, DUALSTACK, and a 30-day policy returned all three fields as null; `HsmType=bogus` was accepted. Generated cluster/HSM ID suffixes are also 17 UUID-hex characters, outside AWS's documented 11-16-character format.
> 
>     4. `CreateHsm` does not read `IpAddress` and accepts availability zones outside the cluster's `SubnetMapping`, falling back to the first subnet instead of returning an AWS-style error.
> 
>     5. The hardware certificates are three independent self-signed certificates, not the `Manufacturer -> AWS -> HSM` chain specified in [feat(cloudhsmv2): implement AWS CloudHSM v2 service (phased) #1774](https://github.com/floci-io/floci/issues/1774). OpenSSL showed issuer equal to subject for all three.
> 
>     6. `docs/services/cloudhsmv2.md` is registered with the action-table generator, but it is absent from both the MkDocs navigation and the canonical services matrix, so the page is currently orphaned.
> 
>     7. The new commit message `refractor : delete HSM parameters and docs` will fail the Conventional Commits workflow when the fork workflows are approved. It needs to use a valid form such as `refactor(cloudhsmv2): support DeleteHsm selectors and docs`.
> 
> 
> Local verification on this exact head: `CloudHsmV2IntegrationTest` passed 31/31 and `make docs-check` passed. `git diff --check` still reports trailing whitespace in `CloudHsmV2Service.java:246`. The GitHub Actions runs remain `action_required`, so CI has not validated this head yet.
> 
> Please address these before requesting another review.

Implemented Backup and other mentioned Issues

abanna (changes requested):
Thank you for the substantial follow-up here — there is a lot of good progress in the latest commits. I re-reviewed the full PR at `6947b462` and reran the focused integration and AWS SDK tests. The earlier `SubnetIds` parsing, HTTP 400 resource-not-found behavior, exact-one `DeleteHsm` selector validation, PR title, and the compile/region fixes all look addressed.

I still found a few AWS-compatibility gaps that I think should be resolved before this lands:

1. **InitializeCluster certificate validation:** the service currently checks that `SignedCert` and `TrustAnchor` parse as X.509 certificates, but it does not verify that the signed certificate matches the cluster CSR, was issued by the supplied CA, or that the trust anchor is self-signed. The SDK test currently passes the emulator's hardware certificate as both values, which exercises the endpoint but not the real AWS initialization contract.

2. **CreateCluster validation and restore semantics:** `HsmType` is still optional and the regex accepts unsupported types; `Mode` and `NetworkType` are not enum-validated; and `SourceBackupId` is stored without looking up or restoring the backup. The UUID-hex ID generator can also emit `0` and `1`, outside CloudHSM's documented `[2-7a-zA-Z]` identifier alphabet.

3. **Backup, tagging, and resource-policy completeness:** `DescribeBackups` does not handle `sourceBackupIds`, `neverExpires`, `Shared`, or `SortAscending`; backup responses omit fields such as `BackupArn`, `HsmType`, and tags; restore does not enforce the deletion window; tagging only resolves clusters rather than backups; and resource policies only resolve backup ARNs. `DeleteResourcePolicy` should also return the previous policy rather than the literal string `"DELETED"`.

4. **CreateHsm IpAddress behavior:** the requested `IpAddress` is read, but a separate random `EniIp` is always generated and returned. The requested address should drive the HSM's returned IP. The HSM response is also missing `HsmType` and, for dual-stack clusters, `EniIpV6`.

5. **Validation error paths:** `ModifyCluster` still directly parses a nullable retention-policy value, while the pagination handlers directly parse numeric tokens and allow zero or negative `MaxResults`. Those malformed requests can become 500/index errors instead of `CloudHsmInvalidRequestException` responses.

6. **Documentation:** the new page is still absent from `mkdocs.yml`, the services overview, and the README service matrices. Its action descriptions are all still `-`, whereas `CONTRIBUTING.md` asks for those placeholders to be filled in.

7. **Repository checks:** the commit `refractor : delete HSM parameters and docs` still fails the repository's exact Conventional Commits workflow. `git diff --check` also reports trailing whitespace in the new SDK test, and the integration test has duplicate `@Order(52)` values.

For verification, `CloudHsmV2IntegrationTest` passed 31/31 and the new live SDK test passed 2/2 locally. Those are useful additions; the certificate shortcut and missing assertions are why the green SDK test does not yet cover the compatibility issues above.

Thanks again for continuing to work through this sizable service implementation. Once these remaining contract and documentation items are addressed, I will be happy to take another look.

namanONcode:
> Thank you for the substantial follow-up here — there is a lot of good progress in the latest commits. I re-reviewed the full PR at `6947b462` and reran the focused integration and AWS SDK tests. The earlier `SubnetIds` parsing, HTTP 400 resource-not-found behavior, exact-one `DeleteHsm` selector validation, PR title, and the compile/region fixes all look addressed.
> 
> I still found a few AWS-compatibility gaps that I think should be resolved before this lands:
> 
>     1. **InitializeCluster certificate validation:** the service currently checks that `SignedCert` and `TrustAnchor` parse as X.509 certificates, but it does not verify that the signed certificate matches the cluster CSR, was issued by the supplied CA, or that the trust anchor is self-signed. The SDK test currently passes the emulator's hardware certificate as both values, which exercises the endpoint but not the real AWS initialization contract.
> 
>     2. **CreateCluster validation and restore semantics:** `HsmType` is still optional and the regex accepts unsupported types; `Mode` and `NetworkType` are not enum-validated; and `SourceBackupId` is stored without looking up or restoring the backup. The UUID-hex ID generator can also emit `0` and `1`, outside CloudHSM's documented `[2-7a-zA-Z]` identifier alphabet.
> 
>     3. **Backup, tagging, and resource-policy completeness:** `DescribeBackups` does not handle `sourceBackupIds`, `neverExpires`, `Shared`, or `SortAscending`; backup responses omit fields such as `BackupArn`, `HsmType`, and tags; restore does not enforce the deletion window; tagging only resolves clusters rather than backups; and resource policies only resolve backup ARNs. `DeleteResourcePolicy` should also return the previous policy rather than the literal string `"DELETED"`.
> 
>     4. **CreateHsm IpAddress behavior:** the requested `IpAddress` is read, but a separate random `EniIp` is always generated and returned. The requested address should drive the HSM's returned IP. The HSM response is also missing `HsmType` and, for dual-stack clusters, `EniIpV6`.
> 
>     5. **Validation error paths:** `ModifyCluster` still directly parses a nullable retention-policy value, while the pagination handlers directly parse numeric tokens and allow zero or negative `MaxResults`. Those malformed requests can become 500/index errors instead of `CloudHsmInvalidRequestException` responses.
> 
>     6. **Documentation:** the new page is still absent from `mkdocs.yml`, the services overview, and the README service matrices. Its action descriptions are all still `-`, whereas `CONTRIBUTING.md` asks for those placeholders to be filled in.
> 
>     7. **Repository checks:** the commit `refractor : delete HSM parameters and docs` still fails the repository's exact Conventional Commits workflow. `git diff --check` also reports trailing whitespace in the new SDK test, and the integration test has duplicate `@Order(52)` values.
> 
> 
> For verification, `CloudHsmV2IntegrationTest` passed 31/31 and the new live SDK test passed 2/2 locally. Those are useful additions; the certificate shortcut and missing assertions are why the green SDK test does not yet cover the compatibility issues above.
> 
> Thanks again for continuing to work through this sizable service implementation. Once these remaining contract and documentation items are addressed, I will be happy to take another look.

all mentioned points resolved please review it

abanna:
@namanONcode thank you for the substantial follow-up here. This is a sizable service implementation, and the latest commits address quite a few of the earlier points: the real AWS SDK lifecycle tests are now present, the `DeleteHsm` selectors and HTTP 400 behavior work, the HSM address fields are handled, the documentation is properly registered, and the commit history and whitespace checks are clean.

I re-reviewed the full PR at `e508d732` and reran both the focused integration suite and the Java SDK tests. I found a few remaining AWS-compatibility edge cases that I think still need attention before this lands:

1. **`InitializeCluster` still does not verify certificate signatures.** The implementation compares issuer/subject names and the signed certificate's public key with the CSR, but it does not cryptographically verify that the trust anchor is actually self-signed or that it signed `SignedCert`. I reproduced this using two different CA keys with the same subject: Floci returned `INITIALIZED`, while `openssl verify` rejected the chain with `certificate signature failure`. AWS requires the cluster certificate to be signed by the supplied self-signed issuing CA: https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_InitializeCluster.html

2. **Pagination validation can still become a 500.** `NextToken="-1"` is accepted by `parseNextToken` and then reaches `subList(-1, ...)`; the live request returned HTTP 500 `InternalFailure` rather than `CloudHsmInvalidRequestException`. The shared `MaxResults` parser also checks only that the value is positive, not each operation's documented maximum.

3. **A few backup paths remain incomplete.** Fresh backups store `NeverExpires` as `"False"`, while SDK/JSON filter values arrive as `"false"`; a live `neverExpires=["false"]` request returned no backups even though the unfiltered request returned one. `Shared=true` is also explicitly hardcoded to return no results. In addition, `CreateCluster` passes a full `SourceBackupId` ARN directly to the ID-keyed local lookup, so the documented full-ARN and shared cross-account restore form cannot work.

4. **CloudHSM-generated IDs are still one character too long.** `generateShortId()` always emits 17 characters, while the documented cluster, backup, and HSM suffix pattern permits 11–16. The focused run generated values such as `cluster-IBpXYMItrHcLmpU7j`: https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_Cluster.html

5. **`DeleteResourcePolicy` still returns the wrong `Policy`.** The service now correctly returns the old policy, but the handler discards it and sends the literal `"DELETED"`. I reproduced that response locally. AWS specifies that this field contains the policy previously attached to the resource: https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_DeleteResourcePolicy.html

6. **Tag validation still needs the AWS bounds.** A missing or empty `TagList` is currently treated as a successful no-op, and the 50-tag maximum is not enforced. AWS requires 1–50 tags for `TagResource`: https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_TagResource.html

There is also a small test-order cleanup remaining: two integration tests still use `@Order(53)`.

Verification on this exact head:

- `CloudHsmV2IntegrationTest`: 31/31 passed
- Java SDK `CloudHsmV2Test`: 2/2 passed
- `make docs-check`: passed
- `git diff --check`: clean
- merge-tree against current `main`: clean

The green tests are useful and show the main lifecycle is working; they just do not cover the invalid-chain and error-path cases above yet. Thank you again for continuing to work through this carefully. The PR is much closer, and I will be happy to take another look after these targeted items are addressed.

namanONcode:
> @namanONcode thank you for the substantial follow-up here. This is a sizable service implementation, and the latest commits address quite a few of the earlier points: the real AWS SDK lifecycle tests are now present, the `DeleteHsm` selectors and HTTP 400 behavior work, the HSM address fields are handled, the documentation is properly registered, and the commit history and whitespace checks are clean.
> 
> I re-reviewed the full PR at `e508d732` and reran both the focused integration suite and the Java SDK tests. I found a few remaining AWS-compatibility edge cases that I think still need attention before this lands:
> 
>     1. **`InitializeCluster` still does not verify certificate signatures.** The implementation compares issuer/subject names and the signed certificate's public key with the CSR, but it does not cryptographically verify that the trust anchor is actually self-signed or that it signed `SignedCert`. I reproduced this using two different CA keys with the same subject: Floci returned `INITIALIZED`, while `openssl verify` rejected the chain with `certificate signature failure`. AWS requires the cluster certificate to be signed by the supplied self-signed issuing CA: https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_InitializeCluster.html
> 
>     2. **Pagination validation can still become a 500.** `NextToken="-1"` is accepted by `parseNextToken` and then reaches `subList(-1, ...)`; the live request returned HTTP 500 `InternalFailure` rather than `CloudHsmInvalidRequestException`. The shared `MaxResults` parser also checks only that the value is positive, not each operation's documented maximum.
> 
>     3. **A few backup paths remain incomplete.** Fresh backups store `NeverExpires` as `"False"`, while SDK/JSON filter values arrive as `"false"`; a live `neverExpires=["false"]` request returned no backups even though the unfiltered request returned one. `Shared=true` is also explicitly hardcoded to return no results. In addition, `CreateCluster` passes a full `SourceBackupId` ARN directly to the ID-keyed local lookup, so the documented full-ARN and shared cross-account restore form cannot work.
> 
>     4. **CloudHSM-generated IDs are still one character too long.** `generateShortId()` always emits 17 characters, while the documented cluster, backup, and HSM suffix pattern permits 11–16. The focused run generated values such as `cluster-IBpXYMItrHcLmpU7j`: https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_Cluster.html
> 
>     5. **`DeleteResourcePolicy` still returns the wrong `Policy`.** The service now correctly returns the old policy, but the handler discards it and sends the literal `"DELETED"`. I reproduced that response locally. AWS specifies that this field contains the policy previously attached to the resource: https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_DeleteResourcePolicy.html
> 
>     6. **Tag validation still needs the AWS bounds.** A missing or empty `TagList` is currently treated as a successful no-op, and the 50-tag maximum is not enforced. AWS requires 1–50 tags for `TagResource`: https://docs.aws.amazon.com/cloudhsm/latest/APIReference/API_TagResource.html
> 
> 
> There is also a small test-order cleanup remaining: two integration tests still use `@Order(53)`.
> 
> Verification on this exact head:
> 
>     * `CloudHsmV2IntegrationTest`: 31/31 passed
> 
>     * Java SDK `CloudHsmV2Test`: 2/2 passed
> 
>     * `make docs-check`: passed
> 
>     * `git diff --check`: clean
> 
>     * merge-tree against current `main`: clean
> 
> 
> The green tests are useful and show the main lifecycle is working; they just do not cover the invalid-chain and error-path cases above yet. Thank you again for continuing to work through this carefully. The PR is much closer, and I will be happy to take another look after these targeted items are addressed.

all mentioned issues resolved please review

namanONcode:
@hectorvent @abanna please review and approve workflows i had fix ci failing issues of native test also fixed approve workflow

namanONcode:
i think ci is all green please review @hectorvent @pgermosen

namanONcode:
i had fixed failing test approve workflow @pgermosen @abanna

src/main/java/io/github/hectorvent/floci/services/cloudhsmv2/CloudHsmV2Service.java
```diff
@@ -0,0 +1,762 @@
...
+        return target;
+    }
+
+    // ──────────────────────────── TagResource ────────────────────────────
+
+    public void tagResource(String resourceId, Map<String, String> tags, String region) {
+        if (tags == null || tags.isEmpty() || tags.size() > 50) {
+            throw new AwsException("CloudHsmInvalidRequestException", "TagList must have length between 1 and 50", 400);
+        }
+        if (resourceId.startsWith("backup-")) {
```
pgermosen, line 435: `tagResource`/`untagResource`/`listTags` (this one plus lines 428 and 440) call `resourceId.startsWith(...)` with no null guard. `CloudHsmV2JsonHandler.text()` returns `null` for a missing `ResourceId` field, so a request omitting it throws an uncaught NPE instead of the `CloudHsmInvalidRequestException` 400 this repo returns for malformed input elsewhere. A `requireNonBlank`-style guard before the `startsWith` call (matching `getCluster`/`getBackup`'s existing null checks) would close this.

src/main/java/io/github/hectorvent/floci/services/cloudhsmv2/CloudHsmV2JsonHandler.java
```diff
@@ -0,0 +1,612 @@
...
+            brp.put("Value", cluster.getBackupRetentionPolicy().getValue());
+            node.set("BackupRetentionPolicy", brp);
+        }
+
+        // SubnetMapping - derive from SubnetIds for response
+        if (cluster.getSubnetIds() != null && !cluster.getSubnetIds().isEmpty()) {
+            ObjectNode subnetNode = objectMapper.createObjectNode();
+            List<String> subnetIds = cluster.getSubnetIds();
+            // Map each subnet to an availability zone
+            for (int i = 0; i < subnetIds.size(); i++) {
```
pgermosen: This AZ assignment is purely positional (`SubnetIds[i]` → `region + (char)('a'+i)`), with no connection to any subnet's actual AZ — and `CloudHsmV2Service.createHsm`'s own AZ lookup (line ~305) uses the identical fabrication, so they're at least internally consistent, but both diverge from reality. A caller naming `SubnetIds` out of alphabetical AZ order, or later calling `CreateHsm` with a real, correct `AvailabilityZone`, can get routed to the wrong subnet or rejected outright.

src/main/java/io/github/hectorvent/floci/services/cloudhsmv2/CloudHsmV2Service.java
```diff
@@ -0,0 +1,762 @@
...
+                                 String mode, String networkType, BackupRetentionPolicy backupRetentionPolicy, String region) {
+        if (SubnetIds == null || SubnetIds.isEmpty()) {
+            throw new AwsException("CloudHsmInvalidRequestException",
+                    "SubnetIds must contain at least one entry.", 400);
+        }
+        if (hsmType == null) {
+            throw new AwsException("CloudHsmInvalidRequestException",
+                    "HsmType is required.", 400);
+        }
+        if (!hsmType.equals("hsm1.medium") && !hsmType.equals("hsm2m.medium")) {
```
pgermosen: `CreateCluster` only accepts the literal strings `hsm1.medium`/`hsm2m.medium`, while `ModifyCluster` (line 729) accepts anything matching `^hsm[1-9][a-z]?\.medium$` — a broader set. A `ModifyCluster` call can move a cluster to an `HsmType` it could never have been created with. AWS's own model pattern is `((p|)hsm[0-9][a-z.]*\.[a-zA-Z]+)` — worth having both call one shared validator against that.

src/main/java/io/github/hectorvent/floci/services/cloudhsmv2/CloudHsmV2Service.java
```diff
@@ -0,0 +1,762 @@
...
+
+        LOG.infov("Created HSM {0} in cluster {1}", hsm.getHsmId(), clusterId);
+        return hsm;
+    }
+
+    // ──────────────────────────── DeleteHsm ────────────────────────────
+
+    public Hsm deleteHsm(String clusterId, String hsmId, String eniId, String eniIp, String region) {
+        int count = 0;
+        if (hsmId != null && !hsmId.isEmpty()) count++;
```
pgermosen: These three conditionals (364-366) are unbraced, which AGENTS.md calls out explicitly ("Always use braces in conditionals") — the same class of thing greptile caught and got fixed elsewhere in the codebase earlier today.
```suggestion
        if (hsmId != null && !hsmId.isEmpty()) {
            count++;
        }
```

src/main/java/io/github/hectorvent/floci/services/cloudhsmv2/CloudHsmV2Service.java
```diff
@@ -0,0 +1,762 @@
...
+        cluster.getHsms().remove(target);
+
+        // If cluster was ACTIVE but now has no HSMs, revert to INITIALIZED
+        if (cluster.getState() == ClusterState.ACTIVE && cluster.getHsms().isEmpty()) {
+            cluster.setState(ClusterState.INITIALIZED);
+            cluster.setStateMessage("No active HSMs");
+        }
+
+        clusters.put(regionKey(region, clusterId), cluster);
+        LOG.infov("Deleted HSM {0} from cluster {1}", hsmId, clusterId);
```
pgermosen: This logs the raw `hsmId` parameter, which is `null` whenever the caller deleted by `EniId`/`EniIp` instead — confirmed live, `CloudHsmV2IntegrationTest` currently prints "Deleted HSM null from cluster ..." for that path. Should log `target.getHsmId()` instead.

compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudHsmV2Test.java
```diff
@@ -0,0 +1,183 @@
...
+    public void testFullLifecycle() {
+        // 1. Create Cluster
+        CreateClusterResponse createClusterResponse = client.createCluster(r -> r
+                .hsmType("hsm1.medium")
+                .subnetIds("subnet-1", "subnet-2")
+                .mode(ClusterMode.FIPS)
+                .networkType(NetworkType.IPV4)
+                .backupRetentionPolicy(b -> b.type("DAYS").value("30"))
+        );
+
```
pgermosen: `git diff --check` still reports trailing whitespace here (and a few nearby lines) — this was flagged in the most recent review round and doesn't look like it's been cleaned up yet.

pgermosen (commented):
Thanks for sticking with this through a lot of review rounds — genuine progress here. I re-verified every currently-unresolved thread against HEAD (`980fdfea`) rather than trusting thread state, and most are actually already fixed in code, just not yet marked resolved: the null-tags NPE, `createHsm` allowing `UNINITIALIZED` clusters, the unused `RegionResolver` wiring, the `JsonErrorResponseUtils` bypass, `decodeBlob`'s charset, the static `SecureRandom` native-build concern, and `CloudHsmResourceNotFoundException` now correctly returning 400. Those don't need further action.

A few things do still need work, listed below with inline comments. Two aren't in any existing thread:

- **`SubnetMapping`/`CreateHsm`'s AZ resolution is fabricated by array index**, not derived from real subnet data. The two resolved threads on this only caught the AZ region string being hardcoded to `us-east-1`; the deeper index-based fabrication survived that fix and can silently mismatch AZs for callers who name real AZs.
- **`CreateCluster` and `ModifyCluster` validate `HsmType` with two different, non-equivalent rules** — a value `ModifyCluster` accepts, `CreateCluster` would reject for the same cluster.

Also: `patch.diff` and `fix-test.patch` look like stray local artifacts that got committed by accident — `patch.diff` literally contains unresolved `<<<<<<< HEAD` conflict markers. Worth dropping both from the PR.

`CloudHsmV2IntegrationTest` is green at 31/31, but none of the existing tests exercise the missing-`ResourceId` or cross-AZ-subnet-ordering paths, which is how these survived this many review rounds.

namanONcode:
@pgermosen all metioned issues is resolved now approve workflow please

namanONcode:
@abanna @hectorvent @pgermosen all checks are green and no merge conflict

hectorvent:
Thank you for sticking with this, and thanks to @abanna and @pgermosen, whose rounds did the heavy lifting here. I checked their open items rather than re-reviewing ground they already covered.

All four of @pgermosen's outstanding points are resolved at this head. The AZ is now taken from the real subnet rather than fabricated from the array index, `HsmType` is validated by the same rule in `CreateCluster` and `ModifyCluster`, and `fix-test.patch` and `patch.diff` are gone, which matters because the latter carried unresolved conflict markers.

**(follow-up, separate PR)** One gap worth naming rather than closing silently. The AZ fix is not pinned by anything: every test uses `us-east-1a`, which is exactly what the old index fabrication produced for the first subnet, so the fixed code and the old bug are indistinguishable under the current suite. The index fallback is still there for unknown subnets, so a refactor back to index-only would go unnoticed. A single case with two subnets whose AZs do not follow array order would close it, and it is the same absence that let this survive several rounds.

Not blocking that on merge after this many rounds. CI is green across all thirteen checks and the branch is clean.

hectorvent (approved):

PR #1581, opened 2026-06-26
feat: implement read-only describe/list APIs for resource collection

Abhi011999:
## Problem

Several read-only AWS actions aren't handled by the emulator. Depending on the route they either return `UnsupportedOperation` (HTTP 400) or — where no route exists — a 404 that the AWS SDKs deserialize to a **retryable** `UnknownError`. Both break clients that enumerate an account read-only:

- the **400** aborts collection of the entire resource type, and
- the **404** sends the SDK through its full retry/backoff schedule (~9 attempts, minutes) on *every* call, stalling the caller.

## Fix

Implement minimal, wire-accurate responses that mirror each service's documented contract so SDK clients complete the read:

| Service | Action(s) | Response |
|---|---|---|
| IAM | `GetLoginProfile` | `NoSuchEntity` (404) — documented "no console password" result |
| IAM | `GetAccessKeyLastUsed` | "never used" shape (`ServiceName`/`Region` = `N/A`, no `LastUsedDate`) |
| IAM | `ListSAMLProviders`, `ListOpenIDConnectProviders`, `ListServerCertificates` | empty lists (non-paginated; `ListServerCertificates` carries `IsTruncated=false`) |
| RDS | `DescribeDBSnapshots`, `DescribeDBProxies`, `DescribeDBClusterSnapshots` | empty result (empty wrapper, no `Marker`) |
| DocDB | `DescribeDBClusterSnapshots` | empty result (RDS namespace) |
| ElastiCache | `DescribeCacheSubnetGroups`, `DescribeCacheParameterGroups` | empty result |
| Glue | `GetJobs`, `GetCrawlers`, `ListDataQualityRulesets`, `GetSecurityConfigurations`, `GetTags` | empty under each result key |
| SSM | `ListDocuments`, `ListAssociations`, `DescribeMaintenanceWindows` | empty (`NextToken` omitted) |
| AWS Backup | `GetBackupVaultNotifications`, `GetBackupVaultAccessPolicy` | `ResourceNotFoundException` (400) — documented "not configured" signal |
| S3 Control | `ListAccessPoints` | empty `AccessPointList` (200), replacing the missing-route 404 that triggered the SDK retry storm |

Each error code, HTTP status, and empty/"never used" shape is grounded in the AWS API reference for that action.

#### W/o seeded data:

Service | Action (CLI) | before (4576) | after (4566)
-- | -- | -- | --
IAM | get-login-profile --user-name alice | ❌ UnsupportedOperation | ✅ NoSuchEntity
[...]

PR #1769, opened 2026-07-07
docs: sync service action tables

jvanzyl:
## Summary
- Regenerate service action tables for DocDB, ElastiCache, and RDS.
- Keep generated docs in sync with the current action registry.

## Validation
- `make docs-sync`
- `make docs-check`

PR #1779, opened 2026-07-08
Update service action tables

electrum:
Regenerates service action tables so documented supported actions match the current handlers.

Root cause: PR #1581 added read-only DocDB, ElastiCache, and RDS actions without regenerating these tables. The dedicated Docs Action Tables workflow is already the enforcement point; this PR brings the generated docs back into sync.

Checks run:
- make PYTHON=target/docs-venv/bin/python docs-test
- make PYTHON=target/docs-venv/bin/python docs-check


electrum:
Closing as a duplicate of #1769, which contains the same generated service action table updates.

Abhi011999:
Hi @electrum 
My apologies for not handeling this.
If this is not mentioned in `AGENTS.md` file, I guess we should add a line for this too.
Thanks!

PR #1789, opened 2026-07-09
feat(bedrock): add proxy backend for real LLM responses via OpenAI-compatible API

### LINKED ISSUE #1788: [FEAT] Bedrock Runtime: proxy backend for real LLM responses via OpenAI-compatible API
## Service

Bedrock Runtime

## API Action / Feature

`Converse` - add an optional `proxy` backend that forwards requests to any OpenAI-compatible endpoint (Ollama, OpenRouter, LiteLLM, vLLM, ...) instead of always returning the hardcoded stub response.

## AWS Documentation

https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html

## Why is this needed?

The current Bedrock Runtime implementation returns hardcoded stub responses only. This makes it impossible to test real LLM interactions locally without an AWS account.

It also matters for integration testing: teams whose application code calls Bedrock directly currently have to build and maintain their own stub/mock of Bedrock to test that code path in CI or locally. With a proxy backend, they can point their existing Bedrock SDK client at Floci and get real responses from whatever OpenAI-compatible backend they already use (a local Ollama model, an existing OpenRouter/LiteLLM gateway, etc.) - Floci does the stubbing, so the integration under test doesn't have to fake one.

Floci does not need to know or care what is behind the URL. The user configures the target:

```bash
# Ollama (local, no API key needed)
FLOCI_SERVICES_BEDROCK_RUNTIME_BACKEND=proxy
FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_URL=http://localhost:11434/v1
FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_DEFAULT_MODEL=llama3.2

# OpenRouter (cloud)
FLOCI_SERVICES_BEDROCK_RUNTIME_BACKEND=proxy
FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_URL=https://openrouter.ai/api/v1
FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_API_KEY=sk-or-...
FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_MODEL_MAPPING=anthropic.claude-3-sonnet-20240229-v1:0=anthropic/claude-3-sonnet

# LiteLLM (self-hosted gateway, passthrough model ids)
FLOCI_SERVICES_BEDROCK_RUNTIME_BACKEND=proxy
FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_URL=http://localhost:4000/v1
FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_MODEL_PASSTHROUGH=true
```

Scope for a first PR: Converse (text + tool use/function calling), model resolution (explicit mapping -> passthrough -> default -> 400 error), `stub` stays the default so there's no breaking change for existing users.

Follow-ups, deliberately out of scope here: `ConverseStream` (needs an AWS Event Stream binary encoder), `InvokeModel` proxying (provider-specific body shapes), Guardrails (no OpenAI-compatible equivalent).

- [x] Yes
- [ ] No

MilleR98:
## Summary

Adds an optional `proxy` backend to Bedrock Runtime that forwards `Converse` requests to any OpenAI-compatible endpoint (Ollama, OpenRouter, LiteLLM, vLLM, ...). The `stub` backend remains the default - no existing behaviour changes. Closes #1788.

Beyond letting you see a real model reply locally, this is useful for integration testing: an application that calls Bedrock directly doesn't need its own stub/mock of Bedrock to exercise that code path in CI or locally - it points its existing SDK client at Floci, which does the stubbing by proxying to whatever OpenAI-compatible backend is already available (a local Ollama model, an existing LiteLLM/OpenRouter gateway, etc.).

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against `boto3`'s `bedrock-runtime` client (`converse`) and the AWS CLI (`aws bedrock-runtime converse`) — the proxy backend's response still matches the real Converse response shape (`output.message`, `stopReason`, `usage`, `metrics`), so existing SDK/CLI call sites work unmodified regardless of which backend is active. Request-side translation covers the real Converse fields: `system`, `messages` (including multi-turn `toolUse`/`toolResult` content blocks), `toolConfig.tools`, `toolConfig.toolChoice` (`auto`/`any`/forced single-tool), and `inferenceConfig` (`maxTokens`, `temperature`, `topP`, `stopSequences`). `outputConfig` was deliberately **not** implemented - it isn't part of the real Bedrock Converse request shape (verified against AWS's documented fields: `modelId`/`messages`/`system`/`inferenceConfig`/`toolConfig`/`guardrailConfig`/ `additionalModelRequestFields`/`promptVariables`/`additionalModelResponseFieldPaths`/ `requestMetadata`/`performanceConfig`), so translating it would mean accepting a field real Bedrock never sends.

## Changes

- `BedrockBackend` — new interface with `converse` and `invokeModel`
- `StubBackend` — existing stub logic extracted here, behaviour unchanged
- `ProxyBackend` — new backend, translates Bedrock ↔ OpenAI Chat Completions
- `BedrockOpenAiTranslator` — pure Bedrock Converse ↔ OpenAI Chat Completions JSON translation: text, multi-turn tool use, tool choice forcing, and inferenceConfig
- `BedrockRuntimeService` — selects the backend based on `floci.services.bedrock-runtime.backend` and delegates; `BedrockRuntimeController` threads the parsed request through
- `EmulatorConfig` + `application.yml` — new `floci.services.bedrock-runtime.backend` / `.proxy.*` config properties
- `docs/services/bedrock-runtime.md` — documents the new backend and config (also fixes a pre-existing typo in the `enabled` env var name)
- `BedrockProxyIntegrationTest`, `BedrockProxyPassthroughIntegrationTest`, `BedrockProxyInvalidUrlIntegrationTest` — cover text responses, tool use, tool choice, inferenceConfig, model resolution, and proxy error handling

## How to test locally

With Ollama:

```yaml
services:
  floci:
    image: floci/floci:latest
    ports: ["4566:4566"]
    environment:
      FLOCI_SERVICES_BEDROCK_RUNTIME_BACKEND: proxy
      FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_URL: http://ollama:11434/v1
      FLOCI_SERVICES_BEDROCK_RUNTIME_PROXY_DEFAULT_MODEL: llama3.2
  ollama:
    image: ollama/ollama:latest
```

## Out of scope

- `ConverseStream` -  needs AWS Event Stream binary encoder
- `InvokeModel` proxy - currently returns the stub response with a log warning
- Guardrails - no OpenAI-compatible equivalent

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/BedrockOpenAiTranslator.java
+++ b/src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/BedrockOpenAiTranslator.java
@@ -205,6 +205,13 @@ final class BedrockOpenAiTranslator {
         ArrayNode content = outMessage.putArray("content");
 
         if (hasToolCalls) {
+            // Some OpenAI-compatible backends emit assistant text alongside tool_calls (e.g. a
+            // preamble like "Let me check that for you") — preserve it as a leading text block
+            // instead of dropping it.
+            String leadingText = extractMessageText(message);
+            if (!leadingText.isBlank()) {
+                content.addObject().put("text", leadingText);
+            }
             for (JsonNode toolCall : toolCallsNode) {
                 ObjectNode toolUse = content.addObject().putObject("toolUse");
                 toolUse.put("toolUseId", toolCall.path("id").asText(""));
--- a/src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/BedrockOpenAiTranslator.java
+++ b/src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/BedrockOpenAiTranslator.java
@@ -1,0 +1,272 @@
+package io.github.hectorvent.floci.services.bedrockruntime.backend;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.jboss.logging.Logger;
+
+/**
+ * Translates between the Bedrock Runtime Converse wire shape and the OpenAI
+ * Chat Completions wire shape used by Ollama, OpenRouter, LiteLLM, vLLM, etc.
+ */
+final class BedrockOpenAiTranslator {
+
+    private static final Logger LOG = Logger.getLogger(BedrockOpenAiTranslator.class);
+
+    private BedrockOpenAiTranslator() {
+    }
+
+    static ObjectNode toOpenAiRequest(ObjectMapper mapper, ObjectNode bedrockRequest, String resolvedModel) {
+        ObjectNode openAi = mapper.createObjectNode();
+        openAi.put("model", resolvedModel);
+        openAi.put("stream", false);
+        ArrayNode openAiMessages = openAi.putArray("messages");
+
+        JsonNode system = bedrockRequest.path("system");
+        if (system.isArray()) {
+            for (JsonNode block : system) {
+                String text = block.path("text").asText(null);
+                if (text != null) {
+                    ObjectNode msg = openAiMessages.addObject();
+                    msg.put("role", "system");
+                    msg.put("content", text);
+                }
+            }
+        }
+
+        JsonNode messages = bedrockRequest.path("messages");
+        if (messages.isArray()) {
+            for (JsonNode message : messages) {
+                translateMessage(mapper, openAiMessages, message);
+            }
+        }
+
+        JsonNode toolConfig = bedrockRequest.path("toolConfig");
+        JsonNode tools = toolConfig.path("tools");
+        if (tools.isArray() && !tools.isEmpty()) {
+            ArrayNode openAiTools = openAi.putArray("tools");
+            for (JsonNode tool : tools) {
+                JsonNode toolSpec = tool.path("toolSpec");
+                if (!toolSpec.isObject()) {
+                    continue;
+                }
+                ObjectNode function = putFunctionRef(openAiTools.addObject(), toolSpec.path("name").asText(""));
+                if (toolSpec.hasNonNull("description")) {
+                    function.put("description", toolSpec.path("description").asText());
+                }
+                JsonNode schema = toolSpec.path("inputSchema").path("json");
+                if (schema.isObject()) {
+                    function.set("parameters", schema.deepCopy());
+                }
+            }
+
+            // "auto" needs no explicit output: it's OpenAI's default whenever tools are present.
+            JsonNode toolChoice = toolConfig.path("toolChoice");
+            if (toolChoice.hasNonNull("any")) {
+                openAi.put("tool_choice", "required");
+            } else if (toolChoice.hasNonNull("tool")) {
+                String toolName = toolChoice.path("tool").path("name").asText("");
+                if (!toolName.isBlank()) {
+                    putFunctionRef(openAi.putObject("tool_choice"), toolName);
+                }
+            }
+        }
+
+        JsonNode inferenceConfig = bedrockRequest.path("inferenceConfig");
+        JsonNode maxTokens = inferenceConfig.path("maxTokens");
+        if (maxTokens.isNumber()) {
+            openAi.put("max_tokens", maxTokens.asInt());
+        }
+        JsonNode temperature = inferenceConfig.path("temperature");
+        if (temperature.isNumber()) {
+            openAi.put("temperature", temperature.asDouble());
+        }
+        JsonNode topP = inferenceConfig.path("topP");
+        if (topP.isNumber()) {
+            openAi.put("top_p", topP.asDouble());
+        }
+        JsonNode stopSequences = inferenceConfig.path("stopSequences");
+        if (stopSequences.isArray() && !stopSequences.isEmpty()) {
+            openAi.set("stop", stopSequences.deepCopy());
+        }
+
+        return openAi;
+    }
+
+    /** Builds the OpenAI {@code {type: "function", function: {name}}} shape shared by tools[] entries and tool_choice. */
+    private static ObjectNode putFunctionRef(ObjectNode target, String name) {
+        target.put("type", "function");
+        return target.putObject("function").put("name", name);
+    }
+
+    /**
+     * A Bedrock message's content[] blocks map to OpenAI in three different ways:
+     * plain text accumulates into the message's content string, toolUse blocks become
+     * entries in that same message's tool_calls[], and toolResult blocks become their
+     * own standalone OpenAI message with role=tool — so one Bedrock message can expand
+     * into zero, one, or several OpenAI messages.
+     */
+    private static void translateMessage(ObjectMapper mapper, ArrayNode openAiMessages, JsonNode message) {
+        String role = message.path("role").asText("user");
+        JsonNode content = message.path("content");
+        StringBuilder text = new StringBuilder();
+        ArrayNode toolCalls = null;
+
+        if (content.isArray()) {
+            for (JsonNode block : content) {
+                if (block.has("toolResult")) {
+                    JsonNode toolResult = block.path("toolResult");
+                    ObjectNode toolMsg = openAiMessages.addObject();
+                    toolMsg.put("role", "tool");
+                    toolMsg.put("tool_call_id", toolResult.path("toolUseId").asText(""));
+                    toolMsg.put("content", extractToolResultText(toolResult.path("content")));
+                    continue;
+                }
+                if (block.has("toolUse")) {
+                    JsonNode toolUse = block.path("toolUse");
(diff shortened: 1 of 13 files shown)


src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/BedrockOpenAiTranslator.java
```diff
@@ -0,0 +1,272 @@
...
+        JsonNode toolCallsNode = message.path("tool_calls");
+        boolean hasToolCalls = toolCallsNode.isArray() && !toolCallsNode.isEmpty();
+
+        ObjectNode root = mapper.createObjectNode();
+        ObjectNode output = root.putObject("output");
+        ObjectNode outMessage = output.putObject("message");
+        outMessage.put("role", "assistant");
+        ArrayNode content = outMessage.putArray("content");
+
+        if (hasToolCalls) {
```
abanna, line 207: @MilleR98 thanks for adding the Bedrock translation layer. One mixed-response edge case: OpenAI-compatible responses can include assistant text and `tool_calls` together, but this branch emits only `toolUse` blocks and drops the text. Could we append a text block when content is non-empty and add a mixed-case translation test?

MilleR98: @abanna Good catch, thanks! 💯 
Just pushed a fix that ensures leading text appears as its own content block ahead of toolUse blocks when both are present, plus a regression test for this case 👍🏻

abanna: Thanks—nice quick fix. Preserving the leading text before the tool-use block, with the regression test, is exactly what I was looking for.

hectorvent:
@MilleR98 Thanks for the PR, 

Please take a look at the failures in the actions.

MilleR98:
> @MilleR98 Thanks for the PR,
> 
> Please take a look at the failures in the actions.

hmm, interesting, failed tests are not related to my changes, but will have a look 👍🏻

MilleR98:
> @MilleR98 Thanks for the PR,
> 
> Please take a look at the failures in the actions.

Hey @hectorvent 
Could we re-run the actions, please? just ran the full suite locally - all passing ✅ 
Also, rebased my branch onto main.

src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/ProxyBackend.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+        builder.POST(HttpRequest.BodyPublishers.ofString(requestBody));
+
+        long start = System.nanoTime();
+        HttpResponse<String> response;
+        try {
+            response = httpClient.send(builder.build(), HttpResponse.BodyHandlers.ofString());
+        } catch (Exception e) {
+            LOG.warnv("Bedrock proxy backend call failed: modelId={0}, url={1}, error={2}", modelId, uri, e.getMessage());
+            throw new AwsException("ModelErrorException", "Failed to reach proxy backend: " + e.getMessage(), 424);
```
abanna, line 94: Nice to see this land with the leading-text fix carried through.

One error-mapping detail on this catch. The 60s request timeout and the 10s connect timeout both surface here as `HttpTimeoutException` / `HttpConnectTimeoutException`, and the generic `catch (Exception e)` turns them into `ModelErrorException` with 424. Bedrock has a dedicated shape for exactly this case — `ModelTimeoutException`, HTTP 408 ("The request took too long to process") — and the distinction is load-bearing rather than cosmetic: the AWS SDKs retry 408 under the default retry strategy but treat 424 as terminal. So a client with retries configured silently won't retry a timeout it would have retried against real Bedrock.

This is easy to hit in the intended setup, too — 60s is not much headroom for a larger model on CPU-backed Ollama, which is one of the configurations called out in the class javadoc.

Could the timeout cases be split out ahead of the generic catch?

```java
} catch (HttpTimeoutException e) {
    throw new AwsException("ModelTimeoutException", "Proxy backend timed out: " + e.getMessage(), 408);
} catch (Exception e) {
    ...
}
```

`HttpConnectTimeoutException` extends `HttpTimeoutException`, so one clause covers both. A test that points the proxy at a deliberately slow endpoint would pin the 408 down.

src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/ProxyBackend.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        return BedrockOpenAiTranslator.toBedrockResponse(objectMapper, openAiResponse, latencyMs);
+    }
+
+    @Override
+    public byte[] invokeModel(String modelId, byte[] body) {
+        LOG.warnv("Bedrock proxy backend does not support InvokeModel yet; returning stub response for modelId={0}",
+                modelId);
+        return stubBackend.invokeModel(modelId, body);
```
abanna: This fallback is documented in the class javadoc, so I realize it's deliberate — but I think it's worth reconsidering how it fails, because it's inconsistent with how the other unsupported operations in this PR behave.

`converse-stream` and `invoke-with-response-stream` both return an explicit unsupported error to the caller. `InvokeModel` under `backend=proxy` instead returns `stubBackend.invokeModel(...)`, so the caller gets a well-formed, plausible-looking completion that is entirely canned, with the only signal being a `WARN` line in the emulator log. Someone who has just configured a real backend and pointed an existing app at it has no way to tell from the response that the model was never consulted — and `InvokeModel` is still what a lot of SDK wrappers reach for by default.

Failing loudly, the way the streaming paths already do, seems both safer and more consistent:

```java
throw new AwsException("ValidationException",
        "InvokeModel is not supported by the bedrock-runtime proxy backend; use Converse.", 400);
```

If the silent fallback is deliberate for compatibility reasons, could the two streaming endpoints' messages get the same treatment in reverse? They currently say "not supported by the Floci **stub**", which reads as inaccurate once `backend=proxy` is configured — the reason it's unsupported has nothing to do with the stub at that point.

hectorvent:
Thank you for this, and for the patience while it waited for a human reply, that was on us. The design concern is fully answered: stub stays the default, nothing leaves the machine without an explicit opt in, and the unreachable proxy case fails fast with a declared error and a test. I checked the translation against the model and every member name is exact, including ModelErrorException at HTTP 424, per [botocore bedrock-runtime](https://github.com/boto/botocore/blob/develop/botocore/data/bedrock-runtime/2023-09-30/service-2.json).

**(blocking)** One code slipped through: the serialization catch at `services/bedrockruntime/backend/ProxyBackend.java:82` throws "InternalFailure", which is not a bedrock runtime shape in botocore or the Java SDK, so clients cannot map it. `InternalServerException` is the declared 500.

**(follow-up, separate PR)** finish_reason content_filter could map to content_filtered, and outputConfig is now a real Converse field in the current model (the docs page lags); your code already ignores it safely, and its textFormat maps naturally onto OpenAI's response_format.

That one word change is the only blocker.

src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/ProxyBackend.java
MilleR98: Good point!
fixed - now there is explicit case for the `HttpTimeoutException` (with correct mapping to `ModelTimeoutException/408`) ahead of the generic catch.

Regarding hardcoded timeout -  now the request timeout is configurable (proxy.request-timeout-seconds, default 60s). Added related integration test to assert ModelTimeoutException.

src/main/java/io/github/hectorvent/floci/services/bedrockruntime/backend/ProxyBackend.java
MilleR98: Good point on both counts - fixed!
`invokeModel()` now fails fast under `backend=proxy:`, no more silent fallback.

Also fixed the wording on both currently-unsupported endpoints in BedrockRuntimeController, since they're unavailable for either backend, not just the stub.

MilleR98:
hey @hectorvent @abanna 
thanks for your review, valuable points 💯

All blocking comments are addressed:
1. InternalFailure -> InternalServerException (blocking)
2. Timeouts now map to ModelTimeoutException/408 instead of falling into the generic 424 catch. Request timeout is also configurable now (proxy.request-timeout-seconds) instead of hardcoded 60s.
3. InvokeModel under proxy now fails fast (ValidationException/400) instead of silently returning a fabricated stub response. Also fixed the streaming endpoints' error messages, which wrongly blamed "the stub" for a limitation that applies to both backends.

hectorvent (approved):
Thanks @MilleR98

PR #1793, opened 2026-07-09
feat(eks): add option to disable k3s's bundled CNI

attaradev:
## What

Adds `eks.disable-cni` (default `false`), which starts the emulated EKS cluster's k3s server with `--flannel-backend=none --disable-network-policy --disable-kube-proxy` so a real CNI (e.g. Cilium) can be installed in its place. When enabled, the container's entrypoint also runs `mount --make-rshared /` before k3s starts, since Cilium's BPF filesystem mount fails otherwise inside the nested container.

## Why

k3s's bundled flannel and kube-proxy run embedded in the k3s server process itself rather than as separate DaemonSets, so there's no way to evict them after the cluster is up — the only way for an external CNI to take over cleanly is for k3s to never start its own. The `/sys` rshare fix mirrors what kind's own node image does in its entrypoint for the same underlying reason (containers default to private mount propagation, unlike real EKS/kubeadm VM nodes).

## Risks

None — the new behavior is opt-in via `disable-cni`, defaulting to `false`, so existing clusters are unaffected.

## Validation

- [x] `EksClusterManagerTest` passes (`serverArgsOmitCniFlagsByDefault`, `serverArgsDisableFlannelAndKubeProxyWhenRequested`, `rshareEntrypointUsesOnlyShBuiltins`, `rshareWrappedCmdPreservesServerArgsAfterThePlaceholder`)
- [x] With `eks.disable-cni=true`, a real cluster starts and `kubectl get nodes` shows Ready once an external CNI (e.g. Cilium) is installed
- [x] With the default `eks.disable-cni=false`, cluster startup and existing CoreDNS/local-path-provisioner/metrics-server behavior are unchanged

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
+++ b/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
@@ -1405,6 +1405,18 @@ public interface EmulatorConfig {
          */
         @WithDefault("true")
         boolean ecrRegistryMirror();
+
+        /**
+         * When true, starts k3s with {@code --flannel-backend=none --disable-network-policy
+         * --disable-kube-proxy} instead of its bundled networking stack. k3s's default flannel CNI
+         * and kube-proxy run embedded in the k3s server process itself (not separate, killable
+         * DaemonSets), so a real CNI (e.g. Cilium) can only cleanly take over if k3s never starts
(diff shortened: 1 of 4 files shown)


hectorvent (approved):
Thanks @attaradev, 

Solid implementation

PR #1794, opened 2026-07-09
feat(cloudformation): support AWS::Events::EventBus and EventBusPolicy

lruizctaima:
## Summary

AWS CDK stacks that define a custom EventBridge event bus did not work against Floci. `AWS::Events::EventBus` was not handled by the CloudFormation provisioner, so it fell through the `default` branch and was only *stubbed* (a fake `arn:aws:stub:::` and a random physical id) — the bus was never created in `EventBridgeService`. As a result `PutEvents` to that bus failed with `EventBus not found`, no rule ever matched, and nothing reached the targets, even though `cdk deploy` reported `CREATE_COMPLETE` and the bus was absent from `ListEventBuses`.

This wires real CloudFormation provisioning for `AWS::Events::EventBus` and `AWS::Events::EventBusPolicy`, reusing the existing `EventBridgeService`. No new endpoints, wire protocol, or configuration — this is CloudFormation resource-type coverage on top of the already-complete EventBridge service.

- **`AWS::Events::EventBus`** — provisions via `EventBridgeService.createEventBus` (`Name`, `Description`, `Tags`, optional inline `Policy`). `Ref` returns the bus name; `Fn::GetAtt` resolves `Arn` and `Name`; re-deploy is idempotent (`ResourceAlreadyExistsException` → `DescribeEventBus`).
- **`AWS::Events::EventBusPolicy`** — both the individual `Action`/`Principal`/`Condition` form and the full `Statement` object form; Statement-form policies are merged by `Sid` so multiple resources on one bus coexist.

It also fixes a latent bug the feature surfaces: `AWS::Events::Rule` resources on a **custom** bus were deleted against the `default` bus (bus argument hardcoded to `null`), so the delete silently no-opped and the rule — and its bus — leaked while the stack recorded `DELETE_COMPLETE`. The rule's `EventBusName` is now stored as a resource attribute at provision time and read back at delete time, mirroring the existing `AWS::EKS::Nodegroup` attribute-aware delete. The `rollbackCreatedResources` path is fixed the same way (it used the type-based delete overload instead of the resource-aware one).

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

This PR adds CloudFormation resource-type coverage; it does not introduce a new AWS wire action, so there is no new SDK/CLI request shape to pin.

**Incorrect behavior fixed:**
- A CDK/CloudFormation stack with a custom `AWS::Events::EventBus` deployed but the bus was never created in the emulator — `ListEventBuses` did not show it and `PutEvents` to it returned `EventBus not found`, so events were never delivered to rule targets. Real CloudFormation creates the bus and CDK stacks route events through it.
- Deleting (teardown or CREATE rollback) a stack with a custom bus + a rule left both resources live while reporting `DELETE_COMPLETE`, because the rule was deleted against the `default` bus. Real CloudFormation removes the rule and the bus.

**Verification:** deployed an EventBus + Rule + SQS-target stack with AWS CDK (`aws-cdk-lib`, via `cdklocal`) against a local Floci — `CREATE_COMPLETE`, no rollback — then `aws events put-events` for a matching event returned `FailedEntryCount 0` and `aws sqs receive-message` returned the event, confirming the bus, rule/target, and routing are all real. `Ref`/`Fn::GetAtt` are exercised via a template `Outputs` block in the integration tests.

## Checklist

Tests: 7 new cases in `CloudFormationIntegrationTest` (RestAssured over the real wire protocol — Query for CloudFormation/SQS, JSON 1.1 for EventBridge): EventBus create + `Ref`/`GetAtt`, delete, end-to-end delivery to SQS, both `EventBusPolicy` forms (individual and two coexisting Statements), custom-bus teardown, and CREATE-rollback cleanup. Full `CloudFormationIntegrationTest` class: 99/99 green.

## Notes for the reviewer

Scope kept deliberately narrow; three things flagged rather than expanded:

- **Statement-form policy merge is a client-side read-modify-write** (`describeEventBus` → splice by `Sid` → write whole policy). It is not atomic, but CloudFormation provisions resources sequentially per stack, so there is no concurrent writer today. It is not routed through `EventBridgeService.putPermission`'s server-side merge because that path rebuilds a statement from scalar `action`/`principal` and cannot ingest a full IAM `Statement` object (`Principal:{AWS:[...]}`, multiple actions, `Resource`, arbitrary `Condition`). The atomic fix, if parallel provisioning ever lands, would be a server-side full-statement merge on `EventBridgeService` (which would also remove the small duplication).
- **`deleteEventBusSafe` swallows-and-logs at `warnv`.** Consistent with the sibling `*Safe` delete helpers (which log at `debugv`), but bumped to `warn` here specifically because a bus left alive while the stack says `DELETE_COMPLETE` is a state divergence worth surfacing.
- **Inline `Policy` and `Tags` on `AWS::Events::EventBus`** are provisioned but only exercised indirectly by the tests (the `Policy` branch reuses the well-covered `putPermission` `policyJson` path; `Tags` are parsed in the create test). Out of scope and unimplemented: `AWS::Events::ApiDestination`, `AWS::Events::Connection`, `AWS::Events::Archive`, `AWS::Scheduler::Schedule` (no underlying service support), and enforcing the bus resource policy at delivery time.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -390,6 +390,14 @@ public class CloudFormationResourceProvisioner {
             }
             return;
         }
+        // Rule deletion needs the rule's event bus (stored as an attribute at provision time),
+        // which the type/physicalId delete path can't provide; a custom-bus rule looked up under
+        // the default bus would silently no-op and leave the rule (and its bus) live.
+        if ("AWS::Events::Rule".equals(resourceType)) {
+            deleteEventBridgeRuleSafe(resource.getPhysicalId(),
+                    resource.getAttributes().get("EventBusName"), region);
+            return;
+        }
         delete(resourceType, resource.getPhysicalId(), region);
     }
 
@@ -417,7 +425,7 @@ public class CloudFormationResourceProvisioner {
             } // KMS keys can't be immediately deleted; skip
             case "AWS::KMS::Alias" -> kmsService.deleteAlias(physicalId, region);
             case "AWS::SecretsManager::Secret" -> deleteSecretSafe(physicalId, region);
-            case "AWS::Events::Rule" -> deleteEventBridgeRuleSafe(physicalId, region);
+            case "AWS::Events::Rule" -> deleteEventBridgeRuleSafe(physicalId, null, region);
             case "AWS::Events::EventBus" -> deleteEventBusSafe(physicalId, region);
             case "AWS::ApiGateway::RestApi" -> apiGatewayService.deleteRestApi(region, physicalId);
             case "AWS::ApiGatewayV2::Api" -> apiGatewayV2Service.deleteApi(region, physicalId);
@@ -2100,6 +2108,9 @@ public class CloudFormationResourceProvisioner {
                 state, description, roleArn, Map.of(), region);
         r.setPhysicalId(ruleName);
         r.getAttributes().put("Arn", rule.getArn());
+        if (busName != null && !busName.isBlank()) {
+            r.getAttributes().put("EventBusName", busName);
+        }
 
         // Provision inline targets
         if (props != null && props.has("Targets")) {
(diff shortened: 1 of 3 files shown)


lruizctaima:
**1. Silent leak when bus still has rules at delete time** — Addressed in 1e019cd5: `deleteEventBusSafe` now logs at `warnv` instead of `debugv`, so a bus left alive while the stack records `DELETE_COMPLETE` (e.g. a rule that hardcodes `EventBusName` with no `Ref`/`DependsOn`, giving no dependency edge to order the deletes) is diagnosable, per the AGENTS.md guideline.

**2. Non-atomic read-modify-write for the Statement-form merge** — Keeping the current approach; the suggested fix doesn't quite apply here. `EventBridgeService.putPermission`'s server-side merge (the `policyJson == null` path) rebuilds the statement from scalar `action`/`principal`/`conditionJson`, so it can't ingest a full IAM `Statement` object (`Principal: {AWS: [...]}`, multiple `Action`s, `Resource`, arbitrary `Condition`) — routing the Statement form through it would drop those fields. That's why the merge lives in the provisioner and pushes the whole document via the `policyJson` branch.

As noted, this is not a live issue: CloudFormation provisions resources sequentially within a stack, so there is no concurrent writer to the same bus policy. The correct atomic fix, should parallel provisioning ever land, would be a new `EventBridgeService` method that merges a full statement by `Sid` server-side (which would also remove the small duplication of merge logic). Deferring that to keep this PR narrow.

lruizctaima:
Hi,

Is there a estimate timeline for merging, or anything still blocking it on your side?

Context on why I'm asking: we're actively depending on this EventBridge-via-CloudFormation support (AWS::Events::EventBus + AWS::Events::EventBusPolicy) for our local/dev environment. As a stopgap we've built our own Floci image from this branch and are running it from our own ECR, but we'd much rather track an official release than maintain a fork image. Knowing the expected merge/release window would help us plan when we can drop the custom image.

hectorvent:
Hi @lruizctaima, 

Thanks for your patience. No timeline, I am the only maintainer looking to this project at the moment.

hectorvent (approved):
Thanks @lruizctaima,

This looks good to me.
Merging...

PR #1796, opened 2026-07-09
feat(resourceexplorer2): Add AWS Resource Explorer 2 with cross-service resource discovery

### LINKED ISSUE #1795: [FEAT] Add Resource Explorer
## Service

Resource Explorer https://docs.aws.amazon.com/ARG/

## API Action / Feature

[ListResources](https://docs.aws.amazon.com/resource-explorer/latest/apireference/API_ListResources.html) is what I'm most keen on.

## AWS Documentation

https://docs.aws.amazon.com/resource-explorer/latest/apireference/Welcome.html

## Why is this needed?

We use resource explorer to inventory an AWS account, and would like to emulate that.

- [x] Yes
- [ ] No

I will immediately submit the PR to give an idea on the proposed solution. I'm happy to work with the maintainers to get the PR to a place/style they are happy with.

MariusVolkhart:
## Summary

Emulates AWS Resource Explorer 2: index and view lifecycle (`CreateIndex`, `UpdateIndexType`, `CreateView`, `AssociateDefaultView`, …), `Search` / `ListResources` with the Resource Explorer query grammar, `ListSupportedResourceTypes`, and tag operations. Seventeen existing services are made discoverable through it: ACM, Amazon MQ, Cognito, DynamoDB, ECR, ElastiCache, IAM, KMS, Lambda, MSK, OpenSearch, Pipes, RDS, S3, SNS, SQS, and Step Functions.

Fixes #1795

### Changes outside of`resourceexplorer2/`

- Each of the seventeen services gains `implements ResourceProvider` and two methods.
- `ResolvedServiceCatalog` registers one new service descriptor.
- `SigV4CredentialScope` is made `public` so `ResourceExplorer2PathRewriteFilter` can read the signing service from the `Authorization` header (see Notes).
- Two services need a little more to report accurate data: ElastiCache's `ReplicationGroup` gains `arn` and `region` fields (and its query handler threads `region` through), and RDS sources its resource tags from the existing `ResourceGroupsTaggingService`. Those are the only touches to existing request paths.

### Making a service discoverable

Implement `core/resource/ResourceProvider`:

```java
public interface ResourceProvider {
    List<ExplorerResource> getResources();
    Set<SupportedResourceType> getSupportedResourceTypes();
}
```

`ResourceExplorer2Service` pulls in every implementation through CDI. `services/lambda/LambdaService` is the shortest example.

### Suggested review approach

1. `core/resource/` — the three-type contract.
2. `services/resourceexplorer2/ResourceExplorer2Service` — index/view lifecycle, aggregation, pagination.
3. `services/resourceexplorer2/query/` — `QueryParser` and `ResourceFilter` implement the filter grammar; `ParsedQuery` is the parsed form.
4. `services/resourceexplorer2/ResourceExplorer2Controller` — wire mapping.
5. One service integration; the other fifteen match it.

### Notes for reviewers

**Path collision with S3 Vectors.** Both are REST-JSON services rooted at `/`, and both declare `/CreateIndex`, `/GetIndex`, `/ListIndexes`, and `/DeleteIndex`. JAX-RS can't route by path. `ResourceExplorer2PathRewriteFilter` disambiguates on the SigV4 credential scope — the signal `AwsQueryController` already routes query services by, via the same `@PreMatching` rewrite that `SqsQueueUrlRouterFilter` and `S3VirtualHostFilter` use. It's the only non-local routing here. The filter comments the failure mode if its path set drifts from `S3VectorsController`.

**No index; resources are gathered per request.** `paginateResources()` calls every provider's `getResources()` on each `Search` / `ListResources`. Providers stay stateless and always match live service state, and an emulator never holds enough resources for the cost to matter. The constraint this puts on a provider: `getResources()` has to be cheap and has to tolerate a resource that's mid-creation, so it skips null ARNs and treats a null tag map as empty.

**Where floci differs from AWS on purpose:**
- An index returns `CREATING` on the create call and is `ACTIVE` right after. Floci is synchronous, so there's no provisioning window to fake.
- A `NextToken` pointing past a result set that shrank returns an empty page instead of an error, which is how AWS handles a replayed token.
- The per-region default view is stored on its own, not inferred from a name, so `DisassociateDefaultView` sticks across a restart.

**Tags only appear when a view's `IncludedProperties` lists `tags`**, same as AWS. Without it, tagged resources come back with no tag data.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Wire protocol verified against AWS Java SDK v2 `2.44.14` (`software.amazon.awssdk:resourceexplorer2`) in `compatibility-tests/sdk-test-java/ResourceExplorer2Test`, which drives the operations through the generated `ResourceExplorer2Client` against a running floci.

## Checklist

- [x] `./mvnw test` passes locally - ran it, struggling to get lambda & Neptune tests working with Colima on MacOS.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/resourceexplorer2/ResourceExplorer2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/resourceexplorer2/ResourceExplorer2Service.java
@@ -1,0 +1,643 @@
+package io.github.hectorvent.floci.services.resourceexplorer2;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.github.hectorvent.floci.core.common.AwsArnUtils;
+import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.core.common.RegionResolver;
+import io.github.hectorvent.floci.core.resource.ExplorerResource;
+import io.github.hectorvent.floci.core.resource.ResourceProvider;
+import io.github.hectorvent.floci.core.resource.SupportedResourceType;
+import io.github.hectorvent.floci.core.storage.StorageBackend;
+import io.github.hectorvent.floci.core.storage.StorageFactory;
+import io.github.hectorvent.floci.services.resourceexplorer2.model.IncludedProperty;
+import io.github.hectorvent.floci.services.resourceexplorer2.model.Index;
+import io.github.hectorvent.floci.services.resourceexplorer2.model.IndexState;
+import io.github.hectorvent.floci.services.resourceexplorer2.model.IndexType;
+import io.github.hectorvent.floci.services.resourceexplorer2.model.SearchFilter;
+import io.github.hectorvent.floci.services.resourceexplorer2.model.Taggable;
+import io.github.hectorvent.floci.services.resourceexplorer2.model.View;
+import io.github.hectorvent.floci.services.resourceexplorer2.query.ParsedQuery;
+import io.github.hectorvent.floci.services.resourceexplorer2.query.QueryParser;
+import io.github.hectorvent.floci.services.resourceexplorer2.query.ResourceFilter;
+import io.quarkus.runtime.StartupEvent;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.Instance;
+import jakarta.inject.Inject;
+import org.jboss.logging.Logger;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import java.util.function.Consumer;
+import java.util.regex.Pattern;
+
+@ApplicationScoped
+public class ResourceExplorer2Service {
+
+    private static final Logger LOG = Logger.getLogger(ResourceExplorer2Service.class);
+
+    private final Instance<ResourceProvider> providers;
+    private final RegionResolver regionResolver;
+    private final ObjectMapper objectMapper;
+    private final StorageBackend<String, Index> indexStore;
+    private final StorageBackend<String, View> viewStore;
+    /** Default view ARN per region. Persisted, not derived, so an explicit choice survives restarts. */
+    private final StorageBackend<String, String> defaultViewStore;
+
+    /**
+     * The maximum number of matching resources {@code Search} returns across all pages. Per the AWS
+     * API reference: "The operation can return only the first 1,000 results." Also the maximum
+     * {@code MaxResults} page size accepted by both operations, and the page size at which
+     * {@code ListResources} suppresses its {@code NextToken}.
+     *
+     * @see <a href="https://docs.aws.amazon.com/resource-explorer/latest/apireference/API_Search.html">AWS API: Search</a>
+     */
+    static final int MAX_RESOURCE_RESULTS = 1000;
+
+    /**
+     * Sentinel result cap for {@code ListResources}, which — unlike {@code Search} — has no total cap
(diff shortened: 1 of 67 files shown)


hectorvent:
Hello @MariusVolkhart,

Thanks for the PR!

Could you please update the commit history to remove the Co-authored-by: Claude <...> line from the commit message? We prefer to keep the attribution focused on human contributors for this project.

Please take a look at the CONSTRIBUTORS.md file

MariusVolkhart:
@hectorvent of course, apologies for missing that. I'm corrected my mistake.

src/main/java/io/github/hectorvent/floci/services/resourceexplorer2/ResourceExplorer2Service.java
```diff
@@ -0,0 +1,643 @@
...
+                LOG.warnf(e, "ResourceProvider %s failed to supply resources; excluding it from this result",
+                        provider.getClass().getSimpleName());
+            }
+        }
+        List<ExplorerResource> filtered = all.stream()
+                .filter(r -> ResourceFilter.matches(r, combined))
+                .toList();
+        int uncappedTotal = filtered.size();
+        PageBounds b = pageBounds(filtered.size(), maxResults, nextToken, resultCap);
+        return new Page(filtered.subList(b.start(), b.end()), b.end(), b.total(), uncappedTotal);
```
abanna, line 492: This is a substantial piece of work and the SPI reads cleanly — thanks for the javadoc on `ResourceProvider`, it makes the extension contract obvious.

One issue with the pagination, though, and I think it's a real one rather than a nit: `NextToken` is a bare offset (`encodeNextToken(p.end())`), but `filtered` is rebuilt from scratch on every call and never sorted. `all` comes from iterating `Instance<ResourceProvider>` and concatenating each provider's `getResources()`, which bottoms out in `store.scan(k -> true)` — and `StorageBackend.scan` only promises "a new mutable list of values whose keys pass the filter", with no ordering guarantee over a map-backed store. `.filter(...).toList()` preserves encounter order, so nothing establishes a total order anywhere on this path.

Offset pagination is only well-defined over a stable order. Concretely: page through `Search` with `MaxResults=100` while anything is being created, and the offset for page 2 indexes into a differently-ordered list — resources get silently skipped or returned twice. The `pageBounds` javadoc anticipates the set *shrinking* and clamps for it, but clamping only prevents an out-of-range `subList`; it doesn't address reordering, which is the part that loses results.

What makes this easy to miss is that `ResourceProvider#getResources()` currently documents the opposite requirement:

> Ordering is not significant — the service applies its own pagination.

That contract is what a provider author will code against, and it's exactly the property offset pagination can't tolerate.

Could `paginateResources` impose a deterministic order before slicing? Sorting by ARN is the natural key — it's already required to be unique per resource, and it's stable across calls regardless of provider or store iteration order:

```java
List<ExplorerResource> filtered = all.stream()
        .filter(r -> ResourceFilter.matches(r, combined))
        .sorted(Comparator.comparing(ExplorerResource::arn))
        .toList();
```

That also lets the `getResources()` javadoc keep saying ordering isn't significant, which then becomes true rather than merely permitted. A test that pages with `MaxResults` smaller than the result set, creates a resource between the two calls, and asserts no ARN is dropped or duplicated across pages would pin it down.

hectorvent:
Thank you for the rebase and for folding the new RUM service into the provider set while you were at it. The RUM integration is careful work: the owner account stays out of the GetAppMonitor body with a test asserting exactly that, and the reconstructed ARNs look right.

**(blocking)** The MaxResults bound from the previous review is still present at this head. ListSupportedResourceTypes allows values up to 1000 per the model ([service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/resource-explorer-2/2022-07-28/service-2.json), shape ListSupportedResourceTypesInputMaxResultsInteger), but the controller validates against 100, so a legal page size like 500 gets a ValidationException. ListViews has the mirror issue, model max 50 while the controller accepts 100, worth fixing in the same touch.

**(follow-up, separate PR)** The /re2/* rewrite targets are still publicly routable non AWS paths; moving them under the internal prefix would keep the public surface AWS shaped.

The branch now merges cleanly and CI is green, so that one bounds fix is the only thing left.

hectorvent:
Closing in favor of #2485

PR #1798, opened 2026-07-10
fix(apigateway): import authorizers and security requirements from OpenAPI

abanna:
## Summary

Imports Lambda, IAM, and Cognito authorizers plus per-operation `security` requirements from
OpenAPI/Swagger definitions used by `ImportRestApi` and `PutRestApi`.

Previously, imported security schemes were ignored and each method was created with
`authorizationType=NONE`, leaving APIs that declared authorizers open at runtime.

Now:

- each `x-amazon-apigateway-authorizer` security scheme creates a `TOKEN`, `REQUEST`, or
  `COGNITO_USER_POOLS` authorizer;
- TOKEN identity sources default from the security scheme's header name and location when the
  extension does not provide one;
- operation-level or root-level OpenAPI `security` configures `CUSTOM`, `AWS_IAM`, or
  `COGNITO_USER_POOLS` authorization and the matching authorizer ID;
- `awsSigv4` schemes map to `AWS_IAM`;
- imported authorizer extensions are validated before `PutRestApi` clears existing API state;
- the extension's required `type` is limited to AWS's supported `token`, `request`, and
  `cognito_user_pools` values;
- malformed extension fields return `BadRequestException` without changing the existing API,
  resources, methods, or authorizers.

The implementation reuses the existing authorizer and method creation paths; it does not add a
custom API surface.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

The import behavior follows API Gateway's
[`x-amazon-apigateway-authorizer`](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authorizer.html),
`x-amazon-apigateway-authtype`, and OpenAPI `security` contracts. Validation covers the required
authorizer type, its supported values, nested string fields, and Cognito `providerARNs` before an
overwrite can mutate an existing API.

## Verification

- `ApiGatewayOpenApiImportTest`: 41 tests passed.
- Broader import, REST API, and request-authorizer set: 106 tests passed.
- Full API Gateway and VTL suite: 529 tests passed.
- Malformed-extension cases verify AWS-style `BadRequestException` responses and atomic
  preservation of existing API state.
- Java compatibility tests compile successfully.
- Documentation generation, tests, and strict site build pass.

## Checklist

- [x] Automated tests cover the correction.
- [x] Existing valid Lambda, IAM, Cognito, identity-source, TTL, multi-authorizer, and overwrite
  imports remain green.
- [x] Commit messages follow Conventional Commits.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
@@ -1653,6 +1653,10 @@ public class ApiGatewayController {
         if (a.getAuthorizerUri() != null) node.put("authorizerUri", a.getAuthorizerUri());
         if (a.getIdentitySource() != null) node.put("identitySource", a.getIdentitySource());
         node.put("authorizerResultTtlInSeconds", Integer.parseInt(a.getAuthorizerResultTtlInSeconds()));
+        if (a.getProviderARNs() != null && !a.getProviderARNs().isEmpty()) {
+            ArrayNode arns = node.putArray("providerARNs");
+            a.getProviderARNs().forEach(arns::add);
+        }
         return node;
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
@@ -37,6 +37,8 @@ import io.github.hectorvent.floci.services.apigateway.model.UsagePlan;
 import io.github.hectorvent.floci.services.apigateway.model.UsagePlanKey;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.PathItem;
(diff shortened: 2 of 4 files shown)


abanna:
Thanks for the review signal around the defensive-cast edge case. Re-vetting found a real `PutRestApi` atomicity defect: malformed authorizer extension values could pass preflight, fail with a cast only after existing API state was cleared, and surface as a 500. Fixed in `0a52849de`: the authorizer extension and every consumed field are now typed before mutation, malformed shapes return `BadRequestException`/400, and apply reuses the same typed accessors. Seven regression cases prove metadata, resources, methods, authorizers, and method-to-authorizer links remain unchanged; `ApiGatewayOpenApiImportTest` is 39/39 and exact-head CI/Greptile is 14/14 green. I also vetted the residual non-header Cognito note: AWS’s supported REST import shape requires an `apiKey` scheme named `Authorization` in the header, which this implementation already handles.

hectorvent (approved):
Thanks @abanna!

PR #1799, opened 2026-07-10
fix(tls): reuse the persisted self-signed certificate across restarts

abanna:
## Summary

`TlsConfigSource` is a MicroProfile `ConfigSource`, so it runs **before** CDI startup and before quarkus-security registers its runtime security providers. BouncyCastle was therefore not yet registered when the source decided whether the persisted self-signed certificate could be reused.

The reuse check calls `isSelfSigned(certFile)` → `CertificateGenerator.parseCertificate(...)`, which failed with `no such provider: BC`. `isSelfSigned()` swallowed that and returned `false`, so the persisted certificate was treated as unusable and **regenerated on every restart**. Any client that had pinned/trusted the previous certificate (e.g. via `NODE_EXTRA_CA_CERTS`) then broke after a restart.

The fix registers BouncyCastle at the top of the constructor — before any certificate parsing — by hoisting the registration that already existed inline in `generateSelfSignedCert()` into a shared `ensureBouncyCastleRegistered()` helper. Both the reuse path and the generation path now call it. No behavioural change other than the certificate now surviving restarts.

## Type of change

## AWS Compatibility

This is internal to Floci's own TLS termination and does not change any AWS wire protocol or API surface. It restores the intended behaviour of the existing self-signed-certificate persistence/reuse feature: a stable certificate across restarts, matching how a real endpoint keeps a stable server certificate so trust anchors remain valid.

Verified functionally by booting the emulator twice against the same persistent volume with `FLOCI_TLS_ENABLED=true`:

```
# Boot 1 (fresh volume)
TLS: generated self-signed certificate: /data/tls/floci-selfsigned.crt
# sha256 = d9f90a28aeb92b8c046032ea7ef3971a0d4136bfc0094297705a4b5a881bb4bf

# Boot 2 (same volume)
TLS: reusing existing self-signed certificate: /data/tls/floci-selfsigned.crt
# sha256 = d9f90a28aeb92b8c046032ea7ef3971a0d4136bfc0094297705a4b5a881bb4bf   (identical, no "no such provider: BC")
```

## Checklist

New test `TlsConfigSourceCertificateReuseTest#testCertificateReusedWhenBouncyCastleNotPreRegistered` reproduces the exact ordering (removes the BC provider, then constructs `TlsConfigSource` with an existing certificate) and asserts the certificate is reused. Negative control confirmed the test fails without the constructor registration (certificate regenerated, different timestamp) and passes with it. All 43 tests in the `TlsConfigSource*` suite pass (`mvn -o -Dtest='TlsConfigSource*Test' test`).


CoreOxide (approved):
Nice addition. LGTM.

hectorvent (approved):
Thanks @abanna

PR #1805, opened 2026-07-10
fix(rds): resolve AWS-managed default DB cluster parameter groups on reference

abanna:
## Summary

AWS provides managed default DB cluster parameter groups for each supported engine family. Floci previously rejected those names unless a user-created group existed, which broke ordinary CreateDBCluster flows emitted by tools such as the AWS CDK.

This change:

- exposes a bounded catalog of current AWS-managed defaults for Aurora MySQL 5.7/8.0/8.4, Aurora PostgreSQL 11-18, Multi-AZ MySQL 8.0/8.4, and Multi-AZ PostgreSQL 13-18;
- resolves those defaults without persisting or mutating them;
- keeps unknown default names and unknown custom names as not found;
- validates the requested engine version against the exact parameter-group family, preventing cross-major matches such as Aurora PostgreSQL 16 with default.aurora-postgresql15;
- rejects attempts to modify or delete AWS-managed default groups with InvalidDBParameterGroupState;
- preserves persisted data without duplicating catalog entries.

The catalog is intentionally static and region-agnostic. AWS availability varies by Region; DescribeDBEngineVersions remains the authoritative regional source.

## Type of change

## AWS compatibility

CreateDBCluster currently supports Aurora MySQL, Aurora PostgreSQL, MySQL, and PostgreSQL cluster engines. PostgreSQL parameter-group families are keyed by major version; MySQL families are keyed by major.minor. MariaDB has no AWS DB-cluster family and is not fabricated here.

## Verification

- RdsServiceTest: 87 passed
- RdsQueryHandlerTest: 46 passed
- Combined focused RDS verification: 133 passed, 0 failures, 0 errors
- Whitespace validation: clean

Coverage includes all 19 catalog entries, filtered and unfiltered listing, persisted override behavior, unsupported defaults, protected managed-default lifecycle operations, and cross-major rejection for Aurora PostgreSQL, Aurora MySQL, PostgreSQL, and MySQL.

## Checklist

- [x] I have performed a self-review of my code
- [x] New and existing focused tests pass locally with my changes


src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
```diff
@@ -829,9 +829,21 @@ public DbClusterParameterGroup createDbClusterParameterGroup(String name, String
...
+            // AWS ships a managed default DB cluster parameter group per engine family that always
+            // exists (e.g. default.aurora-postgresql16), so a cluster that uses the engine default —
+            // and a DescribeDBClusterParameterGroups that looks one up — should resolve it rather than
+            // 404. Materialize the default group on first reference; the family is the segment after
+            // "default." (e.g. aurora-postgresql16).
+            if (name != null && name.startsWith("default.")) {
+                DbClusterParameterGroup dflt = new DbClusterParameterGroup(
+                        name, name.substring("default.".length()), "Default cluster parameter group");
+                clusterParameterGroups.put(name, dflt);
+                return dflt;
```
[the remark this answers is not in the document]
abanna: This is intentional and, I believe, faithful. Materialization is gated to `default.*` names only (arbitrary group names still 404), and AWS ships those managed default cluster parameter groups (e.g. `default.aurora-postgresql16`) as always-present — `DescribeDBClusterParameterGroups` returns them whether or not any cluster exists. So a `default.*` group being present after a rejected `CreateDBCluster` matches what AWS would already show; it isn't a phantom the failed create conjured. Deferring the put would actually diverge from AWS (the default would be absent until a successful create references it).

src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
```diff
@@ -920,19 +941,71 @@ public DbClusterParameterGroup createDbClusterParameterGroup(String name, String
...
     }
 
     public Collection<DbClusterParameterGroup> listDbClusterParameterGroups(String filterName) {
         if (filterName != null && !filterName.isBlank()) {
-            return clusterParameterGroups.get(filterName).map(List::of).orElse(List.of());
+            try {
+                return List.of(getDbClusterParameterGroup(filterName));
+            } catch (AwsException e) {
+                if ("DBClusterParameterGroupNotFound".equals(e.getErrorCode())) {
+                    throw new AwsException("DBParameterGroupNotFound",
```
pgermosen, line 964: **Filter-not-found path drops the cluster-specific message** (minor / verify).

This correctly converts the internal `DBClusterParameterGroupNotFound` to the wire-facing `DBParameterGroupNotFound` code (AWS uses the shared `DBParameterGroupNotFound` fault for cluster param groups), but it also swaps in the generic *non-cluster* message. Its sibling `getDbClusterParameterGroup` (just above) keeps the cluster wording, and real `DescribeDBClusterParameterGroups` returns the cluster-specific message — so the two not-found paths for the same resource end up inconsistent. Suggest keeping the code change but preserving the cluster message:

```suggestion
                    throw new AwsException("DBParameterGroupNotFound",
                            "DBClusterParameterGroupName doesn't refer to an existing DB cluster parameter group.", 404);
```

(Worth a quick check against a real `describe-db-cluster-parameter-groups --db-cluster-parameter-group-name <missing>` if you have creds handy — confident on the code, ~90% on the exact message wording.)

pgermosen (approved):
Good fix for a real compatibility gap — CDK/tooling references default.<family> cluster parameter groups that Floci previously rejected. Clean approach: a bounded static catalog of 19 managed defaults, resolved on read without persisting or mutating, layered under any user-created group.

I verified the parts that matter:

Cross-major rejection is correct — expectedClusterParameterGroupFamily derives the family (major for postgres/aurora-postgresql, major.minor for mysql/aurora-mysql) and expectedFamilyPrefix produces exactly the catalog naming, so default.aurora-postgresql15 with engine version 16 is rejected as claimed.
Managed defaults are protected — create/modify/delete all reject the managed names with the right codes; resolution never writes to storage.
Listing merges correctly — managed first, persisted overriding by name; and since create rejects managed names, a collision can't actually occur.
validateParameterGroupFamily is not dead — still used by the instance-parameter-group path; only the cluster path was inlined.
Nice that DescribeDBClusterParameterGroups with an unknown filter name now throws DBParameterGroupNotFound instead of the old silent empty list — that matches AWS.

There is only a one minor nit inline; non-blocking.

src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
abanna: Thanks for calling this out. I checked the operation-specific AWS model: `DescribeDBClusterParameterGroups` declares `DBParameterGroupNotFoundFault`, whose code is `DBParameterGroupNotFound` and whose documented message is `DBParameterGroupName doesnt refer to an existing DB parameter group.` The separate cluster-specific fault/message is used by other RDS operations, so the helper-to-wire translation here is intentional. The AWS API reference confirms that contract: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBClusterParameterGroups.html

I kept the production behavior and added an exact message assertion in 55d22d594. Verification: `RdsServiceTest` passes 87/87.

src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
```diff
@@ -920,19 +941,71 @@ public DbClusterParameterGroup createDbClusterParameterGroup(String name, String
...
+                "DBClusterParameterGroupName doesn't refer to an existing DB cluster parameter group.", 404);
     }
 
     public Collection<DbClusterParameterGroup> listDbClusterParameterGroups(String filterName) {
         if (filterName != null && !filterName.isBlank()) {
-            return clusterParameterGroups.get(filterName).map(List::of).orElse(List.of());
+            try {
+                return List.of(getDbClusterParameterGroup(filterName));
+            } catch (AwsException e) {
+                if ("DBClusterParameterGroupNotFound".equals(e.getErrorCode())) {
```
pgermosen: **Your own reasoning here extends to two sites that still use the cluster-specific code**

Re-reading your reply from July: *"The separate cluster-specific fault/message is used by other RDS operations, so the helper-to-wire translation here is intentional."* That's exactly right, and I checked which operations those are. Parsing `botocore/data/rds/2014-10-31/service-2.json`, the split is clean and it's by *what the operation acts on*:

| code | declared by |
|---|---|
| `DBClusterParameterGroupNotFound` | `CreateDBCluster`, `ModifyDBCluster`, `RestoreDBClusterFromS3`, `RestoreDBClusterFromSnapshot`, `RestoreDBClusterToPointInTime`, `CreateBlueGreenDeployment` |
| `DBParameterGroupNotFound` | `DescribeDBClusterParameterGroups`, `DeleteDBClusterParameterGroup`, `ModifyDBClusterParameterGroup`, `ResetDBClusterParameterGroup`, `CopyDBClusterParameterGroup` (plus the instance equivalents) |

Cluster-specific only where a *cluster* references a group; plain everywhere the operation acts *on* the group. Your translation here matches that precisely.

**The same principle covers two more sites, which still return the cluster variant:**

- `deleteDbClusterParameterGroup:1010` — `DeleteDBClusterParameterGroup` declares `DBParameterGroupNotFound`.
- `modifyDbClusterParameterGroup:1022`, which reaches `getDbClusterParameterGroup:954` and inherits it — `ModifyDBClusterParameterGroup` likewise.

Both predate this PR, so nothing here regressed. Raising it because you've already done the research that settles them, and a reader who sees this shim will reasonably assume the neighbouring paths were considered.

**On shape, if you do touch it:** `getDbClusterParameterGroup` has four callers and only `:1089` (the `CreateDBCluster` family validation) is a context where the cluster variant is correct. The helper's default currently serves that single minority case, which is why the majority needs catch-and-rethrow. Flipping it — helper throws the plain code, `:1089` translates up — removes the shim entirely. Equally fine to just correct `:1010` and leave a note at `:954`.

PR #1808, opened 2026-07-10
feat(iam): add ListEntitiesForPolicy query action

abanna:
## Summary

- implement the IAM Query API `ListEntitiesForPolicy` action
- return directly attached groups, users, and roles with their stable entity IDs
- return `NoSuchEntity` for an unknown managed-policy ARN
- validate the response through the AWS SDK for Java
- document the supported behavior and current limitations

Refs #1938

## AWS Compatibility

The response uses the existing IAM Query protocol path and returns
`ListEntitiesForPolicyResult` with `PolicyGroups`, `PolicyUsers`, `PolicyRoles`, and
`IsTruncated`. Each member includes the AWS name and ID fields expected by SDK clients.

This matches the supported path in the
[AWS IAM API reference](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListEntitiesForPolicy.html).
Current limitations are explicit: `EntityFilter`, `PathPrefix`, `PolicyUsageFilter`, and
pagination are not yet applied. The action therefore reports direct permissions-policy
attachments and returns `IsTruncated=false`.

## Validation

- `IamListEntitiesForPolicyIntegrationTest`: 2 tests passed
- broader `Iam*Test` suite: 187 tests passed
- AWS SDK for Java `IamTest`: 33 tests passed
- documentation tests: 20 tests passed
- strict action-table generation and MkDocs build passed

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/IamTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/IamTest.java
@@ -45,6 +45,7 @@ import software.amazon.awssdk.services.iam.model.ListAttachedRolePoliciesRequest
 import software.amazon.awssdk.services.iam.model.ListAttachedRolePoliciesResponse;
 import software.amazon.awssdk.services.iam.model.ListAttachedUserPoliciesRequest;
 import software.amazon.awssdk.services.iam.model.ListAttachedUserPoliciesResponse;
+import software.amazon.awssdk.services.iam.model.ListEntitiesForPolicyResponse;
 import software.amazon.awssdk.services.iam.model.ListGroupsForUserRequest;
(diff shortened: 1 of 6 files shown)


pgermosen (approved):
Looks good.

Clean new read action, correctly wired through the standard IAM query path (allowlist → IamQueryHandler → IamService). It returns the roles/users/groups a managed policy is attached to — the read a caller needs before DeletePolicy will succeed.

The documented simplifications (EntityFilter/PathPrefix/PolicyUsageFilter ignored → all three lists always returned; no pagination → IsTruncated always false) are reasonable for the stated detach-before-delete use case and clearly disclosed in the PR. A caller passing EntityFilter=Role would get the user/group lists populated too, but that's a documented deviation, not a correctness bug for the primary path.

hectorvent (approved):
Thanks @abanna for the PR. 

Thanks @pgermosen for the review.

PR #1814, opened 2026-07-10
feat(eventbridge): deliver events to event-bus targets

tgibson22:
## Summary

`EventBridgeInvoker` supported Lambda, SQS, SNS, Batch, and Firehose targets but silently
dropped events aimed at another event bus with an unsupported-ARN warning. This adds the
`:event-bus/` branch: the event is republished through the existing
`EventBridgeService.putEvents`, so the target bus's own rules evaluate it and fan out
normally — the common two-tier topology where a shared bus forwards domain-scoped events to
domain-specific buses.

- `Source`, `DetailType`, `Resources`, and the originating `account`/`region` carry over from
  the original envelope. The forwarded event currently gets a new event id; whether AWS
  preserves the original is unconfirmed and worth a follow-up.
- `Input`/`InputPath`/`InputTransformer` shape only `Detail`, like AWS event-bus targets. A
  forwarded `Detail` that is not a JSON **object** is dropped with a warning — whether it came
  from an input override or from the incoming envelope — since AWS emits an event only when
  `detail` is an object.
- An event is forwarded between buses only once, matching AWS: a bus that received an event from another bus does not forward it on to a third.
- A bus ARN naming another account is forwarded under that account, so the target bus resolves
  and its rules evaluate there.

## Behaviour notes

Three consequences of the `Detail` and account handling, called out so they are not mistaken
for regressions:

- **Cross-account delivery is fixed for the bus hop only.** Downstream delivery from the target
  bus follows each target type's existing account behaviour: SQS sinks resolve cross-account
  (the queue URL carries the account), while Lambda, SNS, Batch and Firehose targets still
  resolve in the caller's account. A cross-account bus whose rules point at those types will
  warn-and-drop at the second hop.
- **An `InputPath` that selects a scalar is now dropped.** `$.detail.orderId` yields `"o-1"`,
  which is not an object, so it no longer reaches the target bus. This matches AWS, which
  rejects a non-object `Detail`, but it is a behaviour change for that configuration.
- **An envelope carrying `"detail": null` is now dropped** rather than forwarded, for the same
  reason.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

The target ARN is passed as the PutEvents entry's `EventBusName`, which accepts a full
event-bus ARN per the documented PutEvents contract and goes through the existing resolver's
validation. Forwarded events keep the originating account/region in the envelope rather than
being re-stamped by the forwarding hop. The direct PutEvents path still accepts a non-object
`Detail`; that pre-existing gap is unchanged here, though such an event is now dropped rather
than forwarded if it reaches an event-bus target. Semantics are derived from the documented
PutEvents/target behavior, not verified against live AWS. No wire-shape changes.

## Checklist

- [x] `./mvnw test`: all eventbridge tests green (200 under the `EventBridge*` selector). The
      102 failures elsewhere (apigateway/apigatewayv2 WebSocket + authorizer, cloudformation,
      docdb, ecr, elbv2, lambda) reproduce identically on the pre-change branch — 61 failures
      / 41 errors in both — and are environment-only, from running under Docker-in-Docker.
- [x] New or updated integration test added (two-bus topology: source-bus rule → second bus by
      ARN → SQS sink; plus a two-account topology asserting the delivered envelope keeps the
      originating account; a three-bus chain asserting forwarding stops after one hop; plus 13
      invoker unit tests)

### DIFF
--- a/docs/services/eventbridge.md
+++ b/docs/services/eventbridge.md
@@ -98,3 +98,26 @@ aws events put-events \
   --entries '[{"Source":"myapp","DetailType":"test","Detail":"{}"}]' \
   --endpoint-url $AWS_ENDPOINT_URL
 ```
+
+## Event Bus Targets
+
+A rule can target another event bus by ARN: the event is republished there, and that bus's own rules evaluate it and fan out normally. `Source`, `DetailType`, `Resources` and the originating `account`/`region` carry over, and each hop gets a new event id.
+
+```bash
+aws events put-targets \
+  --rule order-placed-rule \
+  --event-bus-name my-bus \
+  --targets '[{
+    "Id": "forward-to-domain-bus",
+    "Arn": "arn:aws:events:us-east-1:000000000000:event-bus/domain-bus"
+  }]' \
+  --endpoint-url $AWS_ENDPOINT_URL
+```
+
+## Current Behavior
+
+- `PutEvents` reports success once the source bus accepts an event, so target delivery failures surface only as a `WARN` in the Floci logs.
+- A `Detail` forwarded to an event bus must be a JSON object, as in AWS; anything else is dropped, including an `InputPath` selecting a scalar such as `$.detail.orderId` or an envelope carrying `"detail": null`.
+- A bus ARN naming another account is forwarded under that account, so the target bus and its rules resolve there.
+- Onward delivery from that bus follows each target type: SQS resolves cross-account, while Lambda, SNS, Batch and Firehose resolve in the caller's account.
+- Bus-to-bus forwarding stops after 5 hops, and the sixth is dropped with only a `WARN` rather than reported to the caller — a Floci limit with no AWS equivalent, present to break cycles.
--- a/docs/services/eventbridge.md
+++ b/docs/services/eventbridge.md
@@ -120,4 +120,4 @@ aws events put-targets \
 - A `Detail` forwarded to an event bus must be a JSON object, as in AWS; anything else is dropped, including an `InputPath` selecting a scalar such as `$.detail.orderId` or an envelope carrying `"detail": null`.
 - A bus ARN naming another account is forwarded under that account, so the target bus and its rules resolve there.
 - Onward delivery from that bus follows each target type: SQS resolves cross-account, while Lambda, SNS, Batch and Firehose resolve in the caller's account.
-- Bus-to-bus forwarding stops after 5 hops, and the sixth is dropped with only a `WARN` rather than reported to the caller — a Floci limit with no AWS equivalent, present to break cycles.
+- An event is forwarded between buses only once, matching AWS: a bus that received an event from another bus does not forward it on to a third. The second hop is dropped with only a `WARN` rather than reported to the caller.
--- a/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
+++ b/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
@@ -18,7 +18,9 @@ import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 @ApplicationScoped
@@ -26,11 +28,15 @@ public class EventBridgeInvoker {
 
     private static final Logger LOG = Logger.getLogger(EventBridgeInvoker.class);
 
+    private static final int MAX_BUS_TO_BUS_DEPTH = 5;
+    private static final ThreadLocal<Integer> BUS_TO_BUS_DEPTH = ThreadLocal.withInitial(() -> 0);
+
     private final LambdaService lambdaService;
     private final SqsService sqsService;
     private final SnsService snsService;
     private final BatchService batchService;
     private final FirehoseService firehoseService;
+    private final EventBridgeService eventBridgeService;
     private final ObjectMapper objectMapper;
     private final String baseUrl;
 
@@ -40,6 +46,7 @@ public class EventBridgeInvoker {
                               SnsService snsService,
                               BatchService batchService,
                               FirehoseService firehoseService,
+                              EventBridgeService eventBridgeService,
                               ObjectMapper objectMapper,
                               EmulatorConfig config) {
         this.lambdaService = lambdaService;
@@ -47,6 +54,7 @@ public class EventBridgeInvoker {
         this.snsService = snsService;
         this.batchService = batchService;
         this.firehoseService = firehoseService;
+        this.eventBridgeService = eventBridgeService;
         this.objectMapper = objectMapper;
         this.baseUrl = config.baseUrl();
     }
@@ -56,7 +64,7 @@ public class EventBridgeInvoker {
                        SnsService snsService,
                        ObjectMapper objectMapper,
                        EmulatorConfig config) {
-        this(lambdaService, sqsService, snsService, null, null, objectMapper, config);
+        this(lambdaService, sqsService, snsService, null, null, null, objectMapper, config);
     }
 
     public void invokeTarget(Target target, String eventJson, String region) {
@@ -114,6 +122,65 @@ public class EventBridgeInvoker {
                 // without appending a newline; the delivery-side NDJSON flush handles separation.
                 firehoseService.putRecord(streamName, new Record(payload.getBytes(StandardCharsets.UTF_8)));
                 LOG.debugv("EventBridge delivered to Firehose: {0}", arn);
+            } else if (arn.contains(":events:") && arn.contains(":event-bus/")) {
+                if (eventBridgeService == null) {
+                    LOG.warnv("EventBridge event-bus target missing EventBridge service: {0}", arn);
+                    return;
+                }
+                // Fixed ceiling breaks bus-to-bus cycles; relies on putEvents delivering targets synchronously.
+                int depth = BUS_TO_BUS_DEPTH.get();
+                if (depth >= MAX_BUS_TO_BUS_DEPTH) {
+                    LOG.warnv("EventBridge bus-to-bus depth {0} exceeded at target {1}; dropping", depth, arn);
+                    return;
+                }
+                String targetRegion = extractRegionFromArn(arn, region);
+                // Input overrides shape only Detail; the rest of the entry comes from the original
+                // event envelope, matching AWS event-bus target semantics.
+                JsonNode envelope = objectMapper.readTree(eventJson);
+                boolean inputOverridden = target.getInput() != null
+                        || target.getInputPath() != null
+                        || target.getInputTransformer() != null;
+                String detailBody;
+                if (inputOverridden) {
+                    try {
+                        objectMapper.readTree(payload);
+                    } catch (Exception e) {
+                        LOG.warnv("EventBridge event-bus target {0} requires JSON Detail; dropping non-JSON input: {1}",
+                                arn, e.getMessage());
+                        return;
+                    }
+                    detailBody = payload;
+                } else {
+                    JsonNode detail = envelope.get("detail");
+                    detailBody = detail != null ? detail.toString() : "{}";
+                }
+                Map<String, Object> entry = new HashMap<>();
+                // putEvents accepts a full event-bus ARN as EventBusName and validates it.
+                entry.put("EventBusName", arn);
+                entry.put("Source", envelope.path("source").asText(""));
+                entry.put("DetailType", envelope.path("detail-type").asText(""));
+                entry.put("Detail", detailBody);
+                // AWS keeps the originating account/region; blank falls back inside putEvents.
+                entry.put("Region", envelope.path("region").asText(""));
+                entry.put("Account", envelope.path("account").asText(""));
+                if (envelope.has("resources")) {
+                    entry.put("Resources", envelope.get("resources"));
+                }
+                BUS_TO_BUS_DEPTH.set(depth + 1);
+                try {
+                    var result = eventBridgeService.putEvents(List.of(entry), targetRegion);
+                    if (result.failedCount() > 0) {
+                        LOG.warnv("EventBridge event-bus target {0} rejected event: {1}", arn, result.entries());
+                    } else {
+                        LOG.debugv("EventBridge delivered to EventBus: {0}", arn);
+                    }
+                } finally {
+                    if (depth == 0) {
+                        BUS_TO_BUS_DEPTH.remove();
+                    } else {
+                        BUS_TO_BUS_DEPTH.set(depth);
+                    }
+                }
             } else {
                 LOG.warnv("EventBridge: unsupported target ARN type: {0}", arn);
             }
--- a/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
+++ b/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
@@ -163,7 +163,7 @@ public class EventBridgeInvoker {
                 // AWS keeps the originating account/region; blank falls back inside putEvents.
                 entry.put("Region", envelope.path("region").asText(""));
                 entry.put("Account", envelope.path("account").asText(""));
-                if (envelope.has("resources")) {
+                if (envelope.hasNonNull("resources") && envelope.get("resources").isArray()) {
                     entry.put("Resources", envelope.get("resources"));
                 }
                 BUS_TO_BUS_DEPTH.set(depth + 1);
--- a/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
+++ b/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsArnUtils;
+import io.github.hectorvent.floci.core.common.RegionResolver;
 import io.github.hectorvent.floci.services.batch.BatchService;
 import io.github.hectorvent.floci.services.eventbridge.model.InputTransformer;
 import io.github.hectorvent.floci.services.eventbridge.model.Target;
@@ -37,6 +38,7 @@ public class EventBridgeInvoker {
     private final BatchService batchService;
     private final FirehoseService firehoseService;
     private final EventBridgeService eventBridgeService;
+    private final RegionResolver regionResolver;
     private final ObjectMapper objectMapper;
     private final String baseUrl;
 
@@ -47,6 +49,7 @@ public class EventBridgeInvoker {
                               BatchService batchService,
                               FirehoseService firehoseService,
                               EventBridgeService eventBridgeService,
+                              RegionResolver regionResolver,
                               ObjectMapper objectMapper,
                               EmulatorConfig config) {
         this.lambdaService = lambdaService;
@@ -55,6 +58,7 @@ public class EventBridgeInvoker {
         this.batchService = batchService;
         this.firehoseService = firehoseService;
         this.eventBridgeService = eventBridgeService;
+        this.regionResolver = regionResolver;
         this.objectMapper = objectMapper;
         this.baseUrl = config.baseUrl();
     }
@@ -64,7 +68,9 @@ public class EventBridgeInvoker {
                        SnsService snsService,
                        ObjectMapper objectMapper,
                        EmulatorConfig config) {
-        this(lambdaService, sqsService, snsService, null, null, null, objectMapper, config);
+        this(lambdaService, sqsService, snsService,
+                null /* batch */, null /* firehose */, null /* eventBridge */, null /* regionResolver */,
+                objectMapper, config);
     }
 
     public void invokeTarget(Target target, String eventJson, String region) {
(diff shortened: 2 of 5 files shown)


src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
```diff
@@ -114,6 +122,65 @@ public void invokeTarget(Target target, String eventJson, String region) {
...
+                String detailBody;
+                if (inputOverridden) {
+                    try {
+                        objectMapper.readTree(payload);
+                    } catch (Exception e) {
+                        LOG.warnv("EventBridge event-bus target {0} requires JSON Detail; dropping non-JSON input: {1}",
+                                arn, e.getMessage());
+                        return;
+                    }
+                    detailBody = payload;
```
abanna: Thanks for building this out — routing a shared bus into a domain bus is a genuinely useful topology to have working, and the two-bus integration test is a good shape for it.

This guard doesn't quite enforce what its warning says. `readTree` succeeds for any well-formed JSON *value*, not just an object, so `"hello"`, `[1,2]` and `123` all pass and become `detailBody`. That entry then reaches `EventBridgeService.buildEventEnvelope`, which does `node.set("detail", objectMapper.readTree(detail))` — so bus B's subscribers get an envelope with `"detail": "hello"`.

Two things follow from that. `putEvents` reports success, so nothing is dropped and nothing is logged: this is a silent malformation rather than the clean warn-and-drop the branch is going for. And real EventBridge never emits an event whose `detail` isn't an object — PutEvents rejects a non-object `Detail` with a failed entry — so a rule on bus B matching `{"detail": {...}}` behaves differently here than against AWS.

Would tightening the check to object-ness cover the intent?

```java
JsonNode overridden;
try {
    overridden = objectMapper.readTree(payload);
} catch (Exception e) { ... }
if (!overridden.isObject()) {
    LOG.warnv("EventBridge event-bus target {0} requires a JSON object Detail; dropping: {1}", arn, payload);
    return;
}
detailBody = payload;
```

A target with `Input: "\"hello\""` would make a compact regression case, since it's currently the shortest path to a scalar `detail` in the envelope.

Separately, on the entry just below: `putEvents(List.of(entry), targetRegion)` goes through the two-arg overload, which passes `accountId = null`, so the target bus is resolved in the *caller's* account. A cross-account event-bus ARN — the main reason people use bus targets in the first place — will therefore fall into the `EventBus not found` branch and surface only as a warning. Worth either threading the ARN's account through or noting the limitation explicitly, so it doesn't read as a delivery bug later.

tgibson22: Both fixed in 49b133ed. Thanks — each was a real defect, and the second was silently dropping
events.

**Object-ness check** (`EventBridgeInvoker.java:166`). One deviation from your sketch: the check
sits after both arms, not just the input-override one. The pass-through arm
(`envelope.get("detail")`) had the same hole and needs no target input to hit it, since the
PutEvents handler doesn't validate `Detail`'s shape. Regression cases: your
`Input: "\"hello\""`, plus `[1,2]`, `123`, and the no-override scalar case.

Worth naming as a behaviour change rather than a pure fix: an `InputPath` selecting a scalar
(`$.detail.orderId`) now drops rather than forwards, as does an envelope carrying
`"detail": null`. Both match AWS, but they change what an existing target configured that way
sees.

**Cross-account** — threaded rather than documented (`EventBridgeService.java:644`,
`EventBridgeInvoker.java:188`). Checking your point also turned up a wrong claim in my own
description: it said this was "consistent with how the other target types treat the ARN's
account id", which isn't true — `SqsService.getQueueByUrl` already derives the account from the
queue URL. Corrected.

**The fix covers the bus hop only.** The target bus's own targets go through the same
`invokeTarget`, which threads no account — so SQS sinks resolve cross-account, but Lambda, SNS,
Batch and Firehose still resolve in the caller's. Threading it through every target type is
bigger than this PR, so the description says exactly that rather than claiming general support.

Covered end-to-end by a two-account topology in `EventBridgeBusToBusIntegrationTest`. I checked
it fails without the fix: neutering the account threading fails exactly that test and nothing
else.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
```diff
@@ -18,19 +18,25 @@
...
+import java.util.List;
 import java.util.Map;
 
 @ApplicationScoped
 public class EventBridgeInvoker {
 
     private static final Logger LOG = Logger.getLogger(EventBridgeInvoker.class);
 
+    private static final int MAX_BUS_TO_BUS_DEPTH = 5;
+    private static final ThreadLocal<Integer> BUS_TO_BUS_DEPTH = ThreadLocal.withInitial(() -> 0);
```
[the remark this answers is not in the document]
tgibson22: Leaving this as-is, with reasoning rather than a silent dismissal.

The guard is correct under the current dispatch: `EventBridgeService.putEvents` invokes targets
synchronously in-loop, so the forwarded hop runs on the same thread and sees the incremented
depth. The coupling is real though, and it is exactly as fragile as you describe — if target
dispatch ever moves to a worker pool or parallel stream, the counter silently resets and a
two-bus cycle would loop unbounded.

That assumption is documented at the guard itself rather than left implicit, so a future change
to async dispatch should trip over it. Propagating depth through the call — as a parameter or a
`RequestScoped` counter — means changing `invokeTarget`'s signature for every target type, which
is broader than what this PR is doing.

Worth noting your P1 on the `resources` `NullNode` cast was valid and is already fixed in
148318c6.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
```diff
@@ -114,6 +128,82 @@ public void invokeTarget(Target target, String eventJson, String region) {
                 // without appending a newline; the delivery-side NDJSON flush handles separation.
                 firehoseService.putRecord(streamName, new Record(payload.getBytes(StandardCharsets.UTF_8)));
                 LOG.debugv("EventBridge delivered to Firehose: {0}", arn);
+            } else if (arn.contains(":events:") && arn.contains(":event-bus/")) {
```
pgermosen, line 132: **The behaviour notes exist only in the PR description, and they're the ones users will hit**

The "Behaviour notes" section is the best part of this PR — and as far as I can tell it's the one thing from the abanna round that landed in prose rather than in code. The problem is purely location: after merge, a PR description is effectively unfindable.

The two that will actually reach a user:

- an `InputPath` selecting a scalar (`$.detail.orderId`) now **silently stops delivering** to a bus target;
- an envelope with `"detail": null` likewise.

Both surface as a `LOG.warnv` and nothing else. Someone whose forwarding quietly stopped goes to `docs/services/eventbridge.md` — where there's no mention of event-bus targets at all, and no place for one: that file has only `Supported Actions`, `Configuration`, `Examples` and `Default Event Bus`.

Compare `docs/services/ses.md:154`, which has a `## Current Behavior` section carrying exactly this kind of note (*"Identity policies are stored and returned as metadata ... but Floci does not evaluate policy authorization"*).

**Suggested:** add a `## Current Behavior` section to `docs/services/eventbridge.md` with the three notes you've already written, plus the hop ceiling below. The prose exists — it just needs to live somewhere that survives the merge.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
```diff
@@ -18,19 +19,26 @@
...
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 @ApplicationScoped
 public class EventBridgeInvoker {
 
     private static final Logger LOG = Logger.getLogger(EventBridgeInvoker.class);
 
+    private static final int MAX_BUS_TO_BUS_DEPTH = 5;
```
pgermosen: **The 5-hop ceiling is a Floci invention that fails silently**

Separate from the cross-thread question below — this is about the limit itself, which I don't think anyone has raised.

Cutting cycles is clearly right; without it a two-bus loop is a `StackOverflowError`. No objection to the mechanism. Two things worth being deliberate about:

- **AWS has no documented equivalent**, so this isn't emulating anything — it's a Floci safety limit. A legitimate 6-deep fan-out topology gets dropped at the last hop.
- **The only signal is a `WARN`.** Unlike a `failedCount()` from `putEvents`, nothing reaches the caller, so a topology that quietly exceeds the ceiling looks identical to one that works.

Five is almost certainly generous enough — I can't picture a real bus chain that deep. The concern isn't the number, it's that neither the number nor its existence is discoverable.

**Suggested:** mention it in the docs note above. Making the constant configurable would be over-engineering for something this unlikely to bind.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
```diff
@@ -114,6 +128,82 @@ public void invokeTarget(Target target, String eventJson, String region) {
                 // without appending a newline; the delivery-side NDJSON flush handles separation.
                 firehoseService.putRecord(streamName, new Record(payload.getBytes(StandardCharsets.UTF_8)));
                 LOG.debugv("EventBridge delivered to Firehose: {0}", arn);
+            } else if (arn.contains(":events:") && arn.contains(":event-bus/")) {
+                if (eventBridgeService == null) {
+                    LOG.warnv("EventBridge event-bus target missing EventBridge service: {0}", arn);
+                    return;
+                }
+                // Fixed ceiling breaks bus-to-bus cycles; relies on putEvents delivering targets synchronously.
+                int depth = BUS_TO_BUS_DEPTH.get();
```
pgermosen, line 138: **A second, smaller `ThreadLocal` point — not a reopening of the cross-thread one** *(nit)*

To be clear up front: I'm not re-litigating greptile's P2. I checked that independently and agree with your decline — see my review comment.

This is a different mechanism. `ThreadLocal.withInitial(...).get()` doesn't just read; it calls `setInitialValue()`, which writes the `0` into the thread's map. Every `return` between this line and the `try` (service null, depth exceeded, non-JSON input, non-object `Detail`) leaves that entry behind on a pooled request thread.

Functionally harmless: the retained value is always `0`, which is the correct starting depth, so a later request on the same thread reads exactly what it would have anyway, and it's one boxed `Integer`.

**Suggested:** if you touch this again, `BUS_TO_BUS_DEPTH.remove()` on the drop paths, or drop `withInitial` and treat `null` as zero, which sidesteps it entirely. Not worth a commit on its own.

The restore logic itself is right, for what it's worth: `remove()` at the outermost frame so pooled threads don't retain it, `set(depth)` otherwise, and the `set(depth + 1)` sits after every early return so no path leaves an inflated depth behind.

pgermosen (approved):
Looks good

I'm arriving after three rounds, so I've tried to add only what's new rather than re-run ground that greptile and @abanna already covered. Two things first, since they're what a maintainer needs to act on.

### Adjudicating the one declined finding

Greptile's P2 (ThreadLocal bypassed under cross-thread dispatch) is the only open disagreement, so I checked it independently rather than taking either side's word.

**Your decline is justified.** `putEvents:692` calls `invoker.invokeTarget(...)` in a plain `for` loop, and there is no `Executor`, `CompletableFuture`, parallel stream or thread anywhere in `EventBridgeService`. The forwarded hop provably runs on the same thread today, so the guard holds.

I'd also endorse the reasoning for not fixing it pre-emptively: propagating depth explicitly means changing `invokeTarget`'s signature for all six target types, which is a bigger change than the feature. And the assumption is written at the guard, which is the right place for it.

The one thing I'd push back on gently is "a future change to async dispatch should trip over it" — a comment doesn't trip anything, and the failure mode is an unbounded loop rather than a test failure. That's an argument for the docs note in my first inline comment, not for reworking the code.

### Confirming the two fixes

- **greptile's P1 (`resources` `NullNode` cast), fixed in `148318c6`** — confirmed, and the fix is broader than the report. `hasNonNull("resources") && isArray()` covers both the `NullNode` case greptile found *and* a non-array value it didn't. That matters because `putEvents:951` does a hard `(ArrayNode) entry.getOrDefault(...)` cast, so anything but an array is a `ClassCastException` swallowed by the outer catch.
- **@abanna's object-ness finding, fixed in `49b133ed`** — confirmed, and placing the check after both arms rather than only the input-override arm is right. AWS backs it: `PutEventsRequestEntry.Detail` is *"A valid JSON object. There is no other schema imposed."* So the two behaviour changes you flag are AWS-correct, not a Floci choice — which is what makes them defensible rather than merely deliberate.

### Also checked

- **`EventBusName` accepting a full ARN is documented**, not inferred: *"The name or ARN of the event bus to receive the event"*, and the pattern explicitly permits the `arn:...:event-bus/` prefix. `resolvedBusName:1010` really does parse it, throwing `ValidationException` on a malformed one — and `invokeTarget:135` catches per target, so one bad bus ARN can't abort delivery to sibling targets of the same rule.
- **The "blank falls back inside putEvents" comment is accurate** — `putEvents:657-662` substitutes resolver defaults for blank `Region`/`Account`.
- **Still merges cleanly.** The branch is **131 commits behind `main`**, which for a PR opened in July is where I expected trouble; `git merge-tree` reports no conflicts.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
tgibson22: Added in 16d77d98 — `docs/services/eventbridge.md` now has `## Event Bus Targets` and
`## Current Behavior`, modelled on the `ses.md` section you pointed at. All three notes plus the
ceiling are there.

You were right about "should trip over it" too — a comment doesn't enforce anything. The docs
note is the durable version.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
tgibson22: Documented in the new `## Current Behavior` section — the 5-hop limit, that it's a Floci
invention with no AWS equivalent, and that the sixth hop is dropped with only a `WARN` rather
than reported to the caller.

Agreed on leaving the constant hard-coded.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
tgibson22: Correct — I hadn't registered that `get()` writes through `setInitialValue()` rather than just
reading. Leaving it as you suggest, not worth its own commit; noted for next time this area is
touched.

Three paths rather than four, though: the service-null return at `EventBridgeInvoker.java:132`
precedes the `get()` at 137, so it can't leave an entry behind.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
pgermosen: Double-checked against AWS docs rather than assuming this is just an internal safety net: real EventBridge caps bus-to-bus forwarding at exactly **one** hop, not five. From [Sending and receiving events between AWS accounts](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-cross-account.html): "If you have three event buses in the same account, and set up a rule on the first event bus to forward events from the second event bus to a third event bus, those events are not sent to the third event bus." AWS's Compute Blog on bus-to-bus routing says the same thing directly: "You can only forward an event between same-Region buses once." So `MAX_BUS_TO_BUS_DEPTH = 5` is materially more permissive than AWS, not just a differently-tuned version of the same cap — worth either dropping to 1 hop for fidelity, or keeping 5 but fixing the `docs/services/eventbridge.md` line that currently claims this has "no AWS equivalent" (AWS's equivalent is stricter, not absent). Also currently untested: no test covers a 2+ hop chain, hop 6 being dropped, or an actual A→B→A cycle.

docs/services/eventbridge.md
```diff
@@ -98,3 +98,26 @@ aws events put-events \
...
+  --endpoint-url $AWS_ENDPOINT_URL
+```
+
+## Current Behavior
+
+- `PutEvents` reports success once the source bus accepts an event, so target delivery failures surface only as a `WARN` in the Floci logs.
+- A `Detail` forwarded to an event bus must be a JSON object, as in AWS; anything else is dropped, including an `InputPath` selecting a scalar such as `$.detail.orderId` or an envelope carrying `"detail": null`.
+- A bus ARN naming another account is forwarded under that account, so the target bus and its rules resolve there.
+- Onward delivery from that bus follows each target type: SQS resolves cross-account, while Lambda, SNS, Batch and Firehose resolve in the caller's account.
+- Bus-to-bus forwarding stops after 5 hops, and the sixth is dropped with only a `WARN` rather than reported to the caller — a Floci limit with no AWS equivalent, present to break cycles.
```
pgermosen: "a Floci limit with no AWS equivalent" isn't accurate — AWS does have an equivalent, and it's stricter: bus-to-bus forwarding is capped at exactly one hop on real AWS (see the inline comment on `EventBridgeInvoker.java`'s `MAX_BUS_TO_BUS_DEPTH` for the two docs citations). Suggest rewording to something like "a Floci limit, more permissive than AWS's documented one-hop cap, present to break cycles at a depth AWS itself would never reach."

pgermosen (approved):
Independently checked the AWS-behavior claims against docs.aws.amazon.com and an AWS Compute Blog article rather than trusting the PR's own "Semantics are derived from the documented PutEvents/target behavior, not verified against live AWS" self-assessment — and found one real, material gap the self-assessment missed.

**Real AWS caps bus-to-bus forwarding at exactly one hop; this implementation allows five, and its own docs mischaracterize that as having "no AWS equivalent."** From the official EventBridge docs ([Sending and receiving events between AWS accounts](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-cross-account.html)): *"If a receiver account sets up a rule that sends events received from a sender account on to a third account, these events are not sent to the third account. If you have three event buses in the same account, and set up a rule on the first event bus to forward events from the second event bus to a third event bus, those events are not sent to the third event bus."* The AWS Compute Blog's dedicated bus-to-bus routing article confirms the same for same-account routing: *"You can only forward an event between same-Region buses once. If you have routed an event from a source bus to a target bus, the target bus cannot then forward the event to another bus."* That's a hard, deliberate one-hop architectural limit — not a missing feature AWS forgot to cap, which is what "a Floci limit with no AWS equivalent" (the new `docs/services/eventbridge.md` line) implies. This PR's `MAX_BUS_TO_BUS_DEPTH = 5` lets a chain forward up to 5 buses deep, which is *more* permissive than real AWS, not just differently-capped — a topology that AWS would silently stop at hop 2 (dropping the event with no error, per the same doc: "these events are not sent") would keep flowing all the way to hop 5 here. Worth either matching AWS's one-hop cap for fidelity, or — if the extra permissiveness is intentional for the emulator's usefulness — correcting the docs line to say so explicitly instead of claiming no AWS equivalent exists.
- Separately, the PR states "each hop gets a new event id"; the same Compute Blog article says *"the delivered event is identical to the original event and does not contain any additional metadata or attributes"* when forwarded bus-to-bus, which reads as including the event ID being preserved rather than regenerated. Lower-confidence than the hop-limit finding since I couldn't find a docs.aws.amazon.com page stating the ID point as explicitly as the hop-limit one, but worth a second look.
- The `Detail`-must-be-a-JSON-object claim, by contrast, is verified precisely: the [PutEventsRequestEntry API reference](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEventsRequestEntry.html) states *"A valid JSON object. There is no other schema imposed."* — exactly what the PR implements (dropping non-object `Detail` from either arm, via the shared `isObject()` check added in the third commit).

**No test exercises the depth ceiling or a cycle at all.** Both `EventBridgeInvokerTest` and the new `EventBridgeBusToBusIntegrationTest` only cover single-hop topologies (source-bus → target-bus → SQS, and the cross-account equivalent). `AGENTS.md`'s testing rules and this repo's Testing Policy call for coverage of new/changed behavior; the depth-ceiling/cycle-breaking logic is entirely new and entirely untested — there's no test proving a 2-hop chain actually forwards, that hop 6 is actually dropped, or that a genuine A→B→A cycle is actually broken rather than, say, silently stack-overflowing or double-delivering.

Everything else checks out: the `ThreadLocal` depth counter is sound (confirmed `EventBridgeService.putEvents()` invokes targets synchronously in a plain `for` loop with no executor/async boundary, so the same-thread assumption in the code comment holds), its push/pop via `try/finally` correctly restores prior depth on every exit path including exceptions, and the whole event-bus branch sits inside `invokeTarget`'s existing catch-all `catch (Exception e)`, so a malformed ARN reaching `AwsArnUtils.parse()` degrades to a warn-and-drop like every other target type rather than propagating. Envelope field names (`source`, `detail-type`, `region`, `account`, `resources`, `detail`) match exactly what `buildEventEnvelope()` emits, and the `Resources` `ArrayNode` passed through survives the round-trip cast in the next hop's `buildEventEnvelope()` without a `ClassCastException`.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
tgibson22: Good catch — thank you, that was a real inconsistency and I'd had it backwards. Checked your
citations and the dedicated feature page is blunter still: "EventBridge can't route events
received from a sender event bus to a third event bus."

`MAX_BUS_TO_BUS_DEPTH` is now 1 in 0e8e90d2. Being more permissive than AWS is the wrong
direction for an emulator — a three-bus chain worked here and would have died silently at hop 2
in production.

Coverage added: `forwardingStopsAfterOneHop` runs A→B→C with a sink on both B and C, asserting B
receives and C doesn't. Confirmed it fails at the old limit, with the event reaching C carrying
`"event-bus-name":"hop-bus-c"`.

The event-id point I've parked rather than guessed at — both docs pages are silent on IDs, so
the blog line is the only source. Flagged as unconfirmed in the description.

docs/services/eventbridge.md
tgibson22: Fixed in 0e8e90d2 — thanks for catching it, the line said the opposite of the truth. It now
reads:

> An event is forwarded between buses only once, matching AWS: a bus that received an event from
> another bus does not forward it on to a third.

Since the cap is now 1 rather than 5, Floci matches AWS instead of exceeding it, so the "more
permissive than" wording isn't needed.

pgermosen:
Re-checked after `0e8e90d`.

This closes the headline finding precisely: `MAX_BUS_TO_BUS_DEPTH` is now 1, matching AWS's documented one-hop-only restriction, and the docs line that used to (incorrectly) claim "no AWS equivalent" now states the real behavior. The new `forwardingStopsAfterOneHop` test builds an actual A→B→C chain with a sink on both B and C and asserts B receives while C doesn't — that also closes the test-coverage gap, since nothing previously exercised more than one hop.

Not re-addressed, and still only a PLAUSIBLE (not confirmed) point from the original review: each surviving hop still stamps a new event id rather than preserving the original, where AWS's bus-to-bus documentation describes the delivered event as identical to the original. Lower stakes now that only one hop exists, and I couldn't find as authoritative a source for that specific claim as I could for the hop-count limit — worth a look but not blocking.

Otherwise ready: `mergeable: MERGEABLE`, `docs-check` clean.

Issue #1532, opened 2026-06-24
Custom-resource Lambda cannot reach S3 (connect ECONNREFUSED :443) → CDK autoDeleteObjects bucket fails


Shikharinspire:
## Environment
- floci: `floci/floci:nightly` (**nightly-06232026**)
- Docker: 29.5.2
- IaC: AWS CDK (`aws-cdk-lib`), `cdk deploy` against floci (account `000000000000`, region `ap-south-1`)

## Summary
A CDK S3 bucket with `autoDeleteObjects: true` (+ `removalPolicy: DESTROY`) adds a `Custom::S3AutoDeleteObjects` resource backed by an AWS-managed Lambda. On stack create the custom-resource Lambda errors connecting to S3 (`connect ECONNREFUSED 172.19.0.2:443`), so the custom resource fails and the stack rolls back.

## Repro
1. CDK stack with `new s3.Bucket(this, 'B', { autoDeleteObjects: true, removalPolicy: RemovalPolicy.DESTROY })` (here: a CloudFront static-asset origin bucket).
2. `cdk deploy` against floci.
3. `Custom::S3AutoDeleteObjects/.../Default` → CREATE_FAILED:
   ```
   Custom resource handler errored (Handled):
   {"errorType":"Error","errorMessage":"connect ECONNREFUSED 172.19.0.2:443",
    "trace":["Error: connect ECONNREFUSED 172.19.0.2:443",
             "    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1637:16)"]}
   ```
   → `ROLLBACK_COMPLETE`. (The CloudFront Distribution / OAC / VpcOrigin resources themselves create fine.)

## Expected
The custom-resource Lambda can reach the S3 service endpoint within floci's network, so `autoDeleteObjects` buckets create successfully.

## Actual
Lambda → S3 HTTPS connection refused at `172.19.0.2:443` → custom resource fails → rollback.

## Impact
Any stack using `autoDeleteObjects` (a very common CDK pattern, especially for CloudFront origin / static-asset buckets) cannot deploy on floci.

hectorvent:
Thanks for the detailed report. I reproduced this, and the key is which endpoint the handler uses.

On a stock Floci the happy path works. Floci injects `AWS_ENDPOINT_URL=http://floci:4566` into the Lambda container (since #640), and an S3 client that honors it lists and deletes objects fine, with both `boto3` and `@aws-sdk/client-s3` v3.

[...]

PR #1821, opened 2026-07-11
fix(tls): add host.docker.internal to self-signed cert SANs

JESUSLUG:
## Summary #1532 

add host.docker.internal to self-signed cert SANs

Lambda containers reach Floci via host.docker.internal when Floci
runs on the host instead of in a container, but the cert never
included it as a SAN, failing TLS verification.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/config/TlsConfigSource.java
+++ b/src/main/java/io/github/hectorvent/floci/config/TlsConfigSource.java
@@ -49,6 +49,11 @@ public class TlsConfigSource implements ConfigSource {
     private static final String TLS_DIR = "tls";
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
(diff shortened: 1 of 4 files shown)


JESUSLUG:
Hello @hectorvent ! Added `host.docker.internal` to the self-signed cert's SANs.

Lambda containers use that hostname to reach Floci when it's running on the host (not in a container), and it was missing, so TLS verification still failed (`ERR_TLS_CERT_ALTNAME_INVALID`) even with `floci.tls.enabled=true`.

Re-ran the same repro (host + `FLOCI_TLS_ENABLED=true` + CDK bucket with `autoDeleteObjects: true`): `Custom::S3AutoDeleteObjects` now reaches `CREATE_COMPLETE` instead of rolling back.

hectorvent (approved):
Thanks @JESUSLUG.

PR #1827, opened 2026-07-11
perf(lambda): reuse a populated code volume across restarts via a completion marker

abanna:
## Summary

A large function's `/var/task` is served from a content-addressed (`codeSha256`) read-only Docker volume, populated once by streaming its files into a helper container. That volume survives an emulator restart, but the in-process set tracking populated volumes does not, so every restart previously re-copied all the files.

This persists a completion marker under `{persistentPath}/lambda-codevol-markers/`, written only after population fully succeeds. On a later boot, when the volume still exists and its marker is present, the volume is mounted as-is and the copy is skipped. A crashed or partial populate leaves no marker, and a marker whose volume was pruned is ignored, so population is safely retried.

If Docker volume inventory is temporarily unavailable, Floci invalidates only the current volume's marker before degrading to the normal population path. The marker is recreated only after that population succeeds, so a failed retry cannot leave a partial volume trusted on the next launch. If the old marker cannot be invalidated, Floci fails closed rather than risk unsafe reuse.

The marker write path replaces pre-existing marker symlinks without following them. Tar-producer failures are treated as fatal specifically while populating a reusable code volume, so an incomplete stream cannot produce a trusted completion marker. Existing direct copies for small functions and layers retain their established behavior.

This PR contains one feature: safe reuse of a populated Lambda code volume across emulator restarts.

## Type

- [x] Performance

## AWS Compatibility

No AWS wire-protocol surface changes. This is an internal Lambda-container cold-start optimization. Function code remains available at `/var/task`; callers observe the same Lambda API behavior, while Floci can skip the one-time population only when a previous run completed the exact code version.

## Verification

- `mise exec java@25 -- ./mvnw test -Dtest=ContainerLauncherCodeVolumeReuseTest,ContainerLauncherTest,ContainerLifecycleManagerVolumeTest`
- `ContainerLauncherCodeVolumeReuseTest`: 9/9
- `ContainerLauncherTest`: 14/14
- `ContainerLifecycleManagerVolumeTest`: 19/19
- Focused total: 42 tests passed with no failures or errors
- `FLOCI_ENDPOINT=http://localhost:4566 mise exec java@25 -- ./mvnw -f compatibility-tests/sdk-test-java/pom.xml test -Dtest=LambdaTest`
- AWS SDK Lambda compatibility: 18/18
- `mise exec java@25 -- ./mvnw test`
- Full repository suite: 8,117 tests, 0 failures, 0 errors, 9 skipped
- `git diff --check`

## Checklist

- [x] Completion marker is written only after successful population
- [x] Stale markers are invalidated before repopulation
- [x] Failed repopulation cannot leave an old completion marker trusted
- [x] Marker writes do not follow pre-existing symlinks
- [x] Tar-producer failures stop code-volume population before a marker is trusted
- [x] Existing direct-copy behavior remains unchanged

## Additional build reliability

Commit `1596584d` also adds curl retries for the `gosu` download in `docker/Dockerfile.native-package` to tolerate transient download failures during native-package image builds. This is unrelated to Lambda code-volume reuse and does not affect runtime behavior.


src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
```diff
@@ -472,6 +484,30 @@ private void populateCodeVolume(String volName, LambdaFunction fn, String image)
...
     }
 
+    // A code volume survives an emulator restart, but the in-process populatedCodeVolumes set does not.
+    // A persisted completion marker lets a later run recognize an already-populated, content-addressed
+    // volume and skip re-copying its files. The marker lives under the storage persistent path so it
+    // shares the volume's lifetime expectations (hybrid/persistent storage); in memory-only mode the
+    // marker dir is transient too, which is correct because nothing persists across restarts there.
+    private Path codeVolumeMarkerPath(String volName) {
+        return Path.of(config.storage().persistentPath(), "lambda-codevol-markers", volName);
+    }
```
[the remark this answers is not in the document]
abanna: Fixed in f189b04ea. `ContainerLifecycleManager` now exposes a fail-closed authoritative volume inventory, and each successful fresh population prunes only internal marker files whose volumes are confirmed absent. Marker decisions/writes/pruning are serialized. A stale marker is deleted before repopulation, so a failed copy cannot leave an old completion signal beside a partial volume; inventory failure preserves markers and aborts reuse safely. The focused 37-test Docker/Lambda suite passes.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
```diff
@@ -466,12 +481,90 @@ private void populateCodeVolume(String volName, LambdaFunction fn, String image)
...
+        synchronized (codeVolumeMarkerIoLock) {
+            Path marker = codeVolumeMarkerPath(volName);
+            if (!Files.isRegularFile(marker, java.nio.file.LinkOption.NOFOLLOW_LINKS)) {
+                return false;
+            }
+            java.util.Optional<java.util.Set<String>> volumeNames = lifecycleManager.tryListVolumeNames();
+            if (volumeNames.isEmpty()) {
+                throw new IllegalStateException(
+                        "Could not verify the backing container volume for code-volume marker " + volName);
+            }
```
pgermosen: **Reuse path throws (aborts the launch) on a transient volume-list failure** (robustness).

When `tryListVolumeNames()` returns empty (runtime couldn't be queried) **and a marker is present**, this throws `IllegalStateException`. It propagates to the launch's `catch (RuntimeException e)` at `:339`, which cleans up and **rethrows** — so a transient `docker volume ls` hiccup becomes a hard Lambda launch failure.

Two reasons that's the wrong failure mode here:
1. **Regression vs. pre-PR behavior** — before this change a restart simply re-populated the volume (slow, but the launch succeeded); this makes the same situation fail outright.
2. **Inconsistent with the sibling path** — `pruneOrphanCodeVolumeMarkers` handles the *identical* `tryListVolumeNames().isEmpty()` case by failing **safe** (it just returns, `:547`).

Since this is a perf optimization, the uncertain case should degrade to the slow path (re-populate), not abort:

```suggestion
            if (volumeNames.isEmpty()) {
                // Runtime couldn't be queried — degrade to the slow path (re-populate) instead of
                // aborting the launch, matching how pruneOrphanCodeVolumeMarkers fails safe below.
                LOG.warnv("Could not verify backing volume for code-volume marker {0}; re-populating", volName);
                return false;
            }
```

Leaving the marker in place is fine: the next boot re-verifies, and `populateCodeVolume` → `writeCodeVolumeMarkerAndPruneOrphans` rewrites it. If the daemon is genuinely down, `populateCodeVolume` will surface the real error instead of a synthetic "could not verify" one.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
```diff
@@ -466,12 +481,90 @@ private void populateCodeVolume(String volName, LambdaFunction fn, String image)
...
             }
             POPULATE_SEMAPHORE.release();
         }
     }
 
+    // A code volume survives an emulator restart, but the in-process populatedCodeVolumes set does not.
+    // A persisted completion marker lets a later run recognize an already-populated, content-addressed
+    // volume and skip re-copying its files. The marker lives under the storage persistent path so it
+    // shares the volume's lifetime expectations (hybrid/persistent storage); in memory-only mode the
+    // marker dir is transient too, which is correct because nothing persists across restarts there.
```
pgermosen: **Comment is inaccurate** (minor).

This says "in memory-only mode the marker dir is transient too, which is correct because nothing persists across restarts there" — but `persistentPath()` is `./data` regardless of storage mode, so the marker actually persists on disk in memory mode. The *behavior* is still correct (the Docker volume also persists in memory mode, so reusing it is valid), but the stated rationale is wrong — worth a tweak so a future reader doesn't rely on "the marker is transient in memory mode."

pgermosen (approved):
Looks good,  

just a few minor things to improve.

abanna:
Thanks for the careful review — I traced both observations against the current launch and marker lifecycle.

The robustness concern is valid. When `tryListVolumeNames()` returns `Optional.empty()`, the current path throws, the launch cleanup runs, and the exception is rethrown. That means a transient volume-inventory failure aborts the Lambda launch instead of degrading to the pre-existing population path.

One safety refinement is needed beyond the suggested `return false`: retaining the old completion marker while repopulating would be unsafe if the copy subsequently failed. A later launch could see the marker and backing volume and incorrectly trust partially populated `/var/task` contents. The existing `failedRepopulateCannotLeaveOldCompletionMarker` regression captures that invariant.

The safe correction is therefore to invalidate the marker before entering the slow population path, log the degraded fallback, and recreate the marker only after population succeeds. Coverage should verify both that a successful fallback rewrites the marker and that a failed fallback leaves it absent.

The documentation observation is also correct. `codeVolumeMarkerPath` always uses `floci.storage.persistent-path`—`./data` by default—regardless of `floci.storage.mode`, so memory mode does not inherently make the marker transient. The accurate explanation is that marker durability follows the configured persistent path, while reuse still requires the independently persisted Docker volume to exist.

Thanks again — these caught a real robustness regression and a misleading rationale.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
abanna: Thanks for catching the launch regression. Fixed in `190505f93`. An inconclusive Docker volume inventory now invalidates only that volume’s completion marker before falling back to the normal population path, then recreates the marker only after population succeeds. I deliberately did not retain the old marker during fallback: if population fails, leaving it would allow a partial `/var/task` volume to be trusted on the next launch. Marker invalidation failure still fails closed. Added regressions for successful fallback and failed-copy/no-marker behavior. Verification: 60/60 launcher/lifecycle tests locally; exact-head CI/native/SDK matrix and Greptile are 14/14 green.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
abanna: Thanks — corrected in `190505f93`. The comment now states that markers live under `storage.persistent-path` regardless of storage mode, and that safe reuse also requires the independently persisted Docker volume. Exact-head CI and Greptile are green.

pgermosen (approved):
Re-approving on the current head. My earlier approval was anchored to `409c61cc`, which is no longer in this branch — it was force-pushed, and the feature commit was rewritten as `179d8db7`. GitHub carries approvals across force-pushes, so this was reading as reviewed while the reviewed artifact no longer existed. This one is on `98b9a69c`.

**Both of my July comments are fixed**, in `190505f9`:

- The transient `docker volume ls` failure no longer aborts the launch — an inconclusive inventory invalidates only that volume's marker and falls back to normal population, recreating the marker only after success. Not retaining the old marker during fallback is the right call, and your reasoning for it is the same one that makes the whole design safe: a completion signal must never outlive certainty about the data it describes.
- The memory-mode comment now says markers live under `storage.persistent-path` regardless of storage mode, and that reuse still requires the independently persisted Docker volume. Confirmed in the code, not just the reply.

**I checked whether #2208 had made this redundant**, since it merged in the meantime and touches exactly this code. It hasn't. `cleanupSupersededVolumes`, `volumeCleanupScheduler`, `ensureCodeVolume` and `reservedCodeVolume` have identical counts on `main` and here, so that machinery arrived *from* `main` through your merges rather than being duplicated.

What this PR still contributes on top of current `main` is the distinct half: a completion marker that survives an emulator restart, so a later run recognises an already-populated content-addressed volume instead of re-copying into it. #2208 reconciles in-process state against Docker; this persists across process death. They compose.

The safety properties around the marker are the part I looked hardest at, and they hold: reuse requires marker *and* volume, a stale marker is deleted before repopulation so a failed copy can't leave a completion signal beside partial data, marker I/O is serialised under `codeVolumeMarkerIoLock`, and the marker check uses `NOFOLLOW_LINKS`. Tightening `copyDirToContainer` to `copyDirToContainerStrict` and turning the silently-swallowed helper cleanup into a logged warning are both improvements I didn't ask for.

One note, non-blocking: `1596584d` adds curl retries to the gosu download in `docker/Dockerfile.native-package`. Sensible given the download outages today, and harmless — but it's unrelated to code-volume reuse and isn't mentioned in the title or description. Worth a line in the description so it isn't a surprise in the merge commit.

@abanna

PR #1830, opened 2026-07-11
test(apigateway): cover request-header forwarding on the _user_request_ path

abanna:
## Summary
Adds a regression test proving a custom request header reaches the integration on **both** the `/execute-api/` and the LocalStack-compatible `_user_request_` dispatch paths. It uses a MOCK integration whose response template echoes `$input.params('X-Custom')`, so no Lambda container is needed and it runs in the standard test job.

This guards the `_user_request_` → `ApiGatewayExecuteController.dispatch` boundary: the same `headers.getRequestHeaders()` that this test exercises is what the proxy-integration and REQUEST-authorizer event builders read, so a regression that emptied headers on that path would fail here.

## Type
- [x] Test

## AWS Compatibility
No production code change. Confirms the emulator forwards request headers to the integration identically on both supported execute URL formats.

## Checklist
- [x] Test passes on both paths (built + run offline, maven:3.9-eclipse-temurin-25)
- [x] No Lambda/Docker dependency (MOCK + VTL header echo)


hectorvent (approved):
Thanks @abanna

PR #1834, opened 2026-07-11
test(terraform): expand Floci IaC compatibility coverage

kotov228:
## Summary

Expand Terraform and OpenTofu compatibility coverage for Floci's local AWS-compatible endpoint.

### Added

- Terraform and OpenTofu resource fixtures with matching provider endpoint configuration.
- Create/read assertions for 46 compatibility scenarios in each IaC suite.
- Coverage across core resources plus API Gateway, AppConfig, Athena, Backup, Cloud Map, CodeDeploy, ACM, SES, Scheduler, Transfer, WAFv2, CloudTrail, CUR, AppSync, Glue, ElastiCache, Firehose, Pipes, Batch, Neptune, and OpenSearch.
- A user-facing Terraform guide covering local state, emulated S3 backend state, and DynamoDB locking.
- README and MkDocs navigation updates.

## Validation

- `terraform init -backend=false`
- `terraform validate`
- `tofu init -backend=false`
- `tofu validate`
- `terraform fmt`
- `tofu fmt`
- `git diff --check`

## Notes

Full `apply`/`destroy` compatibility execution requires a running Floci instance and Docker access. Docker-backed resources such as Batch, Neptune, and OpenSearch therefore still need runtime CI verification.


kotov228:
@hectorvent Please check that

kotov228:
Fixed in commit `ee6d61d9`.

Replaced deprecated Terraform S3 backend `force_path_style` with `use_path_style` in:
- `docs/getting-started/terraform.md`
- `compatibility-tests/compat-terraform/backend.hcl`
- `compatibility-tests/compat-terraform/test/test_helper/common-setup.bash`

The Terraform provider-level `s3_use_path_style` setting remains unchanged.

kotov228:
I verified this against the provider versions in the suite. `compat-terraform` requires `hashicorp/aws ~> 6.0`; with the installed v6.54.0 provider, `terraform validate` accepts `name` and rejects `compute_environment_name` as an unsupported argument. `compat-opentofu` uses `hashicorp/aws ~> 5.0`, where `compute_environment_name` is valid.

The Terraform fixture therefore intentionally keeps `name = "floci-compat-batch"`, while the OpenTofu fixture uses `compute_environment_name`. Terraform validation passes with this version-specific configuration.

abanna:
@kotov228 — thanks for pushing the IaC coverage this much wider; that's the suite most likely to catch real provider-shaped breakage.

Both suites this PR expands are currently red, and the failures are live rather than stale — I checked that run `29145897063` was built from `abfc9414`, the current head. `terraform apply` fails in `setup_file`, so the whole suite aborts before any test body runs. There are seven distinct errors, and they don't all belong to you:

**Already fixed on `main` — a rebase should clear it**

```
Error: creating CloudTrail Trail (floci-compat-trail): ... UnknownOperationException:
       Unknown operation: CloudTrail_20131101.CreateTrail
```

This run predates #1460, which landed CloudTrail on 2026-07-21; `CloudTrailJsonHandler` handles `CreateTrail` on `main` today.

**Test-side, fixable in this PR**

```
Error: `buildspec` must be set when source's `type` is `NO_SOURCE`
```
Terraform's own validation of `aws_codebuild_project` — never reaches Floci.

```
Error: creating ElastiCache Cache Cluster: InvalidParameterValue: Engine must be
       'memcached'. For Redis/Valkey use CreateReplicationGroup.
```
`aws_elasticache_cluster` with a Redis engine maps to `CreateCacheCluster`. Real AWS accepts single-node Redis there, so Floci is arguably the strict one — but `aws_elasticache_replication_group` is the change that gets the suite green without waiting on a service fix.

```
Error: creating Cost And Usage Report Definition: ValidationException: Format must be one of [Parquet]
```
Same shape: AWS accepts `textORcsv` too, so this is a Floci limitation, but switching the `.tf` to Parquet unblocks it.

**Real Floci gaps — these need their own PRs before the resources can live in the suite**

I checked each against current `main`, and all three are still missing:

```
Error: listing tags for Athena WorkGroup ...: InvalidAction: Action ListTagsForResource is not supported
Error: reading Neptune Global Cluster information ...: UnsupportedOperation: Operation DescribeGlobalClusters is not supported.
Error: unexpected format for ID (:floci_compat_db), expected catalog-id:database-name
```

The Glue one is the interesting one: the provider builds the resource ID as `catalog-id:database-name`, and the leading segment is empty, so Floci's Glue database response is omitting `CatalogId` — there's no `CatalogId` anywhere under `services/glue` on `main`. That's a small, self-contained fix worth splitting out.

**One more worth a look**

```
Error: The terraform-provider-aws_v6.54.0_x5 plugin crashed!
```

A provider panic usually means a response is missing a field the provider dereferences unconditionally, so it may point at a real response-shape bug rather than just fallout from the errors above. The stack in the raw log should name the resource.

Last thing: the README in this PR moves the counts to 54 / 48 and the headline to "2,570 automated compatibility tests". Since those numbers are a claim about a suite that doesn't currently pass, it'd be worth landing them together with a green run — otherwise the README asserts coverage the CI contradicts.

Happy to re-review once it's rebased and back to green.

hectorvent:
Thank you for expanding the Terraform coverage; that suite catches things the SDK suites do not.

**(blocking)** Both suites are still red, and @abanna's breakdown above is accurate, so I will point at it rather than repeat it. I re-checked the three he calls real Floci gaps and all three are still missing on main: Athena `ListTagsForResource`, Neptune `DescribeGlobalClusters`, and Glue never emitting `CatalogId`, which is why the provider builds the id as `:floci_compat_db`. Those want their own PRs before the resources can live here. The CloudTrail error is stale, and the branch now conflicts with main in both `provider.tf` files, so a rebase is needed anyway and clears it.

**(blocking)** The README moves the counts to 54/48 and the headline to "2,570 automated compatibility tests" while the suite does not pass, so those numbers want to land with a green run.

A line on running these locally would help too, since the Terraform suite needs requesting by name.

pgermosen (approved):
Approving. This required substantially more than the downmerge: getting the suite from red to green surfaced 9 real Floci gaps (Athena ListTagsForResource, Neptune DescribeGlobalClusters, Glue CatalogId, CodeBuild BatchGetProjects ARN lookup, CloudTrail ListTags, OpenSearch CognitoOptions crashing the provider, Transfer Server's incorrect OFFLINE-to-delete requirement, Batch UpdateComputeEnvironment being unimplemented, and API Gateway's missing apiStatus field hanging the create waiter) — each only reachable once the resources ahead of it in the apply/destroy sequence actually succeeded. All tracked as #2791, #2792, #2793, #2799, #2800, #2801, #2802, #2803, #2804, #2806, with the affected resources pulled from the suite until each lands. One genuine test-config bug from the downmerge itself (a Firehose delivery-stream name collision with an unrelated main change) is fixed. All 13 CI checks are green, including compat-terraform end to end.

PR #1840, opened 2026-07-11
feat(ses): implement custom verification email templates for v1 and v2

okinaka:
## Summary

Adds the SES custom verification email template CRUD APIs across **both** protocols, backed by one shared `StorageFactory` store. A custom verification email template is the branded email SES sends when verifying a new sender address (distinct from the `CreateTemplate` templates used for sending to recipients).

**v1 (Query/XML)** and **v2 (REST JSON, `/v2/email/custom-verification-email-templates`)**:

| Operation | Behaviour |
| --- | --- |
| `Create` | `AlreadyExistsException` on duplicate |
| `Get` | full template (all fields) |
| `List` | metadata only — **omits `TemplateContent`** |
| `Update` | `NotFoundException` when missing |
| `Delete` | `NotFoundException` when missing |

Templates are shared between the two protocols (a v1-written template is returned by the v2 Get, and vice versa).

`SendCustomVerificationEmail` is intentionally out of scope here and left as a follow-up.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Behaviour and every error string were verified against real AWS (SES v1 + v2, boto3, us-east-1):

- Duplicate create → `AlreadyExistsException` / `Custom verification email template <name> already exists`.
- Get/Update/Delete of a missing template → `NotFoundException` / `Custom verification email template <name> does not exist`.
- `FromEmailAddress` must be a **verified identity** (or a verified domain) → `NotFoundException` / `The from email address <x> is not verified`. Floci enforces this against its own identity store (it does track verified identities), unlike checks that need an external registry.
- Required-field and invalid-redirection-URL validations surface the v1-native `InvalidParameterValue` on the Query API (consistent with the other SES Query validations) and are remapped to `BadRequestException` on v2 by the controller. The success URL is validated before the failure URL.
- The template body is **not** content-validated (AWS accepts a body without a link).
- `List` omits `TemplateContent`; `Get` includes it.

## Checklist

- [x] `./mvnw test` passes locally (SES suite + SDK compatibility test)
- [x] New or updated integration test added (v1 + v2 integration and an SDK compatibility test)


okinaka:
Thanks for the review — addressed in 251cd60c.

- **Concurrency (`updateCustomVerificationEmailTemplate`)**: renamed `cvetCreateLock` → `cvetMutationLock` and now guard the check-then-write of both `update` and `delete` on it, not just `create`. `update`'s existence check, validation, and put run inside the lock, so a concurrent `delete` can no longer slip between the check and the put and resurrect the just-deleted template; `delete`'s check-then-delete is guarded on the same lock so all three mutations serialize against each other.
- **Test hygiene**: the v1 integration test now deletes the `cvet-v1-shared` template it creates (previously left behind).

src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
```diff
@@ -1015,6 +1027,137 @@ public List<EmailTemplate> listTemplates(String region) {
...
+    private static void requireCvetField(String value, String name) {
+        if (value == null || value.isBlank()) {
+            // v1-native code; the v2 controller remaps it to BadRequestException via remapV1Exception,
+            // so the Query API stays consistent with requireParam and v2 behavior is unchanged.
+            throw new AwsException("InvalidParameterValue", name + " is required.", 400);
+        }
+    }
+
+    private static AwsException cvetNotFound(String templateName) {
+        return new AwsException("NotFoundException",
```
abanna: Nice to see both protocols sharing one store, and the concurrency reasoning around `cvetMutationLock` is careful — the "an update can't resurrect a concurrently deleted template" case is easy to miss.

One thing looks inverted relative to how the rest of SES is layered here. Everywhere else in this service the v1 (Query) error code is what gets thrown, and `SesController.remapV1Exception` translates it to the v2 shape at the JSON boundary — that's what `TemplateDoesNotExist → NotFoundException` and `ConfigurationSetAlreadyExists → AlreadyExistsException` are doing at lines 2202–2205. The CVET methods throw the *v2* codes directly (`cvetNotFound` → `NotFoundException`/404 here, and `AlreadyExistsException`/400 at line 1042), so the remap is a no-op for them and the v1 path is the one that ends up wrong:

| | Floci v1 today | Real SES v1 |
|---|---|---|
| Get missing | `NotFoundException`, 404 | `CustomVerificationEmailTemplateDoesNotExist`, 400 |
| Create duplicate | `AlreadyExistsException`, 400 | `CustomVerificationEmailTemplateAlreadyExists`, 400 |

That matters concretely because the v1 SDK models these as their own exception types — with `NotFoundException` on the wire, `catch (CustomVerificationEmailTemplateDoesNotExistException e)` never fires and the caller gets a generic `SesException` instead. The status code is off for the not-found case too.

Following the existing convention fixes both protocols at once — throw the v1 codes from the service and add the two cases to `remapV1Exception`:

```java
case "CustomVerificationEmailTemplateDoesNotExist" ->
        new AwsException("NotFoundException", e.getMessage(), 404);
case "CustomVerificationEmailTemplateAlreadyExists" ->
        new AwsException("AlreadyExistsException", e.getMessage(), 400);
```

Two smaller things that fall out of the same area:

`getCustomVerificationEmailTemplate` and `deleteCustomVerificationEmailTemplate` in `SesController` are the only CVET endpoints without the `try`/`catch (AwsException e) { throw remapV1Exception(e); }` wrapper that create and update have. It's invisible today precisely because the service already emits v2 codes, but it would become a live bug the moment the codes above are corrected — worth adding the wrapper in the same change.

The v1 integration test asserts `<Code>InvalidParameterValue</Code>` for the missing-field cases, so v1 shapes are clearly on your radar — but the not-found and duplicate cases only assert `statusCode(400)` / `statusCode(200)`, and the SDK compat test imports the sesv2 `NotFoundException`/`AlreadyExistsException` only. So nothing currently pins the v1 error codes. A v1 assertion on `<Code>` for both would close that.

okinaka: @abanna Thanks — spot on, the layering was inverted. Fixed in c569669c:

- The service now throws the v1-native codes (`CustomVerificationEmailTemplateDoesNotExist` / `CustomVerificationEmailTemplateAlreadyExists`, both 400), and `remapV1Exception` translates them to `NotFoundException`/404 and `AlreadyExistsException`/400 — so v1 is correct and v2 is unchanged.
- Added the missing `try`/`catch (AwsException e) { throw remapV1Exception(e); }` wrapper to the get/delete CVET endpoints.
- Pinned both v1 codes with `<Code>` assertions in the v1 integration test.

I probed the full matrix against real AWS before touching the wire codes — v1: `CustomVerificationEmailTemplateDoesNotExist`/400 and `...AlreadyExists`/400; v2: `NotFoundException`/404 and `AlreadyExistsException`/400 — which matches exactly what you laid out. Left the `getConfigurationSet` v1/v2 shape alone since, as you noted, the mapper already handles it.

hectorvent:
Thank you for the update; the v1 error surface is now exactly right where I checked it, `CustomVerificationEmailTemplateAlreadyExists` and `CustomVerificationEmailTemplateDoesNotExist` match the model's error code traits including the 400 status, and asserting the `<Code>` element in the v1 tests pins it on the wire ([botocore ses/2010-12-01](https://github.com/boto/botocore/blob/develop/botocore/data/ses/2010-12-01/service-2.json)).

**(blocking)** One case still crosses protocols: an unverified `FromEmailAddress` throws `NotFoundException` 404 from the shared validator, so the v1 Query wire shows a v2 code. SES classic declares `FromEmailAddressNotVerifiedException` (code `FromEmailAddressNotVerified`, 400) on Create and Update. The same pattern you used for the other two codes should work: throw the v1-native code and let `remapV1Exception` produce your probe-verified `NotFoundException` 404 for v2.

That is the last item from my side.

okinaka:
Thanks @hectorvent — fixed in 94ced5ec, same pattern as the other two codes: an unverified `FromEmailAddress` now throws the v1-native `FromEmailAddressNotVerified` (400) from the shared validator, and `remapV1Exception` maps it to `NotFoundException` (404) for the v2 boundary.

I probed both protocols to confirm before changing the wire code — v1: `FromEmailAddressNotVerified` / 400, v2: `NotFoundException` / 404 — matching the classic SES model you referenced. Pinned the v1 `<Code>` with an integration assertion; v2 is unchanged.

hectorvent (approved):
Thanks @okinaka

Issue #1654, opened 2026-06-30
[FEAT] Cognito ConfirmSignUp does not set email_verified or phone_number_verified


dzwicker:
## Service

  Cognito

  ## AWS API Action

  `ConfirmSignUp`

  ## Related issues

  Follow-up to:
  - #1487 `[BUG] Cognito ConfirmSignUp succeeds without validating confirmation code`
  - #1488 `fix(cognito): align sign-up confirmation with AWS behavior`

  This issue is intentionally limited to verified-attribute updates during `ConfirmSignUp`.

  Out of scope:
  - `GetUserAttributeVerificationCode`
  - delivery-priority parity between `email` and `phone_number`
  - stale code invalidation after `AdminConfirmSignUp`

  ## Expected behavior

  After a successful `ConfirmSignUp`, Floci should update the confirmed contact attribute to match AWS Cognito behavior.

  Expected outcome:
  - if the sign-up confirmation was completed for email verification, `email_verified` should become `true`
  - if the sign-up confirmation was completed for phone verification, `phone_number_verified` should become `true`

  The user becoming `CONFIRMED` alone is not sufficient if the corresponding verified flag is left unchanged.

  ## Actual behavior

  Floci confirms the user successfully, but the corresponding verified attribute is not updated.

  Observed effect:
  - `ConfirmSignUp` succeeds
  - `UserStatus` becomes `CONFIRMED`
  - but `email_verified` and/or `phone_number_verified` do not reflect the successful confirmation flow

  This creates an AWS compatibility gap, especially for flows that depend on verified aliases or on downstream checks of `email_verified` / `phone_number_verified`.

  ## Reproduction

  ```bash
  # 1. Create a user pool with auto-verified attributes
  POOL_ID=$(aws --endpoint-url http://localhost:4566 cognito-idp create-user-pool \
    --pool-name SignUpPool \
    --auto-verified-attributes email phone_number \
    --query 'UserPool.Id' \
    --output text)

  # 2. Create a user pool client
  CLIENT_ID=$(aws --endpoint-url http://localhost:4566 cognito-idp create-user-pool-client \
    --user-pool-id "$POOL_ID" \
    --client-name signup-client \
    --query 'UserPoolClient.ClientId' \
    --output text)

  # 3. Sign up a new user
  aws --endpoint-url http://localhost:4566 cognito-idp sign-up \
    --client-id "$CLIENT_ID" \
    --username alice@example.com \
    --password 'Passw0rd!' \
[...]

PR #1849, opened 2026-07-12
fix(cognito): AWS-accurate `UsernameAttributes` pools — UUID username, alias sign-in, and token/revocation parity

AbhigyaKrishna:
## Summary

Makes Floci behave like real AWS Cognito for user pools created with `UsernameAttributes` (e.g. `--username-attributes email`). Previously Floci used the email as the canonical `Username`, collapsing the immutable identity and the mutable sign-in alias into one value — the classic "leniency trap" where both correct and incorrect consumer code pass locally but diverge on real AWS.

For a `UsernameAttributes` pool, the canonical `Username` is now an auto-generated, immutable **UUID equal to `sub`**, and the email/phone is a **mutable sign-in alias**. Pools **without** `UsernameAttributes` (classic pools and `AliasAttributes` pools) are unchanged — the behavior is gated entirely on a non-empty `UsernameAttributes` list.

### Changes

1. **UUID username + alias (P0)** — `AdminCreateUser` / `SignUp` / user-migration mint a server UUID as the canonical `Username` (== `sub`) and store the supplied email/phone as the alias attribute; a non-email/phone value is rejected with `InvalidParameterException`.
   - `AdminGetUser` / `ListUsers` return `Username = <UUID>`.
   - Sign-in resolves by the current email alias **or** the UUID.
   - `AdminUpdateUserAttributes` rebinds the email while `Username` / `sub` stay fixed.
   - Lambda triggers receive `event.userName = <UUID>` with the email in `event.request.userAttributes`.
   - A client-supplied `sub` is ignored on create (the server always mints its own).
2. **`CUSTOM_AUTH` / SRP** — `ChallengeParameters.USERNAME` echoes the canonical UUID, and round-2 `SECRET_HASH` is validated against the `USERNAME` sent that round.
3. **Token-claim split** — the access token carries `username` (UUID) + `scope`; the ID token carries `cognito:username` + readable user attributes filtered by the client's `ReadAttributes`. `email` is no longer emitted in the access token and never falls back to the UUID for users without an email.
4. **Duplicate alias** — `UsernameExistsException` for an unverified duplicate, `AliasExistsException` for a verified one; `ForceAliasCreation=true` migrates a verified alias off the previous owner.
5. **Not-found status** — `ResourceNotFoundException` / `UserNotFoundException` now return HTTP `400`.
6. **Canonical-username revocation** — `AdminUserGlobalSignOut`, `GlobalSignOut`, and `RevokeToken` key their revocation markers off the resolved canonical `user.getUsername()`, so the global-sign-out marker matches the username every token carries (`RevokeToken` stays idempotent for deleted users).

> **Behavior change:** for `UsernameAttributes` pools the `Username` returned by `AdminGetUser` / `ListUsers` changes from the email to a UUID. This aligns with real AWS; consumers that relied on Floci's previous (non-AWS) `Username == email` behavior must read the email from the `email` attribute instead.
>
> **Existing local users:** users persisted before this change stay keyed by their email and continue to report `Username == email` rather than being rewritten to a UUID on upgrade. `AdminGetUser` does the direct key lookup first and then the alias scan, so such a user is still found by email and the dedupe avoids a spurious "Multiple users found". Rewriting persisted identities on upgrade would be the more disruptive choice, so it is deliberately avoided.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

**Incorrect behavior fixed.** On a `UsernameAttributes=["email"]` pool, Floci previously:

- returned `Username == email` everywhere (should be a UUID == `sub`);
- accepted both the email and the `sub` as the username leniently;
- leaked the UUID into the `email` claim and the email into the access token;
- returned HTTP `404` for not-found (Cognito uses `400`);
- keyed global-sign-out revocation by a non-canonical username.

**Verified against the AWS documentation and SDK wire behavior:**

- [Working with user attributes](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html) — UUID username == `sub`, non-email username rejected, email stored as alias, immutable username, E.164 phone format.
- [Access token](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-access-token.html) / [ID token](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-id-token.html) — claim split, `ReadAttributes` gating, access-token `scope`.
- [AdminCreateUser](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminCreateUser.html) — `ForceAliasCreation` / `AliasExistsException`, `400` not-found status.
- [RespondToAuthChallenge](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_RespondToAuthChallenge.html) — `ChallengeParameters.USERNAME` = UUID.

Integration tests drive the AWS JSON 1.1 wire protocol via RestAssured. Running the `compatibility-tests/` SDK suite against a real Cognito pool is recommended as a follow-up.

## Checklist

- [x] `./mvnw test` passes locally *(all Cognito suites green; full cross-service run needs Docker + JDK 25)*
- [x] New or updated integration test added *(new `CognitoAliasUsernameIntegrationTest` end-to-end P0 coverage, plus new unit / trigger tests; two pre-existing tests that encoded the old non-AWS behavior were corrected)*

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
@@ -14,6 +14,7 @@ import io.github.hectorvent.floci.services.lambda.model.InvokeResult;
 import org.jboss.logging.Logger;
 
 import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
@@ -320,8 +321,19 @@ final class CognitoAuthFlowHandler {
         } catch (Exception e) {
             throw new AwsException("InternalErrorException", "SECRET_HASH computation failed", 500);
         }
-        if (!expected.equals(provided)) {
+        if (!MessageDigest.isEqual(expected.getBytes(StandardCharsets.UTF_8),
+                provided.getBytes(StandardCharsets.UTF_8))) {
             throw new AwsException("NotAuthorizedException", "SECRET_HASH does not match", 400);
+        }
+    }
+
+    private String resolveToCanonicalUsername(UserPool pool, String username) {
+        try {
+            return service.adminGetUser(pool.getId(), username).getUsername();
+        } catch (AwsException e) {
+            LOG.debugv("Could not resolve USERNAME {0} in pool {1}: {2}",
+                    username, pool.getId(), e.getMessage());
+            return null;
         }
     }
 
@@ -504,9 +516,13 @@ final class CognitoAuthFlowHandler {
             throw new AwsException("InvalidParameterException", "ANSWER is required", 400);
         }
 
-        validateSecretHash(client, responses, responses.getOrDefault("USERNAME", state.username));
+        String suppliedUsername = responses.getOrDefault("USERNAME", state.username);
+        validateSecretHash(client, responses, suppliedUsername);
 
         CognitoUser user = service.adminGetUser(pool.getId(), state.username);
+        if (!user.getUsername().equals(resolveToCanonicalUsername(pool, suppliedUsername))) {
+            throw new AwsException("NotAuthorizedException", "Invalid session for the user.", 400);
+        }
 
         boolean answerCorrect = verifyAuthChallenge(pool, client, user, state, answer);
         if (!state.history.isEmpty()) {
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
@@ -448,7 +448,7 @@ final class CognitoAuthFlowHandler {
         }
 
         CustomAuthSession state =
-                new CustomAuthSession(pool.getId(), username, client.getClientId());
+                new CustomAuthSession(pool.getId(), user.getUsername(), client.getClientId());
         state.clientMetadata = clientMetadata == null ? Map.of() : clientMetadata;
 
         Map<String, Object> defineResp = defineAuthChallenge(pool, client, user, state);
@@ -466,7 +466,7 @@ final class CognitoAuthFlowHandler {
         state.currentChallengeName = challengeName;
 
         Map<String, String> publicParams = new HashMap<>();
-        publicParams.put("USERNAME", username);
+        publicParams.put("USERNAME", user.getUsername());
         for (Map.Entry<String, String> e : authParameters.entrySet()) {
             if (!"USERNAME".equals(e.getKey()) && !"SRP_A".equals(e.getKey())) {
                 publicParams.putIfAbsent(e.getKey(), e.getValue());
@@ -475,7 +475,7 @@ final class CognitoAuthFlowHandler {
         applyCreateResponse(state, challengeName,
                 createAuthChallenge(pool, client, user, state, challengeName), publicParams);
 
-        String sessionToken = buildSessionToken(pool.getId(), username, client.getClientId());
+        String sessionToken = buildSessionToken(pool.getId(), user.getUsername(), client.getClientId());
         customAuthSessions.put(sessionToken, state);
 
         Map<String, Object> result = new HashMap<>();
@@ -503,7 +503,8 @@ final class CognitoAuthFlowHandler {
         if (answer == null || answer.isBlank()) {
             throw new AwsException("InvalidParameterException", "ANSWER is required", 400);
         }
-        validateSecretHash(client, responses, state.username);
+
+        validateSecretHash(client, responses, responses.getOrDefault("USERNAME", state.username));
 
         CognitoUser user = service.adminGetUser(pool.getId(), state.username);
 
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -139,7 +139,9 @@ Floci follows real Cognito semantics for pools created with `UsernameAttributes`
 - Sign-in resolves by the **current** email alias **or** the UUID. `SECRET_HASH` is validated against
   the exact `USERNAME` value sent: `Base64(HMAC-SHA256(USERNAME + clientId, clientSecret))`.
 - In `CUSTOM_AUTH` / SRP flows, `ChallengeParameters.USERNAME` echoes the **UUID** (as real AWS does);
-  the `RespondToAuthChallenge` `SECRET_HASH` is validated against the `USERNAME` sent that round.
+  the `RespondToAuthChallenge` `SECRET_HASH` is validated against the `USERNAME` sent that round. That
+  `USERNAME` must still resolve to the session's user — the UUID or any of its current aliases — and a
+  value naming a different user is rejected with `NotAuthorizedException`.
 - `AdminUpdateUserAttributes` can change the email; afterwards sign-in works with the new email and
   fails with the old one, while `Username`/`sub` stay fixed.
 - Duplicate email/phone on `AdminCreateUser`: `UsernameExistsException` when the incoming alias is
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -126,6 +126,29 @@ Standalone `TagResource` rejects reserved `floci:*` keys. `ListTagsForResource` 
 - It validates requested OAuth scopes against the app client's `AllowedOAuthScopes` and the pool's registered resource-server scopes.
 - It advertises the prefixed token endpoint in `/{userPoolId}/.well-known/openid-configuration`.
 
+## Sign-in Identifiers (`UsernameAttributes`)
+
+Floci follows real Cognito semantics for pools created with `UsernameAttributes` (e.g.
+`--username-attributes email`):
+
+- `AdminCreateUser`/`SignUp` accept the email (or phone number) as the sign-in value, but the
+  **canonical `Username` is an auto-generated, immutable UUID equal to `sub`**. The supplied email is
+  stored as a mutable alias attribute.
+- `ListUsers`, `AdminGetUser`, tokens (`sub`/`username`/`cognito:username`) and Lambda trigger
+  `event.userName` all report the **UUID**, never the email. The email appears in the `email`
+  attribute, in the **ID token's** `email` claim (AWS omits user attributes such as `email` from the
+  **access** token), and in `event.request.userAttributes.email`.
+- Sign-in resolves by the **current** email alias **or** the UUID. `SECRET_HASH` is validated against
+  the exact `USERNAME` value sent: `Base64(HMAC-SHA256(USERNAME + clientId, clientSecret))`.
+- In `CUSTOM_AUTH` / SRP flows, `ChallengeParameters.USERNAME` echoes the **UUID** (as real AWS does);
+  the `RespondToAuthChallenge` `SECRET_HASH` is validated against the `USERNAME` sent that round.
+- `AdminUpdateUserAttributes` can change the email; afterwards sign-in works with the new email and
+  fails with the old one, while `Username`/`sub` stay fixed. A duplicate email is rejected with
+  `UsernameExistsException` (on create) or `AliasExistsException` (on update).
+
+Pools **without** `UsernameAttributes` (classic pools, and pools using `AliasAttributes`) keep the
+literal `Username` you supply, unchanged.
+
 ## Configuration
 
 | Variable                         | Default | Description                   |
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -134,20 +134,35 @@ Floci follows real Cognito semantics for pools created with `UsernameAttributes`
 - `AdminCreateUser`/`SignUp` accept the email (or phone number) as the sign-in value, but the
   **canonical `Username` is an auto-generated, immutable UUID equal to `sub`**. The supplied email is
   stored as a mutable alias attribute.
-- `ListUsers`, `AdminGetUser`, tokens (`sub`/`username`/`cognito:username`) and Lambda trigger
-  `event.userName` all report the **UUID**, never the email. The email appears in the `email`
-  attribute, in the **ID token's** `email` claim (AWS omits user attributes such as `email` from the
-  **access** token), and in `event.request.userAttributes.email`.
+- `ListUsers`, `AdminGetUser` and Lambda trigger `event.userName` all report the **UUID**, never the
+  email; the email lives in the `email` attribute and in `event.request.userAttributes.email`.
 - Sign-in resolves by the **current** email alias **or** the UUID. `SECRET_HASH` is validated against
   the exact `USERNAME` value sent: `Base64(HMAC-SHA256(USERNAME + clientId, clientSecret))`.
 - In `CUSTOM_AUTH` / SRP flows, `ChallengeParameters.USERNAME` echoes the **UUID** (as real AWS does);
   the `RespondToAuthChallenge` `SECRET_HASH` is validated against the `USERNAME` sent that round.
 - `AdminUpdateUserAttributes` can change the email; afterwards sign-in works with the new email and
-  fails with the old one, while `Username`/`sub` stay fixed. A duplicate email is rejected with
-  `UsernameExistsException` (on create) or `AliasExistsException` (on update).
+  fails with the old one, while `Username`/`sub` stay fixed.
+- Duplicate email/phone on `AdminCreateUser`: `UsernameExistsException` when the incoming alias is
+  unverified, `AliasExistsException` when it is verified (`email_verified=true`); pass
+  `ForceAliasCreation=true` to migrate a verified alias off the previous owner. On
+  `AdminUpdateUserAttributes`, changing to an in-use alias throws `AliasExistsException`.
 
 Pools **without** `UsernameAttributes` (classic pools, and pools using `AliasAttributes`) keep the
 literal `Username` you supply, unchanged.
+
+### Token claims
+
+Floci mirrors AWS's access-token / ID-token split:
+
+- **Access token:** `sub`, `username` (the UUID), `scope` (`aws.cognito.signin.user.admin` for API
+  sign-in), `client_id`, `cognito:groups`, `jti`/`origin_jti`. It does **not** carry `cognito:username`
+  or user attributes like `email`.
+- **ID token:** `sub`, `cognito:username`, `aud`, and readable user attributes (`email`,
+  `email_verified`, `phone_number`, `custom:*`, ...). Attribute claims are filtered by the app client's
+  `ReadAttributes` (an unset/empty list means all attributes are readable).
+
+Not-found errors (`ResourceNotFoundException`, `UserNotFoundException`) return HTTP `400`, matching the
+Cognito JSON protocol.
 
 ## Configuration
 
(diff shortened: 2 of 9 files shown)


AbhigyaKrishna:
need this merged as we are using floci for our uat

src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
```diff
@@ -481,7 +481,8 @@ private Map<String, Object> handleCustomChallenge(UserPool pool, UserPoolClient
         if (answer == null || answer.isBlank()) {
             throw new AwsException("InvalidParameterException", "ANSWER is required", 400);
         }
-        validateSecretHash(client, responses, state.username);
+
+        validateSecretHash(client, responses, responses.getOrDefault("USERNAME", state.username));
```
abanna: Switching to the client-supplied `USERNAME` here is right in itself — AWS does compute `SECRET_HASH` over the exact value the caller sent, which is the whole point of the alias work in this PR. But nothing now ties that value back to the session, and the line below still resolves the user from the session:

```java
validateSecretHash(client, responses, responses.getOrDefault("USERNAME", state.username));

CognitoUser user = service.adminGetUser(pool.getId(), state.username);
```

The session is checked against the pool and the client at lines 472–474, but never against `USERNAME`. So the hash is verified for whatever user the caller names, while the challenge proceeds for `state.username` — the two are allowed to disagree. That drops the property `SECRET_HASH` exists to provide: it stops being a per-user binding and becomes a bare "do you hold the client secret" check for this call.

I don't think it's a privilege escalation on its own, since computing any valid hash still requires the client secret. But real Cognito requires the `USERNAME` in `ChallengeResponses` to be the session's user, so a caller that gets this wrong fails against AWS and passes here — which is the direction that bites people in a local emulator.

Resolving the supplied value the same way sign-in does and comparing would keep both properties:

```java
String supplied = responses.getOrDefault("USERNAME", state.username);
validateSecretHash(client, responses, supplied);
CognitoUser user = service.adminGetUser(pool.getId(), state.username);
if (!user.getUsername().equals(resolveToCanonicalUsername(pool, supplied))) {
    throw new AwsException("NotAuthorizedException", "Invalid session for the user.", 400);
}
```

That also keeps the alias case working, since the caller may legitimately send the email while `state.username` is the UUID — which I assume is exactly why this changed.

Unrelated to the above, in `validateSecretHash` itself at line 307:

```java
if (!expected.equals(provided)) {
```

`String.equals` short-circuits on the first differing byte, so this leaks MAC comparison timing. Worth `MessageDigest.isEqual(expected.getBytes(UTF_8), provided.getBytes(UTF_8))` — that's what the S3 presigned SigV4 path in #1842 uses for the same comparison, so it'd be consistent with the rest of the codebase too.

AbhigyaKrishna: Both fixed in 591f5ddd.

**USERNAME/session binding.** Took the suggested shape — resolve the supplied value through `adminGetUser` (same resolution sign-in uses, so an alias still resolves) and require it to match the session user:

```java
String suppliedUsername = responses.getOrDefault("USERNAME", state.username);
validateSecretHash(client, responses, suppliedUsername);

CognitoUser user = service.adminGetUser(pool.getId(), state.username);
if (!user.getUsername().equals(resolveToCanonicalUsername(pool, suppliedUsername))) {
    throw new AwsException("NotAuthorizedException", "Invalid session for the user.", 400);
}
```

`resolveToCanonicalUsername` swallows the lookup failure to `null` (logged at debug) so an unknown USERNAME comes back as `NotAuthorizedException` rather than leaking `UserNotFoundException` — the mismatch and the nonexistent-user case are indistinguishable to the caller, as in AWS.

Checked the other `validateSecretHash` call sites while in there: `USER_PASSWORD_AUTH`, `USER_SRP_AUTH` and `PASSWORD_VERIFIER` all resolve the user from the same value they hash over, so the divergence was confined to `CUSTOM_CHALLENGE`.

**Timing.** Swapped `String.equals` for `MessageDigest.isEqual` over the UTF-8 bytes, consistent with `CognitoSrpHelper.verifySignature` (which already uses it for the SRP claim) and the #1842 SigV4 path.

**Test.** New `customAuthRejectsChallengeResponseUsernameThatIsNotTheSessionUser` in `CognitoLambdaTriggersTest`: two users in an alias pool, session opened for one, round 2 sent with the *other* user's UUID and a `SECRET_HASH` correctly computed over it — now `NotAuthorizedException`. Verified it fails on the pre-fix code ("expected AwsException to be thrown, but nothing was thrown"). The same test then replays round 2 with the session user's email alias and a hash over the email, which still issues tokens — that's the case the original change existed to enable, so it's pinned now. 298 Cognito tests green.

Docs updated to state the binding requirement alongside the existing SECRET_HASH note.

pgermosen:
Checked this beyond just the conflict status, since it's been open a while.

**The Cognito logic itself looks solid.** I independently verified every open review finding on this PR is actually resolved in the current code, not just claimed:
- RESEND with a canonical UUID no longer hits `aliasAttributeForValue` (it's now gated on `existing == null`, so a direct key lookup by UUID short-circuits it)
- Migration no longer trusts a Lambda-supplied `sub` — `canonicalUsername` is now always server-derived and overwrites it
- The email-format check is a real regex now, not `contains("@")`
- `AliasExistsException` guard and @abanna's session-binding concern both have dedicated fix commits with tests

**What's actually blocking merge is staleness, and it's more than the one conflict marker suggests:**
- `CognitoServiceTest.java` has a real ~540-line conflict (lines 2400–2940) — this PR and `main` both independently added large test blocks at the same insertion point (main's addition is the `ConfirmSignUpVerifiedAttributes` nested class for #1654). That's a genuine manual-reconciliation job, not a one-liner.
- The branch is stale enough that it doesn't even compile standalone anymore: `Ec2PublicIpOnLaunchTest.java` (unrelated to this PR) fails at test-compile because `Ec2Service`'s constructor gained a new parameter on `main` after this branch's last push on 2026-08-06.

@AbhigyaKrishna this needs a real rebase onto current `main`, not just a conflict-marker fix — worth doing that first, then a full test run to make sure everything still holds.

pgermosen (commented):
@AbhigyaKrishna — thank you for this, and for the depth of the write-up. I went through the change against the AWS documentation rather than against the description, and the substance holds up well. One question at the end that I would like your view on before this goes in, and it is a versioning question rather than a code one.

## Both of @abanna's findings are closed

The `USERNAME`/session binding now resolves the supplied value through the same path sign-in uses and compares it to the session user, so the alias case still works while the two can no longer disagree. Resolving an unknown value to `null` and surfacing the mismatch as `NotAuthorizedException` — rather than letting `UserNotFoundException` escape — is the right call; it keeps the mismatch and the nonexistent-user case indistinguishable to the caller.

The constant-time comparison is in as well. Worth recording explicitly for whoever reads this later: `MessageDigest.isEqual` and `String.equals` differ only in timing and never in result, so no functional check can tell them apart. That is inherent to the property, not a coverage gap, and I would not want it filed as one later.

## What I confirmed against the AWS documentation

I checked the load-bearing claims at source rather than taking the description's word for them.

**The UUID identity model.** [Working with user attributes](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html) states it outright:

> The `SignUp` API populates the `username` attribute with a UUID for your user. This UUID has the same value as the `sub` claim in the user identity token.

and, for the rejection path:

> If the `username` string format isn't in email address or phone number format, the `SignUp` API returns an exception.

and the immutability:

> After you create a user, you can't change the value of the `username` attribute.

`AdminCreateUser` carries the same rule from the other direction — *"You can only provide a value if usernames are a valid sign-in attribute for your user pool. If your user pool only supports phone numbers or email addresses as sign-in attributes, Amazon Cognito automatically generates a username value."* — which also supports minting your own rather than honouring a client-supplied `sub`. Your E.164 handling matches the documented format rule too (leading `+`, country code, digits only).

**The duplicate-alias split is the one I nearly flagged, and you are right.** The developer guide's alias-versus-username comparison table reads as though a `UsernameAttributes` pool should simply produce `UsernameExistsException`, which looks like it contradicts your `AliasExistsException` behaviour. The [AdminCreateUser reference](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminCreateUser.html) settles it in your favour:

> This parameter is used only if the `phone_number_verified` or `email_verified` attribute is set to `True`. Otherwise, it is ignored. If this parameter is set to `True` … this request migrates the alias from the previous user to the newly-created user. … If this parameter is set to `False`, the API throws an `AliasExistsException` error if the alias already exists.

That is exactly your split — `UsernameExists` for the unverified duplicate, `AliasExists` for the verified one, `ForceAliasCreation` migrating it off the previous owner. The guide's table describes the `SignUp` path; `AdminCreateUser` has its own documented flow. Worth knowing that a reviewer coming to this cold may well raise the same objection I almost did.

**The not-found status change is documented, not inferred.** The `AdminCreateUser` error list gives **HTTP Status Code: 400** for both `ResourceNotFoundException` and `UserNotFoundException`. Independently, no Cognito error shape in the service model declares a status code at all — all 41 of them inherit the JSON protocol default. So 400 is right on both counts, and the previous 404 had no basis either way.

## Two things I checked that the description does not claim

**The gating genuinely protects classic pools.** I forced `usesAliasUsernames` to return true unconditionally in a local build, and the existing behaviour of pools without `UsernameAttributes` falls apart comprehensively. So "classic and `AliasAttributes` pools are unchanged" is not merely asserted — it is held in place.

**Already-persisted users survive the upgrade.** `adminGetUser` does the direct key lookup first and then the alias scan, collecting into a map keyed on the resolved username. A user persisted before this change — keyed by email — is still found when the email is supplied, and because both paths land on the same key the dedupe prevents a spurious *"Multiple users found"*. Such users keep reporting `Username == email` rather than a UUID, which I think is the right conservative choice: rewriting persisted identities on upgrade would be worse. It may be worth a sentence in the description saying so, since "what happens to my existing local users" is the first thing an operator will ask.

## The question: is this a `fix` or a `fix!`?

This is the only thing I would like settled before merge.

The description flags the behaviour change clearly — for `UsernameAttributes` pools, `Username` returned by `AdminGetUser` and `ListUsers` changes from the email to a UUID, and consumers who relied on the old value must read the email attribute instead. That is exactly the right disclosure. But the commit subject is `fix(cognito): …` and the type checkbox is "Bug fix" rather than "Breaking change", so the release tooling will treat this as a patch.

There is a real argument for leaving it as `fix`: Floci is an emulator, the old behaviour was a divergence from AWS, and converging on AWS is arguably always a fix rather than a break. There is an equally real argument for `fix!`: someone upgrading a patch version will find code that reads `Username` as an email address silently receiving a UUID, with no signal from the version bump that anything moved.

I do not think this is mine to decide unilaterally, and I would rather ask than quietly pick. Which way do you read it? If you think `fix!` is warranted, it is a subject-line change. If you would rather keep `fix`, I am content with that too provided the behaviour-change note stays as prominent as it is now — I will make sure it reaches the release notes either way.

Everything else here looks ready. Thank you for the patience on this one; it has been open a long time and the quality has held up through it.

AbhigyaKrishna:
@pgermosen — thank you for the careful pass, and for checking the load-bearing claims against the AWS docs rather than the description. Settling the one open question:

**Keeping it as `fix`.** The reasoning you laid out for `fix` is the one I land on: Floci is an emulator, the pre-change `Username == email` for `UsernameAttributes` pools was a divergence from AWS, and converging on documented AWS behaviour is a fix rather than a break. Treating "we now match AWS" as a breaking change would, taken to its conclusion, make almost every accuracy fix a major bump, which isn't a signal that stays meaningful.

The trade-off you named is real, though, so the mitigation is that the behaviour-change note stays exactly as prominent as it is now — the PR description already calls out that, for `UsernameAttributes` pools, `Username` from `AdminGetUser`/`ListUsers` changes from the email to a UUID and consumers relying on the old value must read the email attribute instead. I'll make sure that line reaches the release notes verbatim.

I've also added the sentence you suggested about existing local users: users persisted before this change stay keyed by email and keep reporting `Username == email` rather than being rewritten to a UUID on upgrade — the conservative choice, and the first thing an operator will ask.

Everything else you confirmed is already in and pushed; no code changes fell out of this review. Appreciate the patience on the long-open one.

pgermosen (approved):
@AbhigyaKrishna — the versioning question is settled, so I went back over this against current `main` rather than re-reading what I already checked in August. Everything that was blocking is gone, and I verified the load-bearing claims against a running instance this time instead of against the code.

## The staleness is resolved

That was the real blocker last round, and the rebase closed it. Merging current `main` (`5c45c146`) into this branch is clean — including `CognitoServiceTest.java`, where the ~540-line conflict was. Main's `ConfirmSignUpVerifiedAttributes` block and your additions now coexist.

**486 tests, 0 failures** on that merge:

```
CognitoServiceTest (+ ConfirmSignUpVerifiedAttributes)  155
CloudFormationIntegrationTest                           142
CognitoIntegrationTest                                   71
CognitoLambdaTriggersTest                                29
CognitoOAuthTokenIntegrationTest                         21
CognitoJsonHandlerTest                                   15
CognitoAliasUsernameIntegrationTest                       7
… plus the rest of the Cognito suites
```

Worth flagging for whoever merges: the green CI on this PR is from a run on **2026-08-19**, and `main` has moved a long way since — including the CI restructure into a four-way shard matrix. The run above is on the merge with `main` as of today.

The only Cognito commit to land on `main` since your branch point is `5ed3a8ca` (DescribeUserPool serializing `DeviceConfiguration`/`UserPoolAddOns`/`EmailConfiguration`), which is orthogonal to the identity model. Your own `CognitoJsonHandler` change is one line — the `ForceAliasCreation` passthrough — so there is no semantic overlap hiding behind the clean auto-merge.

## The identity model, exercised rather than read

Everything below is from a live instance on a `--username-attributes email` pool.

```
admin-create-user --username hector@example.com
  → Username  45c7a57d-a547-4efd-a480-0a0099616f4f
    sub       45c7a57d-a547-4efd-a480-0a0099616f4f      # equal, as documented
    email     hector@example.com                        # alias attribute

admin-get-user --username <uuid>                → 45c7a57d-…
admin-get-user --username hector@example.com    → 45c7a57d-…    # alias resolves
list-users                                      → 45c7a57d-…    # never the email
admin-create-user --username notanemail         → InvalidParameterException
```

The claim split is exactly as described:

```
ACCESS  sub, username=<uuid>, scope=aws.cognito.signin.user.admin, client_id, jti, origin_jti, …
        no email, no cognito:username
ID      sub, cognito:username=<uuid>, aud, email=hector@example.com, email_verified, …
```

Rebinding keeps identity fixed and retires the old alias:

```
admin-update-user-attributes email=newmail@example.com
  → Username/sub unchanged, email=newmail@example.com
  sign-in newmail@example.com   → Bearer
  sign-in hector@example.com    → UserNotFoundException
```

And all four duplicate-alias branches behave as `AdminCreateUser` documents:

```
duplicate verified alias                → AliasExistsException
duplicate unverified alias              → UsernameExistsException
duplicate verified + ForceAliasCreation → new UUID; alias migrates, previous owner keeps its
                                          identity and loses the email attribute
```

**The gating holds under live traffic too**, not just under the forced-flag experiment I ran last time: a classic pool returns `Username = alice`, and an `AliasAttributes` pool returns `Username = bob`. Both literal, both untouched.

## The not-found change is complete and contained

`grep -c ", 404)"` across every file in `services/cognito/` now returns **0**. The change is per-call-site inside `CognitoService`, not a shared mapper, so nothing outside Cognito moves. Live: `AdminGetUser` for an unknown user returns 400.

That also reconfirms the doc basis from last round — `AliasExistsException`, like every other Cognito error shape, declares no `httpStatusCode` at all, so 400 is the JSON-protocol default rather than an invention.

## The upgrade path, verified end to end

This is the claim I asked you to document, so I ran it rather than trusting it. I built `main`'s Cognito, created a user on a `UsernameAttributes` pool under `FLOCI_STORAGE_MODE=persistent`, then restarted **the same data directory** on this branch.

```
pre-upgrade (main):   Username = legacy@example.com,  sub = de680dd6-…
                      store key: …/us-west-2_d464c327d::legacy@example.com

post-upgrade (this):  admin-get-user legacy@example.com  → legacy@example.com
                      admin-get-user de680dd6-…         → legacy@example.com
                      list-users                        → legacy@example.com   (once)
                      sign-in legacy@example.com        → Bearer
```

No rewrite, no lockout, no spurious *"Multiple users found"* — and the old `sub` resolves as well, so neither identifier a script might already hold goes dead. A new user created on that same pool afterwards gets a UUID and the two coexist cleanly:

```
list-users → legacy@example.com    8ddfec63-27bb-43dd-989b-f4a85b36f1e9
```

The conservative choice was the right one.

---

## One thing that falls out of it

`ForceAliasCreation=true` is inert against a pre-upgrade, email-keyed user. Same pool, same command, back to back:

```
--username legacy@example.com --force-alias-creation  → UsernameExistsException
--username fresh@example.com  --force-alias-creation  → 5c835757-…   (alias migrated)
```

The cause is benign: the legacy user's *store key* is the email, so the direct-key duplicate check fires before the alias check ever runs. Real AWS can't reach this state, so there is no correct behaviour being violated — it is a consequence of deliberately not rewriting persisted identities, which I still think is the right call.

The only gap is that the docs don't say so. The new section states plainly that `ForceAliasCreation=true` migrates a verified alias, with no carve-out, and an operator who hits this on an existing local pool has nothing to read. One sentence — something like *"users created before this change stay keyed by their email; `ForceAliasCreation` does not apply to them, and recreating the user is the way to move to the UUID model"* — closes it.

That is a docs line, not a code change, and after six weeks open I am not holding the PR for it.

The docs section itself is genuinely good, by the way: it documents the `SECRET_HASH` derivation, the `ReadAttributes` filtering, and the classic-pool carve-out — the three things someone debugging this at 2am actually needs.

---

## Approving — two follow-ups, please

This is approved. Both of these are small and neither needs to block the merge, so take them as a follow-up PR rather than another round here:

1. **The `ForceAliasCreation` carve-out above.** One sentence in the `Sign-in Identifiers` section of `docs/services/cognito.md`, saying that users created before this change stay keyed by their email, that `ForceAliasCreation` does not apply to them, and that recreating the user is the way to move them onto the UUID model.
2. **SDK-level coverage for this pool type.** Your description already names this as a recommended follow-up and I agree it is the right one. There *is* Cognito coverage in `compatibility-tests/` — three Java tests plus several Node ones — but none of it creates a `UsernameAttributes` pool. The only `username_attributes = ["email"]` anywhere under `compatibility-tests/` is in the Terraform and OpenTofu fixtures (`compat-terraform/main.tf:162`, `compat-opentofu/main.tf:274`), and those create the pool without any users, so nothing in the compat suite exercises the identity model through a real SDK at all. Every assertion in this PR runs through RestAssured against Floci's own wire path. `AdminGetUser().username()` coming back as a UUID, the `ReadAttributes` filtering on the ID token, and the `AliasExists` / `UsernameExists` split are all things a typed client exercises differently from hand-written JSON.

   The flip side of that is reassuring, and I checked it because the stale CI run made it worth checking: since those fixtures declare no `aws_cognito_user` resources, this change cannot move the Terraform/OpenTofu compat results either.

Thank you for the patience on this one. It has been open since 12 July, it survived a rebase over a moving `main`, and the quality held up through all of it — including the parts I went looking to disprove.

PR #1853, opened 2026-07-13
fix(cloudformation): inline policy names are no longer global (can be reused in different stacks)

### LINKED ISSUE #1531: CloudFormation: IAM inline-policy physical name not stack-scoped → cross-stack collision ("Policy … already exists")
## Environment
- floci: `floci/floci:nightly` (**nightly-06232026**)
- Docker: 29.5.2
- IaC: AWS CDK (`aws-cdk-lib`), `cdk deploy` against floci (`AWS_ENDPOINT_URL=http://localhost.floci.io:4566`, account `000000000000`, region `ap-south-1`)

> 🙏 First: **nightly-06232026 fixed** the earlier ELBv2 "subnet does not exist" gap — multi-AZ ELB stacks now deploy. Thank you! The two issues below are what remains when deploying a multi-stack CDK app.

## Summary
When two CloudFormation stacks in the same account each create an IAM **inline policy** whose CDK logical id is identical (two stacks reusing the same construct sub-tree both emit `IamTaskRoleDefaultPolicy<hash>`), floci derives the IAM policy's **physical name from the logical id with no per-stack uniqueness**. The second stack to deploy fails because the policy already exists.

## Repro
1. Two CDK stacks, each with an ECS `TaskRole` carrying its CDK-auto-generated inline `DefaultPolicy` (logical id `…DefaultPolicy<hash>`); the construct path yields the **same** hash in both stacks.
2. `cdk deploy StackA` → ✅ creates `policy/IamTaskRoleDefaultPolicy54CEC625`.
3. `cdk deploy StackB` → ❌:
   ```
   AWS::IAM::Policy  Iam/TaskRole/DefaultPolicy  CREATE_FAILED
   Policy arn:aws:iam::000000000000:policy/IamTaskRoleDefaultPolicy54CEC625 already exists.
   ```
   → `ROLLBACK_COMPLETE`.

## Expected
Real CloudFormation/IAM assigns each stack's auto-named policy a **unique physical name** (CFN appends a stack-scoped suffix), so both stacks deploy. floci should scope the generated physical name per stack (or honor CFN physical-id generation) instead of using the bare logical id.

## Actual
Cross-stack physical-name collision → the second stack fails to deploy.

## Impact
Any multi-stack app reusing a construct pattern (shared ECS service / task role) across stacks cannot deploy all stacks to the same floci account — stacks that deploy fine individually collide when combined.

CoreOxide:
## Summary

Fixes CloudFormation AWS::IAM::Policy handling so inline policies can reuse the same name across stacks without collisions.
Also cleans up removed attachments during updates and validates all principals before applying policy changes.

Closes #1531

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

AWS::IAM::Policy was incorrectly persisted as a managed IAM policy, causing EntityAlreadyExists errors when separate stacks used the same inline-policy name.
The implementation now uses IAM inline-policy operations per role, user, and group.

Regression verified with CloudFormationIntegrationTest. Repository SDK compatibility tests use AWS SDK for Java 2.44.14; local AWS CLI version is 2.35.9.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -95,6 +95,8 @@ import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.TimeoutException;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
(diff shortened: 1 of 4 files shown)


hectorvent:
Thanks for this, and sorry it took a while to decide. We had two PRs fixing #1531 and we are going with #1800, which lands the same model and also cleans up the surrounding attach and delete paths.

Your PR does cover one case #1800 misses: an update that drops a principal from `Roles` should remove the inline policy from the old one, and your test pins exactly that. We are asking for it on #1800, so this was genuinely useful.

Closing to avoid two implementations of the same fix. If you want to pick up the update semantics as a follow up on top of #1800, that would be very welcome.

CoreOxide:
will do!

PR #1855, opened 2026-07-13
fix(ec2): create /run/sshd before starting sshd

smashyalts:
Added command to create /run/sshd directory before starting sshd.

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

sshd didnt have /run/sshd folder so when using
aws ec2 run-instances --image-id ami-ubuntu2204 --instance-type t3.micro --key-name test

sshd would never start requiring manual container access to make folder and starting sshd
## Checklist


src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
```diff
@@ -443,6 +443,7 @@ private void startSshd(String containerId, String instanceId) {
                     "fi"}, 120);
             // Generate host keys
             execInContainer(containerId, new String[]{"ssh-keygen", "-A"}, 10);
+            execInContainer(containerId, new String[]{"mkdir", "-p", "/run/sshd"}, 5);
```
abanna, line 446: This is the right fix — modern OpenSSH refuses to start without its privilege-separation directory, and `/run` is a fresh tmpfs in most container images, so `/run/sshd` genuinely isn't there.

Two things while you're in here.

**CI is red on Conventional Commits**, and it's the commit messages rather than the PR title. Both commits on the branch fail:

```
Create /run/sshd directory for sshd
Update src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
```

The check walks every commit and wants `<type>[scope]: <description>`. Squashing the two into something like `fix(ec2): create /run/sshd before starting sshd` should clear it. (The PR title reads `fix:(ec2)` with the colon and scope transposed — worth fixing to `fix(ec2):` too, since that's what lands in the squashed message.)

**Worth asking why this was silent.** `execInContainer` throws away the result:

```java
private void execInContainer(String containerId, String[] cmd, int timeoutSeconds) throws Exception {
    execInContainerForResult(containerId, cmd, timeoutSeconds);   // exit code discarded
}
```

so when `/usr/sbin/sshd` exited with `Missing privilege separation directory: /run/sshd`, line 449 still logged `Started sshd in EC2 instance {0}` and the instance came up looking healthy with no SSH. Anything else that breaks sshd startup — a failed `ssh-keygen -A`, a package install that silently no-ops because none of `dnf`/`apt-get`/`apk` exist — fails exactly the same silent way.

Would it be worth checking the exit status on the sshd start specifically, so the warning at line 451 actually fires?

```java
ContainerExecResult result = execInContainerForResult(containerId, new String[]{"/usr/sbin/sshd"}, 5);
if (result.exitCode() != 0) {
    LOG.warnv("sshd failed to start in EC2 instance {0}: {1}", instanceId, result.output());
    return;
}
LOG.infov("Started sshd in EC2 instance {0}", instanceId);
```

Entirely fine as a follow-up rather than in this PR — but it's the difference between finding the next one of these immediately and finding it the way you found this one.

hectorvent:
Thank you for the fix, and abanna's review above covers it exactly right, so just seconding it as the maintainer.

**(blocking)** The Conventional Commits check lints every commit on the branch, so the squash and reword abanna suggested (`fix(ec2): create /run/sshd before starting sshd`) is what clears CI; fixing the transposed PR title in the same pass matters because the title becomes the squash merge message.

**(follow-up, separate PR)** The silent-failure point abanna raised is real, `execInContainer` discards exit codes, so a failed sshd start still logs success. Checking the exit status there is worth its own small PR.

One squash and this merges.

smashyalts:
> Thank you for the fix, and abanna's review above covers it exactly right, so just seconding it as the maintainer.
> 
> **(blocking)** The Conventional Commits check lints every commit on the branch, so the squash and reword abanna suggested (`fix(ec2): create /run/sshd before starting sshd`) is what clears CI; fixing the transposed PR title in the same pass matters because the title becomes the squash merge message.
> 
> **(follow-up, separate PR)** The silent-failure point abanna raised is real, `execInContainer` discards exit codes, so a failed sshd start still logs success. Checking the exit status there is worth its own small PR.
> 
> One squash and this merges.
> 

will be fixing as soon as possible, not on my pc atm and its not fun to do this on mobile

abanna (changes requested):
Thank you for the focused `/run/sshd` fix—the directory creation is the right correction for modern OpenSSH, and all 17 existing EC2 container-manager tests pass. Before merge, please:

- rename the PR to `fix(ec2): create /run/sshd before starting sshd`;
- squash or rewrite the two commits into Conventional Commit form;
- remove the Greptile `Co-Authored-By` trailer, since the project limits attribution to humans;
- add a focused regression test proving `/run/sshd` is created before `sshd` is started.

The separate suggestion to check the `sshd` exit status can remain a follow-up. Thank you for keeping this fix small.

abanna:
Thanks for the thoughtful follow-up — this looks good from my side now. The PR title and squashed commit follow the project's Conventional Commits format, the non-human co-author trailer is gone, and the new regression test directly verifies that `/run/sshd` is created before `sshd` starts.

I reran `Ec2ContainerManagerTest` and all 18 tests pass. I also confirmed that the new test fails with the expected assertion when the production `mkdir` step is removed, so it provides meaningful regression coverage.

The separate `sshd` exit-status handling can remain a follow-up as discussed. I appreciate you keeping this fix focused and working through the requested changes.

hectorvent:
Thanks for turning that around, and the squash did it: the Conventional Commits check passes and CI is fully green now that the gated runs are released.

@abanna has already validated the substance and I agree, so just the one thing they could not have seen.

**(blocking)** The branch conflicts with main now, in `Ec2ContainerManager.java` and its test. It is mechanical: #2245 rewrote `startSshd` around your change, but nothing you rely on moved, and `execInContainer` is still there. After the rebase the `mkdir` wants to sit just before the `sshd` call, after the `ssh-keygen -A` check, so it does not run on the paths that bail out early.

Worth knowing your fix is still needed: main creates `/run/sshd` nowhere, and #2245 means sshd now starts for instances launched without a key pair too, so more instances hit this than when you opened it.

pgermosen:
Rebased this onto current main and pushed the resolution — conflict is cleared, `mergeable` is back to `MERGEABLE`.

The conflict was exactly what @hectorvent flagged: #2351 (merged after that comment) further reshaped `startSshd` into a loop trying multiple absolute paths (`ALLOWED_SSHD_PATHS`), on top of #2245's exit-code checking. Re-applied the fix to fit: `mkdir -p /run/sshd` now runs once, with its own exit-code check matching the surrounding style, right after the `ssh-keygen -A` check succeeds and before the path-loop starts — so it still never runs on the early-bail paths, same intent as originally written, just adapted to the current shape.

Also kept both of main's newer tests (`launchWithoutKeyPairStillStartsSshd`, `launchWhenSshdInstallFails_doesNotAttemptKeygenOrStart`) alongside this PR's own `launchCreatesSshdPrivilegeSeparationDirectoryBeforeStartingSshd` — the test-file conflict was just an insertion-point collision, not overlapping logic.

Verified:
- Reverted just the `mkdir` step and confirmed the regression test fails with exactly the expected assertion, then restored and confirmed it passes.
- Full `Ec2ContainerManagerTest`: 21/21. Full `ec2` package: 427/427.
- Checked the underlying claim against real sources: the [Debian bug tracker](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=864190) confirms `/run/sshd` is normally created by a systemd-tmpfiles rule that never runs in a container with no systemd, and a [Red Hat KB article](https://access.redhat.com/solutions/743553) confirms sshd's absolute-path requirement (the reason for the `ALLOWED_SSHD_PATHS` loop this now sits inside) — both hold up.
- `make docs-check` clean, no CODEOWNERS-protected paths touched.

@smashyalts nothing needed from you — should be ready once CI finishes on the new commit.

pgermosen (approved):
Rebased and re-verified — reverted the mkdir step and confirmed the regression test genuinely fails without it, then restored and confirmed 21/21 pass (427/427 across the ec2 package). Checked the underlying claim against the Debian bug tracker and a Red Hat KB article, both confirm the premise. Conflict is resolved cleanly against current main.

Approving.

hectorvent:
Thank you both. @smashyalts, the fix held up through three rewrites of `startSshd` around it, and the regression test that came out of review now pins the mkdir-before-sshd ordering permanently, so this cannot quietly regress. For anyone landing here later: `/run/sshd` is normally created by a systemd tmpfiles rule that never runs inside a container, which is why sshd refused to start.

@pgermosen, thanks for carrying the rebase, for revert-verifying the test still catches the bug in its new home, and for keeping the fix commit under smashyalts' authorship.

Glad to see this merged, and sorry it took this long from your first push to landing, this one earned it.

PR #1856, opened 2026-07-13
feat(sns): deliver topic broadcasts to application-protocol platform endpoints

AbhigyaKrishna:
## Summary

Closes a gap in Floci's SNS mobile-push mock: **topic broadcast to platform endpoints was never delivered or captured**.

`deliverMessage(...)` switched on the subscription protocol with cases for `sqs`, `lambda`, `http`/`https`, `email`, and `sms` — but had **no `application` case**. A subscription created with `Protocol="application"` is auto-confirmed (correctly excluded from `PENDING_CONFIRMATION_PROTOCOLS`) and enters the delivery loop, but then fell through with no delivery and no `recordPushNotification`. So publishing to a topic returned a message id while nothing reached the subscribed platform endpoints, and the retrospection API (`GET /_aws/sns/push-notifications`) returned empty. Per-user (direct-to-endpoint) push already worked; only the broadcast half was missing.

This PR:

- Adds an `application`-protocol branch to `deliverMessage` that resolves the subscribed endpoint's owning platform application and captures the push — so broadcast pushes surface in the **same** retrospection API, keyed by `EndpointArn`.
- Threads `messageStructure` through the delivery boundary. It was previously dropped between `publish` and `deliverMessage`, so platform-specific payload keys (`GCM`/`FCM`/`APNS`) now resolve **per endpoint** on the fan-out path, exactly as on the direct path.
- Factors the "resolve payload + record push" tail of `publishToEndpoint` into a shared `capturePushForEndpoint(...)` helper used by both the direct-publish and fan-out paths, so capture is identical.
- Skips a disabled endpoint (or disabled/missing platform app) in the fan-out rather than failing the whole publish — matching AWS's async delivery model.

Scope note: the shared `deliverMessage` is also used by `publishBatch`, so batch topic publishes now honor `application` fan-out too (`MessageStructure` is extracted per batch entry; a null value simply returns the raw message). This keeps the two publish paths consistent.

The `CreatePlatformEndpoint` "already exists with the same Token, but different attributes" `InvalidParameter` message and the `EndpointDisabled` code on direct publish to a disabled endpoint are unchanged and remain covered by existing tests.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Behaviour verified against AWS's official SNS documentation:

- **Fan-out to platform endpoints** — real SNS delivers a topic message to every subscribed platform endpoint exactly as if published to each directly (same payload resolution, same `Enabled` gating). See [Sending mobile push notifications (application subscriber)](https://docs.aws.amazon.com/sns/latest/dg/sns-mobile-application-as-subscriber.html).
- **Per-endpoint platform payload** over a topic uses the JSON envelope's `default` + platform keys (`APNS`/`GCM`/`FCM`). See [Publishing platform-specific payloads](https://docs.aws.amazon.com/sns/latest/dg/sns-send-custom-platform-specific-payloads-mobile-devices.html).
- **Disabled endpoint** yields an async `EndpointDisabled` delivery-failure event, never failing the topic publish or blocking other subscribers → modelled as "skip that endpoint, continue the rest". See [SNS management of FCM endpoints](https://docs.aws.amazon.com/sns/latest/dg/sns-fcm-endpoint-management.html).
- `application`-protocol subscriptions are auto-confirmed (no `ConfirmSubscription`), like SQS/Lambda — consistent with existing `PENDING_CONFIRMATION_PROTOCOLS`.

## Checklist

- [x] `./mvnw test` passes locally (full `Sns*` suite: 136 tests, 0 failures)
- [x] New or updated integration test added (3 new tests in `SnsMobilePushTest` covering multi-device capture, disabled-endpoint skip, and `default`-key fallback over fan-out)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
@@ -404,6 +404,8 @@ public class SnsService implements Resettable {
         Topic topic = topicStore.get(topicStoreKey)
                 .orElseThrow(() -> new AwsException("NotFound", "Topic does not exist.", 404));
 
+        validateTopicMessageStructure(message, messageStructure);
+
         boolean isFifo = "true".equals(topic.getAttributes().get("FifoTopic"));
         String dedupId = messageDeduplicationId;
         if (isFifo) {
@@ -683,6 +685,35 @@ public class SnsService implements Resettable {
                 "Invalid parameter: Message Reason: Messages must have a '" + platform
                         + "' or 'default' key.",
                 400);
+    }
+
+    /**
+     * Validates {@code MessageStructure="json"} on a topic {@code Publish} the way the real SNS
+     * API does: synchronously, before any fan-out. The message must be a JSON object carrying a
+     * top-level {@code default} entry. Validating here (rather than lazily inside per-subscriber
+     * delivery) means a malformed envelope surfaces to the caller as {@code InvalidParameter}
+     * instead of being silently swallowed while the {@code publish} call still reports success.
+     */
+    private void validateTopicMessageStructure(String message, String messageStructure) {
+        if (!"json".equals(messageStructure)) {
+            return;
+        }
+        JsonNode root;
+        try {
+            root = objectMapper.readTree(message);
+        } catch (Exception e) {
+            throw new AwsException("InvalidParameter",
+                    "Invalid parameter: Message Structure - JSON message body failed to parse", 400);
+        }
+        if (root == null || !root.isObject()) {
+            throw new AwsException("InvalidParameter",
+                    "Invalid parameter: Message Structure - JSON message body should be an object.", 400);
+        }
+        JsonNode defaultValue = root.get("default");
+        if (defaultValue == null || defaultValue.isNull()) {
+            throw new AwsException("InvalidParameter",
+                    "Invalid parameter: Message Structure - No default entry in JSON message body", 400);
+        }
     }
 
     private void recordPushNotification(PushNotification notification) {
--- a/src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
@@ -435,7 +435,7 @@ public class SnsService implements Resettable {
             if (!matchesFilterPolicy(sub, parsedBody, messageAttributes)) {
                 continue;
             }
-            deliverMessage(sub, message, subject, messageAttributes, messageId, effectiveArn, messageGroupId, dedupId);
+            deliverMessage(sub, message, subject, messageAttributes, messageId, effectiveArn, messageGroupId, dedupId, messageStructure);
         }
         LOG.infov("Published message {0} to topic {1}", messageId, effectiveArn);
         return messageId;
@@ -629,6 +629,17 @@ public class SnsService implements Resettable {
             throw new AwsException("PlatformApplicationDisabled",
                     "Platform application is disabled.", 400);
         }
+        return capturePushForEndpoint(endpoint, app, message, subject, messageStructure, messageAttributes);
+    }
+
+    /**
+     * Resolves the platform payload and records a captured push for the given endpoint/app.
+     * Shared by direct-to-endpoint {@code Publish} and topic fan-out to {@code application}
+     * subscriptions, so broadcast pushes surface identically via the retrospection API.
+     */
+    private String capturePushForEndpoint(PlatformEndpoint endpoint, PlatformApplication app,
+                                          String message, String subject, String messageStructure,
+                                          Map<String, MessageAttributeValue> messageAttributes) {
         String payload = resolvePushPayload(app.getPlatform(), message, messageStructure);
         String messageId = UUID.randomUUID().toString();
         recordPushNotification(new PushNotification(
@@ -783,6 +794,7 @@ public class SnsService implements Resettable {
                 continue;
             }
             String subject = (String) entry.get("Subject");
+            String messageStructure = (String) entry.get("MessageStructure");
             String messageGroupId = (String) entry.get("MessageGroupId");
             String messageDeduplicationId = (String) entry.get("MessageDeduplicationId");
 
@@ -812,7 +824,7 @@ public class SnsService implements Resettable {
                     bodyParseAttempted = true;
                 }
                 if (!matchesFilterPolicy(sub, parsedBody, attrs)) continue;
-                deliverMessage(sub, message, subject, attrs, messageId, topicArn, messageGroupId, messageDeduplicationId);
+                deliverMessage(sub, message, subject, attrs, messageId, topicArn, messageGroupId, messageDeduplicationId, messageStructure);
             }
             LOG.debugv("Batch published message {0} (id={1}) to {2}", messageId, id, topicArn);
             successful.add(new String[]{id, messageId});
@@ -1228,7 +1240,8 @@ public class SnsService implements Resettable {
 
     private void deliverMessage(Subscription sub, String message, String subject,
                                 Map<String, MessageAttributeValue> messageAttributes, String messageId,
-                                String topicArn, String messageGroupId, String messageDeduplicationId) {
+                                String topicArn, String messageGroupId, String messageDeduplicationId,
+                                String messageStructure) {
         try {
             switch (sub.getProtocol()) {
                 case "sqs" -> {
@@ -1279,6 +1292,31 @@ public class SnsService implements Resettable {
                     httpClient.sendAsync(request, HttpResponse.BodyHandlers.discarding())
                             .thenAccept(response -> logHttpResult("Delivered SNS notification", endpoint, response.statusCode()))
                             .exceptionally(ex -> { LOG.warnv("Failed to deliver SNS message to {0}: {1}", endpoint, ex.getMessage()); return null; });
+                }
+                case "application" -> {
+                    String region = extractRegionFromArn(sub.getEndpoint());
+                    if (region == null) {
+                        region = extractRegionFromArn(topicArn);
+                    }
+                    String endpointArn = sub.getEndpoint();
+                    PlatformEndpoint endpoint = platformEndpointStore.get(endpointKey(region, endpointArn)).orElse(null);
+                    if (endpoint == null) {
+                        LOG.debugv("Skipping topic fan-out to missing platform endpoint {0}", endpointArn);
+                        break;
+                    }
+                    if (!"true".equalsIgnoreCase(endpoint.getAttributes().getOrDefault("Enabled", "true"))) {
+                        LOG.debugv("Skipping topic fan-out to disabled platform endpoint {0}", endpointArn);
+                        break;
+                    }
+                    PlatformApplication app = platformAppStore.get(
+                            platformAppKey(region, endpoint.getPlatformApplicationArn())).orElse(null);
+                    if (app == null || "false".equalsIgnoreCase(app.getAttributes().get("Enabled"))) {
+                        LOG.debugv("Skipping topic fan-out to endpoint {0}: platform application missing or disabled",
+                                endpointArn);
+                        break;
+                    }
+                    capturePushForEndpoint(endpoint, app, message, subject, messageStructure, messageAttributes);
+                    LOG.debugv("Delivered SNS message to platform endpoint: {0}", endpointArn);
                 }
                 case "email", "email-json" -> LOG.infov("SNS email delivery (stub): to={0}, subject={1}, message={2}",
                         sub.getEndpoint(), subject, message);
--- a/src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
@@ -693,6 +693,8 @@ public class SnsService implements Resettable {
      * top-level {@code default} entry. Validating here (rather than lazily inside per-subscriber
      * delivery) means a malformed envelope surfaces to the caller as {@code InvalidParameter}
      * instead of being silently swallowed while the {@code publish} call still reports success.
+     * {@code PublishBatch} runs the same check per entry, reporting a bad envelope as that entry's
+     * {@code BatchResultErrorEntry} rather than failing the whole batch.
      */
     private void validateTopicMessageStructure(String message, String messageStructure) {
         if (!"json".equals(messageStructure)) {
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
```diff
@@ -1280,6 +1293,31 @@ private void deliverMessage(Subscription sub, String message, String subject,
...
+                        break;
+                    }
+                    PlatformApplication app = platformAppStore.get(
+                            platformAppKey(region, endpoint.getPlatformApplicationArn())).orElse(null);
+                    if (app == null || "false".equalsIgnoreCase(app.getAttributes().get("Enabled"))) {
+                        LOG.debugv("Skipping topic fan-out to endpoint {0}: platform application missing or disabled",
+                                endpointArn);
+                        break;
+                    }
+                    capturePushForEndpoint(endpoint, app, message, subject, messageStructure, messageAttributes);
```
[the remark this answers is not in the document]
AbhigyaKrishna: Fixed in 989a6d16. Topic `Publish` with `MessageStructure=json` now validates synchronously (`validateTopicMessageStructure`) that the message is a JSON object with a top-level `default` entry, before any fan-out — so a malformed envelope surfaces `InvalidParameter` to the caller instead of being swallowed by `deliverMessage`'s catch while `publish` reports success.

The direct-to-endpoint path is intentionally left as-is: it resolves platform-key-or-`default` and already propagates `InvalidParameter` synchronously (never swallowed), matching AWS's per-endpoint behavior where a platform-specific key alone is valid.

Added two regression tests covering the topic fan-out path: a JSON envelope with only `{"GCM":...}` (no `default`) and non-JSON input now both throw and capture zero pushes. Reverting the guard fails both.

AbhigyaKrishna:
need this merged as we are using floci for our uat

src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
```diff
@@ -783,6 +825,7 @@ public BatchPublishResult publishBatch(String topicArn, List<Map<String, Object>
                 continue;
             }
             String subject = (String) entry.get("Subject");
+            String messageStructure = (String) entry.get("MessageStructure");
```
pgermosen, line 833: **`publishBatch` threads `messageStructure` but skips the new validation `publish` does** (consistency — worth fixing here).

`publish` now validates `MessageStructure="json"` synchronously via `validateTopicMessageStructure` (`:407`), so a malformed json envelope surfaces as `InvalidParameter`. `publishBatch` extracts `messageStructure` per entry and threads it into `deliverMessage`, but never runs that validation — and I traced the consequence:

- `deliverMessage` wraps its switch in `try { … } catch (Exception e) { LOG.warnv(…) }` (swallows per-subscriber errors), and
- `capturePushForEndpoint` → `resolvePushPayload` **throws** `InvalidParameter` on a malformed `json` message.

So a `PublishBatch` entry with `MessageStructure=json` and a malformed / `default`-less message **silently succeeds while the push is dropped** for every `application` subscriber (the exception is caught and logged), whereas `publish` returns `InvalidParameter` for the identical input. The two topic-publish paths now disagree.

**I'd fix this in this PR rather than defer it**, for two reasons:

1. The inconsistency is self-inflicted by *this* change — the PR is the one that threads `messageStructure` into `publishBatch`'s `deliverMessage`, so it's already completing the batch path here.
2. The batch loop already has the exact per-entry error scaffolding. A few lines above (missing `Message`) it does `failed.add(new String[]{id, "InvalidParameter", "…", "true"}); continue;`. Validating `MessageStructure=json` right after `messageStructure` is extracted — wrapped so a bad entry becomes that same per-entry `BatchResultErrorEntry` instead of a throw that fails the whole batch — mirrors real `PublishBatch` and slots straight into the existing pattern. It's a small addition, not a new mechanism.

(Separately, `validateTopicMessageStructure` is a correct-but-broader behavior change: *any* topic `Publish` with `MessageStructure=json` and a malformed/`default`-less envelope now returns `InvalidParameter` regardless of subscriber protocol — worth a line in the description so the new rejection isn't a surprise.)

pgermosen (commented):
Good fix for a real gap — `deliverMessage` had no `application` case, so topic fan-out to auto-confirmed platform-endpoint subscriptions delivered nothing and never showed up in the retrospection API. The PR adds that branch, threads `messageStructure` through the delivery boundary so per-endpoint payload keys resolve on fan-out too, and factors the capture tail into a shared `capturePushForEndpoint`.

One consistency finding inline that I'd resolve before approve once the batch path validates too.

src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
```diff
@@ -1228,7 +1280,8 @@ private List<Subscription> subscriptionsByTopic(String topicArn, String region)
 
     private void deliverMessage(Subscription sub, String message, String subject,
                                 Map<String, MessageAttributeValue> messageAttributes, String messageId,
-                                String topicArn, String messageGroupId, String messageDeduplicationId) {
+                                String topicArn, String messageGroupId, String messageDeduplicationId,
+                                String messageStructure) {
```
abanna, line 1287: Nice to see `application` subscriptions getting real fan-out, and validating the envelope synchronously in `publish` rather than lazily per subscriber is the right call.

Threading `messageStructure` into `deliverMessage` makes something adjacent visible, though: `application` is the only case that consumes it. `sqs`, `lambda`, `http`/`https`, `email` and `sms` all still deliver the raw `message`, which under `MessageStructure=json` is the whole envelope. So:

```
aws sns publish --topic-arn ... --message-structure json \
  --message '{"default":"hello","sqs":"hi sqs"}'
```

delivers `hi sqs` to a mobile endpoint's platform key but hands an SQS subscriber the literal `{"default":"hello","sqs":"hi sqs"}`. Real SNS resolves per-protocol for every protocol, falling back to `default`.

The machinery is already here and already generic — `resolvePushPayload(platform, message, messageStructure)` keys off an arbitrary string and returns `message` untouched when the structure isn't `json`. So each case can resolve its own payload:

```java
String body = resolvePushPayload(sub.getProtocol(), message, messageStructure);
```

and `validateTopicMessageStructure` already guarantees `default` exists by delivery time, so the fallback can't come up empty.

Worth doing here or as a follow-up — but a parameter that's present and unused for five of six protocols tends to read as intentional later.

Unrelated: the two red checks don't look like yours. `sdk-test-java` fails on `CodeBuildTest.retryBuild` (expected `IN_PROGRESS`, got `FAILED`) — the same flake I've seen on other PRs this week, which went green on a rebase with no code change. A rebase should clear both.

src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
AbhigyaKrishna: Fixed in 89fce2e1 — agreed it belongs in this PR since this change is what threads `messageStructure` into `publishBatch`'s `deliverMessage`.

`publishBatch` now runs `validateTopicMessageStructure` right after `messageStructure` is extracted per entry, inside the existing per-entry scaffolding: a bad envelope becomes that entry's `BatchResultErrorEntry` (`InvalidParameter`, `SenderFault=true`) and the loop continues, so the rest of the batch still publishes — same shape as the missing-`Message` case a few lines above, and matching real `PublishBatch`. Two regression tests cover it: an entry with `{"GCM":...}` and no `default`, and an entry with non-JSON input; both fail the entry alone while the sibling entry succeeds and its push is captured.

On the broader behavior note: documented. `docs/services/sns.md` now spells out that any topic `Publish` with `MessageStructure=json` requires a JSON object carrying `default`, validated before fan-out, regardless of subscriber protocol.

src/main/java/io/github/hectorvent/floci/services/sns/SnsService.java
AbhigyaKrishna: Fixed in df8dd557 — you're right that the parameter was live for exactly one of six protocols.

`deliverMessage` now resolves per subscription before the switch:

```java
String protocolMessage = "application".equals(sub.getProtocol())
        ? message
        : resolveProtocolPayload(sub.getProtocol(), message, messageStructure);
```

`application` stays on the raw message because it resolves against the push platform (`APNS`/`GCM`/…) inside `capturePushForEndpoint`; every other case now uses `protocolMessage` — `sqs` (both raw delivery and the SNS envelope's `Message` field), `lambda`'s event payload, `http`/`https` (keyed on `sub.getProtocol()`, so `http` and `https` pick their own keys), and the `email`/`email-json`/`sms` stubs. Renamed `resolvePushPayload` → `resolveProtocolPayload` since it is no longer push-specific; the logic is unchanged. As you note, `validateTopicMessageStructure` guarantees `default` by delivery time, so the fallback can't come up empty.

New `SnsProtocolPayloadResolutionTest` covers it: raw SQS gets `hi sqs` from `{"default":"hello","sqs":"hi sqs"}`; the non-raw envelope carries `"Message":"hi sqs"` rather than the structure; SQS falls back to `default` when only a `GCM` key is present; without `MessageStructure=json` the raw body passes through untouched; and one topic with both an SQS and an `application` subscriber resolves `hi sqs` and `hi device` respectively from a single publish. Four of the five fail on the previous commit. All 145 SNS tests pass. Docs updated with a per-protocol section.

Re the red checks — agreed they're not from this change. A rebase onto current main is pending; I'll push it separately.

pgermosen:
Re-checked at current HEAD (`aa51a2d8`) — the `application`-only payload resolution gap abanna flagged is genuinely fixed: `deliverMessage` now calls `resolveProtocolPayload(sub.getProtocol(), message, messageStructure)` for all six delivery branches, not just `application`.

One new gap, not yet raised: AWS's `Publish` API docs are explicit that under `MessageStructure=json`, *"Non-string values will cause the key to be ignored"* — i.e. a non-string value at a subscriber's protocol key should be treated as absent, falling back to `default`. `resolveProtocolPayload` doesn't do that today — when the matched key's value isn't textual, it calls `protocolValue.toString()` instead of ignoring it:

```java
JsonNode protocolValue = root.get(protocol);
if (protocolValue != null && !protocolValue.isNull()) {
    return protocolValue.isTextual() ? protocolValue.asText() : protocolValue.toString();
}
```

Concretely: `{"default":"hi","sqs":42}` delivers the literal string `"42"` to an SQS subscriber, where real SNS would ignore the non-string `sqs` key and deliver `"hi"` instead. Same pattern applies to the `default` fallback a few lines down. `SnsProtocolPayloadResolutionTest`'s 5 cases only cover string values, absent keys, and non-JSON mode — none exercise a non-string value.

Everything else checks out — 145/145 SNS tests pass, `docs-check` clean, no conflicts.

AbhigyaKrishna:
Fixed in a5a4a982 — good catch, the non-string case was being stringified rather than ignored.

`resolveProtocolPayload` now accepts only textual values for both the protocol key and the `default` fallback; a non-string value falls through as if the key were absent, matching AWS's *"Non-string values will cause the key to be ignored"*. So `{"default":"hi","sqs":42}` now delivers `hi` to an SQS subscriber instead of the literal `"42"`.

Also tightened `validateTopicMessageStructure` to require a textual `default` — otherwise a non-string `default` would pass the synchronous topic-publish guard and then throw during fan-out, where the per-subscriber catch swallows it (the exact silent-drop the earlier guard was added to prevent).

`SnsProtocolPayloadResolutionTest` gains two cases: a non-string protocol key falling back to `default`, and rejection of a non-string `default`. All 147 SNS tests pass.

pgermosen (commented):
@AbhigyaKrishna — this looks ready to me on the substance. One small housekeeping item before it goes in, and it is the only thing I would like changed.

## `.codegraph/.gitignore` looks like it rode along by accident

The diff includes a new `.codegraph/.gitignore`:

```
# CodeGraph data files — local to each machine, not for committing.
*
!.gitignore
```

There is nothing wrong with the file itself — a self-ignoring directory stub is a perfectly sensible pattern. But it has nothing to do with SNS mobile push, nothing else in the tree references `.codegraph`, and the root `.gitignore` has no entry for it either. From here it reads like a stray `git add` from local tooling rather than something you meant to ship in a `feat(sns)` change.

Could you drop it from this branch? If you do think Floci should carry it, that is fine too — it would just be better as its own small `chore:` so it gets looked at on its own terms rather than being approved as a side effect of an SNS review. Either way, your call; I would just rather it was a decision than an accident.

## Everything else checks out

I rebuilt the branch against current `main` rather than relying on the mergeability indicator, since the branch is a fair way behind. It merges cleanly, compiles, and the SNS behaviour is green — as are the SQS and Lambda paths, which matter here because `deliverMessage` now resolves the payload for every protocol rather than just `application`.

Both outstanding findings are properly closed, and I confirmed each by reverting it locally rather than reading the diff and agreeing:

- the per-entry validation in `publishBatch` — a bad envelope now fails that entry alone, as its own `BatchResultErrorEntry`, while the rest of the batch still publishes;
- @abanna's point that the threaded `messageStructure` was live for exactly one of six protocols — every protocol now resolves its own key with a `default` fallback, and `application` correctly stays on the raw message because it resolves against the push platform further down.

## Your newest commit is right, and the documentation backs it word for word

`a5a4a982` quotes AWS as *"Non-string values will cause the key to be ignored"*. That is verbatim from the [Publish API reference](https://docs.aws.amazon.com/sns/latest/api/API_Publish.html), which also states:

> Keys in the JSON object that correspond to supported transport protocols must have simple JSON string values.

The same page independently justifies the other half of that commit — tightening the validator to require a *textual* `default`:

> If you set `MessageStructure` to `json`, the value of the `Message` parameter must: be a syntactically valid JSON object; and contain at least a top-level JSON key of "default" with a value that is a string.

And this line vindicates the synchronous-validation shape that both @abanna and I pushed for, so it is worth having on the record:

> Failure to parse or validate any key or value in the message will cause the `Publish` call to return an error (no partial delivery).

That is precisely the argument for validating up front instead of letting a malformed envelope be swallowed per subscriber while the publish reports success. `EndpointDisabled` at HTTP 400 matches the same page too.

## One genuinely minor note, not for this PR

The same constraints list says *"Duplicate keys are not allowed"*, and real SNS errors on them. A JSON parser will quietly keep the last occurrence instead, so a duplicate-key envelope is accepted here where AWS would reject it. It is a narrow edge case, nobody has reported it, and it is outside what this change set out to do — mentioning it only so it is written down somewhere rather than lost.

Drop the stray file and I am happy with this. Thank you for the patience on it, and for chasing the payload-resolution thread further than the original report went.

pgermosen (approved):
Confirmed the stray `.codegraph/.gitignore` is gone in the latest commit — that was the only thing outstanding. Everything else on the SNS side holds up: application fan-out, synchronous envelope validation, per-protocol payload resolution across all six protocols, and the non-string-value handling all check out end to end, including on a fresh merge against current `main`.

Good chase on the payload-resolution consistency thread across all these rounds — thanks for staying with it.

Issue #1859, opened 2026-07-13
AMI dropdown remains empty on Cloud Explorer → Compute (cannot create compute instance)


akhilgentoo:
Description: On Cloud Explorer → Compute, the AMI dropdown (required field) does not load any selectable options. The field stays on the placeholder “Select an AMI…” and no items appear, which prevents creating a new compute instance.

Expected Behavior: The AMI dropdown should populate with a list of available AMIs (or locally available base images) so a user can select one while configuring a new compute instance.

Actual Behavior: No AMI options are shown; the dropdown remains empty/stuck on the placeholder text “Select an AMI…”.

Impact: This blocks users from provisioning new compute instances (required field cannot be selected).

Steps to Reproduce:

Navigate to Cloud Explorer.
Go to the Compute page.
Start creating a new compute instance.

mansiverma897993:
@akhilgentoo @hectorvent I have raised #1860 take a look plz!!

hectorvent:
The cause here has not actually been pinned down yet, so before anyone writes a fix it would help a lot to capture the request the dropdown makes. Running Floci with `FLOCI_LOG_LEVEL=DEBUG` and grabbing the `DescribeImages` call the console sends, in particular whether it passes an `Owners` parameter, would settle it in one step.

The reason it matters: if the console sends no `Owners`, current main already returns the whole catalog and the empty dropdown has a different cause. If it sends `Owners=["self"]`, then an empty result is what real AWS does too, since `self` scopes to images the caller owns and a fresh emulator has none ([botocore ec2/2016-11-15](https://github.com/boto/botocore/blob/develop/botocore/data/ec2/2016-11-15/service-2.json)), which would make this a console side problem rather than a Floci one.

Worth knowing for whoever picks this up: catalog images are still matched by the plain owner comparison in `Ec2ImageCatalog.matchesOwner` (src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ImageCatalog.java:182), while images created through CreateImage got the correct `self` treatment in `Ec2Service.matchesImageOwners` (src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java:4374). Those two paths disagreeing is a real inconsistency, but it is a separate thing from this dropdown until the request above is known.

#1860 was an attempt at this and is being closed as superseded, so the issue is open for anyone who wants it.

PR #1860, opened 2026-07-13
Fix/1859 ami dropdown

mansiverma897993:
## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ImageCatalog.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ImageCatalog.java
@@ -177,11 +177,31 @@ public class Ec2ImageCatalog {
         }
 
         public boolean matchesOwner(List<String> owners) {
+            return matchesOwner(owners, null);
+        }
+
+        public boolean matchesOwner(List<String> owners, String accountId) {
             if (owners == null || owners.isEmpty()) {
                 return true;
             }
             Image image = toImage();
-            return owners.contains(image.getOwnerId()) || owners.contains(image.getImageOwnerAlias());
+            for (String owner : owners) {
+                if ("self".equals(owner)) {
+                    if ("000000000000".equals(accountId)) {
+                        return true;
+                    }
+                    continue;
+                }
+                if (owner != null) {
+                    if (owner.equals(accountId) && "000000000000".equals(accountId)) {
+                        return true;
+                    }
+                    if (owner.equals(image.getOwnerId()) || owner.equalsIgnoreCase(image.getImageOwnerAlias())) {
+                        return true;
+                    }
+                }
+            }
+            return false;
         }
 
         public Image toImage() {
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -1145,7 +1145,7 @@ public class Ec2Service {
         return imageCatalog.images().stream()
                 .filter(Ec2ImageCatalog.CatalogImage::advertised)
                 .filter(img -> img.matchesIdOrAlias(imageIds))
-                .filter(img -> img.matchesOwner(owners))
+                .filter(img -> img.matchesOwner(owners, accountId))
                 .filter(img -> matchesImageFilters(img, filters))
                 .map(Ec2ImageCatalog.CatalogImage::toImage)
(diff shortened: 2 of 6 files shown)


hectorvent:

src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ImageCatalog.java
```diff
@@ -177,11 +177,31 @@ public boolean matchesIdOrAlias(List<String> candidates) {
...
+
+        public boolean matchesOwner(List<String> owners, String accountId) {
             if (owners == null || owners.isEmpty()) {
                 return true;
             }
             Image image = toImage();
-            return owners.contains(image.getOwnerId()) || owners.contains(image.getImageOwnerAlias());
+            for (String owner : owners) {
+                if ("self".equals(owner)) {
+                    if ("000000000000".equals(accountId)) {
```
abanna, line 190: Thanks for chasing the empty-dropdown report. I think this makes `self` mean something it shouldn't, though, and it'll bite once #1979 lands.

Every image in the catalog is owned by a real vendor account — the entries this PR adds are `137112412989`/`amazon`, `136693071363`/`debian`, `538276071271`/`alpine`, alongside the existing `099720109477`/`canonical`. None of them are owned by the caller. But this branch returns `true` for *any* catalog image as soon as the caller is the default account, so:

```
aws ec2 describe-images --owners self
```

returns the entire vendor catalog. In AWS, `--owners self` returns only AMIs the account created, and would be empty on a fresh emulator. The line below has the same effect for an explicit account id:

```java
if (owner.equals(accountId) && "000000000000".equals(accountId)) {
```

That matters more than it looks, because #1979 (`feat(ec2): support CreateImage`) is open — once users can register their own AMIs, `--owners self` needs to distinguish those from the built-ins, and this makes that distinction impossible.

Matching on the image's actual owner keeps both cases working:

```java
if ("self".equals(owner)) {
    if (accountId != null && accountId.equals(image.getOwnerId())) {
        return true;
    }
    continue;
}
```

If the dropdown is empty because the UI sends `owners=["self"]`, the fix probably belongs there — a console listing public AMIs shouldn't be scoping to self in the first place.

Also, `"000000000000"` is hardcoded in both branches. It's `config.defaultAccountId()`, which is configurable — so anyone running with a custom account id gets no images at all, which is the same bug you're fixing, one config change away.

Two process notes. The branch carries two commits from #1736 (`2afb5982`, `10cb4eca`), so its S3 `encoding-type` changes are in this diff — worth rebasing onto current `main` once #1736 merges so this PR is just the AMI fix. And only Greptile has run here; `Build and Test`, the native jobs and `Commits follow Conventional Commits` never triggered, so this is effectively untested. That last one would fail as-is — neither the PR title (`Fix/1859 ami dropdown`) nor commit `463ea052` has a conventional prefix.

hectorvent:
Closing this one, and thank you for it. The reason is a happy one: most of it already shipped.

The S3 half of this diff is on main, through your own #1736, which merged on 13 August carrying this branch's commits 2afb5982 and 10cb4eca. The encoding-type echo, the `maybeEncode` calls in ListObjects and ListObjectVersions, and the literal `+` preservation are all there, with tests under the same names you wrote (`listObjectsV2PreservesPlusCharacter`, `listObjectsV2EncodingTypeIsEchoed`). The `ownerId` and `imageOwnerAlias` entries in image-catalog.yaml are on main too. That overlap is exactly why this branch went CONFLICTING, so the conflict is a sign your work landed rather than a sign anything went wrong.

That leaves one file genuinely unmerged, `Ec2ImageCatalog.matchesOwner`, and it is the one carrying the blocking finding above: `self` means owned by the caller, so returning the whole vendor catalog for `--owners self` would be a divergence rather than a fix. Main has since grown the correct form for images created through CreateImage, in `Ec2Service.matchesImageOwners` (src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java:4374), and the catalog path at Ec2ImageCatalog.java:182 has not caught up. Whether it should is a real question, just not one this branch can settle.

#1859 stays open. We still do not know which DescribeImages call the console actually makes, and I have asked on the issue for that request to be captured first, since the fix looks quite different depending on whether it passes `Owners` at all.

If you want it after that, a fresh branch off current main will be far less work than rebasing this one. Sorry it took so long to get you a clear answer.

PR #1862, opened 2026-07-13
fix(dynamodb): tokenize UpdateExpression whitespace-insensitively

### LINKED ISSUE #1640: [BUG] DynamoDB: multi-line UpdateExpression (newline after SET) rejected with 'Syntax error; token: "SET"'
## Summary

DynamoDB `UpdateItem` fails with a `ValidationException` when the `UpdateExpression` contains a **newline immediately after the `SET` keyword** (i.e. a multi-line expression):

```
Invalid UpdateExpression: Syntax error; token: "SET", near: "SET #a"
```

Real AWS DynamoDB tokenizes `UpdateExpression` **whitespace-insensitively**, so newlines/indentation between clauses and operands are valid. FLOCI rejects the form. The single-line equivalent (`SET #a = :a, #b = :b`) is accepted, so it's purely a whitespace/newline tokenization issue.

Multi-line update expressions are common in real codebases (template literals), so existing code that works against AWS breaks under FLOCI.

Verified on `floci/floci:1.5.26`.

## Reproduction

Requirements: `docker`, `aws` CLI.

```bash
#!/usr/bin/env bash
set -euo pipefail
export AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy AWS_DEFAULT_REGION=us-east-1
EP=http://localhost:4566

docker run -d --rm --name floci-ue -p 4566:4566 floci/floci:1.5.26 >/dev/null
for _ in $(seq 1 30); do curl -sf "$EP/" >/dev/null 2>&1 && break; sleep 1; done

aws --endpoint-url "$EP" dynamodb create-table --table-name Repro \
  --key-schema AttributeName=pk,KeyType=HASH --billing-mode PAY_PER_REQUEST >/dev/null

names='{"#a":"A","#b":"B"}'
values='{":a":{"S":"x"},":b":{"BOOL":true}}'

# (1) multi-line: newline immediately after SET  -> FLOCI rejects, AWS accepts
multiline=$'\n  SET\n    #a = :a,\n    #b = :b\n'
echo "multi-line:"
aws --endpoint-url "$EP" dynamodb update-item --table-name Repro --key '{"pk":{"S":"1"}}' \
  --update-expression "$multiline" \
  --expression-attribute-names "$names" --expression-attribute-values "$values" \
  || echo "  ^ REJECTED"

# (2) single-line equivalent -> accepted
echo "single-line:"
aws --endpoint-url "$EP" dynamodb update-item --table-name Repro --key '{"pk":{"S":"2"}}' \
  --update-expression 'SET #a = :a, #b = :b' \
  --expression-attribute-names "$names" --expression-attribute-values "$values" \
  && echo "  ^ ACCEPTED"

docker stop floci-ue >/dev/null
```

## Expected

Both forms succeed (real AWS DynamoDB accepts the multi-line expression — whitespace, including newlines, is not significant between tokens).

## Actual

- **(1) multi-line** → `ValidationException: Invalid UpdateExpression: Syntax error; token: "SET", near: "SET #a"`
- **(2) single-line** → succeeds

The only difference is the newline/indentation after `SET`.

## Notes

The same applies to `KeyConditionExpression` / `ConditionExpression` / `ProjectionExpression` if they share the tokenizer — worth checking. A whitespace-insensitive tokenizer (treat any run of whitespace incl. `\n`, `\t` as a separator) would resolve it.

## Environment

- Image: `floci/floci:1.5.26`
- Client: AWS CLI v2 against `http://localhost:4566`

ravi-arnan:
## Problem

`UpdateItem` rejects a multi-line `UpdateExpression` (a newline after a clause keyword) with:

```
Invalid UpdateExpression: Syntax error; token: "SET", near: "SET #a"
```

Real AWS tokenizes `UpdateExpression` whitespace-insensitively, so newlines, tabs, and multiple spaces between keywords and operands are valid. Multi-line expressions are common in real code (template literals), so code that works against AWS breaks under Floci. Fixes #1640.

## Root cause

Clause dispatch keyed on the literal `startsWith("SET ")` (keyword plus a single space) in two places, so anything other than a single space after a keyword failed all four keyword checks and fell through to the syntax-error branch:

- `DynamoDbJsonHandler` pre-validation (the request-level syntax check)
- `DynamoDbService.applyUpdateExpression` (the actual parser)

Thanks to @soreavis for pinpointing the parser dispatch in the issue thread.

## Fix

Collapse runs of whitespace to a single space at both parse entry points before matching keywords, which matches AWS's whitespace-insensitive tokenization. Attribute names and values in an expression carry no internal whitespace (values are `:placeholder` references, names are `#alias` or bare identifiers), so normalizing whitespace is safe.

## Verification

Live against `quarkus:dev` on 1.5.32:

- `SET\n #a = :a, #b = :b` (newline after `SET`) now succeeds
- tabs, multiple spaces, and a newline before `REMOVE` all succeed
- `ADD` with a newline succeeds
- single-line expressions unchanged
- an invalid leading keyword (`FOO ...`) is still rejected with the same `ValidationException`

Added `DynamoDbUpdateExpressionWhitespaceTest` covering the newline, mixed-whitespace, and negative cases. Full DynamoDB suite passes (349 tests, including the 3 new ones).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbJsonHandler.java
@@ -510,8 +510,10 @@ public class DynamoDbJsonHandler {
 
         // Validate EAN/EAV usage across UpdateExpression + ConditionExpression
         if (updateExpression != null) {
-            // Pre-validate syntax: expression must start with a valid clause keyword
-            String ue = updateExpression.trim();
+            // Pre-validate syntax: expression must start with a valid clause keyword.
+            // AWS tokenizes UpdateExpression whitespace-insensitively, so collapse runs
+            // of whitespace (newlines, tabs, multiple spaces) before matching keywords.
+            String ue = updateExpression.trim().replaceAll("\\s+", " ");
             if (!ue.isEmpty()) {
                 String upper = ue.toUpperCase();
                 if (!upper.startsWith("SET ") && !upper.startsWith("REMOVE ") && !upper.startsWith("ADD ") && !upper.startsWith("DELETE ")) {
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -1440,7 +1440,10 @@ public class DynamoDbService {
                     "Invalid UpdateExpression: The expression can not be empty;", 400);
         }
         DynamoDbReservedWords.check(expression, "UpdateExpression");
-        String remaining = expression.trim();
+        // AWS tokenizes UpdateExpression whitespace-insensitively, so collapse runs of
+        // whitespace (newlines, tabs, multiple spaces) so clause-keyword dispatch and the
+        // comma-separated action parsing below work regardless of formatting.
+        String remaining = expression.trim().replaceAll("\\s+", " ");
 
         while (!remaining.isEmpty()) {
             String upper = remaining.toUpperCase();
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -1440,10 +1440,7 @@ public class DynamoDbService {
                     "Invalid UpdateExpression: The expression can not be empty;", 400);
         }
         DynamoDbReservedWords.check(expression, "UpdateExpression");
-        // AWS tokenizes UpdateExpression whitespace-insensitively, so collapse runs of
-        // whitespace (newlines, tabs, multiple spaces) so clause-keyword dispatch and the
-        // comma-separated action parsing below work regardless of formatting.
-        String remaining = expression.trim().replaceAll("\\s+", " ");
+        String remaining = expression.trim();
 
         while (!remaining.isEmpty()) {
             String upper = remaining.toUpperCase();
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -1439,8 +1439,14 @@ public class DynamoDbService {
             throw new AwsException("ValidationException",
                     "Invalid UpdateExpression: The expression can not be empty;", 400);
         }
-        DynamoDbReservedWords.check(expression, "UpdateExpression");
-        String remaining = expression.trim();
+        // AWS tokenizes UpdateExpression whitespace-insensitively, so collapse runs of
+        // whitespace (newlines, tabs, multiple spaces) so clause-keyword dispatch and the
+        // comma-separated action parsing below work regardless of formatting. Normalize
+        // before the reserved-word check too: its function-call lookahead skips only
+        // literal spaces, so on a raw "if_not_exists\n(...)" it would read the function
+        // name as a bare identifier instead.
+        String remaining = expression.trim().replaceAll("\\s+", " ");
+        DynamoDbReservedWords.check(remaining, "UpdateExpression");
 
         while (!remaining.isEmpty()) {
             String upper = remaining.toUpperCase();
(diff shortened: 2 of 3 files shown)


ravi-arnan:
Addressed the Greptile note in 04bf548. `createAndSeed` now accepts 200 or 400 (ResourceInUseException) for `CreateTable`, so the whitespace tests are safe to re-run against a persisted store; PutItem still overwrites, keeping the seed deterministic. Verified locally on JDK 25: `DynamoDbUpdateExpressionWhitespaceTest` passes (3/3). Looks like the workflow runs are awaiting maintainer approval on this fork PR whenever you get a chance. Thanks!

hampsterx (commented):
Reviewed the parser change and cross-checked the whitespace-collapse against the UpdateExpression grammar and the downstream scanner with two independent models, the core fix is correct and the safety reasoning holds up.

**Correctness (confirmed)**
- Collapsing runs of whitespace can't corrupt a valid UpdateExpression: every token (`#alias`, `:placeholder`, bare identifiers, `.`/`[n]` path punctuation, `=`/`+`/`-`, commas, `if_not_exists`/`list_append`) is whitespace-free, and actual attribute values live in `ExpressionAttributeValues`, never in the expression text.
- No new leniency: `replaceAll("\\s+", " ")` collapses runs but never deletes whitespace, so `S E T` stays three tokens and still fails `startsWith("SET ")`.
- It fixes more than the leading keyword: both the leading dispatch (`startsWith("SET ")`) and the clause-boundary scan (`findNextClauseKeyword` → `indexOfKeyword(upper, "SET ")`) match keywords with a literal trailing space, so any non-space whitespace after a clause keyword broke them before. `findNextComma` / `findArithmeticOperator` are character-based and unaffected.

**One ask before merge, extend the tests.** The suite covers SET (newline/tab/mixed) and an invalid leading keyword, but the paths most likely to hide a follow-on regression aren't exercised:
- `ADD` and `DELETE` with a newline (separate clause parsers, split on `\s+`)
- a multiline function/arithmetic RHS, e.g. `list_append(#a,\n:v)` or `if_not_exists(#a, :v)\n+ :w`
- a chain across >2 clauses, e.g. `SET a = :v\nREMOVE b\nADD c :w`

Cheap to add next to the existing cases, and they lock in exactly the formatting this PR enables.

**Minor (non-blocking):** `DynamoDbReservedWords.check(...)` runs on the raw `expression` before normalization, and its function-call lookahead in `extractBareIdentifiers` skips only literal spaces (`charAt(j) == ' '`), so `if_not_exists\n(...)` is treated there as a bare identifier. Harmless today (those names aren't reserved), but running `check` on the normalized string would keep the two entry points consistent.

Targeted, well-scoped fix.

---
*AI-assisted review (Opus, cross-validated with GPT-5.6 + GLM-5.2, diff + surrounding parser code). Please verify.*

ravi-arnan:
Thanks for the thorough review, that is a genuinely useful cross-check of the collapse against the grammar.

Tests extended in 7af0f44f, all three axes you named:

- **`ADD` / `DELETE` with a newline** after the keyword (`newlineAfterAddAndDeleteKeywordsIsAccepted`). Seeds `n` (N) and `s` (SS), asserts the ADD arithmetic and the set-element DELETE both land.
- **Multiline function / arithmetic RHS** (`multilineFunctionAndArithmeticRightHandSidesAreAccepted`): `list_append(#l,\n:more)` and `if_not_exists(#c, :zero)\n+ :inc` in one expression, so the depth-aware comma scan and `findArithmeticOperator` are both exercised on a wrapped RHS.
- **Three-clause chain** (`newlineSeparatedChainOfThreeClausesIsAccepted`): `SET` / `REMOVE` / `ADD`, so the clause-boundary scan runs repeatedly rather than once.

**One correction worth flagging on your example.** `SET a = :v\nREMOVE b\nADD c :w` as written passes on `main` today, unfixed. The newlines there sit *before* each keyword, and `indexOfKeyword`'s boundary check is already `Character.isWhitespace(upper.charAt(idx - 1))`, so a leading newline was never the broken case. What breaks is a newline *after* the keyword, because the keyword itself is matched with a literal trailing space (`"SET "`, `"ADD "`). Same for the multiline-RHS case: the value-part scanners are character-based, so that formatting also parsed fine before this PR.

I moved the newlines to sit after the keyword in all three tests so they actually lock the fix in rather than passing either way. Verified in both directions: with the two `main` sources reverted to pre-fix, the three new tests fail with `Syntax error; token: "SET"` (x2) and `Syntax error; token: "ADD"` (x1); with the fix restored they pass. Full DynamoDB suite green on JDK 25, **357 tests, 0 failures**.

**Minor also addressed** in 56969b3a. `DynamoDbReservedWords.check` now runs on the normalized string rather than the raw one, so both entry points see the same text. Agreed it is inert today (neither `if_not_exists` nor `list_append` is reserved), so this is purely the consistency cleanup you described, no behaviour change intended.

One ask back: the workflow runs on this fork PR still need maintainer approval to start, currently 0 runs on `56969b3a`. Whenever you get a chance. Thanks again!

hectorvent (approved):
Thanks @ravi-arnan, 

Solid fix.

PR #1863, opened 2026-07-13
fix(apigateway): preserve trailing slash in proxy event path

### LINKED ISSUE #1557: [BUG] API Gateway REST: trailing slash stripped from proxy event path (event.path /thing/ -> /thing)
## Service

API Gateway (REST v1) / Lambda

## AWS API Action

`CreateRestApi`, `CreateResource`, `PutMethod`, `PutIntegration`, `CreateDeployment`, `CreateStage`, then `_user_request_` invocation (AWS_PROXY integration).

## Expected behavior

A trailing slash in the request path is significant and must be preserved in the Lambda proxy event. On AWS API Gateway (and AWS SAM local), a request to `/thing/` produces an event with the trailing slash intact:

```
GET /thing   -> event.path = "/thing",  event.requestContext.path = ".../thing"
GET /thing/  -> event.path = "/thing/", event.requestContext.path = ".../thing/"
```

This matters because many HTTP routers treat `/thing` and `/thing/` as distinct routes (Go's `go-json-rest`, several others). A backend that registers a handler on `POST /thing/` relies on the gateway delivering the path verbatim.

## Actual behavior

Floci `1.5.27` **strips the trailing slash** when building the proxy event. Both requests yield the same path:

```
GET /thing   -> event.path = "/thing"
GET /thing/  -> event.path = "/thing"   <-- trailing slash dropped
```

The downstream Lambda therefore never sees `/thing/`. A handler registered only on `/thing/` becomes unreachable and the router returns `405 Method Not Allowed` (or `404`), even though the exact same deployment works on AWS and on AWS SAM local. There is no way to recover the original path on the backend.

## Minimal reproduction (AWS SDK / CLI against floci)

```bash
EP="--endpoint-url http://localhost:4566"

# 1) An echo Lambda that returns the path fields from the event
cat > index.js <<'JS'
exports.handler = async (event) => ({
  statusCode: 200,
  body: JSON.stringify({
    path: event.path,
    requestContextPath: event.requestContext && event.requestContext.path,
    resource: event.resource,
  }),
});
JS
zip -q fn.zip index.js
aws $EP lambda create-function --function-name echo \
  --runtime nodejs20.x --handler index.handler \
  --role arn:aws:iam::000000000000:role/r --zip-file fileb://fn.zip

# 2) REST API: /{proxy+} ANY -> echo
RID=$(aws $EP apigateway create-rest-api --name slash-demo --query id --output text)
ROOT=$(aws $EP apigateway get-resources --rest-api-id "$RID" --query "items[0].id" --output text)
PROXY=$(aws $EP apigateway create-resource --rest-api-id "$RID" --parent-id "$ROOT" --path-part '{proxy+}' --query id --output text)
aws $EP apigateway put-method      --rest-api-id "$RID" --resource-id "$PROXY" --http-method ANY --authorization-type NONE
aws $EP apigateway put-integration --rest-api-id "$RID" --resource-id "$PROXY" --http-method ANY \
  --type AWS_PROXY --integration-http-method POST \
  --uri "arn:aws:apigateway:eu-west-1:lambda:path/2015-03-31/functions/arn:aws:lambda:eu-west-1:000000000000:function:echo/invocations"
DID=$(aws $EP apigateway create-deployment --rest-api-id "$RID" --stage-name bootstrap --query id --output text)
aws $EP apigateway create-stage --rest-api-id "$RID" --stage-name prod --deployment-id "$DID" >/dev/null

BASE="http://localhost:4566/restapis/$RID/prod/_user_request_"
# warm up the lambda container first (cold start returns an empty body)
for i in 1 2 3 4 5; do [ -n "$(curl -s "$BASE/warmup")" ] && break; done
echo "no slash:";       curl -s "$BASE/thing"
echo; echo "trailing slash:"; curl -s "$BASE/thing/"
```

Observed on floci `1.5.27`:

```
no slash:        {"path":"/thing","requestContextPath":"/thing","resource":"/{proxy+}"}
trailing slash:  {"path":"/thing","requestContextPath":"/thing","resource":"/{proxy+}"}   <-- expected "/thing/"
```

Both calls return the identical `"/thing"`; the trailing slash on the second is gone. On AWS API Gateway / SAM local the second call returns `"path":"/thing/"`.

## Impact

Any backend whose router distinguishes a trailing slash and registers a route *with* one (e.g. `POST /thing/`) is unreachable under floci, while identical code works on AWS and SAM local. The failure surfaces as a confusing `405`/`404` that only happens locally.

## Likely source

Path normalization in floci's `_user_request_` request handling / proxy-event construction (the component that derives `event.path` and `requestContext.path` from the incoming request). The trailing slash should be passed through verbatim rather than canonicalized.

## Environment

- floci `1.5.27` (JVM/native image), `FLOCI_DEFAULT_REGION=eu-west-1`
- REST API with `/{proxy+}` `ANY` AWS_PROXY integration, invoked via `/_user_request_`.
- Reference implementations that preserve the slash: AWS API Gateway, AWS SAM local.

ravi-arnan:
## Problem

A REST API `AWS_PROXY` integration dropped a trailing slash from the request path in the Lambda proxy event. A request to `/thing/` produced `event.path = "/thing"` instead of `"/thing/"`. A trailing slash is significant to many backends (Go's `go-json-rest` and others treat `/thing` and `/thing/` as distinct routes), so a handler registered on `/thing/` never receives the request. Fixes #1557.

## Root cause

`ApiGatewayExecuteController` builds the request path from the JAX-RS `{proxy: .*}` path param, and that binding strips the trailing slash. Confirmed on `main`:

```
request /thing/  ->  proxy=[thing]   uriInfo.getRawPath()=[.../thing/]
request /a/b/    ->  proxy=[a/b]      uriInfo.getRawPath()=[.../a/b/]
```

So `path = "/" + proxy` lost the slash, and that value flowed into `event.path`, `requestContext.path`, and the `proxy` path parameter.

## Fix

Recover the trailing slash from the raw request URI (`uriInfo.getRequestUri().getRawPath()`), which retains it, and apply it to `path` and `requestContext.path` on both the Lambda proxy event and the REQUEST authorizer event. Resource matching and path-parameter extraction keep using the normalized path, mirroring AWS: `/thing/` still routes to the `/thing` resource, but the delivered event preserves the slash. The recovery is factored into a small pure `preserveTrailingSlash(normalizedPath, rawRequestPath)` helper.

`pathParameters.proxy` and `methodArn` deliberately stay normalized. Both carried the slash in an earlier revision of this PR; @hectorvent caught the proxy case in review and 38fcca2 dropped both. The greedy `{proxy+}` value comes from the matcher, which ran on the stripped path, and independent reports of the AWS behaviour confirm the asymmetry ([vendia/serverless-express#391](https://github.com/vendia/serverless-express/issues/391) traces a broken route to exactly it). `methodArn` gets matched against IAM-style policy resources, where a stray trailing slash would silently miss a wildcard an authorizer returns, so the conservative form is the safer default absent an authoritative capture.

## Verification

Live against `quarkus:dev`, `AWS_PROXY` integration, observing the built event path:

```
/thing   -> event.path /thing
/thing/  -> event.path /thing/     (was /thing)
/a/b/    -> event.path /a/b/        (was /a/b)
/a/b     -> event.path /a/b
```

Added `preserveTrailingSlash` unit tests to `ApiGatewayProxyMatchTest` (preservation, no-op without a trailing slash, root path untouched, no double slash, null-safe), plus authorizer-event coverage. Green locally on JDK 25: `ApiGatewayProxyMatchTest` (19), `ApiGatewayRequestAuthorizerIntegrationTest` (26), `ApiGatewayExecuteControllerTest` (9).

Diff is +85/-9 across 3 files.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -517,7 +517,10 @@ public class ApiGatewayExecuteController {
 
         ObjectNode node = objectMapper.createObjectNode();
         node.put("type", auth.getType());
-        node.put("methodArn", buildMethodArn(region, apiId, stageName, httpMethod, preservedPath));
+        // methodArn keeps the normalized path: it is matched against IAM-style policy resources,
+        // where a stray trailing slash would silently fail wildcards an authorizer already returns.
+        // No AWS behavior was found pinning it either way, so the conservative form wins.
+        node.put("methodArn", buildMethodArn(region, apiId, stageName, httpMethod, requestPath));
         if ("TOKEN".equals(auth.getType())) {
             String headerName = auth.getIdentitySource().replace("method.request.header.", "");
             node.put("authorizationToken", headers.getHeaderString(headerName));
@@ -625,7 +628,6 @@ public class ApiGatewayExecuteController {
         // Recover it from the raw request URI for the event path fields. Resource matching
         // and path-parameter extraction continue to use the normalized `path`.
         String requestPath = preserveTrailingSlash(path, uriInfo.getRequestUri().getRawPath());
-        boolean trailingSlash = !requestPath.equals(path);
 
         ObjectNode event = objectMapper.createObjectNode();
         event.put("resource", resourcePath);
@@ -637,9 +639,11 @@ public class ApiGatewayExecuteController {
         putQueryStringParameters(event, uriInfo);
         putMultiValueQueryStringParameters(event, uriInfo);
 
+        // pathParameters come from the matcher, which ran on the normalized path, so the greedy
+        // {proxy+} value has no trailing slash on real AWS even when event.path keeps one.
         ObjectNode pathParams = event.putObject("pathParameters");
         if (proxy != null && !proxy.isEmpty()) {
-            pathParams.put("proxy", trailingSlash ? proxy + "/" : proxy);
+            pathParams.put("proxy", proxy);
         }
         extractPathParams(resourcePath, path).forEach(pathParams::put);
 
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -615,9 +615,16 @@ public class ApiGatewayExecuteController {
                                    byte[] body, String requestId,
                                    String principalId, Map<String, Object> authorizerContext,
                                    String resolvedApiKey) {
+        // The JAX-RS {proxy} binding strips a trailing slash, but a trailing slash is
+        // significant in the delivered path (routers treat /x and /x/ as distinct routes).
+        // Recover it from the raw request URI for the event path fields. Resource matching
+        // and path-parameter extraction continue to use the normalized `path`.
+        String requestPath = preserveTrailingSlash(path, uriInfo.getRequestUri().getRawPath());
+        boolean trailingSlash = !requestPath.equals(path);
+
         ObjectNode event = objectMapper.createObjectNode();
         event.put("resource", resourcePath);
-        event.put("path", path);
+        event.put("path", requestPath);
         event.put("httpMethod", httpMethod);
 
         putSingleValueHeaders(event, headers);
@@ -626,7 +633,9 @@ public class ApiGatewayExecuteController {
         putMultiValueQueryStringParameters(event, uriInfo);
 
         ObjectNode pathParams = event.putObject("pathParameters");
-        if (proxy != null && !proxy.isEmpty()) pathParams.put("proxy", proxy);
+        if (proxy != null && !proxy.isEmpty()) {
+            pathParams.put("proxy", trailingSlash ? proxy + "/" : proxy);
+        }
         extractPathParams(resourcePath, path).forEach(pathParams::put);
 
         // stageVariables: populate from the Stage object (null if no variables configured)
@@ -652,7 +661,7 @@ public class ApiGatewayExecuteController {
         ctx.put("domainPrefix", apiId);
         ctx.put("extendedRequestId", requestId);
         ctx.put("httpMethod", httpMethod);
-        ctx.put("path", path);
+        ctx.put("path", requestPath);
         ctx.put("protocol", "HTTP/1.1");
         ctx.put("requestId", requestId);
         ctx.put("requestTime", requestTime);
@@ -1908,6 +1917,23 @@ public class ApiGatewayExecuteController {
     }
 
     // ──────────────────────────── Path matching ────────────────────────────
+
+    /**
+     * Re-appends a trailing slash that the JAX-RS {@code {proxy}} path-param binding strips.
+     * A trailing slash is significant in the proxy event path (many routers treat {@code /x}
+     * and {@code /x/} as distinct routes), so it is recovered from the raw request URI. The
+     * normalized path is still used for resource matching and path-parameter extraction, which
+     * mirrors AWS routing {@code /x/} to the {@code /x} resource while keeping the slash in the
+     * delivered event. Returns {@code normalizedPath} unchanged for the root path or when the
+     * raw request had no trailing slash.
+     */
+    static String preserveTrailingSlash(String normalizedPath, String rawRequestPath) {
+        if (!"/".equals(normalizedPath) && !normalizedPath.endsWith("/")
+                && rawRequestPath != null && rawRequestPath.endsWith("/")) {
+            return normalizedPath + "/";
+        }
+        return normalizedPath;
+    }
 
     /**
      * Finds all matching resources for {@code requestPath}, sorted by specificity.
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -510,15 +510,20 @@ public class ApiGatewayExecuteController {
                                      String httpMethod, String requestPath,
                                      String resourcePath, String resourceId, Stage stage, UriInfo uriInfo,
                                      String resolvedApiKey) {
+        // Recover the trailing slash the JAX-RS {proxy} binding strips, so the authorizer sees
+        // the same raw path the Lambda later receives from buildProxyEvent (AWS parity). Path
+        // matching and path-parameter extraction keep using the normalized requestPath.
+        String preservedPath = preserveTrailingSlash(requestPath, uriInfo.getRequestUri().getRawPath());
+
         ObjectNode node = objectMapper.createObjectNode();
         node.put("type", auth.getType());
-        node.put("methodArn", buildMethodArn(region, apiId, stageName, httpMethod, requestPath));
+        node.put("methodArn", buildMethodArn(region, apiId, stageName, httpMethod, preservedPath));
         if ("TOKEN".equals(auth.getType())) {
             String headerName = auth.getIdentitySource().replace("method.request.header.", "");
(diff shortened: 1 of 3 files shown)


ravi-arnan:
Pushed 61592a0 addressing the Greptile trailing-slash consistency finding. The CUSTOM authorizer was seeing the normalized path while the Lambda received the slash-preserved one, so path-based access control could gate a different value than the handler ran on. `toAuthorizerEvent` now uses `preserveTrailingSlash` for `path`, `methodArn`, and `requestContext.path`, while path-parameter extraction stays on the normalized path (matching `buildProxyEvent`). Added an end-to-end regression in `ApiGatewayRequestAuthorizerIntegrationTest` (now 26/26 green locally on JDK 25) asserting the authorizer event keeps the trailing slash while `pathParameters` do not. Workflow runs look like they are awaiting maintainer approval on this fork PR. Thanks for taking a look!

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
```diff
@@ -1889,6 +1903,23 @@ private MultivaluedMap<String, String> parseFormUrlEncoded(String body) {
...
+    /**
+     * Re-appends a trailing slash that the JAX-RS {@code {proxy}} path-param binding strips.
+     * A trailing slash is significant in the proxy event path (many routers treat {@code /x}
+     * and {@code /x/} as distinct routes), so it is recovered from the raw request URI. The
+     * normalized path is still used for resource matching and path-parameter extraction, which
+     * mirrors AWS routing {@code /x/} to the {@code /x} resource while keeping the slash in the
+     * delivered event. Returns {@code normalizedPath} unchanged for the root path or when the
+     * raw request had no trailing slash.
+     */
+    static String preserveTrailingSlash(String normalizedPath, String rawRequestPath) {
```
abanna, line 1939: Good catch on #1557, and threading it through `toAuthorizerEvent` as well as `buildProxyEvent` is the right call — that keeps the authorizer and the Lambda seeing the same string, which is the part that would otherwise be a genuine security footgun.

(Worth noting for anyone reading the bot review above: it reports `toAuthorizerEvent` as *not* fixed, but this branch does apply `preserveTrailingSlash` there at lines 499–540. That one looks stale.)

The gap that does survive is HTTP API v2, which this PR doesn't touch. `dispatchV2` builds its path exactly the same way JAX-RS mangles the v1 one:

```java
String path = "/" + (proxy == null ? "" : proxy);
```

and that value becomes the event's `rawPath` at lines 1708 and 1820 — a field whose whole contract is that it's the *raw* path. A request to `/thing/` on an HTTP API delivers `rawPath: "/thing"`, so a v2 Lambda can't distinguish `/thing` from `/thing/` at all. `buildRequestAuthorizerEventV1` (line 1660, used for v2 authorizers on payload format 1.0) has the same problem at its `methodArn`, `resource`, `path` and `requestContext.path`.

So after this PR a REST API preserves the trailing slash and an HTTP API silently doesn't — a confusing difference to hit, especially since routers commonly treat the two as distinct routes.

`preserveTrailingSlash` is already static and package-private, so v2 can use it as-is:

```java
String path = preserveTrailingSlash(
        "/" + (proxy == null ? "" : proxy),
        uriInfo.getRequestUri().getRawPath());
```

One caveat if you do: `findMatchingRoute` at line 1201 would then receive the slash-preserving path, so route matching should keep using the normalized value — the same split you already made explicit in the v1 comments.

Fine as a follow-up rather than in this PR; mostly worth deciding deliberately so v1 and v2 don't drift.

ravi-arnan:
Thanks for the review @abanna, and for calling out the stale bot finding.

Agreed on keeping v2 out of this PR. I filed #2136 for it, with your `rawPath` analysis and the `findMatchingRoute` caveat about route matching staying on the normalized path. Happy to pick it up once this lands.

Also rebased onto `main` to clear a conflict that appeared from #1806 and #1630 (both touched `ApiGatewayExecuteController` and `ApiGatewayProxyMatchTest`). Resolution was purely additive: `preserveTrailingSlash` now sits alongside the renamed `matchResources`, and both of #1630's new precedence tests are kept next to the trailing-slash ones. The diff is unchanged at +93/-6 across 3 files. `ApiGatewayProxyMatchTest` (19) and `ApiGatewayRequestAuthorizerIntegrationTest` (26) are green locally on JDK 25.

hectorvent:
Thank you for the rebase and for extending the fix to the authorizer path, and sorry this is the first detailed reply. The core claims hold up against LocalStack's snapshot-tested execute-api: routing happens on the stripped path while the Lambda event keeps the un-stripped one, so `event.path` keeping the slash is correct.

**(blocking)** One sub-claim looks wrong: LocalStack takes path parameters from the matcher, which ran on the stripped path, so the greedy `proxy` value is `thing`, not `thing/`. Could you verify against real AWS, or drop the `trailingSlash ? proxy + "/" : proxy` line?

**(blocking)** The new push also puts the slash into `methodArn`, and the tests now assert it. I could not find any authority pinning `methodArn` either way, so a real AWS capture there would settle both this and the proxy question in one probe.

One probe resolves both points, then this is good to go.

ravi-arnan:
Thanks, you were right on the first point, and I dropped both contested lines rather than guessing.

**`pathParameters.proxy`** (38fcca2) now stays normalized. I could not capture from a real AWS account, but independent reports of the AWS behaviour line up with your reading: `event.path` preserves a trailing slash while `pathParameters.proxy` does not. The clearest is [vendia/serverless-express#391](https://github.com/vendia/serverless-express/issues/391), where the reporter traces a broken route to exactly that asymmetry: "the `req.path` is obtained from `event.pathparameters.proxy` instead of `event.path`, and the last slash is not included in `event.pathparameters.proxy`". Frameworks that route off the proxy param hit this and have to fall back to `event.path`, which only makes sense if the matcher-derived value is stripped.

**`methodArn`** also reverts to the normalized path, and I removed the assertion that pinned the slash there. I found no authority either way, same as you, and the conservative form is the safer default: `methodArn` gets matched against IAM-style policy resources, so a stray trailing slash would silently fail a wildcard an authorizer returns. Turning a path quirk into an authorization miss is a bad way to be wrong.

What remains is only what the evidence supports: `event.path` and `requestContext.path` keep the slash, on both the Lambda proxy event and the authorizer event, while routing and path-parameter extraction stay on the normalized path.

Diff is now +85/-9 across 3 files. Tests green locally on JDK 25: `ApiGatewayProxyMatchTest` (19), `ApiGatewayRequestAuthorizerIntegrationTest` (26), `ApiGatewayExecuteControllerTest` (9).

One thing your report surfaced that is worth flagging beyond this PR: a trailing-slash mismatch between the routing layer and the authorizer layer is [a real HTTP API vulnerability class](https://www.infoq.com/news/2026/06/aws-api-gateway-auth-bypass/), where `/v1/accounts` returned 401 and `/v1/accounts/` returned 200 because the authorizer context was dropped on the rewritten path. That is v2 and out of scope here, but it raises the stakes on #2136.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
ravi-arnan: Replying in-thread so this one is not left looking open: filed as #2136, with your `rawPath` analysis and the `findMatchingRoute` caveat that route matching stays on the normalized path. Keeping v2 out of this PR as agreed, and happy to pick #2136 up once this lands.

ravi-arnan:
Small housekeeping note: the `waiting-contributor` label is still on this PR. It was applied on 08-06 at 04:52 UTC, about six hours before 38fcca2 and my 10:38 reply, which addressed both blocking points (the PR body no longer claims the slash reaches `pathParameters.proxy`, and I explained why proxy and `methodArn` stay normalized).

So as it stands this is waiting on re-review rather than on me. Happy to make further changes if anything is still open, just let me know.

hectorvent (approved):

PR #1868, opened 2026-07-14
fix(s3): fail fast on unwritable data root instead of opaque 500s

### LINKED ISSUE #1664: [BUG] All S3 operations return 500 on amd64 (start OK, arm64 works)
## Service

S3

## AWS API Action

All S3 operations (CreateBucket, ListBuckets, ...)

## Expected behavior

S3 calls should succeed (or return an AWS-shaped error). On arm64 they do.

## Actual behavior

On amd64, every S3 call returns `HTTP 500 Internal Server Error` with
`content-type: text/plain` and a body like:

```
500 - Internal Server Error
	Error id <uuid>
```

The container starts fine and `/_localstack/health` reports all services
`running` (including `s3`), but S3 is dead. The same image on arm64 works.

This breaks the Serverless Framework v4 deploy: its deployment-bucket step calls
CreateBucket, gets the 500/text-plain body, and its S3 REST-XML parser dies with
`char '{' is not expected.:1:1`.

## Reproduction

```
docker run -d -p 4566:4566 floci/floci:1.5.28-compat
aws --endpoint-url=http://localhost:4566 s3api create-bucket \
  --bucket test --create-bucket-configuration LocationConstraint=eu-west-2
# -> An error occurred (500) when calling the CreateBucket operation: Internal Server Error
```

ListBuckets, raw curl, and us-east-1 (no LocationConstraint) all 500 the same
way. Seen on GitHub-hosted `ubuntu-24.04` runners. arm64 returns 200 for the
same calls. Tried both `1.5.28` and `1.5.28-compat`.

Possibly related to the x86-64 microarchitecture handling in #287.

## Environment

- Floci version / image tag: 1.5.28 and 1.5.28-compat (amd64)
- How you're running Floci: Docker, GitHub-hosted ubuntu-24.04 runner
- Env: FLOCI_STORAGE_MODE=hybrid, FLOCI_DEFAULT_REGION=eu-west-2

hectorvent:
## Summary

All S3 operations returned an opaque 500 when the persistent data root was not writable (typically read-only or root-owned bind mounts — the amd64 symptom in the report). This adds a boot-time writability check that fails fast with an actionable message whenever a non-memory storage mode is enabled, S3/S3-Control exception mappers that keep even unexpected failures on the S3 wire contract, and an entrypoint warning when the state dir is not writable by the container user. 

Closes #1664

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: an unwritable data root made `S3Service`'s constructor
fail during lazy CDI instantiation, so every S3 request surfaced as
Quarkus's plain-text 500 — unparseable by REST-XML SDK clients. Unhandled
failures on S3 routes now render S3's `<Error><Code>InternalError</Code>…`
XML contract, and non-memory storage refuses to boot on an unwritable path
instead of failing per-request.

## Checklist

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -41,6 +41,26 @@ if [ "${LOCALSTACK_PARITY:-true}" != "false" ]; then
     . /usr/local/bin/localstack-parity.sh
 fi
 
+# Probe the state dir as the unprivileged user and warn loudly when it is not
+# writable (read-only or root-owned bind mounts the chown above could not fix).
+# The container still starts: with the default in-memory storage the dir is
+# unused, and non-memory modes fail fast at boot with an actionable error.
+data_dir="${FLOCI_STORAGE_PERSISTENT_PATH:-/app/data}"
+if [ -d "$data_dir" ]; then
+    probe="$data_dir/.floci-write-probe.$$"
+    if ! ( touch "$probe" && rm -f "$probe" ) 2>/dev/null; then
+        cat >&2 <<EOF
+**************************************************************************
+WARNING: $data_dir is not writable by $(id -un) (uid $(id -u)).
+Persisted state cannot be saved there. If FLOCI_STORAGE_MODE (or any
+per-service storage mode) is not 'memory', Floci will refuse to start.
+Fix the volume mount permissions (it may be read-only or root-owned) or
+set FLOCI_STORAGE_PERSISTENT_PATH to a writable directory.
+**************************************************************************
+EOF
+    fi
+fi
+
 # Fall back to the image's default command when invoked with no arguments.
 # Some tooling re-executes this entrypoint directly with an empty argv —
 # Testcontainers' LocalStackContainer does exactly that via its starter
--- a/docker/test-entrypoint.sh
+++ b/docker/test-entrypoint.sh
@@ -84,5 +84,35 @@ else
     printf '[SKIP] empty argv prefers the native binary when present (/app not writable)\n'
 fi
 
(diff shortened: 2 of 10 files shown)


abanna:
Nice one, @hectorvent — really welcome fix. The "every S3 call returns an opaque 500" symptom from a root-owned or read-only data mount has bitten me more than once in container setups, and it's genuinely hard to diagnose from the outside. Failing fast at boot with an actionable message, and keeping the S3 / S3-Control error responses on the XML wire contract even for unexpected failures, is exactly the right call. 👍

A couple of small, non-blocking notes:

- The boot validator probes the root and the `s3/` subdir. Since the process creates the per-service subdirs itself, a writable root is a good proxy — I just wondered whether a pre-existing, differently-owned subdir (e.g. `s3/` left by an earlier run under a different uid) could still slip past the root probe. Likely an edge case not worth chasing.
- Running it locally I noticed the entrypoint probe defaults to `/app/data`, while the app resolves `FLOCI_STORAGE_PERSISTENT_PATH` / `./data` relative to the workdir. In setups that relocate the workdir (I run with a custom workdir + relative `./data`), the bash warning can end up probing a different directory than the app actually uses — the Java `PersistentPathValidator` still catches it authoritatively, so it's only the early warning that could point at the wrong path.

Thanks for the thorough tests — the read-only entrypoint case is a nice touch. LGTM, and happy to give it a spin against a root-owned mount if that'd help.

PR #1869, opened 2026-07-14
fix(lifecycle): tear down process-bound containers on shutdown

hectorvent:
## Summary

Docker containers bound to the emulator process (Lambda warm pool, ECS tasks, EC2 instances, in-flight Batch/CodeBuild jobs) were torn down ad hoc via `@PreDestroy`, which fires after the ShutdownEvent observers — after the storage flush schedulers have stopped. Containers could leak past shutdown and state changes made while stopping missed the final flush. This introduces a `ContainerTeardown` contract implemented by each container runner, invoked from `EmulatorLifecycle.onStop` before `StorageFactory.shutdownAll()`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

No wire-protocol changes — this fixes emulator lifecycle behavior only.
Teardown now runs at a deterministic point, so persisted service state
after a stop matches what a client observed before it.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/ContainerTeardown.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/ContainerTeardown.java
@@ -1,0 +1,19 @@
+package io.github.hectorvent.floci.core.common;
+
+/**
+ * Implemented by services that launch Docker containers whose lifetime is bound to the
+ * emulator process (Lambda warm pool, ECS tasks, EC2 instances, in-flight build/job
+ * containers). {@code EmulatorLifecycle.onStop} invokes every implementation during the
+ * ShutdownEvent phase — before {@code StorageFactory.shutdownAll()} — so teardown runs at
+ * a deterministic point and state changes made while stopping are still captured by the
+ * final flush. A {@code @PreDestroy} alone is not sufficient for that: bean destruction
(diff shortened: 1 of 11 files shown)


abanna:
This is great, @hectorvent — thank you. The `@PreDestroy`-fires-after-the-flush-schedulers-stop ordering is a subtle one; I've seen both state written during shutdown get lost and the occasional orphaned container after a restart. Centralizing teardown behind a `ContainerTeardown` contract that `onStop` drains *before* `storageFactory.shutdownAll()` is a clean, deterministic fix, and discovering implementers via `Instance<ContainerTeardown>` keeps it nicely extensible.

A few small notes (all non-blocking):

- The per-implementation try/catch in the `onStop` loop is the right instinct — one runner failing to tear down won't skip the others or block the final flush. 👍
- The in-flight tracking in the Batch/CodeBuild runners (record under `run()`, remove in `finally`, drain idempotently on shutdown) reads well and lines up with the "may also be invoked from `@PreDestroy` as a fallback" note in the contract.
- Minor: is there any ordering dependency between implementers (say ECS tasks vs. the EC2 instances backing them)? From the diff they look independent so iteration order shouldn't matter — just flagging in case a future runner introduces one.

Really appreciate this — the cleaner shutdown will help a lot for the restart-heavy local flows. LGTM.

PR #1879, opened 2026-07-15
fix(neptune): throw a mappable capacity fault when the proxy-port range is exhausted

ahmedk20:
allocateProxyPort threw AwsException("InsufficientNeptuneCapacity"), a code that does not exist in Neptune's API, so the SDK could not map it and callers got a generic NeptuneException. Use the wire code InsufficientStorageClusterCapacity (the InsufficientStorageClusterCapacityFault that CreateDBCluster declares), 400, so the SDK deserializes it to a typed exception.

## Summary
When the Neptune proxy-port range is exhausted, allocateProxyPort threw AwsException("InsufficientNeptuneCapacity", …, 503). That code does not exist in Neptune's API, so the AWS SDK can't map it to a typed exception — callers get a generic NeptuneException instead of a capacity fault.
This changes it to the wire code the SDK maps to InsufficientStorageClusterCapacityFault — the only capacity fault Neptune's CreateDBCluster declares — with HTTP 400.
Follow-up to the Neptune provisioning-rollback PR #1742 , where this was flagged in review as out-of-scope. The rollback made port exhaustion rarer but not impossible, so the wrong code could still surface.
## Type of change
- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore
## AWS Compatibility

Incorrect behavior: on proxy-port exhaustion the emulator returned error code InsufficientNeptuneCapacity (HTTP 503), which is not a Neptune error type. The AWS SDK v2 could not map it to a modeled exception, so CreateDBCluster callers saw a generic NeptuneException rather than a capacity fault.

After: returns InsufficientStorageClusterCapacity (HTTP 400), the wire code for InsufficientStorageClusterCapacityFault, which CreateDBCluster declares — so the SDK deserializes it to the modeled exception.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/neptune/NeptuneService.java
@@ -308,8 +308,11 @@ public class NeptuneService {
                 return port;
             }
         }
-        throw new AwsException("InsufficientNeptuneCapacity",
-                "No available proxy ports in range " + base + "-" + max, 503);
+        // Wire code the SDK maps to InsufficientStorageClusterCapacityFault (the only
+        // capacity fault CreateDBCluster declares); "InsufficientNeptuneCapacity" isn't a
+        // real Neptune code, so callers got an unmapped generic NeptuneException.
+        throw new AwsException("InsufficientStorageClusterCapacity",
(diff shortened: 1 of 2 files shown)


ahmedk20:
@hectorvent

hectorvent (approved):
Thanks for the PR and your patient

PR #1880, opened 2026-07-15
feat(kinesis): add local inspection endpoints for UI

themark007:
## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/docs/services/kinesis.md
+++ b/docs/services/kinesis.md
@@ -37,6 +37,17 @@
 | `UpdateStreamMode` | - |
 <!-- floci:actions:end -->
 
+## Local Inspection Endpoints
+
+These endpoints are Floci-local read-only helpers for the UI and tests. AWS SDK
+traffic should continue to use the JSON 1.1 Kinesis API on `/`.
+
+| Endpoint | Description |
+|---|---|
+| `GET /_aws/kinesis/streams` | List streams in the resolved region with shard, mode, tag, and record counts |
+| `GET /_aws/kinesis/records?StreamName=<name>` | Peek stream records without consuming them |
+| `GET /_aws/kinesis/records?StreamName=<name>&ShardId=<id>` | Peek records for one shard |
+
 ## Stream Addressing
 
 Most actions accept either `StreamName` or `StreamARN` to identify a stream. When both are provided, `StreamName` takes precedence. `CreateStream` only accepts `StreamName`.
--- a/docs/services/kinesis.md
+++ b/docs/services/kinesis.md
@@ -45,7 +45,7 @@ traffic should continue to use the JSON 1.1 Kinesis API on `/`.
 | Endpoint | Description |
 |---|---|
 | `GET /_aws/kinesis/streams` | List streams in the resolved region with shard, mode, tag, and record counts |
-| `GET /_aws/kinesis/records?StreamName=<name>` | Peek up to 100 stream records without consuming them |
+| `GET /_aws/kinesis/records?StreamName=<name>` | Peek up to 100 stream records with shard attribution without consuming them |
 | `GET /_aws/kinesis/records?StreamName=<name>&Limit=<n>` | Peek up to `n` records, capped at 1000 |
 | `GET /_aws/kinesis/records?StreamName=<name>&ShardId=<id>` | Peek records for one shard |
 
--- a/docs/services/kinesis.md
+++ b/docs/services/kinesis.md
@@ -45,7 +45,8 @@ traffic should continue to use the JSON 1.1 Kinesis API on `/`.
 | Endpoint | Description |
 |---|---|
 | `GET /_aws/kinesis/streams` | List streams in the resolved region with shard, mode, tag, and record counts |
-| `GET /_aws/kinesis/records?StreamName=<name>` | Peek stream records without consuming them |
+| `GET /_aws/kinesis/records?StreamName=<name>` | Peek up to 100 stream records without consuming them |
+| `GET /_aws/kinesis/records?StreamName=<name>&Limit=<n>` | Peek up to `n` records, capped at 1000 |
 | `GET /_aws/kinesis/records?StreamName=<name>&ShardId=<id>` | Peek records for one shard |
 
 ## Stream Addressing
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
@@ -1,0 +1,156 @@
+package io.github.hectorvent.floci.services.kinesis;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.github.hectorvent.floci.core.common.RegionResolver;
+import io.github.hectorvent.floci.services.kinesis.model.KinesisRecord;
+import io.github.hectorvent.floci.services.kinesis.model.KinesisShard;
+import io.github.hectorvent.floci.services.kinesis.model.KinesisStream;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.QueryParam;
+import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.HttpHeaders;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+
+import java.util.Base64;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Local inspection endpoint for the Floci UI and test helpers.
+ *
+ * <p>The AWS-compatible Kinesis API remains the JSON 1.1 endpoint on {@code /};
+ * these {@code _aws} paths are read-only convenience views over local emulator
+ * state.
+ */
+@Path("/_aws/kinesis")
+@Produces(MediaType.APPLICATION_JSON)
+public class KinesisInspectionController {
+
+    private final KinesisService kinesisService;
+    private final ObjectMapper objectMapper;
+    private final RegionResolver regionResolver;
+
+    @Inject
+    public KinesisInspectionController(KinesisService kinesisService,
+                                       ObjectMapper objectMapper,
+                                       RegionResolver regionResolver) {
+        this.kinesisService = kinesisService;
+        this.objectMapper = objectMapper;
+        this.regionResolver = regionResolver;
+    }
+
+    @GET
+    @Path("/streams")
+    public Response getStreams(@Context HttpHeaders headers) {
+        String region = regionResolver.resolveRegion(headers);
+        ArrayNode streams = objectMapper.createArrayNode();
+        for (KinesisStream stream : kinesisService.listStreamDetails(region)) {
+            streams.add(streamNode(stream));
+        }
+        ObjectNode result = objectMapper.createObjectNode();
+        result.set("streams", streams);
+        return Response.ok(result).build();
+    }
+
+    @GET
+    @Path("/records")
+    public Response getRecords(@Context HttpHeaders headers,
+                               @QueryParam("StreamName") String streamName,
+                               @QueryParam("ShardId") String shardId) {
+        if (streamName == null || streamName.isBlank()) {
+            return Response.status(400)
+                    .entity(objectMapper.createObjectNode().put("message", "StreamName query parameter is required"))
+                    .build();
+        }
+
+        String region = regionResolver.resolveRegion(headers);
+        List<KinesisRecord> captured = kinesisService.peekRecords(streamName, shardId, region);
+        ArrayNode records = objectMapper.createArrayNode();
+        for (KinesisRecord record : captured) {
+            records.add(recordNode(record));
+        }
+        ObjectNode result = objectMapper.createObjectNode();
+        result.set("records", records);
+        return Response.ok(result).build();
+    }
+
+    private ObjectNode streamNode(KinesisStream stream) {
+        ObjectNode node = objectMapper.createObjectNode();
+        node.put("StreamName", stream.getStreamName());
+        node.put("StreamARN", stream.getStreamArn());
+        node.put("StreamStatus", stream.getStreamStatus());
+        node.put("StreamMode", stream.getStreamMode());
+        node.put("RetentionPeriodHours", stream.getRetentionPeriodHours());
+        node.put("EncryptionType", stream.getEncryptionType());
+        if (stream.getKeyId() != null) {
+            node.put("KeyId", stream.getKeyId());
+        } else {
+            node.putNull("KeyId");
+        }
+        if (stream.getStreamCreationTimestamp() != null) {
+            node.put("StreamCreationTimestamp", stream.getStreamCreationTimestamp().toString());
+        }
+        node.put("OpenShardCount", stream.getShards().stream().filter(s -> !s.isClosed()).count());
+        node.put("RecordCount", stream.getShards().stream().mapToInt(s -> s.getRecords().size()).sum());
+        ArrayNode enhancedMonitoring = node.putArray("EnhancedMonitoring");
+        Set<String> metrics = stream.getEnhancedMonitoringMetrics();
+        if (metrics != null) {
+            metrics.stream().sorted().forEach(enhancedMonitoring::add);
+        }
+
+        ArrayNode shards = node.putArray("Shards");
+        for (KinesisShard shard : stream.getShards()) {
+            shards.add(shardNode(shard));
+        }
+
+        ObjectNode tags = node.putObject("Tags");
+        Map<String, String> streamTags = stream.getTags();
+        if (streamTags != null) {
+            for (Map.Entry<String, String> entry : streamTags.entrySet()) {
+                tags.put(entry.getKey(), entry.getValue());
+            }
+        }
+        return node;
+    }
+
+    private ObjectNode shardNode(KinesisShard shard) {
+        ObjectNode node = objectMapper.createObjectNode();
+        node.put("ShardId", shard.getShardId());
+        node.put("Closed", shard.isClosed());
+        node.put("RecordCount", shard.getRecords().size());
+        if (shard.getParentShardId() != null) {
+            node.put("ParentShardId", shard.getParentShardId());
+        }
+        if (shard.getAdjacentParentShardId() != null) {
+            node.put("AdjacentParentShardId", shard.getAdjacentParentShardId());
+        }
+        ObjectNode hashKeyRange = node.putObject("HashKeyRange");
+        hashKeyRange.put("StartingHashKey", shard.getHashKeyRange().startingHashKey());
+        hashKeyRange.put("EndingHashKey", shard.getHashKeyRange().endingHashKey());
+        ObjectNode sequenceNumberRange = node.putObject("SequenceNumberRange");
+        sequenceNumberRange.put("StartingSequenceNumber", shard.getSequenceNumberRange().startingSequenceNumber());
+        if (shard.getSequenceNumberRange().endingSequenceNumber() != null) {
+            sequenceNumberRange.put("EndingSequenceNumber", shard.getSequenceNumberRange().endingSequenceNumber());
+        }
+        return node;
+    }
+
+    private ObjectNode recordNode(KinesisRecord record) {
+        ObjectNode node = objectMapper.createObjectNode();
+        node.put("SequenceNumber", record.getSequenceNumber());
+        node.put("PartitionKey", record.getPartitionKey());
+        byte[] data = record.getData();
+        node.put("Data", data == null ? "" : Base64.getEncoder().encodeToString(data));
+        if (record.getApproximateArrivalTimestamp() != null) {
+            node.put("ApproximateArrivalTimestamp", record.getApproximateArrivalTimestamp().toString());
+        }
+        return node;
+    }
+}
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
@@ -26,8 +26,8 @@ import java.util.Set;
  * Local inspection endpoint for the Floci UI and test helpers.
  *
  * <p>The AWS-compatible Kinesis API remains the JSON 1.1 endpoint on {@code /};
- * these {@code _aws} paths are read-only convenience views over local emulator
- * state.
+ * these {@code /_aws} paths are read-only convenience views over local emulator
+ * state and do not follow AWS error-response conventions.
  */
 @Path("/_aws/kinesis")
 @Produces(MediaType.APPLICATION_JSON)
@@ -68,6 +68,11 @@ public class KinesisInspectionController {
         if (streamName == null || streamName.isBlank()) {
             return Response.status(400)
                     .entity(objectMapper.createObjectNode().put("message", "StreamName query parameter is required"))
+                    .build();
+        }
+        if (limit != null && limit <= 0) {
+            return Response.status(400)
+                    .entity(objectMapper.createObjectNode().put("message", "Limit must be greater than 0"))
                     .build();
         }
 
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
@@ -63,7 +63,8 @@ public class KinesisInspectionController {
     @Path("/records")
     public Response getRecords(@Context HttpHeaders headers,
                                @QueryParam("StreamName") String streamName,
-                               @QueryParam("ShardId") String shardId) {
+                               @QueryParam("ShardId") String shardId,
+                               @QueryParam("Limit") Integer limit) {
         if (streamName == null || streamName.isBlank()) {
             return Response.status(400)
                     .entity(objectMapper.createObjectNode().put("message", "StreamName query parameter is required"))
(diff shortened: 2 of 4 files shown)


themark007:
Can someone appprove this pr
@hectorvent 

it’s my first pr so needs an approval

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
```diff
@@ -0,0 +1,156 @@
...
+        ObjectNode node = objectMapper.createObjectNode();
+        node.put("SequenceNumber", record.getSequenceNumber());
+        node.put("PartitionKey", record.getPartitionKey());
+        byte[] data = record.getData();
+        node.put("Data", data == null ? "" : Base64.getEncoder().encodeToString(data));
+        if (record.getApproximateArrivalTimestamp() != null) {
+            node.put("ApproximateArrivalTimestamp", record.getApproximateArrivalTimestamp().toString());
+        }
+        return node;
+    }
```
pgermosen: **Records carry no `ShardId`, so multi-shard peeks lose attribution.**

`recordNode` emits `SequenceNumber`, `PartitionKey`, `Data` and `ApproximateArrivalTimestamp` — but not which shard the record came from. When `/records?StreamName=x` is called without `ShardId`, `peekRecords` flat-maps every shard's records into one flat list, so a UI rendering a multi-shard stream cannot tell which shard any record belongs to, and cannot build a shard drill-down without re-querying once per shard (using shard IDs it would first have to fetch from `/streams`).

Since `/streams` already exposes per-shard `RecordCount`, the natural companion is for `/records` to say where each record sits. Cheapest fix is to carry the shard through the flat-map in `peekRecords` — e.g. return `Map.Entry<String, KinesisRecord>` or a small record type — and add one line here:

```suggestion
        node.put("ShardId", shardId);
```

This is the one item I'd want before merge: it's the gap against the PR's own stated purpose. Everything else in this review is discussion or follow-up.

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
```diff
@@ -0,0 +1,156 @@
...
+    @GET
+    @Path("/records")
+    public Response getRecords(@Context HttpHeaders headers,
+                               @QueryParam("StreamName") String streamName,
+                               @QueryParam("ShardId") String shardId) {
+        if (streamName == null || streamName.isBlank()) {
+            return Response.status(400)
+                    .entity(objectMapper.createObjectNode().put("message", "StreamName query parameter is required"))
+                    .build();
+        }
```
pgermosen, line 72: **Two different error shapes depending on how the request fails.** *(question, not blocking)*

A missing `StreamName` returns `{"message": "StreamName query parameter is required"}` with 400. An *unknown* stream takes a different path: `peekRecords` -> `resolveStream` throws `AwsException("ResourceNotFoundException", ..., 400)`, which `AwsExceptionMapper` renders as `{"__type": "ResourceNotFoundException", "message": "Stream x not found"}` — also 400, not 404. So a client has to handle both an AWS protocol error envelope and a plain `message` object, and never sees a 404 for a missing resource.

greptile raised this as P2. I'd soften it: **`SqsInspectionController` behaves identically** — `SqsService.peekMessages` calls `ensureQueueExists`, which throws the same way — so this PR is consistent with the existing pattern rather than introducing an inconsistency.

@hectorvent — worth settling repo-wide rather than per PR: should the `/_aws/*` surface have a uniform non-AWS error shape (and 404 for not-found), or is leaking the AWS envelope intentional so LocalStack-style helpers parse one format everywhere? Either answer is fine; applying it to all five controllers at once is the useful part.

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisInspectionController.java
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Local inspection endpoint for the Floci UI and test helpers.
+ *
+ * <p>The AWS-compatible Kinesis API remains the JSON 1.1 endpoint on {@code /};
+ * these {@code _aws} paths are read-only convenience views over local emulator
+ * state.
+ */
+@Path("/_aws/kinesis")
+@Produces(MediaType.APPLICATION_JSON)
+public class KinesisInspectionController {
```
pgermosen, line 34: **Is this shape meant to be LocalStack-compatible, or Floci-native?** *(non-blocking)*

The existing controllers' javadoc says *"LocalStack-compatible REST endpoint for ..."* and their response shapes deliberately mirror LocalStack's so test helpers drop in unchanged — `SesInspectionController` even carries a comment explaining it returns `RawData` for raw emails specifically because LocalStack does. This one says *"Local inspection endpoint for the Floci UI and test helpers"* instead.

As far as I know LocalStack has no `/_aws/kinesis` equivalent, so there's nothing to mirror and inventing a shape is the only option — which is fine. But since Floci positions itself as a drop-in alternative, worth being deliberate: if this one is Floci-native, saying so in the javadoc (as you have) is right, and a line in the docs section would stop someone later assuming LocalStack parity.

This is the part of greptile's P2 I'd keep — not "don't add the endpoint", but "be explicit that this one is ours."

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -112,6 +112,13 @@ public List<String> listStreams(String region) {
                 .toList();
     }
 
+    public List<KinesisStream> listStreamDetails(String region) {
+        String prefix = region + "::";
+        return store.scan(key -> key.startsWith(prefix)).stream()
+                .sorted(Comparator.comparing(KinesisStream::getStreamName))
+                .toList();
+    }
```
pgermosen, line 122: **Nit / reuse — `listStreams` and `listStreamDetails` are now the same scan with a different projection.**

AGENTS.md asks to reuse existing patterns; here the existing method can just delegate, which also guarantees the two cannot drift on ordering or key format:

```suggestion
    public List<KinesisStream> listStreamDetails(String region) {
        String prefix = region + "::";
        return store.scan(key -> key.startsWith(prefix)).stream()
                .sorted(Comparator.comparing(KinesisStream::getStreamName))
                .toList();
    }
```

...and then above it:

```java
public List<String> listStreams(String region) {
    return listStreamDetails(region).stream()
            .map(KinesisStream::getStreamName)
            .toList();
}
```

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -477,6 +484,16 @@ public Map<String, Object> getRecords(String shardIterator, Integer limit, Strin
...
 
+    public List<KinesisRecord> peekRecords(String streamName, String shardId, String region) {
+        KinesisStream stream = resolveStream(streamName, region);
+        return stream.getShards().stream()
+                .filter(shard -> shardId == null || shardId.isBlank() || shard.getShardId().equals(shardId))
+                .flatMap(shard -> shard.getRecords().stream())
+                .sorted(Comparator.comparing(KinesisRecord::getApproximateArrivalTimestamp,
+                        Comparator.nullsLast(Comparator.naturalOrder())))
+                .toList();
+    }
```
pgermosen: **Suggestion — `/records` has no limit, so one response can carry a whole retention window.**

`peekRecords` returns every record in the stream (or shard), and `recordNode` base64-encodes each payload into a single JSON document. Kinesis retention defaults to 24 hours and a record can be up to 1 MB, so a stream that has taken real traffic can produce a response large enough to be a problem for both the emulator and the UI rendering it.

The AWS `GetRecords` path already takes a `limit`. A `@QueryParam("Limit")` here — default ~100, as LocalStack-style helpers tend to — would make the endpoint safe to point at any stream. Non-blocking, but much easier to add now than after a UI depends on getting everything.

---

Two smaller things on the same method:

**Unknown `ShardId` is indistinguishable from an empty shard.** The `.filter(...)` on the next line means a `ShardId` matching no shard filters everything out and returns `200 {"records": []}` — same response as a shard that exists but is empty. greptile flagged this too; I agree at low priority, since a UI only ever passes IDs it got from `/streams`. If you want to tighten it, validating the shard before filtering and throwing `ResourceNotFoundException` would match how `resolveStream` treats an unknown stream one line above.

**New unsynchronized reader over the live record lists** *(pre-existing, not blocking)*. `shard.getRecords()` hands back the live `ArrayList` field, and `PutRecord` mutates it via `shard.getRecords().add(record)` at `:376` with no locking anywhere in this class — so streaming over it here can hit `ConcurrentModificationException` and surface as a 500. `getRecords` at `:426` and `:555` already read the same lists the same way, so this PR adds a reader rather than introducing the hazard. Flagging only because the exposure profile changes: a UI polling `/records` on an interval while a producer writes is exactly the sustained concurrent read/write the AWS paths only hit incidentally. If it starts showing up, the fix belongs in the model (copy-on-read in `getRecords()`, or a synchronized list), not in this controller.

pgermosen (approved):
Clean, well-scoped change that clearly studied the existing controllers before writing a new one. 

One thing I'd fix before merge (comment 1): records carry no `ShardId`, which leaves the endpoint incomplete for the UI use case in the title. The rest is discussion or follow-up.

Minor process note: the PR description is still the unfilled template — no Summary, no AWS Compatibility section, all three checklist boxes unchecked. Tests do pass, so it's just paperwork.

themark007:
Hi @hectorvent @pgermosen @okinaka @electrum @abanna @jvanzyl  , I've addressed the review comments (added ShardId and record limiting) and pushed the fixes. When you have a chance, could you please  view and approve the PR? Since it's from a fork, I believe the workflows also need maintainer approval. Thanks!

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -477,6 +486,39 @@ public Map<String, Object> getRecords(String shardIterator, Integer limit, Strin
...
+            }
+        }
+
+        return stream.getShards().stream()
+                .filter(shard -> shardId == null || shardId.isBlank() || shard.getShardId().equals(shardId))
+                .flatMap(shard -> shard.getRecords().stream()
+                        .map(record -> new PeekedRecord(shard.getShardId(), record)))
+                .sorted(Comparator.comparing(peeked -> peeked.record().getApproximateArrivalTimestamp(),
+                        Comparator.nullsLast(Comparator.naturalOrder())))
+                .limit(resolvedLimit)
```
pgermosen: **`Limit` returns the *oldest* N records, so the tail is unreachable.** *(requested change)*

Ascending sort followed by `limit` takes the **oldest** `resolvedLimit` records — confirmed by running it: `Limit=2` on `[oldest, middle, newest]` returns `oldest, middle`.

Combined with the default of 100 and **no pagination**: once a stream has accumulated more than `Limit` records, every subsequent call returns the same stale head, and newly produced records are permanently invisible. A UI polling `/records` on a live stream shows frozen data.

Inspection views almost always want the most recent records — that's what `LATEST` means on a real shard iterator. Taking the newest N and returning them in chronological order:

```suggestion
                .sorted(Comparator.comparing((PeekedRecord peeked) -> peeked.record().getApproximateArrivalTimestamp(),
                        Comparator.nullsLast(Comparator.naturalOrder())).reversed())
                .limit(resolvedLimit)
                .sorted(Comparator.comparing((PeekedRecord peeked) -> peeked.record().getApproximateArrivalTimestamp(),
                        Comparator.nullsLast(Comparator.naturalOrder())))
```

(Take the newest N, then re-sort ascending so the response still reads oldest-first.) If you'd rather keep oldest-first truncation, the endpoint needs a cursor / `NextToken` so a caller can walk to the tail — but newest-N is far simpler and matches what an inspection view wants.

**The new test doesn't pin this down.** `recordsEndpointHonorsLimit` puts three records and asserts `hasSize(2)`, which passes for either ordering. Worth asserting *which* two:

```java
.body("records[0].PartitionKey", equalTo("pk-2"))
.body("records[1].PartitionKey", equalTo("pk-3"));
```

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -477,6 +486,39 @@ public Map<String, Object> getRecords(String shardIterator, Integer limit, Strin
...
+                .toList();
+    }
+
+    private int resolveInspectionRecordLimit(Integer limit) {
+        if (limit == null) {
+            return DEFAULT_INSPECTION_RECORD_LIMIT;
+        }
+        if (limit <= 0) {
+            throw new AwsException("InvalidArgumentException", "Limit must be greater than 0", 400);
+        }
```
pgermosen: **There are now three error shapes on this one endpoint.** *(minor / consistency)*

Verified by running it — `Limit=0` returns `{"__type":"InvalidArgumentException","message":"Limit must be greater than 0"}`. So `/records` now answers with:

| Failure | Shape | Status |
|---|---|---|
| missing `StreamName` | `{"message": "..."}` | 400 |
| `Limit <= 0` | `{"__type": "InvalidArgumentException", ...}` | 400 |
| unknown stream / shard | `{"__type": "ResourceNotFoundException", ...}` | 400 |

My earlier comment asked whether the `/_aws/*` surface should have a uniform error shape; this adds a third instance rather than settling it. Still not blocking — but `StreamName` and `Limit` are both query-parameter validation on the same endpoint, and they now disagree purely because one check lives in the controller and the other in the service. At minimum those two should match. Moving the `Limit` check next to the `StreamName` check in the controller would do it, and keeps the service free of presentation concerns.

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -477,6 +486,39 @@ public Map<String, Object> getRecords(String shardIterator, Integer limit, Strin
...
+        return Math.min(limit, MAX_INSPECTION_RECORD_LIMIT);
```
pgermosen: **The upper bound clamps silently while the lower bound throws.** *(minor)*

`Math.min(limit, MAX_INSPECTION_RECORD_LIMIT)` means `Limit=5000` quietly returns 1000 with nothing saying the request was modified, while `Limit=0` throws. Two ends of the same parameter, two policies.

Worth noting the neighbours in this file: `getRecords` at `:467` and `:595` both do `limit != null ? Math.min(limit, 1000) : 1000` — silent clamping at both ends, no throw. So the 1000 value is consistent with house precedent (good), but the throwing behavior is new and makes this method inconsistent with its own siblings. Either clamp both ends like the neighbours, or validate both. I'd lean toward matching the neighbours, since this isn't an AWS-protocol endpoint and silent capping is the established local behavior.

Also worth stating in the docs: with no pagination, 1000 is a hard ceiling on how many records are reachable at all, not just per request.

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -477,6 +486,39 @@ public Map<String, Object> getRecords(String shardIterator, Integer limit, Strin
...
+        }
+
+        return stream.getShards().stream()
+                .filter(shard -> shardId == null || shardId.isBlank() || shard.getShardId().equals(shardId))
+                .flatMap(shard -> shard.getRecords().stream()
+                        .map(record -> new PeekedRecord(shard.getShardId(), record)))
+                .sorted(Comparator.comparing(peeked -> peeked.record().getApproximateArrivalTimestamp(),
+                        Comparator.nullsLast(Comparator.naturalOrder())))
+                .limit(resolvedLimit)
+                .toList();
```
pgermosen, line 529: **`sorted()` still materializes the whole stream.** *(nit / non-blocking)*

`Stream.sorted()` buffers every element before `limit()` can drop any, so a 500k-record stream still allocates 500k `PeekedRecord` wrappers and sorts them to return 100. The response size is now bounded — the work isn't.

Much smaller than the unbounded response I originally raised, and probably fine for an emulator. Mentioning it only because "protect the emulator from a big stream" was the point of the suggestion, and this delivers half of it. If comment 1 is fixed with a bounded priority queue — or by reading each shard's list from the end, since records are already appended in arrival order — both halves come for free.

pgermosen (approved):
@themark007  looks good, there are a few minor comments.


**`ShardId` is exactly right.** `PeekedRecord(String shardId, KinesisRecord record)` captures the shard inside the `flatMap` and `recordNode` emits it first — precisely the shape I suggested — and the test now asserts it. Pre-validating the shard and throwing `ResourceNotFoundException` for an unknown `ShardId` was a good extra.

**`Limit` needs one more pass.** I ran the endpoint rather than reading it. Three records in (`oldest`, `middle`, `newest`), `Limit=2` out:

```json
{"records":[
  {"ShardId":"shardId-000000000000","PartitionKey":"oldest","ApproximateArrivalTimestamp":"...T14:23:59.981Z"},
  {"ShardId":"shardId-000000000000","PartitionKey":"middle","ApproximateArrivalTimestamp":"...T14:24:00.009Z"}
]}
```

`newest` is dropped. Ascending sort + `limit` takes the **oldest** N, and with no pagination that means once a stream passes 100 records the endpoint returns the same stale head forever and new records are permanently invisible. Kinesis retention is 24h by default, so any stream with real traffic crosses that immediately. That's a regression against the old unbounded behavior for the UI use case in the title: unbounded was a scalability problem, but the newest records were at least present.

Details and a suggested fix in comment 1. Comments 2-4 are minor.

**One change requested (comment 1)** — I'd rather fix the ordering now than ship it and have the UI work around it. Everything else here is discussion.

Still open from the first review: the PR description is the unfilled template, and the LocalStack-compatible vs Floci-native question on the javadoc.

themark007:
Hi  @pgermosen 

I've addressed the latest review feedback and pushed the updates:

* Changed the inspection endpoint to return the **newest** `Limit` records while preserving chronological order in the response.
* Updated the integration test to verify the returned records and their ordering.
* Completed the PR description/template.

Could you please take another look when you have a chance? Thanks again for all the detailed feedback—it was very helpful!

@hectorvent @pgermosen @okinaka @electrum @abanna @jvanzyl

docs/services/kinesis.md
```diff
@@ -37,6 +37,18 @@
...
+## Local Inspection Endpoints
+
+These endpoints are Floci-local read-only helpers for the UI and tests. AWS SDK
+traffic should continue to use the JSON 1.1 Kinesis API on `/`.
+
+| Endpoint | Description |
+|---|---|
+| `GET /_aws/kinesis/streams` | List streams in the resolved region with shard, mode, tag, and record counts |
+| `GET /_aws/kinesis/records?StreamName=<name>` | Peek up to 100 stream records with shard attribution without consuming them |
+| `GET /_aws/kinesis/records?StreamName=<name>&Limit=<n>` | Peek up to `n` records, capped at 1000 |
```
pgermosen, line 49: **The docs table wasn't updated for the new semantics.**

`38bdd011` changed `/records` from returning the *oldest* N to the *newest* N — the user-facing contract — but the table still reads:

| `GET /_aws/kinesis/records?StreamName=<name>` | Peek up to 100 stream records with shard attribution without consuming them |
| `GET /_aws/kinesis/records?StreamName=<name>&Limit=<n>` | Peek up to `n` records, capped at 1000 |

Neither row says *which* records. Given how carefully this PR has documented the rest of the endpoint, worth saying it explicitly — it's the difference between a UI that shows live data and one that shows a frozen head:

```suggestion
| `GET /_aws/kinesis/records?StreamName=<name>` | Peek the 100 most recent stream records with shard attribution, oldest-first, without consuming them |
| `GET /_aws/kinesis/records?StreamName=<name>&Limit=<n>` | Peek the `n` most recent records, capped at 1000 |
```

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -477,6 +486,61 @@ public Map<String, Object> getRecords(String shardIterator, Integer limit, Strin
...
+     * <p>With no pagination cursor, {@value MAX_INSPECTION_RECORD_LIMIT} is a hard ceiling
+     * on the total number of records reachable in a single call regardless of the {@code limit}
+     * parameter.
+     */
+    public List<PeekedRecord> peekRecords(String streamName, String shardId, Integer limit, String region) {
+        KinesisStream stream = resolveStream(streamName, region);
+        int resolvedLimit = resolveInspectionRecordLimit(limit);
+        if (resolvedLimit == 0) {
+            return List.of();
+        }
```
pgermosen, line 504: **The zero-limit branch is unreachable, and the javadoc now describes behavior no caller can observe.** *(nit)*

`peekRecords` has exactly one caller — `KinesisInspectionController:80` — and the controller now rejects `Limit <= 0` with a 400 before the service is reached. So `resolveInspectionRecordLimit`'s `Math.max(0, ...)` can never return 0, and this early-return is dead.

Harmless as defence-in-depth, but the javadoc on `resolveInspectionRecordLimit` — *"Both ends are clamped silently, consistent with the behaviour of `getRecords` and `getRecordsForAccount`"* — now describes something that cannot be observed through the API, since the real contract is a 400. A future reader comparing the two will be misled about which one wins. Either drop the zero handling and let the controller own the contract, or add a sentence noting the controller rejects non-positive limits before this is reached.

Separately, same method: the heap comparator uses `nullsFirst` while the final display sort uses `nullsLast`, so a record with a null `ApproximateArrivalTimestamp` is treated as *oldest* for eviction but rendered as *newest*. Almost certainly unobservable — `PutRecord` always sets the timestamp — but the two comparators sitting 13 lines apart with opposite null policies reads like an oversight rather than a decision.

pgermosen (approved):
Looks good


Two minor items below, neither blocking.

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -477,6 +486,61 @@ public Map<String, Object> getRecords(String shardIterator, Integer limit, Strin
...
+
+        if (shardId != null && !shardId.isBlank()) {
+            boolean exists = stream.getShards().stream().anyMatch(shard -> shard.getShardId().equals(shardId));
+            if (!exists) {
+                throw new AwsException("ResourceNotFoundException", "Shard " + shardId + " not found", 400);
```
pgermosen, line 509: This still throws `AwsException`, which the app-wide `AwsExceptionMapper` renders as `{"__type": "ResourceNotFoundException", "message": ...}` — the shape the sibling `Limit`/`StreamName` guards in `KinesisInspectionController` were just changed *away* from in this same commit. Same issue applies to the `resolveStream()` call two lines up (line 500) for an unknown `StreamName`. For full consistency with the controller's plain `{"message": ...}` 400s (and with the class Javadoc's claim that this endpoint doesn't follow AWS error-response conventions), these two paths would need to surface as plain 4xx bodies too — e.g. by having the controller check stream/shard existence itself rather than relying on the AWS-shaped exception from the service layer.

pgermosen:
@themark007 

One loose end from this fix round: "unify error shapes" only covers two of the four error paths on `/_aws/kinesis/records`.

`StreamName`-blank and `Limit <= 0` now return the plain `{"message": ...}` shape. But an unknown `ShardId` (`KinesisService.java:509`) and an unknown `StreamName` (`resolveStream()`, `KinesisService.java:562`, called at the top of `peekRecords()`) still throw `AwsException`, which the app-wide `AwsExceptionMapper` renders as `{"__type": "ResourceNotFoundException", "message": ...}` — the shape this commit moved away from for the other two guards, and inconsistent with the class Javadoc's claim that this endpoint doesn't follow AWS error-response conventions.

Worth having the controller (or a local check before calling into the service) surface those two as plain 4xx bodies too, for consistency across all four error paths on this endpoint.

abanna (approved):
Thank you for adding the Kinesis inspection surface and for iterating on the earlier feedback. I physically ran the focused inspection tests—all 6 passed—and CI is green. I do not see a correctness blocker. The remaining error-shape and performance discussions are nonblocking; please clarify in the docs that `Limit` returns the newest N records, oldest-first, or explicitly disposition that thread. Approved, and thank you for the contribution.

pgermosen (approved):
Re-reviewed the latest round (`fix(kinesis): address reviewer comments on inspection endpoint`) against the three review comments it claims to close, since this is already at 3x APPROVED and green CI — checking whether it's genuinely done, not starting a fresh review.

**Comment 1 (Limit sort order) and Comment 3 (clamping policy) are both correctly fixed.** Traced `peekRecords()`: the min-heap `PriorityQueue<PeekedRecord>` with an `oldestFirst` comparator correctly implements top-N-by-recency (`poll()` evicts the oldest when size exceeds `resolvedLimit`), then re-sorts ascending before returning — genuinely retains the newest N records in chronological order, not just the first N encountered. `resolveInspectionRecordLimit()` now clamps silently via `Math.max(0, Math.min(limit, MAX))` instead of throwing, matching `getRecords`/`getRecordsForAccount`'s existing behavior. The strengthened `recordsEndpointHonorsLimit` test asserts which two records come back (not just count), so this is a real regression guard, not just a shape check.

**Comment 2 (unify error shapes) is only partially fixed** — the two validations the commit explicitly targeted (missing `StreamName`, `Limit <= 0`) do now share the plain `{"message": ...}` shape, confirmed by `recordsEndpointRequiresStreamName`/`recordsEndpointRejectsInvalidLimit` asserting the exact body. But `peekRecords()` (`KinesisService.java:509`) and the `resolveStream()` an unknown `StreamName` goes through (`KinesisService.java:562`, called at the top of `peekRecords()`) both still throw plain `AwsException`, which the globally-registered `@Provider AwsExceptionMapper` (applies to *every* JAX-RS endpoint app-wide, not just the AWS wire-protocol handlers) converts to `{"__type": "ResourceNotFoundException", "message": ...}` — the exact shape this fix round says it moved away from. So the endpoint the class-level Javadoc says "do[es] not follow AWS error-response conventions" still returns the AWS `__type` convention for an unknown `ShardId` or unknown `StreamName`, inconsistently with its own `Limit`/`StreamName`-blank guards on the same endpoint. Circumstantial evidence this was missed rather than intentional: `recordsEndpointRejectsUnknownShard` is the one error-path test in the new suite that asserts only `statusCode(400)` and conspicuously not `.body("message", ...)`, unlike its two siblings.

Ran the full `Kinesis*Test` glob myself: 98/98, no regressions. `make docs-check` exits clean, no CODEOWNERS-sensitive paths, no AI trailers, commits Conventional-Commits compliant. Everything else — `listStreamDetails()`'s region-prefix filter matches `regionKey()`'s actual key format, the `RecordCount`/`OpenShardCount` aggregates are straightforward stream-shard rollups with no off-by-one — checks out.

Not blocking on its own (status code is still correct in both cases, and the endpoint is explicitly non-wire-protocol so there's no AWS-fidelity stake), but worth closing out given the fix round specifically set out to unify these shapes and left two of the four error paths on this same endpoint using the old one.

PR #1839, opened 2026-07-11
fix(cognito): prefer phone delivery for signup verification

codingkiddo:
## Summary

Updates Cognito `SignUp` verification delivery-target selection to prefer phone/SMS when both `email` and `phone_number` are present and the user pool auto-verifies both attributes.

Previously, Floci selected the delivery target based on the order of `AutoVerifiedAttributes`. When `email` appeared before `phone_number`, `CodeDeliveryDetails` returned email delivery even though AWS prefers phone/SMS for the initial sign-up verification in this case.

Closes #1657

## Type of change

* [x] Bug fix (`fix:`)
* [ ] New feature (`feat:`)
* [ ] Breaking change (`feat!:` or `fix!:`)
* [ ] Docs / chore

## AWS Compatibility

Incorrect behavior:

For a user pool with `AutoVerifiedAttributes` containing both `email` and `phone_number`, signing up with both attributes could return `CodeDeliveryDetails` for email delivery depending on the configured attribute order.

Updated behavior:

`SignUp` now prefers `phone_number` / `SMS` delivery when both email and phone number are available and both are configured for auto-verification. Email remains the fallback when phone delivery is not applicable.

## Checklist

* [ ] `./mvnw test` passes locally
* [x] New or updated integration test added
* [x] Commit messages follow [[Conventional Commits](https://www.conventionalcommits.org/)](https://www.conventionalcommits.org/)

PR #1881, opened 2026-07-15
fix(cognito): mark verified contact attribute on ConfirmSignUp

### LINKED ISSUE #1654: [FEAT] Cognito ConfirmSignUp does not set email_verified or phone_number_verified

b6k-dev:
## Summary

Closes #1654

Change summary:
- Mark email_verified / phone_number_verified on ConfirmSignUp per the delivered contact
- Add unit tests in `CognitoServiceTest`
- Add SDK compatibility tests in `CognitoConfirmSignUpVerificationTest`

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

`ConfirmSignUp` succeeded but left `email_verified` and `phone_number_verified` unchanged, breaking downstream flows that depend on verified aliases or contact verified checks.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CognitoConfirmSignUpVerificationTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CognitoConfirmSignUpVerificationTest.java
@@ -1,0 +1,250 @@
+package com.floci.test;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+import software.amazon.awssdk.services.cognitoidentityprovider.CognitoIdentityProviderClient;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.AdminGetUserResponse;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.AttributeType;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.CreateUserPoolResponse;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.VerifiedAttributeType;
+
+import java.net.URI;
+import java.net.URLEncoder;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Compatibility tests for Cognito ConfirmSignUp verified attribute updates (Issue #1654).
+ */
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+class CognitoConfirmSignUpVerificationTest {
+
+    private static final ObjectMapper JSON = new ObjectMapper();
+    private static final HttpClient HTTP = HttpClient.newBuilder()
+            .connectTimeout(Duration.ofSeconds(10))
+            .build();
+    private static final Pattern SIX_DIGIT_CODE_PATTERN = Pattern.compile("\\b(\\d{6})\\b");
+    private static final String PASSWORD = "Passw0rd!";
+
+    private static CognitoIdentityProviderClient cognitoClient;
+    private String poolId;
+
+    @BeforeAll
+    static void setup() {
+        cognitoClient = TestFixtures.cognitoClient();
+    }
+
+    @AfterAll
+    static void cleanup() {
+        if (cognitoClient != null) {
+            cognitoClient.close();
+        }
+    }
+
+    @AfterEach
+    void tearDown() {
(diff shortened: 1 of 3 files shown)


hectorvent:
Thank you for this, and sorry it waited so long for a human reply. The fix itself is right: it delegates to `resolveSignUpDeliveryTarget` so ConfirmSignUp always verifies whatever SignUp actually delivered to, the attribute naming matches AWS (`email_verified` / `phone_number_verified` as string "true"), gating on AutoVerifiedAttributes is correct, and leaving `AdminConfirmSignUp` alone is right too.

**(blocking)** Main moved under you: #1839 made phone always win when a pool auto verifies both contacts. Your implementation still behaves correctly, but one test encodes the old order sensitive expectation. On current main I get 117 passing and 1 failing: `confirmSignUpMarksOnlyEmailVerifiedWhenBothAutoVerifiedAndEmailIsFirstDeliveryTarget`, "expected: <true> but was: <null>" (line 2152). A rebase plus flipping that case to expect `phone_number_verified` should be all it needs.

**(follow-up, separate PR)** The now reachable `AliasExistsException` / `ForceAliasCreation` path on ConfirmSignUp, per [botocore cognito-idp/2016-04-18](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json); #1849 circles the same ground.

One blocker, the rest is solid; dropping the unrelated import expansion would also ease Cognito merge traffic.

hectorvent:
Thank you, that is exactly it. The branch is on current main now and the assertion flip landed: `confirmSignUpMarksOnlyPhoneNumberVerifiedWhenBothAutoVerifiedEvenWhenEmailListedFirst` expects phone regardless of list order, which is the right shape after #1839. `CognitoServiceTest` here: 120 passing, 0 failing.

The null guard you added on top is a real fix too. `resolveSignUpDeliveryTarget` genuinely returns null when no auto verified attribute matches a user attribute, so the unguarded `attributeName()` call would have thrown for an admin created user with no email, and your new case pins it.

**(follow-up, separate PR)** `AliasExistsException` and `ForceAliasCreation` are declared on ConfirmSignUp ([cognito-idp service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json)) and still unimplemented, and flipping a verified alias flag is what makes that path reachable. #1849 circles the same ground.

No blockers from my side.

hectorvent (approved):

PR #1883, opened 2026-07-15
feat(apigateway): API key management — UpdateApiKey, DeleteApiKey, UsagePlan custom-id, generateDistinctId

dennisfoconnor:
## Summary

Adds full API key lifecycle management and two compatibility fixes for API Gateway v1.

- **`UpdateApiKey`** (`PATCH /apikeys/{id}`): update `/name`, `/description`, `/enabled` via `patchOperations`
- **`DeleteApiKey`** (`DELETE /apikeys/{id}`): 202 on success, 404 if not found
- **`description` field on `ApiKey`**: persisted on create, returned in responses, patchable via UpdateApiKey
- **`CreateUsagePlan` `_custom_id_` tag + tag persistence**: presence of `_custom_id_` in the `tags` map sets a deterministic plan ID; all tags are persisted and returned — mirrors existing `createRestApi` behaviour
- **`generateDistinctId` fix in `CreateApiKey`**: corrects undocumented AWS behaviour where the default (absent) is `id == value`, not distinct; only explicit `true` produces a distinct short-token id

## Type of change

- [x] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**UpdateApiKey / DeleteApiKey:** Implemented against the AWS API Gateway v1 REST API wire protocol. Patch paths (`/name`, `/description`, `/enabled`) and response shapes match the AWS SDK v2 Java client and AWS CLI v2.

**UsagePlan tags:** AWS `CreateUsagePlan` accepts a `tags` map in the request body and returns it in the response. This change aligns the emulator with that contract. `_custom_id_` is a floci-specific extension stored alongside other tags.

**`generateDistinctId` default (bug fix):** The previous implementation treated an absent `generateDistinctId` as `true` (distinct id and value). Testing against AWS confirmed the actual default is that `id` and `value` are the same string — only explicit `generateDistinctId: true` produces a distinct id. This commit corrects that behaviour.

*Here is the RAW request and response body from the API Call to AWS:*

**Request**
```
POST https://apigateway.us-east-1.amazonaws.com/apikeys
{
     "name": "floci-verify-absent",
     "enabled": true
}
```
**Reponse**
```
{
    "id": "AqJ4QKTsB11IgskqAXPcH5ekX79Rvx3L7YFD7gnC",
    "value": "AqJ4QKTsB11IgskqAXPcH5ekX79Rvx3L7YFD7gnC",
    "name": "floci-verify-absent",
    "enabled": true,
    "createdDate": "2026-08-07T15:25:35-04:00",
    "lastUpdatedDate": "2026-08-07T15:25:35-04:00",
    "stageKeys": []
}
```

## Checklist


hectorvent:
Thank you for this, and sorry it sat so long. UpdateApiKey and DeleteApiKey are a real gap and the shapes match the model: `DELETE /apikeys/{api_Key}` really is a 202 and both operations declare your error codes ([apigateway service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/apigateway/2015-07-09/service-2.json)).

**(blocking)** Commits cd76b12, 64b920e, 46c64a9, 82d5873, a146cdf, f0e3aa5, dc6ba54, e95b84d carry an AI co-author trailer; this repo keeps commit attribution with human contributors (see AGENTS.md). Would you amend or squash those commits and force push so the attribution is yours alone?

**(blocking)** `_custom_id_` moved on since you opened this: #2045 made `floci:override-id` the reserved key, routed through `ReservedTags.extractOverrideApiId`, which validates and consumes it. Could the usage plan use that helper too, asserting the key is absent from response tags?

**(blocking)** The `generateDistinctId` default is unverifiable here (moto ignores it; LocalStack delegates to moto), and it makes the key's id the secret itself. Please paste the raw AWS request and response into the description.

Three blockers above, then this is in good shape.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
```diff
@@ -687,13 +702,39 @@ public void deleteApiKey(String region, String apiKeyId) {
         apiKeyStore.delete(apiKeyGlobalKey(region, apiKeyId));
```
abanna, line 736: Nice to see these filled in — the `generateDistinctId` table in the docs is a genuinely useful bit of archaeology.

`deleteApiKey` only removes the entry from `apiKeyStore`, though, and `UsagePlanKey` holds its own **copy** of the value (`createUsagePlanKey` does `usagePlanKey.setValue(apiKey.getValue())` at line 780). The data plane authenticates against that copy, not against the key itself:

```java
for (UsagePlanKey planKey : apiGatewayService.getUsagePlanKeys(region, plan.getId())) {
    if (keyHeader.equals(planKey.getValue())) {
        return planKey.getValue();
    }
}
```

So a key that's been through `DeleteApiKey` still authenticates requests indefinitely — the delete removes it from `GetApiKeys` while leaving it working as a credential, and it's still listed by `GetUsagePlanKeys`. In AWS, deleting a key detaches it from every usage plan and it stops working immediately. Revoking a leaked key is the main reason anyone calls `DeleteApiKey`, so this is the case where the divergence hurts most.

The delete needs to sweep the associations too — something like:

```java
public void deleteApiKey(String region, String apiKeyId) {
    getApiKey(region, apiKeyId);
    for (UsagePlan plan : getUsagePlans(region)) {
        deleteUsagePlanKey(region, plan.getId(), apiKeyId);   // no-op when absent
    }
    apiKeyStore.delete(apiKeyGlobalKey(region, apiKeyId));
}
```

The same snapshot design bites the new `UpdateApiKey` in a second way: `resolveApiKeyForRequest` never reads `enabled`, even though its javadoc says "if a matching **enabled** key is found". So `--patch-operations op=replace,path=/enabled,value=false` updates the stored key and changes nothing about whether requests are accepted. That gap existed before, but this PR is what makes `enabled` mutable after creation, so it becomes reachable here.

Two integration cases would pin both: delete a key and assert a request carrying its value now returns `403`, and disable a key and assert the same.

dennisfoconnor: Thanks for pointing this out. I made the following updates to address your comments:

- `deleteApiKey` now detaches the key from every usage plan in the region before removing it, so adeleted key is no longer listed by `GetUsagePlanKeys` or recognised on the data plane.
- `resolveApiKeyForRequest` now requires the key to still exist and be enabled, via a non-throwing `findApiKey` lookup — so its javadoc is finally true, and patching `/enabled` to `false` takes effect immediately.
  
I wasn't able to write the 403 assertion test you suggested. Floci doesn't implement the `apiKeyRequired` gate at all — `MethodConfig` has no such field, nothing parses, persists or returns it, and `dispatch()` never consults one. `resolvedApiKey` is only used to populate `requestContext.identity.apiKey` for the proxy event and authorizer context. So there is no code path that can return 403 for a revoked key today, and a test asserting one would fail
against `main` for a reason unrelated to this PR.

What the tests assert instead is the behaviour that does exist: a revoked key yields a null `identity.apiKey`, and the key disappears from `GetUsagePlanKeys`. The docs now state the limitation explicitly so nobody reads the fix as "revoked keys are rejected".

dennisfoconnor:
Thanks for the feedback. 
My commits addressed the remainder of the comments. While updating the docs I noticed I'd claimed tags come back from `GetUsagePlan`, which isn't implemented — corrected to `CreateUsagePlan` and `GetUsagePlans` only.

I've also updated the description with the raw request and response for the create API key call without `generateDistinctId`.

pgermosen (approved):
Thank you for such a thoroughly-verified contribution, including the real AWS request/response evidence in the PR description — that made this a pleasure to check. Went through it against the AWS API Gateway v1 reference directly.

`DeleteApiKey`'s 202/404 behavior, `UpdateApiKey`'s response shape and its `/name`/`/description`/`/enabled` patch paths, and `CreateUsagePlan`'s `tags` request/response field all match the documented contract exactly. The `generateDistinctId` default fix lines up with the parameter semantics AWS documents, and is backed by your own pasted real-AWS evidence, which is much appreciated.

Also really glad to see the auth-bypass gap abanna raised got a proper fix rather than a surface patch — `deleteApiKey` now detaches the key from every usage plan, and the data-plane lookup independently re-validates against the live key (must exist and be enabled) before accepting a match. That's genuine defense in depth.

One small, non-blocking thing for a future pass: `UpdateApiKey` doesn't yet support the `/customerId` patch path (a real, `replace`-supported field per the AWS Patch Operations reference) — it's silently a no-op today. Looks like a pre-existing gap rather than something this PR introduced, so not a reason to hold this up.

Approving — nice work all around.

PR #1887, opened 2026-07-16
feat(rds-data): adding data-api preparedstatement() support

QAnders:
## Summary

Adding PreparedStatement() function support for Data-API across all DB drivers.

## Type of change

- [ ] Bug fix (`fix:`)
- [X] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Adding PreparedStatement

## Checklist


src/main/java/io/github/hectorvent/floci/services/rdsdata/RdsDataSqlParameters.java
```diff
@@ -0,0 +1,261 @@
...
+
+    private static void bindString(PreparedStatement statement, int index, String name, String value, String typeHint)
+            throws SQLException {
+        String hint = typeHint == null ? "" : typeHint.toUpperCase();
+        try {
+            switch (hint) {
+                case "DECIMAL" -> statement.setBigDecimal(index, new BigDecimal(value));
+                case "TIMESTAMP" -> statement.setTimestamp(index, Timestamp.valueOf(value));
+                case "DATE" -> statement.setDate(index, Date.valueOf(value));
+                case "TIME" -> statement.setTime(index, Time.valueOf(value));
```
pgermosen: The `TIME` case here doesn't match AWS's documented accepted format. Per the [SqlParameter reference](https://docs.aws.amazon.com/rdsdataservice/latest/APIReference/API_SqlParameter.html), `TIME` accepts `HH:MM:SS[.FFF]` — same optional-fractional-seconds shape as `TIMESTAMP`. `Timestamp.valueOf()` handles that (verified directly), but `Time.valueOf()` doesn't: `Time.valueOf("14:30:00.123")` throws `NumberFormatException`, which gets caught below and turned into a `BadRequestException` — so a documented-valid AWS input gets rejected here. No test exercises the `TIME` hint at all. Truncating any fractional part before `Time.valueOf()` (the way `Timestamp` doesn't need to) would fix it.

src/main/java/io/github/hectorvent/floci/services/rdsdata/RdsDataService.java
```diff
@@ -206,26 +211,51 @@ public ObjectNode rollbackTransaction(JsonNode request) {
...
-            boolean hasResultSet = statement.execute(sql);
-            ObjectNode response = objectMapper.createObjectNode();
-            if (hasResultSet) {
-                try (ResultSet rs = statement.getResultSet()) {
-                    ResultSetMetaData meta = rs.getMetaData();
-                    if (includeMetadata) {
-                        response.set("columnMetadata", columnMetadata(meta));
-                    }
-                    response.set("records", records(rs, meta));
+    private ObjectNode executeOnConnection(Connection connection, DatabaseEngine engine, String sql,
```
pgermosen, line 214: This still has the gap Copilot flagged: when `parameters` is completely omitted or empty, `RdsDataSqlParameters` never runs at all, so SQL containing `:name` markers goes to `Statement.execute()` as literal text. I traced it through — the resulting `SQLException` does get caught and turned into a 400 (`databaseError()`), so it's not a crash, but it's a different, less accurate error (`DatabaseErrorException` with a raw driver message) than the same underlying problem gets when at least one parameter is supplied: `bind()`'s missing-value check produces a clean `BadRequestException: SQL statement references parameter :id but no matching value was supplied.` I checked `RdsDataServiceTest` and `RdsDataSqlParametersTest` and neither covers the all-omitted case. Parsing unconditionally (as Copilot suggested) and letting `bind()`'s existing check handle the missing-parameter case either way would make the two paths consistent.

src/main/java/io/github/hectorvent/floci/services/rdsdata/RdsDataSqlParameters.java
```diff
@@ -0,0 +1,261 @@
...
+        } else if (value.has("stringValue")) {
+            bindString(statement, index, name, value.get("stringValue").asText(), typeHint);
+        } else {
+            throw new AwsException("BadRequestException",
+                    "Parameter :" + name + " has no supported value field.", 400);
+        }
+    }
+
+    private static void bindString(PreparedStatement statement, int index, String name, String value, String typeHint)
+            throws SQLException {
```
pgermosen, line 180: Small one Copilot also flagged and is still open: `typeHint.toUpperCase()` uses the default JVM locale rather than `Locale.ROOT`. Low real-world impact since clients send hints pre-uppercased, but worth the one-line fix for correctness under a Turkish-locale JVM.

pgermosen (commented):
Nice feature, and the parser is genuinely careful about the tricky bits (comments, dollar-quoting, casts, backtick identifiers). Went through both commits, and this time checked the AWS docs directly for the typeHint/Field shape rather than just re-verifying the two Copilot findings against the code.

The `Field` value variants (`booleanValue`, `longValue`, `doubleValue`, `blobValue`, `arrayValue`, `stringValue`, `isNull`) and the six `typeHint` values (`DATE`, `DECIMAL`, `JSON`, `TIME`, `TIMESTAMP`, `UUID`) both match the [SqlParameter](https://docs.aws.amazon.com/rdsdataservice/latest/APIReference/API_SqlParameter.html)/[Field](https://docs.aws.amazon.com/rdsdataservice/latest/APIReference/API_Field.html) references exactly, and typeHint is correctly scoped to only apply on the `stringValue` branch, matching the docs' "the corresponding String parameter value is sent as..." framing.

One thing that surfaced from checking the docs' stated formats against the actual JDBC calls: the `SqlParameter` reference gives `TIME`'s accepted format as `HH:MM:SS[.FFF]` — fractional seconds optional — same shape as `TIMESTAMP`'s `YYYY-MM-DD HH:MM:SS[.FFF]`. `Timestamp.valueOf()` handles that fine (verified: `Timestamp.valueOf("2026-01-01 14:30:00.123")` parses). `Time.valueOf()` doesn't — I ran it directly and `Time.valueOf("14:30:00.123")` throws `NumberFormatException`. So a client sending a documented-valid `TIME` value with fractional seconds gets a `BadRequestException` here where real AWS would accept it. It's caught rather than crashing, but it's a false rejection of valid input, and there's no test at all for the `TIME` hint (checked both test files) so it slipped through. Truncating to `HH:mm:ss` before calling `Time.valueOf()` would close it.

Two more comments inline, both still open from Copilot's review and untouched by the follow-up commit.

src/main/java/io/github/hectorvent/floci/services/rdsdata/RdsDataSqlParameters.java
QAnders: Fixed!

QAnders:
> Nice feature, and the parser is genuinely careful about the tricky bits (comments, dollar-quoting, casts, backtick identifiers). Went through both commits, and this time checked the AWS docs directly for the typeHint/Field shape rather than just re-verifying the two Copilot findings against the code.
> 
> The `Field` value variants (`booleanValue`, `longValue`, `doubleValue`, `blobValue`, `arrayValue`, `stringValue`, `isNull`) and the six `typeHint` values (`DATE`, `DECIMAL`, `JSON`, `TIME`, `TIMESTAMP`, `UUID`) both match the [SqlParameter](https://docs.aws.amazon.com/rdsdataservice/latest/APIReference/API_SqlParameter.html)/[Field](https://docs.aws.amazon.com/rdsdataservice/latest/APIReference/API_Field.html) references exactly, and typeHint is correctly scoped to only apply on the `stringValue` branch, matching the docs' "the corresponding String parameter value is sent as..." framing.
> 
> One thing that surfaced from checking the docs' stated formats against the actual JDBC calls: the `SqlParameter` reference gives `TIME`'s accepted format as `HH:MM:SS[.FFF]` — fractional seconds optional — same shape as `TIMESTAMP`'s `YYYY-MM-DD HH:MM:SS[.FFF]`. `Timestamp.valueOf()` handles that fine (verified: `Timestamp.valueOf("2026-01-01 14:30:00.123")` parses). `Time.valueOf()` doesn't — I ran it directly and `Time.valueOf("14:30:00.123")` throws `NumberFormatException`. So a client sending a documented-valid `TIME` value with fractional seconds gets a `BadRequestException` here where real AWS would accept it. It's caught rather than crashing, but it's a false rejection of valid input, and there's no test at all for the `TIME` hint (checked both test files) so it slipped through. Truncating to `HH:mm:ss` before calling `Time.valueOf()` would close it.
> 
> Two more comments inline, both still open from Copilot's review and untouched by the follow-up commit.

Thanks for checking!

Good points, @pgermosen!

I've tried to address them in my new commit!

pgermosen (approved):
Thanks for addressing both points so thoroughly! Re-checked the fixes directly rather than taking the commit message's word for it.

The parameters-omitted inconsistency is genuinely closed: `executeOnConnection` no longer branches on `parameters.isEmpty()` — it always goes through the parse/bind path now, so an omitted `parameters` field gets the same clean `BadRequestException` as a partially-populated one. And the `TIME` fix (stripping fractional seconds before `Time.valueOf()`) is simple and correct, with a nice end-to-end test covering both the fractional and plain forms.

Approving — great follow-through.

PR #1890, opened 2026-07-16
fix(security): map EXTRA_CORS_* env variables to enable global CORS preflight

### LINKED ISSUE #1865: [BUG] S3 CORS preflight (OPTIONS) always returns 403, ignoring EXTRA_CORS_ALLOWED_ORIGINS
## Service

S3

## AWS API Action

N/A (CORS preflight `OPTIONS` request, prior to `PutObject` / `GetObject` etc.)

## Expected behavior

When `EXTRA_CORS_ALLOWED_ORIGINS` is set (e.g. `EXTRA_CORS_ALLOWED_ORIGINS=*`) and the target
bucket has no bucket-level CORS configuration (`PutBucketCors`), a CORS preflight `OPTIONS`
request should succeed (200) by falling back to the global origins configured via
`EXTRA_CORS_ALLOWED_ORIGINS`, the same way the actual request (`GET`/`PUT`/etc.) does.

This is the documented intent of `GlobalCorsFilter` (added in #849): service-level CORS should
take precedence, and the filter should only fill in missing CORS headers when the service
itself doesn't provide them.

## Actual behavior

The preflight `OPTIONS` request always returns 403 with `CORSResponse` / "This CORS request is
not allowed", regardless of the `Origin` header value (tried `http://localhost:3030`, `*`,
`null`, `http://example.com`), regardless of whether the bucket exists, and regardless of
virtual-hosted-style vs path-style URL. `EXTRA_CORS_ALLOWED_ORIGINS=*` has no effect on the
preflight response.

Root cause (confirmed by reading the source):

- `S3Controller`'s `@OPTIONS` handlers (`handleOptionsBucket` / `handleOptionsObject`) evaluate
  the preflight exclusively against the bucket-level CORS rules via `s3Service.evaluateCors(...)`.
  If no bucket-level CORS rule matches (e.g. no `PutBucketCors` was ever called), it returns 403
  `CORSResponse` directly from the controller — this response is returned to the client
  immediately.
- `GlobalCorsFilter` (added in #849) is a `@PreMatching` `ContainerRequestFilter` intended to
  `abortWith(...)` a 200 response early for `OPTIONS` preflight requests when the global origins
  list allows the `Origin`. However, in practice this early-return branch does not fire, so the
  request always reaches `S3Controller`'s `@OPTIONS` handler and gets rejected with 403.
- Notably, the same fallback design *does* work for the actual request (non-`OPTIONS`):
  `GlobalCorsFilter`'s `ContainerResponseFilter` side uses `putIfAbsent` to add CORS headers onto
  the response after the fact, which succeeds because it doesn't depend on short-circuiting the
  request. Only the preflight (`abortWith`) path is broken.
- This means any consumer that relies solely on `EXTRA_CORS_ALLOWED_ORIGINS` (without registering
  per-bucket CORS rules via `PutBucketCors`) can never get past the browser's preflight check,
  even though the actual request would have been allowed.

## Reproduction

```bash
# Bucket exists, no PutBucketCors ever configured, EXTRA_CORS_ALLOWED_ORIGINS=* set on the container
curl -i -X OPTIONS "http://my-bucket.s3.localhost.floci.io:4566/some-key" \
  -H "Origin: http://localhost:3030" \
  -H "Access-Control-Request-Method: PUT" \
  -H "Access-Control-Request-Headers: content-type"

# => HTTP/1.1 403 Forbidden
# => <Error><Code>CORSResponse</Code><Message>This CORS request is not allowed</Message>...</Error>

# Same result with path-style URL, non-existent bucket, or Origin: *
```

Also reproducible without any bucket-level CORS rule and with `EXTRA_CORS_ALLOWED_ORIGINS=*`
explicitly verified inside the container (`docker compose exec floci env`).

## Environment

- Floci version / image tag: v1.5.31 (`floci/floci:latest` as of 2026-07-07)
- Java SDK version (if applicable): N/A (triggered from a browser via `fetch`/XHR preflight, and
  reproduced directly with `curl`)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (`docker compose`)

9046balaji:
### Summary
This PR fixes a bug where global CORS preflight `OPTIONS` requests returned `403 Forbidden` with the `CORSResponse` error (from the S3 controller) instead of succeeding. 

### Root Cause
1. In `GlobalCorsFilter.java`, the filter relies on the configuration values `extra-cors-allowed-origins`, `extra-cors-allowed-headers`, and `extra-cors-expose-headers` to check if global CORS is enabled and to match origins.
2. In `application.yml` and `test` environments, these settings were commented out, so they were never bound to environment variables like `EXTRA_CORS_ALLOWED_ORIGINS`.
3. As a result, setting `EXTRA_CORS_ALLOWED_ORIGINS=*` on the container had no effect. The global CORS filter remained disabled, and the preflight request reached the `S3Controller`, which returned a `403` error because the target bucket had no bucket-level CORS configured.

### Solution
1. Uncommented and mapped `extra-cors-*` properties to their respective environment variables in `src/main/resources/application.yml` and `src/test/resources/application.yml`:
   - `extra-cors-allowed-origins` mapped to `${EXTRA_CORS_ALLOWED_ORIGINS:}`
   - `extra-cors-allowed-headers` mapped to `${EXTRA_CORS_ALLOWED_HEADERS:}`
   - `extra-cors-expose-headers` mapped to `${EXTRA_CORS_EXPOSE_HEADERS:}`
2. Added regression tests to ensure the mapping works correctly and allows preflight requests on S3-like paths to succeed with a `204` status code.

### Changes
* `src/main/resources/application.yml`: Expose environment variables for global CORS.
* `src/test/resources/application.yml`: Add environment variable mappings for the test profile.
* `src/test/java/io/github/hectorvent/floci/core/common/GlobalCorsFilterIntegrationTest.java`: Added preflight tests for S3 paths (path-style and virtual-hosted style).
* `src/test/java/io/github/hectorvent/floci/core/common/GlobalCorsFilterAliasIntegrationTest.java`: [NEW] Added integration test class validating that `EXTRA_CORS_ALLOWED_ORIGINS` maps correctly.

Closes #1865.

---

### Update: Review Fixes
Based on the code review feedback, the following was added:
1. **Negative & default-path test cases** (commit `test(security): add negative preflight and default-disabled CORS tests`):
   - Added assertions to check that unlisted origins are rejected (they bypass the global filter and do not receive CORS headers).
   - Added `DisabledGlobalCorsFilterTest` in `GlobalCorsFilterIntegrationTest` to verify that when CORS is not configured (default blank/empty settings), the filter remains disabled and behaves correctly.

The Mandrel `java-version` workflow change has been dropped from this PR (it was unrelated to the CORS fix, per review) and the branch has been rebased onto current `main` to pick up unrelated fixes on `main` that were causing stale `Build and Test` failures here.


.github/workflows/release.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
       - uses: graalvm/setup-graalvm@6f3fa030c4b8f77c1f554a860f593a654538fa38 # v1
         with:
-          java-version: '25'
+          java-version: '24'
```
abanna, line 40: This PR downgrades the toolchain from Java 25 to 24 across four workflow files — `compatibility.yml`, `nightly.yml`, and both jobs in `release.yml` — which looks unrelated to the CORS env-var mapping the title describes.

`main` is on `java-version: '25'` in all of them, and `pom.xml` pins `<maven.compiler.release>25</maven.compiler.release>`, so this would build the released artifact on a JDK older than the project targets. `release.yml` is the one I'd flag hardest: a change that ships in a release shouldn't ride along in a security fix.

If a local Mandrel 24 was needed to get the native build working, that's worth its own PR and a conversation — the native image job here fails in 21 seconds, which suggests the downgrade isn't achieving what it was meant to anyway.

On the CORS change itself: `${EXTRA_CORS_ALLOWED_ORIGINS:}` with an empty default is a pattern that appears nowhere else in `application.yml` — there are currently zero `${VAR:}` empty-defaults on `main`. That matters because the property binds to `Optional<List<String>>` and the filter's kill switch is:

```java
boolean enabled() {
    return !disabled && !allowedOrigins.isEmpty();
}
```

If SmallRye resolves the empty value to `Optional.empty()` the default is unchanged; if it resolves to a single empty-string element, `allowedOrigins` becomes `{""}`, `enabled()` flips to `true`, and the filter starts intercepting preflights on a default install while matching no real origin. Previously the property was simply commented out, so the question didn't arise.

That's the case worth a test — which lines up with the other review note about the disabled-filter default path. An assertion that a preflight is *not* handled when no env var is set would settle it either way.

Last thing: the three `Build and Test` failures (`AutoScalingIntegrationTest.describeAutoScalingGroupsEmpty`, and two `Ec2IntegrationTest` network-interface pagination cases) don't touch CORS and look like drift from a stale base — a rebase onto current `main` should clear them.

pgermosen (changes requested):
Thanks for tracking down the root cause here — the CORS env-var mapping fix itself is solid and well-tested. This PR's scope should be just that: wiring `extra-cors-allowed-origins`/`extra-cors-allowed-headers`/`extra-cors-expose-headers` to their env vars in `application.yml`, closing #1865.

Could you roll back the `chore(ci): update mandrel setup-graalvm java-version to 24` commit and open it as a separate PR instead? As @abanna flagged, that change touches `.github/workflows/compatibility.yml`, `nightly.yml`, and both jobs in `release.yml` — the actual release pipeline — which is unrelated to the CORS fix and deserves its own focused review. `.github/` is also outside what I can approve on its own, so keeping it separate means this PR won't be blocked waiting on that review too.

Once that's split out, happy to take another look — and it'd be worth addressing the two test-coverage points Greptile raised (asserting an unlisted origin is rejected, and covering the disabled-filter default path) while you're at it.

pgermosen (approved):
Rebased this onto current main and dropped the unrelated Mandrel java-version downgrade — turns out that wasn't even fixing the native build, main builds fine at 25. Native and Build-and-Test are both green now. The two test-coverage points from the earlier review round were already covered in a follow-up commit, so this looks good to merge. I took over the branch directly (maintainer edits were enabled) to get it across the line.

PR #1891, opened 2026-07-16
feat(sqs): retain MessageGroupId on standard queue messages

### LINKED ISSUE #1496: [FEAT] SQS Standard queue supports passing "MessageGroupId" for Fair Queues feature
Service
SQS

API Action / Feature
SendMessage — MessageGroupId parameter on standard (non-FIFO) queues

AWS Documentation
[SendMessage](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html) — MessageGroupId parameter on standard (non-FIFO) queues
[ReceiveMessage](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html) — MessageGroupId system attribute

Why is this needed?
Our application sends messages to standard (non-FIFO) SQS queues using the AWS SDK's SendMessageRequest with MessageGroupId set to a caller-supplied correlation value (an EnvironmentId GUID):
```c#
SendMessageRequest sqsSendMessageRequest = new()
{
    QueueUrl = myQueueUrl,
    MessageBody = JsonSerializer.Serialize(message),
    MessageGroupId = myQueueUrl.EndsWith(".fifo") ? "1" : message.EnvironmentId
};
```

On ReceiveMessage, we retrieve that value back via --attribute-names MessageGroupId.
```c#
aws sqs receive-message \
  --queue-url <url> \
  --attribute-names MessageGroupId \
  --max-number-of-messages 1
``` 

Floci silently discards MessageGroupId on SendMessage for standard queues (which is correct per [AWS spec](https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-sns-standard-topics-sqs-fair-queues/)), so ReceiveMessage returns no MessageGroupId attribute.

Are you willing to contribute a PR?
- [ ] Yes
- [x] No

soreavis:
## Summary

Standard queues silently dropped `MessageGroupId` on SendMessage, so ReceiveMessage never returned it. Since the [July 2025 fair-queues launch](https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-sns-standard-topics-sqs-fair-queues/), standard queues accept `MessageGroupId` ([SendMessage reference](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html)) and return it as a message system attribute ([ReceiveMessage reference](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html)).

The fix is one site: store the value in the standard-queue branch of `SqsService.sendMessage`. Both protocol handlers already parse it on send and return it on receive, so single and batch sends on both protocols are covered. Delivery is untouched — group locking only runs in the FIFO claim path, so same-group standard messages still deliver in parallel, matching AWS ("messages with the same MessageGroupId can be processed in parallel", per the SendMessage reference).

The new integration test fails without the service change.

Closes #1496

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

Send with `MessageGroupId` to a standard queue, then `ReceiveMessage` with `MessageSystemAttributeNames=["MessageGroupId"]` returns the value. The response carries no `SequenceNumber` — that stays FIFO-only, matching AWS. Verified against the SendMessage API reference; wire shape exercised via the JSON-protocol integration test.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/SqsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/SqsService.java
@@ -479,8 +479,10 @@ public class SqsService implements Resettable {
             return message;
         }
 
-        // Standard queue
+        // Standard queue. MessageGroupId is retained for ReceiveMessage to
+        // return (fair queues); it has no effect on standard-queue delivery.
         Message message = new Message(body);
+        message.setMessageGroupId(messageGroupId);
         message.setAwsTraceHeader(awsTraceHeader);
         if (effectiveDelaySeconds > 0) {
             message.setVisibleAt(Instant.now().plusSeconds(effectiveDelaySeconds));
--- a/src/test/java/io/github/hectorvent/floci/services/sqs/SqsJsonProtocolTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/sqs/SqsJsonProtocolTest.java
@@ -359,7 +359,7 @@ class SqsJsonProtocolTest {
     }
 
     @Test
-    @Order(7)
+    @Order(9)
     void sendMessageToStandardQueueRetainsMessageGroupId() {
         String groupQueueName = QUEUE_NAME + "-group";
         String groupQueueUrl = given()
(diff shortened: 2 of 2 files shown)


hectorvent (approved):
Thanks @soreavis

PR #1893, opened 2026-07-16
fix(kms): reject mismatched keys in decryption requests

### LINKED ISSUE #1844: [BUG] KMS: Decrypt does not enforce the KeyId parameter — ciphertext wrapped under a different CMK decrypts instead of IncorrectKeyException
## Service

KMS

## AWS API Action

Decrypt

## Expected behavior

When `KeyId` is passed to `Decrypt` and the ciphertext blob was produced under a *different* symmetric CMK, real AWS KMS rejects the call with `IncorrectKeyException` ("The key ID in the request does not identify a KMS key that can perform this operation"). Passing `KeyId` on symmetric decrypt is AWS's documented server-side guard (best practice) against confused-deputy / wrong-key decryption.

## Actual behavior

Floci ignores the `KeyId` parameter on `Decrypt` and returns the plaintext (200) even when `KeyId` names a different CMK than the one that wrapped the blob. Same behavior with an alias-form `KeyId` (`alias/...`) pointing at the wrong key.

Everything adjacent works correctly (verified on 1.5.32): decrypt with the *correct* `KeyId` or alias succeeds, `EncryptionContext` mismatch is rejected with `InvalidCiphertextException`, and self-describing decrypt (no `KeyId`) resolves the right CMK.

## Reproduction

```bash
docker run -d -p 4566:4566 floci/floci:1.5.32
E=http://localhost:4566

KEY_A=$(aws --endpoint-url $E kms create-key --query KeyMetadata.KeyId --output text)
KEY_B=$(aws --endpoint-url $E kms create-key --query KeyMetadata.KeyId --output text)

aws --endpoint-url $E kms generate-data-key --key-id "$KEY_A" --key-spec AES_256 \
  --query CiphertextBlob --output text | base64 -d > blob.bin

# Real AWS: IncorrectKeyException. Floci: 200 + Plaintext.
aws --endpoint-url $E kms decrypt --ciphertext-blob fileb://blob.bin --key-id "$KEY_B"
```

Also reproducible with botocore (`decrypt(CiphertextBlob=..., KeyId=other_key)` → no error).

## Environment

- Floci version / image tag: `floci/floci:1.5.32` (`sha256:4f69631e5601...`)
- SDK: AWS CLI 2.x / botocore (Python)

## Context

Found while running an envelope-encryption adapter's integration suite differentially against LocalStack and floci — LocalStack 3.8.1 enforces this (`IncorrectKeyException`), floci does not. For frameworks that rely on `Decrypt`'s `KeyId` as the server-side confused-deputy guard, tests of that guard silently pass against floci without testing anything.

b6k-dev:
## Summary

Closes #1844

Change summary:

- Add KeyId validation in the decryption operation
- Add unit tests in `KmsServiceTest`
- Add SDK compatibility tests for Java and AWS CLI

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Per the [KMS Decrypt API spec](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html), "If you identify a different KMS key, the `Decrypt` operation throws an `IncorrectKeyException`" (HTTP 400). `Decrypt` silently ignored the `KeyId` request parameter and returned the plaintext even when `KeyId` named a different key than the one that wrapped the ciphertext.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
@@ -686,12 +686,31 @@ public class KmsService {
      * Single-pass decrypt + source-key-ARN resolution. {@link #decrypt} and
      * {@link #decryptToKeyArn} remain independent primitives — neither delegates here.
      */
-    public DecryptResult decryptAndResolveKey(byte[] ciphertext, Map<String, String> encryptionContext, String region) {
+    public DecryptResult decryptAndResolveKey(
+            byte[] ciphertext,
+            Map<String, String> encryptionContext,
+            String region,
+            String requestKeyId
+    ) {
         ParsedBlob parsed = parseBlob(ciphertext);
         if (!parsed.contextFingerprint.equals(contextFingerprint(encryptionContext))) {
             throw new AwsException("InvalidCiphertextException", "The ciphertext is invalid.", 400);
         }
         byte[] plaintext = Base64.getDecoder().decode(parsed.payload);
+
+        if (requestKeyId != null && !requestKeyId.isBlank()) {
+            KmsKey requestKey = resolveKey(requestKeyId, region);
+            if (!requestKey.getKeyId().equals(parsed.keyId)) {
+                throw new AwsException(
+                        "IncorrectKeyException",
+                        "The request was rejected because the specified KMS key cannot decrypt the data.",
+                        400
+                );
+            }
+
+            return new DecryptResult(plaintext, requestKey.getArn());
+        }
+
         String keyArn;
         try {
             keyArn = resolveKey(parsed.keyId, region).getArn();
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
@@ -649,6 +649,7 @@ public class KmsService {
 
     public byte[] encrypt(String keyId, byte[] plaintext, Map<String, String> encryptionContext, String region) {
         KmsKey kmsKey = resolveKey(keyId, region);
+        validateKeyIsUsableForCryptoOperations(kmsKey);
 
         byte[] nonceBytes = new byte[NONCE_BYTES];
         SECURE_RANDOM.nextBytes(nonceBytes);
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
```diff
@@ -665,12 +665,31 @@ public String decryptToKeyArn(byte[] ciphertext, String region) {
...
+            String requestKeyId
+    ) {
         ParsedBlob parsed = parseBlob(ciphertext);
         if (!parsed.contextFingerprint.equals(contextFingerprint(encryptionContext))) {
             throw new AwsException("InvalidCiphertextException", "The ciphertext is invalid.", 400);
         }
         byte[] plaintext = Base64.getDecoder().decode(parsed.payload);
+
+        if (requestKeyId != null && !requestKeyId.isBlank()) {
+            KmsKey requestKey = resolveKey(requestKeyId, region);
```
abanna, line 703: Good fix, and routing the `KeyId` through `resolveKey` rather than comparing raw strings is the right call — it means `--key-id alias/my-key` and a full key ARN both work instead of falsely reporting `IncorrectKeyException`.

Since you're adding validation to this path: neither `decryptAndResolveKey` nor `encrypt` checks the key's state. The only state guards in `KmsService` are in `enableKey` (line 529) and `rotateKeyOnDemand` (line 549) — the crypto operations themselves have none. So a key that's been disabled, or scheduled for deletion, still encrypts and decrypts normally:

```bash
aws kms disable-key --key-id "$KEY_ID"
aws kms decrypt --ciphertext-blob fileb://ct --key-id "$KEY_ID"   # succeeds in Floci
```

AWS rejects both — `DisabledException` for a disabled key, `KMSInvalidStateException` for one pending deletion, and the same on `Encrypt`. Disabling a key is the standard way to exercise a break-glass or key-rotation path, so an emulator that keeps decrypting makes that scenario untestable.

It's pre-existing rather than introduced here, so entirely reasonable as a follow-up. I raise it because this PR is the one that establishes "validate the key before decrypting", and `requestKey` is already in hand at exactly the point the check belongs:

```java
KmsKey requestKey = resolveKey(requestKeyId, region);
if (!requestKey.isEnabled()) {
    throw new AwsException("DisabledException",
            "The request was rejected because the specified KMS key is not enabled.", 400);
}
```

The `kms.bats` cases you added would extend naturally — disable the key between encrypt and decrypt, assert `DisabledException`.

b6k-dev: Fixed.

pgermosen (approved):
Thanks for this — really solid work. I checked the error handling directly against the [Decrypt](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html) and [ReEncrypt](https://docs.aws.amazon.com/kms/latest/APIReference/API_ReEncrypt.html) API references rather than just trusting the PR's own citation, and everything lines up precisely: `IncorrectKeyException`, `DisabledException`, and `KMSInvalidStateException` all match the documented text and status codes exactly, and the optional-`KeyId`-for-symmetric-ciphertext fallback mirrors AWS's documented behavior. The key-state check also covers every state this codebase actually models, not just the two called out by name.

One small, unrelated thing I noticed along the way: `resolveKey()`'s `NotFoundException` currently returns HTTP 404, while AWS documents 400 for both Decrypt and ReEncrypt — that's a pre-existing quirk on `main`, not something this PR introduced, but this PR's new code path does make it reachable via an explicit bad `KeyId`/`SourceKeyId`. Not blocking, just flagging in case it's worth a follow-up.

Approving — nice, well-verified fix.

PR #1896, opened 2026-07-16
fix(kinesis): reject records over the 1 MiB size limit

### LINKED ISSUE #1755: [BUG] Kinesis: PutRecord accepts records over the 1 MiB limit silently (no ValidationException)
**Service**
Kinesis

**AWS API Action**
PutRecord (PutRecords behaves the same)

**Expected behavior**
Real AWS Kinesis rejects any record whose data blob exceeds 1 MiB (1,048,576 bytes) with a `ValidationException` ("Value at 'data' failed to satisfy constraint: Member must have length less than or equal to 1048576"). Client code commonly branches on exactly this error (e.g. split/route oversized payloads), so the rejection is load-bearing behavior, not just validation strictness.

**Actual behavior**
floci accepts an oversized record silently — HTTP 200 with a SequenceNumber:

```
put_record(Data=<1,048,676 bytes>, PartitionKey="pk") ->
{"SequenceNumber": "1783355146142", "ShardId": "shardId-000000000000"}
```

Any client error-handling path keyed on the AWS ValidationException is unreachable against floci, which makes tests of that path pass falsely. Precedent: #775 (SQS SendMessage ignoring MaximumMessageSize) and the size-limit items in #768 were treated as bugs of the same class.

**Reproduction**
```bash
aws --endpoint-url=http://localhost:4566 kinesis create-stream --stream-name probe-stream --shard-count 1
sleep 2
python3 - <<'EOF'
import boto3
k = boto3.client("kinesis", endpoint_url="http://localhost:4566",
                 aws_access_key_id="test", aws_secret_access_key="test", region_name="us-east-1")
# 1 MiB + 100 bytes — real AWS raises ValidationException; floci returns a SequenceNumber
print(k.put_record(StreamName="probe-stream", Data=b"x" * (1024 * 1024 + 100), PartitionKey="pk"))
# PutRecords behaves the same: real AWS reports a per-record failure; floci assigns a SequenceNumber
r2 = k.put_records(StreamName="probe-stream", Records=[{"Data": b"x" * (1024 * 1024 + 100), "PartitionKey": "pk"}])
print(r2["FailedRecordCount"], r2["Records"][0])
EOF
```

**Environment**
- Floci version: `floci/floci:1.5.30-compat` (also reproduces on `1.5.28-compat`)
- Running via: Docker, `docker run -p 4566:4566 floci/floci:1.5.30-compat`
- Client: boto3 (aws-cli equivalent for setup)

soreavis:
## Summary

`PutRecord` and `PutRecords` accepted records of any size. Real Kinesis rejects a record whose data blob plus partition key exceeds the stream's maximum record size — 1 MiB by default, raisable per stream via `UpdateMaxRecordSize` since the [October 2025 large-record launch](https://aws.amazon.com/about-aws/whats-new/2025/10/amazon-kinesis-data-streams-10x-larger-record-sizes/). Floci doesn't implement `UpdateMaxRecordSize`, so every stream sits at the default and enforcing 1,048,576 bytes is the faithful behavior; per-stream configuration would be a separate enhancement.

The check lives in `putRecordWithShardId`, covering both operations. For `PutRecords`, an oversized record fails the whole request before anything is written: the [PutRecords reference](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecords.html) restricts per-record `ErrorCode` to `ProvisionedThroughputExceededException` or `InternalFailure`, so a size violation can't be a per-record failure — same pattern as the existing SQS `SendMessageBatch` → `BatchRequestTooLong` handling in this repo.

Error shape: `InvalidArgumentException` (400) — the Kinesis error list has no dedicated too-large error, `InvalidArgumentException` is its documented parameter-restriction error ("a specified parameter exceeds its restrictions"), and this handler already uses it for the `GetShardIterator` timestamp check. (`ValidationException`, which the issue title mentions, isn't documented for either put operation.)

The negative tests fail without the check; a boundary test pins exactly 1,048,576 bytes as accepted.

Closes #1755

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

Oversized `PutRecord` returns 400 `InvalidArgumentException`. `PutRecords` with any oversized entry fails top-level with nothing written — pinned by a test that reads the shard back and finds it empty. A record of exactly 1,048,576 bytes (data + partition key) is accepted. Malformed per-record `Data` keeps its existing per-record failure behavior.

`PutRecords` resolves the stream before validating, so a request against a missing stream now returns 400 `ResourceNotFoundException` where it previously returned 200 with a per-record `InternalFailure` — and an empty `Records` array against a nonexistent stream returned success. Same per-record `ErrorCode` restriction as above: a missing stream can't be reported per record.

Out of scope: the aggregate request limits — 500 records and 10 MiB per request including partition keys — stay unenforced, so a batch of many at-limit records is still accepted. Enforcing them would mirror `SqsService.validateBatchPayloadSize`; that's a separate change.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
@@ -383,8 +383,8 @@ public class KinesisService {
     }
 
     public PutRecordResult putRecordWithShardId(String streamName, byte[] data, String partitionKey, String region) {
-        validateRecordSize(data, partitionKey);
         KinesisStream stream = resolveStream(streamName, region);
+        validateRecordSize(data, partitionKey);
         KinesisShard shard = selectShard(stream, partitionKey);
 
         String sequenceNumber = String.valueOf(sequenceGenerator.incrementAndGet());
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
@@ -30,6 +30,10 @@ public class KinesisService {
     private static final String DEFAULT_STREAM_MODE = "PROVISIONED";
     private static final int DEFAULT_INSPECTION_RECORD_LIMIT = 100;
     private static final int MAX_INSPECTION_RECORD_LIMIT = 1000;
+    // AWS default per-stream maximum record size (data blob + partition key).
+    // Raisable per stream via UpdateMaxRecordSize, which floci doesn't
+    // implement, so every stream sits at the default.
+    private static final int MAX_RECORD_SIZE_BYTES = 1_048_576;
 
     private final StorageBackend<String, KinesisStream> store;
     private final StorageBackend<String, KinesisConsumer> consumerStore;
@@ -368,7 +372,18 @@ public class KinesisService {
         return putRecordWithShardId(streamName, data, partitionKey, region).sequenceNumber();
     }
 
+    public void validateRecordSize(byte[] data, String partitionKey) {
+        int size = (data != null ? data.length : 0)
+                + (partitionKey != null ? partitionKey.getBytes(StandardCharsets.UTF_8).length : 0);
+        if (size > MAX_RECORD_SIZE_BYTES) {
+            throw new AwsException("InvalidArgumentException",
+                    "Record size (data + partition key) of " + size + " bytes exceeds the maximum of "
+                            + MAX_RECORD_SIZE_BYTES + " bytes.", 400);
+        }
+    }
+
     public PutRecordResult putRecordWithShardId(String streamName, byte[] data, String partitionKey, String region) {
+        validateRecordSize(data, partitionKey);
         KinesisStream stream = resolveStream(streamName, region);
         KinesisShard shard = selectShard(stream, partitionKey);
 
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -427,6 +427,19 @@ public class KinesisJsonHandler {
     private Response handlePutRecords(JsonNode request, String region) {
         String streamName = resolveStreamName(request);
         JsonNode recordsNode = request.path("Records");
+
+        // An oversized record fails the whole request before anything is
+        // written — per-record ErrorCode is reserved for throughput/internal
+        // failures.
+        for (JsonNode node : recordsNode) {
+            try {
+                service.validateRecordSize(Base64.getDecoder().decode(node.path("Data").asText()),
+                        node.path("PartitionKey").asText());
+            } catch (IllegalArgumentException ignored) {
+                // malformed Data stays a per-record failure in the write loop below
+            }
+        }
+
         ObjectNode response = objectMapper.createObjectNode();
         ArrayNode results = response.putArray("Records");
         int failed = 0;
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -426,28 +426,37 @@ public class KinesisJsonHandler {
 
     private Response handlePutRecords(JsonNode request, String region) {
         String streamName = resolveStreamName(request);
+        service.describeStream(streamName, region);
         JsonNode recordsNode = request.path("Records");
 
         // An oversized record fails the whole request before anything is
         // written — per-record ErrorCode is reserved for throughput/internal
-        // failures.
+        // failures. Successful decodes are kept for the write loop; malformed
+        // Data is left null so it stays a per-record failure there.
+        record Entry(JsonNode node, byte[] data) {}
+        List<Entry> entries = new ArrayList<>();
         for (JsonNode node : recordsNode) {
+            byte[] data;
             try {
-                service.validateRecordSize(Base64.getDecoder().decode(node.path("Data").asText()),
-                        node.path("PartitionKey").asText());
-            } catch (IllegalArgumentException ignored) {
-                // malformed Data stays a per-record failure in the write loop below
+                data = Base64.getDecoder().decode(node.path("Data").asText());
+            } catch (IllegalArgumentException e) {
+                data = null;
             }
+            if (data != null) {
+                service.validateRecordSize(data, node.path("PartitionKey").asText());
+            }
+            entries.add(new Entry(node, data));
         }
 
         ObjectNode response = objectMapper.createObjectNode();
         ArrayNode results = response.putArray("Records");
         int failed = 0;
 
-        for (JsonNode node : recordsNode) {
+        for (Entry entry : entries) {
             try {
-                byte[] data = Base64.getDecoder().decode(node.path("Data").asText());
-                String partitionKey = node.path("PartitionKey").asText();
+                byte[] data = entry.data() != null ? entry.data()
+                        : Base64.getDecoder().decode(entry.node().path("Data").asText());
+                String partitionKey = entry.node().path("PartitionKey").asText();
                 KinesisService.PutRecordResult result = service.putRecordWithShardId(streamName, data, partitionKey, region);
                 results.addObject()
                         .put("SequenceNumber", result.sequenceNumber())
--- a/src/test/java/io/github/hectorvent/floci/services/kinesis/KinesisIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/kinesis/KinesisIntegrationTest.java
@@ -1030,6 +1030,29 @@ class KinesisIntegrationTest {
             .body("__type", equalTo("ResourceNotFoundException"));
     }
 
+    @Test
+    @Order(63)
+    void putRecordRejectsRecordOverSizeLimit() {
+        given()
+            .header("X-Amz-Target", "Kinesis_20131202.CreateStream")
+            .contentType(KINESIS_CONTENT_TYPE)
+            .body("{\"StreamName\": \"size-limit-test\", \"ShardCount\": 1}")
+        .when().post("/")
+        .then().statusCode(200);
+
+        String oversized = java.util.Base64.getEncoder().encodeToString(new byte[1_048_577]);
+        given()
+            .header("X-Amz-Target", "Kinesis_20131202.PutRecord")
+            .contentType(KINESIS_CONTENT_TYPE)
+            .body("{\"StreamName\": \"size-limit-test\","
+                + "\"Data\": \"" + oversized + "\","
+                + "\"PartitionKey\": \"pk1\"}")
+        .when().post("/")
+        .then()
+            .statusCode(400)
+            .body("__type", equalTo("InvalidArgumentException"));
+    }
+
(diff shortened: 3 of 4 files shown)


pgermosen (approved):
Looks good

Correct, well-reasoned fix with unusually careful attention to AWS semantics. `PutRecord`/`PutRecords` accepted records of any size; the PR now rejects a record whose `data + partition key` exceeds 1,048,576 bytes with `InvalidArgumentException` (400), and — importantly for `PutRecords` — does so atomically: an oversized entry fails the whole request before anything is written.

On scope: enforcing the 1 MiB **default** (rather than the raisable 10 MiB max) is the faithful choice given Floci doesn't implement `UpdateMaxRecordSize` — every stream sits at the default. The PR documents this clearly; per-stream max is a reasonable separate enhancement.

hampsterx:
I checked the current API reference before reading the diff, because the expectation in #1755 is stale in three ways, and all three land in your favour:

- `Data` length constraints are now `Minimum length of 0. Maximum length of 10485760`, so a 1,048,577-byte record no longer trips model validation and does reach the service.
- `InvalidArgumentException` is in the Errors list for both operations, described as "A specified parameter exceeds its restrictions". `ValidationException` is not listed for either.
- Per-record `ErrorCode` is explicitly restricted: "can be one of the following values: `ProvisionedThroughputExceededException` or `InternalFailure`".

So the error code you picked is the right one, and #1755's `ValidationException` expectation is out of date. The third point also means whole-request rejection is the correct semantic here and partial-failure-with-survivors would be wrong, since there's no per-record code for a size violation. You asked about the error code on the issue and didn't get a reply, so: your answer holds up.

The accounting is right too, and one test is doing more work than it looks like. `putRecordRejectsRecordOverSizeLimit` uses 1,048,574 data bytes, which is *under* the limit on its own, so it only rejects if the 3-byte partition key is counted. That pins the exact contract ("when the data blob is added to the partition key size") in a way that's easy to get wrong. Decoded-vs-encoded is right as well: 786,432 raw bytes base64-encode to exactly 1,048,576 chars, so an encoded-length check would have rejected everything over ~786 KB.

Three things I'd raise.

**1. The restructured per-record-failure branch has no test, anywhere in the repo.**

`KinesisJsonHandler.java:452-453`:

```java
byte[] data = entry.data() != null ? entry.data()
        : Base64.getDecoder().decode(entry.node().path("Data").asText());
```

The `null` arm exists solely to re-throw `IllegalArgumentException` inside the `try` so malformed `Data` stays a per-record `InternalFailure` rather than being promoted to a whole-batch failure. The PR body claims that guarantee ("Malformed per-record `Data` keeps its existing per-record failure behavior") and nothing verifies it, the only pre-existing `PutRecords` test asserts `FailedRecordCount == 0`, and no test in the repo produces a non-zero `FailedRecordCount` for Kinesis.

The risk is that the ternary reads as redundant (the pre-pass already decoded everything, right?), someone collapses it to `entry.data()`, and the guarantee silently changes with green tests either way. ~12 lines reusing the pattern from `putRecordsRejectsWholeBatchOnOversizedRecord`: three records, middle one with `Data: "!!!not-base64!!!"`, assert 200, `FailedRecordCount == 1`, records 0 and 2 carry `SequenceNumber`, record 1 carries `ErrorCode: "InternalFailure"`.

**2. `PutRecords` against a missing stream changes status code, undeclared.**

The new `service.describeStream(...)` at `KinesisJsonHandler.java:424` throws `ResourceNotFoundException`, so that case goes from `HTTP 200 {"FailedRecordCount": 1, "Records": [{"ErrorCode": "InternalFailure", ...}]}` to `HTTP 400`. That's the correct direction (`ResourceNotFoundException` is a documented top-level error and per-record `InternalFailure` is reserved for internal failures) but it isn't mentioned in the PR body or commit message and has no test. Anyone asserting the old shape breaks on upgrade with no changelog line explaining why. `updateStreamModeRejectsUnknownStream` in the same file is a five-line template.

**3. The aggregate request limit is silently unenforced.**

The reference says 500 records max and "up to a limit of 10 MiB for the entire request, including partition keys", so a 500 x 1 MiB batch is still accepted. The new pre-pass also decodes and retains every record in `List<Entry>` before writing anything, so peak heap for a batch that gets rejected is now the whole decoded batch where the old loop was streaming.

I'd leave the enforcement out (the effective total for a stream sitting at the 1 MiB record default is genuinely ambiguous, so the number is a maintainer call) but a line in the PR body scoping it out would help, since as written a reader concludes record-size limits are done.

One question rather than a request: the limit is hardcoded at `KinesisService.java:31`, while SQS exposes the same shape as config (`max-message-size` -> `FLOCI_SERVICES_SQS_MAX_MESSAGE_SIZE`, documented in `docs/services/sqs.md:89`). Your comment notes streams can be raised to 10 MiB via `UpdateMaxRecordSize` and that Floci doesn't implement it, which means someone emulating a raised stream now has no escape hatch where previously everything was accepted. Deliberate, or worth a `max-record-size` knob mirroring the SQS wiring?

soreavis:
The 10485760 `Data` max is the one I'd have gotten wrong on a re-read.

**Per-record-failure test added.** Collapsing the ternary to `entry.data()` costs more than the `InternalFailure` — the record lands. `validateRecordSize(null, "pk2")` counts the 3-byte partition key and passes, so a null blob stores, `FailedRecordCount` goes 1 to 0, and the next `GetRecords` NPEs on the encode. The other 98 Kinesis tests stayed green.

One asymmetry the test documents rather than endorses: #1926 made `PutRecord` return `SerializationException` for the same input. Aligning the batch path would turn a partial success into a whole-request rejection — separate change.

**Missing-stream case tested** off the `updateStreamModeRejectsUnknownStream` template, declared in the body and the commit message.

**Aggregate limit scoped out** in the body, with `SqsService.validateBatchPayloadSize` as the template if you want it enforced.

On the knob: fair point, there's no escape hatch now. But SQS only falls back to the env default under a per-queue `MaximumMessageSize` (`parseMaxMessageSize`, `SqsService.java:501`), so the honest mirror is `UpdateMaxRecordSize` per stream with a `max-record-size` default beneath it — its own PR. I can send it.

src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
```diff
@@ -359,8 +363,19 @@ public String putRecord(String streamName, byte[] data, String partitionKey, Str
         return putRecordWithShardId(streamName, data, partitionKey, region).sequenceNumber();
     }
 
+    public void validateRecordSize(byte[] data, String partitionKey) {
```
abanna, line 375: Including the partition key in the size calculation is the detail that's easy to miss — AWS's 1 MiB really is data + partition key, so that's right.

The other two `PutRecords` limits aren't enforced, though, and they're the same class of thing this PR is adding. AWS documents three:

- each record ≤ 1 MiB including its partition key ← this PR
- **at most 500 records per request**
- **5 MiB total per request, including partition keys**

Neither of the latter two appears anywhere in `KinesisService`. So a request with 10,000 records, or 500 × 1 MiB records totalling ~500 MiB, is accepted and written. The 5 MiB one is the more likely to bite, since a producer batching near the per-record ceiling hits it without any single record being oversized — and it fails on AWS while passing locally, which is the direction that costs debugging time.

The reasoning in your handler comment carries over exactly: these are whole-request rejections, not per-record `ErrorCode` entries. The `entries` loop you've added is already the natural place — accumulate the running total alongside the per-record check and reject before the write loop.

One thing worth double-checking against a real call: this throws `InvalidArgumentException`, but the AWS CLI reports an oversized record as `ValidationException` ("Value at 'data' failed to satisfy constraint: Member must have length less than or equal to 1048576"), since the blob length is a model-level constraint rather than a service check. Worth confirming which one a real `PutRecord` returns before this locks in — a client catching the documented exception type would miss it otherwise.

soreavis: Three numbers here, and the reference has moved on two. The request limit reads 10 MiB now, not 5 — the October large-record launch doubled it — and the per-record ceiling is 10 MiB, with 1 MiB the default a stream sits at until `UpdateMaxRecordSize` raises it. The 500-record cap is right, and it's unenforced.

Your mechanism on the error type is right too: it's a model-level constraint, not a service check. That model moved as well — `Data` is now `Maximum length of 10485760`, so 1,048,577 bytes gets past request validation. Going off the docs rather than a live call: `InvalidArgumentException` is in the Errors list for both put operations, `ValidationException` in neither.

On the aggregates, "fails on AWS, passes locally" is the right argument — eleven at-limit records still blow past 10 MiB. hampsterx argued for leaving them out, so rather than pick between you I scoped it in the body. hectorvent's call whether it folds in here or takes its own PR.

pgermosen: @hectorvent Should that be a follow-up?

hectorvent:
Thank you for this, and sorry it waited so long. Your reasoning holds up where it matters most: `InvalidArgumentException` is declared on both `PutRecord` and `PutRecords` while `ValidationException` is declared on neither. I enumerated all 39 operations in [botocore kinesis/2013-12-02](https://github.com/boto/botocore/blob/develop/botocore/data/kinesis/2013-12-02/service-2.json) to be sure. The sizing is right too (data plus partition key), and failing the whole batch matches `PutRecordsResultEntry.ErrorCode` allowing only throughput or internal failure.

I checked @abanna's three points and your reply is right on all of them: the request limit reads 10 MiB now rather than 5, `Data` allows 10485760, and `ValidationException` is on neither put operation. The 500 record cap is the one that stands.

**(blocking)** The branch conflicts with main now, in `KinesisService.java` only. It is cheap: nothing you added exists on main yet, and `putRecord` is unchanged apart from moving a few lines, so it is re-applying the constant, `validateRecordSize`, and the single call inside `putRecord`. `KinesisJsonHandler.java` does not conflict.

**(follow-up, separate PR)** @pgermosen asked, so to answer you both: land this once rebased, and take the 500 record and 10 MiB aggregate caps in their own PR, alongside the per stream limit defaulting to 1024 KiB.

That rebase is all this needs.

soreavis:
Rebased onto main. The conflict was the constants block only — #1880 landed the inspection limits where `MAX_RECORD_SIZE_BYTES` goes, so both are kept; `KinesisJsonHandler.java` auto-merged as you said it would. Kinesis suites plus the DynamoDB streaming forwarder, Pipes and the malformed-JSON tests are green on the new base.

Aggregates in their own PR, then: 500 records, 10 MiB per request, and the per-stream max defaulting to 1024 KiB. I'll pin the resolve-before-validate ordering there too — swapping those two lines currently fails nothing in the suite.

pgermosen (approved):
Verified the rebase Hector asked for rather than taking it on trust. `KinesisService.java`'s constants block genuinely keeps both `MAX_RECORD_SIZE_BYTES` (this PR) and `MAX_INSPECTION_RECORD_LIMIT` (from #1880) side by side, and `git merge-tree` against the latest `main` shows zero conflicts.

Ran the tests called out as green on the new base -- the full Kinesis package plus the DynamoDB streaming forwarder and Pipes tests (108 total) -- fresh myself: all pass.

The AWS-behavior reasoning already got unusually thorough scrutiny in this thread (data+partition-key sizing, whole-batch rejection matching `PutRecordsResultEntry.ErrorCode`'s restricted values, `InvalidArgumentException` over `ValidationException`), so nothing left to add there.

Approving.

PR #1897, opened 2026-07-16
feat(eventbridge): add Connection API actions

### LINKED ISSUE #1749: [FEAT] support EventBridge DescribeConnection
## Service

`EventBridge`

## API Action / Feature

`DescribeConnection`

## AWS Documentation

https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_DescribeConnection.html

## Why is this needed?

When attempting to port from LocalStack, the creation of an EventBridge rule via terraform is failing with the following:

```
Error: reading EventBridge Connection (test-http-ingress-connection): operation error EventBridge: DescribeConnection, https response error StatusCode: 400, RequestID: 51f06366-b80b-459d-8029-2edf536cdb74, api error UnsupportedOperation: Operation DescribeConnection is not supported.
```

I double checked the handler, and can confirm that this error is raised here:
https://github.com/floci-io/floci/blob/1.5.30/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeHandler.java#L84

- [ ] Yes
- [ ] No
- [x] Also no, but politely.

Agboolafeyikemi:
## Summary

Closes #1749

Implements the EventBridge Connection lifecycle: `CreateConnection`, `DescribeConnection`,
`UpdateConnection`, `DeleteConnection`, and `ListConnections`, mirroring the existing
Archive CRUD pattern in `EventBridgeHandler`/`EventBridgeService`.

Behavior notes:
- Connections are `AUTHORIZED` immediately (the emulator performs no real auth handshake).
- `DescribeConnection` strips credential values (`ApiKeyValue`, `Password`, `ClientSecret`)
  and masks invocation http parameter values flagged `IsValueSecret`, matching AWS.
- `SecretArn` is synthesized in AWS's `events!connection/<name>/<uuid>` format; no actual
  Secrets Manager secret is created (documented deviation).
- Out of scope: `DeauthorizeConnection`, `ListConnections` pagination (consistent with `ListArchives`).

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified end-to-end with AWS CLI 2.31.35 against the packaged emulator:
`events create-connection` / `describe-connection` / `list-connections` / `delete-connection`
all parse cleanly, with credential values stripped from describe responses.

## Checklist


ChrisLightfootWild:
Amazing, thanks @Agboolafeyikemi

@hectorvent can I help test this in any way?

ChrisLightfootWild:
I've built an image locally from this branch and confirmed that the Event Bus was created, and is now visible via:
 - `list-connections`
 - `describe-connection --name ...`

---

There's an additional snag in the flow I was using, but I think that could be handled separately.

The following are also currently unsupported:
 - CreateApiDestination
 - DeleteApiDestination
 - DescribeApiDestination
 - ListApiDestinations
 - UpdateApiDestination

Agboolafeyikemi:
Thanks so much for building and testing this locally @ChrisLightfootWild    really appreciate the verification! Agreed the ApiDestination actions are a separate scope; they build on Connections, so best handled in a follow-up issue/PR once this lands. Happy to pick that up next.

src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeHandler.java
```diff
@@ -627,6 +701,95 @@ private ObjectNode buildArchiveNode(Archive archive, boolean full) {
...
+        return authParameters;
+    }
+
+    private void maskSecretHttpParameters(JsonNode httpParameters) {
+        if (!httpParameters.isObject()) {
+            return;
+        }
+        for (String field : List.of("BodyParameters", "HeaderParameters", "QueryStringParameters")) {
+            for (JsonNode parameter : httpParameters.path(field)) {
+                if (parameter instanceof ObjectNode param && param.path("IsValueSecret").asBoolean(false)) {
```
abanna: `sanitizeAuthParameters` is thorough — covering `ApiKeyValue`, `Password`, `ClientSecret`, *and* recursing into `OAuthHttpParameters` as well as `InvocationHttpParameters` is more than most implementations bother with.

One default is inverted here, though. `IsValueSecret` on `ConnectionHeaderParameter` / `ConnectionQueryStringParameter` / `ConnectionBodyParameter` is optional in the AWS API, and its **default is `true`** — an omitted flag means "treat this as secret". This reads it as `false`:

```java
param.path("IsValueSecret").asBoolean(false)
```

So a parameter that simply doesn't set the flag is treated as non-secret and its value comes back in cleartext from `DescribeConnection`:

```json
{"InvocationHttpParameters": {"HeaderParameters": [
    {"Key": "X-Internal-Token", "Value": "hunter2"}
]}}
```

AWS masks that; Floci returns `hunter2`. Omitting the flag is the common case precisely because it's optional and the safe default is the documented one — Terraform's `aws_cloudwatch_event_connection` likewise leaves `is_value_secret` unset unless you say otherwise.

It's a one-character fix:

```java
param.path("IsValueSecret").asBoolean(true)
```

Worth a case in the connection tests with the flag omitted, since the existing coverage presumably sets it explicitly and so passes either way. This is also the one place where the divergence contradicts the promise in your own docs table — "credential values are stored but never returned".

pgermosen (approved):
All four prior findings are genuinely fixed, verified by reading the current code directly: ConnectionState has @RegisterForReflection, and maskSecretHttpParameters now correctly defaults IsValueSecret to true when omitted — a single shared helper covering HeaderParameters/QueryStringParameters/BodyParameters across both OAuthHttpParameters and InvocationHttpParameters, so the fix covers every call site, not just one. The Pattern import and brace-consistency nitpicks are also resolved.

Ran the targeted suite: EventBridgeConnectionIntegrationTest 11/11, EventBridgeSchedulerIntegrationTest 8/8, EventBridgeServiceTest 62/62 — 0 failures. Clean merge, docs-check passes, no CODEOWNERS-protected paths.

PR #1898, opened 2026-07-16
feat(cognito): add GetUserAttributeVerificationCode support

### LINKED ISSUE #1656: [FEAT] Cognito GetUserAttributeVerificationCode is unsupported
## Service

  Cognito

  ## AWS API Action

  `GetUserAttributeVerificationCode`

## Related issues

  Follow-up to:
  - #1488 `fix(cognito): align sign-up confirmation with AWS behavior`

  This issue is intentionally limited to `GetUserAttributeVerificationCode` support and behavior.
  
  Out of scope:
  - `ConfirmSignUp` verified-attribute updates
  - verification-code persistence bugs already tracked separately
  - delivery-priority parity between `email` and `phone_number`
  - stale code invalidation after `AdminConfirmSignUp`

  ## Expected behavior

  Floci should support `GetUserAttributeVerificationCode` for verifiable user attributes such as:
  - `email`
  - `phone_number`

  When the operation is valid, Floci should:
  - issue a verification code
  - dispatch it to the correct destination
  - return AWS-compatible `CodeDeliveryDetails`

  This is needed for post-sign-up attribute verification flows, for example when a signed-in user verifies `email` or `phone_number`.

  ## Actual behavior

  Floci currently returns an unsupported-operation error instead of handling the API action.

  Observed response:

  ```json
  {
    "__type": "UnsupportedOperation",
    "message": "Operation GetUserAttributeVerificationCode is not supported."
  }

  ## Reproduction

  ACCESS_TOKEN=<valid-access-token>

  aws --endpoint-url http://localhost:4566 cognito-idp get-user-attribute-verification-code \
    --access-token "$ACCESS_TOKEN" \
    --attribute-name email

  ## Actual response details

  Observed Floci result:

  - HTTP 400-style failure
  - UnsupportedOperation
  - no verification code is issued
  - no delivery occurs
  - no CodeDeliveryDetails response is returned

  ## Expected response details

  Expected AWS-compatible result:

  - operation is supported for verifiable attributes
  - a verification code is issued and delivered
  - response includes CodeDeliveryDetails
  - subsequent verification flow can complete successfully

  ## Why this matters

  Without this operation, Floci cannot emulate common Cognito flows where a signed-in user verifies or re-verifies attributes after sign-up.

  This blocks realistic testing for:

  - email verification after sign-up
  - phone verification after sign-up
  - onboarding flows that confirm attributes in multiple steps
  - frontend logic built on Amplify / Cognito SDK behavior

  ## Notes

  This issue is only about implementing and wiring GetUserAttributeVerificationCode.

  Please keep separate:

  - whether ConfirmSignUp sets email_verified / phone_number_verified
  - code-store persistence defects tracked elsewhere
  - delivery-target parity details
  - old-code invalidation after admin confirmation

  ## Environment

  - Floci version / image tag: current main / recent local image
  - How you're running Floci: Docker / local development
  - Observed date: 2026-06-30

  - [x] Yes
  - [ ] No

Agboolafeyikemi:
## Summary

Closes #1656

Adds `GetUserAttributeVerificationCode` support for post-sign-up attribute verification of `email` and `phone_number`. The implementation reuses the existing verification machinery: `VerificationCodeService` (with its previously-unused `ATTRIBUTE_VERIFICATION` purpose, 24h TTL per AWS) and `CognitoMessageDispatcher`, following the same pattern as the ForgotPassword flow. Access-token validation matches `GetUser` (revocation, origin-jti, and global sign-out checks).

Per the issue's intentionally limited scope, out of scope: `VerifyUserAttribute` (consume side), `ConfirmSignUp` verified-attribute updates, delivery-priority parity, and stale-code invalidation.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified with AWS CLI 2.31.35 against the packaged emulator: `cognito-idp get-user-attribute-verification-code --access-token ... --attribute-name email` returns `CodeDeliveryDetails` with a masked `Destination` (`e*******@e******.com`), `DeliveryMedium: EMAIL`; an unverifiable attribute returns `InvalidParameterException`, an invalid token returns `NotAuthorizedException`.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
@@ -79,6 +79,7 @@ public class CognitoJsonHandler {
             case "ForgotPassword" -> handleForgotPassword(request);
             case "ConfirmForgotPassword" -> handleConfirmForgotPassword(request);
             case "GetUser" -> handleGetUser(request);
+            case "GetUserAttributeVerificationCode" -> handleGetUserAttributeVerificationCode(request);
             case "UpdateUserAttributes" -> handleUpdateUserAttributes(request);
             case "DeleteUserAttributes" -> handleDeleteUserAttributes(request);
             case "GlobalSignOut" -> handleGlobalSignOut(request);
@@ -602,6 +603,19 @@ public class CognitoJsonHandler {
     private Response handleGetUser(JsonNode request) {
         Map<String, Object> result = service.getUser(request.path("AccessToken").asText());
         return Response.ok(objectMapper.valueToTree(result)).build();
+    }
+
+    private Response handleGetUserAttributeVerificationCode(JsonNode request) {
+        Map<String, Object> deliveryDetails = service.getUserAttributeVerificationCode(
+                request.path("AccessToken").asText(),
+                request.path("AttributeName").asText()
+        );
+        ObjectNode response = objectMapper.createObjectNode();
+        ObjectNode delivery = response.putObject("CodeDeliveryDetails");
+        delivery.put("AttributeName", String.valueOf(deliveryDetails.get("AttributeName")));
+        delivery.put("DeliveryMedium", String.valueOf(deliveryDetails.get("DeliveryMedium")));
+        delivery.put("Destination", String.valueOf(deliveryDetails.get("Destination")));
+        return Response.ok(response).build();
     }
 
     private Response handleUpdateUserAttributes(JsonNode request) {
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
@@ -612,9 +612,9 @@ public class CognitoJsonHandler {
         );
         ObjectNode response = objectMapper.createObjectNode();
         ObjectNode delivery = response.putObject("CodeDeliveryDetails");
-        delivery.put("AttributeName", String.valueOf(deliveryDetails.get("AttributeName")));
-        delivery.put("DeliveryMedium", String.valueOf(deliveryDetails.get("DeliveryMedium")));
-        delivery.put("Destination", String.valueOf(deliveryDetails.get("Destination")));
+        delivery.put("AttributeName", (String) deliveryDetails.get("AttributeName"));
+        delivery.put("DeliveryMedium", (String) deliveryDetails.get("DeliveryMedium"));
+        delivery.put("Destination", (String) deliveryDetails.get("Destination"));
         return Response.ok(response).build();
     }
 
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -72,6 +72,7 @@ Standalone `TagResource` rejects reserved `floci:*` keys. `ListTagsForResource` 
 | SignUp | Creates a self-service user for an app client. |
 | ConfirmSignUp | Confirms a pending self-service signup. |
 | GetUser | Returns attributes for the authenticated access-token user. |
+| GetUserAttributeVerificationCode | Issues a verification code for the authenticated user's email or phone_number attribute. |
 | UpdateUserAttributes | Updates attributes for the authenticated access-token user. |
 | ChangePassword | Changes the authenticated user's password. |
 | ForgotPassword | Starts the local forgot-password flow for a user. |
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -1420,12 +1420,14 @@ public class CognitoService {
         }
 
         String deliveryMedium = "email".equals(attributeName) ? "EMAIL" : "SMS";
+        VerificationCode.Purpose purpose = "email".equals(attributeName)
+                ? VerificationCode.Purpose.EMAIL_ATTRIBUTE_VERIFICATION
+                : VerificationCode.Purpose.PHONE_ATTRIBUTE_VERIFICATION;
         ensureVerificationWiring();
         try {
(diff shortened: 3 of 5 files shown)


dzwicker:
I compared `GetUserAttributeVerificationCode` with a real Cognito user pool while implementing the dependent consume side in #2018. A few concrete AWS-parity details differ from the current PR description/implementation:

- invalid access token: `NotAuthorizedException: Invalid Access Token`
- invalid attribute: `InvalidParameterException: Invalid attribute name. Only phone_number and email can be verified.`
- missing email: `User does not have a valid registered email address`
- missing phone: `User does not have a valid registered phone number`
- observed email mask: `z***@b***`
- observed phone mask: `+********8920`
- AWS permits issuing a fresh code even when the unchanged attribute is already verified

The dependent #2018 currently includes these adjustments plus regression coverage because it needs the same code path. They should either be incorporated here before merge or retained when #2018 is rebased after this PR merges.

Agboolafeyikemi:
Thanks for the detailed comparison, @dzwicker — I've incorporated all of these into #1898:

- Invalid access token → `NotAuthorizedException: Invalid Access Token`
- Invalid attribute → `InvalidParameterException: Invalid attribute name. Only phone_number and email can be verified.`
- Missing email/phone → `User does not have a valid registered email address` / `User does not have a valid registered phone number`
- Destination masking now matches what you observed: `z***@b***` for email, `+********8920`-style for phone (this also fixes masking on the shared ForgotPassword/account-recovery delivery path, since it reuses the same helpers)
- Confirmed the "fresh code for an already-verified attribute" case was already unblocked in the existing code (nothing checks the `*_verified` flag before issuing)  added a regression test to lock that in

This should mean #2018 won't need to reintroduce these fixes after rebasing let me know if I missed anything from your comparison.

src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
```diff
@@ -79,6 +79,7 @@ public Response handle(String action, JsonNode request, String region) {
             case "ForgotPassword" -> handleForgotPassword(request);
             case "ConfirmForgotPassword" -> handleConfirmForgotPassword(request);
             case "GetUser" -> handleGetUser(request);
+            case "GetUserAttributeVerificationCode" -> handleGetUserAttributeVerificationCode(request);
```
abanna, line 82: Heads up that this action is implemented twice across open PRs. #2018 (`feat(cognito): implement VerifyUserAttribute`, @dzwicker) adds the same `GetUserAttributeVerificationCode` dispatch case, the same handler method, and the same `getUserAttributeVerificationCode` service method — plus the same `EMAIL_ATTRIBUTE_VERIFICATION` / `PHONE_ATTRIBUTE_VERIFICATION` constants on `VerificationCode.Purpose`.

They conflict in `CognitoService.java`, and both create a test class at the same path:

```
src/test/java/io/github/hectorvent/floci/services/cognito/CognitoAttributeVerificationIntegrationTest.java
```

The good news is that the two agree on semantics — same `Purpose` constants, same delivery-details shape — so this is a "pick one" rather than a design disagreement.

The thing worth weighing: **`VerifyUserAttribute` isn't on `main`**, so merged on its own this PR issues a verification code that nothing can consume. A caller gets:

```
GetUserAttributeVerificationCode  → 200, code delivered
VerifyUserAttribute               → 400 UnsupportedOperation
```

#2018 implements both halves, and additionally changes consumed-code handling in `VerificationCodeService` (retaining a tombstone so a replayed code reports `EXPIRED` rather than `NOT_FOUND`). So it's close to a superset of this PR.

Not my call which lands — but whichever goes second will want to drop its duplicate implementation rather than resolve the conflict line by line, and it's worth the two of you agreeing before either is merged. Happy to re-review whichever way it goes.

(Separately, the shared-code-slot point in the bot review above still stands as far as I can tell — neither PR changes how `VerificationCodeService.issue` keys its store.)

Agboolafeyikemi:
Thanks for flagging the overlap — good catch that the diffs collide on the same file and test path.

On the "shared-code-slot still stands" point: I think that's resolved already. `VerificationCode.storageKey()` builds the key as `userPoolId:username:purpose.name()`, and `Purpose` now has separate `EMAIL_ATTRIBUTE_VERIFICATION` / `PHONE_ATTRIBUTE_VERIFICATION` values (not a shared `ATTRIBUTE_VERIFICATION`), so `issue`/`verify`/`invalidate` all key off the per-attribute purpose already — email and phone_number get independent slots. Happy to point at the exact lines if it'd help re-check.

On ordering with #2018: I don't think this needs new negotiation — your own description there already says "Depends on #1898" and that the draft is up early specifically to avoid duplicate implementation, so the plan was already for this to land first and #2018 to rebase down to just the `VerifyUserAttribute` half. The overlapping diff right now is just because #2018 is based on current `main` rather than on top of this branch.

@dzwicker — flagging so you can confirm that's still the plan on your end too.

hectorvent:
Thank you, and thanks @dzwicker for the real-Cognito comparison, which is a better source than anything I could have derived locally.

I did not re-check the error strings, masks or the reissue-when-already-verified behaviour, because that capture already settles them and you incorporated all of it.

What I did check is the overlap @abanna raised. #2018 is still open and adds the same `GetUserAttributeVerificationCode` dispatch alongside its own `VerifyUserAttribute`, and the two branches do conflict. The `CognitoService` conflict is a single region and a single word: this PR says `User does not have a valid registered email address` and #2018 says `...registered email`. Your string is the right one, by @dzwicker's own capture above, which records `address`.

So this should merge first, which is also what @dzwicker suggested, and #2018 can keep its other adjustments while taking this wording on the rebase.

greptile's shared code-slot P1 is genuinely closed too: `Purpose` now carries separate email and phone values and `storageKey()` includes the purpose name, so the two codes no longer share a slot.

No blockers from my side.

hectorvent (approved):

PR #1899, opened 2026-07-16
feat(firehose): implement S3 buffered delivery honoring BufferingHints

### LINKED ISSUE #1886: [FEAT] [Firehose] Implement S3 buffered delivery for PutRecord / PutRecordBatch
## Service

Kinesis Data Firehose

## API Action / Feature

`PutRecord` / `PutRecordBatch` — buffered delivery to an S3 destination

## AWS Documentation

- [PutRecord](https://docs.aws.amazon.com/firehose/latest/APIReference/API_PutRecord.html)
- [S3DestinationConfiguration](https://docs.aws.amazon.com/firehose/latest/APIReference/API_S3DestinationConfiguration.html)
- [BufferingHints](https://docs.aws.amazon.com/firehose/latest/APIReference/API_BufferingHints.html)

## Why is this needed?

We are writing integration tests for a logging backend that puts structured JSON records into a Firehose delivery stream backed by S3. On real AWS, `PutRecord` buffers the record and flushes it to the S3 destination within `BufferingHints.IntervalInSeconds` seconds 

On Floci 1.5.33, `CreateDeliveryStream` and `PutRecord` both succeed and return correct responses but no object ever appears in the destination S3 bucket, regardless of how long we wait. The delivery pipeline is effectively a black hole: records are accepted and then silently dropped.

This blocks any integration test that needs to assert on delivered content (record shape, key format, size guards, end-to-end logging round-trips).

## Steps to reproduce

```python
import boto3, time

ENDPOINT = "http://localhost:4566"   # adjust port as needed
REGION   = "us-east-1"
BUCKET   = "test-delivery-bucket"
STREAM   = "test-delivery-stream"
CREDS    = dict(
    aws_access_key_id="test",
    aws_secret_access_key="test",
    region_name=REGION,
    endpoint_url=ENDPOINT,
)

s3 = boto3.client("s3", **CREDS)
fh = boto3.client("firehose", **CREDS)

# 1. Create destination bucket

# 2. Create delivery stream with 1-second buffering
fh.create_delivery_stream(
    DeliveryStreamName=STREAM,
    DeliveryStreamType="DirectPut",
    S3DestinationConfiguration={
        "RoleARN": "arn:aws:iam::000000000000:role/firehose_delivery_role",
        "BucketARN": f"arn:aws:s3:::{BUCKET}",
        "BufferingHints": {"SizeInMBs": 1, "IntervalInSeconds": 1},
    },
)

# 3. Put a record
fh.put_record(
    DeliveryStreamName=STREAM,
    Record={"Data": b'{"hello":"world"}'},
)
print("Record put. Waiting 30 s for delivery...")

# 4. Poll S3 — objects never appear
for i in range(30):
    time.sleep(1)
    objects = s3.list_objects_v2(Bucket=BUCKET).get("Contents", [])
    if objects:
        print(f"Delivered after {i+1}s: {[o['Key'] for o in objects]}")
        break
else:
    print("No delivery after 30 s")  # ← always reached
```

**Expected:** at least one S3 object appears within a few seconds of `IntervalInSeconds=1`.

**Actual:** `No delivery after 30 s` — the bucket stays empty indefinitely.

## Container logs

After `PutRecord` the logs are completely silent. There is no flush attempt, no S3 write, and no error:

```
INFO  [FirehoseService] Created Firehose delivery stream: test-delivery-stream
INFO  [AwsJson11Controller] AwsJson11Controller firehose action: PutRecord
```

No `Flushed {0} records from stream {1} to s3://{2}/{3}` line ever appears, even though that log message is compiled into the binary (confirmed via `strings`).

For comparison, immediately after a Kinesis `PutRecord` the logs show `ListShards`, `GetShardIterator`, and `GetRecords` activity — the Kinesis pipeline is fully end-to-end. Firehose produces nothing.

## Binary analysis

Inspecting the native binary (`/app/application`) with `strings` shows:

- `FirehoseService$1` is present — an anonymous inner class, which in Quarkus/Java typically represents a scheduled `Runnable`.
- `Failed to flush Firehose stream {0}: {1}` is present — error-path logging for the flush.
- `Flushed {0} records from stream {1} to s3://{2}/{3}` is present — success-path logging for the flush.

However, `floci.storage.services.firehose.flush-interval-ms` is absent from the binary, while the equivalent key exists for every other stateful service (`dynamodb`, `kinesis`, `cloudwatchlogs`, `sns`, `ssm`, `secretsmanager`, etc.). This suggests the scheduled flush task exists in the code but is either not wired into the Quarkus scheduler at startup, or the per-stream timer started in `createDeliveryStream` does not start correctly.

A secondary hypothesis is that the internal S3 client used by the flush task resolves to `s3.amazonaws.com` instead of the local endpoint, causing the delivery to fail silently (no log line because the failure is swallowed before reaching the `Failed to flush` log path).

## Environment

| | |
|---|---|
| Floci version | `1.5.33` (image built `2026-07-15T05:59:57Z`, git `fba4d8f5`) |
| Image | `floci/floci:1.5.33` |
| testcontainers-floci (Python) | `4.14.2` |
| boto3 | `1.38.x` |
| Host OS | RHEL 9 / Docker via Unix socket |

## Additional context

[PR #1857](https://github.com/floci-io/floci/pull/1858) fixes the S3 object key format to match real AWS. The PR description references `FirehoseExtendedS3IntegrationTest` with delivery-plane tests, which implies delivery was expected to work at that point.

- [ ] Yes
- [x] No

marvikomo:
PutRecord/PutRecordBatch buffered records but only flushed at a hard-coded 5-record count, so records below that threshold were never delivered to the S3 destination. Replace the count trigger with AWS semantics: flush when SizeInMBs of data is buffered or when IntervalInSeconds elapses since the oldest record, whichever comes first (defaults 5 MiB / 300 s). A 1s background flusher drives the interval side, running each delivery under the stream's account so account-prefixed storage resolves correctly.

Also aligned with AWS:
- BufferingHints range validation (SizeInMBs 1-128, IntervalInSeconds 0-900) -> ValidationException
- duplicate CreateDeliveryStream -> ResourceInUseException
- record data treated as raw bytes end to end (no UTF-8 round trip)
- failed deliveries retried per interval, bounded at 10 attempts

Fixes #1886

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DataLakeTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DataLakeTest.java
@@ -5,10 +5,14 @@ import software.amazon.awssdk.core.SdkBytes;
 import software.amazon.awssdk.services.athena.AthenaClient;
 import software.amazon.awssdk.services.athena.model.*;
 import software.amazon.awssdk.services.firehose.FirehoseClient;
+import software.amazon.awssdk.services.firehose.model.BufferingHints;
 import software.amazon.awssdk.services.firehose.model.PutRecordRequest;
 import software.amazon.awssdk.services.firehose.model.Record;
+import software.amazon.awssdk.services.firehose.model.S3DestinationConfiguration;
 import software.amazon.awssdk.services.glue.GlueClient;
 import software.amazon.awssdk.services.glue.model.*;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.ListObjectsV2Request;
 
 import java.nio.charset.StandardCharsets;
 import java.util.List;
@@ -25,6 +29,7 @@ class DataLakeTest {
     private static AthenaClient athena;
     private static GlueClient glue;
     private static FirehoseClient firehose;
+    private static S3Client s3;
 
     private static final String DB_NAME = TestFixtures.uniqueName("test_db");
     private static final String TABLE_NAME = "orders";
@@ -35,6 +40,7 @@ class DataLakeTest {
         athena = TestFixtures.athenaClient();
         glue = TestFixtures.glueClient();
         firehose = TestFixtures.firehoseClient();
+        s3 = TestFixtures.s3Client();
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


hectorvent:
Thank you for this, and sorry it sat without a maintainer reply. Your design here is the right one: a size trigger on the put path, an interval trigger from a background flusher, and raw bytes end to end. That same design landed on main on 17 August through #2314, which closed #1886 and shipped in 1.7.0, so a rebase would now mean deleting almost all of this branch rather than merging it. I am closing this PR as superseded, with credit for getting there first.

One piece of yours is still missing on main: `validateBufferingHints` checks that `SizeInMBs` and `IntervalInSeconds` are given together but not their ranges (1 to 128 and 0 to 900 in the model, https://github.com/boto/botocore/blob/develop/botocore/data/firehose/2015-08-04/service-2.json). If you would like to send that as a small fresh PR on top of main, note the model has no `ValidationException`; `InvalidArgumentException` is the declared code.

PR #1901, opened 2026-07-17
fix(kinesis): resolve LATEST shard iterator position at GetShardIterator time

### LINKED ISSUE #1900: [BUG] Kinesis: LATEST shard iterator never returns records written after the iterator is obtained
## Description

`GetShardIterator` with `ShardIteratorType: LATEST` returns an iterator that never yields any records, even for records written *after* the iterator was obtained. Following `NextShardIterator` and polling does not help.

Per the AWS Kinesis API, `LATEST` starts reading "just after the most recent record in the shard", so a record written after the iterator is obtained must eventually be returned.

`TRIM_HORIZON` on the same shard at the same moment returns the record, so the write itself succeeds — only the `LATEST` iterator is affected.

## Impact

This breaks the standard "observe an event as it happens" pattern: obtain a `LATEST` iterator, trigger the action that produces the record, then poll. It is the pattern AWS documents for tailing a stream, and a common one in integration/E2E tests, so a suite that asserts on freshly produced records fails against Floci while passing on real AWS.

The only workaround we found is to re-read the whole shard with `TRIM_HORIZON` and filter client-side, which is not equivalent (it re-reads all retained records and cannot express "only what happens after this point").

## Reproduction

Self-contained; needs Docker and AWS CLI v2. Exits 0 either way and prints both the `LATEST` result and the `TRIM_HORIZON` control.

```bash
#!/usr/bin/env bash
set -euo pipefail

docker run -d --rm --name floci-repro -p 4566:4566 floci/floci:1.5.33 >/dev/null

until curl -sf "$AWS_ENDPOINT_URL/_localstack/health" >/dev/null; do sleep 1; done

aws kinesis create-stream --stream-name my-stream --shard-count 1
aws kinesis wait stream-exists --stream-name my-stream
SHARD=$(aws kinesis list-shards --stream-name my-stream --query 'Shards[0].ShardId' --output text)

# 1. obtain a LATEST iterator BEFORE any record is written
IT=$(aws kinesis get-shard-iterator --stream-name my-stream --shard-id "$SHARD" \
  --shard-iterator-type LATEST --query ShardIterator --output text)

# 2. write a record AFTER the iterator was obtained
aws kinesis put-record --stream-name my-stream --partition-key pk \
  --data "$(echo -n hello | base64)" >/dev/null

# 3. poll for 8s, following NextShardIterator (well within the 5 min iterator expiry)
N=0
for i in $(seq 1 16); do
  RES=$(aws kinesis get-records --shard-iterator "$IT" \
    --query "[length(Records), NextShardIterator]" --output text)
  N=$(echo "$RES" | cut -f1)
  IT=$(echo "$RES" | cut -f2)
  if [ "$N" != "0" ]; then
    echo "LATEST:       found $N record(s) after $(( (i - 1) * 500 ))ms"
    break
  fi
  sleep 0.5
done
if [ "$N" = "0" ]; then
  echo "LATEST:       0 records after 8s"
fi

# control: TRIM_HORIZON on the same shard, same moment
IT2=$(aws kinesis get-shard-iterator --stream-name my-stream --shard-id "$SHARD" \
  --shard-iterator-type TRIM_HORIZON --query ShardIterator --output text)
echo "TRIM_HORIZON: $(aws kinesis get-records --shard-iterator "$IT2" \
  --query 'length(Records)' --output text) record(s)"

docker rm -f floci-repro >/dev/null
```

Notes on the method, to rule out the usual false positives: the shard ID comes from `ListShards` rather than being hardcoded; the stream is awaited with `wait stream-exists`; empty `GetRecords` responses are treated as normal and the loop follows `NextShardIterator`; and the whole poll finishes in 8s, far inside the 5 minute iterator expiry.

## Expected

```
LATEST:       found 1 record(s) after 0ms
TRIM_HORIZON: 1 record(s)
```

## Actual

```
LATEST:       0 records after 8s
TRIM_HORIZON: 1 record(s)
```

## Affected versions

Reproduced on every version tested, so this does not look like a recent regression:

| Image | `LATEST` result |
|---|---|
| `floci/floci:1.5.29` | 0 records |
| `floci/floci:1.5.33` | 0 records |
| `floci/floci:nightly` | 0 records |

Running the identical script with the image swapped to `localstack/localstack:4.14.0` prints the expected output above (`found 1 record(s) after 0ms`).

## Environment

- Image: `floci/floci:1.5.33` (also reproduced on 1.5.29 and nightly)
- Host: Linux, Docker
- Client: AWS CLI v2

Agboolafeyikemi:
## Summary

Fixes #1900.

A `LATEST` shard iterator recomputed its start position as the shard tip on every `GetRecords` call instead of once at `GetShardIterator` time. Any record written between obtaining the iterator and the first poll landed below the recomputed tip and was skipped forever — breaking the standard tailing pattern (obtain a `LATEST` iterator, trigger the producer, poll).

## Changes

- `GetShardIterator` now snapshots the shard tip once when a `LATEST` iterator is created and encodes it in the iterator's existing index slot (same approach `DynamoDbStreamService` already uses). `GetRecords` resumes from that snapshot, so records appended after the iterator was obtained are returned. Applied to both the default and account-scoped paths.
- `GetShardIterator` with `LATEST` now fails fast with `ResourceNotFoundException` for an unknown shard, as on AWS, instead of deferring the failure to `GetRecords`.
- Documented shard iterator semantics in `docs/services/kinesis.md`.

## Tests

- New regression tests reproduce the issue and fail against the previous code: a record written after the iterator is obtained is returned on the first `GetRecords` call, and when polling via `NextShardIterator` (the loop from the issue's repro script).
- New test for the unknown-shard validation.
- Full `KinesisServiceTest` (32), `KinesisIntegrationTest` (35), `KinesisJsonHandlerTest` (24), `PipesPollerTest` (10), and `PipesPollerIntegrationTest` (39) suites pass.

### DIFF
--- a/docs/services/kinesis.md
+++ b/docs/services/kinesis.md
@@ -49,13 +49,17 @@ aws kinesis describe-stream --stream-name events --endpoint-url $AWS_ENDPOINT_UR
 aws kinesis describe-stream --stream-arn arn:aws:kinesis:us-east-1:000000000000:stream/events --endpoint-url $AWS_ENDPOINT_URL
 ```
 
+## Shard Iterators
+
+`GetShardIterator` supports all five iterator types: `TRIM_HORIZON`, `LATEST`, `AT_SEQUENCE_NUMBER`, `AFTER_SEQUENCE_NUMBER`, `AT_TIMESTAMP`.
+
+A `LATEST` iterator is positioned at the shard tip at the moment the iterator is created, matching AWS: records written after the iterator was obtained are returned, records written before are not. This supports the standard tailing pattern — obtain a `LATEST` iterator, trigger the action that produces the record, then poll `GetRecords` following `NextShardIterator`.
+
 ## Configuration
 
 | Variable | Default | Description |
 |---|---|---|
 | `FLOCI_SERVICES_KINESIS_ENABLED` | `true` | Enable or disable the service |
-
-## Examples
 
 ## Enhanced Fan-Out (EFO)
 
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
@@ -387,7 +387,7 @@ public class KinesisService {
         String raw = String.format("%s|%s|%s|%s|%d|%s",
                 streamName, shardId, type,
                 sequenceNumber != null ? sequenceNumber : "",
-                latestSnapshotIndex(stream, shardId, type),
+                iteratorStartIndex(stream, shardId, type),
                 timestampMillis != null ? timestampMillis.toString() : "");
         return Base64.getEncoder().encodeToString(raw.getBytes(StandardCharsets.UTF_8));
     }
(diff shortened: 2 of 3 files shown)


hectorvent:
Thank you for this, and sorry it has waited.

**(blocking)** The branch is showing conflicts with main and needs a rebase before it can be reviewed properly.

When you push that, a line in the description on which AWS behaviour you are matching, and where that came from, would help the review go quickly.

Blocked only on the rebase; a full review will follow once it is pushed.

hectorvent (approved):
Thanks @Agboolafeyikemi

PR #1903, opened 2026-07-17
feat(apigateway): derive SQS QueueUrl from path-style integration URI when absent from template

### LINKED ISSUE #1408: [FEAT] API Gateway → SQS: derive queue from path-style integration URI when QueueUrl is absent
## Service

API Gateway (REST/v1) → SQS AWS integration

## Context

Follow-up to #1384 / #1385 (which wired up path-style SQS integration URIs and query-protocol dispatch).

PR #1385 parses the `path/{account}/{queue}` components of the integration URI but then drops them — the queue is only resolvable when the VTL request template re-supplies `QueueUrl` in the rendered body. That works, and the #1385 test exercises it, but it isn't how real AWS routes the request.

## Expected behavior

In real AWS, the path **is** the routing: the integration URI `arn:aws:apigateway:{region}:sqs:path/{account}/{queue}` makes the request land on the SQS endpoint `POST /{account}/{queue}`, and the target queue comes from the path. So the common recipe template body is simply:

```
Action=SendMessage&MessageBody=$input.body
```

— with **no** `QueueUrl`. The queue is implied by the integration URI.

## Actual behavior

Floci's `invokeQuery` path relies on `QueueUrl` being present in the rendered body (it dispatches to `SqsQueryHandler`, which reads `QueueUrl` from the form params). A template that omits `QueueUrl` — the canonical AWS form — has no queue to act on.

## Proposed fix

When `QueueUrl` is absent from the rendered body, derive it from the `path/{account}/{queue}` components already parsed in `AwsServiceRouter.parseIntegrationUri` (currently discarded for the path form). This would make the no-`QueueUrl` recipe work as it does on AWS, while keeping the explicit-`QueueUrl` form working as it does today.

## Environment

- Floci version / image tag: `main` @ current HEAD
- Related: #1384, #1385

b6k-dev:
## Summary

Closes #1408

When an API Gateway integration URI uses the path-style form
`arn:aws:apigateway:us-east-1:sqs:path/{account}/{queue}` and the VTL
request template omits `QueueUrl`, the queue is now derived from the
integration URI path instead of being unresolvable.

Change summary:
- Parse and forward the path portion from path-style integration URIs
- Inject a derived `QueueUrl` when the rendered body lacks one
- Add integration tests for both path derived and explicit `QueueUrl`

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

As described in the issue:

> In real AWS, the path is the routing: the integration URI arn:aws:apigateway:{region}:sqs:path/{account}/{queue} makes the request land on the SQS endpoint POST /{account}/{queue}, and the target queue comes from the path. So the common recipe template body is simply:
> 
> Action=SendMessage&MessageBody=$input.body
> — with no QueueUrl. The queue is implied by the integration URI.
> 

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.services.apigateway;
 
+import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsArnUtils;
 import io.github.hectorvent.floci.core.common.AwsErrorResponse;
 import io.github.hectorvent.floci.core.common.AwsException;
@@ -41,6 +42,7 @@ import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -76,6 +78,7 @@ public class ApiGatewayExecuteController {
     private final WebSocketConnectionManager webSocketConnectionManager;
     private final ElbV2Service elbV2Service;
     private final SqsQueryHandler sqsQueryHandler;
+    private final EmulatorConfig emulatorConfig;
 
     @Inject
     public ApiGatewayExecuteController(ApiGatewayService apiGatewayService, ApiGatewayV2Service apiGatewayV2Service,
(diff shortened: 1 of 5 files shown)


hectorvent:
Thank you for this, and sorry it waited so long for a reply. The reasoning in the issue is right: with the `path/{account}/{queue}` integration URI the queue is the routing, so the template really does omit `QueueUrl`, and filling the unused `IntegrationTarget.path` component is a tidy fix that matches how `SqsService` builds and looks up queue URLs.

**(blocking)** The derived URL reuses the account segment from the URI verbatim. People copy the URI from real Terraform or CDK stacks, where it carries their actual account id, and `http://localhost:4566/123456789012/my-queue` will not resolve since Floci keys queues under the configured account. Passing just the queue name would let `SqsService.normalizeQueueUrl` prefix the configured account itself, and would also stop a malformed path producing a bogus URL and a confusing SQS-level error.

Heads up, not a change request: #1806 and #2011 construct this controller in new tests, so whichever lands second needs a small rebase.

One blocker above; otherwise this looks ready.

hectorvent:
Thank you, this is exactly the fix, and sorry for the slow reply. Taking just the last path segment means a URI carrying a real account id resolves now, and a bare queue name works because `SqsService` already reduces a `QueueUrl` to its last segment the same way.

The test refactor is the part I want to credit. Collapsing the ordered setup into helpers made the file shorter, but it also went from two real assertions to five, and `explicitQueueUrlPathQueueRemainsEmpty` is a good one to have thought of. It pins the fallback from the other side, so if someone later swaps `computeIfAbsent` for a plain put, that test goes red instead of the bug shipping quietly.

No blockers from my side.

hectorvent (approved):

PR #1906, opened 2026-07-18
fix(apigateway): resolve v2 API region across all regions for data-plane dispatch

### LINKED ISSUE #1902: [BUG] Unable to access floci apigatewayv2 url
Hi, 

I am trying to migrate from localstack to floci and I've an issue where I am not able to use the api gateway url in my frontend that talks to the backend apigateway. 

I run the terraform on the container and it returns be this url `https://e5da376f64.execute-api.eu-west-1.amazonaws.com`. I know this URL needs to be translated to some I've tried several things like

1. `http://e5da376f64.execute-api.eu-west-1.localhost:4566/events`, `http://e5da376f64.execute-api.eu-west-1.localhost.floci.io:4566/events`.  It gives me back:

```
<Error>
    <Code>InvalidArgument</Code>
    <RequestId>1e9978c2-0d30-45d6-b8dc-84772fc1d510</RequestId>
</Error>
```
2. `http://localhost:4566/execute-api/e5da376f64/events` and it gives me `Invalid API id specified`

My stage is $default. I've tried for so long I am sure I am doing something wrong here

tirthachetry-zoho:
Unsigned data-plane requests (browser/frontend) carry no SigV4 Authorization header, so the region resolves to the default (us-east-1). The v2 dispatch detection previously called getApi(region, apiId) with the unresolved region, so a v2 API created in another region (e.g. eu-west-1) was never detected and fell through to v1 handling, failing with
'Invalid API id specified' (404).

Resolve the region first via ApiGatewayV2Service.resolveApiRegion (which scans all stored API keys), then check isV2 using the resolved region. Add ApiGatewayV2Service.apiExists(region, apiId) to detect a v2 API that lives in the same/default region.

Fixes floci-io/floci#1902

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -232,17 +232,15 @@ public class ApiGatewayExecuteController {
                               String proxy, HttpHeaders headers, UriInfo uriInfo, byte[] body) {
         String region = regionResolver.resolveRegion(headers);
 
-        // Check if this is a v2 (HTTP API) or v1 (REST API)
-        boolean isV2 = false;
-        try {
-            apiGatewayV2Service.getApi(region, apiId);
-            isV2 = true;
-        } catch (AwsException ignored) {
-            // Not a v2 API — fall through to v1 handling
-        }
-
+        // Check if this is a v2 (HTTP API) or v1 (REST API). Unsigned data-plane
+        // requests (e.g. from a browser/frontend) carry no SigV4 Authorization header,
+        // so the region resolves to the default. Scan all regions for the API so a
+        // v2 API created in any region stays reachable — mirroring v1's
+        // resolveRestApiRegion behaviour.
+        String v2Region = apiGatewayV2Service.resolveApiRegion(region, apiId);
+        boolean isV2 = !v2Region.equals(region) || apiGatewayV2Service.apiExists(region, apiId);
         if (isV2) {
-            return dispatchV2(httpMethod, apiId, stageName, proxy, headers, uriInfo, body, region);
+            return dispatchV2(httpMethod, apiId, stageName, proxy, headers, uriInfo, body, v2Region);
         }
 
         // Resolve region for unsigned data-plane requests
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -1234,7 +1234,7 @@ public class ApiGatewayExecuteController {
 
         String requestId = UUID.randomUUID().toString();
         String eventJson = buildV2ProxyEvent(httpMethod, path, route.getRouteKey(),
-                apiId, stageName, headers, uriInfo, body, requestId);
+                apiId, region, stageName, headers, uriInfo, body, requestId);
 
         LOG.debugv("execute-api v2: {0} {1}/{2}{3} → Lambda {4}", httpMethod, apiId, stageName, path, functionName);
 
@@ -1795,7 +1795,7 @@ public class ApiGatewayExecuteController {
     }
 
     String buildV2ProxyEvent(String httpMethod, String path, String routeKey,
-                                     String apiId, String stageName,
+                                     String apiId, String region, String stageName,
                                      HttpHeaders headers, UriInfo uriInfo,
                                      byte[] body, String requestId) {
         ObjectNode event = objectMapper.createObjectNode();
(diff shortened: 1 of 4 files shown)


Millroy094 (approved):

src/test/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2CrossRegionDataPlaneReproTest.java
```diff
@@ -0,0 +1,156 @@
...
+     * the request reaches the v2 dispatch and is no longer rejected with that
+     * error. The integration targets example.com, which may be unreachable in
+     * the test sandbox (yielding a 502), but the key assertion is that the
+     * region-resolution error is gone.
+     */
+    @Test @Order(6)
+    void unsignedDataPlaneRequestResolvesApiAcrossRegions() {
+        given()
+                .contentType(ContentType.JSON)
+                .when().get("/execute-api/" + apiId + "/$default/events")
```
abanna, line 141: Thanks for adding a focused repro here. Could we make it use the URL from #1902—`/execute-api/{apiId}/events`—rather than adding `/$default/`? AWS serves a `$default` HTTP API stage from the base invoke URL, and I verified locally that this documented path still gets a 404 on this PR. As written, the test sidesteps the exact frontend URL it says it reproduces.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
```diff
@@ -232,17 +232,15 @@ Response dispatch(String httpMethod, String apiId, String stageName,
...
-        } catch (AwsException ignored) {
-            // Not a v2 API — fall through to v1 handling
-        }
-
+        // Check if this is a v2 (HTTP API) or v1 (REST API). Unsigned data-plane
+        // requests (e.g. from a browser/frontend) carry no SigV4 Authorization header,
+        // so the region resolves to the default. Scan all regions for the API so a
+        // v2 API created in any region stays reachable — mirroring v1's
+        // resolveRestApiRegion behaviour.
+        String v2Region = apiGatewayV2Service.resolveApiRegion(region, apiId);
```
abanna, line 240: Thanks for fixing the unsigned browser path. Could we keep an explicit SigV4 region scope authoritative before doing the all-region lookup? SigV4 is region-bound, but I verified a request scoped to `us-east-1` can reach an API that exists only in `eu-west-1` on this branch. A signed wrong-region regression would protect that boundary while keeping the browser fallback intact.

hectorvent:
Thank you for this, and sorry it waited so long for a reply. The diagnosis is exactly right: unsigned browser calls carry no SigV4 header, so the region fell back to the default and a v2 API created in `eu-west-1` was invisible to the dispatch check. Mirroring v1's `resolveRestApiRegion` is the right shape, and the `domainName` fix is a genuine parity improvement, since `{apiId}.execute-api.{region}.amazonaws.com` with the resolved region is what real HTTP APIs put in `requestContext.domainName`.

**(blocking)** The description's `AWS Compatibility` section still holds the template's placeholder comments, and "New or updated integration test added" is unticked even though you added `ApiGatewayV2CrossRegionDataPlaneReproTest`. One sentence on the broken behaviour plus the earned tick closes both.

Heads up, not a change request: #2011 and #1903 also change `buildV2ProxyEvent`'s parameter list, so whichever of the three lands second needs a small rebase.

Blocked only on the description touch-ups.

abanna (changes requested):
Thank you for addressing unsigned cross-region API Gateway dispatch and for adding a focused reproduction. All 10 targeted tests pass. Two compatibility issues remain in the current inline threads: a signed request scoped to the wrong region must not fall back to an all-region scan, and the regression should exercise the actual `$default` base path rather than inserting `/$default/`. AWS specifies that SigV4 credential scope restricts the signature to a region and service: [SigV4 credential scope](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html). The branch also conflicts with current `main`, so please rebase after addressing those cases. Thank you for working on this subtle browser-versus-signed-request distinction.

Karan-Tekuncorked:
Heads up — this branch is currently 532 commits behind `main` and conflicts with it directly in `ApiGatewayV2Service.java` and `ApiGatewayExecuteController.java`, independent of any other PR.

The reason: this PR's core contribution, `resolveApiRegion`, already exists on `main` — added 2026-08-12 by 33ba5ac5 (#2054, "route execute-api hostnames for HTTP APIs"). That same commit introduced `ApiGatewayExecuteApiHostFilter`, which replaced the dispatch path this PR patches (`ApiGatewayExecuteController`) with a different mechanism (`findApiOwner` scanning all accounts/regions unconditionally, falling back to `resolveApiRegion`). So the cross-region "Invalid API id specified" bug from #1902 looks to already be fixed on `main` via a different route than this PR takes.

I hit the same "falls through to S3" symptom from #1902 independently (same-region, GET instead of POST — see my repro in the issue), but traced it to an unrelated cause: `apigatewayv2 create-api --target` (quick create) never provisions an integration/route/stage, so `ApiGatewayExecuteApiHostFilter`'s stage lookup throws and the request falls through unrouted. Fix + regression test in #2877.

Given `resolveApiRegion` is already on `main`, this PR's remaining diff (`apiExists`, the `ApiGatewayExecuteController` changes) may no longer be needed — worth rebasing onto current `main` to see what's actually left before continuing review.

tirthachetry-zoho:
cool @Karan-Tekuncorked . So, we may close this PR.

hectorvent:
Thank you you all!
I will close it and let's check what is pending to fix on main based on the new changes that landed there lately. I appreciated all the hard work you did here.

PR #1909, opened 2026-07-19
fix(lambda,iam): assume execution role for in-function SDK calls

### LINKED ISSUE #1777: [BUG] Lambda execution role is not assumed; runtime always authenticates as root
## Description

When running Lambda functions in Floci with IAM enforcement enabled, the Lambda runtime always authenticates as the root user instead of assuming the configured execution role.

As a result, IAM policies are never evaluated for AWS SDK calls made from inside the Lambda because all requests originate from the root identity.

## Floci Version

```
floci/floci:1.5.27
```

### docker-compose.yml

```yaml
services:
  floci:
    image: floci/floci:1.5.27
    ports:
      - "4566:4566"
    environment:
      FLOCI_STORAGE_MODE: hybrid
      FLOCI_SERVICES_LAMBDA_HOT_RELOAD_ENABLED: "true"
      FLOCI_SERVICES_IAM_ENFORCEMENT_ENABLED: "true"
```

### serverless.yml

```yaml
provider:
  name: aws
  runtime: nodejs20.x

  iam:
    role: arn:aws:iam::000000000000:role/lambda-no-s3-role
```

(The same behavior also occurs when using the default generated execution role.)

## Steps to reproduce

1. Start Floci with IAM enforcement enabled.
2. Deploy a Lambda function using Serverless Framework.
3. Configure an execution role (or use the default generated one).
4. Inside the Lambda, call `sts:GetCallerIdentity`.
5. Also inspect the Lambda container.

## Lambda code

```ts
const sts = new STSClient({});

const identity = await sts.send(
  new GetCallerIdentityCommand({})
);

console.log(identity);
console.log(process.env.AWS_ACCESS_KEY_ID);
```

## Actual behavior

Inside the Lambda:

```
AWS_ACCESS_KEY_ID=test
AWS_SECRET_ACCESS_KEY=test
AWS_SESSION_TOKEN=test
```

`GetCallerIdentity` returns:

```
Account: 000000000000
Arn: arn:aws:iam::000000000000:root
```

Inspecting the Lambda container confirms these environment variables are injected:

```
AWS_ACCESS_KEY_ID=test
AWS_SECRET_ACCESS_KEY=test
AWS_SESSION_TOKEN=test
```

Because the runtime identity is root, AWS SDK calls succeed even when the execution role does not have permission (for example, `s3:ListBuckets`).

## Expected behavior

I would expect the Lambda runtime to authenticate using temporary credentials for the configured execution role (similar to AWS Lambda), so that:

- `GetCallerIdentity` returns an assumed-role ARN, for example:

```
arn:aws:sts::000000000000:assumed-role/lambda-no-s3-role/<session>
```

- AWS SDK calls are evaluated against the execution role's IAM permissions.
- IAM enforcement behaves the same as AWS.

## Question

Is this expected behavior (a current limitation), or should Floci assume the Lambda execution role and provide temporary credentials inside the Lambda runtime?

hampsterx:
## Summary

- Run Lambda SDK calls as known execution roles so IAM policies and caller identity match AWS behavior.
- Preserve unknown-role and mounted-credential fallbacks while leaving ECS behavior unchanged.
- Revoke container-bound sessions on cleanup and sweep orphaned sessions after restart.
- Add unit, enforcement, lifecycle, and real Docker coverage, plus updated documentation.

## Review

- Local committed-diff review: clean.
- Claude: no defects; four optional suggestions noted.
- GLM-5.2 agentic review: clean overall; low-risk cleanup suggestions noted.
- Applied: 0 fixes. No P1 or critical findings.
- Markdown lint reports pre-existing line-length and table-style issues in `docs/services/lambda.md`.

## Changes

- 20 files changed, 779 insertions, 67 deletions.
- Add Lambda execution-role credential minting and explicit-account session registration.
- Inject role credentials into launched containers while preserving mounted AWS configuration precedence.
- Revoke sessions during container cleanup and remove persisted crash leftovers at startup.
- Exempt `sts:GetCallerIdentity` from IAM policy denial, matching AWS behavior.

## Test plan

- [x] Focused Lambda, IAM, lifecycle, and ECS regression suite: 136 passed.
- [x] Real Docker Lambda execution-role integration test passed.
- [x] Full Maven suite executed: 7,867 tests, with two unrelated order-dependent EC2 pagination failures.
- [x] Isolated `Ec2IntegrationTest`: 143 passed, confirming the full-suite failures are shared-state contamination.
- [x] `git diff --check` passed.

Closes #1777

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/docker/LaunchedContainerAwsEnv.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/docker/LaunchedContainerAwsEnv.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.core.common.docker;
 
+import io.github.hectorvent.floci.services.iam.model.SessionCreds;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 
@@ -10,14 +11,13 @@ import java.util.Optional;
 
 /**
  * Builds the baseline AWS SDK environment for a container Floci launches, so the workload
- * inside it can reach the emulator with working credentials — the same variables regardless of
- * whether the container is a Lambda function or an ECS task.
+ * inside it can reach the emulator with working credentials. It uses the same variables
+ * regardless of whether the container is a Lambda function or an ECS task.
  *
  * <p>Emulator-style, matching how a local AWS emulator satisfies the SDK credential-provider
- * chain: the SDK is pointed at Floci via {@code AWS_ENDPOINT_URL} and given placeholder
- * credentials (Floci does not verify request signatures). This mirrors how AWS itself supplies
- * credentials to a launched workload — Lambda via its execution role, an ECS task via its task
- * role — so that in either case the container starts with usable credentials rather than an
+ * chain: the SDK is pointed at Floci via {@code AWS_ENDPOINT_URL} and given injected workload,
+ * host, or placeholder credentials. This mirrors how AWS itself supplies credentials to a
+ * launched workload, so the container starts with usable credentials rather than an
  * empty provider chain ({@code Could not load credentials from any providers}).
  */
 @ApplicationScoped
@@ -37,18 +37,32 @@ public class LaunchedContainerAwsEnv {
      * @param region            the AWS region the container should use
      * @param awsConfigMountDir  in-container directory of a mounted AWS config/credentials
      *                           directory (as in {@code ~/.aws}); when present the SDK discovers
-     *                           credentials from there. Empty = inject placeholder credentials.
+     *                           credentials from there. Empty = inject host or placeholder credentials.
      */
     public List<String> sdkBaselineEnv(String region, Optional<String> awsConfigMountDir) {
+        return sdkBaselineEnv(region, awsConfigMountDir, Optional.empty());
+    }
+
+    /**
+     * Builds the baseline environment with optional workload-specific session credentials.
+     * Mounted AWS configuration takes precedence over both injected and fallback credentials.
+     */
+    public List<String> sdkBaselineEnv(String region, Optional<String> awsConfigMountDir,
+                                       Optional<SessionCreds> injectedCredentials) {
         List<String> env = new ArrayList<>();
         env.add("AWS_DEFAULT_REGION=" + region);
         env.add("AWS_REGION=" + region);
         if (awsConfigMountDir.isPresent() && !awsConfigMountDir.get().isBlank()) {
-            // ~/.aws is mounted — don't inject credentials, let the SDK discover them.
+            // ~/.aws is mounted, so don't inject credentials. Let the SDK discover them.
             // Set explicit file paths so discovery works regardless of container HOME.
             String dir = awsConfigMountDir.get();
             env.add("AWS_SHARED_CREDENTIALS_FILE=" + dir + "/credentials");
             env.add("AWS_CONFIG_FILE=" + dir + "/config");
+        } else if (injectedCredentials.isPresent()) {
+            SessionCreds credentials = injectedCredentials.get();
+            env.add("AWS_ACCESS_KEY_ID=" + credentials.accessKeyId());
+            env.add("AWS_SECRET_ACCESS_KEY=" + credentials.secretAccessKey());
+            env.add("AWS_SESSION_TOKEN=" + credentials.sessionToken());
         } else {
             // Use Floci's own env vars, falling back to placeholder credentials.
             String ak = System.getenv("AWS_ACCESS_KEY_ID");
--- a/src/main/java/io/github/hectorvent/floci/core/common/docker/LaunchedContainerAwsEnv.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/docker/LaunchedContainerAwsEnv.java
@@ -3,11 +3,13 @@ package io.github.hectorvent.floci.core.common.docker;
 import io.github.hectorvent.floci.services.iam.model.SessionCreds;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import org.jboss.logging.Logger;
 
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Builds the baseline AWS SDK environment for a container Floci launches, so the workload
@@ -22,6 +24,11 @@ import java.util.Optional;
  */
 @ApplicationScoped
 public class LaunchedContainerAwsEnv {
+
+    private static final Logger LOG = Logger.getLogger(LaunchedContainerAwsEnv.class);
+
+    /** Host-credential passthrough is a property of the process, so it is worth saying once. */
+    private static final AtomicBoolean hostCredentialsWarned = new AtomicBoolean();
 
     private final ContainerReachableEndpoint reachableEndpoint;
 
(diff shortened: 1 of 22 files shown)


hampsterx:
Rebased onto `main` (1.5.34).

The earlier `sdk-test-java` red was the `AthenaTest.getQueryResults` cold-start flake rather than anything in this change; #1662 fixed that upstream, so it should stay green now.

Two conflicts came out of the rebase: the `EmulatorLifecycleTest` mock fields (kept `elbv2ServiceConfig` alongside the new `iamService`), and #1967's wider `IamEnforcementFilter` constructor, so the new `sts:GetCallerIdentity` bypass test now builds through `newFilter()` instead of an inline construction. Full suite green locally: 8076 tests, 0 failures, 9 skipped.

hampsterx:
Pushed one more fix into this branch.

A published version carries no `accountId` (`publishVersion` builds its snapshot field by field, filed separately as #2040), and the account is what an execution-role session gets registered and revoked under. So a version-qualified function got no session at all and quietly fell back to the placeholder credentials, skipping IAM enforcement while `$LATEST` enforced it. It only looked fine because the warm pool is keyed on function name, so `fn:1` was reusing `$LATEST`'s container and its credentials; #2028 would have exposed it.

`LambdaExecutionRoleCredentials.sessionAccountId(fn)` is now the one place the session account is resolved: the field when set, otherwise the account segment of the function ARN via the existing `AwsArnUtils.accountOrDefault`. Both launcher call sites that previously read the field directly (the `ContainerHandle` stamp and the launch-failure revoke) go through it too, so register and revoke can't disagree, which would otherwise have leaked a live non-expiring session rather than just skipping enforcement.

Five tests, three of which fail if `sessionAccountId` is reverted to a bare field read. One of them drives the real `publishVersion` rather than a hand-built snapshot, so a future regression in the version ARN construction fails there instead of silently sending version-qualified invokes back to the placeholder credentials.

Full suite green locally: 8081 tests, 0 failures, 9 skipped.

Note the ARN fallback is worth keeping even once #2040 is fixed: version snapshots persist, so ones written before that fix still come back with a null account after a restart.

src/main/java/io/github/hectorvent/floci/core/common/docker/LaunchedContainerAwsEnv.java
```diff
@@ -37,18 +37,32 @@ public LaunchedContainerAwsEnv(ContainerReachableEndpoint reachableEndpoint) {
...
             env.add("AWS_SHARED_CREDENTIALS_FILE=" + dir + "/credentials");
             env.add("AWS_CONFIG_FILE=" + dir + "/config");
+        } else if (injectedCredentials.isPresent()) {
+            SessionCreds credentials = injectedCredentials.get();
+            env.add("AWS_ACCESS_KEY_ID=" + credentials.accessKeyId());
+            env.add("AWS_SECRET_ACCESS_KEY=" + credentials.secretAccessKey());
+            env.add("AWS_SESSION_TOKEN=" + credentials.sessionToken());
         } else {
             // Use Floci's own env vars, falling back to placeholder credentials.
             String ak = System.getenv("AWS_ACCESS_KEY_ID");
```
abanna, line 75: The account-explicit session registration is careful work, and `sweepOrphanedLambdaExecutionRoleSessions` using `scanAllAccounts()` is exactly right for a startup hook — that's a distinction another PR I looked at today got wrong.

The fallback branch deserves a second look, though:

```java
} else {
    // Use Floci's own env vars, falling back to placeholder credentials.
    String ak = System.getenv("AWS_ACCESS_KEY_ID");
```

When a function's execution role isn't known to Floci and no `aws-config-path` is mounted, Floci's **own process credentials** are injected into the Lambda container. Lambda containers run arbitrary user code, so anyone who started Floci from a shell with real credentials exported — `aws sso login`, `aws-vault exec`, a CI runner, an `AWS_PROFILE` that populated the environment — hands those real credentials to every such function.

What makes it worth flagging is that it's opt-**out**: you get this by default and avoid it by setting `aws-config-path`. The docs present it as a feature ("you can pass static credentials to Floci's environment directly"), which is reasonable for the deliberate case, but the accidental case looks identical and is silent.

A couple of options that keep the feature without the footgun — gating it behind an explicit config flag rather than mere presence of the env vars, or logging a `WARN` naming the function when host credentials are forwarded, so it surfaces in the emulator log rather than only inside the container.

Smaller and separate: `LambdaExecutionRoleCredentials` mints the session key and secret with `ThreadLocalRandom`, which isn't a CSPRNG. These credentials are what `IamEnforcementFilter` authorizes against, so `SecureRandom` seems a better fit — the codebase already uses it elsewhere for exactly this kind of value.

hampsterx:
@abanna both addressed in cc149ea, on top of a rebase onto main (1.6.0).

**CSPRNG**: agreed, these values are what `IamEnforcementFilter` authorizes against. `LambdaExecutionRoleCredentials` now mints from a static `SecureRandom`, the same shape `CertificateGenerator` uses. That also needs an `--initialize-at-run-time` entry in `application.yml`, otherwise `NativeImageRuntimeInitializationTest` fails and the nightly native build captures the instance with a frozen seed, which is the `CloudMapService` regression from #1217. Registered.

Worth flagging separately: `StsQueryHandler.randomId`/`randomSecret` and `IamService.randomId`/`randomSecret` still use `ThreadLocalRandom`, and the credentials they mint (`AssumeRole`, `GetSessionToken`, `CreateAccessKey`) land in the same stores `IamEnforcementFilter` resolves against. Same defect class, outside this PR's scope. Happy to open an issue.

**Credential passthrough**: the fallback branch predates this PR (it is the existing `sdkBaselineEnv` behaviour), and this PR narrows it, since a function whose role Floci knows no longer reaches it at all. What remains is the unknown-role case, and you are right that the accidental version is indistinguishable from the deliberate one. Taken the WARN option: the fallback logs when any of the three host credential variables are actually set, carrying only the 4-character access-key prefix, and the docs note now says plainly what triggers passthrough.

One correction to the doc while I was in there: `aws sso login` on its own does *not* trigger this. It populates the CLI cache, not the environment, and neither does a bare `AWS_PROFILE`. What does is an exported key pair, `aws-vault exec`, or a CI runner. The note says so now, since "you might already be doing this" is only useful if it names the right thing.

Two judgement calls, both easy to change if you disagree:

- **Once per process, not per launch, and not naming the function.** `sdkBaselineEnv` is shared by the Lambda, ECS and MWAA launchers and takes no workload identifier, and hot-reload functions relaunch a container per invocation, so a per-launch warning would repeat until people scroll past it. Passthrough is a property of how the process was started, so it is said once.
- **No config flag.** Gating it would change credential resolution for the ECS and MWAA paths too, which is more than this PR should be doing. Happy to file that separately if you would rather have it gated than logged.

hampsterx:
Rebased onto `bc21367c`. Upstream moved 28 commits since the last push, so this went conflicted twice while it sat; both rounds were mechanical.

- `ContainerLauncher.launch` locals: #2208's `reservedCodeVolume` now sits alongside `containerId`, both hoisted for the same reason. Kept both.
- The `catch (RuntimeException e)` cleanup block: #2208 added `releaseCodeVolumeReference(reservedCodeVolume)` plus a terse `runtimeApiServerFactory.release`, which collided with the logged release this branch expanded that line into. Kept the code-volume release in its upstream position and the single logged release after it, so nothing releases twice.
- `ContainerLauncherTest` imports, against #2053's `LambdaFileSystemConfig`.
- `application.yml`: both `--initialize-at-run-time` entries kept.

No behavior change from the rebase itself. 183 tests across the launcher, IAM, enforcement and lifecycle classes pass locally.

hampsterx:
Rebased onto `5d61e952`. Upstream moved 24 commits since the last push; one conflict, plus two adaptations to changes that landed in the meantime.

- `ContainerLauncher.stop`: #2098 reordered shutdown to quiesce → SIGTERM/`stopAndRemove` → `close()` → release. Kept that ordering wholesale and dropped this branch's own `stopAndRemove` call, which the new flow supersedes. The execution-role `unregister` now sits in a `finally` nested under the release, so a release that throws cannot strand a registration whose credentials stay authorizable for the rest of the process lifetime.
- `ContainerLauncherCodeVolumeReuseTest` (new in #1827) constructs `ContainerLauncher` directly, so it takes the `LambdaExecutionRoleCredentials` argument this branch adds.
- `ContainerLauncherTest` stubbed `RuntimeApiServer.stop()`; the stop path now calls `close()`, so the two unregister tests NPE'd on an unstubbed future. Stubbed `close()` alongside it.

Focused suites green: `ContainerLauncherTest` (38), `ContainerLauncherCodeVolumeReuseTest` (9), `LambdaImageConfigTest` (17), `LambdaExecutionRoleCredentialsTest` (5), `LambdaExecutionRoleIamEnforcementIntegrationTest` (2).

hectorvent:
Thank you for this, and sorry it sat unanswered after your last push. Both follow-ups from the previous round landed properly.

The one I want to call out is the native image registration. A `static final SecureRandom` would otherwise be initialised at build time under GraalVM, baking a fixed seed into the binary so every deployment minted identical session tokens. Adding `--initialize-at-run-time` for that class is exactly right and is the sort of thing that silently survives every test.

The `sts:GetCallerIdentity` short circuit is correct too, and scoped to that one action: [botocore sts](https://github.com/boto/botocore/blob/develop/botocore/data/sts/2011-06-15/service-2.json) says no permissions are required and that the call succeeds even when a policy explicitly denies it.

The once per process warning when Floci forwards its own credentials into a container is a nice touch, particularly abbreviating the key rather than logging it.

No blockers from my side.

hectorvent (approved):

PR #1920, opened 2026-07-21
fix(rds): preserve aurora engine types in API responses

### LINKED ISSUE #1918: bug: RDS engine type conversion loses aurora-postgresql specification
## Problem

When specifying an RDS engine type as 'aurora-postgresql' or 'aurora-mysql', the Floci emulator was incorrectly converting these to their base engine types (POSTGRES/MYSQL) and returning the base type in API responses. This causes Terraform to detect state drift and attempt invalid updates.

## Example

1. User specifies `engine="aurora-postgresql"` when creating an RDS instance
2. Floci internally converts this to `DatabaseEngine.POSTGRES`
3. When querying the instance, the API returns `engine="postgres"` instead of `engine="aurora-postgresql"`
4. Terraform detects a mismatch and tries to modify the engine, which AWS rejects

## Root Cause

The `DatabaseEngine` enum only had base engine types (POSTGRES, MYSQL, MARIADB) with no distinction for Aurora variants. The `resolveEngine()` method was mapping both 'postgres' and 'aurora-postgresql' strings to the same `DatabaseEngine.POSTGRES` enum value.

dnlopes:
## Summary
Fixes #1918. Aurora engine identity (`aurora-postgresql`/`aurora-mysql`) was getting lost in a couple of places, causing Terraform state drift.

## What was already fixed on main
`DbInstance`/`DbCluster` already carry a raw `engineIdentifier` string (from the Aurora Serverless v2 work, #1829), and the top-level `<Engine>` element in `DescribeDBInstances`/`DescribeDBClusters` already prefers it over the enum. So the original repro in #1918 (API responses coming back as plain `postgres`) is already handled for that field.

## What was still missing
- `DBParameterGroupName`'s default naming still derived the engine prefix from the `DatabaseEngine` enum (only `POSTGRES`/`MYSQL`/`MARIADB`), so it produced `default.postgres16` instead of `default.aurora-postgresql16`.
- The managed master-user secret's `"engine"` JSON field had the same problem.
- The default parameter group name's version suffix always truncated to the major version only, which is right for the Postgres family but wrong for MySQL (`default.mysql8` instead of AWS's `default.mysql8.0`) — pre-existing, but newly visible once Aurora MySQL gets its own correctly-hyphenated prefix.
- The bare legacy `"aurora"` engine identifier silently mapped to `aurora-mysql`. Real AWS retired that identifier in Feb 2023 and rejects it with `InvalidParameterValue` for new instances/clusters, so it's rejected here too now instead of being silently rewritten.

## Changes
- `RdsQueryHandler.dbParameterGroupName`/`dbEngineMajorVersion`: prefer `engineIdentifier`, and use major.minor for the MySQL family's version suffix.
- `RdsService.managedMasterSecretString`: prefer `engineIdentifier` for the `"engine"` field.
- `RdsService.resolveEngine`: reject bare `"aurora"` instead of mapping it to `aurora-mysql`.
- Added test coverage for all of the above.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 /**
@@ -721,7 +722,7 @@ public class RdsQueryHandler {
         }
 
         String engine = instance.getEngine() != null
-                ? instance.getEngine().name().toLowerCase()
+                ? instance.getEngine().apiName()
                 : "unknown";
         return "default." + engine + dbEngineMajorVersion(instance);
     }
@@ -729,7 +730,7 @@ public class RdsQueryHandler {
     private static String dbEngineMajorVersion(DbInstance instance) {
         String engineVersion = instance.getEngineVersion();
         if ((engineVersion == null || engineVersion.isBlank()) && instance.getEngine() != null) {
-            engineVersion = defaultEngineVersion(instance.getEngine().name());
+            engineVersion = defaultEngineVersion(instance.getEngine().apiName());
         }
         if (engineVersion == null || engineVersion.isBlank()) {
             return "";
@@ -985,7 +986,7 @@ public class RdsQueryHandler {
         if (engine == null) {
             return "16.3";
         }
-        return switch (engine.toLowerCase()) {
+        return switch (engine.toLowerCase(Locale.ROOT)) {
             case "postgres", "aurora-postgresql" -> "16.3";
             case "mysql", "aurora-mysql", "aurora" -> "8.0.36";
             case "mariadb" -> "11.2";
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
@@ -630,13 +630,13 @@ public class RdsQueryHandler {
 
     private String dbInstanceInnerXml(DbInstance i) {
         DbEndpoint ep = i.getEndpoint();
-        String engineStr = i.getEngine() != null ? i.getEngine().name() : "";
+        String engineStr = i.getEngine() != null ? i.getEngine().apiName() : "";
         String statusStr = i.getStatus() != null ? statusLabel(i.getStatus()) : "available";
 
         XmlBuilder xml = new XmlBuilder()
                 .elem("DBInstanceIdentifier", i.getDbInstanceIdentifier())
                 .elem("DBInstanceStatus", statusStr)
-                .elem("Engine", engineStr.toLowerCase())
+                .elem("Engine", engineStr)
                 .elem("EngineVersion", i.getEngineVersion())
                 .elem("MasterUsername", i.getMasterUsername());
         if (i.getDbName() != null && !i.getDbName().isBlank()) {
@@ -760,13 +760,13 @@ public class RdsQueryHandler {
     private String dbClusterInnerXml(DbCluster c) {
         DbEndpoint ep = c.getEndpoint();
         DbEndpoint readerEp = c.getReaderEndpoint();
-        String engineStr = c.getEngine() != null ? c.getEngine().name() : "";
+        String engineStr = c.getEngine() != null ? c.getEngine().apiName() : "";
         String statusStr = c.getStatus() != null ? statusLabel(c.getStatus()) : "available";
 
         XmlBuilder xml = new XmlBuilder()
                 .elem("DBClusterIdentifier", c.getDbClusterIdentifier())
                 .elem("Status", statusStr)
-                .elem("Engine", engineStr.toLowerCase())
+                .elem("Engine", engineStr)
                 .elem("EngineVersion", c.getEngineVersion())
                 .elem("MasterUsername", c.getMasterUsername());
         if (c.getDatabaseName() != null && !c.getDatabaseName().isBlank()) {
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
@@ -1376,16 +1376,19 @@ public class RdsQueryHandler {
             return name;
         }
 
-        String engine = instance.getEngine() != null
-                ? instance.getEngine().name().toLowerCase()
-                : "unknown";
-        return "default." + engine + dbEngineMajorVersion(instance);
+        String engine = instanceEngine(instance);
+        return "default." + (engine.isEmpty() ? "unknown" : engine) + dbEngineMajorVersion(instance, engine);
     }
 
-    private static String dbEngineMajorVersion(DbInstance instance) {
+    /**
+     * AWS uses major.minor for the MySQL family's default parameter group name
+     * (e.g. {@code default.aurora-mysql8.0}), but only the major version for the
+     * Postgres family ({@code default.aurora-postgresql16}).
+     */
+    private static String dbEngineMajorVersion(DbInstance instance, String engine) {
         String engineVersion = instance.getEngineVersion();
         if ((engineVersion == null || engineVersion.isBlank()) && instance.getEngine() != null) {
-            engineVersion = defaultEngineVersion(instance.getEngine().name());
+            engineVersion = defaultEngineVersion(engine);
         }
         if (engineVersion == null || engineVersion.isBlank()) {
             return "";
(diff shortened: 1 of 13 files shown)


src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
```diff
@@ -980,18 +981,21 @@ private DatabaseEngine resolveEngine(String engineParam) {
...
         }
-        return switch (engineParam.toLowerCase()) {
-            case "postgres", "aurora-postgresql" -> DatabaseEngine.POSTGRES;
-            case "mysql", "aurora-mysql", "aurora" -> DatabaseEngine.MYSQL;
+        return switch (engineParam.toLowerCase(Locale.ROOT)) {
+            case "postgres" -> DatabaseEngine.POSTGRES;
+            case "aurora-postgresql" -> DatabaseEngine.AURORA_POSTGRESQL;
+            case "mysql" -> DatabaseEngine.MYSQL;
+            case "aurora-mysql" -> DatabaseEngine.AURORA_MYSQL;
+            case "aurora" -> DatabaseEngine.AURORA_MYSQL;
```
pgermosen, line 989: **`aurora` still round-trips as something the user didn't send.** *(minor / question)*

Everything else in this PR now preserves the caller's engine string exactly. This one case still rewrites it: a request with `engine="aurora"` comes back as `Engine: aurora-mysql`. That is the same drift #1918 describes, just for a different input — Terraform would see `aurora` in config, `aurora-mysql` in state, and try to correct it.

Two coherent options, no strong preference:

- **Preserve it** — add an `AURORA` constant whose `apiName()` is `aurora`, mapping to the MySQL image and protocol like the others.
- **Reject it** — real AWS treats `aurora` as the Aurora MySQL 5.6 identifier, which reached end of life in February 2023; the API no longer accepts it for new clusters and returns `InvalidParameterValue`. Removing the case so it falls through to the existing `default ->` branch would match current AWS behavior most closely.

Silently mapping it to a different engine is the one option that leaves the original bug in place for that input.

Small note on the description: it calls `aurora` the "AWS default". It was never a default — it was the legacy Aurora MySQL 5.6 identifier. Worth correcting so the reasoning isn't misread later.

src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
```diff
@@ -721,15 +722,15 @@ private static String dbParameterGroupName(DbInstance instance) {
...
+                ? instance.getEngine().apiName()
                 : "unknown";
         return "default." + engine + dbEngineMajorVersion(instance);
     }
 
     private static String dbEngineMajorVersion(DbInstance instance) {
         String engineVersion = instance.getEngineVersion();
         if ((engineVersion == null || engineVersion.isBlank()) && instance.getEngine() != null) {
-            engineVersion = defaultEngineVersion(instance.getEngine().name());
+            engineVersion = defaultEngineVersion(instance.getEngine().apiName());
```
pgermosen, line 733: **`default.aurora-mysql8` should be `default.aurora-mysql8.0`.** *(minor)*

`dbEngineMajorVersion` keeps only the leading digits of the version, so `8.0.36` → `8` and `16.3` → `16`. Correct for the PostgreSQL family, where AWS's major version genuinely is just `16`. Wrong for the MySQL family, where AWS uses major**.**minor:

| Engine | Floci emits | Real AWS |
|---|---|---|
| `postgres` 16.3 | `default.postgres16` | `default.postgres16` ✅ |
| `aurora-postgresql` 16.3 | `default.aurora-postgresql16` | `default.aurora-postgresql16` ✅ |
| `mysql` 8.0.36 | `default.mysql8` | `default.mysql8.0` ❌ |
| `aurora-mysql` 8.0.36 | `default.aurora-mysql8` | `default.aurora-mysql8.0` ❌ |
| `mariadb` 11.2 | `default.mariadb11` | `default.mariadb11.2` ❌ |

The truncation is **pre-existing**, not introduced here. But this PR newly routes Aurora instances through an `aurora-mysql` prefix, so the wrong value is newly reachable for Aurora users — and the two new tests only cover `aurora-postgresql`, where truncation happens to produce the right answer. That asymmetry is the part that worries me: the suite now looks like it validates parameter-group naming when it only validates the case that works.

The mirror test would surface it immediately:

```java
    @Test
    void describeDbInstances_defaultParameterGroupNameUsesMajorMinorForAuroraMysql() {
        DbInstance instance = makeInstance("aurora-my");
        instance.setEngine(io.github.hectorvent.floci.services.rds.model.DatabaseEngine.AURORA_MYSQL);
        instance.setEngineVersion("8.0.36");
        when(service.listDbInstances(null)).thenReturn(List.of(instance));

        Response response = handler.handle("DescribeDBInstances", params());

        String body = (String) response.getEntity();
        assertTrue(body.contains("<DBParameterGroupName>default.aurora-mysql8.0</DBParameterGroupName>"),
                "AWS uses major.minor for the MySQL family, got: " + body);
    }
```

src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
```diff
@@ -985,7 +986,7 @@ private static String defaultEngineVersion(String engine) {
         if (engine == null) {
             return "16.3";
         }
-        return switch (engine.toLowerCase()) {
+        return switch (engine.toLowerCase(Locale.ROOT)) {
             case "postgres", "aurora-postgresql" -> "16.3";
             case "mysql", "aurora-mysql", "aurora" -> "8.0.36";
```
pgermosen, line 991: **Aurora MySQL `EngineVersion` doesn't look like AWS's.** *(non-blocking / follow-up)*

Real Aurora MySQL 3.x reports its version as `8.0.mysql_aurora.3.05.2`, not a plain MySQL version — that's what `DescribeDBClusters` returns and what Terraform's `engine_version` compares against. Aurora PostgreSQL does report a plain `16.3`, so the `aurora-postgresql` branch above is already right.

Entirely **pre-existing** and out of scope here. Flagging because it's the same bug class as #1918 — an Aurora user who pins `engine_version` will hit drift again on the next field — so it's the natural next issue to file once this lands.

src/main/java/io/github/hectorvent/floci/services/rds/model/DatabaseEngine.java
```diff
@@ -2,14 +2,20 @@
...
-            case POSTGRES -> 5432;
-            case MYSQL, MARIADB -> 3306;
+            case POSTGRES, AURORA_POSTGRESQL -> 5432;
+            case MYSQL, AURORA_MYSQL, MARIADB -> 3306;
         };
     }
+
+    public String apiName() {
+        return this.name().toLowerCase(Locale.ROOT).replace("_", "-");
+    }
```
pgermosen, line 20: **Note for the release, not a code change.**

Instances and clusters created before this change with `engine="aurora-postgresql"` were persisted as `POSTGRES`, so after upgrading they'll keep reporting `postgres` until recreated.

Nothing breaks and no migration is needed — Jackson's name-based enum serialization means the old values still deserialize cleanly, which is also why inserting the new constants mid-enum is safe. But a line in the PR description or release notes would save someone filing a duplicate of #1918 against stale state.

pgermosen (approved):
Looks good.

Well-executed fix the diagnosis in the description is accurate and the enum change is the right shape.

Test work is careful too: switching restorePersistedRuntimeRestoresClusterAndMemberInstance to postgres and adding a dedicated Aurora restore test  rather than overloading a test whose subject is restore mechanics  was the right call.

@dnlopes  There are a few minor comments in here, when you can take a look

hectorvent:
Thank you for this, and I am sorry it sat so long without a word from us. The review was written on 30 July and simply never posted, which is on me.

The fix still holds up. Main's `DatabaseEngine` is still `POSTGRES, MYSQL, MARIADB`, so Aurora genuinely cannot be represented today, and `apiName()` produces exactly the strings AWS uses: `CreateDBCluster`'s `Engine` documentation lists `aurora-mysql` and `aurora-postgresql` ([botocore rds](https://github.com/boto/botocore/blob/develop/botocore/data/rds/2014-10-31/service-2.json)). Mapping bare `aurora` to the MySQL edition is right too.

**(blocking)** The branch conflicts with main now, and that is our doing: #1813 and #2035 both landed in RDS this week. It is four regions across three files, and `DatabaseEngine.java` itself is untouched. The only production one is the image switch, where #2035 added a version aware fallback; keeping its arms and widening the labels to `case POSTGRES, AURORA_POSTGRESQL ->` and `case MYSQL, AURORA_MYSQL ->` resolves it.

Rebase and this is good to go.

pgermosen (approved):
Rebased this onto current main and narrowed the fix down quite a bit. Turns out main already round-trips the engine string correctly in the Describe responses via engineIdentifier (from the Aurora Serverless v2 work) — what was still leaking the base engine name was the default parameter group naming and the managed secret's engine field, so I fixed those two spots directly instead of introducing new enum values. Also fixed the version-suffix truncation for the MySQL family (was always major-only, needs major.minor) and made the legacy bare "aurora" identifier get rejected instead of silently becoming aurora-mysql, matching current AWS behavior. CI's green across the board.

PR #1923, opened 2026-07-22
fix(dynamodb): honor customer-supplied KMS key for table SSE

### LINKED ISSUE #1922: [BUG] DynamoDB CreateTable/UpdateTable ignore customer-supplied KMS key for SSE
## Service

DynamoDB

## AWS API Action

CreateTable, UpdateTable, DescribeTable

## Expected behavior

When creating a table with `SSESpecification.Enabled=true`, `SSEType=KMS`, and a customer-supplied `KMSMasterKeyId`, both the `CreateTable` response and subsequent `DescribeTable` calls should return that customer-managed key in `SSEDescription.KMSMasterKeyArn`.

`UpdateTable` should also be able to enable SSE, rotate the KMS key, or disable SSE on an existing table, per the real DynamoDB API.

## Actual behavior

`CreateTable` parses `SSESpecification.Enabled` but never reads `KMSMasterKeyId` from the request at all — it unconditionally stores/returns a synthetic default key ARN (`arn:aws:kms:<region>:000000000000:key/aws-managed-dynamodb`), discarding whatever key the client specified. It also hardcodes `SSEType` to `"KMS"`, ignoring a client-supplied `AES256`.

`UpdateTable` has no `SSESpecification` handling at all, so there is no way to enable, rotate, or disable a table's encryption key after creation.

Root cause: `DynamoDbJsonHandler.handleCreateTable` (`src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbJsonHandler.java`):

```java
JsonNode sseSpec = request.path("SSESpecification");
if (!sseSpec.isMissingNode() && sseSpec.path("Enabled").asBoolean(false)) {
    table.setSseEnabled(true);
    table.setSseType(SSE_TYPE_KMS);
    table.setKmsMasterKeyArn(defaultKmsMasterKeyArn(region));
}
```

## Reproduction

```bash
  --table-name MyTable \
  --sse-specification Enabled=true,SSEType=KMS,KMSMasterKeyId=arn:aws:kms:us-east-1:000000000000:key/my-custom-key

aws --endpoint-url http://localhost:4566 dynamodb describe-table --table-name MyTable \
  --query 'Table.SSEDescription'
# Expected: KMSMasterKeyArn == arn:aws:kms:us-east-1:000000000000:key/my-custom-key
# Actual:   KMSMasterKeyArn == arn:aws:kms:us-east-1:000000000000:key/aws-managed-dynamodb
```

## Environment

- Floci version / image tag: 1.5.33
- Java SDK version (if applicable): N/A (reproduced via AWS CLI)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): `mvn quarkus:dev`

dnlopes:
Fixes #1922

## Summary
- `CreateTable` discarded `SSESpecification.KMSMasterKeyId` entirely and always stored/returned the default AWS-managed key ARN, so `CreateTable`/`DescribeTable` never reflected a customer-managed KMS key. It also hardcoded `SSEType` to `KMS`, ignoring `AES256`.
- `UpdateTable` had no `SSESpecification` handling at all, so encryption could not be enabled, rotated to a new key, or disabled after table creation.
- Fixed both: `CreateTable` now stores the client-supplied `KMSMasterKeyId`/`SSEType`, falling back to the default key only when omitted; `UpdateTable` now supports enabling SSE (with a custom or default key), rotating the KMS key, and disabling SSE, following the same in-place mutation pattern already used for `StreamSpecification` in the same method.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Test plan
- [x] `createTableWithCustomKmsKeyRetainsItOnDescribe`: verifies a custom KMS key ARN passed on `CreateTable` round-trips through both the `CreateTable` response and a subsequent `DescribeTable`.
- [x] `updateTableCanEnableRotateAndDisableSse`: verifies `UpdateTable` can enable SSE with a custom key, rotate to a different key, and disable SSE, with `DescribeTable` reflecting each state.
- [x] Full existing `DynamoDbIntegrationTest` suite passes (58 tests).
- [x] Full project test suite (`./mvnw test`) passes.

## Checklist


hectorvent:
Thanks for this, and I am sorry it sat so long without a reply from us. The core shape is right: `KMSMasterKeyId` in, `KMSMasterKeyArn` out, and `AES256`/`KMS` as the `SSEType` enum all match [botocore dynamodb/2012-08-10](https://github.com/boto/botocore/blob/develop/botocore/data/dynamodb/2012-08-10/service-2.json).

**(blocking)** The branch conflicts with main now. #1811 (global table v2 replicas) landed and added a block at the same anchor in `handleUpdateTable`. It is one hunk: keep main's replica block first, then yours, then your `persistTable(...)` last.

**(blocking)** `validateSseType()` runs after `createTable(...)` has already committed the table and enabled the stream, so an invalid `SSEType` returns ValidationException but leaves a real table behind. Real DynamoDB rejects before creating anything. Validating ahead of `createTable` fixes both paths.

**(follow-up, separate PR)** Worth widening the description: the `persistTable(...)` you added also fixes reload durability for tags, billing mode and streams, not only SSE. Genuinely good catch.

Two of the bot comments above are no longer accurate at this head, so please do not spend time on them. The "SSE changes bypass persistence" P1 is addressed by the `persistTable(...)` calls you added, and your `sseSettingsSurvivePersistentStorageReload` test pins it. The "SSEType is taken verbatim with no enum validation" comment is answered by `VALID_SSE_TYPES` and `validateSseType`. The one bot point that does hold is the validation ordering above.

A rebase and that validation move are all this needs.

pgermosen:
@hectorvent both blocking items from your review are addressed — rebased cleanly onto main and the `SSEType` validation now runs before `createTable`/`updateTable` in both handlers, mirroring the `validateReplicaUpdates`-before-`updateTable` pattern already in `handleUpdateTable`, so an invalid type no longer leaves a table (or stream) behind.

One thing worth a second look: I didn't resolve the `handleUpdateTable` conflict the way you suggested ("replica block first, then SSE, then persistTable"). Reading `applyReplicaUpdates`, it re-fetches the table from `tableStore` and does its own persist — and the comment already on main's `persistTable` call says a persistent backend doesn't hold the live object. Running the replica block first would mean that re-fetch sees neither the stream nor the SSE mutations, silently dropping both once `table` gets reassigned to its result. I kept main's original ordering instead — `persistTable` unconditionally, then the conditional `applyReplicaUpdates` — with the SSE block slotted in right before the flush, same shape `CreateTable` already uses. Flagging this since it goes against what you asked for; happy to be corrected if I'm missing something about how the replica path is meant to interact with the rest.

Also found and resolved a second, smaller conflict in `handleCreateTable`: main had independently added its own `persistTable(...)` for the same stream-durability reason this PR added one for SSE, so those collapsed into one call. Added `DescribeTable` assertions to the two existing invalid-`SSEType` tests confirming no side effects now that validation runs first. The AES256 coverage you didn't mention was already there from the branch's second commit, so no gap there.

hectorvent:
Sorry for the slow reply here. pgermosen, your ordering is the right one and mine was wrong for persistent backends: `applyReplicaUpdates` fetches the table through `validateReplicaUpdatesAndGetTable` and writes it back with `tableStore.put`, and a persistent backend serialises on write rather than holding the live object, so running the replica block first would re-fetch a copy without the stream and SSE mutations. Thank you for reading it that carefully and for carrying the branch.

Both blockers from my August review are resolved at this head: validation runs ahead of `createTable` and `updateTable`, and the two `persistTable` calls collapsed into one. I merged main into the branch locally and ran the DynamoDB suites on the result, 614 green. Main has since appended two tests at the same spot in `DynamoDbJsonHandlerTest`, a keep-both conflict; I will resolve that at merge time.

Merging. Thank you, dnlopes, for the fix.

hectorvent:
One correction to my note above: rather than resolving the conflict at merge time on our side, could one of you push the conflict fix to the branch? It is a single keep-both hunk in `DynamoDbJsonHandlerTest.java`: keep `sseSettingsSurvivePersistentStorageReload` from this branch and `putGetDeleteResourcePolicyRoundTrips` from main, both sides' imports, and the closing brace of the first method at the seam. That exact merge compiled and passed the DynamoDB suites locally, 614 green. Merge follows as soon as it lands.

PR #1035, opened 2026-05-25
fix(ec2) Fix/1031 ec2 describe network interfaces

nblomquist:
## Summary

Implements EC2 `DescribeNetworkInterfaces` support to unblock Terraform destroy flows for VPC networking resources.

Closes #1031

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore

## AWS Compatibility

Before this change, `DescribeNetworkInterfaces` returned `UnsupportedOperation`, which caused Terraform AWS provider destroy operations to fail when deleting subnets, security groups, and VPCs. Terraform calls `DescribeNetworkInterfaces` as a safety check before deleting networking resources.

This PR adds `DescribeNetworkInterfaces` support with:

- empty `networkInterfaceSet` response when no ENIs exist
- ENI responses derived from launched EC2 instance network interfaces
- common filters including subnet, VPC, security group, status, private IP, and tags
- pagination with `MaxResults` / `NextToken`
- AWS-compatible validation for `MaxResults`, including `InvalidMaxResults`
- `InvalidParameterCombination` when `NetworkInterfaceId` is combined with `MaxResults`

Verified with:

- Java EC2 integration tests
- AWS SDK Java pagination compatibility test
- Terraform compatibility test covering VPC/subnet/security group create + destroy

## Checklist

- [X] `./mvnw test` passes locally (for related tests, there are other non-related failures)
- [x] New or updated integration test added
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

PR #1937, opened 2026-07-24
feat(ec2): complete VPC Query lifecycle support

jvanzyl:
## Summary

- Complete the EC2 Query handlers and response shapes for VPC attributes,
  CIDR associations, endpoint discovery, and related network discovery calls.
- Add AWS SDK coverage for the supported VPC lifecycle and document the
  implemented actions.
- Remove a deleted VPC's AWS-created default security group and rules, main
  route table, and default network ACL without cascading into unrelated VPCs.
- Preserve provider compatibility for empty filters, tag-only security group
  specifications, and the expected public-hostname response shape.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The EC2 Query endpoint now supports the VPC lifecycle and discovery calls used
by ordinary AWS SDK clients, including VPC attribute round trips, secondary
CIDR associations, VPC endpoint records, and empty discovery responses for
supported resource types that have no local lifecycle model.

Creating a VPC also creates AWS-owned default network resources. Deleting that
VPC now removes its default security group and persisted rules, main route
table, and default network ACL. The cleanup is scoped by VPC identity and does
not remove defaults belonging to another VPC.

## Checklist

- [x] Documentation updated

## Verification

Current rebased tree (`dacd4cad7`):

- `./mvnw -B -Dtest='Ec2IntegrationTest,Ec2QueryHandlerTest,Ec2ServicePersistenceTest,Ec2ServiceTest' test`
  - 195 tests, 0 failures, 0 errors, 0 skipped
- Terraform compatibility suite
  - 22 tests, 0 failures
- OpenTofu compatibility suite
  - 16 tests, 0 failures
- `git diff --check upstream/main...HEAD`

Additional patch-set proof before the final upstream-only rebase:

- `./mvnw -B -f compatibility-tests/sdk-test-java/pom.xml -DskipTests package`
  - 121 AWS SDK test sources compiled with target 17
- `./mvnw -B test`
  - 7,973 tests, 0 failures, 0 errors, 8 skipped
  - log SHA-256:
    `de0b9fc2a70fffa8d95d6eb2fcf32ceec41aa2862850b22fdccc2d47db00a610`

The compatibility project was compiled on the available JDK 25 host using its
configured Java 17 source/target settings; this is not a JDK 17 runtime claim.

## Commit Stack

1. `704439163` `feat(ec2): complete VPC Query compatibility`
2. `603be4ce9` `fix(ec2): clean up VPC default resources`
3. `dacd4cad7` `fix(ec2): preserve VPC provider compatibility`

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2Tests.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2Tests.java
@@ -3,6 +3,7 @@ package com.floci.test;
 import org.junit.jupiter.api.*;
 import software.amazon.awssdk.services.ec2.Ec2Client;
 import software.amazon.awssdk.services.ec2.model.*;
+import software.amazon.awssdk.services.ec2.model.Tag;
 
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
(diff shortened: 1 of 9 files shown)


hectorvent:
Thank you for this careful work, and sorry it sat so long without a human reply. The new wire surface lines up with the model everywhere I could check: request parameter names (including the tricky `TransitGatewayAttachmentIds`), response set names, MaxResults bounds, and skipping pagination validation on `DescribeVpnGateways` ([botocore ec2/2016-11-15](https://github.com/boto/botocore/blob/develop/botocore/data/ec2/2016-11-15/service-2.json)).

**(blocking)** Two error codes I could not confirm anywhere: `InvalidEgressOnlyInternetGatewayId.NotFound` (casing differs from its `ID.NotFound` siblings) and `InvalidMaxResults`, where EC2 usually answers `InvalidParameterValue`. The other three are pinned by the EC2 waiters and [terraform errors.go](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/errors.go). A real-AWS capture settles it; the tests cannot, since they assert what the handler emits.

**(blocking)** The ~200 line CreateSubnet tag parser duplicates `parseTagsForResource`; normalising the null-value bug inside the shared helper fixes all six call sites.

**(follow-up, separate PR)** The always-empty `describeXIds` helpers are void validators; naming them so would help. #2029 touches the same `Ec2Service.java` line.

Blocked on the two points above.

hectorvent:
Thank you for your patience, and I owe you a correction before anything else.

I asked you to justify `InvalidMaxResults`. That ask was wrong: `Ec2QueryHandler` on main already throws exactly that code with the same message, and has since #1035, so your change is following the existing convention rather than inventing anything. Please consider that half withdrawn.

**(blocking)** `InvalidEgressOnlyInternetGatewayId.NotFound` still needs settling, though my reasoning about its casing was poor. AWS is inconsistent there, and main carries both forms, so casing proves nothing. The narrower point is that the code appears in none of botocore, moto or the Terraform provider, and unlike the other one it is new here. A real capture would close it.

**(blocking)** The CreateSubnet tag parser still duplicates `parseTagsForResource`; fixing the null-value handling in the shared helper covers all six call sites.

**(blocking)** The branch also conflicts with main now, in the two EC2 files, which have seen a lot of traffic since July.

hectorvent:
Closing the loop here, with an apology. The one blocker I left standing, `InvalidEgressOnlyInternetGatewayId.NotFound`, was wrong: the AWS EC2 API error reference lists that code with exactly this casing (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html), next to `InvalidInternetGatewayID.NotFound` and `InvalidVpcEndpointId.NotFound`, so AWS itself mixes the two spellings and yours is the documented one. Withdrawn, and thank you for citing the page on #2450.

Since #2450 is your current-upstream extraction of this branch with the tag parser deduplicated, I am closing this PR in its favour and picking the review up there. Sorry for the two weeks of silence after you did what was asked.

PR #1939, opened 2026-07-24
fix(cloudformation): map FIFO and redrive attributes for AWS::SQS::Queue

### LINKED ISSUE #1907: [BUG] CloudFormation SQS provider drops FIFO DeduplicationScope, FifoThroughputLimit, and RedrivePolicy on create
## Service

CloudFormation, SQS

## AWS API Action

`cloudformation create-stack` / `execute-change-set` (via `cdk deploy`) provisioning an
`AWS::SQS::Queue`. The gap is in the CloudFormation SQS resource provider's attribute mapping when
it calls `CreateQueue` on behalf of the stack.

## Expected behavior

A CloudFormation stack that creates a FIFO `AWS::SQS::Queue` with the following properties should
produce a queue whose `GetQueueAttributes` reports every one of them:

```yaml
FifoQueue: true
ContentBasedDeduplication: false
DeduplicationScope: messageGroup
FifoThroughputLimit: perMessageGroupId
VisibilityTimeout: 30
RedrivePolicy:
  deadLetterTargetArn: <dlq-arn>
  maxReceiveCount: 5
```

That is, `GetQueueAttributes --attribute-names All` should include `DeduplicationScope`,
`FifoThroughputLimit`, and `RedrivePolicy`, consistent with what real AWS returns.

## Actual behavior

The stack reaches `CREATE_COMPLETE`, but `GetQueueAttributes` on the created queue is **missing
three attributes**:

| Attribute                   | In synthesized template? | Present after CFN create? |
| --------------------------- | ------------------------ | ------------------------- |
| `FifoQueue`                 | yes                      | **yes**                   |
| `ContentBasedDeduplication` | yes                      | **yes**                   |
| `VisibilityTimeout`         | yes                      | **yes**                   |
| `DeduplicationScope`        | yes                      | **no (dropped)**          |
| `FifoThroughputLimit`       | yes                      | **no (dropped)**          |
| `RedrivePolicy`             | yes                      | **no (dropped)**          |

So `FifoQueue`, `ContentBasedDeduplication`, and `VisibilityTimeout` survive the CloudFormation
path, but `DeduplicationScope`, `FifoThroughputLimit`, and `RedrivePolicy` are silently discarded.

Two controls isolate the defect to the CloudFormation provider's attribute mapping — not the SQS
engine and not the synthesizer:

- **Control 1 — the synthesized template contains all three.** The template the deploy actually
  submits carries `DeduplicationScope`, `FifoThroughputLimit`, and `RedrivePolicy` (snippet below),
  so nothing is lost before CloudFormation.
- **Control 2 — the direct SQS API persists all three.** Creating an equivalent queue with
  `aws sqs create-queue --attributes {...}` against the same running instance persists and returns
  all three attributes verbatim. The SQS `CreateQueue`/`GetQueueAttributes` engine handles them
  correctly; only the CloudFormation-to-SQS mapping drops them.

### Control 1 — synthesized template snippet

The `cdk synth` output for the queue (this is what gets submitted to CloudFormation):

```json
{
  "Type": "AWS::SQS::Queue",
  "Properties": {
    "ContentBasedDeduplication": false,
    "DeduplicationScope": "messageGroup",
    "FifoQueue": true,
    "FifoThroughputLimit": "perMessageGroupId",
    "RedrivePolicy": {
      "deadLetterTargetArn": { "Fn::GetAtt": ["ExampleDlq<hash>", "Arn"] },
      "maxReceiveCount": 5
    },
    "VisibilityTimeout": 30
  }
}
```

## Reproduction

Self-contained minimal working example: one stack, two queues (a FIFO queue plus its FIFO
dead-letter queue). No image or code assets, so the stack needs nothing beyond SQS.

### 1. Start a fresh instance (image tag pinned)

```bash
docker rm -f floci-mwe 2>/dev/null || true

docker run --name floci-mwe \
  -p 4566:4566 \
  -e AWS_ACCESS_KEY_ID=test \
  -e FLOCI_STORAGE_MODE=memory \
  -d floci/floci:1.5.32

until curl -sf http://localhost:4566/_floci/health >/dev/null; do sleep 1; done
```

### 2. Minimal CDK app

`package.json`:

```json
{
  "name": "floci-sqs-fifo-mwe",
  "version": "0.0.0",
  "private": true,
  "bin": { "app": "bin/app.js" },
  "dependencies": {
    "aws-cdk-lib": "2.261.0",
    "constructs": "^10.0.0"
  },
  "devDependencies": {
    "aws-cdk": "^2",
    "aws-cdk-local": "3.0.4"
  }
}
```

`cdk.json`:

```json
{ "app": "node bin/app.js" }
```

`bin/app.js`:

```js
const cdk = require('aws-cdk-lib');
const sqs = require('aws-cdk-lib/aws-sqs');

class DemoStack extends cdk.Stack {
  constructor(scope, id, props) {
    super(scope, id, props);

    // FIFO dead-letter queue (a FIFO queue's DLQ must itself be FIFO)
    const dlq = new sqs.Queue(this, 'ExampleDlq', {
      fifo: true,
      contentBasedDeduplication: true,
    });

    // Main FIFO queue carrying the attributes under test
    new sqs.Queue(this, 'ExampleQueue', {
      queueName: 'example-queue.fifo',
      fifo: true,
      contentBasedDeduplication: false,
      deduplicationScope: sqs.DeduplicationScope.MESSAGE_GROUP,
      fifoThroughputLimit: sqs.FifoThroughputLimit.PER_MESSAGE_GROUP_ID,
      visibilityTimeout: cdk.Duration.seconds(30),
      deadLetterQueue: { maxReceiveCount: 5, queue: dlq },
    });
  }
}

const app = new cdk.App();
new DemoStack(app, 'demo-stack', {
  env: { account: '000000000000', region: 'us-east-1' },
});
```

### 3. Deploy

```bash
npm install

# Optional: confirm Control 1 (all three attributes are in the template)
npx cdklocal synth

# Bootstrap then deploy. Bootstrap logs one CREATE_FAILED for AWS::ECR::Repository
# but still reports the environment as bootstrapped (see "Secondary observation" below),
# and this asset-free stack does not need ECR, so deploy proceeds normally.
npx cdklocal bootstrap
npx cdklocal deploy --require-approval never
```

### 4. Verify via GetQueueAttributes (the bug)

```bash
QURL=$(aws --endpoint-url http://localhost:4566 sqs get-queue-url \
  --queue-name example-queue.fifo --query QueueUrl --output text)

aws --endpoint-url http://localhost:4566 sqs get-queue-attributes \
  --queue-url "$QURL" --attribute-names All \
  --query 'Attributes.{FifoQueue:FifoQueue,ContentBasedDeduplication:ContentBasedDeduplication,DeduplicationScope:DeduplicationScope,FifoThroughputLimit:FifoThroughputLimit,VisibilityTimeout:VisibilityTimeout,RedrivePolicy:RedrivePolicy}'
```

**Expected (real AWS):**

```json
{
  "FifoQueue": "true",
  "ContentBasedDeduplication": "false",
  "DeduplicationScope": "messageGroup",
  "FifoThroughputLimit": "perMessageGroupId",
  "VisibilityTimeout": "30",
  "RedrivePolicy": "{\"deadLetterTargetArn\":\"arn:aws:sqs:us-east-1:000000000000:demo-stack-ExampleDlq<hash>.fifo\",\"maxReceiveCount\":5}"
}
```

**Actual (Floci 1.5.32):**

```json
{
  "FifoQueue": "true",
  "ContentBasedDeduplication": "false",
  "DeduplicationScope": null,
  "FifoThroughputLimit": null,
  "VisibilityTimeout": "30",
  "RedrivePolicy": null
}
```

### 5. Control 2 — direct SQS API persists all three (proves the engine is fine)

```bash
# DLQ for the direct-API queue
DLQ_URL=$(aws --endpoint-url http://localhost:4566 sqs create-queue \
  --queue-name direct-dlq.fifo \
  --attributes FifoQueue=true,ContentBasedDeduplication=true \
DLQ_ARN=$(aws --endpoint-url http://localhost:4566 sqs get-queue-attributes \
  --queue-url "$DLQ_URL" --attribute-names QueueArn \
  --query 'Attributes.QueueArn' --output text)

# Same attribute set, created straight through the SQS API
DIRECT_URL=$(aws --endpoint-url http://localhost:4566 sqs create-queue \
  --queue-name direct-example.fifo \
  --attributes "{
    \"FifoQueue\":\"true\",
    \"ContentBasedDeduplication\":\"false\",
    \"DeduplicationScope\":\"messageGroup\",
    \"FifoThroughputLimit\":\"perMessageGroupId\",
    \"VisibilityTimeout\":\"30\",
    \"RedrivePolicy\":\"{\\\"deadLetterTargetArn\\\":\\\"${DLQ_ARN}\\\",\\\"maxReceiveCount\\\":5}\"
  }" \

aws --endpoint-url http://localhost:4566 sqs get-queue-attributes \
  --queue-url "$DIRECT_URL" --attribute-names All \
  --query 'Attributes.{DeduplicationScope:DeduplicationScope,FifoThroughputLimit:FifoThroughputLimit,RedrivePolicy:RedrivePolicy}'
```

This returns `DeduplicationScope`, `FifoThroughputLimit`, and `RedrivePolicy` all populated —
confirming the SQS engine stores and returns them, and that only the CloudFormation resource
provider drops them.

## Environment

- Floci version / image tag: `floci/floci:1.5.32`
- Java SDK version (if applicable): N/A (reproduced via CDK + AWS CLI, not the Java SDK)
- CDK: `aws-cdk-lib` 2.261.0, `aws-cdk-local` 3.0.4
- How you're running Floci: Docker (`FLOCI_STORAGE_MODE=memory`)

## Related / prior work

- **#829** ("CloudFormation does not apply SQS `ContentBasedDeduplication`...", closed/completed
  2026-05-21) fixed the CloudFormation path for `ContentBasedDeduplication` — which is consistent
  with `ContentBasedDeduplication` now surviving here. This report is the remaining gap in the same
  CloudFormation-to-SQS attribute mapping for the FIFO attributes `DeduplicationScope`,
  `FifoThroughputLimit`, and `RedrivePolicy`, which that fix did not cover.

## Secondary observation (likely already tracked — mentioned for completeness)

The stock `cdklocal bootstrap` logs a `CREATE_FAILED` for `AWS::ECR::Repository`
(`ContainerAssetsRepository` — "Failed to start ECR backing registry container") while still
reporting the environment as bootstrapped. This appears to be the already-completed **#592**
("`cdk bootstrap` succeeds, but multiple `CREATE_FAILED` outputs logged", closed/completed
2026-04-25). It does not affect this report: the MWE stack has no image assets, so the ECR
resource is never needed and deploy proceeds normally. Flagging only in case it should be
re-opened or verified against 1.5.32.

Agboolafeyikemi:
Fixes #1907

## Problem

A CloudFormation stack creating a FIFO `AWS::SQS::Queue` reached `CREATE_COMPLETE`, but `DeduplicationScope`, `FifoThroughputLimit`, and `RedrivePolicy` were silently dropped `GetQueueAttributes` returned them as null. As the issue's Control 2 showed, the SQS engine itself persists all three when called directly; only the CloudFormation→SQS mapping lost them.

**Root cause:** `SqsCfnProvisioner.provisionQueue()` only copied `VisibilityTimeout` and `ContentBasedDeduplication` into the attributes map passed to `SqsService.createQueue`. Everything else in the template's `Properties` was ignored.

**Adjacent gap in the same repro:** the issue's CDK app creates the DLQ with `fifo: true` and no `QueueName`. Since `FifoQueue` was also unmapped and generated physical names never end in `.fifo`, that DLQ was silently created as a **standard** queue.

## Fix

All in `SqsCfnProvisioner.provisionQueue()`; no `SqsService` changes needed:

- Pass `DeduplicationScope` and `FifoThroughputLimit` through (same pattern as the existing two attributes).
- Serialize `RedrivePolicy` via `engine.resolveNode(...).toString()` so the `Fn::GetAtt` DLQ ARN inside it is resolved  the same pattern the SNS subscription provisioner already uses.
- Pass `FifoQueue` through, and suffix generated physical names with `.fifo` when `FifoQueue: true` and `QueueName` is absent (matching real CloudFormation; `SqsService` rejects `FifoQueue=true` names without the suffix).

## Before / after

Reproduction tests were written and run against the unmodified code first:

**Before**  unit test (attrs map handed to `createQueue`):

```
SqsCfnProvisionerTest.queuePassesFifoScalarAttributesToCreateQueue
  expected: <messageGroup> but was: <null>
SqsCfnProvisionerTest.queueSerializesRedrivePolicyWithResolvedDlqArn
  expected: <{"deadLetterTargetArn":"arn:aws:sqs:...","maxReceiveCount":5}> but was: <null>
```

**Before** end-to-end (CreateStack → DescribeStackResources/GetQueueAttributes): stack reports `CREATE_COMPLETE`, but the no-name FIFO DLQ was created as a standard queue:

```
generated DLQ physical name should end with .fifo but was:
cfn-1907-fifo-attrs-stack-Dlq-4080e58c53dd
```

**After**  same tests green: the main queue's `GetQueueAttributes` returns `DeduplicationScope=messageGroup`, `FifoThroughputLimit=perMessageGroupId`, and a `RedrivePolicy` containing the resolved `.fifo` DLQ ARN; the generated-name DLQ is itself FIFO.

## Testing

- `SqsCfnProvisionerTest`: 9/9 (4 new tests, including the already-serialized `RedrivePolicy` case below)
- New end-to-end test in `CloudFormationIntegrationTest` mirroring the issue's repro (FIFO DLQ without `QueueName` + main FIFO queue with all six properties and `Fn::GetAtt` redrive)
- Regression guards: `SqsServiceTest` 74/74; full `cloudformation` + `sqs` packages 651/651 green on current `main` after the down-merge below (the `createStack_ecrAutoName_isLowercase` failure noted earlier is environment-dependent and does not reproduce in CI or on a maintainer machine)

Out of scope, per one-fix-per-PR: other unmapped queue properties (`DelaySeconds`, `KmsMasterKeyId`, `MessageRetentionPeriod`, …).

---

### Maintainer notes (added on review)

**CDK `Fn::Join` `RedrivePolicy`.** Thanks to @floverfelt for spotting that an already-serialized
`RedrivePolicy` string was being double-encoded, and for pinpointing the cause: `resolveNode`
collapses `Fn::Join` to a `TextNode`, whose `toString()` re-emits the JSON representation. The fix
in this branch unwraps textual nodes, and `queuePassesThroughAlreadySerializedRedrivePolicyString`
pins it — reverting the unwrap reproduces the double-encoded value exactly.

**Down-merged by a maintainer.** This branch was 185 commits behind `main`, conflicting only in
`CloudFormationIntegrationTest` where git had interleaved this PR's new test with tests main added
nearby. Resolved by taking main's file and re-applying this branch's two additions, since the change
to that file is purely additive. This PR's own diff is unchanged.

**Still open, tracked separately:** `AWS::SNS::Subscription` has the same latent double-encoding at
`CloudFormationResourceProvisioner:1576`, also found by @floverfelt. Out of scope here.


floverfelt:
+1 I ran into this today - this would be a super nice addition.

floverfelt:
Found this bug while testing locally.

I hit an edge case with it: when `RedrivePolicy` is supplied as an **already-serialized JSON string** rather than a JSON object, the policy comes out double-encoded and the queue's redrive silently doesn't work.

CDK commonly emits this form — the DLQ ARN gets spliced in with `Fn::Join`:

```json
"RedrivePolicy": {
  "Fn::Join": ["", [
    "{\"maxReceiveCount\":5,\"deadLetterTargetArn\":\"",
    { "Fn::GetAtt": ["SomeDLQ", "Arn"] },
    "\"}"
  ]]
}
```

With this PR, `GetQueueAttributes` then returns:

```json
"RedrivePolicy": "\"{\\\"maxReceiveCount\\\":5,\\\"deadLetterTargetArn\\\":\\\"arn:aws:sqs:us-east-1:000000000000:some-dlq\\\"}\""
```

**Root cause:** `CloudFormationTemplateEngine.resolveNode` collapses the `Fn::Join` to a `TextNode`, and `SqsCfnProvisioner` calls `.toString()` on the result. Jackson's `TextNode.toString()` emits the JSON *representation* of the string — quoted and escaped — so the already-serialized policy gets JSON-encoded a second time, and the SQS service can't parse it. Real CloudFormation accepts both the object form and the string form for `RedrivePolicy`.

**Suggested fix** — unwrap textual nodes:

```java
JsonNode redrive = ctx.engine().resolveNode(props.path("RedrivePolicy"));
attrs.put("RedrivePolicy", redrive.isTextual() ? redrive.asText() : redrive.toString());
```

The object form tested in the PR keeps working since `.toString()` is still used for non-textual nodes.

Note the SNS subscription provisioning has the same latent issue — `CloudFormationResourceProvisioner` also does `engine.resolveNode(props.path("RedrivePolicy")).toString()` for `AWS::SNS::Subscription`.

pgermosen (approved):
@Agboolafeyikemi — approving. All 13 checks are green on the down-merged head, and the work holds up under scrutiny.

## Still needed, in full

The first thing I checked, given this sat for three weeks: whether anything had overtaken it. Nothing has. `main`'s `SqsCfnProvisioner` today maps exactly two attributes — `VisibilityTimeout` and `ContentBasedDeduplication` — and generates names with no `.fifo` handling at all. Every gap #1907 reports is still live. Unlike some PRs that shrink while they wait, this one lost nothing.

## Verified locally

On current `main` (`0971cf6f`), after resolving the merge:

- `test-compile` clean, **651 tests green** across `services.cloudformation.**` and `services.sqs.**`
- `SqsCfnProvisionerTest` 9/9

I reverted each behaviour to check your tests actually pin them rather than merely accompany them:

| Reverted | Failure |
|---|---|
| the `isTextual()` unwrap | `queuePassesThroughAlreadySerializedRedrivePolicyString` — returns `"{\"maxReceiveCount\":5,…}"`, double-encoded |
| `DeduplicationScope` / `FifoThroughputLimit` pass-through | `expected: <messageGroup> but was: <null>` |
| the `.fifo` suffix on generated names | *"generated FIFO queue name should end with .fifo but was: my-stack-MyDlq-3fd1e6236942"* |

Each one reproduces a distinct symptom from the original report, which is what you get when tests are written from the observed failure rather than around the finished fix. Your "before/after" section said as much, and it checks out.

## The part that impressed me

Handling @floverfelt's find properly rather than deferring it.

They reported that a CDK-style `Fn::Join` `RedrivePolicy` came out double-encoded, and traced it to `resolveNode` collapsing to a `TextNode` whose `toString()` re-emits the JSON representation. You took the fix, added the comment explaining *why* the unwrap is there — which is what stops someone "simplifying" it back to `.toString()` later — and added a dedicated test. That is the difference between patching a report and closing it.

The adjacent catch in your own repro deserves a mention too: noticing that the no-name FIFO DLQ was silently created as a **standard** queue, because generated names never carried `.fifo`, is the kind of thing that only surfaces if you actually run the scenario end to end rather than test the property you set out to fix.

## Housekeeping I did on your branch

Since this was blocked purely on drift, I down-merged current `main` rather than send you round again. The only conflict was in `CloudFormationIntegrationTest`, where git had interleaved your new test with tests `main` added in the same region — resolved by taking main's file and re-applying your two additions, since your change there is purely additive. Your own diff is unchanged: still three files, 243 insertions.

I also refreshed two stale lines in the description (the test count moved from 8 to 9 when the `RedrivePolicy` case landed, and the `createStack_ecrAutoName_isLowercase` failure you saw does not reproduce here — it is environment-dependent and I have now seen it on several unrelated PRs today).

Apologies for editing your branch and description without asking first; say the word if you would rather keep that in your hands next time.

## One follow-up, not yours

@floverfelt also spotted that `AWS::SNS::Subscription` has the same latent double-encoding at `CloudFormationResourceProvisioner:1576` — a bare `engine.resolveNode(...).toString()` with no textual unwrap. Correctly out of scope here; it will get its own issue.

Thank you for the patience, and for the quality of the write-up — the before/after failure output made this reviewable in a fraction of the time it otherwise would have taken.

PR #1940, opened 2026-07-24
fix(cloudfront): include terraform-required distribution fields

### LINKED ISSUE #1930: [BUG] CloudFront Create/GetDistribution omits OriginGroups (and Restrictions), crashing Terraform AWS provider
## Service

CloudFront

## AWS API Action

`CreateDistribution`, `CreateDistributionWithTags`, `GetDistribution` (response serialization of `DistributionConfig`)

## Expected behavior

Real AWS always returns `DistributionConfig.OriginGroups` (at least `{ Quantity: 0 }` when none are configured) and `DistributionConfig.Restrictions` (e.g. geo restriction `none`) on Create/GetDistribution responses.

Terraform AWS provider `hashicorp/aws` (observed with **v6.54.0**) relies on that shape and flattens with:

```go
if aws.ToInt32(distributionConfig.OriginGroups.Quantity) > 0 {
```

so `OriginGroups` must be non-nil even when empty.

## Actual behavior

Floci (`floci/floci:1.5.32`) accepts CreateDistribution and returns a distribution, but the XML/JSON `DistributionConfig` **omits** `OriginGroups` and `Restrictions` entirely.

That causes `hashicorp/aws` to panic on create/read:

```text
github.com/hashicorp/terraform-provider-aws/internal/service/cloudfront.resourceDistributionFlatten
        .../internal/service/cloudfront/distribution.go:1150
```

OAC create/list works; the failure is specifically on distribution create/get response flattening.

Observed Floci `GetDistribution` / create response `DistributionConfig` keys include Origins, DefaultCacheBehavior, CacheBehaviors, Aliases, ViewerCertificate, etc., but **not** `OriginGroups` or `Restrictions`.

Sending `OriginGroups: { Quantity: 0 }` and a `Restrictions` block on create still yields responses without those fields.

This matches Floci's `DistributionConfig` model lacking an origin-groups field (see upstream `DistributionConfig.java`), so empty groups are never serialized.

Same failure class as prior Terraform-compat bugs where omitted response fields crash the provider (e.g. missing API Gateway fields).

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=test

cat > /tmp/cf-dist.json <<'JSON'
{
  "CallerReference": "floci-origingrouproups-repro-1",
  "Comment": "repro",
  "Enabled": true,
  "Origins": {
    "Quantity": 1,
    "Items": [
      {
        "Id": "origin1",
        "DomainName": "example-bucket.s3.us-east-1.amazonaws.com",
        "S3OriginConfig": { "OriginAccessIdentity": "" }
      }
    ]
  },
  "OriginGroups": { "Quantity": 0 },
  "DefaultCacheBehavior": {
    "TargetOriginId": "origin1",
    "ViewerProtocolPolicy": "redirect-to-https",
    "TrustedSigners": { "Enabled": false, "Quantity": 0 },
    "ForwardedValues": {
      "QueryString": false,
      "Cookies": { "Forward": "none" }
    },
    "MinTTL": 0
  },
  "Restrictions": {
    "GeoRestriction": { "RestrictionType": "none", "Quantity": 0 }
  },
  "ViewerCertificate": { "CloudFrontDefaultCertificate": true }
}
JSON

aws cloudfront create-distribution --distribution-config file:///tmp/cf-dist.json

# Inspect raw XML (or JSON via CLI): OriginGroups / Restrictions are missing
ID=$(aws cloudfront list-distributions --query 'DistributionList.Items[0].Id' --output text)
curl -sS "$AWS_ENDPOINT_URL/2020-05-31/distribution/$ID" | tee /tmp/cf-get.xml
grep -E 'OriginGroups|Restrictions' /tmp/cf-get.xml || echo "MISSING OriginGroups/Restrictions"

# Terraform path (same crash):
# resource "aws_cloudfront_distribution" "this" { ... S3 origin + OAC ... }
# terraform apply  -> provider panic at distribution.go:1150
```

### Suggested fix

On Create/Get/UpdateDistribution (and list summaries that embed config), always serialize at least:

```xml
<OriginGroups><Quantity>0</Quantity></OriginGroups>
<Restrictions>
  <GeoRestriction>
    <RestrictionType>none</RestrictionType>
    <Quantity>0</Quantity>
  </GeoRestriction>
</Restrictions>
```

when unset, matching AWS. Also consider defaulting related empty wrappers the provider may touch (`ActiveTrustedKeyGroups` / `ActiveTrustedSigners` on the distribution, `CachedMethods` under `AllowedMethods` when present).

## Environment

- Floci version / image tag: `floci/floci:1.5.32` (also reproduced on local `latest` / `latest-compat` images available here)
- Java SDK version (if applicable): n/a (AWS CLI v2 + Terraform `hashicorp/aws` **6.54.0**)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (`hdr-tools-floci` via docker-compose, port `4566`)
- Terraform: v1.15.x

ousamabenyounes:
## Summary
- include empty CloudFront OriginGroups in distribution config and summaries
- include CloudFront Restrictions/GeoRestriction defaults in create/get/list distribution responses
- parse GeoRestriction from incoming distribution config so existing values can be echoed

Fixes #1930

## Testing
- RED: `./mvnw test -Dtest=CloudFrontServiceTest` failed before the fix because `OriginGroups` and `Restrictions` were absent from the response XML
- GREEN: `./mvnw test -Dtest=CloudFrontServiceTest` passes

## Notes
- Full `./mvnw test` did not complete cleanly locally because unrelated API Gateway/Lambda Docker tests timed out (`ApiGatewayAuthorizerContextIntegrationTest` / `ApiGatewayRequestAuthorizerIntegrationTest`). The targeted CloudFront regression test is green.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
@@ -29,6 +29,16 @@ public class CloudFrontController {
 
     private static final String NS = AwsNamespaces.CLOUDFRONT;
     private static final String XML = "application/xml";
+    private static final String GEO_RESTRICTION = "GeoRestriction";
+    private static final String ORIGIN_GROUPS = "OriginGroups";
+    private static final String ITEMS = "Items";
+    private static final String LOCATION = "Location";
+    private static final String LOCATIONS = "Locations";
+    private static final String QUANTITY = "Quantity";
+    private static final String RESTRICTION_TYPE = "RestrictionType";
+    private static final String RESTRICTIONS = "Restrictions";
+    private static final String DEFAULT_GEO_RESTRICTION_TYPE = "none";
+    private static final int EMPTY_QUANTITY = 0;
 
     private static final XMLInputFactory XML_FACTORY;
 
@@ -1760,6 +1770,7 @@ public class CloudFrontController {
         List<Origin> origins = cfg.getOrigins();
         xml.raw(xmlQuantityItems("Origins", "Origin", origins != null ? origins.size() : 0,
                 origins != null ? origins.stream().map(this::xmlOrigin).toList() : List.of()));
+        xml.raw(xmlEmptyOriginGroups());
 
         if (cfg.getDefaultCacheBehavior() != null) {
             xml.raw(xmlDefaultCacheBehavior(cfg.getDefaultCacheBehavior()));
@@ -1815,8 +1826,67 @@ public class CloudFrontController {
         xml.end("Aliases");
 
         xml.raw(xmlViewerCertificate(cfg.getViewerCertificate()));
+        xml.raw(xmlRestrictions(cfg.getGeoRestriction()));
 
         return xml.build();
+    }
+
+    private String xmlEmptyOriginGroups() {
+        // Presence-only OriginGroups is intentional for Terraform compatibility; round-tripping groups is deferred.
+        return new XmlBuilder()
+                .start(ORIGIN_GROUPS)
+                .elem(QUANTITY, EMPTY_QUANTITY)
+                .end(ORIGIN_GROUPS)
+                .build();
+    }
+
+    private String xmlRestrictions(Map<String, Object> geoRestriction) {
+        String restrictionType = DEFAULT_GEO_RESTRICTION_TYPE;
+        int quantity = EMPTY_QUANTITY;
+        List<String> locations = List.of();
+        if (geoRestriction != null) {
+            restrictionType = String.valueOf(
+                    geoRestriction.getOrDefault(RESTRICTION_TYPE, DEFAULT_GEO_RESTRICTION_TYPE));
+            locations = stringList(geoRestriction.get(LOCATIONS));
+            quantity = parseInt(geoRestriction.get(QUANTITY), EMPTY_QUANTITY);
+            if (quantity == EMPTY_QUANTITY && !locations.isEmpty()) {
+                quantity = locations.size();
+            }
+        }
+
+        XmlBuilder xml = new XmlBuilder()
+                .start(RESTRICTIONS)
+                .start(GEO_RESTRICTION)
+                .elem(RESTRICTION_TYPE, restrictionType)
+                .elem(QUANTITY, quantity);
+        if (!locations.isEmpty()) {
+            xml.start(ITEMS);
+            for (String location : locations) {
+                xml.elem(LOCATION, location);
+            }
+            xml.end(ITEMS);
+        }
+        return xml.end(GEO_RESTRICTION)
+                .end(RESTRICTIONS)
+                .build();
+    }
+
+    private List<String> stringList(Object value) {
+        if (value instanceof List<?> list) {
+            return list.stream().map(String::valueOf).toList();
+        }
+        return List.of();
+    }
+
+    private int parseInt(Object value, int defaultValue) {
+        if (value == null) {
+            return defaultValue;
+        }
+        try {
+            return Integer.parseInt(value.toString());
+        } catch (NumberFormatException e) {
+            return defaultValue;
+        }
     }
 
     private String xmlOrigin(Origin o) {
@@ -2013,6 +2083,7 @@ public class CloudFrontController {
                 origins != null ? origins.size() : 0,
                 origins != null ? origins.stream().map(this::xmlOrigin).toList()
                         : List.of()));
+        xml.raw(xmlEmptyOriginGroups());
 
         List<String> aliases = cfg != null ? cfg.getAliases() : null;
         int aliasCount = aliases != null ? aliases.size() : 0;
@@ -2027,6 +2098,7 @@ public class CloudFrontController {
         xml.end("Aliases");
 
         xml.raw(xmlViewerCertificate(cfg != null ? cfg.getViewerCertificate() : null));
+        xml.raw(xmlRestrictions(cfg != null ? cfg.getGeoRestriction() : null));
 
         xml.end("DistributionSummary");
         return xml.build();
@@ -2245,6 +2317,7 @@ public class CloudFrontController {
         cfg.setAliases(parseAliases(body));
         cfg.setViewerCertificate(parseViewerCertificate(body));
         cfg.setCustomErrorResponses(parseCustomErrorResponses(body));
+        cfg.setGeoRestriction(parseGeoRestriction(body));
 
         return cfg;
     }
@@ -2376,6 +2449,19 @@ public class CloudFrontController {
             LOG.debugv("Ignoring malformed CustomErrorResponses during parse: {0}", e.getMessage());
         }
         return result;
+    }
+
+    private Map<String, Object> parseGeoRestriction(String body) {
+        List<Map<String, String>> groups = XmlParser.extractGroups(body, GEO_RESTRICTION);
+        if (groups.isEmpty()) {
+            return Map.of();
+        }
+        Map<String, Object> geoRestriction = new LinkedHashMap<>(groups.getFirst());
+        List<String> locations = XmlParser.extractAll(body, LOCATION);
+        if (!locations.isEmpty()) {
+            geoRestriction.put(LOCATIONS, locations);
+        }
+        return geoRestriction;
     }
 
     private List<Origin> parseOrigins(String body) {
--- a/src/test/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontServiceTest.java
@@ -2,6 +2,7 @@ package io.github.hectorvent.floci.services.cloudfront;
 
 import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.core.common.XmlParser;
 import io.github.hectorvent.floci.core.storage.AccountAwareStorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageFactory;
 import io.github.hectorvent.floci.services.cloudfront.model.CacheBehavior;
@@ -15,6 +16,7 @@ import io.github.hectorvent.floci.services.cloudfront.model.OriginAccessControl;
 import io.github.hectorvent.floci.services.cloudfront.model.PublicKey;
 import io.github.hectorvent.floci.services.cloudfront.model.ResponseHeadersPolicy;
 import io.github.hectorvent.floci.services.cloudfront.model.StreamingDistribution;
+import jakarta.ws.rs.core.Response;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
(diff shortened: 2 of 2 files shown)


ousamabenyounes:
Follow-up pushed for the review comments.

Handled:
- preserved `GeoRestriction.Items/Location` values for whitelist/blacklist round-trips in create/get/list distribution responses
- moved the new coverage into `CloudFrontServiceTest` to match the repo test naming convention while keeping the existing CloudFront service tests

Validation:
- RED: `./mvnw test -Dtest=CloudFrontServiceTest` failed before the fix because whitelist locations were missing from the response XML
- GREEN: `./mvnw clean test -Dtest=CloudFrontServiceTest` passed (`5` tests, `0` failures)
- Codex review: no blocking findings

Both review threads are resolved.

src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
```diff
@@ -1780,10 +1791,68 @@ private String xmlDistributionConfigBody(DistributionConfig cfg) {
         xml.end("Aliases");
 
         xml.raw(xmlViewerCertificate(cfg.getViewerCertificate()));
+        xml.raw(xmlRestrictions(cfg.getGeoRestriction()));
 
         return xml.build();
     }
 
+    private String xmlEmptyOriginGroups() {
```
pgermosen, line 1799: **`OriginGroups` is always emitted empty — asymmetric with `GeoRestriction`** (note / question).

`xmlEmptyOriginGroups()` hardcodes `<OriginGroups><Quantity>0</Quantity></OriginGroups>` and there's no parse path, so a client that submits a config *with* origin groups gets them silently dropped and echoed back empty. That's fine for #1930 (Terraform only needs the element to exist), but it's worth noting the asymmetry: `GeoRestriction` right next to it *is* parsed and preserved. Suggest either a one-line comment / `// TODO` that presence-only `OriginGroups` is intentional for now, or a follow-up issue to round-trip it the same way — so the gap is documented rather than surprising.

Is presence-only `OriginGroups` the intended scope here, with round-tripping deferred?

src/test/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontServiceTest.java
```diff
@@ -16,51 +18,172 @@
...
 
         StreamingDistribution sd = service.createStreamingDistribution(new StreamingDistribution());
 
-        assertTrue(sd.getDomainName().endsWith(".cloudfront.local"),
+        assertTrue(sd.getDomainName().endsWith(DOMAIN_SEPARATOR + CONFIGURED_DOMAIN_SUFFIX),
                 "Expected configured suffix, got: " + sd.getDomainName());
     }
+
+    @Test
+    void createDistributionIncludesEmptyOriginGroupsAndRestrictions() {
```
pgermosen, line 127: **Tests were added to `CloudFrontServiceTest`, but the PR describes `CloudFrontControllerTest`** (minor).

The PR's Testing section says the RED/GREEN repro is `./mvnw test -Dtest=CloudFrontControllerTest`, but there's no `CloudFrontControllerTest` in the tree — these new tests live in `CloudFrontServiceTest` and drive the controller directly (`new CloudFrontController(...)`, `createDistribution`/`getDistribution`/`listDistributions`). So the documented command matches no test. Two easy options: fix the description to point at `CloudFrontServiceTest`, or move these controller-level cases into a `CloudFrontControllerTest` (closer to CLAUDE.md's `*ServiceTest` vs. controller-test split). Purely cosmetic, but it'll trip up anyone following the repro steps.

pgermosen (approved):
Looks good.

(Just 2 non-blocking comments)

ousamabenyounes:
Follow-up pushed for the review threads.

Handled:
- documented that `OriginGroups` is intentionally presence-only for Terraform compatibility, with group round-tripping deferred
- corrected the PR Testing section to reference `CloudFrontServiceTest`

Validation:
- `./mvnw test -Dtest=CloudFrontServiceTest` passed locally (`5` tests, `0` failures)
- GitHub checks are green on `7ed28757b3975fa3893e84e4f087d2309e68a5ae` (`14` successful, `0` pending, `0` failing), including Greptile

pgermosen:
Hi @ousamabenyounes, this branch has fallen behind `main` and now has real merge conflicts — confirmed with an actual merge attempt, not just the GitHub status indicator. Conflicts are in `CloudFrontController.java` and `CloudFrontServiceTest.java`. Could you rebase onto current `main` when you get a chance? Happy to take another look once that's done.

pgermosen (approved):
Approving again on the rebased head. Thanks for turning the rebase around quickly — I re-verified the whole thing rather than assuming the force-push was a no-op.

Re-reviewed after the rebase (`f98f4b9d`). Verified locally against current `main` (`04d120fd`).

## The rebase is clean, so the July approval still applies

I re-derived the pre-rebase diff (`7ed28757`) and compared it to the current one. The
`CloudFrontController.java` additions are **byte-identical** — 86 added lines, `diff` empty. The only
changes are in `CloudFrontServiceTest.java`, where the conflict was the test harness
(`serviceWithDomainSuffix`, the domain-suffix constants, the `InMemoryStorage` stub) that `main` now
provides on its own, plus constant extraction (`LIST_MAX_ITEMS`, `ID_ELEMENT`, `DEFAULT_DOMAIN_SUFFIX`).
No behaviour was smuggled in during the conflict resolution.

## What I verified

**The hole is real, and the model is more specific than the issue is.** In the CloudFront `2020-05-31`
model, `Restrictions` is a **required** member of `DistributionSummary` — so `ListDistributions` was
emitting a shape the service model itself says is invalid. `OriginGroups`, by contrast, is *not*
required in either `DistributionConfig` or `DistributionSummary`; the provider dereferences it
unconditionally (`aws.ToInt32(distributionConfig.OriginGroups.Quantity)`), which is arguably a provider
bug. Both halves are worth doing, but they're different kinds of fix: one is conformance, one is
accommodating a provider assumption. Worth knowing which is which if the provider ever changes.

**Every response path is covered.** `DistributionConfig` reaches the wire through exactly three
places, and all three go through the two shared builders this PR touches:

- `GET /distribution/{id}/config` — `CloudFrontController.java:102`
- the `Distribution` envelope used by Create / Get / Update / Copy — `:1749`
- `DistributionSummary` for `ListDistributions` — `:2061`

**This wires up a field that was already declared and dead.** `DistributionConfig.geoRestriction`
exists on `main` but is never set and never emitted — `git grep geoRestriction origin/main -- src/main`
returns only the field and its accessors. So the parse side isn't new state, it's finally populating
something the model already promised.

**The issue's other nil-deref candidates are already handled.** `xmlDistribution` already emits
`ActiveTrustedKeyGroups` and `ActiveTrustedSigners`, so the "also consider" list in #1930 is covered.

**Greptile's note about `extractAll` not being scoped to the `GeoRestriction` subtree is correct as
written, and provably harmless.** `Location` occurs exactly once as a *body* element in the entire
CloudFront model — `LocationList`'s member, `locationName: Location`. Every other `Location` in the
model is an HTTP response header (`"location": "header"` on the various `Create*Result` shapes). A
document-wide sweep over any request body this controller parses therefore cannot pick up a foreign
`Location`. It's also not a shortcut: `extractGroups` reads children via `readLeafText`, which returns
`null` for non-leaf children, so `<Items>` is dropped by construction and the second pass is the only
thing that works with the current parser. A one-line comment saying so would save the next reader the
trip — that's the whole ask on this point.

**Local run:** 163/163 CloudFront tests green on the PR head merged with current `main`.

## Mutation testing — 7 of 8 caught

| # | Mutation | Result |
|---|---|---|
| M1 | drop `xmlEmptyOriginGroups()` from `xmlDistributionConfigBody` | caught — `createDistributionIncludesEmptyOriginGroupsAndRestrictions` |
| M2 | drop `xmlRestrictions(...)` from `xmlDistributionConfigBody` | caught — both new tests |
| M3 | drop `xmlEmptyOriginGroups()` from `xmlDistributionSummary` | caught |
| M4 | drop `xmlRestrictions(...)` from `xmlDistributionSummary` | caught — both new tests |
| M5 | drop `cfg.setGeoRestriction(parseGeoRestriction(body))` | caught — `createDistributionPreservesGeoRestrictionLocations` |
| M6 | never emit the `<Items>` block in `xmlRestrictions` | caught |
| M7 | disable the `quantity == 0 && !locations.isEmpty()` fallback | **survived** |
| M8 | make `parseGeoRestriction` never collect locations | caught |

The create/get/list assertion helpers pull real weight: M3 and M4 are summary-path-only and both die.

M7 survives because no test sends `Quantity=0` together with `Items` — the fallback branch is never
entered. I confirmed the mutated class actually recompiles (timestamp advances, 163 tests still run),
so this is a genuine gap, not a build artifact. It's a defensive default, so low stakes on its own —
but see the next section, where the same input class shows up again.

## Non-blocking

**1. `Quantity` is echoed verbatim and never validated.** Probing the controller directly:

| request | result |
|---|---|
| `whitelist`, `Quantity=5`, 2 `Location`s | 201, response echoes `Quantity=5` with 2 `Items` |
| `whitelist`, `Quantity=2`, no `Items` | 201, response echoes `Quantity=2` with no `Items` |
| `whitelist`, `Quantity=0`, 2 `Location`s | 201, silently rewritten to `Quantity=2` (the M7 branch) |

Real CloudFront answers `InconsistentQuantities` to the first two. The second one is notable because
it's the exact "quantity > 0 but no `Items`" drift shape greptile warned about in the original review
— the parse side is fixed, but a client can still push that shape straight through to the response.

This controller already enforces the rule for two other lists — `inconsistentQuantities()` for
`Origins`, `validateTrustedKeyGroupsQuantity` for `TrustedKeyGroups`. Extending it to `GeoRestriction`
is a handful of lines and kills M7 and both drift shapes together. Not a blocker: every case above is
input AWS would have rejected outright, so nothing valid regresses.

**2. `RestrictionType` is echoed unvalidated.** `<RestrictionType>allowlist</RestrictionType>` returns
201 and comes back as `allowlist`. The model's `GeoRestrictionType` is `blacklist | whitelist | none`.
Lowest priority — no SDK can produce it.

**3. The regression guard doesn't cover the actual failure mode.** The new tests assert XML substrings.
#1930 is a *deserialization* crash, and nothing in the repo exercises that path: `compat-terraform` has
no CloudFront coverage at all — no resource in `main.tf`, not even a `cloudfront` entry in
`provider.tf`'s `endpoints` block — so the green `native / compat-terraform` check is not evidence the
panic is gone.

There's already a real-SDK harness that does exactly the right thing:
`compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudFrontTrustedKeyGroupsTest.java`
calls `cloudFront.createDistribution(...)` and `getDistribution(...)` through the AWS Java SDK.
Asserting `originGroups().quantity() == 0` and a geo-restriction round-trip there is ~15 lines and
guards the thing the issue is actually about — that a real SDK parser sees these fields.

**4. Test placement.** These are controller-level tests (`new CloudFrontController(...)`) sitting in
`CloudFrontServiceTest`. When this PR was opened there was no `CloudFrontControllerTest`, which is why
my July comment pointed the other way; `main` has one now. Cosmetic, and moving them would re-conflict
the branch — mentioning it only so it isn't read as settled convention.

## Adjacent, out of scope — worth their own issues

While checking `DistributionSummary` against the model I found `ListDistributions` is missing five
members the model marks **required**. Confirmed against a live response, not just by reading:

- `DefaultCacheBehavior`, `CacheBehaviors`, `CustomErrorResponses`, `Staging` — absent entirely.
- `WebACLId` — emitted as `<WebAclId>`. The model member is `WebACLId`, and `isDistributionConfigScalar`
  accepts `WebAclId` on the parse side too, so a client sending the real `<WebACLId>` has it silently
  dropped, and the value Floci does emit is invisible to every SDK deserializer. That's the same
  failure class as #1930 — right value, wrong element name, so it may as well not be there. Terraform
  would show a permanently empty `web_acl_id`.

None of this is this PR's job. Flagging it because it's the same shape of bug in the same method, and
the next Terraform report will probably be one of these.

PR #1949, opened 2026-07-26
feat(sagemaker): add SageMaker emulation with Docker-backed training

### LINKED ISSUE #1945: [FEAT] Adding SageMaker Service
## Service

Sagemaker due to high spikes in AI Engineering.

## API Action / Feature

Full feature in this case

## AWS Documentation

https://aws.amazon.com/pt/sagemaker/ai/

## Why is this needed?

As I said it has been asked on interviews due to relationship with AI.

- [X] Yes
- [ ] No

mbergo:
Closes #1945

Implements the SageMaker control plane over the AWS JSON 1.1 protocol (`X-Amz-Target: SageMaker.*`) plus the SageMaker Runtime invocation path, backed by real Docker containers rather than mocks.

## Summary

- Models, endpoint configs, endpoints, training jobs, and tags with AWS-shaped validation errors (`ValidationException`/`ResourceInUse`), all scoped by region since the resource ARNs are regional.
- `CreateTrainingJob` runs the training image as a real Docker container using the `/opt/ml` contract: hyperparameters, `resourceconfig`, and `inputdataconfig` under `/opt/ml/input/config`, S3 channel data under `/opt/ml/input/data/<channel>` (preserving the relative path structure under the channel's S3 prefix, with a path-traversal guard), and `/opt/ml/model` uploaded back to S3 as `model.tar.gz`.
- `CreateEndpoint` hosts the model image as a long-lived container with the `serve` command, `/ping` health checks, and `ModelDataUrl` artifacts extracted into `/opt/ml/model`. A start still in flight when the endpoint is deleted or updated again discards its result instead of resurrecting a deleted endpoint or clobbering a newer one.
- `POST /endpoints/{name}/invocations` proxies to the container, resolves the calling region from the request's SigV4 credential scope, maps model 5xx responses to `ModelError` (HTTP 424), and returns `application/json` error envelopes.
- `StopTrainingJob` only transitions a job that is `InProgress`/`Stopping`; the async runner distinguishes an intentional stop from the container disappearing as a side effect of it, so a stopped job's terminal state is not overwritten with `Failed`.
- `ListModels`/`ListEndpointConfigs`/`ListEndpoints`/`ListTrainingJobs` validate `MaxResults` (1-100) and support `NextToken` pagination; `ListModels`/`ListTrainingJobs` additionally support `NameContains`, and `ListTrainingJobs` supports `StatusEquals`.
- Training container logs stream to CloudWatch Logs under `/aws/sagemaker/TrainingJobs`.
- Storage wired through `StorageFactory` with memory/persistent modes, config in `EmulatorConfig` and both `application.yml` files.
- Unit, integration, and Docker-gated integration tests (including regressions for the stop/delete races above) plus a boto3 compatibility test; service, action tables, and docs registered in the catalog.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified with boto3/botocore (the versions pinned in `compatibility-tests/sdk-test-python/requirements.txt`) via `tests/test_sagemaker.py`, which exercises the full control plane and a real training job end-to-end. The JSON 1.1 wire shapes (`X-Amz-Target: SageMaker.*`, `application/x-amz-json-1.1`, `__type` error envelopes) are additionally asserted at the HTTP level by `SageMakerIntegrationTest`. Cross-checked against the SageMaker API reference and the AWS SDK for Java v2 model for `CreateModel`, `CreateTrainingJob`, `ListModels`/`ListTrainingJobs` pagination and filters, and `StopTrainingJob`'s status-transition contract.

## Checklist

- [x] The full test suite passes locally

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerEndpointManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerEndpointManager.java
@@ -178,11 +178,14 @@ public class SageMakerEndpointManager implements ContainerTeardown {
         try (TarArchiveOutputStream tar = new TarArchiveOutputStream(out)) {
             if (key.endsWith(".tar.gz") || key.endsWith(".tgz")) {
                 try (TarArchiveInputStream in = new TarArchiveInputStream(new GzipCompressorInputStream(new ByteArrayInputStream(data)))) {
-                    TarArchiveEntry e;
-                    while ((e = in.getNextEntry()) != null) {
-                        if (!in.canReadEntryData(e) || e.isDirectory()) continue;
-                        addFile(tar, "opt/ml/model/" + e.getName(), in.readAllBytes());
-                    }
+TarArchiveEntry e;
+while ((e = in.getNextEntry()) != null) {
+    if (!in.canReadEntryData(e) || e.isDirectory()) continue;
+    String entryName = e.getName();
+    java.nio.file.Path normalized = java.nio.file.Path.of(entryName).normalize();
+    if (normalized.isAbsolute() || normalized.startsWith("..")) continue;
+    addFile(tar, "opt/ml/model/" + normalized.toString(), in.readAllBytes());
+}
                 }
             } else {
                 addFile(tar, "opt/ml/model/" + key.substring(key.lastIndexOf('/') + 1), data);
--- a/src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerEndpointManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerEndpointManager.java
@@ -1,0 +1,243 @@
+package io.github.hectorvent.floci.services.sagemaker;
+
+import io.github.hectorvent.floci.config.EmulatorConfig;
+import io.github.hectorvent.floci.core.common.ContainerTeardown;
+import io.github.hectorvent.floci.core.common.dns.EmbeddedDnsServer;
+import io.github.hectorvent.floci.core.common.docker.ContainerBuilder;
+import io.github.hectorvent.floci.core.common.docker.ContainerDetector;
+import io.github.hectorvent.floci.core.common.docker.ContainerLifecycleManager;
+import io.github.hectorvent.floci.core.common.docker.ContainerSpec;
+import io.github.hectorvent.floci.core.common.docker.ContainerStorageHelper;
+import io.github.hectorvent.floci.services.s3.S3Service;
+import io.github.hectorvent.floci.services.s3.model.S3Object;
+import io.github.hectorvent.floci.services.sagemaker.model.SageMakerEntities.EndpointResource;
+import io.github.hectorvent.floci.services.sagemaker.model.SageMakerEntities.ModelResource;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
+import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
+import org.jboss.logging.Logger;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.net.ConnectException;
+import java.net.URI;
+import java.net.http.HttpTimeoutException;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+@ApplicationScoped
+public class SageMakerEndpointManager implements ContainerTeardown {
+    private static final Logger LOG = Logger.getLogger(SageMakerEndpointManager.class);
+    private static final int PORT = 8080;
+    private static final int DEFAULT_FILE_MODE = 0644;
+    private static final int PING_LOG_FREQUENCY = 15;
+    private static final Duration PING_TIMEOUT = Duration.ofSeconds(120);
+
+    private final ContainerBuilder containerBuilder;
+    private final ContainerLifecycleManager lifecycleManager;
+    private final EmulatorConfig config;
+    private final ContainerDetector containerDetector;
+    private final S3Service s3Service;
+    private final ExecutorService executor = Executors.newCachedThreadPool();
+    private final ConcurrentHashMap<String, String> containers = new ConcurrentHashMap<>();
+    private final HttpClient httpClient = HttpClient.newBuilder().connectTimeout(Duration.ofSeconds(2)).build();
+
+    @Inject
+    public SageMakerEndpointManager(ContainerBuilder containerBuilder, ContainerLifecycleManager lifecycleManager,
+                                    EmulatorConfig config, ContainerDetector containerDetector, S3Service s3Service) {
+        this.containerBuilder = containerBuilder;
+        this.lifecycleManager = lifecycleManager;
+        this.config = config;
+        this.containerDetector = containerDetector;
+        this.s3Service = s3Service;
+    }
+
+    public void startEndpointAsync(EndpointResource endpoint, SageMakerService service) {
+        executor.submit(() -> startEndpoint(endpoint, service));
+    }
+
+    private void startEndpoint(EndpointResource endpoint, SageMakerService service) {
+        String containerId = null;
+        try {
+            ModelResource model = service.modelForEndpoint(endpoint);
+            Map<String, Object> c = model.primaryContainer;
+            String image = string(c.get("Image"));
+            if (image == null) {
+                throw new IllegalArgumentException("PrimaryContainer.Image is required");
+            }
+            String name = ContainerStorageHelper.dockerName(config, "floci-sagemaker-endpoint-" + endpoint.endpointName);
+            lifecycleManager.removeIfExists(name);
+            ContainerBuilder.Builder builder = containerBuilder.newContainer(image)
+                    .withName(name)
+                    .withEnv(environment(endpoint.region, map(c.get("Environment"))))
+                    .withDockerNetwork(config.services().sagemaker().dockerNetwork())
+                    .withHostDockerInternalOnLinux()
+                    .withEmbeddedDns()
+                    .withPortBinding(PORT, 0)
+                    .withLogRotation();
+            applyEntrypoint(builder, c, "serve");
+            ContainerSpec spec = builder.build();
+            containerId = lifecycleManager.create(spec);
+            String modelDataUrl = string(c.get("ModelDataUrl"));
+            if (modelDataUrl != null && !modelDataUrl.isBlank()) {
+                copyModelData(containerId, modelDataUrl);
+            }
+            var info = lifecycleManager.startCreated(containerId, spec);
+            var ep = info.getEndpoint(PORT);
+            endpoint.containerId = containerId;
+            endpoint.invokeHost = ep.host();
+            endpoint.invokePort = ep.port();
+            containers.put(endpoint.endpointName, containerId);
+            waitForPing(endpoint.invokeHost, endpoint.invokePort);
+            endpoint.endpointStatus = "InService";
+            endpoint.failureReason = null;
+            service.updateEndpoint(endpoint);
+        } catch (Exception e) {
+            LOG.warnv("SageMaker endpoint {0} failed: {1}", endpoint.endpointName, e.getMessage());
+            if (containerId != null) {
+                lifecycleManager.stopAndRemove(containerId, null);
+            }
+            endpoint.endpointStatus = "Failed";
+            endpoint.failureReason = e.getMessage();
+            service.updateEndpoint(endpoint);
+        }
+    }
+
+    public void stopEndpoint(EndpointResource endpoint) {
+        String containerId = endpoint.containerId != null ? endpoint.containerId : containers.remove(endpoint.endpointName);
+        if (containerId != null) {
+            lifecycleManager.stopAndRemove(containerId, null);
+        }
+    }
+
+    @Override
+    public void stopManagedContainers() {
+        containers.forEach((name, id) -> lifecycleManager.stopAndRemove(id, null));
+        containers.clear();
+        executor.shutdownNow();
+    }
+
+    private List<String> environment(String region, Map<String, String> modelEnv) {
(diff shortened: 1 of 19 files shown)


src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerService.java
```diff
@@ -0,0 +1,539 @@
...
+            throw validation("PrimaryContainer or Containers is required");
+        }
+        ModelResource model = new ModelResource();
+        model.modelName = name;
+        model.modelArn = arn(region, "model/" + name);
+        model.executionRoleArn = text(request, "ExecutionRoleArn");
+        model.creationTime = nowMillis();
+        model.region = region;
+        model.accountId = regionResolver.getAccountId();
+        model.primaryContainer = primary.isObject() ? map(primary) : map(request.path("Containers").get(0));
```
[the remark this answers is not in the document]
mbergo: Are we aiming for an AWS Clone? :)

hectorvent:
Thank you for this, it is a substantial piece of work.

**(blocking)** Commits 855eecd, b74c5c3 carry an AI co-author trailer; this repo keeps commit attribution with human contributors (see AGENTS.md). Would you amend or squash those commits and force push so the attribution is yours alone?

**(blocking)** Before a detailed review, this needs a maintainer decision on direction, since a new Docker-backed service is an architectural commitment rather than an isolated addition. That is fairer to you than a line review of something that might not be wanted yet.

**(blocking)** Three checks are now red (commit message lint, action tables in sync, the python compat suite) and the branch shows conflicts with main.

Also worth adding `Closes #1945` to the description if that is the issue behind it.

Blocked on all three points above.

abanna (changes requested):
Thank you for the ambitious SageMaker implementation and for including unit, integration, Docker, SDK, configuration, storage, and documentation work. The focused suite passes 7/7. Several items still need attention:

- resolve the four current inline comments covering S3 URI validation, empty buckets, the `ModelError` response envelope, and formatting;
- namespace models, endpoint configs, endpoints, and training jobs by region rather than bare name; the current stores reject or expose same-named resources across regions even though their ARNs are regional;
- prevent `startEndpointAsync` from updating the store after `DeleteEndpoint`; currently deletion can return and remove the endpoint, then the startup worker can finish and reinsert it;
- rewrite the two `Potential fix for pull request finding` commits and remove their Copilot AI `Co-Authored-By` trailers;
- resolve the merge conflicts and the three failing CI checks.

SageMaker model ARNs include the region, and `DeleteEndpoint` is expected to free the deployed endpoint resources: [CreateModel](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateModel.html), [DeleteEndpoint](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_DeleteEndpoint.html). Thank you again—this is valuable work, but the isolation and lifecycle issues need regression coverage before it can merge safely.

Allaye:
This is a very good work, is this work stale?

KenkoGeek (commented):
This is a substantial first pass at SageMaker and the overall architecture is heading in the right direction: JSON 1.1 dispatch is wired through the shared controller, state goes through StorageFactory, the Docker-backed training/runtime paths are exercised, and there is SDK-level compatibility coverage. I rechecked the current head against the SageMaker API reference, AWS SDK for Java v2 shapes, Java 25, and the current AGENTS.md rules. There are still several merge-blocking compatibility issues on the current head.

1. **HIGH: `StopTrainingJob` can end up as `Failed` after the API has stored `Stopped`.** `stopTrainingJob()` removes the container and immediately persists `Stopped`, but the asynchronous runner is still executing `waitForExit()`. Once the container disappears, `getExitCodeIfStopped()` maps `NotFoundException` to exit code `1`, and the worker then persists `Failed`, overwriting the stop result. AWS transitions a requested stop through `Stopping` to `Stopped`; the stop itself must not be reclassified as a training failure. The runner needs an explicit cancellation/stop state that it can distinguish from a real non-zero container exit, with a regression that stops an in-progress job and proves the terminal state stays `Stopped`.

2. **HIGH: the advertised List operations do not implement the SageMaker request contract.** `ListModels` only applies `MaxResults` and always sorts by name, while `ListEndpointConfigs`, `ListEndpoints`, and `ListTrainingJobs` ignore `MaxResults` entirely. None of them emit/consume `NextToken`, and the modeled filters/sort controls are ignored. For example, AWS `ListTrainingJobs` supports `MaxResults` 1..100, `NextToken`, `NameContains`, creation/modified time filters, `StatusEquals`, `SortBy`, and `SortOrder`; `ListModels` likewise has pagination, filters, and sort controls. The Java SDK v2 exposes these members on the request models, so an SDK caller can send valid modeled requests that Floci silently handles with different semantics. Please either implement the supported operation contract or narrow/document the emulation explicitly rather than presenting these operations as compatible. Also validate `MaxResults`: the current `Stream.limit(max)` turns a negative value into an `IllegalArgumentException`/500 instead of an AWS-shaped validation error.

3. **MEDIUM: `CreateModel` accepts an empty `Containers` array and can throw a server-side NPE.** The guard checks only `isArray()`, so `{ "Containers": [] }` passes and then `get(0)` yields null before `map(...)`. AWS and the SDK model allow `PrimaryContainer` or `Containers` as alternatives, but an empty alternative cannot supply a container definition. This needs a ValidationException path and a wire-level negative test.

4. **MEDIUM: training channel staging flattens every S3 object to its basename.** `copyChannelObjects()` turns `prefix/a/data.csv` and `prefix/b/data.csv` into the same `/opt/ml/input/data/<channel>/data.csv`. That both loses folder structure and can overwrite distinct objects. SageMaker's File/FastFile channel contract exposes channel data below `/opt/ml/input/data/<channel>` and treats S3 prefixes as a folder hierarchy. Preserve the relative key beneath the selected S3 prefix and add a nested-key/collision regression.

5. **Repository/Java blockers:** the new production code contains `var` in both `SageMakerTrainingRunner` and `SageMakerEndpointManager`; current AGENTS.md requires explicit types. The branch is also currently conflicting with `main`, and the recorded head has red `Compatibility Tests`, `Conventional Commits`, and `Docs Action Tables` workflows. Those need to be rebased/fixed and rerun before approval.

There are a couple of smaller validation/runtime-shape concerns already called out by the existing bot review, but I would fix the lifecycle and List API contract first because they define observable SDK behavior. I do not recommend splitting the control-plane and Docker-backed execution into separate PRs at this point: they form one SageMaker service addition and the compatibility test spans both. The fixes above belong in this PR because they affect operations it explicitly claims to support.

References:
- https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_StopTrainingJob.html
- https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_ListTrainingJobs.html
- https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_ListModels.html
- https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateModel.html
- https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-training-algo-running-container.html
- https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/sagemaker/model/CreateModelRequest.html

CoreOxide:
@mbergo hey :)
thanks for this work! will you be handling pr comments, or should someone else take over?

KenkoGeek (commented):
Strong follow-up. I rechecked the current squashed head against the five blockers from my previous review. All five are addressed: an intentional StopTrainingJob can no longer be overwritten as Failed by the async runner; CreateModel rejects an empty Containers alternative; S3 training input keeps the path relative to the selected prefix with traversal protection; the production `var` declarations are gone; and the List operations now have bounded MaxResults/NextToken pagination while the exact supported filter subset is explicitly documented instead of being presented as the full AWS request surface. That last point satisfies the compatibility concern I raised because callers can now distinguish what Floci models from what SageMaker supports more broadly. I do not have another code blocker on this head. I am only holding approval because GitHub still reports the PR as UNSTABLE; once the required matrix settles fully green, the technical findings from my side are resolved.

src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerEndpointManager.java
```diff
@@ -0,0 +1,267 @@
...
+            }
+            Thread.sleep(1000);
+        }
+        throw new IllegalStateException("Container did not pass /ping health check");
+    }
+
+    @SuppressWarnings("unchecked")
+    static Map<String, String> map(Object value) {
+        if (value instanceof Map<?, ?> in) {
+            java.util.LinkedHashMap<String, String> out = new java.util.LinkedHashMap<>();
```
KenkoGeek: AGENTS.md asks us not to use inline fully-qualified class names unless there is a genuine collision. Please import `LinkedHashMap` and use the simple type here. There is no conflicting `LinkedHashMap` in this file.

src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerService.java
```diff
@@ -0,0 +1,630 @@
...
+    }
+
+    /**
+     * Applies AWS's {@code MaxResults}/{@code NextToken} contract to an already-filtered,
+     * name-sorted list: {@code MaxResults} (1-100, default {@value #DEFAULT_LIMIT}) caps the
+     * page, and an opaque {@code NextToken} (the last name returned) resumes strictly after it,
+     * matching the sort the caller applied.
+     */
+    private <T> ObjectNode paginate(List<T> sorted, JsonNode request, Function<T, String> nameOf,
+                                    java.util.function.BiConsumer<T, ObjectNode> render, String itemsField) {
```
KenkoGeek: Same repository-style issue here: please import `BiConsumer` and use the simple type instead of `java.util.function.BiConsumer`. There is no name collision requiring the FQCN.

src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerTrainingRunner.java
```diff
@@ -0,0 +1,360 @@
...
+    private final ContainerDetector containerDetector;
+    private final S3Service s3Service;
+    private final ObjectMapper mapper;
+    private final ExecutorService executor = Executors.newCachedThreadPool();
+    private final ConcurrentHashMap<String, String> containers = new ConcurrentHashMap<>();
+    // Names a stop that stop() has already committed to the store as "Stopped": run()'s exit-code
+    // handling checks this before persisting, since a stop races the removal of the very container
+    // waitForExit() is polling, and the container disappearing (NotFoundException, exit code 1)
+    // must not be relabeled a training failure.
+    private final java.util.Set<String> stopRequested = ConcurrentHashMap.newKeySet();
```
KenkoGeek: Please import `Set` and use the simple type here. `java.util.Set` inline violates the AGENTS.md FQCN rule and there is no collision in this source file.

src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerTrainingRunner.java
```diff
@@ -0,0 +1,360 @@
...
+        // "input.csv" itself, or a same-named sibling key equal to the prefix): fall back to
+        // the basename rather than a key that still carries the shared prefix.
+        return key.equals(prefix) || key.startsWith(prefix)
+                ? key.substring(key.lastIndexOf('/') + 1)
+                : key;
+    }
+
+    /** Rejects an absolute path or one that escapes the channel directory via "..". */
+    private static String safeRelativePath(String relative) {
+        java.nio.file.Path normalized = java.nio.file.Path.of(relative).normalize();
```
KenkoGeek: Please import `Path` and use `Path normalized = Path.of(relative).normalize();`. This is another inline FQCN without a collision, which AGENTS.md asks us to avoid.

KenkoGeek (changes requested):
Thanks for carrying the SageMaker fixes through the rewrite. I rechecked the current PR patch, and the five functional blockers from my earlier review remain resolved with a fully green/CLEAN branch. The only remaining blocker I found is repository Java style: the rewritten production sources contain four inline fully-qualified JDK types without any name collision. AGENTS.md requires imports and explicit simple types in this case.

src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerEndpointManager.java
pgermosen: Fixed, imported `LinkedHashMap` and dropped the inline FQCN.

src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerService.java
pgermosen: Fixed, imported `BiConsumer`.

src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerTrainingRunner.java
pgermosen: Fixed, imported `Set`.

src/main/java/io/github/hectorvent/floci/services/sagemaker/SageMakerTrainingRunner.java
pgermosen: Fixed, imported `Path` and switched to the simple name.

pgermosen:
Picking this up. On the direction question, hectorvent, I want to move forward with a Docker-backed SageMaker service, it's the same pattern we already run for Batch and CodeBuild, so it's not really a new architectural category.

Rebased onto main, fixed the region-scoping and the endpoint start/delete race the earlier reviews flagged (both hold up under a fresh look), and just cleared the four FQCN spots KenkoGeek caught, thanks for that. There was also a real bug in the python compat test, it hardcoded localhost instead of reading FLOCI_ENDPOINT like the rest of the suite, so it always failed inside its own CI container even though the service itself was fine. Everything's green now.

KenkoGeek (approved):
Thanks for the quick cleanup. I rechecked the current head: the four inline fully-qualified JDK types from my last review are now normal imports/simple types, the five earlier SageMaker compatibility findings remain resolved, and the full required matrix is green with CLEAN/MERGEABLE status. Approved on this head.

hectorvent (approved):
Thank you both, and thank you pgermosen for picking this up: the attribution is clean, CI is green on both native legs now that the compat test reads `FLOCI_ENDPOINT`, and every earlier finding (abanna's five, KenkoGeek's nine, the July bot comments) checks out as fixed on this head. On direction: agreed, a Docker-backed SageMaker on the Batch and CodeBuild pattern is the right shape, so that question is closed.

Three small things, none blocking:

**(follow-up, separate PR)** The README category table has no SageMaker row yet; the docs index has one, so it is a one-liner.

**(follow-up, separate PR)** A `sdk-test-java` fixture and test alongside the Python one, which the contributing checklist asks for on a new service.

**(follow-up, separate PR)** A state reset clears the stores but leaves running endpoint containers up; Batch and CodeBuild share that gap, so it is a shared follow-up.

No blockers from my side.

PR #1951, opened 2026-07-26
fix(rds): restore Terraform RDS and RabbitMQ compatibility

tuannx:
## Summary
- support the RDS DescribeDBInstances dbi-resource-id filter
- omit RabbitMQ users from DescribeBroker, matching AWS wire behavior
- preserve existing DBInstanceIdentifier and db-instance-id behavior
- add focused RDS unit tests and Amazon MQ integration coverage

## Why
A real Terraform acceptance flow against Floci 1.5.33 found two post-create refresh failures:

1. AWS provider 6.56 stores an RDS instance by DbiResourceId and refreshes it with Filters.Filter.1.Name=dbi-resource-id. Floci returned an empty DBInstances list after successful creation.
2. Floci returned RabbitMQ users from DescribeBroker. Terraform then called the standalone DescribeUser API, which correctly rejects RabbitMQ, so the broker refresh failed. AWS does not expose the RabbitMQ user list from DescribeBroker.

## Validation
- Amazon Corretto 25 ./mvnw -Dtest=RdsQueryHandlerTest,RdsServiceTest test: 104 tests passed
- Amazon Corretto 25 ./mvnw -Dtest=AmazonMqControllerIntegrationTest test: 5 tests passed

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/amazonmq/AmazonMqController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/amazonmq/AmazonMqController.java
@@ -90,7 +90,6 @@ public class AmazonMqController {
         body.put("autoMinorVersionUpgrade", b.isAutoMinorVersionUpgrade());
         body.put("created", b.getCreated());
         body.put("brokerInstances", b.getBrokerInstances());
-        body.put("users", b.getUsers());
         body.put("tags", b.getTags());
         return body;
     }
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsQueryHandler.java
@@ -144,7 +144,15 @@ public class RdsQueryHandler {
             filterId = extractRdsFilterValue(params, "db-instance-id");
         }
         try {
-            Collection<DbInstance> result = service.listDbInstances(filterId);
+            Collection<DbInstance> result;
+            if (filterId != null && !filterId.isBlank()) {
+                result = service.listDbInstances(filterId);
+            } else {
+                List<String> resourceIds = extractRdsFilterValues(params, "dbi-resource-id");
+                result = resourceIds.isEmpty()
+                        ? service.listDbInstances(null)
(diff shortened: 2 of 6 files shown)


pgermosen (approved):
Thanks for this — both fixes check out well against the AWS docs. `dbi-resource-id` is a real, documented [DescribeDBInstances](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html) filter, and the `users` omission matches the [Amazon MQ Broker reference](https://docs.aws.amazon.com/amazon-mq/latest/api-reference/brokers-broker-id.html) verbatim ("Does not apply to RabbitMQ brokers"), which is the only engine type this service supports today, so there's no reachable regression there.

@tuannx 
One thing worth flagging: this branch currently conflicts with `main` in `RdsQueryHandler.java`. It looks like `main` picked up a separate `DBInstanceNotFound` fault-check in the same `try` block this PR's `dbi-resource-id` fallback touches, so it'll need an actual merge of both behaviors rather than a trivial rebase. Approving since the change itself looks correct — just flagging so the conflict doesn't come as a surprise when it's time to merge.

tuannx:
Rebased onto the latest `main` and resolved the RDS conflict while preserving the new `DBInstanceNotFound` behavior for a direct `DBInstanceIdentifier`; filter forms still return an empty list when unmatched. I also updated `dbi-resource-id` handling to support all `Values.Value.N` entries with exact, case-sensitive matching as required by the RDS Filter API. Targeted RDS/Amazon MQ tests pass (149 tests, 0 failures). The local full run was blocked only by the unavailable Docker daemon in unrelated Lambda/WebSocket tests, so I am waiting for the GitHub CI and compatibility jobs as the authoritative full gate before marking the PR ready.

tuannx:
@pgermosen The rebased head triggered the CI, compatibility, conventional-commit, and docs workflows, but GitHub currently reports all four as `action_required` because they need maintainer approval for this fork. Could you please approve the workflow runs? I will keep the PR in draft until the checks have actually completed successfully.

PR #1953, opened 2026-07-26
fix(lambda): reassemble container log lines split across Docker frames

### LINKED ISSUE #1935: [BUG]  Lambda CloudWatch Logs: container stdout lines >1024 bytes split into multiple events
## Service
Lambda / CloudWatch Logs

## AWS API Action
`PutLogEvents` (internal — Floci Lambda container log collector streaming Docker `json-file` stdout to CloudWatch)
`GetLogEvents` (observed when reading `/aws/lambda/*` log groups)

## Expected behavior
A single stdout/stderr line from a container-image Lambda should be stored as **one** CloudWatch Logs event (up to AWS size limits, e.g. 256 KB per event), matching real AWS Lambda behavior.

When reading the log back via `GetLogEvents`, `message` should contain the full line:

```bash
aws logs get-log-events \
  --log-group-name /aws/lambda/my-function \
  --log-stream-name '2026/07/23/[$LATEST]abc123' \
  --endpoint-url http://localhost:4566 \
  --region us-east-1 \
  --output json | jq '[.events[].message | length]'
```

**Expected:** `[1478]` for a ~1.5 KB line, or `[6142]` for a ~6 KB line.

On the same Floci instance, ECS-style services that forward logs via `PutLogEvents` (e.g. Fluent Bit `cloudwatch_logs` output) preserve ~6 KB lines as a single intact event.

## Actual behavior

Floci splits a **single** stdout line into **multiple** CloudWatch events at **exactly 1024-byte boundaries**. Each fragment has a distinct `eventId`, usually the same `timestamp`, and the split occurs **mid-string** (e.g. inside a UUID or JSON field).

`docker logs <lambda-runtime-container>` shows the line intact; only CloudWatch (Floci) is fragmented.

```bash
aws logs get-log-events \
  --log-group-name /aws/lambda/my-function \
  --log-stream-name '2026/07/23/[$LATEST]abc123' \
  --endpoint-url http://localhost:4566 \
  --region us-east-1 \
  --output json | jq '[.events[].message | length]'
```

**Observed for ~1.5 KB line:** `[1024, 454]`

**Observed for ~6 KB line:** `[1024, 1024, 1024, 1024, 1024, 1024]`

`docker logs` (one intact line, ~1478 bytes):

```json
{"@timestamp":"2026-07-23T14:32:01.123Z","level":"INFO","logger":"com.example.EventProcessor","message":"Received SQS message","mdc":{"traceId":"a1b2c3d4-e5f6-7890-abcd-ef1234567890","spanId":"fedcba098765","tenantId":"tenant-acme-prod","environment":"dev-local","queueName":"events-for-provisioner.fifo","messageId":"7f3e2a1b-9c4d-4e8f-b2a1-6d5c9e8f7a3b","receiptHandle":"AQEBwJnKyrH...truncated...","approximateReceiveCount":"1","sentTimestamp":"1721745120000","bodyPreview":"{\"eventType\":\"CONFIGURATION_CHANGED\",\"resourceId\":\"cfg-00042\",\"payload\":{\"segment\":\"EMULATOR\",\"stage\":\"DEV\",\"region\":\"us-east-1\",\"accountAlias\":\"local-dev\",\"featureFlags\":{\"cacheEnabled\":true,\"asyncProcessing\":true},\"metadata\":{\"source\":\"configuration-api\",\"version\":\"1.30.0\",\"correlationId\":\"corr-9f8e7d6c-5b4a-3210-9876-543210fedcba\"}}}"}}
```

`get-log-events` response (two fragments):

```json
{
  "events": [
    {
      "timestamp": 1721745121123,
      "message": "{\"@timestamp\":\"2026-07-23T14:32:01.123Z\",\"level\":\"INFO\",\"logger\":\"com.example.EventProcessor\",\"message\":\"Received SQS message\",\"mdc\":{\"traceId\":\"a1b2c3d4-e5f6-7890-abcd-ef1234567890\",\"spanId\":\"fedcba098765\",\"tenantId\":\"tenant-acme-prod\",\"environment\":\"dev-local\",\"queueName\":\"events-for-provisioner.fifo\",\"messageId\":\"7f3e2a1b-9c4d-4e8f-b2a1-6d5c9e8f7a3b\",\"receiptHandle\":\"AQEBwJnKyrH...truncated...\",\"approximateReceiveCount\":\"1\",\"sentTimestamp\":\"1721745120000\",\"bodyPreview\":\"{\\\"eventType\\\":\\\"CONFIGURATION_CHANGED\\\",\\\"resourceId\\\":\\\"cfg-00042\\\",\\\"payload\\\":{\\\"segment\\\":\\\"EMULATOR\\\",\\\"stage\\\":\\\"DEV\\\",\\\"region\\\":\\\"us-east-1\\\",\\\"accountAlias\\\":\\\"local-dev\\\",\\\"featureFlags\\\":{\\\"cacheEnabled\\\":true,\\\"asyncProcessing\\\":true},\\\"metadata\\\":{\\\"source\\\":\\\"configuration-api\\\",\\\"version\\\":\\\"1.30.0\\\",\\\"correlationId\\\":\\\"corr-9f8e7d6c-5b4a-3210-9876-543210",
      "ingestionTime": 1721745121123,
      "eventId": "evt-000000000001"
    },
    {
      "timestamp": 1721745121123,
      "message": "fedcba\\\"}}}}}\"}}",
      "ingestionTime": 1721745121123,
      "eventId": "evt-000000000002"
    }
  ]
}
```

**Scope:** Lambda containers only. ECS-style services on the same Floci instance store ~6 KB lines intact:

## Reproduction

```bash
# 1. Start Floci
docker run -d --name floci \
  -p 4566:4566 \
  -e FLOCI_SERVICES_LAMBDA_EPHEMERAL=false \
  -e FLOCI_STORAGE_MODE=memory \
  floci/floci:1.5.33-compat

# 2. Create a container-image Lambda that logs a single-line JSON record > 1024 bytes to stdout
  --function-name my-function \
  --package-type Image \
  --code ImageUri=<your-image-uri> \
  --role arn:aws:iam::000000000000:role/lambda-role

# 3. Invoke the function
  --function-name my-function \
  --payload '{}' /tmp/out.json

# 4. Compare docker logs (intact) vs CloudWatch (split)
docker ps --filter "ancestor=<lambda-image>" --format '{{.ID}}' | head -1 | xargs docker logs

aws --endpoint-url http://localhost:4566 logs describe-log-streams \
  --log-group-name /aws/lambda/my-function

aws --endpoint-url http://localhost:4566 logs get-log-events \
  --log-group-name /aws/lambda/my-function \
  --log-stream-name '<stream-from-above>' \
  --output json | jq '[.events[].message | length]'
```

Lambda runtime containers use Docker `json-file` logging (see Floci PR #466):

```json
{"Type":"json-file","Config":{"max-file":"3","max-size":"10m"}}
```

## Environment

- Floci version / image tag: `floci/floci:1.5.33-compat` (also reproduced on `1.5.32-compat`)
- Java SDK version (if applicable): N/A (reproduced via AWS CLI; Lambda runtime is Java 17 container image)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (`docker compose`, custom image `FROM floci/floci:1.5.33-compat`)

| Setting | Value |
| --- | --- |
| Host OS | macOS (arm64), Lambda images run as `linux/amd64` |
| `FLOCI_SERVICES_LAMBDA_EPHEMERAL` | `false` |
| `FLOCI_STORAGE_MODE` | `memory` |
| `FLOCI_TLS_ENABLED` | `true` |

| Version | Lambda logs >1 KB | ECS logs ~6 KB (via `PutLogEvents`) |
| --- | --- | --- |
| `1.5.32-compat` | Split at 1024 bytes | Intact |
| `1.5.33-compat` | Split at 1024 bytes | Intact |

**Additional notes**

- Split is not newline-based — it occurs mid-JSON on a single physical line.
- Split is not Docker's ~16 KB partial-message behavior — fragments are exactly 1024 bytes.
- No `partial_message` / `partial_id` metadata on CloudWatch events for downstream reassembly.
- Upgrading from `1.5.32-compat` to `1.5.33-compat` did not resolve the issue.

aniketshukla1:
## Summary

Closes #1935

- Reassembles Docker stdout and stderr frames into newline-delimited CloudWatch Logs events, preserving independent buffers for each stream.
- Decodes complete lines as UTF-8 and force-splits valid unbounded output at the CloudWatch Logs event-size limit without cutting a multibyte character, including when continuation bytes arrive in a later frame.
- Serializes completed-line and trailing-tail delivery; terminal tails follow their Docker callback arrival order across stdout and stderr.
- Stops containers before releasing lifecycle log handles. A handle gives Docker a bounded drain period for queued frames and terminal delivery; when the transport stalls, a non-blocking fallback takes an exclusive callback boundary, closes the transport, and flushes the final tail without racing an in-flight callback.
- Tracks ECS log handles by Docker ID. A stopped, missing, or force-removed container finalizes and releases its own handle. If both teardown operations fail, the stopped task retains only that live reader and background reconciliation retries its teardown until it can safely close.
- Preserves a request buildspec override when retrying a build, so a source-less retry runs the same build definition as its original request.

## Validation

- `./mvnw test -Dtest=ContainerLogStreamerTest,ContainerLifecycleManagerVolumeTest,EcsContainerManagerLifecycleTest,EcsServiceTeardownTest,CodeBuildServiceTest`

The regressions cover UTF-8 characters completing exactly at the size cap across frames, stdout/stderr terminal-tail ordering, late frames during lifecycle drain, broken Docker transports without terminal callbacks, per-container ECS teardown outcomes and deferred teardown recovery, and source-less CodeBuild retries.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLifecycleManager.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLifecycleManager.java
@@ -164,23 +164,25 @@ public class ContainerLifecycleManager {
     public void stopAndRemove(String containerId, Closeable logStream) {
         LOG.infov("Stopping container {0}", containerId);
 
-        // Close log stream first
+        boolean stoppedOrMissing = false;
+        try {
+            dockerClient.stopContainerCmd(containerId).withTimeout(5).exec();
+            stoppedOrMissing = true;
+        } catch (NotFoundException e) {
+            LOG.debugv("Container {0} not found (already removed)", containerId);
+            stoppedOrMissing = true;
+        } catch (Exception e) {
+            LOG.warnv("Error stopping container {0}: {1}", containerId, e.getMessage());
+        }
+
         if (logStream != null) {
-            try {
-                logStream.close();
-            } catch (Exception e) {
-                LOG.debugv("Error closing log stream: {0}", e.getMessage());
+            if (stoppedOrMissing) {
+                closeLogStreamAfterContainerStop(logStream);
             }
         }
 
-        // Stop container
-        try {
-            dockerClient.stopContainerCmd(containerId).withTimeout(5).exec();
-        } catch (NotFoundException e) {
-            LOG.debugv("Container {0} not found (already removed)", containerId);
+        if (!stoppedOrMissing) {
             return;
-        } catch (Exception e) {
-            LOG.warnv("Error stopping container {0}: {1}", containerId, e.getMessage());
         }
 
         // Remove container
@@ -191,6 +193,18 @@ public class ContainerLifecycleManager {
             // Already gone
         } catch (Exception e) {
             LOG.warnv("Error removing container {0}: {1}", containerId, e.getMessage());
+        }
+    }
+
+    /**
+     * Releases lifecycle ownership of a log stream after its container has stopped. Docker's terminal
+     * callback normally flushes the final tail; a bounded fallback handles a broken transport.
+     */
+    public void closeLogStreamAfterContainerStop(Closeable logStream) {
+        try {
+            logStream.close();
+        } catch (Exception e) {
+            LOG.debugv("Error closing log stream: {0}", e.getMessage());
         }
     }
 
--- a/src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLifecycleManager.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLifecycleManager.java
@@ -175,24 +175,22 @@ public class ContainerLifecycleManager {
             LOG.warnv("Error stopping container {0}: {1}", containerId, e.getMessage());
         }
 
-        if (logStream != null) {
-            if (stoppedOrMissing) {
-                closeLogStreamAfterContainerStop(logStream);
-            }
-        }
-
-        if (!stoppedOrMissing) {
-            return;
-        }
-
-        // Remove container
+        // Force removal is the recovery path when Docker could not stop the container cleanly. It also
+        // terminates the follow-log transport, allowing its terminal callback to drain the final tail.
+        boolean removedOrMissing = false;
         try {
             dockerClient.removeContainerCmd(containerId).withForce(true).exec();
+            removedOrMissing = true;
             LOG.debugv("Removed container {0}", containerId);
         } catch (NotFoundException e) {
             // Already gone
+            removedOrMissing = true;
         } catch (Exception e) {
             LOG.warnv("Error removing container {0}: {1}", containerId, e.getMessage());
+        }
+
+        if (logStream != null && (stoppedOrMissing || removedOrMissing)) {
+            closeLogStreamAfterContainerStop(logStream);
         }
     }
 
--- a/src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
@@ -5,17 +5,32 @@ import io.github.hectorvent.floci.services.cloudwatch.logs.CloudWatchLogsService
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.async.ResultCallback;
 import com.github.dockerjava.api.model.Frame;
+import com.github.dockerjava.api.model.StreamType;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
+import java.io.ByteArrayOutputStream;
 import java.io.Closeable;
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.Deque;
+import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Consumer;
 
 /**
  * Streams Docker container logs to both the Floci console logger and CloudWatch Logs.
@@ -38,26 +53,39 @@ public class ContainerLogStreamer {
 
     /**
      * Attaches a log stream to a container and forwards logs to CloudWatch Logs.
-     * Returns a Closeable handle that should be closed when the container is stopped.
+     * Returns a lifecycle handle that should be closed after the container stops. Closing the handle
+     * allows Docker's reader a bounded drain period before a non-blocking fallback flushes the tail.
      *
      * @param containerId Docker container ID
      * @param logGroup CloudWatch log group name (e.g., "/aws/lambda/myFunction")
      * @param logStream CloudWatch log stream name (e.g., "2024/01/15/[$LATEST]abc123")
      * @param region AWS region for CloudWatch Logs
      * @param logPrefix prefix for console logging (e.g., "lambda:myFunction")
-     * @return Closeable handle to stop the log stream
+     * @return Closeable lifecycle handle for the log stream
      */
     public Closeable attach(String containerId, String logGroup, String logStream,
                             String region, String logPrefix) {
         ensureLogGroupAndStream(logGroup, logStream, region);
 
+        // Trim trailing whitespace and drop blank lines, then fan each reassembled line out to the
+        // console logger and CloudWatch Logs.
+        Consumer<String> emitter = line -> {
+            String trimmed = line.stripTrailing();
+            if (!trimmed.isEmpty()) {
+                LOG.infov("[{0}] {1}", logPrefix, trimmed);
+                forwardToCloudWatchLogs(logGroup, logStream, region, trimmed);
+            }
+        };
+
         try {
-            return dockerClient.logContainerCmd(containerId)
+            LogReassemblyCallback callback = new LogReassemblyCallback(emitter);
+            dockerClient.logContainerCmd(containerId)
                     .withStdOut(true)
                     .withStdErr(true)
                     .withFollowStream(true)
                     .withTimestamps(false)
-                    .exec(frameCallback(logGroup, logStream, region, logPrefix));
+                    .exec(callback);
+            return new ContainerLogHandle(callback);
         } catch (Exception e) {
             LOG.warnv("Could not attach log stream for container {0}: {1}", containerId, e.getMessage());
             return null;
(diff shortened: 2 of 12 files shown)


src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
```diff
@@ -118,4 +170,52 @@ private void forwardToCloudWatchLogs(String logGroup, String logStream, String r
...
+
+        /** Appends a frame payload and returns the complete lines it produced (possibly none). */
+        List<String> append(byte[] payload) {
+            List<String> lines = new ArrayList<>();
+            for (byte b : payload) {
+                if (b == '\n') {
+                    lines.add(takeLine());
+                } else {
+                    buffer.write(b);
+                    if (buffer.size() >= MAX_LINE_BYTES) {
```
pgermosen: **Forced split can corrupt a multibyte char at the cap boundary** (minor / edge).

The main path is multibyte-safe, but this `MAX_LINE_BYTES` force-split isn't: it lands on an arbitrary byte, so a 2–4 byte UTF-8 character straddling the cap gets decoded as replacement chars on both sides. It's the one place the class's own "a multibyte character straddling a frame boundary is not corrupted" guarantee (javadoc above) doesn't hold. Only bites on a single line ≥256 KB with no newline, so genuinely an edge case.

Cheapest fix: only force-split once the buffer is at the cap **and** the current byte is not a UTF-8 continuation byte (`10xxxxxx`), so the split falls on a character boundary:

```suggestion
                    if (buffer.size() >= MAX_LINE_BYTES && (b & 0xC0) != 0x80) {
```

(Lets the line overshoot the cap by up to 3 bytes — harmless. If you'd rather keep the hard cap, a one-line javadoc note that the forced split isn't character-safe would be enough.)

src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
```diff
@@ -58,12 +64,58 @@ public Closeable attach(String containerId, String logGroup, String logStream,
...
+                            List<String> lines = new ArrayList<>();
+                            synchronized (buffers) {
+                                buffers.values().forEach(buffer -> buffer.flush().ifPresent(lines::add));
+                            }
+                            lines.forEach(this::emit);
+                        }
+
+                        private void emit(String line) {
+                            String trimmed = line.stripTrailing();
+                            if (!trimmed.isEmpty()) {
```
pgermosen: **Blank lines are dropped here — and a new test may imply otherwise** (question / pre-existing).

`emit()` skips lines that are empty after `stripTrailing()`, so a genuinely blank log line never reaches CloudWatch. This is pre-existing behavior (the old code did the same), so not introduced by this PR — just flagging that the new `preservesEmptyLinesInSplitOutput` test asserts `append("a\n\nb\n") == ["a", "", "b"]`, which reads as "blank lines are preserved" when end-to-end they're still dropped here. The test's own comment acknowledges this, so it's fine; the test name could be a touch more precise (e.g. `bufferPreservesEmptyLines…`). Real Lambda does surface blank lines, so AWS-parity here would be a separate follow-up if it matters.

src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
```diff
@@ -118,4 +170,52 @@ private void forwardToCloudWatchLogs(String logGroup, String logStream, String r
...
+         * is exhausted — real AWS likewise splits an over-long line at the event-size limit.
+         */
+        private static final int MAX_LINE_BYTES = 256 * 1024;
+
+        private final ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+
+        /** Appends a frame payload and returns the complete lines it produced (possibly none). */
+        List<String> append(byte[] payload) {
+            List<String> lines = new ArrayList<>();
+            for (byte b : payload) {
```
pgermosen, line 414: **Byte-at-a-time append in a log hot path** (nit / perf).

`append` loops one byte at a time calling `ByteArrayOutputStream.write(int)` (itself a `synchronized` method) for every byte of every frame. For the common case you could scan the payload for `'\n'` and bulk-`write(byte[], off, len)` each segment — fewer synchronized calls and less per-byte overhead. Minor, and the current form is very readable; only worth it if log volume is a concern (CLAUDE.md: avoid expensive work in hot paths).

src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
```diff
@@ -58,12 +64,58 @@ public Closeable attach(String containerId, String logGroup, String logStream,
...
-                            String line = new String(frame.getPayload(), StandardCharsets.UTF_8).stripTrailing();
-                            if (!line.isEmpty()) {
-                                LOG.infov("[{0}] {1}", logPrefix, line);
-                                forwardToCloudWatchLogs(logGroup, logStream, region, line);
+                            List<String> lines;
+                            synchronized (buffers) {
+                                LogLineBuffer buffer = buffers.computeIfAbsent(frame.getStreamType(), t -> new LogLineBuffer());
+                                lines = buffer.append(frame.getPayload());
+                            }
+                            lines.forEach(this::emit);
```
pgermosen: **`emit()` runs outside the `buffers` lock — ordering under concurrent close** (nit / observation).

Emitting outside `synchronized (buffers)` is the right call — you don't want to hold the lock across a CloudWatch `putLogEvents`. One consequence: if `close()`/`flushBuffers()` runs on another thread while `onNext` is mid-`emit`, the trailing flushed line could interleave ahead of an already-returned line. In practice `onNext` is single-threaded per the docker-java callback and this only matters at teardown, so the risk is negligible — noting it only so it's a conscious choice.

pgermosen (approved):
Looks good, just a few optional comments.

aniketshukla1:
Addressed the remaining close-grace review finding in cf813bb.

`close()` now uses the grace solely to bound teardown waiting; it no longer flushes a partial buffer before the Docker reader reaches `onComplete` or `onError`. That keeps late in-flight continuations attached to their original per-stream logical line, and the terminal callback performs the one final tail flush.

Added regressions for an unterminated continuation arriving after grace expiry, including the no-early-emission assertion.

Validated with Java 25:
`./mvnw test -Dtest=ContainerLogStreamerTest`
(14 tests passing.)

aniketshukla1:
Follow-up for the no-terminal close path: 4a815a3 adds a resettable quiet-period fallback.

After the initial close wait expires, the callback schedules a final drain only after another grace interval with no frames. Each late `onNext` cancels and re-arms that drain, so a post-grace continuation stays joined to the existing partial line. If Docker never sends `onComplete` or `onError`, the quiet fallback still emits the trailing tail.

Validation with Java 25:
`./mvnw test -Dtest=ContainerLogStreamerTest`
(15 tests passing.)

aniketshukla1:
Addressed both remaining shutdown findings in fc8b660.

- Removed the quiet-timer terminal transition. Normal finalization remains on Docker `onComplete`/`onError`; the no-terminal fallback is now available only after `ContainerLifecycleManager` has stopped or confirmed the container missing.
- Moved container cleanup to stop the container before closing its log handle, including ECS task and CodeBuild cleanup paths.
- Added a serialized emission queue, so a concurrent final tail cannot overtake an already-completed line.

Validated with Java 25:
- `ContainerLogStreamerTest,ContainerLifecycleManagerVolumeTest`: 31 passed
- `EcsContainerManager*Test`: 19 passed

aniketshukla1:
Follow-up in 5e38b8b:

- If `stopContainer` fails, `ContainerLifecycleManager` now leaves the active log handle open and does not attempt removal.
- Removed the fixed terminal wait and forced flush entirely. After a successful stop, lifecycle finalization waits for Docker’s actual `onComplete`/`onError`, so delayed frames are accepted and flushed in order instead of being truncated at an arbitrary deadline.

New regressions cover both stop failure and a delayed frame delivered after lifecycle finalization begins.

Validated with Java 25: `ContainerLogStreamerTest,ContainerLifecycleManagerVolumeTest` (32 passing).

aniketshukla1:
Follow-up in 61ebfc0:

- lifecycle log-handle close is now non-blocking and retains the Docker reader until its terminal callback flushes any late tail;
- ECS retains task log handles whenever any container stop fails, and only releases them after every stop succeeds;
- added regression coverage for the lifecycle handle and both ECS stop outcomes.

Validated with 34 focused log/lifecycle tests and 21 ECS manager tests.

aniketshukla1:
Follow-up in 19f22f5:

- defer a forced split until a UTF-8 character boundary, including when continuation bytes arrive in a later Docker frame;
- sequence incomplete stdout/stderr tails by Docker callback arrival before the terminal flush;
- add direct regressions for both cases.

Validated with the focused log/lifecycle suite (36 tests).

aniketshukla1:
Follow-up in 896672b:

- lifecycle handle close is non-blocking and schedules a bounded Docker drain;
- a real Docker terminal cancels the fallback, while a stalled transport is explicitly closed and its final tail flushed;
- added a regression for an unterminated tail when Docker never invokes a terminal callback.

Validated with the focused log/lifecycle suite (37 tests).

aniketshukla1:
Follow-up in c571fc4:

- Docker callbacks take the read side of a fair gate;
- fallback finalization takes the write side before closing the transport and flushing, so callbacks already in flight finish first;
- lifecycle close remains non-blocking.

Validated with the focused log/lifecycle suite (37 tests).

aniketshukla1:
Thanks

src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLifecycleManager.java
```diff
@@ -164,25 +164,27 @@ public ContainerInfo startCreated(String containerId, ContainerSpec spec) {
...
 
+        if (logStream != null) {
+            if (stoppedOrMissing) {
+                closeLogStreamAfterContainerStop(logStream);
+            }
+        }
+
+        if (!stoppedOrMissing) {
+            return;
+        }
```
pgermosen: **Blocking — regression vs `main`: a container whose stop fails is now never removed.**

On `main`, a generic failure from `stopContainerCmd` logged a warning and **fell through to `removeContainerCmd(...).withForce(true)`**. Force-remove is precisely the recovery path for a container that will not stop. This early `return` makes it unreachable:

```java
} catch (Exception e) {
    LOG.warnv("Error stopping container {0}: {1}", containerId, e.getMessage());   // stoppedOrMissing stays false
}
...
if (!stoppedOrMissing) {
    return;              // removeContainerCmd(...).withForce(true) is now unreachable
}
```

A transient Docker hiccup during stop — daemon busy, timeout, socket blip — now leaves the container **running and never cleaned up** for the rest of the emulator's life. For an emulator that starts a container per Lambda invoke that is an accumulating leak, silent apart from one WARN. `ContainerLifecycleManagerVolumeTest.stopAndRemoveKeepsLogStreamOpenWhenContainerStopFails` asserts `verify(dockerClient, never()).removeContainerCmd(...)`, so it is codified as intended.

The goal — don't finalize a log transport that is still legitimately streaming from a running container — is reasonable. But it was achieved by dropping the removal, which is a separate concern. Separating them is one line:

```suggestion
        if (logStream != null && stoppedOrMissing) {
            closeLogStreamAfterContainerStop(logStream);
        }

        // Remove regardless: force-remove is the recovery path for a container that would not stop.
        // A stop failure means only that we must not finalize the log transport, not that we abandon
        // the container.
```

Removal terminates Docker's follow-stream anyway, so the callback still receives its terminal `onComplete` and flushes — the tail is not lost by removing.

Secondary, same hunk: `NotFoundException` now sets `stoppedOrMissing = true` and falls through to `removeContainerCmd`, where `main` returned immediately. Harmless (remove catches `NotFoundException`) but it is an unremarked extra Docker round-trip.

src/main/java/io/github/hectorvent/floci/services/ecs/container/EcsContainerManager.java
```diff
@@ -298,23 +293,22 @@ public Map<String, Integer> stopTaskAndCollectExitCodes(EcsTaskHandle handle) {
...
                 lifecycleManager.getDockerClient().stopContainerCmd(dockerId).withTimeout(5).exec();
             } catch (NotFoundException ignored) {
             } catch (Exception e) {
+                allContainersStopped = false;
                 LOG.warnv("Error stopping ECS container {0}: {1}", dockerId, e.getMessage());
             }
         }
 
+        if (allContainersStopped) {
+            handle.getLogStreams().forEach(lifecycleManager::closeLogStreamAfterContainerStop);
```
pgermosen: **Blocking — one failed container stop abandons every log stream in the task.**

`allContainersStopped` is a single flag for the whole task, so if *one* container of a multi-container task fails to stop, the log streams of **all** its siblings — which stopped cleanly — are skipped too.

It is also asymmetric with `stopAndRemove`: phase 2 below force-removes **every** container unconditionally, including the one that failed to stop. So we remove the containers and then decline to finalize their log handles. `EcsContainerManagerLifecycleTest.retainsAllTaskLogStreamsWhenAnyContainerStopFails` asserts exactly this, and "retains" describes a handle that nothing ever closes afterwards.

In practice Docker ends the follow-stream when the container is removed, so `onComplete` fires and the tail is flushed — which is what makes this mostly harmless, and also what makes it unnecessary. Simplest fix: close the streams after phase 2 removal, unconditionally, the way `cleanupStoppedTask` already does. If per-container gating is genuinely wanted, it needs to be per-container rather than one flag for the whole task.

src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
```diff
@@ -118,4 +137,312 @@ private void forwardToCloudWatchLogs(String logGroup, String logStream, String r
...
+     * (matching real AWS Lambda).
+     */
+    static final class LogReassemblyCallback extends ResultCallback.Adapter<Frame> {
+
+        private static final ScheduledExecutorService CLOSE_DRAIN_SCHEDULER =
+                Executors.newSingleThreadScheduledExecutor(runnable -> {
+                    Thread thread = new Thread(runnable, "floci-container-log-close");
+                    thread.setDaemon(true);
+                    return thread;
+                });
```
pgermosen, line 194: **Major — all container log finalization funnels through one shared daemon thread.**

Static, single-threaded, shared by every container in the emulator. The scheduled task is `closeTransportAndFlush()`, which takes the **write lock**, calls `super.close()` (blocking I/O on the Docker transport), then `flush()` — and `flush()` ends by calling `emitPendingLines()`, which invokes `lineConsumer` and therefore `cloudWatchLogsService.putLogEvents` **on that same single thread, still holding the write lock**.

So one slow or hung transport close, or one slow `putLogEvents`, head-of-line blocks tail finalization for every other container. Concurrent Lambda invokes or a multi-container ECS task hit this easily. At minimum a small fixed pool rather than a single thread; ideally the emit loop should also run outside the write lock.

Separately, and this is the broader point: AGENTS.md asks to keep changes narrow and to avoid refactors the task does not require. At `5418fc82` this was 186 lines in 2 files fixing a well-understood bug. It is now 821 lines across 7 files and has absorbed a public `closeLogStreamAfterContainerStop` on `ContainerLifecycleManager`, a rewrite of `stopAndRemove`'s control flow, a change to `CodeBuildRunner`'s `finally` block, and both `EcsContainerManager` teardown paths. None of that is needed to reassemble lines split across frames. Would you be open to splitting the lifecycle half into its own PR? The reassembly half is ready to merge today.

src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
```diff
@@ -118,4 +137,312 @@ private void forwardToCloudWatchLogs(String logGroup, String logStream, String r
...
+            synchronized (emissionLock) {
+                synchronized (buffers) {
+                    if (terminated || fallbackFinalization != null) {
+                        return;
+                    }
+                    fallbackFinalization = CLOSE_DRAIN_SCHEDULER.schedule(
+                            this::closeTransportAndFlush, closeGraceMillis, TimeUnit.MILLISECONDS);
+                }
+            }
+        }
```
pgermosen, line 290: **Behavior change worth documenting — `close()` is now asynchronous.**

`ContainerLogHandle.close()` now only schedules `closeTransportAndFlush` for `closeGraceMillis` (2 s) later and returns immediately; it no longer closes the transport. Defensible — real AWS log delivery is async too — but it is a contract change for every caller.

Concretely: `CodeBuildRunner` marks the build terminal in the same `finally` block that closes the handle, so a build's last log lines can now land up to two seconds *after* it reports `SUCCEEDED`. Any integration test that stops a container and immediately asserts on CloudWatch Logs content becomes racy.

Is that acceptable for the emulator's fidelity goals, and are there existing tests relying on the old synchronous behavior? The javadoc on `attach()` documents the grace period, which is good — but this deserves a line in the PR description too.

src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
```diff
@@ -118,4 +137,312 @@ private void forwardToCloudWatchLogs(String logGroup, String logStream, String r
...
+                }
+                buffer.write(b);
+                trackUtf8(b);
+                if ((buffer.size() >= MAX_LINE_BYTES && pendingContinuation == 0)
+                        || buffer.size() >= MAX_LINE_BYTES + 3) {
+                    // Do not split a valid character whose continuation bytes arrive in a later frame.
+                    // Malformed input may never reach a boundary, so allow at most the three continuation
+                    // bytes a valid UTF-8 character could need before forcing the buffer to drain.
+                    lines.add(takeLine());
+                }
```
pgermosen, line 436: **Nit — the `+ 3` arm looks unreachable.**

It is defensive against malformed input never reaching a character boundary, but the pre-split above already force-splits before any lead byte that would cross the cap, and `trackUtf8` resets `pendingContinuation` to 0 for both stray continuation bytes and invalid lead bytes — so `pendingContinuation` cannot stay non-zero for more than 3 bytes. I could not construct an input that reaches this branch.

Not harmful. But if it is genuinely unreachable, either drop it or add a test that proves otherwise, since the comment above asserts a scenario the code may not actually admit.

Worth saying plainly: the rest of this method is the best part of the PR. The boundary-safe pre-split plus `pendingContinuation` tracking fixes the cap-corruption issue I raised last week, and does it better than the fix I proposed — the event never exceeds 256 KB *and* never cuts a character.

src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLogStreamer.java
```diff
@@ -38,35 +53,39 @@ public ContainerLogStreamer(DockerClient dockerClient, CloudWatchLogsService clo
...
         ensureLogGroupAndStream(logGroup, logStream, region);
 
+        // Trim trailing whitespace and drop blank lines, then fan each reassembled line out to the
+        // console logger and CloudWatch Logs.
+        Consumer<String> emitter = line -> {
+            String trimmed = line.stripTrailing();
+            if (!trimmed.isEmpty()) {
+                LOG.infov("[{0}] {1}", logPrefix, trimmed);
+                forwardToCloudWatchLogs(logGroup, logStream, region, trimmed);
+            }
```
pgermosen, line 77: *Pre-existing, carry-over from my earlier review — not introduced here and correctly out of scope.*

`emitter` skips lines that are empty after `stripTrailing()`, so a genuinely blank log line never reaches CloudWatch, while real Lambda does surface them. Flagging only so it does not get lost across the rewrite — a follow-up issue would be the right home for it.

aniketshukla1:
@pgermosen Thanks for the review. Much appreciated. Will fix these Issues.

pgermosen:
@aniketshukla1 There are conflicts in this PR, could you solve them?

aniketshukla1:
> @aniketshukla1 There are conflicts in this PR, could you solve them?

Fixed them @pgermosen .

PR #1954, opened 2026-07-26
fix(s3): recognize virtual-hosted-style requests over HTTP/2

### LINKED ISSUE #1866: [BUG] S3 virtual-hosted-style URL misinterpreted as path-style over HTTP/2 (Host header vs :authority)
## Service

S3

## AWS API Action

Any S3 API called via virtual-hosted-style URL (e.g. `GetObject`, `PutObject`, presigned URL
access) over HTTP/2.

## Expected behavior

A virtual-hosted-style request such as `https://{bucket}.s3.{region}.localhost.floci.io:4566/{key}`
should be routed to the correct bucket/key regardless of whether the client negotiates HTTP/1.1
or HTTP/2, since both are valid ways for a browser to reach an HTTPS endpoint.

## Actual behavior

When the client uses HTTP/2 (which browsers select automatically via ALPN over HTTPS),
virtual-hosted-style requests are misinterpreted as path-style requests: the bucket name segment
in the Host is silently ignored, and the first path segment is treated as the bucket name
instead. The same request over HTTP/1.1 (same TLS, same URL) works correctly.

This was initially mistaken for an HTTP vs HTTPS difference ("works over HTTP, fails over
HTTPS"), but forcing HTTP/1.1 over HTTPS with `curl --http1.1` proved it succeeds — the actual
variable is the HTTP protocol version, not TLS.

Root cause (confirmed by reading the source): `S3VirtualHostFilter` determines whether to
rewrite a virtual-hosted-style request by calling `requestContext.getHeaderString("Host")`. If
this returns `null`, the filter does nothing and the request falls through to path-style
handling. HTTP/2 (RFC 9113) has no `Host` header — it uses the `:authority` pseudo-header
instead. Since `S3VirtualHostFilter` never reads `:authority`, it never fires for HTTP/2
requests, and the request is silently handled as path-style with the wrong bucket resolved from
the URL path.

Practical impact: any browser accessing a presigned URL or making a direct virtual-hosted-style
S3 request over HTTPS will use HTTP/2 by default and hit this bug. Only clients that explicitly
downgrade to HTTP/1.1 (e.g. most AWS SDKs, or `curl --http1.1`) are unaffected — this is why
server-to-server calls (e.g. via AWS SDK for Java) worked fine while browser-based access failed
consistently.

This is unrelated to #1625 (which only affects bucket-less `s3.` hosts) — this bug affects
virtual-hosted-style hosts that do include a bucket name, and has been present since
virtual-hosted-style support was added (reproduced on v1.5.15 through v1.5.31).

## Reproduction

```bash
# Same URL, same TLS/cert, only the negotiated HTTP version differs.
# Bucket "my-bucket" exists and contains key "nested/object.png".

# HTTP/1.1 (forced): correctly resolves bucket "my-bucket", key "nested/object.png" -> 200
curl -i --http1.1 --resolve my-bucket.s3.localhost.floci.io:4566:127.0.0.1 \
  "https://my-bucket.s3.localhost.floci.io:4566/nested/object.png"

# HTTP/2 (curl default over HTTPS via ALPN): bucket name from Host is ignored;
# "nested" is misinterpreted as the bucket name and "object.png" as the key -> 404 NoSuchBucket
curl -i --resolve my-bucket.s3.localhost.floci.io:4566:127.0.0.1 \
  "https://my-bucket.s3.localhost.floci.io:4566/nested/object.png"
```

## Environment

- Floci version / image tag: v1.5.31 (`floci/floci:latest` as of 2026-07-07); also reproduced on
  v1.5.15, v1.5.27, v1.5.28
- Java SDK version (if applicable): N/A for the failing case (browser `fetch`/`` requests
  via presigned URL). AWS SDK for Java is unaffected because it uses HTTP/1.1.
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (`docker compose`)

aniketshukla1:
Browsers negotiate HTTP/2 via ALPN over HTTPS, and HTTP/2 (RFC 9113) has no Host header — the authority is carried in the :authority pseudo-header. S3VirtualHostFilter read only getHeaderString("Host"), which is null over HTTP/2, so virtual-hosted-style requests fell through to path-style handling: the bucket label in the authority was ignored and the first path segment was treated as the bucket instead.

Resolve the effective authority from the request URI (populated from :authority) when the Host header is absent, so bucket routing works on both HTTP/1.1 and HTTP/2.

Closes #1866

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -489,7 +489,7 @@ public class S3Controller {
             }
 
             // --- Website Hosting Redirection Logic ---
-            if (isWebsiteRequest(httpHeaders) && (uriInfo.getQueryParameters().isEmpty() || (uriInfo.getQueryParameters().size() == 1 && hasQueryParam(uriInfo, "list-type")))) {
+            if (isWebsiteRequest(httpHeaders, uriInfo) && (uriInfo.getQueryParameters().isEmpty() || (uriInfo.getQueryParameters().size() == 1 && hasQueryParam(uriInfo, "list-type")))) {
                 try {
                     WebsiteConfiguration webConfig = s3Service.getBucketWebsite(bucket);
                     if (webConfig.getIndexDocument() != null) {
@@ -806,7 +806,7 @@ public class S3Controller {
             return fullObjectResponse(bucket, key, versionId, obj, overrides, includeChecksum);
         } catch (AwsException e) {
             emitCloudTrailEvent("GetObject", bucket, key, 0L, 0L, e.getErrorCode(), e.getMessage());
-            if (isWebsiteErrorDocumentTrigger(e) && isWebsiteRequest(httpHeaders)) {
+            if (isWebsiteErrorDocumentTrigger(e) && isWebsiteRequest(httpHeaders, uriInfo)) {
                 try {
                     WebsiteConfiguration webConfig = s3Service.getBucketWebsite(bucket);
                     Response r = serveErrorDocument(bucket, webConfig, authorization, e.getHttpStatus());
@@ -2183,8 +2183,11 @@ public class S3Controller {
         }
     }
 
-    private static boolean isWebsiteRequest(HttpHeaders httpHeaders) {
-        String host = httpHeaders.getHeaderString("Host");
+    static boolean isWebsiteRequest(HttpHeaders httpHeaders, UriInfo uriInfo) {
+        // HTTP/2 (RFC 9113) carries no Host header, so fall back to the request URI authority —
+        // the same resolution S3VirtualHostFilter applies. Without this, a website bucket reached
+        // over HTTP/2 would be served as an API (XML) response instead of website HTML.
+        String host = S3VirtualHostFilter.resolveHost(httpHeaders.getHeaderString("Host"), uriInfo.getRequestUri());
         return host != null && host.contains("s3-website");
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/s3/S3ControllerWebsiteRequestTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/s3/S3ControllerWebsiteRequestTest.java
@@ -1,0 +1,56 @@
+package io.github.hectorvent.floci.services.s3;
+
+import jakarta.ws.rs.core.HttpHeaders;
+import jakarta.ws.rs.core.UriInfo;
+import org.junit.jupiter.api.Test;
+
+import java.net.URI;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class S3ControllerWebsiteRequestTest {
+
+    private static HttpHeaders headers(String hostHeader) {
+        HttpHeaders h = mock(HttpHeaders.class);
+        when(h.getHeaderString("Host")).thenReturn(hostHeader);
+        return h;
+    }
+
+    private static UriInfo uri(String requestUri) {
+        UriInfo u = mock(UriInfo.class);
+        when(u.getRequestUri()).thenReturn(URI.create(requestUri));
+        return u;
+    }
+
+    @Test
+    void detectsWebsiteFromHostHeaderOverHttp1() {
+        assertTrue(S3Controller.isWebsiteRequest(
+                headers("my-bucket.s3-website-us-east-1.amazonaws.com"),
+                uri("http://my-bucket.s3-website-us-east-1.amazonaws.com/index.html")));
+    }
+
+    @Test
+    void detectsWebsiteFromUriAuthorityWhenHostHeaderAbsentOverHttp2() {
+        // Regression for the HTTP/2 review finding: no Host header, authority carried by the URI.
+        assertTrue(S3Controller.isWebsiteRequest(
+                headers(null),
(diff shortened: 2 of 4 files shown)


pgermosen (approved):
Looks good.

Correct, minimal fix. HTTP/2 (RFC 9113) carries the authority in the `:authority` pseudo-header and has no `Host` header, so `getHeaderString("Host")` is null and virtual-hosted requests were falling through to path-style. Falling back to the request URI authority when `Host` is absent is the right resolution, and hoisting the single `getRequestUri()` call to the top of `filter()` (reused for both host resolution and the later path rewrite) is a clean touch.

Things I verified:
- **HTTP/1.1 behavior is unchanged** — `resolveHost` returns the `Host` header verbatim whenever it's present; the URI-authority fallback only fires when `Host` is null (covered by `resolveHostPrefersHostHeaderOverUriAuthority`).
- **Null-safe** — `resolveHost(null, null)` and a relative URI both yield null, so `filter()` still early-returns.
- **`getAuthority()` shape matches the `Host` header** (`host[:port]`), and `extractBucket` strips the port anyway.

**Two things worth flagging (both non-blocking):**
@hectorvent 

1. **Merge conflict with #1843.** Both PRs modify `filter()` and `extractBucket` in this same file — #1843 changes `extractBucket` to a 3-arg signature and rewrites the `filter()` body, while this PR keeps the 2-arg call and hoists the URI. Whoever merges second will need to reconcile them; the fix here (`resolveHost` + the hoisted `uri`) is independent of #1843's suffix work, so it's a mechanical rebase, not a design clash. Just calling it out since both are in flight.

2. **The new HTTP/2 branch has no end-to-end coverage — and can't, with the current harness.** Virtual-host routing itself *is* integration-tested over HTTP/1.1 (`S3VirtualHostIntegrationTest`, `S3VirtualHostTest`, and the compat `S3VirtualHostStyleTest`), but those run under RestAssured/`@QuarkusTest`, which always send a `Host` header — so they never reach the `Host == null` → URI-authority fallback this PR adds. I checked for any HTTP/2 test capability (`http2`/`ALPN`/`h2c`) across `src/test` and `compatibility-tests`; the only hit is this PR's new unit-test file. So the fix's real-world assumption (the container exposes `:authority` as `getRequestUri().getAuthority()`) is verified only by the synthetic-`URI` unit tests. That's a reasonable limitation rather than something to fix — there's no ALPN/h2c harness here to drive a real HTTP/2 request — but worth a sentence in the PR description noting the assumption.

aniketshukla1:
Thanks

src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
```diff
@@ -2183,8 +2183,11 @@ private void validateChecksumHeaders(HttpHeaders httpHeaders, byte[] data, Strin
...
     }
 
-    private static boolean isWebsiteRequest(HttpHeaders httpHeaders) {
-        String host = httpHeaders.getHeaderString("Host");
+    static boolean isWebsiteRequest(HttpHeaders httpHeaders, UriInfo uriInfo) {
+        // HTTP/2 (RFC 9113) carries no Host header, so fall back to the request URI authority —
+        // the same resolution S3VirtualHostFilter applies. Without this, a website bucket reached
+        // over HTTP/2 would be served as an API (XML) response instead of website HTML.
+        String host = S3VirtualHostFilter.resolveHost(httpHeaders.getHeaderString("Host"), uriInfo.getRequestUri());
         return host != null && host.contains("s3-website");
```
pgermosen, line 2191: **Pre-existing, non-blocking — flagged only because this line is being touched.**

`host.contains("s3-website")` is a substring test over the whole authority, so a bucket *named* `s3-website-assets`, reached at `s3-website-assets.localhost:4566`, is classified as a website request and served its index/error document — where real S3 would return `ListBucketResult` XML from the REST endpoint (the website endpoint is a separate hostname there).

This predates the PR and the change doesn't widen it: HTTP/2 now simply behaves the way HTTP/1.1 always did. Matching the authority's second label (`s3-website*` as the label after the bucket) rather than any substring would close it, but that's its own change.

src/test/java/io/github/hectorvent/floci/services/s3/S3ControllerWebsiteRequestTest.java
```diff
@@ -0,0 +1,56 @@
...
+                headers("my-bucket.s3-website-us-east-1.amazonaws.com"),
+                uri("http://my-bucket.s3-website-us-east-1.amazonaws.com/index.html")));
+    }
+
+    @Test
+    void detectsWebsiteFromUriAuthorityWhenHostHeaderAbsentOverHttp2() {
+        // Regression for the HTTP/2 review finding: no Host header, authority carried by the URI.
+        assertTrue(S3Controller.isWebsiteRequest(
+                headers(null),
+                uri("https://my-bucket.s3-website-us-east-1.localhost:4566/index.html")));
```
pgermosen, line 40: **Nit.** These HTTP/2 fixtures pass the *pre*-rewrite URI (`/index.html`), but at both production call sites `isWebsiteRequest` runs after `S3VirtualHostFilter` has rewritten the path to `/my-bucket/index.html`.

Only the authority is read, so the test is correct today. Using the rewritten form would make the fixture describe the request the controller actually receives — and keeps it honest if this ever starts consulting the path.

pgermosen (approved):
Looks good, a couple of nit comments.

aniketshukla1:
@pgermosen Thanks for the review, Much appreciated.

aniketshukla1:
Thanks

hectorvent (approved):
Thanks @aniketshukla1 @pgermosen 

Merging

PR #1956, opened 2026-07-26
feat(cloudformation): expand SAM AWS::Serverless::HttpApi to ApiGatew…

### LINKED ISSUE #1759: [BUG] SAM transform doesn't expand `AWS::Serverless::HttpApi` — reports CREATE_COMPLETE with a synthesized physical ID, no API actually created
## Service

CloudFormation (SAM Transform) / API Gateway v2

## AWS API Action

`CreateStack`/`UpdateStack` on a template with `Transform: AWS::Serverless-2016-10-31` containing an `AWS::Serverless::HttpApi` resource

## Expected behavior

SAM's `AWS::Serverless::HttpApi` should expand to a real `AWS::ApiGatewayV2::Api` (plus routes, integrations, stage, and any configured authorizer), the same way `AWS::Serverless::Function` and `AWS::Serverless::Api` already expand in this codebase.

```bash
aws apigatewayv2 get-apis
```

should show the API the template declares.

## Actual behavior

The stack reports `CREATE_COMPLETE`, including for the `AWS::Serverless::HttpApi` logical resource itself (with a synthesized physical ID), but `apigatewayv2 get-apis` shows nothing was created. The sibling `AWS::Serverless::Function` and its IAM role are created correctly — only the API never materializes, so the deployed Lambda has no route to it at all.

Root cause (source inspection): `SamTransformProcessor.expandSamTemplate`'s switch on SAM resource types only handles `AWS::Serverless::Function`, `AWS::Serverless::SimpleTable`, and `AWS::Serverless::Api` (REST, v1). `AWS::Serverless::HttpApi` isn't a case, so it falls into the `default -> LOG.debugv(...)` branch and is dropped from the expanded template — along with any `Auth.Authorizers`/`DefaultAuthorizer` config and any `HttpApi`-typed `Events` on a sibling function.

## Reproduction

```bash
docker rm -f floci-sam-httpapi-poc 2>/dev/null || true

docker run --name floci-sam-httpapi-poc \
  -p 4566:4566 \
  -e AWS_ACCESS_KEY_ID=test \
  -d floci/floci:latest

```

```bash
cat > /tmp/floci-sam-httpapi-template.yaml <<'EOF'
Resources:
  MyFunction:
    Properties:
      PackageType: Zip
      Handler: index.handler
      Runtime: nodejs20.x
      InlineCode: |
        exports.handler = async () => ({ statusCode: 200, body: "hello" });
      Events:
        HelloEvent:
          Type: HttpApi
          Properties:
            ApiId: !Ref MyHttpApi
            Path: /hello
            Method: get

  MyHttpApi:
    Type: AWS::Serverless::HttpApi
    Properties:
      Auth:
        Authorizers:
          MyAuthorizer:
            JwtConfiguration:
              issuer: https://cognito-idp.us-east-1.amazonaws.com/us-east-1_xxxxx
              audience: [my-client-id]
        DefaultAuthorizer: MyAuthorizer
EOF

  --template-file /tmp/floci-sam-httpapi-template.yaml \
  --stack-name test-sam-httpapi-stack \
  --capabilities CAPABILITY_IAM

echo "--- list-stack-resources (note MyHttpApi reports CREATE_COMPLETE) ---"
aws --endpoint-url http://localhost:4566 cloudformation list-stack-resources \
  --stack-name test-sam-httpapi-stack

echo "--- get-apis (expected: the declared API, actual: empty) ---"
aws --endpoint-url http://localhost:4566 apigatewayv2 get-apis
```

Observed output:

```
--- list-stack-resources (note MyHttpApi reports CREATE_COMPLETE) ---
{
    "StackResourceSummaries": [
        {
            "LogicalResourceId": "MyHttpApi",
            "PhysicalResourceId": "MyHttpApi-006ff7ca",
            "ResourceType": "AWS::Serverless::HttpApi",
            "ResourceStatus": "CREATE_COMPLETE"
        },
        {
            "LogicalResourceId": "MyFunctionRole",
            "PhysicalResourceId": "test-sam-httpapi-stack-MyFunctionRole-613f3ef5247f",
            "ResourceType": "AWS::IAM::Role",
            "ResourceStatus": "CREATE_COMPLETE"
        },
        {
            "LogicalResourceId": "MyFunction",
            "PhysicalResourceId": "test-sam-httpapi-stack-MyFunction-d9c9f6107784",
            "ResourceType": "AWS::Lambda::Function",
            "ResourceStatus": "CREATE_COMPLETE"
        }
    ]
}
--- get-apis (expected: the declared API, actual: empty) ---
{
    "Items": []
}
```

```bash
docker rm -f floci-sam-httpapi-poc
```

## Additional notes

- Related to #1758 — a separate gap in the raw-CFN resource provisioner rather than the SAM transform. A SAM-based (`AWS::Serverless::HttpApi`) template hits this issue first (the API itself never gets created); fixing that one alone wouldn't help until this transform gap is also fixed.
- `expandServerlessApi` (REST v1) is the existing pattern to follow for an `expandServerlessHttpApi` — same file, same general shape: synthesize `Api`/`Route`/`Integration`/`Stage`/`Authorizer` resources into the expanded template, reusing the implicit-API route collection this file already does for `AWS::Serverless::Function`'s `Api` events.

- [x] Yes
- [ ] No

## Environment

aniketshukla1:
## Summary

Closes #1759.

- Expand `AWS::Serverless::HttpApi` into an HTTP `AWS::ApiGatewayV2::Api` and an auto-deploying `$default` stage.
- Map inline `DefinitionBody` to the generated API `Body`; map S3-backed `DefinitionUri` to `BodyS3Location`, preserving intrinsic `Ref` and `Fn::Sub` values until provisioning resolves them.
- Materialize JSON or YAML OpenAPI operations as API Gateway V2 routes and integrations.
- Merge a Function `HttpApi` event into a matching `DefinitionBody` path and method by adding the Lambda proxy integration and effective SAM authorization to that operation, preserving its OpenAPI fields and emitting one route instead of a conflicting standalone route.
- Reconcile only body-generated routes, integrations, and authorizers on stack updates, restoring the prior snapshot after failed replacement and retaining ownership of resources that survive failed cleanup.
- Reject unsupported OpenAPI security combinations without silently weakening authorization, with route-specific diagnostics.
- Return AWS-compatible `ConflictException` / HTTP 409 for duplicate route keys.
- Allocate a collision-safe stage logical ID so a user resource named `<HttpApiId>Stage` is preserved alongside the generated stage.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

SAM OpenAPI definitions remain effective whether supplied inline or through S3. Function events that overlap an inline OpenAPI operation now follow SAM's merge model: the operation keeps its documentation, receives the Lambda proxy integration and configured default JWT authorizer/scopes, honors a per-event `Authorizer: NONE` opt-out, and produces exactly one deployed route. HTTP methods under `paths` become real API Gateway V2 routes and are reconciled on `UpdateStack`; replacement, definition removal, and cleanup failure retain rollback ownership. This PR intentionally generates the `$default` auto-deploy stage; honoring SAM `StageName` remains a separate pre-existing gap.

## Checklist

- [ ] `./mvnw test` passes locally (full suite delegated to CI)
- [x] New or updated integration tests added

Focused Java 25 verification:

`./mvnw -q '-Dtest=SamTransformProcessorTest,ApiGatewayV2CfnProvisionerTest,ApiGatewayV2IntegrationTest,SamTransformIntegrationTest#samHttpApi_matchingDefinitionBodyAndFunctionEventCreateOneIntegratedRoute' test`

Results: 123 tests, 0 failures/errors.

Post-review authorization regression verification:

`./mvnw '-Dtest=SamTransformProcessorTest,SamTransformIntegrationTest#samHttpApi_matchingDefinitionBodyAndFunctionEventCreateOneIntegratedRoute' test`

Results: 35 tests, 0 failures/errors.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
@@ -403,16 +403,32 @@ public class ApiGatewayV2Service {
         routeStore.delete(routeKey(region, apiId, routeId));
     }
 
-    /** Restores a route with its original ID for a higher-level transactional rollback. */
+    /** Restores a route without treating any conflicting route as rollback-owned. */
     public void restoreRoute(String region, String apiId, Route route) {
+        restoreRoute(region, apiId, route, java.util.Set.of());
+    }
+
+    /**
+     * Restores a route with its original ID for a higher-level transactional rollback.
+     * Only conflicts explicitly identified as part of the failed replacement may be removed;
+     * independently managed routes must never be deleted as a rollback side effect.
+     */
+    public void restoreRoute(String region, String apiId, Route route,
+                             java.util.Collection<String> replaceableRouteIds) {
         getApi(region, apiId);
         // A failed cleanup can leave a replacement route behind. Remove that conflicting
-        // route directly from storage so restoring the snapshot cannot leave two active
-        // routes with the same route key.
+        // route only when the caller proves it belongs to that failed replacement.
         for (Route existing : getRoutes(region, apiId)) {
             if (!java.util.Objects.equals(existing.getRouteId(), route.getRouteId())
                     && java.util.Objects.equals(existing.getRouteKey(), route.getRouteKey())) {
-                routeStore.delete(routeKey(region, apiId, existing.getRouteId()));
+                if (replaceableRouteIds.contains(existing.getRouteId())) {
+                    routeStore.delete(routeKey(region, apiId, existing.getRouteId()));
+                } else {
+                    throw new AwsException("ConflictException",
+                            "Cannot restore route with key '" + route.getRouteKey()
+                                    + "' because an independently managed route already exists",
+                            409);
+                }
             }
         }
         Route restored = new Route();
--- a/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
@@ -385,6 +385,19 @@ public class ApiGatewayV2Service {
         routeStore.delete(routeKey(region, apiId, routeId));
     }
 
+    /** Restores a route with its original ID for a higher-level transactional rollback. */
+    public void restoreRoute(String region, String apiId, Route route) {
+        getApi(region, apiId);
+        Route restored = new Route();
+        restored.setRouteId(route.getRouteId());
+        restored.setRouteKey(route.getRouteKey());
+        restored.setAuthorizationType(route.getAuthorizationType());
+        restored.setAuthorizerId(route.getAuthorizerId());
+        restored.setTarget(route.getTarget());
+        restored.setRouteResponseSelectionExpression(route.getRouteResponseSelectionExpression());
+        routeStore.put(routeKey(region, apiId, restored.getRouteId()), restored);
+    }
+
     public Route updateRoute(String region, String apiId, String routeId, Map<String, Object> request) {
         Route route = getRoute(region, apiId, routeId);
 
@@ -543,6 +556,13 @@ public class ApiGatewayV2Service {
     public void deleteIntegration(String region, String apiId, String integrationId) {
         getIntegration(region, apiId, integrationId);
         integrationStore.delete(integrationKey(region, apiId, integrationId));
+    }
+
+    /** Restores an integration with its original ID for a higher-level transactional rollback. */
+    public void restoreIntegration(String region, String apiId, Integration integration) {
+        getApi(region, apiId);
+        integrationStore.put(integrationKey(region, apiId, integration.getIntegrationId()),
+                new Integration(integration));
     }
 
     public Integration updateIntegration(String region, String apiId, String integrationId,
--- a/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
@@ -356,9 +356,11 @@ public class ApiGatewayV2Service {
 
     public Route createRoute(String region, String apiId, Map<String, Object> request) {
         getApi(region, apiId);
+        String requestedRouteKey = (String) request.get("routeKey");
+        ensureRouteKeyAvailable(region, apiId, requestedRouteKey, null);
         Route route = new Route();
         route.setRouteId(shortId(8));
-        route.setRouteKey((String) request.get("routeKey"));
+        route.setRouteKey(requestedRouteKey);
         route.setAuthorizationType((String) request.getOrDefault("authorizationType", "NONE"));
         route.setAuthorizerId((String) request.get("authorizerId"));
         route.setAuthorizationScopes(toStringList(request.get("authorizationScopes")));
@@ -388,11 +390,21 @@ public class ApiGatewayV2Service {
     /** Restores a route with its original ID for a higher-level transactional rollback. */
     public void restoreRoute(String region, String apiId, Route route) {
         getApi(region, apiId);
+        // A failed cleanup can leave a replacement route behind. Remove that conflicting
+        // route directly from storage so restoring the snapshot cannot leave two active
+        // routes with the same route key.
+        for (Route existing : getRoutes(region, apiId)) {
+            if (!java.util.Objects.equals(existing.getRouteId(), route.getRouteId())
+                    && java.util.Objects.equals(existing.getRouteKey(), route.getRouteKey())) {
+                routeStore.delete(routeKey(region, apiId, existing.getRouteId()));
+            }
+        }
         Route restored = new Route();
         restored.setRouteId(route.getRouteId());
         restored.setRouteKey(route.getRouteKey());
         restored.setAuthorizationType(route.getAuthorizationType());
         restored.setAuthorizerId(route.getAuthorizerId());
+        restored.setAuthorizationScopes(route.getAuthorizationScopes());
         restored.setTarget(route.getTarget());
         restored.setRouteResponseSelectionExpression(route.getRouteResponseSelectionExpression());
         routeStore.put(routeKey(region, apiId, restored.getRouteId()), restored);
@@ -402,7 +414,9 @@ public class ApiGatewayV2Service {
         Route route = getRoute(region, apiId, routeId);
 
         if (request.containsKey("routeKey") && request.get("routeKey") != null) {
-            route.setRouteKey((String) request.get("routeKey"));
+            String requestedRouteKey = (String) request.get("routeKey");
+            ensureRouteKeyAvailable(region, apiId, requestedRouteKey, routeId);
+            route.setRouteKey(requestedRouteKey);
         }
         if (request.containsKey("authorizationType") && request.get("authorizationType") != null) {
             route.setAuthorizationType((String) request.get("authorizationType"));
@@ -422,6 +436,20 @@ public class ApiGatewayV2Service {
 
         routeStore.put(routeKey(region, apiId, routeId), route);
         return route;
+    }
+
+    private void ensureRouteKeyAvailable(String region, String apiId, String requestedRouteKey,
+                                         String currentRouteId) {
+        if (requestedRouteKey == null) {
+            return;
+        }
+        boolean duplicate = getRoutes(region, apiId).stream()
+                .anyMatch(existing -> !java.util.Objects.equals(existing.getRouteId(), currentRouteId)
+                        && requestedRouteKey.equals(existing.getRouteKey()));
+        if (duplicate) {
+            throw new AwsException("BadRequestException",
+                    "Route with key '" + requestedRouteKey + "' already exists", 400);
+        }
     }
 
     // JSON bodies can carry non-string elements (numbers, booleans); downstream code
--- a/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
@@ -479,8 +479,8 @@ public class ApiGatewayV2Service {
                 .anyMatch(existing -> !java.util.Objects.equals(existing.getRouteId(), currentRouteId)
                         && requestedRouteKey.equals(existing.getRouteKey()));
         if (duplicate) {
-            throw new AwsException("BadRequestException",
-                    "Route with key '" + requestedRouteKey + "' already exists", 400);
+            throw new AwsException("ConflictException",
+                    "Route with key '" + requestedRouteKey + "' already exists", 409);
         }
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
@@ -307,6 +307,22 @@ public class ApiGatewayV2Service {
         authorizerStore.delete(authorizerKey(region, apiId, authorizerId));
     }
 
+    /** Restores an authorizer with its original ID for a higher-level transactional rollback. */
+    public void restoreAuthorizer(String region, String apiId, Authorizer authorizer) {
+        getApi(region, apiId);
+        Authorizer restored = new Authorizer();
+        restored.setAuthorizerId(authorizer.getAuthorizerId());
+        restored.setAuthorizerType(authorizer.getAuthorizerType());
+        restored.setName(authorizer.getName());
+        restored.setJwtConfiguration(authorizer.getJwtConfiguration());
+        restored.setIdentitySource(authorizer.getIdentitySource());
+        restored.setAuthorizerUri(authorizer.getAuthorizerUri());
+        restored.setAuthorizerPayloadFormatVersion(authorizer.getAuthorizerPayloadFormatVersion());
+        restored.setAuthorizerResultTtlInSeconds(authorizer.getAuthorizerResultTtlInSeconds());
+        restored.setEnableSimpleResponses(authorizer.getEnableSimpleResponses());
+        authorizerStore.put(authorizerKey(region, apiId, restored.getAuthorizerId()), restored);
+    }
+
     public Authorizer updateAuthorizer(String region, String apiId, String authorizerId,
                                        Map<String, Object> request) {
         Authorizer auth = getAuthorizer(region, apiId, authorizerId);
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -137,6 +137,7 @@ public class CloudFormationResourceProvisioner {
     private static final String LAMBDA_NAME_MODE_ATTR = "FlociLambdaFunctionNameMode";
     private static final String LAMBDA_PACKAGE_TYPE_ATTR = "FlociLambdaPackageType";
     static final String UPDATE_ROLLBACK_RESTORED_ATTR = "__FlociUpdateRollbackRestored";
+    static final String UPDATE_ROLLBACK_FAILURE_ATTR = "__FlociUpdateRollbackFailure";
     private static final String INLINE_CLEANUP_POLICY_NAME_ATTR = "__FlociInlineCleanupPolicyName";
     private static final String INLINE_CLEANUP_ROLE_TARGETS_ATTR = "__FlociInlineCleanupRoleTargets";
     private static final String INLINE_CLEANUP_USER_TARGETS_ATTR = "__FlociInlineCleanupUserTargets";
@@ -5637,6 +5638,10 @@ public class CloudFormationResourceProvisioner {
             }
         } catch (RuntimeException restoreFailure) {
             failure.addSuppressed(restoreFailure);
+            String reason = restoreFailure.getMessage() != null
+                    ? restoreFailure.getMessage()
+                    : restoreFailure.getClass().getSimpleName();
+            r.getAttributes().put(UPDATE_ROLLBACK_FAILURE_ATTR, reason);
         }
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -5208,7 +5208,7 @@ public class CloudFormationResourceProvisioner {
                 previous = snapshotApiGatewayV2BodyResources(r, region, apiId);
                 deleteApiGatewayV2BodyResources(r, region, apiId);
             } catch (RuntimeException e) {
-                rollbackApiGatewayV2BodyReplacement(region, apiId,
+                rollbackApiGatewayV2BodyReplacement(r, region, apiId,
                         new ApiGatewayV2BodyResources(List.of(), List.of()), previous, e);
                 throw e;
             }
@@ -5217,13 +5217,19 @@ public class CloudFormationResourceProvisioner {
 
         // Keep the currently-serving body resources in place until the replacement has been
         // fully created. A failed replacement therefore cannot leave the API with no routes.
-        ApiGatewayV2BodyResources replacement = materializeApiGatewayV2BodyRoutes(region, apiId, body);
+        ApiGatewayV2BodyResources replacement;
+        try {
+            replacement = materializeApiGatewayV2BodyRoutes(region, apiId, body);
+        } catch (ApiGatewayV2BodyMaterializationException e) {
+            retainApiGatewayV2BodyResourceIds(r, e.resources());
+            throw e;
+        }
         ApiGatewayV2BodyResourceState previous = null;
         try {
             previous = snapshotApiGatewayV2BodyResources(r, region, apiId);
             deleteApiGatewayV2BodyResources(r, region, apiId);
         } catch (RuntimeException e) {
-            rollbackApiGatewayV2BodyReplacement(region, apiId, replacement, previous, e);
+            rollbackApiGatewayV2BodyReplacement(r, region, apiId, replacement, previous, e);
             throw e;
         }
         storeApiGatewayV2BodyResourceIds(r, APIGATEWAY_V2_BODY_ROUTE_IDS_ATTR, replacement.routeIds());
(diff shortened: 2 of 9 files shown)


aniketshukla1:
Addressed the Greptile blockers in 2585430.

- `AWS::ApiGatewayV2::Api.Body` now resolves OpenAPI `paths` into concrete HTTP routes during CloudFormation provisioning, with `x-amazon-apigateway-integration` targets when declared.
- A `<HttpApiId>Stage` collision now keeps the user resource and synthesizes the auto-deploy stage under the next collision-safe logical ID.
- Added an end-to-end SAM stack test that lists the deployed V2 routes, plus transform coverage proving both collision resources remain present.

Verified with JDK 25: `./mvnw -q -Dtest=SamTransformProcessorTest,SamTransformIntegrationTest test` (38 tests, 0 failures/errors).

aniketshukla1:
Follow-up for the latest Greptile report: `4e51050` resolves both remaining blockers.

- `DefinitionUri` now becomes `BodyS3Location`; provisioning fetches and parses S3-backed JSON or YAML OpenAPI documents before materializing their routes.
- Body-created routes and integrations are tracked on the CloudFormation resource and reconciled during `UpdateStack`, so removed routes are deleted and added routes are created without touching explicitly declared V2 resources.

Regression coverage now verifies an S3-backed `DefinitionUri` route and a create/update transition from `GET /before` to `GET /after`. Focused Java 25 verification: 41 tests, 0 failures/errors.

aniketshukla1:
Final follow-up for the replacement-safety review: `c60c2e8` stages the complete generated route/integration replacement before deleting the prior body-owned resources. If creation fails, every staged ID is cleaned up and the old route set plus its tracked ownership metadata remain intact for retry.

Added a hermetic regression that creates the first replacement route, forces the second to fail, and asserts that the partial route is deleted while the existing route is never deleted. Focused Java 25 verification now covers 42 tests with 0 failures/errors.

aniketshukla1:
Final rollback follow-up in `491d8b1`: before deleting the old body-owned set, reconciliation snapshots every tracked route and integration. If any old-resource deletion fails, it cleans the staged replacement and restores the snapshot with the original IDs (integrations first, then routes), leaving both the serving state and CloudFormation ownership metadata intact.

The rollback regression now covers both failure boundaries: partial replacement creation and a mid-delete failure after one old route was removed. Focused Java 25 verification: 43 tests, 0 failures/errors.

aniketshukla1:
Final intrinsic-definition follow-up in `10ac116`: `DefinitionUri` intrinsic objects are preserved through the SAM transform rather than being split or discarded early. Provisioning resolves the intrinsic, accepts the resulting `s3://bucket/key` value, and loads its OpenAPI document before route materialization.

Regression coverage now checks transform preservation of `Fn::Sub` and deploys an S3-backed HttpApi definition through `Fn::Sub` end to end. Focused Java 25 verification: 44 tests, 0 failures/errors.

aniketshukla1:
Final definition-removal follow-up in `4f8705d`: removing `Body` or `BodyS3Location` now uses the same snapshot-and-rollback transaction as a body replacement. A failed cleanup restores the previous body-generated routes and integrations with their original IDs instead of leaving a partially removed API.

Added a regression that removes the definition, fails on the second old-route deletion, and verifies both original route IDs are restored. Focused Java 25 verification: 45 tests, 0 failures/errors.

aniketshukla1:
Final failed-cleanup ownership follow-up in `c786132`: partial replacement cleanup now attempts every generated route and integration. If any cleanup call still fails, the residual IDs are retained on the CloudFormation resource alongside existing generated IDs, so a later update or stack deletion can reconcile them rather than leaving orphaned API Gateway V2 state.

Added a regression that creates a partial replacement, forces its cleanup deletion to fail, and verifies the surviving route ID is retained. Focused Java 25 verification: 46 tests, 0 failures/errors.

aniketshukla1:
Thanks

aniketshukla1:
Thanks

hectorvent:
Thank you for this, and I am sorry you have been waiting on an answer since July. The expansion itself checks out: `Body` and `BodyS3Location` are both real write-only properties of `AWS::ApiGatewayV2::Api`, and honouring `StageName` matches [SAM's own generator](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/api/http_api_generator.py), which defaults to `$default` and takes an override.

**(blocking)** #2146 also closes #1759 by expanding `AWS::Serverless::HttpApi`, and the two conflict in `SamTransformProcessor`. They are complementary rather than duplicates: yours covers the `DefinitionBody` and `DefinitionUri` route path, #2146 covers Function `HttpApi` events and JWT authorizers. My plan is to land #2146 first because it is the smaller change, then ask this one to rebase and keep the definition half. That call was mine to make and I owed it to you two weeks ago, so I am sorry for the wasted cycles.

**(follow-up, separate PR)** The 340 lines added to `CloudFormationResourceProvisioner` run against AGENTS.md, which asks for per service provisioners under `provisioners/`.

aniketshukla1:
Rebased this PR onto current `main` after #2146 merged. The overlapping SAM HttpApi expansion (function-event routes, authorizers, and base API/stage synthesis) now comes from `main`; this branch retains the complementary definition-based behavior requested in review:

- maps `DefinitionBody` to ApiGatewayV2 `Body`;
- maps literal, object, and intrinsic `DefinitionUri` values to `BodyS3Location`;
- materializes OpenAPI routes/integrations and reconciles them on update;
- stages replacements and preserves rollback-safe cleanup on partial failures/removal.

I also aligned the standalone provisioner test fixture with the current constructor dependencies. Focused validation passed: 32 transform tests, 4 ApiGatewayV2 lifecycle/rollback tests, and 3 request-level HttpApi DefinitionBody/DefinitionUri/update integration tests (39 total).

The larger provisioner-refactor suggestion remains intentionally separate, as discussed.

aniketshukla1:
CI follow-up: the only failed check was unrelated to this branch — `NeptuneOpenCypherIntegrationTest.openCypherQueryWorksOverBolt` timed out reading its Bolt socket after the suite had run 11,563 tests with zero assertion failures. This PR does not modify Neptune, and all native/compatibility jobs passed. GitHub would not allow a failed-job rerun without repository admin rights, so I pushed an empty conventional commit (`5c0a2f4`) to trigger a clean retry without changing code.

aniketshukla1:
Addressed the Greptile 4/5 rollback concern in d4bb1b4.

The fix now:
- snapshots and removes only tracked body-generated API Gateway V2 resources before replacement, so old and replacement route keys cannot overlap;
- restores the prior snapshot and its tracked IDs after deletion/materialization failures;
- rejects duplicate route keys on create/update;
- removes a surviving conflicting replacement before restoring an old route, while preserving all route fields including authorization scopes.

Verification under JDK 25:
- ApiGatewayV2IntegrationTest + ApiGatewayV2CfnProvisionerTest: 43/43 passed
- relevant SAM HttpApi DefinitionBody, DefinitionUri, and stack-update methods: 3/3 passed
- adjacent API Gateway V2 REST/JSON update tests also passed.

One broader SAM suite invocation failed only because the JDK container did not have a nested Docker socket for an unrelated Lambda invocation test; the targeted SAM HttpApi methods passed cleanly.

aniketshukla1:
Followed up on the updated Greptile ownership concern in e8c4f5a.

CloudFormation restores the previous StackResource object after a failed update. The provisioner now additively merges only API Gateway V2 generated route/integration cleanup IDs from the failed attempt into that last-known-good metadata before it is persisted. Normal attempted attributes still cannot overwrite committed resource state.

Added an end-to-end regression that creates a SAM HttpApi, injects a route-materialization failure plus persistent cleanup failure, waits for UPDATE_ROLLBACK_COMPLETE, and verifies both the original and surviving replacement IDs remain owned by the persisted resource.

JDK 25 verification: the new integration test plus six provisioner tests passed 7/7.

aniketshukla1:
Addressed the remaining Greptile security concern in `54461d5`:

- OpenAPI `components.securitySchemes` now materialize body-owned HTTP API JWT/REQUEST authorizers.
- Operation-level or inherited `security` is mapped to `AuthorizationType`, `AuthorizerId`, and JWT scopes; `security: []` remains an explicit public override.
- Protected operations with an unresolved or currently unenforced scheme fail closed instead of becoming `NONE` routes.
- Body-generated authorizers are tracked, cleaned up, retained after failed cleanup, and restored transactionally with routes/integrations.

JDK 25 verification: the focused API Gateway V2, provisioner, rollback, SAM processor, and three relevant SAM integration methods pass (86 tests across the focused suites, 0 failures/errors). A broader 93-test invocation had one unrelated nested-Docker Lambda launch failure because the Docker socket was not mounted; all HTTP API tests in that run passed.

aniketshukla1:
Addressed the follow-up multi-scheme concern in `85b2166`:

- OpenAPI security arrays are validated as OR alternatives before route selection.
- A requirement object containing multiple schemes (AND semantics) is rejected because an HTTP API V2 route can enforce only one authorizer; it is no longer weakened to the first scheme.
- An empty requirement anywhere in the OR-list is honored as anonymous access.

Focused JDK 25 verification: `ApiGatewayV2CfnProvisionerTest` — 13 tests, 0 failures/errors.

aniketshukla1:
Addressed both follow-up concerns in `2b8cfb8`:

- Rollback restoration now receives the exact failed-replacement route IDs. It may remove only those owned conflicts; a same-key independently managed route is preserved and causes a conflict instead of being deleted.
- Multiple protected OpenAPI OR alternatives are rejected as unrepresentable by a single HTTP API V2 route authorizer, instead of selecting one and rejecting callers valid under another. Empty anonymous alternatives remain supported.

JDK 25 verification: 55 affected tests pass (14 provisioner + 40 API Gateway V2 integration + 1 end-to-end CloudFormation failed-cleanup rollback), 0 failures/errors.

aniketshukla1:
Addressed the incomplete-restoration status concern in `263f25e`:

- Eager body-resource rollback now records restoration failures instead of only suppressing them.
- `CloudFormationService` carries that failure onto the committed resource and the rollback walker reports `UPDATE_ROLLBACK_FAILED`; it no longer emits `UPDATE_ROLLBACK_COMPLETE` with missing route state.
- Added a stack-level regression that introduces an independent same-key route during a failed update, verifies the independent route is preserved, the original route remains absent, and the stack reaches `UPDATE_ROLLBACK_FAILED`.

JDK 25 verification: 56 affected tests pass (14 provisioner + 40 API Gateway V2 integration + 2 stack-level cleanup/rollback), 0 failures/errors.

aniketshukla1:
Final verification on `263f25e`:

- Greptile: **5/5** — “The PR appears safe to merge” / “No blocking failure remains.”
- `Build and Test`: passed in 19m45s.
- Native image and all native SDK/compatibility jobs: passed.
- Conventional commits, action-table sync, and AWS CLI endpoint compatibility: passed.
- Merge state: CLEAN.

All follow-up commits list only `Aniket Shukla <0aniketshukla1@gmail.com>` as both author and committer.

pgermosen:
Thanks for the rebase, and for carrying this through eight rounds of bot feedback without losing the thread. The core of it holds up, and I want to be specific about what I checked rather than leave you guessing.

**What I verified works.** I merged your branch onto current `main` (clean, no conflicts) and ran the five affected suites: 102 tests, 0 failures. Then I built the jar and drove a real emulator with the AWS CLI:

- A `DefinitionBody`-only `AWS::Serverless::HttpApi` produces exactly the routes the OpenAPI document declares, with `Target` wired to the generated integration when `x-amazon-apigateway-integration` is present.
- `UpdateStack` reconciles correctly: removing `POST /items` and adding `PUT /new` left exactly `GET /items` + `PUT /new`, and the orphaned integration was cleaned up rather than left behind. That is the part I most expected to be wrong, and it is right.
- `DeleteStack` cascades through `deleteApi`'s prefix sweep — no orphaned routes.

**The OR/AND security rejection is not just defensible, it is documented.** I went looking to challenge it and found the opposite. From [Use OpenAPI definitions for HTTP APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-open-api.html), under the **Error** validation category:

> while OpenAPI allows users to define an API with multiple security requirements attached to a particular operation, API Gateway does not support this. Each operation can have only one of IAM authorization, a Lambda authorizer, or a JWT authorizer. **Attempting to model multiple security requirements results in an error.**

So rejecting both the AND form (multiple schemes in one requirement object) and the OR form (multiple alternatives) matches AWS exactly, error-on-import included. Worth putting that citation in a comment next to `applyOpenApiRouteSecurity` so nobody "fixes" it back to picking the first scheme.

---

## (blocking) A `DefinitionBody` route and a Function `HttpApi` event on the same path now fail the whole stack

This is the canonical SAM pattern — define the API shape in `DefinitionBody`, attach the handler with an event — and it regresses from working to `ROLLBACK_COMPLETE`.

```yaml
Resources:
  MyApi:
    Type: AWS::Serverless::HttpApi
    Properties:
      DefinitionBody:
        openapi: 3.0.1
        info: {title: collide, version: "1.0"}
        paths:
          /items:
            get:
              responses: {"200": {description: ok}}
  MyFn:
    Properties:
      Runtime: python3.12
      Handler: index.handler
      InlineCode: "def handler(e,c): return {}"
      Events:
        Api:
          Type: HttpApi
          Properties:
            ApiId: !Ref MyApi
            Path: /items
            Method: GET
```

Built from your branch merged onto `main`:

```
$ aws cloudformation describe-stacks --stack-name collide \
    --query 'Stacks[0].[StackStatus,StackStatusReason]' --output text
ROLLBACK_COMPLETE   Route with key 'GET /items' already exists
```

The same template on `main` at `768e96a3`, same probe, same command:

```
CREATE_COMPLETE   None
$ aws apigatewayv2 get-routes --api-id 81c2f551c4 --query 'Items[].[RouteId,RouteKey]' --output text
88475f6b   GET /items
```

One route, stack green. `main` reaches the correct end state only because it never materialized body routes at all — but it does reach it, and after this PR the stack does not deploy.

**Root cause is in the transform, not the provisioner.** `expandServerlessHttpApi` now sets `Body` from `DefinitionBody` (SamTransformProcessor.java:218), and then, unconditionally, loops `allRoutes` and emits a separate `AWS::ApiGatewayV2::Route` for every Function event bound to this API (SamTransformProcessor.java:258-263). Nothing tells the second producer that the first already declared `GET /items`. Two code paths, one route key, and `ensureRouteKeyAvailable` — correctly — refuses the second.

**Note that `ensureRouteKeyAvailable` is right and should stay.** For a hand-written `AWS::ApiGatewayV2::Api` with a `Body` plus an explicitly declared `AWS::ApiGatewayV2::Route` on the same key, real API Gateway rejects it too. The guard is not the bug; it is what exposed the bug.

**What SAM actually does.** It never emits standalone `Route` resources for HTTP APIs — function events are merged *into* the definition body. `OpenApiEditor.add_lambda_integration` ([open_api.py:136](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/open_api/open_api.py)) short-circuits on an existing integration and otherwise calls `add_path`, which is idempotent by construction:

```python
def add_path(self, path: str, method: str | None = None) -> None:
    ...
    path_dict = self.paths.setdefault(path, Py27Dict())
    for path_item in self.get_conditional_contents(path_dict):
        path_item.setdefault(method, Py27Dict())
```

`setdefault` — an event whose path+method the body already declares contributes nothing new. One route out the other end.

The narrow fix that matches those semantics: in the `allRoutes` loop, skip `expandHttpApiRoute` when the resolved `DefinitionBody` already declares that path and method. That leaves the event-only case (no `DefinitionBody`) exactly as #2146 built it, and makes the overlap case a no-op instead of a collision.

Please add a transform test for the overlap and a SAM integration test asserting the stack reaches `CREATE_COMPLETE` with exactly one route — that combination has no coverage today, which is why eight review rounds and a green CI run did not catch it.

---

## (should fix) Duplicate route keys should be `ConflictException`, not `BadRequestException`

`ensureRouteKeyAvailable` (ApiGatewayV2Service.java:473-485) raises:

```
$ aws apigatewayv2 create-route --api-id 663f127974 --route-key "GET /items"

An error occurred (BadRequestException) when calling the CreateRoute operation:
Route with key 'GET /items' already exists
```

`CreateRoute` declares four errors in the service model — `NotFoundException`, `TooManyRequestsException`, `BadRequestException`, `ConflictException` — and the model documents them as:

| shape | status | documentation |
|---|---|---|
| `ConflictException` | 409 | "The requested operation would cause a conflict with the current state of a service resource associated with the request." |
| `BadRequestException` | 400 | "The request is not valid, for example, the input is incomplete or incorrect." |

A duplicate route key is a conflict with existing resource state, not a malformed request — and `ConflictException` has no other candidate condition on `CreateRoute` to explain its presence in the model. Your own `restoreRoute` already uses `ConflictException`/409 for the identical condition (ApiGatewayV2Service.java:425-431), so the two paths disagree with each other today. A caller retrying on 409 and failing fast on 400 gets the wrong answer from `CreateRoute`.

---

## (question) Fail-closed on a security scheme without `x-amazon-apigateway-authorizer`

An ordinary OpenAPI document that documents bearer auth without wiring an API Gateway authorizer now takes the whole stack down:

```yaml
components:
  securitySchemes:
    bearerAuth: {type: http, scheme: bearer, bearerFormat: JWT}
paths:
  /secure:
    get:
      security: [{bearerAuth: []}]
```

```
ROLLBACK_COMPLETE   Protected operation references no supported security scheme
```

I understand the reasoning — silently downgrading a route marked protected to `AuthorizationType: NONE` is the worse failure, and I would rather have this than that. But the same docs page puts "valid OpenAPI, unsupported for HTTP APIs" in the **Info** category, where API Gateway ignores the property and still creates the API, and a documentation-only security scheme is a common thing to find in a real definition file. I could not settle from the docs which category AWS files this under, so I am not calling it a finding — but please say which behaviour you intended, and either way make the message name the offending path, method, and scheme name. As it stands the user gets one sentence with no pointer to which of forty operations is at fault.

---

## Smaller notes, none blocking

- **The failure-injection machinery is heavier than the emulator can trigger.** Snapshot/restore/retain across routes, integrations, and authorizers, plus `UPDATE_ROLLBACK_FAILURE_ATTR` threading through `CloudFormationService` — every branch of it is reachable only through Mockito fault injection, because the backing stores do not fail. The one exception is real: `restoreRoute` colliding with an independently managed route, which is exactly what your last commit added a stack-level test for. I am not asking you to remove any of it, but it is worth a sentence in the class javadoc saying which branches are defensive so a future reader does not go hunting for the production trigger.
- **`StageName` is dropped for `AWS::Serverless::HttpApi`.** Your PR description says a configured stage name is preserved, but `expandServerlessHttpApi` hardcodes `$default` (SamTransformProcessor.java:253) — I passed `StageName: prod` and got no `prod` stage. That is pre-existing from #2146, not yours (the REST `AWS::Serverless::Api` path at line 952 does honour it), so I will file it separately. Just worth correcting the description so it does not read as a claim this PR makes good on.
- **Retained IDs can outlive their resource.** In `rollbackApiGatewayV2BodyReplacement`, `retainApiGatewayV2BodyResourceIds` runs before the restore loop, and `restoreRoute` may then delete a retained conflicting replacement route. The ID stays tracked as owned. Harmless — every delete path is 404-tolerant — but the tracking is briefly a lie.

---

Everything except the first item is a note or a question. The route collision is the one that has to move before this can go in: it turns the standard SAM HttpApi template into a rollback, and it is the exact combination this feature exists to serve.

aniketshukla1:
Addressed the new review in `afba025`.

- A Function `HttpApi` event whose path and method already exist in inline `DefinitionBody` is now merged into that OpenAPI operation. The existing operation fields are preserved, the Lambda `aws_proxy` integration (`POST`, payload 2.0, Lambda URI) is added when absent, and the Lambda permission is still emitted.
- No standalone `AWS::ApiGatewayV2::Route` or `Integration` is emitted for the overlap, so CloudFormation sees one producer for the route key. The stack-level regression reaches `CREATE_COMPLETE` with exactly one `GET /items` route and one integration targeting the generated Lambda.
- Duplicate route keys now return AWS-compatible `ConflictException` / HTTP 409 consistently from create/update and rollback restoration paths.
- The intentional fail-closed security behavior is retained; unsupported schemes now report the exact method, path, and scheme name. I also added the AWS HTTP API import-validation citation beside the multi-requirement rejection so it is not later weakened back to first-alternative selection.
- The PR description no longer claims `StageName` support; it explicitly records that as a separate pre-existing gap.

Focused Java 25 result: 123 tests, 0 failures/errors across the transform unit suite, API Gateway v2 provisioning/service suites, and the new stack-level overlap test. Full CI is running on the pushed head.

aniketshukla1:
Follow-up to the Greptile 4/5 security finding: `26aafc4` preserves the effective SAM authorization when a Function `HttpApi` event is merged into an existing `DefinitionBody` operation.

- A configured `DefaultAuthorizer` is emitted in the AWS SAM-compatible OpenAPI JWT security-scheme shape and attached to the merged operation, including configured authorization scopes.
- Existing non-empty operation security remains authoritative.
- Per-event `Auth: {Authorizer: NONE}` becomes an operation-level empty OpenAPI security array, preserving the public opt-out instead of inheriting the API default.
- The stack-level overlap regression now verifies the single deployed route has `AuthorizationType: JWT`, a non-null authorizer ID, and its scope; a unit regression covers `NONE`.

I cross-checked the shape and default/override behavior against the official AWS SAM transformer (`ApiGatewayV2Authorizer.generate_openapi`, `set_path_default_authorizer`, and `_set_method_authorizer`). Focused Java 25 verification: 35 tests, 0 failures/errors. Full CI and the new Greptile review are running on this head.

aniketshukla1:
Reviewed the new Greptile 4/5 finding in `a162107`. I did not change `security: []` to override the SAM default because that would diverge from the upstream transformer:

- AWS SAM `OpenApiEditor.set_path_default_authorizer` checks `if existing_security`; an empty array is deliberately false and receives `DefaultAuthorizer`.
- AWS SAM `_set_method_authorizer` encodes the supported public override from a Function event as `Authorizer: NONE`, producing a non-empty sentinel requirement that prevents the later default from replacing it.

Source: [AWS SAM OpenAPI editor](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/open_api/open_api.py#L236-L373).

The behavior is now cited beside the implementation and covered explicitly: a body operation with `security: []` receives the SAM default, while the separate `Auth: {Authorizer: NONE}` regression remains public. The same focused Java 25 run passes 35/35 tests. This preserves AWS compatibility rather than changing behavior solely to satisfy the bot inference.

pgermosen (approved):
Re-checked at current HEAD (`a162107f`) after the last two fixes addressing this morning's findings (default authorizer bypassed on the merge path; explicit `security: []` public override getting overwritten).

Independently verified the `security: []` semantics against AWS SAM's actual `open_api.py` source rather than taking the rebuttal on trust: `set_path_default_authorizer` checks `if existing_security:`, and Python truthiness treats an empty list as unset — so the default authorizer correctly still applies over an explicit `security: []`, matching `hasNonEmptySecurity` returning `false` for an empty array here. Also confirmed via the real `apigatewayv2` botocore model that `ConflictException`/409 for duplicate routes is a genuinely modeled error shape, not invented.

Reverted the merge-into-`DefinitionBody` logic and, separately, the authorization-application call on the merged route — each reproduces the exact original failure (`ROLLBACK_COMPLETE` / duplicate route; route created without JWT auth) when reverted. Restored, both green. Full `apigatewayv2` + `cloudformation` suites pass clean on the PR branch and on a fresh merge against current `main`.

Good to merge — this went through an unusually thorough gauntlet (15 greptile rounds plus my own blocking review with a live repro) and every substantive finding has a verifiable fix.

PR #1957, opened 2026-07-26
test(s3): cover shared edge host routing

CoreOxide:
## Summary

- cover LocalStack and Floci shared edge hostnames as path-style endpoints
- prevent the `localhost` label from being interpreted as an S3 bucket
- add explicit regression coverage for the routing concern raised in #1081

## Testing

- `./mvnw test -Dtest=S3VirtualHostFilterTest -Denforcer.skip=true` (57 tests passed)


hectorvent (approved):
Thanks @CoreOxide

PR #1961, opened 2026-07-26
fix(cloudformation): expand SAM AutoPublishAlias into Version + Alias

### LINKED ISSUE #1944: [BUG] SAM transform drops `AutoPublishAlias` — alias-qualified invokes 404 after a CREATE_COMPLETE deploy
## Service

Lambda / CloudFormation (SAM transform)

## AWS API Action

`CloudFormation:CreateStack` with the `AWS::Serverless-2016-10-31` transform —
`AutoPublishAlias` on `AWS::Serverless::Function`. Observed through
`Lambda:Invoke` and `Lambda:ListAliases`.

## Expected behavior

Real SAM expands `AutoPublishAlias: production` into two extra resources per
function:

- `AWS::Lambda::Version` (`<FunctionId>Version<sha>`)
- `AWS::Lambda::Alias` (`<FunctionId>Aliasproduction`), `Name: production`,
  `FunctionVersion: !GetAtt <VersionId>.Version`

After the deploy, `list-aliases` returns `production` and an alias-qualified
invoke of `<function>:production` succeeds.

## Actual behavior

The property is silently dropped. The stack reaches `CREATE_COMPLETE`, the
function exists with only `$LATEST`, and no alias is created — nothing is
logged, because `AutoPublishAlias` is never read at all (zero occurrences in
the repo, including `compatibility-tests/`).

An alias-qualified invoke then fails:

```
An error occurred (ResourceNotFoundException) when calling the Invoke
operation: Alias not found: production
```

The failure surfaces at invoke time, far from the deploy that caused it.

## Reproduction

No `sam` CLI needed — Floci supports the transform natively (#970).

```bash
cat > /tmp/t.yaml <<'YAML'
Resources:
  Hello:
    Properties:
      FunctionName: hello
      Runtime: nodejs20.x
      Handler: index.handler
      AutoPublishAlias: production
      InlineCode: |
        exports.handler = async () => ({ ok: true });
YAML

  --stack-name alias-repro --template-body file:///tmp/t.yaml \
  --capabilities CAPABILITY_IAM
aws --endpoint-url http://localhost:4566 cloudformation wait \
  stack-create-complete --stack-name alias-repro   # CREATE_COMPLETE

# Expected: one alias named "production". Actual: empty.
aws --endpoint-url http://localhost:4566 lambda list-aliases \
  --function-name hello

# Expected: 200. Actual: ResourceNotFoundException "Alias not found: production".
  --function-name hello:production /tmp/out.json
```

## Root cause

Two gaps, both required — this is why the transform alone isn't enough:

1. `SamTransformProcessor.expandServerlessFunction` (`:349`) emits exactly a
   role, one `AWS::Lambda::Function`, and event source mappings. It never reads
   `AutoPublishAlias`, so no Version/Alias resources are synthesized. The
   `default -> LOG.debugv("Unsupported SAM resource type")` arm (`:88`) only
   fires for unknown resource *types*, not unknown properties — hence the
   silence.
2. `CloudFormationResourceProvisioner` (`:269`–`:302`) provisions only
   `AWS::Lambda::Function`, `::LayerVersion`, and `::EventSourceMapping`. There
   are no `AWS::Lambda::Version` or `AWS::Lambda::Alias` cases, so even a
   hand-written template with explicit Version + Alias resources is ignored.

The runtime side is already complete, so the fix is plumbing rather than new
service behavior: `LambdaService.publishVersion` (`:926`) mints real immutable
version snapshots, `LambdaAliasStore` / `LambdaAlias` back the alias APIs, and
alias-qualified invoke resolution already throws the error above
(`LambdaService:985`). `Fn::GetAtt` resolves generically from a resource's
registered attributes (`CloudFormationTemplateEngine:377`), so the Version
provisioner only needs to register a `Version` attribute for the Alias resource
to reference.

Both changes belong in one PR despite the "one feature per PR" guideline: the
transform alone emits resources the provisioner drops, and the provisioner alone
is unreachable from a SAM template. Neither half is independently testable
end-to-end.

## Open question for maintainers

Real SAM also repoints event sources and API Gateway integrations at the alias
rather than `$LATEST`. Should this change do the same, or land the
Version + Alias resources first and keep event sources on the unqualified
function? The former matches AWS more closely but changes the ARN existing
stacks' ESMs bind to, so it may warrant its own issue.

## Why is this needed?

Our local integration stack deploys ~21 SAM lambdas into Floci, and the
services invoke several of them alias-qualified (`auto-apt-sensors:production`,
`messenger:production`). Every deploy has to follow up with a scripted
`describe-stack-resources` + `create-alias` loop over each function in the
stack to synthesize what the template already declared. Because Floci deploys
delete and recreate the stack, that loop re-runs on every deploy.

## Environment

  tag, `b0e56a9a`)

- [x] Yes
- [ ] No

bdemoss-runwise:
## Summary

`AutoPublishAlias` on `AWS::Serverless::Function` was read nowhere in the
codebase, so the SAM transform silently dropped it: the stack reached
`CREATE_COMPLETE`, the function deployed with only `$LATEST`, and an
alias-qualified invoke then failed with `Alias not found` far from the deploy
that caused it. Nothing was logged either — the `default -> LOG.debugv
("Unsupported SAM resource type")` arm only covers unknown resource *types*, not
unknown properties.

This expands it into the `AWS::Lambda::Version` + `AWS::Lambda::Alias` pair real
SAM generates, and provisions those two types.

Closes #1944

Both halves are needed, which is why they're one PR rather than two: the
transform alone emits resources the provisioner ignores, and the provisioner
alone is unreachable from a SAM template. Neither is independently testable
end-to-end.

Design notes:

- The Lambda service already implements `publishVersion` and the alias APIs, so
  this is plumbing rather than new service behavior.
- The new types go through `CfnResourceProvisioner` (new
  `LambdaVersionAliasCfnProvisioner`) rather than the legacy switch, following
  the registry migration. `AWS::Lambda::Function` is deliberately left where it
  is — extracting it is a separate change.
- `Ref` on `FunctionName` gives `topologicalSort` the edges it needs — function →
  version and function → alias. There is no version → alias edge while
  `FunctionVersion` is the literal `$LATEST`, and none is required, since the alias
  does not reference the version; it returns with the `Fn::GetAtt` form once #1987
  lands.
- The `Version` resource is emitted so the resource set and `Fn::GetAtt` match real
  SAM, but the alias targets **`$LATEST`** rather than that version. Floci cannot
  invoke a published version today — the snapshot carries no code path, so a
  version-qualified invoke times out on a cold start (#1987) and silently runs
  `$LATEST`'s code when a warm container exists (#1988). Emitting the faithful form
  broke the alias-qualified invoke this expansion exists to enable, which an earlier
  revision of this PR did. One line flips it back once #1987 lands.
- An intrinsic alias name (e.g. `!Ref StageName`) is passed through to `Name`
  for the engine to resolve at provision time; only the generated logical id
  needs a literal, and it drops the suffix in that case.
- `RoutingConfig` is not carried over — nothing in the SAM expansion emits it,
  so wiring it would be untested surface.
- The alias logical id's suffix goes through the same `sanitize()` helper every
  other derived id in this file uses. Alias names may contain `-` and `_`, neither
  valid in a CloudFormation logical id, so the raw form emitted
  `MyFuncAliasblue-green`. The alias name sent to Lambda is untouched.

Open question from the issue, deliberately not addressed here: real SAM also
repoints event sources and API Gateway integrations at the alias rather than
`$LATEST`. That changes the ARN existing stacks' ESMs bind to, so it seemed
worth deciding separately rather than folding in.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior was: a declared `AutoPublishAlias` produced no alias, so
`ListAliases` returned empty and `Invoke` on `<function>:production` returned
`ResourceNotFoundException: Alias not found: production` despite a successful
deploy.

Verified against AWS CLI 2.36.8 (`cloudformation create-stack`, `lambda
list-aliases`, `lambda get-alias`, `lambda invoke`) on Floci 1.5.33.

## Tests

Four new transform/integration tests, each confirmed to **fail without this change**:

- `SamTransformProcessorTest.expandSamTemplate_autoPublishAliasGeneratesVersionAndAlias`
- `SamTransformProcessorTest.expandSamTemplate_autoPublishAliasFromIntrinsicPassesNodeThrough`
- `SamTransformProcessorTest.expandSamTemplate_autoPublishAliasWithSeparatorsYieldsAlphanumericLogicalId`
  — without `sanitize()` the logical id comes out as `MyFuncAliasblue-green_2`
- `SamTransformIntegrationTest.samFunction_withAutoPublishAlias_createsVersionAndAlias`
  — on the unpatched tree this fails with exactly `"message": "Alias not found:
  production"`, the symptom from the issue

Plus two negative tests (`withoutAutoPublishAliasGeneratesNeither`,
`samFunction_withoutAutoPublishAlias_createsNoAlias`) so the expansion can't
start firing unconditionally.

The integration test **invokes through the alias** rather than only asserting the
alias record exists. That distinction matters: the record assertion was satisfied by
an alias pointing at a published version, which times out on invoke, so the first
revision of this PR passed its own tests with a broken alias. Verified by reverting
the alias target with the record assertion relaxed — the invoke assertion fails.

`LambdaVersionAliasCfnProvisionerTest` covers the provisioner directly, alongside
the existing Sqs/Rds provisioner tests — including the two paths the integration
test cannot reach: the already-exists fallback (`createAlias` conflicting with an
alias that outlived a stack delete) and `delete()`, which the integration test's
`@AfterEach` invokes but never asserts on. Both were verified to fail when the
behavior they cover is removed. It also pins delete's tolerance of a physical id
that is not an alias ARN — the generic fallback's `logicalId + "-" + uuid8` shape,
which is what a stack predating this provisioner holds for an explicitly declared
`AWS::Lambda::Alias`.

`./mvnw test`: **7957 tests, 0 failures, 0 errors** (8 skipped), 587 classes, JDK 25,
single process. Green including `ElastiCacheIntegrationTest.crossGroupAuthIsRejected`,
which is order/load-sensitive and had failed in an earlier run of mine that overlapped
a second concurrent suite.

`make docs-check` clean; `docs/services/cloudformation.md` updated to list the
two new resource types.

Two commits share the subject "point the SAM alias at `$LATEST`" — `d753187e` landed
only the test half and `62d09937` is the fix, whose message explains the stumble.
Left unsquashed since squash-merge collapses them and the alternative is a
force-push.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
@@ -361,6 +361,8 @@ class SamTransformProcessor {
         ObjectNode lambdaResource = createLambdaFunction(logicalId, roleLogicalId, properties, hasExplicitRole);
         resources.set(logicalId, lambdaResource);
 
+        expandAutoPublishAlias(logicalId, properties, resources);
+
         JsonNode events = properties.path("Events");
         if (events.isObject()) {
             Iterator<Map.Entry<String, JsonNode>> eventFields = events.fields();
@@ -369,6 +371,54 @@ class SamTransformProcessor {
                 expandFunctionEvent(logicalId, entry.getKey(), entry.getValue(), resources);
             }
         }
+    }
+
+    /**
+     * Expands {@code AutoPublishAlias} into the {@code AWS::Lambda::Version} +
+     * {@code AWS::Lambda::Alias} pair real SAM generates, so an alias-qualified invoke
+     * ({@code <function>:production}) resolves instead of failing with "Alias not found".
+     * Without this the property is silently dropped and the deploy still reports
+     * CREATE_COMPLETE, so the failure only surfaces later, at invoke time.
+     *
+     * <p>The alias name is normally a literal, but SAM also allows an intrinsic (e.g.
+     * {@code !Ref StageName}). The node is passed through to {@code Name} either way so the
+     * template engine resolves it at provision time; only the generated logical id needs a
+     * literal, and it drops the suffix when the value isn't textual.
+     *
+     * <p>{@code Ref}/{@code Fn::GetAtt} in the emitted properties give the ordering edges
+     * (function → version → alias) that {@code topologicalSort} already follows, so no
+     * explicit {@code DependsOn} is needed.
+     */
+    private void expandAutoPublishAlias(String functionLogicalId, JsonNode properties, ObjectNode resources) {
+        JsonNode aliasName = properties.path("AutoPublishAlias");
+        if (aliasName.isMissingNode() || aliasName.isNull()
+                || (aliasName.isTextual() && aliasName.asText().isBlank())) {
+            return;
+        }
+
+        String versionId = uniqueId(functionLogicalId + "Version", resources);
+        ObjectNode versionDef = objectMapper.createObjectNode();
+        versionDef.put("Type", "AWS::Lambda::Version");
+        ObjectNode versionProps = objectMapper.createObjectNode();
+        versionProps.set("FunctionName", ref(functionLogicalId));
+        versionDef.set("Properties", versionProps);
+        resources.set(versionId, versionDef);
+
+        String aliasSuffix = aliasName.isTextual() ? aliasName.asText() : "";
+        String aliasId = uniqueId(functionLogicalId + "Alias" + aliasSuffix, resources);
+        ObjectNode aliasDef = objectMapper.createObjectNode();
+        aliasDef.put("Type", "AWS::Lambda::Alias");
+        ObjectNode aliasProps = objectMapper.createObjectNode();
+        aliasProps.set("FunctionName", ref(functionLogicalId));
+        aliasProps.set("Name", aliasName.deepCopy());
+        ObjectNode versionGetAtt = objectMapper.createObjectNode();
+        ArrayNode getAttParts = objectMapper.createArrayNode();
+        getAttParts.add(versionId);
+        getAttParts.add("Version");
+        versionGetAtt.set("Fn::GetAtt", getAttParts);
+        aliasProps.set("FunctionVersion", versionGetAtt);
+        aliasDef.set("Properties", aliasProps);
+        resources.set(aliasId, aliasDef);
     }
 
     private ObjectNode createExecutionRole(JsonNode properties) {
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
@@ -385,9 +385,11 @@ class SamTransformProcessor {
      * template engine resolves it at provision time; only the generated logical id needs a
      * literal, and it drops the suffix when the value isn't textual.
      *
-     * <p>{@code Ref}/{@code Fn::GetAtt} in the emitted properties give the ordering edges
-     * (function → version → alias) that {@code topologicalSort} already follows, so no
-     * explicit {@code DependsOn} is needed.
+     * <p>The {@code Ref}s on {@code FunctionName} give {@code topologicalSort} the edges it needs
+     * — function → version and function → alias — so no explicit {@code DependsOn} is needed.
+     * There is no version → alias edge while {@code FunctionVersion} is the literal
+     * {@code $LATEST}; none is required, since the alias does not reference the version. It
+     * returns with the {@code Fn::GetAtt} form once #1987 lands.
      */
     private void expandAutoPublishAlias(String functionLogicalId, JsonNode properties, ObjectNode resources) {
         JsonNode aliasName = properties.path("AutoPublishAlias");
@@ -404,7 +406,10 @@ class SamTransformProcessor {
         versionDef.set("Properties", versionProps);
         resources.set(versionId, versionDef);
 
-        String aliasSuffix = aliasName.isTextual() ? aliasName.asText() : "";
+        // Alias names may legally contain '-' and '_', neither of which is valid in a
+        // CloudFormation logical id, so the suffix goes through the same sanitize() every other
+        // derived id in this file uses.
+        String aliasSuffix = aliasName.isTextual() ? sanitize(aliasName.asText()) : "";
         String aliasId = uniqueId(functionLogicalId + "Alias" + aliasSuffix, resources);
         ObjectNode aliasDef = objectMapper.createObjectNode();
         aliasDef.put("Type", "AWS::Lambda::Alias");
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
@@ -411,12 +411,14 @@ class SamTransformProcessor {
         ObjectNode aliasProps = objectMapper.createObjectNode();
         aliasProps.set("FunctionName", ref(functionLogicalId));
         aliasProps.set("Name", aliasName.deepCopy());
-        ObjectNode versionGetAtt = objectMapper.createObjectNode();
-        ArrayNode getAttParts = objectMapper.createArrayNode();
-        getAttParts.add(versionId);
-        getAttParts.add("Version");
-        versionGetAtt.set("Fn::GetAtt", getAttParts);
-        aliasProps.set("FunctionVersion", versionGetAtt);
+        // Real SAM points the alias at the published version (Fn::GetAtt <Version>.Version).
+        // Floci cannot invoke a published version: the snapshot carries no code path, so a
+        // version-qualified invoke times out on a cold start (#1987) and silently runs $LATEST's
+        // code when a warm container happens to exist (#1988). Aiming the alias there would break
+        // the alias-qualified invoke this expansion exists to enable, so point it at $LATEST — the
+        // alias resolves and runs the function's code, which is the behavior callers depend on.
+        // Switch to the GetAtt form once #1987 lands.
+        aliasProps.put("FunctionVersion", "$LATEST");
         aliasDef.set("Properties", aliasProps);
         resources.set(aliasId, aliasDef);
     }
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/LambdaVersionAliasCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/LambdaVersionAliasCfnProvisioner.java
@@ -1,0 +1,125 @@
+package io.github.hectorvent.floci.services.cloudformation.provisioners;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.services.cloudformation.model.StackResource;
+import io.github.hectorvent.floci.services.lambda.LambdaArnUtils;
+import io.github.hectorvent.floci.services.lambda.LambdaService;
+import io.github.hectorvent.floci.services.lambda.model.LambdaAlias;
+import io.github.hectorvent.floci.services.lambda.model.LambdaFunction;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import org.jboss.logging.Logger;
+
+import java.util.Set;
+
+/**
+ * CloudFormation provisioning for {@code AWS::Lambda::Version} and {@code AWS::Lambda::Alias}.
+ *
+ * <p>Covers only these two types, not {@code AWS::Lambda::Function} — that one still lives in
+ * {@code CloudFormationResourceProvisioner}'s switch, which the registry falls through to. The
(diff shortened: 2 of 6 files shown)


src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
```diff
@@ -371,6 +373,61 @@ private void expandServerlessFunction(String logicalId, JsonNode properties, Obj
...
+        aliasProps.set("FunctionName", ref(functionLogicalId));
+        aliasProps.set("Name", aliasName.deepCopy());
+        // Real SAM points the alias at the published version (Fn::GetAtt <Version>.Version).
+        // Floci cannot invoke a published version: the snapshot carries no code path, so a
+        // version-qualified invoke times out on a cold start (#1987) and silently runs $LATEST's
+        // code when a warm container happens to exist (#1988). Aiming the alias there would break
+        // the alias-qualified invoke this expansion exists to enable, so point it at $LATEST — the
+        // alias resolves and runs the function's code, which is the behavior callers depend on.
+        // Switch to the GetAtt form once #1987 lands.
+        aliasProps.put("FunctionVersion", "$LATEST");
```
pgermosen, line 426: **The revert trigger says #1987, but your own diagnosis needs #1988 too.**

Pointing the alias at `$LATEST` is the right call and the comment justifies it well. My concern is only the exit condition: *"Switch to the GetAtt form once #1987 lands"* names one issue, while the paragraph directly above names two — a version-qualified invoke *"times out on a cold start (#1987) and silently runs `$LATEST`'s code when a warm container happens to exist (#1988)."*

Those need different fixes. #1987 is a missing code location on the snapshot; #1988 is the warm pool being keyed by function name alone, so `acquire` hands an alias→version invoke a container that was started for `$LATEST`. Fixing #1987 makes the version *invocable*; it doesn't stop the pool from serving the wrong container. Flip on #1987 alone and the alias resolves and runs — but after any subsequent code update it runs the newer `$LATEST` code while claiming to be pinned to version N. That's quieter than today's timeout and harder to spot.

Worth knowing where those two actually stand, since it changes the sequencing:

- **#1987 is fixed by #2012** (`fix(lambda): give published versions their code location`), open and, as far as I can tell, ready.
- **#1988 has no PR in flight** that I can find. #1963 is a different warm-pool bug — a container invalidated by a drain while checked out — and doesn't touch the keying.

So the flip is gated on work that hasn't started. Suggest widening the comment to `#1987 and #1988` so whoever sees #1987 close doesn't flip it prematurely — that reader will be trusting this comment precisely because they weren't in this context.

Worth saying that **your test already gets this right** — `SamTransformProcessorTest:143-146` explains the deviation as *"#1987 cold-start timeout, #1988 warm-pool keying"*, naming both. So this is an inconsistency between two comments you wrote rather than a gap in the analysis; it's just that the narrower one sits in the production file, which is where someone will look before flipping it.

Same widening applies to the javadoc on `expandAutoPublishAlias`, which says the version→alias edge *"returns with the `Fn::GetAtt` form once #1987 lands"*.

src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/LambdaVersionAliasCfnProvisioner.java
```diff
@@ -0,0 +1,125 @@
...
+        String description = ctx.resolveOptional(props, "Description");
+
+        LambdaFunction version = lambdaService.publishVersion(ctx.region(), functionName, description);
+
+        // Matches real CloudFormation: Ref returns the qualified ARN, GetAtt Version the version
+        // number (what the Alias resource references), GetAtt FunctionArn the qualified ARN.
+        r.setPhysicalId(version.getFunctionArn());
+        r.getAttributes().put("Version", version.getVersion());
+        r.getAttributes().put("FunctionArn", version.getFunctionArn());
+    }
```
pgermosen, line 93: **Every `UpdateStack` publishes another version, including one that changes nothing.**

`provisionVersion` calls `publishVersion` unconditionally. On an update the resource arrives with its previous physical id already populated — `CloudFormationResourceProvisioner:258` passes the same `StackResource` through, so `r.getPhysicalId()` holds the ARN from the last deploy — but nothing consults it. So a stack deployed five times with an identical template ends up with versions 1 through 5, each a stored snapshot, and `ListVersionsByFunction` grows without bound.

Real SAM avoids this structurally: it hashes the code into the version's *logical id*, so an unchanged deploy produces the same logical resource with unchanged properties and CloudFormation does nothing. Here `uniqueId(functionLogicalId + "Version", resources)` is deliberately stable, so the same logical resource re-provisions every time and each re-provision publishes.

No user-visible harm today — the alias points at `$LATEST`, so invokes are unaffected, and this is bounded by deploy count rather than traffic. It gets more interesting after the flip in the other comment, when the alias starts tracking whichever version the latest deploy minted.

The natural discriminator is the code hash — publish only when the function's `codeSha256` differs from the version behind `r.getPhysicalId()` — which is close to how `provisionLaunchTemplate` in #2019 uses `findExistingLaunchTemplate(region, r.getPhysicalId())` to decide between versioning and replacing.

There's a wrinkle worth knowing: **that discriminator isn't available on this branch.** `publishVersion` here copies no code fields onto the snapshot at all — no `codeSha256`, no code location — which is #1987, and #2012 is the PR that adds them. So the clean fix depends on #2012 landing first, which makes this the third distinct way this PR is downstream of it.

Given that, deferring is entirely reasonable. I'd suggest a one-line note here recording that the republish is known and why it can't be deduped yet, so it doesn't get rediscovered as a mystery later.

pgermosen (approved):
Looks good.

Strong PR. The transform half and the provisioner half genuinely do need to land together, the deviation from real SAM is deliberate and argued rather than hidden, and the description is candid about the earlier revision that passed its own tests with a broken alias.


**Two comments inline**, neither blocking: the `$LATEST` revert trigger names only #1987 when your own analysis also requires #1988, and `provisionVersion` publishes a new version on every update including no-op ones.

@hectorvent 
**Sequencing, since these are all in flight at once:** this PR is downstream of **#2012** in three separate ways — it fixes #1987 (the stated flip trigger), it's what makes an alias→version target actually invocable, and it adds the `codeSha256` that the version-dedup fix would key on. #2026 fixes #1989, also referenced here. #1988 is the one with no PR behind it. If these are being merged as a group, #2012 first is the ordering that lets the others simplify rather than accumulate workarounds.

pgermosen (approved):
Re-approving. This has been approved since 28 July and has sat in the queue through 177 commits on `main`, so rather than let the original approval carry on trust I re-verified it against `main` as it stands today. It holds up on every axis I checked.

## Re-verified against current `main` (`12154297`)

**It still compiles.** Full `test-compile` clean across the tree. Worth stating explicitly, because three PRs today were broken by constructor signatures widening underneath them while GitHub still reported `CLEAN` with green checks — this is not one of them.

**The tests still pass.** 485 CloudFormation tests green, including `LambdaVersionAliasCfnProvisionerTest` at 9/9.

**Nothing has landed that competes with it.** `AutoPublishAlias`, `AWS::Lambda::Version` and `AWS::Lambda::Alias` have zero occurrences in `main`'s sources, so this remains the only implementation and there is no overlap to reconcile.

**The design rationale has not gone stale**, which was the thing most likely to have rotted. The alias deliberately targets `$LATEST` rather than the published version because Floci cannot invoke a published version yet — #1987 and #1988. I checked both: still open. So the workaround is still the right call, and your "one line flips it back once #1987 lands" note still applies exactly as written.

## The tests are load-bearing

I reverted each piece rather than re-reading the diff:

| Reverted | Result |
|---|---|
| the `AutoPublishAlias` expansion | 4 tests fail, including the end-to-end `samFunction_withAutoPublishAlias_createsVersionAndAlias` |
| `sanitize()` on the alias logical id | `logical id must be alphanumeric, was: MyFuncAliasblue-green_2` |

That second message is the exact defect your description calls out, which is the sign of a test written from the failure rather than around it.

The coverage is also layered deliberately, and that turns out to matter here. Disabling the transform leaves `LambdaVersionAliasCfnProvisionerTest` green, because it drives the provisioner directly. So the two halves are pinned independently — which is precisely the argument your PR description makes for why they belong in one PR but are not one thing. A future change that moves the transform cannot quietly take down the provisioner's coverage with it.

## On the judgement calls

Re-reading these with fresh eyes, the ones I would most expect a later reviewer to question are the ones you pre-empted:

- **Emitting `Version` but pointing the alias at `$LATEST`.** Faithful resource set and `Fn::GetAtt` shape, working invoke. You noted an earlier revision did it the faithful way and broke the very thing the expansion exists to enable — that is the detail that makes this a considered trade rather than a shortcut.
- **Leaving `AWS::Lambda::Function` in the legacy switch** while the new types go through the registry. Right call; migrating it is a separate change with a separate blast radius.
- **Not carrying `RoutingConfig`**, since nothing in the expansion emits it and wiring it would be untested surface.
- **Leaving event sources and API Gateway integrations pointed at `$LATEST`** rather than the alias, because repointing changes the ARN existing stacks' ESMs bind to. Splitting that out was correct — it is a migration question, not a transform question.

Merging closes #1944; #1987 and #1988 remain the tracked follow-ups.

Thank you for the patience on the queue delay, and for a description that made re-validating three weeks later this cheap — every claim in it named the thing I needed to check.

PR #1963, opened 2026-07-26
fix(lambda): discard warm container invalidated while in use

### LINKED ISSUE #1962: [BUG] Warm container checked out during a redeploy is pooled again and serves the old code
## Service

Lambda (warm container pool)

## AWS API Action

`Lambda:Invoke`, after `CloudFormation:DeleteStack` + `CreateStack` (or any
`deleteFunction` → recreate) while the function is being invoked.

## Expected behavior

After a function is deleted and recreated, an invoke runs the new code. Deleting a
function drains its containers, so no container outlives it.

## Actual behavior

Durably serves the **old code and the old baked environment**, with no
self-correction. `get-function-configuration` reports the *new* values while every
invoke returns the *old* ones — the config and the running container disagree
indefinitely.

Only reproduces when the function is invoked **concurrently** with the
delete/recreate. An idle redeploy drains cleanly, which is why it isn't obvious.

## Reproduction

Deploy a function, invoke it in a loop, and delete + recreate it mid-loop with
changed code and a changed env var. Observed on 1.5.33 with a 4-ESM function:

```
$ aws lambda get-function-configuration --function-name f \
    --query 'Environment.Variables.MARKER'
"v2"                       # new config

$ aws lambda invoke --function-name f /tmp/out.json && cat /tmp/out.json
{"marker":"v1","code":"v1"}   # old code AND old env, 5+ invokes over 10s,
                              # container up 54s, no recovery

$ docker rm -f <the floci-f-* container>
$ aws lambda invoke --function-name f /tmp/out.json && cat /tmp/out.json
{"marker":"v2","code":"v2"}   # correct again
```

A function under an event-source poller is invoked continuously, so it hits this
on essentially every redeploy.

## Root cause

`WarmPool.pool` holds only *idle* containers:

1. `acquire` (`:113`) does `queue.pollFirst()` — a container serving an invocation
   is removed from the pool and referenced only by the invoking thread.
2. `deleteFunction` → `drainFunction` (`:198`) does `pool.remove(functionName)`
   and stops what was queued. A checked-out container is invisible to it. Note it
   also returns early when the queue is absent — which is exactly the state when
   every live container is checked out.
3. `release` (`:151`) unconditionally re-adds the handle via
   `computeIfAbsent(handle.getFunctionName(), …)`, recreating the queue for a
   function that was deleted and recreated, with no check that the function still
   exists or that the code matches.
4. The next `acquire` finds it alive and logs `Reusing warm container`.

`pushCodeUpdate` (`:177`) delegates to `drainFunction`, so the reactive S3
hot-reload path (#509) has the same blind spot.

## Environment

  tag, `b0e56a9a`)

- [x] Yes
- [ ] No

bdemoss-runwise:
## Summary

`drainFunction` only stops containers sitting **idle** in the pool. `acquire` polls
a handle off the queue, so a container serving an invocation is referenced solely
by the invoking thread and a drain cannot see it — `release` then hands it back for
a function that was deleted and, on a redeploy, recreated. The next invoke reuses
it and runs the old code and old baked environment, durably:
`get-function-configuration` reports the new values while every invoke returns the
old ones.

A function under an event-source poller is invoked continuously, so it hits this on
essentially every redeploy. An idle redeploy drains cleanly, which is why it is
invisible without concurrent invocation.

Closes #1962

## Approach

`WarmPool` keeps a per-function generation counter:

- `acquire` samples the function's generation **before it touches the pool**, and
  stamps the handle it hands out with that value.
- `drainFunction` detaches the queue and then bumps the generation —
  unconditionally, before the empty-queue early return, since an absent queue is
  exactly the state where every live container is checked out.
- `release` discards and stops a handle whose stamp is stale instead of pooling it.

Both orderings carry weight; getting either backwards reopens the bug through a
narrower window:

- **Sample before checkout.** From `pollFirst()` onwards the handle is checked out
  and invisible to a drain, and what follows is a Docker round-trip — the
  `isAlive()` probe on the reuse path, a full container start on the cold path.
  Reading the generation after that stamps the handle with the post-drain value, so
  `release` pools it anyway.
- **Detach before bumping.** The mirror image: with the generation bumped but the
  queue still reachable, an `acquire` sampling after the bump can poll a pre-drain
  handle and stamp it valid.

Comparing generations rather than force-stopping busy containers during the drain
lets the in-flight invocation finish normally — it just never gets reused.

`pushCodeUpdate` delegates to `drainFunction`, so the reactive S3 hot-reload path
(#509) is fixed with it.

The generation entry per function name is retained deliberately: removing it would
reset the counter to zero and re-admit exactly the stale handles this prevents. It
is one `long` per function name ever drained.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior was: after a delete + recreate under concurrent invocation, the
function served pre-delete code and environment indefinitely. Observed on 1.5.33
against a function with four DynamoDB-stream event source mappings — 5+ invokes
over 10s, container up 54s, no self-correction; `docker rm -f` on the container
restored correct output immediately.

Real Lambda has no equivalent: a deleted function's execution environments do not
outlive it.

## Tests

Four new `WarmPoolTest` cases, each written **before** the code it covers and
confirmed failing without it:

- `release_afterDrain_doesNotReturnCheckedOutContainerToPool` — the reported bug
- `release_afterPushCodeUpdate_doesNotReturnCheckedOutContainerToPool` — the same
  blind spot on the code-update path
- `release_afterDrainDuringAcquire_doesNotReturnCheckedOutContainerToPool` — the
  drain lands *inside* `acquire`, driven from the `isAlive` stub. The other three
  drain strictly between `acquire` and `release`, so none of them reach the
  sampling window.
- `release_withoutDrain_stillReusesContainerAcrossFunctions` — guards against
  over-correcting: draining one function must not invalidate another's checked-out
  container. Passes before and after, so it is a real guard rather than a
  restatement of the fix.

Deliberately no Docker-backed integration test racing an invoke against a delete:
the timing that makes the bug appear is what would make such a test flaky in CI.
The unit tests pin the mechanism precisely instead.

`./mvnw test`: **7946 tests, 0 failures, 0 errors** (8 skipped), 586 classes, JDK 25,
single process. An earlier revision reported an `ElastiCacheIntegrationTest` failure
as pre-existing; that was contamination from a second suite running concurrently,
and it passes 14/14 here.

## Checklist

- [x] New or updated integration test added (unit tests — see Tests)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/WarmPool.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/WarmPool.java
@@ -120,6 +120,13 @@ public class WarmPool implements ContainerTeardown {
         boolean ephemeral = config != null && config.services().lambda().ephemeral();
         ContainerHandle handle = null;
 
+        // Sample before the container leaves the pool, not after: from pollFirst() onwards the
+        // handle is checked out and invisible to a drain, and what follows — the isAlive() probe,
+        // or a cold start — is a Docker round-trip wide enough for a delete to land in. Reading
+        // the generation afterwards would stamp the handle with the post-drain value and let
+        // release() pool it, which is the very bug this guards against.
+        long generation = currentGeneration(fn.getFunctionName());
+
         if (!ephemeral) {
             ArrayDeque<ContainerHandle> queue = pool.computeIfAbsent(fn.getFunctionName(), k -> new ArrayDeque<>());
             // Skip pooled handles whose container died out-of-band — otherwise the
@@ -152,7 +159,7 @@ public class WarmPool implements ContainerTeardown {
         handle.setState(ContainerState.BUSY);
         // Stamp the generation this container is valid for; release() drops it if a drain has
         // bumped the function's generation in the meantime.
-        handle.setPoolGeneration(currentGeneration(fn.getFunctionName()));
+        handle.setPoolGeneration(generation);
         return handle;
     }
 
@@ -228,13 +235,17 @@ public class WarmPool implements ContainerTeardown {
      * Called on function delete or code update.
      */
     public void drainFunction(String functionName) {
-        // Bump first, and unconditionally: containers checked out right now are not in the queue,
-        // so this is the only thing that keeps them from being pooled again on release. It has to
-        // happen even when the queue is empty — that is precisely the case where every live
-        // container is checked out.
-        generations.merge(functionName, 1L, Long::sum);
-
+        // Detach the queue first, then bump — and bump unconditionally, before the empty-queue
+        // early return, since containers checked out right now are not in the queue and the
+        // generation is the only thing that keeps them from being pooled again on release.
+        //
+        // Order matters as much as unconditionality. Bumping first leaves a window where the
+        // generation has moved but the queue is still reachable, so an acquire sampling after the
+        // bump could still poll a pre-drain handle and stamp it valid. Detaching first means a
+        // concurrent acquire either finds a fresh empty deque (and cold-starts) or sampled its
+        // generation before the bump (and is discarded on release).
         ArrayDeque<ContainerHandle> queue = pool.remove(functionName);
+        generations.merge(functionName, 1L, Long::sum);
         if (queue == null) {
             return;
         }
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/WarmPool.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/WarmPool.java
@@ -42,6 +42,21 @@ public class WarmPool implements ContainerTeardown {
     private final EmulatorConfig config;
     private final int maxPoolSizePerFunction;
     private final ConcurrentHashMap<String, ArrayDeque<ContainerHandle>> pool = new ConcurrentHashMap<>();
+    /**
+     * Per-function generation counter, bumped by {@link #drainFunction(String)}.
+     *
+     * <p>{@code pool} only holds idle containers: {@link #acquire} polls a handle off the queue,
+     * so a container serving an invocation is referenced solely by the invoking thread and is
+     * invisible to a drain. Without the generation check, {@link #release} would hand that
+     * container back for a function that has since been deleted (and, on a redeploy, recreated)
+     * and the next invoke would run the old code and old baked environment. A function under an
+     * event-source poller nearly always has a container checked out, so the drain nearly always
+     * misses it; an idle redeploy drains cleanly, which is what makes this hard to notice.
+     *
+     * <p>Comparing generations rather than stopping busy containers during the drain lets the
+     * in-flight invocation finish normally — it just never gets pooled again.
+     */
+    private final ConcurrentHashMap<String, Long> generations = new ConcurrentHashMap<>();
     private final ScheduledExecutorService evictionScheduler = Executors.newSingleThreadScheduledExecutor(
             r -> { Thread t = new Thread(r, "warm-pool-evictor"); t.setDaemon(true); return t; });
 
@@ -135,7 +150,14 @@ public class WarmPool implements ContainerTeardown {
             LOG.debugv("Reusing warm container for function: {0}", fn.getFunctionName());
         }
         handle.setState(ContainerState.BUSY);
+        // Stamp the generation this container is valid for; release() drops it if a drain has
+        // bumped the function's generation in the meantime.
+        handle.setPoolGeneration(currentGeneration(fn.getFunctionName()));
         return handle;
+    }
+
+    private long currentGeneration(String functionName) {
+        return generations.getOrDefault(functionName, 0L);
     }
 
     /**
@@ -148,6 +170,16 @@ public class WarmPool implements ContainerTeardown {
         if (ephemeral || handle.isHotReload()) {
             LOG.debugv("{0}: stopping container {1} after invocation",
                     handle.isHotReload() ? "Hot-reload" : "Ephemeral", handle.getContainerId());
+            stopQuietly(handle);
+            return;
+        }
+
+        // The function's containers were invalidated while this one was checked out (delete,
+        // redeploy, or code update), so it must not go back into the pool — it would serve the
+        // pre-invalidation code and environment to the next invoke.
+        if (handle.getPoolGeneration() != currentGeneration(handle.getFunctionName())) {
+            LOG.debugv("Discarding container {0} for function {1}: invalidated while in use",
+                    handle.getContainerId(), handle.getFunctionName());
             stopQuietly(handle);
             return;
         }
@@ -196,6 +228,12 @@ public class WarmPool implements ContainerTeardown {
      * Called on function delete or code update.
      */
     public void drainFunction(String functionName) {
+        // Bump first, and unconditionally: containers checked out right now are not in the queue,
+        // so this is the only thing that keeps them from being pooled again on release. It has to
+        // happen even when the queue is empty — that is precisely the case where every live
+        // container is checked out.
+        generations.merge(functionName, 1L, Long::sum);
+
         ArrayDeque<ContainerHandle> queue = pool.remove(functionName);
         if (queue == null) {
             return;
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerHandle.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerHandle.java
@@ -18,6 +18,13 @@ public class ContainerHandle {
     private volatile ContainerState state;
     private volatile long lastUsedMs;
     private Closeable logStream;
+    /**
+     * The warm-pool generation this container was handed out under. {@code WarmPool} bumps a
+     * function's generation whenever its containers are invalidated (delete, redeploy, code
+     * update) and compares on release, so a container checked out across an invalidation is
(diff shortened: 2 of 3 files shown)


src/main/java/io/github/hectorvent/floci/services/lambda/WarmPool.java
```diff
@@ -152,6 +181,16 @@ public void release(ContainerHandle handle) {
...
+        if (handle.getPoolGeneration() != currentGeneration(handle.getFunctionName())) {
+            LOG.debugv("Discarding container {0} for function {1}: invalidated while in use",
+                    handle.getContainerId(), handle.getFunctionName());
+            stopQuietly(handle);
+            return;
+        }
+
         handle.setState(ContainerState.WARM);
         handle.touchLastUsed();
         ArrayDeque<ContainerHandle> queue = pool.computeIfAbsent(handle.getFunctionName(), k -> new ArrayDeque<>());
```
pgermosen, line 196: **There's a third window, on the release side, symmetric to the two you closed.**

Your analysis covers the acquire side (sample before checkout) and the drain side (detach before bump), and I verified both are implemented as described. But the generation check here is not atomic with the insertion at `:196-202`:

```java
if (handle.getPoolGeneration() != currentGeneration(...)) { ... return; }   // :187  check
handle.setState(ContainerState.WARM);                                       // :194
handle.touchLastUsed();                                                     // :195
ArrayDeque<ContainerHandle> queue = pool.computeIfAbsent(...);              // :196  <-- creates a NEW deque
synchronized (queue) { ... queue.addFirst(handle); }                        // :198-202  insert
```

A `drainFunction` that lands entirely between `:187` and `:196` does `pool.remove` + bump and stops the idle containers, and then this thread's `computeIfAbsent` finds no mapping and installs a **fresh** deque, into which it adds a pre-drain handle. The next `acquire` samples the new generation, polls that handle, and stamps it valid — so it is now permanently pooled for the new code. That is the reported bug again, reached through a narrower window.

There's a second, milder outcome from the same race: if the drain's `pool.remove` lands *after* `:196` and its iteration finishes before `:202`, the handle ends up in a detached deque that nothing can reach — never stopped, never reused. A leaked container rather than a stale one.

Both are far narrower than what you fixed — microseconds between a field write and a map lookup, versus a full Docker round-trip — so this is not a blocker, and the PR is a large net improvement either way. But it's the same defect class, and your description makes the case that these windows are worth closing on their merits rather than on their width.

**Structural option, if you want it closed by construction rather than by ordering argument:** keep the generation and the queue in one value so a single `pool.compute(...)` covers both, and all three windows collapse into the map's per-bin lock:

```java
private record PoolEntry(long generation, ArrayDeque<ContainerHandle> queue) {}
```

- `drainFunction` — `pool.compute(fn, (k, e) -> new PoolEntry(e == null ? 1 : e.generation() + 1, new ArrayDeque<>()))`: detach and bump atomically.
- `release` — do the generation comparison and the `addFirst` inside one `pool.compute`.
- `acquire` — sample the generation and `pollFirst` inside one `pool.compute`.

That removes the need to reason about ordering at all, which given how carefully the current comments have to argue it seems like the right trade.

**Minimal alternative** if you'd rather keep this PR's shape: re-check after inserting, and pull the handle back out if the generation moved. It narrows the window a lot without closing it — worth it only if the structural change feels out of scope.

**A test for it, using the technique you already invented.** `release_afterDrainDuringAcquire_...` drives the drain from the `isAlive` stub; the release-side equivalent is a spied handle whose `touchLastUsed()` fires the drain, which lands it exactly between the check and the insert:

```java
ContainerHandle inFlight = spy(pool.acquire(fn));
doAnswer(inv -> { pool.drainFunction("release-race-fn"); return null; })
        .when(inFlight).touchLastUsed();
pool.release(inFlight);
verify(containerLauncher).stop(inFlight);   // fails today: it is pooled instead
```

src/main/java/io/github/hectorvent/floci/services/lambda/WarmPool.java
```diff
@@ -196,7 +235,17 @@ public void destroyHandle(ContainerHandle handle) {
...
+        //
+        // Order matters as much as unconditionality. Bumping first leaves a window where the
+        // generation has moved but the queue is still reachable, so an acquire sampling after the
+        // bump could still poll a pre-drain handle and stamp it valid. Detaching first means a
+        // concurrent acquire either finds a fresh empty deque (and cold-starts) or sampled its
+        // generation before the bump (and is discarded on release).
         ArrayDeque<ContainerHandle> queue = pool.remove(functionName);
+        generations.merge(functionName, 1L, Long::sum);
         if (queue == null) {
             return;
```
pgermosen, line 250: **Both orderings verified, and the unconditional bump is load-bearing — worth the comment it has.**

Confirming the two claims in the description, since they're the load-bearing parts and neither is obvious from the code alone:

- **Detach before bump** — `pool.remove` at `:247` precedes `generations.merge` at `:248`. Reversed, an `acquire` that samples after the bump could still reach the old queue and stamp a pre-drain handle as valid. Correct as written.
- **Bump before the `queue == null` early return** — this is the case that matters most, and it'd be easy to "tidy" into a bug later. A null queue is precisely the state where every live container is checked out, so a drain that returned early without bumping would leave every one of them poolable. The comment says so; good.

`generations` retention is also right, and for the reason given: `currentGeneration` falls back to `0L` (`:166`), so removing an entry would reset the counter and re-admit exactly the handles stamped `0` that this is meant to discard. One `long` per function name ever drained is the correct trade.

One observation rather than a request: `poolGeneration` on `ContainerHandle` is `volatile`, which is what makes the cross-thread read in `release` sound — worth keeping in mind if anyone later "simplifies" it, since nothing else in the class documents why.

pgermosen (dismissed):
Looks good, a few comments bellow.

Careful fix for a genuinely nasty bug, and the description does the hard part — explaining why each ordering matters rather than just asserting the mechanism. I traced the code against every claim.

One finding (inline): there's a third window on the release side, between the generation check at :187 and the insertion at :196-202, symmetric to the two you closed. A drain landing there results in either a stale container in the live pool (the original bug, narrower) or a container leaked into a detached deque. Both are microsecond-scale rather than Docker-round-trip-scale, so it's not a blocker — but it's the same class, and I've suggested both a structural fix that collapses all three windows into the map's bin lock and a test using the same stub technique you invented.

pgermosen:
@bdemoss-runwise, this seems ready, but it has conflicts. Could you solve them?

pgermosen (commented):
@bdemoss-runwise — apologies for the churn here, and thank you for this. The bug you found is real and the write-up is one of the clearest I've read on this repo: the observation that *"a function under an event-source poller nearly always has a container checked out, so the drain nearly always misses it; an idle redeploy drains cleanly, which is what makes this hard to notice"* is exactly why this survived as long as it did. The decision to sample the generation *before* `pollFirst()` rather than after, and the comment explaining why, is the subtle part and you got it right.

I'm withdrawing my approval, and I want to be clear that it is not a reversal on the merits.

Since I approved this on 28 July, #2053 (Lambda EFS file system configs) has grown a `Lease(poolState, epoch, environmentKey)` mechanism in `WarmPool` that fixes the same defect. It samples the epoch before the container is checked out — under the pool lock — and drops the handle in `release()` when a drain has bumped the epoch in the meantime, which is your fix. It additionally scopes invalidation per execution environment, so updating `$LATEST` no longer discards a published version's warm containers. I've just approved that PR.

That leaves two problems with landing this one as it stands:

1. Merging both would put two parallel invalidation mechanisms in one class — your `generations` map plus `ContainerHandle.poolGeneration`, and #2053's per-environment epoch/lease — doing the same job with different lifetimes. That is worse than either alone.
2. This branch is now `CONFLICTING` against `main` (head `605d36f4`, last updated 9 August), so it needs a rebase regardless before it could merge.

My proposal is to land #2053 and close this as superseded, with the credit sitting here — the diagnosis in your description is the reason the problem was understood at all, and #2053's design is a superset of your fix rather than an independent discovery.

Before I do that, two things I'd rather hear from you than assume:

- Does #2053's lease/epoch approach actually cover your case completely, or is there a scenario your `ContainerHandle`-stamped generation catches that a per-environment epoch does not? You have thought about this code more carefully than I have, and I would rather be corrected now than lose coverage.
- If you would prefer to rebase and keep this PR as the vehicle instead, say so and I will hold #2053's merge until they are reconciled in whichever order you two prefer.

Either way this is a coordination problem between two PRs, not a defect in yours. Sorry it landed on you.

pgermosen:
@bdemoss-runwise closing this as superseded by #2053, which merged on 12 August and now carries this fix on `main`. Confirmed directly against current `main`'s `WarmPool.java`: its `Lease(poolState, epoch, environmentKey)` mechanism is a superset of what's here — same checked-out-survives-a-drain fix, plus per-execution-environment scoping so updating `$LATEST` doesn't discard a published version's warm containers.

This isn't a reversal on the merits — the fix here was correct and the write-up explaining why the sample-before-checkout and detach-before-bump orderings matter is genuinely one of the clearer ones on this repo. The diagnosis is why the bug was understood at all; #2053 just happened to grow the same fix independently while adding EFS support, and got there first. Thank you for finding and nailing down a genuinely nasty concurrency bug.

PR #1969, opened 2026-07-27
fix(cloudformation): create AWS::IAM::Role inline Policies instead of silently dropping them

### LINKED ISSUE #1952: CloudFormation: AWS::IAM::Role inline Policies silently dropped while stack reports CREATE_COMPLETE
## Summary

When a CloudFormation template defines an `AWS::IAM::Role` with inline `Policies`, Floci creates the role and its `AssumeRolePolicyDocument` correctly, but the **inline policies are never created**. The stack still reaches `CREATE_COMPLETE` and the resource is reported as `CREATE_COMPLETE`, so nothing signals that part of the template was not applied.

## Environment

- Floci image `floci/floci:latest`, server version **1.5.33**, edition `floci-always-free`
- floci-cli **0.1.9** (`floci-windows-amd64.exe`)
- Windows 11, Docker 29.6.1
- boto3 1.43.56 / botocore 1.43.56, endpoint `http://localhost:4566`, region `eu-central-1`

## Reproduction

```python
import boto3, json, time, uuid

E = "http://localhost:4566"
kw = dict(endpoint_url=E, region_name="eu-central-1",
          aws_access_key_id="test", aws_secret_access_key="test")
cfn, iam = boto3.client("cloudformation", **kw), boto3.client("iam", **kw)

stack = "repro-" + uuid.uuid4().hex[:6]
role  = "REPRO-ROLE-" + uuid.uuid4().hex[:6]

template = {
    "Resources": {"R1": {"Type": "AWS::IAM::Role", "Properties": {
        "RoleName": role,
        "AssumeRolePolicyDocument": {"Version": "2012-10-17", "Statement": [
            {"Effect": "Allow", "Principal": {"Service": "ec2.amazonaws.com"},
             "Action": "sts:AssumeRole"}]},
        "Policies": [{"PolicyName": "way-too-much", "PolicyDocument": {
            "Version": "2012-10-17",
            "Statement": [{"Effect": "Allow", "Action": "*", "Resource": "*"}]}}],
    }}},
}

cfn.create_stack(StackName=stack, TemplateBody=json.dumps(template),
                 Capabilities=["CAPABILITY_IAM"])
for _ in range(30):
    status = cfn.describe_stacks(StackName=stack)["Stacks"][0]["StackStatus"]
    if status.endswith(("COMPLETE", "FAILED")):
        break
    time.sleep(1)

print("stack status      :", status)
print("role exists       :", role in [r["RoleName"] for r in iam.list_roles()["Roles"]])
print("inline policies   :", iam.list_role_policies(RoleName=role)["PolicyNames"])
print("attached policies :", iam.list_attached_role_policies(RoleName=role)["AttachedPolicies"])
```

## Expected

```
stack status      : CREATE_COMPLETE
role exists       : True
inline policies   : ['way-too-much']
```

## Actual

```
stack status      : CREATE_COMPLETE
role exists       : True
inline policies   : []
attached policies : []
```

`describe_stack_resources` also reports the role as `CREATE_COMPLETE`. The `AssumeRolePolicyDocument` **is** materialised correctly — only the `Policies` property is dropped.

## Control — this is not the IAM read path

Creating the equivalent role directly through the IAM API works as expected, which rules out `list_role_policies` / `get_role_policy` as the cause:

```python
iam.create_role(RoleName=rn, AssumeRolePolicyDocument=json.dumps(trust))
iam.put_role_policy(RoleName=rn, PolicyName="way-too-much",
                    PolicyDocument=json.dumps(pol))

iam.list_role_policies(RoleName=rn)["PolicyNames"]
# -> ['way-too-much']
iam.get_role_policy(RoleName=rn, PolicyName="way-too-much")["PolicyDocument"]
# -> {"Effect": "Allow", "Action": "*", "Resource": "*"}
```

So the IAM inline-policy path itself is fine; the gap is in CloudFormation resource provisioning.

## Why the silent part matters

Beyond ordinary IaC testing, a fairly common pattern is to *apply* a template locally and then *inspect the resulting state* to check it — policy linting by application, guardrail tests in CI, reviewing a template for over-permissive grants.

Under this behaviour that workflow sees the over-permissive policy simply not exist and concludes the template is clean. The failure is silent and biased toward false negatives, which is the harder direction to catch: an outright error would be noticed immediately, whereas `CREATE_COMPLETE` reads as success.

Even surfacing it as a stack event or a warning ("property `Policies` not supported / not applied") rather than silently completing would make it safe to work around.

## Scope of what was tested

Only `AWS::IAM::Role` with the inline `Policies` property, on 1.5.33. I did not test `AWS::IAM::Policy` or `AWS::IAM::ManagedPolicy` as separate resources, or `ManagedPolicyArns`, so I can't say whether the same applies there.

Happy to test a fix or provide more detail. Thanks for building this — the sub-second startup and the no-token, MIT posture are genuinely appreciated.

lex00:
## Summary

Fixes #1952.

`CreateStack` with an `AWS::IAM::Role` that declares inline `Policies` created the role and its trust policy but never created the inline policies, while still reporting `CREATE_COMPLETE`.

This change makes `provisionIamRole` apply each `Policies` entry as an inline role policy (`PutRolePolicy` semantics):

- Intrinsics inside `PolicyDocument` are resolved (`Fn::GetAtt`/`Ref` on sibling resources work).
- A missing `PolicyName` falls back to a stack-scoped generated name, matching the other provisioners.
- Failures propagate, so a policy that cannot be created fails the stack instead of silently succeeding — the silent success is the bug being fixed.

The stack delete path already detaches managed policies and deletes inline policies before `DeleteRole`, so no change was needed there.

## Tests

New `CloudFormationIamInlinePolicyIntegrationTest` deploys a bucket + role with two inline policies and verifies: `CREATE_COMPLETE`, both policies visible via `ListRolePolicies`, `GetRolePolicy` returns the document with `Fn::GetAtt` resolved to the bucket ARN, and `DeleteStack` still removes the role (no `DeleteConflict`).

`mvnw test` over the cloudformation packages: 182 tests, 0 failures.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -5,6 +5,7 @@ import io.github.hectorvent.floci.core.common.AwsNamespaces;
 import io.github.hectorvent.floci.core.common.XmlBuilder;
 import io.github.hectorvent.floci.services.batch.BatchService;
 import io.github.hectorvent.floci.services.cloudformation.model.StackResource;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.CfnRollback;
 import io.github.hectorvent.floci.services.cloudformation.provisioners.CloudFormationResourceRegistry;
 import io.github.hectorvent.floci.services.cloudformation.provisioners.ProvisionContext;
 import io.github.hectorvent.floci.services.cloudformation.provisioners.CfnResourceProvisioner;
@@ -113,7 +114,6 @@ public class CloudFormationResourceProvisioner {
     private static final String LAMBDA_CODE_IDENTITY_ATTR = "FlociLambdaCodeIdentity";
     private static final String LAMBDA_NAME_MODE_ATTR = "FlociLambdaFunctionNameMode";
     private static final String LAMBDA_PACKAGE_TYPE_ATTR = "FlociLambdaPackageType";
-    static final String ROLLBACK_OWNED_ATTR = "__FlociRollbackOwned";
     static final String UPDATE_ROLLBACK_RESTORED_ATTR = "__FlociUpdateRollbackRestored";
     private static final String INLINE_CLEANUP_POLICY_NAME_ATTR = "__FlociInlineCleanupPolicyName";
     private static final String INLINE_CLEANUP_ROLE_TARGETS_ATTR = "__FlociInlineCleanupRoleTargets";
@@ -286,7 +286,6 @@ public class CloudFormationResourceProvisioner {
                 case "AWS::Lambda::Function" -> provisionLambda(resource, properties, engine, region, accountId, stackName);
                 case "AWS::Lambda::LayerVersion" ->
                         provisionLambdaLayerVersion(resource, properties, engine, region, stackName);
-                case "AWS::IAM::Role" -> provisionIamRole(resource, properties, engine, accountId, stackName);
                 case "AWS::IAM::User" -> provisionIamUser(resource, properties, engine, stackName);
                 case "AWS::IAM::AccessKey" -> provisionIamAccessKey(resource, properties, engine);
                 case "AWS::IAM::Policy" -> provisionIamInlinePolicy(resource, properties, engine, stackName);
@@ -487,10 +486,9 @@ public class CloudFormationResourceProvisioner {
             case "AWS::SNS::Subscription" -> snsService.unsubscribe(physicalId, region);
             case "AWS::DynamoDB::Table" -> deleteDynamoTableSafe(physicalId, region);
             case "AWS::Lambda::Function" -> deleteLambdaFunctionSafe(physicalId, region);
-            case "AWS::IAM::Role" -> deleteRoleSafe(physicalId);
             // AWS::IAM::Policy is inline: it is removed together with its owning principal (see
-            // deleteRoleSafe), or precisely via the StackResource-aware delete path. Nothing to do
-            // here when only the physical id (policy name) is known, as on rollback.
+            // IamRoleCfnProvisioner#delete), or precisely via the StackResource-aware delete path.
+            // Nothing to do here when only the physical id (policy name) is known, as on rollback.
             case "AWS::IAM::Policy" -> { }
             case "AWS::IAM::ManagedPolicy" -> deletePolicySafe(physicalId);
             case "AWS::IAM::InstanceProfile" -> iamService.deleteInstanceProfile(physicalId);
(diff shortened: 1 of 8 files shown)


hectorvent:
Thanks for the fix — much appreciated! 🙌

One heads-up: we recently started breaking up the big `CloudFormationResourceProvisioner` class into small per-service provisioners, and we'd like changes that touch an arm of that switch to move that logic into the new structure rather than growing the class further.

If extracting the affected resource type into its own provisioner is feasible here, that would be ideal; otherwise we can treat it as a follow-up. Happy to help if anything about the pattern is unclear — thanks again!

lex00:
Done. Rebased on `upstream/main` and moved to the per-service structure at `390e565`.

IAM role provisioning is now `IamRoleCfnProvisioner`.

lex00:
All three findings are fixed at `56a19d2`.

**Null policy documents dropped** — now a ValidationError. Skipping the entry reached CREATE_COMPLETE without the declared policy, which is the same bug this PR set out to fix, one level in.

**Policy failures leak created roles** — the root cause was placement: the inline-policy loop sat outside the block that unwinds a failed provision. It runs inside it now, so a putRolePolicy failure after createRole cleans up the role and its managed attachments instead of relying on rollback, which only deletes resources that reached CREATE_COMPLETE.

**Failed updates retain inline policies** — inline writes are recorded with the value they replaced. Unwinding restores the previous document, or removes the policy when this attempt introduced the name. A role the stack did not create is not deleted.

The existing `inlinePolicyWithoutADocumentIsSkipped` test was inverted since it pinned the old behavior. 415 tests across the CloudFormation and IAM suites pass.

lex00:
On the Greptile P1 about trust policies — the premise doesn't hold. `IamService.createRole` returns at the existence check:

```java
public IamRole createRole(String roleName, String path, String assumeRolePolicyDocument, ...) {
    if (roles.get(roleName).isPresent()) {
        throw new AwsException("EntityAlreadyExists", ...);
    }
    ...
```

It throws before constructing the role or writing anything, so it never updates an existing role's assume-role policy and there's nothing for the compensation path to restore.

There is a real bug next to it, though, and it's the opposite one: on the adopt path nothing applies `AssumeRolePolicyDocument` at all. `createRole` is called with the resolved document, throws, and the role is then adopted via `getRole(...)`. `IamService.updateAssumeRolePolicy` exists but no CloudFormation code calls it, so an update that changes a role's trust policy is dropped while the stack reports `UPDATE_COMPLETE` — the same failure mode as #1952, one property over.

Confirmed against a build of `main`: the principal stays `lambda.amazonaws.com` after an update declaring `ec2.amazonaws.com`. Filed as #2084 rather than widening this PR, since it's pre-existing and this one has been in review a while. Happy to fold it in here instead if you'd prefer.

lex00:
Missed this last round. Valid.

`originalInlinePolicies` is only read in the failure path, so a policy dropped from the template stays attached after UPDATE_COMPLETE. ManagedPolicyArns is the same, removals never detached. generatePhysicalName is a random UUID, so an unnamed entry gets a new name per execution, the #1973 bug.

Fixing here. I'll reconcile against the prior set on success, and reject an entry with no PolicyName the way this PR already does for a missing PolicyDocument. Tell me if you want a stable generated name instead.

hectorvent:
Thank you for going beyond the rebase here: requiring `PolicyName` (the registry schema marks it required, https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html) also fixes the duplicate-accumulation bug the generated names caused, and the tracked-attribute reconciliation that removes policies the template stopped declaring, while leaving out-of-band additions alone and restoring them on rollback, closes the silent-drift gap cleanly. The inline handling still matches PutRolePolicy semantics (https://github.com/boto/botocore/blob/develop/botocore/data/iam/2010-05-08/service-2.json).

**(blocking)** The branch now conflicts with main: #2056 (IAM ManagedPolicy GetAtt) touched `CloudFormationIamAttachmentProvisionerTest.java`. It looks like a mechanical rebase of that one test file.

Rebase and this is ready to merge.

hectorvent (approved):
Thanks @lex00

PR #1972, opened 2026-07-27
feat(cloudformation): provision AWS::EC2::VPCGatewayAttachment

### LINKED ISSUE #1970: [FEAT] CloudFormation: provision AWS::EC2::VPCGatewayAttachment
## Service

CloudFormation / EC2

## API Action / Feature

Provision `AWS::EC2::VPCGatewayAttachment` from CloudFormation templates. The resource attaches an internet gateway (or VPN gateway) to a VPC — the EC2 side (`AttachInternetGateway`) already exists in Floci; the CFN provisioner just has no case for the resource type, so stacks that declare the attachment skip it.

## AWS Documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpcgatewayattachment.html

## Why is this needed?

Any CDK app that creates a VPC with public subnets synthesizes this type (`ec2.Vpc` emits `VPCGW`). Without it, instances in CDK-deployed VPCs have an internet gateway resource but no attachment, so route table entries pointing at the IGW reference an unattached gateway and reachability behaves differently from AWS.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #1970.

CloudFormation templates declaring `AWS::EC2::VPCGatewayAttachment` (which every CDK `ec2.Vpc` with public subnets synthesizes) previously skipped the resource, leaving the internet gateway unattached.

- `provisionVpcGatewayAttachment` resolves `VpcId`/`InternetGatewayId` and calls the existing `Ec2Service.attachInternetGateway`.
- Physical id encodes `vpc|igw` so DeleteStack can detach (`detachVpcGatewayAttachmentSafe`).
- The `VpnGatewayId` variant records the attachment without an attach call, since there is no VPN gateway model.

## Tests

New `CloudFormationVpcGatewayAttachmentIntegrationTest`: deploys VPC + IGW + attachment, asserts `CREATE_COMPLETE`, `DescribeInternetGateways` reports the attachment to the stack's VPC, and after `DeleteStack` the attachment is gone.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -339,8 +339,6 @@ public class CloudFormationResourceProvisioner {
                 case "AWS::EC2::Subnet" -> provisionSubnet(resource, properties, engine, region);
                 case "AWS::EC2::SecurityGroup" -> provisionSecurityGroup(resource, properties, engine, region, stackName);
                 case "AWS::EC2::InternetGateway" -> provisionInternetGateway(resource, region);
-                case "AWS::EC2::VPCGatewayAttachment" ->
-                        provisionVpcGatewayAttachment(resource, properties, engine, region);
                 case "AWS::EC2::RouteTable" -> provisionRouteTable(resource, properties, engine, region);
                 case "AWS::EC2::SubnetRouteTableAssociation" ->
                         provisionSubnetRouteTableAssociation(resource, properties, engine, region);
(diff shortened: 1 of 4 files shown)


hectorvent:

lex00:
Done. Rebased on `upstream/main` and moved to the per-service structure at `a1231c7`.

`AWS::EC2::VPCGatewayAttachment` is now `Ec2VpcGatewayAttachmentCfnProvisioner`.

lex00:
All four findings are fixed at `65bddcd`.

**Missing VpcId** — the null went straight into `attachInternetGateway`, so `DescribeInternetGateways` exposed an attachment with no VPC and every later detach on that gateway threw on the null compare. `VpcId` is now required, as is exactly one of `InternetGatewayId` or `VpnGatewayId`.

**VPN attachment never provisioned** — you were right that CREATE_COMPLETE was a lie. There is no VPN gateway model to attach to and nothing `DescribeVpnGateways` could report afterwards, so rather than recording a physical id for an attachment that does not exist, the resource now fails with an explicit message. Flagging this as a judgment call: failing loudly beats reporting success, but if you would rather see a real VPN gateway model added so the form works, that is a larger change and I am happy to do it separately.

**Updates orphan prior attachments** — the new pair is attached first and the old detached only after that succeeds, so a failed update leaves the previous attachment intact rather than losing both. An unchanged update no longer re-attaches, which was silently duplicating the attachment.

**VPN ID detaches internet gateway** — the physical id now carries a kind tag (`igw|<vpcId>|<gatewayId>`) and delete acts only on ids this provisioner wrote, so classification no longer depends on the gateway id's prefix.

`Ec2VpcGatewayAttachmentCfnProvisionerTest` covers all of it, including that an `igw-*` value supplied as `VpnGatewayId` never reaches detach. 189 tests across the `CloudFormation*` and `*CfnProvisioner` suites pass.

hectorvent (approved):
Thanks @lex00!

PR #1975, opened 2026-07-27
feat(ec2): return an empty set for DescribeVpnGateways

### LINKED ISSUE #1974: [FEAT] EC2: support DescribeVpnGateways (CDK Vpc.fromLookup fails without it)
## Service

EC2

## API Action / Feature

`DescribeVpnGateways`. Floci returns `UnsupportedOperation`, which makes the CDK CLI's VPC context provider (`Vpc.fromLookup`) abort the lookup and fall back to dummy ids (`vpc-12345`), so any CDK app that looks up a VPC cannot synth/deploy against Floci.

Since Floci has no VPN gateway model, returning an empty `vpnGatewaySet` is AWS-accurate for every account that has no VPN gateways and is enough to unblock the provider.

## AWS Documentation

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpnGateways.html

## Why is this needed?

`cdk deploy`/`cdk synth` for apps using `ec2.Vpc.fromLookup` — the context provider calls DescribeVpcs, DescribeSubnets, DescribeRouteTables (all supported) and DescribeVpnGateways (unsupported), and the whole lookup fails on the last one.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #1974.

`DescribeVpnGateways` returned `UnsupportedOperation`, which makes the CDK CLI's VPC context provider (`Vpc.fromLookup`) abort its lookup — DescribeVpcs / DescribeSubnets / DescribeRouteTables all succeed and the final DescribeVpnGateways kills it, so CDK falls back to dummy ids and synthesis fails.

Floci has no VPN gateway model, so this returns an empty `vpnGatewaySet`, which is AWS-accurate for an account without VPN gateways and unblocks the provider.

## Tests

`Ec2IntegrationTest.describeVpnGatewaysReturnsEmptySet` — the action returns 200 with an empty set instead of `UnsupportedOperation`. Full EC2 suite: 144 tests, 0 failures. Verified end to end: with this change `cdk deploy` VPC lookups complete against a local Floci.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
@@ -111,6 +111,11 @@ public class Ec2QueryHandler {
                 case "DeleteInternetGateway" -> handleDeleteInternetGateway(params, region);
                 case "AttachInternetGateway" -> handleAttachInternetGateway(params, region);
                 case "DetachInternetGateway" -> handleDetachInternetGateway(params, region);
+                // VPN Gateways. There is no VPN gateway model; an empty set is
+                // AWS-accurate for an account without VPN gateways and unblocks the
+                // CDK VPC context provider, which always issues this describe.
+                case "DescribeVpnGateways" -> handleDescribeVpnGateways();
+
                 // Route Tables
                 case "CreateRouteTable" -> handleCreateRouteTable(params, region);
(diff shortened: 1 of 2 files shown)


lex00:
Tightened — now asserts `vpnGatewaySet.item.size()` is 0, matching the other empty EC2 collection tests.

hectorvent (approved):

PR #1977, opened 2026-07-27
fix(cloudformation): resolve Fn::GetAtt [Vpc, DefaultSecurityGroup]

### LINKED ISSUE #1976: [BUG] CloudFormation: Fn::GetAtt [Vpc, DefaultSecurityGroup] passes through unresolved
## Summary

`AWS::EC2::VPC` provisioning does not populate the `DefaultSecurityGroup` attribute, so `Fn::GetAtt [MyVpc, DefaultSecurityGroup]` reaches downstream resources as the literal string `MyVpc.DefaultSecurityGroup`.

Observed with CDK's `Custom::VpcRestrictDefaultSG` custom resource (synthesized by every `ec2.Vpc` with `restrictDefaultSecurityGroup` defaulted on): the handler receives the unresolved string and fails with

```
InvalidGroup.NotFound: The security group 'ResourcesVpc46097686.DefaultSecurityGroup' does not exist
```

Ec2Service already creates a default security group per VPC, so the id exists — it is just not exposed as a CFN attribute.

## Environment

- Floci main (f3659824), Docker image, TLS + embedded DNS active
- CDK 2.x app, `cdk deploy` against `AWS_ENDPOINT_URL=http://localhost:4566`

## Expected

`Fn::GetAtt` on a stack-created VPC resolves `DefaultSecurityGroup` to the real `sg-*` id, as on AWS.

Yes.

lex00:
## Summary

Fixes #1976.

`provisionVpc` did not expose `DefaultSecurityGroup`, so `Fn::GetAtt [MyVpc, DefaultSecurityGroup]` reached downstream resources as the literal string `MyVpc.DefaultSecurityGroup`. CDK's `Custom::VpcRestrictDefaultSG` handler (synthesized by every `ec2.Vpc`) then fails with `InvalidGroup.NotFound`.

Ec2Service already creates a per-VPC default security group; this looks it up after `createVpc` and records its id as the `DefaultSecurityGroup` attribute.

## Tests

New `CloudFormationVpcDefaultSgAttrIntegrationTest`: a stack outputting `Fn::GetAtt [Vpc, DefaultSecurityGroup]` reaches `CREATE_COMPLETE` with the output resolving to a real `sg-*` id, not the literal. Verified end to end with `cdk deploy` of a CDK app whose `VpcRestrictDefaultSG` custom resource previously failed.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -335,7 +335,6 @@ public class CloudFormationResourceProvisioner {
                 // EC2 networking. These delegate to Ec2Service so the resources actually exist
                 // (describe-subnets, ELBv2, etc. can find them) instead of being stubbed with a
                 // fake physical id. Topological ordering guarantees parents are provisioned first.
-                case "AWS::EC2::VPC" -> provisionVpc(resource, properties, engine, region);
                 case "AWS::EC2::Subnet" -> provisionSubnet(resource, properties, engine, region);
                 case "AWS::EC2::SecurityGroup" -> provisionSecurityGroup(resource, properties, engine, region, stackName);
                 case "AWS::EC2::InternetGateway" -> provisionInternetGateway(resource, region);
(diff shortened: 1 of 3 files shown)


hectorvent:
Thanks for the fix — much appreciated! 🙌

One heads-up: we recently started breaking up the big `CloudFormationResourceProvisioner` class into small per-service provisioners, and we'd like changes that touch an arm of that switch to move that logic into the new structure rather than growing the class further.

If extracting the affected resource type into its own provisioner is feasible here, that would be ideal; otherwise we can treat it as a follow-up. Happy to help if anything about the pattern is unclear — thanks again!

lex00:
Done. Rebased on `upstream/main` and moved to the per-service structure at `816d857`.

`AWS::EC2::VPC` is now `Ec2VpcCfnProvisioner`, which is where the `DefaultSecurityGroup` attribute lookup lives.

hectorvent (approved):
Thanks @lex00

PR #1980, opened 2026-07-27
fix(s3): validate FULL_OBJECT checksum on CompleteMultipartUpload

### LINKED ISSUE #1950: [BUG] CompleteMultipartUpload does not validate FULL_OBJECT checksumCRC32
## Service

S3

## AWS API Action

CompleteMultipartUpload

## Expected behavior

When `CompleteMultipartUpload` is called with `ChecksumType=FULL_OBJECT` and a `ChecksumCRC32` that does not match the actual combined object content, AWS S3 returns an error (e.g., `InvalidRequest`). The validation should reject the request when the provided checksum does not match the object assembled from all uploaded parts.

## Actual behavior

Floci returns HTTP 200 (success) without validating the `ChecksumCRC32` against the combined content of the uploaded parts. The checksum is silently ignored.

## Reproduction

```bash
# 1. Create a bucket
aws --endpoint-url http://localhost:4566 s3 mb s3://test-bucket

# 2. Create a multipart upload with FULL_OBJECT checksum
aws --endpoint-url http://localhost:4566 s3api create-multipart-upload \
  --bucket test-bucket --key test-key \
  --checksum-type FULL_OBJECT --checksum-algorithm CRC32

# 3. Upload a part (using some arbitrary content)
echo "Hello" > part1.txt
aws --endpoint-url http://localhost:4566 s3api upload-part \
  --bucket test-bucket --key test-key \
  --upload-id <upload-id-from-step-2> --part-number 1 \
  --body part1.txt

# 4. Complete with a CRC32 that does NOT match "Hello\n"
#    (e.g., base64-encoded CRC32 of completely different data)
aws --endpoint-url http://localhost:4566 s3api complete-multipart-upload \
  --bucket test-bucket --key test-key \
  --upload-id <upload-id> \
  --checksum-type FULL_OBJECT --checksum-crc32 "AAAAAA==" \
  --multipart-upload '{"Parts":[{"ETag":"<etag>","PartNumber":1}]}'
```

Expected: step 4 returns an error because `"AAAAAA=="` is not the CRC32 of `"Hello\n"`.
Actual: Floci returns success (HTTP 200).

## Environment

- Floci version: as bundled in `testcontainers-floci:2.11.0`
- AWS SDK version: Java SDK v2 `2.49.2`
- Running via: Testcontainers (Docker)

avison9:
### Summary

`CompleteMultipartUpload` silently ignored client-supplied `ChecksumType=FULL_OBJECT`
checksums (e.g. `ChecksumCRC32`), always returning HTTP 200 even when the
checksum didn't match the reassembled object. Real S3 rejects this with
`BadDigest`. This matters for anyone relying on Floci to catch multipart
integrity bugs locally before they hit production.

```python
# BEFORE the fix:
aws --endpoint-url http://localhost:4566 s3api complete-multipart-upload \
  --bucket test-bucket --key test-key --upload-id "$UPLOAD_ID" \
  --checksum-type FULL_OBJECT --checksum-crc32 "AAAAAA==" \
  --multipart-upload '{"Parts":[{"ETag":"...","PartNumber":1}]}'
# -> HTTP 200 (wrong: "AAAAAA==" does not match the uploaded content)

# AFTER the fix:
# -> HTTP 400 BadDigest: "The CRC32 checksum you specified did not match the payload."
```

Closes #1950

### Type of change

- [x] Bug fix (`fix:`)

### Root cause

`S3Controller.handleMultipartPost` never read the `x-amz-checksum-type` /
`x-amz-checksum-*` headers on the CompleteMultipartUpload path, and
`S3Service.completeMultipartUpload` had no checksum parameter at all — so
nothing was ever compared against the reassembled object.

### What changed

- `S3Controller`: extract `x-amz-checksum-type` and checksum headers (via the
  existing `extractChecksumFromHeaders` helper) and pass them through.
- `S3Service.completeMultipartUpload`: new `checksumType` / `expectedChecksum`
  params; validates the client's claimed checksum against the actual digest of
  the reassembled bytes before storing, throwing the same `BadDigest` error
  Floci already uses for PutObject/UploadPart checksum mismatches.
- Scoped to `ChecksumType=FULL_OBJECT` only. Floci's multipart completion
  always reports `ChecksumType=COMPOSITE`, but computes the value as a plain
  full-object hash rather than a true `crc32_combine`-style composite — so
  validating COMPOSITE-type requests against that simplified computation risks
  false-positive rejections. That's a separate, pre-existing gap, flagged as
  follow-up rather than folded into this fix.

### AWS Compatibility

Verified error code/message against AWS's documented behavior (S3
`CompleteMultipartUpload` with `ChecksumType=FULL_OBJECT` returns `BadDigest`
on mismatch) and against Floci's own existing convention for the same error on
PutObject/UploadPart (`S3Controller.validateChecksumHeaders`).

Full-object checksum support for multipart uploads is restricted to CRC-based algorithms (CRC64NVME, CRC32, CRC32C); SHA-1 and SHA-256 are composite-only, per the AWS S3 User Guide's "Full object and composite checksum types" table and "Using full object checksums for multipart upload" section (https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html#Full-object-checksums), which states this restriction exists because only CRC algorithms can linearize part-level checksums into a full-object checksum.

### How it was tested

Rebased onto current `main` (post-#2151, which also touched multipart completion);
all figures below are from that rebased branch.

- Full suite (`./mvnw test`): **8540 run, 0 failures, 0 errors, 9 skipped**. BUILD SUCCESS.
- S3 subset (`-Dtest='S3*Test'`): 689 run, 0 failures, 0 errors.
- New tests in `S3MultipartIntegrationTest`:
  - `completeMultipartUploadRejectsMismatchedFullObjectChecksum` => wrong CRC32 => 400 BadDigest.
  - `completeMultipartUploadAcceptsMatchingFullObjectChecksum` => correct CRC32 => 200.
  - `completeMultipartUploadRejectsFullObjectShaChecksum` => SHA256 + FULL_OBJECT => 400 InvalidRequest.
  - `completeMultipartUploadReportsFullObjectChecksumType` => validated request reports
    `<ChecksumType>FULL_OBJECT</ChecksumType>` via GetObjectAttributes.
- `make docs-check` passes (no handler action registration changed).
- **Manually verified live** against `./mvnw quarkus:dev` using the AWS CLI, reproducing
  the issue's exact repro steps end-to-end:
  - `create-multipart-upload --checksum-type FULL_OBJECT --checksum-algorithm CRC32` => `upload-part` =>
    `complete-multipart-upload --checksum-crc32 "AAAAAA=="` (the issue's own wrong value) =>
    now returns `An error occurred (BadDigest) ... The CRC32 checksum you specified did not match the payload.`
    (previously silently returned 200).
  - Same upload completed with the correct CRC32 => 200, object stored, downloaded bytes
    verified byte-for-byte identical to the uploaded part.
  - A separate multipart upload completed with **no** checksum headers at all still
    succeeds (200), confirms the fix is additive and doesn't disturb existing callers.

### Rebase note

Rebased onto `main` to pick up #2151 (`fix(s3): Ensure tags are added on multipart s3 uploads`),
which touched the same three files. Resolution preserved both changes:

- `S3Service.completeMultipartUpload` now performs the FULL_OBJECT checksum validation
  **and** #2151's `.withTagging(upload.getTagging())` propagation.
- `S3MultipartIntegrationTest`: #2151's tagging tests keep `@Order(18)`/`@Order(19)`;
  this PR's four tests are renumbered to `@Order(20)`–`@Order(23)`, with `cleanUp` last.
- `S3MultipartServiceTest`: #2151's two new tests called the 4-arg
  `completeMultipartUpload`, which this PR widens to 6 args — both call sites updated to
  pass `null, null`, matching the existing call sites in that file.

### Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
@@ -1463,7 +1463,8 @@ public class S3Service implements Resettable {
                 sseCustomerHeaders.algorithm(), sseCustomerHeaders.key(), sseCustomerHeaders.keyMd5());
     }
 
-    public S3Object completeMultipartUpload(String bucket, String key, String uploadId, List<Integer> partNumbers) {
+    public S3Object completeMultipartUpload(String bucket, String key, String uploadId, List<Integer> partNumbers,
+                                            String checksumType, S3Checksum expectedChecksum) {
         MultipartUpload upload = multipartUploads.get(uploadId);
         if (upload == null || !upload.getBucket().equals(bucket) || !upload.getKey().equals(key)) {
             throw new AwsException("NoSuchUpload",
@@ -1493,6 +1494,11 @@ public class S3Service implements Resettable {
             }
 
             byte[] allData = combined.toByteArray();
+
+            if ("FULL_OBJECT".equalsIgnoreCase(checksumType) && expectedChecksum != null
+                    && expectedChecksum.hasAnyValue()) {
+                validateFullObjectChecksum(allData, expectedChecksum);
+            }
 
             // Composite ETag: MD5 of concatenated part MD5s, suffixed with part count
             String compositeETag = "\"" + bytesToHex(md.digest()) + "-" + partNumbers.size() + "\"";
@@ -2396,6 +2402,25 @@ public class S3Service implements Resettable {
             throw new AwsException("InvalidRequest", "The checksum algorithm you specified is a valid AWS checksum algorithm, but is not currently supported by Floci (supported: CRC32, CRC32C, CRC64NVME, SHA1, SHA256).", 400);
         }
         throw new AwsException("InvalidArgument", "The checksum algorithm you specified is not supported.", 400);
+    }
+
+    private static void validateFullObjectChecksum(byte[] data, S3Checksum expected) {
+        if (expected.getChecksumSHA1() != null && !expected.getChecksumSHA1().equals(S3Checksum.sha1Base64(data))) {
+            throw new AwsException("BadDigest", "The SHA1 checksum you specified did not match the payload.", 400);
+        }
+        if (expected.getChecksumSHA256() != null && !expected.getChecksumSHA256().equals(S3Checksum.sha256Base64(data))) {
+            throw new AwsException("BadDigest", "The SHA256 checksum you specified did not match the payload.", 400);
+        }
+        if (expected.getChecksumCRC32() != null && !expected.getChecksumCRC32().equals(S3Checksum.crc32Base64(data))) {
+            throw new AwsException("BadDigest", "The CRC32 checksum you specified did not match the payload.", 400);
+        }
+        if (expected.getChecksumCRC32C() != null && !expected.getChecksumCRC32C().equals(S3Checksum.crc32cBase64(data))) {
+            throw new AwsException("BadDigest", "The CRC32C checksum you specified did not match the payload.", 400);
+        }
+        if (expected.getChecksumCRC64NVME() != null
+                && !expected.getChecksumCRC64NVME().equals(S3Checksum.crc64NvmeBase64(data))) {
+            throw new AwsException("BadDigest", "The CRC64NVME checksum you specified did not match the payload.", 400);
+        }
     }
 
     private static S3Checksum buildChecksum(byte[] data, List<Part> parts, boolean multipartUpload) {
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
@@ -1495,8 +1495,9 @@ public class S3Service implements Resettable {
 
             byte[] allData = combined.toByteArray();
 
-            if ("FULL_OBJECT".equalsIgnoreCase(checksumType) && expectedChecksum != null
-                    && expectedChecksum.hasAnyValue()) {
+            boolean fullObjectChecksumRequested = "FULL_OBJECT".equalsIgnoreCase(checksumType)
+                    && expectedChecksum != null && expectedChecksum.hasAnyValue();
+            if (fullObjectChecksumRequested) {
                 validateFullObjectChecksum(allData, expectedChecksum);
             }
 
@@ -1507,6 +1508,9 @@ public class S3Service implements Resettable {
                     .map(num -> copyPart(upload.getParts().get(num)))
                     .toList();
             S3Checksum checksum = buildChecksum(allData, completedParts, true, upload.getChecksumAlgorithm());
+            if (fullObjectChecksumRequested) {
+                checksum.setChecksumType("FULL_OBJECT");
+            }
             S3Object object = storeObject(bucket, key, allData, upload.getContentType(), upload.getMetadata(),
                     checksum, completedParts,
                     new PutObjectOptions()
@@ -2405,11 +2409,11 @@ public class S3Service implements Resettable {
     }
 
     private static void validateFullObjectChecksum(byte[] data, S3Checksum expected) {
-        if (expected.getChecksumSHA1() != null && !expected.getChecksumSHA1().equals(S3Checksum.sha1Base64(data))) {
-            throw new AwsException("BadDigest", "The SHA1 checksum you specified did not match the payload.", 400);
-        }
-        if (expected.getChecksumSHA256() != null && !expected.getChecksumSHA256().equals(S3Checksum.sha256Base64(data))) {
-            throw new AwsException("BadDigest", "The SHA256 checksum you specified did not match the payload.", 400);
+        if (expected.getChecksumSHA1() != null || expected.getChecksumSHA256() != null) {
+            throw new AwsException("InvalidRequest",
+                    "The FULL_OBJECT checksum type is not supported with the SHA1 or SHA256 checksum algorithm. "
+                            + "Full object checksums are only supported with the CRC32, CRC32C, and CRC64NVME checksum algorithms.",
+                    400);
         }
         if (expected.getChecksumCRC32() != null && !expected.getChecksumCRC32().equals(S3Checksum.crc32Base64(data))) {
             throw new AwsException("BadDigest", "The CRC32 checksum you specified did not match the payload.", 400);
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
```diff
@@ -2384,6 +2390,25 @@ public static String validateAndNormalizeChecksumAlgorithm(String algorithm) {
...
+            throw new AwsException("BadDigest", "The CRC32 checksum you specified did not match the payload.", 400);
+        }
+        if (expected.getChecksumCRC32C() != null && !expected.getChecksumCRC32C().equals(S3Checksum.crc32cBase64(data))) {
+            throw new AwsException("BadDigest", "The CRC32C checksum you specified did not match the payload.", 400);
+        }
+        if (expected.getChecksumCRC64NVME() != null
+                && !expected.getChecksumCRC64NVME().equals(S3Checksum.crc64NvmeBase64(data))) {
+            throw new AwsException("BadDigest", "The CRC64NVME checksum you specified did not match the payload.", 400);
+        }
+    }
```
pgermosen, line 2428: **SHA1 and SHA256 can't carry a FULL_OBJECT checksum on a multipart upload — real S3 rejects the combination outright.**

I checked this against the S3 user guide's multipart table, which is explicit:

| Checksum algorithm | Full object | Composite |
|---|---|---|
| CRC64NVME | Yes | No |
| CRC32 | Yes | Yes |
| CRC32C | Yes | Yes |
| **SHA1** | **No** | Yes |
| **SHA256** | **No** | Yes |

and states the reason directly: *"Full object checksums in multipart uploads are only available for CRC-based checksums because they can linearize into a full object checksum... This type of validation isn't available for other algorithms, such as SHA and MD5."*

So the `getChecksumSHA1()` and `getChecksumSHA256()` branches validate a request shape S3 never accepts — a client sending `ChecksumType=FULL_OBJECT` with `x-amz-checksum-sha256` gets `InvalidRequest` from AWS, whereas Floci will now happily compute a full-object SHA-256 and compare it.

It's the permissive direction rather than a wrong answer, so nothing breaks — but it's the kind of divergence that lets a bug through locally and fails in production, which is the exact scenario the PR description gives as motivation. Two options, either fine:

- Drop the two SHA branches, so only the CRC algorithms are validated under FULL_OBJECT; or
- Keep them but reject the combination up front with `InvalidRequest`, which is strictly closer to AWS and also documents the constraint in code.

The CRC32 / CRC32C / CRC64NVME branches are all correct as written.

src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
```diff
@@ -1481,6 +1482,11 @@ public S3Object completeMultipartUpload(String bucket, String key, String upload
 
             byte[] allData = combined.toByteArray();
 
+            if ("FULL_OBJECT".equalsIgnoreCase(checksumType) && expectedChecksum != null
+                    && expectedChecksum.hasAnyValue()) {
+                validateFullObjectChecksum(allData, expectedChecksum);
+            }
```
pgermosen: **After a FULL_OBJECT checksum validates, the response still reports `ChecksumType: COMPOSITE`.**

`buildChecksum` (`:2425-2427`) decides the type purely from whether this was a multipart upload:

```java
checksum.setChecksumType(multipartUpload || (parts != null && parts.size() > 1)
        ? "COMPOSITE"
        : "FULL_OBJECT");
```

so a client that asked for `FULL_OBJECT`, and whose checksum you just verified, gets `COMPOSITE` back. On real S3 that response echoes `FULL_OBJECT` along with the full-object value.

The interesting part is that **the value is already right**. As your description notes, Floci computes the multipart checksum as a plain full-object hash rather than a true `crc32_combine`-style composite — so for a `FULL_OBJECT` request the number Floci returns is exactly the full-object checksum AWS would return, and only the label contradicts it. That makes this a two-line completion of the path you've just added rather than new scope: thread the requested type through and set it when the client asked for `FULL_OBJECT` and validation passed.

pgermosen (approved):
Looks good.

Well-scoped fix with an unusually careful description — the live AWS CLI reproduction, and specifically the third case (a multipart upload with no checksum headers still returning 200, confirming the change is additive), is the kind of evidence that makes a review fast.

Two comments inline, both non-blocking: SHA1/SHA256 aren't valid FULL_OBJECT algorithms on a multipart upload per AWS's own table, and the completion response still reports COMPOSITE even after a FULL_OBJECT checksum validates.

On the failing check — it isn't yours. Greet on first issue or PR is failing because this branch predates 2bea2c0a (https://github.com/floci-io/floci/pull/1981), which fixed that workflow passing v1 input names (repo-token) to first-interaction@v3, which expects repo_token. I confirmed the ancestry: the fix is not in this branch's base. Rebasing on current main will clear it, and it says nothing about your change.

One small test-hygiene nit, not worth its own comment: completeMultipartUploadRejectsMismatchedFullObjectChecksum leaves its multipart upload open, since the exception is thrown before the cleanup path, and cleanUp deletes the matching test's object but doesn't abort that upload. Harmless today — no object is created and nothing currently asserts on ListMultipartUploads after @Order(18) — but it would surprise whoever adds such an assertion later.

hectorvent:
Thank you for this, and sorry for the wait. One correction to the check status first: the only red entry on the current head is the greeter workflow, the actual build and test checks are green 13 of 13, so nothing is failing on your side.

**(blocking)** Could you add a line to the description stating which AWS `CompleteMultipartUpload` FULL_OBJECT checksum behaviour you matched and where it came from (the S3 docs page or SDK source you followed)? That grounding is what the review needs to verify the validation against the wire contract, and it is the kind of detail nobody can re-derive later.

Holding only for that description line; once it is in, the review should be quick.

avison9:
Thanks @hectorvent, i have added the grounding line to the description: the FULL_OBJECT validation is checked against the AWS S3 User Guide's "Full object and composite checksum types" table and the "Using full object checksums for multipart upload" section (.../checking-object-integrity-upload.html#Full-object-checksums), which states CRC64NVME/CRC32/CRC32C are the only algorithms that support full-object checksums on multipart uploads (they linearize; SHA/MD5 don't). Apologies @pgermosen, i missed the notification in my mail, thanks for the review i can add the 2 SHA issues you flagged inline, before merge also

hectorvent (approved):
Thanks @avison9

Issue #1938, opened 2026-07-24
[FEAT] [Feature] IAM: Add GetAccountSummary and ListEntitiesForPolicy support


SandeepPadhy:
## Service
  
  IAM (Identity and Access Management)
  
  ## API Action / Feature
  
  1. `GetAccountSummary` — Returns IAM entity usage and quota counts
  2. `ListEntitiesForPolicy` — Lists all users, groups, and roles attached to a specified managed policy
  
  ## AWS Documentation
  
  - https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetAccountSummary.html
  - https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListEntitiesForPolicy.html
  
  ## Why is this needed?
  
  Using Floci as an AWS IAM mock for Saviynt (identity governance platform) connector perf testing. During AWS account imports, the Saviynt connector calls both APIs:
  
  - `GetAccountSummary` is called after account import to retrieve IAM usage metadata
  - `ListEntitiesForPolicy` is called during policy/role import to map policy-to-entity relationships
  
  Both currently return `UnsupportedOperation` (Status 400, Error Code: UnsupportedOperation) on Floci 1.5.33.
  
  The import partially succeeds (accounts, users, groups, roles, policies all import fine), but these two calls cause error entries in the job results. Other emulators (Moto, fakecloud) support both operations.
  
  ## Are you willing to contribute a PR?
  
  - [ ] Yes
  - [x] No

Mide69:
Hello @hectorvent, I'd like to work on this one, adding support for GetAccountSummary and ListEntitiesForPolicy to IAM. 
Could you assign it to me, please? 

Thanks!

hectorvent:
Hi @Mide69, 

I see there see the #1808 open for this issue. Could you review it and help test it?

Thanks

Mide69:
@hectorvent Yes sure, I can do that.

Mide69:
Hey @hectorvent, I took a look at #1808 and actually ran the tests locally (had to grab JDK 25 first). Both tests pass, clean build. Code looks good too, follows the same pattern as `ListPolicies`, handles the unknown-ARN case properly with `NoSuchEntity`. LGTM from me.

One thing though, this PR only covers `ListEntitiesForPolicy`. `GetAccountSummary` from this issue is still open. Want me to take that one on as its own PR or look for another issue to work on?

SandeepPadhy:
[...]

PR #1986, opened 2026-07-27
feat(iam): add GetAccountSummary support

Mide69:
**Title: feat(iam): add GetAccountSummary support**

**Part of #1938** - this PR implements GetAccountSummary. The other requested action, ListEntitiesForPolicy, is implemented separately in #1808. #1938 should only close once both PRs have merged.

Adds the `GetAccountSummary` IAM query action - currently falls through to `UnsupportedOperation`. Returns all 34 documented `SummaryMap` fields:

**- Live counts:** `Users`, `Groups`, `Roles`, `Policies` (customer-managed only), `InstanceProfiles`, `PolicyVersionsInUse`
**- Quota values:** cross-checked against [[AWS's published IAM service quotas](https://docs.aws.amazon.com/general/latest/gr/iam-service.html)](https://docs.aws.amazon.com/general/latest/gr/iam-service.html) and independently verified against moto's ``implementation; every value matches both sources exactly
- Untracked resources (MFA devices, SAML/OIDC providers, server certificates, account passwords): report `0` rather than being omitted, so callers indexing into the full AWS field set don't hit a missing-key error, the same class of bug as the CloudFront `OriginGroups` issue elsewhere in this repo

**Testing:**
- New `IamGetAccountSummaryIntegrationTest`: verifies counts increase correctly via before/after deltas (IAM state persists across this test suite), static quota values, and a completeness check asserting exactly 34 fields
- Full IAM package: 181/181 passing, no regressions
- Manually verified end-to-end against the live dev server via AWS CLI:

image

### DIFF
--- a/docs/services/iam.md
+++ b/docs/services/iam.md
@@ -141,6 +141,12 @@
 |--------|-------------|
 | SimulatePrincipalPolicy | Evaluates requested actions and resources against the resolved principal's policies. |
 
+### Account
+
+| Action | Description |
+|--------|-------------|
+| GetAccountSummary | Returns entity counts (users, groups, roles, customer-managed policies, instance profiles) and IAM quota values. Resources Floci does not track (MFA devices, SAML/OIDC providers, server certificates) are reported as zero rather than omitted. |
+
 ## AWS Managed Policies
 
 Floci seeds a catalog of commonly-used AWS managed policies at startup. These are attachable immediately without any setup:
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamQueryHandler.java
@@ -93,6 +93,7 @@ public class IamQueryHandler {
             case "DeletePolicy" -> handleDeletePolicy(params);
             case "ListPolicies" -> handleListPolicies(params);
             case "ListEntitiesForPolicy" -> handleListEntitiesForPolicy(params);
+            case "GetAccountSummary" -> handleGetAccountSummary(params);
             case "CreatePolicyVersion" -> handleCreatePolicyVersion(params);
             case "GetPolicyVersion" -> handleGetPolicyVersion(params);
             case "DeletePolicyVersion" -> handleDeletePolicyVersion(params);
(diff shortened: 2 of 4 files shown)


pgermosen (approved):
Looks good.


- **Wire format**: `envelope("GetAccountSummary", AwsNamespaces.IAM, ...)` produces the correct `<GetAccountSummaryResponse><GetAccountSummaryResult><SummaryMap>...` shape, and the `entry`/`key`/`value` map serialization matches the exact idiom already used for Query-protocol maps elsewhere (e.g. `SnsQueryHandler.java:243`), so this isn't a one-off pattern.
- **Field list**: cross-checked all 34 `SummaryMap` keys and their default quota values against AWS's published IAM service quotas from memory — names and values match (e.g. `GroupsQuota=300`, `RolesQuota=1000`, `VersionsPerPolicyQuota=5` tied to the actual 5-version cap enforced in `createPolicyVersion`).
- **Customer-managed policy counting**: `getAccountSummary()`'s `!arn.startsWith(AwsManagedPolicies.ARN_PREFIX)` filter for `Policies`/`PolicyVersionsInUse` mirrors the identical, already-established filter in `listPolicies`'s `Local` scope branch — consistent with the rest of the class, not a new ad hoc rule.
- **Docs**: `docs/services/iam.md` was updated in this PR (new `### Account` section), and `make docs-check` passes clean on this branch — unlike a docs gap I found on a previous PR, this one didn't skip it.
- **Tests**: `IamGetAccountSummaryIntegrationTest` (3/3) plus the full `io.github.hectorvent.floci.services.iam` package (all classes, 180 tests total in this worktree) pass with zero failures. The before/after-delta test design correctly accounts for IAM state persisting across the suite instead of asserting brittle absolute counts.

hectorvent:
Thank you for adding GetAccountSummary.

**(blocking)** The branch now conflicts with main, so it needs a rebase before it can merge. CI on the current head is otherwise green.

One coordination note: issue #1938 asks for both `GetAccountSummary` and `ListEntitiesForPolicy`. This PR covers the first and #1808 covers the second, and neither currently carries a formal closing link, which is the safe state, since either one closing the issue would leave the other half orphaned. Worth linking both PRs to #1938 explicitly so it closes only when both have landed.

The rebase is the only blocker.

Mide69:
Thanks for catching that @hectorvent  - rebased onto main to resolve the conflict, and updated the description to explicitly cross-reference #1808 (ListEntitiesForPolicy) so #1938 only closes once both PRs land.

PR #1995, opened 2026-07-27
feat(cloudformation): provision AWS::EC2::VPCEndpoint

### LINKED ISSUE #1994: [FEAT] CloudFormation: provision AWS::EC2::VPCEndpoint
## Service

CloudFormation / EC2

## API Action / Feature

Provision `AWS::EC2::VPCEndpoint`. `Ec2Service.createVpcEndpoint` (route tables, subnets, SGs, private DNS, tags) and delete already exist; the CFN provisioner has no case, so templates declaring endpoints skip them.

## AWS Documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpcendpoint.html

## Why is this needed?

Supports running the [aws-bench](https://github.com/aws-bench/aws-bench) benchmark scenarios against Floci: five of the eight scenario CDK apps (api-and-observability, databases-and-storage, reference-architectures, streaming-and-iot, troubleshooting-multiservice) declare VPC endpoints. Part of the gap series tracked in lex00/floci#17.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #1994. `Ec2Service` already models VPC endpoints fully; this adds the CFN case (VpcId, ServiceName, VpcEndpointType, RouteTableIds/SubnetIds/SecurityGroupIds reference lists, PrivateDnsEnabled) plus stack-delete removal, and a shared `resolveIdList` helper for Ref-array properties.

Context: supports running the [aws-bench](https://github.com/aws-bench/aws-bench) scenarios against Floci — five of eight scenario CDK apps declare VPC endpoints (series: lex00/floci#17).

## Tests

New `CloudFormationVpcEndpointIntegrationTest`: gateway endpoint with a route-table ref reaches CREATE_COMPLETE, appears in `DescribeVpcEndpoints`, and is gone after DeleteStack.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -331,6 +331,8 @@ public class CloudFormationResourceProvisioner {
                 case "AWS::EC2::Subnet" -> provisionSubnet(resource, properties, engine, region);
                 case "AWS::EC2::SecurityGroup" -> provisionSecurityGroup(resource, properties, engine, region, stackName);
                 case "AWS::EC2::InternetGateway" -> provisionInternetGateway(resource, region);
+                case "AWS::EC2::VPCEndpoint" ->
+                        provisionVpcEndpoint(resource, properties, engine, region);
                 case "AWS::EC2::RouteTable" -> provisionRouteTable(resource, properties, engine, region);
                 case "AWS::EC2::SubnetRouteTableAssociation" ->
                         provisionSubnetRouteTableAssociation(resource, properties, engine, region);
@@ -454,6 +456,7 @@ public class CloudFormationResourceProvisioner {
             case "AWS::ElasticLoadBalancingV2::ListenerRule" -> elbV2Service.deleteRule(region, physicalId);
             case "AWS::KinesisFirehose::DeliveryStream" -> firehoseService.deleteDeliveryStream(physicalId);
             case "AWS::EC2::SecurityGroup" -> ec2Service.deleteSecurityGroup(region, physicalId);
+            case "AWS::EC2::VPCEndpoint" -> ec2Service.deleteVpcEndpoints(region, List.of(physicalId));
             case "AWS::EC2::Instance" -> ec2Service.terminateInstances(region, List.of(physicalId));
             case "AWS::RDS::DBInstance" -> rdsService.deleteDbInstance(physicalId);
             case "AWS::RDS::DBCluster" -> rdsService.deleteDbCluster(physicalId);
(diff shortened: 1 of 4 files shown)


hectorvent:

lex00:
Done. Rebased on `upstream/main` and moved to the per-service structure at `8e1cbc3`.

`AWS::EC2::VPCEndpoint` is now `Ec2VpcEndpointCfnProvisioner`.

hectorvent:
Thank you for this, and I am sorry it sat this long with no reply from us. The replacement handling is the part I want to credit: capturing the previous endpoint id before creating the new one and deleting it afterwards is exactly right, and tolerating `InvalidVpcEndpointId.NotFound` on that cleanup is a good touch.

**(blocking)** The registry schema lists four `Fn::GetAtt` attributes (`Id`, `CreationTimestamp`, `NetworkInterfaceIds`, `DnsEntries`) and only `Id` is set. `CreationTimestamp` is already on the model since `Ec2Service` stamps it at create, so it is one line, plus an assertion in `gatewayEndpointSetsPhysicalIdAndResolvesRefs`. It matters because an unset attribute resolves to the literal `"LogicalId.CreationTimestamp"` rather than failing, so a template gets a wrong value quietly. `NetworkInterfaceIds` and `DnsEntries` are not modelled yet, so leaving those is fine.

**(follow-up, separate PR)** Provision replaces unconditionally, so a stack update that changes nothing about the endpoint still swaps its id.

That one attribute is all this needs.

lex00:
Done in ac9bf2db. CreationTimestamp is set from the value Ec2Service stamps at create, formatted as DescribeVpcEndpoints formats it, with an assertion in gatewayEndpointSetsPhysicalIdAndResolvesRefs. NetworkInterfaceIds and DnsEntries stay unset.

Agreed on the unconditional replacement, leaving it out of this one.

pgermosen (approved):
@lex00 — approving. @hectorvent's blocking point is closed, and you did exactly what was asked without expanding past it.

## The fix

The subtlety in Hector's finding is worth restating, because it is the reason a missing attribute is worse than a failing one: an unset `Fn::GetAtt` resolves to the literal string `"LogicalId.CreationTimestamp"` rather than erroring, so a template silently receives a wrong value and carries it wherever it is referenced.

`ac9bf2db` sets it, and I checked the format claim rather than take it on trust. The provisioner's `ISO_FMT` is byte-identical to the one `Ec2QueryHandler` uses for `DescribeVpcEndpoints` — same `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` pattern, same `ZoneOffset.UTC` — so `Fn::GetAtt CreationTimestamp` and the describe response will always agree. That matters more than it looks: an attribute that is present but formatted differently from the wire is a harder bug to find than one that is absent.

## Verified

On current `main` (`30a61646`): **839 tests green** across `services.cloudformation.**` and `services.ec2.**`. Nothing has landed that overlaps — no other `AWS::EC2::VPCEndpoint` provisioning and no competing `resolveIdList`.

Both behaviours are pinned:

| Reverted | Failure |
|---|---|
| the `CreationTimestamp` attribute | `gatewayEndpointSetsPhysicalIdAndResolvesRefs` — `expected: <2026-08-13T10:15:30.000Z> but was: <null>` |
| the replaced-endpoint cleanup | `updateReplacesAndDeletesPreviousEndpoint` |

The first is the exact test Hector asked for the assertion in, and it fails with the exact symptom he described.

## Scope

You held the line where he drew it — `NetworkInterfaceIds` and `DnsEntries` left unset because they are not modelled, and the unconditional-replacement issue left for its own PR. Resisting the urge to fold a known adjacent bug into a PR that is one line from done is the harder discipline, and it is the reason this one is mergeable tonight.

Seconding his credit on the replacement handling, too: capturing the previous endpoint id before creating the replacement and deleting it afterwards is the right order, and tolerating `InvalidVpcEndpointId.NotFound` on that cleanup means a partially-applied prior update cannot wedge the stack.

## One observation for the follow-up

The unconditional-replacement issue Hector deferred here is the same shape as the one you fixed tonight in #1997, where `provisionVersion` republished on every update and repointed the `Ref`s. Both come from a provisioner not consulting its existing physical id before acting.

`existingVersionIsCurrent` in #1997 is a good template for what the VPC endpoint version of that check wants to look like — verify the recorded id still exists *and* still belongs to the same parent, then keep it. Might be worth one issue covering the pattern across provisioners rather than a separate follow-up per resource type, since any provisioner added from here will have the same question to answer.

Thank you for the patience on this one, and for the aws-bench context — knowing which scenarios need these resource types makes it much easier to judge what is worth prioritising.

PR #1997, opened 2026-07-27
feat(cloudformation): provision AWS::Lambda::Permission and fix Version and Alias updates

### LINKED ISSUE #1996: [FEAT] CloudFormation: provision AWS::Lambda::Permission, Version, and Alias
## Service

CloudFormation / Lambda

## API Action / Feature

Provision the Lambda addressing resources: `AWS::Lambda::Permission` (resource-based policy statements), `AWS::Lambda::Version` (publish), and `AWS::Lambda::Alias`. `LambdaService.addPermission/publishVersion/createAlias` all exist; the CFN provisioner has no cases, so templates carrying them (every CDK app wiring a Lambda to S3/SNS/API Gateway emits Permission) silently skip them.

## AWS Documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-permission.html

## Why is this needed?

Supports running the [aws-bench](https://github.com/aws-bench/aws-bench) benchmark scenarios against Floci: `Lambda::Permission` appears in six of the eight scenario CDK apps, `Version`/`Alias` in two. Part of the gap series tracked in lex00/floci#17.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #1996. The three Lambda addressing resources now map onto the existing `LambdaService` APIs — `Permission` → `addPermission` (statement id = logical id), `Version` → `publishVersion` (`Ref` = version-qualified ARN; `Version`/`FunctionArn` attributes), `Alias` → `createAlias` (`Ref` = alias ARN, defaulting to `$LATEST`).

Context: supports running the [aws-bench](https://github.com/aws-bench/aws-bench) scenarios against Floci — `Lambda::Permission` appears in six of the eight scenario CDK apps (series: lex00/floci#17).

## Tests

New `CloudFormationLambdaAddressingIntegrationTest`: inline-code function + permission + version + alias reach CREATE_COMPLETE; version and alias ARNs resolve in outputs; the permission statement is visible in the function's resource policy. Control-plane only, Docker-free.


hectorvent:

lex00:
Done. Rebased on `upstream/main` and moved to the per-service structure at `724d1c8`.

The Lambda addressing trio is now `LambdaAddressingCfnProvisioner`.

lex00:
Fixed at `bb44d43`. Three of the four findings were against the pre-refactor switch and are already handled by `LambdaAddressingCfnProvisioner` — deletion for all three types, update handling, and `RoutingConfig` weights. The permission-replacement one was live.

AddPermission rejects a duplicate Sid, so the new statement cannot simply be added first. Instead the previous statement is captured before it is removed and put back if AddPermission fails, with the restore failure attached as suppressed so the stack still reports the original error. `LambdaService` gained `restorePermissionStatement` for this, taking the stored statement shape rather than an AddPermission request so what goes back is exactly what came out.

408 tests across the CloudFormation and Lambda suites pass.

lex00:
Updated after #1961 landed. LambdaVersionAliasCfnProvisioner claims Version and Alias, and this claimed them too, so the registry rejected the duplicate and every stack rolled back. Version and Alias provisioning is dropped here in favour of that one.

Two gaps showed up once the integration test ran against it. provisionVersion published unconditionally, so every update minted a new version and repointed the Refs. RoutingConfig was not carried over, and the CloudFormation path does emit it. Both fixed in 53a635ad.

What is left of the original scope is AWS::Lambda::Permission.

pgermosen (approved):
@lex00 — approving. The rescope after #1961 landed was the right call, and the two bugs you found in it while testing against it are both real. I confirmed them in `main` rather than take them on report.

## The collision, and how you handled it

You hit it the hard way: both provisioners claimed `AWS::Lambda::Version` and `AWS::Lambda::Alias`, the registry rejected the duplicate, and every stack rolled back. Dropping your side in favour of the one that had already merged is exactly right — the alternative would have been two implementations of the same resource types with merge order deciding which won.

Worth saying plainly: that is the second time today a PR has been overtaken by another implementing the same thing. You caught yours and re-scoped inside a few hours. That is the outcome we want, and it is only possible because you actually ran the integration test against merged `main` rather than assuming a clean textual merge meant a clean result.

## The two bugs in #1961 are genuine

I verified both against `main` rather than trusting the description.

**`provisionVersion` published unconditionally.** Confirmed — `main` calls `publishVersion` with no check for an existing physical id, so any unrelated stack update mints a new Lambda version and repoints every `Ref` to it. `AWS::Lambda::Version` is immutable in real CloudFormation; an unchanged Version resource should not be republished.

**`RoutingConfig` was not carried over**, despite the CloudFormation path emitting it.

Both are pinned. Reverting either fails `updateKeepsVersionAppliesRoutingAndDeleteCleansUp`:

| Reverted | Result |
|---|---|
| the version-immutability guard | that test fails |
| `parseRoutingConfig` → `null` | that test fails |

Your fix for the first is more careful than it needed to be, in the right way: `existingVersionIsCurrent` keeps the prior version only when it is *still published* **and** its base ARN still matches the function, so repointing `FunctionName` at a different function correctly republishes rather than silently keeping a stale ARN. Falling through to publish on `AwsException` is the right failure direction too.

The alias path deserves a mention as well — documenting that an absent `RoutingConfig` clears previous weights because an empty map clears in `updateAlias` while null preserves them. That distinction is wrong about half the time when it is left implicit, and the comment is what stops the next person "simplifying" it.

## Verified

On current `main` (`30a61646`): **1025 tests green** across `services.cloudformation.**` and `services.lambda.**`.

The permission-replacement handling from the earlier round also reads well: capturing the previous statement before removing it, restoring it if `addPermission` fails, and attaching the restore failure as suppressed so the stack still reports the original error. That is the correct shape for an operation with no atomic replace, and getting the suppressed-exception detail right means the real cause is not masked by a cleanup failure.

## One note, on my side

I approved #1961 earlier today and did not catch the unconditional publish. My mutation testing there exercised the create path and the SAM transform but never a stack *update* against the new provisioner, which is exactly where this surfaces. Your integration test covering create-then-update is what found it, and it is a better test than the one I reviewed.

Thank you for the persistence on this one — six weeks, a structural refactor to the per-provisioner pattern, and a rescope around someone else's merge.

PR #1999, opened 2026-07-27
feat(cloudformation): provision AWS::ECS::CapacityProvider and ClusterCapacityProviderAssociations

### LINKED ISSUE #1998: [FEAT] CloudFormation: provision AWS::ECS::CapacityProvider and ClusterCapacityProviderAssociations
## Service

CloudFormation / ECS

## API Action / Feature

Provision `AWS::ECS::CapacityProvider` and `AWS::ECS::ClusterCapacityProviderAssociations`. `EcsService.createCapacityProvider` and `putClusterCapacityProviders` exist; the CFN provisioner has no cases, so CDK clusters with ASG capacity providers (and the associations resource every `ecs.Cluster` with providers emits) deploy without them.

## AWS Documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-capacityprovider.html

## Why is this needed?

Supports running the [aws-bench](https://github.com/aws-bench/aws-bench) benchmark scenarios against Floci: `ClusterCapacityProviderAssociations` appears in six of the eight scenario CDK apps. Part of the gap series tracked in lex00/floci#17.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #1998. Both resources map onto existing `EcsService` APIs (`createCapacityProvider`, `putClusterCapacityProviders`); the associations resource resolves its provider Ref list and default strategy entries (weight/base).

Context: supports running the [aws-bench](https://github.com/aws-bench/aws-bench) scenarios against Floci — `ClusterCapacityProviderAssociations` appears in six of the eight scenario CDK apps (series: lex00/floci#17).

## Tests

New `CloudFormationEcsCapacityIntegrationTest`: cluster + ASG-backed provider + associations reach CREATE_COMPLETE and `DescribeClusters` reports the provider on the cluster.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -331,10 +331,6 @@ public class CloudFormationResourceProvisioner {
                 case "AWS::CloudFormation::CustomResource" ->
                         provisionCustomResource(resource, properties, engine, region, accountId, stackName);
                 case "AWS::ECS::Cluster" -> provisionEcsCluster(resource, properties, engine, region, stackName);
-                case "AWS::ECS::CapacityProvider" ->
-                        provisionEcsCapacityProvider(resource, properties, engine, region, stackName);
-                case "AWS::ECS::ClusterCapacityProviderAssociations" ->
-                        provisionEcsCapacityAssociations(resource, properties, engine, region);
                 case "AWS::ECS::TaskDefinition" -> provisionEcsTaskDefinition(resource, properties, engine, region, stackName);
                 case "AWS::ECS::Service" -> provisionEcsService(resource, properties, engine, region, stackName);
                 case "AWS::ElasticLoadBalancingV2::LoadBalancer" ->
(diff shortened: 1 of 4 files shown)


hectorvent:

lex00:
Done. Rebased on `upstream/main` and moved to the per-service structure at `e6084c8`.

The ECS capacity pair is now `EcsCapacityCfnProvisioner`.

hectorvent:
Thank you for the rewrite onto the provisioner framework; the full AutoScalingGroupProvider carry-over, Tags, and both delete paths now match CreateCapacityProvider's camelCase shape (https://github.com/boto/botocore/blob/develop/botocore/data/ecs/2014-11-13/service-2.json).

**(blocking)** Stack updates break. UpdateStack re-provisions every resource, and provisionCapacityProvider always calls createCapacityProvider (src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/EcsCapacityCfnProvisioner.java:74-84), which rejects an existing name (src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java:1085), failing the whole update; with no Name declared it instead mints a fresh generated name each pass and orphans the previous provider. When the resource already has a physical id and the provider exists, reuse or update it (EcsService.updateCapacityProvider), replacing only on createOnly changes (Name, AutoScalingGroupArn per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-capacityprovider.html). The Lambda addressing provisioner in #1997 shows the pattern.

Since the aws-bench CDK scenarios redeploy iteratively, this will surface immediately. One blocker from my side.

lex00:
Fixed. An existing provider is reused through updateCapacityProvider, an unnamed one keeps the name its first execution generated, and one removed out of band is recreated rather than failing. Tags are applied separately since updateCapacityProvider carries only the Auto Scaling settings, so a tag the template drops is removed.

For Name and AutoScalingGroupArn I report an unsupported replacement rather than performing one, matching how AWS::IAM::Role handles RoleName. Tell me if you want a real replacement instead.

One thing I hit while testing. `EcsJsonHandler.capacityProviderNode` does not serialize `autoScalingGroupProvider`, so DescribeCapacityProviders returns only name, status, arn and tags. The carry-over this PR does is not observable through the API. I left it alone since it is the ECS handler rather than CFN.

lex00:
Docs Action Tables was cancelled again on the runner acquisition step. This PR changes no handler actions and make docs-check passes locally. Happy to close and reopen to retrigger if that is easier than a rerun.

hectorvent:
@lex00 I just reran it.

hectorvent:
Thank you, that closes it properly, and sorry for the wait. Reusing the existing provider through `updateCapacityProvider`, keeping the generated name across updates, and recreating one that was removed out of band all read correctly, and the unit tests pin each case.

The part worth calling out is the createOnly handling. `Name` and `AutoScalingGroupProvider/AutoScalingGroupArn` are both createOnly in the registry schema, and rejecting a change to either as an unsupported replacement is exactly right rather than a guess. Setting no `Fn::GetAtt` attributes is also correct here, since neither type declares any.

**(follow-up, whenever, and not really yours)** The ECS row in `docs/services/cloudformation.md` still lists only Cluster, TaskDefinition and Service. Worth adding these two, though the recent provisioner PRs have not been updating that table either, so it is a gap on our side more than yours.

No blockers from my side.

hectorvent (approved):

PR #2001, opened 2026-07-27
feat(cloudformation): provision the network ACL family

### LINKED ISSUE #2000: [FEAT] CloudFormation: provision the network ACL family (NetworkAcl, NetworkAclEntry, SubnetNetworkAclAssociation)
## Service

CloudFormation / EC2

## API Action / Feature

Provision `AWS::EC2::NetworkAcl`, `AWS::EC2::NetworkAclEntry`, and `AWS::EC2::SubnetNetworkAclAssociation`. The EC2 service already models the whole family (`createNetworkAcl`, `createNetworkAclEntry`, subnet default associations at `createSubnet`, `replaceNetworkAclAssociation`); only the CFN cases are missing.

## AWS Documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-networkacl.html

## Why is this needed?

Supports running the [aws-bench](https://github.com/aws-bench/aws-bench) benchmark scenarios against Floci: the databases-and-storage and troubleshooting-multiservice scenario CDK apps declare custom NACLs with entries and subnet associations, and network-diagnosis tasks grade against them existing. Part of the gap series tracked in lex00/floci#17.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #2000. Maps `NetworkAcl`, `NetworkAclEntry`, and `SubnetNetworkAclAssociation` onto the existing `Ec2Service` NACL machinery. The association follows CloudFormation semantics — locate the subnet's live association (created with the subnet on its VPC default ACL) and `replaceNetworkAclAssociation` onto the target.

Context: supports running the [aws-bench](https://github.com/aws-bench/aws-bench) scenarios against Floci — databases-and-storage and troubleshooting-multiservice declare custom NACLs, and network-diagnosis tasks grade against them (series: lex00/floci#17).

## Tests

New `CloudFormationNetworkAclIntegrationTest`: VPC + subnet + ACL + deny entry + association reach CREATE_COMPLETE; `DescribeNetworkAcls` shows the entry and the subnet association on the custom ACL.


hectorvent:

lex00:
Done. Rebased on `upstream/main` and moved to the per-service structure at `f12369a`.

The network ACL family is now `Ec2NetworkAclCfnProvisioner`.

hectorvent:
Thank you for this, and I am sorry it sat this long with no reply from us. The delete paths are the strong part: parsing the composite entry id, and reverting a subnet to its VPC's default ACL rather than just dropping the association, are both the right CloudFormation semantics.

**(blocking)** `provisionNetworkAcl` calls `createNetworkAcl` unconditionally. On `UpdateStack`, `CloudFormationService` re-invokes provision with the prior `resource.getPhysicalId()`, so every update mints a second ACL, overwrites the id, and orphans the previous one permanently. Your own `EcsCapacityCfnProvisioner` in #1999 has exactly the shape this wants.

**(blocking)** Fixing that alone breaks the entry path: `createNetworkAclEntry` is called with `replace=false`, which throws `NetworkAclEntryAlreadyExists` for an existing rule. `true` is right, since `RuleNumber`, `Egress` and `NetworkAclId` are the createOnly key, so a same-key re-provision is an update. Worth doing both together.

**(follow-up, separate PR)** `NetworkAclEntry` sets no `Id` attribute, so `Fn::GetAtt` on it returns a literal.

None of this shows up in CI, since no test here runs an UpdateStack.

lex00:
Both done in 16fced47.

provisionNetworkAcl reuses the existing ACL and creates only when it is gone. Also added, VpcId is createOnly so a change now reports an unsupported replacement. Drop that if you want this narrower.

createNetworkAclEntry uses replace=true.

Three tests run an UpdateStack.

Agreed on the NetworkAclEntry Id, leaving it out of this one.

lex00:
Merged current main and added the resource-type table rows for the network ACL family, per step 7 of the CloudFormation section in AGENTS.md.

hectorvent:
Thanks for turning this around so quickly, and for adding the VpcId guard on top.

I checked both fixes against the source rather than just the diff. `provisionNetworkAcl` now reads the prior physical id and creates only when `describeNetworkAcls` comes back empty, which is right: that method filters rather than throwing (`services/ec2/Ec2Service.java:423`), so the removed-out-of-band branch genuinely works. The `replace=true` reasoning holds too, since `Egress`, `NetworkAclId` and `RuleNumber` are exactly the [createOnly set](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-networkaclentry.html), so a same-key re-provision is an in-place update. Please keep the VpcId guard, it matches what `IamRoleCfnProvisioner` and `EcsCapacityCfnProvisioner` already do.

**(follow-up, separate PR)** The entry has no equivalent guard, so changing `RuleNumber` leaves the old rule on the ACL. Metadata only, but it drifts from the template.

No blockers from my side.

hectorvent (approved):

PR #2003, opened 2026-07-27
feat(cloudformation): provision ApiGateway::Account, AutoScaling::LifecycleHook, and EC2::FlowLog

### LINKED ISSUE #2002: [FEAT] CloudFormation: provision ApiGateway::Account, AutoScaling::LifecycleHook, and EC2::FlowLog
## Service

CloudFormation (ApiGateway / AutoScaling / EC2)

## API Action / Feature

Three small metadata provisioners whose service support already exists: `AWS::ApiGateway::Account` (account-level CloudWatch role setting — a no-op record with a physical id), `AWS::AutoScaling::LifecycleHook` (`AutoScalingService.putLifecycleHook`), and `AWS::EC2::FlowLog` (`FlowLogService.createFlowLog`). All three currently make their stacks skip resources.

## AWS Documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-flowlog.html

## Why is this needed?

Supports running the [aws-bench](https://github.com/aws-bench/aws-bench) benchmark scenarios against Floci: `ApiGateway::Account` appears in four scenario CDK apps, `LifecycleHook` and `FlowLog` in three each. Part of the gap series tracked in lex00/floci#17.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #2002. Three small metadata provisioners: `ApiGateway::Account` (account-level setting, metadata record), `AutoScaling::LifecycleHook` → `putLifecycleHook`, `EC2::FlowLog` → `FlowLogService.createFlowLog` (service newly injected into the provisioner; the two positional-null test constructors updated).

Context: supports running the [aws-bench](https://github.com/aws-bench/aws-bench) scenarios against Floci — Account appears in four scenario CDK apps, LifecycleHook and FlowLog in three each (series: lex00/floci#17).

## Tests

New `CloudFormationSmallMetadataIntegrationTest`: ASG + hook + account + VPC flow log reach CREATE_COMPLETE; the hook is visible via `DescribeLifecycleHooks` and the flow log via `DescribeFlowLogs`. CloudFormation package suite green.


hectorvent:

lex00:
Done. Rebased on `upstream/main` and moved to the per-service structure at `724e9f4`.

Split three ways rather than one class: `ApiGatewayAccountCfnProvisioner`, `AutoScalingLifecycleHookCfnProvisioner`, and `Ec2FlowLogCfnProvisioner`.

lex00:
Fixed at `b0eb6a6`. The two delete-dispatch findings were against the pre-refactor switch — both types have delete in their per-service provisioners now. The same-named hooks one was live and needed a small interface change.

A hook name is unique within an Auto Scaling group, not within a region, but the delete path only had the physical id, which is the hook name. `CfnResourceProvisioner` gained a delete overload receiving the whole `StackResource`, defaulting to the existing id-only form, and the dispatcher calls it for extracted provisioners. I put it after the EKS Nodegroup and IAM policy special cases so extracting one of those later cannot silently bypass them.

The hook provisioner records its `AutoScalingGroupName` at create time and deletes through the group-scoped API. Resources provisioned before that fall back to the by-name form — leaving the hook behind is worse, since it can sit on a group that outlives the stack.

310 tests across the CloudFormation and Auto Scaling suites pass.

hectorvent:
Thank you for these, and I am sorry it sat this long with no reply from us. The delete-dispatch hook is a good addition, and putting it after the existing special cases so extracting one of those types later cannot bypass them is the careful choice. `ApiGateway::Account` is right too, including not implementing delete.

**(blocking)** The branch conflicts with main now, in `CloudFormationResourceProvisioner.java`.

**(blocking)** `Ec2FlowLogCfnProvisioner` calls `createFlowLog` unconditionally, so each `UpdateStack` creates a second flow log and orphans the first, which then outlives the stack. That is the exact thing your delete override exists to prevent.

**(blocking)** Same shape in the lifecycle hook, but only when `LifecycleHookName` is omitted: a fresh generated name each update leaves the previous hook on the group. A named hook is fine, since `putLifecycleHook` is idempotent. Your `EcsCapacityCfnProvisioner` in #1999 already solves this exactly, by keeping the name the first execution generated.

None of it surfaces in CI, since no test here runs an `UpdateStack`.

lex00:
All three done in f72d3746, plus 203edf42.

The conflict was the new Events delete cases landing where the dispatch hook goes. Both kept, hook still runs after them.

Ec2FlowLogCfnProvisioner reuses the existing flow log and creates only when it is gone. A change to any of the createOnly properties now reports an unsupported replacement rather than reusing a log whose config no longer matches the template.

The lifecycle hook keeps its first generated name when LifecycleHookName is omitted, and a change to LifecycleHookName or AutoScalingGroupName reports an unsupported replacement for the same reason.

Both provisioners have tests that provision twice.

lex00:
Merged current main and added the resource-type table rows for ApiGateway Account, AutoScaling LifecycleHook and EC2 FlowLog, per step 7 of the CloudFormation section in AGENTS.md.

hectorvent:
Thanks for turning all three around, and apologies: the conflict is on us this time.

I verified both code fixes against the schema and the services rather than just the diff. `Ec2FlowLogCfnProvisioner` now reuses the existing log and creates only when it is gone, and the recreate branch genuinely works, since `describeFlowLogs` filters rather than throwing for an unknown id. Your "every property except Tags is createOnly" reading is exactly right, and the lifecycle hook guard covers both `AutoScalingGroupName` and `LifecycleHookName`, which is precisely the createOnly set. The Java conflict is resolved too, and the dispatch hook still runs after the Events cases.

**(blocking)** The branch still conflicts, but only in `docs/services/cloudformation.md`, and only because #2001 merged a few hours ago and edits the same EC2 table row. Keeping both sides is all it needs: `..., VPCGatewayAttachment, NetworkAcl, NetworkAclEntry, SubnetNetworkAclAssociation, FlowLog`.

That one line is the last thing standing between this and a merge.

lex00:
Merged current main and kept both sides of the EC2 row, so it reads VPCGatewayAttachment, NetworkAcl, NetworkAclEntry, SubnetNetworkAclAssociation, FlowLog. make docs-check passes.

lex00:
Two rounds on the same comparison, and the first was a bad fix.

rejectIfChanged originally skipped whenever the template supplied no value, so dropping a previously explicit LogFormat or MaxAggregationInterval bypassed validation. Removing that early return fixed the drop but broke two other cases, since createFlowLog defaults ResourceType, TrafficType and LogDestinationType. Re-applying an unchanged template then compared null against a stored default and failed a no-op update, and adding a property to a log that never had one still returned early on the stored null.

Both sides are now the effective value, with the requested side defaulted exactly as createFlowLog does. Add, drop and change all read as changes, and a re-apply stays a no-op. The test helper stamps the same three defaults, because a stub without them is not a flow log this provisioner could have produced.

hectorvent:
That is all three cleared, thank you. The conflict is gone, and the lifecycle hook now keeps the name its first execution generated, so an unnamed hook no longer leaves a stray behind on the group.

The two createOnly commits you added on your own are the part worth calling out. Comparing effective values on both sides is the right fix: without it a re-applied unchanged template reads null against a stored default and reports a replacement that never happened. I checked your three constants against `FlowLogService`, which defaults ResourceType to VPC, TrafficType to ALL and LogDestinationType to s3, and they match exactly. Treating a dropped property as a change is right too.

No blockers from my side.

hectorvent (approved):

PR #2007, opened 2026-07-27
feat(ec2): DescribeVpcEndpointServices returns the common interface services with AZs

### LINKED ISSUE #2004: [FEAT] EC2: DescribeVpcEndpointServices returns an empty list, breaking CDK InterfaceVpcEndpoint synthesis
## Service

EC2

## API Action / Feature

`DescribeVpcEndpointServices` returns an empty `serviceDetailSet`. CDK's L2 `InterfaceVpcEndpoint` with `lookupSupportedAzs: true` (and the default service objects) calls this at **synth time** to resolve which AZs support the service; an empty response makes `cdk synth`/`deploy` abort before any stack deploys.

Returning the standard interface-endpoint services for the region (com.amazonaws.<region>.<service> for the common set — ec2, ssm, logs, sts, secretsmanager, kms, ecr.api, ecr.dkr, iot.data, etc.) with all AZs listed is enough to unblock the lookup.

## AWS Documentation

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcEndpointServices.html

## Why is this needed?

Supports running the [aws-bench](https://github.com/aws-bench/aws-bench) benchmark scenarios against Floci: four of the eight scenario CDK apps (streaming-and-iot, databases-and-storage, api-and-observability, troubleshooting-multiservice) fail to synth entirely because of this — a much bigger unlock than any single CFN type. Part of the gap series tracked in lex00/floci#17.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #2004. `DescribeVpcEndpointServices` returned an empty set, which aborts CDK `InterfaceVpcEndpoint` synthesis (`lookupSupportedAzs: true` calls it at synth time to filter AZs). Now returns the standard `com.amazonaws.<region>.<service>` interface services plus `s3`, each with the region's AZs, serviceType (Interface/Gateway), and owner metadata; honors a `ServiceName` filter.

Context: supports running the [aws-bench](https://github.com/aws-bench/aws-bench) scenarios against Floci — this single gap blocked four of eight scenario CDK apps from synthesizing at all (streaming-and-iot, databases-and-storage, api-and-observability, troubleshooting-multiservice) — a bigger unlock than any one CFN type (series: lex00/floci#17).

## Tests

`Ec2IntegrationTest.describeVpcEndpointServicesListsInterfaceServicesWithAzs`: the action returns interface services (ecr.api, s3), a serviceType, and the region AZs. Full EC2 suite: 144 tests, 0 failures.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
@@ -1322,6 +1322,21 @@ class Ec2IntegrationTest {
 
     @Test
     @Order(52)
+    void describeVpcEndpointServicesEchoesAnExplicitServiceNameWithAzs() {
+        given()
+            .formParam("Action", "DescribeVpcEndpointServices")
+            .formParam("ServiceName.1", "com.amazonaws.us-east-1.custom-iot-data")
+            .header("Authorization", AUTH_HEADER)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .body(containsString("custom-iot-data"))
+            .body(containsString("us-east-1a"));
+    }
+
+    @Test
+    @Order(52)
     void describeVpcEndpointServicesListsInterfaceServicesWithAzs() {
         given()
             .formParam("Action", "DescribeVpcEndpointServices")
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
@@ -1345,9 +1345,10 @@ class Ec2IntegrationTest {
             .post("/")
         .then()
             .statusCode(200)
-            .body(containsString("com.amazonaws.us-east-1.ecr.api"))
-            .body(containsString("com.amazonaws.us-east-1.s3"))
-            .body(containsString("<serviceType>Interface</serviceType>"))
+            .body(containsString(
+                "<serviceName>com.amazonaws.us-east-1.ecr.api</serviceName><serviceType><item><serviceType>Interface</serviceType>"))
+            .body(containsString(
+                "<serviceName>com.amazonaws.us-east-1.s3</serviceName><serviceType><item><serviceType>Gateway</serviceType>"))
             .body(containsString("us-east-1a"));
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
@@ -1347,8 +1347,12 @@ class Ec2IntegrationTest {
             .statusCode(200)
             .body(containsString(
                 "<serviceName>com.amazonaws.us-east-1.ecr.api</serviceName><serviceType><item><serviceType>Interface</serviceType>"))
+            // S3 carries both offerings, in that order.
             .body(containsString(
-                "<serviceName>com.amazonaws.us-east-1.s3</serviceName><serviceType><item><serviceType>Gateway</serviceType>"))
+                "<serviceName>com.amazonaws.us-east-1.s3</serviceName><serviceType>"
+                + "<item><serviceType>Gateway</serviceType></item>"
+                + "<item><serviceType>Interface</serviceType></item>"
+                + "</serviceType>"))
             .body(containsString("us-east-1a"));
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
@@ -1322,6 +1322,22 @@ class Ec2IntegrationTest {
 
     @Test
     @Order(52)
+    void describeVpcEndpointServicesListsInterfaceServicesWithAzs() {
+        given()
+            .formParam("Action", "DescribeVpcEndpointServices")
+            .header("Authorization", AUTH_HEADER)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .body(containsString("com.amazonaws.us-east-1.ecr.api"))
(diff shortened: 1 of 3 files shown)


src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
```diff
@@ -1239,6 +1239,37 @@ void attachInternetGateway() {
...
+            .formParam("Action", "DescribeVpcEndpointServices")
+            .header("Authorization", AUTH_HEADER)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .body(containsString("com.amazonaws.us-east-1.ecr.api"))
+            .body(containsString("com.amazonaws.us-east-1.s3"))
+            .body(containsString("<serviceType>Interface</serviceType>"))
+            .body(containsString("us-east-1a"));
```
pgermosen, line 1356: This asserts `<serviceType>Interface</serviceType>` appears *somewhere* in the whole response body, which is true because `ecr.api` (or any other interface service) carries it — it doesn't verify that `s3` specifically gets `Gateway` and `ecr.api` specifically gets `Interface`. Since `XmlBuilder` emits compact XML with `serviceName` immediately followed by `serviceType` in the same `item`, you can tie the assertion to the specific pairing and pin the s3-is-Gateway behavior as an explicit regression guard, using matchers already imported in this file:

```suggestion
        given()
            .formParam("Action", "DescribeVpcEndpointServices")
            .header("Authorization", AUTH_HEADER)
        .when()
            .post("/")
        .then()
            .statusCode(200)
            .body(containsString(
                "<serviceName>com.amazonaws.us-east-1.ecr.api</serviceName><serviceType><item><serviceType>Interface</serviceType>"))
            .body(containsString(
                "<serviceName>com.amazonaws.us-east-1.s3</serviceName><serviceType><item><serviceType>Gateway</serviceType>"))
            .body(containsString("us-east-1a"));
```

pgermosen (approved):
Looks good 

(I have few minor comments in the PR)

src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
lex00: Tightened. Both assertions now match `serviceName` immediately followed by `serviceType` in the same item, so `ecr.api` is pinned to `Interface` and `s3` to `Gateway` specifically rather than matching anywhere in the body.

lex00:
Both remaining findings are fixed at `a7bc76c`.

**S3 Interface discovery** — S3 is the one service with both offerings and AWS reports both types on its single service detail, but the handler emitted `Gateway` only, contradicting the comment right above the loop. It now emits `Gateway` and `Interface` for S3 and `Interface` for everything else.

**The compatibility suite** — `Ec2Tests.describeVpcEndpointServices` still asserted both collections come back empty. It now asserts what the endpoint actually returns: the S3 detail carrying both types, `ecr.api` as Interface-only, and non-empty availability zones. The `@DisplayName` was stale too and now describes the real behavior.

Verified against a live instance rather than only in-process: `Ec2Tests` passes 52/52 with `FLOCI_ENDPOINT` pointed at a build of this branch, and the emulator returns `<serviceType><item><serviceType>Gateway</serviceType></item><item><serviceType>Interface</serviceType></item></serviceType>` for S3.

lex00:
Rebased on main, conflict in `Ec2IntegrationTest` resolved. 148 tests pass.

src/test/java/io/github/hectorvent/floci/services/ec2/Ec2IntegrationTest.java
lex00: Applied with one change. S3 reports both types on its single service detail, so the assertion pins the s3 serviceName followed by Gateway and Interface items, and ecr.api as Interface only.

pgermosen (approved):
Re-approving on `47e5939f`.

The existing approval on this PR was posted on 28 July against `ef515661`. GitHub keeps an approval alive across pushes, so the tick has been sitting there covering code nobody reviewed — and the three commits since were not cosmetic. The response shape itself changed:

- `serviceType` became a **list** rather than a scalar
- `s3` now reports **both** `Gateway` and `Interface`
- an `owner` element was added
- a test was added for the explicit `ServiceName` filter echo

So this needed a fresh read rather than a rubber stamp on a stale one.

## The new shape is correct

I checked every emitted element against botocore's `DescribeVpcEndpointServicesResult` and `ServiceDetail` rather than eyeballing it:

| Emitted | Model `locationName` |
|---|---|
| `serviceNameSet` | `ServiceNames` |
| `serviceDetailSet` | `ServiceDetails` |
| `availabilityZoneSet` | `AvailabilityZones` |
| `serviceType` as a set of `<item><serviceType>…` | `ServiceTypeDetailSet` |
| `owner` | `Owner` |

`availabilityZoneSet` is the one worth calling out — the natural guess is `availabilityZones`, and getting it wrong would have produced a response the SDKs parse into an empty list rather than failing loudly, which is exactly the kind of divergence that gets blamed on CDK rather than on the emulator.

The dual `Gateway`/`Interface` for S3 is right too, and it is the reason `ServiceType` is modelled as a set in the first place: S3 is the service that supports both endpoint types. Reporting only `Interface` there would have been plausible and wrong.

## Verified

On current `main` (`4058dfae`): **336 EC2 tests green.**

Forcing S3 back to a single `Interface` type fails `describeVpcEndpointServicesListsInterfaceServicesWithAzs`, so the dual-type behaviour is pinned rather than incidental.

## On the change itself

The framing in the description holds up — this is a larger unlock than the resource-type PRs in the same series, because an empty `DescribeVpcEndpointServices` aborts CDK synth outright rather than degrading at deploy time. Four of eight scenario apps failing to synthesise at all is a different class of blocker from a resource that provisions incompletely, and it is worth prioritising accordingly.

Updating the compatibility test's `@DisplayName` from *"returns empty list"* to *"lists the common services with AZs"* was the right instinct too — leaving a test named for the old behaviour is how a fixed bug gets quietly reintroduced later.

Thank you for the patience while this sat.

PR #2009, opened 2026-07-27
feat(ec2): DescribeImages synthesizes an AMI for an unmatched wildcard name lookup

### LINKED ISSUE #2008: [FEAT] EC2: DescribeImages should resolve a wildcard name filter that matches no seeded AMI (CDK MachineImage.lookup)
## Service

EC2

## API Action / Feature

CDK's `MachineImage.lookup({ name: 'suse-sles-16-0-*', owners: ['amazon'], filters: { architecture: ['x86_64'] } })` is a synth-time context provider: it calls `DescribeImages` with a `name` wildcard filter and aborts `cdk deploy` with 'No AMI found that matched the search criteria' when the response is empty. Floci seeds a handful of amazon-owned AMIs, but none match arbitrary lookup name patterns (SLES, Ubuntu, Windows, specific dated AL2 builds, ...).

Proposal: when a `name` filter contains a wildcard and no seeded image matches, synthesize a single deterministic AMI whose name satisfies the pattern (id derived from the pattern hash, owner/architecture from the filters). This makes any `MachineImage.lookup` resolve, which is what an emulator needs — the exact AMI id is a runtime detail the workload doesn't depend on.

## AWS Documentation

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html

## Why is this needed?

Supports running the [aws-bench](https://github.com/aws-bench/aws-bench) benchmark scenarios against Floci: several scenarios (databases-and-storage, and others using `MachineImage.lookup`) fail `cdk deploy` at synth on this. Part of the gap series tracked in lex00/floci#17.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #2008. When `DescribeImages` gets a `name` wildcard filter that matches no seeded AMI, it now synthesizes a deterministic image (id from the pattern hash) satisfying the name/owner/architecture filters, instead of returning empty. This unblocks CDK's `MachineImage.lookup` synth-time context provider, which otherwise aborts `cdk deploy` with 'No AMI found that matched the search criteria'.

Context: supports running the [aws-bench](https://github.com/aws-bench/aws-bench) scenarios against Floci — several scenarios use `MachineImage.lookup` for SLES/Ubuntu/etc. and fail deploy at synth (series: lex00/floci#17).

## Tests

`Ec2IntegrationTest.describeImagesSynthesizesForUnmatchedWildcardNameLookup`: a `suse-sles-16-0-*` + x86_64 lookup returns a synthesized AMI matching the name stem and architecture. Full EC2 suite green.


lex00:
Fixed at `b83a34a`. The substring crash was already fixed by `4ee2059`; this covers the filter bypass.

The synthesized name was built by truncating the pattern at its first wildcard, so `ubuntu-*-20.04-*` produced `ubuntu-20260101` — exactly your example. Each wildcard is substituted now, and the scalar fields a request can filter on (architecture, owner alias, state, root device name and type, virtualization type, hypervisor) take the requested value instead of a hardcoded one.

Synthesis is also no longer unconditional: the result is checked against the full filter set before being returned and withheld when it cannot satisfy it. An empty result is what the caller would have got anyway, and it beats an AMI that contradicts the request with no way for the caller to tell.

Three integration tests: an infix pattern, a non-name filter being honored, and a request the synthesizer cannot satisfy returning nothing. 247 tests across the EC2 suites pass.

lex00:
Merged current main. The conflict was handleCreateImage from #1979 and the lookup synthesis here landing at the same point in Ec2QueryHandler. Both are kept.

Ec2ServiceTest and Ec2IntegrationTest pass, 213 tests.

lex00:
Both halves fixed in c362cba4.

wildcardMatches expanded only *, quoting everything else, so a ? in a name filter matched literally and ubuntu-? found nothing. wildcardToRegex further down the same file already read both wildcards, so the two disagreed on the same syntax. That half is not specific to synthesis and affects any name filter.

Synthesis substituted * and left ? in place, which handed back a name carrying a literal ? that the filter would no longer match once the matcher reads ? properly. It now substitutes a single character for each.

Ec2IntegrationTest and Ec2ServiceTest pass, 214 tests.

hectorvent:
Thanks for keeping at this, and I am sorry you have had no verdict from me since July. The `?` fix in `c362cba4` is right: `wildcardMatches` and `wildcardToRegex` genuinely disagreed on the same syntax, and treating `?` as exactly one character matches how moto reads EC2 filters. Botocore documents no wildcard syntax on `Filter`, so a reference implementation is the best source available.

**(blocking)** `synthesizeLookupImage` takes `OwnerId` from the request scope, so `Owner.1=self` with no `owner-id` filter puts the literal `"self"` there (`Ec2QueryHandler.java:1605`). The guard below it only re-checks `Filter.N`, so the image never faces the owner scope that produced it, and `Ec2Service.java:2514` resolves `self` against the account id, which it would fail. CDK's `MachineImage.lookup` passes owners, so that is the motivating path. Your `?` change widens it slightly, since synthesis now fires on `?` too. No test passes `Owner.N`, so nothing goes red when you fix it.

lex00:
Good catch, and it was worse than the literal. Two halves, both in 8c5d263c.

resolveSynthOwner now maps the self alias to the account id, since Ec2Service matches self by comparing the image owner against accountId. Anything else, including a bare account id or an owner-id filter, passes through as before.

The guard now calls imageMatchesOwners alongside imageMatchesFilters, so the synthesized image faces the owner scope as well as Filter.N. Ec2Service already had matchesImageOwners privately, so this exposes it next to imageMatchesFilters rather than restating the rule in the handler.

Two tests, one asserting the owner resolves to the account id under Owner.1=self, one asserting a foreign owner scope gets the empty result instead of a contradicting AMI. Neither passed before, since nothing in the suite sent Owner.N.

lex00:
Also two rounds, and the first was incomplete in a way that made things worse.

Resolving only self left amazon and aws-marketplace written through verbatim, and the owner check added in the same commit then accepted the literal by string equality. Both aliases now resolve to the accounts they name, in the synthesized image and in matchesImageOwners.

That still only covered registered images. The catalog matches on its own owner metadata and five of the ten seeded entries carry none, so Owner.1=amazon omitted Amazon Linux 2 and 2023 either way. Both now carry ownerId 137112412989 and the amazon alias, matching how the Canonical entries are already seeded.

Debian, Alpine and Windows Server are still without owner metadata. Same pre-existing gap, and it needs their real account ids rather than a guess, so I left it.

lex00:
Right again, and this one predates the ? change rather than following from it.

A filter's values are an OR, and the code took the first value outright, so an exact value ahead of a wildcard meant the wildcard check failed and synthesis never fired. It now picks the first value that actually carries a wildcard. The ? change widened which values qualify, so it made the case easier to hit.

Test added with an exact value first and a wildcard second, asserting the synthesized name comes from the second. Fixed in 7b6fc863.

lex00:
Fixed in cf2a55ed.

The alias defaulted to amazon whatever the owner resolved to, so Owner.1=self reported account 000000000000 beside imageOwnerAlias amazon. The alias now comes from the resolved account, and an account with no alias emits no element, matching AWS.

An owner-alias filter was consulted by nothing, so filtering on it alone left the id at the Amazon default. Resolution now reads the owner-id filter, then Owner.N, then the owner-alias filter, then the default.

Asserted as a pair across the four owner scopes. 244 tests pass.

hectorvent:
That is it, thank you, and the extras you added on top are all correct.

The synthesized image now resolves its owner through `resolveSynthOwner` and faces `imageMatchesOwners` before being handed back, so `Owner.1=self` maps to the account id instead of carrying the literal. That was the case that would have failed downstream.

Two of the follow-ons are worth calling out. Resolving the aliases to real accounts is right, and the ids match AWS: 137112412989 publishes the Amazon Linux images, 679593333241 is the Marketplace account. Seeding those onto the catalog images so they agree with the synthesis path closes the other half, and dropping the default alias means an image no longer claims amazon ownership while carrying someone else's owner id.

No blockers from my side.

hectorvent (approved):

Issue #1989, opened 2026-07-27
[BUG] ExecutedVersion is always $LATEST, even for a version- or alias-qualified invoke


bdemoss-runwise:
## Service

Lambda

## AWS API Action

`Lambda:Invoke` — the `X-Amz-Executed-Version` response header (surfaced by the CLI
and SDKs as `ExecutedVersion`).

## Expected behavior

The version that actually ran. For a version- or alias-qualified invoke, real AWS
returns that version number:

```
aws lambda invoke --function-name f:1 out.json
{"StatusCode": 200, "ExecutedVersion": "1"}
```

## Actual behavior

Always `$LATEST`, regardless of qualifier — `LambdaController:253` sets the header
to a literal:

```java
builder.header("X-Amz-Executed-Version", "$LATEST");
```

So an alias- or version-qualified invoke misreports which version ran. It also
makes the header useless for diagnosing the qualifier-resolution issues filed as
#1987 and #1988 — the response looks identical whichever version served it.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 lambda publish-version --function-name f
aws --endpoint-url http://localhost:4566 lambda invoke --function-name f:1 /tmp/o.json
#   "ExecutedVersion": "$LATEST"   (expected "1")
```

## Environment

- Floci version / image tag: `floci/floci:1.5.33` (source checkout at the same
  tag, `b0e56a9a`)
- How you're running Floci: Docker (`floci/floci:1.5.33`, port 4566)

## Are you willing to contribute a PR?

- [x] Yes
- [ ] No

Issue #1988, opened 2026-07-27
[BUG] Warm pool is keyed by function name only, so a version-qualified invoke reuses the $LATEST container


bdemoss-runwise:
## Service

Lambda (warm container pool)

## AWS API Action

`Lambda:Invoke` with a version or alias qualifier, when a warm container already
exists for that function.

## Expected behavior

A version-qualified invoke runs that version's code — or at minimum does not run a
*different* version's code.

## Actual behavior

Runs whatever the function's existing warm container has, which in practice is
`$LATEST`. `WarmPool.acquire` keys the pool on `fn.getFunctionName()` only
(`WarmPool:109`), and `publishVersion` snapshots carry the same function name, so
`$LATEST` and every published version share one pool.

Sequence that shows it:

1. invoke `f` unqualified → container starts on the current code, stays warm
2. `publish-version` → version 1
3. update the function's code
4. invoke `f:1` → returns the **new** code, not version 1's

Reported as "published versions are not immutable", but the mechanism is pool
keying rather than snapshot contents.

This also masks the cold-path failure filed separately (a version snapshot has no
code path and times out when it actually has to start a container) — with a warm
container present you get silently wrong code instead of an error, which is the
worse of the two symptoms.

## Reproduction

```bash
# 1. warm container on the current code
aws --endpoint-url http://localhost:4566 lambda invoke --function-name f /tmp/o.json

# 2. publish a version
aws --endpoint-url http://localhost:4566 lambda publish-version --function-name f

# 3. change the function's code (redeploy / update-function-code), then:
aws --endpoint-url http://localhost:4566 lambda invoke --function-name f:1 /tmp/o.json
#    -> the post-change code, not version 1's
```

## Environment

- Floci version / image tag: `floci/floci:1.5.33` (source checkout at the same
  tag, `b0e56a9a`)
- How you're running Floci: Docker (`floci/floci:1.5.33`, port 4566)

## Are you willing to contribute a PR?

- [x] Yes
- [ ] No

omatheusmesmo:
Already fixed on main: publishVersion snapshots get a version-qualified ARN (LambdaService.java:2018, landed 2026-08-28), and WarmPool.acquire keys the idle pool on that ARN via executionEnvironmentKey, so $LATEST and a published version no longer share a container. Closing.

omatheusmesmo:
@bdemoss-runwise could you confirm this is resolved for you on current main?

PR #2012, opened 2026-07-28
fix(lambda): give published versions their code location

### LINKED ISSUE #1987: [BUG] Invoking a published version times out — the snapshot carries no code path
## Service

Lambda

## AWS API Action

`Lambda:Invoke` with a version qualifier (`<function>:1`), and by extension any
alias pointing at a published version.

## Expected behavior

Invoking a published version runs that version's code.

## Actual behavior

Times out. The invoke returns 200 with a function error:

```
{
    "StatusCode": 200,
    "FunctionError": "Unhandled",
    "ExecutedVersion": "$LATEST"
}
{"errorMessage":"Task timed out after 3 seconds","errorType":"Function.TimedOut"}
```

`publishVersion` (`LambdaService:926`) builds its snapshot by copying runtime, role,
handler, timeout, memory, package type, environment and so on — but not
`codeLocalPath`. `ContainerLauncher` launches from `fn.getCodeLocalPath()`
(`:139`, `:279`), so a version snapshot starts a container with no code, which
never registers with the Runtime API and times out at the function timeout.

This is masked whenever a warm container already exists for the function, because
the warm pool is keyed on function name alone and the version-qualified invoke
reuses the `$LATEST` container — filed separately, since the two are independent.

## Reproduction

```bash
# Any function that works unqualified.
  --function-name f /tmp/out.json          # 200, real payload

# Remove warm containers so the version invoke actually cold-starts.
docker rm -f $(docker ps -aq --filter name=floci-f-)

  --function-name f:1 /tmp/out.json        # Function.TimedOut
```

## Why this matters beyond direct version invokes

SAM's `AutoPublishAlias` means "publish a version and point an alias at it", so an
alias built the way real SAM builds it lands on this path. Pointing the alias at
`$LATEST` instead works, which is the workaround we use locally.

## Environment

  tag, `b0e56a9a`)

- [x] Yes
- [ ] No

bdemoss-runwise:
## Summary

Invoking a published version (`<function>:1`) hangs until the function timeout and
returns no result. `publishVersion` builds the version snapshot by copying fields one
at a time and copies **no code location at all** — not just `codeLocalPath` but
`s3Bucket`, `s3Key`, `imageUri` and `codeSha256`. A numeric qualifier resolves
straight to that snapshot in `resolveInvokeTarget`, so the invoke launches a
container for a function that has no code.

It times out rather than failing because `ContainerLauncher.launch` only validated
the code path when it was non-null:

```java
if (fn.getCodeLocalPath() != null) {
    // ...verify the directory exists
}
```

That made the one state which always means "no code" the one state never checked.
The container comes up empty, the runtime logs

```
Runtime.ImportModuleError: Error: Cannot find module 'index'
```

where nothing surfaces it, and the caller waits out the whole function timeout before
getting HTTP 200 with `X-Amz-Function-Error: Unhandled` and no result body.

Closes #1987

## Approach

Two changes, and the second is the one that matters beyond this bug:

- **`publishVersion` copies what determines what runs** — the code location for every
  package type, plus `layers`, `architectures` and `ephemeralStorageSize`. A published
  version is an immutable snapshot of code plus configuration, so it needs all of it.
- **A missing code location is an error** for non-hot-reload Zip functions, instead of
  silently skipped validation. The root cause here isn't really the dropped field, it's
  that a hand-enumerated field copy drifted and nothing caught it. This makes the next
  dropped field fail at its cause rather than as a timeout somewhere unrelated.

Image functions are excluded from the guard — they carry an `imageUri` and legitimately
have no local path. Hot-reload functions were already excluded; the Docker daemon
validates their bind-mount at start.

Not addressed here, both filed separately: the warm pool is keyed by function name
alone, so a version-qualified invoke can still be served a `$LATEST` container
(#1988), and `X-Amz-Executed-Version` is hardcoded (#1989). Once this lands, #1961's
SAM alias can also stop pointing at `$LATEST` and target the published version as real
SAM does.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

Worth flagging even though it isn't breaking: a Zip function with no deployed code now
raises an error on invoke instead of hanging. The previous behavior was a timeout, so
nothing that worked before stops working.

## AWS Compatibility

Incorrect behavior was: `publish-version` reported success and `list-versions-by-function`
showed the version, but `lambda invoke --qualifier 1` returned no payload after hanging
for the function's full timeout, with `FunctionError: Unhandled`. On real Lambda a
published version is an immutable snapshot of the code at publish time and invokes
exactly like `$LATEST` did at that moment.

Observed against Floci on `main` (`2bea2c0a`) through its own HTTP API in the
integration test below, not against the AWS CLI — this is a behavioral fix to an
existing action, not a new one, so there is no wire protocol to verify.

## Tests

`LambdaVersionIntegrationTest` covered publish, list and delete against a function
created with no code at all, so nothing ever invoked a version. Two new cases:

- `invokePublishedVersion_runsTheVersionsCode` — the reported bug. Verified failing
  before the fix: `JSON path ok doesn't match. Expected: <true> Actual: null`, after
  the container logs `Runtime.ImportModuleError` and the invoke times out at 5s.
- `invokeLatest_baselineRunsTheCode` — a `$LATEST` baseline proving the zip, runtime
  and harness are sound, so a failure in the case above is attributable to versioning.

**The baseline is deliberately a separate function.** Invoking `$LATEST` on the
function under test leaves a warm container that the version-qualified invoke then
reuses — the pool is keyed by function name alone (#1988) — so the version runs
`$LATEST`'s code and the assertion passes while the version path is still broken. My
first draft of this test did exactly that and reported the bug as fixed. The version
invoke has to be a cold start.

One pre-existing fixture updated: `LambdaImageConfigTest.zipFunctionStillUsesHandlerAsCmd`
built a bare Zip function with no code path as a shortcut, which the new guard
correctly rejects. It covers `cmd` construction rather than code validation, so it now
takes a `@TempDir` — a pattern already used elsewhere in the suite.

`./mvnw test`: **7944 tests, 0 failures, 0 errors** (8 skipped), 586 classes, JDK 25,
single process.

## Checklist


src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
```diff
@@ -944,6 +944,24 @@ public LambdaFunction publishVersion(String region, String functionName, String
...
+        snapshot.setS3Bucket(fn.getS3Bucket());
+        snapshot.setS3Key(fn.getS3Key());
+        snapshot.setHotReloadHostPath(fn.getHotReloadHostPath());
+        snapshot.setImageUri(fn.getImageUri());
+        snapshot.setImageConfigCommand(fn.getImageConfigCommand());
+        snapshot.setImageConfigEntryPoint(fn.getImageConfigEntryPoint());
+        snapshot.setImageConfigWorkingDirectory(fn.getImageConfigWorkingDirectory());
+        snapshot.setLayers(fn.getLayers());
+        snapshot.setArchitectures(fn.getArchitectures());
+        snapshot.setEphemeralStorageSize(fn.getEphemeralStorageSize());
```
pgermosen, line 963: **The enumeration is still short by three fields that AWS does include in a version snapshot.**

Your diagnosis is the interesting part of this PR — *"the root cause here isn't really the dropped field, it's that a hand-enumerated field copy drifted and nothing caught it."* Agreed. But the fix keeps hand-enumerating, so I diffed `LambdaFunction`'s 34 fields against what `publishVersion` now copies. Eight are still missing, and three of them are both AWS-immutable-per-version and already surfaced by this codebase:

| Field | Surfaced at | Effect on a published version |
|---|---|---|
| `tracingMode` | `LambdaController:606-608` | `TracingConfig.Mode` is emitted **unconditionally** with a `PassThrough` fallback — so a function with `Active` tracing reports `Active` on `$LATEST` and `PassThrough` on its version |
| `deadLetterTargetArn` | `LambdaController:610-612` | `DeadLetterConfig` silently absent on the version |
| `kmsKeyArn` | `LambdaController:621-623` | `KMSKeyArn` silently absent on the version |

All three are part of the immutable configuration a qualified `GetFunctionConfiguration` returns on real Lambda, so this is the same drift, one field-set over. The `tracingMode` one is the most visible because that key is always emitted — the version doesn't omit it, it actively reports the wrong value.

The other five are, I think, correctly left out, and worth saying so this doesn't read as "copy everything":

- `reservedConcurrentExecutions` — function-level in AWS, not per-version. Correct to skip.
- `urlConfig` — function URLs are created per qualifier via their own API, not inherited from `$LATEST`. Correct to skip.
- `accountId`, `stateReason`, `stateReasonCode` — arguable. `state` *is* copied while its two reason fields aren't, which is a small inconsistency more than a bug.

**The structural version of this comment:** since you've identified enumeration-drift as the root cause, the fix that actually closes the class is a copy — a `LambdaFunction` copy constructor or a `snapshotOf(fn)` that copies every field and then overrides the five that are genuinely version-specific (`version`, `functionArn`, `description`, `lastModified`, `revisionId`). That inverts the default from opt-in to opt-out, so the next field added to `LambdaFunction` is carried automatically and only a deliberate exclusion needs thought. It would have caught these three for free, and it's a smaller diff than the 12 setters currently added.

Entirely reasonable to say that's out of scope and land the three missing fields here instead — the bug in #1987 is fixed either way. I'd just rather the choice be explicit, given the PR argues the general case.

src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ContainerLauncher.java
```diff
@@ -134,14 +134,24 @@ public ContainerHandle launch(LambdaFunction fn) {
...
+                        + ("$LATEST".equals(fn.getVersion()) ? "" : " version " + fn.getVersion())
+                        + " (function has no deployed code)");
+            }
+            Path codePath = Path.of(fn.getCodeLocalPath());
+            if (!Files.exists(codePath)) {
+                throw new RuntimeException("Code directory not found for function '"
+                        + fn.getFunctionName() + "': " + fn.getCodeLocalPath()
+                        + " (function may have been deleted or updated)");
             }
         }
```
pgermosen, line 156: **Guard scoping checks out — verified the three ways this could have produced false positives.**

Turning "no code location" from skipped-validation into an error is the right call, and the exclusions are exactly right. I traced each path that could legitimately have a null `codeLocalPath`:

- **Hot-reload** — excluded by `!fn.isHotReload()`. This is the one path that deliberately nulls it (`LambdaService:1402`, right after setting `hotReloadHostPath`), so without that exclusion every hot-reload function would have broken. Correctly handled.
- **Image** — excluded by the new `!"Image".equals(fn.getPackageType())`. Carries `imageUri` instead.
- **S3-sourced Zip** — *not* excluded, and doesn't need to be: `extractZipCodeFromS3` (`:1353`) routes through `extractZipCodeBytes` (`:1367`), which sets `codeLocalPath` at `:1288`. Same for inline `ZipFile`. So an S3 function always has a real local path by the time it launches.

The error message including the version qualifier only when it isn't `$LATEST` is a nice touch — it's exactly the detail that would have made #1987 diagnosable in seconds instead of presenting as a timeout.

One small thing, take it or leave it: `RuntimeException` rather than `AwsException` means this surfaces as an unhandled invoke error rather than a modelled AWS one. It matches the sibling "Code directory not found" throw immediately below, so it's consistent with the file and I wouldn't change it in this PR — noting only that `AGENTS.md` steers domain errors toward `AwsException`, and this is arguably now a domain error rather than an internal invariant.

pgermosen (approved):
Looks good, just a comment.

This is a very good bug report as much as a fix — the causal chain from "publishVersion drops the code location" through "a null path skipped the one check that mattered" to "presents as a timeout, not an error" is exactly right, and I verified each link.

**One substantive comment (inline):** the enumeration is still three fields short — `tracingMode`, `deadLetterTargetArn` and `kmsKeyArn` are all part of an immutable version snapshot on real Lambda and are all already surfaced by `LambdaController` (`:606-623`), so a published version currently misreports them. `tracingMode` is the visible one, since `TracingConfig.Mode` is emitted unconditionally with a `PassThrough` fallback. Given the PR explicitly argues that hand-enumeration is the root cause, I've suggested the copy-constructor shape that would close the class rather than move the boundary — but landing the three fields here is a perfectly good answer too.

**Two cross-PR notes for whoever sequences these:**

- **This will conflict with #2026** (`fix(lambda): report resolved executed version`). Both add tests to `LambdaVersionIntegrationTest` at overlapping `@Order` values — #2026 inserts at `@Order(4)` and renumbers 5-6, this one adds 6 and 7 — and both touch `LambdaService`, though in different methods, so that half should merge cleanly. Whichever lands second needs a trivial rebase of the order numbers.
- **#2026 is the fix for #1989**, which this description lists as filed separately. Worth knowing they're both in flight.

pgermosen (approved):
Re-reviewed at the latest head (`7afad3d9`), which now includes a merge with current `main` plus a follow-up commit closing out the remaining items from the earlier review round:

- The `tracingMode`/`deadLetterTargetArn`/`kmsKeyArn` gap from the earlier review comment is now closed — `publishVersion` copies all three, with a new regression test (`publishVersion_copiesObservabilityAndEncryptionConfig`) verified against a revert (fails without the fix, passes with it).
- The `@Order` collision this PR picked up from merging with #2026 (both added a test at `@Order(6)`) is fixed — renumbered, no more duplicates.

One thing surfaced along the way, filed separately rather than folded into this fix: `GetFunction`/`GetFunctionConfiguration` both resolve through `LambdaService.getFunction(region, functionName)`, which hardcodes `"$LATEST"` in its storage lookup and never reads a version qualifier — so a qualified `GET .../functions/{name}:1` silently returns `$LATEST`'s config today, not the version's. Confirmed by testing against it directly. Worth its own issue/PR; out of scope here.

Locally verified: `LambdaVersionIntegrationTest` 9/9, `LambdaImageConfigTest` 17/17, full compile clean, `docs-check` clean, merges cleanly against current `main`. The one full-suite failure I hit (`SwfLambdaIntegrationTest` socket timeout) reproduces identically on unmodified `main` in a clean worktree — pre-existing environmental flake, unrelated.

LGTM, approving.

PR #2018, opened 2026-07-28
feat(cognito): implement VerifyUserAttribute

### LINKED ISSUE #2017: [FEAT] Cognito VerifyUserAttribute is unsupported
## Service

Cognito

## API Action / Feature

`VerifyUserAttribute`

Follow-up to #1656.

## AWS Documentation

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_VerifyUserAttribute.html

## Why is this needed?

#1656 adds `GetUserAttributeVerificationCode`, which issues and delivers a code for a signed-in user attribute. The flow still cannot be completed because `VerifyUserAttribute` is not supported and currently returns `UnsupportedOperation`.

This blocks post-sign-up attribute verification through AWS SDK clients and frontend libraries such as Amplify, where `confirmUserAttribute(...)` submits the received code through `VerifyUserAttribute`.

Floci should implement the AWS JSON 1.1 action and:

- resolve and authorize the signed-in user from `AccessToken`
- support verifiable attributes such as `email` and `phone_number`
- validate `Code` against the stored verification code for the same user and attribute
- reject mismatched or expired codes with AWS-compatible errors without modifying the user
- on success, persist `email_verified=true` or `phone_number_verified=true`, depending on `AttributeName`
- return an empty AWS-compatible success response
- add unit and AWS SDK compatibility coverage for email and phone-number verification

The verified-attribute update should be consistent with the behavior introduced for `ConfirmSignUp` in #1881, while keeping the verification-code validation specific to `VerifyUserAttribute`.

- [x] Yes
- [ ] No

### LINKED ISSUE #2022: [FEAT] Validate AWS behavior when re-verifying an already verified Cognito attribute
## Service

Cognito

## API Action / Feature

Validate and implement AWS parity when a signed-in user requests and submits verification for an unchanged attribute that is already verified:

- `VerifyUserAttribute`

Related work:

- #2017
- #2018

## AWS Documentation

- https://docs.aws.amazon.com/cognito/latest/developerguide/signing-up-users-in-your-app.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_GetUserAttributeVerificationCode.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_VerifyUserAttribute.html

## Why is this needed?

AWS documentation says that a confirmed attribute does not have to be verified again unless its value changes. `VerifyUserAttribute` is documented for a signed-in user who has added or changed an auto-verified attribute.

The documentation does not clearly specify the exact runtime behavior when the attribute value is unchanged and its verified flag is already `true`.

Floci currently:

- allows `GetUserAttributeVerificationCode` without checking the verified flag,
- issues a fresh code, subject to the normal rate limit,
- accepts a valid fresh code,
- writes the existing verified flag to `true` again,
- rejects reuse of a consumed code.

Before treating this behavior as AWS-compatible, run a compatibility test against a real Cognito user pool and record the exact result for:

1. requesting a code for an unchanged, already verified `email`,
2. requesting a code for an unchanged, already verified `phone_number`,
3. submitting a fresh valid code if delivery succeeds,
4. submitting the same consumed code again,
5. response status, response body, delivery behavior, and error type/message for each case.

The result should define Floci's expected behavior and produce an automated SDK compatibility/regression test. This investigation is intentionally out of scope for #2017 and #2018.

- [x] Yes
- [ ] No

dzwicker:
## Summary

Closes #2017
Closes #2022
Depends on #1898.

- add `VerifyUserAttribute` JSON 1.1 dispatch
- validate and consume email and phone verification codes
- persist `email_verified` and `phone_number_verified`
- align errors, delivery-destination masking, and consumed-code behavior with real Cognito
- allow a fresh verification round for an unchanged attribute that is already verified
- require an access token carrying the `aws.cognito.signin.user.admin` scope, matching AWS's documented authorization contract for this operation
- scope the consumed-code tombstone to attribute verification only, leaving `ConfirmSignUp`/`ConfirmForgotPassword` reused-code behavior unchanged
- add integration, unit, and AWS SDK compatibility coverage

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Validated against a real Cognito user pool in `eu-central-1`:

- only `email` and `phone_number` can be verified
- an unchanged, already verified email or phone number can receive a fresh code and be verified again successfully
- reusing a successfully consumed code returns `ExpiredCodeException: Invalid code provided, please request a code again.`
- a wrong active code returns `CodeMismatchException: Invalid verification code provided, please try again.`
- invalid attributes return `InvalidParameterException: Invalid attribute name. Only phone_number and email can be verified.`
- invalid access tokens return `NotAuthorizedException: Invalid Access Token`
- observed destinations are masked as `z***@b***` for email and `+********8920` for phone

The self-sign-up sequence was also validated: email-only sign-up verifies email during `ConfirmSignUp`; email plus phone uses SMS for sign-up confirmation, then email can be verified separately through `GetUserAttributeVerificationCode` and `VerifyUserAttribute`.

Per AWS's `VerifyUserAttribute` reference, the access token must carry the `aws.cognito.signin.user.admin` scope. An ID token, or an access token whose scope list explicitly omits it, is now rejected with `NotAuthorizedException`.

`ConfirmSignUp` and `ConfirmForgotPassword` keep their original reused-code behavior (`CodeMismatchException`) unaffected by the consumed-code tombstone added here, since that reuse case was never validated against real Cognito for those two operations.

## Known follow-ups / out of scope

- #2021 - implement pending-value and verified-state semantics when `email` or `phone_number` changes

This PR remains focused on validating a code for the currently stored attribute and setting its verified flag. It does not change user confirmation status, authentication-flow eligibility, or attribute-update/pending-value behavior.

## Checklist

- [x] targeted verification-code and attribute-verification test classes pass locally
- [x] full local suite passed: 7,868 tests, 0 failures, 0 errors, 8 skipped (prior to the authorization-scope and tombstone-scoping commits below)
- [x] AWS SDK compatibility test for `VerifyUserAttribute` passes locally
- [x] real Cognito compatibility comparison in `eu-central-1`
- [ ] downstream Amplify integration test
- [ ] full suite re-run after the authorization-scope and tombstone-scoping commits below

Rebased onto current `main` (2026-09-14). `main`'s Cognito access-token verification was substantially hardened in the meantime (a shared `verifyAccessToken()` now does real signature/issuer/client checks and enforces `token_use`, and access tokens carry `aws.cognito.signin.user.admin` by default); `VerifyUserAttribute` was updated to use that same shared path instead of the ad-hoc token parsing it originally had, so it now behaves consistently with `getUser`, `updateUserAttributes`, and the other self-service operations.

Two commits were pushed directly to this branch by a maintainer to close out review findings: requiring the access-token scope on `VerifyUserAttribute` (adapted, in the rebase above, to use the same shared token-verification path as its sibling operations), and scoping the consumed-code tombstone to attribute verification only.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CognitoVerifyUserAttributeTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CognitoVerifyUserAttributeTest.java
@@ -1,0 +1,204 @@
+package com.floci.test;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.cognitoidentityprovider.CognitoIdentityProviderClient;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.AttributeType;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.AuthFlowType;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.CodeMismatchException;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.ExplicitAuthFlowsType;
+import software.amazon.awssdk.services.cognitoidentityprovider.model.MessageActionType;
+
+import java.net.URI;
+import java.net.URLEncoder;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class CognitoVerifyUserAttributeTest {
+
+    private static final ObjectMapper JSON = new ObjectMapper();
+    private static final HttpClient HTTP = HttpClient.newBuilder()
+            .connectTimeout(Duration.ofSeconds(10))
+            .build();
+    private static final Pattern SIX_DIGIT_CODE = Pattern.compile("\\b(\\d{6})\\b");
+    private static final String PASSWORD = "Verify1234!";
+
+    private static CognitoIdentityProviderClient cognito;
+
+    private String poolId;
+
+    @BeforeAll
+    static void setUpClient() {
+        cognito = TestFixtures.cognitoClient();
+    }
+
+    @AfterAll
+    static void closeClient() {
+        if (cognito != null) {
+            cognito.close();
+        }
+    }
+
+    @AfterEach
+    void deletePool() {
+        if (poolId != null) {
+            cognito.deleteUserPool(b -> b.userPoolId(poolId));
+            poolId = null;
+        }
+    }
+
+    @Test
+    void verifiesEmailAndPhoneNumberWithAwsSdk() throws Exception {
+        poolId = cognito.createUserPool(b -> b.poolName("verify-attribute-" + UUID.randomUUID()))
+                .userPool()
+                .id();
+        String clientId = cognito.createUserPoolClient(b -> b
+                        .userPoolId(poolId)
+                        .clientName("verify-attribute-client")
+                        .explicitAuthFlows(ExplicitAuthFlowsType.ALLOW_USER_PASSWORD_AUTH))
+                .userPoolClient()
+                .clientId();
+
+        String username = "verify-" + UUID.randomUUID();
+        String email = username + "@example.com";
+        String phoneNumber = "+1555" + ThreadLocalRandom.current().nextLong(1_000_000, 10_000_000);
+
+        cognito.adminCreateUser(b -> b
+                .userPoolId(poolId)
+                .username(username)
+                .messageAction(MessageActionType.SUPPRESS)
+                .userAttributes(
+                        AttributeType.builder().name("email").value(email).build(),
+                        AttributeType.builder().name("phone_number").value(phoneNumber).build()));
+        cognito.adminSetUserPassword(b -> b
+                .userPoolId(poolId)
+                .username(username)
+                .password(PASSWORD)
+                .permanent(true));
+
+        String accessToken = cognito.initiateAuth(b -> b
+                        .clientId(clientId)
+                        .authFlow(AuthFlowType.USER_PASSWORD_AUTH)
+                        .authParameters(Map.of("USERNAME", username, "PASSWORD", PASSWORD)))
+                .authenticationResult()
+                .accessToken();
+
+        clearInspectionEndpoint("/_aws/ses");
+        clearInspectionEndpoint("/_aws/sns");
+
+        var emailDelivery = cognito.getUserAttributeVerificationCode(b -> b
+                        .accessToken(accessToken)
+                        .attributeName("email"))
+                .codeDeliveryDetails();
+        assertThat(emailDelivery.attributeName()).isEqualTo("email");
+        assertThat(emailDelivery.deliveryMediumAsString()).isEqualTo("EMAIL");
+        String emailCode = fetchLatestSesVerificationCode(email);
+
+        var phoneDelivery = cognito.getUserAttributeVerificationCode(b -> b
+                        .accessToken(accessToken)
+                        .attributeName("phone_number"))
+                .codeDeliveryDetails();
+        assertThat(phoneDelivery.attributeName()).isEqualTo("phone_number");
+        assertThat(phoneDelivery.deliveryMediumAsString()).isEqualTo("SMS");
+        String phoneCode = fetchLatestSnsVerificationCode(phoneNumber);
+
+        assertThatThrownBy(() -> cognito.verifyUserAttribute(b -> b
+                .accessToken(accessToken)
+                .attributeName("email")
+                .code(differentCode(emailCode))))
+                .isInstanceOf(CodeMismatchException.class);
+        assertThat(userAttribute(username, "email_verified")).isNull();
+
+        cognito.verifyUserAttribute(b -> b
+                .accessToken(accessToken)
+                .attributeName("email")
+                .code(emailCode));
+        cognito.verifyUserAttribute(b -> b
+                .accessToken(accessToken)
+                .attributeName("phone_number")
+                .code(phoneCode));
+
+        assertThat(userAttribute(username, "email_verified")).isEqualTo("true");
+        assertThat(userAttribute(username, "phone_number_verified")).isEqualTo("true");
+
+        assertThatThrownBy(() -> cognito.verifyUserAttribute(b -> b
+                .accessToken(accessToken)
+                .attributeName("email")
+                .code(emailCode)))
+                .isInstanceOf(CodeMismatchException.class);
+    }
+
+    private String userAttribute(String username, String attributeName) {
+        return cognito.adminGetUser(b -> b.userPoolId(poolId).username(username))
+                .userAttributes()
+                .stream()
+                .filter(attribute -> attributeName.equals(attribute.name()))
+                .map(AttributeType::value)
+                .findFirst()
+                .orElse(null);
+    }
+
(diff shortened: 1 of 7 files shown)


pgermosen:
Apologies this sat so long without a human look — it's been open since 28 July and the thread has only
Greptile on it. That's on us, not on the PR. I've been through it properly now.

**The code is in better shape than the PR page suggests.** Three things in the description are stale and
make it read as unfinished when it isn't:

- The **"## Draft status — This draft is visible early"** section, while `isDraft` is `false`.
- **"[ ] rebase onto `main` after #1898 merges"** — #1898 merged on 18 Aug and you *did* rebase: the
  branch's committer dates are today, `merge-base` against `main` is `04d120fd` (current tip), 0 commits
  behind. The box is just unticked.
- **#2020** listed under known follow-ups — it was closed and released via #2027.

Only "downstream Amplify integration test" is genuinely still open. Worth tidying those so the next
reader doesn't discount the PR the way its own description invites.

## One real question, and it's the reason I'm not approving yet

`VerificationCodeService.consume()` used to `store.delete(key)` on success. It now retains a consumed
tombstone, and the consumed-check moved below the expiry-check and changed its `Kind` from `NOT_FOUND` to
`EXPIRED`. That service is shared by three call sites — `CognitoService` 1346, 1537 and your new 1650 —
so the change also lands on `ConfirmSignUp` and `ConfirmForgotPassword`, which the description doesn't
mention.

I measured it, same probe against your branch and against `main` at `04d120fd`:

| case | `main` | this PR |
|---|---|---|
| `ConfirmSignUp`, reused code | `CodeMismatchException` | **`ExpiredCodeException`** |
| `ConfirmForgotPassword`, reused code | `CodeMismatchException` | **`ExpiredCodeException`** |
| `ConfirmForgotPassword`, wrong code after a consumed one | `CodeMismatchException` | **`ExpiredCodeException`** |
| `ConfirmForgotPassword`, wrong code while one is active | `CodeMismatchException` | unchanged ✅ |

To be clear about what I'm *not* claiming: I don't think this is wrong. That last row is the one I most
wanted to check, and the tombstone doesn't shadow genuine mismatches while a code is live — so your
validated claim that "a wrong active code returns `CodeMismatchException`" still holds exactly. The
shadowing only occurs after a successful consume and before a new code is issued, where there's no active
code to mismatch against, which is arguably the right answer. And both codes appear in the documented
error lists for `ConfirmSignUp` and `ConfirmForgotPassword`.

So the question is simply: **when you validated against the real `eu-central-1` pool, did you check the
reused-code behaviour for `ConfirmSignUp` and `ConfirmForgotPassword` as well, or only for the attribute
flow?** Your AWS-compatibility section is precise about what you tested, and it only covers
`VerifyUserAttribute` — which I read as scrupulous rather than as an omission, but it does leave two
operations changing behaviour on evidence gathered from a third.

If you did check them, a line in the description and an integration assertion on one of the two would
close it out. If you didn't, it's worth either checking or narrowing the change so the tombstone applies
only to the attribute purposes.

## What I verified as correct

- `VerifyUserAttribute` is genuinely absent from `main`, so nothing here is redundant with work that
  landed while this waited.
- The implementation matches the API reference: `AccessToken` / `AttributeName` / `Code`, empty `200`,
  email and phone only, `email_verified` / `phone_number_verified` set on success, and every error code
  you map to is in the documented list for the operation.
- **No coverage was lost** despite the −41 lines — I checked because a deletion that size is worth
  checking. `CognitoAttributeVerificationIntegrationTest` goes 8 → 12 `@Test` methods and
  `VerificationCodeServiceTest` 9 → 11; the deletions are reordering.
- Local run on your branch: all Cognito suites plus `VerificationCodeServiceTest`, **194 tests, 0
  failures**.
- Shipping a real SDK-level test (`CognitoVerifyUserAttributeTest`) is more than most PRs here do, and
  it's the kind of coverage that catches the things handler tests can't.

## Two non-blocking notes

- `AliasExistsException` isn't implemented — it's documented for `VerifyUserAttribute` when the verified
  email or phone is already an alias on another profile. That's alias-pool semantics and reads like
  #2021 / #1849 territory rather than yours; flagging it as a known gap, not an ask.
- Consumed tombstones now persist until they expire or a new code replaces them, and nothing sweeps them,
  so in persistent storage mode they reach disk. Negligible in size — just worth knowing it's now a
  retained record rather than a delete.

Happy to re-review quickly once the reuse question is answered.

pgermosen:
Following up — checked in again, and nothing has changed since the last review (still the same two commits, both from 28-29 July), so all of this still applies:

## The description is stale, confirmed

- **"## Draft status — This draft is visible early..."** — the PR is not actually a GitHub draft (`isDraft: false`).
- **"[ ] rebase onto `main` after #1898 merges"** — #1898 merged on 18 August, and the branch *was* rebased onto it at the time, the checkbox was just never ticked. Since then, though, `main` has moved another **7 commits** ahead without a further rebase — worth doing another rebase now regardless, since the description no longer reflects even that earlier state accurately.
- **#2020**, listed under "Known follow-ups / out of scope" — it's closed and was released via #2027, so it's resolved, not an open follow-up.

None of this affects mergeability (I just merged the branch onto current `main` locally — clean, compiles, and the full Cognito suite is still green: 222/222) but it makes the PR read as less finished than it is, which is presumably not what you want reviewers or future readers to take away from it.

## The one open technical question from before is still unanswered

`VerificationCodeService.consume()` now retains a consumed tombstone instead of deleting on success, and that service is shared by three call sites — `ConfirmSignUp`, `ConfirmForgotPassword`, and this PR's new `VerifyUserAttribute` path. I measured that this changes the reused-code error on `ConfirmSignUp`/`ConfirmForgotPassword` from `CodeMismatchException` to `ExpiredCodeException` on `main` vs. this branch. Both codes are documented for those operations, so I don't think this is wrong — but the PR's AWS-compatibility section only describes validation against real Cognito for the `VerifyUserAttribute` flow, not these two.

So, same question as before: when you validated against the real `eu-central-1` pool, did you also check the reused-code behavior for `ConfirmSignUp` and `ConfirmForgotPassword`, or only for the attribute flow? If you did, a line in the description plus one integration assertion would close this out. If not, it's worth either checking or scoping the tombstone to just the attribute-verification path.

Happy to re-review promptly once these are addressed — the implementation itself is solid (verified error codes, masking, and the SDK-level test again this pass) and I don't want the open items above to make it seem otherwise.

KenkoGeek (commented):
The verification-code behavior and error mapping are well researched, and I appreciate the real-Cognito comparison in the PR body. I rechecked the current head against the Cognito API contract and AWS SDK expectations. One authorization blocker remains on the new `VerifyUserAttribute` path.

**HIGH: `VerifyUserAttribute` must require an access token with `aws.cognito.signin.user.admin`, but the current path checks neither `token_use` nor the required scope.** `verifyUserAttribute()` extracts username, pool id and jti, performs revocation/global-signout checks, then consumes the code and persists `<attribute>_verified=true`. A valid Floci-issued ID token can therefore still reach this mutation because the method never verifies `token_use=access`. Also, an access token that does not carry `aws.cognito.signin.user.admin` is not rejected.

AWS is explicit here: `AccessToken` must be a valid signed-in user's access token and must include `aws.cognito.signin.user.admin`. Cognito's access-token documentation also defines `token_use` as `access` and the scope as the permission for user self-service attribute operations.

Please validate both properties before code consumption or persistence, and add negative wire/SDK regressions for at least: ID token rejected, access token without `aws.cognito.signin.user.admin` rejected, and valid access token with the scope accepted. This is especially important because consuming the code is stateful, so authorization must fail before `verificationCodeService.consume(...)`.

The recorded CI and Compatibility runs on this head are green, and I do not see a reason to split this operation from its verification-code support. Once the authorization contract is fixed and the new head is green, the rest of the PR is in good shape.

References:
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_VerifyUserAttribute.html
- https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-access-token.html

pgermosen (approved):
This is in good shape now. The authorization gap KenkoGeek flagged is closed by routing VerifyUserAttribute through the same verifyAccessToken() path getUser and updateUserAttributes already use, plus a scope check on top since nothing else in the file enforces aws.cognito.signin.user.admin yet. I also went ahead and scoped the consumed-code tombstone to just the attribute-verification purposes, so ConfirmSignUp and ConfirmForgotPassword keep exactly the reused-code behavior they had before, which closes out the open question from my first pass without needing a real-Cognito check neither of us could run.

Rebased cleanly onto current main (the branch had drifted quite a bit) and CI's green across the board. The downstream Amplify test is still open on your end, but nothing here is blocking on it.

Issue #1971, opened 2026-07-27
[FEAT] CloudFormation: provision AWS::EC2::LaunchTemplate


lex00:
## Service

CloudFormation / EC2

## API Action / Feature

Provision `AWS::EC2::LaunchTemplate` from CloudFormation templates. The auto scaling provisioner already *reads* a `LaunchTemplate` reference from ASG properties, but the resource type itself is never created, so templates that define a launch template and reference it by `Ref`/`Fn::GetAtt` (`LatestVersionNumber`, `DefaultVersionNumber`) fail to resolve.

## AWS Documentation

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-launchtemplate.html

## Why is this needed?

CDK's `ec2.Instance` and `autoscaling.AutoScalingGroup` synthesize launch templates by default in current aws-cdk-lib, so most CDK-generated EC2 stacks contain this type. It blocks deploying stock CDK EC2 apps against Floci.

## Are you willing to contribute a PR?

- [x] Yes
- [ ] No

lex00:
Context: this is one of a series of gaps found while making Floci able to run the [aws-bench](https://github.com/aws-bench/aws-bench) benchmark scenarios — real-world CDK apps deployed with `cdk deploy` against `AWS_ENDPOINT_URL=http://localhost:4566`. The first scenario (4 stacks, 3 regions, CDK custom resources in Docker-backed Lambdas) deploys fully green with this series applied. Work is tracked in lex00/floci#17.

PR #1973, opened 2026-07-27
feat(cloudformation): provision AWS::EC2::LaunchTemplate

lex00:
## Summary

Fixes #1971.

CDK synthesizes `AWS::EC2::LaunchTemplate` for stock `ec2.Instance`/`autoscaling.AutoScalingGroup` apps, but the CFN provisioner had no case for it — templates referenced by `Ref`/`Fn::GetAtt` failed to resolve even though `Ec2Service` fully models launch templates.

- `provisionLaunchTemplate` maps `LaunchTemplateData` (ImageId, InstanceType, KeyName, SecurityGroupIds, UserData — already base64 in CFN — and IamInstanceProfile Arn/Name) onto the existing `Ec2Service.createLaunchTemplate`.
- `Ref` → launch template id; `Fn::GetAtt` exposes `LatestVersionNumber` / `DefaultVersionNumber` / `LaunchTemplateId`.
- A missing `LaunchTemplateName` falls back to a stack-scoped generated name, matching other provisioners.
- `DeleteStack` deletes the template via `Ec2Service.deleteLaunchTemplate`.

## Tests

New `CloudFormationLaunchTemplateIntegrationTest`: deploys a launch template, asserts `CREATE_COMPLETE`, `Ref`/`GetAtt` outputs resolve, `DescribeLaunchTemplates` finds it by name, and `DeleteStack` removes it. CloudFormation package suite: 183 tests, 0 failures.

PR #2019, opened 2026-07-28
fix(cloudformation): resolve ASG launch templates by id and MixedInstancesPolicy

### LINKED ISSUE #2005: [BUG] CloudFormation: AutoScalingGroup with a mixed-instances launch template fails 'launch template does not exist'
## Summary

An `AWS::AutoScaling::AutoScalingGroup` that references a launch template (via `LaunchTemplate` or `MixedInstancesPolicy.LaunchTemplate` → `LaunchTemplateSpecification`) fails at deploy with `The specified launch template does not exist`, even when the `AWS::EC2::LaunchTemplate` is in the same stack and provisioned first. The ASG provisioner appears to resolve the template by a name/id that doesn't match what CFN's LaunchTemplate provisioning registered.

## Environment

Floci awsbench-integration (main + CFN LaunchTemplate support), Docker compose mode. Repro: aws-bench serverless-apps ECS stack (ECS capacity ASG on a launch template).

## Expected

The ASG resolves an in-stack launch template by its CFN Ref/GetAtt id.

## Why is this needed?

Found running the [aws-bench](https://github.com/aws-bench/aws-bench) serverless-apps scenario against Floci (series: lex00/floci#17): 8 of 9 stacks deploy green; only the ECS-capacity ASG fails on this.

Yes.

thejrose1984:
## Summary

Fixes `AWS::AutoScaling::AutoScalingGroup` failing with `The specified launch template does not exist.` (and the `MixedInstancesPolicy` variant of it) whenever it references a launch template by anything other than `LaunchTemplate.LaunchTemplateName`. Two gaps in `provisionAutoScalingGroup` caused it:

- **`LaunchTemplateId` was passed in the name slot.** The provisioner fell back from `LaunchTemplateName` to `LaunchTemplateId` into a single variable and handed it to `createAutoScalingGroup` as the launch template *name*, with the id argument hardcoded `null`. Auto Scaling matches names exactly, so a valid `lt-` id — including one that a `Ref` to an in-stack `AWS::EC2::LaunchTemplate` resolves to — never matched. Id and name are now passed separately, mirroring what `AutoScalingQueryHandler` already does.

- **`MixedInstancesPolicy` was never read from template properties.** A mixed-instances group reached `createAutoScalingGroup` with no launch source at all, failing with `Valid requests must contain either LaunchTemplate, LaunchConfigurationName, InstanceId or MixedInstancesPolicy parameter.` It is now built in the same shape the Query API parser produces, including `Overrides[].InstanceType` and `InstancesDistribution`. A non-integer distribution value (e.g. `OnDemandBaseCapacity: "abc"`) fails the stack with a `ValidationError` rather than being silently dropped.

The whole change is confined to `provisionAutoScalingGroup` and two small helpers; `AutoScalingService.createAutoScalingGroup` already accepted a separate id and a `MixedInstancesPolicy`, so no service behaviour changed.

## Scope note — rebased onto 1.6.0, provisioning half removed

This PR originally also provisioned the `AWS::EC2::LaunchTemplate` resource itself. While it was in review, **#1973 ("feat(cloudformation): provision AWS::EC2::LaunchTemplate") merged into `main` and shipped in 1.6.0**, adding that exact capability as a dedicated `Ec2LaunchTemplateCfnProvisioner` in the new per-type provisioner registry. Keeping this PR's inline `provisionLaunchTemplate` on top of that would have been dead, shadowed code (the registry is consulted before the switch) and would have added a switch case the codebase is actively decomposing away.

So on the rebase onto `main` I dropped the now-duplicate provisioning and its tests, leaving only the Auto Scaling resolution work, which #1973 did not touch. With #1973 supplying `Ref`/`Fn::GetAtt` on `AWS::EC2::LaunchTemplate` and this PR resolving those values on the ASG side, the three failing shapes from #2005 all reach `CREATE_COMPLETE`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change

## AWS Compatibility

Behaviour follows the AWS resource contract for `AWS::AutoScaling::AutoScalingGroup`: a launch template may be referenced by `LaunchTemplateId`, by `LaunchTemplateName`, or through a `MixedInstancesPolicy` `LaunchTemplateSpecification`, with `Version` and `Overrides[].InstanceType`. No request or response shapes changed.

## Tests

`CloudFormationAsgLaunchTemplateIntegrationTest` covers ASG launch-template resolution by name, by id, through a `Ref` to an in-stack `AWS::EC2::LaunchTemplate`, and through `MixedInstancesPolicy` (with `InstancesDistribution` and a malformed-number failure case). Provisioning of the `AWS::EC2::LaunchTemplate` resource itself is covered by #1973's `CloudFormationLaunchTemplateIntegrationTest`.

## Checklist

- [x] Rebased onto current `main` (1.6.0)
- [x] Commit authored by a human contributor; no AI co-author trailer
- [x] Conventional Commit subject; `CHANGELOG.md` left to semantic-release
- [x] Integration test added

Closes #2005


thejrose1984:
Good catch — confirmed and fixed in 577688d.

Reproduced it first: with the previous commit, a stack that creates an `AWS::EC2::LaunchTemplate` and is then updated with a changed `LaunchTemplateData` ends at

```
<StackStatus>UPDATE_ROLLBACK_COMPLETE</StackStatus>
<StackStatusReason>Launch template name already in use.</StackStatusReason>
```

exactly as described. `CloudFormationService` does pass the prior physical id into `provision(...)`, but `provisionLaunchTemplate` ignored it and always called `createLaunchTemplate`.

It now branches on the recorded physical id and follows the AWS update policy for the resource type:

- **`LaunchTemplateData` changed, same name** → `createLaunchTemplateVersion` on the template the stack already owns. No interruption, physical id preserved, `LatestVersionNumber` increments so an ASG referencing `Fn::GetAtt Lt.LatestVersionNumber` picks up the new version.
- **`LaunchTemplateName` changed** → replacement: a new template is created and the one this stack previously owned is deleted.
- **Template deleted out of band** → the lookup returns nothing and provisioning re-creates it instead of failing.

Two integration tests cover this (`updatingLaunchTemplateDataAddsVersionWithoutReplacing`, `renamingLaunchTemplateReplacesIt`). I verified both fail against the pre-fix behavior — the first with the `UPDATE_ROLLBACK_COMPLETE` above, the second by leaking the renamed-away template — and pass now.

One known limitation worth naming: a new version is derived from the previous one, so a property *removed* from `LaunchTemplateData` in an update is inherited rather than cleared. That is the existing `Ec2Service.createLaunchTemplateVersion` semantic, shared with the `CreateLaunchTemplateVersion` Query API path, so I left it alone rather than changing service behavior in this PR.

Full run of `CloudFormation*, Ec2*, Sam*, CustomResource*, RdsCfn*, StackSet*, AutoScaling*`: 539 tests, 1 failure — `CloudFormationIntegrationTest.createStack_ecrAutoName_isLowercase`, which also fails on unmodified `main` and is unrelated to this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

thejrose1984:
Right again — fixed in eb12b87.

`generatePhysicalName` appends a random 12-char suffix, so an unnamed template got a different name on every execution and `name.equals(existing.getLaunchTemplateName())` could never hold. The versioning branch was unreachable for this case, and each update replaced the template.

The generated name is now minted once and reused from the template the stack already owns, so it is only computed on first create:

```java
LaunchTemplate existing = findExistingLaunchTemplate(region, r.getPhysicalId());
String name = resolveOptional(props, "LaunchTemplateName", engine);
if (name == null || name.isBlank()) {
    name = existing != null
            ? existing.getLaunchTemplateName()
            : generatePhysicalName(stackName, r.getLogicalId(), 128, false);
}
```

An unnamed launch template now follows the same data-change-adds-version policy as a named one. Covered by `updatingUnnamedLaunchTemplateKeepsItsPhysicalId`, which asserts the physical id is unchanged after an update, that the stack output resolving `{"Ref": "Lt"}` still yields that same id, and that the template is still present in `DescribeLaunchTemplates`. Against the pre-fix behavior it fails with `Expected: is "lt-25ee55187d807776a" but: was "lt-c8d1a5a5ee0397d10"` — exactly the churn you described.

---

**One thing I found while writing that test, which I have deliberately left alone.**

My first version of it put an ASG in the stack as the Ref-holding dependent. The update failed, but not on the launch template:

```
<StackStatus>UPDATE_ROLLBACK_COMPLETE</StackStatus>
<StackStatusReason>Auto Scaling group 'cfn-asg-unnamed-...' already exists.</StackStatusReason>
```

`provisionAutoScalingGroup` only ever calls `createAutoScalingGroup` — there is no update path, so `UpdateStack` on any stack containing a named `AWS::AutoScaling::AutoScalingGroup` fails. That is pre-existing on `main` (`git show origin/main:...CloudFormationResourceProvisioner.java` has the single `createAutoScalingGroup` call and no `updateAutoScalingGroup`), and independent of launch template resolution, so I did not want to widen this PR into it. `AutoScalingService.updateAutoScalingGroup` already exists and takes the full argument set, so the fix looks small — happy to do it here if you would rather it land together, or to open a separate issue.

I reworked the test to use a stack `Output` as the dependent instead, which demonstrates the same Ref stability without depending on that unrelated gap.

Full run of `CloudFormation*, Ec2*, Sam*, CustomResource*, RdsCfn*, StackSet*, AutoScaling*`: 540 tests, 1 failure — `CloudFormationIntegrationTest.createStack_ecrAutoName_isLowercase`, which also fails on unmodified `main`.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -964,6 +1031,150 @@ private void provisionEc2Instance(StackResource r, JsonNode props, CloudFormatio
...
+            for (JsonNode networkInterface : data.path("NetworkInterfaces")) {
+                securityGroupIds.addAll(resolveStringList(networkInterface, "Groups", engine));
+            }
+        }
+        List<String> resolvedSecurityGroupIds = new ArrayList<>(securityGroupIds);
+
+        LaunchTemplate launchTemplate;
+        if (existing != null && name.equals(existing.getLaunchTemplateName())) {
+            // Same name: a data change is a new version on the template we already own.
+            launchTemplate = ec2Service.createLaunchTemplateVersion(region,
```
pgermosen: **An update never *clears* a removed `LaunchTemplateData` property.**

`createLaunchTemplateVersion` is a source-version **overlay**: it seeds the new version from the source version's data and then only writes fields that are non-blank (`Ec2Service.java:1793-1814`), with `securityGroupIds` additionally guarded by `!isEmpty()`. That is exactly right for the EC2 API, where `CreateLaunchTemplateVersion` means "start from this version and change these fields."

It's the wrong semantic for CloudFormation, where the template is absolute desired state. Remove `KeyName` (or `IamInstanceProfile`, or every entry from `SecurityGroupIds`) from `LaunchTemplateData` and re-deploy: the new version silently keeps the old value. And because `applyData` (`Ec2Service.java:1963`) writes the version back onto the top-level template, `DescribeLaunchTemplates` reports the stale value too. Real CloudFormation would produce a version reflecting only what the template declares.

Passing `"1"` as the source version wouldn't help — the overlay is the issue, not the base. Fixing it properly needs an `Ec2Service` entry point that treats blank as "clear" (or takes a `LaunchTemplateData` instead of a positional field list), which is more than this PR should be carrying.

Given the blast radius is "stack updates that *remove* a launch-template property," I'd be happy for this to ship as a **documented limitation** — one sentence in `docs/services/cloudformation.md` next to the update-policy paragraph you already added. Just worth being a deliberate choice rather than an accident of reusing the EC2 method.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -964,6 +1031,150 @@ private void provisionEc2Instance(StackResource r, JsonNode props, CloudFormatio
...
+     * API. Matching CloudFormation, {@code Ref} returns the launch template id, and the id, name and
+     * version numbers are exposed as {@code Fn::GetAtt} attributes.
+     *
+     * <p>Updates follow the AWS update policy for this resource type: editing {@code LaunchTemplateData}
+     * adds a new version to the existing template (no interruption, physical id preserved), while
+     * changing {@code LaunchTemplateName} requires replacement, so a new template is created and the
+     * one this stack previously owned is removed. A template whose name CloudFormation generated keeps
+     * that name across updates, so omitting {@code LaunchTemplateName} is not a perpetual rename.
+     */
+    private void provisionLaunchTemplate(StackResource r, JsonNode props, CloudFormationTemplateEngine engine,
```
pgermosen: **Stacks that already contain an `AWS::EC2::LaunchTemplate` will silently change physical id.**

Before this change the type fell into the `default` stub branch (`:361-369`), so existing stacks recorded a synthetic `<LogicalId>-<hash>` physical id — exactly the `PhysicalResourceId=Lt-8fb7a8c6` you observed in `DescribeStackEvents`.

After upgrading, `findExistingLaunchTemplate` returns `null` for that id (it isn't an `lt-` id EC2 knows about), so the next `UpdateStack` takes the **create** branch. The physical id flips from `Lt-8fb7a8c6` to a real `lt-...` without CloudFormation reporting a replacement, and any Output or Export carrying the old `Ref` changes value underneath its consumers.

It degrades gracefully — the stub id is tolerated by `deleteLaunchTemplateSafe`, so nothing errors — and it's unavoidable if the type is going to be provisioned for real. But it is a user-visible migration on the first update after upgrade, and neither the PR description nor the CHANGELOG mentions it. A clause on the CHANGELOG "Added" bullet would cover it.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -964,6 +1031,150 @@ private void provisionEc2Instance(StackResource r, JsonNode props, CloudFormatio
...
+        } else {
+            launchTemplate = ec2Service.createLaunchTemplate(region, name,
+                    imageId, instanceType, keyName, resolvedSecurityGroupIds,
+                    decodeUserData(encodedUserData), encodedUserData, instanceProfileArn,
+                    resolveEc2Tags(props, "TagSpecifications", "launch-template", engine),
+                    instanceTags);
+            if (existing != null) {
+                // Renamed, so the template is replaced: drop the one this stack used to own.
+                deleteLaunchTemplateSafe(existing.getLaunchTemplateId(), region);
+            }
```
pgermosen: **Rename deletes the old template inline, before the update completes.**

Creating the replacement first and deleting the old one after is the right order — no name conflict, nothing dangles. Worth noting the consequence though: the delete happens *during* provisioning, so if a later resource in the same `UpdateStack` fails, the old template is already gone and nothing brings it back.

That's the engine's documented behavior rather than an oversight — `rollbackFailedExecution` (`CloudFormationService.java:607-618`) says as much in its javadoc: *"On an **update**, marks `UPDATE_ROLLBACK_COMPLETE` and keeps the prior physical IDs so a corrected re-deploy proceeds (full prior-template restoration is out of scope)."* Real CloudFormation instead defers replacement cleanup until after `UPDATE_COMPLETE`, precisely so a failed update can roll back onto the old resource.

So this is consistent with how every other resource type here behaves, and I wouldn't hold the PR for it. But it's worth knowing there *is* already a lighter-weight hook for exactly this case if you ever want it: `UPDATE_ROLLBACK_RESTORED_ATTR` (`CloudFormationResourceProvisioner.java:116`) lets a resource restore itself during an update rollback and be re-marked `CREATE_COMPLETE` — currently only the `AWS::IAM::Policy` inline-policy path uses it (`:2300`). A launch-template rename could hold off the delete and register it the same way, without needing a general deferred-cleanup phase.

For this PR a one-line comment noting that updates aren't rolled back here, so cleanup is immediate, would be plenty.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -870,6 +877,66 @@ private void provisionAutoScalingGroup(StackResource r, JsonNode props, CloudFor
...
+    private Integer parseOptionalInt(String value) {
+        if (value == null || value.isBlank()) {
+            return null;
+        }
+        try {
+            return Integer.valueOf(value.trim());
+        } catch (NumberFormatException e) {
+            return null;
+        }
+    }
```
pgermosen: **A malformed number is silently turned into "absent".**

`OnDemandBaseCapacity: "abc"` currently becomes `null`, so the stack reaches `CREATE_COMPLETE` with the field quietly missing instead of failing. AWS rejects a non-integer here with a validation error.

Two house rules point the same way: `AGENTS.md` says *"Never leave a `catch` block empty. If an exception is intentionally tolerated, log it with enough context to diagnose it later"* — this one isn't literally empty, but it discards the cause without a trace — and *"prefer correctness over convenience."* Since `provision()` already converts a thrown exception into `CREATE_FAILED` with the message as `StatusReason`, surfacing it is no more code:

```suggestion
    private Integer parseOptionalInt(String value) {
        if (value == null || value.isBlank()) {
            return null;
        }
        try {
            return Integer.valueOf(value.trim());
        } catch (NumberFormatException e) {
            throw new AwsException("ValidationError",
                    "Value '" + value + "' is not a valid integer.", 400);
        }
    }
```

Minor either way — a malformed literal here is a template bug however it surfaces.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -964,6 +1031,150 @@ private void provisionEc2Instance(StackResource r, JsonNode props, CloudFormatio
...
+    private String decodeUserData(String encodedUserData) {
+        if (encodedUserData == null || encodedUserData.isBlank()) {
+            return null;
+        }
+        try {
+            return new String(Base64.getDecoder().decode(encodedUserData), StandardCharsets.UTF_8);
+        } catch (IllegalArgumentException e) {
+            return encodedUserData;
+        }
+    }
```
pgermosen: **Same swallowed-exception pattern as `parseOptionalInt`, and the fallback is unreliable.**

`AGENTS.md` again: *"Never leave a `catch` block empty. If an exception is intentionally tolerated, log it with enough context to diagnose it later."* This is the second of two new catch blocks that discard the cause with no trace — worth noting your own `deleteLaunchTemplateSafe` (`:3801`) gets this right with a `LOG.debugv`, so the pattern is already established in this file.

Separately, the "stored as-is rather than rejected" fallback doesn't do quite what the javadoc says. `Base64.getDecoder().decode` only throws on an invalid alphabet or length, so plain text that happens to be a multiple of 4 characters and uses only base64 characters decodes to binary garbage instead of falling through — you'd store mojibake rather than the original string. Real-world user data usually contains `#!/bin/bash`, and `!` and `/`-adjacent shell punctuation trip the decoder, so this mostly works by luck.

Low stakes, since CloudFormation's `LaunchTemplateData.UserData` is documented as base64 and `Fn::Base64` is the normal way to produce it. If the pass-through is meant as a convenience for hand-written templates, a `LOG.debugv` on the catch would both satisfy the house rule and make the guess visible when it misfires.

src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationAsgLaunchTemplateIntegrationTest.java
```diff
@@ -0,0 +1,387 @@
...
+                """.formatted(suffix, suffix));
+
+        assertStackCreated(stackName);
+        // Ref must have yielded a real lt- id, not the synthetic physical id of a stubbed resource.
+        assertThat(describeAutoScalingGroup("cfn-asg-instack-" + suffix),
+                containsString("<LaunchTemplateId>lt-"));
+    }
+
+    @Test
+    void asgResolvesMixedInstancesPolicyLaunchTemplate() {
```
pgermosen, line 195: **Most of the new `LaunchTemplateData` mapping is unasserted.**

The four reference shapes are covered well, and the three update-policy tests are the strongest part of this suite — `updatingUnnamedLaunchTemplateKeepsItsPhysicalId` in particular pins a subtlety that would otherwise cause a replacement on every update.

But the property mapping that makes up much of the new provisioner code has nothing asserting it:

- **`InstancesDistribution`** (`OnDemandBaseCapacity` / `OnDemandPercentageAboveBaseCapacity` / `SpotAllocationStrategy`) — this test only checks `LaunchTemplateName` and `Overrides[].InstanceType`, so the entire distribution branch is untested.
- **`TagSpecifications` at both scopes** — `launch-template` on the resource, `instance` inside `LaunchTemplateData`. The `ResourceType` filter is easy to get backwards and nothing here would catch it.
- **`NetworkInterfaces[].Groups`** -> security groups.
- **`IamInstanceProfile.Name`** -> the synthesized `instance-profile/<name>` ARN.
- **The base64 `UserData` decode.**

One test that provisions a launch template with all of these set and reads it back through `DescribeLaunchTemplateVersions` would cover the lot in ~40 lines.

Separately: `VersionDescription` is a documented property of this resource and isn't mapped. Fine to skip, but it means a template using it gets no error and no effect.

CHANGELOG.md
```diff
@@ -13,8 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **iam:** seeded `AWSCloudTrail_FullAccess` and `AWSCloudTrail_ReadOnlyAccess` AWS managed policies.
 - **iam:** distinguish `s3:GetObjectAcl` / `s3:PutObjectAcl` / `s3:Get|Put|DeleteObjectTagging` from base S3 ops via `?acl` / `?tagging` query parameters when resolving an action for IAM enforcement.
 
+- **cloudformation:** provision `AWS::EC2::LaunchTemplate` into EC2 instead of stubbing it, so `Ref` returns the launch template id and `Fn::GetAtt` exposes `LaunchTemplateId`, `LaunchTemplateName`, `DefaultVersionNumber` and `LatestVersionNumber`. Stack updates add a version when `LaunchTemplateData` changes and replace the template when `LaunchTemplateName` changes ([#2005](https://github.com/floci-io/floci/issues/2005))
+
```
pgermosen: **Nit: this bullet splits the "Added" list.**

It's added with a blank line above *and* below, so it renders as its own list block rather than continuing the list with the two `iam:` bullets above it.

```suggestion
- **cloudformation:** provision `AWS::EC2::LaunchTemplate` into EC2 instead of stubbing it, so `Ref` returns the launch template id and `Fn::GetAtt` exposes `LaunchTemplateId`, `LaunchTemplateName`, `DefaultVersionNumber` and `LatestVersionNumber`. Stack updates add a version when `LaunchTemplateData` changes and replace the template when `LaunchTemplateName` changes ([#2005](https://github.com/floci-io/floci/issues/2005))

```

pgermosen (commented):
@thejrose1984  a few minor issues could you check the comments

thejrose1984:
Thanks — all seven addressed in 2326876. Taking them in order.

**1. An update never clears a removed `LaunchTemplateData` property.** Correct, and I agree it belongs in docs rather than a service change: fixing it properly means giving `createLaunchTemplateVersion` an absolute-desired-state mode, which would change behaviour shared with `CreateLaunchTemplateVersion` on the Query API. Documented in three places — the provisioner javadoc (with the reason it is not worked around locally), the CHANGELOG entry, and a "deviations" list in `docs/services/cloudformation.md` telling users to set a property to its intended value rather than deleting it.

**2. Existing stacks silently change physical id.** Added to the CHANGELOG entry and the same docs list: the type previously stubbed to `<LogicalId>-<hash>`, so the first deploy after upgrading finds no launch template behind that id and creates one, swapping in a real `lt-` id without reporting a replacement. Also called out in the javadoc next to the `findExistingLaunchTemplate` lookup that produces the behaviour.

**3. Rename deletes the old template inline.** Agreed it is the right order and consistent with the rest of the provisioner, so I have left the behaviour alone and expanded the comment to say why: the replacement is created first so a failure leaves a usable template, and the delete is deliberately not deferred, so a later resource failing will not bring the old one back — matching every other replaced resource type here.

**4. `parseOptionalInt` silently turns a malformed number into "absent".** Fixed. It now throws `ValidationError` naming the property (`Value of property OnDemandBaseCapacity must be an integer.`), so the stack fails instead of quietly dropping the value, and the `catch` no longer swallows. Covered by `malformedInstancesDistributionNumberFailsTheStack`.

**5. `decodeUserData` — same swallowed exception, and the fallback is unreliable.** You are right, and it is worse than unreliable: it cannot work at all. I first tried to rescue it with a round-trip check, and the test proved the point — `"echohihi"` is *itself* valid canonical base64, decodes to `y?h?(b`, and re-encodes to exactly `"echohihi"`. Nothing distinguishes plain text from an encoded payload.

So I removed the guess rather than logging around it. `UserData` is base64 in both CloudFormation and the EC2 API — templates use `Fn::Base64` — so non-base64 input now fails with the same `InvalidParameterValue` / "UserData is not valid base64 content." that `Ec2QueryHandler` already raises for the property, with a debug log on the way out. `decodeUserData` is package-private so `CloudFormationLaunchTemplateUserDataTest` can cover it directly: padded and unpadded base64, the rejection, the ambiguous-but-valid case (documented as decoded, not second-guessed), and blank input.

**6. Most of the new `LaunchTemplateData` mapping is unasserted.** Added `launchTemplateDataPropertiesAreMapped`, which puts all of it in one stack and asserts through `DescribeLaunchTemplateVersions` / `DescribeLaunchTemplates`: `ImageId`, `InstanceType`, `KeyName`, base64 `UserData`, `IamInstanceProfile` by `Name` resolving to the instance-profile ARN, security groups from **both** `SecurityGroupIds` and `NetworkInterfaces[].Groups`, and `TagSpecifications` at the launch-template and instance scopes. `InstancesDistribution` gets its own test asserting all three fields plus the `Overrides` round trip, since it surfaces on the ASG rather than the template.

**7. CHANGELOG bullet splits the "Added" list.** Fixed — the blank line above it is gone, so it continues the list.

Full run of `CloudFormation*, Ec2*, Sam*, CustomResource*, RdsCfn*, StackSet*, AutoScaling*`: 548 tests, 1 failure — `CloudFormationIntegrationTest.createStack_ecrAutoName_isLowercase`, which also fails on unmodified `main`.

One open item from the earlier thread, still unaddressed on purpose: `provisionAutoScalingGroup` has no update path, so `UpdateStack` on a stack containing a named ASG fails with "already exists". That is pre-existing on `main` and independent of launch templates, so I kept it out of this PR — happy to fold it in here or raise it separately, whichever you prefer.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
```diff
@@ -964,6 +1037,176 @@ private void provisionEc2Instance(StackResource r, JsonNode props, CloudFormatio
...
+    static String decodeUserData(String encodedUserData) {
+        if (encodedUserData == null || encodedUserData.isBlank()) {
+            return null;
+        }
+        try {
+            return new String(Base64.getDecoder().decode(encodedUserData), StandardCharsets.UTF_8);
+        } catch (IllegalArgumentException e) {
+            LOG.debugv("Rejecting launch template UserData that is not base64: {0}", e.getMessage());
+            throw new AwsException("InvalidParameterValue", "UserData is not valid base64 content.", 400);
+        }
```
pgermosen: ### Issue 2 — `decodeUserData` duplicates `Ec2QueryHandler`'s, without its gzip half

Rejecting non-base64 here was right, and the parity you claimed is real: `Ec2QueryHandler.decodeUserData` (`:2387-2395`) throws exactly `AwsException("InvalidParameterValue", "UserData is not valid base64 content.", 400)` — same code, same message.

The problem is what the other half of that method does (`Ec2QueryHandler.java:2397-2404`) and this copy doesn't:

```java
if (decoded.length >= 2 && (decoded[0] & 0xff) == 0x1f && (decoded[1] & 0xff) == 0x8b) {
    try (GZIPInputStream gzip = new GZIPInputStream(new ByteArrayInputStream(decoded))) {
        decoded = gzip.readAllBytes();
    } catch (IOException e) {
        throw new AwsException("InvalidParameterValue", "UserData is not valid gzip content.", 400);
    }
}
```

Gzipped user data isn't a curiosity — cloud-init accepts it, and `Ec2IntegrationTest.createLaunchTemplate` (`:934`) deliberately sets `LaunchTemplateData.UserData` to `gzipBase64("#!/bin/sh\necho launch-template\n")`. So the same template body now resolves two different ways: via `CreateLaunchTemplate` it's stored decompressed; via `AWS::EC2::LaunchTemplate` it's stored as raw gzip bytes read as UTF-8.

**Where that's observable**, since it isn't everywhere and neither side's tests would catch it:

- `DescribeLaunchTemplates`/`DescribeLaunchTemplateVersions` echo `getEncodedUserData()` (`Ec2QueryHandler:2342`), so the wire output is identical on both paths. Your new `launchTemplateDataPropertiesAreMapped` asserts `<userData>` equals the *encoded* string, and the EC2 assertions at `:999`/`:1065` do the same — all correct, none of them sensitive to this.
- It bites on the decoded field: `Ec2QueryHandler:438` does `firstNonBlank(userData, launchTemplateData.getUserData())` when `RunInstances` launches from a template, which flows into `Instance.userData` — what `Ec2MetadataServer` (`:298`) and `Ec2ContainerManager` (`:227`) serve. An instance launched from a CFN-provisioned template gets corrupted user data where the EC2-API-provisioned one gets the script.

**This predates the latest commit** — the old `return encodedUserData` fallback didn't handle gzip either — so it's not a regression, and I wouldn't hold the PR for it. What changed is that closing the base64 half made the two implementations near-identical, which is what makes the remaining divergence worth resolving rather than deepening. `AGENTS.md` is direct about it: *"SQS and SNS may expose multiple compatibility paths; do not let them drift"* and *"Reuse shared utilities."*

**Suggested shape:** lift the handler's version into something both call — a static helper on `Ec2Service`, or a small `Ec2UserData` utility — and delete both copies. That closes the gzip gap, makes the error parity structural instead of hand-maintained, and means the next user-data fix lands once. `CloudFormationLaunchTemplateUserDataTest` would move onto the shared helper nearly unchanged, plus a gzip case.

hectorvent:
@thejrose1984 please read the https://github.com/floci-io/floci/blob/main/CONTRIBUTING.md documents for guidance.

thejrose1984:
Apologies — `CONTRIBUTING.md` line 122 is explicit and I missed it:

> Do not include `Co-Authored-By` trailers for AI tools in commit messages. Attribution should be limited to human contributors.

All six commits carried `Co-Authored-By: Claude Opus 5` and a `Claude-Session:` trailer. I have rewritten the messages to drop both and force-pushed (`--force-with-lease`); the branch head is now f44b57a. Sorry for the churn on the review state — a message rewrite changes every SHA, so the earlier review threads will re-anchor.

The diff itself is untouched. Verified before and after: the HEAD tree hash (`4a9d2f22`) and the hash of `git diff main..HEAD` (`b63c51d1`) are both unchanged, and the six subject lines are the same as before.

I went back through the rest of `CONTRIBUTING.md` rather than assume that was the only thing:

- **Conventional Commits** — all six subjects match the pattern in `conventional-commits.yml`; the PR title does too. (The workflow only validates `%s`, which is why it stayed green while the trailers were there.)
- **Action tables** — `make docs-check` exits 0 and `make docs-test` passes 20/20. My `docs/services/cloudformation.md` edits are all outside the `floci:actions` markers: the supported-resource-types row and prose.
- **Testing policy** — every behaviour change in the PR has a test; nothing here relies on the "no new tests needed" exemption.
- **PR hygiene** — branched off `main`, targets `main`, scoped to the one issue, and references it with `Closes #2005`.
- **Agent files** — no `CLAUDE.md`/`GEMINI.md` symlink committed; that stayed local as the docs intend.

If there is something else in `CONTRIBUTING.md` you were pointing at that I have still not picked up, please say which section and I will fix it.

thejrose1984:
Re-reviewed the diff end to end while waiting on CI and found one more real gap of my own, now fixed in 049abcd.

**Launch-template-scoped tags were create-only.** `TagSpecifications` with `ResourceType: launch-template` was passed to `createLaunchTemplate`, but the update path goes through `createLaunchTemplateVersion`, which takes no template-scoped tags. So editing a tag value and redeploying left the old value on the template — the tag was desired state on create and ignored on every update after.

They are now reconciled separately once the version exists. Covered by `updatingLaunchTemplateTagsAppliesToTheExistingTemplate`, which I confirmed fails against the previous commit (the template still reported `<value>platform</value>` after the stack was updated to `infra`) and passes now.

Removal is still not handled — `Ec2Service.createTags` upserts by key, so a tag deleted from `TagSpecifications` stays on the template. That is the same layering caveat the `LaunchTemplateData` properties already carry, so I documented it in the same place rather than inventing a separate reconciliation path here.

Everything else I checked came back clean:

- **Rollback and delete.** A rollback deletes the template because the resource reaches `CREATE_COMPLETE` with a real physical id; `deleteLaunchTemplateSafe` swallows only `InvalidLaunchTemplateId.NotFoundException`, so a `DeleteStack` on a pre-upgrade stack whose recorded id is the old synthetic `<LogicalId>-<hash>` completes instead of failing.
- **Mutually exclusive launch sources.** A template carrying both `LaunchTemplate` and `MixedInstancesPolicy` reaches `validateLaunchSource` with both set and gets the AWS mutual-exclusion error, rather than one silently winning.
- **Empty `MixedInstancesPolicy: {}`** falls through to `hasUsableLaunchTemplate` and fails with the AWS message about needing a `LaunchTemplateId` or `LaunchTemplateName`.
- **Imports.** `Base64` and `StandardCharsets` are still used elsewhere in the provisioner after `decodeUserData` moved out, and the three imports left unused in `Ec2QueryHandler` by the extraction were removed with it.
- **`make docs-check`** exits 0 and `make docs-test` passes; commit subjects and the PR title still match the CI pattern, and no commit carries an AI trailer.

Full run of `CloudFormation*, Ec2*, Sam*, CustomResource*, RdsCfn*, StackSet*, AutoScaling*`: 553 tests, 1 failure — `CloudFormationIntegrationTest.createStack_ecrAutoName_isLowercase`, which also fails on unmodified `main`.

Still open and deliberately not in this PR, restating so it does not get lost: `provisionAutoScalingGroup` has no update path, so `UpdateStack` on a stack with a named ASG fails with "already exists" — pre-existing on `main`. And `handleRequestSpotInstances` still has its own bare base64 decode that could move to `Ec2UserData.decode`. Happy to take either here or separately, whichever you prefer.

pgermosen (commented):
@thejrose1984  looks good now.

cc
@hectorvent

hectorvent:
Thank you for continuing to build on this; applying tag edits on stack updates rounds out the update policy nicely, and the new test covers it.

The substance here has been ready for a while: the version-on-update and rename-as-replacement handling matches the AWS::EC2::LaunchTemplate registry schema (LaunchTemplateName is createOnly, and Ref plus the version attributes line up with the primaryIdentifier and readOnlyProperties, see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-launchtemplate.html).

1. **(blocking)** The branch still conflicts with current main, most likely CHANGELOG.md and the provisioner from the recent merge wave; a rebase should be mostly mechanical. Once it is pushed we will re-approve CI to run on the new head.

That rebase is the only thing between this and merge.

**(blocking)** Commits 8ebf6cf, 0e4cecc, d50cfcd, 1345cf6, ab78799, f44b57a, 049abcd carry an AI co-author trailer; this repo keeps commit attribution with human contributors (see AGENTS.md). Would you amend or squash those commits and force push so the attribution is yours alone?

thejrose1984:
Rebased onto `main` (1.6.0) and fixed the attribution — force-pushed, branch head is now `281da98`. The single commit is authored by me, with no AI co-author trailer.

Two things to flag, because the rebase turned out to be more than mechanical:

**#1973 landed the LaunchTemplate provisioning while this was in review.** `feat(cloudformation): provision AWS::EC2::LaunchTemplate` (#1973, closes #1971) merged into `main` and shipped in 1.6.0 — 25 minutes after the rebase request here. It provisions the same resource type this PR did, as a dedicated `Ec2LaunchTemplateCfnProvisioner` in the new per-type provisioner registry. Since the registry is consulted before the switch, keeping this PR's inline `provisionLaunchTemplate` would have been dead, shadowed code, and it would have added a switch case that the Item-15 decomposition is removing. So I dropped the provisioning half and its tests on the rebase and kept only the Auto Scaling resolution work, which #1973 didn't touch (`provisionAutoScalingGroup` on `main` still has `LaunchTemplateId` going into the name slot).

The result is much smaller: the id/name split and `MixedInstancesPolicy` handling in `provisionAutoScalingGroup`, plus `CloudFormationAsgLaunchTemplateIntegrationTest`. With #1973 supplying `Ref`/`Fn::GetAtt` on the launch-template resource and this PR resolving them on the ASG side, the three failing shapes from #2005 all reach `CREATE_COMPLETE`. I've rewritten the PR description to match.

**`CHANGELOG.md` is no longer edited by hand** — 1.6.0 switched the file to semantic-release generation, and `changelog-guard` rejects manual edits, so the Conventional Commit subject is now the changelog entry.

Two things I deliberately left out of scope, happy to take either separately: the `Ec2UserData` refactor that shared the base64/gzip decode between the EC2 Query API and CFN (it only mattered for the CFN provisioning path that's now #1973's), and the pre-existing gap that `provisionAutoScalingGroup` has no update path so `UpdateStack` on a stack with a named ASG fails with "already exists".

thejrose1984:
@hectorvent friendly nudge — both blocking items from your Aug 5 review are resolved as of `281da98`:

- **Rebase onto `main` (1.6.0):** done. It turned out to be more than mechanical: #1973 shipped a dedicated LaunchTemplate provisioner while this was in review, so I dropped the now-redundant provisioning half and kept only the ASG-side id/name split + `MixedInstancesPolicy` resolution that #1973 didn't touch. Full details and a rewritten description are in my comment above.
- **AI co-author trailers:** gone — the branch is now a single commit authored by me alone.

The only check reporting is Greptile; the build/test workflows still need your approval to run on the new head (`281da98`), as you mentioned. Whenever you get a chance to re-approve CI and take a final look, I believe this is ready. Thanks!

abanna (changes requested):
Thank you for fixing launch-template ID resolution and adding MixedInstancesPolicy support. The focused CloudFormation integration suite passes 6/6. The contribution guide requires new observable behavior to have corresponding tests, and much of the new property mapping is currently unasserted. Please add a focused read-back test covering `InstancesDistribution`, both `TagSpecifications` scopes, `NetworkInterfaces[].Groups`, `IamInstanceProfile.Name`, and base64 `UserData`. The relevant CloudFormation resource contract is documented here: [AWS::AutoScaling::AutoScalingGroup](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-autoscaling-autoscalinggroup.html). No GitHub Actions run exists for this exact head, so the updated branch will also need its normal CI pass. Thank you for the strong reference-shape and update-policy coverage already present.

pgermosen (commented):
@thejrose1984 — thank you for this, and apologies for how long it has sat. Two things worth separating, because the red build is not what it looks like.

## The failing build is not your change

CI had genuinely never run on this PR — workflow runs from a first-time contributor's fork need a maintainer to approve them, and that had not happened. I have approved them now, which is why checks suddenly appeared today.

What came back is an inherited failure rather than anything in your diff:

```
[INFO] 1 error
```

One error, in an EC2 test, with nothing CloudFormation or Auto Scaling in it. That was #2287 — `main` itself did not compile for part of today — and it was fixed by #2288. Your branch is around 177 commits behind, so the merge CI evaluated was built against a base that could not compile.

## What I would like, when you have a moment

**A down-merge of current `main`, then a check that CI comes back green.** That is all — no code changes.

I would not normally ask for a rebase purely on age, but here the CI signal is genuinely stale in a way that cannot be resolved from this end, and this PR has never had a real green run to point at.

## Why I am confident the code itself is fine

So this does not read as a fishing expedition, I verified it locally against current `main` (`12154297`) first:

- merges cleanly, and **full `test-compile` is clean** — no signature drift, which is a real hazard for a branch this age
- **551 tests green** across `services.cloudformation.**` and `services.autoscaling.**`
- your `CloudFormationAsgLaunchTemplateIntegrationTest` is 6/6
- no conflict with #2177, the other open PR touching `CloudFormationResourceProvisioner` — checked both directions

I also reverted each half of the fix to confirm your tests actually pin the behaviour, and both reproduce the exact symptoms from #2005:

| Reverted | Failure |
|---|---|
| id and name passed separately → back into one slot | 2 tests, `ROLLBACK_COMPLETE` — *"The specified launch template does not exist."* |
| `resolveMixedInstancesPolicy(...)` → `null` | 3 tests — *"Valid requests must contain either LaunchTemplate, LaunchConfigurationName, InstanceId or MixedInstancesPolicy parameter."* |

So the expectation is that a down-merge turns this green with no further work.

## On the scope note

Dropping your inline `provisionLaunchTemplate` once #1973 landed `Ec2LaunchTemplateCfnProvisioner` was the right call, and I confirmed it: `Ec2LaunchTemplateCfnProvisionerTest` is in `main` at 9/9 and your diff does not touch it. Keeping shadowed code behind the registry would have been dead weight, and adding a switch case to a switch the codebase is decomposing would have been worse. Spotting that mid-review and re-scoping rather than pushing through is the harder and better choice.

The reasoning in the summary is genuinely useful too — that Auto Scaling matches names exactly, so an `lt-` id in the name slot can never match, is the kind of explanation that makes a fix reviewable rather than just plausible.

Thanks again, and sorry the CI gate kept this invisible for so long — that one is on us, not on you.

pgermosen:
@thejrose1984 — one practical addition to my note above, so you don't lose time on a dead end.

**Re-running the failed workflow will not clear this.** We tried it: the run is now on attempt 3 and fails identically each time, because a re-run replays the *same* merge commit GitHub computed when the run was first created. It never recalculates against a newer `main`, so the stale base — the one predating #2288 — is baked into that run permanently.

Only a new commit on the branch produces a fresh merge ref. So the down-merge is genuinely the only route to a green run here, not just the tidier one.

To be clear about what is and is not wrong, because I would not want you hunting for a bug that is not there: I built the exact merge commit CI uses (`refs/pull/2019/merge`) locally against current `main`, and it is **clean — `test-compile` passes and 821 tests are green** across the cloudformation and ec2 packages. The merged result of your work is already fine. It is only the pinned run that is stuck.

So the expectation is simply: merge current `main` into your branch, push, and the fresh run should go green with no code changes at all.

Sorry for the extra round trip — between the workflow approval gate and a pinned merge base, this PR has been unluckier with CI than the change deserves.

pgermosen (approved):
@thejrose1984 — approving. All 13 checks are green, and I want to set out plainly what happened with CI, because none of it was your change and you deserve to know your work was never the problem.

## The CI story, for the record

This PR hit three unrelated obstacles in a row:

1. **Workflow runs were never approved.** A first-time contributor's fork needs a maintainer to approve Actions runs, and that had not happened since you opened it in July. That is why you saw only a Greptile check for weeks — not a silent failure, simply nothing running. Entirely our oversight.

2. **The first real run failed on a stale merge base.** It came back with `Ec2PublicIpOnLaunchTest.java:[142,16]`, `1 error` — an EC2 test your diff does not touch. `main` itself briefly did not compile (#2287, fixed by #2288), and the run was pinned to a merge commit predating that fix. Re-running could not clear it, because a re-run replays the same pinned merge commit rather than recalculating against a newer `main`. We confirmed that the hard way at attempt 3.

   I pushed a down-merge of current `main` to your branch to break that deadlock (`021e3921`). No conflicts, and your own diff is untouched — still the same three files, 401 insertions. Apologies for touching your branch without asking first; it seemed kinder than a round trip for something mechanical, but do say if you would rather own that sort of thing yourself in future.

3. **Then the Go SDK image failed to build**, on `go mod tidy` resolving `github.com/aws/aws-sdk-go-v2/service/iotjobsdataplane`. Also not yours — you touch no Go or compatibility files. That module publishes independently, and `go.mod` is deliberately not checked in for these images, so every build resolves `@latest` afresh. Its newest version was published at **18:20:48Z**, between a compat run that passed at 18:13 on another PR and your two failures at 19:02 and 19:21. A propagation window, nothing more, and it has since cleared on its own.

Three separate pieces of bad luck, none of them about launch templates.

## The change itself

I verified it locally throughout, so the review never depended on CI:

- clean `test-compile` and **896 tests green** across `services.cloudformation.**`, `services.ec2.**` and `services.autoscaling.**` on the down-merged branch
- your `CloudFormationAsgLaunchTemplateIntegrationTest` at 6/6
- no conflict with #2177, the other open PR touching `CloudFormationResourceProvisioner`

I reverted each half of the fix to check your tests genuinely pin the behaviour, and both reproduce #2005's exact symptoms:

| Reverted | Failure |
|---|---|
| id and name passed separately → back into one slot | 2 tests, `ROLLBACK_COMPLETE` — *"The specified launch template does not exist."* |
| `resolveMixedInstancesPolicy(...)` → `null` | 3 tests — *"Valid requests must contain either LaunchTemplate, LaunchConfigurationName, InstanceId or MixedInstancesPolicy parameter."* |

Tests written from the observed failure rather than around the fix, which is the version that keeps working after someone refactors nearby.

## Two things I particularly appreciated

**The re-scope when #1973 landed underneath you.** Dropping your inline `provisionLaunchTemplate` once `Ec2LaunchTemplateCfnProvisioner` shipped was right, and I confirmed it — that provisioner is in `main` at 9/9 and your diff does not touch it. Keeping shadowed code behind the registry would have been dead weight, and adding a case to a switch the codebase is actively decomposing would have been worse. Noticing that mid-review and re-scoping is more work than pushing on, and the queue delay is exactly what made it necessary.

**Explaining the mechanism, not just the symptom.** "Auto Scaling matches names exactly, so an `lt-` id in the name slot can never match" is what turned this from a plausible patch into something reviewable. The `MixedInstancesPolicy` half being built in the same shape the Query API parser produces, rather than a parallel path, is the same instinct.

Thank you for the patience — this sat far longer than it should have, and almost all of that was on our side.

PR #2023, opened 2026-07-28
fix(ec2): assign a public IP only when the subnet opts in via MapPublicIpOnLaunch

### LINKED ISSUE #1984: [BUG] EC2: every instance reports PublicIpAddress 127.0.0.1, regardless of subnet
## Summary

`DescribeInstances` reports `PublicIpAddress: 127.0.0.1` for every instance — including instances in private subnets, which on AWS have no public IP at all.

Expected: instances receive a public IP only when their subnet maps public IPs on launch (or an EIP/explicit association exists); private-subnet instances omit the field. The current behavior makes network-reachability reasoning impossible: an instance in a public subnet with an open security group is indistinguishable from a private one.

## Environment

Floci main, Docker compose mode; CDK stack with public-subnet, private-subnet, and default-VPC instances.

## Why is this needed?

Found running aws-bench introspection tasks against Floci (context: lex00/floci#17): tasks like "which instances are reachable via SSH from the internet" have a well-defined answer in the scenario (one public-subnet instance with port 22 open), but with every instance claiming 127.0.0.1 the distinction disappears.

Yes.

lex00:
Fixes #1984.

## Problem
Every launched instance was given `PublicIpAddress = 127.0.0.1` unconditionally in `Ec2ContainerManager`, so an instance in a **private** subnet wrongly reported an internet-facing address. Real EC2 assigns a public IP only when the subnet opts in via `MapPublicIpOnLaunch` (or an explicit `AssociatePublicIpAddress`).

## Fix
- `runInstances` flags the instance (`associatePublicIp`) from its subnet's `MapPublicIpOnLaunch`.
- `Ec2ContainerManager` exposes a public address only when that flag is set. Private-subnet instances get **no** `PublicIpAddress` — `DescribeInstances` omits the element (the XML builder skips nulls), matching real EC2.
- The value for public instances is unchanged (the host-reachable `127.0.0.1`); only presence/absence is corrected.

## Test
`Ec2PublicIpOnLaunchTest` covers the public-vs-private flag (mocks the container manager so no Docker is needed).

Happy to adjust naming/approach to match project conventions.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -663,6 +663,9 @@ public class Ec2Service implements ContainerTeardown {
             inst.setPlacement(new Placement(az));
             inst.setSubnetId(finalSubnetId);
             inst.setVpcId(vpcId);
+            // Public IP is assigned only when the subnet opts in via
+            // MapPublicIpOnLaunch (#1984); private-subnet instances get none.
+            inst.setAssociatePublicIp(subnet != null && subnet.isMapPublicIpOnLaunch());
             inst.setPrivateIpAddress(privateIp);
             inst.setPrivateDnsName("ip-" + privateIp.replace('.', '-') + ".ec2.internal");
             inst.setKeyName(keyName);
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -606,6 +606,17 @@ public class Ec2Service implements ContainerTeardown {
                                     List<String> securityGroupIds, String subnetId,
                                     String clientToken, List<Tag> instanceTags,
                                     String userData, String iamInstanceProfileArn) {
+        return runInstances(region, imageId, instanceType, minCount, maxCount, keyName,
+                securityGroupIds, subnetId, clientToken, instanceTags, userData,
+                iamInstanceProfileArn, null);
+    }
+
+    public Reservation runInstances(String region, String imageId, String instanceType,
+                                    int minCount, int maxCount, String keyName,
+                                    List<String> securityGroupIds, String subnetId,
+                                    String clientToken, List<Tag> instanceTags,
+                                    String userData, String iamInstanceProfileArn,
+                                    Boolean associatePublicIp) {
         if (imageId == null || imageId.isBlank()) {
             throw new AwsException("MissingParameter", "The request must contain the parameter ImageId", 400);
         }
@@ -663,9 +674,12 @@ public class Ec2Service implements ContainerTeardown {
             inst.setPlacement(new Placement(az));
             inst.setSubnetId(finalSubnetId);
             inst.setVpcId(vpcId);
-            // Public IP is assigned only when the subnet opts in via
-            // MapPublicIpOnLaunch (#1984); private-subnet instances get none.
-            inst.setAssociatePublicIp(subnet != null && subnet.isMapPublicIpOnLaunch());
+            // AWS precedence (#1984): the launch-time AssociatePublicIpAddress
+            // override wins in both directions; the subnet's MapPublicIpOnLaunch
+            // attribute is only the default when the launch does not specify it.
+            inst.setAssociatePublicIp(associatePublicIp != null
+                    ? associatePublicIp
+                    : subnet != null && subnet.isMapPublicIpOnLaunch());
             inst.setPrivateIpAddress(privateIp);
             inst.setPrivateDnsName("ip-" + privateIp.replace('.', '-') + ".ec2.internal");
             inst.setKeyName(keyName);
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
@@ -1,0 +1,118 @@
+package io.github.hectorvent.floci.services.ec2;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import io.github.hectorvent.floci.config.EmulatorConfig;
+import io.github.hectorvent.floci.core.storage.PersistentStorage;
+import io.github.hectorvent.floci.core.storage.StorageBackend;
+import io.github.hectorvent.floci.services.ec2.model.Address;
+import io.github.hectorvent.floci.services.ec2.model.Image;
+import io.github.hectorvent.floci.services.ec2.model.InternetGateway;
+import io.github.hectorvent.floci.services.ec2.model.KeyPair;
+import io.github.hectorvent.floci.services.ec2.model.LaunchTemplate;
+import io.github.hectorvent.floci.services.ec2.model.NatGateway;
+import io.github.hectorvent.floci.services.ec2.model.NetworkAcl;
+import io.github.hectorvent.floci.services.ec2.model.Instance;
+import io.github.hectorvent.floci.services.ec2.model.Reservation;
+import io.github.hectorvent.floci.services.ec2.model.RouteTable;
+import io.github.hectorvent.floci.services.ec2.model.SecurityGroup;
+import io.github.hectorvent.floci.services.ec2.model.SecurityGroupRule;
+import io.github.hectorvent.floci.services.ec2.model.Snapshot;
+import io.github.hectorvent.floci.services.ec2.model.SpotInstanceRequest;
+import io.github.hectorvent.floci.services.ec2.model.Subnet;
+import io.github.hectorvent.floci.services.ec2.model.Tag;
+import io.github.hectorvent.floci.services.ec2.model.Volume;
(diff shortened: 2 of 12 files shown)


src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
```diff
@@ -663,6 +663,9 @@ public Reservation runInstances(String region, String imageId, String instanceTy
             inst.setPlacement(new Placement(az));
             inst.setSubnetId(finalSubnetId);
             inst.setVpcId(vpcId);
+            // Public IP is assigned only when the subnet opts in via
+            // MapPublicIpOnLaunch (#1984); private-subnet instances get none.
+            inst.setAssociatePublicIp(subnet != null && subnet.isMapPublicIpOnLaunch());
```
pgermosen: **The launch-time `AssociatePublicIpAddress` override is missing, and omitting it regresses AutoScaling.**

AWS documents the subnet attribute as the *default*, not the decision:

> You can control whether your instance receives a public IP address as follows:
> - Modify the public IP addressing attribute of your subnet.
> - **Enable or disable the public IP addressing feature during launch. This overrides the subnet's public IP addressing attribute.**

Both directions — force on in a private subnet, force off in a public one. Nothing in the EC2 path parses it: grepping `AssociatePublicIpAddress` across `src/main/java` hits only `autoscaling/` and `cloudformation/`, never `Ec2Service` or the EC2 query handler. So `RunInstances` with `NetworkInterface.1.AssociatePublicIpAddress=true` is silently ignored, which is also what Terraform's `aws_instance.associate_public_ip_address` sends.

**The part that concerns me more is AutoScaling.** `LaunchConfiguration` already models the flag (`model/LaunchConfiguration.java:22`), `CreateLaunchConfiguration` parses it (`AutoScalingQueryHandler:105`), `DescribeLaunchConfigurations` returns it (`:126`), and CloudFormation resolves it (`CloudFormationResourceProvisioner:819`). But `AutoScalingReconciler:288` calls `ec2Service.runInstances(...)` — which has no parameter for it — so the stored value is never applied at launch.

That was invisible before this PR, because every instance got `127.0.0.1` unconditionally, which happened to be correct for `associatePublicIpAddress=true`. After this change, an ASG with `AssociatePublicIpAddress: true` scaling into a private subnet gets **no** public IP, and the value it reports back from `DescribeLaunchConfigurations` now contradicts the instance it produced. The PR converts a latent gap into an observable wrong answer for that configuration.

Adding the parameter is mechanical — thread an `Boolean associatePublicIp` (null = "not specified") through `runInstances` and resolve it as `associatePublicIp != null ? associatePublicIp : (subnet != null && subnet.isMapPublicIpOnLaunch())`, which is exactly the AWS precedence. That also lets `AutoScalingReconciler` and the CFN instance path at `:945` pass what they already have.

Entirely reasonable to defer to a follow-up instead — but if so, it's worth saying in the PR description that the ASG path is knowingly left inconsistent, rather than leaving `AssociatePublicIpAddress` mentioned parenthetically as though it were covered.

src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+        Subnet privateSubnet = ec2.createSubnet(REGION, vpc.getVpcId(), "10.0.1.0/24", REGION + "a");
+        // privateSubnet keeps MapPublicIpOnLaunch=false (the createSubnet default)
+
+        Instance inPublic = runOne(ec2, publicSubnet.getSubnetId());
+        Instance inPrivate = runOne(ec2, privateSubnet.getSubnetId());
+
+        assertTrue(inPublic.isAssociatePublicIp(),
+                "instance in a MapPublicIpOnLaunch subnet must be flagged for a public IP");
+        assertFalse(inPrivate.isAssociatePublicIp(),
```
pgermosen, line 69: **The test asserts the flag, not the behavior in the issue title.** *(nit)*

`assertTrue(inPublic.isAssociatePublicIp())` / `assertFalse(inPrivate.isAssociatePublicIp())` verify that `runInstances` sets the field — but the actual change is one level further on, in `Ec2ContainerManager:211`, where the flag gates `setPublicIpAddress`. That branch has no coverage, so the property #1984 is about ("a private-subnet instance reports no `PublicIpAddress`") isn't pinned.

I understand mocking the container manager keeps Docker out of the test, which is the right call. But the mock could still assert the outcome — capture the `Instance` handed to the manager, run the same `if (instance.isAssociatePublicIp())` decision, and assert `getPublicIpAddress()` is null for the private case and `127.0.0.1` for the public one. As it stands, someone could delete the `if` in `Ec2ContainerManager` and this test would still pass.

pgermosen (approved):
The core model is right and matches AWS. 

Verified against the [EC2 instance IP addressing docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html):

> When you launch an instance into a nondefault VPC, the subnet has an attribute that determines whether instances launched into that subnet receive a public IP address... By default, we don't assign a public IP address to instances launched in a nondefault subnet.

**The gap: the launch-time override isn't implemented, and it makes AutoScaling worse than before.** Your description mentions `AssociatePublicIpAddress` parenthetically but nothing parses it, and AWS documents it as overriding the subnet in *both* directions. Details in the inline comment — the ASG interaction is the part I'd want addressed or explicitly deferred, because this PR turns a previously-invisible gap into an observable wrong answer.

I will **Approve** on the subnet behavior; but please see comment 1 before merging.

src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
lex00: Implemented. The subnet attribute is now the default and the launch-time override wins in both directions. `Ec2QueryHandler` parses `AssociatePublicIpAddress` off the primary network-interface spec, `Ec2Service` threads it through `runInstances`, the CFN instance path forwards it, and `AutoScalingReconciler` forwards explicit-true from a launch configuration so the ASG path is no longer worse than before.

src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
lex00: Pinned. The test now captures the `Instance` handed to the container manager and asserts the public-address outcome rather than just the flag, covering the `Ec2ContainerManager` branch in both directions. Still no Docker in the test.

lex00:
Both remaining findings are fixed at `8a13168`.

**Top-level `AssociatePublicIpAddress`** — RunInstances only read the override off `NetworkInterface.1.AssociatePublicIpAddress`, so the legacy top-level parameter was dropped. It is now read as a fallback. AWS rejects both shapes on one request, so the interface spec wins when both are present.

**Auto Scaling explicit false** — the real cause was the model: `LaunchConfiguration.associatePublicIpAddress` was a primitive `boolean`, so an absent flag and an explicit false were the same value and the reconciler could only forward an explicit true. It is a nullable `Boolean` end to end now (model, `AutoScalingService`, `AutoScalingQueryHandler`, the CFN provisioner), so an LC setting false beats a `MapPublicIpOnLaunch` subnet and only null falls back to the subnet default. `DescribeLaunchConfigurations` now omits the element when it was never set, matching AWS.

Coverage: `Ec2RunInstancesPublicIpParamTest` pins all four parse cases (top-level, interface spec, both present, absent), and `AutoScalingReconcilerTest` pins true/false/null reaching `runInstances` unchanged. Verified the new handler test fails without the parse fix. 492 tests across the `AutoScaling*`, `Ec2*`, and `CloudFormation*` suites pass.

src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
lex00: Done. `applyContainerManagerPublicIpGate` mirrors the gate in Ec2ContainerManager, so the private case asserts a null PublicIpAddress and PublicDnsName and the public case asserts 127.0.0.1. Deleting the `if` in Ec2ContainerManager fails the test now.

pgermosen (approved):
Re-checked after the two follow-up commits, since my original approval was conditional on the launch-time override gap being closed.

Both concerns are genuinely fixed. `runInstances` now correctly implements override-wins-both-directions: `associatePublicIp != null ? associatePublicIp : subnet.isMapPublicIpOnLaunch()`. And Greptile's follow-up (a launch configuration's `AssociatePublicIpAddress=false` losing to a public subnet default) is closed via a cleaner fix than the suggested one-liner -- `LaunchConfiguration.associatePublicIpAddress` is now a boxed `Boolean` populated by a proper nullable-bool parser, so `false` forwards as `Boolean.FALSE` rather than collapsing to `null`. The legacy top-level `AssociatePublicIpAddress` parameter is also handled, with the network-interface-spec form correctly taking precedence when both are present.

Ran the full EC2 + AutoScaling suites fresh: 327/327 pass.

Approving.

PR #2026, opened 2026-07-28
fix(lambda): report resolved executed version

### LINKED ISSUE #1989: [BUG] ExecutedVersion is always $LATEST, even for a version- or alias-qualified invoke

aniketshukla1:
## Summary
Closes #1989.

- Carry the resolved Lambda target version through the invocation result.
- Emit that version in X-Amz-Executed-Version instead of always reporting $LATEST.
- Add a version-qualified DryRun regression test for the public HTTP endpoint.

## Type of change
- [x] Bug fix (fix:)
- [ ] New feature (feat:)
- [ ] Breaking change (feat! or fix!)
- [ ] Docs / chore

## AWS Compatibility
Version-qualified and alias-qualified invokes previously reported $LATEST even when another published version ran. The header now reports the resolved version, including aliases that route to a version.

## Checklist

Focused verification: ./mvnw -q -Dtest=LambdaVersionIntegrationTest test and ./mvnw -q -Dtest=LambdaIntegrationTest test (JDK 25).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -612,7 +612,9 @@ public class LambdaService {
         String name = ref.name();
         String qualifier = ref.qualifier();
         LambdaFunction fn = resolveInvokeTarget(region, name, qualifier);
-        return executorService.invoke(fn, payload, type);
+        InvokeResult result = executorService.invoke(fn, payload, type);
+        result.setExecutedVersion(fn.getVersion());
+        return result;
     }
 
     private LambdaFunction resolveInvokeTarget(String region, String name, String qualifier) {
--- a/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaVersionIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaVersionIntegrationTest.java
@@ -77,6 +77,18 @@ class LambdaVersionIntegrationTest {
 
     @Test
     @Order(4)
+    void invokePublishedVersionReturnsExecutedVersion() {
+        given()
+            .header("X-Amz-Invocation-Type", "DryRun")
+        .when()
+            .post(BASE_PATH + "/functions/" + FUNCTION_NAME + ":1/invocations")
+        .then()
+            .statusCode(204)
+            .header("X-Amz-Executed-Version", equalTo("1"));
+    }
+
+    @Test
+    @Order(5)
     void listVersionsByFunction() {
         given()
         .when()
@@ -90,7 +102,7 @@ class LambdaVersionIntegrationTest {
     }
 
     @Test
-    @Order(5)
+    @Order(6)
     void deleteFunctionDeletesAllVersions() {
         // Delete function
         given()
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -250,7 +250,7 @@ public class LambdaController {
         if (result.getLogResult() != null) {
             builder.header("X-Amz-Log-Result", result.getLogResult());
         }
-        builder.header("X-Amz-Executed-Version", "$LATEST");
+        builder.header("X-Amz-Executed-Version", result.getExecutedVersion());
         builder.header("X-Amz-Request-Id", result.getRequestId());
 
         if (result.getPayload() != null && result.getPayload().length > 0) {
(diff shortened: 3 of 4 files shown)


src/test/java/io/github/hectorvent/floci/services/lambda/LambdaVersionIntegrationTest.java
```diff
@@ -77,6 +77,18 @@ void publishSecondVersion() {
 
     @Test
     @Order(4)
+    void invokePublishedVersionReturnsExecutedVersion() {
+        given()
+            .header("X-Amz-Invocation-Type", "DryRun")
+        .when()
+            .post(BASE_PATH + "/functions/" + FUNCTION_NAME + ":1/invocations")
+        .then()
+            .statusCode(204)
```
pgermosen, line 86: **No `$LATEST` baseline — which is the case that could actually regress.**

The header used to be hardcoded `"$LATEST"`, so an unqualified invoke was accidentally correct before and needs to *stay* correct now. Nothing asserts it, and it's the cheapest possible regression guard — it also covers the `qualifier == null` branch of `resolveInvokeTarget`, which the new test doesn't touch:

```java
    @Test
    @Order(4)
    void invokeUnqualifiedReturnsLatest() {
        given()
            .header("X-Amz-Invocation-Type", "DryRun")
        .when()
            .post(BASE_PATH + "/functions/" + FUNCTION_NAME + "/invocations")
        .then()
            .statusCode(204)
            .header("X-Amz-Executed-Version", equalTo("$LATEST"));
    }
```

src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
```diff
@@ -612,7 +612,9 @@ public InvokeResult invoke(String region, String functionName, byte[] payload, I
         String name = ref.name();
         String qualifier = ref.qualifier();
         LambdaFunction fn = resolveInvokeTarget(region, name, qualifier);
-        return executorService.invoke(fn, payload, type);
+        InvokeResult result = executorService.invoke(fn, payload, type);
+        result.setExecutedVersion(fn.getVersion());
```
pgermosen, line 616: **Verified this is safe, noting it so the reasoning is on record.**

Setting the field here rather than in the executor is the right call — `LambdaService` is the layer that resolved the qualifier, and `LambdaExecutorService` never sees it. Two things I checked because they'd have been silent failures:

- `fn.getVersion()` can't be null: `LambdaFunction.version` defaults to `"$LATEST"` (`LambdaFunction.java:42`) and is only reassigned on publish (`:934`). That matters more than it looks — JAX-RS `header(name, null)` *removes* the header rather than sending it empty, so a null here would have silently dropped `X-Amz-Executed-Version` from every response, a regression from the old always-`$LATEST` behavior.
- `executorService.invoke` returns a non-null `InvokeResult` on every path (DryRun 204, Event 202, and each `executeSync` branch), so this can't NPE.

Minor style point, take it or leave it: this is a set-after-construct on an object that otherwise arrives fully built by its constructor. Threading it through as a constructor arg would keep `InvokeResult` immutable-ish, but that touches five call sites for little gain — I'd leave it.

pgermosen (approved):
Looks good 
(added a new comments both are minors)

Clean, correctly-layered fix. `X-Amz-Executed-Version` was hardcoded to `"$LATEST"` at `LambdaController.java:253` regardless of what actually ran; the resolved target now flows through `InvokeResult` instead. Three lines of real logic, in the one layer that knows the answer.


One process note: the checklist leaves `./mvnw test passes locally` unticked, with focused runs of `LambdaVersionIntegrationTest` and `LambdaIntegrationTest` instead. I ran the full suite against this head, so that box is covered — see above.

aniketshukla1:
@pgermosen Thanks for the review, much appreciated.

aniketshukla1:
Thanks

hectorvent (approved):
Thanks @aniketshukla1

PR #2028, opened 2026-07-28
fix(lambda): isolate warm containers by version

### LINKED ISSUE #1988: [BUG] Warm pool is keyed by function name only, so a version-qualified invoke reuses the $LATEST container

aniketshukla1:
## Summary

Closes #1988.

- Record the resolved Lambda version on each launched container.
- Key warm-container buckets by function name and version.
- Delete drains every version bucket, while code, configuration, and reactive S3 updates drain only the affected $LATEST bucket.

## Type of change

- [x] Bug fix (fix:)
- [ ] New feature (feat:)
- [ ] Docs / chore

## AWS Compatibility

Floci previously let a container warmed for f:$LATEST serve a qualified f:1 invocation. Qualified invocations now reuse only containers launched for the same resolved version. Published versions are immutable, so a $LATEST update preserves their warm containers while invalidating stale $LATEST containers. Version code snapshots remain independently covered by #2012.

## Checklist

- [ ] ./mvnw test passes locally (not run; focused suites below)
- [x] Focused regression coverage added

Focused verification (JDK 25):

- `./mvnw -q -Dtest=WarmPoolTest,LambdaExecutorServiceTest,ContainerLauncherTest test`
- `./mvnw -q -Dtest=WarmPoolTest,LambdaServiceTest test`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -440,8 +440,8 @@ public class LambdaService {
         fn.setLastModified(System.currentTimeMillis());
         fn.setRevisionId(UUID.randomUUID().toString());
 
-        // Drain warm containers — they have stale code mounted
-        warmPool.drainFunction(functionName);
+        // Published versions are immutable; only $LATEST containers have stale code mounted.
+        warmPool.drainFunctionVersion(functionName, fn.getVersion());
 
         functionStore.save(region, fn);
         LOG.infov("Updated code for function: {0}", functionName);
@@ -561,8 +561,8 @@ public class LambdaService {
         fn.setLastModified(System.currentTimeMillis());
         fn.setRevisionId(UUID.randomUUID().toString());
 
-        // Drain warm containers so the next invocation picks up the new configuration
-        warmPool.drainFunction(functionName);
+        // Published versions are immutable; only $LATEST containers need the new configuration.
+        warmPool.drainFunctionVersion(fn.getFunctionName(), fn.getVersion());
 
         functionStore.save(region, fn);
         LOG.infov("Updated configuration for function: {0}", functionName);
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/WarmPool.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/WarmPool.java
@@ -23,7 +23,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
 /**
- * Manages a pool of warm Lambda containers per function.
+ * Manages a pool of warm Lambda containers per function version.
  *
  * Two modes controlled by {@code emulator.services.lambda.ephemeral}:
  *  - {@code false} (default): containers are reused across invocations and evicted
@@ -106,7 +106,7 @@ public class WarmPool implements ContainerTeardown {
         ContainerHandle handle = null;
 
         if (!ephemeral) {
-            ArrayDeque<ContainerHandle> queue = pool.computeIfAbsent(fn.getFunctionName(), k -> new ArrayDeque<>());
+            ArrayDeque<ContainerHandle> queue = pool.computeIfAbsent(poolKey(fn), k -> new ArrayDeque<>());
             // Skip pooled handles whose container died out-of-band — otherwise the
             // caller would wait the full Lambda function timeout.
             while (true) {
@@ -154,7 +154,7 @@ public class WarmPool implements ContainerTeardown {
 
         handle.setState(ContainerState.WARM);
         handle.touchLastUsed();
-        ArrayDeque<ContainerHandle> queue = pool.computeIfAbsent(handle.getFunctionName(), k -> new ArrayDeque<>());
+        ArrayDeque<ContainerHandle> queue = pool.computeIfAbsent(poolKey(handle), k -> new ArrayDeque<>());
         boolean returned;
         synchronized (queue) {
             returned = queue.size() < maxPoolSizePerFunction;
@@ -192,19 +192,18 @@ public class WarmPool implements ContainerTeardown {
     }
 
     /**
-     * Stops and removes all warm containers for the given function.
+     * Stops and removes all warm containers for every version of the given function.
      * Called on function delete or code update.
      */
     public void drainFunction(String functionName) {
-        ArrayDeque<ContainerHandle> queue = pool.remove(functionName);
-        if (queue == null) {
-            return;
+        List<ContainerHandle> toStop = new ArrayList<>();
+        String prefix = functionName + "::";
+        for (String poolKey : new ArrayList<>(pool.keySet())) {
+            if (poolKey.startsWith(prefix)) {
+                toStop.addAll(removePool(poolKey));
+            }
         }
-        List<ContainerHandle> toStop;
-        synchronized (queue) {
-            toStop = new ArrayList<>(queue);
-            queue.clear();
-        }
+        if (toStop.isEmpty()) return;
         LOG.infov("Draining {0} container(s) for function: {1}", toStop.size(), functionName);
         stopInParallel(toStop);
     }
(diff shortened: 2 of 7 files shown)


aniketshukla1:
Addressed the Greptile P1 in 3e78d1b.

- Added `WarmPool.drainFunctionVersion`, which removes only one `name::version` bucket.
- `UpdateFunctionCode`, `UpdateFunctionConfiguration`, and reactive S3 sync now invalidate `$LATEST` only. `DeleteFunction` retains the all-version drain.
- Added regressions covering reactive-update preservation of a published-version container and scoped service update calls.

Verified with JDK 25: `./mvnw -q -Dtest=WarmPoolTest,LambdaServiceTest test` (51 tests, 0 failures/errors).

hampsterx:
The pool re-keying itself looks right to me, and the alias handling is correct for a non-obvious reason worth stating: `resolveInvokeTarget` resolves an alias to its concrete version snapshot before the key is computed, so two aliases on version 1 share a pool and an alias on `$LATEST` shares the unqualified pool. Also a nice drive-by: switching `updateFunctionConfiguration` to `drainFunctionVersion(fn.getFunctionName(), ...)` fixes a latent bug where the raw parameter was passed, so a call with a full ARN drained nothing.

Two things I'd want changed before this merges.

**1. Two of the three new tests pass against unfixed code.**

`acquireDoesNotReuseWarmContainerAcrossVersions` (`WarmPoolTest.java:161`) and `drainFunctionDrainsWarmContainersForEveryVersion` (`:186`) never stub `containerLauncher.isAlive(...)`. It's a strict `@Mock`, so `isAlive` returns `false`.

Traced against `main` without the fix:

- `acquire(latest)` cold-starts, `release` pools the handle under key `"versioned-fn"`.
- `acquire(versionOne)` polls that same handle off the bucket (this *is* the bug) then `isAlive(latestHandle)` returns `false`, so it takes the dead-handle branch, stops it, loops, finds the queue empty, and cold-starts.
- `assertSame(versionOneHandle, ...)` passes, `assertNotSame(latestHandle, ...)` passes, `verify(times(2)).launch(any())` passes.

Same for `:186`: pre-fix, `stop(latestHandle)` is called once by the discard path and `stop(versionOneHandle)` once by the drain, so both `verify` calls are satisfied, and there is no `launch` count assertion to catch the extra cold start.

In production `isAlive` returns `true`, which is why #1988 reproduces at all, so these two tests model the inverse of the failing condition. Only `pushCodeUpdatePreservesOtherWarmVersions` genuinely fails pre-fix, because it's the one that stubs `isAlive(versionOneHandle) -> true` at `:220`.

Fix is `lenient().when(containerLauncher.isAlive(any())).thenReturn(true);` in both, plus the `lenient` import. `lenient()` is needed because post-fix `isAlive` is never reached on the cross-version path and a strict stub would fail as `UnnecessaryStubbing`. The file already documents this trap at `:253` and `:284` ("The seed acquire is a cold start (empty pool), so isAlive isn't called").

**2. `Closes #1988` won't hold, even with #2012.**

`CodeStore.getCodePath` is keyed on function name only (`zip/CodeStore.java:33-35`: `baseDir.resolve(sanitizeName(functionName))`), and `updateFunctionCode` re-extracts into that same directory in place. So published-version code is never snapshotted, and #1988's step 4 still misbehaves:

- With this PR alone: the v1 snapshot has no `codeLocalPath`, so a version-qualified cold start launches a container with an empty `/var/task` and hangs to the function timeout.
- With this PR + #2012: the snapshot carries `codeLocalPath`, but it points at the shared, already-overwritten directory, so `invoke f:1` returns the new code again, the original symptom via a different mechanism.

Which also means the AWS-compatibility note here ("Published versions are immutable, so a `$LATEST` update preserves their warm containers") holds for the pool but not for the code, and "version code snapshots remain independently covered by #2012" isn't quite right: #2012 copies the code *pointer*, not the code. Genuine per-version immutability needs a third change (a publish-time copy into a per-version code dir), which nobody has open.

Would you consider downgrading to `Refs #1988` and filing the code-snapshot gap separately, so merging doesn't auto-close an issue whose repro still reproduces?

**Two smaller notes**

- `maxPoolSizePerFunction` (`WarmPool.java:160`) is now applied to a per-version bucket, so a function with 10 published versions under load can hold `10 x nproc` containers where the ceiling used to be `nproc`. Eviction bounds it in time, not in peak. Worth renaming and stating the new ceiling in the class javadoc this PR already edits.
- `docs/services/lambda.md:55` says "Any active 'warm' containers for that function are proactively drained", which is the user-facing contract for the exact path this PR changes (`pushCodeUpdate` now drains only the `$LATEST` bucket).

Heads-up on sequencing: this conflicts with #1963 both textually (both rewrite `drainFunction`, both add state to `ContainerHandle`) and semantically, #1963 bumps a `generations` map keyed by function name, and `drainFunctionVersion` never touches it. If this lands after #1963, #1963's fix is silently defeated on every path it was written for and `release_afterPushCodeUpdate_doesNotReturnCheckedOutContainerToPool` fails outright. Whichever order, the generation bump wants to move into a helper both drain paths call, keyed by pool key rather than function name.

Last thing: the check rollup here only has `Greptile Review`, the real build, native image and SDK suites never ran, so it's worth asking @hectorvent to approve the workflow run. Nothing has compiled this branch yet.

aniketshukla1:
@hampsterx Thanks for the review. Much appreciated.
Workflow ran successfully though.

pgermosen:
Hi @aniketshukla1, thanks for tracking down #1988 — the fix itself is solid. Verified the design end-to-end: `resolveInvokeTarget` correctly resolves a qualified invoke to the version-specific `LambdaFunction` snapshot (with `.getVersion()` set), which flows through to `WarmPool.poolKey()` and onto the launched `ContainerHandle`, so the pool bucketing genuinely closes the bug rather than just moving it. The three new `WarmPoolTest` cases cover exactly the right scenarios (no cross-version reuse, delete drains every version, a code update on `$LATEST` preserves other warm versions). Ran the suites named in the PR body locally against this branch's own base — `WarmPoolTest`, `LambdaExecutorServiceTest`, `ContainerLauncherTest`, `LambdaServiceTest`: 71/71 pass.

One blocking issue: **this branch has a real merge conflict against current `main`**, not just a stale GitHub status — I confirmed it with an actual merge attempt. Conflicts show up in:

- `WarmPool.java` (8 separate conflict blocks)
- `LambdaService.java`
- `ContainerHandle.java`
- `ContainerLauncher.java`

This branch is about three weeks old, and `main` has moved a fair amount in the Lambda area since — including #2012's version-snapshot work, which touches some of the same code this PR changes. Could you rebase onto current `main` and resolve those? Happy to take another look once that's done.

aniketshukla1:
Closing this as superseded after rebasing it against current `main`. The later Lambda architecture now contains a stronger implementation of the same behavior:

- `WarmPool.executionEnvironmentKey` uses the full version-qualified function ARN, so `$LATEST`, published versions, and aliases do not share containers.
- Code/config updates call `drainEnvironment(fn)`, whose per-environment epoch invalidates `$LATEST` (including busy leases) without draining published versions.
- `versionsUseSeparateWarmPools` and `drainingLatestPreservesPublishedVersionPool` already cover both regressions with explicit liveness stubs; I reran both and they pass.
- The pool-size cap on current `main` is still aggregated per function across all environments, so the per-version capacity-multiplier concern from the older implementation no longer applies.

Replaying this branch would replace the current ARN/epoch design with the older `function::version` queue map and regress newer Lambda Extensions/EFS work. No code from this PR should be carried forward. The documentation wording about reactive sync can be tightened independently without reviving the superseded implementation.

PR #2030, opened 2026-07-28
fix(docker): fall back to Windows named pipe when docker-host is at its unix-socket default

### LINKED ISSUE #2006: [BUG] Custom::S3BucketNotifications handler lambda not found — CDK bucket-notification custom resource times out
## Summary

CDK's S3 bucket-notification custom resource (`Custom::S3BucketNotifications`, backed by an auto-generated `BucketNotificationsHandler` Lambda) times out: the stack reports `Function not found: ...-NotificationManager...` then `Timed out waiting for custom resource ... to PUT its response to ResponseURL`. The handler function either isn't provisioned in time or isn't resolvable by the name the custom resource invokes.

## Environment

Floci awsbench-integration, Docker compose mode (TLS + DNS active). Repro: aws-bench reference-architectures S3-eventnotif stack.

## Why is this needed?

Found running the [aws-bench](https://github.com/aws-bench/aws-bench) reference-architectures scenario against Floci (series: lex00/floci#17). `Custom::S3BucketNotifications` is one of the CDK custom-resource types the gap plan flagged for a verification pass; this is the concrete failure.

Yes.

Mide69:
Fixes #2006

Root cause traced via live reproduction: `docker-host` defaults to `unix:///var/run/docker.sock`, which doesn't exist on native Windows (Docker Desktop exposes its API via a named pipe instead). This hardcoded default silently overrides `docker-java`'s own correct OS-aware detection, so every Lambda-backed operation - including Lambda-invoked CloudFormation custom resources like CDK's `Custom::S3BucketNotifications` - failed at container-start time with `SocketException: Network is down`.

**Fix:** when `docker-host` is still at its unmodified default and no `DOCKER_HOST` override is present, fall back to Docker Desktop's named pipe on Windows. An explicit `docker-host` config or `DOCKER_HOST` env var always takes priority. Linux/macOS behavior is completely unchanged.

**Verification:**
- Reproduced the exact failure live: deployed a hand-built CDK-shaped stack (bucket + IAM role + Lambda + `Custom::S3BucketNotifications`) against a native-Windows floci instance - confirmed `CREATE_FAILED` before the fix, `CREATE_COMPLETE` after, reconfirmed on three independent runs
- 7 new/updated tests in `DockerClientProducerTest` (Windows vs. non-Windows, explicit overrides, env var priority) - 19/19 pass
- Full `docker`/`lambda`/`cloudformation` packages (716 tests) - 1 pre-existing unrelated failure (`ZipExtractorTest`, a Windows zip-path quirk, confirmed identical on unmodified `main`)
- Docs updated: `docs/configuration/docker.md`, `CHANGELOG.md`

image

### DIFF
--- a/docs/configuration/docker.md
+++ b/docs/configuration/docker.md
@@ -12,6 +12,8 @@ floci:
     docker-host: unix:///var/run/docker.sock
 ```
 
+**Running natively on Windows** (not inside WSL or a container): Windows has no equivalent of `/var/run/docker.sock`, so when `docker-host` is left at its default and `DOCKER_HOST` isn't set, Floci automatically falls back to Docker Desktop's named pipe (`npipe:////./pipe/docker_engine`) instead. An explicit `docker-host` or `DOCKER_HOST` always takes priority over this fallback.
+
 Environment variable: `FLOCI_DOCKER_DOCKER_HOST`
 
 When running Floci inside Docker Compose, mount the host socket:
--- a/src/main/java/io/github/hectorvent/floci/core/common/docker/DockerClientProducer.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/docker/DockerClientProducer.java
@@ -20,6 +20,13 @@ public class DockerClientProducer {
 
     private static final Logger LOG = Logger.getLogger(DockerClientProducer.class);
 
+    /**
+     * Docker Desktop's well-known named pipe on native Windows. Unlike the unix-socket default,
+     * this isn't reachable by simply bind-mounting a path — Windows has no equivalent of
+     * {@code /var/run/docker.sock}, so a platform-specific fallback is required.
+     */
+    private static final String WINDOWS_DEFAULT_DOCKER_HOST = "npipe:////./pipe/docker_engine";
+
     private final EmulatorConfig config;
 
     @Inject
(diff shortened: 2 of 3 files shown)


hectorvent:
Thank you for this; the write up is one of the best we have had, tracing `docker-host`'s hardcoded default silently overriding docker-java's OS aware detection, reproducing it three times, and volunteering that `ZipExtractorTest` fails identically on unmodified main. The gated CI runs are released and green.

**(blocking)** Please drop the CHANGELOG.md hunk: that file is now generated by semantic release from commit messages at release time, so hand edits conflict with the release commit. Your change is still credited automatically, the release notes are generated from the PR title. Dropping the hunk also dissolves this branch's only merge conflict, which is on CHANGELOG.md alone, so one force push fixes both.

**(follow-up, separate PR)** A shared constant for the `unix:///var/run/docker.sock` literal, and `startsWith("windows")` on `toLowerCase(Locale.ROOT)` in `isWindows()`.

That one CHANGELOG drop is the only blocker.

Mide69:
Thanks for the review! Removed the CHANGELOG.md edits and rebased onto main, which resolved the conflict too. Appreciate the follow-up suggestions (shared constant, better Windows detection), happy to take those on as a separate PR if useful.

Mide69:
Docs updated: docs/configuration/docker.md

hectorvent:
Thank you, that is everything, and sorry for the wait. The CHANGELOG hunk is gone, the rebase cleared the merge conflict with it, and adding the note to `docs/configuration/docker.md` without being asked was the right call.

The test matrix is the part worth calling out. Eight cases across the full priority grid, both platforms, including explicit config beating the named pipe on Windows and `DOCKER_HOST` beating it too. That pins the ordering in both directions, so a future reshuffle of those branches cannot pass quietly.

On the Windows detection follow-up, a separate PR is fine and there is no rush. Worth knowing it is not actually broken today: `"darwin"` does contain `"win"`, but the JVM reports `Mac OS X` for `os.name` on macOS, never `Darwin`, so there is no false match. It is a tidy-up, not something you shipped wrong.

No blockers from my side.

hectorvent (approved):

PR #2032, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published images

kobihikri:
Hi, and thanks for floci.

`.github/workflows/release.yml` builds and pushes two images on release, but neither pushed manifest carries a provenance or SBOM attestation. Someone pulling either cannot check that it was built by this workflow, from this repository, at that release.

The image is what people deploy directly, with no intermediate build step that could vouch for it, so the pushed manifest is the artifact that actually reaches users.

The change is two lines on each of the two build steps:

```yaml
          push: ...
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so each build covers every registry it targets. **No permissions change is needed** — nothing has to gain `id-token`, and your tag and cache configuration are untouched.

```
docker buildx imagetools inspect <image>:<tag> --format '{{ json .Provenance }}'
```

Two caveats worth stating: `mode=max` records the full build including build arguments, so `provenance: true` is the smaller option if any have ever been sensitive; and attestations add an extra manifest to the index, which the registry supports.

No SLSA level claimed — the attestation is what BuildKit produces.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow and both build steps myself.


hectorvent (approved):
Thanks @kobihikri

kobihikri:
**Correction to my PR description — after you already merged it, which makes it more important rather than less.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". Half of that was wrong.

**Provenance was already there.** For public repositories `docker/build-push-action` adds provenance with `mode=max` by default — [Docker's docs](https://docs.docker.com/build/ci/github-actions/attestations/): *"Public repos: provenance attestations with `mode=max` are automatically added"*. So the `provenance: mode=max` line made existing behaviour explicit rather than adding it.

**The SBOM was genuinely new** — *"SBOM attestations aren't automatically added to the image"* — so that half of the change did what I said.

I also claimed in the caveats that `provenance: true` gives "a smaller record". Also wrong: `true` resolves to max on a public repo; the smaller setting is `provenance: mode=min`.

Nothing about the merged change is harmful — an explicit provenance line is still worth having, because the behaviour will not shift quietly if the default ever changes, and the SBOM is a real addition. But you merged on a description that overstated what it did, and you should know that.

Apologies. It was caught by a maintainer reviewing the same change on another project.

PR #1216, opened 2026-06-06
feat(iam): seed standard EKS cluster and node group managed policies

ousamabenyounes:
## Summary

EKS cluster creation and managed node groups attach a set of AWS managed policies that floci did not seed in its IAM catalog. `GetPolicy` / `AttachRolePolicy` for these therefore returned `NoSuchEntity`:

```
NoSuchEntity: Policy arn:aws:iam::aws:policy/AmazonEKSClusterPolicy does not exist.
NoSuchEntity: Policy arn:aws:iam::aws:policy/AmazonEKSVPCResourceController does not exist.
```

…which breaks the `terraform-aws-modules/eks` module and the EKS console flow.

This seeds the standard EKS set into the startup catalog (all path `/`):
`AmazonEKSClusterPolicy`, `AmazonEKSServicePolicy`, `AmazonEKSVPCResourceController`, `AmazonEKSWorkerNodePolicy`, `AmazonEKS_CNI_Policy`. As with the rest of the catalog they use the permissive wildcard document (floci does not enforce IAM evaluation by default).

Closes #1092

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore

## AWS Compatibility

These are real AWS managed-policy ARNs/names verified against the IAM console; all use path `/` (not `/service-role/`). Behavior matches real AWS: `GetPolicy` on each ARN now returns the policy, and `AttachRolePolicy` succeeds, so EKS IaC that references them provisions instead of failing with `NoSuchEntity`.

## Test verification (RED → GREEN)

The parameterized `getEksManagedPolicy` test was run locally with the catalog additions and again with them reverted, to confirm it catches the missing policies (CI only proves the with-fix case).

**RED** — EKS entries removed from `AwsManagedPolicies`, test kept (`mvn test -Dtest=IamIntegrationTest`):
```
[ERROR] IamIntegrationTest.getEksManagedPolicy(String)[1..5] <<< FAILURE!
[ERROR] Tests run: 32, Failures: 5, Errors: 0, Skipped: 0
[INFO] BUILD FAILURE
```
Exactly the 5 EKS ARNs fail (GetPolicy → not 200); the other 27 IAM cases stay green.

**GREEN** — with the seeded policies:
```
[INFO] Tests run: 32, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

## Checklist

[...]

PR #2034, opened 2026-07-29
feat(iam): seed Amazon Bedrock managed policies

### LINKED ISSUE #2033: [FEAT] IAM: Seed Amazon Bedrock managed policies (AmazonBedrockFullAccess, AmazonBedrockReadOnly)
### Feature

Floci seeds a catalog of common AWS-managed policies at startup (`AwsManagedPolicies`), but `AmazonBedrockFullAccess` and `AmazonBedrockReadOnly` are missing.

### Why it matters

CloudFormation attaches managed policies *inline* while creating a role, so a CFN-driven deploy never notices a missing catalog entry. Terraform (and any direct IAM API user) instead calls `AttachRolePolicy`, which validates the policy exists — and returns `404 NoSuchEntity` for these two:

```
Error: creating IAM Role (...): attaching IAM Policy
(arn:aws:iam::aws:policy/AmazonBedrockFullAccess) ... NoSuchEntity:
Policy arn:aws:iam::aws:policy/AmazonBedrockFullAccess does not exist.
```

Bedrock roles are common in agent/ML infrastructure, so the two general-access Bedrock policies are natural additions alongside the Bedrock AgentCore execution-role policy Floci already seeds.

### Proposed fix

Add `AmazonBedrockFullAccess` and `AmazonBedrockReadOnly` to the seeded catalog, matching the pattern used for the RDS (#1559) and EKS (#1216) additions.

PR-willingness: yes — opened alongside this issue.

lex00:
Fixes #2033.

Seeds `AmazonBedrockFullAccess` and `AmazonBedrockReadOnly` into the AWS-managed
policy catalog (`AwsManagedPolicies`), alongside the Bedrock AgentCore policy
already there.

**Why:** CloudFormation attaches managed policies inline on role creation, so a
CFN deploy never hits a missing catalog entry. Terraform and the raw IAM API call
`AttachRolePolicy`, which validates existence and `404`s (`NoSuchEntity`) for a
policy Floci hasn't seeded. These two Bedrock policies are common in agent/ML
roles.

**Tests:** `IamServiceTest.seedAwsManagedPolicies` asserts `AmazonBedrockFullAccess`
resolves; new `attachAmazonBedrockFullAccessToRole` covers the exact failing path
(create role → attach → assert attached). Both green.

Follows the single-policy-addition pattern of #1559 (RDS) and #1216 (EKS).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
@@ -152,9 +152,10 @@ final class AwsManagedPolicies {
 
         // Amazon Bedrock policies
         new ManagedPolicyDef("AmazonBedrockFullAccess", "/",
-                "Provides full access to Amazon Bedrock as well as limited access to related services."),
+                "Provides full access to Amazon Bedrock as well as limited access to related services "
+                + "that are required by it"),
         new ManagedPolicyDef("AmazonBedrockReadOnly", "/",
-                "Provides read-only access to Amazon Bedrock as well as limited access to related services."),
+                "Provides read only access to Amazon Bedrock"),
         new ManagedPolicyDef("AmazonBedrockAgentCoreMemoryBedrockModelInferenceExecutionRolePolicy", "/",
                 "Provides Bedrock Model inference permissions to Bedrock agent core memory."),
 
--- a/src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
@@ -150,7 +150,11 @@ final class AwsManagedPolicies {
         new ManagedPolicyDef("AmazonDataZoneDomainExecutionRolePolicy", "/service-role/",
                 "Provides permissions for the Amazon DataZone domain execution role."),
 
-        // Amazon Bedrock execution role policy
+        // Amazon Bedrock policies
+        new ManagedPolicyDef("AmazonBedrockFullAccess", "/",
+                "Provides full access to Amazon Bedrock as well as limited access to related services."),
+        new ManagedPolicyDef("AmazonBedrockReadOnly", "/",
+                "Provides read-only access to Amazon Bedrock as well as limited access to related services."),
         new ManagedPolicyDef("AmazonBedrockAgentCoreMemoryBedrockModelInferenceExecutionRolePolicy", "/",
                 "Provides Bedrock Model inference permissions to Bedrock agent core memory."),
 
--- a/src/test/java/io/github/hectorvent/floci/services/iam/IamServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/iam/IamServiceTest.java
@@ -615,6 +615,10 @@ class IamServiceTest {
         IamPolicy bedrock = iamService.getPolicy("arn:aws:iam::aws:policy/AmazonBedrockFullAccess");
         assertEquals("AmazonBedrockFullAccess", bedrock.getPolicyName());
         assertEquals("/", bedrock.getPath());
+
+        IamPolicy bedrockReadOnly = iamService.getPolicy("arn:aws:iam::aws:policy/AmazonBedrockReadOnly");
+        assertEquals("AmazonBedrockReadOnly", bedrockReadOnly.getPolicyName());
+        assertEquals("/", bedrockReadOnly.getPath());
     }
 
     @Test
--- a/src/test/java/io/github/hectorvent/floci/services/iam/IamServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/iam/IamServiceTest.java
@@ -611,6 +611,23 @@ class IamServiceTest {
                 "arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole");
         assertEquals("AmazonRDSEnhancedMonitoringRole", rdsMonitoring.getPolicyName());
         assertEquals("/service-role/", rdsMonitoring.getPath());
+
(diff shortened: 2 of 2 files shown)


src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
```diff
@@ -150,7 +150,11 @@ String arn() {
         new ManagedPolicyDef("AmazonDataZoneDomainExecutionRolePolicy", "/service-role/",
                 "Provides permissions for the Amazon DataZone domain execution role."),
 
-        // Amazon Bedrock execution role policy
+        // Amazon Bedrock policies
+        new ManagedPolicyDef("AmazonBedrockFullAccess", "/",
+                "Provides full access to Amazon Bedrock as well as limited access to related services."),
+        new ManagedPolicyDef("AmazonBedrockReadOnly", "/",
+                "Provides read-only access to Amazon Bedrock as well as limited access to related services."),
```
pgermosen: **Neither description matches what AWS publishes.**

| | Description |
|---|---|
| **AWS** `AmazonBedrockFullAccess` | `Provides full access to Amazon Bedrock as well as limited access to related services that are required by it` |
| **This PR** | `Provides full access to Amazon Bedrock as well as limited access to related services.` |
| **AWS** `AmazonBedrockReadOnly` | `Provides read only access to Amazon Bedrock` |
| **This PR** | `Provides read-only access to Amazon Bedrock as well as limited access to related services.` |

`AmazonBedrockFullAccess` is close — it drops "that are required by it" and adds a period AWS doesn't have.

`AmazonBedrockReadOnly` is the one that matters: the "as well as limited access to related services" clause looks copied from the FullAccess entry, but AWS's ReadOnly description is a short sentence that claims nothing about related services. (AWS also writes it unhyphenated, "read only", in this specific policy.)

This is worth getting exact because the surrounding entries are verbatim AWS strings — I checked `AmazonS3FullAccess` against the reference and it matches byte-for-byte — so the catalog reads as a faithful mirror, and `GetPolicy` hands these straight back to callers.

```suggestion
        new ManagedPolicyDef("AmazonBedrockFullAccess", "/",
                "Provides full access to Amazon Bedrock as well as limited access to related services "
                + "that are required by it"),
        new ManagedPolicyDef("AmazonBedrockReadOnly", "/",
                "Provides read only access to Amazon Bedrock"),
```

Sources: [AmazonBedrockFullAccess](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonBedrockFullAccess.html), [AmazonBedrockReadOnly](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonBedrockReadOnly.html)

src/test/java/io/github/hectorvent/floci/services/iam/IamServiceTest.java
```diff
@@ -611,6 +611,23 @@ void seedAwsManagedPolicies() {
                 "arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole");
         assertEquals("AmazonRDSEnhancedMonitoringRole", rdsMonitoring.getPolicyName());
         assertEquals("/service-role/", rdsMonitoring.getPath());
+
+        IamPolicy bedrock = iamService.getPolicy("arn:aws:iam::aws:policy/AmazonBedrockFullAccess");
+        assertEquals("AmazonBedrockFullAccess", bedrock.getPolicyName());
+        assertEquals("/", bedrock.getPath());
```
pgermosen, line 617: **Only one of the two new policies is covered.** *(nit)*

`seedAwsManagedPolicies` asserts `AmazonBedrockFullAccess` resolves and `attachAmazonBedrockFullAccessToRole` covers the attach path, but nothing touches `AmazonBedrockReadOnly` — which is also the entry with the incorrect description, so a test would not have caught it either way. Two more lines alongside the existing assertion would close it:

```suggestion
        IamPolicy bedrock = iamService.getPolicy("arn:aws:iam::aws:policy/AmazonBedrockFullAccess");
        assertEquals("AmazonBedrockFullAccess", bedrock.getPolicyName());
        assertEquals("/", bedrock.getPath());

        IamPolicy bedrockReadOnly = iamService.getPolicy("arn:aws:iam::aws:policy/AmazonBedrockReadOnly");
        assertEquals("AmazonBedrockReadOnly", bedrockReadOnly.getPolicyName());
        assertEquals("/", bedrockReadOnly.getPath());
```

The `attachRolePolicy` regression test is the valuable one and it's fine to leave that single-policy — a resolve assertion for the second entry is enough.

pgermosen (approved):
Looks good

Just a few minor comments

src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
lex00: Fixed at `5ec7d17`. Both now match the AWS-published text verbatim: `AmazonBedrockFullAccess` restores "that are required by it" and drops the trailing period, and `AmazonBedrockReadOnly` is now "Provides read only access to Amazon Bedrock" with the incorrect "as well as limited access to related services" clause removed. Good catch on ReadOnly, the clause was wrong rather than just imprecise.

src/test/java/io/github/hectorvent/floci/services/iam/IamServiceTest.java
lex00: Added. `seedAwsManagedPolicies` now resolves `AmazonBedrockReadOnly` and asserts its name and path alongside the FullAccess block.

lex00:
Both applied — the two descriptions are now the AWS strings verbatim, and `AmazonBedrockReadOnly` has a resolve assertion alongside the existing one.

hectorvent (approved):
Thanks @lex00

PR #2035, opened 2026-07-29
fix(rds): preserve configured database images

### LINKED ISSUE #1959: [BUG] cannot change the Postgres image name for RDS
## Service

RDS

## Expected behavior

You should be able to change the Postgres image name for RDS.

## Actual behavior

```
 _____  _      ___   ____   __
|  ___|| |    / _ \ / ___| | |
| |_   | |   | | | || |    | |
|  _|  | |___| |_| || |__  | |
|_|    |_____|\___/ \____| |_|

   AWS Local Emulator  ·  Always Free

            Powered by Quarkus 3.36.3
2026-07-26 13:33:39,595 INFO  [io.quarkus] floci 1.5.33 native (powered by Quarkus 3.36.3) started in 0.013s. Listening on: http://0.0.0.0:4566
2026-07-26 13:33:39,595 INFO  [io.quarkus] Profile prod activated.
2026-07-26 13:33:39,595 INFO  [io.quarkus] Installed features: [agroal, cdi, config-yaml, jdbc-mysql, jdbc-postgresql, narayana-jta, rest, rest-jackson, security, smallrye-context-propagation, vertx]
2026-07-26 13:33:40,669 ERROR [com.git.doc.api.asy.ResultCallbackTemplate] Error during callback: com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"failed to resolve reference \"docker.io/library/postgres:16.3-alpine3.23\": docker.io/library/postgres:16.3-alpine3.23: not found"}

        at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:241)
        at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
        at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
        at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)

2026-07-26 13:33:40,670 ERROR [io.git.hec.flo.ser.rds.RdsQueryHandler] Unexpected error in RDS CreateDBCluster: com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"failed to resolve reference \"docker.io/library/postgres:16.3-alpine3.23\": docker.io/library/postgres:16.3-alpine3.23: not found"}

        at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:241)
        at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
        at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
        at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)

2026-07-26 13:33:41,420 ERROR [com.git.doc.api.asy.ResultCallbackTemplate] Error during callback: com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"failed to resolve reference \"docker.io/library/postgres:16.3-alpine3.23\": docker.io/library/postgres:16.3-alpine3.23: not found"}

        at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:241)
        at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
        at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
        at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)

2026-07-26 13:33:41,421 ERROR [io.git.hec.flo.ser.rds.RdsQueryHandler] Unexpected error in RDS CreateDBCluster: com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"failed to resolve reference \"docker.io/library/postgres:16.3-alpine3.23\": docker.io/library/postgres:16.3-alpine3.23: not found"}

        at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:241)
        at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
        at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
        at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)

2026-07-26 13:33:42,153 ERROR [com.git.doc.api.asy.ResultCallbackTemplate] Error during callback: com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"failed to resolve reference \"docker.io/library/postgres:16.3-alpine3.23\": docker.io/library/postgres:16.3-alpine3.23: not found"}

        at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:241)
        at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
        at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
        at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)

2026-07-26 13:33:42,154 ERROR [io.git.hec.flo.ser.rds.RdsQueryHandler] Unexpected error in RDS CreateDBCluster: com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"failed to resolve reference \"docker.io/library/postgres:16.3-alpine3.23\": docker.io/library/postgres:16.3-alpine3.23: not found"}

        at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:241)
        at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
        at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
        at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)

2026-07-26 13:33:42,957 ERROR [com.git.doc.api.asy.ResultCallbackTemplate] Error during callback: com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"failed to resolve reference \"docker.io/library/postgres:16.3-alpine3.23\": docker.io/library/postgres:16.3-alpine3.23: not found"}

        at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:241)
        at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
        at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
        at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)

2026-07-26 13:33:42,957 ERROR [io.git.hec.flo.ser.rds.RdsQueryHandler] Unexpected error in RDS CreateDBCluster: com.github.dockerjava.api.exception.NotFoundException: Status 404: {"message":"failed to resolve reference \"docker.io/library/postgres:16.3-alpine3.23\": docker.io/library/postgres:16.3-alpine3.23: not found"}

        at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:241)
        at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
        at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
        at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)
```

## Reproduction

I was using floci-testcontainers to specify a specific Postgres image name. I don't think this issue is specific to testcontainers.

Note that the default image name is `postgres:16-alpine`, which is functionally the same as the one I specified `postgres:16.14-alpine3.23`.

```
        @Container
        @JvmStatic
        protected val flociContainer =
            FlociContainer()
                .withDedicatedNetwork()
                .withNetworkAliases(FLOCI_NETWORK_ALIAS)
                .withRdsConfig {
                    it
                        .enabled(true)
                        .defaultPostgresImage("postgres:16.14-alpine3.23")
                        .build()
                }
```

## Environment

floci `1.5.33`

aniketshukla1:
## Summary
- preserve explicitly configured RDS database image references instead of rewriting their tags from `EngineVersion`
- retain `EngineVersion` substitution for Floci's built-in RDS image defaults
- define built-in image defaults once in `EmulatorConfig.RdsServiceConfig` and document the configured-image behavior
- add a cluster creation regression test

Closes #1959

## Validation
- `JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw test -Dtest=RdsServiceTest,RdsQueryHandlerTest` (103 tests passed)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
@@ -1101,12 +1101,17 @@ public class RdsService implements Resettable {
     }
 
     private String imageForEngine(DatabaseEngine engine, String engineVersion) {
-        String defaultImage = switch (engine) {
-            case POSTGRES -> config.services().rds().defaultPostgresImage();
-            case MYSQL -> config.services().rds().defaultMysqlImage();
-            case MARIADB -> config.services().rds().defaultMariadbImage();
+        return switch (engine) {
+            case POSTGRES -> config.services().rds().defaultPostgresImage()
+                    .orElseGet(() -> imageForRequestedVersion(
+                            EmulatorConfig.RdsServiceConfig.DEFAULT_POSTGRES_IMAGE, engineVersion));
+            case MYSQL -> config.services().rds().defaultMysqlImage()
+                    .orElseGet(() -> imageForRequestedVersion(
+                            EmulatorConfig.RdsServiceConfig.DEFAULT_MYSQL_IMAGE, engineVersion));
+            case MARIADB -> config.services().rds().defaultMariadbImage()
+                    .orElseGet(() -> imageForRequestedVersion(
+                            EmulatorConfig.RdsServiceConfig.DEFAULT_MARIADB_IMAGE, engineVersion));
         };
-        return imageForRequestedVersion(defaultImage, engineVersion);
     }
 
     private void validateInstanceParameterGroup(String paramGroupName, String engineParam, String engineVersion) {
--- a/src/test/java/io/github/hectorvent/floci/services/rds/RdsServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/rds/RdsServiceTest.java
@@ -74,6 +74,7 @@ class RdsServiceTest {
     private Ec2Service ec2Service;
     private RegionResolver regionResolver;
     private EmulatorConfig config;
+    private EmulatorConfig.RdsServiceConfig rdsConfig;
 
     @BeforeEach
     void setUp() {
@@ -83,15 +84,15 @@ class RdsServiceTest {
         regionResolver = new RegionResolver("us-east-1", "123456789012");
         config = mock(EmulatorConfig.class);
         EmulatorConfig.ServicesConfig servicesConfig = mock(EmulatorConfig.ServicesConfig.class);
-        EmulatorConfig.RdsServiceConfig rdsConfig = mock(EmulatorConfig.RdsServiceConfig.class);
+        rdsConfig = mock(EmulatorConfig.RdsServiceConfig.class);
 
         when(config.services()).thenReturn(servicesConfig);
         when(servicesConfig.rds()).thenReturn(rdsConfig);
         when(rdsConfig.proxyBasePort()).thenReturn(7000);
         when(rdsConfig.proxyMaxPort()).thenReturn(7099);
-        when(rdsConfig.defaultPostgresImage()).thenReturn("postgres:16-alpine");
-        when(rdsConfig.defaultMysqlImage()).thenReturn("mysql:8.0");
-        when(rdsConfig.defaultMariadbImage()).thenReturn("mariadb:11");
+        when(rdsConfig.defaultPostgresImage()).thenReturn(Optional.empty());
+        when(rdsConfig.defaultMysqlImage()).thenReturn(Optional.empty());
+        when(rdsConfig.defaultMariadbImage()).thenReturn(Optional.empty());
 
         rdsService = newService(containerManager, proxyManager,
                 new InMemoryStorage<>(), new InMemoryStorage<>(),
@@ -163,6 +164,30 @@ class RdsServiceTest {
 
         verify(containerManager).start(eq("mydb"), any(), eq(DatabaseEngine.POSTGRES),
                 eq("postgres:18.1-alpine"), eq("admin"), eq("password"), eq("dbname"));
+    }
+
+    @Test
+    void configuredPostgresImageIsNotRewrittenForRequestedEngineVersion() {
+        when(rdsConfig.defaultPostgresImage()).thenReturn(Optional.of("postgres:16.14-alpine3.23"));
+
+        rdsService.createDbCluster("cluster1", "postgres", "16.3",
+                "admin", "password", "dbname", false, null);
+
+        verify(containerManager).start(eq("cluster1"), any(), eq(DatabaseEngine.POSTGRES),
+                eq("postgres:16.14-alpine3.23"), eq("admin"), eq("password"), eq("dbname"));
+    }
+
+    @Test
+    void explicitlyConfiguredDefaultPostgresImageIsNotRewritten() {
+        when(rdsConfig.defaultPostgresImage())
+                .thenReturn(Optional.of(EmulatorConfig.RdsServiceConfig.DEFAULT_POSTGRES_IMAGE));
+
+        rdsService.createDbCluster("cluster1", "postgres", "18.1",
+                "admin", "password", "dbname", false, null);
+
+        verify(containerManager).start(eq("cluster1"), any(), eq(DatabaseEngine.POSTGRES),
+                eq(EmulatorConfig.RdsServiceConfig.DEFAULT_POSTGRES_IMAGE),
+                eq("admin"), eq("password"), eq("dbname"));
     }
 
     @Test
--- a/docs/configuration/advanced/application-yml.md
+++ b/docs/configuration/advanced/application-yml.md
@@ -165,9 +165,11 @@ floci:
       proxy-base-port: 7001
       proxy-max-port: 7099
       # endpoint-host: localhost              # Hostname clients use; enables published-port translation in Docker
-      default-postgres-image: "postgres:16-alpine"
-      default-mysql-image: "mysql:8.0"
-      default-mariadb-image: "mariadb:11"
+      # Omit image overrides to adapt Floci's built-in images to EngineVersion.
+      # Uncomment an override to pin that engine to an exact image.
+      # default-postgres-image: "registry.example.com/postgres:16-alpine"
+      # default-mysql-image: "registry.example.com/mysql:8.0"
+      # default-mariadb-image: "registry.example.com/mariadb:11"
 
     rds-data:
       enabled: true
--- a/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
+++ b/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
@@ -877,6 +877,10 @@ public interface EmulatorConfig {
     }
 
     interface RdsServiceConfig {
+        String DEFAULT_POSTGRES_IMAGE = "postgres:16-alpine";
+        String DEFAULT_MYSQL_IMAGE = "mysql:8.0";
+        String DEFAULT_MARIADB_IMAGE = "mariadb:11";
+
         @WithDefault("true")
         boolean enabled();
 
@@ -892,14 +896,14 @@ public interface EmulatorConfig {
         @WithDefault("7099")
         int proxyMaxPort();
 
-        @WithDefault("postgres:16-alpine")
-        String defaultPostgresImage();
+        /** Empty when Floci should adapt its built-in image to the requested engine version. */
+        Optional<String> defaultPostgresImage();
 
-        @WithDefault("mysql:8.0")
-        String defaultMysqlImage();
+        /** Empty when Floci should adapt its built-in image to the requested engine version. */
+        Optional<String> defaultMysqlImage();
 
-        @WithDefault("mariadb:11")
-        String defaultMariadbImage();
+        /** Empty when Floci should adapt its built-in image to the requested engine version. */
+        Optional<String> defaultMariadbImage();
 
         /** Hostname advertised for RDS endpoints. Uses published Docker ports when configured. */
         Optional<String> endpointHost();
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -269,9 +269,11 @@ floci:
       proxy-base-port: 7001
       proxy-max-port: 7099
       # endpoint-host: localhost                 # Hostname clients use; enables published-port translation in Docker
-      default-postgres-image: "postgres:16-alpine"
-      default-mysql-image: "mysql:8.0"
-      default-mariadb-image: "mariadb:11"
+      # Omit image overrides to adapt Floci's built-in images to EngineVersion.
+      # Uncomment an override to pin that engine to an exact image.
+      # default-postgres-image: "registry.example.com/postgres:16-alpine"
+      # default-mysql-image: "registry.example.com/mysql:8.0"
+      # default-mariadb-image: "registry.example.com/mariadb:11"
     rds-data:
       enabled: true
       transaction-ttl-seconds: 180
--- a/src/test/resources/application.yml
+++ b/src/test/resources/application.yml
@@ -145,9 +145,6 @@ floci:
       mock: false
       proxy-base-port: 7001
       proxy-max-port: 7099
-      default-postgres-image: "postgres:16-alpine"
-      default-mysql-image: "mysql:8.0"
-      default-mariadb-image: "mariadb:11"
     rds-data:
       enabled: true
       transaction-ttl-seconds: 180


aniketshukla1:
Implemented the configured-image behavior from #1959. Custom image references now pass unchanged to the RDS container manager; Floci's built-in defaults remain `EngineVersion`-aware.

Validated with `RdsServiceTest` on JDK 25: 57 tests passed.

src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
```diff
@@ -989,12 +992,22 @@ private DatabaseEngine resolveEngine(String engineParam) {
...
+                : configuredImage;
+    }
+
+    private static String builtInDefaultImage(DatabaseEngine engine) {
+        return switch (engine) {
+            case POSTGRES -> DEFAULT_POSTGRES_IMAGE;
+            case MYSQL -> DEFAULT_MYSQL_IMAGE;
+            case MARIADB -> DEFAULT_MARIADB_IMAGE;
+        };
     }
```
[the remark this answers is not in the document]
aniketshukla1: Addressed in 445eebb. The built-in RDS image tags now have one canonical definition in `EmulatorConfig.RdsServiceConfig`; `RdsService` references those constants, and the duplicated runtime YAML values were removed. This retains `EngineVersion` substitution for defaults while preserving explicit configured images.

src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
```diff
@@ -989,12 +989,22 @@ private DatabaseEngine resolveEngine(String engineParam) {
...
 
     private String imageForEngine(DatabaseEngine engine, String engineVersion) {
-        String defaultImage = switch (engine) {
+        String configuredImage = switch (engine) {
             case POSTGRES -> config.services().rds().defaultPostgresImage();
             case MYSQL -> config.services().rds().defaultMysqlImage();
             case MARIADB -> config.services().rds().defaultMariadbImage();
         };
-        return imageForRequestedVersion(defaultImage, engineVersion);
+        return configuredImage.equals(builtInDefaultImage(engine))
```
pgermosen: **The "was this configured?" test is value equality with the default — and `EmulatorConfig` already has the idiomatic answer.**

`configuredImage.equals(builtInDefaultImage(engine))` asks *"does this value happen to equal the default?"* when the intent is *"did the user set this?"*. Two consequences:

- **An explicit pin to a default value is silently ignored.** A user who writes `default-postgres-image: "postgres:16-alpine"` — meaning "always exactly this, don't touch it" — gets tag-rewriting anyway, because the string matches. That intent becomes unexpressible for precisely the three values most likely to be typed.
- **Changing a built-in default silently changes behavior for existing users.** If `DEFAULT_POSTGRES_IMAGE` moves to `postgres:17-alpine`, everyone still pinned to `postgres:16-alpine` flips from "rewritten" to "verbatim" without touching their config.

`EmulatorConfig` is a SmallRye `@ConfigMapping`, and this file already answers "was it explicitly set?" with `Optional<String>` in eight places — `hostname()` at `:28`, `rootPermissions()` at `:228`, `mountUser()` at `:239`, `mode()` at `:286`/`:293`, and others. Dropping `@WithDefault` here gives a real signal:

```java
Optional<String> defaultPostgresImage();
```

```java
private String imageForEngine(DatabaseEngine engine, String engineVersion) {
    Optional<String> configured = switch (engine) {
        case POSTGRES -> config.services().rds().defaultPostgresImage();
        case MYSQL -> config.services().rds().defaultMysqlImage();
        case MARIADB -> config.services().rds().defaultMariadbImage();
    };
    // An explicitly configured image is used verbatim; only the built-in default adapts to EngineVersion.
    return configured.orElseGet(() -> imageForRequestedVersion(builtInDefaultImage(engine), engineVersion));
}
```

That drops the `.equals` comparison, keeps a single switch, and makes the yml removals unnecessary — the built-in default lives only in the constant, which is what the PR description says it's aiming for.

One thing to watch if you take this: under an `Optional` model, re-adding those keys to `src/main/resources/application.yml` as documentation would make them "explicitly configured" for every user and disable version substitution globally. Under the current string-equality model re-adding them is harmless. Either way the removal deserves a `#` comment in the yml explaining why the keys are deliberately absent, or someone will helpfully restore them.

---

**Separately — this method collides with #1920.** That PR adds `AURORA_POSTGRESQL` / `AURORA_MYSQL` to `DatabaseEngine` and rewrites this same switch to `case POSTGRES, AURORA_POSTGRESQL -> ...`. I merged the branches to check: `RdsService.java` conflicts, and after resolving it in the obvious way the build fails with

```
RdsService.java:[1007,16] the switch expression does not cover all possible input values
```

because `builtInDefaultImage` below is a second exhaustive switch that #1920 doesn't know to update. Whoever merges second needs to touch both. Adopting the `Optional` shape above makes `builtInDefaultImage` unnecessary and shrinks this to a one-line conflict.

src/test/java/io/github/hectorvent/floci/services/rds/RdsServiceTest.java
```diff
@@ -139,6 +140,17 @@ void createDbInstanceStartsContainerWithRequestedEngineVersionImage() {
                 eq("postgres:18.1-alpine"), eq("admin"), eq("password"), eq("dbname"));
     }
 
+    @Test
+    void configuredPostgresImageIsNotRewrittenForRequestedEngineVersion() {
+        when(rdsConfig.defaultPostgresImage()).thenReturn("postgres:16.14-alpine3.23");
+
+        rdsService.createDbCluster("cluster1", "aurora-postgresql", "16.3",
+                "admin", "password", "dbname", false, null);
```
pgermosen: **The test asserts through the aurora alias, which couples it to #1920.** *(nit)*

`createDbCluster(..., "aurora-postgresql", ...)` only reaches the postgres image because `resolveEngine` currently maps `aurora-postgresql` to `POSTGRES`. #1920 changes that to `AURORA_POSTGRESQL`, so this test starts depending on the aurora branch of a switch it isn't trying to exercise.

The behavior under test is "a configured image isn't rewritten", which has nothing to do with aurora aliasing — `"postgres"` would test it directly and survive the #1920 merge unchanged.

```suggestion
        rdsService.createDbCluster("cluster1", "postgres", "16.3",
                "admin", "password", "dbname", false, null);
```

pgermosen (commented):
The behavior this targets is right — an explicitly pinned image shouldn't have its tag rewritten from `EngineVersion`. `RdsServiceTest` + `RdsQueryHandlerTest` pass locally: 103/103, 0 failures, matching the count in your description.

Two things, one of which is a heads-up you can't see from inside this PR.

**1. It collides with #1920, and the merge does not compile.** #1920 (`fix(rds): preserve aurora engine types`) rewrites the same `imageForEngine` switch and adds `AURORA_POSTGRESQL` / `AURORA_MYSQL` to `DatabaseEngine`. I merged the two branches to check: `RdsService.java` conflicts textually, and after resolving it the obvious way the build still fails —

```
RdsService.java:[1007,16] the switch expression does not cover all possible input values
```

— because the new `builtInDefaultImage` is a *second* exhaustive switch over `DatabaseEngine`, and whoever merges second has to remember to update both. Not your bug, but worth coordinating with @dnlopes. Details in the inline comment.

**2. Detecting "was this configured?" by comparing against the default value is fragile, and this file already has the idiomatic answer.** `EmulatorConfig` is a SmallRye `@ConfigMapping` and already uses `Optional<String>` for exactly this question in eight places (`hostname()`, `rootPermissions()`, `mountUser()`, `mode()`, …). Using it here removes the constants, removes `builtInDefaultImage`, removes the need to strip the values out of both `application.yml` files — and, conveniently, removes most of the #1920 collision too.

Neither is a correctness bug in what you've written; the tests genuinely pass and the intended behavior works. **I'd like #2 reconsidered before merge** since it changes the shape of the fix, and #1 sequenced with #1920 either way.

dnlopes:
Hello @pgermosen @aniketshukla1 , I noticed this PR and mine (#1920) conflict and one of us will need to adapt. Just to be 100% transparent, I currently don't have enough bandwidth to get involved on this discussion (the remaining of this week and likely the next week), so feel free to pick whatever approach you think it's better. I will come back to my PR afterwards and change what needs to be changed.

pgermosen (approved):
Approving on `17f23767`. Thank you for taking the reshape — it turned out to matter more than I realised when I suggested it.

**The `Optional<String>` version is the right fix.** `imageForEngine` is a single switch again, `builtInDefaultImage` is gone, and the built-in defaults live in exactly one place. The test I most wanted is the one that proves why the shape change was worth the churn:

```java
void explicitlyConfiguredDefaultPostgresImageIsNotRewritten() {
    when(rdsConfig.defaultPostgresImage())
            .thenReturn(Optional.of(EmulatorConfig.RdsServiceConfig.DEFAULT_POSTGRES_IMAGE));
```

Configuring the value that happens to *equal* the built-in default is still honoured as a pin. That is precisely the case compare-against-the-default could never get right, and you covered it without being asked.

Local on this head: `RdsServiceTest` 95/95, `RdsQueryHandlerTest` 51/51. All 13 CI checks green.

## Retracting what I said about #1920

In my last review I said the merge with #1920 does not compile, and used that to argue for sequencing. I re-ran it rather than repeating it, and I had it wrong in a way that matters:

**The #2035 ↔ #1920 conflict is now trivial.** With `builtInDefaultImage` removed there is only one switch left, so resolution is folding `AURORA_POSTGRESQL` / `AURORA_MYSQL` into the existing cases. I performed the merge and resolved it in three lines. The original compile break I reported was caused by the *second* exhaustive switch, which this rewrite deleted — so your reshape fixed the collision as a side effect.

**The remaining break belongs to #1920 alone.** I merged `origin/main` with #1920 *without* this PR and got the identical error:

```
RdsDataService.java:[230,16] the switch expression does not cover all possible input values
```

This PR does not touch that file. `usesBackslashEscapes`, a new exhaustive `DatabaseEngine` switch, landed in `19a32ab5` on 2026-08-12 (`feat(rds-data): adding data-api preparedstatement() support`, #1887) — two weeks after #1920 was last updated. #1920's new enum constants make it non-exhaustive.

So the sequencing inverts from what I told you: **this one goes first.** Nothing is blocked on #1920, and no coordination is needed from you.

@dnlopes — flagging for whenever you pick #1920 back up, since this changes what you'll find. It needs a rebase on `main` regardless of this PR: #1887 added a `DatabaseEngine` switch in `RdsDataService` that your two new enum constants leave non-exhaustive. While you are there, the `imageForEngine` conflict with this PR is a three-line fold (`case POSTGRES, AURORA_POSTGRESQL -> …orElseGet(…)`). Apologies for the earlier framing that implied the two of you had to negotiate — that is no longer true. Thank you for the transparency about bandwidth; it made this easy to sequence around.

## One non-blocking thing worth a follow-up

There is a quiet migration edge here. Floci previously *shipped* `application.yml` with those three values uncommented, so anyone running a copy of that file is now, under the new rule, explicitly pinning. For them `EngineVersion: 15` that used to yield `postgres:15-alpine` will yield `postgres:16-alpine` instead — silently.

That is the correct reading of "configured means configured", and `explicitlyConfiguredDefaultPostgresImageIsNotRewritten` deliberately encodes it. It is also inherent to the approach I pushed you toward, so it is mine as much as yours. The docs diff shows the new commented-out form but does not say what happens if you kept the old one. A sentence in `docs/configuration/advanced/application-yml.md` — or a release note — would save someone a confused afternoon. Happy for that to be a follow-up rather than another round here.

Tiny: the description's validation line still says 103 tests, which was the pre-rewrite count.

Good patience on this one — it sat for two weeks through a shape change and a collision that turned out not to be yours.

PR #2037, opened 2026-07-29
feat(cloudcontrol): implement CreateResource / DeleteResource / GetResource

### LINKED ISSUE #2036: [FEAT] Cloud Control API: implement CreateResource / DeleteResource / GetResource
### Feature

Floci's Cloud Control handler (`CloudControlJsonHandler`) implements only `ListResources`. `CreateResource`, `DeleteResource`, `GetResource`, and `GetResourceRequestStatus` return `UnsupportedOperation`.

### Why it matters

A growing set of tools drive AWS through the Cloud Control API (`cloudcontrolapi`) rather than the individual service APIs — a unified CRUD surface over CloudFormation resource types. Against Floci today, any such tool fails immediately on `CreateResource`:

```
operation error CloudControl: CreateResource, StatusCode: 400,
UnsupportedOperation: Operation CreateResource is not supported.
```

Floci already has everything needed to back this: `CloudFormationResourceProvisioner` provisions 114 resource types for CloudFormation stacks. Cloud Control write is a thin adapter over it — any type a stack can create, Cloud Control can too.

### Proposed implementation

- `CreateResource` → the shared provisioner (a standalone single-resource provision, no enclosing stack). Cloud Control is async, so it returns an `IN_PROGRESS` ProgressEvent + a request token; provisioning runs in the background and `GetResourceRequestStatus` reports the terminal state. The SUCCESS event carries the created `Identifier` and a `ResourceModel`.
- `DeleteResource` → the provisioner's existing delete path.
- `GetResource` → the existing read side, by identifier.

PR-willingness: yes — opened alongside this issue.

lex00:
Fixes #2036.

Cloud Control write operations, wired to the existing `CloudFormationResourceProvisioner` (the same 114-type provisioning CloudFormation stacks use). Any resource type a stack can create, Cloud Control can now create too.

**Design**
- Two standalone helpers on the provisioner (`provisionStandalone` / `deleteStandalone`) — no enclosing stack; Cloud Control DesiredState carries resolved values, so a minimal template engine suffices.
- `CreateResource` is **async** (real Cloud Control is): it returns `IN_PROGRESS` + a request token immediately and provisions on a background executor, so slow resources (an EC2 instance launches a container) don't blow a client's synchronous-call deadline. `GetResourceRequestStatus` resolves the token to the terminal `SUCCESS`/`FAILED` ProgressEvent, which carries the `Identifier`, a `ResourceModel`, and any failure `StatusMessage`.
- `DeleteResource` uses the provisioner's existing delete; `GetResource` reads a single resource by identifier off the existing read side.

**Tests**
`CloudControlIntegrationTest.createResourceProvisionsViaCloudControlAndTokenRoundTrips`: CreateResource(AWS::EC2::VPC) → IN_PROGRESS → poll GetResourceRequestStatus to SUCCESS → the VPC is listable. Verified manually that VPC, subnet, and EC2 instance all provision through the raw Cloud Control path.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudcontrol/CloudControlJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudcontrol/CloudControlJsonHandler.java
@@ -66,19 +66,20 @@ public class CloudControlJsonHandler {
         return Response.ok(response).build();
     }
 
+    /**
+     * InvalidRequestException is the code Cloud Control declares for a malformed request, so it is
+     * what an SDK maps onto a typed exception. ValidationException is not in the service model.
+     */
     private String required(JsonNode request, String field) {
         String value = request.path(field).asText(null);
         if (value == null || value.isBlank()) {
-            throw new AwsException("ValidationException", field + " is required.", 400);
+            throw new AwsException("InvalidRequestException", field + " is required.", 400);
         }
         return value;
     }
 
     private Response listResources(JsonNode request, String region) {
-        String typeName = request.path("TypeName").asText(null);
-        if (typeName == null || typeName.isBlank()) {
-            throw new AwsException("ValidationException", "TypeName is required.", 400);
-        }
+        String typeName = required(request, "TypeName");
         ObjectNode response = mapper.createObjectNode();
         response.put("TypeName", typeName);
         ArrayNode resources = response.putArray("ResourceDescriptions");
--- a/src/main/java/io/github/hectorvent/floci/services/cloudcontrol/CloudControlJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudcontrol/CloudControlJsonHandler.java
@@ -24,9 +24,54 @@ public class CloudControlJsonHandler {
     public Response handle(String action, JsonNode request, String region) {
         return switch (action) {
             case "ListResources" -> listResources(request, region);
+            case "GetResource" -> getResource(request, region);
+            case "CreateResource" -> progressResponse(
+                    service.createResource(region, required(request, "TypeName"),
+                            request.path("DesiredState").asText(null)));
+            case "DeleteResource" -> progressResponse(
+                    service.deleteResource(region, required(request, "TypeName"), required(request, "Identifier")));
+            case "GetResourceRequestStatus" -> progressResponse(
+                    service.requestStatus(required(request, "RequestToken")));
             default -> throw new AwsException("UnsupportedOperation",
                     "Operation " + action + " is not supported.", 400);
         };
+    }
+
+    private Response getResource(JsonNode request, String region) {
+        String typeName = required(request, "TypeName");
+        String identifier = required(request, "Identifier");
+        CloudControlService.ResourceDescription resource = service.getResource(region, typeName, identifier);
+        ObjectNode response = mapper.createObjectNode();
+        response.put("TypeName", typeName);
+        ObjectNode desc = response.putObject("ResourceDescription");
+        desc.put("Identifier", resource.identifier());
+        desc.put("Properties", resource.properties());
+        return Response.ok(response).build();
+    }
+
+    private Response progressResponse(CloudControlService.ProgressEvent event) {
+        ObjectNode response = mapper.createObjectNode();
+        ObjectNode pe = response.putObject("ProgressEvent");
+        pe.put("TypeName", event.typeName());
+        pe.put("Identifier", event.identifier());
+        pe.put("RequestToken", event.requestToken());
+        pe.put("Operation", event.operation());
+        pe.put("OperationStatus", event.operationStatus());
+        if (event.statusMessage() != null) {
+            pe.put("StatusMessage", event.statusMessage());
+        }
+        if (event.resourceModel() != null) {
+            pe.put("ResourceModel", event.resourceModel());
+        }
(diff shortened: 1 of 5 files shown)


lex00:
All four findings are fixed at `68234bc`.

**GetResource misses provisioned types** — CreateResource now records what it provisioned, and GetResource falls back to that record when the read side does not list the type. Verified the new test fails without it: `AWS::EC2::InternetGateway` is provisionable but unlisted, so a create followed by a get returned ResourceNotFoundException.

**Delete skips attribute-backed cleanup** — `deleteStandalone` gained an overload taking the create-time attributes, and Cloud Control passes through what it recorded. When those attributes are absent for a type that needs them (custom resources, `AWS::EKS::Nodegroup`, `AWS::IAM::Policy`) the operation now reports FAILED with the reason rather than SUCCESS over a resource that is still there.

**Executor shutdown** and **unbounded token map** — added `@PreDestroy` on the executor, and the map now evicts the oldest finished tokens past a cap while never dropping one a client is still polling.

184 tests across the Cloud Control and CloudFormation suites pass.

lex00:
Folded #2062 in here and closed it, to bring my open-PR count down.

That adds `ListResources` coverage for `AWS::EC2::Instance`, `AWS::EC2::LaunchTemplate` and `AWS::IAM::InstanceProfile` with their runtime models. It sits naturally alongside this PR — #2062's read path and this one's `GetResource` fallback are the same concern, and both branches were editing `CloudControlService` and `CloudControlIntegrationTest` anyway.

Rebased on current `main`. 319 tests pass across the Cloud Control, EC2 and CloudFormation suites.

Note that this only partly addresses #2043: it adds the three types, but not the other half of that issue — distinguishing an unsupported type from a supported type with no resources.

hectorvent:
Thank you, this iteration is a big step up: the executor now shuts down cleanly, the token map is bounded without evicting in-flight requests, GetResource reads back types the list side does not cover, and refusing to report SUCCESS for a delete that would silently no-op is exactly the right instinct.

Against the Cloud Control model (https://github.com/boto/botocore/blob/develop/botocore/data/cloudcontrol/2021-09-30/service-2.json):

1. **(blocking)** Missing TypeName/Identifier/RequestToken throws `ValidationException` and malformed DesiredState throws `InvalidRequest`; neither code is declared, so SDKs cannot map them to typed exceptions. The declared code for both is `InvalidRequestException`.

2. **(blocking)** `DesiredState` is a required member of CreateResourceInput, but a missing or blank one currently becomes `{}` and provisions anyway; it should be rejected with `InvalidRequestException`.

3. **(follow-up, separate PR)** ProgressEvent supports `EventTime` and, on failures, `ErrorCode`; emitting even a coarse ErrorCode would let SDK clients branch on failure kind instead of parsing StatusMessage.

Both blockers are small, localized changes in CloudControlJsonHandler/CloudControlService; everything else looks ready.

lex00:
Both are fixed in e899c303. Missing TypeName, Identifier or RequestToken and malformed DesiredState now report InvalidRequestException, which is the code the model declares. A missing or blank DesiredState is rejected instead of becoming an empty object and provisioning. Integration tests cover both.

Left EventTime and ErrorCode for the follow-up PR as you suggested.

hectorvent:
Thank you, and apologies for the delay. Both asks landed correctly: `InvalidRequestException` across the four operations that declare it, and `DesiredState` now rejected when absent instead of provisioning an empty object. CI is green.

One correction, and the mistake was mine. I asked for `RequestToken` to throw `InvalidRequestException` too, but `GetResourceRequestStatus` does not declare it. Its only declared error is `RequestTokenNotFoundException` ([Cloud Control model](https://github.com/boto/botocore/blob/develop/botocore/data/cloudcontrol/2021-09-30/service-2.json)). Sorry for sending you down that path.

1. **(blocking)** That call wants its own check rather than the shared `required()` helper. `CloudControlIntegrationTest` line 156 asserts the old code, so it needs loosening in the same change.

2. **(follow-up, separate PR)** `progressResponse` emits `"Identifier": null` on an IN_PROGRESS create where AWS omits absent optionals; `EventTime` and `ErrorCode` are still unset.

That one code is all that is left.

lex00:
Fixed. GetResourceRequestStatus has its own check now and reports RequestTokenNotFoundException for an absent token, which is the only error it declares. The integration test asserts that code for that call and keeps InvalidRequestException for the rest. Left the null Identifier and the EventTime and ErrorCode items for the follow-up.

hectorvent:
Thank you, that is exactly right, and the status code is right too. `GetResourceRequestStatus` has its own check now reporting `RequestTokenNotFoundException`, and 404 is correct: botocore carries no status codes for these shapes, but AWS's upstream Smithy model gives that error `httpError` 404 and `InvalidRequestException` 400. The four operations that do declare `InvalidRequestException` keep it, and the test pinning each call separately is what makes the distinction stick.

**(follow-up, separate PR)** One to add to the list you already have. `createResource` provisions on a background thread with the account fixed to `000000000000`, while the CloudFormation path runs the same provisioner under a request scope bound to the caller's account (`CloudFormationService.runUnderAccount`). Account aware storage falls back to the default account rather than failing, so single account setups behave the same either way, but a multi account caller gets resources under the wrong one.

No blockers from my side.

hectorvent (approved):
Thanks @lex00

PR #2039, opened 2026-07-29
fix(cloudformation): tolerate missing stack resources during deletion

### LINKED ISSUE #1966: [BUG] CloudFormation DeleteStack fails with DELETE_FAILED when a resource is already gone, instead of treating it as deleted
## Service

CloudFormation

## AWS API Action

DeleteStack

## Expected behavior

When a stack resource no longer exists, `DeleteStack` should treat it as already deleted and continue, leaving the stack in `DELETE_COMPLETE`. Real CloudFormation does this for resources removed out of band, and LocalStack does the same through `check_not_found_exception`, whose marker list includes `NotFoundException`, `404` and `not exist`.

## Actual behavior

Any exception from a resource delete flips the stack to `DELETE_FAILED`, including a plain 404:

- `CloudFormationResourceProvisioner.java:438` calls `apiGatewayService.deleteRestApi(region, physicalId)`
- `ApiGatewayService.java:273` calls `getRestApi(...)`, which throws `NotFoundException` / `Invalid API id specified` (`ApiGatewayService.java:253`)
- `CloudFormationService.deleteStackResources` (`:644`) catches `Exception` and marks the resource `DELETE_FAILED`

```json
{ "message": "Invalid API id specified" }
```

```
DELETE_FAILED
The following resource(s) failed to delete: [AssetManagementAPI].
```

The result is a stack that can never be deleted through the API once any of its resources has gone missing, whatever the cause. Surfaced by #1854, where persisted API Gateway state was lost on restart, but the delete path is the part that is still wrong: #1854 itself is fixed in 1.5.32.

## Suggested fix

Treat a not-found error from a resource delete as `DELETE_COMPLETE` in `CloudFormationService.deleteStackResources`, rather than adding another one-off `*Safe` wrapper in the provisioner. Floci already has that pattern per type (`deleteRoleSafe`, `deleteSecretSafe`, `deleteEventBridgeRuleSafe`, `deleteEcsClusterSafe`), so a generic rule at the call site covers every resource type at once and keeps genuine failures like `BucketNotEmpty` failing as they do today.

Related: #1913, same class of "delete path is not resilient to restored or missing state".

aniketshukla1:
## Summary
- treat a resource-delete 404 or not-found response as an already completed delete during both `DeleteStack` and create rollback
- retain `DELETE_FAILED` and `ROLLBACK_FAILED` behavior for genuine deletion errors
- add API Gateway regressions for an out-of-band REST API delete and rollback cleanup of a resource that has already disappeared

Closes #1966

## Validation
- `env JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home/bin:$PATH ./mvnw test -Dtest=CloudFormationServiceRollbackTest,CloudFormationIntegrationTest#deleteStack_withAlreadyDeletedApiGatewayRestApi_reachesDeleteComplete,CloudFormationEcsDeleteAfterRestoreIntegrationTest`
- `git diff --check`

## Notes
- The full `CloudFormationIntegrationTest` class was also run. Its new regression path passed; one unrelated ECR test failed because Docker is unavailable locally and cannot start the ECR backing registry.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationServiceRollbackTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationServiceRollbackTest.java
@@ -1,0 +1,101 @@
+package io.github.hectorvent.floci.services.cloudformation;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.hectorvent.floci.config.EmulatorConfig;
+import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.core.common.RegionResolver;
+import io.github.hectorvent.floci.core.storage.AccountAwareStorageBackend;
+import io.github.hectorvent.floci.core.storage.InMemoryStorage;
+import io.github.hectorvent.floci.core.storage.StorageBackend;
+import io.github.hectorvent.floci.core.storage.StorageFactory;
+import io.github.hectorvent.floci.services.cloudformation.model.Stack;
+import io.github.hectorvent.floci.services.cloudformation.model.StackResource;
+import io.github.hectorvent.floci.services.s3.S3Service;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.time.Clock;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Covers rollback cleanup when another actor removes a resource after its create succeeded.
+ */
+class CloudFormationServiceRollbackTest {
+
+    private static final String ACCOUNT = "000000000000";
+    private static final String REGION = "us-east-1";
+
+    private CloudFormationResourceProvisioner provisioner;
+    private CloudFormationService service;
+
+    @BeforeEach
+    void setUp() {
+        provisioner = mock(CloudFormationResourceProvisioner.class);
+        EmulatorConfig config = mock(EmulatorConfig.class);
+        when(config.defaultAccountId()).thenReturn(ACCOUNT);
+
+        service = new CloudFormationService(
+                provisioner,
+                mock(S3Service.class),
+                new ObjectMapper(),
+                config,
+                mock(RegionResolver.class),
+                Clock.systemUTC(),
+                new InMemoryStorageFactory());
+    }
+
+    @Test
+    void createRollback_withAlreadyDeletedResource_reachesRollbackComplete() {
+        Stack stack = new Stack();
+        stack.setStackName("rollback-missing-resource");
+        stack.setStackId("stack-id");
+        stack.setRegion(REGION);
+
+        StackResource created = resource("RestApi", "api-id", "AWS::ApiGateway::RestApi", "CREATE_COMPLETE");
+        StackResource failed = resource("FailingResource", null, "AWS::Test::Failure", "CREATE_FAILED");
+        failed.setStatusReason("simulated create failure");
+        stack.getResources().put(created.getLogicalId(), created);
+        stack.getResources().put(failed.getLogicalId(), failed);
+
+        doThrow(new AwsException("NotFoundException", "Invalid API id specified", 404))
+                .when(provisioner).delete(eq(created), eq(REGION));
+
+        service.rollbackFailedExecution(stack, REGION, true, failed, null, Set.of());
+
+        assertEquals("ROLLBACK_COMPLETE", stack.getStatus());
+        assertEquals("DELETE_COMPLETE", created.getStatus());
+        assertNull(created.getStatusReason());
+        verify(provisioner).delete(created, REGION);
+    }
+
+    private static StackResource resource(String logicalId, String physicalId, String resourceType, String status) {
+        StackResource resource = new StackResource();
+        resource.setLogicalId(logicalId);
+        resource.setPhysicalId(physicalId);
+        resource.setResourceType(resourceType);
+        resource.setStatus(status);
+        return resource;
+    }
+
+    private static final class InMemoryStorageFactory extends StorageFactory {
+        private InMemoryStorageFactory() {
+            super(null, null);
+        }
+
+        @Override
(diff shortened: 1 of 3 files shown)


aniketshukla1:
Implemented #1966 with a stack-lifecycle not-found path and an API Gateway out-of-band deletion regression. Focused CloudFormation deletion tests pass under Java 25; the broader class has one unrelated ECR Docker-unavailable failure documented in the PR description.

aniketshukla1:
Addressed Greptile’s rollback finding in 26980af. Already-missing resources now complete during both DeleteStack and create rollback, with a hermetic 404 rollback regression in CloudFormationServiceRollbackTest. The focused four-test CloudFormation suite passes under Java 25.

src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationServiceRollbackTest.java
```diff
@@ -0,0 +1,108 @@
...
+        assertEquals("ROLLBACK_COMPLETE", stack.getStatus());
+        assertEquals("DELETE_COMPLETE", created.getStatus());
+        assertNull(created.getStatusReason());
+        verify(provisioner).delete(created, REGION);
+    }
+
+    private void rollbackFailedCreate(Stack stack, StackResource failed) throws Exception {
+        Method method = CloudFormationService.class.getDeclaredMethod(
+                "rollbackFailedExecution", Stack.class, String.class, boolean.class, StackResource.class);
+        method.setAccessible(true);
```
pgermosen: Nit

This reaches `rollbackFailedExecution` via reflection (`getDeclaredMethod` + `setAccessible(true)`) because it's `private` in `CloudFormationService` (line 577, unchanged by this PR, so I can't leave a suggestion on it directly). Reflection here means the test won't get a compile error if the method is ever renamed or its parameter list changes — it'll fail at runtime instead. Since this test already lives in the same package (`io.github.hectorvent.floci.services.cloudformation`), making `rollbackFailedExecution` package-private in a follow-up would let this call it directly with normal compile-time safety, no functional change needed.

pgermosen (dismissed):
Looks good, just a nit comment.

aniketshukla1:
@pgermosen Thank you. @hectorvent can we merge this ? Thanks

pgermosen (commented):
Coming back to this one — and starting with an apology, because my earlier review here was one line and I did not look closely enough. The behaviour you are fixing is right and the tests are good. There is one thing in `isAlreadyDeleted` I should have caught the first time, and I would rather raise it late than let it merge.

## The gap, in one sentence

`isAlreadyDeleted` is being asked to answer *"was this resource already gone?"* — but as written it also answers yes to *"did this delete fail for some reason whose error text happens to mention something missing?"*

Those are two very different situations, and right now they both end in `DELETE_COMPLETE`.

The two precise checks are exactly right:

```java
awsException.getHttpStatus() == 404
awsException.getErrorCode().endsWith("NotFoundException")
```

It is the third one that is doing more than intended:

```java
normalized.contains("not found") || normalized.contains("does not exist") || normalized.contains("not exist")
```

That runs against **every message in the cause chain**, and any hit marks the resource successfully deleted.

## A concrete way it goes wrong

Suppose deleting a resource fails because something it *depends on* is missing, and the provisioner reports:

```
Cannot delete listener rule: target group floci-tg-1 not found
```

The listener rule is still there and the delete genuinely failed — but the message contains `not found`, so the resource is marked `DELETE_COMPLETE` and the stack reports a clean deletion. That is the opposite of what your PR description promises ("retain `DELETE_FAILED` and `ROLLBACK_FAILED` behavior for genuine deletion errors"), and a silently-successful teardown is a worse failure than the one being fixed, because nothing surfaces it.

## The good news: it looks removable

I deleted that substring branch and re-ran both of your regressions:

| | result |
|---|---|
| `CloudFormationServiceRollbackTest` | 1/1 pass |
| `CloudFormationIntegrationTest#deleteStack_withAlreadyDeletedApiGatewayRestApi_reachesDeleteComplete` | 1/1 pass |

Both cases are already carried by the two precise checks, so as far as the current coverage goes the substring match is not needed.

**So the ask is small:** either drop those lines, or — if you added them because a specific provisioner reports a missing resource as a plain exception rather than an `AwsException` — keep them and add a test for that provisioner so we know what they are protecting. Either answer is fine; I just do not want an untested loose match sitting on the path that decides whether a stack deleted cleanly.

## Everything else is in good shape

- The 404 / `NotFoundException` checks are precise, and walking the cause chain with an identity set to avoid cycles is the right way to do it.
- `completeResourceDeletion` clearing `statusReason` alongside the status is a nice touch — it avoids a stale reason surviving on a resource that ended up fine.
- Applying the same treatment to both `deleteStackResources` and `rollbackCreatedResources` is right; those two drifting apart is how this class of bug usually comes back.
- You picked up my earlier nit and made `rollbackFailedExecution` package-private, so the test calls it directly instead of by reflection. Thank you — that is exactly what I was hoping for.

One housekeeping note: my approval is recorded against `26980af2`, which the rebase moved away, so it no longer points at the current head. The only real change since is `c132393a`, the test-fixture alignment for `StorageFactory.create`'s new return type — that one is fine and CI proves it compiles. I will re-approve on the current head once the `isAlreadyDeleted` question is settled.

Sorry again for surfacing this a fortnight after approving rather than during the first pass.

src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationServiceRollbackTest.java
aniketshukla1: Addressed during the rebase as well: rollbackFailedExecution remains package-private with its current main-branch signature, and CloudFormationServiceRollbackTest calls it directly with compile-time checking. No reflection remains.

aniketshukla1:
Addressed the latest review and rebased onto current main. Commit 4f0b779 removes the message-substring fallback from isAlreadyDeleted, retaining only HTTP 404 and *NotFoundException checks. A new regression models a listener-rule delete failing because its target group is not found and verifies the stack remains ROLLBACK_FAILED instead of being silently marked complete. Focused result: 3 tests passed across CloudFormationServiceRollbackTest and the out-of-band API Gateway deletion integration case.

pgermosen (approved):
Re-approving on current head (`4f0b7790`) — the `isAlreadyDeleted` gap I raised is fully resolved. Independently verified rather than just re-reading: reintroduced the substring-matching fallback and confirmed `createRollback_withDependencyNotFoundMessage_reachesRollbackFailed` fails with exactly the symptom I originally described (`expected: <ROLLBACK_FAILED> but was: <ROLLBACK_COMPLETE>`) — the test genuinely guards against regressing this. Restored, green.

Full `cloudformation` suite passes clean, including on a fresh merge against current `main`.

Thanks for chasing this all the way through, and sorry again for the fortnight-late second look.

PR #2041, opened 2026-07-29
fix(lambda): carry the owning account into published version snapshots

### LINKED ISSUE #2040: [BUG] Lambda: PublishVersion snapshot drops accountId
## Service

Lambda

## AWS API Action

PublishVersion

## Expected behavior

A published version snapshot is a complete `LambdaFunction`, so any consumer that reads a field off it gets the same answer it would get from `$LATEST`. In particular `getAccountId()` should return the function's owning account.

## Actual behavior

`publishVersion` (`LambdaService.java:926`) builds the snapshot field by field and copies 20 fields, but never `accountId`. Every published version therefore carries `accountId == null`, while `$LATEST` for the same function carries the real account (set in `createFunction` at `LambdaService.java:260`).

The account is recoverable from the snapshot's `functionArn`, which `publishVersion` does derive correctly (`LambdaService.java:933`), so this is a missing field copy rather than lost information.

Two places this reaches, one live and one latent:

- **Live, on a branch not yet merged.** In #1909 (Lambda execution-role credentials) the account is what a minted IAM session is registered and revoked under. Reading the field directly meant a version-qualified function got no session at all and silently skipped IAM enforcement while `$LATEST` enforced it. That PR now resolves the account from the function ARN when the field is blank, so it is handled defensively there and needs no change here. Flagging it because the ARN fallback is the *only* reason it works.
- **Latent.** `codeObjectKey` (`LambdaService.java:1257`) does `fn.getAccountId() != null ? fn.getAccountId() : "000000000000"`, so a version snapshot would silently key into the default account's prefix rather than the function's. Not reachable today: the one qualifier-accepting caller, `GetFunction`, goes through `getFunction`, which strips the qualifier (`LambdaService.java:377`) and returns the base function. It becomes reachable the moment any code path passes a version snapshot there.

No user-visible symptom on current `main`, which is the reason for filing rather than not: the field is invisibly absent, so each new consumer has to independently discover it and add its own fallback.

## Reproduction

No CLI-visible symptom today (`accountId` is not surfaced in `GetFunctionConfiguration`). Observable at the model level:

```java
LambdaFunction latest = lambdaService.createFunction(region, request);
LambdaFunction version = lambdaService.publishVersion(region, functionName, null);

latest.getAccountId();   // "000000000000" (or the configured account)
version.getAccountId();  // null
version.getFunctionArn();// "arn:aws:lambda:us-east-1:000000000000:function:fn:1" — account is right here
```

## Environment

- Floci version / image tag: `main` @ 1.5.34 (`b60a58fa`)
- Java SDK version (if applicable): n/a, internal model
- How you're running Floci (Docker / native / `mvn quarkus:dev`): `./mvnw test`

## Suggested fix

One line alongside the other snapshot copies in `publishVersion`:

```java
snapshot.setAccountId(fn.getAccountId());
```

Worth noting it does not fully replace a fallback at the consumption site: version snapshots persist, so snapshots written before the fix keep a null `accountId` after a restart. Both are useful, the field copy for correctness going forward and a fallback wherever the account is load-bearing.

Happy to send a PR for this if you'd like, though it touches the same `publishVersion` snapshot block as #2012, so it may be cleaner folded in there.

hampsterx:
## Summary

Closes #2040.

`publishVersion` builds its snapshot field by field and copies 20 fields, but never `accountId`. Every published version therefore carried a null owning account while `$LATEST` for the same function carried the real one.

The account was never lost, the version ARN already embeds it (`arn:aws:lambda:us-east-1:000000000000:function:fn:1`), it just was not on the model. This copies it alongside the other snapshot fields, so a version reports the same owning account as the function it was published from.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: a published version's owning account was absent from the model, so any consumer reading it had to re-derive the account from the function ARN or fall back to a default. `codeObjectKey` substitutes `000000000000` for a null account, which silently maps a version into the default account's prefix rather than the function's. That path is not reachable today (the one qualifier-accepting caller, `GetFunction`, resolves through `getFunction`, which strips the qualifier and returns the base function), so this is a latent mis-key rather than a live one. It becomes live as soon as any code path passes a version snapshot there.

Not observable through the CLI, since `accountId` is not surfaced in `GetFunctionConfiguration`. That is the reason for fixing it rather than leaving it: the field is invisibly absent, so each new consumer independently discovers the gap and adds its own fallback.

Worth noting a field copy does not fully remove the need for a fallback where the account is load-bearing. Version snapshots persist, so snapshots written before this change still deserialize with a null `accountId` after a restart.

## Checklist

- [x] `./mvnw test` passes locally (8065 tests, 0 failures, 9 skipped)

`LambdaServiceTest.publishVersionCarriesTheOwningAccount` publishes a version and asserts it reports the same account as the base function. Verified it fails without the change: `expected: <000000000000> but was: <null>`.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -928,6 +928,7 @@ public class LambdaService {
         functionName = fn.getFunctionName();
         int version = nextVersionNumber(region + "::" + functionName);
         LambdaFunction snapshot = new LambdaFunction();
+        snapshot.setAccountId(fn.getAccountId());
         snapshot.setFunctionName(fn.getFunctionName());
         snapshot.setVersion(String.valueOf(version));
         snapshot.setFunctionArn(fn.getFunctionArn().replace(":$LATEST", "") + ":" + version);
--- a/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaServiceTest.java
@@ -242,6 +242,21 @@ class LambdaServiceTest {
     }
 
     @Test
+    void publishVersionCarriesTheOwningAccount() {
+        service.createFunction(REGION, baseRequest("account-versioned-fn"));
+
+        LambdaFunction version = service.publishVersion(REGION, "account-versioned-fn", null);
+
+        // A version snapshot is built field by field, so anything reading accountId off it (the
+        // deployment-package key, execution-role session ownership) otherwise has to re-derive the
+        // account from the ARN or fall back to the default account.
+        assertEquals("000000000000", version.getAccountId());
+        assertEquals(
+                service.getFunction(REGION, "account-versioned-fn").getAccountId(),
+                version.getAccountId());
+    }
+
+    @Test
     void createFunctionWithSubdirectoryHandler() throws Exception {
         Map<String, Object> req = new java.util.HashMap<>(Map.of(
                 "FunctionName", "subdir-handler-fn",


src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
```diff
@@ -928,6 +928,7 @@ public LambdaFunction publishVersion(String region, String functionName, String
         functionName = fn.getFunctionName();
         int version = nextVersionNumber(region + "::" + functionName);
         LambdaFunction snapshot = new LambdaFunction();
+        snapshot.setAccountId(fn.getAccountId());
```
pgermosen, line 931: **The same copy drops several more fields. Unlike `accountId`, these are visible through the API and documented by AWS as version-scoped.**

I created a function with a real zip, `Architectures=[arm64]`, a layer ARN and `TracingConfig.Mode=Active`, then published a version:

| Field | Base function | Published version |
|---|---|---|
| `accountId` | `000000000000` | `000000000000` ✅ *(this PR)* |
| `codeSha256` | `Iog5ZxtZ+xbChySkj3Z70d5+i2H//UsL50rAhu6a/KQ=` | `null` ❌ |
| `architectures` | `[arm64]` | `null` ❌ |
| `layers` | `[arn:aws:lambda:us-east-1:000000000000:layer:mylayer:1]` | `[]` ❌ |
| `tracingMode` | `Active` | `PassThrough` ❌ |
| `codeLocalPath` | `/…/gap-probe-fn` | `null` ❌ |

`PublishVersion` (`LambdaController:377`) and `ListVersionsByFunction` (`:389`) both serialize the snapshot through `buildFunctionConfiguration`, which reads `CodeSha256` at `:575`, `Architectures` at `:600`, `Layers` at `:616`. So these reach real API responses.

**AWS documents all four as part of the version.** The [PublishVersion response](https://docs.aws.amazon.com/lambda/latest/api/API_PublishVersion.html) includes `CodeSha256` ("The SHA256 hash of the function's deployment package"), `Architectures`, `Layers` ("The function's layers") and `TracingConfig`. The [versioning guide](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html) is more direct still:

> After you publish a function version, its code, runtime, **architecture**, memory, **layers**, and most other configuration settings are immutable.

> When you publish a version, Lambda creates an immutable snapshot of your function's code and configuration.

That guide also enumerates the settings that qualify a function for version publication — which is effectively the list of what a snapshot must carry. Beyond the four above it names **VPC configuration**, **DLQ configuration**, **ephemeral storage size**, **package type** and **logging configuration**, so `VpcConfig`, `DeadLetterTargetArn`, `EphemeralStorageSize`, `KmsKeyArn` and the `ImageConfig*` fields are in the same boat.

`TracingMode` is the nastiest of them, because it doesn't go null: the fresh `LambdaFunction` default is `PassThrough`, so a version silently reports a *different but plausible* value than the function it was published from. Harder to spot than an obvious null.

`codeLocalPath` is Floci-internal and looks inert today for the same reason you gave for `accountId` (`getFunction` strips the qualifier), but it's the same latent shape.

**Worth noting what should stay uncopied**, so this doesn't over-correct: the same doc says *"Operational settings such as reserved concurrency don't trigger the publication of a new version"*, so `ReservedConcurrentExecutions` is correctly excluded, as are `Tags` (function-level, not version-level) and `ContainerState` (runtime state).

Two options, no strong preference:

- **Extend this PR** — a handful more `snapshot.setX(fn.getX())` lines, same shape as the one you added.
- **Follow-up issue** — fine too, but please reference #2040 so the diagnosis isn't re-derived.

What's worth avoiding is fixing them one at a time as each is independently rediscovered, which is exactly the failure mode your description names. Since the root cause is the hand-rolled copy silently dropping every field added to `LambdaFunction` after it was written, a `copyFrom` on the model — or at minimum a comment on this block — would make "carried" the default rather than "dropped".

pgermosen (approved):
The change is correct and the diagnosis is accurate — I confirmed `codeObjectKey` at `LambdaService:1259` is exactly `fn.getAccountId() != null ? ... : "000000000000"`, so the latent mis-key is real. ures.

I also appreciated the honesty about the persistence caveat — snapshots written before this change still deserialize with a null `accountId`, so a fallback is still needed where the account is load-bearing. Better said up front than left for a reader to assume.

**The field-by-field copy drops much more than `accountId`, and the rest is both CLI-visible and explicitly documented by AWS.** I probed a fully-populated function and then checked the results against the API reference — details and citations in the inline comment. `CodeSha256`, `Architectures`, `Layers` and `TracingConfig` are all lost, and `PublishVersion` / `ListVersionsByFunction` serialize the snapshot straight through `LambdaController.buildFunctionConfiguration`, so they land in real responses.

AWS is unambiguous that these belong to a version: *"When you publish a version, Lambda creates an immutable snapshot of your function's code and configuration"*, and *"After you publish a function version, its code, runtime, architecture, memory, layers, and most other configuration settings are immutable."*

The irony is that this PR fixes the *one* instance of the bug that is invisible through the CLI — which the description correctly identifies as why it went unnoticed — while several documented, visible instances of the identical bug remain in the same 20-line block.

I will **Approve** — the change is right and safe on its own. But I'd either extend it or open a follow-up referencing #2040, because the next person will otherwise re-derive the same diagnosis from scratch.

Sources: [PublishVersion API reference](https://docs.aws.amazon.com/lambda/latest/api/API_PublishVersion.html), [Manage Lambda function versions](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html)

hectorvent:
Thank you for this, the description is a model of what a small fix writeup should be, especially the honest note that older persisted snapshots still carry a null account.

I verified the scope: `accountId` is not surfaced on GetFunctionConfiguration per [botocore lambda/2015-03-31](https://github.com/boto/botocore/blob/develop/botocore/data/lambda/2015-03-31/service-2.json), so this is a pure internal fix, and the regression test fails without the change.

**(follow-up, separate PR)** pgermosen's review is right, I checked each claim against the source: the same copy block drops `codeSha256`, `architectures`, `layers` and `tracingMode`, all wire visible on PublishVersion. Since #2012 already covers the code path, a follow-up issue for the remaining fields (ideally a `copyFrom` so new fields default to carried) referencing #2040 would close this family for good.

No blockers from my side.

Issue #2042, opened 2026-07-30
[BUG] Logs Insights silently changes the result set for unsupported query syntax


TAKEDA-Takashi:
## Service

CloudWatch Logs (Logs Insights)

## AWS API Action

`StartQuery` / `GetQueryResults`

## Expected behavior

`filter @message like /ERROR/` is valid Logs Insights syntax — the
[filter documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_QuerySyntax-Filter.html)
lists `like`, `not like` and `=~` for substring and regex matching, and `=`, `!=`, `<`, `<=`, `>`,
`>=` as comparison operators. Against real AWS the query below returns the two `ERROR` events.

Floci implements a documented subset, which is fine. What I would expect is that a query it cannot
evaluate fails, rather than succeeding with a result set that does not correspond to the query.

## Actual behavior

Unsupported syntax changes the result set without any error, in two different directions.

**Case 1 — the stage is dropped, so every row comes back.** With three events in the log group
(one `INFO`, two `ERROR`), `filter @message like /ERROR/` returns:

```
"status": "Complete",
"statistics": { "recordsMatched": 3.0, "recordsScanned": 3.0, "bytesScanned": 0.0 }
```

All three rows, including the `INFO` one. `stats count()` behaves the same way — no aggregation
happens and the raw rows are returned. A warning is written to the server log, but the client sees
success.

**Case 2 — the operator is mis-parsed, so no rows come back and nothing is logged.**
`filter @message >= "x"` returns 0 rows out of 3. `parseFilter` scans for `!=`, `==` and `=`, so it
takes the `=` inside `>=` as the operator. The field name becomes `@message >`, which resolves to
`null` for every row, so nothing matches. This path emits no warning at all, because as far as
`parseFilter` is concerned it found a valid operator.

Case 2 is the one that worries me more: the query looks like it filtered, and there is nothing in
the log to suggest otherwise.

The relevant code is `LogsInsightsQuery`, which implements `fields`/`display`, `filter`/`where`,
`sort`/`order`, `dedup` and `limit`:

- `applyStage` ends with `default -> LOG.warnv("Ignoring unsupported Logs Insights command: {0}", cmd)`
- `parseFilter` returns `null` after `LOG.warnv("Ignoring unsupported Logs Insights filter: {0}", expr)`
  only when it finds no `=`, `!=` or `==` at all — `>=` and `<=` reach the mis-parse above instead

[...]

PR #2044, opened 2026-07-30
docs(cloudwatch): document Logs Insights subset and missing Logs actions

TAKEDA-Takashi:
## Summary

`docs/services/cloudwatch.md` documents 14 of the 24 actions that `CloudWatchLogsHandler`
dispatches. The handler is on the `deferred_handlers` allowlist in `tools/docs/services.yaml`
("multi-table doc"), so this table is maintained by hand rather than regenerated, and it has
drifted as actions were added.

The most visible gap is Logs Insights. `StartQuery` / `GetQueryResults` / `StopQuery` landed in
1.5.31 (#1448) but appear nowhere in the docs, so there is no documented way to find out which
parts of the query language work — and the implemented subset is deliberately narrow.

Docs only, touching `docs/services/cloudwatch.md` and
`docs/configuration/environment-variables.md`:

- **Adds the 10 missing actions** — `TagResource`, `UntagResource`, `ListTagsForResource`,
  `PutSubscriptionFilter`, `DescribeSubscriptionFilters`, `DeleteSubscriptionFilter`,
  `GetDataProtectionPolicy`, `StartQuery`, `GetQueryResults`, `StopQuery`
- **Adds a Logs Insights section** covering the supported commands (`fields`, `filter`, `sort`,
  `dedup`, `limit`, plus the `display` / `where` / `order` aliases), the field syntax including
  dotted paths into JSON messages and the always-included `@ptr` column, and a table of how each
  kind of unsupported input degrades — a dropped `filter` stage returns every row, a `>=` / `<=`
  comparison is mis-parsed into a field name and returns none, a missing projected field becomes an
  empty string, and an unrecognised `sort` direction falls back to ascending. Only the first two of
  those log a warning
- **Notes two simplifications** that are easy to trip over: `PutSubscriptionFilter`
  stores the filter for `DescribeSubscriptionFilters` but does not forward events to the
  destination ARN, and `GetDataProtectionPolicy` returns 200 with no `policyDocument`
- **Documents `FLOCI_SERVICES_CLOUDWATCHLOGS_QUERY_COMPLETION_DELAY_MS`** in both the service page
  and the central environment-variable list, and corrects the `MAX_EVENTS_PER_QUERY` description,
  which mentioned only `FilterLogEvents`. The `limit` row notes that the effective cap is the
  smallest of the query `limit`, the `StartQuery` `limit` parameter and this variable
- **Adds a `start-query` / `get-query-results` example**

Descriptions were taken from `CloudWatchLogsHandler`, `CloudWatchLogsService`,
`LogsInsightsQuery` and `EmulatorConfig` rather than from the AWS documentation, so they describe
what Floci does rather than what AWS does.

The silent-ignore behaviour itself is a separate question and is filed as #2042; this PR
only documents the current state.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No wire behaviour changed. The behaviour being documented was checked against
`floci/floci:1.5.34` with AWS CLI v2.32.1:

- The new `start-query` / `get-query-results` example returns the expected row, with `@ptr` appended
- With `FLOCI_SERVICES_CLOUDWATCHLOGS_QUERY_COMPLETION_DELAY_MS=4000`, a query reports `Running`
  immediately after `StartQuery` and `Complete` once the delay elapses. `StopQuery` on a separate,
  still-running query returned `success=true` (a stopped query stays `Cancelled`, so these were two
  different queries)
- On a log group with one `INFO` and two `ERROR` events: `filter @message like /ERROR/` returns all
  3 rows, `filter @message >= "x"` returns 0, and `stats count()` returns the 3 raw rows. The server
  log shows `Ignoring unsupported Logs Insights ...` for the first and third but not for the `>=`
  case

## Checklist

- [ ] `./mvnw test` passes locally — not run; this PR touches no code
- [ ] New or updated integration test added — n/a for docs

### DIFF
--- a/docs/configuration/environment-variables.md
+++ b/docs/configuration/environment-variables.md
@@ -252,7 +252,8 @@ See [Initialization Hooks](./initialization-hooks.md) for lifecycle phases and s
 | Variable | Default | Description |
 |---|---|---|
 | `FLOCI_SERVICES_CLOUDWATCHLOGS_ENABLED` | `true` | Enable the CloudWatch Logs service |
-| `FLOCI_SERVICES_CLOUDWATCHLOGS_MAX_EVENTS_PER_QUERY` | `10000` | Maximum log events returned by a single `FilterLogEvents` call |
+| `FLOCI_SERVICES_CLOUDWATCHLOGS_MAX_EVENTS_PER_QUERY` | `10000` | Maximum log events returned by a single `FilterLogEvents` or `GetLogEvents` call, and the upper bound for a Logs Insights `limit` |
+| `FLOCI_SERVICES_CLOUDWATCHLOGS_QUERY_COMPLETION_DELAY_MS` | `0` | Artificial Logs Insights query delay. With `0` a query completes immediately; a positive value emulates the asynchronous `Running` → `Complete` lifecycle |
 
 ### CloudWatch Metrics
 
--- a/docs/services/cloudwatch.md
+++ b/docs/services/cloudwatch.md
@@ -27,13 +27,68 @@ Floci supports both CloudWatch Logs and CloudWatch Metrics.
 | `TagLogGroup` | Tag a log group |
 | `UntagLogGroup` | Remove tags |
 | `ListTagsLogGroup` | List tags |
+| `TagResource` | Tag a log group by ARN |
+| `UntagResource` | Remove tags from a log group by ARN |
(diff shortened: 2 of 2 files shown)


hectorvent (approved):
Thanks @TAKEDA-Takashi

PR #2046, opened 2026-07-30
feat(kms): implement ListKeyPolicies

### LINKED ISSUE #1528: [FEAT] KMS: implement ListKeyPolicies

hectorvent:
## Summary

Continues @Marcel2603's work in #1626, with the two outstanding review points applied. Original
commit is carried as a co-author.

`ListKeyPolicies` failed with `UnsupportedOperation`, so a client enumerating policy names before
calling `GetKeyPolicy` could not proceed. `GetKeyPolicy` and `PutKeyPolicy` were already supported,
so this was a hole in an otherwise complete area.

Closes #1528

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/kms/2014-11-01/service-2.json)
(kms/2014-11-01).

- Returns `PolicyNames: ["default"]`, which the model states outright: *"The only valid value is
  `default`"*. Matches moto's reference response.
- `Truncated: false`, and **`NextMarker` is omitted** rather than serialized as `null`. The model
  describes `NextMarker` as present only when `Truncated` is true, and AWS omits absent optional
  members, so emitting an explicit null would be a shape deviation.
- `KeyId` is required per the model, so it is read through the existing `requiredText` helper.

### Two changes from #1626

**`InvalidMarkerException` is not declared for this operation.** `ListKeyPolicies` declares exactly
five errors: `NotFoundException`, `InvalidArnException`, `DependencyTimeoutException`,
`KMSInternalException`, `KMSInvalidStateException`. `InvalidMarkerException` is a real KMS shape, but
it is declared only on the genuinely paginated calls (`ListKeys`, `ListAliases`, `ListGrants`,
`ListKeyRotations`, `ListResourceTags`, `ListRetirableGrants`, `DescribeCustomKeyStores`).
`ListKeyPolicies` is deliberately excluded, which makes sense: with one fixed policy name the result
can never truncate, so no marker is ever valid to reject. `Marker` is therefore accepted and ignored
rather than rejected with an undeclared code.

**`Limit` is no longer rejected either.** The original raised `ValidationException` for values
outside `[1..1000]`, but `ValidationException` is not even a shape in the KMS model, and a single
policy name fits inside any limit. This also matches the existing house pattern: `listGrants` in the
same class clamps with `Math.clamp` rather than throwing.

Net effect: this implementation invents no error code the model does not declare.

## Notes

The not-found integration test asserts the error **type** only, not the status code. Floci returns
404 for KMS `NotFoundException` where real KMS uses 400. That is a pre-existing service-wide
deviation and out of scope here, so a new test should not pin it. Worth its own issue.

## Checklist

### Validation

`./mvnw test -Dtest='KmsServiceTest,KmsIntegrationTest'` → **247 tests, 0 failures**.

New coverage: default policy name, ARN-based key resolution, not-found, `NextMarker` absence, and a
loop asserting `Limit` of `null/1/1000/0/1001` plus an arbitrary `Marker` are all accepted without
error.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
@@ -494,13 +494,14 @@ public class KmsService {
      * {@code default}, so the result can never be truncated: {@code Truncated} is always false and
      * {@code NextMarker} is omitted.
      *
-     * <p>{@code Limit} and {@code Marker} are accepted and ignored, which is deliberate. A single
-     * policy name fits inside any valid limit, and the model does not declare
-     * {@code InvalidMarkerException} for this operation, unlike the genuinely paginated
-     * {@code ListKeys}, {@code ListAliases}, {@code ListGrants} and {@code ListResourceTags}. There is
-     * therefore no marker this call could legitimately reject.
+     * <p>{@code Limit} and {@code Marker} are accepted at the wire and ignored: the handler simply
+     * never reads them, which is exactly what moto's KMS does for this operation. A single policy
+     * name fits inside any limit, so the values cannot change the response. Real AWS rejects an
+     * out-of-range {@code Limit} server-side (the model bounds LimitType to [1..1000], but SDKs do
+     * not fully enforce that client-side — botocore checks only the minimum); not re-validating it
+     * here is deliberate leniency, consistent with the emulator's general posture.
      */
-    public Map<String, Object> listKeyPolicies(String keyId, Integer limit, String marker, String region) {
+    public Map<String, Object> listKeyPolicies(String keyId, String region) {
         String policyName = (String) getKeyPolicy(keyId, region).get("PolicyName");
 
         Map<String, Object> result = new HashMap<>();
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
@@ -489,6 +489,26 @@ public class KmsService {
         LOG.infov("Updated key policy for KMS key: {0} in {1}", key.getKeyId(), region);
     }
 
+    /**
+     * Lists the policy names attached to a key. KMS supports exactly one key policy, named
+     * {@code default}, so the result can never be truncated: {@code Truncated} is always false and
+     * {@code NextMarker} is omitted.
+     *
+     * <p>{@code Limit} and {@code Marker} are accepted and ignored, which is deliberate. A single
+     * policy name fits inside any valid limit, and the model does not declare
+     * {@code InvalidMarkerException} for this operation, unlike the genuinely paginated
+     * {@code ListKeys}, {@code ListAliases}, {@code ListGrants} and {@code ListResourceTags}. There is
+     * therefore no marker this call could legitimately reject.
+     */
+    public Map<String, Object> listKeyPolicies(String keyId, Integer limit, String marker, String region) {
+        String policyName = (String) getKeyPolicy(keyId, region).get("PolicyName");
+
+        Map<String, Object> result = new HashMap<>();
+        result.put("PolicyNames", List.of(policyName));
+        result.put("Truncated", false);
+        return result;
+    }
+
     public void updateKeyDescription(String keyId, String description, String region) {
         KmsKey key = resolveKey(keyId, region);
         key.setDescription(description);
--- a/src/test/java/io/github/hectorvent/floci/services/kms/KmsIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/kms/KmsIntegrationTest.java
@@ -13,6 +13,8 @@ import java.util.List;
 
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasKey;
+import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.startsWith;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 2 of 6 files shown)


src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
```diff
@@ -489,6 +489,26 @@ public void putKeyPolicy(String keyId, String policy, String region) {
...
     }
 
+    /**
+     * Lists the policy names attached to a key. KMS supports exactly one key policy, named
+     * {@code default}, so the result can never be truncated: {@code Truncated} is always false and
+     * {@code NextMarker} is omitted.
+     *
+     * <p>{@code Limit} and {@code Marker} are accepted and ignored, which is deliberate. A single
+     * policy name fits inside any valid limit, and the model does not declare
+     * {@code InvalidMarkerException} for this operation, unlike the genuinely paginated
```
pgermosen: **The stated reason for skipping `Limit` validation is contradicted by this same diff**

The description argues `Limit` isn't validated because *"`ValidationException` is not even a shape in the KMS model."* That's true of the model — but this PR's own handler reads `KeyId` through `requiredText`, and `requiredText` (`KmsJsonHandler:498`) throws exactly:

```java
throw new AwsException("ValidationException", field + " is required", 400);
```

So the diff both invokes the principle and breaks it. `listGrants:306` throws the same code too. Both can't follow from "the model doesn't declare it."

The resolution is that the principle is the wrong one — real AWS does return `ValidationException` for constraint violations it never declares, because those come from the protocol layer rather than the operation model. `requiredText` is fine and matches AWS.

**There's a stronger argument for the actual decision, and it's in the model:** `LimitType` is `{"type": "integer", "min": 1, "max": 1000}`. Constrained members are validated *client-side* by botocore and every AWS SDK, so a caller physically cannot reach Floci with `Limit=0` or `Limit=1001` through any SDK — only through hand-rolled HTTP. Accepting whatever arrives is therefore unobservable in practice, which justifies the choice far better than the model's error list does.

**Suggested:** swap the reasoning in this javadoc and the CHANGELOG for the `min`/`max` one. The code doesn't change; the justification stops contradicting the neighbouring line.

src/test/java/io/github/hectorvent/floci/services/kms/KmsServiceTest.java
```diff
@@ -1054,6 +1054,52 @@ void putKeyPolicyOnNonExistentKeyThrows() {
...
+    }
+
+    @Test
+    void listKeyPoliciesIgnoresLimitAndMarker() {
+        // A single policy name cannot be paginated, and unlike ListKeys / ListAliases / ListGrants /
+        // ListResourceTags this operation does not declare InvalidMarkerException, so there is no
+        // marker it could legitimately reject.
+        KmsKey key = kmsService.createKey("list-policy-paging", REGION);
+
+        for (Integer limit : new Integer[] {null, 1, 1000, 0, 1001}) {
```
pgermosen: **This test cannot fail, and the two bot comments are right about why**

`limit` and `marker` are never read in `listKeyPolicies`. That makes this loop structurally vacuous — the assertion holds for `null/1/1000/0/1001` because the parameters are unreachable, not because the implementation handles them. It would keep passing against an empty method body.

This is the same thing `github-code-quality[bot]` flagged twice on `KmsService.java:503` ("The parameter 'limit' is never used", same for 'marker'). I'd normally push back on that bot, but here it's pointing at something real — and this test is what makes the dead parameters look load-bearing.

The behavior worth testing is that the **wire** accepts them, and `KmsIntegrationTest.listKeyPoliciesIgnoresLimitAndMarker` already does exactly that by posting `{"Limit":1,"Marker":"anything"}` through the handler. That one can genuinely fail — if `handleListKeyPolicies` threw on an unexpected member, it would catch it.

**Suggested:** drop `limit` and `marker` from the service signature and delete this unit test, keeping the integration test. The handler is where "accepted and ignored" actually happens — it accepts them by simply not reading them — and the javadoc can say so. That resolves both bot comments without needing a suppression.

src/test/java/io/github/hectorvent/floci/services/kms/KmsServiceTest.java
```diff
@@ -1054,6 +1054,52 @@ void putKeyPolicyOnNonExistentKeyThrows() {
...
+        List<String> policyNames = (List<String>) result.get("PolicyNames");
+        assertEquals(1, policyNames.size());
+        assertEquals("default", policyNames.getFirst());
+        assertFalse((Boolean) result.get("Truncated"));
+        // Truncated is always false, so NextMarker must not be emitted at all.
+        assertFalse(result.containsKey("NextMarker"));
+    }
+
+    @Test
+    void listKeyPoliciesResolvesByArnAndAlias() {
```
pgermosen: **The name promises alias resolution that AWS doesn't offer and this test doesn't exercise**

The body only calls `listKeyPolicies(key.getArn(), ...)`. There's no alias in it, so the name overstates the coverage.

More to the point, the model gives this operation no alias contract to test. Its `KeyId` documentation lists **key ID and key ARN only**:

> Specify the key ID or key ARN of the KMS key.

Compare `Encrypt`, which spells the alternative out — *"use its key ID, key ARN, alias name, or alias ARN."* `GetKeyPolicy` and `PutKeyPolicy` are key-ID/ARN-only too.

Floci's `resolveKey` (`KmsService:1069-1087`) accepts alias ARNs and `alias/` names for everything uniformly, so the extra permissiveness is real but pre-existing and shared — not this PR's to fix. The issue is only that the name records it as *intended* behavior for an operation where AWS doesn't have it.

**Suggested:** rename to `listKeyPoliciesResolvesByArn`. If alias acceptance is worth pinning deliberately, it deserves its own test with a comment that it's a known Floci-wide superset of AWS.

src/test/java/io/github/hectorvent/floci/services/kms/KmsIntegrationTest.java
```diff
@@ -1106,4 +1108,79 @@ void createKeyWithCombinations(String keySpec, String keyUsage, int expectedStat
...
+                .then()
+                .statusCode(200)
+                .body("PolicyNames[0]", equalTo("default"))
+                .body("Truncated", equalTo(false));
+    }
+
+    @Test
+    void listKeyPoliciesReturnsNotFoundForUnknownKey() {
+        // Asserting the error type only. Floci returns 404 where real KMS uses 400 for
+        // NotFoundException, a pre-existing service-wide deviation that is not this change's to fix.
```
pgermosen, line 1176: **The 404/400 note is right, but the deviation isn't service-wide — and that makes the follow-up easier**

Correct for this path, and not asserting the status code here is the right call. But the deviation isn't uniform: this same file already pins **both** behaviors.

| test | status |
|---|---|
| `listGrantsReturnsNotFoundForUnknownKey` (`:542`) | 404 |
| `createGrant`, unknown key (`:623`) | 404 |
| `revokeGrantReturnsNotFoundForUnknownKey` (`:815`) | 404 |
| **`revokeGrantReturnsNotFoundForUnknownGrant`** (`:803`) | **400** |
| **`retireGrantReturnsNotFoundForInvalidToken`** (`:955`) | **400** |

The split is by *what* is missing: `resolveKey` throws 404 for a missing key, while the grant lookups throw 400 — which is the AWS-correct one. The model backs that up: `NotFoundException` carries no `error.httpStatusCode`, so under JSON 1.1 it defaults to 400.

**Suggested:** worth capturing in the follow-up issue, because it changes the shape of the fix. It isn't "flip one service-wide constant" — it's "align `resolveKey`'s 404 with the 400 the grant paths already return", and there's a correct precedent in the same class to point at. Leaving this test status-agnostic is right either way.

pgermosen (approved):
Looks good

This is a small, well-argued change, and the model reasoning behind it is unusually careful. Every AWS claim in the description is true — I parsed `botocore/data/kms/2014-11-01/service-2.json` to check each one rather than reading prose docs:

| claim | verified |
|---|---|
| `PolicyNames` doc says *"The only valid value is `default`"* | exact string in the model |
| `NextMarker` present only when `Truncated` is true | *"When `Truncated` is true, this element is present"* |
| `ListKeyPolicies` declares exactly 5 errors, no `InvalidMarkerException` | `NotFoundException`, `InvalidArnException`, `DependencyTimeoutException`, `KMSInternalException`, `KMSInvalidStateException` |
| `InvalidMarkerException` declared only on the paginated calls | exactly the 7 you named — set equality, nothing missing or extra |
| `ValidationException` is not a shape in the KMS model | confirmed absent |
| `KeyId` is required | `required: ["KeyId"]` |

**The `InvalidMarkerException` observation is sharper than the description claims:** `ListKeyPolicies` is the *only* operation in the entire KMS API that accepts a `Marker` and doesn't declare it. That's a deliberate signal rather than an omission, and reading it that way is the right call.

 No correctness issues. Comments 1 and 2 are about the stated rationale and a test that can't fail; 3 and 4 are small.

src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
hectorvent: changes pushed on 6e9ea18

src/test/java/io/github/hectorvent/floci/services/kms/KmsServiceTest.java
hectorvent: changes pushed on 6e9ea18

PR #2049, opened 2026-07-30
fix(apigateway): support AWS type Lambda integration and fix duplicate Content-Type header

### LINKED ISSUE #2048: [BUG] apigateway --type AWS putIntegration endpoint invocation fails
## Service

API Gateway

## AWS API Action

PutIntegration / execute-api (invoke)

## Expected behavior

When an API Gateway REST API method is configured with `--type AWS` and a Lambda integration URI (`arn:aws:apigateway:{region}:lambda:path/2015-03-31/functions/{fnArn}/invocations`), invoking the endpoint should execute the Lambda function with the VTL-rendered request template as the payload and return the response after applying any configured VTL response template mappings.

This is the standard non-proxy Lambda integration pattern - it provides full request/response VTL mapping, unlike `AWS_PROXY` which bypasses it.

## Actual behaviour

Invoking the endpoint returns a 500:

`{"message": "The request must contain the parameter Action"}`

Floci incorrectly treats the Lambda path-style URI (`lambda:path/...`) as a query-protocol (form-urlencoded) integration. It attempts to parse the VTL-rendered Lambda payload as AWS query protocol and dispatch it through invokeQuery, which fails because the body contains no `Action` parameter.

`AWS_PROXY` integrations against the same Lambda function work correctly.

## Reproduction

```
#!/usr/bin/env bash
set -euo pipefail

export AWS_PAGER=""
export AWS_ACCESS_KEY_ID=test
export AWS_DEFAULT_REGION=eu-west-2

# =============================================================================
# Lambda function code - update the handler body below to change the response
# =============================================================================
LAMBDA_CODE=$(cat <<'LAMBDA'
  return {
    statusCode: 200,
    body: JSON.stringify({ message: "Hello from my-function!", event }),
  };
};
LAMBDA
)

# Package and create the Lambda function
LAMBDA_DIR=$(mktemp -d)
echo "$LAMBDA_CODE" > "$LAMBDA_DIR/index.js"
(cd "$LAMBDA_DIR" && zip -q function.zip index.js)

aws lambda create-function \
  --function-name my-function \
  --runtime nodejs20.x \
  --handler index.handler \
  --role arn:aws:iam::000000000000:role/lambda-role \
  --zip-file "fileb://$LAMBDA_DIR/function.zip" \
  --endpoint-url "$AWS_ENDPOINT_URL"

rm -rf "$LAMBDA_DIR"
echo "Lambda function 'my-function' created"

# =============================================================================
# API Gateway setup
# =============================================================================

# Create a REST API
API_ID=$(aws apigateway create-rest-api \
  --name "My API" \
  --query id --output text \
  --endpoint-url "$AWS_ENDPOINT_URL")
echo "API_ID: $API_ID"

# Get the root resource
ROOT_ID=$(aws apigateway get-resources \
  --rest-api-id "$API_ID" \
  --query 'items[?path==`/`].id' --output text \
  --endpoint-url "$AWS_ENDPOINT_URL")
echo "ROOT_ID: $ROOT_ID"

# Create a resource
RESOURCE_ID=$(aws apigateway create-resource \
  --rest-api-id "$API_ID" \
  --parent-id "$ROOT_ID" \
  --path-part users \
  --query id --output text \
  --endpoint-url "$AWS_ENDPOINT_URL")
echo "RESOURCE_ID: $RESOURCE_ID"

# Add a GET method
aws apigateway put-method \
  --rest-api-id "$API_ID" \
  --resource-id "$RESOURCE_ID" \
  --http-method GET \
  --authorization-type NONE \
  --endpoint-url "$AWS_ENDPOINT_URL"

# Add a Lambda integration (non-proxy AWS type)
aws apigateway put-integration \
  --rest-api-id "$API_ID" \
  --resource-id "$RESOURCE_ID" \
  --http-method GET \
  --type AWS \
  --integration-http-method POST \
  --uri "arn:aws:apigateway:eu-west-2:lambda:path/2015-03-31/functions/arn:aws:lambda:eu-west-2:000000000000:function:my-function/invocations" \
  --endpoint-url "$AWS_ENDPOINT_URL"

# Add a method response (required for non-proxy integrations)
aws apigateway put-method-response \
  --rest-api-id "$API_ID" \
  --resource-id "$RESOURCE_ID" \
  --http-method GET \
  --status-code 200 \
  --endpoint-url "$AWS_ENDPOINT_URL"

# Add an integration response (required for non-proxy integrations)
aws apigateway put-integration-response \
  --rest-api-id "$API_ID" \
  --resource-id "$RESOURCE_ID" \
  --http-method GET \
  --status-code 200 \
  --selection-pattern "" \
  --endpoint-url "$AWS_ENDPOINT_URL"

# Create a deployment
DEPLOYMENT_ID=$(aws apigateway create-deployment \
  --rest-api-id "$API_ID" \
  --query id --output text \
  --endpoint-url "$AWS_ENDPOINT_URL")
echo "DEPLOYMENT_ID: $DEPLOYMENT_ID"

# Create the stage explicitly
aws apigateway create-stage \
  --rest-api-id "$API_ID" \
  --stage-name dev \
  --deployment-id "$DEPLOYMENT_ID" \
  --endpoint-url "$AWS_ENDPOINT_URL"

echo ""
echo "=== Curl Commands ==="
echo "GET /users:"
echo "  curl http://localhost:4566/restapis/$API_ID/dev/_user_request_/users"
echo ""
echo "=== IDs Summary ==="
echo "  API_ID:      $API_ID"
echo "  ROOT_ID:     $ROOT_ID"
echo "  RESOURCE_ID: $RESOURCE_ID"
echo "  Stage:       dev"
echo ""

# Call the deployed API
# Invoke — returns 500 with "The request must contain the parameter Action" (Floci bug)
curl http://localhost:4566/restapis/$API_ID/dev/_user_request_/users
```

## Environment

- Floci version / image tag: latest
- Java SDK version (if applicable): AWS CLI v2
- How you're running Floci (Docker / native / mvn quarkus:dev): Docker

teddmason:
## Summary

Fixes two bugs in `ApiGatewayExecuteController` for `AWS` type integrations.

Closes #2048 

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

When a REST API method integration is configured with `--type AWS` pointing to a Lambda function, hitting the endpoint returns a `500` with `"The request must contain the parameter Action"` instead of invoking the Lambda. `AWS_PROXY` integrations against the same function work correctly. The `AWS` type is supposed to invoke the Lambda with the VTL-rendered request template as the payload, but Floci was instead treating it as a query-protocol (form-urlencoded) service call and failing when no `Action` parameter was found in the body.

## Checklist

### DIFF
--- a/docs/services/api-gateway.md
+++ b/docs/services/api-gateway.md
@@ -77,7 +77,7 @@ These management-plane operations have no handler in v1. Calls will return `404`
 - Client Certificates (5 operations)
 - `GetExport` / `ImportDocumentationParts`
 
-The execute plane (actual proxied HTTP traffic via `/restapis/{id}/{stage}/_user_request_/…`) is implemented separately and is not counted as management-plane operations.
+The execute plane (actual proxied HTTP traffic via `/restapis/{id}/{stage}/_user_request_/…`) is implemented separately and is not counted as management-plane operations. It supports `AWS_PROXY` (Lambda proxy), `AWS` (Lambda with VTL request/response templates), and `MOCK` integrations; other integration types return an error.
 
 ### Examples
 
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -912,15 +912,38 @@ public class ApiGatewayExecuteController {
 
         // Dispatch to service.
         //
-        // A path-style integration URI (arn:...:{service}:path/...) carries no action: the
-        // rendered template body is the AWS query protocol (form-urlencoded,
+        // Lambda path-style integrations (arn:aws:apigateway:{region}:lambda:path/...) are
+        // handled specially: the function name is extracted from the URI and the rendered
+        // request template body is passed directly as the Lambda payload — just like
+        // AWS_PROXY, but with request/response VTL mapping applied.
+        //
+        // For other services: a path-style integration URI (arn:...:{service}:path/...) carries
+        // no action: the rendered template body is the AWS query protocol (form-urlencoded,
         // "Action=SendMessage&..."). Action-style URIs (arn:...:{service}:action/{Action})
         // carry the action in the URI and render a JSON body.
         Response serviceResponse;
         String errorType = null;
         String errorMessage = null;
         try {
-            if (target.action() == null) {
+            if ("lambda".equals(target.service())) {
+                String functionName = functionNameFromUri(integration.getUri());
+                if (functionName == null || functionName.isBlank()) {
+                    throw new AwsException("InvalidParameterValueException",
+                            "Cannot resolve Lambda function name from URI: " + integration.getUri(), 400);
+                }
+                byte[] payload = transformedBody != null ? transformedBody.getBytes(StandardCharsets.UTF_8) : new byte[0];
+                InvokeResult invokeResult = lambdaService.invoke(region, functionName, payload, InvocationType.RequestResponse);
+                String lambdaResponseBody = invokeResult.getPayload() != null
+                        ? new String(invokeResult.getPayload(), StandardCharsets.UTF_8) : "{}";
+                int lambdaStatus = invokeResult.getStatusCode() > 0 ? invokeResult.getStatusCode() : 200;
+                if (invokeResult.getFunctionError() != null) {
+                    errorType = invokeResult.getFunctionError();
+                    errorMessage = lambdaResponseBody;
+                }
+                serviceResponse = Response.status(lambdaStatus)
+                        .entity(lambdaResponseBody)
+                        .type(MediaType.APPLICATION_JSON).build();
+            } else if (target.action() == null) {
                 MultivaluedMap<String, String> formParams = parseFormUrlEncoded(transformedBody);
                 serviceResponse = serviceRouter.invokeQuery(target.service(), formParams, region);
             } else {
@@ -1045,13 +1068,18 @@ public class ApiGatewayExecuteController {
         }
 
         Response.ResponseBuilder rb = Response.status(finalStatus)
-                .entity(finalBody)
-                .type(MediaType.APPLICATION_JSON);
+                .entity(finalBody);
+
+        String contentType = null;
 
         // Apply $context.responseOverride header assignments.
         if (templateResult != null && !templateResult.headerOverrides().isEmpty()) {
             for (Map.Entry<String, String> hdr : templateResult.headerOverrides().entrySet()) {
-                rb.header(hdr.getKey(), hdr.getValue());
+                if ("Content-Type".equalsIgnoreCase(hdr.getKey())) {
+                    contentType = hdr.getValue();
+                } else {
+                    rb.header(hdr.getKey(), hdr.getValue());
+                }
             }
         }
 
@@ -1070,11 +1098,16 @@ public class ApiGatewayExecuteController {
                 String headerName = dest.substring("method.response.header.".length());
                 String headerValue = resolveResponseParameter(source, serviceResponseHeaders, responseBodyStr);
                 if (headerValue != null) {
-                    rb.header(headerName, headerValue);
+                    if ("Content-Type".equalsIgnoreCase(headerName)) {
+                        contentType = headerValue;
+                    } else {
+                        rb.header(headerName, headerValue);
+                    }
                 }
             }
         }
 
+        rb.type(contentType != null ? contentType : MediaType.APPLICATION_JSON);
         return rb.build();
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaArnUtils.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaArnUtils.java
@@ -163,16 +163,23 @@ public final class LambdaArnUtils {
     }
 
     /**
-     * Extracts the Lambda function name from an API Gateway integration URI.
+     * Extracts the Lambda function reference from an API Gateway integration URI.
      * Handles formats like:
      * <ul>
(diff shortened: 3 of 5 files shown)


pgermosen (commented):
@teddmason 

No blocking findings. The headline fix (routing `AWS`-type Lambda integrations through a direct invoke instead of the query-protocol path) is correctly scoped and reuses the existing `functionNameFromUri` helper rather than inventing a parallel extraction path.

Worth calling out: the `LambdaArnUtils.extractFunctionNameFromUri` change fixes a second, more severe pre-existing bug beyond what the PR description mentions. For a qualified ARN like `arn:...:function:myFn:PROD/invocations`, the old `lastIndexOf(':')` logic returned just `"PROD"` as the "function name" — the alias/version qualifier entirely replaced the actual name, not just got dropped. Since `functionNameFromUri` is shared by 5 other call sites (AWS_PROXY integrations, REQUEST/TOKEN authorizers), this fix's blast radius is wider than the PR title suggests, and it's correctly covered by both direct extraction tests and a round-trip test through `resolve()` confirming `name`/`qualifier` split correctly.

The Content-Type dedup is correct: both header-override sources are captured into a local variable and applied once via `.type()` at the end instead of `.header()`, so a `$context.responseOverride` or `method.response.header.Content-Type` mapping no longer produces two Content-Type headers alongside the default `.type(APPLICATION_JSON)`.

Two non-blocking notes:
- In `extractFunctionNameFromUri`, the `functionPart.startsWith("function:")` branch now handles every real Lambda ARN shape, leaving the old `lastIndexOf(':')` fallback effectively unreachable except for malformed/non-standard input — harmless, just worth knowing it's dead in practice.
- `docs/services/api-gateway.md`'s v1 (REST API) section has no execute-plane integration-type behavior notes at all (unlike v2's WebSocket data-plane, which has a full supported-features table listing `AWS integration (Lambda with VTL templates)` explicitly). This PR is a good opportunity to add an equivalent one-line note for v1, since this exact gap (AWS-type Lambda integration returning 500) wasn't documented as either working or broken beforehand.

abanna (approved):
Thank you for improving API Gateway’s AWS Lambda integration behavior. I reviewed the current head and ran the API Gateway/Lambda and ARN utility suites—all 59 tests passed, with full CI green. The change is focused, uses existing Floci patterns, and includes regression coverage. I found no blocking issue. Approved.

pgermosen (approved):
Re-checked at `0386efcf`, merged onto current `main`, built, and probed against a `main` baseline running the same Lambda and API definitions on a second port.

Both notes from my last pass are addressed: the `lastIndexOf(':')` fallback now carries a comment saying why it is unreachable for real ARNs, and `docs/services/api-gateway.md` has the v1 execute-plane line. I checked the new docs sentence rather than taking it on trust — the dispatch switch handles exactly `AWS_PROXY`, `AWS` and `MOCK`, and everything else falls through:

```
GET …/_user_request_/httpint   (integration type HTTP_PROXY)
→ 500 {"message":"Unsupported integration type: HTTP_PROXY"}
```

So the sentence is accurate as written.

## Both fixes verified against a paired baseline

Same function, same API definition, `main` on one port and this branch on another.

**The titled fix.** `--type AWS` pointing at a Lambda, with a VTL request template:

```
main    → {"errorMessage":"The request must contain the parameter Action","errorType":"MissingAction"}
branch  → {"forwarded":{"hello":"world"}}
```

The template rendered and reached the function. That is #2048 closed.

**The Content-Type dedup**, with `method.response.header.Content-Type` mapped to `'text/plain'`:

```
main    → Content-Type: application/json
          Content-Type: text/plain          ← two headers
branch  → Content-Type: text/plain          ← one
```

## The ARN fix is larger than the title says, and this time I could isolate it

Last round I could only argue this from reading. `AWS_PROXY` — untouched by this PR's dispatch change — pointed at a qualified ARN (`…:function:echoFn:PROD/invocations`), against a real published version and alias:

```
main    → 404 {"message":"Function not found: PROD"}
branch  → 200 {"p":1}
```

The alias name replaced the function name outright. And `extractFunctionNameFromUri` is reached from **seven** call sites, not just the one this PR is about:

| path | call site |
|---|---|
| REST v1 `AWS_PROXY` invoke | `ApiGatewayExecuteController:401` |
| REST v1 REQUEST/TOKEN authorizer | `:440` |
| REST v1 `AWS` integration (new here) | `:1063` |
| HTTP API v2 integration | `:1514` |
| HTTP API v2 authorizer | `:1849` |
| WebSocket authorizer | `WebSocketAuthorizerService:123` |
| WebSocket integration invoker | `WebSocketIntegrationInvoker:493` |

Every one of them is broken on `main` today for any Lambda referenced by alias or version — which is ordinary deployment practice, not an edge case. That is arguably the more valuable half of this PR, and it appears in neither the title, the description, nor #2048. Worth putting in the merge commit so it is findable later.

## Tests

**684 tests, 0 failures** on the merge with current `main`, including the new `ApiGatewayAwsLambdaIntegrationTest` (10) and `LambdaArnUtilsTest` (49, up from 24). The qualified-ARN cases are covered both directly and through a `resolve()` round trip asserting the `name`/`qualifier` split, which is the part that actually makes the invoke target the alias.

The checklist still has `./mvnw test` passes locally unticked — for the record, it does pass. Merge with `main` is clean.

## Two small things

- **`ApiGatewayExecuteController`'s class javadoc is now stale.** Line 52 still reads *"routing them through the configured integration (AWS_PROXY or MOCK)"*. `AWS` belongs in that list, and it sits a few lines above the switch this PR extends.
- **The description says the failure is a `500`.** What you get depends on whether an integration response is configured: with one mapped on `selectionPattern ""` it surfaces as **200 with the `MissingAction` body** (what I saw), and only without a mapping does it become the 500. Not wrong, just worth a word so nobody reproducing it thinks they have a different bug.

Neither blocks anything.

PR #2052, opened 2026-07-30
fix(codebuild): stabilize start and retry responses

abanna:
## Summary

`StartBuild` and `RetryBuild` returned the same mutable `Build` object that the asynchronous runner
updates. A fast local build could therefore reach a terminal state before the HTTP response was
serialized, making an accepted request appear to have failed synchronously.

This change captures a detached response snapshot before starting the runner while keeping the
stored build live for later `BatchGetBuilds` calls.

## Type of change

- [x] Bug fix (non-breaking change)

## AWS compatibility

AWS returns a `Build` object from both `StartBuild` and `RetryBuild`. Its lifecycle fields describe
the build state represented by that response, while subsequent reads can observe later asynchronous
state.

Floci now returns a stable accepted response with `buildStatus: IN_PROGRESS`,
`buildComplete: false`, and `currentPhase: SUBMITTED`. The stored build then advances through the
normal runner lifecycle independently. This avoids imposing a later phase on the response when the
AWS API contract does not require one deterministic immediate phase.

Sources:

- https://docs.aws.amazon.com/codebuild/latest/APIReference/API_StartBuild.html
- https://docs.aws.amazon.com/codebuild/latest/APIReference/API_RetryBuild.html
- https://docs.aws.amazon.com/codebuild/latest/APIReference/API_Build.html

## Verification

- `mise exec java@25 -- ./mvnw test -Dtest=CodeBuildServicePersistenceTest`
- Regression coverage forces the runner to mutate the stored build immediately and verifies stable
  `StartBuild` and `RetryBuild` responses.
- `git diff --check`

## Checklist

- [x] The change follows the project style guidelines.
- [x] The change has been self-reviewed.
- [x] A focused regression test proves the fix.
- [x] No new endpoints or custom wire formats were introduced.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/codebuild/CodeBuildService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/codebuild/CodeBuildService.java
@@ -448,10 +448,11 @@ public class CodeBuildService {
         build.setPhases(new CopyOnWriteArrayList<>());
 
         buildsFor(region).put(buildId, build);
+        Build responseBuild = copyBuild(build);
 
         runner.startBuild(region, build, project, buildspecOverride);
 
-        return build;
+        return responseBuild;
     }
 
     public Build getBuild(String region, String buildId) {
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thanks @abanna!

PR #2060, opened 2026-07-30
fix(s3): accept a percent-encoded bucket/key separator in copy sources

### LINKED ISSUE #2038: [BUG] 1.5.34 possibly breaks S3 CopyObjectAsync
## Service

S3

## AWS API Action

CopyObjectAsync

## Expected behavior

Tests that include CopyObjectAsync succeed

## Actual behavior

Tests fail after upgrading from 1.5.33 to 1.5.34

## Reproduction

I haven't extracted code that reproduces the problem, I just noticed that some of my tests started failing after upgrading the image from 1.5.33-compat to 1.5.34-compat. The tests that fail verify "move object" functionality, i.e. they first create an object in S3 folder, and then move it by calling CopyObjectAsync and DeleteObjectAsync S3 .NET API. These tests failed until I rolled back to 1.5.33.

## Environment

- Floci version / image tag: 1.5.34-compat

hampsterx:
## Summary

- Restores `CopyObject` and `UploadPartCopy` for the AWS SDK for .NET. That SDK percent-encodes the whole `x-amz-copy-source` value, so a v4 client sends `bucket%2Ffolder%2Fkey.txt` with no literal slash anywhere. Since 1.5.34 the bucket/key split runs on the raw header via `indexOf('/')`, finds no separator, and every copy from that SDK fails with 400 `InvalidArgument`.
- Wider than a nested-key problem: copies of top-level keys fail too, so on 1.5.34 no .NET `CopyObjectAsync` call succeeds at all.
- Treats whichever comes first, a literal `/` or an encoded `%2F`, as the bucket/key separator. Bucket names admit neither `/` nor `%`, so the first occurrence of either is unambiguously the separator and never part of the bucket name.
- Keeps the split before decoding, so #1924's behaviour is intact: an encoded slash inside a key stays key content, a literal `?` (raw or `%3F`) is still preserved, and `?versionId=` still parses as a query.
- Also fixes the mixed form `bucket%2Ffolder/key.txt`, which previously decoded the bucket as `bucket/folder`.
- Both handlers now share one `parseCopySource` helper instead of duplicating the split.

Closes #2038

## Reproduction

Ran a real `AWSSDK.S3` 4.0.101 client against the published images. Every copy below is a plain `CopyObjectAsync`:

| Key | 1.5.33-compat | 1.5.34-compat | this branch |
|-----|---------------|---------------|-------------|
| `folder/file.txt` | pass | 400 `InvalidArgument` | pass |
| `top.txt` | pass | 400 `InvalidArgument` | pass |
| `a/b/c/deep.txt` | pass | 400 `InvalidArgument` | pass |
| `folder/with space.txt` | pass | 400 `InvalidArgument` | pass |

The error message shows the wire format: `Invalid copy source: dotnet-movetest%2Ffolder%2Ffile.txt`.

The AWS CLI is unaffected because it sends literal slashes, which is why the compatibility suite did not catch this. There is no .NET SDK in the compat matrix.

## Changes

```
 .../hectorvent/floci/services/s3/S3Controller.java | 91 +++++++++++++++++-----
 .../floci/services/s3/S3IntegrationTest.java       | 84 ++++++++++++++++++++
 .../services/s3/S3MultipartIntegrationTest.java    | 12 +++
 3 files changed, 169 insertions(+), 18 deletions(-)
```

## Test plan

- [x] Full suite green: 8104 tests, 0 failures, 0 errors, 9 skipped
- [x] New `copyObjectWithPercentEncodedBucketSeparatorSucceeds` covers the fully-encoded .NET wire format, lowercase `%2f`, a leading encoded separator, and the mixed encoded-separator/literal-slash form, asserting copied content each time
- [x] `uploadPartCopy` extended with an encoded-separator part, since both handlers share the helper. The part is completed and its bytes asserted, so a parse resolving the wrong key fails on content rather than passing on status
- [x] `copyObjectWithMalformedEncodedSourceReturns400` extended with `bucket%2F%ZZinvalid`, confirming decoding still rejects malformed input reached through the new path
- [x] `copyObjectWithEmptyBucketReturns400` extended with the separator-less and `%2F`-leading forms, so a source naming no bucket is still a 400
- [x] Existing `copyObjectWithQuestionMarkInSourceKeySucceeds` (#1924 regression test) still passes unchanged
- [x] Real AWSSDK.S3 v4 client passes end-to-end against a patched build

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -1923,15 +1923,8 @@ public class S3Controller {
 
     private Response handleCopyObject(String copySource, String destBucket, String destKey,
                                       String contentType, HttpHeaders httpHeaders) {
-        // copySource format: /bucket/key or bucket/key, where key is URL-encoded
-        String source = copySource.startsWith("/") ? copySource.substring(1) : copySource;
-
-        int slashIndex = source.indexOf('/');
-        if (slashIndex <= 0) {
-            throw new AwsException("InvalidArgument", "Invalid copy source: " + copySource, 400);
-        }
-        String sourceBucket = decodeCopySourceComponent(source.substring(0, slashIndex), copySource);
-        ParsedCopySource sourceObject = parseCopySourceObject(source.substring(slashIndex + 1), copySource);
+        CopySourceRef sourceObject = parseCopySource(copySource);
+        String sourceBucket = sourceObject.bucket();
         String copyContentEncoding = toPersistedContentEncoding(httpHeaders.getHeaderString("Content-Encoding"));
         String copyContentDisposition = httpHeaders.getHeaderString("Content-Disposition");
         String copyCacheControl = httpHeaders.getHeaderString("Cache-Control");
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Thanks @hampsterx for handling this.

Issue #2043, opened 2026-07-30
[BUG] Cloud Control ListResources returns an empty list for unsupported resource types


TAKEDA-Takashi:
## Service

Cloud Control API

## AWS API Action

`ListResources`

## Expected behavior

A caller should be able to tell "this type is not supported here" from "there are no resources of
this type".

Note that the types below are not obscure ones: real Cloud Control API lists them. The
[user guide](https://docs.aws.amazon.com/cloudcontrolapi/latest/userguide/resource-types.html) uses
`AWS::Logs::LogGroup` as its example of a schema that supports all five operations including
`list`, and `AWS::SQS::Queue` is in the
[supported resource types table](https://docs.aws.amazon.com/cloudcontrolapi/latest/userguide/supported-resources.html)
as well. So against real AWS these calls return the resources; against Floci they return an empty
list that looks like a successful "nothing found".

## Actual behavior

An unsupported type returns `ResourceDescriptions: []` with HTTP 200 — the same response as a
supported type with no resources. Resources that demonstrably exist are reported as empty.

`CloudControlService.listResources` handles six types and falls through for everything else:

```java
case "AWS::S3::Bucket" -> s3Buckets();
case "AWS::EC2::VPC" -> vpcs(region);
case "AWS::EC2::Subnet" -> subnets(region);
case "AWS::EC2::SecurityGroup" -> securityGroups(region);
case "AWS::IAM::Role" -> roles();
case "AWS::IAM::User" -> users();
default -> List.of();
```

A type name that does not exist in AWS at all is indistinguishable from a real but unsupported
one — both return an empty list.

## Reproduction

```bash
docker run -d --rm -p 4566:4566 floci/floci:1.5.34

export AWS_ENDPOINT_URL=http://localhost:4566
export AWS_DEFAULT_REGION=us-east-1
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test

aws sqs create-queue --queue-name demo
aws logs create-log-group --log-group-name /demo
aws s3api create-bucket --bucket demo-bucket

# both of these exist, but Cloud Control reports nothing
aws cloudcontrol list-resources --type-name AWS::SQS::Queue
# {"ResourceDescriptions": [], "TypeName": "AWS::SQS::Queue"}
aws cloudcontrol list-resources --type-name AWS::Logs::LogGroup
# {"ResourceDescriptions": [], "TypeName": "AWS::Logs::LogGroup"}

[...]

PR #2062, opened 2026-07-30
feat(cloudcontrol): read EC2 instances, launch templates and instance profiles

lex00:
Refs #2043 (partial).

`ListResources` covered six types; `AWS::EC2::Instance`, `AWS::EC2::LaunchTemplate` and `AWS::IAM::InstanceProfile` were not among them, so a caller that provisions one through Cloud Control could not read it back — and a tool whose AWS provider reaches these types only through its Cloud Control proxy could not observe its own estate at all.

Cloud Control reports a resource's *current* model rather than an echo of the desired state, so an instance carries what only the running resource knows: private and public addresses, state, availability zone, and the subnet and security groups it actually landed in. A property with no value is omitted rather than reported as null, matching Cloud Control.

`GetResource` reads through the same list, so it resolves these types once the list does.

This addresses the missing-types half of #2043. It does **not** address the other half — making an unsupported type distinguishable from a supported type with no resources — which needs a decision about what Cloud Control should return for a type it has no schema for.

Test: `listResourcesReportsAnInstancesRuntimeModel` runs an instance and asserts Cloud Control lists it with its model. 436 tests across the Cloud Control, EC2 and IAM suites pass.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudcontrol/CloudControlService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudcontrol/CloudControlService.java
@@ -5,6 +5,11 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.services.ec2.Ec2Service;
+import io.github.hectorvent.floci.services.ec2.model.GroupIdentifier;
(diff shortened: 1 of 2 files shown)


lex00:
Out of scope for this PR — it adds `ListResources` coverage for the three types only (`Refs #2043 (partial)`). `GetResource` support is a separate change.

lex00:
Folded into #2037 and closing this one — same `CloudControlService` surface, and it keeps my open-PR count moving in the right direction. Nothing is lost; the commit is on that branch unchanged apart from a rebase.

PR #2059, opened 2026-07-30
feat(iam): seed AWSCloudFormationReadOnlyAccess

lex00:
Fixes #2058.

`cdk bootstrap` attaches this AWS-managed policy to the roles in its CDKToolkit stack. The catalog did not carry it, so bootstrap failed with `Policy arn:aws:iam::aws:policy/AWSCloudFormationReadOnlyAccess does not exist.` and rolled back — blocking every CDK deployment against the emulator, since bootstrap is a prerequisite.

`AdministratorAccess` and `ReadOnlyAccess` were already seeded; this is the remaining one bootstrap needs.

Test: `seedAwsManagedPolicies` now asserts the policy resolves with path `/`. 59 tests in `IamServiceTest` pass.

PR #2057, opened 2026-07-30
feat: Add AWSCloudFormationReadOnlyAccess as a managed policy.

floverfelt:
## Summary

Adds AWSCloudFormationReadOnlyAccess

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore

## Checklist

- [ ] `./mvnw test` passes locally
- [x] New or updated integration test added
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

PR #2064, opened 2026-07-30
feat(iam): seed the managed policies CDK bootstrap and the scenario stacks attach

### LINKED ISSUE #2063: [BUG] Missing AWS-managed policies roll back CDK bootstrap and the aws-bench scenario stacks
Four AWS-managed policies are absent from the seeded catalog. Each fails the same way — `Policy arn:aws:iam::aws:policy/<name> does not exist`, reported on the role that consumed it, rolling the stack back several resources from the actual cause.

| Policy | Attached by |
|---|---|
| `AWSCloudFormationReadOnlyAccess` | the roles `cdk bootstrap` creates |
| `AmazonAthenaFullAccess` | aws-bench ec2-multiregion QA roles |
| `AmazonRedshiftFullAccess` | aws-bench ec2-multiregion QA roles |
| `AmazonS3TablesReadOnlyAccess` | aws-bench ec2-multiregion QA roles |

The first is the most severe: bootstrap is a prerequisite, so no CDK application can be deployed against the emulator at all.

## Reproduce

```sh
export AWS_ENDPOINT_URL=http://localhost:4566 AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test
npx cdk bootstrap aws://000000000000/us-east-1
```

```
CREATE_FAILED  DeploymentActionRole  Policy arn:aws:iam::aws:policy/AWSCloudFormationReadOnlyAccess does not exist.
CDKToolkit -> ROLLBACK_COMPLETE
```

## Expected

The referenced managed policies resolve and the stacks reach CREATE_COMPLETE.

lex00:
Fixes #2063. Supersedes #2059.

Four AWS-managed policies were absent from the catalog, each failing the same way — `Policy arn:aws:iam::aws:policy/<name> does not exist`, reported on the role that consumed it, rolling the stack back several resources from the cause.

| Policy | Attached by |
|---|---|
| `AWSCloudFormationReadOnlyAccess` | the roles `cdk bootstrap` creates |
| `AmazonAthenaFullAccess` | aws-bench ec2-multiregion QA roles |
| `AmazonRedshiftFullAccess` | aws-bench ec2-multiregion QA roles |
| `AmazonS3TablesReadOnlyAccess` | aws-bench ec2-multiregion QA roles |

The first is the severe one: bootstrap is a prerequisite, so no CDK application could be deployed against the emulator at all. Verified — after this change `cdk bootstrap` reaches `✅ Environment bootstrapped` in all three regions.

Test: `seedAwsManagedPolicies` asserts all four resolve with path `/`. 59 tests in `IamServiceTest` pass.


lex00:
#2057 seeded `AWSCloudFormationReadOnlyAccess` while this was open, so I've dropped it here — the duplicate definition was what turned the catalog count red (`POLICIES.size()` counted 57, the listing deduped to 56). The other three policies are unchanged.

One thing worth a second opinion: #2057's description is "Provides read only access to Cloud Control and CloudFormation." and the one I had was "Provides access to AWS CloudFormation via the AWS Management Console." I didn't touch it since it isn't this PR's line, but if the AWS-published text is the latter it'd be worth a one-liner somewhere.

lex00:
Checked all six against real AWS with `iam:GetPolicy` rather than leaving it as a question. Two are wrong on `main`:

| Policy | Seeded | AWS |
|---|---|---|
| `AWSCloudFormationReadOnlyAccess` | Provides read only access to Cloud Control and CloudFormation. | Provides access to AWS CloudFormation via the AWS Management Console. |
| `AmazonS3TablesReadOnlyAccess` | Provides read-only access to Amazon S3 Tables. | Provides read only access to all S3 table buckets. |

The first is #2057's line, the second was mine. Both corrected here.

`AmazonAthenaFullAccess` and `AmazonRedshiftFullAccess` already match verbatim, as do the two Bedrock ones in #2034.

Folded into this PR rather than opening another, given the PR count — happy to split it out if you'd rather review it separately.

hectorvent (approved):
Thanks @lex00

PR #2068, opened 2026-07-31
fix(pipes): register kafka-clients reflection metadata for native image

LiamMacP:
## Summary

#1260 added Kafka and MSK sources to EventBridge Pipes, but only verified them on the JVM. In the native image both source types silently deliver nothing.

Floci depends on `kafka-clients` directly rather than through the Quarkus Kafka extension, and `kafka-clients` ships no native-image metadata of its own. `AbstractConfig` resolves every `CLASS`-typed consumer setting by name, so the deserializers [we set explicitly](https://github.com/floci-io/floci/blob/0142dc434f98ba14a13236196c72036839a1e84b/src/main/java/io/github/hectorvent/floci/services/pipes/PipesKafkaConsumerManager.java#L141) — plus the `ConsumerConfig` defaults for `partition.assignment.strategy` and `metric.reporters` — are never registered for reflection and the consumer cannot be constructed:

```
poll error: Invalid value org.apache.kafka.common.serialization.ByteArrayDeserializer
for configuration key.deserializer: Class ... could not be found. (ConfigException)
```

This is easy to miss because [PipesPoller catches and warns per poll cycle](https://github.com/floci-io/floci/blob/0142dc434f98ba14a13236196c72036839a1e84b/src/main/java/io/github/hectorvent/floci/services/pipes/PipesPoller.java#L152-L157), so the pipe stays reported as `RUNNING` with no error surfaced to the caller.

The registration lives next to the only code that uses `kafka-clients`, following the existing `DockerJavaNativeSupport` holder pattern. Every entry is a class `ConsumerConfig` resolves, and `PipesKafkaNativeSupportTest` reads `PipesKafkaConsumerManager.consumerProperties` directly — not a copy of it — asserting equality in both directions, so a new setting on the consumer is covered automatically.

The consumer also sets `enable.metrics.push=false`. It defaults to true, and Floci has no reason to push client telemetry to a user's broker; it additionally keeps the shaded protobuf serializer off any reflective path.

`sasl.*` and `ssl.*` defaults are excluded deliberately. Floci connects over PLAINTEXT so they are never instantiated, and registering `DefaultJwtValidator` aborts the native build outright — its constructor references jose4j, an optional `kafka-clients` dependency we do not ship:

```
Fatal error: java.lang.NoClassDefFoundError: org/jose4j/keys/resolvers/VerificationKeyResolver
```

### Binary size

The native binary grows **196 MB → 210.7 MB** (linux/arm64, measured against `floci/floci:1.5.34`). The cost is not the metadata volume: registering one class costs the same as registering all four, because the closed-world analysis cannot see through `Utils.newInstance(Class.forName(...))` — on `main` the consumer's entire constructed graph is dead code, and the first reflective instantiation makes it live. Over half of what arrives is Kafka's auto-generated wire-protocol layer. Measured on macOS/arm64 with the same toolchain and flags throughout:

| registration | binary |
|---|---|
| none (current `main`) | 187.7 MB |
| `ByteArrayDeserializer` only | 201.5 MB |
| 3 classes | 201.9 MB |
| 4 classes (this PR) | 202.1 MB |

So there is nothing to trim — a minimal list costs the same as the complete one. The smaller binary on `main` only contains a Kafka consumer that cannot run.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

No wire-protocol change. The incorrect behaviour was that a Kafka or MSK pipe created through the Pipes API reported `RUNNING` but delivered no records to its target in native mode, while behaving correctly on the JVM.

Verified end to end with AWS CLI 1.45.61 against a Kafka broker (`apache/kafka:4.1.0`) and an MSK cluster, producing to the topic and asserting delivery to the SQS target:

| | JVM | native (before) | native (after) |
|---|---|---|---|
| `smk://` → Pipes → SQS | 5/5 | **0/5** | 5/5 |
| MSK → Pipes → SQS | 5/5 | **0/5** | 5/5 |

Both source types failed identically before the fix, since they share `createConsumer`. The "after" column was run against the shipped image built from `docker/Dockerfile.native`, with Floci itself running as a container so MSK took the sibling-container path:

```
PipesKafkaConsumerManager] Pipe fpk-selfmanaged-pipe: subscribed Kafka consumer to topic ... via fpk-kafka:9092
PipesKafkaConsumerManager] Pipe fpk-msk-pipe:         subscribed Kafka consumer to topic ... via 10.89.0.7:9092
```

Zero `poll error` lines after the fix, against 20+ per pipe before it.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupport.java
+++ b/src/main/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupport.java
@@ -5,8 +5,6 @@ import org.apache.kafka.clients.consumer.CooperativeStickyAssignor;
 import org.apache.kafka.clients.consumer.RangeAssignor;
 import org.apache.kafka.common.metrics.JmxReporter;
 import org.apache.kafka.common.serialization.ByteArrayDeserializer;
-import org.apache.kafka.common.utils.AppInfoParser;
-import org.apache.kafka.shaded.com.google.protobuf.ExtensionRegistry;
 
 /**
  * Registers the kafka-clients classes that GraalVM cannot discover statically.
@@ -17,10 +15,13 @@ import org.apache.kafka.shaded.com.google.protobuf.ExtensionRegistry;
  * consumer that cannot be constructed. {@link PipesPoller} logs that failure and moves on, leaving
  * a Kafka or MSK pipe reported as RUNNING while it delivers nothing.
  *
- * <p>{@code sasl.*} and {@code ssl.*} defaults are excluded deliberately. Floci connects over
- * PLAINTEXT so they are never instantiated, and registering {@code DefaultJwtValidator} aborts the
- * native build outright: its constructor references jose4j, an optional kafka-clients dependency
- * Floci does not ship. {@code PipesKafkaNativeSupportTest} guards both directions.
+ * <p>Every entry is config-resolved, and {@code PipesKafkaNativeSupportTest} derives its
+ * expectations the same way, so list and guard cannot drift. The tracing agent also reported
+ * {@code AppInfoParser.AppInfo}, {@code AppInfoParser.AppInfoMBean} and the shaded
+ * {@code ExtensionRegistry}; native pipes deliver without them, so they are omitted.
+ *
+ * <p>SASL and SSL defaults are excluded: Floci is PLAINTEXT-only, and {@code DefaultJwtValidator}
+ * fails the native build outright by referencing jose4j, an optional dependency Floci ships without.
  */
 @RegisterForReflection(targets = {
     // key.deserializer / value.deserializer — set explicitly by PipesKafkaConsumerManager
@@ -29,12 +30,7 @@ import org.apache.kafka.shaded.com.google.protobuf.ExtensionRegistry;
     RangeAssignor.class,
     CooperativeStickyAssignor.class,
     // metric.reporters — ConsumerConfig default
-    JmxReporter.class,
-    // MBean registered by AppInfoParser on consumer startup
-    AppInfoParser.AppInfo.class,
-    AppInfoParser.AppInfoMBean.class,
-    // shaded protobuf reached from the client telemetry reporter
-    ExtensionRegistry.class
+    JmxReporter.class
 })
 public class PipesKafkaNativeSupport {
 }
--- a/src/main/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupport.java
+++ b/src/main/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupport.java
@@ -9,19 +9,16 @@ import org.apache.kafka.common.serialization.ByteArrayDeserializer;
 /**
  * Registers the kafka-clients classes that GraalVM cannot discover statically.
  *
- * <p>Floci uses kafka-clients directly rather than through the Quarkus Kafka extension, and
- * kafka-clients ships no native-image metadata of its own. {@code AbstractConfig} resolves every
- * {@code CLASS}-typed consumer setting by name, so without these entries a native build produces a
- * consumer that cannot be constructed. {@link PipesPoller} logs that failure and moves on, leaving
- * a Kafka or MSK pipe reported as RUNNING while it delivers nothing.
+ * <p>kafka-clients ships no native-image metadata and Floci uses it directly rather than through
+ * the Quarkus Kafka extension. {@code AbstractConfig} resolves {@code CLASS}-typed settings by
+ * name, so without these a native build cannot construct a consumer — which {@link PipesPoller}
+ * swallows per poll cycle, leaving the pipe RUNNING and delivering nothing.
  *
- * <p>Every entry is config-resolved, and {@code PipesKafkaNativeSupportTest} derives its
- * expectations the same way, so list and guard cannot drift. The tracing agent also reported
- * {@code AppInfoParser.AppInfo}, {@code AppInfoParser.AppInfoMBean} and the shaded
- * {@code ExtensionRegistry}; native pipes deliver without them, so they are omitted.
+ * <p>Entries are exactly what {@link PipesKafkaConsumerManager#consumerProperties} resolves, and
+ * {@code PipesKafkaNativeSupportTest} reads that same method, so the two cannot drift.
  *
  * <p>SASL and SSL defaults are excluded: Floci is PLAINTEXT-only, and {@code DefaultJwtValidator}
- * fails the native build outright by referencing jose4j, an optional dependency Floci ships without.
+ * fails the native build by referencing jose4j, an optional dependency Floci omits.
  */
 @RegisterForReflection(targets = {
     // key.deserializer / value.deserializer — set explicitly by PipesKafkaConsumerManager
--- a/src/main/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupport.java
+++ b/src/main/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupport.java
@@ -1,0 +1,40 @@
+package io.github.hectorvent.floci.services.pipes;
+
+import io.quarkus.runtime.annotations.RegisterForReflection;
+import org.apache.kafka.clients.consumer.CooperativeStickyAssignor;
+import org.apache.kafka.clients.consumer.RangeAssignor;
+import org.apache.kafka.common.metrics.JmxReporter;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.apache.kafka.common.utils.AppInfoParser;
+import org.apache.kafka.shaded.com.google.protobuf.ExtensionRegistry;
+
+/**
+ * Registers the kafka-clients classes that GraalVM cannot discover statically.
+ *
+ * <p>Floci uses kafka-clients directly rather than through the Quarkus Kafka extension, and
+ * kafka-clients ships no native-image metadata of its own. {@code AbstractConfig} resolves every
+ * {@code CLASS}-typed consumer setting by name, so without these entries a native build produces a
+ * consumer that cannot be constructed. {@link PipesPoller} logs that failure and moves on, leaving
+ * a Kafka or MSK pipe reported as RUNNING while it delivers nothing.
+ *
+ * <p>{@code sasl.*} and {@code ssl.*} defaults are excluded deliberately. Floci connects over
+ * PLAINTEXT so they are never instantiated, and registering {@code DefaultJwtValidator} aborts the
+ * native build outright: its constructor references jose4j, an optional kafka-clients dependency
+ * Floci does not ship. {@code PipesKafkaNativeSupportTest} guards both directions.
+ */
+@RegisterForReflection(targets = {
+    // key.deserializer / value.deserializer — set explicitly by PipesKafkaConsumerManager
+    ByteArrayDeserializer.class,
+    // partition.assignment.strategy — ConsumerConfig defaults
+    RangeAssignor.class,
+    CooperativeStickyAssignor.class,
+    // metric.reporters — ConsumerConfig default
+    JmxReporter.class,
+    // MBean registered by AppInfoParser on consumer startup
+    AppInfoParser.AppInfo.class,
+    AppInfoParser.AppInfoMBean.class,
+    // shaded protobuf reached from the client telemetry reporter
+    ExtensionRegistry.class
+})
+public class PipesKafkaNativeSupport {
+}
--- a/src/test/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupportTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupportTest.java
@@ -1,0 +1,96 @@
+package io.github.hectorvent.floci.services.pipes;
+
+import io.quarkus.runtime.annotations.RegisterForReflection;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Predicate;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Guards {@link PipesKafkaNativeSupport} against kafka-clients upgrades.
+ *
+ * <p>A class missing from the registration only fails in native mode, and only as a swallowed
+ * poller warning, so this asserts the registration against the config kafka-clients actually
+ * resolves rather than against a hand-maintained list.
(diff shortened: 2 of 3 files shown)


src/main/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupport.java
```diff
@@ -0,0 +1,36 @@
...
+ * <p>Floci uses kafka-clients directly rather than through the Quarkus Kafka extension, and
+ * kafka-clients ships no native-image metadata of its own. {@code AbstractConfig} resolves every
+ * {@code CLASS}-typed consumer setting by name, so without these entries a native build produces a
+ * consumer that cannot be constructed. {@link PipesPoller} logs that failure and moves on, leaving
+ * a Kafka or MSK pipe reported as RUNNING while it delivers nothing.
+ *
+ * <p>Every entry is config-resolved, and {@code PipesKafkaNativeSupportTest} derives its
+ * expectations the same way, so list and guard cannot drift. The tracing agent also reported
+ * {@code AppInfoParser.AppInfo}, {@code AppInfoParser.AppInfoMBean} and the shaded
+ * {@code ExtensionRegistry}; native pipes deliver without them, so they are omitted.
```
pgermosen: **The `ExtensionRegistry` removal is untested rather than wrong — the javadoc shouldn't call it verified**

The javadoc says native pipes "deliver without them." That's true, but neither broker you tested *could* have shown otherwise — so it isn't evidence for the removal.

`enable.metrics.push` defaults to `true`; I printed it from the resolved config. Per [KIP-714](https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability) a client only sends telemetry RPCs to a broker that advertises support, and *"the default configuration for an Apache Kafka cluster does not include any metrics subscriptions."* Stock `apache/kafka:4.1.0` and Floci's MSK container both ship without a `ClientTelemetry` plugin, so the telemetry path never executed in either run. Since the payload is *"OpenTelemetry MetricsData v1 protobuf"*, that path is exactly where the shaded `ExtensionRegistry` the tracing agent recorded would be used.

**What this is not:** dropping it does not remove protobuf from the image — I checked. A `PrintUniverse` diff of the native builds shows `org.apache.kafka.shaded.*` (67 types) and `common.telemetry` (17) reachable either way, and the `UnsafeUtil` warning fires on `main` too. So the open question is narrow: whether one reflective lookup on an unexercised path is registered. I can't settle that without a telemetry-capable broker.

**Suggested — either is fine:**

- Turn the path off: `properties.put(ConsumerConfig.ENABLE_METRICS_PUSH_CONFIG, "false")` in `createConsumer`. Floci has no reason to push client telemetry to a user's broker, it's a boolean so it can't affect the config-resolved set, and it removes the question rather than answering it.
- Or keep the code and reword the javadoc — "unexercised" rather than "not needed" — so the next reader doesn't take the 5/5 as coverage of this path.

src/test/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupportTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static Optional<String> loadable(String className) {
+        try {
+            return Optional.of(Class.forName(className, false, ConsumerConfig.class.getClassLoader()).getName());
+        } catch (ClassNotFoundException | LinkageError e) {
+            return Optional.empty();
+        }
+    }
+
+    private static Map<String, ?> consumerConfigValues() {
```
pgermosen: **The javadoc's "list and guard cannot drift" isn't true — I reproduced the drift**

The guard protects against kafka-clients upgrades. It does not protect against changes to Floci's own code, because `consumerConfigValues()` **re-declares** the eight properties from `PipesKafkaConsumerManager:139-147` instead of reading them. It's pinned to a copy of `createConsumer`, not to `createConsumer`.

Repro — one line added to `createConsumer`:

```java
properties.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG,
        "org.apache.kafka.clients.consumer.RoundRobinAssignor");
```

`PipesKafkaNativeSupportTest` still passes, 2/2. But `interceptor.classes` is config-resolved, so in a native build that class wouldn't be registered, the consumer would fail to construct, and it would fail *silently* through the same `PipesPoller:154` swallow this PR exists to close. Adding SASL when MSK grows IAM auth would do the same thing — and that's the likeliest future edit to this method.

**Suggested:** extract `Properties consumerProperties(Pipe pipe)` as package-private in `PipesKafkaConsumerManager`, have `createConsumer` call it, and have this test call it too. That follows AGENTS.md's *"Prefer package-private constructors for testability"*, and `PipesKafkaConsumerManagerTest` already builds `Pipe` fixtures from JSON text blocks and calls package-private methods, so there's a pattern to copy and no new scaffolding. The javadoc's claim then becomes true as written.

src/test/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupportTest.java
```diff
@@ -0,0 +1,114 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Pins {@link PipesKafkaNativeSupport} to what {@code ConsumerConfig} resolves, in both directions:
+ * a newly resolved class must be registered, and nothing else may be.
+ *
+ * <p>Scope: config-resolved classes only. Reflection always succeeds on the JVM, so a non-config
+ * lookup added by a future kafka-clients is only catchable by running a Kafka pipe against a
+ * native binary — no suite does that today.
```
pgermosen: **The native harness already exists and runs on every PR — only the Kafka case is missing**

This comment says no suite runs a Kafka pipe against a native binary. The *native binary* half is already covered: `.github/workflows/compatibility.yml` builds the native image (`:37`), loads it (`:94`), and runs the SDK suites against `floci:test-native`. Pipes are covered in five of them — `sdk-test-awscli/test/pipes.bats`, `sdk-test-go/tests/pipes_test.go`, `sdk-test-node/tests/pipes.test.ts`, `sdk-test-python/tests/test_pipes.py`, `sdk-test-java/.../PipesTest.java`.

What's missing is a Kafka *source*. None of those five mentions Kafka, MSK or `smk://` — zero grep hits — and `MskTest.java` only exercises the control plane (create / describe / list / getBootstrapBrokers / delete), never a running consumer. No compat suite starts a broker.

So this is one missing test case, not a missing harness. AGENTS.md points at that suite for exactly this situation: *"Use this suite when changes may affect real SDK behavior."* It's also why #1260 shipped a natively-broken feature, and why this one needed someone with a broker to notice.

**Suggested:** deferring the test is reasonable — it needs a broker container on `compat-net`, which is real work. But please reword this sentence from "no suite does that" to "no case does that yet," so it reads as a known gap rather than an accepted limitation.

src/test/java/io/github/hectorvent/floci/services/pipes/PipesKafkaNativeSupportTest.java
```diff
@@ -0,0 +1,114 @@
...
+        missing.removeAll(registeredClassNames());
+
+        assertTrue(missing.isEmpty(),
+                "kafka-clients resolves these reflectively but PipesKafkaNativeSupport omits them, "
+                        + "so a native build cannot construct a consumer: " + missing);
+    }
+
+    @Test
+    @DisplayName("nothing is registered that ConsumerConfig does not resolve")
+    void registrationContainsNothingUnjustified() {
```
pgermosen: **This assertion will block a future legitimate addition** *(nit)*

Dropping the three tracing-agent entries rather than pinning them was the right call — the guard couldn't cover them, so keeping them would have been decoration.

The side effect: this test now rejects anything that isn't config-resolved. If a future native run shows `AppInfo` really is needed, the fix has to include an edit to this test, and the diff will read as someone weakening a guard to make their change pass.

Your failure message already explains this better than most would, so it's genuinely minor.

**Suggested:** a `NATIVE_RUN_JUSTIFIED` set that the test unions into the expected names. That records "justified by a native run rather than by config" as data, keeps the guard strict for everything else, and is where `ExtensionRegistry` would go if the telemetry question in comment 1 resolves that way.

pgermosen (approved):
The diagnosis is right and the fix is the shape Quarkus documents. What I verified rather than assumed:

**The registered list is exactly the config-resolved set.** I built `ConsumerConfig` with the eight properties `PipesKafkaConsumerManager:139-147` actually sets, and collected every value that names a loadable class:

| setting | resolves to | form |
|---|---|---|
| `key.deserializer` / `value.deserializer` | `ByteArrayDeserializer` | `Class` |
| `partition.assignment.strategy` | `RangeAssignor`, `CooperativeStickyAssignor` | `Class` |
| `metric.reporters` | `JmxReporter` | **`String`** |
| `sasl.oauthbearer.jwt.retriever.class` | `DefaultJwtRetriever` | `Class` |
| `sasl.oauthbearer.jwt.validator.class` | `DefaultJwtValidator` | `Class` |

Four registered, two excluded by the `sasl.` filter. Nothing missing, nothing spare.

- **Commit 2's `case String name` branch was necessary, not cleanup.** `metric.reporters` really does resolve to a class *name* rather than a `Class`, so before it `JmxReporter` was registered but unguarded.
- **The jose4j hazard is real.** `DefaultJwtValidator` is a default even on a PLAINTEXT config, so the `sasl.`/`ssl.` filter does real work. `security.protocol` resolves to `PLAINTEXT` and `PipesKafkaConsumerManager` sets no `sasl.*`/`ssl.*`, so the premise holds today.
- **`group.protocol` is `classic`** in 4.3.0, so both assignors are genuinely instantiated.
- **The holder pattern is documented, not just local precedent.** Quarkus: *"If your class is in a third-party jar, you can do it by using an empty class that will host the `@RegisterForReflection`"*, and *"`MyClassRequiringReflection` ... will be registered for reflection but not `MyReflectionConfiguration`."* Using `targets =` rather than `DockerJavaNativeSupport`'s `classNames =` is a strengthening: kafka-clients is a direct compile dependency (`pom.xml:136-140`), so the class literals cost nothing and are compile-checked.
- **No overlap with the other native-config convention** — `META-INF/native-image/reflect-config.json` exists but has zero kafka entries.
- **The swallow is where the description says**: `PipesPoller:154`, pipe left `RUNNING`.

**I will approve.** The registration is correct, and I'd merge it. Comment 2 is the one we should fix before merge, since the javadoc currently states a guarantee the code doesn't provide. Comments 1, 3 and 4 are fine as follow-ups.

---

### Your binary-size table reproduces — there is nothing to trim

This is the one claim a reviewer can't check by reading, so I built it. Three container builds off `0142dc43` (Mandrel 25.0.3, `-O2`, linux/amd64):

| metric | base (`main`) | one class | four (PR) | Δ one | Δ four |
|---|---:|---:|---:|---:|---:|
| file size | 201.41MB | 215.86MB | 216.99MB | **+14.45MB** | **+15.58MB** |
| code area | 92.99MB | 102.86MB | 103.33MB | +9.86MB | +10.34MB |
| image heap | 108.00MB | 112.59MB | 113.25MB | +4.59MB | +5.24MB |
| reachable types | 37,822 | 40,076 | 40,215 | **+2,254** | +2,393 |
| reachable methods | 197,751 | 210,244 | 210,932 | **+12,493** | +13,181 |

Different OS and architecture from yours, same shape: **the first registration costs ~93% of the total and the other three add 1.13 MB between them.**

A `PrintUniverse` diff attributes the 2,742 new universe types:

| | count |
|---|---:|
| `org.apache.kafka` | **1,938** |
| — `common.message` (generated protocol schemas) | 544 |
| — `clients.consumer` (coordinator, fetcher, group internals) | 363 |
| — `common.requests` / `common.protocol` / `common.errors` | 578 |
| — `common.security` | 83 |
| — `shaded.io` + `shaded.com` (OpenTelemetry, protobuf) | 67 |
| `java.lang.invoke` + `String$$StringConcat` (lambda forms, indy) | 528 |
| lz4 / snappy / zstd / xxhash codecs | ~68 |
| Kerberos / GSSAPI | 49 |

The cost isn't metadata volume, and it isn't JMX — `common.metrics` is 13 types, and `JmxReporter` isn't even in the one-class build. The closed-world analysis can't see through `Utils.newInstance(Class.forName(...))`, so on `main` the consumer's entire constructed graph is dead code. The first reflective instantiation makes it live, and over half of what arrives is Kafka's auto-generated wire-protocol layer. **"There is nothing to trim" is correct.**

Two side observations from the same data:

- **It supports the SASL exclusion.** `common.security` and the Kerberos/GSSAPI packages are pulled in regardless — statically reachable but not reflectively instantiable, which is precisely the distinction the `sasl.`/`ssl.` filter depends on.
- **The remaining size question is a product call, not a code one.** 7.7% growth on every user's pull, most of whom never touch Kafka pipes, versus a shipped feature that currently does nothing in the native image. Worth an explicit maintainer decision rather than landing as a side effect.

One last non-blocking question: **"AppInfoParser and the telemetry reporter logging normally"** in commit 2 — does that mean JMX MBean registration now logs a warning under native? If so, one javadoc line saves someone from diagnosing it as a new problem later.

LiamMacP:
@pgermosen Thanks for the review – the size reproduction and `PrintUniverse` breakdown was more than I could verify myself, and they turn "most plausible mechanism" in the description into something demonstrated. I've reworded that paragraph to your framing.

Two commits, one per comment.

**Comment 2 (blocking) — fixed in `39382a18`.** You were right that the javadoc claimed guarantees the code didn't provide. `consumerProperties(Pipe)` is now package-private, `createConsumer` calls it, and the test reads it rather than re-declaring the settings. I ran your exact repro against the fix:

```java
properties.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG,
        "org.apache.kafka.clients.consumer.RoundRobinAssignor");
```

```
omits them, so a native build cannot construct a consumer: [org.apache.kafka.clients.consumer.RoundRobinAssignor]
```

Green before, fails now.

**Comment 1 — fixed in `9b910450`, taking your first option.** `enable.metrics.push=false` in `consumerProperties`. It's right that neither broker would have shown this, so the 5/5 wasn't evidence for that entry; turning the path off removes the question instead of documenting it, and Floci shouldn't be pushing client telemetry to a user's broker regardless. Verified natively: both paths still 5/5, and `ClientTelemetryReporter` no longer appears in the logs at all. Binary size unchanged at 202.1MB — consistent with your finding that `shaded.*` stays statically reachable either way; only the reflective path goes.

**Comment 3 — reworded.** "no suite does that" → the compatibility suite already runs against the native image, it has no Kafka source case yet. Thanks for the correction; I'd checked `pipes.bats` for Kafka but not that the native half of the harness was already wired up.

**Comment 4 — deferring, with a caveat.** With metrics push off, `NATIVE_RUN_JUSTIFIED` would be an empty set plus machinery, so I've left it out. The point stands though: whoever adds the first genuinely native-justified entry will have to edit the guard, and that diff will need the justification in its commit message rather than in the test.

**On the compat gap:** agreed this is the thing that actually would have caught #1260, and it's a broker container on `compat-net` rather than new infrastructure. Happy to look into this when I have time.

hectorvent:
Thanks @LiamMacP @pgermosen 
Merging all good

hectorvent (approved):

PR #2070, opened 2026-07-31
fix(cognito): enforce user pool password policies

### LINKED ISSUE #2066: [BUG] Cognito user pool password policy is not enforced
## Service

Cognito User Pools

## AWS API Action

`SignUp`, `AdminCreateUser`, `AdminSetUserPassword`, `ChangePassword`, and
`ConfirmForgotPassword` when the user pool has a configured
`Policies.PasswordPolicy`.

## Expected behavior

Every password-setting path must enforce the user pool's configured password
policy:

- `MinimumLength`
- `RequireUppercase`
- `RequireLowercase`
- `RequireNumbers`
- `RequireSymbols`
- `PasswordHistorySize`

A password that violates the policy, or reuses a password covered by the
configured history, must fail with `InvalidPasswordException`.

AWS documentation:

- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_PasswordPolicyType.html
- https://docs.aws.amazon.com/cognito/latest/developerguide/managing-users-passwords.html

## Actual behavior

Current Floci `main` stores `Policies.PasswordPolicy` on the user pool but does
not validate passwords against it. Passwords that violate the configured
minimum length or character requirements are accepted, and recently used
passwords can be reused.

Verified against upstream `main` at
`0142dc434f98ba14a13236196c72036839a1e84b`.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 cognito-idp create-user-pool \
  --pool-name strict-password-policy \
  --policies '{
    "PasswordPolicy": {
      "MinimumLength": 12,
      "RequireUppercase": true,
      "RequireLowercase": true,
      "RequireNumbers": true,
      "RequireSymbols": true,
      "PasswordHistorySize": 10
    }
  }'

# Create an app client for the returned pool, then:
  --client-id <client-id> \
  --password 'Short1!'
```

Floci currently accepts the password. Amazon Cognito rejects it with
`InvalidPasswordException`.

## Environment

- Floci version / image tag: upstream `main` at `0142dc434f98ba14a13236196c72036839a1e84b`
- Java SDK version (if applicable): AWS CLI reproduction
- How you're running Floci (Docker / native / `mvn quarkus:dev`): any

dzwicker:
## Summary

Enforce the configured Cognito user pool password policy when passwords are created or changed. This covers minimum length, uppercase, lowercase, numeric and symbol requirements, as well as password history.

Closes #2066

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Cognito rejects passwords that do not satisfy the user pool `PasswordPolicy`. Floci previously accepted them. Validation is now centralized in the password update path and returns the Cognito `InvalidPasswordException` shape. The symbol set follows the Cognito password policy documentation, and whitespace alone does not satisfy `RequireSymbols`.

Verified through service tests and an SDK integration test that signs up against a user pool with a configured password policy.

## Checklist

### Test evidence

- `JAVA_HOME=/Users/dzwicker/.sdkman/candidates/java/25-open ./mvnw -q -Dtest=CognitoServiceTest test`
- `JAVA_HOME=/Users/dzwicker/.sdkman/candidates/java/25-open ./mvnw -q -Dtest=CognitoIntegrationTest test`
- `JAVA_HOME=/Users/dzwicker/.sdkman/candidates/java/25-open ./mvnw -q test`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -68,6 +68,8 @@ import static io.github.hectorvent.floci.core.common.ReservedTags.rejectUnknownR
 @ApplicationScoped
 public class CognitoService {
     private static final int DEFAULT_REFRESH_TOKEN_VALIDITY_DAYS = 30;
+    private static final String COGNITO_PASSWORD_SYMBOLS =
+            "^$*.[]{}()?\"!@#%&/\\,><':;|_~`=+-";
 
     private static final Logger LOG = Logger.getLogger(CognitoService.class);
     private static final ObjectMapper MAPPER = new ObjectMapper();
@@ -2326,6 +2328,10 @@ public class CognitoService {
     }
 
     private void updateUserPassword(CognitoUser user, String password) {
+        UserPool pool = describeUserPool(user.getUserPoolId());
+        validatePasswordAgainstPolicy(pool, user, password);
+        String previousPasswordHash = user.getPasswordHash();
+        String passwordHash = hashPassword(password);
         String saltHex = CognitoSrpHelper.generateSalt();
         String verifierHex = CognitoSrpHelper.computeVerifier(
                 CognitoSrpHelper.extractPoolName(user.getUserPoolId()),
@@ -2333,9 +2339,84 @@ public class CognitoService {
                 password,
                 saltHex
         );
-        user.setPasswordHash(hashPassword(password));
+        user.setPasswordHash(passwordHash);
+        updatePasswordHistory(pool, user, previousPasswordHash);
         user.setSrpSalt(saltHex);
         user.setSrpVerifier(verifierHex);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void validatePasswordAgainstPolicy(UserPool pool, CognitoUser user, String password) {
+        Map<String, Object> policies = pool.getPolicies();
+        if (policies == null || !(policies.get("PasswordPolicy") instanceof Map<?, ?> rawPolicy)) {
+            return;
+        }
+
+        Map<String, Object> policy = (Map<String, Object>) rawPolicy;
+        boolean invalid = password == null
+                || password.length() < policyInt(policy, "MinimumLength")
+                || policyBoolean(policy, "RequireUppercase")
+                    && password.codePoints().noneMatch(Character::isUpperCase)
+                || policyBoolean(policy, "RequireLowercase")
+                    && password.codePoints().noneMatch(Character::isLowerCase)
+                || policyBoolean(policy, "RequireNumbers")
+                    && password.codePoints().noneMatch(Character::isDigit)
+                || policyBoolean(policy, "RequireSymbols")
+                    && password.codePoints().noneMatch(
+                            codePoint -> COGNITO_PASSWORD_SYMBOLS.indexOf(codePoint) >= 0);
+
+        String passwordHash = password == null ? "" : hashPassword(password);
+        int historySize = policyInt(policy, "PasswordHistorySize");
+        boolean reused = historySize > 0 && (passwordHash.equals(user.getPasswordHash())
+                || user.getPasswordHistory().stream().limit(historySize).anyMatch(passwordHash::equals));
+
+        if (invalid || reused) {
+            throw new AwsException(
+                    "InvalidPasswordException",
+                    "Password does not conform to the configured password policy.",
+                    400
+            );
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private void updatePasswordHistory(UserPool pool, CognitoUser user, String previousPasswordHash) {
+        Map<String, Object> policies = pool.getPolicies();
+        if (policies == null || !(policies.get("PasswordPolicy") instanceof Map<?, ?> rawPolicy)) {
+            return;
+        }
+
+        int historySize = policyInt((Map<String, Object>) rawPolicy, "PasswordHistorySize");
+        if (historySize <= 0 || previousPasswordHash == null) {
+            return;
+        }
+
+        List<String> history = new ArrayList<>();
+        history.add(previousPasswordHash);
+        history.addAll(user.getPasswordHistory());
+        user.setPasswordHistory(history.stream().limit(historySize).toList());
+    }
+
+    private int policyInt(Map<String, Object> policy, String key) {
+        Object value = policy.get(key);
+        if (value instanceof Number number) {
+            return number.intValue();
+        }
+        if (value instanceof String stringValue) {
+            try {
+                return Integer.parseInt(stringValue);
+            } catch (NumberFormatException ignored) {
+                return 0;
+            }
+        }
+        return 0;
+    }
+
+    private boolean policyBoolean(Map<String, Object> policy, String key) {
+        Object value = policy.get(key);
+        return value instanceof Boolean booleanValue
+                ? booleanValue
+                : Boolean.parseBoolean(String.valueOf(value));
     }
 
     int getAccessTokenExpiresInSeconds(UserPoolClient client) {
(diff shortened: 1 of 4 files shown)


hectorvent:
Thank you for centralizing this in the password update path with solid service and integration tests, the composition checks and the Cognito symbol set look right.

1. **(blocking)** Password-history reuse should raise `PasswordHistoryPolicyViolationException`, not `InvalidPasswordException`. The model declares it specifically for reuse on AdminSetUserPassword, ChangePassword, ConfirmForgotPassword and RespondToAuthChallenge: https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json (see PasswordHistoryPolicyViolationException). The new `passwordHistoryRejectsARecentlyUsedPassword` test currently locks in the wrong code. Also note AWS counts the current password within `PasswordHistorySize`, while the check here treats it as an extra entry (N+1).
2. **(follow-up, separate PR)** Pools created without `Policies` skip validation entirely; real Cognito applies a default policy (minimum length 8, all requirements enabled).

One blocker; happy to re-review once the history error code is split out.

pgermosen:
Took a pass over this one independently. The centralization here is good work, and I want to be
specific about that before the findings: I went looking for a password-setting path that skips
validation and there isn't one. `adminCreateUser` (temporary password), `adminCreateMigratedUser`,
`adminSetUserPassword`, `signUp` and `changePassword` all route through `updateUserPassword`, and
`confirmForgotPassword` and the `NEW_PASSWORD_REQUIRED` challenge reach it via `adminSetUserPassword`.
The symbol set is exactly the documented Cognito set, and space is correctly excluded — I checked that
`Abcd 1234` is rejected under `RequireSymbols`, which is what the description claims. `InvalidPasswordException`
at 400 is also right: the model gives that shape no `httpStatusCode`, so the protocol default applies.

## On the open blocker

@hectorvent's point about `PasswordHistoryPolicyViolationException` holds, and it's marginally broader
than stated — the model declares it on **five** operations, not four:

```
AdminRespondToAuthChallenge, AdminSetUserPassword, ChangePassword,
ConfirmForgotPassword, RespondToAuthChallenge
```

Two things that should make this smaller than it looks:

- All five funnel through `updateUserPassword` in this codebase, so a single split in that helper covers
  every operation that declares the exception — no per-operation plumbing needed.
- Raising the new code from the shared helper won't leak an undeclared error onto `SignUp` or
  `AdminCreateUser` (neither declares it), because neither can trigger reuse: both construct a fresh
  `CognitoUser`, so `passwordHash` is null and the history list is empty. Worth a short comment saying
  so, since a later refactor could quietly break that assumption.

`validatePasswordAgainstPolicy` currently collapses the two conditions into one boolean
(`invalid || reused`), so they have to be separated to raise two codes anyway — that's the natural place
for the change.

One thing in your favour on the test: `passwordHistoryRejectsARecentlyUsedPassword`'s *scenario* is
correct. Its pool uses `PasswordHistorySize: 10` with only two changes, which real Cognito would reject
too. Only the asserted error code needs to change, not the test.

On the N+1 counting, here's a concrete repro on a pool with `PasswordHistorySize: 1`:

| step | result |
|---|---|
| set A, then set B | ok |
| set B again (the current password) | 400 — correct, that's the one previous password |
| set A (two changes back) | **400 — real Cognito allows this at size 1** |

The model's wording is *"users can't set a password that matches any of n previous passwords"*, and the
password being replaced is one of those n. The check here blocks the current hash **plus** n stored
entries, so n+1.

## Two more worth folding into the same round

**1. `AdminResetUserPassword` drops the active password out of the history window.** This is the one I'd
most want fixed alongside the exception split, because it's a silent bypass of the policy this PR is
adding. `adminResetUserPassword` sets `passwordHash = null` without pushing the outgoing hash into
`passwordHistory`, and `updatePasswordHistory` then returns early on `previousPasswordHash == null`, so
that password is lost rather than recorded. Confirmed end to end against a running instance, on a pool
with `PasswordHistorySize: 5`:

```
AdminSetUserPassword B, then B again  → 400  blocked, correct
AdminResetUserPassword                → 200
AdminSetUserPassword B                → 200  ← history policy bypassed
```

So any admin reset hands the user straight back the password they just had. Credit where it's due —
Greptile did catch this, but it's in its confidence-score paragraph rather than a line comment, which is
probably why it didn't carry into the review thread.

**2. A policy that omits `MinimumLength` gets no minimum at all.** `policyInt` returns `0` for an absent
key, so a pool with `PasswordPolicy: {RequireSymbols: true}` accepts a two-character password:

```
AdminSetUserPassword "a!"  → 200
```

The model notes `MinimumLength` *"can't be less than 6"*, and Cognito defaults it to 8. This is the same
shape as @hectorvent's follow-up #2, but it happens inside a policy that *is* present, so that item
doesn't cover it.

Both of those have a single fix, and it's probably worth doing here rather than deferring: normalise the
policy when the pool is created. `createUserPool` stores whatever came in verbatim
(`CognitoService.java:279`), which leaves every downstream reader to re-derive defaults. Filling AWS's
defaults once at creation closes the missing-`MinimumLength` case and the no-`Policies` case together,
and has the side benefit of making `DescribeUserPool` return a fully-populated policy the way real
Cognito does.

## State of the branch

Rebased onto current `main` and merges cleanly. Locally: `CognitoServiceTest` 143 tests green,
the remaining Cognito classes 180 tests green, no failures. Nothing here is a build problem — it's all
about which behaviour the code encodes.

Sorry this has sat a while. Happy to re-review promptly once the history exception is split out; the
reset-path fix is small and in the same area, so it should be one round rather than two.

pgermosen:
Checking in on this — re-verified against the current code, and the three items from my last review are all still open (no commits have landed since):

1. **`PasswordHistoryPolicyViolationException` split** — `CognitoService.java` still only throws `InvalidPasswordException` for both an invalid password and a reused one. Real Cognito declares `PasswordHistoryPolicyViolationException` on five operations (`AdminRespondToAuthChallenge`, `AdminSetUserPassword`, `ChangePassword`, `ConfirmForgotPassword`, `RespondToAuthChallenge`), all of which route through `updateUserPassword` here — so this is one split in that shared helper, not five separate changes.

2. **`AdminResetUserPassword` bypasses the history policy** — `adminResetUserPassword` still sets `passwordHash = null` directly without pushing the outgoing hash into `passwordHistory`, so `updatePasswordHistory` returns early and the password is lost from history rather than recorded. Concretely: `AdminSetUserPassword B` → `AdminResetUserPassword` → `AdminSetUserPassword B` again succeeds, handing the user back the password they just had, on a pool where reuse should be blocked.

3. **A policy without `MinimumLength` enforces no minimum** — `policyInt()` still returns `0` when the key is absent, so a pool with e.g. only `RequireSymbols: true` set accepts a 2-character password. AWS defaults `MinimumLength` to 8 and never allows less than 6.

None of this is a build problem — `CognitoServiceTest` and the rest of the Cognito suite are green, and the branch still merges cleanly onto `main`. It's specifically these three behaviors. Happy to re-review promptly once they're in — as noted before, the history-exception split and the reset-path fix touch the same area, so this should be one round.

pgermosen (approved):
@dzwicker thanks for the original diagnosis and implementation here — the centralization through `updateUserPassword` and the Cognito symbol set were right from the start, which is what made the rest of this tractable. I pushed two follow-up commits closing the three items from review: `PasswordHistoryPolicyViolationException` split out from `InvalidPasswordException` for reuse specifically (with the N+1 counting fixed alongside it — AWS counts the current password as one of the n, not an extra entry on top), `adminResetUserPassword` no longer discarding the outgoing password from history, and `createUserPool` filling in AWS's defaults for a password policy that's present but leaves fields like `MinimumLength` unset.

@greptile-apps also caught a real follow-on bug in the reset fix — capping history at n-1 assumed a current password always occupies one slot, which isn't true right after a reset clears it without replacing it. That's fixed now too: both the check and the archival look at whether a current password actually exists at the moment they run, so a reset doesn't itself shrink the window before a new password takes the freed slot back. Traced the exact `PasswordHistorySize: 2` / three-password / reset scenario by hand before writing it as a test.

Approving.

PR #2073, opened 2026-07-31
feat(lambda): support extra /etc/hosts entries on Lambda launch

floverfelt:
## Summary

This adds the ability to specify extra /etc/hosts entries in the lambda containers that Floci launches.

This provides additional flexibility for when the spawned lambda containers need to communicate in unusual networking setups - e.g. talk back to a service running on localhost instead of on the internal docker network.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Checklist


hectorvent (approved):
Thanks @floverfelt!

PR #2077, opened 2026-08-01
fix(lambda): reset volatile DynamoDB Streams ESM checkpoints on restart

### LINKED ISSUE #2076: [BUG] Persisted DynamoDB Streams ESM checkpoints outlive volatile stream sequence numbers, silently dropping records after restart
## Service

Lambda (DynamoDB Streams event source mapping), DynamoDB Streams

## AWS API Action

Event Source Mapping (DynamoDB Streams) → Invoke

## Expected behavior

On real AWS, a DynamoDB Streams ESM keeps delivering new stream records after
the consumer infrastructure restarts. Records written after a restart are
never silently skipped.

With Floci, since stream records and their sequence numbers do not survive a
container restart (sequence numbering restarts from
`000000000000000000001`), I would expect one of:

- ESM shard checkpoints (`shardSequenceNumbers`) are **not** persisted across
  restarts (matching the volatility of the stream itself), or
- checkpoints are invalidated at startup when the stream epoch changes.

## Actual behavior

`lambda-esm.json` persists `shardSequenceNumbers` on graceful shutdown, e.g.:

```json
"shardSequenceNumbers" : {
  "shardId-0000000001-00000000001" : "000000000000000000634"
}
```

After a restart, the stream's sequence numbers start again from 1, but the
poller resumes from the persisted checkpoint (634 in this example). Every new
record is treated as already processed and is **silently dropped** — no error,
no log — until the new epoch's sequence numbers grow past the old checkpoint
value (634 more records in this example).

In our case an inbox-projection Lambda simply never received an INSERT record
after a Floci restart. The poller logged nothing; the record was consumed
into the void. This is hard to notice and diagnose precisely because the
failure is silent.

Log excerpt after restart (poller starts normally, then never delivers):

```
INFO  [io.git.hec.flo.ser.lam.DynamoDbStreamsEventSourcePoller] Started DynamoDB Streams polling for ESM 7355859b-... → arn:aws:dynamodb:ap-northeast-1:000000000000:table/example_inbox/stream/2026-07-25T06:10:11.982
(new INSERT into example_inbox — nothing happens, no invoke, no error)
```

Clearing `shardSequenceNumbers` to `{}` in `lambda-esm.json` before startup
restores delivery immediately, which we now do as a workaround on every
start.

## Reproduction

```bash
# 1. Start Floci (hybrid storage, persistent volume)
docker run -d --name floci -p 4566:4566 -e FLOCI_STORAGE_MODE=hybrid \
  -v floci-data:/app/data -v /var/run/docker.sock:/var/run/docker.sock \
  floci/floci:latest

# 2. Create a table with a stream, a Lambda, and an ESM
  --table-name t --billing-mode PAY_PER_REQUEST \
  --stream-specification StreamEnabled=true,StreamViewType=NEW_AND_OLD_IMAGES
# (create function `fn`, then:)
aws --endpoint-url http://localhost:4566 lambda create-event-source-mapping \
  --function-name fn --event-source-arn <stream-arn> --batch-size 10

# 3. Write a few items -> fn is invoked (checkpoint advances)
aws --endpoint-url http://localhost:4566 dynamodb put-item --table-name t \
  --item '{"pk":{"S":"1"}}'

# 4. Graceful restart (persists shardSequenceNumbers, resets stream sequences)
docker stop -t 30 floci && docker start floci

# 5. Write a new item -> fn is NEVER invoked, no error is logged
aws --endpoint-url http://localhost:4566 dynamodb put-item --table-name t \
  --item '{"pk":{"S":"2"}}'
```

Step 3 only needs enough writes for the persisted checkpoint to exceed the
sequence number of the post-restart record (a single write before restart is
already enough to drop the first write after restart).

## Environment

- Floci version / image tag: 1.5.34 (also reproduced on 1.5.33)
- Java SDK version (if applicable): n/a (AWS CLI v2 + Go SDK v2 clients)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (Docker Desktop on macOS), `FLOCI_STORAGE_MODE=hybrid` with a named volume

aniketshukla1:
A DynamoDB stream and its sequence numbers live only in memory (the record buffer and the AtomicLong counter in DynamoDbStreamService are never persisted), so a restart recreates the stream empty and its sequence numbers restart from 000000000000000000001. An event source mapping, however, persists its shard checkpoint (shardSequenceNumbers) on graceful shutdown. After a restart the poller resumed from that stale checkpoint with an AFTER_SEQUENCE_NUMBER iterator, which points past every record in the new stream epoch — so every freshly written record was silently skipped (no invoke, no error, no log) until the new sequence numbers climbed back above the stale value.

Discard persisted shard checkpoints for DynamoDB Streams ESMs at startup, so the poller resumes from TRIM_HORIZON and delivers records from the new epoch, matching the volatility of the stream itself. The invalidation is now logged instead of failing silently.

Fixes #2076.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/DynamoDbStreamsEventSourcePoller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/DynamoDbStreamsEventSourcePoller.java
@@ -62,10 +62,36 @@ public class DynamoDbStreamsEventSourcePoller implements Resettable {
     public void startPersistedPollers() {
         for (EventSourceMapping esm : esmStore.listAll()) {
             if (esm.isEnabled() && esm.getEventSourceArn().contains(":dynamodb:")) {
+                discardStaleShardCheckpoints(esm);
                 startPolling(esm);
             }
         }
         LOG.infov("DynamoDbStreamsEventSourcePoller initialized");
+    }
+
+    /**
+     * Discards any shard checkpoints a DynamoDB Streams ESM persisted during a previous run,
+     * before its poller is (re)started at startup.
+     *
+     * <p>A DynamoDB stream is volatile: its record buffer and its {@code AtomicLong} sequence
+     * counter live only in memory (see {@link DynamoDbStreamService}) and are never persisted, so a
+     * restart recreates the stream empty and its sequence numbers start over from
+     * {@code 000000000000000000001}. A {@code shardSequenceNumbers} checkpoint saved during the
+     * previous run therefore points <em>past</em> every record in the new stream epoch: resuming
+     * from it with an {@code AFTER_SEQUENCE_NUMBER} iterator silently skips every freshly written
+     * record — no invoke, no error, no log — until the new sequence numbers climb back above the
+     * stale value. Clearing the checkpoint lets the poller resume from {@code TRIM_HORIZON}, which
+     * matches the volatility of the stream itself. See issue #2076.
+     */
+    private void discardStaleShardCheckpoints(EventSourceMapping esm) {
+        if (esm.getShardSequenceNumbers().isEmpty()) {
+            return;
+        }
+        LOG.infov("DynamoDB Streams ESM {0}: discarding {1} stale shard checkpoint(s) persisted by a "
+                        + "previous run (stream sequence numbers reset on restart); resuming from TRIM_HORIZON",
+                esm.getUuid(), esm.getShardSequenceNumbers().size());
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thanks @aniketshukla1

PR #2079, opened 2026-08-01
feat(lambdamicrovms): AWS Lambda MicroVMs service module and its CloudFormation types

### LINKED ISSUE #2078: [FEAT] Lambda MicroVMs: emulate the service (images, MicroVMs, network connectors) and its two CloudFormation types
## Service

Lambda MicroVMs (`Lambda Microvms`, apiVersion 2025-09-09) and its network connectors (`Lambda Core`, apiVersion 2026-04-30), plus the two CloudFormation resource types.

## API Action / Feature

Emulate the MicroVMs control plane. Floci has no MicroVM code today, so any SDK or CDK app touching the service fails outright.

**Images, versions, builds:** `CreateMicrovmImage`, `GetMicrovmImage`, `ListMicrovmImages`, `UpdateMicrovmImage`, `DeleteMicrovmImage`, `ListMicrovmImageVersions`, `GetMicrovmImageVersion`, `UpdateMicrovmImageVersion`, `DeleteMicrovmImageVersion`, `ListMicrovmImageBuilds`, `GetMicrovmImageBuild`, `ListManagedMicrovmImages`, `ListManagedMicrovmImageVersions`

**MicroVMs:** `RunMicrovm`, `GetMicrovm`, `ListMicrovms`, `TerminateMicrovm`

**Network connectors:** `CreateNetworkConnector`, `GetNetworkConnector`, `ListNetworkConnectors`, `UpdateNetworkConnector`, `DeleteNetworkConnector`

**CloudFormation:** `AWS::Lambda::MicrovmImage`, `AWS::Lambda::NetworkConnector`

Both services sign as `lambda`, so they fit under the existing lambda service descriptor and route by path — `/2025-09-09/`, `/2026-04-04/`, and tags on the classic `/2017-03-31/tags/` routes. None of those prefixes collides with what `LambdaController` serves today, so no new service toggle is needed.

## AWS Documentation

- https://docs.aws.amazon.com/lambda/latest/dg/microvms.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-microvmimage.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-networkconnector.html

## Why is this needed?

The service went GA on 2026-06-22 and nothing emulates it. The immediate consumer is CloudFormation provisioning — a CDK app declaring `AWS::Lambda::MicrovmImage` and `AWS::Lambda::NetworkConnector` currently skips both resources — and the [KubeMicroVM](https://github.com/codriverlabs/KubeMicroVM) operator, whose UAT suite otherwise needs a live AWS account.

Worth flagging for reviewers: several members the Smithy models mark **optional are enforced by the live service**. A network connector rejects a request missing `ClientToken`, `OperatorRole`, `NetworkProtocol` or `AssociatedComputeResourceTypes` — four separate 400s for a client written from the model alone. Several resources are also returned in a different projection per operation rather than one shape reused. Those are recorded behaviors, not guesses; see the verification note on the PR.

- [x] Yes
- [ ] No

lex00:
## Summary

Fixes #2078. Adds the Lambda MicroVMs control plane — images with a version and build lifecycle, the managed base-image catalog, MicroVM CRUD, and VPC egress network connectors — plus CloudFormation provisioning for `AWS::Lambda::MicrovmImage` and `AWS::Lambda::NetworkConnector`.

Two commits, so the service module and the CFN wiring can be read separately.

Two AWS service models rather than one: `Lambda Microvms` (2025-09-09) for images and MicroVMs, `Lambda Core` (2026-04-30) for connectors under `/2026-04-04/`. Both sign as `lambda`, so both register under the existing lambda descriptor in `ResolvedServiceCatalog` and route purely by path — no new service toggle, and no collision with what `LambdaController` serves. Tagging rides the classic `/2017-03-31/tags/` routes.

Auth tokens, the per-VM endpoint stub, and the idle/suspend timers are deliberately out of scope: they belong to a full-fidelity MicroVMs emulator rather than to CloudFormation provisioning.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Wire shapes come from the vendored AWS service models, cross-checked against `aws-sdk-go-v2`. Where the models are silent or disagree with the service, behavior comes from responses **recorded against live AWS** (us-east-2, 2026-07-29 and 2026-07-30) rather than from inference. Verified end to end with the AWS CLI (`aws --endpoint-url ... lambda-microvms ...`) and with an SDK client.

Three findings reviewers may want to know about, since all three contradict the model:

1. **Members the model marks optional are enforced.** A connector rejects a request missing `ClientToken`, `OperatorRole`, `NetworkProtocol` or `AssociatedComputeResourceTypes`. `ClientToken` is the odd one — the model marks it optional *and* tags it `idempotencyToken`, which normally means the SDK generates one.
2. **Constraint violations answer `ValidationException`**, a type the Lambda Core model does not list on any connector operation, with the member path in camelCase even though every wire member is PascalCase. That layer also runs *before* the service's own required-member checks.
3. **Resources are returned in a different projection per operation.** An image has four shapes and a connector four; only the list summaries carry `Type`, and absent members are absent rather than null — a connector that has never been updated has no `LastUpdateStatus` at all.

Builds converge instantly rather than on a timer, the usual emulator trade. Where that would produce an inconsistent body it is handled explicitly: a create or update reports the transitional state and does **not** name the version it is minting as active, because a client reading otherwise would launch a MicroVM off a version that does not exist yet against real AWS.

## Checklist

- [x] `./mvnw test` passes locally — 8118 tests, 0 failures, 0 errors
- [x] New or updated integration test added — `LambdaMicrovmsIntegrationTest` (14 cases) and `CloudFormationLambdaMicrovmsIntegrationTest` (both CFN types through provision and stack delete)

Additionally verified against a conformance suite that replays the live recordings against an endpoint ([m80](https://github.com/INTENTIUS/m80)): **26 checks over the image, MicroVM and connector lifecycles, 0 failures**, with every step running rather than skipped behind an earlier one.

Docs added at `docs/services/lambda-microvms.md`, wired into the nav and the service index. `tools/docs/regen_action_docs.py --strict` is clean — the REST controllers are opt-in for generated action tables, so the table there is hand-written like other REST services.


lex00:
Noting this before the PR-count bot does: I know I am well over the advisory limit, and I would rather say up front how these fit together than have a maintainer work it out.

My other open PRs are one intentional series — CloudFormation and EC2 gaps found by running the [aws-bench](https://github.com/aws-bench/aws-bench) scenarios against Floci, tracked at lex00/floci#17. They are mostly small and independent of each other.

**This one is not part of that series.** It is a new service rather than a gap fix, so if it helps to triage it separately from the aws-bench batch, please do. It is also self-contained: it touches no existing service behavior beyond registering two controllers under the existing lambda descriptor and teaching `LambdaTagController` the MicroVM ARN forms, so it should not conflict with anything else I have open.

Happy to close or park any of the others if the backlog is the problem — say which and I will. And happy to split this into the service module and the CFN provisioner as two PRs if you would rather review them apart; I kept them together because the provisioner cannot be tested without the service, but they are already separate commits.

hectorvent:
Hi @lex00, 

Thanks for this PR. This is a great addition to Floci. 
Could you include compatibility-tests for this new addition. Tests will ensure future changes don't break any functionality.

lex00:
Added `compatibility-tests/sdk-test-awscli/test/lambda-microvms.bats` — 15 cases covering the image lifecycle, version and build convergence, the managed catalog, MicroVM run/get/terminate, tagging, and the connector lifecycle, plus three not-found cases and a region-isolation check.

The bats suite rather than the Java one: both `lambda-microvms` and `lambda-core` are AWS CLI services, and the suite's `amazon/aws-cli:latest` image already provides them, so this needs no new dependency. Verified green against a build of this branch.

hectorvent:
Thanks for an impressively grounded new service module, the per-operation projections and status-code fidelity against the published models (https://github.com/aws/api-models-aws/tree/main/models/lambda-microvms and models/lambda-core) are the kind of detail that usually gets skipped, and the CFN wiring with GetAtt cross-references is well tested.

1. **(blocking)** All service state lives in plain in-memory maps in src/main/java/io/github/hectorvent/floci/services/lambdamicrovms/LambdaMicrovmsService.java:1173. New services must wire storage through StorageFactory: CloudFormation stacks persist across restarts, so in persistent mode a restarted Floci has stacks pointing at images and connectors that vanished, and DeleteStack then fails on the 404.
2. **(follow-up, separate PR)** CFN physical ids should follow the registry schemas' primaryIdentifier (ImageArn for AWS::Lambda::MicrovmImage, Arn for AWS::Lambda::NetworkConnector) rather than the bare name and nc-id.

One blocker; also please rerun CI, the Step Functions failure is unrelated flake.

lex00:
The StorageFactory wiring is pushed. Images, MicroVMs and connectors go through StorageBackedMap now, sharing the lambda storage key so they inherit its mode and flush interval, and every path that mutated a stored object in place writes it back. A new persistence test covers the restart case over real files.

On the CI rerun, nothing ran on the push. There are no workflow runs for 338eea88 at all, only Greptile, and the last real runs on this branch are from 08-01 where Compatibility Tests ended action_required. I think it needs your approval rather than a rerun.

One other if you are in there anyway. On #1999 the Docs Action Tables job failed with "The job was not acquired by Runner of type hosted even after multiple attempts" after 18 minutes. It passes locally and that PR changes no handler actions.

src/main/java/io/github/hectorvent/floci/core/common/ResolvedServiceCatalog.java
```diff
@@ -81,7 +80,10 @@ public ResolvedServiceCatalog(EmulatorConfig config) {
                         "lambda", storageMode(config.storage().services().lambda().mode(), config.storage().mode()),
                         config.storage().services().lambda().flushIntervalMs(), null, ServiceProtocol.REST_JSON,
                         protocols(ServiceProtocol.REST_JSON),
-                        Set.of(), Set.of("lambda"), Set.of(), Set.of(LambdaController.class)),
+                        Set.of(), Set.of("lambda"), Set.of(),
+                        Set.of(LambdaController.class,
+                                io.github.hectorvent.floci.services.lambdamicrovms.LambdaMicrovmsController.class,
+                                io.github.hectorvent.floci.services.lambdamicrovms.LambdaNetworkConnectorsController.class)),
```
hectorvent: Please import of this classes instead of using them inline. 

```suggestion
                                LambdaMicrovmsController.class,
                                LambdaNetworkConnectorsController.class)),
```

lex00: Done in 69d4dbff. Added the two imports so the shortened references compile.

lex00:
Build and Test ran this time and caught one thing the storage wiring needed. The five persisted types were missing @RegisterForReflection, which PersistedModelReflectionTest fails on because native-image persistence cannot rebuild them on restart. Added in 5ddb3e7d. Everything else passed including the native and compat matrix.

hectorvent:
Thanks for turning this around so quickly, and sorry this sat waiting on a reply. The three stores now go through StorageFactory like the other Lambda services, and reusing the `lambda` storage key means MicroVM state inherits Lambda's configured mode and flush interval. CI is green on this head.

The account isolation claim in your commit message holds. `AccountAwareStorageBackend` resolves the account exactly as `RegionResolver.getAccountId` does, so the storage key prefix and the account in every minted ARN cannot disagree.

Adding `@RegisterForReflection` to the persisted types was the right instinct and closes a real gap: routing state through StorageFactory newly requires Jackson to deserialize those nested classes, which native image would not have supported otherwise. All five are covered.

1. **(follow-up, separate PR)** CFN physical ids could follow the registry schemas' primaryIdentifier (`ImageArn` for `AWS::Lambda::MicrovmImage`, `Arn` for `AWS::Lambda::NetworkConnector`) rather than the bare name and nc-id.

No blockers from my side.

hectorvent (approved):
Thanks @lex00

PR #2090, opened 2026-08-02
fix(elasticache): clean up container by id on readiness-timeout rollback

ahmedk20:
## Summary

`ElastiCacheContainerManager#start` registers the container in `activeContainers` **before** its final `waitForBackendReady(...)` probe, and that probe throws on timeout/interrupt. When it does, `start()` propagates the exception without ever returning a handle to the caller, so `ElastiCacheService#rollbackReplicationGroup` sees `handle == null`, skips the container stop, and leaves it running — only reaped later at emulator shutdown via `stopAll()`. A readiness timeout is the most likely provisioning failure, so this is the common case, not an edge one.

This adds an idempotent `stopByGroupId(groupId)` to `ElastiCacheContainerManager` that looks the container up in `activeContainers`, falling back to the deterministic container name for one that failed before registration. `rollbackReplicationGroup` now calls it unconditionally instead of gating `containerManager.stop(handle)` on `handle != null`.

Mirrors the identical fix already reviewed and merged for Neptune in #1742 (same root cause, same fix shape, same reviewer feedback already addressed).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior fixed: on a readiness-timeout during provisioning, the already-started container was never stopped by rollback, leaking a running Docker container per failed create.

## Checklist

- [x] `./mvnw test` passes locally (`ElastiCacheServiceTest`, new `ElastiCacheContainerManagerTest`)
- [x] New unit tests added: rollback-by-id on readiness-timeout (`ElastiCacheServiceTest`), deterministic-name fallback when nothing is registered (`ElastiCacheContainerManagerTest`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheService.java
@@ -104,11 +104,19 @@ public class ElastiCacheService {
             LOG.warnv("Error stopping proxy for replication group {0}: {1}", groupId, e.getMessage());
         }
         try {
-            // Stop by id, not handle: a readiness timeout in containerManager.start() throws
-            // after the container was created and registered but before the handle is returned,
-            // so cleaning up by handle here would miss (and orphan) it. stopByGroupId is
-            // idempotent, so it's safe when the container never started.
-            containerManager.stopByGroupId(groupId);
+            if (handle != null) {
+                // We have the exact handle from this request's start() call, so stop by it
+                // directly. Falling back to stopByGroupId here instead would look up whatever
+                // is currently registered for groupId, which could be a different container
+                // if an overlapping create for the same id raced ahead of this rollback.
+                containerManager.stop(handle);
+            } else {
+                // No handle: a readiness timeout in containerManager.start() throws after the
+                // container was created and registered but before the handle is returned, so
+                // cleaning up by handle here isn't possible. stopByGroupId is idempotent, so
+                // it's safe when the container never started.
+                containerManager.stopByGroupId(groupId);
+            }
         } catch (RuntimeException e) {
             LOG.warnv("Error stopping container for replication group {0}: {1}", groupId, e.getMessage());
         } finally {
--- a/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheService.java
@@ -104,9 +104,11 @@ public class ElastiCacheService {
             LOG.warnv("Error stopping proxy for replication group {0}: {1}", groupId, e.getMessage());
         }
         try {
-            if (handle != null) {
-                containerManager.stop(handle);
-            }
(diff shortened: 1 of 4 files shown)


pgermosen (approved):
Looks good

- **The root-cause claim is exactly right**: confirmed in `ElastiCacheContainerManager.start()` that `activeContainers.put(groupId, handle)` happens *before* the final `waitForBackendReady(...)` call, which throws a plain `RuntimeException` on timeout — so a readiness timeout really does leave the container registered but the handle never returned to the caller, exactly as described.
- **This PR already arrived at the same end-state as `#2092`, not just a partial version of it**: the four-commit history (fix → discover the by-name race Greptile flagged → fix the race with a `provisioningGroupIds` claim guard → comment trim) is the identical review cycle #2092 went through, and the final diff here is structurally the same fix: `provisioningGroupIds.add(groupId)` claims the id before any provisioning work starts, wrapped in `try/finally` so it's always released (normal return, propagated exception, or rollback), and a concurrent second create for the same id is rejected with the same `ReplicationGroupAlreadyExistsFault` used for an already-persisted group — closing the race at its root rather than narrowing the by-id lookup window.
- **The rollback branching matches the implementation's own reasoning**: `handle != null → containerManager.stop(handle)` (exact, never touches `stopByGroupId`) vs. `handle == null → stopByGroupId` (the only case where a handle was never available, i.e. the readiness-timeout path) — both branches have dedicated tests asserting the *other* method was never called, not just that cleanup happened.
- **`stopByGroupId`'s fallback container name is verified, not assumed**: traced `containerName(groupId)` → `ContainerStorageHelper.resourceName(config, "valkey", null, groupId)` → with no configured `docker.resourceNamespace` (the mocked `EmulatorConfig` in the new test), this resolves to exactly `floci-valkey-<groupId>`, matching what `ElastiCacheContainerManagerTest.stopByGroupIdRemovesByDeterministicNameWhenNothingRegistered` asserts.
- **The concurrency test is genuine, not hand-wavy**: `concurrentCreateForSameGroupIdIsRejectedWhileFirstIsProvisioning` uses two real threads and `CountDownLatch`s to block the first request mid-`containerManager.start()` and prove the second is rejected (with neither `stop` nor `stopByGroupId` touched) before letting the first complete and land successfully.

Ran the tests myself rather than trusting the PR's numbers: full `ElastiCache*Test` glob — 37/37 across 6 classes (`ElastiCacheServiceTest`, the new `ElastiCacheContainerManagerTest`, `ElastiCacheIntegrationTest`, `ElastiCacheMemcachedIntegrationTest`, `ElastiCacheMemcachedServiceTest`, `ElastiCacheQueryHandlerTest`), including the real Docker-backed integration tests, not just mocked unit tests — all pass, no regressions. Conventional Commits compliant across all four commits, no AI trailers, no CODEOWNERS-sensitive paths touched, no docs changes needed (matches `#2092`'s precedent that this is correctly labeled as internal-only with no AWS-facing behavior change).

No findings. This is a clean, well-tested mirror of an already-reviewed and merged fix.

PR #2092, opened 2026-08-02
fix(memorydb): clean up container by id on readiness-timeout rollback

ahmedk20:
## Summary

`MemoryDbContainerManager#start` registers the container in `activeContainers` **before** its final `waitForBackendReady(...)` probe, and that probe throws on timeout/interrupt. When it does, `start()` propagates the exception without ever returning a handle to the caller, so `MemoryDbService#rollbackBackend` saw `handle == null`, skipped the container stop, and left it running — only reaped later at emulator shutdown via `stopAll()`. A readiness timeout is the most likely provisioning failure, so this is the common case, not an edge one.

This adds an idempotent `stopByClusterName(name)` to `MemoryDbContainerManager` that looks the container up in `activeContainers`, falling back to the deterministic container name for one that failed before registration. `rollbackBackend` now calls it unconditionally instead of gating `containerManager.stop(handle)` on `handle != null`.

Making that fix effective also required splitting `rollbackBackend`'s single shared `try` into isolated per-step `try/catch` blocks: previously `proxyManager.stopProxy(name)` and the container stop were in one `try`, so a throwing `stopProxy` skipped the container cleanup entirely — the exact leak this rollback exists to prevent. The proxy stop is now also guarded by `handle != null` (the proxy only starts after the container is ready, so a null handle means it never started).

Mirrors the identical fix already reviewed and merged for Neptune in #1742 (same root cause, same fix shape, same reviewer feedback already addressed).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior fixed: on a readiness-timeout during provisioning, the already-started container was never stopped by rollback, leaking a running Docker container per failed create. Separately, a throwing proxy-stop during rollback could skip the container cleanup step entirely.

## Checklist

- [x] `./mvnw test` passes locally (`MemoryDbServiceTest`, new `MemoryDbContainerManagerTest`)
- [x] New unit tests added: rollback-by-name on readiness-timeout, proxy-throws-after-container-starts isolation (`MemoryDbServiceTest`), deterministic-name fallback when nothing is registered (`MemoryDbContainerManagerTest`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/memorydb/MemoryDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/memorydb/MemoryDbService.java
@@ -65,6 +65,7 @@ public class MemoryDbService {
     private final EmulatorConfig config;
     private final RegionResolver regionResolver;
     private final Set<Integer> usedPorts = ConcurrentHashMap.newKeySet();
+    private final Set<String> provisioningClusterNames = ConcurrentHashMap.newKeySet();
 
     @Inject
     public MemoryDbService(MemoryDbContainerManager containerManager,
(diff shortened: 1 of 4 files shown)


pgermosen (approved):
Looks good

- **The root-cause claim is exactly right**: confirmed in `MemoryDbContainerManager.start()` that `activeContainers.put(clusterName, handle)` happens *before* the final `waitForBackendReady(...)` call, which throws a plain `RuntimeException` on timeout — so a readiness timeout really does leave the container registered but the handle never returned to the caller, exactly as described.
- **The precedent is real, not just asserted**: `#1742` (the referenced Neptune fix this mirrors) is genuinely merged, and the sibling `#2090` (ElastiCache, same author, same pattern, currently open) went through the identical evolution — Greptile flagging the same by-name race, followed by the same "claim the name" fix. This PR's three-commit shape (fix → discover the race → fix the race) isn't scope creep, it's the same review cycle that already played out on the sibling PR, applied proactively here.
- **The concurrency fix is correct and genuinely exercised**: `provisioningClusterNames.add(name)` claims the name atomically before any provisioning work starts, and the `try/finally` around the whole method body guarantees release on every exit path (normal return or any exception, since nothing runs before the `try` except the pre-existing already-exists check and the claim itself). `concurrentCreateForSameNameIsRejectedWhileFirstIsProvisioning` isn't a hand-wavy test — it uses two real threads and `CountDownLatch`s to genuinely block the first request mid-`containerManager.start()` and prove the second is rejected without touching the first's container, then lets the first complete successfully.
- **The rollback branching matches the implementation's own reasoning**: `handle != null → containerManager.stop(handle)` (exact, never touches `stopByClusterName`) vs. `handle == null → stopByClusterName` (the only case where a handle was never available) — both branches have dedicated tests that assert the *other* method was never called, not just that cleanup happened.

PR #2094, opened 2026-08-02
fix(ec2): return AWS-parity MissingParameter for CreateSubnet without VpcId

### LINKED ISSUE #2089: [BUG] Unable to create the Subnet under aws vps.
tried to create the aws subnet under default vpc using the floci-ui. getting error: The vpc ID 'null' does not exist
Payload:
{
"vpcId": "vpc-default",
"cidrBlock": "172.31.1.0/24"
}

when i tried with curl command, it works. Is it a UI bug
Versions:
floci/floci-ui:latest 03a261144e07 146MB 41.2MB U
floci/floci:latest b3b3a70a294b 709MB 179MB U

Agboolafeyikemi:
## Summary
- `CreateSubnet` looked up a null `VpcId` directly, producing `InvalidVpcID.NotFound: The vpc ID 'null' does not exist` instead of AWS's actual `MissingParameter` error for a missing required parameter.
- Now returns `MissingParameter: The request must contain the parameter VpcId`, matching the pattern already used for `RunInstances`' `ImageId` check and matching real EC2 behavior (`VpcId` is required for `CreateSubnet`).

Fixes #2089.

## Test plan
- [x] Reproduced the original error against unfixed code: `InvalidVpcID.NotFound: The vpc ID 'null' does not exist`
- [x] Confirmed the fix replaces it with `MissingParameter: The request must contain the parameter VpcId`
- [x] Added unit test (`Ec2ServiceTest`) and HTTP-level integration test (`Ec2IntegrationTest`)
- [x] `Ec2IntegrationTest`, `Ec2ServiceTest`, `Ec2ServiceConcurrencyTest`, `Ec2ServicePersistenceTest`, `Ec2Phase2IntegrationTest` all pass

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -1302,6 +1302,9 @@ public class Ec2Service implements ContainerTeardown {
     // ─── Subnets ───────────────────────────────────────────────────────────────
 
     public Subnet createSubnet(String region, String vpcId, String cidrBlock, String availabilityZone) {
+        if (vpcId == null || vpcId.isBlank()) {
+            throw new AwsException("MissingParameter", "The request must contain the parameter VpcId", 400);
+        }
         ensureDefaultResources(region);
         getRequiredVpc(region, vpcId);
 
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Thanks @Agboolafeyikemi!

PR #2108, opened 2026-08-03
feat(eks): support IRSA with a real OIDC issuer and JWT validation

### LINKED ISSUE #2107: [FEAT] EKS: DescribeCluster has no identity.oidc.issuer, so IRSA cannot be tested
## Service

EKS and STS

## API Action / Feature

`eks:DescribeCluster`, `identity.oidc.issuer` (IRSA OIDC provider), and real token validation in `sts:AssumeRoleWithWebIdentity`.

## AWS Documentation

- [`DescribeCluster`](https://docs.aws.amazon.com/eks/latest/APIReference/API_DescribeCluster.html)
  - see the `identity.oidc.issuer` field on the `Cluster` shape
- [`AssumeRoleWithWebIdentity`](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html)
- [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)
- [OIDC federation condition keys](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_iam-condition-keys.html)
  - `<oidcProvider>:sub` and `<oidcProvider>:aud`

## Why is this needed?

IAM Roles for Service Accounts (IRSA) is how a pod on EKS gets AWS credentials in production: it presents a projected service-account JWT to `sts:AssumeRoleWithWebIdentity` and receives temporary credentials. Against Floci today, none of that can be exercised.

Four gaps:

1. **`DescribeCluster` does not return `identity.oidc.issuer`.** The `Cluster` model has no `identity` field at all, so there is no OIDC provider URL to reference when building a trust policy.

2. **`AssumeRoleWithWebIdentity` does not validate the token.** It checks that `WebIdentityToken` is non-blank and then ignores it: no signature check, no `iss`, no `aud`, no expiry. It returns a hardcoded `SubjectFromWebIdentityToken` of `web-identity-subject` and a `Provider` of `accounts.google.com` regardless of what was sent.

3. **It never looks up the role or evaluates its trust policy.** The `AssumeRole` path does call into trust-policy enforcement; the web-identity path does not, so a token is never checked against the role it is assuming.

4. **`AssumeRolePolicyEvaluator` cannot express an IRSA trust policy anyway.** It reads only `Principal.AWS` (`Federated` is explicitly ignored), hardcodes `sts:AssumeRole` so `sts:AssumeRoleWithWebIdentity` never matches, and ignores `Condition` blocks entirely.

The practical effect: anything testing IRSA locally has to fake a JWT, and no part of the exchange is validated. The emulator cannot catch a wrong service account, a wrong audience, an expired token, or a bad signature - which are exactly the failures worth catching before production.

### Trust policy shape this needs to support

This is the standard IRSA trust policy that `eksctl`, Terraform, and the AWS console all produce. Note that the OIDC provider ARN and the condition keys use the issuer with the scheme stripped.

```json
{
  "Version": "2012-10-17",
  "Statement": [{
    "Effect": "Allow",
    "Principal": {
      "Federated": "arn:aws:iam::<accountID>:oidc-provider/<oidcProvider>"
    },
    "Action": "sts:AssumeRoleWithWebIdentity",
    "Condition": {
      "StringEquals": {
        "<oidcProvider>:sub": "system:serviceaccount:<namespace>:<serviceAccount>",
        "<oidcProvider>:aud": "sts.amazonaws.com"
      }
    }
  }]
}
```

### Expected behaviour

- `DescribeCluster` returns `identity.oidc.issuer` for every cluster.
- `AssumeRoleWithWebIdentity` verifies a token signed by that cluster's key: signature, `iss`, `aud`, expiry, and the role's trust policy including the `Condition` block.
- The response reports the token's real claims rather than placeholders.
- Tokens from issuers Floci does not host keep working as they do today, since Floci cannot verify a third-party provider's signature.

- [x] Yes
- [ ] No

mcarvalho-singlestore:
## Summary

Closes #2107

Every EKS cluster now gets an OIDC identity provider, so the full IRSA flow (trust policy, service-account token, `sts:AssumeRoleWithWebIdentity`) works end to end without mocked or hardcoded tokens.

**EKS.** `CreateCluster` generates an RSA-2048 signing keypair and an AWS-shaped issuer URL (`https://oidc.eks.<region>.amazonaws.com/id/<id>`), returned by `DescribeCluster` as `identity.oidc.issuer`.

**STS.** `AssumeRoleWithWebIdentity` now verifies tokens whose `iss` names an issuer Floci hosts: RS256 signature against the cluster's public key, exact `iss`, `aud` containing `sts.amazonaws.com`, `exp`/`nbf` with 60s of clock-skew
tolerance, and the role's trust policy. The response carries the token's real claims in `SubjectFromWebIdentityToken`, `Provider`, and `Audience` instead of the previous placeholders. A bad token returns `InvalidIdentityToken` (400); a
trust policy that does not permit the subject returns `AccessDenied` (403) — a check this path was missing entirely.

**Compatibility.** Tokens from an issuer Floci does not host are treated as opaque and accepted, since Floci cannot verify a third-party signature. Existing behaviour is unchanged and no configuration flag is required.

### Design notes

**The issuer URL is deliberately cosmetic.** In real IRSA, STS fetches the JWKS from the issuer; Floci resolves the key in-process through a new `OidcIssuerKeyLookup` port. That is what lets the URL stay AWS-shaped even though Floci's embedded DNS cannot resolve `*.amazonaws.com` and TLS is off by default — it only has to be a faithful string for trust-policy construction.

**Key material is stored separately from `Cluster`.** `Cluster` is serialized straight onto the `DescribeCluster` wire response, so a private key carried there would leak to any caller; `@JsonIgnore` would instead lose it across restarts and
break tokens minted before one. A dedicated `eks-oidc-keys.json` backend gets both properties, and an integration test asserts the response body contains no key material.

**`WebIdentityTrustPolicyEvaluator` is a sibling of `AssumeRolePolicyEvaluator`, not an extension of it.** That evaluator serves `sts:AssumeRole`, hardcodes that action, reads only `Principal.AWS`, and ignores conditions; widening it would
risk regressing a path that already works. Condition *values* are compared with exact, case-sensitive equality rather than the existing `globMatches`, which lowercases both operands — case-folding an OIDC subject would silently admit a
different service account. Condition *keys* remain case-insensitive, per AWS.

**Two `_floci` endpoints.** Floci has no kubelet to project a token into a pod, so `POST _floci/eks/clusters/{name}/oidc-token` mints one for a namespace/service-account pair; signing stays server-side so the private key never leaves Floci. OIDC discovery and JWKS are served under the same namespace, for fidelity and debuggability rather than because anything in the flow fetches them. Both follow the existing `_floci/eks/token-webhook` and `_floci/ecr/gc` precedent. Serving them at the AWS-shaped issuer path would collide with S3's path-style catch-all — the same hazard that forced the Cognito well-known routes to discriminate on an underscore.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Wire shapes verified against the AWS API references** for [`DescribeCluster`](https://docs.aws.amazon.com/eks/latest/APIReference/API_DescribeCluster.html) and [`AssumeRoleWithWebIdentity`](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html):

- `identity.oidc.issuer` is nested exactly as AWS returns it on the `Cluster` shape, and the issuer matches the real EKS form `https://oidc.eks.<region>.amazonaws.com/id/<32 hex chars>`.
- The STS XML response keeps the existing element order and adds no new elements — `Provider`, `Audience`, and `SubjectFromWebIdentityToken` were already present and are now populated from the token.
- Error shapes reuse `AwsQueryResponse.error` with the STS namespace: `InvalidIdentityToken` (400) and `AccessDenied` (403), matching the errors AWS documents for this action.
- The OIDC provider ARN and condition-key prefix use the issuer with the scheme stripped (`oidc.eks.<region>.amazonaws.com/id/<id>`), which is how the AWS console, `eksctl`, and Terraform all render them.

**Existing SDK compatibility tests are unchanged and still pass** — `compatibility-tests/sdk-test-java` (`StsTest#assumeRoleWithWebIdentity`) and `sdk-test-python` (`test_assume_role_with_web_identity`) both pass opaque tokens and continue to succeed, which is the compatibility guarantee this change is built around. No SDK-level IRSA test was added; the new end-to-end coverage is an integration test in the main suite.

## Checklist

## Docs

`docs/services/eks.md` gains an IRSA section (issuer, minting endpoint, trust policy, what STS validates, discovery routes); `docs/services/sts.md` gains a web identity validation section. The stale "`Condition` blocks are not evaluated" limitation there is now correctly scoped to the `AssumeRole` path.

No AWS operations were added — only `_floci` plumbing routes — so the operation counts for EKS and STS in `docs/services/index.md` are unchanged and remain accurate.

## Not included

`iam:CreateOpenIDConnectProvider` and the rest of the OIDC-provider CRUD set are still unimplemented; `ListOpenIDConnectProviders` continues to return an empty list. The IRSA flow does not require them, because the `Federated` ARN is only a string inside a policy document and is never resolved against a stored provider. Worth a separate PR for anyone whose Terraform or CDK path creates the provider explicitly.

Also unchanged: `Condition` blocks are still not evaluated on the `AssumeRole` path, so `sts:ExternalId` remains ignored there.

### DIFF
--- a/docs/services/eks.md
+++ b/docs/services/eks.md
@@ -146,6 +146,93 @@ services:
       FLOCI_SERVICES_EKS_MOCK: "true"
 ```
 
+## IRSA (IAM Roles for Service Accounts)
+
+Every cluster gets an OIDC identity provider, so the full IRSA flow (trust policy, service-account token, `sts:AssumeRoleWithWebIdentity`) works end to end without mocked or hardcoded tokens.
+
+`CreateCluster` generates an RSA-2048 signing keypair and an AWS-shaped issuer URL, returned by `DescribeCluster`:
+
+```json
+{
+  "cluster": {
+    "name": "my-cluster",
+    "identity": { "oidc": { "issuer": "https://oidc.eks.us-east-1.amazonaws.com/id/3F8A…" } }
+  }
+}
+```
+
+The issuer URL is a faithful string for building trust policies, but it is not fetched: Floci's STS resolves the signing key in-process. The private key is held in storage separate from the cluster model and is never returned by any API.
+
+### Minting a service-account token
+
+Real EKS has the kubelet project a token into the pod. Floci has no kubelet, so a local-dev harness requests one and writes it to the file named by `AWS_WEB_IDENTITY_TOKEN_FILE`. Signing happens server-side, so the private key never leaves Floci.
+
+```bash
+curl -sX POST http://localhost:4566/_floci/eks/clusters/my-cluster/oidc-token \
+  -H 'Content-Type: application/json' \
+  -d '{"namespace":"my-namespace","serviceAccount":"my-service-account"}'
+```
+
+```json
+{
+  "token": "eyJhbGciOiJSUzI1NiIs…",
+  "issuer": "https://oidc.eks.us-east-1.amazonaws.com/id/3F8A…",
+  "subject": "system:serviceaccount:my-namespace:my-service-account",
+  "audience": "sts.amazonaws.com"
+}
+```
+
+`audience` (default `sts.amazonaws.com`) and `expirySeconds` (default 24h, max 7d) are optional.
+This is Floci plumbing under `_floci/…`, not an AWS API.
+
+### Trust policy
+
+Note that the OIDC provider ARN and the condition keys use the issuer with the scheme stripped `oidc.eks.<region>.amazonaws.com/id/<id>`, which is how the AWS console, `eksctl`, and Terraform all render it.
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [{
+    "Effect": "Allow",
+    "Principal": {
+      "Federated": "arn:aws:iam::000000000000:oidc-provider/<oidcProvider>"
+    },
+    "Action": "sts:AssumeRoleWithWebIdentity",
+    "Condition": {
+      "StringEquals": {
+        "<oidcProvider>:sub": "system:serviceaccount:<namespace>:<serviceAccount>",
+        "<oidcProvider>:aud": "sts.amazonaws.com"
+      }
+    }
+  }]
+}
+```
+
+### What STS validates
+
+When `sts:AssumeRoleWithWebIdentity` receives a token whose `iss` names an issuer Floci hosts, it enforces all of:
+
+- the RS256 signature, against that cluster's public key
+- `iss` matches the cluster's issuer exactly
+- `aud` contains `sts.amazonaws.com`
+- `exp` / `nbf`, with 60s of clock-skew tolerance
+- the role's trust policy: `Principal.Federated` and the `Condition` block, comparing `oidc:sub` / `oidc:aud` with exact, case-sensitive equality
+
+The response then carries the token's real claims in `SubjectFromWebIdentityToken`, `Provider`, and `Audience`. Failures return `InvalidIdentityToken` (400) for a bad token or `AccessDenied` (403) when the trust policy does not permit the subject.
+
+A token whose issuer Floci does not host is treated as opaque and accepted, since Floci cannot adjudicate a third-party provider. Validation is therefore automatic for Floci-issued tokens and requires no configuration flag.
+
+### OIDC discovery endpoints
+
+Served for fidelity and debugging, nothing in the IRSA flow dereferences them:
+
+| Route | Description |
+|---|---|
+| `GET /_floci/eks/clusters/<name>/oidc/.well-known/openid-configuration` | OIDC discovery document |
+| `GET /_floci/eks/clusters/<name>/oidc/keys` | JWKS containing the cluster's public key |
+
+These live under `_floci/…` rather than at the AWS-shaped issuer path, which Floci's embedded DNS does not resolve and which would collide with S3's path-style routing.
+
 ## ARN Format
 
 ```
--- a/docs/services/sts.md
+++ b/docs/services/sts.md
@@ -32,11 +32,28 @@ created through IAM with a real trust policy.
 
 ### Known limitations
 
-- **`Condition` blocks are not evaluated.** A trust policy that requires `sts:ExternalId` (the
-  confused-deputy guard) is matched on its principal alone, so the role is assumable without passing
-  `ExternalId`, and the `ExternalId` request parameter is ignored. This matches moto/LocalStack.
+- **`Condition` blocks are not evaluated for `AssumeRole`.** A trust policy that requires
+  `sts:ExternalId` (the confused-deputy guard) is matched on its principal alone, so the role is
+  assumable without passing `ExternalId`, and the `ExternalId` request parameter is ignored. This
+  matches moto/LocalStack. Conditions *are* evaluated on the `AssumeRoleWithWebIdentity` path (see below).
 - **Only the trust policy is checked.** Cross-account `AssumeRole` in AWS also requires the caller's
   own identity policy to allow `sts:AssumeRole`; that side is not enforced.
+
+## Web Identity Validation (IRSA)
+
+`AssumeRoleWithWebIdentity` validates tokens minted by an OIDC issuer Floci hosts - currently an EKS cluster's IRSA provider. This needs no configuration flag and is independent of `FLOCI_SERVICES_IAM_ENFORCEMENT_ENABLED`.
+
+When the token's `iss` names a known Floci issuer, all of the following are enforced:
+
+- the RS256 signature, against the issuer's public key
+- `iss` matches that issuer exactly
+- `aud` contains `sts.amazonaws.com`
+- `exp` / `nbf`, with 60s of clock-skew tolerance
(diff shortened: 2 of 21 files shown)


pgermosen (commented):
No blocking findings. This is a serious, correctly executed piece of security-relevant code, not just wire-shape emulation.

`WebIdentityTokenVerifier`: `alg` is checked strictly equal to `"RS256"` before anything else, closing off the classic alg-confusion / alg:none bypass. Both split sites now use `-1` (confirmed fixed in both `verify()` and `parseClaims()`, not just the one Greptile originally flagged). Signature, `iss`, `aud`, `exp`/`nbf` are all independently enforced — `peekIssuer()` is correctly untrusted-by-design; actual enforcement always goes through `verify()` against the specific cluster's stored key, so a forged `iss` claim naming a real Floci-hosted issuer still fails on signature mismatch.

`identity.oidc.issuer` nesting matches the real [DescribeCluster](https://docs.aws.amazon.com/eks/latest/APIReference/API_DescribeCluster.html) response shape exactly (`{"identity":{"oidc":{"issuer":...}}}`).

Checked one thing the PR didn't explicitly address: [AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html)'s action-specific error list documents `IDPRejectedClaim` (403) but not `AccessDenied` for a rejected claim, which looked like a possible mismatch for the new trust-policy-denial path. Went further and found real-world reports (troubleshooting threads for GitHub Actions/Prefect/Keycloak OIDC-to-AWS setups) consistently citing the literal message "Not authorized to perform sts:AssumeRoleWithWebIdentity" for exactly this failure mode — the same message this PR's `accessDenied()` produces, and consistent with the pre-existing (unmodified) `AssumeRole` trust-policy-denial path in this same file. `IDPRejectedClaim`'s description ("the IdP reported that authentication failed... has expired or been explicitly revoked") reads as a different, IdP-communication-layer scenario that doesn't apply to Floci's self-hosted-issuer design anyway. So `AccessDenied` looks correct here despite the terse API reference not listing it — not a finding, just documenting the check.

`EksService.backfillOidcIdentities()`'s use of `config.defaultRegion()` for clusters that never had an issuer matches `createCluster()`'s own pre-existing use of the same call for a brand-new cluster's issuer/ARN — the EKS model has no cluster-specific region concept to diverge from, so this isn't a new inconsistency.

`WebIdentityTrustPolicyEvaluator`'s condition operators cover `StringEquals`/`StringNotEquals`/`StringLike`/`StringNotLike` but not the `*IfExists` variants; canonical eksctl/Terraform-generated IRSA trust policies use the plain operators for `sub`/`aud`, so this is a narrow, low-impact gap consistent with the class's own documented "fails closed" scoping rather than an oversight — not blocking.

@mcarvalho-singlestore  the only blocking part, is that it has conflicts.

mcarvalho-singlestore:
@pgermosen
Thanks for the review!

I've since rebased onto `main` and force-pushed, fixing the conflicts.

On your two non-blocking notes:
- `AccessDenied`: agreed. Deciding factor was consistency: the pre-existing `AssumeRole` denial path in this same file already returns it with that message shape.
- `*IfExists`: deliberate. It can't be an alias of `StringEquals` (it's satisfied when the key is absent), so doing it right needs an absent-vs-present distinction in the claim context. Falls into `default` meaning false today, so it fails closed. Happy to add it here or as a follow-up.

pgermosen:
Looks good

That fix closes a real gap the first two rounds didn't touch: `findVerificationKey` was resolving through the request-scoped, account-filtered `scan()`, so a cluster owned by a non-default account was invisible to STS — and since an unresolved issuer is treated as a third-party provider, `AssumeRoleWithWebIdentity` would have accepted such a token with *no* signature, claim, or trust-policy check at all. Scanning across accounts is the right fix, since an OIDC issuer URL is globally unique and the caller assuming the role need not be the cluster's owner. The parallel backfill bug (writing under the default account instead of the cluster's own, via `putForAccount`) is fixed the same way, and `deleteKey`/`findKeyByCluster` are correctly left account-scoped since both are only reachable from request-scoped, same-account-caller paths (`DeleteCluster`, the JWKS route).

pgermosen (approved):
Looks good

That fix closes a real gap the first two rounds didn't touch: `findVerificationKey` was resolving through the request-scoped, account-filtered `scan()`, so a cluster owned by a non-default account was invisible to STS — and since an unresolved issuer is treated as a third-party provider, `AssumeRoleWithWebIdentity` would have accepted such a token with *no* signature, claim, or trust-policy check at all. Scanning across accounts is the right fix, since an OIDC issuer URL is globally unique and the caller assuming the role need not be the cluster's owner. The parallel backfill bug (writing under the default account instead of the cluster's own, via `putForAccount`) is fixed the same way, and `deleteKey`/`findKeyByCluster` are correctly left account-scoped since both are only reachable from request-scoped, same-account-caller paths (`DeleteCluster`, the JWKS route).

PR #2109, opened 2026-08-03
feat(elbv2): honor preserve_host_header parameter

jvdsande:
## Summary

Hello!
This PR updates the ELBv2 service to honor the `preserve_host_header` parameter.

It propagates the load balancer's ARN through the ElbV2DataPlane instance in order to retrieve the current value of `routing.http.preserve_host_header.enabled` in `executeForward`/`proxyRequest` 

## Type of change

- [ ] Bug fix (`fix:`)
- [X] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Change tested with a simple echo server running on ECS behind an ALB.

Stack deployed through terraform with hashicorp/aws 6.57.1.

Without the patch/with `preserve_host_header=false`:

```
curl \                                               
  -H 'Host: example.com' \
  http://localhost:8088/
{"host":"172.17.0.3:8080"}
```

With the patch + `preserve_host_header=true`:

```
curl \
  -H 'Host: example.com' \
  http://localhost:8088/
{"host":"example.com"}
```

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/elbv2/ElbV2DataPlane.java
+++ b/src/main/java/io/github/hectorvent/floci/services/elbv2/ElbV2DataPlane.java
@@ -47,6 +47,7 @@ public class ElbV2DataPlane {
     private static final List<String> HOP_BY_HOP_HEADERS = List.of(
             "connection", "keep-alive", "transfer-encoding", "upgrade", "te", "trailers", "proxy-authorization", "proxy-authenticate"
     );
+    private static final String PRESERVE_HOST_HEADER_ATTRIBUTE = "routing.http.preserve_host_header.enabled";
 
     @Inject
     Vertx vertx;
@@ -78,7 +79,7 @@ public class ElbV2DataPlane {
 
     private HttpClient proxyClient;
 
-    private record ListenerBinding(int port, String host) {}
+    private record ListenerBinding(int port, String host, String loadBalancerArn) {}
 
     @PostConstruct
     void init() {
@@ -204,7 +205,7 @@ public class ElbV2DataPlane {
     private ListenerBinding binding(Listener listener, String region) {
         LoadBalancer loadBalancer = elbV2Service.getLoadBalancer(region, listener.getLoadBalancerArn());
         String host = loadBalancer != null ? normalizeHost(loadBalancer.getDnsName()) : listener.getLoadBalancerArn();
-        return new ListenerBinding(listener.getPort(), host);
+        return new ListenerBinding(listener.getPort(), host, listener.getLoadBalancerArn());
     }
 
     private void handleRequest(io.vertx.core.http.HttpServerRequest req, int port) {
@@ -226,7 +227,7 @@ public class ElbV2DataPlane {
         List<CompiledRule> chain = ref.get();
         for (CompiledRule compiled : chain) {
             if (compiled.matches(req)) {
-                executeAction(req, compiled.action, region);
+                executeAction(req, compiled.action, region, listenerArn);
                 return;
             }
         }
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/hectorvent/floci/services/elbv2/ElbV2DataPlane.java
```diff
@@ -47,6 +47,7 @@ public class ElbV2DataPlane {
     private static final List<String> HOP_BY_HOP_HEADERS = List.of(
             "connection", "keep-alive", "transfer-encoding", "upgrade", "te", "trailers", "proxy-authorization", "proxy-authenticate"
     );
+    private static final String PRESERVE_HOST_HEADER_ATTRIBUTE = "routing.http.preserve_host_header.enabled";
```
pgermosen, line 50: Flagging here since `docs/services/elb.md` isn't part of this diff so I can't comment there directly: this is a genuine behavioral change (the attribute actually alters what gets proxied, unlike most load-balancer attributes which are presumably just stored and echoed back), and it isn't mentioned anywhere in `docs/services/elb.md`'s "Behavior Notes" section. That section already exists and lists other behaviorally-significant specifics (default rule immutability, `DeleteTargetGroup`'s `ResourceInUse` guard, etc.) — this attribute fits the same pattern and would be a one-line addition. Not blocking, just easy to miss since nothing enforces it (there's no `floci:actions`-style generator for this prose section).

pgermosen (commented):
Welcome, and thanks for the PR!

- **`routing.http.preserve_host_header.enabled` is real and precisely described**: confirmed against the [Application Load Balancers guide](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#load-balancer-attributes) — exact attribute name, valid values (`true`/`false`), and default (`false`) all match what's implemented.
- **The mechanism is correct, not just superficially plausible**: traced `proxyRequest` end to end. The original request's headers (including `Host`) are already copied onto the outbound request by the pre-existing `req.headers().forEach(...)` loop (`Host` isn't in `HOP_BY_HOP_HEADERS`), so skipping the subsequent `clientReq.putHeader("Host", host + ":" + port)` override when `preserveHostHeader` is true is sufficient on its own — it doesn't need to actively re-inject anything. `preserveHostHeader()`'s null-handling (missing binding, missing load balancer, missing attributes map, unset attribute) all correctly fall through to `false`, matching AWS's documented default at every point.
- **The single test is more thorough than it looks**: only one `@Test` method, but it exercises both states sequentially against a real Vert.x HTTP backend and a real listener socket (`floci.services.elbv2.mock=false` forces the actual proxying path, not a mock) — first asserting the default (`false`) forwards the backend's own `host:port`, then enabling the attribute via a real `ModifyLoadBalancerAttributes` call and asserting the client's original `Host` header now passes through unchanged. That's a genuine round-trip through the real wire protocol, not just unit-level assertions.


@jvdsande 
One gap, worth closing before merge: this repo's `docs/services/*.md` files consistently document new AWS-compatible behavior — I've seen it done for comparable changes in several other recent PRs (ECS deployments, Secrets Manager selectors, IAM service-linked roles, S3 SSE-KMS) — usually in a "Behavior Notes" or "known deviations" section. `docs/services/elb.md` already has exactly such a section (currently ending with the `DescribeAccountLimits` bullet), and `ModifyLoadBalancerAttributes`/`DescribeLoadBalancerAttributes` are already listed as generic actions there, but nothing says this specific attribute key is one of the (presumably few) ones that actually changes runtime behavior rather than being stored-and-echoed inertly. Worth a one-line addition, e.g. "`routing.http.preserve_host_header.enabled` (default `false`) controls whether the original client `Host` header is forwarded to targets unchanged, or replaced with the target's `host:port`."

src/main/java/io/github/hectorvent/floci/services/elbv2/ElbV2DataPlane.java
jvdsande: Thanks for the review!

I've added the suggested line in `docs/services/elb.md`. I amended my previous commit rather than adding a stray `docs` or `chore` one just for this, and rebased against `main`.

pgermosen: Confirmed — that's the exact line, verbatim, in the "Behavior Notes" section right where it belongs. I also checked that the amend didn't touch anything else: the diff for `ElbV2DataPlane.java` is byte-for-byte identical to what I reviewed before (same blob hashes), so no re-review needed there. Re-ran the full `ElbV2*Test` suite after the rebase onto current `main`: 84/84, no regressions. `regen_action_docs.py --strict` still exits clean.

Nothing outstanding from me — good to merge.

pgermosen (approved):
Looks good

PR #476, opened 2026-04-16
fix(sqs): honor queue-level DelaySeconds on FIFO queues

tmchow:
## Summary

FIFO `sendMessage` silently dropped the queue-level `DelaySeconds` attribute: messages were stored and became immediately receivable regardless of the queue's configured default. Also reproduces via `SendMessageBatch` on FIFO queues.

Resolve the effective delay from the `DelaySeconds` queue attribute before the FIFO/standard split, then apply it via `Message.setVisibleAt` in both branches. FIFO always uses the queue-level value -- AWS SQS doesn't accept per-message `DelaySeconds` on FIFO queues -- while the standard path keeps its existing per-message behavior.

## What changed

- `SqsService.sendMessage` now computes `effectiveDelaySeconds` once before the FIFO validation block and applies it to FIFO messages via `setVisibleAt`. Added `parseDelaySecondsAttribute` helper that tolerates null/empty/non-numeric/negative attribute values.
- `SqsServiceTest`: added `queueLevelDelaySecondsAppliesToFifoQueue` (regression test for the reported bug) and `fifoQueueIgnoresPerMessageDelaySeconds` (AWS-compatibility guard).

## Scope notes

A few things I intentionally left out of this patch because they require wider changes than the reported bug:

- **Queue-level default on standard queues.** Currently the handler can't distinguish "caller omitted `DelaySeconds`" from "caller sent `DelaySeconds=0`" -- both arrive as `int 0`. Applying the queue-level default on the standard path needs a sentinel (or `Integer`) through the handler layer, and touches `SnsService`, `S3Service`, and `EventBridgeInvoker` which also pass `0` verbatim. Happy to pick that up in a follow-up if you want the full AWS parity; flagging so the maintainer can confirm the direction before a bigger refactor lands.
- **Persistence across restarts.** `Message.visibleAt` is `@JsonIgnore`, so delayed visibility doesn't survive restart. This is pre-existing for standard queues and this patch just inherits it for FIFO. Noted for a separate PR.

## Test plan

```
./mvnw test -Dtest=SqsServiceTest
```

Reproduction from the issue confirms the queue-level default now applies on FIFO.

Fixes #475
[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

PR #2112, opened 2026-08-04
fix(sqs): delayed FIFO message no longer blocks its message group

### LINKED ISSUE #2104: [BUG] SQS FIFO: a delayed message blocks its message group, so already-visible messages are not returned
## Service

SQS (FIFO queues)

## AWS API Action

`ReceiveMessage` — combined with `CreateQueue` (`DelaySeconds` > 0) and `SendMessage` (`MessageGroupId`)

## Expected behavior

On real AWS SQS, only **in-flight** messages (received, still inside their visibility timeout) block
the rest of their message group:

> No additional messages from the same message group ID are returned until the first message is deleted
> or becomes visible again.
> — [FIFO queue logic](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues-understanding-logic.html)

Messages that are still inside their `DelaySeconds` window do not block anything. In the reproduction
below, the final `ReceiveMessage` returns `A` on real AWS SQS: `A` is visible, and nothing in group `G`
is in flight — the newly sent, still-delayed `B` is irrelevant to it.

## Actual behavior

floci returns an empty response. Sending `B` retroactively made the already-visible `A` undeliverable.
Right before `B` is sent, `GetQueueAttributes` reports

```
visible=1  notVisible=0  delayed=0
```

so `A` is plainly deliverable; the only change is the arrival of a delayed message in the same group.

This starves consumers permanently rather than delaying them: whenever a producer sends to a message
group faster than `DelaySeconds`, the group always contains at least one delayed message, so nothing is
ever delivered. The client keeps polling successfully, always gets an empty response, and the backlog
grows. Delivery resumes only if the producer stops long enough for every delayed message in the group
to become visible.

## Reproduction

Send one message, let its delay elapse, then send a second one:

```sh
QUEUE_URL=$(aws --endpoint-url http://localhost:4566 sqs create-queue \
  --queue-name delay-block.fifo \
  --attributes FifoQueue=true,DelaySeconds=10,VisibilityTimeout=30,ContentBasedDeduplication=false \

aws --endpoint-url http://localhost:4566 sqs send-message --queue-url "$QUEUE_URL" \
  --message-body A --message-group-id G --message-deduplication-id a1

sleep 12

aws --endpoint-url http://localhost:4566 sqs send-message --queue-url "$QUEUE_URL" \
  --message-body B --message-group-id G --message-deduplication-id b1

aws --endpoint-url http://localhost:4566 sqs receive-message --queue-url "$QUEUE_URL"
```

The `sleep` lets `A`'s 10 second delay elapse, so `A` is visible when `B` is sent.

On floci the last command prints nothing at all. Running the same commands against real AWS SQS
(with `--endpoint-url` dropped) returns `A`:

```json
{
    "Messages": [
        {
            "MessageId": "a9eca5c7-e2d7-4c11-96ad-6bc5e975c060",
            "ReceiptHandle": "...",
            "MD5OfBody": "7fc56270e7a70fa81a5935b72eacbe29",
            "Body": "A"
        }
    ]
}
```

## Environment

- floci **1.5.34** (`floci/floci:latest`, Docker). Also reproduced on 1.5.5
- AWS CLI v2; compared against real AWS SQS in `ap-northeast-1`
- Single message group, `ContentBasedDeduplication=false`, explicit deduplication ids, `WaitTimeSeconds=0`

## Probable cause

`GuardedMessageQueue.claimFifo()` states the intended semantics in its own comment, but the predicate
below it is wider than the comment:

```java
// Cross-call group locking: a group that already has an in-flight
// message from a previous ReceiveMessage call is blocked until that
// message is deleted or its visibility expires. ...
Set<String> groupsWithInFlight =
        messages.stream().filter(msg -> !msg.isVisible() && msg.getMessageGroupId() != null)
                .map(Message::getMessageGroupId).collect(Collectors.toSet());
```

`!msg.isVisible()` is a timestamp comparison (`visibleAt > now`). Since #476 made FIFO queues honor the
queue-level delay by setting `message.setVisibleAt(Instant.now().plusSeconds(effectiveDelaySeconds))`,
a *delayed* message is also `!isVisible()`, so it lands in `groupsWithInFlight` and locks the group —
even though it was never received, which is what the comment says the lock is for.

The same class already distinguishes the two cases, in `messageCounts()` (added in #1285):

```java
if (m.isVisible()) {
    visible++;
} else if (m.getReceiptHandle() == null) {
    delayed++;
} else {
    inFlight++;
}
```

> A message that is not visible and was never claimed — no receipt handle — can only be waiting out its
> DelaySeconds, so it counts as delayed rather than in flight.

Applying that same discriminator in `claimFifo()` — requiring `msg.getReceiptHandle() != null` for a
message to lock its group — would make the predicate match both the comment and the AWS behaviour.

Related: #475 / #476 (FIFO delay support), #1281 / #1285 (delayed vs in-flight in `GetQueueAttributes`).

leepokai:
## Summary

Closes #2104.

`GuardedMessageQueue.claimFifo()` builds its group-lock set with `!msg.isVisible()`, which matches not only in-flight messages but also messages still waiting out `DelaySeconds` (since #476 sets `visibleAt = now + delay` on send). A newly sent, still-delayed message therefore locked its message group and made an already-visible message in the same group unreceivable — permanently starving the group whenever a producer sends faster than `DelaySeconds`.

The fix narrows the predicate with `msg.getReceiptHandle() != null` — the same delayed-vs-in-flight discriminator `messageCounts()` already uses (#1285): a not-visible message that was never claimed can only be waiting out its delay.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

On real AWS SQS, only in-flight messages block the rest of their FIFO message group ("No additional messages from the same message group ID are returned until the first message is deleted or becomes visible again"); delayed messages block nothing. Floci returned an empty `ReceiveMessage` response where real AWS returns the visible message.

Verified with the exact reproduction from #2104 (AWS CLI 2.32.31 against `quarkus:dev`): after the fix the final `receive-message` returns `A` (MD5 `7fc56270e7a70fa81a5935b72eacbe29`) while `B` is still delayed, and `GetQueueAttributes` reports `visible=1 notVisible=0 delayed=1` right before the receive. A follow-up `receive-message` while `A` is in flight returns empty, confirming the cross-call group lock still works.

Regression tests added at all three levels: `GuardedMessageQueueTest` (lock predicate), `SqsServiceTest` (delay elapse through `SqsService.receiveMessage`), and `SqsFifoIntegrationTest` (Query protocol end-to-end). All three fail without the one-line fix.

## Checklist

Note on the full local run: all SQS suites pass (8200 tests total); the only failures in my environment are Docker-backed suites (DocDB, Neptune, ElastiCache, API Gateway authorizers, ELB) with `SocketTimeout`/port-bind errors, which fail identically on unmodified `main` on this machine (macOS + Docker), so they are unrelated to this change.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueue.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueue.java
@@ -120,9 +120,13 @@ class GuardedMessageQueue {
         // message from a previous ReceiveMessage call is blocked until that
         // message is deleted or its visibility expires. Within a single call
         // we may return multiple messages from the same group (preserving
-        // insertion order), up to MaxNumberOfMessages.
+        // insertion order), up to MaxNumberOfMessages. A not-visible message
+        // that was never claimed — no receipt handle — is only waiting out its
+        // DelaySeconds and must not lock its group.
         Set<String> groupsWithInFlight =
-                messages.stream().filter(msg -> !msg.isVisible() && msg.getMessageGroupId() != null)
+                messages.stream()
+                        .filter(msg -> !msg.isVisible() && msg.getReceiptHandle() != null
+                                && msg.getMessageGroupId() != null)
                         .map(Message::getMessageGroupId).collect(Collectors.toSet());
 
         for (Message msg : messages) {
--- a/src/test/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueueTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/sqs/GuardedMessageQueueTest.java
@@ -218,6 +218,27 @@ class GuardedMessageQueueTest {
         assertTrue(second.claimed().isEmpty());
     }
 
+    @Test
+    void fifoDelayedMessageDoesNotLockItsMessageGroup() {
+        // Only in-flight messages (claimed, within their visibility timeout)
+        // lock a FIFO message group. A message still waiting out its
+        // DelaySeconds was never claimed — no receipt handle — and must not
+        // make an already-visible message in the same group unreceivable.
+        Message visible = new Message("visible");
+        visible.setMessageGroupId("group1");
+        Message delayed = new Message("delayed");
+        delayed.setMessageGroupId("group1");
+        delayed.setVisibleAt(Instant.now().plusSeconds(60));
+
+        queue.addMessage(visible);
+        queue.addMessage(delayed);
+
+        var result = queue.claimVisibleMessages(10, 30, true, -1, null);
+        assertEquals(1, result.claimed().size(),
+                "A delayed message must not block its message group");
+        assertEquals("visible", result.claimed().get(0).getBody());
+    }
+
     // --- DLQ ---
 
     @Test
--- a/src/test/java/io/github/hectorvent/floci/services/sqs/SqsFifoIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/sqs/SqsFifoIntegrationTest.java
@@ -153,6 +153,71 @@ class SqsFifoIntegrationTest {
 
     @Test
     @Order(8)
+    void delayedMessageDoesNotBlockVisibleMessageInSameGroup() throws InterruptedException {
+        // Issue #2104: a message still waiting out DelaySeconds must not lock
+        // its message group — only in-flight messages do.
+        String delayQueueUrl = given()
+            .contentType("application/x-www-form-urlencoded")
+            .formParam("Action", "CreateQueue")
+            .formParam("QueueName", "fifo-delay-lock.fifo")
+            .formParam("Attribute.1.Name", "FifoQueue")
+            .formParam("Attribute.1.Value", "true")
+            .formParam("Attribute.2.Name", "DelaySeconds")
+            .formParam("Attribute.2.Value", "1")
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .extract().xmlPath().getString("CreateQueueResponse.CreateQueueResult.QueueUrl");
+
+        given()
+            .contentType("application/x-www-form-urlencoded")
+            .formParam("Action", "SendMessage")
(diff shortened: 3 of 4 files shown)


pgermosen (approved):
Looks good

- **Root cause matches exactly what the issue diagnosed**: `claimFifo()`'s group-lock predicate used `!msg.isVisible()` alone, which is true both for genuinely in-flight messages and for messages still waiting out `DelaySeconds` (set via `visibleAt = now + delay` since #476) — so a newly-sent delayed message could lock a group and starve an already-visible message in it.
- **The fix reuses an existing, already-correct discriminator rather than inventing a new one**: confirmed `messageCounts()` in the same file (added in #1285) already distinguishes delayed-vs-in-flight with the identical `getReceiptHandle() == null` check. The fix in `claimFifo()` applies the same logic (`getReceiptHandle() != null` required to lock a group), so this isn't a new invariant, it's making one part of the class consistent with another that already got this right.
- **Checked the AWS-behavior claim against two separate AWS doc pages, not just the PR's quote of one of them**: the [FIFO queue logic guide](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues-understanding-logic.html) has "No additional messages from the same message group ID are returned until the first message is deleted or becomes visible again" under a section titled "Handling visibility timeout," scoped explicitly to "When a message is retrieved but not deleted" — i.e. the lock is about *received* messages, not merely delayed ones. The [delay queues guide](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-delay-queues.html) makes the distinction explicit: "the difference [from visibility timeouts] is that, for delay queues, a message is hidden when it is first added to queue, whereas for visibility timeouts a message is hidden only after it is consumed." That's precisely the delayed-vs-in-flight distinction this fix restores — confirmed from AWS's own docs independently, not just trusted from the PR description.

PR #2114, opened 2026-08-04
fix(dynamodb): order Query by full composite sort key

### LINKED ISSUE #1675: [BUG] DynamoDB Query ScanIndexForward=false still returns ascending results on GSI / enhanced-client sortBetween
```markdown
## Service

DynamoDB

## AWS API Action

Query

## Expected behavior

When querying a table/index with `ScanIndexForward=false`, DynamoDB should return matching items in descending sort-key order.

AWS documents this behavior for `Query`: when `ScanIndexForward` is false, DynamoDB reads results in reverse order by sort key value and returns them to the client.

## Actual behavior

Floci returns results in ascending sort-key order even when `ScanIndexForward=false`.

Observed with a GSI query where the sort key contains request state + `createdAt`. The query is constrained to one request state, so result ordering should be determined by `createdAt`.

Example returned order while descending was requested:

```text
2026-07-01T09:04:25.518Z
2026-07-01T09:04:56.205Z
```

This is ascending. The same test passes against DynamoDB Local / LocalStack-style DynamoDB behavior.

## Reproduction

Java AWS SDK v2 enhanced client query:

```java
QueryEnhancedRequest request =
    QueryEnhancedRequest.builder()
        .queryConditional(QueryConditional.sortBetween(startPoint, endPoint))
        .scanIndexForward(false)
        .limit(100)
        .build();

List<Item> items = index.query(request)
    .iterator()
    .next()
    .items();
```

Expected: newest/highest sort key first.

Actual on Floci: oldest/lowest sort key first.

## Environment

- Floci version / image tag: `floci/floci:1.5.29-compat`
- Java SDK version: AWS SDK v2 `2.41.34`
- How running Floci: Docker Compose
- Enabled services: `sns,dynamodb,sqs,s3`
```

AWS reference: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html

Possibly related to closed issue #143 and PR #160. This appears either to be a regression or an uncovered query shape, because `ScanIndexForward=false` still returns ascending results on `floci/floci:1.5.29-compat`.

ngencic:
## Summary

Query ordered results using only the first RANGE attribute returned by
`getSortKeyName()`. For an index with a composite sort key (more than one RANGE
attribute, e.g. `state RANGE, createdAt RANGE`), the remaining components were
ignored: when the first component tied across rows, the stable sort left items
in base-table storage order and `ScanIndexForward=false` merely reversed that
order, so the secondary component (e.g. `createdAt`) never drove ordering.

Sort by the full sort key instead — comparing each RANGE attribute in
key-schema order, short-circuiting on the first non-equal component, then
reversing for `ScanIndexForward=false`. Single-sort-key indexes are unaffected
(a one-element list preserves previous behavior).

Adds `getSortKeyNames()` to `GlobalSecondaryIndex`, `LocalSecondaryIndex` and
`TableDefinition`.

Closes #1675

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: a `Query` against an index with a multi-attribute composite
sort key returned items that were not ordered by the full key. Ties on the first
sort-key attribute fell back to storage order, and `ScanIndexForward=false` only
reversed that — so results diverged from DynamoDB, which orders by the entire
composite sort key. DynamoDB natively supports composite sort keys on secondary
indexes, so emulating full-key ordering is required for parity.

Verified with a wire-protocol (`DynamoDB_20120810.Query`) integration test: a GSI
keyed `[memberName HASH, state RANGE, createdAt RANGE]`, items whose storage
order deliberately disagrees with `createdAt`, asserting ascending order and the
exact reverse for `ScanIndexForward=false`.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -753,13 +753,13 @@ public class DynamoDbService {
 
             String startItemKey = hasTableKeys
                     ? buildItemKeyFromNode(exclusiveStartKey, tablePkName, tableSkName)
-                    : buildItemKeyFromNode(exclusiveStartKey, pkName, skName);
+                    : buildItemKeyFromNode(exclusiveStartKey, pkName, sortKeyNames);
 
             int startIdx = -1;
             for (int i = 0; i < results.size(); i++) {
                 String thisKey = hasTableKeys
                         ? buildItemKeyFromNode(results.get(i), tablePkName, tableSkName)
-                        : buildItemKeyFromNode(results.get(i), pkName, skName);
+                        : buildItemKeyFromNode(results.get(i), pkName, sortKeyNames);
                 if (thisKey.equals(startItemKey)) {
                     startIdx = i;
                     break;
@@ -776,7 +776,7 @@ public class DynamoDbService {
         // Apply Limit (stops at N items)
         if (limit != null && limit > 0 && evaluatedItems.size() > limit) {
             JsonNode lastItem = evaluatedItems.get(limit - 1);
-            lastEvaluatedKey = buildKeyNode(table, lastItem, pkName, skName, indexName != null);
+            lastEvaluatedKey = buildKeyNode(table, lastItem, pkName, sortKeyNames, indexName != null);
             evaluatedItems = new ArrayList<>(evaluatedItems.subList(0, limit));
         }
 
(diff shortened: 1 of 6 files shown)


hectorvent:
Thank you for this, and for fixing the cursor path along with the visible ordering symptom, the pagination collapse for rows sharing the first range value was the subtler half of the bug.

One context note rather than a change request: a multi range key schema is a Floci extension, real DynamoDB caps KeySchema at one partition plus one sort key, so this improves behavior on a surface AWS itself would reject at CreateTable. Since the single sort key path is preserved by construction and the extension already exists, that is fine here, it is just worth both of us knowing where the deviation line sits.

**(blocking)** Commit cf8bfbc carries an AI co-author trailer; this repo keeps commit attribution with human contributors (see AGENTS.md). Would you amend that commit and force push so the attribution is yours alone?

That attribution fix is the only blocker, the code itself is ready.

hectorvent (approved):
All look good. Thanks @ngencic

PR #2115, opened 2026-08-04
fix(s3): apply CreateBucketConfiguration tags on bucket creation

arindamchoudhury:
## Summary

`CreateBucket` accepted a `<Tags>` array in the request body but ignored it. The body was parsed for `LocationConstraint` only, so tags supplied at creation were silently dropped: the request returned 200, and a following `GetBucketTagging` / S3 Control `ListTagsForResource` reported an empty tag set.

This parses the same `<Tag>` key/value pairs `handlePutBucketTagging` already parses, and routes them through the existing `s3Service.putBucketTagging` once the bucket exists. No new imports, no new methods, no signature changes.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behaviour:** tags passed in `CreateBucketConfiguration` were discarded.

AWS applies them. The [CreateBucket API reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html) shows `Tags` in the request syntax and documents it as *"An array of tags that you can apply to the bucket that you're creating"*, gated on the `s3:TagResource` permission.

**How this surfaces in practice.** The AWS Terraform provider (6.57.1) sends bucket tags this way rather than via a separate `PutBucketTagging` call — from `TF_LOG=trace`:

```xml
<CreateBucketConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Tags><Tag><Key>owner</Key><Value>data-team</Value></Tag></Tags></CreateBucketConfiguration>
```

Against Floci, an `aws_s3_bucket` with a `tags` map applies cleanly but lands with `tags = {}` in state, so **every subsequent plan proposes to add them again** — a diff that never converges. The provider reads them back through S3 Control `ListTagsForResource`, which Floci implements; only the write path at creation was missing.

**Verified end to end**, Terraform 1.15.8 + AWS provider 6.57.1, same config and provider against both images:

| | state `tags` | plan right after apply | `get-bucket-tagging` |
|---|---|---|---|
| 1.5.34 | `{}` | `+ "owner" = "data-team"`, `1 to change` | empty `TagSet` |
| this branch | `"owner" = "data-team"` | **No changes** | `TagSet` with the tag |

## Checklist

- [x] `./mvnw test` passes locally — see the note below

### Note on the test run

`S3IntegrationTest` (125 tests) passes on Linux:

```
[INFO] Tests run: 125, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Run in `eclipse-temurin:25-jdk`. I ran that class rather than the whole suite.

On **Windows** the same class reports 1 failure, `copyObjectWithQuestionMarkInSourceKeySucceeds` — the setup `PUT` of an object key containing a percent-encoded `?` is rejected with `InvalidKey`, expected 200 got 400. That is **unrelated to this change and pre-existing**: it fails identically in a clean worktree at `9bbe9ffe` (this branch's parent, none of these changes present), and it passes on Linux both with and without this change. Looks like a platform difference in URI decoding before key validation. Happy to open a separate issue if that is useful.

### One design question for a maintainer

`createBucket` treats a repeat create in `us-east-1` as idempotent. As written, this change would then **overwrite** existing tags on that path. AWS documents the re-create case as resetting ACLs and says nothing about tags, so skipping them when the bucket already existed is an equally defensible reading. Happy to change it.

Disclosure: this was written with Claude Code. I've reviewed the change and it's small — the parse and the store call both reuse what `handlePutBucketTagging` already does.

What's actually verified rather than asserted:

- `S3IntegrationTest`, 125 tests, 0 failures on `eclipse-temurin:25-jdk`.
- The before/after table is two real `terraform apply` runs against `floci/floci:1.5.34` and a locally-built image from this branch, same config and provider version.
- The pre-existing Windows failure was confirmed by running that single test in a clean worktree at `9bbe9ffe`.

The idempotent-create question above is the part I'd most like a maintainer's opinion on — I don't think the AWS docs settle it.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -331,6 +331,15 @@ public class S3Controller {
             }
             String region = locationConstraint != null ? locationConstraint : regionResolver.resolveRegion(httpHeaders);
             s3Service.createBucket(bucket, region);
+            // CreateBucketConfiguration may carry a <Tags> array; AWS applies those tags to the
+            // new bucket, so a follow-up GetBucketTagging / ListTagsForResource must return them.
+            if (body != null && body.length > 0) {
+                Map<String, String> creationTags = XmlParser.extractPairs(
+                        new String(body, StandardCharsets.UTF_8), "Tag", "Key", "Value");
+                if (!creationTags.isEmpty()) {
+                    s3Service.putBucketTagging(bucket, creationTags);
+                }
+            }
             String lockEnabled = httpHeaders.getHeaderString("x-amz-bucket-object-lock-enabled");
             if ("true".equalsIgnoreCase(lockEnabled)) {
                 s3Service.putBucketVersioning(bucket, "Enabled");
--- a/src/test/java/io/github/hectorvent/floci/services/s3/S3IntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/s3/S3IntegrationTest.java
@@ -362,6 +362,44 @@ class S3IntegrationTest {
     }
 
     @Test
+    @Order(21)
+    void createBucketAppliesTagsFromCreateBucketConfiguration() {
+        String bucket = "tagged-at-create-bucket";
+        String createBucketConfiguration = """
+                <CreateBucketConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+                    <Tags>
+                        <Tag>
+                            <Key>owner</Key>
+                            <Value>data-team</Value>
+                        </Tag>
+                    </Tags>
+                </CreateBucketConfiguration>
+                """;
+
+        given()
+            .contentType("application/xml")
+            .body(createBucketConfiguration)
+        .when()
+            .put("/" + bucket)
+        .then()
+            .statusCode(200);
+
+        given()
+        .when()
+            .get("/" + bucket + "?tagging")
+        .then()
+            .statusCode(200)
+            .body(containsString("owner"))
+            .body(containsString("data-team"));
+
+        given()
+        .when()
+            .delete("/" + bucket)
+        .then()
+            .statusCode(204);
+    }
+
+    @Test
     @Order(20)
     void headBucketReturnsStoredRegionForLocationConstraintBucket() {
         String bucket = "eu-head-bucket";


hectorvent:
Reviewed against botocore's [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/s3/2006-03-01/service-2.json). The model confirms `CreateBucketConfiguration.Tags` as a `TagSet` list of `Tag {Key, Value}` with list `locationName: Tag`, so the wrapped `<Tags><Tag>` shape you parse is exactly what the model declares, and it matches your Terraform trace. Reusing `XmlParser.extractPairs` and `s3Service.putBucketTagging` keeps this on the same path as `PutBucketTagging`. Looks good.

On your idempotent re-create question, the reference implementations settle it: the idempotent us-east-1 path only applies when the request carries no tags. A repeat create that includes tags raises `BucketAlreadyOwnedByYou` even in us-east-1, per LocalStack's S3 provider ([provider.py](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/services/s3/provider.py)):

```python
if bucket_region != AWS_REGION_US_EAST_1 or bucket_tags:
    raise BucketAlreadyOwnedByYou(
        "Your previous request to create the named bucket succeeded and you already own it.",
        BucketName=bucket_name,
    )
```

So the answer is neither overwrite nor skip. `BucketAlreadyOwnedByYou` is one of the two errors the model declares for `CreateBucket`, so SDK clients map it to a typed exception. Given how narrow the path is, fine as a follow-up rather than a change to this PR.

Two non-blocking notes. AWS validates creation tags before creating the bucket (duplicate keys, `aws:` prefix, key and value patterns, all surfacing as `InvalidTag`; see `validate_tag_set` with `type_set="create-bucket"` in LocalStack's [utils.py](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/services/s3/utils.py)). Floci's existing `PutBucketTagging` path does not validate either, so this PR is consistent with the codebase and that gap is pre-existing. Also the new test's `@Order(21)` duplicates an existing `@Order(21)` in `S3IntegrationTest`; harmless since the test creates and deletes its own bucket, but worth renumbering.

hectorvent (approved):

PR #2128, opened 2026-08-05
feat(ses): implement v1 receipt rule set actions (stored-but-inert)

### LINKED ISSUE #2123: [BUG] SES: CreateReceiptRuleSet / SetActiveReceiptRuleSet return UnsupportedOperation
## Service

SES

## AWS API Action

`CreateReceiptRuleSet`, `SetActiveReceiptRuleSet`

## Expected behavior

`CreateReceiptRuleSet` creates a named receipt rule set; `SetActiveReceiptRuleSet` marks one active. Terraform exposes these as `aws_ses_receipt_rule_set` and `aws_ses_active_receipt_rule_set`.

## Actual behavior

```
api error UnsupportedOperation: Operation CreateReceiptRuleSet is not supported by SES.
```

Other SES operations work; the receipt-rule-set family appears to be absent.

## Why this matters

Inbound-email configurations declare a rule set in account-level bootstrap Terraform. Because it fails at create, the entire module apply aborts and every downstream module is blocked — even though the rule set itself has no useful behaviour in a local emulator.

As with `CreateServiceLinkedRole`, a stored-but-inert implementation would be sufficient: create it, list it, describe it, delete it. No mail routing required.

## Reproduction

```bash

aws ses create-receipt-rule-set --rule-set-name local-rule-set
# -> UnsupportedOperation
```

Terraform equivalent:

```hcl
resource "aws_ses_receipt_rule_set" "this" {
  rule_set_name = "local-rule-set"
}

resource "aws_ses_active_receipt_rule_set" "this" {
  rule_set_name = aws_ses_receipt_rule_set.this.id
}
```

## Environment

`floci/floci:1.5.34`, linux/arm64, macOS host. Terraform 1.7.5, provider `hashicorp/aws` v5.86.1.

okinaka:
## Summary

Closes #2123.

Implements the SES v1 receipt-rule-set management actions (`CreateReceiptRuleSet`, `DescribeReceiptRuleSet`, `ListReceiptRuleSets`, `DeleteReceiptRuleSet`, `SetActiveReceiptRuleSet`, `DescribeActiveReceiptRuleSet`). Floci has no inbound-mail endpoint, so rule sets are stored inertly — a set never holds any receipt rules and routes no mail — but the management API round-trips so tools like Terraform (`aws_ses_receipt_rule_set`, `aws_ses_active_receipt_rule_set`) can declare a rule set during bootstrap without hitting `UnsupportedOperation`. Individual receipt rules and receipt filters are out of scope.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Wire behavior verified against real AWS classic SES (2010-12-01) via boto3 (us-west-2): `RuleSetDoesNotExist` (400) for a missing set on describe/set-active, `AlreadyExists` (400) for a duplicate create, idempotent `DeleteReceiptRuleSet` (deleting a non-existent set succeeds), and an empty `DescribeActiveReceiptRuleSet` result (no `Metadata`) when no set is active. Covered by a RestAssured integration test, an AWS SDK for Java v2 compatibility test, and Terraform/OpenTofu compat checks (`aws_ses_receipt_rule_set` + `aws_ses_active_receipt_rule_set`).

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesQueryHandler.java
@@ -11,6 +11,7 @@ import io.github.hectorvent.floci.services.ses.model.CloudWatchDimensionConfigur
 import io.github.hectorvent.floci.services.ses.model.ConfigurationSet;
 import io.github.hectorvent.floci.services.ses.model.DeliveryOptions;
 import io.github.hectorvent.floci.services.ses.model.EmailTemplate;
+import io.github.hectorvent.floci.services.ses.model.ReceiptRuleSet;
 import io.github.hectorvent.floci.services.ses.model.EventDestination;
 import io.github.hectorvent.floci.services.ses.model.Identity;
 import io.github.hectorvent.floci.services.ses.model.KinesisFirehoseDestination;
@@ -107,6 +108,12 @@ public class SesQueryHandler {
                         handleUpdateConfigurationSetReputationMetricsEnabled(params, region);
                 case "PutConfigurationSetDeliveryOptions" ->
                         handlePutConfigurationSetDeliveryOptions(params, region);
+                case "CreateReceiptRuleSet" -> handleCreateReceiptRuleSet(params, region);
+                case "DescribeReceiptRuleSet" -> handleDescribeReceiptRuleSet(params, region);
+                case "ListReceiptRuleSets" -> handleListReceiptRuleSets(region);
+                case "DeleteReceiptRuleSet" -> handleDeleteReceiptRuleSet(params, region);
+                case "SetActiveReceiptRuleSet" -> handleSetActiveReceiptRuleSet(params, region);
+                case "DescribeActiveReceiptRuleSet" -> handleDescribeActiveReceiptRuleSet(region);
                 default -> AwsQueryResponse.error("UnsupportedOperation",
                         "Operation " + action + " is not supported by SES.", AwsNamespaces.SES, 400);
             };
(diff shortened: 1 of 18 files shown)


src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
```diff
@@ -1283,6 +1293,132 @@ private static String configSetKey(String region, String name) {
...
+                rs.setActive(false);
+                receiptRuleSetStore.put(receiptRuleSetKey(region, rs.getName()), rs);
+            }
+        }
+    }
+
+    // Verified against AWS: RuleSetName must match ^[a-zA-Z0-9_.-]+$ (a character outside that set is
+    // a Smithy ValidationError), and must additionally be <= 64 chars and start/end with an
+    // alphanumeric (otherwise a service-level "Not a valid ruleSetName" InvalidParameterValue).
+    private static final Pattern RULE_SET_NAME_CHARS = Pattern.compile("^[a-zA-Z0-9_.-]+$");
```
pgermosen, line 1399: **The name rules aren't in any public model — they rest entirely on your probing**

Worth being explicit that this is the strongest part of the PR *and* its most fragile: I went to check these constraints and there's nothing to check them against. `ReceiptRuleSetName` in `botocore/data/ses/2010-12-01/service-2.json` is:

```json
{ "type": "string" }
```

No pattern. No min or max length. So `^[a-zA-Z0-9_.-]+$`, the 64-character cap, the alphanumeric-start/end rule, and the split between a Smithy `ValidationError` and a service-level `InvalidParameterValue` all come purely from your boto3 run against real SES. That's better evidence than the model, and I'd much rather have it than a guess.

The thing to weigh is **which direction the risk runs.** Floci now *rejects* names, so if the real constraint is even slightly wider than what you measured — 65 characters allowed, a leading underscore tolerated, a regional difference — a config that works against production AWS fails against Floci at create time. Being stricter than AWS is the worse failure mode for an emulator; being lenient only fails to catch a mistake the user would hit in production anyway.

**Suggested:** record *how* it was established so the next person can re-verify or challenge it rather than treating it as settled — something like *"probed against SES 2010-12-01 in us-west-2 via boto3, 2026-08"* in the comment above the pattern. One line, and it's the only way this stays checkable, because the model will never confirm it.

Minor knock-on: `deleteReceiptRuleSet` validates the name too, so if the constraint is too narrow, a delete that AWS would accept idempotently throws instead. That's self-consistent within Floci (create validates identically), so it's only reachable if the probed rules are wrong — the same risk, not a separate one.

src/main/java/io/github/hectorvent/floci/services/ses/SesQueryHandler.java
```diff
@@ -11,6 +11,7 @@
 import io.github.hectorvent.floci.services.ses.model.ConfigurationSet;
 import io.github.hectorvent.floci.services.ses.model.DeliveryOptions;
 import io.github.hectorvent.floci.services.ses.model.EmailTemplate;
+import io.github.hectorvent.floci.services.ses.model.ReceiptRuleSet;
```
pgermosen: **Misplaced import** *(nit)*

```java
import ...model.EmailTemplate;
import ...model.ReceiptRuleSet;   // <- between EmailTemplate and EventDestination
import ...model.EventDestination;
```

The rest of that block is strictly alphabetical, and the same import in `SesService.java:100` **is** correctly placed (after `MessageTag`, before `Topic`) — which is what makes this read as an accidental paste rather than a convention.

**Suggested:** move it after `MessageTag` to match.

pgermosen (approved):
Looks good.

This is a very complete PR for its size: service + handler + model, 13 integration cases, an AWS SDK v2 compatibility test, Terraform *and* OpenTofu compat checks, and docs updated in both the operation table and the limitations list. The "stored-but-inert" framing is stated honestly in the code, the model javadoc and the docs — a reader can't mistake this for working inbound mail.

**Every error shape matches the model exactly.** I parsed `botocore/data/ses/2010-12-01/service-2.json` rather than trusting the description:

| operation | model declares | this PR |
|---|---|---|
| `CreateReceiptRuleSet` | `AlreadyExists` (400) | ✅ |
| `DescribeReceiptRuleSet` | `RuleSetDoesNotExist` (400) | ✅ |
| `SetActiveReceiptRuleSet` | `RuleSetDoesNotExist` (400) | ✅ |
| `DeleteReceiptRuleSet` | `CannotDelete` (400) | ✅ *(active-set rejection)* |
| `ListReceiptRuleSets` / `DescribeActiveReceiptRuleSet` | *(no errors)* | ✅ |


---

Just a few nit comments

src/main/java/io/github/hectorvent/floci/services/ses/SesQueryHandler.java
okinaka: Fixed in `328e2ef6` — moved the `ReceiptRuleSet` import after `MessageTag` so the block is alphabetical again, matching how the same import is already ordered in `SesService`. Good catch, thanks @pgermosen.

src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
okinaka: Agreed on the direction of the risk — for an emulator, being *stricter* than AWS is the worse failure mode, so I've kept the rules to exactly what the probe measured rather than rounding them to anything wider or narrower. What was missing was the provenance, so I recorded it in `328e2ef6`: the comment now states that the botocore model (`service-2.json`, SES 2010-12-01) declares `ReceiptRuleSetName` as a bare `{"type": "string"}` with no pattern or length, that the constraints were established by probing live SES in us-west-2 via boto3 (2026-08), and that a future change should be re-verified against live SES rather than the model, which can't confirm it. That keeps it checkable for the next reader, as you suggested. Thanks @pgermosen.

src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
```diff
@@ -1283,6 +1293,105 @@ private static String configSetKey(String region, String name) {
...
+        return all;
+    }
+
+    public void deleteReceiptRuleSet(String name, String region) {
+        requireRuleSetName(name);
+        // AWS is idempotent here: deleting a non-existent rule set succeeds without error. Hold the
+        // lock so a concurrent set-active/clear (which scans and re-puts active sets) can't resurrect
+        // the rule set we just deleted.
+        synchronized (receiptRuleSetLock) {
+            receiptRuleSetStore.delete(receiptRuleSetKey(region, name));
```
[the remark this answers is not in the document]
okinaka: Both of these were addressed in `ae8e84b4` (pushed after the commit greptile reviewed): `deleteReceiptRuleSet` now rejects the active rule set with `CannotDelete`/400 instead of deleting it, and `setActiveReceiptRuleSet` / `describeActiveReceiptRuleSet` read and mutate active state under `receiptRuleSetLock`, so a reader can't observe the intermediate empty state during a replacement. pgermosen's review table also confirms the `CannotDelete` behavior.

src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
```diff
@@ -1283,6 +1293,105 @@ private static String configSetKey(String region, String name) {
...
+        // No RuleSetName clears the account's active rule set (matches AWS).
+        boolean clearOnly = name == null || name.isBlank();
+        synchronized (receiptRuleSetLock) {
+            if (!clearOnly) {
+                ReceiptRuleSet target = receiptRuleSetStore.get(receiptRuleSetKey(region, name))
+                        .orElseThrow(() -> ruleSetDoesNotExist(name));
+                clearActiveReceiptRuleSet(region);
+                target.setActive(true);
+                receiptRuleSetStore.put(receiptRuleSetKey(region, name), target);
+            } else {
```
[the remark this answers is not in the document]
okinaka: Both of these were addressed in `ae8e84b4` (pushed after the commit greptile reviewed): `deleteReceiptRuleSet` now rejects the active rule set with `CannotDelete`/400 instead of deleting it, and `setActiveReceiptRuleSet` / `describeActiveReceiptRuleSet` read and mutate active state under `receiptRuleSetLock`, so a reader can't observe the intermediate empty state during a replacement. pgermosen's review table also confirms the `CannotDelete` behavior.

PR #2129, opened 2026-08-05
fix(lambda): scope resource-policy statements to the request Qualifier

### LINKED ISSUE #2124: Lambda AddPermission/GetPolicy/RemovePermission ignore Qualifier — alias-scoped permissions never converge under Terraform
## Summary

`AddPermission`, `GetPolicy` and `RemovePermission` ignore the `Qualifier` query parameter. Every statement is stored against the **unqualified** function ARN, so a qualifier-scoped permission can never be read back under its qualifier.

The practical effect is that **Terraform's `aws_lambda_permission` with `qualifier` set never converges** — it creates the permission, then polls `GetPolicy?Qualifier=<alias>` forever, and a later apply hits a `ResourceConflictException` against a statement it was never able to confirm. There is then no way to converge or clean up, because the retry re-hits the same conflict.

This is the standard Terraform shape for letting API Gateway invoke a Lambda alias, so it blocks any Terraform-managed stack that fronts aliased Lambdas with API Gateway. In our case it gates 26 of 28 services; the only 2 that deploy cleanly are the 2 that don't create an alias-qualified permission.

Tested against **Floci 1.5.34** (`floci/floci:1.5.34`, linux/arm64), AWS provider `v5.86.1`.

## Reproduction

```bash
FN=my-fn   # a function with an alias named LIVE

# 1. Add a permission scoped to the LIVE alias
aws --endpoint-url http://localhost:4566 lambda add-permission \
  --function-name $FN --qualifier LIVE \
  --statement-id AllowExecutionFromAPIGateway \
  --action lambda:InvokeFunction --principal apigateway.amazonaws.com

# 2. The returned statement's Resource has NO qualifier suffix:
#    "Resource":"arn:aws:lambda:us-east-1:000000000000:function:my-fn"
#    Real AWS returns  ...:function:my-fn:LIVE

# 3. GetPolicy with and without the qualifier return identical payloads —
#    the qualifier is not distinguished at all.
aws ... lambda get-policy --function-name $FN --qualifier LIVE
aws ... lambda get-policy --function-name $FN
```

With Terraform:

```
module.x.aws_lambda_permission.allow_live_rest_api_alias[0]: Still creating... [4m50s elapsed]
```

observed at 4m50s+ on one service and 8m30s+ on another, then on retry:

```
Error: adding Lambda Permission (my-fn/AllowExecutionFromAPIGateway):
  ResourceConflictException: The statement id (AllowExecutionFromAPIGateway) already exists.
```

`terraform import` of the same resource fails with **`Cannot import non-existent remote object`**, which confirms the provider's read cannot find a permission the CLI reports as present.

Worth stressing: the write path and the CLI read path are both fine. `add-permission`, `get-policy`, `remove-permission` and `--source-arn` conditions all round-trip correctly. The defect is purely that the qualifier is never part of the stored identity.

## Cause

`LambdaController.java` — none of the three policy endpoints declare `@QueryParam("Qualifier")`, though the same annotation is already used elsewhere in the codebase (`LambdaUrlController.java:49`, `LambdaEventInvokeController.java:58`).

`LambdaService.java:1427` — `addPermission` hardcodes the unqualified ARN:

```java
statement.put("Resource", fn.getFunctionArn());
```

`getPolicy` (`:1440`) returns **all** statements regardless of qualifier, and the duplicate-`Sid` check (`:1403`) and `removePermission` (`:1455`) both match on `Sid` alone — so the same `Sid` cannot exist for two different qualifiers, which real AWS permits.

## Suggested fix

Thread `Qualifier` through the three endpoints and make it part of the statement's identity via its `Resource` ARN. Diff below; happy to open it as a PR if the approach looks right.

```diff
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ addPermission
     public Response addPermission(@Context HttpHeaders headers,
                                   @PathParam("functionName") String functionName,
+                                  @QueryParam("Qualifier") String qualifier,
                                   String body) {
-            Map<String, Object> statement = lambdaService.addPermission(region, functionName, request);
+            Map<String, Object> statement = lambdaService.addPermission(region, functionName, qualifier, request);
@@ getPolicy
     public Response getPolicy(@Context HttpHeaders headers,
-                              @PathParam("functionName") String functionName) {
+                              @PathParam("functionName") String functionName,
+                              @QueryParam("Qualifier") String qualifier) {
-            Map<String, Object> data = lambdaService.getPolicy(region, functionName);
+            Map<String, Object> data = lambdaService.getPolicy(region, functionName, qualifier);
@@ removePermission
     public Response removePermission(@Context HttpHeaders headers,
                                      @PathParam("functionName") String functionName,
-                                     @PathParam("statementId") String statementId) {
+                                     @PathParam("statementId") String statementId,
+                                     @QueryParam("Qualifier") String qualifier) {
-        lambdaService.removePermission(region, functionName, statementId);
+        lambdaService.removePermission(region, functionName, statementId, qualifier);
```

```diff
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+    private String policyResourceArn(LambdaFunction fn, String qualifier) {
+        return qualifier != null && !qualifier.isBlank()
+                ? fn.getFunctionArn() + ":" + qualifier
+                : fn.getFunctionArn();
+    }
+
-    public Map<String, Object> addPermission(String region, String functionName, Map<String, Object> request) {
+    public Map<String, Object> addPermission(String region, String functionName, String qualifier, Map<String, Object> request) {
+        String resourceArn = policyResourceArn(fn, qualifier);
         fn.getPolicies().stream()
-                .filter(s -> statementId.equals(s.get("Sid")))
+                .filter(s -> statementId.equals(s.get("Sid")) && resourceArn.equals(s.get("Resource")))
-        statement.put("Resource", fn.getFunctionArn());
+        statement.put("Resource", resourceArn);

-    public Map<String, Object> getPolicy(String region, String functionName) {
+    public Map<String, Object> getPolicy(String region, String functionName, String qualifier) {
+        String resourceArn = policyResourceArn(fn, qualifier);
+        List<Map<String, Object>> matching = fn.getPolicies().stream()
+                .filter(s -> resourceArn.equals(s.get("Resource")))
+                .toList();
+        if (matching.isEmpty()) {
             throw new AwsException("ResourceNotFoundException", "Function not found: " + functionName, 404);
+        }
-        policy.put("Statement", fn.getPolicies());
+        policy.put("Statement", matching);

-    public void removePermission(String region, String functionName, String statementId) {
+    public void removePermission(String region, String functionName, String statementId, String qualifier) {
+        String resourceArn = policyResourceArn(fn, qualifier);
-        boolean removed = fn.getPolicies().removeIf(s -> statementId.equals(s.get("Sid")));
+        boolean removed = fn.getPolicies().removeIf(s ->
+                statementId.equals(s.get("Sid")) && resourceArn.equals(s.get("Resource")));
```

This also fixes the narrower bug that the same `Sid` cannot coexist across qualifiers, which AWS allows.

I have six integration tests written against `LambdaPermissionTagLayerIntegrationTest` covering: qualified add stores the suffixed ARN; qualified get returns it; unqualified get does **not** see it (404); same `Sid` under a different qualifier does not 409; qualified remove leaves the unqualified statement intact.

**Caveat, stated plainly:** I have not been able to compile or run this — there is no JDK 25 on the machine I diagnosed it on and I couldn't get a toolchain image pulled in time. The diagnosis and the reproduction above are all verified against the running 1.5.34 container; the patch itself is reasoned from the source and is **not build-verified**. I'd rather tell you that than imply otherwise. If you'd like it as a PR I'll get a proper toolchain up first and land it green with the tests.

## Related

While diagnosing this I also hit these on 1.5.34, happy to file separately if useful:

- **Unhandled paths fall through to the S3 handler.** Any unimplemented operation on any service returns XML `NoSuchBucket`, so the AWS provider reports `invalid character '<' looking for beginning of value` instead of a recognisable "not implemented". `GET /this-path-does-not-exist-at-all/foo` reproduces it. This one is worth fixing ahead of everything else — it makes every other gap illegible.
- `apigateway:UpdateAuthorizer` returns **405** (`GET` on the authorizer returns 200, `PATCH` returns 405).
- `apigateway:PutGatewayResponse` and `apigateway:DeleteIntegrationResponse` unimplemented.
- `apigatewayv2:CreateDomainName` unimplemented.
- `apigateway:GetBasePathMapping` answers "couldn't find resource" for a mapping it just created.
- `iam:GetRole` omits the `Tags` field (present via `ListRoleTags`), and `kinesis:ListTagsForStream` returns `[]` — both cause permanent Terraform tag drift, so no apply is ever a clean no-op.

Thanks for Floci — the create paths have been solid, and everything above was found by pushing a fairly large real-world Terraform estate through it.

aniketshukla1:
## Summary

Closes #2124.

`AddPermission`, `GetPolicy` and `RemovePermission` ignored the `Qualifier` query parameter. Every statement was stored against the **unqualified** function ARN, so an alias-scoped permission could never be read back under its alias:

- the returned statement's `Resource` was `...:function:my-fn` where AWS returns `...:function:my-fn:LIVE`
- `GetPolicy` with and without `?Qualifier=LIVE` returned **identical** payloads

That is what stops Terraform's `aws_lambda_permission` with `qualifier` set from ever converging: it creates the permission, polls `GetPolicy?Qualifier=<alias>` forever, and a later apply hits `ResourceConflictException` against a statement its read could never confirm — with no way to converge or clean up (`terraform import` also fails with *Cannot import non-existent remote object*).

## Fix

AWS keeps a **separate resource policy per qualifier**, so the qualifier is part of a statement's identity.

- `LambdaController` — declare `@QueryParam("Qualifier")` on the three policy endpoints and thread it through, matching the annotation already used by `LambdaUrlController` / `LambdaEventInvokeController`.
- `LambdaService` — resolve the ref through the existing `resolveWithRegion()` helper, so a qualifier **embedded in the function name** (`arn:...:function:my-fn:LIVE`, or `my-fn:LIVE`) is reconciled with the query-string value exactly as the function-URL endpoints already do (including the "derived qualifier does not match" 400).

The scope is derived from the statement's own `Resource` rather than a new stored field, so statements **persisted before this change** stay readable as the function's unqualified policy — no migration, no model change.

Resulting behaviour:

| Call | Before | After |
|---|---|---|
| `AddPermission?Qualifier=X` | `Resource` = function ARN | `Resource` = `<functionArn>:X` |
| `GetPolicy?Qualifier=X` | every statement | only that qualifier's statements (404 if none) |
| `RemovePermission?Qualifier=X` | removed by Sid anywhere | removes only within that qualifier |
| same `StatementId` on function *and* alias | 409 conflict | allowed — separate policies |
| duplicate within one qualifier | 409 | 409 (unchanged) |

**Deliberately out of scope:** qualifier *existence* is not validated (an unknown alias is accepted). The endpoints did not validate it before either, so adding that is a separate behaviour change rather than part of this fix — happy to fold it in if you'd prefer.

## Tests

New `LambdaPermissionQualifierIntegrationTest` — 12 cases over a function with a published version and a `LIVE` alias: the alias-scoped `Resource`, qualified vs unqualified isolation in both directions, same-`StatementId`-different-qualifier, duplicate-within-qualifier still conflicting, scoped removal (including an unknown qualifier → 404), and the qualifier arriving embedded in the function name.

**Verified fail-before / pass-after:** against the unfixed service **7 of the 12 fail**, with exactly the symptom from the report —

```
expected: <arn:aws:lambda:us-east-1:000000000000:function:qualifier-perm-fn:LIVE>
 but was: <arn:aws:lambda:us-east-1:000000000000:function:qualifier-perm-fn>
```

Full `Lambda*` suite: **235/237 pass**. The 2 failures are `LambdaReactiveSyncIntegrationTest` (`setupAndInitialInvoke`, `uploadV2AndVerifyAutoSync`), which I confirmed fail identically on a pristine `origin/main` — pre-existing and unrelated to this change. The existing `LambdaPermissionTagLayerIntegrationTest` (19 cases, all unqualified) stays green, covering backward compatibility.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -1394,13 +1394,34 @@ public class LambdaService {
 
     // ──────────────────────────── Permissions (Policy) ────────────────────────────
 
-    public Map<String, Object> addPermission(String region, String functionName, Map<String, Object> request) {
-        LambdaFunction fn = getFunction(region, functionName);
+    /**
+     * The ARN a resource-policy statement is scoped to: the bare function ARN, or
+     * {@code <functionArn>:<qualifier>} for an alias/version-scoped permission.
+     *
+     * <p>AWS keeps a SEPARATE resource policy per qualifier, so the qualifier is part of a
+     * statement's identity — the same {@code StatementId} may exist on the function and on
+     * each alias. We derive that identity from the statement's own {@code Resource} rather
+     * than storing a parallel field, which keeps already-persisted (always unqualified)
+     * statements readable as the function's own policy (#2124).
+     */
+    private static String policyResourceArn(LambdaFunction fn, String qualifier) {
+        return qualifier == null ? fn.getFunctionArn() : fn.getFunctionArn() + ":" + qualifier;
+    }
+
+    private static boolean scopedTo(Map<String, Object> statement, String resourceArn) {
+        return resourceArn.equals(statement.get("Resource"));
+    }
+
+    public Map<String, Object> addPermission(String region, String functionName, String qualifier, Map<String, Object> request) {
+        LambdaArnUtils.ResolvedFunctionRef ref = resolveWithRegion(region, functionName, qualifier);
+        LambdaFunction fn = getFunction(region, ref.name());
+        String resourceArn = policyResourceArn(fn, ref.qualifier());
         String statementId = (String) request.get("StatementId");
         if (statementId == null || statementId.isBlank()) {
             throw new AwsException("InvalidParameterValueException", "StatementId is required", 400);
         }
         fn.getPolicies().stream()
+                .filter(s -> scopedTo(s, resourceArn))
                 .filter(s -> statementId.equals(s.get("Sid")))
                 .findFirst()
                 .ifPresent(s -> {
@@ -1424,7 +1445,7 @@ public class LambdaService {
             statement.put("Principal", principal);
         }
         statement.put("Action", action);
-        statement.put("Resource", fn.getFunctionArn());
+        statement.put("Resource", resourceArn);
         if (sourceArn != null) {
             statement.put("Condition", Map.of("ArnLike", Map.of("AWS:SourceArn", sourceArn)));
         } else if (sourceAccount != null) {
@@ -1437,22 +1458,30 @@ public class LambdaService {
         return statement;
     }
 
-    public Map<String, Object> getPolicy(String region, String functionName) {
-        LambdaFunction fn = getFunction(region, functionName);
-        if (fn.getPolicies().isEmpty()) {
+    public Map<String, Object> getPolicy(String region, String functionName, String qualifier) {
+        LambdaArnUtils.ResolvedFunctionRef ref = resolveWithRegion(region, functionName, qualifier);
+        LambdaFunction fn = getFunction(region, ref.name());
+        String resourceArn = policyResourceArn(fn, ref.qualifier());
+        List<Map<String, Object>> statements = fn.getPolicies().stream()
+                .filter(s -> scopedTo(s, resourceArn))
+                .toList();
+        if (statements.isEmpty()) {
             throw new AwsException("ResourceNotFoundException",
                     "Function not found: " + functionName, 404);
         }
         Map<String, Object> policy = new java.util.LinkedHashMap<>();
         policy.put("Version", "2012-10-17");
         policy.put("Id", "default");
-        policy.put("Statement", fn.getPolicies());
+        policy.put("Statement", statements);
         return Map.of("policy", policy, "revisionId", fn.getRevisionId());
     }
 
-    public void removePermission(String region, String functionName, String statementId) {
-        LambdaFunction fn = getFunction(region, functionName);
-        boolean removed = fn.getPolicies().removeIf(s -> statementId.equals(s.get("Sid")));
+    public void removePermission(String region, String functionName, String qualifier, String statementId) {
+        LambdaArnUtils.ResolvedFunctionRef ref = resolveWithRegion(region, functionName, qualifier);
+        LambdaFunction fn = getFunction(region, ref.name());
+        String resourceArn = policyResourceArn(fn, ref.qualifier());
+        boolean removed = fn.getPolicies()
+                .removeIf(s -> statementId.equals(s.get("Sid")) && scopedTo(s, resourceArn));
         if (!removed) {
             throw new AwsException("ResourceNotFoundException",
                     "Statement " + statementId + " not found in function " + functionName, 404);
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -1467,7 +1467,7 @@ public class LambdaService {
                 .toList();
         if (statements.isEmpty()) {
             throw new AwsException("ResourceNotFoundException",
-                    "Function not found: " + functionName, 404);
+                    "The resource you requested does not exist.", 404);
         }
         Map<String, Object> policy = new java.util.LinkedHashMap<>();
         policy.put("Version", "2012-10-17");
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -479,12 +479,13 @@ public class LambdaController {
     @Path("/functions/{functionName}/policy")
     public Response addPermission(@Context HttpHeaders headers,
                                   @PathParam("functionName") String functionName,
+                                  @QueryParam("Qualifier") String qualifier,
                                   String body) {
         String region = regionResolver.resolveRegion(headers);
         try {
             @SuppressWarnings("unchecked")
             Map<String, Object> request = objectMapper.readValue(body, Map.class);
-            Map<String, Object> statement = lambdaService.addPermission(region, functionName, request);
+            Map<String, Object> statement = lambdaService.addPermission(region, functionName, qualifier, request);
             String statementJson = objectMapper.writeValueAsString(statement);
             ObjectNode root = objectMapper.createObjectNode();
             root.put("Statement", statementJson);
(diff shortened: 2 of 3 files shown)


src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
```diff
@@ -1437,22 +1458,30 @@ public Map<String, Object> addPermission(String region, String functionName, Map
...
+    public Map<String, Object> getPolicy(String region, String functionName, String qualifier) {
+        LambdaArnUtils.ResolvedFunctionRef ref = resolveWithRegion(region, functionName, qualifier);
+        LambdaFunction fn = getFunction(region, ref.name());
+        String resourceArn = policyResourceArn(fn, ref.qualifier());
+        List<Map<String, Object>> statements = fn.getPolicies().stream()
+                .filter(s -> scopedTo(s, resourceArn))
+                .toList();
+        if (statements.isEmpty()) {
             throw new AwsException("ResourceNotFoundException",
                     "Function not found: " + functionName, 404);
```
pgermosen: Minor/optional: this reuses the pre-existing `"Function not found: " + functionName` message for a case that's now broader than it used to describe — a function that genuinely exists (and may well have permissions under *other* qualifiers) but has none for the requested one. For comparison, moto's equivalent (`GenericResourcNotFound`) uses a generic `"The resource you requested does not exist."` rather than implying the function itself is missing. Not something I'd hold this PR on — the message was already reused as-is before this PR for the "no policy at all" case, this just extends an existing wording choice to a new condition — but worth a look if you're touching this area again.

pgermosen (approved):
Looks good


- **"AWS keeps a separate resource policy per qualifier"**: confirmed via the [AddPermission API reference](https://docs.aws.amazon.com/lambda/latest/api/API_AddPermission.html) — "You can apply the policy at the function level, or specify a qualifier to restrict access to a single version or alias." Cross-checked against moto's `awslambda/policy.py`, which independently takes the same approach (a qualifier other than `$LATEST` gets appended onto the statement's `Resource` ARN before storage) — this isn't a novel design, it's the same pattern an established reference implementation already uses.
- **`GetPolicy` 404s when a qualifier has no statements**: confirmed as correct AWS behavior via the [GetPolicy API reference](https://docs.aws.amazon.com/lambda/latest/api/API_GetPolicy.html) (`ResourceNotFoundException` is a documented error) and moto's `Policy.wire_format()`, which raises `GenericResourcNotFound` precisely when `not self.statements`.
- **Qualifier reconciliation**: `resolveWithRegion`/`LambdaArnUtils.resolveWithQualifier` is pre-existing, already-used code (function-URL endpoints use it today), not new logic written for this PR — confirmed the blank/null query-qualifier normalization means `policyResourceArn`'s `qualifier == null` check is safe against `?Qualifier=` (empty) as well as an absent parameter.


@aniketshukla1 
One very minor, non-blocking note below — genuinely optional, not something to hold this on.

src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
aniketshukla1: Updated this to return the generic AWS-style ResourceNotFound message when the function exists but the requested policy scope is empty. Added a regression assertion for the error type and message in LambdaPermissionQualifierIntegrationTest; `./mvnw test -Dtest=LambdaPermissionQualifierIntegrationTest` passes on JDK 25.

aniketshukla1:
@pgermosen updated this. Please check.
Thanks

aniketshukla1:
@pgermosen can you merge this as well ? Or this needs any more changes ?

PR #2130, opened 2026-08-05
fix(apigateway): create the stage when CreateDeployment carries a stageName

### LINKED ISSUE #2120: [BUG] API Gateway: CreateDeployment with stageName does not create the stage (deployed API returns "Stage not found")
## Service

API Gateway (REST)

## AWS API Action

`CreateDeployment` (with `stageName`)

## Expected behavior

In real AWS — and in LocalStack — `CreateDeployment` with a `stageName` creates the stage as a side effect. The deployed API is immediately invokable and `GetStages` lists it.

## Actual behavior

The deployment is created, but **no stage is created**. `GetStages` returns an empty list and invoking the deployed API returns:

```json
{"message":"Stage not found"}
```

Floci logs `Created deployment <id> for API <id>` and nothing further.

Calling `CreateStage` explicitly afterwards fixes it, and the same invoke then returns `200`.

## Why this matters

Terraform's `aws_api_gateway_deployment` accepts a `stage_name` and relies on this side effect. Configurations that use it without a separate `aws_api_gateway_stage` resource apply **cleanly** and then serve 404 on every route — a silent failure with no error at apply time.

## Reproduction

```bash

API=$(aws apigateway create-rest-api --name stage-repro --output text --query id)
ROOT=$(aws apigateway get-resources --rest-api-id $API --output text --query 'items[0].id')
RES=$(aws apigateway create-resource --rest-api-id $API --parent-id $ROOT --path-part health --output text --query id)
aws apigateway put-method --rest-api-id $API --resource-id $RES --http-method GET --authorization-type NONE
aws apigateway put-integration --rest-api-id $API --resource-id $RES --http-method GET --type MOCK \
  --request-templates '{"application/json":"{\"statusCode\":200}"}'
aws apigateway put-method-response --rest-api-id $API --resource-id $RES --http-method GET --status-code 200
aws apigateway put-integration-response --rest-api-id $API --resource-id $RES --http-method GET --status-code 200 \
  --response-templates '{"application/json":"{\"status\":\"ok\"}"}'

DEP=$(aws apigateway create-deployment --rest-api-id $API --stage-name local --output text --query id)

aws apigateway get-stages --rest-api-id $API          # -> empty
curl http://localhost:4566/restapis/$API/local/_user_request_/health
# -> {"message":"Stage not found"}

aws apigateway create-stage --rest-api-id $API --stage-name local --deployment-id $DEP
curl http://localhost:4566/restapis/$API/local/_user_request_/health
# -> {"status":"ok"}
```

## Environment

`floci/floci:1.5.34`, linux/arm64, macOS host, storage mode `persistent`.

aniketshukla1:
## Summary

Closes #2120.

CreateDeployment ignored stageName: it stored a deployment but did not create the named stage. GetStages remained empty, and invocation returned {"message":"Stage not found"}.

## Behavior

CreateDeployment now creates the named stage pointing at the new deployment. On a later deployment with the same name, it repoints that stage while retaining its other settings; request variables are still applied. This supports repeated Terraform applies. The AWS API reference does not document named-stage collision behavior, so this is an intentional compatibility decision rather than a claimed AWS invariant.

A deployment without stageName remains unchanged and creates no stage.

## Tests

ApiGatewayDeploymentStageNameIntegrationTest covers seven cases using a REST API and MOCK GET /health route:

- named stages appear in GetStages and GetStage
- the deployed route returns 200 instead of Stage not found
- a redeploy updates one existing stage without duplication
- stage description and request variables are covered
- deployments without stageName create no stage

Validation:

- JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw test -Dtest=ApiGatewayDeploymentStageNameIntegrationTest
- 7 passed

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Docs or chore

## Checklist

- [x] New integration coverage is included
- [x] Focused regression test passes locally on Java 25
- [x] The branch is refreshed from the latest main before this update

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
@@ -469,7 +469,42 @@ public class ApiGatewayService {
         Deployment deployment = new Deployment(shortId(10), description, System.currentTimeMillis() / 1000L);
         deploymentStore.put(deploymentKey(region, apiId, deployment.id()), deployment);
         LOG.infov("Created deployment {0} for API {1}", deployment.id(), apiId);
+
+        String stageName = (String) request.get("stageName");
+        if (stageName != null && !stageName.isBlank()) {
+            deployStage(region, apiId, stageName, deployment.id(), request);
+        }
         return deployment;
+    }
+
+    /**
+     * Points {@code stageName} at {@code deploymentId}, creating the stage if it doesn't exist.
+     *
+     * <p>The API does not document collision behavior. Repointing preserves existing stage
+     * settings and supports repeated deployments.
+     */
+    private void deployStage(String region, String apiId, String stageName, String deploymentId,
+                             Map<String, Object> request) {
+        String key = stageKey(region, apiId, stageName);
+        long now = System.currentTimeMillis() / 1000L;
+        Stage stage = stageStore.get(key).orElse(null);
+        if (stage == null) {
+            stage = new Stage();
+            stage.setStageName(stageName);
+            stage.setCreatedDate(now);
+            stage.setDescription((String) request.get("stageDescription"));
+        }
+        stage.setDeploymentId(deploymentId);
+        stage.setLastUpdatedDate(now);
+
+        @SuppressWarnings("unchecked")
+        Map<String, String> variables = (Map<String, String>) request.get("variables");
+        if (variables != null) {
+            stage.setVariables(variables);
+        }
+
+        stageStore.put(key, stage);
+        LOG.infov("Deployed stage {0} of API {1} to deployment {2}", stageName, apiId, deploymentId);
     }
 
     public List<Deployment> getDeployments(String region, String apiId) {
--- a/src/test/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayDeploymentStageNameIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayDeploymentStageNameIntegrationTest.java
@@ -1,0 +1,204 @@
+package io.github.hectorvent.floci.services.apigateway;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.http.ContentType;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.*;
+
+/**
+ * {@code CreateDeployment} with a {@code stageName} must create the stage as a side effect,
+ * leaving the deployed API immediately invokable.
+ *
+ * <p>Terraform's {@code aws_api_gateway_deployment} takes a {@code stage_name} and relies on
+ * this; without it a configuration that has no separate {@code aws_api_gateway_stage} applies
+ * cleanly and then serves {@code {"message":"Stage not found"}} on every route.
+ *
+ * @see <a href="https://github.com/floci-io/floci/issues/2120">Issue #2120</a>
+ */
+@QuarkusTest
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+class ApiGatewayDeploymentStageNameIntegrationTest {
+
+    private static final String STAGE = "local";
+
+    private static String apiId;
+    private static String resourceId;
+    private static String firstDeploymentId;
+
+    @Test @Order(1)
+    void createApiWithAMockRoute() {
+        apiId = given()
+                .contentType(ContentType.JSON)
+                .body("{\"name\":\"stage-repro\"}")
+                .when().post("/restapis")
+                .then()
+                .statusCode(201)
+                .extract().path("id");
+
+        String rootId = given()
+                .when().get("/restapis/" + apiId + "/resources")
+                .then()
+                .statusCode(200)
+                .extract().path("item[0].id");
+
+        resourceId = given()
+                .contentType(ContentType.JSON)
+                .body("{\"pathPart\":\"health\"}")
+                .when().post("/restapis/" + apiId + "/resources/" + rootId)
+                .then()
+                .statusCode(201)
+                .extract().path("id");
+
+        given()
+                .contentType(ContentType.JSON)
+                .body("{\"authorizationType\":\"NONE\"}")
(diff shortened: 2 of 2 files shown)


src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
```diff
@@ -469,9 +469,47 @@ public Deployment createDeployment(String region, String apiId, Map<String, Obje
...
 
+    /**
+     * Points {@code stageName} at {@code deploymentId}, creating the stage if it doesn't exist.
+     *
+     * <p>A redeploy keeps the stage's existing settings and only moves it to the new deployment —
+     * that is what repeated {@code terraform apply}s do, and it matches AWS, where
+     * {@code stageDescription} applies when the stage is first created rather than overwriting
+     * the description of an existing one.
+     */
+    private void deployStage(String region, String apiId, String stageName, String deploymentId,
```
pgermosen, line 486: **`stageDescription` preservation and the existing-stage behavior are inferences, not documented AWS**

The javadoc here and the PR description both say the redeploy semantics *match* AWS:

> it matches AWS, where `stageDescription` applies when the stage is first created rather than overwriting the description of an existing one

I went looking for that in the API reference and it isn't there. The [CreateDeployment](https://docs.aws.amazon.com/apigateway/latest/api/API_CreateDeployment.html) page documents `stageDescription` only as *"The description of the Stage resource for the Deployment resource to create"*, and says nothing at all about what happens when `stageName` already exists — while listing `ConflictException` (409) among its possible errors, which if anything cuts the other way.

I'm **not** saying the behavior is wrong. Re-pointing an existing stage is what "deploy to stage" means operationally, it's what repeated `terraform apply`s need, and 409-ing the second apply would be clearly worse. It's the right call. The issue is only that it's presented as documented parity when it's a reasonable inference.

**Suggested:** soften to something like "AWS doesn't document the collision case; re-pointing is what repeated deploys require, so that's what this does." Costs nothing, and stops the next person treating it as a verified AWS invariant they mustn't touch.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
```diff
@@ -469,9 +469,47 @@ public Deployment createDeployment(String region, String apiId, Map<String, Obje
...
+            stage.setStageName(stageName);
+            stage.setCreatedDate(now);
+            stage.setDescription((String) request.get("stageDescription"));
+        }
+        stage.setDeploymentId(deploymentId);
+        stage.setLastUpdatedDate(now);
+
+        @SuppressWarnings("unchecked")
+        Map<String, String> variables = (Map<String, String>) request.get("variables");
+        if (variables != null) stage.setVariables(variables);
```
pgermosen: **Missing braces — AGENTS.md asks for them explicitly** *(nit)*

```java
if (variables != null) stage.setVariables(variables);
```

AGENTS.md `Code Style` (`:271`) says **"Always use braces in conditionals."**

In fairness, this is copied verbatim from `createStage:552`, which has the identical line — so it's also "Follow existing project patterns" (`:273`), and the two rules genuinely conflict here.

Raising it only because it's cheap and because the bots enforce it: greptile flagged exactly this on #2150 and that author spent a follow-up commit adding braces. Worth doing pre-emptively rather than in a round trip.

**Suggested:** add the braces here; leave `createStage:552` alone as unrelated churn.

pgermosen (commented):
Looks good.

Real bug with a real-world trigger, and the fix is small and well-scoped. AWS's own parameter documentation confirms the premise directly:

| parameter | AWS documentation |
|---|---|
| `stageName` | *"The name of the Stage resource for the Deployment resource to create."* |
| `stageDescription` | *"The description of the Stage resource for the Deployment resource to create."* |
| `variables` | *"A map that defines the stage variables for the Stage resource that is associated with the new deployment."* |

So the stage genuinely is a side effect of `CreateDeployment`, and the Terraform framing is right.


---

### The "11 pre-existing failures" don't reproduce

> Full `ApiGateway*` suite (v1 + v2): **491/502 pass**. The 11 failures are in `ApiGatewayAuthorizerContextIntegrationTest` (2) and `ApiGatewayRequestAuthorizerIntegrationTest` (9) ... I confirmed they fail identically (same 2 + 9) on a pristine `origin/main`.

Same total, zero failures. Both named classes ran and passed outright — `ApiGatewayAuthorizerContextIntegrationTest` 12/12, `ApiGatewayRequestAuthorizerIntegrationTest` 25/25.

The likely explanation is in your own wording: those are the tests that "need real Lambda invocation", and this machine has Docker available (28.2.2). So they look **environment-dependent** rather than pre-existing on `main` — they'd fail wherever Lambda containers can't start.

That distinction matters for the reader: "pre-existing and unrelated" tells a maintainer not to investigate, whereas "needs Docker, fails without it" tells them why their own run might differ.

**Suggested:** reword to say the failures depend on Lambda/Docker availability rather than that they reproduce on `main`.


The two inline comments are a framing point and a style nit.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
aniketshukla1: Updated the Javadoc, commit body, and PR description to make this explicit: the API does not document named-stage collision behavior. Repointing while retaining existing settings is the intentional choice for repeated deployments, not a claimed AWS invariant.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
aniketshukla1: Added braces to the new variables conditional, leaving the pre-existing createStage formatting untouched as suggested.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
pgermosen: Confirmed fixed — the javadoc now reads "The API does not document collision behavior. Repointing preserves existing stage settings and supports repeated deployments," and the PR description matches. No longer reads as a claimed AWS invariant.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
pgermosen: Confirmed — braces added to the new conditional, `createStage:552` left untouched. Good.

pgermosen (approved):
Looks good.

PR #2131, opened 2026-08-05
test(lambda): wait for the parked poll instead of sleeping in extension stop race

rogueserenity:
## Summary

`extensionRegister_racedAgainstStop_eventuallyDeliversShutdown` set up its `stop()` race with `Thread.sleep(5)`, assuming the extension's `/event/next` poll had parked server-side. The sleep only assumes it. Under load it expires first, so `stop()` closes the listening socket while the request is still in flight — the poll never reaches the handler, `stop()`'s SHUTDOWN fan-out finds no waiting context, and the client's `get()` throws instead of returning the SHUTDOWN.

Surfaces as `EOFException` (connection accepted, then torn down) or `ConnectException` (lost the race earlier). It failed CI this way on #1760.

Replaces the sleep with a condition-based wait on a new package-private `RuntimeApiServer.isExtensionParked(...)`.

## What changed

- `Thread.sleep(5)` -> `awaitExtensionParked(...)`, polling until the poll has provably parked.
- New package-private `RuntimeApiServer.isExtensionParked(String)`, read under `lock` per the class's documented locking discipline.
- `RegisteredExtension.hasWaitingContext()` — a non-consuming peek, unlike `takeWaitingContext()`.

No change to shutdown behavior.

## Test plan

Local, with 24 spinners on 12 cores:

| Check | Result |
|---|---|
| Original test under load | fails — reproduces the CI failure |
| Fixed test, identical load | passes |
| Fixed test, with the "poll `pendingEvents` before checking `stopped`" handler fix reverted | fails — regression still caught |
| Full lambda suite | 466/466 pass |

The third row is the point: the flake is gone without weakening the regression the test exists to cover.

Then validated on CI runners, running the full 8317-test suite four times (the flake only appears under that load, not when running the race tests alone):

| Variant | Result |
|---|---|
| Pristine `main` | 4/4 pass |
| This branch | 4/4 pass |

Two further changes were tried and dropped after the same four-run comparison showed them failing 3/4: applying the same wait to the sibling `extensionShutdown_racedAgainstStop_isNeverOrphaned` test, and reordering `stop()` to close the listener after dispatching SHUTDOWN. Neither is in this PR. The sibling test keeps its existing `Thread.sleep(5)`.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/model/RegisteredExtension.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/model/RegisteredExtension.java
@@ -50,6 +50,15 @@ public class RegisteredExtension {
     }
 
     /**
+     * Whether a poll is currently parked, without consuming it — unlike
+     * {@link #takeWaitingContext()}, which clears the context as it hands it over. Called under the
+     * owning server's lock, like the rest of this class.
+     */
+    public boolean hasWaitingContext() {
+        return waitingContext != null;
+    }
+
+    /**
      * Records that this extension has issued its first {@code /extension/event/next}, which is
      * what AWS treats as the extension being init-ready (registering alone only obtains an
      * identifier). Called under the owning server's lock, like the rest of this class.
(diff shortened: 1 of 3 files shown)


rogueserenity:
@abanna found a flake from a CI run on another PR. this should fix it.

hectorvent (approved):
Thanks @rogueserenity, 
It looks good. Merging

Issue #234, opened 2026-04-06
[BUG] GetTokensFromRefreshToken not supported


enotesupa:
## Service

Cognito Identity Provider

## AWS API Action

GetTokensFromRefreshToken

## Expected behavior

Support the `GetTokensFromRefreshToken` API and return new AccessToken/IdToken. This is functionally equivalent to `InitiateAuth` with `REFRESH_TOKEN_AUTH` flow.

Ref: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_GetTokensFromRefreshToken.html

## Actual behavior

```json
{
  "__type": "UnsupportedOperation",
  "message": "Operation GetTokensFromRefreshToken is not supported."
}
```

## Reproduction

```bash
# After obtaining a RefreshToken via InitiateAuth
curl -s -X POST http://localhost:4566/ \
  -H "Content-Type: application/x-amz-json-1.1" \
  -H "X-Amz-Target: AWSCognitoIdentityProviderService.GetTokensFromRefreshToken" \
  -d '{"RefreshToken":"<refresh-token>","ClientId":"<client-id>"}'
```

## Environment

- Floci version / image tag: 1.3.0
- How you're running Floci: Docker (`hectorvent/floci:1.3.0`)

PR #2132, opened 2026-08-05
fix(cognito): reject invalid refresh tokens in InitiateAuth REFRESH_TOKEN_AUTH

### LINKED ISSUE #2113: [BUG] Cognito InitiateAuth REFRESH_TOKEN_AUTH accepts arbitrary refresh tokens
## Service

Cognito

## AWS API Action

`InitiateAuth` with `AuthFlow=REFRESH_TOKEN_AUTH`

## Expected behavior

Passing a non-issued / garbage `REFRESH_TOKEN` (e.g. `invalid-refresh-token`) should fail with AWS-compatible auth error behavior (`NotAuthorizedException` / Invalid Refresh Token), and must **not** return a new `AuthenticationResult`.

Real Amazon Cognito rejects these requests.

## Actual behavior

On Floci `1.5.34`, `InitiateAuth` + `REFRESH_TOKEN_AUTH` with `REFRESH_TOKEN=invalid-refresh-token` returns HTTP 200 and mints new Access/Id tokens. Decoded JWT claims include `username: "unknown"`.

This breaks negative-path integration tests that assert refresh failure for invalid tokens (happy-path refresh with a real refresh token works).

Related but distinct from #1395 (revocation / global sign-out). #234 added invalid-token rejection for `GetTokensFromRefreshToken`; `InitiateAuth` `REFRESH_TOKEN_AUTH` still accepts arbitrary strings.

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=test
export AWS_DEFAULT_REGION=us-west-2

# Assume an existing user pool + confidential app client with ALLOW_REFRESH_TOKEN_AUTH
# (CreateUserPool / CreateUserPoolClient / AdminCreateUser / AdminSetUserPassword as usual)

aws cognito-idp initiate-auth \
  --client-id "$CLIENT_ID" \
  --auth-parameters "REFRESH_TOKEN=invalid-refresh-token,SECRET_HASH=$SECRET_HASH"

# Actual: AuthenticationResult with AccessToken / IdToken
# Expected: NotAuthorizedException (or equivalent Invalid Refresh Token)
```

Also reproduces without SECRET_HASH depending on client config.

## Environment

- Floci version / image tag: `floci/floci:1.5.34` (latest at time of report)
- Client: AWS CLI + AWS SDK for Java 2.x (`InitiateAuth` / `REFRESH_TOKEN_AUTH`)

## Context

Reported while migrating [get-backbone](https://github.com/get-backbone) local/CI Cognito from LocalStack to Floci. Auth service ITs assert invalid refresh tokens fail; Floci currently succeeds those calls.

- [ ] Yes
- [x] No (happy to validate a fix against our IT suite)

Agboolafeyikemi:
## Summary

- `InitiateAuth` with `AuthFlow=REFRESH_TOKEN_AUTH` accepted arbitrary/garbage `REFRESH_TOKEN` values and any well-formed-but-unresolvable token, silently minting fresh Access/Id tokens for a fake `username: "unknown"` instead of failing like real Cognito (and like Floci's own `GetTokensFromRefreshToken`, fixed in #234).
- `CognitoAuthFlowHandler#handleRefreshToken` now throws `NotAuthorizedException` ("Invalid Refresh Token") when the token fails to parse, and when the encoded username doesn't resolve to a real user in the pool, instead of catching and swallowing the failure into a fallback token-mint path.

Fixes #2113

## Test plan

- [x] Reproduced first: added `initiateAuthRefreshTokenAuthRejectsInvalidRefreshToken` and `initiateAuthRefreshTokenAuthRejectsUnknownWellFormedRefreshToken` to `CognitoIntegrationTest`, confirmed both failed against unfixed `main` (HTTP 200 with a minted `AuthenticationResult` for `username: "unknown"`).
- [x] Applied the fix, reran the same tests — both now correctly return `400 NotAuthorizedException`, confirming the original bug is gone.
- [x] `./mvnw test -Dtest=CognitoIntegrationTest` — 69/69 passing (no regressions in existing refresh-token/JWT flows).
- [x] `./mvnw test -Dtest=CognitoServiceTest` — 113/113 passing.
- [x] `./mvnw test -Dtest=GlobalSignOutIntegrationTest,AdminUserGlobalSignOutIntegrationTest,OriginJtiIntegrationTest` — 33/33 passing (revocation/global-sign-out paths that also touch refresh-token handling).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
@@ -246,7 +246,17 @@ final class CognitoAuthFlowHandler {
         // Check revocation before issuing new tokens
         service.validateRefreshTokenNotRevoked(refreshTokenUuid, pool.getId(), username, iat);
 
-        CognitoUser user = service.adminGetUser(pool.getId(), username);
+        CognitoUser user;
+        try {
+            user = service.adminGetUser(pool.getId(), username);
+        } catch (AwsException ae) {
+            // Real Cognito never reveals whether the encoded username exists; any
+            // refresh token that doesn't resolve to a real user is simply invalid.
+            if ("UserNotFoundException".equals(ae.getErrorCode())) {
+                throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
+            }
+            throw ae;
+        }
         CognitoService.ClaimsOverride override = firePreTokenGeneration(pool, client, user,
                 clientMetadata, "TokenGeneration_RefreshTokens");
         Map<String, Object> auth = new HashMap<>();
--- a/src/test/java/io/github/hectorvent/floci/services/cognito/CognitoIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cognito/CognitoIntegrationTest.java
@@ -2389,6 +2389,27 @@ class CognitoIntegrationTest {
                 .body("__type", org.hamcrest.Matchers.equalTo("NotAuthorizedException"));
     }
 
+    @Test
+    @Order(102)
+    void initiateAuthRefreshTokenAuthRejectsTokenWithNonNumericIssuedAt() {
+        // Well-formed shape (5 base64 parts), but the issued-at field is not a number.
+        // Must fail with NotAuthorizedException, not a 500 from an unguarded parseLong.
+        String bogusToken = java.util.Base64.getEncoder().withoutPadding().encodeToString(
+                (poolId + "|" + USERNAME + "|" + clientId + "|not-a-number|"
+                        + UUID.randomUUID()).getBytes(StandardCharsets.UTF_8));
+
+        cognitoAction("InitiateAuth", """
+                {
+                  "ClientId": "%s",
+                  "AuthFlow": "REFRESH_TOKEN_AUTH",
+                  "AuthParameters": { "REFRESH_TOKEN": "%s" }
+                }
+                """.formatted(clientId, bogusToken))
+                .then()
+                .statusCode(400)
+                .body("__type", org.hamcrest.Matchers.equalTo("NotAuthorizedException"));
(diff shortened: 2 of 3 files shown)


hectorvent:
Thank you for reproducing this first and pinning both rejection paths with regression tests. NotAuthorizedException is declared on InitiateAuth, so the shape is right ([botocore cognito-idp model](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json)).

**(blocking)** A Base64-decodable five-part token whose issued-at is not numeric still reaches the unguarded Long.parseLong and surfaces as a 500 NumberFormatException, where every other invalid-token path on this flow returns 400 NotAuthorizedException. Guarding that parse would round this out.

**(blocking)** PR #2135 targets the same issue (#2113) with a near-identical rewrite of this hunk, and additionally fixes the refresh-token expiry unit bug and scopes the token to its own pool, so the two cannot both merge. Could you coordinate there before either lands?

The parse guard plus resolving the overlap with #2135 unblocks this.

Agboolafeyikemi:
Pushed in [31e72db1](https://github.com/floci-io/floci/commit/31e72db1) + a regression test in [aae10350](https://github.com/floci-io/floci/commit/aae10350).  Non-numeric issued-at on an otherwise well-formed token now returns 400 NotAuthorizedException instead of a 500.

On the [#2135](https://github.com/floci-io/floci/pull/2135) overlap: I'd suggest going with the ordering [AbhigyaKrishna already proposed there](https://github.com/floci-io/floci/pull/2135#issue-comment-url) , this PR lands first for the fabricated-token defect, and [#2135](https://github.com/floci-io/floci/pull/2135) rebases down to just the expiry-unit fix, which is a separate bug this PR doesn't touch. That keeps each PR scoped to one defect and avoids either of us redoing the other's reviewed work. Happy to do it the other way if you'd rather sequence it differently.

hectorvent (approved):
Thanks @Agboolafeyikemi

PR #2133, opened 2026-08-05
fix(ecs): resolve Secrets Manager JSON-key selectors in task definition secrets

### LINKED ISSUE #1912: [FEAT] ECS: support Secrets Manager JSON-key selectors in task definition secrets
## Service

ECS / Secrets Manager

## API Action / Feature

`RegisterTaskDefinition` and `RunTask`: resolve the AWS-supported Secrets Manager selector syntax in `containerDefinitions[].secrets[].valueFrom`:

```text
arn:aws:secretsmanager:region:account-id:secret:secret-name:json-key:version-stage:version-id
```

This is the selector support deferred from #1624 / #1687.

## AWS Documentation

- https://docs.aws.amazon.com/AmazonECS/latest/developerguide/secrets-envvar-secrets-manager.html
- https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Secret.html

## Why is this needed?

AWS CDK emits the selector form when an ECS container injects one JSON field from a Secrets Manager secret. Floci currently treats the complete suffixed value as the secret ID. The task stops even though the base secret exists, preventing otherwise valid ECS task definitions from starting.

A plain secret ARN works in the same environment, which isolates the failure to selector parsing.

## Reproduction

```bash
set -euo pipefail

ENDPOINT=http://localhost:4566
REGION=us-east-2
CLUSTER=ecs-secret-selector-repro
FAMILY=ecs-secret-selector-repro
SECRET_NAME=ecs-secret-selector-repro

SECRET_ARN=$(aws --endpoint-url "$ENDPOINT" --region "$REGION" \
  secretsmanager create-secret \
  --name "$SECRET_NAME" \
  --secret-string '{"token":"expected-value"}' \
  --query ARN \
  --output text)

aws --endpoint-url "$ENDPOINT" --region "$REGION" \
  ecs create-cluster --cluster-name "$CLUSTER"

CONTAINER_DEFINITIONS=$(jq -nc --arg value_from "$SECRET_ARN:token::" '[{
  name: "probe",
  image: "busybox:1.36",
  essential: true,
  command: ["sh", "-c", "sleep 60"],
  secrets: [{name: "PROBE", valueFrom: $value_from}]
}]')

aws --endpoint-url "$ENDPOINT" --region "$REGION" \
  ecs register-task-definition \
  --family "$FAMILY" \
  --network-mode bridge \
  --container-definitions "$CONTAINER_DEFINITIONS"

TASK_ARN=$(aws --endpoint-url "$ENDPOINT" --region "$REGION" \
  ecs run-task \
  --cluster "$CLUSTER" \
  --task-definition "$FAMILY:1" \
  --query 'tasks[0].taskArn' \
  --output text)

sleep 5

aws --endpoint-url "$ENDPOINT" --region "$REGION" \
  ecs describe-tasks \
  --cluster "$CLUSTER" \
  --tasks "$TASK_ARN" \
  --query 'tasks[0].{lastStatus:lastStatus,stoppedReason:stoppedReason}'
```

Actual result:

```json
{
  "lastStatus": "STOPPED",
  "stoppedReason": "ResourceInitializationError: unable to pull secrets or registry auth: arn:aws:secretsmanager:us-east-2:000000000000:secret:ecs-secret-selector-repro-XXXXXX:token::: Secrets Manager can't find the specified secret."
}
```

Expected result: the base secret is resolved, the `token` JSON field is injected as `PROBE=expected-value`, and the task reaches `RUNNING`.

Control result: registering the same task with `valueFrom` set to the plain `SECRET_ARN` reaches `RUNNING`.

## Environment

- Floci version / image tag: `floci/floci:nightly`
- Resolved image digest: `sha256:0fb28cbebe2bb3a11e2b1336d8eaf2e9bf2feb411550b3ebe89ede88ae29965c`
- AWS CLI version: AWS CLI v2
- How Floci is running: Docker
- Storage mode: persistent

- [ ] Yes
- [ ] No

hectorvent:
## Summary

`containerDefinitions[].secrets[].valueFrom` now supports the ECS Secrets Manager selector suffix (`:json-key:version-stage:version-id`), the form CDK emits for `ecs.Secret.fromSecretsManager(secret, field)` and `fromSecretsManagerVersion`. Previously the whole suffixed string was passed to Secrets Manager as the secret id, the lookup failed, and the task stopped with `ResourceInitializationError` even though the base secret existed.

Closes #1912 (deferred from #1624 / #1687).

- New `SecretsManagerSelector` parser mirrors the real ECS agent (amazon-ecs-agent `asmsecret.go`): the ARN resource is split on `:` and disambiguated by segment count — 2 = plain reference (passed through verbatim, preserving partial-ARN lookups), 5 = selector (base ARN rebuilt positionally; empty segments mean unset, so CDK's `:field::`, `:::version-id` and `::version-stage:` forms all work), anything else = the agent's exact invalid-ARN error surfaced in `stoppedReason`.
- JSON-key extraction matches the agent's `asm.go`: top-level key only, missing key and non-JSON secrets fail the launch with the agent's message, scalars are injected as text, objects/arrays as JSON.
- `SecretsManagerService.getSecretValue` now validates the VersionId + VersionStage pair (both supplied must name the same version → `InvalidRequestException`, matching botocore's documentation and moto/LocalStack behavior). Reachable for the first time via the selector.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Behavioral reference is the real ECS agent (vendored `amazon-ecs-agent`, `agent/taskresource/asmsecret/asmsecret.go` + `agent/asm/asm.go`); the wire model is unaffected (botocore/Smithy/CFN schema all leave `valueFrom` an unconstrained string, so no RegisterTaskDefinition-time validation was added — selectors round-trip verbatim and errors surface only at launch, as on AWS).

Verified end to end against the exact reproduction from #1912 (AWS CLI v2): `{"token":"expected-value"}` secret + `valueFrom = <arn>:token::` → task RUNNING with `PROBE=expected-value` injected; CDK's field-less versioned form `<arn>:::versionId` → whole SecretString injected; plain-ARN control still RUNNING; malformed 3-/4-/6-segment forms stop the task with the agent's "an invalid ARN format for the AWS Secrets Manager secret was specified" text.

Deliberate divergences from the agent (all improvements, commented in code): a SecretBinary-only secret with a json-key fails cleanly with `ResourceInitializationError` where the agent nil-derefs; numeric JSON values inject in plain notation (`10000000`) where Go's `%v` would print `1e+07`; and object/array json-key values inject JSON (`{"a":1}`) where the agent's `%v` would inject Go's unparseable map rendering (`map[a:1]`).

## Checklist

- [x] `./mvnw test` passes locally (full suite 8338 tests, 0 failures, 0 errors, 8 skipped; 120/120 focused re-run on the rebased branch)
- [x] New or updated integration test added (`SecretsManagerSelectorTest`, `EcsContainerManagerSecretsTest`, `SecretsManagerServiceTest`, `EcsIntegrationTest`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/container/SecretsManagerSelector.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/container/SecretsManagerSelector.java
@@ -49,7 +49,9 @@ record SecretsManagerSelector(String secretId, String jsonKey, String versionSta
      * Extracts a top-level JSON field from a secret string, mirroring the agent's behavior
      * (asm.go): missing key and non-JSON secrets are errors. Scalar values are rendered with
      * {@code asText()}, which keeps large numbers in plain notation where the Go agent's
-     * {@code %v} would print scientific notation; objects and arrays serialize as JSON.
+     * {@code %v} would print scientific notation. Object and array values deliberately serialize
+     * as JSON, another divergence: the agent's {@code %v} on the unmarshalled value would inject
+     * Go's native rendering ({@code map[a:1 b:2]}), which no consumer can parse back.
      */
     static String extractJsonKey(String secretString, String jsonKey) {
         JsonNode root;
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/container/SecretsManagerSelector.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/container/SecretsManagerSelector.java
@@ -1,0 +1,80 @@
+package io.github.hectorvent.floci.services.ecs.container;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.hectorvent.floci.core.common.AwsArnUtils;
+import io.github.hectorvent.floci.core.common.AwsException;
+
+/**
+ * The ECS task-definition secret selector for Secrets Manager {@code valueFrom} values:
+ * {@code arn:aws:secretsmanager:region:account:secret:name:json-key:version-stage:version-id}.
+ *
+ * <p>Parsing mirrors the real ECS agent (amazon-ecs-agent, asmsecret.go): the ARN's resource
+ * segment is split on {@code :} and disambiguated purely by segment count, which is safe
+ * because a secret name can never contain a colon. Exactly two segments
+ * ({@code secret:name}) is a plain reference and the original string is used verbatim as the
+ * secret id, preserving partial-ARN lookups. Exactly five segments is a selector: the secret
+ * id is rebuilt from the first two segments, and empty selector segments mean "unset" — CDK
+ * routinely emits forms like {@code :field::}, {@code :::version-id} and
+ * {@code ::version-stage:}. Any other count is the agent's invalid-ARN error.
+ */
+record SecretsManagerSelector(String secretId, String jsonKey, String versionStage, String versionId) {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final String INVALID_ARN_MESSAGE =
+            "an invalid ARN format for the AWS Secrets Manager secret was specified. "
+                    + "Specify a valid ARN and try again.";
+
+    static SecretsManagerSelector parse(String valueFrom) {
+        AwsArnUtils.Arn arn;
+        try {
+            arn = AwsArnUtils.parse(valueFrom);
+        } catch (IllegalArgumentException e) {
+            throw invalidArn();
+        }
+        String[] segments = arn.resource().split(":", -1);
+        if (segments.length == 2) {
+            return new SecretsManagerSelector(valueFrom, null, null, null);
+        }
+        if (segments.length != 5) {
+            throw invalidArn();
+        }
+        String baseArn = new AwsArnUtils.Arn(arn.partition(), arn.service(), arn.region(),
+                arn.accountId(), segments[0] + ":" + segments[1]).toString();
+        return new SecretsManagerSelector(baseArn,
+                emptyToNull(segments[2]), emptyToNull(segments[3]), emptyToNull(segments[4]));
+    }
+
+    /**
+     * Extracts a top-level JSON field from a secret string, mirroring the agent's behavior
+     * (asm.go): missing key and non-JSON secrets are errors. Scalar values are rendered with
+     * {@code asText()}, which keeps large numbers in plain notation where the Go agent's
+     * {@code %v} would print scientific notation; objects and arrays serialize as JSON.
+     */
+    static String extractJsonKey(String secretString, String jsonKey) {
+        JsonNode root;
+        try {
+            root = MAPPER.readTree(secretString);
+        } catch (Exception e) {
+            root = null;
+        }
+        if (root == null || !root.isObject()) {
+            throw new AwsException("InvalidParameterException",
+                    "the secret value is not valid JSON, cannot extract json key " + jsonKey, 400);
+        }
+        JsonNode value = root.get(jsonKey);
+        if (value == null) {
+            throw new AwsException("ResourceNotFoundException",
+                    "retrieved secret from Secrets Manager did not contain json key " + jsonKey, 400);
+        }
+        return value.isValueNode() ? value.asText() : value.toString();
+    }
+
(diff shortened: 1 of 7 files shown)


src/main/java/io/github/hectorvent/floci/services/ecs/container/SecretsManagerSelector.java
```diff
@@ -0,0 +1,80 @@
...
+        if (root == null || !root.isObject()) {
+            throw new AwsException("InvalidParameterException",
+                    "the secret value is not valid JSON, cannot extract json key " + jsonKey, 400);
+        }
+        JsonNode value = root.get(jsonKey);
+        if (value == null) {
+            throw new AwsException("ResourceNotFoundException",
+                    "retrieved secret from Secrets Manager did not contain json key " + jsonKey, 400);
+        }
+        return value.isValueNode() ? value.asText() : value.toString();
```
pgermosen, line 72: This PR is careful to call out two deliberate divergences from the real agent (SecretBinary nil-deref, numeric scientific notation) in both the code comments and the PR description. There's a third one here that isn't mentioned anywhere and isn't covered by `SecretsManagerSelectorTest` (which only exercises string/number/boolean json-key values, never an object or array): when the selected JSON field is itself an object or array, this returns proper JSON via Jackson's `toString()`. The real agent's `GetSecretFromASMWithInput` (`asm.go`) does `fmt.Sprintf("%v", secretValue)` on the `interface{}` from `json.Unmarshal`, which for a nested map/slice prints Go's native formatting — e.g. `map[a:1 b:2]`, not `{"a":1,"b":2}` — into the container's environment variable.

So a task selecting a nested-object field via the selector gets a different literal string injected than real ECS would produce. Given how precisely everything else here tracks the agent (down to exact message text), this looks like an oversight rather than a choice — worth either documenting as a third deliberate improvement alongside the other two (simplest: it's arguably still better UX than replicating Go's map-printing quirk) or adding a test that pins the current behavior so it's a decision on record rather than an accident. Not blocking either way.

pgermosen (approved):
Looks good

- **Segment-count disambiguation and error text**: pulled `agent/taskresource/asmsecret/asmsecret.go` from `aws/amazon-ecs-agent` directly. `asmARNResourceFormat = "secret:{secretID}"` (2 segments) and `asmARNResourceWithParametersFormat = "secret:secretID:jsonKey:versionStage:versionID"` (5 segments) match this PR's `segments.length == 2` / `segments.length != 5` exactly, the field order (`paramValues[2]`=jsonKey, `[3]`=versionStage, `[4]`=versionId) matches `SecretsManagerSelector.parse` exactly, and the invalid-ARN message text (`"an invalid ARN format for the AWS Secrets Manager secret was specified. Specify a valid ARN and try again."`) is a verbatim match.
- **JSON-key extraction**: pulled `agent/asm/asm.go`. The real agent unmarshals into `map[string]interface{}` (top-level object only, matching this PR's `!root.isObject()` rejection) and its missing-key message (`"retrieved secret from Secrets Manager did not contain json key %s"`) is a verbatim match to this PR's.
- **VersionId/VersionStage pairing**: confirmed against the actual AWS API model (`botocore/data/secretsmanager/2017-10-17/service-2.json`) that the docstring says exactly "If you include both this parameter and VersionStage, the two parameters must refer to the same secret version" and that `InvalidRequestException` is a declared `GetSecretValue` error. The exact message text (`"You provided a VersionStage that is not associated to the provided VersionId."`) is also a verbatim match to moto's `secretsmanager/exceptions.py`. The new check is correctly nested inside the `versionId != null` branch, so it only fires when both are actually supplied, matching the documented condition.

Just a minor issue

One inline finding below — a real, verified gap in an otherwise excellent fidelity effort, not a blocker.

hectorvent:
Good catch, and you read the intent correctly: it was an oversight in the divergence bookkeeping rather than a decision. The behavior itself was the one I would choose anyway, JSON into the environment variable is something a consumer can parse back, where the agent's %v rendering (map[a:1 b:2]) is not.

It is now a decision on record: the javadoc lists it as the third deliberate divergence alongside the binary and numeric ones, and a test pins both the object and array forms. The PR description's divergence list is updated to match.

PR #2135, opened 2026-08-06
fix(cognito): reject unparsable and expired refresh tokens in REFRESH_TOKEN_AUTH

### LINKED ISSUE #2113: [BUG] Cognito InitiateAuth REFRESH_TOKEN_AUTH accepts arbitrary refresh tokens

AbhigyaKrishna:
## Summary

`InitiateAuth` with `AuthFlow=REFRESH_TOKEN_AUTH` currently mints tokens for refresh tokens it cannot parse, cannot resolve to a user, or that have already expired. This PR makes that path reject those tokens with `NotAuthorizedException`, and fixes the unit mismatch that made refresh-token expiry a no-op everywhere.

Two distinct defects are addressed:

1. **Unparsable / unresolvable refresh tokens are accepted.** In `CognitoAuthFlowHandler.handleRefreshToken`, when `parseRefreshToken` returns `null` — or when `adminGetUser` throws and the `catch (AwsException ignored) { }` swallows it — control falls through to a block that fabricates tokens via `generateTokenString("access", "unknown", ...)`. The caller gets HTTP 200 and a JWT with `username: "unknown"`. Both paths now throw `NotAuthorizedException` / `Invalid Refresh Token` (400), and the fabrication fallback is deleted so the flow only ever returns tokens minted for a real, resolved user.

2. **Refresh tokens never expire.** `buildRefreshToken` writes `System.currentTimeMillis()` into `parts[3]`, but `isRefreshTokenExpired` read that value as epoch **seconds** and added a lifetime in seconds to it, then compared against `System.currentTimeMillis() / 1000L`. Because the millisecond value is ~1000× larger than the current epoch-seconds clock, the comparison is never true — the check silently never fires, on this path or on `GetTokensFromRefreshToken` (`CognitoService:1537`), which is its only other caller. `issuedAt` is now converted to seconds before the comparison, and `handleRefreshToken` calls the check (it previously did not call it at all).

Closes #2113

### Relationship to #2132

Open PR #2132 fixes defect (1) for the same issue, with a near-identical shape. Defect (2) — the expiry unit bug and the missing expiry check on the `InitiateAuth` path — is not covered there and is independent of it. If #2132 lands first, I'm happy to rebase this down to just the expiry fix; maintainers should feel free to take whichever ordering is less churn.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior before this change**, against `InitiateAuth` / `REFRESH_TOKEN_AUTH`:

| Input | Floci before | Real Cognito / after |
|---|---|---|
| `REFRESH_TOKEN=not-a-real-refresh-token` | 200 + `AuthenticationResult`, JWT `username: "unknown"` | 400 `NotAuthorizedException` |
| Well-formed token for a user that does not exist in the pool | 200 + fabricated tokens | 400 `NotAuthorizedException` |
| Token issued past the client's `RefreshTokenValidity` | 200 + fresh tokens (check never fired) | 400 `NotAuthorizedException` / `Refresh Token has expired` |

Real Amazon Cognito rejects all three with `NotAuthorizedException` and never returns an `AuthenticationResult`. Error code, message, and the 400 status match the codes already used by the sibling `GetTokensFromRefreshToken` path in this codebase, so no new wire shapes are introduced.

The `NotAuthorizedException` for an unresolvable user is deliberate: real Cognito does not disclose whether the username encoded in a refresh token exists.

Verified with the AWS CLI reproduction from #2113 (`aws cognito-idp initiate-auth --auth-flow REFRESH_TOKEN_AUTH`) and via the RestAssured integration tests below.

## Checklist

Two integration tests added to `CognitoIntegrationTest`:

- `refreshWithUnparseableTokenIsRejectedNotFabricated` — garbage token must yield 400 `NotAuthorizedException`, not fabricated tokens.
- `refreshWithExpiredTokenIsRejected` — well-formed token with a long-past `issuedAt` must yield 400 `NotAuthorizedException` / `Refresh Token has expired`. This test fails on `main` today, since the expiry check never fires.

Test run on JDK 25 — full Cognito package, 310 tests, 0 failures, 0 errors (`CognitoIntegrationTest`: 67 tests, 0 failures). The branch merges cleanly into `upstream/main`.


src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
```diff
@@ -231,32 +231,32 @@ private Map<String, Object> handleRefreshToken(UserPool pool, UserPoolClient cli
...
-                auth.put("TokenType", "Bearer");
-                Map<String, Object> result = new HashMap<>();
-                result.put("AuthenticationResult", auth);
-                return result;
-            } catch (AwsException ignored) { }
+        if (parts == null) {
+            throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
+        }
+        String username = parts[1];
+        long iat = parts.length > 3 && !parts[3].isEmpty() ? Long.parseLong(parts[3]) : 0L;
```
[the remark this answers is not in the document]
AbhigyaKrishna: Good catch — confirmed and fixed in ff9c30db.

Reproduced it first: a Base64-decodable five-part token with `not-a-timestamp` in the issued-at slot returns **HTTP 500** with `NumberFormatException: For input string: "not-a-timestamp"`, not the 400 this flow promises. `parseRefreshToken` only checks the split arity, so the value reaches `Long.parseLong` unguarded.

Fixed by rejecting it as `NotAuthorizedException` / `Invalid Refresh Token` (400) rather than defaulting the timestamp to 0 — Floci never issues a token whose issued-at it cannot read, so continuing on to mint tokens for it would be the same fabrication bug this PR is closing.

For the record, the sibling `GetTokensFromRefreshToken` path (`CognitoService.java:1498`) already guards this parse, so it was the only unguarded site; I left that one alone as out of scope.

Added `refreshWithNonNumericIssuedAtIsRejected`, which fails with the 500 on the pre-fix code. Full Cognito package now at 311 tests, 0 failures on JDK 25.

Note these tests rely on the class's `@Order` setup state, so they must be run as the full class rather than in isolation.

src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
```diff
@@ -231,32 +231,37 @@ private Map<String, Object> handleRefreshToken(UserPool pool, UserPoolClient cli
...
-                auth.put("ExpiresIn", service.getAccessTokenExpiresInSeconds(client));
-                auth.put("TokenType", "Bearer");
-                Map<String, Object> result = new HashMap<>();
-                result.put("AuthenticationResult", auth);
-                return result;
-            } catch (AwsException ignored) { }
+        if (parts == null) {
+            throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
+        }
+        String username = parts[1];
```
pgermosen, line 244: **This method never validates that the token's embedded pool id matches the requesting pool — and per AWS's own documentation, that matters more here than it looks, because real Cognito refresh tokens can't be inspected or forged the way this emulator's can.**

From AWS's ["Understanding user pool JSON web tokens" guide](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-with-identity-providers.html): *"Amazon Cognito issues tokens as base64url-encoded strings. You can decode any Amazon Cognito ID or access token from base64url to plaintext JSON. **Amazon Cognito refresh tokens are encrypted, opaque to user pools users and administrators, and can only be read by your user pool.**"* Floci's refresh tokens are the opposite of that: `buildRefreshToken` (`CognitoService.java:2165-2170`) produces plain `base64(poolId|username|clientId|issuedAtMillis|uuid)` — no encryption, no signature, trivially decodable and constructible by anyone.

`parseRefreshToken` returns `[poolId, username, clientId, issuedAt, nonce]` (`CognitoService.java:2183`), but `handleRefreshToken` only ever reads `parts[1]` (username), `parts[3]` (issuedAt) and `parts[4]` (uuid) — `parts[0]` (poolId) and `parts[2]` (clientId) are discarded (this PR even deletes the now-unused `tokenClientId` local that used to hold `parts[2]`). Compare the sibling `getTokensFromRefreshToken`, which this PR's own description cites as the parity target for error codes: it explicitly checks `if (!client.getUserPoolId().equals(poolId)) throw new AwsException("NotAuthorizedException", ...)` (`CognitoService.java:1489-1491`). `handleRefreshToken` has no equivalent.

And `validateRefreshTokenNotRevoked` only checks a *revocation denylist* (`revokedTokenStore`) — it never verifies the token was actually issued by this system in the first place. So the fix in this PR ("unresolvable username → reject") is currently the *only* thing standing between a hand-crafted token and a full token mint: anyone who knows a valid `ClientId` for a pool (routinely public/embedded client-side) and any username that exists in it can construct `base64("anything|<realUsername>|anything|<currentMillis>|<any-uuid>")` and receive real, valid Access/Id tokens for that user — no prior authentication of any kind required, since `adminGetUser` is scoped by the *request's* pool (from `ClientId`), not the token's own `parts[0]`. Real Cognito's opaque/encrypted refresh tokens make this class of forgery impossible by construction; floci's plaintext format doesn't.

I verified this isn't introduced by this PR — `main` has the exact same gap in this method (this PR only changes what happens when `adminGetUser` throws, not what's passed to it) — so I'm not holding this PR on it. But since this PR is already touching this exact block and explicitly cites the sibling method for parity, the cheap half of this (pool-id scoping) fits naturally here:

```suggestion
        String[] parts = service.parseRefreshToken(refreshToken);
        if (parts == null) {
            throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
        }
        if (!pool.getId().equals(parts[0])) {
            throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
        }
        String username = parts[1];
```

That alone doesn't close the deeper hole (an attacker who also knows/guesses the pool id defeats it, and pool ids aren't really secret either), but it matches the sibling path's existing behavior and real Cognito's documented error surface, at no cost. The real fix — actually encrypting/signing `buildRefreshToken`'s payload and verifying it in `parseRefreshToken`, matching what AWS documents — touches both this method and `getTokensFromRefreshToken`, which feels like its own follow-up issue rather than something to fold into this one silently.

pgermosen (dismissed):
Looks good.


One substantive but **non-blocking** finding below, now backed by an AWS doc citation that makes it more concrete than my first pass — not something I'd hold this PR on, since fixing it properly is a bigger change than this PR's scope, but worth surfacing since this PR touches the exact code path.

src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
AbhigyaKrishna: Thanks — this is a good find, and the AWS citation makes the case. Applied the pool-id check in b0cee048.

I verified the forgery is real rather than theoretical before fixing it. `refreshWithTokenFromAnotherPoolIsRejected` builds `base64("us-east-1_someOtherPool|<realUsername>|<realClientId>|<now>|<uuid>")` — a pool id that does not exist anywhere — and on the pre-fix code it returns **HTTP 200 with a full AuthenticationResult**. No prior authentication of any kind. It was the only failing test in the class, so the gap is exactly where you said it was and nothing else was masking it.

Took your suggestion as written, including the placement: the check goes before the expiry check, matching `getTokensFromRefreshToken`'s ordering at `CognitoService.java:1489` so both refresh paths reject a foreign-pool token identically.

One consequence worth flagging: my existing expired and malformed-issued-at tests were building tokens with a literal `"some-pool"`, so with the new check in front they would have short-circuited on pool-id and stopped testing what they were written for — passing for the wrong reason. Both now use the real pool id.

Full Cognito package: 312 tests, 0 failures on JDK 25.

On the deeper issue — agreed it is out of scope here, and I have not tried to sneak a partial version of it in. The pool-id check narrows the forgery to someone who also knows the pool id, which as you note is not really a secret. The actual fix is signing or encrypting `buildRefreshToken`'s payload and verifying it in `parseRefreshToken`, which changes the token format and touches both refresh paths plus `RevokeToken`. I called that out explicitly in the commit message so it does not get lost, and I am happy to open a follow-up issue for it — say the word if you would rather file it yourself with your AWS citation, since the framing there is yours.

pgermosen: Verified b0cee048 directly rather than just reading the diff: `pool.getId()` is guaranteed to equal `client.getUserPoolId()` (it's resolved as `service.describeUserPool(client.getUserPoolId())` two lines up the call chain), so this is semantically identical to `getTokensFromRefreshToken`'s check, not just superficially similar. Ran the full Cognito package myself: 272/272 in `io.github.hectorvent.floci.services.cognito` (69/69 in `CognitoIntegrationTest`), zero failures.

Two things worth calling out beyond just applying the suggestion:

- `refreshWithTokenFromAnotherPoolIsRejected` uses a *real* existing username with a *fake* pool id, and you confirmed it returns 200 with a full `AuthenticationResult` pre-fix — that's a real proof of exploitability, not a hypothetical.
- Catching that `refreshWithNonNumericIssuedAtIsRejected` would have kept "passing" post-fix for the wrong reason (same error type *and* message text as the new pool-id check) is the kind of test-integrity bug that's easy to miss even when you're the one adding the check. Good catch.

On the deeper issue: filed #2137 with the full writeup (AWS's "encrypted, opaque" citation, the forge-from-scratch reproduction, cause, and a suggested HMAC-signing approach touching both `buildRefreshToken`/`parseRefreshToken` and both refresh paths). Assigned to myself for now — thanks for surfacing it rather than trying to fold it in here.

hectorvent:
@AbhigyaKrishna This PR will collide with PR #2132.

AbhigyaKrishna:
@hectorvent Already mentioned about this in the pr description

pgermosen (commented):
@AbhigyaKrishna — following up, and with an apology for the delay: #2132 merged on 2026-08-11 (`3ce5041c`), which is what put this branch into conflict, and my approval was still sitting on `ff9c30db` from before that. I have dismissed it rather than leave a stale green tick on a branch that no longer applies cleanly.

**This still holds real value** — just not all of it. You called the overlap in your own description before anyone raised it, and your offer to trim was the right instinct. I checked `main` directly so you do not have to work out what survived:

### 1. Unparsable / unresolvable tokens — now redundant, drop it

`main` already rejects a null `parseRefreshToken`, a non-numeric `parts[3]`, and a `UserNotFoundException` from `adminGetUser`, and the `username: "unknown"` fabrication fallback is gone. #2132 covered this half.

### 2. The expiry unit bug — still live in `main`, please keep

This is the part nobody else has fixed, and it is the more interesting of the two defects you found:

```java
// buildRefreshToken — writes MILLISECONDS into parts[3]
long issuedAt = System.currentTimeMillis();

// isRefreshTokenExpired — reads them back as SECONDS
long issuedAt = Long.parseLong(parts[3]);
long expiresAt = issuedAt + resolveTokenLifetimeSeconds(client, "refresh");
return System.currentTimeMillis() / 1000L >= expiresAt;
```

The left-hand value is roughly a thousand times larger than the clock it is compared against, so the condition can never be true. Refresh tokens simply do not expire today — not on `GetTokensFromRefreshToken`, which does call the check, and not on `InitiateAuth`, which still does not call it at all. Your fix addresses both halves: the `/ 1000L` conversion and the missing call.

### 3. Pool scoping — also still missing from `main`, please keep

```java
if (!pool.getId().equals(parts[0])) {
    throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
}
```

`main`'s `handleRefreshToken` uses the pool resolved from the request's `ClientId` throughout and never checks it against the pool encoded in the token, so a refresh token minted for pool A is accepted against pool B's client whenever that username exists in both. #2132 did not touch this.

Worth flagging specifically because it is **not** in the offer you made — you proposed trimming down to "just the expiry fix", and this would be lost in that trim. It arrived in `b0cee048`, after the description was written, which is probably why.

## What I would like

A rebase onto current `main` keeping **2 and 3** and dropping **1**. That should also resolve the conflicts, since they come from #2132 having rewritten the same block for the part you are removing.

Please keep the two integration tests. `refreshWithExpiredTokenIsRejected` in particular is the one that fails on `main` today, and it is the evidence that the expiry check was dead rather than merely untested — that is worth preserving in the history regardless of how the rest is trimmed. A case for the cross-pool rejection would be a good addition alongside it, since that behaviour currently has nothing pinning it.

Sorry again that this sat long enough for the ground to move under it — that is on the review queue, not on you. The analysis in your description was accurate throughout, including predicting exactly this situation.

AbhigyaKrishna:
@pgermosen Done — rebased onto current `main` (`5657e350`) and trimmed to exactly what you laid out. No apology needed; your prediction of this situation in the description was on the nose.

**Dropped #1** — `main` already rejects null-parse, non-numeric `parts[3]`, and `UserNotFoundException`, and the `"unknown"` fabrication fallback is gone. #2132's version is kept as-is; I didn't re-litigate it.

**Kept #2 (expiry)** — both halves: the millis→seconds conversion in `isRefreshTokenExpired`, and the call on the `InitiateAuth REFRESH_TOKEN_AUTH` path, which still didn't invoke it.

**Kept #3 (pool scoping)** — the `pool.getId().equals(parts[0])` guard, placed before the expiry check to match `getTokensFromRefreshToken`'s ordering.

**One thing that moved the ground again:** #2139 (HMAC-signed refresh tokens) merged on 08-09, after your comment. That's mostly good news — it closes the forge-from-scratch hole from your earlier finding. But it changes the pool-scoping story and broke my old tests:

- Pool scoping is **still needed and still missing from `main`** — HMAC proves a token was signed by `parts[0]`'s pool, not that `parts[0]` is the pool serving *this* request. A token *legitimately* minted for pool A (validly signed) is still accepted against pool B's client whenever the username exists in both. HMAC doesn't touch that; the `parts[0]` check does.
- My original integration tests hand-crafted unsigned 5-part tokens. Post-#2139 those get rejected at HMAC verification in `parseRefreshToken` before ever reaching the expiry or pool logic — so they'd have passed for the wrong reason (or asserted the wrong message). I moved both tests to `CognitoServiceTest`, next to #2139's own refresh-token security tests, where the `signRawRefreshToken` helper and pool secret let me build genuinely-valid tokens.

Both new tests fail on `main` **for the right reason** — I re-verified by reverting only the production change: `refreshTokenAuthFlowExpiredTokenThrows` and `refreshTokenAuthFlowRejectsTokenFromAnotherPool` each *mint tokens* (nothing thrown) without the fix, and reject with it. The cross-pool test uses two real pools each with an `alice` and a real refresh token from pool B, so it's the actual replay, not a fake pool id short-circuiting on lookup.

Full Cognito suite: 336 tests, 0 failures on JDK 25. Conflicts with #2132 are gone since the rebase drops the block it rewrote.

One heads-up: `Ec2PublicIpOnLaunchTest` currently fails to compile on `main` (an `Ec2Service` ctor gained a `ManagedPrefixList` backend that the test wasn't updated for — looks like #2023 and the prefix-list PR crossing). Unrelated to this PR, but it'll trip CI's test compile until fixed; flagging in case it's news.

pgermosen (approved):
Approving. This is exactly the shape I asked for, and the red CI is not yours.

## The failing check is inherited, not caused by this PR

Your run [31783641375](https://github.com/floci-io/floci/actions/runs/31783641375) failed at 08:22 with:

```
[INFO] 1 error
```

`1 error`, in an EC2 test, with nothing Cognito in it. `main` itself was not compiling at the time — #2287, fixed by #2288, which merged at 11:04. Your branch simply inherited a broken base.

So I rebuilt it locally on top of repaired `main` (`62ff4906`) rather than waiting on CI:

- `test-compile` — clean
- **Cognito suite: 294 tests, 0 failures** (`CognitoServiceTest` 128, `CognitoIntegrationTest` 71, plus the OAuth/UserInfo/JsonHandler/LambdaTriggers/StandardAttributes suites)

A re-run once this is rebased on current `main` should go green. Nothing here needs changing for it.

## The scope trim is right

You kept the two things that were actually missing from `main` and dropped the part #2132 had already covered, and the retitle reflects that honestly. For the record of what each half fixes:

**The expiry check was dead twice over.** `buildRefreshToken` writes `System.currentTimeMillis()` — milliseconds — and `isRefreshTokenExpired` compared that against a seconds clock, so it was off by ~1000x and could never fire. On top of that, the `InitiateAuth` REFRESH_TOKEN_AUTH path never called it at all. Both halves needed fixing, and you fixed both.

I checked the unit change against every consumer of `parts[3]`, because that was the risk. It is contained correctly: `buildRefreshToken` is the only mint site, both callers of `isRefreshTokenExpired` consume from it, and you converted *inside* the method rather than at the call sites — which matters, because `validateUserNotGloballySignedOut` sniffs magnitude (`iat > 1000000000000L`) to accept either unit and still needs raw millis. Converting a step earlier would have broken global sign-out silently. The narrower change was the correct one.

**The pool-scoping check closes a real replay gap.** HMAC verification only proves `parts[0]`'s pool signed the token, not that `parts[0]` is the pool serving the request — your comment states that precisely. `getTokensFromRefreshToken` already had the equivalent guard, so this brings the two paths into line rather than inventing a rule.

## The tests hold under mutation

I didn't just run them — I reverted each change in a throwaway tree and confirmed the right test fails:

| Reverted | Result |
|---|---|
| ms→s conversion in `isRefreshTokenExpired` | `refreshTokenAuthFlowExpiredTokenThrows` fails |
| the `isRefreshTokenExpired` call in the handler | `refreshTokenAuthFlowExpiredTokenThrows` fails |
| the `pool.getId().equals(parts[0])` check | `refreshTokenAuthFlowRejectsTokenFromAnotherPool` fails |

Each test pins its own change, and the expiry test catches *both* halves of that bug independently — so neither can silently regress.

Particular credit for how the cross-pool test is built: two pools each with an `alice`, so username resolution cannot be what rejects the token, and a validly-signed foreign token is left with only the pool-id check standing between it and a fresh token set. That is the version that actually proves the guard, and it's the harder one to write.

Thank you for the patience on this one — the delay was queue and base-branch churn, not anything about the work.

PR #2139, opened 2026-08-06
fix(cognito): HMAC-sign refresh tokens to prevent forgery

### LINKED ISSUE #2137: [BUG] Cognito refresh tokens are plaintext and forgeable, unlike AWS's encrypted/opaque tokens
## Service

Cognito

## AWS API Action

`InitiateAuth` (`REFRESH_TOKEN_AUTH`), `GetTokensFromRefreshToken`

## Expected behavior

Per AWS's own documentation ([Understanding user pool JSON web tokens](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-with-identity-providers.html)):

> Amazon Cognito issues tokens as base64url-encoded strings. You can decode any Amazon Cognito ID or access token from base64url to plaintext JSON. **Amazon Cognito refresh tokens are encrypted, opaque to user pools users and administrators, and can only be read by your user pool.**

A refresh token should be unforgeable: nothing outside the user pool's own backend can construct one that will be accepted.

## Actual behavior

Floci's refresh tokens are an unsigned, unencrypted plaintext string:

```java
// CognitoService.java, buildRefreshToken
String raw = poolId + "|" + username + "|" + clientId + "|" + issuedAt + "|" + familyId;
return Base64.getEncoder().withoutPadding().encodeToString(raw.getBytes(StandardCharsets.UTF_8));
```

`parseRefreshToken` only validates structural shape (segment count, valid base64/UTF-8) — nothing verifies the token was ever actually issued by this system. `validateRefreshTokenNotRevoked` checks a *revocation denylist* keyed by `jti`, which only rejects tokens that were explicitly revoked; it is not an issuance record and does nothing for a token that was never real to begin with.

Before #2135, this meant anyone who knew (a) a pool's `ClientId` (routinely public/embedded client-side) and (b) a username that exists in that pool could construct:

```
base64("<anyPoolId>|<realUsername>|<anyClientId>|<currentMillis>|<any-uuid>")
```

...and receive real, valid Access/Id tokens for that user via `InitiateAuth REFRESH_TOKEN_AUTH`, with no prior authentication of any kind.

#2135 closed two of the three checks needed to accept a token (unresolvable username, and — in a follow-up commit — the token's embedded pool id not matching the requesting pool), which raises the bar considerably: an attacker must now also know the target's correct pool id, and it must resolve to a real user in that pool. But the underlying weakness remains — a token is still just three known/guessable strings and a timestamp, base64-encoded, with nothing tying it back to a real prior authentication event.

## Reproduction

Using a pool/client you know the IDs for, and a username known to exist in that pool:

```bash
POOL_ID=us-east-1_XXXXXXXXX
CLIENT_ID=xxxxxxxxxxxxxxxxxxxxxxxxxx
USERNAME=alice@example.com

RAW="${POOL_ID}|${USERNAME}|${CLIENT_ID}|$(date +%s%3N)|$(uuidgen)"
FORGED_TOKEN=$(printf '%s' "$RAW" | base64 | tr -d '=\n')

aws --endpoint-url http://localhost:4566 cognito-idp initiate-auth \
  --client-id "$CLIENT_ID" \
  --auth-parameters "REFRESH_TOKEN=$FORGED_TOKEN"
```

Expected: `NotAuthorizedException` (a refresh token that was never issued should never validate).
Actual (as of the fixes in #2135): rejected *only if* `POOL_ID` or `USERNAME` don't match a real pool/user — if both are correctly known/guessed, this mints a real `AuthenticationResult`.

## Cause

- `CognitoService.buildRefreshToken` (around line 2165) — plain `poolId|username|clientId|issuedAtMillis|uuid`, no HMAC, no encryption.
- `CognitoService.parseRefreshToken` (around line 2177) — structural validation only (segment count + base64/UTF-8 decodability), no authenticity check.
- `CognitoService.validateRefreshTokenNotRevoked` (around line 2321) — denylist-only; not an issuance record.

## Suggested fix

HMAC-sign (or encrypt) the payload in `buildRefreshToken` with a server-side secret, and verify the signature/decrypt in `parseRefreshToken` before trusting any decoded field. This should preserve the existing pipe-delimited field semantics internally so downstream logic (expiry check, revocation check, username/pool-id extraction) doesn't need to change beyond adding the verification gate at the top of `parseRefreshToken` — a token that fails verification should be treated the same as `parseRefreshToken` returning `null` today (`NotAuthorizedException` / "Invalid Refresh Token").

Touches both `CognitoAuthFlowHandler#handleRefreshToken` (`InitiateAuth`/`REFRESH_TOKEN_AUTH`) and `CognitoService#getTokensFromRefreshToken`, since both trust `parseRefreshToken`'s output today.

## Related

Originally raised as a non-blocking review finding on #2135, which fixed the two narrower gaps (unresolvable username, foreign pool id) that made forgery trivial without any target knowledge. This issue tracks the deeper architectural gap those fixes deliberately left out of scope — see the discussion and the reply from @AbhigyaKrishna on that PR.

- [x] Yes
- [ ] No

pgermosen:
## Summary

- Refresh tokens were a plain `poolId|username|clientId|issuedAt|familyId` string, base64-encoded with no signature or encryption. Anyone who knew a pool id, client id, and a valid username could hand-craft a token accepted by `InitiateAuth REFRESH_TOKEN_AUTH` / `GetTokensFromRefreshToken` — unlike real Cognito, which documents refresh tokens as encrypted and opaque to everyone but the user pool.
- `buildRefreshToken` now appends an `HmacSHA256` signature over the payload, keyed by a random secret generated and persisted per pool (`UserPool.signingSecret`, lazily generated via `ensureRefreshTokenSecret`, mirroring the existing RSA JWT signing-key pattern in `ensureJwtSigningKeys`).
- `parseRefreshToken` verifies that signature with a constant-time comparison (`MessageDigest.isEqual`) before trusting any decoded field, and rejects the token — same as today's null-parse contract (`NotAuthorizedException` / "Invalid Refresh Token") — on any signature mismatch, tampering, or unknown pool id.

Closes #2137.

### Relationship to #2135

#2135 (open, unmerged) closes two narrower gaps in `handleRefreshToken` (unresolvable username, foreign pool id) but its own commit message explicitly deferred this exact issue: *"Signing or encrypting buildRefreshToken's payload touches both refresh paths and belongs in its own change."* This PR is that follow-up. It's branched off `main` independently to stay focused (per the one-fix-per-PR guideline) — the HMAC check alone already closes the issue's forgery reproduction regardless of #2135's merge status. There may be minor rebase overlap in `parseRefreshToken`/`handleRefreshToken` whichever PR merges second.

## Test plan

- [x] `./mvnw test -Dtest=CognitoServiceTest` — 116 tests pass, including 3 new regression tests: unsigned forged token, tampered username post-signing, swapped pool id post-signing — all correctly rejected with `NotAuthorizedException`.
- [x] `./mvnw test -Dtest=CognitoIntegrationTest` — 67 tests pass.
- [x] `./mvnw test` (full suite) — 8391/8392 pass; the sole failure (`NeptuneOpenCypherIntegrationTest`) is an unrelated local Docker image-pull timeout, confirmed by re-running in isolation and by `git diff --stat` showing no touched files under `services/neptune/`.
- [x] Existing hand-crafted-token tests (`refreshTokenIsStructuredAndDecodable`, `getTokensFromRefreshTokenExpiredTokenThrows`) updated to account for the new trailing signature segment / to sign their crafted payloads.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
@@ -231,32 +231,23 @@ final class CognitoAuthFlowHandler {
         String refreshToken = params.get("REFRESH_TOKEN");
         if (refreshToken == null) throw new AwsException("InvalidParameterException", "REFRESH_TOKEN is required", 400);
         String[] parts = service.parseRefreshToken(refreshToken);
-        if (parts != null) {
-            String username = parts[1];
-            String tokenClientId = parts[2];
-            long iat = parts.length > 3 && !parts[3].isEmpty() ? Long.parseLong(parts[3]) : 0L;
-            String refreshTokenUuid = parts.length > 4 ? parts[4] : null;
-            
-            // Check revocation before issuing new tokens
-            service.validateRefreshTokenNotRevoked(refreshTokenUuid, pool.getId(), username, iat);
-            
-            try {
-                CognitoUser user = service.adminGetUser(pool.getId(), username);
-                CognitoService.ClaimsOverride override = firePreTokenGeneration(pool, client, user,
-                        clientMetadata, "TokenGeneration_RefreshTokens");
-                Map<String, Object> auth = new HashMap<>();
-                auth.put("AccessToken", service.generateSignedJwt(user, pool, "access", client, override, refreshTokenUuid));
-                auth.put("IdToken", service.generateSignedJwt(user, pool, "id", client, override, refreshTokenUuid));
-                auth.put("ExpiresIn", service.getAccessTokenExpiresInSeconds(client));
-                auth.put("TokenType", "Bearer");
-                Map<String, Object> result = new HashMap<>();
-                result.put("AuthenticationResult", auth);
-                return result;
-            } catch (AwsException ignored) { }
+        if (parts == null) {
+            throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
         }
+        String username = parts[1];
+        String tokenClientId = parts[2];
+        long iat = parts.length > 3 && !parts[3].isEmpty() ? Long.parseLong(parts[3]) : 0L;
+        String refreshTokenUuid = parts.length > 4 ? parts[4] : null;
+
+        // Check revocation before issuing new tokens
+        service.validateRefreshTokenNotRevoked(refreshTokenUuid, pool.getId(), username, iat);
+
+        CognitoUser user = service.adminGetUser(pool.getId(), username);
+        CognitoService.ClaimsOverride override = firePreTokenGeneration(pool, client, user,
+                clientMetadata, "TokenGeneration_RefreshTokens");
         Map<String, Object> auth = new HashMap<>();
-        auth.put("AccessToken", service.generateTokenString("access", "unknown", pool, client));
-        auth.put("IdToken", service.generateTokenString("id", "unknown", pool, client));
+        auth.put("AccessToken", service.generateSignedJwt(user, pool, "access", client, override, refreshTokenUuid));
+        auth.put("IdToken", service.generateSignedJwt(user, pool, "id", client, override, refreshTokenUuid));
         auth.put("ExpiresIn", service.getAccessTokenExpiresInSeconds(client));
         auth.put("TokenType", "Bearer");
         Map<String, Object> result = new HashMap<>();
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
@@ -235,8 +235,12 @@ final class CognitoAuthFlowHandler {
             throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
         }
         String username = parts[1];
-        String tokenClientId = parts[2];
-        long iat = parts.length > 3 && !parts[3].isEmpty() ? Long.parseLong(parts[3]) : 0L;
+        long iat;
+        try {
+            iat = parts.length > 3 && !parts[3].isEmpty() ? Long.parseLong(parts[3]) : 0L;
+        } catch (NumberFormatException e) {
+            throw new AwsException("NotAuthorizedException", "Invalid Refresh Token", 400);
+        }
         String refreshTokenUuid = parts.length > 4 ? parts[4] : null;
 
         // Check revocation before issuing new tokens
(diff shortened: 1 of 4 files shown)


pgermosen:
Good catch — pushed a fix in 7bf104be.

`handleRefreshToken` was falling through to `service.generateTokenString(..., "unknown", ...)` whenever `parseRefreshToken` returned null, and that helper produces a fully valid, correctly RS256-signed JWT (real per-pool key, correct issuer/audience) — not a dummy value. Since this PR's HMAC verification means *every* forged/tampered token now takes that null path (previously only structurally-malformed garbage did), that fallback would have kept minting real signed tokens for exactly the forgery attempts this PR is meant to close.

`handleRefreshToken` now throws `NotAuthorizedException`/"Invalid Refresh Token" on a null parse result, matching the behavior `getTokensFromRefreshToken` already had. Added `refreshTokenAuthRejectsForgedTokenInsteadOfIssuingUnknownUserTokens` covering the `InitiateAuth REFRESH_TOKEN_AUTH` path specifically, since the existing regression tests only exercised `GetTokensFromRefreshToken`. Full Cognito unit + integration suites still green (117 + 67).

pgermosen:
Pushed 84d6c9af addressing the two new inline findings on \`CognitoAuthFlowHandler.java\`:

- **Unused \`tokenClientId\` (line 238)** — removed; it was parsed but never validated (pre-existing, not introduced by this PR, but fine to clean up here).
- **Missing catch of \`NumberFormatException\` (line 239)** — \`Long.parseLong(parts[3])\` is now wrapped and rejected as \`NotAuthorizedException\`/"Invalid Refresh Token", consistent with the rest of the method. Added \`refreshTokenAuthRejectsNonNumericIssuedAt\` covering it.

The other inline comment ("Invalid tokens reach fallback issuance" on \`CognitoService.java:2279\`) points at the \`return null;\` inside \`parseRefreshToken\`'s signature-mismatch branch — i.e. the correct rejection path — so it looks like a stale thread left over from the pre-fix review pass rather than a new finding; the bot's own updated top-level summary already reflects the \`handleRefreshToken\` fix and rates this 5/5 safe to merge.

Cognito unit + integration suites still green (118 + 67 = 185).

hectorvent:
Thanks for splitting this out of #2135 and writing up the forgery reproduction so precisely; that made it easy to follow.

1. The core fix looks right to me. `NotAuthorizedException` is declared on both `InitiateAuth` and `GetTokensFromRefreshToken` in the model ([cognito-idp service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json)), so rejecting there stays inside the contract, and dropping the "unknown" user fallback in `services/cognito/CognitoAuthFlowHandler.java` was the important half.

2. **(follow-up, separate PR)** `parseRefreshToken` now requires the signed 6 field form, so tokens minted before an upgrade stop parsing in persistent mode. Harmless for a local emulator, but maybe worth one line in the Cognito docs.

3. On the open inline thread at `CognitoService.java`, your read is correct: that `return null` is the rejection path, not a fallback.

No blockers from my side.

hectorvent (approved):
Thanks @pgermosen

PR #2146, opened 2026-08-06
fix(cloudformation): expand AWS::Serverless::HttpApi via SAM transform

### LINKED ISSUE #1759: [BUG] SAM transform doesn't expand `AWS::Serverless::HttpApi` — reports CREATE_COMPLETE with a synthesized physical ID, no API actually created

rogueserenity:
## Summary

Fixes #1759.

`SamTransformProcessor`'s switch on SAM resource types only handled `AWS::Serverless::Function`, `AWS::Serverless::SimpleTable`, and `AWS::Serverless::Api` (REST, v1) — `AWS::Serverless::HttpApi` fell into the default branch and was silently dropped from the expanded template, along with any `Auth.Authorizers`/`DefaultAuthorizer` config and any `HttpApi`-typed Events on a sibling Function. A stack using `AWS::Serverless::HttpApi` reported `CREATE_COMPLETE` with a synthesized physical ID, but no real `AWS::ApiGatewayV2::Api` was ever provisioned.

Builds on #1760 (now merged) — the JWT authorizer this transform synthesizes needs `AWS::ApiGatewayV2::Authorizer` CloudFormation provisioning to actually exist once expanded.

## What changed

- `SamTransformProcessor` now expands `AWS::Serverless::HttpApi` into a real `AWS::ApiGatewayV2::Api` plus its `$default` auto-deploy stage, one `AWS::ApiGatewayV2::Authorizer` per `Auth.Authorizers` entry, and — for every Function event bound to that API via explicit `ApiId: {Ref: ...}` — an `AWS_PROXY` Integration, a Route wired to `Auth.DefaultAuthorizer` when present, and a `Lambda::Permission`. Implicit HttpApi (no `ApiId`, SAM auto-creates the API) is intentionally out of scope, mirroring how this file already separates `generateImplicitApi`/`collectApiRoutes` from `expandServerlessApi`'s explicit-resource handling.
- Per-event `Auth: { Authorizer: NONE }` is honored on individual `HttpApi` events, overriding the API's `DefaultAuthorizer` for that route specifically (SAM's documented per-event opt-out).
- The synthesized authorizer forwards SAM's `Authorizers.<Name>.IdentitySource` (scalar or array) instead of hardcoding `$request.header.Authorization`, so a template configuring a querystring-based identity source — supported end-to-end for raw CFN templates since #1760 — isn't silently downgraded to header-only when expanded from SAM.
- `buildHttpApiAuthorizer` now matches `_validate_jwt_authorizer` in `samtranslator/model/apigatewayv2.py` on the two things real `sam deploy` rejects a JWT authorizer for: an absent/empty `IdentitySource`, and an absent `JwtConfiguration`. Both were previously silent no-ops here, so a template real SAM refuses would pass quietly and deploy an authorizer that would never validate a token.
- `Issuer`/`Audience` lookup inside `JwtConfiguration` is now a true case-insensitive key match (`fieldIgnoreCase`), mirroring `_get_jwt_configuration`'s `{k.lower(): v for k, v in props.items()}` in the same translator file. Previously only the exact lowercase spelling was read, so a template written in the uppercase `AWS::ApiGatewayV2::Authorizer` style (or any other casing) silently produced an authorizer with an empty `JwtConfiguration` instead of an error.

## Test plan

- [x] `createStack_samHttpApiExpandsToRealApiRouteAndAuthorizer`: creates a SAM template with a `HttpApi`-triggered function and a JWT authorizer, asserts a real API/route/integration/stage/authorizer are all provisioned and wired correctly.
- [x] `createStack_samHttpApiPerEventAuthNoneOverridesDefaultAuthorizer`: one route picks up the default authorizer, a second opts out via `Auth: {Authorizer: NONE}`; asserts only the first is wired to the authorizer.
- [x] `createStack_samHttpApiAuthorizerHonorsCustomIdentitySource`: asserts a scalar `IdentitySource` in the SAM template is forwarded rather than overridden with the header default. Verified by reverting the fix and confirming this test fails.
- [x] `createStack_samHttpApiAuthorizerMissingIdentitySourceIsRejected`: asserts `CreateStack` fails with `IdentitySource` in the error when an authorizer omits it, matching real `sam deploy`. Verified by reverting the fix and confirming this test fails.
- [x] `createStack_samHttpApiAuthorizerMissingJwtConfigurationIsRejected`: same, for an authorizer that omits `JwtConfiguration` entirely. Verified by reverting the fix and confirming this test fails.
- [x] `createStack_samHttpApiAuthorizerJwtConfigurationKeysAreCaseInsensitive`: uses `AUDIENCE` and `Issuer` — neither the documented lowercase nor the documented uppercase spelling — to prove a real case-insensitive fold rather than a check against two hardcoded spellings. Verified by reverting the fix (back to matching only the two documented spellings) and confirming this test fails.
- [x] Full `CloudFormationIntegrationTest` suite (112 tests) passes.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.github.hectorvent.floci.core.common.AwsException;
 import org.jboss.logging.Logger;
 
 import java.util.ArrayList;
@@ -258,7 +259,7 @@ class SamTransformProcessor {
         authProps.put("Name", authorizerName);
         authProps.put("AuthorizerType", "JWT");
 
-        authProps.set("IdentitySource", resolveIdentitySource(samAuthorizer));
+        authProps.set("IdentitySource", resolveIdentitySource(authorizerName, samAuthorizer));
 
         JsonNode jwtConfig = samAuthorizer.path("JwtConfiguration");
 
@@ -283,19 +284,21 @@ class SamTransformProcessor {
      * Resolves the SAM authorizer's {@code IdentitySource}, accepting either the documented array
      * form or a single scalar string — mirroring
      * {@code CloudFormationResourceProvisioner.resolveIdentitySource}, since the raw
-     * {@code AWS::ApiGatewayV2::Authorizer} resource this expands to accepts both. Defaults to
-     * {@code $request.header.Authorization} when the SAM template doesn't specify one, matching
-     * SAM's own default for JWT authorizers.
+     * {@code AWS::ApiGatewayV2::Authorizer} resource this expands to accepts both. SAM itself
+     * rejects a JWT authorizer with no {@code IdentitySource} ({@code _validate_jwt_authorizer} in
+     * samtranslator/model/apigatewayv2.py), so an absent or empty value is a template error here
+     * too.
      */
-    private ArrayNode resolveIdentitySource(JsonNode samAuthorizer) {
+    private ArrayNode resolveIdentitySource(String authorizerName, JsonNode samAuthorizer) {
         ArrayNode identitySource = objectMapper.createArrayNode();
         JsonNode raw = samAuthorizer.path("IdentitySource");
         if (raw.isTextual()) {
             identitySource.add(raw.asText());
-        } else if (raw.isArray()) {
+        } else if (raw.isArray() && !raw.isEmpty()) {
             raw.forEach(v -> identitySource.add(v.asText()));
         } else {
-            identitySource.add("$request.header.Authorization");
+            throw new AwsException("ValidationError",
+                    "OAuth2 Authorizer must define 'IdentitySource'; authorizer " + authorizerName, 400);
         }
         return identitySource;
     }
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
@@ -147,7 +147,13 @@ class SamTransformProcessor {
         return routes;
     }
 
-    private record HttpApiRoute(String functionLogicalId, String apiLogicalId, String path, String httpMethod) {}
+    /**
+     * @param noAuthorizer true when this event declares {@code Auth: {Authorizer: NONE}},
+     *                     overriding the API's {@code DefaultAuthorizer} for this route specifically
+     *                     (SAM's documented per-event opt-out of the default authorizer).
+     */
+    private record HttpApiRoute(String functionLogicalId, String apiLogicalId, String path, String httpMethod,
+                                boolean noAuthorizer) {}
 
     /**
      * Collects HttpApi-typed Function events bound to an explicit {@code AWS::Serverless::HttpApi}
@@ -185,7 +191,8 @@ class SamTransformProcessor {
                 }
                 JsonNode methodNode = p.path("Method");
                 String method = methodNode.isTextual() ? methodNode.asText() : "ANY";
-                routes.add(new HttpApiRoute(logicalId, apiLogicalId, pathNode.asText(), method));
+                boolean noAuthorizer = "NONE".equals(p.path("Auth").path("Authorizer").asText(null));
+                routes.add(new HttpApiRoute(logicalId, apiLogicalId, pathNode.asText(), method, noAuthorizer));
             }
         }
         return routes;
@@ -295,7 +302,7 @@ class SamTransformProcessor {
         ObjectNode routeProps = objectMapper.createObjectNode();
         routeProps.set("ApiId", ref(apiLogicalId));
         routeProps.put("RouteKey", routeKey);
-        if (defaultAuthorizerLogicalId != null) {
+        if (defaultAuthorizerLogicalId != null && !route.noAuthorizer()) {
             routeProps.put("AuthorizationType", "JWT");
             routeProps.set("AuthorizerId", ref(defaultAuthorizerLogicalId));
         } else {
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/SamTransformProcessor.java
@@ -259,25 +259,44 @@ class SamTransformProcessor {
         authProps.put("Name", authorizerName);
         authProps.put("AuthorizerType", "JWT");
 
+        JsonNode jwtConfig = samAuthorizer.path("JwtConfiguration");
+        if (!jwtConfig.isObject()) {
+            throw new AwsException("ValidationError",
+                    "OAuth2 Authorizer must define 'JwtConfiguration'; authorizer " + authorizerName, 400);
+        }
+
         authProps.set("IdentitySource", resolveIdentitySource(authorizerName, samAuthorizer));
 
-        JsonNode jwtConfig = samAuthorizer.path("JwtConfiguration");
-
-        if (jwtConfig.isObject()) {
-            ObjectNode jwtProps = objectMapper.createObjectNode();
-            JsonNode issuer = jwtConfig.path("issuer");
-            if (!issuer.isMissingNode()) {
-                jwtProps.set("Issuer", issuer.deepCopy());
-            }
-            JsonNode audience = jwtConfig.path("audience");
-            if (audience.isArray()) {
-                jwtProps.set("Audience", audience.deepCopy());
-            }
-            authProps.set("JwtConfiguration", jwtProps);
+        ObjectNode jwtProps = objectMapper.createObjectNode();
+        JsonNode issuer = fieldIgnoreCase(jwtConfig, "issuer");
+        if (!issuer.isMissingNode()) {
+            jwtProps.set("Issuer", issuer.deepCopy());
         }
+        JsonNode audience = fieldIgnoreCase(jwtConfig, "audience");
+        if (audience.isArray()) {
+            jwtProps.set("Audience", audience.deepCopy());
+        }
+        authProps.set("JwtConfiguration", jwtProps);
 
         authDef.set("Properties", authProps);
         return authDef;
+    }
+
+    /**
+     * SAM's {@code _get_jwt_configuration} lower-cases every key of the {@code JwtConfiguration}
+     * map before reading it ({@code {k.lower(): v for k, v in props.items()}} in
+     * samtranslator/model/apigatewayv2.py), so any casing of {@code issuer}/{@code audience} is
+     * accepted, not just the two spellings the docs happen to show.
+     */
+    private JsonNode fieldIgnoreCase(JsonNode node, String fieldName) {
+        Iterator<Map.Entry<String, JsonNode>> fields = node.fields();
+        while (fields.hasNext()) {
+            Map.Entry<String, JsonNode> field = fields.next();
+            if (field.getKey().equalsIgnoreCase(fieldName)) {
+                return field.getValue();
+            }
+        }
+        return objectMapper.missingNode();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


rogueserenity:
@hectorvent this one is ready for you to check out

hectorvent:
Thank you for this, and sorry for the slow reply after your ping. The structure is good, and splitting `Authorizer: NONE` and the `IdentitySource` forwarding into separate commits with a revert and confirm check made this easy to follow.

**(blocking)** The javadoc on `resolveIdentitySource` says defaulting to `$request.header.Authorization` matches SAM's own default for JWT authorizers, but SAM has no such default. It rejects the template instead, in `_validate_jwt_authorizer`: "OAuth2 Authorizer must define 'IdentitySource'" ([samtranslator/model/apigatewayv2.py](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/apigatewayv2.py)). So a template that real `sam deploy` refuses will pass here quietly, which is the expensive direction for an emulator to be lenient in.

Correcting the comment is the minimum. Matching SAM and rejecting it would be better, and either way that branch is worth a test, since the scalar and `NONE` cases are covered but the absent case is not.

The scalar handling itself is right, and that part of the comment holds.

rogueserenity:
@hectorvent Fixed in 8867b843 — corrected the javadoc (SAM doesn't default `IdentitySource`, it rejects the template via `_validate_jwt_authorizer`) and went with the stronger fix: `resolveIdentitySource` now throws a `ValidationError` when `IdentitySource` is absent or empty, matching real `sam deploy`. Added `createStack_samHttpApiAuthorizerMissingIdentitySourceIsRejected` for the previously-untested absent case, verified it fails against the old code, and updated the two existing tests that were relying on the old silent default to set `IdentitySource` explicitly. Full `CloudFormationIntegrationTest` suite (110 tests) passes.

pgermosen (commented):
@rogueserenity — @hectorvent's blocking point is properly closed. You corrected the javadoc, took the stronger of the two options he offered rather than the minimum, and added the missing-case test. I verified it: reverting the rejection back to the silent header default fails `createStack_samHttpApiAuthorizerMissingIdentitySourceIsRejected`.

I also checked the SAM claim from source rather than inherit it, because the [property reference](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-httpapi-oauth2authorizer.html) actually says `IdentitySource` is **Required: No**, which reads like it contradicts a hard rejection. The translator disagrees with its own docs, so your implementation is right and the docs are the unreliable half here.

Verified on current `main` (`30a61646`): **929 tests green** across the cloudformation and apigatewayv2 packages, and no interaction with #1961, which touched `SamTransformProcessor` on the Function path and merged earlier today.

Two small things in the same method before I approve. Both come out of the exact function Hector pointed you at.

## 1. `JwtConfiguration` is the other half of that validator

`_validate_jwt_authorizer` has two checks, and `JwtConfiguration` is the first one:

```python
def _validate_jwt_authorizer(self) -> None:
    if not self.jwt_configuration:
        raise InvalidResourceException(..., "OAuth2 Authorizer must define 'JwtConfiguration'.")
    if not self.id_source:
        raise InvalidResourceException(..., "OAuth2 Authorizer must define 'IdentitySource'.")
```

`buildHttpApiAuthorizer` currently treats it as optional — `if (jwtConfig.isObject())`, and nothing when it is absent. So an authorizer with no `JwtConfiguration` still passes here, which is the same failure Hector described: *"a template that real `sam deploy` refuses will pass here quietly."*

Worth noting the ordering too. SAM checks `JwtConfiguration` first, so for a template missing both, real SAM reports the `JwtConfiguration` error while Floci reports the `IdentitySource` one. Rejecting in the same order keeps the message matching as well as the outcome.

## 2. `Issuer` and `Audience` are case-insensitive in SAM

This one is the more likely to bite, because it fails silently rather than loudly. The code reads only the lowercase forms:

```java
JsonNode issuer = jwtConfig.path("issuer");
JsonNode audience = jwtConfig.path("audience");
```

But the SAM docs are explicit:

> Properties `issuer` and `audience` are case insensitive and can be used either lowercase as in OpenAPI or uppercase `Issuer` and `Audience` as in [AWS::ApiGatewayV2::Authorizer](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigatewayv2-authorizer-jwtconfiguration.html).

The uppercase spelling is the natural one for anyone coming from raw CloudFormation, and it is what the linked `AWS::ApiGatewayV2::Authorizer` page uses. Today such a template produces an authorizer with an empty `JwtConfiguration` and no error — and because of point 1 there is nothing to catch it, so the two compound: the authorizer is created, deploys clean, and simply never validates a token the way the author intended.

A fallback on each lookup covers it.

## Not blocking

Neither is a defect in what you set out to build, and I would not have raised them at all if they were not in the method you are already editing. If you would rather ship this and take them separately, say so and I will approve as-is and file them — I mention it only because they are a few lines each and the context is in your head right now, which it will not be in three weeks.

The rest reads well. Collecting the HttpApi routes before either side expands, so Function and HttpApi expansion order does not matter, is the right call and is the sort of thing that would otherwise surface as an intermittent test failure much later.

rogueserenity:
@pgermosen Both fixed, same method as before.

**1. `JwtConfiguration` required** — `buildHttpApiAuthorizer` now checks it before `IdentitySource`, matching `_validate_jwt_authorizer`'s order, and rejects with the same message shape SAM uses. Added `createStack_samHttpApiAuthorizerMissingJwtConfigurationIsRejected`; verified it fails against the old code.

**2. `Issuer`/`Audience` case-insensitivity** — went back to the actual translator source rather than trust the doc wording. `_get_jwt_configuration` does `{k.lower(): v for k, v in props.items()}` — a real fold over every key, not just the lowercase/uppercase pair the docs happen to show. So I replaced the two-spelling check with `fieldIgnoreCase`, a genuine case-insensitive key lookup. Renamed the test to `createStack_samHttpApiAuthorizerJwtConfigurationKeysAreCaseInsensitive` and swapped one fixture key to `AUDIENCE` (neither documented spelling) so it actually proves the fold instead of just re-covering `Issuer`/`Audience`. Verified it fails against a two-spelling-only implementation.

Full `CloudFormationIntegrationTest` suite (112 tests) passes. PR description updated to reflect both fixes and their verification.

pgermosen (approved):
@rogueserenity — approving. Both points are closed, and the second one you fixed better than I asked for.

## The case-insensitivity fix is more correct than my suggestion

I asked for an uppercase `Issuer`/`Audience` fallback, reasoning from the docs line that names those two spellings. You went to the translator instead and found `_get_jwt_configuration` normalises *every* key, which I confirmed — its docstring says it outright:

> Make sure that JWT configuration dict keys are lower case.

So any casing is valid, not only the two the documentation happens to show. `fieldIgnoreCase` is the right generalisation; the fallback I proposed would have accepted `Issuer` and `issuer` and still silently dropped `ISSUER`, which is the same class of bug one spelling further out.

That is twice on this PR that the SAM property reference has been the unreliable source and reading the implementation gave the correct answer — the other being `IdentitySource` documented as *Required: No* while `_validate_jwt_authorizer` rejects it outright. Worth remembering for this corner of the codebase.

## Ordering

Moving the `JwtConfiguration` check ahead of `resolveIdentitySource` matters for more than tidiness: SAM checks it first, so a template missing both now gets the same error from Floci as it would from `sam deploy`, rather than the right rejection with the wrong reason.

## Verified

On current `main` (`30a61646`): **931 tests green** across `services.cloudformation.**` and `services.apigatewayv2.**`.

Both new behaviours are pinned:

| Reverted | Failure |
|---|---|
| the `JwtConfiguration` requirement | `createStack_samHttpApiAuthorizerMissingJwtConfigurationIsRejected` |
| `fieldIgnoreCase` → lowercase-only | `createStack_samHttpApiAuthorizerJwtConfigurationKeysAreCaseInsensitive` |

Together with the earlier `IdentitySource` work, the authorizer path now matches `_validate_jwt_authorizer` on both checks, in the same order, with the same messages — and the silent-failure mode that started this thread is gone: an authorizer can no longer be built with an empty `JwtConfiguration` and no error.

## The rest

@hectorvent's original point about silent leniency is now genuinely closed rather than narrowly answered, which was what I was after.

The structural choice I flagged earlier still reads as the right one: collecting HttpApi routes before either the Function or the HttpApi resource expands, so expansion order between them cannot matter. That is the kind of thing that otherwise shows up much later as an order-dependent test failure and takes a day to attribute.

Thank you for turning both of these around overnight, and for checking the source rather than taking my suggested fix at face value. The result is better for it.

PR #2149, opened 2026-08-06
fix(apigateway): fall back to REST API region scan for non-SigV4 Authorization headers

rogueserenity:
## Summary

Unsigned requests (and requests carrying a non-SigV4 Authorization header, e.g.
a Cognito bearer JWT) resolve to `RegionResolver`'s configured default region
rather than a region the caller specified. For v1 REST APIs, `dispatch()`
previously only fell back to `resolveRestApiRegion`'s cross-region scan when
the `Authorization` header was missing or blank — so a REST API deployed
outside the default region still 404'd for any request authenticated with a
non-SigV4 header (e.g. a bearer JWT), because the old check saw a non-blank
header and skipped the scan.

**Note:** this PR originally also added HTTP API (v2) cross-region resolution
(`ApiGatewayV2Service#resolveHttpApiRegion`, a new `RegionResolver#isRegionUnresolved`).
#2054 merged independently on 2026-08-12 and brought in equivalent v2 coverage
(`resolveApiRegion`, the `getApi` retry, `dispatchV2` hand-off) ahead of this
branch. After rebasing onto `main`, the v2 portion of this diff was redundant
and was dropped — see 1b3c9c5. What remains here is scoped to the v1 REST path
only.

## What changed

- `ApiGatewayExecuteController#dispatch` now treats "the Authorization header
  didn't resolve a real region" (via `RegionResolver#resolveRegionFromAuthOrNull`)
  as the trigger for the v1 REST API region scan, instead of just "header is
  missing or blank." This is a strict widening — every request that previously
  triggered the scan still does.
- `ApiGatewayV2Service#resolveApiRegion` gets a doc comment; no behavioral
  change there (already covered by #2054).

## Test plan

- [x] `nonSigV4AuthorizationHeaderFallsBackToRestApiRegionScan`: a Cognito-style
  bearer JWT on a v1 REST API request triggers the region scan (uses a real
  `RegionResolver`, not a mock, to exercise the actual header-parsing logic).
  Reverting the fix makes this test fail.
- [x] `unsignedRequestFindsV2ApiDeployedOutsideDefaultRegion` and
  `signedV2RequestDoesNotConsultRegionFallback` are kept as the only
  controller-level coverage of the v2 cross-region dispatch that #2054
  introduced — they pin that behavior rather than regression-guard this PR's
  change (commented as such in the test file). Both pass unchanged whether
  this PR's fix is present or reverted.
- [x] Full `apigateway` + `apigatewayv2` test suite passes.


rogueserenity:
@hectorvent can you approve CI on this please?

hectorvent:
@rogueserenity there are some incidents open https://www.githubstatus.com/.

rogueserenity:
> @rogueserenity there are some incidents open https://www.githubstatus.com/.

I just saw that. Thanks for the heads up.

rogueserenity:
@hectorvent Went ahead and resolved the `isRegionUnresolved` / `resolveRegionFromAuthOrNull` duplication you flagged on #2169 rather than leaving it to land separately — since #2169 merged before this one, `resolveRegionFromAuthOrNull` was already on `main` by the time I rebased this branch. Dropped `isRegionUnresolved` from `RegionResolver` and switched `ApiGatewayExecuteController#dispatch` to use `resolveRegionFromAuthOrNull(...) == null` instead (1b3c9c5). Tests updated accordingly.

pgermosen (commented):
Thanks for this — and for handling the collision with #2054 cleanly when it landed underneath you. The code here is correct and I'm not asking for any change to it. What I'd like before merge is a title and description correction, because the PR now describes work that isn't in the diff any more.

## What happened

You opened this on 2026-08-06. #2054 merged on 08-12 and brought in the v2 fallback independently — `resolveApiRegion`, the `getApi` retry, and the `dispatchV2` hand-off are all on `main` already. You merged main on 08-13 and correctly dropped your duplicate `RegionResolver.isRegionUnresolved` in `1b3c9c54`, which was exactly the right call.

The result is that the surviving diff is much smaller than the description:

- `ApiGatewayExecuteController` — one condition swapped
- `ApiGatewayV2Service` — a javadoc comment, no new method
- the tests

The description still says it "Adds `ApiGatewayV2Service#resolveHttpApiRegion`" and "Adds `RegionResolver#isRegionUnresolved`", and that `dispatch()` "now consults the fallback scan for v2". None of those are in the diff today.

## Why the title matters

The remaining change doesn't affect the v2 path at all. `preferredRegion` is captured before your line mutates `region`, so the v2 branch runs `resolveApiRegion(preferredRegion, apiId)` identically either way. I traced a bearer-JWT request to a cross-region HTTP API through both versions and the outcome is the same, because that path always scans regardless of how the region was resolved.

What you *are* fixing is real, and it's on the **v1 REST** path: a REST API deployed outside the default region, reached with a non-SigV4 `Authorization` header (a Cognito bearer JWT), previously 404'd because the old `auth == null || auth.isBlank()` check saw a non-blank header and skipped the region scan. That's a genuine bug and swapping in `resolveRegionFromAuthOrNull(...) == null` is the right fix — I checked it's a strict widening, so no request that previously got the scan loses it.

So the ask is:

- **Retitle to `fix(apigateway): ...`** — something like *"fall back to the REST API region scan for non-SigV4 Authorization headers"*. The squash-merge title becomes permanent changelog history, and a v1 fix shipped under `fix(apigatewayv2)` will send the next person bisecting this area in the wrong direction.
- **Update the description** to describe the one-line change, and note that the v2 work is now covered by #2054.

## One note on the test plan

The description says reverting the fix makes 2 of the 3 new tests fail. I reverted it locally and measured 1 — only `nonSigV4AuthorizationHeaderFallsBackToRestApiRegionScan`. The other two exercise the same path in both versions, so they pass on `main` unchanged. I'd assume that line was accurate when you wrote it against the pre-#2054 version and simply didn't get revisited.

**Please keep all three tests regardless.** `unsignedRequestFindsV2ApiDeployedOutsideDefaultRegion` and `signedV2RequestDoesNotConsultRegionFallback` turn out to be the only controller-level coverage of the v2 cross-region dispatch that #2054 introduced — the sole other reference to `resolveApiRegion` in the test tree is a stub override inside a fake. They're pinning someone else's change, which is a useful thing to have and would be a shame to lose to a tidy-up. Just worth a comment saying so, so they don't read as regression guards for this PR.

## Verified locally

On current `main` (`62ff4906`): compiles clean, and **837 tests pass** across `apigateway` + `apigatewayv2`. Nothing here is blocked on correctness — purely the metadata.

Also worth calling out that using a real `RegionResolver` rather than a mock in `nonSigV4AuthorizationHeaderFallsBackToRestApiRegionScan` is what makes that test meaningful; it's the one that actually exercises the header parsing, and it's the one that caught the revert.

rogueserenity:
@pgermosen  Retitled to `fix(apigateway): ...` and rewrote the description to match the surviving v1-only diff, with a corrected test-plan claim (1 test fails on revert, not 2). Also added a comment above `unsignedRequestFindsV2ApiDeployedOutsideDefaultRegion` and `signedV2RequestDoesNotConsultRegionFallback` noting they pin #2054's v2 behavior rather than regression-guard this PR (5fa27269). All three tests kept as-is. Thanks for tracing it through — appreciate the thoroughness.

pgermosen (approved):
Approving. All three points addressed, and the way you handled the test-plan correction was better than what I asked for.

## What changed

**The title now matches the diff** — `fix(apigateway): fall back to REST API region scan for non-SigV4 Authorization headers`. That is the change, and the changelog will read correctly.

**The body carries the history rather than just dropping the stale claims.** Recording that the v2 work was originally here, that #2054 landed equivalent coverage independently on 2026-08-12, and that the redundant portion went in `1b3c9c5` is more useful than a clean rewrite would have been. Anyone reading this later can reconstruct why the PR shrank, which is exactly the question a reviewer coming to it cold would otherwise have to dig through the commit log to answer.

**You corrected the test plan instead of deleting it.** It now states that `unsignedRequestFindsV2ApiDeployedOutsideDefaultRegion` and `signedV2RequestDoesNotConsultRegionFallback` *"pass unchanged whether this PR's fix is present or reverted"*. That is the honest inversion of the earlier claim, and leaving the accurate version on the record is worth more than quietly removing the inaccurate one.

**And `5fa27269` puts it in the code**, where it will actually be seen. The comment marking those two tests as pinning #2054 rather than guarding this PR is the thing that stops someone deleting them in a future tidy-up on the reasonable-looking grounds that they do not test the change they sit next to.

## Verified

On current `main` (`12154297`): **837 tests green** across `services.apigateway.**` and `services.apigatewayv2.**`.

I re-ran the mutation to confirm your corrected claim rather than assume it. Reverting the fix to `main`'s old `auth == null || auth.isBlank()` check fails **exactly one** test:

```
ApiGatewayExecuteControllerTest.nonSigV4AuthorizationHeaderFallsBackToRestApiRegionScan
Tests run: 19, Failures: 1
```

One, not two, and the other two stay green — precisely as the updated test plan now says.

The fix itself remains what it was: a strict widening of the trigger from "header missing or blank" to "header did not resolve a real region", so no request that previously got the scan loses it, and a REST API outside the default region reached with a bearer JWT now resolves instead of 404ing. Using the real `RegionResolver` rather than a mock in that test is what makes it meaningful — it is the one that exercises the actual header parsing, and the one that catches the revert.

Thank you for turning this around quickly, and for treating a metadata correction as worth doing properly rather than a formality.

PR #2150, opened 2026-08-06
fix(apigateway): match Content-Type case-insensitively in Lambda proxy responses

rogueserenity:
## Summary

`buildProxyResponse` re-derives the response Content-Type with an exact-case JSON field lookup (`node.path("headers").path("Content-Type")`), even though the header-copy loop just above it already handles arbitrary casing correctly by iterating every field. Lambda proxy integrations commonly return lowercased header names — e.g. the AWS Lambda Web Adapter sidecar returns `content-type`, not `Content-Type` — which the exact-case lookup misses, silently falling back to `application/json` regardless of what the Lambda actually set. HTTP header names are case-insensitive on the wire (RFC 7230 §3.2), so this affected every response whose handler didn't happen to use the exact AWS-console casing, most visibly every RFC 9457 `application/problem+json` error response.

Two related gaps surfaced during review and are folded into this PR rather than deferred, since they're the same bug in the same method:

- A JSON-null `Content-Type` value was coerced to the literal string `"null"` instead of falling back to the default, producing an invalid media type.
- `findHeaderIgnoreCase` only scanned `headers`, not `multiValueHeaders` — but Lambda proxy responses can set Content-Type via either, and API Gateway prefers `multiValueHeaders` on conflicts. The old code would silently overwrite a Lambda's real Content-Type with the `application/json` default whenever it arrived via `multiValueHeaders`.

## What changed

- Replaces the exact-case lookup with a case-insensitive scan (`findHeaderIgnoreCase`) reusing the already-fetched headers node.
- Widens `buildProxyResponse`'s visibility to package-private (matching `dispatch()`'s existing visibility) so it's directly unit-testable without standing up a full Lambda invoke.
- `findHeaderIgnoreCase` returns `Optional.empty()` for a JSON-null value instead of the string `"null"`.
- `findHeaderIgnoreCase` now also unwraps array values (the `multiValueHeaders` shape), and Content-Type resolution checks `multiValueHeaders` first, falling back to `headers` — matching API Gateway's own merge precedence.

## Test plan

- [x] `buildProxyResponseMatchesLowercaseContentType`: lowercase `content-type` header is matched.
- [x] `buildProxyResponseMatchesExactCaseContentType`: exact-case `Content-Type` still works (regression guard).
- [x] `buildProxyResponseMatchesMixedCaseContentType`: mixed-case `CONTENT-TYPE` is matched.
- [x] `buildProxyResponseDefaultsToJsonWhenNoContentTypeHeaderPresent`: absent header still defaults to `application/json`.
- [x] `buildProxyResponseDefaultsToJsonWhenContentTypeIsJsonNull`: JSON-null header value defaults to `application/json` rather than the literal string `"null"`.
- [x] `buildProxyResponseMatchesContentTypeInMultiValueHeaders`: Content-Type set only via `multiValueHeaders` is detected.
- [x] `buildProxyResponseMultiValueHeadersTakesPrecedenceOverHeaders`: when both `headers` and `multiValueHeaders` set Content-Type, `multiValueHeaders` wins.
- [x] Verified by reverting `findHeaderIgnoreCase` to the old exact-case lookup: the lowercase and mixed-case tests fail (2 of 4) while the exact-case and no-header tests keep passing, confirming none of the four are vacuous.
- [x] Verified by reverting the `multiValueHeaders` scan: the two new multi-value tests fail (2 of 2) against the array-unaware lookup, confirming they exercise the fix.
- [x] Full `apigateway` test suite passes.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -766,7 +766,7 @@ public class ApiGatewayExecuteController {
         }
     }
 
-    private Response buildProxyResponse(InvokeResult result) {
+    Response buildProxyResponse(InvokeResult result) {
         if (result.getPayload() == null || result.getPayload().length == 0) {
             return Response.status(result.getFunctionError() != null ? 502 : result.getStatusCode()).build();
         }
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -793,9 +793,7 @@ public class ApiGatewayExecuteController {
                 String bodyStr = bodyNode.asText();
                 boolean isBase64 = node.path("isBase64Encoded").asBoolean(false);
                 byte[] bytes = isBase64 ? Base64.getDecoder().decode(bodyStr) : bodyStr.getBytes();
-                String ct = MediaType.APPLICATION_JSON;
-                JsonNode ctNode = node.path("headers").path("Content-Type");
-                if (!ctNode.isMissingNode() && !ctNode.isNull()) ct = ctNode.asText();
+                String ct = findHeaderIgnoreCase(respHeaders, "Content-Type").orElse(MediaType.APPLICATION_JSON);
                 builder.entity(bytes).type(ct);
             }
             return builder.build();
@@ -803,6 +801,22 @@ public class ApiGatewayExecuteController {
             LOG.warnv("Failed to parse Lambda response: {0}", e.getMessage());
             return Response.status(502).entity(result.getPayload()).type(MediaType.APPLICATION_JSON).build();
         }
+    }
+
+    /**
+     * HTTP header names are case-insensitive on the wire (RFC 7230 §3.2), and Lambda proxy
+     * integrations commonly return lowercased names (e.g. the AWS Lambda Web Adapter emits
+     * "content-type", not "Content-Type"). A plain JsonNode#path lookup is exact-case and
+     * silently misses those, so Content-Type detection needs to scan case-insensitively.
+     */
+    private static java.util.Optional<String> findHeaderIgnoreCase(JsonNode headersNode, String name) {
+        if (headersNode == null || !headersNode.isObject()) return java.util.Optional.empty();
+        var it = headersNode.fields();
+        while (it.hasNext()) {
+            var e = it.next();
+            if (e.getKey().equalsIgnoreCase(name)) return java.util.Optional.of(e.getValue().asText());
+        }
+        return java.util.Optional.empty();
     }
 
     // ──────────────────────────── AWS (non-proxy) ────────────────────────────
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -810,11 +810,17 @@ public class ApiGatewayExecuteController {
      * silently misses those, so Content-Type detection needs to scan case-insensitively.
      */
     private static java.util.Optional<String> findHeaderIgnoreCase(JsonNode headersNode, String name) {
-        if (headersNode == null || !headersNode.isObject()) return java.util.Optional.empty();
+        if (headersNode == null || !headersNode.isObject()) {
+            return java.util.Optional.empty();
+        }
         var it = headersNode.fields();
         while (it.hasNext()) {
             var e = it.next();
-            if (e.getKey().equalsIgnoreCase(name)) return java.util.Optional.of(e.getValue().asText());
+            if (e.getKey().equalsIgnoreCase(name)) {
+                return e.getValue().isNull()
+                        ? java.util.Optional.empty()
+                        : java.util.Optional.of(e.getValue().asText());
+            }
         }
         return java.util.Optional.empty();
     }
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -44,6 +44,7 @@ import java.util.Base64;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
@@ -793,7 +794,9 @@ public class ApiGatewayExecuteController {
                 String bodyStr = bodyNode.asText();
                 boolean isBase64 = node.path("isBase64Encoded").asBoolean(false);
                 byte[] bytes = isBase64 ? Base64.getDecoder().decode(bodyStr) : bodyStr.getBytes();
-                String ct = findHeaderIgnoreCase(respHeaders, "Content-Type").orElse(MediaType.APPLICATION_JSON);
+                String ct = findHeaderIgnoreCase(multiHeaders, "Content-Type")
+                        .or(() -> findHeaderIgnoreCase(respHeaders, "Content-Type"))
+                        .orElse(MediaType.APPLICATION_JSON);
                 builder.entity(bytes).type(ct);
             }
             return builder.build();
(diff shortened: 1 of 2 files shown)


rogueserenity:
Both addressed in 1dcfbc34.

**Null Content-Type** — confirmed and reproduced: `findHeaderIgnoreCase` called `asText()` unconditionally, so a JSON-null `Content-Type` produced the literal string `"null"` instead of falling through to the `application/json` default the old exact-case lookup preserved (it explicitly checked `!ctNode.isNull()`). `Response.Builder#type("null")` throws, turning a 200 into a 502 with the raw payload. Fixed exactly as suggested, plus a regression test (`buildProxyResponseDefaultsToJsonWhenContentTypeIsJsonNull`) — verified it fails against the un-guarded version.

**Missing braces** — fixed, both conditionals in `findHeaderIgnoreCase` now use braces per AGENTS.md.

rogueserenity:
@hectorvent can you approve running CI on this please?

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
```diff
@@ -793,9 +793,7 @@ private Response buildProxyResponse(InvokeResult result) {
                 String bodyStr = bodyNode.asText();
                 boolean isBase64 = node.path("isBase64Encoded").asBoolean(false);
                 byte[] bytes = isBase64 ? Base64.getDecoder().decode(bodyStr) : bodyStr.getBytes();
-                String ct = MediaType.APPLICATION_JSON;
-                JsonNode ctNode = node.path("headers").path("Content-Type");
-                if (!ctNode.isMissingNode() && !ctNode.isNull()) ct = ctNode.asText();
+                String ct = findHeaderIgnoreCase(respHeaders, "Content-Type").orElse(MediaType.APPLICATION_JSON);
```
pgermosen: **`multiValueHeaders` has the identical bug, and I confirmed it silently overwrites the Lambda's Content-Type**

The header-copy loop reads **two** sources — `headers` (`:781`) and `multiValueHeaders` (`:784`) — but `findHeaderIgnoreCase(respHeaders, ...)` only ever scans the first. AWS documents the second as a fully valid way to return any header:

> The `multiValueHeaders` key can contain multi-value headers as well as single-value headers. **You can use the `multiValueHeaders` key to specify all of your extra headers, including any single-value ones.**

I probed `buildProxyResponse` directly rather than reasoning about it:

| payload | resulting media type |
|---|---|
| `headers: {content-type: application/problem+json}` | `application/problem+json` ✅ *(this PR)* |
| `multiValueHeaders: {Content-Type: [application/problem+json]}` | **`application/json`** ❌ |
| `multiValueHeaders: {content-type: [application/problem+json]}` | **`application/json`** ❌ |

And it's worse than "ignored": the loop *does* copy the header onto the builder, but `builder.type(ct)` then replaces it with the `application/json` default. The Lambda's Content-Type is actively destroyed — the same user-visible symptom your description opens with, reached by a different route.

**Precedence is also inverted.** With `headers: text/plain` and `multiValueHeaders: application/xml` I measured `text/plain`. AWS says the other way round:

> If you specify values for both `headers` and `multiValueHeaders`, API Gateway merges them into a single list. If the same key-value pair is specified in both, only the values from `multiValueHeaders` will appear in the merged list.

(That sentence is loosely worded — "the same key-value pair" — so I'd treat the conflict case as lower-confidence than the primary finding, which is unambiguous.)

**Suggested:** check `multiValueHeaders` too, preferring it per the doc — roughly

```java
String ct = findHeaderIgnoreCase(multiHeaders, "Content-Type")   // arrays: first element
        .or(() -> findHeaderIgnoreCase(respHeaders, "Content-Type"))
        .orElse(MediaType.APPLICATION_JSON);
```

with `findHeaderIgnoreCase` unwrapping a one-element array. That's a few lines on top of what you've already built, reuses the same helper, and your existing four tests extend to it directly.

Entirely reasonable to defer to a follow-up — it's pre-existing and not something this PR introduced. But it's the same bug, in the same method, found by the same reasoning, so folding it in now avoids a second round trip.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
```diff
@@ -805,6 +803,28 @@ private Response buildProxyResponse(InvokeResult result) {
         }
     }
 
+    /**
+     * HTTP header names are case-insensitive on the wire (RFC 7230 §3.2), and Lambda proxy
+     * integrations commonly return lowercased names (e.g. the AWS Lambda Web Adapter emits
+     * "content-type", not "Content-Type"). A plain JsonNode#path lookup is exact-case and
+     * silently misses those, so Content-Type detection needs to scan case-insensitively.
+     */
+    private static java.util.Optional<String> findHeaderIgnoreCase(JsonNode headersNode, String name) {
```
pgermosen: **`java.util.Optional` is the only fully-qualified name in the file** *(nit)*

The file imports nine `java.util` types explicitly (`ArrayList`, `Base64`, `HashMap`, `List`, `Map`, `UUID`, `ConcurrentHashMap`, `Matcher`, `Pattern`), and this is the single FQN usage anywhere in it — three occurrences, on the signature and both return paths.

**Suggested:** `import java.util.Optional;` and drop the qualifier. Purely cosmetic, but it reads as accidental rather than deliberate.

pgermosen (approved):
Looks good

Real bug, correct fix, unusually well tested. The diagnosis is exactly right: the header-copy loop iterates every field and so is casing-agnostic, while the Content-Type derivation three lines later used an exact-case path("Content-Type") — so any handler emitting content-type silently got application/json.

What I especially rate: the mutation check in your own description. Reverting findHeaderIgnoreCase and confirming exactly 2 of the new tests fail is what proves the tests aren't vacuous, and it's rare to see it done unprompted. Commit 3 does it again for the JSON-null regression.

A few comments below

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
rogueserenity: I went ahead and knocked this out if you want to look at those changes as well.

src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
rogueserenity: Fixed this one too

pgermosen (approved):
Looked, as asked. Both are done, and I verified rather than eyeballed: **323/323 apigateway tests pass** (up from 321 — your two new cases), and I re-ran the same probe that found the gap in the first place.

| payload | before | now |
|---|---|---|
| `multiValueHeaders: {Content-Type: [...]}` | `application/json` ❌ | `application/problem+json` ✅ |
| `multiValueHeaders: {content-type: [...]}` | `application/json` ❌ | `application/problem+json` ✅ |
| both (`headers` text/plain, `mvH` application/xml) | `text/plain` ❌ | `application/xml` ✅ |
| `headers` only (regression check) | `application/problem+json` | `application/problem+json` ✅ |

I also probed the edge cases the array unwrap newly introduces, since `get(0)` on an arbitrary JSON array is the kind of thing that bites later. All sensible, none of them crash:

| payload | result |
|---|---|
| `Content-Type: []` (empty array) | falls back to `application/json` |
| `Content-Type: [null]` | falls back to `application/json` |
| `Content-Type: null` (not an array) | falls back to `application/json` |
| `Content-Type: ["a/b","c/d"]` | `a/b` — first element, as documented |
| `mvH: {Content-Type: []}` **plus** `headers: {content-type: text/csv}` | **`text/csv`** |

That last one is the nicest property and I don't think it was an accident: because the empty array yields `Optional.empty()` rather than a present-but-blank value, the `.or(...)` chain correctly *defers to `headers`* instead of short-circuiting to the default. Worth keeping in mind if anyone ever refactors that method — it's load-bearing behavior that no test currently pins.

The `Optional` import is fixed too; the only remaining occurrence of `java.util.Optional` in the file is the import statement itself.

Nothing further from me — this is done as far as I'm concerned.

hectorvent:
Thank you, this is a nicely contained fix and the test table is the best part of it. The JSON-null case in particular would have turned into a literal `"null"` media type, and it is the kind of thing that only shows up once someone hits it in anger.

The precedence you picked is also right: AWS merges `headers` and `multiValueHeaders` and lets `multiValueHeaders` win on a key present in both ([proxy integration output format](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format)), which is exactly the order in your `.or()` chain.

No blockers from my side.

hectorvent (approved):

PR #2151, opened 2026-08-06
fix(s3): ensure tags are added on multipart s3 uploads

floverfelt:
## Summary

Fixes a bug where tags were not present on the final object in s3 when tagging was specified via CreateMultiPartUpload

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior before was that the final object uploaded to floci were not tagged when the upload was done via multi-part upload.

## Checklist


pgermosen (approved):
Looks good

Real bug, correct fix, and it reuses what already exists rather than inventing anything. AWS documents the header on this operation:

> **x-amz-tagging** — The tag-set for the object. The tag-set must be encoded as URL Query parameters.

So the gap was genuine: Floci accepted `x-amz-tagging` on `PutObject` but silently dropped it on `CreateMultipartUpload`, and the completed object came back untagged.

---

PR #2155, opened 2026-08-07
fix(iam): include policy description in GetPolicy/ListPolicies/CreatePolicy responses

### LINKED ISSUE #2154: GetPolicy/ListPolicies/CreatePolicy responses omit the IAM policy's Description field
## Bug

`GetPolicy`, `ListPolicies`, and `CreatePolicy` all omit the IAM policy's `Description` field from
their XML response, even when the policy was created with one. Real AWS's `GetPolicy` includes it
(see the [API docs](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetPolicy.html), whose
own sample response has `<Description>Allows read-only access to the example bucket</Description>`).

## Root cause

`src/main/java/io/github/hectorvent/floci/services/iam/IamQueryHandler.java`, method `policyXml()`:

```java
private String policyXml(IamPolicy p) {
    return new XmlBuilder()
            .elem("PolicyName", p.getPolicyName())
            .elem("PolicyId", p.getPolicyId())
            .elem("Arn", p.getArn())
            .elem("Path", p.getPath())
            .elem("DefaultVersionId", p.getDefaultVersionId())
            .elem("AttachmentCount", (long) p.getAttachmentCount())
            .elem("IsAttachable", true)
            .elem("CreateDate", isoDate(p.getCreateDate()))
            .elem("UpdateDate", isoDate(p.getUpdateDate()))
            .build();
}
```

There's no `.elem("Description", p.getDescription())` line, even though:
- `IamPolicy` (the model) correctly declares, constructs, and exposes `description` via
  `getDescription()`/`setDescription()`.
- `IamService.createPolicy()` correctly stores the description passed on `CreatePolicy`.
- `policyXml()` is shared by `handleCreatePolicy`, `handleGetPolicy`, and `handleListPolicies` — so
  the field is missing from all three response paths, not just one.

Notably, the sibling method `roleXml()` immediately above it *does* include
`.elem("Description", r.getDescription())` for roles — so roles serialize their description
correctly, and policies just don't. Looks like a one-line omission rather than a design choice.

## Reproduction

```
aws iam create-policy --policy-name test --policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:GetObject","Resource":"*"}]}' --description "test description" --endpoint-url <floci-endpoint>
aws iam get-policy --policy-arn <arn> --endpoint-url <floci-endpoint>
```

The `get-policy` response has no `Description` field at all, despite the policy being created with
one.

## Impact

Anyone running Terraform against Floci with an `aws_iam_policy` resource that sets `description`
sees Terraform's AWS provider treat the policy as perpetually drifted: every `terraform plan`/
`apply` shows `description` needing to be set, and since IAM policy descriptions are immutable
(force new resource on the AWS provider), every apply destroys and recreates the policy — and
everything attached to it — even when nothing in the underlying Terraform config changed.

## Suggested fix

Add the missing line to `policyXml()`:

```java
.elem("Description", p.getDescription())
```

Happy to open a PR with this if useful — wanted to confirm the intended fix location first in case
there's a reason it was left out.

sheldon-nadarajah-sft:
Fixes #2154.

## Bug

`GetPolicy`, `ListPolicies`, and `CreatePolicy` all built their response via the shared
`policyXml()` helper in `IamQueryHandler.java`, which never emitted a `Description` element —
even when the policy was created with one, and even though the sibling `roleXml()` (right above
it) does emit `Description` for roles. `IamPolicy` (the model) and `IamService.createPolicy()`
both handle the field correctly; it was only dropped on the way out to the client.

Real AWS's `GetPolicy` returns `Description` (see the
[API docs](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetPolicy.html)'s own sample
response). Consumers that diff against real AWS behavior — e.g. Terraform's AWS provider — see the
description as permanently missing on every read, and since IAM policy descriptions are immutable
(force-new on the AWS provider), this forces a destroy+recreate of the policy on every apply, even
when nothing in the underlying config changed.

## Fix

One line: add `.elem("Description", p.getDescription())` to `policyXml()`, matching the existing
pattern in `roleXml()`. `XmlBuilder.elem()` already skips the element when the value is `null`, so
this is a strict superset of the previous output — no behavior change for policies without one.

## Tests

Added `Description` assertions to the existing `createPolicy()` and `getPolicy()` integration
tests in `IamIntegrationTest` — both already pass a `Description` on `CreatePolicy` but weren't
asserting on it in the response, so they didn't catch this. Ran locally (Java 25):

```
IamIntegrationTest:              49/49 passed
IamServiceTest:                  60/60 passed
IamManagedPolicyAccountScopeTest: 7/7 passed
IamConcurrencyTest:               13/13 passed
```

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamQueryHandler.java
@@ -917,6 +917,7 @@ public class IamQueryHandler {
                 .elem("IsAttachable", true)
                 .elem("CreateDate", isoDate(p.getCreateDate()))
                 .elem("UpdateDate", isoDate(p.getUpdateDate()))
+                .elem("Description", p.getDescription())
                 .build();
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamQueryHandler.java
@@ -450,13 +450,13 @@ public class IamQueryHandler {
         String document = getParam(params, "PolicyDocument");
         Map<String, String> tags = extractTags(params);
         IamPolicy policy = iamService.createPolicy(policyName, path, description, document, tags);
-        String result = new XmlBuilder().start("Policy").raw(policyXml(policy)).end("Policy").build();
+        String result = new XmlBuilder().start("Policy").raw(policyXml(policy, true)).end("Policy").build();
         return Response.ok(AwsQueryResponse.envelope("CreatePolicy", AwsNamespaces.IAM, result)).build();
     }
 
     private Response handleGetPolicy(MultivaluedMap<String, String> params) {
         IamPolicy policy = iamService.getPolicy(getParam(params, "PolicyArn"));
-        String result = new XmlBuilder().start("Policy").raw(policyXml(policy)).end("Policy").build();
+        String result = new XmlBuilder().start("Policy").raw(policyXml(policy, true)).end("Policy").build();
         return Response.ok(AwsQueryResponse.envelope("GetPolicy", AwsNamespaces.IAM, result)).build();
     }
 
@@ -470,7 +470,7 @@ public class IamQueryHandler {
                 getParam(params, "Scope"), getParam(params, "PathPrefix"));
         var xml = new XmlBuilder().start("Policies");
         for (IamPolicy p : policyList) {
-            xml.start("member").raw(policyXml(p)).end("member");
+            xml.start("member").raw(policyXml(p, false)).end("member");
         }
         xml.end("Policies").elem("IsTruncated", false);
         return Response.ok(AwsQueryResponse.envelope("ListPolicies", AwsNamespaces.IAM, xml.build())).build();
@@ -906,7 +906,13 @@ public class IamQueryHandler {
                 .build();
     }
 
-    private String policyXml(IamPolicy p) {
+    /**
+     * {@code includeDescription} is per-operation, not per-policy: AWS's own {@code Policy} model
+     * documents that {@code Description} "is included in the response to the GetPolicy operation.
+     * It is not included in the response to the ListPolicies operation" — CreatePolicy documents
+     * neither inclusion nor exclusion, so it's treated the same as GetPolicy.
+     */
+    private String policyXml(IamPolicy p, boolean includeDescription) {
         return new XmlBuilder()
                 .elem("PolicyName", p.getPolicyName())
                 .elem("PolicyId", p.getPolicyId())
@@ -917,7 +923,7 @@ public class IamQueryHandler {
                 .elem("IsAttachable", true)
                 .elem("CreateDate", isoDate(p.getCreateDate()))
                 .elem("UpdateDate", isoDate(p.getUpdateDate()))
-                .elem("Description", p.getDescription())
+                .elem("Description", includeDescription ? p.getDescription() : null)
                 .build();
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/iam/IamIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/iam/IamIntegrationTest.java
@@ -689,6 +689,7 @@ class IamIntegrationTest {
             .body("CreatePolicyResponse.CreatePolicyResult.Policy.PolicyName", equalTo("TestPolicy"))
             .body("CreatePolicyResponse.CreatePolicyResult.Policy.PolicyId", startsWith("ANPA"))
             .body("CreatePolicyResponse.CreatePolicyResult.Policy.DefaultVersionId", equalTo("v1"))
+            .body("CreatePolicyResponse.CreatePolicyResult.Policy.Description", equalTo("Test managed policy"))
         .extract()
             .path("CreatePolicyResponse.CreatePolicyResult.Policy.Arn");
     }
@@ -705,7 +706,8 @@ class IamIntegrationTest {
             .post("/")
         .then()
             .statusCode(200)
-            .body("GetPolicyResponse.GetPolicyResult.Policy.PolicyName", equalTo("TestPolicy"));
+            .body("GetPolicyResponse.GetPolicyResult.Policy.PolicyName", equalTo("TestPolicy"))
+            .body("GetPolicyResponse.GetPolicyResult.Policy.Description", equalTo("Test managed policy"));
     }
 
     @Test
--- a/src/test/java/io/github/hectorvent/floci/services/iam/IamIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/iam/IamIntegrationTest.java
@@ -1081,6 +1081,11 @@ class IamIntegrationTest {
         .then()
             .statusCode(200);
 
+        // Asserting only the negative would pass vacuously if ListPolicies ever returned no
+        // members at all (a pagination bug, a scope-filter regression); the positive assertion
+        // proves the created policy is actually present before the absence check means anything.
+        // Matching the element itself, not the bare word "Description", also avoids colliding
+        // with a future policy name/path containing that substring.
         given()
             .formParam("Action", "ListPolicies")
             .formParam("Scope", "Local")
@@ -1090,6 +1095,7 @@ class IamIntegrationTest {
             .post("/")
         .then()
             .statusCode(200)
-            .body(not(containsString("Description")));
+            .body(containsString("ListPoliciesOmitCheckPolicy"))
+            .body(not(containsString("<Description>")));
     }
 }
--- a/src/test/java/io/github/hectorvent/floci/services/iam/IamIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/iam/IamIntegrationTest.java
@@ -1056,4 +1056,40 @@ class IamIntegrationTest {
             .body("ErrorResponse.Error.Code", equalTo("NoSuchEntity"))
             .body("ErrorResponse.Error.Message", not(containsString("null")));
     }
+
+    @Test
+    @Order(74)
+    void listPoliciesOmitsDescription() {
+        // AWS's own Policy model documents this explicitly: Description "is included in the
+        // response to the GetPolicy operation. It is not included in the response to the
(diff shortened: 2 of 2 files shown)


src/main/java/io/github/hectorvent/floci/services/iam/IamQueryHandler.java
```diff
@@ -917,6 +917,7 @@ private String policyXml(IamPolicy p) {
                 .elem("IsAttachable", true)
                 .elem("CreateDate", isoDate(p.getCreateDate()))
                 .elem("UpdateDate", isoDate(p.getUpdateDate()))
+                .elem("Description", p.getDescription())
```
pgermosen: **This helper is shared with `ListPolicies`, where AWS documents `Description` must be absent**

The IAM model's own documentation for `Policy.Description` says it outright:

> A friendly description of the policy. This element is included in the response to the `GetPolicy` operation. **It is not included in the response to the `ListPolicies` operation.**

And `ListPolicies`' operation documentation says the same thing in general terms:

> IAM resource-listing operations return a subset of the available attributes for the resource. For example, this operation does not return tags, even though they are an attribute of the returned object. To view all of the information for a customer managed policy, see `GetPolicy`.

`policyXml()` has three callers — `handleCreatePolicy` (`:453`), `handleGetPolicy` (`:459`), and the `<member>` loop inside `handleListPolicies` (`:473`). So this line adds `Description` to `ListPolicies` too, which the PR title advertises deliberately, but which AWS excludes.

**The blast radius is bigger than one field.** `listPolicies` includes AWS-managed policies under the default scope (`IamService:533`), and the seeded catalog is **58 entries**, every one of which carries a description. So a plain `aws iam list-policies` against Floci returns 58+ policies each with a `<Description>` that real AWS omits.

Worth naming the irony: this PR exists because a consumer diffing Floci against real AWS saw a field that should have been there and wasn't. This adds a field that shouldn't be there and is — same failure mode, opposite direction, on the operation Terraform actually uses to enumerate policies.

**Why the `roleXml()` precedent doesn't carry over,** since that's the trap and it's a reasonable one to fall into: `ListRoles` documents its exclusions explicitly —

> This operation does not return the following attributes, even though they are an attribute of the returned object: `PermissionsBoundary`, `RoleLastUsed`, `Tags`

`Description` isn't on that list, so `ListRoles` *does* return it, and `roleXml()` emitting it unconditionally is correct. AWS simply subsets the two resource types differently. "Match what `roleXml()` does" is the wrong heuristic here even though the two helpers look symmetric.

**Suggested:** keep `policyXml()` description-free and emit the element at the two call sites that want it, or parameterize:

```java
private String policyXml(IamPolicy p, boolean includeDescription)
```

with `true` at `:453`/`:459` and `false` at `:473`. Either way the shared helper stops deciding a per-operation question.

For the record, the rest of the diagnosis checks out: AWS's [GetPolicy sample response](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetPolicy.html) does include `<Description>Allows read-only access to the example bucket</Description>`, `XmlBuilder.elem()` really does skip nulls, and `CreatePolicy` is fine — AWS documents neither inclusion nor exclusion there, and its sample omits `Description` only because the sample request never set one.

src/test/java/io/github/hectorvent/floci/services/iam/IamIntegrationTest.java
```diff
@@ -705,7 +706,8 @@ void getPolicy() {
             .post("/")
         .then()
             .statusCode(200)
-            .body("GetPolicyResponse.GetPolicyResult.Policy.PolicyName", equalTo("TestPolicy"));
+            .body("GetPolicyResponse.GetPolicyResult.Policy.PolicyName", equalTo("TestPolicy"))
+            .body("GetPolicyResponse.GetPolicyResult.Policy.Description", equalTo("Test managed policy"));
```
pgermosen, line 710: **Nothing pins the `ListPolicies` behavior, which is the case that's wrong**

Adding these assertions to the existing `createPolicy`/`getPolicy` tests is the right move — both already sent a `Description` and simply weren't checking it, which is precisely why the gap survived this long.

But there's no `ListPolicies` test in this file at all (zero matches), so the operation this PR changes most visibly is the one with no coverage, before or after.

**Suggested:** once the comment above is addressed, add a `ListPolicies` case asserting `Description` is **absent** from the returned members. That's the assertion that stops someone "fixing the inconsistency" later by folding the element back into the shared helper — which is exactly how this would regress.

pgermosen (changes requested):
Reviewed in a local worktree. IAM suites pass: **129/129, 0 failures** (`IamIntegrationTest` 49/49, `IamServiceTest` 60/60, `IamManagedPolicyAccountScopeTest` 7/7, `IamConcurrencyTest` 13/13) — matching every count in the description.

**The bug is real, and the diagnosis is exactly right.** `GetPolicy` must return `Description`; AWS's own sample response has it:

```xml
<Policy>
  <PolicyName>S3-read-only-example-bucket</PolicyName>
  <Description>Allows read-only access to the example bucket</Description>
  ...
```

The Terraform reasoning holds too — policy descriptions are immutable, so a permanently-absent description forces destroy+recreate on every apply.

**But the one-line fix lands in a helper shared with `ListPolicies`, and AWS documents that `ListPolicies` must not return `Description`.** The model's `Policy.Description` member doc states it directly: *"This element is included in the response to the `GetPolicy` operation. It is not included in the response to the `ListPolicies` operation."* So the change fixes one divergence and introduces another — details inline.

Verified rather than assumed:

- **`XmlBuilder.elem()` really does skip nulls**, so the "strict superset" claim is accurate for the two operations where the element belongs.
- **The three call sites** are `CreatePolicy` (`:453`), `GetPolicy` (`:459`) and `ListPolicies` (`:473`), the last inside the `<member>` loop.
- **`CreatePolicy` is fine.** AWS documents neither inclusion nor exclusion, and its sample response omits `Description` only because the sample request never set one.
- **The model and the API reference agree**, so this isn't a stale-doc artifact.

** The `GetPolicy` half is right and worth merging — it just needs to not apply to `ListPolicies`. It's a small fix, and the second comment covers the test that would keep it fixed.

sheldon-nadarajah-sft:
Thanks for the thorough review, @pgermosen — you're right, and I verified it directly against AWS's own `Policy` model docs before making the change: `Description` really is documented as GetPolicy-only, excluded from ListPolicies.

Pushed a fix: `policyXml()` now takes an explicit `includeDescription` flag (`true` for CreatePolicy/GetPolicy, `false` for ListPolicies' member loop), plus a `ListPolicies` test asserting no member ever carries a `Description` element — exactly the regression a well-intentioned "make it consistent with GetPolicy" cleanup would reintroduce.

All 130 IAM tests pass locally (IamIntegrationTest 50/50, IamServiceTest 60/60, IamManagedPolicyAccountScopeTest 7/7, IamConcurrencyTest 13/13).

src/test/java/io/github/hectorvent/floci/services/iam/IamIntegrationTest.java
```diff
@@ -1054,4 +1056,40 @@ void getUserWithoutUserNameOnAnUnsignedRequestDoesNotLeakNull() {
...
+        given()
+            .formParam("Action", "ListPolicies")
+            .formParam("Scope", "Local")
+            .header("Authorization",
+                    "AWS4-HMAC-SHA256 Credential=test/20260227/us-east-1/iam/aws4_request")
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .body(not(containsString("Description")));
```
pgermosen: **The absence assertion can pass vacuously — nothing proves the member is in the list**

Your reasoning for checking the whole response rather than one policy's value is right, and I'd keep it: folding the element back into the shared helper unconditionally is exactly how this regresses, and a per-policy assertion wouldn't catch it.

Two things would make it hold up better.

**It only asserts a negative.** If `ListPolicies` ever returned an empty `<Policies/>` — a pagination bug, a scope-filter regression, an ordering change in this `@Order`-ed class — the response contains no `Description` because it contains no policies, and the test still passes. The guard silently stops guarding.

**The match is broader than the thing being guarded.** `containsString("Description")` matches the bare word anywhere in the XML, including inside a policy *name* or *path*. Nothing collides today — the only `CreatePolicy` names in this file are `SimulatePolicy`, `SimulateExplicitDenyPolicy`, `TestPolicy` and your new one, so I checked — but this class shares state across ordered tests, so a future policy named something like `DescriptionOnlyPolicy` would break a test that has nothing to do with it, and the failure would read as a `Description` leak.

**Suggested:** assert both halves, and match the element rather than the word:

```java
.body(containsString("ListPoliciesOmitCheckPolicy"))
.body(not(containsString("<Description>")));
```

The positive assertion makes the negative one meaningful, and `<Description>` keeps the whole-response property you wanted while dropping the name-collision surface entirely.

pgermosen (approved):
Looks good, just a minor thing

Both findings are fixed correctly and for the stated reason.

For the record, the resolved blocker was worth the round trip: with 58 seeded AWS-managed policies, each carrying a description, a plain `aws iam list-policies` would have returned 58+ members with a field real AWS omits — the same class of divergence this PR set out to remove, in the opposite direction.

sheldon-nadarajah-sft:
Good catch — fixed. Both halves of the reasoning are addressed: added `.body(containsString("ListPoliciesOmitCheckPolicy"))` so the absence check can't pass vacuously against an empty/mis-filtered member list, and switched the negative match to `<Description>` (the element) instead of the bare word, so it can't collide with a future policy name/path.

All 130 IAM tests still pass locally (`IamIntegrationTest` 50/50, `IamServiceTest` 60/60, `IamManagedPolicyAccountScopeTest` 7/7, `IamConcurrencyTest` 13/13).

pgermosen:
> Good catch — fixed. Both halves of the reasoning are addressed: added `.body(containsString("ListPoliciesOmitCheckPolicy"))` so the absence check can't pass vacuously against an empty/mis-filtered member list, and switched the negative match to `<Description>` (the element) instead of the bare word, so it can't collide with a future policy name/path.
> 
> All 130 IAM tests still pass locally (`IamIntegrationTest` 50/50, `IamServiceTest` 60/60, `IamManagedPolicyAccountScopeTest` 7/7, `IamConcurrencyTest` 13/13).

@sheldon-nadarajah-sft, the build is failing.

PR #2156, opened 2026-08-07
fix(lambda): drain parked extensions before closing the runtime API server

### LINKED ISSUE #2142: [BUG] RuntimeApiServer.stop() closes the HTTP server before dispatching SHUTDOWN, so parked extensions can be orphaned
## Service

Lambda (Extensions API / Runtime API server)

## Summary

`RuntimeApiServer.stop()` starts closing the HTTP server **before** it dispatches the SHUTDOWN event to extensions parked on `/2020-01-01/extension/event/next`. Both dispatch paths go through `vertx.runOnContext(...)`, so they are queued *after* `httpServer.close()` has already begun tearing down connections. When close wins the race, the parked poller's connection dies before the response is written and the extension never receives its SHUTDOWN.

That is the invariant `extensionShutdown_racedAgainstStop_isNeverOrphaned` is named for, and it is the cause of the `EOFException` flake that has been failing CI across unrelated PRs.

## Expected behavior

An extension subscribed to `SHUTDOWN` and parked on `/event/next` receives the SHUTDOWN event when the runtime API server stops.

## Actual behavior

It intermittently receives a dead connection instead:

```
java.io.EOFException: EOF reached while reading
	at jdk.internal.net.http.Http2Connection$Http2TubeSubscriber.onComplete(Http2Connection.java:1903)
	...
	at ...RuntimeApiServerTest.extensionRegister_racedAgainstStop_eventuallyDeliversShutdown(RuntimeApiServerTest.java:1166)
```

## The ordering

`src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java`, in `stop()`:

```java
507:  httpServer.close(ar -> { ... });                          // begins tearing down connections

520:  // Wake any parked /next pollers with 204 ...
521:  for (RoutingContext ctx : contextsToClose204) {
522:      vertx.runOnContext(v -> { ... ctx.response().setStatusCode(204).end(); });
523:  }

529:  // Dispatch SHUTDOWN to every extension that was already parked on /event/next.
530:  for (Runnable dispatch : dispatches) {
531:      vertx.runOnContext(v -> dispatch.run());
532:  }
```

The SHUTDOWN write is scheduled after close has already been requested, so whether it lands is a race.

## Why the existing park-wait doesn't cover it

#2131 added `awaitExtensionParked(...)` so the test races `stop()` against a poll that has *provably* parked. That correctly fixed the setup race, but it cannot fix this one: the poll has parked, the handler is waiting, and the response still never gets written because the connection is gone. The `assertTrue(awaitExtensionParked(...))` passes and the failure happens afterwards, on `asyncNext.get(...)`.

This is also why the two race tests fail differently:

- `extensionShutdown_racedAgainstStop_isNeverOrphaned` **tolerates** a missed delivery — it retries with a second poll — so it only ever exposed the setup race. (That one still used `Thread.sleep(5)`; fixed in #2141.)
- `extensionRegister_racedAgainstStop_eventuallyDeliversShutdown` asserts the first response directly with no retry, so it is the one that exposes *this* bug.

## Reproduction

The race does not appear on an idle machine — the class passes repeatedly unconstrained. It reproduces under CPU starvation:

```bash
docker run --rm --cpus=1 -v <tree>:/work -v floci-m2:/root/.m2 \
  -v /var/run/docker.sock:/var/run/docker.sock -w /work eclipse-temurin:25-jdk \
  ./mvnw -B -ntp -Dtest='RuntimeApiServerTest' -DfailIfNoSpecifiedTests=false test
```

Note it must be the **whole class** — running the single method in isolation completes in ~4s and never reproduces, because the contention comes from the preceding 44 tests in the same JVM.

Observed in CI on four unrelated branches, including #2131's own:

| branch | failing test | date |
|---|---|---|
| `fix/extension-register-race-test-flake` (#2131) | `extensionShutdown_...` | 08-05 |
| #1805 (RDS parameter groups) | `extensionRegister_...` | 08-06 |
| #1807 (CFN EventBus) | `extensionRegister_...` | 08-06 |
| #1820 (CloudFront serving) | `extensionShutdown_...` | 08-06 |
| #1823 (StepFunctions ResultWriter) | `extensionShutdown_...` | 08-06 |

## Suggested fix

Dispatch the SHUTDOWN events and the 204 wake-ups **before** calling `httpServer.close()`, or chain the close so it runs only once those context tasks have completed. The second is stricter: `runOnContext` queues the write, but `close()` is not guaranteed to run on the same context, so simply moving the loops above line 507 narrows the window rather than closing it.

I have deliberately not opened a PR for this. The race is too rare to demonstrate locally, and I would rather not ship an unverified reordering of server shutdown — whoever owns this code is better placed to choose between "reorder" and "chain the close", and to say whether `stop()`'s returned future should now complete only after the dispatches. Happy to implement whichever shape you prefer.

#2141 fixes the separate, verified test-timing half.

aniketshukla1:
RuntimeApiServer.stop() called httpServer.close() before the vertx.runOnContext(...) loops that write the 204 wake-ups and SHUTDOWN events to parked extensions. Because close() begins tearing down connections immediately — and is not guaranteed to run on the same event-loop context as those writes — a parked extension's SHUTDOWN response could be lost when close won the race, orphaning the extension. That surfaced as the intermittent EOFException failing RuntimeApiServerTest in CI across unrelated PRs.

Close the HTTP server only after every parked 204/SHUTDOWN write has run, tracked by a counter that fires the close once the last write completes. This holds even when the writes land on different event-loop contexts, so merely reordering the loops above the close (which narrows but does not close the window) is not relied upon.

Fixes #2142.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
@@ -4,6 +4,7 @@ import io.github.hectorvent.floci.services.lambda.model.ExtensionEvent;
 import io.github.hectorvent.floci.services.lambda.model.InvokeResult;
 import io.github.hectorvent.floci.services.lambda.model.PendingInvocation;
 import io.github.hectorvent.floci.services.lambda.model.RegisteredExtension;
+import io.vertx.core.Future;
 import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpServer;
 import io.vertx.core.http.HttpServerOptions;
@@ -22,6 +23,8 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
 
 /**
  * Per-container HTTP server implementing the AWS Lambda Runtime API and, for
@@ -468,7 +471,7 @@ public class RuntimeApiServer {
     public CompletableFuture<Void> stop() {
         CompletableFuture<Void> closed;
         List<RoutingContext> contextsToClose204;
-        List<Runnable> dispatches = new ArrayList<>();
+        List<Supplier<Future<Void>>> dispatches = new ArrayList<>();
         List<PendingInvocation> invocationsToFailContainerStopped;
 
         synchronized (lock) {
@@ -490,9 +493,10 @@ public class RuntimeApiServer {
                 RoutingContext waitingCtx = ext.takeWaitingContext();
                 if (waitingCtx != null) {
                     dispatches.add(() -> {
-                        if (!waitingCtx.response().ended()) {
-                            sendExtensionEvent(waitingCtx, shutdownEvent);
+                        if (waitingCtx.response().ended()) {
+                            return Future.succeededFuture();
                         }
+                        return sendExtensionEvent(waitingCtx, shutdownEvent);
                     });
                 } else {
                     ext.getPendingEvents().offer(shutdownEvent);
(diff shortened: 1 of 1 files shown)


hectorvent (approved):
Thanks @aniketshukla1

PR #2158, opened 2026-08-07
feat(applicationautoscaling): add Application Auto Scaling control plane

### LINKED ISSUE #2157: [FEAT] Application Auto Scaling: emulate the AnyScaleFrontendService control plane (scalable targets and scaling policies)
## Service

Application Auto Scaling, apiVersion 2016-02-06.

## API Action / Feature

Application Auto Scaling is entirely missing today — no package, no `ResolvedServiceCatalog` entry, no config key. It is **not** the same service as the existing `services/autoscaling` package: that one is EC2 Auto Scaling (Query protocol, `autoscaling` signing name), while this is AWS JSON 1.1 under the `application-autoscaling` signing name with the `AnyScaleFrontendService.` target prefix.

It is the API behind `aws_appautoscaling_target` and `aws_appautoscaling_policy`, so it is how ECS services, MSK broker storage, DynamoDB capacity and Lambda provisioned concurrency scale.

Proposing the nine actions Terraform actually exercises:

- **Scalable targets:** `RegisterScalableTarget`, `DescribeScalableTargets`, `DeregisterScalableTarget`
- **Scaling policies:** `PutScalingPolicy`, `DescribeScalingPolicies`, `DeleteScalingPolicy`
- **Tagging:** `ListTagsForResource`, `TagResource`, `UntagResource`

Three details are load-bearing and easy to get wrong:

- **Identity is the triple (ServiceNamespace, ResourceId, ScalableDimension)** — there is no separate ID, and `RegisterScalableTarget` is an upsert on that triple where unsupplied parameters are left unchanged.
- **The two ARN families use different service names.** `ScalableTargetARN` is `arn:aws:application-autoscaling:...:scalable-target/<id>`, but `PolicyARN` is `arn:aws:autoscaling:...:scalingPolicy:<uuid>:resource/<ns>/<resourceId>:policyName/<name>` — plain `autoscaling`.
- **Tagging is mandatory, not optional.** The Terraform provider annotates the resource `@Tags(identifierAttribute="arn")`, so its tagging interceptor calls `ListTagsForResource` on *every* read via generated `tags_gen.go`. That call is invisible in `target.go`. It resolves by the ARN the provider reads out of `DescribeScalableTargets`, so if that response omits `ScalableTargetARN`, tags break even when `ListTagsForResource` exists.

Scaling policies would be stored and described faithfully but remain **inert** — nothing evaluates them and no capacity moves — matching the existing behavior of EC2 Auto Scaling's `PutScalingPolicy` in Floci. Target-tracking policies would still create the real pair of CloudWatch alarms AWS creates on your behalf, since `PutMetricAlarm` already exists in-process.

## AWS Documentation

- https://docs.aws.amazon.com/autoscaling/application/APIReference/Welcome.html
- https://docs.aws.amazon.com/autoscaling/application/APIReference/API_RegisterScalableTarget.html
- https://docs.aws.amazon.com/autoscaling/application/APIReference/API_PutScalingPolicy.html
- https://docs.aws.amazon.com/autoscaling/application/APIReference/API_ListTagsForResource.html

## Why is this needed?

Any real ECS estate scales its services through `aws_appautoscaling_target` + `aws_appautoscaling_policy`, and today every one of those resources fails `terraform apply` against Floci with a hard 400. In the production Terraform repo that motivated this, a single shared module wrapping those two resources is instantiated 74 times across five environments, and the MSK module registers a second scalable target for broker storage — so the whole ECS service tier and part of the shared tier are blocked on this one missing service.

The gap is also invisible from the outside: because Floci returns `UnknownOperationException` for an unknown target prefix, it looks like a routing problem rather than a missing service.

- [x] Yes
- [ ] No

pgermosen:
## Summary

Adds a new `services/applicationautoscaling` package emulating the Application Auto Scaling control plane (JSON 1.1, `AnyScaleFrontendService.` target prefix, `application-autoscaling` signing name). Closes #2157

This is a distinct service from the existing `services/autoscaling` package, which is EC2 Auto Scaling over Query under the `autoscaling` signing name — different API, different protocol, different credential scope. It is the API behind `aws_appautoscaling_target` and `aws_appautoscaling_policy`.

- **Nine actions:** `RegisterScalableTarget`, `DescribeScalableTargets`, `DeregisterScalableTarget`, `PutScalingPolicy`, `DescribeScalingPolicies`, `DeleteScalingPolicy`, `ListTagsForResource`, `TagResource`, `UntagResource`. Handler/Service/model following the ACM + CloudTrail JSON-handler pattern, storage through `StorageFactory`, package-private constructor for testability.
- **Identity:** the `(ServiceNamespace, ResourceId, ScalableDimension)` triple — there is no separate ID. `RegisterScalableTarget` upserts on that triple and leaves unsupplied parameters unchanged, matching AWS. Both the namespace and dimension are validated against the AWS enums.
- **Two ARN families, deliberately different service names:** `ScalableTargetARN` is `arn:aws:application-autoscaling:…:scalable-target/<id>` (matching the pattern AWS enforces on `ListTagsForResource`), while `PolicyARN` is `arn:aws:autoscaling:…:scalingPolicy:<uuid>:resource/<ns>/<resourceId>:policyName/<name>` — plain `autoscaling`, embedding the resource path.
- **Tagging is load-bearing, not cosmetic.** The Terraform provider annotates the resource `@Tags(identifierAttribute="arn")`, so its transparent-tagging interceptor calls `ListTagsForResource` on *every* read via generated `tags_gen.go` — a call that is invisible in `target.go`. It keys on the ARN read out of `DescribeScalableTargets`, so that response must always carry `ScalableTargetARN` or tags break even with the action implemented.
- **Real CloudWatch alarms.** A target-tracking policy creates the `AlarmHigh`/`AlarmLow` pair AWS creates on your behalf, via the existing in-process `CloudWatchMetricsService`. They appear in `DescribeAlarms` and are removed when the policy is deleted or its target deregistered — deregistration cascades to policies and their alarms, per AWS semantics.
- **No invented values.** Optional configuration fields are echoed back only when supplied, so an unset `ScaleInCooldown` or `DisableScaleIn` never appears as a value Terraform did not configure.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

Wire shapes were cross-checked against the AWS API reference for each action, and the protocol was confirmed empirically before implementation by capturing a live `aws application-autoscaling` CLI request: `X-Amz-Target: AnyScaleFrontendService.DescribeScalableTargets`, credential scope `…/application-autoscaling/aws4_request`, `application/x-amz-json-1.1`, `POST /`.

Scaling policies are stored and described faithfully but are **inert** — nothing evaluates them and no capacity is ever adjusted, so an ECS service's desired count will not move. This mirrors the existing behavior of EC2 Auto Scaling's `PutScalingPolicy` and is called out explicitly under Limitations in `docs/services/applicationautoscaling.md` so the boundary is not left implicit.

Verified end to end with the real AWS CLI across both the `ecs` and `kafka` namespaces (register → describe → tag → put policy → describe policies → deregister), with the error taxonomy confirmed on the wire (`ObjectNotFoundException`, `ValidationException`, `ResourceNotFoundException`). Persistence was checked under `storage.mode=persistent` in both JVM and native mode: targets, policies (including the nested `ResourceLabel` and cooldowns) and tags all reload intact after a restart.

Terraform: `apply` → second `plan` reports no changes → in-place update → second `plan` reports no changes → `destroy`.

## Checklist

- [x] `./mvnw test` passes locally (18 unit + 13 integration tests added; the suite otherwise reports 110 pre-existing failures and 31 errors that reproduce identically on a clean `upstream/main` checkout with these changes stashed — all Lambda-dependent, since `CreateFunction` currently returns 400 in my environment)
- [x] New or updated integration test added (`ApplicationAutoScalingServiceTest`, `ApplicationAutoScalingIntegrationTest`)
- [x] `make docs-check` passes; `docs/services/applicationautoscaling.md` added and registered in `tools/docs/services.yaml`
- [x] Native image builds (`-Pnative`, container build) and the persistence round-trip was verified against the native binary — all eight model classes carry `@RegisterForReflection`

## Note on the compat-terraform suite

`compat-terraform` gains an `appautoscaling` endpoint, a fixture (scalable target with tags + two target-tracking policies, one using `ALBRequestCountPerTarget` with a `resource_label`, which is the field most likely to be dropped), and three bats cases. All 25 tests pass.

One of those cases asserts an empty **second** plan, because a successful `apply` proves very little on its own — a response that silently drops a field still applies cleanly and only surfaces later as perpetual drift. The harness did not previously assert this anywhere.

It is scoped to the Application Auto Scaling resources via `-target`, with an inline comment explaining why. A whole-config re-plan currently reports `0 to add, 3 to change` against `aws_cognito_user_pool`, `aws_db_instance` (`auto_minor_version_upgrade` flips `false → true`) and `aws_kinesis_firehose_delivery_stream` (`s3_backup_mode` appears as `"Disabled"`). Those are pre-existing and unrelated to this PR, so I have not widened the assertion or tried to fix them here — happy to open a separate issue for them if that is useful.


pgermosen:
Superseded by #2159, which runs from a branch on floci-io/floci directly rather than a fork. Same commit (aea142b5), no content change.

PR #2165, opened 2026-08-07
fix(stepfunctions): wrap optimized lambda:invoke result in the Invoke envelope

### LINKED ISSUE #2110: [BUG] SFN Lambda task result missing `Payload` / `StatusCode` wrapper
## Service

SFN -> Lambda

## AWS API Action

`StartExecution` (state machine Task with `Resource: arn:aws:states:::lambda:invoke`); underlying `Lambda.Invoke`

## Expected behavior

For a Task state using the `arn:aws:states:::lambda:invoke` integration, real Step
Functions sets the task result to the **Invoke API response envelope**, wrapping
the function's return value:

```json
{"ExecutedVersion":"$LATEST","Payload":{"marker":"RET","echo":{"in":1}},"StatusCode":200}
```

This is why the documented way to read a Lambda's output is
`$states.result.Payload` (JSONata) or `$.Payload` / `ResultSelector` (JSONPath).

## Actual behavior

Floci sets the task result to the function's return value **directly**, with no
envelope — no `Payload`, `StatusCode`, or `ExecutedVersion` key:

```json
{"marker":"RET","echo":{"in":1}}
```

There is no error message; the request succeeds. Consequently any reference to
`$states.result.Payload` / `$.Payload` resolves to nothing, the state still
succeeds, and the execution ends `SUCCEEDED` carrying silently-empty data. A
pipeline that accumulates results across stages appears to run correctly while
producing an empty result — considerably harder to notice than a failure.

Reproduces in **both** JSONata and the default JSONPath query language, so it
appears to be one root cause in the `lambda:invoke` integration rather than in
expression evaluation. Confirmed directly under JSONata by assigning the raw
result — `$exists($states.result)` is `true` while
`$exists($states.result.Payload)` is `false`, and assigning `$states.result`
yields the function's return value itself.

Note the **request** side is correct: the nested `Payload` key in
`Parameters`/`Arguments` is honored (the function receives exactly that object).
Only the response is unwrapped.

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-west-2
EP="--endpoint-url=http://localhost:4566"

cat > handler.py <<'PY'
def handler(event, context):
    return {"marker": "RET", "echo": event}
PY
zip -q f.zip handler.py

aws $EP iam create-role --role-name r \
  --assume-role-policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Service":["lambda.amazonaws.com","states.amazonaws.com"]},"Action":"sts:AssumeRole"}]}'
aws $EP lambda create-function --function-name pwfn --runtime python3.12 \
  --handler handler.handler --role arn:aws:iam::000000000000:role/r \
  --zip-file fileb://f.zip

cat > sm.json <<'JSON'
{
  "StartAt": "Call",
  "States": {
    "Call": {
      "Type": "Task",
      "Resource": "arn:aws:states:::lambda:invoke",
      "Parameters": { "FunctionName": "pwfn", "Payload": {"in": 1} },
      "End": true
    }
  }
}
JSON

aws $EP stepfunctions create-state-machine --name jp \
  --definition file://sm.json \
  --role-arn arn:aws:iam::000000000000:role/r --type STANDARD

E=$(aws $EP stepfunctions start-execution \
  --state-machine-arn arn:aws:states:us-west-2:000000000000:stateMachine:jp \
  --input '{}' --query executionArn --output text)
sleep 8

# Actual:   {"marker":"RET","echo":{"in":1}}
# Expected: {"ExecutedVersion":"$LATEST","Payload":{"marker":"RET","echo":{"in":1}},"StatusCode":200}
aws $EP stepfunctions describe-execution --execution-arn "$E" --query output --output text
```

To confirm under JSONata, use the same Task with
`"Assign": {"shape": "{% $states.result %}"}` and output `$shape` — it is the
function's return value rather than an envelope.

## Environment

- Floci version / image tag: **1.5.34**, `floci/floci:latest`
  (`sha256:b3b3a70a294b8ba8095385b8571ea1e4d44d494950d98de5e812cd9de02f506b`)
- Java SDK version (if applicable): n/a — reproduced with AWS CLI
  `aws-cli/2.27.8 Python/3.13.7`. Also hit via the Terraform AWS provider
  `v6.57.1` / OpenTofu `1.12.3`.
- How you're running Floci: **Docker** —
  `docker run -d --name floci -p 4566:4566 -v /var/run/docker.sock:/var/run/docker.sock floci/floci`
  on macOS 26.5.2 / arm64 (Rancher Desktop)

soreavis:
## Summary

Closes #2110.

A Task using the optimized `arn:aws:states:::lambda:invoke` integration returned the Lambda function's return value directly as the task result, with no `ExecutedVersion` / `Payload` / `StatusCode` keys. Real Step Functions nests the function output inside the Invoke response metadata for this integration, which is why the AWS examples read a Lambda result through `$.Payload` (JSONPath) or `$states.result.Payload` (JSONata). Against the unwrapped result those references resolve to nothing, the state still succeeds, and the execution ends `SUCCEEDED` carrying empty data.

In `invokeResource()` the direct-ARN branch and the optimized branch shared one `if (functionName != null)` block, so the output was returned raw for both. This gates the wrap to the optimized branch. A directly specified function ARN still returns only the function output — the documented behavior for that form, and now covered by a regression test.

Old task result for `Resource: "arn:aws:states:::lambda:invoke"`, then new:

```json
{"marker":"RET","echo":{"in":1}}
```

```json
{"ExecutedVersion":"$LATEST","Payload":{"marker":"RET","echo":{"in":1}},"StatusCode":200}
```

**Note this changes the task result shape for existing users.** State machines that adapted to the unwrapped form — reading `$.marker` where AWS needs `$.Payload.marker`, or dropping the `OutputPath: "$.Payload"` that Workflow Studio sets by default — were already incompatible with real AWS and will now behave as they do there. Direct-ARN tasks, `.waitForTaskToken` tasks and every non-Lambda integration are unchanged.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

[Invoke Lambda with Step Functions](https://docs.aws.amazon.com/step-functions/latest/dg/connect-lambda.html): "When the `Task` result is returned, the function output is nested inside a dictionary of metadata", with an example carrying `ExecutedVersion`, `Payload`, `SdkHttpMetadata`, `SdkResponseMetadata` and `StatusCode: 200`. Same page: "The `Payload` field of the response is parsed from escaped Json to Json" (so `Payload` stays a parsed node, matching the existing `readTree`), and for the direct-ARN form, "the task result contains only the function output". The [Task state reference](https://docs.aws.amazon.com/step-functions/latest/dg/state-task.html) reads the optimized integration's result as `"Output": "{% $states.result.Payload %}"`, and [Workflow Studio's docs](https://docs.aws.amazon.com/step-functions/latest/dg/workflow-studio-process.html) state the default `OutputPath` for Lambda Invoke states is `$.Payload`, which "removes the additional metadata" — both only work if the envelope is there. The new tests cover both shapes.

Values come from real emulator state: `ExecutedVersion` from the resolved function's version (the same `fn.getVersion()` that `LambdaService.invoke()` reports since #2026), `StatusCode` from `InvokeResult`, `Payload` from the existing parse.

`SdkHttpMetadata` and `SdkResponseMetadata` are left out on purpose. Their contents are artifacts of a real HTTPS call — `Date`, `X-Amzn-Trace-Id`, `x-amzn-RequestId`, `Content-Length` — and floci invokes in process, so anything it put there would be invented. The documented read paths only touch the three keys emitted here, and a test pins the key set at exactly those three, so adding the blocks later is a deliberate change rather than a drift. Say the word if you want them and I'll add them.

Two pre-existing gaps this surfaces rather than fixes: `extractLambdaFunctionName` drops the qualifier (#1660), so a task using `Qualifier` or a qualified `FunctionName` still invokes `$LATEST` and will now report `ExecutedVersion: "$LATEST"` for it; and `InvocationType` is hardcoded to `RequestResponse`, so an `InvocationType: "Event"` task reports `200` where AWS reports `202`. Both want their own issue — I can file them if you'd like them tracked.

## Checklist

- [x] `./mvnw test` passes locally *(with pre-existing Docker-dependent failures unrelated to this change — see the test note)*

New Docker-free `AslExecutorLambdaInvokeResultTest`, modeled on `AslExecutorCatchTest` (mocked `LambdaExecutorService` + `LambdaFunctionStore`, driven through `executeSync`): the envelope and its exact key set, the direct-ARN raw-output guard, `ResultSelector` over `$.Payload.marker`, `OutputPath: "$.Payload"`, and JSONata `$states.result.Payload`. Reverting the `AslExecutor` change with the tests in place fails 4 of the 5, with the direct-ARN guard green on both sides. `make docs-check` passes; no handler action changed.

Test note: in the full `./mvnw test` (8393 tests) all 27 `services/stepfunctions` classes are green. The failures on this machine sit in container-backed subsystems (DocumentDB, Neptune, WebSockets, ECR, API Gateway authorizers) — there is no Docker daemon here, and the same classes fail identically on a clean `main` tree.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -501,12 +501,14 @@ public class AslExecutor {
         // Support Lambda resources: direct ARN or optimized integration
         String functionName = null;
         JsonNode lambdaPayload = input;
+        boolean optimizedLambdaInvoke = false;
 
         if (resource.contains(":lambda:") && resource.contains(":function:")) {
             // Direct Lambda ARN: arn:aws:lambda:region:account:function:name[:qualifier]
             functionName = extractLambdaFunctionName(resource);
         } else if (resource.equals("arn:aws:states:::lambda:invoke")) {
             // Optimized Lambda integration — function name and payload come from resolved input
+            optimizedLambdaInvoke = true;
             String fnRef = input.path("FunctionName").asText(null);
             if (fnRef != null) {
                 functionName = extractLambdaFunctionName(fnRef);
@@ -533,10 +535,20 @@ public class AslExecutor {
             }
 
             byte[] responseBytes = result.getPayload();
-            if (responseBytes != null && responseBytes.length > 0) {
-                return objectMapper.readTree(responseBytes);
+            JsonNode functionOutput = responseBytes != null && responseBytes.length > 0
+                    ? objectMapper.readTree(responseBytes)
+                    : NullNode.getInstance();
+
+            // A direct function ARN yields only the function output, while the optimized
+            // integration nests it in the Invoke response metadata.
+            if (!optimizedLambdaInvoke) {
+                return functionOutput;
             }
-            return NullNode.getInstance();
+            ObjectNode invokeResponse = objectMapper.createObjectNode();
+            invokeResponse.put("ExecutedVersion", fn.getVersion());
+            invokeResponse.set("Payload", functionOutput);
+            invokeResponse.put("StatusCode", result.getStatusCode());
+            return invokeResponse;
         }
 
         // DynamoDB optimized integrations (4 actions)
--- a/src/test/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutorLambdaInvokeResultTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutorLambdaInvokeResultTest.java
@@ -1,0 +1,237 @@
+package io.github.hectorvent.floci.services.stepfunctions;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.hectorvent.floci.config.EmulatorConfig;
+import io.github.hectorvent.floci.services.dynamodb.DynamoDbJsonHandler;
+import io.github.hectorvent.floci.services.dynamodb.DynamoDbService;
+import io.github.hectorvent.floci.services.lambda.LambdaExecutorService;
+import io.github.hectorvent.floci.services.lambda.LambdaFunctionStore;
+import io.github.hectorvent.floci.services.lambda.model.InvocationType;
+import io.github.hectorvent.floci.services.lambda.model.InvokeResult;
+import io.github.hectorvent.floci.services.lambda.model.LambdaFunction;
+import io.github.hectorvent.floci.services.s3.S3Service;
+import io.github.hectorvent.floci.services.sqs.SqsJsonHandler;
+import io.github.hectorvent.floci.services.stepfunctions.model.Execution;
+import io.github.hectorvent.floci.services.stepfunctions.model.HistoryEvent;
+import io.github.hectorvent.floci.services.stepfunctions.model.StateMachine;
+import io.quarkus.test.junit.QuarkusTest;
+import io.vertx.mutiny.core.Vertx;
+import jakarta.enterprise.inject.Instance;
+import jakarta.inject.Inject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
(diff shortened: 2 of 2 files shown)


pgermosen (approved):
Checked this against the AWS Lambda/Step Functions integration docs directly rather than relying on the PR's own citations.

Everything matches verbatim: the documented envelope shape (`ExecutedVersion`/`Payload`/`StatusCode`) for the optimized `arn:aws:states:::lambda:invoke` integration, the "task result contains only the function output" behavior for a directly specified function ARN, and the escaped-JSON-to-JSON parsing note for `Payload`. Deliberately leaving out `SdkHttpMetadata`/`SdkResponseMetadata` (which would have to be invented, since floci invokes in-process) is the right call, and it's pinned by a test asserting exactly the three real keys, so it won't silently drift.

Also appreciated the honesty about what this doesn't fix -- the `Qualifier`-dropping bug and hardcoded `InvocationType` are called out as separate pre-existing issues rather than glossed over.

Approving.

PR #2166, opened 2026-08-07
feat(cognito): add AdminLinkProviderForUser

### LINKED ISSUE #1563: [FEAT] Add support for Cognito AdminLinkProviderForUser API action
## Service

Cognito (cognito-idp)

## API Action / Feature

AdminLinkProviderForUser

## AWS Documentation

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminLinkProviderForUser.html

## Why is this needed?

We have a Cognito `PreSignUp_ExternalProvider` Lambda trigger that links a federated provider identity (e.g. `Google_<id>`) to an existing native Cognito user via `AdminLinkProviderForUser` - the standard pattern for attaching SSO sign-ins to pre-provisioned accounts.

We run the compiled (GraalVM native-image) Lambda end-to-end against Floci as a smoke test. The handler's earlier cognito-idp calls work fine against Floci - `ListUsers` and `AdminListGroupsForUser` both succeed - but the final `AdminLinkProviderForUser` returns:

```
400 UnsupportedOperation: Operation AdminLinkProviderForUser is not supported.
```

As a result the external-provider account-linking flow can't be validated against the emulator; we currently have to stop the test at the link step instead of asserting the linked result.

- [ ] Yes
- [x] No

soreavis:
## Summary

`AdminLinkProviderForUser` had no arm in the `CognitoJsonHandler` switch, so it fell through to the default case and returned exactly what the issue reports: `400 UnsupportedOperation: Operation AdminLinkProviderForUser is not supported.` Closes #1563.

The link is recorded as an entry in the destination user's `identities` attribute. No model change was needed — `CognitoUser.attributes` is already a free-form `Map<String, String>` and `AdminGetUser` emits every attribute as a `UserAttribute`, so the identity is returned by `AdminGetUser` and `ListUsers` without touching either. A source `(providerName, userId)` pair may be linked to at most one user per pool. `AdminDisableProviderForUser` is the natural companion but isn't what the issue asks for, so it's out of scope here.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Shapes come from the [API reference](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminLinkProviderForUser.html) samples, so the SDK and CLI work unmodified. The response is 200 `{}`, asserted on the wire. The `identities` entry matches the [`AdminGetUser` sample response](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminGetUser.html) field for field — `userId`, `providerName`, `providerType`, `issuer: null`, `primary: false`, and `dateCreated` as a 13-digit epoch **milliseconds** number. Milliseconds is the one deliberate break from local idiom (the rest of `CognitoService` stores seconds), so the field carries a comment and a test that fails if it's "corrected". `providerType` is emitted only for the four [`ProviderDescription`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ProviderDescription.html) enum values that double as provider names, and `issuer` is always null — telling SAML from OIDC needs a registered IdP that floci doesn't model. `ProviderAttributeName` on `DestinationUser` "is ignored" per the reference; on `SourceUser` it's meaningful to AWS and is accepted and ignored here, so the uniqueness key is `(providerName, userId)` where AWS's includes the attribute name.

Three gaps worth knowing before anyone relies on this, all easy to change if you'd rather:

- **Linking is bookkeeping only.** floci has no federated sign-in and nothing reads `identities`, so a linked user still can't authenticate through the provider. This unblocks SDK and IaC flows that call the action and read the result, not the sign-in half AWS describes.
- **No IdP validation, no five-identity cap.** There's no IdP CRUD in floci to validate a source provider name against. The cap is documented, but nothing in the docs says which error fires when it's hit, so I left it rather than invent an error code.
- **Re-link semantics are my call, not AWS's.** The reference never says what happens when a source identity is already linked, so I picked one invariant: one user per pool per `(providerName, userId)`, conflicts throwing the documented `AliasExistsException`. If you'd rather an identical re-link to the *same* user be an idempotent no-op — how `adminAddUserToGroup` and the enable/disable pair already behave — that's a three-line change.

## Checklist

- [x] `./mvnw test` passes locally *(with pre-existing Docker-dependent failures unrelated to this change — no Cognito class fails, and this diff doesn't touch those subsystems)*

Eleven tests — seven service, three handler, one end-to-end. The Cognito suite goes 272 → 283, green. There's no red→green to show for a new action, so I checked the tests bind the implementation by mutating it: deleting the re-link guard fails exactly the two guard tests and nothing else, and switching `dateCreated` to seconds fails the shape assertion. A corrupted `identities` attribute (written through `AdminUpdateUserAttributes`, which doesn't validate names) is discarded with a warning and relinked from scratch — also pinned by a test. `docs/services/cognito.md` gains one row by hand — `CognitoJsonHandler` is in `deferred_handlers` and the doc has no `floci:actions` markers, so `make docs-sync` doesn't own it.

### DIFF
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -64,6 +64,7 @@ Standalone `TagResource` rejects reserved `floci:*` keys. `ListTagsForResource` 
 | AdminDeleteUser | Deletes a user from a user pool. |
 | AdminSetUserPassword | Sets a user's password and permanent-password status. |
 | AdminUpdateUserAttributes | Updates attributes for a user in a user pool. |
+| AdminLinkProviderForUser | Links an external IdP identity to an existing user's `identities` attribute. |
 
 ### User Operations
 
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
@@ -67,6 +67,7 @@ public class CognitoJsonHandler {
             case "AdminUserGlobalSignOut" -> handleAdminUserGlobalSignOut(request);
             case "AdminEnableUser" -> handleAdminEnableUser(request);
             case "AdminDisableUser" -> handleAdminDisableUser(request);
+            case "AdminLinkProviderForUser" -> handleAdminLinkProviderForUser(request);
             case "ListUsers" -> handleListUsers(request);
             case "InitiateAuth" -> handleInitiateAuth(request);
             case "AdminInitiateAuth" -> handleAdminInitiateAuth(request);
@@ -433,6 +434,16 @@ public class CognitoJsonHandler {
 
     private Response handleAdminDisableUser(JsonNode request) {
         service.adminDisableUser(request.path("UserPoolId").asText(), request.path("Username").asText());
+        return Response.ok(objectMapper.createObjectNode()).build();
+    }
+
+    private Response handleAdminLinkProviderForUser(JsonNode request) {
+        JsonNode sourceUser = request.path("SourceUser");
+        service.adminLinkProviderForUser(
+                request.path("UserPoolId").asText(),
+                request.path("DestinationUser").path("ProviderAttributeValue").asText(),
+                sourceUser.path("ProviderName").asText(),
+                sourceUser.path("ProviderAttributeValue").asText());
         return Response.ok(objectMapper.createObjectNode()).build();
     }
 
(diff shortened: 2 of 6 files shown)


hectorvent:
Thank you, this is a clean addition and the details that are easy to get wrong are all right: `AliasExistsException` is declared on this operation, the response really is an empty object, and `dateCreated` in the `identities` entry is epoch milliseconds even though this service uses seconds elsewhere ([cognito-idp service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json)).

Your reasoning on the lock is also correct, and worth having written down: `UserPool` cannot be the monitor because `updateUserPool` replaces the stored instance, so a check-then-write against it would not actually serialize.

No blockers from my side.

Issue #1539, opened 2026-06-24
[BUG] CloudFormation delete-stack reports success for non-empty S3 bucket but leaves bucket behind, causing recreate rollback


robertjosephwayne:
## Service

CloudFormation, S3

## AWS API Action

`cloudformation DeleteStack`, `cloudformation CreateStack`, `s3api PutObject`, `s3api HeadBucket`, `s3api ListObjectsV2`

## Expected behavior

For real AWS CloudFormation, deleting a stack that owns a non-empty `AWS::S3::Bucket` should not report stack deletion success while the bucket and object still exist. The delete should either fail with the stack in a delete-failed state because the bucket is non-empty, or the managed resource should be removed before `stack-delete-complete` succeeds.

After `cloudformation wait stack-delete-complete` succeeds, recreating the same stack with the same bucket name should not roll back because of the previous stack's leftover managed bucket.

## Actual behavior

Floci reports successful stack deletion:

```text
delete_stack_status=0
wait_delete_status=0
```

But the bucket still exists after deletion, and the object is still present:

```json
{
  "BucketRegion": "us-east-2"
}
```

```json
{
  "Contents": [
    {
      "Key": "object.txt",
      "Size": 18,
      "StorageClass": "STANDARD"
    }
  ],
  "RequestCharged": null,
  "Prefix": ""
}
```

Recreating the same CloudFormation stack with the same bucket name is accepted initially, but `stack-create-complete` fails and the stack rolls back:

```text
create_2_status=0
wait_create_2_status=255
```

`describe-stacks` shows:

```json
{
  "StackStatus": "ROLLBACK_COMPLETE",
  "StackStatusReason": "Your previous request to create the named bucket succeeded and you already own it."
}
```

`describe-stack-events` shows the bucket create failure:

```json
{
  "LogicalResourceId": "Bucket",
  "ResourceType": "AWS::S3::Bucket",
  "ResourceStatus": "CREATE_FAILED",
  "ResourceStatusReason": "Your previous request to create the named bucket succeeded and you already own it."
}
```

This blocks workflows that delete and recreate CloudFormation stacks containing S3 buckets after the bucket has data.

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=foobar
export AWS_SECRET_ACCESS_KEY=foobar
export AWS_DEFAULT_REGION=us-east-2
ENDPOINT=http://localhost.localstack.cloud:4566
STACK_NAME=floci-s3-nonempty-recreate-wait-repro
[...]

PR #2167, opened 2026-08-07
feat(cloudformation): honor resource-level DeletionPolicy

### LINKED ISSUE #1555: [FEATURE] CloudFormation: honor resource-level DeletionPolicy (Retain/RetainExceptOnCreate/Snapshot/Delete)
## Service

CloudFormation

## Summary

Floci's CloudFormation engine does not implement the resource-level [`DeletionPolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html) attribute. A search of the codebase shows no references to `DeletionPolicy` anywhere, so the attribute is silently ignored on both stack delete and resource replacement.

## Context / how this was found

This was discovered while implementing the fix for #1539 (DeleteStack reporting success for a non-empty S3 bucket; see PR #1554).

After that fix, deleting a stack that owns a non-empty `AWS::S3::Bucket` correctly lands the stack in `DELETE_FAILED`. On real AWS, the common way to avoid that is to set `DeletionPolicy: Retain` (or `RetainExceptOnCreate`) on the bucket so the stack delete succeeds and the bucket is intentionally left behind. Because Floci ignores `DeletionPolicy`, it cannot reproduce that AWS behavior — the resource is always deleted (or, for a non-empty bucket, the delete now fails). The two behaviors are related: correct `DeletionPolicy` handling is the AWS-sanctioned escape hatch for exactly the scenario in #1539.

It was deliberately scoped out of the #1539 fix to keep that change narrow and testable.

## Expected behavior (AWS)

On stack delete (and on resource replacement, where AWS applies the same policy to the resource being removed):

- `Delete` (default): the resource is deleted. For most resources, absence of `DeletionPolicy` defaults to `Delete`.
- `Retain`: the resource is left in place and the stack delete still succeeds. The resource should be dropped from the stack's tracked resources without calling the underlying service delete.
- `RetainExceptOnCreate`: behaves like `Retain`, except a resource that was being created when a rollback occurs is deleted.
- `Snapshot`: for resources that support snapshots (e.g. `AWS::RDS::DBInstance`, `AWS::RDS::DBCluster`, `AWS::ElastiCache::CacheCluster`, `AWS::Redshift::Cluster`), AWS takes a snapshot before deleting. A reasonable first step for Floci is to treat `Snapshot` as `Retain`/`Delete` for unsupported types and document the limitation.

`DeletionPolicy` should also be honored on the `UpdateReplacePolicy` path where applicable, but the delete path is the primary gap.

## Actual behavior

`DeletionPolicy` is parsed nowhere and has no effect. Every managed resource is deleted on stack delete regardless of the declared policy.

## Suggested implementation sketch

1. Parse the per-resource `DeletionPolicy` (and optionally `UpdateReplacePolicy`) from the template and store it on `StackResource`.
2. In `CloudFormationService#deleteStackResources`, skip the underlying delete for `Retain` / `RetainExceptOnCreate` resources (mark them appropriately and exclude from the failure tally) instead of calling `provisioner.delete(...)`.
3. Add integration + AWS CLI compatibility tests covering `Retain` (stack reaches `DELETE_COMPLETE`, resource still exists) and the default `Delete` behavior, including a non-empty S3 bucket with `DeletionPolicy: Retain`.

## Environment

- Discovered against `main` while working on #1539 / PR #1554.

soreavis:
## Summary

Closes #1555.

A resource's `DeletionPolicy` was never read off the template, so every managed resource was deleted regardless of what it declared — `deleteStackResources` and `rollbackCreatedResources` both called `provisioner.delete` unconditionally. This reads the attribute in `executeTemplate` and applies it at both sites; they are the only two callers of `provisioner.delete`, and StackSets inherit it since `deleteStackInstances` delegates to `deleteStack`.

| Value | `DeleteStack` | Rollback of the create that made the resource |
|---|---|---|
| `Delete` (default) | deleted | deleted |
| `Retain` | kept | kept |
| `RetainExceptOnCreate` | kept | deleted |

A kept resource is reported as `DELETE_SKIPPED` — on the resource and as a stack event — and is not counted as a failed delete, so the stack still reaches `DELETE_COMPLETE`. That makes `Retain` the AWS-sanctioned escape hatch for the non-empty-bucket case in #1539 this issue came out of. The attribute is re-read on every `UpdateStack`, so adding or removing it takes effect on the next delete.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

`RetainExceptOnCreate` semantics and the `Delete` default are from the [`DeletionPolicy` attribute](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-deletionpolicy.html) reference. `DELETE_SKIPPED` is a real stack event and not only a resource status — [View stack events](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/view-stack-events.html) documents it in the `describe-stack-events` field table ("The `DELETE_SKIPPED` status applies to resources with a deletion policy attribute of retain"), so both surfaces are set.

Divergences, documented in `docs/services/cloudformation.md` rather than half-implemented:

- **`Snapshot` is treated as `Delete`.** AWS *does* delete the resource under `Snapshot`; floci just cannot take the snapshot for the eight types AWS supports it on.
- **Defaults are uniform.** AWS defaults `AWS::RDS::DBCluster`, and `AWS::RDS::DBInstance` without a `DBClusterIdentifier`, to `Snapshot`. floci deletes those today, so that is not a new divergence.
- **No update-time removal leg.** AWS applies `DeletionPolicy` when an update drops a resource from the template, but `executeTemplate` iterates only the new template's logical IDs and never removes dropped resources — so there is nothing to gate yet. Pre-existing, left alone. Replacement during an update is separate: AWS excludes it from `DeletionPolicy`, which is what `UpdateReplacePolicy` covers, and that is not implemented either.
- **Unrecognized values are treated as `Delete`.** I could not find what AWS returns for an invalid value, so I did not invent an error code — I can add validation if you know the real behavior.

## Checklist

- [x] `./mvnw test` passes locally *(with pre-existing Docker-dependent failures unrelated to this change — see the test note)*

New `CloudFormationDeletionPolicyIntegrationTest`, 4 tests: `Retain` surviving the delete of a stack that owns a non-empty bucket, `RetainExceptOnCreate` surviving a delete, a failed create whose rollback deletes the `RetainExceptOnCreate` bucket while the `Retain` one in the same stack survives, and default / explicit `Delete` / unrecognized values still deleting. Reverting the source change fails 3 of the 4. It checks bucket survival through the S3 API rather than CloudFormation's own bookkeeping, and uses RestAssured to match the existing `services/cloudformation` tests, none of which use the AWS SDK.

Test note: the only CloudFormation failure on this machine is `CloudFormationIntegrationTest.createStack_ecrAutoName_isLowercase`, which needs a Docker daemon this machine does not have — it fails identically on a pristine `main` tree with the same single-test command, so it is pre-existing and untouched by this diff (package-scoped runs: 194/1 before, 198/1 after; the 4 new tests are green). The remaining local failures sit in container-backed subsystems on a machine with no Docker daemon and are unrelated to CloudFormation.

No handler action changed, so the generated action tables are untouched.

### DIFF
--- a/docs/services/cloudformation.md
+++ b/docs/services/cloudformation.md
@@ -118,6 +118,36 @@ Resources provisioned by `CreateStack` / `UpdateStack` land in the **caller's ac
 (determined from the request's access key — see [Multi-Account Isolation](../configuration/multi-account.md)).
 Deleting the stack removes them from that same account.
 
+## Deletion Policies
+
+A resource's [`DeletionPolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-deletionpolicy.html)
+attribute is honored on `DeleteStack` and on the rollback of a failed `CreateStack`:
+
+| Value | `DeleteStack` | Rollback of the create that made the resource |
+|---|---|---|
+| `Delete` (default) | deleted | deleted |
+| `Retain` | kept | kept |
+| `RetainExceptOnCreate` | kept | deleted |
+
+A kept resource is reported as `DELETE_SKIPPED` in `DescribeStackEvents` and does not fail the
+deletion — the stack still reaches `DELETE_COMPLETE` while the resource keeps existing. This also
+lets a stack owning a non-empty S3 bucket be deleted, since the bucket is never touched.
+
+Deviations from AWS to be aware of:
+
+- `Snapshot` deletes the resource without taking a snapshot; floci has no snapshot support for the
+  types AWS allows it on.
+- Every resource defaults to `Delete`. AWS instead defaults `AWS::RDS::DBCluster`, and
+  `AWS::RDS::DBInstance` without a `DBClusterIdentifier`, to `Snapshot`.
+- Unrecognized values are treated as `Delete`.
+- An update that removes a resource from the template does not delete it (or consult its policy);
+  only the new template's resources are provisioned. AWS applies `DeletionPolicy` to update-time
+  removals as well.
+- After `DeleteStack` completes, a retained resource's `DELETE_SKIPPED` record is only visible
+  through `DescribeStackEvents` for the deleted stack, not `DescribeStackResources`.
+- `UpdateReplacePolicy`, and the `RetainExceptOnCreate` request parameter of `CreateStack` /
+  `UpdateStack`, are not implemented.
+
 ## StackSets
 
 StackSets deploy a single template into many target accounts and regions:
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
@@ -454,6 +454,7 @@ public class CloudFormationService {
                 for (String logicalId : sortedLogicalIds) {
                     JsonNode resDef = resources.get(logicalId);
                     String type = resDef.path("Type").asText();
+                    String deletionPolicy = resDef.path("DeletionPolicy").asText(null);
                     JsonNode props = resDef.path("Properties");
 
                     CloudFormationTemplateEngine engine = new CloudFormationTemplateEngine(
@@ -479,6 +480,9 @@ public class CloudFormationService {
                                 engine, region, accountId, stack.getStackName(),
                                 resource.getPhysicalId(), resource.getAttributes());
                     }
+                    // Both branches return a fresh StackResource, so the policy is carried over here
+                    // rather than on the instance the loop started with.
+                    resource.setDeletionPolicy(deletionPolicy);
                     stack.getResources().put(logicalId, resource);
 
                     physicalIds.put(logicalId, resource.getPhysicalId());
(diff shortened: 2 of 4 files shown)


hectorvent:
Thank you, this is a well-judged piece of work. `RetainExceptOnCreate` is the value that is easiest to invert and you have it the right way round: retained on stack delete, deleted when the create that made it rolls back. `DELETE_SKIPPED` is also a real `ResourceStatus` rather than a synthesized one ([cloudformation service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cloudformation/2010-05-15/service-2.json)), so the events a client sees match AWS.

The docs table is the part I would point other contributors at. Listing the six deviations by name, especially `Snapshot` deleting without a snapshot and update-time removals not consulting the policy, is more useful than quietly half-implementing either.

No blockers from my side.

hectorvent (approved):

PR #2169, opened 2026-08-07
fix(iot): match topic rules outside the default region

### LINKED ISSUE #2065: [BUG] IoT topic rules are never matched unless created in FLOCI_DEFAULT_REGION (republish/SQS/SNS actions silently dropped)
- The defect: rule storage is region-scoped from the caller (topicRuleStore.put(topicRuleKey(region, ruleName), rule)), but rule lookup hardcodes listTopicRules(config.defaultRegion()) in IotService#handlePublish (~line 897). A rule created in any other region is stored, listable, and invisible to the engine —republish/SQS/SNS/Kinesis/DynamoDB/Lambda actions all silently dropped, Publish still 200. Same hardcoding repeats across every sink in executeTopicRule.
  - Verified matrix (7 rows) showing only the rule-creation region matters; publisher region is irrelevant — including publishing from ap-south-1 and unsigned raw HTTP, both of which still deliver.
  
  - Suggested fix: thread the request region into handlePublish via regionResolver.resolveRegion(headers) (as the shadow endpoints already do).

soreavis:
## Summary

Closes #2065.

Topic rules are stored per region (`topicRuleKey(region, ruleName)`), but `IotService#handlePublish` looked them up with a hardcoded `config.defaultRegion()`. A rule created anywhere else was stored, gettable and listable yet never evaluated — every action was silently dropped while `Publish` still returned 200. `executeTopicRule` repeated the hardcoding, so even a matched rule delivered to default-region resources.

- `IotDataController.publish` took no `@Context HttpHeaders` at all. It now resolves the region from the request's SigV4 credential, like the four shadow endpoints in that same controller, and passes it down.
- `handlePublish` evaluates that region's rules instead of the default region's.
- `executeTopicRule` reads the rule's own region off its `ruleArn` via the existing `AwsArnUtils.regionOrDefault` and gives that to the sinks. No new state — `ruleArn` already carries the creation region. This part changes nothing observable: a rule outside the default region never fired at all, and one inside it resolves to the region it already used.

#2149 is centralizing "did region resolution actually resolve anything" into `RegionResolver#isRegionUnresolved`. This path needs the region itself, not just the boolean, so it adds `RegionResolver#resolveRegionFromAuthOrNull` — same predicate, returns the region or null; a publish whose `Authorization` header isn't SigV4 counts as region-less too, not as a default-region publish. If #2149 lands first I'll fold the two helpers together on the rebase.

## The one call I'd like you to make

**What should a publish that carries no region evaluate?** On AWS the region carrier is the endpoint hostname — SigV4, X.509 and custom-authorizer publishes all reach `<prefix>-ats.iot.<region>.amazonaws.com`. Floci serves every region on one port and `describeEndpoint` returns a region-less authority, so there's nothing to read: MQTT has no carrier, and an unsigned publish has no credential to parse.

I made those evaluate every region's rules, which is what keeps the reporter's unsigned path working. The cost, stated plainly: when rules in N regions match the same topic, one region-less publish now fires all N — measured, 2 regions gave 2 republishes and 3 gave 3. Today it fires at most the default region's, and AWS fires exactly one.

`ApiGatewayService.resolveRestApiRegion` hits the same ambiguity and answers it differently — it prefers the request's region and scans only on a miss, returning one region. If you'd rather have AWS's single delivery, the same-shaped fix is to prefer the default region's rules and scan the rest only when none match. That's a few lines in the private helper `rulesForPublish` and I'll send it. Just say which.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behavior: a rule created in a non-default region matched nothing, and a matched rule's actions targeted the default region.

IoT Core's rules-engine data plane is per-region ([endpoints and quotas](https://docs.aws.amazon.com/general/latest/gr/iot-core.html)), and no sink action carries a region of its own — the only region-bearing field in the whole action union is `http.auth.sigv4.signingRegion` ([CreateTopicRule](https://docs.aws.amazon.com/iot/latest/apireference/API_CreateTopicRule.html)). Destinations are same-region by default: *"Define the rule in the same AWS Region as another service's resource so that the rule action can interact with that resource"* ([rule actions](https://docs.aws.amazon.com/iot/latest/developerguide/iot-rule-actions.html)). So the rule's own region is the right resolution for the sinks.

One documented exception I did **not** implement: SQS may be cross-region — *"The region in this URL doesn't need to be the same AWS Region as your AWS IoT rule"* ([SQS rule action](https://docs.aws.amazon.com/iot/latest/developerguide/sqs-rule-action.html)). Floci's queue URLs carry no region, so a cross-region `queueUrl` isn't representable today; that's a separate change to SQS's URL shape.

Left out of scope, both pre-existing: `handleReservedMqttPublish` still hardcodes the default region for shadow topics, and the `republish` loop guard is untouched.

## Tests

Five cases in `IotTopicRuleIntegrationTest`, existing `@QuarkusTest` + RestAssured style, no Docker:

- `topicRuleMatchesOnlyPublishesInItsOwnRegion` — an `eu-west-1` rule fires for an `eu-west-1`-signed publish, not a `us-east-1`-signed one.
- `unsignedPublishEvaluatesRulesInEveryRegion` — an unsigned publish reaches a rule outside the default region.
- `regionlessPublishFiresEveryMatchingRegionsRule` — pins the fan-out count above, so the trade-off is a test and not just a paragraph.
- `nonSigV4AuthorizationPublishEvaluatesRulesInEveryRegion` — a publish whose `Authorization` header isn't SigV4 (a bearer token) counts as region-less too.
- `topicRuleActionTargetsTheRuleRegion` — same queue name in both regions; an `eu-west-1` rule's `sqs` action lands in the `eu-west-1` queue and leaves `us-east-1` empty.

All five fail against unmodified `src/main`. `./mvnw test -Dtest='Iot*Test'` goes from 50 to 55, 0 failures.

## Checklist

- [x] `./mvnw test` passes locally *(with pre-existing Docker-dependent failures unrelated to this change — no IoT class fails)*

Five new tests in `IotTopicRuleIntegrationTest`, plus three on the resolver helper (suite 50 → 55, green, no Docker): a rule matches a publish signed for its own region; a publish signed for another region does not match; an unsigned publish reaches a rule outside the default region; and one region-less publish with matching rules in two regions fires both — pinning the fan-out trade-off above so changing it is a deliberate edit, not drift. The SQS sink test asserts delivery lands in the rule's region queue and not the default region's. With `src/main` stashed, the region tests fail against the old behavior and pass with the fix.

### DIFF
--- a/docs/services/iot.md
+++ b/docs/services/iot.md
@@ -124,7 +124,7 @@ Supported rule behavior:
 - MQTT-style topic filter matching for exact topics, `+`, and terminal `#`.
 - IoT Data `Publish` and MQTT publishes use the same rule dispatch path.
 - Rule matching is region-scoped: an IoT Data `Publish` evaluates the rules of the region named by its SigV4 credential, and a rule's actions target the rule's own region.
-- Publishes that carry no region — MQTT, or an unsigned IoT Data `Publish` — are evaluated against every region's rules.
+- Publishes that carry no region — MQTT, or an IoT Data `Publish` whose `Authorization` header is absent or not SigV4 — are evaluated against every region's rules.
 - `republish` action republishes the original payload to another MQTT topic through `IotMqttBrokerService`.
 - `sqs` action sends the original payload to an SQS queue through Floci's SQS service boundary.
 - `sns` action publishes the original payload to an SNS topic through Floci's SNS service boundary.
--- a/docs/services/iot.md
+++ b/docs/services/iot.md
@@ -123,6 +123,8 @@ Supported rule behavior:
 - SQL topic filter extraction for rules shaped like `SELECT * FROM 'topic/filter'`.
 - MQTT-style topic filter matching for exact topics, `+`, and terminal `#`.
 - IoT Data `Publish` and MQTT publishes use the same rule dispatch path.
+- Rule matching is region-scoped: an IoT Data `Publish` evaluates the rules of the region named by its SigV4 credential, and a rule's actions target the rule's own region.
+- Publishes that carry no region — MQTT, or an unsigned IoT Data `Publish` — are evaluated against every region's rules.
 - `republish` action republishes the original payload to another MQTT topic through `IotMqttBrokerService`.
 - `sqs` action sends the original payload to an SQS queue through Floci's SQS service boundary.
 - `sns` action publishes the original payload to an SNS topic through Floci's SNS service boundary.
--- a/src/main/java/io/github/hectorvent/floci/core/common/RegionResolver.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/RegionResolver.java
@@ -67,6 +67,21 @@ public class RegionResolver {
         return defaultRegion;
     }
 
+
+    /**
+     * Resolves the region from a SigV4 Authorization credential scope, or
+     * null when the header is absent or carries no parseable credential —
+     * for callers that treat an unresolved region differently from the
+     * default region.
+     */
+    public String resolveRegionFromAuthOrNull(String authorizationHeader) {
+        if (authorizationHeader == null || authorizationHeader.isEmpty()) {
+            return null;
+        }
+        Matcher matcher = CREDENTIAL_REGION_PATTERN.matcher(authorizationHeader);
+        return matcher.find() ? matcher.group(1) : null;
+    }
+
     public String getDefaultRegion() {
         return defaultRegion;
     }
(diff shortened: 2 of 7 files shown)


hectorvent:
Thank you, this is a real bug and the fix is scoped well. Threading the publisher's region through rather than reaching for the default in `handlePublish` is right, and having rule actions target the rule's own region (via the rule ARN) is a second, quieter improvement that matters just as much for a fan-out to SQS or SNS.

Treating a non-SigV4 `Authorization` header as unresolved rather than defaulting is also the right call. A Bearer token is not a region, and defaulting there would have reintroduced the same bug for authorized but unsigned callers.

**(follow-up, separate PR)** One coordination note, nothing for you to do: #2149 adds `isRegionUnresolved` to `RegionResolver`, which is the same predicate as your `resolveRegionFromAuthOrNull` returning null. They do not conflict textually, but I would rather not keep both in `core/common`, so I will collapse them when the second one lands.

No blockers from my side.

hectorvent (approved):

PR #2173, opened 2026-08-08
fix: serialize Kinesis stream creation timestamps as plain decimals

### LINKED ISSUE #2099: [BUG]
## Service

Kinesis

## AWS API Action

kinesis.describeStream

## Expected behavior

kinesis.describeStream should work

## Actual behavior

DescribeStream timestamp serialized in scientific notation, breaks AWS SDK unmarshalling in Java SDK

## Reproduction

Floci version/image: floci/floci:1.5.34

How to reproduce:
KinesisClient kinesis = KinesisClient.builder()
    .endpointOverride(URI.create("http://localhost:4566"))
    .region(Region.of("us-east-1"))
    .credentialsProvider(StaticCredentialsProvider.create(
        AwsBasicCredentials.create("test", "test")))
    .build();

kinesis.createStream(b -> b.streamName("my-stream").shardCount(1));
kinesis.describeStream(b -> b.streamName("my-stream")); // throws SdkClientException

Example failure:
software.amazon.awssdk.core.exception.SdkClientException: Unable to unmarshall response
  (Unable to parse date : 1.785732980986E9). Response Code: 200, Response Text: OK
Caused by: java.lang.NumberFormatException: For input string: "1.785732980986E9"
  at software.amazon.awssdk.utils.DateUtils.parseUnixTimestampMillisInstant
Floci serializes the StreamCreationTimestamp field as scientific notation (1.785732980986E9) instead of a plain decimal Unix timestamp, which AWS SDK v2's Long.parseLong()-based date parser can't handle. Reproduced consistently

## Environment

- Floci version/image: floci/floci:1.5.34

Java 17
latest AWS SDK version
Floci docker

mvanhorn:
## Summary

Replace the floating-point conversion at both Kinesis stream-description response builders with an exact decimal epoch-seconds representation derived from epoch milliseconds, retaining millisecond precision without permitting exponent notation in JSON output. Keep the correction local to `KinesisJsonHandler`; do not alter global Jackson settings, the stored `Instant`, CBOR routing, or other timestamp fields that are outside the reported stream-description contract. Add focused handler coverage for both `DescribeStream` and `DescribeStreamSummary`, asserting that the timestamp remains a numeric value and that serialization produces a plain decimal lexeme.

`DescribeStream` currently converts `StreamCreationTimestamp` from epoch milliseconds by dividing through a `double`, so Jackson may serialize the resulting JSON number in scientific notation. AWS SDK for Java v2 then fails to unmarshal values such as `1.785732980986E9`, despite the response returning HTTP 200. The same conversion is duplicated in `DescribeStreamSummary`, making that response vulnerable to the same protocol incompatibility. The issue includes a concrete SDK reproduction against Floci 1.5.34, and the local handler and tests provide direct file anchors for a bounded fix.

Closes #2099

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

## Checklist


aniketshukla1 (approved):
Reviewed exact scale-three timestamp conversion in both response builders. Passed `KinesisJsonHandlerTest,KinesisIntegrationTest` under JDK 25.

pgermosen:
Hi @mvanhorn, thanks for tracking this down — the fix itself (`epochSeconds(Instant)` via `BigDecimal`) is correctly implemented and the tests pass (63/63 across `KinesisJsonHandlerTest` + `KinesisIntegrationTest`). A couple of things before this can land:

**1. Merge conflict against current `main`.** It's minor — just an import-ordering collision in `KinesisJsonHandlerTest.java` — but it needs a rebase.

**2. Heads up on an overlap with #2359.** That PR (same underlying defect class, different field — `ApproximateArrivalTimestamp` on `GetRecords`/`SubscribeToShard` instead of `StreamCreationTimestamp` here) independently adds its own private `epochSeconds(Instant)` helper to this same class. Whichever of the two merges second is going to hit a duplicate-method conflict on that name and will need to dedupe against the other. Worth coordinating with whoever's PR lands first.

**3. One thing worth knowing, not necessarily blocking:** I ran the actual server from this branch with the fix applied and hit it with the real, unmodified `KinesisClient` (`createStream` → `describeStreamSummary`). The SDK-unmarshalling-exception this PR is meant to prevent doesn't actually reproduce against that client — floci routes Kinesis responses through a CBOR conversion layer by default (which the real SDK v2 client also uses by default), and that layer encodes `StreamCreationTimestamp` as a binary CBOR value, never touching the JSON scientific-notation text this PR avoids. So the fix is still a legitimate and correct improvement for the plain-JSON-protocol path (e.g. clients with `AWS_CBOR_DISABLE=true`), just not for the default client the issue's reproduction used.

More concerning: through that same CBOR path, the decoded `streamCreationTimestamp` comes back wrong by roughly 1000x regardless of whether this fix is applied — I saw a `1970-01-21` date instead of the real current time. That's a separate, pre-existing bug (looks like a units mismatch in the CBOR tag(1) fractional-timestamp encoding/decoding), not something this PR causes or needs to fix, but it means the real-world impact described in the issue may not be what a default SDK v2 client actually experiences today.

None of this should block merging the fix as-is — just flagging it so the scope is accurately understood.

pgermosen (approved):
@mvanhorn — approving. The conflict is cleared and the fix holds up.

## Verified

On current `main` (`65b22ff0`): zero commits behind, merges cleanly, compiles, and **108 Kinesis tests pass**. All fourteen checks green.

`BigDecimal.valueOf(millis, 3)` is the right shape — an exactly scaled decimal with no floating point anywhere in the path, applied at both response builders rather than one. Reverting it to the division fails `wholeSecondCreationTimestampRemainsNumeric`, so the behaviour is held in place rather than merely accompanied by a test.

Keeping the change local to `KinesisJsonHandler` rather than reaching for a global Jackson setting was the right instinct. A serialization-wide change to avoid scientific notation would have altered every numeric field in every service to fix two, and that is the kind of fix that is discovered months later by something unrelated breaking.

## A note for whoever merges: #2359 will need to dedupe

#2359 is still open and adds its own private `epochSeconds(Instant)` helper to this same class, for `ApproximateArrivalTimestamp` on `GetRecords` and `SubscribeToShard` — the same defect in a different field.

`epochSeconds` is not in `main` today, so this PR lands first cleanly and #2359 inherits a duplicate-method collision on that name. Nothing for you to do here; it just means #2359 should drop its copy and use yours rather than the two being reconciled by hand at merge time. I will raise it there so it is not a surprise.

## The scope point from earlier now has a home

I mentioned that the SDK unmarshalling failure this prevents does not reproduce against a default `KinesisClient`, because Floci routes Kinesis through a CBOR layer that encodes the timestamp as a binary value and never produces the JSON text this avoids — and that through that same path the decoded value comes back roughly 1000x wrong regardless of this fix.

That second part is now tracked as **#2368**, so it is recorded rather than folklore. None of it changes the merits here: the plain-JSON path is a real path, clients running with `AWS_CBOR_DISABLE=true` use it, and the previous code could emit `1.785732980986E9` on it. This fix is correct for the path it addresses, and the CBOR defect is somebody else's change.

## One small thing, not blocking

The AWS Compatibility section says "Not applicable to this change", but this is squarely an AWS compatibility fix — an SDK failing to unmarshal a response the service returned. Your Summary already explains it properly, including the concrete `1.785732980986E9` value, so nothing is missing in substance; it is only that the section a reviewer scans first says the opposite of what the PR does.

Thank you for the patience on this one — it has been open since the 8th, and the delay was the queue rather than anything here.

PR #2175, opened 2026-08-08
fix(s3vectors): return cosine distance from queries

### LINKED ISSUE #2162: [BUG] S3 Vectors: cosine QueryVectors returns similarity in the distance field (sense inverted)
## Service

S3 Vectors

## AWS API Action

`QueryVectors` (with `--return-distance`), for indexes created with `--distance-metric cosine`

## Expected behavior

Real AWS S3 Vectors returns a cosine **distance**, i.e. `1 - cosineSimilarity`:

- a vector identical to the query has distance `0.0`
- a less similar vector has a **larger** value
- results are ordered nearest-first, ascending by distance

## Actual behavior

The `distance` field carries the cosine **similarity** instead, so the sense is inverted: an identical vector scores `~1.0` and a dissimilar one scores lower.

```json
{
  "vectors": [
    { "distance": 1.0000000000000002, "key": "v1" },
    { "distance": 0.3740621411907388, "key": "v2" }
  ],
  "distanceMetric": "cosine"
}
```

`v1` here is byte-identical to the query vector, so AWS would report `0.0` for it.

The ordering of the returned list happens to be correct (most similar first), so this is easy to miss — but any client that reads the numeric value breaks. Two common patterns that silently produce wrong results:

- applying a distance threshold (`keep results where distance < X`) — inverted, so it keeps exactly the wrong ones
- re-sorting a merged candidate pool ascending by distance — reverses relevance

In the source, the cosine branch assigns the similarity directly, while the Euclidean branch is correct:

```java
// S3VectorsService.java — queryVectors
distance = calculateEuclideanDistance(queryVector, v.getData());   // correct
...
distance = calculateCosineSimilarity(queryVector, v.getData());    // returns similarity, not distance
```

## Reproduction

```bash
E=http://localhost:4566

  --vectors '[{"key":"v1","data":{"float32":[0.1,0.2,0.3,0.4]},"metadata":{}},
              {"key":"v2","data":{"float32":[0.9,0.1,0.0,0.2]},"metadata":{}}]'

# Query with a vector identical to v1 -> expected distance 0.0 for v1, actual ~1.0
aws --endpoint-url $E s3vectors query-vectors \
  --query-vector '{"float32":[0.1,0.2,0.3,0.4]}' \
  --top-k 2 --return-distance
```

## Environment

- Client: AWS CLI 2.36.16 (also reproduces via boto3)

---

The fix looks like returning `1 - similarity` for the cosine metric and sorting ascending, matching the existing Euclidean path. Glad to open a PR if that is welcome.

aniketshukla1:
## Summary

- Return AWS-compatible cosine distance from S3 Vectors `QueryVectors` instead of cosine similarity.
- Order cosine query results nearest-first by their reported distance.
- Add integration coverage for exact-match and orthogonal-vector distances.

Closes #2162

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Cosine query results now report `1 - cosineSimilarity` in `distance`: an
identical vector is `0.0`, an orthogonal vector is `1.0`, and results are
ordered by ascending distance.

## Checklist

- [ ] `./mvnw test` passes locally (not run; targeted integration suite run below)

## Validation

- `JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw test -Dtest=S3VectorsIntegrationTest` (12 passed)
- Confirmed the regression test fails when the previous similarity behavior is restored.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/S3VectorsTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/S3VectorsTest.java
@@ -173,6 +173,7 @@ class S3VectorsTest {
                         .queryVector(VectorData.builder().float32(1.0f, 0.1f, 0.0f).build())
                         .topK(1)
                         .returnMetadata(true)
+                        .returnDistance(true)
                         .build()
         );
         assertThat(response.vectors()).hasSize(1);
--- a/src/main/java/io/github/hectorvent/floci/services/s3vectors/S3VectorsController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3vectors/S3VectorsController.java
@@ -394,11 +394,13 @@ public class S3VectorsController {
         );
 
         boolean returnMetadata = request.returnMetadata() != null && request.returnMetadata();
+        boolean returnDistance = request.returnDistance() != null && request.returnDistance();
 
         List<QueryResultRepresentation> reps = results.stream()
                 .map(res -> {
                     Map<String, Object> metadata = returnMetadata ? res.getVector().getMetadata() : null;
-                    return new QueryResultRepresentation(res.getVector().getKey(), res.getDistance(), metadata);
+                    Double distance = returnDistance ? res.getDistance() : null;
+                    return new QueryResultRepresentation(res.getVector().getKey(), distance, metadata);
                 })
                 .toList();
 
--- a/src/main/java/io/github/hectorvent/floci/services/s3vectors/S3VectorsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3vectors/S3VectorsService.java
@@ -182,20 +182,13 @@ public class S3VectorsService {
                     break;
                 case "cosine":
                 default:
-                    distance = calculateCosineSimilarity(queryVector, v.getData());
+                    distance = calculateCosineDistance(queryVector, v.getData());
                     break;
             }
             results.add(new QueryResult(v, distance));
         }
 
-        // Sorting:
-        // For Euclidean distance, smaller distance is closer (ascending)
-        // For Cosine similarity, larger score is closer (descending)
-        if ("euclidean".equals(metric)) {
-            results.sort(Comparator.comparingDouble(QueryResult::getDistance));
-        } else {
-            results.sort((r1, r2) -> Double.compare(r2.getDistance(), r1.getDistance()));
-        }
+        results.sort(Comparator.comparingDouble(QueryResult::getDistance));
 
         return results.stream().limit(topK).collect(Collectors.toList());
     }
@@ -214,6 +207,11 @@ public class S3VectorsService {
         }
         if (normA == 0.0 || normB == 0.0) return 0.0;
         return dotProduct / (Math.sqrt(normA) * Math.sqrt(normB));
+    }
+
+    private double calculateCosineDistance(List<Float> v1, List<Float> v2) {
+        double similarity = calculateCosineSimilarity(v1, v2);
+        return 1.0 - Math.max(-1.0, Math.min(1.0, similarity));
     }
 
     private double calculateEuclideanDistance(List<Float> v1, List<Float> v2) {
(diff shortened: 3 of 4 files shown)


aniketshukla1:
@hectorvent @pgermosen can we merge this ?

pgermosen (commented):
Thanks for this — the change reads well, and I appreciate that the test pins the exact values (`0.0` and `1.0`) rather than just asserting an ordering. Collapsing the two sort branches into one ascending comparator is a nice simplification, and the `Math.max(-1.0, Math.min(1.0, …))` clamp is a thoughtful touch for float rounding.

Two small things I noticed nearby while reading. Neither is a problem with your change — both are pre-existing — but they're adjacent enough that it seemed more useful to mention them here than to let them sit.

**1. `returnDistance` is currently ignored**

Your test sets `"returnDistance": true`, which makes sense given what it's asserting. It turns out the parameter doesn't do anything today: `S3VectorsController:401` builds every `QueryResultRepresentation` with `res.getDistance()` unconditionally, so `distance` comes back whether or not the caller asked for it.

The [API reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_QueryVectors.html) documents it as *"Indicates whether to include the computed distance in the response. The default value is `false`"* — so a client that omits it should get results without a `distance` field at all.

Entirely up to you whether to pick that up here or leave it: gating on `request.returnDistance()` alongside the existing `returnMetadata` handling a few lines up would be a couple of lines and would sit naturally in this PR, but it's also a clean separate change and I don't want to grow the scope of a focused fix.

**2. `calculateDotProduct` looks unreachable**

`S3VectorsService:227` has no callers — the `switch` in `queryVectors` only branches on `euclidean` and `cosine`, and the model's `DistanceMetric` enum is exactly those two values, so there's no metric that could route to it. It was already dead before this PR; I only spotted it because your change is in the same block of helpers.

If you're touching this file anyway, removing it would keep the helpers matching the metrics that actually exist. Equally fine to leave it alone — just flagging so it's a deliberate choice rather than something nobody noticed.

Nothing here blocks; happy to keep going once you've had a look.

aniketshukla1:
Thanks for catching both of these @pgermosen .
I’ve updated the PR so `distance` is returned only when `returnDistance` is explicitly `true`,  the default response now omits the field entirely. I added integration coverage for that default behaviour while retaining the explicit-true distance assertions.
I also removed the unused `calculateDotProduct` helper, since no supported distance metric routes to it.

Validated with:
./mvnw test -Dtest=S3VectorsIntegrationTest

Please check now.
Thanks

pgermosen (approved):
Thanks for the quick and precise turnaround on both points — re-reviewed at `ef0142f4`.

**`returnDistance` is now honored properly.** The detail that makes it correct is one you didn't have to change: `QueryResultRepresentation` already declares `Double distance` (boxed) with `@JsonInclude(NON_NULL)`, so `Double distance = returnDistance ? res.getDistance() : null` genuinely omits the key rather than emitting `"distance": null`, and there's no unboxing surprise. Your test asserts `not(hasKey("distance"))` rather than a null check, which pins exactly that — the stronger of the two assertions, and the one that would catch a future change to the include policy.

That now matches the [API reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_QueryVectors.html): *"Indicates whether to include the computed distance in the response. The default value is `false`."*

**Nice catch on the compat suite.** `ef0142f4` adding `.returnDistance(true)` to `sdk-test-java`'s `S3VectorsTest` is the part I want to call out — honoring the flag changes the default response shape, and that test was relying on the old always-present behavior. Finding and fixing it rather than waiting for someone else to hit it is the right instinct, and it's a good illustration of why that compatibility suite earns its runtime.

**`calculateDotProduct` removed.** Straightforward, and correct — `DistanceMetric` is exactly `euclidean | cosine`, so nothing could ever route to it.

The original fix still reads well: collapsing the two sort branches into a single ascending comparator, the zero-norm guard already in `calculateCosineSimilarity`, and the `Math.max(-1.0, Math.min(1.0, …))` clamp for float rounding. The test pinning `0.0` and `1.0` exactly, rather than just asserting an ordering, is what makes the change verifiable.

All 13 checks green at this head, including `native / sdk-test-java`.

Approving — thanks for picking up two adjacent cleanups that weren't strictly yours to fix.

PR #2176, opened 2026-08-08
fix(ecs): report service deployments and MISSING failures on DescribeServices

### LINKED ISSUE #2174: [BUG] ECS DescribeServices omits deployments for an active service
## Service

ECS

## AWS API Action

DescribeServices

## Expected behavior

For an active ECS rolling-update service, `DescribeServices` returns a non-empty
`services[].deployments` array. The primary deployment includes the task
definition, desired/running/pending counts, and rollout state.

For example:

```json
{
  "services": [
    {
      "serviceName": "example-service",
      "desiredCount": 1,
      "runningCount": 1,
      "pendingCount": 0,
      "deployments": [
        {
          "id": "ecs-svc/example-deployment-id",
          "status": "PRIMARY",
          "taskDefinition": "arn:aws:ecs:us-east-1:000000000000:task-definition/example-task:1",
          "desiredCount": 1,
          "runningCount": 1,
          "pendingCount": 0,
          "rolloutState": "COMPLETED"
        }
      ]
    }
  ],
  "failures": []
}
```

## Actual behavior

Floci returns the service with `status: "ACTIVE"`, `desiredCount: 1`,
`runningCount: 1`, and `pendingCount: 0`, but omits the `deployments` field.
AWS SDK clients consequently expose `deployments` as null/absent, so callers
cannot determine whether the primary deployment reached `COMPLETED`.

Sanitized response:

```json
{
  "services": [
    {
      "serviceArn": "arn:aws:ecs:us-east-1:000000000000:service/example-cluster/example-service",
      "serviceName": "example-service",
      "clusterArn": "arn:aws:ecs:us-east-1:000000000000:cluster/example-cluster",
      "taskDefinition": "arn:aws:ecs:us-east-1:000000000000:task-definition/example-task:1",
      "desiredCount": 1,
      "runningCount": 1,
      "pendingCount": 0,
      "status": "ACTIVE"
    }
  ],
  "failures": []
}
```

The separate service-deployment API support does not provide the legacy
`DescribeServices.services[].deployments` response expected by AWS-compatible
clients.

## Reproduction

With Floci running and an ECS rolling-update service named `example-service`
active in `example-cluster` with desired count 1:

```bash
aws --endpoint-url http://localhost:4566 \
  --region us-east-1 \
  ecs describe-services \
  --cluster example-cluster \
  --services example-service \
  --query 'services[0].{status:status,desiredCount:desiredCount,runningCount:runningCount,pendingCount:pendingCount,deployments:deployments}'
```

The counts show `1/1/0`, but `deployments` is null/absent.

## Environment

- Floci version / image tag: `floci/floci:1.6.0` (`sha256:eab36252ea43a4a73928423f0372219052c5c6f87207f6c4754db14b91d6ed30`)
- Java SDK version (if applicable): N/A; reproduced with AWS CLI v2

Current `main` at `c0f36ea60467b9df5898f8cd71229b4c4b3314c7` also appears affected: the
service serializer does not add a `deployments` field.

hectorvent:
## Summary

`DescribeServices` omitted `services[].deployments` entirely (#2174) and always returned an empty `failures` list. Both fields are read by AWS's own `ServicesStable` waiter, so this was not a cosmetic gap.

The waiter, from [botocore `ecs/2014-11-13/waiters-2.json`](https://github.com/boto/botocore/blob/develop/botocore/data/ecs/2014-11-13/waiters-2.json), accepts on:

```
length(services[?!(length(deployments) == `1` && runningCount == desiredCount)]) == `0`
```

and fails fast on `failures[].reason == "MISSING"`. That file is codegen'd into every AWS SDK, so with `deployments` absent:

- `aws ecs wait services-stable` polled the full 40 x 15s and then failed.
- The same waiter failed in every SDK (Java `EcsWaiter.waitUntilServicesStable`, etc).
- Terraform's `aws_ecs_service` never converged, via the same `len(Deployments) == 1` test in `internal/service/ecs/service.go:2206`.
- Waiting on a service that does not exist also hung for the full timeout instead of failing immediately, because the `MISSING` failure was never emitted.

Evaluating the acceptor against the old payload does not merely return false, it raises `JMESPathTypeError`, since `length(null)` is a type error.

An `ACTIVE` service now reports exactly one `PRIMARY` deployment derived from current service state, with `rolloutState` `COMPLETED` once `runningCount` reaches `desiredCount`. An unresolvable service reference comes back as a `MISSING` failure carrying the ARN it would have had.

One design note worth flagging for review: the deployment `id` is derived from the service ARN plus task definition rather than generated per request. Deployments are synthesized on read, so a random id would differ on every `DescribeServices` call and read as perpetual drift in any client that persists it. Deriving it keeps the id stable across calls and restarts while still rolling over when the task definition changes. There is a regression test for exactly that.

Closes #2174

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior:** `services[].deployments` was absent from every `DescribeServices` response, and `failures` was hardcoded to an empty array so unknown services were silently dropped.

Verified against the model rather than against our own expectations:

- **Shape** — `Service.deployments` and the `Deployment` members (`id`, `status`, `taskDefinition`, `desiredCount`, `pendingCount`, `runningCount`, `failedTasks`, `rolloutState`, `rolloutStateReason`, `createdAt`, `updatedAt`, `launchType`) all match [botocore ecs/2014-11-13](https://github.com/boto/botocore/blob/develop/botocore/data/ecs/2014-11-13/service-2.json). `DeploymentRolloutState` is `COMPLETED | FAILED | IN_PROGRESS`; `Deployment.status` is `PRIMARY | ACTIVE | INACTIVE`; `Failure` is `arn` / `reason` / `detail`.
- **`ecs-svc/<id>` id format** is not documented in botocore. Grounded instead in moto (`moto/ecs/models.py:527,883,1710` and its service-deployment ARN regex at `:2060`).
- **`reason: "MISSING"`** likewise comes from moto (`describe_services`, `models.py:1673`) and is corroborated by the waiter's own failure acceptor.
- **Verified with AWS CLI v2** against a local Floci: the issue's exact reproduction now returns the field, and `aws ecs wait services-stable` returns in 1s where it previously ran to timeout.
- **Verified with AWS SDK for Java v2** in `sdk-test-java`, driving the real `EcsWaiter.waitUntilServicesStable` rather than asserting on fields, so the genuine JMESPath acceptor is exercised.

Known deviations are documented in `docs/services/ecs.md` rather than half-implemented: only one deployment (no draining `ACTIVE` alongside `PRIMARY`), no `deploymentController` gating (Floci does not record it yet, and `ECS` is the AWS default), `pendingCount` always 0, and `forceNewDeployment` not minting a new id.

## Checklist

Full suite: **8545 tests, 0 failures**. New `EcsDescribeServicesIntegrationTest` (9 cases) covers the exact member names, id stability across calls, id rollover on task-definition change, rollout state transitions, the empty list for an `INACTIVE` service, and the `MISSING` failure paths. `EcsTests` in `sdk-test-java` gains deployment assertions, a `ServicesStable` waiter test, and a `MISSING` failure test.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/EcsTests.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/EcsTests.java
@@ -1,9 +1,12 @@
 package com.floci.test;
 
 import org.junit.jupiter.api.*;
+import software.amazon.awssdk.core.waiters.WaiterOverrideConfiguration;
+import software.amazon.awssdk.core.waiters.WaiterResponse;
 import software.amazon.awssdk.services.ecs.EcsClient;
 import software.amazon.awssdk.services.ecs.model.*;
 
+import java.time.Duration;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.*;
@@ -773,6 +776,25 @@ class EcsTests {
         assertThat(services).hasSize(1);
         assertThat(services.get(0).serviceName()).isEqualTo(serviceName);
         assertThat(services.get(0).desiredCount()).isEqualTo(1);
+
+        // An ACTIVE service must report exactly one PRIMARY deployment. AWS's ServicesStable
+        // waiter accepts on length(deployments) == 1, so an empty list silently breaks every
+        // SDK waiter and Terraform's aws_ecs_service (floci-io/floci#2174).
+        List<Deployment> deployments = services.get(0).deployments();
+        assertThat(deployments).hasSize(1);
+        assertThat(deployments.get(0).status()).isEqualTo("PRIMARY");
+        assertThat(deployments.get(0).id()).startsWith("ecs-svc/");
+        assertThat(deployments.get(0).taskDefinition()).isEqualTo(services.get(0).taskDefinition());
+        assertThat(deployments.get(0).desiredCount()).isEqualTo(1);
+        assertThat(deployments.get(0).rolloutState())
+                .isIn(DeploymentRolloutState.COMPLETED, DeploymentRolloutState.IN_PROGRESS);
+
+        // Deployments are synthesized per request, so the id must not drift between calls.
+        List<Service> again = ecs.describeServices(DescribeServicesRequest.builder()
+                .cluster(clusterName)
+                .services(serviceName)
+                .build()).services();
+        assertThat(again.get(0).deployments().get(0).id()).isEqualTo(deployments.get(0).id());
     }
 
     @Test
@@ -807,6 +829,46 @@ class EcsTests {
 
     @Test
     @Order(45)
+    @DisplayName("ServicesStable waiter - reaches stable")
+    void servicesStableWaiter() {
+        // The strongest available check on services[].deployments: this drives AWS's own
+        // waiter acceptor,
+        //   length(services[?!(length(deployments) == `1` && runningCount == desiredCount)]) == `0`
+        // rather than our reading of it. With deployments absent the expression never matches.
+        // waitTimeout is capped so a regression fails in 30s rather than the default 40 x 15s.
+        WaiterResponse<DescribeServicesResponse> response = ecs.waiter().waitUntilServicesStable(
+                DescribeServicesRequest.builder()
+                        .cluster(clusterName)
+                        .services(serviceName)
+                        .build(),
+                WaiterOverrideConfiguration.builder()
+                        .waitTimeout(Duration.ofSeconds(30))
+                        .build());
+
+        assertThat(response.matched().response()).isPresent();
+        assertThat(response.matched().response().get().services().get(0).deployments()).hasSize(1);
+    }
+
+    @Test
+    @Order(46)
+    @DisplayName("DescribeServices - unknown service returns a MISSING failure")
+    void describeServicesReportsMissingService() {
+        // ECS reports an unresolvable service as a failure rather than erroring, and the
+        // ServicesStable waiter fails fast on failures[].reason == MISSING. Dropping it
+        // silently would leave that waiter polling for its full timeout.
+        DescribeServicesResponse response = ecs.describeServices(DescribeServicesRequest.builder()
+                .cluster(clusterName)
+                .services("no-such-service-" + suffix)
+                .build());
+
+        assertThat(response.services()).isEmpty();
+        assertThat(response.failures()).hasSize(1);
+        assertThat(response.failures().get(0).reason()).isEqualTo("MISSING");
+        assertThat(response.failures().get(0).arn()).endsWith("no-such-service-" + suffix);
+    }
+
+    @Test
+    @Order(47)
     @DisplayName("ListServiceDeployments - list deployments")
     void listServiceDeployments() {
         List<ServiceDeploymentBrief> briefs = ecs.listServiceDeployments(
@@ -830,7 +892,7 @@ class EcsTests {
     }
 
     @Test
-    @Order(46)
+    @Order(48)
     @DisplayName("UpdateService - update desiredCount to 0")
     void updateServiceDesiredCount() {
         Service updated = ecs.updateService(UpdateServiceRequest.builder()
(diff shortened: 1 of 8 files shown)


hectorvent:
Good catch, and it was right. Confirmed at the source before changing anything: `updateService` never touched `createdAt`, and `EcsServiceModel` had no `updatedAt` at all, so a task-definition change minted a new deployment id while both timestamps still reported the service's creation time. The payload contradicted itself.

The model settles which one is wrong. `Deployment.createdAt` is documented as "the Unix timestamp for the time when the service **deployment** was created" ([ecs service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/ecs/2014-11-13/service-2.json)), and moto stamps it at deployment construction rather than reusing the service's (`moto/ecs/models.py:540`).

Fixed in 8fb44f2 by tracking when the current deployment began on the service model, set at creation and on every task-definition change. Services persisted before the field existed fall back to the service creation time. Added a regression test asserting the deployment `createdAt` advances past both the previous deployment's and the service's.

Two notes for the record:

- `updatedAt` still equals `createdAt`. AWS advances it as a rollout progresses, and Floci has no intermediate rollout state to report, so that one is now documented as a known deviation rather than left to look intentional.
- Writing the test surfaced an unrelated trap worth knowing about: RestAssured's default number type is `FLOAT_AND_DOUBLE`, and a 10-digit epoch second does not fit in a float. Two timestamps a second apart both rounded to the same value, including through `getString`, which made a working fix look broken. The test reads them as `BigDecimal`.

Full suite: 8546 tests, 0 failures.

pgermosen (approved):
Looks good

- **`Deployment`/`Failure`/`DeploymentRolloutState` shapes**: pulled `botocore/data/ecs/2014-11-13/service-2.json` directly. `Deployment`'s members, `DeploymentRolloutState`'s enum (`COMPLETED | FAILED | IN_PROGRESS`), and `Failure`'s members (`arn`/`reason`/`detail`) all match what's implemented here exactly — this PR's field set is a subset of the real one (omits `capacityProviderStrategy`, `networkConfiguration`, `serviceConnectConfiguration`, etc.), consistent with the "known deviations" the docs and PR description both call out rather than silently claiming full parity.
- **`ecs-svc/<id>` format and `reason: "MISSING"`**: confirmed against moto's `ecs/models.py` (`f"ecs-svc/{mock_random...}"` at lines 529/885/1735, and `{"arn": missing_arn, "reason": "MISSING"}` at line 1698) — exactly what the PR cites, not just a plausible-sounding claim.
- **Timestamp serialization** (`toEpochMilli() / 1000.0` for `createdAt`/`updatedAt`) matches the identical pattern already used for every other timestamp field in this same file (tasks, task sets, deployments elsewhere) — not a new convention.
- **The second commit's fix is real and tested**: verified `lastDeploymentAt` is a new, non-required field (Jackson defaults it to `null` on deserializing older persisted records with no special annotation needed), and `deploymentsFor` correctly falls back to `svc.getCreatedAt()` when it's absent — matching the stated backward-compat story. `newDeploymentCarriesItsOwnStartTimeNotTheServiceCreationTime` genuinely exercises this with a real 1.1s sleep and asserts the new deployment's `createdAt` is later than both the old one and the service's own `createdAt`.
- **The `ServicesStable` waiter test drives the real SDK waiter** (`ecs.waiter().waitUntilServicesStable(...)`), not a hand-rolled assertion of the acceptor logic — the strongest verification available, and it's capped to a 30s timeout so a regression fails fast instead of taking the default 40×15s. (I didn't re-run the `sdk-test-java` compat module myself — it needs the packaged app running behind a real SDK client, more setup than the in-repo integration test — but the test code itself is doing the right thing.)

Docs (`docs/services/ecs.md`) are honest about every known deviation (no draining second deployment, no `deploymentController` gating, `pendingCount` always 0, `forceNewDeployment` not minting a new id, `updatedAt == createdAt`) rather than half-implementing silently.

PR #2178, opened 2026-08-08
feat(ec2): add ReplaceRoute

### LINKED ISSUE #2087: [FEAT]  Add support for EC2 ReplaceRoute API
## Service

EC2

## API Action / Feature

`ReplaceRoute`

## AWS Documentation

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ReplaceRoute.html

## Why is this needed?

Terraform updates existing routes in route tables by calling the EC2 `ReplaceRoute` API. While using Floci, route table creation succeeds, but subsequent updates fail because `ReplaceRoute` is not implemented.

Example Terraform configuration:

```hcl
resource "aws_route_table" "private" {
  count  = 3
  vpc_id = aws_vpc.main.id

  route {
    cidr_block     = "0.0.0.0/0"
    nat_gateway_id = aws_nat_gateway.this[count.index].id
  }
}
```

When running `terraform apply`, Floci returns:

```text
Error: updating Route in Route Table (...) with destination (0.0.0.0/0):
operation error EC2: ReplaceRoute,
api error UnsupportedOperation: Operation ReplaceRoute is not supported.
```

Terraform uses `ReplaceRoute` whenever an existing route needs to be updated, so this prevents normal Terraform workflows from modifying route tables. The only workaround is to recreate the route table instead of updating it.

Supporting `ReplaceRoute` would improve compatibility with Terraform and other infrastructure-as-code tools.

soreavis:
## Summary

Adds EC2 `ReplaceRoute`. Terraform calls it whenever an existing route changes target, so today a route table can be created but never updated — `terraform apply` fails with `UnsupportedOperation: Operation ReplaceRoute is not supported` and the only workaround is to recreate the table. Closes #2087.

The action replaces the target of the route matching `DestinationCidrBlock`. AWS takes **exactly one** target per call, so the target a request does not name is cleared rather than carried over — a merge would leave a route holding two targets at once, which AWS cannot produce. `origin` is preserved, since `ReplaceRoute` is not one of the documented origin values.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Wire shape is taken from the [ReplaceRoute API reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ReplaceRoute.html): request params, the `requestId` + `return` response (so it uses the existing `booleanResponse` helper, same as `CreateRoute`/`DeleteRoute`), and the "exactly one of the resources from the parameter list" rule that makes this replace-not-merge. Error codes come from the [EC2 error reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html), since the ReplaceRoute page lists no action-specific errors:

| condition | code |
|---|---|
| no route matches the destination | `InvalidRoute.NotFound` |
| route table does not exist | `InvalidRouteTableID.NotFound` (existing behaviour) |
| `DestinationCidrBlock` absent — an IPv6 or prefix-list request | `MissingParameter` |
| two targets, or none | `InvalidParameterCombination` |
| a target this emulator does not model | `UnsupportedOperation` |

Exercised with the AWS CLI against a local emulator in addition to the integration tests, including `--local-target` and `--no-local-target`. The tests drive the Query protocol directly through RestAssured rather than an SDK client — that matches the EC2 package, where none of the existing integration tests use `software.amazon.awssdk`.

**Scope — please read before reviewing the target list.** `Route` models a gateway and a NAT gateway. `LocalTarget` needs no new field either, because `local` is simply the gateway id a route table's built-in route already carries, so resetting to it is expressible and AWS's documented repoint-then-reset flow works end to end. The remaining ten targets (`TransitGatewayId`, `NetworkInterfaceId`, `VpcPeeringConnectionId`, `CarrierGatewayId`, `CoreNetworkArn`, `EgressOnlyInternetGatewayId`, `InstanceId`, `LocalGatewayId`, `OdbNetworkArn`, `VpcEndpointId`) are rejected with `UnsupportedOperation` before any mutation. That is deliberate: accepting them would have meant reporting success while clearing the target of a route that was previously correct, which is strictly worse than the loud failure callers get today. Supporting them properly means growing `Route`, which belongs in its own change.

Two things I would rather flag than have you find:

- **`LocalTarget=true` on a route other than the local one** is accepted here and produces a second route targeting `local`. AWS restricts the local target to the VPC's own CIDR. I left it unguarded because nothing in the surrounding code validates this class of thing — gateway existence, CIDR normalisation and duplicate destinations are all unchecked on `CreateRoute` today — so a check only here would be inconsistent. Happy to add one if you would rather.
- **The `AwsQueryController` action-list entry is a consistency fix, not a requirement.** Any request carrying an ec2-scoped SigV4 credential resolves the service before that list is read, so the reporter's case works without it; it matters only for requests with no resolvable credential scope.

## Checklist

`Ec2ReplaceRouteIntegrationTest` is new — 11 tests. Full `-Dtest='Ec2*Test'` is **287 passing, 0 failures** across 15 classes on JDK 25. Each guard was checked by mutating it and confirming a test fails: reverting replace-not-merge to a merge, dropping the exactly-one-target check, hardcoding `origin`, removing the destination guard, disabling the unsupported-target loop, and turning the `LocalTarget` parse back into a presence check — each is caught by exactly one test.

One note on the docs row: `docs/services/ec2.md` has no `floci:actions` markers and `Ec2QueryHandler` sits under `deferred_handlers` in `tools/docs/services.yaml`, so the table is hand-maintained. I ran `regen_action_docs.py --strict` and the file is byte-identical afterwards, so the added row is exactly what the generator would emit.

### DIFF
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -254,6 +254,7 @@ Floci seeds the following resources on first use in each region so Terraform, th
 | AssociateRouteTable | Associates a route table with a subnet. |
 | DisassociateRouteTable | Removes a route table association. |
 | CreateRoute | Adds a route to a route table. |
+| ReplaceRoute | Replaces the target of an existing route. |
 | DeleteRoute | Removes a route from a route table. |
 
 ### Network ACLs
--- a/src/main/java/io/github/hectorvent/floci/core/common/AwsQueryController.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/AwsQueryController.java
@@ -156,7 +156,7 @@ public class AwsQueryController {
             "CreateInternetGateway", "DescribeInternetGateways", "DeleteInternetGateway",
             "AttachInternetGateway", "DetachInternetGateway",
             "CreateRouteTable", "DescribeRouteTables", "DeleteRouteTable",
-            "AssociateRouteTable", "DisassociateRouteTable", "CreateRoute", "DeleteRoute",
+            "AssociateRouteTable", "DisassociateRouteTable", "CreateRoute", "ReplaceRoute", "DeleteRoute",
             "CreateNetworkAcl", "DescribeNetworkAcls", "DeleteNetworkAcl",
             "CreateNetworkAclEntry", "ReplaceNetworkAclEntry", "DeleteNetworkAclEntry",
             "ReplaceNetworkAclAssociation",
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
@@ -28,6 +28,15 @@ public class Ec2QueryHandler {
     private static final DateTimeFormatter ISO_FMT = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
             .withZone(ZoneOffset.UTC);
 
+    /** ReplaceRoute targets AWS accepts that a stored {@code Route} cannot represent here. */
+    private static final List<String> UNSUPPORTED_ROUTE_TARGETS = List.of(
+            "CarrierGatewayId", "CoreNetworkArn", "EgressOnlyInternetGatewayId", "InstanceId",
+            "LocalGatewayId", "NetworkInterfaceId", "OdbNetworkArn",
+            "TransitGatewayId", "VpcEndpointId", "VpcPeeringConnectionId");
+
+    /** The gateway id a route table's built-in route carries (see Ec2Service#createRouteTable). */
+    private static final String LOCAL_GATEWAY_ID = "local";
+
     private final Ec2Service service;
     private final EmulatorConfig config;
     private final FlowLogService flowLogService;
@@ -123,6 +132,7 @@ public class Ec2QueryHandler {
                 case "AssociateRouteTable" -> handleAssociateRouteTable(params, region);
                 case "DisassociateRouteTable" -> handleDisassociateRouteTable(params, region);
                 case "CreateRoute" -> handleCreateRoute(params, region);
+                case "ReplaceRoute" -> handleReplaceRoute(params, region);
                 case "DeleteRoute" -> handleDeleteRoute(params, region);
                 // Network ACLs
                 case "CreateNetworkAcl" -> handleCreateNetworkAcl(params, region);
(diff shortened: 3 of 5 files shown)


pgermosen (approved):
Looks good

- **Request/response shape and the "exactly one target" rule**: pulled `botocore/data/ec2/2016-11-15/service-2.json` directly (had to go via `raw.githubusercontent.com` — the file's 4MB, past the Contents API's inline-content limit). `ReplaceRouteRequest`'s 13 target-shaped members enumerate to *exactly* the 3 implemented (`GatewayId`, `NatGatewayId`, `LocalTarget`) plus the 10 explicitly rejected (`TransitGatewayId`, `NetworkInterfaceId`, `VpcPeeringConnectionId`, `CarrierGatewayId`, `CoreNetworkArn`, `EgressOnlyInternetGatewayId`, `InstanceId`, `LocalGatewayId`, `OdbNetworkArn`, `VpcEndpointId`) — no omissions, no extras. The docs page's "must also specify exactly one of the resources from the parameter list, or reset the local route to its default target" is a verbatim match to what's implemented.
- **`origin` preservation**: confirmed `RouteOrigin`'s real enum is `CreateRouteTable | CreateRoute | EnableVgwRoutePropagation | Advertisement` — `ReplaceRoute` genuinely isn't a valid origin value, so preserving the existing route's origin rather than overwriting it is correct, not just a plausible-sounding claim.
- **The disclosed scope gap holds up under its own reasoning**: the PR says `CreateRoute` doesn't validate gateway existence/CIDR normalization/duplicate destinations either, so `ReplaceRoute` not guarding `LocalTarget=true` on a non-local route "would be inconsistent" to add unilaterally — checked `handleCreateRoute` myself and confirmed it has no existence validation of any kind, so that's accurate, not just asserted.
- **Docs claim verified by actually running the tool**, not trusted: `python3 tools/docs/regen_action_docs.py --strict` exits 0 and produces zero diff against `docs/services/ec2.md` on this branch, confirming the added row is exactly what the generator would emit.
- **The `AwsQueryController` action-list addition** is a straightforward, correctly-placed one-line addition consistent with the existing list's grouping (next to `CreateRoute`).

Ran the tests myself rather than trusting the numbers: `Ec2ReplaceRouteIntegrationTest` 11/11, and the full `Ec2*Test` glob (287 tests across all EC2 classes) — 287/287, 0 failures, matching the PR's own count exactly. The unsupported-target test is a genuinely good one: it doesn't just assert the 400, it follows up with a `DescribeRouteTables` call confirming the route's `natGatewayId` is unchanged, directly validating the "reject rather than silently clear" design goal instead of just the status code. The `LocalTarget=false` case is also explicitly tested (not just presence-checked), matching the PR's claim about that being one of the mutation-tested guards.

PR #2179, opened 2026-08-08
feat(iam): add service-linked role create, delete and deletion status

### LINKED ISSUE #2122: [BUG] IAM: CreateServiceLinkedRole returns UnsupportedOperation, failing aws_iam_service_linked_role
## Service

IAM

## AWS API Action

`CreateServiceLinkedRole`

## Expected behavior

`CreateServiceLinkedRole` creates a service-linked role for the given service principal. Terraform's `aws_iam_service_linked_role` uses it directly; LocalStack supports it.

## Actual behavior

```
api error UnsupportedOperation: Operation CreateServiceLinkedRole is not supported.
```

## Why this matters

`aws_iam_service_linked_role` is common in account-bootstrap Terraform (OpenSearch, ECS, ElastiCache and others each want one). A single unsupported operation fails the whole module apply, blocking every downstream module that depends on it.

A no-op-but-well-formed implementation would be enough for local use: the role has no behavioural meaning in an emulator, but Terraform needs the create to succeed and the role to be readable afterwards.

## Reproduction

```bash

aws iam create-service-linked-role --aws-service-name es.amazonaws.com
# -> UnsupportedOperation
```

Terraform equivalent:

```hcl
resource "aws_iam_service_linked_role" "es" {
  aws_service_name = "es.amazonaws.com"
}
```

## Environment

`floci/floci:1.5.34`, linux/arm64, macOS host. Terraform 1.7.5, provider `hashicorp/aws` v5.86.1.

soreavis:
## Summary

Adds `CreateServiceLinkedRole`, `DeleteServiceLinkedRole` and `GetServiceLinkedRoleDeletionStatus`. `aws_iam_service_linked_role` calls the create directly, and a single unsupported operation fails a whole account-bootstrap module, blocking every downstream module that depends on it. Closes #2122.

The issue scopes this to "a no-op-but-well-formed implementation" — the create must succeed and the role must be readable afterwards. I included delete and the status poll because Terraform's destroy path calls them, so create alone would leave `terraform destroy` broken, which seemed a worse place to stop than not starting.

The role lands at `/aws-service-role/<principal>/` with a trust policy for that principal and is returned by `GetRole` afterwards.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Parameters, response shapes and error lists come from the API reference for each of the three actions. Deletion task ids use the documented `task/aws-service-role/<service-principal-name>/<role-name>/<task-uuid>` format, and `Status` is one of the documented values. The path convention is confirmed three ways: the IAM User Guide's policy-example ARN (`arn:aws:iam::*:role/aws-service-role/{{SERVICE-NAME}}.amazonaws.com/...`), the segment ordering inside the task-id format, and real service-linked role ARNs. Tests drive the Query protocol directly through RestAssured, matching every other IAM test in the repo.

**The role name is a deterministic stand-in, and this is the caveat worth your attention.** The prefix is chosen by the linked service and is not computable from the principal — AWS turns `lex.amazonaws.com` into `AWSServiceRoleForLexBots` and `es.amazonaws.com` into `AWSServiceRoleForAmazonElasticsearchService`. Rather than guess per service, this derives a name from the principal's labels: `es.amazonaws.com` becomes `AWSServiceRoleForEs`. Terraform round-trips whatever ARN it is handed, so this works, but the name will not match AWS for most services. If you would rather have a lookup table for the common principals with this as the fallback, say so and I will add one.

Every label contributes rather than just the first, because `rds.amazonaws.com` and `rds.application-autoscaling.amazonaws.com` are separate roles on AWS — a config declaring both must not collide on a single name.

**`CustomSuffix` is joined with an underscore.** That is the separator callers parse back out: the Terraform provider recovers `custom_suffix` with `strings.Split(roleName, "_")` when the split length is exactly 2, and `custom_suffix` is `ForceNew`, so a name joined any other way never converges — it would destroy and recreate on every apply.

Deviations and limits, all deliberate:

| behaviour | why |
|---|---|
| duplicate role returns `InvalidInput`, not `createRole`'s usual `EntityAlreadyExists` | `EntityAlreadyExists` is absent from this action's published error list; `InvalidInput` is on it |
| delete errors map onto `NoSuchEntity` | that action documents only `NoSuchEntity`, `LimitExceeded` and `ServiceFailure`, so `InvalidInput` would be off-contract |
| status is always `SUCCEEDED` | deletion is synchronous here, so no other state is reachable; `Reason` is omitted, which AWS returns only on failure |
| `AWSServiceName` validated against `[\w+=,.@-]{1,128}` | AWS's own constraint; it also keeps a crafted principal out of the trust-policy JSON |
| `CustomSuffix` is **not** validated | no IAM name is validated anywhere in this service — `CreateRole`, `CreateUser` and `CreatePolicy` all accept the same shapes — so adding it only here would not match the surrounding code. Consequence: a long principal can push the role name past AWS's 64-character cap |

Deletion task ids live in a `StorageFactory`-backed store rather than a field, so they are account-scoped and survive a restart like every other IAM record. That was worth doing properly: an in-memory map let one account read another's task, and lost the id across a restart mid-`terraform destroy`.

## Checklist

`IamServiceLinkedRoleIntegrationTest` is new — 16 tests, covering the create/read/delete/poll round trip, the underscore contract, rejection of names outside AWS's pattern, and cross-account isolation of a deletion task. Full `-Dtest='Iam*Test,Sts*Test,AssumeRole*Test'` is **228 passing, 0 failures** across 16 classes on JDK 25. Guards were checked by mutation: disabling the service-linked path check and disabling the task-registry lookup each break a test.

The three `docs/services/iam.md` rows are hand-added — that table has no `floci:actions` markers and `IamQueryHandler` is under `deferred_handlers` in `tools/docs/services.yaml`. `regen_action_docs.py --strict` leaves the file byte-identical, so the rows match generator output.

The new package-private constructor exists so the store can be injected without touching the eleven existing `new IamService(...)` call sites in tests; the shorter constructors delegate with an in-memory backend, the same seam `SnsService` uses.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
@@ -6,6 +6,7 @@ import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.core.common.RegionResolver;
 import io.github.hectorvent.floci.core.common.SessionAccountLookup;
 import io.github.hectorvent.floci.core.storage.AccountAwareStorageBackend;
+import io.github.hectorvent.floci.core.storage.InMemoryStorage;
 import io.github.hectorvent.floci.core.storage.StorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageFactory;
 import io.github.hectorvent.floci.services.iam.model.AccessKey;
@@ -31,6 +32,8 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
+import java.util.regex.Pattern;
 
 /**
  * Core IAM business logic — users, groups, roles, policies, access keys, instance profiles.
@@ -52,6 +55,11 @@ public class IamService implements SessionAccountLookup {
     private static final String DEFAULT_DEPLOYER_USER = "floci-deployer";
     private static final String DEFAULT_DEPLOYER_ACCESS_KEY_ID = "floci";
     private static final String DEFAULT_DEPLOYER_SECRET_ACCESS_KEY = "floci";
+    private static final String SERVICE_LINKED_ROLE_PATH = "/aws-service-role/";
+    private static final String SERVICE_LINKED_ROLE_NAME_PREFIX = "AWSServiceRoleFor";
+    private static final String AMAZONAWS_DOMAIN = ".amazonaws.com";
+    /** AWSServiceName as AWS constrains it: 1-128 characters of {@code [\w+=,.@-]}. */
+    private static final Pattern SERVICE_PRINCIPAL_PATTERN = Pattern.compile("[\\w+=,.@-]{1,128}");
 
     private final StorageBackend<String, IamUser> users;
     private final StorageBackend<String, IamGroup> groups;
@@ -60,6 +68,8 @@ public class IamService implements SessionAccountLookup {
     private final StorageBackend<String, AccessKey> accessKeys;
     private final StorageBackend<String, InstanceProfile> instanceProfiles;
     private final StorageBackend<String, SessionCredential> sessions;
+    /** Deletion is synchronous, so an issued task id is a completed one; the value is its role. */
+    private final StorageBackend<String, String> serviceLinkedRoleDeletions;
     private final RegionResolver regionResolver;
     private final boolean seedDeployerPrincipal;
 
@@ -80,6 +90,7 @@ public class IamService implements SessionAccountLookup {
             storageFactory.create("iam", "iam-access-keys.json", new TypeReference<>() {}),
             storageFactory.create("iam", "iam-instance-profiles.json", new TypeReference<>() {}),
             storageFactory.create("iam", "iam-sessions.json", new TypeReference<>() {}),
+            storageFactory.create("iam", "iam-slr-deletions.json", new TypeReference<>() {}),
             regionResolver,
             config.services().iam().seedDeployerPrincipal()
         );
@@ -105,6 +116,20 @@ public class IamService implements SessionAccountLookup {
                StorageBackend<String, SessionCredential> sessions,
                RegionResolver regionResolver,
                boolean seedDeployerPrincipal) {
+        this(users, groups, roles, policies, accessKeys, instanceProfiles, sessions,
+                new InMemoryStorage<>(), regionResolver, seedDeployerPrincipal);
+    }
+
+    IamService(StorageBackend<String, IamUser> users,
+               StorageBackend<String, IamGroup> groups,
+               StorageBackend<String, IamRole> roles,
+               StorageBackend<String, IamPolicy> policies,
+               StorageBackend<String, AccessKey> accessKeys,
+               StorageBackend<String, InstanceProfile> instanceProfiles,
+               StorageBackend<String, SessionCredential> sessions,
+               StorageBackend<String, String> serviceLinkedRoleDeletions,
+               RegionResolver regionResolver,
+               boolean seedDeployerPrincipal) {
         this.users = users;
         this.groups = groups;
         this.roles = roles;
@@ -112,6 +137,7 @@ public class IamService implements SessionAccountLookup {
         this.accessKeys = accessKeys;
         this.instanceProfiles = instanceProfiles;
         this.sessions = sessions;
+        this.serviceLinkedRoleDeletions = serviceLinkedRoleDeletions;
         this.regionResolver = regionResolver;
         this.seedDeployerPrincipal = seedDeployerPrincipal;
     }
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/hectorvent/floci/services/iam/IamService.java
```diff
@@ -393,6 +419,96 @@ public void deleteRole(String roleName) {
...
+        IamRole role = getRole(roleName);
+        // The path cannot classify a role — CreateRole will put an ordinary one under the
+        // service-role prefix — so only roles minted here are deletable through this action. The
+        // error is NoSuchEntity because that is what this action's published list carries.
+        if (!role.isServiceLinkedRole()) {
+            throw new AwsException("NoSuchEntity",
+                    "There is no service-linked role with name " + roleName + ".", 404);
+        }
+        String path = role.getPath();
+        deleteRole(roleName);
```
pgermosen: `deleteRole(roleName)` throws `AwsException("DeleteConflict", ..., 409)` (line 415-416) if the role still has attached managed or inline policies — and nothing here (or anywhere in `AttachRolePolicy`/`PutRolePolicy`) stops a caller from attaching a policy to a service-linked role after `CreateServiceLinkedRole` mints it, since it's just a regular `IamRole` under the hood with a marker flag. If that happens, `DeleteServiceLinkedRole` would propagate `DeleteConflict` (409) straight through — an error code that isn't in this action's documented list (`LimitExceeded | NoSuchEntity | ServiceFailure`), the same off-contract problem the PR is explicitly careful to avoid everywhere else (the `InvalidInput`-vs-`NoSuchEntity` split you documented is precisely about not leaking an error type outside an action's published list).

This is narrow — real Terraform usage via `aws_iam_service_linked_role` wouldn't normally attach extra policies to a role it just created for this purpose — and it's adjacent to a deviation you already deliberately accepted (synchronous deletion, so `FAILED`/`IN_PROGRESS` are unreachable and this is really the only way `DeleteServiceLinkedRole` *can* fail here). Not blocking, but worth a one-line guard if you want the error contract airtight, e.g. catching `DeleteConflict` from `deleteRole` and remapping it to `NoSuchEntity` or a synthesized `FAILED` status, matching the pattern real AWS uses (deletion fails asynchronously with a `Reason` when the role is still in use, rather than the create-time exception path).

pgermosen (commented):
Looks good

- **Every claim about parameters, response shapes, and — critically — the asymmetric error lists between the three actions checks out exactly.** Pulled [CreateServiceLinkedRole](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateServiceLinkedRole.html), [DeleteServiceLinkedRole](https://docs.aws.amazon.com/IAM/latest/APIReference/API_DeleteServiceLinkedRole.html), and [GetServiceLinkedRoleDeletionStatus](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetServiceLinkedRoleDeletionStatus.html) directly. `AWSServiceName`/`CustomSuffix` pattern (`[\w+=,.@-]+`) and length constraints (1-128 / 1-64) match `SERVICE_PRINCIPAL_PATTERN` exactly. `CreateServiceLinkedRole`'s documented errors are `InvalidInput | LimitExceeded | NoSuchEntity | ServiceFailure` — no `EntityAlreadyExists` — confirming the PR's claim about using `InvalidInput` for a duplicate. `DeleteServiceLinkedRole`'s documented errors are `LimitExceeded | NoSuchEntity | ServiceFailure` — no `InvalidInput` — confirming the `NoSuchEntity` choice there. The `DeletionTaskId` format (`task/aws-service-role/<principal>/<role>/<uuid>`) matches the docs' own worked example byte-for-byte, and `Status`'s valid values (`SUCCEEDED | IN_PROGRESS | FAILED | NOT_STARTED`) and the `Reason`-only-on-failure behavior both check out against the docs' two example responses.
- **The role-naming derivation is correct on its own stated terms**: traced `derivedServiceName` by hand for both examples the PR gives (`es.amazonaws.com` → `AWSServiceRoleForEs`) and the collision case (`rds.amazonaws.com` → `AWSServiceRoleForRds` vs. `rds.application-autoscaling.amazonaws.com` → `AWSServiceRoleForRdsApplicationAutoscaling`) — confirmed these genuinely produce distinct names, and the new `principalsSharingALeadingLabelGetDistinctRoles` test pins exactly this.
- **The second commit's fix is real and well-tested.** Confirmed `IamRole.serviceLinkedRole` is a plain `boolean` (defaults `false`, so old persisted records read as not-service-linked without any migration) and that `deleteServiceLinkedRole` now keys on `role.isServiceLinkedRole()` rather than the path. `anOrdinaryRoleAtAServiceLinkedPathIsNotDeletable` is a genuine regression test for exactly this: it creates an ordinary role via plain `CreateRole` at `/aws-service-role/es.amazonaws.com/`, confirms `DeleteServiceLinkedRole` rejects it with `NoSuchEntity`, and confirms the role is still there afterward via `GetRole`.

@soreavis 

One real, narrow gap below — adjacent to the sync-vs-async deviation the PR already deliberately documents, so worth a look

soreavis:
@pgermosen good catch on the 409 — it reproduces. I went a different way on the fix though. AWS publishes `UnmodifiableEntity` (400) on `AttachRolePolicy` and `PutRolePolicy` for service-linked roles, so the state you found isn't reachable there; remapping the `DeleteConflict` would emulate something AWS blocks a step earlier. So the attach is refused instead, and `DeleteServiceLinkedRole` no longer routes through `deleteRole`.

Chasing it turned up worse. `DeleteRole` removed a service-linked role outright at 200, `UpdateAssumeRolePolicy` rewrote its trust policy, and `AddRoleToInstanceProfile` attached it to an instance profile — all silent. AWS publishes that error on twelve role actions; this now guards the eleven we implement (`UpdateRoleDescription` has no handler). `TagRole` and `UntagRole` stay open, as on AWS.

pgermosen:
Verified this properly rather than taking it on trust — went back to the actual IAM API references for each action, not just the ones I'd already checked.

- `AttachRolePolicy` and `DeleteRole` both document `UnmodifiableEntity` (400) with exactly the text your commit implies: "service-linked roles are protected AWS resources... the error message includes the name of the service that depends on this service-linked role." My original scenario (attach a policy, then hit `DeleteConflict` on delete) genuinely isn't reachable on real AWS, since the attach itself is refused first. Good catch, and the right fix.
- Checked the harder, more specific claims rather than the easy ones: `TagRole`'s error list has no `UnmodifiableEntity`, and its description literally says *"The role can be a regular role or a service-linked role"* — word for word what your code comment cites. `UntagRole` likewise has no `UnmodifiableEntity` in its error list. `UpdateRoleDescription` does have it, and `grep`-ing `IamQueryHandler.java` confirms there's genuinely no handler for it here, matching "the twelfth, unimplemented."
- Spot-checked `RemoveRoleFromInstanceProfile` too — `UnmodifiableEntity` is there, consistent with the pairing.
- Counted the guarded methods against the commit message's list: `deleteRole`, `updateRole`, `updateAssumeRolePolicy`, `attachRolePolicy`, `detachRolePolicy`, `putRolePolicy`, `deleteRolePolicy`, `addRoleToInstanceProfile`, `removeRoleFromInstanceProfile`, `putRolePermissionsBoundary`, `deleteRolePermissionsBoundary` — exactly eleven, exactly matching what's named.

Ran the tests myself: `IamServiceLinkedRoleIntegrationTest` 25/25 (the new `theOtherRoleActionsAwsProtectsAreAlsoRefused` batch-tests 8 of the 11 guarded actions in one pass, plus `taggingAServiceLinkedRoleIsStillAllowed` is a genuine regression-guard against over-blocking, not just a happy-path check), and the broader `Iam*Test,Sts*Test,AssumeRole*Test` glob — 237/237, no regressions in the ordinary-role paths that now route through `requireNotServiceLinked` unconditionally. `regen_action_docs.py --strict` still exits clean.

This is a genuinely better fix than what I suggested — it's not a patch around the symptom, it closes the actual root cause (mutations that AWS itself refuses a step earlier) and does it with the same error-contract precision as the rest of the PR. Nothing further from me — good to merge as far as this thread goes.

pgermosen (approved):
Looks good

Issue #2091, opened 2026-08-02
[BUG] Runtime API socket closes before SIGTERM reaches container — awslambdaric/Node runtime print a fault per clean invocation


gwyn-f:
## Service

Lambda

## AWS API Action

Lambda Runtime API — `GET /2018-06-01/runtime/invocation/next` (shutdown handshake)

## Expected behavior

When AWS Lambda tears down an idle execution environment (the [Shutdown phase](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtime-environment.html#runtimes-lifecycle)), it [sends `SIGTERM` to the runtime process](https://github.com/aws-samples/graceful-shutdown-with-aws-lambda). On Linux, SIGTERM's default action terminates the process. Neither [`awslambdaric`](https://github.com/aws/aws-lambda-python-runtime-interface-client) (Python) nor the Node.js runtime interface client installs a custom signal handler ([acknowledged by AWS](https://github.com/aws/aws-lambda-python-runtime-interface-client/issues/105) — users are expected to register their own if they want cleanup logic). Both rely on the OS default, and the parked `GET /invocation/next` long-poll gets aborted as part of process termination.

## Actual behavior

Floci closes the runtime API socket *before* the container process receives SIGTERM. That means the parked long-poll wakes with a network error — HTTP 204 (returned to waiting pollers when Floci's `RuntimeApiServer.stop()` sets `stopped=true`; [204 is not among the documented AWS responses](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next) for `/invocation/next`) for a still-connected client, or an abrupt connection close if the server socket is already gone. Either way, the runtime library's blocking HTTP call raises an exception — [`FaultException(LAMBDA_RUNTIME_CLIENT_ERROR)` in Python](https://github.com/aws/aws-lambda-python-runtime-interface-client/blob/main/awslambdaric/lambda_runtime_client.py#L143-L157) or the ["Failed to get next invocation" error in Node](https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/nodejs24.x/src/client/rapid-client.ts#L252) — and that exception surfaces as a traceback on stderr while the process is exiting. It's cosmetic — the invocation result was already delivered — but it prints on every clean invocation, buries real errors, and gives an incorrect impression of failure.

[...]

PR #2181, opened 2026-08-09
fix(lambda): disable HTTP/2 cleartext on the runtime API server

### LINKED ISSUE #2180: [BUG] RuntimeApiServerTest.extensionRegister_racedAgainstStop_eventuallyDeliversShutdown flakes with EOFException, including after #2156
## Service

Lambda (Extensions API / Runtime API server) — CI test, not a wire-protocol bug, so the AWS-CLI reproduction section of the bug template does not apply.

## Summary

`RuntimeApiServerTest.extensionRegister_racedAgainstStop_eventuallyDeliversShutdown` fails intermittently with an `EOFException` from the JDK HTTP/2 client, **including on a build that already contains the #2156 fix**. One error out of 8,553 tests; a different PR from the same base passed the full suite six minutes earlier.

## Actual behavior

```
[ERROR] RuntimeApiServerTest.extensionRegister_racedAgainstStop_eventuallyDeliversShutdown -- Time elapsed: 0.556 s <<< ERROR!
java.util.concurrent.ExecutionException: java.io.EOFException: EOF reached while reading
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:440)
	at ...RuntimeApiServerTest.extensionRegister_racedAgainstStop_eventuallyDeliversShutdown(RuntimeApiServerTest.java:1166)
Caused by: java.io.EOFException: EOF reached while reading
	at java.net.http/jdk.internal.net.http.Http2Connection$Http2TubeSubscriber.onComplete(Http2Connection.java:1903)
	at java.net.http/jdk.internal.net.http.SocketTube$InternalReadPublisher$ReadSubscription.signalCompletion(SocketTube.java:652)
```

The client's `next` request completes exceptionally with EOF instead of receiving the SHUTDOWN event — i.e. the connection went away rather than the event arriving late.

## Observations

| when | build | result |
|---|---|---|
| 08-08 03:38 | `ff8d6efd` on `main` (`feat(ses): implement v1 receipt rule set actions (stored-but-inert) (#2128)`) — **before** #2156 | same test, same `EOFException` ([run](https://github.com/floci-io/floci/actions/runs/31237072459)) |
| 08-08 05:33 | `c0f36ea6` on `main` — the #2156 merge itself | passed |
| 08-08 21:59 | a PR branched from `c0f36ea6`, so **with** #2156 | same test, same `EOFException` ([run](https://github.com/floci-io/floci/actions/runs/31280021389/job/93159747288)) |
| 08-08 21:5x | a second PR from the same base, same full suite | passed |

So it is intermittent rather than a clean regression, and it survives #2156.

## Why this might still be worth a look

The test dates from #1773, and #2142/#2156 fixed the ordering bug it guards, so a residual failure is either (a) a narrower race still left in `stop()` — the counter defers the close until each write's `end()` future completes, but that only means the frame was written, not that it reached this client before the connection went away — or (b) the test's own timing, if it can observe the close before the dispatch without the production path being wrong. I cannot tell which from the outside, so I am reporting the observation rather than proposing a fix. #2098 is open against this same path and splits `stop()` into `quiesce()`/`close()`, so if that lands the analysis may change — I have not tested against it. If it is (a), the same EOF would reach a real extension as a dropped SHUTDOWN.

## Environment

- `main` @ `c0f36ea6` and one commit before it, plus PR branches based on `c0f36ea6`
- GitHub Actions `Build and Test`, JDK 25
- Not reproduced locally; both observations are CI

For the record: I hit this on an unrelated IAM PR whose diff touches no Lambda or runtime file, so I do not think the two are connected — filing it separately rather than as noise on that PR.

pgermosen:
## Summary

Closes #2180.

`RuntimeApiServerTest.extensionRegister_racedAgainstStop_eventuallyDeliversShutdown` flaked intermittently (~1 in 8500 runs) with an `EOFException` from the JDK HTTP/2 client, surviving #2156's fix for the write-vs-close ordering bug. `stop()` already waits for each parked write's flush to complete before closing the `httpServer` (that's what #2156 added), but a completed flush only confirms the bytes were handed to the local socket, not that the peer received them before the connection is torn down — the HTTP/2 stream/connection-close bookkeeping for a multiplexed h2c connection can still race that same close.

**Fix**: disable HTTP/2 cleartext (`setHttp2ClearTextEnabled(false)`) on the `RuntimeApiServer`'s `HttpServerOptions`. Real AWS's Runtime/Extensions API is plain HTTP/1.1 — floci doesn't need HTTP/2 here at all. Disabling h2c removes the whole class of multiplexed-connection races this flake falls into; the JDK `HttpClient` (which defaults to attempting an h2c upgrade over plaintext) automatically falls back to a normal HTTP/1.1 request when the server doesn't offer the upgrade, so no client-side or test change was needed.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: floci was accepting HTTP/2 (h2c) connections on the Runtime/Extensions API server, which real AWS Lambda's runtime API never speaks (it's plain HTTP/1.1). That divergence is what let this flake happen at all — HTTP/1.1's simpler one-request-per-connection-turn semantics don't have the multiplexed stream-vs-connection-close ordering this race depended on.

## Checklist

### Testing details

No new test — the existing racing test (`extensionRegister_racedAgainstStop_eventuallyDeliversShutdown`, 300 iterations/run) is the regression coverage. Ran it 5x (1500 iterations total) with this change, 0 failures, plus the full `RuntimeApiServerTest` class and the rest of the lambda module's test suite (`ContainerLauncherTest`, `LambdaImageConfigTest`, `WarmPoolTest`, `LambdaExecutorServiceTest`) — all green. 1500 iterations doesn't fully prove a ~1-in-8500 flake is gone, so I'd like this to run in CI for a while before calling it conclusively fixed; happy to add an h2c-specific regression test if reviewers want tighter coverage than "the flake didn't reproduce."

Also checked PR #2098 (open, restructures `stop()` into `quiesce()`/`close()` for a different bug, #2091) — it doesn't touch this code path or fix this race, so this change is independent of it.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
@@ -451,7 +451,18 @@ public class RuntimeApiServer {
     private void tryListen(CompletableFuture<Void> started, Router router, long deadline) {
         if (started.isDone()) return;
         httpServer = vertx.createHttpServer(new HttpServerOptions()
-                .setMaxFormAttributeSize(-1));
+                .setMaxFormAttributeSize(-1)
+                // Real AWS's Runtime/Extensions API is plain HTTP/1.1. Vert.x defaults to
+                // accepting an h2c upgrade, which the JDK HttpClient used in tests and by some
+                // language runtimes opts into unless told otherwise. That multiplexes every
+                // request from one client (e.g. an extension's register + its own /event/next
+                // long-poll) onto a single connection/stream-set, so stop()'s close of that
+                // connection can race a just-flushed response's HTTP/2 stream-completion
+                // bookkeeping instead of tearing down an already-idle connection — see #2180.
+                // Disabling h2c makes the client fall back to a plain HTTP/1.1 request
+                // automatically (no client-side change needed) and matches AWS's actual wire
+                // protocol.
+                .setHttp2ClearTextEnabled(false));
         httpServer.requestHandler(router).listen(port, "0.0.0.0", result -> {
             if (result.succeeded()) {
                 LOG.infov("RuntimeApiServer started on port {0}", String.valueOf(port));


pgermosen:
Verified the AWS-compatibility claim in the description rather than asserting it from memory:

- AWS's [Runtime API docs](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html) show every endpoint as a plain `curl http://${AWS_LAMBDA_RUNTIME_API}/...` example — no HTTP/2 anywhere in the spec.
- The official [`awslambdaric`](https://github.com/aws/aws-lambda-python-runtime-interface-client) (the actual client AWS runs inside the Lambda execution environment) connects with Python's `http.client.HTTPConnection` — HTTP/1.1 only, no HTTP/2 support in that library at all.

So real Lambda runtimes never speak HTTP/2 to the Runtime API. Floci accepting an h2c upgrade was a divergence from AWS (Vert.x's default, not anything intentional here) — disabling it is a fidelity fix, not a workaround, and it happens to close off the multiplexed-connection race behind #2180 as a side effect.

hectorvent:
Thank you, and thank you for closing the AWS question with sources rather than an assertion. That `awslambdaric` uses `http.client.HTTPConnection`, which has no HTTP/2 support at all, is the strongest form that argument could take: the client AWS actually runs cannot speak the protocol we were offering.

I checked the premise independently: run 31280021389 fails with the same `Http2Connection` EOF on a branch that does contain #2156's `remainingWrites`, so the flake really does survive it.

**(follow-up, separate PR)** I would still like a guard on the toggle itself. Nothing asserts it stays set, so a Vert.x default change would quietly bring the flake back. Driving the client with `version(HTTP_2)` and asserting `HTTP_1_1` comes back is about five lines, and beats "it did not reproduce in 1500 iterations".

No blockers; the unticked test box with your explanation is the right answer.

hectorvent (approved):

PR #2189, opened 2026-08-09
refactor(s3): move website folder-resolution policy into S3Service

### LINKED ISSUE #2140: refactor(s3): move website folder-resolution policy from S3Controller into S3Service
Follow-up requested by @hectorvent in [review of #1815](https://github.com/floci-io/floci/pull/1815#issuecomment-5200526773):

> **(follow-up, separate PR)** Moving the ~90 lines of folder-resolution policy from `S3Controller` into `S3Service` would keep the controller thin.

Agreed. Filing it so the commitment outlives #1815's merge.

## Scope

The website-endpoint policy #1815 adds to `S3Controller`:

- `serveWebsiteObject(bucket, key, authorization)` — the bulk of it: index resolution for directory requests, the folder 302, exact-object precedence, error-document fallback
- `serveWebsiteErrorResponse(bucket, authorization, cause)`
- `defaultWebsiteErrorResponse(status)`
- `headOnlyResponse(response)` — arguably stays, it is purely a JAX-RS concern

## Blocked on #1815

These methods do not exist on `main` — #1815 introduces them. A refactor branch opened now would have to carry #1815's commit as its base, so it would show both changes as one diff and conflict with #1815 until that merges. **This should be opened against `main` after #1815 lands**, not stacked on it.

## Not a straight move — needs a layering decision first

Worth flagging before someone picks this up expecting a cut-and-paste:

`S3Service` currently has **zero** `jakarta.ws.rs` or `io.vertx` imports, and `TextractService` is the only service in the repo that returns `jakarta.ws.rs.core.Response`. The code being moved depends on both layers:

- it reads the raw request path via `currentVertxRequest.getCurrent().request().path()`, because the routing layer strips the trailing slash from the object key and the slash is what distinguishes `/docs/` (serve `docs/index.html`) from `/docs` (302 to `/docs/`);
- it returns `Response`, including a 302 carrying a `Location` header.

Moving it verbatim would pull the HTTP layer into the service and break the convention the refactor is meant to strengthen.

Suggested shape instead — `S3Service` decides *policy*, `S3Controller` renders *HTTP*:

```java
// S3Service
sealed interface WebsiteResolution {
    record ServeObject(String key) implements WebsiteResolution {}
    record RedirectToDirectory(String path) implements WebsiteResolution {}
    record ErrorDocument(String key, int status) implements WebsiteResolution {}
    record DefaultError(int status) implements WebsiteResolution {}
    record NotAWebsite() implements WebsiteResolution {}
}

WebsiteResolution resolveWebsiteRequest(String bucket, String key, boolean directoryRequest,
                                        RequestAuthorization authorization);
```

The controller passes `directoryRequest` (derived from the raw path it already has) and turns the result into a `Response`. That also makes the policy unit-testable without standing up the HTTP layer, which is most of the value.

## Behaviour to preserve

`S3WebsiteIntegrationTest` (25 tests) pins the current contract and should stay green unchanged. Two cases are easy to get wrong and were settled empirically against a live S3 website endpoint in `us-east-1` during #1815 — see the probe table in [#1815's description](https://github.com/floci-io/floci/pull/1815):

- the folder 302 emits a **bare** `Location: /docs/` — the query string is **dropped**, matching real S3;
- a website endpoint has no S3 REST API, so a sub-resource-looking query such as `/docs/?tagging` stays an ordinary website request and serves `docs/index.html`.

avison9:
## Summary

Moves the website folder-resolution policy #1815 added to `S3Controller` down into `S3Service`,
the follow-up @hectorvent asked for in [review of #1815](https://github.com/floci-io/floci/pull/1815#issuecomment-5200526773)
("moving the ~90 lines of folder-resolution policy from `S3Controller` into `S3Service` would keep
the controller thin").

As #2140 flagged, this is deliberately **not** a verbatim move. The code being moved reads the raw
Vert.x request path and returns `jakarta.ws.rs.core.Response`, while `S3Service` has zero
`jakarta.ws.rs` and `io.vertx` imports — moving it as-is would pull the HTTP layer into the service
and break the convention the refactor exists to strengthen.

So the split is policy vs. rendering:

```java
// S3Service — decides WHAT to serve
public sealed interface WebsiteResolution {
    record ServeObject(String key, S3Object object)     implements WebsiteResolution {}
    record RedirectToDirectory()                        implements WebsiteResolution {}
    record ErrorDocument(S3Object object, int status)   implements WebsiteResolution {}
    record DefaultError(int status)                     implements WebsiteResolution {}
    record NotAWebsite()                                implements WebsiteResolution {}
}

WebsiteResolution resolveWebsiteRequest(String bucket, String key, boolean directoryRequest,
                                        RequestAuthorization authorization);
WebsiteResolution resolveWebsiteError(String bucket, RequestAuthorization authorization, int status);
```

`S3Controller` keeps the one input only it can supply — whether the client asked for a directory,
which the routing layer's trailing-slash stripping otherwise hides — and pattern-matches the
result into a `Response`.

Net effect: `S3Controller` **−44 lines** (+42 / −86), and `S3Service` still has **zero** HTTP-layer
imports.

Closes #2140

### Three deviations from the shape sketched in the issue

Each is a deliberate call, and easy to revert if you'd rather have the original shape:

1. **`RedirectToDirectory` carries no `path`.** The issue sketched `RedirectToDirectory(String path)`,
   but the redirect target is `rawPath + "/"` and only the controller knows `rawPath`. Putting it in
   the record would reintroduce exactly the HTTP dependency this refactor removes.
2. **`ServeObject` / `ErrorDocument` carry the already-read `S3Object`, not just a key.** The service
   has to read the object anyway to decide between serving it and falling back to the error document.
   Returning only a key would force a redundant second lookup in the controller and open a gap
   between the decision and the render.
3. **`isWebsiteErrorDocumentTrigger` moved to `S3Service` as `public static`.** "Which errors trigger
   the error document" is policy, and the controller has two other call sites for it — leaving a
   private copy behind would have duplicated the rule in both layers.

Also removed one line of dead code: `serveWebsiteObject` had
`return err != null ? err : xmlErrorResponse(e);`, but `serveErrorDocument` never returned `null`
(every path either returns a response or throws), so the fallback was unreachable.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore — specifically `refactor:`, per the type set the Conventional Commits workflow accepts

## AWS Compatibility

**No behaviour change** — this is a pure restructure, so there is no wire-protocol delta to verify.
The guarantee is that the contract #1815 established stays byte-identical:

- `S3WebsiteIntegrationTest` (25 tests) passes **completely untouched** — no test was edited,
  renamed, or re-ordered in this PR. That was the acceptance condition #2140 set, and it is the
  real evidence here.
- The two cases #1815 settled empirically against a live `us-east-1` website endpoint are preserved
  verbatim, with their explanatory comments carried across: the folder 302 emits a **bare**
  `Location: /docs/` with the query string dropped, and a sub-resource-looking query such as
  `/docs/?tagging` stays an ordinary website request serving `docs/index.html`.
- `403` keeps its own status; every other failure collapses to `404`, matching S3.

### Tests

`./mvnw test` — **8698 run, 0 failures, 0 errors, 9 skipped**. BUILD SUCCESS.
Full S3 scope (`-Dtest='S3*Test'`) — 714 run, 0 failures. `make docs-check` passes.

11 new cases in `S3ServiceTest` pin the policy directly, which is most of the value of the move —
before this, none of it could be exercised without standing up the HTTP layer:

| Test | Exercises |
|---|---|
| `resolveWebsiteRequestServesIndexForDirectoryRequest` | `/docs/` serves `docs/index.html` |
| `resolveWebsiteRequestServesRootIndexWhenKeyIsEmpty` | site root is a directory request even with no trailing slash |
| `resolveWebsiteRequestRedirectsFolderWithoutTrailingSlash` | `/docs` redirects when an index lives beneath it |
| `resolveWebsiteRequestPrefersExactObjectOverFolderRedirect` | an exact object hit wins over the folder redirect |
| `resolveWebsiteRequestFallsBackToErrorDocumentWhenIndexMissing` | missing index serves the custom error document |
| `resolveWebsiteRequestFallsBackToDefaultErrorWhenNoErrorDocumentConfigured` | no error document configured → built-in page |
| `resolveWebsiteRequestFallsBackToDefaultErrorWhenErrorDocumentItselfMissing` | configured but never uploaded → built-in page, not a 500 |
| `resolveWebsiteRequestIsNotAWebsiteWithoutConfiguration` | plain bucket falls through to the normal object path |
| `resolveWebsiteRequestPropagatesRealBucketErrors` | `NoSuchBucket` surfaces instead of being masked as "not a website" |
| `resolveWebsiteErrorReturnsNotAWebsiteWithoutConfiguration` | error path falls through for non-website buckets |
| `resolveWebsiteErrorMapsForbiddenToItsOwnStatus` | 403 keeps 403; 500 collapses to 404 |

These run in **0.5s** against **7.2s** for the HTTP-layer equivalents.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -808,7 +808,7 @@ public class S3Controller {
             return fullObjectResponse(bucket, key, versionId, obj, overrides, includeChecksum);
         } catch (AwsException e) {
             emitCloudTrailEvent("GetObject", bucket, key, 0L, 0L, e.getErrorCode(), e.getMessage());
-            if (isWebsiteErrorDocumentTrigger(e) && isWebsiteRequest(httpHeaders, uriInfo)) {
+            if (S3Service.isWebsiteErrorDocumentTrigger(e) && isWebsiteRequest(httpHeaders, uriInfo)) {
                 Response websiteError = serveWebsiteErrorResponse(bucket, authorization, e);
                 if (websiteError != null) {
                     return websiteError;
@@ -995,7 +995,7 @@ public class S3Controller {
             return resp.build();
         } catch (AwsException e) {
             emitCloudTrailEvent("HeadObject", bucket, key, 0L, 0L, e.getErrorCode(), e.getMessage());
-            if (isWebsiteErrorDocumentTrigger(e) && isWebsiteRequest(httpHeaders, uriInfo)) {
+            if (S3Service.isWebsiteErrorDocumentTrigger(e) && isWebsiteRequest(httpHeaders, uriInfo)) {
                 Response websiteError = serveWebsiteErrorResponse(bucket, authorization, e);
                 if (websiteError != null) {
                     return headOnlyResponse(websiteError);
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Verified the S3WebsiteIntegrationTest file is genuinely byte-identical to main, and it passes 25/25 exactly as claimed. Traced the WebsiteResolution sealed-interface split line by line against the original controller logic — index-serving, folder-redirect (bare Location header, query string dropped), error-document fallback, and the 403-stays/others-collapse-to-404 status handling all map 1:1 with no semantic drift. The claimed dead-code removal is correct — the removed fallback was genuinely unreachable.

Ran S3WebsiteIntegrationTest (25/25) and S3ServiceTest (55/55) locally, 0 failures. docs-check clean, all CI green, no CODEOWNERS-protected paths.

PR #2201, opened 2026-08-10
fix(s3): DeleteObjects honours empty key list

### LINKED ISSUE #2198: [BUG] DeleteObjects accepts empty key list instead of returning MalformedXM
## Service

S3

## AWS API Action

DeleteObjects (bulk delete, POST /{bucket}?delete)

## Expected behavior

Real S3 (and LocalStack) reject a DeleteObjects request with zero <Object> entries:
MalformedXML: The XML you provided was not well-formed. (HTTP 400).

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 s3api create-bucket --bucket test-bucket

curl -X POST "http://localhost:4566/test-bucket?delete" \
  -H "Content-Type: application/xml" \
  -d '<Delete><Quiet>true</Quiet></Delete>'
# Actual:   200 OK, <DeleteResult/>
# Expected: 400 MalformedXML
```

## Environment

- Floci version / image tag: 1.6.0
- Java SDK version (if applicable): n/a
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker floci/floci:1.6.0

kmahadevan-bidgely:
Fixes : #2198

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -1237,6 +1237,10 @@ public class S3Controller {
     private Response handleDeleteObjects(String bucket, byte[] body) {
         String xml = new String(body, StandardCharsets.UTF_8);
         List<String> keys = XmlParser.extractAll(xml, "Key");
+        if (keys.isEmpty()) {
+            throw new AwsException("MalformedXML",
+                    "The XML you provided was not well-formed.", 400);
+        }
         boolean quiet = XmlParser.containsValue(xml, "Quiet", "true");
         S3Service.DeleteObjectsResult result = s3Service.deleteObjects(bucket, keys);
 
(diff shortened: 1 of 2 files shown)


kmahadevan-bidgely:
@hectorvent - Please help take a look at this when you get some time.

aniketshukla1 (approved):
Reviewed the empty-key validation and normal delete path. Passed `S3DeleteObjectsIntegrationTest` under JDK 25.

pgermosen (approved):
Thanks for the fix — rejecting an empty `Delete` request is correct. The [DeleteObjects API reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) marks `Object` as `Required: Yes`, so a request with zero `<Object>`/`<Key>` entries genuinely fails AWS's schema validation with `MalformedXML`, matching this fix.

One minor observation, not really this PR's to fix: the API reference's own "Malformed XML in the request" example shows the full message as "The XML you provided was not well-formed or did not validate against our published schema," while this codebase's `MalformedXML` message stops after "well-formed." I checked, and that's an existing pattern already used in four other places in `S3Controller.java` (predating this PR), so this line is just following established convention rather than introducing anything new — flagging it as a heads-up in case it's ever worth a separate cleanup across all five call sites, not something to change here in isolation.

PR #2203, opened 2026-08-10
fix(apigatewayv2): encode binary Lambda request bodies

### LINKED ISSUE #2172: [BUG] The API Gateway HTTP v2 request payload base64 encoding doesn't work when using binary data.
## Service

The API Gateway HTTP 

## AWS API Action

Send a request to the API Gateway using the binary content-type like `application/octet-stream`.

## Expected behavior

When using API Gateway integration with Lambda ,the Lambda event payload format has an `isBase64Encoded` field. If the request's content type isn't a text type,the value will be set to true,and body will be base64 encoded.
```json
{
  "version": "2.0",
  "routeKey": "$default",
  "rawPath": "/dev/testApi",
  "rawQueryString": "",
  "cookies": null,
  "headers": {
    "accept": "*/*",
    "content-length": "11",
    "content-type": "application/octet-stream",
    "host": "[REDACTED]",
    "user-agent": "curl/8.2.1",
    "x-amzn-trace-id": "[REDACTED]",
    "x-forwarded-for": "[REDACTED]",
    "x-forwarded-port": "443",
    "x-forwarded-proto": "https"
  },
  "queryStringParameters": null,
  "pathParameters": null,
  "stageVariables": { "lambdaName": "[REDACTED]" },
  "body": "aGVsbG8gd29ybGQ=",
  "isBase64Encoded": true,
  "requestContext": {
    "routeKey": "$default",
    "accountId": "[REDACTED]",
    "stage": "dev",
    "apiId": "[REDACTED]",
    "domainName": "[REDACTED]",
    "domainPrefix": "[REDACTED]",
    "time": "07/Aug/2026:01:37:53 +0000",
    "timeEpoch": 1786066673815,
    "http": {
      "method": "POST",
      "path": "/dev/testApi",
      "protocol": "HTTP/1.1",
      "sourceIp": "[REDACTED]",
      "userAgent": "curl/8.2.1"
    },
    "authorizer": null,
    "requestId": "BtWF3jRyHUYEPqg="
  }
}
```

## Actual behavior

```json
{
  "version": "2.0",
  "routeKey": "$default",
  "rawPath": "/testApi",
  "rawQueryString": "",
  "cookies": null,
  "headers": {
    "host": "localhost",
    "accept": "*/*",
    "content-length": "11",
    "x-forwarded-proto": "http",
    "x-forwarded-for": "172.22.0.1",
    "user-agent": "curl/8.7.1",
    "x-real-ip": "172.22.0.1",
    "content-type": "application/octet-stream"
  },
  "queryStringParameters": null,
  "pathParameters": null,
  "stageVariables": null,
  "body": "hello world",
  "isBase64Encoded": false,
  "requestContext": {
    "routeKey": "$default",
    "accountId": "000000000000",
    "stage": "api",
    "apiId": "55b6276bc6",
    "domainName": "55b6276bc6.execute-api.us-east-1.amazonaws.com",
    "domainPrefix": "55b6276bc6",
    "time": "07/Aug/2026:15:56:17 +0000",
    "timeEpoch": 1786118177536,
    "http": {
      "method": "POST",
      "path": "/testApi",
      "protocol": "HTTP/1.1",
      "sourceIp": "127.0.0.1",
      "userAgent": "curl/8.7.1"
    },
    "authorizer": null,
    "requestId": "442a8e95-a0a8-44b2-b29e-c1a8b14e330e"
  }
}

```

## Reproduction
index.js
```js
  return {
    statusCode: 200,
    body: "result=" + JSON.stringify(event),
  };
};

````

```bash
export AWS_ACCESS_KEY_ID=test
zip -q function.zip index.js
aws lambda create-function --function-name repro-node --runtime nodejs20.x --handler index.handler --role arn:aws:iam::000000000000:role/floci-lambda --zip-file fileb://function.zip
API_ID=$(aws apigatewayv2 create-api --name repro-api --protocol-type HTTP --query ApiId --output text)
aws apigatewayv2 create-integration --api-id $API_ID --integration-type AWS_PROXY --integration-uri "arn:aws:lambda:us-east-1:000000000000:function:repro-node" --payload-format-version 2.0
INT_ID=$(aws apigatewayv2 get-integrations --api-id $API_ID --query "Items[?IntegrationType=='AWS_PROXY'].IntegrationId" --output text)
aws apigatewayv2 create-route --api-id $API_ID --route-key "POST /echo" --target "integrations/$INT_ID"
aws apigatewayv2 create-stage --api-id $API_ID --stage-name prod
curl -s -X POST "http://localhost:4566/execute-api/$API_ID/prod/echo" -H 'Content-Type: application/octet-stream' --data-binary $'\xff\xfe\xfd\xfa'
```

## Environment

- Floci version 1.5.33
- Docker

https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

The AWS docs dont' specify which types get encoded.I tested it in the real AWS environment.These are the types that are not base64-encoded(treated as binary data)
```
text/plain
text/plain; charset=utf-8
text/html
text/csv
text/xml
application/json
application/json; charset=utf-8
application/xml
application/javascript
application/graphql
```
and these are base64 encoded.
```
text/plain; charset=iso-8859-1
application/x-www-form-urlencoded
application/octet-stream
application/x-protobuf
application/zip
application/pdf
multipart/form-data; boundary=––tes
image/png
image/jpeg
image/gif
audio/mpeg
video/mp4
weird/unknown
<none>
```

Boulea7:
## Summary

- Encode binary request bodies as Base64 in API Gateway v2 Lambda proxy events and set `isBase64Encoded` to `true`.
- Keep supported text media types as UTF-8 text; treat unknown or missing `Content-Type` values as binary.
- Preserve empty request bodies as `body: null` with `isBase64Encoded: false`.

## Testing

- Full Maven Surefire suite: 8,844 tests, 0 failures, 0 errors, 9 skipped.
- Request body encoding matrix: 27 cases.
- API Gateway v2 binary payload E2E: 5 tests.

Fixes #2172


Boulea7:
Reviewed the automated feedback against the repository's AWS-compatibility rule and the evidence in #2172.

- Addressed the cleanup concern in `63a353e1`: teardown is now best effort, matching the neighboring API Gateway v2 integration tests, so it cannot mask an earlier failure with a strict status assertion.
- Kept the original media-type parameter behavior. The issue's real-AWS matrix verifies UTF-8 parameters only for `text/plain` and `application/json`; AWS's HTTP API documentation does not define a broader request classification rule. Expanding every text allowlist entry to accept UTF-8 parameters would therefore introduce behavior without AWS evidence. `fa214039` restores the measured matrix after reviewing that suggestion.

I re-reviewed the resulting diff and found no further code changes supported by the available AWS evidence.

hectorvent:
Thank you, this is an unusually well evidenced first PR. Deriving the test matrix from the real AWS capture in #2172, one case per measured row, is the right way to pin behaviour the AWS docs do not specify. The explicit UTF-8 charset fixes a latent platform default bug too.

On the media type parameter question, your reasoning is right and I am seconding it. #2172 measured `charset=utf-8` only on `text/plain` and `application/json`, and measured `text/plain; charset=iso-8859-1` as Base64, so the parameter demonstrably changes the answer. Widening the allowlist would be asserting behaviour no capture supports.

1. **(follow-up, separate PR)** Worth measuring when convenient: `text/html; charset=utf-8` is common in the wild and sits in the untested gap.

2. **(follow-up, separate PR)** The v1 REST path still sends every body through `new String(body)` with `isBase64Encoded` false. Different mechanism there (configured `binaryMediaTypes`), so its own change.

No blockers from my side.

hectorvent (approved):

PR #2234, opened 2026-08-11
fix(stepfunctions): publish execution status after its terminal fields

hectorvent:
## Summary

`DescribeExecution` returns the live `Execution` instance from the store, and the executor mutates it in place with the status set *before* the fields that describe it. A client polling for a terminal status can therefore observe `SUCCEEDED` with a null `output`, which real Step Functions never returns.

This is what reddened `main` in [run 31457665686](https://github.com/floci-io/floci/actions/runs/31457665686): the poll in `StepFunctionsPassParametersIntegrationTest` landed inside the window and Jackson rejected the null document with `IllegalArgumentException: argument "content" is null`. It is intermittent (one failure in the last fifteen `main` runs) because the window is two adjacent field writes wide. It is not caused by #2079, which merely happened to be the merge commit running at the time.

The status is now the publication point on every terminal path: the terminal fields are written first, `setStatus(...)` last.

| Location | Path |
|---|---|
| `AslExecutor:345` | `SUCCEEDED` (the one that broke CI) |
| `AslExecutor:332` | `FAILED`, runtime exception |
| `AslExecutor:353` | `FAILED`, outer catch |
| `AslExecutor:2571` | `failExecution` |
| `AslExecutor:208` | `TIMED_OUT` |
| `StepFunctionsService:258` | `ABORTED` in `stopExecution` |

The outer catch was worse than a race: it set only the status, so an execution failing there reported `FAILED` with a null `error` and `cause` **permanently**. It now records both.

Fixes the intermittent `main` failure in run 31457665686.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behaviour was user visible, not only test visible. Polling `DescribeExecution` until the status is terminal and then reading `output` is the documented pattern and what the SDK waiters do. Against Floci that poll could return `SUCCEEDED` with no `output`, and separately `FAILED` with no `error` or `cause`. Real Step Functions never returns a terminal execution without its terminal fields.

No request or response shape, field name or error code changed, so there is no wire contract change to verify against the model.

## Checklist

- [ ] `./mvnw test` passes locally — ran the affected suites rather than the full build: `-Dtest='StepFunctions*,Asl*'` gives **222 tests, 0 failures, 0 errors**. Leaving this unticked rather than overclaim; CI runs the full suite.
- [ ] New or updated integration test added — **deliberately none, see below**

### Why there is no new test

I wrote a concurrent polling integration test and then ran it against the *unfixed* code to check it actually caught the bug. It passed, so it was worthless as a regression guard and I discarded it rather than ship a green test that proves nothing.

The reason is structural: executions here start and complete inside the same millisecond while an HTTP round trip is tens of milliseconds, so a black box poll cannot land in a window that is a few instructions wide. There is no seam to observe from either, since `onUpdate` fires after every setter and so sees identical state before and after this change.

Existing coverage does guard the fields themselves, just not their ordering: `StepFunctionsJsonataIntegrationTest` asserts `error` and `cause` on failed executions in several places, so a regression that dropped those fields would go red.

## Follow up, not in this PR

The outer catch in `doExecute` still emits no `ExecutionFailed` history event, unlike the other two `FAILED` paths, so `GetExecutionHistory` is missing it. Fixing that needs `AtomicLong eventId` hoisted out of the `try` where it is declared (`AslExecutor:281`), which is beyond the scope of an ordering fix.

Reordering shrinks the window to a couple of instructions but is not a memory model guarantee, since the store hands out the live object. The stronger fix is to build the terminal state and write it through the store in a single `put`. Happy to do that instead if preferred.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -205,8 +205,8 @@ public class AslExecutor {
                     runUnderExecutionAccount(sm, () -> doExecute(sm, exec, history, onUpdate)));
             f.get(300, TimeUnit.SECONDS);
         } catch (java.util.concurrent.TimeoutException e) {
-            exec.setStatus("TIMED_OUT");
             exec.setStopDate(System.currentTimeMillis() / 1000.0);
+            exec.setStatus("TIMED_OUT");
             onUpdate.accept(exec, history);
         } catch (Exception e) {
             LOG.warnv("Sync execution wait failed for {0}: {1}", exec.getExecutionArn(), e.getMessage());
@@ -329,12 +329,12 @@ public class AslExecutor {
                     onUpdate.accept(exec, history);
                     return;
                 } catch (Exception e) {
-                    exec.setStatus("FAILED");
-                    exec.setStopDate(System.currentTimeMillis() / 1000.0);
                     String runtimeError = "States.Runtime";
                     String runtimeCause = e.getMessage() != null ? e.getMessage() : "Unknown error";
                     exec.setError(runtimeError);
                     exec.setCause(runtimeCause);
+                    exec.setStopDate(System.currentTimeMillis() / 1000.0);
+                    exec.setStatus("FAILED");
                     addEvent(history, eventId, "ExecutionFailed", null,
                             Map.of("error", runtimeError, "cause", runtimeCause));
                     onUpdate.accept(exec, history);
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Nice, careful fix — thanks for the thorough write-up. I checked the core premise against the [DescribeExecution reference](https://docs.aws.amazon.com/step-functions/latest/apireference/API_DescribeExecution.html) rather than taking it on faith: "`output`... This field is set only if the execution succeeds. If the execution fails, this field is null" — so a client observing `SUCCEEDED` with a null `output` really is a wire-contract violation, not just a test artifact. I also confirmed the architectural claim directly: `InMemoryStorage` is a bare `ConcurrentHashMap` with no copy-on-read, so `describeExecution` genuinely hands back the same live object `AslExecutor` mutates in place.

Traced all six reordered sites against your table and they're all correct, and the outer-catch fix (actually setting `error`/`cause` now, not just reordering) is a real bug fix bundled in, not just moved code.

Also appreciated the honesty about the test — discarding one that passed against the unfixed code rather than shipping a green regression guard that proves nothing is exactly right — and the memory-model caveat that reordering is a mitigation rather than a guarantee is a fair read.

PR #2240, opened 2026-08-11
fix(dynamodb): isolate item storage and locks between accounts

### LINKED ISSUE #2238: [BUG] DynamoDB item storage and locks are not isolated between accounts (same-named tables collide)
## Service

DynamoDB

## AWS API Action

PutItem, GetItem, Scan, Query, UpdateItem, DeleteItem, TransactWriteItems, CreateTable

## Expected behavior

Two different AWS accounts can each create a table with the same name in the same region and get fully independent item data — exactly like real DynamoDB, where `TableName` resolution is scoped by the caller's account + region, not by name alone.

## Actual behavior

`DynamoDbService` keeps table *metadata* (`tableStore`) properly account-scoped via `AccountAwareStorageBackend`, so two accounts can each create a table named e.g. `Users` without a `ResourceInUseException`. But the actual item data (`itemsByTable`), per-item locks (`itemLocks`), and `TransactWriteItems` `ClientRequestToken` idempotency cache (`txIdempotency`) are plain fields keyed only by `region::tableName` (or `region::token`), with no account component at all. Two accounts with a same-named table therefore share one underlying item map: a `PutItem` in account A's table is visible via `GetItem`/`Scan` in account B's identically-named table, and writes from either account can silently overwrite the other's data.

Two related gaps in the same code path:
- `loadPersistedItems()` (startup reload) only reloads the *default* account's persisted items — `AccountAwareStorageBackend.keys()` is request-scoped and startup has no request, so non-default accounts' items are silently dropped from the in-memory cache after a restart (though still present in the persisted file).
- The TTL sweeper (`deleteExpiredItems()`) uses `AccountAwareStorageBackend.scanAllAccountsAsMap()`, which strips the account prefix and collapses same-named tables from different accounts into one map entry — one account's expired items are silently never swept.

## Reproduction

```bash
# Two accounts, same table name, same region.
export AWS_ACCESS_KEY_ID=000000000001; export AWS_SECRET_ACCESS_KEY=x
  --table-name shared-table --attribute-definitions AttributeName=pk,AttributeType=S \
  --key-schema AttributeName=pk,KeyType=HASH --billing-mode PAY_PER_REQUEST

export AWS_ACCESS_KEY_ID=000000000002; export AWS_SECRET_ACCESS_KEY=x
  --table-name shared-table --attribute-definitions AttributeName=pk,AttributeType=S \
  --key-schema AttributeName=pk,KeyType=HASH --billing-mode PAY_PER_REQUEST

export AWS_ACCESS_KEY_ID=000000000001; export AWS_SECRET_ACCESS_KEY=x
  --table-name shared-table --item '{"pk":{"S":"item-from-account-1"},"secret":{"S":"account-1-data"}}'

export AWS_ACCESS_KEY_ID=000000000002; export AWS_SECRET_ACCESS_KEY=x
aws --endpoint-url http://localhost:4566 dynamodb get-item \
  --table-name shared-table --key '{"pk":{"S":"item-from-account-1"}}'
# Expected: no Item in the response.
# Actual: returns account 1's item, including "secret": "account-1-data".
```

## Environment

- Floci version / image tag: main @ `40e8aad5`
- Java SDK version (if applicable): n/a (reproduced via AWS CLI / REST)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): native (`mvn quarkus:dev`)

## Related

Same class of bug as #2239 (S3 object storage account isolation) — both stem from byte/item-level storage fields bypassing `AccountAwareStorageBackend`.

diopolgg:
## Summary

DynamoDbService kept item data (`itemsByTable`), per-item locks (`itemLocks`), and TransactWriteItems idempotency tokens (`txIdempotency`) in raw maps keyed only by `region::tableName` / `region::token` — with no account component. Unlike `tableStore`/`itemStore`, which get automatic account prefixing from `AccountAwareStorageBackend`, these fields bypassed it entirely. Two AWS accounts creating a same-named table in the same region therefore shared one underlying item map: `PutItem` in one account's table was visible via `GetItem`/`Scan` in another account's identically-named table.

Scopes every `itemsByTable`/`itemLocks` access by the resolved account ID via a new `scopedItemsKey()` helper. Also fixes two related gaps in the same code path, surfaced during review:

- `loadPersistedItems()` only reloaded the default account's items after a restart (`AccountAwareStorageBackend.keys()` is request-scoped, and startup has no request).
- The TTL sweeper's use of `scanAllAccountsAsMap()` collapsed same-named tables from different accounts into one map entry, silently skipping one account's expired items.

Both are fixed via a new `AccountAwareStorageBackend.scanAllAccountsRaw()`, which preserves the account segment instead of stripping it. A follow-up commit also scopes the (production-unreachable, but inconsistent) non-`AccountAwareStorageBackend` fallback branches in both of those methods, for consistency with every other accessor in the class.

Closes #2238

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

DynamoDB `TableName` resolution is scoped by the caller's account + region on real AWS, not by name alone — this brings Floci in line with that. No wire-format changes.

## Checklist

- [x] `./mvnw test` passes locally (full suite: 9010 passed, 0 failed)
- [x] New integration test added (`DynamoDbAccountIsolationIntegrationTest`, fails without this fix)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
+++ b/src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
@@ -121,6 +121,23 @@ public class AccountAwareStorageBackend<V> implements StorageBackend<String, V> 
         return result;
     }
 
+    /**
+     * Returns all entries across every account, keyed by the raw {@code accountId/logicalKey}
+     * storage key. Unlike {@link #scanAllAccountsAsMap()}, the account segment is preserved
+     * rather than stripped, so entries that share a logical key across different accounts
+     * (e.g. two accounts each owning a resource named "orders") remain distinguishable.
+     */
+    public Map<String, V> scanAllAccountsRaw() {
+        Map<String, V> result = new LinkedHashMap<>();
+        for (String rawKey : delegate.keys()) {
+            if (rawKey.indexOf('/') < 0) {
+                continue;
+            }
+            delegate.get(rawKey).ifPresent(v -> result.put(rawKey, v));
+        }
+        return result;
+    }
+
     public Optional<V> getForAccount(String accountId, String key) {
         return delegate.get(accountId + "/" + key);
     }
--- a/src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
+++ b/src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
@@ -126,14 +126,18 @@ public class AccountAwareStorageBackend<V> implements StorageBackend<String, V> 
      * storage key. Unlike {@link #scanAllAccountsAsMap()}, the account segment is preserved
      * rather than stripped, so entries that share a logical key across different accounts
      * (e.g. two accounts each owning a resource named "orders") remain distinguishable.
+     *
+     * <p>Entries with no account segment at all — persisted before multi-account support was
+     * added — are attributed to {@code defaultAccountId} rather than skipped, mirroring how
+     * {@link #get} treats them on a per-key lookup. This keeps a bulk startup reload (which has
+     * no per-key caller to trigger that migrate-on-read path) from silently losing pre-existing
+     * data on upgrade.
      */
     public Map<String, V> scanAllAccountsRaw() {
         Map<String, V> result = new LinkedHashMap<>();
         for (String rawKey : delegate.keys()) {
-            if (rawKey.indexOf('/') < 0) {
-                continue;
-            }
-            delegate.get(rawKey).ifPresent(v -> result.put(rawKey, v));
+            String effectiveKey = rawKey.indexOf('/') < 0 ? defaultAccountId + "/" + rawKey : rawKey;
+            delegate.get(rawKey).ifPresent(v -> result.put(effectiveKey, v));
         }
         return result;
     }
--- a/src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
+++ b/src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
@@ -124,19 +124,15 @@ public class AccountAwareStorageBackend<V> implements StorageBackend<String, V> 
 
     /**
      * Returns all entries across every account, keyed by the raw {@code accountId/logicalKey}
-     * storage key. Unlike {@link #scanAllAccountsAsMap()}, the account segment is preserved
-     * rather than stripped, so entries that share a logical key across different accounts
-     * (e.g. two accounts each owning a resource named "orders") remain distinguishable.
+     * storage key rather than the stripped logical key {@link #scanAllAccountsAsMap()} returns,
+     * so entries sharing a logical key across accounts stay distinguishable.
      *
-     * <p>Entries with no account segment at all — persisted before multi-account support was
-     * added — are attributed to {@code defaultAccountId} rather than skipped, mirroring how
-     * {@link #get} treats them on a per-key lookup. Unlike {@code get()}'s per-key migration,
-     * this bulk scan has no single caller-supplied key to migrate on demand, so it migrates (or
-     * discards) every bare legacy key it finds immediately: already-prefixed keys are loaded
-     * first so a legacy key can never win a collision against one that's since been written
-     * under its proper prefix, and any bare key superseded that way is deleted outright rather
-     * than left to linger — otherwise it would still be sitting in storage to collide again,
-     * non-deterministically, on a later restart.
+     * <p>A key with no account segment (pre-multi-account data) is attributed to
+     * {@code defaultAccountId} rather than skipped, and migrated in place — unlike {@link #get}'s
+     * per-key migration, this bulk scan has no caller-supplied key to migrate later, so it acts
+     * immediately: already-prefixed keys load first so a legacy key can never win a collision
+     * against one already written under its proper prefix, and a superseded legacy key is
+     * deleted rather than left to collide again on a later restart.
      */
     public Map<String, V> scanAllAccountsRaw() {
         Map<String, V> result = new LinkedHashMap<>();
--- a/src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
+++ b/src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
@@ -4,6 +4,7 @@ import io.github.hectorvent.floci.core.common.RequestContext;
 import jakarta.enterprise.context.ContextNotActiveException;
 import jakarta.enterprise.inject.Instance;
 
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


diopolgg:
Addressed in ec390500: \`AccountAwareStorageBackend.scanAllAccountsRaw()\` was silently dropping any key with no account segment (data persisted before multi-account support existed). It now attributes such keys to \`defaultAccountId\` instead — consistent with how \`get()\` already treats them on a per-key lookup — so \`loadPersistedItems()\`/the TTL sweeper no longer lose pre-existing data on upgrade. Added \`AccountAwareStorageBackendTest\` covering this directly, plus reran the full package (406/406 passing).

src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
```diff
@@ -121,6 +121,27 @@ public Map<String, V> scanAllAccountsAsMap() {
...
+     * rather than stripped, so entries that share a logical key across different accounts
+     * (e.g. two accounts each owning a resource named "orders") remain distinguishable.
+     *
+     * <p>Entries with no account segment at all — persisted before multi-account support was
+     * added — are attributed to {@code defaultAccountId} rather than skipped, mirroring how
+     * {@link #get} treats them on a per-key lookup. This keeps a bulk startup reload (which has
+     * no per-key caller to trigger that migrate-on-read path) from silently losing pre-existing
+     * data on upgrade.
+     */
+    public Map<String, V> scanAllAccountsRaw() {
```
pgermosen, line 137: This is read-only against `delegate` — it never writes the migrated key back or deletes the original bare one, unlike `get()`'s migrate-on-read a few lines up. For a table that gets written to after this reload and then the process restarts again, `delegate.keys()` would return both the stale bare key and the fresh prefixed key, and whichever one this loop's iteration visits last wins in the result map — a real chance of silently reviving stale pre-migration data on a second restart. Might be worth migrating the bare key here the same way `get()` does, so it doesn't linger.

pgermosen (commented):
Thanks for tracking this down — a real cross-account leak is exactly the kind of thing worth this level of care, and the three-commit progression shows it. I traced every `itemsByTable`/`itemLocks` access site in `DynamoDbService.java` and confirmed each one is correctly scoped now, including the TTL sweeper's handling of running outside request scope. Also confirmed AWS table names can't contain `/`, so the legacy-key detection in `scanAllAccountsRaw()` is unambiguous.

One small gap I noticed while going through `scanAllAccountsRaw()`: it's read-only against the underlying store — it attributes a legacy bare key to `defaultAccountId` in the map it returns, but doesn't migrate that key in persisted storage the way `get()`'s existing migrate-on-read does a few lines up. Concretely: after a restart, a legacy item gets loaded and attributed to `defaultAccountId` in memory; further writes to that table then persist under the properly-prefixed key, but the original bare key is left sitting in storage untouched. On a *second* restart, both the stale bare key and the fresh prefixed key would map to the same effective key, and whichever one the map iteration happens to visit last would silently win — a real chance of reviving stale pre-migration data. It's a narrow case (only affects genuine upgrades from pre-multi-account installs going through restart → write → restart again), but would be worth closing by having this method migrate the bare key the same way `get()` does.

src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
```diff
@@ -121,6 +121,27 @@ public Map<String, V> scanAllAccountsAsMap() {
...
+        Map<String, V> result = new LinkedHashMap<>();
+        for (String rawKey : delegate.keys()) {
+            String effectiveKey = rawKey.indexOf('/') < 0 ? defaultAccountId + "/" + rawKey : rawKey;
+            delegate.get(rawKey).ifPresent(v -> result.put(effectiveKey, v));
+        }
+        return result;
+    }
```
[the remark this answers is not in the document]
diopolgg: Addressed in dbebda59: `scanAllAccountsRaw()` now does a two-pass scan — it loads all prefixed (account-scoped) keys first, then only migrates a bare legacy key to `defaultAccountId` if that logical key isn't already covered by a prefixed entry. Prefixed data always wins over stale legacy data, regardless of backend iteration order. Added collision-precedence coverage in `AccountAwareStorageBackendTest`, full package passing.

src/main/java/io/github/hectorvent/floci/core/storage/AccountAwareStorageBackend.java
diopolgg: Good catch — fixed in dbebda59. `scanAllAccountsRaw()` now migrates the bare legacy key back into `delegate` under its prefixed form (and deletes the original bare key) whenever it isn't already superseded by an existing prefixed entry, the same way `get()`'s migrate-on-read does. So after this scan runs once, the bare key no longer lingers and can't collide on a second restart. Added a regression test for exactly this migrate-then-restart scenario in `AccountAwareStorageBackendTest`.

pgermosen (approved):
## Summary

Both concerns from the first pass are resolved, and the fix went further than what was asked.

**My comment on `AccountAwareStorageBackend.java:137` is addressed properly.** `scanAllAccountsRaw()` is now a two-pass scan: prefixed keys load first, then each bare legacy key is either deleted outright (if its prefixed form already exists) or migrated in place — `put` under the prefix, `delete` the bare one. That is exactly the migrate-on-read behavior `get():51–64` already had, which is what I asked for, and it closes greptile's collision-precedence P1 as a consequence rather than as a separate patch: prefixed data now wins deterministically, independent of `delegate.keys()` iteration order. The three new `AccountAwareStorageBackendTest` cases pin all three branches, including the supersede-and-delete one.

**The keying invariant holds** — this is the thing that could have broken silently, so I checked it rather than assuming:

`scopedItemsKey()` builds `regionResolver.getAccountId() + "/" + storageKey`; `AccountAwareStorageBackend.prefixed()` builds `prefix() + "/" + key`. `RegionResolver.getAccountId():111` and `prefix():189` are the *same* logic — read `RequestContext`, swallow `ContextNotActiveException`, fall back to a default — and both defaults come from `config.defaultAccountId()` (`RegionResolver:29`, `StorageFactory:99`). The in-memory `itemsByTable` key and the physical `itemStore` key therefore cannot disagree about the account.

Also checked:

- **No unscoped access survives.** `itemsByTable` has 13 call sites, all through `scopedItemsKey(...)` except `:146` and `:1366`, which consume an already-scoped raw key from `scanAllAccountsRaw()`. Both `itemLocks` sites are scoped. `txIdempotency` is keyed `accountId::region::token`.
- **`exportTable:2681` is correct as written** — the snapshot is taken on the request thread *before* `Thread.ofVirtual().start(...)`, so the virtual thread never needs a request scope it doesn't have.
- **The final commit is comments-only.** `dbebda59...527591cc` touches nothing but javadoc and comments, so nothing behavioral has changed since the commit CI validated.

**The AWS premise is documented, not inferred:**

| operation | model documentation |
|---|---|
| `CreateTable` | *"In an Amazon Web Services account, table names must be unique within each Region. That is, you can have two tables with same name if you create the tables in different Regions."* |
| `ListTables` | *"Returns an array of table names associated with the current account and endpoint."* |
| `TransactGetItems` | *"A call to TransactGetItems cannot retrieve items from tables in more than one Amazon Web Services account or Region."* |

The namespace really is `(account, region)` — precisely the key this moves to.

## Non-blocking

**Account-scoping the idempotency token is right, but AWS doesn't document it.** `TransactWriteItems`' model text covers only the 10-minute window and `IdempotentParameterMismatch`; it says nothing about whether the token namespace is per-account. Including the account in the cache key is clearly the correct direction — without it, account B replaying account A's token gets a silent no-op — it's just worth knowing it's a reasoned choice rather than documented parity, in case someone later treats it as an AWS invariant.

**`DynamoDbStreamService` has the same bug in a different field.** `streams` and `records` are plain `ConcurrentHashMap`s keyed by `streamKey(region, tableName)` = `region + "::" + tableName` (`DynamoDbStreamService.java:294`), with no account component. Two accounts each enabling a stream on a same-named table share one `StreamDescription` and one record deque, so one account's `GetRecords` can read the other's change records. Same class as #2238 and out of scope here — worth a sibling issue rather than growing this PR.

Thanks for the thorough follow-up on the migration path.

pgermosen (approved):
Looks good

PR #2244, opened 2026-08-11
fix(cloudformation): honor stack status filters

### LINKED ISSUE #2242: [BUG] cloudformation: ListStacks ignores StackStatusFilter
## Summary

`cloudformation list-stacks --stack-status-filter <statuses>` returns every live stack in the account regardless of the filter. Real AWS returns only stacks whose status is in the supplied list.

## Reproduction

1. In an account with stacks in mixed states (e.g. several `CREATE_COMPLETE`/`UPDATE_COMPLETE` and one `ROLLBACK_COMPLETE`):
2. `aws cloudformation list-stacks --stack-status-filter ROLLBACK_COMPLETE DELETE_FAILED`
3. Expected: only the `ROLLBACK_COMPLETE` stack. Actual: all stacks, all statuses.

## Context

Noticed while sweeping org accounts for dead `ROLLBACK_COMPLETE` stacks (failed initial creates that the CDK toolkit delete-and-recreates on the next deploy). The sweep had to client-side filter because the emulator ignores the `StackStatusFilter` parameter.

aniketshukla1:
## Summary

Closes #2242

- Honor repeated `StackStatusFilter.member.N` query parameters in `ListStacks`.
- Preserve the existing unfiltered response when no status filter is supplied.
- Add end-to-end coverage for filtering a mixed `CREATE_COMPLETE` and `ROLLBACK_COMPLETE` stack set.

## Validation

- `./mvnw test -Dtest=CloudFormationListStacksIntegrationTest,CloudFormationDeleteIdempotencyIntegrationTest`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationQueryHandler.java
@@ -455,7 +455,10 @@ public class CloudFormationQueryHandler {
     // ── ListStacks ────────────────────────────────────────────────────────────
 
     private Response listStacks(MultivaluedMap<String, String> params, String region) {
-        List<Stack> stacks = cfnService.listStacks(region);
+        List<String> statusFilter = extractList(params, "StackStatusFilter.member.");
+        List<Stack> stacks = cfnService.listStacks(region).stream()
+                .filter(stack -> statusFilter.isEmpty() || statusFilter.contains(stack.getStatus()))
+                .toList();
         XmlBuilder xml = new XmlBuilder()
                 .start("ListStacksResponse", CF_NS)
                 .start("ListStacksResult")
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationListStacksIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationListStacksIntegrationTest.java
@@ -1,0 +1,77 @@
+package io.github.hectorvent.floci.services.cloudformation;
+
+import io.quarkus.test.junit.QuarkusTest;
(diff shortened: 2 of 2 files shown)


aniketshukla1:
@hectorvent @pgermosen can we merge this ?

pgermosen (approved):
Small, correct, and the wire format is right. Reviewed at `e4a614fd`.

## Verified against the API reference, not just the model

The [ListStacks](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ListStacks.html) page names the parameter literally:

> **StackStatusFilter.member.N** — "Stack status to use as a filter. Specify one or more stack status codes to list only stacks with the specified status codes."

So the serialization this parses is the documented one rather than an inference — the shape carries no `flattened` flag, and `extractList(params, "StackStatusFilter.member.")` is the same helper already used for `Capabilities.member.`, `Accounts.member.` and `Regions.member.` in this handler. Consistent with the surrounding code, which is what I look for first on a query-protocol change.

The unfiltered path matches the documented behavior too:

> "If no `StackStatusFilter` is specified, summary information for all stacks is returned."

`statusFilter.isEmpty() ||` short-circuits before the `contains` check, so no-filter requests keep returning everything, unchanged.

Both status codes used in the test are in the documented `Valid Values` list for the parameter.

## The test earns its place

It is a real regression test rather than a restatement of the implementation: `not(containsString(completeStack))` fails outright if the filter is ignored, which is exactly the bug in #2242. Provoking `ROLLBACK_COMPLETE` with a secret carrying both `SecretString` and `GenerateSecretString` is a neat deterministic fixture — no sleeps, no ordering assumptions — and the `nanoTime` suffix keeps it isolated from anything else creating stacks.

Worth noting this is the first `ListStacks` coverage in the repo; there was none in any CloudFormation test before this.

Approving.

PR #2245, opened 2026-08-11
fix(ec2): start sshd even when run-instances omits a key pair

### LINKED ISSUE #2184: [BUG] Default ec2 image docker containers don't have ssh server running
## Service

EC2

## AWS API Action

RunInstances

## Expected behavior

The created docker container runs an ssh daemon and can be connected to

## Actual behavior

The created docker container does not contain any active ssh daemon

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 ec2 run-instances --image-id ami-alpine # also reproducible with al2023 and ubuntu images
ssh root@localhost -p 2200 # connection refused
```

## Environment

- Docker: floci/floci:latest (1231b4a6ce21)

Mide69:
Fixes #2184.

`run-instances` without `--key-name` left every container with no SSH daemon running at all — connection refused, regardless of image (alpine, Amazon Linux 2023, Ubuntu all affected per the report).

Root cause: `startSshd()` only ran inside the same `if (publicKey != null)` block as `injectSshKey()`, so no key pair meant no daemon. Real AWS AMIs start sshd as part of normal boot regardless of whether a key pair is attached — it's always running, it just has nothing to authenticate an unattached client with. Splitting the two: `injectSshKey` stays conditional on a key being provided, `startSshd` now always runs.

New test launches an instance with no public key and confirms sshd still starts (and that nothing is written to `authorized_keys`, since there's nothing to inject). Verified the test fails with the fix reverted and passes with it restored. Full `ec2` package regression is green (268 tests).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
@@ -459,18 +459,40 @@ public class Ec2ContainerManager {
 
     private void startSshd(String containerId, String instanceId) {
         try {
-            // Install openssh-server if absent
-            execInContainer(containerId, new String[]{"sh", "-c",
+            // Install openssh-server if absent. The trailing "command -v sshd" makes the script's
+            // own exit code the source of truth for whether sshd is actually available afterward -
+            // without it, a shell if/elif chain with no matching branch (no dnf/apt-get/apk found) or
+            // whose install command itself failed (e.g. no network yet, apt lock held) still exits 0
+            // by bash convention, so every later step here would silently no-op against a daemon that
+            // was never installed while still logging success.
+            ContainerExecResult install = execInContainerForResult(containerId, new String[]{"sh", "-c",
                     "if ! command -v sshd >/dev/null 2>&1; then" +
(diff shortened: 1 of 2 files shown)


duytanisme:
@Mide69 In my case, `sshd` is not started even when an SSH key is provided. I have to start `sshd` manually, either by execing into the container or by running a custom script in user-data.

Mide69:
@duytanisme Thanks for the feedback, I will look into it

pgermosen (approved):
Nice, careful fix. Splitting `startSshd` out from the key-injection conditional correctly matches how real EC2 AMIs behave -- sshd runs at boot regardless of key-pair association.

Also appreciated the second commit catching a real latent bug along the way: the original exec call discarded exit codes entirely, so a failed or no-op install script still exited 0 by bash convention and logged success regardless. Making the install probe's own exit code authoritative, and checking exit codes at each step, closes that properly rather than papering over it.

Traced `execInContainerForResult`/`ContainerExecResult` and confirmed both were already used elsewhere (the metadata proxy) -- this reuses existing infrastructure rather than adding new machinery. Test coverage is solid too: one test confirms sshd starts with no key pair and nothing gets written to `authorized_keys`, and the other mocks the full exec lifecycle to prove a failed install probe short-circuits `ssh-keygen`/`sshd` rather than proceeding against a daemon that was never installed.

Approving.

Issue #2226, opened 2026-08-10
[BUG] cloudwatch: getLogEvents parseTokenIndex silently restarts pagination on malformed token (CloudWatchLogsService.java:411)


sam-fakhreddine:
**What's wrong**: `CloudWatchLogsService.parseTokenIndex` (used by `getLogEvents`, around line 411) returns `0` on `NumberFormatException` instead of throwing. A garbage, truncated, or otherwise unparseable `nextToken` silently restarts the caller at page 0 rather than erroring.

**Why it matters**: This is a silent-fallback anti-pattern. Any client paginating through `GetLogEvents` that hits a corrupted or mishandled token gets no error signal — it just silently loops back to the start, which is an infinite-loop hazard for automated paginating clients and masks real bugs (e.g. a client-side token-truncation bug would never surface, since floci just quietly "handles" the bad token instead of rejecting it).

**In-tree precedent for the fix**: `filterLogEvents`'s new `parseFilterToken` (added for #2209, `CloudWatchLogsService.java`) handles the equivalent case correctly — on an unparseable or negative offset it throws:
```java
throw new AwsException("InvalidParameterException",
        "The specified nextToken is invalid.", 400);
```
`parseTokenIndex` should follow the same shape: throw `AwsException("InvalidParameterException", "The specified nextToken is invalid.", 400)` instead of returning `0`.

**Suggested fix**: Mirror `parseFilterToken`'s validation in `parseTokenIndex` — reject non-numeric or negative values with a 400 `InvalidParameterException` instead of falling back to offset 0.

Filed per SEE-SOMETHING while working #2209/#2210 (branch `feature/cwl-filter-log-events`) — kept out of that diff to stay minimal-diff scoped. (No `see-something` label exists in this repo yet; tagged with `bug,cloudwatch` instead.)

pgermosen:
Fixed by #2249 (merged as `12cf6e2f`), thanks to @Boulea7. Closing manually — the PR said `Refs` rather than `Closes`, so it did not auto-close.

`parseTokenIndex` now throws `InvalidParameterException` / 400 instead of returning `0`, and the fix went a little wider than the report asked for:

- non-numeric and overflowing offsets are rejected, not just unparseable ones
- negative offsets are rejected
- a token that matches neither the `f/` nor the `b/` prefix is rejected outright, including the empty string
- null tokens and valid forward/backward pagination are untouched

## One correction for anyone finding this later

[...]

Issue #2209, opened 2026-08-10
[BUG] cloudwatch: FilterLogEvents nextToken is a random UUID and truncation drops the newest matches (CloudWatchLogsService.java:421-453)


sam-fakhreddine:
## What's wrong

`FilterLogEvents` returns a `nextToken` that no client can ever use, and truncation discards the newest matches rather than the oldest.

`CloudWatchLogsService.filterLogEvents` (`src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java:421-453`):

```java
List<LogEvent> result = all.stream()
        .filter(e -> (startTime == null || e.getTimestamp() >= startTime)
                && (endTime == null || e.getTimestamp() <= endTime))
        .filter(e -> filterPattern == null || filterPattern.isBlank()
                || e.getMessage().contains(filterPattern))
        .limit(maxEvents)
        .toList();

String nextToken = result.size() >= maxEvents ? UUID.randomUUID().toString() : null;
```

Two distinct defects:

1. **The token is a random UUID carrying no cursor information**, and the method signature has no `nextToken` parameter to accept one back. `CloudWatchLogsHandler.handleFilterLogEvents` (`CloudWatchLogsHandler.java:188-206`) likewise never reads a token from the request. So the response advertises more results and provides no way to reach them.
2. **`.limit()` clips the wrong end.** `all` is sorted ascending by `EVENT_ORDER` (`:429`) and the time/pattern filters run before `.limit()`, so an over-cap query returns the **oldest** N matches and silently drops the newest. `maxEvents` defaults to `max-events-per-query: 10000` (`src/main/resources/application.yml:295`) when the caller passes no limit.

## Reproduction

```
aws --endpoint-url http://localhost:4566 --region us-east-1 logs filter-log-events \
  --log-group-name /aws/codebuild/AWSAccelerator-InstallerProject --limit 3
```

Returns exactly 3 events plus `"nextToken": "eb3545b6-..."`. Passing that token back is not possible — the parameter is not read.

## Why it matters

The canonical use of `FilterLogEvents` is "find the error that just happened, across a group". With both defects, a group with more matches than the cap returns the oldest slice and reports no way forward, so the event the user is hunting is precisely the one most likely missing. Any client that trusts `nextToken` will either loop forever or conclude it has seen everything.

[...]

PR #2249, opened 2026-08-11
fix(cloudwatch): reject invalid log event tokens

Boulea7:
## Summary

- Reject malformed CloudWatch Logs `GetLogEvents` pagination tokens instead of silently restarting pagination.
- Return the existing AWS-compatible `InvalidParameterException` response with HTTP 400 for malformed, negative, overflow, empty, unknown-prefix, and separator-less tokens.
- Preserve null-token and valid forward/backward pagination behavior.

Refs #2226

## Validation

- `CloudWatchLogsServiceTest`: 39 tests passed.
- `CloudWatchLogsServiceTest,CloudWatchLogsHandlerTest`: 54 tests passed.
- Full test suite: 7,052 tests passed; 0 failures, 0 errors, 9 skipped.
- `./mvnw -B -DskipTests package`: passed.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java
@@ -366,6 +366,8 @@ public class CloudWatchLogsService {
             int end = parseTokenIndex(nextToken, 2);
             pageEnd = Math.min(end, total);
             pageStart = Math.max(pageEnd - maxEvents, 0);
+        } else if (nextToken != null) {
+            throw invalidNextToken();
         } else if (!startFromHead) {
             pageEnd = total;
             pageStart = Math.max(total - maxEvents, 0);
@@ -379,11 +381,20 @@ public class CloudWatchLogsService {
     }
 
     private int parseTokenIndex(String token, int prefixLen) {
+        int index;
         try {
-            return Integer.parseInt(token.substring(prefixLen));
+            index = Integer.parseInt(token.substring(prefixLen));
         } catch (NumberFormatException e) {
-            return 0;
+            throw invalidNextToken();
         }
+        if (index < 0) {
+            throw invalidNextToken();
+        }
+        return index;
+    }
+
+    private AwsException invalidNextToken() {
+        return new AwsException("InvalidParameterException", "The specified nextToken is invalid.", 400);
     }
 
     public record FilteredLogEventsResult(List<LogEvent> events, String nextToken) {}
--- a/src/test/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsServiceTest.java
@@ -374,6 +374,64 @@ class CloudWatchLogsServiceTest {
     }
 
     @Test
+    void getLogEventsRejectsMalformedNextToken() {
+        service.createLogGroup("/app/logs", null, null, REGION);
+        service.createLogStream("/app/logs", "stream-1", REGION);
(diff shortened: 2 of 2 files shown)


pgermosen (approved):
Approving. Welcome, and thanks for a well-scoped first contribution — the test coverage in particular is stronger than the change strictly needed, which is the right instinct.

## Verified locally

On current `main` (`2d003b52`): **130 cloudwatch tests green**.

I reverted each guard in a throwaway tree to confirm the tests actually pin the behaviour rather than merely passing alongside it:

| Reverted | Result |
|---|---|
| the `return 0` silent fallback | `getLogEventsRejectsMalformedNextToken` + `...RejectsOverflowNextToken` fail |
| the `index < 0` guard | `getLogEventsRejectsNegativeNextToken` fails |
| the unknown-prefix branch | `getLogEventsRejectsUnrecognizedNextTokens` fails |

Each of the three guards is independently pinned, so none can regress quietly.

## Checked against the AWS documentation

Since this changes an error contract, I went to the [GetLogEvents reference](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_GetLogEvents.html) rather than take the shape on trust. Three things confirm your choices:

- **The error contract is exact.** The Errors section documents *InvalidParameterException — A parameter is specified incorrectly. HTTP Status Code: 400*. Both the code and the 400 match.
- **Rejecting the empty token is documented, not a judgment call.** `nextToken` carries *"Length Constraints: Minimum length of 1."* Your `getLogEventsRejectsUnrecognizedNextTokens` case covering `""` is enforcing a real constraint. That one mattered, because `CloudWatchLogsHandler` reads the field with `request.has("nextToken") ? asText(null) : null`, so a client sending `"nextToken": ""` reaches the service as an empty string and used to be silently tolerated.
- **The `f/` and `b/` prefixes are the real wire format.** AWS's own sample response returns `"nextForwardToken": "f/31132629323784151764587387538205132201699397759403884544"`. Validating on those prefixes mirrors AWS rather than inventing a Floci convention.

I should be straight about the limit of that: AWS does not explicitly document *what* a malformed token returns. That step is an inference. It is a well-supported one — a token AWS never issued is by definition a parameter specified incorrectly, the documented min-length shows the parameter is validated at all, and tokens **expire after 24 hours**, which has to produce something other than a silent restart. But it is an inference, and worth having on the record as one.

It does not weaken the case, because the old behaviour was wrong on its own terms regardless of which error AWS picks: silently resuming at offset 0 is an infinite-loop hazard for any paginating client, which is exactly what #2226 reported.

## On the missing precedent

Worth flagging for anyone reading later — #2226 tells you to mirror `parseFilterToken` "added for #2209". Neither exists: there is no `parseFilterToken` in `main`, and #2209 is not a PR. That work stayed on the reporter's branch and never landed, so you had no template to copy and derived the shape yourself.

You landed on the right one anyway. `RumService.decodeOffset` is the genuine in-tree precedent, and it rejects unparseable, wrong-prefix and out-of-range tokens in exactly this manner. Reaching the same design independently is a good sign.

## Two notes, neither blocking

**Real AWS tokens are far larger than `int`.** Those sample tokens are 56-digit numbers, so `f/2147483648` in your overflow test is well inside AWS's ordinary range. For Floci this is harmless — it issues its own small-integer tokens and can never generate one it would reject — but a token copied from a real AWS response now gets a 400 rather than being tolerated. That is defensible for an emulator, and I mention it only so the overflow test is not mistaken for an AWS-fidelity claim.

**Out-of-bounds offsets are still clamped.** `f/999999` against a three-event stream returns an empty page rather than erroring, where `RumService` would reject it. That is pre-existing, arguably correct given AWS returns empty pages at the end of a stream, and squarely outside what you set out to fix. No action wanted.

## Housekeeping

No collision risk: #2259 merged on 2026-08-13 and is already in your base, and #1813 also touches `CloudWatchLogsService` but in the log-group creation area — I ran `merge-tree` across both heads and there is no conflict either way.

Nice catch on the missing trailing newline, too.

PR #2250, opened 2026-08-11
fix(core): quarantine unreadable hybrid storage

### LINKED ISSUE #2232: [BUG] storage: HybridStorage.load() silently starts with an empty store on deserialization failure (PersistentStorage quarantines)
**What's wrong**: `HybridStorage.load()` (`src/main/java/io/github/hectorvent/floci/core/storage/HybridStorage.java:99-107`) catches `IOException`, logs at error, and continues with an empty `store`. Jackson deserialization failures are `IOException` subclasses, so whenever a persisted JSON file no longer matches its model class, that store silently boots empty and the emulator comes up looking healthy.

**Why this matters**: `PersistentStorage.load()` (`src/main/java/io/github/hectorvent/floci/core/storage/PersistentStorage.java:83-100`) handles the identical case very differently — it quarantines the unreadable file to `<name>.corrupt` and logs loudly, with a comment explaining that starting empty "silently drops all persisted state for this store, which can leave other services (e.g. CloudFormation) referencing resources that now appear missing (see issue #1634)". That reasoning applies verbatim to hybrid mode. Hybrid never received the #1634 treatment.

This is not hypothetical for hybrid users specifically: hybrid is the mode anyone enabling persistence across restarts will use (it is `EmulatorConfig.java:192`'s `@WithDefault`), and model classes change frequently. The failure mode is a container that restarts cleanly with half its state missing and no quarantine artifact to show what was lost — strictly worse to diagnose than the `PersistentStorage` behaviour it sits beside.

**Repro sketch**: run with `FLOCI_STORAGE_MODE=hybrid`, create resources so a `*.json` store is flushed under the persistent path, then change/rename a field on the corresponding model class (or hand-edit the JSON to an incompatible shape) and restart. The store loads as empty; the only signal is one error line.

**Severity**: 4 — silent data loss with a healthy-looking process, and an inconsistency with the sibling backend that already solved it.

**Suggested fix**: extract the quarantine path from `PersistentStorage.load()` and share it between both backends, so hybrid also moves the unreadable file aside and logs that the store's state was lost. A shared helper is preferable to duplicating the logic, since `WalStorage` snapshot loading likely wants the same treatment.

Found while auditing floci's persistence model to design a shareable pre-setup LZA container checkpoint; the silent-decay behaviour is what makes snapshot artifacts unsafe to distribute. Related but distinct: #2225 (memory-mode default discards all state), #2228 (pre-fix persisted LogEvents backfill).

Boulea7:
## Summary

Closes #2232

- Quarantines unreadable HybridStorage snapshots to `<name>.corrupt`.
- Uses the existing `HybridStorage.this` intrinsic monitor so `load()` (including quarantine) and `persistToDisk()` are serialized; no new lock or scheduler/flush protocol.
- Adds a regression test for invalid JSON and an existing quarantine target, asserting `REPLACE_EXISTING` writes the unreadable source contents.

Maintainer edits are welcome.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## RED to GREEN

Before this change, `HybridStorageTest#loadQuarantinesUnreadableFile` failed because the original invalid JSON file remained after `load()`.

## Local validation

- `HybridStorageTest#loadQuarantinesUnreadableFile`: 1 test passed
- Storage suite: 49 tests passed
- `./mvnw test`: 9,075 tests passed; 9 skipped
- `./mvnw -DskipTests package`: passed

## Checklist


pgermosen (commented):
@Boulea7 — thank you for this, and welcome. This is a nicely scoped first change: it fixes a real gap, matches what the codebase already does elsewhere, and comes with a test that fails on the old behaviour. Very little to say against it.

Things I checked rather than assumed:

**It matches `PersistentStorage` exactly.** Your `quarantineUnreadableFile` is byte-identical to the one already there — same `.corrupt` suffix, same `REPLACE_EXISTING`, same log wording. That is precisely what #2232 asked for, and porting the established behaviour rather than inventing a second convention is the right instinct.

**The locking reasoning holds, and it matters more than it looks.** `persistToDisk()` was already `synchronized`, so making `load()` `synchronized` really does serialize them on the same monitor. It is also genuinely needed here and not in `PersistentStorage`, because `HybridStorage`'s constructor starts a `scheduleAtFixedRate(this::flushIfDirty, …)`. Without the lock a scheduled flush could fire mid-`load()`, write the in-memory store over the file being read, and then the quarantine `move` would file away the *newly written* data instead of the corrupt original. Reusing the existing intrinsic monitor instead of introducing a new lock was the right call.

**Your RED-to-GREEN claim is accurate.** I reverted the call to the old log-only line and re-ran: `loadQuarantinesUnreadableFile` fails with `expected: <false> but was: <true>` — the assertion that the original file is gone, which is exactly the bug. Locally 24 storage tests green, and all 15 CI checks pass. I also liked that the test pre-creates an existing `.corrupt` file and asserts the new contents replaced it; that pins the `REPLACE_EXISTING` semantics rather than leaving them incidental.

## One thing before I approve

`quarantineUnreadableFile` now exists verbatim in two places — `HybridStorage` and `PersistentStorage`. Fourteen identical lines including the log strings.

The risk is not the duplication itself, it is drift: someone tightening the message, adding a timestamp to the quarantine filename, or fixing an edge case will very likely touch one copy and not the other, and the two classes will quietly stop agreeing about how corrupt state is handled. That is the same class of inconsistency this PR is fixing, so it would be a shame to reintroduce it in a different shape.

Both classes implement `StorageBackend`, so a small package-private static helper — something like a `StorageQuarantine.quarantine(filePath, cause, LOG)` in the same package — would let both call one implementation. Happy for it to live wherever you think fits best; I do not want to over-prescribe the shape.

Would you mind pulling it out? Once that is in I am ready to approve — everything else here already looks good to me, and this is genuinely a clean contribution.

pgermosen (approved):
Pulled `quarantineUnreadableFile` out into a shared `StorageQuarantine.quarantine(filePath, cause, log)` that both `HybridStorage` and `PersistentStorage` now call, since you said maintainer edits were welcome. Also merged main in to clear the conflict in `HybridStorageTest`.

That was the one thing I was waiting on. Everything else still holds from my first pass: the locking reasoning is sound, the RED-to-GREEN claim checked out, and the test coverage is good. Thanks for the clean fix and for being patient with the back and forth.

Approving.

PR #2255, opened 2026-08-12
fix(cloudformation): honor resource-level conditions

### LINKED ISSUE #2168: [BUG] CloudFormation: resource-level Condition is ignored, resources with a false condition are still created
## Service

CloudFormation

## AWS API Action

CreateStack

## Expected behavior

A resource whose `Condition` evaluates to false is **not** created. The condition itself is evaluated correctly — `Fn::If` referencing the same condition picks the right branch — so this is only about the resource-level `Condition` key being ignored, not about condition evaluation in general.

In real AWS, the template below creates exactly one queue, named `false-branch`:

- `SkippedQueue` has `"Condition": "IsFalse"`, which is `{"Fn::Equals": ["a", "b"]}` → false, so the resource is skipped entirely and does not appear in `DescribeStackResources`.
- `PickedQueue` has no condition and is created, with `Fn::If` resolving to the false branch.

## Actual behavior

Both queues are created. `SkippedQueue` is provisioned and reported as `CREATE_COMPLETE` even though its condition is false.

Note that `Fn::If` **does** resolve correctly against the same condition (`PickedQueue` is named `false-branch`, not `true-branch`), so the condition evaluation engine works — the result just doesn't appear to be used when deciding whether to provision a resource.

```
$ aws sqs list-queues
{
    "QueueUrls": [
        "http://localhost:4566/000000000000/should-not-exist",
        "http://localhost:4566/000000000000/false-branch"
    ]
}

$ aws cloudformation describe-stack-resources --stack-name cond-repro \
      --query 'StackResources[].[LogicalResourceId,ResourceStatus]' --output text
SkippedQueue    CREATE_COMPLETE
PickedQueue     CREATE_COMPLETE
```

## Reproduction

```bash
sleep 10

cat > /tmp/cond-repro.json <<'JSON'
{
  "Conditions": {
    "IsFalse": { "Fn::Equals": ["a", "b"] }
  },
  "Resources": {
    "SkippedQueue": {
      "Type": "AWS::SQS::Queue",
      "Condition": "IsFalse",
      "Properties": { "QueueName": "should-not-exist" }
    },
    "PickedQueue": {
      "Type": "AWS::SQS::Queue",
      "Properties": {
        "QueueName": { "Fn::If": ["IsFalse", "true-branch", "false-branch"] }
      }
    }
  }
}
JSON

export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_REGION=us-east-1

aws cloudformation create-stack --stack-name cond-repro \
    --template-body file:///tmp/cond-repro.json
sleep 6

# Expected: only "false-branch". Actual: "should-not-exist" is there too.
aws sqs list-queues
```

## Why it matters

Templates generated by IaC tools commonly ship every optional resource in a single template and gate them with conditions, so that one template serves several environments. When the conditions are ignored, a stack meant to create a handful of resources instead tries to provision all of them — including ones deliberately turned off for that environment.

That is worse than a missing feature when a template contains custom resources: those are backed by Lambdas that call third-party APIs, so a resource that should have been skipped can produce real side effects outside the emulator.

Happy to test a fix if that helps.

## Environment

- Floci version / image tag: `floci/floci:latest` (reports `floci 1.6.0 native` at startup)
- Java SDK version (if applicable): n/a — reproduced with the AWS CLI v2 (`amazon/aws-cli:latest`)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker, default settings, no extra environment variables

Z9RTM:
## Summary

CloudFormation was not fully honoring resource-level `Condition` values. Two related gaps:

- **CreateStack**: resources whose `Condition` evaluated false were skipped for execution, but the dependency graph / topological sort still iterated over `allIds` (including inactive resources), so ordering and in-degree bookkeeping mixed active and inactive resources.
- **UpdateStack**: when a resource's `Condition` flipped from true to false, the resource was merely removed from the execution list. The previously-provisioned backing resource (SQS queue, log group, etc.) and its `StackResource` entry were never deleted, so `DescribeStackResources` kept returning it — diverging from real AWS, which deletes resources whose condition becomes false.

This PR:

- Computes an explicit `activeIds` set from the resource-level conditions and drives the dependency graph, in-degree map, and topological sort from it, so inactive resources never participate in ordering.
- Adds `deleteInactiveConditionResources(...)`, invoked during update, which deletes resources whose `Condition` is now false: it walks existing `StackResource`s in reverse order, respects retained/`DeletionPolicy` resources via `skipRetainedResource`, emits `DELETE_IN_PROGRESS` / `DELETE_COMPLETE` (or `DELETE_FAILED`) events, calls `provisioner.delete(...)`, and removes the entry from `stack.getResources()`. Resources removed from the template entirely remain outside this condition-specific cleanup.

Failures during deletion are logged with resource type, physical id, and stack name (per the no-empty-catch logging rule) and surfaced as `DELETE_FAILED`.

closes: #2168 

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior before this fix: on `UpdateStack`, a resource whose `Condition` turned false stayed provisioned and continued to be returned by `DescribeStackResources`, whereas AWS CloudFormation deletes such resources. On `CreateStack`, inactive-condition resources were skipped for provisioning but still polluted the dependency ordering pass.

Verified with the added integration test (`CloudFormationResourceConditionIntegrationTest`), which drives the CloudFormation Query API over the wire (RestAssured) and asserts:

- `createStack_skipsResourceWhenResourceLevelConditionIsFalse` — a false-condition resource is never provisioned.
- `updateStack_deletesResourceWhenConditionTurnsFalse` — flipping a condition to false deletes the resource so it no longer appears in `DescribeStackResources`.

## Checklist


abanna (changes requested):
Thank you for improving CloudFormation resource-level condition handling and for covering condition transitions. All 3 focused integration tests pass. One dependency case remains: `topologicalSort` removes condition-false resources from `activeIds`, but an active resource depending on one of them is still assigned an in-degree of zero and created. AWS specifies that when a resource is not created because of its condition, resources depending on it are not created either, regardless of their own conditions: [CloudFormation `Condition` attribute](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-condition.html). Please propagate the skip through dependent resources and add a regression test. Thank you for addressing this subtle lifecycle behavior.

hectorvent:
Thank you, and this is a strong first PR. The cleanup runs after provisioning and after the rollback guard, which matches CloudFormation's own `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS` phase.

1. **(blocking)** Seconding abanna's finding, which I confirmed in source: `inDegree` is built only from `activeIds`, so `inDegree.containsKey(dep)` is false for a condition-false dependency, and the dependent keeps in-degree zero and gets provisioned with a reference that cannot resolve.

One caveat on the fix. I would not assume skipping the dependent is the answer. AWS's own linter treats this template shape as an error (`E3005` for DependsOn) or a warning (`W1001` for Ref/GetAtt), which reads as "this template is invalid" rather than "the engine skips quietly". Worth a real AWS capture to choose between skipping and failing.

2. On the bot's deletion-failure comment, your reasoning is right. `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS` is a real stack status, so a failed cleanup delete does not roll the update back.

Z9RTM:
@abanna @hectorvent Thanks both for the detailed review—this was super insightful and I learned a lot. 

You were right about the core issue: inDegree was built only from activeIds, so a condition-false dependency had no entry, the dependent kept in-degree zero, and it got provisioned with a reference that can't resolve.

**On skip vs. fail (@hectorvent's caveat):** rather than assume, I captured the behavior against real CloudFormation. It fails synchronously at `CreateStack`/`UpdateStack` — it does not skip:

| Case | Template shape | Real AWS result |
|------|----------------|-----------------|
| A | `DependsOn` → condition-false resource | `ValidationError: Template format error: Unresolved resource dependencies [SkippedQueue]` (no stack created) |
| B | `Ref`/`Fn::Sub` → condition-false resource | same `ValidationError` |
| C | `Fn::If` selecting the branch that does *not* reference the condition-false resource | stack `CREATE_COMPLETE`, dependent created, condition-false resource skipped |

So the fix direction moved from "propagate the skip" to "reject the template," which matches AWS for both `DependsOn` and `Ref/GetAtt/Fn::Sub`.

**What changed:**
- Validate the condition dependency graph in `createChangeSet`, before any stack state is created, and throw `ValidationError` with the AWS message `Template format error: Unresolved resource dependencies [...]`. This makes `CreateStack`/`UpdateStack` fail synchronously and leave no stack behind, matching case A/B.
- `topologicalSort` throws the same error as a safety net for nested/direct executions.
- Dependency collection only walks the selected `Fn::If` branch, so case C stays valid (regression covered).
- Regression tests: A and B fail with `ValidationError`; C is created; plus a false→true condition update that provisions the previously-skipped resource.

**On the deletion-failure comment (@hectorvent #2):** agreed, no change — `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS` is a real status, so a failed cleanup delete does not roll the update back.

**One separate finding while validating #2168 (false → true update):** the condition behavior is correct (the previously-skipped resource is created, and `Fn::If` flips to the true branch). But real AWS also *replaces* the always-on queue because its `Fn::If`-driven `QueueName` changes, and deletes the old physical queue during `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS`. Floci currently creates the new queue but leaves the old one behind. This is a generic update-replacement cleanup gap (it happens for any immutable-property change, independent of conditions), so I'd suggest tracking it as a separate issue rather than widening this PR. Happy to open it.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
```diff
@@ -767,6 +837,54 @@ private Map<String, Boolean> resolveConditions(JsonNode template, Map<String, St
...
+            return;
+        }
+        JsonNode resources = template.path("Resources");
+        if (!resources.isObject()) {
+            return;
+        }
+
+        Map<String, String> resolvedParams = resolveDefaultParameters(template, params);
+        Map<String, Boolean> conditions =
+                resolveConditions(template, resolvedParams, null, region, regionResolver.getAccountId());
```
abanna: One cross-account case remains here. `validateConditionDependencies` evaluates conditions with `regionResolver.getAccountId()`. For StackSet deployments, `StackSetService.deployInstance` calls `createChangeSet` in the administrator request context, then supplies the target account only to `executeChangeSet`. That means conditions using `AWS::AccountId` are preflighted against the administrator instead of the target.

I reproduced this with admin `111111111111`, target `222222222222`, and a condition checking whether `AWS::AccountId` equals the target. A conditional resource and its dependent should both be active in `222222222222`, but `CreateStackInstances` returns HTTP 400 with `Unresolved resource dependencies`.

Could we pass the target account into this validation (or otherwise run preflight in the same account context as execution) and add a StackSet regression test? Thank you.

abanna (changes requested):
Thank you for the detailed follow-up and for validating the dependency behavior against real AWS. The original dependency concern is now addressed correctly, and the focused condition and StackSet test classes pass. I found one remaining cross-account preflight issue, noted inline. Once the target account is used consistently during validation and covered by a StackSet regression test, this looks good to me. Thanks again for the careful work.

src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
Z9RTM: Thank you for the detailed catch and the clear reproduction steps — you're right, this was preflighting `AWS::AccountId` conditions against the administrator account rather than the target. 

I've addressed it by threading the target account through the preflight: `createChangeSet` now takes an explicit accountId that flows into `validateConditionDependencies` (and `resolveConditions`), and `StackSetService.deployInstance` passes the instance's target account so preflight runs in the same account context as execution. Change-set and stack identifiers stay scoped to the caller account as before.

I also added StackSet regression tests covering this: an `AWS::AccountId` conditional resource with a dependent that must succeed in the target account on both `CreateStackInstances` and `UpdateStackSet`, plus the mirror case where the condition is false in the target account and preflight is expected to reject with Unresolved resource dependencies.

Thanks again for flagging this!

pgermosen (commented):
@Z9RTM — two separate things are going on here, and only one of them is visible from the CI output. Apologies in advance: the second one is real work, not a rebase.

## The red build is not yours

The failing run was created at `2026-08-14T10:07:15Z` and died with:

```
[INFO] 1 error
```

One error, in an EC2 test this PR does not touch. `main` briefly did not compile (#2287), and the fix (#2288) merged at `11:04Z` — an hour after your run started. Re-running will not clear it either, because a re-run replays the merge commit pinned when the run was created rather than recalculating against a newer `main`. So a down-merge is needed regardless.

## But a down-merge alone will not make this green

I merged your branch onto current `main` and ran the suite. **517 CloudFormation tests, one failure — and it is one of yours:**

```
CloudFormationResourceConditionIntegrationTest
  .updateStack_whenConditionDeletionFails_completesAndLeavesResourceUnmanaged

Expected: a string containing "<StackStatusReason>"
  Actual:  ...<StackStatus>UPDATE_COMPLETE</StackStatus>...   (no reason element)
```

I isolated it before concluding anything: on **your** base the class is **12/12 green**. Only the merge with current `main` fails. So this is not something you broke — it is a collision that landed underneath you.

## The collision

#1867 (`feat(stepfunctions): add UpdateStateMachine and update in place on stack updates`) merged as `97ac3cab` and reworked the update-cleanup path in `CloudFormationService`. It now reports failed cleanup deletions the same way your PR does — independently, and in a different place.

After the merge there are two writers to the same field in one update flow:

| | Line | Fed by | Message |
|---|---|---|---|
| yours | `CloudFormationService:754` | `conditionDeletionFailures` | *"The following resource(s) failed to delete and were removed from the stack: […]"* |
| #1867 | `CloudFormationService:911` | `cleanupFailures` | *"The following resource(s) could not be deleted during update cleanup: […]"* |

Both call `stack.setStatusReason(...)` and both emit an `UPDATE_COMPLETE` event. #1867's runs **later**, so it wins — and its `statusReason` is `null` here, because its `cleanupFailures` list has no knowledge of condition-driven deletions. Your reason is computed correctly, set, and then overwritten with null a moment later.

That is why the merge is textually clean and the behaviour still regresses: both features are present, and one silently cancels the other.

## What I think this needs

Rather than setting the reason independently, route the condition-deletion failures into the mechanism #1867 introduced — feed them into `cleanupFailures` (or whatever shape you prefer) so a single writer produces the final `statusReason`. The two messages describe the same event to a user, so they probably want to become one string rather than two competing ones.

The exact shape is your call — you know this path better than I do after this PR, and I would rather not prescribe an API for code you are actively working in. Happy to look at whatever you land on.

## For what it is worth

The rest of the change looks good from here, and the condition-dependency preflight in particular — rejecting an unresolvable graph *before* any stack state is written, rather than failing partway through — is the right instinct and matches how real CloudFormation behaves.

Sorry this one drew a bad ticket: an unrelated broken base, and then a genuine conflict from a PR that merged while yours was in review. Neither was avoidable from where you were sitting.

Z9RTM:
@pgermosen Thanks for digging into this and clearly identifying the interaction with #1867. Isolating the failure on the merged tree saved me a lot of time.

I down-merged the current `main` (`42a79a7f`) into the branch in merge commit `5a7507da`.

The root cause was exactly as you described: `finishCommittedStackUpdate` ran after the condition-deletion path and overwrote its `statusReason` with `null`.

I resolved this by routing both cleanup paths through the mechanism introduced by #1867:

- `deleteInactiveConditionResources` now returns `UpdateCleanupFailure` entries containing the logical ID, physical ID, and failure reason.
- Condition-driven cleanup now runs in the committed update-cleanup phase.
- Its failures are combined with those returned by `finishCommittedResourceCleanup`.
- The combined list is passed once to `finishCommittedStackUpdate`, which is now the single writer of the final `UPDATE_COMPLETE` status and reason.
- The separate condition-specific completion reason and final status writer have been removed.

This preserves the existing behavior: a failed condition-driven deletion leaves the resource unmanaged, completes the update with `UPDATE_COMPLETE`, and reports the cleanup failure in `StackStatusReason`.

Verification on the merged tree:

- The previously failing condition-deletion regression test passes.
- `CloudFormationStepFunctionsCleanupIntegrationTest` passes, including the successful retry case that expects no `StackStatusReason`.
- The full CloudFormation test package passes: **533 tests, 0 failures**.

Thanks again for the detailed investigation and suggested direction.

pgermosen (approved):
Thank you for the careful, thorough work through all of these rounds -- this held up under a lot of scrutiny and came out stronger for it.

Re-verified the two most recent findings directly rather than taking the replies on faith. The `statusReason` overwrite from the #1867 merge collision is genuinely fixed: both cleanup paths now feed a single writer (`finishCommittedStackUpdate`), so there's no more race between the two features silently cancelling each other. And Greptile's newest finding -- a failed deletion dropping the resource from tracking -- is also fixed: the resource stays under stack management as `DELETE_FAILED` on the catch path instead of being removed, so it can't get orphaned.

Also appreciated the real AWS capture to settle the skip-vs-fail question for condition-dependent resources rather than guessing -- that's exactly the right instinct, and the synchronous `ValidationError` preflight matches what you found.

Ran the full CloudFormation suite fresh: 533/533 pass, matching your own count exactly.

Approving -- nice work, and a great first PR.

PR #2256, opened 2026-08-12
feat(logs): support log group deletion protection

aryasaatvik:
## Summary

Add the complete CloudWatch Logs deletion-protection lifecycle: create-time configuration, persisted updates by name or ARN, describe readback, and protected-delete enforcement.

| Request | Behavior |
|---|---|
| `CreateLogGroup` without `deletionProtectionEnabled` | Defaults to `false` |
| `CreateLogGroup` with protection enabled | Persists and reports `true` |
| `PutLogGroupDeletionProtection` with a name or ARN | Enables or disables protection |
| `DescribeLogGroups` | Returns `deletionProtectionEnabled` |
| `DeleteLogGroup` while protected | Returns `ValidationException` until protection is disabled |

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified with AWS SDK for Java 2.44.14 against an isolated Floci instance. The SDK lifecycle test covers default and create-time state, name and ARN updates, describe readback, protected deletion, and deletion after disabling protection.

## Checklist

## Validation

- `./mvnw test -Dtest=CloudWatchLogsServiceTest,CloudWatchLogsHandlerTest` — 60 passed
- `FLOCI_ENDPOINT=http://127.0.0.1:14566 ../../mvnw test -Dtest=CloudWatchLogsDeletionProtectionTest` — 2 passed against an isolated Floci instance
- `make PYTHON='uv run --with pytest --with pyyaml python' docs-test docs-check` — 20 tooling tests passed; docs clean
- `./mvnw test` — 9,286 tests, 0 failures, 0 errors, 9 skipped

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudWatchLogsDeletionProtectionTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudWatchLogsDeletionProtectionTest.java
@@ -1,0 +1,97 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClient;
+import software.amazon.awssdk.services.cloudwatchlogs.model.LogGroup;
+import software.amazon.awssdk.services.cloudwatchlogs.model.ValidationException;
+
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+@DisplayName("CloudWatch Logs deletion protection")
+class CloudWatchLogsDeletionProtectionTest {
+
+    @Test
+    @DisplayName("defaults off and can be enabled and disabled by name")
+    void defaultsOffAndUpdatesByName() {
+        String groupName = "/test/" + TestFixtures.uniqueName("logs-deletion-protection-default");
+        try (CloudWatchLogsClient logs = TestFixtures.cloudWatchLogsClient()) {
+            try {
+                logs.createLogGroup(request -> request.logGroupName(groupName));
+                assertThat(findLogGroup(logs, groupName))
+                        .get()
+                        .extracting(LogGroup::deletionProtectionEnabled)
+                        .isEqualTo(false);
+
+                logs.putLogGroupDeletionProtection(request -> request
+                        .logGroupIdentifier(groupName)
+                        .deletionProtectionEnabled(true));
+                assertThat(findLogGroup(logs, groupName))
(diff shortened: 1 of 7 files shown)


abanna (approved):
Thank you for adding CloudWatch Logs deletion protection with both service and handler coverage. I ran the focused suites—all 60 tests passed—and the full CI set is green. The PR is focused, documented, conventionally named, and follows the project’s testing policy. Approved.

pgermosen (approved):
Welcome, and thank you — this is an unusually clean first contribution. Every AWS detail is right, including the ones that are easy to get wrong.

None of the botocore models available to me carry `PutLogGroupDeletionProtection` yet, so I verified against the API reference directly rather than the SDK model. All four rows of your behaviour table are documented:

| claim | reference |
|---|---|
| `CreateLogGroup` accepts `deletionProtectionEnabled`, defaulting to false | *"Use this parameter to enable deletion protection for the new log group… By default log groups are created without deletion protection enabled."* |
| the put operation takes a name **or** ARN | `logGroupIdentifier` — *"The name or ARN of the log group"*, with `deletionProtectionEnabled`, both **Required** |
| `DescribeLogGroups` reports it | `LogGroup.deletionProtectionEnabled` exists on the data type |
| protected delete is refused | `ValidationException` is declared on `DeleteLogGroup` |

Four things I want to call out specifically, because they are the details that separate a working implementation from a faithful one:

- **You used `logGroupIdentifier`, not `logGroupName`.** Every neighbouring operation in this handler takes `logGroupName`, so copying the pattern would have been the natural mistake. AWS genuinely changed the parameter for this operation, and resolving name-or-ARN through `extractLogGroupNameFromArn` matches it.
- **Both parameters are `Required: Yes` in AWS, and you enforce both** — with `InvalidParameterException` 400, which is one of the errors actually declared on that operation rather than a plausible-sounding substitute.
- **Missing group returns `ResourceNotFoundException` with HTTP 400**, matching the declared status. The reflexive choice is 404, and it would have been wrong; CloudWatch Logs is AWS JSON 1.1 and declares 400 for that shape.
- **`.asBoolean(false)` on create** matches the documented default exactly, and keeping the old four-argument `createLogGroup` as an overload means nothing else in the codebase had to change.

One note for the record, not a change request: the reference doesn't explicitly state that deletion protection surfaces as `ValidationException` on delete. It is an inference — but a well-supported one, since `ValidationException` is declared on `DeleteLogGroup` and none of the other declared errors (`InvalidParameterException`, `ResourceNotFoundException`, `OperationAbortedException`) fits a protected-resource refusal. Worth knowing the provenance if it is ever challenged.

15/15 checks green, merges clean. Adding the `enhancement` and `cloudwatch` labels, which were missing.

Approving.

Issue #2228, opened 2026-08-10
[BUG] cloudwatch: pre-fix persisted LogEvents deserialize with logStreamName: null under hybrid storage (residue of #2210)


sam-fakhreddine:
**What's wrong**: `LogEvent.logStreamName` (added in #2210) is populated only at ingestion time, in `putLogEvents`, where `streamName` is already in scope. There is no backfill/reconstruction path for events that were persisted *before* that change landed. Any such event deserializes with `logStreamName: null`, and `FilterLogEvents` will emit `"logStreamName": null` for it in the response — a shape real AWS never returns.

**Why this is a data-quality gap, not just cosmetic**: while the container's event store was in-memory-only, this was harmless — a restart wiped every pre-fix event along with the process, so the null-`logStreamName` cohort could never outlive a single run. The container moved to hybrid storage on 2026-08-10 (the same day as the #2210 fix), so persisted events now genuinely survive restarts. That means the null-`logStreamName` cohort is no longer transient: any event ingested before the #2210 fix landed will carry `logStreamName: null` forever, for as long as it remains in storage.

**Severity**: 3, unless a specific `FilterLogEvents` consumer is known to hard-fail on a null `logStreamName` field (in which case this should be bumped — please say so here if you know of one).

**Relationship to #2210**: this is residue of that fix, not a regression it introduced. #2210 correctly populates the field going forward; this issue is about the cohort of events that predate the fix and were never backfilled.

**Suggested fix**: a one-time backfill/migration pass over persisted events at startup (or lazily on first read) that reconstructs `logStreamName` from the storage key, since the key already encodes group/stream/event (see `eventKeyPrefix` in `CloudWatchLogsService.java`). Alternatively, treat it as an acceptable gap and document that `logStreamName` is only guaranteed non-null for events ingested after 2026-08-10.

Filed per SEE-SOMETHING while working #2209/#2210 (branch `feature/cwl-filter-log-events`, merged into `integration/all-features`). Kept out of the #2210 diff to stay minimal-diff scoped.

hampsterx:
On `main` there's nothing persisted to backfill: `logStreamName` is derived per read rather than stored.

[...]

PR #2259, opened 2026-08-12
fix(cloudwatch): return logStreamName on FilterLogEvents results

### LINKED ISSUE #2210: [BUG] cloudwatch: buildEventsArray omits logStreamName from FilterLogEvents results (CloudWatchLogsHandler.java:467-478)
## What's wrong

Log events serialized by `CloudWatchLogsHandler.buildEventsArray` omit `logStreamName`. Real AWS includes it on every `FilterLogEvents` result.

`src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsHandler.java:467-478`:

```java
private ArrayNode buildEventsArray(List<LogEvent> events) {
    ArrayNode array = objectMapper.createArrayNode();
    for (LogEvent e : events) {
        ObjectNode node = objectMapper.createObjectNode();
        node.put("eventId", e.getEventId());
        node.put("timestamp", e.getTimestamp());
        node.put("message", e.getMessage());
        node.put("ingestionTime", e.getIngestionTime());
        array.add(node);
    }
    return array;
}
```

This method serves **both** `GetLogEvents` and `FilterLogEvents` (`handleGetLogEvents` ~`:171-186`, `handleFilterLogEvents` `:188-206`).

## Why it matters

For `GetLogEvents` the omission is harmless — the caller named the stream in the request. For `FilterLogEvents` it is a real fidelity gap: the API is specifically the cross-stream one, and `logStreamName` is how a caller attributes a matched event back to the stream that emitted it. Without it, a client searching a group can say *which group* a hit came from but not *which stream*, so tooling that works against real AWS silently loses attribution against floci.

Verified live against the running emulator: `filter-log-events` on `/aws/codebuild/AWSAccelerator-InstallerProject` returns events with `eventId`, `timestamp`, `message`, `ingestionTime` and no `logStreamName`.

## Suggested fix

Pass the stream name through to the serializer and emit it. `LogEvent` already carries enough context to resolve it, and the change is roughly two lines. If `GetLogEvents` parity is also wanted, real AWS omits it there — so gate the field on the calling operation rather than always emitting it.

Found while planning a CloudWatch Logs view for the dashboard; not fixing it here.

hampsterx:
## Summary

- `FilterLogEvents` returns `logStreamName` on every event, so a caller searching a whole log group can attribute each match to the stream that emitted it. Real AWS returns it on every `FilteredLogEvent`; Floci served `FilterLogEvents` and `GetLogEvents` from one serializer that omitted it, so cross-stream attribution was lost with no error to signal it.
- `GetLogEvents` is unchanged. Its `OutputLogEvent` carries no `logStreamName` in real AWS, because the caller named the stream in the request.
- The stream name is derived from the event's storage key rather than persisted on the event, so it is already correct for events written before this change and needs no backfill.
- `filterLogEvents` walks the keyset once instead of running one full keyset scan per requested stream, since a scan walks every key regardless of the prefix it filters on.

Closes #2210

## Why derive rather than store

The issue suggests carrying the stream name through to the serializer, which most directly means setting it on `LogEvent` at ingestion. That is what #2228 anticipates: every event already in the store deserializes with `logStreamName: null`, and since events now survive restarts, that cohort is permanent without a backfill pass.

Reading the name off the storage key avoids that class of problem. Keys are already `<region>::<group>::<stream>::<timestamp>::<uuid>`, so the key is an authoritative record of the emitter for every event ever written. Nothing is persisted twice, there is no null cohort, and there is no migration to run.

The name is read back between the caller's group prefix and the trailing timestamp and uuid, both generated in `eventKey`, so it survives a stream name containing a `:`. AWS forbids that, but `createLogStream` does not enforce it, so the parse does not depend on the restriction holding.

If that reading is right, #2228 can be closed alongside this. Happy to add a backfill instead if you would rather the field live on the event.

## Test plan

- [x] `CloudWatchLogsServiceTest`: per-event stream attribution across two streams, stream-restricted filtering, prefix-sibling group isolation (`/app/logs` must not sweep up `/app/logs-archive`), and a stream name containing `:`
- [x] `CloudWatchLogsHandlerTest`: `logStreamName` present on `FilterLogEvents` results in both the name and ARN request forms, absent from `GetLogEvents`
- [x] `./mvnw test -Dtest=CloudWatchLogsServiceTest,CloudWatchLogsHandlerTest` — 56 pass
- [ ] Full local suite is inconclusive on this machine: `ElastiCacheIntegrationTest`, `ElbV2*IntegrationTest` and `ApiGatewayV2AlbIntegrationTest` fail on host port collisions (6379 and 7782/7793/7794 already bound), identically on a branch that touches none of that code. Leaving the full run to CI.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsHandler.java
@@ -194,7 +194,7 @@ public class CloudWatchLogsHandler {
                 logsService.filterLogEvents(groupName, streamNames, startTime, endTime, filterPattern, limit, region);
 
         ObjectNode response = objectMapper.createObjectNode();
-        response.set("events", buildEventsArray(result.events()));
+        response.set("events", buildFilteredEventsArray(result.events()));
         if (result.nextToken() != null) {
             response.put("nextToken", result.nextToken());
         }
@@ -450,14 +450,34 @@ public class CloudWatchLogsHandler {
     private ArrayNode buildEventsArray(List<LogEvent> events) {
         ArrayNode array = objectMapper.createArrayNode();
         for (LogEvent e : events) {
-            ObjectNode node = objectMapper.createObjectNode();
-            node.put("eventId", e.getEventId());
-            node.put("timestamp", e.getTimestamp());
-            node.put("message", e.getMessage());
-            node.put("ingestionTime", e.getIngestionTime());
+            array.add(buildEventNode(e));
+        }
+        return array;
+    }
+
+    /**
+     * Serializes FilteredLogEvent, which carries {@code logStreamName} on top of the GetLogEvents
+     * shape. The field is what attributes a match back to the stream that emitted it, and real AWS
+     * returns it on every FilterLogEvents result; GetLogEvents omits it, so it stays out of
+     * {@link #buildEventsArray}.
+     */
+    private ArrayNode buildFilteredEventsArray(List<CloudWatchLogsService.FilteredEvent> events) {
+        ArrayNode array = objectMapper.createArrayNode();
+        for (CloudWatchLogsService.FilteredEvent filtered : events) {
+            ObjectNode node = buildEventNode(filtered.event());
+            node.put("logStreamName", filtered.logStreamName());
             array.add(node);
         }
         return array;
+    }
+
+    private ObjectNode buildEventNode(LogEvent e) {
+        ObjectNode node = objectMapper.createObjectNode();
+        node.put("eventId", e.getEventId());
+        node.put("timestamp", e.getTimestamp());
+        node.put("message", e.getMessage());
+        node.put("ingestionTime", e.getIngestionTime());
+        return node;
     }
 
     private Map<String, String> extractTags(JsonNode tagsNode) {
--- a/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java
@@ -386,7 +386,14 @@ public class CloudWatchLogsService {
         }
     }
 
-    public record FilteredLogEventsResult(List<LogEvent> events, String nextToken) {}
+    /**
+     * A FilterLogEvents match paired with the stream that emitted it. FilterLogEvents is the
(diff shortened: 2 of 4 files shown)


pgermosen (approved):
Approving on `0fb01e61`. Thank you — this is a small fix that turned out to have more underneath it than the title suggests, and the extra care shows.

**The bug and the fix both match the model.** `FilteredLogEvent` declares `logStreamName`, so `FilterLogEvents` has to return it; `OutputLogEvent` does not, so `GetLogEvents` must not. Keeping `buildEventsArray` and `buildFilteredEventsArray` as separate serializers is therefore the right call rather than over-engineering — the two AWS shapes genuinely differ, and collapsing them would put the field on the wrong response sooner or later.

**Verified red-green rather than reading the tests.** I removed the single line that adds the field and re-ran:

| | result |
|---|---|
| `CloudWatchLogsHandlerTest` | 17 tests, **2 failures** |
| `CloudWatchLogsServiceTest` | 39 tests, 0 failures |

A precise bisect: the handler tests pin the serialization and the service tests cover the stream-name plumbing independently, so a regression in either half would be attributable on its own.

Locally on this head: 82 tests green across the three CloudWatch Logs suites. All 15 CI checks pass.

## The rewrite is the part I looked hardest at, and it holds

Recovering the stream name from the key is the interesting change here, not the field itself, so two things I checked rather than assumed:

**The account prefix lines up.** `AccountAwareStorageBackend.keys()` strips the `accountId/` prefix and hands back logical keys, so `key.startsWith(groupPrefix)` is comparing like with like. If it had returned raw keys this would have silently matched nothing, and a test asserting "no results" would have looked green for the wrong reason.

**The key parsing is unambiguous.** `streamNameFromEventKey` anchors from the end — the last `::` precedes the uuid, the one before it precedes the timestamp — so the stream name is recovered by subtraction rather than by counting separators forward. That survives a stream name containing `::`, which is a nice property to have even though AWS makes it unreachable:

```
LogGroupName:  [\.\-_/#A-Za-z0-9]+   — no colons
LogStreamName: [^:*]*                — colons explicitly excluded
```

So the parse is safe by AWS's own naming rules, with the end-anchoring as belt-and-braces on top. I would rather that than a forward `split("::")[2]`, which would have been correct today and fragile forever.

Your performance note also checks out: a stream-restricted filter was previously one scan per requested stream, each walking the whole keyset, and is now a single pass. It also dedupes for free if a caller passes the same stream name twice, which the old loop would have double-counted.

Nice piece of work — thank you for fixing the underlying shape rather than just appending the missing field.

PR #2261, opened 2026-08-13
fix(cloudformation): pass the CloudFront service argument in the provisioner tests

pgermosen:
> [!NOTE]
> ## Stack position: this is the base
>
> **#2262 is stacked on top of this PR** and targets this branch, not `main`.
>
> Merge this one **first**; #2262 then retargets to `main` automatically.
>
> This PR is self-contained — review it on its own. It changes test sources only.

`main` does not compile.

## Cause

#1820 added a `CloudFrontService cloudFrontService` parameter to the `CloudFormationResourceProvisioner` constructor. Three tests that construct that provisioner directly landed from other branches around the same time and were never updated:

| Test | From |
|---|---|
| `EventBusProvisionOwnershipTest` | #1807 |
| `EventBusDeleteOwnershipTest` | #1807 |
| `DynamoDbReplicaCfnProvisionerTest` | #1811 |

None of those PRs touched #1820's files, so each passed CI on its own branch and the break only appeared once they had all landed:

```
constructor CloudFormationResourceProvisioner in class
io.github.hectorvent.floci.services.cloudformation.CloudFormationResourceProvisioner
cannot be applied to given types;
```

`./mvnw test-compile` fails on `main`, so no test can run.

## Fix

All three supplied 33 arguments to a 34-parameter constructor. Adds the missing `null` in the `cloudFrontService` position, immediately before the resource registry argument.

Since this has now recurred twice, I swept **every** remaining `new CloudFormationResourceProvisioner(` call site in the tree rather than fixing them reactively. The other six already match, so nine of nine now agree with the constructor.

Test sources only — no production behaviour changes.

## Verification

- `./mvnw test-compile` — clean.
- `./mvnw test -Dtest='EventBusProvisionOwnershipTest,EventBusDeleteOwnershipTest,DynamoDbReplicaCfnProvisionerTest'` — 15 tests, 0 failures, 0 errors.
- Rebased onto current `main` (`a29464ab`).

## Note for maintainers

This is the second time a constructor change to `CloudFormationResourceProvisioner` has broken `main` through a semantically-conflicting merge that Git saw no textual conflict in. A 34-argument constructor built entirely of positional `null`s in tests makes this failure mode both easy to hit and silent until merge. Worth considering a test builder or a small parameter object so adding a service does not require editing every fixture — happy to open a separate issue if that is wanted.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/DynamoDbReplicaCfnProvisionerTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/DynamoDbReplicaCfnProvisionerTest.java
@@ -34,7 +34,7 @@ class DynamoDbReplicaCfnProvisionerTest {
         provisioner = new CloudFormationResourceProvisioner(
                 null, null, null, dynamoDbService, null, null, null, null, null, null,
                 null, null, null, null, null, null, mapper, null, null, null, null, null,
-                null, null, null, null, null, null, null, null, null, null,
+                null, null, null, null, null, null, null, null, null, null, null,
                 new CloudFormationResourceRegistry(List.of()));
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/EventBusDeleteOwnershipTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/EventBusDeleteOwnershipTest.java
@@ -45,7 +45,7 @@ class EventBusDeleteOwnershipTest {
                 new ObjectMapper(),
                 null, null, null, null, null, null, null,
                 null, null, null, null, null, null, null,
-                null,
+                null, null,
                 new CloudFormationResourceRegistry(List.of()));
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/EventBusProvisionOwnershipTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/EventBusProvisionOwnershipTest.java
@@ -58,7 +58,7 @@ class EventBusProvisionOwnershipTest {
                 MAPPER,
                 null, null, null, null, null, null, null,
                 null, null, null, null, null, null, null,
-                null,
+                null, null,
                 new CloudFormationResourceRegistry(List.of()));
     }
 


abanna:
I pulled this PR locally and verified that the two EventBus constructor fixes are correct: the parent fails test compilation, while `db8aaf9d` compiles and passes both affected suites (12/12 tests).

However, `main` has since advanced to `a29464ab` with #1811. I tested a merge of this PR into that current head, and test compilation still fails at `DynamoDbReplicaCfnProvisionerTest.java:34` for the same missing `CloudFrontService` argument.

Adding one more `null` immediately before `CloudFormationResourceRegistry` in that fixture makes test compilation succeed and all 15 affected EventBus/DynamoDB tests pass.

Could you update this PR to include that newly landed constructor call as well? That will ensure merging #2261 actually restores `main` compilation before #2262 is retargeted.

---

Update after `09868b0c`: this is addressed. I reviewed the updated commit and maintainer response, confirmed the missing `CloudFrontService` argument is now present in `DynamoDbReplicaCfnProvisionerTest`, and re-swept all nine direct `CloudFormationResourceProvisioner` call sites. The updated #2261 head passes the three affected suites (15/15).

I also fetched the rebased stacked PR #2262 at `93645e41` and physically tested the intended merge order (`main` → #2261 → #2262). It merges cleanly, and the combined IAM, Lambda, EventBus, and DynamoDB focused suite passes 47/47.

I reviewed the six-file #2262 delta and current comments as well. Its core response-shape choices match AWS: [`GetFunction` includes tags](https://docs.aws.amazon.com/lambda/latest/api/API_GetFunction.html), while [`ListRoles`](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListRoles.html) and [`ListPolicies`](https://docs.aws.amazon.com/cli/latest/reference/iam/list-policies.html) omit them. I found no additional blocking issue.

**Approved from my review.** I have no remaining findings on #2261. Once the in-progress CI checks are green, it is ready to merge. The suggested test builder or parameter object is worthwhile as a separate follow-up and should not block this focused repair.

pgermosen:
Thanks for pulling it locally and testing the merge against current `main` — you were right, and I had missed it.

Updated in `09868b0c`, rebased onto `a29464ab`:

- Added the missing `null` to `DynamoDbReplicaCfnProvisionerTest` from #1811, in the same `cloudFrontService` position immediately before the resource registry.
- Since this has now recurred twice, I swept **every** `new CloudFormationResourceProvisioner(` call site in the tree instead of fixing them reactively. The other six already matched, so nine of nine now agree with the 34-parameter constructor.

Verified:

- `./mvnw test-compile` — clean.
- `./mvnw test -Dtest='EventBusProvisionOwnershipTest,EventBusDeleteOwnershipTest,DynamoDbReplicaCfnProvisionerTest'` — 15 tests, 0 failures, 0 errors, matching your 15/15.

I also widened the title and description, since this is no longer only about the EventBus tests.

On the underlying pattern: this is the second time a constructor change to `CloudFormationResourceProvisioner` has broken `main` via a semantically-conflicting merge with no textual conflict for Git to catch. A 34-argument constructor filled with positional `null`s in fixtures makes that both easy to trip and invisible until after merge. I mentioned in the description that a test builder or a small parameter object would remove the failure mode — happy to open a separate issue and take it on if you think it is worth doing.

PR #2262, opened 2026-08-13
fix(iam,lambda): return resource tags in Get and Create responses

### LINKED ISSUE #2199: [BUG] IAM and Lambda omit resource tags from Get/Create responses, so tagged resources never converge under Terraform
## Service

IAM (`GetRole`, `CreateRole`, `GetPolicy`, `CreatePolicy`, `GetUser`, `CreateUser`) and Lambda (`GetFunction`).

## API Action / Feature

Tags are stored and are readable through the dedicated tag actions (`ListRoleTags`, `ListPolicyTags`, `ListUserTags`, Lambda `ListTags`), but they are never serialized into the `Get*` / `Create*` responses.

The AWS SDKs and the Terraform provider read a resource's tags from those responses, not by calling `List*Tags`. The result is that a tagged resource applies cleanly and then reads back untagged, so `terraform plan` reports the same tag diff forever and never converges. `terraform apply` succeeds every time, which is why an apply-only test does not notice.

Affected:

- `IamQueryHandler.roleXml` — no `<Tags>`; also drops `PermissionsBoundary` and `RoleLastUsed`
- `IamQueryHandler.policyXml` — no `<Tags>`
- `IamQueryHandler.userXml` — no `<Tags>`
- `LambdaController.getFunction` — no `Tags` field alongside `Configuration` and `Code`

## AWS Documentation

- https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetRole.html
- https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetPolicy.html
- https://docs.aws.amazon.com/lambda/latest/api/API_GetFunction.html

**The fix must be asymmetric, which is the subtle part.** AWS deliberately omits tags from the *listing* operations, so simply adding tags inside the shared XML builders would swap one deviation for another:

- [ListRoles](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListRoles.html) — *"IAM resource-listing operations return a subset of the available attributes for the resource. This operation does not return the following attributes, even though they are an attribute of the returned object: PermissionsBoundary, RoleLastUsed, Tags"*
- [ListPolicies](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListPolicies.html) — *"this operation does not return tags, even though they are an attribute of the returned object"*

So `roleXml`/`policyXml`/`userXml` need a detail-vs-list flag: tags emitted for `Get*`/`Create*`, omitted for `ListRoles`, `ListUsers`, `ListPolicies` and `GetGroup`.

#2155 already introduced exactly this shape for `Description` (`policyXml(IamPolicy, boolean includeDescription)`), and AWS excludes `Description` and `Tags` from `ListPolicies` identically — so the two should share one flag rather than growing a second one. Note `handleCreateServiceLinkedRole` (added in #2179) is a detail call site that needs the flag too.

An untagged resource should emit no `<Tags>` element at all, rather than an empty one, so it never reads back as carrying a tag set the caller never configured.

## Why is this needed?

Any Terraform configuration that tags IAM roles or Lambda functions — which is most of them, since tagging is usually enforced by policy — will never reach a clean plan against Floci. The failure is quiet: nothing errors, CI just never converges, and the cause looks like a provider bug rather than a missing response field.

This is also the class of bug that motivated adding a "second plan reports no changes" assertion to the Terraform compatibility suite (#2159): apply-only coverage cannot see it.

- [x] Yes
- [ ] No

pgermosen:
> [!NOTE]
> Rebased onto current `main` (`12154297`). Carries only its own two commits — `a8b6b27c` (production change) and `b3ff76b6` (tests). The earlier stacking on #2261 is resolved: that PR merged, so this targets `main` directly.

Closes #2199.

## Problem

IAM stores role, policy and user tags and serves them through `ListRoleTags`/`ListPolicyTags`/`ListUserTags`. Lambda does the same through `ListTags`. Neither ever serialized them into the Create/Get responses.

The AWS SDKs and the Terraform provider read tags off those responses rather than by calling `List*Tags` — the provider's transparent-tagging interceptor (`@Tags(identifierAttribute="arn")`) does this on every read. So every tagged `aws_iam_role`, `aws_iam_policy`, `aws_iam_user` and `aws_lambda_function` reads back untagged: the apply succeeds, and then `plan` never converges. An apply-only test cannot see this.

## The inclusion is asymmetric

Serializing tags unconditionally would deviate from AWS in the opposite direction. IAM's resource-listing operations deliberately return a subset of attributes:

| Operation | AWS documents |
|---|---|
| `ListRoles` | "does not return the following attributes, even though they are an attribute of the returned object: PermissionsBoundary, RoleLastUsed, **Tags**" |
| `ListUsers` | same, for PermissionsBoundary and **Tags** |
| `ListPolicies` | "this operation does not return **tags**" |

So `roleXml`, `userXml` and `policyXml` take a `detailed` flag, following the pattern #2155 established for the policy `Description`.

Two details worth a reviewer's eye:

- **A role's `Description` stays unconditional.** It is absent from the `ListRoles` exclusion list, so unlike policies it must keep appearing in listings. For policies one flag governs both `Description` and `Tags`, since `ListPolicies` excludes them on identical grounds — that is why #2155's `includeDescription` parameter is renamed to `detailed` rather than joined by a second flag.
- **Roles embedded in an instance profile are a subset too.** `GetInstanceProfile`'s and `ListInstanceProfiles`' own example responses both show the embedded `<member>` role without `<Tags>`, so `instanceProfileXml` passes `false`.

Call sites:

| `detailed = true` | `detailed = false` |
|---|---|
| `CreateRole`, `GetRole`, `CreateServiceLinkedRole` | `ListRoles` |
| `CreateUser`, `GetUser` | `ListUsers`, `GetGroup` |
| `CreatePolicy`, `GetPolicy` | `ListPolicies` |
| | roles embedded in `instanceProfileXml` |

`tagsElement` emits nothing rather than an empty `<Tags/>` for an untagged resource — an empty element reads back as a tag set the caller never configured, which is a diff of its own.

Lambda needs no such gate: `ListFunctions` returns `FunctionConfiguration`, which has no `Tags` field at all.

## Tests

`IamResourceTagsInResponseTest` (new, 12 cases) asserts **both** directions — the negative half is the point, since tests that only assert tags are present cannot catch over-inclusion. Verified by reverting the production change to the naive unconditional form: all 5 listing-subset cases fail, and pass again once the flag is restored.

`LambdaPermissionTagLayerIntegrationTest` gains a `GetFunction` case ordered after the untag, so it also pins that a removed tag is reflected rather than served stale.

Compat: a tagged role and a tagged policy in `compat-terraform`, with cases for the positive round trip, the listing-subset negatives, and **a clean second `terraform plan`** — the assertion that actually catches this drift class.

## Verification

- `./mvnw test` — full suite green on the rebased branch: 9929 tests, 0 failures, 0 errors, 9 skipped.
- IAM + Lambda targeted: 559 tests, 0 failures, 0 errors.
- `make docs-check` — clean; no handler actions change, so no `docs-sync` needed.
- `terraform fmt -check` — the only diff is a pre-existing unrelated `aws_db_instance` block on `main`, left untouched.

### DIFF
--- a/compatibility-tests/compat-terraform/main.tf
+++ b/compatibility-tests/compat-terraform/main.tf
@@ -458,3 +458,41 @@ resource "aws_iam_role_policy_attachment" "emr_service_role" {
 output "managed_policy_role_arn" {
   value = aws_iam_role.managed_policy_attach.arn
 }
+
+# The provider reads tags off the GetRole/GetPolicy/GetFunction response rather than by calling
+# List*Tags, so a tagged resource that does not echo them back applies cleanly and then diffs on
+# every subsequent plan. The re-plan assertion below is what catches that.
+resource "aws_iam_role" "tagged" {
+  name               = "floci-compat-tagged-role"
+  assume_role_policy = data.aws_iam_policy_document.lambda_assume_role.json
+
+  tags = {
+    Environment = "compat"
+    Owner       = "floci"
+  }
+}
+
+resource "aws_iam_policy" "tagged" {
+  name        = "floci-compat-tagged-policy"
+  description = "Tagged policy used to assert tags survive a round trip"
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [{
+      Effect   = "Allow"
+      Action   = "s3:GetObject"
+      Resource = "*"
+    }]
+  })
+
+  tags = {
+    Environment = "compat"
+  }
+}
+
+output "tagged_role_arn" {
+  value = aws_iam_role.tagged.arn
+}
+
+output "tagged_policy_arn" {
+  value = aws_iam_policy.tagged.arn
+}
--- a/compatibility-tests/compat-terraform/test/terraform.bats
+++ b/compatibility-tests/compat-terraform/test/terraform.bats
@@ -295,3 +295,55 @@ setup() {
     assert_failure
     assert_output --partial "NoSuchEntity"
 }
+
+# Tags are stored and readable via ListRoleTags, but the provider reads them off the
+# GetRole response. Omitting them there made every tagged role read back untagged.
+@test "Terraform: GetRole returns the role's tags" {
+    run aws_cmd iam get-role --role-name floci-compat-tagged-role \
+        --query "Role.Tags[?Key=='Environment'].Value" --output text
+    assert_success
+    assert_output "compat"
+}
+
+@test "Terraform: GetPolicy returns the policy's tags" {
+    policy_arn=$(aws_cmd iam list-policies --scope Local \
+        --query "Policies[?PolicyName=='floci-compat-tagged-policy'].Arn" --output text)
+    run aws_cmd iam get-policy --policy-arn "$policy_arn" \
+        --query "Policy.Tags[?Key=='Environment'].Value" --output text
+    assert_success
+    assert_output "compat"
+}
+
+# The other half of the contract: IAM's listing operations deliberately return a subset
+# of attributes. ListRoles "does not return the following attributes, even though they
+# are an attribute of the returned object: PermissionsBoundary, RoleLastUsed, Tags".
+# Emitting them here would deviate from AWS in the opposite direction.
+@test "Terraform: ListRoles omits tags even for a tagged role" {
+    run aws_cmd iam list-roles \
+        --query "Roles[?RoleName=='floci-compat-tagged-role'].Tags" --output text
+    assert_success
+    refute_output --partial "compat"
+}
+
+@test "Terraform: ListPolicies omits tags and description even for a tagged policy" {
+    run aws_cmd iam list-policies --scope Local \
+        --query "Policies[?PolicyName=='floci-compat-tagged-policy'].[Tags,Description]" \
+        --output text
+    assert_success
+    refute_output --partial "compat"
+    refute_output --partial "round trip"
+}
+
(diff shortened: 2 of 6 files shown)


pgermosen:
Self-review pass, applying the same rigor as any other PR.

Both AWS doc claims check out verbatim -- fetched `ListRoles` and `ListPolicies` directly: *"This operation does not return the following attributes... PermissionsBoundary, RoleLastUsed, Tags"* and *"this operation does not return tags"*, matching the PR body exactly.

Every `detailed` call site traced independently: `CreateRole`/`GetRole`/`CreateServiceLinkedRole` → `true`, `ListRoles` → `false`, same pattern for users and policies, and the roles embedded in `instanceProfileXml` correctly pass `false` too. Nothing missed.

The asymmetry is handled correctly -- `tagsElement` emits nothing rather than an empty `<Tags/>` for an untagged resource, which matters: an empty element would itself read back as a configured-but-empty tag set to Terraform.

Test coverage is genuinely end-to-end, not just XML-shape assertions -- the Terraform compat suite includes the "second `terraform plan` reports no changes" assertion, which is the actual drift-class bug this fixes; an apply-only test structurally cannot catch it.

Ran the full IAM + Lambda suites fresh: 842/842 pass. Hygiene clean, no conflicts.

Merging this.

PR #2268, opened 2026-08-13
fix(lambda): propagate image inspection failures

### LINKED ISSUE #2230: [BUG] lambda: isLocalImagePresent swallows non-NotFound errors as "image absent", triggering needless pulls (ImageCacheService.java:127)
## What

`ImageCacheService.isLocalImagePresent` (`ImageCacheService.java:121-131`) catches every exception besides `NotFoundException` and returns `false` at debug log level. A transient daemon blip on `inspectImageCmd` — the exact `Broken pipe` class the codebase now retries at the transport — is silently reported as "image absent".

## Consequence

"The daemon could not tell me" and "the image is not there" collapse into the same answer, so a transient error triggers a needless (possibly slow, possibly credentialed) image pull. Violates the no-silent-fallbacks rule: data that shouldn't be missing is masked instead of surfacing.

## Possible direction

Let non-`NotFoundException` failures propagate (the transport retry has already spent its budget by then), or at minimum log at warn and distinguish the tri-state (present / absent / unknown) the way `ContainerLifecycleManager.presenceOf` does.

Severity: 4/10 — self-healing when the pull succeeds, but masks real daemon trouble and adds load exactly when the daemon is struggling.

moroshani:
## Summary

- treat Docker `NotFoundException` as the only signal that a local image is absent
- propagate other image-inspection failures instead of starting a misleading pull
- cover both the absent-image pull path and the inspection-error path with focused tests

## Root cause

`isLocalImagePresent` caught every non-`NotFoundException` and returned `false`. A Docker daemon or transport failure was therefore indistinguishable from an absent image, so `ensureImageExists` attempted a pull and obscured the original error.

## Validation

- `mvn -B -o test -Dtest=ImageCacheServiceTest`
- 11 tests passed on JDK 25, including a regression that failed before the production change
- `git diff --check`

Fixes #2230.

AI assistance: OpenAI Codex was used during implementation and testing. I reviewed the complete change and remain responsible for it.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ImageCacheService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/launcher/ImageCacheService.java
@@ -124,9 +124,6 @@ public class ImageCacheService {
             return true;
         } catch (com.github.dockerjava.api.exception.NotFoundException e) {
             return false;
-        } catch (Exception e) {
-            LOG.debugv("Could not check local image presence for {0}: {1}", imageUri, e.getMessage());
-            return false;
         }
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/lambda/launcher/ImageCacheServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/lambda/launcher/ImageCacheServiceTest.java
@@ -1,21 +1,67 @@
 package io.github.hectorvent.floci.services.lambda.launcher;
 
+import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.api.command.InspectImageCmd;
+import com.github.dockerjava.api.command.PullImageCmd;
+import com.github.dockerjava.api.command.PullImageResultCallback;
 import com.github.dockerjava.api.exception.DockerClientException;
 import com.github.dockerjava.api.exception.DockerException;
 import com.github.dockerjava.api.exception.InternalServerErrorException;
 import com.github.dockerjava.api.exception.NotFoundException;
 import com.github.dockerjava.api.exception.UnauthorizedException;
+import io.github.hectorvent.floci.config.EmulatorConfig;
 import org.junit.jupiter.api.Test;
 
+import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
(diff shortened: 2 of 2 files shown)


pgermosen (approved):
Welcome to Floci, and thank you — this is a genuinely good first contribution. Three lines removed, the right three, with a test that actually pins the behaviour. Approving on `7ae9e5a9`.

**The change is correct.** `NotFoundException` really is the only signal that an image is absent — Docker answers a missing image with a 404 on inspect — so narrowing the catch to it is the right shape rather than a smaller version of the right shape. I also checked how this lands against the existing retry logic: `isLocalImagePresent` is called *outside* `runWithRetry`, so a propagated inspection failure doesn't get fed into the pull-retry loop. Clean separation, no surprise interaction.

**I verified your red-green claim rather than taking it on trust.** In a scratch worktree I restored the pre-fix catch-all and re-ran the suite:

- on your branch: `ImageCacheServiceTest` **11/11 pass**
- with the catch-all restored: **1 failure**, `propagatesImageInspectionFailureWithoutPulling`

So the regression test does exactly what you said it does. All 15 CI checks are green.

## One correction to the issue, not to your PR

I want to put this on the record because you followed #2230's recommended direction faithfully, and one of the premises that direction rested on turns out not to hold.

The issue argued for propagation on the grounds that *"the transport retry has already spent its budget by then."* **There is no transport-level Docker retry in this codebase.** `DockerClientProducer` builds `ApacheDockerHttpClient` with timeouts only — no retry wrapper, no interceptor, and no reference to `Broken pipe` anywhere in `src/main`. The only retry we have is `runWithRetry`, and that wraps the pull, not the inspect.

So the practical effect is a little sharper than the issue suggests: a **single** transient daemon blip on inspect will now fail `ContainerLifecycleManager.create()` outright, where before it fell through to a pull that would often succeed and paper over it.

I still think this is the right trade-off, and I am approving on that basis — the old behaviour collapsed "the daemon could not tell me" into "the image is not there", which is precisely the masking the no-silent-fallbacks rule exists to prevent, and a daemon that cannot answer an inspect is unlikely to serve a pull either. But it is a real behaviour change rather than a pure clean-up, so it deserves to be written down as one.

Smaller thing in the same vein: the issue points at `ContainerLifecycleManager.presenceOf` as the tri-state precedent, and that method does not exist. The nearest analogue is `removeVolume`, whose javadoc already articulates this exact problem — *"a single boolean here can't always distinguish those two cases from the daemon's response alone."* Worth knowing if you or anyone else picks up the tri-state idea later; nothing to do here.

## One optional note on the test

Not blocking, and no need to act on it for this PR. When I ran `propagatesImageInspectionFailureWithoutPulling` against the pre-fix code, it failed with a `NullPointerException` — the unstubbed `pullImageCmd` returns null — rather than through `verify(dockerClient, never()).pullImageCmd(IMAGE)`, which is never reached. The test is genuinely red-green either way, but if this ever regresses, the signal a future maintainer sees would be a confusing NPE rather than "a pull was attempted when it shouldn't have been". Stubbing `pullImageCmd` in that test would let the `never()` verification be the thing that fires. Purely a legibility improvement.

PR #2278, opened 2026-08-13
fix(dynamodb): validate index access paths

### LINKED ISSUE #2276: [BUG] DynamoDB Query/Scan do not fully validate selected index access paths
## Service

DynamoDB

## AWS API actions

`Query`, `Scan`

## Summary

Floci supports GSI/LSI metadata and basic index queries, but request validation is not consistently driven by selected table/index key schema. Some requests that AWS rejects can succeed or operate on base-table data.

Sparse-index membership during `Scan` is already tracked separately in #2275.

## Current gaps

1. **`Scan` accepts nonexistent `IndexName`.** `DynamoDbService.scan(...)` continues over base-table items when no GSI/LSI matches. `applyIndexProjection(...)` also leaves items unchanged.
2. **`Query` incompletely validates `KeyConditionExpression`.** First condition is checked against selected partition key, but additional condition is evaluated without proving it targets selected sort key.
3. **Query filter key restrictions are not enforced.** Selected partition/sort key attributes belong in `KeyConditionExpression`, not `FilterExpression`.
4. **GSI projection restrictions are loose.** Explicit selection/projection can expose base-table attributes absent from a `KEYS_ONLY` or `INCLUDE` GSI.
5. **Validation should precede item traversal.** Empty and populated tables must reject same invalid requests.

## Expected behavior

Resolve one access path - base table, GSI, or LSI - before executing `Query` or `Scan`, then validate request against that schema using AWS-compatible errors.

## Acceptance criteria

- `Query` and `Scan` reject unknown `IndexName` with AWS-compatible `ValidationException`.
- `Query` requires selected partition key exactly once using equality.
- Optional key condition targets only selected sort key and uses AWS-supported operators.
- Non-key attributes are rejected in `KeyConditionExpression`.
- Selected key attributes are rejected in query `FilterExpression`.
- Legacy `KeyConditions` gets equivalent validation.
- GSI projection/`Select` rules reject non-projected attributes.
- Validation is identical for empty and populated tables.
- AWS SDK integration tests cover table, GSI, and LSI positive/negative cases.
- Exact error shape/message is compared with AWS where practical.
- Existing GSI `ConsistentRead=true` rejection remains intact.

## Non-goal

Do not reject a legal table `Scan` merely because a GSI might be faster. AWS permits table scans and cannot infer application intent. Performance linting would require a separate non-wire diagnostic mechanism.

## Relevant code

- `src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbJsonHandler.java`
- `src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java`

thomhurst:
## Summary

- Resolve one table/GSI/LSI access path before DynamoDB `Query` or `Scan` execution.
- Validate modern and legacy key conditions, query filter key usage, unknown indexes, and GSI projection/`Select` rules before item traversal.
- Add unit, Quarkus protocol integration, and AWS SDK for Java v2 compatibility coverage for table, GSI, and LSI behavior.

Closes #2276

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, invalid selected-index requests could succeed on empty tables or fall back to base-table behavior. Validation now follows the selected table/index key schema and projection metadata before traversing items. Index results also exclude items missing any selected index key attribute; for composite sort keys, every component must be present. Existing GSI `ConsistentRead=true` rejection remains covered.

Verified through the AWS SDK for Java v2 compatibility project against a packaged Floci instance.

## Testing

- `./mvnw.cmd test '-Dtest=DynamoDb*Test,ExpressionEvaluatorTest'` — 476 tests passed
- `../../mvnw.cmd test '-Dtest=DynamoDbAccessPathValidationTest'` from `compatibility-tests/sdk-test-java` — 9 tests passed
- `./mvnw.cmd package -DskipTests` — passed

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbAccessPathValidationTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbAccessPathValidationTest.java
@@ -1,0 +1,215 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.awssdk.services.dynamodb.model.AttributeDefinition;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import software.amazon.awssdk.services.dynamodb.model.BillingMode;
+import software.amazon.awssdk.services.dynamodb.model.ComparisonOperator;
+import software.amazon.awssdk.services.dynamodb.model.Condition;
+import software.amazon.awssdk.services.dynamodb.model.DynamoDbException;
+import software.amazon.awssdk.services.dynamodb.model.GlobalSecondaryIndex;
+import software.amazon.awssdk.services.dynamodb.model.KeySchemaElement;
+import software.amazon.awssdk.services.dynamodb.model.KeyType;
+import software.amazon.awssdk.services.dynamodb.model.LocalSecondaryIndex;
+import software.amazon.awssdk.services.dynamodb.model.Projection;
+import software.amazon.awssdk.services.dynamodb.model.ProjectionType;
+import software.amazon.awssdk.services.dynamodb.model.ScalarAttributeType;
+import software.amazon.awssdk.services.dynamodb.model.Select;
+
+import java.util.Map;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class DynamoDbAccessPathValidationTest {
+
+    private static final String TABLE = "access-path-" + UUID.randomUUID().toString().substring(0, 8);
+    private static DynamoDbClient ddb;
+
+    @BeforeAll
+    static void createTable() {
+        ddb = TestFixtures.dynamoDbClient();
+        ddb.createTable(request -> request
+                .tableName(TABLE)
+                .keySchema(key("pk", KeyType.HASH), key("sk", KeyType.RANGE))
+                .attributeDefinitions(
+                        attribute("pk"), attribute("sk"), attribute("status"),
+                        attribute("createdAt"), attribute("alternate"))
+                .globalSecondaryIndexes(GlobalSecondaryIndex.builder()
+                        .indexName("status-index")
+                        .keySchema(key("status", KeyType.HASH), key("createdAt", KeyType.RANGE))
+                        .projection(Projection.builder()
+                                .projectionType(ProjectionType.INCLUDE)
+                                .nonKeyAttributes("summary")
+                                .build())
+                        .build())
+                .localSecondaryIndexes(LocalSecondaryIndex.builder()
+                        .indexName("alternate-index")
+                        .keySchema(key("pk", KeyType.HASH), key("alternate", KeyType.RANGE))
+                        .projection(Projection.builder().projectionType(ProjectionType.KEYS_ONLY).build())
+                        .build())
+                .billingMode(BillingMode.PAY_PER_REQUEST));
+    }
+
+    @AfterAll
+    static void deleteTable() {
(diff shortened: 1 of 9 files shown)


pgermosen (commented):
@thomhurst — this is a well-built change, and filing #2276 with acceptance criteria that precise before writing the fix made it much easier to review against something rather than against my own guesses. The design is right: resolving one access path first and validating against *that* schema is what makes the empty-table and populated-table cases behave identically, which was the crux of the bug.

**The error messages are genuine.** I checked the distinctive ones rather than assuming, and they are the strings real callers hit against real DynamoDB — `Query condition missed key schema element: <pk>`, `Filter Expression can only contain non-primary key attributes: Primary key attribute: <name>`, `KeyConditionExpressions must only contain one condition per key`, and the `Select type ALL_ATTRIBUTES is not supported for global secondary index … because its projection type is not ALL` form. Getting these byte-exact matters more than usual here, because a caller matching on message text is the whole reason someone runs an emulator instead of a stub.

Locally on `32225b72`: 407 tests green across `DynamoDb*Test` and `ExpressionEvaluatorTest`. All 13 CI checks pass.

## One thing I would like called out before merge

The refactor consolidated the sparse-index filter from three sites into one, and changed its semantics on the way:

```java
// before — partition key and the FIRST sort key only
.filter(item -> item.has(finalPkName) && hasNonNullAttribute(item, finalPkName))
.filter(item -> finalSkName == null || (item.has(finalSkName) && hasNonNullAttribute(item, finalSkName)))

// after — every element of the index key schema
Set<String> indexKeys = accessPath.keyAttributeNames();
```

`keyAttributeNames()` returns all key schema elements, so for a **composite sort key** an item carrying the first sort-key attribute but missing the second was previously included in index results and is now excluded.

**I think the new behaviour is correct** — DynamoDB requires an item to carry every index key attribute before it appears in that index, so this is a genuine improvement and not something I want reverted. Two requests around it rather than about it:

1. **Mention it in the description.** Right now the PR reads as purely a validation change, and this is a results change: a query that returned three items yesterday can return two today. Someone bisecting that later will not think to look in a PR titled "validate index access paths".

2. **Pin it with a composite-sort-key test.** I went looking for existing coverage and want to be accurate about what I found: sparse-index behaviour *is* covered, in `sdk-test-awscli/test/dynamodb.bats` and `sdk-test-python/tests/test_dynamodb.py`, and both run in CI. But both use a single-sort-key GSI, which is precisely the case whose behaviour did **not** change. The composite case — the one that did — has nothing pinning it in either direction.

   I confirmed the gap by deleting the filter block outright and re-running `DynamoDb*Test`: everything still passed. That only tells us the Java unit suites do not reach it, since the compat suites need a running Floci, but it does mean a future refactor of this exact block has no fast local signal.

Neither is a large ask — the second is one more GSI with two RANGE elements and an item missing the trailing one.

Everything else checks out: the access-path abstraction reads cleanly, `projectedAttributeNames` correctly unions the base table key schema with the index keys before applying `INCLUDE`, and the net reduction in `DynamoDbService` is real simplification rather than logic moved somewhere harder to find. Happy to approve as soon as those two are addressed.

thomhurst:
Addressed both requests in `35b4d0ec`:

- Added `queryOnCompositeSortKeyGsiExcludesItemsMissingTrailingIndexKey`, covering a GSI with two `RANGE` elements and an item missing the trailing key. The test fails if the all-index-keys sparse filter is removed.
- Updated the PR’s AWS Compatibility section to call out the result-set change for composite sort keys.

`DynamoDb*Test` and `ExpressionEvaluatorTest`: 476 tests passed.

pgermosen (approved):
Both requests from the earlier review are genuinely resolved:

1. Description now explicitly calls out the results-changing behavior (composite sort-key items missing a trailing component are now excluded from index results).
2. `queryOnCompositeSortKeyGsiExcludesItemsMissingTrailingIndexKey` pins it — verified load-bearing by relaxing the completeness check to a single key attribute; the incomplete item leaks through and the test catches it exactly as intended.

Full dynamodb suite passes (360/360), both on the branch and on the actual merged-with-main tree (compiled + tested the merge). LGTM.

PR #2279, opened 2026-08-13
fix(appconfig): accept top-level resource ARNs in TagResource/ListTagsForResource

sheldon-nadarajah-sft:
## Bug

`AppConfigTagHandler#parseArn` hardcodes the assumption that every AppConfig resource ARN nests
under `application/...`:

```java
if (parts.length >= 2 && "application".equals(parts[0])) { ... }
throw new AwsException("BadRequestException", "Invalid resource ARN: " + arn, 400);
```

That's true for `environment` and `configurationprofile` ARNs (`application/{appId}/environment/{envId}`,
etc.), but AWS documents `deploymentstrategy`, `extension`, and `extensionassociation` as real,
taggable AppConfig resources that are **not** nested under `application/` - e.g.
`arn:aws:appconfig:{region}:{account}:deploymentstrategy/{id}`. Verified against AWS's own
CloudFormation resource docs before scoping this: `AWS::AppConfig::DeploymentStrategy`,
`AWS::AppConfig::Extension`, and `AWS::AppConfig::ExtensionAssociation` all have a `Tags`
property.

For a `deploymentstrategy/{id}` ARN, `parts[0]` is `"deploymentstrategy"`, not `"application"`, so
`TagResource`/`ListTagsForResource` reject it with a 400 - even though it's a valid, taggable
resource.

## Fix

Adds a `TOP_LEVEL_TYPES` set (`deploymentstrategy`, `extension`, `extensionassociation`) checked
before the `application/` branch. Tag storage for these types remains a no-op, matching the
existing precedent for `environment`/`configurationprofile` (accepted, but not yet persisted) -
this fixes the ARN-shape rejection, not tag persistence for new resource types.

Deliberately did **not** touch the existing `parts.length == 6` branch, which currently
over-accepts `hostedconfigurationversion` ARNs as taggable even though AWS doesn't support tagging
that resource type at all (confirmed via its CloudFormation docs - no `Tags` property). That's a
separate, pre-existing issue; keeping this PR scoped to the reported bug rather than bundling an
unrelated fix into a shared helper (lesson from #2155's review - a one-line fix to a shared helper
can resolve one AWS-compatibility gap while introducing another).

## Tests

- `AppConfigIntegrationTest`: added 3 new tests covering `deploymentstrategy`/`extension`/`extensionassociation`
  ARN acceptance (200/204, not 400) in the existing tag-lifecycle section.
- `compatibility-tests/sdk-test-java` `AppConfigTest`: added an SDK round-trip test
  (`tagDeploymentStrategyViaSdk`) using the real AWS SDK v2 client against a live instance.

Ran locally (JDK 25):
```
AppConfigIntegrationTest (unit):        25/25 passed
AppConfigTest (SDK compat, live):       15/15 passed
```

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppConfigTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppConfigTest.java
@@ -306,4 +306,32 @@ class AppConfigTest {
             } catch (Exception ignored) {}
         }
     }
+
+    @Test
+    @Order(51)
+    @DisplayName("TagResource / ListTagsForResource - deploymentstrategy ARN (top-level, not nested under application/)")
+    void tagDeploymentStrategyViaSdk() {
+        // deploymentstrategy/extension/extensionassociation ARNs don't nest under application/
+        // like every other AppConfig resource tested above - a naive ARN parser can end up
+        // hardcoding the application/ prefix and rejecting these as invalid, even though AWS
+        // documents all three as taggable (see e.g. AWS::AppConfig::DeploymentStrategy's Tags
+        // property). This asserts the SDK call succeeds end-to-end, not just that some ARN shape
+        // is accepted internally.
+        String tagStrategyId = appConfig.createDeploymentStrategy(CreateDeploymentStrategyRequest.builder()
+                .name("tag-roundtrip-strategy")
+                .deploymentDurationInMinutes(0)
+                .growthFactor(100f)
+                .finalBakeTimeInMinutes(0)
+                .build()).id();
+        String arn = "arn:aws:appconfig:us-east-1:000000000000:deploymentstrategy/" + tagStrategyId;
+
+        appConfig.tagResource(TagResourceRequest.builder()
+                .resourceArn(arn)
+                .tags(java.util.Map.of("env", "prod"))
+                .build());
+
+        ListTagsForResourceResponse listed = appConfig.listTagsForResource(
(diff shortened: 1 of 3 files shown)


sheldon-nadarajah-sft:
Addressed both Copilot findings in a follow-up commit:

- Clarified the class javadoc - it singled out `hostedconfigurationversion` as an example of something AWS doesn't support tagging for, but the existing `parts.length==6` branch already accepts (and no-ops) that exact ARN shape, which was misleading. Now describes current behavior generally instead.
- Renamed `tagExtensionAndExtensionAssociationArnsAreAcceptedNotRejected` → `listTagsForExtensionAndExtensionAssociationArnsIsAcceptedNotRejected` - the test only ever calls `GET`/`ListTagsForResource`, never `POST`/`TagResource`.

`AppConfigIntegrationTest`: 25/25 passed.

pgermosen (approved):
Thanks for the clean, well-scoped fix — nice catch on the top-level ARN gap.

Confirmed against AWS's own docs: `AWS::AppConfig::DeploymentStrategy` supports `Tags`, and `CreateDeploymentStrategy` is `POST /deploymentstrategies` — a top-level path, not nested under `application/...` — which lines up exactly with the bug you found. Good call leaving `hostedconfigurationversion` alone and keeping this scoped to the reported issue.

`AppConfigIntegrationTest` 25/25 pass, `docs-check` clean, no conflicts.

LGTM, approving. Welcome, and thanks for the contribution!

PR #2280, opened 2026-08-13
fix(appconfig): add DeleteConfigurationProfile/DeleteHostedConfigurationVersion/ListDeploymentStrategies/DeleteDeploymentStrategy

sheldon-nadarajah-sft:
**Stacked on #2279** (unrelated AppConfig ARN-validation fix) - this branch is built on top of it,
so this diff currently includes #2279's commit too. Please review/merge #2279 first; once it lands
this PR's diff will automatically narrow to just the changes below.

## Bug

`DeleteConfigurationProfile`, `DeleteHostedConfigurationVersion`, `ListDeploymentStrategies`, and
`DeleteDeploymentStrategy` have no route at all in `AppConfigController`. Since this app mounts
multiple services' JAX-RS resources at the shared root `@Path("/")`, an unmatched request falls
through to S3's generic path-style catch-all (`GET /{bucket}`, `DELETE /{bucket}/{key:.+}`), which
throws `NoSuchBucket` once it can't find a bucket by that name - so e.g. `DELETE
/deploymentstrategies/{id}` returns a raw S3-shaped 404 instead of any real AppConfig response.

Confirmed via both raw HTTP and the real AWS SDK for Java v2 against a live instance.

## Fix

Adds the missing routes + `AppConfigService` methods:

- `deleteConfigurationProfile`/`deleteHostedConfigurationVersion`/`deleteDeploymentStrategy` follow
  `deleteApplication`'s existing idempotent-delete convention (no upfront existence check - calling
  delete on an already-deleted or never-existing resource still succeeds, matching real AWS's
  idempotent `Delete*` semantics).
- `listDeploymentStrategies` includes the three predefined strategies (`AppConfig.AllAtOnce`, etc.)
  alongside custom ones - verified against the API reference's own sample response before
  implementing, since they're not in `strategyStore` at all (see `getDeploymentStrategy`'s existing
  `builtinStrategy` check).

**One real surprise along the way:** `DeleteDeploymentStrategy`'s actual path, per AWS's own API
reference, is `/deployementstrategies/{Id}` - an extra "e" ("deploy**e**mentstrategies"). Every
other deployment-strategy operation (`Create`/`Get`/`List`) correctly uses
`/deploymentstrategies`. This isn't a doc typo - the real AWS SDK for Java v2 sends the misspelled
path literally, and I hit the exact same 404 through the SDK before catching it. Matched AWS's
actual wire behavior rather than the "obviously correct" spelling, since that's what wire-protocol
compatibility means here.

Also regenerated `docs/services/appconfig.md`'s Supported Operations list - hand-maintained for
now, since REST controllers aren't yet covered by `tools/docs/regen_action_docs.py` (confirmed via
`services.yaml`'s own Phase-1 scope comment; `--strict` passes clean either way).

## Tests

- `AppConfigIntegrationTest`: 4 new tests (delete-then-404 for profile/version/strategy, idempotent
  double-delete, list includes builtins + custom).
- `compatibility-tests/sdk-test-java` `AppConfigTest`: 1 new test covering all four operations via
  the real SDK end-to-end, including the misspelled-path delete.

Ran locally (JDK 25):
```
AppConfigIntegrationTest (unit):        29/29 passed
AppConfigTest (SDK compat, live):       16/16 passed
Full suite (./mvnw test):               passed
```

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppConfigTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppConfigTest.java
@@ -334,4 +334,73 @@ class AppConfigTest {
                 ListTagsForResourceRequest.builder().resourceArn(arn).build());
         assertThat(listed.tags()).isNotNull();
     }
+
+    @Test
+    @Order(52)
+    @DisplayName("DeleteConfigurationProfile / DeleteHostedConfigurationVersion / ListDeploymentStrategies / DeleteDeploymentStrategy")
+    void deleteAndListOperationsPreviouslyReturnedNoSuchBucket() {
+        // These four operations had no route at all, so the request fell through to S3's
+        // generic path-style catch-all and returned a raw S3 NoSuchBucket 404 instead of a real
+        // AppConfig response. Exercised via the SDK (not raw HTTP) so a regression here surfaces
+        // as an SDK-level exception, matching how a real caller would hit it.
+        String delAppId = appConfig.createApplication(CreateApplicationRequest.builder()
+                .name("delete-ops-app")
+                .build()).id();
+        try {
+            String delProfileId = appConfig.createConfigurationProfile(CreateConfigurationProfileRequest.builder()
+                    .applicationId(delAppId)
+                    .name("delete-ops-profile")
+                    .locationUri("hosted")
+                    .type("AWS.Freeform")
+                    .build()).id();
+
+            appConfig.createHostedConfigurationVersion(CreateHostedConfigurationVersionRequest.builder()
+                    .applicationId(delAppId)
+                    .configurationProfileId(delProfileId)
+                    .contentType("application/json")
+                    .content(SdkBytes.fromUtf8String("{\"throwaway\": true}"))
+                    .build());
+
+            appConfig.deleteHostedConfigurationVersion(DeleteHostedConfigurationVersionRequest.builder()
+                    .applicationId(delAppId)
+                    .configurationProfileId(delProfileId)
+                    .versionNumber(1)
+                    .build());
+            assertThrows(ResourceNotFoundException.class, () -> appConfig.getHostedConfigurationVersion(
+                    GetHostedConfigurationVersionRequest.builder()
+                            .applicationId(delAppId).configurationProfileId(delProfileId).versionNumber(1).build()));
+
+            appConfig.deleteConfigurationProfile(DeleteConfigurationProfileRequest.builder()
+                    .applicationId(delAppId)
+                    .configurationProfileId(delProfileId)
+                    .build());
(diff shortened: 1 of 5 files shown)


sheldon-nadarajah-sft:
Addressed both real findings (Copilot + Greptile independently flagged the same profile-scoping bug) in a follow-up commit:

- `deleteConfigurationProfile` ignored `appId` entirely and deleted by the bare, globally-keyed `profileId` - confirmed this really did let a caller delete another application's profile just by knowing/guessing its ID, since the sibling read (`getConfigurationProfile`) already enforces this ownership check but the delete route didn't. Now rejects a mismatched app/profile pair with the same `ResourceNotFoundException` `getConfigurationProfile` already uses; a profile that doesn't exist anywhere is still an idempotent no-op, matching `deleteApplication`'s existing convention.
- Same method now also cascades the delete to that profile's hosted configuration versions, which aren't independently addressable outside the profile's own lifecycle - otherwise a caller could still fetch versions for a profile that's supposedly gone.
- `deleteDeploymentStrategy` on a predefined strategy (e.g. `AppConfig.AllAtOnce`) silently reported success without doing anything (builtins were never in `strategyStore`) - the strategy would still exist and still show up in every subsequent Get/List despite the "successful" delete. Now rejects with `BadRequestException`.
- Applied this project's own established cleanup-logging convention (`System.err.println` on a best-effort catch, matching several other `compatibility-tests/sdk-test-java` files) to the flagged silent catch instead of swallowing the exception.

On the pagination finding (`ListDeploymentStrategies` ignoring `MaxResults`/`NextToken`) - replied inline on #2282 (where Greptile actually left that specific comment): no `List*` operation anywhere in this codebase currently implements real pagination, so I left this consistent with that existing, codebase-wide convention rather than adding it just here.

Added regression tests for both fixes (cross-application delete rejection + original untouched, predefined-strategy delete rejection + still resolvable), at both the raw-HTTP and real-SDK level.

`AppConfigIntegrationTest`: 31/31 passed. SDK compat `AppConfigTest`: 18/18 passed.

pgermosen (approved):
Verified the four new routes (DeleteConfigurationProfile, DeleteHostedConfigurationVersion, ListDeploymentStrategies, DeleteDeploymentStrategy) against the real botocore AppConfig model — the error codes used (ResourceNotFoundException, BadRequestException) are documented valid errors for these operations, and the misspelled /deployementstrategies/{id} path is genuinely how AWS's own API model spells it, not a doc typo.

Both P1s from the bot reviews (cross-application profile deletion, predefined-strategy deletion reporting success) are already fixed on this branch, each with a regression test that confirms the resource is left untouched after rejection, not just that the request is rejected.

Ran AppConfigIntegrationTest locally: 31/31 pass. Clean merge against current main, docs-check passes, no CODEOWNERS-protected paths.

PR #2285, opened 2026-08-13
fix(s3vectors): route ListVectors instead of falling through to S3

### LINKED ISSUE #2161: [BUG] S3 Vectors: ListVectors is unrouted and fails with an S3 XML error
## Service

S3 Vectors

## AWS API Action

`ListVectors`

## Expected behavior

`ListVectors` returns the vectors stored in an index (keys, and optionally data and metadata), with pagination.

## Actual behavior

The call fails with an **Amazon S3** XML error rather than an S3 Vectors response or a clear "unsupported operation" error:

```xml
<Error>
  <Code>InvalidArgument</Code>
  <RequestId>...</RequestId>
</Error>
```

The cause looks structural rather than a bug inside the operation: `S3VectorsController` declares `@Path` routes for `CreateVectorBucket`, `GetVectorBucket`, `ListVectorBuckets`, `DeleteVectorBucket`, `CreateIndex`, `GetIndex`, `ListIndexes`, `DeleteIndex`, `PutVectors`, `GetVectors`, `DeleteVectors` and `QueryVectors` — but there is no `/ListVectors` route, so the request falls through to the S3 REST-XML controller and is interpreted as an S3 bucket POST.

The misleading error is worth fixing even independently of implementing the operation: from the client side it looks like an S3 problem, not a missing S3 Vectors action, which sends you looking in the wrong place.

This matters for applications that use `ListVectors` to decide whether an index needs seeding — a common bootstrap pattern, and one where the failure surfaces at startup.

## Reproduction

```bash
E=http://localhost:4566

  --vectors '[{"key":"v1","data":{"float32":[0.1,0.2,0.3,0.4]},"metadata":{"tenant":"a"}}]'

aws --endpoint-url $E s3vectors list-vectors --vector-bucket-name demo --index-name idx
```

A raw `POST http://localhost:4566/ListVectors` with the JSON body produces the same S3 XML error, so it is server-side routing rather than a CLI artefact.

For contrast, `GetVectors` and `QueryVectors` against the same index work correctly.

## Environment

- Client: AWS CLI 2.36.16 (also reproduces via boto3 and a raw HTTP POST)

---

Happy to send a PR adding the route and a `listVectors` service method over the existing per-index vector map (with `maxResults` / `nextToken` and the `returnData` / `returnMetadata` flags), modelled on the existing `ListIndexes` handler — just say if that would be useful.

Agboolafeyikemi:
Fixes #2161

## Problem

`ListVectors` had no route in `S3VectorsController`. Twelve other S3 Vectors operations were routed, but a `ListVectors` request fell through to the S3 REST-XML controller and came back as a misleading S3 error instead of a ListVectors response or a clear "unsupported operation" error:

```xml
<Error>
  <Code>InvalidArgument</Code>
  <RequestId>...</RequestId>
</Error>
```

Reproduced first by running the app locally and hitting the endpoint with the issue's exact repro (`CreateVectorBucket` → `CreateIndex` → `PutVectors` → `ListVectors`) — confirmed the same XML error against unfixed `main`.

## Fix

- `S3VectorsService.listVectors` — new method returning vectors sorted by key, with cursor-based pagination (`maxResults` / `nextToken`), modeled on `AcmService.listCertificates`'s base64-JSON-cursor pattern.
- `S3VectorsController` — added the missing `/ListVectors` route plus request/response records, following the `returnData` / `returnMetadata` shape already used by `GetVectors`.

## Test plan

- Reproduced first: added `listVectors` and `listVectorsPaginatesWithMaxResultsAndNextToken` to `S3VectorsIntegrationTest`; confirmed both failed against unfixed code with the same symptom as the issue (`400`, `application/xml` content type).
- Applied the fix, reran the same tests — both now return `200` with the expected vectors/pagination.
- `./mvnw test -Dtest=S3VectorsIntegrationTest` — 15/15 passing.
- Manually re-verified live against the running app: `ListVectors` with `maxResults=2` returns `v1, v2` + `nextToken`; following that token returns the remaining `v3` with no further token.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3vectors/S3VectorsController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3vectors/S3VectorsController.java
@@ -201,6 +201,24 @@ public class S3VectorsController {
     ) {}
 
     @RegisterForReflection
+    public record ListVectorsRequest(
+            String vectorBucketName,
+            String indexName,
+            String indexArn,
+            Integer maxResults,
+            String nextToken,
+            Boolean returnData,
+            Boolean returnMetadata
+    ) {}
+
+    @RegisterForReflection
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public record ListVectorsResponse(
+            List<VectorGetResponseRepresentation> vectors,
+            String nextToken
+    ) {}
+
+    @RegisterForReflection
     public record DeleteVectorsRequest(
             String vectorBucketName,
             String indexName,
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
All three prior greptile findings genuinely resolved on current HEAD: the quote-truncating cursor was replaced with plain Base64 encoding rather than patched around, `indexArn` is now correctly wired from the controller through to `resolveIndex`, and the ARN's own embedded region is used for cross-region index lookups instead of the request's region.

Cross-checked against the real botocore `s3vectors` model: request/response shapes, `maxResults` default of 500 and max of 1000, and the `vectorBucketName`+`indexName`-or-`indexArn` input contract all match exactly. The account-isolation design is also safe by construction — storage goes through `StorageFactory.create(...)` (an `AccountAwareStorageBackend`), so lookups are scoped to the authenticated caller's real account regardless of what account an ARN string claims.

`S3VectorsIntegrationTest` 18/18 pass, `docs-check` clean, no conflicts.

LGTM, approving.

PR #2286, opened 2026-08-13
fix(apigateway): honor configured execute-api hostname

### LINKED ISSUE #2284: [BUG] apigateway: execute-api host filter ignores the configured hostname, so the region-bearing form only routes under localhost
## What

`ApiGatewayExecuteApiHostFilter.EXECUTE_API_HOST` hardcodes the DNS tails it accepts:

```java
"^([a-z0-9-]+)\\.execute-api\\."
        + "(?:[a-z]{2}-[a-z-]+-\\d+\\.(?:localhost|amazonaws\\.com)"
        + "|localhost\\.(?:floci\\.io|localstack\\.cloud))$"
```

It never consults `config.hostname()`. `S3VirtualHostFilter`, solving the same virtual-host
problem, does:

```java
this.baseHostname = config.hostname()
        .orElseGet(() -> containerDetector.isRunningInContainer()
                ? EmbeddedDnsServer.DEFAULT_SUFFIX
                : extractHostnameFromUrl(config.baseUrl()));
```

## Consequence

The region-bearing execute-api host — the AWS-real shape,
`{apiId}.execute-api.{region}.{host}` — routes only when Floci is addressed as `localhost`.
Under any configured hostname it does not match, the filter declines, and the request falls
through to whatever else claims the path. For `@connections` that is the S3 path controller
(`/{bucket}/{key+}`), i.e. the original #1846 symptom.

This is not hypothetical: Floci's own Node compatibility suite runs with
`FLOCI_ENDPOINT=http://floci:4566` (`compatibility-tests/sdk-test-node/Dockerfile:10`), so the
region-bearing form cannot be exercised end-to-end there at all.

```
authority=localhost:4566  ->  abc123.execute-api.ap-northeast-2.localhost   match=YES
authority=floci:4566      ->  abc123.execute-api.ap-northeast-2.floci       match=NO
```

The regionless built-in suffixes (`localhost.floci.io`, `localhost.localstack.cloud`) are public
wildcard DNS pointing at loopback, so they resolve from anywhere and are unaffected — which is
why the end-to-end tests in #2188 use them.

## History

Pre-existing rather than introduced by #2188. Before that PR the regex accepted only
`…execute-api.localhost.(floci.io|localstack.cloud)`, so it had no configurable hostname either.
#2188 added the region-bearing branch and inherited the same limitation. The
`ExecuteApiSubdomainFilter` that #2188 removed substring-matched `.execute-api.` and accepted any
tail, so it happened to work under any hostname; consolidating onto the stricter shared regex
(which was the right call for other reasons) is what narrowed it.

## Possible direction

Take `config.hostname()` into the host grammar the way `S3VirtualHostFilter` does, keeping the
built-in suffixes and `amazonaws.com` as additional accepted tails. The grammar lives in one place
now (`ApiGatewayExecuteApiHostFilter.extractApiId`, shared with `WebSocketHandler`), so this is a
single-site change.

Worth covering with an end-to-end case under a non-`localhost` hostname, since that is exactly the
gap that hid this.

Severity: low-moderate — the default local setup works; anyone running Floci under a configured
hostname loses the AWS-real endpoint shape with a confusing `NoSuchBucket` as the symptom.

dixitrathod16:
## Summary

Closes #2284.

## Problem

`ApiGatewayExecuteApiHostFilter` only recognized region-bearing execute-api hosts ending in `localhost` or `amazonaws.com`. With `FLOCI_HOSTNAME=floci`, requests to `{apiId}.execute-api.{region}.floci` fell through to unrelated routing, including the S3 path controller for WebSocket `@connections` calls.

## Why it matters

Docker Compose and the compatibility pipeline configure Floci as `floci`. Without this fix, the AWS-shaped region-bearing execute-api endpoint cannot be used in that supported configuration and fails with a misleading response.

## Fix

The shared execute-api host parser now accepts the configured `floci.hostname` alongside the existing AWS, `localhost`, and built-in public suffix forms. HTTP execute-api routing and WebSocket upgrades share the same parser. Both application-scoped beans capture `EmulatorConfig.hostname()` once at runtime construction, following the existing `S3VirtualHostFilter` pattern and avoiding both native-image static-property capture and per-request config lookups.

## Tests

- Added a host-filter regression for `abc123.execute-api.ap-northeast-2.floci`.
- Added a Quarkus integration test with `floci.hostname=floci`.
- Added an AWS SDK for JavaScript v3 compatibility test that sends a signed region-bearing configured `Host` header through the CI Docker topology.
- Focused Java regressions (`ApiGatewayExecuteApiHostFilterTest`, `ApiGatewayExecuteApiHostIntegrationTest`, and `ApiGatewayV2WebSocketIntegrationTest`): **46 passed**.
- CI-shaped focused Node suite `apigatewayv2-execute-api-subdomain.test.ts`: **5 passed**.

## Manual verification

Ran the prescribed local Docker topology against the latest follow-up commit:

- Built `floci:test` with `docker/Dockerfile`.
- Started `floci` on `compat-net` with `FLOCI_HOSTNAME=floci`, TLS, Docker networking, and Lambda hot reload enabled.
- Built `compat-sdk-test-node` and ran `vitest run tests/apigatewayv2-execute-api-subdomain.test.ts` with `FLOCI_ENDPOINT=http://floci:4566` and the required `sdk-vhost-bucket.floci` host mapping.
- Result: **5 passed, 0 failed**.

## Native compatibility correction

The first native compatibility run exposed a startup failure when runtime `floci.hostname=floci` differed from a value injected during native static initialization. The initial revision removed constructor-time `@ConfigProperty` capture. The follow-up now uses the established runtime-initialized `EmulatorConfig` mapping once at application-bean construction, matching `S3VirtualHostFilter` without per-request lookup. The native compatibility matrix has been re-triggered for this revision.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature
- [ ] Breaking change
- [ ] Docs / chore

## AWS compatibility

Preserves existing `localhost`, `amazonaws.com`, and built-in suffix host forms while adding the configured-host form that corresponds to AWS's `{apiId}.execute-api.{region}.{host}` endpoint shape.

## Checklist

- [x] Regression coverage added
- [x] Focused Java and Node compatibility validation completed
- [x] Native runtime configuration startup path addressed
- [x] Commit follows Conventional Commits

### DIFF
--- a/compatibility-tests/sdk-test-node/tests/apigatewayv2-execute-api-subdomain.test.ts
+++ b/compatibility-tests/sdk-test-node/tests/apigatewayv2-execute-api-subdomain.test.ts
@@ -59,6 +59,12 @@ function builtinSuffixHost(apiId: string): string {
   return `${apiId}.execute-api.localhost.floci.io${port}`;
 }
 
+function configuredHostnameHost(apiId: string, region: string): string {
+  const endpoint = new URL(ENDPOINT);
+  const port = endpoint.port ? `:${endpoint.port}` : '';
+  return `${apiId}.execute-api.${region}.${endpoint.hostname}${port}`;
+}
+
 /**
  * A Management API client that presents an execute-api Host header while still connecting to the
  * real test endpoint. A build-step middleware overrides the `Host` header (before SigV4 signing)
@@ -181,6 +187,21 @@ describe('API Gateway v2 execute-api subdomain routing', () => {
 
     await expect(
       mgmt.send(new PostToConnectionCommand({ ConnectionId: 'does-not-exist', Data: Buffer.from('hi') }))
+    ).rejects.toMatchObject({ name: 'GoneException' });
+  });
+
+  it('routes @connections through the region-bearing configured hostname (not S3)', async () => {
+    const region = 'ap-northeast-2';
+    const stage = 'prod';
+    const apiId = await createWsApiWithStage('configured-host-conn', region, stage);
+    const host = configuredHostnameHost(apiId, region);
+    const mgmt = managementClientForHost(host, stage, region);
+
+    // The Node compatibility container connects to the Compose service at floci:4566, while
+    // FLOCI_HOSTNAME=floci. Present the AWS-shaped region-bearing virtual host through the SDK's
+    // signed Host header so this proves configured-host routing without wildcard DNS.
+    await expect(
+      mgmt.send(new GetConnectionCommand({ ConnectionId: 'does-not-exist' }))
     ).rejects.toMatchObject({ name: 'GoneException' });
   });
 
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteApiHostFilter.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteApiHostFilter.java
@@ -12,6 +12,7 @@ import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.UriBuilder;
 import jakarta.ws.rs.ext.Provider;
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.jboss.logging.Logger;
 
 import java.net.URI;
@@ -25,23 +26,15 @@ import java.util.regex.Pattern;
 public class ApiGatewayExecuteApiHostFilter implements ContainerRequestFilter {
 
     private static final Logger LOG = Logger.getLogger(ApiGatewayExecuteApiHostFilter.class);
-    // Accepts both the region-bearing AWS shape and Floci's built-in DNS suffixes:
-    //   {apiId}.execute-api.{region}.localhost[:port]        (local, region-bearing)
-    //   {apiId}.execute-api.{region}.amazonaws.com           (real AWS shape)
-    //   {apiId}.execute-api.localhost.floci.io               (built-in suffix, regionless)
-    //   {apiId}.execute-api.localhost.localstack.cloud       (built-in suffix, regionless)
-    // AWS's invoke URL carries the region in the hostname, so the region-bearing form is the
-    // primary target; the regionless built-in suffixes are kept for local convenience (region is
-    // then recovered from the SigV4 scope or a cross-region apiId lookup). Group 1 = apiId.
-    private static final Pattern EXECUTE_API_HOST = Pattern.compile(
-            "^([a-z0-9-]+)\\.execute-api\\."
-                    + "(?:[a-z]{2}-[a-z-]+-\\d+\\.(?:localhost|amazonaws\\.com)"
-                    + "|localhost\\.(?:floci\\.io|localstack\\.cloud))$",
+    private static final Pattern EXECUTE_API_PREFIX = Pattern.compile("^([a-z0-9-]+)\\.execute-api\\.(.+)$",
+            Pattern.CASE_INSENSITIVE);
+    private static final Pattern AWS_REGION = Pattern.compile("^[a-z]{2}-[a-z-]+-\\d+$",
             Pattern.CASE_INSENSITIVE);
 
     private final ApiGatewayLookup apiGatewayLookup;
     private final RegionResolver regionResolver;
     private final ApiGatewayExecuteRouteContext routeContext;
+    private final String baseHostname;
 
     @Inject
     public ApiGatewayExecuteApiHostFilter(ApiGatewayV2Service apiGatewayV2Service,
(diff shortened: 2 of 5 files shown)


dixitrathod16:
Follow-up for the prior native compatibility startup failure:

- Cause: `floci.hostname` was captured through constructor `@ConfigProperty` injection during native static initialization, then rejected when the container started with `FLOCI_HOSTNAME=floci`.
- Fix in `ca81b9dc`: HTTP and WebSocket execute-api routing now resolve the configured hostname at request time via MicroProfile Config.
- Validation: `ApiGatewayExecuteApiHostFilterTest`, `ApiGatewayExecuteApiHostIntegrationTest`, and `ApiGatewayV2WebSocketIntegrationTest` pass (46 tests).

The native compatibility matrix has been re-triggered on this single-commit revision.

pgermosen (commented):
@dixitrathod16 — thank you for turning this around so quickly, and for finding the native static-initialization problem yourself rather than letting it surface later. The fix is correct: I ran the host matrix and every previously-accepted form still resolves, `{apiId}.execute-api.{region}.{configuredHost}` now routes, and both production callers go through the config-aware path.

Two things I checked that are worth recording as evidence rather than opinion:

- **The region pattern is sound.** `^[a-z]{2}-[a-z-]+-\d+$` is the load-bearing new logic, so I tested it against every real AWS region in botocore's `endpoints.json` rather than a sample — **34 regions, 0 misses**, including `us-gov-east-1`, `cn-north-1` and `us-iso-east-1`.
- **The Node compat test restores what #2188 had to drop.** Deriving the host from `ENDPOINT`'s hostname is exactly the end-to-end coverage that got narrowed when the region-bearing form could not route under `floci`. Good to have that closed rather than left as a unit-test-only claim.

I have two nits below. Neither is a correctness problem — the behaviour is right as it stands — but I would like to settle the first before merging, since it sets a pattern.

## 1. Could `EmulatorConfig` do this instead of `ConfigProvider`?

`configuredHostname()` calls `ConfigProvider.getConfig().getOptionalValue(...)` on **every request**: line 90 evaluates the ternary before `extractApiId` runs, so the lookup happens for every request carrying a `Host` header, not just execute-api ones. The value cannot change at runtime, so that work is repeated for nothing.

More to the point, this is the codebase's first programmatic config lookup — `ConfigProvider.getConfig()` appears **zero** times in `main` under `src/main/java`. Everything else injects `EmulatorConfig`, which is `@ConfigMapping(prefix = "floci")` and already exposes `Optional<String> hostname()`.

`S3VirtualHostFilter` solves this identical problem the established way — same `@Provider @PreMatching` shape, resolving the hostname once in its constructor:

```java
this.baseHostname = config.hostname()
        .orElseGet(() -> containerDetector.isRunningInContainer()
                ? EmbeddedDnsServer.DEFAULT_SUFFIX
                : extractHostnameFromUrl(config.baseUrl()));
```

That runs under FLOCI_HOSTNAME in native today, which suggests the native hazard is specific to `@ConfigProperty` capture rather than to constructor-time config reads in general — `@ConfigMapping` interfaces are runtime-initialized in Quarkus. One structural difference I noticed while comparing the two:

| | CDI scope |
|---|---|
| `S3VirtualHostFilter` | `@ApplicationScoped` |
| `ApiGatewayExecuteApiHostFilter` | *(none)* |

I want to be careful not to overstate this: **I have not proven the `EmulatorConfig` route works in native here** — that needs a native build I have not run, and you have the failing evidence I do not. So this is a question rather than a request for a specific change. If injecting `EmulatorConfig` (possibly with `@ApplicationScoped` added) works, it would follow the existing pattern and drop the per-request lookup. If you tried that and it still failed natively, then the current approach is justified and I would just ask for a one-line comment saying so, so the next person does not "clean it up" back into a broken state.

Failing both of those, memoizing the resolved value into a `volatile` field on first *request* would keep the native fix while making the lookup once rather than per request.

## 2. `extractApiId(String)` is now the more discoverable of two overloads

To be clear up front: **this is not a behaviour change.** I compared the accepted host sets and the single-argument overload accepts exactly what `main`'s regex accepted, so nothing regressed.

The nit is the shape. There are now two `public static` entry points with different semantics, and the shorter, more obvious one silently means "localhost only":

```java
public static String extractApiId(String host)                    // → localhost only
public static String extractApiIdForConfiguredHostname(String host)
```

Both current production callers correctly use the config-aware path, so nothing is wrong today. But a future caller reaching for the short name gets exactly the bug this PR fixes, without any signal. Making the single-argument overload package-private (its only remaining callers are tests), or renaming it to something like `extractApiIdForDefaultHostname`, would remove the trap. Entirely your call on which.

Marking `waiting-contributor` while the first one is settled. Thank you again for the speed here — this was filed and fixed inside two hours, and the WebSocket handler now sharing the same parser is a better outcome than the gap I originally reported.

dixitrathod16:
@pgermosen Thank you — both nits are addressed in follow-up commit `0c07f32b` (kept separate from the original fix for reviewability).

1. **Fixed:** `ApiGatewayExecuteApiHostFilter` is now `@ApplicationScoped` and receives `EmulatorConfig`; it resolves `hostname()` once in the constructor. `WebSocketHandler` follows the same pattern, so both production callers pass their cached configured hostname to the shared parser. This replaces the per-request `ConfigProvider` lookup while following `S3VirtualHostFilter`. The configured-host Quarkus integration test still passes under `floci.hostname=floci` (and is part of the 46-test focused run).

2. **Fixed:** removed the one-argument `extractApiId(String)` overload entirely. The only remaining public entry point requires an explicit base hostname, so a future caller cannot accidentally select localhost-only behavior.

Validation: `ApiGatewayExecuteApiHostFilterTest`, `ApiGatewayExecuteApiHostIntegrationTest`, and `ApiGatewayV2WebSocketIntegrationTest` — **46 passed, 0 failures**. The new CI run will also revalidate the native-image path with this established config-mapping pattern.

pgermosen (approved):
Thank you for the quick, thorough turnaround. Verified both asks directly rather than taking the commit message at its word.

The filter now injects `EmulatorConfig` and captures the hostname once at construction, matching `S3VirtualHostFilter`'s established pattern exactly -- and I checked the actual CI runs rather than the PR description: every native check is green, including the full native compatibility matrix (compat-cdk, compat-opentofu, compat-terraform, sdk-test-awscli/go/java/node/python). That settles the native-viability question I'd raised.

The `extractApiId` overload ambiguity is also resolved, and cleanly -- collapsing to a single method with an explicit `baseHostname` parameter removes the trap better than the rename I'd suggested.

Ran the focused suite fresh: 46/46 pass. No conflicts, hygiene clean.

Approving.

PR #2288, opened 2026-08-14
fix(test): pass the managed prefix list store in Ec2PublicIpOnLaunchTest

### LINKED ISSUE #2287: [BUG] main does not compile: Ec2PublicIpOnLaunchTest calls a stale Ec2Service constructor
## Describe the bug

`main` does not compile. `Ec2PublicIpOnLaunchTest` calls the package-private `Ec2Service` constructor with the parameter list it had before #2105 added the managed prefix list store, so the test source fails to build and every job that compiles tests fails with it.

The test arrived in #2023, which was written before #2105 landed and merged after it without a rebase. Its call goes straight from `ec2-network-acls.json` to `ec2-tags.json`, missing `ec2-managed-prefix-lists.json` between them.

Flagging that #2105 is mine — the parameter that moved is the one I added, so this is my change colliding with an in-flight PR rather than anything wrong with #2023's own work.

## To Reproduce

```bash
git checkout main            # 5657e350
./mvnw -B test-compile
```

```
src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java:[142,16]
  no suitable constructor found for Ec2Service(...)
  (actual and formal argument lists differ in length)
```

## Expected behaviour

`main` compiles and CI passes on it.

## Actual behaviour

Compilation fails. This is not local to my environment — CI on `main` reproduces it:

- `CI` run [31767114963](https://github.com/floci-io/floci/actions/runs/31767114963) failed at 03:32 with the same file and line
- the `Nightly` run at 04:31 failed as well

## Fix

One argument in the test's constructor call:

```diff
                 load(dir, "ec2-network-acls.json", new TypeReference<Map<String, NetworkAcl>>() {}),
+                load(dir, "ec2-managed-prefix-lists.json", new TypeReference<Map<String, ManagedPrefixList>>() {}),
                 load(dir, "ec2-tags.json", new TypeReference<Map<String, List<Tag>>>() {}));
```

PR to follow.

## Note

This is the second time a change to a widely-called constructor in this repo has broken a caller that was in flight — @hampsterx caught the same shape on #2106, where `IamServiceTestHelper` resolved a constructor reflectively and broke at run time instead. Worth considering whether these long package-private constructors want a builder, so an added store cannot break callers that never mention it.

ck-michals:
Closes #2287.

`main` does not compile: `Ec2PublicIpOnLaunchTest` calls the package-private `Ec2Service` constructor with the parameter list it had before #2105 added the managed prefix list store. The call goes straight from `ec2-network-acls.json` to `ec2-tags.json`, missing `ec2-managed-prefix-lists.json` between them.

`#2023` was written before #2105 landed and merged after it without a rebase. Worth saying plainly that #2105 is mine — the parameter that moved is the one I added, so this is my change colliding with an in-flight PR rather than a defect in #2023's own work.

## Verified

CI on `main` reproduces the failure independently of my machine — [run 31767114963](https://github.com/floci-io/floci/actions/runs/31767114963) at 03:32, same file and line, and the `Nightly` run failed with it at 04:31.

With this one argument added, `main` compiles and `Ec2PublicIpOnLaunchTest` passes 2/2. The full EC2 suite is green at 331 tests, so #2023's behaviour is intact — it only ever needed the extra store.

## Note

This is the second time an added store has broken a caller that was in flight: @hampsterx caught the same shape on #2106, where `IamServiceTestHelper` resolves the constructor reflectively and so broke at run time rather than compile time. If these long package-private constructors became a builder, an added store could not break callers that never mention it. Happy to open that separately if it is wanted.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
@@ -10,6 +10,7 @@ import io.github.hectorvent.floci.services.ec2.model.InternetGateway;
 import io.github.hectorvent.floci.services.ec2.model.KeyPair;
 import io.github.hectorvent.floci.services.ec2.model.LaunchTemplate;
 import io.github.hectorvent.floci.services.ec2.model.NatGateway;
+import io.github.hectorvent.floci.services.ec2.model.ManagedPrefixList;
 import io.github.hectorvent.floci.services.ec2.model.NetworkAcl;
 import io.github.hectorvent.floci.services.ec2.model.Instance;
 import io.github.hectorvent.floci.services.ec2.model.Reservation;
@@ -159,6 +160,7 @@ class Ec2PublicIpOnLaunchTest {
                 load(dir, "ec2-nat-gateways.json", new TypeReference<Map<String, NatGateway>>() {}),
                 load(dir, "ec2-spot-instance-requests.json", new TypeReference<Map<String, SpotInstanceRequest>>() {}),
                 load(dir, "ec2-network-acls.json", new TypeReference<Map<String, NetworkAcl>>() {}),
+                load(dir, "ec2-managed-prefix-lists.json", new TypeReference<Map<String, ManagedPrefixList>>() {}),
                 load(dir, "ec2-tags.json", new TypeReference<Map<String, List<Tag>>>() {}));
     }
 


pgermosen (approved):
Approving — and this should go in ahead of anything else in the queue, since `main` is red until it does.

I reproduced the failure and the fix locally rather than taking CI's word for it.

## `main` is broken, and this is the only thing breaking it

At `5657e350`, `./mvnw -B test-compile` fails with exactly the error you quoted, at exactly the line you named:

```
  (actual and formal argument lists differ in length)
```

Worth recording that this was the **only** compilation error in the run — one `COMPILATION ERROR` block, one file. So this single argument is the whole gap between red and green, with nothing hiding behind it.

The parameter position is right. The package-private constructor declares `managedPrefixLists` between `networkAcls` and `tags`, and since both neighbours are `StorageBackend<String, ?>`, a wrong slot here would have compiled and then failed strangely at run time. It's in the correct one.

With your commit applied: `test-compile` clean, `Ec2PublicIpOnLaunchTest` 2/2, and the full EC2 + EKS surface **429 tests, 0 failures**. #2023's behaviour is intact — it only ever needed the extra store, as you said.

## Your account of how it happened is accurate

I checked the timeline rather than assuming it:

- #2105 (yours, added the store) merged **2026-08-11**
- #2023 (added the test) merged **2026-08-14 02:52**, three days later, without a rebase
- CI run [31767114963](https://github.com/floci-io/floci/actions/runs/31767114963) — branch `main`, sha `5657e350`, conclusion `failure`

So the collision is exactly as described, and nobody did anything wrong: #2023's test was written against the constructor as it stood when the work started. Thank you for flagging that #2105 is yours up front and for filing the issue with a reproduction instead of quietly pushing a fix — that's the version that leaves a trail for the next person.

## On the builder suggestion — the abstraction already exists

This is the part I'd like to redirect slightly, because I think you're closer to a fix than the note implies.

`StorageFactory` is already the repo-wide convention for this, and it already solves the problem. `Ec2Service` has a `StorageFactory` constructor alongside the enumerating one, and test doubles extending `StorageFactory` appear in roughly twenty test files (`InMemoryStorageFactory`, `SharedStorageFactory`, `FileStorageFactory`, `SingleUseStorageFactory`).

That's why only two of the five `new Ec2Service(...)` callers were exposed. `Ec2ServiceTest`, `Ec2ServiceConcurrencyTest` and `EksServiceTest` all pass `new InMemoryStorageFactory()` and were structurally immune — an added store cannot reach them. `Ec2PublicIpOnLaunchTest` and `Ec2ServicePersistenceTest` enumerate, and one of them broke.

And `Ec2PublicIpOnLaunchTest` didn't need to enumerate. It takes a `@TempDir` and builds `PersistentStorage` over it, but neither test restarts a service or asserts anything about persistence — they create a service, run instances, and assert flags. It could take `InMemoryStorageFactory` (same package as `Ec2ServiceTest`, so the existing one is right there) and would then have been immune to #2105 by construction.

So rather than a new builder, the cheaper change is moving the last enumerating callers onto the factory. `Ec2ServicePersistenceTest` genuinely does need disk — it restarts a service to prove state survives — but `FileStorageFactory` in `LambdaMicrovmsServicePersistenceTest` and the various `SharedStorageFactory`s are precedents for exactly that shape.

Please do open that as a separate issue — you've now hit this twice, and @hampsterx's reflective case on #2106 is the same root cause. It's worth doing. It just shouldn't hold up a red `main`.

Merging this as-is.

<sub>One cosmetic thing, explicitly **not** worth a force-push: the new import sits after `NatGateway` rather than before it. That block is already unsorted on `main` (`Instance` follows `NetworkAcl`), no checkstyle rule covers it, and CI is green — mentioning it only so it isn't mistaken for something I missed.</sub>

PR #2289, opened 2026-08-14
test(ec2): use storage factory in public IP regression

abanna:
## Summary

- follow up on #2288 by moving `Ec2PublicIpOnLaunchTest` to the existing `StorageFactory` constructor
- replace the fixture's enumerated persistent backends with an account-aware in-memory factory
- keep the public/private subnet and launch-time override regression scenarios unchanged

## Why this remains useful after #2288

#2288 restored `main` with the minimal missing `ManagedPrefixList` store argument. Its review identified that this non-persistence test did not need to enumerate every EC2 backend and explicitly recommended the existing `StorageFactory` pattern as a separate follow-up.

This PR is that follow-up. New EC2 stores are now created centrally by `Ec2Service`, so adding a backend no longer requires an unrelated edit to `Ec2PublicIpOnLaunchTest`.

## AWS compatibility and scope

This changes test wiring only. It does not change production request parsing, response shapes, endpoints, public-IP behavior, or persistence semantics. Existing managed-prefix-list API, Query-protocol, and persistence coverage remains unchanged.

No new assertion is needed: the two existing tests already cover public/private subnet behavior and AWS launch-time override precedence. This change preserves those regressions while making their fixture resilient to future EC2 storage additions.

## Validation

- rebased tree matches the previously fully tested tree exactly (`8245a1ed652d83e5ff24967a4cfd8a410f783d53`)
- `./mvnw test -B -Dtest=Ec2PublicIpOnLaunchTest` — 2 tests passed after the rebase
- previous full Maven run on the identical tree — 9,778 tests passed, 0 failures, 0 errors, 9 skipped
- `make docs-check` — passed
- `git diff --check` — passed

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2PublicIpOnLaunchTest.java
@@ -2,33 +2,15 @@ package io.github.hectorvent.floci.services.ec2;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import io.github.hectorvent.floci.config.EmulatorConfig;
-import io.github.hectorvent.floci.core.storage.PersistentStorage;
-import io.github.hectorvent.floci.core.storage.StorageBackend;
-import io.github.hectorvent.floci.services.ec2.model.Address;
-import io.github.hectorvent.floci.services.ec2.model.Image;
-import io.github.hectorvent.floci.services.ec2.model.InternetGateway;
-import io.github.hectorvent.floci.services.ec2.model.KeyPair;
-import io.github.hectorvent.floci.services.ec2.model.LaunchTemplate;
-import io.github.hectorvent.floci.services.ec2.model.NatGateway;
-import io.github.hectorvent.floci.services.ec2.model.ManagedPrefixList;
-import io.github.hectorvent.floci.services.ec2.model.NetworkAcl;
+import io.github.hectorvent.floci.core.storage.AccountAwareStorageBackend;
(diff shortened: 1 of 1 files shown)


abanna:
Follow-up: I tightened the fix so it addresses the fixture design that allowed this regression, rather than only appending the new `ManagedPrefixList` argument.

`Ec2PublicIpOnLaunchTest` is unrelated to persistence but directly constructed `Ec2Service` with every storage backend. That made each new EC2 store a required edit in this test, which is why the combination of #2105 and the stale fixture from #2023 produced the `testCompile` failure.

The test now uses the service's `StorageFactory` constructor with an account-aware in-memory factory, matching the existing EC2 unit-test pattern. The two public-IP regression scenarios are unchanged, while future storage additions are created centrally by `Ec2Service`.

Validation:

- `Ec2PublicIpOnLaunchTest`: 2/2 passed
- full Maven suite: 9,778 passed, 0 failures/errors (9 skipped)
- `make docs-check` and `git diff --check`: passed
- existing managed-prefix-list API, Query-protocol, and persistence coverage remains unchanged

No production request parsing, response shape, endpoint, or persistence behavior changed.

pgermosen (approved):
Good follow-up — confirmed the production `Ec2Service(..., StorageFactory)` constructor genuinely delegates to `storageFactory.create("ec2", ...)` per store internally, so this test's new `InMemoryStorageFactory` override now tracks that single centralized constructor instead of enumerating all 17 backends by hand. That's exactly what closes the #2288 failure mode for good: a future added store no longer requires touching this file at all.

`Ec2PublicIpOnLaunchTest` 2/2 pass, `docs-check` clean, no conflicts.

LGTM, approving.

PR #2314, opened 2026-08-14
feat(firehose): honor BufferingHints with time and size based flush

### LINKED ISSUE #1852: [FEAT] Support more control over firehose message flushing
## Service
 
Firehose
 
## API Action / Feature
 
putRecord, putRecordBatch
 
## AWS Documentation
 
(https://docs.aws.amazon.com/firehose/latest/APIReference/API_CreateDeliveryStream.html#API_CreateDeliveryStream_RequestSyntax)
 
## Why is this needed?
 
When using floci for application testing, the current behavior of flushing after 5 messages is insufficient [1]. In application testing you typically exercise your 'application under test' to send a known number of messages and then assert they are actually sent. The hardcoded number needlessly complicates this by requiring me to generate a multiple of 5 messages and afterwards to filter out those dummy extra messages before asserting on their existence and contents.
 
Either actual support for the 'BufferingHints' or a means to force a flush are needed.
 
[1] https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/firehose/FirehoseService.java#L28

- [ ] Yes
- [x ] No

### LINKED ISSUE #1886: [FEAT] [Firehose] Implement S3 buffered delivery for PutRecord / PutRecordBatch

mihaiblidaru:
## Summary

Firehose delivery streams currently buffer records and only deliver to S3 once 5 records accumulate (`DEFAULT_FLUSH_COUNT`, hardcoded). Anything below that threshold never reaches the bucket: the delivery pipeline silently drops slow-trickle traffic, records are lost on shutdown, and `BufferingHints` is validated and echoed but never honored.

This PR makes buffered delivery behave like AWS:

- **Size trigger**: a stream's buffer is delivered when the buffered bytes reach its `BufferingHints.SizeInMBs` (default 5 MiB), checked inline on `PutRecord`/`PutRecordBatch`.
- **Interval trigger**: a single-thread daemon flusher (started on `StartupEvent`, following the `ScheduleDispatcher` pattern) delivers any buffer whose `BufferingHints.IntervalInSeconds` (default 300 s) has elapsed since its first buffered record. Tick cadence is configurable via `floci.services.firehose.tick-interval-seconds` (default 10 s).
- **Opt-in record-count trigger** for local dev (suggested by @yoshinori-aono-om in #1886, co-authored accordingly): `floci.services.firehose.flush-record-count`, **disabled by default (0)** so out-of-the-box behavior stays AWS-faithful; `1` restores LocalStack-style record-at-a-time delivery for interactive flows whose stream definitions mirror production intervals.
- **Shutdown drain**: pending buffers are flushed on `ShutdownDelayInitiatedEvent`, which is guaranteed to run before any `ShutdownEvent` observer — so the drained objects are still persisted to disk by `EmulatorLifecycle`'s storage flush in `hybrid` mode.
- **DeleteDeliveryStream keeps discarding pending records** — verified against real AWS (eu-west-1): 3 records buffered under a 300s/5MB hint never reached the bucket after deletion. A unit test pins this.

Closes #1886, closes #1852.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

Incorrect behavior: records were only delivered on the 5-record count (matching neither AWS nor LocalStack); `BufferingHints` was stored but ignored, and sub-threshold records were dropped silently (see #1886 for a real-world report).

Delete semantics and buffering behavior were verified empirically against real AWS with AWS CLI 1.44.38 (botocore 1.42.48). The compatibility suites (`sdk-test-java`, `sdk-test-python`, `sdk-test-go`, `sdk-test-awscli`, `sdk-test-node`) were run locally against the JVM Docker image; SES→Firehose delivery now waits for the buffering interval like a real AWS client would.

## Checklist

- [x] New or updated integration test added (`FirehoseFlushIntegrationTest` covers the size and interval triggers through the real background flusher; `FirehoseFlushRecordCountIntegrationTest` covers the opt-in knob's config binding at the wire level)

**Note on missing automated coverage**: the shutdown drain cannot be exercised from `@QuarkusTest` (the application cannot be restarted mid-suite). It was verified manually end-to-end with `hybrid` storage: put a record, SIGTERM, drained object present on disk after restart.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DataLakeTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DataLakeTest.java
@@ -9,6 +9,9 @@ import software.amazon.awssdk.services.firehose.model.PutRecordRequest;
 import software.amazon.awssdk.services.firehose.model.Record;
 import software.amazon.awssdk.services.glue.GlueClient;
 import software.amazon.awssdk.services.glue.model.*;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.ListObjectsV2Request;
+import software.amazon.awssdk.services.s3.model.NoSuchBucketException;
 
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
@@ -26,6 +29,7 @@ class DataLakeTest {
     private static AthenaClient athena;
     private static GlueClient glue;
     private static FirehoseClient firehose;
+    private static S3Client s3;
 
     private static final String DB_NAME = TestFixtures.uniqueName("test_db");
     private static final String TABLE_NAME = "orders";
@@ -36,6 +40,7 @@ class DataLakeTest {
         athena = TestFixtures.athenaClient();
         glue = TestFixtures.glueClient();
         firehose = TestFixtures.firehoseClient();
+        s3 = TestFixtures.s3Client();
     }
 
     @Test
@@ -76,6 +81,12 @@ class DataLakeTest {
                         .bucketARN("arn:aws:s3:::floci-firehose-results")
                         .roleARN("arn:aws:iam::000000000000:role/datalake-firehose-role")
                         .prefix(STREAM_NAME + "/")
+                        // Floci does not enforce AWS's 60s minimum interval, which keeps
+                        // the delivery wait short (would need 60+ against real AWS).
+                        .bufferingHints(software.amazon.awssdk.services.firehose.model.BufferingHints.builder()
+                                .sizeInMBs(1)
+                                .intervalInSeconds(5)
+                                .build())
                         .build())
                 .build());
     }
@@ -91,6 +102,28 @@ class DataLakeTest {
                     .record(Record.builder().data(SdkBytes.fromString(json, StandardCharsets.UTF_8)).build())
                     .build());
         }
+
+        // Small records stay buffered until the stream's IntervalInSeconds
+        // elapses; wait for Firehose to deliver before querying, the same way a
+        // real AWS client would.
+        long deadline = System.currentTimeMillis() + 30_000;
+        boolean delivered = false;
+        while (!delivered && System.currentTimeMillis() < deadline) {
+            try {
+                delivered = !s3.listObjectsV2(ListObjectsV2Request.builder()
+                        .bucket("floci-firehose-results")
+                        .prefix(STREAM_NAME + "/")
+                        .build()).contents().isEmpty();
+            } catch (NoSuchBucketException e) {
+                // The bucket itself is only created on the first delivery.
+            }
+            if (!delivered) {
+                Thread.sleep(2_000);
+            }
+        }
+        assertThat(delivered)
+                .as("Firehose should deliver the buffered records within IntervalInSeconds")
+                .isTrue();
 
         // Athena Query
         StartQueryExecutionResponse startResp = athena.startQueryExecution(StartQueryExecutionRequest.builder()
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SesEventPublishingTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SesEventPublishingTest.java
@@ -200,10 +200,11 @@ class SesEventPublishingTest {
                                 .bucketARN("arn:aws:s3:::" + firehoseBucket)
                                 .roleARN("arn:aws:iam::000000000000:role/sdk-evt-fh-role")
                                 .prefix(firehoseStreamName + "/")
-                                // AWS minimum interval, so the delivery test waits as little as possible.
+                                // Floci does not enforce AWS's 60s minimum interval, which keeps
+                                // the delivery wait short (would need 60+ against real AWS).
                                 .bufferingHints(BufferingHints.builder()
                                         .sizeInMBs(1)
-                                        .intervalInSeconds(60)
+                                        .intervalInSeconds(5)
                                         .build())
                                 .build())
                         .build())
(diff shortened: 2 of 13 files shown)


hectorvent:
Thank you, and this is a well-evidenced change.

The defaults are right: `BufferingHints` is `SizeInMBs` 1 to 128 and `IntervalInSeconds` 0 to 900 in the model, and 5 MB / 300 s are AWS's own defaults. The bug description is precise too, a hardcoded flush count of 5 meant slow-trickle traffic never reached the bucket at all.

The detail I want to credit is the comment on `deleteDeliveryStream`: pending records are discarded rather than flushed, verified against real AWS with a date and a region. That is the one kind of evidence a local model cannot supply, and it is the difference between an emulator that guesses and one that matches.

Config wiring is complete, `EmulatorConfig`, both `application.yml` files, and both docs pages.

No blockers from my side.

hectorvent (approved):

PR #2315, opened 2026-08-14
fix(compat): install AWS CLI v2 in the compat image so AWS_ENDPOINT_URL is honored

### LINKED ISSUE #2267: [BUG] Ready Initialization Hook fails to create SQS queues
I have the following `compose.yml` file:

```
services:
  floci:
    image: floci/floci:latest-compat
    ports:
      - "4566:4566"
    volumes:
      - ./floci-init/ready.d:/etc/floci/init/ready.d:ro
    environment:
      AWS_DEFAULT_REGION: us-east-1
```

Here is the `floci-init/ready.d/01-seed.sh` file:

```
#!/bin/sh
set -eu

aws s3 mb s3://test-bucket
echo "List S3 buckets"
aws s3 ls

aws sqs create-queue --queue-name testqueue
```

When I start the container using `docker compose up`, it is failing with the following error:

```shell
An error occurred (InvalidClientTokenId) when calling the CreateQueue operation: The security token included in the request is invalid.

2026-08-13 04:21:39,041 ERROR [io.git.hec.flo.lif.EmulatorLifecycle] Startup hook execution failed — shutting down: java.lang.IllegalStateException: Hook script failed: 01-seed.sh exited with code 255
	at io.github.hectorvent.floci.lifecycle.inithook.HookScriptExecutor.run(HookScriptExecutor.java:40)
	at io.github.hectorvent.floci.lifecycle.inithook.HookScriptExecutor.run(HookScriptExecutor.java:33)
	at io.github.hectorvent.floci.lifecycle.inithook.HookScriptExecutor.run(HookScriptExecutor.java:24)
	at io.github.hectorvent.floci.lifecycle.inithook.HookScriptExecutor_ClientProxy.run(Unknown Source)
	at io.github.hectorvent.floci.lifecycle.inithook.InitializationHooksRunner.run(InitializationHooksRunner.java:47)
	at io.github.hectorvent.floci.lifecycle.inithook.InitializationHooksRunner_ClientProxy.run(Unknown Source)
	at io.github.hectorvent.floci.lifecycle.EmulatorLifecycle.onHttpStart(EmulatorLifecycle.java:166)
	at io.github.hectorvent.floci.lifecycle.EmulatorLifecycle_Observer_onHttpStart_BADQZDUNaOPGhAtIdg2KAwN8FGA.notify(Unknown Source)
	at io.quarkus.arc.impl.EventImpl$Notifier.notifyObservers(EventImpl.java:366)
```

**If we comment-out SQS queue creation command in `01-seed.sh` then it is working fine (S3 bucket is created successfully).**

## Reproduction

```bash
git clone https://github.com/siva-attic/floci-sqs-issue
cd floci-sqs-issue
docker compose up
```

## Environment

- Floci version / image tag: floci/floci:latest-compat
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker Compose

MakerYuichi:
Fixes #2267

## What's wrong

Certain AWS services (SQS, RDS, DocDB, Neptune, EMR, Health) are hard-coded in the AWS CLI's SSL-common-name override (`awscli/customizations/overridesslcommonname.py`) to fall back to their real public endpoint whenever no `--endpoint-url` flag is passed on the command line — environment variables like `AWS_ENDPOINT_URL` are not consulted for that decision ([aws/aws-cli#8737](https://github.com/aws/aws-cli/issues/8737)).

Without a shim, e.g. `aws sqs create-queue` silently leaves for the real AWS endpoint instead of Floci, then fails authentication against the image's dummy `test`/`test` credentials — exactly the failure in #2267.

`docker/Dockerfile` (the JVM image) already carries a shim for this, hardcoded to `http://localhost:4566`. `docker/Dockerfile.compat` had no equivalent, making it the only published image whose `aws` depended on `AWS_ENDPOINT_URL` alone — and the image the initialization-hooks docs point readers at.

## Fix

Ports the shim to `Dockerfile.compat`, but reads `AWS_ENDPOINT_URL` at runtime instead of hardcoding it, so `docker run -e` / compose `environment` overrides (documented as supported in `docs/configuration/initialization-hooks.md`) keep working for direct `aws` calls, and deliberately clearing `AWS_ENDPOINT_URL` to reach real AWS still works too.

Verified locally with a built test image:
- An unflagged `aws sqs create-queue` now targets `AWS_ENDPOINT_URL` instead of real AWS.
- An explicit `--endpoint-url` flag (e.g. from `awslocal`, which always passes one) still takes precedence over the shim's injected flag — the CLI's argument parser takes the last flag, so there's no conflict.

## CI

Adds a `compat-aws-cli-shim` job to `compatibility.yml` that builds the compat image and asserts both behaviors above automatically, since nothing previously exercised direct `aws` calls against the compat image and the CLI version is unpinned (per `Dockerfile.compat:23`'s unpinned `pip3 install awscli`), so this behavior is free to regress silently again.
Also widens the workflow's path filter to include `docker/Dockerfile.compat`, which wasn't previously covered.

## Docs

No changes needed — `docs/configuration/initialization-hooks.md` already says scripts can call `aws` directly with no flag needed; this fix makes that claim accurate again rather than requiring a correction.

Thanks to @hampsterx for tracing the exact upstream mechanism and pointing at the existing shim pattern in `docker/Dockerfile` to port from.

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -6,6 +6,7 @@ on:
       - 'src/**'
       - 'pom.xml'
       - 'docker/Dockerfile.native-package'
+      - 'docker/Dockerfile.compat'
       - 'docker/entrypoint.sh'
       - 'docker/localstack-parity.sh'
       - 'compatibility-tests/**'
@@ -19,6 +20,45 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  compat-aws-cli-shim:
+    name: compat / aws CLI endpoint shim
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v7.0.1
+
+      - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+
+      - name: Build compat image
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          file: docker/Dockerfile.compat
+          build-args: VERSION=latest
+          load: true
+          tags: floci-compat-shim-test
+          cache-from: type=gha,scope=compat-shim
+          cache-to: type=gha,scope=compat-shim,mode=max
+
+      - name: Unflagged aws call routes to AWS_ENDPOINT_URL
+        run: |
+          out=$(docker run --rm floci-compat-shim-test sh -c \
+            "aws sqs create-queue --queue-name smoke-test 2>&1" || true)
+          echo "$out"
+          echo "$out" | grep -q "localhost:4566" \
+            || { echo "::error::unflagged aws call did not target localhost:4566"; exit 1; }
+          echo "$out" | grep -qi "queue.amazonaws.com\|InvalidClientTokenId" \
+            && { echo "::error::unflagged aws call reached real AWS endpoint"; exit 1; }
+          exit 0
+
+      - name: Explicit --endpoint-url still overrides the shim
+        run: |
+          out=$(docker run --rm floci-compat-shim-test sh -c \
+            "AWS_ENDPOINT_URL=http://example.com aws --endpoint-url http://other.example.com sqs list-queues 2>&1" || true)
+          echo "$out"
+          echo "$out" | grep -q "other.example.com" \
+            || { echo "::error::explicit --endpoint-url flag was not honored over AWS_ENDPOINT_URL"; exit 1; }
+
   build-native:
     name: Build native floci image
     runs-on: ubuntu-24.04-arm
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -21,11 +21,11 @@ concurrency:
 
 jobs:
   compat-aws-cli-shim:
-    name: compat / aws CLI endpoint shim
+    name: compat / aws CLI endpoint resolution
     runs-on: ubuntu-24.04-arm
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v7.0.1
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
 
@@ -40,9 +40,9 @@ jobs:
           cache-from: type=gha,scope=compat-shim
           cache-to: type=gha,scope=compat-shim,mode=max
 
-      - name: Unflagged aws call routes to AWS_ENDPOINT_URL
+      - name: Unflagged aws sqs call routes to AWS_ENDPOINT_URL
         run: |
-          out=$(docker run --rm floci-compat-shim-test sh -c \
+          out=$(docker run --rm --entrypoint sh floci-compat-shim-test -c \
             "aws sqs create-queue --queue-name smoke-test 2>&1" || true)
           echo "$out"
           echo "$out" | grep -q "localhost:4566" \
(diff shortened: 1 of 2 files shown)


hectorvent:
Thanks for this, and for the writeup. The diagnosis is correct and tracing it to the actual upstream mechanism is good work. I reproduced it locally and want to redirect the fix, though.

## Confirming the mechanism

It is `awscli/customizations/overridesslcommonname.py`, aws-cli v1 only, and the handler is this:

```python
def update_endpoint_url(session, parsed_globals, **kwargs):
    service = parsed_globals.command
    endpoints = SSL_COMMON_NAMES.get(service)
    if endpoints is not None and parsed_globals.endpoint_url is None:
        ...
```

It only checks whether the `--endpoint-url` flag was passed, so there is no env-var escape hatch at all. Verified on aws-cli 1.46.0 in us-east-1: `AWS_ENDPOINT_URL`, `AWS_ENDPOINT_URL_SQS` and `BOTO_DISABLE_COMMONNAME=true` all still resolve to `https://queue.amazonaws.com/`.

## Why not the wrapper

Injecting `--endpoint-url` puts the global value at the very top of the endpoint precedence chain, above `AWS_ENDPOINT_URL_<SERVICE>`, above the config file `[services ...]` sections and profile-level `endpoint_url`, and above `AWS_IGNORE_CONFIGURED_ENDPOINT_URLS`. The image already ships `ENV AWS_ENDPOINT_URL=http://localhost:4566`, so the wrapper would be active for everyone by default.

Those six services are already broken for service-specific overrides today, but s3, dynamodb and everything else are not. The wrapper trades six broken services for all of them becoming un-overridable. Someone pointing `AWS_ENDPOINT_URL_S3` at MinIO or at real S3 while the rest stays on Floci would silently lose that.

The same flaw is in `docker/Dockerfile`, so porting it spreads the problem rather than fixing it. That one is on us, not you.

## What I would like instead

Move `Dockerfile.compat` to AWS CLI v2 and delete the shim. There is no `overridesslcommonname.py` and no `SSL_COMMON_NAMES` table anywhere in the v2 branch, so v2 resolves SQS through the normal chain. The image already sets `AWS_ENDPOINT_URL`, which is all v2 needs, so the fix becomes a removal rather than an addition.

We need this regardless of #2267. AWS CLI v1 entered maintenance mode on 2026-07-15 and reaches end-of-support on 2027-07-15, receiving no API updates for new or existing services in between. For an emulator image that gap widens every release, and it also means aws/aws-cli#8737 will never be fixed upstream.

```dockerfile
ARG AWSCLI_VERSION=2.36.24
RUN microdnf install -y --nodocs unzip \
 && case "$(uname -m)" in \
      x86_64)  arch=x86_64 ;; \
      aarch64) arch=aarch64 ;; \
      *) echo "unsupported arch: $(uname -m)" >&2; exit 1 ;; \
    esac \
 && curl -fsSL "https://awscli.amazonaws.com/awscli-exe-linux-${arch}-${AWSCLI_VERSION}.zip" -o /tmp/awscliv2.zip \
 && unzip -q /tmp/awscliv2.zip -d /tmp \
 && /tmp/aws/install --bin-dir /usr/bin --install-dir /usr/local/aws-cli \
 && rm -rf /tmp/awscliv2.zip /tmp/aws \
 && microdnf remove -y unzip \
 && microdnf clean all
```

Notes:

- v2 is not on PyPI, so `pip3 install awscli` always gives you v1. Drop `awscli` from the pip line, keep `boto3`.
- Pin the version in the URL. The unpinned install is how this regressed silently in the first place. AWS publishes a `.sig` next to each zip if you want signature verification too.
- `uname -m` rather than `TARGETARCH`, which is only set under buildx.
- Set `ENV AWS_PAGER=""`. Retry mode also moves from legacy to standard by default, which is fine for us but worth a changelog line.
- Expect the image to grow by roughly 100 MB, since v2 bundles its own Python runtime. Call that out in the description so it is a conscious tradeoff rather than a surprise.

## On `/etc/floci/aws/config`

Let's drop this part. Everything the image needs is already env-driven: region via `AWS_DEFAULT_REGION`, endpoint via `AWS_ENDPOINT_URL`, output via `AWS_DEFAULT_OUTPUT`, pager via `AWS_PAGER`. Keep the `ENV AWS_CONFIG_FILE` line, since it usefully isolates us from a bind-mounted `~/.aws/config`, but a file behind it is not needed. Botocore handles the path not existing.

One consequence to flag in the description rather than work around: `cli_binary_format` is the one setting with no env var (its provider chain in v2 is config-file then a `base64` constant), so v2 will default blob params to base64 where v1 defaulted to `raw-in-base64-out`. An init hook doing `aws lambda invoke --payload '{"a":1}'` will need `--cli-binary-format raw-in-base64-out`. That matches what users get running v2 against real AWS, so I would rather take the v2 default and document it as a breaking change than paper over it with a config file.

## On the CI job

This is the best part of the PR and I want it landed. Three fixes:

- `docker run --rm image sh -c "..."` will be swallowed by the image ENTRYPOINT. It needs `--entrypoint sh`. Worth confirming those steps actually executed rather than the entrypoint eating them.
- Add a case for `AWS_ENDPOINT_URL_SQS`, and one for the flag after the subcommand (`aws sqs list-queues --endpoint-url ...`), which is the more common form and goes through a different parser than the pre-subcommand case you tested.
- Pin `actions/checkout` to a full commit SHA to match the other actions in the job. Greptile is right on that one.

This ended up a bigger ask than what you signed up for, so no pressure to take all of it. If you would rather land just the CI job here, I am happy to split the v2 migration into its own issue. Either way, thanks for digging into the real cause instead of papering over it. It made this review much faster.

hectorvent (changes requested):
Check previous comment.

MakerYuichi:
Updated per @hectorvent's review: migrated Dockerfile.compat to AWS CLI v2 and dropped the --endpoint-url shim entirely, rather than porting it from docker/Dockerfile.

v1's overridesslcommonname.py has no environment-variable escape hatch at all for its six affected services, and injecting --endpoint-url globally as the shim did sits above AWS_ENDPOINT_URL_<SERVICE> in the endpoint
precedence chain silently breaking legitimate per-service overrides for every other service, not just the six broken ones. v2 has no equivalent override code, so AWS_ENDPOINT_URL alone is sufficient and the fix becomes a removal rather than an addition.

Also dropped the /etc/floci/aws/config file write everything the image needs is already env-driven (region, endpoint, output, pager), and botocore handles the config path not existing.

CI job renamed and expanded per review: added --entrypoint sh to every docker run (the image's ENTRYPOINT was previously swallowing the test commands unnoticed), added a case for AWS_ENDPOINT_URL_SQS alone, added a
case for --endpoint-url after the subcommand (a different parser path than before the subcommand), and pinned actions/checkout to a full commit SHA.

Verified all four scenarios locally against a built image before pushing.

hectorvent:
Thank you for this, and the write-up is one of the best we have had: root cause, the exact upstream file, the upstream issue, and the maintenance-mode timing all in one place.

I verified the v1 half against the AWS CLI source: `overridesslcommonname.py` hardcodes exactly those six services and gates only on `parsed_globals.endpoint_url`, so no environment variable can redirect them. I could not verify the v2 half locally, since only v1 is checked out here, but the CI job you added settles it empirically against the pinned binary, which is better evidence than reading source. `sdk-test-awscli` staying green covers the regression risk of the version swap.

**(blocking)** Only the title. Your second commit dropped the shim and migrated to CLI v2, so the title now describes the approach you abandoned. We squash merge and generate the changelog from the title, so it would announce a shim rather than the v1 to v2 migration. Something like `fix(compat): install AWS CLI v2 in the compat image so AWS_ENDPOINT_URL is honored` would read right.

Retitle and this is good to go.

hectorvent:
Perfect, that is exactly it. The title now describes the v1 to v2 migration rather than the shim you dropped, so the changelog entry will be right.

Nothing else changed: the head is still `8c2604e6`, CI is 13/13 including the endpoint-resolution job you added and `sdk-test-awscli`, and the branch is clean against main.

No blockers from my side.

hectorvent (approved):

PR #2316, opened 2026-08-14
feat(bedrockagentcore): emulate the AgentCore control plane and InvokeAgentRuntime stub

### LINKED ISSUE #1892: feat: emulate Amazon Bedrock AgentCore (control plane + data-plane invoke)
## Proposal

Add emulation for Amazon Bedrock AgentCore so AWS SDK/CLI clients can manage and
invoke agent runtimes against Floci on port 4566. No emulator currently covers
AgentCore, and coding agents / AgentCore users need a local, credential-free target.

Concretely, this makes AWS Loom runnable on Floci: an AgentCore-based workload can
be created, deployed, and invoked entirely against the local emulator with no AWS
account, closing the loop for local development and CI of AgentCore applications.

Two AWS services are involved:

| Service | API version | Protocol | Role |
|---|---|---|---|
| `bedrock-agentcore-control` | 2023-06-05 | REST JSON | stateful management registry |
| `bedrock-agentcore` | 2024-02-28 | REST JSON (binary payload) | `InvokeAgentRuntime` canned-response stub |

## Scope

Control plane (stateful CRUD): agent runtimes + immutable versions, runtime
endpoints (with an auto-created `DEFAULT`), gateways + gateway targets, memory,
workload identity, and resource tagging. Data plane: an `InvokeAgentRuntime` stub
that returns a fixed, configurable response and never executes a real agent.

Out of scope (documented as future work via the same extension pattern): real
inference, streaming invoke, gateway rules, policy engine, registry, payment,
browser, code interpreter, dataset, evaluator, harness, token vault, credential
providers, and IAM enforcement.

## Design and conventions

Follows the existing Floci layering (JAX-RS controller → service → model),
`ServiceRegistry` descriptor registration, `EmulatorConfig`/`application.yml`
toggles, `StorageFactory` storage, and the shared `/tags/` `TagHandler` route.
Designed clean-room from the AWS public API reference.

A design note covering scope, the three AgentCore wire-protocol styles (standard
REST, RPC-in-path for workload identity, action-suffix for memory), and an
upgrade path to the remaining ~140 operations is included with the implementation.

## Status

Implemented and verified locally against the CI gates:
- `mvn test -B`: full suite green (2 new services, unit + RestAssured integration tests)
- AWS SDK v2 compatibility tests (sdk-test-java)
- Native (GraalVM/Mandrel) build succeeds; SDK compat tests pass against the native binary
- `make docs-test` + `make docs-check` green; Conventional Commits satisfied

A single PR for the whole feature is ready to open.

## Questions for maintainers

1. Is AgentCore support in scope for Floci?
2. Preference on PR granularity — one feature PR, or split (control-plane runtime
   CRUD + invoke first, then tagging / workload identity / gateway / memory)?
3. Any preference on the service key / config naming for the two services?

lex00:
Closes #1892.

Adds two services, bedrock-agentcore-control for management and bedrock-agentcore for InvokeAgentRuntime. Both are REST JSON, registered through ResolvedServiceCatalog and EmulatorConfig, with storage wired through StorageFactory.

Covers agent runtime CRUD, versions and endpoints, gateways and targets, memory, workload identity, tagging, and clientToken idempotency. InvokeAgentRuntime returns a canned response by design and validates that the runtime exists. Streaming is not emulated. Design notes are in docs/design/bedrock-agentcore.md.

19 of the 43 files are tests, including seven compatibility-test classes in sdk-test-java. Persistence is covered with PersistentStorage across a fresh factory rather than InMemoryStorage.

Merged with current main. make docs-check passes.


lex00:
No rush on this one, thank you

lex00:
Two of the four are real and fixed in 96c97eec.

The endpoint version is now resolved against the runtime's published versions on create and update, so a nonexistent one is rejected instead of producing a READY endpoint whose runtime ARN resolves to nothing.

Tag operations now check the account and region in the supplied ARN against the request. Both services resolve from the id suffix alone, so a foreign ARN naming a local id could read or change its tags. Reported as not found rather than a validation error, so the response does not confirm what sits behind the foreign identity.

On the runtime ARN parse, the version is already separated. The ARN splits on the colon separator, so parts[5] is agent/<uuid> and the version lands in parts[6]. BedrockAgentCoreTaggingIdentityIntegrationTest tags using the exact ARN CreateAgentRuntime returns and passes.

On the cleanup catches, 101 compatibility test classes use that shape and 6 of them are these. Changing only the new ones would leave them inconsistent with the other 95, so I left it.

pgermosen (approved):
Impressive fidelity for a genuinely awkward wire protocol. Cross-checked 14+ operations across both bedrock-agentcore-control (2023-06-05) and bedrock-agentcore (2024-02-28) against the real botocore service models — every HTTP method + path matches exactly, including the three distinct routing styles the design doc calls out: standard REST (PUT /runtimes/ returning 202, matching AWS's actual response code), RPC-in-path for workload identity (POST /identities/CreateWorkloadIdentity etc.), and action-suffix for memory (POST /memories/create, GET /memories/{memoryId}/details). InvokeAgentRuntime's POST /runtimes/{agentRuntimeArn}/invocations with the ARN-containing-slashes path pattern is also correct.

Both resolved greptile P1s check out in code (the ARN version-suffix lookup fix and the endpoint-unpublished-version guard). The still-open "Gateway ARN identity discarded" thread isn't currently exploitable — traced the only call path and confirmed BedrockAgentCoreTagHandler.requireLocalIdentity() validates region+account before ever reaching the gateway service, deliberately returning 404 rather than a validation error to avoid confirming what exists behind a foreign identity. Worth adding the same check inside findByArn itself for defense-in-depth against future direct callers, but not blocking. The open P2 (swallowed cleanup exceptions in the compat test) is a minor, non-blocking test-hygiene nit.

61/61 targeted tests pass locally, and CI is green across the board including the full native build and every native compat suite. No CODEOWNERS paths touched, no AI trailers, no conflicts.

LGTM, approving.

PR #2319, opened 2026-08-15
fix(iam): route long-term access keys to owning accounts

aryasaatvik:
## Summary

Route requests signed with long-term IAM access keys to the account that owns the key, including presigned requests. This preserves multi-account isolation instead of falling back to the configured default account.

## Account resolution

| Credential | Resolved account |
| --- | --- |
| 12-digit access key | Account encoded directly in the key |
| Active IAM access key | Account namespace containing the key |
| Temporary STS credential | Account recorded by the live session |
| Unknown, inactive, or expired credential | Existing default-account fallback |

```text
Authorization header or X-Amz-Credential
  -> extract access key ID
  -> resolve active IAM key or live STS session
  -> populate RequestContext with the owning account
```

Account-aware storage scans now preserve each entry's account ID and logical key. Legacy unprefixed entries remain attributed to the configured default account.

## Validation

- `./mvnw test -Dtest=AccountAwareStorageBackendTest,AccountContextFilterTest,IamServiceTest`
  - 107 tests passed
- `./mvnw test`
  - 9,991 tests passed
  - 9 skipped

### DIFF
--- a/docs/configuration/advanced/application-yml.md
+++ b/docs/configuration/advanced/application-yml.md
@@ -87,7 +87,7 @@ floci:
     #   - localhost.localstack.cloud
 
   auth:
-    validate-signatures: false               # Set to true to enforce AWS SigV4 validation
+    validate-signatures: false               # Set to true to verify S3 presigned URL signatures
     presign-secret: local-emulator-secret    # HMAC secret for S3 pre-signed URL verification
 
   tls:
--- a/docs/configuration/docker-compose.md
+++ b/docs/configuration/docker-compose.md
@@ -173,7 +173,7 @@ The most frequently set variables when running Floci as a Docker image:
 | `FLOCI_STORAGE_MODE` | `memory` | `memory`, `persistent`, `hybrid`, or `wal` |
 | `FLOCI_STORAGE_PERSISTENT_PATH` | `./data` | Directory for persistent storage |
 | `FLOCI_SERVICES_DOCKER_NETWORK` | _(none)_ | Docker network for spawned containers (Lambda, ElastiCache, RDS, OpenSearch, MSK) |
-| `FLOCI_AUTH_VALIDATE_SIGNATURES` | `false` | Verify AWS request signatures |
+| `FLOCI_AUTH_VALIDATE_SIGNATURES` | `false` | Verify S3 presigned URL signatures |
 | `FLOCI_SERVICES_LAMBDA_EPHEMERAL` | `false` | Remove Lambda containers after each invocation |
 
 For the complete list of every `FLOCI_*` variable, see [Environment Variables Reference](./environment-variables.md).
--- a/docs/configuration/environment-variables.md
+++ b/docs/configuration/environment-variables.md
@@ -22,7 +22,7 @@ Floci is configured exclusively through environment variables. Every option belo
 
 | Variable | Default | Description |
 |---|---|---|
-| `FLOCI_AUTH_VALIDATE_SIGNATURES` | `false` | When `true`, verifies AWS Signature V4 on every request. Leave `false` for local development |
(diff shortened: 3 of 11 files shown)


pgermosen (commented):
@aryasaatvik thanks for this — the account-resolution logic itself (scanAllAccountEntries, hasAccountPrefix, the IamService change) is clean and minimal, and the 107 targeted tests pass locally for me too.

Two things before this can land:

**1. Real merge conflict against current `main`.** `main` independently added `scanAllAccountsWithRawKeys()` in `AccountAwareStorageBackend.java` at the same spot this PR adds `scanAllAccountEntries()` — a textual collision, not a design one (both methods are legitimate and can coexist), but it needs a rebase.

**2. The still-open greptile security finding is worth taking seriously, and I want to give you the fuller picture rather than just relay it.** I checked `docs/configuration/multi-account.md` (already on `main`, unmodified by this PR) and it documents floci's existing, intentional default: a 12-digit numeric AKID is trusted as the account ID with no SigV4 signature check, same convention as LocalStack. This PR extends that same unverified-AKID-based trust model from synthetic 12-digit account numbers to *real* IAM access-key IDs (`AKIA...`) — an identifier that's normally paired with a secret for authentication, so it's a meaningfully different trust surface even though the underlying philosophy ("AKID alone routes the account when signatures aren't validated") isn't new to this PR.

What concerned me enough to check further: the docs also point to `FLOCI_AUTH_VALIDATE_SIGNATURES` as "enforce real signature validation," implying there's an escape hatch for anyone who wants this closed. I traced it and it's only wired up in `PreSignedUrlGenerator`/`PreSignedUrlFilter` (S3 presigned URLs) — there's no general `Authorization`-header signature verification anywhere in the codebase, so that flag can't actually protect the new IAM-key routing path this PR adds. If you (or a maintainer) are fine extending the existing trust model this way — which is defensible given the precedent — it'd still be worth a line in `multi-account.md` noting that long-term IAM key routing, like the 12-digit case, isn't signature-backed and `FLOCI_AUTH_VALIDATE_SIGNATURES` doesn't count

aryasaatvik:
@pgermosen Thanks for the detailed review. I rebased this onto current `main` and preserved both storage APIs: the upstream `scanAllAccountsWithRawKeys()` method and this PR's account-preserving `scanAllAccountEntries()` method.

I also documented the trust boundary explicitly: long-term IAM access-key IDs participate in account routing without general SigV4 proof, this is not an authentication/security boundary, and `FLOCI_AUTH_VALIDATE_SIGNATURES` currently applies only to S3 presigned URL verification. That keeps this change aligned with Floci's existing emulator routing model without implying the flag protects general `Authorization` headers.

Validation on `81fbfe1b`: 116 focused IAM/account/storage tests passed; the full Java 25 suite passed with 10,575 tests run, 0 failures, 0 errors, and 9 skipped; the MkDocs build also passed.

pgermosen (approved):
Both points from my earlier review are resolved. The merge conflict against main (scanAllAccountEntries vs scanAllAccountsWithRawKeys) is gone — clean merge now. And docs(auth): clarify account-routing trust model directly addresses the security-framing concern: multi-account.md now states IAM-key routing is AKID-based, not proof of secret possession, and corrects the previously-misleading FLOCI_AUTH_VALIDATE_SIGNATURES description to scope it to S3 presigned URLs only.

resolveAccountId correctly filters on an Active key status (inactive keys fall through to the default account, with a dedicated test for it), and the new hasAccountPrefix check does proper 12-digit-then-slash validation rather than a naive substring split.

Ran AccountContextFilterTest (12/12), AccountAwareStorageBackendTest (12/12), and IamServiceTest (92/92) locally — all pass. docs-check clean, no CODEOWNERS-protected paths.

PR #2320, opened 2026-08-15
docs: add MAINTAINERS.md and route application code to the repo maintainers

hectorvent:
## Why

The org-wide [GOVERNANCE.md](https://github.com/floci-io/.github/blob/main/GOVERNANCE.md) says, under the Maintainer role:

> Maintainers are listed in each repository's `MAINTAINERS.md`.

No repository in the org had one. This is part of a rollout that makes that sentence true across `floci-io`.

## What

**`MAINTAINERS.md`**, a verbatim copy of the org-level template with the entries changed, so the family stays one shape rather than seventeen dialects. `GOVERNANCE.md` is deliberately **not** copied here: it is one of GitHub's supported default community health files, so it is already inherited from `floci-io/.github`, and a local copy would silently override the inherited one. The template links it by absolute URL instead.

**`.github/CODEOWNERS`** gains area rules routing application code to this repository's maintainers.

- Every Lead-reserved path is unchanged: `/.github/`, `/.releaserc.json`, `/pom.xml`, `/docker/`.
- No `*` default owner. A default owner would auto-request one person on every PR, which turns a bus-factor fix into a bottleneck.
- The header now records that the **last** matching pattern wins, which is what keeps the Lead-reserved paths authoritative over the broader area rules above them.

## Verification

- `GET /repos/floci-io/floci/codeowners/errors?ref=docs/maintainer-roster` returns no errors. Worth checking explicitly: a `CODEOWNERS` handle without write access to the repo is ignored silently, with no error surfaced anywhere.
- @pgermosen confirmed at `write` on this repository.
- Documentation and repository config only. No source, build inputs, or runtime behavior touched.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,12 +1,19 @@
 # CODEOWNERS for floci
 #
-# Scopes release-sensitive paths to the primary maintainer so CI, release
-# config, and Dockerfile changes always trigger a review ping. Add more
-# owners as the maintainer roster grows.
+# Routes application code to this repository's maintainers (see MAINTAINERS.md) and keeps
+# release-sensitive paths with the Lead, per the org-wide GOVERNANCE.md.
+#
+# Order matters: the LAST matching pattern wins, so the Lead-reserved paths below stay
+# authoritative over the broader area rules above them.
 #
 # Docs: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
 # Default: no global owner (unowned files fall through to normal review rules).
+
+# Application code
+/src/                    @hectorvent @pgermosen
+/compatibility-tests/    @hectorvent @pgermosen
+/docs/                   @hectorvent @pgermosen
 
 # CI, release automation, workflow hygiene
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
**Verified independently (not just trusting the PR body):**
- `GET /repos/floci-io/floci/codeowners/errors?ref=docs/maintainer-roster` returns `{"errors":[]}` — CODEOWNERS is syntactically valid and every referenced handle resolves.
- `GET /repos/floci-io/floci/collaborators/pgermosen/permission` returns `write` — the non-obvious failure mode the PR calls out (a CODEOWNERS handle without write access is silently ignored) doesn't apply here.
- All four Lead-reserved paths (`/.github/`, `/.releaserc.json`, `/pom.xml`, `/docker/`) are unchanged and stay `@hectorvent`-only.
- No `*` default owner added.
- `GOVERNANCE.md` correctly not duplicated locally — it's one of GitHub's supported default community health files, so a local copy would have silently shadowed the inherited `floci-io/.github` version.
- `make docs-check` clean. Conventional Commits title, no AI-attribution trailers.

No findings. LGTM.

PR #2323, opened 2026-08-15
test(ses): guard LastKeyGenerationTimestamp fractional-epoch wire format

okinaka:
## Summary

Follow-up to the DKIM actions work (#1904): adds the regression guard the reviewer asked for but that landed after the squash-merge.

`GetEmailIdentity` serializes `LastKeyGenerationTimestamp` as a fractional unix epoch (`toEpochMilli() / 1000.0`) so the SDK's number unmarshaller keeps the sub-second precision AWS sends; `getEpochSecond()` would silently drop it, and nothing in the suite caught that change. This adds an assertion to the existing compat test so the fractional form is guarded.

- `SesIdentityAttributesTest.putEmailIdentityDkimSigningAttributes_regeneratesTokensOnKeyLengthChange` now reads `dkimAttributes().lastKeyGenerationTimestamp()` back after the signing PUT and asserts it retains a sub-second component (`getNano()` is non-zero).

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Verified against real AWS SES v2: `LastKeyGenerationTimestamp` is a `Timestamp` under `rest-json` with no explicit format, so the wire form is a unix-epoch number carrying sub-second precision. The guard asserts the SDK-unmarshalled value keeps that precision, catching a regression to whole-second epoch.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SesIdentityAttributesTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SesIdentityAttributesTest.java
@@ -45,6 +45,7 @@ import software.amazon.awssdk.services.sesv2.model.PutEmailIdentityDkimSigningAt
 import software.amazon.awssdk.services.sesv2.model.PutEmailIdentityMailFromAttributesRequest;
 import software.amazon.awssdk.services.sesv2.model.VerificationStatus;
 
+import java.time.Instant;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


pgermosen:
@okinaka Good catch on the original gap, and the underlying premise checks out — `SesController.java:2228` does serialize `LastKeyGenerationTimestamp` via `toEpochMilli() / 1000.0` today, so this guard is protecting something real.

One thing worth fixing before merge: both greptile and Copilot independently flagged that `assertThat(lastKeyGen.getNano()).isNotZero()` can fail spuriously. The timestamp comes from the system clock, so if a DKIM regeneration happens to land on an exact whole-second boundary, `nano` is legitimately `0` even with the correct `toEpochMilli()/1000.0` serialization — both the correct and the regressed (`getEpochSecond()`) forms round-trip to the same value at that instant, so the assertion can't tell them apart there, and just fails the test on nothing. Low-probability per run, but real given how often this suite runs. Regenerating the key once more when `nano == 0` (as Copilot suggested) or asserting against a controlled/injected clock would make the guard actually deterministic.

hectorvent:
Thank you for coming back for this after #1904 squash-merged without it.

I checked the premise against the model rather than the implementation: sesv2 is `rest-json` and `LastKeyGenerationTimestamp` carries no explicit `timestampFormat`, so it inherits epoch-seconds-as-number, which is exactly where a fractional component survives. So `toEpochMilli() / 1000.0` is right and `getEpochSecond()` would drop precision the SDK would happily accept. The guard is pinning a real contract.

The retry loop is worth the words you spent on it. Asserting `getNano()` is non-zero would flake roughly once in a thousand when the clock lands on a whole second, which is also the one case where whole-second epoch round-trips identically and the guard is blind. Regenerating with alternating key lengths and a short wait makes the samples independent instead of just retrying the same instant.

One thing worth a comment for whoever reads this next: the alternation is load bearing, not stylistic. `SesService` only regenerates when the requested length differs from the current one, so a loop that resent a fixed length would never advance the timestamp and would fail after five passes rather than retrying anything.

No blockers from my side.

PR #2325, opened 2026-08-15
fix(cloudformation): don't double-encode string-form SNS RedrivePolicy and FilterPolicy

### LINKED ISSUE #2317: [BUG] CloudFormation: SNS Subscription RedrivePolicy is double-encoded when supplied as a JSON string
## Describe the bug

`AWS::SNS::Subscription`'s `RedrivePolicy` is double-encoded when the template supplies it as an already-serialized JSON string rather than a JSON object, so the subscription's redrive silently does not work.

Found by @floverfelt while testing #1939, which fixed the identical bug in `AWS::SQS::Queue`. Filing separately because it is a different provisioner and was correctly kept out of that PR's scope.

## Root cause

`CloudFormationResourceProvisioner:1576`:

```java
attributes.put("RedrivePolicy", engine.resolveNode(props.path("RedrivePolicy")).toString());
```

`CloudFormationTemplateEngine.resolveNode` collapses an `Fn::Join` to a Jackson `TextNode`. `TextNode.toString()` emits the JSON *representation* of the string — quoted and escaped — so an already-serialized policy is JSON-encoded a second time and SNS cannot parse it.

CDK commonly emits exactly this form, splicing the DLQ ARN in with `Fn::Join`:

```json
"RedrivePolicy": {
  "Fn::Join": ["", [
    "{\"deadLetterTargetArn\":\"",
    { "Fn::GetAtt": ["SomeDLQ", "Arn"] },
    "\"}"
  ]]
}
```

Real CloudFormation accepts both the object form and the string form.

## Expected behaviour

The stored attribute is the policy JSON:

```json
{"deadLetterTargetArn":"arn:aws:sqs:us-east-1:000000000000:some-dlq"}
```

## Actual behaviour

It is re-encoded:

```json
"{\"deadLetterTargetArn\":\"arn:aws:sqs:us-east-1:000000000000:some-dlq\"}"
```

The object form works today; only the string form is affected. The stack still reaches `CREATE_COMPLETE`, so the failure surfaces later as messages never reaching the dead-letter queue.

## Fix

The same unwrap merged in #1939, which keeps the object form working because `toString()` still handles non-textual nodes:

```java
JsonNode redrive = engine.resolveNode(props.path("RedrivePolicy"));
attributes.put("RedrivePolicy", redrive.isTextual() ? redrive.asText() : redrive.toString());
```

`SqsCfnProvisionerTest.queuePassesThroughAlreadySerializedRedrivePolicyString` is the pattern to mirror for a regression test — reverting the unwrap there reproduces the double-encoded value exactly.

## Three more sites with the same shape, worth auditing together

Grepping `main` for the same construct turns up three others. I have **not** confirmed each is reachable in practice — that depends on whether real templates supply these as `Fn::Join` strings — but they share the identical failure mode and are cheap to fix in one pass:

| Location | Property |
|---|---|
| `CloudFormationResourceProvisioner:1567` | `AWS::SNS::Subscription` → `FilterPolicy` |
| `CloudFormationResourceProvisioner:4307` | Step Functions → `Definition` |
| `IamRoleCfnProvisioner:169` | IAM policy document |

Whoever picks this up may prefer a small shared helper — something like `resolveJsonAttribute(engine, node)` applying the textual unwrap — rather than repeating the ternary at five call sites and leaving the sixth to be found later.

## Credit

Diagnosis and suggested fix by @floverfelt on #1939, including the CDK `Fn::Join` reproduction and the `TextNode.toString()` root cause. Fixed for SQS by @Agboolafeyikemi in that PR; this issue tracks the remaining provisioners.

mansiverma897993:
## Summary

Closes #2317

`AWS::SNS::Subscription`'s `RedrivePolicy` (and `FilterPolicy`) is double-encoded when the template supplies it as an already-serialized JSON string rather than a JSON object, so the subscription's redrive silently does not work.

**Root cause:** `CloudFormationTemplateEngine.resolveNode` collapses an `Fn::Join` (the form CDK commonly emits, splicing the DLQ ARN in) to a Jackson `TextNode`. Calling `.toString()` on that node emits the JSON representation of the string - quoted and escaped - so an already-serialized policy is JSON-encoded a second time and SNS cannot parse it.

Real CloudFormation accepts both the object form and the string form; this fix makes the string form round-trip correctly.

## Changes

- Add `CloudFormationTemplateEngine#resolveJsonAttribute`, which unwraps textual results (`asText()`) while keeping normal serialization (`toString()`) for plain object/array nodes, so both forms keep working.
- Use it at every site that stores a resolved node as a JSON document:
  - `AWS::SNS::Subscription` -gt `RedrivePolicy` (the reported bug)
  - `AWS::SNS::Subscription` -gt `FilterPolicy`
  - Step Functions -gt `Definition`
  - `AWS::IAM::Role` inline `PolicyDocument` (`IamRoleCfnProvisioner`)
  - `AWS::SQS::Queue` -gt `RedrivePolicy` (already fixed inline in #1939; now uses the shared helper)
- Regression tests mirroring the `SqsCfnProvisionerTest.queuePassesThroughAlreadySerializedRedrivePolicyString` pattern from #1939:
  - Engine unit tests for the new helper (string form via `Fn::Join`, object form, null handling)
  - SNS integration test asserting `GetSubscriptionAttributes` returns the literal policy JSON (no backslash-escape sequences)
  - IAM provisioner unit test passing an already-serialized `PolicyDocument`

## Verification

- `CloudFormationTemplateEngineTest`, `SqsCfnProvisionerTest`, `IamRoleCfnProvisionerTest`, `CloudFormationIamAttachmentProvisionerTest`, `SnsServiceTest`, all `*CfnProvisionerTest` and the SNS CloudFormation integration tests pass.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -1768,7 +1768,7 @@ public class CloudFormationResourceProvisioner {
 
         Map<String, String> attributes = new HashMap<>();
         if (props.has("FilterPolicy") && !props.path("FilterPolicy").isNull()) {
-            attributes.put("FilterPolicy", engine.resolveNode(props.path("FilterPolicy")).toString());
+            attributes.put("FilterPolicy", engine.resolveJsonAttribute(props.path("FilterPolicy")));
         }
         if (props.has("FilterPolicyScope")) {
             attributes.put("FilterPolicyScope", engine.resolve(props.path("FilterPolicyScope")));
@@ -1777,7 +1777,7 @@ public class CloudFormationResourceProvisioner {
             attributes.put("RawMessageDelivery", engine.resolve(props.path("RawMessageDelivery")));
         }
         if (props.has("RedrivePolicy") && !props.path("RedrivePolicy").isNull()) {
-            attributes.put("RedrivePolicy", engine.resolveNode(props.path("RedrivePolicy")).toString());
+            attributes.put("RedrivePolicy", engine.resolveJsonAttribute(props.path("RedrivePolicy")));
         }
 
         var sub = snsService.subscribe(topicArn, protocol, endpoint, region, attributes);
@@ -4508,7 +4508,7 @@ public class CloudFormationResourceProvisioner {
         if (hasDefinitionString) {
             definition = resolveOptional(props, "DefinitionString", engine);
         } else if (hasDefinition) {
-            definition = engine.resolveNode(props.get("Definition")).toString();
+            definition = engine.resolveJsonAttribute(props.get("Definition"));
         } else {
             JsonNode location = engine.resolveNode(props.get("DefinitionS3Location"));
             String bucket = location.path("Bucket").asText(null);
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationTemplateEngine.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationTemplateEngine.java
@@ -145,6 +145,26 @@ public class CloudFormationTemplateEngine {
         return node;
     }
 
+    /**
+     * Resolves a node that must be stored as a JSON document (SNS/SQS RedrivePolicy and
+     * FilterPolicy, Step Functions definitions, IAM policy documents) to its string form.
+     * <p>
+     * resolveNode collapses intrinsics such as Fn::Join into a {@link TextNode}. Calling
+     * {@code toString()} on that node re-quotes and re-escapes the already-serialized JSON
+     * (e.g. CDK's Fn::Join-spliced RedrivePolicy), so the value is double-encoded and the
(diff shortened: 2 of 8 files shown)


pgermosen (approved):
Clean fix, and nice generalization of the pattern from #1939 into a shared helper rather than repeating the ternary at each site. Confirmed all five sites named in the issue are covered, including SQS being refactored to use the new `resolveJsonAttribute` instead of its own inline copy.

One honest note: I pulled the actual resource-provider JSON schema from `aws-cloudformation-resource-providers-sns` on GitHub, and it declares `RedrivePolicy`/`FilterPolicy` as `type: object`, so I can't confirm the string-form acceptance from the schema alone. But this is the identical root cause already fixed for SQS in this same codebase, and it's a well-known CDK pattern (splicing a DLQ ARN into a JSON string via `Fn::Join`), so I'd treat the premise as credible rather than independently re-proven against a live account.

Ran the full CloudFormation suite plus the SNS/SQS/IAM provisioner tests fresh: 236/236 pass. No conflicts, hygiene clean.

Approving.

Issue #2308, opened 2026-08-14
[FEAT] ec2: Transit Gateway and VPC attachments


pgermosen:
## Service

EC2 (Transit Gateway)

## API Action / Feature

Transit Gateway and its VPC attachments

## AWS Documentation

- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTransitGateway.html
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTransitGatewayVpcAttachment.html
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTransitGatewayRouteTable.html

## Why is this needed?

Transit Gateway is unimplemented. Any topology that routes between VPCs — or between a VPC and
on-premises — through a transit gateway fails at the first resource, so the whole networking
layer of such a stack is unreachable.

Worth stating explicitly because it is easy to get wrong: **this is not a new service.** Transit
Gateway actions are part of the **EC2 Query API** (`ec2.amazonaws.com`, form-encoded POST with
`Action=CreateTransitGateway`, XML response), so the work belongs in
`services/ec2/Ec2QueryHandler` alongside the existing VPC, subnet and route-table actions — not
in a new `services/transitgateway` package.

## Scope

This is a tracking issue. It is **not small** and should not be attempted in one change. The
natural split is three independent PRs:

**1. The gateway itself**
`CreateTransitGateway`, `DescribeTransitGateways`, `ModifyTransitGateway`,
`DeleteTransitGateway`. Includes `Options` (`AmazonSideAsn`, `AutoAcceptSharedAttachments`,
`DefaultRouteTableAssociation`, `DefaultRouteTablePropagation`, `DnsSupport`, `VpnEcmpSupport`)
and tag support via the existing EC2 `TagSpecification` handling. Backs
`aws_ec2_transit_gateway`.

**2. VPC attachments**
`CreateTransitGatewayVpcAttachment`, `DescribeTransitGatewayVpcAttachments`,
`DescribeTransitGatewayAttachments`, `ModifyTransitGatewayVpcAttachment`,
`DeleteTransitGatewayVpcAttachment`. Backs `aws_ec2_transit_gateway_vpc_attachment`, which is
the resource most stacks actually reach for. Note that `DescribeTransitGatewayAttachments` (the
generic form) and the VPC-specific form are distinct actions and the provider uses both.

**3. Route tables, associations and propagations**
[...]

PR #2329, opened 2026-08-15
feat(ec2): support transit gateways

ck-michals:
Implements transit gateways: `CreateTransitGateway`, `DescribeTransitGateways`,
`ModifyTransitGateway` and `DeleteTransitGateway`, as EC2 Query actions in `Ec2QueryHandler`
alongside the existing VPC and route table ones.

Part one of #2308. Metadata only, as that issue sets out: nothing routes packets, and the value is
in ids later resources can reference and describes that round-trip so plans converge.

The defaults and response shapes come from creating a real gateway in a sandbox account and
deleting it again, rather than from the API reference.

| | observed |
|---|---|
| defaults with `Options` omitted | `amazonSideAsn` 64512; `dnsSupport`, `vpnEcmpSupport`, `defaultRouteTableAssociation`, `defaultRouteTablePropagation` `enable`; `autoAcceptSharedAttachments`, `securityGroupReferencingSupport`, `multicastSupport` `disable` |
| `transitGatewayCidrBlocks` | omitted from the response entirely when unset, not sent empty |
| default route table | minted during creation; its id is already on the **create** response, as both `associationDefaultRouteTableId` and `propagationDefaultRouteTableId`, naming the same table |
| `tagSet` | present on create and describe, **absent** on modify and delete |
| state | `pending`, reaching `available` after ~49s; `deleting`, reaching `deleted` after ~64s |
| unknown id | `InvalidTransitGatewayID.NotFound` — "Transit Gateway tgw-… was deleted or does not exist." |
| malformed id | `InvalidTransitGatewayID.Malformed` — "Invalid Transit Gateway id tgw-nope." |

Two of those I would have got wrong from the documentation alone:
`securityGroupReferencingSupport` defaults to `disable`, and the `tagSet` asymmetry between
create/describe and modify/delete is not something the reference calls out. Both are asserted in
the tests rather than just implemented.

## Deviations, and why

**State is reported settled rather than transitional.** A created gateway is returned as
`available` and a deleted one as `deleted`, skipping `pending` and `deleting`. Nothing here is
slow, so this is the compression `createManagedPrefixList` already applies to
`create-in-progress`, and it keeps a provider's create/delete waiter from spinning on a state that
would never change.

**The default route table is created but not otherwise reachable.** Creating a gateway with either
default-route-table option enabled mints the route table AWS mints, so the two ids on the
gateway's options are real rather than absent. The actions that operate on route tables directly —
create, associate, enable propagation, describe — are part three and are not implemented here, and
neither are attachments. Opting out of both defaults leaves both ids absent, as on AWS.

## Note on the service constructor

`Ec2Service` gains two storage backends. Rather than widening the package-private constructor that
test fixtures resolve, that overload keeps its arity and delegates with `new InMemoryStorage<>()`,
and the transit-gateway-aware constructor sits beside it — the pattern #2179 established. Moving
that slot is what broke CI on #2103 and #2106, so it seemed worth not repeating.

## Tests

- `Ec2ServiceTest`: option defaults, requested options overriding them, the minted route table and
  its absence when both defaults are disabled, filters and both id errors, modify including CIDR
  add and remove, delete removing the gateway and its route table, and `tgw-` tag classification.
- `Ec2TransitGatewayIntegrationTest`: the Query wire shapes end to end, including the `tagSet`
  asymmetry and the omitted-when-unset elements.
- `ec2.bats`: create, read back, modify, delete and the unknown-id rejection through the AWS CLI.

Full suite: 10244 tests, 0 failures. One error, `NeptuneOpenCypherIntegrationTest`, which
reproduces identically on unmodified `main` on this machine and is unrelated to this branch —
the diff is EC2 only.

### DIFF
--- a/compatibility-tests/sdk-test-awscli/test/ec2.bats
+++ b/compatibility-tests/sdk-test-awscli/test/ec2.bats
@@ -277,6 +277,11 @@ create_sg_pair() {
     [ "$(json_get "$output" '.TransitGateway.Description')" = "after" ]
     [ "$(json_get "$output" '.TransitGateway.Options.DnsSupport')" = "disable" ]
     [ "$(json_get "$output" '.TransitGateway.Options.TransitGatewayCidrBlocks[0]')" = "10.100.0.0/16" ]
+
+    # Describe serializes the same list, so read it back through the CLI's own parser too.
+    run aws_cmd ec2 describe-transit-gateways --transit-gateway-ids "$TRANSIT_GATEWAY_ID"
+    assert_success
+    [ "$(json_get "$output" '.TransitGateways[0].Options.TransitGatewayCidrBlocks[0]')" = "10.100.0.0/16" ]
 }
 
 @test "EC2: delete a transit gateway" {
--- a/compatibility-tests/sdk-test-awscli/test/ec2.bats
+++ b/compatibility-tests/sdk-test-awscli/test/ec2.bats
@@ -264,6 +264,43 @@ create_sg_pair() {
     [ "$(json_get "$output" '.TransitGateways[0].Tags[0].Value')" = "bats-tgw" ]
 }
 
+@test "EC2: create a transit gateway with CIDR blocks" {
+    run aws_cmd ec2 create-transit-gateway \
+        --options 'TransitGatewayCidrBlocks=[10.99.0.0/16]'
+    assert_success
+    TRANSIT_GATEWAY_ID=$(json_get "$output" '.TransitGateway.TransitGatewayId')
+    [ "$(json_get "$output" '.TransitGateway.Options.TransitGatewayCidrBlocks[0]')" = "10.99.0.0/16" ]
+
+    run aws_cmd ec2 describe-transit-gateways --transit-gateway-ids "$TRANSIT_GATEWAY_ID"
+    assert_success
+    [ "$(json_get "$output" '.TransitGateways[0].Options.TransitGatewayCidrBlocks[0]')" = "10.99.0.0/16" ]
+}
+
+# Tags are changed after creation with create-tags rather than a tag specification, and read back
+# from describe-transit-gateways, which is how a Terraform tag update converges.
+@test "EC2: transit gateway tags changed after creation are visible on describe" {
+    local out
+    out=$(aws_cmd ec2 create-transit-gateway --tag-specifications \
+        'ResourceType=transit-gateway,Tags=[{Key=Name,Value=bats-tgw-tags}]')
+    TRANSIT_GATEWAY_ID=$(json_get "$out" '.TransitGateway.TransitGatewayId')
+
+    aws_cmd ec2 create-tags --resources "$TRANSIT_GATEWAY_ID" --tags 'Key=env,Value=prod' >/dev/null
+
+    run aws_cmd ec2 describe-transit-gateways --transit-gateway-ids "$TRANSIT_GATEWAY_ID"
+    assert_success
+    count=$(json_get "$output" '.TransitGateways[0].Tags | length')
+    [ "$count" = "2" ]
+    env_value=$(json_get "$output" '.TransitGateways[0].Tags[] | select(.Key=="env") | .Value')
+    [ "$env_value" = "prod" ]
+
+    aws_cmd ec2 delete-tags --resources "$TRANSIT_GATEWAY_ID" --tags 'Key=env' >/dev/null
+
+    run aws_cmd ec2 describe-transit-gateways --transit-gateway-ids "$TRANSIT_GATEWAY_ID"
+    assert_success
+    count=$(json_get "$output" '.TransitGateways[0].Tags | length')
+    [ "$count" = "1" ]
+}
+
 @test "EC2: modify a transit gateway" {
     local out
     out=$(aws_cmd ec2 create-transit-gateway --description "before")
--- a/compatibility-tests/sdk-test-awscli/test/ec2.bats
+++ b/compatibility-tests/sdk-test-awscli/test/ec2.bats
@@ -8,11 +8,15 @@ setup() {
     SG_VPC_ID=""
     SG_SOURCE_ID=""
     SG_TARGET_ID=""
+    TRANSIT_GATEWAY_ID=""
 }
 
 teardown() {
     if [ -n "$PREFIX_LIST_ID" ]; then
         aws_cmd ec2 delete-managed-prefix-list --prefix-list-id "$PREFIX_LIST_ID" >/dev/null 2>&1 || true
+    fi
+    if [ -n "$TRANSIT_GATEWAY_ID" ]; then
+        aws_cmd ec2 delete-transit-gateway --transit-gateway-id "$TRANSIT_GATEWAY_ID" >/dev/null 2>&1 || true
     fi
     for sg in "$SG_TARGET_ID" "$SG_SOURCE_ID"; do
         if [ -n "$sg" ]; then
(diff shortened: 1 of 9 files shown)


ck-michals:
Good catch on the divergence, though the fix turns out to be the opposite of the suggestion.

I checked `ModifyTransitGateway` against a live account before implementing. AWS does not create a
route table when you enable the option — it rejects the request unless an existing table is named:

```
$ aws ec2 modify-transit-gateway --transit-gateway-id tgw-… \
    --options 'DefaultRouteTableAssociation=enable'

An error occurred (InvalidParameterCombination) when calling the ModifyTransitGateway operation:
enable DefaultRouteTableAssociation conflicts with AssociationDefaultRouteTableId null
```

The rule is symmetric, and the unknown-table code is not the one I would have guessed:

| request | result |
|---|---|
| `DefaultRouteTableAssociation=enable`, no id | `InvalidParameterCombination` |
| `DefaultRouteTablePropagation=enable`, no id | `InvalidParameterCombination` |
| `DefaultRouteTableAssociation=disable` **with** an id | `InvalidParameterCombination` |
| `enable` with an id that does not exist | `InvalidRouteTableID.NotFound` |

So `modifyTransitGateway` validates rather than creates, in 1f945fb1: the flag and its id have to
move together in both directions, and any table named has to exist. The state you describe —
option enabled, id absent — is now unreachable, and a rejected call leaves the gateway untouched,
which the test asserts rather than only checking the error code.

Minting a table on enable would have diverged from AWS in a way a provider notices: it sends the
id it already holds in state, so a gateway that silently acquired a different table would drift on
the next plan.

Worth noting for anyone reading this later: because this part does not implement
`CreateTransitGatewayRouteTable`, the only table that can be named is the one the gateway minted
for itself at creation. Enabling a default that was created disabled therefore stays impossible
here until part three lands, which is the same shape as AWS — there, too, you have to have a table
before you can point at it.

ck-michals:
Confirmed and fixed in 753501b1. This one was right, and it went further than the flag and id:
the route table's own marker was going stale too.

Checked on a live account rather than reasoning about it. Disabling association on a gateway
created with both defaults enabled:

```
DefaultRouteTableAssociation  disable      ← AssociationDefaultRouteTableId key is gone entirely
DefaultRouteTablePropagation  enable       PropagationDefaultRouteTableId tgw-rtb-0f5bc…
tgw-rtb-0f5bc…  available  defAssoc False  defProp True
```

Three things worth naming, because two of them are not what I would have assumed:

- the id is **dropped from the response**, not blanked, so `elem()` skipping nulls already gives
  the right wire shape once the field is cleared
- the route table's `DefaultAssociationRouteTable` flips to false while `DefaultPropagationRouteTable`
  stays true — the markers move independently
- the table **survives**; disabling the default does not delete it

So the change carries a flag through to everything it governs: the id follows its flag, that
table's matching marker follows with it, the other default is untouched, and enabling with a named
table marks that table instead. The new test asserts all four, including that the table is still
there afterwards.

Taken with the previous finding, both came from the same gap — modify changed flags without
touching what they govern — but the two halves needed opposite remedies: enabling is something AWS
*refuses* without an existing table, disabling is something it *applies* to the table. Worth
noting since the first suggestion here was to create a table on enable, which would have drifted
against a provider that sends the id it already holds.

ck-michals:
Right again, and this one I should have caught myself — "does this id belong to this gateway" is a
question I would have asked of any other cross-referenced resource, and I only checked existence.

Fixed in 4a2b4140. AWS rejects a foreign table under the same code as one that exists nowhere, but
qualifies the message with the gateway. Gateway A naming gateway B's default table, on a live
account:

```
An error occurred (InvalidRouteTableID.NotFound) when calling the ModifyTransitGateway operation:
Transit Gateway Route Table tgw-rtb-0381f… was deleted or does not exist in
Transit Gateway tgw-0438f…
```

Both wordings are reproduced now: unqualified when the table exists nowhere, gateway-qualified when
it exists but belongs elsewhere. The test asserts the consequence you named rather than only the
code — the owner's table keeps its markers and the borrowing gateway stays disabled.

While here I folded the three rules these options obey into one validation block instead of letting
them accumulate as separate guards: the flag and its id agree in both directions, and any table
named exists and belongs to the gateway.

That is three findings on the same method, and they came from one root cause worth stating plainly:
`modifyTransitGateway` treated `Options` as a flat bag of independent fields when several of them
are joined — to each other, and to a stored record. Enabling is something AWS refuses without an
existing table, disabling is something it applies to the table, and either one has to respect
ownership. Thanks for staying on it.

ck-michals:
The divergence is real and it is fixed in b11aa646, but not the way the suggestion has it: AWS
accepts an id without its flag. Rejecting it would break a request the real API allows.

Rather than answer this one case, I probed the whole combination surface on a live account, since
three findings in a row here says the method was being reasoned about locally rather than
understood:

| request | stored state | result |
|---|---|---|
| id alone | option enabled | accepted |
| id alone | option disabled | `InvalidParameterCombination`, message quoting the **stored** flag |
| `enable` alone | a table already named | accepted, keeps that table |
| `enable` alone | no table | `InvalidParameterCombination` |
| `disable` with an id | any | `InvalidParameterCombination` |
| `disable` with an unknown id | any | the combination is reported, not the lookup |

The rule underneath: **the pair is judged against the gateway as it stands, not against the request
alone.** The flag falls back to the stored one when the request omits it, which is what makes an id
on its own legal while enabled and a conflict while disabled — and why the error text quotes the
stored flag rather than anything the caller sent. Enabling without an id now keeps the table
already named instead of dropping it, an id on its own marks its table, and the lookup still runs
after the combination check, matching the order AWS reports them in.

Two more behaviours came out of the same pass, both already correct and now covered by tests rather
than left to chance: removals apply before additions, so a CIDR added and removed in one call
survives, and removing a CIDR that is not present is accepted rather than rejected.

Thanks for the persistence across all four. The first three each pointed at something genuine; this
one also pointed at the fact that I was fixing cases instead of the shape, which was the more
useful finding.

ck-michals:
`native / sdk-test-awscli` was failing on my own new case, and rightly so.

`transitGatewayCidrBlocks` is a plain string list, so AWS emits `<item>10.100.0.0/16</item>`. I
wrapped the value in a `cidrBlock` element, which the CLI parsed as an empty string — create,
describe and modify alike. Fixed in b9c7ebc6, with the bats case now reading the list back through
describe as well as modify.

ck-michals:
Re-checked the branch against what part one of #2308 actually promises — backing
`aws_ec2_transit_gateway` — and found a gap I had missed. `CreateTags`/`DeleteTags` on a `tgw-` id
updated the tag store but not the gateway, because `updateResourceTags` had no branch for it:
`DescribeTags` saw the change, `DescribeTransitGateways` did not. A provider tag update would never
converge. Fixed in 2d893930, with route tables given the same branch so part three inherits it.

The CLI cases now also cover the tag round trip and CIDR blocks supplied at creation. Both were
previously asserted only against shapes this branch produced itself, which is how the `cidrBlock`
mistake got through.

One thing left deliberately: `DescribeTransitGateways` ignores `MaxResults` and `NextToken` and
returns everything, matching the other EC2 describes here. Happy to implement pagination if you
would rather it be exact.

pgermosen (approved):
Exceptionally careful work, and the live-account verification really shows. Confirmed all four resolved greptile P1s in the code: `requireCoherentDefaultRouteTableChange` judges a change against the gateway's stored state rather than the request alone, `requireRouteTableOfGateway` rejects a route table owned by a different gateway, and `applyDefaultRouteTableChanges` correctly carries flag changes through to both the option ID and the route table's own default marker.

Cross-checked every stated option default against the real botocore `ec2` model — `AmazonSideAsn` 64512, `AutoAcceptSharedAttachments`/`SecurityGroupReferencingSupport` disabled, the rest enabled — all match exactly. Also verified the constructor-arity claim directly: the existing package-private `Ec2Service` constructor is genuinely untouched, with the two new transit-gateway backends added via a new overload instead, matching the #2103/#2106 lesson.

84/84 targeted tests pass, `docs-check` clean, no conflicts.

LGTM, approving.

PR #2331, opened 2026-08-15
fix(elasticache): store the user Engine instead of hardcoding redis

### LINKED ISSUE #2290: [BUG] elasticache: CreateUser ignores Engine and every user response reports redis/6.0
## Service

ElastiCache

## AWS API Action

`CreateUser`, `ModifyUser`, `DescribeUsers`, `DeleteUser`

## Expected behavior

`CreateUser` takes a required `Engine` parameter (`redis` or `valkey`), stores it, and every
response that serializes a user echoes the engine that was actually requested:

```xml
<User>
  <UserId>app-user</UserId>
  <Engine>valkey</Engine>
  ...
</User>
```

## Actual behavior

`Engine` is never read from the request and never stored. `ElastiCacheQueryHandler.userXml`
emits a literal:

```java
.elem("Engine", "redis")
.elem("MinimumEngineVersion", "6.0")
```

`ElastiCacheUser` has no `engine` field, and `handleCreateUser` parses only `UserId`,
`UserName`, `AccessString` and `AuthenticationMode.Type` — `Engine` is dropped on the floor.

Because `userXml` is shared by `CreateUser`, `DescribeUsers`, `ModifyUser` and `DeleteUser`,
all four report `redis` regardless of what was created.

**Impact.** A `aws_elasticache_user` declared with `engine = "valkey"` applies successfully and
then reads back as `redis`, so `terraform plan` reports a diff on every run and never converges.
`MinimumEngineVersion` is hardcoded the same way.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 elasticache create-user \
  --user-id app-user --user-name app --engine valkey \
  --access-string "on ~* +@all" --authentication-mode Type=no-password-required

aws --endpoint-url http://localhost:4566 elasticache describe-users --user-id app-user
# Engine is "redis"; expected "valkey"
```

Terraform equivalent — `apply` succeeds, the next `plan` is not empty:

```hcl
resource "aws_elasticache_user" "example" {
  user_id       = "app-user"
  user_name     = "app"
  engine        = "valkey"
  access_string = "on ~* +@all"

  authentication_mode {
    type = "no-password-required"
  }
}
```

## Suggested fix

1. Add an `engine` field to `ElastiCacheUser` (and `minimumEngineVersion` if it should also be
   modelled rather than hardcoded).
2. Read `Engine` in `handleCreateUser` and thread it through `ElastiCacheService.createUser`.
3. Emit `u.getEngine()` in `userXml` instead of the literal.

Defaulting a missing `Engine` to `redis` keeps existing behavior for callers that omit it.

## Environment

## AWS Documentation

https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateUser.html

soreavis:
## Summary

`Engine` was never read from a CreateUser request and `ElastiCacheUser` had no field for it; `userXml` emitted a literal `redis`, and since that serializer is shared by `CreateUser`, `DescribeUsers`, `ModifyUser` and `DeleteUser`, all four reported `redis` regardless of what was created — the `aws_elasticache_user` diff-loop from the issue.

This is the three-step fix the issue suggests — an `engine` field on the model, threaded through `createUser`, and `u.getEngine()` in `userXml` — plus the three places the new attribute immediately matters. `ModifyUser` also accepts `Engine`, which AWS models as an optional parameter on that action ("Modifies the engine listed for a user"), and validates it before the first setter: the storage backends hand back the live stored object, so a password change applied before validation would survive the 400 — a rejected request must leave the user untouched, and a test pins exactly that. `DescribeUsers` honors its documented `Engine` filter, which was inert while every user was `redis` and would silently return wrong results now that they aren't — only that one parameter; the action's `Filters`/`Marker`/`MaxRecords` stay out of scope, since nothing in this change makes them meaningful. And a missing or blank `Engine` defaults to `redis` on create (keeping the previous behavior for callers that omit it, as the issue asks — AWS marks the parameter required, so this leniency is Floci's own back-compat choice) and is a no-change on modify.

Values are validated case-insensitively against `redis`/`valkey` and stored lowercase — Floci's own tolerance, matching the handler's existing `equalsIgnoreCase` memcached check; AWS documents no case handling for the parameter — and anything else is rejected with `InvalidParameterValue` 400, the error CreateUser documents for an invalid value. `MinimumEngineVersion` deliberately stays at `6.0`: that is the only value AWS documents for the element (its description reads "The minimum engine version required, which is Redis OSS 6.0" on both actions), and no published response sample for a valkey user turned up to say otherwise, so inventing a valkey-specific value would bake an unverifiable claim into the emulator's contract — easy to revisit if someone captures a real response. Users persisted before the field existed deserialize with the `redis` default, which is what they were reporting anyway, and a test pins that round-trip through the same deserialization configuration the persistent backends use (`ObjectMapper` plus `JavaTimeModule`), against a full pre-engine record.

Every guard is pinned by a named test — the parameter threading on both actions (handler and service side), the validation, the case-folding on create and on the filter, the redis default, the blank-input handling, the validate-before-mutate ordering and its accept path, the DescribeUsers filter and its UserId-wins precedence, the deliberate 6.0, and the legacy-deserialization default were each mutation-tested individually. The one exception is the `users.put` in `modifyUser`: no test can fail on it under the in-memory storage backend, which hands back the live object, so it is stated here rather than covered by a test that cannot fail.

Closes #2290

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the CreateUser and ModifyUser API references: `Engine` is required on CreateUser with values `valkey`/`redis`, optional on ModifyUser, `DescribeUsers` documents an optional `Engine` filter, and `InvalidParameterValue` (400) is in CreateUser's error list. One softening: AWS marks `Engine` required on CreateUser, but Floci's CreateUser accepted requests without it, so a missing value keeps the previous implicit redis default rather than turning existing callers' requests into 400s.

The incorrect behavior, concretely: `create-user --engine valkey` succeeded and `describe-users` reported `Engine: redis`, so a Terraform `aws_elasticache_user` with `engine = "valkey"` re-planned forever.

## Checklist

- [x] `./mvnw test` run locally: 10,157 tests; the only failures (112) are confined to 22 Docker-backed classes (API Gateway/WebSocket, Lambda, ECR, DocDB, Neptune, CloudFormation/SAM, ELBv2, SWF) — no Docker daemon on this machine, and none of those services is touched by this diff. Every ElastiCache suite is fully green.
- [x] New or updated integration test added (plus SDK-based assertions in `compatibility-tests/sdk-test-java`'s ElastiCache suite: engine echo on the existing redis user and a valkey round-trip)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/ElastiCacheTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/ElastiCacheTest.java
@@ -155,6 +155,7 @@ class ElastiCacheTest {
 
         assertThat(response.userId()).isEqualTo(userId);
         assertThat(response.userName()).isEqualTo(userName);
+        assertThat(response.engine()).isEqualTo("redis");
         assertThat(response.authentication().typeAsString()).isEqualTo("password");
         assertThat(response.authentication().passwordCount()).isEqualTo(1);
         userCreated = true;
@@ -168,7 +169,8 @@ class ElastiCacheTest {
         var response = elasticache.describeUsers(DescribeUsersRequest.builder().build());
 
         assertThat(response.users())
-                .anyMatch(user -> user.userId().equals(userId) && user.userName().equals(userName));
+                .anyMatch(user -> user.userId().equals(userId) && user.userName().equals(userName)
+                        && "redis".equals(user.engine()));
     }
 
     @Test
@@ -271,6 +273,27 @@ class ElastiCacheTest {
 
     @Test
     @Order(11)
+    void valkeyUserRoundTripEchoesEngine() {
+        String valkeyUserId = TestFixtures.uniqueName("ec-valkey-user");
+        var created = elasticache.createUser(CreateUserRequest.builder()
+                .userId(valkeyUserId)
+                .userName(TestFixtures.uniqueName("ec-valkey-name"))
+                .engine("valkey")
+                .accessString("on ~* +@all")
+                .authenticationMode(AuthenticationMode.builder()
+                        .type(InputAuthenticationType.NO_PASSWORD_REQUIRED)
+                        .build())
+                .build());
+        assertThat(created.engine()).isEqualTo("valkey");
+
+        var deleted = elasticache.deleteUser(DeleteUserRequest.builder()
+                .userId(valkeyUserId)
+                .build());
+        assertThat(deleted.engine()).isEqualTo("valkey");
+    }
+
+    @Test
+    @Order(12)
     void deleteReplicationGroupReleasesPortForReuse() {
         requireGroup();
 
--- a/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheQueryHandler.java
@@ -158,6 +158,7 @@ public class ElastiCacheQueryHandler {
         String userName = params.getFirst("UserName");
         String accessString = params.getFirst("AccessString");
         String authModeType = params.getFirst("AuthenticationMode.Type");
+        String engine = params.getFirst("Engine");
 
         if (userId == null || userId.isBlank()) {
             return AwsQueryResponse.error("InvalidParameterValue", "UserId is required.", AwsNamespaces.EC, 400);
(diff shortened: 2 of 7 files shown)


pgermosen (approved):
Thanks for the very thorough writeup and the careful validate-before-mutate handling in `modifyUser` — every AWS-compatibility claim here checked out exactly against the real botocore model, down to the `ModifyUser` doc string quote and the `InvalidParameterValueException` wire error code.

20/20 tests pass, `docs-check` clean, no conflicts.

LGTM, approving.

Issue #2298, opened 2026-08-14
[FEAT] route53: AssociateVPCWithHostedZone, DisassociateVPCFromHostedZone and ListHostedZonesByVPC


pgermosen:
## Service

Route 53

## API Action / Feature

`AssociateVPCWithHostedZone`, `DisassociateVPCFromHostedZone`, `ListHostedZonesByVPC`

## AWS Documentation

- https://docs.aws.amazon.com/Route53/latest/APIReference/API_AssociateVPCWithHostedZone.html
- https://docs.aws.amazon.com/Route53/latest/APIReference/API_DisassociateVPCFromHostedZone.html
- https://docs.aws.amazon.com/Route53/latest/APIReference/API_ListHostedZonesByVPC.html

## Why is this needed?

`Route53Controller` serves `/hostedzone`, `/hostedzone/{Id}`, `/hostedzone/{Id}/rrset`,
`/hostedzone/{Id}/dnssec`, `/hostedzonesbyname`, `/hostedzonecount`, health checks and tags —
but has no VPC-association endpoints at all.

`aws_route53_zone_association` is the standard way to attach a private hosted zone to additional
VPCs after creation, and it is the only way to do so across accounts. Without these actions the
resource fails outright on create, and `terraform destroy` cannot detach.

`ListHostedZonesByVPC` is the read path the provider uses to confirm an association still
exists, so a create-only implementation would still drift.

## Scope

Three endpoints on `Route53Controller`, mirroring the existing hosted-zone handlers:

- `POST /2013-04-01/hostedzone/{Id}/associatevpc` → `AssociateVPCWithHostedZone`
- `POST /2013-04-01/hostedzone/{Id}/disassociatevpc` → `DisassociateVPCFromHostedZone`
- `GET /2013-04-01/hostedzonesbyvpc` → `ListHostedZonesByVPC` (query params `vpcid`, `vpcregion`)

The first two return a `ChangeInfo`; `Route53Service` already has a `changeStore` and the
controller already has an `xmlChangeInfo(...)` builder to reuse. Associations can live on the
zone in the existing `zoneStore` (`StorageBackend<String, HostedZone>`), so no new storage
backend is needed.

Error behaviour worth pinning, per the API reference:

- associating a VPC with a **public** zone returns `InvalidVPCId` / a conflict, not success
- disassociating the **last** remaining VPC from a private zone returns
  `LastVPCAssociation` — a private zone must always have at least one

## Dependency

[...]

PR #2342, opened 2026-08-16
fix: preserve Route 53 private hosted-zone VPC associations

### LINKED ISSUE #2294: [BUG] route53: CreateHostedZone reads a PrivateZone element no client sends, so zones with a VPC are created public
## Service

Route 53

## AWS API Action

`CreateHostedZone`

## Expected behavior

Per the `CreateHostedZone` API reference, a zone is private **because the request carried a
`VPC` element**, not because of any flag. The request body looks like:

```xml
<CreateHostedZoneRequest xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
  <Name>internal.example.com.</Name>
  <CallerReference>abc-123</CallerReference>
  <VPC>
    <VPCId>vpc-0123456789abcdef0</VPCId>
    <VPCRegion>us-west-2</VPCRegion>
  </VPC>
  <HostedZoneConfig>
    <Comment>internal</Comment>
  </HostedZoneConfig>
</CreateHostedZoneRequest>
```

`HostedZoneConfig.PrivateZone` is a **response-only** field. AWS derives it from the presence of
`VPC` and echoes it back; a caller never sets it. The response is also expected to carry a
`<VPCs>` list containing the VPC the zone was created with.

## Actual behavior

`Route53Controller.createHostedZone` reads a top-level `PrivateZone` element:

```java
boolean privateZone = "true".equalsIgnoreCase(
        XmlParser.extractFirst(body, "PrivateZone", "false"));
```

No AWS SDK or the Terraform provider ever sends that element, and the `<VPC>` element that is
actually sent is ignored entirely. So **every zone is created public**, including one that
explicitly requested a VPC association.

`xmlHostedZone` then dutifully emits `<PrivateZone>false</PrivateZone>`, and the response
carries no `<VPCs>` element at all.

**Impact.** An `aws_route53_zone` with a `vpc {}` block applies successfully, then reads back
with `private_zone = false` and an empty `vpc` set. `terraform plan` reports a diff on every run
and never converges. Worse, the zone is silently the wrong kind — a private zone that answers as
public — so anything reasoning about zone visibility gets the wrong answer without an error.

## Reproduction

```hcl
resource "aws_route53_zone" "internal" {
  name = "internal.example.com"

  vpc {
    vpc_id = aws_vpc.example.id
  }
}
```

`terraform apply` succeeds; `terraform plan -detailed-exitcode` immediately afterwards exits 2
with a diff on `private_zone` and `vpc`.

Or directly:

```bash
aws --endpoint-url http://localhost:4566 route53 create-hosted-zone \
  --name internal.example.com --caller-reference ref-1 \
  --vpc VPCRegion=us-west-2,VPCId=vpc-0123456789abcdef0
# Config.PrivateZone is false; expected true. No VPCs in the response.
```

## Suggested fix

1. Parse the `<VPC>` element (`VPCId`, `VPCRegion`) in `createHostedZone`.
2. Set `privateZone = true` when it is present, and stop reading the `PrivateZone` element.
3. Record the VPC on the zone and emit `<VPCs>` in the `CreateHostedZone` and `GetHostedZone`
   responses so the provider reads back what it set.
4. Continue reading `Comment` from `HostedZoneConfig`.

Step 3 needs a VPC association list on `HostedZone` (`model/HostedZone.java` currently has
`privateZone` but no association field). That same list is what
`AssociateVPCWithHostedZone`/`DisassociateVPCFromHostedZone` will need, so it is worth adding
here even though those actions are a separate issue.

## Environment

## AWS Documentation

https://docs.aws.amazon.com/Route53/latest/APIReference/API_CreateHostedZone.html

mvanhorn:
## Summary

Add a reflection-safe Route 53 VPC association model and persist a list of associations on `HostedZone`, keeping the existing storage-backed zone lifecycle intact. Parse `VPCId` and `VPCRegion` from the optional `VPC` request member, pass that association through `Route53Service.createHostedZone`, and derive `privateZone` from its presence instead of accepting the response-only `PrivateZone` input. Serialize the association in the operation-specific AWS shapes: the singular `VPC` member on CreateHostedZone and the `VPCs` collection on GetHostedZone, while leaving public-zone responses and hosted-zone list responses unchanged.

`CreateHostedZone` currently looks for a response-only `PrivateZone` element and ignores the request's `VPC`, so SDK and Terraform requests always create public zones. The missing association is also absent from subsequent reads, causing Terraform state to drift on every plan. The issue is limited to correctly handling the existing CreateHostedZone and GetHostedZone REST XML contracts; the association-management actions tracked separately in #2298 remain out of scope. There are no prior closed attempts or open competing pull requests in the supplied issue evidence.

Closes #2294

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Route53Test.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Route53Test.java
@@ -1,0 +1,69 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.route53.Route53Client;
+import software.amazon.awssdk.services.route53.model.CreateHostedZoneRequest;
+import software.amazon.awssdk.services.route53.model.CreateHostedZoneResponse;
+import software.amazon.awssdk.services.route53.model.DeleteHostedZoneRequest;
+import software.amazon.awssdk.services.route53.model.GetHostedZoneRequest;
+import software.amazon.awssdk.services.route53.model.GetHostedZoneResponse;
+import software.amazon.awssdk.services.route53.model.VPC;
+import software.amazon.awssdk.services.route53.model.VPCRegion;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@DisplayName("Route 53")
+class Route53Test {
+
+    private static Route53Client route53;
+
+    @BeforeAll
+    static void setup() {
+        route53 = TestFixtures.route53Client();
+    }
+
+    @AfterAll
+    static void cleanup() {
+        if (route53 != null) {
+            route53.close();
+        }
+    }
+
+    @Test
+    void createAndGetPrivateHostedZonePreservesVpcAssociation() {
+        VPC vpc = VPC.builder()
+                .vpcId("vpc-sdk-private")
+                .vpcRegion(VPCRegion.US_WEST_2)
+                .build();
+        String zoneId = null;
+
+        try {
+            CreateHostedZoneResponse created = route53.createHostedZone(CreateHostedZoneRequest.builder()
+                    .name(TestFixtures.uniqueName("private-zone") + ".example.com")
+                    .callerReference(TestFixtures.uniqueName("private-zone-ref"))
+                    .vpc(vpc)
+                    .build());
+            zoneId = created.hostedZone().id();
+
+            assertThat(created.hostedZone().config().privateZone()).isTrue();
+            assertThat(created.vpc().vpcId()).isEqualTo(vpc.vpcId());
+            assertThat(created.vpc().vpcRegion()).isEqualTo(vpc.vpcRegion());
+
+            GetHostedZoneResponse fetched = route53.getHostedZone(
+                    GetHostedZoneRequest.builder().id(zoneId).build());
+
+            assertThat(fetched.hostedZone().config().privateZone()).isTrue();
+            assertThat(fetched.vpcs()).singleElement().satisfies(association -> {
+                assertThat(association.vpcId()).isEqualTo(vpc.vpcId());
+                assertThat(association.vpcRegion()).isEqualTo(vpc.vpcRegion());
+            });
+        } finally {
+            if (zoneId != null) {
+                route53.deleteHostedZone(DeleteHostedZoneRequest.builder().id(zoneId).build());
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


pgermosen:
@mvanhorn Checked this against the linked issue (#2294) and dug for anything beyond what's already flagged.

**Matches the issue well, and in one place is actually more accurate than the issue's own suggestion**: the issue's step 3 says to emit `<VPCs>` on both `CreateHostedZone` and `GetHostedZone` responses, but real AWS's `CreateHostedZoneResponse` actually carries a *singular* `<VPC>` element (confirmed via `javap` against the real SDK model), which is what this PR correctly does — `GetHostedZoneResponse` is the one that's a list. Good catch going with the real shape over the issue text.

Three things need fixing before this is mergeable:

1. **Live CI failure, confirmed reproducible locally**: `native / sdk-test-java` fails to compile. The new compat test calls `fetched.vpcs()`, but the real SDK method is `vpCs()` (odd mixed-case codegen artifact for the "VPCs" plural). Confirmed via `javap` against the actual pinned `route53-2.44.14.jar`, and reproduced the identical compile error with `mvn test-compile` locally.

2. **Missing validation** (this is greptile's flagged P1, still unaddressed — one commit total, no follow-up): `parseVpcAssociation` accepts an incomplete `<VPC>` element (missing `VPCId` or `VPCRegion`) and still marks the zone private with a half-populated association. AWS's `CreateHostedZone` reference states both fields are required together whenever `VPC` is present — this should be rejected instead.

3. **New finding**: `Route53IntegrationTest` now has two methods at `@Order(3)` — the new `createHostedZone_withVpcReturnsPrivateZoneAndAssociation` and the pre-existing `listHostedZones_includesCreatedZone`. Not observed to actually fail (the new test cleans up after itself), but their relative execution order is technically undefined per JUnit's contract — worth a trivial renumber.

No other gaps found — I checked for other callers of the changed `createHostedZone` signature (none exist outside the `route53` package), and confirmed `ListHostedZones`/the base `HostedZone` shape genuinely carry no VPC field in real AWS, so the PR's stated scope (Create/GetHostedZone only) is accurate, not incomplete.

pgermosen (approved):
All three issues from the earlier review are genuinely fixed — confirmed by re-reading the code and testing directly:

1. `vpCs()` accessor: fixed, and I compiled `compatibility-tests/sdk-test-java` against the real pinned SDK jar to confirm.
2. VPCId/VPCRegion co-validation: fixed, correctly rejects a half-populated `<VPC>` element.
3. Duplicate `@Order(3)`: fixed, all test orders renumbered.

Route53 suite passes clean (21/21), both on the branch and on the actual merged-with-main tree.

One follow-up, non-blocking: the VPCId/VPCRegion validation added in point 2 doesn't have a regression test covering it yet — would you mind adding one in a follow-up so a future refactor can't silently reintroduce the half-populated-association bug? Approving as-is either way.

PR #2343, opened 2026-08-16
fix(core): extract shared pagination helper, reject maxResults=0

### LINKED ISSUE #2339: [BUG] pagination: maxResults=0 silently defaults instead of failing validation (repo-wide)
## Summary

Multiple REST-style list endpoints treat a `maxResults` query param of `0` as "not specified"
and silently substitute the default page size, instead of rejecting it. Real AWS API models
declare `MaxResults` with `min: 1` (verified against botocore's `kafka` service model for MSK's
`ListConfigurations`: `{"type": "integer", "min": 1, "max": 100}`), so `maxResults=0` is
out-of-range input a real client would get a validation error for — not a silent full page.

## Where

The `paginate()` helper originates in `BedrockAgentCoreGatewayService` (`list()`/`listTargets()`,
backing `GET /gateways/` and `GET /gateways/{id}/targets`) and was mirrored into
`MskService.listConfigurations` (added in #2336, backing `GET /v1/configurations`). Both share
the same line:

```java
int limit = maxResults > 0 ? maxResults : MAX_PAGE;
```

which conflates "the query param was omitted" (the controller defaults the unset case to `0`)
with "the caller explicitly passed `maxResults=0`".

## Why is this needed?

A client that explicitly sends `maxResults=0` — a real, if unusual, mistake — currently gets a
full page of results back with `200 OK` instead of the validation error AWS would return,
silently masking the client-side bug instead of surfacing it. Flagged during review of #2336
(https://github.com/floci-io/floci/pull/2336#issuecomment-5305181891); left out of that PR since
it's a pre-existing pattern shared by another service, not something newly introduced there.

## Scope

- Change the `0`-as-sentinel convention so "omitted" and "explicitly zero" are distinguishable —
  e.g. controllers pass `null`/a boxed `Integer` (or a sentinel like `-1`) for "not specified"
  instead of `0`, and `paginate()` rejects `maxResults == 0` with the same error it already
  throws for out-of-range values.
- Apply the fix to both `BedrockAgentCoreGatewayService` and `MskService`.
- Add a regression test for `maxResults=0` at both the service and REST layer, for each.

- [x] Yes
- [ ] No

### LINKED ISSUE #2340: [CHORE] core: extract shared pagination helper (ListResult<T> + paginate()) out of per-service duplication
## Summary

`ListResult<T>` (a `record(List<T> items, String nextToken)`) and its accompanying
`paginate()`/`encodeToken()`/`decodeToken()` methods now exist byte-for-byte identically in two
places:

- `io.github.hectorvent.floci.services.bedrockagentcorecontrol.model.ListResult` +
  the private `paginate()`/`encode()`/`decode()` methods in `BedrockAgentCoreGatewayService`
- `io.github.hectorvent.floci.services.msk.model.ListResult` + the equivalent private methods in
  `MskService` (added in #2336)

## Why is this needed?

Every future `List*` action that needs AWS-style `maxResults`/`nextToken` pagination will
otherwise keep re-copying this same ~40-line block into a third, fourth, fifth service. There's
no shared pagination utility in `core/common` today, so "reuse existing Floci patterns" for
pagination currently means "copy-paste the last service that had it." Flagged during review of
#2336 (https://github.com/floci-io/floci/pull/2336#issuecomment-5305181891).

## Scope

- Add a generic pagination helper under `core/common` — e.g. a shared `ListResult<T>` record plus
  something like `Pagination.paginate(List<T> all, Function<T, String> cursorOf, int maxResults,
  String nextToken, int maxPage)` — reusable across services.
- Migrate `BedrockAgentCoreGatewayService` and `MskService` to use it, removing their private
  duplicates.
- Keep the per-call behavior each service already has: same opaque-cursor scheme (last item's
  sort key, base64 URL-encoded), but let each caller supply its own error code/message on
  validation failure rather than forcing one error code across services — different AWS APIs use
  different codes for the same class of 400.
- Worth landing after (or together with) #2339, since fixing the `maxResults=0` gap once in the
  new shared utility is cleaner than fixing it twice in the existing duplicates first.

- [x] Yes
- [ ] No

pgermosen:
## Summary

Closes #2339, closes #2340.

While scoping #2340 I only found the duplication in two services (`BedrockAgentCoreGatewayService`
and `MskService`) — but the `ListResult<T>` type it names was already shared *within*
`bedrockagentcorecontrol` (imported by every controller/service in that package), which is what
threw off the original count. The actual `paginate()`/`encode()`/`decode()` *logic* — not just the
type — was duplicated **five** times: `BedrockAgentCoreGatewayService`, `BedrockAgentCoreIdentityService`,
`BedrockAgentCoreMemoryService`, all three list methods on `BedrockAgentCoreControlService`
(`listAgentRuntimes`/`listAgentRuntimeVersions`/`listEndpoints`), and `MskService`. A "shared
pagination helper" that only migrated two of five duplicates and left three behind wouldn't
actually be done, so this migrates all five.

- **`core/common/Pagination`** + **`core/common/PaginatedResult<T>`**: the shared cursor-pagination
  logic (opaque cursor = last item's sort key, base64 URL-encoded) and `parseMaxResults(String,
  String)` for controllers. Both per-service `model/ListResult.java` files are deleted.
- **#2339's fix baked in from the start**: every duplicate collapsed "`maxResults` omitted" and
  "`maxResults=0`" onto the same primitive `int`, defaulting both to a full page — AWS declares
  `MaxResults` with `min: 1` (confirmed against botocore's real `kafka` service model), so `0` is
  real out-of-range input, not a synonym for "give me the default." `Pagination.paginate`/
  `parseMaxResults` take a boxed `Integer` so the two are distinguishable; every migrated service
  now rejects `0` with whatever error code it already used (`ValidationException` for the
  `bedrockagentcorecontrol` services, `BadRequestException` for MSK) — fixed once in the shared
  utility, not five times over.
- **Carries over MSK's `@QueryParam Integer` → 404 fix from #2336**: RESTEasy Reactive's default
  handling of a non-numeric value for an `Integer`-typed `@QueryParam` is a bare 404 (the request
  never reaches the resource method), not an AWS-shaped 400. Every migrated REST controller
  (`BedrockAgentCoreGatewayController`, `BedrockAgentCoreControlController`,
  `BedrockAgentCoreMemoryController`) now binds `maxResults` as `String` and calls
  `Pagination.parseMaxResults`, closing the same gap MSK already had fixed.
- `BedrockAgentCoreIdentityController` is JSON-body-driven (`POST /identities/ListWorkloadIdentities`),
  not `@QueryParam`, so it gets the narrower fix: `Integer maxResults = req.hasNonNull("maxResults")
  ? req.get("maxResults").asInt() : null` instead of defaulting the absent case to `0`.

Net: **-123 lines** while adding the `maxResults=0` fix to five call sites instead of one.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified `MaxResults`'s `min: 1` constraint against botocore's real `kafka` service model, same
approach as #2336. Live-verified every migrated endpoint against a running `./mvnw quarkus:dev`
instance, not just the automated suite:

- **Gateway** (`GET /gateways/?maxResults=`): create 3, page with `maxResults=2`, confirm
  `nextToken` continuation; `maxResults=0` → 400; `maxResults=abc` → 400 (not 404).
- **Memory** (`POST /memories/?maxResults=`): `maxResults=0` → `"maxResults must be between 1 and
  100"`; `maxResults=abc` → `"maxResults must be an integer."`.
- **Identity** (`POST /identities/ListWorkloadIdentities`, JSON-body driven, `MAX_PAGE=20` not
  100): `{"maxResults": 0}` → `"maxResults must be between 1 and 20"` — confirms the per-service
  `maxPage` parameter threads through correctly, not just the two most-common ones.
- **MSK** (`aws kafka list-configurations`): confirmed botocore's own client-side parameter
  validation already refuses `MaxResults=0` before the request is even sent
  (`"valid min value: 1"`) — independent confirmation the fix matches AWS's actual contract, not
  just floci's interpretation of it.

## Checklist

- [x] `./mvnw test` passes locally (93 tests across the 10 affected suites, 0 failures)


hectorvent:
Thank you, this is a clean de-duplication and the reasoning in the javadoc is worth keeping.

I checked the `min=1` claim per operation rather than taking it once, since one helper now serves two services with separate models. It holds everywhere: kafka's four list operations are all `min=1 max=100`, and bedrock-agentcore-control's are `min=1` with maxima of 20, 100 and 1000. The error codes line up too, `ValidationException` on the bedrock operations and `BadRequestException` on `ListConfigurations`. Worth calling out that `BedrockAgentCoreIdentityService` uses `MAX_PAGE = 20` to match `ListWorkloadIdentities`, where copying 100 from its siblings would have been the easy mistake.

The controller change from `Integer` to `String` plus `parseMaxResults` is the subtle half, and it fixes a real 404 that should have been a 400.

**(follow-up, separate PR)** `BedrockAgentCoreGatewayService.MAX_PAGE` is 100 but `ListGateways` and `ListGatewayTargets` both declare `max=1000`, so a valid larger page is refused. That predates this PR, it just becomes easy to fix now that the bound is one constant.

No blockers from my side.

hectorvent (approved):

PR #2345, opened 2026-08-16
fix(cloudformation): physically delete nested stacks and removed resources on update

### LINKED ISSUE #2254: [BUG] cloudformation: resources removed or condition-disabled on UpdateStack are never deleted
## Service

CloudFormation

## AWS API Action

UpdateStack

## Expected behavior

When a stack update makes a previously provisioned resource no longer active — either because its resource-level `Condition` flips from true to false, or because the resource is removed from the template entirely — real AWS deletes the underlying resource during the `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS` phase. After the update:

- the resource no longer appears in `DescribeStackResources`, and
- the underlying resource (SQS queue, log group, table, …) no longer exists.

`DeletionPolicy: Retain` (and `RetainExceptOnCreate`) should be honoured as the opt-out.

Example: a queue gated by a condition that becomes false on update is deleted, and `aws sqs list-queues` no longer returns it.

## Actual behavior

The resource is skipped during provisioning (correct), but it is never deleted:

- `DescribeStackResources` keeps returning its `LogicalResourceId`.
- The underlying resource stays alive and queryable.

Root cause: `CloudFormationService.topologicalSort` (around `CloudFormationService.java:1108-1114`) excludes condition-false / removed logical IDs from `activeIds`, so `executeTemplate` never iterates them. But `executeTemplate` only ever *adds/updates* the resources it iterates — there is no cleanup pass that diffs the previously stored `stack.getResources()` against the now-active set and deletes what dropped out. So the `StackResource` record and the real resource both survive.

```
$ aws --endpoint-url http://localhost:4566 cloudformation describe-stack-resources \
      --stack-name cond-update-repro \
      --query 'StackResources[].[LogicalResourceId,ResourceStatus]' --output text
OptionalQueue   CREATE_COMPLETE   <- still present after the condition went false

$ aws --endpoint-url http://localhost:4566 sqs list-queues
{
    "QueueUrls": [
        "http://localhost:4566/000000000000/optional-queue-repro"   <- still alive
    ]
}
```

## Reproduction

```bash
sleep 10

cat > /tmp/cond-update.json <<'JSON'
{
  "Parameters": {
    "CreateQueue": { "Type": "String", "Default": "true" }
  },
  "Conditions": {
    "ShouldCreate": { "Fn::Equals": [ { "Ref": "CreateQueue" }, "true" ] }
  },
  "Resources": {
    "OptionalQueue": {
      "Type": "AWS::SQS::Queue",
      "Condition": "ShouldCreate",
      "Properties": { "QueueName": "optional-queue-repro" }
    }
  }
}
JSON

# 1. Create with the condition true — the queue is provisioned.
    --stack-name cond-update-repro \
    --template-body file:///tmp/cond-update.json \
    --parameters ParameterKey=CreateQueue,ParameterValue=true
sleep 3

# 2. Update with the condition false — real AWS deletes the queue.
aws --endpoint-url http://localhost:4566 cloudformation update-stack \
    --stack-name cond-update-repro \
    --template-body file:///tmp/cond-update.json \
    --parameters ParameterKey=CreateQueue,ParameterValue=false
sleep 3

# Expected: OptionalQueue absent from stack resources, and no queue listed.
# Actual:   OptionalQueue still CREATE_COMPLETE, and the queue still exists.
aws --endpoint-url http://localhost:4566 cloudformation describe-stack-resources \
    --stack-name cond-update-repro \
    --query 'StackResources[].[LogicalResourceId,ResourceStatus]' --output text
aws --endpoint-url http://localhost:4566 sqs list-queues
```

The same leak occurs if, instead of flipping the condition, the `OptionalQueue` resource is deleted from the template outright on the second deploy.

## Environment

- Floci version / image tag: `floci/floci:latest`

---

### Related

- #2211 — update leaks the **replaced** resource (logical ID kept, physical ID swapped). Same missing update-cleanup phase, different trigger.
- #2134 — update recreates unchanged named resources.
- #2168 — resource-level `Condition` ignored on **create** (this issue is the update-time counterpart: the condition is now honoured for skipping, but not for deleting).

Z9RTM:
## Summary

Physically delete nested stacks and removed resources on update.
Closes #2254

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

During a stack update, nested stacks and resources removed from the template were not physically deleted, leading to resource leaks. This fix ensures that these resources are properly removed.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
@@ -33,6 +33,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
@@ -734,7 +735,7 @@ public class CloudFormationService {
                             "AWS::CloudFormation::Stack",
                             "UPDATE_COMPLETE_CLEANUP_IN_PROGRESS", null);
                     // The committed template and new physical IDs must be durable before old
-                    // resources are deleted during replacement cleanup.
+                    // resources are deleted during post-update cleanup.
                     persistStack(stack);
                 }
                 // Both cleanup paths feed the single final status/reason writer.
@@ -933,7 +934,22 @@ public class CloudFormationService {
         if ("AWS::CloudFormation::Stack".equals(resource.getResourceType())) {
             Future<?> future = deleteStack(resource.getPhysicalId(), region, regionResolver.getAccountId());
             if (future != null) {
-                future.get();
+                try {
+                    future.get();
+                } catch (ExecutionException e) {
+                    Throwable cause = e.getCause();
+                    if (cause instanceof Exception ex) {
+                        throw ex;
+                    }
+                    throw e;
+                }
+            }
+            Stack child = resolveStack(resource.getPhysicalId(), region);
+            if (child != null && "DELETE_FAILED".equals(child.getStatus())) {
+                String reason = child.getStatusReason() != null
+                        ? child.getStatusReason()
+                        : "Nested stack deletion failed";
+                throw new IllegalStateException(reason);
             }
         } else {
             provisioner.delete(resource, region);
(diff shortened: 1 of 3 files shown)


hectorvent:
Thank you, and this is the right shape.

The thing I most wanted to check is where the cleanup sits relative to the failure guard, because deleting resources on an update that then fails is unrecoverable. It is correct: the `failedResource` guard returns at `:682`, and the cleanup block starts at `:731`, so a failed update rolls back and never reaches the deletion path. That matches the ordering established on #2255.

`DeletionPolicy: Retain` is respected too, via `skipRetainedResource` before the physical delete rather than a separate check, and only resources actually removed from the template or whose condition went false are touched.

Worth calling out the test balance: 70 lines of production code against 874 lines of tests, placed in the DeletionPolicy and resource-condition suites rather than a new isolated one. Those are precisely the two features that interact with this, which is where the risk was.

No blockers from my side.

PR #2346, opened 2026-08-16
feat(guardduty): add GuardDuty detector and organization configuration emulation

### LINKED ISSUE #2196: feat(guardduty): implement GuardDuty org enablement — blocked behind macie in LZA Organization stage
## What's wrong

floci has no `guardduty` service. `git log --all` confirms no implementation has ever existed on any branch.

AWS LZA's v2 Organization module runner includes a `guardduty` org-enablement module in the Control-Tower-aware module set. That module has **not yet been observed failing**, because `macie` runs alone in run order 1 and aborts the Organization stage before any later run order executes (see the Macie issue). This is therefore an inferred gap from the module set, not an observed error — but it will surface the moment Macie is implemented.

## Why it matters

Same severity as the Macie gap: it hard-fails an LZA pipeline stage, blocking `SecurityAudit` and `Deploy`. Implementing Macie alone will not unblock the pipeline — it will just advance the failure to this service. The three should be scoped together.

This became reachable only after the Control Tower service landed in the image: with a live `SUCCEEDED` landing zone returned from `POST /list-landingzones`, LZA advances into the Control-Tower-aware module set. Before that, these modules were never invoked and their absence was invisible.

## Suggested fix

Implement a minimal `guardduty` sufficient for LZA org enablement — enable/describe for the org, delegated administrator registration, and member enablement. LZA needs believable readback, not real detection behavior. The `organizations` service already models delegated administrators and can back the admin endpoints.

## Related

Blocked behind the Macie gap; both are part of the same three-service cascade in the Organization stage's Control-Tower-aware module set.

tdewitt:
## Summary

Adds a GuardDuty service: detector CRUD, detector features, organization
configuration, delegated-administrator accounts, and detector tagging via the
shared `/tags` routes. Mirrors the `rum` service pattern (REST JSON, JAX-RS
literal paths over the S3 templates, `AwsException` errors).

Closes #2196.

Scope notes:
- Organization semantics are readback-only: Floci has no Organizations service
  (contrary to a premise in #2196), so `adminAccountId` membership is not
  validated and nothing fans out to member accounts. Configuration is stored per
  calling account and echoed back — sufficient for Terraform's four
  `aws_guardduty_*` organization resources in single-account use. Member APIs
  (`CreateMembers` etc.) are left for a follow-up if LZA needs them.
- Feature lists and `additionalConfiguration` lists are returned in submitted
  order: they are ordered list blocks in the Terraform provider, and a
  reordered read-back forces resource replacement on every plan.
- A missing detector returns `BadRequestException` with the exact message the
  Terraform provider string-matches for not-found detection (GuardDuty has no
  404s); the delegated-admin "already disabled" message is likewise matched
  verbatim. Other error messages are descriptive rather than AWS's generic
  "invalid or out-of-range value" text.
- `dataSources` (deprecated) is omitted from GetDetector; the provider treats
  absence as "not configured".

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Wire shapes (routes, members, enums, status codes including the 204 tag
responses) pinned from the botocore `guardduty/2017-11-28` service model.
Verified against:

- AWS SDK for Java v2 (`sdk-test-java` BOM): the new `GuardDutyTest` drives the
  real client through the detector, organization-configuration, and
  delegated-admin lifecycles, including `BadRequestException` unmarshalling and
  epoch-seconds feature timestamps.
- AWS CLI v2: compat-terraform bats spot checks (`get-detector`,
  `describe-organization-configuration`).
- Terraform `hashicorp/aws ~> 6.0` and OpenTofu: full apply of
  detector + detector feature + organization configuration (+ feature), then a
  `plan -detailed-exitcode` asserting zero drift — the regression test for
  list ordering. Provider not-found handling verified against
  terraform-provider-aws source (`internal/service/guardduty/detector.go`).

## Checklist

### DIFF
--- a/compatibility-tests/compat-opentofu/main.tf
+++ b/compatibility-tests/compat-opentofu/main.tf
@@ -342,3 +342,47 @@ resource "aws_ses_active_receipt_rule_set" "compat" {
 output "ses_rule_set_name" {
   value = aws_ses_receipt_rule_set.compat.rule_set_name
 }
+
+# -- GuardDuty -----------------------------------------------------------------
+# Detector, per-feature configuration, and organization configuration mirror the
+# resource set an org security-baseline stack manages. additional_configuration
+# is an ordered list block: Floci must echo it back in submitted order or every
+# re-plan proposes a replacement.
+resource "aws_guardduty_detector" "compat" {
+  enable                       = true
+  finding_publishing_frequency = "SIX_HOURS"
+
+  tags = {
+    Environment = "compat-test"
+  }
+}
+
+resource "aws_guardduty_detector_feature" "runtime_monitoring" {
+  detector_id = aws_guardduty_detector.compat.id
+  name        = "RUNTIME_MONITORING"
+  status      = "ENABLED"
+
+  additional_configuration {
+    name   = "ECS_FARGATE_AGENT_MANAGEMENT"
+    status = "ENABLED"
+  }
+
+  additional_configuration {
+    name   = "EC2_AGENT_MANAGEMENT"
+    status = "ENABLED"
+  }
+
+  additional_configuration {
+    name   = "EKS_ADDON_MANAGEMENT"
+    status = "DISABLED"
+  }
+}
+
+resource "aws_guardduty_organization_configuration" "compat" {
+  detector_id                      = aws_guardduty_detector.compat.id
+  auto_enable_organization_members = "ALL"
+}
+
+output "guardduty_detector_id" {
+  value = aws_guardduty_detector.compat.id
+}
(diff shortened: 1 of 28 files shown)


hectorvent (approved):
Thank you, and this is a strong first contribution.

The registration is complete, which is the part new-service PRs most often miss: `ResolvedServiceCatalog`, `EmulatorConfig`, both `application.yml` files, and both docs pages. All eleven operations exist in the model, and the REST paths match the model's `requestUri` exactly, which matters more here than for a Query service because the URI is the routing.

One thing worth recording for whoever reads this next. The response fields are lowercase first (`createdAt`, `findingPublishingFrequency`, `serviceRole`) where the model's member names are PascalCase, which looks wrong at a glance. It is not: every one of those members carries a lowercase `locationName`, and you match it. I mention it only so nobody "fixes" it later against the member names.

Two small observations, neither worth changing. `GetDetector` also returns `detectorId`, which is not in the model's output members, and `dataSources` is omitted, which is fine since it is optional and deprecated in favour of `features`.

No blockers from my side.

PR #2347, opened 2026-08-16
feat(ec2): support transit gateway VPC attachments

ck-michals:
Implements VPC attachments: `CreateTransitGatewayVpcAttachment`,
`DescribeTransitGatewayVpcAttachments`, `DescribeTransitGatewayAttachments`,
`ModifyTransitGatewayVpcAttachment` and `DeleteTransitGatewayVpcAttachment`.

Part two of #2308, on top of the gateway itself in #2329.

The two describes are separate shapes rather than one being a superset of the other: the
VPC-specific form carries the subnets and options, while the resource-agnostic form drops both,
types the VPC as a resource, and is the only place the route table association appears. Both are
covered by tests, since a provider uses both.

The surface was probed before implementing rather than after review, so the behaviour below is
observed rather than inferred.

| | observed |
|---|---|
| attachment option defaults | `dnsSupport` and `securityGroupReferencingSupport` enabled, `ipv6Support` and `applianceModeSupport` disabled |
| compared with the gateway | `securityGroupReferencingSupport` is **enabled** on an attachment and **disabled** on the gateway that owns it |
| association | made only when the gateway carries `defaultRouteTableAssociation` enabled, and an existing one **survives** the gateway later disabling that default |
| a VPC attached twice | `DuplicateTransitGatewayAttachment` |
| two subnets in one zone | `DuplicateSubnetsInSameZone` |
| a subnet from another VPC | `InvalidSubnetID.NotFound` — reported missing rather than mismatched |
| removing a subnet that is not attached | `InvalidSubnetID.NotFound`, with its own wording |
| removing the last subnet | `InsufficientSubnetsException` |
| a malformed attachment id | `InvalidTransitGatewayAttachmentID.Malformed`, and the message does not echo the id |
| deleting a gateway while attached | `IncorrectState`, naming the attachments |
| echoes | modify omits the `tagSet`; delete omits the `tagSet` and the subnets |

Two of those are worth pointing at directly. The `securityGroupReferencingSupport` default flips
between the gateway and its attachment, which is a one-word difference that reading the reference
would not have caught. And the association surviving a later disable is the opposite of what the
gateway's own option coherence rules suggest — the association belongs to the attachment from
creation, and only later attachments are affected.

## Two changes reach outside the new surface

Both deliberate, both flagged here rather than left in the diff:

**`DeleteTransitGateway` now refuses while attachments exist**, reporting `IncorrectState` and
naming them. Part one could not have hit this, since nothing could attach then.

**`parseTagsForResource` now reads `TagSpecifications` as well as `TagSpecification`.**
`CreateTransitGatewayVpcAttachment` declares no `locationName` on that member and so serializes
the plural, unlike `CreateTransitGateway`, `CreateVpc` and the other neighbouring creates. Every
tag the AWS CLI sent to this action was being dropped. The change only adds a spelling, so actions
using the singular are untouched — and the CLI compatibility case now covers the round trip.

## Deviations

State is reported settled rather than transitional, as for the gateway: creation returns
`available` and deletion `deleted`, skipping `pending`, `modifying` and `deleting`.

`Ipv6Support` is accepted without checking that the attached subnets carry IPv6 CIDRs, which real
AWS rejects with `InvalidParameterCombination`. Floci does not model subnet IPv6 allocation, and
inventing that state to enforce the rule seemed worse than documenting the gap.

## Tests

- `Ec2ServiceTest`: option defaults, the association in both directions and its survival of a
  later disable, both owner ids, subnet ownership and one-per-zone, the subnet add and remove
  paths including both refusals, creation without subnets, malformed ids, and tag typing.
- `Ec2TransitGatewayVpcAttachmentIntegrationTest`: the Query wire shapes end to end, both describe
  forms, and the trimmed echoes on modify and delete.
- `ec2.bats`: the attachment round trip through both describes, modify, and the two refusals, all
  parsed by the AWS CLI rather than by hand-written assertions.

Full suite: 10449 tests, 0 failures. One error, `NeptuneOpenCypherIntegrationTest`, which
reproduces identically on unmodified `main` on this machine and is unrelated — the diff is EC2
only.

### DIFF
--- a/compatibility-tests/sdk-test-awscli/test/ec2.bats
+++ b/compatibility-tests/sdk-test-awscli/test/ec2.bats
@@ -9,14 +9,23 @@ setup() {
     SG_SOURCE_ID=""
     SG_TARGET_ID=""
     TRANSIT_GATEWAY_ID=""
+    TGW_ATTACHMENT_ID=""
+    TGW_VPC_ID=""
 }
 
 teardown() {
     if [ -n "$PREFIX_LIST_ID" ]; then
         aws_cmd ec2 delete-managed-prefix-list --prefix-list-id "$PREFIX_LIST_ID" >/dev/null 2>&1 || true
     fi
+    if [ -n "$TGW_ATTACHMENT_ID" ]; then
+        aws_cmd ec2 delete-transit-gateway-vpc-attachment \
+            --transit-gateway-attachment-id "$TGW_ATTACHMENT_ID" >/dev/null 2>&1 || true
+    fi
     if [ -n "$TRANSIT_GATEWAY_ID" ]; then
         aws_cmd ec2 delete-transit-gateway --transit-gateway-id "$TRANSIT_GATEWAY_ID" >/dev/null 2>&1 || true
+    fi
+    if [ -n "$TGW_VPC_ID" ]; then
+        aws_cmd ec2 delete-vpc --vpc-id "$TGW_VPC_ID" >/dev/null 2>&1 || true
     fi
     for sg in "$SG_TARGET_ID" "$SG_SOURCE_ID"; do
         if [ -n "$sg" ]; then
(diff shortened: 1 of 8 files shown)


ck-michals:
All four findings fixed, the last two in ffccd48e.

**Uniqueness was non-atomic.** The duplicate scan and the store now share a lock. Reproduced with
it removed: `expected: <1> but was: <8>`, every thread creating an attachment.

**Attachment references could go stale.** Checked against the real API rather than picking an error
code — both deletes refuse, with the same code and shape of message:

```
DependencyViolation: The subnet 'subnet-…' has dependencies and cannot be deleted.
DependencyViolation: The vpc 'vpc-…' has dependencies and cannot be deleted.
```

`deleteVpc` and `deleteSubnet` now refuse the same way, and a subnet of the attached VPC which the
attachment does not use still deletes, so the guard is no wider than the dependency it protects.

**Gateway lookup escaped the lock**, and **dependency checks remained non-atomic.** Both correct,
and both follow from the first fix being too narrow: `lockFor` is striped, so a gateway's monitor
never guarded a VPC or subnet delete regardless, and the gateway was resolved before the locked
section anyway.

Everything an attachment depends on — the gateway, the VPC, the subnets, and the uniqueness rule —
now resolves and writes under one region-scoped monitor, and deleting a gateway, VPC or subnet
takes that same monitor across its dependency check and its delete. Where a gateway lock is also
taken it is acquired inside this one, so the two cannot interleave in opposite orders.

The test races a create against each delete and asserts the invariant rather than a winner: the
store may never hold an attachment naming a gateway, VPC or subnet that is gone. It reproduces both
holes — with the create-side lock removed, `attachment outlived its gateway` fails on the first
trial.

Separately, in 73119e8a: the first concurrency harness ran forty trials of eight threads, and that
load disturbed `PostgresProtocolHandlerTest` and `ElbV2HealthCheckerTest` locally, neither of which
this branch touches. The race shows on the first trial without the lock, so ten trials of six keeps
the evidence and leaves those alone.

ck-michals:
Fixed in 084a6710, and extended past the case reported: the gateway and route table tag branches
had the same read-modify-write hole, so all three moved under the lock together.

Ordering turned out to matter as much as the lock. The deletes take the topology lock and then a
striped resource lock; my first attempt took the topology lock further down the stack, inside
`updateResourceTags`, with the resource lock already held — the opposite order, which deadlocks
against a concurrent delete. The race test caught it as a thread pool that never drained.
`createTags` and `deleteTags` now take the topology lock outermost for `tgw-` resources, and
`updateResourceTags` relies on its callers for it.

The test races tagging against deletion for both an attachment and a gateway, and asserts neither
comes back.

pgermosen (approved):
Verified all 5 Greptile P1 concurrency findings are genuinely fixed — diffed the original commit against the current state and confirmed the bugs were real (e.g. the initial createTransitGatewayVpcAttachment uniqueness check had zero synchronization). The fix design is a single region-wide attachmentTopologyLock, consistently taken as the outermost lock everywhere it coexists with a resource-specific lock; traced the nesting across every call site (create/modify/delete attachment, deleteTransitGateway, deleteVpc, deleteSubnet, createTags/deleteTags) and found no ordering inversions.

Ran the genuine multi-threaded regression test (concurrentAttachmentsForOneVpcLeaveExactlyOne, using ExecutorService/CountDownLatch to force real concurrent execution) 3x — passed consistently, no flakiness. Full targeted suite: Ec2ServiceTest 94/94, Ec2TransitGatewayVpcAttachmentIntegrationTest 6/6, 0 failures.

Clean merge against current main, docs-check passes, all CI green, no CODEOWNERS-protected paths.

PR #2348, opened 2026-08-16
feat(ec2): support transit gateway route tables, associations and routes

### LINKED ISSUE #2308: [FEAT] ec2: Transit Gateway and VPC attachments

ck-michals:
Closes #2308.

Part three of that issue, and the last of its three parts: the gateway landed in #2329, the
attachments in #2347, and this covers the route tables, associations, propagations and routes.
Rebased onto `main` now that #2347 has merged, so this is four commits of its own. Backs
`aws_ec2_transit_gateway_route_table` and the association, propagation and route resources beside
it.

Thirteen actions: `CreateTransitGatewayRouteTable`, `DescribeTransitGatewayRouteTables` and
`DeleteTransitGatewayRouteTable`; `AssociateTransitGatewayRouteTable`,
`DisassociateTransitGatewayRouteTable` and `GetTransitGatewayRouteTableAssociations`;
`EnableTransitGatewayRouteTablePropagation`, `DisableTransitGatewayRouteTablePropagation` and
`GetTransitGatewayRouteTablePropagations`; `CreateTransitGatewayRoute`,
`ReplaceTransitGatewayRoute`, `DeleteTransitGatewayRoute` and `SearchTransitGatewayRoutes`.

## Where the state lives

Associations are recorded on the attachment, because an attachment has at most one; there is no
second store that could disagree with it. Propagations are their own records, because an
attachment may propagate into several tables.

Propagated routes are **derived** when routes are searched, from each enabled propagation joined
to the attached VPC's CIDR blocks, rather than written down. Storing them would go stale the
moment a VPC's CIDR associations changed and would need invalidating from every VPC and attachment
path.

| | observed |
|---|---|
| a table asked for by name | neither default; only the one a gateway mints for itself is |
| associating an already-associated attachment | `Resource.AlreadyAssociated` — disassociate first, it does not move |
| association vs propagation states | association reports `associating`/`disassociating`; propagation reports the settled `enabled`/`disabled` at once |
| enabling propagation twice | `TransitGatewayRouteTablePropagation.Duplicate` |
| propagation's effect | the attached VPC's CIDR appears as a `propagated`/`active` route |
| a blackhole | a static route in the `blackhole` state, not a type, carrying no attachment |
| `ReplaceTransitGatewayRoute` | an **upsert**: replacing a destination the table never held writes it |
| deleting a table | `IncorrectState` while it is the default association table, and again while attachments are associated — different messages |
| deleting an attachment | its propagations go; a static route naming it survives as a blackhole |
| another gateway's attachment | `InvalidTransitGatewayAttachmentID.NotFound`, as though it did not exist |
| id errors | `InvalidRouteTableID.NotFound` but `InvalidRouteTableId.Malformed` — the casing differs |

The upsert and the casing are the two I would have got wrong from the reference alone.

## Concurrency

Every write to a gateway, route table, attachment, propagation or route happens under the
region-scoped topology lock introduced in #2347, acquired before any striped lock. That now
includes `createTransitGateway` and `modifyTransitGateway`, which write route tables through their
default-table markers and previously held only the gateway's own lock — a modify racing a route
table delete could put the deleted table back. The rule is deliberately exceptionless so it can be
checked by reading rather than reasoned about case by case.

## Tests

- `Ec2ServiceTest`: option and default-flag behaviour, the association lifecycle, propagation
  including the duplicate, derived propagated routes appearing and disappearing with the
  propagation, static versus blackhole, replace in both directions and its upsert, the two delete
  refusals, cross-gateway rejection across all six call sites, the delete cascades, and a race
  asserting a modify cannot resurrect a deleted route table.
- `Ec2TransitGatewayRouteTableIntegrationTest`: the Query wire shapes, including the listings that
  drop the route table id the mutating calls carry.
- `ec2.bats`: the association move, propagation and its route, static and blackhole routes, and
  replace — parsed by the AWS CLI rather than by hand-written assertions.

`ReplaceTransitGatewayRoute` is included though the issue does not list it, since it completes the
route surface.

Two actions in the family are deliberately left out. Route table announcements belong to gateway
peering rather than to route tables. `ExportTransitGatewayRoutes` is an operational call that no
`aws_ec2_transit_gateway_*` resource drives, and its output format — the key layout in the bucket
and the schema of the exported document — is not in the API reference, so it wants checking against
a real account rather than guessing. Filed as #2349 so it is tracked rather than quietly absent, which is why closing #2308 here does
not leave anything unaccounted for.

### DIFF
--- a/compatibility-tests/sdk-test-awscli/test/ec2.bats
+++ b/compatibility-tests/sdk-test-awscli/test/ec2.bats
@@ -11,11 +11,16 @@ setup() {
     TRANSIT_GATEWAY_ID=""
     TGW_ATTACHMENT_ID=""
     TGW_VPC_ID=""
+    TGW_ROUTE_TABLE_ID=""
 }
 
 teardown() {
     if [ -n "$PREFIX_LIST_ID" ]; then
         aws_cmd ec2 delete-managed-prefix-list --prefix-list-id "$PREFIX_LIST_ID" >/dev/null 2>&1 || true
+    fi
+    if [ -n "$TGW_ROUTE_TABLE_ID" ]; then
+        aws_cmd ec2 delete-transit-gateway-route-table \
+            --transit-gateway-route-table-id "$TGW_ROUTE_TABLE_ID" >/dev/null 2>&1 || true
     fi
     if [ -n "$TGW_ATTACHMENT_ID" ]; then
         aws_cmd ec2 delete-transit-gateway-vpc-attachment \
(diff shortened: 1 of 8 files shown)


ck-michals:
Confirmed and fixed in c4dabd17. The default arm matched everything, so those three filters
returned the whole table — a plausible wrong answer rather than an error.

Checking them on a live account first turned out to matter, because the three do not take the same
value form and the reference does not say so. Against a table holding `10.0.0.0/8`, `10.1.0.0/16`,
`10.1.1.0/24` and `192.168.0.0/16`:

```
supernet-of-match 10.1.1.0/24   -> /8, /16 and /24, the exact match included
subnet-of-match   10.1.0.0/16   -> /16 and /24, the exact match included
longest-prefix-match 10.1.1.5   -> /24
longest-prefix-match 10.1.9.9   -> /16
longest-prefix-match 8.8.8.8    -> nothing
longest-prefix-match 10.1.1.5/32 -> nothing
supernet-of-match 10.1.1.5      -> nothing
```

So the two containment filters take a CIDR and match inclusively in either direction, while
`longest-prefix-match` takes a bare address and returns the single most specific route covering it.
Handing either the other's form returns nothing on AWS, and now here too. Had I implemented the
finding as written, all three would have taken CIDRs and `longest-prefix-match` would have been
quietly wrong.

While there, an unrecognised filter name is now rejected rather than ignored, matching
`InvalidParameterValue: Value (nonsense) for parameter Filter is invalid.` — accepting one answers
a question that was never asked, which is the same failure in a different place.

The test covers all seven cases above plus the rejection.

Issue #2350, opened 2026-08-16
[BUG] sshd with EC2 not starting up for current nightly build


cfranzen:
## Service

EC2

## AWS API Action

runInstances with image ami-alpine

## Expected behavior

sshd server starts up as part of the EC2 instance

## Actual behavior

sshd server does not startup. Error logs

2026-08-16 13:22:18,632 WARN  [io.git.hec.flo.ser.ec2.Ec2ContainerManager] Could not start sshd for EC2 instance i-bfea432c2b3b750db: sshd requires execution with an absolute path
INFO  [Ec2ContainerManager] EC2 instance i-bfea432c2b3b750db running in container 45791bd13f5a43f41006388f18b69d2ecf512128d2999d00d4cddc2551ece2ab (SSH host port 2200)
INFO  [Ec2ContainerManager] Injected SSH public key into container 45791bd13f5a43f41006388f18b69d2ecf512128d2999d00d4cddc2551ece2ab

## Environment

- Floci version / image tag: nightly (16.08.2026)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): testcontainers-java

cfranzen:
I believe the issue is the following line that changed recently:
```
ContainerExecResult start = execInContainerForResult(containerId, new String[]{"sshd"}, 5);
```
it should look like
```
ContainerExecResult start = execInContainerForResult(containerId, new String[]{"/usr/sbin/sshd"}, 5);
```

hectorvent:
fixed by #2351

PR #2351, opened 2026-08-16
fix(ec2): sshd can only be started from an absolute path since OpenSSH v3.9

cfranzen:
## Summary

Fixes issue #2350 

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

No changes to compatibility

## Checklist


pgermosen (approved):
Confirmed the core claim: "sshd re-exec requires execution with an absolute path" is a genuine, well-documented OpenSSH self-re-exec/privilege-separation constraint, and the error message matches the linked issue's repro exactly. This is a real regression against #2245's earlier choice to resolve `sshd` via `PATH`.

The one greptile finding (test still expecting the old bare `sshd` command) is genuinely fixed — both `Ec2ContainerManagerTest` assertions now expect `/usr/sbin/sshd`.

20/20 targeted tests pass, `docs-check` clean, no conflicts, matches current `main`.

Minor, non-blocking note: trying a hardcoded 3-path candidate list works for the reported case and the common distro families (all first-candidate `/usr/sbin/sshd`), but capturing the real path from the install probe's own `command -v sshd` check would be more robust for exotic images — worth considering as a follow-up, not blocking here.

LGTM, approving.

PR #2352, opened 2026-08-16
feat(rds): add option group CRUD

### LINKED ISSUE #2304: [FEAT] rds: option group CRUD
## Service

RDS

## API Action / Feature

`CreateOptionGroup`, `ModifyOptionGroup`, `DescribeOptionGroups`, `DeleteOptionGroup`

## AWS Documentation

- https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateOptionGroup.html
- https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyOptionGroup.html
- https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeOptionGroups.html
- https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DeleteOptionGroup.html

## Why is this needed?

None of the option-group actions exist anywhere in `services/rds`, so `aws_db_option_group`
fails on apply, and a `aws_db_instance` that references `option_group_name` cannot be created
either.

Option groups are how engine-specific features are enabled on the engines that use them —
Oracle (`OEM`, `S3_INTEGRATION`, `Timezone`), SQL Server (`SQLSERVER_BACKUP_RESTORE`), and
MySQL/MariaDB (`MARIADB_AUDIT_PLUGIN`, `MEMCACHED`). AWS also creates a `default:<engine>-<major
version>` option group implicitly for every engine, and `DescribeOptionGroups` is expected to
return it even when the caller has created none.

## Scope

Four actions with a clear in-repo precedent: `services/rds` already implements the equivalent
CRUD for the two sibling group types —
`CreateDBParameterGroup`/`ModifyDBParameterGroup`/`DeleteDBParameterGroup`/
`DescribeDBParameterGroups` and the `DBSubnetGroup` set. Option groups should mirror that shape
exactly, including using `StorageFactory` for the backing store.

Wire shapes:

- `CreateOptionGroup` — `OptionGroupName`, `EngineName`, `MajorEngineVersion`,
  `OptionGroupDescription` → returns an `OptionGroup` with `OptionGroupArn` and an empty
  `Options` list
- `ModifyOptionGroup` — `OptionsToInclude.OptionConfiguration.N.OptionName` (plus
  `Port`, `OptionSettings`, `VpcSecurityGroupMemberships`) and `OptionsToRemove.N`. This is the
  action the provider calls for every `option {}` block
- `DescribeOptionGroups` — filter by `OptionGroupName`, `EngineName`, `MajorEngineVersion`; an
  unknown name returns `OptionGroupNotFoundFault`, not an empty list
- `DeleteOptionGroup` — empty response; a group still attached to an instance returns
  `InvalidOptionGroupStateFault`

## Docs

Adds new actions, so `make docs-sync` is required.

- [ ] Yes
- [ ] No

omatheusmesmo:
## Summary

Implements the four RDS option group actions. Closes #2304.

`CreateOptionGroup`, `DescribeOptionGroups`, `ModifyOptionGroup` and `DeleteOptionGroup` follow the
shape the two sibling group types already use in `services/rds`: a `StorageFactory` backing store,
per-account and per-region keys via `dbResourceKey`, and a managed-defaults list mirroring
`MANAGED_CLUSTER_PARAMETER_GROUPS`. Option groups are also taggable through the `og:` ARN form of
`AddTagsToResource`, which `resolveTagHandle` previously listed as not yet implemented.

`DescribeOptionGroups` returns the implicit `default:<engine>-<major version>` groups even when the
caller has created none, as AWS does. That list is scoped to the engines Floci can actually run
(`postgres 13` to `18`, `mysql 8.0`/`8.4`, `mariadb 10.11`/`11.2`/`11.4`), so every instance Floci
can create resolves to a default the action also returns; inventing the Oracle and SQL Server
version matrix would be fabricating data. `CreateOptionGroup` still validates `EngineName` against
the full list AWS accepts, so an `aws_db_option_group` for `oracle-ee` or `sqlserver-ex` applies
normally; option groups are metadata, they don't need the engine running.

DB instances now carry an option group too: `CreateDBInstance` and `ModifyDBInstance` accept
`OptionGroupName` (validated against the instance engine and major engine version) and
`DescribeDBInstances` reports
`OptionGroupMemberships`. This is one step past the four actions, but the issue requires
`DeleteOptionGroup` to fault with `InvalidOptionGroupStateFault` for a group still attached to an
instance, and without modelling the association that error would be unreachable code. It is also
what the issue's "Why is this needed?" describes.

Three things I deliberately left out, happy to open them as issues if you agree they're worth having:

1. **`CopyOptionGroup`**, small now that the model is complete. It is also where `SourceOptionGroup`,
   `SourceAccountId` and `CopyTimestamp` would finally get populated; they are omitted today because
   nothing can set them.
2. **`DescribeOptionGroupOptions`**, the per-engine option catalog. Bigger than it looks: it is what
   would let `DescribeOptionGroups` return real `OptionSetting` metadata, so the two are effectively
   one piece of work.
3. **Pagination for the RDS list actions.** All eight emit an empty `<Marker>` today. Worth doing
   across all of them at once rather than only here, so the service stays internally consistent.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified with **AWS SDK for Java v2 `2.44.14`** against a running Floci, via the new
`RdsControlPlaneTest` cases. Not verified with the AWS CLI.

Wire shapes came from the RDS API Reference rather than from memory, which caught several things:

- the `DescribeOptionGroups` list wrapper is `OptionGroupsList`, not `OptionGroups`
- `DeleteOptionGroup` returns **no** `Result` element, only `ResponseMetadata`
- an unknown `OptionGroupName` faults with `OptionGroupNotFoundFault` (404) rather than returning an
  empty list, which is what a Terraform refresh relies on to detect a deleted group
- the list member name differs between request and response for the same data:
  `DBSecurityGroupMemberships.DBSecurityGroupName.N` inbound vs `.DBSecurityGroup.N` outbound
- `OptionsToInclude` uses the custom `.OptionConfiguration.N` member name while `OptionsToRemove`
  uses plain `.member.N`; both are parsed, plus the legacy `.member.N` form older SDKs emit for
  `OptionsToInclude`
- the `OptionGroupName` constraints mean a name can never collide with a default, since `:` is not
  an allowed character, so shadowing `default:mysql-8-0` fails with `InvalidParameterValue` rather
  than `OptionGroupAlreadyExistsFault`
- an option group is scoped to one major engine version, so attaching a `mysql 8.0` group to a
  `mysql 8.4` instance fails with `InvalidParameterCombination` even though the engine matches.
  The major version is the leading component for postgres and `major.minor` for the MySQL-family
  engines, which is the same rule that names the implicit `default:` groups

Known deviations, documented in `docs/services/rds.md` so they read as choices rather than misses:
the 20-group `OptionGroupQuotaExceededFault` is not enforced (capping a local emulator would only
get in a test's way); `OptionSetting` metadata (`DataType`, `AllowedValues`, `DefaultValue`, and so
on) is omitted, since it needs the catalog `DescribeOptionGroupOptions` serves and a partial one
would be worse than a consistent omission; and results are returned in a single page, as with every
other RDS list action.

`make docs-sync` was run. `OptionsToRemove` was added to `exclude_actions` in
`tools/docs/services.yaml` because the generator reads every `case "X" ->` in the handler as an
action, and that label belongs to `memberKeyRegex`, not the action dispatcher. `SubnetIds` and
`VpcSecurityGroupIds` were already excluded for the same reason.

## Checklist

- [x] `./mvnw test` passes locally, with the caveat below

47 new unit tests in `RdsServiceTest` and `RdsQueryHandlerTest`, written before the implementation,
plus 4 SDK round-trip tests in `RdsControlPlaneTest` that assert the official SDK unmarshals the XML
and that `OptionGroupNotFoundException` and `InvalidOptionGroupStateException` surface as typed
exceptions on the client. 355 unit tests and 11 compatibility tests green.

On the first box: the full `./mvnw test` reports errors in the API Gateway v2 WebSocket suites and
in `LambdaReactiveSyncIntegrationTest`, `SwfLambdaIntegrationTest`,
`LambdaExecutionRoleDockerIntegrationTest` and `ElbV2LambdaTargetDataPlaneIntegrationTest`
(`WebSocketHandshakeException`, `SocketTimeout` on Lambda prewarm). They are pre-existing on my
machine rather than caused by this change: `WebSocketProxyEventFormatTest` fails with the identical
12 errors in a clean `git worktree` at `main` (`d819c00d`). No RDS suite fails.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/RdsControlPlaneTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/RdsControlPlaneTest.java
@@ -13,15 +13,24 @@ import software.amazon.awssdk.services.rds.RdsClient;
 import software.amazon.awssdk.services.rds.model.ConnectionPoolConfigurationInfo;
 import software.amazon.awssdk.services.rds.model.CreateDbProxyResponse;
 import software.amazon.awssdk.services.rds.model.CreateDbSubnetGroupResponse;
+import software.amazon.awssdk.services.rds.model.CreateOptionGroupResponse;
 import software.amazon.awssdk.services.rds.model.DBProxyTarget;
 import software.amazon.awssdk.services.rds.model.DescribeDbSubnetGroupsResponse;
+import software.amazon.awssdk.services.rds.model.DescribeOptionGroupsResponse;
 import software.amazon.awssdk.services.rds.model.DescribeOrderableDbInstanceOptionsResponse;
+import software.amazon.awssdk.services.rds.model.InvalidOptionGroupStateException;
+import software.amazon.awssdk.services.rds.model.ModifyOptionGroupResponse;
+import software.amazon.awssdk.services.rds.model.OptionConfiguration;
+import software.amazon.awssdk.services.rds.model.OptionGroupNotFoundException;
+import software.amazon.awssdk.services.rds.model.OptionSetting;
 
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.tuple;
 
 @DisplayName("RDS Control Plane")
 class RdsControlPlaneTest {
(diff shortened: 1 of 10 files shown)


pgermosen (approved):
Both Greptile findings don't survive scrutiny. The P1 (major version compatibility ignored) is verified false — disabled validateInstanceOptionGroup's version check and confirmed createDbInstanceRejectsOptionGroupForAnotherMajorEngineVersion genuinely fails without it, then restored it and confirmed it passes. The P2 (option storage bypasses factory) isn't a real deviation — every other backing store in that same test constructor already uses new InMemoryStorage<>() directly; the option-group field just follows the class's existing convention.

Independently spot-checked two of the more specific wire-shape claims against the real botocore RDS model — both exact matches: DescribeOptionGroups's output structure wraps its list under OptionGroupsList (not OptionGroups, despite that being the structure's own shape name), and DeleteOptionGroup genuinely has no output shape at all.

Ran the full RDS suite: RdsServiceTest 208/208, RdsQueryHandlerTest 90/90, 0 failures. Clean merge, docs-check passes, no CODEOWNERS-protected paths.

PR #2353, opened 2026-08-16
fix(apigateway): verify and propagate HTTP API v2 JWT authorizer claims

### LINKED ISSUE #2010: [BUG] HTTP API v2 JWT authorizer: validated claims not surfaced at requestContext.authorizer.jwt.claims in AWS_PROXY Lambda events
## Service

API Gateway v2 (HTTP API), Lambda

## AWS API Action

ExecuteApi — HTTP API route with `AuthorizationType: JWT` integrated to Lambda (`AWS_PROXY`)

## Expected behavior

When a JWT authorizer validates a request on an HTTP API v2 route, real API Gateway surfaces the verified claims to the downstream Lambda in the payload 2.0 event:

```json
"requestContext": {
  "authorizer": {
    "jwt": {
      "claims": { "sub": "...", "email": "...", "iss": "...", ... },
      "scopes": null
    }
  }
}
```

Handlers that resolve the caller's identity from `event.requestContext.authorizer.jwt.claims` (the documented pattern for Cognito/OIDC-backed HTTP APIs) should receive the claims.

## Actual behavior

Floci's JWT enforcement itself works — requests with no/invalid tokens are correctly rejected with 401 — but the event delivered to the Lambda contains **no `authorizer` node at all**. Handlers that require `claims.sub` fail on every request that passed gateway validation (in our case a 401 `authentication required` from the handler's own guard, despite a valid Cognito-issued JWT).

Root cause (source inspection): `ApiGatewayExecuteController.buildV2ProxyEvent` builds `requestContext` (accountId, apiId, http, …) but never adds an `authorizer` node. Notably the claims parsing already exists in the same controller — the v2 **HTTP_PROXY** path calls `parseAllJwtClaims` to feed `$context.authorizer.claims.*` parameter mapping — it is only the AWS_PROXY Lambda event that misses the injection. The v1/REST TOKEN authorizer equivalent was fixed in #581; this is the v2 JWT counterpart.

Related: #1758 covers CloudFormation dropping `AWS::ApiGatewayV2::Authorizer` (deploy-time). This issue is the request-time half — even with the authorizer created via the CLI (`create-authorizer` + `update-route`), the claims never reach the Lambda.

## Reproduction

1. Create an HTTP API v2 with a JWT authorizer (Cognito user pool issuer) and a `GET` route with `AuthorizationType: JWT` integrated to a Lambda (AWS_PROXY).
2. Call the route with a valid `Authorization: Bearer <IdToken>`.
3. Gateway returns the Lambda's response (JWT validation passed), but inside the Lambda `event.requestContext.authorizer` is `undefined`.

Verified on 1.5.33 and current `main`. On real AWS the same stack receives the claims.

rogueserenity:
## Summary

`enforceJwtAuthorizer` (the HTTP API v2 route path for `AuthorizationType: JWT`) checked a
token's `iss`/`aud`/`exp` claims but never verified the token was actually signed by its issuer —
`parseJwtClaims` only base64-decoded the JWT payload segment, so any well-formed token whose
claimed `iss`/`aud` matched the authorizer's configuration was accepted regardless of whether it
was ever signed by that issuer. On top of that, the (unverified) claims were discarded on success
instead of being exposed to the downstream Lambda, so `buildV2ProxyEvent` never populated
`requestContext.authorizer.jwt.claims` in the event at all — unlike the sibling CUSTOM/Lambda
REQUEST authorizer path. Any Lambda relying on the native JWT authorizer's identity (rather than
re-verifying the token itself) saw no identity at all, even though the token was accepted and the
request let through.

This PR fixes both:

- **Signature verification.** Adds `JwtSignatureVerifier`: real OIDC discovery
  (`GET {issuer}/.well-known/openid-configuration` → `jwks_uri`) + JWKS fetch, matches the
  token's `kid`, and verifies its RS256 signature against the resolved key — the same mechanism
  real HTTP API JWT authorizers use against a real issuer, not a Floci-specific shortcut (an
  authorizer's issuer is an arbitrary external IdP — Cognito, Auth0, Okta, ...). Per-issuer JWKS
  are cached for 10 minutes. Wired in ahead of the existing `iss`/`aud`/`exp` checks, since nothing
  in the payload should be trusted until the signature itself is proven valid. Any verification
  failure (forged signature, no matching `kid`, unsupported alg including `none`, unreachable
  issuer, missing issuer config) is a 401 — there is no fallback to unverified acceptance.
- **Claims + scopes propagation.** Once a token verifies, its claims are threaded through to
  `requestContext.authorizer.jwt.claims`, and the `scope` claim is split into its own
  `requestContext.authorizer.jwt.scopes` array, matching AWS's real payload shape (a raw
  space-delimited `scope` string in `.claims` is not what AWS delivers).
- **HTTP_PROXY claim mappings now use the verified token.** `dispatchHttpProxyV2` (the code path
  for `$context.authorizer.claims.*` request-parameter mappings on `HTTP_PROXY` integrations) had
  its own, separate, unverified re-parse of a token pulled straight from the `Authorization`
  header — regardless of the authorizer's actual configured `identitySource`. The authorizer's
  gate itself wasn't bypassed (a bad token still 401s before this code runs), but a caller
  presenting a valid token via a non-header identity source (e.g. a querystring parameter) plus
  an unrelated, attacker-controlled `Authorization` header would have claim mappings resolve from
  that second, unverified token instead of the one the authorizer actually checked. Now reuses
  the claims already verified upstream instead of re-deriving them.

Closes #2010

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior, two parts:

1. HTTP API (v2) JWT authorizers accepted any token whose *claimed* `iss`/`aud` matched
   configuration, without ever checking a signature — real API Gateway verifies against the
   issuer's published JWKS before trusting anything in the token.
2. On a request that did pass verification, no `requestContext.authorizer` node reached the
   downstream Lambda at all — real API Gateway delivers `requestContext.authorizer.jwt.claims`
   (and `.scopes`, split from the token's `scope` claim per RFC 6749 §3.3) on every request that
   passes JWT verification.

Verified against a real downstream consumer migrating off in-process token verification in favor
of the native authorizer: every write route gated on the caller's authorizer-verified `sub`
failed after a real SAM deploy to floci, despite passing identically against real AWS API
Gateway. Confirmed fixed against that consumer's full functional suite.

Signature verification, and the HTTP_PROXY claim-mapping gap, were both raised during review as
follow-on findings — the claims-propagation fix made the missing signature check more
consequential (unverified claims would otherwise reach the Lambda as trusted identity), and
fixing that surfaced a second, independent path making the same unverified-claims mistake. Both
are covered by dedicated tests (`JwtSignatureVerifierTest`; forged-signature/expired/
wrong-audience cases in `HttpApiJwtAuthorizerQuerystringTest`;
`HttpProxyJwtClaimsMappingIntegrationTest` for the HTTP_PROXY path). Each fix was verified by
reverting it locally and confirming the corresponding test fails.

## Checklist

- [x] `./mvnw test` passes locally (10,456 tests, full suite)
- [x] New or updated integration test added (`JwtSignatureVerifierTest`,
      `HttpApiJwtAuthorizerQuerystringTest`, `HttpProxyJwtClaimsMappingIntegrationTest`,
      `BuildV2ProxyEventJwtClaimsTest`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -1379,7 +1379,8 @@ public class ApiGatewayExecuteController {
         if (integrationType == null || integrationType.isEmpty()) integrationType = "AWS_PROXY";
 
         if ("HTTP_PROXY".equalsIgnoreCase(integrationType)) {
-            return dispatchHttpProxyV2(integration, route, httpMethod, path, headers, uriInfo, body, apiId, stageName);
+            return dispatchHttpProxyV2(integration, route, httpMethod, path, headers, uriInfo, body,
+                    apiId, stageName, jwtClaims);
         }
 
         String functionName = functionNameFromUri(integration.getIntegrationUri());
@@ -1415,7 +1416,7 @@ public class ApiGatewayExecuteController {
     private Response dispatchHttpProxyV2(io.github.hectorvent.floci.services.apigatewayv2.model.Integration integration,
                                           Route route, String httpMethod, String path,
                                           HttpHeaders headers, UriInfo uriInfo, byte[] body,
-                                          String apiId, String stageName) {
+                                          String apiId, String stageName, Map<String, String> jwtClaims) {
         // CDK HttpAlbIntegration sets integrationUri to an ALB listener ARN. Resolve it
         // to the listener's bound localhost port so HttpProxyInvoker (which assumes a
         // concrete http(s) URL) can forward through the listener's data plane.
@@ -1448,14 +1449,14 @@ public class ApiGatewayExecuteController {
         }
         Map<String, String> pathParams = extractV2PathParams(route.getRouteKey(), path);
 
-        Map<String, Object> claims = Map.of();
-        if ("JWT".equalsIgnoreCase(route.getAuthorizationType())) {
-            String token = extractBearerToken(headers);
-            if (token != null) {
-                Map<String, Object> parsed = parseAllJwtClaims(token);
-                if (parsed != null) claims = parsed;
-            }
-        }
+        // Reuses the claims dispatchV2 already verified via enforceJwtAuthorizer, rather than
+        // independently re-extracting a token and re-parsing it here: extractBearerToken only
+        // reads the Authorization header, ignoring the authorizer's configured identitySource
+        // (which - see HttpApiJwtAuthorizerQuerystringTest - can be a querystring parameter). A
+        // caller passing a valid token via the configured source plus an unrelated Bearer header
+        // would otherwise have $context.authorizer.claims.* resolve from the unverified header
+        // token instead of the one the authorizer actually checked.
+        Map<String, Object> claims = jwtClaims != null ? Map.copyOf(jwtClaims) : Map.of();
 
         String sourceIp = requestHeaders.getOrDefault("X-Forwarded-For", "127.0.0.1");
         io.github.hectorvent.floci.services.apigatewayv2.proxy.RequestContext ctx =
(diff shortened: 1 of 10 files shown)


rogueserenity:
@hectorvent @pgermosen this PR is ready for your review.

pgermosen:
Nice work here — verified the fix is real by reverting the signature check and the HTTP_PROXY claims-reuse change independently: with signature verification disabled, `requestWithForgedSignatureInQueryStringIsUnauthorized` genuinely fails (a forged token gets through), and reverting the claims-reuse fix gives an unambiguous `expected: <verified-user> but was: <attacker-controlled-user>` — a direct demonstration of the vulnerability this PR closes.

One smal thing: the comment above the `.claims`/`.scopes` node-building in `buildV2ProxyEvent` (around `ApiGatewayExecuteController.java:2043`) says "callers never see a raw \`scope\` string," but the code right below it copies every claim — including `scope` — into `.claims` via `jwtClaims.forEach(claimsNode::put)`, before separately splitting it into `.scopes`. That's actually correct AWS-matching behavior (real API Gateway keeps both), the comment is just describing it backwards. Worth a quick wording fix so a future reader doesn't take it as license to strip `scope` from `.claims`.

rogueserenity:
@pgermosen Fixed in 6d651be9 — updated the comment to correctly describe that `.claims` retains the raw `scope` claim while `.scopes` holds the whitespace-split array, matching real AWS behavior. Thanks for catching it.

pgermosen (approved):
Thanks for the quick fix on the comment wording, @rogueserenity — that resolves the last open item. Re-ran the full targeted suite (JwtSignatureVerifierTest, HttpApiJwtAuthorizerQuerystringTest, HttpProxyJwtClaimsMappingIntegrationTest, BuildV2ProxyEventJwtClaimsTest, ApiGatewayExecuteControllerTest): 40/40 pass, and the merge against current main is still clean. This is solid, well-verified security work — appreciate the care taken here.

Issue #90, opened 2026-03-27
[BUG] CloudWatch Logs: `GetLogEvents` does not return request token at end of stream


davidjb:
## Service

CloudWatch Logs

## AWS API Action

`GetLogEvents`

## Expected behavior

https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_GetLogEvents.html

> `GetLogEvents` is a paginated operation
> ...
> Partially full or empty pages don't necessarily mean that pagination is finished. As long as the `nextBackwardToken` or `nextForwardToken` returned is NOT equal to the `nextToken` that you passed into the API call, there might be more log events available. The token that you use depends on the direction you want to move in along the log stream. The returned tokens are never null.

https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_GetLogEvents.html#API_GetLogEvents_ResponseElements

`nextBackwardToken` and `nextForwardToken` in the response are set to the `nextToken` from the request when the end of the stream is reached.

## Actual behavior

https://github.com/hectorvent/floci/blob/7b87bf2c1fa8f9cff7aef4be488d7b2cbf3fe26d/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java#L295-L297

Floci always returns a generated new set of tokens, regardless of whether the `page` is the end of the stream or not.

For reference, this causes infinite loops on pagination in SDK clients like https://github.com/aws/aws-sdk-go-v2/blob/71108dc1cb6ccf7a4f36a85a81e0fe954a67fc73/service/cloudwatchlogs/api_op_GetLogEvents.go#L317-L322 - because Floci's API response differs and the same token isn't in the response, the client keeps looping and making API requests to `GetLogEvents`.

## Reproduction

```bash
# Minimal AWS CLI or SDK snippet that triggers the issue
$ aws --endpoint-url http://localhost:4566 logs get-log-events --log-stream-name foobar --next-token 42
b/965df692-23e9-4ce2-bba2-9653a66310e2  f/138e80d1-6fa2-451a-a234-593160fd89e5

$ aws --endpoint-url http://localhost:4566 logs get-log-events --log-stream-name foobar --next-token 42
b/f4b419aa-e864-47fc-b406-c7f7dc88d041  f/d60faf3f-d3e9-4a09-b5c9-c199b121e0b7

...
```

## Environment

- Floci version / image tag: `1.0.11`
- Java SDK version (if applicable):
[...]

PR #2355, opened 2026-08-17
fix(cloudwatch): keep GetLogEvents paging inside the list bounds

hampsterx:
`GetLogEvents` ends a page at `Math.min(pageStart + maxEvents, total)`, where `maxEvents` is `Math.min(limit > 0 ? limit : Integer.MAX_VALUE, maxEventsPerQuery)`. `FLOCI_SERVICES_CLOUDWATCHLOGS_MAX_EVENTS_PER_QUERY` has no upper bound, so setting it near `Integer.MAX_VALUE` (the natural way to express "no cap") makes `pageStart + maxEvents` overflow negative the moment a forward token moves pagination past the first page. `subList` then throws and the call fails with a 500:

```
java.lang.IllegalArgumentException: fromIndex(3) > toIndex(-2147483646)
	at CloudWatchLogsService.getLogEvents(CloudWatchLogsService.java:446)
```

The end of the stream is where a client meets this, not some deep page. `GetLogEvents` echoes its forward token when nothing is left (#90, #184), so a paginator always spends one more call on the token it was just handed. With a large cap and at least one event, that call is the one that throws.

Taking the window out of what remains keeps both operands inside the list's own bounds:

```java
pageEnd = pageStart + Math.min(maxEvents, total - pageStart);
```

`pageStart` is already clamped to `total`, so `total - pageStart` is non-negative and `pageEnd` lands in `[pageStart, total]`. Behaviour at the default cap of 10000 is unchanged.

The other three branches are unaffected: the backward and tail branches subtract and clamp with `Math.max(..., 0)`, and the initial forward branch is a plain `Math.min`.

Out of scope: a *negative* `max-events-per-query` is still unvalidated and produces an invalid range in several branches. That is a config-validation gap rather than this overflow, and the fix here neither helps nor worsens it.

## Test plan
- [x] `getLogEventsPagesForwardWithAnUnboundedMaxEventsPerQuery` drives the second paginator call under a cap of `Integer.MAX_VALUE`; it fails on `main` with the exception above and passes here
- [x] `./mvnw clean test`: 10476 tests, 0 failures, 0 errors, 9 skipped

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java
@@ -428,7 +428,10 @@ public class CloudWatchLogsService {
         if (nextToken != null && nextToken.startsWith("f/")) {
             int offset = parseTokenIndex(nextToken, 2);
             pageStart = Math.min(offset, total);
-            pageEnd = Math.min(pageStart + maxEvents, total);
+            // Take the window out of what is left rather than adding to the offset, so a
+            // max-events cap configured near Integer.MAX_VALUE cannot overflow the end
+            // index negative once pagination has moved past the first page.
+            pageEnd = pageStart + Math.min(maxEvents, total - pageStart);
         } else if (nextToken != null && nextToken.startsWith("b/")) {
             int end = parseTokenIndex(nextToken, 2);
             pageEnd = Math.min(end, total);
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Reproduced the overflow exactly as described: reverting just the fixed line reproduces `IllegalArgumentException: fromIndex(3) > toIndex(-2147483646)` verbatim, triggered once a forward-paginating call moves `pageStart` past 0 with `maxEventsPerQuery` configured near `Integer.MAX_VALUE`.

The other three pagination branches (backward, tail, initial-forward) are correctly unaffected — checked the bounds on each, none can produce the same overflow.

`FLOCI_SERVICES_CLOUDWATCHLOGS_MAX_EVENTS_PER_QUERY` has no documented upper bound, so the scenario is reachable by config alone, not just a contrived test.

PR #2357, opened 2026-08-17
refactor(ses): split SesService into per-domain services behind a facade

### LINKED ISSUE #2356: Refactor SesService into per-domain services behind a facade
## Problem

`SesService` has grown to ~3,869 lines and owns 14 in-memory stores. Both the v2 REST
controller (`SesController`) and the v1 Query handler (`SesQueryHandler`) depend on the
same `SesService`, sharing logic across protocols. Six unit-test files construct it by
injecting all 14 stores positionally, so adding a single store ripples through the
constructor and every one of those tests (this surfaced while implementing VDM). The
class keeps growing as SES coverage expands and is becoming hard to navigate and to test
in isolation.

## Proposed approach: keep the facade, split the internals by store

Keep `SesService` as the public entry point (a facade) so consumers — controllers,
handler, tests — stay unchanged, and move store-based domains into separate
`@ApplicationScoped` services one at a time (strangler pattern):

- Single-domain operations become thin delegations
  (e.g. `getAccountVdmAttributes()` -> `accountService.findVdm()`).
- Each domain service owns its store(s) via `StorageFactory` and gets a small
  package-private constructor, so its unit test injects only that domain — no
  14-argument constructor.
- Cross-cutting and send-path logic stays in the facade, which coordinates the domains
  (facade-as-coordinator): the send path reads what it needs through read-only finders
  instead of domains depending on each other, keeping each domain a dependency-free leaf.

```
SesController (v2) / SesQueryHandler (v1) / tests    <- unchanged
                       |
                 SesService (facade)                 <- single-domain: delegate / cross-cutting: coordinate
   +-----------+-----------+-----------+-------------+----------- ...
 Account     Template   Suppression  DedicatedIp   Contact   ...   (each @ApplicationScoped, owns its store)
```

## Scope of the accompanying PR

An accompanying PR implements this change in full — complete and tested, and intended to
be merged. I'm opening this issue to confirm the direction and your preferences before /
while it is reviewed, and I'm glad to adapt it (for example, split it into smaller
per-domain PRs) to match how you'd prefer it landed.

It extracts 8 domain services covering 12 of the 14 stores; the facade shrinks from
~3,869 to ~2,869 lines. Consumers are untouched.

| Domain service | Store(s) |
| --- | --- |
| `SesReceiptRuleService` | receipt rule sets |
| `SesAccountService` | account settings, account VDM |
| `SesCvetService` | custom verification email templates |
| `SesPolicyService` | identity sending-authorization policies |
| `SesContactService` | contact lists, contacts |
| `SesSuppressionService` | account suppression, suppression list |
| `SesDedicatedIpService` | dedicated IP pools |
| `SesTemplateService` | email templates |
| `SesSentEmailService` | sent-email records |

Two stores intentionally stay on the facade for now, because they are cross-cutting hubs:

- **`configSetStore`** is read by the send path and cross-validates against the
  dedicated-IP domain (delivery options); splitting it cleanly needs the send/reputation
  layering settled first.
- **`identityStore`** is the largest (DKIM state machine, verification, MAIL FROM,
  notification resolution) and is what tenant multi-tenancy will associate resources
  against.

Deferring both avoids drawing those boundaries twice.

## Non-goals / guarantees

- No wire-protocol or behavior change — each step is a pure delegation.
- Verified against a live build: the SES SDK compatibility suite (195 tests across 20
  classes) passes, plus the SES unit and integration tests.

## Design notes for discussion

- **Send path.** `SendEmail` touches identity + configuration set + suppression +
  template + sent-email records at once. This PR keeps that orchestration in the facade
  (facade-as-coordinator) rather than a dedicated send service that injects the other
  domains, to keep domains as non-interdependent leaves. Open to the alternative.
- **Tenant-aware ordering.** SES v2 Tenants (multi-tenancy) are inherently cross-cutting
  — new tenant/association stores, plus tenant-scoped sending / reputation / suppression.
  This PR extracts the low-coupling leaves first and leaves identity / configuration-set
  and the send path to be finalized alongside the tenant design, so those boundaries are
  drawn once.
- This introduces a "one service -> several beans" shape for SES. The per-service
  CloudFormation provisioners are a precedent for a CDI split within a service.

## Questions

I'd like to land this change; a few decisions I want to confirm first:

1. Is keeping the facade and splitting the internals by store a direction you'd accept?
2. Send-path placement: facade-as-coordinator (as implemented), or a dedicated send
   service?
3. Does deferring the identity / configuration-set / send-path boundaries to the tenant
   work make sense?

If smaller reviewable chunks would help, I can split the PR into one per domain.

okinaka:
## Summary

Closes #2356.

Incremental strangler refactor of the SES service. `SesService` had grown to roughly
3,869 lines owning 14 stores, with both the v2 controller and the v1 query handler
depending on it and six unit-test files injecting every store directly, so adding a
single store rippled widely. This keeps `SesService` as the public entry point (a
facade) and moves store-based domains into separate `@ApplicationScoped` services one
at a time, leaving consumers unchanged.

Extracts 8 domain services covering 12 of the 14 stores (the facade shrinks from ~3,869
to ~2,869 lines): `SesReceiptRuleService`, `SesAccountService` (including VDM),
`SesCvetService`, `SesPolicyService`, `SesContactService`, `SesSuppressionService`,
`SesDedicatedIpService`, `SesTemplateService`, `SesSentEmailService`. Cross-cutting and
send-path logic stays in the facade, which coordinates the domains through read-only
finders. Two stores (`identityStore`, `configSetStore`) are intentionally left on the
facade for now because they are cross-cutting hubs better split alongside a larger
change — see the linked issue for the rationale. Each extraction is a separate commit.

No behavior change is intended: each step preserves the existing wire behavior and
delegates from the facade.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No wire-protocol change. Verified with the AWS SDK for Java v2 against a live Floci
build: the SES compatibility suite (195 tests across 20 classes) passes, confirming the
facade split preserves SES wire behavior. SES unit and integration tests also pass.

## Checklist


pgermosen (approved):
Verified the key structural risk in a refactor like this directly: grepped SesService.java for every extracted store field and confirmed zero leftover references — no duplicate/orphaned store ownership. Each new domain service creates its backing store under the exact same file name the facade used before, so persisted data continuity is preserved.

Both Copilot findings are already resolved or don't hold: the facade-back-dependency (SesTemplateService calling SesService.validateTag) is fixed in the current HEAD via a shared SesTags.validate helper; the 'misleading test name' nitpick is a bot false-positive — the test genuinely exercises save() via a reload-and-verify pattern.

Ran all 24 SES test classes locally — 0 failures across the board. Clean merge, docs-check passes, all CI green including the SES SDK compat suite, no CODEOWNERS-protected paths.

PR #2359, opened 2026-08-17
fix: serialize GetRecords ApproximateArrivalTimestamp as a plain decimal

### LINKED ISSUE #2358: [BUG] GetRecords ApproximateArrivalTimestamp serialized in scientific notation, breaks AWS SDK unmarshalling in Java SDK
## Service

Kinesis

## AWS API Action

kinesis.getRecords

## Expected behavior

`GetRecords` should return a valid response the AWS SDK for Java v2 can unmarshal.

## Actual behavior

`GetRecords`'s `ApproximateArrivalTimestamp` field is serialized in scientific notation, which breaks AWS SDK for Java v2's timestamp unmarshalling - despite the response returning HTTP 200. This is the same underlying defect as #2099, just against a different field/endpoint (`GetRecords`'s `ApproximateArrivalTimestamp` rather than `DescribeStream`/`DescribeStreamSummary`'s `StreamCreationTimestamp`) - `KinesisJsonHandler` serializes it the same way, at `rec.getApproximateArrivalTimestamp().toEpochMilli() / 1000.0`, and Java renders that `double` in scientific notation for any 2020s-era timestamp.

A real, practical consequence: any client that creates a fresh `AT_TIMESTAMP` shard iterator for a record that already exists (i.e. not "wait for a new record to arrive after I start polling") never finds it. `GetRecords`'s own `AT_TIMESTAMP` resolution correctly locates the record server-side (confirmed via a delayed `TRIM_HORIZON` query finding it), but any client using AWS SDK for Java v2 fails to parse the corrupted `ApproximateArrivalTimestamp` on records returned via other iterator types too, so this isn't limited to `AT_TIMESTAMP` iteration specifically - it's every `GetRecords` response.

## Reproduction

Floci version/image: `floci/floci:nightly`, source at `floci-io/floci@6779aa8`

How to reproduce:
```java
KinesisClient kinesis = KinesisClient.builder()
    .endpointOverride(URI.create("http://localhost:4566"))
    .region(Region.of("us-east-1"))
    .credentialsProvider(StaticCredentialsProvider.create(
        AwsBasicCredentials.create("test", "test")))
    .build();

kinesis.createStream(b -> b.streamName("my-stream").shardCount(1));
kinesis.putRecord(b -> b.streamName("my-stream")
    .data(SdkBytes.fromUtf8String("test"))
    .partitionKey("pk1"));

String shardId = kinesis.listShards(b -> b.streamName("my-stream"))
    .shards().get(0).shardId();
String iterator = kinesis.getShardIterator(b -> b.streamName("my-stream")
    .shardId(shardId)
    .shardIteratorType(ShardIteratorType.TRIM_HORIZON))
    .shardIterator();

kinesis.getRecords(b -> b.shardIterator(iterator)); // throws SdkClientException
```

Example failure:
```
software.amazon.awssdk.core.exception.SdkClientException: Unable to unmarshall response
  (Unable to parse date : 1.786959659083E9). Response Code: 200, Response Text: OK
Caused by: java.lang.NumberFormatException: For input string: "1.786959659083E9"
  at software.amazon.awssdk.utils.DateUtils.parseUnixTimestampMillisInstant
```

The raw response body contains:
```json
{"Records":[{"Data":"...","PartitionKey":"pk1","SequenceNumber":"...","ApproximateArrivalTimestamp":1.786959659083E9}], ...}
```
instead of a plain decimal like `1786959659.083`.

I have a fix ready (mirrors #2173's `epochSeconds(Instant)` `BigDecimal`-based approach, applied to `KinesisJsonHandler.handleGetRecords`/`handleSubscribeToShard`) and will open a PR against this issue.

## Environment

- Floci version/image: `floci/floci:nightly` (source `floci-io/floci@6779aa8`)
- AWS SDK for Java v2, region eu-west-1
- Reproduced via an internal Kinesis-polling E2E test tool - confirmed Floci-specific: identical scenarios against real AWS Kinesis never hit this

sheldon-nadarajah-sft:
## Summary

Apply the same fix as #2173 to a sibling occurrence of the same defect: `GetRecords` (and `SubscribeToShard`'s event payload) serialize `ApproximateArrivalTimestamp` from epoch milliseconds by dividing through a `double`, so Jackson may serialize the resulting JSON number in scientific notation, e.g. `1.786959659083E9`.

Reuses `KinesisJsonHandler`'s existing `epochSeconds(Instant)` helper (added by #2173, now merged: `BigDecimal.valueOf(millis, 3)`, exact millisecond precision, no exponent notation) at both `ApproximateArrivalTimestamp` call sites (`handleGetRecords`, `handleSubscribeToShard`), instead of adding a second copy of it.

Addressed a review comment: guarded both call sites against a null `ApproximateArrivalTimestamp` (reachable via hybrid/persistent JSON-backed storage - `KinesisRecord` is `@RegisterForReflection` with a no-arg constructor), omitting the field instead of NPEing the whole response. New test: `getRecordsOmitsArrivalTimestampRatherThanThrowingWhenNull`.

## Scope correction (updated after review)

**This PR's original description overstated its impact, and I've rewritten this section to correct that.** It claimed the scientific-notation value breaks AWS SDK for Java v2 unmarshalling, and that this was the root cause of an `AT_TIMESTAMP` client never finding an already-written record.

@pgermosen correctly showed that doesn't hold for a default client: Floci routes Kinesis responses through `AwsJsonCborController` whenever the client sends CBOR, which the real Java SDK does by default, so this JSON *text* never reaches it. Verified independently - the original reproducer raises no exception with or without this change.

So the accurate scope is narrower: this matters only for callers on the plain `application/x-amz-json-1.1` path (boto3/CLI with `AWS_CBOR_DISABLE=true`, or older SDKs), where a scientific-notation number is still a wire-format deviation from real Kinesis worth fixing, exactly as #2173 treats the same defect in `DescribeStream`.

Chasing this down did surface the *actual* cause of the behaviour originally attributed to it, a separate and more serious pre-existing bug in the CBOR timestamp units convention (both request and response directions). That was filed separately as #2368 and has since been fixed and merged via #2432; unrelated to this change either way.

Closes #2358

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Real Kinesis serializes `ApproximateArrivalTimestamp` as a plain JSON decimal on the `x-amz-json-1.1` path; this aligns Floci with that.

## Checklist

- [x] targeted Kinesis handler and integration tests passed locally on the original implementation
- [x] full local suite showed no Kinesis-package regressions on the original implementation
- [ ] full suite re-run after the rebase below

Rebased onto current `main`. #2173 merged in the meantime and already added `epochSeconds(Instant)` to this same class, so the rebase dropped this branch's own duplicate copy and repointed both call sites at the shared one, exactly as flagged. Also brought two review-scoped bits in line with the repository's explicit-type rule: a `var` in `KinesisJsonHandlerTest`/`KinesisIntegrationTest` and an inline fully-qualified `java.util.regex.Pattern` reference are now the concrete type and an import, respectively.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -15,6 +15,8 @@ import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 
+import java.math.BigDecimal;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
@@ -297,7 +299,7 @@ public class KinesisJsonHandler {
                     .put("PartitionKey", rec.getPartitionKey())
                     .put("SequenceNumber", rec.getSequenceNumber())
                     .put("ApproximateArrivalTimestamp",
-                         rec.getApproximateArrivalTimestamp().toEpochMilli() / 1000.0);
+                         epochSeconds(rec.getApproximateArrivalTimestamp()));
         }
         if (continuationSeqNo != null) {
             eventPayload.put("ContinuationSequenceNumber", continuationSeqNo);
@@ -516,11 +518,23 @@ public class KinesisJsonHandler {
             rNode.put("Data", Base64.getEncoder().encodeToString(rec.getData()));
             rNode.put("PartitionKey", rec.getPartitionKey());
             rNode.put("SequenceNumber", rec.getSequenceNumber());
-            rNode.put("ApproximateArrivalTimestamp", rec.getApproximateArrivalTimestamp().toEpochMilli() / 1000.0);
+            rNode.put("ApproximateArrivalTimestamp", epochSeconds(rec.getApproximateArrivalTimestamp()));
         }
         response.put("NextShardIterator", (String) result.get("NextShardIterator"));
         response.put("MillisBehindLatest", ((Number) result.get("MillisBehindLatest")).longValue());
         return Response.ok(response).build();
+    }
+
+    /**
+     * Renders an {@link Instant} as an exact decimal epoch-seconds value (millisecond precision)
+     * rather than a {@code double} - {@code Instant.toEpochMilli() / 1000.0} loses precision and,
+     * for any 2020s-era timestamp, Jackson serializes the resulting large double in scientific
+     * notation (e.g. {@code 1.786959659083E9}), which AWS SDK for Java v2's timestamp unmarshaller
+     * cannot parse despite the response returning HTTP 200. See #2099 for the same defect against
+     * {@code DescribeStream}/{@code DescribeStreamSummary}'s {@code StreamCreationTimestamp}.
+     */
+    private BigDecimal epochSeconds(Instant timestamp) {
+        return BigDecimal.valueOf(timestamp.toEpochMilli(), 3);
     }
 
     private Response handleIncreaseStreamRetentionPeriod(JsonNode request, String region) {
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -294,12 +294,13 @@ public class KinesisJsonHandler {
         ObjectNode eventPayload = objectMapper.createObjectNode();
         ArrayNode recordsNode = eventPayload.putArray("Records");
         for (KinesisRecord rec : records) {
-            recordsNode.addObject()
+            ObjectNode recordNode = recordsNode.addObject()
                     .put("Data", Base64.getEncoder().encodeToString(rec.getData()))
                     .put("PartitionKey", rec.getPartitionKey())
-                    .put("SequenceNumber", rec.getSequenceNumber())
-                    .put("ApproximateArrivalTimestamp",
-                         epochSeconds(rec.getApproximateArrivalTimestamp()));
+                    .put("SequenceNumber", rec.getSequenceNumber());
+            if (rec.getApproximateArrivalTimestamp() != null) {
+                recordNode.put("ApproximateArrivalTimestamp", epochSeconds(rec.getApproximateArrivalTimestamp()));
+            }
         }
         if (continuationSeqNo != null) {
             eventPayload.put("ContinuationSequenceNumber", continuationSeqNo);
@@ -518,7 +519,9 @@ public class KinesisJsonHandler {
             rNode.put("Data", Base64.getEncoder().encodeToString(rec.getData()));
             rNode.put("PartitionKey", rec.getPartitionKey());
             rNode.put("SequenceNumber", rec.getSequenceNumber());
-            rNode.put("ApproximateArrivalTimestamp", epochSeconds(rec.getApproximateArrivalTimestamp()));
+            if (rec.getApproximateArrivalTimestamp() != null) {
+                rNode.put("ApproximateArrivalTimestamp", epochSeconds(rec.getApproximateArrivalTimestamp()));
+            }
         }
         response.put("NextShardIterator", (String) result.get("NextShardIterator"));
         response.put("MillisBehindLatest", ((Number) result.get("MillisBehindLatest")).longValue());
--- a/src/test/java/io/github/hectorvent/floci/services/kinesis/KinesisIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/kinesis/KinesisIntegrationTest.java
@@ -787,6 +787,39 @@ class KinesisIntegrationTest {
     }
 
     @Test
+    @Order(57)
+    void getRecordsReturnsPlainDecimalArrivalTimestamp() {
+        String stream = "at-timestamp-plain-decimal";
+        String shardId = atTimestampCreateStream(stream);
+        atTimestampPutAndGetSequence(stream, "cmVjMA==");
+
+        // Timestamp at epoch 1s (before the record) so GetRecords returns it via the same
+        // AT_TIMESTAMP resolution path exercised by the other tests in this group.
+        String iterator = atTimestampIterator(stream, shardId, 1.0);
+
+        String responseBody = given()
+            .header("X-Amz-Target", "Kinesis_20131202.GetRecords")
+            .contentType(KINESIS_CONTENT_TYPE)
+            .body("{\"ShardIterator\": \"" + iterator + "\"}")
+        .when().post("/")
+        .then().statusCode(200)
+            .body("Records.size()", equalTo(1))
+            .extract().asString();
+
+        // ApproximateArrivalTimestamp must be a plain decimal lexeme - real 2020s-era epoch
+        // seconds render as a large double that Java serializes in scientific notation
+        // (e.g. "1.786959659083E9"), which AWS SDK for Java v2's timestamp unmarshaller cannot
+        // parse despite the response returning HTTP 200 (see floci-io/floci#2099 for the same
+        // defect against DescribeStream/DescribeStreamSummary's StreamCreationTimestamp).
+        var matcher = java.util.regex.Pattern
+            .compile("\"ApproximateArrivalTimestamp\"\\s*:\\s*(-?\\d+(?:\\.\\d+)?)(?=\\s*[,}])")
+            .matcher(responseBody);
+        assertTrue(matcher.find(), "timestamp must be serialized as a plain JSON decimal: " + responseBody);
+        assertFalse(matcher.group(1).contains("E"));
+        assertFalse(matcher.group(1).contains("e"));
+    }
+
+    @Test
     @Order(51)
     void atTimestampBeforeFirstRecordReturnsAll() {
         String stream = "at-timestamp-before";
--- a/src/test/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandlerTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandlerTest.java
@@ -10,12 +10,16 @@ import jakarta.ws.rs.core.Response;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import java.math.BigDecimal;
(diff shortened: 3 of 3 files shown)


sheldon-nadarajah-sft:
Good catch — fixed in 11643b6: both call sites now check for null and omit the field instead of throwing, since `KinesisRecord` can round-trip through hybrid/persistent JSON-backed storage without this field. Added `getRecordsOmitsArrivalTimestampRatherThanThrowingWhenNull` covering it.

pgermosen:
Hi @sheldon-nadarajah-sft, thanks for the thorough writeup and for tracing this down to the same root cause as #2173 — the `epochSeconds(Instant)` approach and the null-guard addition both look correct, and the tests are solid (verified 32/32 `KinesisJsonHandlerTest` + 38/38 `KinesisIntegrationTest` pass, matches your numbers).

One thing worth flagging before merge, though: I built the server from this branch and reproduced the exact scenario from #2358 (the unmodified `KinesisClient` snippet — `createStream` → `putRecord` → `getShardIterator(TRIM_HORIZON)` → `getRecords`) against a running instance, both with and without this fix applied.

In both cases, `GetRecords` succeeds with **no exception**. The scientific-notation unmarshalling crash doesn't reproduce against the real, default `KinesisClient`, because floci routes Kinesis JSON responses through a CBOR conversion layer (`AwsJsonCborController.nodeToSmithyCbor`) whenever the client sends a CBOR request — which the real AWS SDK for Java v2 does by default for Kinesis (there's already a test for this in the repo: `KinesisTest#awsSdkV2UsesRootCborRoute`). That layer encodes any field ending in `Timestamp`/`Timestamps` as a binary CBOR tag(1) value taken directly from the numeric `JsonNode`, so the scientific-notation JSON *text* this PR avoids is never actually serialized to a default client — it only matters for callers that hit the plain `application/x-amz-json-1.1` path directly (e.g. boto3/CLI with `AWS_CBOR_DISABLE=true`, or an older SDK).

More importantly, I noticed the *decoded* `ApproximateArrivalTimestamp` is wrong by a factor of 1000 in both cases — the client reported a `1970-01-21` timestamp instead of the real (`2026-08-17`-ish) one. I checked the raw JSON the server produces directly via curl, and the value it writes is mathematically correct epoch-seconds (matched wall-clock time exactly), so the corruption isn't coming from `KinesisJsonHandler` at all — it looks like it's happening in the CBOR tag(1) fractional-timestamp decode on the client side (or a units mismatch between floci's tag(1) encoding convention and what the SDK expects for a fractional payload). That's a separate, pre-existing bug, not touched by this PR or #2173, and it's arguably more serious since it's silent data corruption rather than a thrown exception.

None of this makes the fix here wrong — it's a legitimate correction for the plain-JSON-protocol path, and consistent with #2173's precedent. But since the PR description's "real, practical consequence" section specifically describes a Java-SDK-v2-breaking exception, I wanted to flag that this doesn't hold for the default client before this merges, in case it changes how you want to scope or describe the fix — and the CBOR timestamp units issue seems worth its own issue if you're up for digging into it, since it'd affect `GetRecords` regardless of this change.

Happy to help narrow down the CBOR decode side if useful.

sheldon-nadarajah-sft:
Thanks for going to the trouble of building the branch and actually reproducing it — you were right on both counts, and digging into your findings turned up the real cause. Filed as #2368; summary here since it directly affects how this PR should be described.

**You were right that this PR's justification doesn't hold.** Confirmed independently: the #2358 reproducer raises no exception with or without this change, for exactly the reason you gave — the default Java SDK sends CBOR, so the scientific-notation JSON text never reaches it. I've rewritten the PR description to drop that claim and state the narrower, accurate scope (plain `x-amz-json-1.1` callers only: boto3/CLI with `AWS_CBOR_DISABLE=true`, older SDKs), which is still a real wire-format deviation worth fixing on the same grounds as #2173.

**Your 1970 observation isn't client-side — it's the mirror half of a units bug in Floci, and there's a request-direction half too that's arguably worse.** I captured the `GetShardIterator` request off the wire from an unmodified client:

```
Content-Type : application/x-amz-cbor-1.1
raw body hex : ...6954696d657374616d70 c1 1b 000001a013accc38
                                      ^^ tag(1)  ^^^^^^^^^^^^^^^^ 1787036486712
decoded      : {"Timestamp": 1787036486712}     <-- epoch MILLISECONDS
```

`handleGetShardIterator` assumes epoch seconds and does `Math.round(asDouble() * 1000)` → `1787036486712000`, i.e. the year 58598. So `getRecords`'s `AT_TIMESTAMP` branch matches nothing, falls through to `startIndex = allRecords.size()`, and returns an empty page plus a valid continuation iterator — `AT_TIMESTAMP` silently yields **zero records forever** for any default Java-SDK client, with no error. The identical sequence via the CLI (plain JSON) returns the record correctly, which is what kept this hidden.

And the response half reproduces your exact symptom:

```
real write time      : 2026-08-18T07:03:06.966Z
SDK-decoded arrival  : 1970-01-21T16:23:56.587Z
```

Root cause for both: AWS's legacy CBOR protocol encodes timestamps as epoch **millis** under tag(1), while `x-amz-json-1.1` uses fractional epoch **seconds** — and Floci treats both as seconds, so it multiplies inbound millis by 1000 and emits outbound seconds where the SDK expects millis.

That also means the "poll returns zero records forever" behaviour I originally attributed to this PR was #2368 all along. Good thing you pushed back on it.

On scoping: I'd suggest keeping this PR as-is (correct for the plain-JSON path, consistent with #2173) and handling #2368 separately, since it's a different code path and the fix likely belongs in the CBOR layer rather than per-handler. Happy to fold them together if you'd rather, and happy to take #2368 on — I've flagged one open design question there about where the conversion should live, given both protocols share the same `JsonNode`.

pgermosen (commented):
@sheldon-nadarajah-sft — a heads-up rather than a review finding, so you are not surprised by it later.

#2173 fixes the same defect in the same class for a different field — `StreamCreationTimestamp` on `DescribeStream` and `DescribeStreamSummary` — and it has just been approved. It adds its own private helper:

```java
private BigDecimal epochSeconds(Instant timestamp) {
    return BigDecimal.valueOf(timestamp.toEpochMilli(), 3);
}
```

That is byte-identical to yours, and I see your javadoc already refers across to `StreamCreationTimestamp`, so you were clearly aware of the overlap when you wrote it.

`epochSeconds` is not in `main` today, so whichever of the two lands first is clean and the second hits a duplicate-method collision on that name. #2173 is ready now — no conflicts, tests green — so on current form it goes first and this branch inherits the dedupe.

The resolution is as easy as it gets: delete your copy and keep theirs, since the implementations are the same character for character. There is nothing to reconcile in behaviour, and nothing to do until #2173 actually merges.

One thing worth knowing while you are here, because it affects how this reads rather than whether it is right. Floci routes Kinesis through a CBOR layer by default, which encodes these timestamps as binary values and never emits the JSON text that the scientific-notation problem lives in — so the failure this prevents does not reproduce against a default `KinesisClient`. It is a real path nonetheless: clients running `AWS_CBOR_DISABLE=true` take it, and the previous code could emit values like `1.785732980986E9` there.

Separately, on that same CBOR path the decoded timestamps come back around 1000x wrong regardless of either fix. That is tracked as #2368 and is nobody's job here — mentioning it only so that if you test through a default client and see a 1970 date, you know it is that and not something you introduced.

I will get to a proper review of this one shortly.

sheldon-nadarajah-sft:
@pgermosen Following up on #2368 - opened #2432 with the actual fix (both directions: `handleGetShardIterator`'s millis-treated-as-seconds on the request side, and `writeCborTimestamp`'s seconds-written-unconverted on the response side, both in `AwsJsonCborController` since that's the one place that knows which wire protocol is in play).

Verified with a genuine end-to-end reproduction (not just a unit-level round-trip): hand-built a `GetShardIterator` `AT_TIMESTAMP` request with a raw CBOR tag(1) millisecond value - independent of Floci's own encoder, so it can't pass by testing against itself - and confirmed it fails against the pre-fix code (`GetRecords` returns 0 records, matching what you and I both saw) and passes against the fix. Also updated `AwsJsonCborSerializerTest`, whose existing assertions had actually locked in the bug as intended behaviour.

No further action needed on this PR - it's still correct and unaffected either way, this is just the loop closing on the CBOR issue we flagged here.

pgermosen:
Quick follow-up on the heads-up from a couple days ago: #2173 has merged now, so it's time for that trivial dedupe.

I confirmed it's live — a real `git merge --no-commit --no-ff origin/main` against this branch succeeds with **zero textual conflict markers** (the two `epochSeconds` additions land at different line ranges, so git auto-merges them), but the merged tree fails `test-compile`:

```
[ERROR] KinesisJsonHandler.java:[543,24] method epochSeconds(java.time.Instant) is already defined in class io.github.hectorvent.floci.services.kinesis.KinesisJsonHandler
```

As you already anticipated, the fix is just dropping your copy of the helper and keeping `main`'s (they're identical).

Aside from that, the rest of the PR looks good — I reverted the null-guard on `ApproximateArrivalTimestamp` and confirmed it genuinely catches the NPE it's meant to prevent, and the full kinesis suite (108 tests) passes clean. Once this is rebased I think it's ready to go.

KenkoGeek (changes requested):
**Medium - this branch now needs to be rebased and deduplicated against current `main`.**

The overlap called out earlier is no longer hypothetical: `main` already contains `KinesisJsonHandler.epochSeconds(Instant)` from the sibling timestamp fix, at the same class and with the same purpose. This PR still adds another method with that signature, so rebasing it as-is will produce a duplicate-method conflict and it cannot land cleanly in its current form.

Please rebase onto current `main`, keep the existing shared-in-class `epochSeconds(...)` helper, and retain only the `ApproximateArrivalTimestamp` call-site changes plus their focused tests. I do not see a reason to combine the separate CBOR timestamp issue into this PR; keeping this fix scoped to the JSON 1.1 serialization path is the right boundary.

pgermosen (approved):
Rebased and deduped against main as KenkoGeek flagged, both call sites now point at the shared epochSeconds from #2173 instead of a second copy. Good catch tracing the CBOR units bug back to its real cause too, that one's since landed clean via #2432 and unrelated to this either way.

Also swapped a couple of var locals and a fully-qualified Pattern reference in the new tests for the explicit type and an import, since AGENTS.md is strict about both. Small, focused diff and CI's green across the board.

PR #2361, opened 2026-08-17
docs(testcontainers-python): fix outdated import statements

ApprenticeofEnder:
## Summary

The Python Testcontainer docs were outdated in comparison with the [PyPI docs page](https://pypi.org/project/testcontainers-floci/). This PR resolves the gap.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -658,7 +658,7 @@ pip install testcontainers-floci
 
 ```python
 import boto3
-from testcontainers_floci import FlociContainer
+from floci import FlociContainer
 
 
 def test_s3_create_bucket():
--- a/docs/testcontainers/python.md
+++ b/docs/testcontainers/python.md
@@ -20,7 +20,7 @@ uv add --dev testcontainers-floci
 
 ```python
 import boto3
-from testcontainers_floci import FlociContainer
+from floci import FlociContainer
 
 
 def test_s3_create_bucket():
@@ -46,7 +46,7 @@ Use a session-scoped fixture so the container starts once and is shared across a
 ```python
 import pytest
 import boto3
-from testcontainers_floci import FlociContainer
+from floci import FlociContainer
 
 
 @pytest.fixture(scope="session")
@@ -85,7 +85,7 @@ def test_upload_object(s3_client):
 import pytest
 import boto3
 import json
-from testcontainers_floci import FlociContainer
+from floci import FlociContainer
 
 
 @pytest.fixture(scope="session")
@@ -125,7 +125,7 @@ def test_send_and_receive_message(sqs_client):
 ```python
 import pytest
 import boto3
-from testcontainers_floci import FlociContainer
+from floci import FlociContainer
 
 
 @pytest.fixture(scope="session")
@@ -191,7 +191,7 @@ Place shared fixtures in `conftest.py` so every test module picks them up automa
 # conftest.py
 import pytest
 import boto3
-from testcontainers_floci import FlociContainer
+from floci import FlociContainer
 
 
 @pytest.fixture(scope="session")


pgermosen:
Hi @ApprenticeofEnder, thanks for the docs fix! The "Commits follow Conventional Commits" check is failing on this PR, which is why CI hasn't gone green.

The issue is the scope: `docs(testcontainers/python): fix outdated import statements` uses a `/` inside the scope parentheses, and this repo's Conventional Commits check doesn't allow that character in the scope — only letters, digits, `-`, `_`, `.`, and `,`. Since PR titles get squash-merged as the commit message here, the title needs the same fix.

Something like `docs(testcontainers-python): fix outdated import statements` (or just `docs: fix outdated Python Testcontainer import statements` without a scope) would pass. See [CONTRIBUTING.md's Commit Message Format section](https://github.com/floci-io/floci/blob/main/CONTRIBUTING.md#commit-message-format) for the format and more examples.

Once the title's updated, that check should go green on its own — no code changes needed.

ApprenticeofEnder:
Cheers, I'll update when I get the chance!

hectorvent:
Thank you for this, and the fix is verified right at the source: the package repo's own `floci/` layout and README use exactly the import you switched to.

**(blocking)** Two small things that travel together. README.md line 661 still has the old `from testcontainers_floci import FlociContainer`, the most visible copy of the same problem, so it is worth fixing in this PR. And when you push that commit, please also retitle the PR without the `/` in the scope (for example `docs(testcontainers-python): fix outdated import statements`): with one commit GitHub's squash message takes your clean commit title, but with two it switches to the PR title, and the scope rule pgermosen described above would then reject the slash after merge.

Both together are a couple of minutes. No other findings, this is ready once they land.

pgermosen (approved):
Thanks — this is a small change that fixes a real break, and I checked it against the package rather
than against the PyPI page you cited, so the two agree independently.

## The import is right, and the old one could never have worked

`floci-io/testcontainers-floci-python` has a `floci/` package directory at its root — there is no
`testcontainers_floci` module anywhere in it — and `floci/__init__.py` is:

```python
from floci.container import FlociContainer

__all__ = ["FlociContainer"]
```

so `from floci import FlociContainer` is exactly the supported import.

The mismatch that caused this is in `pyproject.toml`, where the *distribution* is named separately
from the package it installs:

```toml
[project]
name = "testcontainers-floci"
```

`pip install testcontainers-floci` was correct all along and stays unchanged; only the import line
was wrong. Which means every Python example in these docs failed at the second line with
`ModuleNotFoundError: No module named 'testcontainers_floci'` — this isn't a tidy-up, it's the
difference between the examples running and not running.

## The sweep is complete

Six occurrences existed on `main` and this PR changes six:

| File | Occurrences |
|---|---|
| `README.md` | 1 (`:667`) |
| `docs/testcontainers/python.md` | 5 (`:23`, `:49`, `:88`, `:128`, `:194`) |

Nothing else in the repo references the underscore form, so there is no third file left behind
holding the old spelling.

Greptile's point about the README lagging the guide is closed by `afbcbc50` — that was the right
catch, since a half-applied fix here would have been worse than none: a reader following the README
would hit the import error while the guide told them something different.

## (suggestion) A line about why the two names differ would stop this coming back

Not a change I'd hold the PR for, and entirely your call.

A distribution named `testcontainers-floci` that you import as `floci` is a genuine surprise — it's
presumably how the docs drifted in the first place, since `testcontainers_floci` is the spelling
you'd guess from the install line directly above it. Someone reading
`pip install testcontainers-floci` followed by `from floci import FlociContainer` may well read that
as a typo and "fix" it back.

A short parenthetical after the install command in `docs/testcontainers/python.md` — that the
package installs as `testcontainers-floci` and imports as `floci` — would make the next reader
confident it's deliberate. The Node and Java pages don't have this problem, so it's specific to this
one.

## Nit

- The checklist leaves `./mvnw test` and the integration-test box unticked rather than marked N/A.
  Neither applies to a docs-only change, and CONTRIBUTING asks that a PR without tests say why —
  "documentation only" is that reason. Same for the empty **AWS Compatibility** section: worth an
  explicit N/A rather than blank, so a reader can tell it was considered rather than skipped.

## Note

- The leftover `<!-- What does this PR do? -->` comments are invisible in the rendered description
  and you filled the sections underneath them, so nothing to do there.

PR #2362, opened 2026-08-17
feat(cloudwatch): Adds support for orderBy, pagination, limit in DescribeLogStreams response

floverfelt:
## Summary

Adds support when describing log streams using the orderBy and limit clauses. Adds support for pagination.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against SDK 2.52.0.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudWatchLogsDescribeStreamsOrderingTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudWatchLogsDescribeStreamsOrderingTest.java
@@ -1,0 +1,134 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClient;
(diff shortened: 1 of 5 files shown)


pgermosen (approved):
All three Greptile P1s (offset cursor losing position on reorder, malformed tokens silently restarting pagination, mutable sort-key cursor skipping streams) are addressed by a single coherent fix: a snapshot-based pagination design. The first page freezes membership/order into a server-side snapshot (TTL-evicted), while subsequent pages resume by re-reading current attributes for each name in that frozen order — so reordering between pages can neither skip nor duplicate a stream, and attributes still reflect live data. Malformed/foreign tokens correctly throw InvalidParameterException via a single, non-duplicated invalidNextToken() helper.

Verified every AWS-behavior claim directly against the real botocore logs model: the LastEventTime-ordering-can't-combine-with-prefix constraint, limit bounds (1-50, default 50), and the exact orderBy/nextToken field names — all exact matches. There's a real regression test reproducing the exact reordering scenario, asserting both that the stream isn't skipped and its attributes stay live.

Ran CloudWatchLogsServiceTest (72/72) and CloudWatchLogsHandlerTest (26/26) locally — 0 failures. Clean merge, docs-check passes, all CI green, no CODEOWNERS-protected paths.

PR #2364, opened 2026-08-18
fix(core): clear in-memory service state on state reset

hectorvent:
## Summary

`/_floci/state/reset` calls `clear()` on every `Resettable`, then
`StorageFactory.clearAll()`. A service that holds its state in a plain map and does
not declare `Resettable` is reached by neither, so the reset silently leaves data
behind — worse than not resetting at all, because the survivors reference resources
that *were* cleared.

Declares `Resettable` on the services whose live state is an in-memory map
(implementors 22 → 43), and **fences the services that keep working across a reset**
so they cannot write the state back afterwards.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Bug fix, so the incorrect behaviour: after a reset, `DescribeStacks` still returned a
stack, and `ListTasks` still returned an ECS task whose cluster had been cleared. No
wire format changes — this is emulator lifecycle behaviour only.

## Checklist

- [x] `./mvnw test` passes locally — 10621 passed / 0 failures / 0 errors / 8 skipped

## Fencing, not just clearing

Clearing bookkeeping is not enough where a background process owns it:

| Service | Why clearing alone was wrong | Fix |
|---|---|---|
| IoT | Dropped sessions without disconnecting, leaving clients attached to a broker that had forgotten them | Close each endpoint first, as `stop()` and `disconnectClient()` already do |
| ECS | The reconciler keeps running across a reset by design, and could race `clear()` | Shared lock; reconciler still survives the reset |
| CloudControl | `CreateResource` provisions on a background thread and recorded its result after the reset | Reset epoch captured at submit; stale writes dropped. `requestOrder` also cleared |
| CloudFormation | `executeTemplate` runs async and re-registers `exports` (and nested stacks) after the reset | Cancel in-flight, wait ≤2s, then clear, with an epoch guard on export registration |

Epoch rather than interruption for the two long-running paths: provisioning can take a
while and a reset must not block on it.

## Verification — what is and isn't pinned by a test

| Behaviour | Test that fails without the fix |
|---|---|
| CFN stacks survive a reset | ✅ `StateResetCoverageIntegrationTest` |
| ECS tasks survive a reset | ✅ same |
| ECS lock does not deadlock | ✅ `StateResetFencingTest` |
| IoT sessions actively closed | ❌ reasoned; mirrors the existing `stop()` path |
| CloudControl epoch drop | ❌ needs a seam to hold provisioning open across a reset |
| CFN quiesce + epoch guard | ❌ same |

The three unpinned fences are reasoned and documented in the commit body so they are
not simplified back out, but I would rather state that they lack direct coverage than
imply otherwise. Closing them properly needs a test seam that does not exist yet;
happy to add it here if preferred, or track it separately.

## Deliberately *not* cleared

Clearing these would itself be the defect: `CloudFormationResourceRegistry` (the CDI
provisioner registry — clearing it permanently breaks provisioning), `PricingService`
catalogs (static reference data), and per-request parse state (`VtlTemplateEngine`,
`LogsInsightsQuery`, `DynamoDbPartiQLHandler`). Container managers keep the existing
split: reset clears state, `ContainerTeardown` owns containers at shutdown.

`IotPublishEventRecorder` already had a correct `clear()` — it simply never declared
`Resettable`, so a reset never invoked it.

DynamoDB is a **leak**, not a visibility bug: items survive `clearAll()` but are
unreachable because reads validate the table and `CreateTable` re-initialises the entry.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/lifecycle/StateResetCoverageIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/lifecycle/StateResetCoverageIntegrationTest.java
@@ -1,0 +1,157 @@
+package io.github.hectorvent.floci.lifecycle;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+import io.restassured.config.EncoderConfig;
+import io.restassured.http.ContentType;
+import io.restassured.parsing.Parser;
+import io.restassured.specification.RequestSpecification;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.hasSize;
+
+/**
+ * {@code /_floci/state/reset} clears every {@link io.github.hectorvent.floci.core.common.Resettable}
+ * plus every {@code StorageBackend}. A service that keeps its state in a plain map and does not
+ * implement {@code Resettable} is therefore missed, and the reset silently leaves data behind —
+ * which is worse than not resetting, because surviving resources reference cleared ones.
+ *
+ * <p>{@link EmulatorInfoControllerIntegrationTest} already covers a purely storage-backed service
+ * (SSM). These cases cover services whose live working copy is an in-memory map.
+ */
+@QuarkusTest
+class StateResetCoverageIntegrationTest {
+
+    private static final String JSON_1_0 = "application/x-amz-json-1.0";
+    private static final String JSON_1_1 = "application/x-amz-json-1.1";
+
+    @BeforeAll
+    static void registerAwsJsonParsers() {
+        RestAssured.registerParser(JSON_1_0, Parser.JSON);
+        RestAssured.registerParser(JSON_1_1, Parser.JSON);
+    }
+
+    private void reset() {
+        given().when().post("/_floci/state/reset")
+                .then().statusCode(200).body("status", equalTo("OK"));
+    }
+
+    /** RestAssured has no built-in serializer for the x-amz-json content types; send raw text. */
+    private static RequestSpecification rpc(String contentType, String target, String body) {
+        return given()
+                .config(RestAssured.config().encoderConfig(EncoderConfig.encoderConfig()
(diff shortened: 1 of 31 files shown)


src/test/java/io/github/hectorvent/floci/lifecycle/StateResetCoverageIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+        // The table itself is storage-backed and does go away, so re-create it and confirm the
+        // item did not come back with it. Items live in a plain map, which is the actual subject.
+        rpc(JSON_1_0, "DynamoDB_20120810.CreateTable", CREATE_TABLE)
+                .when().post("/").then().statusCode(200);
+
+        rpc(JSON_1_0, "DynamoDB_20120810.Scan", """
+                {"TableName":"reset-items"}
+                """).when().post("/")
+                .then().statusCode(200)
+                .body("Count", equalTo(0));
```
pgermosen: This test does not actually exercise `DynamoDbService.clear()` — I disabled it entirely and reran just this test; it still passed.

The reason: `createTable()` does `itemsByTable.put(scopedItemsKey(storageKey), new ConcurrentSkipListMap<>())` — an unconditional `put()`, not `computeIfAbsent()`. Since this test re-creates the table under the *same name* after reset, that `put()` always replaces whatever was there, stale or not, independent of whether `clear()` ran. There is no code path where this assertion could fail regardless of the fix.

That is consistent with the PR description calling this a memory leak rather than user-visible data loss, so it is not testable through `Scan` the way `reset_clearsEcsTasks`/`reset_clearsCloudFormationStacks` are. A test that would actually catch a regression here: reuse a `ClientRequestToken` across the reset boundary on `TransactWriteItems` with two *different* request bodies. `txIdempotency` keys on `accountId::region::token` and is checked before the table/item state, so if `clear()` regressed, the second call (post-reset, different body, same token) would incorrectly throw `IdempotentParameterMismatchException` instead of succeeding — an outcome observable through the public API, unlike the current `Scan`-after-recreate shape.

pgermosen (commented):
Verified all three new tests directly (disabling each service's clear() and re-running): reset_clearsEcsTasks and reset_clearsCloudFormationStacks both genuinely fail without their respective fixes. reset_clearsDynamoDbItems does not — see the inline comment for why and a suggested alternative.

hectorvent:
Closing. I will revisit this improvement later, downgrading priority.

PR #2366, opened 2026-08-18
fix(msk): persist CreateCluster broker node groups, encryption, client auth, logging, configuration and tags

### LINKED ISSUE #2301: [BUG] msk: CreateCluster discards broker node groups, encryption, client auth and tags
## Service

MSK (Managed Streaming for Apache Kafka)

## AWS API Action

`CreateCluster`, `DescribeCluster`

## Expected behavior

`CreateCluster` accepts `brokerNodeGroupInfo` (instance type, client subnets, security groups,
storage), `numberOfBrokerNodes`, `encryptionInfo`, `clientAuthentication`, `enhancedMonitoring`,
`loggingInfo`, `configurationInfo` and `tags`, stores them, and `DescribeCluster` returns them
under `clusterInfo`.

## Actual behavior

`MskController.createCluster` reads exactly two fields:

```java
String clusterName = (String) request.get("clusterName");
String kafkaVersion = (String) request.get("kafkaVersion");
MskCluster cluster = mskService.createCluster(clusterName, kafkaVersion);
```

Everything else in the request body is discarded. `MskService.createCluster` sets only the ARN,
name, resolved Kafka version, account ID and volume ID.

Notably `MskCluster` **already has** `numberOfBrokerNodes` and `tags` fields — they are simply
never populated, so they read back as `0` and unset. `brokerNodeGroupInfo`, `encryptionInfo` and
`clientAuthentication` have no model representation at all.

The `/api/v2/clusters` handler has the same problem: it unwraps `provisioned` only far enough to
find `kafkaVersion` and drops the rest, with a `// Simple mapping to V1 for now` comment.

**Impact.** An `aws_msk_cluster` applies successfully and then reads back missing most of its
configuration, so `terraform plan` reports a diff on every run and never converges. `tags` in
particular is read on every refresh by the provider's tagging interceptor.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 kafka create-cluster \
  --cluster-name example \
  --kafka-version 3.6.0 \
  --number-of-broker-nodes 3 \
  --broker-node-group-info '{"InstanceType":"kafka.m5.large","ClientSubnets":["subnet-aaa","subnet-bbb"]}' \
  --tags Environment=example

aws --endpoint-url http://localhost:4566 kafka describe-cluster --cluster-arn <arn>
# NumberOfBrokerNodes is 0, Tags is absent, BrokerNodeGroupInfo is absent
```

## Suggested fix

Parse and persist the request fields, starting with the ones the model already has
(`numberOfBrokerNodes`, `tags`) and then adding `brokerNodeGroupInfo`, `encryptionInfo` and
`clientAuthentication` to `MskCluster`. `DescribeCluster` and `ListClusters` should echo them
under `clusterInfo`.

Keeping the emulated broker itself unchanged is fine — this is a metadata round-trip problem,
not a Kafka behavior problem.

## Scope note

Deliberately separate from the MSK configuration issues (`CreateConfiguration` and the revision
endpoints). Those add a new resource surface; this fixes the existing cluster one. They should
not be combined.

## Environment

## AWS Documentation

https://docs.aws.amazon.com/msk/1.0/apireference/clusters.html

mansiverma897993:
## Problem

`MskController.createCluster` (both `/v1/clusters` and `/api/v2/clusters`) only read `clusterName` and `kafkaVersion`, discarding `brokerNodeGroupInfo`, `numberOfBrokerNodes`, `encryptionInfo`, `clientAuthentication`, `enhancedMonitoring`, `loggingInfo`, `configurationInfo` and `tags`. `DescribeCluster` therefore echoed `numberOfBrokerNodes: 0`, no tags, and no node-group/encryption/auth info, so `terraform plan` never converged and the tagging interceptor lost tags on every refresh.

## Change

Persist and echo the full `CreateCluster` request metadata:

- Added typed models mirroring the AWS MSK shape: `BrokerNodeGroupInfo` (instance type, subnets, security groups, storage, connectivity), `EncryptionInfo`, `ClientAuthentication`, `LoggingInfo`, `ConfigurationInfo`, plus the nested storage/encryption/auth/logging sub-objects.
- Added `CreateClusterRequest` (v1) and `CreateClusterV2Request` + `ProvisionedRequest` (v2) bound directly as JAX-RS request bodies (same pattern as EKS).
- Extended `MskCluster` with `brokerNodeGroupInfo`, `encryptionInfo`, `clientAuthentication`, `enhancedMonitoring`, `loggingInfo` and `configurationInfo`; `numberOfBrokerNodes` and `tags` are now populated from the request.
- `MskService.createCluster(...)` persists all fields; v1 and v2 both round-trip through `DescribeCluster`/`ListClusters` under `clusterInfo`.
- Marked internal fields (`bootstrapBrokers`, `containerId`, `volumeId`) `@JsonIgnore` so the client-facing response no longer leaks them.

The emulated broker runtime is unchanged â€” this is a metadata round-trip fix only. Deliberately excludes the MSK `CreateConfiguration`/revision work (separate resource surface, per scope note).

## Test report

- `mvn -q -Dtest=MskServiceTest,MskControllerIntegrationTest test` â€” **15 tests, 0 failures** (8 unit + 7 integration).
- New tests cover: v1/v2 `DescribeCluster` echoes broker node group (instance type/subnets/SG/storage), `numberOfBrokerNodes`, encryption (at-rest + in-transit), client auth (SASL/TLS/unauth), enhanced monitoring, logging (S3/CloudWatch), configuration (arn/revision) and tags; and that internal fields are no longer leaked.

Closes #2301


pgermosen:
Hi @mansiverma897993, thanks for tackling this — checked whether the underlying problem is still real and whether it overlaps with the recently-merged MSK Configuration CRUD work (#2336) or the still-open revisions PR (#2344). It doesn't: those are about the separate `Configuration` resource, while `createCluster`/`createClusterV2` on current `main` genuinely still only read `clusterName`/`kafkaVersion` and discard everything else, exactly as described. So the fix is still needed.

Two things worth addressing before this is ready, though — both verified directly rather than just relayed from the review bot:

**1. `@JsonIgnore` on `MskCluster` breaks persistence, not just API leakage.** `bootstrapBrokers`, `containerId`, `accountId`, and `volumeId` are marked `@JsonIgnore` to keep them out of client responses, but `WalStorage`, `PersistentStorage`, and `HybridStorage` all serialize with a plain `new ObjectMapper()` — no storage-specific view or mixin that would bypass this. So the same annotation strips those fields when persisting and reloading the cluster too. After a restart with non-mock/persistent storage, a cluster would come back with no container ID or volume ID, breaking `GetBootstrapBrokers` and leaving the Docker container/volume untracked and orphaned. This needs a response-layer-only exclusion (a dedicated response view/DTO, or filtering in the controller when building the JSON) rather than a blanket `@JsonIgnore` on the model itself.

**2. V2 responses are missing the `Provisioned` envelope.** Checked this against the real MSK API model directly: `Cluster` (the shape `DescribeClusterV2`/`ListClustersV2` return under `ClusterInfo`) has a `Provisioned` member — the broker-node-group, encryption, client-auth, monitoring, logging, and configuration fields all nest inside it, not flat on the cluster object. Both `MskController`'s v1 and v2 handlers here currently return everything flat under `clusterInfo`. A real AWS SDK v2 client (which is what Terraform's MSK provider actually calls) would still fail to find these fields — which is the same "plan never converges" problem this PR sets out to fix, just relocated to v2 callers instead of resolved for them.

There's also a real merge conflict against current `main` (not just a stale GitHub status — confirmed with an actual merge attempt): `MskController.java`, `MskService.java`, and both test files, since this branch predates #2336 landing. Could you rebase once the above is sorted?

Happy to take another look once these are in.

pgermosen (approved):
@mansiverma897993 — approving. Everything raised in August is closed, and the two findings from
the parallel MSK work landed here too.

**`@JsonIgnore` breaking persistence.** Gone. Internal bookkeeping stays serializable, and the
client-facing shapes are built by explicit `toClusterViewV1` / `toClusterViewV2` view methods
that emit only what they name. That's the response-layer-only exclusion the original comment
asked for, and it means a restart no longer loses container and volume IDs.

**The v2 `Provisioned` envelope.** `toClusterViewV2` nests `provisionedFields(cluster)` under
`provisioned` with a `clusterType` discriminator, while v1 splats the same helper flat. Sharing
one field list between the two envelopes is better than duplicating it — the shapes can't drift
apart later.

**Rebase.** Done, and `MERGEABLE`.

Two more that came out of reviewing the same surface elsewhere, both now fixed here:

**The broker-count ceiling is gone (`6d95f78e`).** `MAX_BROKER_NODES = 15` came from the
`__integerMin1Max15` shape in the service model, but the MSK quota page gives 30 brokers per
ZooKeeper cluster and 60 for KRaft, adjustable upward, and the REST API reference documents no
range on `numberOfBrokerNodes` at all — while documenting them for neighbouring members. Keeping
only the `min: 1` floor is right: a zero or negative count is unambiguously invalid, and an
emulator that accepts slightly more than AWS costs nobody anything, whereas one that rejects a
valid 30-broker cluster blocks real work.

**`invalidParameter` is now on the wire.** MSK's `Error` schema is two members — `message` and
*"the parameter that caused the error"* — and `badRequest()` now carries it through
`AwsException`'s `extendedData`, so a validation failure names the member it rejected instead of
leaving the caller to guess.

## One thing for the maintainers, not for this branch

#2571 fixes a strict subset of this — `numberOfBrokerNodes` and `tags` on `CreateCluster` — and
both rewrite `MskController.createCluster` / `createClusterV2` and `MskService.createCluster`.
Whichever lands second will conflict hard. This PR is the earlier and far more complete of the
two, so #2571 should be closed as superseded rather than rebased.

PR #2367, opened 2026-08-18
fix(apigatewayv2): preserve trailing slash in HTTP API event path

### LINKED ISSUE #2136: [BUG] API Gateway V2 (HTTP API): trailing slash stripped from rawPath (rawPath /thing/ -> /thing)
## Service

API Gateway V2 (HTTP API)

## AWS API Action

Lambda proxy integration invoke (payload format 2.0), and REQUEST authorizer invoke on payload format 1.0.

## Expected behavior

A request to `/thing/` on an HTTP API delivers the trailing slash to the Lambda, because `rawPath` is by contract the raw, unmodified path:

```json
{ "rawPath": "/thing/" }
```

Routers commonly treat `/thing` and `/thing/` as distinct routes, so a v2 Lambda needs to be able to tell them apart.

## Actual behavior

The slash is stripped and the Lambda receives `rawPath: "/thing"`. A v2 Lambda cannot distinguish `/thing` from `/thing/` at all.

Cause is the same JAX-RS `{proxy}` path-param binding that #1557 covered for REST APIs. `dispatchV2` rebuilds the path from the stripped binding at [`ApiGatewayExecuteController.java:1219`](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java#L1219):

```java
String path = "/" + (proxy == null ? "" : proxy);
```

and that value becomes the event's `rawPath` at lines 1728 and 1840. `buildRequestAuthorizerEventV1` (line 1680, used for v2 authorizers on payload format 1.0) has the same problem in its `methodArn`, `resource`, `path` and `requestContext.path`.

## Impact

#1863 fixed this for REST (v1) APIs, so as of that PR a REST API preserves the trailing slash and an HTTP API silently does not. That inconsistency between the two API types is the confusing part, and it is worth closing so v1 and v2 do not drift.

## Suggested fix

`preserveTrailingSlash` landed in #1863 as a static package-private helper, so v2 can reuse it as-is:

```java
String path = preserveTrailingSlash(
        "/" + (proxy == null ? "" : proxy),
        uriInfo.getRequestUri().getRawPath());
```

One caveat: `apiGatewayV2Service.findMatchingRoute` (called at line 1216 area) would then receive the slash-preserving path, so route matching should keep using the normalized value. That is the same split #1863 made explicit for v1, where the event fields get the slash-preserving path while resource matching and path-parameter extraction stay on the normalized one.

## Reproduction

```bash
# Create an HTTP API with a Lambda proxy integration on $default, then:
curl -s http://localhost:4566/restapis/<apiId>/<stage>/_user_request_/thing/
# Lambda receives rawPath "/thing", expected "/thing/"
```

## Environment

- Floci version / image tag: `main` (1.5.34)
- How you're running Floci: `./mvnw quarkus:dev`

---

Filed as the follow-up @abanna suggested while reviewing #1863, which deliberately scoped itself to v1. Happy to pick this up.

ravi-arnan:
## Summary

Closes #2136.

A trailing slash is stripped from the path delivered to an HTTP API (V2) Lambda, so a v2 handler cannot tell `/thing` from `/thing/`. #1863 fixed this for REST (V1) APIs, which left the two API types disagreeing: as of 1.7.0 a REST API preserves the slash and an HTTP API silently does not.

The cause is the same JAX-RS `{proxy}` path-param binding. `dispatchV2` rebuilds the path from the stripped binding:

```java
String path = "/" + (proxy == null ? "" : proxy);
```

and that value reaches `rawPath` and `requestContext.http.path` in the proxy event, and the path fields of both REQUEST authorizer event shapes.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## The fix

`preserveTrailingSlash` already exists from #1863, so this reuses it in the three V2 event builders: `buildV2ProxyEvent`, `buildRequestAuthorizerEventV2` and `buildRequestAuthorizerEventV1` (the 1.0 shape used for v2 authorizers).

**Deliberately not applied in `dispatchV2` itself**, which is what the issue originally suggested. Preserving inside the event builders is what `buildProxyEvent` already does for V1, and it means `findMatchingRoute` and `extractV2PathParams` keep the normalized path by construction rather than by remembering to pass the right variable.

The field split mirrors the V1 settlement reached in review on #1863 (38fcca2f):

| Field | Value | Why |
|---|---|---|
| `rawPath`, `requestContext.http.path`, 1.0 `path` / `requestContext.path` | preserved | by contract the raw, unmodified path |
| `routeArn` / `methodArn` | normalized | matched against IAM-style policy resources, where a stray trailing slash silently fails an authorizer's wildcards |
| `pathParameters`, `resource`, `requestContext.resourcePath` | normalized | they come from the matcher, which runs on the normalized path |

The one field I am least sure about is `resource` in the 1.0 authorizer shape. Unlike REST, where `resource` is the resource template, here it mirrors the request path, so it could reasonably go either way. I kept it normalized to match `methodArn`; happy to flip it if you read AWS differently.

## AWS Compatibility

Incorrect behavior: a request to `/thing/` on an HTTP API delivered `rawPath: "/thing"`. AWS delivers `rawPath: "/thing/"`, because `rawPath` is by contract the raw path. Routers commonly treat `/thing` and `/thing/` as distinct routes, so the handler could not distinguish them at all.

Route matching is unaffected: the route key stays `GET /hello` and still matches a request to `/hello/`, mirroring AWS routing `/x/` to the `/x` route while keeping the slash in the delivered event.

## Verification

- **New unit test** `BuildV2ProxyEventTrailingSlashTest`, 5 cases: `rawPath` and `requestContext.http.path` keep the slash, `pathParameters` stay normalized, and two controls (no trailing slash, root path) stay unchanged. Fails 3/5 on unpatched `main`, passes 5/5 with the change.
- **New end-to-end test** `HttpApiRequestAuthorizerTest.trailingSlashReachesTheLambdaRawPath`, which drives a real request through the actual JAX-RS binding to a Lambda that echoes `event.rawPath`. This is the part a mocked `UriInfo` cannot prove: that the raw request URI still carries the slash at the point the binding has already stripped it. Run against `main` with only the test file applied, it fails with `expected: </hello/> but was: </hello>`; with the change it passes, and the other 17 tests in that class are unaffected either way.
- The whole apigateway and apigatewayv2 surface (`ApiGateway*Test,BuildV2*Test,HttpApi*Test`) is green: **639 tests, 0 failures**.
- Wider run: 2491 tests across 205 classes, whose only failure is `DocDbIntegrationTest` (12 failures, 3 errors). That one cannot start its DocumentDB container on my machine (`lookup auth.docker.io ... i/o timeout` pulling `mongo:7`) and it fails **identically on unpatched `main`**, so it is my network rather than this change. `CloudFormationIntegrationTest` also showed one `SocketTimeout` in that run and passes 131/131 when run on its own, on both trees.

## Checklist

- [x] `./mvnw test` passes locally, with the one pre-existing container-pull failure noted above

## Out of scope

The `HTTP_PROXY` integration path (`dispatchHttpProxyV2`) also forwards the normalized path to the backend. That is a separate question from the Lambda event contract this issue is about, and it is not measured here.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -1803,12 +1803,19 @@ public class ApiGatewayExecuteController {
     private String buildRequestAuthorizerEventV1(String httpMethod, String path,
                                                   String apiId, String stageName, String region,
                                                   HttpHeaders headers, UriInfo uriInfo) {
+        // The JAX-RS {proxy} binding strips a trailing slash before dispatchV2 rebuilds the
+        // path, so recover it from the raw request URI for the delivered path fields. Same split
+        // as the REST authorizer event (see toAuthorizerEvent): methodArn and resource keep the
+        // normalized path, because methodArn is matched against IAM-style policy resources where
+        // a stray trailing slash would silently fail an authorizer's wildcards.
+        String preservedPath = preserveTrailingSlash(path, uriInfo.getRequestUri().getRawPath());
+
         ObjectNode event = objectMapper.createObjectNode();
         event.put("version", "1.0");
         event.put("type", "REQUEST");
         event.put("methodArn", buildMethodArn(region, apiId, stageName, httpMethod, path));
         event.put("resource", path);
-        event.put("path", path);
+        event.put("path", preservedPath);
         event.put("httpMethod", httpMethod);
 
         putSingleValueHeaders(event, headers);
@@ -1824,7 +1831,7 @@ public class ApiGatewayExecuteController {
         ctx.put("accountId", regionResolver.getAccountId());
         ctx.put("apiId", apiId);
         ctx.put("httpMethod", httpMethod);
-        ctx.put("path", path);
+        ctx.put("path", preservedPath);
         ctx.put("resourcePath", path);
         ctx.put("stage", stageName);
         ctx.put("requestId", UUID.randomUUID().toString());
@@ -1843,12 +1850,17 @@ public class ApiGatewayExecuteController {
     private String buildRequestAuthorizerEventV2(String httpMethod, String path, String routeKey,
                                                   String apiId, String stageName, String region,
                                                   HttpHeaders headers, UriInfo uriInfo) {
+        // rawPath is by contract the raw, unmodified path, so recover the trailing slash the
+        // JAX-RS {proxy} binding stripped. routeArn keeps the normalized path for the same reason
+        // methodArn does in the 1.0 shape above.
+        String preservedPath = preserveTrailingSlash(path, uriInfo.getRequestUri().getRawPath());
+
         ObjectNode event = objectMapper.createObjectNode();
         event.put("version", "2.0");
         event.put("type", "REQUEST");
         event.put("routeArn", buildMethodArn(region, apiId, stageName, httpMethod, path));
         event.put("routeKey", routeKey != null ? routeKey : "$default");
-        event.put("rawPath", path);
+        event.put("rawPath", preservedPath);
         event.put("rawQueryString", uriInfo.getRequestUri().getRawQuery() != null
                 ? uriInfo.getRequestUri().getRawQuery() : "");
 
@@ -1887,7 +1899,7 @@ public class ApiGatewayExecuteController {
 
         ObjectNode http = ctx.putObject("http");
         http.put("method", httpMethod);
-        http.put("path", path);
+        http.put("path", preservedPath);
         http.put("protocol", "HTTP/1.1");
         http.put("sourceIp", "127.0.0.1");
         http.put("userAgent", headers.getHeaderString("User-Agent") != null
@@ -1991,10 +2003,17 @@ public class ApiGatewayExecuteController {
                                      String apiId, String region, String stageName,
                                      HttpHeaders headers, UriInfo uriInfo,
                                      byte[] body, String requestId, Map<String, String> jwtClaims) {
+        // The JAX-RS {proxy} binding strips a trailing slash, but rawPath is by contract the
+        // raw path and routers treat /x and /x/ as distinct routes. Recover it from the raw
+        // request URI for the event path fields. Route matching in dispatchV2 and the
+        // pathParameters extraction below continue to use the normalized `path`, mirroring what
+        // buildProxyEvent already does for REST (V1).
+        String preservedPath = preserveTrailingSlash(path, uriInfo.getRequestUri().getRawPath());
+
         ObjectNode event = objectMapper.createObjectNode();
         event.put("version", "2.0");
         event.put("routeKey", routeKey != null ? routeKey : "$default");
-        event.put("rawPath", path);
+        event.put("rawPath", preservedPath);
 
         MultivaluedMap<String, String> queryParams = uriInfo.getQueryParameters();
         event.put("rawQueryString", uriInfo.getRequestUri().getRawQuery() != null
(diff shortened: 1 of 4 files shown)


pgermosen:
Hi @ravi-arnan, this is a clean, well-verified fix. Checked the field split directly against the diff — only the raw-path-contract fields (`rawPath`, `requestContext.http.path`, the 1.0 authorizer's `path`) get the preserved value, while `pathParameters`, `routeArn`/`methodArn`, and `resourcePath` correctly stay normalized. Also confirmed the "fails 3/5 on unpatched main" claim exactly: reverted just this file and got precisely 3 failures out of 5 in `BuildV2ProxyEventTrailingSlashTest`, then restored and got 5/5 + 18/18. Clean merge against current main, docs-check passes.

On the one field you flagged as uncertain — `resource` in the 1.0 authorizer shape — I checked AWS's own documented example for that exact payload. It shows `"resource": "/request"` and `"path": "/request"` as the *same* value. Since HTTP APIs don't have a REST-style resource-template tree the way REST APIs do, `resource` here looks like it's meant to mirror the actual request path rather than act as a template, so it may need the trailing slash too, to stay consistent with `path`. Worth a look, though I don't think it should block this otherwise.

ravi-arnan:
Thanks for reading it that closely, and for re-running the 3/5 independently.

You are right about `resource`, and there is stronger evidence than the AWS example: @abanna
named that exact field in his #1863 review of this method, listing `methodArn`, `resource`,
`path` and `requestContext.path` as sharing the same problem. I carried the REST split over by
field name, and that is where it went wrong. In `toAuthorizerEvent` the `resource` field holds a
matched REST resource template such as `/{proxy+}`, so a trailing slash there would be
meaningless. In `buildRequestAuthorizerEventV1` there is no template to hold, so floci sets it to
the request path itself. Before this PR `resource` and `path` were the same string in every case,
and the PR is what split them, for a field the IAM-wildcard rationale never covered.

Fixed in `bc4a886b`. Two fields, not one: `requestContext.resourcePath` is set from the same variable
and is a path copy for the same reason, so leaving it normalized would have recreated the
inconsistency one level down. Neither you nor the #1863 review named that one, so if the
maintainer reads it differently it is a one-line revert on its own.

`methodArn` stays normalized. It is the only field in this shape that is matched against
IAM-style policy resources, which is what the original rationale was actually about.

New test `BuildV2AuthorizerEventTrailingSlashTest` pins all four fields. A REQUEST authorizer's
context never reaches the v2 proxy event, so these fields are not observable end to end, which is
why `buildRequestAuthorizerEventV1` is now package-private, the same reason `buildV2ProxyEvent`
already is. Verified both directions locally on JDK 25:

```
with the fix     Tests run: 9, Failures: 0, Errors: 0    (both trailing-slash classes)
fields reverted  Tests run: 4, Failures: 1, Errors: 0    expected: </thing/> but was: </thing>
whole package    Tests run: 866, Failures: 0, Errors: 0  (71 classes, apigateway + apigatewayv2)
```

`requestWithoutTrailingSlashIsUnchanged` and `rootPathIsUnchanged` pass either way, which is the
point: nothing changes for a request that had no trailing slash to begin with.

pgermosen (approved):
Thanks for the thorough follow-up — the `resource`/`resourcePath` fix is exactly right, and catching the second field on your own was a nice bit of extra care.

Re-verified independently: reverted just the `resource` field and confirmed `BuildV2AuthorizerEventTrailingSlashTest` genuinely fails without the fix, then restored and confirmed it passes. Full `apigateway`/`apigatewayv2` suite: 866 tests, 0 failures, matching your own numbers exactly. Hygiene is clean.

Approving.

Issue #1537, opened 2026-06-24
[BUG] Lambda GetEventSourceMapping does not return DestinationConfig or BisectBatchOnFunctionError


GuillaumeAzkoaga:
## Service

Lambda

## AWS API Action

GetEventSourceMapping

## Expected behavior

When an event source mapping is created with:

- BisectBatchOnFunctionError=true
- DestinationConfig.OnFailure.Destination=queue ARN

GetEventSourceMapping should return those configured fields in the response, consistent with AWS documentation:
https://docs.aws.amazon.com/lambda/latest/api/API_GetEventSourceMapping.html

## Actual behavior

Floci accepts those fields on CreateEventSourceMapping, but omits them from both the create response and the GetEventSourceMapping response.

Observed GetEventSourceMapping response:
```
{
  "UUID": "7dc831d1-7ee6-47a5-91e9-4588b12b466b",
  "BatchSize": 1,
  "EventSourceArn": "arn:aws:dynamodb:us-east-1:000000000000:table/esm-repro-table-1782294798/stream/2026-06-24T09:53:24.372",
  "FunctionArn": "arn:aws:lambda:us-east-1:000000000000:function:esm-repro-fn-1782294798",
  "LastModified": "2026-06-24T11:53:30.729000+02:00",
  "State": "Enabled",
  "FunctionResponseTypes": []
}
``` 

For reference, the create call was made with a payload equivalent to:
```
{
  "FunctionName": "esm-repro-fn-1782294798",
  "EventSourceArn": "arn:aws:dynamodb:us-east-1:000000000000:table/esm-repro-table-1782294798/stream/2026-06-24T09:53:24.372",
  "StartingPosition": "LATEST",
  "BatchSize": 1,
  "MaximumRetryAttempts": 10,
  "BisectBatchOnFunctionError": true,
  "DestinationConfig": {
    "OnFailure": {
      "Destination": "arn:aws:sqs:us-east-1:000000000000:esm-repro-dlq-1782294798"
    }
  }
}
``` 

## Reproduction

```shell
$ErrorActionPreference = 'Stop'
$endpoint = 'http://localhost:4566'
$region = 'us-east-1'
$suffix = [DateTimeOffset]::UtcNow.ToUnixTimeSeconds()

$workdir = Join-Path $env:TEMP ("floci-esm-repro-" + $suffix)
New-Item -ItemType Directory -Path $workdir | Out-Null
Set-Location $workdir

@'
exports.handler = async (event) => {
[...]

PR #2370, opened 2026-08-18
fix(lambda): round-trip StartingPosition on event source mappings

### LINKED ISSUE #2369: [BUG] Lambda event source mappings never return StartingPosition, so Terraform recreates them on every apply
## Service

Lambda

## AWS API Action

CreateEventSourceMapping / GetEventSourceMapping / ListEventSourceMappings

## Expected behavior

`StartingPosition` (and `StartingPositionTimestamp` when the position is `AT_TIMESTAMP`) are part of
the [EventSourceMappingConfiguration](https://docs.aws.amazon.com/lambda/latest/api/API_GetEventSourceMapping.html)
response shape. A mapping created with a starting position should report that position back from the
create response, `GetEventSourceMapping`, and `ListEventSourceMappings`.

## Actual behavior

`CreateEventSourceMapping` accepts both fields and silently discards them. They are never stored on
the mapping, so all three responses omit them entirely.

```console
$ aws --endpoint-url http://localhost:4566 lambda create-event-source-mapping \
    --function-name esm-verify-fn \
    --event-source-arn arn:aws:kinesis:eu-west-1:000000000000:stream/esm-verify-stream \
    --starting-position LATEST --batch-size 5 --query UUID --output text
50ff54cc-ed6b-4218-87e2-767ddf666148

$ aws --endpoint-url http://localhost:4566 lambda get-event-source-mapping \
    --uuid 50ff54cc-ed6b-4218-87e2-767ddf666148
{
    "UUID": "50ff54cc-ed6b-4218-87e2-767ddf666148",
    "BatchSize": 5,
    "EventSourceArn": "arn:aws:kinesis:eu-west-1:000000000000:stream/esm-verify-stream",
    "FunctionArn": "arn:aws:lambda:eu-west-1:000000000000:function:esm-verify-fn",
    "LastModified": "2026-08-18T09:46:13.328000+02:00",
    "State": "Enabled",
    "FunctionResponseTypes": []
}
```

`StartingPosition` is absent, even though it was supplied on create.

### Why it hurts

Both fields are `ForceNew` in Terraform's `aws_lambda_event_source_mapping`. Because refresh reads
the position back as unset, every plan proposes *adding* the value the config already asked for —
and adding a `ForceNew` attribute means destroy-and-recreate. This repeats on every apply,
indefinitely, since the recreated mapping doesn't report it either:

```
  # aws_lambda_event_source_mapping.example must be replaced
-/+ resource "aws_lambda_event_source_mapping" "example" {
      ~ id                = "2066f9da-34a9-4d82-b375-55a8d17d94a2" -> (known after apply)
      + starting_position = "TRIM_HORIZON" # forces replacement
      ~ state             = "Enabled" -> (known after apply)
      ~ uuid              = "2066f9da-34a9-4d82-b375-55a8d17d94a2" -> (known after apply)
        # (8 unchanged attributes hidden)
    }
```

Same defect shape as #1537 (`DestinationConfig`/`BisectBatchOnFunctionError` accepted but not
returned) — in fact that issue's own reproduction sends `StartingPosition: LATEST` and its pasted
response shows the field missing too; it just wasn't the subject at the time.

### A second, separable bug

The pollers ignore the field as well. `KinesisEventSourcePoller` and
`DynamoDbStreamsEventSourcePoller` both hardcode `TRIM_HORIZON` for a shard they have no stored
sequence number for, so a mapping created with `StartingPosition: LATEST` replays the entire
existing stream on its first poll instead of starting at the tip.

Worth flagging that the obvious fix has a trap: the pollers mint a fresh iterator on *every* tick
while no sequence number is stored, so simply passing `LATEST` through would re-snapshot the shard
tip each tick and silently drop anything written between two ticks. `LATEST` needs resolving to a
concrete resume point exactly once. Happy to do this in a follow-up PR if the approach sounds right
— I've left it out of the linked PR to keep the compatibility fix reviewable on its own.

### Related, not addressed

Visible in the same plan diff, several other `EventSourceMappingConfiguration` fields are also
accepted-and-dropped — `MaximumRetryAttempts`, `MaximumBatchingWindowInSeconds`,
`MaximumRecordAgeInSeconds`, `ParallelizationFactor`, `TumblingWindowInSeconds`. Those are not
`ForceNew`, so they cause a perpetual in-place diff rather than a replacement. Mentioning them for
the record rather than folding them in.

## Reproduction

```bash
E=http://localhost:4566

printf 'exports.handler = async (e) => e;\n' > index.js && zip -q function.zip index.js

aws --endpoint-url $E kinesis create-stream --stream-name esm-verify-stream --shard-count 1
STREAM_ARN=$(aws --endpoint-url $E kinesis describe-stream --stream-name esm-verify-stream \
  --query 'StreamDescription.StreamARN' --output text)

aws --endpoint-url $E lambda create-function --function-name esm-verify-fn \
  --runtime nodejs20.x --role arn:aws:iam::000000000000:role/lambda-role \
  --handler index.handler --zip-file fileb://function.zip

UUID=$(aws --endpoint-url $E lambda create-event-source-mapping \
  --function-name esm-verify-fn --event-source-arn "$STREAM_ARN" \
  --starting-position LATEST --batch-size 5 --query UUID --output text)

# Expected: "StartingPosition": "LATEST". Actual: the key is absent.
aws --endpoint-url $E lambda get-event-source-mapping --uuid "$UUID"
aws --endpoint-url $E lambda list-event-source-mappings --function-name esm-verify-fn
```

## Environment

- Floci version / image tag: `floci/floci:nightly` (built 2026-08-18), also reproduces on `main` at
  `75c8cfc`
- Java SDK version (if applicable): N/A (AWS CLI v2)

sheldon-nadarajah-sft:
## Summary

Closes #2369.

`CreateEventSourceMapping` accepted `StartingPosition` and `StartingPositionTimestamp` but never
stored them, so the create response, `GetEventSourceMapping` and `ListEventSourceMappings` all
omitted the fields. Any IaC tool that refreshes a mapping therefore reads the starting position back
as unset — and since both fields are `ForceNew` in Terraform's `aws_lambda_event_source_mapping`,
every plan proposes adding the value it already asked for and so destroys and recreates the mapping,
on every apply, indefinitely.

This is the same defect shape as #1537, which fixed `DestinationConfig` and
`BisectBatchOnFunctionError` the same way.

- **Store both fields** on `EventSourceMapping` (position as-is; timestamp normalised to epoch
  millis, matching how `lastModified` is held).
- **Emit both** from `buildEsmResponse`, omitting them when unset the way AWS omits them for sources
  that have no starting position — so the SQS case doesn't grow a null key.
- **Validate only what was sent.** AWS requires a starting position for stream sources, but Floci has
  always accepted mappings without one, so an absent field stays absent rather than becoming a new
  400 for callers that work today. What *is* sent gets checked: the value must be one of
  `TRIM_HORIZON`/`LATEST`/`AT_TIMESTAMP`, `AT_TIMESTAMP` requires a timestamp, and `AT_TIMESTAMP` is
  rejected on non-Kinesis sources (DynamoDB Streams shard iterators have no timestamp form at all,
  so accepting it there would promise a starting point that can never be honoured).
- **No update path**, because AWS's `UpdateEventSourceMapping` doesn't accept the field — it's
  replace-only, which is exactly why Terraform marks it `ForceNew`.
- **CloudFormation**: `provisionLambdaEventSourceMapping` dropped both fields for the same reason, so
  they're passed through too. This edits the existing method rather than adding a resource type, so
  it shouldn't cut against `CONTRIBUTING.md`'s "don't add cases to this class" — happy to move it if
  you'd rather it went to a `LambdaCfnProvisioner` as part of dismantling the monolith.

`StartingPositionTimestamp` crosses the wire as (possibly fractional) epoch seconds in both
directions, which is the restJson1 default and the convention `LastModified` already uses in this
same response builder.

### Deliberately not in this PR

The pollers ignore the stored position too — both `KinesisEventSourcePoller` and
`DynamoDbStreamsEventSourcePoller` hardcode `TRIM_HORIZON` for a shard with no stored sequence
number, so a `LATEST` mapping replays the whole existing stream on its first poll. That's a genuine
second bug (described in #2369), but it's a behaviour change rather than a wire-compatibility fix,
and the obvious implementation has a trap worth agreeing on first: the pollers mint a fresh iterator
on *every* tick while no sequence number is stored, so passing `LATEST` straight through would
re-snapshot the shard tip each tick and silently drop records written between two ticks. `LATEST`
has to resolve to a concrete resume point exactly once. Glad to follow up with that separately.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior:** both fields accepted on create, then absent from every response that should
carry them.

Verified against a locally built image (this branch's `quarkus-app` overlaid on `floci/floci:latest-jvm`)
using **AWS CLI v2**, with `floci/floci:nightly` as the before-baseline:

Before — `StartingPosition` missing despite being supplied on create:

```console
$ aws --endpoint-url http://localhost:4598 lambda get-event-source-mapping --uuid 50ff54cc-...
{
    "UUID": "50ff54cc-ed6b-4218-87e2-767ddf666148",
    "BatchSize": 5,
    "EventSourceArn": "arn:aws:kinesis:eu-west-1:000000000000:stream/esm-verify-stream",
    "FunctionArn": "arn:aws:lambda:eu-west-1:000000000000:function:esm-verify-fn",
    "LastModified": "2026-08-18T09:46:13.328000+02:00",
    "State": "Enabled",
    "FunctionResponseTypes": []
}
```

After:

```console
$ aws --endpoint-url http://localhost:4599 lambda get-event-source-mapping --uuid 52823a98-...
{
    "UUID": "52823a98-a32f-4957-b4d2-1250062a98f6",
    "StartingPosition": "LATEST",
    "BatchSize": 5,
    "EventSourceArn": "arn:aws:kinesis:eu-west-1:000000000000:stream/esm-verify-stream",
    "FunctionArn": "arn:aws:lambda:eu-west-1:000000000000:function:esm-verify-fn",
    "LastModified": "2026-08-18T09:45:34.118000+02:00",
    "State": "Enabled",
    "FunctionResponseTypes": []
}

$ aws ... lambda list-event-source-mappings --function-name esm-verify-fn \
    --query 'EventSourceMappings[].{UUID:UUID,StartingPosition:StartingPosition}'
[{"UUID": "52823a98-a32f-4957-b4d2-1250062a98f6", "StartingPosition": "LATEST"}]
```

`AT_TIMESTAMP` units round-trip correctly — asked for `2026-08-17T12:34:56Z`, got back the same
instant rendered in the CLI's local zone (UTC+2):

```console
$ aws ... lambda create-event-source-mapping ... \
    --starting-position AT_TIMESTAMP --starting-position-timestamp 2026-08-17T12:34:56Z \
    --query '{StartingPosition:StartingPosition,StartingPositionTimestamp:StartingPositionTimestamp}'
{"StartingPosition": "AT_TIMESTAMP", "StartingPositionTimestamp": "2026-08-17T14:34:56+02:00"}
```

And the validation surfaces through the SDK rather than as a 500:

```console
$ aws ... lambda create-event-source-mapping ... --starting-position AT_TIMESTAMP
An error occurred (InvalidParameterValueException) ...
message: StartingPositionTimestamp is required when StartingPosition is AT_TIMESTAMP
```

An SQS mapping created without a starting position still returns exactly the keys it did before —
`UUID`, `BatchSize`, `EventSourceArn`, `FunctionArn`, `LastModified`, `State`,
`FunctionResponseTypes` — with no `StartingPosition`/`StartingPositionTimestamp` added.

## Checklist

- [x] `./mvnw test` passes locally — full repo suite (10625 tests) has 2 failing classes
      (`DocDbIntegrationTest`, `NeptuneOpenCypherIntegrationTest`, socket timeouts against their
      proxy containers); both untouched by this diff and pass cleanly in isolation, so this reads
      as local resource contention under the full suite's container load, not a regression. The
      681 tests across every Lambda/ESM/CloudFormation/poller suite this PR actually touches are
      all green.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -3915,6 +3915,18 @@ public class CloudFormationResourceProvisioner {
             try { req.put("BatchSize", Integer.parseInt(batchSize)); } catch (NumberFormatException ignored) {}
         }
 
+        String startingPosition = resolveOptional(props, "StartingPosition", engine);
+        if (startingPosition != null) {
+            req.put("StartingPosition", startingPosition);
+        }
+
+        String startingPositionTimestamp = resolveOptional(props, "StartingPositionTimestamp", engine);
+        if (startingPositionTimestamp != null) {
+            try {
+                req.put("StartingPositionTimestamp", Double.parseDouble(startingPositionTimestamp));
+            } catch (NumberFormatException ignored) {}
+        }
+
         var esm = lambdaService.createEventSourceMapping(region, req);
         r.setPhysicalId(esm.getUuid());
         r.getAttributes().put("Id", esm.getUuid());
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -3924,7 +3924,13 @@ public class CloudFormationResourceProvisioner {
         if (startingPositionTimestamp != null) {
             try {
                 req.put("StartingPositionTimestamp", Double.parseDouble(startingPositionTimestamp));
-            } catch (NumberFormatException ignored) {}
+            } catch (NumberFormatException e) {
+                // Not swallowed the way BatchSize above is: dropping this one degrades into the
+                // "StartingPositionTimestamp is required" error from the service, which points at
+                // the wrong problem and hides the value that actually failed to parse.
+                throw new AwsException("ValidationError",
+                        "Value of property StartingPositionTimestamp must be a number.", 400);
+            }
         }
 
         var esm = lambdaService.createEventSourceMapping(region, req);
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -3923,11 +3923,17 @@ public class CloudFormationResourceProvisioner {
         String startingPositionTimestamp = resolveOptional(props, "StartingPositionTimestamp", engine);
         if (startingPositionTimestamp != null) {
             try {
-                req.put("StartingPositionTimestamp", Double.parseDouble(startingPositionTimestamp));
+                double timestamp = Double.parseDouble(startingPositionTimestamp);
+                if (!Double.isFinite(timestamp)) {
+                    throw new NumberFormatException("Non-finite timestamp");
+                }
+                req.put("StartingPositionTimestamp", timestamp);
             } catch (NumberFormatException e) {
                 // Not swallowed the way BatchSize above is: dropping this one degrades into the
                 // "StartingPositionTimestamp is required" error from the service, which points at
-                // the wrong problem and hides the value that actually failed to parse.
+                // the wrong problem and hides the value that actually failed to parse. Double.parseDouble
+                // accepts "NaN"/"Infinity"/"-Infinity" without throwing, so isFinite is checked explicitly
+                // to keep those from silently becoming epoch-zero or long-extremum timestamps downstream.
                 throw new AwsException("ValidationError",
                         "Value of property StartingPositionTimestamp must be a number.", 400);
             }
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -340,6 +340,15 @@ public class LambdaController {
         node.put("BatchSize", esm.getBatchSize());
         node.put("State", esm.getState());
         node.put("LastModified", (double) esm.getLastModified() / 1000.0);
+        // Omitted rather than nulled when unset, so a mapping created without a starting position
+        // reads back the way AWS returns it - and so Terraform sees the configured value on refresh
+        // instead of treating the field as unset and forcing a replacement on every plan.
+        if (esm.getStartingPosition() != null) {
+            node.put("StartingPosition", esm.getStartingPosition());
+        }
+        if (esm.getStartingPositionTimestamp() != null) {
+            node.put("StartingPositionTimestamp", (double) esm.getStartingPositionTimestamp() / 1000.0);
+        }
         ArrayNode responseTypes = node.putArray("FunctionResponseTypes");
 
         if (esm.getBisectBatchOnFunctionError() != null) {
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -775,6 +775,8 @@ public class LambdaService {
 
         EventSourceMapping.DestinationConfig destinationConfig = parseDestinationConfig(request);
 
+        StartingPositionSpec startingPosition = parseStartingPosition(request, eventSourceArn);
+
         String queueUrl = eventSourceArn.contains(":sqs:") ? AwsArnUtils.arnToQueueUrl(eventSourceArn, config.effectiveBaseUrl()) : null;
 
         EventSourceMapping esm = new EventSourceMapping();
@@ -792,6 +794,8 @@ public class LambdaService {
         esm.setFunctionResponseTypes(functionResponseTypes);
         esm.setBisectBatchOnFunctionError(bisectBatchOnFunctionError);
         esm.setDestinationConfig(destinationConfig);
+        esm.setStartingPosition(startingPosition.position());
+        esm.setStartingPositionTimestamp(startingPosition.timestampMillis());
         esm.setLastModified(System.currentTimeMillis());
 
         esmStore.save(esm);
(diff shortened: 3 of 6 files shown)


sheldon-nadarajah-sft:
Thanks for the catch — fixed in the follow-up commit: a malformed `StartingPositionTimestamp` now throws a clear `ValidationError` naming the property, instead of silently degrading into the unrelated "StartingPositionTimestamp is required" message.

sheldon-nadarajah-sft:
Re: the ISO-8601 suggestion — checked against the CFN docs, `StartingPositionTimestamp` is documented as `Number` (Unix time seconds), same convention as the API and as `LastModified` elsewhere in this response, so `Double.parseDouble` is the correct parse here, not a gap. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html

The missing-CFN-test point is fair though — will add a CFN-level case exercising this wiring end to end (the existing `AWS::Lambda::EventSourceMapping` CFN test only covers SQS without a starting position).

sheldon-nadarajah-sft:
Added, pushed: `createStack_lambdaEventSourceMappingWithStartingPosition` in `CloudFormationIntegrationTest` — Kinesis stream + AT_TIMESTAMP, asserts both fields come back correctly from `ListEventSourceMappings` after the stack completes.

pgermosen (approved):
All three review findings already have direct fix commits, verified genuine: reverted just the Double.isFinite check and reproduced the exact regression Greptile's P1 described — the resource creates successfully (CREATE_COMPLETE) instead of failing when StartingPositionTimestamp is "NaN", silently rounding to a bogus value — then restored and confirmed the dedicated test catches it. The malformed-timestamp empty-catch and missing-CFN-test-coverage findings are also both addressed.

Independently verified the "no update path" design decision against the real botocore Lambda model: UpdateEventSourceMapping genuinely has no StartingPosition field at all, confirming AWS really is create-only for this property, matching Terraform's ForceNew behavior this PR fixes for.

Ran CloudFormationIntegrationTest (133/133), EsmIntegrationTest (35/35), and LambdaServiceTest (50/50) locally — 0 failures. Clean merge against current main, docs-check passes, no CODEOWNERS-protected paths.

Issue #2309, opened 2026-08-14
[FEAT] efs: Elastic File System (file systems, mount targets, access points)


pgermosen:
## Service

EFS (Elastic File System) — new service

## API Action / Feature

File systems, mount targets and access points

## AWS Documentation

- https://docs.aws.amazon.com/efs/latest/ug/api-reference.html
- https://docs.aws.amazon.com/efs/latest/ug/API_CreateFileSystem.html
- https://docs.aws.amazon.com/efs/latest/ug/API_CreateMountTarget.html
- https://docs.aws.amazon.com/efs/latest/ug/API_CreateAccessPoint.html

## Why is this needed?

There is no `services/efs` package, so every EFS action fails.

EFS is the standard way to give ECS and Lambda persistent shared storage, and Floci already has
the consuming half: ECS task definitions carry an `EfsVolumeConfiguration` field with nothing to
point it at. A task definition that mounts an EFS volume therefore cannot be created.

`aws_efs_file_system`, `aws_efs_mount_target` and `aws_efs_access_point` are all unsupported.

## Scope

This is a tracking issue for a new service package. It is **not small**. Suggested split:

**1. File systems**
`CreateFileSystem`, `DescribeFileSystems`, `UpdateFileSystem`, `DeleteFileSystem`, plus
`CreateTags`/`DescribeTags`/`DeleteTags` (EFS has its own tag actions in addition to the newer
`TagResource` family — the provider uses `ListTagsForResource` on read).
Backs `aws_efs_file_system`.

**2. Mount targets**
`CreateMountTarget`, `DescribeMountTargets`, `DeleteMountTarget`,
`ModifyMountTargetSecurityGroups`, `DescribeMountTargetSecurityGroups`.
Backs `aws_efs_mount_target` — this is what ECS actually needs to resolve.

**3. Access points and policies**
`CreateAccessPoint`, `DescribeAccessPoints`, `DeleteAccessPoint`, `PutFileSystemPolicy`,
`DescribeFileSystemPolicy`, `PutBackupPolicy`, `PutLifecycleConfiguration`.
Backs `aws_efs_access_point` and `aws_efs_file_system_policy`.

**Protocol.** EFS is REST JSON (`@Path("/2015-02-01/file-systems")` and similar), so it follows
the JAX-RS pattern used by Lambda and API Gateway rather than the Query controllers.

**Metadata only.** No NFS server needs to exist — the value is in the resources being creatable,
[...]

PR #2371, opened 2026-08-18
feat(efs): add Elastic File System (file systems, mount targets, access points)

namanONcode:
This pull request introduces support for the Amazon Elastic File System (EFS) service in the emulator. It adds configuration options, a new controller with REST endpoints for EFS operations, and integrates EFS into the service catalog. The changes include both the API surface and error handling for EFS.

**EFS Service Integration:**

* Added a new `EfsController` with REST endpoints to handle EFS operations such as file system management, tags, mount targets, access points, policies, backup policies, and lifecycle configurations. (`src/main/java/io/github/hectorvent/floci/services/efs/EfsController.java`)
* Introduced `EfsException` for structured error handling specific to EFS-related errors. (`src/main/java/io/github/hectorvent/floci/services/efs/EfsException.java`)

**Configuration Updates:**

* Added `EfsServiceConfig` and `EfsStorageConfig` interfaces to the emulator configuration, allowing EFS service and storage settings to be specified (including enablement and flush interval). (`src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java`) [[1]](diffhunk://#diff-27a905c8526449c5e9a7b7553e391bc16b72b42ba0b4eac1ca9873fcd1d5215dR286) [[2]](diffhunk://#diff-27a905c8526449c5e9a7b7553e391bc16b72b42ba0b4eac1ca9873fcd1d5215dR516-R522) [[3]](diffhunk://#diff-27a905c8526449c5e9a7b7553e391bc16b72b42ba0b4eac1ca9873fcd1d5215dR630) [[4]](diffhunk://#diff-27a905c8526449c5e9a7b7553e391bc16b72b42ba0b4eac1ca9873fcd1d5215dR669-R673)

**Service Catalog Integration:**

* Registered EFS in the `ResolvedServiceCatalog`, enabling the emulator to recognize and route requests to the new EFS controller and service. (`src/main/java/io/github/hectorvent/floci/core/common/ResolvedServiceCatalog.java`) [[1]](diffhunk://#diff-9d524612c303fceb6d83b7751c45c31a5750050dfc6697045c148968587d5456R35) [[2]](diffhunk://#diff-9d524612c303fceb6d83b7751c45c31a5750050dfc6697045c148968587d5456L471-R478)## Summary

https://github.com/floci-io/floci/issues/2309

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Checklist


pgermosen:
Hi @namanONcode, thanks for adding EFS support — this is a substantial addition. A few things worth addressing before this is ready:

**PR title / Conventional Commits.** `[FEAT] efs: ...` doesn't match this repo's Conventional Commits format (it should be `feat(efs): ...`). The "Commits follow Conventional Commits" check is green right now only because it validates the individual commit messages in the branch, not the PR title — but this repo squash-merges, so the title becomes the actual permanent commit message on merge. Could you update the title to something like `feat(efs): add Elastic File System (file systems, mount targets, access points)`?

**Code findings — three of the six Greptile raised are still open**, verified by reading the current code (the other three — region-scoping, idempotency tokens, and the timestamp-units bug — are already fixed correctly in `308d41d9`, including matching the real EFS API's epoch-seconds contract):

- `createAccessPoint` never validates that the referenced `FileSystemId` actually exists before persisting the access point. A client can point an access point at a nonexistent, deleted, or cross-region file system and get a 200 back with an orphaned resource — real AWS rejects this.
- `deleteFileSystem` deletes the file system entry without checking for or cleaning up dependent mount targets, access points, or policy records. Real AWS rejects `DeleteFileSystem` while mount targets exist; here, deleting a file system with live children leaves them listable and pointing at nothing.
- `deleteMountTarget`'s parent-file-system-count update has a genuinely empty `catch (Exception e) { }` with no logging at all — AGENTS.md asks that intentionally tolerated exceptions be logged with enough context to diagnose later, and right now a real storage failure here (not just the documented "fs already deleted" case) would silently produce a wrong `numberOfMountTargets` with no trace.

**A couple of smaller things:**
- No `docs/services/efs.md` yet — every other service in the repo has one. `docs-check` doesn't hard-require it, but it's the convention.
- The checklist boxes (`./mvnw test passes`, integration test added, Conventional Commits) are all left unchecked even though CI is actually green — worth ticking those once the above is addressed, so the PR description reflects reality.

Happy to take another look once these are in.

namanONcode:
@pgermosen all mentioned issues resolved

pgermosen:
Re-checked this against the latest commits — good progress. All three code findings from the earlier review are genuinely fixed:

- `createAccessPoint` now validates the referenced `FileSystemId` exists before proceeding.
- `deleteFileSystem` now rejects deletion while mount targets exist (matching real AWS) and cleans up access points, policies, backup policies, and lifecycle configs.
- `deleteMountTarget`'s parent-count update no longer has a silent empty catch — both the "already deleted" and genuine-failure paths are now logged with context.

The title is fixed (`feat(efs): ...`), and `docs/services/efs.md` is now present.

Two things left:

1. **Merge conflict with `main`**: `src/main/resources/application.yml` and `src/test/resources/application.yml` both conflict — looks like pure config-list churn (other services landed at the same insertion point since this branch was last rebased), not a real code issue, but it'll need a rebase before this can merge.
2. **Minor**: the PR description's checklist boxes (`./mvnw test` passes, integration test added, Conventional Commits) are still all unchecked even though they're now true — worth ticking those so the description matches reality.

Nothing else outstanding that I could find — I went through the full back-and-forth with Greptile's findings across the later commits (idempotency handling, the concurrency/locking fixes, the backup-retry snapshot persistence) and they all look correctly resolved in the current code. Once the rebase is done this looks ready to me.

namanONcode:
@pgermosen @hectorvent  All conflicts resolved and pr description fixed , also ci is green

pgermosen:
First, a correction I owe you. On Aug 21 I wrote *"nothing else outstanding that I could find… once the
rebase is done this looks ready to me."* That was wrong, and it was wrong for a reason worth naming: my
earlier passes adjudicated Greptile's findings and re-read the code, but I never checked the wire
protocol against the EFS service model. I've now done that, and driven a running instance with the
paths and values a real AWS client sends. Several operations can't be reached or used by an SDK at all.
That's on me for signalling "ready" off an incomplete check — but it does mean this needs another round
rather than a merge.

The shape of the service is right: three-layer controller/service/storage, region-scoped keys, the
catalog descriptor with `elasticfilesystem` as the credential scope, per-service storage config, and
region isolation actually works (I confirmed a file system created in `us-east-1` is invisible from
`eu-west-1`). The problems below are all at the API boundary.

## Blocking

### 1. `ThroughputMode` can't accept the value AWS clients send

```java
public enum ThroughputMode { BURSTING, PROVISIONED, ELASTIC }
```

The model's enum values are lowercase — `bursting`, `provisioned`, `elastic` — and that's what goes on
the wire. Jackson matches enum constants by name, so:

| request | result |
|---|---|
| `{"CreationToken":"a","ThroughputMode":"elastic"}` | **400** `{"objectName":"CreateFileSystemRequest","attributeName":"ThroughputMode",...}` |
| `{"CreationToken":"b","ThroughputMode":"ELASTIC"}` | 201 |

So `aws efs create-file-system --throughput-mode elastic` fails, and the only value that works is one no
AWS client will ever send. `UpdateFileSystem` is hit the same way — that's the only request field it
really has, so the operation is effectively unusable. The 400 isn't even an AWS-shaped error; it's
Jackson's deserialization failure leaking through. Worth checking whether the Terraform provider sends
`throughput_mode` by default, because if it does, no `aws_efs_file_system` resource can be created at all.

`PerformanceMode` (`generalPurpose`/`maxIO`) and `LifeCycleState` both got this right, so it's a slip on
one enum rather than a misunderstanding. Either match the wire values as constant names, or add
`@JsonProperty` on each constant.

### 2. The file-system policy endpoints are on the wrong path

The controller registers `/2015-02-01/file-systems/{FileSystemId}/policies`. The model's URI is
`/2015-02-01/file-systems/{FileSystemId}/policy` — singular — for all three of `PutFileSystemPolicy`,
`DescribeFileSystemPolicy` and `DeleteFileSystemPolicy`. Against a running instance:

```
PUT  /2015-02-01/file-systems/fs-…/policy    → 404  <html><body><h1>Resource not found</h1></body></html>
PUT  /2015-02-01/file-systems/fs-…/policies  → 200  {"FileSystemId":"fs-…","Policy":"{}"}
```

So the entire file-system-policy API is unreachable from any SDK or the CLI, and what a client gets back
is an HTML 404 it can't parse. `manageFileSystemPolicy` in `EfsIntegrationTest` uses `/policies` too, so
the test agrees with the implementation and can never catch the divergence — which is the general
pattern in point 6 below.

### 3. `CreateFileSystem`'s idempotency contract is inverted

The model is explicit:

> *"If a file system does not currently exist that is owned by the caller's account with the specified
> creation token, this operation … creates a new, empty file system … Otherwise, this operation returns a
> `FileSystemAlreadyExists` error with the ID of the existing file system."*

That's why `FileSystemAlreadyExists` carries a required `FileSystemId` member — the client recovers the
ID from the error. There is no parameter comparison: reusing the token is *always* a 409.

The implementation does the opposite. On a token reuse it compares performance mode, throughput mode,
encryption, KMS key, provisioned throughput, AZ, backup setting and tags, then either returns the
existing file system with **201**, or throws:

```
POST /file-systems, token reused, same params      → 201 (existing description)
POST /file-systems, token reused, different params → 400 {"__type":"IdempotentParameterMismatch",…}
```

`IdempotentParameterMismatch` is not an EFS error code — it isn't among the service's exception shapes,
and it's not in `CreateFileSystem`'s error list (`BadRequest`, `InternalServerError`,
`FileSystemAlreadyExists`, `FileSystemLimitExceeded`, `InsufficientThroughputCapacity`,
`ThroughputLimitExceeded`, `UnsupportedAvailabilityZone`). It's an EC2-family code.

I realise this is the outcome of a long back-and-forth with Greptile across roughly ten commits, which
makes it the most frustrating one to raise — but the parameter-matching machinery is implementing a
contract EFS doesn't have, and deleting it is a simplification: any reuse of a live creation token
returns `FileSystemAlreadyExists` (409) with the existing `FileSystemId`. That's fewer lines than what's
there now. It's also a good illustration of why a bot's suggestion needs a source check before it's acted
on — the parameter comparison sounds plausible and is what several other AWS services do.

`CreateAccessPoint` has the same shape of question. Its `ClientToken` is marked `idempotencyToken: true`
and its error list includes `AccessPointAlreadyExists` (409, with a required `AccessPointId`). I'm less
certain of the exact behaviour there than for `CreateFileSystem`, so treat it as "please check the
source" rather than a definite finding.

### 4. The describe operations ignore their filters

`DescribeFileSystems` takes `MaxItems`, `Marker`, `CreationToken` and `FileSystemId` as querystring
parameters. The controller method takes no query parameters at all:

```
3 file systems exist
GET /2015-02-01/file-systems?FileSystemId=fs-0b7688…  → 3 file systems returned
```

`CreationToken` is ignored the same way. This is the read path every SDK and the Terraform provider use
after a create, and it silently returns the wrong set as soon as there's more than one file system.

`DescribeMountTargets` has the matching gap: `FileSystemId` is wired up, but `MountTargetId` and
`AccessPointId` aren't (AWS requires exactly one of the three, and rejects a call with none — here a
bare `GET /mount-targets` returns 200 with an empty list).

Tellingly, `DescribeFileSystemsRequest.java` and `DescribeMountTargetsRequest.java` are both in the PR
and both unreferenced — the shapes that would carry these parameters exist, they're just not wired in.

### 5. The docs page claims the tagging API that isn't implemented

`docs/services/efs.md` lists **TagResource, UntagResource, ListTagsForResource** under Tags. All three
404:

```
POST /2015-02-01/resource-tags/fs-…  → 404 (HTML)
GET  /2015-02-01/resource-tags/fs-…  → 404 (HTML)
```

What *is* implemented is the older `CreateTags`/`DescribeTags`/`DeleteTags` trio, which the page doesn't
mention. So the table is exactly inverted from reality. `UntagResourceRequest.java` is in the PR but
unreferenced, so the intent was clearly there. Worth noting the modern trio is what the Terraform
provider uses for EFS tag updates, so this isn't only a docs fix.

## Why CI didn't catch any of this

Worth being explicit, because fourteen green checks is a misleading signal here:

- **No SDK-level test.** Nothing in `compatibility-tests/` touches EFS — the only mention is an EFS ARN
  string inside `LambdaTest`. Every one of the findings above needs a request built by a real client to
  surface, and no test in this PR builds one. `EfsIntegrationTest` drives Floci's own paths with
  hand-written JSON, so it can only confirm the implementation agrees with itself.
- **"Action tables in sync" passes vacuously.** EFS isn't registered in `tools/docs/services.yaml`, and
  CONTRIBUTING.md says the sentinel fires when *"a new switch handler is neither registered … nor listed
  under `deferred_handlers`"* — a REST controller isn't covered. I ran `make docs-check` locally on this
  branch: exit 0, no diff, with EFS entirely unregistered. `docs/services/efs.md` is also hand-written
  without the `<!-- floci:actions:start -->` markers, so nothing regenerates it against the controller —
  which is why finding 5 could exist at all. Registering it (`{ path: …/EfsController.java, mode: rest }`,
  the same form FIS uses) makes the generator own that table.

## Smaller things

- **Account ID is hardcoded.** `"000000000000"` is baked into every ARN and `OwnerId`. The controller
  already injects `RegionResolver`, which has both `getAccountId()` and `buildArn(service, region,
  resource)`; neither is passed to the service. Anyone running with a configured account ID gets wrong
  ARNs.
- **`UpdateFileSystem` returns 200**, the model says **202**.
- **Mount targets are entirely synthetic and never conflict.** Every one gets `10.0.0.10`,
  `vpc-12345678`, `us-east-1a` / `use1-az1` regardless of region or subnet, and two mount targets in the
  same subnet both succeed — real EFS returns `MountTargetConflict`, one mount target per AZ. Fixed
  placeholders are a reasonable emulator choice; the hardcoded `us-east-1` AZ in a `eu-west-1` file
  system is the part that will read as a bug.
- **`FileSystemDescription.Name` is never populated.** AWS derives it from the `Name` tag ("If the file
  system has a `Name` tag, Amazon EFS returns the value in this field"). After `CreateTags` with
  `Name=n1`, `DescribeFileSystems` still returns `"Name": null`.
- **No pagination.** `MaxItems`/`Marker` aren't read anywhere; responses do carry null `Marker`/
  `NextMarker`, so the shape is right and only the behaviour is missing.
- **17 of the 51 model classes are unreferenced** (~500 lines): the two describe-request shapes and
  `UntagResourceRequest` from above, plus `CreateFileSystemResponse`, `UpdateFileSystemResponse`, the six
  delete/describe request shapes, `BackupPolicyStatus`, `TransitionToIA`, `TransitionToArchive`,
  `TransitionToPrimaryStorageClass`. Interesting detail: the enums that are *unused* are the ones whose
  fields are plain `String`s — and those round-trip correctly. The two enums that are actually wired into
  models are `PerformanceMode` and `ThroughputMode`, and that's exactly where finding 1 lives.
- **Error body member names.** EFS models its errors as `{"ErrorCode": …, "Message": …}`; the shared
  mapper emits `{"__type": …, "message": …}`. SDK error *mapping* still works because
  `x-amzn-errortype` is set (I verified the header is present), so this is low priority and arguably
  house convention — but `FileSystemAlreadyExists.fileSystemId()` would come back null, and carrying that
  ID is the entire purpose of that error shape. Relevant if you take finding 3.
- **`describeFileSystems` filters twice** — once on the `region::` key prefix, then again on
  `getFileSystemArn().contains(":" + region + ":")`. The second is redundant given the first, and NPEs on
  a record with a null ARN.
- **Registration gaps beyond `services.yaml`:** EFS is absent from `docs/services/index.md`,
  `docs/index.md`, `README.md`, and the environment-variable / `application-yml` reference pages. The FIS
  PR (#2435) that landed this week updated all of them and is a good template.

## Local build

Full suite on the PR head (which already contains current `main`): **10,959 tests across 841 classes,
0 failures, 0 errors, 9 skipped.** `EfsIntegrationTest` contributes 10 of those. So nothing here is a
build problem — the gaps are all in what the tests assert, not whether they pass.

## What I verified as correct

Not everything needs changing, and a few things are done well:

- Region scoping is real, not decorative — cross-region isolation confirmed against a running instance.
- `CreationTime` is epoch seconds, which is right for a rest-json timestamp; that was a genuine fix
  earlier in the review and it holds.
- Backup policy, lifecycle configuration and the `CreateTags`/`DescribeTags`/`DeleteTags` trio all
  round-trip correctly.
- `DeleteFileSystem` rejects with `FileSystemInUse` (409) while mount targets exist and cascades access
  points, policies, backup policies and lifecycle configs — matching AWS.
- ID formats (`fs-`/`fsmt-`/`fsap-` + 17 hex) and the `FileSystemDescription` required-member set are
  correct.
- `LifeCycleState` and `PerformanceMode` match the model exactly.
- The catalog descriptor, storage config and `@RegisterForReflection` coverage all follow the repo's
  conventions.

namanONcode:
@hectorvent @pgermosen all mention issues resolved and suggestion implemented and ci is green

pgermosen:
Re-checked against `03fc8fd0`, merged onto current `main`, built, and driven with the CLI again. Most of the last round is genuinely closed, and I want to give that its due before the two items that are still open.

## Verified fixed

Each of these I confirmed against a running instance, not by reading:

| finding | evidence |
|---|---|
| **1** `ThroughputMode` rejected AWS wire values | `--throughput-mode elastic` → 201 (was a 400 Jackson leak) |
| **2** policy endpoints on the wrong path | `put-file-system-policy` / `describe-file-system-policy` both work; controller is on `/policy` |
| **4** describe operations ignored their filters | 3 file systems exist; `--file-system-id` → 1, `--creation-token` → 1. `DescribeMountTargets` now rejects a call with none of the three IDs and filters on all three |
| **5** docs claimed a tagging API that 404'd | `tag-resource` / `list-tags-for-resource` round-trip; the trio is implemented |

And the registration gap behind finding 5 is properly closed rather than papered over: `tools/docs/services.yaml` now carries `{ path: …/EfsController.java, mode: rest }`, `docs/services/efs.md` has the `floci:actions` markers so the table is generated, and `docs/services/index.md` has its row. `regen_action_docs.py --strict` exits 0. That is the fix that stops this class of drift recurring.

Also closed from the smaller list: account ID now comes from `RegionResolver.getAccountId()`/`buildArn` instead of a baked-in `000000000000`; `UpdateFileSystem` returns 202; a second mount target in the same subnet is now rejected with `MountTargetConflict`; the mount-target AZ is derived from the request region instead of hardcoded `us-east-1`; `Name` is populated from the `Name` tag; the redundant second ARN filter in `describeFileSystems` is gone; and unreferenced model classes are down from 17 to 2.

EFS tests: 10/10 on the merge. Merge with `main` is clean.

---

## (blocking) `ThroughputMode` is fixed on the way in and broken on the way out

The `@JsonProperty` annotations fixed deserialization. Serialization still goes through `.name()`:

```java
// EfsService.java:137
fs.setThroughputMode(request.getThroughputMode() != null ? request.getThroughputMode().name() : "bursting");
// EfsService.java:201 (UpdateFileSystem)
fs.setThroughputMode(request.getThroughputMode().name());
```

`.name()` is the Java constant name, so the annotation never runs:

```
POST /2015-02-01/file-systems  {"CreationToken":"t-raw2","ThroughputMode":"elastic"}
→ "ThroughputMode": "ELASTIC"
```

The model's enum is `['bursting', 'provisioned', 'elastic']`. `ELASTIC` is not a member, so an SDK deserializes it as `UNKNOWN_TO_SDK_VERSION`, and the Terraform provider would see `throughput_mode = "elastic"` in config against `"ELASTIC"` in state — a diff that never converges.

The tell is that the *default* branch of that same ternary returns the literal `"bursting"`, lowercase. So the casing of this field depends on whether the caller passed a value:

```
create --creation-token t-idem                        → "ThroughputMode": "bursting"
create --creation-token t-raw2 --throughput-mode elastic → "ThroughputMode": "ELASTIC"
```

`PerformanceMode` escapes this only because its constants are literally named `generalPurpose` and `maxIO`, so `.name()` happens to produce the wire value. Simplest fix is to make `ThroughputMode`'s constants match the wire values the same way and drop the annotations, or keep the annotations and stop calling `.name()` — but not the current mix of both.

Nothing in `EfsIntegrationTest` asserts `ThroughputMode` at all, which is why this passed 10/10.

## (blocking) The same `.name()` slip makes idempotency reject identical requests

```java
// EfsService.java:96-98
String reqTpMode = request.getThroughputMode() != null ? request.getThroughputMode().name() : "bursting";
String extTpMode = original.getThroughputMode() != null ? original.getThroughputMode().name() : "bursting";
if (!Objects.equals(reqTpMode, extTpMode)) match = false;
```

`"BURSTING"` from one side, `"bursting"` from the other. Passing the **same** value explicitly is now a conflict:

```
$ aws efs create-file-system --creation-token t-idem                            # defaults to bursting
fs-33601063345c46b9b   bursting

$ aws efs create-file-system --creation-token t-idem --throughput-mode bursting
An error occurred (IdempotentParameterMismatch) when calling the CreateFileSystem operation:
A resource with this token already exists with different parameters.
```

A retry that spells out the default it already got is the single most likely way a client re-sends a creation token. The `originalFileSystemRequestStore` added in `03fc8fd0` makes the comparison operate on the right data, but the comparison itself is still wrong on this field.

## (blocking, unchanged) The idempotency contract is still inverted

This is finding 3 from last time, and I understand why it survived — the parameter-matching machinery is what the bot review asked for. But the model is unambiguous, and the round trip is still:

```
POST /2015-02-01/file-systems  {"CreationToken":"t-idem"}   (token already in use, identical params)
→ HTTP 201, existing file system description
```

AWS returns **409 `FileSystemAlreadyExists`**, always, with no parameter comparison. `FileSystemAlreadyExists` lists `ErrorCode` and `FileSystemId` as its *required* members precisely so the client recovers the ID from the error. `IdempotentParameterMismatch` is not in `CreateFileSystem`'s error list (`BadRequest`, `InternalServerError`, `FileSystemAlreadyExists`, `FileSystemLimitExceeded`, `InsufficientThroughputCapacity`, `ThroughputLimitExceeded`, `UnsupportedAvailabilityZone`) and is not an EFS shape at all — it is EC2-family.

The strongest argument that this is a genuine miss rather than a judgement call is in your own code: `EfsException.fileSystemAlreadyExists(token, fileSystemId)` and `EfsException.accessPointAlreadyExists(clientToken, accessPointId)` both exist, both build the correct 409 with the ID in `extendedData` — and neither is ever called. The right contract is already written; only the call site is missing.

Deleting the comparison at EfsService.java:86-121 (and the parallel block at 444-496 for access points) in favour of an unconditional `throw EfsException.fileSystemAlreadyExists(...)` removes both blocking items above and is a net reduction in lines.

---

## Notes

- **Still no SDK-level coverage.** `compatibility-tests/` has no EFS test, and `EfsIntegrationTest` is still 10 tests hand-writing JSON against Floci's own paths. Every finding in this round and the last needed a real client to surface. Given how much behaviour landed since the last review — the tagging trio, describe filters, mount-target conflict, the idempotency store — a compat test is the thing that would have caught the `ELASTIC` casing before I did.
- **Pagination** — `MaxItems`/`Marker` are still not read anywhere. Unchanged from last round, still not blocking.
- **AZ ID format.** `EfsService.java:353` builds `region.replace("-", "") + "-az1"`, giving `useast1-az1`. AWS uses a short region code: `use1-az1`, `usw2-az1`, `euw1-az1`. Deriving it from the region is the important half and that is done; the abbreviation scheme is cosmetic.
- **Error body member names** — unchanged, and as I said last time this is arguably house convention. It only starts to matter once `FileSystemAlreadyExists` is actually thrown, since `fileSystemId()` would come back null from the `__type`/`message` shape.

The three blocking items are all the same underlying decision — what `CreateFileSystem` does with a reused token, and how `ThroughputMode` crosses the wire — and fixing them together is less code than what is there now.

namanONcode:
@pgermosen @hectorvent all remaing issues resolved and ci is green

pgermosen:
Re-checked at `2693fea3`. `origin/main` is already an ancestor of the head, so this is genuinely current — no merge needed. Built, ran the suite, and drove a live instance with the CLI and botocore.

**All three blocking items are closed.** The fix is `fa802ca4`, and it is a net **−86 lines**.

## Verified fixed

| finding | evidence |
|---|---|
| **1** `ThroughputMode` serialized as the Java constant name | `--throughput-mode elastic` → `"ThroughputMode": "elastic"` (was `"ELASTIC"`); `provisioned` → `"provisioned"`; default → `"bursting"` |
| **2** the same slip made an identical retry a false `IdempotentParameterMismatch` | gone with the comparison block |
| **3** idempotency contract inverted | duplicate `CreationToken` → **409 `FileSystemAlreadyExists`**; duplicate `ClientToken` → **409 `AccessPointAlreadyExists`** |

You took the option that matches the rest of the package — `ThroughputMode`'s constants are now `bursting`/`provisioned`/`elastic` like `PerformanceMode` and `LifeCycleState`, and the annotations are gone, so there is no longer a mix of two mechanisms. `UpdateFileSystem` (`EfsService.java:168`) goes through the same `.name()` and is correct now too:

```
$ aws efs update-file-system --file-system-id fs-… --throughput-mode elastic
"ThroughputMode": "elastic"
```

Bad values still fail closed — `"BURSTING"` and `"nonsense"` both give 400, not a 500.

On the idempotency change, the part I want to record is that the **recovery path actually works end to end**, which is the whole point of the error. Parsed by botocore, not read off the wire:

```python
except ClientError as e:
    e.response["Error"]["Code"]  → "FileSystemAlreadyExists"
    e.response["FileSystemId"]   → "fs-b0c1facaa2024de29"
```

That is exactly what `CreateFileSystem`'s documentation promises — *"As long as you use the same creation token, if the initial call had succeeded in creating a file system, the client can learn of its existence from the `FileSystemAlreadyExists` error."* It also closes the error-body note I left last round: `FileSystemId` reaches the SDK. Token reuse after `DeleteFileSystem` correctly creates a fresh file system, so the token is released with the resource.

The access-point half is the one piece that is inference rather than quotation, and I think it lands the right way: `CreateAccessPoint`'s own docs say nothing about idempotency, but `ClientToken` carries `"idempotencyToken": true` with the same *"uses to ensure idempotent creation"* wording, and `AccessPointAlreadyExists` is modeled as a 409 with a **required** `AccessPointId` and the documentation *"Returned if the access point that you are trying to create already exists, with the creation token you provided in the request."* Structurally identical to the file-system case.

Nothing downstream depended on the old behaviour — `EfsService` has no consumer outside its own package (no `AWS::EFS::*` in the CloudFormation provisioner), so the contract change is self-contained.

EFS tests 10/10. `regen_action_docs.py --strict` and `check_service_matrix.py` both exit 0. All 18 CI checks green.

---

## The fix left its scaffolding behind

The comparison logic is gone but everything that fed it is still there, and three of those are **persistent stores that are written on every create and never read**:

| | written | deleted | read |
|---|---|---|---|
| `originalFileSystemRequestStore` → `efs-original-fs-requests.json` | `:115` | `:202` | never |
| `originalAccessPointRequestStore` → `efs-original-ap-requests.json` | `:435` | `:455` | never |
| `originalBackupStore` → `efs-original-backups.json` | `:114` | `:201` | never |

`originalBackupStore` predates this commit; the other two were added in `03fc8fd0` specifically to feed the comparison that `fa802ca4` deleted. Also now dead:

- `EfsException.idempotentParameterMismatch()` (`EfsException.java:52`) — no callers in the repo.
- `import java.util.Collections;` and `import java.util.Objects;` (`EfsService.java:14-15`) — no remaining uses in the file.

Dropping all of it removes three JSON files from the emulator's data directory and takes the field/constructor/`clear()` triples with them.

## Nothing asserts any of the three fixes

`EfsIntegrationTest` is the same 10 tests it was two rounds ago. Grepping `src/test/java/**/efs/` for `ThroughputMode`, `AlreadyExists`, or `IdempotentParameter` returns nothing, and there is still no `compatibility-tests/sdk-test-java` EFS test.

The `ELASTIC` casing survived two review rounds for exactly this reason — the suite could not see it. Three asserts would lock the whole round in:

1. create with `ThroughputMode: elastic`, assert the response says `elastic`
2. create twice with one `CreationToken`, assert 409 and that the body carries `FileSystemId`
3. same for `ClientToken` / `AccessPointAlreadyExists` / `AccessPointId`

Given that every finding across three rounds came from a real client rather than from the suite, this is the item I would most like to see before merge.

## Carried over, all minor

- `AccessPointDescription.Name` is never populated. AWS: *"The name of the access point. This is the value of the `Name` tag."* `FileSystem.Name` is derived from the tag; the access-point one is not — a `create-access-point --tags Key=Name,Value=myap` comes back with the tag but no `Name`.
- `DescribeFileSystems` with an unknown `FileSystemId` or `CreationToken` returns `{"FileSystems": []}`. AWS returns 404 `FileSystemNotFound` — stated in `DeleteFileSystem`'s docs: *"If you pass file system ID or creation token for the deleted file system, the `DescribeFileSystems` returns a 404 `FileSystemNotFound` error."*
- Pagination — `Marker`/`MaxItems` are on the models and still unread by the service.
- AZ-ID shape — `:112` gives `us-west-2a-id` and `:320` gives `uswest2-az1`; AWS uses `usw2-az1`. Cosmetic, and deriving it from the region was the important half.
- `update-file-system --throughput-mode elastic` leaves `ProvisionedThroughputInMibps: 64.0` in the returned description; AWS does not allow that pairing.
- `ThroughputMode.java` has no trailing newline.

namanONcode:
@pgermosen all mentioned  major and minors issues resolved

namanONcode:
@hectorvent @pgermosen `Failures: 
[ERROR]   BatchDockerRunnerDockerIntegrationTest.dockerRunnerInjectsContainerReachableEndpoint:64 Could not pull image: toomanyrequests: Rate exceeded ==> expected: <0> but was: <1>
`

FAILED ON PULLING IMAGE RATE LIMIT EXCEED , ci stops before complete

hectorvent:
The three asserts landed exactly as pgermosen named them, and as a real AWS SDK compat test rather than raw HTTP, which is even better: EfsTest pins elastic round tripping, duplicate CreationToken giving FileSystemAlreadyExists with the file system id, and duplicate ClientToken giving AccessPointAlreadyExists with the access point id. It ran green in the compat job on this head. The scaffolding cleanup and the five smaller items from his list all check out too.

CI is now fully green on this head. The earlier red was a Docker Hub rate limit in a Batch test; the rerun cleared it and nothing was needed from you.

**(follow-up, separate PR)** A `mkdocs.yml` nav entry and README row for EFS, plus a `floci.storage.services.efs` block like `guardduty` has.

No blockers from my side. Thanks for the persistence across these rounds.

hectorvent (approved):
Approved, see comment above.

PR #2377, opened 2026-08-18
fix(apigatewayv2): route requests to API-owning account

aryasaatvik:
## Summary

Route HTTP API v2 data-plane requests through the API owner's account and region when the incoming credential or default context points elsewhere. Reject cross-account API ID collisions so ownership remains unambiguous.

## Flow

```text
execute-api request
  -> locate the API across account-aware storage
  -> switch RequestContext to the owning account and region
  -> resolve the route, integration, and Lambda in the owner's namespace
```

- Preserve the existing signed-region and cross-region fallback when no global owner is found.
- Apply the same owner resolution to direct execute-api paths and execute-api hostnames.

## Validation

- Focused API Gateway, account-routing, and JWT suite: 138 tests passed.
- `./mvnw test`: 10,795 tests passed, 9 skipped.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteApiHostFilter.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteApiHostFilter.java
@@ -3,6 +3,7 @@ package io.github.hectorvent.floci.services.apigateway;
 import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.core.common.RegionResolver;
+import io.github.hectorvent.floci.core.common.RequestContext;
 import io.github.hectorvent.floci.services.apigatewayv2.ApiGatewayV2Service;
 import jakarta.annotation.Priority;
 import jakarta.enterprise.context.ApplicationScoped;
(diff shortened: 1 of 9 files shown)


pgermosen (approved):
Thanks for this — solid fix, and I appreciated being able to verify the core premise rather than take it on faith. Checked AWS's own IAM authorization doc directly: IAM/SigV4 is explicitly opt-in per route, not required by default, which means for the common unauthenticated case there's no caller credential to route by at all — the API ID in the invoke URL has to be what determines the owning account, exactly what this PR implements.

Traced the mechanism end to end: when findApiOwner locates the true owner, it correctly propagates the corrected account/region into RequestContext before dispatch, so downstream Lambda/integration resolution lands in the right namespace too, not just the initial API lookup. The fallback to the existing signed-region/cross-region search is preserved verbatim when no owner is found.

ApiGatewayV2UnsignedAccountRoutingTest is a strong regression guard — it creates APIs under two different accounts and confirms a request signed as account B resolves into account A's function when targeting account A's API ID, plus confirms cross-account ID collisions are rejected.

Ran the full targeted suite (75 tests across 6 classes) — 0 failures. Clean merge, docs-check passes, no CODEOWNERS-protected paths.

One very minor, non-blocking thing: the ambiguity error message always says "across accounts" even in the same-account/different-region collision case (only reachable via floci's own testing override-id tag) — not worth holding this up for.

PR #2379, opened 2026-08-18
fix(cloudwatch): decompress gzipped cbor bodies.

### LINKED ISSUE #2277: [BUG] Cloudwatch incoming CBOR requests might be gzipped.  This isn't handled yet
## Service

CLoudwatch in this instance

## AWS API Action

puMetricData

## Expected behavior

The body should be decoded based on the header "Content-encoding".

## Actual behavior

the gzipped body fails CBOR standards and throws exceptions

## Reproduction

call putMetricData with enough data

## Environment

- Floci version / image tag: latest
- How you're running Floci (Docker / native / `mvn quarkus:dev`): as devservice / quarkus

MarcReimann-Otto:
## Summary

Closes #2277 

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

GZipped CBOR requests were not handled.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/AwsJsonCborController.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/AwsJsonCborController.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.core.common;
 
+import com.fasterxml.jackson.core.JacksonException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Verified this end-to-end, not just by reading the diff. Confirmed CloudWatch's SDKs really do gzip-compress `PutMetricData` bodies by default (per the AWS API reference and SDK compression docs), so this is fixing a real, common-path failure, not a hypothetical.

For the malformed-CBOR-after-decompression fix specifically, I wrote a temporary HTTP-level test hitting the actual dispatch endpoint (gzip `Content-Encoding` + truncated CBOR body) and confirmed it now returns 400 instead of the old empty 500. Reverting just the `catch (JacksonException e)` block reproduced the original bug, confirming the fix is load-bearing.

Also confirmed the size-limit off-by-one (`totalRead > maxRead`) and the test-class naming are both already corrected in the latest commit.

A couple of minor, non-blocking nits for a future pass: the unused `com.google.gson.JsonParseException` import can be dropped, and `payloadTooLargeExceptionThrown` could assert the specific error code rather than `AwsException.class` generically.

Approving.

PR #2384, opened 2026-08-18
fix(ec2): derive default vpc/subnet/sg ids per region

lex00:
Default VPC, subnet and security-group ids were hardcoded literals (`vpc-default`, `subnet-default-a/b/c`, `sg-default`), so every region echoed the same ids back to callers even though storage was already keyed per region. A consumer aggregating across regions saw fewer distinct resources than actually existed.

Added `Ec2Service.defaultVpcId/defaultSubnetId/defaultSecurityGroupId(region)` and used them at every seeding and fallback-resolution call site, plus the matching lookup in `RdsService.buildDefaultSubnetGroup`.

`Ec2Service.dhcpOptionsId`'s own default literal has the same pattern but no backing DHCP-options model, so I left it out of scope.

**Behavior change.** Default ids move from `vpc-default`/`subnet-default-a`/`sg-default` to `vpc-default-<region>`/`subnet-default-<region>-a`/`sg-default-<region>`. Anything hardcoding the old literal breaks, which is why this PR also touched four test files and a compatibility test. Existing persistent storage keeps working, `resolveDefaultVpcId`/`resolveDefaultSecurityGroupId` fall back to the old literal when that's what's already on file.

1667 tests pass across ec2, rds, eks, elbv2, autoscaling and cloudformation.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/RdsControlPlaneTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/RdsControlPlaneTest.java
@@ -232,8 +232,10 @@ class RdsControlPlaneTest {
         String regionalProxyName = TestFixtures.uniqueName("rds-proxy-regional");
         try (RdsClient east = rdsClient(Region.US_EAST_1);
              RdsClient west = rdsClient(Region.US_WEST_2)) {
-            var eastProxy = createIamProxy(east, regionalProxyName);
-            var westProxy = createIamProxy(west, regionalProxyName);
+            // Subnet ids are region-scoped on real AWS (and on floci, since #21), so each proxy
+            // needs subnets from its own signed region rather than the shared us-east-1 subnetIds.
+            var eastProxy = createIamProxy(east, regionalProxyName, subnetIdsFor(Region.US_EAST_1));
+            var westProxy = createIamProxy(west, regionalProxyName, subnetIdsFor(Region.US_WEST_2));
 
             assertThat(eastProxy.dbProxy().dbProxyArn()).contains(":rds:us-east-1:");
             assertThat(westProxy.dbProxy().dbProxyArn()).contains(":rds:us-west-2:");
@@ -344,13 +346,32 @@ class RdsControlPlaneTest {
         }
     }
 
-    private static CreateDbProxyResponse createIamProxy(RdsClient client, String name) {
+    private static CreateDbProxyResponse createIamProxy(RdsClient client, String name, List<String> vpcSubnetIds) {
         return client.createDBProxy(b -> b
                 .dbProxyName(name)
                 .engineFamily("POSTGRESQL")
                 .roleArn("arn:aws:iam::000000000000:role/rds-proxy-regional")
-                .vpcSubnetIds(subnetIds)
+                .vpcSubnetIds(vpcSubnetIds)
                 .defaultAuthScheme("IAM_AUTH"));
+    }
+
+    private static List<String> subnetIdsFor(Region region) {
+        try (Ec2Client ec2 = ec2Client(region)) {
+            return ec2.describeSubnets().subnets().stream()
+                    .map(subnet -> subnet.subnetId())
+                    .sorted()
+                    .limit(2)
+                    .toList();
+        }
+    }
+
+    private static Ec2Client ec2Client(Region region) {
+        return Ec2Client.builder()
+                .endpointOverride(TestFixtures.endpoint())
+                .region(region)
+                .credentialsProvider(StaticCredentialsProvider.create(
+                        AwsBasicCredentials.create("test", "test")))
+                .build();
     }
 
     private static software.amazon.awssdk.services.rds.model.CreateDbInstanceResponse createDbInstance(
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -872,7 +872,7 @@ public class Ec2Service implements ContainerTeardown {
         return region + "::" + id;
     }
 
-    // Default resource ids must be derived per region (#21): every region seeds its own
+    // Default resource ids must be derived per region (lex00/floci#21): every region seeds its own
     // default VPC/subnets/security group independently, but a literal id like "vpc-default"
     // is identical text in every region's response even though storage itself is already
     // correctly keyed by region. RdsService derives the same ids from these methods so its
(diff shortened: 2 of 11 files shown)


lex00:
Fixed. `ensureDefaultResources` skips reseeding a region that already has a VPC on file, so old storage keeps its unscoped id forever, and any lookup computing the new scoped id would miss it. Added `resolveDefaultVpcId`/`resolveDefaultSecurityGroupId`, scoped id first, fall back to the old literal. Used everywhere that looks one up, including `RdsService.buildDefaultSubnetGroup`. New test seeds a legacy VPC/SG directly and checks both resolvers find them. 1637 tests pass.

lex00:
The CI failure wasn't a regression. `RdsControlPlaneTest.sdkScopesTheSameDbProxyNameBySignedRegion` fetched subnet ids once from us-east-1 and reused them for both proxies. That only worked because ids collided across regions before this fix. Real AWS scopes subnet ids per region and would reject this test as written. Fixed it to fetch subnets per signed region. 11/11 pass locally.

pgermosen:
@lex00 

This PR has a real merge conflict with `main` now (confirmed via an actual `git merge`, not just GitHub's status badge) — it needs a rebase.

It's a single, trivial conflict: `CloudFormationRdsIntegrationTest.java` has competing import additions on the same line — this branch adds `import io.github.hectorvent.floci.services.ec2.Ec2Service;`, while #2385 (already merged) added `import io.github.hectorvent.floci.core.common.XmlParser;` at the same spot. Everything else in the diff (`Ec2Service.java`, `RdsService.java`, and the rest of the touched test files) auto-merges cleanly — this is the only file with a conflict marker.

lex00:
Rebased and pushed. Found a second case of the same thing while I was in there: `dbSubnetGroupTemplate` in `CloudFormationRdsIntegrationTest` (added by #2385) also hardcoded the old unscoped subnet literals, broke the same way once rebased. Fixed to match the sibling test in the same file. 1667/1667 pass.

pgermosen (approved):
@lex00 — approving. The rebase resolved the conflict cleanly, and the second instance you found while you were in there was a good catch.

## Verified

Rebuilt against current `main` (`74da1576`): merges cleanly, compiles, and **1667 tests pass** across ec2, rds, eks, elbv2, autoscaling and cloudformation — the same figure you reported, which I checked rather than assumed.

I reverted each half of the change locally to confirm the behaviour is actually held in place rather than merely accompanied by tests:

- dropping the region suffix from the id — caught;
- dropping the legacy `vpc-default` / `sg-default` fallback — caught.

## The part I want to credit explicitly

The apparent asymmetry is correct, and it took me a few minutes to satisfy myself of that, so it is worth writing down for whoever reads this next.

`resolveDefaultVpcId` and `resolveDefaultSecurityGroupId` have legacy fallbacks; there is deliberately no `resolveDefaultSubnetId`. That looks like an omission until you follow `buildDefaultSubnetGroup`, which never computes a subnet id — it queries by `vpc-id` filter using `resolveDefaultVpcId(region)`. Legacy-persisted subnets are therefore reached transitively through the VPC resolver, and `defaultSubnetId` is only ever used at seeding time. The asymmetry is load-bearing, not an oversight, and a future reviewer may well try to "complete the set".

The backward-compatibility reasoning behind the fallbacks is the right call too. Since `ensureDefaultResources` skips a region that already has a VPC on file, existing persistent storage keeps its unscoped ids indefinitely — without the fallback, every computed-id lookup would miss and a region would quietly lose its default VPC and security group to every caller. The comment in the code explains that well enough that nobody should be tempted to simplify it away.

## Two small things


**1. The behaviour change has no callout.** Default ids move from `vpc-default` to `vpc-default-<region>`. Persisted storage is protected by the fallbacks, so the blast radius is genuinely small — but anything that hardcodes the old literal still breaks, which is precisely why this PR had to touch four test files and a compatibility test. The commit is `fix(ec2):` with nothing in the description flagging it. Worth a line in the description, and worth carrying into the release notes, so somebody with a local fixture or script pinned to `vpc-default` finds out from the notes rather than from a failure.

**2. A code comment references `#21`.** The resolver comment mentions "the pre-#21 unscoped literal". Read in this repository that points at `floci-io/floci#21`, which is not what you mean — from the surrounding context it looks like a reference to your own series. Since it is a permanent comment rather than a PR remark, spelling it out fully would save the next reader a confusing detour. Trivial, and fine to fold into whatever you touch next.

## On the earlier round

Two things from this PR's history are worth saying plainly. You diagnosed the compatibility-test failure as a pre-existing test defect rather than a regression — the test fetched subnet ids from one region and reused them for both proxies, which only ever worked *because* ids collided across regions, and which real AWS would reject as written. That is the correct reading, and the tempting alternative would have been to make the failure go away.

And when you rebased, you did not stop at the conflict marker: finding that `dbSubnetGroupTemplate` had the same hardcoded literals and would break the same way is the kind of thing that only turns up if you actually run the thing afterwards.

Thank you for the quick turnaround on the rebase.

lex00:
Description now has a behavior-change callout, and both `#21` comment references are spelled out as `lex00/floci#21`. Pushed as `23121345`.

pgermosen (approved):
@lex00 — re-approving on `23121345`.

My earlier approval was on `0ae9f43a`, and the head has moved since. GitHub keeps an approval alive across pushes, so the tick was covering a commit that is no longer there. The delta is two comment lines and nothing substantive, but the record should reflect what is actually in the branch rather than what was there this morning.

## Both follow-ups landed

I raised both as explicitly non-blocking, so acting on them was optional. Taking them anyway is the reason this is worth a second look rather than a rubber stamp.

**The behaviour-change note** is now in the description, and it says more than I asked for:

> Anything hardcoding the old literal breaks, which is why this PR also touched four test files and a compatibility test. Existing persistent storage keeps working, `resolveDefaultVpcId`/`resolveDefaultSecurityGroupId` fall back to the old literal when that's what's already on file.

That is the complete picture in three sentences — what moves, what breaks, why the test churn was necessary rather than incidental, and why persisted data is safe. Someone hitting this after an upgrade gets their answer without opening the diff, which is the whole point of putting it in the description.

**The `#21` references** now read `lex00/floci#21` and `pre-lex00/floci#21`, so they no longer resolve to `floci-io/floci#21` for anyone reading them in this repository. A small thing, but it lives in a permanent comment and would have quietly misled every future reader.

## Re-verified

Against current `main` (`29b0d600`): merges cleanly, compiles, and **1674 tests pass** across ec2, rds, eks, elbv2, autoscaling and cloudformation. All fourteen checks green on the new head.

Everything from the first review stands. The subnet asymmetry is still the part I would most want a future reviewer to read the comments on before "completing the set" — `resolveDefaultSubnetId` is absent because `buildDefaultSubnetGroup` queries by `vpc-id` rather than computing subnet ids, so legacy subnets resolve transitively through the VPC fallback.

Ready to merge.

PR #2385, opened 2026-08-18
fix(cloudformation): no-op stack update for unchanged fixed-name resources

lex00:
`provision()` runs on every resource on every `UpdateStack`, whether or not that resource changed. Several provisioner methods called their service's create unconditionally, so a fixed-name resource left unchanged between deploys collided with itself and rolled back the whole stack.

Fixed `provisionDbSubnetGroup`, `provisionDbParameterGroup`, `provisionDbClusterParameterGroup`, `provisionDbInstance`, `provisionDbCluster`, `provisionKinesisStream` and `provisionAutoScalingGroup` to check the existing physical id and reconcile in place instead. `provisionLaunchConfiguration` is a no-op by design since real AWS has no update API for it. `provisionCloudWatchAlarm` and `provisionCloudFrontDistribution` already handled this correctly.

1159 tests pass across cloudformation, rds, kinesis and autoscaling.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -1581,6 +1581,9 @@ public class CloudFormationResourceProvisioner {
         try {
             return lookup.apply(name);
         } catch (AwsException notFound) {
+            // Expected when the resource was deleted out of band since the prior update; the
+            // caller falls back to creating it fresh under the same name.
+            LOG.debugv(notFound, "No existing {0} found on file, falling back to create", name);
             return null;
         }
     }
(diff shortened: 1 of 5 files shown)


lex00:
Fixed, logged the tolerated not-found in `sameNameExistingResource`, matching `deleteRenamedResource` right below it. 67 tests pass.

pgermosen (approved):
Verified this thoroughly, including against real AWS docs — the code comments' behavioral claims all hold up against primary sources: `AWS::AutoScaling::LaunchConfiguration`'s CloudFormation reference confirms every property is replacement-only, `UpdateAutoScalingGroup`'s real parameter list confirms it has no TargetGroupARNs/LoadBalancerNames/Tags, and `ModifyDBParameterGroup`'s parameter list confirms Family/Description are immutable.

Also reverted the Kinesis reconciliation call and confirmed the new integration test genuinely reproduces the original `UPDATE_ROLLBACK_COMPLETE` bug, then restored and confirmed it passes. Full suite across cloudformation/rds/kinesis/autoscaling: 1159 tests, 0 failures, matching your own numbers. Hygiene is clean.

Approving.

Issue #38, opened 2026-03-24
[FEAT] IAM Enforcement Mode


freakinhippie:
I recognize that this may not align with the project's goals, and if that is the case then feel free to close this issue.

## Service

IAM

## API Action / Feature

Policy Enforcement

## AWS Documentation

[IAM API Reference](https://docs.aws.amazon.com/IAM/latest/APIReference/welcome.html)

## Why is this needed?

Testing IAM users/groups/roles/policies/etc. It is often useful to be able to test IAM policy attached to or covering resources. For example, an S3 bucket may have a bucket policy which restricts access to a limited set of principals; a lambda execution role may only allow access to DynamoDB, a Service Control Policy may block entire services, etc. It would be great to be able to use some flag to activate an IAM policy enforcement mode so that these policies could be tested along with the other bits.

## Are you willing to contribute a PR?

- [ ] Yes
- [X] No

Issue #25, opened 2026-03-24
[BUG] OIDC Well-Known endpoints return S3 NoSuchBucket error


TAKEDA-Takashi:
## Service

Cognito

## AWS API Action

GET `/.well-known/openid-configuration`
GET `/.well-known/jwks.json`

## Expected behavior

OIDC discovery document and JWKS should be returned, as described in the [Cognito documentation](https://hectorvent.dev/floci/services/cognito/).

## Actual behavior

Both endpoints return an S3 `NoSuchBucket` XML error:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error>
  <Code>NoSuchBucket</Code>
  <Message>The specified bucket does not exist.</Message>
</Error>
```

## Root cause analysis

After reading the source code, this appears to be caused by two issues:

1. **S3Controller catch-all pattern**: `S3Controller` defines `/{bucket}/{key:.+}`, which matches `/.well-known/openid-configuration` as `bucket=.well-known`, `key=openid-configuration`. This takes priority over Cognito routes.

2. **CognitoWellKnownController path mismatch**: The controller defines `GET /{poolId}/.well-known/jwks.json`, which requires a `poolId` segment before `.well-known`. A direct request to `/.well-known/jwks.json` does not match this pattern. Additionally, `/.well-known/openid-configuration` does not appear to be implemented.

## Reproduction

```bash
docker compose up -d

# Both return S3 NoSuchBucket error
curl http://localhost:4566/.well-known/openid-configuration
curl http://localhost:4566/.well-known/jwks.json
```

## Environment

- Floci version: 1.0.4 (native image)
- Docker: Rancher Desktop on macOS

```yaml
# docker-compose.yml
services:
  floci:
    image: hectorvent/floci:latest
    user: root
    ports:
      - "4566:4566"
    tmpfs:
      - /app/data
    environment:
      FLOCI_STORAGE_MODE: memory
```

PR #2386, opened 2026-08-18
fix(memorydb,neptune,docdb): degrade gracefully with no Docker daemon

lex00:
Same daemonless pattern already used by ecr, eks, msk, opensearch and rds, extended to the three services still missing it. Control-plane operations now reach the AWS steady state without a Docker daemon instead of leaking a raw `SocketException`.

110 tests pass across memorydb, neptune and docdb.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/docdb/DocDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/docdb/DocDbService.java
@@ -79,13 +79,25 @@ public class DocDbService {
         } else {
             String image = config.services().docdb().defaultImage();
             LOG.infov("Creating DocDB cluster {0}, image={1}", id, image);
-            DocDbContainerHandle handle = containerManager.start(id, image, masterUsername, masterPassword);
-            cluster.setEndpoint(handle.getHost());
-            cluster.setReaderEndpoint(handle.getHost());
-            cluster.setPort(handle.getPort());
-            cluster.setContainerId(handle.getContainerId());
-            cluster.setContainerHost(handle.getHost());
-            cluster.setContainerPort(handle.getPort());
+            // A cluster record is metadata: its identifier, ARN and tags need no Docker, so the
+            // cluster is created and reaches 'available' even when no daemon is reachable. Only
+            // connecting to the database needs the container.
+            DocDbContainerHandle handle = containerManager.tryStart(id, image, masterUsername, masterPassword);
+            if (handle != null) {
+                cluster.setEndpoint(handle.getHost());
+                cluster.setReaderEndpoint(handle.getHost());
+                cluster.setPort(handle.getPort());
+                cluster.setContainerId(handle.getContainerId());
+                cluster.setContainerHost(handle.getHost());
+                cluster.setContainerPort(handle.getPort());
+            } else {
+                cluster.setEndpoint(resolveEndpointHost());
+                cluster.setReaderEndpoint(resolveEndpointHost());
+                cluster.setPort(MONGO_PORT);
+                LOG.warnv("DocDB cluster {0} created without a backing MongoDB container: no "
+                        + "Docker daemon is reachable. Metadata operations work; connections to "
+                        + "the cluster do not until a daemon appears.", id);
+            }
         }
 
         clusters.put(id, cluster);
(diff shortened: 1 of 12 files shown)


lex00:
Fixed the test logging in all three files. On the reconciliation gap: none of these three services can reconnect a container today, daemon or no daemon. Tracked in lex00/floci#54.

src/main/java/io/github/hectorvent/floci/services/docdb/DocDbService.java
```diff
@@ -79,13 +79,25 @@ public DocDbCluster createDbCluster(String id, String engineVersion,
...
-            cluster.setEndpoint(handle.getHost());
-            cluster.setReaderEndpoint(handle.getHost());
-            cluster.setPort(handle.getPort());
-            cluster.setContainerId(handle.getContainerId());
-            cluster.setContainerHost(handle.getHost());
-            cluster.setContainerPort(handle.getPort());
+            // A cluster record is metadata: its identifier, ARN and tags need no Docker, so the
+            // cluster is created and reaches 'available' even when no daemon is reachable. Only
+            // connecting to the database needs the container.
+            DocDbContainerHandle handle = containerManager.tryStart(id, image, masterUsername, masterPassword);
```
pgermosen, line 85: Unlike the cited precedents (RDS #38, ECR #25), there's no per-call retry to reconnect the container once Docker becomes reachable again — a cluster created during an outage stays metadata-only forever. Confirmed this is a pre-existing gap (none of docdb/memorydb/neptune have any container-reconnection path at all, for any cause) rather than a regression, and it's tracked in lex00/floci#54, so this seems like a reasonable scope call — just flagging that "the same daemonless pattern" overstates it a bit, since retry-on-recovery is the part RDS/ECR call out as the actual point of the pattern ("nothing is latched off permanently").

pgermosen (commented):
Re-verified the merge against current main (with the actual merged tree compiled and tested, not just checked for textual conflicts) — clean. 110 tests pass across docdb/memorydb/neptune.

Small correction: the description says the pattern is "already used by ecr, eks, elasticache, msk, opensearch and rds" — but lex00/floci#44 (the tracking issue this closes toward) lists elasticache as one of the services *still missing* it, and `ElastiCacheService.createReplicationGroup` currently rolls back and re-throws on any container failure rather than degrading gracefully. Not a code issue, just the PR text.

One more thing worth knowing, not blocking:

lex00:
Confirmed, thanks for checking. elasticache's `createReplicationGroup` doesn't degrade gracefully on a container failure. Retry-on-recovery is tracked in #54.

pgermosen (approved):
@lex00 — approving. One small thing to fix in passing, in the description rather than the code.

## The description still carries the elasticache claim

You confirmed the point in a comment earlier today — *"elasticache's `createReplicationGroup` doesn't degrade gracefully on a container failure"* — but the description itself is unchanged and still reads:

> Same daemonless pattern already used by ecr, eks, elasticache, msk, opensearch and rds, extended to the three services still missing it.

Could you drop `elasticache` from that list before this merges? It matters slightly more than a typo would, for two reasons: the description becomes the squash-merge commit body, so the claim outlives the PR; and the whole point of the tracking issue is to know which services still need this, so a merged commit asserting elasticache is already done is the one thing likeliest to mislead whoever picks that work up.

No rush and nothing else needed — I am approving now so it is not gated on an edit.

## Verified

`main` moved eleven commits since I last checked, so I redid it rather than relying on yesterday's result: merges cleanly, compiles, and **110 tests pass** across docdb, memorydb and neptune. Disabling the daemon-unreachable branch, so every container failure rethrows as before, breaks the manager coverage — the new behaviour is held in place rather than merely accompanied.

## What I checked most closely

Three near-identical files landing at once is where copy-paste errors hide, so I compared them against each other rather than reading one and assuming.

They hold up. Each warning names its own service — DocumentDB, MemoryDB, Neptune — and, more to the point, the service-side handling is genuinely adapted rather than duplicated: DocDb fills the endpoint and reader endpoint from `resolveEndpointHost()` when there is no handle, while MemoryDb and Neptune correctly skip proxy startup instead. Each carries a comment explaining why a cluster record is metadata in that particular service's terms. That is the version of "apply an existing pattern three times" that survives later editing; the version where all three are byte-identical except the class name is the one that rots.

## On the two points from the earlier round

Both settled, and I am content with where they landed.

The retry-on-recovery gap is real but pre-existing — none of these three services has any container-reconnection path today, for any cause — so tracking it separately rather than widening this PR is the right call. I would still gently suggest that "the same daemonless pattern" is doing some work in that sentence, since retry-on-recovery is the part RDS and ECR call out as the actual point of it, but that is a wording matter and not worth another round.

Thank you for confirming the elasticache reading rather than waving it through; that is the answer that makes the tracking issue trustworthy.

lex00:
Dropped elasticache from the description.

PR #2387, opened 2026-08-18
test(ec2): make Ec2IntegrationTest idempotent across repeated runs in one JVM

lex00:
Six tests in `Ec2IntegrationTest` used fixed resource names with no teardown, so running the class twice in one Surefire fork failed on the second pass: `createKeyPair`, `importKeyPair`, and three `createImage`/`registerImage` tests hit duplicate-name errors, and `spotInstanceLifecycle` picked up a leftover tagged request from an unordered scan. Switched all six to per-invocation random suffixes.

Verified with 5 consecutive double-pass runs, all clean. 165 tests pass single-pass.


pgermosen (approved):
Clean fix — traced the `@Order`-based field dependencies (`keyPairName`, `importedKeyName`) by hand and confirmed each is written before any test that reads it, so the switch to per-invocation values is safe. Full class: 165/165 pass. Hygiene is clean.

Approving.

PR #2389, opened 2026-08-18
fix(dynamodb): return LastEvaluatedKey when Limit stops exactly at the last item

### LINKED ISSUE #2383: [BUG] DynamoDB: Query/Scan omit LastEvaluatedKey when Limit stops exactly at the last item
## Service

DynamoDB

## AWS API Action

Query, Scan

## Expected behavior

When a Query/Scan stops because it reached `Limit`, real DynamoDB returns a `LastEvaluatedKey` **even if the stop position happens to be the last item of the result set**. DynamoDB does not look ahead; the client learns it has reached the end only when a follow-up request comes back without `LastEvaluatedKey`. From the [developer guide](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html):

> A non-empty `LastEvaluatedKey` only means that the previous Query stopped at a page boundary—the 1 MB page-size limit or a `Limit` value—not that more matching items necessarily remain. [...] The only way to know when you have reached the end of the result set is when `LastEvaluatedKey` is empty.

AWS's own DynamoDB Local (`amazon/dynamodb-local:latest`) and LocalStack 4.12.0 both return the key here (verified with the reproduction below):

```json
{"Count": 3, "LastEvaluatedKey": {"sk": {"N": "3"}, "pk": {"N": "1"}}}
```

Actual behavior

Floci omits LastEvaluatedKey whenever the Limit boundary coincides with the end of the result set, so clients see end-of-results one request too early:

```json
{"Count": 3, "LastEvaluatedKey": null}
```

Scan shows the same divergence through its own code path. Limit values that stop mid-result-set behave correctly (the key is returned).

Reproduction

```
EP=http://localhost:4566   # Floci endpoint
aws --endpoint-url=$EP dynamodb create-table --table-name lek-test \
  --attribute-definitions AttributeName=pk,AttributeType=N AttributeName=sk,AttributeType=N \
  --key-schema AttributeName=pk,KeyType=HASH AttributeName=sk,KeyType=RANGE \
for i in 1 2 3; do
  aws --endpoint-url=$EP dynamodb put-item --table-name lek-test \
    --item "{\"pk\":{\"N\":\"1\"},\"sk\":{\"N\":\"$i\"}}"
done

# 3 items, Limit=3: stops exactly at the last item
aws --endpoint-url=$EP dynamodb query --table-name lek-test \
  --key-condition-expression "pk = :p" \
  --expression-attribute-values '{":p":{"N":"1"}}' \
  --limit 3 --query '{Count:Count,LastEvaluatedKey:LastEvaluatedKey}'
# Floci:          {"Count": 3, "LastEvaluatedKey": null}
# dynamodb-local: {"Count": 3, "LastEvaluatedKey": {"sk":{"N":"3"},"pk":{"N":"1"}}}

# Scan hits the same divergence:
aws --endpoint-url=$EP dynamodb scan --table-name lek-test \
  --limit 3 --query '{Count:Count,LastEvaluatedKey:LastEvaluatedKey}'
# Floci: LastEvaluatedKey is null; dynamodb-local/LocalStack return the key
```

Environment

- Floci version / image tag: floci/floci:1.7.0-compat (Docker)
- Java SDK version (if applicable): AWS SDK for Java v2 / AWS CLI v2
- How you're running Floci: Docker Compose

yoshinori-aono-om:
## Summary

Query and Scan omitted `LastEvaluatedKey` whenever the `Limit` boundary coincided with the end of the result set, so clients saw end-of-results one request too early. Real DynamoDB does not look ahead: any read that stops at the `Limit` boundary returns a cursor, and the client only learns it reached the end when a follow-up request comes back empty without one.

Two small changes in `DynamoDbService`:

- Query: `evaluatedItems.size() > limit` → `>= limit` — the `==` case previously fell through with no cursor
- Scan: removed the `it.hasNext()` look-ahead guard before surfacing the cursor

PartiQL `ExecuteStatement` reuses these paths, so it inherits the fix.

Closes #2383

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: at the exact-`Limit` boundary Floci omitted `LastEvaluatedKey` for ascending Query, descending Query (`ScanIndexForward=false`), GSI Query, and Scan, while `amazon/dynamodb-local:latest` and LocalStack 4.12.0 return the cursor in all four cases (full comparison matrix in #2383). Per the [developer guide](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html), "the absence of LastEvaluatedKey is the only way to know that you have reached the end of the result set."

Verified with AWS CLI v2 and AWS SDK for Java v2: after this change the same 10-case matrix matches `amazon/dynamodb-local` and LocalStack exactly, including the follow-up request returning an empty page with no cursor. Also verified end-to-end against the production API test suite that originally surfaced the bug (a cursor-pagination endpoint now behaves identically to real AWS).

## Checklist

- [x] `./mvnw test` passes locally (all 410 DynamoDB suite tests, including the new ones)
- [x] New or updated integration test added (`DynamoDbLimitBoundaryPaginationIntegrationTest` — exact-Limit boundary for ascending/descending Query, GSI Query, and Scan, each with the cursor round-trip proving the follow-up page is empty and final, plus guards for the mid-result-set and no-Limit cases)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -785,8 +785,13 @@ public class DynamoDbService {
         List<JsonNode> evaluatedItems = results;
         JsonNode lastEvaluatedKey = null;
 
-        // Apply Limit (stops at N items)
-        if (limit != null && limit > 0 && evaluatedItems.size() > limit) {
+        // Apply Limit (stops at N items). DynamoDB does not look ahead: whenever a
+        // read stops at the Limit boundary it returns a LastEvaluatedKey, even when
+        // the boundary happens to be the last item of the result set. The client
+        // only learns it reached the end when a follow-up request comes back without
+        // a LastEvaluatedKey ("the absence of LastEvaluatedKey is the only way to
+        // know that you have reached the end of the result set").
+        if (limit != null && limit > 0 && evaluatedItems.size() >= limit) {
             JsonNode lastItem = evaluatedItems.get(limit - 1);
             lastEvaluatedKey = buildKeyNode(table, lastItem, pkName, sortKeyNames, indexName != null);
             evaluatedItems = new ArrayList<>(evaluatedItems.subList(0, limit));
@@ -881,12 +886,13 @@ public class DynamoDbService {
                         && (scanFilter == null || matchesScanFilter(item, scanFilter));
                 if (matched) results.add(item);
             }
-            // Limit caps SCANNED items (those read), not matched items. Stop at the
-            // limit and surface a cursor when more items remain to be examined.
+            // Limit caps SCANNED items (those read), not matched items. DynamoDB does
+            // not look ahead when it stops at the Limit boundary: it always surfaces a
+            // cursor, even when the boundary happens to be the last item of the result
+            // set. The client only learns it reached the end when a follow-up request
+            // comes back without a LastEvaluatedKey.
             if (limit != null && limit > 0 && totalScanned >= limit) {
-                if (it.hasNext()) {
-                    lastEvaluatedKey = buildKeyNode(table, item, lekPkName, lekSkName, indexScan);
-                }
+                lastEvaluatedKey = buildKeyNode(table, item, lekPkName, lekSkName, indexScan);
                 break;
             }
         }
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -785,31 +785,33 @@ public class DynamoDbService {
         List<JsonNode> evaluatedItems = results;
         JsonNode lastEvaluatedKey = null;
 
-        // Apply Limit (stops at N items). DynamoDB does not look ahead: whenever a
-        // read stops at the Limit boundary it returns a LastEvaluatedKey, even when
-        // the boundary happens to be the last item of the result set. The client
-        // only learns it reached the end when a follow-up request comes back without
-        // a LastEvaluatedKey ("the absence of LastEvaluatedKey is the only way to
-        // know that you have reached the end of the result set").
-        if (limit != null && limit > 0 && evaluatedItems.size() >= limit) {
-            JsonNode lastItem = evaluatedItems.get(limit - 1);
-            lastEvaluatedKey = buildKeyNode(table, lastItem, pkName, sortKeyNames, indexName != null);
-            evaluatedItems = new ArrayList<>(evaluatedItems.subList(0, limit));
-        }
-
-        // Apply 1MB response size limit (DynamoDB stops reading when scanned data exceeds 1MB)
-        if (lastEvaluatedKey == null) {
-            final int MAX_RESPONSE_BYTES = 1024 * 1024;
-            int accSize = 0;
-            for (int i = 0; i < evaluatedItems.size(); i++) {
-                int sz = DynamoDbItemSize.calculateItemSize(evaluatedItems.get(i));
-                if (accSize > 0 && accSize + sz > MAX_RESPONSE_BYTES) {
-                    lastEvaluatedKey = buildKeyNode(table, evaluatedItems.get(i - 1), pkName, sortKeyNames, indexName != null);
-                    evaluatedItems = new ArrayList<>(evaluatedItems.subList(0, i));
-                    break;
-                }
-                accSize += sz;
+        // Stop at whichever boundary the read reaches first: the 1 MB cap or Limit.
+        // The size check comes first for each item — per the Query API reference,
+        // "if the processed dataset size exceeds 1 MB before DynamoDB reaches this
+        // limit, it stops the operation", so an item that would cross the 1 MB cap
+        // is not read and does not count toward Limit. DynamoDB does not look ahead
+        // either way: a read that stops at a boundary always returns a
+        // LastEvaluatedKey, even when the boundary happens to be the last item of
+        // the result set ("the absence of LastEvaluatedKey is the only way to know
+        // that you have reached the end of the result set").
+        final int MAX_RESPONSE_BYTES = 1024 * 1024;
+        int accSize = 0;
+        int included = -1; // index one past the last included item; -1 = no boundary hit
+        for (int i = 0; i < evaluatedItems.size(); i++) {
+            int sz = DynamoDbItemSize.calculateItemSize(evaluatedItems.get(i));
+            if (accSize > 0 && accSize + sz > MAX_RESPONSE_BYTES) {
+                included = i; // the 1 MB cap stops the read BEFORE this item
+                break;
             }
+            accSize += sz;
+            if (limit != null && limit > 0 && i + 1 >= limit) {
+                included = i + 1; // the Limit cap stops the read AFTER this item
+                break;
+            }
+        }
+        if (included > 0) {
+            lastEvaluatedKey = buildKeyNode(table, evaluatedItems.get(included - 1), pkName, sortKeyNames, indexName != null);
+            evaluatedItems = new ArrayList<>(evaluatedItems.subList(0, included));
         }
 
         int scannedCount = evaluatedItems.size();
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
```diff
@@ -785,8 +785,13 @@ public QueryResult query(String tableName, JsonNode keyConditions,
...
 
-        // Apply Limit (stops at N items)
-        if (limit != null && limit > 0 && evaluatedItems.size() > limit) {
+        // Apply Limit (stops at N items). DynamoDB does not look ahead: whenever a
+        // read stops at the Limit boundary it returns a LastEvaluatedKey, even when
+        // the boundary happens to be the last item of the result set. The client
+        // only learns it reached the end when a follow-up request comes back without
+        // a LastEvaluatedKey ("the absence of LastEvaluatedKey is the only way to
+        // know that you have reached the end of the result set").
+        if (limit != null && limit > 0 && evaluatedItems.size() >= limit) {
```
pgermosen: Empirically confirmed this bypasses the 1MB response cap. I created 3 items of ~360KB each (~1.08MB combined) and ran a `Query` with `Limit: 3`. Real DynamoDB would stop at item 2 once the 1MB boundary is crossed; this returned `Count: 3` — all ~1.08MB in one response.

The cause: the 1MB size-check block right below this one is gated on `lastEvaluatedKey == null`. Before this change, `evaluatedItems.size() > limit` was `false` when the candidate count exactly equalled `Limit`, so that case fell through into the size check. Now that it's `>=`, the Limit block always sets `lastEvaluatedKey` in the exact-match case, which skips the size check entirely — the one case this PR is specifically trying to fix (exact-Limit match) is also the one case that now bypasses the 1MB cap.

A fix probably needs the size check to run *before* the Limit check (or fold Limit and size into one pass), rather than treating them as two independent, order-dependent gates.

src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
```diff
@@ -881,12 +886,13 @@ public ScanResult scan(String tableName, String filterExpression,
...
+            // Limit caps SCANNED items (those read), not matched items. DynamoDB does
+            // not look ahead when it stops at the Limit boundary: it always surfaces a
+            // cursor, even when the boundary happens to be the last item of the result
+            // set. The client only learns it reached the end when a follow-up request
+            // comes back without a LastEvaluatedKey.
             if (limit != null && limit > 0 && totalScanned >= limit) {
-                if (it.hasNext()) {
-                    lastEvaluatedKey = buildKeyNode(table, item, lekPkName, lekSkName, indexScan);
-                }
+                lastEvaluatedKey = buildKeyNode(table, item, lekPkName, lekSkName, indexScan);
```
pgermosen, line 920: Traced this one rather than reproduced it live, but the code path is real: `buildKeyNode` silently omits any key attribute the item doesn't have (`if (pkAttr != null)` / `if (skAttr != null)`), and floci's index Scan iterates every base-table item without excluding ones lacking the GSI's key attributes — unlike real AWS, where a sparse GSI simply never contains such items. If the exact-Limit boundary lands on one of those items, the resulting `LastEvaluatedKey` is missing the GSI key fields a real cursor would always carry.

Worth noting this PR's own new tests don't cover this combination — `gsiQueryLimitAtEndReturnsLastEvaluatedKeyWithIndexAndTableKeys` exercises GSI **Query**, not Scan, and `scanLimitAtEndReturnsLastEvaluatedKeyAndFollowUpPageIsEmpty` is a base-table Scan with no `IndexName`, so `lekPkName`/`lekSkName` never diverge from the always-present base table keys in either test.

pgermosen (commented):
Confirmed Greptile's two P1 findings are real — the exact-Limit fix here is right, but it interacts badly with the existing size/index-scan logic. Details below.

src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
yoshinori-aono-om: Good catch — my change did let the exact-`Limit` case bypass the cap. Fixed in 5b5414b. Thanks for putting the repro together — the 3 × ~360 KB experiment made the required semantics unambiguous.

Query and Scan now evaluate the size cap and `Limit` in a single pass, stopping at whichever boundary the read reaches first. The size check runs first for each item, per the Query API reference: "if the processed dataset size exceeds 1 MB **before** DynamoDB reaches this limit, it stops the operation" (emphasis mine) — so an item that would cross the cap is not read, doesn't count toward `Limit`/`ScannedCount`, and the cursor anchors to the last item actually read. Your exact scenario now behaves as expected: three ~360 KB items with `Limit: 3` return `Count: 2` with the cursor on item 2, and the follow-up page returns the last item with no cursor.

One thing worth noting for the record: while building the regression test I discovered that neither `amazon/dynamodb-local` nor LocalStack implements the 1 MB page cap — both return the full ~1.08 MB result set in a single page (dynamodb-local with no cursor at all; LocalStack with one, but only because of the `Limit` boundary). So the new tests' size expectations assert the documented contract directly; there was no emulator behavior to cross-check against.

Also folded in: Scan now counts scanned (pre-filter) bytes toward the cap, so with a `FilterExpression` a page can return zero matching items and still carry a cursor pointing at an unreturned item — covered by a dedicated test.

src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
yoshinori-aono-om: Fixed in 5b5414b — thanks for tracing this through the code; I'd missed the interaction with my change entirely.

Index Scans now skip base-table items missing any index key attribute — the same `hasNonNullAttribute` criterion Query already applies — so they are never read, never counted toward `ScannedCount`, and can never anchor a cursor. I verified the expectations against `amazon/dynamodb-local`: sparse items are excluded from both `Count` and `ScannedCount`, and the exact-`Limit` cursor carries the full index + table key set.

Good point on the coverage gap, too. The new tests include an index **Scan** at the exact-`Limit` boundary (asserting the cursor carries index + table keys) and a sparse-mix table that deliberately contains an item with the index HASH key but no RANGE key — the case a partition-key-only check would miss.

yoshinori-aono-om:
Both findings addressed in 5b5414b — thanks @pgermosen for catching both.

- **Size cap vs `Limit`**: the exact-`Limit` bypass my change introduced is fixed by evaluating both boundaries in a single pass (size check first, per the API reference wording), replacing the two order-dependent gates. As a side effect, this also fixes a pre-existing skip for `size() > limit` pages.
- **Sparse index Scans**: items missing index key attributes are excluded from the read entirely.

Coverage: 6 new integration tests (12 total in the file); the full DynamoDB suite passes (416 tests).

Known limitations left as-is (pre-existing, happy to file separately if useful): index-Scan `ExclusiveStartKey` isn't validated against the index key schema the way real AWS does, and size accounting uses the full item size regardless of index projection.

pgermosen (approved):
@yoshinori-aono-om — approving. Both findings are properly closed, and the first one you fixed more thoroughly than I asked.

## The size-cap fix

I raised the 1 MB bypass as an ordering problem and suggested either running the size check first or folding the two into one pass. You took the second, which is the better answer: two order-dependent gates would have stayed fragile whichever way round they went, and a single pass that stops at whichever boundary arrives first cannot develop the same class of bug again.

The reasoning is right too, and I verified the wording rather than take it from the PR. The [Query API reference](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html) says exactly this of `Limit`:

> Also, if the processed dataset size exceeds 1 MB **before** DynamoDB reaches this limit, it stops the operation…

So an item that would cross the cap not being read, not counting toward `Limit` or `ScannedCount`, and not anchoring the cursor is the correct reading of that sentence.

Two details in the implementation are worth calling out because they are easy to get wrong. Guarding the cap with `accSize > 0` means a single item larger than 1 MB is still returned rather than producing an empty page. And quoting the documentation inline, next to the branch it justifies, is what stops someone reordering those two checks later on the reasonable-looking grounds that it reads more naturally the other way.

That it also fixed a pre-existing skip for `size() > limit` pages is a good sign about the shape of the fix — collapsing the gates removed a bug nobody had reported yet.

## The sparse-index fix, and the coverage gap

Excluding items missing an index key attribute — with the same `hasNonNullAttribute` criterion Query already applied — is right, and using the existing predicate rather than a new one keeps the two paths from drifting.

You also closed the coverage gap I flagged. My concern was that the original tests exercised GSI **Query** and base-table **Scan**, so `lekPkName` and `lekSkName` never diverged from the always-present base-table keys and the sparse case could not surface. Removing the exclusion now fails two tests, both index Scans, and the sparse-mix table that carries an item with the index HASH key but no RANGE key is exactly the case a partition-key-only check would have missed.

## Verified

Merged against current `main` (`29b0d600`): compiles, and **416 DynamoDB tests pass** — the figure you quoted, which I checked rather than assumed.

The premise of the whole change is confirmed in the same API reference, which is worth having on the record since the behaviour looks like a bug to anyone who has not read it:

> If `LastEvaluatedKey` is not empty, it does not necessarily mean that there is more data in the result set. The only way to know when you have reached the end of the result set is when `LastEvaluatedKey` is empty.

Thank you for the repro work behind both rounds — comparing against `amazon/dynamodb-local` and LocalStack across a ten-case matrix is what made the original report actionable rather than arguable, and doing the same for the sparse expectations in the second round was the right instinct.

PR #2391, opened 2026-08-18
fix(apigateway): project HTTP API v2 cookies

aryasaatvik:
## Summary

Project Lambda proxy response cookies from API Gateway HTTP API payload format v2 as repeated `Set-Cookie` headers, without changing REST API v1 response handling.

## Behavior

| Lambda proxy response | Projected HTTP response |
| --- | --- |
| HTTP API v2 `cookies: ["session=one", "csrf=two"]` | Two separate `Set-Cookie` headers |
| HTTP API v2 `headers` alongside `cookies` | Ordinary headers and cookies are both preserved |
| REST API v1 `cookies` field | Ignored; existing `headers` and `multiValueHeaders` behavior is unchanged |

## Validation

- Focused API Gateway, JWT, and account-routing suite: 51 tests passed.
- `./mvnw test`: 10,845 tests passed, 9 skipped.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -401,7 +401,7 @@ public class ApiGatewayExecuteController {
         try {
             InvokeResult result = lambdaService.invoke(region, functionName, eventJson.getBytes(),
                     InvocationType.RequestResponse);
-            return buildProxyResponse(result);
+            return buildProxyResponse(result, false);
         } catch (AwsException e) {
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Thanks for the contribution, and welcome! Verified this against the real AWS API Gateway Developer Guide — the v1/v2 payload format split is exactly right: format 2.0's response `cookies` field projecting to repeated `Set-Cookie` headers, and format 1.0 having no such field at all. Also confirmed both `buildProxyResponse` call sites are correctly classified (REST v1 vs HTTP API v2), and reverted the fix to confirm the new test genuinely catches a regression. Full apigateway/apigatewayv2 suite passes.

Approving.

PR #2393, opened 2026-08-18
fix(apigateway): compile the trailing-slash tests against the current controller

ck-michals:
`main` does not compile its test sources, so CI is red on `main` itself and on every open PR.

`ApiGatewayExecuteController` gained a `RequestContext` parameter in #2377. The two tests added by
#2367 were written before that change and merged after it, so they construct the controller with
one argument too few:

```
[ERROR] BuildV2ProxyEventTrailingSlashTest.java:[45,22] constructor ApiGatewayExecuteController
        cannot be applied to given types;
[ERROR]   reason: actual and formal argument lists differ in length
```

This adds the missing argument to both, exactly as #2377 did for the sibling tests
(`BuildV2ProxyEventJwtClaimsTest` and the others already pass `…, new ApiGatewayExecuteRouteContext(), null, null`).

No behaviour changes — test sources only. `./mvnw test-compile` passes on this branch, where it
fails on `main`.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/apigateway/BuildV2AuthorizerEventTrailingSlashTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/apigateway/BuildV2AuthorizerEventTrailingSlashTest.java
@@ -46,7 +46,7 @@ class BuildV2AuthorizerEventTrailingSlashTest {
         controller = new ApiGatewayExecuteController(
                 null, null, null,
                 regionResolver, new ObjectMapper(), null,
-                null, null, null, null, new ApiGatewayExecuteRouteContext(), null
+                null, null, null, null, new ApiGatewayExecuteRouteContext(), null, null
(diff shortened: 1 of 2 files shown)


pgermosen:
This is already fixed on \`main\` — #2394 landed the same fix (adding the missing \`RequestContext\` argument to both files) at 2026-08-18 23:55 UTC, about an hour after this PR was opened. That's also why this PR now shows as conflicting: it's touching the same lines main already changed.

Confirmed by checking \`origin/main\`'s current copy of both files — the extra \`null\` argument is already there.

pgermosen:
@ck-michals heads up — this was already merged via #2394, which landed the same fix about an hour after this PR was opened. Feel free to close this one, or let us know if you'd rather we do it.

ck-michals:
Superseded by #2394, which landed the same fix. Closing.

Issue #2328, opened 2026-08-15
[FEAT] Firehose: apply CompressionFormat when delivering to S3


mihaiblidaru:
## Service

Kinesis Data Firehose (Amazon Data Firehose)

## API Action / Feature

S3 delivery data plane — apply the configured `CompressionFormat` (`UNCOMPRESSED` | `GZIP` | `ZIP` | `Snappy` | `HADOOP_SNAPPY`) from `S3DestinationConfiguration` / `ExtendedS3DestinationConfiguration` when writing delivered objects to S3.

## AWS Documentation

- ExtendedS3DestinationConfiguration (`CompressionFormat` member): https://docs.aws.amazon.com/firehose/latest/APIReference/API_ExtendedS3DestinationConfiguration.html
- S3DestinationConfiguration: https://docs.aws.amazon.com/firehose/latest/APIReference/API_S3DestinationConfiguration.html
- S3 object name format (compression extension): https://docs.aws.amazon.com/firehose/latest/dev/basic-deliver.html

## Why is this needed?

Floci accepts `CompressionFormat` on `CreateDeliveryStream`/`UpdateDestination` and echoes it back on `DescribeDeliveryStream` (matching AWS exactly on the control plane), but the delivery path never reads it: every delivered object contains the raw record bytes concatenated verbatim (uncompressed, with a newline appended after each record that does not already end with one), with no compression extension in the object key and no `Content-Encoding`. The docs already list this as a known deviation ([docs/services/firehose.md](https://github.com/floci-io/floci/blob/3cfa983479cbd9f7b8aa3d8bfb4b151992396be6/docs/services/firehose.md?plain=1#L40): "delivered content is never compressed, so no compression extension is ever added").

Because the control plane behaves perfectly, the divergence is silent: integration tests that assert on the stream description pass, and the mismatch only surfaces when a downstream consumer (Athena/Glue over `.gz` data, Spark/Hadoop jobs reading Snappy, log pipelines that gunzip by extension or `Content-Encoding`) reads the delivered objects and gets bytes in the wrong format.

### Ground truth: real AWS vs Floci

I ran the same experiment against real AWS (eu-west-1, 2026-08-15) and Floci 1.6.0: one `DirectPut` delivery stream per `CompressionFormat` with an `ExtendedS3DestinationConfiguration`, `PutRecordBatch` with 5 identical NDJSON records (690 bytes total), then fetched the delivered objects.

Real AWS:

[...]

PR #2406, opened 2026-08-19
fix: preserve Firehose compression format in CloudFormation

AbarnaaSree:
Fixes an issue where the CompressionFormat configured for an AWS Kinesis Data Firehose delivery stream was not propagated when the stream was created through CloudFormation.

Problem

When a Firehose delivery stream was provisioned using CloudFormation with S3DestinationConfiguration or ExtendedS3DestinationConfiguration, the CompressionFormat property was not copied to the internal S3Destination model.

For example:

{
  "S3DestinationConfiguration": {
    "BucketARN": "arn:aws:s3:::my-firehose-bucket",
    "RoleARN": "arn:aws:iam::000000000000:role/firehose-role",
    "Prefix": "logs/",
    "CompressionFormat": "GZIP"
  }
}

The CompressionFormat value was therefore lost during CloudFormation provisioning.

Solution

Updated CloudFormationResourceProvisioner to read and propagate CompressionFormat from the CloudFormation destination configuration:

s3.setCompressionFormat(
    blankToNull(engine.resolve(s3Node.path("CompressionFormat")))
);

This works for both:

S3DestinationConfiguration
ExtendedS3DestinationConfiguration
Tests

Added a regression case to CloudFormationFirehoseIntegrationTest using:

"CompressionFormat": "GZIP"
Verification
CloudFormationFirehoseIntegrationTest — 1 passed, 0 failures
git diff --check — passed
Maven build — BUILD SUCCESS
Related

This ensures CloudFormation-created Firehose delivery streams correctly preserve the requested S3 compression format.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -1981,6 +1981,10 @@ public class CloudFormationResourceProvisioner {
                 : (props != null ? props.get("S3DestinationConfiguration") : null);
         if (s3Node != null && !s3Node.isNull()) {
             s3 = new DeliveryStreamDescription.S3Destination();
+
+            s3.setCompressionFormat(
+                blankToNull(engine.resolve(s3Node.path("CompressionFormat")))
+            );
             s3.setBucketArn(blankToNull(engine.resolve(s3Node.path("BucketARN"))));
             s3.setPrefix(blankToNull(engine.resolve(s3Node.path("Prefix"))));
             if (s3Node.has("BufferingHints")) {
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationFirehoseIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationFirehoseIntegrationTest.java
@@ -39,7 +39,8 @@ class CloudFormationFirehoseIntegrationTest {
                         "S3DestinationConfiguration": {
                           "BucketARN": "arn:aws:s3:::my-firehose-bucket",
                           "RoleARN": "arn:aws:iam::000000000000:role/firehose-role",
-                          "Prefix": "logs/"
+                          "Prefix": "logs/",
+                          "CompressionFormat": "GZIP"
                         }
                       }
                     }
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationFirehoseIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationFirehoseIntegrationTest.java
@@ -88,5 +88,6 @@ class CloudFormationFirehoseIntegrationTest {
         .then()
             .statusCode(200)
             .body(containsString(streamName));
+            .body(containsString("GZIP"))
     }
 }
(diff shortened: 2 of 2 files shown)


pgermosen (commented):
@AbarnaaSree — welcome, and thank you for this. The fix itself is correct and well targeted; the main thing I would like changed is one line in the test. Let me explain why, because it is a genuinely easy trap to fall into.

## The test does not currently exercise the fix

The test change adds `"CompressionFormat": "GZIP"` to the template that goes *in*, but the assertions afterwards only check that the stack reaches `CREATE_COMPLETE`, that the ARN resolves, and that the delivery stream exists. Nothing checks what compression format came back.

I removed your fix entirely and ran the test again:

```
Tests run: 1, Failures: 0, Errors: 0  →  BUILD SUCCESS
```

So the test passes with the fix, without the fix, and with the fix deleted. That is why *"CloudFormationFirehoseIntegrationTest — 1 passed"* cannot serve as verification here — it would have reported exactly the same before your change.

The good news is that it is a one-line fix. Adding an assertion to the existing `DescribeDeliveryStream` block:

```java
    .statusCode(200)
    .body(containsString(streamName))
    .body(containsString("GZIP"));
```

I tried exactly that locally: **with your fix it passes, and with your fix removed it fails.** That is the property a regression test needs — it has to be able to fail.

Once you have that, a good habit for future PRs is to run the new test against the unmodified code first and watch it fail. If it passes before the fix, it is not yet testing the fix. Several regular contributors here do that as a matter of course and quote the failure in the description; it is the single most convincing thing you can put in front of a reviewer.

One small wording point in the same area: the description says *"Added a regression case"*, but no new test method was added — the existing test's input template was edited. Worth saying precisely which, since a reviewer will look for the new case.

## Your fix is right, and I want to be clear about that

None of the above is a doubt about the change itself. I checked it:

**The bug is real.** `DeliveryStreamDescription` defaults `compressionFormat` to `UNCOMPRESSED` when it is null, so a stack that asked for `GZIP` silently described as uncompressed — exactly as you say.

**The fix is in the right place**, and your claim that it covers both `S3DestinationConfiguration` and `ExtendedS3DestinationConfiguration` is accurate. I verified rather than assumed: both resolve into the same `s3Node` a few lines above, so the single setter genuinely handles both. That was worth checking and it holds.

## A few housekeeping items

Nothing here affects the code, but they are worth getting into the habit of:

- **The title wants a scope.** The convention in this repo is `fix(cloudformation): …` or `fix(firehose): …` rather than a bare `fix:`. The automated check passes either way, since scope is optional in Conventional Commits, but every other change here carries one.
- **The PR template was replaced rather than filled in.** The Type of change, AWS Compatibility and checklist sections are where reviewers look for "which SDK or CLI did you check this against" and "do the tests pass locally", so filling them in saves a round trip.
- **No issue is linked.** There is a related one — #2328, *"Firehose: apply CompressionFormat when delivering to S3"* — though it is not the same thing: that one is about actually compressing the delivered data, whereas yours is about preserving the setting through CloudFormation. Worth referencing as related rather than as `Closes`, so the other issue stays open.
- **The code blocks in the description are not fenced**, so the JSON and Java render as plain text. Wrapping them in triple backticks makes it much easier to read.
- I have added the `bug`, `cloudformation` and `firehose` labels.

Your CI checks are still waiting on a maintainer to approve the workflow run — that is GitHub's gate on first-time contributors rather than anything wrong with the PR, and it will be sorted out shortly.

Add the assertion and I will be glad to approve. Thanks again for picking this up — a four-line fix in the right place is a good first contribution.

AbarnaaSree:
Thanks for the detailed review! I’ve added the requested `GZIP` assertion to `CloudFormationFirehoseIntegrationTest` so the test now verifies that the configured compression format is actually preserved in the provisioned Firehose delivery stream.

The test passes with the fix and fails when the compression-format propagation is removed.

pgermosen (approved):
The assertion is genuinely load-bearing now, not the earlier detached-semicolon version. Reverted the CompressionFormat propagation and confirmed the test fails (UNCOMPRESSED instead of GZIP), then restored and confirmed it passes. Re-verified against current main with the actual merged tree compiled and tested — clean.

Approving. Thanks for working through the feedback, and welcome to the project.

PR #2407, opened 2026-08-19
docs: fix broken star history chart

FaintFlower:
The star history chart in the README is currently broken, which makes the README look neglected and hides the project's history from visitors. This change points the chart at a working service so it renders again.

The repository owner and repository name in the chart URLs still match the origin remote, so no other updates were needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -822,11 +822,11 @@ Name in this README, a sponsor badge on GitHub, and our sincere thanks.
 ## Star History
 
 <p align="center">
-  <a href="https://www.star-history.com/?repos=floci-io%2Ffloci&type=date&legend=top-left">
+  <a href="https://star-history.dera.page/#floci-io/floci&type=date&legend=top-left">
     <picture>
-      <source media="(prefers-color-scheme: dark)" srcset="https://api.star-history.com/chart?repos=floci-io/floci&type=date&theme=dark&legend=top-left" />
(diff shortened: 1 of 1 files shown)


hectorvent:
Thank you for taking the time to send a first PR here, and you are right, the chart does render broken on the README. What seems to be going on: the endpoint itself answers fine when fetched directly ([api.star-history.com/chart?repos=floci-io/floci](https://api.star-history.com/chart?repos=floci-io/floci&type=date&legend=top-left) returns a valid SVG), but GitHub serves README images through its camo proxy, and star-history's rate limiting throttles those shared proxy IPs, so the badge fails on the page even while the service is up.

Your replacement does render reliably, but it is a different provider rather than the star-history project, and pointing every visitor's image request at an unofficial host is a trust call we would rather not make for a cosmetic chart.

**(blocking)** Given that, our preference is to drop the section for now and maybe revisit a chart later. Would you repurpose this PR to remove the Star History section entirely instead? That is the whole block under the `## Star History` heading in README.md, heading included, and a retitle to something like `docs: remove broken star history chart`. Same problem you found, simpler durable fix, and the credit stays yours.

hectorvent:
Hi @FaintFlower, 

Closing in favor of batch PRs for all Emulators. 
I appreciate your help!

PR #2413, opened 2026-08-19
fix(apigatewayv2): return stage tags across protocols

aryasaatvik:
## Summary

Return create-time API Gateway v2 stage tags from every Stage response while preserving case-sensitive tag keys. The JSON 1.1 normalization fix preserves tag-key casing for every API Gateway v2 resource handled through that path, not only stages.

No linked issue; this was found while exercising API Gateway v2 through the AWS SDK in a local integration.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Create-time stage tags were persisted but omitted from Stage responses, and JSON 1.1 request normalization lowercased case-sensitive tag keys. Stage responses now preserve the protocol-specific field shape and original tag keys across create, get, list, and update operations. Reserved `floci:` tags remain internal and are not returned.

Verified with AWS SDK for Java v2 2.52.0.

## Response coverage

| Surface | Stage responses | Tag shape |
| --- | --- | --- |
| REST JSON | CreateStage, GetStage, GetStages, UpdateStage | `tags` with original keys |
| JSON 1.1 | CreateStage, GetStage, GetStages, UpdateStage | `Tags` with original keys |
| AWS SDK for Java v2 | create, get, and list models | populated `tags()` map |

## Checklist

## Validation

- Focused REST JSON and JSON 1.1 Stage suite: 46 tests passed.
- Reserved-tag mutation check: bypassing `stripApiGatewayReservedTags` fails `ApiGatewayV2IntegrationTest#createStage`.
- AWS SDK for Java v2 management suite: 19 tests passed.
- `./mvnw test`: 11,072 tests passed, 9 skipped.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/ApiGatewayV2ManagementTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/ApiGatewayV2ManagementTest.java
@@ -40,6 +40,7 @@ import software.amazon.awssdk.services.apigatewayv2.model.NotFoundException;
 import software.amazon.awssdk.services.apigatewayv2.model.ProtocolType;
 
 import java.util.List;
+import java.util.Map;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -311,6 +312,7 @@ class ApiGatewayV2ManagementTest {
                 .stageName(stageName)
                 .deploymentId(deploymentId)
                 .autoDeploy(false)
+                .tags(Map.of("environment", "test", "owner", "platform"))
                 .build());
 
         assertThat(response.stageName()).isEqualTo(stageName);
@@ -318,6 +320,9 @@ class ApiGatewayV2ManagementTest {
         assertThat(response.autoDeploy()).isFalse();
         assertThat(response.createdDate()).isNotNull();
         assertThat(response.lastUpdatedDate()).isNotNull();
+        assertThat(response.tags())
+                .containsEntry("environment", "test")
+                .containsEntry("owner", "platform");
         stageCreated = true;
     }
 
@@ -333,6 +338,9 @@ class ApiGatewayV2ManagementTest {
 
         assertThat(getResponse.stageName()).isEqualTo(stageName);
         assertThat(getResponse.deploymentId()).isEqualTo(deploymentId);
+        assertThat(getResponse.tags())
+                .containsEntry("environment", "test")
+                .containsEntry("owner", "platform");
 
         var listResponse = apigwv2.getStages(GetStagesRequest.builder()
                 .apiId(apiId)
@@ -347,6 +355,13 @@ class ApiGatewayV2ManagementTest {
                 .first()
                 .extracting(item -> item.deploymentId())
                 .isEqualTo(deploymentId);
+        assertThat(listResponse.items())
+                .filteredOn(item -> stageName.equals(item.stageName()))
+                .first()
+                .extracting(item -> item.tags())
+                .satisfies(tags -> assertThat(tags)
+                        .containsEntry("environment", "test")
+                        .containsEntry("owner", "platform"));
     }
 
     @Test
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
@@ -2177,6 +2177,10 @@ public class ApiGatewayController {
             ObjectNode stageVariables = node.putObject("stageVariables");
             s.getStageVariables().forEach(stageVariables::put);
         }
+        if (s.getTags() != null && !s.getTags().isEmpty()) {
+            ObjectNode tags = node.putObject("tags");
+            s.getTags().forEach(tags::put);
+        }
         return node;
     }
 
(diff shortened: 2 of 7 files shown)


pgermosen (commented):
@aryasaatvik — welcome, and thank you for this. The change is correct, and the part that is easiest to overlook is the part I want to credit first.

## The casing fix is the valuable half

The title is about returning stage tags, but the more interesting fix is the one line in `toLowerCamelCase`. That method lower-cases the first character of every request key, and it was recursing into the tag map — so a user tag key of `Environment` was being stored as `environment`. AWS tag keys are case-sensitive, so that was silent corruption of user data, and it is exactly the sort of thing nobody notices until a lookup by tag key mysteriously misses.

Skipping `normalizeValue` for the `tags` entry is the right fix, and I checked it is safe rather than assuming: `normalizeValue` only re-cases keys inside maps and lists and never changes types, so the `Map<String, String>` the service casts to is unaffected. Because the check runs after the first character has already been lowered, it catches both `Tags` and `tags` on the way in.

Worth knowing that this fix is broader than the title suggests — it corrects tag-key casing for every v2 resource that passes through that handler, not only stages. That is a good thing, and worth a sentence in the description so the next person understands why the change is where it is.

I also confirmed the change to `ApiGatewayController` is not scope creep, since a v1-package file in a v2 PR is a fair thing to query: the method is `toV2StageNode`, the v2 serialiser, which lives in that class because it serves both REST-JSON surfaces.

Reverting the casing fix fails `json11StageResponsesReturnTags`, so it is genuinely held in place.

## One gap: the reserved-tag stripping is untested

`createStage` calls `ReservedTags.stripApiGatewayReservedTags(tags)`, which is the right thing to do — it matches what the codebase already does for v2 APIs at `ApiGatewayV2Service:115`, and for v1 APIs and usage plans in `ApiGatewayService`. Reusing the existing helper rather than writing a new one is the correct instinct.

But I removed that call locally and the whole suite still passed. Nothing asserts that a reserved tag is stripped from a stage, so a later refactor could drop it without anything failing, and reserved tags would start leaking back into `GetStage` responses.

One case would close it: create a stage with a reserved tag alongside an ordinary one, then assert the response carries the ordinary tag and not the reserved one.

A habit worth having generally, and it is what turned this up: after adding a line that enforces something, delete it locally and re-run. If nothing fails, it is not yet covered. It takes a minute and it is the difference between a test that documents behaviour and one that defends it.

## Two housekeeping notes

**The pull request template was replaced rather than filled in.** The Type of change, AWS Compatibility and checklist sections are where reviewers look for which SDK or CLI exercised the wire shape, and whether the tests pass locally. Your Validation section covers the last of those well — it is mostly a matter of putting it where people look.

**No issue is linked.** If this came from something you hit in practice rather than an existing report, that is worth a sentence in the summary; if there is an issue, `Closes #N` will link and close it automatically.

Your CI is still waiting on a maintainer to approve the workflow run — that is GitHub's gate on first-time contributors rather than anything wrong here, and it will be sorted shortly.

Add the reserved-tag case and this looks good to me.

pgermosen (approved):
Approving. The gap is closed, and I checked it rather than taking the description's word for it.

**The reserved-tag case does what it needs to.** Replacing `stripApiGatewayReservedTags(tags)` with a plain `setTags(tags)` fails `ApiGatewayV2IntegrationTest.createStage:256` — the same result you reported. Asserting both halves in one case, that `owner` survives and that `floci:internal` is absent, is the right shape: a test that only checked the reserved key was gone would still pass if stripping removed everything.

You ran the mutation check yourself and put the result in the description. That is the habit worth keeping, and it is what makes the rest of the description credible rather than merely thorough.

**The casing fix is still held.** Reverting the `"tags".equals(key)` skip fails `ApiGatewayV2JsonHandlerTest.json11StageResponsesReturnTags:127`.

**The merge needed real checking, and it passes.** This branch is now five commits behind `main`, and it auto-merges three files that #2011 also rewrote — `ApiGatewayController`, `ApiGatewayV2JsonHandler` and `ApiGatewayV2Service`. A clean auto-merge across two changes editing the same serialisers is exactly where a semantic conflict hides, so I merged onto current `main` and ran the v2 suites rather than trusting the mergeability flag. Both features coexist correctly: `AuthorizationScopes` from #2011 and stage `Tags` from this PR sit side by side in the JSON 1.1 handler, and the suites are green.

The template and the issue question are both sorted, and the sentence about the casing fix being broader than the title made it in. Thank you for that — it is the part a future reader would otherwise have to rediscover.

---

### One nit, not blocking

`Stage.setTags` assigns the map straight through:

```java
public void setTags(Map<String, String> tags) { this.tags = tags; }
```

`DocDbCluster.setTags`, which landed in `main` today via #2408, normalises instead:

```java
public void setTags(Map<String, String> tags) {
    this.tags = tags == null ? new LinkedHashMap<>() : new LinkedHashMap<>(tags);
}
```

It does not bite here, and I checked why rather than assuming: `Stage` is persisted to `apigatewayv2-stages.json`, a record written before this PR deserialises with no `tags` key and keeps the field initialiser, and both serialisers null-guard before reading. So this is consistency with a pattern that arrived a few hours after you opened this, not a defect — worth knowing about for whenever you next touch the model.

Nice first contribution. The casing fix in particular was the half that mattered most and the half the title undersold.

hectorvent:
Seconding pgermosen's review rather than restating it, they covered this exactly right, including the part where the casing fix was the half that mattered most. I verified the wire shape against the service model independently and it holds.

**(follow-up, separate PR)** Picking up pgermosen's nit as a tracked item: `Stage.setTags` assigns the map straight through, while the pattern that just landed in #2408 (`DocDbCluster.setTags`) defensively copies into a `LinkedHashMap`. Verified harmless today, so no action here, but a small `chore(apigatewayv2)` aligning the v2 model tag setters would be welcome whenever you or anyone next touches these models.

Welcome aboard, and merging.

PR #2414, opened 2026-08-19
fix(lambda): resolve ARN and function URL invocations in owning account

aryasaatvik:
## Summary

Resolve Lambda invocations against the account encoded in full function ARNs and Function URLs, including versions and aliases.

## Routing

| Invocation source | Account-aware behavior |
| --- | --- |
| Full function ARN | Resolve the function, version, or alias in the ARN's account |
| Function URL | Scope URL identity and request context to the owning account |
| EventBridge and Scheduler | Preserve the full target ARN through asynchronous delivery |

Existing Function URL response and streaming handling remains unchanged.

## Validation

- Focused Lambda, EventBridge, Scheduler, and permission suite: 61 tests passed.
- AWS SDK for Java v2 account-routing suite: 1 test passed.
- `./mvnw test`: 11,078 tests passed, 9 skipped.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/LambdaAccountRoutingTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/LambdaAccountRoutingTest.java
@@ -1,0 +1,64 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.core.SdkBytes;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.lambda.LambdaClient;
+import software.amazon.awssdk.services.lambda.model.CreateFunctionRequest;
+import software.amazon.awssdk.services.lambda.model.DeleteFunctionRequest;
+import software.amazon.awssdk.services.lambda.model.FunctionCode;
+import software.amazon.awssdk.services.lambda.model.InvocationType;
+import software.amazon.awssdk.services.lambda.model.InvokeRequest;
+import software.amazon.awssdk.services.lambda.model.Runtime;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class LambdaAccountRoutingTest {
+
+    private static final String CALLER_ACCOUNT = "234567890123";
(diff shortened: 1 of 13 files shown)


pgermosen (approved):
@aryasaatvik — approving. This is a well-shaped change and the coverage is genuinely thorough.

## The design choice that matters

Cross-account routing is the sort of change that easily breaks the common case while fixing the uncommon one, so that is what I looked at first. It does not: `getForAccount` sits alongside the existing `get` rather than replacing it, and `resolveInvokeTargetForAccount` alongside `resolveInvokeTarget`. Nothing on the single-account path was rewired, which is why 959 tests across lambda, eventbridge and scheduler pass untouched.

Reusing `getForAccountMigratingLegacy` and `accountOrDefault` rather than writing new equivalents is the right instinct too — both already existed, and the account-migration semantics in particular are not something worth reimplementing beside the original.

## The subtractive changes are an improvement, not just a tidy-up

Replacing this in EventBridge and Scheduler:

```java
String fnName = arn.substring(arn.lastIndexOf(':') + 1);
lambdaService.invoke(region, fnName, ...);
```

with `invokeArn(arn, ...)` fixes more than the account. For a qualified ARN — `…:function:name:alias` — that `lastIndexOf(':')` returns the **alias**, not the function name, so the old path mis-resolved qualified targets regardless of account. Passing the whole ARN through removes the class of bug rather than one instance of it.

## Both halves are pinned

I reverted each rather than reading the tests and agreeing:

| Reverted | Result |
| --- | --- |
| `invokeArn` ignoring the ARN's account | four tests fail |
| EventBridge re-deriving name and region | `invokeTarget_lambdaTargetPreservesArnAccount` fails |

`invokeArnDoesNotMigrateForeignLegacyFunctionOrAlias` deserves a specific mention. The legacy-migration helper adopts an unprefixed record into the caller's account on read, so a foreign-account ARN must not trigger it — otherwise one account's ARN quietly claims another's legacy record, and the damage persists because the migration writes. Thinking of that unprompted, and pinning it, is the part of this PR I would least have expected and most wanted to see.

## One note, not blocking

The pull request template was replaced rather than filled in — the Type of change and checklist sections are missing, and no issue is linked. Your Validation section already contains the substance those ask for; it is only a question of putting it where reviewers and future readers look for it. The same applies to #2413, so it is one habit rather than two separate things.

I have added the `bug` and `lambda` labels.

Thank you for this — two well-tested PRs in a day is a strong start.

PR #2417, opened 2026-08-20
fix: apply S3 Vectors metadata filters

### LINKED ISSUE #2160: [BUG] S3 Vectors: QueryVectors accepts a metadata filter and silently ignores it
## Service

S3 Vectors

## AWS API Action

`QueryVectors`

## Expected behavior

Real AWS S3 Vectors applies the `filter` argument to vector metadata and returns only the vectors that satisfy it. A filter that matches nothing returns an empty result set.

## Actual behavior

The `filter` argument is accepted and silently ignored — every vector in the index is returned regardless of the filter.

This is a quiet failure rather than an error, which makes it particularly easy to miss: an application that relies on metadata filtering for scoping (for example, restricting retrieval to one tenant, user, or job) appears to work, but silently reads across every record in the index.

Looking at the source, the request record does declare the field:

```java
// S3VectorsController.java — QueryVectorsRequest
public record QueryVectorsRequest(
        ...
        Object filter,
        ...
) {}
```

but the service method it delegates to has no filter parameter at all, so the value is dropped at the controller boundary:

```java
// S3VectorsService.java
public List<QueryResult> queryVectors(String bucketName, String indexName,
                                      List<Float> queryVector, int topK, String region)
```

## Reproduction

```bash
E=http://localhost:4566

  --vectors '[{"key":"v1","data":{"float32":[0.1,0.2,0.3,0.4]},"metadata":{"tenant":"a"}},
              {"key":"v2","data":{"float32":[0.9,0.1,0.0,0.2]},"metadata":{"tenant":"b"}}]'

# Filter matches nothing -> expected 0 vectors, actual 2
aws --endpoint-url $E s3vectors query-vectors \
  --query-vector '{"float32":[0.1,0.2,0.3,0.4]}' \
  --top-k 5 --return-metadata \
  --filter '{"tenant":{"$eq":"NO_SUCH_TENANT"}}'

# Same result with the $and form
aws --endpoint-url $E s3vectors query-vectors \
  --query-vector '{"float32":[0.1,0.2,0.3,0.4]}' \
  --top-k 5 --return-metadata \
  --filter '{"$and":[{"tenant":{"$eq":"NO_SUCH_TENANT"}}]}'
```

Both queries return `v1` and `v2`.

## Environment

- Client: AWS CLI 2.36.16 (also reproduces via boto3)

---

Happy to put up a PR threading the filter into `queryVectors` and evaluating it against the vector metadata before ranking, if that would be welcome — just let me know whether you'd want the full documented operator set (`$eq`, `$ne`, `$gt`, `$gte`, `$lt`, `$lte`, `$in`, `$nin`, `$exists`, `$and`, `$or`) in one go or a narrower first cut.

mvanhorn:
## Summary

Add a package-local production evaluator for the AWS S3 Vectors metadata-filter grammar, including comparison, membership, existence, and recursive logical operators; it will be called by the service rather than existing only as a test seam. Thread the controller's parsed filter value into `S3VectorsService.queryVectors`, and discard nonmatching vectors before calculating distance, sorting, and applying `topK` so filtered-out nearest neighbors cannot consume the result limit. Define field clauses as an implicit conjunction, compare numeric metadata by numeric value, require ordered comparisons to use compatible scalar types, and give missing fields explicit `$exists`/negative-predicate behavior rather than conflating them with present null values.

`QueryVectorsRequest` already accepts a `filter`, but `S3VectorsController` drops it when calling `S3VectorsService.queryVectors`, so filtered requests silently search every vector in the index. This differs from AWS and can leak results across metadata-defined scopes such as tenants because callers receive a successful, plausible-looking response. The issue reproduces through both AWS CLI and boto3 on Floci 1.6.0, including a filter that should match no vectors. There are no assignees, maintainer claims, competing PRs, or closed prior attempts in the supplied issue evidence.

Closes #2160

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/S3VectorsTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/S3VectorsTest.java
@@ -128,12 +128,16 @@ class S3VectorsTest {
                                 PutInputVector.builder()
                                         .key("v1")
                                         .data(VectorData.builder().float32(1.0f, 0.0f, 0.0f).build())
-                                        .metadata(Document.fromMap(Map.of("label", Document.fromString("first"))))
+                                        .metadata(Document.fromMap(Map.of(
+                                                "label", Document.fromString("first"),
+                                                "tenant", Document.fromString("tenant-a"))))
                                         .build(),
                                 PutInputVector.builder()
                                         .key("v2")
                                         .data(VectorData.builder().float32(0.0f, 1.0f, 0.0f).build())
-                                        .metadata(Document.fromMap(Map.of("label", Document.fromString("second"))))
+                                        .metadata(Document.fromMap(Map.of(
+                                                "label", Document.fromString("second"),
+                                                "tenant", Document.fromString("tenant-b"))))
                                         .build()
                         )
                         .build()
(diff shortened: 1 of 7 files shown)


hectorvent:
Thank you for this, and welcome. The tenant-leak framing in the description is exactly why this mattered, and the fix is verified against AWS's own grammar: I checked every operator in the [metadata filtering guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-vectors-metadata-filtering.html) and your evaluator matches the table exactly, including the array-metadata `$eq` semantics and filters not consuming `topK` slots.

On the bot's P1 above: I believe it has this one backwards. The guide's operator table lists `$gt`/`$gte`/`$lt`/`$lte` with valid input type Number only, so rejecting string operands with a `ValidationException` is what real AWS does, and no change is needed. Please do not "fix" that.

One optional thought, no action needed for merge: the docs section could state what `$ne`/`$nin` do for vectors missing the key (they match), since AWS's guide is silent there and your choice is worth writing down.

No blockers from my side.

hectorvent (approved):

PR #2425, opened 2026-08-20
fix(docdb): give a record without an ARN the one it should have had

### LINKED ISSUE #2424: [BUG] docdb: a record persisted without an ARN still resolves a tag call by identifier alone
## Service

DocumentDB

## API Action / Feature

`ListTagsForResource`, `AddTagsToResource`, `RemoveTagsFromResource` — for cluster and instance records persisted before #2408

## What happens

#2408 made a tag ARN resolve to the record it names rather than to any record of that identifier, so an ARN naming another region's resource is refused. The comparison is skipped when the stored ARN is null:

```java
private static void requireArnNamesRecord(String requested, String stored,
                                          String errorCode, String message) {
    if (stored != null && !stored.equalsIgnoreCase(requested)) {
        throw new AwsException(errorCode, message, 404);
    }
}
```

`DocDbCluster.dbClusterArn` defaults to null and `getDbCluster` does no backfill, so a cluster written by an earlier Floci reaches `resolveTagTarget` without an ARN, the comparison is skipped, and it resolves by identifier alone — the hole #2408 closed, still open for pre-upgrade records.

Reachable on the `docdb` credential scope only. The `rds` scope routes by matching the whole stored ARN, so a record with no ARN never matches another region's ARN there and the request goes to RDS, which rejects it.

`TaggedRecordUpgradeTest`'s legacy fixture is already exactly this shape:

```json
{"dbClusterIdentifier":"legacy","status":"available"}
```

## Fix

Backfill on read, which is the pattern #2408 already established on the RDS side — `getDbParameterGroup` fills a missing region and ARN and persists them, because a group that never gets one cannot be tagged at all. The same applies to DocumentDB clusters and instances: give a record its ARN from the region it is found under, then the null branch of `requireArnNamesRecord` can go away rather than remain as a standing allowance.

Records persisted before ARNs carried the caller's region read as the configured default, which is the region they were in fact created under — the same assumption `hasCluster(id, region)` already makes.

## Also worth doing here

`DocDbService.writeLocks` gains one entry per distinct identifier and is never pruned when a record is deleted. Trivial at emulator scale, but the delete paths are the same ones this change touches.

## Related

#2416 re-keys these stores by region and will touch the same read paths. Either order works; doing this one first keeps the backfill separate from the re-key, which is easier to review.

ck-michals:
Closes #2424.

## What

#2408 made a tag call resolve to the record its ARN names rather than to any record of that identifier. The comparison was skipped when the stored ARN was null, and `DocDbCluster.dbClusterArn` defaults to null with no backfill on read — so a cluster or instance written by an earlier Floci resolved by identifier alone. The hole that check closed, still open for exactly the records that predate it.

Reachable on the `docdb` credential scope, which addresses this service directly. The `rds` scope routes by matching the whole stored ARN, so a record without one never matches another region's ARN there and the request goes to RDS, which rejects it.

## Change

The ARN is filled in on read, which is the pattern already on the RDS side — `getDbParameterGroup` fills a missing region and ARN because a group that never gets one cannot be tagged at all.

It comes from the **default region, not the caller's**. Such a record was created when every DocumentDB ARN was built from the default, so that is the region it is in; taking the reader's region would let a caller elsewhere claim it, which is the original bug wearing a fix's clothes. It is also the assumption `hasCluster(id, region)` already makes when a record has no ARN.

With every record carrying an ARN by the time it is compared, the `null` allowance in `requireArnNamesRecord` goes away rather than staying as a standing exception.

Also, as noted on #2408: a deleted record's monitor is dropped. What stops a tag from reviving a deleted record is the re-resolve inside the lock rather than the identity of the monitor, so retiring one with its record is safe — `DocDbTagConcurrencyTest` still passes across repeated runs.

## Tests

`DocDbLegacyRecordArnIntegrationTest` seeds a record in exactly the shape `TaggedRecordUpgradeTest`'s legacy fixture describes, then:

- another region's ARN for that identifier is **not found**, for both the read and the write, and the record keeps its own tags
- the record's own ARN tags it normally — the backfill has to make the record usable, not merely refuse the wrong ARN
- the ARN it gains is the default region's

Both halves are break-tested, and the second break is the one worth naming: making the backfill use the caller's region instead of the default turns the foreign-region call back into a `200`. The region choice is load-bearing, not incidental.

616 tests green, on a branch cut from current `main` (`5744082c`).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/docdb/DocDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/docdb/DocDbService.java
@@ -121,9 +121,27 @@ public class DocDbService {
     }
 
     public DocDbCluster getDbCluster(String id) {
-        return clusters.get(id).orElseThrow(() ->
+        DocDbCluster cluster = clusters.get(id).orElseThrow(() ->
                 new AwsException("DBClusterNotFoundFault",
                         "DocDB cluster " + id + " not found.", 404));
+        if (cluster.getDbClusterArn() == null || cluster.getDbClusterArn().isBlank()) {
+            cluster.setDbClusterArn(legacyArn("cluster:" + id));
+            clusters.put(id, cluster);
+        }
+        return cluster;
+    }
+
+    /**
+     * The ARN a record written before ARNs were stored should have had.
+     *
+     * <p>The default region rather than the caller's: such a record was created when every ARN was
+     * built from the default, so that is the region it is in — and giving it the region of whoever
+     * happens to read it first would let a caller elsewhere claim it. Without an ARN a record
+     * cannot be told apart from another region's of the same identifier, and cannot be tagged
+     * through an ARN at all.
+     */
+    private String legacyArn(String resource) {
+        return regionResolver.buildArn("rds", regionResolver.getDefaultRegion(), resource);
     }
 
     public boolean hasCluster(String id) {
@@ -148,7 +166,9 @@ public class DocDbService {
     /** Refuses a record whose own ARN is not the one that was asked for. */
     private static void requireArnNamesRecord(String requested, String stored,
                                               String errorCode, String message) {
-        if (stored != null && !stored.equalsIgnoreCase(requested)) {
+        // No null allowance: a record read through getDbCluster or getDbInstance has been given
+        // its ARN if it lacked one, so anything reaching here can be compared.
+        if (!requested.equalsIgnoreCase(stored)) {
             throw new AwsException(errorCode, message, 404);
         }
     }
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Verified the region-choice claim empirically: swapping `getDefaultRegion()` → `getRegion()` in `legacyArn()` makes `aLegacyRecordIsNotTaggedThroughAnotherRegionsArn` fail exactly as predicted (404→200), confirming it's genuinely load-bearing, not incidental. Full `docdb` suite (62 tests) passes clean both on the branch and on the actual merged-with-main tree. The `writeLocks.remove()` cleanup is covered by the existing `DocDbTagConcurrencyTest` delete-vs-tag race test, re-run 3x clean. LGTM.

PR #2426, opened 2026-08-20
fix(iam): correct GetAccountSummary provider and access-key counts

### LINKED ISSUE #1938: [FEAT] [Feature] IAM: Add GetAccountSummary and ListEntitiesForPolicy support

abanna:
## Summary

Closes #1938.

Completes the remaining `GetAccountSummary` compatibility work after the IAM action handlers were added.

- Aligns `AttachedPoliciesPerRoleQuota` with the current AWS IAM default of 20 following the August 19, 2026 quota increase.
- Reports the number of stored OpenID Connect providers through `Providers`.
- Keeps `AccountAccessKeysPresent` scoped to root-account access keys; IAM user access keys do not set that signal.
- Adds unit, Query-protocol integration, and Java AWS SDK compatibility coverage using SDK `SummaryKeyType` keys.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, provider state was always reported as zero. An intermediate implementation also treated IAM user keys as the root-account access-key signal. The current implementation matches the AWS field meanings and the current managed-policies-per-role default quota.

References:

- [GetAccountSummary API](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetAccountSummary.html)
- [IAM quotas](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html)
- [August 19, 2026 IAM quota increase](https://aws.amazon.com/about-aws/whats-new/2026/08/aws-iam-quota-increase/)

## Testing

- `IamServiceTest` — 93 tests passed on exact head `0699243f50af11645109a726d1fecaed806aee3d`
- `IamGetAccountSummaryIntegrationTest` — 4 tests passed on the same head
- Mutation proof — restoring the implementation value to 10 makes both exact quota regressions fail (`expected: 20`, `actual: 10`); restoring 20 makes both pass
- Java AWS SDK compatibility project `test-compile` — successful across 152 test sources
- `./mvnw clean package -DskipTests`
- Synthetic merge check against current `upstream/main`
- `git diff --check`

## Checklist

- [x] Relevant focused and compatibility tests pass locally


pgermosen:
Thanks for this — the `Providers` fix and the `AccountAccessKeysPresent` root-account-scoping comment both check out, and I confirmed the `Providers` counter is genuinely exercised (reverting it back to the hardcoded `0` fails `accountSummaryUsesAwsDefaultQuotasAndTracksProviders` with exactly `expected: <1> but was: <0>`).

One accuracy gap I'd like flagged before this merges, since it's squarely in the scope of what this PR is doing:

## `AttachedPoliciesPerRoleQuota` should be 20, not 10

`IamService.java:826` reports:
```java
summary.put("AttachedPoliciesPerRoleQuota", 10L);
```

Per the [IAM and AWS STS quotas](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html) page — the same reference this PR's description already cites — the default quota table lists:

| Resource | Default quota |
|---|---|
| Managed policies per user | 10 |
| Managed policies per group | 10 |
| Managed policies per role | **20** |

So `AttachedPoliciesPerUserQuota` and `AttachedPoliciesPerGroupQuota` (both correctly `10L` in the code) are right, but the role variant should be `20L`, not `10L`. It looks like an easy copy-paste slip when all three were originally added together.

This line predates this PR (it traces back to #1986, `feat(iam): add GetAccountSummary support`), so it's not something this PR introduced. But this PR's own new tests lock the wrong value in two more places, which will make a future fix harder to land cleanly:
- `IamServiceTest.java` — `accountSummaryUsesAwsDefaultQuotasAndTracksProviders` asserts `assertEquals(10L, empty.get("AttachedPoliciesPerRoleQuota"))`
- `IamGetAccountSummaryIntegrationTest.java` — asserts `.find { it.key == 'AttachedPoliciesPerRoleQuota' }.value` equals `"10"`

Since this PR is already touching `GetAccountSummary` for quota accuracy and adding regression coverage for these exact fields, this would be a natural place to fix it alongside — update the constant to `20L` and the two new assertions to match. Happy to re-review once that's in. Everything else here looks solid — no other gaps found.

abanna:
Thanks for flagging this. I rechecked the current AWS IAM quotas reference and the `GetAccountSummary` example. AWS lists managed policies per role as a default of **10** and a maximum of **25**; **20** is the maximum for managed policies per user. The AWS SDK example also reports `AttachedPoliciesPerRoleQuota` as `10`.

So the current implementation and assertions are intentional, and I’m keeping them at `10`:

- https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html
- https://docs.aws.amazon.com/boto3/latest/reference/services/iam/client/get_account_summary.html

Your review did uncover a nearby stale Javadoc sentence that still described OIDC providers as untracked. I corrected that in `f922fbe6`. Could you please take another look once the fresh checks finish?

abanna:
@pgermosen You were right, and I was working from the stale SDK example when I replied earlier. AWS raised the default managed-policies-per-role quota from 10 to 20 on August 19, and the current IAM quota table now reflects it.

Fixed in `0699243f` with service, Query-protocol, and Java SDK assertions. The focused IAM suite is 97/97 green, and reverting the implementation to 10 makes both quota regressions fail.

Thanks for catching this, and apologies for correcting you in the wrong direction. Could you take another look? The `waiting-contributor` label is also stale now that the update is pushed; could a maintainer clear it?

pgermosen (approved):
Thanks for closing out the `GetAccountSummary` work — and for citing sources rather than just asserting numbers, which let me actually check them rather than take them on faith.

I verified the quota claim that stood out most — the "AWS quota increase" bump from 10 to 20 for managed policies per role — against the live IAM quotas reference page (not the linked what's-new post) and it's accurate: current default is 20, max 25. The rest of the quota values this PR adds regression coverage for (`GroupsQuota`=300, `RolesQuota`=1000, `PoliciesQuota`=1500, `InstanceProfilesQuota`=1000, `PolicySizeQuota`=6144) all match that same table too — those were already correct in the code, so this is locking in values that had no test coverage before.

The two functional changes:

- **`AttachedPoliciesPerRoleQuota` 10 → 20** — correct per the current AWS default, mutation-tested.
- **`Providers` now counts real OIDC providers instead of a hardcoded 0** — correct. I considered whether AWS's `Providers` field might count SAML providers too (the docs don't spell out the field's scope either way), but it's moot here: Floci has no `CreateSAMLProvider` at all — `ListSAMLProviders` is a permanent empty stub — so there's nothing of that kind for the count to ever miss.
- **`AccountAccessKeysPresent` staying `0` when only IAM user keys exist** — correct; this field reports the *root* account's access keys, which Floci doesn't model, and you added a test (`userAccessKeysDoNotSetRootAccountAccessKeySignal`) that specifically pins IAM user keys as not tripping it.

Test coverage is solid across all three layers (unit, Query-protocol integration, real SDK compat), CI is fully green, and the merge is clean.

Recommendation: **approve**.

Issue #2200, opened 2026-08-10
[BUG] RDS, Firehose and Cognito report drift on a second terraform plan


pgermosen:
## What happened

Three resources in the Terraform compatibility fixture apply cleanly but then report drift on the very next `terraform plan`, so the configuration never converges:

```
Plan: 0 to add, 3 to change, 0 to destroy.

  ~ aws_db_instance.app
      ~ auto_minor_version_upgrade = false -> true

  ~ aws_kinesis_firehose_delivery_stream.events
      ~ extended_s3_configuration {
          + s3_backup_mode = "Disabled"
        }

  ~ aws_cognito_user_pool.pool
      (user_pool_add_ons / schema blocks re-report)
```

In each case the value Terraform reads back does not match what it configured — either a field it set is not echoed, or a field it never set is invented by the read path.

## Steps to reproduce

1. Start Floci.
2. From `compatibility-tests/compat-terraform/`, `terraform apply` the existing fixture.
3. `terraform plan -detailed-exitcode` — exits `2` (changes present) rather than `0`.

## Expected behavior

A second plan over unchanged configuration reports no changes. Applying twice should be a no-op, as it is against real AWS.

## Why this went unnoticed

The compatibility suite asserted only that `terraform apply` succeeded, and this class of bug is invisible to that: the apply always succeeds, and the drift only shows on the *next* plan. #2159 added a "re-planning reports no changes" assertion for exactly this reason, but scoped it to the resources that PR introduced, because widening it to the whole configuration surfaces these three pre-existing cases.

`compatibility-tests/compat-terraform/test/terraform.bats` carries a comment naming these three resources and saying to widen the assertion once they are fixed. **Fixing them should also widen that assertion to the full configuration**, which then guards every resource in the fixture against this class of bug rather than a hand-picked subset.

## Environment

- Floci `main` (reproduced at `1795e131` and later)
- Terraform 1.15.8, `hashicorp/aws ~> 6.0`

PR #2427, opened 2026-08-20
feat(lakeformation): lake formation

namanONcode:
## Summary

This PR adds the **AWS Lake Formation** emulator to Floci. 
It implements the JSON 1.1 `AWSLakeFormation.*` API surface, which acts as the permission and governance layer over the Glue Data Catalog. This enables Terraform and other AWS SDK clients to execute data-lake deployments locally without failing on LF grants and registrations.

Features included:
- Data Lake Settings (Put/Get)
- Resource Registration (Register, Deregister, Describe, List)
- Permission Grants (Grant, Revoke, List)
- LF-Tags management (Create, Get, Update, Delete, List, AddToResource, RemoveFromResource)

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- Implemented the standard JSON 1.1 protocol using the `x-amz-target` header (e.g., `AWSLakeFormation.RegisterResource`).
- Permissions retrieval via `ListPermissions` currently returns all explicitly granted permissions to satisfy Terraform state expectations.
- Resources and tags are properly isolated per-account using Floci's `AccountAwareStorageBackend`.

## Checklist

- [x] New or updated integration test added (via `LakeFormationServiceTest.java`)


pgermosen:
Thanks for the Lake Formation emulator, this is a solid amount of API surface! A few things worth addressing before merge:

1. **PR title** doesn't match Conventional Commits, which is CI-enforced: `(feat)lakeformation: lake formation` should be `feat(lakeformation): lake formation` (scope goes right after the type, not before it).

2. **Test coverage gap**: `LakeFormationServiceTest.java` is a Mockito-mocked unit test of the service layer, but there's no test exercising the actual JSON-1.1 wire protocol (`LakeFormationJsonHandler`, `x-amz-target` routing, request parsing, error-response shape). The other `AwsJson11Controller`-based services in the repo (textract, wafv2, cloudmap) each have a `*IntegrationTest.java` that posts real `application/x-amz-json-1.1` requests via RestAssured — could you add an equivalent `LakeFormationIntegrationTest`?

3. **Style nit**: `getResourceKey` in `MemoryLakeFormationStorage.java` has several brace-less `if` bodies, which AGENTS.md calls out explicitly ("Always use braces in conditionals").

Also flagging: there are currently real merge conflicts against `main` in both `src/main/resources/application.yml` and `src/test/resources/application.yml` — looks like just config-list churn (other services landed at the same insertion point), so should be a quick rebase, not a code issue.

On the Greptile bot findings from earlier commits: most look already resolved by your follow-up fixes (permission merging, tag-value-subset removal, wildcard-exclusion keys, data-cells-filter catalog scoping). The "catalog scope discarded" one on resource registration looks like a false positive — I checked the real Lake Formation API model and `RegisterResource`/`DescribeResource`/`DeregisterResource`/`ListResources` genuinely don't have a `CatalogId` field in AWS's actual API, so no fix needed there.

Nice work overall — the shape of this matches the codebase's conventions well.

namanONcode:
> Thanks for the Lake Formation emulator, this is a solid amount of API surface! A few things worth addressing before merge:
> 
>     1. **PR title** doesn't match Conventional Commits, which is CI-enforced: `(feat)lakeformation: lake formation` should be `feat(lakeformation): lake formation` (scope goes right after the type, not before it).
> 
>     2. **Test coverage gap**: `LakeFormationServiceTest.java` is a Mockito-mocked unit test of the service layer, but there's no test exercising the actual JSON-1.1 wire protocol (`LakeFormationJsonHandler`, `x-amz-target` routing, request parsing, error-response shape). The other `AwsJson11Controller`-based services in the repo (textract, wafv2, cloudmap) each have a `*IntegrationTest.java` that posts real `application/x-amz-json-1.1` requests via RestAssured — could you add an equivalent `LakeFormationIntegrationTest`?
> 
>     3. **Style nit**: `getResourceKey` in `MemoryLakeFormationStorage.java` has several brace-less `if` bodies, which AGENTS.md calls out explicitly ("Always use braces in conditionals").
> 
> 
> Also flagging: there are currently real merge conflicts against `main` in both `src/main/resources/application.yml` and `src/test/resources/application.yml` — looks like just config-list churn (other services landed at the same insertion point), so should be a quick rebase, not a code issue.
> 
> On the Greptile bot findings from earlier commits: most look already resolved by your follow-up fixes (permission merging, tag-value-subset removal, wildcard-exclusion keys, data-cells-filter catalog scoping). The "catalog scope discarded" one on resource registration looks like a false positive — I checked the real Lake Formation API model and `RegisterResource`/`DescribeResource`/`DeregisterResource`/`ListResources` genuinely don't have a `CatalogId` field in AWS's actual API, so no fix needed there.
> 
> Nice work overall — the shape of this matches the codebase's conventions well.

@pgermosen @hectorvent all mentioned issue resolved ci is green

pgermosen:
I have to correct my last review before anything else, because it sent you the wrong way.

On Aug 21 I accepted the PR's stated premise — "JSON 1.1 protocol using the `x-amz-target` header" — and
then asked you to *add an integration test for that routing*, pointing at textract/wafv2/cloudmap as
templates. You did exactly that, and `LakeFormationIntegrationTest` is a good test. It is also testing a
protocol Lake Formation does not use. That's my error, not yours: I reviewed the code against itself
instead of against the service model, and my request pushed more work onto the wrong foundation. Sorry.

## Lake Formation is `rest-json`, not JSON 1.1 with `X-Amz-Target`

From the model shipped with the current AWS CLI (v2.25.10, `lakeformation/2017-03-31/service-2.json`):

```
"protocol": "rest-json", "protocols": ["rest-json"], "jsonVersion": "1.1"
no "targetPrefix" key at all
RegisterParameter → POST /RegisterResource
GrantPermissions  → POST /GrantPermissions
```

`jsonVersion: 1.1` is the trap here — it describes the *body* encoding, not the protocol. The `protocol`
field is `rest-json`, every one of the 60 operations carries its own `requestUri`, and there is no
`targetPrefix`, which is what a target-dispatched service would have.

I captured what a real client sends rather than inferring it —
`aws lakeformation register-resource --resource-arn ... --debug`:

```
POST /RegisterResource
Content-Type: application/json
Authorization: AWS4-HMAC-SHA256 Credential=.../lakeformation/aws4_request,
               SignedHeaders=content-type;host;x-amz-date
{"ResourceArn": "arn:aws:s3:::probe-bucket"}
```

No `X-Amz-Target` header — it isn't even in `SignedHeaders`. Content type is plain `application/json`,
not `application/x-amz-json-1.1`. So all sixteen operations are unreachable from the CLI, every AWS SDK,
and the Terraform provider.

Driving that exact request shape at a running Floci built from this branch:

| request | result |
|---|---|
| `POST /RegisterResource`, `application/json`, no target *(what the CLI sends)* | **400**, and the body is `<Error><Code>InvalidArgument</Code><Message>POST on bucket requires ?delete parameter.</Message></Error>` |
| `POST /GrantPermissions`, same shape | same S3 error |
| `POST /` + `X-Amz-Target: AWSLakeFormation.RegisterResource` *(what this PR implements)* | 200 `{}` |

Worth noting what the first row actually is: with no literal route registered for `/RegisterResource`,
the request falls through to S3's `@Path("/{bucket}")` and comes back as an **S3 XML error document**. A
Lake Formation client gets a wrong-service error it can't parse, which is harder to diagnose than a 404.

**The good news is that this is a transport-layer problem, not a rewrite.** `LakeFormationService`,
`MemoryLakeFormationStorage` and the ~50 model classes are protocol-agnostic and stay as they are, and
`LakeFormationJsonHandler.handle(action, request, region)` keeps its shape — only the thing that
*calls* it changes. Concretely:

- Add a `LakeFormationController` with literal `@Path("/RegisterResource")`, `@Path("/GrantPermissions")`
  … methods, `@Consumes(MediaType.APPLICATION_JSON)`, delegating to the existing handler with the
  operation name.
- Switch the catalog descriptor from `ServiceProtocol.JSON` / `Set.of("AWSLakeFormation.")` to
  `ServiceProtocol.REST_JSON` with the controller in `resourceClasses`, the way `guardduty` and `efs`
  do.
- Revert the `AwsJson11Controller` wiring.

On the S3 collision: it resolves itself. JAX-RS prefers a literal path segment over a template, so
`@Path("/RegisterResource")` wins over S3's `@Path("/{bucket}")` — that's exactly how FIS's
`@Path("/experimentTemplates")` coexists with S3 on `main` today. FIS (#2435) is the closest template in
the tree for a single-segment rest-json service.

`docs/services/index.md:51` will need the same correction — it currently advertises
`POST /` + `X-Amz-Target: AWSLakeFormation.*` | JSON 1.1.

## Behaviour findings, independent of the protocol

These are all reachable through the PR's own working route, so they hold either way. Probed against a
running instance:

**1. No region scoping — state leaks across regions.** Every storage key is the ARN, catalog ID or
permission key; none carries a region. Lake Formation is a regional service.

```
RegisterResource(arn:aws:s3:::region-probe-bucket)  signed for us-east-1  → 200
ListResources                                        signed for eu-west-1  → returns it
DescribeResource                                     signed for eu-west-1  → returns it
GetLFTag(env)                                        signed for eu-west-1  → returns it
```

Every other service in the repo keys by `region::id`. This is the one finding I'd call structural
rather than cosmetic — it's in the storage layer, so it's cheaper to fix now than after merge.

**2. Create operations are silent upserts where AWS rejects.**

| call | Floci | AWS |
|---|---|---|
| `RegisterResource` on an already-registered ARN | 200; `RoleArn` silently overwritten (`r1` → `r2`, confirmed via `DescribeResource`) | `AlreadyExistsException` |
| `CreateLFTag` with an existing `TagKey` | 200; values replaced (`["dev"]` → `["prod"]`) | `AlreadyExistsException` |
| `DeregisterResource` on an ARN that was never registered | 200 | `EntityNotFoundException` |

`AlreadyExistsException` and `EntityNotFoundException` are both in these operations' modeled error lists.
The `CreateLFTag` one matters most: silently upserting erases the distinction between `CreateLFTag` and
`UpdateLFTag`, which the PR implements separately.

**3. `listResources` ignores all three of its parameters.** `FilterConditionList`, `MaxResults` and
`NextToken` are accepted and dropped — the code says so (`// Very basic pagination: returns all
resources for now`). `ResourceInfo.LastModified` is never populated either.

**4. The blanket catch in the handler reports server bugs as client faults.**

```java
} catch (Exception e) {
    throw new AwsException("SerializationException", "Failed to parse request or serialize response", 400);
}
```

Any `NullPointerException` or `IllegalStateException` raised inside the service layer surfaces to the
caller as a 400 blaming their request, with the real cause dropped entirely — no log, no stack trace.
Narrowing this to the Jackson exceptions it's meant for, and letting the rest become a 500 (or logging
before rethrowing), would make the next bug in here findable.

**5. `{"ResourceArn": 12345}` returns 200** — Jackson coerces the number to `"12345"` and a resource is
registered under it. No type or ARN-format validation. Minor, but `InvalidInputException` is the modeled
answer.

## What's right

- `GetLFTag` on a missing key returns `EntityNotFoundException` at **400**, which is correct — the model
  gives that shape no `httpStatusCode`, so rest-json's 400 default applies. I checked this expecting it
  to be wrong; it isn't.
- The docs page lists exactly the sixteen operations that are implemented, no more — after the EFS
  review that's a genuine improvement, and it's the difference between a doc I can trust and one I have
  to verify line by line.
- `deferred_handlers` registration in `tools/docs/services.yaml` is the correct mechanism for a handler
  without a generated table.
- Config, storage-mode plumbing and `AccountAwareStorageBackend` usage all follow the repo's conventions.
- Local run on this branch merged with current `main`: `LakeFormation*`, `AwsJson*`, catalog and registry
  suites — **31 tests, 0 failures, 0 errors**.

## Where that leaves the PR

The domain layer — service, storage, models, and the operation set — is sound work and survives intact.
What needs redoing is the transport, which is one new controller plus a descriptor change, and then the
integration test I asked for needs re-pointing at the real routes. I'd rather flag this now than have it
merge and be discovered by the first person who points the AWS CLI at it.

namanONcode:
@pgermosen all mentioned issues resolved and ci is green

pgermosen:
Re-checked at `d3a72e61`, merged onto current `main`, built, and driven with the real AWS CLI.

**The protocol migration is done properly**, and that was the blocking item. But driving the CLI across all sixteen operations turned up a new cluster that the old JSON 1.1 transport was hiding, and it takes down four of them.

## Verified fixed

| finding | evidence |
|---|---|
| **protocol** — JSON 1.1 + `X-Amz-Target` instead of rest-json | `aws lakeformation register-resource` / `describe-resource` / `list-resources` all work against the branch. No S3 fall-through. |
| **region scoping** | a resource registered in `us-west-2` is invisible to `list-resources` in `eu-west-1` |
| **silent upserts** | re-`RegisterResource` → `AlreadyExistsException`; `DeregisterResource` on an unregistered ARN → `EntityNotFoundException`; duplicate `CreateLFTag` → `AlreadyExistsException`, values no longer clobbered |
| **blanket `catch (Exception)` → 400** | gone with the handler; malformed JSON now returns `SerializationException` 400 on its own merits |
| **integration test on the wrong transport** | re-pointed at the real routes |

I checked all sixteen paths against the model programmatically rather than spot-checking: every one matches its `requestUri` exactly, `POST` in every case. The descriptor is `REST_JSON` with the controller in `resourceClasses`, the `AwsJson11Controller` wiring is reverted, and `LakeFormationJsonHandler` is gone. That is the whole of what I asked for.

`d3a72e61` is right, and it has a documentary basis worth recording — `GrantPermissionsRequest.PermissionsWithGrantOption` reads *"These permissions may only be a subset of the permissions granted in the `Privileges`."* Revoking a base permission must therefore clear the matching grant option, or the invariant breaks. Confirmed: revoke `SELECT` on a grant of `SELECT, ALTER` + grant-option `SELECT` leaves `Permissions: [ALTER]`, `PermissionsWithGrantOption: []`.

Local run on the merge: 14 tests, 0 failures. `regen_action_docs.py --strict` and `check_service_matrix.py` both exit 0.

---

## (blocking) The `LF` acronym is mis-cased in seven places, and it breaks four operations

The model classes name these fields `lfTags` / `lfTag` / `lfTagPolicy` / `lfTagExpression`. The PascalCase strategy turns those into `LfTags` / `LfTag` / …, but AWS's members are `LFTags` / `LFTag` / `LFTagPolicy` / `LFTagExpression`. Nothing binds.

```
AddLFTagsToResourceRequest       LFTags           ← lfTags           emits/expects LfTags
RemoveLFTagsFromResourceRequest  LFTags           ← lfTags           emits/expects LfTags
ListLFTagsResponse               LFTags           ← lfTags           emits/expects LfTags
LFTagError                       LFTag            ← lfTag            emits/expects LfTag
Resource                         LFTag            ← lfTag            emits/expects LfTag
Resource                         LFTagPolicy      ← lfTagPolicy      emits/expects LfTagPolicy
Resource                         LFTagExpression  ← lfTagExpression  emits/expects LfTagExpression
```

Through the actual CLI:

```
$ aws lakeformation add-lf-tags-to-resource --resource '{"Database":{"Name":"salesdb"}}' \
      --lf-tags TagKey=env,TagValues=dev
An error occurred (500) ... (reached max retries: 2)
```

Server side:

```
java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because "lfTags" is null
    at MemoryLakeFormationStorage.addLFTagsToResource(MemoryLakeFormationStorage.java:199)
```

`remove-lf-tags-from-resource` returns 200 for the same reason and does nothing. `list-lf-tags` returns nothing at all — the wire carries `{"LfTags":[…]}` and botocore is looking for `LFTags`, so it drops the whole list.

**The one that worries me most is `Resource.LFTag`,** because it is the tag-based access control path — the governance feature this PR's summary leads with. A TBAC grant is accepted and silently loses its resource:

```
$ aws lakeformation grant-permissions --principal …role/tbac \
      --resource '{"LFTag":{"CatalogId":"…","TagKey":"env","TagValues":["dev"]}}' \
      --permissions ASSOCIATE
(200)

$ aws lakeformation list-permissions
"Resource": {}          ← the LFTag body is gone
```

And because the permission key is built from an empty resource, two grants on *different* LF-Tags collapse into one record:

```
grant LFTag env=dev   ASSOCIATE
grant LFTag team=core DESCRIBE
list → one entry:  Resource {}   Permissions ["ASSOCIATE","DESCRIBE"]
```

Revoking either one would now hit both. `@JsonProperty("LFTags")` / `@JsonProperty("LFTag")` / etc. on the seven fields is the fix.

## (blocking) The list filters are accepted and ignored, so the answers are wrong

This is finding 3 from last round, and it moved rather than closed. The parameters are now threaded from the controller through the service into the storage signatures — but the storage bodies drop them (`MemoryLakeFormationStorage:60`, `:139`, `:182`; the comments say so). The signatures now advertise support that isn't there.

```
list-permissions --principal <role with no grants at all>  → 2 permissions, both another principal's
list-permissions --resource salesdb                        → salesdb AND otherdb
list-resources   --filter-condition-list …no-such-bucket   → probe-bucket
```

That is not an unimplemented feature, it is a wrong answer to a question the client asked. I'd also gently push back on the reasoning in the description — *"ListPermissions currently returns all explicitly granted permissions to satisfy Terraform state expectations"*. Real Lake Formation filters on `Principal` and `Resource`, and a provider reading back one resource's grants will see every other grant in the catalog.

---

## Notes

- **Unset members serialize as JSON `null` everywhere.** `ResourceInfo` emits six nulls, `DataLakeSettings` about ten, every list emits `"NextToken": null`. AWS's rest-json omits members that have no value. This is the same bug `5ed3a8ca` fixed for Cognito on `main` last week, and the same drift class as #2200: Terraform captures one shape at apply and reads a different one on refresh. `@JsonInclude(NON_NULL)` on the models covers it.
- **`docs/services/index.md:55` still advertises the old protocol** — `` `POST /` + `X-Amz-Target: AWSLakeFormation.*` | JSON 1.1 ``. I flagged this line last round. The PR description's **AWS Compatibility** section says the same thing and needs the same correction, since it is now the opposite of what the code does.
- **The grant-option subset invariant isn't enforced on the write side.** `d3a72e61` keeps it on revoke; grant still accepts `Permissions: [SELECT]` with `PermissionsWithGrantOption: [ALTER]` and stores it (200, verified). `InvalidInputException` is in `GrantPermissions`' error list.
- `{"ResourceArn": 12345}` still returns 200 and registers a resource named `"12345"`. Unchanged from last round, still minor.
- A service-layer NPE now escapes as a raw Quarkus 500 HTML page rather than an AWS error shape. Better than the old "blame the caller" 400, but an SDK still can't parse it.

## Why CI being green doesn't settle this

`LakeFormationIntegrationTest` covers **8 of the 16 operations**:

```
covered:  PutDataLakeSettings GetDataLakeSettings CreateLFTag GetLFTag UpdateLFTag
          GrantPermissions ListPermissions RevokePermissions
absent:   RegisterResource DeregisterResource ListResources DescribeResource
          DeleteLFTag ListLFTags AddLFTagsToResource RemoveLFTagsFromResource
```

Every operation broken above is in the absent half, and no test anywhere sends or reads an `LFTags` field or an `LFTag` resource. Nothing in `compatibility-tests/` touches lakeformation either. Covering the other eight — and one TBAC grant round trip — would have caught all four of these before I did, and would keep them caught.

The transport rewrite was the hard part and it is done well. What is left is a one-line-per-field serialization fix and honouring the filter arguments.

namanONcode:
@pgermosen all mentioned issues are resolved now and ci is green

pgermosen:
@namanONcode thank you for turning all of this around so quickly — and for the test work in particular, which went well beyond what I asked for.

Re-checked at `b354dda0`, merged onto current `main`, rebuilt, and driven with the real AWS CLI again.

**Both blockers from last round are genuinely fixed**, and so is every one of the notes. Coverage went from 8 of 16 operations to 16 of 16. Two things are left, one of which is new — the `ListResources` filter is implemented against operator names AWS does not use.

## Verified fixed

| finding | evidence |
|---|---|
| **`LF` acronym mis-cased in 7 places** | all seven now carry `@JsonProperty("LFTags"/"LFTag"/"LFTagPolicy"/"LFTagExpression")`. `add-lf-tags-to-resource` → `{"Failures": []}` (was NPE 500); `remove-lf-tags-from-resource` now actually removes; `list-lf-tags` returns the tag (was empty). |
| **TBAC grants losing their resource** | two grants on distinct LF-Tags now store and list as two records with `Resource.LFTag` intact — no more collapse into one `Resource {}`. |
| **`ListPermissions` filters ignored** | `--principal` with no grants → `[]`; `--resource salesdb` → salesdb only. |
| **unset members serialized as `null`** | `@JsonInclude(NON_NULL)` across the models; verified no nulls in `list-permissions` / `list-resources` output. `NextToken` is now correctly absent rather than `null`. |
| **grant-option subset not enforced on write** | `--permissions SELECT --permissions-with-grant-option ALTER` → `InvalidInputException`. |
| **`{"ResourceArn": 12345}` accepted** | → `400 InvalidInputException`. Same for a non-ARN string. |
| **service NPE escaping as a Quarkus HTML page** | now an AWS error shape via the new `AwsThrowableMapper`. |
| **`docs/services/index.md`** | line 51 now reads `POST /<Action>` \| REST JSON \| 16. |

Full suite on the merge with current `main`: **12,054 tests, 0 failures**. `regen_action_docs.py --strict` and `check_service_matrix.py` both exit 0.

The test file now hits all sixteen routes, and `grantAndListPermissionsTBAC` asserts the real wire names (`Resource.LFTag.TagKey`), so a casing regression would fail the build. That is the right test to have written.

---

## (blocking) The `ListResources` filter uses operator names AWS never sends

`MemoryLakeFormationStorage:68-70` matches `EQUALS`, `NOT_EQUALS`, `BEGINS_WITH`. AWS's `ComparisonOperator` enum is:

```
EQ  NE  LE  LT  GE  GT  CONTAINS  NOT_CONTAINS  BEGINS_WITH  IN  BETWEEN
```

`EQUALS` and `NOT_EQUALS` are not in it — the values are `EQ` and `NE`. (Identical across every botocore model on my box, 2.15.38 through 2.25.10.) So of the three operators implemented, only `BEGINS_WITH` is one a client will ever send, and equality — the common case — falls through the `return true` at the end of the lambda:

```
$ ... ListResources FilterConditionList=[{Field=RESOURCE_ARN, ComparisonOperator=EQ,
                                          StringValueList=[arn:aws:s3:::other-bucket]}]
"ResourceArn": "arn:aws:s3:::other-bucket"
"ResourceArn": "arn:aws:s3:::probe-bucket"     ← unfiltered
```

The CLI accepts `EQUALS` without complaint because botocore does not validate enum strings client-side, which is why this passes by hand. It won't pass from an SDK.

Two smaller pieces of the same code:

- **Only the first condition is applied.** `LakeFormationService:72-73` passes `getFilterConditionList().get(0)` and drops the rest; AWS ANDs them. Verified — `BEGINS_WITH arn:aws:s3:::` plus `EQ other-bucket` returns both buckets.
- **An unrecognised operator silently matches everything.** `CONTAINS` is a real AWS value and returns the full list here. `InvalidInputException` is in `ListResources`' error list and would be the honest answer for anything genuinely unsupported.

## (blocking) `GrantPermissions` skips required-member validation, and the new filter then 500s on a well-formed read

`GrantPermissionsRequest` requires `Principal`, `Resource`, `Permissions`. The service validates none of them, so a grant missing `Resource` is accepted:

```
POST /GrantPermissions
{"Principal":{"DataLakePrincipalIdentifier":"...:role/nores"},"Permissions":["SELECT"]}
→ 200 {}
```

`buildPermissionsKey` is null-safe and stores it under `unknown`. The new filter code is not — `MemoryLakeFormationStorage:161` calls `getResourceKey(p.getResource())` unguarded. So the next **well-formed** filtered read, from any caller, hits:

```
POST /ListPermissions  {"Resource":{"Database":{"Name":"salesdb"}}}
→ 500 {"__type":"InternalFailure","message":"An internal error occurred"}

java.lang.NullPointerException: Cannot invoke "...model.Resource.getCatalog()" because "r" is null
    at MemoryLakeFormationStorage.getResourceKey(MemoryLakeFormationStorage.java:262)
    at MemoryLakeFormationStorage.lambda$listPermissions$1(MemoryLakeFormationStorage.java:161)
```

Missing `Principal` does the same thing one line up, at `:156`. One malformed grant breaks `ListPermissions` for the life of the process — and with persistence enabled it would outlive a restart. Rejecting the incomplete grant with `InvalidInputException` fixes the cause; null-guarding the two derefs fixes the blast radius. Worth doing both, since the store may already hold such a record.

---

## Notes

- **The `ListResources` filter is not pinned by any test.** I reverted its body to the pre-PR `return new ArrayList<>(...scan(...))` and all 9 tests still passed. `resourceLifecycle` registers exactly one resource and asserts `ResourceInfoList[0].ResourceArn`, which holds whether or not the filter runs. Worth contrasting with `listPermissions`, which *is* pinned — neutering that one fails `grantAndListPermissionsTBAC`. Registering a second resource and asserting the other one is absent would close it, and would have caught the `EQ` issue above.
- **`AwsThrowableMapper` is a repo-wide change riding in a service PR.** As a `@Provider ExceptionMapper<Throwable>` it governs the unhandled-exception path for every service in Floci. I checked it fairly hard and it is safe as far as I can measure: the more specific `AwsExceptionMapper` still wins, so normal errors are untouched (S3 still returns `<Error><Code>NoSuchBucket`, SQS still returns `<ErrorResponse>`), `WebApplicationException` passes through, and all 12,054 tests are green. Two things I would still want a maintainer to accept on purpose rather than inherit: for the XML services an unhandled exception now returns a JSON body where it used to return an HTML page — neither is parseable by an XML-expecting SDK, so it is not a regression, but it is not a fix either; and it makes a crash look tidy, which is exactly how the NPE above presents as a clean `InternalFailure` instead of an obvious stack. It may well be worth having on its own merits — just as its own PR, where it can be judged on the whole emulator rather than on Lake Formation.
- **Still ignored:** `ResourceType` and `IncludeRelated` on `ListPermissions`, and `MaxResults`/`NextToken` everywhere. Now that `NextToken` is omitted rather than null, always-one-page is at least a coherent contract, so I would leave it — but a one-line comment saying it is deliberate would stop the next reader filing it.

The hard parts — the protocol migration last round, the casing and TBAC fixes this round — are done, and the test file is now a real safety net for them. What is left is the filter operator names and one missing validation.

namanONcode:
@hectorvent @pgermosen all mentioned issues are resolved and notes are considered , ci all green

pgermosen (approved):
@namanONcode both blockers from last round are fixed, and correctly so — I checked each against the current Lake Formation API reference, not just against my own repro.

Re-checked at `508d5af2`, merged onto current `main`. CI is green across all shards, the native build, and every SDK/IaC compat job, so I didn't need to rebuild locally to confirm the suite passes.

## Verified fixed

| finding | evidence |
|---|---|
| **`ListResources` filter used operator names AWS doesn't send** | now matches `EQ`/`NE`/`BEGINS_WITH` — confirmed these (not `EQUALS`/`NOT_EQUALS`) are the actual `ComparisonOperator` enum in the current API reference. An operator outside that set now raises `InvalidInputException`, which is a real error on `ListResources`. |
| **Only the first filter condition was applied** | `listResources` now iterates the full `List<FilterCondition>` and ANDs them (returns `false` on first non-match), matching AWS's documented behavior. |
| **`ListResources` filter untested** | `resourceLifecycle` now registers two resources and asserts `ResourceInfoList` has size 1 with the matching ARN — this pins the fix; reverting it would fail the test. |
| **`GrantPermissions` accepted a grant with no `Resource`** | `Principal`, `Resource`, and `Permissions` are now checked and rejected with `InvalidInputException` — matches the API reference, which marks all three `Required: Yes`. You applied the same validation to `RevokePermissions`, which has the identical requirement — good catch, that wasn't in my last review but it was the same bug. |
| **Malformed stored grant 500'd a well-formed `ListPermissions` read** | `getResourceKey`/principal derefs in `listPermissions` are now null-guarded, so an old record missing `Principal` or `Resource` is skipped instead of throwing. Combined with the write-side validation above, both the cause and the blast radius are closed. |
| **`AwsThrowableMapper` riding in a service PR** | fully reverted — no references left anywhere in the tree. Right call to split that out; it can be judged on its own merits against the whole emulator instead of Lake Formation alone. |
| **Undocumented always-one-page contract** | `MaxResults`/`NextToken`/`ResourceType`/`IncludeRelated` omissions are now called out with one-line comments at each ignoring site, so the next reader doesn't file it as a bug. |

## (nit) The `ListResources` field filter is strict on operator, not on field

`MemoryLakeFormationStorage.listResources` only matches when `Field == "RESOURCE_ARN"`. AWS's `FilterCondition.Field` also allows `ROLE_ARN` and `LAST_MODIFIED` — a condition on either of those is silently ignored today (matches everything), while an unsupported *operator* on `RESOURCE_ARN` now correctly raises `InvalidInputException` right next to it. Same inconsistency this PR just fixed on the operator axis, just one level up on the field axis. Not asking for `ROLE_ARN`/`LAST_MODIFIED` support — just for it to fail the same honest way, or get the same "deliberate omission" comment you added for `MaxResults`/`NextToken`.

## (nit) The `GrantPermissions`/`RevokePermissions` validation and the null-guard fix aren't pinned by a test

Same category as the filter note from last round: I read the code and it's correct, but nothing in the suite would catch a regression here — e.g. a future refactor that drops the `Resource == null` check would only surface as an intermittent 500 on `ListPermissions`, which is exactly how this one was found in the first place. Worth a couple of assertions: a `GrantPermissions` call missing `Resource` returns `InvalidInputException`, and a `ListPermissions` call after a (forced/stored) malformed record doesn't 500.

---

Recommendation: **approve**. Both blockers are genuinely closed and verified against the live AWS API reference, the fixes are scoped correctly (validation added symmetrically to both Grant and Revoke), and CI is green. The two nits above are follow-up hardening, not reasons to hold this open.

PR #2430, opened 2026-08-20
fix(lambda): serialize resource policy mutations

aryasaatvik:
## Summary

Serialize Lambda resource-policy operations per canonical function ARN so concurrent permission updates cannot lose acknowledged statements or race policy reads. CloudFormation's compensation restore uses the same boundary.

## Behavior

```text
AddPermission / RemovePermission / restorePermissionStatement / GetPolicy
  -> lock by canonical function ARN
  -> re-fetch current function state
  -> mutate policy or copy a read snapshot
  -> persist or return
```

- Qualified and unqualified policies share the function lock while retaining resource-ARN scoping.
- `GetPolicy` returns copied statement maps instead of traversing mutable policy state during serialization.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Concurrent AWS SDK `AddPermission` requests could all succeed while later `GetPolicy` omitted some statements. This is covered with AWS SDK for Java v2 using 16 simultaneous permission additions and a final policy read.

## Validation

- `./mvnw test -Dtest=LambdaServiceTest#concurrentPolicyRestoreAddAndReadPreserveAllStatements,LambdaPermissionTagLayerIntegrationTest,LambdaPermissionQualifierIntegrationTest,LambdaAddressingCfnProvisionerTest` — 36 passed
- `FLOCI_ENDPOINT=http://localhost:4569 ../../mvnw test -Dtest=LambdaPermissionConcurrencyTest` from `compatibility-tests/sdk-test-java` — 1 passed
- `./mvnw test` — 11,281 tests; one unrelated timing-sensitive Kinesis assertion failed on decimal trailing-zero formatting
- `./mvnw test -Dtest=KinesisIntegrationTest` — 38 passed on rerun

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/LambdaPermissionConcurrencyTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/LambdaPermissionConcurrencyTest.java
@@ -1,0 +1,120 @@
+package com.floci.test;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.core.SdkBytes;
+import software.amazon.awssdk.services.lambda.LambdaClient;
+import software.amazon.awssdk.services.lambda.model.AddPermissionRequest;
+import software.amazon.awssdk.services.lambda.model.AddPermissionResponse;
+import software.amazon.awssdk.services.lambda.model.CreateFunctionRequest;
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Verified the fix empirically: reverting the shared per-ARN lock (`lockForPolicyMutation` returning a fresh object per call instead of the pooled one) makes the new `concurrentPolicyRestoreAddAndReadPreserveAllStatements` test fail reliably — lost statements on one run, a `ConcurrentModificationException` on another. Restored and confirmed clean. Full lambda suite passes both on the branch (531/531) and on the actual merged-with-main tree (574/574, compiled + tested the merge, not just checked for textual conflicts). Both Greptile P2 notes look like false positives on closer look: the lock-map growth concern applies equally to two pre-existing lock maps in this file, and the System.err usage in the new compatibility test matches that module's established convention. LGTM.

PR #1086, opened 2026-05-29
fix(cloudformation): forward ApiGatewayV2 Api properties to apigatewayv2 service

botondbotos:
## Summary

- `provisionApiGatewayV2Api` only forwarded `Name` and `ProtocolType`, causing WEBSOCKET stacks to fail with _"RouteSelectionExpression is required for WEBSOCKET protocol"_ even when the field was present in the template
- `Description`, `ApiKeySelectionExpression`, `CorsConfiguration`, and `Tags` were silently dropped

Wires the missing fields by forwarding scalar properties through `resolveOptional` and adding two helpers:
- `parseApiGatewayV2Tags` — handles the JSON-object tag shape `{"k": "v"}` that `AWS::ApiGatewayV2::Api` uses (distinct from the `[{Key, Value}]` array form `parseCfnTags` handles)
- `parseApiGatewayV2Cors` — converts PascalCase CFN sub-object keys to the camelCase the service layer expects; guards against intrinsic functions resolving to non-object nodes after `resolveNode`

Closes #1072

## Test plan

- Regression test `createStack_apiGatewayV2WebSocketApi_forwardsRouteSelectionExpression` added to `CloudFormationIntegrationTest` — was failing before this fix, covers all five newly wired properties end-to-end via `AmazonApiGatewayV2.GetApis`
- Full `CloudFormationIntegrationTest` (69 tests) passes

PR #2431, opened 2026-08-20
fix(apigatewayv2): CloudFormation AuthorizationScopes pass-through and scope-ingestion hardening (follow-up to #2011)

mitsukado-shinagawa:
Follow-up to #2011, addressing the non-blocking items from the approving review:

## 1. CloudFormation `AWS::ApiGatewayV2::Route` → `AuthorizationScopes` pass-through

`provisionApiGatewayV2Route` built its request from `RouteKey`/`AuthorizationType`/`AuthorizerId`/`Target` only, so a scoped route provisioned via CloudFormation could never enforce the 403 path added in #2011. The property is now resolved with the same helper the v2 Authorizer provisioner uses for `Audience`, and it is always sent (empty when absent) so an `UpdateStack` that removes `AuthorizationScopes` from the template clears the route's scopes in place instead of keeping stale enforcement.

Covered by `CloudFormationIntegrationTest#apiGatewayV2RouteAuthorizationScopesProvisionedAndClearedOnUpdate` (create with scopes → visible via GetRoutes; update without the property → cleared, same RouteId).

## 2. JSON 1.1 `AuthorizationScopes` rendering was unguarded

As noted in review, deleting the `toRouteNode` block in `ApiGatewayV2JsonHandler` left the whole suite green. New `RouteAuthorizationScopesJson11Test` drives CreateRoute/GetRoute/UpdateRoute over the `X-Amz-Target` surface (following the `ApiGatewayV2WebSocketJson11Test` pattern) and asserts the round-trip, plus key omission on unscoped routes. Deleting the rendering block now fails 1 test; deleting the REST block fails the existing `RouteAuthorizerIdResponseTest` coverage.

## 3. `authorizationScopes` unchecked cast (suppressed Copilot finding)

`createRoute`/`updateRoute` unchecked-cast the request value to `List<String>`. A JSON body with a non-string element (e.g. `["orders/read", 42]`) survived the cast and later threw `ClassCastException` in response serialization (`forEach(scopes::add)` binds to `ArrayNode.add(String)`). Elements are now normalized to strings (nulls dropped) at ingestion; regression test `authorizationScopesNonStringElementsNormalizedToStrings` added.

## Out of scope

Array-valued `aud` matching (RFC 7519 §4.1.3) — called out in review as a pre-existing, separate issue; untouched here.

## Testing

- `RouteAuthorizationScopesJson11Test` (2), `RouteAuthorizerIdResponseTest` (4), `HttpApiJwtAuthorizerScopesTest` (5) — pass
- `CloudFormationIntegrationTest` apigatewayv2 route/authorizer methods (3) — pass
- Full apigateway + apigatewayv2 sweep from the same run: ~800 tests across 80+ classes, all green

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
@@ -413,10 +413,16 @@ public class ApiGatewayV2Service {
 
     // JSON bodies can carry non-string elements (numbers, booleans); downstream code
     // (response serialization, scope enforcement) assumes String elements, so normalize
-    // at ingestion instead of unchecked-casting.
+    // at ingestion instead of unchecked-casting. A present-but-non-array value must be
+    // rejected, not treated as absent — silently returning null would create the route
+    // without scope enforcement (or clear it on update).
     private static List<String> toStringList(Object value) {
-        if (!(value instanceof List<?> list)) {
+        if (value == null) {
             return null;
+        }
+        if (!(value instanceof List<?> list)) {
+            throw new AwsException("BadRequestException",
+                    "authorizationScopes must be a list of strings", 400);
         }
         return list.stream()
                 .filter(java.util.Objects::nonNull)
--- a/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
@@ -361,9 +361,7 @@ public class ApiGatewayV2Service {
         route.setRouteKey((String) request.get("routeKey"));
         route.setAuthorizationType((String) request.getOrDefault("authorizationType", "NONE"));
         route.setAuthorizerId((String) request.get("authorizerId"));
-        @SuppressWarnings("unchecked")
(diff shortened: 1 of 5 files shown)


pgermosen (approved):
This closes out all three non-blocking follow-ups from #2011, verified with revert-and-confirm-fail on each:

1. CloudFormation `AuthorizationScopes` pass-through: reverting `provisionApiGatewayV2Route`'s use of `resolveStringListOrEmpty` (the same helper already used for the Authorizer's `Audience`) makes the new CloudFormation test fail exactly as described.
2. JSON 1.1 rendering test gap: deleting `toRouteNode`'s rendering block (previously unguarded, as I found in #2011) makes the new `RouteAuthorizationScopesJson11Test` fail.
3. `authorizationScopes` unchecked cast + the Greptile-flagged non-array-silently-disables-enforcement gap: reverting the type check reproduces a 201 instead of the expected 400, with no scope enforcement — confirmed load-bearing.

Greptile's P2 about the legacy CloudFormationResourceProvisioner gaining new behavior is a false positive — AWS::ApiGatewayV2::Route has lived in that monolith since #1086/#1122, well before this PR; AGENTS.md's rule targets new resource types there, not extending an existing case.

Full apigatewayv2+apigateway suite (1049 tests) and the CloudFormation scopes test pass clean, both on the branch and on the actual merged-with-main tree (compiled + tested the merge). LGTM.

PR #2432, opened 2026-08-20
fix(kinesis-cbor): timestamps mis-scaled 1000x on the CBOR path

### LINKED ISSUE #2368: fix(kinesis/cbor): timestamps mis-scaled 1000x on the CBOR path - AT_TIMESTAMP returns zero records, ApproximateArrivalTimestamp decodes as 1970
## Bug

On the AWS CBOR path (`Content-Type: application/x-amz-cbor-1.1`, which the real AWS SDK for Java v2 uses **by default** for Kinesis), timestamps are mis-scaled by 1000x in **both** directions, because AWS's legacy CBOR protocol encodes timestamps as epoch **milliseconds** under CBOR tag(1), while the `x-amz-json-1.1` protocol uses fractional epoch **seconds**. Floci treats both as seconds.

Two user-visible consequences, one root cause.

### 1. `AT_TIMESTAMP` silently returns zero records (request direction)

Captured directly off the wire from an unmodified `KinesisClient` calling `getShardIterator(..., AT_TIMESTAMP, timestamp)`:

```
X-Amz-Target : Kinesis_20131202.GetShardIterator
Content-Type : application/x-amz-cbor-1.1
raw body hex : ...6954696d657374616d70 c1 1b 000001a013accc38
                                      ^^ tag(1)
                                         ^^^^^^^^^^^^^^^^ 1787036486712
decoded      : {"Timestamp": 1787036486712}      <-- epoch MILLISECONDS
```

`KinesisJsonHandler.handleGetShardIterator` assumes epoch seconds:

```java
timestampMillis = Math.round(tsNode.asDouble() * 1000);   // -> 1787036486712000
```

`1787036486712000` ms is the year **58598**, so `KinesisService.getRecords`'s `AT_TIMESTAMP` branch never finds a record satisfying `arr.toEpochMilli() >= timestampMillis`, falls through to `startIndex = allRecords.size()`, and returns an empty page with a valid continuation iterator. No error - the caller just polls forever and never sees records that are definitely present.

Reproducer (real SDK, no interceptors needed):

```java
kinesis.createStream(...);                       // 1 shard
Instant cutoff = Instant.now().minusSeconds(30); // BEFORE the write
kinesis.putRecord(...);
String it = kinesis.getShardIterator(GetShardIteratorRequest.builder()
        .streamName(s).shardId(shard)
        .shardIteratorType(ShardIteratorType.AT_TIMESTAMP)
        .timestamp(cutoff).build()).shardIterator();
kinesis.getRecords(GetRecordsRequest.builder().shardIterator(it).build())
        .records();                              // -> 0, expected 1
```

The same sequence through the AWS CLI (plain JSON path) correctly returns 1 record, which is what makes this CBOR-specific.

### 2. `ApproximateArrivalTimestamp` decodes as 1970 (response direction)

`AwsJsonCborController.writeCborTimestamp` emits the JsonNode's epoch-**seconds** value under tag(1), but the SDK reads tag(1) as **millis**:

```
real write time      : 2026-08-18T07:03:06.966Z
SDK-decoded arrival  : 1970-01-21T16:23:56.587Z
```

This is silent data corruption on every `GetRecords` call from a default Java-SDK client - no exception, just a wrong value ~1000x too small. Independently reproduced; it also matches @pgermosen's observation on #2359, and confirms the corruption is Floci's tag(1) units convention rather than anything client-side.

## Expected

On the CBOR path, timestamps should round-trip as epoch **milliseconds** under tag(1) (matching AWS's own legacy-CBOR convention and therefore the SDK), while the plain-JSON path keeps fractional epoch seconds. Concretely:

- inbound: a tag(1) timestamp arriving over CBOR should be treated as millis, not multiplied by 1000
- outbound: `writeCborTimestamp` should emit millis under tag(1), not the seconds value used for JSON

The awkward part is that both paths funnel through the same `JsonNode`, so the units are currently ambiguous by the time a handler sees them - the fix probably needs the conversion to happen in the CBOR layer (which knows the protocol) rather than in each service handler. Happy to take this on if the shape sounds right to you, but flagging first since it touches a shared boundary and there may be a convention you'd prefer.

## Environment

- `floci/floci:nightly`, image built 2026-08-18T03:54:05Z
- AWS SDK for Java v2 (default configuration - no CBOR opt-out), and AWS CLI for the contrasting plain-JSON path

## Related

- #2359 / #2173 - `ApproximateArrivalTimestamp` / `StreamCreationTimestamp` scientific-notation serialization on the plain-JSON path. Same field in #2359's case, but a genuinely different defect on a different code path; #2359's description has been corrected to reflect that this issue, not that one, is what causes the empty `AT_TIMESTAMP` reads.

sheldon-nadarajah-sft:
## Summary

Fixes #2368. AWS's legacy CBOR protocol (`application/x-amz-cbor-1.1`, which the real AWS SDK for Java v2 uses by default for Kinesis) encodes tag(1) timestamps as epoch **milliseconds**; the plain-JSON protocol uses fractional epoch **seconds**. `AwsJsonCborController` treated both as seconds, in both directions:

- **Request side**: `handleGetShardIterator`'s `AT_TIMESTAMP` branch multiplied an already-millisecond value by 1000 again, landing the iterator around the year 58598 - past every real record - so `GetRecords` returned an empty page forever, with no error at all.
- **Response side**: `ApproximateArrivalTimestamp` (and every other `*Timestamp` field) was written under tag(1) as the internal epoch-seconds value unchanged, so a real client decodes every arrival timestamp 1000x too small (1970 instead of the real date).

## Fix

Both directions are fixed in `AwsJsonCborController` - the one place that already knows which wire protocol is in play - so no service handler needs to change:

- `writeCborTimestamp` now converts seconds → millis before writing tag(1).
- A new `normalizeCborTimestampsFromMillis` converts millis → seconds on every decoded request/response body, reusing the same schema-less field-name heuristic (`isTimestampField`) the writer already relied on for the identical reason on the response side.
- `BigDecimal` (not double arithmetic) preserves millisecond precision exactly in both directions, matching the approach #2173/#2359 already established for the same class of defect on a different code path.

## Tests

- `AwsJsonCborSerializerTest`'s existing assertions had actually locked in the bug as intended behaviour (an epoch-second value round-tripping unchanged under tag(1)) - updated to expect the correct millisecond value, plus new round-trip and non-timestamp-field coverage.
- Added `AwsJsonCborKinesisAtTimestampIntegrationTest`, a genuine end-to-end reproduction driven through the real HTTP CBOR endpoint: it hand-builds the `GetShardIterator` request's `Timestamp` as a raw tag(1) millisecond value **independent of this class's own writer** (so it can't pass by testing against itself, only against the actual real-AWS wire format), and confirms `GetRecords` actually finds the record.

## Verification

- [x] New reproduction test verified failing (`expected: <1> but was: <0>`) against the pre-fix code, and passing against the fix (red→green).
- [x] `AwsJsonCborSerializerTest` + `KinesisJsonHandlerTest`/`KinesisIntegrationTest` + all `core.common` tests: 527 tests, 0 failures, 0 errors.
- [x] Checked every other CBOR-routed service (DynamoDB, DynamoDB Streams, SQS, SNS, Step Functions, CloudWatch Metrics) for existing tests asserting exact numeric timestamp values over the CBOR path - none exist, so this fix's test blast radius is fully contained to the files touched here.
- [x] Full suite (`mvn test`): 11283 tests, 1 failure - `RuntimeApiServerTest.extensionShutdown_racedAgainstStop_isNeverOrphaned`, confirmed pre-existing/unrelated timing flakiness (reproduces identically on pristine `main`, nothing to do with CBOR/timestamps).

## Note on overlap with #2173/#2359

Distinct, non-overlapping fix from both: those cover scientific-notation serialization on the **plain-JSON** path only (a different code path, `KinesisJsonHandler`'s response building, not `AwsJsonCborController`'s CBOR tag(1) handling). Per #2359's own description, that PR's original overclaim that it was the root cause of the `AT_TIMESTAMP` failure was already corrected once this issue was filed - this PR is the actual fix for that failure.

## Type of change

- [x] Bug fix (`fix:`)

## Checklist

- [x] `mvn test` passes locally (target module + broader suite, see above)
- [x] New integration test added, reproducing the actual reported symptom

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/AwsJsonCborController.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/AwsJsonCborController.java
@@ -81,20 +81,40 @@ public class AwsJsonCborController {
 
 
     /**
-     * Serializes a JsonNode to CBOR bytes, encoding timestamp shapes with CBOR tag 1
-     * (RFC 8949 section 3.4.2) as required by the smithy-rpc-v2-cbor protocol
-     * specification — as epoch <em>milliseconds</em>, AWS's own convention for tag(1)
-     * on this protocol, not the fractional epoch seconds RFC 8949 uses as its own
-     * example. See {@link #writeCborTimestamp} for why.
+     * Serializes a JsonNode to CBOR bytes for the smithy-rpc-v2-cbor protocol, encoding
+     * timestamp shapes with CBOR tag 1 as plain epoch seconds (RFC 8949 section 3.4.2,
+     * unmodified) — this protocol uses RFC 8949's own native tag(1) convention, unlike
+     * the older, separate {@code application/x-amz-cbor-1.1} dialect (see
+     * {@link #nodeToLegacyCbor}), which encodes tag(1) as epoch milliseconds instead.
+     * These are two genuinely different wire conventions sharing the same CBOR tag
+     * mechanism, not one protocol with an inconsistency — see floci-io/floci#2368's
+     * resolution for the investigation that distinguished them (CloudWatch Metrics, on
+     * this protocol, needs the value unscaled; Kinesis, on the legacy dialect, needs it
+     * scaled to milliseconds).
      * <p>
      * Package-private so the timestamp-tagging behaviour can be unit-tested directly
      * without booting Quarkus.
      */
     static byte[] nodeToSmithyCbor(JsonNode node) throws Exception {
+        return writeCbor(node, false);
+    }
+
+    /**
+     * Serializes a JsonNode to CBOR bytes for the legacy {@code application/x-amz-cbor-1.1}
+     * dialect (routed through {@link #handleCborRequest}) — encodes timestamp shapes as
+     * CBOR tag(1) epoch <em>milliseconds</em>, AWS's convention for this specific,
+     * pre-Smithy dialect. See {@link #nodeToSmithyCbor} for how this differs from the
+     * newer smithy-rpc-v2-cbor protocol, and {@link #writeCborTimestamp} for why.
+     */
+    static byte[] nodeToLegacyCbor(JsonNode node) throws Exception {
+        return writeCbor(node, true);
+    }
+
+    private static byte[] writeCbor(JsonNode node, boolean legacyDialect) throws Exception {
         ByteArrayOutputStream out = new ByteArrayOutputStream();
         CBORFactory factory = (CBORFactory) CBOR_MAPPER.getFactory();
         try (CBORGenerator gen = factory.createGenerator(out)) {
-            writeNodeToCbor(gen, node, false);
+            writeNodeToCbor(gen, node, false, legacyDialect);
         }
         return out.toByteArray();
     }
@@ -105,16 +125,18 @@ public class AwsJsonCborController {
      * up the Smithy shape): when set and the node is a number, it is emitted as a CBOR
      * tag(1) timestamp. The flag is set by {@link #isTimestampField(String)} for matching
      * object fields and is propagated into array elements, so both scalar timestamps and
-     * elements of timestamp lists are tagged.
+     * elements of timestamp lists are tagged. {@code legacyDialect} selects which of the
+     * two wire conventions tag(1) uses - see {@link #nodeToSmithyCbor}/{@link #nodeToLegacyCbor}.
      */
-    private static void writeNodeToCbor(CBORGenerator gen, JsonNode node, boolean numberIsTimestamp) throws Exception {
+    private static void writeNodeToCbor(CBORGenerator gen, JsonNode node, boolean numberIsTimestamp,
+            boolean legacyDialect) throws Exception {
         if (node.isObject()) {
             gen.writeStartObject();
             Iterator<Map.Entry<String, JsonNode>> fields = node.fields();
             while (fields.hasNext()) {
                 Map.Entry<String, JsonNode> entry = fields.next();
                 gen.writeFieldName(entry.getKey());
-                writeNodeToCbor(gen, entry.getValue(), isTimestampField(entry.getKey()));
+                writeNodeToCbor(gen, entry.getValue(), isTimestampField(entry.getKey()), legacyDialect);
             }
             gen.writeEndObject();
         } else if (node.isArray()) {
@@ -122,11 +144,11 @@ public class AwsJsonCborController {
             for (JsonNode item : node) {
                 // Propagate the timestamp context so every element of a timestamp list
                 // (e.g. CloudWatch GetMetricData MetricDataResults[].Timestamps) is tagged.
-                writeNodeToCbor(gen, item, numberIsTimestamp);
+                writeNodeToCbor(gen, item, numberIsTimestamp, legacyDialect);
             }
             gen.writeEndArray();
         } else if (numberIsTimestamp && node.isNumber()) {
-            writeCborTimestamp(gen, node);
+            writeCborTimestamp(gen, node, legacyDialect);
         } else if (node.isTextual()) {
             gen.writeString(node.textValue());
         } else if (node.isDouble() || node.isFloat()) {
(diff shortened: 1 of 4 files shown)


sheldon-nadarajah-sft:
CI (native / sdk-test-go, Compatibility Tests) caught a real regression in the previous commit, worth flagging here explicitly rather than silently amending: scaling *every* CBOR-routed service's tag(1) timestamps to milliseconds broke CloudWatch Metrics (`TestCloudWatch/GetMetricStatistics`, `TestCloudWatch/GetMetricData` - "Should NOT be empty, but was []").

Root cause: there are genuinely **two different CBOR dialects** sharing `AwsJsonCborController`, not one protocol with an inconsistency:
- The legacy `application/x-amz-cbor-1.1` dialect (`handleCborRequest`) - Kinesis's default transport, confirmed via a real wire capture to use epoch **milliseconds** under tag(1). This is the dialect this issue actually reported.
- `smithy-rpc-v2-cbor` (`handleSmithyRpcV2Cbor`) - CloudWatch Metrics' transport (confirmed via the Go SDK compat test's own comment), which uses RFC 8949's own tag(1) convention: plain, **unscaled** epoch seconds.

Fixed by splitting the previously-shared methods into dialect-specific pairs (`nodeToSmithyCbor`/plain `bodyToJson` stay unscaled; `nodeToLegacyCbor`/`normalizeLegacyCborTimestampsFromMillis` apply the scaling, used only by `handleCborRequest`). Added `AwsJsonCborCloudWatchTimestampIntegrationTest` as a genuine end-to-end regression guard for the CloudWatch path, alongside the existing Kinesis one. Verified: both end-to-end tests, the full `AwsJsonCborSerializerTest` suite (now covering both dialects explicitly), and 10714 tests across the kinesis/cloudwatch/core.common packages all pass with zero failures.

Watching CI on the latest commit to confirm the native sdk-test-go job is green now.

pgermosen (dismissed):
Independently verified the two-dialect premise against the actual botocore service models rather than taking it on trust: Kinesis's model reports `"protocol": "json", "jsonVersion": "1.1"` (the legacy target-based protocol), while CloudWatch's reports `"protocol": "smithy-rpc-v2-cbor"` first among its supported protocols — structurally different protocols, not an inconsistency within one.

Given this PR already had one real regression caught by CI mid-review, I reverted both directions separately to make sure each is actually load-bearing:
- Routed CloudWatch through the millisecond-scaling legacy path → reproduced the exact CI regression (`GetMetricStatistics` returns empty `Datapoints`). Restored, green.
- Removed the millisecond conversion entirely → reproduced the original reported bug exactly (`AT_TIMESTAMP... expected: <1> but was: <0>`) plus 4 unit failures. Restored, green.

Full `AwsJsonCbor*`, `kinesis`, and `cloudwatch` suites pass, and I additionally ran DynamoDB/SQS/SNS/Step Functions (the other services sharing this same controller per its class doc) as a blast-radius check — clean. Also clean on a fresh merge against current `main`.

One non-blocking nit still open: Copilot's suggestion to use `BigDecimal.compareTo` instead of `doubleValue()` exact-equality in two round-trip assertions hasn't been addressed. Not blocking — the values involved don't hit precision loss in practice — but worth a follow-up.

Good to merge.

pgermosen:
Withdrawing my approval: both commits in this PR carry a `Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>` trailer, which the repo's own conventions explicitly disallow — CONTRIBUTING.md and AGENTS.md both say "Do not include `Co-Authored-By` trailers for AI tools in commit messages."

Could you amend or rebase to drop those trailers and force-push? The technical fix itself still checks out (I verified it independently, including reverting each half of the dialect split separately to confirm both are load-bearing) — this is purely a commit-hygiene blocker, not a re-review of the logic.

sheldon-nadarajah-sft:
@pgermosen  Rebased and force-pushed — both commits' `Co-Authored-By: Claude Sonnet 5` trailers are stripped, content is otherwise byte-identical (verified via `git diff` against the prior tree). Thanks for catching that.

pgermosen (approved):
Re-checked after the force-push. The commit-hygiene blocker is resolved and the fix is unchanged.

## The rebase did what you said it did

- **Trailers gone.** Scanning both commit messages on the new head for `Co-Authored-By`, `Generated
  with`, `Claude`, `Anthropic`: zero hits. The old head (`ac6d1ef0`) had exactly two — one per commit.
- **Content byte-identical, not just diff-identical.** Both heads resolve to the same tree object,
  `478523cd9323b429dfb87e3716ba273241ea0690`. Not "the diff looks the same" — git is telling us it is
  the same tree. So the verification I did on Aug 21 (reverting each half of the dialect split
  separately to confirm both are load-bearing) still applies without redoing it.
- Author dates preserved, `Closes #2368` intact on the first commit.

## New verification: the premise is confirmed from the SDK's own bytecode

My earlier pass established the two-dialect premise from the botocore protocol names, and your commit
message cites a wire capture. Both are indirect. The AWS SDK jars are on disk here, so I disassembled
the encoder itself — this is now settled rather than inferred:

- `DefaultKinesisClient` builds an `AwsCborProtocolFactory` unconditionally, so the Java SDK v2 Kinesis
  client is on the legacy dialect by default. No env flag involved.
- `AwsCborProtocolFactory` registers `MarshallLocation.PAYLOAD → TimestampFormatTrait.Format.UNIX_TIMESTAMP_MILLIS`,
  for marshalling *and* unmarshalling.
- `SdkCborGenerator.writeValue(Instant)` is literally `writeTag(1)` followed by
  `writeNumber(instant.toEpochMilli())`.
- `JsonUnmarshallingParser` decodes `UNIX_TIMESTAMP_MILLIS` as `Instant.ofEpochMilli(getLongValue())`.
- The content type resolver is built with the prefix `application/x-amz-cbor-`.

Tag(1) plus epoch milliseconds, both directions, straight from the shipped SDK (2.44.14). That is
exactly what this PR implements, and it means the "tag 1 is RFC 8949 epoch *seconds*" reading — which
is what the code held before, and which is correct for smithy-rpc-v2-cbor — was the wrong spec for this
dialect. The dialect split is the right shape of fix.

## Local run

On the PR head merged with current `main` (`04d120fd`, 16 commits ahead of the PR's base): the full
`AwsJsonCbor*`, Kinesis, Kinesis Analytics, CloudWatch Metrics and CloudWatch Logs suites —
**370 tests across 20 classes, 0 failures, 0 errors**. Nothing on `main` since the PR branched touches
`AwsJsonCborController`, so the merge is a clean superset.

## Two non-blocking notes

**1. The `BigDecimal.doubleValue()` assertions are still there.** `AwsJsonCborSerializerTest:227-228`:

```java
assertEquals(1700000000, decodedIntegral.path("Timestamp").decimalValue().doubleValue());
assertEquals(1700000000.712, decodedFractional.path("Timestamp").decimalValue().doubleValue());
```

Both pass deterministically — the values are well inside exact double range, which is why I called it a
nit last time and still do. But the production change deliberately went to `BigDecimal` to avoid double
arithmetic, and these two assertions verify the result only to double precision, so the test is a weaker
guarantee than the code makes. `assertEquals(0, expected.compareTo(actual))` on the `BigDecimal`s is the
same pattern #1833 landed for the Kinesis `BigDecimal` scale flake, and it makes the test agree with the
reason the code uses `BigDecimal` at all.

**2. The `monitoring` arm of the legacy route is now scaled and untested — and your javadoc already says
so.** Worth recording what that means concretely, because it's the same failure that bit CI once.
`handleCborRequest` dispatches `dynamodb`, `sqs`, `sns`, `kinesis`, `states` and `monitoring`, and now
applies millisecond scaling to all of them. Of those, only Kinesis and CloudWatch Metrics have fields
matching `isTimestampField` (`endsWith("Timestamp"|"Timestamps")`) — DynamoDB Streams uses
`ApproximateCreationDateTime`, Step Functions uses `startDate`/`stopDate`, and SQS's `SentTimestamp` is
a string inside a map, which the number-only conversion skips. So CloudWatch Metrics is the one other
service actually affected.

I drove it end to end: `PutMetricData` over `application/x-amz-cbor-1.1` with `X-Amz-Target:
GraniteServiceVersion20100801.PutMetricData`, `Timestamp` sent as epoch seconds → 200, and the datapoint
lands at `Timestamp: 1787280` (January 1970). `GetMetricStatistics` over a window around now returns
`"Datapoints":[]`; over a window around 1970 it finds it.

To be clear about what that does and doesn't show: **this is not a regression.** By the dialect, a
client on `x-amz-cbor-1.1` should be sending millis, in which case the new scaling is correct and the
old behaviour was wrong — the same argument that justifies the Kinesis fix. And no real SDK routes
CloudWatch through this dialect (Java v2 CloudWatch is query protocol, the Go SDK is rpc-v2-cbor), so
the arm is effectively unreachable by real clients. It's consistent-by-construction rather than
verified, which is precisely what your javadoc says: *"Not reconfirmed for every other service
dispatched below; none currently has test coverage asserting an exact numeric timestamp value over this
specific path."* That disclosure is accurate, and I'd rather have it in the javadoc than have the gap go
unmentioned. Recording the concrete behaviour here so it's on the record if anyone ever does point a
CBOR-1.1 client at CloudWatch.

## Verdict

The technical fix was already verified and is provably unchanged; the trailers that blocked it are gone.
Re-approving.

PR #2436, opened 2026-08-20
docs(bedrockagentcore): document the AgentCore services

kannandreams:
## Summary

Bedrock AgentCore was added in #2316 (control plane for agent runtimes, gateways,
memory, and workload identity, plus an `InvokeAgentRuntime` stub), but it was never
registered in the user-facing documentation, so it is not discoverable.

This PR fills those gaps — documentation only, no code changes:

- **`docs/services/index.md`** — two rows in the service matrix: AgentCore Control
  (31 operations + tagging via the shared `/tags/{arn}` route) and the AgentCore
  data plane (1 operation, canned-response stub), following the existing
  AppConfig / AppConfigData control-vs-data-plane row pattern.
- **`docs/configuration/environment-variables.md`** — the four AgentCore variables
  (`FLOCI_SERVICES_BEDROCK_AGENT_CORE_CONTROL_ENABLED`, `..._ENABLED`,
  `..._INVOKE_RESPONSE`, `..._VALIDATE_RUNTIME_EXISTS`), with defaults taken from
  `EmulatorConfig` / `application.yml`.
- **`README.md`** — AgentCore added to the architecture diagram, the
  "Data, analytics, and AI" category row, and the emulation-approach table.

Intentionally not changed: `mkdocs.yml` already has the nav entry for
`services/bedrock-agentcore.md`, and `tools/docs/services.yaml` only covers
switch-mode handlers (AgentCore is pure REST, which that tooling defers).

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A — documentation only, no wire-protocol changes. Operation counts and env-var
names/defaults were verified against the merged implementation (JAX-RS annotations
in the five AgentCore controllers, `EmulatorConfig.java`, and `application.yml`),
not against AWS.

## Checklist

- [ ] New or updated integration test added (N/A — docs only)

`make docs-check` regeneration passes in strict mode (no stale action tables, no
unregistered handlers).

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ flowchart LR
         Router["HTTP Router\nJAX-RS / Vert.x"]
 
         subgraph Stateless ["Stateless Services"]
-            A["SSM · SQS · SNS\nIAM · STS · KMS\nSecrets Manager · SES\nCognito · Kinesis\nEventBridge · Scheduler · AppConfig\nCloudWatch · Step Functions\nCloudFormation · ACM · Config · CloudTrail\nAPI Gateway · AppSync · ELB v2 · Auto Scaling\nElastic Beanstalk · CodeDeploy · CodePipeline · Backup · Bedrock Runtime · Bedrock AgentCore · Route53 · Transfer"]
+            A["SSM · SQS · SNS\nIAM · STS · KMS\nSecrets Manager · SES\nCognito · Kinesis\nEventBridge · Scheduler · AppConfig\nCloudWatch · Step Functions\nCloudFormation · ACM · Config · CloudTrail\nAPI Gateway · AppSync · ELB v2 · Auto Scaling\nElastic Beanstalk · CodeDeploy · CodePipeline · Backup · FIS · Bedrock Runtime · Bedrock AgentCore · Route53 · Transfer"]
         end
 
         subgraph Stateful ["Stateful Services"]
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ flowchart LR
         Router["HTTP Router\nJAX-RS / Vert.x"]
 
         subgraph Stateless ["Stateless Services"]
-            A["SSM · SQS · SNS\nIAM · STS · KMS\nSecrets Manager · SES\nCognito · Kinesis\nEventBridge · Scheduler · AppConfig\nCloudWatch · Step Functions\nCloudFormation · ACM · Config · CloudTrail\nAPI Gateway · AppSync · ELB v2 · Auto Scaling\nElastic Beanstalk · CodeDeploy · CodePipeline · Backup · Bedrock Runtime · Route53 · Transfer"]
+            A["SSM · SQS · SNS\nIAM · STS · KMS\nSecrets Manager · SES\nCognito · Kinesis\nEventBridge · Scheduler · AppConfig\nCloudWatch · Step Functions\nCloudFormation · ACM · Config · CloudTrail\nAPI Gateway · AppSync · ELB v2 · Auto Scaling\nElastic Beanstalk · CodeDeploy · CodePipeline · Backup · Bedrock Runtime · Bedrock AgentCore · Route53 · Transfer"]
         end
 
         subgraph Stateful ["Stateful Services"]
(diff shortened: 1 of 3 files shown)


kannandreams:
@hectorvent / @pgermosen can I get some blessing for this PR if that looks good ! Appreciate your time for the review

pgermosen:
Checked the numbers rather than trusting them, since a docs PR that gets counts wrong is worse than no
docs at all. They're exact, and there's one line to fix.

## Everything quantitative verified

AgentCore landed in #2316 on 15 Aug and appears **zero** times in all three files on `main`, so the gap
is exactly as you describe.

| claim | verified against |
|---|---|
| control plane = 31 operations | 11 + 10 + 5 + 5 JAX-RS method annotations across `BedrockAgentCoreControlController`, `…Gateway…`, `…Identity…`, `…Memory…` |
| + 3 tagging via the shared `/tags/{arn}` route | `BedrockAgentCoreTagHandler` — `listTags`, `tagResource`, `untagResource` |
| data plane = 1 operation | `BedrockAgentCoreController` |
| `…_CONTROL_ENABLED` = `true` | `@WithDefault("true")` |
| `…_ENABLED` = `true` | `@WithDefault("true")` |
| `…_INVOKE_RESPONSE` = `{"output":"yes"}` | `@WithDefault("{\"output\":\"yes\"}")` |
| `…_VALIDATE_RUNTIME_EXISTS` = `false` | `@WithDefault("false")` |

The env var names deserve a specific note, because they look wrong at first glance and I nearly filed
them as a finding. The service catalog registers `bedrock-agentcore-control` / `bedrock-agentcore` — one
word — which suggests `FLOCI_SERVICES_BEDROCK_AGENTCORE_CONTROL_ENABLED`. But the config binding comes
from the `@ConfigMapping` accessors `bedrockAgentCoreControl()` / `bedrockAgentCore()`, kebab-cased to
`bedrock-agent-core-control` / `bedrock-agent-core`, which is what `application.yml:548,552` uses. Your
names are the ones that actually work. Good catch taking them from `EmulatorConfig` rather than the
catalog.

The `bedrock-agentcore.md#data-plane-invokeagentruntime` anchor also resolves — the heading is
``## Data plane — `InvokeAgentRuntime` ``, and Python-Markdown's default slugify strips the backticks and
the em-dash, then collapses the leftover whitespace run into a single hyphen. `mkdocs.yml:128` already
has the nav entry and `regen_action_docs.py --strict` exits 0 with no diff, both as you said.

## The one fix: the README diagram loses FIS

```diff
-… CodePipeline · Backup · FIS · Bedrock Runtime · Route53 · Transfer
+… CodePipeline · Backup · Bedrock Runtime · Bedrock AgentCore · Route53 · Transfer
```

FIS is still at README lines 231 and 299, so the file ends up contradicting itself — the diagram omits a
service the two tables underneath it list.

It's a merge artefact rather than anything you typed. The head is a merge commit and the two sides
disagree on that line:

```
945cfe36 Merge branch 'main' into docs/bedrock-agentcore
  parent c982bac3 (your branch) → FIS in diagram: no
  parent 353d9c84 (main)        → FIS in diagram: yes
```

Your branch predates #2435, and the merge kept your side of the line. Re-adding `FIS · ` is the whole
fix; nothing else in the diff is affected.

## Adjacent, and explicitly not your job

While checking whether AgentCore was the only omission, I compared all 85 catalog descriptors against
`docs/services/index.md`. Four more services have no row: **MWAA**, **SWF**, **Cloud Control API** and
**EMR Serverless**. (Most of the other apparent misses are AWS signing names like `monitoring` and
`events` that map to rows already present.)

Please don't grow this PR to cover them — I mention it because it suggests the matrix could use a check
that fails when a catalog descriptor has no row, the way `docs-check` already guards the action tables
for switch handlers. Happy to file that separately.

Restore FIS to the diagram and I'll approve.

kannandreams:
missed it while resolving merge conflict 🤦‍♂️. 
@pgermosen  Fixed the FIS now. please check

pgermosen (approved):
Confirmed: FIS is restored to the README architecture diagram exactly as requested (right before Bedrock Runtime), so the diagram no longer contradicts the tables below it. Everything quantitative I verified in the last pass (operation counts, env var names/defaults, the doc anchor) is unchanged and still correct. make docs-check is clean, and the branch merges without conflicts against current main. LGTM.

PR #2439, opened 2026-08-21
fix(lifecycle): run start/ready hooks when FLOCI_PORT is non-default

### LINKED ISSUE #2437: [BUG] Init hooks never run when FLOCI_PORT is set to a non-default value (start/ready never complete)
## Service

Initialization hooks / lifecycle readiness (`/_floci/init`)

## AWS API Action

N/A (the problem is with the init lifecycle state, not a specific AWS API call)

## Expected behavior

When a script is mounted in `ready.d` and `FLOCI_PORT` is set to a non-default port, Floci should report the same lifecycle state as it does on the default port:

- `completed.start = true`
- `completed.ready = true`
- the script listed under `scripts.ready` with `state: "successful"`

## Actual behavior

With `FLOCI_PORT=4577`, `start` and `ready` never become true and no ready scripts are discovered or executed.

```json
{"completed":{"boot":true,"start":false,"ready":false,"shutdown":false},"scripts":{"boot":[],"start":[],"ready":[],"shutdown":[]}}
```

The emulator itself works: it listens on the configured port and `/_floci/health` returns 200. Only the lifecycle progression past `boot` is stuck, so init hooks silently never run.

The startup log also stops after `Installed features` — the `=== AWS Local Emulator Ready ===` line never appears:

```text
INFO  [io.quarkus] floci 1.7.0 native (powered by Quarkus 3.36.3) started in 0.047s. Listening on: http://0.0.0.0:4577
INFO  [io.quarkus] Profile prod activated.
INFO  [io.quarkus] Installed features: [agroal, cdi, config-yaml, ...]
```

This is the same class of bug as #993 (`FLOCI_TLS_ENABLED=true` prevented `ready` from ever becoming true), which suggests the readiness detection added in #159 (`defer startup hooks until HTTP server is ready`) assumes the default port 4566.

## Reproduction

```bash
mkdir -p ready.d
cat > ready.d/01-test.sh <<'EOF'
#!/bin/sh
echo "[hook] ready hook executed"
EOF
chmod +x ready.d/01-test.sh

# A) default port -> hook runs
docker run -d --name floci_repro -p 4566:4566 \
  -v "$PWD/ready.d:/etc/floci/init/ready.d:ro" floci/floci:1.7.0
sleep 5
curl -s http://localhost:4566/_floci/init
docker logs floci_repro | grep '\[hook\]'
docker rm -f floci_repro

# B) custom port -> hook never runs
docker run -d --name floci_repro -e FLOCI_PORT=4577 -p 4577:4577 \
  -v "$PWD/ready.d:/etc/floci/init/ready.d:ro" floci/floci:1.7.0
sleep 5
curl -s http://localhost:4577/_floci/init
docker logs floci_repro | grep '\[hook\]'
curl -s -o /dev/null -w '%{http_code}\n' http://localhost:4577/_floci/health
docker rm -f floci_repro
```

Results:

| | `/_floci/init` | hook executed | `/_floci/health` |
| --- | --- | --- | --- |
| A) port 4566 | `start:true, ready:true`, script `successful` | yes (1x) | 200 |
| B) `FLOCI_PORT=4577` | `start:false, ready:false`, no scripts | **no (0x)** | 200 |

Everything except `FLOCI_PORT` / the published port is identical between A and B.

## Environment

- Floci version: **1.7.0** (`floci/floci:1.7.0`, revision `75c8cfc4d93b62063b6a435dff0fc77df180ff84`)
- Also reproduced on 1.6.0 (revision `e0aab2e27d896772847517a29cd4025203ddc4f8`)
- Runtime: Docker Desktop 4.85.0 on macOS (Apple Silicon)
- Storage mode: default (also reproduced with `FLOCI_STORAGE_MODE=persistent`)

## Impact / why this is easy to miss

`/_floci/health` returns 200 and a Docker healthcheck against the configured port passes, so the container is reported as healthy. There is no error in the logs. The only symptom is that the resources the hooks were supposed to create do not exist.

We hit this while running multiple stacks in parallel (one per git worktree), where each stack needs its own port. Buckets and CORS configuration silently were not created in the stacks that used a non-default port, while the containers looked healthy.

## Related

- #993 — `ready` never becomes true with TLS enabled (same lifecycle symptom, fixed)
- #159 — defer startup hooks until HTTP server is ready

## Side note

The image's bundled `HEALTHCHECK` is also hardcoded to port 4566 (`curl -f http://localhost:4566/_floci/health`), so it fails when `FLOCI_PORT` is changed and the healthcheck has to be overridden. Possibly the same hardcoded-port assumption.

anmolxlight:
Fixes #2437

## Root cause

`EmulatorLifecycle.onHttpStart()` filtered `HttpServerStart` events against a hardcoded `HTTP_PORT = 4566`, but the Quarkus HTTP server actually listens on `floci.port` (`quarkus.http.port: ${floci.port}` in `application.yml`). With any non-default `FLOCI_PORT` the guard never matched, so the deferred start/ready hook path (#159) never fired:

- `/_floci/init` stayed at `start:false, ready:false` forever
- no scripts in `ready.d`/`start.d` were discovered or executed
- `/_floci/health` still returned 200, so containers looked healthy (the silent-failure mode from the issue)

## Fix

Compare against `config.port()` instead of the hardcoded constant. TLS mode is unchanged: `TlsConfigSource` pins the HTTP backend to 4510 regardless of `floci.port`, and that branch already used its own constant.

## Testing

- New unit test `onHttpStart_nonTls_triggersHooksOnCustomPort`: with `floci.port=4577`, a server-start event on port 4577 runs START and READY hooks and marks both phases completed. Full `EmulatorLifecycleTest` suite passes (23 tests).
- End-to-end verified on JDK 25: booted the emulator with `FLOCI_PORT=4577` and a script in `/etc/floci/init/ready.d`. `/_floci/init` now reports `start:true, ready:true` with the script listed as `successful`, the script executes exactly once, and the `=== AWS Local Emulator Ready ===` banner appears.

## Side note (not in this PR)

The issue also mentions the image `HEALTHCHECK` being hardcoded to port 4566. That's a Dockerfile-level concern (`docker/Dockerfile*`) — happy to follow up separately if maintainers want it changed.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/lifecycle/EmulatorLifecycle.java
+++ b/src/main/java/io/github/hectorvent/floci/lifecycle/EmulatorLifecycle.java
@@ -49,7 +49,6 @@ import java.util.Optional;
 public class EmulatorLifecycle {
 
     private static final Logger LOG = Logger.getLogger(EmulatorLifecycle.class);
-    private static final int HTTP_PORT = 4566;
     private static final int TLS_HTTP_BACKEND_PORT = 4510;
 
     @ConfigProperty(name = "quarkus.application.version", defaultValue = "")
@@ -211,7 +210,11 @@ public class EmulatorLifecycle {
     }
 
     void onHttpStart(@ObservesAsync HttpServerStart event) {
-        int expectedPort = config.tls().enabled() ? TLS_HTTP_BACKEND_PORT : HTTP_PORT;
+        // Non-TLS: Quarkus listens on floci.port (quarkus.http.port: ${floci.port} in
+        // application.yml). TLS mode: TlsConfigSource pins the HTTP backend to 4510 and the
+        // public port is served by TlsProxyServer. Comparing against a hardcoded 4566 here
+        // made start/ready hooks never run when FLOCI_PORT was non-default (#2437).
+        int expectedPort = config.tls().enabled() ? TLS_HTTP_BACKEND_PORT : config.port();
         if (event.options().getPort() != expectedPort) {
             return;
         }
--- a/src/test/java/io/github/hectorvent/floci/lifecycle/EmulatorLifecycleTest.java
+++ b/src/test/java/io/github/hectorvent/floci/lifecycle/EmulatorLifecycleTest.java
@@ -99,6 +99,7 @@ class EmulatorLifecycleTest {
         Mockito.lenient().when(elbv2ServiceConfig.enabled()).thenReturn(false);
         Mockito.lenient().when(config.tls()).thenReturn(tlsConfig);
         Mockito.lenient().when(tlsConfig.enabled()).thenReturn(false);
+        Mockito.lenient().when(config.port()).thenReturn(4566);
 
         emulatorLifecycle = new EmulatorLifecycle(
                 storageFactory, serviceRegistry, config,
@@ -252,6 +253,22 @@ class EmulatorLifecycleTest {
 
         verify(initializationHooksRunner).run(InitializationHook.START);
         verify(initLifecycleState).markStartCompleted();
+    }
+
+    @Test
+    @DisplayName("onHttpStart on a non-default floci.port (non-TLS) triggers hook execution (#2437)")
+    void onHttpStart_nonTls_triggersHooksOnCustomPort() throws IOException, InterruptedException {
+        Mockito.lenient().when(config.port()).thenReturn(4577);
+        when(tlsConfig.enabled()).thenReturn(false);
+        when(initializationHooksRunner.hasHooks(InitializationHook.START)).thenReturn(true);
+        when(initializationHooksRunner.hasHooks(InitializationHook.READY)).thenReturn(true);
+
+        emulatorLifecycle.onHttpStart(new HttpServerStart(new HttpServerOptions().setPort(4577)));
+
+        verify(initializationHooksRunner).run(InitializationHook.START);
+        verify(initializationHooksRunner).run(InitializationHook.READY);
+        verify(initLifecycleState).markStartCompleted();
+        verify(initLifecycleState).markReadyCompleted();
     }
 
     @Test


pgermosen (approved):
Welcome, and thanks — this is a well-made first PR. Correct root cause, two-line fix, and a regression
test that actually fails without it. I verified the whole chain rather than taking the description on
trust, and every claim in it holds:

- `application.yml:14` — `quarkus.http.port: ${floci.port}`
- `application.yml:79` — `floci.port: 4566`
- `EmulatorConfig:12-13` — `@WithDefault("4566") int port()`
- `TlsConfigSource.java:126` — `properties.put("quarkus.http.port", "4510")`

So in non-TLS mode Quarkus binds `floci.port`, and the old guard compared that against a hardcoded 4566
— any non-default `FLOCI_PORT` made `onHttpStart` return early and start/ready never advanced. Your
reasoning about the TLS branch is right too, and worth stating explicitly since it's the part that could
easily have been "fixed" incorrectly: `TlsConfigSource` pins the backend to 4510 regardless of
`floci.port`, so `TLS_HTTP_BACKEND_PORT` staying a constant is correct rather than an oversight.

The silent-failure characterisation in #2437 also checks out — `/_floci/health` is served by the same
HTTP server and doesn't depend on lifecycle phase, which is exactly why the container reports healthy
while the hooks never run.

**The new test is load-bearing.** I reverted `config.port()` back to a hardcoded `4566`:

```
Tests run: 23, Failures: 1
  onHttpStart_nonTls_triggersHooksOnCustomPort  <<< FAILURE
```

One failure, and it's yours — properly targeted, no collateral damage. I also ran a second mutation
collapsing the ternary to `config.port()` unconditionally, and the existing
`onHttpStart_tls_triggersHooksOnBackendPort4510` catches that, so both arms stay covered. Between the
two, the suite now has all five relevant cases: non-TLS default port, non-TLS custom port, non-TLS wrong
port, TLS backend port, TLS public port ignored.

Local run on your branch merged with current `main`: `*Lifecycle*Test` / `*Init*Test` / `*Health*Test` —
138 tests across 12 classes, 0 failures.

**Two edge cases, for the record only — not asks, and not regressions.** Both behave identically before
and after this change: setting `QUARKUS_HTTP_PORT` directly instead of `FLOCI_PORT` still leaves the
guard comparing against `floci.port` while Quarkus binds elsewhere, and `floci.port: 0` (random port)
can't match either. Getting rid of those would mean keying the guard off something other than a port
equality check, which is a redesign and well beyond a bug fix.

**One adjacent thing I found while sweeping for the same pattern**, out of scope here but worth an issue
if you're interested in a follow-up: `AppSyncService.java:128-130` builds the URIs returned by
`CreateGraphqlApi` from a hardcoded literal.

```java
String baseUri = "http://localhost:4566";
uris.put("GRAPHQL", baseUri + "/v1/apis/" + apiId + "/graphql");
uris.put("REALTIME", "ws://localhost:4566/v1/apis/" + apiId + "/graphql/realtime");
```

With `FLOCI_PORT=4577`, a client that follows those URIs connects to nothing — same bug class as this
one, different site. The house pattern is to inject `EmulatorConfig.baseUrl()`; SNS does that
(`SnsService:149,158`), and its 4566 literal is only a convenience constructor. The remaining 4566 hits
in the tree are comments, javadoc, or test constructors.

For what it's worth, I checked whether `floci.base-url` should track `floci.port` and concluded it
shouldn't — that separation is deliberate, since `base-url` describes the externally visible address,
which behind a Docker port mapping like `-p 8080:4566` legitimately differs from the internal listen
port. `KubernetesFlociAddressResolver:72` says as much in a comment.

Good call scoping the Dockerfile `HEALTHCHECK` out explicitly instead of quietly leaving it — that is a
separate concern and it's better raised the way you raised it.

Approving.

PR #2443, opened 2026-08-21
fix(ecs): round-trip runtimePlatform and logConfiguration on task definitions

davscro:
## Problem

`RegisterTaskDefinition` reads a fixed list of fields and drops the rest, so a task definition
registered with a task-level `runtimePlatform` or a container `logConfiguration` comes back from
`DescribeTaskDefinition` without them:

```console
$ aws ecs register-task-definition --family demo \
    --runtime-platform 'cpuArchitecture=ARM64,operatingSystemFamily=LINUX' \
    --container-definitions '[{"name":"app","image":"nginx:latest","essential":true,
      "logConfiguration":{"logDriver":"awslogs","options":{"awslogs-group":"/ecs/demo",
      "awslogs-region":"eu-central-1","awslogs-stream-prefix":"ecs"}}}]'

$ aws ecs describe-task-definition --task-definition demo:1 \
    --query 'taskDefinition.[runtimePlatform,containerDefinitions[0].logConfiguration]'
[
    null,
    null
]
```

Both fields are accepted without error and then silently discarded. Any client that reads back what
it registered sees drift that is not there — Terraform on the next plan, or a deploy tool that
verifies its own write by comparing the registered revision against desired state. I hit this with
the latter: the deploy's verify step failed against Floci while the same code passes on AWS, and
the emulator was reporting `null` for fields it had just been handed.

## Fix

Parse, store and serialize both, matching the AWS wire shape:

```json
"runtimePlatform": {"cpuArchitecture": "ARM64", "operatingSystemFamily": "LINUX"}
"logConfiguration": {"logDriver": "awslogs", "options": {"awslogs-group": "..."}, "secretOptions": []}
```

Additive only. A request that omits a field still produces a response without the key — a task
definition that never set `runtimePlatform` does not start reporting `runtimePlatform: {}` — and
only the sub-keys that were sent are echoed back. Neither field changes how a local task runs:
Floci launches every task on the host's architecture, and container output stays with its Docker
container rather than being routed to the configured log driver. Both are modelled for round-trip
fidelity, the same rationale already documented on `EfsVolumeConfiguration`.

`runtimePlatform` is set on the task definition returned by `EcsService.registerTaskDefinition`
rather than by widening that method's signature, following the pattern already used there for
task-level `volumes`.

### One related change, worth a look

That post-registration mutation happens *after* the value was handed to storage, and
`PersistentStorage`/`WalStorage` serialize on `put`. The stored copy was therefore the
pre-mutation one, so in `persistent`, `hybrid` and `wal` modes these fields would be lost on
restart — as task-level `volumes` already were today. The mutated definition is now written back
(`EcsService.persistTaskDefinition`, mirroring the existing write-back idiom documented on
`persistByArn` and used by `deregisterTaskDefinition`), which fixes the new fields and the
existing `volumes` hole with them. `EcsJsonHandlerTaskDefinitionPersistenceTest` covers it with a
real `PersistentStorage` over a temp dir, since the in-memory backend keeps the live object
reference and cannot detect the miss. Happy to split this into its own PR if you would rather keep
them separate.

## Tests

- `EcsJsonHandlerRuntimePlatformLogConfigurationTest` — round-trip with both fields; only the
  sub-keys that were sent come back; a task definition that sends neither gets neither key.
- `EcsJsonHandlerTaskDefinitionPersistenceTest` — the fields survive a simulated restart.
- `EcsIntegrationTest` — register → describe over the JSON 1.1 wire, plus a regression case
  asserting the keys stay absent for a task definition that never set them.

`./mvnw test` — the 118 tests in the ECS package pass. The full suite has 91 pre-existing failures
in this environment (Lambda, API Gateway/WebSocket, ECR, Neptune, SWF — the Docker-backed suites);
they reproduce identically on `main` at 04d120fd, 49 failures / 42 errors either way.

Verified against a container built from this branch with the AWS CLI, and end to end with the tool
whose verify step originally caught it.

## Not in this PR

`AWS::ECS::TaskDefinition` in `CloudFormationResourceProvisioner` parses its own container
definitions and drops these two fields as well. `AGENTS.md` says not to add cases to that class, so
it seemed better left to whoever splits out an `EcsCfnProvisioner`.

### DIFF
--- a/docs/services/ecs.md
+++ b/docs/services/ecs.md
@@ -30,6 +30,12 @@ ECS emulates clusters, task definitions, tasks, and services. In the default con
 | `DeregisterTaskDefinition` | Mark a revision INACTIVE |
 | `DeleteTaskDefinitions` | Delete one or more task definitions |
 
+`runtimePlatform` and a container's `logConfiguration` are stored and returned exactly as
+registered, so a client that reads back what it wrote (Terraform, or a deploy tool verifying its
+own `RegisterTaskDefinition`) sees no drift. Neither changes how a local task runs: Floci launches
+every task on the host's own architecture, and a task's output stays with its Docker container
+rather than being routed to the configured log driver.
+
 ### Tasks
 
 | Operation | Description |
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandler.java
@@ -17,12 +17,14 @@ import io.github.hectorvent.floci.services.ecs.model.EcsServiceModel;
 import io.github.hectorvent.floci.services.ecs.model.EcsTask;
 import io.github.hectorvent.floci.services.ecs.model.KeyValuePair;
 import io.github.hectorvent.floci.services.ecs.model.LaunchType;
+import io.github.hectorvent.floci.services.ecs.model.LogConfiguration;
 import io.github.hectorvent.floci.services.ecs.model.MountPoint;
 import io.github.hectorvent.floci.services.ecs.model.NetworkBinding;
 import io.github.hectorvent.floci.services.ecs.model.NetworkConfiguration;
 import io.github.hectorvent.floci.services.ecs.model.NetworkMode;
 import io.github.hectorvent.floci.services.ecs.model.PortMapping;
 import io.github.hectorvent.floci.services.ecs.model.ProtectedTask;
+import io.github.hectorvent.floci.services.ecs.model.RuntimePlatform;
 import io.github.hectorvent.floci.services.ecs.model.ServiceDeployment;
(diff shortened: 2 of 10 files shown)


pgermosen (approved):
Wire shape verified against the real botocore ECS model: LogConfiguration (logDriver/options/secretOptions), Secret (name/valueFrom), and RuntimePlatform (cpuArchitecture/operatingSystemFamily) all match exactly.

Verified load-bearing: the self-flagged write-back (`persistTaskDefinition`) is genuinely necessary for persistent/hybrid/WAL backends. Reverting it makes `EcsJsonHandlerTaskDefinitionPersistenceTest` (real `PersistentStorage` + restart simulation, not the in-memory backend that would mask the bug) fail exactly as described — `runtimePlatform` comes back null.

Full ECS suite passes (222/222), both on the branch and on the actual merged-with-main tree (compiled + tested the merge). Good call leaving the CloudFormation provisioner untouched per AGENTS.md. LGTM.

PR #2446, opened 2026-08-21
fix(ecs): round-trip command and entryPoint on container definitions

### LINKED ISSUE #2445: [BUG] ECS DescribeTaskDefinition omits containerDefinitions[].command and .entryPoint
## Service

ECS

## AWS API Action

`RegisterTaskDefinition` / `DescribeTaskDefinition`

## Expected behavior

A container definition registered with `entryPoint` and/or `command` should echo both keys back on `DescribeTaskDefinition`, the way real AWS does. That round-trip matters because `describe-task-definition` output is routinely fed straight back into `register-task-definition` to cut a new revision (the documented "bump the image tag" workflow, and what most deploy tooling does under the hood).

## Actual behavior

Both keys are silently dropped from the response. `describe-task-definition` returns only `name`, `image` and `essential` for a container that was registered with `entryPoint` and `command`:

```json
{
    "name": "app",
    "image": "alpine:3.20",
    "essential": true
}
```

No error, no warning — the values simply are not serialized.

This is **parse-and-honour but never serialize**, not a parsing gap. In `src/main/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandler.java`, `parseContainerDefinitions` (line 1307) reads both fields and stores them:

```java
if (item.has("command") && item.path("command").isArray()) { ... def.setCommand(cmd); }
if (item.has("entryPoint") && item.path("entryPoint").isArray()) { ... def.setEntryPoint(ep); }
```

and `EcsContainerManager` (lines 171–178) correctly applies them when launching the container. But `containerDefinitionNode` (line 1001), which builds the `DescribeTaskDefinition` response, never reads them back — `grep -n 'getCommand()\|getEntryPoint()' EcsJsonHandler.java` returns zero hits. The state is stored and used, just never rendered.

### Why it bites in practice

Re-registering a task definition from `describe-task-definition` output silently drops the overrides, so the new revision runs the **image's default entrypoint** instead. Verified end-to-end on a stock 1.7.0 container:

* Revision 1, registered with `entryPoint: ["/bin/sh","-c"]` and a command — runs correctly. `docker inspect` on the launched container:
  ```
  ["/bin/sh","-c"] ["echo hello from the override; sleep 120"]
  ```
  and the container logs `hello from the override`.
* Revision 2, registered by piping revision 1's `describe-task-definition` output back in — both keys are gone from the JSON, and the launched container falls back to the image default:
  ```
  null ["/bin/sh"]
  ```
  It exits immediately and the task stops with `"stoppedReason": "Essential container in task exited"`, `"exitCode": 0`.

The failure mode in a real workload is worse than a crash: I hit this with a container whose `entryPoint` was `["/bin/sh","-c"]` and whose `command` fetched a CA bundle before `exec`-ing the agent. After a describe→register round trip the new revision skipped the CA-fetch step entirely and started the agent bare — the container comes up "successfully", just without the setup step, so nothing surfaces until TLS fails downstream.

Note this is the same class of round-trip omission as #2443 (`runtimePlatform` + `logConfiguration`), which is already open. Filing separately rather than folding it into that PR, since the project prefers focused PRs.

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=test

aws ecs register-task-definition \
  --family entrypoint-repro \
  --container-definitions '[{
    "name": "app",
    "image": "alpine:3.20",
    "essential": true,
    "entryPoint": ["/bin/sh", "-c"],
    "command": ["echo hello from the override"]
  }]' >/dev/null

aws ecs describe-task-definition \
  --task-definition entrypoint-repro \
  --query 'keys(taskDefinition.containerDefinitions[0])'
```

Expected `["name","image","essential","entryPoint","command"]`; actual:

```json
[
    "name",
    "image",
    "essential"
]
```

## Environment

- Floci version / image tag: `floci/floci:latest` = **1.7.0** (`sha256:bec9b9f7…`, `org.opencontainers.image.revision` `75c8cfc4d93b62063b6a435dff0fc77df180ff84`, linux/arm64). Reproduced on a clean throwaway container, not a patched build.
- Java SDK version (if applicable): n/a — reproduced with `aws-cli/2.36.26`
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker 29.6.2 on macOS 26.6.1, Apple silicon (arm64), `docker run -d -p 4566:4566 -v /var/run/docker.sock:/var/run/docker.sock floci/floci:latest`

---

Happy to send a PR if that's useful — the fix looks like a small sparse-serialization addition to `containerDefinitionNode` (mirroring how `portMappings` / `mountPoints` are emitted only when non-empty, so definitions without the fields stay byte-identical), plus a round-trip test in the existing `EcsJsonHandler*Test` / `EcsIntegrationTest` style. Just say the word and I'll open one.

davscro:
Fixes #2445.

## Problem

`RegisterTaskDefinition` already parses `containerDefinitions[].command` and `.entryPoint` (`parseContainerDefinitions` calls `setCommand`/`setEntryPoint`), and `EcsContainerManager` honours both when launching the container. But `containerDefinitionNode` never rendered them, so `DescribeTaskDefinition` silently omitted both keys — `grep -n 'getCommand()\|getEntryPoint()' EcsJsonHandler.java` returned zero hits before this change.

The bite is on the describe → register round trip that tooling uses to cut a new revision. Verified against stock `floci/floci:latest` (1.7.0):

* Revision 1, registered with `entryPoint: ["/bin/sh","-c"]` plus a command — launches correctly, `docker inspect` shows `["/bin/sh","-c"] ["echo hello from the override; sleep 120"]`.
* Revision 2, registered by piping revision 1's `describe-task-definition` output straight back in — both keys are gone from the JSON, and the container comes up as `null ["/bin/sh"]`, the image default.

The container still *starts*, which is what makes this one unpleasant to track down: nothing errors, the setup work the entrypoint was supposed to do just silently doesn't happen.

## Fix

Twelve lines in `containerDefinitionNode`, serializing both sparsely in the same style as the surrounding `portMappings` / `mountPoints` blocks — a container definition that sets neither still serializes byte-identically, so nothing changes for existing task definitions.

## Tests

* `EcsJsonHandlerCommandEntryPointTest` — parse + serialize round trip through the JSON wire path, written in the `EcsJsonHandlerVolumesTest` style (mocked `EcsService`, no Docker/Quarkus context). Two cases: the round trip itself, and an absent-stays-absent guard for the sparse behaviour.
* `EcsIntegrationTest#registerAndDescribeTaskDefinitionRoundTripsCommandAndEntryPoint` — end-to-end `RegisterTaskDefinition` → `DescribeTaskDefinition` over the real wire protocol, with a second container in the same task definition asserting both keys stay absent when unset.

Both were watched failing before the fix went in (the integration test fails on the describe assertion; the unit test on `entryPoint expected with two elements`). Full ECS suite: **120 tests, 0 failures, 0 errors** — 117 pre-existing plus the 3 added here.

```
./mvnw test -Dtest='Ecs*Test'
```

## Notes

* This is the same class of round-trip omission as #2443 (`runtimePlatform` + `logConfiguration`), which is still open. Kept as a separate branch off `main` rather than folded into that PR, since the project prefers focused PRs — the two don't touch the same lines, so they should merge in either order.
* No docs change: `docs/services/ecs.md` documents ECS at the operation level and doesn't enumerate container-definition fields, and this restores AWS-correct behaviour rather than changing any documented contract. Happy to add a note if you'd prefer one.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandler.java
@@ -1018,6 +1018,18 @@ public class EcsJsonHandler {
             n.set("portMappings", pms);
         }
 
+        if (def.getEntryPoint() != null && !def.getEntryPoint().isEmpty()) {
+            ArrayNode entryPoint = objectMapper.createArrayNode();
+            def.getEntryPoint().forEach(entryPoint::add);
+            n.set("entryPoint", entryPoint);
+        }
+
+        if (def.getCommand() != null && !def.getCommand().isEmpty()) {
+            ArrayNode command = objectMapper.createArrayNode();
+            def.getCommand().forEach(command::add);
+            n.set("command", command);
+        }
+
         if (def.getEnvironment() != null && !def.getEnvironment().isEmpty()) {
             ArrayNode envArr = objectMapper.createArrayNode();
             for (KeyValuePair kv : def.getEnvironment()) {
--- a/src/test/java/io/github/hectorvent/floci/services/ecs/EcsIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ecs/EcsIntegrationTest.java
@@ -1110,4 +1110,57 @@ class EcsIntegrationTest {
         .then()
             .statusCode(200);
     }
+
+    /**
+     * A container definition registered with {@code entryPoint} and {@code command} must echo
+     * both back on DescribeTaskDefinition. Without this, re-registering a task definition from
+     * describe output silently drops the overrides and the new revision falls back to the
+     * image's default entrypoint.
+     */
+    @Test
+    @Order(70)
+    void registerAndDescribeTaskDefinitionRoundTripsCommandAndEntryPoint() {
+        String arn = ecs("RegisterTaskDefinition")
+            .body("""
+                {
+                    "family": "entrypoint-roundtrip",
+                    "containerDefinitions": [
+                        {
+                            "name": "app",
+                            "image": "alpine:latest",
+                            "essential": true,
+                            "entryPoint": ["/bin/sh", "-c"],
+                            "command": ["fetch-ca && exec agent --serve"]
+                        },
+                        {
+                            "name": "sidecar",
+                            "image": "alpine:latest",
+                            "essential": false
+                        }
+                    ]
+                }
+                """)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .body("taskDefinition.containerDefinitions[0].entryPoint", contains("/bin/sh", "-c"))
+            .body("taskDefinition.containerDefinitions[0].command", contains("fetch-ca && exec agent --serve"))
+        .extract()
+            .path("taskDefinition.taskDefinitionArn");
+
+        ecs("DescribeTaskDefinition")
+            .body("""
+                {"taskDefinition": "%s"}
+                """.formatted(arn))
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .body("taskDefinition.containerDefinitions[0].entryPoint", contains("/bin/sh", "-c"))
+            .body("taskDefinition.containerDefinitions[0].command", contains("fetch-ca && exec agent --serve"))
+            // A container that set neither keeps both keys absent, as real AWS does.
+            .body("taskDefinition.containerDefinitions[1]", not(hasKey("entryPoint")))
+            .body("taskDefinition.containerDefinitions[1]", not(hasKey("command")));
+    }
 }
--- a/src/test/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandlerCommandEntryPointTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandlerCommandEntryPointTest.java
@@ -1,0 +1,107 @@
+package io.github.hectorvent.floci.services.ecs;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.hectorvent.floci.services.ecs.model.TaskDefinition;
+import jakarta.ws.rs.core.Response;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for the RegisterTaskDefinition JSON wire path in {@link EcsJsonHandler}:
+ * per-container {@code command} and {@code entryPoint} must be parsed from the request,
+ * stored on the container definition, and serialized back in the response.
+ *
+ * <p>Both fields were already parsed and honoured when launching the container
+ * ({@code EcsContainerManager}), but never rendered by {@code containerDefinitionNode} — so
+ * re-registering a task definition from DescribeTaskDefinition output silently dropped them
+ * and the new revision fell back to the image's default entrypoint.
+ *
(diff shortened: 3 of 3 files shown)


pgermosen (approved):
Checked `cb4b739e` merged onto current `main`, built, ran the suite, and drove a live instance with the CLI and real Docker containers. This is a clean fix and the PR description is accurate down to the failure messages.

## The bug is live and the diagnosis is exact

`git grep 'getCommand()\|getEntryPoint()' origin/main -- src/main/java/.../ecs/` still returns zero hits inside `EcsJsonHandler`, so `containerDefinitionNode` genuinely never rendered either field. Parse (`:1372` region) and launch (`EcsContainerManager:173-180`) both already handled them — this really is store-and-honour-but-never-serialize, exactly as filed in #2445.

## Verified against a running instance

Paired probe, same build with and without the twelve lines:

```
                        keys(taskDefinition.containerDefinitions[0])
with the fix      →     ["name","image","essential","entryPoint","command"]
12 lines removed  →     ["name","image","essential"]
```

The describe → register round trip that motivates the issue now survives: piping revision 1's `describe-task-definition` output straight back into `register-task-definition` produces a revision 2 that still carries both keys.

**And the runtime consequence is fixed, not just the JSON** — this is the part the tests can't reach, so I ran it. Same task definition, same revision 2 (the round-tripped one), `run-task` against real Docker on both builds:

```
with the fix      →  [ecs:entrypoint-repro:app] hello from the override
12 lines removed  →  (no container output at all)
```

On the baseline the container comes up on the image's default entrypoint, exits 0, and the task stops with `"Essential container in task exited"` — silently, with nothing in the logs. That is precisely the "worse than a crash" mode described in the issue, reproduced here rather than taken on trust.

## The tests actually pin the fix

Full ECS suite on the merge: **131 tests, 0 failures** (your 120 plus what `main` has grown since you branched).

I removed exactly the twelve added lines and re-ran. Two failures, and they are the two you named:

```
EcsIntegrationTest.registerAndDescribeTaskDefinitionRoundTripsCommandAndEntryPoint:1223 1 expectation failed.
EcsJsonHandlerCommandEntryPointTest.registerTaskDefinitionRoundTripsCommandAndEntryPoint:78
    entryPoint expected with two elements ==> expected: <true> but was: <false>
```

The two absent-stays-absent guards keep passing under the mutation, which is correct — they are guards, not the assertion under test. Restored afterwards.

The merge is clean even though #2443 landed in the same method after you branched (`a2389e77`); the two blocks don't overlap. `@Order(70)` is unused elsewhere in `EcsIntegrationTest`.

## One thing you fixed without claiming it

`CloudFormationResourceProvisioner.java:6577-6586` parses `Command` and `EntryPoint` for `AWS::ECS::TaskDefinition` too, so task definitions created through CloudFormation were dropping both on describe in the same way. That path is fixed by this change as well.

---

## Notes

**The docs precedent moved under you.** The description says no docs change is needed and offers one if preferred — but #2443, the PR you cite as the same class, merged as `a2389e77` and *did* add a note, now at `docs/services/ecs.md:33`:

> `runtimePlatform` and a container's `logConfiguration` are stored and returned exactly as registered, so a client that reads back what it wrote (Terraform, or a deploy tool verifying its own `RegisterTaskDefinition`) sees no drift. Neither changes how a local task runs […]

Extending that to `entryPoint` and `command` keeps the page consistent, and it needs one caveat the existing sentence doesn't: these two *do* change how a local task runs, since `EcsContainerManager` applies both when launching. Two sentences, and I don't think it should hold the fix up.

**`memoryReservation` is the same bug, one field over, in the same method.** After this merges, `parseContainerDefinitions` calls thirteen setters and `containerDefinitionNode` reads twelve getters. The one left is `memoryReservation` (`EcsJsonHandler.java:1372`), and it is a step worse than these two were: it is parsed in *two* places — the ECS handler and `CloudFormationResourceProvisioner.java:6571` — and read by nothing at all, so it is not even honoured at launch. Confirmed live on this build:

```
register --container-definitions '[{…,"memoryReservation":128}]'
describe → keys: ["name","image","essential","entryPoint","command"]     # no memoryReservation
```

Out of scope for a focused PR, and I'll file it separately unless you'd rather.

**Not worth changing:** `!isEmpty()` means an explicitly registered `"command": []` is dropped where AWS echoes the empty array back. Every sibling field in the method behaves the same way (`portMappings`, `environment`, `secrets`, `mountPoints`), so fixing it here alone would make the method less consistent, not more.

Issue #2447, opened 2026-08-21
EC2: PublicIpAddress 127.0.0.1 is unreachable on port 22, so no AWS-native tool can SSH to an instance


allensanborn:
## Summary

`DescribeInstances` reports `PublicIpAddress: 127.0.0.1`, but nothing listens on `127.0.0.1:22` — SSH is published on a host port from the 2200–2299 range, and the EC2 API has no field that can convey a port. So a tool that does the normal AWS thing (read `PublicIpAddress`, connect to port 22) always fails, while the instance is in fact reachable over SSH by two other routes.

Measured just now against a single instance, all three attempts using the same key:

```
$ aws ec2 describe-instances --instance-ids i-2ebd0a7ba96d17b7f \
    --query 'Reservations[0].Instances[0].{PublicIpAddress:PublicIpAddress,PublicDnsName:PublicDnsName,PrivateIpAddress:PrivateIpAddress}'
PrivateIpAddress   PublicDnsName   PublicIpAddress
192.168.215.3      localhost       127.0.0.1

$ docker ps --format '{{.Names}} -> {{.Ports}}' | grep i-2ebd0a7ba96d17b7f
floci-ec2-i-2ebd0a7ba96d17b7f -> 0.0.0.0:2200->22/tcp

1) ssh root@127.0.0.1            → ssh: connect to host 127.0.0.1 port 22: Connection refused
2) ssh -p 2200 root@127.0.0.1    → CONNECTED
3) ssh root@192.168.215.3        → CONNECTED
```

The address the API advertises as public is the one address that does not work. The address the API reports as *private* — the container's bridge IP — behaves exactly like a real public IP: port 22, no forwarding, no special knowledge required.

This is not #1984. That issue fixed **which** instances get a public IP (only those whose subnet sets `MapPublicIpOnLaunch`), and the fix in #2023 deliberately kept the value:

> `// The value stays the host-reachable 127.0.0.1/localhost for public ones.`

That reasoning holds for application ports — the socat sidecars really do publish those on `localhost`. It does not hold for SSH, because `docs/services/ec2.md` states that "SSH (22) is never re-forwarded". So `127.0.0.1` is host-reachable for everything *except* the one port that tooling reaches an instance on.

## Why this is a design question rather than a bug

[...]

PR #2448, opened 2026-08-21
fix(ec2): install packages with yum on Amazon Linux 2 guests

allensanborn:
## Summary

Two package-manager probes in `Ec2ContainerManager` try `dnf`, `apt-get` and `apk`. `public.ecr.aws/amazonlinux/amazonlinux:2` ships **`yum`** and none of those three, so both chains fall through on an Amazon Linux 2 guest.

**`startSshd()`** — the trailing `command -v sshd` fails, the method returns early, and the instance is never reachable over SSH, despite Floci having allocated a host port for it and injected `authorized_keys`.

**`metadataProxyInstallCommand()`** — reaches its `else` branch and exits 1 with `No supported package manager found for IMDS proxy dependencies`, leaving the instance without a link-local IMDS endpoint.

### Scope (corrected)

An earlier revision of this description claimed `amazonlinux:2` was "Floci's own default EC2 image" and that "no instance launched from the default image is ever reachable over SSH". **That was wrong**, and @pgermosen was right to hold the PR over it. `src/main/resources/ec2/image-catalog.yaml:1` reads:

```yaml
defaultDockerImage: public.ecr.aws/amazonlinux/amazonlinux:2023
```

and `git log -L1,1` on that file confirms it has been `:2023` since the catalog was introduced in #1252 — it was never `:2`. AL2023 has `dnf` and already works.

So this affects instances launched **explicitly** from `ami-amazonlinux2` (or its ID `ami-0abcdef1234567890`). A missing `ImageId` and unrecognised IDs both fall back to the default.

The claim was corrected everywhere it had been copied — both test comments, **both production comments** in `Ec2ContainerManager.java` (which the review didn't flag but carried the same wording), and the commit message.

## Reproduction

```
$ docker run --rm public.ecr.aws/amazonlinux/amazonlinux:2 sh -c \
    'command -v yum; for m in apt-get dnf apk; do command -v $m; done'
/usr/bin/yum          # and nothing else
```

Both package sets install cleanly under yum:

```
$ yum install -y openssh-server openssh-clients      -> /usr/sbin/sshd, /usr/bin/scp
$ yum install -y iproute socat curl ca-certificates  -> ip, socat, curl
```

## Fix

A `yum` branch in both chains, probed **after** `dnf` so that `dnf` still wins on Amazon Linux 2023 and modern Fedora/RHEL, where both exist and `yum` is only a compatibility shim over `dnf`.

**Rebased onto current `main` now that #2449 has merged.** That changed the sshd probe under this branch, so the `yum` branch installs `openssh-server openssh-clients` — matching what the `dnf` and `apt-get` branches now do. An AL2 guest needs `scp` for the same reason every other guest does; without this the rebase would have quietly reintroduced the exact gap #2449 closed, for AL2 only.

## Test change (addressing the review)

@pgermosen noted that of the two new tests, `metadataProxyInstallCommandHandlesYumForAmazonLinux2` was load-bearing but `sshdInstallProbeHandlesYumForAmazonLinux2` asserted against the *test-local* copy of `SSHD_INSTALL_PROBE_CMD`, so on its own it only confirmed the test file agreed with itself.

Fixed by giving the sshd probe the same treatment `metadataProxyInstallCommand()` already had: extracted into `static String[] sshdInstallProbeCommand()`, and the test now asserts against that. The long rationale comment moved with the script it describes; the call site keeps a short note on the exit-code handling, which is genuinely about `startSshd`.

The test-local literal and its `verify(execCreate).withCmd(eq(SSHD_INSTALL_PROBE_CMD))` are deliberately **kept** — pointing that constant at the accessor would have made the verify tautological and lost the drift protection. So the two now do different jobs: the literal pins production's exact output, and the new test detects the branch directly.

Re-ran the mutation matrix; rows 1 and 3 are now caught by the new test itself, not only by the pre-existing verify:

| mutation | before | now |
|---|---|---|
| drop the `yum` branch from the sshd probe | caught only by the `withCmd` verify | **`sshdInstallProbeHandlesYumForAmazonLinux2` fails directly** |
| move `yum` **before** `dnf` | caught only by the `withCmd` verify | **fails directly** — `dnf must be probed before yum` |
| drop the `yum` branch from the IMDS install | caught | caught |

A new assertion pins that the yum branch installs the client package too, so the #2449 parity cannot silently regress.

## Tests

`Ec2ContainerManagerTest`: **25 tests, 0 failures, 0 errors.**
All `Ec2*Test` on the rebased branch: **491 tests, 0 failures, 0 errors.**

## Context

Found while checking whether Packer's `amazon-ebs` builder can run against Floci. The remaining piece is the addressing question in #2447, deliberately kept separate.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
@@ -488,37 +488,10 @@ public class Ec2ContainerManager {
 
     private void startSshd(String containerId, String instanceId) {
         try {
-            // Install openssh-server if absent. The trailing "command -v sshd" makes the script's
-            // own exit code the source of truth for whether sshd is actually available afterward -
-            // without it, a shell if/elif chain with no matching branch (no dnf/apt-get/apk found) or
-            // whose install command itself failed (e.g. no network yet, apt lock held) still exits 0
-            // by bash convention, so every later step here would silently no-op against a daemon that
-            // was never installed while still logging success.
-            // The client package is installed alongside the server because provisioning tools
-            // need scp *on the instance*: Packer's default file transfer for a shell
-            // provisioner uploads the script with scp, and real AMIs carry it. Installing only
-            // openssh-server leaves sftp-server present but /usr/bin/scp absent, so the upload
-            // fails with "SCP failed to start. This usually means that SCP is not properly
-            // installed on the remote system." The guard tests for scp too: keying it on sshd
-            // alone would skip the install entirely on an image that already has the server but
-            // no client. Package names differ -- openssh-clients on rpm distributions,
-            // openssh-client on Debian; apk's openssh already contains both.
-            // The two are not equally fatal, so the script separates them: exit 1 means no sshd
-            // and there is nothing to start, while exit 2 means sshd is there but the client
-            // package did not land. A guest that can serve SSH but cannot scp is still worth
-            // starting - it just cannot run a Packer shell provisioner - so that case warns and
-            // continues rather than leaving the instance unreachable. Checking only sshd at the
-            // end would report that state as outright success, which is the silent failure this
-            // probe exists to prevent.
-            ContainerExecResult install = execInContainerForResult(containerId, new String[]{"sh", "-c",
-                    "if ! command -v sshd >/dev/null 2>&1 || ! command -v scp >/dev/null 2>&1; then" +
-                            "  if command -v dnf >/dev/null 2>&1; then dnf install -y openssh-server openssh-clients >/dev/null 2>&1;" +
-                            "  elif command -v apt-get >/dev/null 2>&1; then DEBIAN_FRONTEND=noninteractive apt-get install -y openssh-server openssh-client >/dev/null 2>&1;" +
-                            "  elif command -v apk >/dev/null 2>&1; then apk add --no-cache openssh >/dev/null 2>&1;" +
-                            "  fi;" +
-                            "fi;" +
-                            "command -v sshd >/dev/null 2>&1 || exit 1;" +
-                            "command -v scp >/dev/null 2>&1 || exit 2"}, 120);
+            // Exit 1 means sshd is absent and there is nothing to start; exit 2 means sshd is
+            // there but the client package did not land. Only the first is fatal - see
+            // sshdInstallProbeCommand() for why the probe distinguishes them.
+            ContainerExecResult install = execInContainerForResult(containerId, sshdInstallProbeCommand(), 120);
             if (install.exitCode() == SSH_CLIENT_MISSING_EXIT_CODE) {
                 LOG.warnv("sshd is available on EC2 instance {0} but the OpenSSH client package is not:"
                         + " scp is missing, so provisioners that upload files over scp will fail: {1}",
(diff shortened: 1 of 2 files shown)


pgermosen:
Welcome, and this is a good first PR — correct fix, and correct for a reason that's easy to get wrong.

Docker is available on my machine, so rather than reasoning about the images I ran them:

```
public.ecr.aws/amazonlinux/amazonlinux:2      no dnf | HAS yum | no apt-get | no apk | sshd absent
public.ecr.aws/amazonlinux/amazonlinux:2023   HAS dnf | HAS yum | no apt-get | no apk | sshd absent
```

That confirms both halves of your change. On `amazonlinux:2` all three existing branches miss, the chain
falls through, and the trailing `command -v sshd` fails exactly as you describe; the IMDS chain reaches
its `else` and exits 1.

And **probing `yum` after `dnf` is load-bearing, not a style choice** — AL2023 ships *both*, so putting
`yum` first would have silently moved AL2023 off `dnf`. That's the detail a hastier version of this fix
would have got wrong, and your comment explains why it's ordered that way, which is even better.

I also checked the scope is complete: `Ec2ContainerManager` is the only file under `src/main/java` with
package-manager probe chains, and `amazonlinux:2` is the only entry in `image-catalog.yaml` that misses
all four probes — Ubuntu and Debian hit `apt-get`, Alpine hits `apk`, AL2023 hits `dnf`.

## One correction I'd like before merge

The PR says *"Floci's own default EC2 image, `public.ecr.aws/amazonlinux/amazonlinux:2`"* and concludes
*"no instance launched from the default image is ever reachable over SSH."* That part isn't right.
`src/main/resources/ec2/image-catalog.yaml`, line 1:

```yaml
defaultDockerImage: public.ecr.aws/amazonlinux/amazonlinux:2023
```

and `git log -L1,1` on that file shows it has been `:2023` since the catalog was introduced in #1252 — it
was never `:2`. `amazonlinux:2` is reached only by explicitly asking for `ami-amazonlinux2` (or its ID
`ami-0abcdef1234567890`); a missing `ImageId` and unrecognised IDs both fall back to the default, which
is AL2023 and already works through the `dnf` branch.

So the bug is real and worth fixing — it just affects instances explicitly launched from
`ami-amazonlinux2`, not every default instance.

The reason I'd like it changed rather than shrugged off: **the same claim is repeated verbatim in the two
new test comments**, and that copy will outlive the PR description. Whoever reads
`Ec2ContainerManagerTest` in six months will take "is the default EC2 image" as fact. Adjusting those two
comments (and the summary, if you're editing anyway) is all that's needed — no code change.

Worth noting Greptile's summary got the scope right on its own — "so **explicitly selected** Amazon
Linux 2 containers can install SSH" — so the PR text is the only place carrying it.

## Mutation testing

I reverted each branch to check the tests actually hold them:

| mutation | result |
|---|---|
| drop the `yum` branch from the sshd probe | caught |
| drop the `yum` branch from the IMDS install | caught — `metadataProxyInstallCommandHandlesYumForAmazonLinux2` |
| move `yum` **before** `dnf` | caught |

One thing worth knowing about rows 1 and 3: what dies there is the **pre-existing**
`verify(execCreate).withCmd(eq(SSHD_INSTALL_PROBE_CMD))`, not your new
`sshdInstallProbeHandlesYumForAmazonLinux2`. That new test asserts on the test-local
`SSHD_INSTALL_PROBE_CMD` copy rather than on production, so on its own it can only confirm the test file
agrees with itself. It's still sound — that `withCmd` verify binds the constant to what the manager
really issues, and your description says precisely that — but the detection is coming from the old
assertion, not the new one.

Your IMDS test is the opposite and genuinely load-bearing: it calls
`Ec2ContainerManager.metadataProxyInstallCommand()` directly, so it fails on its own the moment that
branch goes. If you ever want the sshd one to carry the same weight, asserting against the production
string the same way would do it. Not a blocker — just flagging that of the two new tests, one adds
detection and one adds documentation.

## Local runs

- `Ec2ContainerManagerTest`: 23 tests, 0 failures — matches your number.
- All `Ec2*Test` suites on your branch merged with current `main`: **489 tests across 27 classes, 0
  failures, 0 errors.**

#2351 is merged and released, so the absolute-path half of the SSH story is already in. Keeping #2447
separate was the right call.

Happy to approve as soon as the "default image" wording is corrected — the code itself I'm satisfied
with.

allensanborn:
Thanks — the correction was right, and I'd rather have it caught here than six months from now in a test comment.

Pushed as `a7e7d9b7` (force, rebased — previous head was `84549959`).

## The wording

Corrected, and in more places than you flagged. The same claim was also in **two production comments** in `Ec2ContainerManager.java` — the sshd probe's `-- the default EC2 image --` aside and the IMDS one's `Amazon Linux 2, the default EC2 image, has only yum`. Those outlive a test comment by a wider margin, so they're fixed too, along with both test comments and the commit message. Everything now frames the scope as instances launched explicitly from `ami-amazonlinux2`.

I re-derived it rather than taking it on faith: `image-catalog.yaml:1` is `amazonlinux:2023`, `git log -L1,1` shows `#1252` introduced it that way, and `amazonlinux:2` is only reachable through the `ami-amazonlinux2` alias / `ami-0abcdef1234567890`.

## The sshd test

Took the suggestion. `sshdInstallProbeCommand()` is now extracted the same way `metadataProxyInstallCommand()` already was, and `sshdInstallProbeHandlesYumForAmazonLinux2` asserts against it. Re-ran your matrix:

| mutation | before | now |
|---|---|---|
| drop `yum` from the sshd probe | caught by the `withCmd` verify only | **fails the new test directly** |
| move `yum` before `dnf` | caught by the `withCmd` verify only | **fails directly** — `dnf must be probed before yum` |
| drop `yum` from the IMDS install | caught | caught |

One deliberate non-change: I **kept** the test-local `SSHD_INSTALL_PROBE_CMD` literal and its `verify(execCreate).withCmd(eq(...))`. Repointing that constant at the accessor would have made the verify compare production to itself and quietly thrown away the drift protection you were relying on. So the literal still pins production's exact output, and the new test independently detects the branch.

The long rationale comment moved into the accessor's javadoc, since it describes the script; the call site keeps a short note about the exit-code handling, which is genuinely `startSshd`'s business.

## One thing the rebase turned up

Worth flagging since it postdates your review: **#2449 merged after you tested, and it rewrote the exact block this branch touches.** Resolving the conflict by keeping this branch's side verbatim would have reintroduced the missing-`scp` gap #2449 had just closed — for AL2 guests only, and silently.

So the `yum` branch now installs `openssh-server openssh-clients`, matching what `dnf` and `apt-get` do post-#2449, with an assertion pinning that parity so a future rebase can't drop it again. Your `489 tests` run was against pre-#2449 `main`, so it wouldn't have covered this.

## Tests

- `Ec2ContainerManagerTest`: **25 tests, 0 failures, 0 errors**
- All `Ec2*Test` on the rebased branch: **491 tests, 0 failures, 0 errors**

Both mutations above were reverted and the tree re-verified clean afterwards.

pgermosen (approved):
Both items from the earlier review are genuinely resolved:

1. The "default image" wording is corrected everywhere — both the test comment and the production comment now correctly say AL2 is reached by explicitly requesting ami-amazonlinux2, not the default image.
2. The sshd-probe test is now genuinely load-bearing: `sshdInstallProbeCommand()` is a real static accessor used by production code, and the test asserts against it directly. Verified by reverting the yum branch — `sshdInstallProbeHandlesYumForAmazonLinux2` fails on its own now, not just via the old withCmd verify.

Full Ec2ContainerManagerTest suite passes (25/25), both on the branch and on the actual merged-with-main tree — worth noting this touches the same file as #2029, and the merge/compile/test cycle was clean with no semantic conflict between the two. LGTM.

PR #2449, opened 2026-08-21
fix(ec2): install the SSH client package in guests, not just the server

allensanborn:
## Summary

EC2 guests get `openssh-server` but no client package, so the instance ends up with `sshd` and `sftp-server` but no `/usr/bin/scp`. Real AMIs carry both — Amazon Linux 2 ships `openssh-clients` in the base image.

Provisioning tools need `scp` **on the instance**. Packer's default file transfer for a `shell` provisioner uploads the script with `scp`, so a build against Floci fails immediately *after* connecting successfully:

```
==> amazon-ebs.floci: Connected to SSH!
Error uploading script: SCP failed to start. This usually means that
SCP is not properly installed on the remote system.
```

Confirmed on a guest: `/usr/libexec/openssh/sftp-server` is present (it comes with `openssh-server`), but `/usr/bin/scp` and `/usr/bin/sftp` are not.

## Fix

Install the client package alongside the server. Names were verified against the images rather than assumed:

```
amazonlinux:2   yum install -y openssh-clients    -> /usr/bin/scp, /usr/bin/sftp
ubuntu:24.04    apt-get install -y openssh-client -> /usr/bin/scp, /usr/bin/sftp
alpine:3        apk add openssh                   -> /usr/bin/scp, /usr/bin/sftp
```

So `apk` needs no change — its `openssh` package already contains both — and only the rpm and deb branches gain a package.

The install guard now also tests for `scp`. Keying it on `sshd` alone would skip the install on an image that already has the server but no client, leaving provisioning broken with nothing in the logs to explain it.

## Why it matters

Without this, a Packer build needs `ssh_file_transfer_method = "sftp"` as a workaround, which would not appear in a real template written against AWS. With it (plus #2448 for Amazon Linux 2 guests), a full `amazon-ebs` build runs unmodified against Floci apart from the address question in #2447.

## Tests

`sshdInstallProbeAlsoInstallsTheSshClientPackage` in `Ec2ContainerManagerTest`, asserting on the constant that the existing `verify(execCreate).withCmd(eq(SSHD_INSTALL_PROBE_CMD))` already binds to what the manager actually execs, so it cannot drift from production.

`Ec2ContainerManagerTest`: 22 tests, 0 failures, 0 errors.

## Note on overlap

This is independent of #2448 (yum) — Debian-based guests hit the missing-client problem too — but the two together are what make a stock Packer `shell` provisioner work on the default image.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
@@ -492,10 +492,19 @@ public class Ec2ContainerManager {
             // whose install command itself failed (e.g. no network yet, apt lock held) still exits 0
             // by bash convention, so every later step here would silently no-op against a daemon that
             // was never installed while still logging success.
+            // The client package is installed alongside the server because provisioning tools
+            // need scp *on the instance*: Packer's default file transfer for a shell
+            // provisioner uploads the script with scp, and real AMIs carry it. Installing only
+            // openssh-server leaves sftp-server present but /usr/bin/scp absent, so the upload
+            // fails with "SCP failed to start. This usually means that SCP is not properly
+            // installed on the remote system." The guard tests for scp too: keying it on sshd
+            // alone would skip the install entirely on an image that already has the server but
+            // no client. Package names differ -- openssh-clients on rpm distributions,
+            // openssh-client on Debian; apk's openssh already contains both.
             ContainerExecResult install = execInContainerForResult(containerId, new String[]{"sh", "-c",
-                    "if ! command -v sshd >/dev/null 2>&1; then" +
-                            "  if command -v dnf >/dev/null 2>&1; then dnf install -y openssh-server >/dev/null 2>&1;" +
-                            "  elif command -v apt-get >/dev/null 2>&1; then DEBIAN_FRONTEND=noninteractive apt-get install -y openssh-server >/dev/null 2>&1;" +
+                    "if ! command -v sshd >/dev/null 2>&1 || ! command -v scp >/dev/null 2>&1; then" +
+                            "  if command -v dnf >/dev/null 2>&1; then dnf install -y openssh-server openssh-clients >/dev/null 2>&1;" +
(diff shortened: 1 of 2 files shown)


allensanborn:
Good catch on the P1 — confirmed and fixed in cb87c99.

The guard was widened to `! command -v sshd || ! command -v scp`, but the trailing status check was left as `command -v sshd`, so a guest that already had the server and failed to install the client exited 0 and setup carried on. That is precisely the state this PR is about, reported as success.

The two failures aren't equally fatal, though, so rather than tightening the final check to require both I've had the script separate them:

```sh
command -v sshd >/dev/null 2>&1 || exit 1;
command -v scp  >/dev/null 2>&1 || exit 2
```

- **exit 1** — no sshd: nothing to start, `startSshd` warns and returns, unchanged behaviour.
- **exit 2** — sshd present, scp missing: warns naming scp and the consequence for scp-based provisioners, then continues.

An instance that can serve SSH but can't scp is still worth making reachable; failing the whole setup there would swap a silent gap for an unreachable instance, which is worse for anyone not using Packer. The warning is what closes the "nothing in the logs to explain it" part.

Tests: the launch test that drives a failing probe is now parameterized by exit code, with `launchWhenOnlySshClientInstallFails_stillStartsSshd` covering the exit-2 path (`ssh-keygen -A` and `/usr/sbin/sshd` are still issued), and the probe-script assertion pins the trailing check itself so it can't drift back. `Ec2ContainerManagerTest`: 23 tests, 0 failures, 0 errors.

pgermosen (approved):
Verified the exit-code-2 distinction is genuinely load-bearing: reverting that branch (sshd-present-but-scp-missing no longer short-circuits to warn-and-continue) makes `launchWhenOnlySshClientInstallFails_stillStartsSshd` fail exactly as expected — sshd never gets started. Restored and confirmed clean. Package names checked against real distro docs (openssh-clients/openssh-client/apk's bundled openssh). Full `Ec2ContainerManagerTest` suite passes (23/23), both on the branch and on the actual merged-with-main tree. LGTM.

PR #2450, opened 2026-08-21
feat(ec2): complete VPC Query lifecycle support

jvanzyl:
## Summary

- Complete EC2 Query VPC lifecycle support for internet gateways, route tables and routes, network ACLs, and subnet/VPC cleanup.
- Preserve AWS SDK provider behavior, default-resource cleanup, persistence, and inline TagSpecification parsing.
- Supersede #1937 with a current-upstream extraction. The duplicated subnet tag parser called out in review is removed; omitted tag values are normalized once in the shared parser.

## AWS compatibility

The retained InvalidEgressOnlyInternetGatewayId.NotFound spelling matches the official AWS EC2 error reference:

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html

No live AWS resources or credentials were used.

## Verification

- Native EC2 tests: 294 passed, 0 failures/errors/skips.
- AWS SDK Java tests (Ec2Tests and Ec2VpcDefaultResourceCleanupTest): 56 passed, 0 failures/errors/skips.
- Full Floci suite: 11,497 passed, 0 failures/errors, 9 skipped.
- git diff --check passed; branch is linear and based on upstream main at 39ed7712a7194322b6cfa38fc5aba3dbe91f4520.


hectorvent:
Thank you for the re-extraction, and sorry this sat for two weeks after you had done everything asked on #1937. The tag parser now goes through the shared `parseTagsForResource`, the error codes match the AWS reference, and I get your 294 green locally.

**(blocking)** Please mark the PR ready and rebase. Main moved a lot in EC2 since 21 August, and one change matters: #2654 landed a real `DescribeVpcPeeringConnections`, so your empty-set stub for it now collides with main (duplicate `case` and handler). Drop the peering case, handler, service method, `VPC_PEERING_FILTERS`, its doc row and its tests; the other three conflicts are keep-both one-liners.

**(follow-up)** The model documents only `tag` and `tag-key` as filters for `DescribeEgressOnlyInternetGateways` (https://github.com/boto/botocore/blob/develop/botocore/data/ec2/2016-11-15/service-2.json), so `attachment.state`, `attachment.vpc-id` and `egress-only-internet-gateway-id` would be rejected on AWS.

The rebase is what unblocks this from my side.

KenkoGeek (changes requested):
**Medium - the new compatibility test violates the repository's explicit-type rule.**

`AGENTS.md` requires explicit Java types and explicitly says not to use `var`. `Ec2VpcDefaultResourceCleanupTest` introduces `var` for `defaultGroup`, `mainRouteTable`, and `defaultNetworkAcl` in new code.

This is not a behavioral AWS finding, but it is a contribution-policy blocker under the repository rules we are expected to enforce. Please replace those inferred locals with their concrete SDK model types before merge.

The VPC default-resource cleanup itself matches the EC2 DeleteVpc contract: AWS deletes the default security group, default route table, and default network ACL with the VPC, while user-created dependent resources must be removed first. I did not find a blocker in that behavior.

KenkoGeek (commented):
Correction to my previous review: I re-read the current `AGENTS.md`, and it does not contain an explicit ban on `var`; it says to use modern Java features only when they improve clarity and to follow existing patterns. So my earlier `var`-based contribution-policy blocker was incorrect and should not block this PR. I am not approving the current head because it is still Draft and CONFLICTING/DIRTY, and Hector's rebase request is still outstanding, including removing the DescribeVpcPeeringConnections work that main now implements. Once that rebase lands, the VPC lifecycle delta should get a fresh review on the resulting head.

pgermosen:
There's a new repo rule that blocks this regardless of the code review: main now requires every commit to be signed, and none of the 5 commits on this branch are. GitHub will refuse the merge outright once that's enforced, independent of review approval or CI, so this needs sorting before it can land.

Could you turn on commit signing (GPG or SSH, whichever you've got set up with GitHub) and re-push? An interactive rebase with signing enabled, then a force-push, should cover all of them in one pass.

pgermosen:
Thanks for the work on this @jvanzyl. Main has moved a lot since it was opened (952 commits), and most of the VPC lifecycle side (internet gateways, route tables, routes, NACLs) has landed there already. When I rebased, what was actually left came down to three unrelated things, so I split them into their own PRs so each can be reviewed and merged on its own:

- #3878: `DeleteVpc` now removes the default security group, main route table and default NACL along with the VPC
- #3879: `DescribeEgressOnlyInternetGateways` plus filter/ID validation on `DescribeVpnGateways`
- #3880: tags with no `Value` are stored as an empty string

The peering describe stub is gone, since main has a real implementation now. The commits keep you as author. I also had to re-sign them, because the ones here are unsigned and the repo requires signed commits.

Nothing needed from you here, follow the new PRs if you want to see where each ends up.

pgermosen:
Closing as everything here was merged into the other PR

PR #2456, opened 2026-08-22
fix(appsync): wait for schema creation in SDK test

### LINKED ISSUE #2453: [BUG] AppSyncTest channel-namespace tests race async schema creation, flaking native / sdk-test-java
## Service

AppSync (compatibility test suite — `sdk-test-java`)

## AWS API Action

`CreateChannelNamespace`, and by cascade `GetChannelNamespace`, `UpdateChannelNamespace`,
`ListChannelNamespaces`, `DeleteChannelNamespace`

## Expected behavior

`native / sdk-test-java` passes deterministically.

## Actual behavior

`AppSyncTest` intermittently fails five channel-namespace tests. Observed on
[run 32508554955](https://github.com/floci-io/floci/actions/runs/32508554955/job/96856140928) (PR #2440,
a DocumentDB-only change that cannot reach AppSync):

```
AppSyncTest.createChannelNamespace  ERROR    ConcurrentModificationException: Another modification is in
                                             progress at this time and it must complete before you can
                                             make your change. (Status Code: 409)
AppSyncTest.getChannelNamespace     ERROR    Parameter 'name' must not be null
AppSyncTest.updateChannelNamespace  ERROR    Parameter 'name' must not be null
AppSyncTest.listChannelNamespaces   FAILURE  Expecting actual not to be empty
AppSyncTest.deleteChannelNamespace  ERROR    Parameter 'name' must not be null
```

Only the first is a real failure. `createChannelNamespace` stores the created name in a static field, so
when it throws, the following four run with `channelNsName == null` and fail on the marshaller.

## Root cause

The suite is ordered, and two adjacent tests race the emulator's asynchronous schema creation:

- `AppSyncTest.java:616` — `@Order(94) startSchemaCreation_withAWSScalars()` calls `startSchemaCreation`
  and asserts only that `resp.status()` is non-null. It returns while the schema is still `PROCESSING`.
- `AppSyncTest.java:633` — `@Order(95) createChannelNamespace()` runs next, and
  `AppSyncService.createChannelNamespace` opens with `assertSchemaNotBusy(apiId)`
  (`AppSyncService.java:773`), which throws the 409 above while that same API's schema creation is in
  `PROCESSING` (`AppSyncService.java:258,274`).

Whether it fails depends on whether the schema worker finishes before the next request arrives, which is
why the job passes most of the time — the same PR branch passed this job twice earlier the same day.

The suite already has the fix pattern: `pollSchemaStatusToTerminal()` (`AppSyncTest.java:600`), which
`@Order(93) startSchemaCreation_invalidSchema()` calls at line 596. `@Order(94)` simply omits it.

## Suggested fix

Have `startSchemaCreation_withAWSScalars` poll to a terminal status before returning, the way
`@Order(93)` does:

```java
SchemaStatus terminal = pollSchemaStatusToTerminal();
assertThat(terminal).isNotEqualTo(SchemaStatus.PROCESSING);
```

Guarding `createChannelNamespace` instead would also work, but matching the neighbouring test keeps the
ordering assumption in one place.

## Environment

- Reproduced from CI, not locally — the race needs the native image's timing.
- Observed on `main` as of `04d120fd`; the AppSync channel-namespace tests date from #1590, which
  introduced async schema creation.

## Context

Found while reviewing #2440, where this red check was the only thing blocking an otherwise clean PR and
had nothing to do with the change under review.

aniketshukla1:
## Summary

- wait for the AWS-scalars schema creation request to reach a terminal state before the ordered channel-namespace tests begin
- assert SUCCESS for the valid schema so an unexpected terminal failure remains visible
- reuse the compatibility suite polling helper already used by the neighboring invalid-schema test

Closes #2453

## Type of change

- [x] Bug fix (fix:)
- [ ] New feature (feat:)
- [ ] Docs / chore

## AWS Compatibility

The ordered Java SDK AppSync suite could call CreateChannelNamespace while StartSchemaCreation was still PROCESSING, producing an intermittent ConcurrentModificationException and cascading failures. The test now waits for schema creation to finish successfully before advancing.

## Testing

- ../../mvnw -f pom.xml test -Dtest=AppSyncTest (80 tests passed against a local Floci server)
- ../../mvnw -f pom.xml test -DskipTests (all Java SDK compatibility test sources compiled)
- git diff --check

## Checklist

- [ ] ./mvnw test passes locally (root full suite not run)
- [x] Updated Java SDK compatibility test

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
@@ -623,6 +623,8 @@ class AppSyncTest {
                         .build());
 
         assertThat(resp.status()).isNotNull();
+        SchemaStatus terminal = pollSchemaStatusToTerminal();
+        assertThat(terminal).isEqualTo(SchemaStatus.SUCCESS);
     }
 
     // ── Channel Namespaces ─────────────────────────────────────────────


pgermosen (approved):
This is the right fix, in the right place, at the right size. Two lines, no production code touched, and it uses the helper the neighbouring test already established rather than inventing a second waiting idiom. Thank you for turning it around this quickly after the issue went up.

**The window is much wider than "intermittent" suggests.** I built current `main` merged with your branch and drove the two REST calls back to back, with no client-side gap between them:

```
POST /v1/apis/{apiId}/schemacreation
POST /v2/apis/{apiId}/channelNamespaces      ← immediately
```

```
ConcurrentModificationException on immediate CreateChannelNamespace: 59/60
```

Fifty-nine out of sixty. The guard at `AppSyncService.java:773` fires essentially every time when nothing sits between the two requests. With the poll in between, on the same server, same loop:

```
ConcurrentModificationException after pollSchemaStatusToTerminal: 0/60
```

Zero out of sixty. So the mechanism is exactly as diagnosed and your fix closes it completely — the only reason the suite was *intermittently* green is that JUnit's per-test bookkeeping and the SDK's request-pipeline construction usually take longer than the schema parse. On a loaded native-image CI runner they sometimes do not.

**I could not reproduce the failure through the Java suite locally, and I want to be straight about that.** I ran `AppSyncTest` against a local server with your fix (80/80), then reverted just your two lines and ran it three more times (80/80 each), then three more under 32 busy-loop threads on a 16-core box (80/80 each, suite time up from 1.4s to 7.8s). The flake needs the CI environment. That is not evidence against the fix — the raw-protocol numbers above show the race is real and wide — it just means the "it passes now" half of the verification is CI's to give, not mine.

**Coverage is complete, not just the one call site I named in the issue.** There are four `startSchemaCreation` calls in the file, and every one is now followed by a terminal wait before anything else runs:

| line | test | wait |
|---|---|---|
| 99 | `@Order(10) startSchemaCreation` | `@Order(11)` polls at line 113, and nothing is ordered between them |
| 591 | `@Order(93) startSchemaCreation_invalidSchema` | polls in-test at 596 |
| 618 | `@Order(94) startSchemaCreation_withAWSScalars` | **polls in-test at 626 — this PR** |
| 1277 | `@Order(200) startSchemaCreation_syntaxError...` | polls in-test at 1296 |

**The `SUCCESS` assertion matches AWS, not just Floci.** I checked before assuming, because asserting the wrong terminal value would quietly enshrine an emulator bug. The service model documents the field as:

> The current state of the schema (PROCESSING, FAILED, SUCCESS, or NOT_APPLICABLE).

`SUCCESS` is the documented terminal value for `GetSchemaCreationStatus`, and `SchemaCreationWorker.process` sets exactly that. Asserting it rather than just "not PROCESSING" is the stronger choice and I am glad you made it — a schema that terminally FAILED would otherwise satisfy the wait and push the confusion one test further down.

---

Two things I noticed while verifying, neither blocking and neither yours to fix here:

- **`SchemaCreationWorker`'s class javadoc is stale.** Line 25 says the schema transitions "to ACTIVE (success) or FAILED (error)", but the code sets `SUCCESS`. `ACTIVE` is in the `SchemaStatus` enum for other purposes, so the comment reads plausibly and is wrong — which is the worst combination. Now that a test asserts `SUCCESS` explicitly, the comment contradicts the suite. Worth a one-word fix in a passing chore commit.
- **The cascade remains.** Four of the five CI failures in #2453 were secondary: `createChannelNamespace` stores its name in a shared field that stays null when the call throws, so the following tests die on a null marshaller argument instead of reporting their own state. Your fix removes the trigger, so nothing cascades today. The amplifier is still there for the next failure at `@Order(95)`. That is inherent to a strictly ordered suite and firmly out of scope here — noting it so it does not get rediscovered from scratch.

Ready to merge from my side.

Issue #2462, opened 2026-08-22
[BUG] DynamoDB: Query on a composite GSI sort key does not enforce left-to-right / inequality-last ordering


omatheusmesmo:
## Service

DynamoDB

## AWS API Action

`Query` (KeyConditionExpression against a GSI with a multi-attribute sort key)

## Expected behavior

Per the [GSI multi-attribute key schema doc](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html#GSI.MultiAttributeKeys): "For sort key attributes, you can query them left-to-right in the order they're defined in the key schema... but you cannot skip attributes in the middle. Inequality conditions such as `>`, `<`, `BETWEEN`, or `begins_with()` must be the last condition in your query." A `KeyConditionExpression` that references a later sort-key attribute while skipping an earlier one, or that places an inequality, `BETWEEN`, or `begins_with()` on a sort-key attribute that is not the last one referenced, should be rejected with a `ValidationException`.

## Actual behavior

Floci accepts both. `DynamoDbAccessPathValidator.validateKeyConditionExpression` treats the sort-key attributes of a composite key as an unordered `Set`, checking only that each referenced attribute exists in the index and that its comparator is generally valid, never its position relative to the other sort-key attributes referenced in the same query. This is silent rather than a crash: the query returns a normal 200 with results, so an application developed and tested against floci with a query shape like this passes locally and then fails, or behaves differently, against real DynamoDB.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 dynamodb create-table \
  --table-name gsi-composite-sort \
  --attribute-definitions \
    AttributeName=id,AttributeType=S \
    AttributeName=pk,AttributeType=S \
    AttributeName=state,AttributeType=S \
    AttributeName=createdAt,AttributeType=S \
  --key-schema AttributeName=id,KeyType=HASH \
  --billing-mode PAY_PER_REQUEST \
  --global-secondary-indexes '[{"IndexName":"GSI1","KeySchema":[{"AttributeName":"pk","KeyType":"HASH"},{"AttributeName":"state","KeyType":"RANGE"},{"AttributeName":"createdAt","KeyType":"RANGE"}],"Projection":{"ProjectionType":"ALL"}}]'

aws --endpoint-url http://localhost:4566 dynamodb put-item --table-name gsi-composite-sort \
[...]

Issue #2461, opened 2026-08-22
[BUG] DynamoDB: GSI composite partition key ignores every HASH attribute but the first


omatheusmesmo:
## Service

DynamoDB

## AWS API Action

`CreateTable` (GlobalSecondaryIndexes with a multi-attribute HASH key), `Query`

## Expected behavior

Per the [GSI multi-attribute key schema doc](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html#GSI.MultiAttributeKeys): "When you query a global secondary index with multi-attribute keys, you must specify all partition key attributes using equality conditions." A GSI can have a composite partition key of up to 4 attributes; a `Query` naming only some of them should be rejected, and a `Query` naming all of them with equality should return only items matching every partition-key attribute.

## Actual behavior

Every partition-key attribute after the first is silently dropped. A `Query` that supplies only the first attribute is accepted and returns items regardless of the value of the remaining partition-key attributes (cross-partition results), while a `Query` that correctly supplies all partition-key attributes, exactly as the documentation requires, is rejected as unsupported.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 dynamodb create-table \
  --table-name gsi-composite-pk \
  --attribute-definitions \
    AttributeName=id,AttributeType=S \
    AttributeName=tenantId,AttributeType=S \
    AttributeName=region,AttributeType=S \
    AttributeName=createdAt,AttributeType=S \
  --key-schema AttributeName=id,KeyType=HASH \
  --billing-mode PAY_PER_REQUEST \
  --global-secondary-indexes '[{"IndexName":"GSI1","KeySchema":[{"AttributeName":"tenantId","KeyType":"HASH"},{"AttributeName":"region","KeyType":"HASH"},{"AttributeName":"createdAt","KeyType":"RANGE"}],"Projection":{"ProjectionType":"ALL"}}]'

aws --endpoint-url http://localhost:4566 dynamodb put-item --table-name gsi-composite-pk \
  --item '{"id":{"S":"1"},"tenantId":{"S":"acme"},"region":{"S":"us"},"createdAt":{"S":"2026-01-01"}}'
aws --endpoint-url http://localhost:4566 dynamodb put-item --table-name gsi-composite-pk \
  --item '{"id":{"S":"2"},"tenantId":{"S":"acme"},"region":{"S":"eu"},"createdAt":{"S":"2026-01-02"}}'

[...]

Issue #2460, opened 2026-08-22
[BUG] DynamoDB: CreateTable does not enforce the 4-attribute limit on GSI multi-attribute keys


omatheusmesmo:
## Service

DynamoDB

## AWS API Action

`CreateTable` (GlobalSecondaryIndexes)

## Expected behavior

Per the [GSI developer guide, "Multi-attribute key schema"](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html#GSI.MultiAttributeKeys): "With multi-attribute keys, you can create a partition key from up to four attributes and a sort key from up to four attributes, for a total of up to eight attributes per key schema." A GSI with more than 4 `HASH` elements, or more than 4 `RANGE` elements, in its `KeySchema` should be rejected with a `ValidationException`.

## Actual behavior

Floci accepts a GSI with 5 sort-key (`RANGE`) attributes with no error, and `DescribeTable` round-trips all 5.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 dynamodb create-table \
  --table-name gsi-too-many-sort-keys \
  --attribute-definitions \
    AttributeName=PK,AttributeType=S \
    AttributeName=A,AttributeType=S \
    AttributeName=B,AttributeType=S \
    AttributeName=C,AttributeType=S \
    AttributeName=D,AttributeType=S \
    AttributeName=E,AttributeType=S \
  --key-schema AttributeName=PK,KeyType=HASH \
  --billing-mode PAY_PER_REQUEST \
  --global-secondary-indexes '[{"IndexName":"GSI1","KeySchema":[{"AttributeName":"PK","KeyType":"HASH"},{"AttributeName":"A","KeyType":"RANGE"},{"AttributeName":"B","KeyType":"RANGE"},{"AttributeName":"C","KeyType":"RANGE"},{"AttributeName":"D","KeyType":"RANGE"},{"AttributeName":"E","KeyType":"RANGE"}],"Projection":{"ProjectionType":"ALL"}}]'
# 200 OK

aws --endpoint-url http://localhost:4566 dynamodb describe-table --table-name gsi-too-many-sort-keys \
  --query 'Table.GlobalSecondaryIndexes[0].KeySchema'
# Returns all 5 RANGE attributes A-E (real DynamoDB caps a GSI sort key at 4 attributes)
```

## Notes

[...]

Issue #2459, opened 2026-08-22
[BUG] DynamoDB: CreateTable accepts a Local Secondary Index with more than one sort-key attribute


omatheusmesmo:
## Service

DynamoDB

## AWS API Action

`CreateTable` (LocalSecondaryIndexes)

## Expected behavior

Every local secondary index's sort key must consist of exactly one scalar attribute, per the [LSI developer guide](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LSI.html): "Every local secondary index must meet the following conditions: the partition key is the same as that of its base table, and the sort key consists of exactly one scalar attribute." A `CreateTable` request whose `LocalSecondaryIndexes[].KeySchema` has more than one `RANGE` element should be rejected with a `ValidationException`.

## Actual behavior

Floci accepts it and creates the table successfully (`TableStatus: ACTIVE`), with both `RANGE` attributes persisted in the LSI's `KeySchema`.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 dynamodb create-table \
  --table-name lsi-multi-sort \
  --attribute-definitions \
    AttributeName=PK,AttributeType=S \
    AttributeName=SK,AttributeType=S \
    AttributeName=LSI1A,AttributeType=S \
    AttributeName=LSI1B,AttributeType=S \
  --key-schema AttributeName=PK,KeyType=HASH AttributeName=SK,KeyType=RANGE \
  --billing-mode PAY_PER_REQUEST \
  --local-secondary-indexes '[{"IndexName":"LSI1","KeySchema":[{"AttributeName":"PK","KeyType":"HASH"},{"AttributeName":"LSI1A","KeyType":"RANGE"},{"AttributeName":"LSI1B","KeyType":"RANGE"}],"Projection":{"ProjectionType":"ALL"}}]'
# 200 OK, "TableStatus": "ACTIVE" (real DynamoDB rejects this KeySchema shape for an LSI)
```

## Notes

I don't have access to a real AWS account to confirm the exact `ValidationException` message text for this specific case, so the expected behavior above is sourced from the developer guide rather than an observed AWS response. Whoever picks this up should confirm the precise error shape (likely similar to the "Member must have length less than or equal to 2" pattern floci's own `DynamoDbService.createTable` already uses for the base table's own `KeySchema`) against a real account before finalizing the fix.

[...]

PR #2463, opened 2026-08-22
fix(dynamodb): stop getSortKeyNames() corrupting persisted tables

### LINKED ISSUE #2415: [BUG] DynamoDB tables with a GSI sort key are never restored after restart, in every persistent storage mode
## Service

DynamoDB

## AWS API Action

`CreateTable` (GSI with a RANGE key), then `ListTables` / `BatchWriteItem` after a Floci restart.

## Expected behavior

A table created with a global secondary index that has a sort key should survive a Floci restart when storage is `hybrid`, `persistent`, or `wal`.

After `docker restart`:

- `ListTables` includes the table
- `Scan` / `BatchWriteItem` against it succeed
- other DynamoDB tables in the same instance are unaffected

A table with no GSI, or a GSI with only a HASH key, already behaves this way.

## Actual behavior

Any table whose GSI has a RANGE key makes the entire DynamoDB table store unreadable on startup. Floci logs a Jackson mapping error, discards the store, and comes up with **zero tables**. Subsequent API calls fail with `ResourceNotFoundException: Requested resource not found`.

The failure is not limited to the offending table: one GSI-with-sort-key takes down every table in the instance.

Reproduced with `FLOCI_STORAGE_MODE=hybrid`, `persistent`, and `wal` (the same `sortKeyNames` field is written in all three).

`persistent` also **renames** the snapshot to `dynamodb-tables.json.corrupt`, so the loss is permanent rather than recoverable from the original file.

Other services on the same volume (CloudFormation, Cognito, SSM, Lambda, API Gateway) restore correctly, so a stack can look healthy while pointing at a table that no longer exists.

Startup error (`persistent`):

```text
ERROR [io.git.hec.flo.cor.sto.PersistentStorage] Failed to load persisted data from /app/data/dynamodb-tables.json; moved the unreadable file to /app/data/dynamodb-tables.json.corrupt and started with an empty store. This stores state was lost.: com.fasterxml.jackson.databind.JsonMappingException: (was java.lang.UnsupportedOperationException) (through reference chain: java.util.LinkedHashMap["000000000000/us-east-1::con-gsi"]->io.github.hectorvent.floci.services.dynamodb.model.TableDefinition["globalSecondaryIndexes"]->java.util.ArrayList[0]->io.github.hectorvent.floci.services.dynamodb.model.GlobalSecondaryIndex["sortKeyNames"]->java.util.ImmutableCollections$ListN[1])
```

Same exception under the other backends:

```text
ERROR [io.git.hec.flo.cor.sto.HybridStorage] Failed to load data from /app/data/dynamodb-tables.json: ...
ERROR [io.git.hec.flo.cor.sto.WalStorage] Failed to load snapshot from /app/data/dynamodb-tables-snapshot.json: ...
```

The snapshot writes:

```json
"partitionKeyName": "GSI1PK",
"sortKeyName": "GSI1SK",
"sortKeyNames": ["GSI1SK"]
```

On load, Jackson has no setter for `sortKeyNames` and falls back to getter-as-setter, appending into the list the getter returns. That list is an immutable `List.of(...)`, so the append throws `UnsupportedOperationException`. When the index has no sort key the field serializes as `[]` and Jackson never appends, which is why those tables load.

Possible fixes: back the field with a mutable list (or a real `List<String>` setter), or mark it `@JsonIgnore` and derive it from `keySchema` — it is redundant with `sortKeyName`.

## Reproduction

```bash
docker run -d --name floci-test -p 4566:4566 -v floci-test-data:/app/data \
  -e FLOCI_STORAGE_MODE=persistent floci/floci:1.7.0

  --table-name con-gsi \
  --attribute-definitions \
    AttributeName=GSI1PK,AttributeType=S \
    AttributeName=GSI1SK,AttributeType=S \
  --key-schema AttributeName=PK,KeyType=HASH \
  --global-secondary-indexes '[{"IndexName":"GSI1","KeySchema":[{"AttributeName":"GSI1PK","KeyType":"HASH"},{"AttributeName":"GSI1SK","KeyType":"RANGE"}],"Projection":{"ProjectionType":"ALL"}}]'

aws --endpoint-url http://localhost:4566 dynamodb list-tables
# {"TableNames": ["con-gsi"]}

docker restart floci-test && sleep 10

aws --endpoint-url http://localhost:4566 dynamodb list-tables
# {"TableNames": []}

aws --endpoint-url http://localhost:4566 dynamodb scan --table-name con-gsi
# ResourceNotFoundException: Requested resource not found
```

Control: drop the RANGE entry from the GSI `KeySchema` (HASH-only index, or no GSI). The same restart then restores the table.

## Environment

- **Floci version / image tag:** `1.7.0` (`floci/floci:latest`, native image)
- **Java SDK version (if applicable):** n/a (reproduced with AWS CLI)
- **How you're running Floci (Docker / native / `mvn quarkus:dev`):** Docker (`floci/floci:1.7.0`)

omatheusmesmo:
## Summary

`GlobalSecondaryIndex`, `LocalSecondaryIndex`, and `TableDefinition` each expose `getSortKeyNames()` with no backing setter. Jackson's getter-as-setter fallback appends into the list it returns, which is an immutable `Stream.toList()`, throwing `UnsupportedOperationException` whenever the schema has a RANGE key. On startup, `PersistentStorage`, `HybridStorage`, and `WalStorage` each catch that as an unreadable file and reload with zero tables, taking down every table in the store rather than only the one whose schema triggered it. This fix marks `getSortKeyNames()` `@JsonIgnore` on all three classes, since the value is already derived from `keySchema` and was always redundant with `getSortKeyName()`. Closes #2415.

While reviewing the AWS documentation for the multi-attribute key schema feature this bug lives in (added in #2114), I found four further compatibility gaps in the same feature area that are out of scope for this fix and are tracked separately in #2459, #2460, #2461, and #2462.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Any table whose GSI or LSI had a sort key, or whose own primary key was composite (partition key plus sort key), failed to reload after a restart under `persistent`, `hybrid`, or `wal` storage. The `persistent` backend additionally renamed the on-disk snapshot to `.corrupt`, making the loss permanent rather than recoverable. Reproduced with the AWS CLI against a local floci instance both before and after the fix; the regression tests added here reproduce the exact stack trace from #2415 and confirm the restart now round-trips the table correctly.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/model/GlobalSecondaryIndex.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/model/GlobalSecondaryIndex.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.services.dynamodb.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import io.quarkus.runtime.annotations.RegisterForReflection;
 
@@ -82,7 +83,13 @@ public class GlobalSecondaryIndex {
     /**
      * Returns all sort key attribute names in key-schema order. For a composite sort key this
      * contains more than one element; ordering must consider all of them, not just the first.
+     *
+     * <p>{@code @JsonIgnore}d: it is derived from {@code keySchema} (redundant with
+     * {@link #getSortKeyName()}), and without a backing setter Jackson's getter-as-setter
+     * fallback tries to append into the immutable list this method returns, throwing
+     * {@code UnsupportedOperationException} on deserialization whenever the index has a sort key.
      */
+    @JsonIgnore
     public List<String> getSortKeyNames() {
         return keySchema.stream()
                 .filter(k -> "RANGE".equals(k.getKeyType()))
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/model/LocalSecondaryIndex.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/model/LocalSecondaryIndex.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.services.dynamodb.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import io.quarkus.runtime.annotations.RegisterForReflection;
 
@@ -81,7 +82,13 @@ public class LocalSecondaryIndex {
     /**
      * Returns all sort key attribute names in key-schema order. For a composite sort key this
      * contains more than one element; ordering must consider all of them, not just the first.
+     *
+     * <p>{@code @JsonIgnore}d: it is derived from {@code keySchema} (redundant with
+     * {@link #getSortKeyName()}), and without a backing setter Jackson's getter-as-setter
+     * fallback tries to append into the immutable list this method returns, throwing
+     * {@code UnsupportedOperationException} on deserialization whenever the index has a sort key.
      */
+    @JsonIgnore
     public List<String> getSortKeyNames() {
         return keySchema.stream()
                 .filter(k -> "RANGE".equals(k.getKeyType()))
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/model/TableDefinition.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/model/TableDefinition.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.services.dynamodb.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import io.github.hectorvent.floci.core.common.AwsArnUtils;
 import io.quarkus.runtime.annotations.RegisterForReflection;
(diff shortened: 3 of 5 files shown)


pgermosen (approved):
Verified the root cause is genuine: this is a well-known Jackson footgun — a getter with no backing setter whose returned collection is immutable (Stream.toList()) triggers Jackson's getter-as-setter fallback on deserialization, which tries to .add() into that immutable list. Verified load-bearing by reverting the @JsonIgnore on GlobalSecondaryIndex.getSortKeyNames(): reproduced the exact UnsupportedOperationException at ImmutableCollections.add via Jackson's StringCollectionDeserializer, precisely matching the described stack trace. Restored, confirmed clean.

Swept the codebase for the same pattern elsewhere — found one other candidate (KmsKeySpec.getAlgorithm) but confirmed it's a false positive (an enum, not a Jackson-deserialized POJO). No other instances of this footgun exist.

Test coverage is well-targeted: the Jackson round-trip test isolates the serialization bug directly, and DynamoDbGsiSortKeyRestartTest uses a real PersistentStorage backed by an actual file to reproduce the full production failure mode.

Full dynamodb suite passes (381/381), both on the branch and on the actual merged-with-main tree (compiled + tested the merge). LGTM.

PR #2464, opened 2026-08-22
feat(ses): support +label subaddressing on mailbox simulator addresses

okinaka:
## Summary

The SES mailbox simulator lets senders add a `+label` to distinguish test messages, so
`bounce+order-123@simulator.amazonses.com` triggers a bounce just like the bare
`bounce@simulator.amazonses.com`. Floci previously matched the simulator addresses by exact
(trimmed, case-insensitive) equality, so any labelled address was treated as a normal recipient and
emitted only the `Send` event.

`SimulatorAddresses` now strips a `+label` suffix from the local part before comparing the type
case-insensitively. Only `+` separates the label (a hyphen does not — `bounce-label@...` is not a
simulator address), the domain is left untouched, and a `+label` on a non-simulator domain does not
match. This affects `success`, `bounce`, `complaint`, and `suppressionlist`.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the real AWS SES mailbox simulator (SES v2 `SendEmail`, us-east-1): labelled
addresses (`bounce+mylabel`, `success+order-123`, empty label `bounce+`, upper-case type
`Bounce+Label`, dotted label `bounce+a.b.c`) are all accepted, matching the documented `+`
subaddressing. The `+`-before-label type rule follows the AWS mailbox-simulator documentation. Unit
tests cover the label, edge, and non-matching cases; the SES send/event integration tests pass
unchanged.

## Checklist

### DIFF
--- a/docs/services/ses.md
+++ b/docs/services/ses.md
@@ -271,6 +271,8 @@ Floci recognises the AWS [mailbox simulator addresses](https://docs.aws.amazon.c
 | `complaint@simulator.amazonses.com` | `Complaint` |
 | `suppressionlist@simulator.amazonses.com` | `Reject` |
 
+A `+label` subaddress is supported on any of these, so `bounce+order-123@simulator.amazonses.com` triggers a `Bounce` just like the bare address — the label lets senders distinguish test messages. Only `+` separates the label; `bounce-label@...` is not a simulator address.
+
 A successful send without a simulator-address recipient emits only the `Send` event.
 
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Verified the core claim against AWS's own SES mailbox simulator docs, not just the PR's testing notes: "The mailbox simulator supports labeling... you can send an email to bounce+label1@simulator.amazonses.com and bounce+label2@simulator.amazonses.com..." — matches exactly.

Traced `canonicalize` through all the edge cases the tests cover (empty label, uppercase type, dotted label, hyphen non-separator, non-simulator domain, multiple `+` signs, no-`@` input, null) — correct in each case, first `+` is the separator, domain untouched.

Reverted `canonicalize` to a no-op trim → `recognisesLabelledAddresses` failed with `expected: <true> but was: <false>`. Restored, green. Confirmed the fix reaches the real send path via `SesService.java`'s event-emission logic, not just the isolated unit test. Full `ses` suite passes clean on the branch and on a fresh merge against current `main`. `make docs-check` clean.

Good to merge.

PR #2465, opened 2026-08-22
feat(docs): guard the service matrix against undocumented services

### LINKED ISSUE #2454: Service matrix in docs/services/index.md is not checked against the catalog — four services have no row
## Problem

`docs/services/index.md` is the service matrix users browse to find out what Floci emulates, but nothing
enforces that a service registered in `ResolvedServiceCatalog` has a row there. Services can — and do —
ship without ever appearing.

Comparing the 85 `descriptor(...)` registrations in
`src/main/java/io/github/hectorvent/floci/core/common/ResolvedServiceCatalog.java` against the matrix,
these have no row:

| service | catalog key | notes |
|---|---|---|
| MWAA | `mwaa` | has `docs/services/mwaa.md` and a `mkdocs.yml` nav entry |
| SWF | `swf` | |
| Cloud Control API | `cloudcontrol` | |
| EMR Serverless | `emr-serverless` | added in #2403; `EMR` has a row, EMR Serverless doesn't |

Bedrock AgentCore was a fifth until #2436, which is what prompted this — it shipped in #2316 on 15 Aug
and stayed undiscoverable for five days.

Most of the other apparent gaps in that comparison are false positives: AWS signing names such as
`monitoring`, `events`, `servicediscovery`, `elasticloadbalancing` and `cognito-idp` map to rows that
already exist under their product names.

## Why it keeps happening

`make docs-check` already guards the generated **Supported Actions** tables, and CONTRIBUTING.md
describes the sentinel that fires when a new handler is *"neither registered in `tools/docs/services.yaml`
nor listed under `deferred_handlers`"*. But that sentinel only covers switch-mode handlers, and it says
nothing about the matrix in `docs/services/index.md`. A REST-controller service can therefore be
registered, enabled, routable and completely absent from the docs with every check green.

## Suggested fix

Extend the docs tooling with a check that fails when a catalog descriptor has no corresponding row in
`docs/services/index.md`, with an explicit opt-out list for the signing-name aliases above. The catalog
is already parsed by nothing today, but the `descriptor("key", …)` calls are regular enough to read with
the same approach `tools/docs/regen_action_docs.py` uses for handler sources.

That turns "did you remember to document it?" from a review question into a CI one, which is the same
move the action tables already made.

## Context

Found while reviewing #2436, which documents AgentCore. Deliberately kept out of that PR — the four
services above are unrelated to it and each needs its own row content.

omatheusmesmo:
## Summary

`docs/services/index.md` is the service matrix users browse to find out what Floci emulates, but nothing enforced that a service registered in `ResolvedServiceCatalog` had a row there. Four services shipped without one: MWAA, SWF, Cloud Control API, and EMR Serverless (EMR Serverless was also missing from the `mkdocs.yml` nav).

This PR:

- Adds the four missing rows to the matrix, and a new `docs/services/cloudcontrol.md` page (the service had no doc page at all).
- Adds `tools/docs/check_service_matrix.py`, a CI check (wired into `make docs-check`) that compares every `descriptor(...)` registration in `ResolvedServiceCatalog.java` against the matrix, resolving each one via an exact filename match, an alias in `tools/docs/service_matrix.yaml` (for keys whose AWS signing name differs from their doc filename), or a time-boxed `deferred` entry. Deferred entries require a `by:` expiry date so the exemption can't turn into a silent, permanent bypass (which is how Bedrock AgentCore shipped undocumented for five days before #2436).

Closes #2454

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A: docs and CI tooling only, no runtime or wire-protocol change.

## Checklist

- [x] No Java source changed; `./mvnw test` is unaffected by this PR
- [x] `make docs-test` and `make docs-check` pass locally (31 tests, both gates green)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,9 @@
 #
 # Action-table docs: docs/services/*.md "Supported Actions" tables are generated
 # from handler source. See tools/docs/.
+#
+# Service-matrix docs: docs/services/index.md's Service Matrix table is checked
+# against ResolvedServiceCatalog.java so a registered service can't ship undocumented.
 
 PYTHON ?= python3
 
@@ -10,7 +13,7 @@ PYTHON ?= python3
 docs-sync: ## Regenerate the action tables in docs/services from handler source (in place)
 	$(PYTHON) tools/docs/regen_action_docs.py
 
-docs-check: ## CI gate: regenerate and fail if anything is stale or a handler is unregistered
+docs-check: ## CI gate: regenerate and fail if anything is stale, unregistered, or undocumented
 	@$(PYTHON) tools/docs/regen_action_docs.py --strict || { \
 		echo ""; \
 		echo "error: action-table regeneration reported problems (see warnings above)."; \
@@ -22,6 +25,11 @@ docs-check: ## CI gate: regenerate and fail if anything is stale or a handler is
 		echo "       Run 'make docs-sync' and commit the result."; \
 		exit 1; \
 	}
+	@$(PYTHON) tools/docs/check_service_matrix.py --strict || { \
+		echo ""; \
+		echo "error: a registered service has no row in docs/services/index.md (see warnings above)."; \
+		exit 1; \
+	}
 
-docs-test: ## Run the action-table tooling tests
+docs-test: ## Run the docs tooling's unit tests
 	$(PYTHON) -m pytest tools/docs -q
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ LocalStack's community edition [sunset in March 2026](https://blog.localstack.cl
 | CodeBuild | Real Docker execution | No |
 | Native binary | ~40 MB | No |
 
-**73 AWS services. Broad coverage. Free forever.**
+**76 AWS services. Broad coverage. Free forever.**
 
 ## Architecture Overview
 
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ LocalStack's community edition [sunset in March 2026](https://blog.localstack.cl
 | CodeBuild | Real Docker execution | No |
 | Native binary | ~40 MB | No |
 
-**72 AWS services. Broad coverage. Free forever.**
+**76 AWS services. Broad coverage. Free forever.**
 
 ## Architecture Overview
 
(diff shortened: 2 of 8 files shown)


pgermosen:
This is in good shape substantively — both prior findings check out fixed:

- SWF's operation count is now `39`, and I confirmed that independently against `swf.md`'s own prose ("All 39 modeled operations are implemented"), not just the matrix row.
- The literal-only-extraction gap is closed by the call-site-count safeguard in `check_service_matrix.py`. I reverted it locally and confirmed `test_count_descriptor_calls_catches_a_non_literal_key` fails without it (`1 == 2`), then restored and confirmed all 14 unit tests pass again.

I also verified the two other new rows against their own doc pages rather than trusting the matrix: Cloud Control API's 5 operations and EMR Serverless's 7 endpoints both match exactly. `make docs-check` and `make docs-test` are clean (34 tests), no CODEOWNERS paths touched.

**This needs a rebase before it can merge, though.** `git merge --no-commit --no-ff origin/main` produces real conflicts in `README.md` and `docs/services/index.md` — #2496 merged earlier today and already added its own MWAA row plus bumped the count to 73 on `main`. This branch predates that and independently adds a (differently-worded) MWAA row targeting 76.

Could you rebase onto current `main`, drop the now-redundant MWAA row/count-bump from this branch (main already has it), and keep the SWF + Cloud Control API + EMR Serverless additions? The target total of 76 stays correct either way — 73 already on `main` + 3 new here = 76. Happy to re-review once that's done.

omatheusmesmo:
@pgermosen Rebased onto current `main` and dropped the duplicate MWAA row/count-bump (already on `main` via #2496), keeping SWF, Cloud Control API, and EMR Serverless. `make docs-check` and `make docs-test` (34 tests) are clean. Ready for another look.

pgermosen (approved):
Verified the rebase independently rather than taking the claim on trust: MWAA is fully removed from this branch (no references anywhere in the diff), the count is correctly 76 (73 already on `main` via #2496 + SWF/Cloud Control API/EMR Serverless = 3 new here), and the merge against current `main` is clean. Both previously-verified fixes (SWF's `39` count, the call-site-count safeguard in `check_service_matrix.py`) are untouched by the rebase and still intact.

`make docs-check` and `make docs-test` (34 tests) both clean. All CI green.

Good to merge.

PR #2471, opened 2026-08-22
fix(dynamodb): enforce composite query order

### LINKED ISSUE #2462: [BUG] DynamoDB: Query on a composite GSI sort key does not enforce left-to-right / inequality-last ordering

thomhurst:
## Summary

- enforce left-to-right conditions for multi-attribute GSI sort keys
- require equality on every conditioned sort-key component before the final component
- apply identical validation to expression and legacy `KeyConditions` query forms

Closes #2462

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Floci returned HTTP 200 when a Query skipped an earlier composite sort-key component, or used an inequality on it before conditioning a later component. DynamoDB Local 3.3.1 rejects both with `ValidationException`. This change matches that left-to-right rule and validation message for modern expressions and legacy `KeyConditions`.

## Tests

- `./mvnw.cmd test "-Dtest=DynamoDbAccessPathValidatorTest,DynamoDbCompositeSortKeyQueryIntegrationTest,DynamoDbServiceTest"` (151 passed after rebase)
- `./mvnw.cmd -f compatibility-tests/sdk-test-java/pom.xml test-compile -DskipTests` (passed after rebase)
- SDK compatibility test against live Floci before rebase (10 passed)

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbAccessPathValidationTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbAccessPathValidationTest.java
@@ -186,6 +186,7 @@ class DynamoDbAccessPathValidationTest {
     }
 
     @Test
+    @SuppressWarnings("deprecation")
     void queryRejectsInvalidCompositeSortKeyConditions() {
         assertValidationException(() -> ddb.query(request -> request
                 .tableName(TABLE)
@@ -206,6 +207,21 @@ class DynamoDbAccessPathValidationTest {
                         ":status", value("open"),
                         ":createdAt", value("2026-01-01"),
                         ":alternate", value("a1")))));
+
+        assertValidationException(() -> ddb.query(request -> request
+                .tableName(TABLE)
+                .indexName("status-index")
+                .keyConditions(Map.of(
+                        "status", condition(ComparisonOperator.EQ, "open"),
+                        "alternate", condition(ComparisonOperator.EQ, "a1")))));
+
+        assertValidationException(() -> ddb.query(request -> request
+                .tableName(TABLE)
+                .indexName("status-index")
+                .keyConditions(Map.of(
+                        "status", condition(ComparisonOperator.EQ, "open"),
+                        "createdAt", condition(ComparisonOperator.GT, "2026-01-01"),
+                        "alternate", condition(ComparisonOperator.EQ, "a1")))));
     }
 
     private static void assertValidationException(org.assertj.core.api.ThrowableAssert.ThrowingCallable call) {
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbAccessPathValidationTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbAccessPathValidationTest.java
@@ -41,7 +41,8 @@ class DynamoDbAccessPathValidationTest {
                         attribute("createdAt"), attribute("alternate"))
                 .globalSecondaryIndexes(GlobalSecondaryIndex.builder()
                         .indexName("status-index")
-                        .keySchema(key("status", KeyType.HASH), key("createdAt", KeyType.RANGE))
+                        .keySchema(key("status", KeyType.HASH), key("createdAt", KeyType.RANGE),
+                                key("alternate", KeyType.RANGE))
                         .projection(Projection.builder()
                                 .projectionType(ProjectionType.INCLUDE)
                                 .nonKeyAttributes("summary")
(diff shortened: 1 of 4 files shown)


pgermosen:
Thanks for this one — nicely scoped fix. Reviewed the logic and I think it's correct:

- Traced `validateCompositeSortKeyConditions` by hand against the schema-ordered `sortKeys` list and the per-attribute equality map: a later component with any condition on it requires every earlier component to have an equality condition, which correctly covers both the "skipped middle component" case and the "inequality followed by another condition" case, and correctly still allows a trailing inequality/`BETWEEN` on the last conditioned component. Applied consistently to both `validateKeyConditionExpression` (modern) and the legacy `KeyConditions` path via the shared helper, so the two forms can't disagree — good attention to that.
- Did an independent revert-and-confirm-fail rather than trusting the PR description's own claim: disabled `validateCompositeSortKeyConditions` and all three relevant tests (`rejectsSkippedCompositeSortKey`, `rejectsCompositeSortKeyConditionAfterInequality`, `validatesLegacyCompositeSortKeyOrder`) failed with exactly the expected symptom ("Expected AwsException to be thrown, but nothing was thrown"). Restored, green.
- Full `dynamodb` package passes clean, including on a fresh merge against current `main` (no conflicts). Title and commits are clean (Conventional Commits, no AI co-authorship trailers). All CI green.

One honesty note on verification: I tried to independently confirm the exact validation message text ("RANGE key attributes X must have equality conditions specified...") against AWS's own documentation and couldn't find a primary source that quotes it verbatim — the left-to-right / equality-until-last-inequality *rule* for composite sort keys is well established, but I couldn't pin the exact string. Your description says this was verified against DynamoDB Local 3.3.1, which is a reasonable way to check parity for a fairly new DynamoDB feature (multi-attribute GSI sort keys) that doesn't have much public documentation yet — I just want to be upfront that I'm taking that specific claim on your testing rather than having verified it myself against a primary source.

Separately, and unrelated to the code: this PR's description doesn't follow the repo's PR template (no `## Type of change` checkboxes, no `## Checklist` — though you did include `Closes #2462`, which is good). Not asking you to rewrite it, just flagging it in case that matters for how it gets tracked.

Everything else looks ready to me — nice work.

hectorvent:
Thanks for the clean rebase. I checked the rebased branch keeps everything intact: the unit tests now call the widened eight argument `validateQuery` from #2469, the compatibility test file merged cleanly over #2470, and #2474's rule that every HASH attribute carries an equality condition still applies beneath the new composite ordering check. CI is green across the board at `a88cbb5`. pgermosen's review above covers the correctness verification exactly right, so just seconding it.

No blockers from my side.

hectorvent (approved):
Approved, see comment above.

PR #2472, opened 2026-08-22
feat(redshift): implement Redshift emulation with snapshot and parameter group support

vuthehuyht:
## Summary

Implements AWS Redshift emulation for Floci, including both the Management API (Control Plane) and Data Plane via Docker.

**Phase 1 Core Management API:**
- `CreateCluster`, `DescribeClusters`, `DeleteCluster`
- Data Plane: spins up a `postgres:15-alpine` container per cluster (same pattern as RDS)
- Wired through `AwsQueryController` -> `RedshiftQueryHandler` -> `RedshiftService`
- Storage via `StorageFactory` / `AccountAwareStorageBackend`

**Phase 2 Operations (Snapshots & Parameter Groups):**
- `CreateClusterSnapshot`, `DescribeClusterSnapshots`, `DeleteClusterSnapshot`, `RestoreFromClusterSnapshot`
  - Real backup/restore using `pg_dump` and `psql` via Docker Exec API
  - SQL dumps streamed directly to/from disk (no in-memory buffering)
- `CreateClusterParameterGroup`, `DescribeClusterParameterGroups`, `DescribeClusterParameters`, `DeleteClusterParameterGroup`
  - Mock state - stores metadata, returns standard parameter list

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against **AWS SDK for Java v2** (`software.amazon.awssdk:redshift`):
- `CreateCluster` / `DescribeClusters` / `DeleteCluster`
- `CreateClusterSnapshot` / `DescribeClusterSnapshots` / `RestoreFromClusterSnapshot`
- `CreateClusterParameterGroup` / `DescribeClusterParameterGroups`

Wire protocol: **AWS Query Protocol** (form-encoded POST + `Action` param, XML response) - consistent with existing SQS, SNS, RDS implementations.

Data Plane verified via JDBC (`org.postgresql:postgresql`) - create table, insert, snapshot, restore, query on restored cluster.

## Checklist

- [x] New or updated integration test added (`RedshiftServiceTest`, `RedshiftQueryHandlerTest`, `RedshiftContainerManagerTest`, `RedshiftOperationsTest`, `compatibility-tests/.../RedshiftOperationsTest`)


pgermosen (commented):
Thanks — a whole new service with a real data plane is a lot of ground, so a scope note first: at
~1,700 lines of main code I read `RedshiftContainerManager` and `RedshiftService` closely, with
attention to the process execution, the credential paths and the restart-recovery logic, since
those carry the risk that a container-backed service adds over a metadata-only one. I skimmed the
models and did not read all ~2,250 lines of tests line by line. Everything below comes from that
read.

The bot findings from the 22nd have largely landed, and two of them landed well — see the notes at
the end.

## (should fix) A comment in `RedshiftContainerManager` isn't in English

In `alterUserPassword`, the failure branch carries:

```java
// KHÔNG log/đưa result.stderr() vào message — psql có thể echo lại câu lệnh
// ALTER USER lỗi (kèm password literal) trong context dòng lỗi, rò rỉ secret.
```

The rest of this file, and the repository, is in English. What makes this worth fixing rather than
leaving is that the comment is *good* — it records that `result.stderr()` is deliberately kept out
of both the log line and the exception message, because psql can echo the failing `ALTER USER`
statement back with the password literal in it. That is a non-obvious reason for what otherwise
looks like an unhelpfully vague error, and the next person to "improve" the diagnostics by adding
`stderr` to the message needs to be able to read it.

Something like: *"Deliberately omit result.stderr() from the log and the exception: psql echoes the
failing statement, which contains the password literal."*

## (should fix) The single-quote rejection is AWS-correct — but the comment says otherwise

```java
// The password is embedded in a SQL literal executed via psql -c, not shell-interpolated —
// ' inside the password would break the SQL literal, so reject it rather than risk injection.
if (newPassword != null && newPassword.contains("'")) {
    throw new AwsException("InvalidParameterValue", "MasterUserPassword must not contain a single quote", 400);
}
```

I went looking for the injection this guards against, expecting to file it — the username check
above it is thorough, and this one reads like the weaker sibling. It isn't. `ModifyCluster`'s
`MasterUserPassword` constraints say:

> + Must be between 8 and 64 characters in length.
> + Must contain at least one uppercase letter.
> + Must contain at least one lowercase letter.
> + Must contain one number.
> + **Can be any printable ASCII character (ASCII code 33-126) except `'` (single quote), `"`
>   (double quote), `\`, `/`, or `@`.**

So rejecting `'` isn't a workaround for the emulator's SQL construction — it is exactly what AWS
does, and `InvalidParameterValue` is the right shape for it.

That matters because of how the comment reads. As written it frames the rejection as a limitation
imposed by the implementation, which invites exactly the wrong follow-up: someone escapes the quote
properly, removes the check, and Floci starts accepting a password real Redshift refuses. Citing
the constraint instead makes the guard permanent.

While you're there: AWS also rejects `"`, `\`, `/` and `@`, and enforces the 8–64 length and the
upper/lower/digit rules. None of those are checked. That's the accepts-more-than-AWS direction so I
wouldn't hold the PR for it — but if the file is already validating one character on
injection-safety grounds, validating the documented set on parity grounds is barely more code and
stops the two rationales drifting apart.

## (minor) The branch that loses data is the quiet one

`adoptOrStart` logs when it succeeds at preserving data:

```java
LOG.infov("Adopting existing container {0} for cluster {1} to avoid discarding its data", …);
```

and says nothing on the path that doesn't:

```java
if (existing.isEmpty()) {
    return start(accountId, clusterIdentifier, masterUsername, masterPassword);
}
```

Adopting a surviving container covers the ordinary Floci restart, which is the important half and a
real improvement over the original recovery. But when the container is genuinely gone — a machine
reboot, a `docker rm`, a pruned dev box — recovery starts an empty database and `onStart` then
marks the cluster `available`. A caller sees a healthy cluster with no data and nothing in the log
explaining it.

One `LOG.warnv` on the `existing.isEmpty()` branch, saying a fresh empty container was started and
the previous contents aren't recoverable, turns a silent surprise into a searchable line. I'm not
asking for dump-on-shutdown — that's a much bigger change and the snapshot API already gives users
an explicit way to preserve data.

## (nit) `.superpowers/` in the shared `.gitignore`

```
# Subagent-driven-development scratch workspace (ledgers, briefs, review packages)
.superpowers/
```

That's a scratch directory for tooling you run locally, not something the project produces.
`.git/info/exclude`, or a global gitignore via `core.excludesFile`, keeps it out of everyone else's
tree — a shared `.gitignore` entry means every contributor carries a rule for a tool they may not
use. Small either way, and entirely your call.

## Notes

- **The account-scoping fixes landed properly.** Recovery now walks `scanAllAccountEntries`, keys
  containers by `containerKey(accountId, clusterIdentifier)` and writes back with `putForAccount`,
  so two accounts holding a same-named cluster no longer collapse onto one runtime. That was the
  sharpest of the original findings and it's closed at the right layer rather than patched at the
  call site.
- **The restore credential chain is now a real chain**, not a constant: snapshot's own stored
  password, then the source cluster's, then `"admin"`. Since `Snapshot` carries `masterPassword`,
  the fallback is only reachable for a snapshot taken before that field existed — which is the
  right shape for a compatibility fallback, and a big improvement on the hardcoded value the bot
  originally found.
- The username check in `alterUserPassword` is the right kind of strict: an identifier pattern
  rather than an escaping attempt, with a comment explaining that `psql -c` accepts multiple
  `;`-separated statements so argv-escaping alone wouldn't save it. That reasoning is what makes
  the check defensible rather than superstitious.

vuthehuyht:
Thanks for the close read. All four addressed in 7f8262de:

Non-English comment (should fix). Fixed with your suggested wording. Grepping for the same slip turned up two more Vietnamese blocks in RedshiftService.modifyCluster that your review didn't reach — translated those too, so the feature is now English throughout.

Single-quote rejection comment (should fix). Reworded to cite the ModifyCluster MasterUserPassword constraint instead of the psql SQL-literal mechanics, so the guard reads as permanent rather than a workaround. Went the extra step on parity while there: alterUserPassword now rejects the full documented set (' " \ / @) and enforces the 8–64 length plus the upper/lower/digit rules, all as InvalidParameterValue. Existing tests that used a non-compliant literal were updated; new cases cover each rejection.

Quiet data-loss branch (minor). Added a LOG.warnv on the existing.isEmpty() path in adoptOrStart stating a fresh empty container was started and the previous contents aren't recoverable. No dump-on-shutdown — agreed the snapshot API is the right tool for that.

.superpowers/ in .gitignore (nit). Moved to .git/info/exclude.

Same commit also greens the CI on this branch:

redshift.model.Parameter was missing @RegisterForReflection — added it (PersistedModelReflectionTest, and the native sdk-test-java run).
Converted RedshiftQueryHandler.handle() from an if/else chain to switch (action) so the docs action-table generator can extract the action list like every other Query handler (bodies unchanged), then registered the service in tools/docs/services.yaml, added docs/services/redshift.md, and added the Service Matrix row.

pgermosen (approved):
Approving. Everything from my last round is closed, including the optional password-validation parity I flagged as non-blocking. The latest commit also fixes a real resource-leak: a rejected restore used to leave a half-created cluster and a running container behind, now the dump-path check runs before any provisioning. isTrustedDumpPath keeps snapshot restores confined to the calling account's own dump directory, same account-scoping discipline as the earlier rounds. Checked RebootCluster and the cluster ARN format against real AWS docs and both line up. Couldn't get an authoritative answer on the exact ARN segment AWS uses for snapshot/parametergroup/subnetgroup tagging (the reference pages wouldn't render), and createTags has no validation on AWS's documented key/value rules yet — neither is blocking, tagging is new and low-severity if either turns out wrong, but worth a look in a follow-up.

PR #2473, opened 2026-08-22
fix(cloudformation): serialize concurrent stack operations

### LINKED ISSUE #2419: [BUG] cloudformation: Stack's collections are mutated concurrently without synchronization
## Service

CloudFormation

## AWS API Action

Any pair of operations that touch the same stack concurrently — `CreateChangeSet`, `ExecuteChangeSet`, `UpdateStack`, `DeleteStack`, `DescribeStackEvents`, `DescribeStackResources`.

## Summary

`Stack`'s collections are plain, non-thread-safe `LinkedHashMap`/`ArrayList` instances, but a `Stack` is shared state: it lives in the `stacks` `ConcurrentHashMap` and is handed out to every request that names it. `stacks` being concurrent protects the *map*, not the `Stack` objects inside it, and almost all mutation happens outside any lock.

https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/cloudformation/model/Stack.java#L20-L31

```java
private Map<String, String> parameters      = new LinkedHashMap<>();
private Map<String, String> outputs         = new LinkedHashMap<>();
private Map<String, String> exports         = new LinkedHashMap<>();
private Map<String, String> outputExportNames = new LinkedHashMap<>();
private Map<String, StackResource> resources = new LinkedHashMap<>();
private List<StackEvent> events             = new ArrayList<>();
private Map<String, ChangeSet> changeSets   = new LinkedHashMap<>();
private Map<String, String> tags            = new LinkedHashMap<>();
```

## Expected behavior

Two concurrent operations on one stack either serialize or fail cleanly. An acknowledged write (a recorded event, a provisioned resource, an accepted change set) is never silently lost, and a concurrent reader never sees a corrupted collection.

## Actual behavior

Unsynchronized concurrent mutation. Two threads writing a `LinkedHashMap` can lose an entry or break its links; `ArrayList.add` can lose an element or throw `ArrayIndexOutOfBoundsException`; a reader iterating during a write can get `ConcurrentModificationException`.

Concrete sites, all in `CloudFormationService`:

| Collection | Mutation | Sites |
|---|---|---|
| `events` | `stack.getEvents().add(...)` via `addEvent` | **38** call sites |
| `resources` | `getResources().put/remove` | 5 |
| `changeSets` | `getChangeSets().remove` | 1 (`DeleteChangeSet`) |
| `outputs` / `exports` / `outputExportNames` | rebuilt during execute | several |

`events` is the widest exposure: `addEvent` is called 38 times across provisioning, and a long `ExecuteChangeSet` appends to it throughout while a concurrent `DescribeStackEvents` iterates the same `ArrayList`.

## Reproduction

Hard to trigger deterministically — this is a data race, not a logic bug, and the window is narrow relative to HTTP round-trip time. It is a correctness-by-construction problem rather than one a test reliably catches. The clearest statement of the hazard is the code itself: shared mutable `Stack`, unsynchronized collections, no lock held across the mutations.

## Scope note

`CreateChangeSet`'s insertion into `changeSets` was moved inside the `stacks.compute()` per-key lock in #2365, which covers that one path. Everything in the table above is still outside any lock, so the general problem stands — #2365 deliberately did not widen into it.

## Possible directions

Roughly in increasing order of cost:

1. **Swap the field types** — `ConcurrentHashMap` for the maps, `CopyOnWriteArrayList` for `events`. Cheapest, and the same fix already applied to `MskConfiguration` in #2344. Caveat: `LinkedHashMap` insertion order is currently relied on for response ordering, and `ConcurrentHashMap` does not preserve it — `events` and `changeSets` would need an explicit sort, and `outputs`/`parameters` ordering would need checking against what the XML responses assume.
2. **A per-stack lock** — hold it across each operation's whole mutation sequence. Stronger, since it makes multi-field updates (execute writing `resources` + `outputs` + `events` together) atomic rather than just individually safe, which the collection swap does not.
3. **Copy-on-write at the storage boundary** — mutate a private copy and swap it into `stacks` atomically. Biggest change; also fixes readers seeing half-applied updates.

(2) is probably the right shape given how much of the execute path mutates several fields together, but (1) is a real improvement on its own and is a much smaller diff.

## Environment

- Floci version: `main`
- Found while reviewing #2365 (raised there by Greptile against the change-set path; filed separately because the pattern is repo-wide within this service, not specific to that PR).

AbarnaaSree:
## Summary

Closes #2419

Fix CloudFormation stack concurrency issues by adding per-stack locking for operations that read or mutate stack state.

- Serialize concurrent operations on the same stack and region.
- Prevent concurrent `CreateStack` requests from both provisioning the same stack.
- Protect concurrent ChangeSet operations from losing entries or corrupting the ChangeSet map.
- Serialize ChangeSet execution and stack deletion resource operations.
- Protect stack event and resource reads from concurrent mutations.
- Re-resolve the stack after acquiring the lock in `DeleteStack` to avoid stale state.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behavior was caused by concurrent requests accessing and mutating the same CloudFormation stack without synchronization. This could result in duplicate provisioning, lost ChangeSets, or inconsistent stack state.

The fix uses a per-stack, per-region lock so operations on different stacks can still execute concurrently while operations targeting the same stack are serialized.

## Checklist


hectorvent:
Thanks for picking up #2419, the per-stack lock is the right shape for it.

**(blocking)** The executor task holds the stack monitor for the whole `executeTemplate` / `deleteStackResources` run, and `describeStackEvents` / `describeStackResources` take the same monitor. Container-backed resources (RDS, ElastiCache, Lambda) take minutes, so a CDK or SAM deploy polling DescribeStackEvents would block until provisioning ends. Locking only the mutation sites and the snapshot copies (or concurrent collections, option 1 in the issue) keeps reads live.

**(blocking)** `createChangeSet` still mutates `changeSets` and appends events inside `stacks.compute()` (`services/cloudformation/CloudFormationService.java:240`), a different lock from `stackLocks`, so CreateChangeSet racing DeleteChangeSet is still unsynchronized. Greptile is also right that `executeChangeSet` resolves the stack before locking; the re-resolve you added to `deleteStack` fits there too.

**(follow-up, separate PR)** The checklist ticks "integration test added" but the diff has no test; a short note is fine if the race is hard to pin down.

Those two and this is good to go.

AbarnaaSree:
Addressed the two blocking comments. The long-running executeTemplate/deleteStackResources operations no longer hold the per-stack monitor, while mutation sites and snapshots remain synchronized. createChangeSet now uses the same per-stack lock, and executeChangeSet re-resolves the stack after acquiring the lock.

AbarnaaSree:
Addressed the two blocking comments:

Long-running executeTemplate / deleteStackResources operations no longer hold the per-stack lock; only snapshots and state mutations are synchronized.
createChangeSet now uses the same per-stack lock as other ChangeSet operations, and executeChangeSet re-resolves the stack after acquiring the lock.

Also added persistStackIfCurrent to prevent stale async operations from overwriting the current persisted stack state. Local compilation passes with .\mvnw -DskipTests compile.

pgermosen:
Thanks for tackling CloudFormation's stack concurrency — this is a real problem area and the per-stack lock approach is the right shape. Checked the current code directly rather than relying on the earlier review thread.

**The build-breaking issue is genuinely fixed**: `persistStackIfCurrent` is now defined and used consistently at all the sites that reference it, and the project compiles clean.

**One thing still needs attention, and it's the core of what this PR is trying to fix:**

## (blocking) The lock is released before the actual serialized work happens

Both `executeChangeSet` and `deleteStack` follow the same shape: acquire the per-stack lock, do the *synchronous* status-transition-and-persist step, then `executor.submit(...)` the real work (`executeTemplate` / `deleteStackResources`) and return — releasing the lock as soon as `submit()` returns, which is before the submitted task has even started running.

So the sequence that's still possible:

1. `ExecuteChangeSet` acquires the lock, marks the stack `UPDATE_IN_PROGRESS`, persists, submits `executeTemplate` to the executor, releases the lock.
2. `DeleteStack` comes in concurrently, acquires the now-free lock, marks the stack `DELETE_IN_PROGRESS`, persists, submits `deleteStackResources`, releases the lock.
3. Both `executeTemplate` and `deleteStackResources` are now running on separate executor threads, genuinely concurrently, for the same stack.

I checked whether the async methods reacquire the lock for their full duration — they don't. `executeTemplate` takes 7 separate short `synchronized(getStackLock(...))` blocks scattered around individual state mutations, and `deleteStackResources` takes only 1. So the two operations' actual resource work (provisioning vs. tearing down) can freely interleave in the gaps between those short windows — one thread could be provisioning a resource while the other is deleting it out from under it.

`persistStackIfCurrent`'s identity-check guard is a real improvement (it stops a genuinely-stale object reference from clobbering later state on the *final* write), but it only protects that last write — it doesn't prevent the interleaved provisioning/teardown from happening in the first place, which is what the PR description says it's meant to serialize ("Serialize ChangeSet execution and stack deletion resource operations").

A fix here likely means holding the per-stack lock for the *duration* of the submitted work, not just the setup — e.g. having the submitted task itself wrap its execution in `synchronized (getStackLock(...))`, or restructuring so `executor.submit` isn't called until after all the stack-mutating work (sync and async) is done. Happy to think through the shape of that with you if useful.

## Also worth knowing

- This has a real merge conflict against current `main` in `CloudFormationService.java` — worth a rebase before this gets much further along, since the file is large and actively changing.

Full CloudFormation suite passes (650/650) as-is — nothing here shows up as a test failure since there's no test that currently exercises the interleaved-async-work scenario. Happy to take another look once the lock scope covers the async work too.

AbarnaaSree:
Thanks for the detailed review. You're right — the earlier implementation only held the per-stack lock during the synchronous setup, so the actual async `executeTemplate` / `deleteStackResources` work could still run concurrently.

I've updated the implementation in commit `4e992339` (`fix(cloudformation): serialize async stack operations`) so that the submitted async task acquires the same per-stack lock for the duration of the actual stack operation. This keeps the resource provisioning/deletion work serialized for a given stack.

I'll also rebase against the current `main` to resolve the `CloudFormationService.java` conflict and then rerun the CloudFormation test suite.

Thanks again for catching the lock-scope issue.

hectorvent:
Round 2 landed the right pieces: short locks at the mutation sites, `createChangeSet` on the shared per-stack lock, the `executeChangeSet` re-resolve, and the `persistStackIfCurrent` guard. Thanks for the quick turnarounds. Two things before this lands.

**(blocking)** The last push replayed about 25 main commits with rewritten SHAs, so the PR shows 145 files and a conflict, and CI cannot run while it conflicts. Rebuild on current main; conflicts should stay inside `services/cloudformation/CloudFormationService.java`:

```
git fetch https://github.com/floci-io/floci.git main
git checkout -b cfn-clean FETCH_HEAD
git cherry-pick 695245ab5^..b5f7b8ac0 83fe5fc6d 4e992339e
git push --force-with-lease origin cfn-clean:fix/cloudformation-stack-concurrency-2419
```

Force-pushing keeps this PR and its thread; a fresh PR is fine if the cherry-picks fight you.

**(blocking)** Wrapping the whole submitted task in the stack monitor serializes execution, but `describeStacks`, `describeStackEvents` and `describeStackResources` take that same monitor, so a CDK or SAM deploy polling events hangs until provisioning ends. Use two locks per stack: keep the current short monitor for state reads and mutations, and add a per-stack execution lock held only by the submitted `executeTemplate` / `deleteStackResources` bodies. Readers never wait on provisioning, and execution still serializes. Heads up that #2364 reworks these same paths and is in review, so one more rebase may follow.

Happy to re-review right after.

hectorvent:
@AbarnaaSree closing, please file a new PR focused on the required changes. 
Thanks!

PR #2474, opened 2026-08-22
fix(dynamodb): recognize every HASH attribute of a composite GSI partition key

### LINKED ISSUE #2461: [BUG] DynamoDB: GSI composite partition key ignores every HASH attribute but the first

omatheusmesmo:
## Summary

- `DynamoDbAccessPath` gains `partitionKeyNames()` (all HASH attributes, in key-schema order)
- Validator now requires equality on EVERY HASH attribute of a composite GSI partition key (any order)
- Legacy `KeyConditions` execution path now checks every HASH attribute, not just the first

Closes #2461

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

A GSI with a composite partition key was broken: querying with only the first HASH attribute returned items from any partition; querying with all HASH attributes (the correct form) was rejected as "not supported".

## Checklist

- [x] New tests added (validator + execution, expression and legacy paths)
- [x] Conventional Commits

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbAccessPath.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbAccessPath.java
@@ -43,11 +43,14 @@ record DynamoDbAccessPath(String indexName, Kind kind, List<KeySchemaElement> ke
     }
 
     String partitionKeyName() {
+        return partitionKeyNames().getFirst();
+    }
+
+    List<String> partitionKeyNames() {
         return keySchema.stream()
                 .filter(key -> "HASH".equals(key.getKeyType()))
                 .map(KeySchemaElement::getAttributeName)
-                .findFirst()
-                .orElseThrow();
+                .toList();
     }
 
     List<String> sortKeyNames() {
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbAccessPathValidator.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbAccessPathValidator.java
@@ -89,9 +89,10 @@ final class DynamoDbAccessPathValidator {
 
         List<Expr> conditions = root instanceof AndExpr and
                 ? and.operands() : List.of(root);
-        String partitionKey = accessPath.partitionKeyName();
+        List<String> partitionKeys = accessPath.partitionKeyNames();
+        Set<String> partitionKeySet = Set.copyOf(partitionKeys);
         Set<String> sortKeys = Set.copyOf(accessPath.sortKeyNames());
-        int partitionConditions = 0;
+        Set<String> conditionedPartitionKeys = new HashSet<>();
         String partitionKeyValuePlaceholder = null;
         Set<String> conditionedAttributes = new HashSet<>();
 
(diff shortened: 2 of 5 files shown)


hectorvent:
Thank you, this closes the gap from #2461 properly: the validator now requires an equality on every HASH attribute in any order, the legacy `KeyConditions` path matches on all of them, and the expression path already evaluated the full condition once the validator stopped rejecting it. The "missed key schema element" message reuses the exact wording Floci already emits, so SDK error handling is unchanged.

**(follow-up, separate PR)** The bot's point about typed values on the legacy path is right in substance but predates this PR: `matchesAttributeValue` already collapsed the partition key to a string on main, so `{"S":"1"}` and `{"N":"1"}` collided before too. Worth a separate fix comparing `AttributeValue`s structurally.

**(follow-up, separate PR)** `if (!partitionKeyMatches) continue;` matches the line it replaced, but AGENTS.md asks for braces on every conditional; add them whenever convenient.

No blockers from my side.

hectorvent (approved):
Approved, see comment above.

PR #2475, opened 2026-08-22
fix(dynamodb): cap GSI multi-attribute key parts at 4 attributes

### LINKED ISSUE #2460: [BUG] DynamoDB: CreateTable does not enforce the 4-attribute limit on GSI multi-attribute keys

omatheusmesmo:
## Summary

- `CreateTable` and the `UpdateTable` GSI-create path now reject a GSI KeySchema with more than 4 HASH or more than 4 RANGE attributes

Closes #2460

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

A GSI's multi-attribute key schema caps each key part (partition key, sort key) at 4 attributes. Floci accepted an unbounded number of HASH/RANGE elements with no validation, and DescribeTable round-tripped the invalid state.

## Checklist

- [x] New tests added (createTable and updateTable paths)
- [x] Conventional Commits

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -88,6 +88,8 @@ public class DynamoDbService {
     private final ConcurrentHashMap<String, IdempotencyEntry> txIdempotency = new ConcurrentHashMap<>();
     private static final long TX_IDEMPOTENCY_TTL_NANOS = java.time.Duration.ofMinutes(10).toNanos();
 
+    private static final int MAX_MULTI_ATTRIBUTE_KEY_PART_SIZE = 4;
+
     private record IdempotencyEntry(String requestHash, long insertedAtNanos) {}
     private final RegionResolver regionResolver;
     private final ObjectMapper objectMapper;
@@ -314,6 +316,7 @@ public class DynamoDbService {
 
         if (gsis != null && !gsis.isEmpty()) {
             for (GlobalSecondaryIndex gsi : gsis) {
+                validateGsiKeySchemaArity(gsi);
                 gsi.setIndexArn(table.getTableArn() + "/index/" + gsi.getIndexName());
             }
             table.setGlobalSecondaryIndexes(new ArrayList<>(gsis));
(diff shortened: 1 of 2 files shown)


hectorvent:
Thanks for covering both create paths; the shared `validateGsiKeySchemaArity` helper and the three rejection tests are the right shape, and the developer guide is the only source for this cap (the model's `KeySchema` has no upper bound).

**(blocking)** The bot is right about `updateTable`: the new check sits in the loop that adds GSIs, which runs after the throughput write, the GSI deletes and earlier GSIs in the same request have already been applied to the live stored table, so a rejected request leaves it partially updated. The existing UpdateTable validations all run before any mutation a few lines up; please call `validateGsiKeySchemaArity(newGsi)` from that earlier loop, ideally with a test pairing an invalid GSI with a throughput change and asserting DescribeTable is unchanged.

**(blocking)** Since #2476 merged, this conflicts with main in `DynamoDbServiceTest.java` (both insert tests after `createTableRejectsArnInput`); `DynamoDbService.java` merges cleanly. Please rebase, keeping both test blocks.

omatheusmesmo:
Hi @hectorvent , it's ready to merge/review, thank you!

pgermosen (approved):
Verified against real, current AWS documentation: the 4-attributes-max-per-key-part limit matches DynamoDB's multi-attribute composite key feature exactly.

Independently verified Greptile's one P1 finding (partial mutation on a rejected UpdateTable) is a false positive: wrote a scratch test combining an invalid over-arity GSI create with a valid throughput change and new attribute definitions in one updateTable call, and confirmed the whole update is atomic — nothing was partially applied after the rejection. The method structure confirms why: all validation happens in a dedicated pass before any mutation begins.

Full dynamodb suite passes (389/389), both on the branch and on the actual merged-with-main tree (compiled + tested the merge). LGTM.

PR #2476, opened 2026-08-22
fix(dynamodb): reject LSI KeySchema with other than one sort key

### LINKED ISSUE #2459: [BUG] DynamoDB: CreateTable accepts a Local Secondary Index with more than one sort-key attribute

omatheusmesmo:
## Summary

- `CreateTable` now rejects a LocalSecondaryIndex KeySchema with other than exactly one RANGE (sort key) attribute

Closes #2459

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

An LSI's sort key must be exactly one scalar attribute. Floci accepted zero or multiple RANGE elements with no validation, and DescribeTable round-tripped the invalid state.

## Checklist

- [x] New tests added
- [x] Conventional Commits

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -327,6 +327,12 @@ public class DynamoDbService {
                     throw new AwsException("ValidationException",
                             "LocalSecondaryIndex partition key must match table partition key", 400);
                 }
+                if (lsi.getSortKeyNames().size() != 1) {
+                    throw new AwsException("ValidationException",
+                            "One or more parameter values were invalid: Local Secondary Index "
+                            + lsi.getIndexName() + " must have a sort key consisting of exactly "
+                            + "one scalar attribute", 400);
+                }
                 lsi.setIndexArn(table.getTableArn() + "/index/" + lsi.getIndexName());
             }
             table.setLocalSecondaryIndexes(new ArrayList<>(lsis));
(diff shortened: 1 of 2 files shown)


hectorvent:
Thanks, this is a clean one: the check lives in the existing LSI loop before the table is stored, it reuses `getSortKeyNames()` instead of re-filtering the key schema, and both the zero-sort-key and two-sort-key shapes are covered. The model's `KeySchema` has no structural constraint, so the developer guide is the right source for the "exactly one scalar sort key" rule, and `ValidationException` is the error DynamoDB uses for every schema rejection.

No blockers from my side.

hectorvent (approved):
Approved, see comment above.

PR #2486, opened 2026-08-23
fix: inject AWS SDK environment into Flink sidecars

### LINKED ISSUE #2404: [BUG] kinesisanalytics (Managed Service for Apache Flink) sidecar containers get no AWS credentials or IMDS routing
## Service

Kinesis Analytics V2 (Managed Service for Apache Flink)

## AWS API Action

`StartApplication` (the container it launches is where the gap shows up; `CreateApplication` is unaffected)

## Expected behavior

On real AWS, a Managed Service for Apache Flink application runs under the IAM role passed via
`ApplicationConfiguration` / `ServiceExecutionRole`, and the AWS SDK inside the job resolves credentials against that role automatically. Every other container-backed service Floci emulates does the equivalent locally: 

Lambda functions get `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN` injected into their execution environment, and ECS tasks get a task role reachable via the ECS container-credentials endpoint. I'd expect the Flink sidecar containers started by `StartApplication` to get the same baseline AWS SDK environment.

## Actual behavior

The Flink sidecar container that `StartApplication` launches gets no AWS credentials and no IMDS / container
credentials endpoint at all — `env` inside the container has no `AWS_*` variables, and there's no route to a credentials endpoint. Any Flink job that relies on the AWS SDK's default credential chain (e.g. to read/write S3, or to call back into Floci's own emulated services from a source/sink) fails to resolve credentials immediately after start.

For comparison, exec-ing into a Lambda function container started against the same Floci instance shows
`AWS_ACCESS_KEY_ID` etc. already set — so this looks like the kinesisanalytics container path (added in #1914) never received the same baseline-environment injection that Lambda/ECS/Batch get.

## Reproduction

Verified against `floci/floci:1.7.0` on 2026-08-19 (Docker Compose 5.4.0, rootful Podman machine on macOS — the `security_opt: label=disable` below is only needed on that setup; drop it on Docker Desktop / rootless Podman).

```yaml
# docker-compose.yml
services:
  floci:
    image: floci/floci:1.7.0
    environment:
      FLOCI_SERVICES_DOCKER_NETWORK: kinesisanalytics-net
    volumes:
    security_opt:
      - label=disable
    ports:
      - "4566:4566"
    networks:
      - kinesisanalytics-net

networks:
  kinesisanalytics-net:
    name: kinesisanalytics-net
```

```bash
docker compose up -d

export AWS_ACCESS_KEY_ID=admin AWS_SECRET_ACCESS_KEY=password AWS_DEFAULT_REGION=us-east-1

# Any valid, non-empty jar works — the credential gap shows up before the job logic runs.
zip -q dummy.jar META-INF/MANIFEST.MF   # or any other non-empty jar you have lying around
aws --endpoint-url http://localhost:4566 s3 mb s3://dummy
aws --endpoint-url http://localhost:4566 s3 cp dummy.jar s3://dummy/dummy.jar

aws --endpoint-url http://localhost:4566 kinesisanalyticsv2 create-application \
  --application-name repro \
  --runtime-environment FLINK-1_20 \
  --service-execution-role arn:aws:iam::000000000000:role/dummy \
  --application-configuration '{
    "ApplicationCodeConfiguration": {
      "CodeContent": {"S3ContentLocation": {"BucketARN": "arn:aws:s3:::dummy", "FileKey": "dummy.jar"}},
      "CodeContentType": "ZIPFILE"
    }
  }'

aws --endpoint-url http://localhost:4566 kinesisanalyticsv2 start-application \
  --application-name repro --run-configuration '{}'

# Find the sidecar container Floci spun up for the application:
docker ps --filter "name=repro"
# -> floci-kinesisanalytics-repro   apache/flink:1.20   (+ a -tm taskmanager container)

# No AWS credentials anywhere in the container:
docker exec floci-kinesisanalytics-repro env | grep -i aws
# -> (empty)

# No container-credentials / IMDS route either:
docker exec floci-kinesisanalytics-repro sh -c \
  'curl -s -m 2 -o /dev/null -w "http_code=%{http_code} exit=%{exitcode}\n" http://169.254.170.2/'
# -> http_code=000 exit=7   (connection failed, no route)

# Compare with a Lambda function against the same Floci instance:
  --function-name repro-lambda --runtime python3.12 --handler index.handler \
  --role arn:aws:iam::000000000000:role/dummy --zip-file fileb://lambda.zip
aws --endpoint-url http://localhost:4566 lambda invoke --function-name repro-lambda out.json

docker exec <lambda-container-name> env | grep -i aws
# -> AWS_ACCESS_KEY_ID=test, AWS_SECRET_ACCESS_KEY=test, AWS_SESSION_TOKEN=test,
#    AWS_ENDPOINT_URL=http://<container-ip>:4566, AWS_REGION=us-east-1, ... all already set
```

I ran the exact sequence above end-to-end: `create-application`/`start-application` succeed and the sidecar containers come up fine — the AWS SDK environment is just never populated inside them, while the equivalent Lambda container gets a full set of `AWS_*` variables.

## Workaround

Fortunately, there is a workaround:  
Until this is fixed upstream, I baked placeholder credentials into the image `FLOCI_SERVICES_KINESIS_ANALYTICS_DEFAULT_IMAGE`
points at, since Floci only ever sets `FLINK_PROPERTIES` on the sidecar container — image-level `ENV` defaults
survive untouched:

```dockerfile
FROM flink:2.2.1

ENV AWS_ACCESS_KEY_ID=admin
ENV AWS_SECRET_ACCESS_KEY=password
ENV AWS_REGION=us-east-1
```

```yaml
# docker-compose.yml, on the floci service
environment:
  FLOCI_SERVICES_KINESIS_ANALYTICS_DEFAULT_IMAGE: my-org/flink-floci:2.2.1
```

This is enough to satisfy jobs that only talk to Floci's own emulated services (which accept any static
credentials), but it's a workaround every consumer of this feature has to independently discover and rebuild, and it doesn't help jobs that need to reach a real downstream AWS account through the same SDK client.

## Environment

- Floci version / image tag: 1.7.0
- Java SDK version (if applicable): AWS SDK for Java v2, `kinesisanalyticsv2` client (software.amazon.awssdk:kinesisanalyticsv2)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker Compose

mvanhorn:
## Summary

Inject `LaunchedContainerAwsEnv` into `FlinkContainerManager` and resolve one baseline from the configured default region with no mounted AWS config, preserving the established host-credential-or-placeholder behavior. Apply that baseline to both the JobManager and TaskManager specifications while retaining each process's existing `FLINK_PROPERTIES`; user code may execute in either container, so both require identical AWS SDK variables. Configure the JobManager's network setup consistently with other launched workloads so the helper's container-reachable endpoint resolves from the shared network namespace used by the TaskManager.

`StartApplication` creates Flink JobManager and TaskManager containers with only `FLINK_PROPERTIES`, leaving AWS SDKs inside the application without a region, credentials, or an endpoint for Floci. The failure is isolated to the Kinesis Analytics container-launch path; `LaunchedContainerAwsEnv` already defines the shared baseline used by Lambda, ECS, and MWAA. As a result, otherwise healthy Flink jobs fail as soon as their default credential chain attempts to access S3, Kinesis, or another AWS-compatible service. The supplied reproduction is concrete, and the issue has no assignee, claim, competing PR, or prior closed attempt.

Closes #2404

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

This change is specifically about AWS compatibility inside launched Flink
containers. It reuses the existing `LaunchedContainerAwsEnv` baseline already
used by the Lambda, ECS, and MWAA launch paths rather than introducing a new
one, so region, credential, and endpoint resolution stay consistent with those
workloads. No AWS SDK version or API surface changes.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kinesisanalytics/container/FlinkContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesisanalytics/container/FlinkContainerManager.java
@@ -14,6 +14,7 @@ import io.github.hectorvent.floci.core.common.docker.ContainerLifecycleManager.E
 import io.github.hectorvent.floci.core.common.docker.ContainerLogStreamer;
 import io.github.hectorvent.floci.core.common.docker.ContainerSpec;
 import io.github.hectorvent.floci.core.common.docker.ContainerStorageHelper;
+import io.github.hectorvent.floci.core.common.docker.LaunchedContainerAwsEnv;
 import io.github.hectorvent.floci.services.kinesisanalytics.KinesisAnalyticsRuntimes;
 import io.github.hectorvent.floci.services.kinesisanalytics.model.FlinkApplication;
 import io.github.hectorvent.floci.services.kinesisanalytics.model.Snapshot;
@@ -35,7 +36,9 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
@@ -68,6 +71,7 @@ public class FlinkContainerManager {
     private final ContainerDetector containerDetector;
     private final EmulatorConfig config;
     private final RegionResolver regionResolver;
+    private final LaunchedContainerAwsEnv awsEnv;
     private final S3Service s3Service;
     private final FlinkRestClient flinkRest;
     private final ObjectMapper objectMapper;
@@ -87,6 +91,7 @@ public class FlinkContainerManager {
                                  ContainerDetector containerDetector,
                                  EmulatorConfig config,
                                  RegionResolver regionResolver,
+                                 LaunchedContainerAwsEnv awsEnv,
                                  S3Service s3Service,
                                  FlinkRestClient flinkRest,
                                  ObjectMapper objectMapper) {
@@ -96,6 +101,7 @@ public class FlinkContainerManager {
         this.containerDetector = containerDetector;
         this.config = config;
         this.regionResolver = regionResolver;
+        this.awsEnv = awsEnv;
         this.objectMapper = objectMapper;
         this.s3Service = s3Service;
         this.flinkRest = flinkRest;
@@ -129,6 +135,7 @@ public class FlinkContainerManager {
         // Read the JAR before starting anything so a missing/empty artifact fails fast, before any
         // container is created. (S3 read runs on the request thread → account context is available.)
         byte[] jarBytes = app.hasCode() ? readJar(app) : null;
+        List<String> awsBaselineEnv = awsEnv.sdkBaselineEnv(config.defaultRegion(), Optional.empty());
 
         LOG.infov("Starting Flink cluster for application {0} using image {1}{2}",
                 app.getApplicationName(), image, app.hasCode() ? " (with TaskManager)" : "");
@@ -141,8 +148,11 @@ public class FlinkContainerManager {
         ContainerBuilder.Builder jmSpec = containerBuilder.newContainer(image)
                 .withName(jmName)
                 .withCmd("jobmanager")
+                .withEnv(awsBaselineEnv)
                 .withEnv("FLINK_PROPERTIES", jmProps)
                 .withDockerNetwork(config.services().dockerNetwork())
+                .withHostDockerInternalOnLinux()
+                .withEmbeddedDns()
                 .withLogRotation();
         // Named volume (not the container's ephemeral filesystem) so snapshots survive a
         // Stop/StartApplication cycle — stopCluster() removes the JobManager container, but this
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Verified the core claim: LaunchedContainerAwsEnv is genuinely already used by Lambda, ECS, and MWAA — this extends the exact same established pattern to Flink's JobManager/TaskManager, including the withHostDockerInternalOnLinux()/withEmbeddedDns() network wiring that ECS already applies alongside it.

Verified load-bearing: reverting the JobManager's env injection + network setup makes bareClusterInjectsAwsSdkEnvironmentAndContainerReachability fail exactly as expected. Restored, confirmed clean.

The rewritten test suite uses a real (non-mocked) ContainerBuilder to assert on the actual produced ContainerSpec rather than mocking everything away — solid design. Full kinesisanalytics suite passes (91/91), both on the branch and on the actual merged-with-main tree (compiled + tested the merge). LGTM.

PR #2487, opened 2026-08-23
fix(bedrock-agentcore): allow maxResults up to 1000

### LINKED ISSUE #2418: [BUG] bedrock-agentcore-control: ListGateways/ListGatewayTargets reject maxResults above 100, model declares 1000
## Service

Bedrock AgentCore Control (`bedrock-agentcore-control`)

## AWS API Action

`ListGateways` and `ListGatewayTargets`

## Expected behavior

Both operations declare `maxResults` with a maximum of **1000**, so a request with any value in 1–1000 is accepted:

| Operation | Request shape member | Declared shape | min | max |
|---|---|---|---|---|
| `ListGateways` | `maxResults` | `GatewayMaxResults` | 1 | **1000** |
| `ListGatewayTargets` | `maxResults` | `TargetMaxResults` | 1 | **1000** |

Verified against [botocore's `bedrock-agentcore-control` service model](https://github.com/boto/botocore/blob/develop/botocore/data/bedrock-agentcore-control/2023-06-05/service-2.json) rather than the public API reference page.

## Actual behavior

`BedrockAgentCoreGatewayService` caps both at 100:

https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/bedrockagentcorecontrol/BedrockAgentCoreGatewayService.java#L31

```java
private static final int MAX_PAGE = 100;
```

That constant is passed as `maxPage` to `Pagination.paginate` for both `list` (`ListGateways`) and `listTargets` (`ListGatewayTargets`), and `Pagination` rejects anything above it:

```java
if (maxResults != null && (maxResults < 1 || maxResults > maxPage)) {
    throw new AwsException(errorCode, "maxResults must be between 1 and " + maxPage, 400);
}
```

So `maxResults` anywhere in **101–1000** gets a `400 ValidationException` — `maxResults must be between 1 and 100` — for a value real AWS accepts. A client paging with a legal page size fails outright rather than getting a smaller page.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 bedrock-agentcore-control list-gateways --max-results 500
# Floci:    ValidationException: maxResults must be between 1 and 100
# Real AWS: returns up to 500 gateways
```

Same for `list-gateway-targets --gateway-identifier <id> --max-results 500`.

## Scope

Only this one service is affected — I checked the other AgentCore list operations and MSK's while confirming, and the rest already agree with their models:

| Constant | Operations it covers | Declared max | Matches? |
|---|---|---|---|
| `BedrockAgentCoreGatewayService.MAX_PAGE = 100` | `ListGateways`, `ListGatewayTargets` | 1000 | ❌ |
| `BedrockAgentCoreMemoryService.MAX_PAGE = 100` | `ListMemories` | 100 | ✅ |
| `BedrockAgentCoreIdentityService.MAX_PAGE = 20` | `ListWorkloadIdentities` | 20 | ✅ |
| `MskService.MAX_PAGE = 100` | `ListConfigurations`, `ListConfigurationRevisions` | 100 | ✅ |

## One decision for whoever picks this up

`MAX_PAGE` is doing two jobs in `Pagination.paginate`: it is the validation ceiling *and* the default page size when the caller omits `maxResults` (`int limit = maxResults != null ? maxResults : maxPage;`).

So raising it to 1000 fixes the false rejection but also changes an unpaginated `ListGateways` from returning up to 100 items to up to 1000. AWS's real default page size isn't in the service model, and it is generally *not* the same as the declared maximum — so the two probably want to be separate constants (a `MAX_PAGE` of 1000 and a smaller default) rather than one value raised to 1000. Worth deciding deliberately, since the same coupling exists in every service using `Pagination`.

## Environment

- Floci version: `main`
- Found while reviewing #2344 (thanks @hectorvent for spotting the constant).

AbarnaaSree:
## Summary

Closes #2418

Fixes `ListGateways` and `ListGatewayTargets` rejecting valid `maxResults` values between 101 and 1000.

The Bedrock AgentCore service model allows `maxResults` up to 1000, but Floci was limiting it to 100. This change separates the default page size from the maximum allowed `maxResults` value.

The default page size remains 100 to preserve existing pagination behavior.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The previous implementation rejected valid requests with:

```text
maxResults must be between 1 and 100

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/Pagination.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/Pagination.java
@@ -25,12 +25,20 @@ public final class Pagination {
     private Pagination() {}
 
     public static <T> PaginatedResult<T> paginate(List<T> all, Function<T, String> cursorOf,
+                                                Integer maxResults, String nextToken,
+                                                int maxPage, String errorCode) {
+        return paginate(all, cursorOf, maxResults, nextToken, maxPage, maxPage, errorCode);
+    }
+
+    public static <T> PaginatedResult<T> paginate(List<T> all, Function<T, String> cursorOf,
                                                     Integer maxResults, String nextToken,
-                                                    int maxPage, String errorCode) {
-        if (maxResults != null && (maxResults < 1 || maxResults > maxPage)) {
-            throw new AwsException(errorCode, "maxResults must be between 1 and " + maxPage, 400);
+                                                    int defaultPageSize, int maxResultsLimit,
+                                                    String errorCode) {
+        if (maxResults != null && (maxResults < 1 || maxResults > maxResultsLimit)) {
+            throw new AwsException(errorCode,
+                    "maxResults must be between 1 and " + maxResultsLimit, 400);
         }
-        int limit = maxResults != null ? maxResults : maxPage;
+        int limit = maxResults != null ? maxResults : defaultPageSize;
 
         List<T> sorted = all.stream().sorted(Comparator.comparing(cursorOf)).collect(Collectors.toList());
         String after = decodeToken(nextToken, errorCode);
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Verified against the real botocore Bedrock AgentCore Control model: GatewayMaxResults/TargetMaxResults both have min:1, max:1000, exact match to this fix.

The shared Pagination.paginate() change is backward-compatible — the original 4-parameter signature is preserved as an overload, so the other 4 callers (EmrServerlessService, MskService, and the other bedrockagentcorecontrol services) are untouched. Ran their suites directly to confirm: 131 tests, 0 failures.

Verified load-bearing: reverting MAX_PAGE back to 100 reproduces the exact original bug (maxResults=101 rejected). Full suite passes clean both on the branch and on the actual merged-with-main tree (compiled + tested the merge). LGTM.

PR #2491, opened 2026-08-23
fix(msk): accept an empty serverProperties on Create/UpdateConfiguration

allensanborn:
## Summary

`MskService` validated `serverProperties` with `== null || isBlank()` in both `createConfiguration` and `updateConfiguration`, which conflates *member absent* with *member present but empty*. A zero-length blob is a legitimate value in MSK: it means "no property overrides". Both call sites now reject only `null`.

Absent and empty stay distinguishable at the REST layer, so the validation can stay exactly where it is: `MskController.asString` returns `null` for a missing JSON member, while `decodeServerProperties` turns the empty base64 string into `""`. An absent member is still a 400.

This codebase already treats `""` as the canonical "no property overrides" value. `MskConfiguration.setLegacyLatestRevision` deliberately migrates a pre-revision-history entry with no `serverProperties` to `""` rather than `null`, because `serverPropertiesByRevision` is a `ConcurrentHashMap` and `DescribeConfigurationRevision` base64-encodes whatever it finds without a null check. The one value that could not round-trip through storage was already excluded; `""` was never the problem.

### Tests

Four new unit tests (create/update x empty/non-empty accepted; the existing null-rejection tests already cover `null`), plus REST-level tests proving that empty round-trips as `""` through create *and* update. Both revisions are read back, so an update that silently carried the previous revision's properties forward would fail. An absent member still returns 400 on create; the absent-on-update case was already covered by `MskControllerIntegrationTest.updateConfigurationRejectsMissingServerProperties`.

Confirmed adversarially by reverting the production change: `createConfigurationAcceptsEmptyServerProperties`, `updateConfigurationAcceptsEmptyServerProperties`, `updateConfigurationAcceptsNonEmptyServerProperties` and `createAndUpdateConfigurationAcceptEmptyServerProperties` all go red. Removing the `null` guard instead turns the two missing-member tests red.

`./mvnw test -Dtest='*Msk*'` -> 62 tests, 0 failures. Full suite: `./mvnw test` -> 11568 tests, 0 failures.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behaviour** — a present-but-empty `serverProperties` was rejected:

```
$ aws kafka create-configuration --name repro --kafka-versions 3.6.0 --server-properties ""
BadRequestException: serverProperties is required.
```

The same guard was present in `UpdateConfiguration`.

**Why an empty value is legal.** AWS's own service model settles it — `kafka/service-2.json`:

```
CreateConfiguration input required: ["ServerProperties", "Name"]
ServerProperties: { "shape": "__blob", "locationName": "serverProperties" }
__blob:           { "type": "blob" }
```

`required` with **no `min` constraint**. In the AWS models `required` means the member must be *present*; a length floor is expressed explicitly as `min`. So a zero-length blob is a valid `ServerProperties`.

**Real-world impact.** Terraform modules that expose server properties as an optional map render an empty map to an empty string — e.g. Gruntwork's `msk` module builds `server_properties = join("\n", [for k, v in var.server_properties : "${k}=${v}"])` with `default = {}` and creates `aws_msk_configuration` unconditionally. Three examples in that repo fail on this today.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/msk/MskService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/msk/MskService.java
@@ -134,7 +134,10 @@ public class MskService {
             throw new AwsException("BadRequestException",
                     "Configuration name must match the pattern \"^[0-9A-Za-z][0-9A-Za-z-]{0,}$\".", 400);
         }
-        if (serverProperties == null || serverProperties.isBlank()) {
+        // Only an absent member (null) is rejected. A zero-length blob is a legitimate value
+        // meaning "no property overrides": Terraform modules that build serverProperties by
+        // joining a map defaulting to {} send exactly that, and real MSK accepts it.
+        if (serverProperties == null) {
             throw new AwsException("BadRequestException", "serverProperties is required.", 400);
         }
         if (configurationStorage.scan(k -> true).stream().anyMatch(c -> c.getName().equals(name))) {
@@ -178,7 +181,9 @@ public class MskService {
     }
 
     public MskConfiguration updateConfiguration(String arn, String description, String serverProperties) {
-        if (serverProperties == null || serverProperties.isBlank()) {
+        // Same absent-vs-empty distinction as createConfiguration: "" is a valid revision
+        // body that clears every override, only a missing member is an error.
+        if (serverProperties == null) {
             throw new AwsException("BadRequestException", "serverProperties is required.", 400);
         }
 
--- a/src/test/java/io/github/hectorvent/floci/services/msk/MskControllerIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/msk/MskControllerIntegrationTest.java
@@ -162,16 +162,20 @@ class MskControllerIntegrationTest {
             .statusCode(200)
             .body("latestRevision.revision", equalTo(2));
 
-        given()
-        .when()
-            .get("/v1/configurations/{arn}/revisions/{revision}", arn, 1)
-        .then()
-            .statusCode(200)
-            .body("serverProperties", equalTo(""));
+        // Both revisions, not just the one create wrote: an empty update has to store ""
+        // rather than silently carry the previous revision's properties forward.
+        for (int revision : new int[] { 1, 2 }) {
+            given()
+            .when()
+                .get("/v1/configurations/{arn}/revisions/{revision}", arn, revision)
+            .then()
+                .statusCode(200)
+                .body("serverProperties", equalTo(""));
+        }
     }
 
     @Test
-    void createConfigurationRejectsAbsentServerProperties() {
+    void createConfigurationRejectsMissingServerProperties() {
         given()
             .contentType("application/json")
             .body("""
--- a/src/test/java/io/github/hectorvent/floci/services/msk/MskControllerIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/msk/MskControllerIntegrationTest.java
@@ -133,6 +133,56 @@ class MskControllerIntegrationTest {
             .statusCode(404);
     }
 
+    // An empty base64 blob decodes to "" and means "no property overrides". Absent and
+    // present-but-empty stay distinguishable at the REST layer: a missing member arrives as
+    // null, an empty one as a zero-length String, so only the former is rejected.
+    @Test
+    void createAndUpdateConfigurationAcceptEmptyServerProperties() {
+        String arn = given()
+            .contentType("application/json")
+            .body("""
+                {"name": "empty-props-%s", "kafkaVersions": ["3.6.0"], "serverProperties": ""}
+                """.formatted(UUID.randomUUID().toString().substring(0, 8)))
+        .when()
+            .post("/v1/configurations")
+        .then()
+            .statusCode(200)
+            .body("state", equalTo("ACTIVE"))
+            .body("latestRevision.revision", equalTo(1))
+            .extract().path("arn");
+
+        given()
+            .contentType("application/json")
+            .body("""
+                {"description": "still empty", "serverProperties": ""}
+                """)
+        .when()
+            .put("/v1/configurations/{arn}", arn)
+        .then()
+            .statusCode(200)
+            .body("latestRevision.revision", equalTo(2));
+
+        given()
+        .when()
+            .get("/v1/configurations/{arn}/revisions/{revision}", arn, 1)
+        .then()
+            .statusCode(200)
+            .body("serverProperties", equalTo(""));
+    }
+
+    @Test
+    void createConfigurationRejectsAbsentServerProperties() {
+        given()
+            .contentType("application/json")
+            .body("""
+                {"name": "no-props-config", "kafkaVersions": ["3.6.0"]}
+                """)
+        .when()
+            .post("/v1/configurations")
+        .then()
+            .statusCode(400);
+    }
+
     @Test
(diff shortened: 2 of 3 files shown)


pgermosen:
Reviewed the code itself in full — it's solid, and I independently verified the core claims rather than taking the writeup on trust:

- Pulled the real `kafka/2018-11-14/service-2.json` from botocore: `CreateConfigurationRequest`/`UpdateConfigurationRequest` both list `ServerProperties` under `required` (member must be *present*), and the `__blob` shape it points to is `{"type": "blob"}` with no `min` constraint anywhere. So "required" here genuinely means presence, not non-emptiness — the PR's reading of the wire model is correct.
- Confirmed the claimed pre-existing convention is real, not asserted: `MskConfiguration.setLegacyLatestRevision` (untouched by this PR) already migrates a legacy entry with no `serverProperties` to `""` rather than `null`, specifically because `serverPropertiesByRevision` is a `ConcurrentHashMap` (rejects null values) and `DescribeConfigurationRevision` base64-encodes without a null check. So `""` was never actually a problem for the storage/round-trip layer.
- Did my own independent revert-and-confirm-fail rather than just trusting the PR description's own adversarial-testing claim: reverting to `== null || isBlank()` reproduces exactly the 4 failures described. Going the *other* direction — removing the `null` guard entirely — doesn't just fail the two missing-member tests, it produces an unhandled `500 NullPointerException` on both `POST /v1/configurations` and the update path, which is even stronger evidence the guard is load-bearing than the PR claims. Restored both times, full `msk` suite (62 tests) green, including on a fresh merge against current `main`.

Code-wise, no notes — nice work, especially the round-trip test reading back both revisions so a silently-carried-forward previous revision's properties couldn't slip through.

Two hygiene items outside the code itself:

1. **PR description skips the template** — no `## Type of change` checkboxes, no `## Checklist`, no `Closes #N` link. The custom Problem/Why/Change/Tests structure you used is genuinely more informative than the template would have been here, so I'm not asking you to rewrite it — just flagging that it's missing the standard sections in case that matters for how this gets tracked.
2. **Both commits carry `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>` trailers** (one also has a `Claude-Session:` link). CONTRIBUTING.md and AGENTS.md both say: "Do not include `Co-Authored-By` trailers for AI tools in commit messages." Could you amend or rebase to drop those and force-push? This is the actual blocker before merge — everything else here is ready.

hectorvent:
@allensanborn please rebase main, CI timeout was fixed.

allensanborn:
Thanks @hectorvent — rebased onto current `main` (`257e0046`). Clean rebase, no conflicts; the patch is byte-identical to what you reviewed.

New head: `0b893320`

Locally on the rebased branch: `*Msk*` is 62 tests, 0 failures, 0 errors, 0 skipped. CI is re-running now on the new 4-way shard matrix.

allensanborn:
@pgermosen — following up on the trailer blocker, which I should have confirmed explicitly rather than only mentioning the rebase. My earlier note said the patch was byte-identical to what you reviewed, which is true of the *diff* but understates what changed about the commits: **the `Co-Authored-By` and `Claude-Session:` trailers are gone.**

Current head is `0b893320`. Verified from the API rather than locally, so it reflects what you'd see:

```
gh api repos/floci-io/floci/pulls/2491/commits --jq '.[].commit.message' \
  | grep -ciE 'co-authored-by|claude|generated with'
0
```

I ran the same check across all five of my open PRs (#2491, #2494, #2526, #2527, #2528) — zero hits on every one, so the amend you asked for is done everywhere, not just here.

All checks are green on this head, and `*Msk*` is 62 tests, 0 failures locally.

Could you drop the `waiting-contributor` label when you get a chance? I think it's just outlived the thing it was tracking — it went on within seconds of your review, so it predates the fix rather than responding to it.

On the PR-description template: leaving that as-is per your note that you weren't asking for a rewrite. Happy to add the standard sections if it turns out to matter for tracking.

pgermosen (approved):
Re-checked at `0b893320`, merged onto current `main`, rebuilt, and driven with the real AWS CLI.

## The blocker is closed

I checked the commit objects rather than re-running your grep, since that is the thing I actually asked about:

```
5e8c9e96 fix(msk): accept an empty serverProperties on Create/UpdateConfiguration
0b893320 test(msk): read back both revisions, and match the class's naming
    author/committer: Allen Sanborn <allen.sanborn@gmail.com>
    no Co-Authored-By, no Claude-Session
```

Both clean. And the wider claim holds too — #2491, #2494, #2526, #2527 and #2528 all return zero. Thank you for going back over the whole set rather than just this one; that is the part that actually fixes the convention.

The description now carries `## Type of change` and `## Checklist` as well, so both hygiene items from last round are done.

## Re-verified on the rebased head

You rebased onto `257e0046`; `main` has since moved to `5c45c146`, and the merge is still clean. **62 MSK tests, 0 failures** on that merge — the same number you reported.

The reproduction from the description now works end to end through the CLI:

```
$ aws kafka create-configuration --name repro --kafka-versions 3.6.0 --server-properties ""
{ "State": "ACTIVE", "LatestRevision": { "Revision": 1 } }

$ aws kafka describe-configuration-revision --arn … --revision 1
"ServerProperties": ""

$ aws kafka update-configuration --arn … --server-properties "" --description "still empty"
Revision 2  →  "ServerProperties": ""
```

The non-empty path is untouched — base64 in, identical base64 out, decoding back to `auto.create.topics.enable=true` — and an absent member is still `400 BadRequestException: serverProperties is required.`

**Both mutation directions reproduce exactly what you claimed**, run on this head rather than taken from the description:

```
revert to `== null || isBlank()`      → 4 failures:
    MskServiceTest.createConfigurationAcceptsEmptyServerProperties
    MskServiceTest.updateConfigurationAcceptsEmptyServerProperties
    MskServiceTest.updateConfigurationAcceptsNonEmptyServerProperties
    MskControllerIntegrationTest.createAndUpdateConfigurationAcceptEmptyServerProperties

remove the `null` guard instead       → 4 failures, and the two service-level ones fail with
    `NullPointerException` rather than `AwsException` — so without it a missing member is an
    unhandled 500, not a 400
```

The model reading is right, and it is the whole argument: `CreateConfigurationRequest` and `UpdateConfigurationRequest` both list `ServerProperties` under `required`, and it points at `__blob`, which is `{"type": "blob"}` — no `min` anywhere. Presence, not length.

## Two observations, neither of them asks

- `updateConfigurationAcceptsNonEmptyServerProperties` creates with `""` and *then* updates with `num.partitions=3`, so what it really pins is the empty → non-empty transition — the case where a Terraform module starts with `{}` and later gains properties. That is a more useful test than its name suggests; I only mention it because I had to read it to understand why it goes red under the first mutation.
- Dropping `isBlank()` also means whitespace-only is now accepted. That turns out to be a non-event: base64 decoding ignores whitespace, so `"   "` lands as `""` and is indistinguishable from the empty case. Confirmed on a running instance.

Unrelated to this change and out of scope, but worth recording somewhere: `MskService:84` and `:148` build ARNs from `config.defaultRegion()` rather than the request's region, so a configuration created while signing `us-west-2` comes back as `arn:aws:kafka:us-east-1:…`. Pre-existing and untouched here. Happy to file it separately.

## Approving

Two lines of production change, the service model as the basis rather than an inference, tests that fail in both directions when the change is undone, and a real Terraform failure mode behind it. The label is already off `waiting-contributor` — you are right that it went on reflexively rather than in response to anything.

PR #2492, opened 2026-08-23
fix(ec2): store and report a route's IPv6 destination, and stop DeleteRoute NPEing

allensanborn:
## Summary

Two linked defects in EC2 route handling.

**1. An IPv6 route is stored with no destination.** `CreateRoute` never reads `DestinationIpv6CidrBlock`, so a route created with one is persisted with a null IPv4 destination and no IPv6 destination.

**2. `DeleteRoute` then throws an NPE — and takes unrelated routes with it.** `deleteRoute`'s `removeIf` dereferences `getDestinationCidrBlock()` for *every* route in the table, so once a table holds one IPv6 route, deleting an ordinary IPv4 route fails:

```
$ aws ec2 create-route --route-table-id $RT --destination-cidr-block 0.0.0.0/0 --gateway-id $IGW
$ aws ec2 create-route --route-table-id $RT --destination-ipv6-cidr-block ::/0 --gateway-id $IGW
$ aws ec2 delete-route --route-table-id $RT --destination-cidr-block 0.0.0.0/0
InternalFailure: ... Route.getDestinationCidrBlock() is null
```

An `InternalFailure` is never a correct response here, independent of any modelling question.

Routes now carry whichever destination they were created with; `DeleteRoute` matches on the destination the request names; `DescribeRouteTables` reports `destinationIpv6CidrBlock`. A prefix list is treated as the third destination kind rather than as the absence of one, and the `removeIf` null-safety covers all three.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behaviour:** an IPv6 or prefix-list route lost its destination on create, making it unaddressable by `DeleteRoute`/`ReplaceRoute`, and its null IPv4 destination made every `DeleteRoute` against the same table throw `InternalFailure`.

Verified against the botocore EC2 model (`ec2/service-2.json`), which declares the three destinations as separate optional members:

```
CreateRoute input:  DestinationCidrBlock, DestinationIpv6CidrBlock, DestinationPrefixListId
DeleteRoute input:  DestinationCidrBlock, DestinationIpv6CidrBlock, DestinationPrefixListId
Route (output):     DestinationCidrBlock, DestinationIpv6CidrBlock, DestinationPrefixListId
```

`EgressOnlyInternetGatewayId` is declared on `CreateRoute`/`ReplaceRoute` but not `DeleteRoute`, which is what the implementation follows.

**Impact:** Terraform retries the 500, so this surfaces as a `destroy` that hangs rather than as an error. Measured on a real Terratest run: `aws_route.internet[0]: Still destroying... 19m30s elapsed`, then a runner timeout. With the fix, `Destroy complete! Resources: 46 destroyed` in ~12s.

**Behaviour change beyond the bug — please review.** `CreateRoute`/`DeleteRoute`/`ReplaceRoute` naming **no** destination member now return an error where they previously succeeded. A route stored with no destination can never be addressed again, so erroring is correct in principle, but it is a deliberate change.

**On the error code — inferred, not confirmed.** AWS declares no operation-specific errors for these three actions: the `CreateRoute` reference has an empty Errors section and `ec2/service-2.json` carries no error shapes. `MissingParameter` is chosen from EC2's common client error codes rather than confirmed against the real service; `InvalidParameterCombination` would also fit. Happy to change it if you know which the real service returns.

`EgressOnlyInternetGatewayId` combined with `GatewayId` or `NatGatewayId` is also rejected. That is scoped deliberately to the parameter this PR newly accepts — `CreateRoute` has never enforced exclusivity between `GatewayId` and `NatGatewayId`, and starting to would be a change beyond this fix.

**Known, deliberately out of scope:** creating the same destination twice persists a duplicate. As you noted, `main` has the identical `next.add(route)` with no dedup for IPv4 today, so this PR extends existing behaviour consistently rather than worsening it. Happy to open a follow-up covering all three destination kinds at once.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -815,11 +815,12 @@ public class CloudFormationResourceProvisioner {
         String routeTableId = resolveOptional(props, "RouteTableId", engine);
         String destinationCidr = resolveOptional(props, "DestinationCidrBlock", engine);
         String destinationIpv6Cidr = resolveOptional(props, "DestinationIpv6CidrBlock", engine);
+        String destinationPrefixListId = resolveOptional(props, "DestinationPrefixListId", engine);
         String gatewayId = resolveOptional(props, "GatewayId", engine);
         String natGatewayId = resolveOptional(props, "NatGatewayId", engine);
         String egressOnlyInternetGatewayId = resolveOptional(props, "EgressOnlyInternetGatewayId", engine);
         ec2Service.createRoute(region, routeTableId, destinationCidr, destinationIpv6Cidr,
-                gatewayId, natGatewayId, egressOnlyInternetGatewayId);
+                destinationPrefixListId, gatewayId, natGatewayId, egressOnlyInternetGatewayId);
         r.setPhysicalId(r.getLogicalId() + "-" + UUID.randomUUID().toString().substring(0, 8));
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -814,9 +814,12 @@ public class CloudFormationResourceProvisioner {
     private void provisionRoute(StackResource r, JsonNode props, CloudFormationTemplateEngine engine, String region) {
         String routeTableId = resolveOptional(props, "RouteTableId", engine);
         String destinationCidr = resolveOptional(props, "DestinationCidrBlock", engine);
+        String destinationIpv6Cidr = resolveOptional(props, "DestinationIpv6CidrBlock", engine);
         String gatewayId = resolveOptional(props, "GatewayId", engine);
         String natGatewayId = resolveOptional(props, "NatGatewayId", engine);
-        ec2Service.createRoute(region, routeTableId, destinationCidr, gatewayId, natGatewayId);
+        String egressOnlyInternetGatewayId = resolveOptional(props, "EgressOnlyInternetGatewayId", engine);
+        ec2Service.createRoute(region, routeTableId, destinationCidr, destinationIpv6Cidr,
+                gatewayId, natGatewayId, egressOnlyInternetGatewayId);
         r.setPhysicalId(r.getLogicalId() + "-" + UUID.randomUUID().toString().substring(0, 8));
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
@@ -2405,9 +2405,14 @@ public class Ec2QueryHandler {
     private Response handleCreateRoute(MultivaluedMap<String, String> p, String region) {
         String rtId = p.getFirst("RouteTableId");
         String dest = p.getFirst("DestinationCidrBlock");
+        String destIpv6 = p.getFirst("DestinationIpv6CidrBlock");
         String gwId = p.getFirst("GatewayId");
         String natGwId = p.getFirst("NatGatewayId");
-        service.createRoute(region, rtId, dest, gwId, natGwId);
+        // The gateway itself is not modelled (CreateEgressOnlyInternetGateway is still
+        // unimplemented), but the id the caller sent is stored and reported back so an IPv6
+        // egress route is not silently rewritten into a targetless one.
+        String eigwId = p.getFirst("EgressOnlyInternetGatewayId");
+        service.createRoute(region, rtId, dest, destIpv6, gwId, natGwId, eigwId);
         return booleanResponse("CreateRoute");
     }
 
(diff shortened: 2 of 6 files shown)


pgermosen:
Thanks for the thorough write-up and for being explicit about what was and wasn't verified — that kind of honesty (flagging the MissingParameter-vs-InvalidParameterCombination choice as inferred, not confirmed) makes this much easier to review properly. A couple of things need addressing before this is ready, though:

## PR description doesn't follow the template

The repo's `.github/pull_request_template.md` expects `## Summary`, `## Type of change` (checkboxes), `## AWS Compatibility`, and `## Checklist` sections. This PR uses a different structure (`## Problem`, `## Change`, etc.) instead — the content itself is excellent, it's just not in the format the template (and this repo's squash-merge convention) expects. Could you restructure it to match?

## (blocking) A commit carries an AI co-authorship trailer

The first commit (`1ff7242e`) has:

```
Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_015aTLzgTHNshpoVYygqmeFi
(cherry picked from commit ad2a5af1f2cc405ce13c921147f3a918e6eda0a5)
```

CONTRIBUTING.md/AGENTS.md don't allow `Co-Authored-By` trailers (or the session-link line) for AI tools in commit messages. This needs an amend/rebase and force-push before merge — happy to re-review as soon as that's cleaned up.

## The fix itself checks out

I verified the core claims independently rather than trusting them:

- Verified against the real botocore EC2 model: `DestinationCidrBlock`/`DestinationIpv6CidrBlock`/`DestinationPrefixListId`/`EgressOnlyInternetGatewayId` are exactly the members declared on `Route`, `CreateRouteRequest`, and `ReplaceRouteRequest` (the last one absent from `DeleteRouteRequest`, matching what you implemented). Also confirmed `CreateRoute`/`DeleteRoute`/`ReplaceRoute` genuinely have no operation-specific error shapes in the model — your disclosure about `MissingParameter` being inferred rather than confirmed is accurate.
- Verified load-bearing: reverted `matchesDestination` to dereference the stored route's field instead of comparing from the request side, and `deletingAnIpv4RouteWorksWhileAnIpv6RouteIsPresent` failed exactly as expected. Restored, confirmed clean.
- Full EC2 suite passes (442/442), both on the branch and on the actual merged-with-main tree (compiled + tested the merge — no conflicts).

One thing worth knowing, not a new issue: Greptile flagged that calling `CreateRoute` twice with the same IPv6/prefix-list destination persists a duplicate route. That's real, but it's not something this PR introduces — `main`'s existing `createRoute` has the identical `next.add(route)` with no dedup check for the IPv4 case today. This PR just extends the same (pre-existing) behavior consistently to the other two destination kinds rather than making it worse. Worth a follow-up issue if you'd like to fix it for all three at once, but I wouldn't block this PR on it.

Once the description and the commit trailer are sorted, this looks ready to me.

allensanborn:
Both addressed — thank you for the review, and particularly for verifying the claims independently rather than taking them on trust.

**Commit trailer (blocking) — fixed.** You were right, and it was a rule I should have read rather than discovered. `CONTRIBUTING.md` is explicit ("Attribution should be limited to human contributors"), and `AGENTS.md` says the same. Both commits rewritten and force-pushed: `1ff7242e` → `c41fc50a`, `3379e6cd` → `e846285b`. The `Claude-Session:` line and the stray `(cherry picked from commit ...)` noise are gone too.

Only the messages changed — I verified `git diff` between the old and new heads was empty before pushing, so the tree you reviewed is byte-identical.

I have audited our other open PRs for the same defect and am cleaning them up; #2491, #2493 and #2494 all carried it. Apologies for the churn across your review queue.

**Description template — fixed.** Restructured into `## Summary` / `## Type of change` / `## AWS Compatibility` / `## Checklist`. I had not looked at `.github/pull_request_template.md`; the "content is excellent, just not in the format" framing was generous, and I have kept the substance intact, including the `MissingParameter` disclosure.

**On the duplicate destinations:** agreed, and thank you for checking whether this PR introduced it. I had independently reached the same conclusion — the `createRoute` on `main` has the identical `next.add(route)` with no dedup for IPv4 — so it is pre-existing rather than a regression. I actually had a fix drafted here and have **backed it out** on your guidance, to keep this PR to its stated scope.

Worth noting for whoever picks up the follow-up: `RouteAlreadyExists` already exists in this file on the transit gateway route path, so there is in-repo precedent for both the error code and the shape of the check. The reason it is more than cosmetic is that duplicates make the destination-based matching this PR relies on ambiguous — `ReplaceRoute` updates only the first copy while `DeleteRoute` removes every copy. Happy to open that as an issue, or as a PR once I am back under the open-PR limit.

pgermosen (approved):
Both blocking items from the earlier review are genuinely resolved, verified independently:

- **AI co-authorship trailer** — confirmed removed from both rewritten commits (`c41fc50a`, `e846285b`); no `Co-Authored-By`/`Claude-Session`/cherry-pick noise anywhere in the current history. Diffed the old commit tree against the new rewritten one directly (`git diff --stat` old-head vs. new-head) — empty, confirming the rewrite was message-only and no code changed underneath.
- **PR description template** — now properly structured (`## Summary` / `## Type of change` / `## AWS Compatibility` / `## Checklist`).

Re-verified against current `main` (which has moved since the last pass): clean auto-merge with real overlapping content in `Ec2Service.java` and `CloudFormationResourceProvisioner.java`, `test-compile` clean, and the full EC2 suite passes on the merged-but-uncommitted tree. LGTM.

PR #2493, opened 2026-08-23
fix(s3): resolve virtual-hosted buckets whose names contain dots

allensanborn:
## Summary

A virtual-hosted-style request for any bucket whose name contains a dot resolves to the service root and answers **200**, reporting that a bucket exists when it does not. `S3VirtualHostFilter` now matches the endpoint host as a suffix instead of assuming the bucket is a single label, so dotted bucket names resolve.

```
$ curl -sI -o /dev/null -w '%{http_code}\n' http://localhost:4566/ -H 'Host: nodots-nonexistent.localhost'
404      # correct
$ curl -sI -o /dev/null -w '%{http_code}\n' http://localhost:4566/ -H 'Host: one.dot-nonexistent.localhost'
200      # wrong — this bucket does not exist
```

**Cause.** `extractBucket` split the Host header at the **first** dot and required the remainder to equal the endpoint host:

```java
String firstLabel = hostname.substring(0, firstDot);
String remainder  = hostname.substring(firstDot + 1);
if (baseHostname != null && matchesEndpointHost(remainder, baseHostname)) return firstLabel;
```

For `my.bucket.name.localhost` the remainder is `bucket.name.localhost`, which matches no known endpoint host, so `extractBucket` returned `null`, the request degraded to a path-style request against `/`, and `HEAD /` answered 200. Dots are legal in S3 bucket names, and naming a website or log bucket after a domain is the conventional pattern.

**Why it is easy to miss.** Path-style addressing is correct throughout, and the AWS CLI defaults to path-style against a custom endpoint — so this cannot be reproduced with the CLI. Only virtual-hosted-style clients hit it. The Terraform AWS provider calls `HeadBucket` before `CreateBucket` and reads the 200 as "bucket exists", failing with `BucketAlreadyExists` for a bucket that never existed.

**Change.** The bucket is everything in front of `".<endpoint-host>"`, minus an optional S3 endpoint qualifier. All four resolution paths (configured base hostname, the always-on `localhost`, configured DNS suffixes, and the AWS-domain fallback) share one implementation, so region-qualified forms work with dotted names everywhere. Where several known suffixes match, the longest wins, so `bucket.localhost.floci.io` is not read as `bucket.localhost`. Preserved unchanged: `null` for path-style requests, the IPv4 skip, port stripping, and the `isS3ServiceEndpointHost` guard.

The diff is larger than the one-line bug suggests, because the first-label assumption was duplicated across four resolution paths; fixing one and leaving the others would have produced inconsistent behaviour between `bucket.localhost` and `bucket.s3.us-east-1.localhost`. Happy to split it if you would prefer.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behaviour being fixed:** `HeadBucket`/`GetBucket*` against a virtual-hosted Host header for a dotted bucket name were answered by the `ListBuckets` route at `/`, so a nonexistent bucket returned 200 instead of 404. After this change the bucket is parsed out of the Host header and the request routes to the bucket, returning 404 when it does not exist.

**Endpoint grammar** is taken from the Amazon S3 endpoint tables in the AWS General Reference (`docs.aws.amazon.com/general/latest/gr/s3.html`), not inferred. The qualifier tails accepted are exactly:

```
s3                          s3-fips.<region>            s3-accelerate
s3.<region>                 s3-fips.dualstack.<region>  s3-accelerate.dualstack
s3.dualstack.<region>       s3-<region>      (legacy)
s3-website                  s3-website.<region>   (newer regions)
                            s3-website-<region>   (older regions)
```

Note the website endpoint spells its region with a dot in newer regions and a dash in older ones (`s3-website.eu-west-2` vs `s3-website-eu-west-1`); both are in service and both are accepted. `s3-accesspoint` and `s3-control` are deliberately absent — those hosts are prefixed by an access point name or an account id, not a bucket, and `filter` already routes S3 Control away by path.

**The guard, and why it exists.** Naive suffix matching swallows *other services'* virtual hosts. Once the bucket is "everything before the endpoint host", `123456789012.dkr.ecr.us-east-1.localhost` — the registry URI `EcrRegistryManager` hands out — parses as a dotted bucket. So do `<domain>.<region>.es.localhost`, `<id>.iot.<region>.localhost`, and every other regional service. The old first-label rule excluded them only by accident. Two guards, in order of how much they carry:

1. **A real AWS region id in a non-leading position, with no S3 qualifier, means it is not a bucket.** A genuine S3 virtual host always carries an `s3` qualifier; another service's regional host puts a bare region there instead. One structural rule covers every regional service at once, including ones Floci does not model yet. When a qualifier *is* present the region is read from there, so `logs.us-east-1.s3.localhost` stays addressable.
2. **`NON_S3_SERVICE_LABELS`** (`execute-api`, `lambda-url`, `emr-serverless`, `cloudfront`) for the *regionless* forms the region rule cannot catch. Each entry names a filter in this repository that claims that hostname. A label denylist is not a sound guard on its own — it has to enumerate every service, and the first one it misses is silently served as a bucket — so it is deliberately kept to the four cases that need it. Speculative entries cost real bucket names: `my.elb.logs` is AWS's own load-balancer access-log naming convention. `ApiGatewayExecuteApiHostIntegrationTest` and `ApiGatewayV2UnsignedAccountRoutingTest` cover the API Gateway case end to end.

`CloudFrontRequestRouter.bucketFromHost` already resolves dotted hosts correctly and explicitly refuses to truncate an unrecognised one; this brings S3 in line with that precedent.

## Changed in response to review

Two bugs were found in the first three commits of this PR. Both are fixed here, each with regression tests that fail without the production change.

**1. An internal `s3` label truncated the bucket (cross-bucket misrouting).** `bucketFromPrefix` scanned labels one at a time and split at the *last* one matching `s3`/`s3-website*`. For `my.s3.archive.localhost` that found `s3` at index 1 and returned the bucket `my` — so a request for `my.s3.archive` silently read and wrote a **different, existing** bucket. That is worse in kind than the 200 this PR set out to fix: the client gets a plausible success against the wrong data, with no signal that anything went wrong.

The rule now implemented is that a qualifier is a **complete tail, not a label**: labels from index *i* onwards must account for the whole remainder of the prefix and spell one of the published endpoint forms above. Nothing legal follows `s3` except a region, `dualstack.<region>`, or the end of the name — so `s3.archive` is bucket text, and `my.s3.archive` is the bucket. The genuine forms still split: `my.s3.archive.s3.us-east-1.localhost` resolves to `my.s3.archive`.

This also widened qualifier support as a side effect of stating the grammar properly. `s3-fips.<region>`, `s3-accelerate`, the legacy `s3-<region>` dash form and the newer `s3-website.<region>` dot form were previously unrecognised and are now handled.

The remaining ambiguity is narrower than before and is asserted: a bucket whose own name *is* a complete qualifier tail (`foo.s3`, `foo.s3.us-east-1`) is indistinguishable from a qualified reference to `foo` and is read as the latter. A bucket that merely *contains* `s3` is no longer ambiguous and is no longer truncated.

**2. The region guard matched a region *shape*, rejecting legal buckets.** The guard tested each label against `[a-z]{2}-[a-z-]+-\d+`. That is the shape of an AWS region id, but it is also the shape of ordinary bucket labels: `data.my-cd-1` is a legal bucket, and `my-cd-1` matches. The host was rejected, fell back to path-style, and **restored the exact false "bucket exists" 200 this PR was opened to fix** — for a different class of name.

The guard now tests against the finite list of real AWS region ids, added as `AwsRegions.KNOWN_IDS`. This is deliberately distinct from the existing `AwsRegions.ALL`, and a superset of it: `ALL` is what the emulator *advertises* through DescribeRegions, `KNOWN_IDS` is what it *recognises* in a hostname. A test pins `ALL ⊆ KNOWN_IDS`. `data.my-cd-1`, `reports.eu-team-2`, `assets.us-west-9` and `logs.ap-corp-1.archive` now resolve; `123456789012.dkr.ecr.us-east-1.localhost` and `search-x.eu-north-1.es.localhost` are still refused.

Position within the host was considered as an additional discriminator and rejected: it does not separate the two cases. The region label is last in `<acct>.dkr.ecr.<region>` but medial in `<domain>.<region>.es`, so no positional rule covers both without also covering bucket names.

**Residual cost of the region rule** — every rule here has one, and both sides of this one are real:

- A bucket whose name carries a **real** region id in a non-leading position (`data.us-east-1`) is still not reachable in the *unqualified* virtual-hosted form. It remains reachable qualified (`data.us-east-1.s3.localhost`) and path-style, and it was equally unreachable before this PR. This is a recoverable client-side failure, and it is asserted in `aBucketNamedAfterARealRegionIsStillTheDocumentedCost`.
- A regional service host in a region AWS launches **after** `KNOWN_IDS` was last updated would be read as a bucket. That is the maintenance cost of a list over a pattern, and it is the direction in which the previous shape-matching rule was safer. It is bounded: the failure is a wrong route rather than data loss, the four `NON_S3_SERVICE_LABELS` entries still cover the services Floci itself routes by Host, and the list is a single sorted constant with a stated source.

I took the list over the pattern because the pattern's failure mode is the one this PR exists to eliminate — a legal bucket silently answering 200 from the service root — while the list's failure mode is a mis-route that only appears for a region AWS has not launched yet.

## Checklist

**Tests.** `S3VirtualHostFilterTest` 135 → 167 tests: single-label, dotted, deeply dotted, dot+hyphen, case-insensitive suffix with bucket case preserved, path-style, IPv4, service endpoints including region-qualified, every published qualifier form, AWS domains, configured extra suffixes, port-bearing hosts, other services' hosts, and a filter-level path-rewrite assertion. New `AwsRegionsTest` (14 tests) pins `ALL ⊆ KNOWN_IDS` and the id-vs-shape distinction.

**Revert check.** Reverting each production rule while keeping the tests, to confirm the tests actually hold the fix:

- qualifier-tail rule reverted to the per-label scan → **12 failures**, e.g. `expected: <my.s3.archive> but was: <my>`, `expected: <www.s3.example.com> but was: <www>`
- region-id rule reverted to the `[a-z]{2}-[a-z-]+-\d+` pattern → **6 failures**, e.g. `expected: <data.my-cd-1> but was: <null>`

Also, the commit trailers `CONTRIBUTING.md` forbids have been stripped from every commit on this branch (`git filter-branch --msg-filter`, verified to leave the trees byte-identical). Apologies for those — same defect as #2492.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/AwsRegions.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/AwsRegions.java
@@ -1,6 +1,8 @@
 package io.github.hectorvent.floci.core.common;
 
 import java.util.List;
+import java.util.Locale;
+import java.util.Set;
 
 /**
  * Catalog of AWS regions the emulator advertises. Single source of truth for
@@ -14,6 +16,43 @@ public final class AwsRegions {
             "eu-west-1", "eu-west-2", "eu-west-3", "eu-central-1",
             "ap-northeast-1", "ap-northeast-2", "ap-southeast-1", "ap-southeast-2",
             "ap-south-1", "sa-east-1", "ca-central-1");
+
+    /**
+     * Every published AWS region id, across the commercial, GovCloud and China partitions.
+     *
+     * <p>Distinct from {@link #ALL}, and deliberately a superset of it. {@code ALL} is what
+     * this emulator <em>advertises</em> — what DescribeRegions returns. {@code KNOWN_IDS} is
+     * for <em>recognising</em> a region id that a client wrote, most importantly inside a
+     * hostname, where the alternative is a shape like {@code [a-z]{2}-[a-z-]+-\d+} that also
+     * matches perfectly ordinary strings ({@code my-cd-1}, {@code eu-team-2}).
+     *
+     * <p>Sourced from the AWS General Reference endpoint tables. It needs an entry when AWS
+     * launches a region; the cost of a missing one is documented at each call site, because it
+     * differs per caller.
+     */
+    public static final Set<String> KNOWN_IDS = Set.of(
+            "us-east-1", "us-east-2", "us-west-1", "us-west-2",
+            "af-south-1",
+            "ap-east-1", "ap-east-2",
+            "ap-south-1", "ap-south-2",
+            "ap-northeast-1", "ap-northeast-2", "ap-northeast-3",
+            "ap-southeast-1", "ap-southeast-2", "ap-southeast-3", "ap-southeast-4",
+            "ap-southeast-5", "ap-southeast-6", "ap-southeast-7",
+            "ca-central-1", "ca-west-1",
+            "eu-central-1", "eu-central-2",
+            "eu-west-1", "eu-west-2", "eu-west-3",
+            "eu-north-1", "eu-south-1", "eu-south-2",
+            "il-central-1",
+            "me-central-1", "me-south-1",
+            "mx-central-1",
+            "sa-east-1",
+            "us-gov-east-1", "us-gov-west-1",
+            "cn-north-1", "cn-northwest-1");
+
+    /** True when {@code label} is exactly an AWS region id, case-insensitively. */
+    public static boolean isRegionId(String label) {
+        return label != null && KNOWN_IDS.contains(label.toLowerCase(Locale.ROOT));
+    }
 
     private AwsRegions() {
     }
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3VirtualHostFilter.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3VirtualHostFilter.java
@@ -13,6 +13,7 @@ import jakarta.ws.rs.core.UriBuilder;
 import jakarta.ws.rs.ext.Provider;
 
 import java.net.URI;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
@@ -34,6 +35,19 @@ public class S3VirtualHostFilter implements ContainerRequestFilter {
      * alongside plain {@code localhost}. Stored lowercase for case-insensitive matching.
      */
     private final Set<String> serviceHostSuffixes;
+
+    /**
+     * Host labels belonging to other AWS services that virtual-host under the same endpoint
+     * host as S3. A hostname carrying one of these is that service's, never an S3 bucket
+     * whose name happens to contain dots — see {@link #bucketFromPrefix}.
+     */
+    private static final Set<String> NON_S3_SERVICE_LABELS = Set.of(
+            "execute-api",      // API Gateway v1/v2 + WebSocket: <api-id>.execute-api.<region>.<host>
+            "lambda-url",       // Lambda function URLs: <url-id>.lambda-url.<region>.<host>
+            "emr-serverless",   // EMR Serverless: emr-serverless.<host>
+            "cloudfront",       // CloudFront distribution domains: <dist-id>.cloudfront.net
+            "mwaa",
+            "elb");
 
     @Inject
     public S3VirtualHostFilter(EmulatorConfig config, ContainerDetector containerDetector) {
@@ -142,19 +156,39 @@ public class S3VirtualHostFilter implements ContainerRequestFilter {
     /**
      * Extracts a bucket name from a virtual-hosted-style Host header.
      *
-     * A request is considered virtual-hosted-style when the hostname's remainder
-     * after the first label matches the configured Floci base hostname, or when it
-     * matches a well-known AWS S3 domain pattern (for DNS-redirect setups).
+     * <p>A request is considered virtual-hosted-style when the hostname <em>ends with</em> a
+     * known endpoint host — the configured Floci base hostname, plain {@code localhost}, a
+     * configured DNS suffix, or a well-known AWS S3 domain (for DNS-redirect setups). The
+     * bucket is everything in front of that suffix, minus an optional S3 qualifier
+     * ({@code s3}, {@code s3.<region>}, {@code s3-website-<region>}, …).
+     *
+     * <p><strong>The bucket is not just the first label.</strong> S3 bucket names may contain
+     * dots, and naming a website bucket after a domain is the conventional pattern, so
+     * {@code www.example.com.localhost} is the bucket {@code www.example.com} — not the bucket
+     * {@code www} with an unrecognised remainder. Splitting at the first dot made every dotted
+     * bucket fall through to path-style, where {@code HEAD /} answers 200 and the AWS provider
+     * reads that as "bucket already exists".
      *
      * Examples with baseHostname="localhost":
      *   my-bucket.localhost:4566       -> "my-bucket"
      *   my-bucket.localhost            -> "my-bucket"
+     *   www.example.com.localhost      -> "www.example.com"   (dotted bucket)
+     *   my.bucket-logs.localhost:4566  -> "my.bucket-logs"    (dots and hyphens)
+     *   my-bucket.s3.us-east-1.localhost -> "my-bucket"       (region-qualified)
+     *   www.example.com.s3.us-east-1.amazonaws.com -> "www.example.com"
+     *   s3.localhost                   -> null  (service endpoint, not a bucket named "s3")
+     *   s3.us-east-1.localhost         -> null  (region-qualified service endpoint)
      *   floci.svc.cluster.local        -> null  (no bucket prefix, path-style)
-     *   my-svc.floci.svc.cluster.local -> null  (remainder doesn't match "localhost")
+     *   my-svc.floci.svc.cluster.local -> null  (does not end with a known endpoint host)
      *
      * Examples with baseHostname="floci.svc.cluster.local":
      *   my-bucket.floci.svc.cluster.local -> "my-bucket"
      *   floci.svc.cluster.local           -> null  (no bucket prefix, path-style)
+     *
+     * <p>Inherent ambiguity: a bucket whose <em>own</em> name ends in a qualifier label
+     * ({@code foo.s3}) is indistinguishable from a qualified reference to {@code foo} and is
+     * read as the latter. The service endpoint always wins over the bucket reading, so
+     * {@code s3.localhost} stays bucketless.
      *
      * Returns null if the host does not match a virtual-hosted pattern.
      */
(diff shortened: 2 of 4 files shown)


pgermosen:
Thanks for tracking down this bug and for the thorough write-up of the design tradeoffs — the reasoning about why suffix-matching needs the two guards (regional-service labels and the explicit denylist) is genuinely well thought out. A few things need addressing before this is ready, though:

## PR description doesn't follow the template

The repo's `.github/pull_request_template.md` expects `## Summary`, `## Type of change` (checkboxes), `## AWS Compatibility`, and `## Checklist` sections. This PR uses a different, custom structure (`## Problem`, `## Cause`, `## Change`, etc.) instead. The content itself is thorough, it's just not in the format the template — and this repo's squash-merge convention — expects. Could you restructure it to match?

## (blocking) Commit carries an AI co-authorship trailer

The first commit (`82f34a63`) has:

```
Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
(cherry picked from commit 5c302135929238a3d60bb11f4a855f490358dc85)
```

CONTRIBUTING.md/AGENTS.md don't allow `Co-Authored-By` trailers for AI tools in commit messages. This needs an amend/rebase and force-push before merge — happy to re-review as soon as that's cleaned up.

## (blocking) Two real bugs in the qualifier/region-matching logic

I verified both of these directly by calling `S3VirtualHostFilter.extractBucket` — they're not hypothetical:

**1. A bucket with an internal `s3` label gets truncated to the wrong bucket.** `bucketFromPrefix`'s qualifier scan (`S3VirtualHostFilter.java:340-345`) finds *any* label matching `s3`/`s3-website*` anywhere in the prefix and treats it as the trailing S3 qualifier, discarding everything after it — without checking that the match is actually anchored at the tail:

```
extractBucket("my.s3.archive.localhost", "localhost", DEFAULT_SUFFIXES) → "my"
```

The real bucket is `my.s3.archive` (there's no genuine S3 qualifier here — `s3` just happens to appear as a bucket-name segment). A `PutObject`/`GetObject`/`HeadBucket` against this host would silently target the wrong bucket (`my`) instead of failing loudly or hitting the right one. This is exactly the kind of internal-qualifier case Greptile flagged.

**2. A bucket name that merely looks region-shaped gets rejected back into the original bug.** The `REGION_LABEL` regex (`[a-z]{2}-[a-z-]+-\d+`, line 70-71) is much broader than real AWS region names — it matches any hyphenated label ending in digits with a 2-letter prefix:

```
extractBucket("data.my-cd-1.localhost", "localhost", DEFAULT_SUFFIXES) → null
```

`null` means it degrades to path-style and hits the service root — which is precisely the false "bucket exists" 200 this PR sets out to fix, just for a different class of dotted names (anything containing a segment like `my-cd-1`, `ab-test-2`, `ci-build-9`, etc.).

Both are real correctness issues in the core logic this PR is adding, not edge-case nitpicks — happy to help think through a fix if useful (e.g. requiring the qualifier match to be anchored at the tail of the prefix for #1, and narrowing the region heuristic to a real region-name list or requiring it to consume the *entire* remaining prefix for #2).

## Not blocking, but worth confirming

The AWS-provider `elb`/`mwaa` removal from the denylist (second commit) looks right based on the reasoning given — worth a quick sanity check that no filter in this repo actually claims those hostnames before merge, since I didn't independently verify that part.

Everything else — the suffix-consolidation approach, the CloudFront precedent, the four-way path unification — looks solid. Happy to take another pass once the trailer is cleaned up and the two matching bugs are fixed.

pgermosen (approved):
Follow-up to the earlier review — everything flagged then is genuinely resolved now, verified directly rather than taken on trust:

1. **Description now matches the template** (Summary/Type of change/AWS Compatibility/Checklist).
2. **AI co-authorship trailers stripped** — confirmed across all 5 commits, none remain.
3. **Both confirmed bugs fixed** — re-ran my exact original reproduction cases against the new code:
   - `my.s3.archive.localhost` now resolves to `my.s3.archive` (was `my` — the wrong-bucket misrouting bug).
   - `data.my-cd-1.localhost` now resolves to `data.my-cd-1` (was `null` — the false-200 regression this PR exists to fix).

New checks this pass:
- `AwsRegions.KNOWN_IDS` (37 region IDs) verified against a fresh lookup of the current AWS region list — exact match, entry for entry.
- The `s3-website.<region>` vs `s3-website-<region>` dot/dash split verified against the real AWS General Reference endpoint tables, region by region — exact match.

Full S3 suite + AwsRegionsTest (935+14 tests) pass clean, both on the branch and on the actual merged-with-main tree (compiled + tested the merge, no conflicts). LGTM.

pgermosen:
Follow-up note for the record, beyond what's in the approval: this pass I also independently cross-checked `AwsRegions.KNOWN_IDS` against the current published AWS region list (exact match, no missing/extra entries) and walked the `s3-website.<region>` vs `s3-website-<region>` dot/dash split in `isS3QualifierTail` against the AWS General Reference endpoint tables region by region (also exact match). Both were spot checks beyond the two bugs originally flagged, in case the region list or endpoint-table shape drifts in the future — no action needed now, just flagging where the ground truth was sourced from for whoever touches this next.

allensanborn:
Belated answer to the one thing you flagged as unverified — apologies for letting it go past the merge without a reply.

> worth a quick sanity check that no filter in this repo actually claims those hostnames before merge, since I didn't independently verify that part

Checked, and the removal is safe. Every `Host`-dispatching filter on `main`:

```
ApiGatewayCustomDomainFilter   ApiGatewayExecuteApiHostFilter
CloudFrontDistributionFilter   EmrServerlessRouteFilter
LambdaUrlRoutingFilter         S3VirtualHostFilter
SqsQueueUrlRouterFilter        AppConfigConfigurationRouteFilter
```

None is an ELB or MWAA filter, and there is no `*Filter.java` under `services/elb*` or `services/mwaa` at all.

MWAA is the one that looked like it might, so I checked it specifically: its only host-adjacent code is `services/mwaa/proxy/MwaaWebProxy.java`, and it does not claim a hostname on the main listener — it binds its own `proxyPort` and rewrites `Host` to `backendHost:backendPort` on the way *out*. So nothing in the repo routes `*.elb.<host>` or `*.mwaa.<host>` by `Host`, and the two entries were costing real bucket names for nothing. `my.elb.logs` is AWS's own load-balancer access-log naming convention.

Worth stating the shape of the residual risk plainly, since it is the same for the four entries that remain: the denylist is only sound while it matches filters that actually exist. If someone later adds an ELB or MWAA host filter, this guard will not know, and the bucket/service collision comes back silently. The four remaining entries each name a filter in the list above, which is the invariant — but nothing enforces it. A test asserting that `NON_S3_SERVICE_LABELS` corresponds one-to-one with the host-dispatching filters would make it self-maintaining; happy to send that as a small follow-up if you think it earns its keep.

PR #2494, opened 2026-08-23
fix(kinesis): apply CreateStream's Tags instead of dropping them

allensanborn:
## Summary

`CreateStream` parsed `StreamName`, `ShardCount`, `StreamModeDetails` and `MaxRecordSizeInKiB` and silently ignored the optional `Tags` member, so a stream created with tags came back untagged from `ListTagsForStream`. The tag store itself was fine — `AddTagsToStream` on the same stream persisted correctly — only the create path dropped the parameter.

`CreateStream` now applies the request's `Tags`, reusing the existing tag store rather than adding a second path. The parse is factored into a private `parseTags(JsonNode)` shared by `CreateStream` and `AddTagsToStream`, following the `parseTags` helpers `CodeBuildJsonHandler` and `SecretsManagerJsonHandler` already have. Behaviour is unchanged for `AddTagsToStream`; the point is that the two operations cannot answer the same request differently.

This is the same class of bug as #2301 (`msk: CreateCluster discards ... tags`), on a different service.

### Scope — deliberately left out

- **Tag validation.** `asText()` coerces a non-string tag value, and nothing enforces AWS's 50-tags-per-stream cap. Both are pre-existing behaviour of `AddTagsToStream`, not introduced here. The Kinesis service model types `TagValue` as a `string`, so no AWS SDK can emit the input this describes; sharing one parser is what keeps the two paths from answering the same malformed request differently. Rejecting non-string tag values is a separate input-validation fix that should land on its own, across every tag operation.
- **`RegisterStreamConsumer` also declares `Tags` and also drops them**, and `KinesisConsumer` has no tag field to put them in. `TagResource` / `UntagResource` / `ListTagsForResource` are unimplemented. Separate work.

### Tests

New handler-level unit tests, which the class had none of for tags: create with tags, create without, create with an empty `Tags` object, and `AddTagsToStream` applying the same map. Unlike the ordered integration tests, these get a fresh service per test.

Confirmed adversarially by reverting the production change: `KinesisIntegrationTest.createStreamAppliesTagsFromTheRequest` and `KinesisJsonHandlerTest.createStreamAppliesTagsFromTheRequest` both go red, while the untagged-stream guards stay green.

Beyond unit tests, this flips a real end-to-end suite: `terraform-aws-messaging`'s `TestKinesis` goes **FAIL -> PASS**, all four subtests, driven by Terratest against a local emulator.

`./mvnw test -Dtest='*Kinesis*'` -> 248 tests, 0 failures. Full suite: `./mvnw test` -> 11568 tests, 0 failures.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behaviour** — tags supplied to `CreateStream` were accepted and discarded:

```
$ aws kinesis create-stream --stream-name repro --shard-count 1 --tags Foo=Bar
$ aws kinesis list-tags-for-stream --stream-name repro
{"Tags": [], "HasMoreTags": false}
```

`kinesis/service-2.json` declares `Tags` as a member of `CreateStream`:

```
CreateStream members: [StreamName, ShardCount, StreamModeDetails, Tags, WarmThroughputMiBps, MaxRecordSizeInKiB]
required: [StreamName]
```

**Real-world impact.** Terraform's `aws_kinesis_stream` sets tags at create time and reads them back on refresh, so a dropped `Tags` is a permanent diff — every plan after a successful apply reports "tags will be updated in-place" on an unchanged configuration and the resource never converges. That is what fails `terraform-aws-messaging`'s `TestKinesis`, whose four subtests all assert `terraform plan` returns exit code 0 after apply; every other assertion in that test (shard count, encryption type, retention period) already passed.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -120,10 +120,8 @@ public class KinesisJsonHandler {
         // hand-written script but not to Terraform: aws_kinesis_stream sets tags at create
         // time and then reads them back with ListTagsForStream, so an empty read produces a
         // permanent "tags will be updated in-place" diff on an unchanged configuration.
-        JsonNode tagsNode = request.path("Tags");
-        if (tagsNode.isObject() && !tagsNode.isEmpty()) {
-            Map<String, String> tags = new HashMap<>();
-            tagsNode.fields().forEachRemaining(e -> tags.put(e.getKey(), e.getValue().asText()));
+        Map<String, String> tags = parseTags(request);
+        if (!tags.isEmpty()) {
             service.addTagsToStream(streamName, tags, region);
         }
         return Response.ok(objectMapper.createObjectNode()).build();
@@ -398,9 +396,7 @@ public class KinesisJsonHandler {
 
     private Response handleAddTagsToStream(JsonNode request, String region) {
         String streamName = resolveStreamName(request);
-        Map<String, String> tags = new HashMap<>();
-        request.path("Tags").fields().forEachRemaining(entry -> tags.put(entry.getKey(), entry.getValue().asText()));
-        service.addTagsToStream(streamName, tags, region);
+        service.addTagsToStream(streamName, parseTags(request), region);
         return Response.ok(objectMapper.createObjectNode()).build();
     }
 
@@ -424,6 +420,17 @@ public class KinesisJsonHandler {
         });
         response.put("HasMoreTags", false);
         return Response.ok(response).build();
+    }
+
+    // Shared by CreateStream and AddTagsToStream so the two paths cannot drift: the same
+    // request shape has to produce the same stored tags whichever operation carries it.
+    // A missing or non-object Tags member yields an empty map rather than an error, which
+    // is what AddTagsToStream already did before CreateStream started calling this.
+    private Map<String, String> parseTags(JsonNode request) {
+        Map<String, String> tags = new HashMap<>();
+        request.path("Tags").fields()
+                .forEachRemaining(entry -> tags.put(entry.getKey(), entry.getValue().asText()));
+        return tags;
     }
 
     private Response handleStartStreamEncryption(JsonNode request, String region) {
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -116,6 +116,16 @@ public class KinesisJsonHandler {
         }
         service.createStream(streamName, shardCount, streamMode,
                 optionalMaxRecordSize(request), region);
+        // CreateStream's optional Tags member was being dropped. That is invisible to a
+        // hand-written script but not to Terraform: aws_kinesis_stream sets tags at create
+        // time and then reads them back with ListTagsForStream, so an empty read produces a
+        // permanent "tags will be updated in-place" diff on an unchanged configuration.
+        JsonNode tagsNode = request.path("Tags");
+        if (tagsNode.isObject() && !tagsNode.isEmpty()) {
+            Map<String, String> tags = new HashMap<>();
+            tagsNode.fields().forEachRemaining(e -> tags.put(e.getKey(), e.getValue().asText()));
+            service.addTagsToStream(streamName, tags, region);
+        }
         return Response.ok(objectMapper.createObjectNode()).build();
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/kinesis/KinesisIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/kinesis/KinesisIntegrationTest.java
@@ -1133,6 +1133,66 @@ class KinesisIntegrationTest {
             .body("Records.size()", equalTo(0));
     }
 
+    @Test
+    @Order(65)
+    void createStreamAppliesTagsFromTheRequest() {
+        // CreateStream's Tags member used to be dropped, which Terraform surfaces as a
+        // permanent tags diff: aws_kinesis_stream tags at create time, then reads back
+        // with ListTagsForStream and finds nothing.
+        given()
+            .header("X-Amz-Target", "Kinesis_20131202.CreateStream")
+            .contentType(KINESIS_CONTENT_TYPE)
+            .body("""
+                {"StreamName": "create-with-tags", "ShardCount": 1,
+                 "Tags": {"Foo": "Bar", "gw:example": "kinesis"}}
+                """)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200);
+
+        given()
+            .header("X-Amz-Target", "Kinesis_20131202.ListTagsForStream")
+            .contentType(KINESIS_CONTENT_TYPE)
+            .body("""
+                {"StreamName": "create-with-tags"}
+                """)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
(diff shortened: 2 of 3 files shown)


allensanborn:
CI note, not a test failure — could a maintainer re-run **Build and Test**? I do not have permission to.

The job was killed by the workflow timeout rather than by a failing test:

- `ci.yml` sets `timeout-minutes: 20`
- the job ran **20m16s** and the `Run tests` step reports `conclusion: cancelled`, which is how GitHub records a timeout kill rather than a failure

Every other check on this PR is green, including all five native SDK suites and `compat-terraform` / `compat-opentofu`, and the same `Build and Test` job passed on my three sibling PRs (#2491, #2492, #2493) opened within a minute of this one. The change here is a few lines in `KinesisJsonHandler` plus tests, so it is not plausibly responsible for pushing the suite over 20 minutes.

Happy to rebase or push if you would rather re-trigger that way — `upstream/main` has not moved since I opened it, so a rebase would be a no-op and would not start a new run.

pgermosen:
Reviewed the code — it's solid, and I verified the core claims independently rather than trusting the writeup:

- Pulled the real `kinesis/2013-12-02/service-2.json`: `CreateStreamInput` lists `Tags` as a member (only `StreamName` is `required`, matching the PR's claim exactly).
- Confirmed the non-string-tag-value coercion (`asText()`) genuinely predates this PR — traced it back to the very first commit of `KinesisJsonHandler.java` via `git log -S`. So scoping that out as a separate, pre-existing issue is accurate, not a rationalization.
- Did my own revert-and-confirm-fail rather than trusting the PR's own claim: reverted the `CreateStream`→tag-apply call and both `createStreamAppliesTagsFromTheRequest` tests (handler + integration) failed with exactly the expected symptom (`expected: <{gw:example=kinesis, Foo=Bar}> but was: <{}>`). Restored, full `kinesis` suite (248 tests) green, including on a fresh merge against current `main` — no conflicts here.

No code findings. Two hygiene items, same as #2491:

1. **PR description skips the template** — no `## Type of change` checkboxes, no `## Checklist`, no `Closes #N`. Same custom Problem/Why/Change/Scope/Verification structure as your other PRs opened around the same time. Not asking for a rewrite, just flagging it's missing the standard sections.
2. **Both commits carry `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>` trailers** (one also has a `Claude-Session:` link) — same pattern as your other open PR #2491. CONTRIBUTING.md and AGENTS.md both say: "Do not include `Co-Authored-By` trailers for AI tools in commit messages." Could you amend/rebase across this and your other open PRs and force-push? This is the actual blocker before merge — the code itself is ready.

On the CI timeout you flagged: `Build and Test` is green now (ran ~20 minutes on the latest run), so that seems to have resolved on its own.

hectorvent:
@allensanborn please rebase main, CI timeout was fixed.

allensanborn:
Thanks @hectorvent — rebased onto current `main` (`257e0046`). Clean rebase, no conflicts; the patch is byte-identical to what you reviewed.

New head: `07c94e5a`

Locally on the rebased branch: `*Kinesis*` is 248 tests, 0 failures, 0 errors, 0 skipped. CI is re-running now on the new 4-way shard matrix.

allensanborn:
@pgermosen — following up on the trailer blocker, which I should have confirmed explicitly rather than only mentioning the rebase. My earlier note said the patch was byte-identical to what you reviewed, which is true of the *diff* but understates what changed about the commits: **the `Co-Authored-By` and `Claude-Session:` trailers are gone.**

Current head is `07c94e5a`. Verified from the API rather than locally, so it reflects what you'd see:

```
gh api repos/floci-io/floci/pulls/2494/commits --jq '.[].commit.message' \
  | grep -ciE 'co-authored-by|claude|generated with'
0
```

I ran the same check across all five of my open PRs (#2491, #2494, #2526, #2527, #2528) — zero hits on every one, so the amend you asked for is done everywhere, not just here.

All checks are green on this head, and `*Kinesis*` is 248 tests, 0 failures locally.

Could you drop the `waiting-contributor` label when you get a chance? I think it's just outlived the thing it was tracking — it went on within seconds of your review, so it predates the fix rather than responding to it.

On the PR-description template: leaving that as-is per your note that you weren't asking for a rewrite. Happy to add the standard sections if it turns out to matter for tracking.

pgermosen (approved):
Re-checked at `07c94e5a`, merged onto current `main`, built, and probed against a `main` baseline on a second port.

## The blocker is closed

Checked the commit objects rather than re-running your grep, since that is what I actually asked about:

```
07c94e5a refactor(kinesis): parse CreateStream and AddTagsToStream tags with one helper
930f77f5 fix(kinesis): apply CreateStream's Tags instead of dropping them
    author/committer: Allen Sanborn <allen.sanborn@gmail.com>
    trailers: []
```

Both clean. The description also carries `## Type of change` and `## Checklist` now, so both hygiene items from last round are done.

## Verified on the rebased head

The reproduction from the description, run against both builds with identical commands:

```
aws kinesis create-stream --stream-name repro --shard-count 1 --tags Foo=Bar,team=core
aws kinesis list-tags-for-stream --stream-name repro

main    → {"Tags":[],"HasMoreTags":false}
branch  → {"Tags":[{"Key":"Foo","Value":"Bar"},{"Key":"team","Value":"core"}],"HasMoreTags":false}
```

**The refactor really is behaviour-preserving for `AddTagsToStream`** — I checked rather than assuming, since that is the half a shared helper could quietly change:

```
add-tags-to-stream --tags a=1,b=2  →  identical output on main and branch
```

Edge cases on the branch: `"Tags":{}` on a fresh stream returns 200 with no tags, and a request with no `Tags` member at all returns 200. (A 400 I saw first time round was `ResourceInUseException` from reusing a name, not the empty map.)

**Mutation test.** Removed the tag-apply from `CreateStream` and re-ran — two failures, exactly the two you named, and the handler one carries the same message I saw last round:

```
KinesisIntegrationTest.createStreamAppliesTagsFromTheRequest:1164
KinesisJsonHandlerTest.createStreamAppliesTagsFromTheRequest:889
    expected: <{gw:example=kinesis, Foo=Bar}> but was: <{}>
```

The untagged-stream guards stayed green, as you said. Restored afterwards.

**248 Kinesis tests, 0 failures** on the merge with current `main` — the same number you reported. Merge is clean.

## The scope exclusions check out at the model level

Two of these I could not confirm last round: the botocore bundled with my CLI is old enough that `RegisterStreamConsumerInput` has only `StreamARN`/`ConsumerName` and the `TagResource` family does not exist at all. Checking `kinesis-2.52.0` instead settles them:

| claim | model |
|---|---|
| `RegisterStreamConsumer` also declares `Tags` and drops them | `RegisterStreamConsumerRequest` has `tags()` |
| `TagResource` / `UntagResource` / `ListTagsForResource` are unimplemented | both request classes exist in the SDK |
| no SDK can emit a non-string tag value | `TagValue` is typed `string` |

And the `parseTags` convention is real rather than a post-hoc justification — a dozen handlers already have one, including the `CodeBuildJsonHandler` you cite.

## Two things worth a line, neither blocking

- **The 50-tag figure.** The scope note says "AWS's 50-tags-per-stream cap". That is the per-stream limit, but `CreateStream`'s own `Tags` member documents *"a set of up to 10 key-value pairs"* and the `TagMap` shape caps at 200. Since tag-count validation is explicitly out of scope this changes nothing — just noting the create-time number is 10 if anyone picks this up later.
- **Tagging is a second call after creation**, so a failure between `createStream` and `addTagsToStream` would leave a created-but-untagged stream. In-process, against a store that has just accepted the create, that is not a realistic failure, and making it atomic would mean threading tags through the service signature for no practical gain. Recording it rather than asking for it.

No code findings.

PR #2495, opened 2026-08-23
feat(organizations): add CloudFormation resource types

willstjohnbacon:
> Follow-up to #2489, now merged. Rebased onto current `main` — this is a **single commit against `main`**, 5 files, +1,111.

## Summary

Adds `OrganizationsCfnProvisioner` covering all five registry types:

| Resource type | `Ref` | `Fn::GetAtt` |
|---|---|---|
| `AWS::Organizations::Organization` | `o-…` | `Id`, `Arn`, `ManagementAccountArn`, `ManagementAccountId`, `ManagementAccountEmail`, `RootId` |
| `AWS::Organizations::OrganizationalUnit` | `ou-…` | `Id`, `Arn` |
| `AWS::Organizations::Account` | account id | `AccountId`, `Arn`, `Status`, `JoinedMethod`, `JoinedTimestamp` |
| `AWS::Organizations::Policy` | `p-…` | `Id`, `Arn`, `AwsManaged` |
| `AWS::Organizations::ResourcePolicy` | `rp-…` | `Id`, `Arn` |

Attribute names are taken from each type's `readOnlyProperties` in `aws-cloudformation-resource-providers-organizations`, so `Fn::GetAtt` resolves the keys AWS publishes. (The `local/aws/cfn-resource-schemas/` tree CONTRIBUTING points at is gitignored and not present in a fresh clone, so I sourced them from the provider repo — happy to switch if there's a canonical local copy I missed.)

No linked issue — happy to open one if you'd prefer to track it that way.

### One design decision worth a look

`CfnResourceProvisioner.delete` hands over a resource type, a physical id and a region — but **no caller identity**. Every Organizations mutation is management-account scoped, so the provisioner needs an account from somewhere.

Injecting `RegionResolver` alongside `OrganizationsService` would have worked, but CONTRIBUTING says a provisioner should inject *only* the service it wraps, and all 16 existing provisioners do exactly that. So instead `OrganizationsService` gained `findManagementAccountForResource(resourceId)`, which recovers the owning account from the resource id itself. That keeps the provisioner single-dependency and makes it independent of the synthetic request scope CloudFormation provisions under. Happy to switch to injection if maintainers prefer it.

### Create *and* update

`provision` serves both paths, per the contributing guide:

- **OrganizationalUnit** — `ParentId` is create-only in the schema, so an update renames in place. The physical id survives, so `Ref`s pointing at it stay valid rather than being orphaned by a silent recreate.
- **Policy** — `TargetIds` reconciles in both directions: targets the template adds are attached, ones it drops are detached.
- **Account** — moved to its `ParentIds` entry after creation (accounts are always created under the root); a no-op when it is already there.
- **ResourcePolicy** — backed by `PutResourcePolicy`, already create-or-update.
- **Organization** — `FeatureSet` is the only settable property and the only legal transition is the one-way upgrade to `ALL`, which routes through `EnableAllFeatures`.

`delete` tolerates an already-deleted resource, and detaches a policy's remaining targets first since `DeletePolicy` refuses while attached.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

This PR adds **no new AWS API actions** — the CloudFormation wire surface is unchanged, and the Organizations actions it provisions against land in #2489. So what needed verifying here is that the five resource types provision, expose the attribute keys AWS publishes, and update/delete correctly.

- **Resource schemas** — `Ref` targets and every `Fn::GetAtt` key are taken from each type's `readOnlyProperties` in the AWS `aws-cloudformation-resource-providers-organizations` registry schemas, not invented.
- **AWS CLI v2 `2.36.29`** — deployed a real stack against a `quarkus:dev` Floci containing all five types wired together (`Org.RootId` → OU `ParentId`, `Unit.Id` → account `ParentIds` and policy `TargetIds`), with **23 stack outputs each backed by a `Ref` or `Fn::GetAtt`**:
  - `CREATE_COMPLETE`, all five resources provisioned, and **all 23 outputs resolved to real values** — no output fell through to the literal `"LogicalId.Attr"`. This is the assertion that matters, since an unmapped attribute is a silent no-op rather than a failure.
  - `UpdateStack` renaming the OU returned `UPDATE_COMPLETE` and **preserved the OU physical id** (`ou-hao2-7h6euefm` before and after), with the rename visible through `describe-organizational-unit`. A recreate here would silently orphan every `Ref`.
  - `DeleteStack` tore the organization down cleanly — `describe-organization` afterwards returns `AWSOrganizationsNotInUseException`.
- ARNs render in the Organizations format, with no region and the org id embedded: `arn:aws:organizations::555566667777:ou/o-37jln6jgtk/ou-hao2-7h6euefm`, `…:policy/o-…/service_control_policy/p-…`, `…:account/o-…/989941865198`.

## Testing

- **`OrganizationsCfnProvisionerTest`** (19) — the provisioner in isolation with only `OrganizationsService` mocked, following `SqsCfnProvisionerTest`.
- **`OrganizationsCfnIntegrationTest`** (9) — a real stack exercising all five types, plus update-in-place and full teardown.

Every integration assertion goes through a stack `Output` backed by `Fn::GetAtt`, deliberately: an unmapped attribute doesn't fail provisioning, it silently resolves to the literal `"LogicalId.Attr"`. Asserting `CREATE_COMPLETE` — or even that the underlying resource exists — would pass just as happily with every attribute unwired. The test also asserts the OU's physical id is unchanged across `UpdateStack`, which is the thing that would break every `Ref` if update accidentally recreated it.

- `./mvnw test -Dtest='OrganizationsCfn*'` — 28 tests, 0 failures, 0 errors.
- `./mvnw test -Dtest='Organizations*'` — 113 tests, 0 failures, 0 errors (this PR's 28 plus #2489's 85).
- `./mvnw test` — full suite green with Docker running, so the Lambda/WebSocket/Neptune container suites execute rather than skip.
- `make docs-check` — clean. `docs/services/cloudformation.md` gains the table row and a behaviour section.

## Checklist

### DIFF
--- a/docs/services/cloudformation.md
+++ b/docs/services/cloudformation.md
@@ -134,11 +134,17 @@ the rest of the stack's Organizations resources are provisioned into that accoun
 - `Account` — `ParentIds` accepts a single entry and the account is moved there after creation
   (accounts are always created under the root). Because Floci resolves a 12-digit access key id
   straight to an account, `Ref`/`Fn::GetAtt AccountId` yields an id you can immediately use as a
-  caller identity against other services.
+  caller identity against other services. `State` mirrors `Status`, the parameter AWS is retiring
+  it in favour of; Floci does not model the `PENDING_ACTIVATION` or `CLOSED` phases that only
+  `State` can express.
 - `Policy` — `TargetIds` is reconciled on update: targets the template adds are attached and ones
   it drops are detached. Deleting the resource detaches its remaining targets first, since
   `DeletePolicy` refuses while a policy is still attached.
 - `ResourcePolicy` — backed by `PutResourcePolicy`, which is already create-or-update.
+
+`Fn::GetAtt OrganizationalUnit.Path` and `Fn::GetAtt Account.Paths` return the organization path —
+`o-<org>/r-<root>/[ou-<ou>/…]<id>/`. An account has one parent and therefore one path, so the
+list-typed `Paths` holds a single entry; read it with `Fn::Select` as you would on AWS.
 
 Deleting the stack removes the resources in dependency order and finally the organization itself.
 Deleting a resource that is already gone is tolerated.
--- a/docs/services/cloudformation.md
+++ b/docs/services/cloudformation.md
@@ -60,6 +60,7 @@ cross-resource references.
 | DynamoDB | `Table`, `GlobalTable` |
 | Lambda | `Function` (Zip via S3/inline `ZipFile`, and Image), `LayerVersion`, `EventSourceMapping` (SQS, Kinesis, DynamoDB Streams), `Version`, `Alias` (also what SAM's `AutoPublishAlias` expands into) |
 | IAM | `Role`, `User`, `AccessKey`, `Policy`, `ManagedPolicy`, `InstanceProfile` |
+| Organizations | `Organization`, `OrganizationalUnit`, `Account`, `Policy`, `ResourcePolicy` |
 | SSM | `Parameter` |
 | KMS | `Key`, `Alias` |
 | Secrets Manager | `Secret`, `SecretTargetAttachment` |
@@ -121,6 +122,26 @@ Redshift clusters, Redshift Serverless namespaces, and DocumentDB Elastic cluste
 because their backing services are not implemented. A `SecretId` change is applied in place rather
 than reproducing CloudFormation's replacement event sequence; failed changes restore affected secret
 data and attachment ownership.
+
+## Organizations
+
+`AWS::Organizations::Organization` uses the stack's calling account as the management account, so
+the rest of the stack's Organizations resources are provisioned into that account's organization.
+`Fn::GetAtt Org.RootId` is the usual way to root the OU tree in the same template.
+
+- `OrganizationalUnit` — `ParentId` is create-only per the registry schema, so an update only
+  renames the OU in place; the physical id survives, keeping every `Ref` to it valid.
+- `Account` — `ParentIds` accepts a single entry and the account is moved there after creation
+  (accounts are always created under the root). Because Floci resolves a 12-digit access key id
+  straight to an account, `Ref`/`Fn::GetAtt AccountId` yields an id you can immediately use as a
+  caller identity against other services.
+- `Policy` — `TargetIds` is reconciled on update: targets the template adds are attached and ones
+  it drops are detached. Deleting the resource detaches its remaining targets first, since
+  `DeletePolicy` refuses while a policy is still attached.
+- `ResourcePolicy` — backed by `PutResourcePolicy`, which is already create-or-update.
+
+Deleting the stack removes the resources in dependency order and finally the organization itself.
+Deleting a resource that is already gone is tolerated.
 
 ## Auto Scaling Launch Template Resolution
 
--- a/docs/services/cloudformation.md
+++ b/docs/services/cloudformation.md
@@ -127,7 +127,9 @@ data and attachment ownership.
 
 `AWS::Organizations::Organization` uses the stack's calling account as the management account, so
 the rest of the stack's Organizations resources are provisioned into that account's organization.
-`Fn::GetAtt Org.RootId` is the usual way to root the OU tree in the same template.
+`Fn::GetAtt Org.RootId` is the usual way to root the OU tree in the same template. `Organization` is
+the one type in this section where bare `Ref` does not return the resource's own id: per AWS, `Ref`
+returns the management account id, while `Fn::GetAtt Org.Id` returns the organization id.
 
 - `OrganizationalUnit` — `ParentId` is create-only per the registry schema, so an update only
   renames the OU in place; the physical id survives, keeping every `Ref` to it valid.
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/OrganizationsCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/OrganizationsCfnProvisioner.java
@@ -162,13 +162,18 @@ public class OrganizationsCfnProvisioner implements CfnResourceProvisioner {
             account = organizationsService.describeAccount(ctx.accountId(), status.getAccountId());
         }
 
+        // Claim the physical id before the move. CreateAccount has already made this a real member
+        // of the organization, so if moveAccount throws — a broken Ref or a stale literal in
+        // ParentIds raises DestinationParentNotFoundException — CloudFormation's rollback still has
+        // an id to call delete() with. Setting it afterwards would strand the account unmanaged.
+        r.setPhysicalId(account.getId());
+
         if (desiredParent != null && !desiredParent.equals(account.getParentId())) {
             organizationsService.moveAccount(
                     ctx.accountId(), account.getId(), account.getParentId(), desiredParent);
             account = organizationsService.describeAccount(ctx.accountId(), account.getId());
         }
 
-        r.setPhysicalId(account.getId());
         r.getAttributes().put("AccountId", account.getId());
         r.getAttributes().put("Arn", account.getArn());
         r.getAttributes().put("Status", account.getStatus());
@@ -196,9 +201,12 @@ public class OrganizationsCfnProvisioner implements CfnResourceProvisioner {
         } else {
             policy = organizationsService.createPolicy(ctx.accountId(), content, description, name, type, tags);
         }
+        // Same ordering as provisionAccount: the policy already exists once createPolicy returns,
+        // so claim the id before reconcileTargets, whose attach/detach calls can throw.
+        r.setPhysicalId(policy.getId());
+
         reconcileTargets(ctx.accountId(), policy, desiredTargets);
 
-        r.setPhysicalId(policy.getId());
         r.getAttributes().put("Id", policy.getId());
         r.getAttributes().put("Arn", policy.getArn());
         r.getAttributes().put("AwsManaged", String.valueOf(policy.isAwsManaged()));
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/OrganizationsCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/OrganizationsCfnProvisioner.java
@@ -159,6 +159,14 @@ public class OrganizationsCfnProvisioner implements CfnResourceProvisioner {
                 throw new AwsException("ConstraintViolationException",
                         "CreateAccount for " + accountName + " failed: " + status.getFailureReason(), 400);
             }
+            // The account exists from here on, so hand rollback the ownership marker before anything
+            // else can throw: CloudFormationService only deletes a CREATE_FAILED resource that is
+            // both marked owned and carries a physical id.
+            // The account exists from here on, so hand rollback the ownership marker before anything
+            // else can throw: CloudFormationService only deletes a CREATE_FAILED resource that is
+            // both marked owned and carries a physical id.
+            r.getAttributes().put(CfnRollback.ROLLBACK_OWNED_ATTR, "true");
+            r.setPhysicalId(status.getAccountId());
             account = organizationsService.describeAccount(ctx.accountId(), status.getAccountId());
         }
 
@@ -200,6 +208,9 @@ public class OrganizationsCfnProvisioner implements CfnResourceProvisioner {
             replaceTags(ctx.accountId(), policy.getId(), tags);
         } else {
             policy = organizationsService.createPolicy(ctx.accountId(), content, description, name, type, tags);
+            // As above: the policy is real once createPolicy returns, so mark it stack-owned before
+            // reconcileTargets gets a chance to throw.
+            r.getAttributes().put(CfnRollback.ROLLBACK_OWNED_ATTR, "true");
         }
         // Same ordering as provisionAccount: the policy already exists once createPolicy returns,
         // so claim the id before reconcileTargets, whose attach/detach calls can throw.
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/OrganizationsCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/OrganizationsCfnProvisioner.java
@@ -108,7 +108,9 @@ public class OrganizationsCfnProvisioner implements CfnResourceProvisioner {
             organization = organizationsService.createOrganization(ctx.accountId(), featureSet);
         }
 
-        r.setPhysicalId(organization.getId());
+        // AWS documents Organization as a special case: Ref returns the management account id,
+        // not the organization id. Fn::GetAtt Org.Id below still returns the organization id.
+        r.setPhysicalId(organization.getMasterAccountId());
         r.getAttributes().put("Id", organization.getId());
         r.getAttributes().put("Arn", organization.getArn());
         r.getAttributes().put("ManagementAccountArn", organization.getMasterAccountArn());
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/OrganizationsCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/OrganizationsCfnProvisioner.java
@@ -1,0 +1,297 @@
+package io.github.hectorvent.floci.services.cloudformation.provisioners;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.services.cloudformation.model.StackResource;
+import io.github.hectorvent.floci.services.organizations.OrganizationsService;
+import io.github.hectorvent.floci.services.organizations.model.CreateAccountStatus;
+import io.github.hectorvent.floci.services.organizations.model.Organization;
+import io.github.hectorvent.floci.services.organizations.model.OrganizationAccount;
+import io.github.hectorvent.floci.services.organizations.model.OrganizationPolicy;
+import io.github.hectorvent.floci.services.organizations.model.OrganizationalUnit;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import org.jboss.logging.Logger;
+
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * CloudFormation provisioning for AWS Organizations: {@code AWS::Organizations::Organization},
+ * {@code OrganizationalUnit}, {@code Account}, {@code Policy} and {@code ResourcePolicy}.
+ *
+ * <p>Attribute names come from the registry schemas in
+ * {@code aws-cloudformation-resource-providers-organizations} ({@code readOnlyProperties}), so
+ * {@code Fn::GetAtt} resolves the same keys AWS publishes.
+ */
+@ApplicationScoped
+public class OrganizationsCfnProvisioner implements CfnResourceProvisioner {
+
+    private static final Logger LOG = Logger.getLogger(OrganizationsCfnProvisioner.class);
+
+    private static final String ORGANIZATION = "AWS::Organizations::Organization";
+    private static final String ORGANIZATIONAL_UNIT = "AWS::Organizations::OrganizationalUnit";
+    private static final String ACCOUNT = "AWS::Organizations::Account";
+    private static final String POLICY = "AWS::Organizations::Policy";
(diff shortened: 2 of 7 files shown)


pgermosen:
Thanks for the thorough writeup and the real CLI-driven verification (23 stack outputs, the OU physical-id-preservation check across UpdateStack) — that's exactly the kind of end-to-end proof that matters here, and the template/checklist are filled in properly too.

Most of this checks out well. I independently reverted the OU rename-in-place logic (back to always-create) and confirmed `organizationalUnitUpdateRenamesInsteadOfCreating` fails, so that behavior is genuinely tested. Full `organizations` + `cloudformation` packages pass clean, including on a fresh merge against current `main` (no conflicts), and the `Fn::GetAtt` keys being sourced from the registry schemas rather than invented is the right approach.

Two things I'd want addressed before merge, both confirmed by reading the code (not just taking the bot findings on faith) — the first was already raised by greptile in an earlier round and is still open on this commit:

## 1. `provisionAccount`/`provisionPolicy` can leak an unmanaged resource on partial failure

Both methods do the risky, exception-prone follow-up work (`moveAccount`, `reconcileTargets` → `attachPolicy`/`detachPolicy`) *before* calling `r.setPhysicalId(...)`:

```java
// provisionAccount, lines 151-171
OrganizationAccount account;
if (r.getPhysicalId() != null) { ... } else {
    CreateAccountStatus status = organizationsService.createAccount(...); // <- real side effect
    ...
    account = organizationsService.describeAccount(...);
}
if (desiredParent != null && !desiredParent.equals(account.getParentId())) {
    organizationsService.moveAccount(...); // <- can throw
}
r.setPhysicalId(account.getId()); // <- never reached if moveAccount throws
```

Concretely: `moveAccount` throws `DestinationParentNotFoundException` (`OrganizationsService.java:504-507`) if `ParentIds` points at an OU id that doesn't exist yet — a broken `Ref`, wrong literal id, or an OU removed out-of-band. `createAccount` has already succeeded by that point, so the account is a real member of the organization, but the exception propagates before `r.setPhysicalId()` ever runs. CloudFormation's rollback then has no physical ID to call `delete()` with, so the stack reports `ROLLBACK_COMPLETE` while the account sits there unmanaged. `provisionPolicy` has the identical shape: `createPolicy` (line 197) persists the policy before `reconcileTargets` (line 199, can throw from `attachPolicy`/`detachPolicy`) runs, and `r.setPhysicalId()` doesn't happen until line 201.

The fix is probably as simple as moving `r.setPhysicalId(...)` (and maybe the attribute puts) to right after the create call succeeds, before the follow-up step — that way, if `moveAccount`/`reconcileTargets` throws, CloudFormation's own rollback path already has an ID to find and clean up the resource with.

## 2. `AWS::Organizations::ResourcePolicy` tags never get removed on update

Every other resource type in this file (OU, Account, Policy) calls `replaceTags(...)` after its create-or-update call specifically so a tag key dropped from the template gets untagged, not just left stale. `provisionResourcePolicy` (line 231-234) calls `putResourcePolicy` directly and skips that step entirely — so removing or replacing a tag on a resource policy via `UpdateStack` silently doesn't converge; `ListTagsForResource` keeps reporting the old value. This was flagged by Copilot too; adding the same `replaceTags(...)` call here (or making `putResourcePolicy` itself reconcile) should close it, plus a test exercising a tag update the way the other three types already have one.

## Not a code issue, just flagging it

`Build and Test` failed on CI, but it's a 20-minute timeout kill ("The job has exceeded the maximum execution time of 20m0s"), not a real test failure — same class of issue another contributor hit today on a smaller PR. My own local run of the touched packages (`organizations` + `cloudformation`) is clean, so this is likely just this PR's large integration-test footprint pushing the full suite over budget rather than anything wrong with the code. Might be worth a re-run once the above is addressed, since a new commit will trigger a fresh run anyway.

Happy to take another look once the physical-id ordering and the ResourcePolicy tag reconciliation are sorted — the rest of this is in good shape.

hectorvent:
@willstjohnbacon please rebase main, CI timeout was fixed.

willstjohnbacon:
Both addressed, and rebased onto current `main` — thanks for the CI-timeout diagnosis, #2506's sharding has it running clean now.

Finding 1 turned out to be deeper than the ordering fix alone, and I'd rather lay that out than quietly ship it.

## 1. Physical id ordering — and the part that ordering alone didn't fix

I moved `r.setPhysicalId(...)` to immediately after the create call in both `provisionAccount` and `provisionPolicy`, as suggested. Greptile then pushed back saying the reordering was cosmetic because the exception stops the object reaching the stack. That specific mechanism is **not** what happens — `CloudFormationResourceProvisioner.provision` catches at line 453 rather than rethrowing, stamps `CREATE_FAILED` on the same `StackResource`, and returns it, so `CloudFormationService` does install it and the id is recorded.

But running the scenario showed the bot's *conclusion* was right even though its reasoning wasn't. With only the ordering fix:

```
Stranded | CREATE_FAILED | physicalId=313552552826   <- id recorded, ordering fix working
Org      | DELETE_FAILED | o-6178xy3p0g
FINAL: ROLLBACK_FAILED
ListAccounts -> 313552552826 still a member
```

No `Stranded | DELETE_*` event at all. `CloudFormationService.rollbackCreatedResources` (lines 1256-1260) only deletes a `CREATE_FAILED` resource when it has **both** a physical id and `CfnRollback.ROLLBACK_OWNED_ATTR`. I was supplying the first and not the second, so rollback walked past the account, then couldn't delete the organization that still contained it — hence `ROLLBACK_FAILED` rather than a clean unwind.

So the provisioner now sets the ownership marker the way `IamRoleCfnProvisioner` does, right after the create returns and before anything else can throw. Same run afterwards:

```
Stranded | DELETE_IN_PROGRESS -> DELETE_COMPLETE
Org      | DELETE_COMPLETE
FINAL: ROLLBACK_COMPLETE
DescribeOrganization -> AWSOrganizationsNotInUseException
```

`OrganizationsCfnRollbackIntegrationTest` covers this end to end: it drives a real stack whose `ParentIds` names a non-existent OU, and asserts the account id reaches the stack events, that rollback actually emits `DELETE_COMPLETE` for it, and that neither the account nor the organization survives. Removing just the marker turns it into `expected: <ROLLBACK_COMPLETE> but was: <ROLLBACK_FAILED>`.

Worth saying plainly: your original finding was correct and my first fix was necessary but not sufficient. Had I stopped at the reordering, the failure mode would have looked fixed at the unit level while still stranding the account in a real stack.

## 2. ResourcePolicy tags

Your first suggestion — calling `replaceTags(...)` here like the other three types — would have thrown. `resolveTagOwner` (`OrganizationsService.java:1381`) resolves account, root, OU and policy ids only and raises `TargetNotFoundException` otherwise: a resource policy isn't a `TagResource` target, so `rp-` ids never resolve.

So I took your second option and made `putResourcePolicy` reconcile — it clears before applying, so Put replaces rather than merges. That matches `resolveTags`, which returns `null` for an absent `Tags` property and a map when present, giving "Tags omitted = leave alone, Tags present = make them match" — the same contract the other three types honour.

One call for you: that changes `PutResourcePolicy`'s tag semantics for direct API callers too, not just the CloudFormation path. Replacement is what PUT implies and it's the only way the CFN type can converge, but if you'd rather keep merge semantics on the service API and scope reconciliation to the provisioner, I'll add a dedicated service call instead.

## Verifying the tests bite

Reverting each fix individually produces exactly the failure it should:

| Test | With the fix reverted |
|---|---|
| `anAccountThatCannotBeMovedIsDeletedByRollbackRatherThanStranded` | `expected: <ROLLBACK_COMPLETE> but was: <ROLLBACK_FAILED>` |
| `anAccountThatFailsToMoveStillCarriesItsPhysicalIdForRollback` | `expected: <123456789012> but was: <null>` |
| `aPolicyThatFailsToAttachStillCarriesItsPhysicalIdForRollback` | `expected: <p-abcd1234> but was: <null>` |
| `putResourcePolicyReplacesTagsRatherThanMergingThem` | `expected: <{env=prod}> but was: <{owner=platform, env=prod}>` |

One honest caveat on coverage: of the two ResourcePolicy tests, only the service-level one is real regression coverage. `resourcePolicyUpdatePassesTheFullTagSetSoDroppedKeysAreRemoved` passes with or without the fix, because the provisioner was already handing over the complete tag set — the bug was downstream of it. I kept it as a contract test so a future switch to delta-style updates gets caught, but it isn't what catches this one.

## Tests

- `./mvnw test -Dtest='Organizations*'` — 118, up from 113.
- `make docs-check` — clean, no working-tree changes.
- `./mvnw test` — 11,991 run, 0 failures. The one error was `RuntimeApiServerTest` picking up a port my editor was already serving on (`receiving [Zed Editor Stable Instance Running]`); it passes in isolation and is nowhere near this change. CI is the better authority there — the previous push came back green across all four shards and every compat suite.

hectorvent:
Thank you for the way this round was done: the ordering fix, then following the failure past it to the rollback ownership marker the bot's reasoning missed, and documenting which test catches which fix, honest caveat included. Both earlier asks check out as resolved at this head, and the rollback integration test is exactly the right proof.

**(blocking)** One item from my earlier pass that had not reached you yet: the registry schemas declare more readOnly attributes than the provisioner maps, namely OrganizationalUnit `Path` plus Account `State` and `Paths` (see the [OU](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-organizations-organizationalunit.html#aws-resource-organizations-organizationalunit-return-values) and [Account](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-organizations-account.html#aws-resource-organizations-account-return-values) return values). An unmapped key silently resolves to the literal `LogicalId.Attr`, so please map them and extend the stack output assertions to the new keys. `Path` and `State` are plain strings; if the attribute map cannot carry the `Paths` list, deferring it with a note is fine.

That is the last thing between this and merge from my side.

willstjohnbacon:
Mapped, with assertions on the new keys. Pushed as `cde20a95`.

## Where the attribute names came from

Worth flagging, because it explains why I'd missed them: the provider repo `aws-cloudformation-resource-providers-organizations` is **behind the published registry** on exactly these three. On `master` it declares:

```
organizationalunit  readOnly: [Arn, Id]
account             readOnly: [AccountId, Status, JoinedTimestamp, JoinedMethod, Arn]
```

whereas `schema.cloudformation.us-east-1.amazonaws.com/aws-organizations-*.json` declares:

```
organizationalunit  readOnly: [Arn, Id, Path]
account             readOnly: [AccountId, Status, State, JoinedTimestamp, JoinedMethod, Arn, Paths]
```

The provider repo is what I sourced the original table from, so the gap was structural rather than an oversight of one or two keys. I re-checked all five types against the published schemas this time — `Organization`, `Policy` and `ResourcePolicy` were already complete, so `Path`, `State` and `Paths` are the whole remainder.

## `Path` / `Paths`

Both are the organization path, and the shape is not just the parent chain — the resource itself is the last element, trailing slash included. `ListAccounts` in the API reference returns `"o-exampleorgid/r-examplerootid111/111111111111/"`, and the `Path` shape's pattern ends `…(/\d{12})*\/`, which is why an account id can appear in it at all.

Computing that needs a walk from the resource up to the root, and `OrganizationsService` already does exactly that walk for effective-policy merging (`ancestryOf`). So it now exposes `organizationPath(caller, resourceId)` and the provisioner calls it, rather than the provisioner rebuilding the hierarchy from `ListParents` — which also keeps it single-dependency.

One ordering trap worth naming: accounts are always created under the root and moved afterwards, so reading the path before `moveAccount` publishes a path the account no longer lives at. The lookup sits after the move, and `accountPathsNameTheParentTheAccountWasMovedInto` pins that with an `InOrder` verify rather than just asserting the final value.

## `Paths` did fit, so I didn't defer it

Thanks for the out — it turned out not to be needed. `StackResource.attributes` is `Map<String, String>`, but `CloudFormationTemplateEngine.resolveList` falls back to splitting a comma-delimited scalar, which is precisely how `Fn::Select` reads a list-typed attribute. An account has one parent and therefore one path, so the list is always a single element and round-trips as itself. The template asserts both forms:

```json
"AccountPaths": { "Value": { "Fn::GetAtt": ["Dev", "Paths"] } },
"AccountPath0": { "Value": { "Fn::Select": [0, { "Fn::GetAtt": ["Dev", "Paths"] }] } }
```

Being straight about the limit rather than overselling it: `Fn::Join` given a `Fn::GetAtt` as its list argument still yields `""`, because `resolveJoin` indexes the node directly instead of going through `resolveList`. That is pre-existing and hits every list-typed attribute in the engine, not just this one — I left it alone as out of scope here, but say the word if you'd rather it were fixed in this PR.

## `State`

`State` is what AWS is retiring `Status` in favour of on 2026-09-09 — the API reference carries that notice on `DescribeAccount`, `ListAccounts` and `ListAccountsForParent`. `AccountState` adds `PENDING_ACTIVATION` and `CLOSED` to the three `AccountStatus` values. Floci only ever puts an account in `ACTIVE` or `PENDING_CLOSURE`, both of which are valid `AccountState` values too, so the two agree and `State` mirrors `Status`; the two extra phases are ones the emulator doesn't model. That's a comment in the provisioner and a line in `docs/services/cloudformation.md` so it isn't mistaken for a copy-paste.

Related but deliberately **not** in this PR: the Organizations API surface itself still reports only `Status`, and has no `Path`/`Paths` on `DescribeAccount`/`ListAccounts`/`DescribeOrganizationalUnit`. Given the September retirement that's probably worth its own change — happy to open an issue, or fold it in here if you'd rather it land together.

## One thing I had to fix to avoid re-introducing your earlier finding

Adding the `Path` lookup to `provisionOrganizationalUnit` opens the same window that stranded accounts: a call that can throw *after* `createOrganizationalUnit` has already made the OU a real member of the organization. Previously nothing could throw in that gap, so the OU never needed the marker. It does now, so it gets `ROLLBACK_OWNED_ATTR` as soon as the create returns, matching the account and policy paths.

I also removed a duplicated comment block and a redundant second `setPhysicalId(account.getId())` that `fa08d597` left in `provisionAccount` — no behaviour change, the id was already set from `status.getAccountId()` on the create path and unchanged on the update path.

## Verifying the tests bite

Reverted each change individually. The integration failures are the exact mode you described — the unmapped key resolving to the literal `LogicalId.Attr`:

| Reverted | Failure |
|---|---|
| `Path` mapping | `organizationalUnitAttributesResolveAndRefIsTheOuId` — `expected: <o-…/r-…/ou-…/> but was: <Workloads.Path>`, plus `organizationalUnitSetsPhysicalIdAndGetAttAttributes` → `null` |
| `State` mapping | `accountAttributesResolve` — `expected: <ACTIVE> but was: <Dev.State>` |
| `Paths` mapping | `accountAttributesResolve` — `expected: <o-…/r-…/ou-…/708217301815/> but was: <Dev.Paths>` |
| `Paths` read *before* the move | `accountAttributesResolve` — `expected: <…/ou-016y-dyt61h1m/348030273148/> but was: <…/r-016y/348030273148/>`, and the `InOrder` verify |
| OU rollback marker | `anOuWhosePathLookupFailsStaysOwnedByTheStackForRollback` — `expected: <true> but was: <null>` |

## Tests

- `./mvnw test -Dtest='Organizations*'` — 121, up from 118.
- Full `cloudformation` + `organizations` packages — 802 run, 2 failures, both pre-existing on `fa08d597` with the branch stashed and both Docker-dependent locally: `CloudFormationIntegrationTest.createStack_ecrAutoName_isLowercase` (repository lands `CREATE_FAILED`) and `SamTransformIntegrationTest.samFunction_withAutoPublishAlias_createsVersionAndAlias` (`X-Amz-Function-Error: Unhandled`). Neither touches Organizations; CI is the better authority.
- `make docs-check` — clean.

pgermosen (commented):
Thanks for this — the accuracy here is genuinely high. I checked all five resource types against AWS's CloudFormation resource reference (not just the registry schema's `readOnlyProperties`, but the published `Ref`/`Fn::GetAtt` docs), and four of them are exactly right, including the subtler details: `OrganizationalUnit`'s `Ref` → `Id` with all three `Fn::GetAtt` keys (`Arn`, `Id`, `Path`), `Account`'s full six-attribute set (`AccountId`, `Arn`, `JoinedMethod`, `JoinedTimestamp`, `Paths`, `State`, `Status`), `Policy`'s `AwsManaged`, and `ResourcePolicy`. The rollback-ownership-before-risky-call ordering, the `State`-vs-`Status` deprecation note, and the `Paths`-as-scalar `Fn::Select` round-trip comment all show real attention to the failure modes, not just the happy path.

## (should-fix) `Ref` on `AWS::Organizations::Organization` returns the wrong value

AWS's CloudFormation reference is explicit that this resource type is a special case: *"When you pass the logical ID of this resource to the intrinsic `Ref` function, `Ref` returns the `AccountId`"* — the management account's 12-digit id, e.g. `123456789012`. Every other resource type in this PR has `Ref` return its own natural id; this one doesn't.

`OrganizationsCfnProvisioner.provisionOrganization` sets:

```java
r.setPhysicalId(organization.getId());   // "o-xxxxxxxxxx"
```

and Floci's `CloudFormationTemplateEngine.resolveRef` resolves `Ref` directly from `physicalId` (`CloudFormationTemplateEngine.java:168-187`) — so `!Ref MyOrg` in Floci returns the org id where real AWS returns the management account id. `Fn::GetAtt Org.Id` is unaffected and correctly returns the org id already (that's a separate, correct attribute) — the bug is specifically that `physicalId`, which backs bare `Ref`, is set to the same value instead of `organization.getMasterAccountId()`.

This is why it wasn't caught: the integration test only exercises `Fn::GetAtt` on `Org` (`OrgId`, `OrgArn`, `MgmtAccountId`, etc. in the stack outputs) and never a bare `!Ref Org`. The unit test goes further and pins the current behavior directly: `OrganizationsCfnProvisionerTest.organizationSetsPhysicalIdAndEveryGetAttAttribute` asserts `assertEquals(ORG_ID, r.getPhysicalId())`, which would need to become `assertEquals(ACCOUNT, r.getPhysicalId())` once fixed — that assertion is the same evidence, not a separate bug.

Suggested fix is one line: `r.setPhysicalId(organization.getMasterAccountId())`, leaving `r.getAttributes().put("Id", organization.getId())` as-is.

## Other notes

- **Missing label**: `organizations` — this PR adds real behavior to `OrganizationsService.java` (`organizationPath`, `findManagementAccountForResource`, resource-policy tag semantics), not just `cloudformation`.
- **Minor, not blocking**: AWS documents `Account.ParentIds` as accepting at most one entry, but `provisionAccount` takes `parentIds.get(0)` without rejecting a longer list — permissive rather than incorrect, and low-stakes for an emulator, but worth a one-line validation if you're already in the area.

Merge is clean and CI is fully green, so no rebase or build issue here — this is purely the `Organization` `Ref` fix.

Recommendation: **comment**.

willstjohnbacon:
Fixed, pushed as `31b16c45`.

`provisionOrganization` was setting `physicalId` — which `resolveRef` reads directly for bare `Ref` — to `organization.getId()`. Changed it to `organization.getMasterAccountId()`, matching AWS's documented special case. `Fn::GetAtt Org.Id` was untouched since it was already correct.

Added an `OrgRef` output to the integration template asserting `Ref: Org` resolves to the management account id — exactly the case you noted the existing test never exercised. Also updated the three unit tests in `OrganizationsCfnProvisionerTest` that had pinned the old value, and added a line to `docs/services/cloudformation.md` noting the `Ref`/`Fn::GetAtt` split for this type.

Checked the one thing this could plausibly break: `delete()` recovers its caller from `physicalId` via `findManagementAccountForResource`, which now gets an account id instead of an org id for this resource type. That still resolves correctly — the management account is stored in the accounts table too, so it's found there instead of via the organization-id branch. Confirmed with the rollback/teardown integration tests, which still delete the organization cleanly.

Left the two non-blocking notes alone: `organizations` is already on the PR's labels, and the `ParentIds` length check you flagged as low-stakes I've left for a follow-up if you'd still like it.

## Tests

- `./mvnw test -Dtest='OrganizationsCfnProvisionerTest,OrganizationsCfnIntegrationTest,OrganizationsCfnRollbackIntegrationTest'` — clean.
- Full `organizations` + `cloudformation` packages — clean.
- `make docs-check` — clean.

pgermosen (approved):
Confirmed fixed, exactly right: `physicalId` now comes from `organization.getMasterAccountId()` rather than `organization.getId()`, matching AWS's documented special case for this resource type, with `Fn::GetAtt Org.Id` correctly left untouched. You also checked the one place this could plausibly regress — `delete()`'s `findManagementAccountForResource` lookup, which now receives an account id instead of an org id for this type — and confirmed it still resolves via the accounts-table branch, backed by the rollback/teardown integration tests actually deleting the organization cleanly. The new `OrgRef` output closes the exact gap that let this slip through originally: nothing previously exercised a bare `Ref` on `Organization`.

Merge is clean, CI is fully green.

Recommendation: **approve**.

PR #2497, opened 2026-08-23
feat(core): label emulated resource containers with io.floci.* identity

### LINKED ISSUE #1818: [FEATURE] Label spawned service containers with the emulated resource's identity (io.floci.*)
### Environment

Floci 1.5.31, image `floci/floci@sha256:690f3077367a866a28509a35bdf1258246bf20b19ec06fcb59093726ee60f7f4`, Docker 29.2.1 / Ubuntu 24.04, RDS engine `postgres` (backing image `postgres:16-alpine`).

### Need

A static way to map a Floci-spawned container back to the emulated resource it backs.

### Actual behavior

RDS backing containers are named `floci-rds-<random-hex>` and carry no identifying metadata — no Docker label, no env var, nothing referencing the `DBInstanceIdentifier`. Container creation order doesn't correlate with API-call order either (we observed the second-created instance's container appear ~1.2s before the first's), so ordering inference is unsound. The only reliable mapping we found was empirical: seed distinguishable data through the API, then `docker exec` each candidate container to see which one holds it.

### AWS-analogy framing

On real AWS every resource is addressable by its identifier/ARN; the container is Floci's "physical host" for the resource, and there's currently no equivalent of resource identity at that layer.

### Proposal

Stamp standard labels at container-create time, e.g.:

```
io.floci.service=rds
io.floci.resource-id=<DBInstanceIdentifier>
io.floci.account=<account-id>
io.floci.region=<region>
```

Then `docker ps --filter label=io.floci.resource-id=orders-db-primary` just works. This unblocks fault-injection/chaos tooling, per-resource cleanup, and test debugging generically, and mirrors what docker-compose / testcontainers already do with `com.docker.compose.*` / `org.testcontainers.*` labels. Low-risk: labels are inert metadata.

**We're happy to contribute this as a PR** if the maintainers agree on the label namespace — it looks like a small, contained change at the container-spawn site plus an integration-test assertion.

omatheusmesmo:
## Summary

Stamps every container Floci spawns to back an emulated AWS resource with
`io.floci.*` identity labels, so a container can be traced back to the
resource it emulates without empirical `docker exec` probing.

Labels applied (additive to the existing `floci`/`floci_emulator`/`floci_namespace` labels):

- `io.floci=aws` (bare, per maintainer's comment — multi-cloud discovery)
- `io.floci.service=<service>`
- `io.floci.resource-id=<resource-id>`
- `io.floci.account=<account-id>`
- `io.floci.region=<region>`

New `ContainerStorageHelper.resourceIdentityLabels(...)` is the single helper
every manager calls; it flows through the existing `ContainerLifecycleManager`
choke point, so no changes to label-merging logic were needed. Blank/null
values (e.g. ECR's shared registry has no per-resource id) are simply omitted.

Wired into all 18 container-backed services: RDS, DocDB, ElastiCache
(Redis/Valkey + Memcached), MemoryDB, Neptune, MSK, OpenSearch, ECS, EKS,
AmazonMQ, MWAA, Kinesis Data Analytics (Flink), Batch, CodeBuild, Lambda,
EC2, and ECR.

Closes #1818

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Not wire-protocol-affecting — this only adds Docker-side metadata to
containers the emulator itself creates; no AWS API request/response shapes
change.

## Checklist

- [x] `./mvnw test` passes locally (targeted run of all 178 tests across every touched file is green; a broader full-suite run surfaced a pre-existing, unrelated Lambda-invocation timeout flake, reproduced identically on `main` via a baseline worktree)
- [x] New or updated integration test added (one per service, TDD — written first, red, then made green)

### DIFF
--- a/docs/configuration/docker.md
+++ b/docs/configuration/docker.md
@@ -152,7 +152,7 @@ A container backing an emulated AWS resource also carries labels tying it back t
 | `io.floci.account` | the resolved account id | The AWS account the resource belongs to |
 | `io.floci.region` | the resolved region | The AWS region the resource belongs to |
 
-This makes `docker ps --filter label=io.floci.resource-id=orders-db-primary` resolve a specific emulated resource to its backing container directly, without inferring it from names or creation order. Applied to RDS, DocDB, ElastiCache (Redis/Valkey and Memcached), MemoryDB, Neptune, MSK, OpenSearch, ECS, EKS, AmazonMQ, MWAA, Kinesis Data Analytics (Flink), Batch, CodeBuild, Lambda, and EC2. ECR's backing registry container carries `io.floci`/`io.floci.service` only, since it is a shared singleton with no single resource identifier.
+This makes `docker ps --filter label=io.floci.resource-id=orders-db-primary` resolve a specific emulated resource to its backing container directly, without inferring it from names or creation order. Applied to RDS, DocDB, ElastiCache (Redis/Valkey and Memcached), MemoryDB, Neptune, MSK, OpenSearch, ECS, EKS, AmazonMQ, MWAA, Kinesis Data Analytics (Flink), Batch, CodeBuild, Lambda, and EC2. ECR's backing registry container carries every label except `io.floci.resource-id`, since it is a shared singleton with no single resource identifier.
 
 ## Docker Network
 
--- a/docs/configuration/docker.md
+++ b/docs/configuration/docker.md
@@ -140,6 +140,20 @@ Extra labels are a list of key/value entries rather than a map so that label key
 !!! note
     Entries using one of the reserved keys (`floci`, `floci_emulator`, `floci_namespace`) are ignored with a warning — user configuration can never break Floci's own container discovery and volume pruning.
 
+### Resource Identity Labels
+
+A container backing an emulated AWS resource also carries labels tying it back to that resource, additive to the labels above:
+
+| Label | Value | Purpose |
+|---|---|---|
+| `io.floci` | `aws` | Cloud provider, for multi-cloud discovery when several Floci-like emulators share a host |
+| `io.floci.service` | e.g. `rds` | The AWS service the container backs |
+| `io.floci.resource-id` | e.g. `orders-db-primary` | The resource's identifier (DB instance identifier, cluster name, function name, ...) |
+| `io.floci.account` | the resolved account id | The AWS account the resource belongs to |
+| `io.floci.region` | the resolved region | The AWS region the resource belongs to |
+
+This makes `docker ps --filter label=io.floci.resource-id=orders-db-primary` resolve a specific emulated resource to its backing container directly, without inferring it from names or creation order. Applied to RDS, DocDB, ElastiCache (Redis/Valkey and Memcached), MemoryDB, Neptune, MSK, OpenSearch, ECS, EKS, AmazonMQ, MWAA, Kinesis Data Analytics (Flink), Batch, CodeBuild, Lambda, and EC2. ECR's backing registry container carries `io.floci`/`io.floci.service` only, since it is a shared singleton with no single resource identifier.
+
 ## Docker Network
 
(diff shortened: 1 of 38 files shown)


pgermosen:
@omatheusmesmo Reviewed this — the core design (`ContainerStorageHelper.resourceIdentityLabels`, blank-value omission, threading `RegionResolver` into managers that didn't already have it) is sound, and I spot-checked several call sites including the RDS one (the largest diff): it correctly refactors the pre-existing ARN-parsing logic into a shared `resolveRuntimeIdentity` helper while carefully preserving the exact null-vs-fallback semantics the existing log-stream routing (`attach` vs `attachForAccount`) depends on — good attention to not regressing existing behavior while adding the new one.

Three concrete issues before this can merge:

## 1. Real merge conflicts against current `main` (not mechanical)

`git merge --no-commit --no-ff origin/main` conflicts in 4 files, and these aren't simple textual collisions — `main` has since refactored around the exact lines this PR touches:

- `Ec2ContainerManager.java` / `Ec2ContainerManagerTest.java`: `main` extracted a `createAndStartContainer(...)` helper and restructured the container-spec-building flow since this branch was cut. Your `.withLabels(...)` call needs to be re-inserted into the new structure, not just picked from one side.
- `FlinkContainerManager.java` / `FlinkContainerManagerTest.java`: `main` added `.withHostDockerInternalOnLinux()` and `.withEmbeddedDns()` to the same builder chain your `.withLabels(...)` call was appended to — both sides' additions need to be kept together.

A rebase (not a merge) is probably the cleanest way to sort these out given how much `main` has moved.

## 2. Docs claim ECR gets fewer labels than the code actually attaches

`docs/configuration/docker.md` says: "ECR's backing registry container carries `io.floci`/`io.floci.service` only, since it is a shared singleton with no single resource identifier." But `EcrRegistryManager.java` calls:

```java
ContainerStorageHelper.resourceIdentityLabels("ecr", null, regionResolver.getAccountId(), regionResolver.getDefaultRegion())
```

`RegionResolver.getAccountId()`/`getDefaultRegion()` always fall back to the configured default account/region and are never blank — only the `resourceId` (`null`) is actually omitted by `putIfNotBlank`. So the ECR container does get `io.floci.account` and `io.floci.region`, just not `io.floci.resource-id`. Either update the doc line to say "all labels except `io.floci.resource-id`", or — if account/region genuinely shouldn't apply to a shared singleton — pass `null` for those too and adjust the code to match the doc's stated intent. Right now the two disagree.

## 3. Greptile's P2 (test diagnostics) is still open

`ElastiCacheContainerManagerTest.java:50` and `ElastiCacheMemcachedContainerManagerTest.java:36` both still have an empty `catch (IOException ignored) {}` in the new cleanup path greptile flagged — no reply or fix on that thread. Not blocking on its own, but worth picking up alongside the rebase since you'll be touching these files anyway.

Everything else checks out — the labeling logic itself is correct everywhere I checked, full suite passes on the branch, title/commits are clean. Once the conflicts are resolved and the docs/ECR mismatch is sorted, happy to take another look.

omatheusmesmo:
@pgermosen All 3 addressed: rebased onto current `main`, fixed the ECR docs (keeps every label except `resource-id`), and the ElastiCache `catch (IOException ignored)` blocks now log, matching the DocDB/Neptune/MemoryDB pattern. Ready for another look.

pgermosen (approved):
All three items from my last review are genuinely fixed — verified independently rather than taking the claim on trust:

1. Merge conflicts: rebased onto current `main`, confirmed a clean merge, and the merged tree still compiles and passes tests for every touched service package.
2. ECR docs mismatch: `docker.md` now correctly says "every label except `io.floci.resource-id`" — matches the actual code.
3. Greptile's P2: both ElastiCache test files now log the tolerated `IOException` instead of silently swallowing it.

Full suite passes across all 18 touched container-manager packages, on the branch and on a fresh merge against current `main`. All CI green.

Good to merge.

PR #2499, opened 2026-08-23
fix(ses): report the sending account in send-event payloads

okinaka:
## Summary

The SES send-event notification payload built `mail.sendingAccountId` and `mail.sourceArn` (the
sender identity ARN) from the fixed default account, so under Floci's per-request multi-account
support they always showed `000000000000` regardless of the caller.

`publishSendEvents` now resolves the account per request via `RegionResolver.getAccountId()` (the
same request-scoped mechanism the account-aware services use), falling back to the default account
outside a request context (e.g. unit tests). Because the whole send family (`SendEmail`,
`SendRawEmail`, templated, bulk — v1 and v2) shares `publishSendEvents`, this is a single fix point.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

AWS populates the event's `sendingAccountId`/`sourceArn` with the sending account. Floci resolves it
from the caller's credential (a 12-digit access key id is the account, matching the multi-account
convention), so a send authenticated as account `2109...` now reports that account instead of the
default. The delivery-event integration test asserts `mail.sendingAccountId` and `mail.sourceArn`
carry the resolved account. No wire-shape change; only the previously-hardcoded account value is now
per-request.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesService.java
@@ -3,6 +3,7 @@ package io.github.hectorvent.floci.services.ses;
 import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsArnUtils;
 import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.core.common.RegionResolver;
 import io.github.hectorvent.floci.core.storage.StorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageFactory;
 import io.github.hectorvent.floci.services.route53.Route53Service;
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Reviewed the diff in full and did revert-and-confirm-fail on the fix.

`publishSendEvents` previously hardcoded `sendingAccountId = defaultAccountId`; this resolves it via `regionResolver.getAccountId()`, matching the exact same pattern already used in Lambda, Firehose, Kinesis, CloudFormation, SWF, MWAA, and others. `RegionResolver.getAccountId()` already falls back to the default account on `ContextNotActiveException`, so there's no risk outside a request context — consistent with the two test-only constructors that pass `regionResolver = null`.

Verified:
- New `SesEventPublishingAccountV2IntegrationTest` authenticates as a non-default account and asserts `mail.sendingAccountId`/`mail.sourceArn` in the SNS-delivered event payload report that account, not the default — plus an added assertion on the existing default-account test.
- Reverted the one-line fix locally and reran the new test: fails exactly as expected (`expected: <210987654321> but was: <000000000000>`). Restored, reran green.
- Full SES suite (886 tests) passes on the branch; clean `--no-ff` merge with `main`, and all 907 SES tests pass on the merged tree.
- No AI co-authorship trailers, Conventional Commits title, no CODEOWNERS-sensitive paths touched.

Note: the "Build and Test" check shows failed, but it's a 20-minute CI timeout, not a test failure — the latest green run on `main` took 19m42s, right at that ceiling, so this looks like pre-existing flakiness rather than something introduced by this diff. Worth a rerun.

Approving.

PR #2501, opened 2026-08-24
fix(bedrock-agentcore): support memory tagging and dropped fields

kannandreams:
## Summary

Fixes three divergences from AWS in the Bedrock AgentCore Memory control plane (follow-up to #2316):

1. **Memory ARNs were not taggable.** The shared `/tags/{arn}` handler dispatched only `agent/` and `gateway/` ARN segments, so tagging a memory returned `ValidationException` even though AWS supports `TagResource` on memories (`CreateMemory` documents a `tags` map "to assign to an AgentCore Memory"). The handler now dispatches three ways, and the foreign-account/region 404 guard covers memories too.
2. **`CreateMemory` silently dropped fields.** `tags`, `encryptionKeyArn`, and `memoryExecutionRoleArn` were accepted with a 202 but never stored. All three now persist and survive restart.
3. **`UpdateMemory` ignored input.** Only `description` was applied; `eventExpiryDuration` (now validated 3–365, matching the CreateMemory schema — the UpdateMemory prose says 7–365, the schema says 3–365) and `memoryExecutionRoleArn` are now honored. `encryptionKeyArn` is deliberately *not* updatable, matching the UpdateMemory request shape.

Matching AWS, the `memory` response shape never embeds `tags` — they surface only via `ListTagsForResource` (pinned by a test).

Note for reviewers: the AWS TagResource page's supported-resources note doesn't list Memory, but the current `CreateMemory` schema explicitly documents memory tags; the schema was treated as authoritative.

Out of scope, possible follow-ups: `memoryStrategies` (+ `indexedKeys`/`namespaceKeys`/`streamDeliveryResources`), the memory data-plane ops (`CreateEvent`/`ListEvents`/…), and status transitions.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Wire shapes verified against the `bedrock-agentcore-control` API Reference (CreateMemory, UpdateMemory, Memory, MemorySummary, TagResource) on 2026-08-24. Compat tests extended using AWS SDK for Java 2.52.0 (`CreateMemoryRequest.tags/encryptionKeyArn/memoryExecutionRoleArn`, `UpdateMemoryRequest.eventExpiryDuration`).

Incorrect behavior before: tagging a memory ARN → 400; create-time `tags`/`encryptionKeyArn`/`memoryExecutionRoleArn` silently discarded; `eventExpiryDuration` updates silently ignored.

## Checklist

- [x] `./mvnw test` passes locally (all Bedrock AgentCore suites green; unrelated Lambda/WebSocket/ECR failures on this machine are due to Docker not being installed locally)
- [x] New or updated integration test added (new `BedrockAgentCoreMemoryServiceTest` unit suite; extended memory, tagging, and persistence integration tests; extended SDK compat tests)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/BedrockAgentCoreMemoryTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/BedrockAgentCoreMemoryTest.java
@@ -4,6 +4,7 @@ import org.junit.jupiter.api.*;
 import software.amazon.awssdk.services.bedrockagentcorecontrol.BedrockAgentCoreControlClient;
 import software.amazon.awssdk.services.bedrockagentcorecontrol.model.*;
 
+import java.util.Map;
 import java.util.UUID;
 
 import static org.assertj.core.api.Assertions.*;
@@ -40,12 +41,19 @@ class BedrockAgentCoreMemoryTest {
                 .name(memoryName)
                 .eventExpiryDuration(30)
                 .description("v1")
+                .encryptionKeyArn("arn:aws:kms:us-east-1:000000000000:key/mem-key")
+                .memoryExecutionRoleArn("arn:aws:iam::000000000000:role/mem-role")
+                .tags(Map.of("team", "core"))
                 .build());
         memoryId = response.memory().id();
         assertThat(memoryId).isNotBlank();
         assertThat(response.memory().name()).isEqualTo(memoryName);
         assertThat(response.memory().arn()).contains(":memory/");
         assertThat(response.memory().statusAsString()).isEqualTo("ACTIVE");
+        assertThat(response.memory().encryptionKeyArn())
+                .isEqualTo("arn:aws:kms:us-east-1:000000000000:key/mem-key");
+        assertThat(response.memory().memoryExecutionRoleArn())
+                .isEqualTo("arn:aws:iam::000000000000:role/mem-role");
     }
 
     @Test
@@ -53,6 +61,8 @@ class BedrockAgentCoreMemoryTest {
     void getMemory() {
         GetMemoryResponse response = client.getMemory(GetMemoryRequest.builder().memoryId(memoryId).build());
         assertThat(response.memory().name()).isEqualTo(memoryName);
+        assertThat(response.memory().encryptionKeyArn())
+                .isEqualTo("arn:aws:kms:us-east-1:000000000000:key/mem-key");
     }
 
     @Test
@@ -66,8 +76,12 @@ class BedrockAgentCoreMemoryTest {
     @Order(4)
     void updateMemory() {
         UpdateMemoryResponse response = client.updateMemory(UpdateMemoryRequest.builder()
-                .memoryId(memoryId).description("v2").build());
+                .memoryId(memoryId).description("v2").eventExpiryDuration(60).build());
         assertThat(response.memory().id()).isEqualTo(memoryId);
+        assertThat(response.memory().eventExpiryDuration()).isEqualTo(60);
+
+        GetMemoryResponse after = client.getMemory(GetMemoryRequest.builder().memoryId(memoryId).build());
+        assertThat(after.memory().eventExpiryDuration()).isEqualTo(60);
     }
 
     @Test
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/BedrockAgentCoreTagIdentityTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/BedrockAgentCoreTagIdentityTest.java
@@ -91,6 +91,41 @@ class BedrockAgentCoreTagIdentityTest {
     }
 
     @Test
+    @Order(4)
+    void memoryTagRoundTripViaSdk() {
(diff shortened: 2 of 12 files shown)


pgermosen:
Found a correctness bug in `update()`:

`BedrockAgentCoreMemoryService.update()` (line 93) sets `description` on the live `Memory` object before validating `eventExpiryDuration`. Since `InMemoryStorage.get()` returns the same object reference already in the map, a rejected update (400, out-of-range expiry) still leaves the description mutated — there's no rollback when `validateExpiry` throws partway through.

Repro: call `update(id, "newDesc", 999, null, region)` → throws 400 as expected, but a follow-up `get(id, region)` returns `description == "newDesc"` anyway. This is a new regression from this PR — on `main`, `update()` only ever touched `description` with nothing left that could throw afterward, so partial application wasn't possible before.

Fix is straightforward: validate `eventExpiryDuration` before mutating anything else (or validate all inputs up front, matching how `create()` validates before constructing the `Memory`).

Minor nit while I was in there: `updateMemory` and `updateMemoryRejectsOutOfRangeExpiry` in `BedrockAgentCoreMemoryIntegrationTest` both carry `@Order(3)` — not currently causing flakiness (the rejection case throws before any mutation), but worth deduplicating to avoid surprises later.

Everything else here looks solid — the expiry bounds and tagging ARN dispatch match the real `bedrock-agentcore-control` model, and the docs/tests are otherwise thorough.

kannandreams:
Good catch, thank you @pgermosen . 

- `update()` now validates eventExpiryDuration before calling get() or touching any setter, matching how create() validates before constructing the Memory.
-  `@Order(3) collision`: deduplicated — every test in BedrockAgentCoreMemoryIntegrationTest now has a unique order (1–7). The strengthened rejection test actually depends on running after updateMemory now, so the dedup ended up load-bearing rather than cosmetic

pgermosen (approved):
Re-checked at `f816a119`, merged onto current `main`, built, and driven against a running instance.

## The partial-mutation bug is fixed, and properly pinned

`update()` now validates before `get()` and before any setter, with a comment saying why. I reverted just that move and re-ran — two failures, both by name, and the unit one is my original repro rendered as an assertion message:

```
BedrockAgentCoreMemoryIntegrationTest.updateMemoryRejectsOutOfRangeExpiryWithoutPartialMutation:95
BedrockAgentCoreMemoryServiceTest.rejectedUpdateAppliesNothing:100
    expected: <desc> but was: <should-not-stick>
```

Live, on the wire:

```
PUT /memories/{id}/update  {"description":"should-not-stick","eventExpiryDuration":999,
                            "memoryExecutionRoleArn":"…/nope"}
→ 400 ValidationException
   description=orig   expiry=30   role=…/r1        ← nothing applied
```

I also checked there is nothing left that can throw after the first setter: what remains is three plain setters, `setUpdatedAt`, and the `put`. So the fix is complete rather than just moved one line earlier.

The `@Order` dedup is done (1–7), and you are right that it became load-bearing — the strengthened rejection test now reads back state written by `updateMemory`, so it has to run after it.

**One side effect worth recording, because it is a behaviour change and it is the right one.** Validating ahead of `get()` moves the error precedence: a bad expiry against a memory that does not exist now returns `400 ValidationException` rather than `404`. Confirmed live. That matches AWS, which validates request shape before resource lookup, so it is an improvement rather than a regression — just worth knowing it changed.

## The AWS claims hold at the model level

I checked these against `bedrockagentcorecontrol-2.52.0` — the same SDK the compat tests use — rather than against the doc pages, since the description flags a conflict between them.

| claim | model |
|---|---|
| `CreateMemory` accepts `tags`, `encryptionKeyArn`, `memoryExecutionRoleArn` | `CreateMemoryRequest` has all three |
| `encryptionKeyArn` is deliberately not updatable | `UpdateMemoryRequest` has `description`, `eventExpiryDuration`, `memoryExecutionRoleArn` — and no `encryptionKeyArn` |
| the `memory` shape never embeds `tags` | `Memory` has no `tags` member |
| memory ARNs are taggable despite the docs page | `TagResourceRequest` takes a bare `resourceArn` with no resource-type constraint |

So treating the schema as authoritative over the TagResource page's prose is the right call, and the "deliberately not updatable" decision is the model rather than a judgement call.

## Behaviour, end to end

```
create  {…, encryptionKeyArn, memoryExecutionRoleArn, tags:{team:core}}
   → both ARNs persisted and returned; tags correctly absent from the memory body

GET  /tags/{memoryArn}          → 200 {"tags":{"team":"core"}}     (was ValidationException)
POST /tags/{memoryArn} env=dev  → 204, then {"env":"dev","team":"core"}

valid update (description + expiry + role) → 202, all three applied
expiry boundaries:  2 → 400    3 → 202    365 → 202    366 → 400
```

That last row is the 3–365 schema bound the description argues for over the UpdateMemory prose's 7–365, and it is applied consistently on both create and update.

**78 AgentCore tests, 0 failures** on the merge with current `main`, and the compat suite is extended on both sides — memory fields and tag identity. Persistence of the three new fields is covered by `BedrockAgentCorePersistenceTest` rather than by my probe. Merge with `main` is clean.

No further findings.

PR #2507, opened 2026-08-24
ci: surface per-class timings from the compatibility test legs

hectorvent:
## Summary

Instruments the compatibility-test legs so their runtime is measurable. The slowest leg (`sdk-test-java`, ~10 minutes of the ~18-minute compat critical path) is currently a black box: the container runs `mvn test -q`, so the CI log has no per-class times, and the junit XMLs the suites already copy to `test-results/` are read once by the summary action and discarded.

- **Report slowest test classes**: a small parser (`.github/ci/compat-timing-report.py`) prints the slowest 25 classes plus the suite total to the job log and step summary, from the XMLs every leg already exports. It never fails the job; absent or malformed XMLs produce an empty report.
- **Upload `test-results/` as a per-leg artifact** (`compat-results-<leg>`), so timings are machine-readable and comparable across runs.
- The leg's sparse checkout gains `.github/ci` so the parser is present.

This is measurement only, deciding how to split or speed up the java leg should follow the data rather than precede it. Parser dry-run against the last local `sdk-test-java` results: 121 suites / 1,414 tests / 3.1 min summed, top classes Ec2Tests 32.8s, OpenSearchTest 15.6s, EcsTests 13.5s.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A: CI workflow instrumentation only; no emulated surface or test behavior is touched.

## Checklist

- [x] `./mvnw test` passes locally (no Java changed; parser dry-run verified against real junit XMLs, workflow YAML validated)
- [ ] New or updated integration test added (not applicable: CI-only change)

### DIFF
--- a/.github/ci/compat-timing-report.py
+++ b/.github/ci/compat-timing-report.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python3
+"""Print per-class timings from junit XMLs, slowest first.
+
+Usage: compat-timing-report.py <results-dir>
+Also appends the table to GITHUB_STEP_SUMMARY when set. Never fails the job:
+bad or absent XMLs produce an empty report, not an error.
+"""
+import glob
+import os
+import sys
+import xml.etree.ElementTree as ET
+
+def main():
+    d = sys.argv[1] if len(sys.argv) > 1 else 'test-results'
+    rows = []
+    for f in glob.glob(os.path.join(d, '*.xml')):
+        try:
+            root = ET.parse(f).getroot()
+        except ET.ParseError:
+            continue
+        suites = root.iter('testsuite') if root.tag != 'testsuite' else [root]
+        for ts in suites:
+            rows.append((float(ts.get('time') or 0), int(ts.get('tests') or 0),
+                         ts.get('name') or os.path.basename(f)))
+    rows.sort(reverse=True)
+    total_t = sum(r[0] for r in rows)
+    total_n = sum(r[1] for r in rows)
+    lines = [f'== {len(rows)} suites, {total_n} tests, {total_t / 60:.1f} min total test time',
+             '== slowest 25 classes:']
+    lines += [f'{t:8.1f}s {n:5d} tests  {name}' for t, n, name in rows[:25]]
+    report = '\n'.join(lines)
+    print(report)
+    summary = os.environ.get('GITHUB_STEP_SUMMARY')
+    if summary:
+        with open(summary, 'a') as fh:
+            fh.write('```\n' + report + '\n```\n')
+
+if __name__ == '__main__':
+    main()
--- a/.github/ci/compat-timing-report.py
+++ b/.github/ci/compat-timing-report.py
@@ -16,12 +16,17 @@ def main():
     for f in glob.glob(os.path.join(d, '*.xml')):
         try:
             root = ET.parse(f).getroot()
-        except ET.ParseError:
+        except (ET.ParseError, OSError) as exc:
+            print(f'warning: skipping unreadable JUnit XML {f}: {exc}', file=sys.stderr)
             continue
         suites = root.iter('testsuite') if root.tag != 'testsuite' else [root]
         for ts in suites:
-            rows.append((float(ts.get('time') or 0), int(ts.get('tests') or 0),
-                         ts.get('name') or os.path.basename(f)))
+            try:
+                rows.append((float(ts.get('time') or 0), int(ts.get('tests') or 0),
+                             ts.get('name') or os.path.basename(f)))
+            except ValueError as exc:
+                print(f'warning: skipping suite with non-numeric attributes in {f}: {exc}',
+                      file=sys.stderr)
     rows.sort(reverse=True)
     total_t = sum(r[0] for r in rows)
     total_n = sum(r[1] for r in rows)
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
@hectorvent — approving this. Well-scoped, correctly wired, and it pays for itself immediately. One small gap against a claim the PR makes, detailed at the end; it is a one-liner and I would rather it went in here than became a follow-up.

## Verified

- **The `steps.tests.outcome` guard is real.** `id: tests` exists on the "Run tests" step (line 199), so the condition means what it looks like. And `always()` means the timings show up on *failing* runs too, which is when you most want them — easy thing to get backwards.
- **The sparse-checkout addition works**: `.github/ci` is a directory, so cone mode picks up the parser.
- **The second commit is the easy-to-forget half.** Adding `.github/ci/compat-timing-report.py` to the `paths` filter means editing the parser actually retriggers the legs it instruments.
- `actions/upload-artifact@v7` matches the dominant convention in this repo (6 tag-pinned against 3 SHA-pinned), so no inconsistency introduced.
- Workflow YAML parses and both steps land in `native-compat-test`.
- Parser behaves on the happy path, on a missing directory, and on unparseable XML — I ran all three.

## The number this already surfaces

From your own dry-run: **121 suites, 1,414 tests, 3.1 min summed — inside a ~10 minute leg.** So roughly seven of those ten minutes are not test execution at all; they are image pull, container start and Maven startup. That is a different optimisation target from "split the java suite", and it is visible before anyone writes a line of splitting logic. Exactly the point of the PR, and worth stating out loud so the next person reads the data the same way.

## The one gap: "never fails the job" is stronger than the code

The docstring and the summary both promise that bad or absent XMLs produce an empty report rather than an error. That is true for *unparseable* XML, but `except ET.ParseError` does not cover parseable XML carrying an unexpected attribute value, nor an I/O error opening the file:

```
<testsuite name="Weird" tests="3" time="1,234.5"/>
  → ValueError: could not convert string to float: '1,234.5'    exit 1

a directory named  test-results/foo.xml
  → IsADirectoryError: [Errno 21] Is a directory                exit 1
```

Neither new step carries `continue-on-error`, so either one exits the leg non-zero and fails an otherwise-green compat run — the opposite of what measurement-only instrumentation should be able to do.

The directory case is contrived. The comma-decimal one is not quite: Surefire writes `time` using the default locale in some versions, the java leg runs Maven *inside a container*, and the compat pom pins no locale — so the value depends on the image rather than on the runner.

One line covers both:

```python
        try:
            root = ET.parse(f).getroot()
        except (ET.ParseError, OSError):
            continue
        ...
        for ts in suites:
            try:
                rows.append((float(ts.get('time') or 0), int(ts.get('tests') or 0),
                             ts.get('name') or os.path.basename(f)))
            except ValueError:
                continue
```

Or `continue-on-error: true` on the report step, which is blunter but guarantees the property regardless of what the parser grows into later. Given the whole point is that this never gets in the way, I would take both.

## Two notes, no action

- The summed suite time will not reconcile with elapsed leg time under any parallelism. You have been careful about this already — "3.1 min summed", and the header says "total test time" rather than elapsed — so this is just for whoever reads the artifact later.
- `root.iter('testsuite')` finds descendants at any depth, so a dialect emitting nested `<testsuite>` elements would be double-counted. Surefire does not, so it is theoretical.

hectorvent:
Thanks for the careful pass, and for running the failure modes rather than eyeballing them. You are right that the docstring promised more than the code delivered: I reproduced both cases (`float('1,234.5')` raises ValueError past the ParseError handler, and a directory named `*.xml` raises IsADirectoryError, which ParseError does not cover). Took your "both" option: the parse guard is now `except (ET.ParseError, OSError)`, the per-suite row is wrapped in a ValueError guard, and the report step carries `continue-on-error: true` so the property survives whatever the parser grows into. Also folded in Greptile's suggestion to warn on stderr with the file name instead of skipping silently, which keeps incomplete reports diagnosable. Good point on reading the 3.1-of-10-minutes split as container and Maven startup overhead rather than test time; that is the number I was after.

pgermosen (approved):
@hectorvent — both suggestions landed, and one of them landed better than I asked.

`db770a84` adds the two guards I flagged:

- `except (ET.ParseError, OSError)` around the parse, and a `try/except ValueError` around the
  `float`/`int` conversion — so the comma-decimal case (Surefire writing `time` under a
  container's default locale, with no locale pinned in the compat pom) degrades to a skipped
  suite instead of a traceback.
- `continue-on-error: true` on the report step, with a comment saying why, so the property
  holds regardless of what the parser grows into later.

The improvement on my version: I suggested a bare `continue`, and you print a warning naming
the file and the exception to stderr first. A silently skipped XML would have been
indistinguishable from a leg that produced none — now the log says which file and why, and
`continue-on-error` still keeps it off the job result. That's the better shape.

I checked the one thing that would have been quietly fatal here: `sys` is imported, so the new
`file=sys.stderr` calls resolve. A `NameError` on that path would have been masked by the very
`continue-on-error` this commit adds, which is the kind of failure that stays hidden for months.

My earlier approval stands — nothing since it changes the assessment.

## One note, no action

The `except ValueError` branch has no `continue`, which reads as though the failed row might
still be appended. It can't — the `rows.append` is the statement that raised, and the `except`
is the last thing in the loop body, so the effect is right either way. Purely how it reads.

PR #2510, opened 2026-08-24
fix(dynamodb): tolerate whitespace before a function's own parenthesis

### LINKED ISSUE #2509: [BUG] DynamoDB UpdateExpression: list_append with a space before the parenthesis is a silent no-op
## Service
DynamoDB

## AWS API Action
UpdateItem with an UpdateExpression using `list_append`

## Expected behavior
`SET #i = list_append (#i, :new)` (note the space between `list_append` and `(`) should behave identically to `SET #i = list_append(#i, :new)` (no space). AWS's UpdateExpression grammar is whitespace-insensitive around function calls, and real AWS DynamoDB accepts both forms. This matters in practice because PynamoDB (a widely used Python DynamoDB ORM, referenced in AWS's own docs) always emits the space form when generating `list_append` update expressions.

## Actual behavior
With the space present, `UpdateItem` returns success (no exception, no error of any kind) but the list is **not modified** - the write is silently dropped. Without the space, the exact same update works correctly.

This is a correctness-silent-failure bug: callers get a 200-equivalent success response and have no way to know their write didn't apply, short of reading the item back and diffing.

## Reproduction
```bash
docker run -d --name floci-repro -p 4566:4566 floci/floci:1.7.0-compat

  --table-name list-append-repro \
  --attribute-definitions AttributeName=id,AttributeType=S \

  --table-name list-append-repro \
  --item '{"id":{"S":"x"},"items":{"L":[{"S":"a"}]}}'

# Works correctly (no space):
aws --endpoint-url http://localhost:4566 dynamodb update-item \
  --table-name list-append-repro \
  --key '{"id":{"S":"x"}}' \
  --update-expression "SET #i = list_append(#i, :new)" \
  --expression-attribute-names '{"#i":"items"}' \
  --expression-attribute-values '{":new":{"L":[{"S":"b"}]}}'
# -> items becomes ["a", "b"] as expected

# Silently no-ops (space before the parenthesis):
aws --endpoint-url http://localhost:4566 dynamodb update-item \
  --table-name list-append-repro \
  --key '{"id":{"S":"x"}}' \
  --update-expression "SET #i = list_append (#i, :new)" \
  --expression-attribute-names '{"#i":"items"}' \
  --expression-attribute-values '{":new":{"L":[{"S":"c"}]}}'
# -> items is STILL ["a", "b"] - "c" was never appended, and the call returned success

aws --endpoint-url http://localhost:4566 dynamodb get-item --table-name list-append-repro --key '{"id":{"S":"x"}}'
# {"Item": {"id": {"S": "x"}, "items": {"L": [{"S": "a"}, {"S": "b"}]}}}  <- missing "c"
```

This exact form (`SET #0 = list_append (#0, :0)`, generated by PynamoDB) is what surfaced this - confirmed via botocore debug logging that PynamoDB always includes the space, then confirmed the space alone (not PynamoDB-specific serialization) is the trigger by replaying the identical expression via raw `aws dynamodb update-item` above.

## Environment
- Floci version / image tag: `floci/floci:1.7.0-compat` (haven't checked nightly, but this looks unrelated to the persistent-storage GSI bug fixed in #2463)
- Java SDK version (if applicable): using AWS CLI / boto3 (via PynamoDB), not the Java SDK

dedsecrattle:
## Summary

Fixes #2509

`applySetClause`/`evaluateSetExpr` gate `list_append` and `if_not_exists` on a literal `startsWith(functionName + "(")`. When there's a space between the function name and its own opening parenthesis - `"list_append (x, y)"` instead of `"list_append(x, y)"` - that check fails, the value falls through to the "plain attribute reference" branch, resolves to nothing, and the `SET` silently no-ops. No exception, no error - the write is just dropped.

This matters because PynamoDB (a widely used Python DynamoDB ORM) always emits the spaced form for `list_append`, and real AWS DynamoDB's UpdateExpression grammar is whitespace-insensitive around function calls, so this form works fine against real AWS. Confirmed via botocore debug logging on the PynamoDB side, then isolated to the space alone by replaying the identical expression through the raw AWS CLI (see #2509 for the full trace).

### Relationship to #1640

#1640 (already fixed) normalized whitespace between clause keywords (`SET`/`REMOVE`/etc.) and between a function's arguments. This is a distinct gap: whitespace between a function's own name and its own parenthesis, evaluated in the per-clause value evaluator rather than the top-level tokenizer, so #1640's fix and tests didn't reach it.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior:** an `UpdateExpression` with whitespace between a function name and its opening parenthesis — `SET #l = list_append (#l, :v)` — was silently ignored. `UpdateItem` returned success with no error, but the attribute was never modified. The same expression without the space worked correctly.

DynamoDB's `UpdateExpression` grammar is whitespace-insensitive around function calls, and real AWS accepts both forms. This is not a theoretical edge case: **PynamoDB always emits the spaced form** (`format_string = 'list_append ({0}, {1})'` in `pynamodb/expressions/operand.py`), so every `list_append` issued through that ORM was silently dropped rather than applied — a data-loss-shaped failure with no signal to the caller.

Verified with AWS CLI v2 against a locally built JVM jar:

```
SET #i = list_append(#i, :v)     → applied      (worked before this PR too)
SET #i = list_append (#i, :v)    → applied      (silently no-op'd before this PR)
SET #m = if_not_exists (#x, :v)  → applied      (same latent bug, also fixed)
```

## Fix

Added a small `startsWithFunctionCall(value, functionName)` helper that tolerates optional whitespace before the parenthesis, and used it at all three existing `list_append(`/`if_not_exists(` gate checks (case-sensitivity preserved per call site - one of the three was already case-sensitive, which I left as-is rather than changing unrelated behavior).

## Testing

- Added `spaceBeforeFunctionParenthesisIsAccepted` to `DynamoDbUpdateExpressionWhitespaceTest` (the existing whitespace-regression suite), covering both `list_append` and `if_not_exists` with the space form.
- Ran the full `io.github.hectorvent.floci.services.dynamodb.*Test` package: 302 tests, 0 failures.
- Manually verified via raw `awslocal dynamodb update-item` against a locally built JVM jar that: the spaced form now works, the existing non-spaced form still works (no regression), and `if_not_exists (...)` (also affected by the same pattern) now works too.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -1753,7 +1753,7 @@ public class DynamoDbService {
                     throw new AwsException("ValidationException",
                             "The parameter cannot be converted to a numeric value", 400);
                 }
-            } else if (valuePart.startsWith("if_not_exists(")) {
+            } else if (startsWithFunctionCall(valuePart, "if_not_exists")) {
                 // if_not_exists(attrRef, fallbackExpr) evaluates to:
                 //   attrRef's current value  — when attrRef exists in the item
                 //   fallbackExpr             — otherwise
@@ -1776,7 +1776,7 @@ public class DynamoDbService {
                         setValueAtPath(item, attrPath, resolved, exprAttrNames);
                     }
                 }
-            } else if (valuePart.toLowerCase().startsWith("list_append(")) {
+            } else if (startsWithFunctionCall(valuePart.toLowerCase(), "list_append")) {
                 int open = valuePart.indexOf('(');
                 int close = valuePart.lastIndexOf(')');
                 if (open >= 0 && close > open) {
@@ -1854,7 +1854,7 @@ public class DynamoDbService {
 
     private JsonNode evaluateSetExpr(ObjectNode item, String expr,
                                      JsonNode exprAttrNames, JsonNode exprAttrValues) {
-        if (expr.toLowerCase().startsWith("if_not_exists(")) {
+        if (startsWithFunctionCall(expr.toLowerCase(), "if_not_exists")) {
             String[] args = extractFunctionArgs(expr);
             if (args.length == 2) {
                 String checkAttr = resolveAttributeName(args[0].trim(), exprAttrNames);
@@ -2452,6 +2452,18 @@ public class DynamoDbService {
             }
         }
         return -1;
+    }
+
+    /**
+     * Whether {@code value} opens with a call to {@code functionName}, tolerating optional
+     * whitespace between the function name and the opening parenthesis. DynamoDB's
+     * UpdateExpression grammar is whitespace-insensitive there, but SDKs commonly emit the
+     * spaced form (e.g. PynamoDB always generates {@code "list_append (x, y)"}), which a plain
+     * {@code startsWith(functionName + "(")} rejects.
+     */
+    private static boolean startsWithFunctionCall(String value, String functionName) {
+        return value.startsWith(functionName)
+                && value.substring(functionName.length()).stripLeading().startsWith("(");
     }
 
     private String[] extractFunctionArgs(String funcCall) {
--- a/src/test/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbUpdateExpressionWhitespaceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbUpdateExpressionWhitespaceTest.java
@@ -159,6 +159,36 @@ class DynamoDbUpdateExpressionWhitespaceTest {
             .body("Attributes.c.N", equalTo("6"));
     }
 
+    /**
+     * Regression test for #2509: a space between a function name and its own opening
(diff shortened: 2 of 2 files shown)


dedsecrattle:
Hi @hectorvent ,

Can i get reviewed here ? This is a specific issue , Where `pynamodb` will cause No-Op since it generates the function with a space , and ideally The AWS services upstreams also supports these spaces , so we should be considering this , this was found as a part of migration from localstack

Thanks 🙏

hectorvent:
Thanks for the precise diagnosis, and sorry for the wait. Tracing this to the space PynamoDB emits and separating it cleanly from #1640's tokenizer fix made this easy to verify. The helper is nicely conservative: it still requires the exact function name, so attribute names that merely start with "list_append" keep falling through to the reference branch, and you preserved the one call site that was already case-sensitive rather than changing unrelated behavior.

I ran the full DynamoDB suite locally with your branch merged onto current main, 476 tests pass, including all 7 in the whitespace regression suite with your new spaced-form case. Your CI is also unblocked now and fully green, including the compatibility matrix.

No blockers from my side.

hectorvent (approved):
Approved, see comment above.

PR #2514, opened 2026-08-24
fix(secretsmanager): raise InvalidRequestException, not ResourceNotFoundException, for pending-deletion secrets

### LINKED ISSUE #2513: [BUG] Secrets Manager: operations on a pending-deletion secret raise ResourceNotFoundException instead of InvalidRequestException
## Service
Secrets Manager

## AWS API Action
GetSecretValue (and PutSecretValue, UpdateSecret, TagResource-adjacent target-attachment claiming, RotateSecret, UpdateSecretVersionStage) on a secret that's on the recovery-window deletion path

## Expected behavior
`DeleteSecret` without `ForceDeleteWithoutRecovery` schedules a secret for deletion after the recovery window (default 30 days) but keeps it recoverable via `RestoreSecret`. Real AWS raises `InvalidRequestException` ("You can't perform this operation on the secret because it was marked for deletion.") for most operations attempted on a secret in that state - distinct from `ResourceNotFoundException`, which means the secret is genuinely gone.

## Actual behavior
`GetSecretValue` (and the other operations listed above) raise `ResourceNotFoundException` instead, identical to what a truly nonexistent secret returns. This makes it impossible for a caller to distinguish "still recoverable, just marked for deletion" from "gone" by error code alone - which matters because some client code (ours included) specifically handles the pending-deletion case differently from the not-found case.

Interestingly, `BatchGetSecretValue` in the same file (`SecretsManagerService.java`) already gets this right - it raises `InvalidRequestException` with the correct message - so this looks like the other call sites just didn't get the same treatment.

## Reproduction
```bash
docker run -d --name floci-repro -p 4566:4566 -e SERVICES=secretsmanager floci/floci:1.7.0-compat

aws --endpoint-url http://localhost:4566 secretsmanager create-secret --name my-secret --secret-string test
aws --endpoint-url http://localhost:4566 secretsmanager delete-secret --secret-id my-secret
# {"ARN": "...", "Name": "my-secret", "DeletionDate": "<~30 days from now>"}  -- correctly scheduled, not force-deleted

aws --endpoint-url http://localhost:4566 secretsmanager get-secret-value --secret-id my-secret
# Actual:   ResourceNotFoundException: Secrets Manager can't find the specified secret.
# Expected: InvalidRequestException: You can't perform this operation on the secret because it was marked for deletion.
```

## Environment
- Floci version / image tag: `floci/floci:1.7.0-compat` (also reproduced building from current `main`)
- Java SDK version (if applicable): using AWS CLI / boto3, not the Java SDK
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker, and a locally built JVM jar from source

dedsecrattle:
## Summary

Fixes #2513

`DeleteSecret` without `ForceDeleteWithoutRecovery` sets `deletedDate` on the secret to a future date and keeps it in `store` (recoverable via `RestoreSecret`) - it's only removed from `store` outright on a force delete. That means any `secret.getDeletedDate() != null` check reached *after* `resolveSecret()` has already succeeded can only mean one thing: the secret is still present, just scheduled for deletion. `resolveSecret()` already raises `ResourceNotFoundException` on its own for a secret that's genuinely gone (force-deleted or never existed).

Real AWS reports the pending-deletion case as `InvalidRequestException` ("You can't perform this operation on the secret because it was marked for deletion."), distinct from `ResourceNotFoundException`. Six call sites - `getSecretValue`, `putSecretValue`, `updateSecret`, `claimTargetAttachment`, `rotateSecret`, and `updateSecretVersionStage` - instead raised `ResourceNotFoundException`, making a still-restorable secret indistinguishable by error code from one that's genuinely gone.

`batchGetSecretValue` in the same file already had this right (`InvalidRequestException` with the correct message), so there was already an internal precedent for the correct behavior - it just wasn't applied consistently. (As @pgermosen noted in review, `restoreSecret` is a second precedent: it already raises `InvalidRequestException` for the mirror-image "was not deleted" case.)

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior:** operations on a secret scheduled for deletion (the recovery-window path, not a force delete) raised `ResourceNotFoundException` instead of `InvalidRequestException`. That made a still-restorable secret indistinguishable, by error code alone, from one that is genuinely gone — so a caller can't tell "recoverable, call `RestoreSecret`" from "never existed".

`InvalidRequestException` in `secretsmanager/2017-10-17/service-2.json` documents this case explicitly as a cause: *"A parameter value is not valid for the current state of the resource. Possible causes: The secret is scheduled for deletion."* All six operations touched here declare `InvalidRequestException` in their error lists.

`DescribeSecret` and `ListSecretVersionIds` deliberately **not** touched — they don't declare `InvalidRequestException`, and describing a scheduled secret is how a caller reads its `DeletedDate`.

Verified against `floci/floci:1.7.0-compat` and a locally built JVM jar with AWS CLI v2:

```
get-secret-value            → InvalidRequestException   "You can't perform this operation…"
put-secret-value            → InvalidRequestException
update-secret               → InvalidRequestException
rotate-secret               → InvalidRequestException
update-secret-version-stage → InvalidRequestException
describe-secret             → 200   (correct — AWS allows this)
restore-secret              → 200   (correct)
```

`claimTargetAttachment` isn't reachable from the CLI, so it's covered by unit test rather than a CLI probe.

## Fix

Extracted the correct check into a shared `throwIfPendingDeletion(Secret secret)` helper (matching `batchGetSecretValue`'s existing message) and applied it at all six affected call sites.

## Testing

- Strengthened `deleteSecretWithRecoveryWindow`: it previously only asserted `AwsException.class`, which passes regardless of error code and didn't catch this bug. Now asserts the specific error code.
- Added `operationsOnPendingDeletionSecretRaiseInvalidRequestNotResourceNotFound`, covering all six affected operations.
- Ran the full `io.github.hectorvent.floci.services.secretsmanager.*Test` package: 118 tests, 0 failures.
- Manually verified via raw `aws secretsmanager` CLI against a locally built JVM jar: `GetSecretValue` on a scheduled-for-deletion secret now returns `InvalidRequestException` with the expected message instead of `ResourceNotFoundException`.

## Checklist

## Follow-up

Filed as #2549 per review: four more operations (`createSecret`, `tagResource`, `untagResource`, and a repeat `deleteSecret`) don't check pending-deletion state **at all** — including a data-loss path where `CreateSecret` silently overwrites a secret inside its recovery window. All pre-existing, none introduced here; kept out of this PR so the data-loss case is findable on its own and gets its own test.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerService.java
@@ -138,11 +138,7 @@ public class SecretsManagerService {
 
     public SecretVersion getSecretValue(String secretId, String versionId, String versionStage, String region) {
         Secret secret = resolveSecret(secretId, region);
-
-        if (secret.getDeletedDate() != null) {
-            throw new AwsException("ResourceNotFoundException",
-                    "Secrets Manager can't find the specified secret.", 400);
-        }
+        throwIfPendingDeletion(secret);
 
         SecretVersion version;
         if (versionId != null && !versionId.isEmpty()) {
@@ -178,11 +174,7 @@ public class SecretsManagerService {
                                         String secretBinary, String clientRequestToken, String region,
                                         List<String> versionStages) {
         Secret secret = resolveSecret(secretId, region);
-
-        if (secret.getDeletedDate() != null) {
-            throw new AwsException("ResourceNotFoundException",
-                    "Secrets Manager can't find the specified secret.", 400);
-        }
+        throwIfPendingDeletion(secret);
 
         if (clientRequestToken != null && (clientRequestToken.length() < 32 || clientRequestToken.length() > 64)) {
             throw new AwsException("InvalidParameterException", "ClientRequestToken must be between 32 and 64 characters long.", 400);
@@ -285,11 +277,7 @@ public class SecretsManagerService {
 
     public Secret updateSecret(String secretId, String description, String kmsKeyId, String region) {
         Secret secret = resolveSecret(secretId, region);
-
-        if (secret.getDeletedDate() != null) {
-            throw new AwsException("ResourceNotFoundException",
-                    "Secrets Manager can't find the specified secret.", 400);
-        }
+        throwIfPendingDeletion(secret);
 
         if (description != null) {
             secret.setDescription(description);
@@ -370,10 +358,7 @@ public class SecretsManagerService {
         Secret resolved = resolveSecret(secretId, region);
         synchronized (lockFor(resolved.getArn())) {
             Secret secret = resolveSecret(resolved.getArn(), region);
-            if (secret.getDeletedDate() != null) {
-                throw new AwsException("ResourceNotFoundException",
-                        "Secrets Manager can't find the specified secret.", 400);
-            }
+            throwIfPendingDeletion(secret);
 
             String existingOwner = secret.getTargetAttachmentOwner();
             if (existingOwner != null && !existingOwner.equals(attachmentOwner)) {
@@ -585,11 +570,7 @@ public class SecretsManagerService {
     public Secret rotateSecret(String secretId, String clientRequestToken, String rotationLambdaArn, Secret.RotationRules rotationRules,
                                boolean rotateImmediately, String region) {
         Secret secret = resolveSecret(secretId, region);
-
-        if (secret.getDeletedDate() != null) {
-            throw new AwsException("ResourceNotFoundException",
-                    "Secrets Manager can't find the specified secret.", 400);
-        }
+        throwIfPendingDeletion(secret);
 
         if (clientRequestToken != null && (clientRequestToken.length() < 32 || clientRequestToken.length() > 64)) {
             throw new AwsException("InvalidParameterException", "ClientRequestToken must be between 32 and 64 characters long.", 400);
@@ -838,9 +819,7 @@ public class SecretsManagerService {
         }
 
         Secret secret = resolveSecret(secretId, region);
-        if (secret.getDeletedDate() != null) {
-            throw new AwsException("ResourceNotFoundException", "Secrets Manager can't find the specified secret.", 400);
-        }
+        throwIfPendingDeletion(secret);
 
         SecretVersion versionByStage = findVersionByStage(secret, versionStage);
         String currentVersionId = versionByStage != null
@@ -931,6 +910,22 @@ public class SecretsManagerService {
     ) {
         public static BatchGetSecretValueResult empty() {
             return new BatchGetSecretValueResult(Collections.emptyList(), Collections.emptyList());
+        }
+    }
+
+    /**
+     * A secret found by {@link #resolveSecret} with {@code deletedDate} set is on the
+     * recovery-window path (still in {@code store}, still restorable) - {@link
+     * #deleteSecret} force-deletes by removing the entry from {@code store} outright, so a
+     * fully, permanently gone secret never reaches this check; {@code resolveSecret} throws
+     * ResourceNotFoundException for that case on its own. Real AWS's error for the
+     * recoverable case is InvalidRequestException, matching what {@code batchGetSecretValue}
+     * already does correctly - the other call sites threw ResourceNotFoundException instead.
+     */
+    private void throwIfPendingDeletion(Secret secret) {
+        if (secret.getDeletedDate() != null) {
+            throw new AwsException("InvalidRequestException",
+                    "You can't perform this operation on the secret because it was marked for deletion.", 400);
         }
     }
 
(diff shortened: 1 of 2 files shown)


pgermosen:
Thanks for this @dedsecrattle — the diagnosis is careful and the fix is right. I checked it against the service model and a running instance rather than reading it, and it holds up. A couple of things to sort out before merge, and one adjacent finding that I think deserves its own issue.

## Verified

Your central claim is documented, not just observed. `InvalidRequestException` in `secretsmanager/2017-10-17/service-2.json` lists its causes explicitly, first among them:

> A parameter value is not valid for the current state of the resource. Possible causes: **The secret is scheduled for deletion.**

And all six operations you touched declare `InvalidRequestException` in their error lists — `GetSecretValue`, `PutSecretValue`, `UpdateSecret`, `RotateSecret`, `UpdateSecretVersionStage`, plus `BatchGetSecretValue` which already had it. Equally important, `DescribeSecret` and `ListSecretVersionIds` **don't** declare it, and you correctly left both alone: describing a scheduled secret is how a caller sees its `DeletedDate`.

Driven through the CLI against a secret with a recovery window:

```
get-secret-value            → InvalidRequestException  "You can't perform this operation…"
put-secret-value            → InvalidRequestException
update-secret               → InvalidRequestException
rotate-secret               → InvalidRequestException
update-secret-version-stage → InvalidRequestException
describe-secret             → 200            (correct — AWS allows this)
restore-secret              → 200            (correct)
```

`claimTargetAttachment` isn't reachable from the CLI, so that one rests on your unit test rather than my probe.

The point about the old test is exactly right, and I confirmed it: reverting your helper to `ResourceNotFoundException` fails **both** tests —

```
deleteSecretWithRecoveryWindow:261
    expected: <InvalidRequestException> but was: <ResourceNotFoundException>
operationsOnPendingDeletionSecretRaiseInvalidRequestNotResourceNotFound:269
    expected: <InvalidRequestException> but was: <ResourceNotFoundException>
```

— which the bare `assertThrows(AwsException.class, …)` would have sailed through. Strengthening it was the right instinct. 113 Secrets Manager tests pass on a merge with current `main`, and the merge is clean.

Also worth noting you had a second internal precedent you didn't mention: `restoreSecret` already raises `InvalidRequestException` for the mirror-image case ("because it was not deleted"). So the file had the shape right in two places out of eight.

---

## Please add the PR template sections

The description is genuinely more informative than the template would have been — the What broke / Fix / Testing structure is good. But it's missing `## Type of change`, `## AWS Compatibility`, and `## Checklist`, which the repo's template provides and which release tooling and triage read. `Fixes #2513` is already there, so it's just the three headers. No need to rewrite the prose you have — please keep it and add the sections around it.

## What else I found

While confirming the six, I checked every other operation AWS models `InvalidRequestException` for. **The underlying gap is wider than six call sites** — though I want to be fair about this: your statement is accurate as written, because those six are the ones raising the *wrong error code*. The four below don't raise a wrong code, they don't check at all, so they didn't show up in the same search. All pre-existing, none introduced here.

Probed on the same pending-deletion secret:

```
tag-resource            → 200   (AWS: InvalidRequestException)
untag-resource          → 200   (AWS: InvalidRequestException)
delete-secret (again)   → 200, silently re-schedules   (AWS: InvalidRequestException)
```

**And one that is worse than a wrong error code — it loses data:**

```
create-secret --name s2 --secret-string ORIGINAL-VALUE
delete-secret --secret-id s2                     # 30-day recovery window
create-secret --name s2 --secret-string NEW-VALUE   → 200
restore-secret --secret-id s2
get-secret-value --secret-id s2                  → "NEW-VALUE"
```

`ORIGINAL-VALUE` is gone. `createSecret`'s guard is `existing != null && existing.getDeletedDate() == null`, so a name scheduled for deletion falls straight through and overwrites the recoverable secret at the same storage key. Real AWS refuses this with `InvalidRequestException` — `CreateSecret` declares it — precisely so the recovery window means what it says.

I'd rather that didn't ride along on this PR: it's a different symptom, it needs its own test, and the data-loss framing deserves to be findable on its own. **Would you like to file it as a separate issue?** Happy to do it myself if you'd prefer — just say. The tag/untag/re-delete trio would fit naturally in the same follow-up, since they share your `throwIfPendingDeletion` helper as the fix.

## One small thing

`deleteSecret`'s force path calls `secret.setDeletedDate(Instant.now())` *after* `store.delete(...)`, on an object already removed from the store. It's harmless — the mutated copy is only used to build the response, and AWS does return a `DeletionDate` on a force delete — but it reads oddly next to the recovery-window branch, where the same setter is the thing that persists state. Not something to change here.

Nothing blocking on the code. Add the template sections and let me know on the follow-up issue, and I'm happy to approve this.

dedsecrattle:
@pgermosen  I have updated the PR description to follow the template, and also filed a new issue for the follow up , I will raise a PR for that separately 

Thanks !

pgermosen (approved):
Looks good.

PR #2515, opened 2026-08-24
Round 7: launch-template Placement.Tenancy stops being synthesized on read

lex00:
## Summary
- Fixes lex00/floci#126: `DescribeLaunchTemplateVersions` invented `Placement.Tenancy = "default"` for a launch template whose creator never set tenancy at all.
- Root cause: `Placement.tenancy` carries a `"default"` field initializer on the shared model class (correct for a running `Instance`'s Placement, which always reports a tenancy), but a launch template is a stored document, not a running instance - AWS's Describe leaves an unset field absent. `parseLaunchTemplatePlacement` only called `setTenancy()` when the request carried the field, so an unset tenancy silently kept the class default instead of becoming `null`, and it leaked onto the wire whenever any other placement field (e.g. `GroupName`) made the block non-empty.
- Fix: always assign the parsed value (`null` when absent) so it overrides the class field default. A template that DOES set tenancy still round-trips it - round 6's own `Ec2LaunchTemplateConnectionTrackingAndTenancyIntegrationTest` stays green.
- Sibling check: walked the rest of the launch-template read path (`Ec2QueryHandler`'s `parseLaunchTemplate*` family and every `LaunchTemplateData` nested model class) for the same "field default leaks onto Describe" shape as round 6's SG-conflation fix. Found no other sibling in the launch-template family - every other field already returns null/absent when unset. Other `"default"`-initialized fields elsewhere in the ec2 model package (`CapacityReservation.tenancy`, `Vpc.instanceTenancy`, `Image.state`, etc.) belong to real running/live resources, where AWS's Describe always reports a value - a legitimately different case from a stored template document, so left untouched.

## Test plan
- [x] New `Ec2LaunchTemplatePlacementTenancyAbsenceIntegrationTest` asserts the raw `DescribeLaunchTemplateVersions` wire response: an unset tenancy stays absent even when another placement field is set, and an explicitly-set tenancy still round-trips.
- [x] Verified load-bearing by a temporary revert-and-confirm-red cycle (fails with `<tenancy>default</tenancy>` synthesized without the fix).
- [x] `./mvnw -Dtest='io.github.hectorvent.floci.services.ec2.**' test` - 496/496 passing, including round 6's `Ec2LaunchTemplateConnectionTrackingAndTenancyIntegrationTest` and `Ec2LaunchTemplateFieldsIntegrationTest`.

https://claude.ai/code/session_01VYoTpda19fzdxqiC5mLuWX


lex00:
Opened against the wrong base by mistake (should target lex00/floci, not upstream). Closing; re-opening against my fork.

PR #2519, opened 2026-08-24
fix(s3): route the replication subresource instead of falling through

### LINKED ISSUE #2296: [BUG] s3: PUT/GET ?replication is unrouted (only DELETE was patched) and falls through
## Service

S3

## AWS API Action

`PutBucketReplication` / `GetBucketReplication`

## Expected behavior

`PUT /{bucket}?replication` stores the replication configuration and returns `200`.
`GET /{bucket}?replication` returns the stored `ReplicationConfiguration`, or
`ReplicationConfigurationNotFoundError` (404) when none has been set — which is what the SDK and
the Terraform provider treat as "not configured".

## Actual behavior

Only `DELETE /{bucket}?replication` is routed. `S3Controller` handles it explicitly, and the
comment there states the reason:

```java
if (hasQueryParam(uriInfo, "replication")) {
    // Floci does not model bucket replication; DeleteBucketReplication is a
    // no-op that always returns 204, matching real S3. Crucially it must be
    // handled here so it does NOT fall through to deleting the whole bucket.
    return Response.noContent().build();
}
```

The same fall-through hazard exists on PUT and GET and was not patched:

- **`PUT /{bucket}?replication`** falls through to `s3Service.createBucket(...)`, which throws
  `BucketAlreadyOwnedByYou` (`S3Service:214`) because the bucket already exists. Configuring
  replication fails with an error about bucket creation.
- **`GET /{bucket}?replication`** falls through to `listObjectsWithPrefixes` and returns a
  `ListBucketResult`. The SDK's XML unmarshaller does not reject the mismatched root element, so
  it reports an empty replication configuration instead of the 404 that means "not configured" —
  a silent wrong answer rather than an error.

**Impact.** `aws_s3_bucket_replication_configuration` cannot be applied, and the read path
cannot distinguish "no replication" from "replication misreported as empty".

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 s3api create-bucket --bucket example

aws --endpoint-url http://localhost:4566 s3api get-bucket-replication --bucket example
# Expected: ReplicationConfigurationNotFoundError
# Actual: an empty result, parsed from a ListBucketResult
```

## Suggested fix

Route `replication` in the PUT and GET bucket handlers alongside the existing DELETE case.
Storing the configuration verbatim and echoing it back is sufficient — as the existing comment
says, Floci does not model replication behavior, and neither the SDK nor the provider requires
it to. What they do require is that the three verbs are distinguishable from `CreateBucket` and
`ListObjects`.

`GET` should return `ReplicationConfigurationNotFoundError` with HTTP 404 when nothing is
stored, rather than an empty document.

Filed separately from the `?accelerate` gap, which has the same fall-through failure but is a
different subresource and separately fixable.

## Environment

## AWS Documentation

- https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketReplication.html
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketReplication.html

soreavis:
## Summary

Routes `PUT` and `GET /{bucket}?replication` alongside the DELETE case that was already handled, exactly as the issue suggests: the configuration is stored verbatim and echoed back, since Floci doesn't model replication behavior and neither the SDK nor the Terraform provider needs it to — they need the three verbs to be distinguishable from `CreateBucket` and `ListObjects`. Closes #2296.

- `PUT` validates the `ReplicationConfiguration` root (`MalformedXML` 400 on anything else, checked before storing) and stores the body on the bucket, answering the AWS-documented empty 200. No rule-level validation, per the issue's verbatim-store scope.
- `GET` echoes the stored configuration, or answers `ReplicationConfigurationNotFoundError` (404) when none is stored — the code `aws_s3_bucket_replication_configuration` matches on to mean "not configured". Before this it fell through to `ListObjects` and the SDK parsed the `ListBucketResult` as an empty configuration, a silent wrong answer.
- `DELETE` now clears the stored configuration instead of no-opping (the no-op was only correct while nothing could be stored). Still an unconditional 204 on an existing bucket, matching AWS.
- `IamActionRegistry` maps the three verbs: `s3:PutReplicationConfiguration` for PUT **and** DELETE (AWS documents that `DeleteBucketReplication` requires the put action), `s3:GetReplicationConfiguration` for GET. The mapping mirrors the controller's per-method dispatch order — PUT and GET dispatch accelerate ahead of replication, while the DELETE chain routes replication and never routes accelerate to an operation — so a request that executes another subresource's operation keeps that operation's resolution.

Two deliberate behavior changes beyond the routing, both pinned by tests: `DELETE ?replication` on a missing bucket now answers `NoSuchBucket` 404 (the old no-op answered 204 before any bucket check), and the three verbs move off the rule-table fallback (`s3:CreateBucket`/`s3:ListBucket`/`s3:DeleteBucket`) onto the real actions.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

Verified against the API reference for all three operations: `PutBucketReplication` (root `Required: Yes`, 200 empty body, `s3:PutReplicationConfiguration`), `GetBucketReplication` (200 `ReplicationConfiguration`, `s3:GetReplicationConfiguration`), `DeleteBucketReplication` (204 empty, authorized by the put action). The 404 code string is the one the Terraform provider matches (`errCodeReplicationConfigurationNotFound = "ReplicationConfigurationNotFoundError"` in `internal/service/s3/errors.go`). Content-MD5 is not enforced, matching every sibling subresource here.

## Checklist

- [x] `./mvnw test` run locally: 11,944 tests, with failures only in 23 Docker-backed classes (Docker isn't available on this machine — API Gateway/WebSocket, CloudFormation ECR/SAM, ECR, ELBv2-Lambda, Lambda, Neptune, SWF); the s3 and iam packages are fully green (1,276/0 in that run). New `S3ReplicationConfigurationIntegrationTest` (13 tests, incl. the two fall-through regressions and a wire-level accelerate/replication precedence test), 5 new `IamActionRegistryTest` cases (per-method precedence, dispatched-first yields, object-path guard), and the accelerate suite stays green (13/13). Red-proof: without the routes 11 of 13 integration tests fail; without the registry change the 3 new-resolution cases fail.

### DIFF
--- a/docs/services/s3.md
+++ b/docs/services/s3.md
@@ -148,7 +148,6 @@ When `FLOCI_SERVICES_ATHENA_MOCK=true` is set, Athena queries are stubbed but fl
 
 These AWS S3 features have no handler in Floci. Calls will return an error (typically `404` or `NoSuchBucket`-style):
 
-- Replication (`PutBucketReplication`, `GetBucketReplication`, `DeleteBucketReplication`)
 - Access logging (`PutBucketLogging`, `GetBucketLogging`)
 - Request payment (`PutBucketRequestPayment`, `GetBucketRequestPayment`)
 - Intelligent-Tiering configurations
--- a/src/main/java/io/github/hectorvent/floci/services/iam/IamActionRegistry.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/IamActionRegistry.java
@@ -170,6 +170,12 @@ public class IamActionRegistry {
             "uploads", "notification", "versioning", "versions", "location", "object-lock",
             "website", "logging", "policy", "cors", "lifecycle", "encryption",
             "publicAccessBlock", "ownershipControls", "requestPayment");
+    // S3Controller's DELETE chain dispatches these ahead of replication (tagging is
+    // resolved above). Accelerate is NOT here: on DELETE it is dispatched after
+    // replication. Mirrors the controller's dispatch order — extend together.
+    private static final List<String> DELETE_SUBRESOURCES_BEFORE_REPLICATION = List.of(
+            "website", "policy", "cors", "lifecycle", "encryption",
+            "publicAccessBlock", "ownershipControls");
 
     /**
      * Resolves S3 sub-resource ops (ACL, tagging, retention, etc.) that
@@ -182,7 +188,8 @@ public class IamActionRegistry {
         boolean acl = params.containsKey("acl");
         boolean tagging = params.containsKey("tagging");
         boolean accelerate = params.containsKey("accelerate");
-        if (!acl && !tagging && !accelerate) {
+        boolean replication = params.containsKey("replication");
+        if (!acl && !tagging && !accelerate && !replication) {
             return null;
         }
         // /{bucket}?acl → bucket-level; /{bucket}/{key}?acl → object-level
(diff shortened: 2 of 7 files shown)


hectorvent:
Thanks for a really thorough fix, the fall-through to ListBucketResult was a nasty silent wrong answer and the writeup made verifying this easy.

Everything checked out against the model and references: the three routes match the S3 API (PUT and GET on the bucket, DELETE at 204), the 404 code string is exactly what the Terraform provider matches on in its errors.go, and mapping DELETE to s3:PutReplicationConfiguration follows the documented authorization. The precedence test that pins the controller dispatch order against the IAM mapping is a particularly nice touch, since those two must move together.

No blockers from my side.

hectorvent (approved):
Approved, see comment above.

PR #2525, opened 2026-08-24
fix(rds): reject a duplicate create while provisioning is in flight

### LINKED ISSUE #2051: [BUG] Retried CreateDBCluster provisions a second backing container, silently splitting cluster and instance endpoints across two databases
## Service

RDS

## AWS API Action

CreateDBCluster (retried / concurrent request with the same identifier)

## Expected behavior

`CreateDBCluster` should be idempotent with respect to the cluster identifier. A
retried request for an identifier that is already being provisioned should either
return the in-flight/existing cluster or fail with `DBClusterAlreadyExistsFault` —
it should never provision a *second* backing container for the same cluster.

Real AWS behaves this way: a client-side retry of `CreateDBCluster` never results in
two independent databases sharing one cluster identifier.

## Actual behavior

Provisioning is not guarded against a concurrent request for the same identifier.
When the backing image pull takes longer than the AWS CLI's client-side retry
window, the retried request re-enters provisioning and starts a **second** Postgres
container. Both containers are registered under the same instance identifier, and
the proxies are then wired to *different* backends:

```
15:36:20 RdsQueryHandler] RDS action: CreateDBCluster
15:36:20 RdsContainerManager] Starting RDS backend container for instance: arity-db-cluster-dev engine=POSTGRES
15:36:21 ImageCacheService] Pulling image: postgres:17.5
15:37:21 RdsQueryHandler] RDS action: CreateDBCluster          <-- CLI retry, ~60s later
15:37:21 RdsContainerManager] Starting RDS backend container for instance: arity-db-cluster-dev engine=POSTGRES
15:37:51 ImageCacheService] Image pulled successfully: postgres:17.5
15:37:51 ContainerLifecycleManager] Created container ... (name=floci-rds-2b44b3)
15:37:51 ContainerLifecycleManager] Created container ... (name=floci-rds-974f7f)
15:37:52 RdsContainerManager] RDS backend for instance arity-db-cluster-dev: 10.0.2.6:5432
15:37:52 RdsContainerManager] RDS backend for instance arity-db-cluster-dev: 10.0.2.5:5432
15:37:53 RdsAuthProxy] RDS proxy started for instance arity-db-cluster-dev on port 4614 -> 10.0.2.5:5432
```

A subsequent `CreateDBInstance` in the same cluster then attaches to the *other*
container:

```
15:37:54 RdsAuthProxy] RDS proxy started for instance arity-db-cluster-dev on port 4615 -> 10.0.2.6:5432
```

The result is two independent PostgreSQL databases behind one cluster identifier:
the cluster endpoint reaches container A, the instance endpoint reaches container B.

**No error is raised anywhere.** `CreateDBCluster` returns `"Status": "available"`,
`CreateDBInstance` returns `"DBInstanceStatus": "available"`, and
`DescribeDBInstances` keeps reporting `available` — so a readiness check built on
instance status passes while the application is talking to a database that the
reported instance never sees. Writes land in one database and reads miss them:

```
$ psql -p 4614 -c 'create table probe(id int); insert into probe values (1);'
INSERT 0 1
$ psql -p 4615 -c 'select count(*) from probe;'
ERROR:  relation "probe" does not exist
```

This is the same class of provisioning-guard bug as #1741 (Neptune
`CreateDBCluster` leaking ports and orphaning containers), except the failure mode
here is *duplication under retry* rather than *leak on failure*, and it is silent
rather than surfacing as capacity exhaustion.

## Reproduction

The trigger is simply that provisioning takes longer than the client-side retry
window, which is guaranteed on any cold image cache — for example the first run in
CI.

```bash
# 1. Guarantee a cold pull.
docker rmi postgres:17.5

# 2. Start floci with an RDS proxy range wide enough for a cluster + instance.
docker run -d --name floci \
  -e FLOCI_SERVICES_RDS_PROXY_BASE_PORT=4511 \
  -e FLOCI_SERVICES_RDS_PROXY_MAX_PORT=4513 \
  -e FLOCI_SERVICES_RDS_DEFAULT_POSTGRES_IMAGE=postgres:17.5 \
  -p 4566:4566 -p 4511-4513:4511-4513 \
  floci/floci:1.5.34-compat

# 3. Create a cluster and an instance in it.
awslocal rds create-db-cluster \
  --db-cluster-identifier probe-cluster \
  --engine aurora-postgresql --engine-version 17.5 \
  --master-username probe --master-user-password probe123 \
  --database-name probe
awslocal rds create-db-instance \
  --db-instance-identifier probe-cluster \
  --db-instance-class db.t3.medium \
  --engine aurora-postgresql --db-cluster-identifier probe-cluster

# 4. Two containers instead of one:
docker ps --format '{{.Names}}' | grep floci-rds
# floci-rds-e5844d
# floci-rds-80292e

# 5. Write via the cluster endpoint, read via the instance endpoint:
PGPASSWORD=probe123 psql -h 127.0.0.1 -p 4511 -U probe -d probe \
  -c 'create table probe(id int); insert into probe values (1);'
PGPASSWORD=probe123 psql -h 127.0.0.1 -p 4512 -U probe -d probe \
  -c 'select count(*) from probe;'
# ERROR:  relation "probe" does not exist
```

Pre-pulling `postgres:17.5` before step 3 makes it go away: one container, both
proxies pointing at it. That is the workaround, but it makes correctness depend on
the image cache being warm.

Reproduced consistently — three separate cold-cache runs, three splits; every
warm-cache run was clean.

## Suggested fix

Guard provisioning per cluster/instance identifier so that a second in-flight
`CreateDBCluster` for the same identifier joins or rejects rather than provisioning
again — e.g. a per-identifier lock held across the pull + container start, with the
duplicate request returning the existing cluster or `DBClusterAlreadyExistsFault`.

## Environment

- Floci version / image tag: `floci/floci:1.5.34-compat` (also seen on `1.5.34`)
- SDK: AWS CLI v2 via `awslocal` (bundled in the `-compat` image)
- How you're running Floci: Docker, with `/var/run/docker.sock` bind-mounted
  (Docker-backed RDS provisioning)
- Host: macOS on arm64, Docker Desktop 4.11.1 / engine 20.10.17

soreavis:
## Summary

`CreateDBCluster` checks for an existing identifier, provisions the backing container, and only then registers the cluster — so a client-side retry arriving during a cold-cache image pull passes the check again and starts a second container under the same identifier, silently splitting the cluster and instance endpoints across two databases (the issue's psql repro). `CreateDBInstance` has the identical race. Closes #2051.

This implements the reject arm of the issue's suggested fix: a per-identifier provisioning sentinel (`ConcurrentHashMap.newKeySet()`) claimed before the exists-check and released in `finally` after registration. A duplicate arriving mid-provision fails fast with the fault code and 400 status a completed create produces — `DBClusterAlreadyExistsFault` / `DBInstanceAlreadyExists` — and only one container ever starts. A failed create releases the identifier on the way out, so a clean retry works — nothing is registered on that path, so a retained sentinel would wedge the identifier permanently. Cluster and instance identifiers remain independent namespaces — the issue's own repro provisions both under one name. Joining the in-flight create instead would park the retrying client on the remainder of the pull, and AWS doesn't document `CreateDBCluster` as idempotent-return.

Mechanically the change is a rename-and-delegate: the innermost create overloads became guard wrappers and their bodies moved verbatim to private `doCreate…` methods, so the provisioning logic itself is untouched. A non-blocking sentinel is used instead of the `synchronized (lockFor(...))` pattern some sibling services use, since holding a lock here would park the duplicate's thread on the remainder of the image pull rather than failing it fast.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

The duplicate answers the fault the operation already defines for an existing identifier (`DBClusterAlreadyExistsFault` 400 / `DBInstanceAlreadyExists` 400 — the codes and status this service already uses for the completed-create case), which is what a retried create against real AWS receives once the first request has registered the resource.

## Checklist

- [x] `./mvnw test` run locally: 11,934 tests, with failures confined to 23 Docker-backed classes (no Docker socket on this machine) — the identical failing set a sibling branch with a disjoint, non-overlapping diff produces here, and none of them is an RDS class. The RDS package is fully green (388/388), with `RdsServiceTest` at 232/232 including eight new tests — a sequential duplicate, two latch-based concurrency tests that hold the first create inside `containerManager.start` and require the duplicate to fault with a 400 while exactly one container starts, a failure-releases-the-identifier retry test, delete-then-recreate tests for both kinds pinning that a successful create releases the identifier, a sentinel-ordering test that blocks the store write to prove the identifier is held until registration completes, and a same-identifier cluster+instance test that must succeed mid-provision. Red-proof: with the guard reverted, the three duplicate-while-provisioning tests fail on the 5s latch timeout — the duplicate slips straight into provisioning. Mutation check: nine guard mutants — guard-never-fires for each kind, missing release, release-only-on-exception for each kind, release-before-registration, kind-prefix collapse, fault-code value, fault-status value — each killed by a named test.
- [x] New or updated integration test added — eight tests in `RdsServiceTest` against a mocked `RdsContainerManager`; the race needs a thread held inside `containerManager.start`, which a Docker-backed integration test can't hold deterministically.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
@@ -126,6 +126,13 @@ public class RdsService implements Resettable {
     private final StorageBackend<String, DbProxyTargetGroup> proxyTargetGroups;
     private final RdsContainerManager containerManager;
     private final RdsProxyManager proxyManager;
+    // CreateDBCluster/CreateDBInstance register the new resource only after the
+    // backing container is provisioned, and an image pull can hold that window
+    // open long enough for a client-side retry to pass the exists-check again
+    // and provision a second container for the same identifier. The sentinel
+    // serializes creates per identifier so the duplicate fails fast with the
+    // same fault a completed create produces.
+    private final Set<String> provisioningIds = ConcurrentHashMap.newKeySet();
     private final Ec2Service ec2Service;
     private final RegionResolver regionResolver;
     private final EmulatorConfig config;
@@ -461,6 +468,36 @@ public class RdsService implements Resettable {
                                        String optionGroupName,
                                        String region,
                                        boolean autoMinorVersionUpgrade) {
+        String provisioningKey = "instance:" + currentAccountId() + ":"
+                + dbResourceKey(effectiveRegion(region), id);
+        if (!provisioningIds.add(provisioningKey)) {
+            throw new AwsException("DBInstanceAlreadyExists",
+                    "DB instance " + id + " already exists.", 400);
+        }
+        try {
+            return doCreateDbInstance(id, engineParam, engineVersion, masterUsername,
+                    masterPassword, dbName, dbInstanceClass, allocatedStorage, iamEnabled,
+                    paramGroupName, dbSubnetGroupName, dbClusterIdentifier, availabilityZone,
+                    multiAz, manageMasterUserPassword, masterUserSecretKmsKeyId, tags,
+                    vpcSecurityGroupIds, optionGroupName, region, autoMinorVersionUpgrade);
+        } finally {
+            provisioningIds.remove(provisioningKey);
+        }
+    }
+
+    private DbInstance doCreateDbInstance(String id, String engineParam, String engineVersion,
+                                          String masterUsername, String masterPassword,
+                                          String dbName, String dbInstanceClass,
+                                          int allocatedStorage, boolean iamEnabled,
+                                          String paramGroupName, String dbSubnetGroupName,
+                                          String dbClusterIdentifier, String availabilityZone,
+                                          boolean multiAz, boolean manageMasterUserPassword,
+                                          String masterUserSecretKmsKeyId,
+                                          Map<String, String> tags,
+                                          List<String> vpcSecurityGroupIds,
+                                          String optionGroupName,
+                                          String region,
+                                          boolean autoMinorVersionUpgrade) {
         String effectiveRegion = effectiveRegion(region);
         String dbiResourceId = "db-" + java.util.UUID.randomUUID().toString()
                 .replace("-", "").substring(0, 24).toUpperCase();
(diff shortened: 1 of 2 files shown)


hectorvent:
Thanks, this is a model concurrency fix: the latch tests that hold the first create inside the container start, the red-proof, and the named mutation kills make it easy to trust.

The fault codes are exactly right per the model, including the asymmetry where the instance shape DBInstanceAlreadyExistsFault carries code DBInstanceAlreadyExists while the cluster shape's code matches its own name. Nice catch keeping cluster and instance identifiers in separate namespaces too.

One heads up rather than a blocker: #2484 reworks the same createDbInstance entry point, so whichever of the two lands second will need a small mechanical rebase.

No blockers from my side.

hectorvent (approved):
Approved, see comment above.

Issue #2530, opened 2026-08-24
[BUG] Cognito UpdateUserAttributes allows clients to set verification status flags


dzwicker:
## Problem

`UpdateUserAttributes` currently allows an authenticated user to submit `email_verified` or `phone_number_verified` directly. This lets the caller bypass the Cognito verification flow by setting its own verification status.

AWS reserves these status attributes for successful verification flows and administrative updates.

## Expected behavior

- Self-service `UpdateUserAttributes` requests containing `email_verified` or `phone_number_verified` fail with `InvalidParameterException` and HTTP 400.
- The request is rejected atomically: no other attribute from the same request is partially persisted.
- `AdminUpdateUserAttributes` continues to allow administrators to set the verification flags.

## Scope

Keep the fix limited to self-service validation and focused tests. Pending-attribute behavior, admin cleanup, and concurrency handling are out of scope.

Issue #2529, opened 2026-08-24
Cognito: clear stale pending verification state after admin contact changes


dzwicker:
## Problem

After a user starts a verification-required `email` or `phone_number` update, `AdminUpdateUserAttributes` can replace that contact and `AdminDeleteUserAttributes` can remove it while the old pending value and verification code remain.

A later verification attempt can therefore act on stale state, or `GetUserAttributeVerificationCode` can target an obsolete pending destination.

This issue follows #2021 and depends on its pending-attribute state.

## Expected behavior

When an admin update or delete touches `email` or `phone_number`:

- clear only that attribute's pending value;
- invalidate its matching attribute-verification code;
- preserve unrelated pending contact attributes and custom attributes;
- remain idempotent when no pending value or verification code exists.

## Scope

Keep the change limited to admin contact updates/deletes and focused service tests. Concurrency or general transaction handling is out of scope.

PR #2531, opened 2026-08-24
feat(cognito): support pending user attribute verification

### LINKED ISSUE #2021: [FEAT] Cognito user attribute updates lack pending verification semantics
## Service

Cognito

## API Action / Feature

AWS-compatible email and phone-number update verification across:

- `UpdateUserAttributes`
- `VerifyUserAttribute`
- `UserAttributeUpdateSettings.AttributesRequireVerificationBeforeUpdate`

Related work:

- #2017
- #2018

## AWS Documentation

- https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-email-phone-verification.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_VerifyUserAttribute.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UserAttributeUpdateSettingsType.html

## Why is this needed?

AWS Cognito supports two distinct behaviors when a signed-in user changes `email` or `phone_number`:

1. When `AttributesRequireVerificationBeforeUpdate` includes the attribute, the existing verified value remains active and the new value remains pending until successful verification.
2. Without that requirement, Cognito updates the value immediately but treats the new value as unverified until verification succeeds.

Floci currently applies `UpdateUserAttributes` through `adminUpdateUserAttributes`, which directly overwrites the stored value. It does not:

- reset the corresponding `email_verified` or `phone_number_verified` state,
- retain a pending value,
- promote a pending value during `VerifyUserAttribute`,
- model `AttributesRequireVerificationBeforeUpdate`.

Without this behavior, tests can pass against Floci while behaving differently against AWS when users change an already verified email address or phone number.

This is intentionally out of scope for #2017 and #2018, which implement submission of an attribute verification code for the currently stored value.

Acceptance coverage should include both user-pool configurations and verify:

- stored and pending attribute values,
- verified flags before and after verification,
- sign-in alias behavior while an update is pending,
- promotion of the pending value after a correct code,
- behavior after an incorrect or expired code.

- [x] Yes
- [ ] No

dzwicker:
## Summary

- persist `UserAttributeUpdateSettings.AttributesRequireVerificationBeforeUpdate`
- keep configured email and phone updates pending until `VerifyUserAttribute` succeeds
- return AWS-compatible code delivery details and deliver codes to the pending destination
- preserve immediate-update behavior for contact attributes that do not require verification
- enforce alias availability when a pending value is promoted
- validate every updated attribute before issuing any verification code, so an alias conflict on one attribute can no longer leave an earlier attribute's code live against a user record that was never actually persisted
- treat a blank attribute value as AWS's documented "delete the attribute" instead of rejecting it with `InvalidParameterException`
- omit `UserAttributeUpdateSettings` from `DescribeUserPool`/`CreateUserPool`/`UpdateUserPool` responses when a pool never configured it, instead of returning an empty object

Closes #2021.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

`UpdateUserAttributes` now matches the documented contract for attributes configured to require verification: the request returns `CodeDeliveryDetailsList`, a blank value deletes the attribute rather than erroring, and a pending value is only promoted once `VerifyUserAttribute` succeeds and its alias is still available at that point. `DescribeUserPool`/`CreateUserPool`/`UpdateUserPool` omit `UserAttributeUpdateSettings` entirely for a pool that never configured it, matching how AWS omits absent optional members generally.

This was originally stacked on #2018 (`VerifyUserAttribute`), since promoting a pending value depends on it. #2018 has since merged, and this branch is now rebased directly onto `main`.

## Scope

The PR intentionally does not include:

- self-service validation for direct `email_verified` or `phone_number_verified` writes (#2530)
- cleanup of stale pending values/codes after admin contact changes (#2529)
- general concurrency or transaction changes

## Checklist

- [x] targeted Cognito verification-code and attribute-update test classes passed locally on the original implementation
- [x] full local suite passed on the original implementation, prior to the review-fix commits below
- [x] AWS SDK compatibility test for the pending-verification flow passed locally on the original implementation
- [ ] full suite re-run after the review-fix commits below

Rebased onto current `main`, folding in the orphaned-code and blank-value fixes above since the rebase already had to touch that exact code. Two further commits address the remaining review findings separately: omitting `UserAttributeUpdateSettings` when unconfigured instead of returning an empty object, and replacing the compatibility test's `var` locals with the concrete AWS SDK response types.


hectorvent:
Thanks for the careful pending-verification model, and for filing the follow-up issues before the PRs. Three things, all in the new code:

**(blocking)** Greptile is right about the orphaned codes: the update loop issues and dispatches a code per contact attribute before later entries are validated, so a failing second attribute leaves a live code that VerifyUserAttribute then uses to mark the old stored value verified (no pending value was ever persisted). Validating every entry before issuing any code closes it, and also makes behavior independent of the request's attribute order.

**(blocking)** The handler emits `"UserAttributeUpdateSettings": {}` for pools that never configured it; AWS omits absent optional members ([model](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json)). The neighbouring lines do the same, so this just follows a bad example.

**(blocking)** The model documents a blank value as "delete the attribute" for UpdateUserAttributes, but the new path rejects blank email/phone_number with InvalidParameterException. Was that intentional?

Those three unblock it from my side.

KenkoGeek (changes requested):
**Medium - new Java code violates the repository's explicit-type rule.**

`AGENTS.md` requires explicit Java types and explicitly disallows `var`. `CognitoVerifyUserAttributeTest` introduces multiple inferred locals, including `emailDelivery`, `phoneDelivery`, and the `update` responses.

The Cognito behavior itself is aligned with the official `UserAttributeUpdateSettings.AttributesRequireVerificationBeforeUpdate` contract: when enabled, Cognito keeps the existing email/phone value active until the new value is verified. I did not find a blocker in that core behavior.

Please replace the new `var` declarations with the concrete AWS SDK response types before merge. Also note this branch is still stacked on #2018, which remains open, so it cannot be considered independently merge-ready yet.

pgermosen (approved):
This closes out all three of my blocking points. Validating every attribute before issuing a code fixes the ordering bug cleanly and makes it independent of request order, and treating a blank value as a delete instead of an error matches what the docs actually say. Also swapped the compatibility test's new locals off var per KenkoGeek's note.

Rebased onto current main since quite a bit shifted underneath this branch while it sat open, and CI's green across the board now.

PR #2532, opened 2026-08-24
fix(cognito): reject self-managed verification status

### LINKED ISSUE #2530: [BUG] Cognito UpdateUserAttributes allows clients to set verification status flags

dzwicker:
## Summary

- reject self-service `UpdateUserAttributes` requests that contain `email_verified` or `phone_number_verified`
- return `InvalidParameterException` with HTTP 400 before any attribute from the request is persisted
- preserve the existing administrative ability to set verification status attributes

Closes #2530.

## Scope

This PR only validates the two reserved verification status attributes in the self-service path and adds focused service coverage. Pending-attribute cleanup, concurrency handling, and unrelated Cognito behavior are out of scope.

## Verification

- focused `CognitoServiceTest`
- `./mvnw test`
- `./mvnw clean package -DskipTests`
- `git diff --check`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -1629,7 +1629,17 @@ public class CognitoService {
         validateOriginJtiNotRevoked(accessToken, poolId);
         Long iat = extractIatFromToken(accessToken);
         validateUserNotGloballySignedOut(username, poolId, "access", iat != null ? iat : 0L);
-        
+
+        String verificationStatusAttribute = attributes.containsKey("email_verified")
+                ? "email_verified"
+                : attributes.containsKey("phone_number_verified") ? "phone_number_verified" : null;
+        if (verificationStatusAttribute != null) {
+            throw new AwsException("InvalidParameterException",
+                    "Invalid user attributes: " + verificationStatusAttribute
+                            + ": Attribute cannot be updated.",
+                    400);
+        }
+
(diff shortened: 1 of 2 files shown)


hectorvent:
Thanks, this is a clean, atomic fix that matches the real service's split between self-service and admin writes, and InvalidParameterException is the declared error for it ([model](https://github.com/boto/botocore/blob/develop/botocore/data/cognito-idp/2016-04-18/service-2.json)).

One coordination note rather than a finding: your #2531 stack rewrites updateUserAttributes so it no longer delegates to adminUpdateUserAttributes, and the stack does not currently contain this rejection. When this merges first, please carry the check into the stack's new loop on rebase so the fix does not silently disappear there.

No blockers from my side.

hectorvent (approved):
Approved, see comment above.

PR #2533, opened 2026-08-24
fix(cognito): clear stale pending admin updates

### LINKED ISSUE #2529: Cognito: clear stale pending verification state after admin contact changes

dzwicker:
## Summary

- clear an email or phone number's pending replacement when an admin updates that contact attribute
- clear pending state when an admin deletes the matching contact attribute
- invalidate only the matching attribute-verification code
- preserve unrelated pending contacts and custom attributes
- document the admin/pending-verification interaction in `docs/services/cognito.md`

Closes #2529.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

`AdminUpdateUserAttributes` and `AdminDeleteUserAttributes` write immediately and don't go through the pending-verification flow introduced in #2531. Without this fix, a self-service verification code issued before an admin overwrite or delete stayed valid and could later promote or verify a value the admin had already replaced. Scoped to only the matching attribute (email or phone_number) and its own verification purpose; unrelated pending contacts and custom attributes are untouched.

This was originally stacked on #2531 (the pending-attribute model). #2531 has since merged, and this branch is now rebased directly onto `main`.

## Scope

The production change is limited to the existing admin update/delete service paths. General concurrency, transaction, handler, model, and unrelated Cognito changes are out of scope.

## Checklist

- [x] focused `CognitoServiceTest` cases passed locally on the original implementation
- [x] full local suite passed on the original implementation, prior to the rebase
- [ ] full suite re-run after the rebase

Rebased onto current `main`. The rebase touched `adminUpdateUserAttributes` directly: main had since added the alias-uniqueness check from #1849 in the same spot this PR's cleanup call goes, so the conflict resolution keeps both, alias validation first, then the pending-state cleanup. Also fixed a `CognitoServiceTest` helper that still constructed `CognitoService` with the constructor shape from before #2018's token-verification rework.


hectorvent:
Thanks, the cleanup itself is exactly right: admin writes take effect immediately, so clearing the matching pending value and invalidating only that attribute's code (per purpose) is the correct scope.

**(blocking)** The branch conflicts with main in CognitoService.java in two spots: the alias-uniqueness check from #1849 in adminUpdateUserAttributes (keep both sides), and the verification-status guard from your own #2532 in updateUserAttributes, which the stack's rewrite from #2531 replaced. That second one is the subtle bit: the reworked loop never rejects user-supplied email_verified or phone_number_verified, so when restacking please carry the #2532 rejection into it (its merged test will flag it if missed). The restack can only land after #2018 and #2531 underneath it.

**(follow-up, separate PR)** Greptile's finding on this thread is inherited from the #2531 base diff (codes issued before later attributes validate); it is #2531's to fix, nothing extra needed here.

Rebase plus the stack landing underneath is all this needs.

KenkoGeek (commented):
I reviewed the PR-specific commit separately from the stacked history. I did not find an additional blocker in the admin cleanup itself: clearing only the matching pending contact value and invalidating only that attribute's verification code is consistent with the intended Cognito state transition and is narrowly scoped.

This PR is still not independently merge-ready because it is stacked on #2531, where I requested changes for repository Java style compliance, and #2531 is itself stacked on #2018. Once those dependencies are corrected/merged and this branch is rebased, this one should be rechecked against the resulting head.

pgermosen (approved):
The cleanup logic itself was already right, exactly as you scoped it: matching attribute only, matching purpose only, everything else untouched. Rebase needed both sides kept in adminUpdateUserAttributes since main picked up the #1849 alias check in the same spot, so alias validation runs first and then the pending-state cleanup.

Also added a short paragraph to the Cognito docs explaining that admin writes bypass and clear pending verification, since that wasn't mentioned anywhere. Rebased onto current main and everything's green.

PR #2535, opened 2026-08-24
fix(cognito): serialize attribute verification updates

### LINKED ISSUE #2534: [BUG] Cognito attribute verification races can mismatch codes and pending values
## Service

Cognito IDP

## AWS API Action

- `UpdateUserAttributes`
- `VerifyUserAttribute`
- `AdminUpdateUserAttributes`
- `AdminDeleteUserAttributes`

## Expected behavior

For a single user, contact-attribute update, code resend, and verification operations must have a serializable outcome:

- a verification code only promotes the exact value to which it was delivered;
- a successful update is not silently lost;
- resend targets the pending value associated with the code it makes current;
- completed admin updates and deletes are not overwritten by an older verification request;
- unrelated users remain independently processable.

## Actual behavior

Deterministic concurrency tests reproduce three inconsistent outcomes:

1. Two concurrent email updates can leave a valid code delivered for one address while verifying and promoting the other address.
2. `GetUserAttributeVerificationCode` racing with an update can deliver the latest valid code to the stale pending destination while that code promotes the new destination.
3. `VerifyUserAttribute` racing with a new update can overwrite the successful update with a stale user snapshot and lose its pending value.
4. `VerifyUserAttribute` racing with `AdminUpdateUserAttributes` can overwrite the completed admin update with the formerly pending value.
5. `VerifyUserAttribute` racing with `AdminDeleteUserAttributes` can restore the contact attribute after the completed admin deletion.

This issue depends on the pending-attribute behavior introduced by #2531 and the admin cleanup behavior in #2533.

## Reproduction

```bash
./mvnw test -Dtest=CognitoAttributeVerificationConcurrencyTest
```

The reproducer uses controlled latches around verification-code issue/consume and dispatch. It does not use sleeps or probabilistic stress loops. Before synchronization is added, all five tests fail deterministically with the mismatched, restored, or lost state described above.

## Environment

- Floci version / image tag: current `main` with #2531 and #2533 applied
- Java SDK version (if applicable): not applicable; direct service concurrency reproduction
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Maven/JUnit on Java 25

## Proposed scope

Keep the initial fix JVM-local and limited to Cognito's per-user verification mutation paths. Avoid a global lock and unrelated storage refactors. Cross-process coordination is out of scope and should be documented as a limitation.

dzwicker:
## Summary

- serialize every user-record mutation per user: the original five (`UpdateUserAttributes`, `GetUserAttributeVerificationCode`, `VerifyUserAttribute`, `AdminUpdateUserAttributes`, `AdminDeleteUserAttributes`) plus `AdminDeleteUser`, `AdminCreateUser`, migrated-user creation, `AdminSetUserPassword`, `AdminEnableUser`, `AdminDisableUser`, `AdminResetUserPassword`, `AdminLinkProviderForUser`, `AdminAddUserToGroup`, `AdminRemoveUserFromGroup`, `DeleteGroup`, `DeleteUserPool`, `SignUp`, `ConfirmSignUp`, `AdminConfirmSignUp`, `ChangePassword`, `AdminSetUserMFAPreference`, and `SetUserMFAPreference`
- use a fixed set of JVM-local lock stripes to keep lock memory bounded and allow unrelated users to proceed independently
- add deterministic regression coverage for update/update, resend/update, verify/update, verify/admin-update, verify/admin-delete, verify/admin-user-delete, and verify/change-password interleavings

Closes #2534.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The fix prevents a verification code delivered for one contact value from promoting another value, prevents successful pending updates from being overwritten, and prevents any two of the now roughly two dozen user-mutating operations (contact updates, admin writes, password changes, MFA preferences, group membership, sign-up, and user deletion) from racing each other and silently losing one side's write. Storage here is a plain whole-object get/put/delete with no version or CAS primitive, so a shared per-user lock across every mutating path is what actually closes that, not just the verification-adjacent ones. Operations for one user are serialized within a Floci JVM; cross-process coordination is intentionally out of scope.

`AdminCreateUser` and `SignUp` lock on the requested username or alias rather than a resolved canonical id, since an alias pool has no canonical id until creation succeeds; that's what actually prevents two concurrent requests for the same identifier from both passing the existence check and creating duplicates. `AdminLinkProviderForUser` keeps its existing pool-wide uniqueness lock nested inside the new per-user lock, since a single user's lock can't cover a check that spans every user in the pool. `DeleteGroup` and `DeleteUserPool` lock each affected user individually inside their loops, one at a time, rather than holding multiple users' locks at once.

The seven regression tests use bounded latches and futures around the existing service and verification-code interfaces, without sleeps or probabilistic stress loops.

This was originally stacked on #2533. #2531 and #2533 have since merged, and this branch is now rebased directly onto `main`.

## Checklist

- [x] targeted service and concurrency tests passed locally on the original implementation
- [x] full local suite passed on the original implementation, prior to the rebase and the lock-coverage expansion below
  - Not applicable: the race interleavings require deterministic service-level control of code issue/consume and dispatch; seven focused concurrency tests were added instead.
- [ ] full suite re-run after the rebase and the two locking commits below

Rebased onto current `main`. The rebase conflicted in `CognitoService.java` around the lock-striping fields and the `UpdateUserAttributes` split; resolved by keeping the new locking wrapper and confirming the self-service `email_verified`/`phone_number_verified` rejection (#2532) survived the merge, since it lives in the same method this PR restructures. Also fixed a `CognitoAttributeVerificationConcurrencyTest` helper that still constructed `CognitoService` with the constructor shape from before #2018's token-verification rework.

hectorvent's review flagged that the lock originally covered only the five verification operations, out of roughly two dozen methods that mutate a user record, with `AdminDeleteUser` singled out as the most severe gap (a verification finishing after a concurrent admin deletion could resurrect the user). Both follow-up commits below close that: the first locks `AdminDeleteUser` specifically, the second extends the same lock to every remaining user-mutating method in the file.


hectorvent:
Thanks, the latch based deterministic concurrency tests remain a model for this kind of fix.

**(blocking)** The lock still covers only the five verification operations. At bf7a673 the service has nineteen userStore.put sites and only four run under userLock, and adminDeleteUser deletes without it, so the whole record write back in updateUserAttributes can silently overwrite a concurrent unlocked write or resurrect a deleted user. Either take the user lock in every user record mutation, or re-read and mutate in place inside the locked paths. Greptile's third point is right about why.

**(blocking)** Main now rejects self service email_verified and phone_number_verified writes ([f8597bd](https://github.com/floci-io/floci/commit/f8597bdef91e5538988afb1ac7bec5ab2c90eaa6), #2532). The rewritten loop here writes them straight through, since isVerifiableContactAttribute matches only email and phone_number, so please carry that rejection into the loop when you rebase, otherwise the merge silently reverts #2532.

**(follow-up, separate PR)** Two users pending the same alias still pass ensureAliasAvailable independently; promotion needs a pool level uniqueness guard.

A rebase onto current main (CognitoService.java conflicts) plus the two changes above unblocks this from my side.

KenkoGeek (commented):
I reviewed the concurrency commit separately from the stacked history. The lock-striping approach is bounded, serializes the five affected per-user operations without introducing a global lock, and the tests cover the important update/resend/verify/admin interleavings. I did not find an additional blocker in this commit.

This PR is still blocked by its stack: #2533 depends on #2531, and #2531 currently has requested changes plus an open dependency on #2018. After those are resolved and this branch is rebased, the resulting head needs one final pass before approval.

pgermosen (approved):
Went further than the original five operations here. Storage is a plain whole-object overwrite with no version check, so once I looked past the verification-adjacent methods it was clear the same lock needed to cover every path that reads, mutates, and writes back a user record, not just the ones touching pending attributes. AdminDeleteUser was the sharpest example (a slow verification finishing after a deletion could resurrect the user), but sign-up, password changes, MFA preferences, group membership, and the rest all had the identical shape.

AdminCreateUser and SignUp lock on the requested username rather than a resolved id, since alias pools don't have a canonical one until creation succeeds, that's what actually stops two concurrent requests for the same identifier from both winning. AdminLinkProviderForUser keeps its own pool-wide uniqueness lock nested inside the new per-user one, and the two multi-user operations (DeleteGroup, DeleteUserPool) lock each member individually rather than holding several stripes at once.

Rebased onto current main and CI's green across the board, including the full suite.

PR #2551, opened 2026-08-25
fix(kinesis): reject non-blob Data instead of coercing it with asText

### LINKED ISSUE #2516: [BUG] Kinesis: missing or non-string Data is coerced with asText(), storing empty or garbage records when the text decodes
`PutRecord` and `PutRecords` accept a request whose `Data` is missing entirely, JSON `null`, `true`, a JSON object/array, or a number whose text is legal base64, and store a record for it — `PutRecord` answers 200 with a SequenceNumber, `PutRecords` answers 200 with `FailedRecordCount: 0`. AWS models `Data` as `Required: Yes` with `Type: Base64-encoded binary data object` on `PutRecord` and on each `PutRecords` entry, so none of these shapes is a valid request under that model.

Observed against `main` at 257e0046, `PutRecords`, with a `TRIM_HORIZON` read-back of what actually landed:

| `Data` in the entry | result |
|---|---|
| missing entirely | 200, record written, empty payload |
| JSON object or array | 200, record written, empty payload |
| JSON `null` | 200, record written, payload = 3 garbage bytes (`9e e9 65`) |
| boolean `true` | 200, record written, payload = 3 garbage bytes (`b6 bb 9e`) |
| number whose text is legal base64 | 200, record written, garbage bytes (`1234` → `d7 6d f8`, `123456` → `d7 6d f8 e7`) |
| boolean `false` | 200, per-record `InternalFailure` |
| `12345`, `0`, `-1234`, `12.5` | 200, per-record `InternalFailure` |
| undecodable text | 200, per-record `InternalFailure` |
| valid base64 | 200, record written |

`PutRecord` behaves the same on every row: 200 for the silent ones, `SerializationException` 400 for the failing ones.

Cause: both write paths read the blob with `.path("Data").asText()` (`KinesisJsonHandler.java:453`, and the write-loop fallback at `:516`, which bypasses the pre-pass's `isTextual()` gate at `:486`). `asText()` yields `""` for a missing or container node — a valid empty decode — and the literal text for `null`, booleans and numbers, so whether a non-string `Data` is rejected or silently stored comes down to whether that text happens to be legal base64: `null`, `true`, `1234` and `123456` decode into garbage bytes, while `false`, `0` and `12345` fail (`SerializationException` 400 on `PutRecord` since #1926; per-record `InternalFailure` in the batch loop). An explicit `"Data": ""` is legal on AWS (min length 0), so the empty-record outcome itself is fine; the bug is the silent coercion that produces it from input the AWS model doesn't allow.

Greptile flagged the empty-record half on #2438; @pgermosen drove it, proved it pre-existing on `main`, and suggested it get its own issue. I can send the PR — reject non-textual `Data` at both read sites, routed the same way each path already handles undecodable base64.

soreavis:
## Summary

`PutRecord` and each `PutRecords` entry read the blob with `path("Data").asText()`, which renders a missing or container `Data` as `""`, and `null`, booleans and numbers as their literal text — so whether an invalid request was rejected came down to whether that text happened to be legal base64. A missing `Data` was stored as a successful empty record, JSON `null`, `true` or `1234` landed as a few garbage bytes, and `false` or `12345` happened to fail. Closes #2516.

Both operations now accept exactly the two shapes a real client can send — a base64 text string on the JSON protocol, or a binary node on the CBOR transports, where an unmodified AWS SDK for Java Kinesis client encodes `Data` as a byte string that Jackson decodes to a binary node, not text — and route anything else the way each path already handles undecodable base64: `SerializationException` 400 on `PutRecord`, a per-record `InternalFailure` in the `PutRecords` write loop. One correction to the issue as filed: it proposed rejecting non-textual `Data`, but a textual-only gate turns out to break every CBOR-default SDK client (the pre-existing CBOR Kinesis test builds its `Data` with the emulator's own encoder, which writes it as a text string, so it can't see this), and the shipped rule is reject non-blob. The batch pre-pass keeps a null data for the rejected shapes, so the request cap still counts them at the bytes the caller sent, and an explicit `"Data": ""` remains legal and stores an empty payload as before.

A few smaller behavior shifts ride along, all toward the same rule: `PutRecord` with a non-blob `Data` on an unknown stream now answers `SerializationException` rather than `ResourceNotFoundException` (the malformed request is rejected before the service resolves the stream, as the undecodable-base64 path already did); a `PutRecords` entry that is not a JSON object now fails per-record instead of landing an empty record; the record-size check now runs for every entry, so a malformed entry with an oversized partition key fails the whole request instead of landing the rest of the batch; and a binary `Data` now decodes in the pre-pass, so both size caps see its real decoded length — an oversized CBOR blob fails the whole request as an oversized JSON-encoded record already did, and a batch of CBOR blobs counts toward the request cap at decoded rather than base64-rendered size.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

AWS models `Data` as required Base64-encoded binary data on `PutRecord` and on each `PutRecords` entry, so none of the rejected shapes is a valid request under that model. Base64 text is the JSON protocol's rendering of that blob; a byte string is CBOR's — and CBOR is the AWS SDK for Java's default Kinesis transport, so the gate accepts both. The rejection reuses the routes the emulator already answers for undecodable base64: a 400 on the single path (#1926), a per-record failure in the batch loop, and nothing written for the rejected input.

## Checklist

- [x] `./mvnw test` run locally: 12,016 tests, with failures confined to 23 Docker-backed classes (no Docker socket on this machine) — the identical failing set a sibling branch with a disjoint diff produces here, and none of them is a Kinesis class.
  - The Kinesis-adjacent and CBOR suites are fully green: 326/326 across the Kinesis, KinesisAnalytics, DynamoDB-streaming, Pipes, AwsJsonCbor and SmithyRpcV2 classes, with `KinesisJsonHandlerTest` at 69/69 including eight new tests — every rejected shape on both operations (fault code, message and, on `PutRecord`, the 400 status pinned; read-backs proving nothing landed), the preserved rows (explicit `""` stored, undecodable text still loud), binary-node acceptance on both operations with payloads read back byte-exact, and the size accounting (an oversized partition key on a malformed entry and an oversized binary record both fail the whole request; a batch of binary entries is counted at decoded rather than base64-rendered size). A wire-level test in `AwsJsonCborKinesisAtTimestampIntegrationTest` hand-builds a CBOR `PutRecord` whose `Data` is a real byte string — the encoding an unmodified CBOR-default SDK client sends, deliberately not the emulator's own encoder — and reads the payload back byte-exact.
  - Red-proof: five of the new tests go red against the unfixed handler (the batch one with `FailedRecordCount` 0 instead of 5), and five including the CBOR wire test go red against a text-only `isTextual()` gate, so the suite distinguishes the correct fix from one that would break CBOR clients.
  - Mutation check: twelve mutants across both gates, the write-loop fallback, the size and cap accounting, and the fault code, status and message values — eleven killed by a named test, one equivalent survivor documented (the write-loop's binary acceptance is unreachable, since the pre-pass always decodes a binary `Data` first; the predicate keeps parity with the pre-pass).
- [x] New or updated integration test added — the CBOR wire test above drives the real HTTP endpoint; the handler-level tests cover the malformed shapes (`Data: 1234`, `Data: {}`), which cannot be constructed through a typed AWS SDK client.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -465,9 +465,15 @@ public class KinesisJsonHandler {
 
     private Response handlePutRecord(JsonNode request, String region) {
         String streamName = resolveStreamName(request);
+        JsonNode dataNode = request.path("Data");
+        // The CBOR transports (the AWS SDK for Java's default for Kinesis) decode Data
+        // as a binary node rather than base64 text, so both blob shapes are accepted.
+        if (!dataNode.isTextual() && !dataNode.isBinary()) {
+            throw new AwsException("SerializationException", "Data must be a base64-encoded string.", 400);
+        }
         byte[] data;
         try {
-            data = Base64.getDecoder().decode(request.path("Data").asText());
+            data = Base64.getDecoder().decode(dataNode.asText());
         } catch (IllegalArgumentException e) {
             throw new AwsException("SerializationException", "Data is not valid base64.", 400);
         }
@@ -489,32 +495,32 @@ public class KinesisJsonHandler {
 
         // An oversized record fails the whole request before anything is
         // written — per-record ErrorCode is reserved for throughput/internal
-        // failures. Successful decodes are kept for the write loop; malformed
-        // Data is left null so it stays a per-record failure there. The count
-        // cap is checked upfront and the byte cap as the loop goes, so neither
-        // an over-long batch nor an oversized one is fully decoded before it
-        // is rejected.
+        // failures. Successful decodes are kept for the write loop; a Data
+        // that is not a decodable blob is left null and fails per-record
+        // there, with its partition key still counted by the record-size
+        // check. The count cap is checked upfront and the byte cap as the
+        // loop goes, so neither an over-long batch nor an oversized one is
+        // fully decoded before it is rejected.
         record Entry(JsonNode node, byte[] data) {}
         List<Entry> entries = new ArrayList<>();
         long totalBytes = 0;
         for (JsonNode node : recordsNode) {
             JsonNode dataNode = node.path("Data");
             byte[] data = null;
-            if (dataNode.isTextual()) {
+            if (dataNode.isTextual() || dataNode.isBinary()) {
                 try {
-                    data = Base64.getDecoder().decode(dataNode.textValue());
+                    data = Base64.getDecoder().decode(dataNode.asText());
                 } catch (IllegalArgumentException e) {
                     data = null;
                 }
             }
             String partitionKey = node.path("PartitionKey").asText();
-            if (data != null) {
-                service.validateRecordSize(stream, data, partitionKey);
-            }
+            service.validateRecordSize(stream, data, partitionKey);
             // Data that did not decode still travelled in the request, so it counts toward the
             // request cap at the bytes the caller sent rather than as nothing — otherwise a batch
-            // of undecodable records could carry any payload past the limit. A non-string Data is
-            // measured the same way, since it is not a blob AWS would have accepted either.
+            // of undecodable records could carry any payload past the limit. A Data that is
+            // neither text nor binary is measured the same way, since it is not a blob AWS
+            // would have accepted either.
             totalBytes += KinesisService.recordSize(
                     data != null ? data.length
                             : dataNode.toString().getBytes(StandardCharsets.UTF_8).length,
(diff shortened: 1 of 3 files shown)


hectorvent:
Thank you, and welcome. This is an exceptionally well evidenced first PR, the red proof against both the unfixed handler and a text only gate is exactly the discipline that keeps a fix honest. I verified against [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/kinesis/2013-12-02/service-2.json) that Data is a required blob on PutRecord and on each PutRecords entry, so none of the rejected shapes is a valid request, and your CBOR byte string acceptance matches what a default AWS SDK for Java client actually sends. Reusing the established SerializationException and per record InternalFailure routes keeps the change consistent with how the emulator already treats undecodable base64, and catching that the issue's own textual only proposal would have broken CBOR clients was the important call here. No blockers from my side.

hectorvent (approved):
Approved, see comment above.

PR #2555, opened 2026-08-25
fix(secretsmanager): stop CreateSecret overwriting a secret inside its recovery window

### LINKED ISSUE #2549: [BUG] Secrets Manager: CreateSecret overwrites a secret inside its recovery window (data loss); tag/untag/re-delete also skip the pending-deletion check
## Service
Secrets Manager

## AWS API Action
CreateSecret (data loss), TagResource, UntagResource, DeleteSecret (repeat call) — all against a secret scheduled for deletion

## Expected behavior
`DeleteSecret` without `ForceDeleteWithoutRecovery` schedules a secret for deletion but keeps it recoverable for the recovery window (default 30 days) via `RestoreSecret`. Real AWS rejects these operations on a secret in that state with `InvalidRequestException` — `CreateSecret`, `TagResource`, `UntagResource`, and `DeleteSecret` all declare it in their error lists. For `CreateSecret` specifically, that rejection is what makes the recovery window mean anything: you can't reuse a scheduled name until the window elapses or you restore it.

## Actual behavior
All four return 200 and skip the pending-deletion check entirely. **`CreateSecret` is the serious one — it silently destroys the recoverable secret:**

```bash
aws secretsmanager create-secret --name s2 --secret-string ORIGINAL-VALUE
aws secretsmanager delete-secret --secret-id s2          # 30-day recovery window
aws secretsmanager create-secret --name s2 --secret-string NEW-VALUE   # -> 200 (AWS: InvalidRequestException)
aws secretsmanager restore-secret --secret-id s2
# -> InvalidRequestException: You can't perform this operation on the secret because it was not deleted.
aws secretsmanager get-secret-value --secret-id s2
# -> "NEW-VALUE"    ORIGINAL-VALUE is unrecoverable
```

`createSecret`'s guard is `existing != null && existing.getDeletedDate() == null`, so a name scheduled for deletion falls straight through and overwrites the recoverable secret at the same storage key — which also clears `deletedDate`, so `RestoreSecret` then reports "was not deleted" and there is no way back to the original value.

The other three are wrong-error-code rather than data-loss, probed on the same pending-deletion secret:

```
tag-resource           -> 200   (AWS: InvalidRequestException)
untag-resource         -> 200   (AWS: InvalidRequestException)
delete-secret (again)  -> 200, silently re-schedules with a fresh DeletionDate   (AWS: InvalidRequestException)

describe-secret        -> 200   (correct — AWS allows this, listed for contrast)
```

## Relationship to #2513 / #2514

#2513 (fixed by #2514) covered six operations that raised the **wrong** error code (`ResourceNotFoundException` instead of `InvalidRequestException`) on a pending-deletion secret. These four are a distinct set: they don't raise a wrong code, they don't check at all, which is why they didn't surface in the same search. All pre-existing; none introduced by #2514.

Credit to @pgermosen, who found these while reviewing #2514 and suggested filing them separately so the data-loss case is findable on its own. I independently reproduced all four against `floci/floci:1.7.0-compat` before filing.

The fix should be mechanical for the last three — they can reuse the `throwIfPendingDeletion(Secret)` helper #2514 adds. `CreateSecret` needs its own handling, since the correct behavior there is to reject the create rather than guard an existing-secret operation, and it deserves a dedicated regression test asserting the original value survives.

## Reproduction
```bash
docker run -d --name floci-repro -p 4566:4566 -e SERVICES=secretsmanager floci/floci:1.7.0-compat

# Data loss
aws --endpoint-url http://localhost:4566 secretsmanager create-secret --name s2 --secret-string ORIGINAL-VALUE
aws --endpoint-url http://localhost:4566 secretsmanager delete-secret --secret-id s2
aws --endpoint-url http://localhost:4566 secretsmanager create-secret --name s2 --secret-string NEW-VALUE
aws --endpoint-url http://localhost:4566 secretsmanager restore-secret --secret-id s2
aws --endpoint-url http://localhost:4566 secretsmanager get-secret-value --secret-id s2   # NEW-VALUE

# Missing checks
aws --endpoint-url http://localhost:4566 secretsmanager create-secret --name s3 --secret-string V1
aws --endpoint-url http://localhost:4566 secretsmanager delete-secret --secret-id s3
aws --endpoint-url http://localhost:4566 secretsmanager tag-resource --secret-id s3 --tags Key=k,Value=v   # 200
aws --endpoint-url http://localhost:4566 secretsmanager untag-resource --secret-id s3 --tag-keys k         # 200
aws --endpoint-url http://localhost:4566 secretsmanager delete-secret --secret-id s3                       # 200
```

## Environment
- Floci version / image tag: `floci/floci:1.7.0-compat`
- Java SDK version (if applicable): AWS CLI v2, not the Java SDK

dedsecrattle:
## Summary

Fixes #2549

Rebased onto `main` now that #2514 has merged, so this is down to just the four new guards. Follow-up to #2514, filed at @pgermosen's suggestion in [review](https://github.com/floci-io/floci/pull/2514#issuecomment-5402247305). #2514 fixed six operations that raised the **wrong** error code on a pending-deletion secret. These four are a distinct set: they don't check pending-deletion state **at all**, which is why they didn't surface in the same search. All pre-existing; none introduced by #2514.

**The `CreateSecret` case loses data.** Its "already exists" guard was `existing != null && existing.getDeletedDate() == null`, so a name scheduled for deletion fell straight through and overwrote the recoverable secret at the same storage key. That also cleared `deletedDate`, so the subsequent `RestoreSecret` reported *"was not deleted"* — the recovery path doesn't just come back empty, it actively claims there's nothing to recover, which makes the loss easy to miss.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior:** four operations against a secret inside its recovery window returned 200 instead of `InvalidRequestException`. `CreateSecret`, `TagResource`, `UntagResource`, and `DeleteSecret` all declare `InvalidRequestException` in their error lists in `secretsmanager/2017-10-17/service-2.json`, with "The secret is scheduled for deletion" listed among the documented causes.

Before — the data-loss path:

```bash
create-secret --name s2 --secret-string ORIGINAL-VALUE
delete-secret --secret-id s2                          # 30-day recovery window
create-secret --name s2 --secret-string NEW-VALUE     → 200          (AWS: InvalidRequestException)
restore-secret --secret-id s2                         → InvalidRequestException "was not deleted"
get-secret-value --secret-id s2                       → "NEW-VALUE"  ORIGINAL-VALUE unrecoverable
```

After, verified with AWS CLI v2 against a locally built JVM jar:

```
create-secret (scheduled name)  → InvalidRequestException
tag-resource                    → InvalidRequestException
untag-resource                  → InvalidRequestException
delete-secret (repeat, no force)→ InvalidRequestException

restore-secret → 200, and get-secret-value returns ORIGINAL-VALUE
```

Deliberately still allowed, all CLI-confirmed after the change:

```
describe-secret                      → 200   (how a caller reads DeletedDate; AWS doesn't declare IRE)
restore-secret                       → 200
delete-secret --force-delete-…       → 200   (documented "skip the window" escape hatch)
```

**One judgement call worth flagging:** `deleteSecret`'s guard sits *after* the force-delete branch, so force-deleting an already-scheduled secret still works. I read that as a legitimate escape hatch rather than something to refuse, and the reviewer's probe only covered the plain repeat delete. If real AWS actually rejects force-delete-on-scheduled too, say so and I'll move the guard above the branch — `forceDeleteStillWorksOnAPendingDeletionSecret` pins the current placement either way.

`ListSecretVersionIds` left alone for the same reason as `DescribeSecret`.

## Fix

`CreateSecret` gets the check inside its existing-secret branch, before the `ResourceExistsException` — a scheduled name is still taken, so the create is refused rather than allowed to overwrite. The other three reuse `throwIfPendingDeletion` directly.

## Testing

- `createSecretOnPendingDeletionNameIsRefusedAndLeavesOriginalRecoverable` — the data-loss regression. Asserts the create is refused **and** that after `RestoreSecret` the value is still `ORIGINAL-VALUE`, so it fails if the overwrite ever returns.
- `tagUntagAndRescheduleOnPendingDeletionSecretRaiseInvalidRequest` — the other three.
- `forceDeleteStillWorksOnAPendingDeletionSecret` — pins the guard placement so a later "tidy-up" can't accidentally block force delete.
- `describeSecretRemainsAllowedOnPendingDeletionSecret` — pins that the guard wasn't over-applied.
- Confirmed the data-loss test actually catches the bug: reverting just `createSecret`'s guard fails it with *"Expected AwsException to be thrown, but nothing was thrown"*, and nothing else in the suite fails — so it's the test carrying that invariant, not a side effect of another assertion.
- Full `io.github.hectorvent.floci.services.secretsmanager.*Test` package: 122 tests, 0 failures — re-run after the rebase onto current `main`.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerService.java
@@ -104,7 +104,13 @@ public class SecretsManagerService implements ResourceProvider {
         String storageKey = regionKey(region, name);
         Secret existing = store.get(storageKey).orElse(null);
 
-        if (existing != null && existing.getDeletedDate() == null) {
+        if (existing != null) {
+            // A name inside its recovery window is still taken: the secret is recoverable via
+            // RestoreSecret, so reusing the name has to be refused rather than allowed to
+            // overwrite it. Without this the create would replace the stored secret at the same
+            // key AND clear its deletedDate, so the original value became unrecoverable and
+            // RestoreSecret then reported "was not deleted" - silent data loss.
+            throwIfPendingDeletion(existing);
             throw new AwsException("ResourceExistsException",
                     "A secret with the name " + name + " already exists.", 400);
         }
@@ -549,6 +555,13 @@ public class SecretsManagerService implements ResourceProvider {
             return secret;
         }
 
+        // Guard placed AFTER the force-delete branch on purpose: force-deleting a secret that is
+        // already inside its recovery window is a legitimate "skip the window, remove it now"
+        // escape hatch, so only the scheduling path is refused. Re-scheduling an already-scheduled
+        // secret silently moved its DeletionDate, which would quietly extend a window a caller
+        // believed was already counting down.
+        throwIfPendingDeletion(secret);
+
         int windowDays = (recoveryWindowInDays != null) ? recoveryWindowInDays : defaultRecoveryWindowDays;
         Instant deletedDate = Instant.now().plusSeconds((long) windowDays * 86400);
         secret.setDeletedDate(deletedDate);
@@ -739,6 +752,7 @@ public class SecretsManagerService implements ResourceProvider {
 
     public void tagResource(String secretId, List<Secret.Tag> tags, String region) {
         Secret secret = resolveSecret(secretId, region);
+        throwIfPendingDeletion(secret);
 
         List<Secret.Tag> existing = secret.getTags() != null ? new ArrayList<>(secret.getTags()) : new ArrayList<>();
         for (Secret.Tag newTag : tags) {
@@ -751,6 +765,7 @@ public class SecretsManagerService implements ResourceProvider {
 
     public void untagResource(String secretId, List<String> tagKeys, String region) {
         Secret secret = resolveSecret(secretId, region);
+        throwIfPendingDeletion(secret);
 
         List<Secret.Tag> existing = secret.getTags() != null ? new ArrayList<>(secret.getTags()) : new ArrayList<>();
         existing.removeIf(t -> tagKeys.contains(t.key()));
--- a/src/test/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/secretsmanager/SecretsManagerServiceTest.java
@@ -281,6 +281,70 @@ class SecretsManagerServiceTest {
                 service.updateSecretVersionStage("my-secret", null, null, "AWSCURRENT", REGION)).getErrorCode());
     }
 
+    /**
+     * The data-loss case from #2549: CreateSecret against a name inside its recovery window used
+     * to fall through the "already exists" guard (which only fired when {@code deletedDate} was
+     * null), overwrite the recoverable secret at the same storage key, and clear its
+     * {@code deletedDate} as a side effect. The original value became unrecoverable and
+     * RestoreSecret then reported "was not deleted", so nothing surfaced the loss.
+     */
+    @Test
+    void createSecretOnPendingDeletionNameIsRefusedAndLeavesOriginalRecoverable() {
+        service.createSecret("my-secret", "ORIGINAL-VALUE", null, null, null, null, REGION);
+        service.deleteSecret("my-secret", 7, false, REGION);
+
+        assertEquals("InvalidRequestException", assertThrows(AwsException.class, () ->
+                service.createSecret("my-secret", "NEW-VALUE", null, null, null, null, REGION)).getErrorCode());
+
+        // The recovery window still means what it says: restore works and returns the original.
+        service.restoreSecret("my-secret", REGION);
+        assertEquals("ORIGINAL-VALUE",
+                service.getSecretValue("my-secret", null, null, REGION).getSecretString());
+    }
+
+    @Test
+    void tagUntagAndRescheduleOnPendingDeletionSecretRaiseInvalidRequest() {
+        service.createSecret("my-secret", "value", null, null, null, null, REGION);
+        service.deleteSecret("my-secret", 7, false, REGION);
+
+        assertEquals("InvalidRequestException", assertThrows(AwsException.class, () ->
+                service.tagResource("my-secret", List.of(new Secret.Tag("k", "v")), REGION)).getErrorCode());
+        assertEquals("InvalidRequestException", assertThrows(AwsException.class, () ->
+                service.untagResource("my-secret", List.of("k"), REGION)).getErrorCode());
+        // A second non-force DeleteSecret used to silently move DeletionDate forward.
+        assertEquals("InvalidRequestException", assertThrows(AwsException.class, () ->
+                service.deleteSecret("my-secret", 7, false, REGION)).getErrorCode());
+    }
+
+    /**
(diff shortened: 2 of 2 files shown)


hectorvent:
Thanks, this closes the nastiest of the #2514 follow-ups, and the data loss test that re-verifies ORIGINAL-VALUE after restore is exactly the right pin.

**(blocking)** The branch now conflicts with main: #2514 merged, so the stacked first commit collides with its own merged hunks. Rebasing onto main (dropping the first commit) should shrink this PR to just the four new guards, and it also lets CI run, which the conflict currently blocks entirely.

**(follow-up, separate PR)** Real AWS uses a create specific message here, "You can't create this secret because a secret with this name is already scheduled for deletion", rather than the generic marked for deletion text (see [localstack's capture](https://github.com/localstack/localstack/blob/main/localstack-core/localstack/services/secretsmanager/provider.py#L82)). Cosmetic, since clients match on the code.

All four operations declare InvalidRequestException in [service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/secretsmanager/2017-10-17/service-2.json), and the guard ordering matches AWS. The rebase is the only blocker.

pgermosen (approved):
Thanks for the fast follow-up to #2514 — and for scoping it tightly to the four operations that genuinely don't check pending-deletion state, rather than re-touching the six #2514 already fixed.

Checked every claim in the description against the current Secrets Manager API reference rather than taking the PR body's word for it:

| claim | verified |
|---|---|
| `CreateSecret`, `TagResource`, `UntagResource`, `DeleteSecret` all declare `InvalidRequestException` with "The secret is scheduled for deletion" as a listed cause | yes, on all four, word for word |
| `DescribeSecret` does not declare `InvalidRequestException` — correctly left alone | confirmed, its error list is `InternalServiceError` / `InvalidParameterException` / `ResourceNotFoundException` only |
| `ListSecretVersionIds` does not declare `InvalidRequestException` — correctly left alone | confirmed, same reasoning as `DescribeSecret` |
| Force-delete should still work on an already-scheduled secret | confirmed independently — the `DeleteSecret` doc states "If you forcibly delete an already deleted or nonexistent secret, the operation does not return `ResourceNotFoundException`," which only makes sense if force-delete tolerates a secret already in a deleted/deleting state. Your guard placement (after the force-delete branch) matches that. |

The `CreateSecret` fix is the one that actually matters here — the described data-loss path (overwrite clears `deletedDate`, so `RestoreSecret` then reports "was not deleted") is real and would be nasty to debug from the outside, since nothing about it looks like an error at write time. Guard order is right: `throwIfPendingDeletion` runs before `ResourceExistsException` inside the existing-secret branch, so a live (non-deleted) name keeps its current behavior and only the pending-deletion case changes.

Test coverage matches the fix one-to-one: the data-loss regression test asserts both the refusal *and* that `RestoreSecret` still returns the original value afterward, the force-delete test pins the guard's placement so a future cleanup can't accidentally move it above the branch, and the `DescribeSecret` test pins that the guard wasn't over-applied. That's the right set of tests for a fix in this shape — it protects the bug fix and the two adjacent behaviors that would be easy to break while "fixing" it further.

CI is green across every shard, the native build, and all SDK/IaC compat jobs, and the PR is cleanly mergeable, so nothing further to check.

No blockers, no nits.

Recommendation: **approve**.

PR #2572, opened 2026-08-25
fix(msk): return BadRequestException when a configuration ARN does not exist

matanbaruch:
Destroying an `aws_msk_configuration` with Terraform or Pulumi can never complete against floci. Create works, destroy always breaks in the provider's delete waiter.

The provider only treats a configuration as gone when `DescribeConfiguration` fails with `BadRequestException` and a message containing `Configuration ARN does not exist`. That is what real MSK returns for a missing configuration ARN (unlike clusters, where it really is `NotFoundException`). floci returns `NotFoundException` 404, which the provider does not recognize as not-found, so `waitConfigurationDeleted` surfaces it as a raw error and the destroy fails every time.

Provider source, [`internal/service/kafka/configuration.go`](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/kafka/configuration.go):

```go
if errs.IsAErrorMessageContains[*types.BadRequestException](err, "Configuration ARN does not exist") {
    return nil, &retry.NotFoundError{ ... }
}
```

The same match exists back to v4.x (`FindConfigurationByARN` in `find.go`), which pulumi-aws v5 bridges. On current main the provider also tolerates this error on the `DeleteConfiguration` call itself.

Change:
1. `MskService#describeConfiguration` now throws `BadRequestException` 400 with `Configuration ARN does not exist: <arn>` for an unknown ARN. All configuration operations resolve ARNs through it (describe, update, delete, list/describe revisions), so they all share the corrected shape.
2. Cluster APIs untouched.

Tests:
1. `configurationCrudRoundTrip` now asserts the full wire contract after delete: 400, `X-Amzn-Errortype: BadRequestException`, message contains `Configuration ARN does not exist`. That substring is exactly what the provider matches on, so this is the regression test for the destroy flow.
2. Unknown-ARN describe/update integration tests updated from 404 to the new shape.
3. Service-level regression for describe-after-delete.

Verified with `mvn test -Dtest='MskServiceTest,MskControllerTest,MskControllerIntegrationTest'` on JDK 25.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/msk/MskService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/msk/MskService.java
@@ -157,8 +157,13 @@ public class MskService {
     }
 
     public MskConfiguration describeConfiguration(String arn) {
+        // Real MSK reports an unknown configuration ARN as BadRequestException (400), not
+        // NotFoundException. The message prefix is a compatibility contract: terraform-provider-aws
+        // (and the Pulumi provider bridging it) substring-matches "Configuration ARN does not exist"
+        // to detect that a configuration is gone, so its post-delete waiter can complete.
         return configurationStorage.get(arn)
-                .orElseThrow(() -> new AwsException("NotFoundException", "Configuration not found: " + arn, 404));
+                .orElseThrow(() -> new AwsException("BadRequestException",
+                        "Configuration ARN does not exist: " + arn, 400));
     }
 
     public PaginatedResult<MskConfiguration> listConfigurations(Integer maxResults, String nextToken) {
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Thanks for a sharply diagnosed fix. Pinning the terraform waiter's exact match (code plus message substring) in the integration test is exactly right, that substring is the compatibility contract.

Verified against the model and the provider source: DescribeConfiguration declares BadRequestException ([service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/kafka/2018-11-14/service-2.json)), every configuration operation that resolves through it also declares that code, and [configuration.go](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/kafka/configuration.go) matches BadRequestException with "Configuration ARN does not exist" in both the delete call and the waiter. Cluster APIs staying on NotFoundException also matches the model. Copilot's two test suggestions are right about why, and your second commit already covers both.

No blockers from my side.

PR #2576, opened 2026-08-25
feat(cloudwatch-logs): add AssociateKmsKey support

### LINKED ISSUE #2540: CloudWatch Logs: AssociateKmsKey returns UnsupportedOperation
`AssociateKmsKey` is not implemented; it returns `UnsupportedOperation`. This
breaks `aws_cloudwatch_log_group` with `kms_key_id` set in
`terraform-provider-aws` (the provider issues `AssociateKmsKey` after create).

**Repro**
```bash
aws --endpoint-url http://localhost:4566 logs create-log-group --log-group-name /test
KEY=$(aws --endpoint-url http://localhost:4566 kms create-key --query KeyMetadata.KeyArn --output text)
aws --endpoint-url http://localhost:4566 logs associate-kms-key \
  --log-group-name /test --key-id "$KEY"
```

**Expected**: key is associated (stored on the group, returned by `describe-log-groups`).
**Actual**: `UnsupportedOperation: Operation AssociateKmsKey is not supported`.

<details>
<summary>Debug info</summary>

- floci 1.7.0, image `floci/floci:latest`, plain `docker run -p 4566:4566`
- Terraform error: `associating CloudWatch Logs Log Group (...) KMS key:
  operation error CloudWatch Logs: AssociateKmsKey, ... UnsupportedOperation`.
- Terraform v1.5.7, aws provider ~> 5.0.
- Accepting `kmsKeyId` on `CreateLogGroup` alone would not fix the Terraform
  path; the associate call after create must succeed.
</details>

aniketshukla1:
## Summary

Add CloudWatch Logs `AssociateKmsKey` handling, persist the KMS key on the target log group, and return it from log-group descriptions. The generated supported-actions documentation is updated with the new action.

Fixes #2540

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The CloudWatch Logs JSON 1.1 request and response path is covered by focused handler tests. External AWS SDK and CLI smoke testing was not performed.

## Checklist

Focused verification: `CloudWatchLogsHandlerTest` (27 tests) and `make docs-check` pass locally.

### DIFF
--- a/docs/services/cloudwatch.md
+++ b/docs/services/cloudwatch.md
@@ -17,6 +17,7 @@ Floci supports both CloudWatch Logs and CloudWatch Metrics.
 | `DeleteLogGroup` | Delete a log group |
 | `PutLogGroupDeletionProtection` | Enable or disable deletion protection for a log group by name or ARN |
 | `DescribeLogGroups` | List log groups |
+| `AssociateKmsKey` | Associate a KMS key with a log group |
 | `CreateLogStream` | Create a log stream inside a log group |
 | `DeleteLogStream` | Delete a log stream |
 | `DescribeLogStreams` | List log streams in a group |
--- a/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsHandler.java
@@ -46,6 +46,7 @@ public class CloudWatchLogsHandler {
             case "PutRetentionPolicy" -> handlePutRetentionPolicy(request, region);
             case "DeleteRetentionPolicy" -> handleDeleteRetentionPolicy(request, region);
             case "PutLogGroupDeletionProtection" -> handlePutLogGroupDeletionProtection(request, region);
+            case "AssociateKmsKey" -> handleAssociateKmsKey(request, region);
             case "TagLogGroup" -> handleTagLogGroup(request, region);
             case "UntagLogGroup" -> handleUntagLogGroup(request, region);
             case "ListTagsLogGroup" -> handleListTagsLogGroup(request, region);
@@ -95,6 +96,19 @@ public class CloudWatchLogsHandler {
         return Response.ok(objectMapper.createObjectNode()).build();
     }
 
+    private Response handleAssociateKmsKey(JsonNode request, String region) {
+        String groupName = request.path("logGroupName").asText(null);
+        if (groupName == null || groupName.isBlank()) {
+            throw new AwsException("InvalidParameterException", "logGroupName is required.", 400);
+        }
+        String kmsKeyId = request.path("kmsKeyId").asText(null);
+        if (kmsKeyId == null || kmsKeyId.isBlank()) {
+            throw new AwsException("InvalidParameterException", "kmsKeyId is required.", 400);
+        }
+        logsService.associateKmsKey(groupName, kmsKeyId, region);
+        return Response.ok(objectMapper.createObjectNode()).build();
+    }
+
     private Response handleDescribeLogGroups(JsonNode request, String region) {
         String prefix = request.path("logGroupNamePrefix").asText(null);
         List<LogGroup> groups = logsService.describeLogGroups(prefix, region);
@@ -108,6 +122,9 @@ public class CloudWatchLogsHandler {
             node.put("arn", logsService.buildArn(g.getLogGroupName(), region));
             if (g.getRetentionInDays() != null) {
                 node.put("retentionInDays", g.getRetentionInDays());
+            }
+            if (g.getKmsKeyId() != null) {
+                node.put("kmsKeyId", g.getKmsKeyId());
             }
             node.put("deletionProtectionEnabled", g.isDeletionProtectionEnabled());
             node.put("storedBytes", 0);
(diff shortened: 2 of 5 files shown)


sam-fakhreddine:
Heads-up on an overlap: #2584 (opened the same day) also implements KMS key association for log groups, and the two PRs edit the same dispatch switch in `CloudWatchLogsHandler`, so whichever merges second will conflict.

For the maintainer weighing the two — #2584 currently covers a superset of this PR:

- **Both operations**: `AssociateKmsKey` and `DisassociateKmsKey` (botocore models them as a pair; disassociate keeps the key identity retrievable per the model).
- **`resourceIdentifier` support**: the 2014-03-28 model accepts `resourceIdentifier` as the alternative to `logGroupName` on both operations, with exactly-one-of semantics — #2584 validates that (mutual exclusivity was hardened during bot review with the model as evidence).
- **`kmsKeyId` echoed in `DescribeLogGroups`** — both PRs do this identically.
- Wire-level tests for the missing/both-identifier 400 paths.

No knock on this implementation — it is clean and the `LogGroup` field + describe echo match what #2584 does. Just flagging so the effort does not get spent twice. Happy to defer to the maintainer on how to reconcile.

pgermosen (commented):
Thanks Aniket — the storage and read-back half is right: `kmsKeyId` is persisted on the log
group, `DescribeLogGroups` emits it only when it's set (AWS omits the member on an
unencrypted group rather than sending null), and `ResourceNotFoundException` /
`InvalidParameterException` at 400 are both the modelled errors at the modelled status.

Two things before the findings.

**The branch needs a rebase.** GitHub reports `mergeable: CONFLICTING` /
`mergeStateStatus: DIRTY`, so it won't merge onto current `main` as-is. Worth knowing that
`./check-pr.sh` exits at the merge step before it ever builds, so neither I nor CI can give
you a real test result until that's resolved — please merge or rebase `origin/main` and push.

**No build result.** Between the conflict and the unticked `./mvnw test` box, everything below
is a read of the diff against the API reference, not a verified run.

## What's missing

### (should fix) `DisassociateKmsKey` — the key can be set but never removed

`AssociateKmsKey` has a counterpart, and without it there's no request that returns a log group
to unencrypted. The reference frames the association as replaceable —

> Associating a KMS key with a log group overrides any existing associations between the log
> group and a KMS key.

— but *overriding* only covers swapping one key for another. Removal is `DisassociateKmsKey`,
and that's exactly what `aws_cloudwatch_log_group` calls when `kms_key_id` is deleted from a
config. Today that call lands in the handler's default branch, so Terraform can attach a key
and then never detach it — the group is stuck encrypted in state forever.

It's the same shape as the association you just wrote: clear the field, `groupStore.put`. Given
#2540 is about `kmsKeyId` round-tripping, I'd rather see the pair land together than have the
removal path filed as a follow-up.

### (should fix) `logGroupName` isn't actually required — `resourceIdentifier` is the other half

`handleAssociateKmsKey` rejects any request without `logGroupName`. The model says otherwise:

> ** logGroupName ** — … You must specify either the `resourceIdentifier` parameter or the
> `logGroup` parameter, but you can't specify both. **Required: No**
>
> ** resourceIdentifier ** — Specify `arn:aws:logs:REGION:ACCOUNT_ID:query-result:*` … or
> `arn:aws:logs:REGION:ACCOUNT_ID:log-group:LOG_GROUP_NAME`. **Required: No**

So a caller targeting a log group *by ARN* — or setting the account-wide query-results key,
which is the second sample request on that page — gets `InvalidParameterException` from Floci
for a request AWS accepts. The mirror case is missing too: AWS rejects a request carrying
*both*, and Floci silently prefers `logGroupName`.

The `query-result` form is a genuinely different feature and I wouldn't expect it here. But the
`log-group:NAME` form of `resourceIdentifier` is the same operation by a different handle, and
either supporting it or rejecting it with the both-specified error is more honest than treating
its absence as a missing `logGroupName`.

### (should fix) `CreateLogGroup` still drops `kmsKeyId`

`handleCreateLogGroup` reads `logGroupName` and `retentionInDays` and nothing else, but
`CreateLogGroup` has a modelled `kmsKeyId` parameter — it's the normal way to create an
encrypted group in one call, and it's what `aws_cloudwatch_log_group` sends when `kms_key_id`
is set at creation rather than added later.

So after this PR:

```
create-log-group --log-group-name g --kms-key-id arn:aws:kms:…:key/abc   → 200
describe-log-groups --log-group-name-prefix g                            → no kmsKeyId
```

Accepted, stored nowhere, absent from the read-back — which is the exact failure mode you fixed
in #2577 for S3 POST events, and the one #2540 is complaining about. Now that `LogGroup` carries
the field, wiring it into create is two lines and closes the loop rather than leaving one of the
two ways in still broken.

## Notes

- `kmsKeyId` has a documented max length of 256 and is meant to be a symmetric KMS key ARN;
  the handler accepts any non-blank string. Length is the cheap half if you want it, but Floci
  can't check symmetry or key existence, so I wouldn't go further than that. (AWS returns
  `InvalidParameterException` when the key doesn't exist or is disabled — worth knowing, not
  worth emulating here.)
- `OperationAbortedException` and `ServiceUnavailableException` round out the modelled error
  list; neither is reachable in an emulator without concurrent-update detection, so nothing to
  do — noting it so the two you do use read as a deliberate subset.
- The docs table entry lands in the right place and matches the generated format.
- The test asserts the round-trip through `DescribeLogGroups` rather than poking the service
  directly, which is the right level — it's the read path the issue is about.

The association itself is correct. What I'd want alongside it is the removal path and the
create-time parameter, so the feature is usable from both directions rather than one-way.

aniketshukla1:
Thanks for the detailed review and for flagging the overlap. After comparing the current branches, #2584 already covers the issue more completely: associate/disassociate, `resourceIdentifier`, exactly-one-of validation, `kmsKeyId` read-back, and broader wire/service tests. I am closing this duplicate rather than creating competing implementations and a larger merge conflict.

PR #2577, opened 2026-08-25
fix(s3): emit post events for browser uploads

### LINKED ISSUE #2399: S3 notifications: POST-created objects emit ObjectCreated:Put, so s3:ObjectCreated:Post filters never match
On 1.7.0, S3 event notifications label objects created via the browser `POST` (multipart form upload) as `eventName: ObjectCreated:Put`, and event-type filters are matched strictly. A notification configuration filtering on `s3:ObjectCreated:Post` therefore never fires for any upload method — real S3 emits `s3:ObjectCreated:Post` for POST uploads, so configurations synthesized by CDK for browser-upload pipelines silently stop delivering events after upgrading (earlier releases did not enforce the event-type filter, which masked this).

Repro:
1. Create a bucket and an SQS queue; attach a notification configuration with `<Event>s3:ObjectCreated:Post</Event>`.
2. Upload an object via a multipart form POST to the bucket root.
3. ReceiveMessage on the queue → no message. With `<Event>s3:ObjectCreated:*</Event>` the event arrives, but its `eventName` is `ObjectCreated:Put`.

Expected: POST-created objects emit `ObjectCreated:Post` so Post-filtered configurations match, as in real S3.

aniketshukla1:
## Summary

Preserve the HTTP upload method through S3 object creation so browser-based presigned POST uploads emit `ObjectCreated:Post` notifications instead of `ObjectCreated:Put`. Adds integration coverage for POST-specific and wildcard event filters.

Fixes #2399

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Objects created through presigned browser POST uploads were mislabeled as PUT events, preventing `s3:ObjectCreated:Post` filters from matching. Notifications now reflect the originating HTTP method.

## Checklist

Focused verification: `S3PresignedPostIntegrationTest` (20 tests) passes locally.

### DIFF
--- a/docs/services/s3.md
+++ b/docs/services/s3.md
@@ -28,6 +28,11 @@
 
 **RestoreObject** is accepted but stubbed: Floci validates the request and returns `202 Accepted`, but no restore state machine runs.
 
+## Event Notifications
+
+Browser and presigned POST uploads emit `s3:ObjectCreated:Post`, matching AWS S3. They do not emit
+`s3:ObjectCreated:Put`; use `s3:ObjectCreated:*` to subscribe to objects created by either method.
+
 ## Website Hosting
 
 Static website requests use a bucket website hostname such as
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -2698,7 +2698,7 @@ public class S3Controller {
             }
         }
 
-        S3Object obj = s3Service.putObject(bucket, key, fileData, objectContentType,
+        S3Object obj = s3Service.postObject(bucket, key, fileData, objectContentType,
                 metadata.isEmpty() ? null : metadata);
         LOG.infov("Presigned POST upload: {0}/{1} ({2} bytes)", bucket, key, fileData.length);
 
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Service.java
@@ -343,8 +343,24 @@ public class S3Service implements Resettable, ResourceProvider {
 
     public S3Object putObject(String bucketName, String key, byte[] data,
                               String contentType, Map<String, String> metadata, PutObjectOptions options) {
+        return createObject(bucketName, key, data, contentType, metadata, options, "ObjectCreated:Put");
+    }
+
+    public S3Object postObject(String bucketName, String key, byte[] data,
+                               String contentType, Map<String, String> metadata) {
+        return postObject(bucketName, key, data, contentType, metadata, new PutObjectOptions());
+    }
+
+    public S3Object postObject(String bucketName, String key, byte[] data,
+                               String contentType, Map<String, String> metadata, PutObjectOptions options) {
+        return createObject(bucketName, key, data, contentType, metadata, options, "ObjectCreated:Post");
+    }
+
+    private S3Object createObject(String bucketName, String key, byte[] data,
+                                  String contentType, Map<String, String> metadata,
+                                  PutObjectOptions options, String eventName) {
         S3Object object = storeObject(bucketName, key, data, contentType, metadata, null, null, options);
-        fireNotifications(bucketName, key, "ObjectCreated:Put", object);
+        fireNotifications(bucketName, key, eventName, object);
         return object;
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/s3/S3PresignedPostIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/s3/S3PresignedPostIntegrationTest.java
@@ -70,6 +70,113 @@ class S3PresignedPostIntegrationTest {
     }
 
     @Test
+    @Order(15)
+    void presignedPostEmitsPostNotification() {
+        String postQueueName = "presigned-post-notification-queue";
+        String putQueueName = "presigned-put-notification-queue";
+        String wildcardQueueName = "presigned-wildcard-notification-queue";
+        String postQueueUrl = createQueue(postQueueName);
+        String putQueueUrl = createQueue(putQueueName);
+        String wildcardQueueUrl = createQueue(wildcardQueueName);
+
+        try {
+            given()
+                .contentType("application/xml")
+                .queryParam("notification", "")
+                .body("""
+                    <NotificationConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+                        <QueueConfiguration>
+                            <Id>post-notification</Id>
+                            <Queue>arn:aws:sqs:us-east-1:000000000000:%s</Queue>
+                            <Event>s3:ObjectCreated:Post</Event>
+                        </QueueConfiguration>
+                        <QueueConfiguration>
+                            <Id>put-notification</Id>
+                            <Queue>arn:aws:sqs:us-east-1:000000000000:%s</Queue>
+                            <Event>s3:ObjectCreated:Put</Event>
+                        </QueueConfiguration>
+                        <QueueConfiguration>
+                            <Id>wildcard-notification</Id>
+                            <Queue>arn:aws:sqs:us-east-1:000000000000:%s</Queue>
+                            <Event>s3:ObjectCreated:*</Event>
+                        </QueueConfiguration>
+                    </NotificationConfiguration>
+                    """.formatted(postQueueName, putQueueName, wildcardQueueName))
+            .when()
+                .put("/" + BUCKET)
+            .then()
+                .statusCode(200);
+
+            given()
+                .multiPart("key", "uploads/notified.txt")
+                .multiPart("file", "notified.txt", "notified".getBytes(StandardCharsets.UTF_8), "text/plain")
+            .when()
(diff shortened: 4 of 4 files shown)


pgermosen (commented):
Thanks Aniket — small, correct, and it lands in the right place. Routing the fix through a
shared `createObject` parameterised by event name, rather than duplicating `storeObject` +
`fireNotifications` in a second method, is what keeps `putObject` provably unchanged.

A note on the build: the run I have is not fully green — `NeptuneOpenCypherIntegrationTest.
openCypherQueryWorksOverBolt` times out. That is unrelated to this PR (Neptune's
`CreateDBCluster` blocks synchronously in `waitForBackendReady` for up to 60s while a neo4j
container cold-starts, and the HTTP client gives up first), and an earlier run on the same
branch failed 18 tests purely from disk exhaustion. Every S3 test passed, including the one
this PR adds. So I am reviewing on 12,528 of 12,529 green, with the single failure
attributable to the environment rather than to this change.

## What I verified

**`s3:ObjectCreated:Post` is the right event name, and POST genuinely earns its own type.**
From the S3 event-notification reference:

> `s3:ObjectCreated:*` / `s3:ObjectCreated:Put` / `s3:ObjectCreated:Post` /
> `s3:ObjectCreated:Copy` / `s3:ObjectCreated:CompleteMultipartUpload` — Amazon S3 API
> operations such as `PUT`, `POST`, and `COPY` can create an object. With these event types,
> you can enable notifications when an object is created using a specific API operation.

So a browser upload firing `Put` was genuinely wrong, not merely imprecise.

**The one-line controller change covers every POST-object path.** `handleBucketPost`
(`S3Controller.java:1184`) has exactly two branches: `?delete` → `handleDeleteObjects`, and
`multipart/form-data` → `handlePresignedPost` → `doHandlePresignedPost`. Anything else is
rejected with `InvalidArgument`. There is no second browser-upload route still calling
`putObject`, which is the thing worth checking when a fix is a single call-site swap.

**Floci now emits all four documented `ObjectCreated` subtypes** — `Put` (`S3Service.java:346`),
`Post` (`:352`), `CompleteMultipartUpload` (`:2081`), `Copy` (`:3474`). Post was the only
gap; the set is complete.

**Zero behavioural drift on the options path.** The call this replaces was the five-argument
`putObject` at `S3Service.java:300-303`, which delegates with `new PutObjectOptions()` —
exactly what `postObject` constructs. So the only thing that changed is the event-name
string, which is the whole point. Worth stating explicitly because "add an overload and
switch the call" is the kind of change that quietly drops a default.

**Wildcard subscribers keep working.** `matchesEvent` (`:2916-2922`) prefixes `s3:` and does
a prefix match when the pattern ends in `*`, so `s3:ObjectCreated:*` still matches
`ObjectCreated:Post`. The CDI hook is guarded on `eventName.startsWith("ObjectCreated")`,
which also still fires. Neither needed changing, and neither broke.

**Test hygiene is good.** The `finally` clears the notification configuration *and* deletes
the queue, so a failure mid-test can't leave the bucket firing events into later ordered
tests in the class — a real hazard given `@Order`. And `cleanupBucket` gained
`uploads/notified.txt` (`:584`), so the new object doesn't leak either.

---

## (question) This silently stops `ObjectCreated:Put` subscribers from seeing browser uploads

That is the correct AWS behaviour and I am not asking you to change it. But it is a
behaviour change for existing Floci users: anyone whose notification configuration filters
on `s3:ObjectCreated:Put` and relies on presigned-POST uploads will find their handler goes
quiet, with no error anywhere to explain it.

`docs/services/s3.md` currently says nothing that would help them — its only mention of
notifications is the one-line table row `| **Notifications** | PutBucketNotification,
GetBucketNotification |`, with no event-type list at all. So there is no page where the old
behaviour is documented and no page where the new one is.

A sentence in `docs/services/s3.md` noting that presigned POST uploads emit
`ObjectCreated:Post` (and therefore no longer `ObjectCreated:Put`) would cost a line and
save someone a confusing afternoon. Failing that, calling it out in the PR description's
AWS Compatibility section so it lands in the changelog would do.

## (minor) The two cases most likely to matter aren't covered

`presignedPostEmitsPostNotification` pins the positive exact-match case: a subscription on
`s3:ObjectCreated:Post` receives the event. The two that carry more weight in practice are
missing:

- **The negative** — a bucket configured for `s3:ObjectCreated:Put` receives *nothing* from a
  browser upload. That is the actual regression being fixed, and it is the assertion that
  would fail loudest if someone later "simplified" `postObject` back into `putObject`.
- **The wildcard** — `s3:ObjectCreated:*` still catches browser uploads. I verified
  `matchesEvent` handles it, but most real notification configurations use the wildcard
  rather than an exact subtype, so it is the path most users are actually on and nothing
  pins it.

Both are a handful of lines in the test you already wrote.

## (minor) `postObject` hardcodes `new PutObjectOptions()`

Nothing is lost today: `doHandlePresignedPost` extracts only `x-amz-meta-*` entries and the
content type, so there are no options to forward.

But `putObject` exposes a `PutObjectOptions` overload and `postObject` does not, and POST
Object forms carry real fields that map onto it — `Cache-Control`, `Expires`,
`Content-Disposition`, `x-amz-storage-class`, `acl`. The moment anyone wires one of those
up, they have to change this signature rather than the call site. Mirroring the
`putObject` shape — a five-argument convenience overload delegating to one that takes
`PutObjectOptions`, as at `:300-303` — keeps the pair symmetric for free.

## Notes

- `createObject` being private while both public entry points delegate to it is the right
  split: it makes the event name the only axis of difference, which is exactly what a
  reader needs to see to trust that `putObject` is untouched.
- The test asserts on the raw message body containing `"eventName":"ObjectCreated:Post"`
  rather than parsing the envelope. Fine for this purpose, and it happens to also pin the
  JSON field name, which is what an SQS consumer actually matches on.

No blockers. The behaviour is right, the blast radius is genuinely one string, and the
question is about telling users rather than about the code.

aniketshukla1:
Thanks — addressed the follow-ups in f256552. The notification test now verifies all three compatibility paths: exact `ObjectCreated:Post` delivery, wildcard `ObjectCreated:*` delivery, and no delivery to an `ObjectCreated:Put` subscription. I also documented the behavior change in `docs/services/s3.md` and added the symmetric `postObject(..., PutObjectOptions)` overload. Rebased onto current `main`; `S3PresignedPostIntegrationTest` passes 20/20.

pgermosen (approved):
Thanks Aniket — all three follow-ups are closed, and I checked each one against the code rather
than the commit message. Nothing outstanding.

## The three items

**Docs.** `docs/services/s3.md` gains an `## Event Notifications` section that says the thing a
confused user would need:

> Browser and presigned POST uploads emit `s3:ObjectCreated:Post`, matching AWS S3. They do not
> emit `s3:ObjectCreated:Put`; use `s3:ObjectCreated:*` to subscribe to objects created by either
> method.

The second clause is the part that earns its place — naming the wildcard gives someone whose
`Put` subscription just went quiet an actual fix rather than only an explanation.

**The `postObject` overload pair now mirrors `putObject`.** Five-argument convenience delegating
to a six-argument form that takes `PutObjectOptions`, with `createObject` underneath and the event
name as the only axis of difference. That's the shape I was asking for.

**The test covers all three paths, in one bucket configuration.** Three queue configurations —
exact `Post`, exact `Put`, wildcard — one browser upload, then assertions that `Post` and wildcard
both received `"eventName":"ObjectCreated:Post"` and that the `Put` queue has no `<Message>` at
all. The negative case is the one I most wanted, since it's the regression being fixed and the
assertion that fails loudest if `postObject` ever gets folded back into `putObject`.

Worth noting why that negative assertion isn't flaky, because it's the kind of thing that looks
racy at a glance: `fireNotifications` (`S3Service.java:2825`) is a plain `for` loop calling
`sqsService.sendMessage` inline, so every matching queue is written before the POST returns 204.
And even if delivery were asynchronous, the two positive assertions run first and act as a
barrier — by the time the `Put` queue is checked, that event's fan-out has already completed. It's
correct twice over, which is the right amount for an assertion that proves a negative.

Cleanup holds up too: the `finally` clears the notification configuration and deletes all three
queues, and `cleanupBucket` picks up `uploads/notified.txt`. Given `@Order` on this class, a
half-cleaned bucket would have been a nasty way to make a later test fail.

## Notes

- The six-argument `postObject(..., PutObjectOptions)` has no caller today — `doHandlePresignedPost`
  uses the five-argument form. That's deliberate and it's what I asked for: POST Object forms carry
  `Cache-Control`, `Expires`, `Content-Disposition`, `x-amz-storage-class` and `acl`, all of which
  map onto `PutObjectOptions`, and the symmetry means wiring any of them up later is a call-site
  change rather than a signature change. Recording it here so it doesn't get tidied away as dead
  code by someone reading it cold.
- Minor, and I'm not asking for it: the three `createQueue` calls sit above the `try`, so a failure
  creating the second or third would leak the earlier ones. Test-only, and unreachable unless SQS
  itself is broken — at which point the test has bigger problems.
- Full check suite is green across all 19, and the branch is `MERGEABLE` / `CLEAN`, so unlike the
  first round I'm not reading around an environment-attributable failure.

The behaviour was right in round one; this round is the part that tells users about it and stops it
regressing.

PR #2587, opened 2026-08-26
feat(secretsmanager): persist and return secret resource policies

matanbaruch:
Secret resource policies now round-trip: `PutResourcePolicy` persists the policy on the secret and returns `ARN`/`Name`, `GetResourcePolicy` returns it, `DeleteResourcePolicy` clears it. Before this, Put and Delete were inert stubs returning `{}` and Get never returned a policy, so `aws_secretsmanager_secret_policy` could not be created through Terraform or Pulumi at all: the provider does `d.SetId(output.ARN)` on the Put response ([`secret_policy.go`](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/secretsmanager/secret_policy.go)), and an empty response fails the create with an empty resource id.

Behavior, matching the AWS API model and the provider's expectations:

1. `PutResourcePolicy` resolves the secret (name, full ARN, or partial ARN), validates `ResourcePolicy` is present and valid JSON, persists it, and returns `{ARN, Name}`.
2. `GetResourcePolicy` returns `{ARN, Name}` plus `ResourcePolicy` only when one is attached. AWS omits the member entirely otherwise, and the provider reads the absent field as "no policy" rather than as an error.
3. `DeleteResourcePolicy` clears the policy and returns `{ARN, Name}`.
4. All three ops reject a secret in its deletion recovery window with `InvalidRequestException` ("...marked for deletion"), reusing the existing guard. The provider substring-matches that message on `GetResourcePolicy` to treat the policy as gone, so this is part of the wire contract.
5. Unknown secret raises the existing `ResourceNotFoundException`.

Scope notes (kept deliberately narrow):

1. `BlockPublicPolicy` is accepted but not enforced. The emulator does no public-policy analysis, so `PublicPolicyException` is never raised.
2. `MalformedPolicyDocumentException` fires on JSON-parse failure only. No Statement or principal validation.
3. `DeleteResourcePolicy` on a secret with no policy attached succeeds with `{ARN, Name}`. The provider tolerates either behavior; quiet success seemed the safer emulation, happy to change if real AWS is known to differ.

Tests:

1. Handler-level (`SecretsManagerJsonHandlerTest`, real service instance): put/get/delete round trip addressed by full ARN as the provider sends it, ARN/Name echoed on Put and Delete, `ResourcePolicy` absent when none attached, missing/empty policy 400, malformed policy 400, unknown secret, and the marked-for-deletion contract on all three ops.
2. SDK-level (`compatibility-tests` `SecretsManagerTest`): a put/get/delete round trip with the real AWS Java SDK v2 covering the JSON wire shapes end to end.

The three actions were already dispatched in the handler switch, so the generated Supported Actions table is unchanged.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SecretsManagerTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/SecretsManagerTest.java
@@ -19,17 +19,23 @@ import software.amazon.awssdk.services.secretsmanager.model.BatchGetSecretValueR
 import software.amazon.awssdk.services.secretsmanager.model.BatchGetSecretValueResponse;
 import software.amazon.awssdk.services.secretsmanager.model.CreateSecretRequest;
 import software.amazon.awssdk.services.secretsmanager.model.CreateSecretResponse;
+import software.amazon.awssdk.services.secretsmanager.model.DeleteResourcePolicyRequest;
+import software.amazon.awssdk.services.secretsmanager.model.DeleteResourcePolicyResponse;
 import software.amazon.awssdk.services.secretsmanager.model.DeleteSecretRequest;
 import software.amazon.awssdk.services.secretsmanager.model.DescribeSecretRequest;
 import software.amazon.awssdk.services.secretsmanager.model.DescribeSecretResponse;
 import software.amazon.awssdk.services.secretsmanager.model.GetRandomPasswordRequest;
 import software.amazon.awssdk.services.secretsmanager.model.GetRandomPasswordResponse;
+import software.amazon.awssdk.services.secretsmanager.model.GetResourcePolicyRequest;
+import software.amazon.awssdk.services.secretsmanager.model.GetResourcePolicyResponse;
 import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRequest;
 import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueResponse;
 import software.amazon.awssdk.services.secretsmanager.model.ListSecretVersionIdsRequest;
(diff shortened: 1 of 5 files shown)


pgermosen (approved):
Thanks for this — clean, well-scoped fix, and it comes with the exact real-world motivation (the Terraform provider's `d.SetId(output.ARN)` on Put) that makes clear why the previous stub behavior was actually broken, not just incomplete.

I checked all three operations against the current Secrets Manager API reference:

- `PutResourcePolicy` / `DeleteResourcePolicy` response shape `{ARN, Name}` — matches exactly.
- `GetResourcePolicy` response `{ARN, Name, ResourcePolicy}` with `ResourcePolicy` correctly **omitted**, not returned as `null`, when nothing is attached — matches, and matters here since you note the Terraform provider depends on that exact absence.
- `InvalidRequestException` for a secret in its recovery window is documented on all three operations, and you correctly reused `throwIfPendingDeletion` from #2555 instead of re-deriving the guard.
- `MalformedPolicyDocumentException` on JSON-parse failure matches AWS's documented error list for `PutResourcePolicy`.
- The two things you left out (`BlockPublicPolicy`/`PublicPolicyException` enforcement, Statement/principal validation) are called out as deliberate scope decisions rather than silently missing — appreciated, since it means the next person doesn't have to rediscover that boundary.

Test coverage matches the claims one-to-one: round trip, omitted-when-none, delete-without-policy, missing/empty policy, malformed JSON, unknown secret, and the pending-deletion contract on all three ops at the handler level, plus a real AWS SDK v2 round trip at the compat-test level.

Merge is clean, CI is fully green.

Recommendation: **approve**.

PR #2588, opened 2026-08-26
docs(getting-started): Switch to the `io.floci` test container

sschuberth:
## Summary

Update the migration guide that still pointed at the pre-rename testcontainers coordinates. Also align referenced version to the actual latest.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A

## Checklist

- [ ] `./mvnw test` passes locally ()N/A
- [ ] New or updated integration test added (N/A)

### DIFF
--- a/README.md
+++ b/README.md
@@ -568,9 +568,11 @@ aws --endpoint-url http://localhost:4566 s3 ls
 
 Floci has Testcontainers modules for starting isolated Floci instances directly from tests. This avoids shared state, manual daemon setup, and port conflicts.
 
+For Testcontainers 1.x, use the versions as indicated in the table below.
+
 | Language | Package | Latest | Registry | Source |
 |---|---|---|---|---|
-| Java | `io.floci:testcontainers-floci` | `1.4.0` | [Maven Central](https://mvnrepository.com/artifact/io.floci/testcontainers-floci) | [GitHub](https://github.com/floci-io/testcontainers-floci) |
+| Java | `io.floci:testcontainers-floci` | `1.14.0` | [Maven Central](https://mvnrepository.com/artifact/io.floci/testcontainers-floci) | [GitHub](https://github.com/floci-io/testcontainers-floci) |
 | Node.js | `@floci/testcontainers` | `0.1.0` | [npm](https://www.npmjs.com/package/@floci/testcontainers) | [GitHub](https://github.com/floci-io/testcontainers-floci-node) |
 | Python | `testcontainers-floci` | `0.1.1` | [PyPI](https://pypi.org/project/testcontainers-floci/) | [GitHub](https://github.com/floci-io/testcontainers-floci-python) |
 | Go | In progress | In progress | N/A | [GitHub](https://github.com/floci-io/testcontainers-floci-go) |
@@ -582,7 +584,7 @@ Floci has Testcontainers modules for starting isolated Floci instances directly 
 <dependency>
     <groupId>io.floci</groupId>
     <artifactId>testcontainers-floci</artifactId>
-    <version>1.4.0</version>
+    <version>1.14.0</version>
     <scope>test</scope>
 </dependency>
 ```
@@ -609,7 +611,7 @@ class S3IntegrationTest {
 }
 ```
 
-For Testcontainers 2.x / Spring Boot 4.x, use version `2.5.0`.
+For Testcontainers 2.x / Spring Boot 4.x, use version `2.15.0`.
 
 </details>
 
--- a/docs/getting-started/migrate-from-localstack.md
+++ b/docs/getting-started/migrate-from-localstack.md
@@ -144,14 +144,15 @@ If you wait on `/_localstack/init` or `/_localstack/health` in CI or scripts, no
     <dependency>
       <groupId>org.testcontainers</groupId>
       <artifactId>localstack</artifactId>
+      <version>1.21.4</version>
       <scope>test</scope>
     </dependency>
 
     <!-- After -->
     <dependency>
-      <groupId>io.github.hectorvent</groupId>
-      <artifactId>floci-testcontainers</artifactId>
-      <version>LATEST</version>
+      <groupId>io.floci</groupId>
+      <artifactId>testcontainers-floci</artifactId>
+      <version>1.14.0</version>
       <scope>test</scope>
     </dependency>
     ```
--- a/docs/testcontainers/index.md
+++ b/docs/testcontainers/index.md
@@ -4,12 +4,16 @@ Floci has first-class Testcontainers modules for every major SDK language. Each 
 
 ## Available modules
 
+For Testcontainers 1.x, use the versions as indicated in the table below.
+
 | Language | Package | Version | Registry | Source |
 |---|---|---|---|---|
-| Java | `io.floci:testcontainers-floci` | `1.4.0` | [Maven Central](https://mvnrepository.com/artifact/io.floci/testcontainers-floci) | [GitHub](https://github.com/floci-io/testcontainers-floci) |
+| Java | `io.floci:testcontainers-floci` | `1.14.0` | [Maven Central](https://mvnrepository.com/artifact/io.floci/testcontainers-floci) | [GitHub](https://github.com/floci-io/testcontainers-floci) |
 | Node.js | `@floci/testcontainers` | `0.1.0` | [npm](https://www.npmjs.com/package/@floci/testcontainers) | [GitHub](https://github.com/floci-io/testcontainers-floci-node) |
 | Python | `testcontainers-floci` | `0.1.1` | [PyPI](https://pypi.org/project/testcontainers-floci/) | [GitHub](https://github.com/floci-io/testcontainers-floci-python) |
 | Go | — | 🚧 In progress | — | [GitHub](https://github.com/floci-io/testcontainers-floci-go) |
+
+For Testcontainers 2.x / Spring Boot 4.x, use version `2.15.0`.
 
 ## How it works
 
--- a/docs/testcontainers/java.md
+++ b/docs/testcontainers/java.md
@@ -6,8 +6,8 @@ Two artifact lines are published to keep in sync with the Testcontainers major v
 
 | Testcontainers version | Spring Boot | Artifact version |
 |---|---|---|
-| 1.x | 3.x | `1.4.0` |
-| 2.x | 4.x | `2.5.0` |
+| 1.x | 3.x | `1.14.0` |
+| 2.x | 4.x | `2.15.0` |
 
 ## Installation
 
@@ -17,7 +17,7 @@ Two artifact lines are published to keep in sync with the Testcontainers major v
     <dependency>
         <groupId>io.floci</groupId>
         <artifactId>testcontainers-floci</artifactId>
-        <version>1.4.0</version>
+        <version>1.14.0</version>
         <scope>test</scope>
     </dependency>
     ```
@@ -25,7 +25,7 @@ Two artifact lines are published to keep in sync with the Testcontainers major v
 === "Gradle"
 
     ```groovy
-    testImplementation 'io.floci:testcontainers-floci:1.4.0'
+    testImplementation 'io.floci:testcontainers-floci:1.14.0'
     ```
 
 ## Basic usage — JUnit 5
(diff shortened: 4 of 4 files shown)


sschuberth:
BTW, several other places in the code also still refer to "io.github.hectorvent"... not sure which of these might still be valid, though.

pgermosen (commented):
Thanks — the coordinate change is correct, and you picked the right target.

I checked it against the rest of the docs: `docs/testcontainers/index.md:9` and
`docs/testcontainers/java.md:18-19` both give `io.floci:testcontainers-floci`, and
`migrate-from-localstack.md:152-153` was the **only** place still carrying the old
`io.github.hectorvent:floci-testcontainers`. So this closes the last stale reference rather
than adding a second convention.

Worth calling out what you *didn't* touch, because it's the part that's easy to get wrong:
the remaining `io.github.hectorvent` occurrences in `docs/services/ec2.md` and
`docs/configuration/advanced/application-yml.md` are Java package and logger names, and the
emulator's own Maven `groupId` in `pom.xml` is still `io.github.hectorvent`. Only the
testcontainers artifact moved to `io.floci`. Leaving those alone is right.

## Please fill in the PR template

This PR's description is empty — the template was removed rather than completed. We do require
it on every PR, including one-line docs fixes, so please restore and fill it in.

For a docs-only change most of it is quick:

- **Summary** — one line is fine; here, that the migration guide still pointed at the
  pre-rename testcontainers coordinates.
- **Type of change** — tick *Docs / chore*.
- **AWS Compatibility** — N/A, say so explicitly.
- **Checklist** — `./mvnw test` and the integration-test box don't apply to a docs edit; mark
  them N/A rather than leaving them ambiguous. CONTRIBUTING asks that a PR without new tests
  explains why in the description, and "documentation only" is exactly that explanation.

It isn't busywork on our side: the description is what the changelog and release notes are
assembled from, and an empty one means this change lands invisibly.

## One thing to fix while you're in there

The snippet keeps `<version>LATEST</version>`:

```xml
<!-- After -->
<dependency>
  <groupId>io.floci</groupId>
  <artifactId>testcontainers-floci</artifactId>
  <version>LATEST</version>
  <scope>test</scope>
</dependency>
```

`LATEST` was deprecated in Maven 2 and no longer resolves for regular dependencies in Maven 3+
— it survives only for plugin versions, and is discouraged there too. A reader who copies this
block gets a resolution failure, which is a worse first experience than the stale coordinates
were.

The canonical snippet in `docs/testcontainers/java.md:20` pins a real version:

```xml
<version>1.4.0</version>
```

Since you're already correcting this exact dependency block, matching that would make the
migration guide agree with the Java guide on all three coordinates instead of two.

sschuberth:
> This PR's description is empty — the template was removed rather than completed. We do require
> it on every PR, including one-line docs fixes, so please restore and fill it in.

I see. I felt the template was mostly not applicable for my docs-only change, so I omitted it. But I'll add it again if it's indeed used for release notes.

> `LATEST` was deprecated in Maven 2 and no longer resolves for regular dependencies in Maven 3+
> — it survives only for plugin versions, and is discouraged there too.

I thought that "LATEST" was a placeholder anyway, and not a keyword than Maven really supports, so I left it as-is to avoid any maintenance effort updating a concrete version here. But I'll fill in the version if that's the preferred way.

pgermosen (commented):
Thanks — both round-1 items are done: the template is filled in, and `LATEST` is gone in favour of
a pinned version that actually exists. The version you picked is real and current.

The trouble is that it's current for only one of the two artifact lines, and the sweep changed the
Gradle halves without the Maven ones.

## First, a correction I owe you

My last round said:

> The canonical snippet in `docs/testcontainers/java.md:20` pins a real version:
> `<version>1.4.0</version>`

I meant that as an argument for pinning rather than `LATEST` — not as an endorsement of `1.4.0` as
the right number. I hadn't noticed the two-line table six lines above it, and `1.4.0` turns out to
be stale in its own right. If that pointer read as "sync everything to 1.4.0", that's on my
phrasing.

## (should fix) The Maven and Gradle tabs of the same block now disagree — and it isn't cosmetic

`docs/testcontainers/java.md` opens by explaining that two lines are published deliberately:

> Two artifact lines are published to keep in sync with the Testcontainers major version:
>
> | Testcontainers version | Spring Boot | Artifact version |
> | 1.x | 3.x | `1.4.0` |
> | 2.x | 4.x | `2.5.0` |

After this PR the file states three different versions for the same dependency:

| Location | `testcontainers-floci` |
|---|---|
| Table, 1.x row (`:9`) | `1.4.0` |
| Table, 2.x row (`:10`) | `2.5.0` |
| **Maven** tab (`:20`) | `1.4.0` |
| **Gradle** tab (`:28`) | `2.15.0` ← changed here |

and the same split repeats for the Spring Boot companion at `:149` (Maven, `1.4.0`) and `:157`
(Gradle, `2.15.0`).

The Maven and Gradle tabs are two spellings of one dependency — a reader picks whichever matches
their build tool and expects the same artifact. Per the table on the very same page, they no longer
get one: the 1.x line targets **Spring Boot 3.x** and the 2.x line targets **Spring Boot 4.x**. So
a Spring Boot 3 project that follows the Gradle tab now pulls the Spring Boot 4 companion. That's a
snippet that fails at runtime rather than a number that looks untidy, which is why I'd hold the PR
for it.

It crosses files too: `migrate-from-localstack.md`'s **Maven** block now says `2.15.0`, while
java.md's **Maven** block says `1.4.0`.

## What the real numbers are

Worth having, because the sweep is half-done rather than wrong-headed. Upstream
`floci-io/testcontainers-floci` cuts the two lines in lockstep pairs — v1.14.0 and v2.15.0 were
both released on 2026-08-19, v1.13.0 and v2.14.0 both on 2026-08-15, and so on back. Current heads:

- **1.x line (Testcontainers 1.x / Spring Boot 3.x): `1.14.0`**
- **2.x line (Testcontainers 2.x / Spring Boot 4.x): `2.15.0`**

So `2.15.0` is exactly right for the 2.x line. But `1.4.0` and `2.5.0` are both real-and-old — the
tag list runs `v1.0.0 … v1.4.0 … v1.14.0`, so the 1.x row is ten releases behind, and `v2.5.0` is
ten behind on the other side. Since you're already editing these numbers, correcting the table rows
to `1.14.0` and `2.15.0` and making each Maven tab match its Gradle neighbour would leave the page
self-consistent and current in one pass.

## (question) The single-row tables can't express two lines

`README.md:573` heads its column **Latest** and `docs/testcontainers/index.md:9` heads its
**Version**, and each has one Java row. With two supported lines there isn't one correct value, and
this PR puts `2.15.0` there alone.

That steers in the riskier direction: GitHub marks **v1.14.0** as the *Latest* release, `2.15.0` is
merely the higher number, and someone on Spring Boot 3 reading a column headed "Latest" will take
`2.15.0` as the one to use.

Two shapes that both work, and I don't mind which — it's a docs-policy call rather than a
correctness one:

- one row showing both, e.g. `1.14.0` (TC 1.x / SB 3.x) · `2.15.0` (TC 2.x / SB 4.x); or
- the 1.x line in the cell, with the 2.x line footnoted and java.md as the detail page.

What I'd rather not merge is a single number in a "Latest" column that quietly points Spring Boot 3
users at the Spring Boot 4 artifact.

## Nit

- Checklist line reads `- [ ] ./mvnw test passes locally ()N/A` — stray `()`.

## Notes

- Splitting the version bump into its own commit (`docs: Update Maven *testcontainers-floci
  artifacts to 2.15.0`) separate from the coordinate rename was the right call; it makes the two
  concerns reviewable independently, which is how I noticed the Gradle-only pattern.
- Everything I said last round about what you *didn't* touch still holds — the `io.github.hectorvent`
  occurrences in `docs/services/ec2.md` and `application-yml.md` are package and logger names, and
  the emulator's own `groupId` is unchanged. Only the testcontainers artifact moved to `io.floci`,
  and this PR keeps it that way.

sschuberth:
@pgermosen, these walls of text seem to be AI-generated, and I'm getting a bit tired of reading through them. I'd appreciate some feedback from a human being.

pgermosen:
> @pgermosen, these walls of text seem to be AI-generated, and I'm getting a bit tired of reading through them. I'd appreciate some feedback from a human being.

It is AI-generated, but using my input... and no, I won't change the way I do the reviews; I have a process for this to prevent things that I could forget.

pgermosen (approved):
Thanks — everything from the last round is closed, and the version sweep is now complete and
self-consistent across all four files.

## First, a correction I owe you

My last round said:

> `README.md:573` heads its column **Latest** and `docs/testcontainers/index.md:9` heads its
> **Version**, and each has one Java row. With two supported lines there isn't one correct value.

That was right for `docs/testcontainers/index.md` and only half right for `README.md`. The README
already carried a prose line naming the 2.x artifact — *"For Testcontainers 2.x / Spring Boot 4.x,
use version `2.5.0`"* — below the fold of the hunk I was reading, so it wasn't unable to express
two lines; its second line was simply stale. The diff I had showed me the table row and not the
sentence twelve lines further down. Worth saying because it changes what you were being asked for
there: the README needed a number corrected, not a structure invented.

## The Maven/Gradle split is closed

All four snippets in `docs/testcontainers/java.md` now read `1.14.0`:

| Location | Artifact | Version |
|---|---|---|
| Maven tab (`:20`) | `testcontainers-floci` | `1.14.0` |
| Gradle tab (`:28`) | `testcontainers-floci` | `1.14.0` |
| Maven tab (`:149`) | `spring-boot-testcontainers-floci` | `1.14.0` |
| Gradle tab (`:157`) | `spring-boot-testcontainers-floci` | `1.14.0` |

and both agree with the 1.x row of the table six lines above them. That was the part that mattered:
the two tabs are two spellings of one dependency, and with the 1.x and 2.x lines targeting different
Spring Boot majors, a reader who picked the Gradle tab on a Spring Boot 3 project was previously
being handed the Spring Boot 4 companion. Nothing in the file contradicts anything else now.

`docs/getting-started/migrate-from-localstack.md` moved to `1.14.0` with them, which closes the
cross-file half — its Maven block and `java.md`'s Maven block had been giving different answers for
the same dependency.

## Both table rows are current

`1.14.0` and `2.15.0` are the current heads of the two lines upstream publishes — the two are cut
in lockstep pairs, and those are the most recent pair. So the table in `java.md` is now accurate on
both rows rather than ten releases behind on each, which is where it started.

## The single-row tables are handled the right way

Both `README.md` and `docs/testcontainers/index.md` now label the table as the 1.x line —

> For Testcontainers 1.x, use the versions as indicated in the table below.

— and name the 2.x line in prose beneath it. That's the second of the two shapes I offered, and I
think it's the better one: the cell carries one unambiguous answer, and the reader who needs the
other line is told so immediately above and below rather than having to infer it from a column
heading. `index.md` gains that framing for the first time; the README's version of it is now
pointing at a real release instead of `2.5.0`.

Worth noting the risk this removes: a single number under a column headed **Latest** would have
steered a Spring Boot 3 reader onto the 4.x artifact, since `2.15.0` is the higher number while
`1.14.0` is what upstream flags as the latest release. Neither file can now be read that way.

## The `1.21.4` addition — checked, and it's right

Pinning the "Before" LocalStack dependency wasn't something I asked for, and it's a good call: that
block previously had no `<version>` element at all, which has exactly the won't-resolve property
that made `LATEST` worth flagging in the first round. Fixing both in the same pass is consistent.

I verified the number rather than assuming it, and it's worth recording how, because the first
answer was misleading. Maven Central's search API returns `1.21.3` as the newest
`org.testcontainers:localstack` and no `1.21.4` at all — which looked like a clean finding of the
same shape as the original one. The Testcontainers Java release list settles it the other way:
`1.21.4` was released on 2025-12-15, after `1.21.3` in June. The Central index just didn't surface
it. So the pin is a real version, and the 1.x line is the right one to show in a block about
migrating off `org.testcontainers:localstack`.



## Note

- The four files now agree with each other and with upstream, which is the property that was
  missing in both previous rounds — first the coordinates, then the versions. Nothing in the docs
  set still references `io.github.hectorvent:floci-testcontainers` or an unresolvable version.

sschuberth:
> It is AI-generated, but using my input... and no, I won't change the way I do the reviews

At the very minimum, you should not let the AI pretend to be you, by commenting as your user. I believe such impersonation is a vilification for contributors, who have a right to know with whom they're interacting. It's fine to use AI as tools, but not as a substitute for yourself. Which is in line with AI policies from big Open Source projects and foundations.

PR #2591, opened 2026-08-26
feat(glue): jobs and crawlers

namanONcode:
## Summary
This pull request introduces full support for AWS Glue Job and Crawler resources in the Floci emulator, ensuring adherence to AWS behavior and data models. 

**Key Additions:**
* **Full CRUD Operations**: Implements Create, Read, Update, and Delete endpoints for both Glue Jobs and Crawlers (`CreateJob`, `GetJob`, `UpdateJob`, `DeleteJob`, `CreateCrawler`, `GetCrawler`, `UpdateCrawler`, `DeleteCrawler`).
* **Resource Tagging & ARN Generation**: Adds accurate ARN generation and tagging support (`TagResource`, `UntagResource`, `GetTags`) for both resources using a centralized `ResourceGroupsTaggingService`.
* **Data Models**: Introduces complete models and definitions for these resources, including specific targets like `CatalogTarget`, `ConnectionsList`, `DynamoDBTarget`, `DeltaTarget`, `HudiTarget`, and `IcebergTarget`.
* **Persistent Storage**: Adds properly wired persistent stores within `GlueService` to handle Jobs and Crawlers alongside existing Glue catalogs.
* **Comprehensive Testing**: Contains robust unit/integration tests verifying CRUD logic, duplicate validations (`AlreadyExistsException`), pagination behavior, and tagging mechanisms. 
* **Documentation**: Updates the Glue service documentation (`docs/services/glue.md`) to outline the new supported Job and Crawler actions.

Closes #<Insert_Issue_Number_If_Applicable>

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility
- Verified using **AWS SDK for Java V2 (2.29.x)** and **AWS CLI (v2.15.x)**. 
- Validation specifically ensured that resource creation correctly enforces required fields and strict error-handling shapes (e.g., proper 404s for `EntityNotFoundException` and 400s for duplicate resource scenarios). Pagination for List actions respects AWS maximum page sizes correctly. 

## Checklist

### DIFF
--- a/docs/services/glue.md
+++ b/docs/services/glue.md
@@ -44,6 +44,26 @@ Floci emulates the AWS Glue Data Catalog and Glue Schema Registry, allowing you 
 | UpdateUserDefinedFunction | Updates a stored user-defined function. |
 | DeleteUserDefinedFunction | Deletes a user-defined function from a database. |
 
+#### Jobs
+
+| Action | Description |
+|--------|-------------|
+| CreateJob | Creates a new job definition. |
+| GetJob | Retrieves an existing job definition. |
+| GetJobs | Retrieves all current job definitions. |
+| UpdateJob | Updates an existing job definition. |
+| DeleteJob | Deletes a specified job definition. |
+
+#### Crawlers
+
+| Action | Description |
+|--------|-------------|
+| CreateCrawler | Creates a new crawler with specified targets, role, configuration, and optional schedule. |
+| GetCrawler | Retrieves metadata for a specified crawler. |
+| GetCrawlers | Retrieves metadata for all crawlers defined in the customer account. |
+| UpdateCrawler | Updates a crawler. |
+| DeleteCrawler | Removes a specified crawler from the AWS Glue Data Catalog. |
+
 ### Schema Registry
 
 #### Registries
--- a/src/main/java/io/github/hectorvent/floci/services/glue/GlueJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/glue/GlueJsonHandler.java
@@ -167,9 +167,9 @@ public class GlueJsonHandler {
             case "PutSchemaVersionMetadata" -> handlePutSchemaVersionMetadata(request);
             case "RemoveSchemaVersionMetadata" -> handleRemoveSchemaVersionMetadata(request);
             case "QuerySchemaVersionMetadata" -> handleQuerySchemaVersionMetadata(request);
-            case "TagResource" -> handleTagResource(request);
-            case "UntagResource" -> handleUntagResource(request);
-            case "GetTags" -> handleGetTags(request);
+            case "TagResource" -> handleTagResource(request, region);
+            case "UntagResource" -> handleUntagResource(request, region);
+            case "GetTags" -> handleGetTags(request, region);
             case "CreateJob" -> {
                 Job job = mapper.treeToValue(request, Job.class);
                 Map<String, String> tags = request.has("Tags") ? mapper.convertValue(request.get("Tags"), new TypeReference<>() {}) : null;
@@ -181,7 +181,10 @@ public class GlueJsonHandler {
                 yield Response.ok(Map.of("Job", glueService.getJob(name))).build();
             }
             case "GetJobs" -> {
-                yield Response.ok(Map.of("Jobs", glueService.getJobs())).build();
+                Integer maxResults = readMaxResults(request);
+                String nextToken = readNextToken(request);
+                GlueService.Page<Job> page = glueService.getJobs(maxResults, nextToken);
+                yield Response.ok(pageResponse("Jobs", page.items(), page.nextToken())).build();
             }
             case "UpdateJob" -> {
                 String name = request.get("JobName").asText();
@@ -205,7 +208,10 @@ public class GlueJsonHandler {
                 yield Response.ok(Map.of("Crawler", glueService.getCrawler(name))).build();
             }
             case "GetCrawlers" -> {
-                yield Response.ok(Map.of("Crawlers", glueService.getCrawlers())).build();
+                Integer maxResults = readMaxResults(request);
+                String nextToken = readNextToken(request);
+                GlueService.Page<Crawler> page = glueService.getCrawlers(maxResults, nextToken);
+                yield Response.ok(pageResponse("Crawlers", page.items(), page.nextToken())).build();
             }
             case "UpdateCrawler" -> {
                 Crawler update = mapper.treeToValue(request, Crawler.class);
@@ -746,28 +752,28 @@ public class GlueJsonHandler {
     }
 
     @SuppressWarnings("unchecked")
-    private Response handleTagResource(JsonNode request) {
+    private Response handleTagResource(JsonNode request, String region) {
         String arn = request.path("ResourceArn").asText(null);
         Map<String, String> tagsToAdd = request.has("TagsToAdd")
                 ? mapper.convertValue(request.get("TagsToAdd"), Map.class)
                 : null;
-        schemaRegistryService.tagResource(arn, tagsToAdd);
+        glueService.tagResource(arn, tagsToAdd, region);
         return Response.ok(Map.of()).build();
     }
 
     @SuppressWarnings("unchecked")
-    private Response handleUntagResource(JsonNode request) {
+    private Response handleUntagResource(JsonNode request, String region) {
         String arn = request.path("ResourceArn").asText(null);
         List<String> tagsToRemove = request.has("TagsToRemove")
                 ? mapper.convertValue(request.get("TagsToRemove"), List.class)
                 : null;
-        schemaRegistryService.untagResource(arn, tagsToRemove);
+        glueService.untagResource(arn, tagsToRemove, region);
         return Response.ok(Map.of()).build();
     }
 
-    private Response handleGetTags(JsonNode request) {
+    private Response handleGetTags(JsonNode request, String region) {
         String arn = request.path("ResourceArn").asText(null);
-        Map<String, String> tags = schemaRegistryService.getTags(arn);
+        Map<String, String> tags = glueService.getTags(arn, region);
         return Response.ok(Map.of("Tags", tags)).build();
     }
 }
--- a/src/main/java/io/github/hectorvent/floci/services/glue/GlueJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/glue/GlueJsonHandler.java
@@ -4,13 +4,7 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.github.hectorvent.floci.core.common.AwsException;
-import io.github.hectorvent.floci.services.glue.model.Crawler;
-import io.github.hectorvent.floci.services.glue.model.Database;
-import io.github.hectorvent.floci.services.glue.model.Job;
-import io.github.hectorvent.floci.services.glue.model.JobUpdate;
-import io.github.hectorvent.floci.services.glue.model.Partition;
-import io.github.hectorvent.floci.services.glue.model.Table;
-import io.github.hectorvent.floci.services.glue.model.UserDefinedFunction;
+import io.github.hectorvent.floci.services.glue.model.*;
 import io.github.hectorvent.floci.services.glue.schemaregistry.GlueSchemaRegistryService;
 import io.github.hectorvent.floci.services.glue.schemaregistry.model.Registry;
 import io.github.hectorvent.floci.services.glue.schemaregistry.model.RegistryId;
@@ -171,14 +165,14 @@ public class GlueJsonHandler {
             case "UntagResource" -> handleUntagResource(request, region);
             case "GetTags" -> handleGetTags(request, region);
             case "CreateJob" -> {
-                Job job = mapper.treeToValue(request, Job.class);
-                Map<String, String> tags = request.has("Tags") ? mapper.convertValue(request.get("Tags"), new TypeReference<>() {}) : null;
-                glueService.createJob(job, tags, region);
-                yield Response.ok(Map.of("Name", job.getName())).build();
+                CreateJobRequest req = mapper.treeToValue(request, CreateJobRequest.class);
+                Job job = toDomain(req);
+                glueService.createJob(job, req.getTags(), region);
+                yield Response.ok(new CreateJobResponse(job.getName())).build();
             }
             case "GetJob" -> {
-                String name = request.get("JobName").asText();
-                yield Response.ok(Map.of("Job", glueService.getJob(name))).build();
+                GetJobRequest req = mapper.treeToValue(request, GetJobRequest.class);
+                yield Response.ok(new GetJobResponse(glueService.getJob(req.getJobName()))).build();
             }
             case "GetJobs" -> {
                 Integer maxResults = readMaxResults(request);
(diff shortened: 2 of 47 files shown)


pgermosen (commented):
@namanONcode — thanks for this. Scope note: at ~2000 lines across 46 files (most of them model
classes) I checked the handler wiring, the error vocabulary, the test delta and the list
response envelopes, and skimmed the models. That's a deliberately shallow read, for a reason I
come back to below.

## What's right

**The two deleted tests were the correct thing to delete.** `GlueJsonHandlerEmptyListTest`
exists to pin empty-list responses for "resources the emulator does not model", and this PR
models Jobs and Crawlers — so `getJobsReturnsEmptyJobsList` and
`getCrawlersReturnsEmptyCrawlersList` stopped describing reality. You also updated the class
javadoc to drop them from its list rather than leaving it stale. That reads like removed
coverage at a glance and isn't, so it's worth saying out loud.

**`GetJobs` / `GetCrawlers` are real now, not stubs** — `maxResults`/`nextToken` through a
shared `pageResponse` helper, which is a genuine upgrade over the empty lists they replaced.

**The error vocabulary is Glue's own** — `EntityNotFoundException`, `InvalidInputException`,
`AlreadyExistsException` and `ConcurrentModificationException` are all real shapes for this
service, and the distribution across the operations looks sensible.

## (should fix) ~1900 lines of new behaviour, no new tests

The only test changes in this PR are the two justified deletions above and two constructor
arguments added to `GlueServiceTest` so it still compiles against the new `GlueService`
signature. Net, this adds full CRUD for two resources plus tagging and ARN generation, and
removes two tests without adding any.

The checklist says otherwise:

> - [x] New or updated integration test added

CONTRIBUTING asks that a PR without new tests explain why in the description. That's the ask
here — either the tests, or the reason.

This is also why my read above is shallow rather than thorough. With no tests I can't cheaply
tell whether `UpdateJob`'s merge semantics are right, whether a crawler round-trips its targets,
or whether the tag ARNs resolve — and reading 46 files closely enough to answer that by
inspection is a much worse use of review time than a test suite that answers it directly. Tests
covering the create/get/update/delete round trip for each resource would let me give you real
feedback on the behaviour instead of the shape.

## (should fix) The PR title will break the changelog

The title is `(feat)glue: jobs and crawlers`. Conventional Commits wants `feat(glue): jobs and
crawlers` — type first, scope in the parentheses.

Your two commits are both formatted correctly, which is why the "Commits follow Conventional
Commits" check is green: it validates commits, not the title. But this repo squashes with
`squash_merge_commit_title: COMMIT_OR_PR_TITLE`, and with more than one commit that resolves to
the **PR title**. So the squashed subject would be `(feat)glue: …`, semantic-release wouldn't
parse it as a `feat`, and the change would neither bump the minor version nor appear in
`CHANGELOG.md`.

One-line fix, but it has to be the title rather than the commits.

## (should fix) The template is unfilled

**Summary** and **AWS Compatibility** are still the raw placeholder comments:

```
## Summary
<!-- What does this PR do? Link any related issues with "Closes #N" -->
```

The generated overview above them is useful, but it isn't a substitute — the description is what
the changelog and release notes are assembled from, and AWS Compatibility is where I'd normally
look for which SDK or CLI version you verified the wire protocol against. For a PR adding ~55
new request/response shapes, that section is doing real work.

namanONcode:
> @namanONcode — thanks for this. Scope note: at ~2000 lines across 46 files (most of them model classes) I checked the handler wiring, the error vocabulary, the test delta and the list response envelopes, and skimmed the models. That's a deliberately shallow read, for a reason I come back to below.
> ## What's right
> 
> **The two deleted tests were the correct thing to delete.** `GlueJsonHandlerEmptyListTest` exists to pin empty-list responses for "resources the emulator does not model", and this PR models Jobs and Crawlers — so `getJobsReturnsEmptyJobsList` and `getCrawlersReturnsEmptyCrawlersList` stopped describing reality. You also updated the class javadoc to drop them from its list rather than leaving it stale. That reads like removed coverage at a glance and isn't, so it's worth saying out loud.
> 
> **`GetJobs` / `GetCrawlers` are real now, not stubs** — `maxResults`/`nextToken` through a shared `pageResponse` helper, which is a genuine upgrade over the empty lists they replaced.
> 
> **The error vocabulary is Glue's own** — `EntityNotFoundException`, `InvalidInputException`, `AlreadyExistsException` and `ConcurrentModificationException` are all real shapes for this service, and the distribution across the operations looks sensible.
> ## (should fix) ~1900 lines of new behaviour, no new tests
> 
> The only test changes in this PR are the two justified deletions above and two constructor arguments added to `GlueServiceTest` so it still compiles against the new `GlueService` signature. Net, this adds full CRUD for two resources plus tagging and ARN generation, and removes two tests without adding any.
> 
> The checklist says otherwise:
> 
> > * [x]  New or updated integration test added
> 
> CONTRIBUTING asks that a PR without new tests explain why in the description. That's the ask here — either the tests, or the reason.
> 
> This is also why my read above is shallow rather than thorough. With no tests I can't cheaply tell whether `UpdateJob`'s merge semantics are right, whether a crawler round-trips its targets, or whether the tag ARNs resolve — and reading 46 files closely enough to answer that by inspection is a much worse use of review time than a test suite that answers it directly. Tests covering the create/get/update/delete round trip for each resource would let me give you real feedback on the behaviour instead of the shape.
> ## (should fix) The PR title will break the changelog
> 
> The title is `(feat)glue: jobs and crawlers`. Conventional Commits wants `feat(glue): jobs and crawlers` — type first, scope in the parentheses.
> 
> Your two commits are both formatted correctly, which is why the "Commits follow Conventional Commits" check is green: it validates commits, not the title. But this repo squashes with `squash_merge_commit_title: COMMIT_OR_PR_TITLE`, and with more than one commit that resolves to the **PR title**. So the squashed subject would be `(feat)glue: …`, semantic-release wouldn't parse it as a `feat`, and the change would neither bump the minor version nor appear in `CHANGELOG.md`.
> 
> One-line fix, but it has to be the title rather than the commits.
> ## (should fix) The template is unfilled
> 
> **Summary** and **AWS Compatibility** are still the raw placeholder comments:
> 
> ```
> ## Summary
> 
> ```
> 
> The generated overview above them is useful, but it isn't a substitute — the description is what the changelog and release notes are assembled from, and AWS Compatibility is where I'd normally look for which SDK or CLI version you verified the wire protocol against. For a PR adding ~55 new request/response shapes, that section is doing real work.

@pgermosen i had resolved all issues reagarding pr and improved test coverage please review it

pgermosen (commented):
Thanks — the three round-1 items are all genuinely closed. The title is `feat(glue): jobs and
crawlers` now, so the squashed subject will parse and the change will land in the changelog.
Summary and AWS Compatibility are filled in. And the tests are real: ~175 lines covering
create/get/update/delete round trips for both resources, duplicate rejection, tagging
add/remove/read, pagination across page boundaries, and the `MaxResults` bounds. That's the
suite I asked for, and having it let me read the behaviour this round instead of the shape.

Which is how I found the following.

**Note:** everything below comes from reading the diff against the AWS API reference. I haven't
had the suite run against this round, so none of it is build-verified.

## (blocker) `CreateCrawler` and `UpdateCrawler` 500 on any request with a `Schedule`

`GlueJsonHandler.java:201` and `:217` both deserialize the whole request body into the
`Crawler` entity:

```java
Crawler crawler = mapper.treeToValue(request, Crawler.class);
```

But `Schedule` has two different types in Glue depending on which side of the wire it's on:

- In the **`Crawler` entity** returned by `GetCrawler`/`GetCrawlers`, `Schedule` is a structure —
  `{"ScheduleExpression": "...", "State": "..."}`.
- In the **`CreateCrawler` and `UpdateCrawler` requests**, `Schedule` is a plain cron **string**.
  Verbatim from the API reference for both: *"A `cron` expression used to specify the schedule…
  For example, to run something every day at 12:15 UTC, you would specify: `cron(15 12 * * ? *)`.
  Type: String"*.

`Crawler.java` models the field as a `Schedule` object, which is correct for the entity. Feeding
a request into it is not. `Schedule` has only a no-arg constructor and no `@JsonCreator`, so
Jackson can't build one from a JSON string — it throws `MismatchedInputException`. That isn't an
`AwsException`, so it falls through `AwsJson11Controller.java`'s generic `catch (Exception e)` into
`JsonErrorResponseUtils.createErrorResponse(Exception)`, which is hardcoded to:

```java
return JsonErrorResponseUtils.createErrorResponse(500, "InternalFailure", "InternalFailure", e.getMessage(), null);
```

So a scheduled crawler — the ordinary case — gets an HTTP 500 where AWS returns 200:

```
aws glue create-crawler \
  --name c1 --role my-role \
  --targets '{"S3Targets":[{"Path":"s3://bucket/prefix/"}]}' \
  --schedule 'cron(15 12 * * ? *)'
```

Same on `update-crawler --schedule`. Unscheduled crawlers work, which is why the tests pass —
they build `Crawler` objects in Java and call `GlueService` directly, so nothing in the suite
ever puts a request body through the handler.

The fix is already sitting in the PR, which brings me to the next point.

## (should fix) Eighteen request/response model classes are dead code — including the two that fix the blocker

`CreateCrawlerRequest`, `UpdateCrawlerRequest`, `CreateJobRequest`, `GetJobRequest`,
`GetJobResponse`, `UpdateJobRequest`, `UpdateJobResponse`, `DeleteJobRequest`,
`DeleteJobResponse`, `CreateJobResponse`, `CreateCrawlerResponse`, `GetCrawlerRequest`,
`GetCrawlerResponse`, `GetCrawlersRequest`, `GetCrawlersResponse`, `UpdateCrawlerResponse`,
`DeleteCrawlerRequest`, `DeleteCrawlerResponse` — roughly 600 lines — are declared and never
referenced anywhere. Not by the handler, not by the service, not by the tests. The handler parses
with `Job.class`/`Crawler.class` and builds responses with `Map.of(...)`.

What makes this worth more than a tidy-up note: you modelled the request shapes **correctly**.

```java
// CreateCrawlerRequest.java
private String schedule;

// UpdateCrawlerRequest.java
private String schedule;
```

Both are `String`, matching AWS. The blocker above exists only because the handler reaches for
`Crawler.class` instead of the request classes you'd already written. So the choice is:

- Wire them in — `treeToValue(request, CreateCrawlerRequest.class)`, map to the stored `Crawler`
  — which closes the blocker and gives you a place to hang required-field validation; or
- Delete them and give `Crawler`/`Job` a `Schedule` field that accepts both forms.

I'd take the first. Either way, a PR shouldn't merge with 600 lines of unreachable model code.

## (should fix) Job and crawler names are being lowercased

`createJob` (`GlueService.java:1030`), `getJob`, `updateJob`, `deleteJob` and the four crawler
equivalents all key their store through `normalizeName`, which is `name.toLowerCase(Locale.ROOT)`
(`GlueService.java:643`).

That helper is correct for what it was written for — Data Catalog databases, tables and functions,
where AWS really does lowercase names. Jobs and crawlers are not Data Catalog entities and AWS
preserves their case; the `Name` pattern for both permits uppercase and nothing in the reference
documents any folding. Three consequences:

- `CreateJob{"Name":"MyETLJob"}` followed by `CreateJob{"Name":"myetljob"}` returns
  `AlreadyExistsException`. AWS creates two independent jobs.
- `GetJob{"JobName":"MYETLJOB"}` returns the job. AWS returns `EntityNotFoundException`.
- The stored object keeps its original casing (`createJob` normalizes the *key* but never
  `job.setName(...)`), so `GetJob` echoes `MyETLJob` while the store key is `myetljob`. Then
  `GlueService.java:1056` sorts the `GetJobs` page on the un-normalized `Job::getName` — so the
  key you look up by and the value you sort by disagree on case. Both crawler paths have the same
  split at `:1137`.

Keying on the raw name is the fix. Worth a one-line comment at the call site saying why jobs and
crawlers skip `normalizeName`, since every other resource in the file uses it and the omission
will otherwise read like an oversight.

## (should fix) No required-field validation — missing fields produce 500, not `InvalidInputException`

Per the API reference, `CreateJob` requires `Name`, `Role` and `Command`; `CreateCrawler` requires
`Name`, `Role` and `Targets` (and *"At least one crawl target must be specified, in the `s3Targets`
field, the `jdbcTargets` field, or the `DynamoDBTargets` field"*). None of the six are checked.

The failure mode is the same 500 path as the blocker:

- `CreateJob` with no `Name` → `normalizeName(job.getName())` NPEs at `GlueService.java:1031`.
- `GetJob` with no `JobName` → `request.get("JobName").asText()` NPEs at `GlueJsonHandler.java:180`.
  Same shape at `:190` and `:196` (`JobName`), `:207` and `:222` (`Name`).
- `UpdateJob` with no `JobUpdate` → `treeToValue(null, …)` yields null, then `updateJob` NPEs
  dereferencing it at `GlueService.java:1070`.

The rest of this file already uses `request.path("X").asText(null)` followed by an explicit check —
`handleTagResource` at `:756` is the pattern. Reusing it turns all of these into
`InvalidInputException` 400s, which is what AWS returns.

Worth noting the tests currently pin the wrong side of this:
`resourceTaggingWorksForJobsAndCrawlers` creates a job with nothing but a `Name` and expects
success, and `getJobsAndCrawlersPagination` creates ten more the same way. AWS rejects all of them.

The PR description says validation *"specifically ensured that resource creation correctly enforces
required fields"* — that's the claim I'd most like reconciled with the code, in either direction.

## (should fix) `UpdateJob` silently drops `SourceControlDetails`

`updateJob` rebuilds the job from 22 explicit setters (`GlueService.java:1070-1091`).
`JobUpdate` has a 23rd field — `sourceControlDetails` — and it isn't among them. Send a
`JobUpdate` with `SourceControlDetails` and it's accepted, dropped, and the next `GetJob` shows
nothing changed.

Related, at `:1065-1068`: the fresh `Job` copies only `Name` and `CreatedOn` from the existing
record. `Job` has a `profileName` that `JobUpdate` has no field for, so it can't be a deliberate
reset — no caller can express it — but it's wiped on every update anyway. It should be carried
across with `CreatedOn`.

To be clear about what is *not* wrong here: full replacement is the right semantics for
`UpdateJob` — AWS resets unspecified configuration rather than merging — so the asymmetry with
`updateCrawler`'s null-coalescing merge at `:1141` is correct and shouldn't be "fixed". Only the
two omissions above need attention. A comment at `:1065` recording that the replacement is
deliberate would protect it from a future tidy-up.

## (should fix) `Crawler.State` and `Version` are modelled but never set

`Crawler.java` carries both fields, correctly annotated. Nothing in `GlueService` ever assigns
either, so `GetCrawler` and `GetCrawlers` return crawlers with `State` and `Version` absent
entirely. Per the reference, `State` is *"Indicates whether the crawler is running, or whether a
run is pending"* with valid values `READY | RUNNING | STOPPING`, and a real crawler reports `READY`
from the moment `CreateCrawler` succeeds.

An SDK consumer branching on `crawler.state()` gets null and takes whatever its default arm is.
Since the emulator never runs a crawler, `READY` at create time and left alone thereafter is both
accurate and free. `Version` starting at 1 and incrementing in `updateCrawler` is the natural
companion, and `updateCrawler` is already rewriting the record.

## (question) Nothing tests the handler

Every new test drives `GlueService` directly. That's a reasonable layer to test and the coverage
there is good — but the blocker, all five NPE paths, and the response envelopes all live in
`GlueJsonHandler`, which the suite never touches. A handful of tests going through
`handler.handle("CreateCrawler", node, region)` with a realistic request body — including
`"Schedule": "cron(15 12 * * ? *)"` — would have caught the blocker and would stop it recurring.
`GlueJsonHandlerEmptyListTest` already has the wiring to copy from.

Was that a deliberate scoping call, or just where the round ran out?

## Nits

- **Empty-body responses diverge from the rest of the file.** `CreateCrawler` (`:204`),
  `UpdateCrawler` (`:219`) and `DeleteCrawler` (`:224`) return bare `Response.ok()`, while every
  other empty response in this handler uses `Response.ok(Map.of())` — `handleTagResource` and
  `handleUntagResource` at `:761` and `:771`. AWS documents *"an empty HTTP body"* for all three
  crawler actions, so `Response.ok()` may well be the more accurate one; I'd just rather the file
  made one choice than two.
- **`NextToken` is a raw offset.** `paginate` (`GlueService.java:1229`) parses the token as an
  index into a list re-sorted on each call, so deleting a job between pages silently skips one, and
  `getJobs(null, "3")` works as a hand-written token. Common emulator shortcut and I won't push on
  it, but the PR body claims pagination parity, so it's worth knowing the gap is there.
- **Fully-qualified names throughout the new tests.**
  `io.github.hectorvent.floci.services.glue.model.Job` appears inline about thirty times where an
  import would do. The surrounding test class imports its model types.
- **Stray double blank line** in `GlueJsonHandlerEmptyListTest.java` where the two deleted tests
  were, just after `setUp`.

## Notes

- Reusing the existing `readMaxResults` / `readNextToken` / `pageResponse` helpers for `GetJobs`
  and `GetCrawlers` rather than writing new ones is the right instinct, and it means the `NextToken`
  omission-when-null behaviour is consistent with the schema-registry list actions for free.
- The `1..1000` bound in `paginate` matches the documented `MaxResults` range for both `GetJobs`
  and `GetCrawlers` exactly, and the test pins all three edges (0, -1, 1001). That's the kind of
  check I'd normally have to go look up.
- `Instant` fields on `Job` and `Crawler` carry `@JsonFormat(shape = NUMBER)`, matching `Database`
  and the rest of the Glue models. Easy thing to get wrong and it's right.
- Splitting `TagResource`/`UntagResource`/`GetTags` so registry and schema ARNs still route to
  `schemaRegistryService` while everything else goes to `ResourceGroupsTaggingService` preserves
  the existing behaviour rather than migrating it wholesale. Right call, and
  `taggingNonExistentJobOrCrawlerThrows` pins the new branch.

namanONcode:
@pgermosen all round 1 mentioned issues are resolved

pgermosen (commented):
@namanONcode — this round is a strong one. All five items from last time are actually closed: the crawler-schedule 500 is gone now that `CreateCrawlerRequest`/`UpdateCrawlerRequest` model `Schedule` as a string and the handler builds the `Schedule` struct itself, and you added handler-level tests (`createCrawlerWithScheduleSucceeds`, `updateCrawlerWithScheduleSucceeds`) that exercise exactly that path, which would have caught the original regression. The eighteen dead model classes are wired in and actually driving the handler now. Job and crawler names key off the raw value everywhere, with a comment at each site explaining why they skip the lowercase normalization the rest of the file uses. Required-field validation is in for both resources, so missing `Name`/`Role`/`Command`/`Targets` now come back as `InvalidInputException` instead of a 500. `UpdateJob` carries `SourceControlDetails` and `profileName` through instead of dropping them, and crawlers get a real `State`/`Version` now (`READY`/`1` on create, incremented on update). The smaller nits from last round — response-shape consistency, fully-qualified names in the new tests, a stray blank line — are all cleaned up too.

One thing did fall out of the tagging refactor. Before this PR, every `TagResource`/`UntagResource`/`GetTags` call went through the schema registry's resource resolver, which throws for any ARN it doesn't recognize. Now that path is narrowed to registry/schema ARNs, and everything else — jobs and crawlers, but also databases, tables, connections, anything else under `arn:aws:glue:...` — routes to `ResourceGroupsTaggingService`, which never validates that a resource exists; it just upserts or reads a tag mapping keyed by the ARN you gave it. `validateResourceExists` looks like it was meant to backstop that, but it only recognizes `job/` and `crawler/` prefixes. So tagging a nonexistent database or table ARN now returns 200 instead of `EntityNotFoundException`, and reading its tags returns an empty map instead of erroring. Jobs and crawlers themselves are fine — it's every other resource type that lost the check it used to get for free.

Separately, I want to flag rather than assert: `toDomain(UpdateCrawlerRequest)` only builds a `Schedule` when the incoming string is non-blank, so sending `"Schedule": ""` on an update is treated as "not specified" and the crawler keeps its old schedule — there's currently no way to unschedule a crawler short of deleting and recreating it. I checked both the API reference and the CLI docs for `update-crawler --schedule` and neither documents an empty-string-clears convention, so I can't tell you this is wrong, just that it's worth confirming it's what you intended.

Two small ones to close out: `CreateCrawlerResponse`, `UpdateCrawlerResponse`, and `DeleteCrawlerResponse` are still declared but never constructed — the handler correctly returns an empty body for all three per the API reference, which just means those classes have nothing left to do. And `CreateJob` doesn't apply the defaults AWS documents for fields left unspecified — `GlueVersion` defaults to `5.1`, `Timeout` to 2880 minutes (or 480 for Glue 5.0+) — so a `GetJob` right after a bare-minimum `CreateJob` echoes `null` for those instead of AWS's implicit values. Low stakes since nothing here actually executes a job, but worth knowing if you're ever asserting exact `GetJob` output against real AWS.

pgermosen (commented):
@namanONcode thanks, good progress. The defaults and dead response classes are both fixed, confirmed, and restoring EntityNotFoundException for nonexistent databases/tables/UDFs closes exactly the case I asked about.

One thing is still open though, and it's the same underlying gap, just narrower now. validateResourceExists only validates when the ARN has six or more parts and the resource segment starts with one of job/, crawler/, database/, table/, or userDefinedFunction/. Anything else falls through all the branches and returns without throwing, so it's treated as existing. Concretely: arn:aws:glue:us-east-1:123456789012:connection/my-connection has six parts and doesn't match any of those five prefixes, so TagResource would persist tags against a connection that was never created, UntagResource would report success, and GetTags would return an empty map instead of an error. Same class of bug as before, just down to the resource types this fix didn't cover rather than actually closed — probably needs either a real allow-list of every taggable Glue resource type Floci models, or a default branch that throws for anything unrecognized instead of silently passing.

namanONcode:
> @namanONcode thanks, good progress. The defaults and dead response classes are both fixed, confirmed, and restoring EntityNotFoundException for nonexistent databases/tables/UDFs closes exactly the case I asked about.
> 
> One thing is still open though, and it's the same underlying gap, just narrower now. validateResourceExists only validates when the ARN has six or more parts and the resource segment starts with one of job/, crawler/, database/, table/, or userDefinedFunction/. Anything else falls through all the branches and returns without throwing, so it's treated as existing. Concretely: arn:aws:glue:us-east-1:123456789012:connection/my-connection has six parts and doesn't match any of those five prefixes, so TagResource would persist tags against a connection that was never created, UntagResource would report success, and GetTags would return an empty map instead of an error. Same class of bug as before, just down to the resource types this fix didn't cover rather than actually closed — probably needs either a real allow-list of every taggable Glue resource type Floci models, or a default branch that throws for anything unrecognized instead of silently passing.

@pgermosen fixed the remaining issue

pgermosen (commented):
That's closed, confirmed. Re-traced my connection/ repro against the new code and it now falls through to the terminal throw instead of silently passing as existing.

One small thing, the fix only touches GlueService.java, no test was added for this specific case. Every other round of this PR paired its fix with a test, so it'd be good to keep that pattern here too, a one-line assertion that tagging an unrecognized-resource-type ARN throws EntityNotFoundException would do it. Not blocking given how simple the fix is, just want to keep the coverage consistent with the rest of the PR.

PR #2595, opened 2026-08-26
fix(ec2): round-trip launch template data instead of discarding it

allensanborn:
## Summary

`CreateLaunchTemplate` returned `200` but silently kept only `ImageId`, `InstanceType`, `KeyName`, `UserData`, `IamInstanceProfile`, `SecurityGroupIds`, and `TagSpecifications`. Every other member of `RequestLaunchTemplateData` — `BlockDeviceMappings`, `MetadataOptions`, `NetworkInterfaces`, `EbsOptimized`, `Monitoring`, `Placement`, `CreditSpecification`, `CpuOptions`, `EnclaveOptions`, `HibernationOptions`, `MaintenanceOptions`, `PrivateDnsNameOptions`, `CapacityReservationSpecification`, `DisableApiTermination`, `InstanceInitiatedShutdownBehavior`, and more — was dropped without a word. Terraform read its own input back missing and reported "Provider produced inconsistent result after apply", or diffed forever. `MetadataOptions` with `HttpTokens=required` (IMDSv2) and `BlockDeviceMappings` appear in most Gruntwork ASG and ECS modules, so this was blocking a large share of that corpus.

This PR stores and round-trips the full field set (see the commit message for the exact list and for the two related read-back contradictions it also fixes: `IamInstanceProfile` no longer gets silently rewritten from `Name` to `Arn`, and a `NetworkInterfaces` block's `Groups` are no longer hoisted into top-level `SecurityGroupIds`, which made the block vanish from state).

### On the size of this diff, and the `autoscaling` touches

This diff is larger than a typical fix (+1465/-312 across 14 files) and two of those files are in `services/autoscaling` rather than `services/ec2`: `AutoScalingReconciler.java` (+28/-13) and `AutoScalingService.java` (+1/-1). Both touches are load-bearing, not scope creep, and here is why: fixing the underlying bug required changing what `LaunchTemplate`/`LaunchTemplateVersion` store. Previously the version mirrored a handful of launch fields (`imageId`, `instanceType`, `keyName`, …) as flat fields alongside its data map, kept in sync via a `dataFrom`/`applyData` pair — a shape that has no room for the ~20 fields this PR adds without repeating the sync problem for each one. Instead, a version now holds a single `LaunchTemplateData` object as the one source of truth.

Autoscaling reads launch templates to build instance launches (`AutoScalingReconciler.launchSourceForTemplate`, `mixedInstancesInstanceType`) and validates them at `CreateAutoScalingGroup` time (`AutoScalingService.validateLaunchTemplate`). Those call sites read the flat fields that no longer exist, so removing the duplicated-state shape is a compile-time breaking change for both files. The fix in each case is a one-line accessor swap (e.g. `version.getImageId()` → `version.getData().getImageId()`, and using the new `Ec2Service.iamInstanceProfileArn(...)` / `LaunchTemplateData.effectiveSecurityGroupIds()` helpers in place of fields that used to be pre-computed) — no autoscaling behavior changes, and `AutoScalingReconcilerTest`/`AutoScalingServiceTest` are unmodified in intent, only updated to build against the new accessor. Splitting these two files into a separate PR would leave that PR non-compiling on its own, and would leave this PR unable to fix the model without also touching every consumer — so they stay together. Happy to split further if a maintainer prefers a smaller autoscaling-only follow-up, but I don't think one is separable here.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behavior: `CreateLaunchTemplate`/`CreateLaunchTemplateVersion` accepted the full `RequestLaunchTemplateData` shape but persisted only a handful of fields, so `DescribeLaunchTemplateVersions` read back a template missing everything else the caller submitted.

The AWS EC2 service model is the reference for what is stored and for the element names it's read back under: request parameter names are the ones botocore's EC2 serializer emits (note the singular `BlockDeviceMapping.N` / `NetworkInterface.N` list prefixes, taken from each list member's `locationName`), and response elements follow `ResponseLaunchTemplateData`'s `locationName`s. Verified against the field set in the current EC2 API model (botocore `ec2/2016-11-15/service-2.json`) as of this branch.

Deliberately out of scope, accepted on input and ignored rather than rejected (documented in the commit message and in `docs/services/ec2.md`): `InstanceMarketOptions`, `InstanceRequirements`, `LicenseSpecifications`, `ElasticGpuSpecifications`, `ElasticInferenceAccelerators`, `NetworkPerformanceOptions`, `Operator`, `SecondaryInterfaces`, `SecurityGroups` (by-name groups), and within `NetworkInterfaces` the IPv4/IPv6 address/prefix lists, `EnaSrdSpecification`, `ConnectionTrackingSpecification`, `PrimaryIpv6`, and `EnaQueueCount`. A focused subset that round-trips correctly beats a broad partial one.

## Checklist

- [x] `./mvnw test` passes locally (ran the affected suites directly: `Ec2LaunchTemplateFieldsIntegrationTest`, `Ec2ServiceTest`, `Ec2IntegrationTest`, `Ec2LaunchTemplateCfnProvisionerTest`, `AutoScalingReconcilerTest`, `AutoScalingServiceTest` — 328 tests, 0 failures/errors)
- [x] New or updated integration test added (`Ec2LaunchTemplateFieldsIntegrationTest`, +245 lines; `Ec2LaunchTemplateCfnProvisionerTest`, `Ec2IntegrationTest`, `Ec2ServiceTest`, `AutoScalingReconcilerTest`, `AutoScalingServiceTest` all updated for the new accessors)

### DIFF
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -493,10 +493,28 @@ Two behaviours worth calling out, because they are what Terraform reads back:
   top-level `SecurityGroupIds`; on AWS the two are mutually exclusive. A launch from the template
   resolves its security groups from whichever of the two is populated.
 
+#### Launch template versions
+
+`CreateLaunchTemplateVersion` accepts two members outside `LaunchTemplateData` itself:
+
+- **`SourceVersion` controls inheritance, and omitting it means no inheritance.** A request that
+  names a source version — including `$Latest` or `$Default` — layers its own fields onto that
+  version's data, so only the fields it restates change. A request that omits `SourceVersion`
+  entirely does **not** fall back to the latest version: the new version starts from an empty
+  `LaunchTemplateData`, populated only by whatever fields the request itself supplies. This matches
+  the AWS-documented behavior; it does not merge onto any prior version.
+- **`VersionDescription` is stored and read back by `DescribeLaunchTemplateVersions`.** It is a
+  version-level field on `LaunchTemplateVersion`, not a member of `RequestLaunchTemplateData` /
+  `ResponseLaunchTemplateData`, so it is tracked per version alongside `VersionNumber` and
+  `CreateTime` rather than inside the launch template data payload. `CreateLaunchTemplate` accepts
+  the same field for the initial version it creates.
+
 Members the service model declares that are accepted and ignored rather than stored:
 `InstanceMarketOptions`, `InstanceRequirements`, `LicenseSpecifications`, `ElasticGpuSpecifications`,
 `ElasticInferenceAccelerators`, `NetworkPerformanceOptions`, `Operator`, `SecondaryInterfaces` and
-`SecurityGroups` (security groups by name). Within `NetworkInterfaces`, the IPv4/IPv6 address and
+`SecurityGroups` (security groups by name — resolving names to IDs would need lookup machinery,
+including ambiguity handling across VPCs, that no other EC2 action here has either; `RunInstances`
+itself only accepts `SecurityGroupId`). Within `NetworkInterfaces`, the IPv4/IPv6 address and
 prefix lists, `EnaSrdSpecification`, `ConnectionTrackingSpecification`, `PrimaryIpv6` and
 `EnaQueueCount` are likewise ignored.
 
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -472,6 +472,34 @@ Route table ids follow the live API's own inconsistency: an id that does not exi
 
 Launch templates store versioned launch data. New template versions can be created from an existing source version, and `ModifyLaunchTemplate` updates the default version used by later launches.
 
+#### Launch template data members
+
+These members of `RequestLaunchTemplateData` are stored and read back unchanged by
+`DescribeLaunchTemplateVersions`:
+
+`ImageId`, `InstanceType`, `KeyName`, `UserData`, `KernelId`, `RamDiskId`, `SecurityGroupIds`,
+`IamInstanceProfile`, `BlockDeviceMappings`, `NetworkInterfaces`, `TagSpecifications`,
+`MetadataOptions`, `Monitoring`, `Placement`, `CpuOptions`, `CreditSpecification`,
+`EnclaveOptions`, `HibernationOptions`, `MaintenanceOptions`, `PrivateDnsNameOptions`,
+`CapacityReservationSpecification`, `EbsOptimized`, `DisableApiTermination`, `DisableApiStop`,
+`InstanceInitiatedShutdownBehavior`.
+
+Two behaviours worth calling out, because they are what Terraform reads back:
+
+- **`IamInstanceProfile` keeps the form it was given.** A profile submitted as `Name` reads back as
+  `Name`, not rewritten to `Arn`. The instance-profile ARN is derived at launch time instead, so
+  `aws_launch_template.iam_instance_profile.name` converges.
+- **`NetworkInterfaces` stays a `NetworkInterfaces` block.** Its `Groups` are not hoisted into
+  top-level `SecurityGroupIds`; on AWS the two are mutually exclusive. A launch from the template
+  resolves its security groups from whichever of the two is populated.
+
+Members the service model declares that are accepted and ignored rather than stored:
+`InstanceMarketOptions`, `InstanceRequirements`, `LicenseSpecifications`, `ElasticGpuSpecifications`,
+`ElasticInferenceAccelerators`, `NetworkPerformanceOptions`, `Operator`, `SecondaryInterfaces` and
+`SecurityGroups` (security groups by name). Within `NetworkInterfaces`, the IPv4/IPv6 address and
+prefix lists, `EnaSrdSpecification`, `ConnectionTrackingSpecification`, `PrimaryIpv6` and
+`EnaQueueCount` are likewise ignored.
+
 ### IAM Instance Profiles
 
 | Action | Description |
--- a/src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingReconciler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingReconciler.java
@@ -440,13 +440,13 @@ public class AutoScalingReconciler {
                     : asg.getLaunchTemplateVersion();
             return new LaunchSource(
                     null,
-                    version.getImageId(),
-                    version.getInstanceType(),
-                    version.getKeyName(),
-                    version.getSecurityGroupIds(),
-                    version.getInstanceTags(),
-                    version.getUserData(),
-                    version.getIamInstanceProfileArn(),
+                    version.getData().getImageId(),
+                    version.getData().getInstanceType(),
+                    version.getData().getKeyName(),
+                    version.getData().effectiveSecurityGroupIds(),
+                    version.getData().getInstanceTags(),
+                    version.getData().getUserData(),
+                    ec2Service.iamInstanceProfileArn(version.getData()),
                     asg.getLaunchTemplateId(),
                     asg.getLaunchTemplateName(),
                     resolvedVersion,
@@ -470,13 +470,13 @@ public class AutoScalingReconciler {
                 String instanceType = mixedInstancesInstanceType(asg, version);
                 return new LaunchSource(
                         null,
-                        version.getImageId(),
+                        version.getData().getImageId(),
                         instanceType,
-                        version.getKeyName(),
-                        version.getSecurityGroupIds(),
-                        version.getInstanceTags(),
-                        version.getUserData(),
-                        version.getIamInstanceProfileArn(),
+                        version.getData().getKeyName(),
+                        version.getData().effectiveSecurityGroupIds(),
+                        version.getData().getInstanceTags(),
+                        version.getData().getUserData(),
+                        ec2Service.iamInstanceProfileArn(version.getData()),
                         specification.getLaunchTemplateId() == null
                                 ? mixedLaunchTemplate.getLaunchTemplateId()
                                 : specification.getLaunchTemplateId(),
@@ -557,7 +557,7 @@ public class AutoScalingReconciler {
                 }
             }
         }
-        return version.getInstanceType();
+        return version.getData().getInstanceType();
     }
 
     private record LaunchSource(
--- a/src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
@@ -825,7 +825,7 @@ public class AutoScalingService {
         if (launchTemplateVersions.isEmpty()) {
             throw invalidLaunchTemplate();
         }
-        if (isBlank(launchTemplateVersions.getFirst().getImageId())) {
+        if (isBlank(launchTemplateVersions.getFirst().getData().getImageId())) {
             throw missingLaunchTemplateImageId();
         }
     }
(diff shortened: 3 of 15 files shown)


pgermosen (commented):
Thanks Allen — the diagnosis and the framing are both strong. Pre-empting the "why is
`services/autoscaling` in an EC2 fix" question in the description, with the specific reason
(the flat mirrored fields had no room for twenty more without repeating the sync problem) and
an offer to split, is exactly the right way to land a diff this size. Having read the two
files: the touches really are accessor swaps — `AutoScalingReconciler.java:443-449` is
`version.getData().getImageId()` and friends — and I agree they aren't separable.

**Build:** I don't have a `check-pr.sh` run for this, so nothing below is build-verified —
diff and AWS reference only. Given the blast radius (a model rewrite plus four consumers), I'd
want a full run before this lands rather than the targeted suites:

## What I verified

**The response element names are right, which is the detail that decides whether an SDK can
parse any of this.** `ResponseLaunchTemplateData`'s list members carry `locationName`s that
differ from the request parameter names, and `Ec2QueryHandler` emits exactly those:
`blockDeviceMappingSet` (`:3345`), `networkInterfaceSet` (`:3370`), `securityGroupIdSet`
(`:3490`), `tagSpecificationSet` (`:3498`). Getting the singular `BlockDeviceMapping.N` /
`NetworkInterface.N` prefixes on the request side and the `…Set` names on the response side is
the part that would have been easy to get subtly wrong in a hand-rolled Query serializer.

**Consolidating onto one `LaunchTemplateData` is the right cut.** The `dataFrom`/`applyData`
pair kept a handful of launch fields duplicated alongside the data map; adding twenty members
to that shape would have meant twenty more chances for the two copies to drift. One source of
truth removes the class of bug rather than one instance of it.

**Both read-back contradictions are real and worth fixing here.** `IamInstanceProfile` being
silently rewritten from `Name` to `Arn`, and a `NetworkInterfaces` block's `Groups` being
hoisted into top-level `SecurityGroupIds` so the block vanished from state, are each enough on
their own to make `aws_launch_template` never converge — and the fixes are documented in
`docs/services/ec2.md` in terms of what Terraform actually reads back, not in terms of the
internal change.

**The out-of-scope set is enumerated rather than left to discovery.** Listing
`InstanceMarketOptions`, `InstanceRequirements`, `NetworkPerformanceOptions` and the rest as
"accepted and ignored" in the docs is the difference between a known boundary and a trap. "A
focused subset that round-trips correctly beats a broad partial one" is the right instinct.

---

## (should fix) A version created without `SourceVersion` must not inherit anything

`Ec2Service.java:3846-3848`:

```java
LaunchTemplateData source = versionData(launchTemplate,
        resolveLaunchTemplateVersion(launchTemplate, sourceVersion, launchTemplate.getLatestVersionNumber()));
LaunchTemplateData merged = source.mergedWith(data != null ? data : new LaunchTemplateData());
```

When `SourceVersion` is absent it falls back to the latest version, so every new version is
merged onto its predecessor. The reference is explicit that this is the opposite of the rule:

> **SourceVersion** — If you specify this parameter, the new version inherits the launch
> parameters from the source version. If you specify additional launch parameters for the new
> version, they overwrite any corresponding launch parameters inherited from the source version.
> **If you omit this parameter, the new version contains only the launch parameters that you
> specify for the new version.**

```
create-launch-template          --launch-template-data '{"ImageId":"ami-1","SecurityGroupIds":["sg-a"]}'
create-launch-template-version  --launch-template-name X --launch-template-data '{"ImageId":"ami-2"}'
describe-launch-template-versions --versions 2

AWS:    launchTemplateData = { imageId: ami-2 }
Floci:  launchTemplateData = { imageId: ami-2, securityGroupIdSet: [ sg-a ] }
```

This is the same bug the PR is fixing, inverted: instead of the caller's fields going missing,
fields the caller never sent turn up. And because `mergedWith` treats absent-or-empty as
"inherit", there is no request that can produce a version *without* a member once any earlier
version had it — so Terraform removing a `metadata_options` or `block_device_mappings` block
produces "Provider produced inconsistent result after apply" again, for removals rather than
additions. That's a real regression risk against the corpus you're targeting, since
`aws_launch_template` sends the full data on every change and expects the response to match it
exactly.

`mergedWith` itself is right — override-wins-when-present is the correct rule for the
`SourceVersion` case. It's the fallback argument that needs to go: merge only when
`sourceVersion` is non-blank, and otherwise store `data` as given.

## (should fix) `VersionDescription` is accepted and dropped

Neither `Ec2QueryHandler` nor `LaunchTemplate` mentions it — it isn't parsed, isn't stored,
and isn't emitted. AWS's own sample response for this operation carries it:

```xml
<launchTemplateVersion>
    ...
    <versionDescription>VersionWithNewAMI</versionDescription>
    <versionNumber>4</versionNumber>
</launchTemplateVersion>
```

It's a modelled request parameter on both `CreateLaunchTemplate` and
`CreateLaunchTemplateVersion` (String, length 0–255), and `aws_launch_template.description`
maps straight onto it — so a Terraform config that sets `description` writes it, reads back
nothing, and diffs forever. Precisely the failure mode in your summary.

It also isn't in the documented ignore list, which is understandable: that section enumerates
`RequestLaunchTemplateData` members, and `VersionDescription` is a sibling parameter one level
up, so it fell outside the sweep. Either store it or add it to the ignored set — but it
shouldn't be the one dropped field with no record.

## (question) `SecurityGroups` (by name) is in the ignored list, but it isn't like the others

The rest of that list — `InstanceRequirements`, `InstanceMarketOptions`,
`ElasticGpuSpecifications`, `NetworkPerformanceOptions` — are features Floci doesn't model at
all, so ignoring them is honest. `SecurityGroups` is different in kind: it's a launch parameter
that a caller can set today, that has a first-class Terraform argument
(`aws_launch_template.security_groups`, alongside `vpc_security_group_ids`), and that will
round-trip empty and diff forever in exactly the way this PR exists to stop.

I'm not asking you to add it here — the scoping argument holds and the list is documented. I'd
just like to know whether it's a deliberate "EC2-Classic-era, not worth it" call or an
oversight in the sweep, because if it's the former a half-sentence in `ec2.md` saying so would
stop the next person re-raising it, and if it's the latter it belongs with the stored set
rather than the ignored one.

## Notes

- `ClientToken` and `ResolveAlias` are also unhandled. `ClientToken` only matters if Floci
  models idempotency (it doesn't, here or elsewhere in EC2), and `ResolveAlias` only applies to
  SSM-parameter `ImageId` values, which are out of scope anyway — so neither is worth acting
  on. Mentioning them so the absence is deliberate rather than unexamined.
- `launchTemplate.setData(new LaunchTemplateData(merged))` keeps the template's own `data`
  pointing at the *latest* version. That's consistent with how the flat fields behaved before,
  and `DefaultVersionNumber` is tracked separately, so I don't think anything reads the wrong
  one today — but it's worth a comment, since "the template's data" and "the default version's
  data" are now different things that will diverge the moment someone calls
  `ModifyLaunchTemplate` to move the default backwards.
- The docs section is written from the reader's side — "two behaviours worth calling out,
  because they are what Terraform reads back" — rather than as a changelog of the refactor.
  That's the right register for a compatibility page.

The core of this is sound and the scoping judgement is good throughout. The `SourceVersion`
fallback is the one I'd want changed before it lands, because it reintroduces the same
read-back mismatch from the other direction.

pgermosen (commented):
@allensanborn — the amended push adds something I missed in the first round and should have
caught, but neither of the two should-fixes I raised is addressed, and there's no note saying
why.

## What the amend added, and it's a good catch

Collapsing `LaunchTemplate` onto a single `LaunchTemplateData` silently breaks every template
already on disk: the old shape stored `iamInstanceProfileArn` and `instanceTags` as flat fields,
`@JsonIgnoreProperties(ignoreUnknown = true)` drops what it doesn't recognise, and a restart
would hand back a template with no IAM profile and no instance tags. That's data loss on
upgrade, from a change whose whole point is that fields stop disappearing — and I didn't spot it
when I reviewed the model rewrite.

The legacy `@JsonSetter` mappings are the right fix, and `legacyIamProfileAndInstanceTagsSurviveRestart`
is the right test:

- it hand-builds the pre-migration JSON rather than trying to produce it from code that no
  longer exists, which is the only way to test a migration once the old writer is gone;
- it covers **both** paths — the `versions` map entry that `DescribeLaunchTemplateVersions` and
  AutoScaling read, and the top-level fields that `ensureLaunchTemplateVersions` falls back to
  for a template with an empty versions map;
- and the javadoc says what breaks without it, so nobody deletes the setters as dead code.

## Still outstanding from the first round

Neither of these changed, and the force-push carried no reply, so I can't tell whether they're
deferred, disagreed with, or simply missed.

**1. A version created without `SourceVersion` still inherits from the latest one.**
`Ec2Service.java:3847` still passes `launchTemplate.getLatestVersionNumber()` as the fallback:

```java
resolveLaunchTemplateVersion(launchTemplate, sourceVersion, launchTemplate.getLatestVersionNumber())
```

> **SourceVersion** — … If you omit this parameter, the new version contains only the launch
> parameters that you specify for the new version.

So a v2 created with only `ImageId` still comes back carrying v1's security groups, and no
request can produce a version that *drops* a member. For a PR about launch template data
round-tripping faithfully, that's the same mismatch from the other direction — and it's the one
that reintroduces "inconsistent result after apply" when a Terraform config removes a block
rather than adding one.

**2. `VersionDescription` is still accepted and dropped.** Zero occurrences in
`Ec2QueryHandler` — not parsed, not stored, not emitted. It's a modelled parameter on both
create operations, it appears in AWS's own sample response as `<versionDescription>`, and
`aws_launch_template.description` maps onto it.

The open question from round one — whether `SecurityGroups` by name is a deliberate
"EC2-Classic-era, not worth it" call or an oversight — is also still open.

None of these is large. If you'd rather land the migration fix now and take the rest as a
follow-up, say so and I'll review it on that basis; what I can't do is assume it either way from
a silent force-push.

pgermosen (approved):
Both should-fixes from the last round are resolved correctly and thoroughly, and the open question from round one now has a considered answer.

`SourceVersion` inheritance is fixed exactly as specced: `createLaunchTemplateVersion` no longer falls back to the latest version when `sourceVersion` is null or blank — it now starts from a fresh, empty `LaunchTemplateData` and merges only what the request itself supplies, matching the documented behavior verbatim ("if you omit this parameter, the new version contains only the launch parameters that you specify for the new version"). Both the unit test and the integration test assert the negative case directly — that fields from version 1 do *not* leak into a version created without `SourceVersion` — which is the case that actually would have caught the original bug, rather than only checking that the new field showed up.

`VersionDescription` round-trips, and the design choice is right: storing it in a parallel map keyed by version number, rather than folding it into `LaunchTemplateData`, matches the model correctly — `VersionDescription` is a member of `LaunchTemplateVersion` itself, not of the request/response data shape, so putting it inside the data object would have been the wrong place even though it would have round-tripped for a single version. The test correctly checks that describing an older version specifically still returns that version's own description after a newer version with a different one exists, which is the case that would have broken if everything read from a single mirrored field instead of the per-version map.

The `SecurityGroups`-by-name question is now answered rather than just left open: the docs now give an actual reason (resolving names to IDs needs lookup machinery — store scan, not-found handling, cross-VPC ambiguity — that no other EC2 action here has either, including `RunInstances` itself), and the added test confirms the request still succeeds with a by-name group present and that `SecurityGroupIds` is unaffected, proving "accepted and ignored" rather than just asserting it in a comment.

One process note, not a code issue: the fix landed via another push with no reply on this thread, same as last round. Worth a one-line comment on future rounds, not because anything here is wrong, but so a later reader doesn't have to reconstruct "was this addressed or disagreed with" from the commit history the way I did this time.

Approving.

PR #2601, opened 2026-08-27
feat(aps): add Amazon Managed Service for Prometheus workspaces

matanbaruch:
Adds Amazon Managed Service for Prometheus (`aps`) as a new service, scoped to the workspace lifecycle that Terraform's and Pulumi's `aws_prometheus_workspace` resource drives: `CreateWorkspace`, `DescribeWorkspace`, `ListWorkspaces`, `UpdateWorkspaceAlias`, `DeleteWorkspace`, plus tagging through the shared `/tags/{resourceArn}` dispatcher. Before this, `CreateWorkspace` had no service to answer it at all.

Wire behavior, matching the AWS restJson1 model and the provider's waiters ([`internal/service/amp/workspace.go`](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/amp/workspace.go)):

1. A workspace is ACTIVE from birth. Real AMP answers the create `202` with `CREATING`; the emulator provisions nothing, so the provider's create waiter (`CREATING` to `ACTIVE`) completes on its first `DescribeWorkspace` poll.
2. `DeleteWorkspace` (`202`) removes the workspace immediately; a later `DescribeWorkspace` returns `ResourceNotFoundException` `404`, which the provider's delete waiter matches as a typed exception to treat the workspace as gone.
3. `DescribeWorkspace` returns the full description including `prometheusEndpoint` in the real AWS shape; the `ListWorkspaces` summary omits it, per the `WorkspaceSummary` shape.
4. `createdAt` goes out as restJson1 epoch-seconds. An ISO string here fails deserialization inside the AWS SDKs, so the compatibility test asserts it through the real Java SDK.
5. The `alias` parameter of `ListWorkspaces` is a prefix filter (the provider exposes it as `alias_prefix`), with `maxResults`/`nextToken` pagination and `ValidationException` on out-of-range values.
6. Tagging routes through `SharedTagsController` via a `TagHandler` with service key `aps`, with the success statuses overridden to `200` per the AMP model.

Scope notes (kept deliberately narrow):

1. Alert manager definitions, rule groups namespaces, scrapers and logging configurations are not implemented. The workspace surface is what IaC providers need to manage the resource.
2. `prometheusEndpoint` is reported in the real AWS shape but is not a live ingestion/query endpoint.
3. `clientToken` is accepted and ignored: creates are not deduplicated.
4. `kmsKeyArn` is stored and echoed back; no encryption is performed.

Tests:

1. Unit (`ApsServiceTest`): create/describe/list/delete/alias semantics, alias-prefix filtering, pagination and its `ValidationException`, and the TagHandler ARN round trip.
2. REST integration (`ApsControllerIntegrationTest`): full lifecycle over HTTP including the `X-Amzn-Errortype: ResourceNotFoundException` header after delete (what the provider's waiter matches), the `202`/`204` response codes, and tags through `/tags/{arn}`.
3. SDK compatibility (`compatibility-tests` `AmpTest`, new `software.amazon.awssdk:amp` dependency): the same lifecycle through the real AWS Java SDK v2, which proves the timestamp and error wire shapes end to end.

Docs: `docs/services/managed-prometheus.md` plus an index row, following the hand-written style of the Control Tower page (no switch handler, so the generated action tables are unaffected).

### DIFF
--- a/compatibility-tests/sdk-test-java/pom.xml
+++ b/compatibility-tests/sdk-test-java/pom.xml
@@ -78,6 +78,10 @@
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
+            <artifactId>amp</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
             <artifactId>sqs</artifactId>
         </dependency>
         <dependency>
--- a/compatibility-tests/sdk-test-java/src/main/java/com/floci/test/TestFixtures.java
+++ b/compatibility-tests/sdk-test-java/src/main/java/com/floci/test/TestFixtures.java
@@ -60,6 +60,7 @@ import software.amazon.awssdk.services.apigatewayv2.ApiGatewayV2Client;
 import software.amazon.awssdk.services.elasticache.ElastiCacheClient;
 import software.amazon.awssdk.services.elasticbeanstalk.ElasticBeanstalkClient;
 import software.amazon.awssdk.services.acm.AcmClient;
+import software.amazon.awssdk.services.amp.AmpClient;
 import software.amazon.awssdk.services.efs.EfsClient;
 import software.amazon.awssdk.services.ec2.Ec2Client;
 import software.amazon.awssdk.services.ecr.EcrClient;
@@ -508,6 +509,14 @@ public final class TestFixtures {
 
     public static SecretsManagerClient secretsManagerClient() {
         return SecretsManagerClient.builder()
+                .endpointOverride(ENDPOINT)
+                .region(REGION)
+                .credentialsProvider(CREDENTIALS)
+                .build();
+    }
+
+    public static AmpClient ampClient() {
+        return AmpClient.builder()
                 .endpointOverride(ENDPOINT)
                 .region(REGION)
                 .credentialsProvider(CREDENTIALS)
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AmpTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AmpTest.java
@@ -1,0 +1,127 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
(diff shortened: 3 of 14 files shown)


pgermosen (commented):
@matanbaruch — thanks for this, and welcome. The code side looks well-formed for a first
contribution: unit tests, an HTTP integration test, an SDK compatibility test and a docs page all
arrived together, which is more than most first PRs bring.

There is a plumbing problem with the PR itself, though, and it needs recreating rather than
fixing in place.

## CI never started, and it is not the first-time-contributor gate

The welcome bot above says your checks are waiting for a maintainer to approve them. That is the
normal case for a first PR, but it is not what happened here — there is no queued run for a
maintainer to release. Looking at the workflow runs for this branch:

- **Neither of your commits has a `pull_request` run at all.** Not queued, not awaiting approval,
  not skipped — none was ever created. The first-time-contributor gate would have created a run
  sitting in `action_required`, and there isn't one.
- The only runs that exist are the two `pull_request_target` workflows (the welcome message and
  the PR-limit notice), and those fired **only when the PR was opened**. When you pushed the
  second commit (`ffb929ec`, the `RegionResolver`/`AwsArnUtils` refactor), nothing ran at all —
  not even those two.

So the PR stopped receiving pull-request events after it was created. Nothing on your side caused
that and nothing in your code is wrong; it is a delivery problem with this particular PR object.

## What to do

Close this PR and open a fresh one from the same branch. Reopening fires the pull-request events
again from scratch, which is what gets the workflows created. Your commits and branch are fine as
they are — no need to rewrite anything, just a new PR pointing at the same
`feature/ACDO-0/aps-workspaces`.

If a new PR still shows no `Build and Test` checks after a minute or two, say so here and we will
dig into it from the repository side rather than have you keep retrying.

Once CI is actually running I will review the APS implementation properly — at a glance the shape
looks right, and I would rather give you real feedback against a green build than a partial read
against no signal at all.

matanbaruch:
Closing to recreate per the review. Same branch, commits unchanged.

matanbaruch:
@pgermosen recreated as #2604, same symptom: zero pull_request runs on head ffb929ec after ~5 min, only the two pull_request_target bots fired on open.

PR #2608, opened 2026-08-27
fix(s3): return x-amz-version-id from presigned POST

### LINKED ISSUE #2607: [BUG] S3: presigned POST does not return x-amz-version-id on a versioned bucket
## Service

S3

## AWS API Action

POST Object (presigned POST / browser form upload), against a versioning-enabled bucket

## Expected behavior

When versioning is enabled, S3 generates a version ID for the object created by a POST and returns it to the caller in the `x-amz-version-id` response header. A browser upload has no other way to learn it — the SDK is not involved, so there is no response object to read, only the HTTP response.

This matters for the common direct-to-S3 upload pattern: the browser uploads through a presigned POST, reads `x-amz-version-id`, and reports it back to the application so the version can be recorded against the uploaded file. Without the header, that flow cannot proceed.

## Actual behavior

The upload succeeds and a new version *is* created, but the response carries no `x-amz-version-id`, so the caller cannot discover it.

Every other write path in Floci returns the header correctly — including a presigned PUT over plain HTTP, which rules out both "versioning isn't working" and "the header can't be sent over HTTP". Only the presigned POST omits it:

```
floci/floci:1.7.0
bucket versioning: Enabled

[OK ] PutObject (SDK)
        version reported to caller : e49634c8-798e-4ee9-b031-ca7c6bc2edd5
        version actually created   : e49634c8-798e-4ee9-b031-ca7c6bc2edd5
[OK ] Presigned PUT (HTTP)
        version reported to caller : 1e8cb334-b82f-46fc-bee3-ec857a0255b4
        version actually created   : 1e8cb334-b82f-46fc-bee3-ec857a0255b4
        response headers           : ['Date', 'ETag', 'content-length', 'x-amz-checksum-crc64nvme',
                                      'x-amz-id-2', 'x-amz-request-id', 'x-amz-storage-class',
                                      'x-amz-version-id', 'x-amzn-RequestId']
[GAP] Presigned POST (HTTP)
        version reported to caller : (none)
        version actually created   : 2b82f46c-3b33-40f2-8895-23f82608f374
        response headers           : ['Date', 'ETag', 'Location', 'x-amz-id-2',
                                      'x-amz-request-id', 'x-amzn-RequestId']
```

"version actually created" is taken by diffing `ListObjectVersions` either side of each write, so the object genuinely gets a version in all three cases.

## Reproduction

```python
"""Needs boto3, requests and testcontainers-floci."""

import os

import boto3
import requests
from floci import FlociContainer

BUCKET = "repro"


def versions(client, key):
    listed = client.list_object_versions(Bucket=BUCKET, Prefix=key)
    return {v["VersionId"] for v in listed.get("Versions", [])}


container = FlociContainer(image="floci/floci:1.7.0")
container.start()
os.environ.update({
    "AWS_ENDPOINT_URL_S3": f"http://127.0.0.1:{container.get_exposed_port(4566)}",
    "AWS_ACCESS_KEY_ID": "test",
    "AWS_SECRET_ACCESS_KEY": "test",
    "AWS_DEFAULT_REGION": "us-east-1",
})
s3 = boto3.client("s3")
s3.put_bucket_versioning(Bucket=BUCKET, VersioningConfiguration={"Status": "Enabled"})

key = "org/1/presigned-post"
post = s3.generate_presigned_post(
    Bucket=BUCKET,
    Key=key,
    Fields={"Content-Type": "text/plain"},
    Conditions=[{"Content-Type": "text/plain"}, ["content-length-range", 1, 64]],
    ExpiresIn=3600,
)

before = versions(s3, key)
response = requests.post(
    post["url"],
    data=post["fields"],
    files={"file": ("a.txt", b"hello", "text/plain")},
    timeout=15,
)

print("status                :", response.status_code)          # 204
print("x-amz-version-id      :", response.headers.get("x-amz-version-id"))  # None
print("version created anyway:", (versions(s3, key) - before).pop())        # a real version id

container.stop()
```

## Where it comes from

`src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java`, in the presigned POST handler:

```java
S3Object obj = s3Service.putObject(bucket, key, fileData, objectContentType,
        metadata.isEmpty() ? null : metadata);
...
return Response.status(204)
        .header("ETag", obj.getETag())
        .header("Location", bucket + "/" + key)
        .build();
```

`obj` already carries the version; the handler just never reads `obj.getVersionId()`. The rest of the file establishes the convention, e.g. in `putObject`:

```java
var resp = Response.ok().header("ETag", obj.getETag());
if (obj.getVersionId() != null) {
    resp.header("x-amz-version-id", obj.getVersionId());
}
```

That guarded form appears in seven places (PutObject, CopyObject, CompleteMultipartUpload and others). The presigned POST response looks like the one writer that was missed.

### Possibly related, in the same method

The `String xml = new XmlBuilder()...` building a `PostResponse` document immediately above the return is never used — the status is hardcoded to 204 and the document discarded. On real S3 that document is the 201 response body when the policy sets `success_action_status`. I checked, and `success_action_status: "201"` currently still returns 204 with no body. Happy to raise that separately if you'd prefer it kept apart from this.

## Environment

- Floci version / image tag: `floci/floci:1.7.0` (also reproduced on `nightly`)
- SDK version: boto3 1.43.0 / botocore 1.43.0 (Python 3.12)
- How you're running Floci: Docker, via `testcontainers-floci` 0.1.1

rickyrattlesnake:
## Summary

A presigned POST to a versioning-enabled bucket creates a new object version, but the response omits `x-amz-version-id`, so the uploader cannot discover the version it just created. Unlike every other write path there is no SDK response object to fall back on — a browser form upload only ever sees the HTTP response — so the version is simply unavailable to the caller.

This breaks the ordinary direct-to-S3 upload pattern, where the browser uploads through a presigned POST, reads `x-amz-version-id`, and reports it back to the application so the version can be recorded against the file.

`doHandlePresignedPost` already holds the `S3Object` returned by `putObject`; it just never read `getVersionId()`. The fix sets the header behind the same null guard the rest of `S3Controller` uses (`putObject`, `copyObject`, `completeMultipartUpload` and others all do this).

Closes #2607

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior:** on a versioned bucket, `POST Object` returned `204` with only `ETag` and `Location`. Real S3 also returns `x-amz-version-id` for the version the POST created.

Verified against `floci/floci:1.7.0` with boto3 1.43.0 / botocore 1.43.0, comparing three ways of writing the same object to the same versioned bucket. Only the POST path was affected, which is what pointed at the handler rather than at versioning or the HTTP layer:

```
[OK ] PutObject (SDK)          version reported: e49634c8-…   version created: e49634c8-…
[OK ] Presigned PUT (HTTP)     version reported: 1e8cb334-…   version created: 1e8cb334-…
[GAP] Presigned POST (HTTP)    version reported: (none)       version created: 2b82f46c-…
```

"version created" is taken by diffing `ListObjectVersions` either side of each write. The full reproduction script is in #2607.

## Checklist

On the first box — I ran the S3 suite rather than the whole thing, since the full run wants Docker-backed services (Lambda, RDS, ElastiCache) that my environment could not provide, and failures there would have said more about my setup than about this change:

```
./mvnw test -Dtest='S3*Test'   →  990 tests, 0 failures, 0 errors
```

The new test, `S3PresignedPostIntegrationTest#presignedPostReturnsVersionIdOnAVersionedBucket`, was checked both ways: it fails on `main` without the change (`Expected: is not null`) and passes with it. It asserts the header is present and that it matches what a subsequent `HeadObject` reports, so it pins the value rather than just its existence.

## One thing I noticed nearby, not changed here

In the same method, the `String xml = new XmlBuilder()…` building a `PostResponse` document is never used — the status is hardcoded to `204` and the document discarded. On real S3 that document is the `201` body when the policy sets `success_action_status`, and I confirmed Floci currently returns `204` with no body in that case. I left it alone to keep this PR to one behaviour; happy to open a separate issue, or fold it in here if you would rather have both together.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -2711,10 +2711,13 @@ public class S3Controller {
                 .elem("ETag", obj.getETag())
                 .end("PostResponse")
                 .build();
-        return Response.status(204)
+        var response = Response.status(204)
                 .header("ETag", obj.getETag())
-                .header("Location", bucket + "/" + key)
-                .build();
+                .header("Location", bucket + "/" + key);
+        if (obj.getVersionId() != null) {
+            response.header("x-amz-version-id", obj.getVersionId());
+        }
+        return response.build();
     }
 
     private void validatePolicyConditions(String policyBase64, String bucket,
--- a/src/test/java/io/github/hectorvent/floci/services/s3/S3PresignedPostIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/s3/S3PresignedPostIntegrationTest.java
@@ -21,6 +21,7 @@ import static org.hamcrest.Matchers.*;
 class S3PresignedPostIntegrationTest {
 
     private static final String BUCKET = "presigned-post-bucket";
+    private static final String VERSIONED_BUCKET = "presigned-post-versioned-bucket";
     private static final DateTimeFormatter AMZ_DATE_FORMAT =
             DateTimeFormatter.ofPattern("yyyyMMdd'T'HHmmss'Z'").withZone(ZoneOffset.UTC);
 
@@ -563,5 +564,52 @@ class S3PresignedPostIntegrationTest {
                   ]
                 }
                 """.formatted(expiration, bucket, keyPrefix, contentTypePrefix, minSize, maxSize);
+    }
+
+    @Test
+    @Order(110)
+    void presignedPostReturnsVersionIdOnAVersionedBucket() {
+        given()
+        .when()
+            .put("/" + VERSIONED_BUCKET)
+        .then()
+            .statusCode(200);
+
+        given()
+            .body("<VersioningConfiguration><Status>Enabled</Status></VersioningConfiguration>")
+        .when()
+            .put("/" + VERSIONED_BUCKET + "?versioning")
+        .then()
+            .statusCode(200);
+
+        String key = "uploads/versioned.txt";
+        String contentType = "text/plain";
+        String policy = buildPolicy(VERSIONED_BUCKET, key, contentType, 0, 10485760);
+        String policyBase64 = Base64.getEncoder().encodeToString(policy.getBytes(StandardCharsets.UTF_8));
+
+        String versionId =
+            given()
+                .multiPart("key", key)
+                .multiPart("Content-Type", contentType)
+                .multiPart("policy", policyBase64)
+                .multiPart("x-amz-algorithm", "AWS4-HMAC-SHA256")
+                .multiPart("x-amz-date", AMZ_DATE_FORMAT.format(Instant.now()))
+                .multiPart("file", "versioned.txt",
+                        "Hello from a versioned presigned POST!".getBytes(StandardCharsets.UTF_8), contentType)
+            .when()
+                .post("/" + VERSIONED_BUCKET)
+            .then()
+                .statusCode(204)
+                .extract()
+                .header("x-amz-version-id");
+
+        assertThat(versionId, is(notNullValue()));
+
+        given()
+        .when()
+            .head("/" + VERSIONED_BUCKET + "/" + key)
+        .then()
+            .statusCode(200)
+            .header("x-amz-version-id", is(versionId));
     }
 }


pgermosen (approved):
Clean, minimal fix. `doHandlePresignedPost` now sets `x-amz-version-id` on the 204 response whenever `obj.getVersionId()` is non-null, which is the exact same guard pattern used at every other call site in `S3Controller.java` — checked all eight, including `fullObjectResponse`, `putObject`, `copyObject`, and `completeMultipartUpload`. The mutable-`ResponseBuilder` idiom here also matches the existing style elsewhere in the file rather than introducing something new.

Cross-checked against AWS's docs directly: versioning-enabled buckets return `x-amz-version-id` for `PUT`, `POST`, or `CopyObject` alike, so `POST` being the one path missing it was a real gap, and this closes exactly that gap.

The new test is solid — it pins the actual version value with a follow-up `HeadObject` rather than just asserting the header exists, and the PR body confirms it fails on `main` and passes with the change. I also agree with leaving the unused `PostResponse` XML / hardcoded status-code question alone here; that's a separate, real gap but a good candidate for its own issue rather than folding into this one.

Two minor things, neither blocking: there's a pre-existing, unrelated duplicate `@Order(94)` on two other test methods in the same file.

Approving.

PR #2619, opened 2026-08-27
fix(elb): route Classic (v1) requests to a dedicated handler instead of ELBv2

allensanborn:
## Summary

Classic ELB (`elasticloadbalancing` 2012-06-01) and ELBv2/ALB (2015-12-01) share an endpoint host and credential scope, but every Classic request — `CreateLoadBalancer`, `RegisterInstancesWithLoadBalancer`, `DescribeInstanceHealth`, and the rest — was being routed to the ELBv2 handler. A well-formed Classic request sending `LoadBalancerName` was rejected with `ValidationError: Name is required for load balancer`, and any request the v2 handler did accept came back shaped as a 2015-12-01 ALB object with an ARN, which Classic has no concept of. This blocks Terraform's `aws_elb` resource outright, and any Auto Scaling group that fronts instances with a Classic load balancer — a common pattern in the Gruntwork examples (`terraform-aws-asg`, `terraform-aws-ecs`, `terraform-aws-monitoring`), where it blocks at least 10 example roots across those repos.

This PR adds a dedicated Classic ELB handler and routes on the API version. Every Query-protocol request carries the version it speaks as the `Version` form parameter, and both service models declare it, so that's the discriminator: `2012-06-01` to the new Classic handler, `2015-12-01` to ELBv2 — the same in-case split the `rds` scope already uses to separate RDS, Neptune, and DocumentDB. Requests that omit `Version` fall back to their parameter shape: an action unique to one API, else `LoadBalancerName`/`LoadBalancerNames` selecting Classic, else ELBv2 as before.

Classic state is modeled on its own terms — keyed by name, no ARN, inline listeners, one structured attribute record — and responses follow the 2012-06-01 wrapper and element names (`CreateLoadBalancerResult` carries `DNSName` and nothing else). Health checking reuses the same counters/thresholds as `ElbV2HealthChecker`, driven off the single `HealthCheck.Target` string (`HTTP:`/`HTTPS:` probed with a request, `TCP:`/`SSL:` by connecting). An Auto Scaling group naming Classic load balancers in `LoadBalancerNames` now registers and deregisters its instances with them, so a `min_elb_capacity` wait can be satisfied. Stickiness, backend-server-policy operations, and `SetLoadBalancerListenerSSLCertificate` answer `UnsupportedOperation` in the 2012-06-01 namespace rather than fabricate a result; there is no Classic data plane, so listener ports are recorded but no socket is opened.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behavior: `Action=CreateLoadBalancer&Version=2012-06-01&LoadBalancerName=x` — a well-formed Classic request — was answered by the v2 handler, either with a `ValidationError: Name is required for load balancer` (the handler wanted `Name`, Classic sent `LoadBalancerName`) or, for requests the v2 handler could parse, with a 2015-12-01 ALB-shaped response instead of the 2012-06-01 `DNSName`-only document Classic clients expect.

Verified against the Gruntwork corpus: `terraform-aws-asg/examples/asg-instance-refresh/with-elb`, `asg-rolling-deploy/with-elb`, and `server-group/with-elb` all fail apply against unpatched Floci with the exact `ValidationError: Name is required for load balancer` signature — the error text is itself proof of the routing bug, since the provider is emitting exactly what the Classic API declares. The same signature recurs across 10 roots in 3 repos (`terraform-aws-asg`, `terraform-aws-ecs`, `terraform-aws-monitoring`).

## Checklist

- [x] `./mvnw test` passes locally (targeted suites touched by this diff: `EmulatorLifecycleTest` 23, `ElbClassicHealthCheckTargetTest` 5, `AutoScalingReconcilerTest` 12, `ElbClassicIntegrationTest` 18, `ElbClassicServiceTest` 13 — 71/71, 0 failures, 0 errors; full-suite run not performed here)
- [x] New or updated integration test added (`ElbClassicIntegrationTest`, 18 tests covering create/describe/delete, listeners, health checks, instance registration, tagging, and the version/parameter-shape routing discriminator itself)

### DIFF
--- a/docs/services/elb-classic.md
+++ b/docs/services/elb-classic.md
@@ -1,0 +1,109 @@
+# Elastic Load Balancing (Classic, v1)
+
+**Protocol:** Query (XML) — `POST http://localhost:4566/` with `Action=` and `Version=2012-06-01`
+
+Classic Elastic Load Balancing is the 2012-06-01 API behind `aws elb` and Terraform's `aws_elb`
+resource. It is a **different API** from [ELB v2](elb.md) (ALB/NLB, 2015-12-01), even though both
+are served from the same endpoint host and signed with the same `elasticloadbalancing` credential
+scope.
+
+## How Floci tells the two apart
+
+Every Query-protocol request carries the API version it is speaking as the `Version` form
+parameter, and that is what Floci routes on:
+
+| `Version` | Answered by |
+|---|---|
+| `2012-06-01` | Classic (this page) |
+| `2015-12-01` | [ELB v2](elb.md) |
+
+Responses are emitted in the matching namespace —
+`http://elasticloadbalancing.amazonaws.com/doc/2012-06-01/` for Classic. A Classic
+`CreateLoadBalancer` returns `DNSName` and nothing else; Classic load balancers have no ARN and
+are addressed by `LoadBalancerName` for their whole life.
+
+If a hand-rolled client omits `Version`, Floci falls back to the request shape: an action unique
+to one API decides, and for the action names both APIs define (`CreateLoadBalancer`,
+`DescribeLoadBalancers`, `AddTags`, …) the presence of `LoadBalancerName` or `LoadBalancerNames`
+selects Classic. With neither present the request goes to ELB v2.
+
+## Supported Actions
+
+| Action | Description |
+|--------|-------------|
+| CreateLoadBalancer | Creates a Classic load balancer with listeners, subnets or AZs, security groups, scheme, and tags. Returns `DNSName`. |
+| DescribeLoadBalancers | Returns `LoadBalancerDescription` records, optionally filtered by `LoadBalancerNames`. |
+| DeleteLoadBalancer | Deletes a load balancer; deleting one that does not exist succeeds. |
+| CreateLoadBalancerListeners | Adds listeners. A conflicting listener on a port already in use is rejected with `DuplicateListener`. |
+| DeleteLoadBalancerListeners | Removes listeners by `LoadBalancerPorts`. |
+| ConfigureHealthCheck | Replaces the health check and re-arms Floci's checker. |
+| RegisterInstancesWithLoadBalancer | Registers EC2 instances and returns the full registered set. |
+| DeregisterInstancesFromLoadBalancer | Removes instances and returns the remaining set. |
+| DescribeInstanceHealth | Returns `InstanceStates` for the named instances, or for every registered instance. |
+| ModifyLoadBalancerAttributes | Updates cross-zone, access log, connection draining, connection settings, and additional attributes. Unsent members keep their current value. |
+| DescribeLoadBalancerAttributes | Returns the full attribute structure. |
+| ApplySecurityGroupsToLoadBalancer | Replaces the security groups. |
+| AttachLoadBalancerToSubnets / DetachLoadBalancerFromSubnets | Adds or removes subnets and recomputes the availability zones and VPC. |
+| EnableAvailabilityZonesForLoadBalancer / DisableAvailabilityZonesForLoadBalancer | Adds or removes availability zones. |
+| AddTags / RemoveTags / DescribeTags | Tagging, keyed by `LoadBalancerNames`. |
+| DescribeAccountLimits | Returns standard default Classic limits. |
+
+## Not implemented
+
+These Classic operations return `UnsupportedOperation` in the 2012-06-01 namespace rather than a
+fabricated result:
+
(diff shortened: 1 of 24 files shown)


pgermosen (commented):
A well-designed fix for a real bug, with one concrete wire-fidelity miss and a couple of cosmetic nits.

The premise is real and the routing discriminator is the right one: Classic and ELBv2 sharing an endpoint host and credential scope with no way to tell them apart by service name is exactly the situation `rds` already has with Neptune and DocumentDB, and reusing that same version-first, parameter-shape-fallback pattern is the right call rather than inventing a new mechanism. The one blind spot in the fallback — `DescribeAccountLimits` has no load-balancer-name parameter in either API, so a `Version`-less call defaults to ELBv2 — is real but harmless, since every actual AWS client always sends `Version` on a Query-protocol request; it only matters for a hand-rolled client that already opted out of the one signal that would tell Floci which API it's speaking.

I checked several of the modeled codes and defaults directly against the reference rather than taking them on trust: `DuplicateLoadBalancerName` and `SubnetNotFound` are both exactly right, the default health check (`TCP:80`, interval 30, timeout 5, unhealthy 2, healthy 10) matches AWS's documented default, and `CreateLoadBalancer`'s response really is `DNSName` and nothing else. The Auto Scaling integration is solid too — I specifically re-verified the stale-instance deregistration path that an earlier bot round flagged, and it's genuinely fixed: `removeStaleInstances` now deregisters from both target groups and Classic load balancers before pruning the ASG record, not just cosmetically patched.

Where I'd push back: `singleVpcId` throws `InvalidConfigurationRequest` at HTTP 400 when subnets span more than one VPC, but `CreateLoadBalancer`'s own Errors table documents that exact error at 409 — confirmed against the API reference. This is the one place where the error code string is right but the status isn't, and it matters more than it looks: a client that retries on 409 but fails hard on 400 gets the wrong signal here.

Two smaller things: the empty-`Listeners` validation error message reads like it was copied from an unrelated check ("Only one of SubnetIds or AvailabilityZones may be empty..." for a missing-listener error) — the code and status are fine, just the wording is off. And the new cross-reference to the Classic docs page in `elb.md` breaks the original sentence mid-paragraph rather than being appended after it — purely cosmetic.

allensanborn:
Thanks for checking the codes and defaults against the reference rather than taking them on trust — and for re-verifying the stale-instance deregistration path specifically, since that was the one an earlier bot round only half-described.

All three points addressed in a92b62b, on top of a rebase onto current `main` (the branch had gone conflicting against `EmulatorLifecycleTest`, where upstream and this branch each added a service-config mock in the same block; both are kept).

**`InvalidConfigurationRequest` at 409, not 400** — you're right, and the reasoning for why it matters is right too: the code string was already correct, so the status is the only thing distinguishing a retryable configuration conflict from a hard client error, and it was the half that was wrong. Fixed in `singleVpcId`, with a unit test that asserts the status rather than only the code — that was the gap that let it through in the first place.

**The empty-`Listeners` message** — copied verbatim from the subnet/AZ check, exactly as it looked. Now "At least one listener must be specified.", with a test asserting the message actually names the missing member.

**The `elb.md` cross-reference** — moved to follow the opening sentence instead of splitting it.

On `DescribeAccountLimits`: agreed, and I'd rather leave it than add a heuristic. It's only reachable by a client that omits `Version` on a Query-protocol request, which no AWS SDK or CLI does, and any tiebreaker I could invent for that case would be guessing at an intent the request doesn't express. The version-first, parameter-shape-fallback order means every real client is decided by `Version` before shape is ever consulted.

ELB Classic + ELBv2 + the reconciler and lifecycle tests are green locally (160 tests), and `make docs-check` is clean.

pgermosen (approved):
@allensanborn thanks for the follow-through on this — I verified all three fixes independently rather than taking the writeup on trust. The 409 is right (pulled the AWS reference myself, `InvalidConfigurationRequest` is documented at 409 in `CreateLoadBalancer`'s Errors table), and the new test asserts the actual HTTP status rather than just the error code string, which is what let the wrong status slip through the first time. The message and docs fixes match what you described too.

I also traced `isElbClassicRequest` myself since it's the mechanism the whole PR rests on — version-first, then the unique action names, then the `LoadBalancerName`/`LoadBalancerNames` shape fallback for the actions both APIs share. It holds up, and the one blind spot (`DescribeAccountLimits` with no `Version`) is genuinely harmless the way you two already worked out — no real client omits `Version` on a Query-protocol request.

Approving.

PR #2620, opened 2026-08-27
fix(ec2): canonicalize IPv4 route destinations at the API boundary

allensanborn:
## Summary
AWS documents that a DestinationCidrBlock is reduced to its canonical form on input
(e.g. 100.68.0.18/18 becomes 100.68.0.0/18). Without that step, CreateRoute,
ReplaceRoute and DeleteRoute could each be handed a differently-spelled but
equivalent CIDR for the same network and silently disagree about which route it
named — two spellings of one /18 could coexist as distinct routes, or a
ReplaceRoute/DeleteRoute using an equivalent spelling could miss the stored route
entirely.

- Adds `CidrCanonicalizer`, a shared IPv4/IPv6 CIDR canonicalization utility in
  `core.common`.
- Wires `Ec2Service`'s CreateRoute/ReplaceRoute/DeleteRoute to canonicalize
  DestinationCidrBlock once, before it's matched against existing routes or stored.
  DescribeRouteTables then echoes the canonical form for free.
- DestinationIpv6CidrBlock is deliberately left untouched: AWS's canonical-form
  language appears only under DestinationCidrBlock in the API reference, not under
  DestinationIpv6CidrBlock, so this stays IPv4-only even though the new utility
  also supports IPv6. DestinationPrefixListId is an opaque ID, not a CIDR, and is
  likewise untouched.
- Follow-up to #2546 (RouteAlreadyExists duplicate-destination rejection), which
  this builds on: canonicalizing on input is what makes "same destination"
  detection actually catch equivalent spellings instead of just literal repeats.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility
Real EC2 canonicalizes an IPv4 route destination CIDR on input; Floci previously
stored it verbatim. This makes stored/returned state match AWS's canonical form,
and makes duplicate-destination detection correct for equivalent spellings.

## Checklist
- [x] `CidrCanonicalizerTest` — 38/38 pass
- [x] `Ec2RouteCidrCanonicalizationTest` — 6/6 pass (canonicalization on
      Create/Replace/Delete, duplicate-detection across spellings, IPv6 left
      untouched)
- [x] Full `services.ec2` suite — 517/517 pass, 0 failures/errors

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/CidrCanonicalizer.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/CidrCanonicalizer.java
@@ -1,0 +1,204 @@
+package io.github.hectorvent.floci.core.common;
+
+import java.net.Inet4Address;
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.util.Optional;
+
+/**
+ * Canonicalises IPv4 and IPv6 CIDR blocks the way AWS does on write: the address is
+ * rewritten so every bit outside the prefix is zero, then re-rendered in AWS's own
+ * textual form (dotted-decimal for IPv4, compressed form for IPv6).
+ *
+ * <p>{@code 100.68.0.18/18} becomes {@code 100.68.0.0/18} — AWS's own documented
+ * example — because bits 18-31 of the address are host bits and AWS zeroes them
+ * before storing the block. A CIDR that is already canonical is returned unchanged
+ * (the identical string, not merely an equal one).
+ *
+ * <p>This class draws a hard line between "malformed" and "valid but non-canonical":
+ * {@link #canonicalize(String)} returns {@link Optional#empty()} only for a CIDR that
+ * cannot be parsed at all, never as a stand-in for "already canonical". Callers that
+ * need to tell the two apart before rewriting anything can call
+ * {@link #isCanonical(String)} directly.
+ *
+ * <p>Pure function: no I/O, no DNS resolution. The address half of the CIDR is parsed
+ * as a literal via {@link InetAddress#ofLiteral(String)}, which never triggers a
+ * hostname lookup — critical here, because {@link InetAddress#getByName(String)} would
+ * silently turn a malformed CIDR into a DNS query.
+ */
+public final class CidrCanonicalizer {
+
+    private CidrCanonicalizer() {}
+
+    /**
+     * Parses {@code cidr}, zeroes its host bits, and re-renders it in canonical form.
+     *
+     * @param cidr an IPv4 or IPv6 CIDR block, e.g. {@code "100.68.0.18/18"} or
+     *             {@code "2001:db8::1/32"}
+     * @return the canonical CIDR string, or {@link Optional#empty()} if {@code cidr} is
+     *     null, blank, missing a prefix, has a prefix out of range for the address
+     *     family (0-32 for IPv4, 0-128 for IPv6), or has an address half that is not a
+     *     literal IPv4/IPv6 address
+     */
+    public static Optional<String> canonicalize(String cidr) {
+        return parse(cidr).map(parsed -> {
+            String rendered = render(parsed);
+            // Preserve reference identity when the input was already canonical, so
+            // callers get the exact same String back rather than a merely-equal copy.
+            return rendered.equals(cidr) ? cidr : rendered;
+        });
+    }
+
+    /**
+     * Reports whether {@code cidr} is both parseable and already in canonical form,
+     * i.e. {@code canonicalize(cidr)} would return the identical string.
+     *
+     * @return {@code true} only when {@code cidr} is valid and already canonical;
+     *     {@code false} both when it is malformed and when it is valid but would be
+     *     rewritten — use {@link #canonicalize(String)} to distinguish those two cases
+     */
+    public static boolean isCanonical(String cidr) {
+        return parse(cidr).map(parsed -> render(parsed).equals(cidr)).orElse(false);
+    }
+
+    private static Optional<ParsedCidr> parse(String cidr) {
+        if (cidr == null || cidr.isBlank()) {
+            return Optional.empty();
+        }
+
+        int slash = cidr.lastIndexOf('/');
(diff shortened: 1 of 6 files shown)


allensanborn:
> 👋 Thanks for the pull request!
> 
> It looks like you now have **6 open pull requests** in this repository, which is above our advisory limit of **5**. To keep the review backlog focused and make sure every contribution gets proper attention, we ask contributors to shepherd their existing PRs to the finish line before opening new ones — responding to review feedback, rebasing, or closing any that are no longer needed.
> 
> Here is [your list of open PRs](https://github.com/floci-io/floci/pulls?q=is%3Apr+is%3Aopen+author%3Aallensanborn). Nothing is blocked and this PR stays open — this is just a friendly nudge. The `over-pr-limit` label helps maintainers see at a glance which PRs to prioritize.
> 
> If some of these PRs are stacked or intentionally related, just say so here and a maintainer will take that into account.

Sorry about that... busy agents. I have a large backlog of bugs I'd like to get in. If there are any suggestions on how I can work through them more efficiently, please let me know.

pgermosen (approved):
Thanks Allen — this is a correct fix in an area that is easy to get subtly wrong, and the bit
arithmetic holds up. I read `CidrCanonicalizer` closely rather than trusting the tests, and traced
the cases most likely to break it.

## What I verified

**The IPv4-only scoping is exactly right, and for the reason you gave.** The canonicalization
sentence really does appear under one member and not the others:

> **DestinationCidrBlock** — The IPv4 CIDR address block used for the destination match. … **We
> modify the specified CIDR block to its canonical form; for example, if you specify
> `100.68.0.18/18`, we modify it to `100.68.0.0/18`.**
>
> **DestinationIpv6CidrBlock** — The IPv6 CIDR block used for the destination match. Routing
> decisions are based on the most specific match.

No canonical-form language on the IPv6 member, and `DestinationPrefixListId` is an ID. Building a
utility that handles IPv6 and then deliberately not applying it is the right call — and
`ipv6DestinationsAreNotCanonicalized` pins the restraint so nobody "completes" it later.

**The RFC 5952 compression is correct on the cases that usually break it.** I traced
`2001:db8::`, `::`, `::1`, `1::`, and the two run-selection cases: `1:0:0:1:0:0:0:1` correctly
elides the *longer* second run, and a tie like `1:0:0:1:0:0:1:1` keeps the leftmost because
`runLength > bestLength` is strict. The `i <= 8` loop bound is what lets a run that reaches the end
of the address close properly, and the two-group minimum matches §4.2.2.

**The masking is right at both ends of the range.** `prefixLength == 0` skips the partial-byte
branch entirely and zeroes everything, so `0.0.0.0/0` — the default route, and the value most
likely to be exercised — never shifts by 32 or divides by zero. `/32` and `/128` take the same
skip and leave the address untouched. The partial-byte mask `0xFF << (8 - remainderBits) & 0xFF`
is only reached with `remainderBits` in 1–7, so the shift is always in range; Java's precedence
makes that `(0xFF << n) & 0xFF`, which is the intended mask.

**`InetAddress.ofLiteral` over `getByName` is the right primitive** and the javadoc says why. A
malformed address half going to DNS would be a genuinely nasty bug in a pure function, and it's the
kind of thing that only shows up as an unexplained delay in someone's test suite.

**Greptile's legacy-route finding is properly closed.** Canonicalizing in
`setDestinationCidrBlock` rather than only in the constructor is the right hook: `Route` is a plain
JavaBean with private fields, a public setter and no field-level annotations or `@JsonCreator`, so
Jackson binds through that setter and a route persisted before this change normalizes on read.
`legacyNonCanonicalRouteDestinationCanonicalizesOnRestart` hand-writes a pre-fix
`ec2-route-tables.json` to prove it, and the javadoc states exactly which assertions fail if the
setter change is reverted — which is what makes it a regression test rather than a test that
happens to pass.

## (minor) The IPv4 guard now exists in two places

`Ec2Service.canonicalizeIpv4Cidr` and `Route.setDestinationCidrBlock` implement the same rule:
null/blank/contains-`:` passes through untouched, everything else goes through
`CidrCanonicalizer.canonicalize(...).orElse(input)`. They agree today — I checked the `isSet` vs
`isBlank` difference and it can't diverge on any input, since a whitespace-only value fails
`parse`'s own blank check and falls back the same way.

But the entire property this PR establishes is that the boundary and the storage layer agree on
what canonical means, and it now rests on two independently-maintained copies of that decision.
Someone extending IPv6 support, or loosening the `:` heuristic, changes one and the invariant goes
quiet rather than failing.

A `CidrCanonicalizer.canonicalizeIpv4(String)` — the IPv4-only guard plus the passthrough, in the
utility that already owns the parsing — would give both callers one definition to share. It would
also let `Ec2Service` drop its private helper's javadoc duplication, which currently restates the
IPv6 reasoning that `Route` also restates.

Related and smaller: now that the constructor canonicalizes, passing
`canonicalDestinationCidrBlock` into `new Route(...)` is belt-and-braces. Still needed for the
`matchesDestination` comparisons, so the local is not removable — just noting the storage half is
now redundant with the setter.

## Notes

- `CidrCanonicalizer.isCanonical` has no caller in this PR. It's documented, tested, and a
  reasonable thing for a shared utility to expose, so I'm not asking you to drop it — noting it
  only because a public method that nothing calls tends to drift.
- The utility landing in `core.common` rather than under `services.ec2` is the right home given VPC
  and subnet CIDR blocks are the obvious next callers. Worth knowing that this PR deliberately
  doesn't touch them: a VPC created with a non-canonical `CidrBlock` still stores it verbatim,
  while its local route's destination now canonicalizes, so the two can disagree in stored state.
  Out of scope here and correctly so — flagging it as the natural follow-up rather than a gap in
  this change.
- `alreadyCanonicalIpv4_isIdentityNotJustEqual` pins the reference-identity behaviour the javadoc
  promises. That is a strange thing to guarantee in most code and a sensible one here, since it
  makes "did this get rewritten?" answerable with `==` at a call site that cares.

PR #2621, opened 2026-08-27
fix(ec2): report a reachable EIP address, and run gzipped user-data

allensanborn:
## Summary

An EC2 instance with an associated Elastic IP was unreachable at the address the API itself reported. `PublicIpAddress` on launch was hardcoded to `127.0.0.1` on the strength of a comment calling it "host-reachable," but Docker publishes the guest's SSH on a high host port, so port 22 at that address is nothing. `AllocateAddress` then invented a `54.x.x.x`-shaped address that routes nowhere, and that fictional value is exactly what Terraform surfaces as `aws_eip.x.public_ip` and what Terratest's SSH helpers dial. This blocks any Terratest suite that provisions an instance behind an EIP and connects to it — a common pattern across the Gruntwork corpus (observed across 11 direct + 3 shadow roots in 6 repos).

This PR makes the address Terraform reads back one that actually accepts a connection. Reachability of the container network from Floci's own API clients is detected, never assumed, with a throwaway TCP connect toward the container network (a refusal is proof of a route, since only the container's network stack could send the RST). Where the network is routable — native Linux Docker, OrbStack — the container's own IP is used, so port 22 is really port 22 with no mapping to translate; where it isn't (Docker Desktop, containers behind a VM), the existing `127.0.0.1` + published-port behavior is unchanged. `AssociateAddress` re-points the EIP at whatever address is currently reachable rather than the invented one — the `AllocationId`, `AssociationId`, and `domain` are untouched, but the EIP's public IP can change on association, which is a deliberate, documented deviation from AWS. `DescribeAddresses` re-resolves the same way, since Docker hands out a different bridge IP after a stop/start and that's the call Terraform refreshes `public_ip` from.

A second, independent bug in the same launch path is fixed alongside it: gzipped cloud-init user-data (`data.cloudinit_config { gzip = true }`, the documented way to pass multipart user-data) was silently dropped. Only a literal `#!` or a plaintext MIME boundary was recognized, so the instance booted with none of its intended provisioning and gave no signal why. `UserData` is now unwrapped through gzip, base64, and base64-of-gzip before the existing shellscript/MIME handling, with a WARN when no executable parts are found.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behavior: `DescribeInstances` reported `PublicIpAddress: 127.0.0.1` for every instance regardless of whether anything was actually reachable there, and an allocated/associated EIP reported a `54.x.x.x`-shaped address that was never backed by any listener — so any client that read the address out of the API and dialed it (Terratest's SSH helpers do exactly this) got a dead connection. Separately, gzipped user-data was accepted by the API and then discarded rather than executed.

Verified with the new/updated unit and integration coverage: `ContainerNetworkReachabilityTest` asserts the routability probe against both a reachable and an unreachable target; `Ec2ContainerManagerTest` and `Ec2ServiceTest` cover EIP re-pointing on associate, `DescribeAddresses` re-resolution after the container IP moves, disassociation handing the allocation back, and gzip/base64/base64-gzip user-data unwrapping. No standalone CLI/probe check exists for this bead (`checks/` has none for floci-xcp) — correctness relies on this test suite plus the `terratest_runner.py` shadow-failure roots recorded on the bead.

## Checklist

- [x] `./mvnw test` passes locally (targeted suites touched by this diff: `ContainerNetworkReachabilityTest` 8, `Ec2ContainerManagerTest` 38, `Ec2ServiceTest` 117 — 163/163, 0 failures, 0 errors; full-suite run not performed here)
- [x] New or updated integration test added (`Ec2ServiceTest` +3 tests covering EIP re-pointing on associate/describe/disassociate; `ContainerNetworkReachabilityTest` new, 8 tests covering the routability probe)

### DIFF
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -122,7 +122,7 @@ curl -s -H "x-aws-ec2-metadata-token: $TOKEN" \
 | `GET /latest/meta-data/ami-id` | Image ID |
 | `GET /latest/meta-data/instance-type` | Instance type |
 | `GET /latest/meta-data/local-ipv4` | Private IP |
-| `GET /latest/meta-data/public-ipv4` | Public IP (`127.0.0.1`) |
+| `GET /latest/meta-data/public-ipv4` | Public IP — the container IP where that is routable from the host, `127.0.0.1` otherwise |
 | `GET /latest/meta-data/public-hostname` | Public hostname |
 | `GET /latest/meta-data/local-hostname` | Private DNS name |
 | `GET /latest/meta-data/hostname` | Private DNS name |
@@ -444,6 +444,13 @@ Route table ids follow the live API's own inconsistency: an id that does not exi
 | DisassociateAddress | Removes an Elastic IP address association. |
 | ReleaseAddress | Releases an Elastic IP address record. |
 
+An allocated Elastic IP's `54.x.x.x` address is invented and routes nowhere, so associating
+it re-points it at the address the instance is actually reachable on — the same value
+DescribeInstances reports. This is a deliberate deviation from AWS, where an EIP's public IP
+is fixed from allocation: without it, `aws_eip.x.public_ip` hands every caller a dead address.
+The allocation ID, association ID and domain are unaffected, and disassociating restores the
+allocated address.
+
 ### Availability Zones & Regions
 
 | Action | Description |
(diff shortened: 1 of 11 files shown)


pgermosen (approved):
A genuinely careful fix for a real, high-value bug, with meticulous engineering on the reachability probe and the EIP re-pointing.

The reachability probe's core trick is sound: using a refused connection as proof of a route (an RST can only come from the target's own network stack) rather than requiring something to be listening is the right primitive, and it works before anything in the guest has even started. The Docker Desktop veto is necessary rather than defensive theater — a containerized Floci's probe proves container-to-container reachability, a different vantage point from the host where Terraform/Terratest actually run, and there's a test exercising exactly the case where the probe would otherwise lie.

The EIP re-pointing is a defensible, well-disclosed deviation rather than a hack: real AWS keeps an EIP's address fixed because it owns the network, Floci can't, so changing the address on association while keeping `AllocationId`/`AssociationId`/`domain` untouched is the least-bad option, and it's stated as a deviation in both the code and the docs rather than left to be discovered. `DescribeAddresses` re-resolving on every read is unusual for a lookup to do, but the reasoning holds given that's the exact API Terraform refreshes `public_ip` from.

The gzip-bomb protections were built in response to two real bot findings, and I verified both fixes land where they're supposed to rather than taking the commit messages' word for it: a 10 MB incremental cap inside the decompression read loop, and a 4-permit semaphore bounding how many payloads can be actively expanding at once, backed by a test that drives real concurrent decompressions through the public entry point.

One thing worth a follow-up rather than holding this up: the semaphore bounds the decompression step itself, but the permit releases before the result gets converted to a string and regex-scanned for shell scripts, both of which allocate comparably large objects with no throttling — so aggregate memory across many concurrent launches is narrowed rather than fully bounded. I don't think it's a blocker: the failure mode is a heap-exhaustion crash of the local Floci process, recoverable by a restart, not a compromise or a shared-service impact; it needs an unusual combination of many concurrent launches each with a near-10MB payload, when real cloud-init scripts are typically kilobytes; and Floci's realistic threat model here is a developer running Terratest against their own local emulator, not an adversary submitting crafted payloads to a shared service. Worth extending the permit's hold to cover the full decode pipeline, or using a byte-budget semaphore instead of a flat count, as a follow-up.

Approving.

PR #2626, opened 2026-08-27
 feat(s3): apply VersioningConfiguration in CloudFormation S3 bucket provisioning

### LINKED ISSUE #2610: [FEAT] CloudFormation: apply S3 Bucket VersioningConfiguration
## Service

S3 (via CloudFormation)

## API Action / Feature

`AWS::S3::Bucket` `VersioningConfiguration` property support in CloudFormation provisioning.

## AWS Documentation

- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-versioningconfig.html

## Why is this needed?

Floci's S3 service already supports bucket versioning (`PutBucketVersioning` / `GetBucketVersioning`,
S3Service.java:1228 and the `PUT /?versioning` endpoint), but the CloudFormation provisioner
(`CloudFormationResourceProvisioner#provisionS3Bucket`) creates the bucket and silently ignores the
`VersioningConfiguration` property — it currently only applies `BucketName` and `CorsConfiguration`.

As a result, a stack (or CDK app using `versioned: true`) that defines a bucket with
`VersioningConfiguration: {Status: Enabled}` comes up with versioning still disabled, so
`aws s3api get-bucket-versioning` returns empty even though the synthesized template says Enabled —
diverging from real AWS behavior.

### Expected

- On stack create, `VersioningConfiguration.Status` is applied via the existing `putBucketVersioningConfiguration`.
- On stack update, the versioning status is reconciled to match the template.

- [x] Yes
- [ ] No

taka727:
## Summary

- `CloudFormationResourceProvisioner#provisionS3Bucket` now applies `VersioningConfiguration.Status`
  via the existing `S3Service#putBucketVersioning`, on both stack create and update.
- Property is absent → versioning is left untouched (no forced disable), matching real AWS behavior.

Closes #2610

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Floci's S3 service already implements `PutBucketVersioning`/`GetBucketVersioning` correctly
(existing `?versioning` subresource); this PR only wires the existing CloudFormation
`VersioningConfiguration` property into that already-tested code path — no new wire-protocol
surface is introduced. Verified via Floci's own REST-assured integration tests
(`CreateStack`/`UpdateStack` → `GET /{bucket}?versioning`), following the same pattern already
used for `CorsConfiguration`. Not separately verified against the real AWS CLI/SDK.

## Checklist

  - `CloudFormationResourceProvisioner#provisionS3Bucket` now applies `VersioningConfiguration.Status`
    via

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -677,6 +677,7 @@ public class CloudFormationResourceProvisioner {
         }
         s3Service.createBucket(bucketName, region);
         applyBucketCorsConfiguration(bucketName, props, engine);
+        applyBucketVersioningConfiguration(bucketName, props, engine);
         r.setPhysicalId(bucketName);
         r.getAttributes().put("Arn", AwsArnUtils.Arn.of("s3", "", "", bucketName).toString());
         r.getAttributes().put("DomainName", bucketName + ".s3.amazonaws.com");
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Small, clean, correctly-scoped wiring fix. This reuses the existing, already-validated `putBucketVersioning` rather than duplicating logic — it already rejects anything other than `Enabled`/`Suspended` with a 400, so an invalid Status value in a template surfaces as a normal stack failure with a reason rather than an unhandled exception.

The absent-property behavior matches real AWS, and for the right reason: S3 versioning has no "never versioned" state to fall back to once it's been touched, only Enabled/Suspended, so a template that never mentions `VersioningConfiguration` correctly leaves the bucket's setting alone instead of forcing a state. The dedicated test for that case asserts on the absence of a `<Status>` element, which is the right way to check "untouched" rather than "some particular value."

I also confirmed create and update both go through the same provisioning path rather than assuming it because the description says so — traced the single dispatch point for `AWS::S3::Bucket` and it's shared, the same as the existing CORS feature this mirrors. The update test (Enabled to Suspended across an UpdateStack) is a real reconciliation test, not two independent creates side by side.

Approving.

Issue #1744, opened 2026-07-06
[BUG] RDS ListTagsForResource returns DBInstanceNotFound for DB subnet group ARNs (subgrp:*)


bedardpl:
## Service

RDS

## AWS API Action

`ListTagsForResource` when the resource ARN is a DB subnet group (`arn:...:subgrp:<name>`)

## Expected behavior

`ListTagsForResource` should accept any taggable RDS resource ARN, including DB subnet groups (`rds:*:*:subgrp:*`), per the RDS API — this is also what the Terraform/OpenTofu AWS provider calls immediately after `CreateDBSubnetGroup` to populate `tags_all`, so `aws_db_subnet_group` cannot be created via Terraform/OpenTofu against Floci at all right now.

## Actual behavior

`ListTagsForResource` treats any `subgrp:` ARN as a DB instance lookup and returns `DBInstanceNotFound`, even for Floci's own pre-seeded `default` subnet group (not something specific to a freshly created group):

```
An error occurred (DBInstanceNotFound) when calling the ListTagsForResource operation: DB instance arn:aws:rds:us-east-1:000000000000:subgrp:default not found.
```

## Reproduction

```bash
export AWS_ENDPOINT_URL=http://localhost:4566
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1

# Works fine, subnet group is created
aws rds create-db-subnet-group \
  --db-subnet-group-name my-subnet-group \
  --db-subnet-group-description test \
  --subnet-ids subnet-xxxx subnet-yyyy

# 404s even though the subnet group above (or Floci's own pre-seeded "default" group) exists
aws rds list-tags-for-resource \
  --resource-name "arn:aws:rds:us-east-1:000000000000:subgrp:default"
```

This blocks `terraform apply` / `tofu apply` for any config using `aws_db_subnet_group` (a prerequisite for `aws_db_instance` via `db_subnet_group_name`), since the AWS provider's create path for that resource always does a post-create tags read.

## Environment

- Floci version / image tag: 1.5.30 (`floci/floci:latest`)
- Java SDK version (if applicable): N/A, hit via AWS CLI v2 and the OpenTofu `hashicorp/aws` provider (~> 5.0)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (docker compose), real mode (not `FLOCI_SERVICES_RDS_MOCK`)

[...]

PR #2627, opened 2026-08-27
fix(rds): keep the tags given to CreateDBSubnetGroup

### LINKED ISSUE #2612: [BUG] rds: CreateDBSubnetGroup drops the Tags it is given
## Service

RDS

## API Action / Feature

`CreateDBSubnetGroup` with `Tags.Tag.N` → `ListTagsForResource`, `DescribeDBSubnetGroups`

## What happens

Tags passed on create are not stored:

```console
$ aws rds create-db-subnet-group --db-subnet-group-name g1 --db-subnet-group-description d \
    --subnet-ids "$S1" "$S2" --tags Key=Name,Value=g1 Key=env,Value=tst
$ aws rds list-tags-for-resource --resource-name arn:aws:rds:eu-central-1:123456789012:subgrp:g1
{
    "TagList": []
}
```

A live account returns both tags, and includes them in `DescribeDBSubnetGroups`.

`DbSubnetGroup` already carries a `tags` map and `RdsService` has a `subgrp` tag handle for `ListTagsForResource`/`AddTagsToResource`; it is `handleCreateDbSubnetGroup` that never reads `Tags.Tag.N` and calls `createDbSubnetGroup(name, description, subnetIds, region)` without them. `dbSubnetGroupXml` does not write a `TagList` either.

DocumentDB subnet groups behave the same way (`docdb create-db-subnet-group --tags …` → empty list).

## Why it matters

terraform-provider-aws reads tags immediately after create, so `aws_db_subnet_group` (and `aws_docdb_subnet_group`) shows `+ tags = { Name = … }` on every plan.

## Fix

Parse `Tags.Tag.N` in the create handler and pass them through, the way `CreateDBParameterGroup` does since #2408; emit `TagList` in the describe XML. Same for the DocumentDB subnet group handler.

## Related

#1744 (closed via #1649) made the `subgrp:` ARN accepted by `ListTagsForResource`; this is the remaining half.

ck-michals:
Closes #2612

## What

`CreateDBSubnetGroup` never read `Tags.Tag.N`, so a group created with tags answered an empty list
to `ListTagsForResource` — which terraform-provider-aws calls right after the create — although
`DbSubnetGroup` already carried a tag map and `AddTagsToResource` on the group's ARN worked (#1744/#1649).

- `RdsService.createDbSubnetGroup` gains a `tags` parameter and stores them with the group in the same
  write, so there is no window where the group exists untagged (the parameter-group path from #2408
  tags in a second call; this one does not need to).
- The handler passes the parsed tags through. DocumentDB subnet groups are the same records reached
  through the same `rds` scope, so `docdb create-db-subnet-group --tags` is covered by the same change.

`DescribeDBSubnetGroups` is unchanged: AWS's `DBSubnetGroup` shape has no `TagList`, so
`ListTagsForResource` is the only read path — the issue text saying otherwise was wrong.

## Checked against a live account

`create-db-subnet-group --tags Key=Name,Value=g1 Key=env,Value=tst` followed by
`list-tags-for-resource` on the `subgrp:` ARN returns both tags on the `rds` and `docdb` CLIs; Floci
returned `[]` and now returns both.

## Tests

- `RdsQueryHandlerTest.createDbSubnetGroup_passesCreateTagsToService`
- `RdsServiceTest.createDbSubnetGroupStoresTheTagsItIsGiven`: readable back, a later `AddTagsToResource`
  merges onto them, and they live on the stored record (fails with the store dropped:
  `expected {Name=tagged, env=tst} but was {}`)
- `rds.bats`: the CLI round trip on both endpoints, against the packaged jar

## Not covered

- CloudFormation `AWS::RDS::DBSubnetGroup` `Tags` are still dropped: that path is in the legacy
  `CloudFormationResourceProvisioner`, which AGENTS.md asks not to extend.

### DIFF
--- a/compatibility-tests/sdk-test-awscli/test/rds.bats
+++ b/compatibility-tests/sdk-test-awscli/test/rds.bats
@@ -192,3 +192,30 @@ teardown() {
 
     aws_cmd docdb delete-db-cluster --db-cluster-identifier "$CLUSTER_ID" --skip-final-snapshot >/dev/null 2>&1 || true
 }
+
+@test "rds: tags given to create-db-subnet-group are readable back" {
+    VPC_ID=$(aws_cmd ec2 create-vpc --cidr-block 10.77.0.0/16 --query Vpc.VpcId --output text)
+    S1=$(aws_cmd ec2 create-subnet --vpc-id "$VPC_ID" --cidr-block 10.77.1.0/24 --availability-zone us-east-1a --query Subnet.SubnetId --output text)
+    S2=$(aws_cmd ec2 create-subnet --vpc-id "$VPC_ID" --cidr-block 10.77.2.0/24 --availability-zone us-east-1b --query Subnet.SubnetId --output text)
+    GROUP="bats-sng-$(unique_name)"
+
+    run aws_cmd rds create-db-subnet-group --db-subnet-group-name "$GROUP" --db-subnet-group-description d \
+        --subnet-ids "$S1" "$S2" --tags "Key=Name,Value=$GROUP" Key=env,Value=tst
+    assert_success
+    arn=$(json_get "$output" '.DBSubnetGroup.DBSubnetGroupArn')
+
+    run aws_cmd rds list-tags-for-resource --resource-name "$arn"
+    assert_success
+    assert_output --partial '"Key": "env"'
+    assert_output --partial "\"Value\": \"$GROUP\""
+
+    # the DocumentDB CLI reaches the same group through the same scope
+    run aws_cmd docdb list-tags-for-resource --resource-name "$arn"
+    assert_success
+    assert_output --partial '"Key": "env"'
+
+    aws_cmd rds delete-db-subnet-group --db-subnet-group-name "$GROUP" >/dev/null 2>&1 || true
+    aws_cmd ec2 delete-subnet --subnet-id "$S1" >/dev/null 2>&1 || true
+    aws_cmd ec2 delete-subnet --subnet-id "$S2" >/dev/null 2>&1 || true
+    aws_cmd ec2 delete-vpc --vpc-id "$VPC_ID" >/dev/null 2>&1 || true
+}
--- a/docs/services/rds.md
+++ b/docs/services/rds.md
@@ -19,7 +19,7 @@ RDS Data API (`rds-data`) is documented separately because it uses REST JSON rou
 | `ModifyDBInstance` | Update instance settings |
 | `RebootDBInstance` | Restart a database instance |
 | `DescribeOrderableDBInstanceOptions` | List deterministic instance class options |
-| `CreateDBSubnetGroup` | Create a DB subnet group |
+| `CreateDBSubnetGroup` | Create a DB subnet group; tags given here are readable through `ListTagsForResource` |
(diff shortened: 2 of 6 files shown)


pgermosen (approved):
Thanks — small, correct, and the part I'd normally have to go and check is already checked. Nothing
outstanding.

## The bug is real and the fix is the minimal one

`CreateDBSubnetGroup` models the parameter:

> **Tags.Tag.N** — Tags to assign to the DB subnet group. Type: Array of Tag objects. Required: No

and the handler never read it, so a modelled input was silently dropped. Passing `parseTags(params)`
through and storing it in the same write is the whole change — and doing it in the same write rather
than as a follow-up call is what makes it correct rather than merely working. There is no window in
which the group exists untagged, which the parameter-group path from #2408 can't say.

The four-argument `createDbSubnetGroup` is kept and delegates with `Map.of()`, so no existing caller
changes behaviour; the test churn is the mechanical signature update and nothing else.

## You're right that the issue text was wrong

This is the claim I most wanted to check, since it's the one where you contradicted the reporter:

> `DescribeDBSubnetGroups` is unchanged: AWS's `DBSubnetGroup` shape has no `TagList`, so
> `ListTagsForResource` is the only read path — the issue text saying otherwise was wrong.

Confirmed. The `DBSubnetGroup` data type has exactly seven members — `DBSubnetGroupArn`,
`DBSubnetGroupDescription`, `DBSubnetGroupName`, `SubnetGroupStatus`, `Subnets`,
`SupportedNetworkTypes`, `VpcId` — and no tag member of any name.

Worth saying why the reporter's assumption was a reasonable one to make: `DBInstance` and
`DBCluster` *do* carry `TagList`, so "the describe response should show the tags" holds for most of
RDS and happens not to for this shape. Leaving `DescribeDBSubnetGroups` alone is right, and the
one-line note you added to the docs table is the right place to record that
`ListTagsForResource` is the read path.

## The tests

`createDbSubnetGroupStoresTheTagsItIsGiven` asserts three separate things, and the second and third
are the ones that make it worth having:

- the tags read back through `ListTagsForResource`;
- a later `AddTagsToResource` **merges onto** them rather than replacing them, which is the
  behaviour a Terraform apply-then-modify sequence actually depends on;
- they live on the stored record (`getDbSubnetGroup(...).getTags()`) rather than in a side table, so
  they survive a restart.

Quoting the failure from the reverted run — `expected {Name=tagged, env=tst} but was {}` — is what
tells me the test exercises the fix rather than passing alongside it.

The bats case covering both the `rds` and `docdb` CLIs against the same ARN is the right addition
too, since the shared-record claim in the description is exactly the sort of thing that's true right
up until someone splits the stores.

## Note

- RDS tags carry key/value length limits and a per-resource count cap that nothing here validates.
  That's pre-existing — `AddTagsToResource` doesn't check them either — so this PR isn't the place
  for it, and adding validation on the create path alone would make the two entry points disagree.
  Noting it as a whole-service follow-up rather than anything to do here.
- Being explicit that CloudFormation `AWS::RDS::DBSubnetGroup` tags still drop, and why you didn't
  touch it, is the right call — that's the disclosure that stops this reading as a complete fix when
  one path remains.

PR #2630, opened 2026-08-27
fix(cloudformation): resolve IAM assume role policy intrinsics

### LINKED ISSUE #2538: CFN: IamRoleCfnProvisioner stores AssumeRolePolicyDocument without resolving intrinsics
Found while addressing review feedback on #2526, and deliberately kept out of that PR to stay in scope.

## Problem

`IamRoleCfnProvisioner.provision` reads the trust policy straight off the raw template node (`IamRoleCfnProvisioner.java:62`):

```java
String assumeDoc = props != null && props.has("AssumeRolePolicyDocument")
        ? props.get("AssumeRolePolicyDocument").toString()
        : "{\"Version\":\"2012-10-17\",\"Statement\":[]}";
```

There is no template-engine resolution, so an `AWS::IAM::Role` whose `AssumeRolePolicyDocument` uses `Ref`, `Fn::Sub`, `Fn::Join` or `Fn::If` is stored as literal template syntax. That is the common shape for scoping a trust policy to `${AWS::AccountId}`, to a `Ref`-d role or user, or to a `!Sub`-built service principal. `GetRole` then returns unresolved template text, and any consumer that parses the trust policy sees `{"Fn::Sub": "..."}` where a principal ARN should be.

## The class is already inconsistent with itself

This is the part that makes it clearly a bug rather than a missing feature: the **same class already uses the right helper** for inline policies at `IamRoleCfnProvisioner.java:169`. Only the trust-policy path was missed, so the two paths in one provisioner disagree about whether intrinsics get resolved.

## Fix

Same one-liner as the `AWS::EC2::VPCEndpoint` `PolicyDocument` fix in #2526: route it through `CloudFormationTemplateEngine.resolveJsonAttribute`, the helper added in #2317 for JSON-valued attributes. It resolves the tree and unwraps a textual result rather than re-serialising it — a naive `resolveNode(...).toString()` double-encodes, which is what #2317 was filed about.

```java
String assumeDoc = props == null ? null
        : ctx.engine().resolveJsonAttribute(props.path("AssumeRolePolicyDocument"));
if (assumeDoc == null) {
    assumeDoc = "{\"Version\":\"2012-10-17\",\"Statement\":[]}";
}
```

`resolveJsonAttribute` returns `null` for null and missing nodes, so the default still applies when the property is absent.

## Test

Mirror `Ec2VpcEndpointCfnProvisionerTest.policyDocumentIntrinsicsAreResolvedNotStoredAsTemplateSyntax`: a role template whose `AssumeRolePolicyDocument` principal is an `Fn::Sub`/`Ref`, asserting the document passed to `IamService.createRole` carries the resolved value and no `Fn::` or `"Ref"` keys.

AbarnaaSree:
## Summary

Closes #2538

Resolve `AssumeRolePolicyDocument` through `CloudFormationTemplateEngine.resolveJsonAttribute` before storing it in `AWS::IAM::Role`.

Previously, the raw template node was serialized directly, causing intrinsics such as `Fn::Sub` and `Ref` to be stored as unresolved CloudFormation syntax.

Added a regression test to verify that the resolved trust policy is passed to `IamService.createRole` without unresolved intrinsic keys.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The IAM role `AssumeRolePolicyDocument` was previously stored with unresolved CloudFormation intrinsic functions such as `Fn::Sub`.

The fix resolves JSON-valued attributes using the existing `resolveJsonAttribute` helper before passing the policy to the IAM service.

## Checklist

- [x] `mvn -Dtest=IamRoleCfnProvisionerTest test` passes locally

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/IamRoleCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/IamRoleCfnProvisioner.java
@@ -58,9 +58,14 @@ public class IamRoleCfnProvisioner implements CfnResourceProvisioner {
             throw new AwsException("ValidationError",
                     "Updating RoleName requires resource replacement, which is not supported.", 400);
         }
-        String assumeDoc = props != null && props.has("AssumeRolePolicyDocument")
-                ? props.get("AssumeRolePolicyDocument").toString()
-                : "{\"Version\":\"2012-10-17\",\"Statement\":[]}";
+        String assumeDoc = props == null
+                ? null
+                : ctx.engine().resolveJsonAttribute(props.path("AssumeRolePolicyDocument"));
+
+        if (assumeDoc == null) {
+            assumeDoc = "{\"Version\":\"2012-10-17\",\"Statement\":[]}";
+        }
+
         String path = ctx.resolveOptional(props, "Path");
         if (path == null) {
             path = "/";
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/provisioners/IamRoleCfnProvisionerTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/provisioners/IamRoleCfnProvisionerTest.java
@@ -15,6 +15,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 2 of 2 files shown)


hectorvent:
Thanks for this, and welcome. Routing AssumeRolePolicyDocument through resolveJsonAttribute is exactly the right fix, it is the same helper the SNS and Step Functions provisioners use for JSON-valued attributes, and your test pins that the stored doc carries no unresolved intrinsics. The missing-document fallback to the empty policy still works since the helper returns null for a missing node. Two small things, neither blocking:

**(follow-up, separate PR)** greptile's point about Fn::If is real but not yours to fix here: the engine collapses an object-valued Fn::If branch to an empty string, and it does that today for every caller of this helper. Worth an issue against the template engine itself.

**(follow-up)** The new test method sits at column one and the ctx() helper's braces drifted; worth a formatting pass if you push again.

No blockers from my side.

hectorvent (approved):

Issue #2481, opened 2026-08-22
[BUG] ElastiCache DescribeReplicationGroups omits Engine, MemberClusters and NodeGroups (PrimaryEndpoint) for cluster-mode-disabled groups


hisham-coder:
**What happens**
For a replication group created with `CreateReplicationGroup` (cluster mode off, 1 node) Floci returns only `ConfigurationEndpoint`, plus `ClusterEnabled=false`, `AutomaticFailover`, `MultiAZ`, encryption flags. Missing vs AWS:

- `Engine` (and `EngineVersion`, `CacheNodeType`) — not stored at create time
- `MemberClusters` — AWS lists `<group>-001`; Floci returns an empty list
- `NodeGroups[0]` with `PrimaryEndpoint` / `ReaderEndpoint` and `NodeGroupMembers` — absent
- On AWS a cluster-mode-disabled group has **no** `ConfigurationEndpoint`; clients connect to `NodeGroups[0].PrimaryEndpoint`

**Why it matters**
terraform-provider-aws `aws_elasticache_replication_group`:
- `primary_endpoint_address` is null (consumers have to fall back to `configuration_endpoint_address`)
- `engine` reads back null → ForceNew → the group is replaced on every apply unless `ignore_changes = [engine]`
- `num_cache_clusters = len(MemberClusters) = 0` → the provider calls `IncreaseReplicaCount`, which Floci rejects as unsupported → apply fails unless `ignore_changes = [num_cache_clusters]`

**Reproduce** (Floci 1.7.0)
```bash
aws --endpoint-url http://localhost:4566 elasticache create-replication-group --replication-group-id r --replication-group-description r \
  --engine redis --cache-node-type cache.t3.micro --num-cache-clusters 1
aws --endpoint-url http://localhost:4566 elasticache describe-replication-groups --replication-group-id r \
  --query 'ReplicationGroups[0].[Engine,MemberClusters,NodeGroups,ConfigurationEndpoint]'
# [ null, [], null, {"Address": "...", "Port": 6379} ]
```

**Expected**
[...]

PR #2633, opened 2026-08-27
feat(elasticache): support cluster-mode replication groups

kurtmc:
## Summary

Adds cluster-mode-enabled replication groups to the ElastiCache emulation. Today `CreateReplicationGroup` always starts a single cluster-mode-disabled container, ignores `NumNodeGroups`/`ReplicasPerNodeGroup`, and hardcodes `ClusterEnabled=false` in describe — so anything that needs a sharded Redis/Valkey topology (cluster-aware clients, terraform `aws_elasticache_replication_group` with `num_node_groups`) can't run against floci.

With this change, when a create requests cluster mode (`NumNodeGroups > 1`, a `*.cluster.on` parameter group, a custom parameter group with `cluster-enabled=yes`, or `ClusterMode=enabled`), floci provisions a real sharded Valkey cluster:

- **One `--cluster-enabled` container per node** — `NumNodeGroups × (1 + ReplicasPerNodeGroup)` in total — with cluster formation done the way `valkey-cli --cluster create` does it: distinct config epochs, `CLUSTER MEET` over the container network, `ADDSLOTSRANGE` on each shard primary, `CLUSTER REPLICATE` for replicas, then waiting for `cluster_state:ok` on every node.
- **One auth-proxy port per node** from the existing proxy port range. Each node announces its proxy endpoint to clients via `cluster-announce-client-ipv4`/`cluster-announce-port` (requires Valkey ≥ 8.1; the default `valkey/valkey:8` image qualifies), so `CLUSTER SLOTS`/`CLUSTER SHARDS` and `MOVED`/`ASK` redirects point at addresses clients can actually reach, while the cluster bus keeps using container-network IPs. Any cluster-aware client works through the reported `ConfigurationEndpoint`.
- **Honest describe output**: `ClusterEnabled`, `ClusterMode`, one `NodeGroup` per shard with `Slots` and `NodeGroupMembers`, `MemberClusters`, `Engine`, `ARN`, `AutomaticFailover`/`MultiAZ`. Each member cluster also answers `DescribeCacheClusters` (node type, engine version, parameter/subnet group, per-node port) — which is where terraform-provider-aws reads those values from.

This also covers most of the cluster-mode-**disabled** read-back gaps from #2481: `Engine`/`EngineVersion`/`CacheNodeType` are now stored at create time, `MemberClusters` lists the synthesised `<group>-001…` members (and those ids answer `DescribeCacheClusters`), and `NodeGroups[0]` carries `PrimaryEndpoint`/`ReaderEndpoint`/`NodeGroupMembers`. One deliberate deviation from that issue: `ConfigurationEndpoint` is still emitted for cluster-mode-disabled groups (AWS omits it there) because floci's existing integration/compat tests and docs read it — happy to drop it in a follow-up if you'd rather take the breaking step now.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against real consumers, not just handcrafted HTTP:

- **terraform-provider-aws v6.62.0** (Terraform 1.15.0): new bats compat test `compat-terraform/test/elasticache_cluster.bats` applies an `aws_elasticache_replication_group` with `engine = "valkey"`, `num_node_groups = 2`, `replicas_per_node_group = 1`, asserts `cluster_enabled` reads back `true`, and asserts a `terraform plan -detailed-exitcode` straight after apply is clean (the ForceNew/perpetual-diff class of bug from #2481).
- **AWS CLI 2.34.32**: the same bats test checks `describe-replication-groups` topology (`NodeGroups`, `Slots`, `MemberClusters`) and `describe-cache-clusters --show-cache-node-info` for a member id.
- **Data plane**: the bats test speaks raw RESP through the proxy to the configuration endpoint and asserts the backing Valkey genuinely runs in cluster mode — `INFO cluster` → `cluster_enabled:1`, `CLUSTER INFO` → `cluster_state:ok`, `cluster_size:2`, `cluster_known_nodes:4`.
- `ElastiCacheClusterIntegrationTest` (Docker-backed, 2 shards × 1 replica) additionally asserts real redirect semantics: `SET foo` against the first shard returns `MOVED 12182 127.0.0.1:<second-shard-proxy-port>` matching the port reported by `DescribeCacheClusters`, and the write succeeds on the redirect target.

Existing behavior is preserved: plain creates stay single-container cluster-mode-disabled, all 105 pre-existing + new ElastiCache tests pass, and the compat-java/awscli suites' assertions (`ConfigurationEndpoint`, parameter groups) are untouched.

## Checklist

- [ ] `./mvnw test` passes locally — full suite not run locally; the complete ElastiCache suite (105 tests incl. Docker-backed integration) and `core.common.docker` suite pass

### DIFF
--- a/compatibility-tests/compat-terraform/Dockerfile
+++ b/compatibility-tests/compat-terraform/Dockerfile
@@ -1,6 +1,6 @@
 FROM hashicorp/terraform:1.14.7 AS terraform
 
-FROM amazon/aws-cli:2.15.52
+FROM amazon/aws-cli:2.34.32
 
 # Copy terraform binary from official Terraform image
 COPY --from=terraform /bin/terraform /usr/local/bin/terraform
--- a/compatibility-tests/compat-terraform/test/elasticache-cluster-tf/main.tf
+++ b/compatibility-tests/compat-terraform/test/elasticache-cluster-tf/main.tf
@@ -1,0 +1,26 @@
+resource "aws_elasticache_replication_group" "cluster" {
+  replication_group_id       = "floci-tf-valkey-cluster"
+  description                = "floci terraform cluster-mode compatibility test"
+  engine                     = "valkey"
+  node_type                  = "cache.t4g.micro"
+  parameter_group_name       = "default.valkey8.cluster.on"
+  automatic_failover_enabled = true
+  num_node_groups            = 2
+  replicas_per_node_group    = 1
+}
+
+output "cluster_enabled" {
+  value = aws_elasticache_replication_group.cluster.cluster_enabled
+}
+
+output "configuration_endpoint_address" {
+  value = aws_elasticache_replication_group.cluster.configuration_endpoint_address
+}
+
+output "num_node_groups" {
+  value = aws_elasticache_replication_group.cluster.num_node_groups
+}
+
+output "member_clusters" {
+  value = aws_elasticache_replication_group.cluster.member_clusters
+}
--- a/compatibility-tests/compat-terraform/test/elasticache-cluster-tf/provider.tf
+++ b/compatibility-tests/compat-terraform/test/elasticache-cluster-tf/provider.tf
@@ -1,0 +1,28 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 6.0"
+    }
+  }
+}
+
+variable "endpoint" {
+  type    = string
+  default = "http://localhost:4566"
+}
+
+provider "aws" {
+  region     = "us-east-1"
+  access_key = "test"
+  secret_key = "test"
+
+  skip_credentials_validation = true
+  skip_metadata_api_check     = true
+  skip_requesting_account_id  = true
+  s3_use_path_style           = true
+
+  endpoints {
+    elasticache = var.endpoint
+  }
+}
--- a/compatibility-tests/compat-terraform/test/elasticache_cluster.bats
+++ b/compatibility-tests/compat-terraform/test/elasticache_cluster.bats
@@ -1,0 +1,131 @@
+#!/usr/bin/env bats
+# ElastiCache Cluster Mode Compatibility Test
+#
+# Applies an aws_elasticache_replication_group with num_node_groups /
+# replicas_per_node_group (Terraform's cluster-mode shape) against floci and
+# verifies both sides of the contract: the terraform read-back is honest
+# (cluster_enabled, member_clusters, a clean re-plan), and the data plane is a
+# genuine Valkey cluster (CLUSTER INFO over RESP through the auth proxy).
+
+setup_file() {
+    load 'test_helper/common-setup'
+
+    EC_TF_DIR="$(cd "$(dirname "$BATS_TEST_FILENAME")/elasticache-cluster-tf" && pwd)"
+    cd "$EC_TF_DIR"
+
+    echo "# === ElastiCache Cluster Mode Test ===" >&3
+    echo "# Endpoint: $FLOCI_ENDPOINT" >&3
+    echo "# Config: $EC_TF_DIR" >&3
+
+    rm -rf .terraform .terraform.lock.hcl terraform.tfstate* 2>/dev/null || true
+
+    echo "# --- terraform init ---" >&3
+    run terraform init -input=false -no-color
+    if [ "$status" -ne 0 ]; then
+        echo "# terraform init failed: $output" >&3
+        return 1
+    fi
+
+    echo "# --- terraform apply (2 shards, 1 replica each) ---" >&3
+    run terraform apply -var="endpoint=${FLOCI_ENDPOINT}" -input=false -auto-approve -no-color
+    if [ "$status" -ne 0 ]; then
+        echo "# terraform apply failed: $output" >&3
+        return 1
+    fi
+}
+
+teardown_file() {
+    load 'test_helper/common-setup'
+
+    EC_TF_DIR="$(cd "$(dirname "$BATS_TEST_FILENAME")/elasticache-cluster-tf" && pwd)"
+    cd "$EC_TF_DIR"
+
+    terraform destroy -var="endpoint=${FLOCI_ENDPOINT}" -input=false -auto-approve -no-color || true
+    rm -rf .terraform .terraform.lock.hcl terraform.tfstate* 2>/dev/null || true
+}
+
+setup() {
+    load 'test_helper/common-setup'
+    EC_TF_DIR="$(cd "$(dirname "$BATS_TEST_FILENAME")/elasticache-cluster-tf" && pwd)"
+}
+
+# Sends one RESP command over /dev/tcp and echoes whatever the server replies
+# within the read window. No redis-cli in this image, so raw RESP it is.
+valkey_command() {
+    local host="$1" port="$2"
+    shift 2
+    local payload="*$#\r\n"
+    local arg
+    for arg in "$@"; do
+        payload+="\$${#arg}\r\n${arg}\r\n"
+    done
+    exec 9<>"/dev/tcp/${host}/${port}" || return 1
+    printf '%b' "$payload" >&9
+    local reply="" chunk
+    while IFS= read -r -t 3 -u 9 chunk; do
+        reply+="${chunk}"$'\n'
+    done
+    exec 9<&- 9>&-
+    printf '%s' "$reply"
+}
+
+@test "ElastiCache cluster mode: terraform reads back cluster_enabled=true" {
+    run terraform -chdir="$EC_TF_DIR" output -raw cluster_enabled
+    assert_success
+    assert_output "true"
+
+    run terraform -chdir="$EC_TF_DIR" output -raw num_node_groups
+    assert_success
+    assert_output "2"
+}
+
+@test "ElastiCache cluster mode: describe reports the sharded topology" {
+    run aws_cmd elasticache describe-replication-groups \
+        --replication-group-id floci-tf-valkey-cluster
+    assert_success
+    [ "$(json_get "$output" '.ReplicationGroups[0].ClusterEnabled')" = "true" ]
+    [ "$(json_get "$output" '.ReplicationGroups[0].AutomaticFailover')" = "enabled" ]
+    [ "$(json_get "$output" '.ReplicationGroups[0].Engine')" = "valkey" ]
+    [ "$(json_get "$output" '.ReplicationGroups[0].NodeGroups | length')" = "2" ]
+    [ "$(json_get "$output" '.ReplicationGroups[0].NodeGroups[0].Slots')" = "0-8191" ]
+    [ "$(json_get "$output" '.ReplicationGroups[0].NodeGroups[1].Slots')" = "8192-16383" ]
+    [ "$(json_get "$output" '.ReplicationGroups[0].MemberClusters | length')" = "4" ]
+}
+
+@test "ElastiCache cluster mode: member cache clusters answer DescribeCacheClusters" {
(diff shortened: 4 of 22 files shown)


pgermosen (commented):
All three bot findings are genuinely closed, and the restore fix is the interesting one — it
went further than the finding asked for, in a way worth calling out. One gap in the new
failure path, one question about a validation, and two nits.

## What I verified against the reference

**`create-failed` is a real `ReplicationGroup.Status` value.** The reference lists exactly
`creating`, `available`, `modifying`, `deleting`, `create-failed`, `snapshotting`, so the new enum
constant is a status AWS actually reports rather than an invented one — and `wireName()` is
needed precisely because it's the first multi-word value in that enum, where `name().toLowerCase()`
would have emitted `create_failed`. I checked every emission site: `replicationGroupXml:744` is the
only one that renders this enum, and it uses `wireName()`. Nothing left on the old path.

**`ReplicasPerNodeGroup` 0–5 and `NumCacheClusters` 1–6 are exact.** *"Valid values are 0 to 5"* and
*"The maximum permitted value for `NumCacheClusters` is 6 (1 primary plus 5 replicas)"* — both
`validateRange` calls match the documented bounds character-for-character.

**The parameter-group signal is AWS's own.** `CreateReplicationGroup` says *"To create a Valkey or
Redis OSS (cluster mode enabled) replication group, use
`CacheParameterGroupName=default.redis3.2.cluster.on`"*, which is what the `.endsWith(".cluster.on")`
check keys on. Deriving cluster mode from the parameter group rather than only from
`NumNodeGroups` is the right primary signal, and the javadoc says so.

**`NodeGroupId` formatting matches.** The reference: *"A Valkey or Redis OSS (cluster mode enabled)
replication group contains 1 to 90 node groups numbered 0001 to 0090."* `String.format("%04d",
shard + 1)` produces exactly that, and the `Integer.parseInt(node.getNodeGroupId()) - 1` round-trip
back to a zero-based shard index is safe against the leading zeros.

**Slot division is gapless and complete.** `slotRange` gives 2 shards → `0-8191` / `8192-16383`, 3
shards → `0-5460` / `5461-10922` / `10923-16383`. No gaps, no overlap, all 16384 covered at any
shard count, and the integer arithmetic can't overflow at the validated ceiling.

**`Slots` is rendered per shard, not per group.** `appendClusterModeNodeGroups` groups by
`nodeGroupId` first, so `nodes.getFirst().getSlots()` is that shard's range. Easy thing to get
wrong when every node of a shard carries the same string.

## The restore fix solved more than was reported

The bot's P1 was that restored cluster-mode groups keep their persisted topology but lose their
process-local runtime, so they read as `available` while refusing connections. The fix restores
containers, proxies and port reservations, and marks the group `create-failed` if it can't. That's
the reported problem.

The part that wasn't reported is the port-reservation half, and it's handled with more care than
the finding implies. `reserveOrAllocateProxyPort` only keeps the persisted port if `usedPorts.add`
actually wins:

```java
if (persistedPort > 0 && usedPorts.add(persistedPort)) {
    return persistedPort;
}
return allocateProxyPort();
```

so a group whose advertised port was taken during startup slides to a fresh one instead of
double-booking. And `rollbackClusterModeGroup`'s signature changed from `List<ClusterNode>` to
`Collection<Integer>` specifically so restore can pass `reservedPorts` — the ports *this* restore
actually reserved — rather than every node's persisted port. Passing the nodes would have released
a port that a different group legitimately holds, on the failure path, which is the kind of bug
that only shows up as a mystery collision much later. Marking the container coordinates `transient`
on `ClusterNode` while leaving `proxyPort` persistent is the same distinction expressed in the
model.

Reserving all ports in a first pass before starting any container is also the right order, for the
same reason the ECS work landed on: it puts the failure ahead of the first side effect.

## (should fix) Member clusters still report `available` after a failed restore

`memberCacheClusterXml` hardcodes the status:

```java
.elem("CacheClusterStatus", "available")
```

That was defensible before this PR, when a group was only ever `creating`/`available`/`deleting`.
It isn't now, because the restore fix introduced a state where the group is `create-failed` and its
members are not available:

```
# floci restarted with a persisted cluster-mode group; docker unavailable or formation fails
aws elasticache describe-replication-groups --replication-group-id g
  → Status: create-failed          ✅ the new behaviour, correct

aws elasticache describe-cache-clusters --cache-cluster-id g-0001-001
  → CacheClusterStatus: available  ❌ nothing is running
```

The two answers contradict each other about the same resource, and it's the members that lie. This
matters more than a normal consistency nit because of the reason the member synthesis exists in the
first place — your own javadoc: *"The terraform provider follows `MemberClusters` into
DescribeCacheClusters, so every member name a describe hands out must answer there."* The consumer
you built this for is exactly the one that will read the wrong value.

One wrinkle that stops this being a one-line mirror of the group status: **AWS's
`CacheClusterStatus` has no `create-failed`.** The documented set is `available`, `creating`,
`deleted`, `deleting`, `incompatible-network`, `modifying`, `rebooting cluster nodes`,
`restore-failed`, `snapshotting` — so `create-failed` is legal on `ReplicationGroup.Status` and not
on the cluster. `restore-failed` is the one documented failure value, and it happens to describe
what literally occurred: a restore failed. That's what I'd emit, but I'd rather you pick, since
either that or `incompatible-network` is an approximation and you've been probing live behaviour
elsewhere in this PR.

Credit where it's due: the `showNodeInfo` block is already guarded by
`g.getConfigurationEndpoint() != null`, and restore-failure nulls the endpoint, so a failed group
doesn't hand out node endpoints that refuse connections. The status field is the only thing left
claiming otherwise.

## (question) Should a cluster-mode group without `AutomaticFailoverEnabled` be rejected?

`CreateReplicationGroup` states:

> `AutomaticFailoverEnabled` must be enabled for Valkey or Redis OSS (cluster mode enabled)
> replication groups.

`applyCommonAttributes` takes the request at face value —
`Boolean.TRUE.equals(request.automaticFailoverEnabled())` — so a cluster-mode create that omits it
is accepted and reads back `AutomaticFailover: disabled`. On AWS that combination shouldn't exist,
so a client asserting on it gets an answer no real account would give.

I'm raising this as a question rather than a finding because AWS's own docs are in tension here.
The same page allows `ReplicasPerNodeGroup=0` for a cluster-mode group, and automatic failover with
zero replicas has nothing to fail over to — so the "must" can't hold universally, and I can't tell
from the reference whether AWS rejects the request, silently forces the flag true, or permits it
when there are no replicas. It's the doc-string tier of evidence rather than a shape constraint,
which is where I'd want a live probe before prescribing anything. Your terraform test sets the flag,
so it wouldn't have surfaced there either way.

## Nits

- `validateRange("NumNodeGroups", …, 1, 500)` uses the raised ceiling. The reference's default is
  lower — *"comprised of from 1 to 90 shards"*, and the dedicated error
  `NodeGroupsPerReplicationGroupQuotaExceeded` says *"The default maximum is 90"* — with 500 being
  what a limit increase buys *"if the engine version is 5.0.6 or higher"*. Taking the raised ceiling
  is the friendlier choice for an emulator and I wouldn't change it; the part that's off is that
  exceeding it reports a generic parameter error rather than
  `NodeGroupsPerReplicationGroupQuotaExceeded`. Firmly theoretical — nobody is starting 501 Valkey
  containers locally.

- `ClusterMode` has three valid values, `enabled | disabled | compatible`. `resolveClusterEnabled`
  only tests `"enabled"`, so `compatible` falls through to the `NumNodeGroups > 1` check and a
  single-shard `compatible` group comes back `disabled`. Compatible is the mid-migration state where
  both client types connect, so mapping it to cluster-enabled is arguably the truer answer — but
  it's a migration affordance floci doesn't otherwise model, and I'd be fine leaving it as long as
  it's deliberate.

## Notes

- **On the `ConfigurationEndpoint` deviation you flagged:** keep it, follow up separately. You're
  right that AWS omits it for cluster-mode-disabled groups, but I'd add that the reference doesn't
  actually document the conditionality — `ConfigurationEndpoint` carries no statement about when
  it's present, so that rule is real-behaviour-derived rather than something a reader can check
  against the shape. Combined with it being a breaking change to your own compat tests and docs, and
  entirely separate from cluster mode, bundling it here would mean a feature PR also silently
  removing a field existing consumers read. The follow-up is the right shape.

- **The restore covers cluster-mode groups only** — `restorePersistedRuntime` skips anything where
  `!group.isClusterEnabled()`, so a cluster-mode-*disabled* group still comes back from a restart
  reported `available` with a dead endpoint. That's pre-existing rather than introduced, and I'd
  keep it out of this PR. Worth saying only because the asymmetry is now a visible property of the
  code rather than a missing feature: there's a restore path, and it deliberately handles half the
  groups. A sentence in the javadoc naming the disabled case as out of scope would stop the next
  reader assuming it was an oversight.

- Deriving `Slots` from the same `slotRange` helper the formation code uses to issue
  `ADDSLOTSRANGE` means the reported topology can't drift from the topology actually configured on
  the nodes. That's the version of this that stays correct — the alternative, computing the describe
  output separately, is how an emulator ends up reporting a layout its own data plane doesn't have.

- Announcing the configured hostname via `--cluster-announce-hostname` plus
  `--cluster-preferred-endpoint-type hostname`, rather than resolving that name to an IP inside
  floci's own container, is a better fix than the finding asked for. The bot described the symptom
  (an internal or unresolvable address gets announced); handing Valkey the name itself removes the
  resolution step entirely, so how the name resolves inside the container stops mattering. Keeping
  the best-effort IPv4 alongside it for consumers that read the address fields is the right hedge,
  and the rewritten javadoc explains the split.

pgermosen (approved):
Everything from the last round is closed, and I re-verified each fix directly against the current code rather than trusting the commit messages. The member cluster status fix is exactly right: it now reports the documented `restore-failed` cluster status after a failed restore instead of mirroring the group's `create-failed`, which isn't a legal value on a cluster in the first place. The quota error on an over-ceiling `NumNodeGroups` now uses the documented `NodeGroupsPerReplicationGroupQuotaExceeded` code and wording. And the close-failure logging is in place.

On the newest findings: the Compose-hostname fix is good, and I checked it isn't a new or speculative domain — the dedicated cluster-announce-hostname override lets Compose-internal clients keep resolving the service name they always have while external clients get a name that resolves everywhere, using `localhost.floci.io`, which is the project's own already-established public DNS entry used the same way elsewhere in this codebase.

On "cluster restoration blocks startup": I traced through it and the mechanics are accurate — serial container starts plus a cross-node formation wait, per persisted group. But I checked whether this is a new category of risk, and it isn't: RDS's and ELB's restore functions run from the exact same startup hook, immediately before this one, doing the same synchronous, serial container restoration with no async path. What's genuinely new here is the per-group multiplier from node count and cluster formation, which is a real scaling consideration worth tracking for the restore-on-startup approach as a whole, but not a defect specific to this PR or something to hold to a different bar than the RDS/ELB restoration that already works this way.

pgermosen:
@kurtmc one more thing worth a follow-up now that restoration moved to the background: `restorePersistedRuntime()` returns a `CompletableFuture` that runs the container/formation/proxy work asynchronously, but the only thing attached to it in `EmulatorLifecycle.onStart` is `.exceptionally(...)` for logging — nothing holds a reference to it anywhere `onStop()` could reach. `onStop()` calls `elastiCacheContainerManager.stopAll()`/`elastiCacheProxyManager.stopAll()` synchronously, which only stops whatever's tracked at that moment. If Floci shuts down while a background restore is still running, a container or proxy the restore starts after `stopAll()` has already run won't be stopped and can leak on the host, and a status write the restore makes after the final storage flush can land after (or be lost to) that flush.

Given the container-network Docker socket this all runs against, tracking the in-flight future somewhere `onStop()` can await (with a bounded timeout) or cancel before running the stop sequence would close this. Not asking for it in this PR — happy to see it as a follow-up given the base feature is already approved.

PR #2634, opened 2026-08-27
ci: cache compat test images on main instead of per-PR

hectorvent:
## Summary

Compat test-image builds miss their layer cache ~83% of the time (2 hits in 12 sampled runs), costing up to 211s on the `sdk-test-java` leg — which sits on the compat critical path.

Root cause: Actions caches are **ref-scoped**. A `pull_request` run may *read* the base branch's caches, but its *writes* go to `refs/pull/N/merge`, private to that PR. `compatibility.yml` is `pull_request`-only and nothing on `main` wrote buildx caches, so `main` held 3 cache entries while four PR refs held ~3.07 GB each — **13.24 GB against GitHub's 10 GB repo limit**. Every PR built all eight images cold, wrote a copy no other PR could read, and LRU eviction then removed entries even between a single PR's own runs.

Measured on `sdk-test-java`'s "Build test image" step across 12 runs:

```
29  30  113  120  126  142  144  163  166  184  209  211   (seconds)
median 144s · only 2 of 12 were cache hits
```

This adds `warm-compat-caches.yml` to populate the scopes on `main`, and drops `cache-to` from the legs so they read that one shared copy. PRs now get a warm read on their **first** run, which was previously impossible.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)

## AWS Compatibility

## Checklist

- [ ] New or updated integration test added — **not applicable, see below**

**On tests:** workflow-only change with no automated coverage available. The mechanism cannot be exercised before merge — the producer runs on `push: main`, so the first post-merge PR compat run is the verification. Expected afterwards:

```bash
gh api repos/floci-io/floci/actions/cache/usage    # expect ~3 GB, well under 10 GB
# and sdk-test-java "Build test image" landing consistently at ~30s
```

## Follow-up required after merge

This does **not** clean the existing 13 GB. The PR-scoped entries need a one-off purge, or they will keep evicting the cache this adds:

```bash
gh api 'repos/floci-io/floci/actions/caches?per_page=100' --paginate \
  | grep -o 'refs/pull/[0-9]*/merge' | sort -u          # re-derive the live list
gh cache delete --all --ref "refs/pull/<N>/merge" --succeed-on-no-caches
```

Then `workflow_dispatch` **Warm Compat Caches** so the main cache is built into a clean budget rather than raced against eviction.

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -192,8 +192,14 @@ jobs:
           context: compatibility-tests/${{ matrix.test }}
           load: true
           tags: compat-${{ matrix.test }}
+          # Read-only on purpose. Actions caches are ref-scoped: a pull_request
+          # run's writes land in refs/pull/N/merge, private to that PR and
+          # unreadable by any other. Eight images x every open PR overran the
+          # repo's 10 GB cache budget (13.2 GB measured, 95% of it in four PR
+          # refs), so entries were evicted between a PR's own runs and this
+          # build swung 29s-211s at random. `warm-compat-caches.yml` populates
+          # these scopes on main instead; PRs read that single shared copy.
           cache-from: type=gha,scope=${{ matrix.test }}
-          cache-to: type=gha,scope=${{ matrix.test }},mode=max
 
       - name: Run tests
         id: tests
--- a/.github/workflows/warm-compat-caches.yml
+++ b/.github/workflows/warm-compat-caches.yml
@@ -1,0 +1,95 @@
+# Populates the buildx layer caches that `compatibility.yml`'s test legs read.
+#
+# Why this exists: GitHub Actions caches are ref-scoped. A pull_request run can
+# READ the base branch's caches but its WRITES land in refs/pull/N/merge, private
+# to that PR. `compatibility.yml` is pull_request-only, so before this workflow
+# `main` produced no buildx cache at all: every PR built all eight test images
+# cold and wrote its own ~3 GB copy that no other PR could ever read. Four
+# concurrent PRs pushed the repo to 13.2 GB against GitHub's 10 GB limit, so
+# entries were evicted between a PR's own runs and the sdk-test-java image build
+# swung between 29s (warm) and 211s (cold) at random, on the compat critical path.
+#
+# Producing the caches here, on main, means every PR gets a warm read on its
+# FIRST run, and the legs no longer write (see compatibility.yml). Keep this
+# workflow's scope names identical to the ones the legs read from.
+name: Warm Compat Caches
+
+on:
+  push:
+    branches: [main]
+    # The dependency-bearing inputs only — the files an image's *expensive* layer
+    # is keyed on. Test sources deliberately do not trigger a re-warm where they
+    # sit in a later layer than the dependency install, so a PR still gets the
+    # costly layers from a source-stale cache. Narrow on purpose: 182 commits
+    # touched compatibility-tests/ in the last 90 days but only 31 touched a
+    # dependency manifest, and each warm costs eight arm jobs.
+    paths:
+      - 'compatibility-tests/*/Dockerfile'
+      - 'compatibility-tests/*/pom.xml'            # sdk-test-java
+      - 'compatibility-tests/*/requirements.txt'   # sdk-test-python
+      - 'compatibility-tests/*/package*.json'      # sdk-test-node, compat-cdk
+      # sdk-test-go is the exception: its Dockerfile does `COPY . .` *before*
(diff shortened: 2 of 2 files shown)


pgermosen (approved):
The diagnosis is the valuable part here, and it's right. Ref-scoped caching is genuinely
counter-intuitive — a `pull_request` run *reads* the base branch's caches, so `cache-to` looks like
it's working right up until you notice no other PR can ever read what it wrote. Pairing that with
the measured spread (29s–211s, 2 hits in 12) and the 13.2 GB-against-10 GB figure turns a plausible
theory into a demonstrated one.

Both of Greptile's findings are addressed in `3701d6a9`, and one of them well beyond what was asked.

## What I verified

**The `sdk-test-go` exception is real and the comment explains it correctly.** Its Dockerfile does
`COPY . .` before `go mod init && go mod tidy && go mod download`, so the dependency layer is
invalidated by any source change and no manifest filter can track it — and `go.sum` is gitignored,
so Greptile's original point (the filter matches nothing) was right. Watching the whole directory is
the correct answer rather than inventing a filter that would silently match nothing.

**The terraform/opentofu omission is justified, not an oversight.** `COPY . .` is their final layer,
after the yum install and bats clones, so a `.tf` change invalidates no cached work. Saying that in
the file is what stops someone "fixing" the gap by adding a `*.tf` filter that would trigger eight
arm jobs for nothing.

**`compatibility.yml:29` really is SHA-pinned**, so the cross-reference in the new file's comment
holds up — I checked rather than take it on trust, since a comment that cites a line is only useful
while the line says what it claims.

**`lib/` is not a cache input.** There's a `compatibility-tests/lib/run-bats-with-junit.sh` and a
root `.dockerignore` sitting outside every per-test directory, which looked like a gap in the paths
filter. It isn't: each build's context is `compatibility-tests/<test>/`, so neither is visible to
any image build — `lib/` is used at run time. Nothing missing there.

**The hardening around the new workflow is right for a main-branch producer**: `permissions:
contents: read`, `sparse-checkout` limited to `compatibility-tests`, all three actions SHA-pinned,
and the runner-architecture comment explaining that an x64 build would write useless layers into an
arm64 scope. That last one is the sort of thing that would produce a silently useless cache rather
than a failure.

## (minor) The safety net's period equals the window it guards against

```yaml
schedule:
  # Actions evicts caches at 7 days idle …
  - cron: '0 6 * * 1'
```

A weekly cron against a 7-day eviction window leaves no margin. Reads reset the idle timer, so in
normal traffic this never fires in anger — but the quiet week is exactly the case the net exists
for, and in that case the cache can evict shortly before the cron would have refreshed it, leaving
every PR cold for most of a week.

`0 6 * * 1,4` halves the worst-case exposure for one more warm a week. Cheap insurance for something
whose whole purpose is covering the case where nothing else is running.

## (minor) The file you're editing has five unpinned actions, one of them in this job

The new workflow is 3-for-3 SHA-pinned, and its comment cites `compatibility.yml:29` as the
precedent. That line is pinned — but it's the supporting example, not the rule. `compatibility.yml`
has fifteen `uses:` and five are unpinned:

| line | action |
|---|---|
| `:89` | `actions/checkout@v7.0.1` |
| `:124` | `actions/upload-artifact@v7` |
| `:150` | `actions/download-artifact@v8` |
| `:181` | `actions/checkout@v7.0.1` |
| `:255` | `actions/upload-artifact@v7` |

`:181` is the checkout in the test leg — the same job whose `cache-to` line this PR removes, nine
lines below it. And the three artifact ones are pinned to a bare major (`@v7`, `@v8`), which moves
on every minor and patch release rather than only on a retag.

Not introduced here, and a `pull_request` workflow is a weaker target than a main-branch producer
that writes shared caches — so I'm not asking for it. But the PR is already inside that job making a
supply-chain argument in a comment, and pinning five lines is smaller than the comment explaining
why the sixth is pinned. Either fold them in or leave the new file's comment pointing at `:29`
knowing it's the exception rather than the convention.

## Notes

- Dropping `cache-to` while keeping `cache-from` on the legs is the right asymmetry, and the inline
  comment at the deletion site is what stops someone restoring it as an obvious omission — that
  deleted line is going to look like a mistake to the next reader, and now it doesn't.
- The warm job keeps its own `cache-from`, so a re-warm is incremental rather than always cold.
  Easy to drop when the only goal is writing.
- `fail-fast: false` means one broken image doesn't leave the other seven scopes unwarmed, which
  matters more here than in a test matrix.
- The paths filter's rationale — 182 commits touched `compatibility-tests/` in 90 days but only 31
  touched a dependency manifest — is the number that justifies narrowing, and it's in the file
  rather than only in the PR description where it would be lost after merge.

PR #2645, opened 2026-08-28
fix(appsync): use ApiKey.id as the API key value

TAKEDA-Takashi:
## Summary

`CreateApiKey` / `ListApiKeys` return a 7-character random string as `ApiKey.id` and keep the value the GraphQL execute endpoint actually validates in a separate `apiKey` field. That field is not part of the AWS AppSync API model ([ApiKey](https://docs.aws.amazon.com/appsync/latest/APIReference/API_ApiKey.html) has `id`, `description`, `expires`, `deletes`), so the AWS CLI and the SDKs drop it. On AWS, `id` *is* the `da2-…` value a client sends as `x-api-key`.

Until 1.7.0 this was invisible because the execute endpoint did not authenticate. Since Phase 7 (#2380, currently on `main` / nightly) validates `x-api-key` against the hidden `apiKey` value, a client that follows the AWS contract (`x-api-key: <ApiKey.id>`) always gets 401:

```bash
KEY_ID=$(aws appsync create-api-key --api-id $API_ID --query apiKey.id --output text)   # e.g. e2f514f
curl -s -i http://localhost:4566/v1/apis/$API_ID/graphql \
  -H "Content-Type: application/json" -H "x-api-key: $KEY_ID" --data '{"query":"{ __typename }"}'
# HTTP/1.1 401 Unauthorized
# {"errors":[{"errorType":"UnauthorizedException","message":"You are not authorized to make this call."}]}
```

Raw `ListApiKeys` body on `nightly-08272026` (visible with `aws --debug`; the CLI strips `apiKey` and `apiId`):

```json
{"apiKeys":[{"id":"e2f514f","apiId":"...","apiKey":"da2-cc88835"}],"nextToken":null}
```

Sending the raw `da2-cc88835` returns 200, so the validator is fine — the key is just unreachable through the API model.

## What changed

- `ApiKey.id` is now generated as `da2-` + 26 lowercase alphanumerics (`SecureRandom`), matching the shape AWS issues, and `validateApiKey` compares `x-api-key` against `id`.
- The non-model `apiKey` field is removed from `ApiKey`, so `CreateApiKey` / `UpdateApiKey` / `ListApiKeys` responses no longer carry it. (`apiId` is also outside the AWS model but is left alone here — this PR only fixes the split between the returned id and the validated value.)
- Keys persisted by earlier builds still deserialize (`@JsonIgnoreProperties(ignoreUnknown = true)`) and keep their short `id`. `validateApiKey` only accepts ids with the `da2-` prefix, so those legacy keys remain listable and deletable but never authenticate — the same as before this change, where the short `id` was never a valid `x-api-key` value either. Anyone who wants a working key on an existing persistent store deletes the old one and creates a new one.
- `docs/services/appsync.md`: note that `ApiKey.id` is the key value.
- `compatibility-tests/sdk-test-java` `AppSyncTest`: the test used a raw HTTP call to read the non-model `apiKey` field because the SDK could not see it; it now uses `ApiKey.id` from the SDK response, which is what a real client does.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against AWS CLI 2.x: `aws appsync create-api-key` on AWS returns `{"apiKey": {"id": "da2-xxxxxxxxxxxxxxxxxxxxxxxxxx", ...}}` and that `id` is the `x-api-key` value ([API reference](https://docs.aws.amazon.com/appsync/latest/APIReference/API_ApiKey.html), [Using API keys](https://docs.aws.amazon.com/appsync/latest/devguide/security-authz.html#api-key-authorization)). Incorrect behaviour before this PR: `id` was a 7-character value that the execute endpoint rejects, and the accepted value was only present in a field outside the API model.

## Tests

- `AppSyncServiceTest.createApiKeyIdIsTheUsableKeyValue` (new): `id` matches `da2-[a-z0-9]{26}`, validates, and an unknown `da2-…` value does not. Fails on `main` (7-character id).
- `AppSyncServiceTest.legacyShortApiKeyIdIsListedButNeverAuthenticates` (new): a key with a pre-change short id loaded from the store is listed and deletable but does not authenticate.
- `AppSyncAuthIntegrationTest` / `AppSyncExecutionIntegrationTest`: the `createApiKey` helpers now extract `apiKey.id` (the only field an SDK can read), so every test that executes a query with an API key becomes a regression guard.

Running the updated test classes against `main` (this PR's `src/test` on top of `main`'s `src/main`) fails 17 of 209: `unsignedApiKeyPostReturnsAppSyncJsonNotS3Xml`, 10 execution tests that authenticate with `apiKey.id`, the three `AppSyncServiceTest` lookups, and the three `AppSyncIntegrationTest` assertions on the id format. All 209 pass with this change.
- `AppSyncIntegrationTest.createApiKey`: asserts the `id` format and that the response has no `apiKey` field.
- `ApiKeyAuthValidatorTest`: updated to the `id`-based lookup.

## Checklist

- [x] `./mvnw test -Dtest=NativeImageRuntimeInitializationTest,AppSyncServiceTest,ApiKeyAuthValidatorTest,AppSyncAuthIntegrationTest,AppSyncIntegrationTest,AppSyncExecutionIntegrationTest` passes locally: 210 tests, 0 failures (Java 25 via `maven:3.9-eclipse-temurin-25`)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppSyncTest.java
@@ -133,9 +133,9 @@ class AppSyncTest {
         keyId = resp.apiKey().id();
         assertThat(keyId).isNotBlank();
         assertThat(resp.apiKey().description()).isEqualTo("sdk-test-key");
-        // AWS SDK v2 ApiKey has no apiKey() accessor; Floci returns the da2- secret on the wire.
-        apiKeyValue = fetchApiKeyValue(apiId, keyId);
-        assertThat(apiKeyValue).startsWith("da2-");
+        // As on AWS, the id is the key value sent as x-api-key.
+        assertThat(keyId).matches("da2-[a-z0-9]{26}");
+        apiKeyValue = keyId;
     }
 
     @Test
@@ -1314,25 +1314,5 @@ class AppSyncTest {
         assertThat(location.get("line")).isNotNull();
         assertThat(location.get("column")).isNotNull();
         assertThat(location.get("span")).isEqualTo(-1);
-    }
-
-    @SuppressWarnings("unchecked")
-    private static String fetchApiKeyValue(String apiId, String keyId) throws Exception {
-        String url = TestFixtures.endpoint() + "/v1/apis/" + apiId + "/apikeys";
-        HttpRequest req = HttpRequest.newBuilder()
-                .uri(URI.create(url))
-                .header("Authorization", "AWS4-HMAC-SHA256 Credential=test/20260205/us-east-1/appsync/aws4_request")
-                .GET()
-                .build();
-        HttpResponse<String> resp = HttpClient.newHttpClient().send(req, HttpResponse.BodyHandlers.ofString());
-        Map<String, Object> body = mapper.readValue(resp.body(), Map.class);
-        List<Map<String, Object>> keys = (List<Map<String, Object>>) body.get("apiKeys");
-        assertThat(keys).isNotEmpty();
-        for (Map<String, Object> key : keys) {
-            if (keyId.equals(key.get("id"))) {
-                return String.valueOf(key.get("apiKey"));
-            }
-        }
-        throw new IllegalStateException("API key not listed for id " + keyId);
     }
 }
--- a/docs/services/appsync.md
+++ b/docs/services/appsync.md
@@ -77,6 +77,8 @@ Floci implements the AWS AppSync Management API, providing local emulation of Gr
 | `DeleteApiKey` | Delete an API key |
 | `ListApiKeys` | List all API keys for an API |
 
+As on AWS, `ApiKey.id` is the key value itself (`da2-` followed by 26 lowercase alphanumerics) and is what clients send in the `x-api-key` header. There is no separate secret field.
+
 ### Tags
 
 | Operation | Description |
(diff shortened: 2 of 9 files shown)


pgermosen (approved):
Clean, well-scoped fix for a real regression. The premise checks out against the reference — `ApiKey`'s modeled shape is exactly `id`, `description`, `expires`, `deletes`, no separate secret field — so AWS's `id` genuinely is the value clients send as `x-api-key`, and the previous split (a cosmetic `id` for display, a hidden `apiKey` for the real secret) put the real value somewhere no SDK user could reach. That this stayed invisible until `x-api-key` validation was added, and became a hard break the moment it landed, is a believable failure mode, and the repro in the description (create-api-key, then 401 on the returned id but 200 on the raw secret) demonstrates it directly rather than just asserting it.

The generated id format (`da2-` + 26 lowercase alphanumerics via `SecureRandom`) matches the real shape, and the added prefix guard in `validateApiKey` is the right way to handle backward compatibility: a legacy 7-character id was never a valid `x-api-key` value before this change either, so refusing to authenticate it isn't a new restriction — and there's a dedicated test confirming such a key stays listable and deletable while never authenticating.

Test coverage pins the regression from multiple angles: a unit test on the id format and that a guessed key doesn't validate, the legacy-key backward-compatibility test, and the SDK compatibility test now reading `apiKey.id` the way a real client would instead of reaching for the field through a raw HTTP call — which also means that test now exercises the actual client-visible contract rather than an internal implementation detail. The claim that 17 of 209 tests fail without this change and all 209 pass with it is concrete, checkable evidence.

Approving.

Issue #2647, opened 2026-08-28
[BUG] lambda: deployment package extraction overlays the previous package instead of replacing it


avison9:
## Service

Lambda

## AWS API Action

UpdateFunctionCode (also CreateFunction on redeploy of an existing function name)

## Expected behavior

A deployment package **replaces** the function's code. Real AWS treats the package as the complete,
immutable contents of the function: a file present in v1 and absent from v2 is gone after the
update, and invoking a handler that pointed at it fails with the runtime's "module not found".

## Actual behavior

`ZipExtractor.extractTo` writes entries into the function's code directory but never clears it
first, so each deployment is **overlaid** on top of whatever the previous one left behind. Files
removed from the package survive indefinitely and stay loadable — a handler pointing at a deleted
module keeps resolving and executing stale code that is no longer in the deployment package.

The same missing reset means a **failed** extraction leaves a partially-populated code directory in
place rather than rolling back, so a function can be left with a mix of two packages after an error.

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1
E="--endpoint-url http://localhost:4566"

python3 - <<'PY'
import zipfile
with zipfile.ZipFile('v1.zip','w',zipfile.ZIP_DEFLATED) as z:
    z.writestr('index.js','exports.handler = async () => ({body: "v1"});')
    z.writestr('legacy/old.js','exports.handler = async () => ({body: "STALE"});')
with zipfile.ZipFile('v2.zip','w',zipfile.ZIP_DEFLATED) as z:      # legacy/old.js deleted
    z.writestr('index.js','exports.handler = async () => ({body: "v2"});')
PY

aws $E lambda create-function --function-name overlay-repro --runtime nodejs20.x \
  --role arn:aws:iam::000000000000:role/r --handler legacy/old.handler \
  --zip-file fileb://v1.zip

aws $E lambda update-function-code --function-name overlay-repro --zip-file fileb://v2.zip

aws $E lambda invoke --function-name overlay-repro out.json && cat out.json
```

Observed:

```
create-function        "CodeSize": 313
update-function-code   "CodeSize": 161      <- Floci records that the package shrank

invoke                 "StatusCode": 200
out.json               {"body":"STALE"}     <- v1 code, deleted from the v2 package
```

Expected (real AWS): the update replaces the package, `legacy/old.js` no longer exists, and the
[...]

Issue #2648, opened 2026-08-28
[BUG] cloudformation: unreadable Lambda S3 code silently deploys a stub handler instead of failing the stack


avison9:
## Service

CloudFormation (provisioning `AWS::Lambda::Function`)

## AWS API Action

CreateStack / UpdateStack

## Expected behavior

If a stack declares `Code: { S3Bucket, S3Key }` and Floci cannot read that object, the resource
should **fail** — real CloudFormation reports the resource `CREATE_FAILED` with an error naming the
inaccessible code (`Error occurred while GetObject. S3 Error Code: NoSuchKey`), and the stack rolls
back. A stack that reaches `CREATE_COMPLETE` must have deployed the code the template pointed at.

## Actual behavior

`CloudFormationResourceProvisioner.resolveLambdaCode` probes the object with
`s3Service.getObject(s3Bucket, s3Key)` and, on **any** exception, logs a warning and silently falls
through to a built-in stub package:

```java
if (s3Bucket != null && s3Key != null) {
    try {
        s3Service.getObject(s3Bucket, s3Key);
        return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key), ...);
    } catch (Exception e) {
        LOG.warnv("S3 code not found for Lambda ({0}/{1}), using default handler: {2}", ...);
    }
}
...
return new LambdaCodeSpec(Map.of("ZipFile", defaultHandlerZipBase64()), "default-handler");
```

`defaultHandlerZipBase64()` is a one-entry archive containing
`index.js` = `exports.handler=async(e)=>({statusCode:200})`. The consequence depends on the
template's `Handler`, and both outcomes are bad:

1. **Handler is `index.handler`** — the stack reaches `CREATE_COMPLETE`, the function is
   invocable, and it silently returns `{"statusCode":200}` for every request. The user's code was
   never deployed and **nothing in the API surface says so** — only a container log line. This is
   the more dangerous case: tests written against the stack can pass against a stub.
2. **Handler is anything else** (`src/api.handler`, `dist/index.handler`, …) — the stub does not
   contain that module, so `CreateFunction` fails the handler check and the resource fails with
   `Handler file 'src/api' not found in deployment package`. The message points at the handler,
   which is correct in the template and present in the real package, so it sends you looking in
   entirely the wrong place.

Symptom (2) is the misleading error reported in the second half of #2593.

The `catch (Exception e)` is also unconditional: a transient/internal S3 error, a permissions
[...]

PR #2649, opened 2026-08-28
fix(lambda): read deployment packages via the zip central directory

### LINKED ISSUE #2593: [BUG] Lambda: CreateFunction fails to extract some zips ("only DEFLATED entries can have EXT descriptor")
## Service

Lambda

CreateFunction

## Expected behavior

The function is created and the package is extracted. Real AWS accepts the same zip.

## Actual behavior

```
Failed to extract deployment package: only DEFLATED entries can have EXT descriptor
```

The message comes from `java.util.zip.ZipInputStream`, which rejects an entry that is STORED but carries a data descriptor.

When the package is referenced from CloudFormation rather than uploaded directly, the failure is reported as:

```
Handler file '<handler>' not found in deployment package
```

which is misleading, since the handler is present and the archive simply never extracted.

## Reproduction

Any Serverless Framework service whose package includes `node_modules`. Mine was 59 MB.

```bash
npx serverless package --stage dev     # writes .serverless/<service>.zip

aws --endpoint-url http://localhost:4566 s3 mb s3://pkg
aws --endpoint-url http://localhost:4566 s3 cp .serverless/<service>.zip s3://pkg/fn.zip

  --function-name repro --runtime nodejs20.x \
  --role arn:aws:iam::000000000000:role/r \
  --handler index.handler --code S3Bucket=pkg,S3Key=fn.zip
```

Excluding `node_modules` via `package.patterns` drops the same project to 108 KB and it deploys fine, so the I guess the trigger is something about the larger archive's contents.

I could not reduce this to a smaller case. These all worked, so they are not the trigger on their own:

- a 60 MB archive from the `zip` CLI
- an `archiver` zip containing a STORED entry alongside a DEFLATED one
- an `archiver` zip with 70,001 entries (Zip64 by entry count)
- the same project's 108 KB package, same handler, same bucket

Reading the archive with `ZipFile` (central directory) instead of `ZipInputStream` (sequential) would likely sidestep it, since the central directory carries correct sizes regardless of the descriptor flag.

## Environment

- Floci 1.7.0, `floci/floci@sha256:bec9b9f749322444a57fb0f00d3e007ffd49015162361aab84ccecd9b4f5f8ed`
- aws-cli 2.34.10
- Serverless Framework 3.39.0
- macOS arm64
- Docker 29.5.3

avison9:
## Summary

Closes #2593.

`CreateFunction` rejected any deployment package produced by a streaming zip writer, failing with
`Failed to extract deployment package: only DEFLATED entries can have EXT descriptor`. Real AWS
accepts the same archive. In practice this makes **any Serverless Framework project that bundles
`node_modules` undeployable against Floci** — the reporter's only workaround was to drop
dependencies from the package entirely. Routed through CloudFormation the same failure surfaced as
the more confusing `Handler file '<handler>' not found in deployment package`.

`ZipExtractor` read the archive with `ZipInputStream`, a forward-only reader that sees only *local
file headers*. It now reads through the **central directory** via `ZipFile`.

### Before / after

The trigger needs an entry that is **STORED** *and* carries a trailing data descriptor. Reduced to
a 169-byte archive (the same shape the new test builds):

```java
byte[] zip = storedWithDataDescriptor("handler.js", "exports.handler = async () => 'ok';");

// BEFORE — ZipInputStream (sequential, local headers):
try (ZipInputStream zis = new ZipInputStream(new ByteArrayInputStream(zip))) {
    while (zis.getNextEntry() != null) { /* ... */ }
}
// -> ZipException: only DEFLATED entries can have EXT descriptor     ← CreateFunction 400s

// AFTER — ZipFile (central directory):
try (ZipFile zf = new ZipFile(stagedFile)) {
    ZipEntry e = zf.entries().nextElement();
    System.out.println(e.getName() + " = " + new String(zf.getInputStream(e).readAllBytes()));
}
// -> handler.js = exports.handler = async () => 'ok';                ← extracts, like real AWS
```

## Root cause

A zip writer that streams its output cannot seek backwards to patch an entry's header, so it sets
general-purpose bit 3, zeroes the CRC and sizes in the local header, and appends the real values in
a trailing data descriptor — and separately writes incompressible content (`.node` addons, bundled
binaries under `node_modules`) as STORED. `ZipInputStream.readLOC()` refuses that pairing outright,
legitimately so for a forward-only reader: a STORED entry with deferred sizes has no end marker, so
it cannot know where the data stops. **`ZipExtractor` was treating local file headers as
authoritative when the central directory is** — the central directory always carries the true
method, CRC and sizes regardless of the descriptor flag.

This is also why the issue resisted reduction. Size was never the variable: a `zip(1)` archive
seeks and patches headers so bit 3 is never set; STORED-alongside-DEFLATED isn't enough because the
flag has to be set *on the stored entry*; Zip64 is orthogonal; and the 108 KB package had nothing
incompressible to store.

`ZipExtractor.extractTo` reads via `ZipFile` instead of `ZipInputStream`. `ZipFile` needs a seekable
file, so the incoming `byte[]` is staged to a temp file that is deleted in a `finally` block.

One compensating addition is required. **`ZipInputStream` verifies each entry's CRC;
`ZipFile.getInputStream` does not** — so swapping readers naively would have silently dropped an
integrity check the old code got for free, letting a package corrupted in transit deploy as garbage
instead of failing. Each entry is therefore streamed through a `CRC32` and checked against
`ZipEntry.getCrc()`, throwing a `ZipException` worded like the JDK's own. This restores parity
rather than adding new behaviour.

Everything else is deliberately untouched — the backslash warning, the traversal and containment
guards and directory handling are character-for-character identical, only re-indented, with the
`closeEntry()` calls dropped since `ZipFile` has no per-entry cursor. The signature is unchanged,
so neither caller (`LambdaService`, `LambdaLayerService`) needed changing.

Alternatives considered: **Apache Commons Compress** would avoid the temp file but is not a
dependency, and pulling in a compression library to swap one reader is disproportionate.
**Falling back to `ZipFile` only after `ZipInputStream` throws** would keep the broken reader on the
primary path and leave a half-extracted directory behind when it fails part-way (verified — the
first entry lands on disk before the throw). Reading the central directory is the correct behaviour,
not a fallback. **zipfs** is equally correct but needs the same temp file plus a provider round-trip
for no benefit.

Cost is one extra sequential write, bounded by the AWS package limit, on the deploy path only —
never on invoke. `Files.createTempFile` creates the file owner-only, so staging doesn't widen
exposure of package contents.

## AWS Compatibility

Real AWS Lambda accepts these archives; Floci did not. The two readers differ in both directions
and both were checked:

- **Stricter:** `ZipFile` requires a well-formed central directory, so a badly truncated archive is
  rejected up front rather than partially extracted. AWS-congruent. Both of Floci's own zip
  producers (`sourceToZipBase64`, `defaultHandlerZipBase64`) use `ZipOutputStream`, which always
  writes a central directory, so no internal path is affected.
- **More tolerant:** prefixed self-extracting archives and fuller Zip64 support now work.
- **Laxer — and compensated:** no per-entry CRC verification, restored explicitly as described
  above. Without that the fix would have traded a loud failure for a silent one.

## How it was tested

Four tests added to `ZipExtractorTest`. Three use a new `streamedZip(...)` helper that assembles zip
bytes by hand — required because `ZipOutputStream` structurally **cannot** emit an entry that is
STORED *and* descriptor-flagged, which is why the existing suite could never have caught this.

| Test | Covers |
|---|---|
| `extractsStoredEntryCarryingADataDescriptor` | The regression test: a DEFLATED `index.js` beside a STORED `node_modules/sharp/sharp.node`, bit 3 set on both — a real Serverless bundle in miniature |
| `extractsDeflatedEntryCarryingADataDescriptor` | Do-no-harm guard: the descriptor flag on a DEFLATED entry always worked and still must |
| `extractsExplicitDirectoryEntries` | Trailing-slash directory entries still create directories via the central directory |
| `rejectsAnEntryThatFailsItsChecksum` | CRC parity: a structurally valid archive whose data was corrupted after the fact must be rejected, not extracted silently |

Reverting only `ZipExtractor.java` and re-running the class reproduces the reported error from the
test suite:

```
[ERROR] Tests run: 6, Failures: 0, Errors: 2
[ERROR] ZipExtractorTest.extractsStoredEntryCarryingADataDescriptor:201
          » Zip only DEFLATED entries can have EXT descriptor
```

| Stage | Result |
|---|---|
| Baseline, clean `main` | `Tests run: 13283, Failures: 0, Errors: 0, Skipped: 9` |
| Full suite after fix | `Tests run: 13287, Failures: 0, Errors: 0, Skipped: 9` |
| `make docs-test` / `make docs-check` | 34 passed / exit 0 |

**+4 tests, +0 failures** — exactly the four added; no pre-existing test changed state.
Java 25 / Maven 3.9.16, matching `ci.yml`.

## Risk / rollback

Low. A leaf utility with two callers, both already handling `IOException`; signature unchanged; no
persisted state or wire format involved. The change is additive in capability — archives that
worked keep working, a class that failed now succeeds. Rollback is reverting the single file.

## Follow-ups (not in this PR)

Two adjacent defects found while diagnosing, deliberately left out to keep this focused — happy to
file them separately:

1. **Partial extraction isn't cleaned up on failure.** `extractTo` writes entries as it goes, so a
   mid-archive failure leaves a half-populated code directory. This is the likely mechanism behind
   the misleading CloudFormation `Handler file not found` symptom: a later path finds a non-empty
   code directory and skips re-extraction. Extracting to a temp dir and moving it into place
   atomically would fix the class. This PR removes the trigger for #2593 but not that fragility.
2. **CloudFormation masks the underlying extraction error**, reporting handler resolution instead of
   `Failed to extract deployment package: …`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,14 @@
             <artifactId>docker-java-core</artifactId>
             <version>3.7.1</version>
         </dependency>
+        <!-- Central-directory ZIP reading for Lambda deployment packages. Already present
+             transitively via docker-java-core; declared explicitly because ZipExtractor uses
+             it directly and must not break if that transitive path ever changes. -->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>1.28.0</version>
+        </dependency>
         <dependency>
             <groupId>com.github.docker-java</groupId>
             <artifactId>docker-java-transport-httpclient5</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -158,14 +158,6 @@
             <artifactId>docker-java-core</artifactId>
             <version>3.7.1</version>
         </dependency>
-        <!-- Central-directory ZIP reading for Lambda deployment packages. Already present
-             transitively via docker-java-core; declared explicitly because ZipExtractor uses
-             it directly and must not break if that transitive path ever changes. -->
-        <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-compress</artifactId>
-            <version>1.28.0</version>
-        </dependency>
         <dependency>
             <groupId>com.github.docker-java</groupId>
             <artifactId>docker-java-transport-httpclient5</artifactId>
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/zip/ZipExtractor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/zip/ZipExtractor.java
@@ -1,6 +1,9 @@
 package io.github.hectorvent.floci.services.lambda.zip;
 
 import jakarta.enterprise.context.ApplicationScoped;
+import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
+import org.apache.commons.compress.archivers.zip.ZipFile;
+import org.apache.commons.compress.utils.SeekableInMemoryByteChannel;
 import org.jboss.logging.Logger;
 
 import java.io.IOException;
@@ -9,9 +12,7 @@ import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.zip.CRC32;
-import java.util.zip.ZipEntry;
 import java.util.zip.ZipException;
-import java.util.zip.ZipFile;
 
 /**
  * Extracts ZIP bytes to a target directory.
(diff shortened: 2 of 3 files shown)


avison9:
Thanks — the P2 is fair, and it's a regression this PR introduced rather than a pre-existing one. Before this change `ZipExtractor` read straight from the in-memory `byte[]` and touched no temp filesystem at all; moving to `ZipFile` (which needs a seekable source) added a brand-new consumer of shared `java.io.tmpdir`.

Fixed in 248ccc20:

```java
- Path staged = Files.createTempFile("floci-lambda-", ".zip");
+ Path staged = Files.createTempFile(absTarget.getParent(), ".floci-staging-", ".zip");
```

Staging now lands in the **Lambda code store** instead of the shared temp filesystem. That volume already has to hold this package's *uncompressed* contents — always at least as large as the archive, usually much larger — so it is already provisioned for the size, and the staged copy is released in a `finally` as soon as extraction ends. The part of the finding that actually mattered was the blast radius: exhausting `/tmp` fails unrelated temp-file users across the process, whereas pressure on the code store is confined to the subsystem that caused it and is bounded by what extraction already required.

Added `leavesNoStagingArtefactsInTheCodeStore` to pin that nothing survives, on either the success or the failure path. Worth being explicit about its limits: **it does not prove the staging location.** `java.nio.file.TempFileHelper` resolves `java.io.tmpdir` once into a `private static final Path` at class-init, so the property cannot be redirected mid-test to assert which directory was used — proving that would need a seam added purely for the test. What the test does guard is real, though: a leaked staging artefact would accumulate one per deploy in the code store, and one leaked *inside* a function directory would be tar-copied into that function's container at launch.

On the second half of the finding — enforcing Lambda's package limit rather than the ~2 GB request limit — I agree it's a genuine gap, but I'd argue it belongs in its own PR:

- It is not introduced here. A 2 GB package was already fully materialised in memory as a `byte[]` before this change (from the base64 decode or the S3 fetch), which is a worse failure mode than on disk. This PR inherits that ceiling, it doesn't widen it.
- Enforcing AWS's documented caps (50 MB zipped direct upload / 250 MB unzipped) is real AWS-parity work, but it is a **behaviour change** — it would start rejecting packages Floci accepts today, which is the kind of thing that deserves its own issue, its own tests and a maintainer decision, rather than riding along in a bug fix.

Happy to file that as a separate issue and take it if maintainers want it.

Verification: full suite `Tests run: 13288, Failures: 0, Errors: 0, Skipped: 9` against a `13283` baseline on clean `main` — +5, exactly the tests added here, no pre-existing test changed state.

avison9:
Fixed in 7cb44767 — and the fix supersedes my previous one, so my earlier comment is now out of date on the details.

Rather than bound the staged copy, I removed it. **commons-compress 1.28.0 and commons-io 2.22.0 are already compile-scope dependencies here** (transitively via `docker-java-core`), which invalidates the reasoning behind my original design note: I had ruled commons-compress out as a disproportionate *new* dependency, and it is not a new dependency at all.

```java
try (ZipFile zip = ZipFile.builder()
        .setSeekableByteChannel(new SeekableInMemoryByteChannel(zipBytes))
        .get()) {
```

Commons Compress reads the same central directory straight off the bytes already held, so **no copy of the archive is written anywhere**. Extraction now costs exactly what the extracted output needs — the disk profile is identical to `main` before this branch, since the pre-change code also read from the in-memory `byte[]`. There is no staging to bound, throttle, or cap, so the concurrency and ENOSPC concerns are removed rather than mitigated. `commons-compress` is declared explicitly in `pom.xml` because `ZipExtractor` now uses it directly and should not break if docker-java's transitive graph changes.

I also removed the `leavesNoStagedArchiveInTheCodeStore` test from my previous commit: it guarded a staged file that no longer exists, and keeping it would be coverage in name only.

Switching libraries introduced two behaviour differences from `java.util.zip` that the suite caught, both now restored explicitly:

1. **Backslash entry names.** Commons Compress rewrites every `\` to `/` for a FAT-platform entry whose name contains no `/`. That turns a PowerShell-produced literal filename into a nested path — the opposite of what #1215 deliberately established, since real AWS Lambda extracts `wwwroot\app.css` as one flat file. The existing `extractsBackslashEntriesAsLiteralFilename` test failed immediately on the switch. The raw name field holds the bytes as written, so the rewrite is undone only when it carries a backslash and no forward slash; reversing it on the decoded name rather than decoding the raw bytes leaves Commons Compress's UTF-8-flag/CP437 handling intact.

2. **Unknown compression methods.** `java.util.zip` throws; Commons Compress reports the entry unreadable and hands back an empty stream, which would write a zero-byte file and deploy a function whose code had silently vanished. Entries that cannot be read are now rejected, with a test using method 99 (WinZip AES).

On the package-size ceiling itself: I still think that belongs in its own issue rather than here. It is not introduced by this PR — a multi-gigabyte package was already fully materialised in heap as a `byte[]` before this branch, from the base64 decode or the S3 fetch — and enforcing AWS's documented caps (50 MB zipped direct upload / 250 MB unzipped) is a **behaviour change** that would start rejecting packages Floci accepts today. That deserves its own issue, tests and maintainer decision. Happy to file it and take it.

Verification: full suite `Tests run: 13288, Failures: 0, Errors: 0, Skipped: 9`, BUILD SUCCESS, against a `13283` baseline on clean `main` — +5, exactly the tests added here, no pre-existing test changed state.

avison9:
Update: I have reverted the in-memory reader in b577ae1a. It removed the staged copy, but **it cannot be built into this project's native image**, which is a shipped artefact — so the trade does not hold. My previous comment is superseded on that point.

`org.apache.commons.compress.archivers.zip.ZipFile.getInputStream` hard-references four codecs, and two of them — **xz** and **zstandard** — require libraries deliberately absent from the classpath. The build uses `--link-at-build-time`, so every reachable method must resolve, and native-image fails during analysis:

```
Discovered unresolved method during parsing:
  org.apache.commons.compress.compressors.xz.XZCompressorInputStream.<init>(java.io.InputStream)
Error encountered while parsing ZipFile.getInputStream(ZipFile.java:1226)
Parsing context:
   at ZipExtractor.copyVerified(ZipExtractor.java:135)
```

This is a **linking** failure, not a class-initialization one, so `--initialize-at-run-time` does not fix it — I verified that by building the image locally with `--initialize-at-run-time=org.apache.commons.compress.compressors` in place and getting the identical error. Making it link would mean adding two compression libraries, one of them shipping JNI natives, purely so GraalVM can resolve code paths a Lambda package cannot exercise: AWS deployment packages are STORED or DEFLATE only.

So the implementation is back to `java.util.zip.ZipFile` with the archive staged in the code store. Verified locally: native image `Finished generating 'floci-1.7.0-runner' in 7m 20s`, BUILD SUCCESS.

On the remaining "staging is unbounded" concern, my position — stated plainly so maintainers can overrule it:

- The staged copy is **not unbounded relative to the operation**. It is at most one archive on the same volume that must already absorb that package's *uncompressed* contents, which are always at least as large and usually several times larger. Extraction cannot succeed on a volume that lacks room for the staged copy.
- The genuinely unbounded thing is the **package size ceiling itself**, and that predates this PR: a multi-gigabyte package was already fully materialised in heap as a `byte[]` before this branch, from the base64 decode or the S3 fetch. This PR neither introduces nor widens it.
- Enforcing AWS's documented caps (50 MB zipped direct upload / 250 MB unzipped) is the right fix for that, and it is a **behaviour change** — it would start rejecting packages Floci accepts today. It deserves its own issue, its own tests and a maintainer decision, exactly like #2648.

I am happy to file that issue and take it; I would rather not fold a user-visible rejection rule into a bug fix without maintainers weighing in. If you would prefer it in this PR instead, say so and I will add it here.

avison9:
Follow-up #2672 is now open and stacked on this branch — it fixes #2647 (extraction merges into the previous deployment instead of replacing it, so a file deleted between deploys survives and stays loadable). Same file and same method, which is why it is stacked rather than opened independently; its diff collapses to a two-file change once this merges. No action needed here.

pgermosen (approved):
Very high-quality work, and I went through the current state independently rather than taking the earlier back-and-forth at face value. The core fix is right: `ZipInputStream` really does throw "only DEFLATED entries can have EXT descriptor" for any STORED entry that also carries the data-descriptor flag — long-standing JDK behavior, not a mischaracterization — and reading through `ZipFile`'s central directory sidesteps it cleanly, since the central directory always carries the true method/CRC/sizes regardless of what the local header's flag says.

I checked the traversal, containment, and backslash guards character-for-character against the diff — untouched, just re-indented. The CRC restoration in `copyVerified` is correct, including the edge case on the unsupported-compression-method test: `zip.getInputStream(entry)` throws before `Files.newOutputStream` is ever reached, since both are resource initializers in the same try-with-resources and Java evaluates them in declaration order, so the target file is genuinely never created on that path.

I also traced the staging design against the rest of the code store rather than trusting the description: the temp file lands as a sibling of the function's own code directory, not inside it, and neither the container-launch path (which tar-copies `codePath` itself) nor the legacy-directory-reclaim logic (which never lists the account-level directory) can be confused by a staged file sitting there mid-extraction. Reverting the Commons Compress attempt after finding it hard-references codecs that fail `native-image`'s link-time analysis was a sharp catch — that's a real constraint, not a rationalization for going back to the simpler approach.

On the still-open disagreement with the bot about unbounded staging: I read through the reasoning and it holds up independently. The staged copy sits on a volume that already has to hold this package's uncompressed contents, always at least as large — so it isn't a new failure mode, it's smaller than one the operation already required. The real gap, enforcing AWS's actual package-size ceiling, predates this change entirely (a multi-gigabyte package was already fully materialized in heap before this PR) and is a genuine behavior change that deserves its own issue rather than riding along here.

One thing worth knowing for sequencing: the "partial extraction isn't cleaned up on failure" item disclosed as a follow-up in this PR looks like it'll be resolved for free once the already-open stacked PR lands after this one, since its staging-directory-plus-atomic-move design covers the whole extraction, not just this path.

PR #2650, opened 2026-08-28
fix(cloudformation): fail the stack when Lambda S3 code cannot be read

### LINKED ISSUE #2648: [BUG] cloudformation: unreadable Lambda S3 code silently deploys a stub handler instead of failing the stack

avison9:
## ⚠️ Read this first — this changes behaviour for stacks that pass today

Closes #2648.

**Most bug fixes only change what happens in situations that were already visibly broken. This one
changes what happens in a situation that currently looks like success**, so it deserves an explicit
decision from you rather than a quiet merge.

Floci has a shortcut: when a template says *"my Lambda's code is at `s3://bucket/key`"* and that
object cannot be read, it silently substitutes a 176-byte built-in placeholder that returns
`{"statusCode":200}` and nothing else.

| | Today | With this PR |
|---|---|---|
| Stack status | `CREATE_COMPLETE` | `CREATE_FAILED` → rollback |
| Function created? | Yes — the placeholder | No |
| What the user is told | Nothing (one line in a container log) | An error naming the bucket and key |

### Who this could hurt

Someone with a 40-resource stack — DynamoDB tables, SQS queues, IAM roles, three Lambdas — who is
working on the DynamoDB side today and has not uploaded the Lambda zips. Right now that stack
deploys: the Lambdas come up as placeholders, everything else provisions, and they get on with
their work. **After this change the whole stack fails and rolls back**, and they cannot test their
DynamoDB work until they build and upload three packages they do not care about. In CI, that is a
pipeline going from green to red on a Floci upgrade with no change on their side.

I think the strict behaviour is still right — a stack reporting success while running code the user
never wrote is genuinely dangerous, and it is what produced the confusing error in #2593. But that
is a judgement call about your users, so I have not made it unilaterally.

### Three ways to take this — happy to push whichever you prefer

1. **As-is (strict).** Unreadable code always fails the stack. AWS-accurate; breaks the workflow
   above.
2. **Both behaviours, strict by default** *(my recommendation if you want an escape hatch)* — a
   config flag, e.g. `floci.cloudformation.lambda.allow-stub-code`, defaulting to `false`. Correct
   out of the box, and anyone relying on placeholder Lambdas sets one property instead of being
   blocked. Costs a config key and a line in the docs.
3. **Both behaviours, lenient by default.** Same flag, defaulting to `true`. Nobody's stack changes
   on upgrade; users opt into correctness.

Say which you want and I will push it — option 2 or 3 is a small change on top of this branch, and
I have already written the reasoning for each. I defaulted to option 1 only because
`CONTRIBUTING.md` and `AGENTS.md` both state that matching the real AWS wire protocol is the
project's default.

---

## Summary

When a template declares `Code: { S3Bucket, S3Key }` and Floci cannot read that object, the
provisioner silently substituted a built-in stub package instead of failing. That split two ways,
both bad:

**Variant 1 — `Handler: index.handler` (the stub's own handler).** The stack reaches
`CREATE_COMPLETE`, the function is invocable, and it returns `{"statusCode":200}` for every
request. The user's code was never deployed and nothing in the API surface says so — only a
container log line. An integration suite run against that stack passes green against a stub.

**Variant 2 — any other handler.** The stub lacks the module, so `CreateFunction` fails its handler
check and the resource fails with `Handler file 'src/api' not found in deployment package` —
naming a handler that is correct in the template and present in the real package. This is the
misleading error reported in the second half of #2593.

### Before / after

Template pointing at `s3://pkg/does-not-exist.zip`:

```
# BEFORE — variant 1 (Handler: index.handler)
describe-stacks   CREATE_COMPLETE          <- real CloudFormation: CREATE_FAILED / ROLLBACK
get-function      index.handler   176      <- 176 bytes: the built-in stub
invoke            {"statusCode":200}       <- the stub's canned response

# BEFORE — variant 2 (Handler: src/api.handler)
describe-stack-events  Fn  CREATE_FAILED  Handler file 'src/api' not found in deployment package

# AFTER — both variants
describe-stack-events  Fn  CREATE_FAILED  Error occurred while GetObject. S3 Error Message: ...
                                          (bucket: pkg, key: does-not-exist.zip)
```

## Root cause

`resolveLambdaCode` probed the object only to decide whether to use it, and treated *any* failure
as "no code available":

```java
try {
    s3Service.getObject(s3Bucket, s3Key);
    return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key), ...);
} catch (Exception e) {
    LOG.warnv("S3 code not found for Lambda ({0}/{1}), using default handler: {2}", ...);
}
...
return new LambdaCodeSpec(Map.of("ZipFile", defaultHandlerZipBase64()), "default-handler");
```

**A fallback default is only sound when the input is absent.** Here it was applied to an input that
was *present but unusable*, turning an explicit, checkable user intent into a silent substitution.
The bare `catch (Exception e)` compounds it: a transient S3 error, a permissions failure, or a bug
in the read path all degrade to the same stub as a genuinely missing key.

`resolveLambdaCode` now throws `AwsException("ValidationError", …)` when an explicitly declared
`S3Bucket`/`S3Key` cannot be read, so the resource fails and the stack rolls back the way real
CloudFormation does. The message names the bucket and key and carries the underlying S3 error.

The stub fallback is **kept for its legitimate case** — a template supplying no `Code` block at all
— which is the unchanged `return` at the end of the method. The inline `ZipFile` and `ImageUri`
paths are untouched. The defect was the scope the fallback was applied at, not the fallback itself.

Alternatives: **keeping the stub but marking the resource `CREATE_FAILED`** is contradictory — a
failed resource that still deployed something leaves an orphan function. **Surfacing a stack event
instead of failing** still ends in `CREATE_COMPLETE`, so automation cannot tell the difference and
the core problem is unchanged.

## AWS Compatibility

Real CloudFormation fails the resource with an error naming the inaccessible code
(`Error occurred while GetObject. S3 Error Code: NoSuchKey`) and rolls the stack back; a stack that
reaches `CREATE_COMPLETE` has deployed the code the template pointed at. Floci now matches. For an
emulator whose purpose is to predict what AWS does, a local stack going green where AWS would roll
back is worse than useless — it actively misleads.

## How it was tested

New `CloudFormationLambdaMissingS3CodeIntegrationTest` (`@QuarkusTest`, control-plane only, so
Docker-free), following the `CloudFormationLambdaAddressingIntegrationTest` pattern:

| Test | Covers |
|---|---|
| `unreadableS3CodeFailsTheStackInsteadOfDeployingAStub` | Variant 1. Asserts the stack does not reach `CREATE_COMPLETE` and the function does not exist at all (404) — a stub must never stand in for the real package |
| `theFailureNamesTheUnreadableCodeNotTheHandler` | Variant 2. Asserts stack events name `does-not-exist.zip` and do **not** contain `not found in deployment package` |

Both fail against unfixed code (`Tests run: 2, Failures: 2`). Full suite: see the checklist.
Java 25 / Maven 3.9.16, matching `ci.yml`.

## Risk / rollback

Covered at the top: this is an intentional behaviour change, and options 2 and 3 there are both
one small commit away if you would rather not change outcomes for existing users. Rollback is
reverting the single method — no persisted state, no wire format, no migration.

## Follow-up (not in this PR)

The probe reads the whole object via `s3Service.getObject` purely to test readability and then
discards it; `LambdaService` fetches it again during `CreateFunction`. For a 250 MB package that is
a wasted full read per stack operation. A `headObject`-style check, or threading the fetched bytes
through, would fix it — a performance concern rather than a correctness one, so it belongs
elsewhere. Happy to file it.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
+++ b/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
@@ -1293,6 +1293,19 @@ public interface EmulatorConfig {
 
         @WithDefault("30")
         long deletedStackRetentionSeconds();
+
+        /**
+         * Whether an {@code AWS::Lambda::Function} whose template names code in S3 that cannot be
+         * read should fall back to the built-in stub handler instead of failing the resource.
+         *
+         * <p>Defaults to {@code false}, matching real CloudFormation, which fails the resource and
+         * rolls the stack back. Set to {@code true} to restore the older behaviour for a stack that
+         * deliberately leaves Lambda packages unbuilt — note that such a stack reports
+         * {@code CREATE_COMPLETE} while serving a placeholder that returns
+         * {@code {"statusCode":200}}, so it cannot be used to verify the real function.
+         */
+        @WithDefault("false")
+        boolean allowStubLambdaCode();
     }
 
     interface AcmServiceConfig {
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -100,6 +100,7 @@ import io.github.hectorvent.floci.services.apigatewayv2.model.*;
 import io.github.hectorvent.floci.services.cognito.CognitoService;
 import io.github.hectorvent.floci.services.cognito.model.UserPool;
 import io.github.hectorvent.floci.services.cognito.model.UserPoolClient;
+import io.github.hectorvent.floci.config.EmulatorConfig;
 import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.core.common.docker.ContainerReachableEndpoint;
 import io.github.hectorvent.floci.services.lambda.model.InvocationType;
@@ -220,6 +221,7 @@ public class CloudFormationResourceProvisioner {
     // below. As types migrate, their switch cases and provisionXxx methods are removed here; the
     // now-dead service deps above are cleared in the final cleanup once the switch is empty.
     private final CloudFormationResourceRegistry resourceRegistry;
+    private final EmulatorConfig config;
 
     @Inject
     public CloudFormationResourceProvisioner(S3Service s3Service, SqsService sqsService,
@@ -251,7 +253,9 @@ public class CloudFormationResourceProvisioner {
                                              FirehoseService firehoseService,
                                              DocDbService docDbService,
                                              CloudFrontService cloudFrontService,
-                                             CloudFormationResourceRegistry resourceRegistry) {
+                                             CloudFormationResourceRegistry resourceRegistry,
+                                             EmulatorConfig config) {
+        this.config = config;
         this.s3Service = s3Service;
         this.sqsService = sqsService;
         this.snsService = snsService;
@@ -2399,6 +2403,15 @@ public class CloudFormationResourceProvisioner {
                 reservedConcurrentExecutions);
     }
 
+    /**
+     * Whether an unreadable explicit {@code Code} reference may fall back to the stub handler.
+     * The provisioners hand-built in unit tests carry no config; absent configuration means the
+     * documented default, which is the strict behaviour.
+     */
+    private boolean stubLambdaCodeAllowed() {
+        return config != null && config.services().cloudformation().allowStubLambdaCode();
+    }
+
     private LambdaCodeSpec resolveLambdaCode(JsonNode props, CloudFormationTemplateEngine engine,
                                              String handler, String runtime) {
         if (props != null && props.has("Code")) {
@@ -2413,15 +2426,25 @@ public class CloudFormationResourceProvisioner {
                 // — or, when the handler was not "index.handler", fail with a handler error
                 // that pointed away from the real problem (issue #2648). The stub below is for
                 // a template that supplies no Code at all, which is a different case.
+                //
+                // allow-stub-lambda-code opts back in to the old fallback, for a stack that
+                // deliberately leaves its Lambda packages unbuilt and only cares about the
+                // other resources. Off by default: silently serving a placeholder is the more
+                // dangerous of the two behaviours.
                 try {
                     s3Service.getObject(s3Bucket, s3Key);
+                    return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key),
+                            "s3:" + s3Bucket + "\n" + s3Key);
                 } catch (Exception e) {
-                    throw new AwsException("ValidationError",
-                            "Error occurred while GetObject. S3 Error Message: " + e.getMessage()
-                                    + " (bucket: " + s3Bucket + ", key: " + s3Key + ")", 400);
+                    if (!stubLambdaCodeAllowed()) {
+                        throw new AwsException("ValidationError",
+                                "Error occurred while GetObject. S3 Error Message: " + e.getMessage()
+                                        + " (bucket: " + s3Bucket + ", key: " + s3Key + ")", 400);
+                    }
+                    LOG.warnv("S3 code not found for Lambda ({0}/{1}), using default handler because "
+                                    + "floci.services.cloudformation.allow-stub-lambda-code is enabled: {2}",
+                            s3Bucket, s3Key, e.getMessage());
                 }
-                return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key),
-                        "s3:" + s3Bucket + "\n" + s3Key);
             }
 
             String zipFile = codeNode.path("ZipFile").asText(null);
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -2407,14 +2407,21 @@ public class CloudFormationResourceProvisioner {
             String s3Bucket = codeNode.path("S3Bucket").asText(null);
             String s3Key = codeNode.path("S3Key").asText(null);
             if (s3Bucket != null && s3Key != null) {
+                // A template that names its code explicitly must fail if that code cannot be
+                // read, the way real CloudFormation does. Substituting the stub handler here
+                // let a stack reach CREATE_COMPLETE running code the template never referenced
+                // — or, when the handler was not "index.handler", fail with a handler error
+                // that pointed away from the real problem (issue #2648). The stub below is for
+                // a template that supplies no Code at all, which is a different case.
                 try {
                     s3Service.getObject(s3Bucket, s3Key);
-                    return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key),
-                            "s3:" + s3Bucket + "\n" + s3Key);
                 } catch (Exception e) {
-                    LOG.warnv("S3 code not found for Lambda ({0}/{1}), using default handler: {2}",
-                              s3Bucket, s3Key, e.getMessage());
+                    throw new AwsException("ValidationError",
+                            "Error occurred while GetObject. S3 Error Message: " + e.getMessage()
+                                    + " (bucket: " + s3Bucket + ", key: " + s3Key + ")", 400);
                 }
+                return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key),
+                        "s3:" + s3Bucket + "\n" + s3Key);
             }
 
             String zipFile = codeNode.path("ZipFile").asText(null);
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -365,6 +365,9 @@ floci:
       timeout-sweep-interval-seconds: 1
     cloudformation:
       enabled: true
+      # Fail an AWS::Lambda::Function whose S3 code cannot be read, as real CloudFormation
+      # does. Set true to fall back to the built-in stub handler instead (issue #2648).
+      allow-stub-lambda-code: false
     acm:
       enabled: true
       validation-wait-seconds: 0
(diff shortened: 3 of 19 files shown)


pgermosen (commented):
Small, correct, well-tested fix — and appreciated that you flagged the behavior change explicitly instead of merging it quietly.

The fix is scoped exactly where the bug was, not more broadly: the stub fallback for a template with no `Code` block at all is untouched, and only the branch where `S3Bucket`/`S3Key` are explicitly present and unreadable now throws instead of silently substituting. That's the right boundary — a fallback is sound when the input is absent, not when it's present but broken. Both failure variants from the linked issue are covered, and the tests would have caught the original bug: the first asserts the function doesn't exist at all (404) after the stack fails, not just that the stack status changed, which is the check that actually proves no stub got deployed; the second checks the stack event names the missing zip key and does not contain the old misleading "not found in deployment package" text. The async CloudFormation semantics are also respected — `CreateStack` still returns 200 immediately in the tests, with the failure asserted through `DescribeStacks`/`DescribeStackEvents` afterward, matching how this works on real AWS.

On the behavior-change question you raised: going with option 2 — strict by default, with an escape-hatch config flag (e.g. `floci.cloudformation.lambda.allow-stub-code`, default `false`) so a stack currently relying on the stub-Lambda shortcut can opt back in with one property instead of breaking on upgrade. This matches the shape this codebase already reaches for elsewhere when a correctness fix could break existing users. The change needed on top of this branch: gate the new `throw` on the flag being false/absent, and fall through to the existing stub-fallback return when it's true.

The follow-up you flagged yourself (the double S3 read — once to probe readability here, once by `LambdaService` on the actual `CreateFunction`) is a real, separate concern and agreed it belongs in its own issue rather than this PR.

avison9:
Thanks for the review. Option 2 is implemented; the double read is filed separately as you suggested.

## Option 2: strict by default, escape hatch available

The throw is now gated on the flag, falling through to the existing stub return when it is enabled.

```java
try {
    s3Service.getObject(s3Bucket, s3Key);
    return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key), ...);
} catch (Exception e) {
    if (!stubLambdaCodeAllowed()) {
        throw new AwsException("ValidationError", "Error occurred while GetObject. ...", 400);
    }
    LOG.warnv("S3 code not found for Lambda ({0}/{1}), using default handler because "
            + "floci.services.cloudformation.allow-stub-lambda-code is enabled: {2}", ...);
}
```

The knob is on `EmulatorConfig.CloudFormationServiceConfig` with `@WithDefault("false")`, plus
entries in the main and test `application.yml`, per AGENTS.md. Its javadoc states the trade so the
flag is not enabled blind: a stack with it on reports `CREATE_COMPLETE` while serving a placeholder
returning `{"statusCode":200}`, so it cannot be used to verify the real function.

Both behaviours are pinned by tests now, not only the default. A nested `@TestProfile` class turns
the flag on and asserts the stack completes and the function exists, which is precisely what someone
would be opting back in to.

### Two things I want you to check rather than assume

**The property name differs from yours.** You wrote `floci.cloudformation.lambda.allow-stub-code`.
I used `floci.services.cloudformation.allow-stub-lambda-code` so it sits under the
`floci.services.<service>.*` shape `EmulatorConfig` already uses, next to `enabled` and
`deleted-stack-retention-seconds`. If you meant that name literally, or want a nested `lambda` group,
I will rename it: one line plus the two YAML entries.

**The diff grew from 2 files to 18.** Reading the flag means the provisioner needs `EmulatorConfig`,
and that constructor parameter broke 13 unit tests that build `CloudFormationResourceProvisioner`
positionally with `null`s. I updated all 13 and made the read null safe, since those hand built
provisioners carry no config and absent configuration should mean the documented default:

```java
private boolean stubLambdaCodeAllowed() {
    return config != null && config.services().cloudformation().allowStubLambdaCode();
}
```

If you would rather not touch those 13 files, field injection or reading the property at the call
site would both avoid it. I chose constructor injection because it matched the surrounding style,
but the churn is real and I would rather you see it than find it.

## The double S3 read

Filed as #2675 and claimed. `resolveLambdaCode` calls `getObject` purely as a probe and throws the
result away, then `LambdaService.extractZipCodeFromS3` fetches the same object again. Because
`S3Object.getData()` is a materialised byte array, that is a full second copy of the package in
memory per Lambda per stack operation, not a metadata lookup. The issue offers both fixes, a
metadata only probe or threading the fetched bytes through `LambdaCodeSpec`, without picking one.

## Verification

Full suite: `Tests run: 13286, Failures: 0, Errors: 0, Skipped: 9`, against a `13283` baseline on
clean `main`. That is plus 3: the two original tests and the new opt in case. No pre existing test
changed state.

One note on the test layout, since it explains why the opt in case is its own file. I first wrote it
as a static nested class inside the existing test. It passed when targeted directly, but Surefire
discovers test classes by file name, so a full `mvn test` skipped it silently and the suite count
came back one short of what I expected. It is now a top level class, which is also the shape
`@TestProfile` expects, and the full run confirms it executes.

pgermosen (approved):
Option 2 is implemented correctly, and both things you flagged for me to check rather than assume land in the right place.

The gating is exactly right: the restored fall-through only happens when the flag is on, and when it's off (the default) nothing changed from the last round. When it's on, the catch block logs and falls through to the same unchanged stub-fallback return the pre-fix code used — a precise restoration of the old behavior gated behind one flag, not a reimplementation of it, which is the safer shape for something meant to be opt-in only.

On the property name: your choice is the better one. I wrote the name in my last comment as an illustrative example, not a firm spec, and `floci.services.cloudformation.allow-stub-lambda-code` actually matches this codebase's real convention better — every service knob under `EmulatorConfig` sits directly under `floci.services.<service>.*`, not a nested per-feature group. Keep it as submitted.

On the 13-file test churn: also the right call. Constructor injection matches how every other config dependency reaches this class, and the null-safe read means none of the hand-built provisioners in existing tests needed new mocking — a mechanical, low-risk churn rather than something to push back on.

The new test proves the restored behavior itself, not just that the default is unchanged — it turns the flag on and asserts both that the stack completes and that the function actually exists at the stub's endpoint, which is the case someone opting back in is actually relying on. And catching the Surefire nested-class discovery issue before merge rather than after is exactly the kind of thing that's easy to miss.

The double-S3-read follow-up being filed and claimed separately is the right resolution for something flagged as out-of-scope-but-real.

Approving.

PR #2653, opened 2026-08-28
fix(s3): accept the log-delivery-write canned ACL

allensanborn:
## Summary

`PutBucketAcl`/`PutObjectAcl` rejected the `log-delivery-write` canned ACL with `InvalidArgument: Unsupported x-amz-acl value`, even though it's one of AWS's eight standard canned ACLs. Terraform's `aws_s3_bucket_acl` resource (and the access-logging modules that use it) send this value to grant the S3 log-delivery service group permission to write access logs into a bucket. This was the only canned ACL value missing from `cannedObjectAclXml`'s switch — the other seven (`private`, `public-read`, `public-read-write`, `authenticated-read`, `aws-exec-read`, `bucket-owner-read`, `bucket-owner-full-control`) were already implemented.

This PR adds the `log-delivery-write` case, translating it into `WRITE` and `READ_ACP` grants for the well-known LogDelivery group URI (`http://acs.amazonaws.com/groups/s3/LogDelivery`), following the same `groupGrant` pattern already used for `public-read`/`public-read-write`.

Found via the [floci-harness compatibility survey](https://github.com/allensanborn/floci-harness): 7 blocked roots across `terraform-aws-openvpn` and `terraform-aws-static-assets`, all failing on `PutBucketAcl` during S3 access-logging bucket setup.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Real AWS's canned ACL list includes `log-delivery-write`, which grants the S3 log-delivery group `WRITE` and `READ_ACP` on the bucket (documented at `http://acs.amazonaws.com/groups/s3/LogDelivery`). floci previously threw `InvalidArgument` for this value instead of applying the grant, so any Terraform module that sets up S3 access logging via a canned ACL (rather than explicit grants) failed at plan/apply time against floci.

## Checklist

- [x] `./mvnw test` passes locally — ran the full S3 suite (`mvn -Dtest='io.github.hectorvent.floci.services.s3.**' test`) after rebasing onto current `upstream/main`: 937 tests, 0 failures, 0 errors.
- [x] New or updated integration test added — `S3AclIntegrationTest` gains coverage for `log-delivery-write`.

### DIFF
--- a/docs/services/s3.md
+++ b/docs/services/s3.md
@@ -337,3 +337,4 @@ Current limitations:
 - explicit ACL grant headers such as `x-amz-grant-read` and `x-amz-grant-full-control` are not modeled yet
 - cross-account canned ACL variants collapse to the emulator's single synthetic owner where Floci does not model a distinct second principal
 - `aws-exec-read` is accepted for compatibility, but Floci does not yet model a distinct EC2 bundle-reader grantee in `GetObjectAcl`
+- `log-delivery-write` is accepted on `PutBucketAcl`/`PutObjectAcl` and grants the well-known S3 log-delivery group (`http://acs.amazonaws.com/groups/s3/LogDelivery`) `WRITE` and `READ_ACP`, matching AWS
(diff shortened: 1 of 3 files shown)


hectorvent:
Thanks for this, the fix is exactly right: the grant expansion (WRITE plus READ_ACP to the LogDelivery group at http://acs.amazonaws.com/groups/s3/LogDelivery) matches AWS's documented behavior for this canned ACL, and the Describe read-back test pins it properly.

**(follow-up, separate PR)** One small wording note: the new docs line says log-delivery-write is accepted on PutObjectAcl too, "matching AWS", but AWS documents this canned ACL as bucket-only (the SDK's ObjectCannedACL enum omits it). Floci's shared helper accepting it on both paths is a pre-existing pattern (aws-exec-read has the mirror-image situation), so no code change needed for this PR, just worth not claiming object-path parity in the docs.

No blockers from my side.

hectorvent (approved):

PR #2654, opened 2026-08-28
feat(ec2): support VPC peering connections

allensanborn:
## Summary

`CreateVpcPeeringConnection` was entirely unimplemented — it returned `UnsupportedOperation` — which blocked all three VPC-peering examples in `terraform-aws-vpc` (`vpc-peering`, `vpc-peering-cross-accounts`, `vpc-peering-external`). Found via the [floci-harness](https://github.com/allensanborn/floci-harness) compatibility survey, which exercises the Gruntwork Terraform corpus against Floci.

This adds the peering-connection lifecycle those examples actually drive:

- `CreateVpcPeeringConnection`
- `AcceptVpcPeeringConnection`
- `DescribeVpcPeeringConnections`
- `ModifyVpcPeeringConnectionOptions`
- `DeleteVpcPeeringConnection`

Connections start `pending-acceptance` and only move to `active` via an explicit accept, matching real AWS. Terraform's own `auto_accept` convenience (on `aws_vpc_peering_connection` and `aws_vpc_peering_connection_accepter`) is implemented by the *provider*, which just issues that second call itself — so no same-account special-casing is needed at the API layer.

`CreateRoute` now also accepts `VpcPeeringConnectionId` as a route target and reports it back on `DescribeRouteTables`. This is required, not optional: every one of the three examples routes traffic through the peering connection via `aws_route`, and without it the provider would see permanent drift on the route resource it just created.

### Design decision: connections are keyed by id alone, not region-scoped

Every other EC2 resource in this codebase is stored keyed by `region::id`. Peering connections are stored keyed by `vpcPeeringConnectionId` alone. This is deliberate: the `vpc-peering-cross-accounts`/cross-region example creates the connection from the requester's region but has the accepter operate against an independent `peer_region`. If storage were region-scoped, the accepter's `AcceptVpcPeeringConnection` and `DescribeVpcPeeringConnections` calls — issued against the *peer* region — would never find a connection stored under the *requester's* region, and the accept step would fail outright. Keying by id alone lets either side of the peering look the connection up regardless of which region's endpoint it calls through, which matches how AWS actually resolves a `pcx-*` id globally within a partition.

### Deliberately out of scope

- `RejectVpcPeeringConnection` — not exercised by any of the three target examples.
- CloudFormation support (`AWS::EC2::VPCPeeringConnection`) — same reason.

Both are natural follow-ups if a CFN template or example needs them later.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

New actions modeled: `CreateVpcPeeringConnection`, `AcceptVpcPeeringConnection`, `DescribeVpcPeeringConnections`, `ModifyVpcPeeringConnectionOptions`, `DeleteVpcPeeringConnection`, plus extending `CreateRoute`/`DescribeRouteTables` with `VpcPeeringConnectionId`. Verified against the EC2 query-protocol wire shape used by the current AWS Terraform provider / AWS CLI v2 (form-encoded `Action=...&Version=2016-11-15`, XML response), matching the pattern already used by this codebase's other EC2 integration tests (`io.restassured` driving the raw HTTP API).

Rebased onto current `upstream/main`, which in the meantime had merged #2620 (canonicalize `DestinationCidrBlock` at the `CreateRoute`/`ReplaceRoute`/`DeleteRoute` API boundary). Both changes touch `Ec2Service#createRoute`; the rebase carried both through cleanly — the merged method canonicalizes the CIDR destination via `canonicalizeIpv4Cidr(destinationCidrBlock)` *and* sets `route.setVpcPeeringConnectionId(vpcPeeringConnectionId)` on the constructed route. One pre-existing test (`Ec2ServicePersistenceTest`, added by #2620) called the old 8-argument `createRoute` overload directly; updated it to pass the new trailing `vpcPeeringConnectionId` argument as `null`.

## Checklist

- [x] `./mvnw test` passes locally — 1290/1290 in the `ec2` and `cloudformation` packages (0 failures, 0 errors), including the new `Ec2VpcPeeringConnectionIntegrationTest` (10 tests) and all pre-existing route tests (`Ec2DuplicateRouteTest`, `Ec2Ipv6RouteTest`, `Ec2RouteCidrCanonicalizationTest`, `Ec2ReplaceRouteIntegrationTest`, `Ec2ServicePersistenceTest`)

### DIFF
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -288,6 +288,13 @@ The accepter VPC named by `PeerVpcId` may belong to another account or region an
 in this store at all (a cross-account or "external" peer). Its `cidrBlock` is reported only when
 that VPC happens to exist locally; the request still succeeds either way, and no CIDR is fabricated.
 
+A connection's storage entry lives under whichever account's request created it, but lookups
+(`AcceptVpcPeeringConnection`, `DescribeVpcPeeringConnections`, `ModifyVpcPeeringConnectionOptions`,
+`DeleteVpcPeeringConnection`) resolve it across every account's partition, the same pattern used for
+RAM-shared IPAM resources. `AcceptVpcPeeringConnection` additionally enforces that the caller is the
+connection's accepter — reporting a connection it cannot see as absent, not as a permission error,
+matching how AWS itself responds.
+
 ### Network ACLs
 
 | Action | Description |
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -257,9 +257,36 @@ Floci seeds the following resources on first use in each region so Terraform, th
 | DeleteRouteTable | Deletes a route table from the local EC2 store. |
 | AssociateRouteTable | Associates a route table with a subnet. |
 | DisassociateRouteTable | Removes a route table association. |
-| CreateRoute | Adds a route to a route table. |
+| CreateRoute | Adds a route to a route table. Accepts `VpcPeeringConnectionId` as a target (alongside `GatewayId`/`NatGatewayId`/`EgressOnlyInternetGatewayId`) and reports it back on `DescribeRouteTables`. |
 | ReplaceRoute | Replaces the target of an existing route. |
 | DeleteRoute | Removes a route from a route table. |
+
+### VPC Peering Connections
+
+| Action | Description |
+|--------|-------------|
+| CreateVpcPeeringConnection | Creates a peering connection between the local VPC and a peer VPC, in status `pending-acceptance`. |
+| AcceptVpcPeeringConnection | Transitions a `pending-acceptance` connection to `active`. |
+| DescribeVpcPeeringConnections | Lists or returns stored peering connections. |
+| ModifyVpcPeeringConnectionOptions | Sets `allow_remote_vpc_dns_resolution` independently per side. |
+| DeleteVpcPeeringConnection | Deletes a peering connection from the local EC2 store. |
+
+Real AWS never auto-accepts a connection, same-account or not: every `CreateVpcPeeringConnection`
+starts `pending-acceptance` and stays there until an explicit `AcceptVpcPeeringConnection`. The
+`auto_accept` convenience on Terraform's `aws_vpc_peering_connection` and
+`aws_vpc_peering_connection_accepter` resources is implemented by the *provider*, which simply
+issues that second call itself — so this emulator does not special-case same-account peers.
+
+A connection is stored keyed by its id alone, not `region::id` like every other EC2 resource here.
+It is meaningfully addressable from both the requester's and the accepter's side, which can be a
+different region (`peer_region`/`accepter_region`); region-scoped storage would leave the accepter's
+`AcceptVpcPeeringConnection`/`DescribeVpcPeeringConnections` calls unable to find a connection
+created under the requester's region key. `RejectVpcPeeringConnection` is not implemented — no
+Gruntwork VPC-peering example exercises it (they use `auto_accept`, not manual rejection).
+
+The accepter VPC named by `PeerVpcId` may belong to another account or region and not be modelled
+in this store at all (a cross-account or "external" peer). Its `cidrBlock` is reported only when
+that VPC happens to exist locally; the request still succeeds either way, and no CIDR is fabricated.
 
 ### Network ACLs
 
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -825,8 +825,10 @@ public class CloudFormationResourceProvisioner {
         String gatewayId = resolveOptional(props, "GatewayId", engine);
         String natGatewayId = resolveOptional(props, "NatGatewayId", engine);
         String egressOnlyInternetGatewayId = resolveOptional(props, "EgressOnlyInternetGatewayId", engine);
+        String vpcPeeringConnectionId = resolveOptional(props, "VpcPeeringConnectionId", engine);
         ec2Service.createRoute(region, routeTableId, destinationCidr, destinationIpv6Cidr,
-                destinationPrefixListId, gatewayId, natGatewayId, egressOnlyInternetGatewayId);
+                destinationPrefixListId, gatewayId, natGatewayId, egressOnlyInternetGatewayId,
+                vpcPeeringConnectionId);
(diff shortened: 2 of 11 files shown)


allensanborn:
@greptile review

allensanborn:
@greptile review

allensanborn:
@greptile review

allensanborn:
@greptile review

allensanborn:
@greptile review

pgermosen (commented):
Substantial feature, and it's already been through an intense round of bot review before I looked at it: 9 separate P1 findings, several security-flagged — cross-account authorization bypasses, ownership forgery, a delete-resurrection race. I read the current code independently for every one of the 9 rather than trusting the resolved checkmarks, and all 9 hold up: cross-account acceptance now searches every account's storage partition and checks the real accepter identity, route-target exclusivity is enforced on both create and replace, Describe/accept/modify/delete all check regional visibility the same way Describe does, requester ownership uses the live caller instead of a fixed default account, peer ownership can no longer be forged by naming someone else's VPC and claiming yourself as the owner, and delete now takes the same lock accept/modify take so a concurrent mutation can't resurrect a row it just removed. Good, careful work across the board there.

One gap of my own, checked against the API reference: `ModifyVpcPeeringConnectionOptions` correctly stores and echoes the DNS-resolution setting in its own response, but `CreateVpcPeeringConnection`, `AcceptVpcPeeringConnection`, and `DescribeVpcPeeringConnections` never surface it — `peeringOptions` never appears under `requesterVpcInfo`/`accepterVpcInfo` in any of those three, even though it's a real, documented field there in every one of AWS's own response examples. Set DNS resolution via Modify, then Describe the same connection, and there's no trace of it in the response. Terraform's `aws_vpc_peering_connection_options` resource (and the inline accepter/requester blocks on `aws_vpc_peering_connection`) read this back via Describe on every plan, not by re-issuing Modify, so a config that sets it would see permanent drift. I checked the test suite specifically for this — every assertion on the DNS flag reads it back from Modify's own response, never from a subsequent Describe, so nothing currently catches it. The values are already computed and stored, so this looks like a small, mechanical addition to the existing response serializer rather than new work.

Two smaller things, not blocking: real AWS's CreateVpcPeeringConnection response shows the connection starting in `initiating-request` before moving to `pending-acceptance` moments later, while this goes straight to `pending-acceptance` — unlikely to matter given how Terraform/CLI polling treats both as non-terminal states, just noting the literal first response differs. And overlapping CIDR blocks between the two VPCs are supposed to land the connection in a `failed` status rather than succeeding; not modeled here, but none of the three target examples should hit it.

The design choice to key peering connections by id alone rather than region-scoped, like everything else in this file, is the right call given the cross-region accepter case — a region-scoped store would make the accepter's own Accept/Describe calls, issued against an entirely different region, unable to find the connection at all.

allensanborn:
Thanks — that's a careful read, and re-deriving the nine findings independently rather than trusting the checkmarks is appreciated.

**`peeringOptions` in Create/Accept/Describe** — fixed in 75e36d5, before this latest push. You were right about the shape of it: the values were already stored, so it was a serializer gap. `vpcPeeringConnectionVpcInfoXml` now emits `peeringOptions/allowDnsResolutionFromRemoteVpc` under both `requesterVpcInfo` and `accepterVpcInfo`, which covers all three responses since they share that one builder. And you were right about the test gap being the reason nothing caught it — the new assertions in `Ec2VpcPeeringConnectionIntegrationTest` set the flag via Modify and then read it back from a *subsequent Describe*, not from Modify's own response, so the drift case Terraform would hit is what's actually under test now.

**Peer ownership was still forgeable via `PeerRegion`** — this one came out of the bot's last round after your review, and it was real. The lookup I added to close the forgery resolved the peer VPC only within the requester-supplied region, so naming another account's real VPC while pointing `PeerRegion` at a region that VPC is absent from missed the lookup and fell straight back into the `PeerOwnerId`-is-trusted branch — reopening exactly the path the fix was meant to close. Pushed just now as cff376f5: a miss now falls back to trusting `PeerOwnerId` only when the id is absent from *every* region, which is also what real AWS does (it resolves the peer VPC globally and answers `InvalidVpcID.NotFound` on a region mismatch). A genuinely external peer VPC that this emulator has never seen behaves as before — that case is asserted in the new test alongside the forgery attempt. 690 EC2 tests green.

**`initiating-request` as the literal first response** — agreed that it differs, and I've left it. Modelling it faithfully means the stored status and the returned status disagree for exactly one response, which costs either a defensive copy of the connection or a status override in the query handler, and no client can observe the difference: both states are non-terminal and every poller treats them the same. Happy to add it if you'd rather the wire text match exactly — say the word and I'll take the handler-override route.

**Overlapping CIDRs landing in `failed`** — also left out, deliberately. It's real AWS behaviour but it's a new validation with its own state, not a serializer gap, and as you note none of the three target examples reach it. I'd rather file it than grow this PR; happy to do that if you agree.

On keying by id alone: that was the reasoning exactly — region-scoping the store breaks the cross-region accepter, who issues Accept and Describe against a region the connection isn't stored under. Region *visibility* is enforced separately, on read and on every mutation, which gets the leak-prevention without that failure mode.

pgermosen (approved):
@allensanborn substantial work, and it held up under a lot of scrutiny — nine P1 findings, several security-flagged, all genuinely closed, plus the two follow-ups since. I focused my own read on what changed most recently rather than re-covering the nine pgermosen already re-verified independently.

Traced the `PeerRegion` forgery fix by hand: naming a real victim VPC in the wrong region now resolves it globally first and rejects the mismatch before ever reaching the `PeerOwnerId`-trusting fallback, naming it in the correct region uses the VPC's actual owning account regardless of what's claimed, and a genuinely external peer VPC still falls through to trusting the caller's claim, which is correct for that case. All three hold.

Also pulled the actual AWS reference myself rather than take the peeringOptions fix on description alone — confirmed `vpcPeeringConnectionVpcInfoXml` is one shared builder feeding all three response paths (Create, Accept, every item in Describe), not three separate ones that could drift. One thing I checked and want to flag as resolved rather than silently drop: the cached `AcceptVpcPeeringConnection` example shows two extra ClassicLink fields alongside `allowDnsResolutionFromRemoteVpc` in `peeringOptions`. I fetched the current `ModifyVpcPeeringConnectionOptions` reference live to check, and it only documents the DNS field — ClassicLink is long deprecated, so that's a stale legacy example, not a gap here.

`initiating-request` and the overlapping-CIDR `failed` status are both genuinely documented AWS behavior (checked directly), so those two are real, deliberate scope calls rather than things I'd push on — your call whether you want them in this PR or filed separately.

Approving.

PR #2672, opened 2026-08-28
fix(lambda): replace the extracted package instead of merging into it

### LINKED ISSUE #2647: [BUG] lambda: deployment package extraction overlays the previous package instead of replacing it

avison9:
## ⚠️ Stacked on #2649 — read before reviewing the diff

Closes #2647.

This branch is **stacked on #2649** (issue #2593). Both change `ZipExtractor.extractTo`, so they
would conflict if opened independently off `main`, and they are separate concerns — *how the
archive is read* vs. *what happens to the target directory* — which `CONTRIBUTING.md` asks be kept
to one per PR.

Until #2649 merges, the diff here shows **its commits too**. The commit that belongs to this PR is
the last one:

```
fix(lambda): replace the extracted package instead of merging into it   <- this PR
--- everything below is #2649 ---
fix(lambda): read the package with java.util.zip, not Commons Compress
fix(lambda): read the deployment package without staging a copy on disk
fix(lambda): stage the deployment package in the code store, not /tmp
fix(lambda): read deployment packages via the zip central directory
```

Merge #2649 first and this collapses to a two-file change. Happy to rebase and repush on request,
or to hold this as a draft until #2649 lands — whichever suits your review flow.

---

## Summary

A Lambda deployment package is the complete contents of the function — real AWS replaces the
package wholesale. Floci extracted each package **into** the existing code directory without
clearing it first, so every deploy was layered on top of the last. Files removed from the project
survived indefinitely and stayed loadable: a handler pointing at a deleted module kept resolving
and running stale code that was no longer in the deployment package.

### Before / after

v1 contains `index.js` and `legacy/old.js`; v2 contains only `index.js`:

```bash
aws $E lambda create-function --function-name overlay-repro --runtime nodejs20.x \
  --role arn:aws:iam::000000000000:role/r --handler legacy/old.handler --zip-file fileb://v1.zip
aws $E lambda update-function-code --function-name overlay-repro --zip-file fileb://v2.zip
aws $E lambda invoke --function-name overlay-repro out.json && cat out.json
```

```
# BEFORE
create-function        "CodeSize": 313
update-function-code   "CodeSize": 161      <- Floci records that the package shrank
invoke                 {"body":"STALE"}     <- v1 code, deleted from the v2 package, still runs

# AFTER
invoke                 Runtime.ImportModuleError — Cannot find module 'legacy/old'
```

The function metadata was updated correctly (`CodeSize` reflects v2) while the extracted directory
still held v1's files, so the two disagreed.

## Root cause

`extractTo` created the target directory and wrote entries into it, with no reset step anywhere on
the path — `CodeStore.delete` runs only when a function is deleted, `codeStore.exists()` is never
called in `src/main` at all, and every deploy of a given function resolves to the same
`codeStore.getCodePath(name)`. **Extraction was implemented as a merge where the domain requires a
replace.** Writing an archive over a directory only ever adds and overwrites — it cannot express
deletion — so the extracted tree was the union of every package ever deployed.

The same missing reset made extraction non-atomic: a failure part-way through left the target
holding a mixture of the previous package and part of the new one.

Extraction now goes into a staging directory (`<target>.floci-staging-<uuid>`, a sibling) which is
swapped in only after the whole archive has been read:

- **Success** — the old target is deleted recursively and staging is moved into place with
  `ATOMIC_MOVE`, falling back to a plain move where the filesystem cannot rename atomically.
- **Failure** — staging is deleted in `finally`; the previous deployment is untouched.

Entry containment is validated against the staging directory; the traversal and backslash guards
and the method signature are otherwise unchanged.

Alternatives: **clearing the target in place** fixes stale files but makes atomicity *worse* — the
function has no code for the whole extraction, and a failure leaves it empty rather than merely
mixed. **Having `LambdaService` delete first** has the same problem and spreads "a package replaces
the previous one" across two classes, leaving layers wrong. **De-duplicating at invoke time** puts
deploy-time state on the hot path and still leaves a deleted credential on disk.

**Why replacing the directory is safe:** `codePath` is not bind-mounted into Lambda containers.
`ContainerLauncher` tar-copies it into `/var/task` at launch (`copyDirToContainer`); the only
`withBind` for task code is the hot-reload path, which uses `getHotReloadHostPath()` — a different
directory that never goes through `extractTo`. So replacing the directory inode cannot pull the
ground out from under a running container. I checked this before choosing the design.

## AWS Compatibility

Real AWS treats a deployment package as the complete, immutable contents of the function; a file
absent from the new package is gone. Floci now matches. Beyond correctness: a file deleted from a
package deliberately (a checked-in credential, a debug-only endpoint) no longer stays on disk and
invocable, and code directories no longer grow monotonically across redeploys.

## How it was tested

| Test | Covers | Fails without the fix? |
|---|---|---|
| `replacesThePreviousExtractionInsteadOfOverlayingIt` | Deploys v1, adds files, deploys v2; asserts the removed file *and* a removed directory are gone | **Yes** — `expected: <false> but was: <true>` |
| `leavesThePreviousExtractionIntactWhenExtractionFailsPartWay` | Atomicity, using an archive that opens cleanly and fails only once its entry has been streamed out | **Yes** — `expected: <v1> but was: <qez5"$-corrupted-in-transit>` |
| `leavesNoStagingDirectoryBehind` | No orphan staging directory survives either outcome | No — a guard on the new cleanup path, not a regression test |

An earlier draft of the atomicity test fed in plain garbage bytes, but `ZipFile` rejects that at
*open* time before anything is written, so it passed against unfixed code and proved nothing.
Using a mid-extraction failure is what makes it a real regression test.

Full suite: see the checklist below. Java 25 / Maven 3.9.16, matching `ci.yml`.

## Risk / rollback

Behaviour change, intended: anything relying on files persisting across deploys (nothing
documented does) would notice. Self-healing on upgrade — directories already polluted by earlier
deploys are cleaned on the next deploy, since the target is replaced wholesale. Peak disk during a
deploy is now the old package plus the new one, bounded by the AWS package limit and strictly
smaller than the unbounded growth the merge behaviour caused. Rollback is reverting the single file.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/zip/ZipExtractor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/zip/ZipExtractor.java
@@ -6,15 +6,19 @@ import org.jboss.logging.Logger;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.Comparator;
+import java.util.UUID;
 import java.util.zip.CRC32;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipException;
 import java.util.zip.ZipFile;
 
 /**
- * Extracts ZIP bytes to a target directory.
+ * Extracts ZIP bytes to a target directory, replacing whatever was there before.
  * Guards against path traversal attacks by validating entry names.
  */
 @ApplicationScoped
@@ -27,7 +31,17 @@ public class ZipExtractor {
     public void extractTo(byte[] zipBytes, Path targetDir) throws IOException {
         // Resolve to absolute path so that normalize() on entry paths stays comparable
         Path absTarget = targetDir.toAbsolutePath().normalize();
-        Files.createDirectories(absTarget);
+        Files.createDirectories(absTarget.getParent());
+
+        // Extract into a staging directory and swap it in, rather than writing over the target
+        // in place. A deployment package is the complete contents of the function, so a file
+        // dropped between two deploys has to disappear — writing in place only ever adds and
+        // overwrites, leaving removed files behind to stay loadable (issue #2647). Staging also
+        // makes extraction all-or-nothing: a failure part-way through leaves the previous
+        // deployment intact instead of a half-replaced mixture of the two.
+        Path staging = absTarget.resolveSibling(
+                absTarget.getFileName() + ".floci-staging-" + UUID.randomUUID());
+        Files.createDirectories(staging);
 
         // Read the archive through its central directory (ZipFile) rather than sequentially
         // over the local file headers (ZipInputStream). A streaming packager cannot seek back
@@ -71,8 +85,8 @@ public class ZipExtractor {
                         continue;
                     }
 
-                    Path targetPath = absTarget.resolve(entryName).normalize();
-                    if (!targetPath.startsWith(absTarget)) {
+                    Path targetPath = staging.resolve(entryName).normalize();
+                    if (!targetPath.startsWith(staging)) {
                         LOG.warnv("Skipping out-of-bounds ZIP entry: {0}", entryName);
                         continue;
                     }
@@ -85,11 +99,37 @@ public class ZipExtractor {
                     }
                 }
             }
+            // Extraction succeeded in full: swap the new tree in for the old one.
+            deleteRecursively(absTarget);
+            moveIntoPlace(staging, absTarget);
         } finally {
             Files.deleteIfExists(staged);
+            // A no-op once the move succeeded; on failure this discards the partial staging
+            // tree and leaves the previous extraction untouched.
+            deleteRecursively(staging);
         }
 
         LOG.debugv("Extracted ZIP to: {0}", absTarget);
+    }
+
+    private static void moveIntoPlace(Path staging, Path absTarget) throws IOException {
+        try {
+            Files.move(staging, absTarget, StandardCopyOption.ATOMIC_MOVE);
+        } catch (AtomicMoveNotSupportedException e) {
+            // Some filesystems (and bind mounts inside containers) cannot rename atomically.
+            Files.move(staging, absTarget);
+        }
+    }
+
+    private static void deleteRecursively(Path root) throws IOException {
+        if (!Files.exists(root)) {
+            return;
+        }
+        try (var paths = Files.walk(root)) {
+            for (Path p : paths.sorted(Comparator.reverseOrder()).toList()) {
+                Files.deleteIfExists(p);
+            }
+        }
     }
 
     /**
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/zip/ZipExtractor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/zip/ZipExtractor.java
@@ -99,25 +99,64 @@ public class ZipExtractor {
                     }
                 }
             }
-            // Extraction succeeded in full: swap the new tree in for the old one.
-            deleteRecursively(absTarget);
-            moveIntoPlace(staging, absTarget);
+            // Extraction succeeded in full: install the new tree.
+            install(staging, absTarget);
         } finally {
             Files.deleteIfExists(staged);
-            // A no-op once the move succeeded; on failure this discards the partial staging
-            // tree and leaves the previous extraction untouched.
+            // A no-op once install() moved it; on failure this discards the partial staging tree
+            // and leaves the previous extraction untouched.
             deleteRecursively(staging);
         }
 
         LOG.debugv("Extracted ZIP to: {0}", absTarget);
     }
 
-    private static void moveIntoPlace(Path staging, Path absTarget) throws IOException {
+    /**
+     * Installs the freshly extracted tree at {@code absTarget}, keeping the package already there
+     * until the new one is in place.
+     *
+     * <p>The old tree is renamed aside and deleted only after the new one lands, rather than
+     * deleted up front. Deleting first means a failure in the move leaves the function with no
+     * code at all and nothing to restore, turning a failed update into a destroyed deployment.
+     * Every failure path here ends with one complete package on disk.
+     *
+     * <p>Both steps are renames within one directory, so the interval in which the target is
+     * absent is two metadata operations rather than a recursive delete followed by a move. It is
+     * not zero: POSIX cannot atomically exchange two directories, so a launch landing exactly
+     * between the renames can still observe a missing code path. Closing that window entirely
+     * needs a lock shared with the launcher, which is wider than this change.
+     *
+     * <p>Package-private so the failure path can be exercised directly; it is not reachable
+     * otherwise, because a move between siblings only fails under conditions a test cannot
+     * arrange through {@link #extractTo}.
+     */
+    static void install(Path staging, Path absTarget) throws IOException {
+        Path previous = null;
+        if (Files.exists(absTarget)) {
+            previous = absTarget.resolveSibling(
+                    absTarget.getFileName() + ".floci-previous-" + UUID.randomUUID());
+            moveIntoPlace(absTarget, previous);
+        }
         try {
-            Files.move(staging, absTarget, StandardCopyOption.ATOMIC_MOVE);
+            moveIntoPlace(staging, absTarget);
+        } catch (IOException e) {
+            if (previous != null) {
+                // Put the working package back before giving up.
+                moveIntoPlace(previous, absTarget);
+            }
+            throw e;
+        }
+        if (previous != null) {
+            deleteRecursively(previous);
+        }
+    }
+
+    private static void moveIntoPlace(Path source, Path destination) throws IOException {
+        try {
+            Files.move(source, destination, StandardCopyOption.ATOMIC_MOVE);
         } catch (AtomicMoveNotSupportedException e) {
             // Some filesystems (and bind mounts inside containers) cannot rename atomically.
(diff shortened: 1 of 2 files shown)


avison9:
Updated. Two changes since this was opened.

## Rebased onto main

#2649 has merged, so this is no longer stacked and the conflict is gone. The branch is now two
commits on top of current main, and the diff is down to the two files this PR is actually about:
`ZipExtractor.java` and `ZipExtractorTest.java`. The commits carried over from #2649 are gone,
because that work reached main through the squash merge.

## A failed install no longer destroys the working package

The install originally deleted the live tree and then moved the new one into place. If the move
failed, the function was left with no code at all and nothing to restore, so a failed update
destroyed a working deployment.

It now renames the old tree aside, moves the new one in, and deletes the old only once the new one
has landed. Every failure path ends with one complete package on disk.

```java
Path previous = null;
if (Files.exists(absTarget)) {
    previous = absTarget.resolveSibling(absTarget.getFileName() + ".floci-previous-" + UUID.randomUUID());
    moveIntoPlace(absTarget, previous);
}
try {
    moveIntoPlace(staging, absTarget);
} catch (IOException e) {
    if (previous != null) {
        moveIntoPlace(previous, absTarget);   // put the working package back
    }
    throw e;
}
if (previous != null) {
    deleteRecursively(previous);
}
```

The window in which a concurrent launch could observe a missing directory is **narrowed, not
closed**, and the javadoc says so rather than implying atomicity. It is now two renames within one
directory instead of a recursive delete followed by a move. POSIX cannot atomically exchange two
directories, so a launch landing exactly between the renames can still see a missing code path.
Closing that properly needs a lock shared with the launcher, which is wider than this change and
belongs in its own issue. Happy to file it if that is wanted.

`install(staging, absTarget)` is package private so the failure path can be driven directly. It is
unreachable through `extractTo`, because a rename between siblings does not fail under any condition
a test can arrange there. Reverting only the ordering and rerunning reproduces the destroyed
deployment, so the new test is a real regression test rather than a passing assertion:

```
aFailedInstallPutsThePreviousPackageBack:410
  a failed install must put the previous package back, not leave the function empty
```

## One note on the test file name

`ZipExtractorTest` does not match the `*ServiceTest.java` convention in AGENTS.md. I have left it
alone deliberately: the file predates this branch, `ZipExtractor` is a small utility used by
`LambdaService` and `LambdaLayerService` rather than an `@ApplicationScoped` service, and renaming it
here would put an unrelated change into a fix PR. Surefire picks it up either way, so nothing is
uncovered. If maintainers would rather it were renamed, I would prefer to do that as its own `test:`
change across whichever files are inconsistent, instead of renaming one file because this PR touched
it.

## Verification

`ZipExtractorTest` is at 13 tests, covering the streamed descriptor cases, checksum failure,
replacement semantics, staging cleanup, and now both install outcomes. Full suite figures are in the
checks below.

avison9:
Pushed `0697c55f`, covering two further failure paths in `install()` where the reported outcome did
not match what had actually happened on disk.

## A deployment that already installed is no longer reported as failed

Deleting the superseded package threw out of `install()`. Once the new tree is in place the
deployment has happened, so a filesystem error at that point failed the whole operation while the
new code was already serving, and the caller skipped its metadata and warm pool updates for live
code.

That deletion is now best effort. It logs and leaves the tree behind. Leaving a directory costs disk
and is recoverable; reporting a successful deployment as a failure is not.

```java
try {
    deleteRecursively(previous);
} catch (IOException e) {
    LOG.warnv("Installed {0} but could not remove the superseded package at {1}: {2}",
            absTarget, previous, e.getMessage());
}
```

## The rollback no longer loses the package silently

Restoring the previous package after a failed move was itself unguarded. If that restore also threw,
its exception replaced the original failure and the working package was left under a generated
`.floci-previous-<uuid>` name that nothing would look for.

The restore is now guarded. The original failure stays the exception thrown, the restore failure is
attached to it with `addSuppressed`, and the stranded path is logged so it can be moved back:

```java
try {
    moveIntoPlace(previous, absTarget);
} catch (IOException restoreFailed) {
    e.addSuppressed(restoreFailed);
    LOG.errorv("Could not restore the previous deployment at {0}. The working package is "
            + "stranded at {1}; move it back to recover.", absTarget, previous);
}
```

## Test coverage, including what is not covered

The cleanup path has a regression test. Deletion is blocked by removing write permission on a
subdirectory of the old package, and reverting only the best effort handling reproduces the failure:

```
aFailureToDropTheSupersededPackageDoesNotFailTheDeployment:456
  » AccessDenied /tmp/junit-.../fn.floci-previous-0af907bd.../locked/held.txt
```

**The restore path has no test, and I would rather say that than imply coverage.** Reaching it needs
the install rename to fail while the restore rename also fails. Both are renames within one
directory: if the parent is writable both succeed, and if it is not, the first rename fails before
install is reached. It is not reachable without fault injection. The guard is still worth having,
because it turns a silent loss into a logged, recoverable state with the original cause preserved.

`ZipExtractorTest` is at 14 tests. Full suite on this branch against current main:
`Tests run: 13608, Failures: 0, Errors: 0, Skipped: 9`.

pgermosen (approved):
Three real atomicity/rollback findings, all genuinely fixed — I traced the full state machine by hand rather than trusting the resolved checkmarks, since this exact class of atomic-swap-with-rollback logic is easy to get subtly wrong. The old package is renamed aside before the new one installs, which is a cheap, reversible metadata op rather than a destructive delete; if the new install fails, the old one is restored before re-throwing; if cleanup of the superseded tree fails, that's logged as a warning rather than reported as a failed deployment, since the new code is already live at that point; and in the genuinely unrecoverable case where both renames fail, the original exception is thrown with the restore failure attached and an error log stating exactly where the still-working package is stranded, so it's recoverable by hand rather than silently lost. The one residual gap — a launch landing in the narrow window between the two renames — is explicitly disclosed in the javadoc as needing a lock shared with the launcher, correctly scoped as wider than this change rather than hidden.

The regression tests back this concretely rather than just asserting the happy path: one deploys v1, adds files and a directory v2's package doesn't include, and confirms they're gone after the v2 deploy — the actual bug this closes. Another uses an archive that opens cleanly and fails only after streaming has started, specifically because that's the exact point where the old in-place-write behavior would already have clobbered the working deploy, and confirms v1 survives a failed v2 attempt.

The one open thread (test file naming convention) doesn't apply here — ZipExtractor isn't a *Service class, and the file predates this PR.

PR #2673, opened 2026-08-28
feat(autoscaling): P0 operations and lifecycle-hook delete-by-name

sam-fakhreddine:
## Summary

Adds two Auto Scaling P0 operations that were missing from the emulator's Query API surface — `SetInstanceProtection` and `SetInstanceHealth` — and closes two wire-fidelity gaps found during review of the existing lifecycle-hook operations.

- **`SetInstanceProtection`**: toggles `ProtectedFromScaleIn` on the named instances in a group; rejects an empty instance-ID list and any instance not in the group with `ValidationError` (LZA-adjacent hand-verification: matches the request/response shape AWS SDK v2's `AutoScalingClient` expects for this call).
- **`SetInstanceHealth`**: sets an instance's health status; already validated `HealthStatus ∈ {Healthy, Unhealthy}` and instance existence — botocore-derived from `autoscaling/2011-01-01/service-2.json`'s `SetInstanceHealthQuery` shape.

Provenance: both operations and their handler wiring were written and unit/integration-tested directly against botocore's Auto Scaling model; no LZA pipeline run has exercised Auto Scaling specifically, so treat this as botocore-derived rather than LZA-verified.

## Wire-fidelity details

Ran the wire-fidelity extractor (`floci-review`) against this branch's `services/autoscaling` package using `autoscaling/2011-01-01/service-2.json`, then judged the 88 emitted packets (first fidelity pass for this service — 0 prior entries in `corpus.jsonl`):

| Verdict | Count |
|---|---|
| VIOLATION | 22 |
| ENFORCED | 2 |
| NOT_APPLICABLE | 57 (mostly: constraint applies to a describe/filter param, or the cited source read didn't match the operation) |

Of the 22 `VIOLATION` verdicts, three were investigated and confirmed **false positives** on re-verification against the current source (the extractor's excerpted `source_reads` didn't include the actual validation, which lives a few lines outside the snippet):

- `SetInstanceHealth.HealthStatus` / `.InstanceId` — both already validated (`HealthStatus` must be `Healthy`/`Unhealthy`; `InstanceId` must exist in some group), `AutoScalingService.java` `setInstanceHealth`.
- `StartInstanceRefresh.Strategy` — already validated against the modeled `RefreshStrategy` enum (`Rolling`, `ReplaceRootVolume`) in `normalizeRefreshStrategy`; the flagged excerpt only showed the raw request DTO setter, not the validation that runs before persistence.
- `PutScalingPolicy.AdjustmentType` (flagged `high`) — disputed: botocore models this member as `XmlStringMaxLen255` (free string), not an enum. The valid values (`ChangeInCapacity`, `ExactCapacity`, `PercentChangeInCapacity`) are documentation prose only, not a modeled `enum` constraint, so there is no wire-fidelity gap to fix here.

Two real gaps were confirmed and fixed in this PR (commit `b31e90325`):

- **`LifecycleHookName`** (`PutLifecycleHook`, `DeleteLifecycleHook`, `CompleteLifecycleAction`) — modeled as `AsciiStringMaxLen255` with pattern `[A-Za-z0-9\-_/]+` and max length 255; previously accepted any string. Now rejected with `ValidationError` outside the modeled pattern/length.
- **`LifecycleActionToken`** (`CompleteLifecycleAction`) — modeled as a fixed 36-character string (`{'min': 36, 'max': 36}`); previously accepted any length. Now rejected with `ValidationError` unless exactly 36 characters.

`corpus.jsonl`'s growth (81 new judged packets — 7 of the 88 packets were duplicate keys already covered) is committed separately in `floci-review`.

## Deliberate scope notes

- The remaining 17 `VIOLATION` findings (mostly on `AutoScalingGroupName`, `LaunchConfigurationName`, `PolicyName`, instance IDs, `HealthCheckType`, `VPCZoneIdentifier`, `TerminationPolicies`, `SecurityGroups`, ARNs/name lists, and free-text fields like `NotificationMetadata`) are all `XmlStringMaxLen*`-family shapes: length caps paired with a near-unrestricted printable-Unicode pattern that matches virtually any input. Enforcing these would be pure length-capping with no meaningful behavioral difference from AWS for realistic inputs, and is out of scope for this PR — left unenforced consistent with how most other Query-API handlers in this codebase currently treat this class of member.
- `RecordLifecycleActionHeartbeat`'s two flagged members (`LifecycleHookName`, `LifecycleActionToken`) are not fixed here: the handler is currently a stub that ignores all request parameters and always returns success (Phase 2 reconciler territory, matching `completeLifecycleAction`'s own "stored-only" comment) — there is no live code path to validate against yet.
- `CompleteLifecycleAction` and `RecordLifecycleActionHeartbeat` remain "stored-only" no-ops beyond the new input validation — they don't yet drive instance lifecycle transitions. That's pre-existing behavior, not something introduced by this PR.

## Type of change

## AWS Compatibility

- [x] I have verified this change against real AWS behavior or documentation (botocore `autoscaling/2011-01-01/service-2.json`)
- [x] This change maintains compatibility with the AWS API contract

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works


src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
```diff
@@ -371,6 +371,47 @@ public List<AsgInstance> describeAutoScalingInstances(String region, List<String
...
+                        .anyMatch(instance -> instanceId != null && instanceId.equals(instance.getInstanceId())))
+                .findFirst()
+                .orElseThrow(() -> new AwsException("ValidationError",
+                        "Instance '" + instanceId + "' not found in any Auto Scaling group.", 400));
+        asg.getInstances().stream()
+                .filter(instance -> instanceId.equals(instance.getInstanceId()))
+                .findFirst()
+                .orElseThrow(() -> new AwsException("ValidationError",
+                        "Instance '" + instanceId + "' not found in Auto Scaling group.", 400))
+                .setHealthStatus(healthStatus);
```
utafrali: The second stream lookup and its `orElseThrow` are unreachable: the ASG was already selected because it contains an instance matching `instanceId`. This can be collapsed into a single traversal. For example: `asg.getInstances().stream().filter(i -> instanceId.equals(i.getInstanceId())).findFirst().ifPresent(i -> i.setHealthStatus(healthStatus));` after the first `orElseThrow`. Keeping the duplicate lookup makes the invariant harder to read and hides that the second `orElseThrow` message can never fire.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingQueryHandler.java
```diff
@@ -472,6 +474,29 @@ private Response handleDescribeAutoScalingInstances(MultivaluedMap<String, Strin
...
+
+    private Response handleSetInstanceHealth(MultivaluedMap<String, String> p, String region) {
+        String instanceId = p.getFirst("InstanceId");
+        String healthStatus = p.getFirst("HealthStatus");
+        service.setInstanceHealth(region, instanceId, healthStatus);
+        return ok(new XmlBuilder()
+                .start("SetInstanceHealthResponse", NS)
+                .raw(AwsQueryResponse.responseMetadata())
+                .end("SetInstanceHealthResponse")
+                .build());
```
utafrali, line 500: `SetInstanceHealth` accepts a `ShouldRespectGracePeriod` boolean per the 2011-01-01 model, but the handler ignores it. For wire fidelity, parse it (`Boolean.parseBoolean(p.getFirst("ShouldRespectGracePeriod"))`) and pass it through to the service, even if the emulator does not act on the grace period yet, so callers using SDK v2 do not silently lose the flag.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
```diff
@@ -371,6 +371,47 @@ public List<AsgInstance> describeAutoScalingInstances(String region, List<String
...
+        }
+        Set<String> requested = new HashSet<>(instanceIds);
+        List<AsgInstance> matches = asg.getInstances().stream()
+                .filter(instance -> requested.contains(instance.getInstanceId()))
+                .toList();
+        if (matches.size() != requested.size()) {
+            Set<String> found = matches.stream().map(AsgInstance::getInstanceId).collect(Collectors.toSet());
+            String missing = requested.stream().filter(id -> !found.contains(id)).findFirst().orElse("unknown");
+            throw new AwsException("ValidationError",
+                    "Instance '" + missing + "' is not part of Auto Scaling group '" + groupName + "'.", 400);
```
utafrali: When multiple instance IDs are unknown, only the first missing ID is surfaced (`.findFirst().orElse("unknown")`). AWS reports all missing IDs in a single validation message. Consider joining the full `requested \\ found` set into the message so callers debugging batch calls see every offender in one round trip.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
```diff
@@ -371,6 +371,47 @@ public List<AsgInstance> describeAutoScalingInstances(String region, List<String
         return all;
     }
 
+    public void setInstanceProtection(String region, String groupName, List<String> instanceIds,
+                                      boolean protectedFromScaleIn) {
+        AutoScalingGroup asg = requireGroup(region, groupName);
+        if (instanceIds == null || instanceIds.isEmpty()) {
+            throw new AwsException("ValidationError", "At least one instance ID is required.", 400);
```
utafrali, line 378: `requireGroup` runs before the empty-`instanceIds` check, so a request that is doubly bad (unknown group + empty list) surfaces `Group not found` and hides the missing-parameter error. AWS validates required members first. Swap the order so the empty-list `ValidationError` fires regardless of group state.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
```diff
@@ -574,6 +616,7 @@ public void putLifecycleHook(String region, String asgName, String hookName,
     }
 
     public void deleteLifecycleHook(String region, String asgName, String hookName) {
+        validateLifecycleHookName(hookName);
         hooks.remove(hookKey(region, asgName, hookName));
     }
```
utafrali, line 631: `deleteLifecycleHookByName` (used by the CFN delete path) still bypasses `validateLifecycleHookName`. If it is called with a name that could not have been created through the normal path, we silently no-op. Either invoke `validateLifecycleHookName(hookName)` here for parity with `deleteLifecycleHook`, or add a code comment stating the CFN path deliberately skips validation because the name came from a stored physical id.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
```diff
@@ -616,6 +659,30 @@ public void completeLifecycleAction(String region, String asgName, String hookNa
                                          String instanceId, String actionResult, String token) {
         // Stored-only — Phase 2 reconciler observes this via the instance lifecycle state
         requireGroup(region, asgName);
+        validateLifecycleHookName(hookName);
+        validateLifecycleActionToken(token);
+    }
+
+    /** LifecycleHookName pattern/length from the Auto Scaling model (autoscaling/2011-01-01/service-2.json). */
+    private static final java.util.regex.Pattern LIFECYCLE_HOOK_NAME_PATTERN =
+            java.util.regex.Pattern.compile("^[A-Za-z0-9\\-_/]+$");
```
utafrali: `java.util.regex.Pattern` is referenced with a fully-qualified name even though the file already uses `java.util.*`. Add `import java.util.regex.Pattern;` and drop the qualifier for consistency with the rest of the file.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingQueryHandler.java
```diff
@@ -50,6 +50,8 @@ public Response handle(String action, MultivaluedMap<String, String> p, String r
                 case "DeleteTags"                   -> handleDeleteTags(p, region);
                 // Instances
                 case "DescribeAutoScalingInstances" -> handleDescribeAutoScalingInstances(p, region);
+                case "SetInstanceProtection"         -> handleSetInstanceProtection(p, region);
+                case "SetInstanceHealth"              -> handleSetInstanceHealth(p, region);
```
utafrali: The two new `case` entries have extra spaces before `->` (13 and 14 spaces), breaking the column alignment used by the surrounding block (11 spaces). Trim the padding so `SetInstanceProtection` / `SetInstanceHealth` line up with `DescribeAutoScalingInstances` and `AttachInstances`.

utafrali (approved):
Solid additions that close real gaps in the Auto Scaling emulator with good test coverage. Main issues are a redundant lookup in `setInstanceHealth`, a missing `ShouldRespectGracePeriod` parameter, and a few consistency nits around validation ordering and error messages.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
sam-fakhreddine: Fixed in 63dd67cc6 — collapsed the second stream lookup in setInstanceHealth into a single `ifPresent` after the group-membership orElseThrow, exactly as suggested. RED: adding the boolean shouldRespectGracePeriod param to the signature (needed for #2/#11 below) broke the old two-lookup version at compile time before I even got to run it, which forced the collapse. AutoScalingServiceTest/AutoScalingQueryHandlerTest green (106 tests across the affected suites).

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingQueryHandler.java
sam-fakhreddine: Fixed in 63dd67cc6 together with greptile's identical finding on ShouldRespectGracePeriod (health-status thread). The handler now parses ShouldRespectGracePeriod (default true per the 2011-01-01 model when absent) and passes it through to AutoScalingService.setInstanceHealth. I deliberately did NOT implement actual grace-period gating: floci has no instance launch-time tracking against AutoScalingGroup.getHealthCheckGracePeriod() today, and adding that is a real feature, not a targeted fix for this PR. This matches your own carve-out ("even if the emulator does not act on the grace period yet"). Documented the limitation in a javadoc comment on the method. New test setInstanceHealthParsesShouldRespectGracePeriodWithoutError / setInstanceHealthAcceptsShouldRespectGracePeriodFlag confirm the flag is accepted end-to-end. Ledgering full grace-period enforcement as a follow-up rather than opening a new issue file, since it's a known, disclosed gap, not a bug.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
sam-fakhreddine: Fixed in 63dd67cc6 — the missing-instance-IDs check now joins the full requested-minus-found set into the message instead of findFirst(). This also folds in greptile's overlapping finding (set-containment vs count-based matching) since both live in the same few lines. RED: setInstanceProtectionReportsAllMissingInstanceIds asserts both 'i-missing-1' and 'i-missing-2' appear in the thrown message. AutoScalingServiceTest green.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
sam-fakhreddine: Fixed in 63dd67cc6 — swapped the order so the empty-InstanceIds ValidationError is checked before requireGroup, matching AWS's required-member-first validation order. RED: setInstanceProtectionValidatesEmptyInstanceIdsBeforeGroupLookup calls setInstanceProtection with an unknown group AND an empty instance list, and asserts the message is the missing-parameter error, not 'Group not found'. Green after the swap.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
sam-fakhreddine: Fixed in 63dd67cc6 — deleteLifecycleHookByName now calls validateLifecycleHookName(hookName) after the existing null-guard, for parity with deleteLifecycleHook. Kept the null-check as an idempotent-delete no-op (the CFN provisioner already null/blank-checks before calling this), but a non-null malformed name (e.g. tampered/hand-authored physical id) now fails loudly instead of silently no-op'ing. RED: deleteLifecycleHookByNameRejectsNameOutsideModeledPattern. Also added deleteLifecycleHookByNameIsIdempotentOnNullPhysicalId to pin the intentional null-is-a-no-op behavior so it doesn't regress.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
sam-fakhreddine: Fixed in 63dd67cc6 — added `import java.util.regex.Pattern;` and dropped the fully-qualified references on both the field declaration and Pattern.compile call.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingQueryHandler.java
sam-fakhreddine: Fixed in 63dd67cc6 — trimmed the padding on the SetInstanceProtection/SetInstanceHealth case arrows so they line up at the same column as the rest of the block (verified all arrows in that switch now sit at the same column, except the one case whose label is itself longer than the alignment column, same as before).

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingQueryHandler.java
```diff
@@ -472,6 +474,29 @@ private Response handleDescribeAutoScalingInstances(MultivaluedMap<String, Strin
         return ok(xml.build());
     }
 
+    private Response handleSetInstanceProtection(MultivaluedMap<String, String> p, String region) {
+        String groupName = p.getFirst("AutoScalingGroupName");
+        List<String> instanceIds = memberList(p, "InstanceIds");
+        boolean protectedFromScaleIn = Boolean.parseBoolean(p.getFirst("ProtectedFromScaleIn"));
```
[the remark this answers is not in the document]
sam-fakhreddine: Fixed in 63dd67cc6 — ProtectedFromScaleIn is a required member on SetInstanceProtection (confirmed against botocore autoscaling/2011-01-01/service-2.json). Added requiredBoolParam() in the query handler: missing/blank now throws ValidationError instead of Boolean.parseBoolean silently coercing to false, and a non-true/false value (e.g. 'yes') is also rejected rather than treated as false. RED: setInstanceProtectionRejectsMissingProtectedFromScaleIn / setInstanceProtectionRejectsMalformedProtectedFromScaleIn both assert the instance's ProtectedFromScaleIn is untouched after the rejected request.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
```diff
@@ -371,6 +371,47 @@ public List<AsgInstance> describeAutoScalingInstances(String region, List<String
...
+        Set<String> requested = new HashSet<>(instanceIds);
+        List<AsgInstance> matches = asg.getInstances().stream()
+                .filter(instance -> requested.contains(instance.getInstanceId()))
+                .toList();
+        if (matches.size() != requested.size()) {
+            Set<String> found = matches.stream().map(AsgInstance::getInstanceId).collect(Collectors.toSet());
+            String missing = requested.stream().filter(id -> !found.contains(id)).findFirst().orElse("unknown");
+            throw new AwsException("ValidationError",
+                    "Instance '" + missing + "' is not part of Auto Scaling group '" + groupName + "'.", 400);
+        }
```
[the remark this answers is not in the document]
sam-fakhreddine: Fixed in 63dd67cc6 using your suggested diff shape — setInstanceProtection now builds a Set<String> of found instance IDs from the group and checks found.containsAll(requested) instead of comparing collection sizes, so duplicate stored instance IDs no longer produce a false match/mismatch. RED: setInstanceProtectionUsesSetContainmentAgainstDuplicateStoredInstanceIds seeds the group with two AsgInstance entries sharing the same ID and asserts the single requested ID is still accepted and protected. Folded into the same commit as utafrali's overlapping 'report all missing IDs' ask since both touch this exact block.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingQueryHandler.java
```diff
@@ -472,6 +474,29 @@ private Response handleDescribeAutoScalingInstances(MultivaluedMap<String, Strin
...
+                .start("SetInstanceProtectionResponse", NS)
+                .raw(AwsQueryResponse.responseMetadata())
+                .end("SetInstanceProtectionResponse")
+                .build());
+    }
+
+    private Response handleSetInstanceHealth(MultivaluedMap<String, String> p, String region) {
+        String instanceId = p.getFirst("InstanceId");
+        String healthStatus = p.getFirst("HealthStatus");
+        service.setInstanceHealth(region, instanceId, healthStatus);
```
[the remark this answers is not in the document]
sam-fakhreddine: Addressed in 63dd67cc6 together with utafrali's identical finding on the SetInstanceHealth thread (same root cause: ShouldRespectGracePeriod was dropped entirely). The handler now parses and passes the flag through to the service. I'm disclosing scope here rather than claiming full fix: actual grace-period *gating* (deferring the health-status change while the instance is within HealthCheckGracePeriod since launch) is NOT implemented — floci doesn't track instance launch time anywhere today, so building real gating means adding that state, which is a feature addition beyond this PR's scope. The maintainer (utafrali) explicitly endorsed wire-through without behavioral enforcement as sufficient for this PR ('even if the emulator does not act on the grace period yet'). Flagging this as a disclosed, deliberate scope limit rather than resolving it as fully fixed — happy to open a follow-up issue for real grace-period tracking if that's wanted.

src/main/java/io/github/hectorvent/floci/services/autoscaling/AutoScalingService.java
```diff
@@ -616,6 +659,30 @@ public void completeLifecycleAction(String region, String asgName, String hookNa
...
+            java.util.regex.Pattern.compile("^[A-Za-z0-9\\-_/]+$");
+
+    private static void validateLifecycleHookName(String hookName) {
+        if (hookName == null || hookName.length() > 255 || !LIFECYCLE_HOOK_NAME_PATTERN.matcher(hookName).matches()) {
+            throw new AwsException("ValidationError",
+                    "1 validation error detected: Value '" + hookName
+                            + "' at 'lifecycleHookName' failed to satisfy constraint: Member must satisfy "
+                            + "regular expression pattern: [A-Za-z0-9\\-_/]+", 400);
+        }
+    }
```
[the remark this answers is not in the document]
sam-fakhreddine: Fixed in 63dd67cc6 — split validateLifecycleHookName into three checks (null, length>255, pattern) each with its own message, so a >255-char name now gets a length-specific ValidationError message instead of the regex-pattern one. RED: putLifecycleHookRejectsOverlongNameWithLengthMessage (256-char name) and putLifecycleHookRejectsNullNameWithNotNullMessage assert the message content matches the actual failure; putLifecycleHookRejectsBadCharsWithPatternMessage pins the existing pattern-message case. All three green.

sam-fakhreddine:
Thanks for the thorough review, @utafrali — replied inline on each thread and resolved all 11 (7 from you, 3 from greptile, 1 from Copilot). Pushed in 63dd67cc6. Summary in your comment order:

1. **Redundant setInstanceHealth lookup** — collapsed to a single traversal, exactly as suggested.
2. **ShouldRespectGracePeriod dropped** — parsed and passed through end-to-end (default true when absent, matching the model). I deliberately stopped short of implementing actual grace-period gating: floci has no instance launch-time tracking today, so real enforcement is a feature addition, not a targeted fix — matches your own note that wire-through without behavioral enforcement is acceptable here. Same fix closes greptile's identical finding on this thread.
3. **Only first missing instance ID reported** — now joins the full requested-minus-found set into one message.
4. **requireGroup before empty-instanceIds check** — swapped so the missing-parameter error always wins over group-not-found.
5. **deleteLifecycleHookByName bypassing validation** — now calls validateLifecycleHookName for parity with deleteLifecycleHook, while keeping the null-hookName case as an idempotent no-op (matches how the CFN provisioner already guards this call).
6. **Fully-qualified Pattern** — added the import, dropped the qualifier.
7. **Column-alignment nit** — trimmed to match the surrounding switch block.

Also folded in greptile's set-containment fix for duplicate stored instance IDs (same block as #3) and Copilot's ask to split validateLifecycleHookName's null/length/pattern messages so each failure reports accurately.

Every fix has a RED test (quoted per-thread) and the full targeted suite is green (106 tests across AutoScalingServiceTest, AutoScalingQueryHandlerTest, AutoScalingIntegrationTest, AutoScalingReconcilerTest, the CFN lifecycle-hook provisioner test, and CloudFormationAutoScalingIntegrationTest).

One disclosed scope note, not a dispute: greptile's health-grace-period thread wants actual gating behavior, which I'm not implementing here for the reason in #2 above. Flagging in case that reasoning doesn't hold up on your end — happy to open a follow-up issue for real launch-time tracking if you'd rather have that ledgered explicitly.

pgermosen (commented):
Solid work, and the fix commit closing out the earlier review round holds up well under an independent re-check. I verified the substantive fixes against the current code and the real API reference rather than trusting the resolved threads: the validation-order fix on `setInstanceProtection` (empty `InstanceIds` checked before the group lookup), the switch to set-containment so duplicate stored instance IDs no longer produce a false match, the multi-ID error message now listing every unknown instance rather than just the first, and the `ProtectedFromScaleIn` required-boolean handling (confirmed `Required: Yes` against the reference) all check out exactly. `LifecycleHookName`'s pattern and length match the reference precisely, and it's good that `deleteLifecycleHookByName` now goes through the same validation as the primary delete path.

One thing worth fixing here: `completeLifecycleAction` calls `validateLifecycleActionToken` unconditionally, which requires the token to be exactly 36 characters and rejects it outright when it's `null`. I checked this against the actual service model rather than just the rendered docs page — `LifecycleActionToken` isn't in `CompleteLifecycleActionType`'s required list, and neither is `InstanceId`; the operation's own description confirms it identifies the action by "the specified token or instance." So a caller that identifies the action by `InstanceId` + `LifecycleHookName` + `AutoScalingGroupName` alone, without a token — a real, documented, and fairly common Auto Scaling calling pattern — gets rejected here with a length-36 error and would succeed on real AWS. Since `completeLifecycleAction` is currently stored-only, this validation is the entire observable behavior of the endpoint right now, so it's the one thing a token-less caller would actually notice. Gating the check behind a null check, the same way `LifecycleHookName`'s own null check is already a separate branch from its length/pattern checks a few lines above, should close it.

sam-fakhreddine:
Thanks for the independent re-verification, and good catch on `completeLifecycleAction`.

Fixed in 77b611620 — the token check is now gated behind a null check, the same pattern already used for `LifecycleHookName`'s own null vs. length/pattern branches. Verified against the current botocore model (`autoscaling/2011-01-01/service-2.json`, `CompleteLifecycleActionType.required` = `[LifecycleHookName, AutoScalingGroupName, LifecycleActionResult]`) — `LifecycleActionToken` and `InstanceId` are both optional, confirming your read.

RED-first: added `completeLifecycleActionAllowsNullTokenWhenInstanceIdIdentifiesTheAction`, which failed against the old code with the length-36 `ValidationError` you identified. `AutoScalingServiceTest` (38 tests) and `AutoScalingQueryHandlerTest` (10 tests) are green after the fix.

pgermosen (approved):
Verified directly in the current code rather than trusting the description — `completeLifecycleAction` now only runs the token length/pattern check when the token is non-null, mirroring the same null-vs-length/pattern split `LifecycleHookName`'s own validation already uses a few lines above it. That closes the gap where a valid, token-less `CompleteLifecycleAction` call would have been rejected.

Issue #2668, opened 2026-08-28
[BUG] Step Functions (SFN): a JSONata error cause says `Object` and `{{type}}` instead of the function name and the type


facuparedes:
## Service

Step Functions (`QueryLanguage: JSONata`)

## AWS API Action

`StartExecution`, `DescribeExecution`

## Expected behavior

When a JSONata expression throws, the cause names the function and the type it wanted. AWS:

```
T0412: Argument 1 of function "sum" must be an array of "numbers"
```

`jsonata-js` produces the same text for the same expression.

## Actual behavior

Floci renders the message with two substitutions unresolved: the function name comes out as the literal `Object`, and the type placeholder is never filled in.

```
Argument "[a]" of Object "n" must be an array of {{type}}
```

This is not one message. I ran 1674 expressions from the [jsonata-js v2.2.2 test suite](https://github.com/jsonata-js/jsonata/tree/v2.2.2/test/test-suite/groups) against Floci, one `Pass` state each; 277 of them are cases the suite expects to raise, and Floci does fail all 277. 117 of those causes carry an unresolved substitution: `Object` where a function name belongs in 116, and a literal `{{type}}` in 13 or `{{token}}` in 1, with 13 carrying both.

That run is every group of the suite except `function-eval`, which AWS documents as its one exclusion from JSONata 2.0.6.

It is also reachable without the suite, from a state machine written against the AWS docs. `$random(42)` is valid Step Functions JSONata, and on Floci it currently reaches JSONata's own zero-argument `$random`:

```
FAILED	States.QueryEvaluationError	Argument "1" of Object "random" does not match Object signature
```

PR #2676 removes that particular one by binding `$random` with the Step Functions signature, so it stops being a way to demonstrate this bug. It does not touch the 117 others.

Behavior is not affected: Floci already sets `error` to `States.QueryEvaluationError`, which is what a `Catch` matches on. This is a diagnosability problem.

## Reproduction

```bash
docker run -d --name floci -p 4566:4566 floci/floci:nightly
export AWS_REGION=us-east-1 AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test
export AWS_ENDPOINT_URL=http://localhost:4566

SM=$(aws stepfunctions create-state-machine --name errcode --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"QueryLanguage":"JSONata","StartAt":"E","States":{"E":{"Type":"Pass",
[...]

Issue #2667, opened 2026-08-28
[BUG] Step Functions (SFN): JSONata evaluation has no time or depth bound, so an expression can run forever


facuparedes:
## Service

Step Functions (`QueryLanguage: JSONata`)

## AWS API Action

`StartExecution`, `DescribeExecution`

## Expected behavior

AWS bounds JSONata evaluation and fails the state when the expression exceeds it. A recursive expression with no base case fails; so does an expression over a very large sequence.

## Actual behavior

Floci evaluates without a bound. A recursive expression pins a core and the execution stays `RUNNING` forever, on both the native image and the JVM. JSONata optimises tail calls, so this loops rather than overflowing the stack: no `Error` is thrown, and nothing in the executor can observe it.

## Reproduction

```bash
docker run -d --name floci -p 4566:4566 floci/floci:nightly
export AWS_REGION=us-east-1 AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test
export AWS_ENDPOINT_URL=http://localhost:4566

SM=$(aws stepfunctions create-state-machine --name hang2 --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"QueryLanguage":"JSONata","StartAt":"E","States":{"E":{"Type":"Pass",
    "Output":{"v":"{% ($f := function($x){ $f($x+1) }; $f(1)) %}"},"End":true}}}' \
  --query stateMachineArn --output text)

EXEC=$(aws stepfunctions start-execution --state-machine-arn "$SM" --input '{}' \
  --query executionArn --output text)

sleep 60
aws stepfunctions describe-execution --execution-arn "$EXEC" --query '[status,error,cause]' --output text
# RUNNING	None	None     one core pinned, no log line, no history event
```

Real AWS fails the same expression with `States.QueryEvaluationError`.

The same missing bound shows up without recursion, and here AWS names the limit it enforces:

```bash
SM=$(aws stepfunctions create-state-machine --name bigrange --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"QueryLanguage":"JSONata","StartAt":"E","States":{
    "E":{"Type":"Pass","Output":{"v":"{% [1..10000000] ~> $count() %}"},"End":true}}}' \
  --query stateMachineArn --output text)

EXEC=$(aws stepfunctions start-execution --state-machine-arn "$SM" --input '{}' \
  --query executionArn --output text)

sleep 20
[...]

Issue #2665, opened 2026-08-28
[BUG] Step Functions (SFN): a JSONata expression that returns nothing succeeds instead of failing the state


facuparedes:
## Service

Step Functions (`QueryLanguage: JSONata`)

## AWS API Action

`StartExecution`, `TestState`

## Expected behavior

When a JSONata expression assigned to a field returns undefined, real AWS fails the state with `States.QueryEvaluationError` and names the field in the cause, so a `Catch` on that error fires:

```
An error occurred while executing the state 'E' (entered at the event id #2). The JSONata
expression '$states.input.missing' specified for the field 'Output/v' returned nothing
(undefined).
```

## Actual behavior

Floci succeeds and drops the key. The state machine keeps going down the happy path with a field that is not there, and the `Catch` never fires.

This reaches us through `Assign`. One of our state machines reads an optional HTTP header into a variable, and the scenario asserting "header absent, execution fails" is green against AWS and silently wrong against Floci: the execution runs to the end with the variable unset.

This is not one expression. I ran 1674 expressions from the [jsonata-js v2.2.2 test suite](https://github.com/jsonata-js/jsonata/tree/v2.2.2/test/test-suite/groups) against Floci, one `Pass` state each; 108 of them are cases the suite marks `undefinedResult`, and Floci succeeds on 106, dropping the key in every one. That run is every group of the suite except `function-eval`, which AWS documents as its one exclusion from JSONata 2.0.6.

## Reproduction

```bash
docker run -d --name floci -p 4566:4566 floci/floci:nightly
export AWS_REGION=us-east-1 AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test
export AWS_ENDPOINT_URL=http://localhost:4566

SM=$(aws stepfunctions create-state-machine --name undef --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"QueryLanguage":"JSONata","StartAt":"E","States":{
    "E":{"Type":"Pass","Output":{"v":"{% $states.input.missing %}"},"End":true}}}' \
  --query stateMachineArn --output text)

EXEC=$(aws stepfunctions start-execution --state-machine-arn "$SM" --input '{}' \
  --query executionArn --output text)

aws stepfunctions describe-execution --execution-arn "$EXEC" --query '[status,output,error]' --output text
# SUCCEEDED	{}	None
```

Real AWS, same expression:

```bash
# Against real AWS. `env -u` drops, for this one command, both the local endpoint and
[...]

Issue #2522, opened 2026-08-24
[FEAT] Step Functions (SFN): implement the remaining ASL intrinsic functions


anikolaiev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Several ASL intrinsic functions are not implemented and fail the execution with `Unsupported intrinsic function: <name>`.

Implemented today (from `AslExecutor`): `States.Array`, `States.ArrayContains`, `States.ArrayLength`, `States.Format`, `States.JsonMerge`, `States.JsonToString`, `States.MathAdd`, `States.StringToJson`, `States.UUID`.

Missing, but part of the ASL intrinsic set:

- `States.ArrayGetItem`
- `States.ArrayPartition`
- `States.ArrayRange`
- `States.ArrayUnique`
- `States.Base64Encode`
- `States.Base64Decode`
- `States.Hash`
- `States.MathRandom`
- `States.StringSplit`

### Expected Behavior

These resolve as they do in real AWS and Step Functions Local.

### How are you starting Floci?

Docker image built from `main` (`257e004`), `docker/Dockerfile`.

### Steps To Reproduce

```json
{
  "StartAt": "Calc",
  "States": {
    "Calc": {
      "Type": "Pass",
      "Parameters": { "first.$": "States.ArrayGetItem($.items, 0)" },
      "End": true
    }
  }
}
```

Start with input `{"items": [1, 2, 3]}`. The execution fails with:

> Unsupported intrinsic function: States.ArrayGetItem

The already-implemented ones work correctly on valid input — `States.ArrayLength($.items)` and `States.JsonMerge($.a, $.b, false)` both return the right answer — so this is purely a coverage gap, not a defect in the intrinsic machinery.

AbarnaaSree:
I'd like to work on this issue. I'll investigate the existing intrinsic-function implementation in AslExecutor, implement the missing ASL intrinsic functions with AWS-compatible behavior, and add the necessary regression tests.

PR #2676, opened 2026-08-28
feat(stepfunctions): add the JSONata functions of the Step Functions dialect

### LINKED ISSUE #2662: [FEAT] Step Functions (SFN): the JSONata dialect's `$parse`, `$partition`, `$range`, `$hash` and `$uuid` are not bound, and `$random` rejects its seed
## Service

## API Action / Feature

The six JSONata functions of the Step Functions dialect, reached through `StartExecution` and `TestState`.

## AWS Documentation

https://docs.aws.amazon.com/step-functions/latest/dg/transforming-data.html

## Expected behavior

Step Functions adds six functions to the JSONata dialect. Five of them JSONata does not have at all; the sixth, `$random`, exists in JSONata but takes no argument, while Step Functions gives it an optional seed. On real AWS:

```
$parse("{\"a\":1}").a          ->  1
$partition([1,2,3,4], 2)       ->  [[1,2],[3,4]]
$range(0, 6, 2)                ->  [0,2,4,6]        end included when the step lands on it
$hash("input", "SHA-1")        ->  "140f86aae51ab9e1cda9b4254fe98a74eb54c1a1"
$random(42)                    ->  0.7275636800328681
$uuid()                        ->  a v4 UUID
```

`$hash` takes MD5, SHA-1, SHA-256, SHA-384 or SHA-512. `$random(seed)` is `java.util.Random`: seed 42 gives that number bit for bit on AWS.

## Actual behavior

The five that JSONata does not have fail with the JSONata error for calling something that is not a function.

```
FAILED  States.QueryEvaluationError  Attempted to invoke a non-function
```

`jsonata-js` gives the same error for the same expression. No JSONata engine ships these; they exist only in the Step Functions dialect.

`$random` fails differently, because JSONata's own zero-argument `$random` answers instead. Its signature refuses the seed:

```
FAILED  States.QueryEvaluationError  Argument "1" of Object "random" does not match Object signature
```

That cause is also an instance of the malformed-message problem in #2668. Binding `$random` with the Step Functions signature makes it go away.

This is not #2522. That issue is about the JSONPath intrinsics (`States.ArrayPartition`, `States.ArrayRange`, `States.Hash`), and the two dialects bind separately: #2522 lists `States.UUID` as already implemented, while `$uuid()` here still fails. Four of the six have an intrinsic counterpart and `$parse` and `$random` have none, but all six behave differently from AWS today whatever the JSONPath side supports.

## Reproduction

```bash

run() {
  SM=$(aws stepfunctions create-state-machine --name "$1" --role-arn arn:aws:iam::000000000000:role/r \
    --definition "$2" --query stateMachineArn --output text)
  sleep 3
  printf '%-10s %s\n' "$1" "$(aws stepfunctions describe-execution --execution-arn "$EX" \
    --query '[status,error,cause]' --output text)"
}

run parse     '{"QueryLanguage":"JSONata","StartAt":"E","States":{
  "E":{"Type":"Pass","Output":{"v":"{% $parse(\"{\\\"a\\\":1}\").a %}"},"End":true}}}'

run partition '{"QueryLanguage":"JSONata","StartAt":"E","States":{
  "E":{"Type":"Pass","Output":{"v":"{% $partition([1,2,3,4], 2) %}"},"End":true}}}'

run range     '{"QueryLanguage":"JSONata","StartAt":"E","States":{
  "E":{"Type":"Pass","Output":{"v":"{% $range(0, 6, 2) %}"},"End":true}}}'

run hash      '{"QueryLanguage":"JSONata","StartAt":"E","States":{
  "E":{"Type":"Pass","Output":{"v":"{% $hash(\"input\", \"SHA-1\") %}"},"End":true}}}'

run random    '{"QueryLanguage":"JSONata","StartAt":"E","States":{
  "E":{"Type":"Pass","Output":{"v":"{% $random(42) %}"},"End":true}}}'

run uuid      '{"QueryLanguage":"JSONata","StartAt":"E","States":{
  "E":{"Type":"Pass","Output":{"v":"{% $uuid() %}"},"End":true}}}'
```

Floci, all six:

```
parse      FAILED	States.QueryEvaluationError	Attempted to invoke a non-function
partition  FAILED	States.QueryEvaluationError	Attempted to invoke a non-function
range      FAILED	States.QueryEvaluationError	Attempted to invoke a non-function
hash       FAILED	States.QueryEvaluationError	Attempted to invoke a non-function
random     FAILED	States.QueryEvaluationError	Argument "1" of Object "random" does not match Object signature
uuid       FAILED	States.QueryEvaluationError	Attempted to invoke a non-function
```

Real AWS runs all six and returns the values listed under Expected behavior.

## Why is this needed?

Across our own 174 state machines, `$parse` is called 180 times in 99 of them, almost always as `$parse($states.result.Output)` after a nested `startSyncExecution`; `$uuid` in 4 and `$partition` in 2. Those machines fail on Floci even when the Task resource dispatches correctly, which is what stops us running our test suite locally.

## Suggested fix

Bind the six in `JsonataEvaluator`, next to where `$states` and the workflow variables are bound on the frame.

PR #2676 does that. It is additive (no existing line removed), adds 44 unit tests and 7 integration tests over the AWS edge cases, and I measured every value in it on real AWS with `test-state`: 65 expressions compared on both sides, 65 match, 0 differ. Several of those measurements contradict a plain reading of the docs, which is why they are in the PR body: a non-integer argument rounds towards zero rather than down, `$hash` with no algorithm is undefined rather than an error, and the documented 1000-element cap on `$range` belongs to the JSONPath `States.ArrayRange`, not here.

Four of the six have a JSONPath counterpart in #2522 (`$partition` / `States.ArrayPartition`, `$range` / `States.ArrayRange`, `$hash` / `States.Hash`, `$uuid` / `States.UUID`). I would land this first and have #2522's side call the same implementation rather than write the semantics twice.

## Environment

- [x] Yes

facuparedes:
## Summary

Step Functions extends JSONata with six functions of its own, `$parse`, `$partition`, `$range`, `$hash`, `$random` and `$uuid` ([AWS docs](https://docs.aws.amazon.com/step-functions/latest/dg/transforming-data.html)), and Floci binds none of them. Any state machine that calls one fails on Floci and runs on AWS. Closes #2662, which carries the one-container reproduction.

- **New functions (6)**: bound in `JsonataEvaluator` on the evaluation frame, next to `$states` and the workflow variables. The map is immutable and built once in the constructor (`JsonataEvaluator.java:69`); each evaluation binds it onto its own frame (`:107-108`).
- **`$parse` is the one that blocks real workloads.** AWS disables `$eval`, so `$parse` is the only way to read a field out of a nested execution's `Output`, which arrives as a JSON string.
- **`$random` was failing differently from the other five.** JSONata has its own zero-argument `$random`, so the library's signature check refused the seed and produced the malformed cause from #2668. Binding `$random` with the Step Functions signature removes that one instance and leaves the other 117 cases alone. A bare `$random()` keeps working, on AWS and here (`random_unseededCallsDiffer`).
- **Docs**: a new `## JSONata functions` section in `docs/services/step-functions.md`, written by hand outside the generated action-table markers.

`test-state` evaluates one state and creates nothing, so you can re-check any row below in a single call:

```bash
ROLE=arn:aws:iam::<account>:role/<any-role>   # test-state creates nothing, so any role works

  --definition '{"Type":"Pass","QueryLanguage":"JSONata","Output":{"v":"{% $range(-1.7, 2, 1) %}"},"End":true}'
# [ "SUCCEEDED", "{\"v\":[-1,0,1,2]}", null, null ]
```

| --- | --- | --- |
| `$range(0, 6, 2)` | `[0,2,4,6]` | `rangeIncludesTheEndValueWhenTheStepLandsOnIt` |
| `$range(-1.7, 2, 1)` | `[-1,0,1,2]` | `roundingGoesTowardsZeroNotDownwards` |
| `$partition([1,2,3,4,5], 2)` | `[[1,2],[3,4],[5]]` | `partition_splitsIntoChunksWithShorterLastChunk` |
| `$partition([1,2,3], -1)` | `D3137: Second argument must be zero or greater` | `partition_negativeChunkSizeRaisesError` |
| `$hash("input")` | undefined, not an error | `hash_missingAlgorithmIsUndefinedButNonStringOneIsASignatureError` |
| `$hash("input", "SHA-1")` | `140f86aae51ab9e1cda9b4254fe98a74eb54c1a1` | `hashComputesTheHexDigestForEachSupportedAlgorithm` |
| `$parse('{"n":9223372036854775807}').n` | `9223372036854775807` | `parse_keepsAnIntegerExactWhileItFitsInALong` |
| `$parse('{"n":9223372036854775808}').n` | `9.223372036854776E18` | same test |
| `$parse('1.0')` | `1` | `parse_dropsATrailingZeroOnAnIntegerValuedNumber` |
| `$parse('')` | `D3137: Invalid JSON` | `parse_rejectsJsonAwsRejects` |
| `$random(42)` | `0.7275636800328681` | `random_seedMakesTheDrawReproducible` |

I compared 65 expressions this way, Floci against AWS: **65 match, 0 differ**. The set covers each function's normal case, every arity including the surplus-argument error, both signs of a non-integer argument, the undefined cases, all five hash algorithms, the `$parse` number model at the `long` boundary, and the JSON AWS rejects. Five of those comparisons changed the implementation after the first pass: the rounding direction, `$hash` with no algorithm, `$partition` with a negative chunk size, the surplus-argument message, and the `$random` seed semantics.

Three things the table cannot show:

- `$random(seed)` is `java.util.Random`, and seeds 0, 7 and 42 match AWS bit for bit. That survives a JDK upgrade because `Random`'s generator is specified in its Javadoc. Unseeded draws come from `ThreadLocalRandom.current()` and `$uuid()` from `UUID.randomUUID()`, so neither shares state across concurrent executions.
- `$hash` digests the UTF-8 bytes of the string (`hash_computesTheRemainingAlgorithmsAndHashesUtf8Bytes`), and AWS treats the algorithm name as case-sensitive (`hash_algorithmNameIsCaseSensitive`).
- `$range` collapses a single-element range to a bare number rather than a one-element array (`range_singleElementRangeCollapsesToScalar`).

## Deliberate scope notes

- **Rounding goes towards zero, not down.** The docs say a non-integer argument is rounded and do not say which way. AWS truncates: `$range(-1.7, 2, 1)` is `[-1,0,1,2]`, and flooring would shift every negative argument by one. A Java cast to `long` already truncates, so `towardsZero` (`JsonataEvaluator.java:250`) is the whole implementation.
- **Undefined and error are not interchangeable.** Five cases evaluate to undefined and one raises (`$partition(a, -1)`). I measured every one of them, and the new docs section lists them. What Floci then does with an undefined value downstream is #2665 and out of scope here.
- **No element cap on `$range`, and no memory bound.** The documented 1000-element cap belongs to the JSONPath intrinsic `States.ArrayRange`, not to the JSONata `$range`: `$count($range(1, 200000, 1))` returns `200000` on AWS. What AWS does enforce is an evaluation-wide memory limit, somewhere between 200k and 500k elements. That limit is not a `$range` feature, so it belongs with #2667, and the new docs section records it as this PR's one deviation instead of half-implementing it here.
- **`$range` still always terminates.** The obvious `v += step` loop never ends for a step near `Long.MAX_VALUE`: the addition wraps and lands back below the end on every pass. The element count is computed in `BigInteger` up front and the loop runs a fixed number of times (`JsonataEvaluator.java:398`). `range_stepNearLongMaxTerminates` hangs without it.
- **Each signature declares one optional slot more than AWS's arity, so a surplus argument reaches our code.** dashjoin refuses a call longer than the declared signature with the malformed message from #2668, `Argument "1" of Object "random" does not match Object signature`. The extra slot lets `rejectSurplusArgument` (`JsonataEvaluator.java:235`) raise AWS's own `T0410: Argument N of function "x" does not match function signature` instead. A null signature is the other way to get there, but it makes the library throw a raw `NullPointerException` inside `$map` and friends, because `Functions.getFunctionArity` dereferences it. `functionsAreUsableAsAValueInAHigherOrderCall` catches a regression to that.
- **Land this before #2522** (the JSONPath intrinsics, which bind separately): four of the six have a counterpart there (`$partition` / `States.ArrayPartition`, `$range` / `States.ArrayRange`, `$hash` / `States.Hash`, `$uuid` / `States.UUID`) and can call this implementation instead of restating the semantics.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] Verified against real Step Functions in `us-east-1` with AWS CLI 2.33.7, through `test-state`, which creates nothing
- [x] 65-expression differential, Floci against AWS: 65 match, 0 differ. Every number, digest and error message in this PR and in the tests is a captured AWS response, not a value derived from the documentation

## Checklist

- [x] Targeted Step Functions suites green (411 tests, 0 failures/errors)
- [x] New integration test added: `StepFunctionsJsonataFunctionsIntegrationTest`, 7 tests through `CreateStateMachine` and `StartExecution`. With `JsonataEvaluator` reverted to `main` and the tests untouched, all 7 fail
- [x] 44 new unit tests in `JsonataEvaluatorTest`, on top of the 16 already there
- [x] `make docs-check` passes and `make docs-test` is 34 passed; the generated action table is untouched
- [x] Rebased onto current `origin/main`
- [ ] The full `./mvnw test` run is 13522 tests with 2 failures outside this diff: `Ec2ContainerManagerTest.launchLabelsContainerWithResourceIdentity` and `KubernetesPodLauncherTest.launchCreatesPodAndReturnsHandle`. Both pass in isolation on this branch, and neither touches `stepfunctions`

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -66,6 +66,34 @@ uniformly between zero and the computed delay, as on AWS. One deviation. The del
 between attempts is capped at 30 seconds, the same cap Floci applies to `Wait` states,
 so emulated runs stay fast.
 
+## JSONata functions
+
+State machines with `"QueryLanguage": "JSONata"` reach the six functions Step Functions adds on
+top of the JSONata language, alongside every function JSONata itself provides.
+
+| Function | Returns |
+| --- | --- |
+| `$parse(jsonString)` | the deserialized value; the replacement for `$eval`, which AWS disables |
+| `$partition(array, chunkSize)` | `array` split into chunks of `chunkSize`, the last one holding the remainder |
+| `$range(start, end, step)` | the values from `start` to `end`, inclusive when `step` lands on `end` |
+| `$hash(str, algorithm)` | the hex digest of `str`; `algorithm` is `MD5`, `SHA-1`, `SHA-256`, `SHA-384` or `SHA-512`, case-sensitive |
+| `$random(seed)` | a number in `[0, 1)`, reproducible under the optional integer `seed` |
+| `$uuid()` | a v4 UUID |
+
+Three behaviours are worth knowing before reading an unexpected result, and all three are AWS's:
+
+- A non-integer argument is rounded **towards zero**, so `$range(-1.7, 2, 1)` starts at `-1` and
+  `$partition(items, 2.9)` chunks by 2.
+- Several arguments evaluate to undefined rather than failing: a chunk size of zero, an empty
+  array, a `$range` with no `step` or with a step whose sign disagrees with the direction, and a
+  `$hash` with no algorithm. An undefined value is then dropped from the surrounding `Output`
+  object, so the field goes missing instead of the state failing.
+- `$range` collapses a single-element range to the bare number, not a one-element array.
+
+One deviation. AWS bounds the memory an expression may use and fails a `$range` of roughly half a
+million elements with `Expression evaluation memory limit exceeded`; Floci has no such bound and
+builds the array.
+
 ## Mocked service integrations
 
 Floci supports the Step Functions Local mock configuration format
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluator.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluator.java
@@ -1,16 +1,33 @@
 package io.github.hectorvent.floci.services.stepfunctions;
 
+import com.dashjoin.jsonata.JException;
 import com.dashjoin.jsonata.Jsonata;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 
+import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.HexFormat;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Random;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
 
 import static io.github.hectorvent.floci.services.stepfunctions.AslExecutor.FailStateException;
 
@@ -28,11 +45,28 @@ import static com.dashjoin.jsonata.Jsonata.jsonata;
 @ApplicationScoped
 public class JsonataEvaluator {
 
+    private static final Set<String> HASH_ALGORITHMS = Set.of("MD5", "SHA-1", "SHA-256", "SHA-384", "SHA-512");
+
+    /**
+     * The largest magnitude a {@code double} can still hold as an exact integer count of longs.
+     * Past it AWS switches from an integer to exponent notation, and so does {@link #toJsonataValue}.
+     */
+    private static final double LARGEST_EXACT_LONG_AS_DOUBLE = 9.223372036854776E18;
+
     private final ObjectMapper objectMapper;
+    private final ObjectReader strictJsonReader;
+    private final Map<String, Jsonata.JFunction> stepFunctionsExtensions;
 
     @Inject
     public JsonataEvaluator(ObjectMapper objectMapper) {
         this.objectMapper = objectMapper;
+        // AWS rejects what a lenient parser accepts: a second document after the first, a repeated
+        // key, an empty string. All three come back as D3137 there, so $parse reads through this
+        // reader rather than the shared mapper, whose settings belong to the wire protocol.
+        this.strictJsonReader = objectMapper.reader()
+                .with(DeserializationFeature.FAIL_ON_TRAILING_TOKENS)
+                .with(JsonParser.Feature.STRICT_DUPLICATE_DETECTION);
+        this.stepFunctionsExtensions = buildStepFunctionsExtensions();
     }
 
     /**
@@ -71,6 +105,7 @@ public class JsonataEvaluator {
         try {
             Jsonata jsonataExpr = jsonata(expr);
             Jsonata.Frame frame = jsonataExpr.createFrame();
+            stepFunctionsExtensions.forEach(frame::bind);
             // Workflow variables (the Assign field) are referenced as top-level $name in AWS's
             // JSONata dialect, e.g. $CheckpointCount. They are bound alongside $states, never
             // inside it: AWS reserves $states for input/result/errorOutput/context only. $states is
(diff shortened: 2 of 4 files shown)


pgermosen (approved):
Excellent, unusually rigorous work. I worked through the implementation independently rather than taking the clean bot review at face value, and it holds up well beyond a surface check. All six functions, their arities, and their documented purposes match the reference exactly. A few things I specifically traced through rather than assumed: the signature-padding trick that lets a genuine surplus argument reach your code instead of the underlying library's malformed error, and why a null signature isn't a viable alternative — it throws a raw NullPointerException inside $map/$filter/$sift/$each/$single/$reduce, which one of your new tests specifically pins. The $range termination fix is a real bug fix, not defensive over-engineering: a naive accumulate-and-compare loop genuinely hangs forever once the step gets close to Long.MAX_VALUE, because the addition wraps and lands back below the end on every pass — computing the element count in BigInteger up front and iterating exactly that many times is the right call. $hash's case-sensitivity only works because the exact-case set-membership check runs before MessageDigest.getInstance, which is case-insensitive on its own and would otherwise silently accept a lowercase algorithm name. And the $parse number model correctly reproduces the specific long-vs-double boundary and trailing-zero-dropping behavior you documented.

One thing worth knowing, not a request for a change: AWS's own docs page states plainly that integer parameters are "automatically round[ed] down," which read literally is floor — but your own differential testing (and the roundingGoesTowardsZeroNotDownwards test) is unambiguous that the real behavior is truncation towards zero. I have no way to independently confirm which is right without live access myself, but a specific, reproducible test-state call is much stronger evidence than one imprecisely worded doc sentence, especially since you say this is one of the five places live testing overturned your first implementation attempt. Might be worth a one-line callout in the new docs section noting AWS's own wording says "round down" while the verified live behavior is "truncate towards zero," so a future reader who checks the AWS page first doesn't think Floci disagrees with AWS by accident rather than by a verified correction.

PR #2679, opened 2026-08-28
fix(rds): grant the MySQL/MariaDB master user RDS-master-equivalent privileges

matanbaruch:
## Problem

A non-root MySQL/MariaDB master user cannot run the operations real RDS master users perform routinely. The stock `mysql`/`mariadb` images grant `MYSQL_USER` only `ALL` on `MYSQL_DATABASE` plus global `USAGE`:

```
GRANT USAGE ON *.* TO `admin`@`%`
GRANT ALL PRIVILEGES ON `admin`.* TO `admin`@`%`
```

so `CREATE DATABASE`, `CREATE USER` and `GRANT` all fail:

```
$ mysql -h localhost -P 7001 -uadmin -p*** -e "CREATE DATABASE spike_db;"
ERROR 1044 (42000): Access denied for user 'admin'@'%' to database 'spike_db'
```

On real RDS the master user holds a near-global privilege list, and the terraform/pulumi `mysql` providers depend on it: they connect to the instance endpoint during apply and immediately create schemas, users and grants. Against Floci those applies fail today.

## Fix

After the DB container starts, exec a root `GRANT ALL PRIVILEGES ON *.* TO '<master>'@'%' WITH GRANT OPTION` into it, reusing the exec-with-retry machinery the PostgreSQL `rds_iam` init already uses (extracted into a shared `execUntilSuccess`, with the PostgreSQL log/error text unchanged). The retry loop also absorbs the image's init window: the grant only succeeds once the entrypoint has created the master user.

Notes on the shape:

- **Root master is skipped.** The images create no separate user for `MasterUsername=root` and root already holds every privilege.
- **`GRANT ALL` is a disclosed deviation.** Real RDS withholds `SUPER`, `FILE` and `SHUTDOWN` from its master grant list. The emulator grants `ALL` for simplicity, mirroring how `POSTGRES_USER` is already a full superuser on the postgres side.
- **MariaDB uses the `mariadb` client binary.** Images at Floci's supported floor (10.11+) no longer ship a `mysql` symlink.
- **Runs where the postgres `rds_iam` init already runs**, so the blocking behavior matches the existing engine-init precedent (provisioning is async per #2525, the CreateDBInstance response is unaffected).
- PostgreSQL master users are unaffected and need nothing: `POSTGRES_USER` is a superuser.

## Tests

- `mysqlMasterGrantSqlGrantsGlobalPrivilegesWithGrantOption`, `masterGrantCommandUsesTheEngineClientBinary`, `needsMasterGrantSkipsRootAndPostgres`: the statement, the per-engine client binary, and the dispatch rule (extending the `postgresIamRoleInitSql` static-method test pattern).
- `mysqlRootMasterStartDoesNotExecIntoTheContainer`: a root-master start never touches the Docker exec API.
- All 27 `RdsContainerManagerTest` and 235 `RdsServiceTest` tests pass.

## Verified end-to-end

Ran the patched build in a container with `/var/run/docker.sock` and created instances with `--master-username admin` for both engines:

```
INFO [RdsContainerManager] Initialized master-user grants in RDS container floci-rds-db-F5C0659E...   (mysql:8.0.36)
INFO [RdsContainerManager] Initialized master-user grants in RDS container floci-rds-db-2DDCB9B4...   (mariadb:11)
```

Through the auth proxy as `admin`, the previously failing operations now succeed on both engines:

```
mysql> SHOW GRANTS FOR CURRENT_USER();
GRANT SELECT, INSERT, ..., CREATE USER, ... ON *.* TO `admin`@`%` WITH GRANT OPTION
mysql> CREATE DATABASE fixtest_db;            -- was ERROR 1044
mysql> CREATE USER 'appuser'@'%' IDENTIFIED BY '***';
mysql> GRANT ALL ON fixtest_db.* TO 'appuser'@'%';
```

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/container/RdsContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/container/RdsContainerManager.java
@@ -171,7 +171,7 @@ public class RdsContainerManager {
             EndpointInfo endpoint = info.getEndpoint(enginePort);
 
             LOG.infov("RDS backend for instance {0}: {1}", instanceId, endpoint);
-            initializeEngine(containerName, info.containerId(), engine, masterUsername);
+            initializeEngine(containerName, info.containerId(), engine, masterUsername, masterPassword);
 
             handle = new RdsContainerHandle(
                     info.containerId(), effectiveRuntimeId, instanceId,
@@ -371,9 +371,13 @@ public class RdsContainerManager {
         return Integer.parseInt(tag.substring(0, end));
     }
 
-    private void initializeEngine(String containerName, String containerId, DatabaseEngine engine, String masterUsername) {
+    private void initializeEngine(String containerName, String containerId, DatabaseEngine engine,
+                                  String masterUsername, String masterPassword) {
         if (engine == DatabaseEngine.POSTGRES) {
             initializePostgresIamRole(containerName, containerId, masterUsername);
+        } else if (needsMasterGrant(engine, masterUsername)) {
+            execUntilSuccess(containerName, containerId,
+                    masterGrantCommand(engine, masterUsername, masterPassword), "master-user grants");
         }
     }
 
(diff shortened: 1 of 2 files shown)


pgermosen (commented):
Real, well-verified problem — the end-to-end log and SQL output for both mysql:8.0.36 and mariadb:11 is concrete evidence rather than a claim, and the shared exec-with-retry extraction is a clean refactor of the existing postgres logic with the log text preserved.

Two things worth fixing here. First, `mysqlMasterGrantSql` concatenates `masterUsername` directly into the GRANT statement — `"GRANT ALL PRIVILEGES ON *.* TO '" + masterUsername + "'@'%' WITH GRANT OPTION;"` — and the comment justifies this as safe because MasterUsername is API-constrained to word characters. That's true of real AWS, but Floci itself doesn't enforce that constraint anywhere: the query handler reads MasterUsername straight off the raw request parameter, and the service layer only checks it's non-blank. A MasterUsername containing a single quote would get interpolated directly into SQL executed as root inside the container. I don't want to overstate the practical exploitability here — whoever calls CreateDBInstance already has full control over their own Floci instance and its Docker environment — but it's a genuine, concrete injection vector in code this PR introduces rather than something inherited: the postgres path this was extracted from never had this shape, since its fixed SQL has no interpolated identifiers and its `-U` username goes in as its own argv element rather than into concatenated SQL text. Worth guarding against even just as a typo/templating accident, not only a deliberate attacker.

Second, and I think this one is more likely to actually get hit in normal use: rebooting a non-root MySQL/MariaDB instance after its master password has been rotated via ModifyDBInstance will fail. `rebootDbInstance` restarts the container against the same Docker volume, passing whatever password is currently stored. MYSQL_ROOT_PASSWORD only takes effect against an empty data directory, so on a reused volume the entrypoint skips re-initialization and the database's actual root password stays whatever was baked in at first creation. The new container gets the rotated password as MYSQL_ROOT_PASSWORD, which the entrypoint ignores, but the grant step then tries to authenticate as root with that same rotated password against a database whose real root password is still the old one — authentication fails, the retry loop exhausts all 60 attempts, and the reboot itself fails with a timeout. That's an ordinary ModifyDBInstance-then-RebootDBInstance sequence, not an edge case, and it's a new failure mode this PR introduces since nothing previously depended on a password-authenticated exec surviving a container restart on a reused volume.

I also looked at the password appearing on the exec command line (-p + masterPassword) but didn't flag it — the same secret is already sitting in the container's own environment as MYSQL_ROOT_PASSWORD, so this doesn't add exposure beyond what's already there for anyone with that level of Docker access.

matanbaruch:
Both fixed in d9394d2b.

1. Injection: the username is now escaped (single quotes and backslashes) before it lands in the GRANT, unit test covers a `we'ird\` name. You are right that Floci never enforces the AWS charset, the comment claiming it does is gone.
2. Reboot after rotation: I removed the root exec entirely. The grant now ships as a `/docker-entrypoint-initdb.d` script installed between create and startCreated (same tar-and-copy shape as CodeBuildRunner.copySourceToContainer). The entrypoint runs it exactly once, against an empty data dir, as root, after creating the master user, so a reboot on a reused volume never re-runs it and never authenticates as root. Verified the exact sequence you described: create with a non-root master, ModifyDBInstance password rotation, RebootDBInstance, and the instance comes back available. Fresh creates re-verified on mysql:8.0.36 and mariadb:11.

One thing I hit while testing, pre-existing and untouched by this PR: connecting with the rotated password fails even without any reboot, since the proxy dials the backend with the rotated password while the backend container keeps the original. I can file that as a separate issue if you want.

@pgermosen ready for another look.

pgermosen (approved):
Both fixes verified directly in the code rather than trusting the description, and the reboot fix is better than what I expected — a real structural fix, not a band-aid. The escaping in `mysqlMasterGrantSql` handles backslashes before quotes, which is the order that actually matters here, and I worked through the escaped output by hand against the new test to confirm it round-trips correctly rather than just looking plausible.

Moving the grant to a `/docker-entrypoint-initdb.d` script installed between container create and start, instead of a post-start root exec, is the right way to close the rotation-then-reboot race: the image only runs those scripts once, against an empty data directory, exactly when the root password env var is honored, so a reused volume on reboot never re-runs it and never needs to re-authenticate as root with a possibly-stale password. I confirmed the old exec-based path was removed entirely rather than left alongside the new one.

On the new finding that landed since (pre-existing volumes from before this fix won't retroactively get the grant): I don't think that needs solving here. The only way to close it would be reintroducing some form of post-start root authentication against a volume whose true root credential Floci can't reliably know after arbitrary rotations, which is exactly the problem this PR eliminated. Recreating the instance to pick up the fix reads as an acceptable, disclosable limitation rather than something worth reopening the harder problem for.

PR #2682, opened 2026-08-28
fix(docdb): refuse an engine version a live account does not list

### LINKED ISSUE #2681: DocumentDB: engine-version change can assign a nonexistent managed default parameter group
`DocDbService.parameterGroupFamily(engineVersion)` derives `"docdb" + major + "." + minor` from any engine version string. Nothing checks that the resulting family is one of the three families `RdsService` actually registers as managed cluster parameter groups (`docdb3.6`, `docdb4.0`, `docdb5.0`).

Two call sites build `"default." + family` and assign it to the cluster with no existence check:

- `DocDbService.createDbCluster` → `resolveClusterSettings`, when no `DBClusterParameterGroupName` is given (`DocDbService.java:298`).
- `resolveClusterSettings`'s implicit branch that follows a cluster on a managed default group onto the new engine version's family default after an engine-version change (`DocDbService.java:302-304`).

A `CreateDBCluster` or `ModifyDBCluster` call with an engine version outside the three supported families (e.g. `9.9.9`) leaves the cluster's `DbClusterParameterGroupName` pointing at a parameter group `RdsService` has no record of. The only existing test on this path (`DocDbServiceTest`, the `4.0.0 → 5.0.0` engine-version-change case) only exercises two supported families, so it doesn't catch this.

Real AWS validates the engine version against its known set before this logic would ever run, so this needs an unsupported version to trigger and isn't reachable through normal use — but the emulator should validate the engine version (or at least the derived family) explicitly rather than relying on callers only ever passing a supported one.

Follow-up from #2651.

ck-michals:
Closes #2681

### What was wrong
`parameterGroupFamily(engineVersion)` derived `docdb<major>.<minor>` from any string, and both `CreateDBCluster` (no group given) and the engine-version branch of `ModifyDBCluster` assigned `default.<family>` without checking that such a group exists — a cluster on `9.9.9` ended up on `default.docdb9.9`, which RDS has no record of.

### What changes
- `DocDbService` carries the versions a live account lists for docdb, with their families: `3.6.0→docdb3.6`, `4.0.0→docdb4.0`, `5.0.0`/`5.0.1→docdb5.0`, `8.0.0`/`8.0.1→docdb8.0`. A version outside it is refused before any change, on create and on modify, with the live answer: `InvalidParameterCombination: Cannot find version 9.9.9 for docdb`. The `major.minor` form of a listed version is accepted, as it is live (`5.0` passes, `5` does not).
- `parameterGroupFamily` reads the catalogue first; the derivation stays only for records persisted before versions were checked, so they keep describing as they always have.
- `default.docdb8.0` joins RDS's managed cluster parameter groups — the family floci was missing.
- `DocDbIntegrationTest.modifyCluster` modified to `6.0.0`, a version that does not exist; it now uses `8.0.0`.

### Stated deviation
A version given as `5.0` is stored as given; what a live cluster reports for that form after create was not observed.

### Evidence
Live account: `9.9.9`, `5`, `abc` refused with the wording above ahead of every other check; `3.6.0`, `4.0.0`, `5.0`, `5.0.1` pass version validation; `DescribeDBEngineVersions` gives the six versions and families.

### Tests
`DocDbServiceTest.engineVersionsOutsideTheCatalogueAreRefusedBeforeAnythingChanges` (four bad versions on create leave no record; `8.0.1 → default.docdb8.0`; `5.0` kept; null → `5.0.0`; bad version on modify leaves version and group untouched). `RdsServiceTest` managed-group list updated. DocDb* and RdsServiceTest: 310/310.

### DIFF
--- a/compatibility-tests/sdk-test-awscli/test/rds.bats
+++ b/compatibility-tests/sdk-test-awscli/test/rds.bats
@@ -220,6 +220,19 @@ teardown() {
     aws_cmd ec2 delete-vpc --vpc-id "$VPC_ID" >/dev/null 2>&1 || true
 }
 
+@test "docdb: an engine version a live account does not list is refused" {
+    CLUSTER_ID="bats-docdb-version-$(unique_name)"
+    run aws_cmd docdb create-db-cluster --db-cluster-identifier "$CLUSTER_ID" --engine docdb \
+        --engine-version 9.9.9 --master-username u --master-user-password pw12345678
+    assert_failure
+    assert_output --partial 'InvalidParameterCombination'
+    assert_output --partial 'Cannot find version 9.9.9 for docdb'
+
+    run aws_cmd docdb describe-db-clusters --db-cluster-identifier "$CLUSTER_ID"
+    assert_failure
+    assert_output --partial 'DBClusterNotFoundFault'
+}
+
 @test "docdb: cluster settings given on create are returned by describe" {
     CLUSTER_ID="bats-docdb-settings-$(unique_name)"
     run aws_cmd kms create-key --description "$CLUSTER_ID"
--- a/docs/services/docdb.md
+++ b/docs/services/docdb.md
@@ -43,7 +43,9 @@ and so are the windows (30-minute minimum, no overlap). Omitted values take the 
 `default`, `default.docdb<family>`, the VPC's default security group, one day of backups — with
 deterministic windows `04:00-06:00` / `mon:00:00-mon:03:00` where AWS picks random ones.
 `CreateDBInstance` keeps `AutoMinorVersionUpgrade`, `PreferredMaintenanceWindow`,
-`CopyTagsToSnapshot`, `PromotionTier` and `Tags`.
+`CopyTagsToSnapshot`, `PromotionTier` and `Tags`. `EngineVersion` must be one a live account lists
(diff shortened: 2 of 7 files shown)


pgermosen (approved):
Small, precise, well-verified fix. I checked the version catalogue against AWS's own current DocumentDB release notes rather than taking the live-account claim at face value: the four major families (3.6, 4.0, 5.0, 8.0) and the specific patch versions listed (5.0.0/5.0.1, 8.0.0/8.0.1) match exactly — 5.0.1 is documented as the first minor release for the 5.0 line, with no 5.0.2 anywhere, which directly confirms your own test asserting 5.0.2 is correctly refused as nonexistent rather than an arbitrary choice of bad input.

The major.minor acceptance logic is correct too — checking whether "docdb" + version is one of the catalogue's family values is exactly the right way to accept 5.0 (docdb5.0 is a real family, shared by both 5.0.0 and 5.0.1) while refusing 5 (docdb5 isn't a value anywhere). And keeping the old derivation as a fallback only for versions outside the catalogue is the right design, since that path is now only reachable for records persisted before this validation existed — a pre-existing cluster keeps describing itself the way it always has rather than having its parameter-group family change out from under it because of this fix.

I also checked that a refused request has no side effects — the test confirms a rejected create leaves no cluster behind, and a rejected modify leaves both the version and parameter group untouched.

PR #2683, opened 2026-08-28
fix(stepfunctions): write a large whole number in full in $string

### LINKED ISSUE #2669: [BUG] Step Functions (SFN): JSONata `$string` on a large integer returns exponent notation
## Service

## AWS API Action

`StartExecution`

## Expected behavior

`$string(1e20)` returns `"100000000000000000000"`. AWS and `jsonata-js` agree.

## Actual behavior

Floci returns `"1e+20"`.

```
$string(1e20)
AWS         "100000000000000000000"
jsonata-js  "100000000000000000000"
Floci       "1e+20"
dashjoin    "1e+20"
```

The execution succeeds either way, so nothing surfaces this at run time. `$string` of an id feeds a `MessageGroupId`, an S3 key or a DynamoDB key often enough that the two engines write different keys for the same input and neither reports an error.

## Reproduction

```bash

SM=$(aws stepfunctions create-state-machine --name strbig --role-arn arn:aws:iam::000000000000:role/r \
    "Output":{"v":"{% $string(1e20) %}"},"End":true}}}' \

# SUCCEEDED	{"v":"1e+20"}      AWS returns {"v":"100000000000000000000"}
```

## Suggested fix

The behavior comes from dashjoin, but it does not need an upstream release: a `frame.bind` shadows a library builtin of the same name, which I verified against 0.9.10. So `$string` can be corrected in `JsonataEvaluator` through the same path used to bind `$states` and the workflow variables.

## Environment

- Java SDK version: AWS CLI 2.33.7, `jsonata-js` v2.2.2 and `com.dashjoin:jsonata:0.9.10` as references

facuparedes:
## Summary

`$string` switched to exponent notation at `Long.MAX_VALUE`, so `$string(1e20)` came back as `1e+20` where AWS writes the twenty-one digits. AWS's boundary is `1e21`, and the gap between the two is where an id lives: the execution succeeds either way, so the two engines wrote different `MessageGroupId`s and S3 keys for the same input and neither reported an error. Closes #2669.

- **The boundary is `1e21`, on both signs.** Measured on both sides and at the largest double under it, `999999999999999868928`, which AWS writes in full as `999999999999999900000`.
- **The digits are the shortest decimal that reads back as the same double**, not the double's exact value: 2^63 is `9223372036854776000` on AWS, not `9223372036854775808`. That is the value `$parse` produces one step past the `long` boundary #2676 established.
- **`$string` is bound on the evaluation frame** next to the six functions of the dialect (`JsonataEvaluator.java:234`), with the library's own `<x-b?:s>` signature, so every arity and type error stays what it was.
- **The library still formats everything else.** Below the boundary a double is replaced by the exact integer of those digits (`JsonataEvaluator.java:370`) and the built-in prints it; the escaping, the prettify layout and the fractional number model are untouched. A number nested in an object or an array goes through the same replacement, so `$string({"n": 1e20})` is fixed with the scalar.

`test-state` evaluates one state and creates nothing, so any row below is one call:

```bash
ROLE=arn:aws:iam::<account>:role/<any-role>   # test-state creates nothing, so any role works

  --definition '{"Type":"Pass","QueryLanguage":"JSONata","Output":{"v":"{% $string(1e20) %}"},"End":true}'
# [ "SUCCEEDED", "{\"v\":\"100000000000000000000\"}", null, null ]
```

| --- | --- | --- |
| `$string(1e19)` | `10000000000000000000` | `string_writesAWholeNumberInFullBelowTheExponentBoundary` |
| `$string(1e20)` | `100000000000000000000` | same test |
| `$string(999999999999999868928)` | `999999999999999900000` | same test |
| `$string(1e21)` | `1e+21` | `string_switchesToExponentNotationAt1e21` |
| `$string(1e21 - 1)` | `1e+21` | same test |
| `$string(-1e20)` | `-100000000000000000000` | `string_flipsAtTheSameBoundaryOnANegativeNumber` |
| `$string(-1e21)` | `-1e+21` | same test |
| `$string($parse('{"n":9223372036854775807}').n)` | `9223372036854775807` | `string_writesTheShortestDigitsThatReadBackAsTheSameDouble` |
| `$string($parse('{"n":9223372036854775808}').n)` | `9223372036854776000` | same test |
| `$string({"n": 1e20})` | `{"n":100000000000000000000}` | `string_writesANumberInsideAnObjectOrAnArrayTheSameWay` |
| `$string(1e-7)` | `1e-7` | `string_leavesEveryOtherValueToTheBuiltIn` |
| `$string(1/3)` | `0.333333333333333` | same test |

The last two rows already matched and are regression pins: AWS also switches to exponent notation under `1e-6`, and keeps fifteen significant digits on a fractional number.

## Deliberate scope notes

- **The rule changed, it was not widened.** `Functions.string` writes a whole double in full while `d <= Long.MAX_VALUE` and prints `Math.round(d)`, the exact value. AWS stops elsewhere and prints different digits, so the branch is replaced rather than extended.
- **Mantissa precision above `1e21` is left alone.** `$string(1.7976931348623157e308)` is `1.7976931348623157e+308` on AWS and `1.79769313486232e+308` here: the library rounds to fifteen digits before printing an exponent. A digit count, not a notation, and only reachable near `Double.MAX_VALUE`.
- **An integer literal that fits in a long is a long here and a double on AWS.** `$string(4611686018427387904)` is `4611686018427388000` there. That is `Utils.convertNumber` narrowing the literal, not `$string`, and correcting it would move every arithmetic result too.
- **The `&` operator is out of reach.** `1e20 & ''` is `100000000000000000000` on AWS and `1e+20` here, because concatenation calls `Functions.string` directly instead of resolving `$string` through the frame.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] Verified against real Step Functions in `us-east-1` with AWS CLI 2.33.7, through `test-state`, which creates nothing
- [x] 22 expressions compared, Floci against AWS. Every digit string here and in the tests is a captured AWS response, not a value read off the documentation

## Checklist

- [x] Targeted Step Functions suites green: `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.*Test'` is 418 tests, 0 failures, 0 errors, up from 411
- [x] 6 new unit tests in `JsonataEvaluatorTest` and 1 in `StepFunctionsJsonataIntegrationTest`, which runs the expression through `CreateStateMachine` and `StartExecution`
- [x] Red proved before the commit. With `JsonataEvaluator` reverted to `main` and the tests untouched: `Tests run: 105, Failures: 6`, every one exponent notation where AWS writes digits, from `expected: <10000000000000000000> but was: <1e+19>` to `expected: <{"n":100000000000000000000}> but was: <{"n":1e+20}>`. `string_switchesToExponentNotationAt1e21` passes on `main`, so it is a regression pin rather than a claim of a fix
- [x] `make docs-check` passes and the docs suite is 34 passed; the generated action table is untouched
- [x] Based on current `origin/main`

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -90,6 +90,10 @@ Three behaviours are worth knowing before reading an unexpected result, and all 
   object, so the field goes missing instead of the state failing.
 - `$range` collapses a single-element range to the bare number, not a one-element array.
 
+JSONata's own `$string` follows AWS's number notation: a whole number is written out in full below
+`1e21` and in exponent notation from there, on both signs, so `$string(1e20)` is
+`100000000000000000000` and `$string(1e21)` is `1e+21`.
+
 One deviation. AWS bounds the memory an expression may use and fails a `$range` of roughly half a
 million elements with `Expression evaluation memory limit exceeded`; Floci has no such bound and
 builds the array.
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluator.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluator.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.services.stepfunctions;
 
+import com.dashjoin.jsonata.Functions;
 import com.dashjoin.jsonata.JException;
 import com.dashjoin.jsonata.Jsonata;
 import com.fasterxml.jackson.core.JsonParser;
@@ -14,6 +15,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 
+import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
@@ -52,6 +54,13 @@ public class JsonataEvaluator {
      * Past it AWS switches from an integer to exponent notation, and so does {@link #toJsonataValue}.
      */
     private static final double LARGEST_EXACT_LONG_AS_DOUBLE = 9.223372036854776E18;
+
+    /**
+     * The magnitude at which AWS stops writing a whole number in full and switches to exponent
+     * notation, on both signs: {@code $string(1e20)} is {@code "100000000000000000000"} and
+     * {@code $string(1e21)} is {@code "1e+21"}.
+     */
+    private static final double SMALLEST_EXPONENT_NOTATION = 1e21;
 
     private final ObjectMapper objectMapper;
     private final ObjectReader strictJsonReader;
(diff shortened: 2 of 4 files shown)


pgermosen (approved):
I didn't just take the live-tested evidence at face value here — I checked JSONata's own public documentation, which states $string converts numbers via JSON.stringify. That's well-known, independently-verifiable JavaScript behavior with exactly the claimed 1e21 cutoff (JSON.stringify(1e20) writes the full digits, JSON.stringify(1e21) switches to exponent notation), which corroborates the central claim here from an entirely separate authoritative source, not just the test-state probing.

I also traced the implementation directly rather than just running the tests in my head. Converting a whole-valued double under the threshold to a BigInteger via BigDecimal.valueOf is genuinely what produces the "shortest round-trip digits, not the exact binary value" behavior described — BigDecimal.valueOf(double) goes through Double.toString, which is exactly why a double holding 2^63 prints as 9223372036854776000 rather than its exact value. The boundary check is correctly strict (less-than, not less-than-or-equal), so 1e21 itself still falls through to the library's own exponent formatting, matching the test. Recursion into nested objects and arrays is consistent with the existing internal representation from the earlier PR, and keeping the library's own signature for $string rather than applying the surplus-argument trick used for the six AWS-specific extensions is the right call — this one's a genuine JSONata builtin with no AWS-specific arity gap to work around, and the javadoc says so rather than leaving the asymmetry unexplained.

PR #2684, opened 2026-08-28
fix(stepfunctions): keep a JSONata expression's explicit null in the output

### LINKED ISSUE #2664: [BUG] Step Functions (SFN): a JSONata expression evaluating to `null` drops the key instead of keeping it
## Service

## AWS API Action

`StartExecution`, `TestState`

## Expected behavior

An expression that evaluates to JSON `null` keeps its key in the output. Real AWS returns `{"v":null}`.

## Actual behavior

Floci drops the key, exactly as it does for an expression that returns nothing. A legitimate `null` and a missing value become indistinguishable in the output, and also inside the evaluator, which is what blocks any correct handling of the undefined case.

## Reproduction

```bash

SM=$(aws stepfunctions create-state-machine --name nul1 --role-arn arn:aws:iam::000000000000:role/r \
    "Output":{"v":"{% $lookup($states.input, \"bar\") %}"},"End":true}}}' \

EXEC=$(aws stepfunctions start-execution --state-machine-arn "$SM" --input '{"bar": null}' \

# SUCCEEDED	{}
```

Real AWS, same expression and same input:

```bash
# instead of reaching AWS.
  aws stepfunctions test-state --region us-east-1 --role-arn "$ROLE_ARN" --input '{"bar": null}' \
  --definition '{"Type":"Pass","QueryLanguage":"JSONata",
                 "Output":{"v":"{% $lookup($states.input, \"bar\") %}"},"End":true}' \
  --query '[status,output]' --output text
# SUCCEEDED	{"v":null}
```

The reference engines disagree with each other on the same expression, which locates the cause:

```
$lookup($, "bar")  on  {"bar": null}
jsonata-js  null
dashjoin    returns nothing
```

## Suggested fix

dashjoin can tell the two apart, but only when asked. `outputConvertNulls` defaults to `true`, which flattens the `null` sentinel into a Java `null`, and Floci never changes it:

```java
// default:                        null -> Java null,          $x.missing -> Java null
// setOutputConvertNulls(false):   null -> Jsonata.NULL_VALUE,  $x.missing -> Java null
```

One call in `JsonataEvaluator` restores the distinction, and `toJsonNode` can then map `NULL_VALUE` to a JSON null and a Java `null` to "returned nothing".

I added that one line to `evaluate` on `main` and ran the Step Functions suite: of its 360 tests, exactly one turns red, `AslExecutorResultWriterTest.jsonataDestinationNullsAreTypeErrorsRatherThanMissingValues`, and the other 359 stay green.

That test name encodes a decision, not an accident, so it is worth a maintainer's read before flipping: I could not tell whether it reflects real AWS behavior on the distributed-`Map` `ResultWriter` S3 export path, where a written `null` and an omitted key are not the same thing downstream.

## Environment

- Java SDK version: AWS CLI 2.33.7, `jsonata-js` v2.2.2 and `com.dashjoin:jsonata:0.9.10` as references
- AWS side: `test-state`, `us-east-1`

facuparedes:
## Summary

A JSONata expression that evaluates to JSON `null` lost its key, exactly as an expression that returned nothing did. The two were indistinguishable in the output and inside the evaluator. Closes #2664, which carries the one-container reproduction.

- **Root cause:** dashjoin flattens JSONata's null marker into a Java `null` unless `outputConvertNulls` is off, and Floci never changed it (`JsonataEvaluator.java:112`). `toJsonNode` then mapped that Java `null` to a `NullNode`, and `resolveTemplate` dropped every `NullNode`.
- **Both halves of the conversion move.** `$states` and the workflow variables bind through the existing `toJsonataValue`, so an input null reaches the expression as a null rather than as undefined. The result is walked back by `fromJsonataValue`, which maps the marker to a JSON null at any depth; a top-level Java `null` becomes a `MissingNode`, now the single representation of "returned nothing".
- **`resolveTemplate` drops only the missing node**, in objects and in arrays. A literal `null` written straight into an `Output` template is kept too, which it was not before.
- **The missing node does not leave the template walk.** `resolveTemplate` is the boundary for `Output`, `Assign` and `Arguments`, and it answers a JSON value: a whole field that is one expression returning nothing resolves to a `null`, as it did before this PR. The private `resolveTemplateNode` is where "returned nothing" is still itself, in the two positions that read it — an object field it drops, an array element it fails the state on. Without that split a top-level `Output` would reach `exec.setOutput(output.toString())` as a `MissingNode`, and `DescribeExecution` would hand back `""` instead of JSON.
- **`AslExecutor` also calls `evaluate` directly, and five of those six sites are unaffected.** A `Choice` condition reads the result through `isBoolean`, `Wait` `Seconds` through `asInt`, `MaxConcurrency` through `isIntegralNumber`, `Map` `Items` through `isArray`; all four answer the same for a `NullNode` and a `MissingNode`, so those paths keep the behaviour they had. The sixth is `Fail`, which reads `Error` and `Cause` through `asText()`: an expression that returned nothing now yields `""` where it yielded `"null"`. AWS answers neither — it fails the state with `States.QueryEvaluationError`, `The JSONata expression '$states.input.absent' specified for the field 'Error' returned nothing (undefined).` — so this is #2665's ground and no test here blesses either string.
- **Docs:** a new `## JSONata nulls` section in `docs/services/step-functions.md`, outside the generated action-table markers, recording both deviations: `$count()` on a JSON null, and the key that Floci drops where AWS fails the state.

```bash
aws stepfunctions test-state --role-arn "$ROLE" --query '[status,output]' --output json \
  --definition '{"Type":"Pass","QueryLanguage":"JSONata","Output":{"v":"{% null %}"},"End":true}' --input '{}'
# [ "SUCCEEDED", "{\"v\":null}" ]
```

| Position, as sent to AWS | AWS returns | Pinned by |
| --- | --- | --- |
| `Output {"v":"{% $lookup($states.input,\"bar\") %}"}` on `{"bar":null}` | `{"v":null}` | `anExplicitNullKeepsItsKeyAndOnlyWhatReturnedNothingIsDropped` |
| `Output {"a":{"b":"{% null %}","c":1}}` | `{"a":{"b":null,"c":1}}` | `anExplicitNullSurvivesInsideANestedObject` |
| `Output "{% null %}"`, the whole field one expression | `null` | `aWholeTemplateIsANullWhetherItEvaluatedToNullOrReturnedNothing` |
| `Output {"a":["{% null %}",1]}` | `{"a":[null,1]}` | `anExplicitNullIsAnArrayElementWhileNothingStillFailsTheState` |
| `Output {"v":null,"k":1}`, a literal null in the template | `{"v":null,"k":1}` | `aLiteralNullInTheTemplateKeepsItsKey` |
| `$exists($states.input.bar)` and `$type($states.input.bar)` on `{"bar":null}` | `true`, `"null"` | `anInputNullIsAValueForExistsAndType` |
| `Assign {"x":"{% null %}"}`, TRACE | variables `{"x":null}` | `anAssignedNullReadsBackAsANullInTheNextState` |
| Task `Arguments {"Payload":{"v":"{% null %}"}}`, TRACE | `afterArguments` `{"Payload":{"v":null}}` | `jsonataArgumentsSendAnExplicitNullAndOmitOnlyWhatReturnedNothing` |

The last two rows are observed differently on each side. `test-state` runs one state and AWS never shows a state its own `Assign`, so the AWS side there is the variables map TRACE reports while the Floci test reads the variable back in the next state; the Task row is `afterArguments` on AWS and the request body an `http:invoke` Task actually sends on Floci.

## Deliberate scope notes

- **#2665 is still open and untouched.** AWS fails the state on an expression that returned nothing: `Output {"v":"{% $states.input.absent %}"}` answers `States.QueryEvaluationError`, `The JSONata expression '$states.input.absent' specified for the field 'Output/v' returned nothing (undefined).` Floci still drops that key, and the new docs section now names that as a deviation rather than leaving it to read as parity. This PR only makes the two cases distinguishable, which is what #2665 needs before it can fail the right one.
- **`$count` on a JSON null answers `0` here and `1` on AWS.** That is dashjoin, not this code: `$count(null)` is `0` straight out of the library with the marker in place. It is one line in the new docs section as a deviation, and no test blesses it.
- **`AslExecutorResultWriterTest.jsonataDestinationNullsAreTypeErrorsRatherThanMissingValues` is unchanged and green.** The issue reports it going red under `setOutputConvertNulls(false)` alone, and that reproduces: 411 tests, 1 failure, at the `Arguments {"Bucket":"{% null %}"}` block. The cause is not a behavioural disagreement. With only that line the marker reaches Jackson unconverted and `valueToTree` throws `No serializer found for class com.dashjoin.jsonata.Jsonata$2`, which `evaluate`'s catch turns into a `States.QueryEvaluationError` carrying that message instead of `ResultWriter Bucket must resolve to a string`. Adding the `toJsonNode` half restores the message: `Bucket` is now a `NullNode` rather than an absent key, and it still is not textual, so the error and its message match `main`. I did not settle what AWS writes on the distributed `Map` `ResultWriter` S3 export path, because this PR does not move it; that stays open for whoever changes that behaviour.
- **Eleven existing tests moved from `assertTrue(result.isNull())` to `assertTrue(result.isMissingNode())`.** `evaluate_returnsNothingForMissingField`, `parse_noArgumentReturnsUndefined`, `partition_emptyArrayReturnsUndefined`, `partition_chunkSizeZeroReturnsUndefined`, `range_missingStepReturnsUndefined`, `range_stepZeroReturnsUndefined`, `range_wrongSignStepReturnsUndefined`, `hash_missingAlgorithmReturnsUndefined`, `hash_missingAlgorithmIsUndefinedButNonStringOneIsASignatureError`, `roundingGoesTowardsZeroNotDownwards` and `string_leavesEveryOtherValueToTheBuiltIn`. Each is named `...ReturnsUndefined` or asserts a missing field, so the assertion now matches the name instead of the representation that made undefined and null the same thing. No test deleted, no expectation weakened.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] Every row above is a captured `test-state` response from `us-east-1`, not a value derived from the documentation. `test-state` creates nothing, and nothing was created on AWS for this PR
- [x] `$exists`, `$type` and `$count` were measured on both sides, AWS through `test-state` and dashjoin 0.9.10 directly
- [x] No wire-protocol or botocore model work: this is expression evaluation

## Checklist

- [x] Targeted Step Functions suites green: 435 tests, 0 failures, 0 errors, 0 skipped (`mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.*Test'`). The merge base is 424 on the same command, and the 11 new tests are the difference
- [x] Red proved before the fix: with the 10 new tests written and `JsonataEvaluator` still at `main`, all 10 fail on the null they expected to survive, for example `expected: <{"fromInput":null,"kept":1}> but was: <{"kept":1}>` and `States.Runtime: The JSONata expression '{% null %}' at array index 0 returned nothing (undefined).`
- [x] 7 new unit tests in `JsonataEvaluatorTest`, 3 new integration tests in `StepFunctionsJsonataIntegrationTest` through `CreateStateMachine` and `StartExecution`, 1 new Task-`Arguments` test in `AslExecutorHttpInvokeTest`
- [x] `aWholeOutputThatReturnedNothingStaysReadableJson` was red before the boundary split: `Output "{% $states.input.absent %}"` came back from `DescribeExecution` as `""`, and it passes with `JsonataEvaluator` reverted to `main`, so the empty output was this PR's own regression and not a pre-existing one
- [x] `make docs-check` passes and the generated action table is untouched. `make docs-test` fails on this machine with `No module named pytest`; through `uvx --with pytest --with pyyaml pytest tools/docs -q` it is 34 passed
- [x] Merge base with `origin/main` is `acf88794`, Conventional Commits
- [ ] The full `./mvnw test` run was not executed on this branch; the diff is one class under `services/stepfunctions`

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -66,6 +66,22 @@ uniformly between zero and the computed delay, as on AWS. One deviation. The del
 between attempts is capped at 30 seconds, the same cap Floci applies to `Wait` states,
 so emulated runs stay fast.
 
+## JSONata nulls
+
+An expression that evaluates to JSON `null` produces a value, not a missing one. It keeps its key
+in `Output`, in `Assign` and in a Task's `Arguments`, at any nesting depth, and it stays in place as
+an array element:
+
+```
+"Output": {"v": "{% $states.input.bar %}"}   on input {"bar": null}   ->   {"v": null}
+```
+
+The same `null` is a value inside the expression too: `$exists()` on it is true and `$type()` on it
+is `"null"`. Only an expression that returns nothing loses its key, which is what
+`$states.input.absent` and the functions listed below that evaluate to undefined do.
+
+One deviation. `$count()` on a JSON null answers `0`; AWS answers `1`.
+
 ## JSONata functions
 
 State machines with `"QueryLanguage": "JSONata"` reach the six functions Step Functions adds on
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -80,7 +80,10 @@ The same `null` is a value inside the expression too: `$exists()` on it is true 
 is `"null"`. Only an expression that returns nothing loses its key, which is what
 `$states.input.absent` and the functions listed below that evaluate to undefined do.
 
-One deviation. `$count()` on a JSON null answers `0`; AWS answers `1`.
+Two deviations. `$count()` on a JSON null answers `0`; AWS answers `1`. And dropping the key of an
+expression that returned nothing is Floci's own behaviour: AWS fails the state instead, with
+`States.QueryEvaluationError` and `The JSONata expression '$states.input.absent' specified for the
+field 'Output/v' returned nothing (undefined).`
 
 ## JSONata functions
 
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluator.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluator.java
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.MissingNode;
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import jakarta.enterprise.context.ApplicationScoped;
@@ -125,6 +126,11 @@ public class JsonataEvaluator {
         String expr = isExpression(expression) ? unwrap(expression) : expression;
         try {
             Jsonata jsonataExpr = jsonata(expr);
+            // Off, the library keeps JSONata's null marker in the result instead of flattening it
+            // into a Java null. On, an expression that evaluated to JSON null and one that returned
+            // nothing both arrive here as a Java null, and AWS keeps the first while dropping the
+            // second.
+            jsonataExpr.setOutputConvertNulls(false);
             Jsonata.Frame frame = jsonataExpr.createFrame();
             stepFunctionsExtensions.forEach(frame::bind);
             // Workflow variables (the Assign field) are referenced as top-level $name in AWS's
@@ -220,7 +226,7 @@ public class JsonataEvaluator {
     }
 
     JsonNode resolveTemplate(JsonNode template, JsonNode statesVar, JsonNode variables) {
-        if (template == null || template.isNull() || template.isMissingNode()) {
+        if (template == null || template.isMissingNode()) {
             return template;
         }
         if (template.isTextual()) {
@@ -236,9 +242,10 @@ public class JsonataEvaluator {
             while (fields.hasNext()) {
                 Map.Entry<String, JsonNode> entry = fields.next();
                 JsonNode value = resolveTemplate(entry.getValue(), statesVar, variables);
-                // Per JSONata spec: undefined (null) values are omitted from object output,
-                // matching real AWS Step Functions behavior.
-                if (value != null && !value.isNull() && !value.isMissingNode()) {
+                // Per JSONata spec: an expression that returned nothing is omitted from object
+                // output, matching real AWS Step Functions behavior. A JSON null is a value and
+                // keeps its key: AWS answers {"v":null} to Output {"v":"{% null %}"}.
+                if (value != null && !value.isMissingNode()) {
                     resolved.set(entry.getKey(), value);
                 }
             }
@@ -249,9 +256,10 @@ public class JsonataEvaluator {
             for (int i = 0; i < template.size(); i++) {
                 JsonNode element = template.get(i);
                 JsonNode value = resolveTemplate(element, statesVar, variables);
-                // Per real AWS behavior: undefined array elements fail the execution.
+                // Per real AWS behavior: an array element that returned nothing fails the execution.
                 // Unlike object fields (which are omitted), undefined in an array is a runtime error.
-                if (value == null || value.isNull() || value.isMissingNode()) {
+                // A JSON null is a value and stays: AWS answers [null,1] to ["{% null %}",1].
+                if (value == null || value.isMissingNode()) {
                     String expr = element.isTextual() ? element.asText() : element.toString();
                     throw new FailStateException("States.Runtime",
                             "The JSONata expression '" + expr + "' at array index " + i + " returned nothing (undefined).");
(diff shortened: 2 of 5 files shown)


pgermosen (approved):
Nice, careful fix. Turning off the library's null-conversion flag and threading its own null sentinel through to a real `NullNode` while a genuine "nothing found" becomes `MissingNode` internally is the right way to recover the distinction AWS actually makes between an explicit JSON null and an expression that returned nothing — object fields dropping the latter but keeping the former, array elements failing the state on the latter, all lines up with how Step Functions really behaves. The public boundary correctly converts a top-level "returned nothing" result back to a real null before it's handed back to callers, so a whole-field Output that resolves to nothing still comes back as valid, parseable JSON instead of an empty string.

The test coverage is thorough on both sides: the renamed assertions are each individually justified as cases where the value genuinely never existed, and the new unit, integration, and HTTP-invoke tests cover explicit null in every position it can appear plus the exact regression this fix addresses. Good call disclosing the two remaining AWS deviations (the count-on-null quirk and the drop-vs-fail behavior tracked separately) in the docs rather than sweeping them in silently.

Issue #2663, opened 2026-08-28
[FEAT] Step Functions (SFN): seven `Task` resources are not in the dispatch chain although Floci already serves the API


facuparedes:
## Service

Step Functions (`Task` service integrations)

## API Action / Feature

Seven `Task` service integrations, reached through `StartExecution`.

## AWS Documentation

https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html

## Expected behavior

AWS runs these seven `Task` resources:

```
arn:aws:states:::events:putEvents
arn:aws:states:::aws-sdk:sfn:startSyncExecution
arn:aws:states:::aws-sdk:sfn:startExecution
arn:aws:states:::aws-sdk:sfn:sendTaskSuccess
arn:aws:states:::aws-sdk:sfn:sendTaskFailure
arn:aws:states:::aws-sdk:scheduler:createSchedule
arn:aws:states:::aws-sdk:scheduler:updateSchedule
```

## Actual behavior

Every one falls through the resource dispatch chain in `AslExecutor.invokeResource` (`main`, line 667) and fails on the `throw` that closes it at line 812:

```
FAILED  States.TaskFailed  Unsupported resource: <the ARN>
```

Floci already implements all seven APIs. Only the ASL branch is missing, so each fix is a mapping from `Arguments` onto a client that already works.

## Reproduction

```bash
docker run -d --name floci -p 4566:4566 floci/floci:nightly
export AWS_REGION=us-east-1 AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test
export AWS_ENDPOINT_URL=http://localhost:4566

TARGET=$(aws stepfunctions create-state-machine --name target --type EXPRESS \
  --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"StartAt":"P","States":{"P":{"Type":"Pass","End":true}}}' \
  --query stateMachineArn --output text)

run() {
  SM=$(aws stepfunctions create-state-machine --name "$1" --role-arn arn:aws:iam::000000000000:role/r \
    --definition "$2" --query stateMachineArn --output text)
  EX=$(aws stepfunctions start-execution --state-machine-arn "$SM" --input '{}' \
    --query executionArn --output text)
  sleep 2
  printf '%-14s %s\n' "$1" "$(aws stepfunctions describe-execution --execution-arn "$EX" \
    --query '[status,error,cause]' --output text)"
}

[...]

PR #2686, opened 2026-08-28
feat(stepfunctions): dispatch the seven Task service integrations of #2663

facuparedes:
## Summary

Seven `Task` resources fell through the dispatch chain in `AslExecutor.invokeResource` and failed with `Unsupported resource`, although Floci already serves all seven APIs. #2663 has the one-container reproduction and the counts from our own fleet. This adds the six that were missing plus `aws-sdk:sfn:startSyncExecution`, which is the first commit on the branch. Every branch maps the `Task` `Arguments` onto the client the direct call already goes through, so there is no second implementation of any of these APIs.

- **One branch per service family, not per ARN.** `aws-sdk:sfn:*` and `aws-sdk:scheduler:*` each enter through a single prefix test that mirrors the `aws-sdk:dynamodb:` and `aws-sdk:cloudformation:` branches already there, then switch on the action. The dispatch chain grows by three tests rather than by seven.
- **`putEvents` goes through `EventBridgeHandler.handle("PutEvents", …)`**, the same entry point `X-Amz-Target` dispatches to. The task result is literally the bytes the direct API returns, which is what `taskResultCarriesTheSameFieldsThePutEventsApiReturns` asserts by comparing the two side by side in one test.
- **The scheduler pair reuses `SchedulerController.parseScheduleRequest`**, widened from `private` to `public`. The `Task` `Arguments` are the CreateSchedule body with `Name` folded in, so restating the mapping would have been a second parse of the same shape.
- **`SendTaskSuccess` and `SendTaskFailure` now return whether the token resolved a waiting task.** That boolean is what lets the integration fail with `Sfn.InvalidTokenException` instead of reporting a delivery that never happened. The direct API keeps tolerating an unknown token, unchanged.
- **Two measured conventions of `aws-sdk:` results now hold for the whole family**, and they correct the `startSyncExecution` envelope from the first commit: a timestamp is ISO-8601, not the epoch seconds of the wire response, and an error names the SDK exception class, so it always ends in `Exception`.
- **A malformed `EcsParameters` list is refused with the name AWS actually sends.** `mapList` and `stringList` now check the shape before the conversion and raise `SerializationException`, so the Task fails as `Scheduler.SerializationException` and `POST /schedules/{name}` answers 400 instead of a server error. Measured, not assumed: AWS answers every one of the six list fields the same way.

## Wire-fidelity details (real Step Functions in `us-east-1`, AWS CLI 2.33.7, through `test-state`)

`test-state` runs the state without creating an execution, so each row is one call. It does need a role that can reach the target service; I created one, measured, and deleted it along with the two state machines and the schedule.

```bash
aws stepfunctions test-state --role-arn "$ROLE" --input '{}' --definition '{
  "Type":"Task","QueryLanguage":"JSONata","Resource":"arn:aws:states:::aws-sdk:sfn:startExecution",
  "Arguments":{"StateMachineArn":"'"$TARGET"'","Input":"{}"},"End":true}'
# {"output": "{\"ExecutionArn\":\"…:execution:floci-2663-target:c18299dc-…\",
#              \"StartDate\":\"2026-08-28T20:34:59.712Z\"}", "status": "SUCCEEDED"}
```

| Case | AWS returns | Pinned by |
| --- | --- | --- |
| `events:putEvents` | `{"Entries":[{"EventId":"920e0a82-…"}],"FailedEntryCount":0}` | `publishedEventReachesTheRuleTargetAndTheResultCarriesItsEventId` |
| `events:putEvents`, one entry rejected | error `EventBridge.FailedEntry`, cause `{"Entries":[{"EventId":"08cbdc46-…"},{"ErrorCode":"MalformedDetail",…}],"FailedEntryCount":1}` | `oneRejectedEntryFailsTheTaskWithFailedEntryAndTheWholeResponseAsCause` |
| `aws-sdk:sfn:startExecution` | `{"ExecutionArn":"…","StartDate":"2026-08-28T20:34:59.712Z"}`, two fields | `startExecutionReturnsThePascalCaseArnAndAnIsoStartDate` |
| the same, with a `Name` already used | `Sfn.ExecutionAlreadyExistsException` | `startExecutionReusingAnExecutionNameFailsWithTheSdkExceptionName` |
| the same, state machine missing | `Sfn.StateMachineDoesNotExistException` | `startExecutionOnAMissingStateMachineFailsWithTheSdkExceptionName` |
| `aws-sdk:sfn:sendTaskSuccess`, token unknown | `Sfn.InvalidTokenException`, cause `Invalid Token: 'Invalid token' (Service: Sfn, Status Code: 400, Request ID: …)` | `sendTaskSuccessOnATokenNobodyIsWaitingForFailsTheTask` |
| `aws-sdk:sfn:sendTaskFailure`, token unknown | same error | `sendTaskFailureOnATokenNobodyIsWaitingForFailsTheTask` |
| `aws-sdk:scheduler:createSchedule` | `{"ScheduleArn":"arn:aws:scheduler:us-east-1:…:schedule/default/floci-2663-s1"}`, one field | `createScheduleReturnsTheScheduleArnAndTheScheduleIsReadable` |
| `aws-sdk:scheduler:updateSchedule` | the same `ScheduleArn` | `updateScheduleKeepsTheArnAndChangesTheExpression` |
| `aws-sdk:scheduler:createSchedule`, name taken | `Scheduler.ConflictException` | `createScheduleOnANameAlreadyTakenFailsWithConflict` |
| `aws-sdk:scheduler:updateSchedule`, schedule missing | `Scheduler.ResourceNotFoundException` | `updateScheduleOnAScheduleThatDoesNotExistFailsWithResourceNotFound` |
| `aws-sdk:scheduler:createSchedule`, `EventBridgeParameters` without `Source` | `Scheduler.ValidationException`, cause `1 validation error detected: Value null at 'target.eventBridgeParameters.source' …` | `createScheduleWithIncompleteEventBridgeParametersIsCatchableAsAValidationException` |
| `aws-sdk:sfn:startSyncExecution` | `"StartDate":"2026-08-28T20:36:17.454Z"` | `succeedingChildReturnsPascalCaseEnvelopeWithOutputAsJsonString` |
| the same, Standard child | `Sfn.StateMachineTypeNotSupportedException` | `standardChildFailsTheTaskWithTheServicePrefixedError` |

The `EcsParameters` list shapes are the one family the CLI refuses client-side, so those requests were signed directly against `scheduler.us-east-1.amazonaws.com`. All ten answer 400 with `x-amzn-errortype: SerializationException`, never `ValidationException`:

| Body | Message |
| --- | --- |
| `CapacityProviderStrategy`, `PlacementConstraints`, `PlacementStrategy`, `Tags`, `Subnets` or `SecurityGroups` holding a scalar instead of a list | `Expected list or null` |
| `CapacityProviderStrategy: ["FARGATE"]` | `Expected null` |
| `Subnets: [123]` | `NUMBER_VALUE can not be converted to a String` |
| `Subnets: [true]` | `TRUE_VALUE can not be converted to a String` |
| `Subnets: [{"a":1}]` | `Start of structure or map found where not expected.` |

Control: `Subnets: ["subnet-0abc"]` gets past serialization and reaches a `ValidationException` about the request itself, so the type check is what separates the two. No schedule was created by any of the malformed probes.

Three things the tables do not show:

- **`aws-sdk:sfn:startExecution` is not `states:startExecution`.** Same API, two integrations, and only the casing of the result separates them: `{ExecutionArn, StartDate}` against `{executionArn, startDate}`. Neither waits. The measured `StartDate` is an ISO-8601 string because the SDK marshals an `Instant` that way, while Floci's `StartExecution` wire response correctly carries `1.787949115835E9`. Both are right, for different envelopes.
- **The token round trip is real.** `sendTaskSuccessResolvesTheWaitingExecutionWithItsOutput` parks an execution on `sqs:sendMessage.waitForTaskToken`, reads the token off the queue, feeds it to a second state machine whose `Task` is `aws-sdk:sfn:sendTaskSuccess`, and asserts the first execution reached `SUCCEEDED` carrying the output the second one sent. `sendTaskFailure` gets the same treatment against `error` and `cause`.
- **The published event is actually delivered.** The `putEvents` test routes a source to an SQS queue through a rule and target, then asserts the `EventId` in the task result is the `id` of the message that arrived.

## Deliberate scope notes

- **The error message stays Floci's, only the error name is AWS's.** AWS answers a missing state machine with `State Machine Does Not Exist: '<arn>'` and Floci's own API says `State machine does not exist`. `Retry` and `Catch` match on the name, so that is what `sdkExceptionName` reproduces; adopting AWS's message text would mean a second copy of every service's message living in `AslExecutor`. The tests assert the name, and the cause only where it is the whole signal (`Invalid Token: 'Invalid token'`, without the `(Service: …, Request ID: …)` tail the SDK appends and Floci has no source for).
- **`putEvents` key order follows Floci's own PutEvents response**, `FailedEntryCount` before `Entries`, where AWS emits them the other way round. Reusing the handler is what makes the two results identical, which is worth more than reordering two keys, but the order is visible in the `EventBridge.FailedEntry` cause string, so it is worth stating.
- **A missing required argument fails at runtime here, at validation time on AWS.** AWS rejects `Arguments` without `StateMachineArn` when the definition is validated, with `SCHEMA_VALIDATION_FAILED: The field 'StateMachineArn' is required but was missing at /States/StateName/Arguments`. Floci does not schema-validate definitions, so the state fails with `Sfn.InvalidArnException` instead; per-integration argument schemas are a different feature.
- **`startSyncExecution` still omits `BillingDetails`, `InputDetails` and `OutputDetails`.** AWS returns all three. `BillingDetails` is metering Floci has no source for, and shipping two of the three would be arbitrary. Recorded here rather than half-implemented.
- **Floci rejects an entry addressed to a bus that does not exist; AWS accepts it** and returns an `EventId`. That divergence belongs to `EventBridgeService.putEvents`, not to this integration, and it is what makes a non-zero `FailedEntryCount` reachable in a test at all. Documented as a deviation in `docs/services/step-functions.md`, unfixed here.
- **An explicit `null` list stays accepted and dropped**, as on `main`. AWS's serializer accepts it — the message is `Expected list or null` — and then rejects `awsvpcConfiguration.Subnets: null` at validation with `Member must not be null`. Reproducing that needs the Scheduler request validator, not this parser, so the guards keep the file's own `has(f) && !get(f).isNull()` idiom and the gap is unchanged by this PR.
- **Thirteen `AslExecutor*Test` files gain three constructor arguments each**, because the constructor takes three new dependencies and those tests build it positionally. Noise in the diff, but field injection would have broken the constructor-injection rule in `AGENTS.md` to buy a smaller one.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] Every response shape, error name and error message in this PR and in the tests is a captured `test-state` or signed-request response from `us-east-1`, not a value read off the documentation
- [x] Each of the seven wraps an API Floci already serves, called directly against `floci/floci:nightly` first, so the task result and the wire response are known to agree
- [x] All AWS resources created for the measurements were deleted: the probe role and its inline policy, two state machines and one schedule
- [x] No wire protocol changed. `SendTaskSuccess` and `SendTaskFailure` return a boolean to their one internal caller; their HTTP responses are untouched

## Checklist

- [x] Step Functions and Scheduler suites green together: `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.*Test,io.github.hectorvent.floci.services.scheduler.*Test'` is 583 tests, 0 failures, 0 errors, 0 skipped
- [x] New tests: `StepFunctionsAwsSdkTaskIntegrationTest` (14 plus a setup method), `StepFunctionsPutEventsIntegrationTest` (4 plus setup) and `StepFunctionsStartSyncExecutionIntegrationTest` (4 plus setup), all through `CreateStateMachine` and `StartExecution`, plus 3 cases in `SchedulerIntegrationTest` covering the malformed `EcsParameters` shapes through the direct REST entry point
- [x] Red proved with `src/main` reverted to `origin/main` and the new tests untouched: 25 failures — 14 of 15 in `StepFunctionsAwsSdkTaskIntegrationTest`, 4 of 5 in `StepFunctionsPutEventsIntegrationTest`, 4 of 5 in `StepFunctionsStartSyncExecutionIntegrationTest` (the remaining one in each is the setup method) and the 3 new `SchedulerIntegrationTest` cases. Every dispatch case fails on `Unsupported resource: <the ARN>`, for all seven ARNs
- [x] `make docs-check` passes; the generated action table is untouched, the new prose sits outside the markers
- [x] `make docs-test` is 34 passed, run as `uvx --with pytest --with pyyaml pytest tools/docs -q` because this machine's python has no pytest
- [x] Branch is `origin/main` plus nine commits: the two `feat:` commits that add the dispatch, then `79b1aab0`, which moves `parseScheduleRequest` inside the Scheduler exception translation so a `Target.EventBridgeParameters` the parser rejects fails the state with `Scheduler.ValidationException` and reaches `Retry` and `Catch` instead of terminating the execution with `States.Runtime`
- [x] Four commits close the malformed-list hole in `SchedulerController` rather than in the integration, and they moved from `ValidationException` to the `SerializationException` AWS actually sends once it was measured: `6cbda1da` and `c5845519` name the conversion failure for `EcsParameters` and `awsvpcConfiguration`, `37bd84f6` checks a scalar element before Jackson coerces it to a string, and `310880cc` moves the shape check inside `mapList` and `stringList` so all six list fields are covered rather than two. On `main`, `POST /schedules/{name}` answered `java.lang.IllegalArgumentException: Cannot construct instance of java.util.LinkedHashMap ... from String value ('FARGATE')` with a server error, and the Task integration turned it into `States.Runtime`
- [x] `e89e1838` passes the three new dependencies to the executor under test, and `0f71d51c` pins the four malformed shapes as catchable `Scheduler.SerializationException` through `aws-sdk:scheduler:createSchedule`
- [x] `Ec2ContainerManagerTest` fails on this machine in a full `./mvnw test` run and passes in isolation (40 tests, 0 failures). It touches nothing in `stepfunctions`, `eventbridge` or `scheduler`, and all four CI `Build and Test` shards are green on this branch, so it is a local Docker artifact rather than a failure of this diff


facuparedes:
Confirmed with a receipt, and the error name in the suggestion is not the one AWS uses.

**The defect is real.** Red first, on the branch before the fix:

```
SchedulerIntegrationTest.createScheduleWithAScalarSubnetReturnsSerializationException
Expected status code <400> but was <200>.   (Subnets: [123])
Expected status code <400> but was <200>.   (Subnets: [true])
```

Jackson writes the number out as a string, so the schedule was stored with `"123"` as a subnet.

**AWS answers `SerializationException`, not `ValidationException`.** The CLI refuses a non-string client-side, so I signed the request directly against `scheduler.us-east-1.amazonaws.com` to reach the API:

| Body | Status | `x-amzn-errortype` | Message |
| --- | --- | --- | --- |
| `Subnets: [123]` | 400 | `SerializationException` | `NUMBER_VALUE can not be converted to a String` |
| `Subnets: [true]` | 400 | `SerializationException` | `TRUE_VALUE can not be converted to a String` |
| `Subnets: [{"a":1}]` | 400 | `SerializationException` | `Start of structure or map found where not expected.` |
| `Subnets: "subnet-1"` | 400 | `SerializationException` | `Expected list or null` |
| `PlacementConstraints: ["x"]` | 400 | `SerializationException` | `Expected null` |
| `Tags: ["x"]` | 400 | `SerializationException` | `Expected map or null` |

Control: `Subnets: ["subnet-0abc"]` reaches a real `ValidationException` about the execution role, so the type check passes and the request goes further. No schedule was created by any probe.

That makes the earlier `mapList` commit on this branch wrong too, in the same way. The name is load-bearing: a `Catch` on `Scheduler.ValidationException` would never match what AWS actually sends.

Fixed in `37bd84f6`: both lists answer `SerializationException`, and `stringList` checks the element type before the conversion runs. The message still names the field rather than reproducing AWS's per-shape wording, which is a deliberate deviation: the name and the status are what a state machine branches on, and the six shapes above would need a token-name table to reproduce exactly.

`SchedulerIntegrationTest`, `SchedulerServiceTest` and `StepFunctionsAwsSdkTaskIntegrationTest` together: 111 tests, 0 failures. Targeted Step Functions suite: 448 tests, 0 failures.

facuparedes:
The two `AslExecutor` findings and the third `SchedulerController` one are all closed in `310880cc` and `0f71d51c`, and the third one generalised past what the suggestion covered.

**The non-array case was real on all six list fields, not two.** The `isArray()` guard silently dropped `CapacityProviderStrategy`, `PlacementConstraints`, `PlacementStrategy`, `Tags`, `Subnets` and `SecurityGroups` alike. Measured against the Scheduler API, every one of them answers the same:

| Body | Status | `x-amzn-errortype` | Message |
| --- | --- | --- | --- |
| `Subnets: "subnet-1"` | 400 | `SerializationException` | `Expected list or null` |
| `Subnets: 1` | 400 | `SerializationException` | `Expected list or null` |
| `Subnets: {"a":1}` | 400 | `SerializationException` | `Start of structure or map found where not expected.` |
| `SecurityGroups: "sg-1"` | 400 | `SerializationException` | `Expected list or null` |
| `CapacityProviderStrategy: "FARGATE"` | 400 | `SerializationException` | `Expected list or null` |
| `PlacementConstraints: "x"` | 400 | `SerializationException` | `Expected list or null` |
| `PlacementStrategy: "x"` | 400 | `SerializationException` | `Expected list or null` |
| `Tags: "x"` | 400 | `SerializationException` | `Expected list or null` |

So the shape check belongs in `mapList` and `stringList`, where the contract already lives, rather than at six call sites. The guards drop back to the file's own `has(f) && !get(f).isNull()` idiom, which is what `Expected list or null` says AWS accepts. Red first: `createScheduleWithANonListEcsParameterReturnsSerializationException` failed `Expected status code <400> but was <200>` on all six before the change.

**The `States.Runtime` half does not reproduce.** Both `convertValue` calls already sit inside a `catch (IllegalArgumentException)` that raises `AwsException`, and after these commits the malformed shape is refused before the conversion runs at all, so nothing reaches `invokeScheduler` as an `IllegalArgumentException`. Pinned rather than argued: `createScheduleWithAMalformedListIsCatchableAsASerializationException` runs four shapes through `arn:aws:states:::aws-sdk:scheduler:createSchedule` with `Catch: ["Scheduler.SerializationException"]` and asserts the caught error, so a regression to `States.Runtime` would fail the state instead of reaching `Recovered`.

Step Functions and Scheduler suites together: 583 tests, 0 failures, 0 errors, 0 skipped.

facuparedes:
One correction to my earlier comment. I wrote that `Expected list or null` means AWS accepts `null` for these fields. That is only half true, and the half I stated is the wrong one.

The serializer accepts `null`, then the request validator rejects it:

```
awsvpcConfiguration.Subnets: null
  -> 400 ValidationException
     1 validation error detected: Value null at
     'target.ecsParameters.networkConfiguration.awsvpcConfiguration.subnets'
     failed to satisfy constraint: Member must not be null
```

Floci drops the field and creates the schedule. That behaviour is unchanged by this PR: `main`'s `isArray()` guard dropped a null identically, so nothing here regresses it. It belongs to the Scheduler request validator rather than to the two list readers this PR touches, and I have recorded it as a scope note in the body rather than widening the diff.

pgermosen (approved):
Went through the dispatch and the scheduler-side fixes directly in the code rather than taking the resolved threads at face value, and it all holds up. The new `aws-sdk:sfn:*`/`aws-sdk:scheduler:*` branches follow the same one-branch-per-family shape as the existing DynamoDB/CloudFormation integrations, and the Scheduler path now correctly wraps the request parse so a malformed `Target` reaches the calling state as a catchable service error instead of terminating the execution outright. The `mapList`/`stringList` fix is a real one, not cosmetic — it closes the gap uniformly across all six ECS list fields, including the scalar-coercion case where a number or boolean was silently getting written into the stored schedule as a string. I also checked that `sendTaskSuccess`/`sendTaskFailure`'s new return value is discarded by the only other caller, so the direct API's existing tolerance for an unknown token is genuinely unaffected.

The scope notes are honest about what's left — the EventBridge key-order mismatch, the runtime-vs-definition-time validation gap, the missing billing/detail fields on the sync execution response, and the null-list behavior are all real, pre-existing, and reasonably left for separate work rather than folded in here.

PR #2687, opened 2026-08-28
fix(stepfunctions): resolve the substitutions in a JSONata error cause

### LINKED ISSUE #2668: [BUG] Step Functions (SFN): a JSONata error cause says `Object` and `{{type}}` instead of the function name and the type

facuparedes:
## Summary

`JsonataEvaluator` built the cause of a `States.QueryEvaluationError` from `e.getMessage()`, dashjoin's already-rendered template. That template is wrong in three separate ways, and #2668 counts what reaches a caller: a function named as the literal `Object`, a sentence ending in `{{type}}` or `{{token}}`, and no error code anywhere. The cause is now composed from `JException`'s own pieces, so it reads the way jsonata-js and AWS write it. Closes #2668.

- **The root cause is dashjoin's catalog, not its call sites.** Diffing `Jsonata.errorCodes` against jsonata-js v2.2.2 shows 21 templates in which the word `function` was replaced by `Object`. Of the 118 malformed causes I measured, 116 come from one of those templates, so restoring that one word is most of the fix.
- **A third placeholder is never reached, and its value is not in the exception either.** `JException.msg` rewrites the first two `{{...}}` of a template into `%1$s` and `%2$s` from `getCurrent()` and `getExpected()`. Exactly two templates have a third: `T0412` (`{{type}}`, 13 cases) and `T2009` (`{{token}}`, 1 case), the remaining 14. `Signature.validate` throws `T0412` with `(arg, param.subtype)` rather than the argument index and function name jsonata-js passes, and the `T2009` site drops the operator, so both sentences now state what is carried and claim nothing else.
- **The code is prefixed and dashjoin's class name is not.** `JSonataException ` is what dashjoin prepends when the code is not in its catalog, which happens for this class's own six functions and for a few library sites that throw with the whole message in the code slot. Values are JSON-rendered too: dashjoin quotes everything it inserts, so `$sqrt(-1)` reported `: "-1"` where AWS reports `: -1`.

`test-state` creates nothing, so every row is re-checkable in one call. AWS prefixes each cause with `The JSONata expression '...' specified for the field '...' threw an error during evaluation.` and the code sits after it; the table drops that prefix, and the next section says why.

```bash
ROLE=arn:aws:iam::<account>:role/<any-role>

aws stepfunctions test-state --role-arn "$ROLE" --query 'cause' --output text \
  --definition '{"Type":"Pass","QueryLanguage":"JSONata","Output":{"v":"{% $sum([\"a\"]) %}"},"End":true}'
# ... threw an error during evaluation. T0412: Argument 1 of function "sum" must be an array of "numbers"
```

| Expression | Floci now | AWS | Pinned by |
| --- | --- | --- | --- |
| `$each('x', function($v){$v})` | `T0410: Argument 1 of function "each" does not match function signature` | identical | `aFunctionNameIsNamedAsAFunctionAndNotAsAnObject` |
| `$single([1,2], function($x){true})` | `D3138: The $single() function expected exactly 1 matching result.  Instead it matched more.` | identical | `aFunctionNameIsNamedAsAFunctionAndNotAsAnObject` |
| `$split('a','b',-1)` | `D3020: Third argument of split function must evaluate to a positive number` | identical | `theCauseCarriesTheJsonataErrorCode` |
| `$sqrt(-1)` | `D3060: The sqrt function cannot be applied to a negative number: -1` | identical | `aValueSubstitutedIntoTheMessageIsRenderedAsJson` |
| `1 + {}` | `T2002: The right side of the "+" operator must evaluate to a number` | identical | `aValueSubstitutedIntoTheMessageIsRenderedAsJson` |
| `$parse(null)` | `T0410: Argument 1 of function "parse" does not match function signature` | identical | `aMessageThrownInPlaceOfACodeIsNotPrefixedWithJSonataException` |
| `$replace('a','','b')` | `Second argument of replace function cannot be an empty string` | prefixes `D3010: ` | `aMessageThrownInPlaceOfACodeIsNotPrefixedWithJSonataException` |
| `$sum(['a'])` | `T0412: Argument ["a"] must be an array of "numbers"` | says `Argument 1 of function "sum"` | `anArrayElementTypeIsNamedInsteadOfLeftAsATypePlaceholder`, `aFailedJsonataExpressionReportsTheErrorCodeAndTheTypeItWanted` |
| `$average(['a','b'])` | `T0412: Argument ["a","b"] must be an array of "numbers"` | says `Argument 1 of function "average"` | `anArrayElementTypeIsNamedInsteadOfLeftAsATypePlaceholder` |
| `1 < 'a'` | `T2009: The values 1 and "a" either side of the operator must be of the same data type` | says `operator "<"` | `bothComparedValuesAreNamedInsteadOfLeftAsATokenPlaceholder` |
| `'a' < 1` | `T2009: The values "a" and 1 either side of the operator must be of the same data type` | says `operator "<"` | `bothComparedValuesAreNamedInsteadOfLeftAsATokenPlaceholder` |

I re-ran the corpus the issue used: 1674 expressions from the jsonata-js v2.2.2 groups minus `function-eval`, each on its dataset, 285 of which raise here. **Causes carrying an unresolved substitution go from 118 to 1, and causes carrying a `T####`/`D####`/`S####` code from 0 to 220.** The one survivor is a raw `NullPointerException` from a path that builds no message from a template (`Cannot invoke "java.util.Map.get(Object)" because "item" is null`); the word `Object` there is a Java signature, not a substitution.

## Deliberate scope notes

- **The sentence AWS puts in front of the cause is #2665, not this PR.** The field path is not available inside `resolveTemplate`, which walks a template without carrying where it is. Threading it here would put the interesting half of #2665 inside a diagnosability fix.
- **The argument index, the function name and the operator are not in the exception, so they are not invented.** Writing `of function "n"` after restoring the word would call a signature type letter a function name, which is worse than the placeholder it replaces. Both messages are shorter than AWS's and every word in them is carried.
- **`T0410`'s argument number is dashjoin's, and it is not always AWS's.** For `$sum([1],2)` AWS says `Argument 2` and the library says `Argument 1`, because `throwValidationError` walks its regex with `matches()` where jsonata-js uses a prefix test. That is a signature-validation defect rather than a substitution one, and fixing it means reimplementing the validator, so the table only claims rows where the two agree.
- **Value rendering came with the placeholders, not instead of them.** Half-repairing `The sqrt Object ... : "-1"` into `The sqrt function ... : "-1"` leaves a cause that still is not the AWS string, for a reason nobody would guess. `json()` is four lines and it is what makes six of the eleven rows above exact.
- **Restoring `function` runs on the template, never on the message.** Rendering once with markers in the value slots keeps a value that contains the word `Object` out of the replacement; replacing on the rendered string would silently corrupt `Syntax error: "Object"`.
- **Behaviour is unchanged.** `error` is still `States.QueryEvaluationError` on every path, including the non-`JException` failures dashjoin lets out of `$toMillis` and `$number`, which keep the message they came with.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] Verified against real Step Functions in `us-east-1` with AWS CLI 2.33.7, through `test-state`, which creates nothing and left nothing behind
- [x] Every "AWS" cell above is a captured response, not a value derived from the documentation or from jsonata-js
- [x] Catalog diff against jsonata-js v2.2.2 `src/jsonata.js`: 21 templates differ only by `function` becoming `Object`, and exactly 2 have a third placeholder

## Checklist

- [x] Targeted Step Functions suites green: `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.*Test'` is 433 tests, 0 failures, 0 errors, 0 skipped (424 before this PR)
- [x] 8 new unit tests in `JsonataEvaluatorTest`, plus 1 integration test in `StepFunctionsJsonataFunctionsIntegrationTest` that reproduces #2668 through `CreateStateMachine`, `StartExecution` and `DescribeExecution`
- [x] Red proof: with `JsonataEvaluator.java` reverted to `main` and the tests untouched, the 9 run as `Tests run: 9, Failures: 7`, for instance `expected: <T0412: Argument ["a"] must be an array of "numbers"> but was: <Argument "[a]" of Object "n" must be an array of {{type}}>`. The two that stay green are the behaviour-preservation pins, `theErrorNameStaysQueryEvaluationErrorSoACatchStillMatchesIt` and `aNonJsonataFailureKeepsTheMessageItCameWith`
- [x] No docs change: nothing under `docs/` documents a JSONata cause, and the generated action table is untouched
- [x] Rebased onto `main` at acf88794, after #2683; the branch merges into current `origin/main` with no conflict

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluator.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluator.java
@@ -62,6 +62,18 @@ public class JsonataEvaluator {
      */
     private static final double SMALLEST_EXPONENT_NOTATION = 1e21;
 
+    /** What dashjoin prefixes to an error whose code is not in its message catalog. */
+    private static final String UNKNOWN_CODE_PREFIX = "JSonataException ";
+
+    /** Stand-ins for the two values dashjoin substitutes; no catalog template contains them. */
+    private static final String CURRENT_MARKER = "{{floci-current}}";
+    private static final String EXPECTED_MARKER = "{{floci-expected}}";
+
+    /** The name jsonata-js gives each element type a function signature can ask an array for. */
+    private static final Map<Object, String> ARRAY_ELEMENT_TYPES = Map.of(
+            "a", "arrays", "b", "booleans", "f", "functions",
+            "n", "numbers", "o", "objects", "s", "strings");
+
     private final ObjectMapper objectMapper;
     private final ObjectReader strictJsonReader;
     private final Map<String, Jsonata.JFunction> stepFunctionsExtensions;
@@ -131,7 +143,68 @@ public class JsonataEvaluator {
             Object result = jsonataExpr.evaluate(null, frame);
             return toJsonNode(result);
         } catch (Exception e) {
-            throw new AslExecutor.FailStateException("States.QueryEvaluationError", e.getMessage());
+            throw new AslExecutor.FailStateException("States.QueryEvaluationError", queryEvaluationCause(e));
+        }
+    }
+
+    /**
+     * The cause AWS reports for a failed JSONata expression: the error code, then the message
+     * jsonata-js renders for it, as in {@code T0412: Argument 1 of function "sum" must be an array
+     * of "numbers"}. AWS puts a sentence naming the expression and the field in front of that,
+     * which needs the field path {@link #resolveTemplate} does not thread today (#2665).
+     *
+     * <p>The dashjoin port renders the message itself, and three things go wrong on the way. Its
+     * copy of the catalog has the word "function" replaced by "Object" throughout; its substituter
+     * fills the first two placeholders of a template and leaves a third one standing; and it quotes
+     * every value it inserts, where jsonata-js JSON-renders it. So the message is composed here
+     * from the code and the values {@link JException} carries instead.
+     */
+    private String queryEvaluationCause(Exception e) {
+        if (!(e instanceof JException jsonataError) || jsonataError.getError() == null) {
+            return e.getMessage();
+        }
+        String code = jsonataError.getError();
+        Object current = jsonataError.getCurrent();
+        Object expected = jsonataError.getExpected();
+        // Both this class's own functions and a few library call sites throw with the whole
+        // message in the code slot. The catalog lookup then misses and dashjoin prefixes its own
+        // class name, which AWS never emits; the message is the code slot itself.
+        if ((UNKNOWN_CODE_PREFIX + code).equals(jsonataError.getMessage())) {
+            return code;
+        }
+        return switch (code) {
+            // The only two templates in the catalog with a third placeholder. Neither third value
+            // reaches the exception: T0412 carries the offending argument and the element type it
+            // wanted but not the argument index nor the function name, and T2009 carries the two
+            // compared values but not the operator. Both sentences state what is carried.
+            case "T0412" -> "T0412: Argument " + json(current) + " must be an array of "
+                    + json(ARRAY_ELEMENT_TYPES.getOrDefault(expected, String.valueOf(expected)));
+            case "T2009" -> "T2009: The values " + json(current) + " and " + json(expected)
+                    + " either side of the operator must be of the same data type";
+            default -> code + ": " + renderTemplate(code, current, expected);
+        };
+    }
+
+    /**
+     * The catalog template for a code with its values substituted. Rendering it once with markers
+     * in the value slots is what keeps the template's own words apart from a value that happens to
+     * contain them, so restoring "function" cannot reach into a value. A marker comes back quoted
+     * where jsonata-js JSON-renders the value and bare where it inserts it raw.
+     */
+    private String renderTemplate(String code, Object current, Object expected) {
+        return JException.msg(code, -1, CURRENT_MARKER, EXPECTED_MARKER)
+                .replace("Object", "function")
+                .replace('"' + CURRENT_MARKER + '"', json(current))
+                .replace('"' + EXPECTED_MARKER + '"', json(expected))
+                .replace(CURRENT_MARKER, String.valueOf(current))
+                .replace(EXPECTED_MARKER, String.valueOf(expected));
+    }
+
+    private String json(Object value) {
+        try {
+            return objectMapper.writeValueAsString(value);
+        } catch (JsonProcessingException e) {
+            return String.valueOf(value);
         }
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluatorTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/stepfunctions/JsonataEvaluatorTest.java
@@ -618,4 +618,117 @@ class JsonataEvaluatorTest {
         assertEquals("[\"100000000000000000000\",\"1e+21\"]",
                 evaluator.evaluate("{% $map([1e20, 1e21], $string) %}", statesVar).toString());
     }
+
+    // ---- the cause of a States.QueryEvaluationError (#2668) ----
+    // Every expected string below was captured from real AWS through test-state, minus the
+    // "The JSONata expression '...' specified for the field '...' threw an error during
+    // evaluation." prefix AWS puts in front of it, which needs the field path #2665 threads.
+
+    @Test
+    void theCauseCarriesTheJsonataErrorCode() {
+        JsonNode statesVar = objectMapper.createObjectNode();
+        AslExecutor.FailStateException ex = assertThrows(AslExecutor.FailStateException.class,
+                () -> evaluator.evaluate("{% $split('a','b',-1) %}", statesVar));
+        assertEquals("D3020: Third argument of split function must evaluate to a positive number",
+                ex.cause);
+    }
+
+    @Test
+    void theErrorNameStaysQueryEvaluationErrorSoACatchStillMatchesIt() {
+        JsonNode statesVar = objectMapper.createObjectNode();
+        for (String expression : new String[]{"$sum(['a'])", "1 < 'a'", "$sqrt(-1)",
+                "$each('x', function($v){$v})", "$parse(null)"}) {
+            AslExecutor.FailStateException ex = assertThrows(AslExecutor.FailStateException.class,
+                    () -> evaluator.evaluate("{% " + expression + " %}", statesVar),
+                    "expected " + expression + " to fail");
+            assertEquals("States.QueryEvaluationError", ex.error,
+                    "wrong error name for " + expression);
+        }
+    }
+
+    @Test
+    void aFunctionNameIsNamedAsAFunctionAndNotAsAnObject() {
+        // dashjoin ships jsonata-js's message catalog with the word "function" replaced by
+        // "Object" throughout, which is 116 of the 117 malformed causes in #2668.
+        JsonNode statesVar = objectMapper.createObjectNode();
+        assertEquals("T0410: Argument 1 of function \"each\" does not match function signature",
+                assertThrows(AslExecutor.FailStateException.class,
+                        () -> evaluator.evaluate("{% $each('x', function($v){$v}) %}", statesVar)).cause);
+        assertEquals("D3050: The second argument of reduce function must be a function"
+                        + " with at least two arguments",
+                assertThrows(AslExecutor.FailStateException.class,
+                        () -> evaluator.evaluate("{% $reduce([1,2], function($a){$a}) %}", statesVar)).cause);
+        assertEquals("D3138: The $single() function expected exactly 1 matching result. "
+                        + " Instead it matched more.",
+                assertThrows(AslExecutor.FailStateException.class,
+                        () -> evaluator.evaluate("{% $single([1,2], function($x){true}) %}", statesVar)).cause);
+    }
+
+    @Test
+    void anArrayElementTypeIsNamedInsteadOfLeftAsATypePlaceholder() {
+        // T0412 is the one code in the catalog carrying both defects: the "Object" word and a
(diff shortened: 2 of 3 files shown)


pgermosen (commented):
This needs a rebase before it can merge — it now conflicts with an already-merged sibling PR that added an unrelated constant at the same spot in JsonataEvaluator.java. The conflict itself looks mechanical (two independent additions landing at the same line, no logical overlap), so it should collapse cleanly once rebased.

On the fix itself: the marker-then-substitute technique is genuinely correct, not just plausible-looking. Rendering the template once with placeholder markers standing in for the real values, doing the global "Object"-to-"function" word fix on that, and only then substituting the real values in, is exactly the right order — doing the value substitution first would risk corrupting any user value that happens to contain the literal word "Object," which the write-up's own example shows is a real, findable failure mode this avoids by construction. I also cross-checked the detection for messages thrown with the whole formatted string in the code slot against the extension functions added earlier, and it correctly identifies that exact pattern and unwraps it rather than letting the generic catalog machinery mishandle it.

One theoretical edge case, not something I'd ask to be fixed: the bare-marker fallback path uses plain string conversion rather than JSON rendering, which could in principle diverge from AWS if some template ever substituted a complex array or object value through that specific path. The corpus run here (1674 real expressions, reducing malformed causes from 118 down to a single unrelated one) is thorough enough that if this path were actually exercised with a complex value, it would very likely have shown up as a remaining mismatch. Also worth noting positively: the comparison table is upfront about one narrow remaining gap (a missing error-code prefix on one specific case) rather than smoothing it over.

facuparedes:
Rebased onto current `main`, which now carries #2683. The conflict was two independent additions at the same line of `JsonataEvaluator.java` (`SMALLEST_EXPONENT_NOTATION` from #2683, the marker constants here) and the same at the tail of `JsonataEvaluatorTest.java`. Both sides kept, no logic touched. Targeted suite on the rebased head: 433 tests, 0 failures, 0 errors, 0 skipped.

On the bare-marker fallback: you are right that it converts with plain string conversion rather than JSON rendering. I went back and checked which templates can reach it, and it is the two three-placeholder codes only, `T0412` and `T2009`, whose leftover slots dashjoin never fills. `T0412` substitutes the offending argument and `T2009` the two compared values, and both of those go through the JSON renderer, not the fallback. The fallback covers the case where a code carries a placeholder no value was captured for at all, where there is nothing to render either way. So a complex value cannot reach it today, which is why the corpus never surfaced it.

facuparedes:
Correcting my previous comment: I named the wrong codes. `T0412` and `T2009` leave through the `switch` at `JsonataEvaluator.java:180` and never reach `renderTemplate`, so they are not the ones on the bare path. Here is the measured answer.

I rendered every code in dashjoin's catalog with markers in both value slots and classified where each marker lands:

| Marker position | Codes |
| --- | --- |
| Quoted, so the JSON renderer fills it | 41 |
| Bare | **2** (`D3137`, `D3141`) |
| Template takes no value | 53 |

For both bare codes the template is nothing but the value: `Jsonata.java:2452` and `:2456` are `put("D3137", "{{{message}}}")` and `put("D3141", "{{{message}}}")`. Their only throw sites are `Functions.error(String message)` (`Functions.java:1881`) and `Functions.assertFn(boolean, String message)` (`:1894`), both typed `String`. So a complex value cannot arrive there, and JSON-rendering that slot would be the bug: it would wrap the message in quotes AWS does not emit.

Your instinct that the corpus would have caught it was right, for a narrower reason than either of us wrote: those two codes are the only way in, and neither can carry an array or an object.

pgermosen (approved):
Rebase checked directly — clean single commit on top of the sibling PRs, no leftover conflict markers, both constant blocks sitting side by side exactly as expected from what was a purely mechanical conflict.

On the earlier theoretical concern about the bare-marker fallback: this is now definitively closed rather than just reassured. I verified the follow-up independently against the actual pinned dashjoin library source rather than taking it on faith — the two codes that can reach that path have their only throw sites typed as a plain String parameter in the library itself, so it isn't just that the corpus happened not to exercise a complex value there, it's structurally impossible for one to ever reach it.

facuparedes:
Independent verification of the review's open point, measured rather than argued.

**The bare-marker path is exercised, and it is exactly AWS's string.** `D3137` and `D3141` are the only two templates that place a value bare (`{{{message}}}`), and both are reachable from a normal expression. Against real AWS through `test-state`:

```
$assert(false, "boom")  ->  ... threw an error during evaluation. D3141: boom
$error("boom")          ->  ... threw an error during evaluation. D3137: boom
```

This branch produces `D3141: boom` and `D3137: boom`. So the path is not only harmless, it is the correct one: JSON-rendering that slot would add quotes AWS does not emit.

**The `Object`-inside-a-value hazard is closed by construction, and I confirmed it end to end.** `$error('an Object walks in')` yields `D3137: an Object walks in` — the value passes through the `"Object"` restoration untouched, because the restoration runs on the template before substitution.

**Catalog claims re-derived from the jar.** Reflecting over `Jsonata.errorCodes` in `jsonata-0.9.10.jar`: 96 codes, 21 templates carrying `Object`, and exactly 2 leaving a third placeholder after the two-slot substituter (`T0412`, `T2009`) — both special-cased here. Marker-slot classification is 41 quoted / 2 bare / 53 valueless, matching the table above.

**Corpus, before and after, on this branch.** Running every error-expecting case of the jsonata-js v2.2.2 groups (minus `function-eval`) straight through `JsonataEvaluator`, 225 reach `FailStateException`:

| | causes with a `T/D/S` code | causes with a leftover `{{..}}` | causes carrying the word `Object` |
| --- | --- | --- | --- |
| `main` | 0 | 13 | 100 |
| this branch | 179 | **0** | **0** |

**Red proof reproduced.** With `JsonataEvaluator.java` reverted to `acf88794` and the tests untouched: `Tests run: 82, Failures: 7`, all seven among the nine added here, including `expected: <T0412: Argument ["a"] must be an array of "numbers"> but was: <Argument "[a]" of Object "n" must be an array of {{type}}>`. The two that stay green are the named behaviour pins.

**Sizing the one gap the description already admits.** 46 of the 225 causes still carry no code. 36 of them are raw Java messages from dashjoin's own date/number formatting (`Malformed pattern "0,"`, `java.text.ParseException`, `Argument error calling function split`) — there is no template and no code to recover. The other 10 are cases where dashjoin threw a fixed catalog message in the code slot, so the code is recoverable by reverse lookup: `D3010`, `D3011`, `D3131`, `D3134`, `D3136` (x4), `S0401`, `S0402`. That is a separate follow-up — it needs a reverse index of the 53 valueless templates, which is machinery this issue does not ask for.

Also refreshed two stale lines in the description: the targeted suite is 433 tests on the rebased head (was written as 420), and the base is `acf88794` after #2683 (was written as `3bd9bcc6` after #2676). The branch merges into current `origin/main` with no conflict, and all 19 checks pass.

PR #2689, opened 2026-08-28
fix(stepfunctions): fail the state when a JSONata expression returns nothing

### LINKED ISSUE #2665: [BUG] Step Functions (SFN): a JSONata expression that returns nothing succeeds instead of failing the state

facuparedes:
#2684 has merged, so this is no longer stacked: the branch is rebased onto `origin/main` and the diff is this PR's change alone, five commits from `1359bb9b` to `b5b3b6f5`.

## Summary

A JSONata expression that returns nothing is not a missing value a state carries on without. Real AWS fails the state with `States.QueryEvaluationError`, names the field the expression was written in, and a `Catch` on that error fires. Floci succeeded and dropped the key, so the execution kept going down the happy path with a field that is not there. Closes #2665, which carries the reproduction and the seven positions.

- **One guard, eleven positions.** `evaluateField` (`JsonataEvaluator.java:152`) evaluates one expression and fails the state when it comes back as nothing. `resolveTemplate` calls it at every leaf, carrying the field it is inside and extending it with `/key` per object key and `[i]` per array index, which covers `Output`, `Assign`, a Task's `Arguments`, `ItemReader/Arguments` and `ResultWriter/Arguments`. The six positions that called `evaluate` directly (`Choices[n]/Condition`, `Seconds`, `Error`, `Cause`, `MaxConcurrency`, `Items`) call it themselves.
- **A matched `Choice` rule and a matching `Catch` clause name their own fields under their index.** `Choices[1]/Output/v`, `Choices[0]/Assign/x`, `Catch[1]/Output/v` and `Catch[0]/Assign/x`, all four measured on the account; the index is the position in `Choices` or `Catch`, not the position among the entries that matched. `applyJsonataAssignAndOutput` had one hardcoded `Output` and `Assign` for all four of its holders. The `Catch` label was unobservable until this: `doExecute` calls `handleCatch` from inside its own `catch (FailStateException)`, so a failure raised while resolving the clause's own `Output` escaped to the method-level handler and came back as `States.Runtime` with the raw message. The execution now fails with the clause's own error, and no later clause catches it, which is what AWS does.
- **This is a behaviour change, and it makes previously green runs red.** A run of the jsonata-js v2.2.2 suite as one `Pass` state per expression found 108 cases the suite marks `undefinedResult`; 106 succeeded on Floci and dropped the key. All 106 now fail the state, as AWS does with each. Any state machine relying on the silent drop stops working, and that is the point: today it runs on Floci and fails on AWS.
- **Three of the seven positions changed control flow, not data.** `Choice` silently took `Default`, `Wait` did not wait, and a `Task` sent its request with the argument omitted. The red proof below catches the SQS Task actually reaching the queue.
- **Docs**: a new `## An expression that returns nothing fails the state` section in `docs/services/step-functions.md`, and the `## JSONata nulls` section from #2684 loses both the sentence saying an undefined expression only loses its key and the deviation #2684 added for it, which this change removes. The `## JSONata functions` section also loses the sentence saying an undefined argument leaves the field missing instead of failing the state, measured on AWS for all three shapes it lists: `$hash("x")`, `$partition([1,2],0)` and `$range(1,5)` each answer `States.QueryEvaluationError` with `... specified for the field 'Output/v' returned nothing (undefined).`

## Wire-fidelity details (AWS CLI 2.33.7, `test-state` and real executions against `us-east-1`)

Every cause below is a captured AWS response, and `test-state` creates nothing, so an unmarked row is re-checkable in one call:

```bash
aws stepfunctions test-state --role-arn "$ANY_ROLE" --input '{}' --query 'cause' --output text \
  --definition '{"Type":"Pass","QueryLanguage":"JSONata","Output":{"a":{"b":[[1,"{% $states.input.missing %}"]]}},"End":true}'
# The JSONata expression '$states.input.missing' specified for the field 'Output/a/b[0][1]' returned nothing (undefined).
```

Each row uses `{% $states.input.missing %}` on input `{}` and each error is `States.QueryEvaluationError`. The middle column is what AWS puts in `... specified for the field '<here>' returned nothing (undefined).`

| Position | AWS names the field | Pinned by |
| --- | --- | --- |
| `Pass` `Output: {"v": ...}` | `Output/v` | `outputExpressionReturningNothingFailsTheStateAndIsCatchable` |
| `Pass` `Output: {"a": {"b": ...}}` | `Output/a/b` | `anExpressionReturningNothingFailsTheStateNamingTheFieldPath` |
| `Pass` `Output: ...` (whole field) | `Output` | same test, and `aWholeOutputThatReturnedNothingFailsTheState` |
| `Pass` `Output: {"values": [..., 1]}` | `Output/values[0]` | same test |
| `Pass` `Output: {"a": {"b": [[1, ...]]}}` | `Output/a/b[0][1]` | same test |
| `Pass` `Assign: {"x": {"y": ...}}` | `Assign/x/y` | same test |
| `Pass` `Assign: {"x": ...}` | `Assign/x` | `assignExpressionReturningNothingFailsTheState` |
| `Task` `sqs:sendMessage` `Arguments` | `Arguments/MessageGroupId` | `taskArgumentReturningNothingFailsTheStateBeforeTheRequestIsSent` |
| `Choice` first rule | `Choices[0]/Condition` | no test: the same `Choices[i]` literal as the row below |
| `Choice` second rule, first one false | `Choices[1]/Condition` | `choiceConditionReturningNothingFailsTheStateInsteadOfTakingDefault` |
| `Choice` matched rule's own `Output` | `Choices[1]/Output/v` | `aMatchedChoiceRuleNamesItsOwnOutputUnderTheRuleIndex` |
| `Choice` matched rule's own `Assign` | `Choices[0]/Assign/x` | `aMatchedChoiceRuleNamesItsOwnAssignUnderTheRuleIndex` |
| `Catch` matching clause's own `Output` (real execution) | `Catch[1]/Output/v` | `aCatchClauseNamesItsOwnOutputUnderTheClauseIndex` |
| `Catch` matching clause's own `Assign` (real execution) | `Catch[0]/Assign/x` | no test: the same `applyJsonataAssignAndOutput` call as the row above |
| `Wait` `Seconds` | `Seconds` | `waitSecondsReturningNothingFailsTheStateInsteadOfNotWaiting` |
| `Fail` `Error` | `Error` | `failErrorAndCauseReturningNothingFailTheStateWithTheQueryEvaluationError` |
| `Fail` `Cause` | `Cause` | same test |
| `Map` `Items` (real execution) | `Items` | `mapItemsReturningNothingFailsTheStateInsteadOfIteratingNothing` |
| `Map` `MaxConcurrency` (real execution) | `MaxConcurrency` | `mapMaxConcurrencyReturningNothingFailsTheStateNamingTheField` |
| DISTRIBUTED `Map` `ItemReader` `Arguments` (real execution) | `ItemReader/Arguments/Key` | no test: needs a distributed run |
| DISTRIBUTED `Map` `ResultWriter` `Arguments` (real execution) | `ResultWriter/Arguments/Prefix` | `aDestinationExpressionReturningNothingFailsBeforeWriting` |

Four things the table cannot show:

- **A `Choice` stops at the first rule that matches.** With `Choices[0]` at `{% true %}` and `Choices[1]` undefined, AWS answers `SUCCEEDED`. `aChoiceRuleAfterTheMatchingOneIsNeverEvaluated` pins it: a guard over every rule up front would break it.
- **No later `Catch` clause catches what an earlier matching one raised.** A state with `Catch[0]` on `States.QueryEvaluationError` whose own `Output` returns nothing and `Catch[1]` on `States.ALL` ends `FAILED` on AWS with `Catch[0]/Output/v`, not caught by the second clause.
- **An explicit `null` still succeeds.** `Output {"v": "{% null %}"}` is `SUCCEEDED` on AWS. That distinction is what #2684 buys, and `anExplicitNullIsAValueAndKeepsEveryPositionEvaluating` keeps it from being undone here.
- **`test-state` does not reach the `Catch` and `Map` rows.** It supports neither `Map` nor `Parallel`, and on a caught error it answers `CAUGHT_ERROR` and simply omits the output when a `Catch` clause's own `Output` returned nothing, rather than failing. Those rows came from `create-state-machine` plus `start-execution`. Every state machine created was deleted.

## Deliberate scope notes

- **I threaded the field path rather than shipping a guard with no field name.** The smaller option fixes the `Catch`, the behavioural half, and the issue offers it as a first step. Threading costs one `String` parameter on `resolveTemplate` and one literal at each call site, and the measured rows agree on one rule: the ASL field relative to the state, `/` per object key, `[i]` per array index. That is an accumulator, not a mechanism, and a nameless cause would have to be replaced later by exactly this change.
- **The `An error occurred while executing the state '<name>' (entered at the event id #<n>).` prefix belongs to #2668.** Floci returns the bare cause, the form `TestState` returns; a real execution adds the prefix. This PR does not touch the cause-building code #2668 owns, and the docs section records the missing prefix as this change's one deviation.
- **An undefined array element stopped being a `States.Runtime` special case.** `resolveTemplate` threw its own error there, with a different code and the wording `at array index 0`. AWS answers `States.QueryEvaluationError` and `Output/values[0]`, the same shape as every other position, so the leaf guard covers it. A second behaviour change, smaller: that state already failed, with the wrong code and message.
- **`resolveTemplateNode` is gone and `resolveTemplate` is one method again.** `36db065e` split it so a whole field returning nothing could be coerced to a `NullNode` before serialization, a missing node writing itself as the empty string no client can parse; its comment says that stands `until #2665`. Once the state fails there is no output to serialize, and the walk's only source of a missing node was the leaf the guard throws at, so coercion and wrapper are both unreachable.
- **Two `ResultWriter` guards were deleted as dead, not as cleanup.** `ResultWriter Bucket must resolve to a string` and its `Prefix` twin fired only on a key `resolveTemplate` had dropped, which can no longer happen. The type checks below them still cover a non-string value, `{% null %}` included, and `jsonataDestinationNullsAreTypeErrorsRatherThanMissingValues` passes unchanged. `ItemReader/Arguments/Key` and `ResultWriter/Arguments/Prefix` were the two labels this PR first shipped extrapolated; both are now measured on a DISTRIBUTED `Map` execution against a bucket that does not exist, since AWS evaluates the arguments before it reaches S3, and both match what the code emits.
- **`MaxConcurrency` keeps its own non-negative-integer check.** The guard removes only the undefined case; `true`, `-1` and `1.5` still reach it, and `invalidJsonataMaxConcurrencyIsCatchableAsQueryEvaluationError` still covers all four.
- **Every state-machine definition in the new tests is one AWS accepts.** All fourteen pass `validate-state-machine-definition` with `result: OK`. `0706fee3` moved the `Catch` clause under test off a `Pass` state, which AWS refuses with `SCHEMA_VALIDATION_FAILED`, `Field 'Catch' is not supported`, onto the `Task` the measurement used.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

The commit is `fix:`, not `fix!:`: the repository has no `!` subject and no `BREAKING CHANGE` footer in its history, and forcing a major bump on a fidelity fix is a maintainer's call. The box is ticked so the decision is visible.

## AWS Compatibility

- [x] Verified in `us-east-1` with AWS CLI 2.33.7: the `Pass`, `Task`, `Choice`, `Wait` and `Fail` rows through `test-state`, the `Catch`, `Map`, `ItemReader` and `ResultWriter` rows through `create-state-machine` plus `start-execution`, every state machine deleted afterwards
- [x] Every field name and cause here and in the tests is a captured AWS response, not a value read off the documentation
- [x] `Catch` on `States.QueryEvaluationError` confirmed to fire on AWS, and pinned here by `outputExpressionReturningNothingFailsTheStateAndIsCatchable`

## Checklist

- [x] Step Functions suite green: `-Dtest='io.github.hectorvent.floci.services.stepfunctions.*Test'` is 468 tests, 0 failures, 0 errors, 0 skipped
- [x] 21 new tests: 12 in `StepFunctionsJsonataIntegrationTest`, 8 in `JsonataEvaluatorTest` (one parameterized over the six measured path shapes), 1 in `AslExecutorResultWriterTest`
- [x] Red proof. With the `isMissingNode` throw removed from `evaluateField` and the drop, the array throw and the top-level null coercion restored in `resolveTemplate`, tests untouched, the same 468-test run is 20 failures, each for the reason claimed: six `nothing was thrown`, eight `Execution SUCCEEDED`, three `but was: <States.Runtime>` (the two array rows and `Map` `Items`), `MaxConcurrency must resolve to a non-negative integer`, `but was: <>` (a `Fail` whose `Error` came back empty) and `but was: <SQS.AWS.SimpleQueueService.NonExistentQueue>`, which is Floci reaching SQS with `MessageGroupId` omitted
- [x] Seven existing tests changed, none deleted. Six because AWS disagrees with what they asserted: the three `...AndDropsOnlyWhatReturnedNothing` / `...AndOmitOnlyWhatReturnedNothing` tests keep the explicit-null half AWS confirms and lose the dropped-key half, which is the bug; `aWholeTemplateIsANullWhetherItEvaluatedToNullOrReturnedNothing` and `aWholeOutputThatReturnedNothingStaysReadableJson` are the stopgap `36db065e` marked `until #2665`; in `AslExecutorResultWriterTest` the `Prefix` block moved into its own test. The seventh is the crashing evaluator stub in `AslExecutorErrorTerminationTest`, which follows the new `resolveTemplate` signature
- [x] `make docs-check` passes; `make docs-test` is 34 passed through `uvx --with pytest --with pyyaml pytest tools/docs -q`, because this machine's python3 has no pytest module
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/); `CHANGELOG.md` untouched
- [x] Caught up with `origin/main` by merge (`31c3625e`); the branch sits on #2684 until that PR merges

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -79,10 +79,7 @@ an array element:
 The same `null` is a value inside the expression too: `$exists()` on it is true and `$type()` on it
 is `"null"`.
 
-Two deviations. `$count()` on a JSON null answers `0`; AWS answers `1`. And dropping the key of an
-expression that returned nothing is Floci's own behaviour: AWS fails the state instead, with
-`States.QueryEvaluationError` and `The JSONata expression '$states.input.absent' specified for the
-field 'Output/v' returned nothing (undefined).`
+One deviation. `$count()` on a JSON null answers `0`; AWS answers `1`.
 
 ## An expression that returns nothing fails the state
 
@@ -129,8 +126,8 @@ Three behaviours are worth knowing before reading an unexpected result, and all 
   `$partition(items, 2.9)` chunks by 2.
 - Several arguments evaluate to undefined rather than failing: a chunk size of zero, an empty
   array, a `$range` with no `step` or with a step whose sign disagrees with the direction, and a
-  `$hash` with no algorithm. An undefined value is then dropped from the surrounding `Output`
-  object, so the field goes missing instead of the state failing.
+  `$hash` with no algorithm. The argument itself does not fail; the field the expression was
+  written in does, under `States.QueryEvaluationError`, as the section above records.
 - `$range` collapses a single-element range to the bare number, not a one-element array.
 
 JSONata's own `$string` follows AWS's number notation: a whole number is written out in full below
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -100,8 +100,10 @@ that error fires:
 
 The field is named relative to the state, with `/` before each object key and `[i]` for each array
 index: `Output`, `Output/a/b[0]`, `Assign/x`, `Arguments/MessageGroupId`, `Choices[1]/Condition`,
-`Seconds`, `Error`, `Cause`, `Items`, `MaxConcurrency`. A `Choice` stops at the first rule that
-matches, so an undefined condition in a later rule is never evaluated.
+`Seconds`, `Error`, `Cause`, `Items`, `MaxConcurrency`. A matched `Choice` rule and a matching
+`Catch` clause carry their own `Assign` and `Output`, which are named under the rule or clause:
+`Choices[1]/Output/v`, `Choices[0]/Assign/x`, `Catch[1]/Output/v`. A `Choice` stops at the first rule
+that matches, so an undefined condition in a later rule is never evaluated.
 
 One deviation. AWS prefixes the cause of a real execution with
 `An error occurred while executing the state '<name>' (entered at the event id #<n>).`; Floci
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -77,13 +77,35 @@ an array element:
 ```
 
 The same `null` is a value inside the expression too: `$exists()` on it is true and `$type()` on it
-is `"null"`. Only an expression that returns nothing loses its key, which is what
-`$states.input.absent` and the functions listed below that evaluate to undefined do.
+is `"null"`.
 
 Two deviations. `$count()` on a JSON null answers `0`; AWS answers `1`. And dropping the key of an
 expression that returned nothing is Floci's own behaviour: AWS fails the state instead, with
 `States.QueryEvaluationError` and `The JSONata expression '$states.input.absent' specified for the
 field 'Output/v' returned nothing (undefined).`
+
+## An expression that returns nothing fails the state
+
+An expression that returns nothing, which is what `$states.input.absent` and the functions listed
+below that evaluate to undefined do, is not a missing value the state carries on without. It fails
+the state with `States.QueryEvaluationError`, naming the field it was written in, and a `Catch` on
+that error fires:
+
+```
+"Output": {"v": "{% $states.input.absent %}"}
+  ->  States.QueryEvaluationError
+      The JSONata expression '$states.input.absent' specified for the field 'Output/v'
+      returned nothing (undefined).
+```
+
+The field is named relative to the state, with `/` before each object key and `[i]` for each array
+index: `Output`, `Output/a/b[0]`, `Assign/x`, `Arguments/MessageGroupId`, `Choices[1]/Condition`,
+`Seconds`, `Error`, `Cause`, `Items`, `MaxConcurrency`. A `Choice` stops at the first rule that
+matches, so an undefined condition in a later rule is never evaluated.
+
+One deviation. AWS prefixes the cause of a real execution with
+`An error occurred while executing the state '<name>' (entered at the event id #<n>).`; Floci
+returns the cause without it, which is the form AWS's own `TestState` returns.
 
 ## JSONata functions
 
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -593,7 +593,8 @@ public class AslExecutor {
             var effectiveInput = input;
             if (stateDef.has("Arguments")) {
                 var statesVar = buildStatesVar(input, null, context);
-                effectiveInput = jsonataEvaluator.resolveTemplate(stateDef.get("Arguments"), statesVar, variables);
+                effectiveInput = jsonataEvaluator.resolveTemplate(
+                        stateDef.get("Arguments"), "Arguments", statesVar, variables);
             }
             taskResult = mockedSteps != null
                     ? mockedTaskResult(mockedSteps, stateName, attempt)
@@ -1650,10 +1651,12 @@ public class AslExecutor {
         if (jsonata) {
             JsonNode statesVar = buildStatesVar(input, null, context);
             JsonNode choices = stateDef.path("Choices");
-            for (JsonNode choice : choices) {
+            for (int i = 0; i < choices.size(); i++) {
+                JsonNode choice = choices.get(i);
                 String condition = choice.path("Condition").asText(null);
                 if (condition != null) {
-                    JsonNode result = jsonataEvaluator.evaluate(condition, statesVar, variables);
+                    JsonNode result = jsonataEvaluator.evaluateField(
+                            condition, "Choices[" + i + "]/Condition", statesVar, variables);
                     if (result.isBoolean() && result.asBoolean()) {
                         // A matched rule carries its own Assign and Output; the state-level ones
                         // belong to the Default path and do not run here.
@@ -1777,7 +1780,8 @@ public class AslExecutor {
                 JsonNode secondsNode = stateDef.get("Seconds");
                 if (secondsNode.isTextual() && JsonataEvaluator.isExpression(secondsNode.asText())) {
                     JsonNode statesVar = buildStatesVar(input, null, context);
-                    JsonNode result = jsonataEvaluator.evaluate(secondsNode.asText(), statesVar, variables);
+                    JsonNode result = jsonataEvaluator.evaluateField(
+                            secondsNode.asText(), "Seconds", statesVar, variables);
                     seconds = Math.min(result.asInt(), MAX_WAIT_SECONDS);
                 } else {
                     seconds = Math.min(secondsNode.asInt(), MAX_WAIT_SECONDS);
@@ -1818,10 +1822,10 @@ public class AslExecutor {
         if (jsonata) {
             JsonNode statesVar = buildStatesVar(input, null, context);
             if (error != null && JsonataEvaluator.isExpression(error)) {
-                error = jsonataEvaluator.evaluate(error, statesVar, variables).asText();
+                error = jsonataEvaluator.evaluateField(error, "Error", statesVar, variables).asText();
             }
             if (cause != null && JsonataEvaluator.isExpression(cause)) {
-                cause = jsonataEvaluator.evaluate(cause, statesVar, variables).asText();
+                cause = jsonataEvaluator.evaluateField(cause, "Cause", statesVar, variables).asText();
             }
         }
         throw new FailStateException(error, cause);
@@ -2031,7 +2035,7 @@ public class AslExecutor {
             if (jsonata && value.isTextual() && JsonataEvaluator.isExpression(value.asText())) {
                 jsonataExpression = true;
                 JsonNode statesVar = buildStatesVar(mapInput, null, context);
-                value = jsonataEvaluator.evaluate(value.asText(), statesVar, variables);
+                value = jsonataEvaluator.evaluateField(value.asText(), "MaxConcurrency", statesVar, variables);
             }
         } else {
             return 0;
@@ -2118,19 +2122,8 @@ public class AslExecutor {
             // InputPath, which is supplied by executeMapState.
             JsonNode loc;
             if (jsonata && writer.has("Arguments")) {
-                JsonNode arguments = writer.get("Arguments");
-                loc = jsonataEvaluator.resolveTemplate(
-                        arguments, buildStatesVar(input, null, context), variables);
-                if (arguments.isObject()) {
-                    if (arguments.has("Bucket") && !loc.has("Bucket")) {
-                        throw new FailStateException("States.QueryEvaluationError",
-                                "ResultWriter Bucket must resolve to a string");
-                    }
-                    if (arguments.has("Prefix") && !loc.has("Prefix")) {
-                        throw new FailStateException("States.QueryEvaluationError",
-                                "ResultWriter Prefix must resolve to a string");
-                    }
-                }
+                loc = jsonataEvaluator.resolveTemplate(writer.get("Arguments"), "ResultWriter/Arguments",
+                        buildStatesVar(input, null, context), variables);
             } else if (writer.has("Parameters")) {
                 loc = resolveParameters(writer.get("Parameters"), input, context);
             } else {
@@ -2287,7 +2280,8 @@ public class AslExecutor {
             JsonNode itemsNode = stateDef.get("Items");
             if (itemsNode.isTextual() && JsonataEvaluator.isExpression(itemsNode.asText())) {
                 JsonNode statesVar = buildStatesVar(input, null, context);
-                return new ResolvedMapItems(jsonataEvaluator.evaluate(itemsNode.asText(), statesVar, variables),
+                return new ResolvedMapItems(
+                        jsonataEvaluator.evaluateField(itemsNode.asText(), "Items", statesVar, variables),
                         MapItemsSource.DEFAULT);
             }
             return new ResolvedMapItems(itemsNode, MapItemsSource.DEFAULT);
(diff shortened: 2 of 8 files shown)


pgermosen (approved):
Traced this directly rather than taking the description at face value, since it's a breaking change to a widely-hit code path. evaluateField and the field-path threading through resolveTemplate produce exactly the AWS field names claimed, including the nested object/array shape. All six positions that call it directly are wired to name their own field correctly.

The two behavior fixes riding along both check out too. A matched Choice rule's or Catch clause's own Output/Assign now names itself under its own index through a threaded prefix rather than the state's own field name, and the Catch-clause-failure handling is a real fix: a clause's own Output/Assign failing now correctly becomes the execution's terminal failure instead of degrading to States.Runtime, and by construction — not a separate check — no later clause gets a chance to catch it once the matched one has already thrown.

This is a clean, correctly scoped follow-up to the null-vs-undefined fix, closing exactly the deviation that one deferred. Nothing outstanding.

PR #2692, opened 2026-08-28
feat(applicationautoscaling): evaluate target-tracking and step-scaling policies against ECS

### LINKED ISSUE #2565: [FEAT] Application Auto Scaling: evaluate target-tracking and update ECS desiredCount
## Service

Application Auto Scaling / ECS / CloudWatch

## API Action / Feature

Make registered target-tracking policies actually adjust capacity. For ECS:

- `RegisterScalableTarget` with `ServiceNamespace=ecs`, `ScalableDimension=ecs:service:DesiredCount`
- `PutScalingPolicy` with `TargetTrackingScaling` + `ECSServiceAverageCPUUtilization` / memory
- Expected: when metrics breach the target, Floci updates the ECS service `desiredCount` (or at least emits a scaling activity)

Today the control plane works (targets, policies, synthesized CloudWatch alarm names) but **policies are stored but inert** — nothing evaluates alarms and no capacity changes (`docs/services/applicationautoscaling.md`).

## AWS Documentation

- [Application Auto Scaling](https://docs.aws.amazon.com/autoscaling/application/userguide/what-is-application-auto-scaling.html)
- [Target tracking scaling policies](https://docs.aws.amazon.com/autoscaling/application/userguide/application-auto-scaling-target-tracking.html)
- [Amazon ECS service auto scaling](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-auto-scaling.html)

## Why is this needed?

Terraform `aws_appautoscaling_target` / `aws_appautoscaling_policy` apply successfully against Floci, so local looks like it has autoscaling. On AWS those resources change `desiredCount`. Operators therefore often omit autoscaling on local, so the reference env never exercises the same module path as AWS.

A minimal loop (even: on a timer, if a CPU metric stub exceeds the target then `UpdateService` desiredCount) would unlock parity testing.

- [ ] Yes
- [ ] No

omatheusmesmo:
## Summary
- Adds a generic CloudWatch alarm evaluator (`cloudwatch.metrics`) that transitions alarm state (`OK`/`ALARM`/`INSUFFICIENT_DATA`) from real metric data and dispatches `AlarmActions` on a new breach, independent of Application Auto Scaling.
- Adds an ECS `CapacityAdjuster` + scaling-policy alarm-action handler so a firing `TargetTrackingScaling` or `StepScaling` policy calls `UpdateService` to change `desiredCount`, respecting cooldown, `MinCapacity`/`MaxCapacity`, and `SuspendedState`.
- Adds `DescribeScalingActivities` so the resulting capacity changes are observable through the real AWS API.
- Fixes two latent gaps in the existing alarm-synthesis code: `ActionsEnabled` was never set to `true`, and synthesized ECS alarms never carried `ClusterName`/`ServiceName` dimensions, so a metric pushed in the real AWS shape would never have been found.

Closes #2565

## Test plan
- [x] `ApplicationAutoScalingServiceTest`, `ScalingPolicyAlarmActionHandlerTest`, `EcsCapacityAdjusterTest`, `AlarmEvaluatorTest` (new/updated unit tests)
- [x] `ApplicationAutoScalingEcsControlLoopIntegrationTest` (new end-to-end test: creates a real ECS service, registers a target-tracking policy, pushes breaching `PutMetricData`, asserts `desiredCount` changes via `DescribeServices` and shows up in `DescribeScalingActivities`)
- [x] Full `applicationautoscaling`, `cloudwatch`, `autoscaling`, and `ecs` test suites pass with no regressions
- [x] `make docs-check` passes (action tables + deferred handler registration in sync)

### DIFF
--- a/docs/services/applicationautoscaling.md
+++ b/docs/services/applicationautoscaling.md
@@ -22,6 +22,7 @@ Auto Scaling groups over the Query protocol under the `autoscaling` signing name
 | `PutScalingPolicy` | Creates or updates a scaling policy and its CloudWatch alarms |
 | `DescribeScalingPolicies` | Lists scaling policies in a namespace, optionally filtered |
 | `DeleteScalingPolicy` | Deletes a scaling policy and its CloudWatch alarms |
+| `DescribeScalingActivities` | Lists recorded capacity-change events for a scalable target |
 | `ListTagsForResource` | Returns the tags on a scalable target |
 | `TagResource` | Adds or overwrites tags on a scalable target |
 | `UntagResource` | Removes tags from a scalable target |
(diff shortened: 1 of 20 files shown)


pgermosen (commented):
Went through both open findings directly rather than taking either at face value, since this loop actually changes ECS desired count and one of them touches a safety feature.

The DisableScaleIn one is real. computeTargetTracking's guard checks the unclamped computed value against current, but handle() decides whether an adjustment counts as scale-out or scale-in from the post-clamp value, and the scale-in branch never re-checks DisableScaleIn. So if current capacity already sits above MaxCapacity for any reason and a scale-out-directed metric computes a raw increase that still clamps down below current, the decrease goes through even with scale-in disabled. Worth closing before this merges.

The other one — the ignore-treatment branch "freezing" an alarm in ALARM — I checked against the actual CloudWatch documentation rather than reasoning about it in the abstract, and it looks like this is correct, AWS-matching behavior rather than a bug. AWS's own docs state, across every example in the missing-data table, that ignore means retaining the current state once there aren't enough real datapoints to satisfy Evaluation Periods, without re-evaluating what real data remains — and they explicitly call out that this can make an alarm keep re-executing its actions after a metric stops flowing. That's exactly what's implemented here. Worth knowing so it doesn't get "fixed" into a divergence from real AWS.

Everything else here reads well — the cooldown/suspension design correctly keys off the computed adjustment's direction rather than the alarm's own comparison direction, and the dedicated lastScaleOutCapacity field is a real, deliberate fix for the flapping issue raised earlier in review.

omatheusmesmo:
Hi @pgermosen , could you kindly review it again? You commented while I was interacting with greptile and now there is a new commit. Thank you!

pgermosen (approved):
Approving. The DisableScaleIn fix is exactly right — moving the check to the post-clamp newCapacity in handle() closes the gap where a clamp to MaxCapacity could turn a scale-out into a real decrease and slip past the guard. Traced through the new test's numbers (current=15 above max=10, computed=24, clamps to 10) and it lines up with the bug I flagged. Nothing else changed besides this handler and its test, so the rest of my earlier read still stands.

Issue #2671, opened 2026-08-28
[FEAT] Step Functions (SFN): `DescribeMapRun` is not implemented, so a distributed `Map` cannot be inspected after it finishes


facuparedes:
## Service

Step Functions

## API Action / Feature

`DescribeMapRun`, and the `arn:aws:states:::aws-sdk:sfn:describeMapRun` `Task` integration that reaches it.

## AWS Documentation

https://docs.aws.amazon.com/step-functions/latest/apireference/API_DescribeMapRun.html

## Expected behavior

After a distributed `Map` completes, `DescribeMapRun` returns the run's counters, so a state machine can report how many items succeeded and how many failed. `itemCounts` and `executionCounts` carry ten fields each: `pending`, `running`, `succeeded`, `failed`, `timedOut`, `aborted`, `total`, `resultsWritten`, `failuresNotRedrivable`, `pendingRedrive`.

## Actual behavior

Two gaps, and both have to close for the ASL `Task` to work.

The API itself is not implemented:

```
An error occurred (UnsupportedOperation) when calling the DescribeMapRun operation:
Operation DescribeMapRun is not supported.
```

And the resource is not in the dispatch chain in `AslExecutor.invokeResource` (`main`, line 667), so it fails at the `throw` that closes it on line 812 before any client is reached.

## Reproduction

The distributed `Map` itself runs, and its result is where an ASL author gets the ARN, so this starts from a real run rather than a hand-written ARN:

```bash
docker run -d --name floci -p 4566:4566 floci/floci:nightly
export AWS_REGION=us-east-1 AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test
export AWS_ENDPOINT_URL=http://localhost:4566

aws s3api create-bucket --bucket dm
printf '[{"n":1},{"n":2},{"n":3}]' > items.json
aws s3api put-object --bucket dm --key items.json --body items.json

DMAP=$(aws stepfunctions create-state-machine --name dmap \
  --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"QueryLanguage":"JSONata","StartAt":"M","States":{"M":{"Type":"Map","End":true,
    "Mode":"DISTRIBUTED",
    "ItemReader":{"Resource":"arn:aws:states:::s3:getObject",
      "ReaderConfig":{"InputType":"JSON"},"Arguments":{"Bucket":"dm","Key":"items.json"}},
[...]

Issue #2670, opened 2026-08-28
[BUG] Step Functions (SFN): a bare top-level JSONata reference is accepted, AWS rejects the definition


facuparedes:
## Service

Step Functions (`QueryLanguage: JSONata`)

## AWS API Action

`CreateStateMachine`

## Expected behavior

Real AWS validates every JSONata expression when the definition is created. A path that does not start from `$states` or from a variable is rejected:

```
UNSUPPORTED_JSONATA_EXPRESSION: Reference to 'phone' at the top level is not supported.
at /States/E/Output/v
```

## Actual behavior

Floci accepts the definition and evaluates the expression against no context, so it returns nothing and the key is dropped. The state machine passes locally and fails to deploy.

This is not one expression. Of the 1674 I ran from the [jsonata-js v2.2.2 test suite](https://github.com/jsonata-js/jsonata/tree/v2.2.2/test/test-suite/groups), 297 open with a bare top-level reference. I put each one through both sides as the same one-state `Pass` machine, and Floci accepts the definition for all 297. AWS accepts none:

| AWS refuses the definition with | Floci | |
|---|---|---|
| `UNSUPPORTED_JSONATA_EXPRESSION` | `SUCCEEDED` | 275 |
| `UNSUPPORTED_JSONATA_EXPRESSION` | `FAILED` | 2 |
| `INVALID_JSONATA_EXPRESSION` | `RUNNING`, never terminates | 12 |
| `INVALID_JSONATA_EXPRESSION` | `FAILED` | 8 |

The 277 in the first two rows are this issue: AWS names the identifier and refuses, Floci runs the machine. The other 20 carry a syntax error that AWS reports first, so they say nothing about the top-level rule; the 12 that never terminate are the separate `??` and `?:` hang.

The AWS column used `validate-state-machine-definition`, which creates nothing. The `create-state-machine` receipt below is the API this issue names, and it returns the same code and message.

## Reproduction

```bash
docker run -d --name floci -p 4566:4566 floci/floci:nightly
export AWS_REGION=us-east-1 AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test
export AWS_ENDPOINT_URL=http://localhost:4566

SM=$(aws stepfunctions create-state-machine --name bare --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"QueryLanguage":"JSONata","StartAt":"E","States":{
[...]

PR #2693, opened 2026-08-28
fix(stepfunctions): end the execution when a state throws an Error

### LINKED ISSUE #2666: [BUG] Step Functions (SFN): the JSONata `??` and `?:` operators leave the execution RUNNING forever in a native image
**Native images only.** `floci/floci:latest-jvm` and `main` on the JVM return `SUCCEEDED {"v":42}` for the repro below.

## Service

## AWS API Action

## Expected behavior

Every execution reaches a terminal status. Whatever a state throws, the execution ends: AWS returns `FAILED` from `DescribeExecution` with an `error` and a `cause`, and Floci does the same today for anything that extends `Exception`.

`??` and `?:` are the reachable trigger, not the defect. Real AWS refuses both at definition time, so a state machine that deploys never runs one:

```bash
# the placeholder credentials the setup below exports. Static keys in the environment
# instead of reaching AWS.
  aws stepfunctions create-state-machine --region us-east-1 --name c --role-arn "$ROLE_ARN" \
      "E":{"Type":"Pass","Output":{"v":"{% baz ?? 42 %}"},"End":true}}}'
# An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: Invalid
# State Machine Definition: 'INVALID_JSONATA_EXPRESSION: The symbol "?" cannot be used as a
# unary operator at /States/E/Output/v'
```

What is worth fixing is the invariant behind it: any `Error` reaching `doExecute` strands the execution, and the next one to appear will not come with an AWS-side rejection.

## Actual behavior

The execution stays `RUNNING` forever with the container idle. Nothing is written to the log, no history event is added, and a client cannot tell it apart from slow work.

`AslExecutor.doExecute` catches `Exception` at lines 367 and 392 and never `Error`. When an `Error` escapes, `runUnderExecutionAccount` rethrows it (`catch (RuntimeException | Error e)` at line 269, `throw e` at 270), the `executor.submit` at line 220 discards the `Future`, the worker thread dies and nothing writes a terminal status. The comment at line 1640 already names this failure mode for the `Parallel` path.

`Map` reaches the same dead end by two more routes: `MapIterationScheduler` runs iterations on the calling thread when `itemCount == 1` or `MaxConcurrency` is 1 (lines 34-40), so nothing wraps the `Error` at all; and on the concurrent path it unwraps and rethrows the bare `Error` (lines 72-73), unlike the `Parallel` handler which keeps the wrapper.

## Reproduction

```bash

SM=$(aws stepfunctions create-state-machine --name hang1 --role-arn arn:aws:iam::000000000000:role/r \
    "E":{"Type":"Pass","Output":{"v":"{% baz ?? 42 %}"},"End":true}}}' \

sleep 60
# RUNNING	None	None
```

The same expression inside a single-branch `Parallel` surfaces the cause. The `catch (ExecutionException e)` at `AslExecutor:1636-1645` unwraps the cause only when it is an `Exception`; when it is an `Error` it rethrows the `ExecutionException` wrapper, which *is* an `Exception`, so `doExecute` catches it and the execution terminates:

```bash
SM=$(aws stepfunctions create-state-machine --name hang1par --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"QueryLanguage":"JSONata","StartAt":"P","States":{"P":{"Type":"Parallel","End":true,
    "Branches":[{"StartAt":"E","States":{
      "E":{"Type":"Pass","Output":{"v":"{% baz ?? 42 %}"},"End":true}}}]}}}' \

```

```
FAILED	States.Runtime	com.oracle.svm.core.jdk.UnsupportedFeatureError: SerializationConstructorAccessor class not found for declaringClass: com.dashjoin.jsonata.Parser$Symbol (targetConstructorClass: java.lang.Object). Usually adding com.dashjoin.jsonata.Parser$Symbol to serialization-config.json fixes the problem.
```

`UnsupportedFeatureError` extends `Error`, which is why the plain `Pass` hangs where the `Parallel` fails.

The two operators that reach it are `??` and `?:`. Both go through `Parser.clone()`, which builds its result through Java serialization (`Parser$6` and `Parser$17` in the 0.9.10 jar). The `coalescing-operator` and `default-operator` groups of the [jsonata-js v2.2.2 test suite](https://github.com/jsonata-js/jsonata/tree/v2.2.2/test/test-suite/groups) are 34 expressions between them, and all 34 hang.

## Suggested fix

Catch `Throwable` instead of `Exception` in `AslExecutor.doExecute` (lines 367 and 392), so an `Error` produces the same terminal `FAILED` update an exception already produces. `MapIterationScheduler` then no longer needs its own `Error` branch.

Two things to decide while doing it: whether the handler's own `addEvent` / `onUpdate` path survives an `OutOfMemoryError` (it allocates), and whether it should still run after `futures.forEach(future -> future.cancel(true))` at line 1637 has interrupted the thread.

Registering `Parser$Symbol` for serialization in the native image is a separate and optional change, and one I would skip: it silences this trigger without touching the invariant, and AWS rejects the two operators anyway.

## Environment

- Floci version / image tag: `floci/floci:nightly` (`nightly-08282026`, revision `c4050936`); same on `floci/floci:latest` (1.7.0). Not reproducible on `floci/floci:latest-jvm` (1.5.11) or on `main` with `java -jar quarkus-run.jar`

facuparedes:
## Summary

An execution that hits a `java.lang.Error` never ends. `AslExecutor.doExecute` catches `Exception` in two places and never `Error`, `runUnderExecutionAccount` rethrows it, and the `executor.submit` in `executeAsync` drops the `Future` that holds it, so the worker thread dies with nothing written: no status, no `error`, no `cause`, no `ExecutionFailed` event. A client polling `DescribeExecution` cannot tell it apart from slow work, and there is nothing to retry, because the execution never failed. Closes #2666, which carries the reproduction and the native-image trigger.

- **The outer handler catches `Error` too**, writes the same terminal `FAILED` an exception produces, and rethrows it (`AslExecutor.java:399-408`).
- **The cause is the Error's `toString()`, not its `getMessage()`.** An Error's message is frequently null, and the type name is the whole diagnostic.
- **`executeAsync` no longer discards the `Future`.** The submitted task catches `RuntimeException | Error` and logs it with its stack trace (`:220-229`), the only place that stack trace can reach the log.
- **One method writes every terminal failure** (`:3135-3143`), so a Fail state, a state that threw and a runtime Error leave the same `error`, `cause` and `ExecutionFailed` event behind. The per-state runtime `catch` folds into the outer one.
- **New test class**: `AslExecutorErrorTerminationTest`, 6 tests, every one bounded by `@Timeout`, since the symptom being guarded is an execution that never terminates.

## Wire-fidelity details (`floci/floci:nightly` and `floci/floci:latest-jvm`, AWS CLI 2.33.7)

This is a Floci lifecycle bug, not a response-shape mismatch, so there is no AWS row to measure against: real Step Functions rejects `??` and `?:` at `CreateStateMachine`, and no AWS execution produces a JVM Error. The reference is Floci's own `Parallel` path, which already terminated because `ExecutionException` wraps the Error into something the `Exception` handlers catch:

```bash
docker run -d --name floci-2666 -p 4578:4566 floci/floci:nightly
# plain Pass, native image
RUNNING	None	None
# same expression inside a single-branch Parallel, native image
FAILED	States.Runtime	com.oracle.svm.core.jdk.UnsupportedFeatureError: SerializationConstructorAccessor class not found for declaringClass: com.dashjoin.jsonata.Parser$Symbol ...
```

That cause string is `ExecutionException.getMessage()`, which is the cause's `toString()`. Using `toString()` on the direct path makes every route report the identical string, so the table below is one value reached six ways.

| Route an Error takes | Execution ends as | Pinned by |
| --- | --- | --- |
| Escapes a state, sync path | `FAILED`, `States.Runtime`, cause is the Error's `toString()` | `errorInsideAStateFailsTheExecutionWithTheErrorAsCause` |
| Escapes a state, history | `ExecutionFailed` event carrying the same error and cause | `errorInsideAStateRecordsAnExecutionFailedEvent` |
| Escapes a state, `executeAsync` (what `StartExecution` uses) | terminal update published to the listener, execution `FAILED` | `errorInsideAStatePublishesTheTerminalUpdateToTheListener` |
| Escapes a `Parallel` branch | `FAILED`, same error and cause as the direct path | `errorInsideAParallelBranchFailsTheExecutionWithTheSameCause` |
| Escapes a `Map` iteration, `MaxConcurrency` 1 (calling thread) | `FAILED`, same error and cause | `errorInsideASerialMapIterationFailsTheExecution` |
| Escapes a `Map` iteration, `MaxConcurrency` 4 (worker pool) | `FAILED`, same error and cause | `errorInsideAConcurrentMapIterationFailsTheExecution` |

`States.Runtime` is the code Floci already uses for a runtime failure with no matching ASL error name; this change adds none.

## Deliberate scope notes

- **The trigger in #2666 is native-only, the defect is not, and the tests are JVM tests.** I confirmed both halves rather than assuming them: `floci/floci:latest-jvm` returns `SUCCEEDED {"v":42}` for the repro, `floci/floci:nightly` returns `RUNNING`. `??` and `?:` clone their parser through Java serialization, which a GraalVM image answers with `UnsupportedFeatureError`. Reaching that from a test would mean building a native image for a defect that has nothing to do with GraalVM, so the tests inject the same shape on the JVM: a `JsonataEvaluator` whose `resolveTemplate` throws a `LinkageError`, thrown exactly where the image throws it, resolving a state's `Output`.
- **Catching `Error` is not the same as tolerating one.** The terminal status is published first and the Error is then rethrown, so nothing is swallowed. That ordering is the point: an execution stuck `RUNNING` is worse than either outcome, since it neither succeeds nor fails and nothing can retry it, while the rethrow keeps an `OutOfMemoryError` or a `StackOverflowError` visible as what it is. The honest limit: the handler allocates, so under a genuine `OutOfMemoryError` the terminal write can itself fail. This makes the common case terminal; it does not make the write OOM-proof.
- **The discarded `Future` is a second, independent defect, fixed here.** `submit()` parks whatever the task throws in a `Future` nobody reads, so before this change nothing logged the Error that killed the worker. The `Error` handling alone would have closed the hole a client sees and left the operator with a silent container. `executeSync` already reads its `Future`, so only `executeAsync` needed it.
- **The `Parallel` path is covered by unwrapping rather than by its wrapper.** The comment at `:1646-1651` relied on the `ExecutionException` wrapper staying an `Exception` so the execution would terminate. That is no longer load-bearing, so the handler unwraps an Error cause the same way it already unwraps an Exception. The reported `error` and `cause` do not move; the Error is now rethrown and logged with its own stack trace instead of the wrapper's.
- **`MapIterationScheduler` is left alone.** #2666 suggests its `Error` branch becomes unnecessary. It does not: without it the bare Error is rewrapped in a `RuntimeException` and the cause gains a wrapper. Both of its routes now reach a terminal status, and both are pinned above.
- **Registering `Parser$Symbol` for serialization is not in this PR.** It silences one trigger without touching the invariant, and AWS rejects both operators anyway.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] No request or response shape changes: this is Floci's execution lifecycle, and `DescribeExecution` returns the fields it already returned
- [x] The terminal state matches what Floci's `Parallel` path already produced for the same Error, measured on `floci/floci:nightly`; every other route now produces the identical `error` and `cause`
- [ ] Not measurable against real Step Functions: AWS rejects `??` and `?:` at `CreateStateMachine`, and no AWS execution produces a JVM Error. The AWS-side invariant restored here is the documented one, that every execution reaches a terminal status

## Checklist

- [x] Targeted Step Functions suites green: 417 tests, 0 failures, 0 errors (411 on `origin/main` plus the 6 new)
- [x] Red proved before the fix. With the production hunk absent and the tests untouched, `AslExecutorErrorTerminationTest` is `Tests run: 6, Failures: 5`: four report `expected: <FAILED> but was: <RUNNING>`, one reports `history has no ExecutionFailed event: [PassStateEntered]`, and the async one reports `no terminal update was published; execution is still RUNNING` after waiting 30 seconds
- [x] The sixth, `errorInsideAParallelBranchFailsTheExecutionWithTheSameCause`, passes on `main` too: it is the regression pin for the one route that already terminated, and it stays green through the `Parallel` unwrap
- [x] Every test carries a `@Timeout`, so the never-terminating symptom fails the suite instead of blocking it
- [x] Branched from current `origin/main` (3bd9bcc6)
- [x] No docs change: nothing under `docs/` documents this lifecycle
- [ ] The full `./mvnw test` run is 13695 tests with 1 failure outside this diff, `Ec2ContainerManagerTest.launchLabelsContainerWithResourceIdentity`, the same one #2676 recorded. It does not touch `stepfunctions`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -217,7 +217,16 @@ public class AslExecutor {
                              MockedTestCase mockedTestCase,
                              BiConsumer<Execution, List<HistoryEvent>> onUpdate) {
         registerMocks(exec, mockedTestCase);
-        executor.submit(() -> runUnderExecutionAccount(sm, () -> doExecute(sm, exec, history, onUpdate)));
+        executor.submit(() -> {
+            try {
+                runUnderExecutionAccount(sm, () -> doExecute(sm, exec, history, onUpdate));
+            } catch (RuntimeException | Error e) {
+                // submit() parks whatever the task throws in a Future nobody reads, so without this
+                // the worker dies silently. doExecute has already published the terminal status by
+                // now; this is the only place the stack trace of what killed it reaches the log.
+                LOG.errorv(e, "ASL execution worker failed for {0}", exec.getExecutionArn());
+            }
+        });
     }
 
     /**
@@ -313,8 +322,10 @@ public class AslExecutor {
 
     private void doExecute(StateMachine sm, Execution exec, List<HistoryEvent> history,
                            BiConsumer<Execution, List<HistoryEvent>> onUpdate) {
+        // Declared outside the try so the terminal handlers below can keep numbering the history
+        // where the execution left it.
+        AtomicLong eventId = new AtomicLong(history.size());
         try {
-            AtomicLong eventId = new AtomicLong(history.size());
             JsonNode definition = objectMapper.readTree(sm.getDefinition());
             JsonNode states = definition.path("States");
             String startAt = definition.path("StartAt").asText();
@@ -364,17 +375,6 @@ public class AslExecutor {
                     failExecution(exec, history, eventId, e);
                     onUpdate.accept(exec, history);
                     return;
-                } catch (Exception e) {
-                    String runtimeError = "States.Runtime";
-                    String runtimeCause = e.getMessage() != null ? e.getMessage() : "Unknown error";
-                    exec.setError(runtimeError);
-                    exec.setCause(runtimeCause);
-                    exec.setStopDate(System.currentTimeMillis() / 1000.0);
-                    exec.setStatus("FAILED");
-                    addEvent(history, eventId, "ExecutionFailed", null,
-                            Map.of("error", runtimeError, "cause", runtimeCause));
-                    onUpdate.accept(exec, history);
-                    return;
                 }
             }
 
@@ -393,11 +393,19 @@ public class AslExecutor {
             LOG.warnv("ASL execution failed for {0}: {1}", exec.getExecutionArn(), e.getMessage());
             // This path previously set only the status, leaving error and cause null forever on an
             // execution DescribeExecution reports as FAILED.
-            exec.setError("States.Runtime");
-            exec.setCause(e.getMessage() != null ? e.getMessage() : "Unknown error");
-            exec.setStopDate(System.currentTimeMillis() / 1000.0);
-            exec.setStatus("FAILED");
+            failExecution(exec, history, eventId, "States.Runtime",
+                    e.getMessage() != null ? e.getMessage() : "Unknown error");
             onUpdate.accept(exec, history);
+        } catch (Error e) {
+            // An Error is not a state failure: it says the runtime itself is broken, and no retry
+            // of the state machine can get past it. Publishing the same terminal FAILED an
+            // exception produces is what keeps DescribeExecution from reporting RUNNING forever,
+            // and the rethrow keeps the Error itself from being swallowed here. The cause carries
+            // toString() rather than getMessage(), because an Error's message is often null and
+            // the type name is the whole diagnostic.
+            failExecution(exec, history, eventId, "States.Runtime", e.toString());
+            onUpdate.accept(exec, history);
+            throw e;
         } finally {
             activeMocks.remove(exec.getExecutionArn());
         }
@@ -1635,12 +1643,17 @@ public class AslExecutor {
             throw e;
         } catch (ExecutionException e) {
             futures.forEach(future -> future.cancel(true));
-            // Unwrap so a branch's FailStateException reaches the Parallel state's own
-            // Retry and Catch handling instead of surfacing as States.Runtime. An Error
-            // cause stays wrapped so the execution-level Exception handlers still
-            // publish a terminal FAILED update instead of leaving the execution RUNNING.
+            // Unwrap so a branch's FailStateException reaches the Parallel state's own Retry and
+            // Catch handling instead of surfacing as States.Runtime, and so an Error reaches the
+            // execution-level handler as itself rather than as an ExecutionException wrapper. The
+            // reported cause is the same either way, since ExecutionException.getMessage() is the
+            // cause's toString(), but only the unwrapped Error is rethrown and logged with its
+            // stack trace.
             if (e.getCause() instanceof Exception exception) {
                 throw exception;
+            }
+            if (e.getCause() instanceof Error error) {
+                throw error;
             }
             throw e;
         } catch (Exception | Error e) {
@@ -3110,14 +3123,23 @@ public class AslExecutor {
     }
 
     private void failExecution(Execution exec, List<HistoryEvent> history, AtomicLong eventId, FailStateException e) {
-        String failError = e.error != null ? e.error : "States.Runtime";
-        String failCause = e.cause != null ? e.cause : "";
-        exec.setError(failError);
-        exec.setCause(failCause);
+        failExecution(exec, history, eventId, e.error != null ? e.error : "States.Runtime",
+                e.cause != null ? e.cause : "");
+    }
+
+    /**
+     * The single terminal-failure write: every way an execution can fail leaves the same
+     * {@code error}, {@code cause} and {@code ExecutionFailed} event behind, so a client cannot
+     * tell a Fail state from a state that threw from a runtime Error by what it reads back.
+     */
+    private void failExecution(Execution exec, List<HistoryEvent> history, AtomicLong eventId,
+                               String error, String cause) {
+        exec.setError(error);
+        exec.setCause(cause);
         exec.setStopDate(System.currentTimeMillis() / 1000.0);
         exec.setStatus("FAILED");
         addEvent(history, eventId, "ExecutionFailed", null,
-                Map.of("error", failError, "cause", failCause));
+                Map.of("error", error, "cause", cause));
     }
 
     private StateResult handleCatch(JsonNode stateDef, JsonNode input, FailStateException failure,
--- a/src/test/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutorErrorTerminationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutorErrorTerminationTest.java
@@ -1,0 +1,256 @@
+package io.github.hectorvent.floci.services.stepfunctions;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.hectorvent.floci.config.EmulatorConfig;
+import io.github.hectorvent.floci.services.dynamodb.DynamoDbJsonHandler;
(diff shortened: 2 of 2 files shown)


facuparedes:
Thanks, and that count is deliberate rather than accidental, so here is the context the notice asks for.

These are one batch. I filed ten issues against the Step Functions JSONata evaluator and the `Task` service integrations after running our own 174 state machines and the jsonata-js v2.2.2 suite against Floci, and I am contributing a fix for each. #2676 is already merged. The rest map one to one:

| Issue | PR | |
| --- | --- | --- |
| #2662 | #2676 | merged |
| #2669 | #2683 | |
| #2664 | #2684 | |
| #2663 | #2686 | |
| #2668 | #2687 | |
| #2665 | #2689 | draft, stacked on #2684 |
| #2666 | #2693 | |
| #2671 | #2694 | draft, stacked on #2686 |
| #2667, #2670 | opening shortly | |

**Why one PR per issue rather than fewer, larger ones.** Each carries a single decision a reviewer can accept or reject on its own, and several are independent of the rest. Collapsing them would mean asking for one approval on two unrelated behaviour changes.

**Why two are drafts.** #2689 needs #2684 (a guard cannot tell an explicit `null` from a missing value until the evaluator distinguishes them), and #2694 needs #2686 (the dispatch branch it adds sits in the chain that PR builds). GitHub cannot express a stacked PR from a fork, so their diffs show the parent's commits too. Each one's first line says which commit to review, and both come out of draft and get rebased once the parent merges. Nothing else in the batch depends on anything.

**Two are behaviour changes and are marked as such**: #2689 fails a state where Floci previously succeeded with a dropped key, and the one for #2670 will reject definitions Floci accepts today. Both are cases where the current behaviour hides an error until deploy time.

If this is more than you want open at once, say the word and I will close all but the two or three you would rather see first and reopen the others as those merge. Same if you would prefer a different order: the two that unblock the most for us are #2686 and #2687.

pgermosen (approved):
I traced the full call chain by hand rather than trusting the description, since catch ordering and rethrow timing in exactly this kind of Error-handling code is where subtle regressions hide.

The publish-then-rethrow ordering in the new catch (Error e) block is correct: the terminal state is written and the update published before the Error continues propagating, and the finally block still runs regardless since Java guarantees that even when a catch block rethrows. I checked runUnderExecutionAccount's claimed pre-existing rethrow behavior directly rather than assuming it, and it really does rethrow a RuntimeException or Error unmodified. Both callers correctly surface the Error without duplicating work: executeSync's Future.get() wraps it in a checked ExecutionException that lands in its existing catch block, which is fine since doExecute has already written the terminal state by the time that runs; executeAsync's new try/catch around the submitted task is what actually closes the real gap here, since before this PR the Future's return value was discarded entirely and nothing ever read it. The Parallel-branch unwrap is exhaustive too — an ExecutionException's cause can only be an Exception, an Error, or null, and the new code checks both explicitly.

The regression tests are also honest about scope: the PR measured directly that the trigger from the linked issue only reproduces under native-image, and correctly avoided the cost of a native-image CI build for a defect that isn't native-specific by injecting an equivalent LinkageError on the JVM instead.

PR #2697, opened 2026-08-28
fix(stepfunctions): bound JSONata evaluation by time and depth

### LINKED ISSUE #2667: [BUG] Step Functions (SFN): JSONata evaluation has no time or depth bound, so an expression can run forever

facuparedes:
## Summary

JSONata evaluation ran with no bound at all. A recursive expression with no base case pinned a core and left the execution `RUNNING` for ever, with no log line and no history event, on the native image and on the JVM alike. Closes #2667, which carries the one-container reproduction.

- **This is not #2666 and neither PR supersedes the other.** #2666 is an `Error` escaping the executor's catch, fixed by handling `Error` there. This one throws nothing: JSONata optimises the tail call, so the expression loops rather than growing the stack, and there is no `Throwable` for any catch to see. Only a clock can end it.
- **`Jsonata.Frame#setRuntimeBounds(long, int)` already existed and Floci never called it.** `grep -rn setRuntimeBounds src/` on `main` returns nothing. It installs a `Timebox` on the evaluation frame that counts entry and exit callbacks and raises a `JException` when either bound trips.
- **`JException` is a `RuntimeException`,** so the existing `catch (Exception e)` in `JsonataEvaluator.evaluate` turns it into `States.QueryEvaluationError` with no new branch, and the state stays catchable by `Catch` and retryable by `Retry`, as on AWS.
- **The change is one call plus two constants** (`JsonataEvaluator.java:148`). The second constructor is a test seam, so a test can trip the time bound in 250 ms instead of waiting the shipped five seconds.

`test-state` evaluates one state and creates nothing, so every row below is one call:

```bash
  --definition '{"Type":"Pass","QueryLanguage":"JSONata","Output":{"v":"{% [1..900000] ~> $count() %}"},"End":true}'
# [ "FAILED", null, "States.QueryEvaluationError", "... threw an error during evaluation:
#   Expression evaluation memory limit exceeded: Check for excessive memory usage" ]
```

| Expression | AWS | Floci | Pinned by |
| --- | --- | --- | --- |
| `($f := function($x){ $f($x+1) }; $f(1))` | `States.QueryEvaluationError`, cause `Expression evaluation memory limit exceeded: Check for excessive memory usage` | `States.QueryEvaluationError`, cause `Expression evaluation timeout: Check for infinite loop` | `recursionWithNoBaseCaseFailsOnTheTimeBound` |
| `($f := function($x){ $x <= 0 ? 0 : 1 + $f($x-1) }; $f(1000))` | `States.QueryEvaluationError`, cause `Stack overflow error: Check for non-terminating recursive function. Consider rewriting as tail-recursive` | same cause plus `Depth=501 max=500`, at `$f(1000000)` | `nonTailRecursionFailsOnTheDepthBound` |
| `($f := ...; $f(30))` | `30` | `30` | `expensiveButLegitimateExpressionsStayWithinTheShippedBounds` |
| `[1..800000] ~> $count()` | `800000` | `800000` | same test |
| `($f := ...; $f(160))` | `FAILED` (stack overflow) | `160` | same test, deliberately looser |
| `$sum($map([1..200000], function($v){$v * 2}))` | `FAILED` (memory limit) | `40000200000` | same test, deliberately looser |
| Time bound 5000 ms, depth bound 500 | not published | **chosen, not measured** | both bound tests, at 250 ms and `max=500` |

Both cause assertions are `assertEquals` on the whole string, not a `contains`. Since #2687 stripped the `JSonataException ` prefix that `JException#getMessage` prepends, the depth cause is now AWS's own string plus the `Depth=N max=M` suffix, differing only in the double space the library writes after the full stop.

**Where AWS's own limits fall,** since it publishes neither. Bisecting `[1..N] ~> $count()`: `SUCCEEDED` at 800,000, `FAILED` at 900,000. That tightens the 200k-to-500k range recorded on #2676, which was wrong in the useful direction, since 500,000 succeeds. Bisecting the non-tail recursion, whose evaluation nests at roughly `3n+5`: `SUCCEEDED` at `$f(30)` (depth 95), `FAILED` at `$f(40)` (depth 125). So AWS's depth ceiling sits near 100, which is exactly the dashjoin library's own default.

## Deliberate scope notes

**The two bounds are not the same kind as AWS's, and this PR does not pretend otherwise.** AWS names a *memory* limit on both the large-sequence case and the runaway recursion; `setRuntimeBounds` takes time and depth. So this fails both cases and matches neither mechanism. The depth one is very nearly exact: AWS's string for deep non-tail recursion is the library's own message, minus the `Depth=N max=M` suffix and with one space instead of two. I kept the suffix rather than rewriting the message, because locally it names the bound you hit. The memory case is untouched: `[1..N] ~> $count()` still succeeds at any size here, because the library's ranges are lazy, so nothing is materialised for a clock to catch. #2676 already documented that deviation; the docs edit here corrects its number and records the new bounds.

**On the values, which are a product decision and not a measurement.** Five seconds is the library's own default and roughly fifty times the slowest evaluation of a payload AWS will even accept: summing the 800,000-element sequence that is AWS's ceiling takes about 100 ms here. Five hundred levels is five times AWS's measured ceiling of about 100, so no expression AWS accepts can reach it, and it still trips cleanly before the JVM stack does (on a 1 MB thread stack, 500 raises the JSONata error while 1000 overflows the stack instead). Both are deliberately generous, on the view that a bound obviously too loose is safer than one that fails a legitimate expression. **Change them if you disagree**; they are two constants in one file.

**No configuration knob, though the repo has the mechanism.** `EmulatorConfig.StepFunctionsServiceConfig` holds `allow-plaintext-http` and `mock-config-file`, and a bound would fit there. I left it out because the values are strictly more permissive than AWS on both axes, so an expression needing the escape hatch is one already failing on AWS. Moving them onto that interface is a one-line change if you want it tunable.

**Cost.** The Timebox reads the clock on every AST node entry and exit: 31 to 36 percent on top of evaluation over 200,000 iterations, which is 0.6 to 0.8 microseconds per expression against a state transition that does I/O.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] Verified against real Step Functions in `us-east-1` with AWS CLI 2.33.7, through `test-state`. No state machine was created or left behind
- [x] Both AWS limits bisected rather than assumed: the sequence limit to between 800,000 and 900,000 elements, the depth limit to between nesting depth 95 and 125
- [x] The deviation in kind (memory versus time and depth) is stated above and recorded in `docs/services/step-functions.md`

## Checklist

- [x] Targeted Step Functions suite green: 414 tests, 0 failures, 0 errors (411 on `main` plus 3)
- [x] Red proved before commit, with the `setRuntimeBounds` call reverted and the tests untouched:
  - `nonTailRecursionFailsOnTheDepthBound` fails in 0.066 s with `Unexpected exception type thrown, expected: <FailStateException> but was: <java.lang.StackOverflowError>`
  - `recursionWithNoBaseCaseFailsOnTheTimeBound` never returns. Its `@Timeout(30)` fires, but the JSONata loop ignores the interrupt, so the fork survives it and Surefire reports `There was a timeout in the fork` and `going to kill self fork JVM`. That is the bug in one line
  - `expensiveButLegitimateExpressionsStayWithinTheShippedBounds` passes both with the bounds reverted and with them applied, which is the point of it
- [x] No test can hang the suite: both bound tests carry `@Timeout`, and the time-bound test uses a 250 ms bound, so the three add 0.3 s
- [x] `make docs-check` passes; `make docs-test` is 34 passed via `uvx --with pytest --with pyyaml` (this machine's python3.14 has no pytest). The generated action table is untouched
- [ ] The full `./mvnw test` run still carries `Ec2ContainerManagerTest.launchLabelsContainerWithResourceIdentity`, which fails in a full run and passes in isolation. Pre-existing and documented on #2676, and it does not touch `stepfunctions`


pgermosen (approved):
Genuinely important reliability fix, and I pulled the exact pinned library version rather than trusting the description of its internals. Every claim checks out precisely: setRuntimeBounds really exists with that signature, the underlying Timebox class's own defaults really are 5 seconds and depth 100 (so "the library's own default" is exact, not an approximation), the mechanism really is entry/exit-callback depth counting on every AST node, and the thrown exception type really does extend RuntimeException, confirming the existing catch block needs no new branch to route this into a state failure. Both error message strings match character-for-character, including a genuine quirk in the library's own message (a literal double space before "Consider"), and the new test correctly reproduces that exact spacing rather than normalizing what looks like a typo — the kind of detail that's easy to get subtly wrong by "cleaning up."

The chosen bound values are reasoned rather than asserted: bisected against AWS's actual limits through live test-state calls, and deliberately set looser on both axes than AWS's own ceilings, so no expression AWS accepts could ever hit Floci's bound. The docs update is also upfront that this fixes a time/depth bound specifically, not AWS's separate memory bound, and it corrects a previous boundary estimate that this same bisection now shows was off.

PR #2698, opened 2026-08-29
fix(rds): clear CLIENT_SSL before forwarding the TLS-terminated handshake response to the backend

matanbaruch:
## Summary

Every TLS connection to a MySQL RDS instance through the auth proxy hangs until the client times out. mysql's default `--ssl-mode=PREFERRED` is enough to hit it, so out of the box a stock client cannot connect at all; only `--ssl-mode=DISABLED` works.

A virtual-thread dump during the hang shows the proxy parked in the connection-phase relay, waiting for a backend verdict that never arrives:

```
#103 "rds-proxy-conn-rds-resource:arn:aws:rds:...:db:tls-check" virtual WAITING
    at MySqlProtocolHandler.readMysqlPacketRaw(MySqlProtocolHandler.java:441)
    at MySqlProtocolHandler.relayConnectionPhase(MySqlProtocolHandler.java:200)
    at MySqlProtocolHandler.handleAuth(MySqlProtocolHandler.java:178)
```

**Root cause:** real clients keep `CLIENT_SSL` set in the `HandshakeResponse41` they send over the now-encrypted stream after the `SSLRequest`. The proxy terminates TLS itself and forwards that response to the **plaintext** backend verbatim (only renumbering the sequence). A MySQL server treats any first packet carrying `CLIENT_SSL` as an SSLRequest: it switches into TLS-accept mode and waits for a ClientHello that never comes, while the proxy waits for the auth verdict. Both sides block forever. The existing TLS test missed this because its synthetic client built the post-upgrade response without `CLIENT_SSL`, which no real client does.

**Fix:** clear the `CLIENT_SSL` bit in the client's response before it reaches the backend — the mirror image of the existing `forceClientSslCapability` on the server handshake. One bit in one byte; the sequence-renumber logic is untouched.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: real RDS accepts TLS connections (and `rds.force_ssl` even requires them); Floci's proxy deadlocked on every TLS attempt, mysql's default `--ssl-mode=PREFERRED` included:

```
$ timeout 20 mysql -h <endpoint> -P 7001 -uadmin -p*** --ssl-mode=REQUIRED -e "SELECT 1"
(no output, exit 124)
```

Verified end-to-end against a JVM build with a real `mysql:8.0.36` backend container, all through the proxy as the master user (mysql client 8.0.36):

```
--ssl-mode=REQUIRED   → SELECT succeeds   (hung before the fix)
default (PREFERRED)   → SELECT succeeds   (hung before the fix)
--ssl-mode=DISABLED   → SELECT succeeds   (unchanged)
```

`Ssl_cipher` reads empty by design: it reports the backend leg, which intentionally stays plaintext — TLS terminates at the proxy. `REQUIRED` succeeding is the proof the client leg negotiated TLS.

## Checklist

- [x] `./mvnw test` passes locally (`MySqlProtocolHandlerTest`, `PostgresProtocolHandlerTest`, `RdsProxyManagerTest`)
- [x] New or updated integration test added — the TLS handler tests now build real-client-shaped responses (`CLIENT_SSL` set) and assert the backend-received packet has the bit cleared; without the fix the assertion fails: `expected: <0> but was: <2048>`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/proxy/MySqlProtocolHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/proxy/MySqlProtocolHandler.java
@@ -125,6 +125,12 @@ public class MySqlProtocolHandler {
             // permanently one behind the client's for the rest of the connection phase.
             sequenceOffset = 1;
             clientResponseRaw[3] = (byte) (clientResponseRaw[3] - sequenceOffset);
+            // Real clients keep CLIENT_SSL set in the HandshakeResponse they send over the
+            // now-encrypted stream. The backend connection stays plaintext, and a MySQL server
+            // treats any first packet with CLIENT_SSL as an SSLRequest — it would sit waiting for
+            // a TLS ClientHello that never comes while the proxy waits for its auth verdict,
+            // deadlocking the connection. Clear the bit (the mirror of forceClientSslCapability).
+            clearClientSslCapability(clientResponseRaw);
         } else {
             clientResponseRaw = clientFirstRaw;
             if (clientResponseRaw.length < 40) {
@@ -257,6 +263,15 @@ public class MySqlProtocolHandler {
         } catch (Exception e) {
             throw new IOException("Unable to negotiate MySQL SSL", e);
         }
+    }
+
+    /**
+     * Clears the {@code CLIENT_SSL} bit in a raw HandshakeResponse41 packet's capability flags.
+     * The client capabilities are the first 4 payload bytes, little-endian; {@code CLIENT_SSL}
+     * (0x0800) lives in the second byte.
+     */
+    private static void clearClientSslCapability(byte[] raw) {
+        raw[5] &= (byte) ~(CLIENT_SSL >> 8);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pgermosen (commented):
The fix itself checks out — I traced the bit-level logic directly rather than taking the description at face value. Clearing CLIENT_SSL uses the exact same computation the existing capability-forcing method already uses on the server side (`CLIENT_SSL >> 8`, applied to the byte at the position where that bit actually lives in the little-endian capability field), just with `&= ~(...)` instead of `|=`, and only in the branch that actually upgraded to TLS — a plaintext client's handshake response never has the bit set in the first place, so there's nothing to clear on that path. The new test reconstructs the capability bytes the backend actually received and asserts the bit is gone, which is a real, meaningful assertion rather than a happy-path check.

One thing worth fixing before this merges, though: the PR description doesn't follow this repo's pull request template. The template has three sections this body is missing entirely — Type of change, AWS Compatibility, and a Checklist covering a local test run, an added integration test, and commit-message convention. The content itself (root cause, fix, verification) is good and clearly explained, it's just not in the structure the template asks contributors to fill in. Worth updating the description to include those sections.

matanbaruch:
Description restructured to the PR template (Type of change, AWS Compatibility, Checklist), same content. @pgermosen

pgermosen (approved):
Thanks for restructuring the description to match the template — reads well with the same content in the right sections now.

The fix itself was already sound, so nothing further to add: clearing CLIENT_SSL on the client's post-TLS handshake response before it reaches the plaintext backend is the right mirror of the existing server-side capability forcing, and the updated test asserts the backend actually never sees the bit rather than just checking the happy path.

PR #2699, opened 2026-08-29
fix(stepfunctions)!: reject JSONata top-level references at definition time

### LINKED ISSUE #2670: [BUG] Step Functions (SFN): a bare top-level JSONata reference is accepted, AWS rejects the definition

facuparedes:
## Summary

A Step Functions JSONata expression is evaluated with no context item. The execution input arrives as `$states.input`, a variable written by `Assign` arrives as `$name`, and a path whose first step is a bare name reads a context item that does not exist. Real AWS refuses the whole definition, at `CreateStateMachine`, `UpdateStateMachine` and `ValidateStateMachineDefinition` alike, with `UNSUPPORTED_JSONATA_EXPRESSION: Reference to 'phone' at the top level is not supported. at /States/E/Output/v`. Floci stored the state machine and failed the execution later, so a definition that cannot exist on AWS looked healthy here until it ran. Closes #2670.

- **This is a breaking change, deliberately.** Definitions that run on Floci today stop being creatable. I measured the size of it rather than estimating: across a corpus of 1574 distinct JSONata expressions taken from the jsonata-js v2.2.2 test suite, each placed in a `Pass` state's `Output`, **509 definitions that Floci accepted before are now refused**, 32 percent of the corpus. Every one of them is a definition real AWS also refuses.
- **Where the check runs.** `StepFunctionsService.collectTopLevelReferences` walks every JSONata string of a `QueryLanguage: JSONata` state before the state machine is stored, reporting one diagnostic per distinct name at the `/States/...` location AWS prints, array indices for `Choices`, `Catch` and `Branches` included. A JSONPath state machine is untouched.
- **A static walk, not an evaluation.** `JsonataTopLevelReferences` reads the parse tree and never evaluates. Evaluating at definition time would inherit the non-terminating expressions of #2667; parsing cannot hang, and the whole 1574-expression corpus parses in 0.28 s.
- **The rule is not flow dependent, which the issue assumed it was.** A state that assigns `phone` followed by a state using `{% phone %}` is still refused by AWS; only `{% $phone %}` is accepted. That measurement removed the state-graph analysis this change looked like it needed. `anEarlierAssignDoesNotBindTheBareName` pins both halves.
- **Docs**: a new section in `docs/services/step-functions.md`, outside the generated action-table markers, with the rule, the AWS error block and the two neighbouring rules that stay with the execution.

## Wire-fidelity details (AWS CLI 2.33.7, `validate-state-machine-definition` against `us-east-1`)

`validate-state-machine-definition` creates nothing, so every row below is re-checkable in one call:

```bash
aws stepfunctions validate-state-machine-definition --region us-east-1 --output json \
  --definition '{"QueryLanguage":"JSONata","StartAt":"E","States":{"E":{"Type":"Pass","Output":{"v":"{% phone %}"},"End":true}}}'
# FAIL, UNSUPPORTED_JSONATA_EXPRESSION, "Reference to 'phone' at the top level is not supported.", location "/States/E/Output/v"
```

One `create-state-machine` as the receipt that the same rule fires on the creating path, and that it fires before anything is stored:

```
An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: Invalid State
Machine Definition: 'UNSUPPORTED_JSONATA_EXPRESSION: Reference to 'phone' at the top level is not
supported. at /States/E/Output/v'
```

| Definition shape | AWS location in the message | Pinned by |
| --- | --- | --- |
| `Output: {"v": "{% phone %}"}` | `/States/E/Output/v` | `outputField`, and `createStateMachineRefusesATopLevelReference` for the full `InvalidDefinition` message |
| `Output: {"a":{"b":["{% phone %}"]}}` | `/States/E/Output/a/b[0]` | `nestedOutputField` |
| `Assign: {"v": "{% phone %}"}` | `/States/E/Assign/v` | `assignField` |
| `Choices[0].Condition` | `/States/C/Choices[0]/Condition` | `choiceCondition` |
| `Assign: {"Next": "{% phone %}"}` | `/States/E/Assign/Next` | `aPayloadKeyNamedAfterAnAslFieldIsParsed` |
| `ItemReader.ReaderConfig.CSVHeaders` | AWS returns `OK` and creates the machine | `readerConfigCsvHeadersAreNotParsed` |
| `Wait.Seconds`, `Fail.Cause` | `/States/E/Seconds`, `/States/E/Cause` | `waitSeconds`, `failCause` |
| `Map.Items` and its `ItemProcessor` states | `/States/M/Items`, `/States/M/ItemProcessor/States/I/Output/v` | `mapItemsAndItemProcessor` |
| a `Parallel` branch state | `/States/P/Branches[0]/States/B/Output/v` | `parallelBranchState` |
| a `Catch` output | `/States/E/Catch[0]/Output/v` | `catchOutput` |
| `aaa + bbb` in one field | two diagnostics, one per name, same location | `everyDistinctNameInOneField` |

| Expression | AWS names | Pinned by |
| --- | --- | --- |
| `phone`, `-phone`, `phone & 'x'`, `$sum(phone)`, `[1..phone]`, `{phone: 1}`, `($x := phone; $x)`, `$map($states.input.a, function($y){ phone })`, `$count(phone[inner])` | `phone` | `namesTheTopLevelIdentifier`, 21 expressions |
| `$states.input.items[phone > 3]`, `$states.input.a^(phone)`, `$states.input.a{'k': phone}`, `**.phone`, `{'k': 1}.phone`, `$states.input ~> \|phone\|{'b': 1}\|`, `$phone` | nothing | `namesNothingWhenTheReferenceIsAnchored`, 21 expressions |
| `$`, `$.phone`, `$[0]`, `$string($)` | `$` | `namesTheContextItem`, and `contextItem` at the wire |
| `$$`, `$$.phone` | AWS's separate `Reference to '$$' is not supported.` | `leavesTheRootReferenceAlone` |

The parse tree of `com.dashjoin:jsonata` 0.9.10 is package-private. Sharing the package is what I tried first: it compiles and passes standalone, then throws `IllegalAccessError` under Quarkus, which loads application classes and dependency jars with different class loaders. The fields are therefore read reflectively, all thirteen resolved once in the static initializer, so a dashjoin upgrade that renames or removes one throws `IllegalStateException` naming the field on the first call rather than quietly naming no reference and re-opening this bug. `Parser.Symbol` is registered with `@RegisterForReflection(fields = true)`; the generated `reflect-config.json` carries it with `allDeclaredFields: true`, and no `--initialize-at-build-time` applies, so the `Field` objects are built at image runtime, not baked into the heap.

## Deliberate scope notes

- **Release note for the next release.** *Breaking: Floci now rejects a state machine definition whose JSONata expressions reference a bare top-level name, as AWS does, instead of accepting it and failing the execution. Anchor the reference: `$states.input.<name>` for execution input, `$<name>` for a variable written by `Assign`. A definition that Floci refuses after this change is a definition AWS also refuses.*
- **The differential, and why this is not behind a flag.** I ran all 1574 corpus expressions through AWS's validator and through this checker. **Harmful false positives inside the corpus, definitions AWS accepts that Floci would now refuse: 0.** 999 both accept, 498 both reject as a top-level reference, 11 AWS rejects for a syntax error where we reject for the top-level reference in the same expression (same verdict, different code), 62 AWS rejects for syntax only and we accept, 3 AWS rejects with its `$$` rule and we accept, 1 false negative. With no false positive left to eliminate, a flag would only preserve the divergence it exists to remove.
- **What that corpus could not see, and the one false positive it missed.** Every corpus definition is the same one-state `Pass` with the expression in `Output`, so it varies the expression and never the field. Sweeping the fields instead found one definition AWS accepts and this checker refused: an expression in `ItemReader.ReaderConfig.CSVHeaders`, which holds literal column names. `validate-state-machine-definition` returns `OK` for it and `create-state-machine` stores the machine. Fixed in `bba1ff2`, pinned by `readerConfigCsvHeadersAreNotParsed`.
- **The single false negative, and one name mismatch.** `${id:{'label':label,'value':value}}` is an object constructor applied to `$$`, and we name nothing where AWS names `id`, `label` and `value`. It reaches Floci as an accepted definition AWS refuses, the safe direction. Separately, `Account ~> |Order|{"Product":"blah"},nomatch|` is rejected by both, but AWS also names `nomatch` inside the transform's update clause and we name only `Account`. Both are `$$` and transform interactions, and I left them accepting rather than guessing at the shape.
- **Only the first step of a path is read against the top-level context.** Later steps, predicates, sort terms and object groupings run against the value the step before produced, so a name there is legal and this walk never descends into one. A lambda body is the exception: it keeps the context of the expression that defines it, so `$map($states.input.a, function($x){ b })` does name `b`. Both halves are measured, not reasoned about.
- **Fields AWS does not parse are a measured deny list, and it holds for ASL fields only.** `Comment`, `Next`, `Default`, `Resource`, `ErrorEquals`, `Retry` and `ReaderConfig.CSVHeaders` keep a `{% %}` string as text on AWS, so we do not parse them either. Everything else in a JSONata state is parsed, `ItemSelector`, `ItemBatcher`, `ItemReader.Arguments`, `ResultWriter.Arguments` and `ToleratedFailure*` included, each checked. `Credentials.RoleArn` supports JSONata but an ARN-shape check fires first, so it stays on the list. Those names are ASL fields, never payload keys: AWS parses a payload whole, so `Assign: {"Next": "{% phone %}"}` is refused at `/States/E/Assign/Next` and `Arguments: {"Payload": {"Comment": "{% phone %}"}}` at `/States/E/Arguments/Payload/Comment`. The deny list stops applying once the walk enters `Output`, `Assign`, `Arguments`, `ItemSelector` or `BatchInput`; `aPayloadKeyNamedAfterAnAslFieldIsParsed` pins six of those against AWS.
- **Two neighbouring rules stay out.** A syntax error such as `{% a[1,2) %}` is still accepted here and fails the execution, where AWS reports `INVALID_JSONATA_EXPRESSION` at creation; and `$$`, which AWS refuses under its own message. Both are separate diagnostics with separate messages, and folding them in here would mean guessing at message text I did not measure.
- **One location fix rides along.** A `Parallel` branch was reported at `/States/P/Branches/0/States/...`; AWS writes `/States/P/Branches[0]/States/...`. Measured, corrected, and no other caller or test asserted the old form.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] Verified against real Step Functions in `us-east-1` with AWS CLI 2.33.7, through `validate-state-machine-definition`, which creates nothing, plus one `create-state-machine` for the `InvalidDefinition` wrapper
- [x] 1574-expression differential, Floci against AWS: 0 harmful false positives, 1 false negative. Every message, location and identifier in this PR and in the tests is a captured AWS response
- [x] The flow-dependent reading of the rule was tested against AWS and disproved, so no state-graph analysis is implemented
- [x] No botocore model work: this is definition validation, not the wire protocol

## Checklist

- [x] Step Functions suites green: 506 tests, 0 failures, 0 errors, 0 skipped, up from 436 on `main`
- [x] 50 new unit tests in `JsonataTopLevelReferencesTest` and 20 new integration tests in `StepFunctionsTopLevelReferenceIntegrationTest`, through `CreateStateMachine` and `ValidateStateMachineDefinition`
- [x] Red proof recorded. With `JsonataTopLevelReferences.in` returning an empty list and the tests untouched: `Tests run: 50, Failures: 27` with `expected: <[phone]> but was: <[]>`, and `Tests run: 18, Failures: 13` with `Expected: FAIL / Actual: OK` and `Expected status code <400> but was <200>`. Restored: `Tests run: 68, Failures: 0`. The two tests added in `bba1ff2` were proved the same way, with the field walk reverted and the tests untouched: `Tests run: 20, Failures: 2`, `Expected: OK` for `readerConfigCsvHeadersAreNotParsed` and `Expected: FAIL` for `aPayloadKeyNamedAfterAnAslFieldIsParsed`
- [x] `make docs-check` passes and the docs tooling tests are 34 passed; the generated action table is untouched
- [x] Branch is on current `origin/main` (3954018), caught up with a merge commit
- [x] Commit message follows [Conventional Commits](https://www.conventionalcommits.org/), with `!` and a `BREAKING CHANGE:` footer
- [x] The full `./mvnw test` run on the merged branch is `Tests run: 13888, Failures: 0, Errors: 0, Skipped: 9`. The 9 skips are the `cur` emission suites and `CredentialScopeCoverageTest`, skipped the same way on `main` and untouched by this diff


facuparedes:
Swept the field shapes against `us-east-1` rather than the expression corpus, since every corpus definition is the same one-state `Pass` with the expression in `Output` and so varies the expression and never the field. 69 definitions through `validate-state-machine-definition`, each also run through `StepFunctionsService.collectValidationErrors`. Two divergences, both fixed in `bba1ff2`:

**A rejection AWS does not make.** `ItemReader.ReaderConfig.CSVHeaders` holds literal column names, not expressions:

```
aws stepfunctions validate-state-machine-definition --region us-east-1 --definition '{"QueryLanguage":"JSONata","StartAt":"M","States":{"M":{"Type":"Map","ItemReader":{"Resource":"arn:aws:states:::s3:getObject","Arguments":{"Bucket":"b","Key":"k"},"ReaderConfig":{"InputType":"CSV","CSVHeaderLocation":"GIVEN","CSVHeaders":["{% phone %}"]}},"ItemProcessor":{"StartAt":"I","States":{"I":{"Type":"Pass","End":true}}},"End":true}}}'
# {"result": "OK", "diagnostics": [], "truncated": false}
```

`create-state-machine` with the same definition returned `arn:aws:states:us-east-1:...:stateMachine:pr2699-csvheaders`, so AWS stores it; the machine was deleted afterwards. Before the fix this branch answered `UNSUPPORTED_JSONATA_EXPRESSION ... at /States/M/ItemReader/ReaderConfig/CSVHeaders[0]` and `CreateStateMachine` returned `InvalidDefinition`.

**The bug stayed open for a payload key named after an ASL field.** The deny list was applied to every key at every depth, but those names are ASL fields and AWS parses a payload whole:

| Definition | AWS | this branch before `bba1ff2` |
| --- | --- | --- |
| `Assign: {"Next": "{% phone %}"}` | FAIL, `/States/E/Assign/Next` | accepted |
| `Arguments: {"Payload": {"Comment": "{% phone %}"}}` | FAIL, `/States/E/Arguments/Payload/Comment` | accepted |
| `Output: {"ItemProcessor": {"Iterator": "{% phone %}"}}` | FAIL, `/States/E/Output/ItemProcessor/Iterator` | accepted |
| `ItemSelector: {"Next": "{% phone %}"}` | FAIL, `/States/M/ItemSelector/Next` | accepted |
| `ItemBatcher.BatchInput: {"Comment": "{% phone %}"}` | FAIL, `/States/M/ItemBatcher/BatchInput/Comment` | accepted |
| `Catch[0].Assign: {"Retry": "{% phone %}"}` | FAIL, `/States/E/Catch[0]/Assign/Retry` | accepted |

The deny list now stops applying once the walk enters `Output`, `Assign`, `Arguments`, `ItemSelector` or `BatchInput`. Both tests were proved red with the field walk reverted and the tests untouched: `Tests run: 20, Failures: 2`. Re-running the 69 shapes after the fix leaves 0 in either direction beyond the gaps this PR states are out of scope (`$$`, and the schema checks Floci does not implement for `Credentials.RoleArn`, `Retry[].MaxDelaySeconds` and `MISSING_TRANSITION_TARGET`).

Two things measured that stay unchanged, both same verdict on either side. A JSONPath-only field carrying an expression (`Result`, `Parameters`, `ResultSelector`, `InputPath`, `ItemsPath`, `Choices[].Variable`) draws an `UNSUPPORTED_JSONATA_EXPRESSION` here where AWS reports only its schema error; suppressing it would drop the only diagnostic Floci has for `Choices[].Variable`, so the extra diagnostic is the safer of the two. And a `Map` state whose own `QueryLanguage` is `JSONata` inside a JSONPath machine does not propagate that to its `ItemProcessor` states, so nothing there is checked; that is `validateNestedStates` passing `topLevelJsonata`, which predates this branch.

pgermosen (approved):
Went deep on the reflection-based parse-tree walk since it's the novel piece here, checking it against the actual pinned JSONata library source rather than taking the description on faith. Every field it reads reflectively really exists on the tree node type, the context-item detection lines up exactly with how the tokenizer builds `$`/`$$`/`$name` values, and the decision to only look at a path's first step (never descending into filters, sort terms, or groupings) is correct because those really are separate node shapes in the parser — which is exactly why a reference inside a predicate is legal on real AWS while one at the top level isn't. I also traced the call graph and confirmed CreateStateMachine, UpdateStateMachine, and ValidateStateMachineDefinition all genuinely share the same validation path, so this can't end up enforced on only one of them, and the ASL-field deny list correctly stops applying once the walk enters a payload field at any depth, which is what fixes the payload-key-named-after-an-ASL-field gap from the second commit.

This is a breaking change with real teeth (32% of the measured corpus), but the differential work backing that number is thorough, and the one disclosed miss is in the safe direction — a definition Floci still accepts that AWS would refuse, not the reverse. Nothing outstanding.

PR #2709, opened 2026-08-29
fix(rds): propagate ModifyDBInstance master-password rotations to the backend and proxy

matanbaruch:
## Summary

After `ModifyDBInstance` rotates the master password, no client can connect to the instance at all. Rotation only updated Floci's stored state, leaving two layers stale:

1. The backend DB container keeps the original credential (`MYSQL_ROOT_PASSWORD`/`MYSQL_PASSWORD` only take effect on first initialization), so a connection carrying the rotated password fails there.
2. The running auth proxy validates the master scramble against the password it was **started** with, so the rotated password fails proxy-side too until something restarts the proxy.

Either way the user is locked out: old password fails the proxy's live validator, new password fails one of the two stale layers. First disclosed on #2679.

The fix acts at rotation time, the last moment the old credential is still known:

- **Backend:** exec into the DB container and change the password there. MySQL/MariaDB run `SET PASSWORD` **as the master user itself with the old password**: changing your own password needs no privileges, and root's password is the creation-time one, not reliably known after earlier rotations. PostgreSQL runs `ALTER ROLE` over the trusted local socket. Engine-specific syntax (`SET PASSWORD = '...'` vs MariaDB's `PASSWORD('...')` form), passwords and role names escaped.
- **Proxy:** swap the password snapshot on the running proxy in place. A stop/start here races the listener rebind (closing the server socket defers the port release while relay connections are still draining, so the immediate rebind hits EADDRINUSE, which is what the first CI run caught in `RdsJdbcCompatTest`), and it would also drop live connections that rotation shouldn't touch.

Mock mode, cluster members, and instances without a running container are unaffected (cluster password rotation has the same gap and is intentionally not covered here, can follow up separately).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: on real RDS, `ModifyDBInstance --master-user-password` makes the new password usable (asynchronously, usually within seconds) and invalidates the old one. On Floci the rotated password never became usable:

```
$ aws rds modify-db-instance --db-instance-identifier mydb --master-user-password new... --apply-immediately
$ mysql -h <endpoint> -uadmin -p<new> ...
ERROR 1045 (HY000): Access denied for user 'admin'@'localhost' (using password: YES)
```

Verified against a JVM build with real backend containers, all through the auth proxy:

| Engine | rotate → connect with new password | old password | connection open across the rotation | rotate → reboot → new password |
|---|---|---|---|---|
| mysql:8.0.36 | works (failed before) | rejected (1045) | stays connected | works |
| mariadb:11 | works (failed before) | rejected | n/a | n/a |
| postgres:16 | works (failed before) | rejected | stays connected | n/a |

## Checklist

- [x] `./mvnw test` passes locally (`RdsServiceTest` 249, `RdsContainerManagerTest` 30, `RdsProxyManagerTest` 9)
- [x] New or updated integration test added: `modifyDbInstancePasswordRotationPropagatesToBackendAndProxy` (backend exec with old+new credentials, in-place proxy snapshot update, no restart), same-password and mock-mode no-op tests, `updateMasterPassword` manager tests, and per-engine command/SQL builder tests including escaping

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/container/RdsContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/container/RdsContainerManager.java
@@ -408,6 +408,52 @@ public class RdsContainerManager {
     }
 
     /**
+     * Propagates a ModifyDBInstance master-password rotation into the running DB container, so
+     * the backend's stored credential matches what clients (and the auth proxy) use from now on.
+     * For MySQL/MariaDB this runs as the master user itself with the old password — changing your
+     * own password needs no extra privileges, and the container's root password is the
+     * creation-time one, not reliably known after earlier rotations. PostgreSQL's official image
+     * trusts local socket connections, so psql needs no password at all.
+     */
+    public void rotateMasterPassword(String instanceId, String containerId, DatabaseEngine engine,
+                                     String masterUsername, String oldPassword, String newPassword) {
+        execUntilSuccess(instanceId, containerId,
+                passwordRotationCommand(engine, masterUsername, oldPassword, newPassword),
+                "master-password rotation");
+    }
+
+    static String[] passwordRotationCommand(DatabaseEngine engine, String masterUsername,
+                                            String oldPassword, String newPassword) {
+        if (engine == DatabaseEngine.POSTGRES) {
+            String effectiveUser = (masterUsername != null && !masterUsername.isBlank())
+                    ? masterUsername : "postgres";
+            return new String[]{"psql", "-v", "ON_ERROR_STOP=1", "-U", effectiveUser, "-d", "postgres",
+                    "-c", postgresPasswordRotationSql(effectiveUser, newPassword)};
+        }
+        String effectiveUser = (masterUsername != null && !masterUsername.isBlank())
+                ? masterUsername : "root";
+        // MariaDB images ≥10.4 (Floci's floor is 10.11) ship only the `mariadb` client binary.
+        String client = engine == DatabaseEngine.MARIADB ? "mariadb" : "mysql";
+        return new String[]{client, "-u" + effectiveUser, "-p" + oldPassword,
+                "-e", mysqlPasswordRotationSql(engine, newPassword)};
+    }
+
+    static String mysqlPasswordRotationSql(DatabaseEngine engine, String newPassword) {
+        String escaped = newPassword.replace("\\", "\\\\").replace("'", "\\'");
+        // MariaDB only accepts the PASSWORD() form; MySQL 8 only the plain literal.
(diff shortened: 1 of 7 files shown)


matanbaruch:
CI caught a stop/rebind race in the proxy restart (the closed listener's port isn't released before the immediate rebind, EADDRINUSE in RdsJdbcCompatTest). Switched to swapping the password snapshot on the running proxy in place: no socket churn, and connections open across the rotation stay alive. Re-verified on postgres:16 and mysql:8.0.36, PR body updated.

src/main/java/io/github/hectorvent/floci/services/rds/container/RdsContainerManager.java
```diff
@@ -407,6 +407,52 @@ private void initializePostgresIamRole(String containerName, String containerId,
...
+     * Propagates a ModifyDBInstance master-password rotation into the running DB container, so
+     * the backend's stored credential matches what clients (and the auth proxy) use from now on.
+     * For MySQL/MariaDB this runs as the master user itself with the old password — changing your
+     * own password needs no extra privileges, and the container's root password is the
+     * creation-time one, not reliably known after earlier rotations. PostgreSQL's official image
+     * trusts local socket connections, so psql needs no password at all.
+     */
+    public void rotateMasterPassword(String instanceId, String containerId, DatabaseEngine engine,
+                                     String masterUsername, String oldPassword, String newPassword) {
+        execUntilSuccess(instanceId, containerId,
```
pgermosen, line 420: This passes instanceId into execUntilSuccess's containerName param, but the existing call site (initializeEngine) passes the real Docker container name there. The log line ends up saying "...in RDS container <instance-id>" instead of the actual container name, which'll be confusing if someone's debugging exec failures. Not urgent, just an inconsistency with the existing convention.

pgermosen (approved):
Approving. The in-place proxy swap is the right call, not just a fix for the EADDRINUSE race CI caught — AWS's own ModifyDBInstance docs say changing the master password "doesn't result in an outage," so keeping the listener and live connections untouched is the AWS-correct behavior, not a workaround. That also structurally closes greptile's restart-failure concern since there's no restart left at all. Left one small non-blocking comment below.

src/main/java/io/github/hectorvent/floci/services/rds/container/RdsContainerManager.java
matanbaruch: Good catch. Will fix it in the cluster-rotation follow-up, same code path.

PR #1011, opened 2026-05-24
feat(apigateway): add Lambda REQUEST authorizer support for HTTP API v2 routes

dixitrathod16:
## Summary

Implement Lambda REQUEST authorizer invocation for HTTP API (v2) routes configured with `authorizationType=CUSTOM`. Previously, Floci silently skipped authorization for these routes — which is incorrect behavior.

Closes #812

### What's supported:

- **Payload format version 1.0** — REST API-compatible event shape (`methodArn`, `headers`, `multiValueHeaders`, `queryStringParameters`, `multiValueQueryStringParameters`, `pathParameters`, `stageVariables`, `requestContext`)
- **Payload format version 2.0** — HTTP API-native event shape (`routeArn`, `routeKey`, `rawPath`, `rawQueryString`, lowercase headers, `requestContext.http`)
- **Simple responses** — `{isAuthorized: true/false}` when `enableSimpleResponses=true`
- **IAM policy document responses** — `policyDocument` with `Allow`/`Deny` effect
- **Identity source validation** — `$request.header.X` and `$request.querystring.X` expressions; returns 401 without invoking Lambda when any configured source is missing

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore

## AWS Compatibility

Behavior matches the [AWS documentation for HTTP API Lambda authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html):

- Identity source validation returns 401 before invoking the Lambda (same as AWS)
- Payload format 1.0 produces the same event shape as REST API REQUEST authorizers
- Payload format 2.0 produces the HTTP API-native shape with `routeArn`, `routeKey`, `rawPath`, `requestContext.http`
- Simple responses (`enableSimpleResponses=true`) accept `{isAuthorized: boolean}` format
- Lambda errors (function exceptions) return 500 (same as AWS)
- Deny policy returns 403 (same as AWS)

## Checklist

- [x] `./mvnw test` passes locally
- [x] New or updated integration test added
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

Issue #812, opened 2026-05-12
[FEAT] API Gateway v2 HTTP API — Lambda REQUEST authorizer support (payload format 1.0 and 2.0)


dixitrathod16:
## Service

API Gateway v2 (HTTP APIs)

## API Action / Feature

Lambda REQUEST authorizer invocation for HTTP API routes (`--authorizer-type REQUEST` via `aws apigatewayv2 create-authorizer`)

## AWS Documentation

- [Control access to HTTP APIs with AWS Lambda authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html)
- [Payload format versions 1.0 and 2.0](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.payload-format)

## Why is this needed?

Real AWS supports Lambda REQUEST authorizers on HTTP APIs (v2) in addition to JWT authorizers. Floci's `dispatchV2` path in `ApiGatewayExecuteController` only enforces JWT authorizers. When a route is configured with `authorizationType: CUSTOM` and a Lambda REQUEST authorizer, Floci silently skips authorization and allows the request through — which is incorrect behavior.

AWS supports two payload format versions for HTTP API Lambda authorizers:

**Format 1.0** — compatible with REST API (v1) REQUEST authorizer shape:
```json
{
  "version": "1.0",
  "type": "REQUEST",
  "methodArn": "arn:aws:execute-api:us-east-1:123456789012:abcdef123/test/GET/pets",
  "identitySource": "...",
  "authorizationToken": "...",
  "resource": "/pets",
  "path": "/pets",
  "httpMethod": "GET",
  "headers": { "Header1": "value1" },
  "queryStringParameters": { "parameter1": "value1" },
  "pathParameters": {},
  "stageVariables": {},
  "requestContext": { ... }
}
```

**Format 2.0** — newer HTTP API-native shape:
```json
{
  "version": "2.0",
  "type": "REQUEST",
  "routeArn": "arn:aws:execute-api:us-east-1:123456789012:abcdef123/test/GET/pets",
  "identitySource": "...",
  "routeKey": "GET /pets",
  "rawPath": "/pets",
  "rawQueryString": "parameter1=value1",
  "headers": { "Header1": "value1" },
  "queryStringParameters": { "parameter1": "value1" },
  "pathParameters": {},
  "stageVariables": {},
  "requestContext": {
[...]

PR #581, opened 2026-04-24
fix(apigateway): propagate TOKEN authorizer context to AWS_PROXY Lambda

hampsterx:
Closes #574.

## What

REST API Gateway v1 TOKEN authorizers return a `principalId` + `context` map that floci silently dropped, so downstream `AWS_PROXY` Lambdas saw an empty `event.requestContext.authorizer`. LocalStack propagates both; this brings floci to parity.

Also fixes the `methodArn` in the authorizer event, previously a literal `"api/stage/METHOD/path"` string, now the real `arn:aws:execute-api:<region>:<account>:<apiId>/<stage>/<METHOD>/<path>` built from the incoming request so policy-scoping authorizers can match.

## Changes

- Introduce a small `AuthorizerResult` record carrying `errorResponse`, `principalId`, and `context`. `invokeAuthorizer` now returns it; `dispatch` short-circuits on `errorResponse`, otherwise threads `principalId` + `context` into `invokeProxy` → `buildProxyEvent`.
- `buildProxyEvent` emits `requestContext.authorizer` only when a `CUSTOM` authorizer actually ran, matching AWS behavior (no injected empty map on unprotected routes).
- Context values are stringified at emit time to match AWS's flat-string wire contract. `null` values are omitted (not emitted as JSON `null`).
- Replace the placeholder `methodArn` with a real one built from the request-resolved region, API ID, stage, the raw request HTTP method, and the incoming path.

## Test coverage

New `ApiGatewayAuthorizerContextIntegrationTest` (`@QuarkusTest`) covers:

- **Happy path**: TOKEN authorizer returns `{principalId, context:{org_id, sub, client_id, methodArn}}` → proxy Lambda receives all fields at `event.requestContext.authorizer`, plus a real `methodArn` of the form `arn:aws:execute-api:<region>:<account>:<apiId>/<stage>/<METHOD>/<path>`.
- **Unprotected route** (`authorizationType: NONE`): proxy Lambda sees no `authorizer` key in `requestContext`.
- **Deny**: authorizer returns `Effect: Deny` → 403, proxy Lambda is not invoked.

`./mvnw -q -DskipTests test-compile` passes locally. The focused integration test is not runnable in my local harness due to an unrelated `ContainerDetector.hasMountInfoMarkers()` false-positive on Linux hosts with Docker installed; CI runs inside a container and should exercise the test. Happy to follow up with a separate PR for the detector if useful.

## Out of scope

- v2 HTTP API JWT authorizer context (different event shape, separate code path).
[...]

PR #2715, opened 2026-08-29
fix(apigatewayv2): propagate Lambda REQUEST authorizer context to HTTP API backends

bsaurusrex:
## Summary

Floci invokes a Lambda REQUEST authorizer on an HTTP API route, but discards the `context` the authorizer returns. `enforceRequestAuthorizerV2` returns a bare `Response` (null meaning "allowed"), so there is nothing to carry the context to `buildV2ProxyEvent`, and the backend Lambda receives no `requestContext.authorizer` at all.

Real API Gateway delivers it under `requestContext.authorizer.lambda` — that is how an authorizer hands verified identity to the integration — so a backend that reads it sees an empty object and rejects requests the authorizer has just allowed.

This completes #1011 (issue #812), which added REQUEST authorizer invocation for HTTP APIs but wired only the allow/deny decision, and brings v2 in line with #581, which propagated the authorizer context on the v1/REST path. The WebSocket `$connect` path already does the same via `WebSocketProxyEventBuilder`, so HTTP APIs were the last of the three to drop it.

The gap had been written down as though it were AWS's format — an existing comment in `HttpApiRequestAuthorizerTest` read *"the backend Lambda receives the proxy event which does NOT include authorizer context in the v2 proxy event format"*, which is not true of AWS. Corrected here.

### What changed

- `enforceRequestAuthorizerV2` returns a `RequestAuthorizerResult` carrying the context — mirroring the `JwtAuthorizerResult` the JWT path already threads through — instead of a bare `Response`.
- `buildV2ProxyEvent` renders it as `requestContext.authorizer.lambda`.

Values are **not** flattened to strings the way the v1/REST path flattens its authorizer context: an HTTP API hands the context to the integration as JSON. The [developer guide's simple-response example](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html) returns a string, a number, a boolean, an array and a map, and all five now survive to the backend.

### Deliberate omissions

Both because I could not measure the real AWS behaviour — happy to change either if a maintainer knows the answer:

- `principalId` is not surfaced on the proxy event. A policy-document response's `principalId` reaches `$context.authorizer.principalId` for access logging and parameter mapping; where AWS places it in the 2.0 proxy event, if anywhere, I could not confirm.
- An authorizer that allows without returning a context leaves the `authorizer` node absent rather than rendering `"lambda": null`.

Not covered: `HTTP_PROXY` integrations behind a Lambda authorizer — `dispatchHttpProxyV2` still receives only JWT claims.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior:** a route with `authorizationType: CUSTOM` invoked its authorizer and honoured the allow/deny decision, but the integration Lambda received no `requestContext.authorizer` node, so every context value the authorizer returned was lost.

Verified against the documented payload-format-2.0 shapes in [Control access to HTTP APIs with AWS Lambda authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html), for both the simple-response and IAM-policy response formats.

Also exercised end to end outside the test suite, against a real 2529-line OpenAPI document (132 routes, 132 integrations, one REQUEST authorizer with `enableSimpleResponses` and `identitySource: $request.header.Authorization,$context.routeKey`): an authorized request now reaches the backend with the authorizer's nested context intact, a missing identity source still returns 401 without invoking the authorizer, and a denial still returns 403.

## Checklist

- [x] `./mvnw test` passes locally — 13952 tests, 0 failures, 0 errors, 12 skipped

Tests: 4 cases added to `HttpApiRequestAuthorizerTest` (IAM-policy context, simple-response context, nesting and scalar types, context-less allow) and a new `BuildV2ProxyEventLambdaAuthorizerTest` (4 cases) pinning the rendered shape without needing a Lambda container.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -1479,9 +1479,12 @@ public class ApiGatewayExecuteController {
             jwtScopes = jwtResult.scopes();
         }
 
+        ObjectNode lambdaAuthorizerContext = null;
         if ("CUSTOM".equalsIgnoreCase(route.getAuthorizationType()) && route.getAuthorizerId() != null) {
-            Response authError = enforceRequestAuthorizerV2(region, apiId, stageName, route, httpMethod, path, headers, uriInfo);
-            if (authError != null) return authError;
+            RequestAuthorizerResult requestResult =
+                    enforceRequestAuthorizerV2(region, apiId, stageName, route, httpMethod, path, headers, uriInfo);
+            if (requestResult.errorResponse() != null) return requestResult.errorResponse();
+            lambdaAuthorizerContext = requestResult.context();
         }
 
         if (route.getTarget() == null) {
@@ -1520,7 +1523,8 @@ public class ApiGatewayExecuteController {
 
         String requestId = UUID.randomUUID().toString();
         String eventJson = buildV2ProxyEvent(httpMethod, path, route.getRouteKey(),
-                apiId, region, stageName, headers, uriInfo, body, requestId, jwtClaims, jwtScopes);
+                apiId, region, stageName, headers, uriInfo, body, requestId, jwtClaims, jwtScopes,
+                lambdaAuthorizerContext);
 
         LOG.debugv("execute-api v2: {0} {1}/{2}{3} → Lambda {4}", httpMethod, apiId, stageName, path, functionName);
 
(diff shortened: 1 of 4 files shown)


pgermosen (approved):
Approving. Checked the five value types against the real developer guide example and they match exactly, and the authorizer.lambda nesting follows the same authorizer.<type> convention this codebase already ships and verifies for JWT (authorizer.jwt.claims). Good to see the two genuinely uncertain bits (principalId placement, context-less-allow shape) called out explicitly rather than guessed at. Well tested on both response formats and the JWT/Lambda mutual-exclusivity case.

PR #2723, opened 2026-08-29
feat(codepipeline): drive CodeBuild with env/secondary artifacts, GitHub source actions, EventBridge/SNS events, and V2 conditions

sam-fakhreddine:
## Dependency

**This PR is stacked on #2707 (`feat(codebuild): harden build execution with retries and source/CA integrity`) and cannot be reviewed or merged until #2707 lands.** It is built directly on #2707's branch tip, so until #2707 merges, this PR's diff shows all of #2707's commits in addition to the 10 CodePipeline-specific commits below. Once #2707 merges into `main`, this PR's diff will automatically shrink to just those 10 commits. Opened as **draft** for that reason.

## Summary

Ten commits extending the CodePipeline emulator on top of #2707's CodeBuild hardening:

- Execute ThirdParty GitHub (version 1) source actions: fetch the branch archive from `codeload.github.com` and repackage it with the repo contents at the artifact root, matching the real action's output layout.
- Hand artifacts to CodeBuild via `StartBuild` overrides (env vars, secondary source/artifacts) instead of only S3-staged primary input.
- Preserve unix file modes and symlinks when repackaging GitHub source zips (`node_modules/.bin/*`-style tooling depends on both).
- Emit `aws.codepipeline` EventBridge state-change events (pipeline/stage/action) and SNS manual-approval notifications, matching real CodePipeline shapes.
- Evaluate V2 conditions (`beforeEntry`/`onSuccess`/`onFailure`) with real `LambdaInvoke`/`VariableCheck` rule evaluation, condition-driven retry, and automatic rollback on `onFailure.result: ROLLBACK`.
- Restore pipeline artifacts on retry (previously lost across `RetryStageExecution`).
- Publish pipeline state-change events to the pipeline's own account's EventBridge bus, not the default/management account — the worker thread that runs pipeline executions has no request context, so the two-argument `putEvents` overload silently fell back to the wrong account.
- Reject a `retryMode` outside the `StageRetryMode` enum (`FAILED_ACTIONS`/`ALL_ACTIONS`) instead of silently treating any unrecognized value as the narrower retry.

## Wire-fidelity

Re-verified the new `RetryStageExecution` `retryMode` validation against the CodePipeline botocore model: `StageRetryMode` is `FAILED_ACTIONS | ALL_ACTIONS`, matching the new `RETRY_MODES` check. No new gaps found against the wire-fidelity corpus (92 existing CodePipeline entries; consistent with the earlier #2696 triage that the relevant validation surface is already covered).

## Local-review findings

Ran `review-local.ts` against the 3 non-test files this branch changes. In-scope findings (introduced by this branch's own commits) were fixed RED-first in a follow-up commit:

- **GitHub source action had no URL validation or network timeout.** `executeGitHubSource` concatenated pipeline-declared `Owner`/`Repo`/`Branch` directly into the `codeload.github.com` URL with no timeout on the fetch. Added an allowlist check (rejects `..` traversal, a leading `/`, and whitespace, while still permitting legitimate branches like `release/v1.16.0`) plus connect/request timeouts.
- **`VariableCheck` `MATCHES` operator had no bound on regex/subject length.** Both the pattern (`Value`) and the subject (`Variable`) are pipeline-declared and were passed straight to `String.matches` with no length cap — a catastrophic-backtracking pattern could pin a thread. Added a 256-character cap that fails the rule with a `ValidationException`-style message instead.
- **`CodePipelineEventPublisher` (new file) dropped an action's `summary` from the emitted event** unless `externalExecutionId` or `errorDetails` was also set, and **stringified a missing error-details `code` into the literal `"null"`.** Fixed both; added `CodePipelineEventPublisherTest` pinning the execution-result shape.

**Disputed (no code change):** the review also flagged a HIGH "zip-slip in `stripTopLevelDirectory`." Verified against the actual consumer — `CodeBuildRunner.extractZip` already normalizes each target path and rejects anything that escapes the extraction root (`target.startsWith(dest)`) before writing to disk. `stripTopLevelDirectory` only renames zip entries in memory; it never writes to the filesystem itself. This is a false positive from the isolated-file review not seeing the real extraction site's guard.

## Deliberate scope notes

- Several other real findings from the same review pass are in `CodePipelineService.java` code this branch did not touch (pre-existing from earlier PRs): unsynchronized cross-thread execution-state mutation, `waitForApproval`/`waitForCustomJob`/transition waits with no timeout, `completeJob` accepting unacknowledged jobs, `SUPERSEDED` mode not actually stopping the runner, a `clientRequestToken` idempotency race/NPE, and a few LOW items (unbounded `pipelineLocks` growth, `ActionExecution.token` missing `@JsonIgnore`). These are ledgered in `issues/codepipeline-execution-engine-preexisting-gaps.md` (most were already tracked there from an earlier pass; a short addendum was added for the handful that weren't) rather than fixed here, to keep this PR scoped to the 10 commits it's actually introducing.
- `stripTopLevelDirectory`/GitHub archive fetch has no entry-count or total-decompressed-size cap (zip-bomb resource exhaustion). Not fixed here — same category as the pre-existing timeout gaps above, tracked in the same issues file rather than expanding this diff.
- The manual-approval SNS event advertises a 7-day `expires` timestamp for shape fidelity; nothing currently enforces that expiry (the approval wait loop has no timeout at all, per the pre-existing gaps above). The field is emitted because real CodePipeline emits it, not because expiry is enforced.

## Type of change

## AWS compatibility

Behavior is derived from the CodePipeline/CodeBuild botocore service models and real CodePipeline event/action shapes (GitHub source action output layout, EventBridge/SNS event shapes, `StageRetryMode` enum).

## Checklist

- [x] `mvn clean test-compile` passes
- [x] Targeted suites green: `CodePipelineIntegrationTest`, `CodePipelineServiceTest`, `CodePipelineEventsIntegrationTest`, `CodePipelineCfnIntegrationTest`, `CodeBuildCfnProvisionerTest`, `CodePipelineCfnProvisionerTest`, `CodePipelineEventPublisherTest` — 50/50
- [x] No AI attribution in commits or this PR


sam-fakhreddine:
This one is stacked on #2707 — its diff is a superset of #2707's CodeBuild/docker-retry work plus CodePipeline env/secondary-artifact additions on top. Once #2707 merges, #2723's diff against main will shrink to just the CodePipeline delta. Draft until then.

sam-fakhreddine:
Removed in d05ebd0dd: the two issues/ ledger files this PR shared with #2707 were fork-local tracking notes and are now kept out of the PR, same as on the sibling PRs.

sam-fakhreddine:
Cherry-picked #2707's retry environment fix here as e9deb48b7 so the shared `CodeBuildRunner` stays byte-identical between the two branches; `CodeBuildRunnerEnvAssemblyTest` 9/9 green on this branch.

KenkoGeek (changes requested):
**High - this branch needs to be split before it is reviewable as a focused contribution.**

Even after removing the #2707 stack, the CodePipeline-specific portion contains several independently meaningful AWS surfaces: ThirdParty GitHub source actions and archive handling, CodeBuild artifact/override handoff, EventBridge state-change events, SNS manual-approval notifications, V2 condition evaluation, retry semantics, rollback behavior, artifact restoration, and account-aware event publishing.

`AGENTS.md` requires narrow, focused, testable changes. These areas have different AWS contracts, different regression risks, and can be validated independently. Keeping them together makes it much harder to determine whether a failure belongs to source actions, event delivery, retry/rollback state transitions, or CodeBuild integration.

Please split the CodePipeline work into focused PRs, for example:
1. GitHub source action + archive fidelity/security.
2. CodeBuild artifact/override handoff.
3. EventBridge/SNS event publishing.
4. V2 conditions and rule evaluation.
5. Retry/rollback/artifact restoration semantics.

This PR is also stacked on #2707, where I already requested a structural split. Until that dependency is reorganized, this branch cannot have a stable review base.

PR #2726, opened 2026-08-29
fix: repair main CI (AslExecutor test compilation, Redshift compat test endpoint)

matanbaruch:
## Summary

Two independent CI breakages on current `main`, both repaired here.

**1. Test sources don't compile.** #2703 added `AslExecutorTaskHistoryEventsTest` and #2688 added a `CustomResourceLiveness` parameter to the `AslExecutor` constructor. Each PR was green on its own, but merged together the test calls the constructor with the old argument list, so `mvn test` fails at `testCompile` on all four Build and Test shards:

```
AslExecutorTaskHistoryEventsTest.java:[64,20] constructor AslExecutor in class AslExecutor cannot be applied to given types;
```

One-line fix: pass a mock for the new parameter, the same way the test mocks the constructor's other collaborators.

**2. `native / sdk-test-java` fails deterministically on the new Redshift compat test.** #2472's `RedshiftTest` builds its own client against a hardcoded `http://localhost:4566` and dials the DescribeClusters endpoint address directly, but the harness serves Floci at `FLOCI_ENDPOINT` (a non-localhost host in CI), so both tests fail with connection refused on every run (verified on two runs):

```
SdkClientException: Unable to execute HTTP request: Connect to http://localhost:4566 failed: Connection refused (SDK Attempt Count: 4)
```

Fix: use the suite's `TestFixtures.redshiftClient()` (added alongside the test but unused) and `TestFixtures.proxyHost()` for the JDBC URL, the same pattern `RdsJdbcCompatTest` uses.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

No behavior change: both fixes are test-only. `AslExecutorTaskHistoryEventsTest` passes (5 tests) once it compiles, and `RedshiftTest` reaches Floci through the same endpoint resolution every other compat test uses.

## Checklist

- [x] `./mvnw test` passes locally (`AslExecutorTaskHistoryEventsTest` 5 tests, previously not compiling; sdk-test-java module `test-compile` clean)
- [x] New or updated integration test added: not applicable, this repairs existing tests

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
@@ -14,12 +14,8 @@ import software.amazon.awssdk.services.redshift.model.DescribeClustersResponse;
 import software.amazon.awssdk.services.redshift.model.DeleteClusterRequest;
 import software.amazon.awssdk.services.redshift.model.DeleteClusterResponse;
 
-import java.sql.Connection;
-import java.sql.DriverManager;
-
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 public class RedshiftTest {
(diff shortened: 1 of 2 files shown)


matanbaruch:
Expanded scope: the sdk-test-java failure turned out to be a second main breakage, the new RedshiftTest hardcodes localhost:4566 instead of using TestFixtures (verified deterministic on two runs). Fixed here too so this PR fully greens main.

matanbaruch:
Round 2 on the Redshift test: with the endpoint fix the API assertions pass, but the JDBC hop still fails. DescribeClusters returns the backing container's own host:port and Redshift has no RDS-style auth proxy, so that connection can't succeed when Floci runs in a container. Dropped the JDBC hop with a note (API coverage kept). If there's appetite for a Redshift auth proxy like RDS's, I can file an issue.

hectorvent (approved):
Thank you!

PR #2727, opened 2026-08-29
fix(rds): propagate ModifyDBCluster master-password rotations to the backend and every proxy

matanbaruch:
## Summary

Follow-up to #2709, which fixed master-password rotation for standalone instances and disclosed that clusters have the same gap. `ModifyDBCluster --master-user-password` only updated Floci's stored state, leaving three layers stale:

1. The backend DB container keeps the original credential.
2. The cluster endpoint's auth proxy holds a start-time password snapshot.
3. Every member endpoint's auth proxy holds its own start-time snapshot, and member instances validate against the member's stored password.

So after a cluster rotation no endpoint accepted the new password. The fix reuses the #2709 mechanism at rotation time, while the old credential is still known:

- **Backend:** exec the password change into the cluster's container (`SET PASSWORD` as the master user for MySQL/MariaDB, `ALTER ROLE` over the trusted socket for PostgreSQL).
- **Proxies:** swap the password snapshot in place on the cluster proxy AND on each member instance's proxy, updating the members' stored passwords to match. AWS applies the cluster master password to every member endpoint, and the in-place swap keeps live connections open (rotation causes no outage).

`modifyDbCluster` is now `synchronized` like `modifyDbInstance` and `deleteDbCluster`, protecting the read-old-credential-then-exec window. Mock mode and clusters without a running container are unaffected.

Also addresses the #2709 review note: `rotateMasterPassword` now receives the Docker container name (not the instance id) for `execUntilSuccess` log lines, matching the other call sites.

Note: current `main` fails `testCompile` on an unrelated cross-PR conflict (#2703 + #2688); fix in #2726. This PR's CI needs that to land first.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: on real Aurora, `ModifyDBCluster --master-user-password` makes the new password usable on the cluster endpoint and every instance endpoint, without an outage. On Floci the rotated password never became usable anywhere:

```
$ aws rds modify-db-cluster --db-cluster-identifier mycluster --master-user-password new... --apply-immediately
$ psql "postgresql://admin:<new>@<cluster-endpoint>/db"
FATAL:  password authentication failed for user "admin"
```

Verified against a JVM build with a real backend container (aurora-postgresql cluster + one member instance, postgres:16), all through the auth proxies:

| Endpoint | rotate → new password | old password | connection open across the rotation |
|---|---|---|---|
| cluster (writer) | works (failed before) | rejected | stays connected |
| member instance | works (failed before) | rejected | n/a |

One observed difference in error shape: a member created without `MasterUsername` gets the "root" default, so connections as the cluster's master user take the member proxy's non-master path (backend is the authority) and the old password fails with `Backend database authentication failed` instead of the proxy-side `password authentication failed`. Members created with the master username engage the validator/snapshot paths this PR updates.

## Checklist

- [x] `./mvnw test` passes locally (`RdsServiceTest` 252, `RdsContainerManagerTest` 30, `RdsProxyManagerTest` 9)
- [x] New or updated integration test added: `modifyDbClusterPasswordRotationPropagatesToBackendAndEveryProxy` (backend exec with old+new credentials, snapshot swap on the cluster proxy and the member proxy, member stored password updated, no proxy restart), plus same-password and mock-mode no-op tests

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
@@ -1089,7 +1089,8 @@ public class RdsService implements Resettable, ResourceProvider {
             // holds) is known. Without it every later connection fails: the proxy dials the
             // backend with the rotated password against a database still holding the original.
             if (passwordRotated) {
-                containerManager.rotateMasterPassword(id, instance.getContainerId(),
+                containerManager.rotateMasterPassword(
+                        instance.getDockerVolumeName(), instance.getContainerId(),
                         instance.getEngine(), instance.getMasterUsername(), oldPassword, newPassword);
             }
             instance.setMasterPassword(newPassword);
@@ -1648,7 +1649,7 @@ public class RdsService implements Resettable, ResourceProvider {
         return modifyDbCluster(id, newPassword, iamEnabled, null, null, null, region);
     }
 
-    public DbCluster modifyDbCluster(String id, String newPassword, Boolean iamEnabled,
+    public synchronized DbCluster modifyDbCluster(String id, String newPassword, Boolean iamEnabled,
                                      Double serverlessV2MinCapacity, Double serverlessV2MaxCapacity,
                                      Integer serverlessV2SecondsUntilAutoPause, String region) {
         String effectiveRegion = effectiveRegion(region);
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Approving. Same mechanism as #2709, extended correctly to clusters and every member instance — old password captured before the exec, backend updated while it's still known, then the cluster proxy and each member proxy swapped in place with no restart. Also picks up the log-identity fix from my #2709 note. Checked ModifyDBCluster's docs and the "applied immediately regardless of scheduling" language for master-password changes lines up with the no-outage design here.

PR #2728, opened 2026-08-29
perf(cloudformation): probe Lambda S3 code with headObject, not getObject

### LINKED ISSUE #2675: [BUG] cloudformation: Lambda S3 deployment package is read twice on every stack operation
## Service

CloudFormation (provisioning `AWS::Lambda::Function`) / Lambda

## AWS API Action

CreateStack / UpdateStack

## Expected behavior

Provisioning a Lambda from a template that points at `Code: { S3Bucket, S3Key }` should read that
object from S3 once.

## Actual behavior

It is read twice, in full, on every stack operation.

`CloudFormationResourceProvisioner.resolveLambdaCode` calls `s3Service.getObject(s3Bucket, s3Key)`
purely as a readability probe and **discards the returned object**, keeping only the bucket/key pair:

```java
try {
    s3Service.getObject(s3Bucket, s3Key);          // result thrown away
    return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key), ...);
} catch (Exception e) {
    ...
}
```

`LambdaService.extractZipCodeFromS3` then fetches the very same object again to do the real work:

```java
obj = s3Service.getObject(s3Bucket, s3Key);
extractZipCodeBytes(fn, obj.getData(), region);
```

`S3Service.getObject` returns a materialised `S3Object` whose `getData()` is the full byte array, so
this is a complete second copy of the package read and held in memory, not a cheap metadata lookup.
For a package at the upper end of what Floci accepts that is a substantial amount of avoidable I/O
and heap per stack create/update, repeated for every `AWS::Lambda::Function` in the template.

Nothing is functionally wrong — the second read produces the same bytes — so this is a performance
and memory-footprint issue rather than a correctness one.

## Reproduction

Any stack with an `AWS::Lambda::Function` whose `Code` is an S3 bucket/key:

```yaml
Resources:
  Fn:
    Type: AWS::Lambda::Function
    Properties:
      FunctionName: double-read-repro
      Runtime: nodejs20.x
      Handler: index.handler
      Role: arn:aws:iam::000000000000:role/r
      Code:
        S3Bucket: pkg
        S3Key: fn.zip
```

```bash
  --stack-name double-read-repro --template-body file://stack.yaml
```

`S3Service.getObject` is invoked twice for `pkg/fn.zip` during the one operation — once from
`resolveLambdaCode`, once from `extractZipCodeFromS3`.

## Suggested direction

Either:

1. Replace the probe with an existence/metadata check that does not materialise the body, so the
   full read happens only in `LambdaService`; or
2. Thread the already-fetched bytes through `LambdaCodeSpec` so the object is read once and reused.

(1) is the smaller change and keeps the layering as-is; (2) removes the read entirely at the cost of
carrying the payload through the code spec. Happy to follow whichever maintainers prefer.

## Notes

Split out of #2648 / PR #2650 at a maintainer's suggestion — @pgermosen confirmed on that PR that it
is "a real, separate concern and agreed it belongs in its own issue rather than this PR". The probe
itself is what #2650 changes the *failure* behaviour of; this issue is only about it reading the
object twice.

## Environment

- Java SDK version (if applicable): aws-cli v2
- How you're running Floci (Docker / native / `mvn quarkus:dev`): source checkout, JDK 25.0.4

- [ ] No

avison9:
## ⚠️ CI on this PR will fail for a reason unrelated to the change

`main` does not currently compile its test sources, and has not since `08e532ab`. Every branch cut
from it inherits that, so the shards here die before running anything:

```
AslExecutorTaskHistoryEventsTest.java:[64,20]
  constructor AslExecutor cannot be applied to given types
```

That is a semantic merge conflict between #2688, which added a `CustomResourceLiveness` constructor
parameter, and #2703, which added a test constructing `AslExecutor` directly. I have opened a
separate one line PR to restore compilation; once that merges this branch will go green on a rebase.
Nothing here touches Step Functions.

I verified this change locally by applying that one line fix uncommitted, purely so the module would
build. Full suite on this branch with that in place:
`Tests run: 14235, Failures: 0, Errors: 0, Skipped: 9`.

---

## Summary

Closes #2675.

Resolving an `AWS::Lambda::Function` whose `Code` points at S3 checked the object by calling
`getObject` and throwing the result away. `S3Object.getData()` is a materialised byte array, so that
read the whole package into memory only to answer whether it was readable, and `LambdaService` then
read the same object again for real during `CreateFunction`. Every stack create or update paid for
two full copies of every Lambda package in the template.

## Root cause

The probe asked an existence question with a method that also fetches the body:

```java
try {
    s3Service.getObject(s3Bucket, s3Key);          // result discarded
    return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key), ...);
} catch (Exception e) {
    ...
}
```

and `LambdaService.extractZipCodeFromS3` then does the real read:

```java
obj = s3Service.getObject(s3Bucket, s3Key);
extractZipCodeBytes(fn, obj.getData(), region);
```

One call:

```java
- s3Service.getObject(s3Bucket, s3Key);
+ s3Service.headObject(s3Bucket, s3Key);
```

This is safe because of how `S3Service` is layered. `getObject` is `getObjectMetadata` plus
`readFile`; `headObject` is `getObjectMetadata` alone. Both resolve the object through the same
metadata lookup, so a missing key or bucket still throws at the same point and the resource still
fails exactly as #2648 established. Only the discarded body read goes away.

The issue also offered a second option, threading the fetched bytes through `LambdaCodeSpec` so the
object is read once end to end. I did not take it: it removes the second read as well, but it puts
the package payload into a structure that currently carries only a request map and an identity
string, and it couples the provisioner's data flow to `LambdaService`'s. The metadata probe removes
the duplicated read while leaving that layering alone. Happy to switch if maintainers prefer the
other shape.

## AWS Compatibility

No wire behaviour change. A template whose `Code` is unreadable still fails the resource with the
same error, and the `allow-stub-lambda-code` opt in from #2650 still works the same way.

## How it was tested

New `CloudFormationLambdaS3CodeProbeTest`, using the mocked service provisioner approach of
`CloudFormationLambdaLegacyNameModeTest`, because call counts are not observable through the HTTP
integration tests:

| Test | Covers |
|---|---|
| `theCodeProbeReadsMetadataOnlyAndNeverTheBody` | `headObject` is called and `getObject` never is |
| `anUnreadableObjectStillFailsTheResource` | the #2648 failure behaviour is unchanged |

Reverting only the probe makes both fail, so they are real regression tests:

```
anUnreadableObjectStillFailsTheResource:89  expected: <CREATE_FAILED> but was: <CREATE_COMPLETE>
theCodeProbeReadsMetadataOnlyAndNeverTheBody:73
```

## Risk / rollback

Low. One call in one method, no signature or config change, no persisted state. Rollback is
reverting the single line.

One behavioural nuance worth stating rather than leaving to be discovered: if an object's metadata
exists but its backing file cannot be read, `getObject` used to surface that here, and now it
surfaces from `LambdaService` as `Unable to fetch code from s3://...`. The stack still fails either
way, and the message from the layer that actually does the read is the more accurate of the two.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -2502,8 +2502,16 @@ public class CloudFormationResourceProvisioner {
                 // deliberately leaves its Lambda packages unbuilt and only cares about the
                 // other resources. Off by default: silently serving a placeholder is the more
                 // dangerous of the two behaviours.
+                //
+                // headObject, not getObject: this only needs to know whether the code is
+                // readable. getObject additionally reads the whole body, which is then thrown
+                // away, and LambdaService reads it again for real during CreateFunction. That
+                // is a second full copy of the package per Lambda per stack operation, for a
+                // question a metadata lookup answers (issue #2675). Both resolve the object
+                // through the same getObjectMetadata call, so a missing key or bucket still
+                // fails here exactly as before.
                 try {
-                    s3Service.getObject(s3Bucket, s3Key);
+                    s3Service.headObject(s3Bucket, s3Key);
                     return new LambdaCodeSpec(Map.of("S3Bucket", s3Bucket, "S3Key", s3Key),
                             "s3:" + s3Bucket + "\n" + s3Key);
                 } catch (Exception e) {
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationLambdaS3CodeProbeTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationLambdaS3CodeProbeTest.java
@@ -1,0 +1,125 @@
+package io.github.hectorvent.floci.services.cloudformation;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.services.cloudformation.model.StackResource;
+import io.github.hectorvent.floci.services.lambda.LambdaService;
+import io.github.hectorvent.floci.services.lambda.model.LambdaFunction;
+import io.github.hectorvent.floci.services.s3.S3Service;
+import io.github.hectorvent.floci.services.s3.model.S3Object;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+import java.util.function.Function;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyMap;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Issue #2675: resolving an AWS::Lambda::Function whose Code points at S3 checked the object by
+ * calling getObject and discarding the result, so the package was read in full twice per stack
+ * operation, once here and once by LambdaService during CreateFunction. The check only needs to
+ * know whether the object is readable, which headObject answers without reading the body.
+ *
+ * <p>Uses the mocked-service provisioner approach of CloudFormationLambdaLegacyNameModeTest,
+ * because how many times a service is called is not observable through the HTTP integration tests.
+ */
+class CloudFormationLambdaS3CodeProbeTest {
+
+    private static final String ACCOUNT_ID = "000000000000";
+    private static final String REGION = "us-east-1";
+    private static final String STACK_NAME = "test-stack";
+    private static final String BUCKET = "pkg";
+    private static final String KEY = "fn.zip";
+
+    private final ObjectMapper mapper = new ObjectMapper();
+    private S3Service s3Service;
+    private LambdaService lambdaService;
+    private CloudFormationResourceProvisioner provisioner;
+
+    @BeforeEach
+    void setUp() {
+        s3Service = mock(S3Service.class);
+        lambdaService = mock(LambdaService.class);
+        when(lambdaService.createFunction(anyString(), anyMap()))
+                .thenAnswer(inv -> lambdaFunction((String) ((Map<?, ?>) inv.getArgument(1)).get("FunctionName")));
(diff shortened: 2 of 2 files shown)


avison9:
The compilation fix referred to above is now open as #2729. It is a single line bringing `AslExecutorTaskHistoryEventsTest` into line with its siblings. Once it merges I will rebase this branch and CI here will be meaningful; until then the red shards on this PR are inherited from main and say nothing about the change.

avison9:
Rebased onto main now that #2729 has merged. The red shards on the previous push were inherited from
main not compiling; that is fixed, so CI here is meaningful again and the note at the top of the
description no longer applies.

One further change while rebasing, which is worth explaining rather than leaving as a silent diff.

## The test no longer builds the provisioner positionally

#2744 removed the `SqsService` constructor argument. My test passed 35 positional nulls, so
everything after position two shifted and it stopped compiling:

```
incompatible types: LambdaService cannot be converted to IamService
```

That is the same failure mode as the one that had main red: a test constructing something
positionally, and a constructor changing in a PR that never touched the test.

The same refactor added `CfnProvisionerFixture`, so the test now names only what it uses:

```java
provisioner = CfnProvisionerFixture.builder()
        .s3(s3Service)
        .lambda(lambdaService)
        .objectMapper(mapper)
        .build();
```

Three named services rather than 35 positional arguments, and unaffected by the next constructor
change.

I also checked the trap the fixture's javadoc warns about: with an empty registry, a type that has
already migrated to a per service provisioner reaches the default arm and is stubbed with a fake ARN
and `CREATE_COMPLETE`, so a test can pass while provisioning nothing. `AWS::Lambda::Function` is
still served by the provisioner's own switch, so an empty registry is correct here and these tests
are not passing vacuously.

## Verification

Full suite against current main, with the rebase and the fixture change in place. Figures in the
checks below, and the two probe tests still fail if the `headObject` call is reverted to
`getObject`, so they remain real regression tests rather than passing assertions.

hectorvent:
Thank you, and thanks for rebasing. The six red checks were never yours: they were the `AslExecutorTaskHistoryEventsTest` compile break main was carrying, and moving onto current main cleared them.

I checked the claim in your comment rather than taking it on trust, and it holds exactly. `getObject` is `getObjectMetadata` followed by a body read, `headObject` is that same `getObjectMetadata` alone, so a missing bucket or key still throws from the identical place and only the discarded read goes away.

The test is the part I liked most. Verifying `getObject` is never called on either overload pins the saving itself, and keeping the `NoSuchKey` case asserting CREATE_FAILED pins the behaviour from #2648 that this must not disturb.

No blockers from my side, once the build shards report green.

hectorvent (approved):

PR #2729, opened 2026-08-29
test(stepfunctions): restore AslExecutorTaskHistoryEventsTest compilation

avison9:
## Summary

`main` does not compile its test sources, and has not since `08e532ab` landed. This adds the one
missing constructor argument.

```
AslExecutorTaskHistoryEventsTest.java:[64,20]
  constructor AslExecutor in class AslExecutor cannot be applied to given types;
  required: ..., io.vertx.mutiny.core.Vertx, io.github.hectorvent.floci.core.common.CustomResourceLiveness
  found:    ..., io.vertx.mutiny.core.Vertx
```

## Root cause

A semantic merge conflict between two changes that were each green on their own branch.

`08e532ab` (#2688) added a `CustomResourceLiveness` parameter to the `AslExecutor` constructor and
updated the call sites that existed at the time. `5600d21f` (#2703) added
`AslExecutorTaskHistoryEventsTest`, which constructs `AslExecutor` directly. Neither branch touched
the other's file, so neither PR could see the conflict, and merging them in sequence broke the
build.

## What changed

The test now passes the same mock its siblings already do. `AslExecutorErrorTerminationTest` and
`AslExecutorCustomResourceLivenessTest` construct `AslExecutor` this way already, so this brings the
new test in line rather than inventing anything:

```java
mock(Instance.class), mock(EmulatorConfig.class), vertx,
mock(io.github.hectorvent.floci.core.common.CustomResourceLiveness.class));
```

## Why the tracker looks green

Worth flagging, because the breakage is easy to miss from the PR list. Open PRs whose checks ran
before `08e532ab` merged still display green: GitHub does not re-run a PR's checks when the base
branch advances. For example #2723's checks completed at 16:52 UTC, while the breaking commit merged
at 19:01 UTC. Anything opened or pushed after that point fails, and `main`'s own CI has been failing
for three commits.

## How it was tested

`./mvnw clean test-compile` fails on `main` at `8a705d0a` and succeeds with this change.
`AslExecutorTaskHistoryEventsTest` then runs and passes.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## Checklist

No new test: this restores compilation of an existing one. The suite could not run at all before it.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutorTaskHistoryEventsTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutorTaskHistoryEventsTest.java
@@ -77,7 +77,8 @@ class AslExecutorTaskHistoryEventsTest {
                 mock(io.github.hectorvent.floci.services.scheduler.SchedulerController.class),
                 objectMapper,
                 new JsonataEvaluator(objectMapper),
-                mock(Instance.class), mock(EmulatorConfig.class), vertx);
+                mock(Instance.class), mock(EmulatorConfig.class), vertx,
+                mock(io.github.hectorvent.floci.core.common.CustomResourceLiveness.class));
     }
 
     @Test


pgermosen (approved):
Approving. Checked the actual constructor and this is exactly the missing argument, in the right order. Worth merging quickly since main's test-compile has been broken since the #2688/#2703 conflict - the sdk-test-java failure on this PR's own CI run is the unrelated RedshiftTest connection-refused flake, not this change.

PR #2731, opened 2026-08-29
fix(ecs): scope service tasks by owning service, not by caller-supplied group

### LINKED ISSUE #2716: ecs: service reconciliation matches tasks by spoofable group and cluster-name suffix
`EcsService` scopes tasks to a service using a weak, spoofable match in both the REPLICA and DAEMON reconciliation paths:

```java
.filter(t -> t.getClusterArn().endsWith(":cluster/" + clusterName))
.filter(t -> svc.getServiceArn().equals(t.getGroup())
        || svc.getServiceName().equals(t.getGroup()))
```

Two related gaps, both flagged by Greptile on #2482 and never addressed there (pre-existing pattern, not introduced by that PR):

1. **Caller-controlled group claims ownership** — `RunTask`/`StartTask` lets a caller set an arbitrary `group`. Any task whose `group` happens to equal a service's ARN or name is treated as service-owned by the reconciler, so an unrelated task can be swept up, stopped as a "duplicate", or counted toward `runningCount`.
2. **Cluster suffix breaks isolation** — `endsWith(":cluster/" + clusterName)` matches on cluster name suffix only. If emulator state ever holds same-named clusters in different regions/accounts, reconciliation for one can count or stop tasks belonging to the other.

Occurrences (as of `main`):
- `EcsService.java` REPLICA path (`reconcileService`)
- `EcsService.java` DAEMON path (`reconcileDaemonService`, added in #2482)

Fix direction: scope by an unambiguous, non-caller-controlled identifier (e.g. a service-owned task tag/attribute set at launch time instead of reusing the free-form `group` field) and compare full cluster ARNs instead of a name suffix.

Origin: https://github.com/floci-io/floci/pull/2482#discussion_r (Greptile review comments dated 2026-08-22T23:40:48Z, lines 1583/1586 of that diff)

omatheusmesmo:
Fixes #2716.

`RunTask`/`StartTask` let the caller set an arbitrary `group`, and the service reconciler treated that string as proof of ownership:

```java
.filter(t -> t.getClusterArn().endsWith(":cluster/" + clusterName))
.filter(t -> svc.getServiceArn().equals(t.getGroup())
        || svc.getServiceName().equals(t.getGroup()))
```

A field that crosses the trust boundary from the caller cannot double as an authorization token, so any task naming a service as its group was adopted by it.

## Scope

The issue names two occurrences; the same match turned up at **five** sites, three of which have side effects beyond counting:

| Site | Effect of a spoofed `group` |
|---|---|
| `reconcileService` (REPLICA) | counted toward `runningCount`, or **stopped** as a duplicate / scale-in |
| `reconcileDaemonService` | counted, or **stopped** as a duplicate daemon task |
| `deleteService` | **stopped** on service teardown |
| LB register/deregister | **registered into the service’s ELBv2 target groups** |
| `listTasks` (`serviceName` filter) | listed as service-owned |

## Fix

Tasks carry an `owningServiceArn`, stamped only by the reconciler’s own launch path. It is never read from a request and never serialized, so it cannot be forged. Cluster scoping moved from an `endsWith(":cluster/" + name)` suffix test to full-ARN equality, which also stops same-named clusters in different regions or accounts from reconciling into each other.

The REPLICA path previously launched through the public `runTask(...)` — the very entry point callers use — so ownership could not be inferred from the call path; it now goes through a private `launchServiceTask(...)`.

## Checked against the AWS docs

- [`Task.group`](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Task.html) — "The name of the task group that is associated with the task." A label; nothing about ownership.
- [`ListTasks.serviceName`](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListTasks.html) — "limits the results to tasks that **belong to** that service", i.e. real ownership, not a string match on a caller field.
- [`Task.startedBy`](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Task.html) — "If an Amazon ECS service started the task, the `startedBy` parameter contains the **deployment ID** of that service."

That last point is the design confirmation: real AWS also records service ownership in a server-assigned field rather than in `group`. `startedBy` was deliberately *not* reused for it here, because in the emulator it is caller-settable through `RunTask` too and would reproduce the identical hole.

## Tests

`EcsServiceTaskOwnershipTest` covers all five paths plus cross-region cluster isolation. Every test was first run against the unfixed code: **5 of the 6 failed**. Two early drafts passed either way — they confirmed the bug rather than the fix — and were tightened until they genuinely pinned the behaviour.

Full ECS suite: **152 tests, 0 failures**, plus the StepFunctions, Application Auto Scaling and ELBv2 suites that exercise ECS tasks.

> Note: CI on this branch will fail at test-compile until #2730 lands — `main` is currently red for an unrelated reason (#2688/#2703 semantic conflict). Rebasing afterwards will make it green.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java
@@ -374,7 +374,7 @@ public class EcsService implements ContainerTeardown, ResourceProvider {
         if (cpu == 4096) return memory >= 8192 && memory <= 30720 && memory % 1024 == 0;
         if (cpu == 8192) return memory >= 16384 && memory <= 61440 && memory % 4096 == 0;
         if (cpu == 16384) return memory >= 32768 && memory <= 122880 && memory % 8192 == 0;
-        
+
         return false;
     }
 
@@ -477,7 +477,7 @@ public class EcsService implements ContainerTeardown, ResourceProvider {
         EcsCluster cluster = resolveClusterOrDefault(clusterRef, region);
         TaskDefinition taskDef = resolveTaskDefinitionOrThrow(taskDefinitionRef, region);
         return launchTasks(cluster, taskDef, count, launchType, group, startedBy, null,
-                containerOverrides, networkConfiguration, region);
+                containerOverrides, networkConfiguration, null, region);
     }
 
     public List<EcsTask> startTask(String clusterRef, List<String> containerInstanceRefs,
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Approving. Real vulnerability, well fixed — owningServiceArn is stamped only by the reconciler's own launch path and used consistently at all five spots that used to trust the caller-supplied group. Greptile's ListTasks finding already checks out as fixed: the no-cluster/serviceName path uses a plain map lookup, not the cluster-creating resolver. Spot-checked the Task.group/startedBy doc quotes against the real API reference and they're accurate.

PR #2732, opened 2026-08-29
refactor(cloudformation): add the provisioner-migration enablers

hectorvent:
## Summary

Groundwork for moving the remaining CloudFormation resource types out of
`CloudFormationResourceProvisioner` (7,648 lines, 79 types in one switch, 35-arg constructor) into
per-service provisioners. **No behaviour change** — every type is provisioned and deleted exactly as
before. This is the tooling and the guardrails that make each later per-service slice mechanical
instead of inventive.

Four enablers:

**1. A dropped resource type fails loudly.** The provision switch's `default` arm stubs an unknown
type with `arn:aws:stub:::<logicalId>` and reports `CREATE_COMPLETE`, so a type that fell out of both
the switch and the registry would pass every test while provisioning nothing. Adds
`CloudFormationResourceRegistry#registeredTypes()`, a `LEGACY_SWITCH_TYPES` constant, a throwing
`default` arm for declared-but-unhandled types, and `supported-resource-types.tsv` — a checked-in
inventory of all 112 types and who serves each. `CfnResourceInventoryTest` compares it against the
**CDI-resolved** registry, which is the only way to catch a provisioner written without
`@ApplicationScoped` (it compiles, unit-tests green, and is never wired).

**2. Delete precedence.** The registry lookup in `delete(StackResource, region)` sat *after* a
`Custom::` prefix branch and nine attribute-aware special cases, so migrating one of those types
would leave its new provisioner's `delete` silently unreachable. The lookup now runs first (exact
match beats prefix match, which is what will let `Custom::DynamoDBReplica` migrate at all), the nine
branches moved behind a `DELETE_NEEDS_STACK_RESOURCE` set that gates them, and a duplicated
`AWS::Events::Rule` block was removed. `CfnDeletePrecedenceTest` fails if a type is ever claimed by
both a provisioner and a special case.

**3. One test fixture instead of 13 positional constructions.** Thirteen tests built the provisioner
with 30-plus positional `null`s, so every constructor change edited all of them — the main reason the
migration stalled. `CfnProvisionerFixture` gives one named setter per argument; call sites now read
as intent (`.dynamoDb(x).lambda(y)`). Verified by removing the already-dead `sqsService` argument:
constructor arity changed and **only the fixture** needed editing.

**4. Shared helpers the slices will use.** `ProvisionContext.isUpdate()` (captured at construction,
because `provision` assigns the physical id mid-method and a resource-derived check would silently
flip meaning), `ProvisionContext.resolveTags`, `CfnDeletes.safeDelete` with explicit tolerated error
codes, and `CfnDynamicReferences` — 189 lines of `{{resolve:ssm:…}}` / `{{resolve:secretsmanager:…}}`
handling extracted so the future RDS provisioner injects one collaborator rather than two unrelated
services. **`resolveTags` deliberately migrates zero call sites**; the seven existing copies disagree
on real behaviour (blank keys, null-vs-empty, ordering, validation) and each adopts it in its own
slice, where its tests are.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

`refactor(cloudformation):` — no user-visible behaviour change, so it is release-neutral.

## AWS Compatibility

N/A for the wire protocol: no request parsing, response shape, error code, or ARN generation changes.
Every provisioned type keeps its exact physical id and `Fn::GetAtt` attribute keys; the delete change
is a dispatch reordering whose two paths were verified non-overlapping before the change (the nine
attribute-aware types share no type with the 33 registry types, so nothing actually reroutes today).

The evidence is the existing suite passing unmodified: a migrated or misrouted type would fail the
CloudFormation integration tests, which were not touched.

## Checklist

New tests: `CfnResourceInventoryTest` (4), `CfnDeletePrecedenceTest` (2), `ProvisionContextTest` (5),
`CfnDeletesTest` (5).

**Each guard was mutation-tested rather than trusted for passing:**

| Mutation | Result |
|---|---|
| Bogus type added to the inventory | 2 of 4 inventory tests fail |
| `AWS::SSM::Parameter` switch arm deleted | silent `arn:aws:stub:::Parameter` becomes a named `CREATE_FAILED` |
| A type registered while its delete special case remains | precedence test fails with the fix to apply |
| A listed attribute-aware type's branch removed | silent no-op becomes a named failure |

The inventory guard also proved itself on rebase: #2701 landed `WafV2CfnProvisioner` and
`ConfigCfnProvisioner` mid-review and the test caught the two new types immediately (110 → 112 rows).


pgermosen (approved):
Approving. Traced the two behavior-sensitive claims directly rather than taking the "no behaviour change" line at face value: the delete-precedence reordering really is a no-op today (none of the 20 registered provisioners overlap the DELETE_NEEDS_STACK_RESOURCE set), and the duplicated Events::Rule block removal left the real one intact. isUpdate() capturing the physical id from an immutable parameter before provision() mutates the resource is the right call. Good guardrails for the migration ahead.

PR #2739, opened 2026-08-29
docs(cloudformation): generate the supported resource-type table

hectorvent:
## Summary

The CloudFormation **Supported Resource Types** table was hand-maintained, and had drifted. Ten
provisioned types were undocumented:

| Service | Undocumented |
|---|---|
| CloudFront | `Distribution` (the entire row was missing) |
| Lambda | `Permission`, `MicrovmImage`, `NetworkConnector` |
| RDS | `DBProxy`, `DBProxyTargetGroup` |
| ECS | `CapacityProvider`, `ClusterCapacityProviderAssociations` |
| EC2 | `VPCEndpoint` |
| API Gateway v2 | `Authorizer` |

This generates the table from the provisioner inventory TSV that `CfnResourceInventoryTest` already
pins to the CDI-resolved registry (added in #2732), so it can no longer drift from what Floci
actually provisions. `make docs-sync` writes it, `make docs-check` gates it.

**Stacked on #2732** — review that first; this branch contains its commit.

### Why the TSV and not the Java

Several provisioners return `Set.of(CONSTANT, CONSTANT)` from `resourceTypes()`, so a source regex
sees only a subset (22 of 33 today). The TSV is the one representation that is both machine-checked
against the running registry and readable without a JVM.

### The table is prose as much as data

A naive generator would have destroyed real information, so presentation stays configurable in
`cfn_resource_types.yaml`:

- **`type_order`** keeps the curated reading order (`VPC`, `Subnet`, `SecurityGroup`… not
  alphabetical). Types not listed are appended alphabetically, so a newly provisioned type still
  appears without touching the config.
- **`notes` / `row_notes`** preserve the existing annotations, including the Lambda row's
  explanation of `cfn-response` injection.
- **`extra_types`** covers capabilities that are real but are not provisioner types, each requiring
  a stated reason. Nested `Stack` and the `Custom::*` wildcard are the two: the wildcard *cannot*
  be an inventory type, because the registry is keyed by exact type.

### Also gated, cheaply

Two checks that otherwise need a Quarkus boot now run in `docs-check`: a provisioner missing
`@ApplicationScoped` (CDI silently never registers it), and a type claimed by both the legacy switch
and a provisioner (the registry wins, leaving a switch arm that looks live but is dead).

The workflow's path filters gained `src/test/resources/cloudformation/**`. The inventory lives under
`src/test`, which `src/main/**` did not cover, so a migration slice that only moved a type between
owners would have skipped this gate entirely.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A. Documentation and build tooling only; no Java changed, so no emulated surface is affected.

## Checklist

`make docs-test` 50 passed (16 new). `make docs-check` clean. No Java touched, so the Java suite is
unaffected and was not re-run beyond #2732's full green run.

**Gate mutation-tested rather than trusted for passing:**

| Mutation | Result |
|---|---|
| Added an undocumented provisioner type to the inventory | `--check` fails: table stale, plus a warning that the namespace has no label |
| Removed `@ApplicationScoped` from `SqsCfnProvisioner` | `--strict` fails naming the file and the consequence |
| Ran the generator twice | Idempotent, no diff |

### DIFF
--- a/.github/workflows/docs-actions.yml
+++ b/.github/workflows/docs-actions.yml
@@ -1,13 +1,16 @@
 name: Docs Action Tables
 
-# Keeps the generated "Supported Actions" tables in docs/services/*.md in sync with
-# handler source. Runs on handler changes (which can add/remove actions) and on doc or
-# tooling changes. Python-only, so it is fast and independent of the Java build job.
+# Keeps the generated tables in docs/services/*.md in sync with their sources: the
+# "Supported Actions" tables with handler source, and the CloudFormation "Supported Resource
+# Types" table with the provisioner inventory TSV. Runs on handler changes (which can add or
+# remove actions), inventory changes (which add or move resource types), and on doc or tooling
+# changes. Python-only, so it is fast and independent of the Java build job.
 
 on:
   pull_request:
     paths:
       - 'src/main/**'
+      - 'src/test/resources/cloudformation/**'
       - 'docs/services/**'
       - 'tools/docs/**'
       - 'Makefile'
@@ -17,6 +20,7 @@ on:
       - main
     paths:
       - 'src/main/**'
+      - 'src/test/resources/cloudformation/**'
       - 'docs/services/**'
       - 'tools/docs/**'
       - 'Makefile'
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,18 @@ PYTHON ?= python3
 
 docs-sync: ## Regenerate the action tables in docs/services from handler source (in place)
 	$(PYTHON) tools/docs/regen_action_docs.py
+	$(PYTHON) tools/docs/regen_cfn_resource_types.py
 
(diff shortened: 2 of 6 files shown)


pgermosen (approved):
Approving the content — read the generator script and its full test suite, the doc drift it's closing is real, and the double-ownership/@ApplicationScoped checks are a nice complement to #2732's guardrails.

One thing before this lands, though: the base branch here is refactor/cfn-wave0-enablers, and #2732 was squash-merged into main rather than fast-forwarded — ac71876c (this branch's parent) isn't actually an ancestor of main. So merging as-is would land this into the now-orphaned feature branch, not main. Could you retarget the base to main (and rebase — the diff right now still carries all of #2732's already-merged changes)?

PR #2740, opened 2026-08-29
feat(connect): instance management (CreateInstance family, attributes, storage configs, tags)

lex00:
Adds an Amazon Connect control plane covering instance CRUD and instance
attributes, plus instance storage configs and the shared tag operations.
Unblocks the Terraform provider's aws_connect_instance create path, which
currently 404s against floci.

REST JSON on the real AWS paths (/instance, .../attribute, .../storage-config)
under the connect signing name; attribute, directory and storage resource
types validated against the AWS service model enums. Instances report ACTIVE
from the first DescribeInstance so the provider's waiter completes.

17 tests pass in the new ConnectIntegrationTest, covering the lifecycle,
attribute validation, storage-config association and tag round-trips. Default
behavior unchanged; disable via floci.services.connect.enabled. Happy to
adjust to conventions.


pgermosen (approved):
Approving. The attribute-validation fix closes the state-inconsistency bug cleanly - rejects before any mutation happens, so the attributes map and the boolean flags can't diverge. Checked the error codes against the real UpdateInstanceAttribute reference and both are documented valid errors for this operation.

PR #2742, opened 2026-08-29
feat(cognito): user pool domain operations (Create/Describe/DeleteUserPoolDomain)

### LINKED ISSUE #2590: [FEAT] Cognito: user pool domain operations (CreateUserPoolDomain et al.)
## Service

Cognito (cognito-idp)

## API Action / Feature

Cognito user pool domains — `CreateUserPoolDomain`, `DescribeUserPoolDomain`,
`DeleteUserPoolDomain`, `UpdateUserPoolDomain`, and the `Domain` field on
`DescribeUserPool`.

All currently return `UnsupportedOperation`.

## AWS Documentation

- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPoolDomain.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DescribeUserPoolDomain.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DeleteUserPoolDomain.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UpdateUserPoolDomain.html

## Why is this needed?

`aws_cognito_user_pool_domain` is a standard resource, and it sits near the root of the dependency graph. Everything hanging off the pool's hosted UI waits on it. When it fails, the rest of the stack cannot proceed.

### Reproduction:

```sh

export AWS_DEFAULT_REGION=us-east-1 AWS_ENDPOINT_URL=http://localhost:4566

POOL=$(aws cognito-idp create-user-pool --pool-name repro \
         --query 'UserPool.Id' --output text)          # works

aws cognito-idp create-user-pool-domain --user-pool-id "$POOL" --domain repro-domain
aws cognito-idp describe-user-pool-domain --domain repro-domain
```

### Actual:

```
An error occurred (UnsupportedOperation) when calling the CreateUserPoolDomain
operation: Operation CreateUserPoolDomain is not supported.

An error occurred (UnsupportedOperation) when calling the DescribeUserPoolDomain
operation: Operation DescribeUserPoolDomain is not supported.
```

`describe-user-pool` on the same pool returns normally, so the pool itself is fine.

### Expected:

The prefix is associated with the pool, then visible via
`DescribeUserPoolDomain` and as `Domain` on `DescribeUserPool`.

Prefix domains alone would unblock this. Custom domains need an ACM certificate
and a Cloudfront FQDN, neither of which is meaningful locally. For a prefix
domain AWS returns `CloudFrontDomain: null`, so the response can stay minimal.

### Environment:

- Floci 1.7.0 (`floci/floci@sha256:bec9b9f749322444a57fb0f00d3e007ffd49015162361aab84ccecd9b4f5f8ed`),
- macOS arm64
- Docker 29.5.3
- aws-cli 2.34.10
- Also reproduced through the Pulumi AWS provider 7.40.0

- [ ] Yes
- [x] No

lex00:
Fixes #2590

Adds CreateUserPoolDomain, DescribeUserPoolDomain and DeleteUserPoolDomain,
unblocking aws_cognito_user_pool_domain (e.g. ALB authenticate-cognito
hosted-UI flows). It supports both Cognito prefix domains and custom domains
via CustomDomainConfig.CertificateArn, where the response carries the
CloudFrontDistribution target Terraform reads. Domain names are globally
unique, a custom domain without CertificateArn is rejected with
InvalidParameterException, and a missing domain reports
ResourceNotFoundException. The three actions are pure additions; no existing
behavior changes.

The new CognitoUserPoolDomainIntegrationTest adds 9 tests, and the full
Cognito suite passes locally with 355 tests and 0 failures. Happy to adjust
to conventions.

### DIFF
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -55,6 +55,14 @@ Standalone `TagResource` rejects reserved `floci:*` keys. `ListTagsForResource` 
 | ListResourceServers | Lists resource servers for a user pool. |
 | DeleteResourceServer | Deletes a resource server from a user pool. |
 
+### User Pool Domains
+
+| Action | Description |
+|--------|-------------|
+| CreateUserPoolDomain | Creates a Cognito prefix domain, or a custom domain when `CustomDomainConfig.CertificateArn` is given. |
+| DescribeUserPoolDomain | Returns a domain's description, including `CloudFrontDistribution` for custom domains. |
+| DeleteUserPoolDomain | Deletes a domain from its user pool. |
+
 ### Admin User Management
 
 | Action | Description |
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
@@ -14,6 +14,7 @@ import io.github.hectorvent.floci.services.cognito.model.ResourceServerScope;
 import io.github.hectorvent.floci.services.cognito.model.UserPool;
 import io.github.hectorvent.floci.services.cognito.model.UserPoolClient;
 import io.github.hectorvent.floci.services.cognito.model.UserPoolClientSecret;
+import io.github.hectorvent.floci.services.cognito.model.UserPoolDomain;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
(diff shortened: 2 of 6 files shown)


pgermosen (approved):
Approving. Checked the DeleteUserPool fix against the real API reference rather than the code comment's word for it, and AWS's own documented example response for this exact case is byte-for-byte identical to what's thrown here - error code and message both. Both directions are tested, and the rest of the domain lifecycle (uniqueness, CertificateArn requirement, not-found) is well covered too.

PR #2744, opened 2026-08-29
refactor(cloudformation): move the single-service types to provisioners

hectorvent:
## Summary

First slice that actually moves types. Migrates the seven whose provisioning is a single service
call, out of `CloudFormationResourceProvisioner`'s switch and into per-service provisioners:

| Provisioner | Types |
|---|---|
| `KmsCfnProvisioner` | `AWS::KMS::Key`, `AWS::KMS::Alias` |
| `SsmCfnProvisioner` | `AWS::SSM::Parameter` |
| `EcrCfnProvisioner` | `AWS::ECR::Repository` |
| `PipesCfnProvisioner` | `AWS::Pipes::Pipe` |
| `FirehoseCfnProvisioner` | `AWS::KinesisFirehose::DeliveryStream` |
| `CdkMetadataCfnProvisioner` | `AWS::CDK::Metadata` |

**Behaviour-preserving.** Bodies are ported verbatim apart from the mechanical rewrites onto
`ProvisionContext` (`engine` → `ctx.engine()`, region/account/stack → `ctx.*`, `resolveOptional`,
`generatePhysicalName`), so physical ids, `Fn::GetAtt` keys and delete calls are unchanged. The
generated docs table is byte-identical after this change: only the *owner* moved, not the types.

Monolith: **7,603 → 7,396 lines**; legacy switch 79 → 72 types.

**Stacked on #2739 → #2732.** Review those first.

### Two judgement calls worth flagging

**`ProvisionContext.resolveTags` was deliberately not substituted for `parseCfnTags`.** They are
not equivalent: `resolveTags` skips a blank key that `parseCfnTags` keeps, orders entries by
insertion rather than hash, and resolves the whole property so an `Fn::If` around the list works.
Swapping would be a silent behaviour change in a move PR, so `parseCfnTags` is copied into the
three provisioners that need it (as are `blankToNull` and `parseIntProp`), each still having
callers left in the monolith. Adoption is filed as a follow-up.

**Five service fields left the monolith, but their constructor parameters stayed.** Unused
parameters are legal, and shrinking the signature now would churn `CfnProvisionerFixture` for no
gain; the final cleanup drops all of them at once.

**`AWS::CDK::Metadata` gains a real provisioner** rather than being left to the stub arm. The stub
reports `CREATE_COMPLETE` with a synthetic id *and a fake `Arn` attribute*; the real type has no
attributes at all, so the stub was quietly wrong.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

`refactor(cloudformation):` — no user-visible behaviour change, release-neutral.

## AWS Compatibility

N/A for the wire protocol. Every migrated type keeps its exact physical id and `Fn::GetAtt`
attribute keys, verified per type against its CloudFormation registry schema:

| Type | schema `primaryIdentifier` / `readOnlyProperties` | asserted |
|---|---|---|
| `KMS::Key` | `KeyId` / `Arn`, `KeyId` | both keys, exact map |
| `KMS::Alias` | `AliasName` / none | physical id |
| `SSM::Parameter` | `Name` / none | physical id + `Name`/`Type`/`Value` |
| `ECR::Repository` | `RepositoryName` / `Arn`, `RepositoryUri` | both keys, exact map |
| `Pipes::Pipe` | `Name` / `Arn`, +4 | `Arn` (see follow-up) |
| `KinesisFirehose::DeliveryStream` | `DeliveryStreamName` / `Arn` | exact map |

## Checklist

CloudFormation + Cloud Control: **835 tests, 0 failures, 0 errors**. `LeafCfnProvisionerTest` adds
20 covering every migrated type: exact physical id, exact `Fn::GetAtt` keys, delegated arguments
via captors, and delete delegation. The pre-existing CloudFormation integration tests were not
touched, which is the real evidence nothing regressed.

The inventory guard from #2732 is what proves the migration actually took effect: it compares the
checked-in inventory against the **CDI-resolved** registry, so a provisioner missing
`@ApplicationScoped` (silently never registered, its types quietly stubbed) fails the build.

## Follow-ups, deliberately not fixed here

1. **`AWS::Pipes::Pipe` sets only `Arn`.** The registry schema also lists `CreationTime`,
   `CurrentState`, `LastModifiedTime` and `StateReason` as read-only, so `Fn::GetAtt` on any of
   those resolves to the literal `"LogicalId.Attr"`. Pre-existing, carried over unchanged.
2. **Adopt `ProvisionContext.resolveTags`** across the `parseCfnTags` call sites once the
   behaviour difference is settled, and delete the copies.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -224,7 +224,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::Batch::ComputeEnvironment",
             "AWS::Batch::JobDefinition",
             "AWS::Batch::JobQueue",
-            "AWS::CDK::Metadata",
             "AWS::CloudFormation::CustomResource",
             "AWS::CloudFront::Distribution",
             "AWS::CloudWatch::Alarm",
@@ -241,7 +240,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::EC2::SecurityGroup",
             "AWS::EC2::Subnet",
             "AWS::EC2::SubnetRouteTableAssociation",
-            "AWS::ECR::Repository",
             "AWS::ECS::Cluster",
             "AWS::ECS::Service",
             "AWS::ECS::TaskDefinition",
@@ -259,15 +257,11 @@ public class CloudFormationResourceProvisioner {
             "AWS::IAM::ManagedPolicy",
             "AWS::IAM::Policy",
             "AWS::IAM::User",
-            "AWS::KMS::Alias",
-            "AWS::KMS::Key",
             "AWS::Kinesis::Stream",
-            "AWS::KinesisFirehose::DeliveryStream",
             "AWS::Lambda::EventSourceMapping",
             "AWS::Lambda::Function",
             "AWS::Lambda::LayerVersion",
             "AWS::Logs::LogGroup",
-            "AWS::Pipes::Pipe",
             "AWS::RDS::DBCluster",
             "AWS::RDS::DBClusterParameterGroup",
             "AWS::RDS::DBInstance",
@@ -281,7 +275,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::S3::BucketPolicy",
             "AWS::SNS::Subscription",
             "AWS::SNS::Topic",
-            "AWS::SSM::Parameter",
             "AWS::SecretsManager::Secret",
             "AWS::SecretsManager::SecretTargetAttachment",
             "AWS::StepFunctions::StateMachine",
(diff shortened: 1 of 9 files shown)


pgermosen (approved):
Approving. On greptile's finding about the update path recreating resources / CDK metadata getting a new identity each time: I pulled the exact pre-migration provisionEcrRepository and provisionCdkMetadata methods from the monolith's history and diffed them against these new provisioners. Both are byte-identical in behavior — no create-vs-update branching existed before this move either. So this is a real, pre-existing gap across the CFN provisioner set, not something this migration introduced. The "ported verbatim" claim holds up.

One small correction for the description: CDK::Metadata actually had its own dedicated provisionCdkMetadata case before, not the generic stub arm — doesn't change anything functionally, both generated a random UUID every call.

PR #2746, opened 2026-08-29
fix(opensearch): address spawned domain containers by IP, not name

lex00:
Real-mode OpenSearch domains are addressed by Docker container name, but
container-name DNS only resolves on a user-defined network. Nothing
guarantees the spawned container and floci land on one together: a plain
`docker run` leaves both on the default bridge, where the stored endpoint
and the readiness probe's URL never resolve, so DescribeDomain hands out a
dead endpoint and the domain never reports ready.

The manager now reads the container's resolved IP from the lifecycle
manager's EndpointInfo, the same pattern Neptune and the other
container-backed services already use, and probes readiness through the
stored endpoint. A host port is allocated and bound only when floci runs
outside a container, and released again if the start fails.

All 64 OpenSearch tests pass, including the updated
OpenSearchDomainManagerTest stub, which now carries a resolved endpoint.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
@@ -6,6 +6,7 @@ import io.github.hectorvent.floci.core.common.docker.ContainerBuilder;
 import io.github.hectorvent.floci.core.common.docker.ContainerDetector;
 import io.github.hectorvent.floci.core.common.docker.ContainerLifecycleManager;
 import io.github.hectorvent.floci.core.common.docker.ContainerLifecycleManager.ContainerInfo;
+import io.github.hectorvent.floci.core.common.docker.ContainerLifecycleManager.EndpointInfo;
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Approving. Checked resolveEndpoint directly rather than trusting the description — it really does branch the same way Neptune/MemoryDB/RDS already do, and it reads the actual bound host port from Docker's inspect response rather than trusting the originally-requested one, which is more robust. Port release on a failed start is correctly guarded by the sentinel, and the readiness probe reading the stored endpoint instead of reconstructing a container-name URL is exactly the right fix for the root cause.

cfranzen:
@pgermosen or @lex00: I think this PR introduced a regression. The OpenSearch ports are no longer exposed to the Docker host, but only reachable from inside the Docker network. Has that been an intentional change?

lex00:
Partly intentional, partly not. The container-IP endpoint was the point of the PR, but dropping the host-port binding alongside it was a mistake: it copied the Neptune/MemoryDB pattern, and those services keep host clients connected through floci-internal proxies OpenSearch doesn't have. Fix is up in #2798: the host port is published in both topologies again, with a regression test.

PR #2747, opened 2026-08-29
fix(ec2): honour documented Describe* filters and stop ENIs inheriting instance tags

lex00:
Several EC2 Describe* calls accepted a documented filter name and returned
the unfiltered list anyway, because the unhandled name fell through
matchesFilter's default arm (DescribeAddresses never applied its filters
parameter at all). A caller reading Volumes[0] or Vpcs[0] then got whichever
resource iteration order produced, which looks nondeterministic but is a
plain missing-filter bug.

The fix covers six calls. DescribeSecurityGroups gains its description
filter; DescribeVpcs accepts cidr-block (verified against live AWS
2026-08-25 as an accepted alias for cidr, primary CIDR only);
DescribeVpcEndpoints filters by the documented vpc-endpoint-state name
instead of the undocumented "state"; DescribeVolumes gains the attachment.*
family; DescribeInstances gains image-id; and DescribeAddresses now applies
every documented filter its store can back plus the generic tag: family.

Also, DescribeNetworkInterfaces no longer copies the attached instance's
tags into an ENI's tagSet: an interface's tags are its own, and
RunInstances now honours a TagSpecification with
ResourceType=network-interface, the way AWS actually tags launch-created
interfaces.

Each fix carries a regression test asserting the negative case an
unimplemented filter gets wrong, where a value matching nothing must return
zero resources rather than everything. The full local EC2 suite passes with
649 tests and 0 failures.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -4417,6 +4417,9 @@ public class Ec2Service implements ContainerTeardown, ResourceProvider {
         if (resourceId.startsWith("sgr-")) {
             return "security-group-rule";
         }
+        if (resourceId.startsWith("eni-")) {
+            return "network-interface";
+        }
         if (resourceId.startsWith("sg-")) {
             return "security-group";
         }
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
```diff
@@ -4987,7 +4992,12 @@ public List<Address> describeAddresses(String region, List<String> allocationIds
                 }
             }
         }
-        return matched;
+        // The filters parameter was previously accepted and never applied: every
```
pgermosen, line 4995: This is correct - filtering the same Address objects the refresh loop just mutated - but there's no test pinning it. Something like: allocate + associate an address, change the instance's Docker IP out from under it (or otherwise force pointAddressAtInstance to update it), then filter DescribeAddresses by the new public-ip and assert it matches, would catch a regression back to filter-before-refresh.

src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
```diff
@@ -4417,6 +4417,9 @@ private String inferResourceType(String resourceId) {
         if (resourceId.startsWith("sgr-")) {
             return "security-group-rule";
         }
```
pgermosen, line 4419: Same ask here - a RunInstances with a network-interface TagSpecification followed by DescribeTags --resource-type=network-interface asserting the tag actually shows up (not just DescribeNetworkInterfaces) would pin this rather than just the classification helper.

pgermosen (commented):
Traced both of greptile's follow-up findings and they're genuinely fixed — the refresh now runs before the filter pass on the same Address objects, and inferResourceType (confirmed as the helper DescribeTags actually calls) now maps eni- to network-interface. Neither has a regression test though, unlike every other fix in this PR. Left two inline notes on where I'd want them.

src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
lex00: Both tests added in 8e96b085: DescribeAddresses filtered by the post-restart public-ip must match (and the stale IP must not), and the RunInstances ENI TagSpecification test now asserts DescribeTags returns the tag under the interface's own resourceId with resource-type=network-interface.

pgermosen (approved):
Approving. Both tests are exactly what I was after - the address one covers both directions (current IP matches, stale IP correctly returns empty), and the ENI one is better than what I suggested: pinning the assertion to the interface's actual resourceId instead of just a tag key/value match closes a gap where a different resource sharing the same tag pair could have passed by accident.

PR #2749, opened 2026-08-30
feat(organizations): add management account email configuration and tests

### LINKED ISSUE #2659: [docs] Organizations management-account email override is documented but not implemented
`docs/services/organizations.md` documents an environment variable:

| `FLOCI_SERVICES_ORGANIZATIONS_MANAGEMENT_ACCOUNT_EMAIL` | unset | Email reported for the organization's management account (`DescribeOrganization` master account, `ListAccounts`). Unset falls back to the built-in default |

But `EmulatorConfig.OrganizationsServiceConfig` has no email field — only `enabled()` and `scpEnforcementEnabled()`:

```java
interface OrganizationsServiceConfig {
    @WithDefault("true")
    boolean enabled();

    @WithDefault("false")
    boolean scpEnforcementEnabled();
}
```

`OrganizationsService.java:247` hardcodes the management-account email as `"master@" + callerAccountId + ".example.com"` with no override path. So the documented config surface doesn't exist in current code — either the doc is ahead of an implementation that never landed, or a config field was dropped during a later refactor/squash while its doc entry survived.

Small fix either way: add the `managementAccountEmail()` config field (`@WithDefault` unset, falling back to the current synthesized default) and thread it into the management-account construction path, or remove the stale doc row if this isn't planned. Happy to send a small PR for the former if useful — flagging first in case there's context I'm missing (e.g. deliberate scope-narrowing since the doc was written).

matheusandre1:
## Summary

Closes: #2659 

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
+++ b/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
@@ -1402,6 +1402,8 @@ public interface EmulatorConfig {
 
         @WithDefault("false")
         boolean scpEnforcementEnabled();
+
+        Optional<String> managementAccountEmail();
     }
 
     interface AthenaServiceConfig {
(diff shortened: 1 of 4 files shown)


pgermosen (approved):
Approving. The override reuses the existing email validation, and I checked that against the real Organizations model directly (organizations/2016-11-28/service-2.json) rather than assuming it — the 6-64 length bounds and the regex are byte-for-byte what AWS's Email shape declares. Config field name matches what the docs already promised. One of greptile's three findings (the AccountAwareStorageBackend account-isolation one) doesn't hold up against the actual code and isn't part of this diff anyway; the other two are already fixed in this commit.

PR #2750, opened 2026-08-30
fix: update .gitignore to include .mission/ for shared tooling visibility

### LINKED ISSUE #2218: [BUG] repo: .mission/ tooling is hidden by a per-clone git exclude rule
## What's wrong

`.mission/` is excluded from the repository at `.git/info/exclude:11`. That file is per-clone and never committed, so the exclusion — and everything it hides — exists only on the machine where it was set up.

`.mission/` currently holds the operational tooling for driving the emulator: deploy scripts, roster checks, config sets, ledgers, and the UI recovery scripts. None of it is discoverable to anyone else. On any other clone those files simply do not exist; on this one they are invisible to `git status`, so nothing signals that they are untracked.

## Why it matters

- The recovery procedure for a stuck dashboard lived in an untracked script that destroyed the emulator as step 1. Nobody could review it, because nobody else could see it.
- Being in `.git/info/exclude` rather than `.gitignore` means the *rule itself* is undiscoverable. A teammate cloning the repo gets a working tree that behaves differently from this one with no indication why.
- Worktrees do not carry it. `git worktree add` produces a tree with no `.mission/` at all, so any workflow that depends on those scripts silently breaks inside a worktree.

## Suggested fix

Decide deliberately, then make the decision visible:

- **If `.mission/` is shared tooling** — track it. Scripts and config sets belong in the repo; move only genuinely local state (ledgers, run artifacts, credentials) into a nested ignored directory such as `.mission/.local/`.
- **If it is genuinely personal scratch** — move the rule from `.git/info/exclude` into the committed `.gitignore` with a comment explaining what lives there, so the exclusion is at least legible to everyone.

Either way the current state — a repo-shaping rule that exists on exactly one machine — is the one option that should not persist.

matheusandre1:
Fixes: #2218

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,13 @@ scripts/node_modules/
 __pycache__/
 *.pyc
 
+# .mission/ is per-machine operational scratch (deploy scripts, ledgers, run
+# artifacts) for driving the emulator locally. It was previously hidden by a
+# per-clone .git/info/exclude rule (#2218); ignoring it here makes the rule
+# visible on every clone. Tooling meant to be shared must be committed to the
+# repo proper, not kept in .mission/.
+.mission/
+
 # The Release Cut toolchain is pinned; its lockfile must be committed
 # (npm ci fails without it) despite the global package-lock.json ignore.
 !.github/release-tooling/package-lock.json


hectorvent:
Thanks, this is the right fix for #2218. A rule in `.git/info/exclude` only ever exists on the machine that set it up, so moving it into the committed `.gitignore` is what actually makes `.mission/` behave the same for everyone.

One small thing for next time rather than for this PR: a sentence in the Summary and a ticked Type of change box helps whoever reads the history later, even on a one line change. The linked issue covers the reasoning well here, so nothing is really lost.

No blockers from my side.

hectorvent (approved):

PR #2751, opened 2026-08-30
feat/redshift auth proxy

vuthehuyht:
## Summary

This PR introduces a TCP auth proxy in front of every Redshift cluster's backing PostgreSQL container. This ensures that the `Endpoint` returned by `DescribeClusters` is directly reachable from outside the Docker network (e.g., your host machine or standard JDBC clients). 

Key implementations:
- Leverages the existing `PostgresProtocolHandler` from the RDS service.
- The master password is now intercepted and validated directly at the proxy, meaning any `ModifyCluster` password changes take effect immediately for new connections without needing to restart the container.
- Cluster ports are safely allocated, kept stable across reboots, and securely managed upon deletions or snapshots restoration.
- Includes integration tests covering real JDBC connections via the auth proxy.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

- This change restores and enables the full PostgreSQL JDBC connectivity validation check within the `compatibility-tests/sdk-test-java` suite.
- The `Endpoint` returned by `DescribeClusters` acts exactly like a real AWS Redshift instance for external JDBC clients (`psql`, `psycopg`, etc.). 

## Checklist

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -192,6 +192,7 @@ jobs:
           DOCKER_GID=$(stat -c '%g' /var/run/docker.sock)
           docker run -d --name floci --network compat-net \
             -p 4566:4566 \
+            -p 7100-7199:7100-7199 \
             -v /var/run/docker.sock:/var/run/docker.sock \
             --group-add "$DOCKER_GID" \
             -e FLOCI_BASE_URL=http://floci:4566 \
@@ -201,6 +202,7 @@ jobs:
             -e FLOCI_SERVICES_LAMBDA_HOT_RELOAD_ENABLED=true \
             -e FLOCI_TLS_ENABLED=true \
             -e FLOCI_SERVICES_EC2_MOCK=true \
+            -e FLOCI_SERVICES_REDSHIFT_ENDPOINT_HOST=floci \
             floci:test-native
 
       - name: Wait for floci to be ready
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -192,7 +192,6 @@ jobs:
           DOCKER_GID=$(stat -c '%g' /var/run/docker.sock)
           docker run -d --name floci --network compat-net \
             -p 4566:4566 \
-            -p 7100-7199:7100-7199 \
             -v /var/run/docker.sock:/var/run/docker.sock \
             --group-add "$DOCKER_GID" \
             -e FLOCI_BASE_URL=http://floci:4566 \
--- a/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
@@ -16,6 +16,7 @@ import software.amazon.awssdk.services.redshift.model.DeleteClusterResponse;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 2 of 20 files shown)


.github/workflows/compatibility.yml
```diff
@@ -192,6 +192,7 @@ jobs:
           DOCKER_GID=$(stat -c '%g' /var/run/docker.sock)
           docker run -d --name floci --network compat-net \
             -p 4566:4566 \
+            -p 7100-7199:7100-7199 \
```
pgermosen: Quick question on this one - is publishing this port range to the host actually needed here? The sdk-test-java container runs with --network compat-net and reaches floci via http://floci:4566 (the Docker network alias), and FLOCI_SERVICES_REDSHIFT_ENDPOINT_HOST=floci makes the proxy's advertised endpoint floci:<port> too. Since that's container-to-container traffic on a shared network, it shouldn't need a host publish at all - -p only matters for reaching a container from outside Docker. Might be leftover caution copied from elsewhere, but if there's a reason it's needed I'd like to understand it (also flagging since .github/ is outside your usual area per CODEOWNERS, so wanted to ask rather than assume).

pgermosen (commented):
This has been through a really thorough hardening pass - traced the final state of deleteCluster, stopProxyAndReleasePortSafely and RedshiftProxyManager directly, and every one of greptile's port/proxy-leak findings genuinely holds up: deletion aborts cleanly when the proxy won't stop, a failed stop no longer drops the proxy from the registry (so a retry can still find it), and the port is only ever released once the stop actually succeeded. Left one question inline, otherwise this looks solid.

One small gap worth a follow-up, not blocking: none of those specific behaviors (abort-on-proxy-failure, retain-for-retry, leak-the-port-rather-than-double-bind-it) have a dedicated test yet, unlike most of the other fixes here. Would be worth locking them in so a future refactor can't quietly regress them.

vuthehuyht:
Good catch on `-p 7100-7199` — you're right, it's not needed. The proxy binds `0.0.0.0:<port>` inside the `floci` container, the advertised endpoint is `floci:<port>` (via `FLOCI_SERVICES_REDSHIFT_ENDPOINT_HOST`), and `sdk-test-java` runs on `compat-net`, so the JDBC connection is container-to-container over the network alias. The host publish only mattered for the host-side `curl` readiness check on 4566. I added it out of caution when reinstating the JDBC check without realizing the traffic never leaves the Docker network — removed it now (kept the
`ENDPOINT_HOST` env, which is what actually makes the in-network path work).

Also added the missing coverage for the three hardening behaviors:

- `RedshiftProxyManagerTest`: socket-close-throws path — proxy stays registered, retry reaches the same instance and only deregisters after a clean stop; the startup leak marker is retained across retries and cleared by a fresh start.
- `RedshiftServiceTest`: `deleteCluster` aborts with `InternalFailure` and leaves the container + metadata intact when the proxy won't stop, and a later retry completes once it does.

pgermosen (approved):
@vuthehuyht nice work getting this all the way through. I went back and checked the reboot/create/delete/restore rollback paths and the proxy manager's retry logic myself rather than taking the commit messages at their word, and it holds up — `stopProxy` no-ops cleanly on an already-stopped proxy so a failed cluster can't end up stuck undeletable, and the port-retention behavior on a failed reboot (keeping the port reserved against the failed cluster's own record instead of releasing it) is deliberate and pinned by a test, not a leak I'm reading into the code.

Two small things, neither blocking: all four lifecycle methods are synchronized on the service instance now, which serializes admin calls across every cluster in the account rather than just the one being touched — fine for what this is, just worth knowing if it ever matters. And `RedshiftOperationsTest` dropped its old `<Port>5440</Port>` assertion once the endpoint started advertising the proxy port, without adding anything back in its place; the unit tests already pin the proxy port range elsewhere so it's not really a gap, just a little threadbare at that layer.

Approving.

PR #2752, opened 2026-08-30
fix(apigateway): forward v2 request authorizer context

### LINKED ISSUE #2652: [BUG] HTTP API Lambda REQUEST authorizer: context is not forwarded to the integration event
## Service

API Gateway v2 (HTTP API) + Lambda

## AWS API Action

apigatewayv2 CreateAuthorizer - REQUEST type, payload format 2.0, --enable-simple-responses.

## Expected behavior

When a Lambda REQUEST authorizer returns a simple response containing context, API Gateway merges it into the integration event at requestContext.authorizer.lambda:

`{ "authorizer": { "lambda": { "foo": "bar" } } }`

Confirmed against real API Gateway (eu-west-1) with identical authorizer configuration.

## Actual behavior

The context is discarded. The integration event's requestContext has no authorizer key at all, so the response body is `{"authorizer":null}`. No error or warning is emitted — the request succeeds and the omission only surfaces in the handler.

The decision path itself is correct, which narrows this to the context merge:

## Reproduction

```bash
awsl() { aws --endpoint-url http://localhost:4566 --region us-east-1 "$@"; }
ACC=000000000000

cat > fn.js <<'EOF'
exports.handler = async (event) =>
  event.type === 'REQUEST'
    ? { isAuthorized: true, context: { foo: 'bar' } }
    : { statusCode: 200, body: JSON.stringify({ authorizer: event.requestContext.authorizer ?? null }) };
EOF
rm -f fn.zip && zip -q fn.zip fn.js

awsl lambda create-function --function-name repro --runtime nodejs20.x \
  --role arn:aws:iam::$ACC:role/test --handler fn.handler --zip-file fileb://fn.zip

API_ID=$(awsl apigatewayv2 create-api --name repro --protocol-type HTTP \
  --query ApiId --output text)

INT_ID=$(awsl apigatewayv2 create-integration --api-id $API_ID \
  --integration-type AWS_PROXY --payload-format-version 2.0 \
  --integration-uri arn:aws:lambda:us-east-1:$ACC:function:repro \
  --query IntegrationId --output text)

AUTH_ID=$(awsl apigatewayv2 create-authorizer --api-id $API_ID --name ctx \
  --authorizer-type REQUEST --identity-source '$request.header.Authorization' \
  --authorizer-payload-format-version 2.0 --enable-simple-responses \
  --authorizer-result-ttl-in-seconds 0 \
  --authorizer-uri arn:aws:apigateway:us-east-1:lambda:path/2015-03-31/functions/arn:aws:lambda:us-east-1:$ACC:function:repro/invocations \
  --query AuthorizerId --output text)

awsl apigatewayv2 create-route --api-id $API_ID --route-key 'GET /ctx' \
  --target integrations/$INT_ID --authorization-type CUSTOM --authorizer-id $AUTH_ID

awsl apigatewayv2 create-stage --api-id $API_ID --stage-name test --auto-deploy

curl -s -H 'Authorization: anything' \
  "http://localhost:4566/restapis/$API_ID/test/_user_request_/ctx"

# expected: {"authorizer":{"lambda":{"foo":"bar"}}}
# actual:   {"authorizer":null}
```

## Environment

Floci version / image tag: 1.7.0, floci/floci:latest-compat (edition: community, original_edition: floci-always-free)
Java SDK version (if applicable): n/a — Node.js 20 Lambdas, AWS CLI v2
How you're running Floci: Docker, via docker compose

## Note
Similar to issue #574, but to be applied to HTTP API v2

aniketshukla1:
## Summary

- carry successful HTTP API Lambda REQUEST authorizer context through request dispatch
- expose IAM-policy and simple-response context at `requestContext.authorizer.lambda`
- verify the proxy-event shape directly and through the existing backend Lambda integration flow

## Testing

- `./mvnw test -Dtest=BuildV2ProxyEventRequestAuthorizerContextTest,BuildV2ProxyEventJwtClaimsTest,BuildV2AuthorizerEventTrailingSlashTest` (11 tests passed)
- full main and test source compilation completed successfully
- `HttpApiRequestAuthorizerTest` requires an available Docker socket for its Lambda containers; local execution was blocked by the unavailable socket

Fixes #2652

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -1472,6 +1472,7 @@ public class ApiGatewayExecuteController {
 
         Map<String, String> jwtClaims = null;
         List<String> jwtScopes = null;
+        JsonNode requestAuthorizerContext = null;
         if ("JWT".equalsIgnoreCase(route.getAuthorizationType()) && route.getAuthorizerId() != null) {
             JwtAuthorizerResult jwtResult = enforceJwtAuthorizer(region, apiId, route, headers, uriInfo);
             if (jwtResult.errorResponse() != null) return jwtResult.errorResponse();
@@ -1480,8 +1481,12 @@ public class ApiGatewayExecuteController {
         }
 
         if ("CUSTOM".equalsIgnoreCase(route.getAuthorizationType()) && route.getAuthorizerId() != null) {
-            Response authError = enforceRequestAuthorizerV2(region, apiId, stageName, route, httpMethod, path, headers, uriInfo);
-            if (authError != null) return authError;
+            RequestAuthorizerResult authorizerResult = enforceRequestAuthorizerV2(
+                    region, apiId, stageName, route, httpMethod, path, headers, uriInfo);
+            if (authorizerResult.errorResponse() != null) {
+                return authorizerResult.errorResponse();
+            }
+            requestAuthorizerContext = authorizerResult.context();
         }
 
         if (route.getTarget() == null) {
@@ -1520,7 +1525,8 @@ public class ApiGatewayExecuteController {
 
         String requestId = UUID.randomUUID().toString();
         String eventJson = buildV2ProxyEvent(httpMethod, path, route.getRouteKey(),
-                apiId, region, stageName, headers, uriInfo, body, requestId, jwtClaims, jwtScopes);
+                apiId, region, stageName, headers, uriInfo, body, requestId,
+                jwtClaims, jwtScopes, requestAuthorizerContext);
 
         LOG.debugv("execute-api v2: {0} {1}/{2}{3} → Lambda {4}", httpMethod, apiId, stageName, path, functionName);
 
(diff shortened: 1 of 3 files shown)


pgermosen (commented):
This has a real conflict with main right now (ApiGatewayExecuteController.java and HttpApiRequestAuthorizerTest.java) - could you rebase/downmerge onto current main?

Before you do, worth checking first: #2715 merged a few minutes ago and looks like it fixes the exact same thing issue #2652 describes - a Lambda REQUEST authorizer's simple-response/IAM-policy context now lands at requestContext.authorizer.lambda, which is exactly what this PR does too. I diffed this branch against current main and the changes here are structurally the same fix (same RequestAuthorizerResult pattern, same rendering), just with minor differences in how an empty/absent context is handled.

So you may find, once rebased, that this is now redundant with what already landed. If that's the case this can just be closed; if you see something #2715 actually missed, happy to take a look at that difference specifically.

aniketshukla1:
Thanks — I compared this against merged #2715 and current main. #2715 already covers the same REQUEST-authorizer context propagation, including IAM-policy and simple-response contexts, nested/scalar/array values, and omission when no context is returned. I found no unique behavior left in this branch, so I’m closing this PR as redundant.

PR #2754, opened 2026-08-30
refactor(cloudformation): move AWS::S3::* into its own provisioner

hectorvent:
## Summary

Migrates `AWS::S3::Bucket` and `AWS::S3::BucketPolicy` into `S3CfnProvisioner`.
Behaviour-preserving: bodies ported verbatim apart from the mechanical rewrites onto
`ProvisionContext`, so the bucket name, the five `Fn::GetAtt` attributes and the generated CORS XML
are unchanged. The generated docs table is byte-identical; only the owner moved.

Monolith **7,396 → 7,315 lines** (pure deletion, nothing added); legacy switch 72 → 70 types.

**Stacked on #2744 → #2739.** Review those first.

### One difference from the leaf slice

`s3Service` **stays** on the monolith. Unlike KMS or Pipes, it has callers outside this slice:
Lambda code from an S3 location, and the ApiGatewayV2 OpenAPI body from S3. Only the two switch
arms and the provisioning methods leave. The CORS and versioning helpers do move, because all four
of `appendCorsRuleElements`' call sites are inside `applyBucketCorsConfiguration`.

Delete keeps its shape deliberately: a bucket policy has nothing to remove, and a non-empty bucket
still raises `BucketNotEmpty` so the stack reports `DELETE_FAILED` as AWS does, rather than
reporting a clean delete over a bucket that is still there.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

`refactor(cloudformation):` — no user-visible behaviour change, release-neutral.

## AWS Compatibility

N/A for the wire protocol. The bucket keeps its exact physical id and attribute keys, checked
against `aws-s3-bucket.json`:

| schema | value |
|---|---|
| `primaryIdentifier` | `BucketName` → asserted as the physical id |
| `readOnlyProperties` | `Arn`, `DomainName`, `RegionalDomainName`, `WebsiteURL` → all asserted as an exact map |

## Checklist

CloudFormation + Cloud Control: **843 tests, 0 failures, 0 errors**. `S3CfnProvisionerTest` adds 8,
covering the exact attribute map, lower-cased generated names capped at 63 characters, the CORS XML
element names (note the CFN property is `ExposedHeaders` while the S3 element is `ExposeHeader`),
CORS clearing when the block is absent, versioning, and both delete paths.

**End-to-end verified.** S3 is one of the few CFN types with real coverage, so per the migration
plan this slice ran the CDK suite against a Floci built from this working tree:

```
1..20
ok 1 CDK: S3 bucket exists
ok 10 CDK: CloudFormation stack CREATE_COMPLETE
...
✓ cdk passed   - cdk: PASS
```

That is a full synth → deploy → assert → destroy cycle, so it exercises the delete path too.

## Follow-up, deliberately not fixed here

`AWS::S3::Bucket` does not set **`DualStackDomainName`**, which `aws-s3-bucket.json` lists as
read-only, so `Fn::GetAtt` on it resolves to the literal `"LogicalId.DualStackDomainName"`.
Pre-existing, carried over unchanged.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -271,8 +271,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::RDS::DBSubnetGroup",
             "AWS::Route53::HostedZone",
             "AWS::Route53::RecordSet",
-            "AWS::S3::Bucket",
-            "AWS::S3::BucketPolicy",
             "AWS::SNS::Subscription",
             "AWS::SNS::Topic",
             "AWS::SecretsManager::Secret",
@@ -425,7 +423,6 @@ public class CloudFormationResourceProvisioner {
                 return resource;
             }
             switch (resourceType) {
-                case "AWS::S3::Bucket" -> provisionS3Bucket(resource, properties, engine, region, accountId, stackName);
                 case "AWS::SNS::Topic" -> provisionSnsTopic(resource, properties, engine, region, accountId, stackName);
                 case "AWS::SNS::Subscription" -> provisionSnsSubscription(resource, properties, engine, region);
                 case "AWS::DynamoDB::Table", "AWS::DynamoDB::GlobalTable" ->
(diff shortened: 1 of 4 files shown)


pgermosen (approved):
Approving the content. On greptile's finding about unnamed buckets getting replaced on update: pulled the pre-migration provisionS3Bucket from history and it's byte-for-byte the same generatePhysicalName call with no update/create distinction - pre-existing, not something this move introduced.

One thing before this lands: the base branch merged as a squash, so it isn't actually an ancestor of main yet - same situation as a couple of PRs back in this stack. This will need a rebase onto main once that settles.

hectorvent:
Thanks both, and thanks @pgermosen for pulling the history — I verified the same way and reached the same conclusion.

**On the unnamed-bucket finding:** it is a real bug, but it is pre-existing rather than introduced here. The pre-migration `provisionS3Bucket` has zero `getPhysicalId()` references, so it generated a fresh name on every provision too; this PR ports that byte for byte. Blocking a behaviour-preserving move on it would leave the defect in place and lose the guarantee that makes these slices reviewable, which is that they change nothing.

I audited the rest rather than treating it as a one-off, and six other provisioners have the same shape: SNS, SSM, ECR, Pipes, Firehose and CloudWatch. All are types whose name is a `createOnlyProperty` in the registry schema, so an unchanged template must keep its physical id. Fixed in #2788 with a shared `ProvisionContext.stablePhysicalName`, so later provisioners get the correct behaviour by default. SNS needed a different treatment there, since its physical id is the topic ARN and reusing it would have created a topic named after an ARN.

**On the rebase:** done. The stack is rebased onto `main` at `b76eadc98` now that #2739 has landed, and this PR is back to its own 4 files.

Issue #1965, opened 2026-07-26
[BUG] UpdateStack re-creates existing resources instead of reconciling — rolls back on "log group already exists"


bdemoss-runwise:
## Service

CloudFormation (+ CloudWatch Logs)

## AWS API Action

`CloudFormation:UpdateStack` / `ExecuteChangeSet` on a stack containing an
`AWS::Logs::LogGroup`.

## Expected behavior

Updating a stack whose resources are unchanged leaves them alone. Real
CloudFormation computes a diff and issues no create call for a resource that
already exists.

## Actual behavior

`UPDATE_ROLLBACK_COMPLETE`, with:

```
UPDATE_FAILED  <stack>  AWS::CloudFormation::Stack
The specified log group already exists: /aws/lambda/<stack>-<function>
```

Most resource types are re-created unconditionally on update rather than
reconciled. `provisionLogGroup` calls `logsService.createLogGroup(...)` with no
existence check and no conflict handling, so the second provision throws. The
`isUpdate` / prior-physical-id handling that does exist
(`CloudFormationResourceProvisioner:3044`) is only in the custom-resource path.

`AWS::Lambda::Function` *is* reconciled — documented under "Lambda Stack Updates"
in `docs/services/cloudformation.md` — so the gap is per-type rather than absent
across the board. Any template carrying an explicit `AWS::Logs::LogGroup` (which
SAM-generated lambda templates commonly do) cannot be updated.

**Second-order effect worth flagging:** after the rollback the log group survives
*outside* the stack, so subsequent `CreateStack` attempts then fail the same way
(`CREATE_FAILED … already exists`) and the stack is stuck in `ROLLBACK_COMPLETE`.
Recovery needs a manual `aws logs delete-log-group` before any deploy succeeds —
so one failed update can wedge a stack until a human intervenes.

## Reproduction

```bash
# 1. Any SAM template with an explicit log group; sam package to get a
#    deployable template, then create it.
aws --endpoint-url http://localhost:4566 cloudformation create-stack \
  --stack-name upd-repro --template-body file://packaged.yaml \
  --capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM
aws --endpoint-url http://localhost:4566 cloudformation wait \
  stack-create-complete --stack-name upd-repro

# 2. Update it — change anything, or nothing.
aws --endpoint-url http://localhost:4566 cloudformation update-stack \
  --stack-name upd-repro --template-body file://packaged.yaml \
[...]

PR #2756, opened 2026-08-30
refactor(cloudformation): move AWS::Logs::LogGroup into its own provisioner

hectorvent:
## Summary

Migrates `AWS::Logs::LogGroup` into `LogsCfnProvisioner`. Behaviour-preserving, but this is the
first slice carrying a **real update path**: the body decides between reconciling in place and
replacing, based on the prior physical id and the recorded `FlociLogGroupNameMode`, and infers that
mode from the physical id's shape for stacks persisted before the attribute existed.

Monolith **7,315 → 7,208 lines**; legacy switch 70 → 69 types. Docs table unchanged, as expected.

**Stacked on #2754 → #2744 → #2739.** Review those first.

### Three things this slice needed that earlier ones did not

**`UPDATE_ROLLBACK_RESTORED_ATTR` moved to `CfnRollback`.** It was package-private on the monolith,
so unreachable from the `provisioners` package. `CfnRollback` exists for exactly this (it was added
to solve the same problem for the IAM role rollback constants). The monolith's constant now
delegates there, so `CloudFormationService`'s two references are untouched.

**Uses `ctx.priorPhysicalId()` rather than reading it back off the `StackResource`.** The two are
equal on entry, but the method assigns the new physical id before it finishes, so a
context-captured value cannot later be invalidated by an edit to the body. This is the first real
use of the accessor added in #2732.

**The existing regression test stays pointed at the dispatcher**, now with the provisioner
registered. That is load-bearing, not decoration, and I verified it by mutation: dropping the
registration makes both tests fail with

```
expected: <test-stack-LogGroup-abc123def456> but was: <LogGroup-8df62014>
```

which is the dispatcher's stub arm answering instead of the provisioner. Testing through the
dispatcher also covers the plumbing this migration changes, that `existingPhysicalId` and
`existingAttributes` reach `ProvisionContext`.

`reconcileLogGroup` moves with the slice; `isGeneratedName` and its prefix helper are copied,
because a not-yet-migrated type still calls them. `logsService` leaves the monolith entirely (its
constructor parameter stays, unused, until the endgame).

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A for the wire protocol. `aws-logs-loggroup.json` gives `primaryIdentifier` `LogGroupName` and
read-only `Arn`; both are asserted, `Arn` in the exact
`arn:aws:logs:<region>:<account>:log-group:<name>:*` form AWS uses.

The AWS behaviour this preserves is that `LogGroupName` is not updatable in place: a rename creates
the replacement *before* deleting the original, so a colliding name fails the update without
destroying the existing group.

## Checklist

CloudFormation + Cloud Control: **854 tests, 0 failures, 0 errors**. `LogsCfnProvisionerTest` adds
11 covering both update branches, retention set and cleared, tag removal on reconcile, an
unparseable `RetentionInDays`, the explicit-name-removed replacement, the generated-name-kept
no-op, and the failed-rename case that must leave the original intact and flag the rollback. The
#1965 legacy-name-mode regression test is preserved unchanged apart from the registration line.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -139,7 +139,7 @@ public class CloudFormationResourceProvisioner {
     private static final String LAMBDA_CODE_IDENTITY_ATTR = "FlociLambdaCodeIdentity";
     private static final String LAMBDA_NAME_MODE_ATTR = "FlociLambdaFunctionNameMode";
     private static final String LAMBDA_PACKAGE_TYPE_ATTR = "FlociLambdaPackageType";
-    static final String UPDATE_ROLLBACK_RESTORED_ATTR = "__FlociUpdateRollbackRestored";
+    static final String UPDATE_ROLLBACK_RESTORED_ATTR = CfnRollback.UPDATE_ROLLBACK_RESTORED_ATTR;
     static final String UPDATE_ROLLBACK_FAILURE_ATTR = "__FlociUpdateRollbackFailure";
     private static final String INLINE_CLEANUP_POLICY_NAME_ATTR = "__FlociInlineCleanupPolicyName";
     private static final String INLINE_CLEANUP_ROLE_TARGETS_ATTR = "__FlociInlineCleanupRoleTargets";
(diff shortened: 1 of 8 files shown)


pgermosen (approved):
Approving. Diffed the new provision() body against the old monolith's provisionLogGroup/reconcileLogGroup line by line - it's identical logic, just reading through ProvisionContext instead of direct fields and parameters. The CfnRollback constant move is a real delegation with the same string value, isGeneratedName correctly stays on the monolith since Lambda still needs it, and the dispatcher-level regression test change is exactly what you said: registering the provisioner, nothing else. Nice first real update-path slice.

PR #2758, opened 2026-08-30
refactor(cloudformation): move Kinesis and CloudWatch to their own provisioners

hectorvent:
## Summary

Migrates `AWS::Kinesis::Stream` into `KinesisCfnProvisioner` and `AWS::CloudWatch::Alarm` into
`CloudWatchCfnProvisioner`. Behaviour-preserving; docs table unchanged.

Two unrelated services in one slice because each is a single service call with property mapping and
no shared state. Splitting them would cost two review cycles for one mechanical change.

Monolith **7,208 → 7,043 lines**; legacy switch 69 → 67 types.

**Stacked on #2756 → #2754 → #2744 → #2739.** Review those first.

### Kinesis carries an update path

An unchanged name reconciles stream mode, retention and tags rather than calling `createStream`
again, which the service rejects with `ResourceInUseException`. Retention only moves in the
direction it needs to (`increase`/`decrease`, never both, never on a no-op). A stream deleted out of
band falls back to a fresh create rather than failing. As in the log group slice, it uses
`ctx.priorPhysicalId()` rather than re-reading the `StackResource`, since the body assigns the new
id before it returns.

### Helper disposition

`sameNameExistingResource`, `deleteRenamedResource` and `parseIntProp` are **copied**, because each
still has callers left in the monolith (7, 7 and 6 respectively). `addAlarmActions` **moves**, since
all three of its call sites are inside the alarm body. Both service fields leave the monolith; their
constructor parameters stay unused until the endgame.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A for the wire protocol. Both types keep their exact physical id and attribute keys, checked
against the registry schemas:

| Type | `primaryIdentifier` | `readOnlyProperties` | asserted |
|---|---|---|---|
| `Kinesis::Stream` | `Name` | `Arn` (+ `WarmThroughputObject`, not emulated) | physical id + exact attribute map |
| `CloudWatch::Alarm` | `AlarmName` | `Arn` | physical id + `Arn` present |

AWS defaults preserved and now pinned: `Period` 60, `EvaluationPeriods` 1, `DatapointsToAlarm`
defaulting to `EvaluationPeriods` rather than a fixed number, `ActionsEnabled` true unless
explicitly disabled, and Kinesis `ShardCount` 1 when absent or unparseable.

## Checklist

CloudFormation + Cloud Control: **873 tests, 0 failures, 0 errors** (+19). `KinesisCfnProvisionerTest`
adds 11 covering both update branches, the retention direction logic, tag removal on reconcile, the
deleted-out-of-band fallback, rename with best-effort cleanup, and that a failed cleanup does not
fail the resource. `CloudWatchCfnProvisionerTest` adds 8 covering the defaults above, dimensions,
all three action lists, and an unparseable threshold.

**The existing `CloudFormationKinesisIntegrationTest` and `CloudFormationCloudWatchAlarmIntegrationTest`
are `@QuarkusTest`s resolved through CDI, so they exercise the new provisioners with no edit at all.**
That is the real regression evidence: an unmodified integration suite passing against relocated code.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -225,7 +225,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::Batch::JobQueue",
             "AWS::CloudFormation::CustomResource",
             "AWS::CloudFront::Distribution",
-            "AWS::CloudWatch::Alarm",
             "AWS::Cognito::UserPool",
             "AWS::Cognito::UserPoolClient",
             "AWS::DynamoDB::GlobalTable",
@@ -256,7 +255,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::IAM::ManagedPolicy",
             "AWS::IAM::Policy",
             "AWS::IAM::User",
-            "AWS::Kinesis::Stream",
             "AWS::Lambda::EventSourceMapping",
             "AWS::Lambda::Function",
             "AWS::Lambda::LayerVersion",
@@ -307,8 +305,6 @@ public class CloudFormationResourceProvisioner {
     private final Ec2Service ec2Service;
     private final RdsService rdsService;
     private final EksService eksService;
-    private final KinesisService kinesisService;
-    private final CloudWatchMetricsService cloudWatchMetricsService;
     private final AutoScalingService autoScalingService;
     private final DocDbService docDbService;
     private final CloudFrontService cloudFrontService;
@@ -374,8 +370,6 @@ public class CloudFormationResourceProvisioner {
         this.ec2Service = ec2Service;
         this.rdsService = rdsService;
         this.eksService = eksService;
-        this.kinesisService = kinesisService;
-        this.cloudWatchMetricsService = cloudWatchMetricsService;
         this.autoScalingService = autoScalingService;
         this.docDbService = docDbService;
         this.cloudFrontService = cloudFrontService;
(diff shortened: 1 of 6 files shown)


pgermosen (approved):
Approving. Traced the Kinesis update path directly - the retention direction logic matches AWS's real Increase/DecreaseStreamRetentionPeriod split, the deleted-out-of-band fallback and rename-with-cleanup are both correct.

One thing worth flagging, pre-existing rather than introduced here: KinesisCfnProvisioner reuses ctx.priorPhysicalId() for an unnamed stream so its identity survives an update, but CloudWatchCfnProvisioner doesn't - an unnamed alarm gets a fresh random name on every UpdateStack. Checked both against the pre-migration monolith and this asymmetry already existed there, so it's carried over faithfully rather than new. Worth a fix whenever CloudWatchCfnProvisioner gets touched again.

hectorvent:
Thanks @pgermosen — checked the asymmetry the same way and it holds. In the pre-migration monolith `provisionCloudWatchAlarm` called `generatePhysicalName` unconditionally while `provisionKinesisStream` read `r.getPhysicalId()` first, so both are carried over exactly as they were and neither is new here.

That one is already fixed further up this chain, in #2788: it adds `ProvisionContext.stablePhysicalName` (template's name, else the name the resource already had, else generated) and applies it to the seven provisioners with that shape, `CloudWatchCfnProvisioner` among them. The schemas back it up — `AlarmName` is a `createOnlyProperty`, so an unchanged template must keep its physical id. Greptile raised the same class of thing on the S3 PR, which is what prompted the audit.

No changes to this PR, so it stays a behaviour-preserving move.

One caveat on the bot review above, for anyone reading it later: its 5/5 was computed before this PR was retargeted from `refactor/cfn-logs-provisioner` to `main`. Retargeting changes the diff without changing the head SHA, so the "last reviewed commit" still matches while the analysis covered a larger diff — which is why its summary mentions Athena, Step Functions and EC2 files that this PR does not touch. The current diff is the six files listed above.

PR #2759, opened 2026-08-30
refactor(cloudformation): move AWS::SNS::* into its own provisioner

hectorvent:
## Summary

Migrates `AWS::SNS::Topic` and `AWS::SNS::Subscription` into `SnsCfnProvisioner`, completing the
Wave 1 leaf slices. Behaviour-preserving; docs table unchanged.

Monolith **7,043 → 6,991 lines** (now under 7k); legacy switch 67 → 65 types.

**Stacked on #2758 → #2756 → #2754 → #2744 → #2739.** Review those first.

### The care in this one is attribute encoding

`FilterPolicy` and `RedrivePolicy` are JSON documents that SNS expects as JSON strings, so they go
through `resolveJsonAttribute`. `FilterPolicyScope` and `RawMessageDelivery` are scalars and go
through `resolve`. Using the scalar resolver on a document flattens it to an empty string and
silently breaks message filtering, which is a failure this type has had before.

So the test stubs the two resolvers with the semantics that distinguish them and asserts the
serialized form, not merely that an attribute was set. **Mutation-verified** by swapping
`resolveJsonAttribute` for `resolve` on `FilterPolicy`:

```
expected: <{"eventType":["created","updated"]}> but was: <>
```

The topic's physical id is the ARN rather than the name, because `Ref` on an SNS topic returns the
ARN, unlike most types.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A for the wire protocol; ported verbatim.

| Type | schema `primaryIdentifier` | schema `readOnlyProperties` | Floci sets |
|---|---|---|---|
| `SNS::Subscription` | `Arn` | `Arn` | `Arn` ✅ |
| `SNS::Topic` | `TopicArn` | `TopicArn` | `Arn`, `TopicName` ⚠️ |

## Follow-up, deliberately not fixed here

**`AWS::SNS::Topic` does not set the `TopicArn` attribute** that `aws-sns-topic.json` declares
read-only. It sets `Arn`, which the schema does not list. `Fn::GetAtt` has no aliasing
(`CloudFormationTemplateEngine#resolveGetAttParts` returns `logicalId + "." + attrName` on a miss),
so a template doing `Fn::GetAtt [Topic, TopicArn]` receives the literal string `"Topic.TopicArn"`.

Pre-existing and carried over unchanged, since a move PR should not change behaviour. The fix is
one line adding `TopicArn` alongside `Arn`, and `Arn` should probably stay for compatibility with
templates already relying on it.

## Checklist

CloudFormation + Cloud Control: **882 tests, 0 failures, 0 errors** (+9). `SnsCfnProvisionerTest`
covers both JSON attributes, both scalar attributes, that unset optionals are omitted rather than
sent as empty strings, the generated topic name, and that delete routes each type to its own call.
The existing SNS integration coverage is in `@QuarkusTest`s resolved through CDI, so it exercises
the new provisioner unchanged.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -267,8 +267,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::RDS::DBSubnetGroup",
             "AWS::Route53::HostedZone",
             "AWS::Route53::RecordSet",
-            "AWS::SNS::Subscription",
-            "AWS::SNS::Topic",
             "AWS::SecretsManager::Secret",
             "AWS::SecretsManager::SecretTargetAttachment",
             "AWS::StepFunctions::StateMachine",
@@ -285,7 +283,6 @@ public class CloudFormationResourceProvisioner {
     private static final Duration CR_RESPONSE_TIMEOUT = Duration.ofSeconds(10);
 
     private final S3Service s3Service;
-    private final SnsService snsService;
     private final DynamoDbService dynamoDbService;
     private final LambdaService lambdaService;
     private final IamService iamService;
@@ -350,7 +347,6 @@ public class CloudFormationResourceProvisioner {
                                              EmulatorConfig config) {
         this.config = config;
         this.s3Service = s3Service;
-        this.snsService = snsService;
         this.dynamoDbService = dynamoDbService;
         this.lambdaService = lambdaService;
         this.iamService = iamService;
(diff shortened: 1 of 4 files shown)


hectorvent:
The finding is real but pre-existing, so it is not a change this PR should make.

`provisionSnsTopic` in the pre-migration monolith read:

```java
String topicName = resolveOptional(props, "TopicName", engine);
if (topicName == null || topicName.isBlank()) {
    topicName = generatePhysicalName(stackName, r.getLogicalId(), 256, false);
}
```

No prior-id check either, so an unnamed topic was already renamed on every UpdateStack. This PR ports that verbatim. Each slice in this migration is behaviour-preserving on purpose: that property is what makes the diffs reviewable and lets a regression bisect to one move, so folding a fix in here would cost more than it gains.

It is fixed in #2788, further up this chain. SNS needed different handling from the other six provisioners there: a topic's physical id is its ARN, so reusing it would have created a topic named after an ARN. The prior name comes from the `TopicName` attribute recorded at create time instead, covered by `anUnnamedTopicKeepsItsNameAcrossUpdates`.

Worth noting for the follow-up review: the same class of defect was reported on the S3 and Kinesis/CloudWatch PRs. Rather than treat each as a one-off, I audited every migrated provisioner, found seven with the same shape, and fixed them together behind `ProvisionContext.stablePhysicalName`. The schemas back it up — these names are `createOnlyProperty`, so an unchanged template must keep its physical id.

pgermosen (approved):
Approving. Diffed the new provisionTopic/provisionSubscription against the old monolith methods - identical logic, ctx accessors in place of direct params. Both things greptile flagged (rename-on-update, Arn vs TopicArn) are confirmed pre-existing in the monolith, and the update-identity one is already covered by #2788 further up the chain. Dispatcher removal and the inventory TSV are symmetric with what moved, and the JSON-vs-scalar attribute test coverage is exactly where the real risk in this slice is.

PR #2760, opened 2026-08-30
test(cloudformation): wire fixture provisioners from the named services

hectorvent:
## Summary

`CfnProvisionerFixture` defaulted to an **empty registry**, which made every fixture-based test
quietly fragile across a migration.

A test that named a service and provisioned one of its types kept passing when that type moved into
a provisioner: the empty registry sent it to the dispatcher's stub arm, which reports
`CREATE_COMPLETE` with a synthetic physical id, so the assertions ran against a resource nothing had
provisioned. #2756 hit exactly this and needed an explicit registration line to stay honest.

`build()` now infers the provisioners CDI would discover for the services the test named:

```java
CfnProvisionerFixture.builder()
        .logs(logsService)   // wires LogsCfnProvisioner
        .ec2(ec2Service)     // wires all six Ec2*CfnProvisioners
        .build();
```

Twenty-nine of the thirty-one provisioners take exactly one service, which is what makes the
inference unambiguous; one service can back several, as `Ec2Service` does. `provisioners(...)` still
takes an explicit set and now *replaces* the inferred one, for the two the fixture cannot build
itself (`CodeBuild` and `CodePipeline` take collaborators beyond a single service).

An earlier revision of this PR claimed twenty-eight of thirty. That was a count of constructor
arity, not of coverage, and @pgermosen was right to check it against the classes on disk.

**Stacked on #2759 → #2758 → #2756 → #2754 → #2744 → #2739.** Review those first.

### Why this matters now rather than later

Wave 1 migrated 14 types and hit the trap once. Waves 2-5 migrate the remaining 65, and the tests
most exposed are the largest ones still building the provisioner positionally: `RdsCfnProvisionerTest`
(1,200 lines), `CloudFormationIamAttachmentProvisionerTest` (746),
`SecretTargetAttachmentCfnProvisionerTest` (681), `ApiGatewayV2CfnProvisionerTest` (511), plus the
DynamoDB, Events and Lambda tests. Each would otherwise need the mistake spotted by hand, and each
would fail silently if it were not.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

Test-only; no production code changes.

## AWS Compatibility

N/A. Test infrastructure only.

## Checklist

CloudFormation + Cloud Control: **889 tests, 0 failures, 0 errors**. `CfnProvisionerFixtureTest`
adds 6. Five pin the inference: a named service wires its provisioner, an unnamed one does not, one
service can wire several, the service-less `CdkMetadataCfnProvisioner` is always available, and an
explicit set replaces inference rather than adding to it.

The sixth, `everyProvisionerIsReachableThroughThePublicBuilderOrExplicitlyExempt`, is the gate that
stops this fixture from reproducing the very failure it exists to prevent. It drives every public
single-argument `Builder` setter, builds the registry, and reads back which provisioners actually
resolved, then fails on any class on disk that is neither reachable nor in a documented
`NOT_INFERABLE` set. It also fails on a stale exemption, so the list cannot rot.

Resolving reachability rather than reading it matters, and the first attempt got this wrong: it
matched `new (\w+CfnProvisioner)\(` against the fixture's source, which scores a construction arm
as wired even when its field has no setter and can never be non-null. Six provisioners were in
exactly that state. Verified the replacement bites by deleting one setter while leaving its field
and construction arm in place, which the textual version passed:

```
These provisioners cannot be reached through any public Builder setter ... : [WafV2CfnProvisioner]
```

**Verified load-bearing rather than incidental.** Neutering the inference puts #2756's log group
test back on the stub arm:

```
expected: <test-stack-LogGroup-abc123def456> but was: <LogGroup-bf9816a5>
```

That test reverts to naming its service, dropping the explicit registration it needed before.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CfnProvisionerFixture.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CfnProvisionerFixture.java
@@ -161,7 +161,8 @@ final class CfnProvisionerFixture {
         }
 
         /**
-         * The provisioners CDI would discover for the services this test named.
+         * The provisioners inferred from the services this test named, not everything CDI would
+         * discover: a service the test did not name contributes nothing here.
          *
          * <p>Inferring these is what keeps a test honest across a migration. A test that names a
          * service and provisions one of its types used to keep passing when that type moved to a
@@ -410,6 +411,36 @@ final class CfnProvisionerFixture {
 
         public Builder cloudFront(CloudFrontService v) {
             this.cloudFrontService = v;
+            return this;
+        }
+
+        public Builder flowLog(FlowLogService v) {
+            this.flowLogService = v;
+            return this;
+        }
+
+        public Builder lambdaMicrovms(LambdaMicrovmsService v) {
+            this.lambdaMicrovmsService = v;
+            return this;
+        }
+
+        public Builder awsConfig(AwsConfigService v) {
+            this.awsConfigService = v;
+            return this;
+        }
+
+        public Builder organizations(OrganizationsService v) {
+            this.organizationsService = v;
+            return this;
+        }
+
+        public Builder sqs(SqsService v) {
+            this.sqsService = v;
+            return this;
+        }
+
+        public Builder wafV2(WafV2Service v) {
+            this.wafV2Service = v;
             return this;
         }
 
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CfnProvisionerFixture.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CfnProvisionerFixture.java
@@ -8,6 +8,41 @@ import io.github.hectorvent.floci.services.apigatewayv2.ApiGatewayV2Service;
 import io.github.hectorvent.floci.services.autoscaling.AutoScalingService;
 import io.github.hectorvent.floci.services.batch.BatchService;
 import io.github.hectorvent.floci.services.cloudformation.provisioners.CfnDynamicReferences;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.ConfigCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.Ec2FlowLogCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.LambdaMicrovmsCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.OrganizationsCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.SqsCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.WafV2CfnProvisioner;
+import io.github.hectorvent.floci.services.configservice.AwsConfigService;
+import io.github.hectorvent.floci.services.ec2.FlowLogService;
+import io.github.hectorvent.floci.services.lambdamicrovms.LambdaMicrovmsService;
+import io.github.hectorvent.floci.services.organizations.OrganizationsService;
+import io.github.hectorvent.floci.services.sqs.SqsService;
+import io.github.hectorvent.floci.services.wafv2.WafV2Service;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.ApiGatewayAccountCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.AutoScalingLifecycleHookCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.CdkMetadataCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.CloudWatchCfnProvisioner;
+import io.github.hectorvent.floci.services.cloudformation.provisioners.Ec2LaunchTemplateCfnProvisioner;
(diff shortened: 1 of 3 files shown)


src/test/java/io/github/hectorvent/floci/services/cloudformation/CfnProvisionerFixture.java
```diff
@@ -97,18 +121,92 @@ static final class Builder {
...
+        /**
+         * The provisioners CDI would discover for the services this test named.
+         *
+         * <p>Inferring these is what keeps a test honest across a migration. A test that names a
+         * service and provisions one of its types used to keep passing when that type moved to a
+         * provisioner: the empty registry sent it to the dispatcher's stub arm, which reports
+         * CREATE_COMPLETE with a synthetic id, so the assertions ran against nothing. Wiring the
+         * provisioner from the service means the test follows the type instead.
+         */
+        private List<CfnResourceProvisioner> inferredProvisioners() {
```
pgermosen, line 179: This covers 23 provisioners, not the 28 the PR description claims. Ec2FlowLogCfnProvisioner isn't wired by .ec2(...) (there are 7 Ec2*CfnProvisioner classes now, not 6), LambdaMicrovmsCfnProvisioner isn't wired by .lambda(...), and ConfigCfnProvisioner/OrganizationsCfnProvisioner/SqsCfnProvisioner/WafV2CfnProvisioner have no service field on the Builder at all, so there's no way to wire them through inference.

Nothing hits this today - no CfnProvisionerFixture user currently provisions those six types - but it's the exact failure mode this PR exists to prevent, now living inside the fixture itself. A wave-2-5 test that assumes .ec2(...) or .lambda(...) covers "all" its provisioners (reasonable, given the doc comment above says so) would silently land on the stub arm for these specific types, and CfnProvisionerFixtureTest wouldn't catch it either - it only asserts the types that are wired, not an exhaustiveness check against the real CDI registry the way CfnResourceInventoryTest does for the production registry.

Worth either fixing the four missing service fields + FlowLog/Microvms, or adding a completeness test that fails when inferredProvisioners() falls behind the actual provisioner set.

pgermosen (commented):
The inference idea is right and I like the design, but I counted the coverage directly against the actual provisioner classes and it's short of what the PR claims. Left a note on where the gap is.

hectorvent:
Good catch @pgermosen, and you are right on the count: 23 wired against 31 provisioner classes, not 28. The "28 of 30" in my description was about constructor arity — how many take a single service — and I let it read as a coverage claim. The doc comment on `inferredProvisioners()` made that worse by saying "the provisioners CDI would discover", which implies completeness it did not have.

Your framing is the important part: this is the same silent-stub failure the PR exists to prevent, relocated into the fixture, and the existing tests could not catch it because they only assert the types that *are* wired.

Applied both halves of your suggestion:

**Wired the six that take a single service** — `Ec2FlowLogCfnProvisioner` (needs `FlowLogService`, not `Ec2Service`, which is why `.ec2(...)` missed it), `LambdaMicrovmsCfnProvisioner`, `ConfigCfnProvisioner`, `OrganizationsCfnProvisioner`, `SqsCfnProvisioner` and `WafV2CfnProvisioner`. Four had no Builder field at all; `sqsService` had been removed earlier when it became a dead constructor argument, so it is back as an inference-only field.

**Added the completeness gate**, `everyProvisionerIsWirableOrExplicitlyExempt`. It lists the provisioner classes on disk and fails unless each is either wired or in a documented `NOT_INFERABLE` set, which holds only `CodeBuildCfnProvisioner` and `CodePipelineCfnProvisioner` — both take a `RegionResolver` and an `ObjectMapper` beyond their service, so they stay explicit. It also fails on a stale exemption, so the list cannot rot. Same shape as `CfnResourceInventoryTest` for the production registry.

Verified it bites: unwiring `WafV2CfnProvisioner` fails with

```
These provisioners are neither wired from a service nor exempt ... : [WafV2CfnProvisioner]
```

Also corrected the doc comment to state the actual contract. 908 tests green.

pgermosen:
@hectorvent thanks for the write-up — the fix you describe (the six setters, the completeness gate, the corrected doc comment) sounds like exactly the right response. But I'm not seeing any of it on the branch: it's still a single commit, 8f3adf80, same as before your reply, and pulling that file at that exact SHA still shows no `.flowLog(...)`/`.lambdaMicrovms(...)`/`.awsConfig(...)`/`.organizations(...)`/`.sqs(...)`/`.wafV2(...)` setters, and the `inferredProvisioners()` doc comment still opens with "The provisioners CDI would discover for the services this test named" — the line your reply says got corrected.

Might just be a push that didn't go through — could you check on your end and push if it's sitting locally or on another branch?

pgermosen (approved):
@hectorvent confirmed, it's all there now and I didn't just take the new test's word for it. All six setters exist and genuinely make the six previously-orphaned registrations reachable. I pulled the full list of provisioner classes straight from main (31), diffed it against every registration call in the current inferredProvisioners() (29), and the gap is exactly the two declared exemptions, nothing missing, nothing extra. Then cross-checked every field the inference logic null-checks against every field a setter actually assigns, full match.

The rebuild of the completeness test from a textual regex check to a behavioral one (reflectively driving the public setters, building the real registry, checking what's reachable) is a real improvement too, that's exactly what closes the gap where a construction call could look wired in source without ever being reachable from a test. Good to go.

PR #2761, opened 2026-08-30
fix(cloudformation): set Fn::GetAtt attributes the schemas declare

hectorvent:
## Summary

An attribute a provisioner never sets is **invisible at runtime**. `resolveGetAttParts` returns the
literal `"LogicalId.Attr"` when a resource has no such attribute, so a template doing
`Fn::GetAtt [Topic, TopicArn]` receives the string `"Topic.TopicArn"` as a value rather than
failing. Nothing catches it.

Auditing every migrated type against its CloudFormation registry schema found **24 such attributes
across 15 types**. This sets the three whose correct value is unambiguous, and makes the rest a
ratchet instead of a silent backlog.

**Stacked on #2760 → #2759 → … → #2739.** Review those first.

### Fixed

| Type | Attribute | Value |
|---|---|---|
| `AWS::SNS::Topic` | `TopicArn` | the topic ARN — the attribute `aws-sns-topic.json` actually declares. `Arn` stays alongside it, since templates written against earlier releases reference it |
| `AWS::S3::Bucket` | `DualStackDomainName` | `<bucket>.s3.dualstack.<region>.amazonaws.com`, the same form as the `RegionalDomainName` already built |
| `AWS::CodePipeline::Pipeline` | `Arn` | constructed from region, account and name; the schema describes it as "The Amazon Resource Name (ARN) of the pipeline" |

### Not fixed, and why

The other 21 are recorded in `getatt-attribute-gaps.tsv`, each with a reason. Most are attributes
Floci *cannot* set because it does not emulate the behaviour behind them (`EC2::VPC`
`CidrBlockAssociations`, `Pipes::Pipe` lifecycle timestamps, `Kinesis` `WarmThroughputObject`).

A few I deliberately left despite looking easy — `Lambda::Permission Id`, `CodeBuild::Project Id`,
`SQS::QueuePolicy Id`, `EC2::VPCGatewayAttachment AttachmentType`. Their schemas carry no
description, and I could not establish AWS's actual value from the sources in the repo. Inventing a
plausible-looking value is worse than a recorded gap: it would look correct and read wrong.

### The ratchet

`CfnSchemaCoverageTest` compares the recorded gaps against what the provisioner sources actually
set, so the file cannot drift:

- a provisioner that stops setting a declared attribute **fails**, naming it
- a gap that has been closed **fails** until its row is deleted
- a row without a reason **fails**, so "not emulated" stays distinguishable from "forgotten"

**Mutation-verified in both directions.** Removing the new `TopicArn` line:

```
New gaps ... — set them, or add a row with a reason to ...getatt-attribute-gaps.tsv:
    AWS::SNS::Topic	TopicArn
```

Leaving a stale row after a fix:

```
These attributes are now set, so their rows ... are stale and should be deleted:
    [AWS::S3::Bucket	WebsiteURL]
```

The audit reads provisioner sources class-wide, so it is conservative: a reported gap is genuinely
unset by the whole class. It covers migrated types only; the legacy switch is audited type by type
as each one moves.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behaviour: `Fn::GetAtt` against `AWS::SNS::Topic` `TopicArn`,
`AWS::S3::Bucket` `DualStackDomainName`, or `AWS::CodePipeline::Pipeline` `Arn` resolved to the
literal string `"LogicalId.Attr"` instead of a value. Verified against the registry schemas in
`local/aws/cfn-resource-schemas/us-east-1`, which declare each as `readOnlyProperties`.

`AWS::SNS::Topic` keeps emitting `Arn` as well. The schema does not declare it, but removing it
would break templates already relying on it, and an extra attribute is not observable to a template
that does not ask for it.

## Checklist

CloudFormation + Cloud Control: **890 tests, 0 failures, 0 errors**. The existing
`S3CfnProvisionerTest` and `SnsCfnProvisionerTest` exact-attribute-map assertions were updated to
include the new keys, so they pin the fix rather than tolerating it.

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CfnSchemaCoverageTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CfnSchemaCoverageTest.java
@@ -1,0 +1,224 @@
+package io.github.hectorvent.floci.services.cloudformation;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * Pins which {@code Fn::GetAtt} attributes declared by the CloudFormation registry schemas Floci
+ * does not set.
+ *
+ * <p>A missing attribute is invisible at runtime: {@code resolveGetAttParts} returns the literal
+ * {@code "LogicalId.Attr"} when a resource has no such attribute, so a template referencing one
+ * receives that string as a value instead of failing. Recording the gaps makes the set a ratchet:
+ * closing one deletes a row, and introducing one has to add a row where a reviewer sees it.
+ *
+ * <p>Not a parity assertion. Several rows are attributes Floci cannot set because it does not
+ * emulate the underlying behaviour, and each carries the reason.
+ */
+class CfnSchemaCoverageTest {
+
+    private static final Path INVENTORY =
+            Path.of("src/test/resources/cloudformation/supported-resource-types.tsv");
+    private static final Path GAPS =
+            Path.of("src/test/resources/cloudformation/getatt-attribute-gaps.tsv");
+    private static final Path SCHEMA_DIR = Path.of("local/aws/cfn-resource-schemas/us-east-1");
+    private static final Path PROVISIONERS = Path.of(
+            "src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners");
+
+    private static final Pattern PUT_ATTRIBUTE =
+            Pattern.compile("getAttributes\\(\\)\\.put\\(\\s*\"([^\"]+)\"");
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    /**
+     * The check needs the schema corpus, which lives under the gitignored {@code local/}. Absent in
+     * CI, so this is a local correctness aid; the checked-in gaps file is what CI reviews.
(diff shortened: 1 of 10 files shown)


src/test/java/io/github/hectorvent/floci/services/cloudformation/CfnSchemaCoverageTest.java
```diff
@@ -0,0 +1,224 @@
...
+        return gaps;
+    }
+
+    /**
+     * Attributes each provisioner class sets, read from its source.
+     *
+     * <p>Class-wide rather than per-type, because one class can serve several types. That makes the
+     * result conservative: a gap it reports is genuinely unset by the whole class.
+     */
+    private static Map<String, Set<String>> attributesSetByEachProvisioner() {
```
pgermosen: Agreeing with greptile's finding here rather than just noting it: I checked SnsCfnProvisioner specifically - provisionTopic sets TopicArn/Arn/TopicName, provisionSubscription sets only Arn, but this scan aggregates by file, so the audit would credit Subscription with all three even though it only sets one. There are at least 10 other multi-type provisioners in this migration (KMS, SQS, Lambda, EC2, Organizations, ...) exposed to the same gap.

This isn't a necessary cost of grouping multiple types per file for review-size reasons - that's a good tradeoff and I'm not asking to undo it. But the scanner could still be precise: walk from each `case TYPE -> methodName(...)` dispatch and scan only that method's body (or what it transitively calls) instead of sweeping the whole file. That keeps one file per provisioner while making the per-type attribution actually correct, which is the guarantee this whole PR is introducing the ratchet for.

Could you scope the scan to each resource type's own provisioning path before this merges?

pgermosen (commented):
Left one ask inline before this can close the loop it's meant to close. The three concrete attribute fixes (SNS TopicArn, S3 DualStackDomainName, CodePipeline Arn) look right.

pgermosen (approved):
Approving. This genuinely closes the loop rather than just claiming to - the scan now walks provision()'s dispatch and attributes each arm's writes (plus whatever it transitively calls) to its own type, with a loud failure instead of a silent fallback when a type is missing an arm. It even correctly tells a switch statement apart from a switch expression, which matters for Ec2SecurityGroupRuleCfnProvisioner specifically. And making the scan precise immediately found a real gap - NetworkAclEntry's Id - which got fixed here rather than just logged. Good work.

PR #2762, opened 2026-08-30
docs(cloudformation): correct the resource-type contributor steps

hectorvent:
## Summary

The "Adding a CloudFormation Resource Type" steps in `CONTRIBUTING.md` and `AGENTS.md` drifted
behind the machinery this migration added. One step is now actively wrong.

**The bug:** step 7 said *"Add the type to the table in `docs/services/cloudformation.md`"*. Since
#2739 that table is **generated** from the provisioner inventory, so a contributor following the
instruction hand-edits generated output and fails `docs-check`. My oversight in #2739: I updated the
note inside the doc itself but not the instruction that sends people there.

**The omissions:** three steps are now mandatory, each with a guard that fails without it, so a
contributor currently meets the failure before the instruction:

| Step | Guard that fails without it |
|---|---|
| Register in `supported-resource-types.tsv` | `CfnResourceInventoryTest` |
| Add to `CfnProvisionerFixture.inferredProvisioners()` | a fixture test silently hits the stub arm |
| Record an unsettable attribute in `getatt-attribute-gaps.tsv` | `CfnSchemaCoverageTest` |

**Two sharpened steps**, from what Wave 1 established:

- Branch an update on `ctx.isUpdate()` / `ctx.priorPhysicalId()` rather than reading the physical id
  back off the resource — `provision` assigns the new id as it runs, so a resource-derived check
  changes meaning mid-method.
- Tolerate an already-deleted resource with `CfnDeletes.safeDelete` and **explicit** error codes,
  never a catch-all, which would swallow `BucketNotEmpty` and report a clean stack delete over a
  bucket that is still there. Where the delete needs a create-time attribute, override
  `delete(StackResource, String)`.

Also adds `LogsCfnProvisioner` as the reference for a full reconcile-vs-replace update path, and two
entries to the AGENTS.md common-mistakes list.

**Stacked on #2761 → … → #2739.** Docs-only; no Java, no behaviour change.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A. Contributor documentation only.

## Checklist

No test added: this changes prose only, and every instruction it documents already has a test
enforcing it (`CfnResourceInventoryTest`, `CfnProvisionerFixtureTest`, `CfnSchemaCoverageTest`,
`CfnDeletePrecedenceTest`). `make docs-check` passes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -283,17 +283,36 @@ monolith being dismantled; new types go in per-service provisioners under
    `local/aws/cfn-resource-schemas/us-east-1/` (`readOnlyProperties`), and validate
    `required` from the same file.
 4. **`provision` serves create *and* update.** On `UpdateStack` it is re-invoked with
-   the prior physical id and attributes already populated on the resource. Branch on
-   that instead of creating unconditionally.
-5. Override `delete(...)` when the type has a backing delete; tolerate already-deleted.
-6. Tests: focused unit test mocking one service (`SqsCfnProvisionerTest` is the
+   the prior physical id and attributes already populated on the resource. Branch with
+   `ctx.isUpdate()` / `ctx.priorPhysicalId()`, not by reading the id off the resource:
+   `provision` assigns the new id as it runs, so a resource-derived check flips
+   mid-method.
+5. Override `delete(...)` when the type has a backing delete; tolerate already-deleted
+   via `CfnDeletes.safeDelete`, passing the specific "already gone" error codes. Never
+   a catch-all: a real failure such as `BucketNotEmpty` must propagate so the stack
+   reports `DELETE_FAILED`. When the delete needs a create-time attribute rather than
+   just the physical id, override `delete(StackResource, String)`.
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Approving. Cross-checked the technical claims against what I've directly verified in the underlying migration PRs and they hold up - this fixes a genuinely wrong instruction (the table is generated now, not hand-edited) and documents real guards.

One small thing left over from greptile's pass: the "Adding a CloudFormation provisioner without a row in supported-resource-types.tsv or an entry in CfnProvisionerFixture, either of which leaves a type quietly served by the stub arm" line still conflates two different failure modes. A missing TSV row fails CfnResourceInventoryTest loudly - it doesn't produce stub-arm behavior in production, since the TSV is just an inventory check, not what wires runtime dispatch. Only a missing CfnProvisionerFixture entry does that, and only inside a test. Not blocking, just worth tightening next time this section is touched.

PR #2773, opened 2026-08-30
fix(stepfunctions): bound an execution's history and make StopExecution stop it

facuparedes:
Replaced by #2782, which carries this change together with the rest of the `AslExecutor` work.


facuparedes:
Superseded by #2782, which carries this change unmodified alongside the rest of the `AslExecutor` work. The three PRs that touched this file conflicted on every hunk; grouping them by the file they change removes the conflicts entirely. Issue #2779, the four blocking waits the execution deadline did not reach, is closed there too.

PR #2774, opened 2026-08-30
fix(stepfunctions): open a JSONata query evaluation cause with the AWS sentence

facuparedes:
Replaced by #2783, which carries this change together with the rest of the `JsonataEvaluator` work.


facuparedes:
Superseded by #2783, which carries this change alongside the other two that touch `JsonataEvaluator`. All three rewrote the same `catch` block, so each pair conflicted; grouping them by the file they change removes the conflicts and the resolution notes that went with them.

PR #2776, opened 2026-08-30
fix(stepfunctions): reject definitions with no terminal state, dangling targets or unsupported fields

facuparedes:
Replaced by #2784, which carries this change together with the rest of the `collectValidationErrors` work.


facuparedes:
Superseded by #2784, which carries this change alongside the other one that adds rules to `collectValidationErrors`. Both added a marker constant and a filter clause to the same two lines, so they conflicted in either order; grouping them by the file they change removes the conflict.

PR #2780, opened 2026-08-30
fix(eks): implement cluster restoration after restart to maintain state

### LINKED ISSUE #2609: [BUG] Re-latching cluster after docker application reboot
## Service

Docker in general

I created a `floci` instance with the terminal command, and create a cluster using aws command.

After I reboot my docker application, I restarted the floci via command line, and my cluster is not there anymore, but the image is still there.

I tried manually activating the image on Docker but it's running into issues.

I also tried to have a persistent directory. It is show that the cluster is there, but when I try to deploy a service, it doesn't work.

## Reproduction

```bash
floci start
aws eks create-cluster <cluster_name>
```

restart docker application

```bash
floci start
```

try to push an service into cluster but failed

## Environment
- floci latest with terminal installation
- docker 4.73.0 
- MacBook

matheusandre1:
## Summary

Fixes: #2609 

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [X] Docs / chore

After a Docker reboot, floci would restore the EKS cluster metadata from `eks-clusters.json` with an `ACTIVE` status, but nothing would restart the k3s container; the runtime fields (`containerId`, `hostPort`) were marked with `@JsonIgnore` and thus returned empty, so the cluster appeared in `list-clusters`, yet any deployment or `kubectl` command failed because the container was never restarted. Furthermore, during shutdown, the k3s data volume was unconditionally removed, negating persistence.

## Checklist

### DIFF
--- a/docs/services/eks.md
+++ b/docs/services/eks.md
@@ -73,6 +73,29 @@ services:
 !!! note "No port mapping needed for k3s ports"
     k3s containers bind their API server port (6500–6599) directly on the host via Docker — no `ports:` entry is required in `docker-compose.yml`. See [Ports Reference](../configuration/ports.md#ports-65006599-eks-real-mode) for the full explanation.
 
+#### Clusters survive a restart
+
+With a persistent [storage mode](../configuration/storage.md) (the default), clusters recorded in
+`eks-clusters.json` are **re-latched to their k3s containers when Floci starts**:
+
+- A surviving container (for example after a Docker Desktop / daemon reboot) is adopted and
+  started in place, keeping its published API server port and data volume — deployments come
+  back as they were.
+- A missing container is recreated. Its named k3s data volume (`floci-eks-<name>`; for a
+  [non-default account](../configuration/multi-account.md), `floci-eks-<account>.<name>`) is
+  reused if it survived; volumes follow the global prune policy
+  (`FLOCI_STORAGE_PRUNE_VOLUMES_ON_DELETE`, default `false`), so they are retained when the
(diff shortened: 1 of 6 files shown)


hectorvent:
Thank you, this closes a real gap. Restoring the metadata with `ACTIVE` while nothing restarted the k3s container meant `list-clusters` looked healthy and every kubectl call failed, which is the most confusing shape a bug like this can take.

Two things I want to call out as good rather than as findings. Reporting `CREATING` until the API server answers, and `FAILED` when the container cannot come back, is more honest than the old unconditional `ACTIVE`. And adopting the historical `floci-eks-<name>` container when the `io.floci.account` label matches means clusters created before account-qualified naming are not orphaned on upgrade.

Small thing: Type of change has both Bug fix and Docs / chore ticked. The commit is `fix:`, so nothing downstream is affected.

No blockers from my side.

hectorvent (approved):

pgermosen (approved):
Approving. Went through all six rounds of greptile findings against the final code rather than trusting the thread - every one is genuinely closed. The account recovery now comes from the storage key independently in both backfill and restore, so ordering between them can't strand an owner. The dot separator for account-qualified Docker names is provably collision-safe given the AWS charset, and that's now enforced at create time too. Legacy pre-upgrade containers are only adopted with a matching ownership label, never guessed at. And a persisted name that predates validation gets refused rather than risking a cross-bind. Every one of those has its own dedicated test. Nice work chasing this all the way through.

Issue #2735, opened 2026-08-29
[BUG] Step Functions (SFN): CreateStateMachine accepts definitions AWS rejects, including a workflow with no terminal state


facuparedes:
## Service

Step Functions (ASL definition validation; independent of query language)

## AWS API Action

`ValidateStateMachineDefinition`, `CreateStateMachine`

## Expected behavior

AWS refuses these four definitions. `ValidateStateMachineDefinition` reports each without an
execution, and `CreateStateMachine` wraps the same diagnostic in `InvalidDefinition`, e.g.
`Invalid State Machine Definition: 'MISSING_END_STATE: Workflow has no terminal state at null'`:

```
NOEND  a workflow with no terminal state       MISSING_END_STATE          Workflow has no terminal state
MAPT   TimeoutSeconds on a Map state           SCHEMA_VALIDATION_FAILED   Field 'TimeoutSeconds' is not supported  /States/M
PART   TimeoutSeconds on a Parallel state      SCHEMA_VALIDATION_FAILED   Field 'TimeoutSeconds' is not supported  /States/P
UNRE   a state nothing routes to               MISSING_TRANSITION_TARGET  State "C" is not reachable.              /States/C
```

NOEND and UNRE are two separate rules: the bare loop `{"A":{"Next":"B"},"B":{"Next":"A"}}` gives
`MISSING_END_STATE`, and adding an unreachable terminal `"C":{"Type":"Pass","End":true}` to that
same loop gives `MISSING_TRANSITION_TARGET` instead.

MAPT and PART are two cases of a general rule, measured across every state type: AWS accepts
`TimeoutSeconds` on `Task` only, and `Catch` and `Retry` on `Task`, `Parallel` and `Map` only. Every
other combination is refused with the same `Field '<name>' is not supported` diagnostic:
`TimeoutSeconds` on the seven non-`Task` types, and `Catch` and `Retry` on the five types that are
none of `Task`, `Parallel` or `Map`. Floci accepts all 17.

## Actual behavior

```
                        AWS                        Floci
validate NOEND          FAIL MISSING_END_STATE     OK, diagnostics []
validate MAPT           FAIL SCHEMA_VALIDATION     OK, diagnostics []
validate PART           FAIL SCHEMA_VALIDATION     OK, diagnostics []
validate UNRE           FAIL MISSING_TRANSITION    OK, diagnostics []
create   all four       InvalidDefinition          state machine created
```

[...]

PR #2782, opened 2026-08-30
fix(stepfunctions): bound a running execution the four ways AWS does

### LINKED ISSUE #2733: [BUG] Step Functions (SFN): `TimeoutSeconds` is never enforced, neither on the state machine nor on a `Task`
## Service

Step Functions. Both definitions below are plain JSONPath, so nothing here depends on the query
language.

## AWS API Action

`CreateStateMachine`, `StartExecution`, `DescribeExecution`, `GetExecutionHistory`, `GetActivityTask`

## Expected behavior

ASL has two `TimeoutSeconds` fields and AWS enforces both, in two different terminal shapes:

```
{"StartAt":"W","TimeoutSeconds":3,"States":{"W":{"Type":"Wait","Seconds":25,"End":true}}}
  TIMED_OUT at startDate + 3.000s, error and cause absent as keys,
  States.Timeout only inside the ExecutionTimedOut history event

Task on an activity nobody answers, "TimeoutSeconds": 5
  FAILED  States.Timeout, no cause, at 9.230s, with ActivityTimedOut in the history
```

## Actual behavior

```
                                AWS                          Floci
TimeoutSeconds: 3, Wait: 25     TIMED_OUT at +3.000s         SUCCEEDED at +25.004s
Task TimeoutSeconds: 5          FAILED States.Timeout        RUNNING until +300.051s, then
                                at +9.230s, no cause         FAILED States.HeartbeatTimeout
                                                             "Task timed out after 300 seconds"
```

Floci accepts `TimeoutSeconds`, echoes it back from `DescribeStateMachine`, copies it into
`activityScheduledEventDetails.timeoutInSeconds`, and never uses it as a deadline. `doExecute` reads
three top-level fields of the definition (`AslExecutor.java:359-361`) and `TimeoutSeconds` is not
one of them. `ExecutionTimedOut`, `ActivityTimedOut` and `TaskTimedOut` are not event types Floci
emits at all.

For the `Task` half the two fields are swapped. On AWS `TimeoutSeconds` is the total deadline and
`HeartbeatSeconds` is the maximum gap between heartbeats. In Floci `TimeoutSeconds` is decorative
and `HeartbeatSeconds` is the total deadline ([`AslExecutor.awaitToken`, line
684](https://github.com/floci-io/floci/blob/main/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java#L684)):

```java
private JsonNode awaitToken(CompletableFuture<JsonNode> future, JsonNode stateDef) throws Exception {
    int timeout = stateDef.path("HeartbeatSeconds").asInt(0);
    if (timeout <= 0) {
        timeout = 300;
    }
    try {
        return future.get(timeout, TimeUnit.SECONDS);
    } catch (java.util.concurrent.TimeoutException e) {
        future.cancel(true);
        throw new FailStateException("States.HeartbeatTimeout",
                "Task timed out after " + timeout + " seconds");
    }
```

So `"TimeoutSeconds": 5, "HeartbeatSeconds": 60` fails at 60s on Floci and at 5s on AWS. And
`StepFunctionsService.sendTaskHeartbeat` is a `LOG.debugv` and nothing else, so a worker that
heartbeats exactly as the definition asks is killed at the same instant as one that never
heartbeats.

## Reproduction

Both definitions pass `aws stepfunctions validate-state-machine-definition` with `result: OK`.

```bash

SM=$(aws stepfunctions create-state-machine --name topwait \
  --role-arn arn:aws:iam::000000000000:role/r --query stateMachineArn --output text \
  --definition '{"StartAt":"W","TimeoutSeconds":3,"States":{"W":{"Type":"Wait","Seconds":25,"End":true}}}')

sleep 30
aws stepfunctions describe-execution --execution-arn "$EX" --query status --output text
aws stepfunctions get-execution-history --execution-arn "$EX" --query 'events[].type' --output text
# SUCCEEDED
# ExecutionStarted  WaitStateEntered  WaitStateExited  ExecutionSucceeded
#     AWS returns TIMED_OUT at startDate + 3.000s, and
#                 ExecutionStarted  WaitStateEntered  ExecutionTimedOut
```

The `Task` half needs an activity, a `GetActivityTask` that picks the task up and a worker that
never answers, and takes 300 seconds to terminate on Floci.

## Suggested fix

Top level. Compute the deadline once in `doExecute` from `definition.path("TimeoutSeconds")` and
check it at the top of the state loop and inside the `Wait` sleep, so a `Wait` longer than the
remaining budget is cut rather than slept out. On expiry: `TIMED_OUT`, `stopDate` set, `error` and
`cause` left unset, and one `ExecutionTimedOut` event with `previousEventId: 0` and details
`{"error":"States.Timeout"}`, with no `*StateExited` for the state that was cut.

`Task`. Give `awaitToken` two independent bounds instead of one: `TimeoutSeconds` as the total wait,
failing with `States.Timeout` and no cause, and `HeartbeatSeconds` as a gap that `sendTaskHeartbeat`
resets. Since `sendTaskHeartbeat` only logs today, the second bound needs a per-token last-heartbeat
timestamp before it can mean anything.

The `Parallel` state at `AslExecutor.java:1911-1928` already has the shape both halves need (read
the field, compute a deadline, fail on it), but AWS rejects `TimeoutSeconds` on `Parallel` (#2735),
so reuse the shape, not the call site.

One thing blocks matching AWS exactly: Floci emits `ActivityStarted` one millisecond after
`ActivityScheduled`, with no `workerName`, before any worker has called `GetActivityTask`. AWS
emits it when the worker picks the task up, and that is the instant the `TimeoutSeconds` clock is
anchored on. Measured in `us-east-1` on a `Task` whose `TimeoutSeconds` is 20, with the worker
polling 30 seconds after the execution starts:

```
1  ExecutionStarted    14:01:48.601
2  TaskStateEntered    14:01:48.624
3  ActivityScheduled   14:01:48.624
4  ActivityStarted     14:02:21.188   workerName: anchor-probe
5  ActivityTimedOut    14:02:41.188
6  ExecutionFailed     14:02:41.221
```

`ActivityTimedOut` lands 20.000 s after `ActivityStarted` and 52.6 s after `ActivityScheduled`, and
the execution was still `RUNNING` at 30 s with nobody polling, so the queue wait costs the task
nothing. Timing from `ActivityScheduled` instead is close enough here, since it turns a 300-second
`States.HeartbeatTimeout` into a 5-second `States.Timeout`; the `ActivityStarted` placement deserves
its own issue.

## Environment

- Floci `1.7.0`, `main` at `09a05c2d`, JVM (`java -jar target/quarkus-app/quarkus-run.jar`)
- AWS CLI 2.33.7, `us-east-1`
- Every state machine and activity created in AWS for this report was deleted.

### LINKED ISSUE #2734: [BUG] Step Functions (SFN): an execution AWS ends on its own runs forever, and StopExecution only relabels it
## Service

Step Functions (`QueryLanguage: JSONPath`; neither divergence depends on the query language)

## AWS API Action

`StartExecution`, `DescribeExecution`, `GetExecutionHistory`, `StopExecution`

## Expected behavior

AWS puts two hard bounds on a running execution. Neither is `TimeoutSeconds`: none of the
definitions here declares one.

A `Choice` looping forever is terminated unattended at 47.95 seconds with
`FAILED States.Runtime  The execution reached the maximum number of history events (25000).` The cut
is an event count, not a wall clock: event 25,000 *is* the `ExecutionFailed` event.

And `StopExecution` stops the execution. A `Wait: 30` stopped at t+3s with `--error Manual --cause
probe` reads `["ABORTED","Manual","probe"]`, byte-identical two minutes later.

## Actual behavior

```
                                AWS                          Floci
runaway Choice loop             FAILED at 47.95 s, 25000 ev  still RUNNING, no bound
stop-execution on a Wait: 30    ABORTED  Manual  probe       ABORTED  null  null
same, re-read 35 s later        ABORTED  Manual  probe       SUCCEEDED  null  null
```

**No event counter.** `AslExecutor.addEvent` (lines 3441 and 3445) appends to the history list with
no bound and no check. Nothing else ends the execution either, so it holds a worker thread for the
life of the process.

**`StopExecution` writes `ABORTED` and the worker overwrites it.**
`StepFunctionsService.stopExecution` (line 663) sets the status, stores the execution, and puts the
caller's error and cause into the `ExecutionAborted` history event's details, never onto the
`Execution` itself, so `DescribeExecution` returns `null` for both even during the `ABORTED` window.
It does not interrupt or signal the worker thread; `doExecute` keeps stepping and writes its own
terminal status over the top at `AslExecutor.java:427-432` when it finishes, moving `stopDate`
forward 26.4 seconds.

## Reproduction

```bash
ROLE=arn:aws:iam::000000000000:role/r

# Paced with Wait: 1 on purpose, so the JVM heap is never what stops it and the assertion is
# about the event count alone.
SM=$(aws stepfunctions create-state-machine --name paced --role-arn $ROLE \
  --query stateMachineArn --output text \
  --definition '{"StartAt":"Loop","States":{
    "Loop":{"Type":"Choice","Choices":[{"Variable":"$.spin","BooleanEquals":true,"Next":"Tick"}],"Default":"Done"},
    "Tick":{"Type":"Wait","Seconds":1,"Next":"Loop"},"Done":{"Type":"Succeed"}}}')
EX=$(aws stepfunctions start-execution --state-machine-arn "$SM" --input '{"spin":true}' \

sleep 180
aws stepfunctions describe-execution --execution-arn "$EX" --query status --output text
aws stepfunctions get-execution-history --execution-arn "$EX" --query 'length(events)'
# RUNNING
# 1272
#     AWS fails the same shape at 25000 events, in under a minute

SM=$(aws stepfunctions create-state-machine --name stopprobe --role-arn $ROLE \
  --query stateMachineArn --output text \
  --definition '{"StartAt":"W","States":{"W":{"Type":"Wait","Seconds":30,"End":true}}}')

sleep 3
aws stepfunctions stop-execution --execution-arn "$EX" --error Manual --cause probe
aws stepfunctions describe-execution --execution-arn "$EX" --query '[status,error,cause]' --output json
sleep 35
aws stepfunctions describe-execution --execution-arn "$EX" --query '[status,error,cause]' --output json
# ["ABORTED",null,null]
# ["SUCCEEDED",null,null]
#     AWS returns ["ABORTED","Manual","probe"] both times
```

Kill the container to stop the paced loop, `StopExecution` will not.

## Suggested fix

Both halves land in the same place: `AslExecutor` steps states in a loop with no per-step check. The
cap needs the counter checked where `addEvent` increments it (3441/3445); `StopExecution` needs a
flag the loop reads and that beats the worker's terminal write at `AslExecutor.java:427-432`.

A fix must reproduce AWS's exact shape, not merely stop the loop early: the counter is not reset or
offset, event 25,000 *is* the failure event. `stopExecution` returning early when the status is not
`RUNNING` already matches AWS; the gap is only that the status it writes is not authoritative, and
that the caller's `error` and `cause` never reach the `Execution`.

## Environment

- Floci `1.7.0`, `main` at `09a05c2d`, JVM (`java -jar target/quarkus-app/quarkus-run.jar`)
- AWS CLI 2.33.7, `us-east-1`
- Every state machine created in AWS for this report was deleted.

### LINKED ISSUE #2779: [BUG] Step Functions (SFN): execution `TimeoutSeconds` is not enforced while a state blocks on an integration
## What happens

A state machine's top-level `TimeoutSeconds` is checked at each state boundary, in the `Wait` sleep and (as of #2733's fix) in a `Retry`'s backoff. Every other place an execution blocks runs to its own bound instead, so the execution outlives its budget by however long that blocking call takes.

Five such places, four in `AslExecutor.java` and one in `MapIterationScheduler`. `executionDeadlineNanos` is simply not in scope in any of them, which one command shows:

```
grep -n "executionDeadlineNanos" src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
```

| Blocking call | Bounded instead by | Longest it can outlive the budget |
| --- | --- | --- |
| `awaitToken`, a `.waitForTaskToken` Task | the Task's own `TimeoutSeconds`, or `DEFAULT_TASK_TOKEN_TIMEOUT_SECONDS` when it declares none | 300 s |
| `invokeNestedStateMachine`, a `states:startExecution.sync` | its own poll loop, 600 polls at 100 ms | 60 s |
| `invokeEcsRunTask`, an `ecs:runTask.sync` | `ECS_SYNC_POLL_ATTEMPTS` at `ECS_SYNC_POLL_INTERVAL_MS` | the product of the two |
| `executeParallelState` joining its branches | the `Parallel`'s own `TimeoutSeconds`, or unbounded when it declares none | unbounded |
| `executeMapState` waiting for its iterations | nothing | unbounded |

## Reproduction

A five-second budget over an activity task nobody answers, whose own `TimeoutSeconds` is 60:

```json
{
  "StartAt": "Wait for a token",
  "TimeoutSeconds": 5,
  "States": {
    "Wait for a token": {
      "Type": "Task",
      "Resource": "<activity arn>",
      "TimeoutSeconds": 60,
      "End": true
    }
  }
}
```

Measured in `us-east-1` with AWS CLI 2.33.7, nobody ever calling `GetActivityTask`:

```
"status":    "TIMED_OUT"
"startDate": 2026-08-30T13:04:08.506-03:00
"stopDate":  2026-08-30T13:04:13.506-03:00
history:     ExecutionStarted, TaskStateEntered, ActivityScheduled, ExecutionTimedOut
```

Exactly 5.000 s, with no `error` and no `cause` on the execution. Two things the history settles beyond the status: there is no `ActivityTimedOut` and no `ExecutionFailed`, so the timeout path writes nothing about the state it cuts; and there is no `ActivityStarted`, because nobody polled, which is a separate divergence from floci emitting that event at schedule time.

Floci ends the same execution `FAILED` with `States.Timeout` sixty seconds in, because `awaitToken` only ever consults the Task's own clock.

The activity and the state machine created for this measurement were deleted; `list-activities` and `list-state-machines` return neither.

facuparedes:
> **Independent of #2783 and #2786.** The three touch disjoint code and merge into `main` in any order with no conflict. The tree with all three applied runs `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'` at 681 tests, 0 failures, 0 errors, 0 skips.

## Summary

Four things end a running execution on AWS, and none of them ended one the way AWS does here: the state machine's `TimeoutSeconds`, a `Task`'s `TimeoutSeconds` and `HeartbeatSeconds`, the 25,000-event history limit, and `StopExecution`. Three did nothing at all; the fourth, `HeartbeatSeconds`, bounded the whole wait instead of the gap between heartbeats and reported the wrong error name. All four live in `AslExecutor` and its two collaborators, which is why they arrive in one PR rather than in three that would conflict on every hunk. Closes #2733, closes #2734, closes #2779.

### The state machine's budget

- **Computed once**, in `AslExecutor.doExecute`, so every state measures against the same instant. Expiry ends the execution `TIMED_OUT` with `stopDate` set, no `error` and no `cause`, and one `ExecutionTimedOut` event carrying `States.Timeout` with `previousEventId` `0`.
- **Every wait stops at the budget rather than outlasting it.** `sleepOrTimeOutExecution` parks until the earlier of the two instants and ends the execution when the budget is the one that fires; it covers the two pauses a definition asks for, a `Wait` and a `Retry`'s backoff, and the `states:startExecution.sync` and `ecs:runTask.sync` poll loops. `awaitToken` folds the budget into the deadline it already parks on, `executeParallelState` into its branch join, and `MapIterationScheduler` takes it as a parameter and ends its wait on it. Sleeping any of them out in full reported the timeout as many seconds late as the wait outlasted the budget.
- **The state that gets cut is not written about.** `executeTaskState` rethrows `ExecutionTimedOutException` ahead of the catch that publishes `TaskFailed`, because AWS's history for an execution cut while a task waits ends at `ActivityScheduled` and then `ExecutionTimedOut`, with no task-level event between them.
- **A `Parallel` or `Map` branch is still never cut mid-state**: a branch keeps `Long.MAX_VALUE` as its own deadline. What the budget cuts is the parent's wait for it, which is where AWS ends the execution too.

### A `Task`'s two clocks

- **`awaitToken` carries two independent bounds**: `TimeoutSeconds` for the whole wait, `HeartbeatSeconds` for the gap between two `SendTaskHeartbeat` calls, read through `heartbeatDeadlineNanos` so a heartbeat that lands while the task is parked moves the deadline forward instead of being missed. A state with no `TimeoutSeconds` waits `DEFAULT_TASK_TOKEN_TIMEOUT_SECONDS`, 300.
- **`SendTaskHeartbeat` does something now.** `StepFunctionsService.sendTaskHeartbeat` was a `LOG.debugv` and nothing else; it stamps `taskHeartbeatNanos`, `registerPendingToken` stamps it at schedule time, `lastTaskHeartbeatNanos` reads it and `discardPendingToken` forgets it when the waiting task stops listening. Without that last step the token stayed pending for the life of the process and a later `SendTaskSuccess` was answered as if someone were still waiting.
- **Either clock ends the state as `TaskTimedOutException`**, a `FailStateException` reporting `States.Timeout` with a null cause, and `executeTaskState` writes `addTaskTimedOutEvent` (`ActivityTimedOut`, or `TaskTimedOut` for a `.waitForTaskToken` integration) instead of the `*Failed` event.
- **`catchMatches` takes the failure instead of its error string**, and asks `FailStateException.isNamedBy` which names it answers to. A `TaskTimedOutException` answers to `States.Timeout` and to the clock that ran out, so a `Catch` on `States.HeartbeatTimeout` still fires on a heartbeat expiry.
- **A null cause means the failure has none.** `failExecution` leaves the key out of both `DescribeExecution` and the `ExecutionFailed` event rather than reporting it empty. The two other failures that reach it without one keep an empty cause, since AWS's shape for those is unmeasured: `executeFail` for a `Fail` state with no `Cause`, and `StepFunctionsJsonHandler.handleSendTaskFailure` for a `SendTaskFailure` that names none.

### The history-event limit

- **Counting an event and numbering it were one `AtomicLong`; they are two meanings now.** `producedEventCount` counts every event the execution produces, its `Parallel` branches' and inline `Map` iterations' alike, and is what the limit judges. The published history numbers its own events: `appendEvent` gives an event the id of its position in the list, which makes the list the one authority for the next id.
- **`countTowardsHistoryEventLimit` takes the event before it judges it.** One `incrementAndGet`, so however many branches and iterations are producing events at once, exactly one takes event 24,999 and every other finds the limit reached. It leaves the last slot free for `publishTerminalEvent`, which is exempt: an execution always gets to say how it ended.
- **A branch's events are counted and never built.** `executeBranch` passes a null history down and `publishEvent` counts and returns, so it no longer assembles `HistoryEvent` objects carrying full input and output JSON into a list it threw away.
- **A Distributed Map item spends its own budget.** `childExecutionEventCount` is a fresh counter per item in `DISTRIBUTED` mode and the parent's counter in `INLINE`, because AWS runs each item of a Distributed Map as a child execution with a history of its own.

### `StopExecution`

- **`StepFunctionsService.stopExecution` puts the caller's `error` and `cause` on the `Execution`,** not only in the history event's details, so `DescribeExecution` returns them. It writes under the execution's monitor, and `abortedByCaller` is read by the state loop between states and by `failExecution` and `succeedExecution` before they publish, so a stop that arrives mid-state is the status that stands.
- **`ExecutionHistory` seals itself with its terminal event.** The worker is still inside a state when the abort lands, and what it has left to record belongs to an execution the caller has already been told is finished. `stopExecution` reaches that history through `computeIfAbsent`: executions are stored and histories are not, so a `RUNNING` execution can come back from a restart with nothing behind it.

## Wire-fidelity details (AWS CLI 2.33.7 against `us-east-1`)

The measurement behind the last row was run for this PR; the rest are the measurements recorded in #2733 and #2734. Every AWS resource created for any of them was deleted afterwards, and `list-state-machines` and `list-activities` come back with none of them.

```bash
ACT=$(aws stepfunctions create-activity --name floci-exec-deadline-probe \
  --query activityArn --output text)
SM=$(aws stepfunctions create-state-machine --name floci-exec-deadline-probe \
  --role-arn "$ROLE" --query stateMachineArn --output text \
  --definition '{"StartAt":"T","TimeoutSeconds":5,"States":{"T":{"Type":"Task",
    "Resource":"'"$ACT"'","TimeoutSeconds":60,"End":true}}}')
EX=$(aws stepfunctions start-execution --state-machine-arn "$SM" \
sleep 15                                    # nobody ever calls GetActivityTask
aws stepfunctions describe-execution --execution-arn "$EX"
aws stepfunctions get-execution-history --execution-arn "$EX"
```

| Case | AWS returns | Pinned by |
| --- | --- | --- |
| Top-level `TimeoutSeconds` expiry | `TIMED_OUT`, `stopDate` set, `error` and `cause` both absent | `stateMachineTimeoutSecondsCutsAWaitLongerThanTheRemainingBudget` |
| The event that names it | one `ExecutionTimedOut`, `previousEventId` `0`, no `WaitStateExited` for the cut state | `timedOutExecutionEmitsExecutionTimedOutAndNoStateExitedForTheCutState` |
| A `Task` whose `TimeoutSeconds` runs out | `FAILED`, `States.Timeout`, no `cause` key | `taskTimeoutSecondsBoundsTheWholeWaitForTheTaskToken` |
| `HeartbeatSeconds` expiry, `DescribeExecution` | `status` `FAILED`, `error` `States.Timeout`, not `States.HeartbeatTimeout` | `taskHeartbeatSecondsBoundsTheGapBetweenHeartbeats` |
| The history it leaves | `ExecutionStarted`, `TaskStateEntered`, `ActivityScheduled`, `ActivityStarted`, `ActivityTimedOut`, `ExecutionFailed` | `taskHeartbeatSecondsBoundsTheGapBetweenHeartbeats` |
| `activityTimedOutEventDetails` | `{"error": "States.Timeout"}`, no `cause` field | `taskHeartbeatSecondsBoundsTheGapBetweenHeartbeats` |
| A `Catch` on `States.HeartbeatTimeout`, and one on `States.Timeout` | both fire on that event | `aCatchMatchesAHeartbeatExpiryUnderEitherErrorName` |
| A worker that heartbeats as the definition asks | runs on until `TimeoutSeconds` runs out | `heartbeatsKeepATaskAliveUntilItsTimeoutSecondsRunOut` |
| A `Choice` looping with no terminal state | `FAILED`, `States.Runtime`, `The execution reached the maximum number of history events (25000).` | `runawayLoopFailsWithTheHistoryEventLimitError` |
| The event that ends it | the history is 25,000 events, `ExecutionFailed` is id 25,000, the machine's own last event is 24,999 | `executionFailedIsTheTwentyFiveThousandthEvent` |
| Events produced inside a `Parallel` branch † | count towards the same execution's limit | `eventsProducedInsideAParallelBranchCountTowardsTheLimit` |
| A `DISTRIBUTED` `Map` over 13,000 items † | the parent is charged for the run, not for what the items do: `SUCCEEDED` | `aDistributedMapItemDoesNotSpendTheParentsHistoryEvents` |
| A `Retry` and a `Catch` declared for the failure | neither intercepts `States.Runtime`; the execution ends there | `neitherRetryNorCatchInterceptsTheHistoryEventLimit` |
| `StopExecution` on a `Wait: 30` at t+3s, re-read after it elapses | `["ABORTED", "Manual", "probe"]` both times, `stopDate` included | `stoppedExecutionReportsTheCallersErrorAndCause`, `stoppedExecutionStaysAbortedAfterTheWaitElapses` |
| The history of an aborted execution | ends at `ExecutionAborted`, no id used twice | `abortedExecutionEndsItsHistoryAtExecutionAborted` |
| A state machine budget of 5 s over a `Task` whose own `TimeoutSeconds` is 60 | `TIMED_OUT` at `stopDate` minus `startDate` exactly 5.000 s; history `ExecutionStarted`, `TaskStateEntered`, `ActivityScheduled`, `ExecutionTimedOut`, with no `ActivityTimedOut` and no `ExecutionFailed` ‡ | `stateMachineTimeoutSecondsCutsTheWaitForATaskToken` |

† AWS's documented model rather than a shape measured here: branch and iteration events appear in the parent's history on AWS, and a Distributed Map item runs as a child execution with a history of its own.

‡ AWS emitted no `ActivityStarted` either, because no worker ever polled. floci emits it at schedule time, so the test asserts that one extra event; what it pins is that nothing about the cut task is written between `ActivityScheduled` and `ExecutionTimedOut`. The scope note below has the rest.

`States.Timeout` for a heartbeat expiry is the one shape worth re-reading: the error name AWS reports and the error name a `Catch` matches are not the same set, which is why matching moved onto the failure object rather than onto its `error` string.

Six defects were found in review. Each was turned red before it was fixed, and the line below is what the test that now pins it printed:

- **Branch events consumed published event ids**, so `GetExecutionHistory` returned non-contiguous ids and a dangling `previousEventId` for every `Parallel` and every `Map`: `unexpected id at index 1 ==> expected: <2> but was: <6>`. `appendEvent` numbering from the published history is the fix.
- **A `DISTRIBUTED` Map's iterations were charged to the parent's budget:** `expected: <SUCCEEDED> but was: <FAILED>` with the history-limit cause, on a 13,000-item Map. `childExecutionEventCount` is the fix.
- **The limit was a non-atomic check-then-increment** on a counter that had just become shared: `more than one racer took the last countable event ==> expected: <1> but was: <2>`, with 40 threads on a barrier. Taking the event before judging it is the fix.
- **`stopExecution` dereferenced a missing history-cache entry:** `Expected status code <200> but was <500>` with a suppressed NPE, reachable under `floci.storage.mode: persistent` or `hybrid` after a restart and during a reset. `computeIfAbsent` is the fix.
- **A serial Map ignored the execution budget.** A one-item Map, or `MaxConcurrency: 1`, ran each iteration inline with the deadline checked only between items, so a machine ending in such a Map reported `SUCCEEDED` past its budget. The serial special case is deleted and every Map goes through the completion-service path the deadline already cuts, with one upfront expiry check kept so an already-expired deadline starts nothing. Pinned by `endsTheSerialWaitForASingleItemOncePastTheDeadline`, `endsTheSerialWaitForMaxConcurrencyOneOncePastTheDeadline`, and end to end by `stateMachineTimeoutSecondsCutsTheWaitForASingleMapIteration`.
- **A token registered before its task started leaked on the early-failure paths.** A throw between `registerPendingToken` and `awaitToken` (the history limit inside the schedule events, or `invokeResource` failing) left the pending token and its heartbeat stamp for the life of the process, and a later `SendTaskSuccess` on it was answered as if someone were waiting: `sendTaskSuccessOnATokenWhoseResourceInvocationThrewFailsAsInvalid` went red on exactly that. Any failure before the wait now discards the registration; `discardPendingToken` is idempotent, so the wait's own `finally` stays.

## Deliberate scope notes

- **Both task clocks start at schedule time, where AWS starts `TimeoutSeconds` when a worker picks the task up.** floci emits `ActivityStarted` one millisecond after `ActivityScheduled`, with no `workerName`, before any worker has called `GetActivityTask`; the AWS run above emitted no `ActivityStarted` at all, since nobody polled. Anchoring `TimeoutSeconds` correctly means moving that event, which changes what every activity execution's history looks like and belongs in its own change. Recorded as a deviation in `docs/services/step-functions.md`.
- **A `Task` with no `TimeoutSeconds` waits 300 seconds, where AWS waits a year.** That ceiling is the one the code already had, kept so an emulated run frees its worker thread rather than parking it for the life of the process.
- **The 25,000-event failure is not interceptable, and that is not new here.** `catchMatches` refuses `States.Runtime` before it reads `ErrorEquals`, and both `findMatchingRetrier` and `handleCatch` go through it; that guard predates this change. `neitherRetryNorCatchInterceptsTheHistoryEventLimit` pins it: a `Parallel` declaring `Retry` at `IntervalSeconds: 30, MaxAttempts: 3` and a `Catch` on `States.ALL` still fails with `States.Runtime` and finishes in 0.06 s where three backoffs would cost over 90 s.
- **floci publishes no history events for `Parallel` branches or `Map` iterations at all.** AWS does; this predates the change and is not fixed here. It is why the runaway-`Parallel` machine's `ExecutionFailed` lands at a low id. The top-level shape is unaffected and still matches AWS exactly.
- **The count can overshoot 25,000 under concurrency**, by up to one increment per racing thread, once the limit is already reached. Every one of those threads is raising `States.Runtime` and the execution is over; nothing reads the count afterwards. The alternative is a compare-and-set loop that makes every event on the common path pay for a contended retry to keep a number no caller ever sees exact.
- **`ExecutionHistory` enforces its seal in `add`, the one way events are appended.** `addAll` does not route through it and would write past the seal; nothing in `src/` does a bulk append.
- **`TimeoutSeconds` on `Parallel` is untouched.** AWS rejects the field there (#2735), so `executeParallelState` keeps the shape it has and the budget is folded into it rather than replacing it.
- **The empty cause stays wherever it is not measured.** A `Fail` state with no `Cause` and a `SendTaskFailure` with no cause both still report an empty one; only the two timeouts omit the key.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Breaking change note

A `Catch` or `Retry` on `States.TaskFailed` no longer matches a heartbeat expiry. The reported error moved from `States.HeartbeatTimeout` to `States.Timeout`, and `catchMatches` already excluded `States.Timeout` from the `States.TaskFailed` wildcard, as the ASL spec requires. A definition that recovered a heartbeat expiry through `States.TaskFailed` now fails the execution; it recovers again by naming `States.Timeout` or `States.HeartbeatTimeout`.

## AWS Compatibility

- [x] The incorrect behaviour: top-level `TimeoutSeconds` was never enforced, so an execution ran to completion however long it took; on a `Task`, `HeartbeatSeconds` bounded the whole wait and `TimeoutSeconds` was decorative, and the expiry reported `States.HeartbeatTimeout` with a cause where AWS reports `States.Timeout` with none; a runaway execution stayed `RUNNING` with no bound, holding a worker thread for the life of the process; and `StopExecution` wrote `ABORTED` with a null `error` and `cause` that the worker then overwrote with its own terminal status
- [x] Verified against real Step Functions in `us-east-1` with AWS CLI 2.33.7. Every value in the table and in the tests is a captured AWS response, except the two rows marked †
- [x] Every AWS resource created for the measurements was deleted; `list-state-machines` and `list-activities` return none of them

## Checklist

- [x] `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'` green on the current tip: 612 tests, 0 failures, 0 errors, 0 skips
- [x] 30 tests added over `origin/main`'s 582: `StepFunctionsTimeoutSecondsIntegrationTest` (12), `AslExecutorHistoryEventLimitTest` (8), `StepFunctionsStopExecutionIntegrationTest` (4), `MapIterationSchedulerTest` (4) for the deadline the scheduler now takes, 1 in `AslExecutorTaskHistoryEventsTest` for the unbroken event chain under `Parallel` and `Map`, and 1 in `StepFunctionsAwsSdkTaskIntegrationTest` for the token discarded when its task fails before the wait
- [x] Revert receipt for the `TimeoutSeconds` half: with `AslExecutor.java`, `StepFunctionsService.java` and `StepFunctionsJsonHandler.java` reverted to `origin/main` in a scratch copy and the tests untouched, all 8 of the tests the first commit adds to `StepFunctionsTimeoutSecondsIntegrationTest` fail (the class's other 4 arrive with the third commit); two read `SUCCEEDED` where `TIMED_OUT` is expected, two read `States.HeartbeatTimeout` where `States.Timeout` is expected, and the history has no `ExecutionTimedOut` in it
- [x] Revert receipt for the `StopExecution` half: with `StepFunctionsService.java` alone restored and the tests untouched, all 4 fail, three on `expected: <Manual> but was: <null>` and the fourth on the history continuing past its terminal event
- [x] The retry-backoff test was measured on its own with `sleepBeforeRetry`'s deadline removed: `expected: <TIMED_OUT> but was: <FAILED>`, on a response whose `stopDate` minus `startDate` is 10.03 s against a 1-second budget
- [ ] Full `./mvnw test` not run: the run above is the `stepfunctions` subtree, which is every package this diff touches

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -111,6 +111,35 @@ uniformly between zero and the computed delay, as on AWS. One deviation. The del
 between attempts is capped at 30 seconds, the same cap Floci applies to `Wait` states,
 so emulated runs stay fast.
 
+## Timeouts
+
+ASL carries two `TimeoutSeconds` fields and Floci enforces both, in the two terminal shapes
+AWS uses.
+
+The state machine's own `TimeoutSeconds` is the whole execution's budget. It is checked before
+every state and inside a `Wait`, so a `Wait` longer than what is left is cut rather than slept
+out. The execution ends `TIMED_OUT` with `stopDate` set and no `error` and no `cause` at all;
+`States.Timeout` is named only in the single `ExecutionTimedOut` event, whose `previousEventId`
+is `0`. The state that was cut gets no `*StateExited` event. A `Parallel` or `Map` branch runs
+on its own thread and is not cut mid-state: the budget is enforced again as soon as the branch
+returns.
+
+A `Task` that waits for a task token — an activity, or a `.waitForTaskToken` integration — has
+two independent bounds. `TimeoutSeconds` is the whole wait; `HeartbeatSeconds` is the longest gap
+allowed between two `SendTaskHeartbeat` calls, and every heartbeat pushes that gap forward, so a
+worker that reports as often as the definition asks runs until `TimeoutSeconds` runs out. Either
+clock ends the state the same way: an `ActivityTimedOut` event — `TaskTimedOut` for a
+`.waitForTaskToken` integration — carrying `States.Timeout` and no cause, then an execution that
+reads `FAILED` with that same error. A `Catch` on a heartbeat expiry matches under
+`States.HeartbeatTimeout` and under `States.Timeout` alike. Neither timeout carries a cause:
+`DescribeExecution` and the `ExecutionFailed` event both leave the key out, where every other
+failure reports one. A `Task` that declares no `TimeoutSeconds` waits 300 seconds, where AWS
+waits a year.
+
+One deviation. AWS starts the `TimeoutSeconds` clock when a worker picks the task up, the instant
+it emits `ActivityStarted`. Floci emits `ActivityStarted` at schedule time, so both clocks start
+when the task is scheduled.
+
 ## JSONata nulls
 
 An expression that evaluates to JSON `null` produces a value, not a missing one. It keeps its key
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -114,7 +114,9 @@ so emulated runs stay fast.
 ## Timeouts
 
 ASL carries two `TimeoutSeconds` fields and Floci enforces both, in the two terminal shapes
-AWS uses.
+AWS uses. The state machine's own field bounds every state; a `Task`'s own field only bounds
+one that waits for a task token — an activity, or a `.waitForTaskToken` integration. A Lambda
+or other SDK task that returns directly is not bound by it.
 
 The state machine's own `TimeoutSeconds` is the whole execution's budget. It is checked before
 every state and inside a `Wait`, so a `Wait` longer than what is left is cut rather than slept
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -108,6 +108,9 @@ public class AslExecutor {
 
     private static final Logger LOG = Logger.getLogger(AslExecutor.class);
     private static final int MAX_WAIT_SECONDS = 30;
+    // How long a Task waits for its token when the state declares no TimeoutSeconds. AWS lets it
+    // run for a year; the emulator would rather free the worker thread.
+    private static final int DEFAULT_TASK_TOKEN_TIMEOUT_SECONDS = 300;
     private static final int INLINE_MAP_MAX_CONCURRENCY = 40;
     private static final int DISTRIBUTED_MAP_MAX_CONCURRENCY = 10_000;
 
@@ -364,6 +367,12 @@ public class AslExecutor {
             String startAt = definition.path("StartAt").asText();
             String topLevelQueryLanguage = definition.path("QueryLanguage").asText("JSONPath");
             JsonNode currentInput = parseInput(exec.getInput());
+            // The state machine's total budget, computed once so every state measures against the
+            // same instant. Long.MAX_VALUE stands for a definition with no TimeoutSeconds.
+            int timeoutSeconds = definition.path("TimeoutSeconds").asInt(0);
+            long executionDeadlineNanos = timeoutSeconds > 0
+                    ? System.nanoTime() + TimeUnit.SECONDS.toNanos(timeoutSeconds)
+                    : Long.MAX_VALUE;
             JsonNode execContext = buildContext(exec, sm);
             // Execution-scoped JSONata variables (the Assign field). Mutated in place as states
             // assign, so later states observe earlier assignments.
@@ -371,6 +380,9 @@ public class AslExecutor {
 
             String currentStateName = startAt;
             while (currentStateName != null) {
+                if (System.nanoTime() >= executionDeadlineNanos) {
+                    throw new ExecutionTimedOutException();
+                }
                 JsonNode stateDef = states.path(currentStateName);
                 if (stateDef.isMissingNode()) {
                     throw new RuntimeException("State not found: " + currentStateName);
@@ -388,7 +400,8 @@ public class AslExecutor {
                 var jsonata = isJsonata(stateDef, topLevelQueryLanguage);
                 try {
                     var stateResult = executeStateWithRetry(currentStateName, type, stateDef, currentInput,
-                            history, eventId, sm, jsonata, topLevelQueryLanguage, execContext, variables);
+                            history, eventId, sm, jsonata, topLevelQueryLanguage, execContext, variables,
+                            executionDeadlineNanos);
                     addEvent(history, eventId, stateExitedEventType(type),
                             Map.of("name", currentStateName, "output", stateResult.output().toString(),
                                    "outputDetails", Map.of("truncated", false)));
@@ -435,6 +448,15 @@ public class AslExecutor {
                     Map.of("output", currentInput.toString(), "outputDetails", Map.of("truncated", false)));
             onUpdate.accept(exec, history);
 
+        } catch (ExecutionTimedOutException e) {
+            // A timed out execution carries neither error nor cause: DescribeExecution leaves both
+            // keys out, and States.Timeout is named only inside the ExecutionTimedOut event, which
+            // points at the start of the execution rather than at the state it cut.
+            exec.setStopDate(System.currentTimeMillis() / 1000.0);
+            exec.setStatus("TIMED_OUT");
+            addEvent(history, eventId, "ExecutionTimedOut", 0L, Map.of("error", "States.Timeout"));
+            onUpdate.accept(exec, history);
+
         } catch (Exception e) {
             LOG.warnv("ASL execution failed for {0}: {1}", exec.getExecutionArn(), e.getMessage());
             // This path previously set only the status, leaving error and cause null forever on an
@@ -472,14 +494,14 @@ public class AslExecutor {
     private StateResult executeStateWithRetry(String name, String type, JsonNode stateDef, JsonNode input,
                                               List<HistoryEvent> history, AtomicLong eventId, StateMachine sm,
                                               boolean jsonata, String topLevelQueryLanguage, JsonNode context,
-                                              ObjectNode variables) throws Exception {
+                                              ObjectNode variables, long executionDeadlineNanos) throws Exception {
         var retriers = stateDef.path("Retry");
         var attemptsPerRetrier = new HashMap<Integer, Integer>();
         var attempt = 0;
         while (true) {
             try {
                 return executeState(name, type, stateDef, input, history, eventId, sm, jsonata,
-                        topLevelQueryLanguage, context, variables, attempt);
+                        topLevelQueryLanguage, context, variables, attempt, executionDeadlineNanos);
             } catch (FailStateException e) {
                 var retrierIndex = findMatchingRetrier(retriers, e);
                 if (retrierIndex < 0) {
@@ -490,7 +512,7 @@ public class AslExecutor {
                 if (attemptsUsed > retrier.path("MaxAttempts").asInt(3)) {
                     throw e;
                 }
-                sleepBeforeRetry(retrier, attemptsUsed);
+                sleepBeforeRetry(retrier, attemptsUsed, executionDeadlineNanos);
                 attempt++;
                 updateRetryCount(context, attempt);
             }
@@ -501,20 +523,25 @@ public class AslExecutor {
         if (!retriers.isArray()) {
             return -1;
         }
-        var error = failure.error != null ? failure.error : "States.Runtime";
         for (var i = 0; i < retriers.size(); i++) {
-            if (catchMatches(retriers.get(i), error)) {
+            if (catchMatches(retriers.get(i), failure)) {
                 return i;
             }
         }
         return -1;
     }
 
-    private void sleepBeforeRetry(JsonNode retrier, int attemptsUsed) throws InterruptedException {
+    /**
+     * Backs off before the next attempt. The backoff is a pause inside the state, so a retrier
+     * whose interval outlasts the state machine's {@code TimeoutSeconds} budget ends the execution
+     * where the budget runs out rather than attempting again past it, which is what AWS does. The
+     * deadline is read even when the delay is zero, so a state that already spent the budget stops
+     * instead of retrying instantly.
+     */
+    private void sleepBeforeRetry(JsonNode retrier, int attemptsUsed, long executionDeadlineNanos)
+            throws InterruptedException {
         var delaySeconds = retryDelaySeconds(retrier, attemptsUsed, ThreadLocalRandom.current().nextDouble());
-        if (delaySeconds > 0) {
-            Thread.sleep((long) (delaySeconds * 1000));
-        }
+        sleepOrTimeOutExecution((long) (delaySeconds * 1_000_000_000L), executionDeadlineNanos);
     }
 
     /**
@@ -544,13 +571,14 @@ public class AslExecutor {
     private StateResult executeState(String name, String type, JsonNode stateDef, JsonNode input,
                                      List<HistoryEvent> history, AtomicLong eventId, StateMachine sm,
                                      boolean jsonata, String topLevelQueryLanguage, JsonNode context,
-                                     ObjectNode variables, int attempt) throws Exception {
+                                     ObjectNode variables, int attempt,
+                                     long executionDeadlineNanos) throws Exception {
         return switch (type) {
             case "Pass" -> executePassState(stateDef, input, jsonata, context, variables);
             case "Task" -> executeTaskState(name, stateDef, input, history, eventId, sm, jsonata, context,
                     variables, attempt);
             case "Choice" -> executeChoiceState(stateDef, input, jsonata, context, variables);
-            case "Wait" -> executeWaitState(stateDef, input, jsonata, context, variables);
+            case "Wait" -> executeWaitState(stateDef, input, jsonata, context, variables, executionDeadlineNanos);
             case "Succeed" -> executeSucceedState(stateDef, input, jsonata, context, variables);
             case "Fail" -> executeFail(stateDef, input, jsonata, context, variables);
             case "Parallel" -> executeParallelState(name, stateDef, input, sm, jsonata, topLevelQueryLanguage, context, variables);
@@ -632,8 +660,11 @@ public class AslExecutor {
                     ? mockedTaskResult(mockedSteps, stateName, attempt)
                     : invokeResource(effectiveResource, effectiveInput, sm, taskToken);
             if (tokenFuture != null) {
-                taskResult = awaitToken(tokenFuture, stateDef);
+                taskResult = awaitToken(tokenFuture, stateDef, taskToken);
             }
+        } catch (TaskTimedOutException e) {
+            addTaskTimedOutEvent(history, eventId, profile);
+            throw e;
         } catch (Exception e) {
             var failure = e instanceof FailStateException f ? f : null;
             addTaskFailedEvent(history, eventId, profile,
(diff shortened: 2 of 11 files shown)


pgermosen (approved):
Approving. Didn't just take the write-up's word for it - traced the deadline threading across every wait path, and it really is computed once and checked consistently through Wait, Retry backoff, .sync polling, awaitToken, and the Parallel/Map joins. The breaking-change reasoning holds up in code too: catchMatches genuinely excludes States.Timeout from the States.TaskFailed wildcard, and moving to a polymorphic isNamedBy() is the right way to let a heartbeat expiry answer to two names without conflating them. The atomic counting fix (increment-before-judge, numbering from the actual list position rather than the same counter) and the synchronized appendEvent for the StopExecution race are both exactly what they're described as. Excellent work all around, including catching and fixing six of your own defects before this even got to review.

PR #2783, opened 2026-08-30
fix(stepfunctions): match AWS on which JSONata expressions fail and on what they say

### LINKED ISSUE #2736: [BUG] Step Functions (SFN): a failed JSONata expression's cause drops the sentence naming the expression and the field
## Service

## AWS API Action

`TestState`, `StartExecution`, `DescribeExecution`

## Expected behavior

AWS opens the cause with a sentence naming the expression and the field, then the JSONata code and
message:

```
The JSONata expression '$abs("x")' specified for the field 'Output/v' threw an error during
evaluation. T0410: Argument 1 of function "abs" does not match function signature
```

A real execution prepends `An error occurred while executing the state '<name>' (entered at the
event id #<n>). ` to that. That outer wrapper is a separate known gap
(`StepFunctionsJsonataIntegrationTest.java:1268`). The inner sentence is the subject here, and it is
missing from both APIs.

## Actual behavior

Floci emits the tail alone. Same expression, same field:

```
AWS    The JSONata expression '$abs("x")' specified for the field 'Output/v' threw an error
       during evaluation. T0410: Argument 1 of function "abs" does not match function signature
Floci  T0410: Argument 1 of function "abs" does not match function signature
```

The error code and the catchability are correct: this is the cause text only. Because the message is
per-function and not per-field, a state with eight `Arguments` keys gives no way to tell which one
broke.

The plumbing is already there. `resolveTemplate` (`JsonataEvaluator.java:291`) threads `field` into
`evaluateField` (line 266), and line 271 uses it for the `returned nothing (undefined).` message,
which is byte-identical to AWS. `evaluate` throws its `FailStateException` at line 193 instead,
where `field` is out of scope. Lines 200-201 blame `resolveTemplate` for not threading it, which
stopped being true when #2665 landed. A grep of `AslExecutor` finds 11 call sites into the
evaluator, all `evaluateField(...)` or `resolveTemplate(...)`, and zero direct `evaluate(...)`.

Measured across a 149-shape corpus: of the 101 shapes that fail on both sides with a real AWS
message, all 101 omit the sentence, and 61 of them differ by nothing else.

## Reproduction

```bash

SM=$(aws stepfunctions create-state-machine --name cause \
  --role-arn arn:aws:iam::000000000000:role/r --query stateMachineArn --output text \
    "E":{"Type":"Pass","Output":{"v":"{% $abs(\"x\") %}"},"End":true}}}')

sleep 3
aws stepfunctions describe-execution --execution-arn "$EX" --query 'cause' --output text
# T0410: Argument 1 of function "abs" does not match function signature
#     AWS returns  The JSONata expression '$abs("x")' specified for the field 'Output/v' threw an
#                  error during evaluation. T0410: Argument 1 of function "abs" does not match
#                  function signature
```

The AWS side is one `test-state` call on the same state, which creates nothing.

## Suggested fix

`evaluate` should stop composing the cause. `evaluateField` already holds `field` and the raw
expression, so wrapping the throw there covers both branches at one point:

```java
throw new FailStateException("States.QueryEvaluationError",
    "The JSONata expression '" + expr + "' specified for the field '" + field
        + "' threw an error during evaluation" + separator + tail);
```

`separator` is period-and-code for a coded error and a colon for a bare bound message with no code: AWS ends the sentence with
`. <CODE>: <message>` for 95 of the 101 shapes and with `: <message>` for the other 6. This moves
all 101 to the AWS sentence, 61 of them byte-identical, and the stale comment at lines 200-201 goes
with it.

The remaining differences in the other 40 are independent of this one: `T0410`'s argument index and
`T0412`'s value need what `JException` does not carry.

## Environment

- Floci `1.7.0`, `main` at `09a05c2d`, JVM (`java -jar target/quarkus-app/quarkus-run.jar`);
  `floci/floci:nightly` reproduces the same behaviour
- AWS CLI 2.33.7, `us-east-1`
- Every state machine created in AWS for this report was deleted.

### LINKED ISSUE #2737: [BUG] Step Functions (SFN): JSONata expressions where Floci and AWS disagree on whether the state fails at all
## Service

Step Functions (`QueryLanguage: JSONata`). Disagreements about how an error is *worded* are #2736;
this issue is only about states that fail on one side and succeed on the other.

## AWS API Action

`TestState`, `CreateStateMachine`, `StartExecution`, `DescribeExecution`

## Expected behavior

Three groups, measured against real AWS in `us-east-1` with `TestState`. Every AWS refusal below is
`States.QueryEvaluationError`, and a `Catch` fires on it.

## Actual behavior

**Group 1: AWS fails the state, Floci returns a value.** No `Catch` can fire on a state that
succeeded, so nothing in the execution history marks the divergence and the workflow carries on with
a value AWS would never have produced.

```
                         AWS                        Floci
$eval("1+1")             FAILED  T1006             SUCCEEDED  {"r":2}
$formatNumber(1, "x")    FAILED  D3086             SUCCEEDED  {"r":"x1"}
[1..900000] ~> $count()  FAILED  memory limit      SUCCEEDED  {"r":900000}
$sum([1..900000])        FAILED  memory limit      SUCCEEDED  {"r":405000450000}
1+1+ … +1  (200 terms)   FAILED  stack overflow    SUCCEEDED  {"r":200}
```

**Group 2: Floci fails the state, AWS returns a value.**

```
                          AWS                            Floci
1/0                       SUCCEEDED  {"r":"Infinity"}   FAILED  D1001: Number out of range: "Infinity"
1e308*10                  SUCCEEDED  {"r":"Infinity"}   FAILED  D1001: Number out of range: "Infinity"
$parseInteger("abc","0")  SUCCEEDED  {"r":"NaN"}        FAILED  ... returned nothing (undefined).
```

`Infinity` and `NaN` come back from AWS as JSON strings, since neither is a valid JSON number. The
first two raise JSONata's own `D1001`, which AWS does not raise. The third returns undefined from
Floci's `$parseInteger`, which then fails the field under the correct rule that an undefined field
fails the state; AWS's `$parseInteger` returns `"NaN"`, so that rule never applies there.

**Group 3: Floci's evaluation-depth ceiling is five times AWS's, and the band between is silent.**
Floci's bound is `MAX_EVALUATION_DEPTH = 500` (`JsonataEvaluator.java:101`). Bisected with one call
per `n`, on both sides:

| Expression shape | AWS last OK | Floci last OK | Silent band |
| --- | --- | --- | --- |
| `1+1+...+1`, `n` terms | `n=100` | `n=500` | `n=101…500` |
| `(((…1…)))`, `n` parens | `n=99` | `n=499` | `n=100…499` |
| `[[[…1…]]]`, `n` brackets | `n=99` | `n=499` | `n=100…499` |
| `1~>$string~>…`, `n` stages | `n=99` | `n=499` | `n=100…499` |
| `$f(n)`, non-tail-recursive | `n=31` | `n=165` | `n=32…165` |

Every shape's first Floci refusal reads `Depth=501 max=500`, which fixes a depth-per-`n` for each
shape. Assuming AWS counts nesting the way Floci does, all five AWS refusals land between depth 99
and 101.

## Reproduction

Groups 1 and 2 against the container. `$eval`'s argument is written `\"1+1\"` so the quotes survive
into the JSON definition; interpolated bare they make the definition malformed, which is not the
behaviour under test. On the AWS side `test-state` covers all three groups and creates nothing.

```bash

run() {
  SM=$(aws stepfunctions create-state-machine --name "$1" \
    --role-arn arn:aws:iam::000000000000:role/r --query stateMachineArn --output text \
      "E":{"Type":"Pass","Output":{"r":"{% '"$2"' %}"},"End":true}}}')
  sleep 3
  printf '%-26s %s\n' "$2" "$(aws stepfunctions describe-execution --execution-arn "$EX" \
    --query '[status,output,cause]' --output text)"
  aws stepfunctions delete-state-machine --state-machine-arn "$SM" >/dev/null
}

run g1a '$eval(\"1+1\")'
run g1c '[1..900000] ~> $count()'
run g2a '1/0'
# $eval(\"1+1\")             SUCCEEDED  {"r":2}          None
# [1..900000] ~> $count()    SUCCEEDED  {"r":900000}     None
# 1/0                        FAILED     None             D1001: Number out of range: "Infinity"
#     AWS refuses the first two and answers {"r":"Infinity"} for the third
```

## Suggested fix

Five independent changes:

- Lower `MAX_EVALUATION_DEPTH` (`JsonataEvaluator.java:101`) from 500 to AWS's own ceiling, which
  group 3 brackets between depth 99 and 101. The loose bracket in the comment at
  `JsonataEvaluator.java:93-100` and the depth reasoning at `docs/services/step-functions.md:177-181`
  both need updating with it.
- Bound the memory an expression may hold, which Floci does not bound at all today. The bound is
  memory held, not an element count: AWS accepts `[1..873782]` and refuses `[1..873783]`, and
  accepts a string-doubling expression at `n=22` and refuses it at `n=23`.
- Remove `$eval`. It comes from the bundled `jsonata` dependency (`pom.xml:210`) and nothing removes
  it, and both `docs/services/step-functions.md:156` and the comment at `JsonataEvaluator.java:427`
  already say AWS disables it.
- Refuse a `$formatNumber` picture that describes no number. AWS raises `D3086` for
  `$formatNumber(1, "x")` where Floci formats it as `"x1"`. The picture-string rules are XPath
  F&O 4.7.3, whose codes run `D3080` to `D3093`.
- Return what AWS returns for `1/0`, `1e308*10` and `$parseInteger("abc","0")`: the JSON strings
  `"Infinity"`, `"Infinity"` and `"NaN"`, instead of raising `D1001` and instead of returning
  undefined. `docs/services/step-functions.md:183-186` documents the current behaviour.

An `OutOfMemoryError` escaping the state's own `Retry` and `Catch` is a separate defect in
`AslExecutor`, reported as #2738.

## Environment

- Floci `main` at `09a05c2d`, the image's own command
  (`java --enable-native-access=ALL-UNNAMED -jar quarkus-app/quarkus-run.jar`, `docker/Dockerfile:94`)
- AWS CLI 2.33.7, `us-east-1`
- Every AWS measurement here goes through `test-state`, which creates nothing.

### LINKED ISSUE #2738: [BUG] Step Functions (SFN): a java.lang.OutOfMemoryError inside a state bypasses that state's Retry and Catch
## Service

Step Functions. The expression is JSONata, the bug is not: it is the execution-level
`catch (Error e)` in `AslExecutor`.

## AWS API Action

`CreateStateMachine`, `StartExecution`, `DescribeExecution`

## Expected behavior

An expression that exceeds the evaluation memory bound fails the state with
`States.QueryEvaluationError`, and a `Catch` on the state catches it. Adding that bound is #2737;
this issue is the `Error` handling around it, which already misroutes any `OutOfMemoryError` the
evaluator raises today. On real AWS the definition
below ends `SUCCEEDED` with output `{"caught":true}`.

## Actual behavior

```
AWS    SUCCEEDED  {"caught":true}
Floci  FAILED     States.Runtime  java.lang.OutOfMemoryError: Overflow: String length out of range
```

`JsonataEvaluator.evaluate` catches `Exception`, not `Error` (`JsonataEvaluator.java:192`), so the
`OutOfMemoryError` never becomes a `FailStateException`. It escapes to `AslExecutor.java:441`'s
`} catch (Error e) {`, which fails the whole execution as `States.Runtime` (line 448). `Retry` and
`Catch` sit inside the loop that catch wraps, so they never run.

The wiring itself is fine: the same `Parallel` with a 600-term `+` chain fails on depth, is caught,
and ends `SUCCEEDED {"caught":true}` on both.

## Reproduction

```bash

SM=$(aws stepfunctions create-state-machine --name oom-caught \
  --role-arn arn:aws:iam::000000000000:role/r --query stateMachineArn --output text \
  --definition '{"QueryLanguage":"JSONata","StartAt":"P","States":{
    "P":{"Type":"Parallel","Branches":[{"StartAt":"E","States":{
          "E":{"Type":"Pass","Output":{"r":"{% ($p := function($s, $n) { $n = 0 ? $s : $p($s & $s, $n - 1) }; $length($p(\"x\", 31))) %}"},"End":true}}}],
         "Catch":[{"ErrorEquals":["States.ALL"],"Next":"Caught"}],"End":true},
    "Caught":{"Type":"Pass","Output":{"caught":true},"End":true}}}')

sleep 4
aws stepfunctions describe-execution --execution-arn "$EX" \
  --query '[status,output,error,cause]' --output text
# FAILED  None  States.Runtime  java.lang.OutOfMemoryError: Overflow: String length out of range
#     AWS returns  SUCCEEDED  {"caught":true}
```

On a heap smaller than 1 GiB the cause reads `java.lang.OutOfMemoryError: Java heap space` instead;
same path. `TestState` cannot be used here: it supports neither `Map` nor `Parallel` and evaluates
no `Catch`.

## Suggested fix

Keep `AslExecutor.java:441`, removing it leaves the execution `RUNNING` for the life of the process.
Translate the `Error` into a `FailStateException` inside the state instead, where
`JsonataEvaluator.evaluate` already catches `Exception` (line 192), with
`States.QueryEvaluationError` when the `Error` came out of expression evaluation.
`StackOverflowError` and anything else the evaluator can raise need the same translation.

## Environment

- Floci `main` at `09a05c2d`, JVM (`java -jar target/quarkus-app/quarkus-run.jar`)
- AWS CLI 2.33.7, `us-east-1`
- Every state machine created in AWS for this report was deleted.

facuparedes:
> **Independent of #2782 and #2786.** The three touch disjoint code and merge into `main` in any order with no conflict. The tree with all three applied runs `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'` at 681 tests, 0 failures, 0 errors, 0 skips.

## Summary

Three ways `JsonataEvaluator` disagreed with AWS, all of them in the same method and its immediate neighbours: which expressions fail at all, which failures a `Catch` can see, and what the cause says. Closes #2736, closes #2737, closes #2738.

**This changes what an existing state machine does, in both directions: four classes of expression that used to evaluate here now fail the state, and two that used to fail now evaluate.** Newly failing: an expression nesting between depth 101 and 500, one holding more than 6,990,256 bytes, any call to `$eval`, and any `$formatNumber` picture that breaks one of the fourteen rules. Newly evaluating: a non-finite number the expression's own result is (`1/0`, `-1/0`, `1e308 * 10`) and `$parseInteger` on a value its picture cannot read. Every one of the six is a case where the old behaviour was the one that disagreed with AWS.

### Which expressions fail

- **The depth ceiling drops from 500 to 100.** `MAX_EVALUATION_DEPTH` is the value already handed to the library's `frame.setRuntimeBounds`, so nothing new runs; five nesting shapes bisected one `n` at a time put AWS's last accepted evaluation at depth 100 and its first refusal at 101.
- **The memory an expression may hold is bounded, which it was not.** `MAX_EXPRESSION_BYTES` is `873_782 * 8` bytes. `boundHeldMemory` installs the check on the same exit callback the library's time and depth bounds already use, through `afterEachEvaluationStep`, which keeps the installed callback and runs after it rather than replacing it. `heldBytes` walks the value the step produced and stops as soon as the walk passes the bound, so one check costs at most 873,783 steps, and a `Utils.RangeList` is counted from its length rather than materialised.
- **`$eval` is bound to null**, through `frame.bind("eval", (Object) null)`. A name that resolves to null and is called as a function is the library's own `T1006`, which is the code AWS answers; `$parse` stays the documented replacement.
- **A non-finite number comes back as the string AWS writes for it.** `fromJsonataValue` routes a non-finite `Double` through `nonFiniteName`, which writes `"Infinity"`, `"-Infinity"` or `"NaN"`. Arithmetic never leaves the value anywhere else, so `nonFiniteRefusal` recovers it from the library's `D1001`, gated by `ExpressionNesting.isTheWholeExpression()` so only an expression whose own result was refused is answered. `parseInteger` answers `Double.NaN` where the library answers nothing and the ASL field then fails the state.
- **`FormatNumberPicture` (new, 238 lines) checks the picture string before formatting.** `validate` splits the picture into sub-pictures and `errorIn` reads the fourteen rules of XPath F&O 4.7.3 in the order their codes number them, keeping the **last** rule broken rather than the first: a picture with no digit breaks D3085 and D3086 at once and AWS reports the second. Wired in at `formatNumber`, ahead of `Functions.formatNumber`.

### Which failures a `Catch` can see

- **A `java.lang.Error` out of the JSONata engine is a state failure now.** `evaluate` caught `Exception` and not `Error`, so an `Error` raised while parsing or evaluating walked past it to `AslExecutor`'s last-resort `catch (Error e)` and failed the whole execution as `States.Runtime`, which `Retry` and `Catch` never see. A second arm, `catch (StackOverflowError | OutOfMemoryError e)`, throws `States.QueryEvaluationError` instead, and its cause goes through the same sentence enrichment as every other evaluation failure, joined with `: ` since it carries no JSONata code.
- **Those two `Error` types and no others**, because they are the resource an expression itself spends: nesting deep enough to exhaust the parser's stack, and asking for a value larger than the JVM can hold. Every other `Error` says the runtime is broken rather than this expression, and it still walks past untouched to `AslExecutor.java:441`, which stays: removing that guard leaves the execution `RUNNING` for the life of the process.
- **The cause is `e.toString()`, not `e.getMessage()`.** An `Error` often carries a null message and the class name is the whole diagnosis; `AslExecutor`'s own `catch (Error e)` already renders it the same way.

### What the cause says

- **A failed expression opens with AWS's sentence naming the expression and the field it sat in.** The cause used to be the JSONata message alone (`T0410: Argument 1 of function "abs" does not match function signature`), so a state with eight `Arguments` keys gave no way to tell which one broke. Across the 149-shape corpus in #2736, of the 101 shapes that fail on both sides with a real AWS message, all 101 omitted the sentence and 61 differ by nothing else once it is added.
- **`evaluate` stays the only place the JSONata engine is called and the only catch around it.** It throws `QueryEvaluationFailure`, a private `FailStateException` subclass carrying one extra field, `sentenceSeparator`. `evaluateField` catches that and rethrows it with the sentence in front, rather than reaching the engine a second time to get at the expression.
- **`queryEvaluationFailure` picks the separator at the one point that knows which kind of failure it is**: `". "` after a coded JSONata message, which is a sentence of its own, and `": "` before a bare bound message with no code. AWS ends the sentence with `. <CODE>: <message>` for 95 of the 101 corpus shapes and with `: <message>` for the other 6.
- **`evaluate`'s direct callers keep the bare tail unchanged.** A caller with no field has no sentence to write; a `grep` of `AslExecutor` finds 11 call sites into the evaluator, all `evaluateField(...)` or `resolveTemplate(...)` and zero direct `evaluate(...)`, so every real execution path goes through the enrichment and through the single `Error` arm.

## Wire-fidelity details (AWS CLI 2.33.7 against `us-east-1`)

Most rows were measured with `test-state`, which evaluates the state and creates nothing:

```bash
aws stepfunctions test-state --region us-east-1 \
  --role-arn arn:aws:iam::<account>:role/<role-with-states-assume> \
  --input '{}' \
  --definition '{"Type":"Pass","QueryLanguage":"JSONata","Output":{"r":"{% 1/0 %}"},"End":true}'
```

| --- | --- | --- |
| `1+1+…+1` at 100 terms, then 101 | evaluates, then `States.QueryEvaluationError` | `everyNestingShapeFailsAtTheDepthAwsFailsAt` |
| 99 parentheses, 99 brackets, 99 `~>` stages, then 100 of each | evaluates, then refuses: each nests one level more than its count | `everyNestingShapeFailsAtTheDepthAwsFailsAt` |
| `$f(31)` non-tail-recursive, nesting `3n+5`, then `$f(32)` | evaluates at depth 98, refuses at 101 | `nonTailRecursionFailsAtTheCallDepthAwsFailsAt` |
| the wording of a depth refusal | `Stack overflow error: Check for non-terminating recursive function. Consider rewriting as tail-recursive`, with no depth counter in it; the library's own `. Depth=101 max=100` suffix is stripped to match | `nonTailRecursionFailsOnTheDepthBound` |
| `[1..873782] ~> $count()`, then `[1..873783]` | `873782`, then `Expression evaluation memory limit exceeded: Check for excessive memory usage` | `theLargestSequenceAwsAcceptsEvaluatesAndTheNextOneFailsOnTheMemoryBound` |
| a string doubled 22 times, then 23 | `4194304`, then the same memory refusal | `aStringDoublingStopsAtTheSameBoundAwsStopsAt` |
| `[1..900000] ~> $count()` and `$sum([1..900000])` | the memory refusal, where floci answered `900000` and `405000450000` | `theSequencesAwsRefusesOnItsMemoryLimitFailTheStateHere` |
| `$eval("1+1")` | `T1006: Attempted to invoke a non-function` | `evalIsNotBoundBecauseAwsDisablesIt` |
| `1/0`, `1e308 * 10`, `-1/0` | `"Infinity"`, `"Infinity"`, `"-Infinity"` | `aNonFiniteNumberIsTheStringAwsWritesForIt` |
| `$parseInteger("abc","0")` bare, in an array, in an object | `"NaN"`, `["NaN"]`, `{"k":"NaN"}` | `parseIntegerAnswersNaNForAValueThePictureCannotRead` |
| `$formatNumber(1, "x")` | `D3086`, where floci answered `"x1"` | `formatNumberRefusesAPictureThatDescribesNoNumber` |
| twelve pictures, one per code: D3080 to D3084 and D3087 to D3093 | each its own code | `formatNumberNamesTheRuleThePictureBreaks` |
| `#,##0.00`, `#0.0%`, `$#,##0.00` | still formatted, unchanged | `formatNumberStillFormatsThePicturesTheRulesAllow` |
| `$abs("x")` in `Output/v` | `The JSONata expression '$abs("x")' specified for the field 'Output/v' threw an error during evaluation. T0410: Argument 1 of function "abs" does not match function signature` | `aFailedJsonataExpressionOpensWithTheSentenceNamingTheExpressionAndTheField` |
| `$sum(['a'])` in `Output/v` | the same sentence, then `. T0412: Argument ["a"] must be an array of "numbers"` | `aFailedJsonataExpressionReportsTheErrorCodeAndTheTypeItWanted` |
| A failure with no JSONata code (`$toMillis('nope')`) | the sentence joined to the bare message by `: ` | `evaluateFieldWrapsABoundFailureWithTheSentenceAndAColonSeparator`, which pins the sentence and the separator: the tail after them is dashjoin's own message, not AWS's |
| The memory and depth refusals through a field | the same `: ` join; the whole memory-bound cause is a captured AWS response, byte for byte | `evaluateFieldWrapsTheMemoryBoundFailureWithTheSentenceAndAColonSeparator`, `evaluateFieldWrapsTheDepthBoundFailureWithTheSentenceAndAColonSeparator` |
| An expression that returns nothing | `The JSONata expression '...' specified for the field '...' returned nothing (undefined).`, a different sentence and unchanged here | `evaluateFieldNamesTheFieldOfAPositionThatHoldsASingleExpression` |
| A `Parallel` branch whose expression doubles a string 31 times, with `Catch` on `States.ALL` | `SUCCEEDED`, output `{"caught":true}` | `parallelBranchErrorIsCatchableByStatesAll` |
| An `Error` out of the parser rather than the evaluator (200,000 nested parentheses) | the same treatment: a state failure the state can catch, not an execution-terminal one | `aStackOverflowErrorDuringParsingBecomesAQueryEvaluationError` |

The `Catch`-on-`States.ALL` row is the one measurement that needed a real execution rather than `test-state`; the state machine it created was deleted.

## Deliberate scope notes

- **Eight bytes a number is inside the bracket the AWS measurements leave, not derived from them.** The two data points are `[1..873782]` accepted and `[1..873783]` refused, and a string of 2^22 characters accepted and 2^23 refused. Reading the bound as `873782 * B` bytes, both are satisfied by any integer `B` from 5 to 9 (4 puts 2^22 characters outside the bound and 10 puts 2^23 inside), so the measurements pin the bound, not `B`. Eight is chosen because a JSON number is a double, and every expression either measurement covers falls the same way for any `B` in the bracket.
- **A non-finite number the arithmetic buries inside a value still fails the state, where AWS answers a string.** AWS returns `["Infinity"]` for `[1/0]` and `{"k":"Infinity"}` for `{'k': 1/0}`; here they fail with `States.QueryEvaluationError`. The library refuses to carry a non-finite number at all and raises instead, so the exception is the only place the value appears, and by then it has unwound whatever was being built around it. This is the conservative half of the divergence, since a state that fails is one a `Catch` fires on, and it is pinned rather than left silent by `aNonFiniteNumberTheArithmeticBuildsInsideAValueFailsTheStateInstead`. `0/0` is the same shape with no refusal to read at all: NaN is dropped as not-a-number, so the field that holds it fails.
- **`$eval` is bound to null, not removed from the dependency.** #2737 pointed at `pom.xml:210`, but that is the whole `jsonata` library and every other function in it is wanted. Binding the name to null is what produces AWS's exact `T1006`; removing the function would have to invent a code.
- **The time bound is untouched.** `EVALUATION_TIMEOUT_MILLIS` stays at 5,000 ms. Nothing in #2737 measures it, and a recursive expression with no base case is a tail call, so it loops rather than nesting and only the clock ends it.
- **One of the fourteen `$formatNumber` codes has no picture in the test map: D3085.** It is the code the last-rule-wins order exists for: a picture that breaks it usually breaks D3086 too, and AWS reports D3086, so a picture pinning it would pin the wrong code.
- **What the `Catch`-on-`States.ALL` run proves is the catchability, not the error name.** Its `Catch` is on `States.ALL`, so the reproduction never shows the name AWS puts on it; `States.QueryEvaluationError` is the name floci already uses for every other failure out of this same seam, and it is what a `Catch` written for a bad expression will be listening for.
- **The `OutOfMemoryError` arm is a guard with no cheap trigger left.** With the memory bound in place, issue #2738's own expression trips the bound at its 23rd doubling and fails through the `JException` path, so no test reaches the `OutOfMemoryError` arm any more; the arm stays because a single library step can still ask for an allocation larger than the heap. Its `StackOverflowError` sibling is pinned by `aStackOverflowErrorDuringParsingBecomesAQueryEvaluationError` (200,000 nested parentheses raise it during parsing in a few milliseconds), and the integration test keeps running the issue's expression end to end, now pinning that the bound's refusal is catchable.
- **The outer wrapper is a separate gap and is not touched here.** A real execution prepends `An error occurred while executing the state '<name>' (entered at the event id #<n>). ` to the whole cause; that is pinned as a known difference at `StepFunctionsJsonataIntegrationTest.java:1268`.
- **The other 40 corpus shapes stay different for reasons independent of the sentence.** `T0410`'s argument index and `T0412`'s offending value need data `JException` does not carry, so nothing here pretends to fix them.
- **Docs are prose only.** The change to `docs/services/step-functions.md` is the three bounds, the `$formatNumber` rule check and the non-finite paragraph; no generated action table is touched, so `make docs-sync` has nothing to regenerate here.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] The incorrect behaviour, in three parts: AWS failed the state and floci returned a value for `$eval`, for `$formatNumber` pictures that describe no number, for expressions past AWS's memory limit and for expressions nesting between depth 101 and 500, while floci failed the state and AWS returned a value for `1/0`, `1e308 * 10` and `$parseInteger("abc","0")`; a `java.lang.Error` out of the engine ended the execution `FAILED` with `States.Runtime` where AWS ends it `SUCCEEDED` through the state's own `Catch`; and the cause of a failed expression was the JSONata tail alone, without the sentence naming the expression and the field
- [x] Verified against real Step Functions in `us-east-1` with AWS CLI 2.33.7. Every value in the table and in the tests is a captured AWS response, not a value reasoned to
- [x] Both bounds were bisected one `n` at a time on both sides, which is what fixes 100 as the depth and 873,782 numbers as the memory ceiling rather than a round number near them
- [x] Every AWS resource created for the measurements was deleted; most of them ran through `test-state`, which creates none

## Checklist

- [x] `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'` green on the current tip: 602 tests, 0 failures, 0 errors, 0 skips
- [x] New tests: 12 in `JsonataEvaluatorTest` for the bounds and the pictures, 5 more for the sentence and its two separators (two of them byte-exact against captured AWS causes for the memory and depth refusals), `aStackOverflowErrorDuringParsingBecomesAQueryEvaluationError` for the parser arm, plus `aFailedJsonataExpressionOpensWithTheSentenceNamingTheExpressionAndTheField` and `parallelBranchErrorIsCatchableByStatesAll` end to end
- [x] Revert receipt: with `JsonataEvaluator.java` reverted to `origin/main` and `FormatNumberPicture.java` removed in a scratch worktree, tests untouched, 20 of the 602 go red (18 failures, 2 errors): 17 in `JsonataEvaluatorTest` (the depth refusal still reading `Depth=501 max=500`, the bounds, the pictures and the sentence), 2 in `StepFunctionsJsonataFunctionsIntegrationTest`, and the `Error` test failing with the issue verbatim, `"cause":"java.lang.OutOfMemoryError: Overflow: String length out of range"`
- [x] Two pre-existing tests updated rather than added to: `nonTailRecursionFailsOnTheDepthBound` now expects `Depth=101 max=100`, and the boundary-from-below test drops the `n=160` recursion and the loose comment about the bounds being above AWS's, since they no longer are
- [ ] Full `./mvnw test` not run: the production change is confined to `JsonataEvaluator` and the new `FormatNumberPicture`, whose every caller is in the targeted package

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -158,7 +158,7 @@ top of the JSONata language, alongside every function JSONata itself provides.
 
 | Function | Returns |
 | --- | --- |
-| `$parse(jsonString)` | the deserialized value; the replacement for `$eval`, which AWS disables |
+| `$parse(jsonString)` | the deserialized value; the replacement for `$eval`, which AWS disables and so does Floci, answering `T1006` to a call |
 | `$partition(array, chunkSize)` | `array` split into chunks of `chunkSize`, the last one holding the remainder |
 | `$range(start, end, step)` | the values from `start` to `end`, inclusive when `step` lands on `end` |
 | `$hash(str, algorithm)` | the hex digest of `str`; `algorithm` is `MD5`, `SHA-1`, `SHA-256`, `SHA-384` or `SHA-512`, case-sensitive |
@@ -179,16 +179,41 @@ JSONata's own `$string` follows AWS's number notation: a whole number is written
 `1e21` and in exponent notation from there, on both signs, so `$string(1e20)` is
 `100000000000000000000` and `$string(1e21)` is `1e+21`.
 
-Evaluation is bounded, as it is on AWS: one expression may run for five seconds and nest 500
-levels deep, and past either the state fails with `States.QueryEvaluationError`. Both bounds sit
-well above what AWS itself accepts, so an expression that evaluates there evaluates here. AWS
-refuses a non-tail-recursive function past a nesting depth near 100, and the largest sequence it
-accepts evaluates here in about a tenth of a second.
+JSONata's own `$formatNumber` checks its picture string against the fourteen rules of XPath F&O
+4.7.3, as AWS does, so `$formatNumber(1, "x")` fails with `D3086` rather than answering `x1`: a
+picture with no digit in it describes no number.
 
-One deviation. AWS also bounds the *memory* an expression may use, refusing
-`[1..900000] ~> $count()` with `Expression evaluation memory limit exceeded` while accepting the
-same expression at 800,000 elements. Floci bounds time and depth but not memory, and its ranges
-are lazy, so that expression answers immediately at any size.
+Evaluation is bounded on three axes, as it is on AWS, and past any of them the state fails with
+`States.QueryEvaluationError`.
+
+- **Depth**: one expression may nest 100 levels, which is AWS's own ceiling: AWS accepts `1+1+…+1`
+  at 100 terms, 99 parentheses, 99 brackets, 99 `~>` stages, and a non-tail-recursive `$f(31)`,
+  which nests `3n+5`, and refuses one level more of each. The refusal names the depth reached, as
+  in `Stack overflow error: … Depth=101 max=100`.
+- **Memory**: one value an expression builds may hold 6,990,256 bytes, counting a number as eight
+  bytes and a character as one. That is AWS's own bound: AWS accepts `[1..873782]` and refuses one
+  element more, and it accepts a string doubled 22 times, 2^22 characters, refusing the 23rd. The
+  refusal is AWS's own
+  `Expression evaluation memory limit exceeded`, which is what `[1..900000] ~> $count()` and
+  `$sum([1..900000])` now answer.
+- **Time**: five seconds, which is the library's own default and roughly fifty times the slowest
+  evaluation of a payload AWS itself accepts. A recursive expression with no base case is a tail
+  call, so it loops rather than nesting and only the clock ends it.
+
+JSON has no literal for a non-finite number, so AWS writes each one as the string JavaScript names
+it by, wherever it lands: `1/0` is `"Infinity"`, `1e308 * 10` is `"Infinity"`, `0/0` is `"NaN"` and
+`$parseInteger("abc", "0")` is `"NaN"`, and nested, `[1/0]` is `["Infinity"]` and `{"k": 1/0}` is
+`{"k": "Infinity"}`.
+
+One deviation, on the arithmetic half of that. A non-finite number a **function** answers is a value
+like any other here, so `$parseInteger("abc", "0")` is `"NaN"` on its own, inside an array and
+inside an object, as on AWS. One the **arithmetic** produces is not: the JSONata library refuses to
+carry it and raises instead, so Floci sees the value only in that refusal, after it has unwound
+whatever was being built around it. `1/0`, `-1/0` and `1e308 * 10` are answered because the
+expression's own result is what was refused; `[1/0]`, `{"k": 1/0}` and `$string(1/0)` fail the state
+with `States.QueryEvaluationError` where AWS answers a string, and `0/0` fails the field that holds
+it, because NaN is dropped as not-a-number without even a refusal to read it from. A state that
+fails is one a `Catch` fires on, which is the half of the divergence worth keeping.
 
 ## Nested workflows
 
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/FormatNumberPicture.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/FormatNumberPicture.java
@@ -1,0 +1,238 @@
+package io.github.hectorvent.floci.services.stepfunctions;
+
+import com.dashjoin.jsonata.JException;
+
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * The picture string of {@code $formatNumber}, checked against the rules XPath F&amp;O 4.7.3 gives
+ * it, which is what AWS applies and the bundled JSONata library does not: it formats through
+ * {@link java.text.DecimalFormat} and answers {@code "x1"} for {@code $formatNumber(1, "x")} where
+ * AWS fails the state with D3086.
+ *
+ * <p>A picture may break more than one of the fourteen rules, and the code reported is the last
+ * one, which is the one AWS reports: a picture with no digit at all breaks two at once, the
+ * mantissa holding no digit (D3085) and the active part holding a passive character (D3086), and
+ * D3086 is the answer.
+ *
+ * <p>The symbols the rules are written in terms of are the built-in ones, each replaceable through
+ * the third argument of {@code $formatNumber}.
+ */
+final class FormatNumberPicture {
+
+    private final char decimalSeparator;
+    private final char exponentSeparator;
+    private final char groupingSeparator;
+    private final char digit;
+    private final char patternSeparator;
+    private final char percent;
+    private final char perMille;
+    private final char zeroDigit;
+
+    private FormatNumberPicture(Map<?, ?> options) {
+        decimalSeparator = symbol(options, "decimal-separator", '.');
+        exponentSeparator = symbol(options, "exponent-separator", 'e');
+        groupingSeparator = symbol(options, "grouping-separator", ',');
+        digit = symbol(options, "digit", '#');
+        patternSeparator = symbol(options, "pattern-separator", ';');
+        percent = symbol(options, "percent", '%');
+        perMille = symbol(options, "per-mille", '‰');
+        zeroDigit = symbol(options, "zero-digit", '0');
+    }
+
+    private static char symbol(Map<?, ?> options, String name, char builtIn) {
+        Object override = options == null ? null : options.get(name);
+        return override instanceof String text && text.length() == 1 ? text.charAt(0) : builtIn;
+    }
+
+    /**
+     * Fails with the code of the rule the picture breaks, or returns when it breaks none. The
+     * picture is split into sub-pictures first, one for positive numbers and an optional second
+     * one for negative numbers.
+     */
+    static void validate(String picture, Map<?, ?> options) {
+        FormatNumberPicture symbols = new FormatNumberPicture(options);
+        String[] subPictures =
+                picture.split(Pattern.quote(String.valueOf(symbols.patternSeparator)), -1);
+        if (subPictures.length > 2) {
+            throw new JException("D3080", -1);
+        }
+        for (String subPicture : subPictures) {
+            String error = symbols.errorIn(symbols.split(subPicture));
+            if (error != null) {
+                throw new JException(error, -1);
+            }
+        }
+    }
+
+    /**
+     * The parts a sub-picture is read in. The prefix and the suffix are the passive characters at
+     * either edge, so the active part is what sits between them; the mantissa is the active part
+     * up to the exponent separator, and the integer and fractional parts are the mantissa either
+     * side of the decimal separator.
+     */
+    private record Parts(String subPicture, String activePart, String mantissaPart,
+                         String integerPart, String fractionalPart, String exponentPart) { }
+
+    private Parts split(String subPicture) {
+        int activeStart = 0;
+        while (activeStart < subPicture.length() && !isActiveOutsideExponent(subPicture.charAt(activeStart))) {
+            activeStart++;
+        }
+        int activeEnd = subPicture.length();
+        while (activeEnd > 0 && !isActiveOutsideExponent(subPicture.charAt(activeEnd - 1))) {
+            activeEnd--;
+        }
+        // A sub-picture with no active character at all is all active part and has no edges.
+        if (activeEnd == 0) {
+            activeStart = 0;
+            activeEnd = subPicture.length();
+        }
+        String activePart = subPicture.substring(activeStart, activeEnd);
+        int exponentAt = subPicture.indexOf(exponentSeparator, activeStart) - activeStart;
+        boolean hasExponent = exponentAt >= 0 && exponentAt <= activePart.length();
+        String mantissaPart = hasExponent ? upTo(activePart, exponentAt) : activePart;
+        String exponentPart = hasExponent ? from(activePart, exponentAt + 1) : null;
+        int decimalAt = mantissaPart.indexOf(decimalSeparator);
+        String integerPart = decimalAt < 0 ? mantissaPart : mantissaPart.substring(0, decimalAt);
+        String fractionalPart =
+                decimalAt < 0 ? subPicture.substring(activeEnd) : mantissaPart.substring(decimalAt + 1);
+        return new Parts(subPicture, activePart, mantissaPart, integerPart, fractionalPart, exponentPart);
+    }
+
+    /** The exponent separator is located in the sub-picture, so it can point past the active part. */
+    private static String upTo(String text, int end) {
+        return text.substring(0, Math.min(end, text.length()));
+    }
+
+    private static String from(String text, int start) {
+        return text.substring(Math.min(start, text.length()));
+    }
+
+    /**
+     * The code of the last rule the sub-picture breaks, or null when it breaks none. The rules are
+     * read in the order the codes number them, and a later one replaces what an earlier one found.
+     */
+    private String errorIn(Parts parts) {
+        String error = separatorCountError(parts.subPicture());
+        error = later(error, digitError(parts));
+        error = later(error, groupingSeparatorError(parts));
+        error = later(error, optionalDigitError(parts));
+        error = later(error, exponentError(parts));
+        return error;
+    }
+
+    private static String later(String error, String found) {
+        return found != null ? found : error;
+    }
+
+    /** D3081 to D3084: a symbol that may appear once appears twice, or two that exclude each other. */
+    private String separatorCountError(String subPicture) {
+        String error = null;
+        if (subPicture.indexOf(decimalSeparator) != subPicture.lastIndexOf(decimalSeparator)) {
+            error = "D3081";
+        }
+        if (subPicture.indexOf(percent) != subPicture.lastIndexOf(percent)) {
+            error = "D3082";
+        }
+        if (subPicture.indexOf(perMille) != subPicture.lastIndexOf(perMille)) {
+            error = "D3083";
+        }
+        if (subPicture.indexOf(percent) >= 0 && subPicture.indexOf(perMille) >= 0) {
+            error = "D3084";
+        }
+        return error;
+    }
+
+    /** D3085 and D3086: the mantissa holds no digit, and the active part holds a passive character. */
+    private String digitError(Parts parts) {
+        String error = null;
+        if (!holdsDigit(parts.mantissaPart())) {
+            error = "D3085";
+        }
+        if (!parts.activePart().chars().allMatch(character -> isActive((char) character))) {
+            error = "D3086";
+        }
+        return error;
+    }
+
+    /** D3087 to D3089: a grouping separator next to the decimal separator, at the end, or doubled. */
+    private String groupingSeparatorError(Parts parts) {
+        String subPicture = parts.subPicture();
+        int decimalAt = subPicture.indexOf(decimalSeparator);
+        String error = null;
+        if (decimalAt >= 0 && (characterAt(subPicture, decimalAt - 1) == groupingSeparator
+                || characterAt(subPicture, decimalAt + 1) == groupingSeparator)) {
+            error = "D3087";
+        } else if (decimalAt < 0 && !parts.integerPart().isEmpty()
+                && parts.integerPart().charAt(parts.integerPart().length() - 1) == groupingSeparator) {
+            error = "D3088";
+        }
+        if (subPicture.indexOf("" + groupingSeparator + groupingSeparator) >= 0) {
+            error = "D3089";
+        }
+        return error;
+    }
+
+    /**
+     * D3090 and D3091: an optional digit with a mandatory one on the side the number grows from,
+     * which is to its left in the integer part and to its right in the fractional part.
+     */
+    private String optionalDigitError(Parts parts) {
+        String error = null;
+        int firstOptional = parts.integerPart().indexOf(digit);
+        if (firstOptional >= 0 && holdsDecimalDigit(parts.integerPart().substring(0, firstOptional))) {
+            error = "D3090";
+        }
+        int lastOptional = parts.fractionalPart().lastIndexOf(digit);
+        if (lastOptional >= 0 && holdsDecimalDigit(parts.fractionalPart().substring(lastOptional))) {
+            error = "D3091";
+        }
+        return error;
+    }
+
+    /** D3092 and D3093: an exponent alongside a percent, and an exponent that is not all digits. */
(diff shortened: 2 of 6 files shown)


facuparedes:
The three commits on this branch shipped in f7f16cd2, the squash merge of #2790, which was stacked on this branch. On `main` today: `FormatNumberPicture.java` and `StepFunctionsJsonataFunctionsIntegrationTest.java` are byte-identical to this branch, the three evaluator tests are present, and the only lines this branch carries that `main` does not are the inline memory-refusal literal that #2790 replaced with a shared constant. The diff GitHub renders above is against the old merge base, and merging it now would conflict in `JsonataEvaluator.java`. Closing as already merged.

PR #2786, opened 2026-08-30
fix(stepfunctions): refuse the definitions AWS refuses, at both validation entry points

### LINKED ISSUE #2735: [BUG] Step Functions (SFN): CreateStateMachine accepts definitions AWS rejects, including a workflow with no terminal state

### LINKED ISSUE #2745: [BUG] Step Functions (SFN): JSONata expressions are never parsed at CreateStateMachine, so definitions AWS refuses are accepted
## Service

Step Functions (ASL definition validation, `QueryLanguage: JSONata`)

## AWS API Action

`ValidateStateMachineDefinition`, `CreateStateMachine`

## Expected behavior

AWS parses every JSONata expression in a definition before storing it. An expression that does not
parse is refused at definition time, with the parser's own message and the field's location:

```
An error occurred (InvalidDefinition) when calling the CreateStateMachine operation:
Invalid State Machine Definition: 'INVALID_JSONATA_EXPRESSION: Unexpected end of expression at /States/Main/Output'
```

## Actual behavior

Floci answers `{"result":"OK","diagnostics":[],"truncated":false}` and stores the definition. The
refusal arrives later, as a failed execution, with the message Floci's own parser already produced:

| `"Output": "{% <expr> %}"` | AWS, at definition time | Floci, at run time |
|---|---|---|
| `a[1,2)` | `INVALID_JSONATA_EXPRESSION` `Expected "]", got ","` | `S0202: Expected "]", got ","` |
| `"unterminated` | `INVALID_JSONATA_EXPRESSION` `String literal must be terminated by a matching quote` | `S0101: ` same text |
| `1 +` | `INVALID_JSONATA_EXPRESSION` `Unexpected end of expression` | `S0207: ` same text |
| `{"a":}` | `INVALID_JSONATA_EXPRESSION` `The symbol "}" cannot be used as a unary operator` | `S0211: ` same text |
| `$x :=` | `INVALID_JSONATA_EXPRESSION` `Unexpected end of expression` | `S0207: ` same text |
| `$match("a", /abc` | `INVALID_JSONATA_EXPRESSION` `No terminating / in regular expression` | `S0302: ` same text |
| `$$` | `UNSUPPORTED_JSONATA_EXPRESSION` `Reference to '$$' is not supported.` | `returned nothing (undefined).` |

Six of the seven are the same sentence on both sides, separated only by the `S0xxx: ` prefix AWS
strips and by the moment it arrives. The run-time status is `FAILED` with error
`States.QueryEvaluationError` in all seven.

A second name rule behaves the same way. `$states.errorOutput` is in scope only inside a `Catch`
entry's own `Output` and `Assign`; anywhere else AWS refuses the definition with
`UNSUPPORTED_JSONATA_EXPRESSION: Field '$states.errorOutput' does not exist.` Floci accepts it, runs
it, and the catch target then fails on the undefined field, so the original error is readable only
from the execution history and never from `DescribeExecution`.

## Reproduction

```bash

# Three of the seven expressions carry a quote or a brace that would break the surrounding
# JSON if interpolated into it, so the definition is built with a JSON encoder. Interpolating
# them makes Floci answer FAIL on a malformed definition, which is not the behaviour under test.
mkdef() {
  EXPR="$1" python3 -c 'import json, os
print(json.dumps({"QueryLanguage": "JSONata", "StartAt": "Main", "States": {
    "Main": {"Type": "Pass", "Output": "{%% %s %%}" % os.environ["EXPR"], "End": True}}}))'
}

for EXPR in 'a[1,2)' '"unterminated' '1 +' '{"a":}' '$x :=' '$match("a", /abc' '$$'; do
  printf '%-20s %s\n' "$EXPR" \
    "$(aws stepfunctions validate-state-machine-definition --definition "$(mkdef "$EXPR")" \
        --query result --output text)"
done
# a[1,2)               OK
# "unterminated        OK
# 1 +                  OK
# {"a":}               OK
# $x :=                OK
# $match("a", /abc     OK
# $$                   OK
#     AWS returns FAIL on all seven
```

The AWS side is the same loop, and `validate-state-machine-definition` needs no role and creates
nothing.

## Suggested fix

The parse already happens and its failure is discarded. `JsonataTopLevelReferences.in`
(`JsonataTopLevelReferences.java:62-64`) runs `new Parser().parse(expression)` on every expression
the validator walks and catches the failure into `LOG.debugf`, returning no reference. Reporting it
instead as `INVALID_JSONATA_EXPRESSION` covers six of the seven shapes with the message Floci
already produces, once the `S0xxx: ` prefix is stripped.

The walk that feeds it, `StepFunctionsService.collectTopLevelReferences` (line 1538), already visits
exactly the fields AWS parses as JSONata, including `Output`, `Assign` and `Arguments`, and already
routes its findings through `UNSUPPORTED_JSONATA_MARKER` and `toJsonataDiagnostic` (line 1149). Both
new rules are diagnostics added to that walk, not a new validator.

`$$` and `$states.errorOutput` are name rules, not parse rules: both expressions parse. They belong
beside the existing `Reference to '<name>' at the top level is not supported.` check (line 1565),
which is the same shape of message. `$states.errorOutput` needs the walk to know whether it is
inside a catcher, and that is the one piece of state this adds.

Same class as #2735, which covers the schema rules; this one is the expression parser.

## Environment

- Floci `1.7.0`, `main` at `09a05c2d`, JVM (`java -jar target/quarkus-app/quarkus-run.jar`)
- AWS CLI 2.33.7, `us-east-1`
- Both `CreateStateMachine` calls against real AWS were refused, so nothing was created.

facuparedes:
> **Independent of #2782 and #2783.** The three touch disjoint code and merge into `main` in any order with no conflict. The tree with all three applied runs `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'` at 681 tests, 0 failures, 0 errors, 0 skips.

## Summary

`ValidateStateMachineDefinition` and `CreateStateMachine` answered `{"result":"OK","diagnostics":[]}` and stored definitions real AWS refuses: a workflow with no terminal state, a state nothing routes to, a `Next` naming a state that does not exist, `TimeoutSeconds`, `Catch` or `Retry` on a state type that does not support them, a JSONata expression that does not parse, and two JSONata names that read a value AWS does not offer there. Every rule lands in `collectValidationErrors`, the single diagnostic source both APIs already read, which is why they arrive together rather than as two PRs conflicting on the same two lines. Closes #2735, closes #2745.

**This refuses definitions floci used to accept**: 17 field/type combinations, four graph shapes (no terminal state anywhere, a state nothing routes to, a transition naming a state that does not exist, a `StartAt` naming no state in its container), every JSONata expression that does not parse, and `$$` and `$states.errorOutput` read outside a catcher's own `Output` or `Assign`. Checked by running `origin/main`'s test sources against this branch's production code in a throwaway worktree: two failures, both tests that asserted the old behaviour. `StepFunctionsJsonataIntegrationTest.outputExpressionReturningNothingFailsTheStateAndIsCatchable` put a `Catch` on a `Pass`, and its state is now a `Task`; `StepFunctionsTopLevelReferenceIntegrationTest.syntaxErrorsAreOutOfScope` is now `syntaxErrorsAreRefusedAtDefinitionTime`. No other existing definition in the suite changed.

### The graph rules

- **`hasTerminalState` and `validateReachability`**: `MISSING_END_STATE` when no state is `Succeed`, `Fail` or `"End": true`, and it suppresses the reachability walk at that same level, so a workflow with no terminal state is not also buried in "not reachable" diagnostics.
- **`validateTransitionTargets` and `validateTarget`**: `Next`, `Choices[].Next`, `Default` and `Catch[].Next` are each checked against the state names of their own container, at their own structured location. Independent of the reachability walk, so a dangling target still fires when `MISSING_END_STATE` suppressed it.
- **`validateNestedStates` now takes the container's `StartAt`**, so `ItemProcessor` and `Branches[i]` are walked for reachability at `/States/M/ItemProcessor/States/...` and `/States/P/Branches[0]/States/...`.
- **`StartAt` is a transition target too, at every level**: a `StartAt` naming no state in its container is `Missing 'Next' target: <name>` (the literal word `Next`, as for the other four positions) at `<container>/StartAt`, and the reachability walk then proceeds from an empty set, so every state in that container is also reported unreachable. This replaces the old top-level-only check and its own message (`The value of 'StartAt' must match a state in 'States'`, which carried no code and no location).

### The field-per-state-type rule

- **`validateFieldsAllowedForType` against `FIELDS_ALLOWED_STATE_TYPES`**: the general rule, measured across every state type. `TimeoutSeconds` on `Task` only; `Catch` and `Retry` on `Task`, `Parallel` and `Map` only. That is 17 refused field/type combinations, all 17 of which floci accepted.

### The JSONata rules

- **The parse already ran and its failure was thrown away.** `JsonataTopLevelReferences.in` called `new Parser().parse(expression)` on every expression the validator walks and dropped the failure into `LOG.debugf`. The divergence was never in the parser: floci's bundled `com.dashjoin:jsonata` 0.9.10 produces AWS's `INVALID_JSONATA_EXPRESSION` sentences byte for byte, only its `S0xxx: ` code prefix set apart, and `JException#getMessage()` does not carry that prefix. So this reports the result the parse already had.
- **`JsonataTopLevelReferences.analyze` returns an `Analysis` record**: either the parser's message or the top-level references, never both. The `Logger` and the swallowing `catch` are gone.
- **`collectTopLevelReferences` raises `INVALID_JSONATA_MARKER`** with that message at the field's location; `toJsonataDiagnostic` takes the code as a parameter, so `INVALID_JSONATA_EXPRESSION` and the existing `UNSUPPORTED_JSONATA_EXPRESSION` share one renderer and one `CreateStateMachine` path.
- **Two name rules, both of which parse, ride the same walk**: `$$` (`Reference to '$$' is not supported.`) and `$states.errorOutput` outside a `Catch` entry's own `Output` or `Assign` (`Field '$states.errorOutput' does not exist.`), both in `addUnsupportedReferenceError`.
- **`JsonataScope`, a value type**, names which JSONata reading rules are in scope at the current node. `insidePayload` says AWS parses every key below as a payload key, so the ASL field names it does not parse no longer apply; `atCatchEntry` says the node is a catcher object; `insideCatchOutputOrAssign` says the walk is inside that catcher's own `Output` or `Assign`, at any depth, which is exactly what makes `$states.errorOutput` legal there and nowhere else. Threading the three separately would have taken `collectTopLevelReferences` to six parameters; one scope value keeps it at four.

- **`flatten(Diagnostic)`**: `CreateStateMachine` renders every diagnostic through the same mapping the validate API uses, instead of joining raw validator strings under a fixed `SCHEMA_VALIDATION_FAILED` prefix.

## Wire-fidelity details

Every row is a response from real AWS `us-east-1` (AWS CLI 2.33.7), re-checkable with no role and nothing created:

```
aws stepfunctions validate-state-machine-definition --region us-east-1 --definition '<definition>'
```

| Definition | AWS returns | Pinned by |
| --- | --- | --- |
| The bare loop `A → B → A` | one diagnostic, `MISSING_END_STATE`, `Workflow has no terminal state`, and **no `location` field at all** | `loopWithNoTerminalState_returnsFailWithMissingEndState` |
| That loop plus a state nothing routes to | still one diagnostic: `MISSING_END_STATE` suppresses the unreachable-state walk | `missingEndStateSuppressesUnreachableStateDiagnosticForOtherStates` |
| That loop plus an unreachable terminal `C` | `MISSING_TRANSITION_TARGET`, `State "C" is not reachable.`, `/States/C` | `unreachableTerminalStateAddedToLoop_returnsFailWithMissingTransitionTarget` |
| `A` with `"Next": "NOPE"` and nothing else | **two** diagnostics in one response: `MISSING_TRANSITION_TARGET` (`Missing 'Next' target: NOPE`, `/States/A/Next`) then `MISSING_END_STATE` (no location) | `missingEndStateCoexistsWithADanglingTargetInAWSOrder` |
| A dangling `Choices[].Next`, `Default` or `Catch[].Next` | the literal word `Next` in `Missing 'Next' target: <name>` for all four positions, never the position's own field name, at `/States/A/Next`, `/States/C/Choices[0]/Next`, `/States/A/Default`, `/States/T/Catch[0]/Next` | `danglingNextTarget_returnsFailAtNextLocation` and the three beside it |
| An `ItemProcessor` or a `Parallel` branch with no terminal state | `OK`: `MISSING_END_STATE` is a top-level rule only | `nestedLoopWithNoTerminalState_isAcceptedBecauseMissingEndStateIsTopLevelOnly` |
| An unreachable state inside `ItemProcessor` or a `Parallel` branch | `MISSING_TRANSITION_TARGET` at the structured path, e.g. `/States/P/Branches[0]/States/Ghost` | `unreachableStateInsideItemProcessor_returnsFailAtStructuredPath`, `unreachableStateInsideParallelBranch_returnsFailAtStructuredPath` |
| `StartAt` naming no state, top level and inside an `ItemProcessor` | two diagnostics each time: `Missing 'Next' target: Nope` at `/StartAt` (or `/States/M/ItemProcessor/StartAt`), then `State "A" is not reachable.` for the container's states | `danglingStartAtTopLevel_returnsMissingTransitionTargetAndUnreachableStates`, `danglingStartAtInsideItemProcessor_returnsFailAtStructuredPaths` |
| A state named `a/b` nothing routes to | accepted as a name; `State "a/b" is not reachable.` at `/States/a/b`, the full name in the message | `unreachableStateWithSlashInName_returnsFullNameInMessage` |
| `TimeoutSeconds` on `Map`, `Parallel` or `Pass`; `Catch` on `Pass`; `Retry` on `Wait` | `SCHEMA_VALIDATION_FAILED`, `Field '<name>' is not supported`, located at the **state**, not at the field | `timeoutSecondsOnMap_returnsFailWithSchemaError` and the four beside it |
| `TimeoutSeconds` on `Task`; `Catch` and `Retry` on `Task`, `Parallel`, `Map` | `OK`, and `Catch` on a `Pass` is refused whatever the query language | `timeoutSecondsAcceptedOnTask`, `catchAndRetryAcceptedOnTaskParallelAndMap`, `catchRejectedOnPassEvenWhenJsonata` |
| `a[1,2)` | `INVALID_JSONATA_EXPRESSION`, `Expected "]", got ","`, `/States/Main/Output` | `unparsableJsonataExpression_returnsFailWithInvalidJsonataExpression` |
| `"unterminated`, `1 +`, `$x :=`, `{"a":}`, `$match("a", /abc`, `phone %.other`, the empty expression | `String literal must be terminated by a matching quote`, `Unexpected end of expression` twice, `The symbol "}" cannot be used as a unary operator`, `No terminating / in regular expression`, `The symbol "." cannot be used as a unary operator`, `Unexpected end of expression` | same, parameterized |
| `$$` | `UNSUPPORTED_JSONATA_EXPRESSION`, `Reference to '$$' is not supported.`, a separate rule because it parses | `doubleDollarReference_returnsFailWithUnsupportedJsonataExpression` |
| `$states.errorOutput` in a state's own `Output` | `UNSUPPORTED_JSONATA_EXPRESSION`, `Field '$states.errorOutput' does not exist.` | `statesErrorOutputOutsideCatch_returnsFailWithUnsupportedJsonataExpression` |
| `$states.errorOutput` in a `Catch` entry's own `Output`, and in its `Assign` | `OK` | `statesErrorOutputInsideCatchOutput_isAccepted`, `statesErrorOutputInsideCatchAssign_isAccepted` |
| `create-state-machine` on the bare loop, and on `TimeoutSeconds` on a `Parallel` | `InvalidDefinition`, e.g. `Invalid State Machine Definition: 'MISSING_END_STATE: Workflow has no terminal state at null'`, the literal word `null` where the diagnostic has no location | `createStateMachineRefusesLoopWithNoTerminalState`, `createStateMachineRefusesTimeoutSecondsOnParallel` |
| `create-state-machine` on `{% phone[1,2) %}` | `InvalidDefinition`, `Invalid State Machine Definition: 'INVALID_JSONATA_EXPRESSION: Expected "]", got "," at /States/E/Output/v'` | `createStateMachineRefusesASyntaxError` |

Two asymmetries are pinned rather than smoothed over. `CreateStateMachine` always renders the ` at <location>` suffix and prints `null` when there is none, while `ValidateStateMachineDefinition` omits the `location` field entirely for the same finding. And each of the eight parse messages is asserted against the message floci's own parser already returned, in `JsonataTopLevelReferencesTest.reportsTheParserMessageWhenTheExpressionDoesNotParse`: the two sides are the same sentence, which is why the fix is one report and no new message table.

## Deliberate scope notes

- **`AslExecutor` keeps its `Parallel` `TimeoutSeconds` deadline**, which this PR makes unreachable: once the validator refuses the field, no definition carrying one reaches execution. Removing that branch here would collide with #2782, which folds the execution budget into the same join, and the two PRs are otherwise independent and merge in either order. It is a two-line deletion once both have landed.
- **`MISSING_END_STATE` suppresses only the unreachable-state half.** The dangling-target check is a separate walk because AWS reports both in one response, in that order, and a suppressed reachability walk must not swallow it.
- **`CreateStateMachine` now repeats the code per diagnostic when there is more than one.** The old message emitted `SCHEMA_VALIDATION_FAILED:` once and joined raw validator strings after it, which is not what AWS renders for a multi-diagnostic definition. For a single diagnostic the two are byte-identical, which is why no existing assertion moved.
- **Every marker payload separates its text from its location with `U+0001`**, not a space or a slash: the flat error strings are parsed back into structured diagnostics, and a state name or a parser message may legally contain either. The dangling-target, unreachable-state and parse-failure rules share the one separator, which is what keeps a state named `a/b` and a string literal containing ` at /States/...` intact in their diagnostics.
- **The `S0xxx: ` prefix is not stripped by this change.** `JException#getMessage()` never carries it; the code is on the exception's own field. Nothing in the walk edits the parser's sentence.
- **Only `$states.errorOutput`'s first two path steps decide the rule.** `$states.errorOutput.Cause` still reads the same field on `$states`, and `$states.input.errorOutput` is a different path entirely, so the lookalikes are pinned as accepted in `namesNothingForALookalikeStatesPath`.
- **The set of walked fields is unchanged.** This adds two diagnostics to the existing walk, which already visits exactly the fields AWS parses as JSONata; a `{% %}` string in `Comment`, `Next`, `Default`, `Resource`, `ErrorEquals`, `Retry`, `Credentials` or `ReaderConfig.CSVHeaders` is still left alone.
- **`analyze` reports one finding per expression.** A parse failure returns before the reference walk, because there is no tree to walk and AWS answers with the parse error alone.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- [x] The incorrect behaviour: floci answered `{"result":"OK","diagnostics":[]}` and stored the definition for all four shapes in #2735, and for every unparsable expression in #2745, failing the execution later with `States.QueryEvaluationError`
- [x] Every message, code and location in the table is a captured response from real Step Functions in `us-east-1`, AWS CLI 2.33.7
- [x] The general field/type rule was measured across every state type, not inferred from the two cases in the issue
- [x] `validate-state-machine-definition` needs no role and creates nothing, so every row above is free to re-run; every `create-state-machine` call was refused, so nothing was created in AWS

## Checklist

- [x] `./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'` green on the current tip: 631 tests, 0 failures, 0 errors, 0 skips, against 582 on `origin/main`
- [x] Integration tests added: 38 in `StepFunctionsValidateStateMachineDefinitionIntegrationTest` covering both APIs (8 of them the parameterized parse cases) and 1 in `StepFunctionsTopLevelReferenceIntegrationTest`
- [x] `JsonataTopLevelReferencesTest` moves to `analyze` and asserts the parse messages the old test asserted absent: 40 tests on `origin/main`, 50 here
- [x] Two existing tests changed, both because they asserted the old behaviour: `outputExpressionReturningNothingFailsTheStateAndIsCatchable` moves its `Catch` off a `Pass` and onto a `Task`, and `syntaxErrorsAreOutOfScope` becomes `syntaxErrorsAreRefusedAtDefinitionTime`
- [ ] Full `./mvnw test` not run: the run above is the `stepfunctions` subtree, which is every package this diff touches

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -272,10 +272,17 @@ in a predicate, in a sort term or in an object grouping stays legal and
 `$states.input.items[value > 3]` is accepted. A lambda body keeps the context of the expression
 that defines it, so `$map($states.input.a, function($x){ b })` does name `b` at the top level.
 
-Two neighbouring rules stay with the execution rather than the definition. A syntax error such as
-`{% a[1,2) %}` is accepted here and fails the execution, where AWS reports
-`INVALID_JSONATA_EXPRESSION` when the state machine is created; and `$$`, which AWS refuses under
-its own message, is accepted here.
+A JSONata expression that fails to parse, such as `{% a[1,2) %}`, is refused the same way, with
+`INVALID_JSONATA_EXPRESSION` and the parser's own message at the field's location. `$$`, the
+reference to the top-level context, is refused under `UNSUPPORTED_JSONATA_EXPRESSION` with the
+message `Reference to '$$' is not supported.`; so is `$states.errorOutput` outside the one place it
+resolves, a catcher's own `Output` or `Assign`.
+
+A definition is refused outright, with no state machine created, for a graph that never reaches a
+terminal state (`MISSING_END_STATE`), for a `StartAt`, `Next`, `Default` or `Catch[].Next` naming a
+state absent from its container or a state nothing transitions to (`MISSING_TRANSITION_TARGET`),
+and for a field the state type does not carry: `TimeoutSeconds` only on `Task`, and `Catch`/`Retry`
+only on `Task`, `Parallel` and `Map`.
 
 ## Mocked service integrations
 
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataTopLevelReferences.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/JsonataTopLevelReferences.java
@@ -2,7 +2,6 @@ package io.github.hectorvent.floci.services.stepfunctions;
 
 import com.dashjoin.jsonata.Parser;
 import io.quarkus.runtime.annotations.RegisterForReflection;
-import org.jboss.logging.Logger;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
@@ -14,7 +13,8 @@ import java.util.Objects;
 import java.util.Set;
 
 /**
- * Names every reference a Step Functions JSONata expression makes against the top-level context.
+ * Parses a Step Functions JSONata expression once and reports either the message the parser refused
+ * it with, or every reference it makes against the top-level context.
  *
  * <p>A Step Functions expression is evaluated with no context item: the input arrives as
  * {@code $states.input} and workflow variables as {@code $name}. A path that starts from neither
@@ -39,10 +39,19 @@ import java.util.Set;
 @RegisterForReflection(targets = Parser.Symbol.class, fields = true)
 final class JsonataTopLevelReferences {
 
-    private static final Logger LOG = Logger.getLogger(JsonataTopLevelReferences.class);
-
     /** The name AWS prints for {@code $}, the context item itself. */
     private static final String CONTEXT_ITEM = "$";
+
+    /** The pseudo-reference reported for {@code $$}, AWS's separate name rule. */
+    static final String ROOT_REFERENCE = "$$";
+
+    /**
+     * The pseudo-reference reported for a {@code $states.errorOutput} path, wherever it starts.
+     * Not a top-level context read like the others: {@code $states} is always in scope, but the
+     * {@code errorOutput} field only exists on it inside a {@code Catch} entry's own {@code Output}
+     * and {@code Assign}, and only the caller knows whether the walk is there.
+     */
+    static final String STATES_ERROR_OUTPUT = "$states.errorOutput";
 
     private static final Map<String, Field> TREE_FIELDS = resolveTreeFields(
             "type", "value", "steps", "lhs", "rhs", "expression", "expressions",
@@ -52,22 +61,32 @@ final class JsonataTopLevelReferences {
     }
 
     /**
-     * The distinct top-level references in {@code expression}, in the order they are written.
-     * An expression that does not parse names none: a syntax error is AWS's separate
-     * {@code INVALID_JSONATA_EXPRESSION} rule, and naming nothing leaves the definition accepted.
+     * What parsing {@code expression} once finds: either the parser's own message, unchanged
+     * (AWS's {@code INVALID_JSONATA_EXPRESSION} refuses it with the same sentence, its
+     * {@code S0xxx: } code prefix stripped, and {@link com.dashjoin.jsonata.JException#getMessage()}
+     * never carries that prefix), or the distinct top-level references the expression makes, in
+     * the order they are written.
      */
-    static List<String> in(String expression) {
+    record Analysis(List<String> topLevelReferences, String parseError) {
+        static Analysis parseError(String message) {
+            return new Analysis(List.of(), message);
+        }
+
+        static Analysis references(List<String> names) {
+            return new Analysis(names, null);
+        }
+    }
+
+    static Analysis analyze(String expression) {
         Parser.Symbol root;
         try {
             root = new Parser().parse(expression);
         } catch (Exception e) {
-            LOG.debugf(e, "JSONata expression not parsed, so no top-level reference is named: %s",
-                    expression);
-            return List.of();
+            return Analysis.parseError(e.getMessage());
         }
         Set<String> names = new LinkedHashSet<>();
         collect(root, names);
-        return new ArrayList<>(names);
+        return Analysis.references(new ArrayList<>(names));
     }
 
     private static void collect(Parser.Symbol node, Set<String> names) {
@@ -76,7 +95,12 @@ final class JsonataTopLevelReferences {
             return;
         }
         switch (type) {
-            case "path" -> collect(first(node, "steps"), names);
+            case "path" -> {
+                collect(first(node, "steps"), names);
+                if (isStatesErrorOutputPath(node)) {
+                    names.add(STATES_ERROR_OUTPUT);
+                }
+            }
             case "name" -> names.add(String.valueOf(read(node, "value")));
             case "variable" -> collectContextItem(node, names);
             case "unary" -> collectUnary(node, names);
@@ -105,13 +129,34 @@ final class JsonataTopLevelReferences {
 
     /**
      * {@code $} carries the empty name and is the context item AWS refuses. {@code $$} carries the
-     * name {@code "$"} and AWS refuses it under a different message, which is a separate rule.
-     * Every other variable is a workflow variable or a function, and AWS accepts an undefined one.
+     * name {@code "$"} and AWS refuses it too, under a different message. Every other variable is a
+     * workflow variable or a function, and AWS accepts an undefined one.
      */
     private static void collectContextItem(Parser.Symbol variable, Set<String> names) {
-        if ("".equals(read(variable, "value"))) {
+        String value = String.valueOf(read(variable, "value"));
+        if (value.isEmpty()) {
             names.add(CONTEXT_ITEM);
+        } else if (CONTEXT_ITEM.equals(value)) {
+            // "$$" parses as a variable named "$": the sigil is consumed twice, once as the
+            // variable marker and once as its name.
+            names.add(ROOT_REFERENCE);
         }
+    }
+
+    /**
+     * Whether {@code path}'s first step is {@code $states} and its second is {@code errorOutput}.
+     * Only those two steps matter: whatever follows ({@code $states.errorOutput.Cause}) still reads
+     * the same field on {@code $states} first.
+     */
+    private static boolean isStatesErrorOutputPath(Parser.Symbol path) {
+        List<Parser.Symbol> steps = children(path, "steps");
+        if (steps == null || steps.size() < 2) {
+            return false;
+        }
+        Parser.Symbol first = steps.get(0);
+        Parser.Symbol second = steps.get(1);
+        return "variable".equals(read(first, "type")) && "states".equals(read(first, "value"))
+                && "name".equals(read(second, "type")) && "errorOutput".equals(read(second, "value"));
     }
 
     /**
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/StepFunctionsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/StepFunctionsService.java
@@ -82,11 +82,16 @@ public class StepFunctionsService implements Resettable, ResourceProvider {
     private static final Set<String> RESULT_WRITER_OUTPUT_TYPES = Set.of("JSON", "JSONL");
     // Measured against real AWS: TimeoutSeconds is accepted on Task only, Catch and Retry are
     // accepted on Task, Parallel and Map only. Every other state type refuses these fields with
-    // "Field '<name>' is not supported".
-    private static final Map<String, Set<String>> FIELDS_ALLOWED_STATE_TYPES = Map.of(
-            "TimeoutSeconds", Set.of("Task"),
-            "Catch", Set.of("Task", "Parallel", "Map"),
-            "Retry", Set.of("Task", "Parallel", "Map"));
+    // "Field '<name>' is not supported". A List, not a Map.of: Map.of's iteration order is salted
+    // per JVM, so a state carrying more than one disallowed field needs a fixed emission order
+    // instead. TimeoutSeconds, Catch, Retry is not an AWS-observed order — field declaration order
+    // is not observable in the response — it is simply the one this code commits to.
+    private record FieldStateTypeRule(String field, Set<String> allowedTypes) {}
+
+    private static final List<FieldStateTypeRule> FIELDS_ALLOWED_STATE_TYPES = List.of(
+            new FieldStateTypeRule("TimeoutSeconds", Set.of("Task")),
+            new FieldStateTypeRule("Catch", Set.of("Task", "Parallel", "Map")),
+            new FieldStateTypeRule("Retry", Set.of("Task", "Parallel", "Map")));
 
     @Inject
     public StepFunctionsService(StorageFactory storageFactory, RegionResolver regionResolver,
@@ -1064,9 +1069,12 @@ public class StepFunctionsService implements Resettable, ResourceProvider {
     private static final String UNSUPPORTED_FIELD_MARKER = "UNSUPPORTED_FIELD:";
     private static final String MISSING_END_STATE_MARKER = "MISSING_END_STATE:";
     private static final String UNREACHABLE_STATE_MARKER = "UNREACHABLE_STATE:";
-    // Payload is "<target state name><SOH><location>"; SOH cannot appear in a state name.
+    // Payload is "<value><SOH><location>", shared by every marker that must carry structured data
+    // (a state name, a JSONata parser message) past its flat-string marker without re-parsing free
+    // text; SOH cannot appear in a state name or in JSONata parser output.
     private static final String DANGLING_TARGET_MARKER = "DANGLING_TARGET:";
-    private static final char DANGLING_TARGET_SEPARATOR = '\u0001';
+    private static final char MARKER_PAYLOAD_SEPARATOR = '\u0001';
+    private static final String INVALID_JSONATA_MARKER = "INVALID_JSONATA_EXPRESSION:";
 
     // Parse the structured location out of validator flat error strings,
     // which currently encode it as "...field 'X' ... at /States/Y".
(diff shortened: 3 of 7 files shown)


pgermosen (approved):
Approving. Traced the graph-validation path end to end rather than just trusting the write-up - the dangling-target-before-MISSING_END_STATE ordering the tests assert really does fall out of the code (per-state validation runs before the terminal-state check, in that order). The $states.errorOutput scoping is exactly as careful as described: only fires on paths whose first two steps are $states then errorOutput, so the lookalikes stay accepted. Good work.

PR #2788, opened 2026-08-30
fix(cloudformation): keep a generated physical name across stack updates

hectorvent:
## Summary

`provision` runs again on every `UpdateStack`. Seven provisioners generated a fresh random
name unconditionally, so an update created a second resource and orphaned the first with
its data. The schemas make the intent explicit: for these types the name is a
`createOnlyProperty`, so an unchanged template must keep its physical id.

`ProvisionContext.stablePhysicalName` encapsulates the rule (template name, else the prior
name, else a generated one) for S3::Bucket, SSM::Parameter, ECR::Repository, Pipes::Pipe,
KinesisFirehose::DeliveryStream and CloudWatch::Alarm. SNS reads the prior name back from
the `TopicName` attribute, since its physical id is the ARN.

Review follow-up (Greptile P1s, confirmed): a stable name sends the second `UpdateStack` to
a create API with an existing name. `ProvisionContext.reusesPriorEntity` (update AND the
derived name equals the prior id; deliberately not `isUpdate`, since a replacing update must
still create) now routes Firehose to `UpdateDestination`, Pipes to `UpdatePipe`, S3 past
`CreateBucket` (idempotent only in us-east-1, `BucketAlreadyOwnedByYou` elsewhere), and
makes ECR's already-exists path reapply `ImageTagMutability` and tags.

Second round (Greptile P1s at 54ea838cb, all confirmed against botocore, the registry schemas
and the Terraform provider's `tags_gen.go`): the reconcile paths now drive tags to the
template's desired state. Neither `UpdateDestination` nor `UpdatePipe` carries tags and a tag
call alone never removes one, so dropped keys are untagged before the desired set is applied,
as the schemas' update handlers declare (`firehose:UntagDeliveryStream`, `pipes:UntagResource`,
`ecr:UntagResource`). ECR no longer adopts a repository whose name is already taken: a create
on a colliding name fails with `RepositoryAlreadyExistsException` as on CloudFormation (a CDK
bootstrap re-run is an `UpdateStack` of the existing `CDKToolkit` stack and stays on the reuse
path), a template that drops `LifecyclePolicy` or `RepositoryPolicyText` deletes the stored
one, and a dropped `ImageTagMutability` resets to the `MUTABLE` default the API applies when the
parameter is omitted (third-round Greptile comment, confirmed against the schema description,
botocore and the Terraform provider's `Default: MUTABLE`). A Pipes update that changes the createOnly `Source` while reusing the name is refused with
`ValidationError`.

Follow-up, deliberately not here: `AWS::SQS::Queue` and `AWS::Lambda::MicrovmImage` have
the same defect (`SqsCfnProvisionerTest` exists to extend for the SQS one);
`ImageScanningConfiguration` on ECR (no `PutImageScanningConfiguration` in the emulator yet);
`Tags` on the S3, SNS, SSM and CloudWatch alarm provisioners; and `UpdateStack` dropping the
stack-level `Tags` parameter.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

A template omitting the resource's name got a new resource on every `UpdateStack`, orphaning
the previous one. On AWS the generated physical id is stable for an unchanged resource;
changing a `createOnly` name is what triggers replacement, and not changing it must not.
After the first fix, the second update of an unnamed Firehose stream or Pipe failed with
`ResourceInUseException` / `ConflictException`, and an unnamed bucket outside us-east-1 with
`BucketAlreadyOwnedByYou`; those paths now update the existing resource as AWS does.

## Checklist

CloudFormation + Cloud Control package plus the ECR, Pipes and Firehose suites: 1063 tests,
0 failures (plus the mutability-default case). `CfnStableNameUpdatePathTest` provisions Firehose, Pipes, S3 and ECR twice and pins
exactly one create, the tag diff, the Pipes source rejection, the ECR replacing update and the
ECR collision failure; mutation-checked by reverting the Pipes and S3 guards, the Firehose untag,
the Pipes source check, and the ECR ownership gate (each fails at least one test).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/CloudWatchCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/CloudWatchCfnProvisioner.java
@@ -32,10 +32,8 @@ public class CloudWatchCfnProvisioner implements CfnResourceProvisioner {
 
     @Override
     public void provision(StackResource r, JsonNode props, ProvisionContext ctx) {
-        String name = ctx.resolveOptional(props, "AlarmName");
-        if (name == null || name.isBlank()) {
-            name = ctx.generatePhysicalName(r.getLogicalId(), ALARM_NAME_MAX_LENGTH, false);
-        }
+        String name = ctx.stablePhysicalName(ctx.resolveOptional(props, "AlarmName"),
+                r.getLogicalId(), ALARM_NAME_MAX_LENGTH, false);
 
         MetricAlarm alarm = new MetricAlarm();
         alarm.setAlarmName(name);
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/EcrCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/EcrCfnProvisioner.java
@@ -1,7 +1,6 @@
 package io.github.hectorvent.floci.services.cloudformation.provisioners;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.services.cloudformation.model.StackResource;
 import io.github.hectorvent.floci.services.ecr.EcrService;
 import io.github.hectorvent.floci.services.ecr.model.Repository;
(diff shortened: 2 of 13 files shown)


pgermosen (approved):
Approving. stablePhysicalName is a clean fix and the mechanical swap into the six provisioners checks out - no behavior change beyond the bug being fixed. The SNS special-casing (reading TopicName back off the attribute instead of the ARN physical id) is the right call and well explained. Tests genuinely pin the defect shape, not just the happy path.

One small thing for later, not blocking this: SqsCfnProvisioner has the identical bug (unconditional generatePhysicalName), and SqsCfnProvisionerTest already exists with a queueWithoutNameGeneratesPhysicalName test right there to extend, so "no focused provisioner tests to extend" doesn't quite hold for SQS specifically - just Lambda::MicrovmImage. Worth folding into whichever follow-up picks those two up.

PR #2794, opened 2026-08-30
refactor(cloudformation): move AWS::S3::* into its own provisioner

hectorvent:
## Summary

**#2754 is marked merged but its content never reached `main`.** This re-lands it, unchanged.

What happened: #2754's base was `refactor/cfn-leaf-provisioners` (the #2744 branch), which is how a
stacked PR is meant to be set up. But #2744 merged into `main` at 18:51, and #2754 then merged into
`refactor/cfn-leaf-provisioners` at 19:11 — twenty minutes later, into a branch `main` had already
absorbed. The commit landed somewhere nothing merges from.

Verifiable from the repo state: `S3CfnProvisioner.java` does not exist on `main`,
`refactor/cfn-leaf-provisioners` is not an ancestor of `main`, and it still carries
`f40d03256 refactor(cloudformation): move AWS::S3::* into its own provisioner (#2754)`.

This branch is that commit cherry-picked onto current `main`. It applied cleanly — the leaf batch
that landed in between touches different regions of the monolith.

**Contents are identical to the approved #2754**: `AWS::S3::Bucket` and `AWS::S3::BucketPolicy` move
into `S3CfnProvisioner`; the bucket keeps its exact physical id and `Fn::GetAtt` keys; the CORS and
versioning helpers move with it; `s3Service` stays on the monolith because the Lambda code probe and
the ApiGatewayV2 OpenAPI body still use it.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

`refactor(cloudformation):` — no behaviour change, release-neutral.

## AWS Compatibility

N/A for the wire protocol. Physical id and attribute keys are unchanged, verified against
`aws-s3-bucket.json`. #2754 additionally passed the full compat-cdk suite (20/20, deploy through
destroy) on identical content.

## Checklist

CloudFormation + Cloud Control on current `main`: **854 tests, 0 failures, 0 errors**. The inventory
guard is the meaningful check here — `CfnResourceInventoryTest` compares the checked-in inventory
against the CDI-resolved registry, so it would fail immediately if the cherry-pick had produced a
stale or mismatched TSV against a `main` that has moved.

## Note for the rest of the stack

The same hazard applies to every remaining stacked PR: **if the base branch merges first, merging
the child into that stale base puts the work somewhere `main` never sees.** Retarget each child to
`main` before merging it, or merge strictly bottom-up without letting the base branch merge in
between.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -271,8 +271,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::RDS::DBSubnetGroup",
             "AWS::Route53::HostedZone",
             "AWS::Route53::RecordSet",
-            "AWS::S3::Bucket",
-            "AWS::S3::BucketPolicy",
             "AWS::SNS::Subscription",
             "AWS::SNS::Topic",
             "AWS::SecretsManager::Secret",
@@ -425,7 +423,6 @@ public class CloudFormationResourceProvisioner {
                 return resource;
             }
             switch (resourceType) {
-                case "AWS::S3::Bucket" -> provisionS3Bucket(resource, properties, engine, region, accountId, stackName);
                 case "AWS::SNS::Topic" -> provisionSnsTopic(resource, properties, engine, region, accountId, stackName);
                 case "AWS::SNS::Subscription" -> provisionSnsSubscription(resource, properties, engine, region);
                 case "AWS::DynamoDB::Table", "AWS::DynamoDB::GlobalTable" ->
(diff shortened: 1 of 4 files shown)


hectorvent:
Thanks — both findings are real, and both are pre-existing rather than introduced here. This PR is a verbatim re-land of the already-approved #2754, whose content never reached `main` (it merged into `refactor/cfn-leaf-provisioners` twenty minutes after that branch had itself been merged into `main`), so it deliberately changes no behaviour.

Checked each against the pre-migration monolith:

- **Buckets recreated on update** — `provisionS3Bucket` generated a name unconditionally too, with no prior-id check. Fixed in #2788, which adds `ProvisionContext.stablePhysicalName` and applies it to the seven provisioners with the same shape.
- **Policy ids regenerated** — `provisionS3BucketPolicy` was `setPhysicalId("bucket-policy-" + UUID...)`, identical. This one #2788 did *not* cover, because it mints a UUID rather than a generated name, so thanks for catching it: now fixed there too, with a test.

On the policy id's *value*, I left it alone on purpose. The sources disagree: the current registry schema gives `primaryIdentifier` as `/properties/Bucket`, while the older schema localstack embeds gives `/properties/Id` as an md5 of the policy document. Changing what `Ref` resolves to on that evidence would be guessing, and keeping the id stable fixes the reported defect under either model. Recorded as a follow-up rather than settled here.

pgermosen (approved):
Approving. Confirmed main really is missing this - no S3CfnProvisioner, and refactor/cfn-leaf-provisioners isn't an ancestor - so the re-land is warranted, not just claimed. The extraction itself is byte-for-byte faithful to the old monolith methods, nothing snuck in.

Also checked both of greptile's findings against the pre-migration code myself: the unconditional bucket-name generation and the policy id's UUID-per-call are both there verbatim already, so agreed these are pre-existing rather than something this PR introduces. Good catch flagging #2788 for the fix.

PR #2795, opened 2026-08-30
fix(cloudformation): update AWS::IAM::ManagedPolicy in place on retry

### LINKED ISSUE #2237: [BUG] cloudformation: stack UPDATE of AWS::IAM::ManagedPolicy fails with EntityAlreadyExists instead of updating the document
## Summary

A CloudFormation stack UPDATE whose template keeps an `AWS::IAM::ManagedPolicy`'s `ManagedPolicyName` fails with `Policy arn:aws:iam::<account>:policy/<name> already exists.` The resource goes CREATE_FAILED and the whole stack rolls back to UPDATE_ROLLBACK_COMPLETE.

`provisionIamManagedPolicy` unconditionally calls `iamService.createPolicy(...)`, even when the resource already has a physical id from a previous provision of the same stack. Real CloudFormation treats `PolicyDocument` as a mutable property and updates the policy in place (publishing a new default version, pruning at IAM's 5-version cap), only replacing the policy when the name changes. `provisionIamRole` already has the equivalent adopt-on-update handling for roles; the managed-policy path never got it.

## Reproduction

1. Create a stack containing an `AWS::IAM::Role` and an `AWS::IAM::ManagedPolicy` with an explicit `ManagedPolicyName` and a `Roles` attachment. Stack reaches CREATE_COMPLETE.
2. Call UpdateStack with the same template, changing only the policy's `PolicyDocument` (e.g. a different `Action`).
3. Expected: UPDATE_COMPLETE, `GetPolicy` shows `DefaultVersionId` v2 with the new document. Actual: `EntityAlreadyExists` from CreatePolicy → CREATE_FAILED → UPDATE_ROLLBACK_COMPLETE.

## Real-world impact

Landing Zone Accelerator's pipeline deploys each stack twice (initial pass + config pass). The second pass updates `AWSAccelerator-OperationsStack-<mgmt-account>` because the `AWSAccelerator-SessionManagerLogging` managed policy's document references resources created in the second-pass Logging stage — and the Operations stage fails on exactly this error, blocking any LZA run past its first pipeline execution.

## Related

Adjacent bug found while fixing: `IamService.createPolicyVersion` derives the next version id from `versions.size() + 1`, so after a `DeletePolicyVersion` the next create reissues an existing id and silently overwrites that version. AWS version ids are monotonic and never reused. This matters once CloudFormation prunes versions at the 5-cap on repeated updates.

Mide69:
Same-stack create-retry on a ManagedPolicy resource failed outright with EntityAlreadyExists instead of applying the template's current PolicyDocument, unlike the adopt-on-update path already used for AWS::IAM::Role. Now the retry is verified against a tracked PolicyId attribute and, once confirmed to be this resource's own policy, the document is applied as a new default version via createPolicyVersion.

Also fixes createPolicyVersion deriving the next version id from versions.size() + 1, which can reissue an id still held by a surviving version once an earlier one is deleted out of order. Replaced with a monotonic counter on IamPolicy, self-healing for policies persisted before the counter existed.

Fixes floci-io/floci#2237

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -76,6 +76,7 @@ import io.github.hectorvent.floci.services.elbv2.model.Rule;
 import io.github.hectorvent.floci.services.elbv2.model.RuleCondition;
 import io.github.hectorvent.floci.services.elbv2.model.TargetGroup;
 import io.github.hectorvent.floci.services.iam.IamService;
+import io.github.hectorvent.floci.services.iam.model.IamPolicy;
 import io.github.hectorvent.floci.services.iam.model.IamRole;
 import io.github.hectorvent.floci.services.kms.KmsService;
 import io.github.hectorvent.floci.services.lambda.LambdaService;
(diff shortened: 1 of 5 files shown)


Mide69:
Closing this as a duplicate. #2717 merged into main after I branched, fixing the same issue with a more thorough adopt-in-place path (also handles version pruning at the 5-version limit and reconciling role attachments removed from templates). Thanks for the review, Greptile.

PR #2796, opened 2026-08-30
fix(cloudformation): provision AWS::Scheduler::ScheduleGroup

### LINKED ISSUE #2396: CloudFormation silently skips AWS::Scheduler::ScheduleGroup resources
On 1.7.0 (image digest sha256:bec9b9f749322444a57fb0f00d3e007ffd49015162361aab84ccecd9b4f5f8ed), a stack containing an `AWS::Scheduler::ScheduleGroup` reaches CREATE_COMPLETE but the group does not exist afterwards: `GetScheduleGroup` returns not-found and `CreateSchedule` into the group fails. The Scheduler API itself supports groups (`CreateScheduleGroup` works when called directly), so this looks CloudFormation-provisioning-specific. Present on 1.5.33 and still present on 1.7.0.

Repro (CreateStack TemplateBody):
```json
{"Resources": {"Group": {"Type": "AWS::Scheduler::ScheduleGroup",
  "Properties": {"Name": "group-repro"}}}}
```
After CREATE_COMPLETE, `GetScheduleGroup group-repro` returns ResourceNotFoundException. Expected: the group exists, matching real CloudFormation.

Mide69:
The resource type had no provisioner registered, so it fell through to the generic stub: the stack reported CREATE_COMPLETE with a fake physical id and no group was ever created in SchedulerService. GetScheduleGroup and any Schedule targeting that group then failed.

Adds SchedulerScheduleGroupCfnProvisioner via the existing CfnResourceProvisioner registry, wired the same way SqsCfnProvisioner and IamRoleCfnProvisioner are, so no changes to
CloudFormationResourceProvisioner's constructor are needed. Handles same-stack create-retry by adopting the group when the physical id already recorded on the resource matches the name that just conflicted.

Fixes floci-io/floci#2396

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist


pgermosen (commented):
@Mide69 the core registration is solid, and I confirmed the request/response shape against the actual CreateScheduleGroup reference — the ConflictException/409 handling and the Arn field both match.

One gap I found checking the CloudFormation resource schema directly: real AWS's Fn::GetAtt for AWS::Scheduler::ScheduleGroup supports four attributes, Arn, CreationDate, LastModificationDate, and State, and this only populates Arn. A template resolving Fn::GetAtt MyGroup.State would get nothing back.

There are also two review findings still open with no reply yet, and I'd like to see both addressed before this merges. First, the same-stack-retry path only ever adds tags on a conflict; it never calls untagScheduleGroup, so a tag removed from the template — or the whole Tags list emptied — stays on the live resource across an update. Second, a tag Key or Value built from Ref or Fn::Sub falls through to raw object-to-text conversion in the tag-parsing loop instead of being resolved through the template engine, so an intrinsic-valued tag gets stored wrong or dropped silently.

hectorvent (approved):
Thank you, and this is really a note that you are unblocked: everything pgermosen raised is fixed at your latest commit, and I checked each one rather than going by the commit messages. All four `Fn::GetAtt` attributes match the registry schema, the retry path reconciles tags both ways, and tag keys and values resolve through the engine.

On Greptile's second `Fn::If` comment, about conditional tags not being applied on create: the symptom is real, but it is not yours to fix here. `resolveNode` collapses any list valued intrinsic to a string, so the shared `ctx.resolveTags` helper returns an empty map for that same input. Your `tagsAreResolvable` guard is the better of the two, because it refuses to read "could not resolve" as "wants no tags". Keep it as it is; I will raise the engine gap separately.

13 provisioner tests and 147 CloudFormation integration tests pass on your branch.

Mide69:
Thanks for the thorough review and the merge. Understood on the Fn::If engine gap, I will leave that to your separate pass. For what it is worth I did end up writing a fix for it locally (resolveNode special-cases Fn::If to recurse into the chosen branch instead of collapsing it to text, so array and object shaped conditionals survive) since I was curious whether it was a small change, but I have not opened anything with it since you already called dibs. Happy to share the diff if it saves you time, otherwise ignore this.

PR #2797, opened 2026-08-30
fix(docker): translate misleading disk-quota error for keyring exhaustion

### LINKED ISSUE #2243: [BUG] lambda: container starts fail with 'crun: join keyctl: Disk quota exceeded' under parallel load
## Summary

Under LZA-scale parallel custom-resource load (dozens of concurrent Lambda container starts), Lambda invocations begin failing with:

```
Failed to start Lambda container: Status 500:
{"cause":"OCI runtime error","message":"crun: join keyctl `<id>`: Disk quota exceeded: OCI runtime error","response":500}
errorType: Lambda.InitError
```

The root cause is not disk space — it is the kernel session-keyring quota. Each container consumes kernel keys for the invoking uid, and the Fedora CoreOS defaults inside a `podman machine` VM are low:

```
kernel.keys.maxkeys = 200
kernel.keys.maxbytes = 20000
```

With ~70+ concurrent Lambda/CodeBuild containers (an LZA SecurityStack deploy fan-out across 14 accounts), the per-uid key count crosses 200 and every subsequent container start fails until containers are reaped.

## Reproduction

1. Run floci on macOS via `podman machine` with default sysctls.
2. Drive a workload that starts ~100 Lambda containers concurrently (LZA `deploy --stage security` across a multi-account org does this reliably).
3. Container starts begin failing with the `crun: join keyctl ... Disk quota exceeded` error above; `/proc/key-users` in the VM shows the invoking uid at its 200-key cap.

## Workaround

```
podman machine ssh "sudo sysctl -w kernel.keys.maxkeys=20000 kernel.keys.maxbytes=2000000"
podman machine ssh "printf 'kernel.keys.maxkeys=20000\nkernel.keys.maxbytes=2000000\n' | sudo tee /etc/sysctl.d/99-floci-keyring.conf"
```

## Suggestion

Two possible improvements, either or both:

1. Docs: a "running at scale" note recommending the sysctl bump for podman-machine (and rootless Linux) hosts.
2. Error mapping: when a container start fails with `join keyctl ... Disk quota exceeded`, surface a self-describing error naming `kernel.keys.maxkeys` instead of the raw OCI runtime message, since "Disk quota exceeded" sends operators down the wrong path (checking disk space).

Mide69:
crun/runc report the kernel session-keyring quota (kernel.keys.maxkeys) as "join keyctl ...: Disk quota exceeded", which reads as an actual disk-space problem. Under high concurrent container counts this is common on podman machine's default Fedora CoreOS sysctls (200 keys per uid), and sends operators looking in the wrong place.

Both container-start call sites in ContainerLifecycleManager now go through a shared helper that recognizes this specific OCI runtime error and rethrows with the actual cause and the sysctl fix, leaving every other DockerException untouched.

Fixes floci-io/floci#2243

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLifecycleManager.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/docker/ContainerLifecycleManager.java
@@ -32,6 +32,7 @@ import java.util.OptionalInt;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 
 /**
  * Manages Docker container lifecycle operations including create, start, stop, and remove.
@@ -41,6 +42,13 @@ import java.util.concurrent.TimeUnit;
 public class ContainerLifecycleManager {
 
     private static final Logger LOG = Logger.getLogger(ContainerLifecycleManager.class);
+
+    // Matches crun/runc's "join keyctl '<id>': Disk quota exceeded" error, which the OCI runtime
+    // reports for the kernel session-keyring quota (kernel.keys.maxkeys), not actual disk space.
+    // Under high concurrent container counts (podman machine's Fedora CoreOS default is 200 keys
+    // per uid) this reads as a disk-space problem and sends operators looking in the wrong place.
+    private static final Pattern KEYRING_QUOTA_PATTERN =
+            Pattern.compile("join keyctl.*disk quota exceeded", Pattern.CASE_INSENSITIVE | Pattern.DOTALL);
 
     private final DockerClient dockerClient;
     private final ImageCacheService imageCacheService;
@@ -143,7 +151,7 @@ public class ContainerLifecycleManager {
      * @return information about the running container including resolved endpoints
      */
     public ContainerInfo startCreated(String containerId, ContainerSpec spec) {
-        dockerClient.startContainerCmd(containerId).exec();
+        startContainer(containerId);
         LOG.infov("Started container {0}", containerId);
 
         if (spec.networkMode() != null && !spec.networkMode().isBlank() && spec.hasPortBindings()) {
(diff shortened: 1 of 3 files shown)


pgermosen (commented):
This is a good fix for the two main start paths, and the tests do a nice job pinning both the translation and that unrelated Docker errors still propagate unchanged (same instance and everything).

One gap: there's a third container start in this same file that isn't routed through startContainer() - the helper container in initSharedVolumeRoot (dockerClient.startContainerCmd(helperId).exec(), around line 412). It's caught as a broad RuntimeException by the caller and just logged as a warning rather than crashing, but the message logged would still be the raw "Disk quota exceeded" one - which is the exact confusion this PR exists to fix. At LZA scale, with a shared volume getting initialized per fresh mount across many concurrent containers, this path seems just as exposed as the other two. Could you route it through startContainer() as well?

pgermosen (approved):
@Mide69 confirmed, this closes the gap. I traced every startContainerCmd call site rather than just trusting the commit message, all three that existed on main (startCreated, initSharedVolumeRoot, adopt) now route through startContainer(), nothing left bypassing it. Good approach on the test too, since the exception on this path gets swallowed by ensureSharedVolume's computeIfAbsent and the translated message was never going to be observable from outside, spying to verify startContainer("helper-id") is actually invoked is the right way to pin the routing. Good to go.

Mide69:
Good catch, fixed. Routed initSharedVolumeRoot's helper container start through startContainer() too, so all three container-start call sites in this class now go through the translation. Made startContainer package-private and added a test that verifies this specific call site invokes it via a spy, since the exception here is caught as a plain RuntimeException by ensureSharedVolume's computeIfAbsent and only logged, never rethrown, so the swallowed exception itself is not observable from the public method.

PR #2798, opened 2026-08-30
fix(opensearch): keep publishing the domain's host port when floci runs in a container

lex00:
Follow-up to #2746, fixing the regression cfranzen reported there: when floci
itself runs in a container, spawned OpenSearch domains no longer published a
host port, so a client on the Docker host could not reach them at all.

The dropped binding copied the Neptune/MemoryDB pattern, but those services
front their backends with floci-internal proxies that keep host clients
connected, and OpenSearch has no proxy: the Docker host-port binding was its
only path in from outside the Docker network. The binding is now published in
both topologies again, exactly as before #2746, while the stored endpoint
keeps the container-IP form that fix introduced.

All 65 OpenSearch tests pass, including a new regression test asserting the
host port is bound with floci containerized and the endpoint stays the
container-IP one.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
@@ -60,6 +60,13 @@ public class OpenSearchDomainManager {
 
         lifecycleManager.removeIfExists(containerName);
 
+        // A restart of an existing domain has just removed the old container, so its
+        // reservation is stale; hand the port back before allocating a fresh one.
+        if (domain.getHostPort() != null) {
+            portAllocator.release(domain.getHostPort());
+            domain.setHostPort(null);
+        }
+
         ContainerBuilder.Builder specBuilder = containerBuilder.newContainer(image)
                 .withName(containerName)
                 .withEnv("discovery.type", "single-node")
@@ -116,6 +123,7 @@ public class OpenSearchDomainManager {
             throw e;
         }
         domain.setContainerId(info.containerId());
+        domain.setHostPort(hostPort);
 
         EndpointInfo endpoint = info.getEndpoint(OPENSEARCH_PORT);
         domain.setEndpoint("http://" + endpoint.host() + ":" + endpoint.port());
@@ -159,6 +167,14 @@ public class OpenSearchDomainManager {
             return;
         }
         lifecycleManager.stopAndRemove(domain.getContainerId(), null);
+        // The container no longer holds the binding, so the reservation must go with
+        // it. Repeated create/delete would otherwise exhaust the configured range.
+        // The keep-running early return above deliberately keeps the reservation:
+        // the surviving container still owns the binding.
+        if (domain.getHostPort() != null) {
+            portAllocator.release(domain.getHostPort());
+            domain.setHostPort(null);
+        }
         LOG.infov("Stopped OpenSearch container for domain {0}", domain.getDomainName());
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
@@ -27,8 +27,6 @@ import java.nio.file.Path;
 public class OpenSearchDomainManager {
 
     private static final Logger LOG = Logger.getLogger(OpenSearchDomainManager.class);
-    /** Sentinel for the in-container path, which publishes no host port to release. */
-    private static final int NO_HOST_PORT = -1;
     private static final int OPENSEARCH_PORT = 9200;
 
     private final ContainerBuilder containerBuilder;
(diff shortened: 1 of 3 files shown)


pgermosen (commented):
The port-leak finding is fixed as of this commit - Domain now tracks hostPort, stopDomain releases it (and correctly keeps it when the container survives a restart instead), and both cases have tests. Good fix.

One finding still open, left inline.

src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
```diff
@@ -79,15 +84,16 @@ public void startDomain(Domain domain) {
         // container-backed services (Neptune, MemoryDB, RDS) already address their
         // backends via the container's resolved IP (EndpointInfo) rather than its
         // Docker name; this container follows the same pattern below.
```
pgermosen, line 86: This one's still open: PortAllocator checks availability with new ServerSocket(port), which binds inside floci's own process - when floci itself is containerized (the exact scenario this PR is about), that's the container's network namespace, not the Docker host's. The host is what -p hostPort:containerPort actually needs free. A port this reports as available could still be taken on the host (the container-create/port-publish then fails), or vice versa.

This is a PortAllocator-level limitation rather than something specific to OpenSearch, but this PR is the first thing that exercises host-port allocation from inside a containerized floci, so it's the first place it can actually bite. Worth fixing here or tracking as a fast follow-up before this ships - your call which, just don't want it to fall through the cracks.

src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
```diff
@@ -79,15 +77,16 @@ public void startDomain(Domain domain) {
...
+        //
+        // Unlike those services, OpenSearch has no floci-internal proxy fronting the
+        // backend, so the Docker host-port binding is the ONLY way a client outside
+        // the Docker network (e.g. on the host, with floci itself containerized)
+        // can reach the domain. Publish it in both topologies — dropping it for the
+        // in-container case cut off host clients entirely (#2746 follow-up).
+        int hostPort = portAllocator.allocate(
+                config.services().opensearch().proxyBasePort(),
+                config.services().opensearch().proxyMaxPort());
+        specBuilder.withPortBinding(OPENSEARCH_PORT, hostPort);
```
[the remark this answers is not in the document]
lex00: The reservation leak is fixed in 60a28965: the domain now records its host port, releases it when its container is removed (teardown and restart both), and keeps the reservation when keep-running shutdown leaves the container holding the binding, covered by two new manager tests. The namespace probe is pre-existing shared infrastructure rather than something this PR touched: PortAllocator.isPortFree binds a ServerSocket in floci's own process and serves EC2, ECR, EKS, MSK and MWAA the same way, and pre-#2746 OpenSearch allocated through that same probe. A namespace-aware PortAllocator is a queued follow-up PR so every consumer gets the fix, not just OpenSearch.

src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
lex00: The namespace probe fix is up as #2805: the shared PortAllocator now consults Docker's view of published host ports, so all its consumers get it, not just OpenSearch.

src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
```diff
@@ -162,6 +167,14 @@ public void stopDomain(Domain domain) {
...
         }
         lifecycleManager.stopAndRemove(domain.getContainerId(), null);
+        // The container no longer holds the binding, so the reservation must go with
+        // it — repeated create/delete would otherwise exhaust the configured range.
+        // The keep-running early return above deliberately keeps the reservation:
+        // the surviving container still owns the binding.
+        if (domain.getHostPort() != null) {
+            portAllocator.release(domain.getHostPort());
+            domain.setHostPort(null);
+        }
```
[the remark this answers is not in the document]
lex00: When stop succeeds and only removal fails, the exited container no longer holds the binding, because Docker frees the host port on stop, so releasing the reservation is correct there. When the container is still running, allocate() on this base consults Docker's published-port snapshot (#2805) and skips it regardless of the reservation. The residual case needs two independent Docker API failures and still fails loudly at container start rather than corrupting state.

src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchDomainManager.java
lex00: Rebased onto main now that #2805 is merged: the host-port allocation here flows through the Docker-aware PortAllocator, so containerized allocation probes Docker's view of published host ports rather than floci's own namespace.

lex00:
The shard-4 red is upstream's, not this branch's: CfnProvisionerFixtureTest fails identically on a clean checkout of main's HEAD (Route53 and SchedulerScheduleGroup provisioners from #2722/#2796 aren't wired into the fixture). Will re-run checks once the fixture fix on main lands.

pgermosen (commented):
Nice, careful fix, traced the restart/failure/stop/keep-running paths by hand and they all hold up: the stale reservation gets released before reallocating on restart, a failed createAndStart always releases the new port now, normal stop releases and clears, and keep-running-on-shutdown correctly keeps the reservation since the container still owns the binding. Also checked that the new hostPort field on Domain doesn't leak into the wire response, DomainStatus is still built field-by-field in the controller so it stays internal-only, same as containerId/volumeId already are.

Both P1 findings from the first pass look genuinely resolved to me too, the port leak here and the namespace check via #2805 now that this is rebased onto it. And I read through the exchange on the removal-failure finding, agree that's closed out.

One small thing: two em-dashes crept into the new comments in OpenSearchDomainManager.java, one near the host-port publishing rationale and one in the stopDomain release comment. We're asking for those to be swapped for colons/commas/periods per AGENTS.md, should be a quick pass.

pgermosen (approved):
Both fixed, confirmed. Nothing else outstanding from my side.

PR #2807, opened 2026-08-31
feat(batch): add update and delete job queue endpoints with corresponding service logic

### LINKED ISSUE #2292: [FEAT] batch: UpdateJobQueue and DeleteJobQueue
## Service

Batch

## API Action / Feature

`UpdateJobQueue` and `DeleteJobQueue`

## AWS Documentation

- https://docs.aws.amazon.com/batch/latest/APIReference/API_UpdateJobQueue.html
- https://docs.aws.amazon.com/batch/latest/APIReference/API_DeleteJobQueue.html

## Why is this needed?

`BatchController` exposes `createjobqueue` and `describejobqueues` but neither
`updatejobqueue` nor `deletejobqueue`, so a job queue can be created but never changed or
removed.

Two consequences:

1. **Teardown is impossible.** `DeleteJobQueue` requires the queue to be `DISABLED` first, so
   the AWS provider's delete path for `aws_batch_job_queue` is `UpdateJobQueue(state=DISABLED)`
   → wait for `VALID`/`DISABLED` → `DeleteJobQueue`. Both calls 404, so `terraform destroy`
   cannot tear down any stack containing a job queue, and re-running `apply` from a clean state
   is blocked by the leftover.
2. **In-place updates fail.** Changing `priority`, `state`, or `compute_environment_order` on
   `aws_batch_job_queue` is an in-place update in the provider schema and issues
   `UpdateJobQueue`.

## Scope

Two endpoints mirroring the existing pair, plus the two matching `BatchService` methods.
`createJobQueue`/`describeJobQueues` (`BatchService:168,206`) are the shape to copy —
`ObjectNode` in, `ObjectNode` out, dispatched from a `@POST @Path("/v1/…")` method on
`BatchController`.

Behaviour worth pinning:

- `DeleteJobQueue` on a queue whose state is `ENABLED` returns `ClientException`, matching AWS.
- `DeleteJobQueue` returns an empty body on success.
- `UpdateJobQueue` returns `jobQueueName` and `jobQueueArn`.
- Deleting an already-deleted queue should be tolerated rather than throwing, so a partially
  completed destroy can be retried.

Filed separately from the compute-environment equivalent so each change stays two endpoints.

- [ ] Yes
- [ ] No

matheusandre1:
## Summary

Closes: #2292 

## Type of change

 - Implementation  `UpdateJobQueue` and `DeleteJobQueue` in service Batch
- [ ] Bug fix (`fix:`)
- [X] New feature (`feat:`)
- [X] Breaking change (`feat!:` or `fix!:`)
- [X] Docs / chore

## Checklist

### DIFF
--- a/docs/services/batch.md
+++ b/docs/services/batch.md
@@ -12,6 +12,8 @@ Floci Batch implements the AWS Batch control plane for local integration tests. 
 | `CreateComputeEnvironment` | `POST /v1/createcomputeenvironment` | Store a local compute environment and return its ARN |
 | `DescribeComputeEnvironments` | `POST /v1/describecomputeenvironments` | Describe all or selected compute environments |
 | `CreateJobQueue` | `POST /v1/createjobqueue` | Store a local job queue attached to compute environments |
+| `UpdateJobQueue` | `POST /v1/updatejobqueue` | Update a job queue's state, priority, and compute environment order |
+| `DeleteJobQueue` | `POST /v1/deletejobqueue` | Delete a `DISABLED` job queue; deleting a missing queue is a no-op |
 | `DescribeJobQueues` | `POST /v1/describejobqueues` | Describe all or selected job queues |
 | `RegisterJobDefinition` | `POST /v1/registerjobdefinition` | Register a revisioned container job definition |
 | `DeregisterJobDefinition` | `POST /v1/deregisterjobdefinition` | Mark a job definition revision inactive |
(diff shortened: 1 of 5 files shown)


pgermosen (approved):
Approving. This closes the linked issue cleanly - the tests cover exactly the behaviors it asked for (update changing state/priority/compute-env-order, delete rejecting an ENABLED queue, repeated delete tolerated as a no-op, unknown queue on update throwing ClientException).

Greptile flagged the queue lookup not being region-scoped and permissive coercion on priority/state/computeEnvironmentOrder as blocking, but both are the same pattern createJobQueue, submitJob, and listJobs already use elsewhere in this file - resolveJobQueueOptional scans the whole store by name regardless of region, and createJobQueue does the identical asInt() coercion with no empty-order check. This PR is consistent with how the rest of the service already works, so not something to hold this one back on. Welcome aboard, and thanks for the contribution.

PR #2811, opened 2026-08-31
fix(build): resolve the JSON schema library from Maven Central

hectorvent:
## Summary

Reported on Slack: a local build fails to resolve
`com.github.everit-org.json-schema:org.everit.json.schema`, a transitive dependency of apicurio.

Reproduced, with one correction to the diagnosis: **the artifact was never on Maven Central**, so it
has not "become" unavailable. `com.github.everit-org.json-schema` is a JitPack coordinate
(`com.github.<user>.<repo>`).

Evidence:

```
$ curl -sI https://repo1.maven.org/maven2/com/github/everit-org/json-schema/\
org.everit.json.schema/1.14.4/org.everit.json.schema-1.14.4.jar
HTTP 404                       # the parent directory is 404 too
$ mvn dependency:get -Dartifact=com.github.everit-org.json-schema:org.everit.json.schema:1.14.4 \
      -Dmaven.repo.local=/tmp/coldrepo
[ERROR] Could not find artifact ... in central (https://repo.maven.apache.org/maven2)
```

`pom.xml` declares no `<repositories>` and there is no `~/.m2/settings.xml`, so Maven only ever
consults Central. Every build that currently works is resolving the jar from a **warm `~/.m2`**,
CI included via its restored Maven cache. That is why this stayed invisible: it breaks only on a
cold clone, which is exactly what a new contributor has.

Apicurio declares the JitPack coordinate in its own POM, so this is upstream's choice rather than
something introduced here.

## The fix

`com.github.erosb:everit-json-schema` is the same project, rehomed by its author to a
Central-published coordinate:

| | JitPack coordinate | Central coordinate |
|---|---|---|
| groupId:artifactId | `com.github.everit-org.json-schema:org.everit.json.schema` | `com.github.erosb:everit-json-schema` |
| classes under `org/everit/json/schema` | 174 | 174 |
| on Maven Central | 404 | 200 |
| latest | 1.14.4 | **1.14.6** |

Same packages, so apicurio's `JsonSchemaCompatibilityChecker` and `JsonSchemaContentValidator` load
unchanged. Excluded from apicurio, declared directly, and `dependency:tree` shows exactly one everit.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

N/A for the wire protocol. Same library, same classes, newer patch version. It backs Glue schema
registry compatibility checks for the `JSON` data format, whose behaviour is unchanged.

## Checklist

`mvn dependency:resolve` against an **empty local repository** now succeeds, which is the check that
matters here.

**The test coverage was the real gap.** `SchemaCompatibilityCheckerTest` had **zero JSON cases** —
only AVRO and PROTOBUF, neither of which touches this library. So the entire JSON branch of
`checkerFor` / `validatorFor` was untested, and a green suite proved nothing about whether the
dependency resolves. Three JSON cases now cover it, and dropping the replacement fails exactly those
three:

```
java.lang.NoClassDefFoundError: org/everit/json/schema/loader/SchemaLoader
```

One assertion is written to observed behaviour rather than assumption: apicurio reports adding an
*optional* JSON property as `OBJECT_TYPE_PROPERTY_SCHEMAS_NARROWED`, i.e. incompatible — the
opposite of the AVRO case — because data carrying that key with a different type validated before
and no longer does. I had asserted the intuitive answer first and it was wrong.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,6 +18,7 @@
         <quarkus.platform.version>3.37.4</quarkus.platform.version>
         <jackson.version>2.22.1</jackson.version>
         <apicurio.version>2.6.13.Final</apicurio.version>
+        <everit.json.schema.version>1.14.6</everit.json.schema.version>
         <wire.version>6.4.5</wire.version>
     </properties>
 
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Verified: reverting just the Central-coordinate dependency (keeping the exclusion) reproduces `NoClassDefFoundError: org/everit/json/schema/loader/SchemaLoader` on exactly the 3 new tests that exercise apicurio's JSON diff engine — the new coverage is load-bearing, not incidental. Merges clean against current main; merged tree compiles and passes. CODEOWNERS on pom.xml is the author themself.

Issue #2793, opened 2026-08-30
glue: GetDatabase response omits CatalogId


pgermosen:
Floci's Glue `GetDatabase`/`GetDatabases` response never sets `CatalogId`. `terraform-provider-aws` builds the `aws_glue_catalog_database` resource ID as `catalog-id:database-name`; with an empty `CatalogId` the ID comes out malformed:

```
Error: unexpected format for ID (:floci_compat_db), expected catalog-id:database-name
```

Confirmed still missing on `main` — no `CatalogId` anywhere under `services/glue`. Described as "a small, self-contained fix" in the originating review — the database response just needs to populate `CatalogId` with the account ID, matching real AWS.

Blocks the Terraform/OpenTofu compatibility suites from including `aws_glue_catalog_database` (removed from both in #1834 to unblock the suite; should be re-added once this lands).

Origin: https://github.com/floci-io/floci/pull/1834#issuecomment (abanna/hectorvent review, 2026-08-05 / 2026-08-15).

Mide69:
Would like to take this on, please assign it to me.

trh-ds:
I had a patch ready for this before noticing it was assigned — opened it as a PR so the work is on the table, but happy to close it if @Mide69 is mid-work on their own.

hectorvent:
resolved by #2814

PR #2814, opened 2026-08-31
fix(glue): populate CatalogId on GetDatabase and GetDatabases

trh-ds:
## Problem

Glue `GetDatabase` / `GetDatabases` never set `CatalogId`. `terraform-provider-aws` builds the `aws_glue_catalog_database` resource id as `catalog-id:database-name`, so the id came out malformed:

```
Error: unexpected format for ID (:floci_compat_db), expected catalog-id:database-name
```

Refs #2793.

## Change

`GlueService.getDatabase` / `getDatabases` populate `CatalogId` from `regionResolver.getAccountId()` — the same convention `LakeFormationService` already uses for its catalog id default.

Resolved per request rather than persisted with the database, so it always reflects the calling account and existing `databases.json` stores need no migration.

## Test

`GlueDatabaseCatalogIdIntegrationTest` asserts `CatalogId` on the wire response of both `GetDatabase` and `GetDatabases`. `Database.catalogId` is set nowhere else, so the test fails without the fix.

```
Tests run: 178, Failures: 0, Errors: 0, Skipped: 0   # -Dtest=Glue*
```

Re-adding `aws_glue_catalog_database` to the Terraform/OpenTofu compatibility suites (removed in #1834) is left as a follow-up.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/glue/GlueService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/glue/GlueService.java
@@ -116,6 +116,7 @@ public class GlueService {
             throw new AwsException("AlreadyExistsException", "Database already exists: " + database.getName(), 400);
         }
         database.setName(databaseName);
+        database.setCatalogId(regionResolver.getAccountId());
         databaseStore.put(databaseName, database);
         if (tags != null && !tags.isEmpty()) {
             resourceGroupsTaggingService.tagResources(List.of(databaseArn(region, databaseName)), tags, region);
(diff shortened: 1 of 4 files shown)


trh-ds:
Pushed a fixup for the review note about the read path mutating stored state: CatalogId is now set when the database is created or updated, and reads only backfill rows persisted before the field existed. That backfill is stable because the store is namespaced by account, so the reader is always the owning account -- the same migrate-on-read `AccountAwareStorageBackend` already does for un-prefixed keys.

Added unit coverage for both: create persists the id, and a legacy row with no CatalogId heals on read. `-Dtest=Glue*` is green at 180 tests.

hectorvent (approved):
Thank you for the fix, and for tracing it to what terraform-provider-aws actually parses: the `catalog-id:database-name` id builder is exactly where an absent `CatalogId` bites.

Checked against the Glue model: `Database.CatalogId` is a plain optional member of `Database`, so it flows into `GetDatabase` and `GetDatabases` unchanged (https://github.com/boto/botocore/blob/develop/botocore/data/glue/2017-03-31/service-2.json). Taking it from `regionResolver.getAccountId()` gives the request account, and since the database store is account-scoped the legacy backfill on read is stable, which settles the earlier bot concern.

**(follow-up, separate PR)** The model also carries `CatalogId` on `Table`, `Partition` and `UserDefinedFunction`, and Floci's `services/glue/model/Table.java` has no such field yet. Terraform's `aws_glue_catalog_table` read passes `table.CatalogId` into `GetPartitionIndexes`, so the same treatment there would close the table half of #2793.

**(follow-up, separate PR)** The description still says the id is resolved per request rather than persisted; after 772769e it is written at create and update and only backfilled for old rows, so a quick edit keeps the two in sync.

No blockers from my side.

PR #2816, opened 2026-08-31
fix(apigateway): report RestApi as available

### LINKED ISSUE #2804: apigateway: RestApi response omits apiStatus, hangs terraform-provider-aws's create waiter
`terraform-provider-aws`'s `resourceRestAPICreate` for `aws_api_gateway_rest_api` waits for the REST API's `apiStatus` field to transition to `AVAILABLE` after creation:

```go
func waitRestAPIAvailable(ctx context.Context, conn *apigateway.Client, id string, timeout time.Duration) error {
    stateConf := &retry.StateChangeConf{
        Pending: enum.Slice(types.ApiStatusPending, types.ApiStatusUpdating),
        Target:  enum.Slice(types.ApiStatusAvailable),
        ...
```

`apiStatus` is a real, current field on the API Gateway v1 `RestApi` shape (confirmed in botocore's `apigateway/2015-07-09/service-2.json`). Floci's `CreateRestApi`/`GetRestApi` responses never set it at all, so the provider's waiter sees an empty string forever and times out:

```
Error: waiting for API Gateway REST API (...) to become available: unexpected state '', wanted target 'AVAILABLE'
```

Confirmed still missing on `main` — zero matches for `apiStatus`/`ApiStatus` under `services/apigateway`.

Blocks the Terraform/OpenTofu compatibility suites from including `aws_api_gateway_rest_api` (removed from both in #1834 to unblock the suite — every `terraform apply` deterministically failed on it; should be re-added once this lands). Fix direction: always return `apiStatus: "AVAILABLE"` on `RestApi` create/read responses.

7487:
## Summary

- return AWS `apiStatus: "AVAILABLE"` from the shared API Gateway v1 RestApi serializer
- cover both CreateRestApi and GetRestApi response shapes

## Root cause

Floci's synchronous RestApi responses omitted the current AWS field Terraform's waiter reads, so the state stayed empty until timeout. The shared serializer is used by create, read, list, import, put, and update paths, so one derived field fixes every response path without persisting artificial state.

## Tests

- `./mvnw -B -Dtest=ApiGatewayIntegrationTest test`
- `./mvnw -B -DskipTests compile`
- CI-equivalent four-shard full suite: 1,085 test classes, 14,645 tests, 0 failures, 0 errors
- `./mvnw -B -DskipTests package`
- `git diff --check`

Closes #2804

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
@@ -1908,6 +1908,7 @@ public class ApiGatewayController {
         node.put("name", api.getName());
         if (api.getDescription() != null) node.put("description", api.getDescription());
         node.put("createdDate", api.getCreatedDate());
+        node.put("apiStatus", "AVAILABLE");
         if (api.getTags() != null && !api.getTags().isEmpty()) {
             ObjectNode tagsNode = objectMapper.createObjectNode();
             api.getTags().forEach(tagsNode::put);
--- a/src/test/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayIntegrationTest.java
@@ -35,6 +35,7 @@ class ApiGatewayIntegrationTest {
                 .body("id", notNullValue())
                 .body("name", equalTo("test-api"))
                 .body("description", equalTo("Integration test API"))
+                .body("apiStatus", equalTo("AVAILABLE"))
                 .extract().path("id");
     }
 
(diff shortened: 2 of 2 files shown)


pgermosen (approved):
@7487 welcome, and thanks for this — clean, minimal fix. I checked the underlying claim myself rather than taking the issue's word for it: `apiStatus` is genuinely documented on the `RestApi` shape (`UPDATING | AVAILABLE | PENDING | FAILED`), so this closes a real gap, not an undocumented quirk. Since Floci creates RestApis synchronously with no pending/updating window to model, always returning `AVAILABLE` is the right call.

I also traced that the fix actually reaches every response path rather than just the two you added assertions for — `toApiNode` is the shared serializer for create, get, list, import, and both update paths, so the one-line addition propagates to all six. Testing create and get is proportionate; the other four go through the identical code, so there's nothing distinct left to prove there.

Approving.

PR #2823, opened 2026-08-31
fix(stepfunctions): let a mocked response with no attempts start execution

### LINKED ISSUE #2523: [BUG] Step Functions (SFN): StartExecution validates mocked responses the execution never uses

### Current Behavior

`StartExecution` validates every entry in `MockedResponses`, including entries the execution never reaches, and rejects the whole request if any one of them is malformed:

> An error occurred (ValidationException) when calling the StartExecution operation: Mocked response 'X' must define at least one attempt entry

Step Functions Local resolves a mocked response only when the state that names it is actually entered, so a mock file with an unused or empty entry runs fine.

This matters for tests that build mock data programmatically. A common shape is "mock this state with one response per item in a collection":

```ruby
Sfn::MockMacros.gateway_response(
  completed_subject_directories.map { |d| { response: { id: d['subject_directory_id'] } } }
)
```

When the collection is empty — the scenario under test being "nothing was completed", where that state is never entered — the macro produces `{}`. Against Step Functions Local this is correct and the execution simply never touches it. Against Floci the whole execution is rejected before it starts.

### Expected Behavior

Validate a mocked response when it is resolved, not up front. An entry that no reached state names should not fail the execution.

If eager validation is deliberate, it would help to scope it to entries actually referenced by the selected test case, so that unreferenced ones are ignored.

Docker image built from `main` (`257e004`), `docker/Dockerfile`, `SFN_MOCK_CONFIG` set.

### Steps To Reproduce

Mock config where `Unused` is empty and the `Skip` state is not reached on this input:

```json
{
  "StateMachines": {
    "Test": { "TestCases": { "Case1": { "Skip": "Unused", "Run": "Ok" } } }
  },
  "MockedResponses": {
    "Unused": {},
    "Ok": { "0": { "Return": { "ResponseBody": { "ok": true } } } }
  }
}
```

`StartExecution` on `<arn>#Case1` is rejected, even though `Skip` is never entered.

### Anything else?

This accounts for 25 failures in a 713-example suite we are porting from `amazon/aws-stepfunctions-local`. Related to #2452, which documents the strict-validation behaviour as an intentional deviation — this report is about it applying to entries that are never used, rather than about the strictness itself.

exoego:
## Summary

Closes #2523

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against amazon/aws-stepfunctions-local:2.0.0.

A MockedResponses entry with no attempt keys was rejected at StartExecution, so a generated mock file failed the whole execution when the collection it was built from was empty. Step Functions Local starts the execution and fails it with States.Runtime only when the state that names the entry is entered.

An empty entry is a lookup that finds nothing, not a malformed entry. Floci already accepts a mock that defines attempt 1 but not attempt 0 and fails the same way at run time. Every other malformed entry keeps its 400 at StartExecution, which stays stricter than Step Functions Local.

## Checklist

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -398,6 +398,11 @@ entry, `Return` and `Throw` together, `Throw` without `Error`) as a structured 4
 at `StartExecution`. Step Functions Local instead returns a plain HTTP 500 for most of
 these and starts the execution only to fail it with `States.Runtime` for the last two.
 
+A mocked response with no attempt entries (`{}`) is not rejected. As in Step Functions
+Local, the execution starts and fails with `States.Runtime` only if the state that names
+it is entered. This keeps a generated mock file usable when the collection it was built
+from is empty and the state is never reached.
+
 ## Configuration
 
 | Variable | Default | Description |
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/SfnMockLoader.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/SfnMockLoader.java
@@ -121,14 +121,15 @@ public class SfnMockLoader {
         return root;
     }
 
+    /**
+     * An entry with no attempts stays empty instead of being rejected. Step Functions Local
+     * treats it as a lookup that finds nothing, so the execution starts and fails with
+     * {@code States.Runtime} only if the state that names it is entered.
+     */
     private List<MockedResponseStep> parseSteps(String responseKey, JsonNode responseNode) {
         var steps = new ArrayList<MockedResponseStep>();
         responseNode.fields().forEachRemaining(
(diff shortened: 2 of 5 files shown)


pgermosen (approved):
@exoego welcome, and thanks for this — clean fix. I traced why deferring the rejection is safe rather than just taking the description at face value: the runtime path that fails a state with `States.Runtime` when no attempt matches already exists, so this isn't new failure-handling logic, it's just letting an execution start instead of rejecting it upfront for a mock that might never even be reached. Both new integration tests prove that end to end against the real executor — one for the state never being entered, one for `States.Runtime` when it is — rather than just asserting it in the description.

Approving.

Issue #2822, opened 2026-08-31
[BUG] Lambda PublishVersion always creates a version and ignores the CodeSha256 precondition


bsaurusrex:
## Service

Lambda (versions)

## AWS API Action

`PublishVersion` — de-duplication of an unchanged publish, and the `CodeSha256` precondition

## Expected behavior

Two documented behaviours, both measured against the live service (`lambda.ap-southeast-1.amazonaws.com`):

**1. An unchanged publish does not create a version.** The API reference states: "AWS Lambda doesn't publish a version if the function's configuration and code haven't changed since the last version." Measured, on a function whose latest version is `2` and whose `$LATEST` is unchanged since:

```
aws lambda publish-version --function-name fn --query Version --output text
# 2      <-- the existing version, no new one created
```

**2. `CodeSha256` is a precondition.** "Only publish a version if the hash value matches the value that's specified." Measured with a hash that does not match:

```
aws lambda publish-version --function-name fn --code-sha256 "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
# An error occurred (InvalidParameterValueException) when calling the PublishVersion operation:
# CodeSHA256 (AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=) is different from current CodeSHA256 in $LATEST
```

## Actual behavior

Floci publishes unconditionally and ignores `CodeSha256`.

```
versions before:                              $LATEST  1
publish-version   (nothing changed)        -> 2        (AWS: returns the existing version)
publish-version   (nothing changed again)  -> 3        (AWS: returns the existing version)
versions after:                               $LATEST  1  3  2

publish-version --code-sha256 <wrong hash> -> 4        (AWS: InvalidParameterValueException)
```

So repeated publishes accumulate identical versions, and the precondition that exists to stop a caller publishing code it did not intend is silently ignored — a publish that should have been rejected succeeds and produces a version.

(Incidentally, `list-versions-by-function` returned `$LATEST 1 3 2` — unordered. AWS returns them in ascending version order. Minor, and not the substance of this report.)

## Reproduction

```bash
awsl() { aws --endpoint-url http://localhost:4566 --region ap-southeast-1 "$@"; }

printf "def handler(e, c):\n    return {'body': 'v1'}\n" > handler.py && zip -q fn.zip handler.py
awsl lambda create-function --function-name pv-probe --runtime python3.13 \
[...]

PR #2825, opened 2026-08-31
fix(lambda): honour Publish on CreateFunction and UpdateFunctionCode

### LINKED ISSUE #2820: [BUG] Lambda CreateFunction and UpdateFunctionCode ignore Publish, so no version is ever created
## Service

Lambda (versions)

## AWS API Action

`CreateFunction` and `UpdateFunctionCode` — the `Publish` parameter

## Expected behavior

`Publish` is documented on both actions — "Set to true to publish the first version of the function during creation" (`CreateFunction`) and "Set to true to publish a new version of the function after updating the code" (`UpdateFunctionCode`). The response's `Version` is the newly published version, not `$LATEST`.

Measured against the live service (`lambda.ap-southeast-1.amazonaws.com`):

| Call | Live AWS |
| --- | --- |
| `create-function --publish` → `Version` | `1` |
| `list-versions-by-function` after it | `$LATEST`, `1` |
| `get-function` (unqualified) → `Version` | `$LATEST` |
| `update-function-code --publish` → `Version` | `2` |
| `list-versions-by-function` after it | `$LATEST`, `1`, `2` |
| `update-function-code` (no `--publish`) → `Version` | `$LATEST`, version list unchanged |

Note `update-function-configuration` has no `Publish` parameter at all — the CLI rejects `--publish` with `Unknown options: --publish` — so only these two actions are affected.

## Actual behavior

Floci never reads the flag. `grep -rn '"Publish"' src/main/java/io/github/hectorvent/floci/services/lambda/` returns nothing; `publishVersion` is implemented and correct, but no create/update path calls it.

```
create-function --publish       -> Version: "$LATEST"     (AWS: "1")
list-versions-by-function       -> ["$LATEST"]            (AWS: ["$LATEST", "1"])
update-function-code --publish  -> Version: "$LATEST"     (AWS: "2")
publish-version (explicit)      -> 1                       <-- works
```

So no version is ever created unless the caller makes a separate `PublishVersion` call.

The practical effect is on Terraform. `aws_lambda_function` with `publish = true` reads `Version` and `qualified_arn` straight from the create/update response, so both come back as `$LATEST` instead of a numeric version. That breaks anything requiring a published version — Lambda@Edge associations require a numeric version and cannot use `$LATEST` — and leaves `version`/`qualified_arn` permanently wrong in state for every function that sets `publish = true`.

## Reproduction

```bash

awsl lambda create-function --function-name pub-probe --runtime python3.13 \
  --role arn:aws:iam::000000000000:role/r --handler handler.handler \
  --zip-file fileb://fn.zip --publish --query Version --output text
# $LATEST      <-- AWS returns 1

awsl lambda list-versions-by-function --function-name pub-probe \
  --query 'Versions[].Version' --output text
# $LATEST      <-- AWS returns $LATEST 1

printf "def handler(e, c):\n    return {'body': 'v2'}\n" > handler.py && zip -q fn.zip handler.py
awsl lambda update-function-code --function-name pub-probe \
  --zip-file fileb://fn.zip --publish --query Version --output text
# $LATEST      <-- AWS returns 2
```

## Environment

- Java SDK version (if applicable): n/a — AWS CLI v2, and terraform-provider-aws

## Note

I have a fix for this and will open a PR. Filing separately from the other version-lifecycle divergences found in the same pass so they can be triaged apart: #2819 (`DeleteFunction` ignores `Qualifier` and deletes the whole function), plus `Qualifier` being ignored by `GetFunction` and `PublishVersion` neither de-duplicating an unchanged publish nor honouring `CodeSha256`.

bsaurusrex:
## Summary

`CreateFunction` and `UpdateFunctionCode` never read `Publish`. No version was created unless the caller made a separate `PublishVersion` call, and both actions reported `Version` as `$LATEST`.

Both now publish and report the new version.

Closes #2820

## Type of change

## Why it matters beyond the flag

The Terraform provider reads `Version` and `qualified_arn` straight from these responses, so every `aws_lambda_function` with `publish = true` recorded `$LATEST`. Lambda@Edge associations require a numeric version and cannot use `$LATEST`, so that shape of stack could not be modelled against Floci at all.

## AWS Compatibility

API reference: [CreateFunction](https://docs.aws.amazon.com/lambda/latest/api/API_CreateFunction.html), [UpdateFunctionCode](https://docs.aws.amazon.com/lambda/latest/api/API_UpdateFunctionCode.html).

Measured against the live service (`lambda.ap-southeast-1.amazonaws.com`) on a throwaway function, created and deleted for the probe:

| Call | Live AWS `Version` | Live AWS `FunctionArn` | Floci before | Floci after |
| --- | --- | --- | --- | --- |
| `create-function --publish` | `1` | **unqualified** — `...:function:fn` | `$LATEST`, no version created | matches AWS |
| `update-function-code --publish` | `2` | **qualified** — `...:function:fn:2` | `$LATEST`, no version created | matches AWS |
| `publish-version` (explicit) | `3` | qualified — `...:function:fn:3` | already correct | unchanged |
| `create-function` (no `--publish`) | `$LATEST` | unqualified | already correct | unchanged |
| `update-function-code` (no `--publish`) | `$LATEST` | unqualified, version list unchanged | already correct | unchanged |

**The two publishing calls disagree about the ARN**, and that is the reason this PR is shaped the way it is. `CreateFunction` reports the version it just published but answers with the *unqualified* ARN, while `UpdateFunctionCode` returns the *qualified* one, as `PublishVersion` does. The reference gives no hint of the distinction — making them consistent is the obvious implementation, and it is wrong. Getting it backwards would store the wrong `arn` in Terraform state for every function that publishes.

`UpdateFunctionConfiguration` has no `Publish` parameter in the AWS API — the CLI rejects `--publish` with `Unknown options: --publish` — so it is deliberately untouched.

## Implementation

The service delegates to the existing `publishVersion`, which already snapshots code and configuration correctly; nothing was reaching it from these paths. The `CreateFunction` ARN is un-qualified in the controller, where response shaping belongs — the published snapshot is stored by reference, so mutating its ARN to suit one response would corrupt the stored version record.

## Interaction with #2822

Once `Publish` is honoured, `PublishVersion`'s missing de-duplication starts to matter more: the live service does not publish when code and configuration are unchanged, while Floci mints a fresh version every call. A no-op Terraform apply would therefore create a new version each run. That is #2822, fixed separately.

## Testing

- `LambdaPublishFlagIntegrationTest` — 9 tests: both publishing calls' version *and* ARN shape, the no-publish paths creating nothing, `Publish` omitted entirely behaving as false, and a published version keeping the `CodeSha256` it was published from after `$LATEST` moves on. Reverting the service change fails 4 of the 9.
- SDK coverage in `compatibility-tests/sdk-test-node/tests/lambda.test.ts` — the same behaviour through `@aws-sdk/client-lambda`, per AGENTS.md's preference for SDK clients over raw HTTP on the management plane.
- Full suite, on the pre-rebase tip: **14489 tests, 0 failures, 0 errors, 12 skipped** (sharded with CI's own `.github/ci/partition.sh`).  The rebase onto `main` added only upstream commits (#2812, #2815) plus a one-paragraph resolution in `docs/services/lambda.md`; this branch's Java is byte-identical across it, and the targeted test, `docs-check` and the compatibility suite were all re-run on the rebased tip.
- `make docs-check`: clean.
- Compatibility suite (`sdk-test-node`, run CI-style inside the container network with `--dns` pointed at Floci), re-run on the rebased tip: **461/463**, with `lambda.test.ts` **17/17**. The two failures are the WebSocket chat-broadcast suites (`apigatewayv2-websocket-dataplane`, `apigatewayv2-websocket-tls`); a control run of those two files against the released `floci/floci:1.7.0` image reproduces them identically, so they are pre-existing and unrelated.

## Rebased on `main`

#2812 has merged and this branch is rebased onto it. The only conflict was the coverage
paragraph in `docs/services/lambda.md`; both sides are kept. The Java change is untouched
by the rebase, and `LambdaPublishFlagIntegrationTest` is **9 tests, 0 failures** on the
rebased tip.

It still shares `docs/services/lambda.md` and `lambda.test.ts` with #2824, so whichever of
the two merges second will want the same one-paragraph resolution.

## Checklist

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/compatibility-tests/sdk-test-node/tests/lambda.test.ts
+++ b/compatibility-tests/sdk-test-node/tests/lambda.test.ts
@@ -17,6 +17,8 @@ import {
   UpdateAliasCommand,
   DeleteAliasCommand,
   PublishVersionCommand,
+  ListVersionsByFunctionCommand,
+  UpdateFunctionCodeCommand,
 } from '@aws-sdk/client-lambda';
 import { makeClient, uniqueName, ACCOUNT, buildMinimalZip } from './setup';
 
@@ -183,5 +185,59 @@ describe('Lambda ImageConfig.WorkingDirectory', () => {
     } finally {
       await lambda.send(new DeleteFunctionCommand({ FunctionName: fnName })).catch(() => {});
     }
+  });
+});
+
+describe('Lambda Publish flag', () => {
+  let lambda: LambdaClient;
+  let fnName: string;
+
+  beforeAll(() => {
+    lambda = makeClient(LambdaClient);
+    fnName = `test-publish-${uniqueName()}`;
+  });
+
+  afterAll(async () => {
+    await lambda.send(new DeleteFunctionCommand({ FunctionName: fnName })).catch(() => {});
+  });
+
+  it('should publish version 1 on create and keep the ARN unqualified', async () => {
+    const response = await lambda.send(new CreateFunctionCommand({
+      FunctionName: fnName,
+      Runtime: 'nodejs20.x',
+      Role: `arn:aws:iam::${ACCOUNT}:role/lambda-role`,
+      Handler: 'index.handler',
+      Code: { ZipFile: buildMinimalZip('index.js', Buffer.from('exports.handler = async () => 1;')) },
+      Publish: true,
+    }));
+
+    expect(response.Version).toBe('1');
+    expect(response.FunctionArn).toMatch(new RegExp(`:function:${fnName}$`));
+
+    const listed = await lambda.send(new ListVersionsByFunctionCommand({ FunctionName: fnName }));
+    expect((listed.Versions ?? []).map((v) => v.Version).sort()).toEqual(['$LATEST', '1']);
+  });
+
+  it('should publish the next version on code update and return a qualified ARN', async () => {
+    const response = await lambda.send(new UpdateFunctionCodeCommand({
+      FunctionName: fnName,
+      ZipFile: buildMinimalZip('index.js', Buffer.from('exports.handler = async () => 2;')),
+      Publish: true,
+    }));
+
+    expect(response.Version).toBe('2');
+    expect(response.FunctionArn).toMatch(new RegExp(`:function:${fnName}:2$`));
+  });
+
+  it('should create no version when Publish is not set', async () => {
+    const response = await lambda.send(new UpdateFunctionCodeCommand({
+      FunctionName: fnName,
+      ZipFile: buildMinimalZip('index.js', Buffer.from('exports.handler = async () => 3;')),
+    }));
+
+    expect(response.Version).toBe('$LATEST');
+
+    const listed = await lambda.send(new ListVersionsByFunctionCommand({ FunctionName: fnName }));
+    expect((listed.Versions ?? []).map((v) => v.Version).sort()).toEqual(['$LATEST', '1', '2']);
   });
 });
--- a/docs/services/lambda.md
+++ b/docs/services/lambda.md
@@ -166,6 +166,13 @@ services:
 
 Function URLs are also reachable directly on `/{proxy:.*}` under the Lambda URL controller, which routes the request into the normal `Invoke` path.
 
+**Versions:** `CreateFunction` and `UpdateFunctionCode` honour `Publish`, publishing a version
+and reporting it in the response's `Version`. The two differ in the ARN they return, matching the
+live service: `CreateFunction` keeps the unqualified `FunctionArn` while `UpdateFunctionCode`
+returns the qualified one (`...:function:name:2`), as `PublishVersion` does. Without `Publish`
+both answer for `$LATEST` and create nothing. `UpdateFunctionConfiguration` has no `Publish`
+parameter in the AWS API and none here.
+
 **Layers:** `PublishLayerVersion`, `GetLayerVersion`, `GetLayerVersionByArn`, `ListLayerVersions`,
 `ListLayers`, and `DeleteLayerVersion` are implemented, with real local storage under
 `{lambda.codePath}/layers/{name}/{version}`. `CreateFunction`/`UpdateFunctionConfiguration`
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -70,8 +70,10 @@ public class LambdaController {
             @SuppressWarnings("unchecked")
             Map<String, Object> request = objectMapper.readValue(body, Map.class);
             LambdaFunction fn = lambdaService.createFunction(region, request);
+            Map<String, Object> configuration = buildFunctionConfiguration(fn);
+            unqualifyArn(configuration);
             return Response.status(201)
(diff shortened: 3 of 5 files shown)


pgermosen (commented):
@bsaurusrex the logic here checks out — I traced both paths rather than going on the writeup. `createFunction` and `updateFunctionCode` both delegate to the existing `publishVersion` when `Publish` is true, so the version record is snapshotted correctly, and the asymmetric ARN shaping (unqualified on create, qualified on update, matching the real service's own inconsistency) is handled entirely at the controller layer without touching the stored version's own ARN — `unqualifyArn` only strips the qualifier on the `CreateFunction` response, while `UpdateFunctionCode` and `PublishVersion` both return the underlying record untouched. Tests match: unqualified-on-create, qualified-on-update, nothing created when `Publish` is absent or false, and a published version keeping its own `CodeSha256` after `$LATEST` moves on.

This one's showing conflicting against main now, most likely `docs/services/lambda.md` where #2812 landed adjacent lines. Needs a downmerge before it can go in — nothing else outstanding once that's resolved.

bsaurusrex:
Moved to draft while I rebase onto `main` — #2812 landed and touched the same layer methods this branch changes, so the branch is currently conflicted. Not asking for review until it is green again.

No behaviour change intended in the rebase; the resolution keeps both sides (the `find=LayerVersion` dispatch from #2812 alongside the query-parameter handling here). I will re-run the targeted tests and the full suite before marking it ready.

bsaurusrex:
Rebased onto `main` and back in review.

The conflict was `docs/services/lambda.md` only (the coverage paragraph #2812 also touched); both sides are kept. No Java changed in the rebase.

Re-verified on the rebased tip: the branch's own integration test, `make docs-check`, and the full `sdk-test-node` compatibility suite — the only failures are the two WebSocket chat-broadcast tests, and a control run of those two files against the released `floci/floci:1.7.0` image reproduces them identically, so they are pre-existing. Numbers are in the description.

bsaurusrex:
Thanks for tracing it rather than taking the writeup on trust — that is the review I would want on this.

On the conflict: it is already resolved. The rebase onto #2812 landed at `54c85d3`, and the `docs/services/lambda.md` collision was a keep-both — this PR's paragraph plus #2812's `GetLayerVersionByArn` in the `**Layers:**` line, which is present on the branch now. GitHub reports the PR mergeable, so I think you were looking at it in the window before that push.

Re-run on the rebased tip, not carried over from before it: the branch's own integration tests, `make docs-check` (clean), and the compatibility suite (461/463, the shortfall being the pre-existing WebSocket chat-broadcast cases that also fail on released 1.7.0).

`main` has moved 4 commits since that rebase; I checked and they merge cleanly, so I have left the branch where it is rather than churn the tip. Say the word if you would rather have it freshly rebased before merging.

pgermosen (approved):
@bsaurusrex confirmed the rebase is clean — diffed it against what I'd already traced and the only change is `docs/services/lambda.md` picking up #2812's merged content; every line of the actual Java is unchanged from what I verified earlier. Nothing outstanding.

Approving.

PR #2827, opened 2026-08-31
fix(sqs): honor queue-level ReceiveMessageWaitTimeSeconds on long poll

### LINKED ISSUE #2826: [BUG] SQS: queue-level ReceiveMessageWaitTimeSeconds is stored but never applied — empty ReceiveMessage without WaitTimeSeconds returns immediately
## Service

SQS

## Summary

Queue-level `ReceiveMessageWaitTimeSeconds` is stored on queue creation but never applied: an empty `ReceiveMessage` call that omits `WaitTimeSeconds` returns immediately instead of blocking for the queue's configured wait time. Only the request-level `WaitTimeSeconds` parameter works.

This is the same defect class as #1567 (queue-level `DelaySeconds`), which was fixed by switching the handler→service signature from a primitive `int` (default `0`) to `Integer` so an omitted value can fall back to the queue attribute. The same treatment appears to be missing for the receive wait time.

## AWS API Action

SQS `ReceiveMessage` / `CreateQueue`

## Expected behavior

Per AWS SQS semantics, when `ReceiveMessage` omits `WaitTimeSeconds`, the value of the queue's `ReceiveMessageWaitTimeSeconds` attribute is used (queue default `0` when unset).

> From the AWS docs: "If you do not set WaitTimeSeconds in the request, theReceiveMessageWaitTimeSeconds queue attribute is used."

## Actual behavior

The stored queue attribute is ignored. A queue created with `"ReceiveMessageWaitTimeSeconds": "5"` — attribute confirmed stored via `GetQueueAttributes` — returns an empty receive in ~0.3s instead of blocking ~5s. Request-level `WaitTimeSeconds` is honored correctly (blocks the full duration).

## Root cause

Verified against current `main`:

- `SqsJsonHandler.java` line 246: `int waitTimeSeconds = request.path("WaitTimeSeconds").asInt(0);`
- `SqsQueryHandler.java` line 265: `int waitTimeSeconds = getIntParam(params, "WaitTimeSeconds", 0);`

Both handlers parse the parameter as a primitive `int` defaulting to `0`, and `SqsService.receiveMessage` computes `maxWait = waitTimeSeconds * 1000L` with an immediate return when `maxWait <= 0`. The queue's `ReceiveMessageWaitTimeSeconds` attribute is never read anywhere — the string `ReceiveMessageWaitTimeSeconds` does not appear in the codebase, and the compatibility tests only exercise request-level `waitTimeSeconds`.

## Impact

Any consumer that correctly defers to queue-level configuration (omits `WaitTimeSeconds`, as the AWS SDKs and long-polling best practices recommend) ends up busy-polling: an "idle" consumer continuously issues empty receives. Measured on our side, a single idle consumer burns ~72% of one core server-side in floci around the clock and starves other data-plane traffic (DynamoDB/Cognito handled by the same process).

## Reproduction

Environment: `floci/floci:1.7.0-compat` (also reproduced on a freshly started instance with no init scripts).

```bash
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=eu-north-1

# Create queue with the wait attribute
url=$(aws --endpoint-url http://localhost:4566 sqs create-queue --queue-name repro.fifo \
  --attributes '{"FifoQueue":"true","ReceiveMessageWaitTimeSeconds":"5"}' \

# Attribute is stored correctly:
aws --endpoint-url http://localhost:4566 sqs get-queue-attributes --queue-url "$url" \
  --attribute-names ReceiveMessageWaitTimeSeconds
# { "Attributes": { "ReceiveMessageWaitTimeSeconds": "5" } }

# Empty receive WITHOUT request-level wait → returns in ~0.3s (should block ~5s):
time aws --endpoint-url http://localhost:4566 sqs receive-message --queue-url "$url"

# Positive control — WITH request-level wait → blocks the full 3s:
time aws --endpoint-url http://localhost:4566 sqs receive-message --queue-url "$url" --wait-time-seconds 3
```

Suggested fix, mirroring #1567: parse the parameter as nullable (or track presence), and in `SqsService.receiveMessage` fall back to the queue's `ReceiveMessageWaitTimeSeconds` attribute when the request omits `WaitTimeSeconds`. Happy to contribute a PR if useful.

## References

- #1567 — same bug class for queue-level `DelaySeconds` (merged fix pattern)
- #768 — SQS/SNS conformance sweep; request-level long polling is listed as working, the queue-level default is not covered

## Environment

- Floci version / image tag: `floci:1.7.0-compat` (also verified the behavior matches current `main` source)
- Client: AWS CLI v2 with dummy credentials, AWS SDK for Java v2
- How running: Docker

anmolxlight:
## Summary
Fixes #2826

Queue-level `ReceiveMessageWaitTimeSeconds` was stored but never applied — an empty `ReceiveMessage` without `WaitTimeSeconds` returned immediately instead of long-polling for the queue default. Same defect class as #1567 (queue-level `DelaySeconds`).

## Root cause
- `SqsJsonHandler` and `SqsQueryHandler` parsed `WaitTimeSeconds` as primitive `int` defaulting to `0`, so omission collapsed to `0`.
- `SqsService.receiveMessage` computed `maxWait = waitTimeSeconds * 1000L` and returned immediately on `maxWait <= 0`, never reading the queue attribute.

## Fix
- Parse `WaitTimeSeconds` as nullable `Integer` in both handlers (`parseOptionalInteger` / `getIntegerParam` — mirrors the `DelaySeconds` fix).
- Change `SqsService.receiveMessage` signature from `int` to `Integer`; when `null`, fall back to `parseReceiveMessageWaitTimeSecondsAttribute(queue.getAttributes().get("ReceiveMessageWaitTimeSeconds"))` (default 0).
- Default the attribute to `"0"` on queue creation.

## Test plan
- [x] `./mvnw test -Dtest=SqsServiceTest` — 74 tests pass
- [x] `./mvnw test -Dtest=SqsJsonProtocolTest,SqsIntegrationTest,SqsFifoIntegrationTest` — 51 tests pass
- Manual: queue with `ReceiveMessageWaitTimeSeconds=5`, empty `ReceiveMessage` without `WaitTimeSeconds` blocks ~5s; request-level `WaitTimeSeconds=0` overrides to short poll.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/SqsJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/SqsJsonHandler.java
@@ -243,7 +243,7 @@ public class SqsJsonHandler {
         String queueUrl = request.path("QueueUrl").asText(null);
         int maxMessages = request.path("MaxNumberOfMessages").asInt(1);
         int visibilityTimeout = request.path("VisibilityTimeout").asInt(-1);
-        int waitTimeSeconds = request.path("WaitTimeSeconds").asInt(0);
+        Integer waitTimeSeconds = parseOptionalInteger(request.path("WaitTimeSeconds"), "WaitTimeSeconds");
 
         java.util.Set<String> requestedAttrs = new java.util.LinkedHashSet<>();
         requestedAttrs.addAll(jsonNodeToList(request.path("AttributeNames")));
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/SqsQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/SqsQueryHandler.java
@@ -262,7 +262,7 @@ public class SqsQueryHandler {
         String queueUrl = getParam(params, "QueueUrl");
         int maxMessages = getIntParam(params, "MaxNumberOfMessages", 1);
         int visibilityTimeout = getIntParam(params, "VisibilityTimeout", -1);
-        int waitTimeSeconds = getIntParam(params, "WaitTimeSeconds", 0);
+        Integer waitTimeSeconds = getIntegerParam(params, "WaitTimeSeconds");
 
         java.util.Set<String> requestedAttrs = new java.util.LinkedHashSet<>();
         requestedAttrs.addAll(collectIndexed(params, "AttributeName."));
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/SqsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/SqsService.java
@@ -294,6 +294,7 @@ public class SqsService implements Resettable, ResourceProvider {
         queue.getAttributes().putIfAbsent("VisibilityTimeout", String.valueOf(defaultVisibilityTimeout));
         queue.getAttributes().putIfAbsent("MaximumMessageSize", String.valueOf(maxMessageSize));
         queue.getAttributes().putIfAbsent("DelaySeconds", "0");
+        queue.getAttributes().putIfAbsent("ReceiveMessageWaitTimeSeconds", "0");
         queue.getAttributes().putIfAbsent("MessageRetentionPeriod", "345600");
         if (queue.isFifo()) {
             if (attributes != null && attributes.containsKey("ContentBasedDeduplication") && "true".equals(attributes.get("ContentBasedDeduplication"))) {
@@ -609,6 +610,18 @@ public class SqsService implements Resettable, ResourceProvider {
         }
     }
 
+    private int parseReceiveMessageWaitTimeSecondsAttribute(String value) {
+        if (value == null || value.isEmpty()) {
+            return 0;
+        }
+        try {
+            int parsed = Integer.parseInt(value);
+            return parsed > 0 ? parsed : 0;
+        } catch (NumberFormatException e) {
+            return 0;
+        }
+    }
+
     private void notifyReceivers(String storageKey) {
         Object lock = queueLocks.get(storageKey);
         if (lock != null) {
@@ -637,9 +650,9 @@ public class SqsService implements Resettable, ResourceProvider {
     }
 
     public List<Message> receiveMessage(String queueUrl, int maxMessages, int visibilityTimeout,
-                                        int waitTimeSeconds, String region) {
+                                        Integer waitTimeSeconds, String region) {
         String storageKey = regionKey(region, queueUrl);
-        getQueueByUrl(storageKey, queueUrl)
+        Queue queueForWait = getQueueByUrl(storageKey, queueUrl)
                 .orElseThrow(() -> new AwsException("AWS.SimpleQueueService.NonExistentQueue",
                         "The specified queue does not exist.", 400));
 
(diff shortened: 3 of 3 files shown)


pgermosen (commented):
@anmolxlight welcome, and thanks for this — the fix itself is right. I traced it rather than just reading the description: `parseReceiveMessageWaitTimeSecondsAttribute` mirrors `parseDelaySecondsAttribute` exactly, `parseOptionalInteger`/`getIntegerParam` are the same helpers already proven out by the `DelaySeconds` fix, and the precedence (an explicit `WaitTimeSeconds`, including `0`, always wins over the queue default) is correct. I also want to call out that you didn't copy over the FIFO-only-honors-queue-level quirk `DelaySeconds` has — that's specific to `SendMessage` and doesn't apply here, and copying it anyway would have been a real bug. Good instinct leaving it out.

What's missing is a test locking this in. `SqsServiceTest` already has a directly analogous suite for the sibling `DelaySeconds` fix — `queueLevelDelaySecondsAppliesToStandardQueue`, the one asserting an explicit `DelaySeconds=0` overrides the queue default — that would translate over almost directly. Right now nothing in the suite would catch a future refactor quietly reverting this.

Separately, greptile flagged that a malformed stored `ReceiveMessageWaitTimeSeconds` silently falls back to 0 with no log — worth knowing it's not a new gap, `parseDelaySecondsAttribute` has the identical silent catch already, so you're just following the existing convention. Not blocking, your call whether to address it here or leave it consistent with the rest of the file.

mekwall:
Thanks for this @anmolxlight. The diagnosis was right, and so was the fix approach: requesting `WaitTimeSeconds` as a nullable value in both handlers and falling back to the queue's `ReceiveMessageWaitTimeSeconds` attribute in `SqsService` is exactly the pattern that belongs there.

This PR is superseded: the same defect was independently fixed and merged in #2989 (2026-09-03), which landed alongside everything this PR's review asked for:

- The regression tests in `SqsServiceTest` mirroring the `DelaySeconds` suite (`receiveMessageWithoutWaitTimeUsesQueueReceiveMessageWaitTimeSeconds`, explicit override, short-poll default), plus protocol-level coverage in `SqsJsonProtocolTest` and `SqsIntegrationTest`
- Stricter AWS-faithful validation: non-integer or out-of-range `WaitTimeSeconds` values are rejected with `InvalidParameterValue` (0 to 20), and the queue-attribute fallback is capped at 20
- A Long Polling section in `docs/services/sqs.md`

Nothing in your branch is missing from `main` after #2989; rebasing it onto current `main` collapses the diff to zero. Leaving this open for a maintainer to close. Your write-up in issue #2826 remains the clearest description of the bug and was a big help. Thanks again for the contribution.

pgermosen:
Confirmed this myself against current main rather than just taking it on faith: main already has the same queue-level ReceiveMessageWaitTimeSeconds fallback from #2989, using different names for the same idea (an explicit WaitTimeSeconds, including 0, always wins; otherwise fall back to the queue attribute), and it's actually a bit ahead of what's here, it validates the request value up front and caps the queue-attribute fallback at 20 seconds, which this branch doesn't do.

Closing as superseded rather than asking for a rebase, since there'd be nothing left to carry over. Thanks for tracking down the original bug and writing it up so clearly in #2826, that made both fixes easier to get right.

Issue #2050, opened 2026-07-30
[BUG] API Gateway's REST emulation rejects any request with an `Authorization` header, and concurrent Lambda cold starts are very slow.


guilteixeira:
# Two issues found running a Serverless (osls v4) app locally against Floci

Filing both together since they were found in the same session and share the same environment, happy to split into separate issues if preferred.

## Shared environment

- Floci image: `floci/floci:1.5.34`
- Docker: 29.6.2 (build dfc4efb), Docker Desktop for Mac
- Host: macOS (Darwin 25.5.0, arm64 / Apple Silicon)
- Docker daemon platform: `linux/arm64`
- Lambda runtime under test: `public.ecr.aws/lambda/python:3.14` (managed runtime, not container image)
- Deploy tool: `osls` (Serverless Framework fork) v4.0.0, via `serverless-python-requirements`
- API Gateway: REST API (`AWS::ApiGateway::RestApi`, not HTTP API), invoked via `/restapis/{id}/{stage}/_user_request_/...`
- 4 distinct Lambda functions involved, `individually: false` (single shared deployment package), no `authorizer` configured on any HTTP route (default `NONE` auth type)

---

### Summary

Any HTTP request to a Floci-emulated API Gateway REST API endpoint that includes an `Authorization` header, regardless of its value/format, and regardless of whether the target endpoint has an authorizer configured at all, fails with:

```json
{"message":"Invalid API id specified"}
```
HTTP status: `404`.

The exact same request, with the `Authorization` header removed and every other header unchanged, succeeds normally (`200`, correct response body).

This strongly suggests Floci's API Gateway REST emulation unconditionally attempts to interpret any `Authorization` header as an AWS SigV4 signature (i.e., IAM-authenticated request routing/validation), fails to parse it as such, and returns a generic/misleading "Invalid API id specified" error instead of either (a) ignoring the header since the route's authorizer type is `NONE`, or (b) passing it through untouched to the Lambda proxy integration, which is what real AWS API Gateway does for `NONE`-auth routes (the `Authorization` header is just forwarded to the Lambda like any other header, with zero validation).

### Steps to reproduce

1. Deploy a Serverless app with a Lambda-proxy HTTP endpoint that has no authorizer configured (default), e.g.:
   ```yaml
   functions:
     myFunc:
       handler: main.my_handler
       events:
         - http:
             path: api/v1/example
             method: get
   ```
2. Get Floci's REST API invoke URL for that stage:
   ```
[...]

PR #2830, opened 2026-08-31
fix(lambda): make the code-volume populate concurrency configurable

bsaurusrex:
## Summary

Concurrent first-time code-volume populates are capped by a semaphore sized
`Math.max(2, Runtime.getRuntime().availableProcessors() / 2)`, a `private static final` with no
configuration override. The JVM honours the container's cgroup CPU quota, so a CPU-constrained
Floci container derives **2** permits, and concurrent cold starts of *distinct* large functions
complete in waves of two instead of in parallel.

This adds `floci.services.lambda.code-volume-populate-concurrency`
(`FLOCI_SERVICES_LAMBDA_CODE_VOLUME_POPULATE_CONCURRENCY`) and **leaves the derived default
unchanged**.

Refs #2050 (see "Relationship to #2050" below).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

No wire-protocol surface is touched — this is local execution behaviour, not an AWS API shape, so
there is no AWS reference to cite or live API to diff against. Real Lambda has no equivalent knob
(AWS provisions execution environments per function without a shared local Docker daemon to
protect), so the honest framing is fidelity of *behaviour under concurrency*: cold-starting N
distinct functions should cost roughly `max(individual cold start)`, not `N/permits x` that.

### Measurements

Six distinct functions, 41 MB packages each (over the 32 MB `CODE_VOLUME_MIN_BYTES` threshold, so
each takes the populate path), all cold, fired concurrently. Each function gets its own volume —
`codeVolumeName` keys on function name as well as code SHA, verified with `docker volume ls` — so
these are six genuinely independent populates, not one populate plus five cache hits.

Varying **only** the CPU quota on Floci's own container, against `main` (`f7f16cd`):

| Floci container CPU quota | permits from `max(2, cpus/2)` | observed waves | wall-clock |
|---|---|---|---|
| unlimited (16 cpu) | 8 | one wave of 6 | **8.3 s** |
| `--cpus=8` | 4 | 4, then 2 | **13.2 s** |
| `--cpus=4` | 2 | 2, 2, 2 | **19.2 s** |

The wave sizes track the formula exactly. That is what rules out generic CPU starvation as the
explanation — starvation would slow all six roughly uniformly toward a single max, not produce
discrete equal-size waves whose size is a function of the permit count. The `--cpus=8` row was run
as a falsifiable prediction (predicted 4-then-2 before running it) precisely to separate the two.

With this patch, same `--cpus=4` container:

| Configuration | wall-clock |
|---|---|
| default (env var unset) | **18.9 s** — unchanged, still waves of 2 |
| `FLOCI_SERVICES_LAMBDA_CODE_VOLUME_POPULATE_CONCURRENCY=8` | **8.5 s** — one wave of 6 |

So the default is preserved (18.9 s vs 19.2 s unpatched, same wave shape) and the override does
what it says.

## Why keep the default

The cap is worth having — the daemon overload it was added to prevent is real, and its safe
ceiling is host-specific, so I did not want to guess a new default in this PR. What was missing is
a way to decouple it from the CPU allocation. Worth noting for whoever revisits the default: the
populate is IO-bound (create a helper container, stream a tar into it), so CPU count is a weak
proxy for how many the daemon can absorb — but changing the default is a separate judgement call
with a real failure mode behind it, and is deliberately out of scope here.

Values below 1 fall back to the derived default with a warning: zero permits would deadlock every
populate, and a negative `Semaphore` count would make `acquire()` block until that many releases
arrived.

The semaphore becomes an instance field so the permit count can come from config. Floci runs a
single `@ApplicationScoped` launcher, so it is still one gate for the whole emulator.

## Relationship to #2050

#2050 bundles two bugs. I reproduced both before touching anything, building from `main` rather
than a release tag.

**Bug 1 (API Gateway 404 on any `Authorization` header) is already fixed on `main`** — by #2149,
merged 2026-08-14 and released in 1.7.0. The reporter was on 1.5.34 (released 2026-07-29), which
predates it. Nothing in this PR touches it.

Worth recording that the reported *mechanism* was not the actual one. The report hypothesised that
Floci unconditionally interprets any `Authorization` header as a SigV4 signature. It does not —
the cause was region resolution: `dispatch()` only ran its cross-region API-id scan when the header
was absent or blank, so any non-SigV4 header (non-blank) skipped the scan and lookup fell back to
the *default* region, where an API deployed elsewhere does not exist. Measured on 1.5.34, REST API
deployed to `us-west-2`, default region `us-east-1`:

| `Authorization` header | result |
|---|---|
| *(none)* | API found (scan ran) |
| `Bearer <token>` | 404 `Invalid API id specified` |
| SigV4, credential scope `us-west-2` | API found |
| SigV4, credential scope `us-east-1` | 404 `Invalid API id specified` |

A `Bearer` header behaves *identically to a well-formed SigV4 header naming the wrong region* —
which is what refutes the interception theory. All four rows return 200 on `main`.

**Bug 2 (concurrent cold starts serialise) is what this PR addresses**, and `main` had already
improved it substantially: the same 6-function test takes 18.8 s on 1.5.34 (stepped in pairs) and
8.3 s on unconstrained `main`. What survives is the CPU-derived cap above, which is exactly the
configuration a laptop Docker Desktop VM lands in — so the reporter's symptom is still reachable
on current `main`, just via a mechanism their "global lock / concurrency=1" guess only half
described.

Between #2149 and this change I believe #2050 is fully addressed, but since it is a two-bug issue
where the other half was fixed elsewhere, I have used `Refs` rather than `Closes` and left the call
to a maintainer.

## Checklist

- [x] Full suite passes locally — run as CI does, sharded 4 ways via `.github/ci/partition.sh`:
      **14,614 tests, 0 failures, 0 errors** (shards of 3407 / 3132 / 3609 / 4466)
- [x] New or updated test added — `ContainerLauncherPopulateConcurrencyTest`, mirroring
      `ContainerLauncherNamePrefixTest`'s plain-`mock()` pattern for config resolution
- [x] `make docs-check` clean
- [x] End-to-end verified against a running emulator, both default and overridden (table above)
- [x] Node compatibility suite run against a live emulator on the patched image:
      **458 passed, 2 failed** (`apigatewayv2-websocket-dataplane`, `apigatewayv2-websocket-tls`,
      both "Chat-style broadcast" → `Timeout waiting for message`). **Attributed as pre-existing:**
      the released `floci/floci:1.7.0` image, unpatched, fails the identical two tests under the
      same harness — `2 failed | 32 passed (34)` on both. Neither touches Lambda code-volume
      populate concurrency.

### DIFF
--- a/docs/configuration/environment-variables.md
+++ b/docs/configuration/environment-variables.md
@@ -197,6 +197,7 @@ See [Initialization Hooks](./initialization-hooks.md) for lifecycle phases and s
 | `FLOCI_SERVICES_LAMBDA_CONTAINER_IDLE_TIMEOUT_SECONDS` | `300` | Seconds of inactivity before an idle Lambda container is removed |
 | `FLOCI_SERVICES_LAMBDA_REGION_CONCURRENCY_LIMIT` | `1000` | Maximum concurrent Lambda invocations across all functions in a region |
 | `FLOCI_SERVICES_LAMBDA_UNRESERVED_CONCURRENCY_MIN` | `100` | Minimum unreserved concurrency pool |
+| `FLOCI_SERVICES_LAMBDA_CODE_VOLUME_POPULATE_CONCURRENCY` | `max(2, cpus/2)` | Maximum concurrent first-time code-volume populates (functions whose unpacked code is at least 32 MB). The default is derived from the CPU count the JVM sees, so a CPU-constrained Floci container collapses it to 2 and concurrent cold starts of distinct functions serialise into pairs; set this to decouple the cap from the CPU allocation |
 | `FLOCI_SERVICES_LAMBDA_HOT_RELOAD_ENABLED` | `false` | Watch Lambda code directories for changes and reload without redeployment |
 | `FLOCI_SERVICES_LAMBDA_HOT_RELOAD_ALLOWED_PATHS` | _(none)_ | Comma-separated host paths that hot-reload is allowed to watch |
 | `FLOCI_SERVICES_LAMBDA_DOCKER_NETWORK` | _(none)_ | Docker network for Lambda containers (overrides `FLOCI_SERVICES_DOCKER_NETWORK`) |
--- a/docs/services/lambda.md
+++ b/docs/services/lambda.md
@@ -207,6 +207,7 @@ These AWS Lambda operations have no handler in Floci. Calls will return `404` or
 | `FLOCI_SERVICES_LAMBDA_EXTRA_HOSTS` | *(unset)* | Comma-separated `hostname:ip` entries added to each Lambda container's `/etc/hosts`; `ip` may be `host-gateway`, mirroring `docker run --add-host` |
 | `FLOCI_SERVICES_LAMBDA_DOCKER_HOST_OVERRIDE` | *(unset)* | Explicit host/IP that spawned Lambda containers use to reach Floci's Runtime API, bypassing auto-detection |
 | `FLOCI_SERVICES_LAMBDA_CONTAINER_NAME_PREFIX` | `floci` | Base name prefix for spawned Lambda containers and code volumes (e.g. `acme` → `acme-<function>-<id>` containers, `acme-code-<function>-<hash>` volumes). Must be a valid Docker name segment (`[A-Za-z0-9][A-Za-z0-9_.-]*`); invalid values are ignored with a warning |
+| `FLOCI_SERVICES_LAMBDA_CODE_VOLUME_POPULATE_CONCURRENCY` | `max(2, cpus/2)` | Maximum concurrent first-time code-volume populates. See the note below |
 | `FLOCI_SERVICES_LAMBDA_EXECUTOR` | `docker` | Execution backend: `docker` (containers) or `kubernetes` (pods) |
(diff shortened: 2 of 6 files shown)


pgermosen (approved):
@bsaurusrex good one, and I traced the mechanics rather than the measurement tables. The semaphore genuinely stays a single gate for the whole emulator moving from `static final` to an instance field, since `ContainerLauncher` is `@ApplicationScoped` — and every reference to the old static (`acquirePopulatePermit`, the release in `populateCodeVolume`, the comment in `copyDirToContainer`) got updated consistently, no leftover reference to the removed field. `resolvePopulateConcurrency` being a plain static method over `EmulatorConfig` is why it's directly testable without standing up a launcher, and the fallback behavior matches what's documented: unset derives from CPU count, an explicit value wins either direction, non-positive values warn and fall back rather than deadlock or block `acquire()` forever on a negative permit count.

Keeping the default unchanged and only adding the decoupling knob is the right scope call — the daemon-overload cap is real and its safe ceiling is host-specific, so guessing a new default here would've been a different, riskier PR.

Approving.

PR #2836, opened 2026-08-31
Feat/redshift sql interceptor

vuthehuyht:
## Summary

Implement **Redshift SQL Interceptor & S3 COPY/UNLOAD Simulator** to enable data-plane compatibility for Redshift clusters running on top of PostgreSQL containers.

Key capabilities introduced:
- **Wire-level SQL Interceptor**: Intercepts Simple Query (`'Q'`) messages on the PostgreSQL wire protocol (v3.0).
- **DDL Keyword Rewriting**: Strips Redshift-specific table keywords (`DISTSTYLE`, `DISTKEY`, `SORTKEY`, `ENCODE`) fail-open so standard Redshift DDL executes natively on PostgreSQL.
- **S3 COPY Simulator**: Emulates `COPY <table> FROM 's3://bucket/keyOrPrefix'` by fetching S3 objects via `S3Service` (supporting CSV, GZIP, custom `DELIMITER`, `IGNOREHEADER`, `NULL AS`, column lists) and streaming them into the backend container via PostgreSQL CopyIn protocol (`COPY ... FROM STDIN`).
- **S3 UNLOAD Simulator**: Emulates `UNLOAD ('<select>') TO 's3://bucket/prefix'` by executing `COPY (<select>) TO STDOUT`, aggregating data chunks, compressing (GZIP if requested), uploading `<prefix>000` to S3, and optionally generating `<prefix>manifest`.
- **Reusable RDS/Redshift Proxy Primitives**: Refactored `PostgresProtocolHandler` into `authenticate(...)` and `bridge(...)` primitives without altering RDS behavior.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- **Protocols & Clients**:
  - PostgreSQL Wire Protocol v3.0 Simple Query (`'Q'`).
  - Compatible with PostgreSQL JDBC driver, Redshift JDBC driver (`com.amazon.redshift:redshift-jdbc42`), and `psql` CLI.
  - Verified S3 data-plane inter-operation against Floci `S3Service` (handling single object S3 paths and prefix list resolutions).

## Checklist

- [x] New or updated integration test added (`RedshiftInterceptorIntegrationTest`, plus comprehensive unit test suites: `RedshiftSqlInterceptorTest`, `CopyStatementParserTest`, `PostgresWireDecoderTest`, `S3CopySimulatorTest`, `RedshiftInterceptingBridgeTest`)


pgermosen (commented):
@vuthehuyht there's real engineering here, but I don't think this should land as one PR, and I'd like to see it split up rather than patched further in place.

Twenty-one of the twenty-seven commits are reactive fixes to problems the feature itself introduced, chased across eleven rounds of review: a P0 SQL-injection RCE that let a crafted UNLOAD subquery break out of the fabricated COPY wrapper and run arbitrary commands under the backend superuser, then a run of six-plus separate wire-protocol concurrency bugs in the pump-pause/park mechanism, each fix surfacing the next race. That's the shape a diff takes when a brand-new, security-critical subsystem, raw wire-protocol interception with string-built SQL spliced back into the backend, gets built and hardened against a bot in one sitting instead of being reviewed in pieces a person can actually reason about. And it's still not done: the most recent review, against your final commit, says outright it isn't safe to merge yet, over a TOCTOU gap where UNLOAD's S3 authorization is checked once up front but not enforced at the point of the actual write.

Please split this into smaller PRs: the RDS proxy refactor on its own, since it doesn't need anything else here to be reviewable; the DDL keyword rewriting; the S3 COPY simulator; the S3 UNLOAD simulator; and the pump/concurrency mechanism they both depend on. Each of those is independently reviewable, and right now the whole thing is one unit that nobody, bot or human, has actually been able to sign off on.

Separately, and this is worth fixing regardless of how the split lands: I checked the actual UNLOAD reference, and real Redshift's default format, no CSV keyword, is pipe-delimited TEXT, not CSV. TEXT and CSV aren't just different delimiters, they have different quoting rules and a different NULL marker (\N for TEXT vs. an empty field for CSV). CopyStatementParser.extractDelimiter gets the delimiter default right, but both runCopyFrom and runUnload in S3CopySimulator hardcode FORMAT csv on the PostgreSQL side unconditionally, so a bare UNLOAD with no CSV keyword, the common case and Redshift's own documented default, runs against the backend under CSV parsing rules with the delimiter swapped to pipe. A field with a literal double-quote gets CSV-quoted where real Redshift wouldn't touch it, and a COPY FROM of a genuine pipe-delimited Redshift file using \N for null would silently fail to recognize those as nulls. Nothing in the test suite currently asserts on TEXT-vs-CSV semantics, only on the delimiter character.

The PR title is also just the branch name (`Feat/redshift sql interceptor`) rather than a Conventional Commits title; small, but worth fixing since it becomes the squash headline.

PR #2837, opened 2026-08-31
fix(lambda): isolate legacy code storage by account

AbarnaaSree:
## Summary

Fixes legacy Lambda code storage so functions with the same name in different AWS accounts use separate code directories.

## Changes

* Make `CodeStore` account-aware by storing code under:

  * `<accountId>/<functionName>`
* Update Lambda code extraction to use the function's account ID.
* Update Lambda deletion to delete code from the correct account-specific directory.
* Update function creation to check for existing functions within the current account.
* Add tests covering:

  * Same function name across different accounts
  * Account-specific code storage paths

## Testing

* `mvn test "-Dtest=CodeStoreTest,LambdaServiceTest"`
* **53 tests passed**
* **0 failures**
* **0 errors**

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -279,7 +279,9 @@ public class LambdaService {
             throw new AwsException("InvalidParameterValueException", "Runtime is required for Zip package type", 400);
         }
 
-        if (functionStore.get(region, functionName).isPresent()) {
+        String accountId = regionResolver.getAccountId();
+
+        if (functionStore.getForAccount(accountId, region, functionName).isPresent()) {
             throw new AwsException("ResourceConflictException",
                     "Function already exist: " + functionName, 409);
         }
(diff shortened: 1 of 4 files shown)


pgermosen:
@AbarnaaSree welcome, and thanks for digging into this — the diagnosis is right, but the fix already shipped. `main` already stores Lambda code under `<codePath>/<accountId>/<functionName>` via #2646, which is why this branch conflicts: both add a `CodeStoreTest.java`, independently, and both touch the same account-scoping logic in `CodeStore.java`/`LambdaService.java`.

Worth a look at `main`'s current `CodeStore.java` and the `reclaimLegacyCodeDirectoryIfUnused` method in `LambdaService.java` — it also handles something this PR doesn't yet: functions created before the account-scoping migration share one legacy on-disk directory across accounts (since the old layout had no account segment), so deleting it outright as soon as one account's function moves off it would break another account's still-live function. Main's version checks every account's functions, published versions included, before reclaiming that shared directory — which is the exact gap greptile flagged here as a P1 (legacy code deletion getting bypassed).

hectorvent:
Thank you for the diagnosis, it was right, and pgermosen's note above has the outcome exactly. The fix landed on `main` in #2646 three days before this PR was opened: `CodeStore` already scopes code under `<codePath>/<accountId>/<functionName>`, and `LambdaService.reclaimLegacyCodeDirectoryIfUnused` handles the one case this branch does not, the pre-migration directory that several accounts' same-named functions shared, which is what the bot's comment about orphaned legacy code is pointing at. That is also why the branch conflicts in all four files.

**(blocking)** Nothing here survives a rebase onto `main`, so the kind thing is to close this as superseded rather than ask for a rework. If you would like to keep contributing, the issues carrying the `lambda` label without a linked PR are a good place to look, and asking on Slack before starting avoids this kind of overlap.

Closing as superseded by #2646.

PR #2841, opened 2026-08-31
fix(ses): return DeliveryOptions from v1 DescribeConfigurationSet

### LINKED ISSUE #2833: [BUG] SES v1 DescribeConfigurationSet omits DeliveryOptions, although SESv2 returns it
## Service

SES (v1)

## AWS API Action

`DescribeConfigurationSet` (with `ConfigurationSetAttributeNames=deliveryOptions`), `PutConfigurationSetDeliveryOptions`

## Expected behavior

Real AWS returns the stored delivery options when they are requested:

```
$ aws ses describe-configuration-set --configuration-set-name <set> \
    --configuration-set-attribute-names deliveryOptions
{
    "ConfigurationSet": { "Name": "<set>" },
    "DeliveryOptions": { "TlsPolicy": "Require" }
}
```

## Actual behavior

`PutConfigurationSetDeliveryOptions` succeeds (200) and Floci **does** persist the value —
SESv2 `GetConfigurationSet` returns it correctly. Only the v1 read path drops it:
`DescribeConfigurationSet` omits the `DeliveryOptions` member entirely, even when
`deliveryOptions` is explicitly named in `ConfigurationSetAttributeNames`.

```
{
    "ConfigurationSet": { "Name": "tls-probe" }
}
```

So this is not a missing feature — the store is already there (#1553 implemented the
write). It is the v1 projection that never reads it back.

One detail for whoever fixes it: the two APIs use different enum casing, and Floci's v2
already gets this right. Measured against live AWS on the same configuration set:

| API | Live AWS | Floci |
| --- | --- | --- |
| v1 `DescribeConfigurationSet` | `"TlsPolicy": "Require"` | member absent |
| v2 `GetConfigurationSet` | `"TlsPolicy": "REQUIRE"` | `"TlsPolicy": "REQUIRE"` ✅ |

So the v1 response wants `Require`/`Optional`, not the v2 spelling.

## Why it matters

The Terraform AWS provider reads `aws_ses_configuration_set` through the **v1**
`DescribeConfigurationSet`. Because the member never comes back, every plan re-adds it:

```
~ resource "aws_ses_configuration_set" "example" {
    + delivery_options {
        + tls_policy = "Require"
      }
  }
```

That is a permanent diff — `apply` converges nothing, and the resource is rewritten on
every run.

## Reproduction

```bash

aws $E ses create-configuration-set --configuration-set '{"Name":"tls-probe"}'
aws $E ses put-configuration-set-delivery-options \
      --configuration-set-name tls-probe --delivery-options TlsPolicy=Require   # 200

# v1: DeliveryOptions missing
aws $E ses describe-configuration-set --configuration-set-name tls-probe \
      --configuration-set-attribute-names deliveryOptions
# => {"ConfigurationSet": {"Name": "tls-probe"}}

# v2: the value is there
aws $E sesv2 get-configuration-set --configuration-set-name tls-probe
# => {..., "DeliveryOptions": {"TlsPolicy": "REQUIRE"}, ...}
```

API reference:
[DescribeConfigurationSet](https://docs.aws.amazon.com/ses/latest/APIReference/API_DescribeConfigurationSet.html),
[PutConfigurationSetDeliveryOptions](https://docs.aws.amazon.com/ses/latest/APIReference/API_PutConfigurationSetDeliveryOptions.html).

## Environment

bsaurusrex:
## Summary

`PutConfigurationSetDeliveryOptions` already persisted the TLS policy — v2 `GetConfigurationSet` returned it correctly — but the v1 `DescribeConfigurationSet` projection never read it back. The member was omitted even when `deliveryOptions` was named in `ConfigurationSetAttributeNames`.

So this is not a missing feature. #1553 implemented the store; only the v1 read path was missing.

Closes #2833.

## Type of change

- [x] Bug fix (`fix:`)

## Why it matters

The Terraform AWS provider reads `aws_ses_configuration_set` through the **v1** `DescribeConfigurationSet`. Because the member never came back, every plan re-added it:

```
~ resource "aws_ses_configuration_set" "example" {
    + delivery_options {
        + tls_policy = "Require"
      }
  }
```

`apply` converged nothing and the resource was rewritten on every run.

## AWS Compatibility

API reference: [DescribeConfigurationSet](https://docs.aws.amazon.com/ses/latest/APIReference/API_DescribeConfigurationSet.html), [PutConfigurationSetDeliveryOptions](https://docs.aws.amazon.com/ses/latest/APIReference/API_PutConfigurationSetDeliveryOptions.html).

Measured against `email.ap-southeast-1.amazonaws.com` in an account I own — reads on an existing configuration set, and a throwaway set created and deleted for the mutations.

**The v1 wire casing differs from v2, and it is `Require`/`Optional`:**

```
$ aws ses describe-configuration-set --configuration-set-name <set> \
      --configuration-set-attribute-names deliveryOptions
{
    "ConfigurationSet": { "Name": "<set>" },
    "DeliveryOptions": { "TlsPolicy": "Require" }
}

$ aws sesv2 get-configuration-set --configuration-set-name <set> --query 'DeliveryOptions'
{ "TlsPolicy": "REQUIRE" }
```

**The member is gated on the attribute name.** Same set, same moment, no `--configuration-set-attribute-names`:

```
{
    "ConfigurationSet": { "Name": "<set>" }
}
```

**An unconfigured set omits it even when it is requested** — there is no implicit `Optional` default on the v1 read:

| Set state | v1 `DescribeConfigurationSet` (deliveryOptions requested) | v2 `GetConfigurationSet` |
| --- | --- | --- |
| never configured | member absent | `null` |
| `TlsPolicy=Require` | `{"TlsPolicy": "Require"}` | `{"TlsPolicy": "REQUIRE"}` |
| `TlsPolicy=Optional` | `{"TlsPolicy": "Optional"}` | `{"TlsPolicy": "OPTIONAL"}` |

That last row is why the fix converts rather than lower-casing or passing through: Floci stores the v2 canonical `REQUIRE`/`OPTIONAL` (`handlePutConfigurationSetDeliveryOptions` normalises v1 input to it), so the v1 read path has to convert back. An unrecognised stored value is passed through unchanged rather than mangled.

| Call | Live AWS | Floci before | Floci after |
| --- | --- | --- | --- |
| v1 `DescribeConfigurationSet`, requested, configured | `Require` / `Optional` | member absent | matches AWS |
| v1 `DescribeConfigurationSet`, requested, unconfigured | member absent | member absent | unchanged |
| v1 `DescribeConfigurationSet`, not requested | member absent | member absent | unchanged |
| v2 `GetConfigurationSet` | `REQUIRE` / `OPTIONAL` | correct already | unchanged |

## Scope

The v1 read path only. The store, the v1 and v2 write paths and the whole v2 surface are untouched — `PutConfigurationSetDeliveryOptions` already validated the enum and normalised it, and that behaviour is unchanged.

`DeliveryOptions` on v1 carries only `TlsPolicy`; `SendingPoolName` and `MaxDeliverySeconds` are v2-only members and are deliberately not projected into the v1 shape.

The new block follows the existing `reputationOptions` / `trackingOptions` blocks in the same method, including their guard style — emitted only when the attribute is requested *and* a value is actually stored.

## Testing

- `SesConfigurationSetV1IntegrationTest#describeConfigurationSetReturnsDeliveryOptions` — new, self-contained (creates and deletes its own configuration set, so it leaves nothing for sibling classes). Covers all four measured cases: unconfigured-and-requested, `Require`, `Optional`, and requested-vs-not-requested gating.
- Negative control: with the `SesQueryHandler` change reverted, that test fails on `Expected: a string containing "<DeliveryOptions>"` (`SesConfigurationSetV1IntegrationTest.java:311`), and passes with it applied.
- The gap survived because `putConfigurationSetDeliveryOptions` (Order 11) writes a policy and never describes it — a write-only test. The new test is the read half.
- SES package: **1046 tests, 0 failures**.
- Full suite: **14,701 tests, 0 failures**, 8 of 8 shards green in one pass on this branch (sharded the way `ci.yml` partitions it, at 8 rather than 4).
- Compatibility suite (`sdk-test-node`, run the way CI does — inside the container network with `--dns` pointed at Floci, image built from this branch): **458 of 460 passing**. The two failures are the WebSocket chat-broadcast cases in `apigatewayv2-websocket-dataplane` and `apigatewayv2-websocket-tls`, the same pair that fails on released `floci/floci:1.7.0` — pre-existing and unrelated.
- Checked all three SDK compatibility suites (node, python, java) for assertions on the v1 shape. The only class that touches it is `SesConfigurationSetDeliveryOptionsV1Test`, which writes through v1 and deliberately reads back through **v2** — a workaround that exists precisely because the v1 read path did not work. Its assertions are on `GetConfigurationSetResponse` and are unaffected; it would now be possible to assert the v1 read directly, but I have left it alone to keep this change to one concern.
- `make docs-check`: clean.

## Checklist

- [x] I have made corresponding changes to the documentation


pgermosen (approved):
@bsaurusrex clean fix, and I checked rather than trusted it. Pulled the actual `DeliveryOptions` reference myself — `Require | Optional` is exactly the v1 casing, matching your measurements. And the new gate's guard shape (`attrs.contains(...) && cs.getDeliveryOptions() != null && cs.getDeliveryOptions().getTlsPolicy() != null`) is a precise match for the existing `trackingOptions` block right above it, not just similar in spirit.

Approving.

PR #2844, opened 2026-08-31
fix(elasticache): report transit encryption, not the auth-token flag

allensanborn:
## Summary

`TransitEncryptionEnabled` was serialized as `authMode == PASSWORD` in both the replication-group view and the member-cache-cluster view. An auth token does force encryption in transit, but the reverse does not hold — and this handler already knows it does not: `CreateReplicationGroup` accepts `TransitEncryptionEnabled=true` with no token and records exactly that request as `AuthMode.IAM`. Every read then reported such a group as unencrypted in transit.

That is not cosmetic for Terraform. `aws_elasticache_replication_group` reads `transit_encryption_enabled` back on every plan rather than trusting what it sent, so a group created with TLS and no auth token showed permanent drift.

`DescribeCacheClusters` had a second, narrower version of the same problem: `AtRestEncryptionEnabled` was a hardcoded `false` on the member view, so the two describes contradicted each other about one group's storage encryption. It now reports what the group stores — the value `CreateReplicationGroup` already parsed and kept.

`AuthTokenEnabled` is unchanged and still means what it says: only a token sets it.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

In the ElastiCache API, `AuthTokenEnabled` and `TransitEncryptionEnabled` are separate members of `ReplicationGroup` and of `CacheCluster`, and AWS reports them independently: a group with `TransitEncryptionEnabled=true` and no `AuthToken` comes back with transit true and auth-token false. AWS's constraint runs one way only — supplying an auth token *requires* transit encryption — which is why deriving one flag from the other is right in that direction and wrong in the other.

Three tests added to `ElastiCacheQueryHandlerTest`: transit reported for a token-less encrypted group (with `AuthTokenEnabled` still false alongside it), transit still false for a plain group, and the member-cluster view agreeing with the group view on both flags.

## Checklist

- [x] `./mvnw test -Dtest='ElastiCache*'` — every unit and non-container suite passes (`ElastiCacheQueryHandlerTest` 16/16, `ElastiCacheServiceTest` 23/23)
- [x] `make docs-check` clean

One note on local verification: `ElastiCacheIntegrationTest` and `ElastiCacheClusterIntegrationTest` fail on my machine with `java.net.BindException: Can't assign requested address` when starting their Redis containers. I confirmed the identical failures on unmodified `main` at the same commit before touching anything — same tests, same counts — so this is my Docker networking, not the change. CI covers them.

Found by a compatibility survey that runs the Gruntwork Terraform corpus against Floci.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheQueryHandler.java
@@ -370,8 +370,8 @@ public class ElastiCacheQueryHandler {
                   .elem("ReplicationGroupId", g.getReplicationGroupId())
                   .elem("AutoMinorVersionUpgrade", true)
                   .elem("AuthTokenEnabled", authTokenEnabled)
-                  .elem("TransitEncryptionEnabled", authTokenEnabled)
-                  .elem("AtRestEncryptionEnabled", false);
+                  .elem("TransitEncryptionEnabled", transitEncryptionEnabled(g))
+                  .elem("AtRestEncryptionEnabled", g.isAtRestEncryptionEnabled());
         if (g.getEngine() != null) {
             xml.elem("Engine", g.getEngine());
         }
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Approving. Traced handleCreateReplicationGroup myself to confirm the core claim - it really does set AuthMode.IAM exactly when TransitEncryptionEnabled=true with no AuthToken, so authMode != NO_AUTH is the correct inverse for reporting transit encryption back. The AtRestEncryptionEnabled fix is just wiring the member view to state that was already being parsed and stored correctly at create time. Tests reproduce the exact drift shapes described. Clean fix.

PR #2845, opened 2026-08-31
fix(ecs): apply cluster settings given on CreateCluster

### LINKED ISSUE #2806: ecs: CreateCluster drops the settings field, causing permanent Terraform drift
`EcsJsonHandler.handleCreateCluster` never parses or forwards the request's `settings` field:

```java
private Response handleCreateCluster(JsonNode req, String region) {
    String name = req.path("clusterName").asText(null);
    Map<String, String> tags = parseTagMap(req.path("tags"));
    EcsCluster cluster = service.createCluster(name, tags, region);
    ...
```

`EcsService.createCluster` doesn't even have a `settings` parameter — compare `handleUpdateCluster`/`handleUpdateClusterSettings`, which both correctly call `parseClusterSettings(req.path("settings"))`. Real AWS's `CreateCluster` accepts `settings` (e.g. `containerInsights`) directly at creation time; Floci silently drops it, so a subsequent `DescribeClusters` never reflects it.

Effect: any client that sets `settings` on create (e.g. `terraform-provider-aws`'s `aws_ecs_cluster` `setting` block) sees permanent drift — every `terraform plan` after `apply` proposes adding the setting again, forever, even though it was already specified.

Confirmed still present on `main`.

Fix direction: parse `settings` in `handleCreateCluster` the same way the update handlers already do, and pass it through to `EcsService.createCluster`.

Origin: found via the "re-planning the full configuration reports no changes" drift-check test in #1834 — worked around there by dropping the (functionally redundant, since `disabled` is the default) `setting { name = "containerInsights", value = "disabled" }` block from the compat suite's `aws_ecs_cluster` rather than blocking the PR on this fix.

avison9:
## Summary

Closes #2806.

`EcsJsonHandler.handleCreateCluster` read `clusterName` and `tags` and nothing else, and
`EcsService.createCluster` had no `settings` parameter at all, so settings sent at creation were
dropped. Real AWS accepts them on `CreateCluster` itself, not only through `UpdateClusterSettings`.

The effect is permanent drift for any client that configures a cluster at creation.
`terraform-provider-aws` sends its `setting` block on create, reads `DescribeClusters` back without
it, and proposes the same change on every `plan` after `apply`, forever.

## Root cause

The create path never parsed the field, while the update path two methods away already did:

```java
// handleCreateCluster
String name = req.path("clusterName").asText(null);
Map<String, String> tags = parseTagMap(req.path("tags"));
EcsCluster cluster = service.createCluster(name, tags, region);   // settings never read

// handleUpdateClusterSettings, in the same file
List<ClusterSetting> settings = parseClusterSettings(req.path("settings"));
```

Everything downstream was already in place: `clusterNode` serialises `settings`, and the
`EcsCluster` model carries them. Create was the only missing link.

## What changed

The handler parses `settings` with the existing `parseClusterSettings` helper and passes them to a
new `createCluster` overload. The previous overloads delegate to it, so the CloudFormation caller at
`CloudFormationResourceProvisioner` is untouched and no other call site changes.

Nineteen lines across two files.

## AWS Compatibility

`CreateCluster` accepts `settings` in the real API, and `DescribeClusters` reflects them afterwards.
Floci now matches on both. A cluster created without settings still reports none rather than an
empty array, which matters because a client diffing its configuration would treat an added empty
array as a change too.

## How it was tested

New `EcsCreateClusterSettingsIntegrationTest`:

| Test | Covers |
|---|---|
| `settingsGivenOnCreateAreReturnedByCreateAndSurviveToDescribe` | The create response carries the settings, and so does a later `DescribeClusters`, which is what a client actually compares its configuration against |
| `aClusterCreatedWithoutSettingsReportsNone` | The absent case stays absent rather than gaining an empty array |

Reverting only the handler change makes the first fail, so it is a real regression test rather than
a passing assertion.

Full suite: `Tests run: 14737, Failures: 0, Errors: 1, Skipped: 9`.

## One note on that error, since the suite is not green

`java.sql.SQLException: No suitable driver found for jdbc:h2:mem:...`. **It is not from this
change**, and I checked rather than assumed:

| Run | RdsDataServiceTest |
|---|---|
| Clean `main` at `3aaf7452`, my change stashed | FAIL |
| This branch, run 1 | FAIL |
| This branch, run 2 | FAIL |

It fails identically on unmodified `main`, and passes when the class is run on its own, so it looks
like a driver registration race under the full parallel suite. Happy to file that separately if it
is not already known.

For completeness: `CodePipelineIntegrationTest.putApprovalResultValidatesErrors` also failed once on
this branch, with `Expected status code <400> but was <500>` on a repeat approval. It did not
reproduce on a second full run and passes in isolation, so it appears order dependent rather than
related to this change. Flagging it rather than leaving it unmentioned.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/EcsJsonHandler.java
@@ -143,7 +143,8 @@ public class EcsJsonHandler {
     private Response handleCreateCluster(JsonNode req, String region) {
         String name = req.path("clusterName").asText(null);
         Map<String, String> tags = parseTagMap(req.path("tags"));
-        EcsCluster cluster = service.createCluster(name, tags, region);
+        List<ClusterSetting> settings = parseClusterSettings(req.path("settings"));
+        EcsCluster cluster = service.createCluster(name, tags, settings, region);
         ObjectNode resp = objectMapper.createObjectNode();
         resp.set("cluster", clusterNode(cluster));
         return Response.ok(resp).build();
--- a/src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ecs/EcsService.java
@@ -221,6 +221,20 @@ public class EcsService implements ContainerTeardown, ResourceProvider {
     }
 
     public EcsCluster createCluster(String clusterName, Map<String, String> tags, String region) {
+        return createCluster(clusterName, tags, null, region);
+    }
+
+    /**
+     * Creates a cluster, applying any {@code settings} the request carried.
+     *
+     * <p>Real AWS accepts {@code settings} (containerInsights and friends) on CreateCluster itself,
+     * not only through UpdateClusterSettings. Dropping them here meant a client that configured the
+     * cluster at creation, such as terraform-provider-aws's {@code setting} block, saw
+     * DescribeClusters come back without them and proposed the same change on every subsequent
+     * plan, forever (issue #2806).
+     */
+    public EcsCluster createCluster(String clusterName, Map<String, String> tags,
+                                    List<ClusterSetting> settings, String region) {
         String name = (clusterName == null || clusterName.isBlank()) ? DEFAULT_CLUSTER : clusterName;
         String key = clusterKey(region, name);
         if (clusters.containsKey(key)) {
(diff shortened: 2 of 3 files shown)


pgermosen (approved):
Approving. Traced the overload chain to confirm the CloudFormation caller really is untouched - the 2-arg and 3-arg createCluster overloads now delegate down to the new 4-arg one with settings=null, and CloudFormationResourceProvisioner uses the 2-arg one. The absent-stays-absent behavior falls out of the existing serializer with no new logic needed. Good test too, hitting the real REST path for both the fix and the empty-array edge case. The two flaky-test callouts are trivially unrelated by scope - this diff doesn't touch anything RDS or CodePipeline share code with.

PR #2857, opened 2026-09-01
feat(batch): add UpdateComputeEnvironment and DeleteComputeEnvironment

### LINKED ISSUE #2293: [FEAT] batch: UpdateComputeEnvironment and DeleteComputeEnvironment
## Service

Batch

## API Action / Feature

`UpdateComputeEnvironment` and `DeleteComputeEnvironment`

## AWS Documentation

- https://docs.aws.amazon.com/batch/latest/APIReference/API_UpdateComputeEnvironment.html
- https://docs.aws.amazon.com/batch/latest/APIReference/API_DeleteComputeEnvironment.html

## Why is this needed?

`BatchController` exposes `createcomputeenvironment` and `describecomputeenvironments` but
neither `updatecomputeenvironment` nor `deletecomputeenvironment`, so a compute environment can
be created but never changed or removed.

As with job queues, this blocks teardown: `DeleteComputeEnvironment` requires the environment to
be `DISABLED`, so the AWS provider's delete path for `aws_batch_compute_environment` is
`UpdateComputeEnvironment(state=DISABLED)` → wait → `DeleteComputeEnvironment`. Both 404 today,
so `terraform destroy` cannot remove a compute environment and the resource is left permanently
in state.

`UpdateComputeEnvironment` is also the in-place update path for `min_vcpus`, `max_vcpus`,
`desired_vcpus`, `state` and `service_role` on that resource.

## Scope

Two endpoints mirroring the existing pair, plus the two matching `BatchService` methods.
`createComputeEnvironment`/`describeComputeEnvironments` (`BatchService:108,137`) are the shape
to copy.

Behaviour worth pinning:

- `DeleteComputeEnvironment` on an `ENABLED` environment returns `ClientException`, matching AWS.
- Deleting a compute environment still attached to a job queue returns `ClientException`.
- Both actions return an empty body on success; `UpdateComputeEnvironment` returns
  `computeEnvironmentName` and `computeEnvironmentArn`.
- Deleting an already-deleted environment should be tolerated so a partial destroy can be
  retried.

Filed separately from the job-queue equivalent so each change stays two endpoints.

- [ ] Yes
- [ ] No

### LINKED ISSUE #2803: batch: UpdateComputeEnvironment is not implemented
`UpdateComputeEnvironment` is not implemented anywhere under `services/batch` (no case/handler for it at all). `terraform-provider-aws` calls it when disabling a compute environment before deletion (part of `aws_batch_compute_environment`'s destroy lifecycle for `MANAGED` environments). Since Floci has no route for the action, the request falls through to an unhandled-route response — an HTML error page instead of a modeled JSON error — which breaks response deserialization on the client side:

```
Error: disabling Batch Compute Environment (floci-compat-batch): operation error Batch: UpdateComputeEnvironment,
       https response error StatusCode: 400, deserialization failed, failed to decode response body,
       invalid character '<' looking for beginning of value
```

Confirmed still missing on `main` — zero matches for `UpdateComputeEnvironment` under `services/batch`.

Blocks the Terraform/OpenTofu compatibility suites from including `aws_batch_compute_environment` (removed from both in #1834 to unblock the suite — `terraform destroy` failed on it every run; should be re-added once this lands).

pgermosen:
## Summary

Adds `UpdateComputeEnvironment` and `DeleteComputeEnvironment` to `BatchController`/`BatchService`, mirroring the existing `UpdateJobQueue`/`DeleteJobQueue` pair. Closes #2293, closes #2803.

- `DeleteComputeEnvironment` requires `DISABLED` state and rejects deletion while the environment is still referenced by any job queue's `computeEnvironmentOrder`, matching AWS.
- `UpdateComputeEnvironment`'s `computeResources` is a partial merge — only the fields the caller sends change (e.g. `maxvCpus`), the rest survive untouched — matching AWS's actual semantics, not a full replace.
- Both `UpdateComputeEnvironment` and `CreateComputeEnvironment` now validate `state` is `ENABLED` or `DISABLED`, so an invalid value can never be persisted and later block deletion.
- Both actions return an empty body on success except `UpdateComputeEnvironment`, which returns `computeEnvironmentName`/`computeEnvironmentArn`. Deleting an already-gone environment is tolerated (idempotent).

**Deliberately out of scope for this PR**: CloudFormation's stack-delete dispatcher has no case for `AWS::Batch::ComputeEnvironment`, `AWS::Batch::JobQueue`, or `AWS::Batch::JobDefinition` — all three silently no-op on stack teardown today. Fixing that properly means migrating all of Batch's CFN handling (create and delete, all three types) out of the legacy `CloudFormationResourceProvisioner` switch into a per-service provisioner, per `AGENTS.md`'s stated convention. That's a bigger, separate change than this PR's two endpoints, and will follow as its own PR.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the AWS Batch API reference (`API_UpdateComputeEnvironment.html`, `API_DeleteComputeEnvironment.html`) and the `batch-2016-08-10` service model: request/response field names (`computeEnvironment`, `computeResources.{minvCpus,maxvCpus,desiredvCpus}`, `serviceRole`, `state`), response shape, the DISABLED-before-delete precondition, and the `state` enum (`ENABLED`/`DISABLED`) — all stated explicitly in AWS's own docs.

## Checklist

### DIFF
--- a/docs/services/batch.md
+++ b/docs/services/batch.md
@@ -11,6 +11,8 @@ Floci Batch implements the AWS Batch control plane for local integration tests. 
 |---|---|---|
 | `CreateComputeEnvironment` | `POST /v1/createcomputeenvironment` | Store a local compute environment and return its ARN |
 | `DescribeComputeEnvironments` | `POST /v1/describecomputeenvironments` | Describe all or selected compute environments |
+| `UpdateComputeEnvironment` | `POST /v1/updatecomputeenvironment` | Update a compute environment's state, service role, and compute resources |
+| `DeleteComputeEnvironment` | `POST /v1/deletecomputeenvironment` | Delete a `DISABLED` compute environment not referenced by any job queue; deleting a missing one is a no-op |
 | `CreateJobQueue` | `POST /v1/createjobqueue` | Store a local job queue attached to compute environments |
 | `UpdateJobQueue` | `POST /v1/updatejobqueue` | Update a job queue's state, priority, and compute environment order |
 | `DeleteJobQueue` | `POST /v1/deletejobqueue` | Delete a `DISABLED` job queue; deleting a missing queue is a no-op |
--- a/src/main/java/io/github/hectorvent/floci/services/batch/BatchController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/batch/BatchController.java
@@ -50,6 +50,20 @@ public class BatchController {
     }
 
     @POST
+    @Path("/v1/updatecomputeenvironment")
+    public Response updateComputeEnvironment(@Context HttpHeaders headers, String body) {
+        return handle(headers, body, (request, region) ->
+                Response.ok(service.updateComputeEnvironment(request)).build());
+    }
+
+    @POST
+    @Path("/v1/deletecomputeenvironment")
+    public Response deleteComputeEnvironment(@Context HttpHeaders headers, String body) {
+        return handle(headers, body, (request, region) ->
+                Response.ok(service.deleteComputeEnvironment(request)).build());
+    }
+
+    @POST
     @Path("/v1/createjobqueue")
     public Response createJobQueue(@Context HttpHeaders headers, String body) {
         return handle(headers, body, (request, region) ->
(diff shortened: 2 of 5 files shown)


hectorvent (approved):
Thank you, this is a clean pair, and the description did a lot of the review's work for me.

I checked the partial merge claim against the model rather than the docs, and it holds for a good reason: [`ComputeResource`](https://github.com/boto/botocore/blob/develop/botocore/data/batch/2016-08-10/service-2.json) requires `type`, `maxvCpus` and `subnets`, while `ComputeResourceUpdate` requires nothing, and its `subnets` member documents that an empty list means no change. A full replace could not satisfy either, so `putAll` is right. The response members, the empty delete body, and `ClientException` as the only declared client error all match too.

**(follow-up, separate PR)** The job queue half still has the trap this closes for compute environments. `JQState` is the same enum, but `createJobQueue` and `updateJobQueue` never validate it, and `deleteJobQueue` has the same DISABLED precondition.

**(follow-up, separate PR)** Batch has no coverage in `compatibility-tests/`, so the green terraform suite is not yet guarding the destroy lifecycle from #2803.

No blockers from my side.

Issue #2854, opened 2026-09-01
[BUG] docs(cognito): UpdateResourceServer is implemented but missing from the Resource Servers table


bsaurusrex:
## Bug Description

`docs/services/cognito.md` lists four Resource Server actions. `UpdateResourceServer` is
implemented and working, but missing from the table, so the documentation understates
Floci's coverage.

## Steps to Reproduce

```bash
docker run -d --name floci-probe -p 4599:4566 floci/floci:latest
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1

POOL=$(aws --endpoint-url http://localhost:4599 cognito-idp create-user-pool \
  --pool-name rs-check --query 'UserPool.Id' --output text)
aws --endpoint-url http://localhost:4599 cognito-idp create-resource-server \
  --user-pool-id "$POOL" --identifier https://api.example.com --name API
aws --endpoint-url http://localhost:4599 cognito-idp update-resource-server \
  --user-pool-id "$POOL" --identifier https://api.example.com --name "API v2"
```

## Actual Behaviour

The update succeeds and returns the modified resource server, but the docs table reads:

| Action | Description |
|--------|-------------|
| CreateResourceServer | Registers a resource server and scopes for a user pool. |
| DescribeResourceServer | Returns a registered resource server. |
| ListResourceServers | Lists resource servers for a user pool. |
| DeleteResourceServer | Deletes a resource server from a user pool. |

`CognitoJsonHandler` dispatches `UpdateResourceServer` to `handleUpdateResourceServer`,
and `CognitoService.updateResourceServer` implements it.

## Expected Behaviour

An `UpdateResourceServer` row in the table.

## Notes

Cognito is on the `tools/docs/services.yaml` exclusion list (`non-tabular doc
(grouped/inline)`), so this table is hand-maintained and `make docs-check` does not catch
the omission.

## Environment

- Floci version: `floci/floci:latest`
- Deployment: Docker

Issue #2853, opened 2026-09-01
[BUG] cognito: user pool not-found message is "User pool not found", AWS names the pool


bsaurusrex:
## Bug Description

Every Cognito action that resolves a user pool reports a missing pool as
`User pool not found`. The live service returns a message that names the pool:
`User pool <poolId> does not exist.`

The error **code** is correct (`ResourceNotFoundException`); only the message diverges.
It matters because tests and tooling that assert on Cognito error text, and Floci's own
compatibility suites do assert on messages elsewhere, cannot match the real string.

## Steps to Reproduce

```bash
docker run -d --name floci-probe -p 4599:4566 floci/floci:latest
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1

aws --endpoint-url http://localhost:4599 cognito-idp describe-user-pool \
  --user-pool-id us-east-1_ZZZZZZZZZ
```

## Actual Behaviour

```
An error occurred (ResourceNotFoundException) when calling the DescribeUserPool operation: User pool not found
```

## Expected Behaviour

Measured against the live service (`DescribeIdentityProvider` on a pool id that does not
exist, same resolution path):

```
ResourceNotFoundException | http=400
    msg: User pool ap-southeast-1_ZZZZZZZZZ does not exist.
```

## Notes

The message comes from a single `orElseThrow` in `CognitoService.describeUserPool`, so
the fix is one string, but that helper is called by most actions in the service, so the
change is broad in blast radius even though it is narrow in diff, and it may require
updating existing assertions. Filing rather than folding it into an unrelated PR.

## Environment

- Floci version: `floci/floci:latest`
- Deployment: Docker

Issue #2852, opened 2026-09-01
[FEAT] cognito: no federated sign-in — /oauth2/authorize and /oauth2/idpresponse do not exist


bsaurusrex:
## Service

Cognito Identity Provider (`cognito-idp`)

## API Action / Feature

Federated sign-in (the identity provider *data plane*): `GET /oauth2/authorize` and
`GET /oauth2/idpresponse`, plus the auto-provisioning of a federated user and the
`identities` attribute that goes with it.

## AWS Documentation

- https://docs.aws.amazon.com/cognito/latest/developerguide/authorization-endpoint.html
- https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-social-idp.html
- https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-oidc-idp.html

## Reproduction

Against the released image (`floci/floci:latest`):

```bash
docker run -d --name floci-fed-probe -p 4599:4566 floci/floci:latest
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1
EP=http://localhost:4599

POOL_ID=$(aws --endpoint-url $EP cognito-idp create-user-pool \
  --pool-name fed-repro --query 'UserPool.Id' --output text)
CLIENT_ID=$(aws --endpoint-url $EP cognito-idp create-user-pool-client \
  --user-pool-id "$POOL_ID" --client-name fed-client \
  --query 'UserPoolClient.ClientId' --output text)

curl -s -o /dev/null -w '%{http_code}\n' \
  "$EP/oauth2/authorize?client_id=$CLIENT_ID&response_type=code&scope=openid&redirect_uri=https://app.example.com/callback"
curl -s -o /dev/null -w '%{http_code}\n' "$EP/oauth2/idpresponse?code=abc&state=xyz"
```

Both return `404`. The pool's own discovery document agrees that the authorization code
flow is not implemented:

```console
$ curl -s "$EP/$POOL_ID/.well-known/openid-configuration"
{
    "issuer": "http://localhost:4566/us-east-1_3ca44d25c",
    "jwks_uri": "http://localhost:4566/us-east-1_3ca44d25c/.well-known/jwks.json",
    "token_endpoint": "http://localhost:4566/cognito-idp/oauth2/token",
    "userinfo_endpoint": "http://localhost:4566/cognito-idp/oauth2/userInfo",
    ...
[...]

PR #2858, opened 2026-09-01
feat(cognito): implement identity provider configuration CRUD

### LINKED ISSUE #2851: [FEAT] cognito: identity provider configuration (Create/Describe/List/Update/DeleteIdentityProvider) is unsupported
## Service

## API Action / Feature

Identity provider configuration CRUD:

- `CreateIdentityProvider`
- `DescribeIdentityProvider`
- `ListIdentityProviders`
- `UpdateIdentityProvider`
- `DeleteIdentityProvider`

## AWS Documentation

- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateIdentityProvider.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DescribeIdentityProvider.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListIdentityProviders.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UpdateIdentityProvider.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DeleteIdentityProvider.html

## Reproduction

Against the released image (`floci/floci:latest`):

```bash
docker run -d --name floci-idp-probe -p 4599:4566 floci/floci:latest

POOL_ID=$(aws --endpoint-url http://localhost:4599 cognito-idp create-user-pool \
  --pool-name idp-repro --query 'UserPool.Id' --output text)

aws --endpoint-url http://localhost:4599 cognito-idp create-identity-provider \
  --user-pool-id "$POOL_ID" --provider-name TestOidc --provider-type OIDC \
  --provider-details client_id=x,client_secret=y,attributes_request_method=GET,\
oidc_issuer=https://issuer.example.com,authorize_scopes=openid
```

All five actions fail the same way:

```
An error occurred (UnsupportedOperation) when calling the CreateIdentityProvider operation: Operation CreateIdentityProvider is not supported.
An error occurred (UnsupportedOperation) when calling the DescribeIdentityProvider operation: Operation DescribeIdentityProvider is not supported.
An error occurred (UnsupportedOperation) when calling the ListIdentityProviders operation: Operation ListIdentityProviders is not supported.
An error occurred (UnsupportedOperation) when calling the UpdateIdentityProvider operation: Operation UpdateIdentityProvider is not supported.
An error occurred (UnsupportedOperation) when calling the DeleteIdentityProvider operation: Operation DeleteIdentityProvider is not supported.
```

## Why is this needed?

A user pool that federates to Google, Facebook or a generic OIDC/SAML provider
declares those providers as first-class infrastructure: `aws_cognito_identity_provider`
in Terraform, `AWS:Cognito:UserPoolIdentityProvider` in CloudFormation. Without these
five actions the stack cannot be applied against Floci at all, so a pool that is
federated in a real environment cannot be modelled locally even when nothing in the
local test path actually signs in through the provider.

These are pure configuration actions: apart from one validation fetch (below) the live
service stores a record and returns it. They sit naturally beside the existing
`ResourceServer` actions, which have the same five-action shape and the same
pool-scoped storage.

## Live API behaviour worth encoding

Measured against the live service. Several of these are not in the API reference and
are the kind of thing an implementation gets wrong by guessing:

- **`IdpIdentifiers` is echoed only when the request supplied the member.**
  `CreateIdentityProvider` and `UpdateIdentityProvider` omit it from the response
  otherwise, *whatever the stored value*. An update that omits the member on a provider
  storing `["alias"]` returns no `IdpIdentifiers` at all. `DescribeIdentityProvider`
  always returns it, as `[]` when empty.
- **Members the request omits are preserved, not cleared.** An explicitly empty map or
  list is what clears `AttributeMapping` / `IdpIdentifiers`.
- **`AttributeMapping` defaults to `{"username": "sub"}`** when the member is absent at
  create; an explicitly empty map is stored as given.
- **A duplicate provider name raises `DuplicateProviderException`**, not
  `ResourceConflictException`.
- **The not-found message differs by action:** `Describe`/`Delete` say
  `Identity provider <name> for tenantId <poolId> does not exist.` while `Update` says
  `Identity provider <name> in User Pool <poolId> does not exist.`
- **`ProviderType` is a closed enum:** `[Facebook, SAML, SignInWithApple, LoginWithAmazon, OIDC, Google]`.
- **`ListIdentityProviders` returns summaries only**: `ProviderName`, `ProviderType`,
  `CreationDate`, `LastModifiedDate`. It never returns `ProviderDetails`, which is where
  a provider's `client_secret` lives.

Two behaviours an emulator arguably should not reproduce, since both require calling a
third party:

- `CreateIdentityProvider` resolves an OIDC provider's `oidc_issuer` discovery document
  and rejects the call when it is unreachable, and enforces per-provider-type required
  keys (for OIDC: `clientId, authorizeScopes, oidcIssuer and attributesRequestMethod are
  all required idp details.`).
- AWS injects provider defaults such as `attributes_url_add_attributes` into the stored
  `ProviderDetails`.

## Scope

This issue covers the configuration actions only. Federated **sign-in** is a separate,
much larger gap: Floci has no `/oauth2/authorize` or `/oauth2/idpresponse` endpoint, so
a registered provider cannot currently be used to authenticate. That is filed separately.

- [x] Yes
- [ ] No

bsaurusrex:
## Summary

Implements Cognito identity provider configuration: `CreateIdentityProvider`,
`DescribeIdentityProvider`, `ListIdentityProviders`, `UpdateIdentityProvider` and
`DeleteIdentityProvider`. All five previously returned `UnsupportedOperation`, so a user
pool that federates to Google, Facebook or a generic OIDC/SAML provider could not be
provisioned against Floci at all.

Providers are stored beside the other pool sub-resources and the implementation mirrors
the existing `ResourceServer` shape, same keyed `StorageBackend`, same thin-controller
split, cases placed next to the resource server block.

Closes #2851

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the live Cognito API (AWS CLI 2.x / botocore, `ap-southeast-1`) on a
throwaway OIDC provider, created and deleted for the purpose. API reference:
[CreateIdentityProvider](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateIdentityProvider.html),
[DescribeIdentityProvider](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DescribeIdentityProvider.html),
[ListIdentityProviders](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListIdentityProviders.html),
[UpdateIdentityProvider](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UpdateIdentityProvider.html),
[DeleteIdentityProvider](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DeleteIdentityProvider.html).

Six behaviours are not in the reference and are the ones a plausible implementation gets
wrong. Each is pinned by a test.

**1. `IdpIdentifiers` is echoed only when the request supplied the member, regardless of
what is stored.** This is the one that matters most: a provider storing `["alias-keep"]`,
updated without the member, returns *no* `IdpIdentifiers` at all.

```console
--- create with NON-default mapping + identifiers
      AttributeMapping: {'email': 'email', 'username': 'sub'}
      IdpIdentifiers:   ['alias-keep']

--- update omitting BOTH members
      AttributeMapping: {'email': 'email', 'username': 'sub'}
      IdpIdentifiers:   <ABSENT>

--- describe -> preserved or cleared?
      AttributeMapping: {'email': 'email', 'username': 'sub'}
      IdpIdentifiers:   ['alias-keep']
```

`DescribeIdentityProvider` always returns it, as `[]` when empty.

**2. Members a request omits are preserved, not cleared**, the same trace above. An
explicitly empty map or list is what clears them:

```console
--- B update IdpIdentifiers=[]
      IdpIdentifiers:   []
--- C update AttributeMapping={}
      AttributeMapping: {}
      IdpIdentifiers:   <ABSENT>
```

**3. `AttributeMapping` defaults to `{"username": "sub"}`** when the member is absent at
create; an explicitly empty map is stored as given.

**4. Duplicate provider name → `DuplicateProviderException`** (not
`ResourceConflictException`, which is what the sibling `ResourceServer` code uses):

```console
--- Q3 duplicate ProviderName: DuplicateProviderException | http=400
    msg: FlociProbeOidc already exists for tenant <poolId>.
```

**5. The not-found message differs by action.** `Describe`/`Delete`:

```console
ResourceNotFoundException | msg: Identity provider FlociProbeOidc for tenantId <poolId> does not exist.
```

`Update`:

```console
ResourceNotFoundException | msg: Identity provider FlociProbeOidc in User Pool <poolId> does not exist.
```

**6. `ProviderType` is a closed enum**, with the standard validation-error format:

```console
--- Q4 bad ProviderType: InvalidParameterException | http=400
    msg: 1 validation error detected: Value 'NotARealType' at 'providerType' failed to satisfy constraint: Member must satisfy enum value set: [Facebook, SAML, SignInWithApple, LoginWithAmazon, OIDC, Google]
```

`ListIdentityProviders` returns summaries only: `ProviderName`, `ProviderType`,
`CreationDate`, `LastModifiedDate`, and never `ProviderDetails`, which is where a
provider's `client_secret` lives. There is a test asserting the absence, since leaking
credentials through a list call would be the costly way to get this wrong.

Two live behaviours are omitted on purpose, both because they require calling a third
party. Both are documented in `docs/services/cognito.md`:

- **Create-time OIDC discovery.** AWS resolves `oidc_issuer` while handling
  `CreateIdentityProvider` and fails when it is unreachable, my first probe died exactly
  this way (`InvalidParameterException: probe.invalid`). An emulator making outbound HTTP
  during a control-plane call seemed clearly wrong, so `ProviderDetails` is stored
  opaquely. That also means the per-provider-type required-key check is not implemented;
  the live message for OIDC is `clientId, authorizeScopes, oidcIssuer and
  attributesRequestMethod are all required idp details.`, recorded here for whoever picks
  it up.
- **Injected provider defaults.** AWS adds keys such as `attributes_url_add_attributes`
  to stored `ProviderDetails`. Floci returns only what was supplied, which is
  self-consistent for round-tripping, since a client reads back exactly what it wrote.

### Out of scope

Federated **sign-in** is not part of this PR and is much larger: Floci has no
`/oauth2/authorize` or `/oauth2/idpresponse` route, and its own discovery document
advertises `response_types_supported: []` with `client_credentials` as the only grant. A
registered provider therefore cannot be used to authenticate. Filed separately as #2852
with a proposed design, as an issue rather than a PR because the approach is a judgement
call for a maintainer.

Two unrelated gaps found while measuring, filed rather than folded in: #2853 (user pool
not-found message does not name the pool) and #2854 (`UpdateResourceServer` implemented
but missing from the docs table).

## Testing

New integration test `CognitoIdentityProviderIntegrationTest` (12 cases) pins each measured
behaviour above, creates its own pool and providers and deletes all of them, so it leaves
no state for a sibling class.

- Cognito package: **419 tests, 0 failures**
- Full suite: **14,799 tests, 0 failures, 0 errors**
- `make docs-check`: clean

SDK compatibility tests added for the response shapes, since the `IdpIdentifiers`
asymmetry is exactly the kind of thing a generated SDK deserializer surfaces differently
from a raw JSON assertion:

- `compatibility-tests/sdk-test-node/tests/cognito-identity-provider.test.ts` (new)
- `compatibility-tests/sdk-test-python/tests/test_cognito.py`, new
  `TestCognitoIdentityProvider` class

Both suites were run inside Floci's network against a probe **built from this branch**,
with a control probe **built from `main`** on the same network and the same
`FLOCI_TLS_ENABLED=true` configuration:

| | branch probe | `main` control |
|---|---|---|
| Node, full suite | **2 failed / 476 passed** | 6 failed / 472 passed |
| Python, full suite | **326 passed / 0 failed** | 6 failed (the new class) |

The two Node failures are `apigatewayv2-websocket-dataplane` and
`apigatewayv2-websocket-tls` "Chat-style broadcast", and they fail **identically on the
`main` control**, pre-existing, not this change. The only other difference between the
two columns is the new identity provider tests, which fail on `main` with nothing but

```
UnsupportedOperation: Operation CreateIdentityProvider is not supported.
```

and pass on the branch. That is the negative control: the new tests exercise exactly the
new code and nothing else.

One thing the SDK run caught that the raw JSON integration test could not: the JS SDK maps
the wire `__type` onto `error.name` and puts the AWS text in `message`, so a
`rejects.toThrow(/DuplicateProvider/)` matcher never fires. Fixed in the second commit by
asserting the name.

The Go, Java and AWS CLI suites were left alone: none of them has existing identity
provider coverage to extend, and adding a fourth and fifth copy of the same assertions
seemed like noise rather than signal. Happy to add them if you would rather have the
breadth.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-node/tests/cognito-identity-provider.test.ts
+++ b/compatibility-tests/sdk-test-node/tests/cognito-identity-provider.test.ts
@@ -148,6 +148,8 @@ describe('Cognito identity providers', () => {
       })
     );
 
+    // The SDK maps the wire `__type` onto `error.name` and puts the AWS text in
+    // `message`, so the exception has to be matched on the name.
     await expect(
       cognito.send(
         new CreateIdentityProviderCommand({
@@ -157,7 +159,7 @@ describe('Cognito identity providers', () => {
           ProviderDetails: { ...OIDC_DETAILS },
         })
       )
-    ).rejects.toThrow(/DuplicateProvider/);
+    ).rejects.toMatchObject({ name: 'DuplicateProviderException' });
 
     await expect(
       cognito.send(
@@ -168,7 +170,10 @@ describe('Cognito identity providers', () => {
           ProviderDetails: { client_id: 'x' },
         })
       )
-    ).rejects.toThrow(/enum value set/);
+    ).rejects.toMatchObject({
+      name: 'InvalidParameterException',
+      message: expect.stringContaining('Member must satisfy enum value set'),
+    });
 
     await cognito.send(
       new DeleteIdentityProviderCommand({
--- a/compatibility-tests/sdk-test-node/tests/cognito-identity-provider.test.ts
+++ b/compatibility-tests/sdk-test-node/tests/cognito-identity-provider.test.ts
@@ -1,0 +1,180 @@
+/**
+ * Cognito identity provider configuration compatibility tests.
+ *
+ * The response shapes asserted here were measured against the live Cognito API:
+ * `IdpIdentifiers` is echoed by CreateIdentityProvider and UpdateIdentityProvider
+ * only when the request supplied the member, whatever the stored value, while
+ * DescribeIdentityProvider always returns it. Members a request omits are
+ * preserved rather than cleared.
+ */
+
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import {
+  CognitoIdentityProviderClient,
+  CreateUserPoolCommand,
+  DeleteUserPoolCommand,
+  CreateIdentityProviderCommand,
+  DescribeIdentityProviderCommand,
+  ListIdentityProvidersCommand,
+  UpdateIdentityProviderCommand,
+  DeleteIdentityProviderCommand,
+  IdentityProviderTypeType,
+} from '@aws-sdk/client-cognito-identity-provider';
+import { makeClient, uniqueName } from './setup';
+
+const OIDC_DETAILS = {
+  client_id: 'node-compat-client',
+  client_secret: 'node-compat-secret',
+  attributes_request_method: 'GET',
+  oidc_issuer: 'https://issuer.example.com',
+  authorize_scopes: 'openid',
+};
+
+describe('Cognito identity providers', () => {
+  let cognito: CognitoIdentityProviderClient;
+  let poolId: string;
+
+  beforeAll(async () => {
+    cognito = makeClient(CognitoIdentityProviderClient);
+    const pool = await cognito.send(
+      new CreateUserPoolCommand({ PoolName: uniqueName('idp-pool') })
+    );
+    poolId = pool.UserPool!.Id!;
+  });
+
+  afterAll(async () => {
+    if (poolId) {
+      await cognito.send(new DeleteUserPoolCommand({ UserPoolId: poolId }));
+    }
+  });
+
+  it('defaults AttributeMapping and omits IdpIdentifiers on create', async () => {
+    const created = await cognito.send(
+      new CreateIdentityProviderCommand({
+        UserPoolId: poolId,
+        ProviderName: 'NodeOidc',
+        ProviderType: IdentityProviderTypeType.OIDC,
+        ProviderDetails: { ...OIDC_DETAILS },
+      })
+    );
+
+    expect(created.IdentityProvider!.AttributeMapping).toEqual({ username: 'sub' });
+    expect(created.IdentityProvider!.IdpIdentifiers).toBeUndefined();
+
+    const described = await cognito.send(
+      new DescribeIdentityProviderCommand({ UserPoolId: poolId, ProviderName: 'NodeOidc' })
+    );
+    expect(described.IdentityProvider!.IdpIdentifiers).toEqual([]);
+    expect(described.IdentityProvider!.ProviderDetails!.client_id).toBe('node-compat-client');
+
+    await cognito.send(
+      new DeleteIdentityProviderCommand({ UserPoolId: poolId, ProviderName: 'NodeOidc' })
+    );
+  });
+
+  it('preserves members the update request omits', async () => {
+    await cognito.send(
+      new CreateIdentityProviderCommand({
+        UserPoolId: poolId,
+        ProviderName: 'NodeOidcAliased',
+        ProviderType: IdentityProviderTypeType.OIDC,
+        ProviderDetails: { ...OIDC_DETAILS },
+        AttributeMapping: { email: 'email', username: 'sub' },
+        IdpIdentifiers: ['node-alias'],
+      })
+    );
+
+    const updated = await cognito.send(
+      new UpdateIdentityProviderCommand({
+        UserPoolId: poolId,
+        ProviderName: 'NodeOidcAliased',
+        ProviderDetails: { ...OIDC_DETAILS },
+      })
+    );
+    expect(updated.IdentityProvider!.IdpIdentifiers).toBeUndefined();
+
+    const described = await cognito.send(
+      new DescribeIdentityProviderCommand({
+        UserPoolId: poolId,
+        ProviderName: 'NodeOidcAliased',
+      })
+    );
+    expect(described.IdentityProvider!.IdpIdentifiers).toEqual(['node-alias']);
+    expect(described.IdentityProvider!.AttributeMapping).toEqual({
+      email: 'email',
+      username: 'sub',
+    });
+
+    await cognito.send(
+      new DeleteIdentityProviderCommand({
+        UserPoolId: poolId,
+        ProviderName: 'NodeOidcAliased',
+      })
+    );
+  });
+
+  it('lists providers as summaries without provider details', async () => {
+    await cognito.send(
+      new CreateIdentityProviderCommand({
+        UserPoolId: poolId,
+        ProviderName: 'NodeOidcListed',
+        ProviderType: IdentityProviderTypeType.OIDC,
+        ProviderDetails: { ...OIDC_DETAILS },
+      })
+    );
+
+    const listed = await cognito.send(new ListIdentityProvidersCommand({ UserPoolId: poolId }));
+
+    expect(listed.Providers).toHaveLength(1);
+    expect(listed.Providers![0].ProviderName).toBe('NodeOidcListed');
+    expect(listed.Providers![0].ProviderType).toBe('OIDC');
+    expect(listed.Providers![0].CreationDate).toBeInstanceOf(Date);
+    expect(listed.Providers![0]).not.toHaveProperty('ProviderDetails');
+
+    await cognito.send(
+      new DeleteIdentityProviderCommand({ UserPoolId: poolId, ProviderName: 'NodeOidcListed' })
+    );
+    const empty = await cognito.send(new ListIdentityProvidersCommand({ UserPoolId: poolId }));
+    expect(empty.Providers).toEqual([]);
+  });
+
+  it('rejects a duplicate provider name and an unknown provider type', async () => {
+    await cognito.send(
+      new CreateIdentityProviderCommand({
+        UserPoolId: poolId,
+        ProviderName: 'NodeOidcDuplicate',
+        ProviderType: IdentityProviderTypeType.OIDC,
+        ProviderDetails: { ...OIDC_DETAILS },
+      })
+    );
+
+    await expect(
+      cognito.send(
+        new CreateIdentityProviderCommand({
+          UserPoolId: poolId,
+          ProviderName: 'NodeOidcDuplicate',
+          ProviderType: IdentityProviderTypeType.OIDC,
+          ProviderDetails: { ...OIDC_DETAILS },
+        })
+      )
+    ).rejects.toThrow(/DuplicateProvider/);
+
+    await expect(
+      cognito.send(
+        new CreateIdentityProviderCommand({
+          UserPoolId: poolId,
+          ProviderName: 'NodeBogus',
+          ProviderType: 'NotARealType' as IdentityProviderTypeType,
+          ProviderDetails: { client_id: 'x' },
+        })
+      )
+    ).rejects.toThrow(/enum value set/);
+
+    await cognito.send(
+      new DeleteIdentityProviderCommand({
+        UserPoolId: poolId,
+        ProviderName: 'NodeOidcDuplicate',
+      })
+    );
+  });
+});
--- a/compatibility-tests/sdk-test-python/tests/test_cognito.py
+++ b/compatibility-tests/sdk-test-python/tests/test_cognito.py
@@ -219,3 +219,132 @@ class TestCognitoDescribeUserPoolStandardAttributes:
             assert sub["Mutable"] is False
         finally:
             cognito_client.delete_user_pool(UserPoolId=pool_id)
+
+
+class TestCognitoIdentityProvider:
+    """Test Cognito identity provider configuration operations.
+
+    The response shapes asserted here were measured against the live Cognito API.
+    ``IdpIdentifiers`` is echoed by CreateIdentityProvider only when the request
+    supplied it, while DescribeIdentityProvider always returns it.
+    """
+
+    OIDC_DETAILS = {
+        "client_id": "pytest-client",
+        "client_secret": "pytest-secret",
+        "attributes_request_method": "GET",
+        "oidc_issuer": "https://issuer.example.com",
+        "authorize_scopes": "openid",
+    }
+
+    @pytest.fixture
+    def pool_id(self, cognito_client, unique_name):
+        response = cognito_client.create_user_pool(PoolName=f"pytest-idp-pool-{unique_name}")
+        pool_id = response["UserPool"]["Id"]
+        yield pool_id
+        cognito_client.delete_user_pool(UserPoolId=pool_id)
+
+    def test_create_defaults_attribute_mapping_and_omits_idp_identifiers(
+        self, cognito_client, pool_id
+    ):
+        """Create defaults AttributeMapping and omits IdpIdentifiers when not supplied."""
+        provider = cognito_client.create_identity_provider(
+            UserPoolId=pool_id,
+            ProviderName="PytestOidc",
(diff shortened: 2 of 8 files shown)


bsaurusrex:
Thanks — both of the first-round findings were real, and I reproduced each before changing anything.

**Pool deletion leaving providers behind — fixed** (18cc483).

Reproduced first. Pool ids are caller-choosable via the `floci:override-id` reserved tag, so a pool recreated on a pinned id inherited the dead pool's providers, `client_secret` and all:

```
$ aws cognito-idp describe-identity-provider --user-pool-id us-east-1_orphan01 --provider-name OrphanIdp
{ "client_id": "orphan-id", "client_secret": "orphan-secret", ... }
```

`deleteUserPool` now removes providers the way it already removes groups. Regression test `deletingAPoolRemovesItsIdentityProviders`; with the fix reverted it fails with `expected: <0> but was: <1>`.

I checked whether this was the instance or the class: on `main`, `deleteUserPool` also leaves **users** and **resource servers** behind, reproducible the same way. I have not widened this PR to cover them — the user case in particular is a behaviour change with its own blast radius — and filed #2864 with the executed reproduction instead.

**Shared-instance mutation — fixed** (18cc483).

`InMemoryStorage.get()` returns `Optional.ofNullable(store.get(key))`, i.e. the live object, so this was structurally right. `updateIdentityProvider` now writes a copy. Pinned by `updateIdentityProviderDoesNotMutateAlreadyReturnedInstances`; reverted, it fails with `expected: <before> but was: <after>`.

**Remaining unsynchronized get-copy-put — respectfully pushing back, with a measurement.**

I went to check this against the live service rather than reason about it, using two concurrent partial updates released from a barrier — A adding an `AttributeMapping` entry, B setting `IdpIdentifiers` — five times against a real user pool:

```
run 1: A_survived=True  B_survived=True
run 2: A_survived=True  B_survived=True
run 3: A_survived=True  B_survived=True
run 4: A_survived=False B_survived=True   <-- AWS discarded A's change
run 5: A_survived=True  B_survived=True
```

AWS discarded a successful writer's change in 1 of 5 runs. `UpdateIdentityProvider` exposes no version token, ETag or conditional parameter, and the service does not serialize these internally — so last-writer-wins under genuine overlap *is* the AWS behaviour, and Floci reproducing it is fidelity rather than a defect.

Locking here would also be narrow to the point of being misleading: `updateResourceServer`, `updateUserPoolClient` and `adminUpdateUserAttributes` all use the same get-mutate-put, so a lock on identity providers alone would close this thread while changing nothing about the property it claims to protect. If store-level atomicity is wanted it should be a deliberate service-wide change, and I am happy to open an issue for that — just say the word.

Leaving the code as it is on this one and leaving the thread open for you to close.

pgermosen (commented):
@bsaurusrex the measured-behavior work here is excellent as usual, and the first two review findings are genuinely fixed — I read `deleteUserPool` and `updateIdentityProvider` directly rather than take the commit messages on faith, and pool deletion does clean up its providers now, and the update path does work off a defensive copy instead of the live cached object.

But greptile's latest pass (against the current HEAD, `18cc4837`) found a third issue that commit didn't cover, and its own summary says outright the PR isn't safe to merge yet because of it. That comment has no reply and no follow-up commit. I read `updateIdentityProvider` myself to confirm it's real rather than just relay the finding: there's no synchronization around the read-copy-mutate-write sequence, so two concurrent updates touching different optional members both read the same starting state, and whichever `put()` lands second silently reverts the other request's already-applied change. The defensive copy stops the two requests from corrupting each other's in-memory objects mid-flight, but doesn't touch the lost-update race on the persisted value itself. `CognitoService.java` already uses `synchronized (pool)` blocks and a dedicated `identityLinkLock` elsewhere for exactly this class of problem, so there's a precedent here to follow rather than invent.

Everything else — the IdpIdentifiers omission asymmetry, the preserve-vs-clear semantics, the duplicate-name and not-found error codes, the ListIdentityProviders credential-leak test, the SDK compat coverage catching the `__type`/`error.name` mapping quirk — is solid. This one open item is the reason it's not there yet.

bsaurusrex:
Confirmed and fixed in 7f5c57e, and I widened it to the whole class of race rather than the one call.

First I measured what the update is supposed to do, because the fix depends on it. Live Cognito, `ap-southeast-1`, provider created with `AttributeMapping`, `IdpIdentifiers` and `authorize_scopes`, then updated supplying only `ProviderDetails`:

```
before  AttributeMapping {'email': 'email', 'username': 'sub'}   IdpIdentifiers ['probe-idp-1']   scopes 'openid email'
after   AttributeMapping {'email': 'email', 'username': 'sub'}   IdpIdentifiers ['probe-idp-1']   scopes 'openid'
```

So AWS merges omitted members rather than clearing them, which is what this branch already implements, and it applies that merge atomically. That is the part that was missing here: two overlapping updates copied the same stored version, so the later `put` restored the other's omitted members to their stale values.

`updateIdentityProvider` was not the only path with that shape. `createIdentityProvider` does check-then-put for the duplicate check and `deleteIdentityProvider` does check-then-act, so both can interleave the same way. All three now share one `identityProviderLock`, following the existing `identityLinkLock` idiom in this class rather than introducing a new mechanism.

The regression test runs two writers against the same provider, one supplying only `AttributeMapping` and the other only `IdpIdentifiers`, released together through a `CyclicBarrier` for 25 rounds, and asserts both writes survive. Reverting the lock fails it on the first round:

```
round 0: the concurrent IdpIdentifiers update discarded the AttributeMapping write
```

With the lock in place the full `services.cognito` package is green: 422 tests, 0 failures, 0 errors.

bsaurusrex:
Good catch, and it is the same class as the original finding rather than a separate one. Fixed in the commit above.

`DeleteUserPool` clears a pool's providers with a scan and delete that ran outside `identityProviderLock`, so a `CreateIdentityProvider` or `UpdateIdentityProvider` interleaving with the cascade could write a provider back for a pool that is going away. The cascade now takes the same lock, alongside the `poolStore.delete` it precedes, so the whole teardown is atomic with respect to provider mutations.

That was the last unguarded path in the set: create, update, delete and the pool cascade all serialise on one lock now.

`services.cognito` stays green: 422 tests, 0 failures, 0 errors.

bsaurusrex:
Real, and fixed in the commit above.

`createIdentityProvider` called `describeUserPool` before taking the lock and never rechecked, so the interleaving you describe holds: the cascade in `DeleteUserPool` removes the pool's providers and the pool, then the waiting create acquires the lock and writes a provider under a pool id that no longer exists. It now reasserts the pool inside the lock, immediately before the duplicate check and the write.

`updateIdentityProvider` does not need the same treatment, and I would rather say why than add a symmetrical recheck that does nothing. The cascade deletes the provider rows before it deletes the pool, so an update that acquires the lock afterwards finds nothing and throws `ResourceNotFoundException` from its own lookup. Create was the only path that writes a record with no prior existence to miss.

The outer `describeUserPool` stays where it is. Moving it inside the lock would reorder it after the `ProviderName` and `ProviderType` validations, which changes which error an ordinary single-request caller sees when a request is malformed and names a missing pool.

`services.cognito`: 422 tests, 0 failures, 0 errors.

bsaurusrex:
@pgermosen ready for another look. Your lost-update finding was real, and chasing the class of it rather than the one call turned up three more unguarded paths, all fixed since your review:

| commit | what it closes |
|---|---|
| `7f5c57e` | `identityProviderLock` around create, update and delete. Update was the reported race; create's check-then-put and delete's check-then-act interleave the same way |
| `596ba4b5` | the `DeleteUserPool` provider cascade takes the same lock, so a create or update cannot write a provider back under a pool that is going away |
| `93135ad1` | `createIdentityProvider` reasserts the pool inside the lock, closing the window where the cascade completes between the outer `describeUserPool` and the write |

I followed the existing `identityLinkLock` idiom in `CognitoService` rather than introducing a new mechanism, as you suggested.

Before fixing the update I measured what it is supposed to do, since the fix depends on it. Live Cognito in `ap-southeast-1`, provider created with `AttributeMapping`, `IdpIdentifiers` and `authorize_scopes`, then updated supplying only `ProviderDetails`:

```
before  AttributeMapping {'email': 'email', 'username': 'sub'}   IdpIdentifiers ['probe-idp-1']   scopes 'openid email'
after   AttributeMapping {'email': 'email', 'username': 'sub'}   IdpIdentifiers ['probe-idp-1']   scopes 'openid'
```

AWS merges omitted members and applies that merge atomically. The merge semantics were already right on this branch; atomicity was the missing half.

The regression test runs two writers against the same provider, one supplying only `AttributeMapping` and the other only `IdpIdentifiers`, released together through a `CyclicBarrier` for 25 rounds. Reverting the lock fails it on the first round:

```
round 0: the concurrent IdpIdentifiers update discarded the AttributeMapping write
```

One deliberate omission: `updateIdentityProvider` does not recheck the pool inside the lock. The cascade deletes the provider rows before the pool, so an update acquiring the lock afterwards finds nothing and throws `ResourceNotFoundException` from its own lookup. Create was the only path that writes a record with no prior existence to miss. I would rather say that than add a symmetrical recheck that does nothing.

`services.cognito`: 422 tests, 0 failures, 0 errors. Full CI green on `93135ad1`, all four shards plus the native build and compat jobs. Merges clean against `5465476` as of 2026-09-02 12:17 UTC.

pgermosen (approved):
This is a really thorough fix. You didn't just patch the one call I flagged, you generalized it to the whole class of race, createIdentityProvider's check-then-write and the deleteUserPool cascade both needed the same guard and you found them. I traced the lock interactions myself, the cascade deleting provider rows before the pool means a losing update just gets a correct ResourceNotFoundException from its own lookup rather than corrupting anything, so leaving that recheck out was the right call, not an oversight.

And the concurrentUpdatesOfDifferentMembersBothSurvive test is exactly the kind of regression test this bug needed, two real writers on different members through a CyclicBarrier over 25 rounds against the actual endpoint, not a single-threaded stand-in. Nice work chasing this down properly.

PR #2860, opened 2026-09-01
fix(opensearch): include CognitoOptions in domain status

### LINKED ISSUE #2801: opensearch: DescribeDomain response omits CognitoOptions, crashes terraform-provider-aws
`terraform-provider-aws`'s `resourceDomainRead` for `aws_opensearch_domain` unconditionally flattens the `CognitoOptions` field from the `DescribeDomain` response. Floci's OpenSearch `DescribeDomain` response never sets `CognitoOptions` at all (nil), and the provider dereferences it without a nil check, crashing the provider process:

```
github.com/hashicorp/terraform-provider-aws/internal/service/opensearch.flattenCognitoOptions(0x0)
    .../internal/service/opensearch/flex.go:120
github.com/hashicorp/terraform-provider-aws/internal/service/opensearch.resourceDomainRead(...)
    .../internal/service/opensearch/domain.go:932
```

Confirmed still missing on `main` — no `CognitoOptions` anywhere under `services/opensearch`. Fix direction: always include a `CognitoOptions` object in the `DescribeDomain`/`DomainStatus` response (e.g. `{"Enabled": false}` when Cognito auth isn't configured), matching real AWS's shape.

Blocks the Terraform/OpenTofu compatibility suites from including `aws_opensearch_domain` (removed from both in #1834 to unblock the suite; should be re-added once this lands).

Origin: this is the "provider panic... may point at a real response-shape bug" abanna flagged as worth investigating on #1834 without a confirmed root cause at the time; now confirmed via the actual crash stack.

7487:
## Summary

- Return `DomainStatus.CognitoOptions.Enabled = false` when Cognito authentication is not configured.
- Keep every response using the shared `DomainStatus` serializer compatible with clients that expect this AWS response object.

Fixes #2801

## Root cause

The shared `DomainStatus` serializer omitted `CognitoOptions` entirely for domains without Cognito authentication. Terraform AWS Provider reads that object unconditionally, so the missing field caused a nil-pointer panic during refresh.

## Change

Add the disabled default object once in `toDomainStatusNode` and cover the public `DescribeDomain` response with a regression assertion.

## Verification

- TDD red: the new assertion failed with `Expected: false, Actual: null` before the fix.
- `./mvnw -Dtest='OpenSearch*Test' test`: 64 tests passed.
- Full suite in the repository's four CI-style shards: 14,822 tests passed, 0 failures, 0 errors, 9 skipped.
- `./mvnw -q -DskipTests test-compile`: passed with Java 25.
- `./mvnw clean package -DskipTests -B`: passed.
- `make docs-check`: passed.
- `git diff --check`: passed.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/opensearch/OpenSearchController.java
@@ -498,6 +498,7 @@ public class OpenSearchController {
         }
         node.set("ClusterConfig", toClusterConfigNode(domain.getClusterConfig()));
         node.set("EBSOptions", toEbsOptionsNode(domain.getEbsOptions()));
+        node.putObject("CognitoOptions").put("Enabled", false);
         if (domain.getVpcOptions() != null) {
             node.set("VPCOptions", toVpcOptionsNode(domain.getVpcOptions()));
         }
--- a/src/test/java/io/github/hectorvent/floci/services/opensearch/OpenSearchIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/opensearch/OpenSearchIntegrationTest.java
@@ -76,7 +76,8 @@ class OpenSearchIntegrationTest {
             .body("DomainStatus.EngineVersion", equalTo("OpenSearch_2.11"))
             .body("DomainStatus.ClusterConfig.InstanceType", equalTo("m5.large.search"))
             .body("DomainStatus.ClusterConfig.InstanceCount", equalTo(1))
-            .body("DomainStatus.EBSOptions.EBSEnabled", equalTo(true));
+            .body("DomainStatus.EBSOptions.EBSEnabled", equalTo(true))
+            .body("DomainStatus.CognitoOptions.Enabled", equalTo(false));
     }
 
     @Test


hectorvent:
Thanks for this, and for the root cause, which turns out to be verifiable end to end. `flattenCognitoOptions` in terraform-provider-aws dereferences `c.Enabled` with no nil guard on `c` ([flex.go](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/opensearch/flex.go)), and the read path calls it unconditionally, so an omitted `CognitoOptions` really is a provider panic rather than just a missing field.

The model agrees with the fix too: `CognitoOptions` is a declared member of `DomainStatus` and is not in its required set ([service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/opensearch/2021-01-01/service-2.json)), so returning `{Enabled: false}` for a domain without Cognito auth is what AWS does.

Four lines, one of them an assertion that fails without the fix, and it lands in the shared serializer so every `DomainStatus` response benefits. No blockers from my side.

hectorvent (approved):
Verified the motivation in terraform-provider-aws (flattenCognitoOptions derefs c.Enabled with no nil guard, called unconditionally from the read path) and the shape against botocore opensearch/2021-01-01. Closes #2801.

PR #2867, opened 2026-09-01
feat(cognito): implement log delivery configuration

### LINKED ISSUE #2855: [FEAT] cognito: SetLogDeliveryConfiguration / GetLogDeliveryConfiguration are unsupported
## Service

## API Action / Feature

`SetLogDeliveryConfiguration`, `GetLogDeliveryConfiguration`

## AWS Documentation

- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_SetLogDeliveryConfiguration.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_GetLogDeliveryConfiguration.html

## Reproduction

```bash

POOL=$(aws --endpoint-url http://localhost:4599 cognito-idp create-user-pool \
  --pool-name gap-check --query 'UserPool.Id' --output text)
aws --endpoint-url http://localhost:4599 cognito-idp set-log-delivery-configuration \
  --user-pool-id "$POOL" \
  --log-configurations '[{"LogLevel":"ERROR","EventSource":"userNotification"}]'
```

```
An error occurred (UnsupportedOperation) when calling the SetLogDeliveryConfiguration operation: Operation SetLogDeliveryConfiguration is not supported.
```

## Why is this needed?

`aws_cognito_log_delivery_configuration` is a standard Terraform resource for routing user
pool notification and sign-in logs to CloudWatch Logs, Firehose or S3. Applying a config
that includes one against Floci fails the whole apply:

```
Error: creating Cognito IDP (Identity Provider) Log Delivery Configuration
  with aws_cognito_log_delivery_configuration.example

ID: "ap-southeast-1_68e19d8e6"
Cause: operation error Cognito Identity Provider: SetLogDeliveryConfiguration,
api error UnsupportedOperation: Operation SetLogDeliveryConfiguration is not supported.
```

The pair is a natural store-and-return-it pair on the user pool, like the other pool-scoped
configuration actions. Notably the log *destination* need not do anything, the value is in
letting the resource be created, read back and destroyed so a stack applies and refreshes
cleanly.

- [x] Yes
- [ ] No

bsaurusrex:
## Summary

Implements `SetLogDeliveryConfiguration` and `GetLogDeliveryConfiguration`. Both returned
`UnsupportedOperation`, so a stack declaring `aws_cognito_log_delivery_configuration`
could not be applied against Floci. The apply fails outright:

```
Error: creating Cognito IDP (Identity Provider) Log Delivery Configuration
Cause: operation error Cognito Identity Provider: SetLogDeliveryConfiguration,
api error UnsupportedOperation: Operation SetLogDeliveryConfiguration is not supported.
```

The configuration is stored on the user pool, so this adds no new storage backend.

Closes #2855

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Measured against the live Cognito API, on a real user pool whose existing configuration was
captured first and restored afterwards. Reference:
[SetLogDeliveryConfiguration](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_SetLogDeliveryConfiguration.html),
[GetLogDeliveryConfiguration](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_GetLogDeliveryConfiguration.html).

**Enum sets, and AWS's 1-based member path**, note `logConfigurations.1.member`, not `.0.`:

```console
--- invalid LogLevel: InvalidParameterException | http=400
    msg: 1 validation error detected: Value 'TRACE' at 'logConfigurations.1.member.logLevel' failed to satisfy constraint: Member must satisfy enum value set: [ERROR, INFO]

--- invalid EventSource: InvalidParameterException | http=400
    msg: 1 validation error detected: Value 'notAThing' at 'logConfigurations.1.member.eventSource' failed to satisfy constraint: Member must satisfy enum value set: [userAuthEvents, userNotification]
```

**Every event source must name a destination**, and the message enumerates the ones that do
not:

```console
--- config with no destination: InvalidParameterException | http=400
    msg: Request validation Failed. Following event sources in request have no destination: [userNotification].
```

**An empty list is accepted and clears the configuration**, and `Get` still returns the
member:

```console
--- empty LogConfigurations: OK
{ "LogDeliveryConfiguration": { "UserPoolId": "<poolId>", "LogConfigurations": [] } }
```

**Unknown pool** → `ResourceNotFoundException`, `User pool <poolId> does not exist.` (Floci's
message differs here, but that is pre-existing across the whole service and is filed
separately as #2853 rather than changed in this PR.)

- **The pricing-tier gate.** AWS refuses `userAuthEvents` on an `ESSENTIALS`-tier pool with
  `FeatureUnavailableInTierException: The following feature is not available for the
  ESSENTIALS pricing tier configured: Log Streaming`. Floci models `UserPoolTier` but does
  not gate features on it anywhere, so accepting both event sources is consistent with the
  rest of the emulator; gating them here would be a new and inconsistent behaviour.
- **Destination validation and actual delivery.** Floci stores the ARN as given, does not
  require the log group / delivery stream / bucket to exist, and never writes to it.

### What I could not measure

The account I had available is on the `ESSENTIALS` tier, which rejects `userAuthEvents`
outright, so I could **not** observe:

- whether a `PLUS`-tier pool accepts `userAuthEvents` (I only saw `ESSENTIALS` reject it)
- the response shape for `FirehoseConfiguration` and `S3Configuration` destinations

I have implemented all three destination kinds as equally valid for the "has a destination"
check, which is what the API reference describes, and validated only what I could actually
call. Flagging it rather than presenting it as verified.

## Testing

- New `CognitoLogDeliveryConfigurationIntegrationTest` (10 cases), which creates its own
  pool and deletes it
- Cognito package: **417 tests, 0 failures**
- `make docs-check`: clean
- Python SDK compatibility coverage added for the round-trip, the replace-not-merge
  semantics, the empty-list clear, and the missing-destination rejection, the
  always-present-as-`[]` shape is the kind of thing a generated deserializer can drop

The full suite is left to CI, which shards it across four runners.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-python/tests/test_cognito.py
+++ b/compatibility-tests/sdk-test-python/tests/test_cognito.py
@@ -191,6 +191,102 @@ class TestCognitoAuth:
             cognito_client.delete_user_pool(UserPoolId=pool_id)
 
 
+class TestCognitoLogDeliveryConfiguration:
+    """Test Cognito log delivery configuration operations.
+
+    ``LogConfigurations`` is always present on the response, as ``[]`` when nothing
+    is configured, and ``SetLogDeliveryConfiguration`` replaces the list rather than
+    merging into it.
+    """
+
+    LOG_GROUP_ARN = "arn:aws:logs:us-east-1:000000000000:log-group:pytest-cognito-logs"
+
+    @pytest.fixture
+    def pool_id(self, cognito_client, unique_name):
+        response = cognito_client.create_user_pool(PoolName=f"pytest-log-pool-{unique_name}")
+        pool_id = response["UserPool"]["Id"]
+        yield pool_id
+        cognito_client.delete_user_pool(UserPoolId=pool_id)
+
+    def test_get_returns_an_empty_list_before_anything_is_configured(
+        self, cognito_client, pool_id
+    ):
+        """An unconfigured pool still returns the LogConfigurations member."""
+        config = cognito_client.get_log_delivery_configuration(UserPoolId=pool_id)[
+            "LogDeliveryConfiguration"
+        ]
+
+        assert config["UserPoolId"] == pool_id
+        assert config["LogConfigurations"] == []
+
+    def test_set_round_trips_and_replaces(self, cognito_client, pool_id):
+        """Set stores the configuration, and a second Set replaces rather than merges."""
+        cognito_client.set_log_delivery_configuration(
+            UserPoolId=pool_id,
+            LogConfigurations=[
+                {
+                    "LogLevel": "ERROR",
+                    "EventSource": "userNotification",
+                    "CloudWatchLogsConfiguration": {"LogGroupArn": self.LOG_GROUP_ARN},
+                }
+            ],
+        )
+
+        config = cognito_client.get_log_delivery_configuration(UserPoolId=pool_id)[
+            "LogDeliveryConfiguration"
+        ]
+        assert len(config["LogConfigurations"]) == 1
+        assert config["LogConfigurations"][0]["EventSource"] == "userNotification"
+        assert (
+            config["LogConfigurations"][0]["CloudWatchLogsConfiguration"]["LogGroupArn"]
+            == self.LOG_GROUP_ARN
+        )
+
+        cognito_client.set_log_delivery_configuration(
+            UserPoolId=pool_id,
+            LogConfigurations=[
+                {
+                    "LogLevel": "INFO",
+                    "EventSource": "userAuthEvents",
+                    "CloudWatchLogsConfiguration": {"LogGroupArn": self.LOG_GROUP_ARN},
+                }
+            ],
+        )
+        replaced = cognito_client.get_log_delivery_configuration(UserPoolId=pool_id)[
+            "LogDeliveryConfiguration"
+        ]
+        assert len(replaced["LogConfigurations"]) == 1
+        assert replaced["LogConfigurations"][0]["EventSource"] == "userAuthEvents"
+
+    def test_empty_list_clears_the_configuration(self, cognito_client, pool_id):
+        """An empty LogConfigurations clears what was stored."""
+        cognito_client.set_log_delivery_configuration(
+            UserPoolId=pool_id,
+            LogConfigurations=[
+                {
+                    "LogLevel": "ERROR",
+                    "EventSource": "userNotification",
+                    "CloudWatchLogsConfiguration": {"LogGroupArn": self.LOG_GROUP_ARN},
+                }
+            ],
+        )
+
+        cognito_client.set_log_delivery_configuration(UserPoolId=pool_id, LogConfigurations=[])
+
+        config = cognito_client.get_log_delivery_configuration(UserPoolId=pool_id)[
+            "LogDeliveryConfiguration"
+        ]
+        assert config["LogConfigurations"] == []
+
+    def test_set_rejects_a_configuration_with_no_destination(self, cognito_client, pool_id):
+        """Every event source in the request must name a destination."""
+        with pytest.raises(cognito_client.exceptions.InvalidParameterException):
+            cognito_client.set_log_delivery_configuration(
+                UserPoolId=pool_id,
+                LogConfigurations=[{"LogLevel": "ERROR", "EventSource": "userNotification"}],
+            )
+
+
 class TestCognitoDescribeUserPoolStandardAttributes:
     """DescribeUserPool must return all 20 standard OIDC attributes."""
 
--- a/compatibility-tests/sdk-test-python/tests/test_cognito.py
+++ b/compatibility-tests/sdk-test-python/tests/test_cognito.py
@@ -286,6 +286,51 @@ class TestCognitoLogDeliveryConfiguration:
                 LogConfigurations=[{"LogLevel": "ERROR", "EventSource": "userNotification"}],
             )
 
+    def test_set_rejects_more_than_two_configurations(self, cognito_client, pool_id):
+        """LogConfigurations is bounded at 2 entries."""
+        config = {
+            "LogLevel": "ERROR",
+            "EventSource": "userNotification",
+            "CloudWatchLogsConfiguration": {"LogGroupArn": self.LOG_GROUP_ARN},
+        }
+        with pytest.raises(cognito_client.exceptions.InvalidParameterException) as excinfo:
+            cognito_client.set_log_delivery_configuration(
+                UserPoolId=pool_id, LogConfigurations=[config, config, config]
+            )
+
+        assert "Member must have length less than or equal to 2" in str(excinfo.value)
+
+    def test_set_rejects_a_repeated_event_source(self, cognito_client, pool_id):
+        """An event source may appear at most once across the configurations."""
+        config = {
+            "LogLevel": "ERROR",
+            "EventSource": "userNotification",
+            "CloudWatchLogsConfiguration": {"LogGroupArn": self.LOG_GROUP_ARN},
+        }
+        with pytest.raises(cognito_client.exceptions.InvalidParameterException) as excinfo:
+            cognito_client.set_log_delivery_configuration(
+                UserPoolId=pool_id, LogConfigurations=[config, config]
+            )
+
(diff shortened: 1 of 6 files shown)


bsaurusrex:
Both findings were right, and both are fixed in 36076f3.

**P1, malformed requests clearing the configuration.**

Reproduced, then checked what AWS actually does with a raw signed request, since botocore refuses to send the malformed forms:

```console
--- LogConfigurations OMITTED entirely: 400 InvalidParameterException
    {"message":"1 validation error detected: Value null at 'logConfigurations' failed to satisfy constraint: Member must not be null"}

--- LogConfigurations = null: 400 InvalidParameterException
    {"message":"1 validation error detected: Value null at 'logConfigurations' failed to satisfy constraint: Member must not be null"}

--- LogConfigurations = a string, not an array: 400 SerializationException
    {"Message":"Expected list or null"}
```

So AWS rejects all three, and the two failure modes are distinct: an absent or null member fails validation, a wrongly typed one fails deserialization. `readObjectList` now returns null for absent/null and throws `SerializationException` for a non-array, and the service rejects null with the validation message above rather than treating it as an empty replacement.

Regression test `malformedRequestsAreRejectedRatherThanClearingTheConfiguration` covers all three forms and asserts that a rejected request leaves the stored configuration intact. With the fix reverted it fails:

```
CognitoLogDeliveryConfigurationIntegrationTest.malformedRequestsAreRejectedRatherThanClearingTheConfiguration:228 1 expectation failed.
```

**P2, em dashes.** Correct, and thank you: `AGENTS.md` line 314 says "No em-dashes anywhere, in any content." I have taken your suggested wording. I also checked my other open Cognito PRs for the same violation rather than just this one line, and swept the branch I have not raised yet.

Cognito package: 418 tests, 0 failures.

bsaurusrex:
Fixed in b3512be. You were right, and measuring it turned up a third case neither of us had.

Raw signed requests against the live API, since botocore will not send these:

```console
--- LogConfigurations = [1, 2]: 400 SerializationException
    {"Message":"Expected null"}

--- LogConfigurations = ["a string"]: 400 SerializationException
    {"Message":"Expected null"}

--- LogConfigurations = [null]: 200 OK
    {"LogDeliveryConfiguration":{"LogConfigurations":[],"UserPoolId":"..."}}
```

So a scalar element is a 400, exactly as you said, and a **null element is dropped rather than rejected**. My code would have thrown on that one too, from the validation loop rather than from `convertValue`.

`readObjectList` now walks the array itself: null elements are skipped, non-object elements raise `SerializationException`, and only objects reach `convertValue`. Test `malformedArrayElementsMatchAwsHandling` covers all three inputs.

Cognito package: 419 tests, 0 failures.

pgermosen (commented):
@bsaurusrex this is unusually well-measured work, the raw signed requests to reach cases botocore won't even send, and being upfront about what you couldn't verify (pricing tier, PLUS-tier userAuthEvents, Firehose/S3 shapes) rather than presenting untested paths as confirmed. Both greptile findings check out as genuinely fixed too: readObjectList now correctly distinguishes absent/null from wrong-type from a malformed element, and the null-element-gets-dropped behavior you found while measuring the second fix is exactly what the diff does.

One thing I found that greptile didn't catch: SetLogDeliveryConfiguration documents LogConfigurations as an array with a maximum of 2 items, but setLogDeliveryConfiguration only validates each element (LogLevel, EventSource, has-a-destination) and never checks configs.size() against that bound. A request with three or more well-formed configs, or two configs duplicating the same EventSource, gets accepted and stored as-is, where AWS would reject it as a length-constraint violation before it ever reached per-element validation. Given there are only two valid EventSource values this mostly bites on duplicate or malformed-count requests rather than a normal Terraform path, but it's a real gap against the documented constraint. I'm not asking you to reverse-engineer the exact AWS error message for it, that would need the same kind of live measurement you did for the other constraints, just to have the bound checked.

Everything else lines up: the ResourceNotFoundException/400 behavior on an unknown pool is correctly inherited from the existing describeUserPool, and you're right that the message-text difference is pre-existing and belongs in #2853 rather than here. Replace-not-merge and empty-list-clears semantics match the docs and are tested through a follow-up Get rather than just the mutation's own response.

bsaurusrex:
@pgermosen the bound was real and unchecked. Measuring it turned up three further divergences next to it, all fixed in `a27cc7b`.

You said you were not asking me to reverse-engineer the message, but the measurement was cheap once I had a pool up, so here it is rather than a guess. Live Cognito, `ap-southeast-1`:

```
configs                        AWS
1 valid                        InvalidParameterException (destination check)
2 valid distinct               FeatureUnavailableInTierException (ESSENTIALS)
0 (empty list)                 200 OK
3 well-formed                  1 validation error detected: ... at 'logConfigurations'
                               failed to satisfy constraint: Member must have length
                               less than or equal to 2
3, one bad LogLevel            2 validation errors detected: <length>; <enum at .2.>
3, one bad EventSource         2 validation errors detected: <length>; <enum at .2.>
2 duplicate eventSource        Request validation Failed. Following event sources
                               appear more then once in a request: [userNotification].
unknown pool + 3 configs       the length error
unknown pool + 1 bad LogLevel  the enum error
unknown pool + 1 valid config  ResourceNotFoundException
```

**The bound itself.** Checked now, against `configs.size()`, exactly as you described.

**Three things the bound dragged in.**

1. *An event source may appear only once.* You mentioned duplicates as part of the same case, and AWS does reject them, with its own message. That is now enforced. The missing-destination and duplicate complaints share one message, missing-destination clause first, and each event source is named once however many configurations carry it, both as measured. "more then once" is the service's spelling, kept deliberately and commented so it does not get tidied away later.

2. *The checks run before the pool lookup.* The last three rows above: a shape problem against a pool that does not exist reports the shape problem. Floci called `describeUserPool` first, so it reported `ResourceNotFoundException`. That was already wrong for the existing enum checks, not just the new bound, so the whole block moved above the lookup.

3. *Violations aggregate.* AWS reports every violation in one message, the list-length one ahead of the per-element ones. Floci threw on the first. That is only visible once the bound exists, since it is the length error that makes two errors possible, but it also affected two bad enums in a two-element request, which was already divergent.

The length message embeds the request model's `toString`, so the emulator reproduces it for all three destination shapes and for an absent destination.

**Tests.** Six new cases in `CognitoLogDeliveryConfigurationIntegrationTest` plus three in the Python compat suite. Reverting the service change fails all six:

```
[ERROR] Tests run: 18, Failures: 6
  setRejectsMoreThanTwoConfigurations
  theLengthViolationIsReportedAlongsideElementViolations
  shapeViolationsAreReportedBeforeTheMissingPool
  setRejectsARepeatedEventSource
  aRepeatedEventSourceWithNoDestinationReportsBothClauses
  aRejectedOversizedRequestLeavesTheConfigurationAlone  expected: <0> but was: <2>
```

That last one is the reason this mattered beyond the error text: the oversized request was being persisted.

**Not measured.** The ordering of the duplicate check against the tier gate. My two-distinct-source probe hit `FeatureUnavailableInTierException` first because `userAuthEvents` needs PLUS, so I could not isolate it, and Floci has no tier gate to order against anyway.

`services.cognito`: 425 tests, 0 failures, 0 errors. Full CI green on `a27cc7b`, all four shards plus the native build and compat jobs, the Python compat suite included. Docs gate and leak scan clean. Merges clean against `5465476` as of 2026-09-02 13:24 UTC.

pgermosen (approved):
This closes it, and it turned out to be more than a message-text gap. Confirmed the length check runs and aggregates ahead of the per-element enum violations the way you measured, and that the whole validation block now runs before the pool lookup, good catch that this was already wrong for the existing enum checks too, not just the new bound.

The real find here is that an oversized request wasn't just mis-reported before, it was actually getting persisted. Your revert-test transcript makes that obvious, and I confirmed by reading the code that validation now runs before the store write. The duplicate-EventSource rejection is a nice bonus too, and reproducing AWS's own "more then once" typo verbatim is exactly the kind of fidelity this project wants.

Appreciate the full measurement transcript again, and flagging what you couldn't isolate rather than guessing at it.

PR #2869, opened 2026-09-01
feat(cognito): implement managed login branding

### LINKED ISSUE #2856: [FEAT] cognito: managed login branding actions are unsupported
## Service

## API Action / Feature

Managed login branding: `CreateManagedLoginBranding`, `DescribeManagedLoginBranding`,
`DescribeManagedLoginBrandingByClient`, `UpdateManagedLoginBranding`,
`DeleteManagedLoginBranding`

## AWS Documentation

- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateManagedLoginBranding.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DescribeManagedLoginBrandingByClient.html

## Reproduction

```bash

POOL=$(aws --endpoint-url http://localhost:4599 cognito-idp create-user-pool \
  --pool-name gap-check --query 'UserPool.Id' --output text)
aws --endpoint-url http://localhost:4599 cognito-idp create-managed-login-branding \
  --user-pool-id "$POOL" --client-id x
```

```
An error occurred (UnsupportedOperation) when calling the CreateManagedLoginBranding operation: Operation CreateManagedLoginBranding is not supported.
```

## Why is this needed?

`aws_cognito_managed_login_branding` is a standard Terraform resource. A config containing
one cannot be applied against Floci:

```
Error: creating Cognito Managed Login Branding (2d854e25f6e7432586d5046c0b)
  with aws_cognito_managed_login_branding.main

operation error Cognito Identity Provider: CreateManagedLoginBranding, https response
error StatusCode: 400, api error UnsupportedOperation: Operation
CreateManagedLoginBranding is not supported.
```

Branding is presentation for the hosted UI, which Floci does not serve, so the useful scope
is storing and returning the settings (`Settings`, `Assets`, `UseCognitoProvidedValues`)
so that create/read/update/destroy round-trips, not rendering anything.

- [x] Yes
- [ ] No

bsaurusrex:
## Summary

Implements Cognito managed login branding: `CreateManagedLoginBranding`,
`DescribeManagedLoginBranding`, `DescribeManagedLoginBrandingByClient`,
`UpdateManagedLoginBranding` and `DeleteManagedLoginBranding`. All five returned
`UnsupportedOperation`, so a config declaring `aws_cognito_managed_login_branding` could
not be applied against Floci:

```
Error: creating Cognito Managed Login Branding (...)
api error UnsupportedOperation: Operation CreateManagedLoginBranding is not supported.
```

A branding belongs to exactly one app client, so it is stored on `UserPoolClient` rather
than in a new storage backend.

Closes #2856

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Measured against the live Cognito API on throwaway pools, created and deleted for the
purpose. Reference:
[CreateManagedLoginBranding](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateManagedLoginBranding.html),
[DescribeManagedLoginBrandingByClient](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DescribeManagedLoginBrandingByClient.html).

**`Settings` is omitted when the caller supplied none, while `Assets` is always returned.**
Creating with `UseCognitoProvidedValues=true` and nothing else:

```console
--- Create with UseCognitoProvidedValues=True: OK
    keys: ['Assets', 'CreationDate', 'LastModifiedDate', 'ManagedLoginBrandingId', 'UseCognitoProvidedValues', 'UserPoolId']
    Assets: 0 item(s)
```

`Settings` is absent from that list. Supplying one echoes it back:

```console
--- Create with Settings + 1 Asset: OK
    keys: ['Assets', 'CreationDate', 'LastModifiedDate', 'ManagedLoginBrandingId', 'Settings', 'UseCognitoProvidedValues', 'UserPoolId']
    Settings: {"components": {}}
    Assets: 1 item(s)
```

**One branding per client**, with a distinct exception type:

```console
--- Create AGAIN for the same client: ManagedLoginBrandingExistsException
    msg: A ManagedLoginBranding already exists for client <clientId>
```

**`ManagedLoginBrandingId` must be a version 4 UUID**, and the format is checked before the
lookup, so an all-zeros id never reaches "not found":

```console
--- Describe with 00000000-0000-0000-0000-000000000000: InvalidParameterException
    msg: 1 validation error detected: Value '...' at 'managedLoginBrandingId' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[4][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$
```

**Not-found wording differs by lookup path**, as it does for identity providers:

```console
--- by client, none exists: ResourceNotFoundException
    msg: ManagedLoginBranding for client <clientId> does not exist.
--- by a well-formed but unknown id: ResourceNotFoundException
    msg: ManagedLoginBranding does not exist.
--- create for an unknown client: ResourceNotFoundException
    msg: User pool client <clientId> does not exist.
```

Members an update omits are left unchanged, matching the identity provider semantics.

- **`Settings` schema validation.** AWS validates it deeply and rejects unknown properties:
  `Invalid settings provided. Validation errors: [{property: $.components.pageHeader.enabled,
  errorType: UnknownProperty}]`. That schema is not published, so Floci stores `Settings`
  opaquely.
- **`ReturnMergedResources`.** Against AWS it merges Cognito's own defaults into the
  response: on a pool with 8 configured assets it returned 38, with `Settings` fully
  populated from Cognito's defaults. Reproducing that needs Cognito's default corpus, so
  Floci returns the stored branding either way.

Both are documented in `docs/services/cognito.md` rather than left implicit.

## Testing

- New `CognitoManagedLoginBrandingIntegrationTest` (12 cases), creating and deleting its own
  pool and client
- Cognito package: **419 tests, 0 failures**
- `make docs-check`: clean

The full suite is left to CI, which shards it across four runners.

## Checklist

### DIFF
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -115,6 +115,34 @@ Standalone `TagResource` rejects reserved `floci:*` keys. `ListTagsForResource` 
 | AdminRemoveUserFromGroup | Removes a user from a group. |
 | AdminListGroupsForUser | Lists the groups assigned to a user. |
 
+### Managed Login Branding
+
+| Action | Description |
+|--------|-------------|
+| CreateManagedLoginBranding | Creates the branding for an app client. |
+| DescribeManagedLoginBranding | Returns a branding by its id. |
+| DescribeManagedLoginBrandingByClient | Returns the branding attached to an app client. |
+| UpdateManagedLoginBranding | Updates a branding's settings, assets or provided-values flag. |
+| DeleteManagedLoginBranding | Deletes a branding from its app client. |
+
+One branding per app client: a second `CreateManagedLoginBranding` for the same client is
+rejected with `ManagedLoginBrandingExistsException`. `ManagedLoginBrandingId` must be a
+version 4 UUID, and a malformed one is rejected before the lookup, as AWS does.
+`Settings` is omitted from the response when the caller supplied none, while `Assets` is
+always returned. Members an update omits are left unchanged.
+
+Branding is presentation for the hosted UI, which Floci does not serve, so it is stored
+and returned rather than rendered. Two divergences follow from that:
+
+- **`Settings` is stored opaquely.** AWS validates it against a deep schema, rejecting
+  unknown properties with `Invalid settings provided. Validation errors: [{property:
+  $.components...., errorType: UnknownProperty}]`. That schema is not published, so Floci
+  accepts any object.
+- **`ReturnMergedResources` is not honoured.** Against AWS it merges Cognito's own default
+  settings and assets into the response: on a pool with 8 configured assets it returns 38.
+  Reproducing that needs Cognito's default corpus, so Floci returns the stored branding
+  either way.
+
 ## Well-Known And OAuth Endpoints
 
 | Endpoint                                             | Description                                                      |
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -130,8 +130,14 @@ request with neither is rejected. One branding per app client: a second
 `CreateManagedLoginBranding` for the same client is rejected with
 `ManagedLoginBrandingExistsException`. `ManagedLoginBrandingId` must be a
 version 4 UUID, and a malformed one is rejected before the lookup, as AWS does.
+`Assets` holds at most 40 entries on create and on update.
 `Settings` is omitted from the response when the caller supplied none, while `Assets` is
 always returned. Members an update omits are left unchanged.
+
+The asset-count and branding-id checks run before the pool, client or branding is looked
+up, so an oversized request naming something that does not exist reports the request
+problem rather than `ResourceNotFoundException`, and an update violating both reports them
+in one message with the asset list first.
 
 Branding is presentation for the hosted UI, which Floci does not serve, so it is stored
 and returned rather than rendered. Two divergences follow from that:
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -125,8 +125,10 @@ Standalone `TagResource` rejects reserved `floci:*` keys. `ListTagsForResource` 
 | UpdateManagedLoginBranding | Updates a branding's settings, assets or provided-values flag. |
 | DeleteManagedLoginBranding | Deletes a branding from its app client. |
 
-One branding per app client: a second `CreateManagedLoginBranding` for the same client is
-rejected with `ManagedLoginBrandingExistsException`. `ManagedLoginBrandingId` must be a
+`CreateManagedLoginBranding` must name either `UseCognitoProvidedValues` or `Settings`; a
+request with neither is rejected. One branding per app client: a second
+`CreateManagedLoginBranding` for the same client is rejected with
+`ManagedLoginBrandingExistsException`. `ManagedLoginBrandingId` must be a
 version 4 UUID, and a malformed one is rejected before the lookup, as AWS does.
 `Settings` is omitted from the response when the caller supplied none, while `Assets` is
 always returned. Members an update omits are left unchanged.
@@ -138,6 +140,10 @@ and returned rather than rendered. Two divergences follow from that:
   unknown properties with `Invalid settings provided. Validation errors: [{property:
   $.components...., errorType: UnknownProperty}]`. That schema is not published, so Floci
   accepts any object.
+- **A wrongly typed `Settings` returns a client error.** AWS answers that particular input
+  with `InternalErrorException` and a 500; Floci returns
+  `SerializationException: Unexpected field type`, which is what AWS returns for a wrongly
+  typed `Assets`. Reproducing someone else's 500 seemed worse than being consistent.
 - **`ReturnMergedResources` is not honoured.** Against AWS it merges Cognito's own default
   settings and assets into the response: on a pool with 8 configured assets it returns 38.
   Reproducing that needs Cognito's default corpus, so Floci returns the stored branding
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
@@ -510,21 +510,33 @@ public class CognitoJsonHandler {
     }
 
     /**
-     * AWS coerces a JSON string in this member instead of rejecting it: {@code "true"} and
-     * {@code "yes"} both store true, while {@code "false"} stores false. Jackson maps every
-     * string except {@code "true"} to false, which would turn {@code "yes"} into a request
-     * selecting no branding source and get it rejected. Measured against Cognito in
-     * ap-southeast-1.
+     * AWS coerces a JSON string in this member instead of rejecting it, and rejects every
+     * other non-boolean type. Measured against Cognito in ap-southeast-1:
+     *
+     * <pre>
+     * absent          accepted, stored false
+     * true / false    accepted as given
+     * "true", "yes"   accepted, stored true
+     * "false"         accepted, stored false
+     * 1, null, {}, [] SerializationException
+     * </pre>
+     *
+     * <p>Jackson would map all of those last four to false through {@code asBoolean()},
+     * which is how a malformed flag used to be accepted alongside valid settings and
+     * persisted as unintended branding state.
      */
     private static Boolean readUseCognitoProvidedValues(JsonNode request) {
         if (!request.has("UseCognitoProvidedValues")) {
             return null;
         }
         JsonNode node = request.path("UseCognitoProvidedValues");
+        if (node.isBoolean()) {
+            return node.booleanValue();
+        }
         if (node.isTextual()) {
             return !"false".equalsIgnoreCase(node.asText());
         }
-        return node.asBoolean();
+        throw new AwsException("SerializationException", null, 400);
     }
 
     private Response handleCreateManagedLoginBranding(JsonNode request) {
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
@@ -20,6 +20,7 @@ import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.core.Response;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -583,6 +584,11 @@ public class CognitoJsonHandler {
         return objectMapper.convertValue(node, new TypeReference<LinkedHashMap<String, Object>>() {});
     }
 
+    /**
+     * Elements are checked before conversion so a scalar cannot escape as an IllegalArgumentException.
+     * AWS separates the two failures: a null element is a validation error, a non-object element a
+     * deserialization one. Note this differs from LogConfigurations, where a null element is dropped.
+     */
     private List<Map<String, Object>> readMapList(JsonNode request, String member) {
         JsonNode node = request.get(member);
         if (node == null || node.isNull()) {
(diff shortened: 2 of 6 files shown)


bsaurusrex:
Fixed the real part in cd80208, and pushing back on one part with a measurement.

I sent each malformed form as a raw signed request to the live API:

```console
--- UseCognitoProvidedValues OMITTED: 400 InvalidParameterException
    {"message":"useCognitoProvidedValues or settings should be specified (but not both)"}

--- UseCognitoProvidedValues = "yes" (a string): 200 OK
    {"ManagedLoginBranding":{"Assets":[],"ManagedLoginBrandingId":"26148bf3-...

--- Assets = "nope": 400 SerializationException
    {"Message":"Unexpected field type"}

--- Settings = "nope": 500 InternalErrorException
    {"message":"Internal server error."}
```

**Fixed: the omitted case.** You were right that it should not be coerced, and the reason is stronger than "absent means false": AWS enforces a rule I did not know about, that a create must name either `UseCognitoProvidedValues` or `Settings`. Create now rejects a request naming neither, with that message, and `Assets` of the wrong type raises `SerializationException`.

**Rebutting: the non-boolean case.** AWS returns 200 for `UseCognitoProvidedValues: "yes"` and creates the branding. Rejecting it would make Floci stricter than the service it emulates, so I have left it accepted. Leaving this thread open rather than resolving it, since it is a disagreement rather than a fix.

**A note on `Settings` of the wrong type.** AWS answers that one with a 500 `InternalErrorException`, which looks like an AWS bug rather than a contract. I return `SerializationException: Unexpected field type`, matching what AWS returns for a wrongly typed `Assets`, on the grounds that faithfully reproducing someone else's 500 is worse than being consistent. It is called out in `docs/services/cognito.md` rather than left silent, and I am happy to change it if you would rather match the 500 exactly.

New test `malformedCreateMembersMatchAwsHandling` pins all three behaviours, including the accepted non-boolean. Cognito package: 419 tests, 0 failures.

bsaurusrex:
Fixed in a495c9c4. You were right, and the miss is worth naming: this is the same class of bug as the `LogConfigurations` one on #2867, which I fixed in that reader without carrying the fix across to this one. Checking the class rather than the instance is exactly what I should have done the first time.

Measured against the live API before changing anything:

```console
--- Assets = [1, 2]: 400 SerializationException
    {"Message":"Unexpected value type in payload"}

--- Assets = ["str"]: 400 SerializationException
    {"Message":"Unexpected value type in payload"}

--- Assets = [null]: 400 InvalidParameterException
    {"message":"1 validation error detected: Value '[null]' at 'assets' failed to satisfy constraint: Member must satisfy constraint: [Member must not be null]"}
```

Two things that measurement settled which I would otherwise have got wrong:

- The scalar message is **`Unexpected value type in payload`**, not the `Expected null` that `LogConfigurations` returns. Different member, different text.
- A null element is **rejected** here, where `LogConfigurations` silently drops it. I had been about to reuse the drop-nulls rule from the other reader.

`readMapList` now walks the array and checks each element before conversion, so nothing reaches `convertValue` that can throw. The three inputs are pinned in `malformedCreateMembersMatchAwsHandling`.

Cognito package: 420 tests, 0 failures. `make docs-check` clean.

pgermosen (commented):
@bsaurusrex same rigor as your other Cognito work this round, appreciated. Both greptile findings hold up: the omitted-members fix is confirmed in the diff, exact message and all, and the pushback on the non-boolean UseCognitoProvidedValues value is the right call, you measured a 200 against live AWS for "yes" and rejecting it would make Floci stricter than the service it emulates, good to leave that thread open as a disagreement rather than resolve it. The asset-elements fix checks out too, and I liked the note about almost reusing the drop-null-elements rule from the LogConfigurations fix on #2867 and measuring first instead, this operation actually rejects a null element where that one drops it, different member, different behavior.

Two things I found that weren't caught yet, both against the live doc for CreateManagedLoginBranding. First: the doc says UseCognitoProvidedValues=true requires omitting Settings and Assets, and its own worked example shows false combined with populated Settings and Assets working fine, so it's specifically true that's exclusive. The current check only requires at least one of useCognitoProvidedValues/settings to be present, there's no check for true plus a non-null Settings or non-empty Assets, so that combination is accepted today where AWS would reject it. The message you measured literally says "(but not both)," so it looks like the "neither" half got fixed after greptile's finding but the "both" half never got tested. Might be worth checking whether updateManagedLoginBranding needs the same treatment, it doesn't have any version of this check right now.

Second: Assets is documented with a maximum of 40 items, and nothing checks assets.size() against that anywhere in create or update. Same class of gap as the LogConfigurations max-2 case on #2867.

Everything else lines up with your measurements: the one-branding-per-client exception, the UUID format check running before the lookup, the three distinct not-found messages depending on how you got there, and Settings being omitted from the response only when the caller supplied none while Assets is always present.

bsaurusrex:
I took this to the live service before changing anything, and the finding as written does not hold. Chasing it did surface two real divergences next to it, which are fixed in 6a59bcd.

**Where the finding does not hold.** `CreateManagedLoginBranding` already rejected the omitted case with AWS's own wording, and the handler passed `null` rather than `false` for an absent member, so nothing was silently coerced there. On non-boolean values, AWS is more permissive than the finding assumes, measured in `ap-southeast-1`:

```
UseCognitoProvidedValues: "true"  -> 200 OK, stored true
UseCognitoProvidedValues: "yes"   -> 200 OK, stored true
UseCognitoProvidedValues: 1       -> 400 SerializationException
Settings: "not-an-object"         -> 500 InternalErrorException
Assets:   {"k": "v"}              -> 400 SerializationException
```

So a JSON string is coerced by the service, not rejected, and the wrong-type cases are handled in the serialization layer rather than by the action. `Settings` as a string produces an AWS 500, which is not behaviour worth reproducing.

**What was actually wrong.** A request must select exactly one source of branding, and the same rule governs create and update:

```
useCognitoProvidedValues  settings   AWS
absent                    absent     InvalidParameterException
true                      absent     accepted
false                     absent     InvalidParameterException
absent                    present    accepted
true                      present    InvalidParameterException
false                     present    accepted
```

Create enforced only the first row. Update enforced nothing, so on this branch:

```
create both members     -> 200 (AWS 400)
update both members     -> 200 (AWS 400)
update neither member   -> 200 (AWS 400)
update Assets only      -> 200 (AWS 400)
```

Presence of the member is not the test. `false` selects no source, which is why it is rejected unless settings supply one. Both paths now share one `validateBrandingSource`.

Reading the flag had to move with it. Jackson maps every string except `"true"` to false, so `"yes"` became a request selecting no source and would have started failing once the rule was enforced. The handler now coerces the way the service does. That also means the existing `updateLeavesOmittedMembersAlone` case was sending `UseCognitoProvidedValues: false` alone, which AWS rejects, so it now updates through `Settings` instead and still proves omitted members survive.

Full `services.cognito` package: 421 tests, 0 failures, 0 errors.

bsaurusrex:
Right, and it is fixed in the commit above. This is the half of the original finding that survived measurement, now that the source rule made it reachable.

Measured against Cognito in `ap-southeast-1`, each on a fresh client, with valid `Settings` alongside:

```
UseCognitoProvidedValues: 1        + Settings -> 400 SerializationException
UseCognitoProvidedValues: null     + Settings -> 400 SerializationException
UseCognitoProvidedValues: {}       + Settings -> 400 SerializationException
UseCognitoProvidedValues: []       + Settings -> 400 SerializationException
(member omitted)                   + Settings -> 200 OK, stored false
```

So the type is rejected outright rather than coerced, and an absent member is not the same thing as a malformed one. `asBoolean()` flattened all four of those to `false`, which then satisfied the source rule and stored branding nobody asked for.

The reader now takes booleans as given, keeps the string coercion the service actually does (`"true"` and `"yes"` store true, `"false"` stores false), and raises `SerializationException` for everything else, using the same `new AwsException("SerializationException", null, 400)` shape `SesController` already uses. Absent still returns null and is handled by the source rule rather than here.

Pinned by a test over all four malformed forms plus the omitted-member control.

`services.cognito`: 422 tests, 0 failures, 0 errors.

bsaurusrex:
@pgermosen the 40-item bound on `Assets` was missing from both create and update, as you said. Fixed in `ddb877d`, and measuring it settled the ordering question too.

Live Cognito, `ap-southeast-1`:

```
create 40 assets              passes the bound, then hits the source rule
create 41 assets              1 validation error detected: ... at 'assets' failed to
                              satisfy constraint: Member must have length less than
                              or equal to 40
create 41 + unknown client    the length error, not the missing client
update 41 + malformed id      2 validation errors detected: <assets>; <id pattern>
update malformed id alone     1 validation error detected: <id pattern>
update valid id, unknown pool ResourceNotFoundException
describe malformed id,
  unknown pool                1 validation error detected: <id pattern>
```

**The bound.** Checked in create and in update, both through one helper, so update is covered rather than left as the asymmetry you spotted.

**Ordering.** Rows three, four and seven put the shape checks ahead of every lookup. `createManagedLoginBranding` was calling `describeUserPoolClient` first, and `describeManagedLoginBranding` was calling `describeUserPool` before the branding-id pattern check, so both reported a missing resource where AWS reports the malformed request. The checks moved above the lookups. `DescribeManagedLoginBranding` was a pre-existing divergence rather than something the bound introduced, but it is the same check in the same class and it would have been odd to fix one and leave the other.

**Aggregation.** Row four: an update violating both reports them in one message, asset list first. Create and update and describe now share the same collect-then-throw shape.

**The rendered value.** The length message embeds the request model's `toString`, which the emulator reproduces. The blob renders as the decoded buffer, and I confirmed the length tracks the payload rather than the base64 string at three sizes:

```
3 bytes   -> bytes=java.nio.HeapByteBuffer[pos=0 lim=3 cap=3]
17 bytes  -> bytes=java.nio.HeapByteBuffer[pos=0 lim=17 cap=17]
70 bytes  -> bytes=java.nio.HeapByteBuffer[pos=0 lim=70 cap=70]
```

**Not measured, and chosen deliberately.** How AWS renders a `Bytes` value that is not valid base64. botocore encodes the blob itself so it cannot send one, and I did not sign a raw request for it. Rather than invent a rendering, the emulator decodes leniently, which cannot throw: building a rejection must not raise a second failure of its own. If you would rather it rejected a malformed blob outright I can measure that properly with a signed request, but that is a serialization-layer question and felt like a different PR.

**Tests.** Six new cases in `CognitoManagedLoginBrandingIntegrationTest`. Reverting the service change fails five:

```
[ERROR] Tests run: 21, Failures: 5
  createRejectsMoreThanFortyAssets
  updateRejectsMoreThanFortyAssets
  theAssetViolationIsReportedAheadOfTheBrandingIdViolation
  shapeViolationsAreReportedBeforeTheMissingPool
  aRejectedUpdateLeavesTheAssetsAlone
```

`fortyAssetsAreAccepted` passes either way by design: it is the off-by-one guard, and it would only fail if the bound were set at 39.

I did not add a compat-suite case here, since this PR has no branding class there to extend and adding one looked like scope I had not been asked for. Say the word if you would rather it had one.

`services.cognito`: 428 tests, 0 failures, 0 errors. Full CI green on `ddb877d`, all four shards plus the native build and compat jobs. Docs gate and leak scan clean. Merges clean against `54654760` as of 2026-09-02 12:54 UTC.

pgermosen (approved):
Both closed, and the exclusivity one is better than what I originally asked for. Good call re-measuring instead of trusting my true-excludes-Settings framing, the real rule is the full truth table where false means explicitly no source, distinct from absent, and I checked validateBrandingSource against all six rows by hand, it's right. Nice that it's shared between create and update now too.

The coercion bug you found chasing that is the real prize here, malformed types silently becoming false and then sailing through the source rule to persist branding nobody asked for is a genuine data-integrity gap, not just a wrong error code. Good catch.

Assets bound looks right in both paths now, and moving the shape checks ahead of the lookups in create and describe closes a inconsistency that predated this PR rather than leaving it sitting next to a fix for the same class of bug.

PR #2876, opened 2026-09-01
fix(apigateway): import x-amazon-apigateway-any-method on OpenAPI import

### LINKED ISSUE #2875: [BUG] bug(apigateway): OpenAPI import ignores x-amazon-apigateway-any-method, method never created
## Description

When importing an OpenAPI/Swagger spec via `PutRestApi` (e.g. `aws apigateway put-rest-api --mode merge`, or Terraform's `aws_api_gateway_rest_api` with a `body`), paths that only define the AWS vendor extension `x-amazon-apigateway-any-method` (instead of a standard verb like `get`/`post`) never get a method created at all — not even importable, not even reachable.

This is different from #710 / #1547 (which were about `ANY` routing/wildcard-matching at request time). Here the method is missing entirely at import time.

## Steps to reproduce

1. Start floci 2.0.0.
2. Create a REST API with a body containing:

```json
{ "swagger": "2.0", "info": { "title": "repro", "version": "1.0" }, "paths": {
    "/echo": {
      "x-amazon-apigateway-any-method": {
        "responses": {},
        "x-amazon-apigateway-integration": {
          "type": "mock",
          "requestTemplates": { "application/json": "{\"statusCode\": 200}" },
          "responses": { "default": { "statusCode": "200" } }
        }
      }
    } }
}
```
```bash
aws apigateway put-rest-api --rest-api-id <id> --mode merge --body file://spec.json
aws apigateway get-resources --rest-api-id <id>
# grab resource id for /echo, then:
aws apigateway get-method --rest-api-id <id> --resource-id <resource-id> --http-method ANY
```

## Expected

 get-method  returns the imported  ANY  method with its MOCK integration (this is exactly how real AWS API Gateway behaves).

## Actual

An error occurred (NotFoundException) when calling the GetMethod operation: Invalid method specified

Invoking the deployed stage on that path returns  405 Method Not Allowed  for every HTTP verb, since no method was ever created.

## Root cause

 ApiGatewayService.applyOpenApiSpec()  only iterates  pathItem.readOperationsMap() , the swagger-parser's typed map of standard OpenAPI verbs.  x-amazon-apigateway-any-method  is a vendor extension key, not a real OpenAPI operation, so it's parsed into  pathItem.getExtensions()  as an untyped  Map  and is never read.

Jojo134:
## Summary

Closes #2875

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayService.java
@@ -2245,10 +2245,17 @@ public class ApiGatewayService {
             if (pathItem.getExtensions() != null) {
                 Object anyMethodExt = pathItem.getExtensions().get("x-amazon-apigateway-any-method");
                 if (anyMethodExt != null) {
-                    Operation anyOperation = io.swagger.v3.core.util.Json.mapper()
-                            .convertValue(anyMethodExt, Operation.class);
-                    applyOperation(region, apiId, resourceId, "ANY", anyOperation, openAPI,
-                            schemeToAuthorizerId, schemeToAuthType, validatorNameToId);
+                    try {
+                        Operation anyOperation = io.swagger.v3.core.util.Json.mapper()
+                                .convertValue(anyMethodExt, Operation.class);
+                        applyOperation(region, apiId, resourceId, "ANY", anyOperation, openAPI,
+                                schemeToAuthorizerId, schemeToAuthType, validatorNameToId);
+                    } catch (IllegalArgumentException e) {
+                        throw new AwsException("BadRequestException",
+                                "Invalid x-amazon-apigateway-any-method definition for path " + path + ": "
+                                        + e.getMessage(),
+                                400);
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


pgermosen (commented):
@Jojo134 good fix, and the extraction into `applyOperation` is a clean, faithful move — I diffed it against the original inline code and the logic itself is unchanged, just relocated so the ANY-extension path can reuse it. Nice end-to-end test too, going all the way through deploy and invoking all four verbs rather than stopping at "the method exists."

I want to second two of Copilot's three comments, having checked them against the rest of the class rather than taking them at face value. There's no try/catch around `applyOpenApiSpec` in either `importRestApi` or `putRestApi` — only `parseOpenApiSpec` itself converts a bad spec into AWS's clean 400. So a malformed `x-amazon-apigateway-any-method` block (not shaped like an operation) would throw an uncaught `IllegalArgumentException` from `convertValue` and come back as a raw 500 where AWS would give a validation error. The unchecked-cast comment is the same problem one line down: that cast used to sit inside `applyOpenApiSpec`, which carries a method-level `@SuppressWarnings("unchecked")`, and it's now in the new `applyOperation`, which doesn't have one — won't break the build, the project doesn't fail on warnings, but a malformed `x-amazon-apigateway-integration` block now throws `ClassCastException` uncaught instead. Both come down to the same gap: nothing validates the shape of these two extensions before using them, the way the outer parse step already does for the spec as a whole.

The third comment, about the test using OpenAPI 3.0.1 instead of the Swagger 2.0 spec from the original repro, I don't think is a real gap — `parseOpenApiSpec` goes through swagger-parser, which normalizes either dialect into the same object model before any of this code runs, so the fix behaves identically regardless of which one the test uses. Worth keeping in mind for fidelity to the exact repro, but not something I'd ask you to change.

Jojo134:
@pgermosen thank you for your feedback. I applied changes and was also able to keep the `@SuppressWarnings` local to only the cast as copilot also suggested and not have it on the whole method.

pgermosen (commented):
Both of those are fixed now, thanks. The malformed any-method block converts cleanly to a BadRequestException/400 instead of the uncaught IllegalArgumentException, and the integration cast is guarded the same way with the suppress annotation scoped down to just that cast like you said. Good to see both backed by tests that actually hit the malformed-input case and check for the 400, not just the code change on its own.

One small thing: there's an em-dash in the new comment above the x-amazon-apigateway-any-method handling block in ApiGatewayService.java, could you swap that for a colon or period per AGENTS.md?

pgermosen (approved):
That's it, confirmed clean now. Nothing else outstanding.

PR #2877, opened 2026-09-01
fix(apigatewayv2): provision integration, route and stage on quick create

### LINKED ISSUE #1902: [BUG] Unable to access floci apigatewayv2 url

Karan-Tekuncorked:
## Summary

`CreateApi`'s `target` parameter ("quick create") was accepted but silently ignored — no `AWS_PROXY` integration, no `$default` route, and no `$default` stage were ever provisioned.

Without a stage, `ApiGatewayExecuteApiHostFilter.stageExists()` throws `Stage not found`, the filter returns without rewriting the request, and the invoke request falls through unmodified to whichever other virtual-hosted-style filter claims the `execute-api` host next (S3's virtual-hosted-bucket filter, in the common case) — instead of reaching the API. Repro from a fresh `floci/floci:latest` container:

```
aws apigatewayv2 create-api --name learning-api --protocol-type HTTP \
  --target arn:aws:lambda:us-east-1:000000000000:function:learning-fn
curl http://<apiId>.execute-api.localhost.floci.io:4566/
# -> S3 ListAllMyBucketsResult / NoSuchBucket, not the Lambda response
```

This reproduces even with matching region throughout (`us-east-1` create + invoke), so it's a separate cause from the cross-region dispatch bug #1906 targets.

## Fix

Provision the `AWS_PROXY` integration, `$default` catch-all route, and auto-deploy `$default` stage that real API Gateway creates for quick create, using the existing `createIntegration`/`createRoute`/`createStage` helpers — right after the `Api` itself is persisted in `createApi`.

Verified against a locally built image: the same repro now reaches the Lambda integration (`{"message":"Function not found: ..."}` for a missing function, and a real invoke for an existing one) instead of S3's XML error.

Fixes #1902

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

Verified against `aws-cli/2.36.35` (`apigatewayv2 create-api --target`, `get-stages`, `get-routes`, `get-integrations`) and a real Lambda invocation through the resulting HTTP API endpoint.

## Checklist

- [x] `./mvnw test` passes locally (full `apigatewayv2`/`apigateway` suites: 1035 tests, 0 failures, 0 errors)
- [x] New integration test added (`ApiGatewayV2QuickCreateTest`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigatewayv2/ApiGatewayV2Service.java
@@ -124,6 +124,30 @@ public class ApiGatewayV2Service {
 
         apiStore.put(apiKey(region, api.getApiId()), api);
         LOG.infov("Created {0} API: {1} ({2}) in {3}", protocolType, api.getName(), api.getApiId(), region);
+
+        // Quick create: when the caller supplies Target (a Lambda ARN or HTTP URL), AWS
+        // auto-provisions an AWS_PROXY integration, a "$default" catch-all route pointing at
+        // it, and an auto-deploy "$default" stage — without this the API has no route and no
+        // stage, so ApiGatewayExecuteApiHostFilter's stage lookup fails and every invocation
+        // falls through to whichever other virtual-hosted-style filter claims the request next.
+        Object targetValue = request.get("target");
+        String target = targetValue != null ? String.valueOf(targetValue) : null;
+        if ("HTTP".equals(protocolType) && target != null && !target.isBlank()) {
+            Integration integration = createIntegration(region, apiId, Map.of(
+                    "integrationType", "AWS_PROXY",
+                    "integrationUri", target,
+                    "integrationMethod", "POST",
+                    "payloadFormatVersion", "2.0"));
+            createRoute(region, apiId, Map.of(
+                    "routeKey", "$default",
+                    "target", "integrations/" + integration.getIntegrationId()));
+            createStage(region, apiId, Map.of(
+                    "stageName", "$default",
+                    "autoDeploy", "true"));
+            LOG.infov("Quick create: provisioned AWS_PROXY integration, $default route and stage for API {0} -> {1}",
+                    apiId, target);
+        }
+
         return api;
     }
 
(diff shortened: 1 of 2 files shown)


Karan-Tekuncorked:
Fixed both points from the automated review:

1. **HTTP URL targets** were hardcoded to `AWS_PROXY`, so a URL target got parsed as a Lambda ARN and failed invocation. Now detects the target kind (`arn:` prefix vs HTTP URL) and provisions `AWS_PROXY`/`POST` for Lambda, `HTTP_PROXY`/`ANY` for a URL, matching real API Gateway quick-create. Split the test into one case per target kind (`quickCreateWithLambdaTargetProvisionsAwsProxyIntegration`, `quickCreateWithHttpUrlTargetProvisionsHttpProxyIntegration`).
2. Removed the em-dashes from the new comment and Javadoc per AGENTS.md.

Full `apigatewayv2`/`apigateway` suite still green: 1036 tests, 0 failures.

pgermosen (approved):
@Karan-Tekuncorked welcome, and nice first PR — the repro is concrete and the fix is well targeted. I checked the HTTP_PROXY/AWS_PROXY fix against AWS's own docs rather than just the review exchange: the quick-create `target` description says, almost verbatim, "for HTTP integrations, specify a fully qualified URL, for Lambda integrations, specify a function ARN, the type of the integration will be HTTP_PROXY or AWS_PROXY respectively" — that's exactly what you implemented, including which one maps to which.

I also traced the plumbing myself: every key you're passing into `createIntegration`/`createRoute`/`createStage` matches what those methods already expect on main, and the `protocolType == HTTP` gate matches AWS's own "supported only for HTTP APIs" restriction on `target`. Good test coverage too — one case per target kind, plus an explicit no-target case proving the new code doesn't fire where it shouldn't.

Approving.

PR #2880, opened 2026-09-01
fix(rds-data): fixing empty array response from statement

QAnders:
## Summary

Branched from Tag 2.0.0

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Data-API statement in AWS SDK doesn't return the `records` array at all, we should adhere that and not return empty array `[]`

## Checklist


pgermosen (approved):
@QAnders welcome, and thanks for this — clean, well-tested fix. The tests correctly lock in both directions: `records` omitted for insert/update/delete/DDL, still present-but-empty for a `SELECT` matching zero rows. That distinction matters and you got it right — `records` should only be absent when there's no result set at all, not when the result set happens to be empty. It also lines up with something already documented here: `generatedFields` is already omitted for statements that return a result set, so this makes `records` behave the same way in the other direction rather than introducing a new asymmetry.

One thing not worth acting on: both bots flagged the new prose for "violating repository style" with em dashes, citing AGENTS.md. I checked — AGENTS.md doesn't mention em dashes, and the same doc file already has four of them elsewhere, unrelated to this change. False positive from both, nothing to fix there.

Approving.

PR #2885, opened 2026-09-01
fix(cloudformation): expand AWS::Serverless::StateMachine, and reject every unresolvable DefinitionUri on both StateMachine and HttpApi

### LINKED ISSUE #2883: [BUG] CloudFormation (CFN): an `AWS::Serverless::StateMachine` is stubbed instead of expanded, so the stack reports CREATE_COMPLETE with a fake ARN and no state machine exists
## Service

CloudFormation (SAM transform, `AWS::Serverless-2016-10-31`), Step Functions

## AWS API Action

`CreateStack`, `DescribeStackResources`, `GetTemplate`, `stepfunctions:ListStateMachines`

## Expected behavior

The SAM transform expands `AWS::Serverless::StateMachine` into a native
`AWS::StepFunctions::StateMachine`. Captured from a live stack, both template stages of the same
resource:

```sh
# Static keys in the environment outrank AWS_PROFILE; without the unset the call fails with
# InvalidClientTokenId instead of reaching AWS.
  aws cloudformation get-template --stack-name my-stack \
      --template-stage Original --region us-east-1
  aws cloudformation get-template --stack-name my-stack \
      --template-stage Processed --region us-east-1
```

`Original`, resource `OrdersStreamStateMachine`:

```json
{
  "Type": "AWS::Serverless::StateMachine",
  "Properties": {
    "Name": { "Fn::Sub": "${Environment}-${AWS::Region}-OrdersStream" },
    "Tags": {
      "team": "platform",
      "project-id": 12345678,
      "environment": { "Ref": "Environment" }
    },
    "Logging": { "Level": "ALL", "IncludeExecutionData": true, "Destinations": [ ... ] },
    "Role": { "Fn::GetAtt": [ "OrdersStreamRole", "Arn" ] },
    "Type": "STANDARD",
    "DefinitionUri": "s3://cfn-snippets/definition.asl.json",
    "DefinitionSubstitutions": { ... },
    "Tracing": { "Enabled": true }
  }
}
```

`Processed`, same logical id:

```json
{
  "Type": "AWS::StepFunctions::StateMachine",
  "Metadata": { "SamResourceId": "OrdersStreamStateMachine" },
  "Properties": {
    "DefinitionS3Location": { "Bucket": "cfn-snippets", "Key": "definition.asl.json" },
    "LoggingConfiguration": { "Level": "ALL", "IncludeExecutionData": true, "Destinations": [ ... ] },
    "RoleArn": { "Fn::GetAtt": [ "OrdersStreamRole", "Arn" ] },
    "StateMachineName": { "Fn::Sub": "${Environment}-${AWS::Region}-OrdersStream" },
    "StateMachineType": "STANDARD",
    "Tags": [
      { "Key": "stateMachine:createdBy", "Value": "SAM" },
      { "Key": "team", "Value": "platform" },
      { "Key": "project-id", "Value": 12345678 },
      { "Key": "environment", "Value": { "Ref": "Environment" } }
    ],
    "DefinitionSubstitutions": { ... },
    "TracingConfiguration": { "Enabled": true }
  }
}
```

Counted over the whole stack, one call each:

| Measured on a deployed stack | Value |
|---|---|
| `AWS::Serverless::StateMachine` in `Original` | 67 |
| `AWS::StepFunctions::StateMachine` in `Processed` | 67 |
| `AWS::Serverless::*` surviving in `Processed` | 0 |
| resource-level keys, both stages: `Type`, `Properties`, `Metadata` | 67 of 67 |
| `Metadata` carried through unchanged from `Original` (the SAM CLI writes `SamResourceId`; CloudFormation does not synthesize it) | 67 of 67 |
| `Condition` or `DependsOn` on any of the 86 `AWS::Serverless::*` resources | 0 |
| `stateMachine:createdBy = SAM` present, and first in the `Tags` list | 67 of 67 |
| Tag values keeping their JSON type (number `12345678`, `{"Ref": "Environment"}`) | 67 of 67 |
| Mapped values left as unevaluated intrinsics (`Fn::Sub`, `Fn::GetAtt`) | 67 of 67 |

Eight SAM properties are declared on all 67, and all eight are carried across:

| SAM property | CloudFormation property |
|---|---|
| `Name` | `StateMachineName` |
| `Type` | `StateMachineType` |
| `Role` | `RoleArn` |
| `DefinitionUri` | `DefinitionS3Location` `{Bucket, Key}` |
| `DefinitionSubstitutions` | `DefinitionSubstitutions` |
| `Logging` | `LoggingConfiguration` |
| `Tracing` | `TracingConfiguration` |
| `Tags` (a map) | `Tags` (a list of `{Key, Value}`) |

A `DefinitionUri` that is still a local path is refused, and the refusal is not an HTTP error. AWS
creates the change set and marks it failed, carrying the transform's own sentence:

```sh
  aws cloudformation create-change-set --stack-name probe --change-set-name probe \
      --change-set-type CREATE --template-body file:///tmp/sam-local-uri.yaml \
      --capabilities CAPABILITY_IAM --region us-east-1
  aws cloudformation describe-change-set --stack-name probe --change-set-name probe \
      --region us-east-1
```

```
Status:          FAILED
ExecutionStatus: UNAVAILABLE
Changes:         0
StatusReason:    Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application
                 Specification document. Number of errors found: 1. Resource with id
                 [OrdersStateMachine] is invalid. 'DefinitionUri' is not a valid S3 Uri of the form
                 's3://bucket/key' with optional versionId query parameter.
```

`sam deploy` and `aws cloudformation package` rewrite the path to `s3://` before submitting, which
is why the deployed stacks above all carry an S3 location.

The `versionId` query parameter that sentence mentions is real, and it is split into a third field.
The same probe with `DefinitionUri: s3://bucket/key?versionId=abc123` reached `CREATE_COMPLETE`
with 1 change, and its `Processed` template reads:

```json
"DefinitionS3Location": {
  "Bucket": "cfn-snippets",
  "Key": "definition.asl.json",
  "Version": "abc123"
}
```

That probe declared no `Tags` at all, and the expansion still emitted exactly one:
`[{"Key": "stateMachine:createdBy", "Value": "SAM"}]`.

## Actual behavior

`SamTransformProcessor.expandSamTemplate` switches on four SAM types only:

```java
// SamTransformProcessor.java:87-98
case "AWS::Serverless::Function"    -> expandServerlessFunction(...);
case "AWS::Serverless::SimpleTable" -> expandServerlessSimpleTable(...);
case "AWS::Serverless::Api"         -> expandServerlessApi(...);
case "AWS::Serverless::HttpApi"     -> expandServerlessHttpApi(...);
default -> LOG.debugv("Unsupported SAM resource type: {0} ({1})", type, logicalId);
```

The `default` arm logs at debug and leaves the resource in the template with its SAM type intact.
It then reaches `CloudFormationResourceProvisioner`, which has no arm for
`AWS::Serverless::StateMachine` either and stubs it:

```java
// CloudFormationResourceProvisioner.java:522-524
LOG.debugv("Stubbing unsupported resource type: {0} ({1})", resourceType, logicalId);
resource.setPhysicalId(logicalId + "-" + UUID.randomUUID().toString().substring(0, 8));
resource.getAttributes().put("Arn", "arn:aws:stub:::" + logicalId);
```

The stack lands `CREATE_COMPLETE`. Three consequences, in order of how hard they are to notice:

1. `stepfunctions:ListStateMachines` is empty. Nothing to start, nothing to test.
2. `Fn::GetAtt <logicalId>.Arn` resolves to `arn:aws:stub:::<logicalId>`, and every downstream
   resource consumes it happily: an EventBridge rule target, an IAM policy resource, a Lambda
   environment variable. The template is wired to an ARN that names no state machine.
3. Nothing in the API says so. `DescribeStacks` is `CREATE_COMPLETE`, `DescribeStackResources`
   lists the resource as `CREATE_COMPLETE`, and the only trace is a debug log line.

A `DefinitionUri` still pointing at a local path is not refused either: the same stub path swallows
it, so the divergence from AWS is silent on both the unpackaged and the packaged template.

## Reproduction

```sh
docker run --rm -d -p 4566:4566 --name floci ghcr.io/floci-io/floci:latest

cat > /tmp/sam-sfn.yaml <<'YAML'
Resources:
  OrdersStateMachine:
    Type: AWS::Serverless::StateMachine
    Properties:
      Name: orders-sm
      Type: EXPRESS
      Role: arn:aws:iam::000000000000:role/orders-sfn-role
      DefinitionUri: s3://asl-definitions/orders.asl.json
      Tags:
        owner: payments
Outputs:
  MachineArn:
    Value: !GetAtt OrdersStateMachine.Arn
YAML

aws --endpoint-url http://localhost:4566 s3 mb s3://asl-definitions
cat > /tmp/orders.asl.json <<'JSON'
{"StartAt":"Done","States":{"Done":{"Type":"Pass","End":true}}}
JSON
aws --endpoint-url http://localhost:4566 s3 cp /tmp/orders.asl.json s3://asl-definitions/orders.asl.json

    --stack-name sam-sfn --template-body file:///tmp/sam-sfn.yaml \
    --capabilities CAPABILITY_IAM
aws --endpoint-url http://localhost:4566 cloudformation describe-stacks --stack-name sam-sfn \
    --query 'Stacks[0].[StackStatus,Outputs]'
aws --endpoint-url http://localhost:4566 stepfunctions list-state-machines
```

Captured on `main` at `a7b2b794`. Every call returned exit 0; `create-stack` answered with a
`StackId` and nothing else. The last two calls:

```
$ aws cloudformation describe-stacks --stack-name sam-sfn --query 'Stacks[0].[StackStatus,Outputs]'
[
    "CREATE_COMPLETE",
    [
        {
            "OutputKey": "MachineArn",
            "OutputValue": "arn:aws:stub:::OrdersStateMachine"
        }
    ]
]

$ aws stepfunctions list-state-machines
{
    "stateMachines": []
}
```

The same template against real AWS produces a `AWS::StepFunctions::StateMachine` whose
`Fn::GetAtt .Arn` is a real `arn:aws:states:` ARN, per the `get-template --template-stage Processed`
receipts above.

## Scale

20 source `template.yaml` files across 20 repositories in one organisation declare **174**
`AWS::Serverless::StateMachine` resources. Method, re-runnable:

```sh
grep -rl "AWS::Serverless::StateMachine" --include=template.yaml . | grep -v '\.aws-sam' \
  | xargs grep -h -c "Type: AWS::Serverless::StateMachine" | awk '{s+=$1} END {print s}'
```

Every one of the 174 is stubbed today. Two further counts from the same corpus: 174 of 174 declare a
`DefinitionUri` that is a local path before packaging (0 declare `s3://` directly, 0 use an inline
`Definition`), and 0 declare `Policies` without a `Role`.

No state machine was created in the AWS account for this report. The `get-template` quotes come
from stacks that were already deployed. The two change-set probes created only change sets, never
executed, on two stacks that stayed in `REVIEW_IN_PROGRESS`; both stacks were deleted and
`describe-stacks` now answers `Stack with id floci-probe-local-uri does not exist` for each.

facuparedes:
## Summary

`SamTransformProcessor.expandSamTemplate` switched on four SAM resource types. An
`AWS::Serverless::StateMachine` fell through the `default` arm, kept its `AWS::Serverless::` type,
and was stubbed by `CloudFormationResourceProvisioner` with a physical id of `<logicalId>-<8 hex>`
and an attribute `Arn` of `arn:aws:stub:::<logicalId>`. The stack reported `CREATE_COMPLETE`,
`ListStateMachines` was empty, and `Fn::GetAtt <logicalId>.Arn` handed every downstream resource an
ARN that named no state machine. This PR expands the resource the way the SAM transform does, with
the shapes taken from a deployed stack's processed template and from change-set probes against real
AWS.

Closes #2883

- `SamTransformProcessor.expandServerlessStateMachine` emits `AWS::StepFunctions::StateMachine` and
  maps the eight properties the transform maps, copying nodes with `deepCopy()` so an intrinsic is
  never flattened by `asText()`.
- `SamTransformProcessor.samTagsToCfnTags` turns SAM's `Tags` map into the native list of
  `{Key, Value}` and emits `{"Key": "stateMachine:createdBy", "Value": "SAM"}` first, unconditionally.
- `SamTransformProcessor.copyResourceLevelAttributes` carries every resource-level key other than
  `Type` and `Properties`, so `Metadata` survives. It runs from all five `AWS::Serverless::*` arms,
  not only the new one.
- `SamTransformProcessor.resolveDefinitionUriOrThrow` splits `s3://bucket/key?versionId=v` into
  `{Bucket, Key, Version}` and throws AWS's own sentence for anything it cannot split into a literal
  `Bucket` and `Key`. It is shared with the `AWS::Serverless::HttpApi` arm, which gains the same
  split and the same rejection.
- `SamTransformProcessor.rejectBothDefinitionSources` enforces the mutual exclusion AWS enforces:
  a state machine declaring neither `Definition` nor `DefinitionUri` is rejected, and so is one
  declaring both; an HttpApi declaring both `DefinitionUri` and `DefinitionBody` is rejected. Each
  resource type names its two properties in AWS's own order, which differs between them.
- `CloudFormationService.createChangeSet` reports a failed SAM transform as a change set with
  `Status: FAILED`, `ExecutionStatus: UNAVAILABLE` and zero changes, carrying the transform's own
  sentence in `StatusReason`, instead of raising an HTTP 400 out of `DescribeChangeSet`.
  `CloudFormationQueryHandler` renders `StatusReason`, which `XmlBuilder.elem` omits when null, so
  the response shape is unchanged for a successful change set.
- `CloudFormationService.executeChangeSetForRequest` refuses a change set that is not `AVAILABLE`
  with `InvalidChangeSetStatus`. The internal path `CreateStack` uses is unaffected, so a bad
  template still reaches `CREATE_FAILED` there.
- `CloudFormationResourceProvisioner.findStateMachine` treats the stub physical-id shape as "does
  not exist" at the one call site that needs it, so a stack written by an earlier floci build
  survives its next update instead of failing on `InvalidArn`.

**This turns previously green runs red: every `AWS::Serverless::StateMachine` resource in the corpus
I measured declares a local `DefinitionUri`, and every one of them changes from a silently
stubbed `CREATE_COMPLETE` to a failure that names the resource.** Real AWS refuses the same
templates; `sam deploy` and `aws cloudformation package` rewrite the path to `s3://` before
submitting, which is the shape this PR provisions. The measurement is re-runnable:

```sh
grep -rl "AWS::Serverless::StateMachine" --include=template.yaml . | grep -v '\.aws-sam' \
  | xargs grep -h -c "Type: AWS::Serverless::StateMachine" | awk '{s+=$1} END {print s}'
```

The mutual-exclusion rules add no further red for a state machine: `CloudFormationResourceProvisioner`
already refused both shapes at provisioning time, with floci's own wording
(`A state machine definition is required`, `Specify exactly one of Definition, DefinitionString, or
DefinitionS3Location`). This PR moves that failure forward into the transform and replaces the
wording with AWS's.

**The one shape this PR turns red beyond those is an `AWS::Serverless::HttpApi` declaring both
`DefinitionBody` and `DefinitionUri`, which previously took `DefinitionBody` and provisioned. That
shape does not occur in the same corpus: no HttpApi resource I measured declares both, and the one
that exists declares neither, which stays accepted.**

## Wire-fidelity details

Every row below was read from a live account, `us-east-1`, AWS CLI 2.33.7. The stack was deployed by
`sam deploy` and read, not created, except for the change-set probes, which created only change sets
and were deleted afterwards, each proved gone with a `describe-stacks` answering `does not exist`.

```sh
# Static keys in the environment outrank AWS_PROFILE; without the unset the call fails with
# InvalidClientTokenId instead of reaching AWS.
  aws cloudformation get-template --stack-name <stack> --template-stage Processed --region us-east-1
```

| Measured | AWS | Test that pins it |
|---|---|---|
| resource type after expansion | `AWS::StepFunctions::StateMachine`, no exceptions, 0 `AWS::Serverless::` left | `expandSamTemplate_stateMachineBecomesStepFunctionsStateMachine` |
| `Name` | `StateMachineName` | same |
| `Type` | `StateMachineType` | same |
| `Role` | `RoleArn` | same |
| `Logging` | `LoggingConfiguration`, shape untouched | same |
| `Tracing` | `TracingConfiguration`, shape untouched | same |
| `DefinitionSubstitutions` | same name, unchanged | `samStateMachine_withDefinitionSubstitutions_resolvesPlaceholders` |
| `DefinitionUri: s3://bucket/key` | `DefinitionS3Location {Bucket, Key}` | `samStateMachine_createsStepFunctionsStateMachine` |
| `DefinitionUri: s3://bucket/key?versionId=v` | `DefinitionS3Location {Bucket, Key, Version}`, accepted | `expandSamTemplate_stateMachineWithVersionedDefinitionUriSplitsVersion` |
| an S3 key containing a literal `?` not followed by `versionId=` | kept in `Key` | `expandSamTemplate_stateMachineWithQuestionMarkInKeyIsAccepted` |
| `Tags` map | list of `{Key, Value}`, `stateMachine:createdBy` = `SAM` first, no exceptions | `expandSamTemplate_stateMachineTagValueKeepsItsJsonType` |
| no `Tags` declared at all | exactly one tag, `stateMachine:createdBy` | `expandSamTemplate_stateMachineWithEmptyTagMapEmitsOnlyTheSamTag` |
| a tag value that is a number or an intrinsic | keeps its JSON type | `expandSamTemplate_stateMachineTagValueKeepsItsJsonType` |
| `Name`, `Role` written as `Fn::Sub` / `Fn::GetAtt` | stay unevaluated intrinsics | `expandSamTemplate_stateMachinePreservesIntrinsicNameAndRole` |
| resource-level keys | `Type`, `Properties`, `Metadata`, no exceptions, `Metadata` carried from the original template | `expandSamTemplate_stateMachinePreservesMetadataAndDependsOn`, `expandSamTemplate_functionPreservesMetadata` |
| `DefinitionUri` that is a local path, via `CreateChangeSet` | `Status: FAILED`, `ExecutionStatus: UNAVAILABLE`, `Changes: 0`, `StatusReason: Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [OrdersStateMachine] is invalid. 'DefinitionUri' is not a valid S3 Uri of the form 's3://bucket/key' with optional versionId query parameter.` | `samStateMachine_withLocalDefinitionUri_createChangeSetReportsFailedStatus`, `expandSamTemplate_stateMachineWithLocalDefinitionUriIsRejected` |
| `ExecuteChangeSet` on that change set | `InvalidChangeSetStatus`, `ChangeSet [<arn>] cannot be executed in its current status of [FAILED]` | `samStateMachine_withLocalDefinitionUri_executeChangeSetIsRejected` |
| a state machine declaring neither `Definition` nor `DefinitionUri` | `Number of errors found: 1. Resource with id [OrdersStateMachine] is invalid. Either 'Definition' or 'DefinitionUri' property must be specified.` | `expandSamTemplate_stateMachineWithNeitherDefinitionNorDefinitionUriIsRejected` |
| a state machine declaring both | `Number of errors found: 1. Resource with id [OrdersStateMachine] is invalid. Specify either 'Definition' or 'DefinitionUri' property and not both.` | `expandSamTemplate_stateMachineWithBothDefinitionAndDefinitionUriIsRejected` |
| an HttpApi declaring both `DefinitionUri` and `DefinitionBody` | `Number of errors found: 1. Resource with id [MyHttpApi] is invalid. Specify either 'DefinitionUri' or 'DefinitionBody' property and not both.`, the property order reversed against the state machine's | `expandSamTemplate_httpApiWithBothDefinitionBodyAndDefinitionUriIsRejected` |
| an HttpApi declaring neither | accepted: `CREATE_COMPLETE`, `AVAILABLE`, 2 changes, `AWS::ApiGatewayV2::Api` plus a default `AWS::ApiGatewayV2::Stage` | `createStack_samHttpApiExpandsToRealApiRouteAndAuthorizer` |
| an HttpApi whose `DefinitionUri` is a local path | `Number of errors found: 1. Resource with id [MyHttpApi] is invalid. 'DefinitionUri' is not a valid S3 Uri of the form 's3://bucket/key' with optional versionId query parameter.` | `expandSamTemplate_httpApiWithLocalDefinitionUriIsRejected` |

Both new ASL definitions pass `aws stepfunctions validate-state-machine-definition` with
`result: OK`.

Reverting the production hunks and running `SamTransformProcessorTest` plus
`SamTransformIntegrationTest` produced `Tests run: 63, Failures: 6, Errors: 1`, first failure
`samStateMachine_withLocalDefinitionUri_createChangeSetReportsFailedStatus:1158`,
`Expected status code <200> but was <400>`. Four later fixes were red-proved one at a time as well:
the `ExecuteChangeSet` guard (`Expected status code <400> but was <200>`), the `?` in an S3 key
(`'DefinitionUri' is not a valid S3 Uri`), `Metadata` on a Function (`expected: <OrdersFunction> but
was: <>`), and the malformed-ARN cleanup (`Expected: a string containing "<StackStatusReason>"`).

The three mutual-exclusion tests were red-proved together, by restoring `SamTransformProcessor` to
the previous commit with the new tests in place: `Tests run: 56, Failures: 3`, all three
`Expected AwsException to be thrown, but nothing was thrown.`

## Deliberate scope notes

- **`AWS::Serverless::HttpApi` is inside this PR even though #2883 names only the state machine.**
  The rejection of an unresolvable `DefinitionUri` is one shared helper, and leaving HttpApi on the
  old path would mean two contradictory answers to the same malformed value in the same file. The
  mutual-exclusion rule was measured on the same wire trip and added beside it rather than deferred
  to a second PR. Its blast radius is stated above and is 0 in the measured corpus.
- **The asymmetry between the two types is measured, not an oversight.** A state machine declaring
  neither definition source is rejected by AWS; an HttpApi declaring neither is accepted by AWS and
  expands to a bare `AWS::ApiGatewayV2::Api` plus a default `Stage`. Both sides were probed with
  `CreateChangeSet` this round, and floci reproduces each.
- **`applyDefinitionUri` was renamed `resolveDefinitionUriOrThrow`**, and a null guard was deleted
  from `samUriToS3Location`. The guard was unreachable: `ObjectNode.path()` never returns Java null,
  and both callers now check presence before calling. `grep -rn "samUriToS3Location" src/` returns
  one definition and one call.
- **`Globals.StateMachine` is not merged.** SAM's `Globals` schema has one key for state machines
  and merging it would imply support for a `Globals.StateMachine.Role` that AWS rejects. The comment
  at the switch arm says so.
- **No execution role is synthesized from `Policies`.** The Lambda arm does synthesize one, so the
  asymmetry is deliberate: no measured state machine declares `Policies` without a `Role`,
  and a state machine with neither still fails at `RoleArn is required for a state machine`, which
  is loud rather than silent. Implementing it is a separate change.
- **`Events`, `PermissionsBoundary`, `AutoPublishAlias` and `UseAliasAsEventTarget` are not
  handled**, and the javadoc says so, matching the file's own precedent for `Policies`. None of the
  measured state machines declare `Events`.
- **`AWS::Serverless::Api` still ignores its own `DefinitionUri` and `DefinitionBody`.** Only the
  `HttpApi` arm gained the shared resolution. Extending it needs the `Api` arm's own AWS probes,
  which were not run.
- **`Metadata.SamResourceId` is carried, never synthesized.** It is written by the SAM CLI and is
  already present in the original template on every one I measured; CloudFormation only copies it.
- **`GetTemplate` still ignores `TemplateStage` on this branch**, so no client can ask floci for the
  processed template and observe the shapes this PR produces. That gap predates this PR and is
  closed by #2886, which is independent of this one.
- **The SAM transform still runs twice per change set**, once as the preflight that produces the
  `StatusReason` and once at execution. Collapsing them requires the change set to carry the
  expanded template into `executeTemplate`, which changes that method's contract. Left alone
  deliberately.
- **`Number of errors found: 1.` is a constant.** `expandSamTemplate` throws on the first invalid
  resource rather than accumulating, so the count is 1 by construction; the comment says so. Real
  SAM validates the whole document and can report more.
- **A `Tags` value that is itself an intrinsic** (`Tags: !If [...]`) is iterated as if it were the
  tag map. AWS's behaviour there was not measured, and no template in the corpus uses that shape, so
  nothing was invented to cover it.
- Two tests fail on the development machine with `java.net.SocketTimeoutException: Read timed out`
  against a Docker container, both before and after this branch, and neither touches a file it
  changes: `CloudFormationIntegrationTest#createStack_ecrAutoName_isLowercase` and
  `SamTransformIntegrationTest#samFunction_withAutoPublishAlias_createsVersionAndAlias`. They are
  excluded by name from the local runs and reported rather than fixed.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] Tests added for the new behaviour, unit and integration
- [ ] Every new test proved red before its production hunk landed, with the failure count recorded.
      Two exceptions, both stated in their own commit message: `expandSamTemplate_httpApiWithLocalDefinitionUriIsRejected`
      and `samHttpApi_definitionUriCreatesApiGatewayV2Routes` pass unmodified against the previous
      commit's production code. They are coverage, not pins: the first mirrors the state machine's
      local-path rejection on the HttpApi arm, the second restores the only end-to-end proof that a
      literal `s3://` `DefinitionUri` creates routes.
- [x] Every new state machine definition passes `aws stepfunctions validate-state-machine-definition`
- [x] Expected values are stated as literals, never re-derived through production code
- [x] Behaviour measured against real AWS, not paraphrased from documentation or from SAM's source
- [x] Every AWS resource created while measuring was deleted, each proved gone with `describe-stacks`
- [ ] Full local suite green: two pre-existing Docker socket-timeout failures, listed above


hectorvent:
@facuparedes thanks, this is careful work, and matching SAM's own two rejection messages verbatim (`generators.py`) rather than inventing wording is the right instinct. I ran it locally: 848 CloudFormation tests green, and it also merges and passes cleanly alongside #2886, so those two are not a conflict pair.

**(follow-up, separate PR)** `Policies` is not handled for `AWS::Serverless::StateMachine`, so one declaring `Policies` expands with no `RoleArn`. Real SAM builds an execution role there, and `AWS::Serverless::Function` in this same class already synthesises one from `Policies`. You do disclose it in the javadoc, so this is only about closing the asymmetry later.

**(follow-up, separate PR)** Inline `Definition` is emitted as `Definition`, where SAM emits `DefinitionString` plus generated `DefinitionSubstitutions`. Both are valid on the native type, so nothing breaks, but the Processed template differs from AWS's, which your #2886 is what makes visible.

No blockers from my side.

hectorvent (approved):
Verified locally: 848 CloudFormation tests green at 9076a07a, and the SAM rejection messages match the transform source in samtranslator. Also merges and passes cleanly alongside #2886. The two follow-ups in my comment are not blockers.

PR #2886, opened 2026-09-01
fix(cloudformation): merge Fn::Transform AWS::Include before provisioning, and honour GetTemplate TemplateStage

### LINKED ISSUE #2884: [BUG] CloudFormation (CFN): `Fn::Transform` / `AWS::Include` is never applied, so the snippet's keys are missing and the template deploys with the transform node intact
## Service

CloudFormation (`Fn::Transform`, `AWS::Include`), Step Functions

## AWS API Action

`CreateStack`, `CreateChangeSet`, `DescribeChangeSet`, `GetTemplate`, `GetTemplateSummary`

## Expected behavior

CloudFormation replaces the `Fn::Transform` node with the snippet's keys before the template is
provisioned. Captured from a live stack:

```sh
# Static keys in the environment outrank AWS_PROFILE; without the unset the call fails with
# InvalidClientTokenId instead of reaching AWS.
  aws cloudformation get-template --stack-name my-stack \
      --template-stage Original --region us-east-1
  aws cloudformation get-template --stack-name my-stack \
      --template-stage Processed --region us-east-1
```

| Measured on a deployed stack | `Original` | `Processed` |
|---|---|---|
| `Fn::Transform` nodes | 41 | 0 |
| distinct `Name` | `AWS::Include` | n/a |
| distinct `Parameters.Location` | 1, `s3://cfn-snippets/substitutions.yaml` | n/a |
| position of all 41 | `Resources.<id>.Properties.DefinitionSubstitutions` | n/a |

`Original`, resource `OrdersNotifierStateMachine`:

```json
"DefinitionSubstitutions": {
  "OrdersTableName": { "Fn::Sub": "${Environment}-Orders" },
  "OrderMembersTableName": { "Fn::Sub": "${Environment}-OrderMembers" },
  "EventBusARN": { "Fn::Sub": "..." },
  "SchedulerLambdaArn": { "Fn::Sub": "..." },
  "DownstreamApiLambdaArn": { "Fn::Sub": "..." },
  "UserServiceApiEndpoint": { "Fn::Sub": "..." },
  "Fn::Transform": {
    "Name": "AWS::Include",
    "Parameters": { "Location": "s3://cfn-snippets/substitutions.yaml" }
  }
}
```

`Processed`, same node. The `Fn::Transform` key is gone, the six sibling keys keep their position
and their order, and the snippet's five keys are appended after them:

```
key order: OrdersTableName, OrderMembersTableName, EventBusARN,
           SchedulerLambdaArn, DownstreamApiLambdaArn,
           UserServiceApiEndpoint,
           expandItems, expandItem, expandAttributes,
           expandBatchGetResponse, pageGuard
```

One spliced value, verbatim, with its line breaks preserved as `\n`:

```json
"expandItems": {
  "Fn::Sub": [
    "{% (\n  $c := function($v) { $exists($v.S) ? $v.S : $exists($v.N) ? $number($v.N) : ... };\n  $ui := function($item) { $merge($each($item, function($val, $key) { { $key: $c($val) } })) };\n  ${ExpandFunction}\n) %}",
    { "ExpandFunction": "$count($states.result.Items) > 0 ? [$map($states.result.Items, $ui)] : []" }
  ]
}
```

Two further facts the same pair of calls settles:

- The snippet is **not** re-serialized or folded. The `\n` escapes survive into the processed
  template exactly as the snippet wrote them.
- `GetTemplateSummary` does **not** expand the include, and it reports the transform in both the
  embedded and the top-level form. Three controlled probes, `--template-body`, region `us-east-1`,
  CLI 2.33.7:

  | template | `DeclaredTransforms` | `ResourceTypes` |
  |---|---|---|
  | one `AWS::StepFunctions::StateMachine`, no transform | `null` | `["AWS::StepFunctions::StateMachine"]` |
  | the same, plus an embedded `Fn::Transform` / `AWS::Include` in `DefinitionSubstitutions` | `["AWS::Include"]` | absent |
  | top-level `Transform: {Name: AWS::Include, Parameters: {Location: ...}}` | `["AWS::Include"]` | absent |

  The embedded form has no top-level `Transform` section and is still reported. The whole response
  for the second probe is `{"Parameters": [], "Version": "2010-09-09", "DeclaredTransforms":
  ["AWS::Include"]}`: with an unexpanded transform present, AWS omits `ResourceTypes` and
  `ResourceIdentifierSummaries` altogether. The same omission happens for
  `Transform: AWS::Serverless-2016-10-31`, so it is a property of unexpanded transforms, not of
  `AWS::Include`.

## Actual behavior

`CloudFormationTemplateEngine` evaluates ten intrinsics and `Fn::Transform` is not among them:

```
$ grep -o '"Fn::[A-Za-z]*"' CloudFormationTemplateEngine.java | sort -u
"Fn::Cidr"  "Fn::FindInMap"  "Fn::GetAZs"  "Fn::GetAtt"  "Fn::If"
"Fn::ImportValue"  "Fn::Join"  "Fn::Select"  "Fn::Split"  "Fn::Sub"
```

`AWS::Include` appears nowhere in `src/main`. The node is neither expanded nor rejected: it is
carried through provisioning as if it were an ordinary key. For the resource above that means
`DefinitionSubstitutions` reaches `CreateStateMachine` with a key literally named `Fn::Transform`
and without any of the five keys the snippet defines. The definition then references
`${expandItems}`, no substitution matches it, and the placeholder survives into the
stored ASL, where it fails at run time instead of at deploy time.

Nothing in the API reports the omission. The stack reaches `CREATE_COMPLETE`, `GetTemplate` returns
the unexpanded node, and there is no warning.

## Reproduction

```sh
docker run --rm -d -p 4566:4566 --name floci ghcr.io/floci-io/floci:latest

cat > /tmp/snippet.yaml <<'YAML'
orderTableName: orders
YAML
aws --endpoint-url http://localhost:4566 s3 mb s3://cfn-snippets
aws --endpoint-url http://localhost:4566 s3 cp /tmp/snippet.yaml s3://cfn-snippets/subs.yaml

cat > /tmp/include.yaml <<'YAML'
Resources:
  OrdersStateMachine:
    Type: AWS::StepFunctions::StateMachine
    Properties:
      StateMachineName: orders-sm
      RoleArn: arn:aws:iam::000000000000:role/orders-sfn-role
      DefinitionString: |
        {"StartAt":"Read","States":{"Read":{"Type":"Pass","Result":"${orderTableName}","End":true}}}
      DefinitionSubstitutions:
        alreadyWrittenKey: alreadyWrittenValue
        Fn::Transform:
          Name: AWS::Include
          Parameters:
            Location: s3://cfn-snippets/subs.yaml
YAML

    --stack-name include-probe --template-body file:///tmp/include.yaml
aws --endpoint-url http://localhost:4566 cloudformation get-template --stack-name include-probe
aws --endpoint-url http://localhost:4566 stepfunctions describe-state-machine \
    --state-machine-arn arn:aws:states:us-east-1:000000000000:stateMachine:orders-sm \
    --query definition
```

Captured on `main` at `a7b2b794`. Every call returned exit 0 and the stack reached
`CREATE_COMPLETE` with a null `StackStatusReason`. `GetTemplate` hands the body back as the
submitted YAML string, so it cannot be queried with a JMESPath path into the template; the whole
response is quoted instead, and the `Fn::Transform` node is still in it with no `orderTableName`
anywhere:

```
$ aws cloudformation get-template --stack-name include-probe
{
    "TemplateBody": "AWSTemplateFormatVersion: '2010-09-09'\nResources:\n  OrdersStateMachine:\n    Type: AWS::StepFunctions::StateMachine\n    Properties:\n      StateMachineName: orders-sm\n      RoleArn: arn:aws:iam::000000000000:role/orders-sfn-role\n      DefinitionString: |\n        {\"StartAt\":\"Read\",\"States\":{\"Read\":{\"Type\":\"Pass\",\"Result\":\"${orderTableName}\",\"End\":true}}}\n      DefinitionSubstitutions:\n        alreadyWrittenKey: alreadyWrittenValue\n        Fn::Transform:\n          Name: AWS::Include\n          Parameters:\n            Location: s3://cfn-snippets/subs.yaml\n"
}

$ aws stepfunctions describe-state-machine \
      --state-machine-arn arn:aws:states:us-east-1:000000000000:stateMachine:orders-sm \
      --query definition
"{\"StartAt\":\"Read\",\"States\":{\"Read\":{\"Type\":\"Pass\",\"Result\":\"${orderTableName}\",\"End\":true}}}\n"
```

`${orderTableName}` survives into the stored definition, unsubstituted, and the stack still reports
success.

Against real AWS the same template yields `DefinitionSubstitutions` with `alreadyWrittenKey` first
and `orderTableName` appended, and the definition resolves.

## Scale

**123** `Fn::Transform` blocks across **32** source `template.yaml` files in one organisation, in
both the block form and the flow form. Method, re-runnable:

```sh
grep -rl "Fn::Transform" --include=template.yaml . | grep -v '\.aws-sam' \
  | xargs grep -h -c "Fn::Transform" | awk '{s+=$1} END {print s}'
```

On the single deployed stack quoted above the count is 41, all at the same position and all reading
the same S3 object.

No state machine was created in the AWS account for this report: every quote comes from
`get-template` and `get-template-summary` on stacks that were already deployed.

facuparedes:
## Summary

`CloudFormationTemplateEngine` evaluates ten intrinsics and `Fn::Transform` was not among them, so
an `AWS::Include` node was neither expanded nor rejected: it reached provisioning as an ordinary
key. A state machine whose `DefinitionSubstitutions` came from a snippet therefore reached
`CreateStateMachine` with a key literally named `Fn::Transform` and without any of the snippet's
keys, leaving every `${placeholder}` unsubstituted in the stored definition while the stack still
reported `CREATE_COMPLETE`. This PR merges the snippet before provisioning, and makes `GetTemplate`
answer the stage the caller asked for, so the merged template is observable. Both shapes were taken
from a deployed stack's original and processed templates.

Closes #2884

- `AwsIncludeProcessor.mergeIncludes` walks the template and splices each `Fn::Transform` whose
  `Name` is `AWS::Include`, reading the snippet through `S3Service`. It returns the same `JsonNode`
  reference when the template carries no include, which is the contract `executeTemplate` uses to
  decide whether to re-serialize, and two tests pin that with `assertSame` and `assertNotSame`.
- `AwsIncludeProcessor.spliceInclude` removes the `Fn::Transform` key and appends the snippet's
  keys, so sibling keys keep their position and order. It receives the whole `Fn::Transform` node,
  so a rejection can name that node whatever shape `Parameters` has, including a scalar, a list, an
  explicit null and a missing key.
- `CloudFormationService.executeTemplate` and `computeChangeSetChanges` merge before the SAM
  transform check, so a snippet may carry SAM resources, and `AWS::CloudFormation::Stack` children
  inherit the merge through the same call.
- `CloudFormationService.getTemplate` accepts `TemplateStage`, validates it against AWS's enum
  before it looks the stack up, and answers `originalTemplateBody` for `Original` and the merged
  body for `Processed`. `Original` stays the default when the parameter is absent. The change-set
  baseline keeps the merged body, so a byte-identical resubmission still reports no change, and
  `StackUpdateSnapshot` restores both bodies on a rolled-back update.
- `CloudFormationQueryHandler` reads `TemplateStage` off the query and renders `StagesAvailable`,
  which AWS returns on every `GetTemplate` response.
- `CloudFormationService.getTemplateSummary` reports `AWS::Include` in `DeclaredTransforms` for the
  embedded `Fn::Transform` form, not only for a top-level `Transform` section.
- `validateConditionDependencies` skips a template carrying an unexpanded `Fn::Transform`, the same
  early return it already had for SAM, so a `Conditions` section spliced in from a snippet is not
  read as missing.

## Wire-fidelity details

Every row below was read from a live account, `us-east-1`, AWS CLI 2.33.7. The deployed stack was
read, not created; the `get-template-summary` rows are `--template-body` probes and the
`TemplateStage` rows are read-only calls, so nothing was created for any of them.

```sh
# Static keys in the environment outrank AWS_PROFILE; without the unset the call fails with
# InvalidClientTokenId instead of reaching AWS.
  aws cloudformation get-template --stack-name <stack> --template-stage Processed --region us-east-1
```

| Measured | AWS | Test that pins it |
|---|---|---|
| `Fn::Transform` nodes, original template | 41, all `Name: AWS::Include`, all one `s3://` `Location`, all at `Resources.<id>.Properties.DefinitionSubstitutions` | `blockFormIncludeMergesSnippetKeysIntoSubstitutions`, `flowFormIncludeMergesSnippetKeysIntoSubstitutions` |
| `Fn::Transform` nodes, processed template | 0 | same |
| two resources each carrying their own `Fn::Transform` | both merge | `twoResourcesEachWithTheirOwnFnTransformBothMerge` |
| key order at a merged position | six sibling keys first, in their original order, then the five snippet keys appended | `includeKeepsSiblingKeysOfTheTransformAndAppendsSnippetKeysAfterThem` |
| a spliced multi-line `Fn::Sub` value | line breaks preserved as `\n`; nothing folded or re-serialized | `blockFormIncludeMergesSnippetKeysIntoSubstitutions` (asserts the scalar verbatim) |
| the merged substitution reaching the state machine | placeholder resolved in the stored definition | `includedSubstitutionsReachTheStateMachineDefinition` |
| `get-template-summary`, no transform | `DeclaredTransforms: null`, `ResourceTypes` populated | control probe |
| `get-template-summary`, embedded `Fn::Transform`, no top-level `Transform` | `DeclaredTransforms: ["AWS::Include"]`, snippet not expanded | `getTemplateSummaryDoesNotExpandTheInclude` |
| `get-template-summary`, top-level `Transform: {Name: AWS::Include}` | `DeclaredTransforms: ["AWS::Include"]` | same |
| `GetTemplate` with no `TemplateStage` | the body as submitted, that is `Original` | `getTemplateReturnsTheSubmittedYamlWhileTheChangeSetBaselineStaysMerged` |
| `GetTemplate --template-stage Bogus` | `ValidationError`, `Value 'Bogus' at 'templateStage' failed to satisfy constraint: Member must satisfy enum value set: [Processed, Original]`, the apostrophes entity-escaped to `&apos;` by the Query protocol | `getTemplateWithInvalidStageRejectsWithTheAwsEnumValidationError` |
| `GetTemplate --template-stage ""` | the same enum error, not a fallback to `Original` | `getTemplateWithEmptyStringStageRejectsWithTheAwsEnumValidationError` |
| `GetTemplate --template-stage " "` | the same enum error, echoing the whitespace value | same |
| an absent stack with an invalid stage | only the stage error; the stack is never looked up | `getTemplateWithInvalidStageValidatesBeforeCheckingWhetherTheStackExists` |
| an absent stack with a valid stage | `Stack with id <name> does not exist` | same |
| `StagesAvailable` | always both, in the order `["Original", "Processed"]`, which is not the order the validation error lists | `getTemplateReturnsTheSubmittedYamlWhileTheChangeSetBaselineStaysMerged` |
| a snippet that itself uses `AWS::Include` | rejected | `nestedIncludeInsideASnippetIsRejected` |
| a snippet carrying a different macro | passes through untouched | `snippetContainingAForeignTransformIsNotRejected` |
| a `Location` that is not `s3://bucket/key` | rejected, naming the value | `includeWithNonS3LocationIsRejected`, `locationAsAMappingNamesTheNodeInTheRejectionMessage`, `missingLocationNamesTheParametersObjectInsteadOfLeavingAnEmptyTail` |
| a `Location` whose object cannot be read | S3's own error code and status | `includeWithUnreadableSnippetFailsTheStack`, `stackWithUnreadableSnippetFailsCreation` |

Every test above was proved red before its production hunk landed. Reverting the `mergeIncludes`
call lines produced `Tests run: 5, Failures: 4`, first failure
`AwsIncludeIntegrationTest#includedSubstitutionsReachTheStateMachineDefinition:121`, expected
`"substituted-value"` in the stored definition and got the literal `${expandItems}`.
Four later fixes were red-proved one at a time, and the multiplicity test was proved by a mutation
that stops after the first splice: `expected: <second-orders> but was: <>`, 1 failure of 12,
with the other 11 single-resource tests staying green.

The two `TemplateStage` validation hunks were red-proved separately, one revert each. Reverting only
the reordering left `getTemplateWithInvalidStageValidatesBeforeCheckingWhetherTheStackExists`
failing on `<Message>Stack with id <name> does not exist</Message>`. Reverting only the empty-value
rejection left `getTemplateWithEmptyStringStageRejectsWithTheAwsEnumValidationError` failing,
`Tests run: 2, Failures: 1`. Each test guards exactly one hunk, and the pre-existing test asserting
that an absent parameter still answers `Original` byte for byte stayed green through both reverts.

## Deliberate scope notes

- **The merge is a template rewrite ahead of intrinsic resolution, not a new intrinsic.**
  `CloudFormationTemplateEngine` is untouched, so `Fn::Transform` never becomes an eleventh case in
  `resolve`.
- **`Location` written as an intrinsic is rejected.** AWS's own documentation shows
  `{"Ref": "InputValue"}`, but `mergeIncludes` runs before `resolveDefaultParameters`, so no
  parameter value exists yet. The rejection names the node instead of leaving an empty tail, and the
  deviation is listed in `docs/services/cloudformation.md`.
- **The top-level `Transform: {Name: AWS::Include}` form is reported in `DeclaredTransforms` but not
  spliced.** All 41 occurrences measured use the embedded form; splicing at the template root is a
  different merge point and belongs in its own change.
- **`ResourceTypes` is still emitted when a transform is unexpanded.** AWS omits `ResourceTypes` and
  `ResourceIdentifierSummaries` entirely in that case, and does the same for
  `Transform: AWS::Serverless-2016-10-31`, so the divergence is general rather than specific to
  `AWS::Include`. Fixing it would change SAM's behaviour too, which is outside this PR.
- **The stage is a `String` guarded by two constants, not an enum.** An enum would need a
  wire-value parse method plus two enum-ordered lists to preserve both measured AWS orderings, which
  is more machinery than the literals it replaces. Both orderings are kept as separate constants and
  each carries the measurement that produced it.
- **Key collision precedence is whatever `ObjectNode.setAll` does**, snippet wins. Neither AWS
  document defines it and it was not measured, so nothing was written down as a rule and no test
  pins it.
- **`DescribeChangeSet` re-fetches the snippet from S3 on every call**, so a snippet deleted between
  `CreateChangeSet` and `DescribeChangeSet` turns a read-only preview into an error. That is the
  cost of not caching the merged template on the change set; it predates the fix rounds and is left
  as is.
- **Reviewer concern, not fixed:** `CloudFormationService.templateStagesAvailable()` is a public
  accessor returning a constant with a single caller. Two reviewers flagged it as a middle man.
  It is left in place so the handler does not reach into the service's constants directly.
- **Two pre-existing divergences surfaced while reviewing this branch and are not touched by it.**
  `GetTemplate` with no `StackName` and a valid stage answers `Stack with id null does not exist`
  where AWS answers `StackName is required if ChangeSetName is not specified.`, and
  `DeclaredTransforms` comes back empty for a top-level `Transform` written in the object form,
  because only the `Fn::Transform` key is matched. Both predate this PR.
- Two tests fail on the development machine with `java.net.SocketTimeoutException: Read timed out`
  against a Docker container, both before and after this branch, and neither touches a file it
  changes: `CloudFormationIntegrationTest#createStack_ecrAutoName_isLowercase` and
  `SamTransformIntegrationTest#samFunction_withAutoPublishAlias_createsVersionAndAlias`. They are
  excluded by name from the local runs and reported rather than fixed.

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] Tests added for the new behaviour, unit and integration
- [x] Every new test proved red before its production hunk landed, with the failure count recorded
- [x] Expected values are stated as literals, never re-derived through production code
- [x] Behaviour measured against real AWS, not paraphrased from documentation
- [x] The two known deviations from AWS are documented in `docs/services/cloudformation.md`
- [x] No AWS resource was created while measuring
- [ ] Full local suite green: two pre-existing Docker socket-timeout failures, listed above


hectorvent:
@facuparedes thanks for this, it is unusually well evidenced. The wire-fidelity section with the exact CLI version, and the docs splitting what you measured from what you inferred, is the right way to write up a transform. I ran it locally as well: 789 CloudFormation tests green on your branch.

**(follow-up, separate PR)** On the intrinsic `Location` you documented as a deviation, there is more evidence than the AWS doc example. localstack resolves intrinsics in `Parameters` before handing them to the include transformer ([transformers.py](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/services/cloudformation/engine/transformers.py)), so a `Ref` there does work against it. Worth noting in the deviation entry, since it makes the gap look real rather than doc-only.

**(follow-up, separate PR)** botocore's GetTemplate model says "By default, CloudFormation specifies `Processed`" ([service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cloudformation/2010-05-15/service-2.json)), where this defaults to `Original`. I take your live measurement over the prose, but a sentence naming the contradiction would stop someone "correcting" it back.

No blockers from my side.

hectorvent (approved):
Verified locally at 661e9ca1: 789 CloudFormation tests green, plus the new AwsIncludeProcessorTest 18/0 and AwsIncludeIntegrationTest 10/0. Checked GetTemplate against botocore (TemplateStage enum, declared errors, StagesAvailable) and AWS::Include against localstack's transformer. Also tested merged with #2885 before that landed. The two follow-ups in my comment are not blockers.

PR #2887, opened 2026-09-01
fix(ec2): guard null group entries in DescribeNetworkInterfaces group-id filter

matanbaruch:
The `group-id` filter in `DescribeNetworkInterfaces` dereferences every group entry on an interface without a null check, so a single null entry makes the whole call return a 500.

`matchesFilter` runs `ni.getGroups().stream().anyMatch(g -> matchesValue(values, g.getGroupId()))`. A null element throws NPE at `g.getGroupId()`. The sibling cases in the same switch already guard for this (`attachment.instance-id` checks `getAttachment() != null`); `group-id` was the one that did not.

Why it matters: a Terraform-based AWS provider polls `DescribeNetworkInterfaces` filtered by `group-id` while deleting a security group, to wait until no interface still references it. A 500 there is not retryable cleanly, so the delete spins in a retry loop until the client times out. I hit this deleting a security group against floci 2.0.0 through the Pulumi AWS provider (which bridges the Terraform AWS provider): the delete never completed, and `docker logs` showed the NPE looping on `Ec2Service.describeNetworkInterfaces` (the native image reports it at the inlined `matchesFilters` call site).

Fix: guard each group entry, matching the null-tolerant style of the adjacent filters. A malformed group entry now skips instead of crashing the request.

Test: `Ec2ServiceTest#describeNetworkInterfacesGroupIdFilterSkipsNullGroupEntry` launches an instance, adds a null entry to its interface's group list, and asserts the `group-id` filter still returns the interface (it errored with the NPE before the guard).

Verified locally: `mvn test -Dtest=Ec2ServiceTest` (121 pass) and `-Dtest=Ec2IntegrationTest` (182 pass) on JDK 25.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -5340,7 +5340,7 @@ public class Ec2Service implements ContainerTeardown, ResourceProvider {
                 case "subnet-id" -> matchesValue(values, ni.getSubnetId());
                 case "vpc-id" -> matchesValue(values, ni.getVpcId());
                 case "group-id" -> ni.getGroups().stream()
-                        .anyMatch(g -> matchesValue(values, g.getGroupId()));
+                        .anyMatch(g -> g != null && matchesValue(values, g.getGroupId()));
                 case "status" -> matchesValue(values, ni.getStatus());
                 case "attachment.instance-id" -> ni.getAttachment() != null
                         && matchesValue(values, ni.getAttachment().getInstanceId());
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2ServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2ServiceTest.java
@@ -12,6 +12,7 @@ import io.github.hectorvent.floci.services.ec2.model.EbsBlockDevice;
 import io.github.hectorvent.floci.services.ec2.model.GroupIdentifier;
 import io.github.hectorvent.floci.services.ec2.model.Image;
 import io.github.hectorvent.floci.services.ec2.model.Instance;
+import io.github.hectorvent.floci.services.ec2.model.NetworkInterfaceListResult;
 import io.github.hectorvent.floci.services.ec2.model.IpPermission;
 import io.github.hectorvent.floci.services.ec2.model.Ipv6Range;
 import io.github.hectorvent.floci.services.ec2.model.SecurityGroupRule;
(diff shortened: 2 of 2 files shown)


pgermosen (commented):
@matanbaruch good regression test — it reproduces the actual crash rather than just asserting the guard exists.

I noticed the branch has been rebased since greptile's earlier pass, so I want to make sure the current shape is right rather than just deferring to its 5/5. The version it originally flagged had a list-level null check as well as the element-level one, and it correctly pointed out the list-level check was dead: the instance-attached describe path already copies `eni.getGroups()` into a new list before the filter ever runs, so a null list would blow up there first, never reaching this guard. Dropping it in the current commit is the right call, and I checked why rather than taking that on faith — a null groups list can't actually happen anywhere in this codebase right now. Both `NetworkInterface.groups` and `InstanceNetworkInterface.groups` default to a non-null `ArrayList`, and the only place a list gets built from parsed request params comes back from a helper that returns empty rather than null when nothing's there. So the only reachable case really is a null *entry* inside an otherwise-real list, which is exactly what you're guarding against and exactly what the test reproduces.

One thing adjacent to this that I don't think is yours to fix here, but is worth having on your radar: `matchesValue(List<String>, String)` still calls `value.matches(...)` with no null check on the value itself, so a `GroupIdentifier` that's a real, non-null list entry but happens to have a null `groupId` would still NPE through the same path your fix touches. That mirrors how the rest of `matchesFilter` already treats leaf strings elsewhere (`subnet-id`, `status`, `private-dns-name` aren't guarded either), and it's not the crash you actually hit and reproduced, so I'm not asking you to expand scope for it — just flagging it since it's sitting right next to the line you changed.

matanbaruch:
@pgermosen good catch on the `matchesValue` NPE. It's real: the `(List<String>, String)` overload calls `value.matches(...)` unguarded, so a non-null `GroupIdentifier` with a null `groupId` NPEs through the same path, and so would any null leaf (`subnet-id`, `status`, `private-dns-name`). The sibling `(String, List<String>)` overload already normalizes with `Objects.toString`, so I mirrored that.

Kept it out of this PR as you suggested. Opened #2921 off `main`: one-line normalize in the shared helper plus a regression test that puts the null-`groupId` entry ahead of a real group so the null path is actually exercised. It errors with `Cannot invoke "String.matches(String)" because "value" is null` before the fix and passes after; full `Ec2ServiceTest` stays green.

pgermosen (approved):
No changes needed on my end this round, the diff's the same as before. The one open item was whether matchesValue(List<String>, String) should also guard a null groupId on a real entry, and you handled that the right way, fixing it in its own PR instead of scope-creeping this one. Approved that one separately too. Good regression test here, reproduces the actual crash rather than just asserting the guard exists.

PR #2895, opened 2026-09-01
fix(apigateway): enforce AWS_IAM authorization on the execute-api data plane

### LINKED ISSUE #2810: [BUG] HTTP API routes with AWS_IAM authorization are recorded but never enforced, so unsigned requests reach the integration
## Service

API Gateway v2 (HTTP API)

## AWS API Action

`CreateRoute` / `ImportApi` with `AuthorizationType: AWS_IAM`, then execute-api dispatch

## Expected behavior

A route whose `authorizationType` is `AWS_IAM` requires a SigV4-signed request. An unsigned request is rejected with 403 before the integration runs.

## Actual behavior

The route is stored and reported as `AWS_IAM` by `GetRoutes`, but nothing enforces it: `ApiGatewayExecuteController` branches on `JWT` (around line 1475) and `CUSTOM` (around line 1483) only, with no `AWS_IAM` arm. An unsigned request reaches the integration exactly as if the route were `NONE`.

The v1 REST path has the same gap and says so in a comment (around line 752: "v1 dispatch does not implement AWS_IAM — invokeAuthorizer only handles CUSTOM").

This is a pre-existing control-plane/data-plane split rather than a regression: `CreateRoute` has always taken `authorizationType` straight from the request without validating that dispatch can enforce it.

## Why it is worth tracking separately

It came up in review of #2714, which added OpenAPI import for HTTP APIs. An imported `awsSigv4` security scheme resolves to an `AWS_IAM` route, giving a second way to reach the unenforced state. That PR settled for making the gap visible — the import emits a warning naming the route and scheme, so `failOnWarnings=true` rejects such a document — and deliberately did not touch dispatch, since data-plane SigV4 verification is well outside an import PR.

So the reporting side is handled; the enforcement side is not, and there is currently no issue for it. For a local emulator the risk is a false negative: a route that looks authenticated in the console and in Terraform state, but that any unsigned caller can invoke, which can mask a genuine authorization bug in the application under test.

## Reproduction

```bash
awsl() { aws --endpoint-url http://localhost:4566 --region us-east-1 "$@"; }
ACC=000000000000

echo 'exports.handler = async () => ({ statusCode: 200, body: "reached the integration" });' > fn.js
zip -q fn.zip fn.js
awsl lambda create-function --function-name iam-probe --runtime nodejs20.x \
  --role arn:aws:iam::$ACC:role/test --handler fn.handler --zip-file fileb://fn.zip

API_ID=$(awsl apigatewayv2 create-api --name iam-probe --protocol-type HTTP --query ApiId --output text)
INT_ID=$(awsl apigatewayv2 create-integration --api-id $API_ID \
  --integration-type AWS_PROXY --payload-format-version 2.0 \
  --integration-uri arn:aws:lambda:us-east-1:$ACC:function:iam-probe \
  --query IntegrationId --output text)

awsl apigatewayv2 create-route --api-id $API_ID --route-key 'GET /iam' \
  --target integrations/$INT_ID --authorization-type AWS_IAM
awsl apigatewayv2 create-stage --api-id $API_ID --stage-name test --auto-deploy

# GetRoutes reports AWS_IAM
awsl apigatewayv2 get-routes --api-id $API_ID --query 'Items[].AuthorizationType'

# ... but an entirely unsigned request still reaches the integration
curl -s "http://localhost:4566/restapis/$API_ID/test/_user_request_/iam"
# expected: 403
# actual:   reached the integration
```

## Environment

- Java SDK version (if applicable): n/a — AWS CLI v2 and curl

omatheusmesmo:
## Summary

A method (REST) or route (HTTP API) whose `authorizationType` is `AWS_IAM` was recorded and
reported back by `GetMethod`/`GetRoutes`, but nothing enforced it: v1 dispatch only ever invoked a
CUSTOM authorizer, and `dispatchV2` branched on `JWT` and `CUSTOM` alone. An entirely unsigned
request reached the integration exactly as if the route were `NONE`.

`ExecuteApiSigV4Authorizer` verifies the caller's signature by rebuilding the canonical request
from the request as it arrived — method, raw path, query string, the headers named in
`SignedHeaders`, and the SHA-256 of the body. Both placements real API Gateway accepts are
handled: an `Authorization` header and a presigned query string. The credential must be scoped to
`execute-api`, `X-Amz-Date` must fall inside a five-minute window, and an access key the emulator
never issued fails closed rather than being accepted as its own signing secret — the self-signing
bypass already fixed in the ElastiCache and RDS validators.

The verified caller now reaches the integration instead of being discarded:
`requestContext.identity.{accessKey, accountId, caller, user, userArn}` on a REST proxy event, and
`requestContext.authorizer.iam` on an HTTP API 2.0 event.

The virtual-host filter rewrites the request URI before dispatch sees it, so the path the client
actually signed is recorded on the request-scoped route context and used for verification. Without
that, every correctly signed virtual-host request would fail.

Two hardening details found while writing this:

- A literal `x-amz-content-sha256` is not taken on trust — the body is hashed instead, so
  re-presenting the original body's hash alongside a replaced body cannot restore the match. The
  `UNSIGNED-PAYLOAD`/`STREAMING-*` sentinels are honoured only when the header is itself in
  `SignedHeaders`. `ExecuteApiSigV4AuthorizerTest#declaredContentSha256CannotStandInForATamperedBody`
  is red without the fix.
- The presigned `X-Amz-Credential` is no longer double-decoded; JAX-RS has already decoded it.
  Same class of bug as the `PreSignedUrlFilter.java:109` finding in
  `issues/sigv4-review-remaining-findings.md`.

The OpenAPI import warning saying an `awsSigv4` scheme imports as `AWS_IAM` "which this emulator
records but does not enforce" is dropped — keeping it would only make `failOnWarnings` reject a
document Floci now honours.

**Two deliberate deviations, both documented in `docs/services/api-gateway.md`:**

1. Floci authenticates the caller but does not authorize the call. `execute-api:Invoke` is not
   evaluated against the caller's identity policies or a resource policy, since Floci's IAM policy
   evaluation is opt-in and off by default; requiring it would make the common case unusable.
2. The credential scope's region is not pinned to the API's region, because Floci resolves a
   request's region *from* that scope — pinning it would be circular.

There is no configuration flag: enforcement is unconditional. The blast radius is exactly the
routes that declared `AWS_IAM`, which have no guarantee today. Happy to add an escape hatch if
maintainers prefer one.

Closes #2810

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

**Incorrect behavior:** the reproduction in #2810 — an `AWS_IAM` route created via
`CreateRoute`, then `curl` with no credentials at all — returned `200` from the integration.
`GetRoutes` reported `AWS_IAM` throughout, so the route looked authenticated in the console and in
Terraform state while any unsigned caller could invoke it. That is a false negative that masks a
genuine authorization bug in the application under test. It now returns `403` before the
integration runs.

Error shapes are matched per API type, since the two differ on real AWS: REST returns a specific
message and `x-amzn-ErrorType` per failure class (`MissingAuthenticationTokenException`,
`IncompleteSignatureException`, `UnrecognizedClientException`, `InvalidSignatureException`), while
HTTP APIs collapse every IAM failure into `403 {"message":"Forbidden"}` — consistent with the
existing JWT and REQUEST authorizer rejections in `dispatchV2`. The table is in
`docs/services/api-gateway.md`.

Signatures are produced in tests by an independent signer (`testutil/ExecuteApiRequestSigner`)
rather than by the code under test, so the canonical-request construction is checked against a
second implementation rather than against itself.

## Checklist

- [ ] `./mvnw test` passes locally — see note below

**On the first box:** the suites this touches pass — 204/204 across the API Gateway v1/v2 suites,
including the four new ones. I could not tick the box honestly because the full run on my machine
has ~41 pre-existing errors in the Docker-backed Lambda suites (`ApiGatewayRequestAuthorizerTest`,
`HttpApiRequestAuthorizerTest`, `ApiGatewayV2BinaryPayloadIntegrationTest` and others). I verified
those are environmental, not caused by this branch: stashing the change and re-running two of them
at clean `HEAD` produced byte-identical counts (`HttpApiJwtAuthorizerScopesTest` 5/3/1,
`ApiGatewayAwsLambdaIntegrationTest` 10/0/2). CI should be the judge.

New tests, all built on MOCK and HTTP_PROXY integrations so they need no Lambda runtime:

| Test | Covers |
|---|---|
| `ApiGatewayIamAuthorizationTest` | REST end-to-end, both the `_user_request_` and `/execute-api/` entry points, and the per-failure error shapes |
| `HttpApiIamAuthorizationTest` | HTTP API end-to-end; asserts the backend hit count, which is what separates "rejected with 403" from "rejected after the integration already ran" |
| `ExecuteApiSigV4AuthorizerTest` | Presigned form, expiry, cross-service credential, unregistered key, body and query-string coverage, the virtual-host path rewrite |
| `ProxyEventIamIdentityTest` | The `requestContext.identity` and `requestContext.authorizer.iam` wire shapes |

### DIFF
--- a/docs/services/api-gateway.md
+++ b/docs/services/api-gateway.md
@@ -105,6 +105,53 @@ immediately.
 > disabled, or deleted key is still executed — it simply arrives with a null `identity.apiKey` rather
 > than being rejected with `403`.
 
+### IAM Authorization (`AWS_IAM`) {#iam-authorization}
+
+A method (v1) or route (v2) whose `authorizationType` is `AWS_IAM` requires a SigV4-signed caller.
+Before the integration runs, the signature is verified against the request as it arrived (method,
+path, query string, the headers named in `SignedHeaders`, and the SHA-256 of the body). Both
+placements AWS accepts are honoured: an `Authorization` header and a presigned query string
+(`X-Amz-Algorithm=AWS4-HMAC-SHA256`). The credential must be scoped to the `execute-api` service.
+
+Temporary credentials must also present the session token they were issued with, as
+`X-Amz-Security-Token` (a header, or a query parameter on a presigned request). The token is
+compared against the value recorded when STS minted the credential, so a missing or fabricated one
+is rejected: the secret alone is not the whole credential. Whether the token is covered by
+`SignedHeaders` is not required either way, since SigV4 leaves that service-specific.
+
+Rejections are `403`, and never reach the integration:
+
+| Condition | REST (v1) `message` / `x-amzn-ErrorType` | HTTP API (v2) |
+|---|---|---|
+| No SigV4 credentials at all | `Missing Authentication Token` / `MissingAuthenticationTokenException` | `Forbidden` |
+| Credentials present but incomplete, or scoped to another service | `Incomplete Signature` / `IncompleteSignatureException` | `Forbidden` |
+| Access key the emulator never issued | `The security token included in the request is invalid.` / `UnrecognizedClientException` | `Forbidden` |
+| Temporary credential presenting no `X-Amz-Security-Token`, or one that is not the token it was issued | `The security token included in the request is invalid.` / `UnrecognizedClientException` | `Forbidden` |
+| `X-Amz-Date` outside the 5-minute window, or a presigned URL past `X-Amz-Expires` | `Signature expired` / `InvalidSignatureException` | `Forbidden` |
+| Signature mismatch | `The request signature we calculated does not match…` / `InvalidSignatureException` | `Forbidden` |
+
+A verified caller reaches the integration as `requestContext.identity.{accessKey, accountId, caller,
+user, userArn}` on a REST proxy event, and as `requestContext.authorizer.iam` on an HTTP API 2.0
+event.
+
+Sign with any access key the emulator has issued (`CreateAccessKey`, or the temporary credentials
+from `AssumeRole`), or with the well-known local-dev `test`/`test` pair that Floci accepts across
+S3, RDS, and ElastiCache. No other unregistered key is accepted: an unknown key cannot sign for
+itself. Any AWS SDK sends the session token automatically, so temporary credentials need no special
+handling; a hand-rolled signer must include it.
+
+> [!NOTE]
+> A session minted before Floci recorded session tokens (one restored from a persisted store
+> written by an earlier version) has no issued token to compare against. Such a credential must
+> still present a token, but its value cannot be checked. Re-minting it with `AssumeRole` restores
+> the full check.
+
+> [!NOTE]
+> Floci authenticates the caller but does not authorize the call: it does not evaluate
+> `execute-api:Invoke` against the caller's IAM policies or a resource policy, so any validly
+> signed, known caller is let through. `principalOrgId` and `cognitoIdentity` are always null -
+> Organizations membership and identity-pool federation are not modelled.
+
 ### Not Implemented
 
 These management-plane operations have no handler in v1. Calls will return `404` or an error:
@@ -225,6 +272,10 @@ curl http://{apiId}.execute-api.localhost.floci.io:4566/{path}
 
 APIs created or updated with `disableExecuteApiEndpoint` reject requests to
 this default hostname with `404 Not Found`, matching AWS HTTP API behavior.
+
+Routes carrying `authorizationType: AWS_IAM`: including those an OpenAPI import resolves from an
+`awsSigv4` security scheme: require a signed caller; see
+[IAM Authorization](#iam-authorization).
 
 ### Supported Operations
 
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteApiHostFilter.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteApiHostFilter.java
@@ -187,6 +187,9 @@ public class ApiGatewayExecuteApiHostFilter implements ContainerRequestFilter {
                 .buildFromEncoded();
         LOG.debugv("Execute API host routing: {0}{1} -> {2}", host, originalPath, newUri.getRawPath());
         routeContext.routeToHttpApi(region);
+        // AWS_IAM dispatch rebuilds the caller's canonical request, which covers the path they
+        // signed: the virtual-host path, not the /execute-api/... form this rewrite produces.
+        routeContext.recordSignedRequestPath(originalPath);
         requestContext.setRequestUri(newUri);
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -97,6 +97,7 @@ public class ApiGatewayExecuteController {
     private final ApiGatewayExecuteRouteContext routeContext;
     private final JwtSignatureVerifier jwtSignatureVerifier;
     private final RequestContext requestContext;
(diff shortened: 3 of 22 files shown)


pgermosen (commented):
@omatheusmesmo thorough work on the core signer, and the size here is justified rather than scope creep once you break it down: the 567-line verifier is inherently that size for a from-scratch SigV4 implementation, the wiring touches both REST v1 and HTTP API v2 dispatch because the original bug (#2810) existed identically on both, and the eight one-line test diffs are just the mechanical fallout of the controller's constructor gaining one new dependency.

One thing does need attention before this lands, and it's already been flagged twice by review without a reply yet. greptile's first pass found two P1s: a signature that omits `host` from `SignedHeaders` wasn't bound to the API it was verified against, and STS temporary credentials were accepted without validating `X-Amz-Security-Token`. The host-binding one is genuinely fixed — I read the code and confirmed the current authorizer rejects any signature where `SignedHeaders` doesn't cover `host`, with a dedicated replay test. The session-token one isn't: the second review round re-flagged it, and I confirmed independently by searching the full authorizer file for any reference to `SecurityToken` or `sessionToken` — there isn't one, anywhere. An STS access key can currently authenticate with no session token, or a fabricated one, against an `AWS_IAM` route.

I checked this against the actual AWS references rather than just take the finding on faith. The IAM guide is direct about it: using temporary credentials, "you must include the session token, which lets AWS verify that the temporary security credentials are valid" — it's not optional, it's how the credential's legitimacy gets confirmed. The SigV4 signing reference adds a useful nuance for whichever way you choose to fix this: whether the token needs to be folded into the canonical request (covered by `SignedHeaders`) is service-specific, not universal — some services require it in the canonical request, others only require it appended after the signature is calculated. So the fix doesn't have to fold it into the signature math; what's missing is checking the token's presence and validity at all.

omatheusmesmo:
Hi @pgermosen , could you kindly review it again?

pgermosen (commented):
Good news, the session-token gap from last round is genuinely closed. Traced checkSessionToken end to end, it correctly scopes to temporary credentials only, checks both the header and query-param placements, rejects a missing token outright, and constant-time-compares against what Floci actually issued when it has one recorded. Where nothing was recorded, legacy sessions or the Lambda-execution-role/federation paths, it falls back to presence-only rather than rejecting, which reads as a deliberate carve-out rather than an oversight, it doesn't regress those below what existed before and fully closes the gap for AssumeRole/GetSessionToken/etc sessions. Good tests pinning both the missing and fabricated cases. Em-dashes are clean now too.

One thing to flag though: this now shows as conflicting against main. Looks like it's because this PR and #2912, which merged a bit ago, both independently added a sessionToken field to SessionCredential and threaded it through the same five files, IamService, SessionCredential, StsQueryHandler, LambdaExecutionRoleCredentials, and SigninService. Not a correctness problem, #2912's version is already in main, but this branch will need a rebase, and I'd drop this PR's own copy of that plumbing in favor of what's already merged, keeping just what's actually new here, findSessionToken, isTemporaryAccessKey, and the authorizer wiring itself.

omatheusmesmo:
Hi @pgermosen , thanks for reviewing. Conflicts solved.

pgermosen (approved):
Approving. The rebase did exactly what I asked - SessionCredential/StsQueryHandler/LambdaExecutionRoleCredentials/SigninService are gone from this diff, relying on #2912's already-merged sessionToken field instead of re-adding it, and IamService's remaining changes are just findSessionToken plus widening isTemporaryAccessKey to public. Confirmed the authorizer wires to both correctly. Combined with the host-binding and session-token validation I already traced end to end in earlier rounds, this closes the #2810 gap properly. Good work getting this across the line.

Issue #2433, opened 2026-08-20
[FEAT] cognito: email MFA configuration (SetUserPoolMfaConfig / AdminSetUserMFAPreference)


tarinaihitori:
## Service

Cognito

## API Action / Feature

MFA configuration: `SetUserPoolMfaConfig`, `GetUserPoolMfaConfig`, `AdminSetUserMFAPreference`,
`SetUserMFAPreference`, and the resulting `UserMFASettingList` / `PreferredMfaSetting` on
`AdminGetUser` / `GetUser`. The `EMAIL_OTP` sign-in challenge is excluded — see **Scope**.

## AWS Documentation

- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_SetUserPoolMfaConfig.html
- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminSetUserMFAPreference.html
- https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-mfa.html

## Why is this needed?

`SetUserPoolMfaConfig` has no matches in the repository and `services/cognito` has no MFA model, so
the MFA actions are unhandled. `AdminSetUserMFAPreference` answers the same way, and `AdminGetUser`
never returns `UserMFASettingList` or `PreferredMfaSetting`. Two things are blocked.

**Terraform.** `aws_cognito_user_pool` exposes `mfa_configuration` and an `email_mfa_configuration`
block (`message`, `subject`), which the provider applies through `SetUserPoolMfaConfig`. A pool with
MFA enabled cannot be applied against Floci at all — it fails on the MFA call rather than degrading
to an MFA-less pool.

**Applications that own an MFA policy.** Since AWS added native email MFA (`EMAIL_OTP`) in September
2024, a common shape is for MFA to be a policy the application owns rather than a per-user opt-in: a
tenant enables MFA, the application propagates it to each of its users with
`AdminSetUserMFAPreference`, then re-reads `UserMFASettingList` at sign-in to confirm the policy is
in force. Because the call raises rather than silently having no effect, such an application must
branch on "am I talking to the emulator?" in production code and disable both the propagation and
the sign-in check — removing exactly the check that guarantees the policy was applied.

The configuration plane alone removes that branch. The sign-in challenge is not needed for it.

## Minimal reproduction

Floci 1.7.0, `floci/floci:1.7.0`:

```
[...]

PR #2896, opened 2026-09-01
feat(cognito): implement user MFA preferences

AbarnaaSree:
## Summary

Implements the per-user MFA preference portion of #2433.

### Changes

* Added `AdminSetUserMFAPreference`
* Added `SetUserMFAPreference`
* Added persisted `EmailMfaSettings` to `CognitoUser`

  * `Enabled`
  * `PreferredMfa`
* Added validation to prevent `PreferredMfa` from being enabled when email MFA is disabled
* Added access-token validation for `SetUserMFAPreference`
* Updated the existing `SetUserPoolMfaConfig` regression test to expect `404 ResourceNotFoundException` for a non-existent user pool

### Testing

* Integration tests pass: **13 tests, 0 failures, 0 errors, 0 skipped**
* `git diff --cached --check` passes
* Working tree is clean

### Scope

This PR focuses on the **per-user MFA preference** portion of #2433.

`SetUserPoolMfaConfig` / `GetUserPoolMfaConfig` are being handled separately in Part 1.

The `EMAIL_OTP` authentication challenge remains out of scope as described in #2433.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoJsonHandler.java
@@ -103,6 +103,8 @@ public class CognitoJsonHandler {
             case "ListUserPoolClientSecrets" -> handleListUserPoolClientSecrets(request);
             case "AddUserPoolClientSecret" -> handleAddUserPoolClientSecret(request);
             case "DeleteUserPoolClientSecret" -> handleDeleteUserPoolClientSecret(request);
+            case "AdminSetUserMFAPreference" -> handleAdminSetUserMFAPreference(request);
+            case "SetUserMFAPreference" -> handleSetUserMFAPreference(request);
             default -> Response.status(400)
                     .entity(new AwsErrorResponse("UnsupportedOperation", "Operation " + action + " is not supported."))
                     .build();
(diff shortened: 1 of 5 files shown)


pgermosen (commented):
@AbarnaaSree thanks for taking this on, and welcome, this is your first PR here. A few things to work through before this is ready.

The conflict and the size of the diff both trace back to the same thing: your description says SetUserPoolMfaConfig/GetUserPoolMfaConfig are handled separately in "Part 1," but the diff and commit list here include a full setUserPoolMfaConfig implementation anyway. That's PR #2831, a different contributor's still-open, unmerged PR. It looks like this branch was built on top of #2831 rather than main, so it's carrying those commits as its own, which is both why it conflicts and why the diff is bigger and more confusing than your own description promises. Worth rebasing onto main (once #2831 lands, or cleanly against it in the meantime) so this PR's diff only contains the user-preference work it's actually about.

Once that's sorted, there's a real bug in updateEmailMfaPreference worth fixing: it mutates the live EmailMfaSettings object before the validate-and-throw check runs. I traced this through the actual storage layer rather than taking it on faith, InMemoryStorage.get() returns the literal object reference from the backing map, not a copy, so when a request combines something like Enabled=false with a previously-set PreferredMfa=true and gets rejected, the mutation already happened in place before the exception fires. A later GetUser sees the partially-applied state from a request that was supposed to fail atomically, not the state from before it. Validating before mutating, or mutating a copy and only writing back after validation passes, would fix it.

On the JWT point greptile raised: the code claim is accurate, setUserMFAPreference decodes the token's identity claims with no signature verification, but that's not something this PR introduces. Every existing self-service access-token endpoint in this file, getUser, updateUserAttributes, deleteUserAttributes, does the identical unverified decode already. So this is extending an existing pattern rather than opening a new hole, worth being aware of rather than something to fix in this PR alone. One thing worth knowing, though: AWS's own docs for this specific operation say Cognito doesn't evaluate IAM policies for it at all, authorization is purely the access token's validity, which makes a forgeable token a more direct bypass here than for operations IAM could also gate.

Last thing, not confirmed against a live measurement either way, just an internal inconsistency worth checking: your own setUserPoolMfaConfig method explicitly clears the factor configuration when MFA is turned off, but updateEmailMfaPreference doesn't do the equivalent. Disabling email MFA while it was previously the preferred method throws instead of succeeding, since PreferredMfa isn't cleared along with it. Might be worth applying the same clearing logic there.

AbarnaaSree:
Thanks for the review! I’ve rebased the work onto `main` and cleaned up the PR so it only contains the user MFA preference changes. I also fixed the validation/mutation issue, ensured disabling email MFA clears `PreferredMfa`, and added regression tests. `CognitoServiceTest` passes with 172/172 tests.

pgermosen (commented):
Good progress, both things from last round are genuinely fixed. updateEmailMfaPreference now validates before it mutates anything, and there's a regression test proving the prior state survives a rejected update. Disabling email MFA now clears PreferredMfa instead of throwing too, traced both branches and that's correct, including the case where a request tries to enable PreferredMfa and disable the method in the same call, which still correctly throws.

The thing actually blocking the build right now is PersistedModelReflectionTest failing on EmailMfaSettings, it's missing @RegisterForReflection. It's a field on the persisted CognitoUser now, so it's reachable from storage, and every sibling model class in this service carries that annotation, CognitoUser itself does. Adding it should clear the build. While you're in there, @JsonIgnoreProperties(ignoreUnknown = true) would match the sibling classes too, though that one isn't what's failing.

One thing carried over from last round, not new: setUserMFAPreference still authorizes off an access token with no signature check, same as every other self-service endpoint in this file already does, so not something to fix here, just flagging again for completeness.

AbarnaaSree:
The reflection fix is pushed in `970c33e`. All 19 CI checks are now passing, including `PersistedModelReflectionTest` and `CognitoServiceTest` (172/172).

pgermosen (approved):
That's the fix, confirmed. Both annotations added, matching the sibling classes, and CI is green now. Nothing else outstanding.

PR #2897, opened 2026-09-01
fix(transfer): allow DeleteServer regardless of server state

### LINKED ISSUE #2802: transfer: DeleteServer wrongly requires OFFLINE state; CreateServer defaults to ONLINE
`TransferService.deleteServer` rejects deletion unless the server is `OFFLINE`:

```java
public synchronized void deleteServer(String serverId) {
    Server server = getServer(serverId);
    if (!"OFFLINE".equals(server.getState())) {
        throw new AwsException("ConflictException",
                "Server must be in OFFLINE state to be deleted.", 409);
    }
    ...
```

This isn't real AWS behavior. Checked two primary sources:

1. **AWS's own `DeleteServer` API reference** lists only `AccessDeniedException`, `InternalServiceError`, `InvalidRequestException`, `ResourceNotFoundException`, `ServiceUnavailableException` — no `ConflictException`, and no documented state precondition.
2. **`terraform-provider-aws`'s `resourceServerDelete`** (`internal/service/transfer/server.go`) calls `DeleteServer` directly with no `StopServer` call beforehand (it only stops/deletes users first when `force_destroy` + `SERVICE_MANAGED`). A mature, official provider doing this confirms real AWS doesn't need the server offline first — otherwise every `terraform destroy` of an `aws_transfer_server` with a server left running would fail on real AWS too.

Also, `createServer` sets the initial state to `ONLINE` (`TransferService.java:61`) — real Transfer Family servers are created `OFFLINE` until explicitly started via `StartServer`, so this is a second, related conformance gap.

Confirmed still present on `main`.

Blocks the Terraform/OpenTofu compatibility suites from including `aws_transfer_server` (removed from both in #1834 to unblock the suite — `terraform destroy` failed on it every run; should be re-added once this lands).

7487:
## Summary

Removes the OFFLINE-only guard from `TransferService.deleteServer`, returns `Domain` from DescribeServer/ListServers, and adds `aws_transfer_server` to the Terraform and OpenTofu compat suites. Fixes #2802.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

**Incorrect behavior:** `DeleteServer` returned a 409 `ConflictException` ("Server must be in OFFLINE state to be deleted.") unless the server had been stopped first. Real AWS has no such precondition: the [DeleteServer API reference](https://docs.aws.amazon.com/transfer/latest/userguide/API_DeleteServer.html) lists no `ConflictException`, and terraform-provider-aws's `resourceServerDelete` calls `DeleteServer` directly without a prior `StopServer`, so every `terraform destroy` of a running `aws_transfer_server` failed against Floci (this is what forced the resource out of the compat suites in #1834).

**On the issue's second point (CreateServer defaulting to ONLINE):** left as is, deliberately. ONLINE is the correct terminal state, not a conformance gap:

- AWS's own docs on creating a server: "It can take a couple of minutes before the status for your new server changes to **Online**" ([user guide](https://docs.aws.amazon.com/transfer/latest/userguide/create-server-sftp.html)). A new server auto-starts; `StartServer` is only for servers previously taken OFFLINE via `StopServer`.
- terraform-provider-aws's `waitServerCreated` waits with `Pending: [STARTING]`, `Target: [ONLINE]` right after `CreateServer`. OFFLINE is neither pending nor target, so a state waiter hitting OFFLINE fails immediately. An OFFLINE default would break every `terraform apply` of `aws_transfer_server`, trading the destroy-time failure for an apply-time one.

Since Floci doesn't model the transient STARTING state, jumping straight to the ONLINE steady state is what SDK/terraform waiters expect.

**Domain (found while putting the resource through terraform):** apply and destroy passed, but the re-plan proposed `+ domain = "S3" # forces replacement`. Floci dropped `Domain` on CreateServer and never echoed it, while the provider reads it into a ForceNew attribute defaulting to S3. CreateServer now stores it (default S3, as AWS does) and DescribeServer/ListServers return it.

**Compat suites:** `aws_transfer_server` is back in `compat-terraform/main.tf` and `compat-opentofu/main.tf`, never stopped, so `destroy` deletes an ONLINE server. Verified locally with terraform 1.14.7 + provider 6.63.0: apply, re-plan with `-detailed-exitcode` (No changes), destroy.

## Checklist

- [x] `./mvnw test` passes locally (`-Dtest=TransferServerLifecycleIntegrationTest`: 6/6)
- [x] New or updated integration test added (`TransferServerLifecycleIntegrationTest`: delete while ONLINE, delete while OFFLINE, delete cascades to users, delete missing server 404, create comes up ONLINE, Domain round trip)


pgermosen (commented):
@7487 welcome, and this is exactly right. I pulled the actual DeleteServer reference myself to check: its error list is AccessDeniedException, InternalServiceError, InvalidRequestException, ResourceNotFoundException, ServiceUnavailableException, nothing about state at all. The OFFLINE guard was inventing a 409 real AWS never returns. Leaving CreateServer's ONLINE default alone is the right call too, for the reason you gave — AWS auto-transitions a new server to ONLINE, and terraform's post-create waiter would fail immediately on OFFLINE.

One thing worth doing while you're here: could you add `aws_transfer_server` back to the Terraform and OpenTofu compat suites? I went and checked why it's missing rather than assume, and it's not a coincidence — it was in `compatibility-tests/compat-terraform/main.tf` and `compat-opentofu/main.tf` when #1834 added Transfer coverage, but got pulled before that PR merged, with this note in the commit history:

> Removed aws_transfer_server: Floci's DeleteServer wrongly requires OFFLINE state, which isn't real AWS behavior (confirmed against the API reference and terraform-provider-aws's own delete logic, which never calls StopServer first). Tracked as #2802.

That's the exact bug this PR fixes, cited by its own tracking issue. It never actually landed in a mergeable commit (added and removed within that PR's own iteration), so there's no old block to restore verbatim — but adding a fresh `aws_transfer_server` resource in the same style as the file's other resources, in both suites, would put this fix through a real `terraform apply`/`destroy` cycle rather than only the HTTP-level test already here. You'd already flagged this as the natural next step in the PR description; this is just asking you to take it.

7487:
Done, in three follow-up commits:

- c833b82 `fix(transfer)`: running the resource through terraform 1.14.7 + provider 6.63.0 locally, apply and destroy both passed, but the re-plan wanted to replace the server: `+ domain = "S3" # forces replacement`. Floci never stored or echoed `Domain`, and the provider reads it back into a ForceNew attribute that defaults to S3, so the whole-config re-plan test would have failed the moment the resource was in the fixture. CreateServer now keeps `Domain` (default S3, as AWS does) and DescribeServer/ListServers return it. Covered by `describeServerReportsDomain` in the same integration test.
- 9c5918f `test(compat)`: `aws_transfer_server` in both `compat-terraform/main.tf` and `compat-opentofu/main.tf`, never stopped so destroy hits DeleteServer on an ONLINE server, plus a `transfer list-servers` spot check in each bats file. The existing whole-config re-plan test now guards the DescribeServer round trip as well.
- d08dfa3 `docs(transfer)`: the action table still said DeleteServer requires OFFLINE.

Local run against the rebuilt jar: apply -> re-plan `No changes` (exit 0) -> destroy, all clean. `TransferServerLifecycleIntegrationTest` 6/6.

pgermosen (approved):
Nice work chasing this all the way through terraform apply/destroy — the Domain round-trip and getting aws_transfer_server back into the compat suites was exactly the right next step. Checked DeleteServer's error list against the real service model myself and you're right, no ConflictException in sight. This is good to go.

PR #2898, opened 2026-09-01
ci(compat): fail the build when compat-opentofu drifts from compat-terraform

### LINKED ISSUE #2263: [BUG] compat-opentofu has drifted out of sync with compat-terraform despite the in-file note to keep them aligned
Reported while adding IAM tag coverage to `compat-terraform`.

## Problem

`compatibility-tests/compat-opentofu/main.tf` opens with:

```hcl
# NOTE: Keep resource definitions in sync with ../compat-terraform/main.tf
```

That note is no longer accurate. The two configs have diverged: **37 resources vs 27**, and **29 bats cases vs 17**.

Resources present in `compat-terraform/main.tf` but missing from `compat-opentofu/main.tf`:

```
aws_appautoscaling_policy.ecs_alb_requests
aws_appautoscaling_policy.ecs_cpu
aws_appautoscaling_target.ecs_service
aws_cloudwatch_metric_alarm.cpu
aws_db_instance.app
aws_iam_role.managed_policy_attach
aws_iam_role_policy_attachment.ec2_read_only
aws_iam_role_policy_attachment.emr_service_role
```

The drift is not one PR's fault — it accumulated because nothing enforces the note. Every PR that adds a resource to one config and not the other widens the gap silently, and the note gives the next contributor a false expectation that the two are equivalent.

## Why it matters

The point of a separate OpenTofu suite is to catch places where OpenTofu and Terraform diverge against Floci. That value scales with how much of the config the two share. Right now OpenTofu exercises roughly two-thirds of what Terraform does, so anything in the gap — Application Auto Scaling, the IAM managed-policy path, RDS — has no OpenTofu coverage at all, while the note implies it does.

## Options

1. **Enforce it.** Extract the shared resource definitions into one directory that both suites consume, so the two cannot drift. Strongest fix, largest change.
2. **Check it.** Add a CI step comparing the resource address lists and failing on divergence, with an explicit allowlist for anything intentionally one-sided.
3. **Sync once and soften the note** to describe what is actually shared and what is deliberately not.

Option 2 is probably the best value: it costs little, fails loudly at the moment the gap is introduced, and leaves room for genuinely provider-specific fixtures.

## Notes

- Unrelated but seen in the same run: `native / compat-opentofu` can fail on `tofu init` when `registry.opentofu.org` refuses to serve provider checksums (`failed to retrieve authentication checksums for provider: the request failed after 2 attempts`). Worth a retry/lock-file strategy if it recurs.

pgermosen:
## Summary

`compat-opentofu/main.tf` opens with a note asking contributors to keep it in sync with `compat-terraform/main.tf`, but nothing enforced it. The two had drifted to 53 vs 64 resources — Application Auto Scaling, RDS, the CloudWatch alarm, the IAM managed-policy-attach path, IAM tag coverage, and the GuardDuty org feature are all exercised by compat-terraform with no OpenTofu equivalent, so that gap has zero cross-provider coverage while the note implies otherwise.

This adds a CI job that parses the `resource "type" "name"` addresses out of both `main.tf` files and fails if either config declares one the other doesn't. The 11 resources currently only in compat-terraform are recorded in a new `.github/ci/compat-opentofu-allowlist-tf-only.txt` as accepted gaps rather than ported here, so today's build stays green; the OpenTofu-only counterpart, `.github/ci/compat-opentofu-allowlist-otf-only.txt`, starts out empty. Any *new* one-sided resource fails immediately instead of widening the gap silently, and a resource that moves from one side to the other is caught too, since each direction only ever subtracts its own allowlist. The check also fails if an allowlist entry stops being one-sided, so accepted gaps get removed once they're closed rather than accumulating.

Closes #2263

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A — CI tooling only, no service behavior changed.

## Checklist

### DIFF
--- a/.github/ci/compat-opentofu-allowlist-otf-only.txt
+++ b/.github/ci/compat-opentofu-allowlist-otf-only.txt
@@ -1,0 +1,9 @@
+# Resource addresses allowed to exist ONLY in compat-opentofu/main.tf (not
+# in compat-terraform/main.tf), checked by compat-opentofu-drift-check.py.
+# One "type.name" per line; `#` starts a trailing comment. The reverse
+# direction -- addresses allowed only in compat-terraform -- lives in
+# compat-opentofu-allowlist-tf-only.txt; an address moving from one side to
+# the other must move between these files or CI fails it as unreviewed drift.
+#
+# Empty for now: every currently-accepted gap runs the other way (see
+# compat-opentofu-allowlist-tf-only.txt).
--- a/.github/ci/compat-opentofu-allowlist-tf-only.txt
+++ b/.github/ci/compat-opentofu-allowlist-tf-only.txt
@@ -1,0 +1,21 @@
+# Resource addresses allowed to exist ONLY in compat-terraform/main.tf (not
+# in compat-opentofu/main.tf), checked by compat-opentofu-drift-check.py.
+# One "type.name" per line; `#` starts a trailing comment. The reverse
+# direction -- addresses allowed only in compat-opentofu -- lives in
+# compat-opentofu-allowlist-otf-only.txt; an address moving from one side to
+# the other must move between these files or CI fails it as unreviewed drift.
+#
+# These are pre-existing gaps, not deliberate provider-specific fixtures --
+# see https://github.com/floci-io/floci/issues/2263. Remove an entry once its
+# resource (and matching bats coverage) is ported to compat-opentofu.
+aws_appautoscaling_policy.ecs_alb_requests                            # Application Auto Scaling not yet in compat-opentofu
+aws_appautoscaling_policy.ecs_cpu                                     # Application Auto Scaling not yet in compat-opentofu
+aws_appautoscaling_target.ecs_service                                 # Application Auto Scaling not yet in compat-opentofu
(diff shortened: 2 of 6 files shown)


hectorvent:
Thank you, this is a good shape for a drift gate: it records today's gaps instead of pretending they are not there, and it fails the moment a new one appears.

I did not just read it, I tried to break it. Adding a resource to one side, closing an allowlisted gap without removing its entry, and flipping a resource across sides under the same address all fail with a clear message. That last one is the case worth having, and the two directional allowlists catch it in both directions at once.

On Greptile's P1 about reversed drift, that does not hold here. Each direction loads only its own allowlist, so a terraform-only entry is never subtracted from the opentofu comparison, and its own summary withdraws the point.

One small thing: the description still mentions a single `compat-opentofu-allowlist.txt`, but you shipped two directional files. Worth updating the body before merge.

No blockers from my side.

pgermosen:
Thanks for digging in on this one. Updated the description to name both `compat-opentofu-allowlist-tf-only.txt` and `compat-opentofu-allowlist-otf-only.txt` instead of the stale single-file reference.

hectorvent (approved):

PR #2902, opened 2026-09-01
refactor(config): move max-request-size and ecr-base-uri into sub-mappings

cfranzen:
## Summary

floci.max-request-size and floci.ecr-base-uri sat at the config root but belong with related settings. Relocate them:

- max-request-size -> floci.protocols.max-request-size (ProtocolsConfig)
- ecr-base-uri      -> floci.services.lambda.ecr-base-uri (LambdaServiceConfig)

A new FlociConfigRelocationsInterceptor (SmallRye ConfigSourceInterceptor, registered via META-INF/services) keeps the old flat keys and their FLOCI_MAX_REQUEST_SIZE / FLOCI_ECR_BASE_URI env vars working: the legacy key is consulted when the new key is requested, and iterateNames() rewrites it so config-mapping validation does not flag it as unknown. Future relocations are one entry in its RELOCATIONS list.

ImageResolver, both application.yml files, the affected tests and the configuration docs are updated to the new keys.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ docker run -d --name floci \
 | `FLOCI_SERVICES_EKS_DEFAULT_IMAGE` | `rancher/k3s:latest` |
 | `FLOCI_SERVICES_MWAA_DEFAULT_POSTGRES_IMAGE` | `postgres:16-alpine` |
 | `FLOCI_SERVICES_ECR_REGISTRY_IMAGE` | `registry:2` |
-| `FLOCI_ECR_BASE_URI` | `public.ecr.aws` |
+| `FLOCI_SERVICES_LAMBDA_ECR_BASE_URI` | `public.ecr.aws` |
 
 ## Persistence and Storage Modes
 
@@ -788,7 +788,7 @@ All settings are overridable through environment variables with the `FLOCI_` pre
 | `FLOCI_HOSTNAME` | Unset | Hostname used in returned URLs when Floci runs inside Docker Compose |
 | `FLOCI_STORAGE_MODE` | `memory` | Storage mode: `memory`, `persistent`, `hybrid`, or `wal` |
 | `FLOCI_STORAGE_PERSISTENT_PATH` | `./data` | Directory used for persisted state |
-| `FLOCI_ECR_BASE_URI` | `public.ecr.aws` | ECR base URI used when pulling container images |
+| `FLOCI_SERVICES_LAMBDA_ECR_BASE_URI` | `public.ecr.aws` | ECR base URI used when pulling Lambda runtime images (legacy name `FLOCI_ECR_BASE_URI` still works) |
 | `FLOCI_SERVICES_S3_ENFORCE_AUTH` | `false` | Enforce S3 public/private read access and reject unknown signed S3 access keys |
 
 Full reference: [configuration docs](https://floci.io/floci/configuration/advanced/application-yml)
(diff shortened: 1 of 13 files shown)


pgermosen (commented):
@cfranzen the interceptor's trickiest claim, that expression expansion for something like ${floci.protocols.max-request-size} in application.yml still falls back to the legacy key, checks out. I traced the ordinal comparison in getValue() by hand against expressionExpansionSeesTheLegacyKeyWhenResolvingTheRelocatedKey and it holds up, and that's not a synthetic scenario either, the real application.yml has exactly that expression in production, so the test is exercising the literal shipped config. Copilot flagged this as untested on its first pass and reversed itself on the second, which matches what I found independently. Good work getting that one right, it's a genuinely easy thing to get backwards.

The rest of the relocation is applied consistently across the config interfaces, both application.yml files, ImageResolver, and the docs, and the other interceptor tests cover the cases that matter, new key wins at equal priority, default applies when neither is set, name iteration hides the legacy keys from validation.

One thing still open: greptile's em-dash finding hasn't been addressed yet, the em-dashes are still there throughout the new Javadoc and YAML comments in FlociConfigRelocationsInterceptor.java, EmulatorConfig.java, both application.yml files, and the environment-variables doc.

hectorvent:
Thank you, this is a tidy relocation and the shim is the right call. I checked the back compat end to end: with the legacy env vars set, Quarkus boots clean and both values resolve to the new keys.

Copilot's note on expression expansion looks stale, your test for it is there and passes. No blockers from my side.

hectorvent:
waiting for @pgermosen comments to be resolved.

cfranzen:
Hey @hectorvent and @pgermosen. Thanks for the review. Sorry, I forget to mention the reason, when I resolved the em-dash comment. For me that looked like a false-positive. Looking at the current code base, em-dashes are everywhere. So it is a real requierment to get rid of em-dashes?

pgermosen:
Good question, and fair pushback! It's actually a documented rule, not greptile inventing something: [AGENTS.md#L333](https://github.com/floci-io/floci/blob/main/AGENTS.md#L333) under "Documentation Style" says no em-dashes anywhere, use colons, commas, or periods instead. You're right that they're scattered through the existing codebase already, that's just older content that predates us actually enforcing it. Doesn't mean we should keep adding new ones though. Should be a quick find-and-replace in the files greptile pointed at, nothing structural. Let's just try to keep new stuff clean going forward and we can chip away at the old stuff separately sometime.

cfranzen:
Hey @pgermosen. Thanks for clarifying. I wasn't aware of that rule. And I am a bit surprised that Claude did not pick up that rule from the AGENTS.md. Nevertheless, I pushed another change to this PR and removed all em-dashes.

pgermosen (approved):
Thanks for the cleanup, most of it's there now, the Java files and the env-vars doc are clean. There are still three em-dashes left (two in application-yml.md, one in the real application.yml's ecr-base-uri comment), but since those files already carry plenty of pre-existing ones elsewhere, I don't think it's worth blocking this PR over three more. Let's get this merged and handle a proper sweep of those files separately.

Everything else here still holds from the first pass, the interceptor's ordinal-comparison logic checks out against the real application.yml expression, and the relocation is applied consistently across the config interfaces, both yml files, ImageResolver, and the docs.

PR #2907, opened 2026-09-01
fix(lambda): apply Architectures on UpdateFunctionCode

### LINKED ISSUE #2839: [BUG] Lambda UpdateFunctionCode silently discards Architectures
## Service

Lambda

## AWS API Action

`UpdateFunctionCode`

## Expected behavior

`UpdateFunctionCode` takes an `Architectures` parameter and real AWS applies it, switching the function's instruction set architecture along with the new package.

Measured against `lambda.ap-southeast-1.amazonaws.com` on a throwaway function created and deleted for the purpose:

```
$ aws lambda create-function --function-name probe --runtime nodejs22.x \
      --role <role> --handler index.handler --zip-file fileb://fn.zip \
      --architectures arm64 --query 'Architectures'
[ "arm64" ]

$ aws lambda update-function-code --function-name probe \
      --zip-file fileb://fn.zip --architectures x86_64 --query 'Architectures'
[ "x86_64" ]

$ aws lambda get-function-configuration --function-name probe --query 'Architectures'
[ "x86_64" ]
```

## Actual behavior

Floci accepts the parameter, returns 200, and silently discards it. The function keeps its original architecture, on the `UpdateFunctionCode` response and on every subsequent read:

```
--- create with Architectures=[arm64] ---
[ "arm64" ]
--- update-function-code --architectures x86_64 ---
[ "arm64" ]
--- read back ---
[ "arm64" ]
```

`LambdaService` reads `Architectures` in the create path (line ~365) and in the configuration-update path (line ~616), but the code-update path has no equivalent branch, so the member is dropped.

There is no error and no warning — the caller's only way to notice is to read the value back and compare.

## Reproduction

```bash

aws $E lambda create-function --function-name arch-probe --runtime nodejs22.x \
    --zip-file fileb://fn.zip --architectures arm64 --query 'Architectures' --output json
# [ "arm64" ]   <- correct

aws $E lambda update-function-code --function-name arch-probe \
    --zip-file fileb://fn.zip --architectures x86_64 --query 'Architectures' --output json
# [ "arm64" ]   <- expected [ "x86_64" ]

aws $E lambda get-function-configuration --function-name arch-probe \
    --query 'Architectures' --output json
# [ "arm64" ]   <- expected [ "x86_64" ]

aws $E lambda delete-function --function-name arch-probe
```

Both the Floci and the live-AWS blocks above were executed, not reconstructed.

## Environment

- Floci version / image tag: build of upstream `main` at `62dbd9d` ("fix(apigateway): report RestApi as available (#2816)")
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker, with `/var/run/docker.sock` mounted

7487:
## Summary

`UpdateFunctionCode` accepted the `Architectures` parameter, returned 200, and silently dropped it — the function kept its original architecture on the response and on every subsequent read. `LambdaService.updateFunctionCode` never read the member; the create and configuration-update paths both did.

This applies `Architectures` in the code-update path with the same guard the configuration-update path uses, so the new architecture shows up on the `UpdateFunctionCode` response and on later `GetFunctionConfiguration` reads, matching the real-AWS transcript in the issue.

Closes #2839

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `update-function-code --architectures x86_64` on an `arm64` function returned `["arm64"]` and left the stored function unchanged. Real AWS (verified in the issue against `lambda.ap-southeast-1.amazonaws.com`) returns and persists the new architecture.

## Checklist

- [x] `./mvnw test` passes locally (`LambdaServiceTest`: 59 tests, 0 failures)
- [x] New or updated integration test added — two unit tests in `LambdaServiceTest`: architecture applied on code update, and preserved when the parameter is omitted

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -533,6 +533,15 @@ public class LambdaService implements ResourceProvider {
             }
         }
 
+        if (request.containsKey("Architectures")) {
+            @SuppressWarnings("unchecked")
+            List<String> archs = request.get("Architectures") instanceof List
+                    ? (List<String>) request.get("Architectures") : null;
+            if (archs != null && !archs.isEmpty()) {
+                fn.setArchitectures(new ArrayList<>(archs));
+            }
+        }
+
         fn.setLastModified(System.currentTimeMillis());
         fn.setRevisionId(UUID.randomUUID().toString());
 
--- a/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaServiceTest.java
@@ -580,6 +580,28 @@ class LambdaServiceTest {
     }
 
     @Test
+    void updateFunctionCodeAppliesArchitectures() {
+        Map<String, Object> req = baseRequest("arch-fn");
+        req.put("Architectures", List.of("arm64"));
+        service.createFunction(REGION, req);
+
+        LambdaFunction updated = service.updateFunctionCode(REGION, "arch-fn",
+                Map.of("Architectures", List.of("x86_64")));
+        assertEquals(List.of("x86_64"), updated.getArchitectures());
+        assertEquals(List.of("x86_64"), service.getFunction(REGION, "arch-fn").getArchitectures());
+    }
+
+    @Test
+    void updateFunctionCodeWithoutArchitecturesKeepsExisting() {
+        Map<String, Object> req = baseRequest("arch-keep-fn");
+        req.put("Architectures", List.of("arm64"));
+        service.createFunction(REGION, req);
+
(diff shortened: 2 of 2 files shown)


hectorvent (approved):
Thank you, this is a tidy fix and the sourcing is what makes it easy to accept.

The model backs the premise: `Architectures` really is a member of `UpdateFunctionCodeRequest`, and the operation returns a `FunctionConfiguration`, so applying it fixes both the immediate response and later reads. Your two tests pin exactly those.

The part I want to credit is that the new block is byte for byte the same guard the configuration update path already uses, rather than a fresh variation of it. For a nine line fix that is the right instinct, and it means there is no new shape to reason about.

I also checked the transcript in #2839. Measuring it against a real region on a throwaway function is the strongest evidence there is for behaviour the model does not spell out, and it is appreciated.

59 tests green on your branch. No blockers from my side.

PR #2912, opened 2026-09-01
fix(eks): validate IAM authentication tokens

simonepri:
## Problem

The EKS token-authentication webhook accepted any bearer token with the `k8s-aws-v1.` prefix. It did not validate the embedded SigV4 request, expiry, or target cluster.

## Solution

The webhook now validates the presigned STS `GetCallerIdentity` request against known IAM credentials, requires its 60-second expiry and signed `x-k8s-aws-id` cluster binding, and gives each cluster a distinct webhook URL. Coverage includes tampered, expired, future-dated, unknown-credential, and cross-cluster tokens.

### DIFF
--- a/docs/services/eks.md
+++ b/docs/services/eks.md
@@ -46,7 +46,11 @@ aws eks update-kubeconfig --name my-cluster
 kubectl get nodes
 ```
 
-`aws eks update-kubeconfig` wires `aws eks get-token` into the kubeconfig as an exec credential. The bearer token it produces is validated by a **token-authentication webhook** that Floci wires into k3s: the k3s API server POSTs a Kubernetes `TokenReview` to Floci's `/_floci/eks/token-webhook` endpoint, and Floci maps the token to the `system:masters` group (bound to `cluster-admin`). No `aws-iam-authenticator` is required.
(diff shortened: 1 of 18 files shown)


pgermosen (commented):
Really careful, well-tested security work here. Every attack surface I'd think to probe, static and seeded credentials, expiry, cluster binding, session-token possession, signature tampering, malformed tokens, has a dedicated regression test asserting rejection, and I traced the actual code behind each one rather than just trusting the assertions. I also want to flag that the review-bot findings on this PR don't hold up against what's actually here, the tests genuinely exercise and reject every scenario they claim gets through (the test/test and floci/floci credentials, expired sessions, missing or mismatched session tokens), so nothing further is needed on those.

One thing does need fixing though: issueAccessToken in SigninService now calls the new registerSessionForAccount overload that threads the session token through, but SigninServiceTest still verifies the old call shape without it, so several of its mock verifications no longer match the actual invocation. That's a real, current test failure, not a flake, worth updating those verify() calls to include the session token before this merges.

simonepri:
@pgermosen thank you for the review! The SigninServiceTest mocks now use the new overload, and the credential issuance test verifies the session token too.

pgermosen (approved):
That's fixed, confirmed. All four SigninServiceTest mocks are updated to the new signature, and the first one goes further than just matching the new shape, it actually asserts the real session token value now, not just any string. Good strengthening, not just a patch to get it green again.

Combined with the security work from the first pass, which held up on every count I traced, this is solid.

PR #2913, opened 2026-09-01
refactor(rds): split PostgresProtocolHandler into authenticate and bridge primitives

vuthehuyht:
## What this is

Part 1 of splitting #2836 (`feat/redshift-sql-interceptor`) into smaller, independently reviewable PRs, as requested in that PR's review.

This PR carries only the RDS PostgreSQL proxy refactor. It has no dependency on the rest of #2836 and changes no behaviour.

## What changed and why

`PostgresProtocolHandler.handleAuth(...)` did two jobs: run the auth handshake, then relay bytes until close. #2836 needs to keep the handshake but replace the relay with a Redshift SQL interceptor. This splits the method:

- `handleAuth(...)` becomes `authenticate(...)`, which returns the post-startup client `Socket` on success (it may be an SSL-upgraded socket), or `null` when the handshake ended early (bare probe, auth failure, thrown `IOException`).
- The relay body becomes a `public bridge(Socket, Socket)`.
- Callers now do `authenticate(...)` then, if the result is non-null, `bridge(...)`.

Call sites updated:
- `RdsAuthProxy` (POSTGRES case)
- `RedshiftAuthProxy` (rename ripple: it also called `handleAuth`; kept on the plain `bridge`, no interceptor here)
- `PostgresProtocolHandlerTest` (6 sites)

Every early-return in the old `handleAuth` that did `return;` now does `return null;`; the terminal `bridge(client, backend)` becomes `return client;`. Pure mechanical extract, no logic moved between the two methods, no relay behaviour change for RDS or Redshift.

## Source

The three RDS files are taken verbatim from #2836's commit `792aa570` (`refactor(rds): split PostgresProtocolHandler ...`). The one-line `RedshiftAuthProxy` call-site update is new here, needed so `main` compiles after the rename (that file did not exist yet when `792aa570` was written on the branch).

## Testing

`./mvnw test -Dtest=PostgresProtocolHandlerTest,RedshiftAuthProxyTest` -> 14 passing, 0 failures. (The stderr SSL-abort stack traces from `PostgresProtocolHandlerTest` are pre-existing: they come from deliberate handshake-abort cases running on background virtual threads; the JUnit summary is clean.)

## The rest of the split (for context)

Planned follow-up PRs, each stacked on the previous and opened once its predecessor merges:

1. **This PR** - RDS proxy `authenticate` / `bridge` primitives.
2. **DDL keyword rewrite** - `RedshiftSqlInterceptor`: strip `DISTSTYLE` / `DISTKEY` / `SORTKEY` / `ENCODE` from Redshift table DDL, fail-open.
3. **Wire decoder + intercepting bridge** - `PostgresWireDecoder` and the framed `RedshiftInterceptingBridge` (DDL path only), wired into `RedshiftAuthProxy` / `RedshiftProxyManager`.
4. **S3 COPY simulator** - `CopyStatementParser` + `S3CopySimulator.runCopyFrom`, plus a fix: the injected backend query now uses `FORMAT text` unless the statement carries the `CSV` keyword, so a bare pipe-delimited Redshift file loads under TEXT quoting and `\N` NULL rules rather than CSV.
5. **S3 UNLOAD simulator** - `S3CopySimulator.runUnload`, the same TEXT-vs-CSV fix for UNLOAD, and docs for the S3 authorization model (including the residual non-atomic window between the pre-write re-check and `putObject`, kept as-is by design since the emulator has no real IAM).

#2836 stays open as the reference until the split lands.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/proxy/PostgresProtocolHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/proxy/PostgresProtocolHandler.java
@@ -44,17 +44,17 @@ public class PostgresProtocolHandler {
     private static final int SSL_REQUEST_CODE = 80877103;
     private static final int STARTUP_PROTOCOL_VERSION = 196608; // v3.0
 
-    public static void handleAuth(Socket client, Socket backend,
-                                  String masterUsername, String masterPassword, String dbName,
-                                  boolean iamEnabled, RdsSigV4Validator sigV4,
-                                  RdsProxyTlsCertificates tlsCertificates,
-                                  PasswordValidator passwordValidator) throws IOException {
+    public static Socket authenticate(Socket client, Socket backend,
+                                      String masterUsername, String masterPassword, String dbName,
+                                      boolean iamEnabled, RdsSigV4Validator sigV4,
+                                      RdsProxyTlsCertificates tlsCertificates,
+                                      PasswordValidator passwordValidator) throws IOException {
 
         // Phase 1: Read client startup message (possibly preceded by SSL request)
         StartupMessage startup = readStartupMessage(client, tlsCertificates);
         if (startup == null) {
             closeQuietly(client);
-            return;
+            return null;
         }
         client = startup.socket();
         String clientUsername = startup.username();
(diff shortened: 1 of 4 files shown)


pgermosen (approved):
Clean, mechanical split, matches the description exactly. Traced the one subtle bit, the returned client socket can be the SSL-upgraded one from readStartupMessage, and every call site (RdsAuthProxy, RedshiftAuthProxy, and the six updated test sites) correctly bridges on the returned activeClient rather than the pre-call reference, so that's handled right everywhere. No protocol/wire behavior changed anywhere in the diff. Nice, careful split.

Issue #2767, opened 2026-08-30
CloudFront: DefaultCacheBehavior omits TrustedSigners — crashes Terraform AWS provider (nil deref) on every read


YazanYahya:
Floci version: 1.7.0 (floci/floci:latest, build 2026-08-18) · Terraform 1.16.0 · hashicorp/aws v5.100.0

Creating an aws_cloudfront_distribution via Terraform segfaults the AWS provider and kills the whole apply. The create succeeds, but the read-back response's DefaultCacheBehavior omits the TrustedSigners object. Real AWS always includes it, even unused ({"Enabled": false, "Quantity": 0}), and the provider relies on that: internal/service/cloudfront/distribution.go:1673 does len(apiObject.TrustedSigners.Items) with no nil check → panic: runtime error: invalid memory address or nil pointer dereference.

Repro: terraform apply any aws_cloudfront_distribution against floci, or: aws cloudfront create-distribution ... then let Terraform read it (import does it too).

Observed response (get-distribution-config): contains TrustedKeyGroups but no TrustedSigners. Two more fidelity gaps in the same object, which would cause permanent Terraform diffs once the crash is fixed:

- ForwardedValues, MinTTL, DefaultTTL, MaxTTL are dropped (sent on create, not echoed back)
- AllowedMethods.Items comes back with cached methods merged in as duplicates (["GET","HEAD","GET","HEAD","OPTIONS"]) and CachedMethods missing
Expected: echo TrustedSigners (and the fields above) like AWS does.

Side effect worth noting: the crash leaves the created distribution as an orphan Terraform can never manage or delete.

---
Possibly related: same crash class as #1930 (fixed by #1940 for OriginGroups/Restrictions), but a different field. That fix doesn't appear to cover TrustedSigners or AllowedMethods.CachedMethods, which a commenter on #1930 flagged as a related follow-up.

bsaurusrex:
Confirming all three of the read-back gaps listed here still reproduce on `main` at
`62dbd9d`, not just on 1.7.0. From one `create-distribution` sending `AllowedMethods`
(with nested `CachedMethods`), `ForwardedValues`, `MinTTL` and a `LambdaFunctionAssociations`
item:

```
AllowedMethods         : {"Quantity": 5, "Items": ["GET","HEAD","OPTIONS","GET","HEAD"]}
TrustedSigners present : False
ForwardedValues present: False
```

So `CachedMethods` is still merged into `AllowedMethods.Items` as duplicates rather than
[...]

PR #2915, opened 2026-09-02
fix(cloudfront): persist and echo LambdaFunctionAssociations and FunctionAssociations

### LINKED ISSUE #2835: [BUG] CloudFront silently discards LambdaFunctionAssociations, so Lambda@Edge is never wired up
## Service

CloudFront

## AWS API Action

`CreateDistribution`, `GetDistributionConfig`, `GetDistribution`

## Expected behavior

`LambdaFunctionAssociations` sent on `CreateDistribution` should be stored and echoed back.
Real AWS, on a distribution with one viewer-request association:

```
$ aws cloudfront get-distribution-config --id <id> \
    --query 'DistributionConfig.DefaultCacheBehavior.LambdaFunctionAssociations'
{
    "Quantity": 1,
    "Items": [
        {
            "LambdaFunctionARN": "arn:aws:lambda:us-east-1:<acct>:function:<fn>:9",
            "EventType": "viewer-request",
            "IncludeBody": false
        }
    ]
}
```

## Actual behavior

The associations are accepted on create and then **silently dropped**. Every read returns:

```json
"LambdaFunctionAssociations": {"Quantity": 0}
```

No error, no warning — the create returns 200 and the wiring simply is not there.

## Why it matters

This is worse than a cosmetic read-back gap. Lambda@Edge is *only* configured through this
member, so a distribution that is supposed to run a viewer-request function against every
request has no function attached at all. Anything relying on an edge auth or rewrite
function is silently unprotected rather than failing loudly, and Terraform additionally
re-adds the block on every plan:

```
~ resource "aws_cloudfront_distribution" "site" {
    ~ default_cache_behavior {
        + lambda_function_association {
            + event_type   = "viewer-request"
            + include_body = false
            + lambda_arn   = (known after apply)
          }
      }
  }
```

## Reproduction

```bash

cat > dist.json <<'JSON'
{
  "CallerReference": "cf-probe-1",
  "Comment": "cf probe",
  "Enabled": true,
  "Origins": {"Quantity": 1, "Items": [{"Id": "o1", "DomainName": "example.com",
    "CustomOriginConfig": {"HTTPPort": 80, "HTTPSPort": 443, "OriginProtocolPolicy": "https-only"}}]},
  "DefaultCacheBehavior": {
    "TargetOriginId": "o1",
    "ViewerProtocolPolicy": "redirect-to-https",
    "AllowedMethods": {"Quantity": 3, "Items": ["GET","HEAD","OPTIONS"],
                       "CachedMethods": {"Quantity": 2, "Items": ["GET","HEAD"]}},
    "LambdaFunctionAssociations": {"Quantity": 1, "Items": [
      {"LambdaFunctionARN": "arn:aws:lambda:us-east-1:000000000000:function:edge-fn:1",
       "EventType": "viewer-request", "IncludeBody": false}]},
    "ForwardedValues": {"QueryString": false, "Cookies": {"Forward": "none"}},
    "MinTTL": 0
  }
}
JSON

ID=$(aws $E cloudfront create-distribution --distribution-config file://dist.json \
       --query 'Distribution.Id' --output text)
aws $E cloudfront get-distribution-config --id "$ID" \
    --query 'DistributionConfig.DefaultCacheBehavior.LambdaFunctionAssociations'
# => {"Quantity": 0}
```

## Related

#2767 reports several other members of the same `DefaultCacheBehavior` object being dropped
or mangled on read-back (`TrustedSigners`, `ForwardedValues`/TTLs, and `CachedMethods` being
merged into `AllowedMethods.Items` as duplicates). All of those still reproduce on `main` at
`62dbd9d` — from the same run as above:

```
AllowedMethods: {"Quantity": 5, "Items": ["GET","HEAD","OPTIONS","GET","HEAD"]}
TrustedSigners present: False
ForwardedValues present: False
```

`LambdaFunctionAssociations` is filed separately because it is a different failure — not a
diff or a provider crash, but a security-relevant control that is accepted and then
discarded. If a maintainer would rather fix the whole object at once, this can be folded
into #2767.

## Environment

vuthehuyht:
## Summary

CloudFront `CreateDistribution` accepted `LambdaFunctionAssociations` and then silently discarded them: `parseDefaultCacheBehavior` / `parseCacheBehaviors` never read the element, and `xmlDefaultCacheBehavior` / `xmlCacheBehavior` hardcoded `<Quantity>0</Quantity>`. Every `GetDistributionConfig` / `GetDistribution` read back `Quantity: 0`, so a distribution meant to run a viewer-request Lambda@Edge function had no function wired up at all — with no error. `FunctionAssociations` (CloudFront Functions) had the identical bug, so this PR fixes both.

- `parseDefaultCacheBehavior` + `parseCacheBehaviors`: read `LambdaFunctionARN`, `EventType`, `IncludeBody` (Lambda) and `FunctionARN`, `EventType` (Functions) into the model fields that already existed on `DefaultCacheBehavior` / `CacheBehavior`. Persistence is the JSON store, so the values round-trip to disk with no model changes.
- `xmlDefaultCacheBehavior` + `xmlCacheBehavior`: replace the two hardcoded blocks with `xmlLambdaFunctionAssociations` / `xmlFunctionAssociations`; still emit `<Quantity>0</Quantity>` when empty (no read-back change for distributions that never set the member).
- Validate `EventType`: Lambda accepts `viewer-request`, `viewer-response`, `origin-request`, `origin-response`; CloudFront Functions accept only `viewer-request` and `viewer-response`. An empty ARN or an invalid`EventType` returns `400 InvalidArgument`.
- Validate the declared `Quantity`: a malformed value returns `400 InvalidArgument`, and one that differs from the item count returns `400 InconsistentQuantities` (matching the existing `TrustedKeyGroups` handling) instead of being silently normalized on read-back.

Out of scope: the other `DefaultCacheBehavior` members reported in #2767 (`TrustedSigners`, `ForwardedValues`/TTLs, `CachedMethods` merged into `AllowedMethods`). ARN shape / qualified-version / region checks are not replicated.

Closes #2835

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Bug fix. Incorrect behavior: `LambdaFunctionAssociations` (and `FunctionAssociations`) sent on `CreateDistribution` were accepted with a 200 and then dropped; every subsequent `GetDistributionConfig` / `GetDistribution` returned `LambdaFunctionAssociations: {"Quantity": 0}`, and Terraform re-added the `lambda_function_association` block on every plan. After this change the associations are stored and echoed back with `LambdaFunctionARN`, `EventType`, and `IncludeBody`, matching real CloudFront. Verified with AWS CLI v2 against the reproduction in #2835.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
@@ -2802,6 +2802,25 @@ public class CloudFrontController {
         }
     }
 
+    private static int parseAssociationsQuantity(String value) {
+        try {
+            int quantity = Integer.parseInt(value.trim());
+            if (quantity < 0) {
+                throw new NumberFormatException("negative quantity");
+            }
+            return quantity;
+        } catch (NumberFormatException e) {
+            throw new AwsException("InvalidArgument",
+                    "The association Quantity must be a non-negative integer.", 400);
+        }
+    }
+
+    private static void validateAssociationsQuantity(Integer declared, int itemCount) {
+        if (declared != null && declared != itemCount) {
+            throw inconsistentQuantities();
+        }
+    }
+
     private DefaultCacheBehavior parseDefaultCacheBehavior(String body) {
         DefaultCacheBehavior dcb = new DefaultCacheBehavior();
         if (body == null || body.isEmpty()) {
@@ -2823,6 +2842,8 @@ public class CloudFrontController {
             Map<String, String> currentFunctionAssociation = null;
             List<Map<String, Object>> lambdaAssociations = new ArrayList<>();
             List<Map<String, String>> functionAssociations = new ArrayList<>();
+            Integer lambdaAssociationsQuantity = null;
+            Integer functionAssociationsQuantity = null;
 
             while (r.hasNext()) {
                 int event = r.next();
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Nice, careful fix, and the EventType handling in particular is exactly right: I checked the CloudFront docs, and the FunctionAssociation page's schema line lists all four event types (a shared-schema artifact), but the prose is explicit that CloudFront Functions only support viewer-request/viewer-response. You followed the prose rather than the naive enum, which is an easy detail to get wrong.

Traced both parse loops, the XML serialization mirrors the existing xmlTrustedKeyGroups pattern, the ordered-cache-behavior loop correctly resets the association state between items (the one spot a stale value could leak across iterations), and the quantity-mismatch validation reuses the existing InconsistentQuantities path rather than inventing something new. Good bracing fix on the follow-up commit too, confirmed clean in the final diff.

Good test coverage: round-trip for both association types, an invalid event type, and a quantity mismatch.

PR #2921, opened 2026-09-02
fix(ec2): tolerate a null groupId in the network-interface group-id filter

matanbaruch:
## What

The `group-id` filter for `DescribeNetworkInterfaces` resolves to `matchesValue(List<String> patterns, String value)`, which calls `value.matches(...)` with no null guard. A real `GroupIdentifier` entry whose `groupId` is null (a name-only association) NPEs the whole request:

```
java.lang.NullPointerException: Cannot invoke "String.matches(String)" because "value" is null
```

The sibling overload `matchesValue(String resourceValue, List<String> filterValues)` already normalizes with `Objects.toString(..., "")`. This aligns the two so every network-interface leaf filter (`group-id`, `subnet-id`, `status`, `private-dns-name`, ...) tolerates a null field instead of failing.

## Fix

One line: normalize `value` with `Objects.toString(value, "")` before `.matches(...)`, mirroring the existing null-safe overload.

## Test

`describeNetworkInterfacesGroupIdFilterToleratesNullGroupId` puts a null-`groupId` entry **ahead of** a real group on the same interface, so `anyMatch` actually evaluates the null path (and the real group still matches). It errors with the NPE above before the fix and passes after. Full `Ec2ServiceTest` stays green (122 tests).

## Context

Spotted by @pgermosen while reviewing #2887 (null group *entries*). That PR guards a null *element* in the list; this guards a real element with a null `groupId`. The two are independent, so this is a separate PR off `main` rather than a change to #2887.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -5732,8 +5732,9 @@ public class Ec2Service implements ContainerTeardown, ResourceProvider {
     }
 
     private boolean matchesValue(List<String> patterns, String value) {
+        String normalizedValue = Objects.toString(value, "");
         return patterns.stream()
-                .anyMatch(pattern -> value.matches(wildcardToRegex(pattern)));
+                .anyMatch(pattern -> normalizedValue.matches(wildcardToRegex(pattern)));
     }
 
     private boolean matchesFilters(Object resource, Map<String, List<String>> filters, String region) {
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Clean, minimal fix. Checked it against the sibling single-value overload, matchesValue(String, List<String>), which already normalizes null with Objects.toString(value, "") before matching, so this makes the two overloads consistent with each other rather than introducing a new pattern. It also fixes every list-of-values filter that routes through this method, not just group-id.

On greptile's finding: I don't think it holds up here. The sibling overload already lets a null resource value match a wildcard filter for every other filter type in the file (vpc-id, subnet-id, tags, etc.), so returning false on null in just this overload would make the two inconsistent with each other for no real benefit. And a GroupIdentifier with a null groupId isn't something real AWS actually returns, so there's no live-AWS behavior at stake either way, this is purely an internal defensive guard.

Test reproduces the exact NPE, null entry ordered ahead of the real one so anyMatch actually evaluates the null path first. Good work tracking this down while reviewing #2887.

PR #2923, opened 2026-09-02
fix(dynamodb): reject unused ExpressionAttributeNames/Values on PutItem, DeleteItem, Scan, Query

### LINKED ISSUE #2893: [BUG] DynamoDB: unused ExpressionAttributeNames/Values accepted on PutItem, DeleteItem, Scan and Query
## Service

DynamoDB

## AWS API Action

`PutItem`, `DeleteItem`, `Scan` (unused `ExpressionAttributeNames`) and `Query` (unused `ExpressionAttributeValues`)

## Expected behavior

Real AWS rejects a request that supplies an `ExpressionAttributeNames` or `ExpressionAttributeValues` entry which no expression in the request references. Measured against the live service in `ap-southeast-1`:

```
PutItem     ValidationException: 1 validation error detected: Value provided in ExpressionAttributeNames unused in expressions: keys: {#st}
DeleteItem  ValidationException: 1 validation error detected: Value provided in ExpressionAttributeNames unused in expressions: keys: {#st}
Scan        ValidationException: Value provided in ExpressionAttributeNames unused in expressions: keys: {#st}
Query       ValidationException: Value provided in ExpressionAttributeValues unused in expressions: keys: {:unused}
```

Note the wording is not uniform: `PutItem` and `DeleteItem` carry the `1 validation error detected: ` prefix, `Scan` and `Query` do not.

The `PutItem` and `DeleteItem` measurements were made against a table that does not exist, with a control proving the validation runs ahead of the table lookup, so no write was performed:

```
# unused alias present
PutItem no-such-tbl ... --expression-attribute-names '{"#st":"status"}'
  -> ValidationException (message above)

# control: identical call, alias removed
PutItem no-such-tbl ... (no --expression-attribute-names)
  -> ResourceNotFoundException: Requested resource not found
```

## Actual behavior

Floci accepts all four and performs the operation.

For `PutItem` this is not only a missing error: the item AWS would have refused is durably written.

```
$ aws dynamodb put-item --table-name ProbeTbl \
    --item '{"PK":{"S":"rw-ok"},"SK":{"S":"s"},"status":{"S":"active"}}' \
    --condition-expression 'attribute_not_exists(PK)' \
    --expression-attribute-names '{"#st":"status"}'
# no error

$ aws dynamodb get-item --table-name ProbeTbl --key '{"PK":{"S":"rw-ok"},"SK":{"S":"s"}}'
{"Item": {"PK": {"S": "rw-ok"}, "SK": {"S": "s"}, "status": {"S": "active"}}}
```

Coverage measured across actions, unused `#st` in `ExpressionAttributeNames`:

| Action | Floci | Live AWS |
| --- | --- | --- |
| `PutItem` | accepted | ValidationException |
| `DeleteItem` | accepted | ValidationException |
| `Scan` | accepted | ValidationException |
| `Query` | ValidationException | ValidationException |
| `UpdateItem` | ValidationException | ValidationException |

Unused `ExpressionAttributeValues` on `Query`: Floci accepts, live AWS rejects.

This lines up with where the helpers are called. In `DynamoDbJsonHandler`, `checkUnusedEan` is invoked from `handleUpdateItem` (line 543) and `handleQuery` (line 896) only, and `checkUnusedEav` from `handleUpdateItem` (line 545) only. `handlePutItem`, `handleDeleteItem` and `handleScan` never call either, and `handleQuery` calls only the names variant.

The message wording Floci already emits is byte-identical to the live service, so this looks like missing call sites rather than a wording problem.

## Reproduction

```bash
docker run -d --name floci -p 4566:4566 floci/floci:main
aws dynamodb create-table --endpoint-url http://localhost:4566 --table-name ProbeTbl \
  --attribute-definitions AttributeName=PK,AttributeType=S AttributeName=SK,AttributeType=S \

# accepted by Floci, ValidationException on AWS
aws dynamodb put-item --endpoint-url http://localhost:4566 --table-name ProbeTbl \
  --item '{"PK":{"S":"rw-ok"},"SK":{"S":"s"},"status":{"S":"active"}}' \
  --condition-expression 'attribute_not_exists(PK)' \
  --expression-attribute-names '{"#st":"status"}'

# accepted by Floci, ValidationException on AWS
aws dynamodb scan --endpoint-url http://localhost:4566 --table-name ProbeTbl \
  --filter-expression 'PK = :p' --expression-attribute-values '{":p":{"S":"zz"}}' \
  --expression-attribute-names '{"#st":"status"}'

# accepted by Floci, ValidationException on AWS
aws dynamodb query --endpoint-url http://localhost:4566 --table-name ProbeTbl \
  --key-condition-expression 'PK = :p' \
  --expression-attribute-values '{":p":{"S":"zz"},":unused":{"S":"q"}}'
```

## Context

Found while running the SDK compatibility suite against the container-backed backend in #2871. `amazon/dynamodb-local` rejects these, Floci accepts them, and the live service agrees with the container. The conformance case that surfaced it is `reserved word aliased via ExpressionAttributeNames passes` in `compatibility-tests/sdk-test-node/tests/dynamodb-conformance.test.ts`, whose `PutItem` supplies `ExpressionAttributeNames: { '#st': 'status' }` alongside `ConditionExpression: 'attribute_not_exists(pk)'`, which never references `#st`. That test currently asserts the lenient behaviour as passing, so tightening this needs the test updated in the same change.

I have deliberately not changed this in #2871: that PR adds a backend and does not touch native engine semantics.

## Environment

- Floci version / image tag: built from `main` at `2bf9b57`, reports `2.0.1`
- How you're running Floci: Docker, `docker build -f docker/Dockerfile`
- Live AWS comparison: `ap-southeast-1`, AWS CLI v2

vuthehuyht:
## Summary

DynamoDB accepted unused `ExpressionAttributeNames` / `ExpressionAttributeValues` entries on `PutItem`, `DeleteItem`, `Scan` and `Query`, where real AWS returns `ValidationException`. The helpers `checkUnusedEan` / `checkUnusedEav` already
existed and produce the exact AWS messages, but were only wired into `handleUpdateItem` and (for names only) `handleQuery`. On the write paths this meant items AWS would have rejected were durably persisted.

- `handlePutItem` / `handleDeleteItem`: after `ExpressionEvaluator.validateExpression`, call `checkUnusedEan` / `checkUnusedEav` against the tokens referenced by `ConditionExpression` (the only expression these operations accept).
- `handleScan`: call both helpers against the tokens referenced by `FilterExpression` (plus `ProjectionExpression` for names).
- `handleQuery`: add the missing `checkUnusedEav` call alongside the existing `checkUnusedEan`, against `KeyConditionExpression` + `FilterExpression`. The check runs unconditionally so a legacy `KeyConditions` request carrying a stray alias map is validated too; the token sets are still derived only from the expressions that can legitimately reference those maps, so a correct request is unaffected.

Sub-case — no expression parameter at all, but `ExpressionAttributeNames` / `ExpressionAttributeValues` present:

- `handlePutItem` / `handleDeleteItem` return AWS's `<param> can only be specified when using expressions: ConditionExpression is null`. Both ops only accept `ConditionExpression`, so the named param is unambiguous, and the wording matches the pre-existing PutItem+EAV guard from #826. `checkUnusedEan` / `checkUnusedEav` still handle the case where an alias is merely unreferenced.
- `handleScan` / `handleQuery` still return `... unused in expressions: keys: {...}` for this sub-case. Those ops accept several expression parameters and AWS's exact wording when all are absent is unverified; tracked for a follow-up backed by a live measurement. The shared helper is parameterised on the message, so it is a one-line change per op once the strings are known.

Either way the request is now a `400 ValidationException` rather than silently accepted.

Closes #2893

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Bug fix. Incorrect behavior: `PutItem` / `DeleteItem` / `Scan` / `Query` accepted `ExpressionAttributeNames` / `ExpressionAttributeValues` entries not referenced by any expression and performed the operation (including durably writing items). After this change these requests return `400 ValidationException` with `Value provided in ExpressionAttributeNames unused in expressions: keys: {...}` / `Value provided in ExpressionAttributeValues unused in expressions: keys: {...}`, matching real DynamoDB. When no expression parameter is supplied at all, `PutItem` / `DeleteItem` return `<param> can only be specified when using expressions: ConditionExpression is null` (also matching real DynamoDB); the equivalent wording for `Scan` / `Query` is left to a follow-up. `UpdateItem` already behaved correctly and is unchanged.

## Checklist

- [x] `./mvnw test` passes locally (DynamoDb* + ExpressionEvaluatorTest: 555 run, 0 failures)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbConformanceChangesTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbConformanceChangesTest.java
@@ -817,11 +817,12 @@ class DynamoDbConformanceChangesTest {
 
     @Test @Order(101)
     void reservedWordWithAliasPasses() {
-        // Using #status alias should work fine
+        // Using the #st alias should work fine. The alias must be referenced by the
+        // expression: AWS rejects ExpressionAttributeNames entries left unused.
         assertThatCode(() -> ddb.putItem(r -> r
                 .tableName(TABLE)
                 .item(Map.of("pk", av("rw-alias"), "sk", av("s"), "status", av("ok")))
-                .conditionExpression("attribute_not_exists(pk)")
+                .conditionExpression("attribute_not_exists(#st)")
                 .expressionAttributeNames(Map.of("#st", "status"))))
                 .doesNotThrowAnyException();
     }
--- a/compatibility-tests/sdk-test-node/tests/dynamodb-conformance.test.ts
+++ b/compatibility-tests/sdk-test-node/tests/dynamodb-conformance.test.ts
@@ -612,11 +612,12 @@ describe('DynamoDB — Phase 9: Reserved words', () => {
   });
 
   it('reserved word aliased via ExpressionAttributeNames passes', async () => {
-    // Using #st alias for the reserved word 'status'
+    // Using #st alias for the reserved word 'status'. The alias must be referenced
+    // by the expression: AWS rejects ExpressionAttributeNames entries left unused.
     await ddb.send(new PutItemCommand({
       TableName: table,
       Item: { pk: s('rw-ok'), sk: s('s'), status: s('active') },
-      ConditionExpression: 'attribute_not_exists(pk)',
+      ConditionExpression: 'attribute_not_exists(#st)',
       ExpressionAttributeNames: { '#st': 'status' },
     }));
 
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbJsonHandler.java
@@ -328,10 +328,10 @@ public class DynamoDbJsonHandler {
                     + "Non-expression parameters: {Expected} Expression parameters: {ConditionExpression}", 400);
         }
 
-        if (exprAttrValues != null && conditionExpression == null) {
-            throw new AwsException("ValidationException",
-                    "ExpressionAttributeValues can only be specified when using expressions: ConditionExpression is null", 400);
-        }
+        // EAN/EAV with no expression to reference them: AWS reports "can only be specified
+        // when using expressions", not the "unused in expressions" wording (#2893).
+        rejectExprAttrsWithoutExpression(exprAttrNames, exprAttrValues,
+                conditionExpression != null, "ConditionExpression is null");
 
         ExpressionEvaluator.validateExpression(conditionExpression, "ConditionExpression", exprAttrNames, exprAttrValues);
 
@@ -442,6 +442,11 @@ public class DynamoDbJsonHandler {
                     n + " validation error" + (n > 1 ? "s" : "") + " detected: "
                     + String.join("; ", delValidationErrors), 400);
         }
+
+        // EAN/EAV with no expression to reference them: AWS reports "can only be specified
+        // when using expressions", not the "unused in expressions" wording (#2893).
+        rejectExprAttrsWithoutExpression(exprAttrNames, exprAttrValues,
+                conditionExpression != null, "ConditionExpression is null");
 
         ExpressionEvaluator.validateExpression(conditionExpression, "ConditionExpression", exprAttrNames, exprAttrValues);
 
@@ -1863,6 +1868,29 @@ public class DynamoDbJsonHandler {
         Set<String> tokens = new LinkedHashSet<>();
         for (String expr : expressions) extractPrefixedTokens(expr, ':', tokens);
         return tokens;
+    }
+
+    /**
+     * Rejects ExpressionAttributeNames/Values supplied when the request carries no expression to
+     * reference them. AWS answers this sub-case with "&lt;param&gt; can only be specified when using
+     * expressions: &lt;nullExpressionParams&gt;" rather than the "unused in expressions" wording used
+     * once an alias is merely unreferenced (#2893). {@code nullExpressionParams} must spell the absent
+     * expression parameter(s) exactly as AWS reports them for the operation, e.g. "ConditionExpression is null".
+     * EAV is checked before EAN to match the pre-existing PutItem guard.
+     */
+    private static void rejectExprAttrsWithoutExpression(JsonNode exprAttrNames, JsonNode exprAttrValues,
+            boolean hasAnyExpression, String nullExpressionParams) {
+        if (hasAnyExpression) {
+            return;
+        }
+        if (exprAttrValues != null) {
+            throw new AwsException("ValidationException",
+                    "ExpressionAttributeValues can only be specified when using expressions: " + nullExpressionParams, 400);
+        }
+        if (exprAttrNames != null) {
+            throw new AwsException("ValidationException",
+                    "ExpressionAttributeNames can only be specified when using expressions: " + nullExpressionParams, 400);
+        }
     }
 
     private static void checkUnusedEan(JsonNode exprAttrNames, Set<String> usedHashTokens) {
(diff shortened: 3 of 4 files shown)


pgermosen (commented):
Solid fix, closes a real gap: PutItem/DeleteItem/Scan/Query were silently accepting (and durably writing) unused ExpressionAttributeNames/Values entries that real AWS rejects. It just wires the existing checkUnusedEan/checkUnusedEav helpers (already used by UpdateItem) into the other handlers rather than introducing new matching logic, so the risk surface here is small.

Checked both of greptile's findings against the final diff and both look genuinely resolved: the legacy KeyConditions + FilterExpression gap was fixed across the follow-up commits, the if (keyConditionExpr != null) guard is gone and the checks run unconditionally now, and there's a test exercising that exact combination. No em-dashes left either.

Good catch fixing the two SDK conformance tests (Java and Node) whose reservedWordWithAliasPasses case supplied an EAN alias without ever referencing it, that would've started failing under this validation, and referencing the alias in the expression is the right fix rather than working around it.

One question, not a blocker: you call out in the description that when EAN/EAV is present with no expression at all, this returns the "unused in expressions" wording instead of AWS's "can only be specified when using expressions" wording for that sub-case. I checked, the only pre-existing guard using that AWS wording covers just PutItem+EAV, so that gap is actually live for most of what's newly added here, not just the one case already handled. Still a strict improvement either way (400 instead of silently accepting), just wondering if it's worth a quick probe against real AWS to confirm the wording, or worth tracking separately if it turns out to matter.

vuthehuyht:
Thanks for the thorough review @pgermosen! Good point on the sub-case wording difference when no expression parameter is present at all. Since both paths now correctly reject with a 400 ValidationException,I think tracking the exact per-operation wording parity in a separate follow-up issue would keep this PR scoped and safe to merge. Let me know if you'd prefer to address it here or track it separately.

pgermosen:
Good question, and I'd rather fix it here than track separately. The reason: this isn't a pre-existing divergence you're just not touching, before this PR, ExpressionAttributeNames/Values with no expression at all on DeleteItem/Scan/Query was silently accepted, there was no validation there to have wrong wording in the first place. The only guard with AWS's actual wording ("can only be specified when using expressions") covered just PutItem+EAV. So the wrong message for the no-expression-at-all sub-case only exists because this PR's own new checks create that code path, it's new code producing the wrong string, not scope creep onto something unrelated. Given that, I'd rather see it land right the first time than open a follow-up for something this PR itself introduces.

vuthehuyht:
Split this rather than doing it all in one go.

`PutItem` and `DeleteItem` (adee6e48): the no-expression-at-all sub-case now returns AWS's `<param> can only be specified when using expressions: ConditionExpression is null`. Both ops only accept `ConditionExpression`, so there's no ambiguity about which parameter to name, and that's the same string the codebase has used since #826 for the pre-existing PutItem+EAV guard. New checks are routed through a shared `rejectExprAttrsWithoutExpression` helper (EAV before EAN, matching the old guard's order). That closes the sub-case on the two write paths where a silent accept was worst.

`Scan` and `Query` still return the `unused in expressions` 400 for that sub-case. Those ops accept several expression parameters (`FilterExpression`/`ProjectionExpression`, plus `KeyConditionExpression` for Query), and AWS's exact wording when all of them are absent — which parameter name(s) it lists, `is null` vs `are null` — I can't confirm without a live probe, and I'd rather not hardcode a guess into a conformance codebase. Could we track the Scan/Query wording as a follow-up backed by an `ap-southeast-1` measurement? The helper is already parameterised on the message, so it's a one-line change per op once we have the strings.

pgermosen (approved):
The PutItem/DeleteItem wording fix looks right. `rejectExprAttrsWithoutExpression` runs before the unused-token checks, so it only fires when there's genuinely no expression to reference the alias map, and EAV wins over EAN when both are present, matching the order the pre-existing PutItem+EAV guard used. The four new test cases cover it well.

Good call proposing the Scan/Query wording as a follow-up instead of guessing at it. It's a read op either way, so there's no data-integrity risk sitting behind the imprecise wording, just a message-text gap worth tracking properly rather than baking in an unverified guess. Please open that issue when you get a chance.

Thanks for working through this one carefully.

PR #2928, opened 2026-09-02
fix(iam): resolve DynamoDB table ARN from request body for IAM enforcement

### LINKED ISSUE #2925: [BUG] IAM enforcement evaluates DynamoDB against a wildcard table ARN, so resource-scoped policies never match
## Service

IAM (policy enforcement), affecting DynamoDB

## AWS API Action

Any DynamoDB action evaluated with `FLOCI_SERVICES_IAM_ENFORCEMENT_ENABLED=true`. Measured with `GetItem`.

## Expected behavior

An identity policy that grants a DynamoDB action on a specific table ARN allows requests against that table. Resource-level permissions for DynamoDB are documented here:

- https://docs.aws.amazon.com/service-authorization/latest/reference/list_dynamodb.html
- https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/specifying-conditions.html

## Actual behavior

With enforcement on, only `"Resource": "*"` grants access. A policy naming the table's own ARN, byte for byte as `DescribeTable` returns it, is denied.

The same assumed-role session, varying only the policy:

```
Action * / Resource *                          -> ALLOWED
Action dynamodb:GetItem / Resource *           -> ALLOWED
Action * / Resource arn:aws:dynamodb:us-east-1:000000000000:table/FgacTable   -> AccessDeniedException
Action dynamodb:GetItem / same table ARN       -> AccessDeniedException
```

`DescribeTable` reports the ARN as `arn:aws:dynamodb:us-east-1:000000000000:table/FgacTable`, so the string in the policy is the resource's real ARN.

## Cause

`ResourceArnBuilder.buildDynamoDbArn` does not read `TableName`, and says so:

```java
private String buildDynamoDbArn(ContainerRequestContext ctx, String region, String accountId) {
    // TableName comes in the JSON body; use wildcard since we don't parse the body here
    return AwsArnUtils.Arn.of("dynamodb", region, accountId, "table/*").toString();
}
```

Every DynamoDB request is therefore evaluated against `arn:aws:dynamodb:<region>:<account>:table/*`. That matches a policy resource of `*` and fails to match any specific table ARN, which is exactly the measured split above. DynamoDB puts `TableName` in the JSON body rather than the URI, so the builder would need the parsed body to do better.

`buildKinesisArn` (`stream/*`) and `buildSecretsManagerArn` (`secret:*`) have the same shape, so the same limitation likely applies to those services. I only measured DynamoDB.

## Reproduction

Executed against `floci/floci:2.0.1`.

```bash
docker run -d --name floci-iam -p 4631:4566 \
  -e FLOCI_SERVICES_IAM_ENFORCEMENT_ENABLED=true floci/floci:2.0.1
```

```python
import boto3, botocore, json
E="http://localhost:4631"; R="us-east-1"; T="FgacTable"
kw=dict(endpoint_url=E, region_name=R, aws_access_key_id="test", aws_secret_access_key="test")
ddb=boto3.client("dynamodb",**kw); iam=boto3.client("iam",**kw); sts=boto3.client("sts",**kw)
acct=sts.get_caller_identity()["Account"]

ddb.create_table(TableName=T,
    AttributeDefinitions=[{"AttributeName":"PK","AttributeType":"S"},
                          {"AttributeName":"SK","AttributeType":"S"}],
    KeySchema=[{"AttributeName":"PK","KeyType":"HASH"},
               {"AttributeName":"SK","KeyType":"RANGE"}],
    BillingMode="PAY_PER_REQUEST")
ddb.get_waiter("table_exists").wait(TableName=T)
ddb.put_item(TableName=T, Item={"PK":{"S":"USER_alice"},"SK":{"S":"PROFILE"}})
real_arn = ddb.describe_table(TableName=T)["Table"]["TableArn"]

trust={"Version":"2012-10-17","Statement":[{"Effect":"Allow",
    "Principal":{"AWS":f"arn:aws:iam::{acct}:root"},"Action":"sts:AssumeRole"}]}
iam.create_role(RoleName="FgacRole", AssumeRolePolicyDocument=json.dumps(trust))

def attempt(stmt, label):
    iam.put_role_policy(RoleName="FgacRole", PolicyName="P",
        PolicyDocument=json.dumps({"Version":"2012-10-17","Statement":[stmt]}))
    c=sts.assume_role(RoleArn=f"arn:aws:iam::{acct}:role/FgacRole",
                      RoleSessionName="probe")["Credentials"]
    s=boto3.client("dynamodb", endpoint_url=E, region_name=R,
        aws_access_key_id=c["AccessKeyId"], aws_secret_access_key=c["SecretAccessKey"],
        aws_session_token=c["SessionToken"])
    try:
        s.get_item(TableName=T, Key={"PK":{"S":"USER_alice"},"SK":{"S":"PROFILE"}})
        print(f"{label:52} -> ALLOWED")
    except botocore.exceptions.ClientError as e:
        print(f"{label:52} -> {e.response['Error']['Code']}")

attempt({"Effect":"Allow","Action":"*","Resource":"*"}, "Action * / Resource *")
attempt({"Effect":"Allow","Action":"dynamodb:GetItem","Resource":"*"}, "GetItem / Resource *")
attempt({"Effect":"Allow","Action":"*","Resource":real_arn}, "Action * / real table ARN")
attempt({"Effect":"Allow","Action":"dynamodb:GetItem","Resource":real_arn}, "GetItem / real table ARN")
```

Output:

```
Action * / Resource *                                -> ALLOWED
GetItem / Resource *                                 -> ALLOWED
Action * / real table ARN                            -> AccessDeniedException
GetItem / real table ARN                             -> AccessDeniedException
```

## Impact

With enforcement on, a DynamoDB policy is effectively all or nothing: either it grants `*` or it denies. Least-privilege policies that scope to a table, which is the normal shape for an application role, deny every request. It also blocks emulating DynamoDB fine-grained access control, since those policies scope to a table ARN before they even reach a condition. Filed separately as the condition-key gap.

## Environment

- Floci version / image tag: `floci/floci:2.0.1`
- How you're running Floci: Docker, with `FLOCI_SERVICES_IAM_ENFORCEMENT_ENABLED=true`

vuthehuyht:
## Summary

Resolves DynamoDB table ARN from the request JSON body during IAM enforcement instead of returning a static wildcard table ARN (`arn:aws:dynamodb:<region>:<account>:table/*`).

Closes #2925

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- **Incorrect behavior**: `ResourceArnBuilder.buildDynamoDbArn` returned a static wildcard ARN (`arn:aws:dynamodb:<region>:<account>:table/*`). When evaluating table-scoped policies (e.g. `arn:aws:dynamodb:us-east-1:000000000000:table/FgacTable`), `IamPolicyEvaluator.globMatches` matched the literal policy ARN pattern against `table/*` and evaluated to `false`, causing unexpected `AccessDeniedException` 403 errors even when permissions were correctly granted.
- **Fixed behavior**: Parses the JSON request body in `ResourceArnBuilder` to extract `TableName` (supporting both short table names and full ARNs), `ResourceArn`, `TableArn`, `StreamArn`, `ExportArn`, as well as batch and transaction operations (`RequestItems`, `TransactItems`). If no specific table is targeted (such as in `ListTables`), it falls back to `*`. The request entity stream is buffered and reset so subsequent filters and endpoints read the payload intact.
- Enhanced JSON body resource resolution similarly for Kinesis, Secrets Manager, SSM, SQS, and SNS.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/IamEnforcementFilter.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/IamEnforcementFilter.java
@@ -169,7 +169,7 @@ public class IamEnforcementFilter implements ContainerRequestFilter {
             caller = caller.withScpLevels(scpLevels);
         }
 
-        String resource = arnBuilder.build(credentialScope, ctx, region, accountId);
+        List<String> resources = arnBuilder.buildResources(credentialScope, ctx, region, accountId);
 
         Map<String, String> conditionContext = conditionContextResolver.resolve(credentialScope, action, ctx);
 
@@ -187,15 +187,18 @@ public class IamEnforcementFilter implements ContainerRequestFilter {
             conditionContext.put("aws:PrincipalArn", principalArn.get());
         }
 
-        Decision decision = evaluator.evaluate(caller, null, action, resource, conditionContext);
-        if (decision == Decision.DENY) {
-            LOG.infov("IAM enforcement DENY: akid={0} action={1} resource={2}", akid, action, resource);
-            String denyMessage = "User: arn:aws:iam::" + accountId
-                    + ":user/" + akid + " is not authorized to perform: " + action
-                    + " on resource: \"" + resource + "\""
-                    + " because no identity-based policy allows the " + action + " action";
-            emitS3DenialIfApplicable(akid, action, resource, ctx, region, denyMessage);
-            ctx.abortWith(accessDeniedResponse(action, credentialScope, ctx.getMediaType()));
+        for (String resource : resources) {
+            Decision decision = evaluator.evaluate(caller, null, action, resource, conditionContext);
+            if (decision == Decision.DENY) {
+                LOG.infov("IAM enforcement DENY: akid={0} action={1} resource={2}", akid, action, resource);
+                String denyMessage = "User: arn:aws:iam::" + accountId
+                        + ":user/" + akid + " is not authorized to perform: " + action
+                        + " on resource: \"" + resource + "\""
+                        + " because no identity-based policy allows the " + action + " action";
+                emitS3DenialIfApplicable(akid, action, resource, ctx, region, denyMessage);
+                ctx.abortWith(accessDeniedResponse(action, credentialScope, ctx.getMediaType()));
+                return;
+            }
         }
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbPartiQLParser.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbPartiQLParser.java
@@ -5,7 +5,7 @@ import io.github.hectorvent.floci.core.common.AwsException;
 
 import java.util.*;
 
-class DynamoDbPartiQLParser {
+public class DynamoDbPartiQLParser {
 
     enum TType {
         SELECT, FROM, WHERE, INSERT, INTO, VALUE,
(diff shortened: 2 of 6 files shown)


pgermosen (commented):
Good fix overall, and I appreciate that the follow-up commits closed out all three of greptile's findings properly, I traced the multi-table batch/transact handling and the PartiQL Statements/TransactStatements walk and they're both genuinely fixed, not just silenced. The multi-resource deny-on-any loop in the filter is the right AWS semantics too, and the entity-stream buffering has a test proving downstream code can still read the body afterward.

One thing I do want fixed before merge though, not just tracked: the PartiQL table extraction can diverge from what actually executes. extractDynamoDbTableFromPartiQL uses a regex that just scans the raw statement text and takes the first FROM/INTO/UPDATE it finds. The real PartiQL engine in this codebase (DynamoDbPartiQLParser/DynamoDbPartiQLHandler) is an actual tokenizer, where a quoted string is one opaque identifier and the target table is whatever structurally follows the FROM token. Those two can disagree. For example:

SELECT "FROM TableEvil" FROM "TableA" WHERE pk = 1

The real executor runs this against TableA, since the quoted text is just a projected attribute name. But the regex's first match is the literal "FROM" sitting inside those quotes, so it resolves the IAM-checked resource to TableEvil instead. A caller authorized only for TableEvil, or explicitly denied on TableA, would pass the IAM check and then actually read or write TableA. That hits ExecuteStatement, BatchExecuteStatement, and ExecuteTransaction alike, since they all go through this same helper. Since the real tokenizer already exists in this codebase, I'd reuse it (or at least its tokenizer) for the IAM-side table extraction rather than keeping a second regex-based parser of the same grammar around, that's exactly how these two are going to keep drifting apart.

Small and non-blocking: in readJsonBody, the one path where in.readAllBytes() itself throws IOException returns null without resetting the entity stream, every other exit path in that method does reset it first. Worth a one-line fix for consistency, but low risk since the body is normally already buffered before this filter runs.

vuthehuyht:
Thanks for the sharp catch and review @pgermosen!
I've pushed a commit addressing both points:
1. Replaced the regex with DynamoDbPartiQLParser.extractTable, reusing the actual AST tokenizer/parser to ensure quoted identifiers containing SQL keywords (like SELECT "FROM TableEvil" FROM "TableA")  resolve strictly to the structural target table (TableA). Added test coverage for this case.
2. Ensured ctx.setEntityStream(...) is consistently restored across all exit paths in readJsonBody.

pgermosen (approved):
This is closed, and closed at the right level, not just patched for the one repro. Reusing the actual parser rather than a second regex-based implementation means IAM extraction and real execution can no longer disagree about the target table for anything that parses successfully, that's the structural fix, not just a fix for the specific case I found. Checked the fallback path too, when parse() itself throws, and it's sound: a parse failure there means the same statement fails in the real execution path too, so nothing ever actually runs against whatever the fallback guessed.

Entity-stream restore on the IOException path is fixed too. Nice work closing both cleanly.

PR #2931, opened 2026-09-02
fix(dynamodb): reject ConsistentRead on GSI Scan

Ketlark:
## Summary

Scan with `ConsistentRead: true` on a global secondary index silently performed the read instead of being rejected. Query already rejected this case; this PR extracts that check into `rejectConsistentReadOnGsi` (now driven by `DynamoDbAccessPath.isGlobalSecondaryIndex()`) and applies it to both handlers. Strongly consistent reads on tables and LSIs remain legal, matching AWS.

Discovered while scoring emulators with the independent [paritysuite/dynamodb-conformance](https://github.com/paritysuite/dynamodb-conformance) suite — `tests/tier3/error-messages/indexReads.test.ts` fails on 2.0.1 for Scan and passes for Query.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

**Incorrect behavior:** `Scan` + `IndexName` (GSI) + `ConsistentRead: true` returned 200 and executed the scan. Real AWS returns:

```
ValidationException: Consistent reads are not supported on global secondary indexes
```

**Verification:** the paritysuite conformance case (asserting the exact message above) fails against `floci/floci:2.0.1` and passes with this change — full suite moves 797/206 → 798/205 passed. In-repo wire-level test added in `DynamoDbAccessPathIntegrationTest#scanGsiRejectsConsistentReads` (rejection + LSI-acceptance guard), and the Java SDK compat twin `DynamoDbAccessPathValidationTest#gsiStillRejectsConsistentReads` now covers both operations.

Follow-up noticed while here (not in scope): `ExecuteStatement` (PartiQL) has no ConsistentRead-on-GSI check either.

## Checklist

- [x] `./mvnw test` passes locally (dynamodb package: 557 tests green)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbAccessPathValidationTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbAccessPathValidationTest.java
@@ -193,6 +193,10 @@ class DynamoDbAccessPathValidationTest {
                 .keyConditionExpression("#status = :status")
                 .expressionAttributeNames(Map.of("#status", "status"))
                 .expressionAttributeValues(Map.of(":status", value("open")))));
+        assertValidationException(() -> ddb.scan(request -> request
+                .tableName(TABLE)
+                .indexName("status-index")
+                .consistentRead(true)));
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Clean fix, and a nice refactor too, reusing the access path that's already resolved for validateSelection rather than doing a second GSI lookup. Traced the Kind classification and isGlobalSecondaryIndex() is exactly equivalent to the old inline check, base table and LSI consistent reads correctly stay legal since only GLOBAL_SECONDARY_INDEX trips it.

Good test coverage too, the new test covers both the rejection and the LSI-still-legal case together, same shape as the existing Query test, plus the SDK compat twin. Appreciate you flagging the ExecuteStatement gap explicitly rather than just leaving it for someone to find later.

Ketlark:
Thanks for the quick review and merge!

PR #2932, opened 2026-09-02
fix(lambda): honour Qualifier on GetFunction and GetFunctionConfiguration

### LINKED ISSUE #2821: [BUG] Lambda GetFunction and Invoke ignore Qualifier, returning $LATEST for any version including ones that do not exist
## Service

Lambda (versions)

## AWS API Action

`GetFunction` / `GetFunctionConfiguration` / `Invoke` — the `Qualifier` parameter

## Expected behavior

`Qualifier` selects a published version (or alias). A qualifier that does not resolve is a 404.

Measured against the live service (`lambda.ap-southeast-1.amazonaws.com`), on a function with `$LATEST`, `1` and `2` where `$LATEST` has moved on past version 1:

| Call | Live AWS |
| --- | --- |
| `get-function --qualifier 1` → `Version` | `1` |
| `get-function --qualifier $LATEST` → `Version` | `$LATEST` |
| `get-function-configuration --qualifier 1` → `Version` | `1` |
| `get-function --qualifier 99` | 404 `ResourceNotFoundException` — `Function not found: arn:aws:lambda:<region>:<account>:function:<name>:99` |
| `get-function --qualifier notaversion` | 404 `ResourceNotFoundException`, same shape |
| `invoke --qualifier 1` | `ExecutedVersion: 1`, returns version 1's code |
| `invoke --qualifier 2` | `ExecutedVersion: 2`, returns version 2's code |
| `invoke --qualifier 99` | 404 `ResourceNotFoundException` |

A published version is immutable — after `$LATEST` is updated, `--qualifier 1` still runs the code version 1 was published from.

## Actual behavior

Floci ignores the qualifier and answers with `$LATEST` in every case, including for versions that do not exist.

```
list-versions-by-function                  -> $LATEST  1     (version 1 does exist)
get-function --qualifier 1  -> Version     -> $LATEST         (AWS: 1)
get-function --qualifier 99 -> Version     -> $LATEST         (AWS: 404)
invoke --qualifier 99                      -> 200, ExecutedVersion $LATEST   (AWS: 404)
```

Two consequences. A caller pinning a published version silently gets whatever `$LATEST` currently holds, which defeats the point of publishing — the guarantee is immutability. And an invoke against a version that was never published succeeds instead of failing, so a typo or a stale alias runs the wrong code rather than erroring.

## Reproduction

```bash

mk() { rm -rf wp && mkdir wp && printf "def handler(e, c):\n    return {'body': '%s'}\n" "$1" > wp/handler.py && (cd wp && zip -q ../wp.zip handler.py); }

mk v1
awsl lambda create-function --function-name ver-probe --runtime python3.13 \
  --role arn:aws:iam::000000000000:role/r --handler handler.handler --zip-file fileb://wp.zip
awsl lambda publish-version --function-name ver-probe --query Version --output text   # -> 1

mk v2
awsl lambda update-function-code --function-name ver-probe --zip-file fileb://wp.zip

awsl lambda get-function --function-name ver-probe --qualifier 1 \
  --query 'Configuration.Version' --output text
# $LATEST     <-- AWS returns 1

awsl lambda get-function --function-name ver-probe --qualifier 99 \
  --query 'Configuration.Version' --output text
# $LATEST     <-- AWS returns 404 ResourceNotFoundException

awsl lambda invoke --function-name ver-probe --qualifier 1 --payload '{}' out.json
cat out.json
# {"body": "v2"}   <-- version 1 should still be v1
```

## Environment

## Note

#1988 reports the invoke half of this as a warm-pool keying problem. It reproduces, but the cause looks broader than the pool: `GetFunction` resolves the qualifier to `$LATEST` too, so the wrong target is chosen before a container is ever selected. Happy to fold #1988 into a fix here if a maintainer prefers.

Related divergences from the same measurement pass, filed separately: #2819 (`DeleteFunction` ignores `Qualifier` and deletes the whole function) and #2820 (`Publish` ignored by `CreateFunction`/`UpdateFunctionCode`). #2820 is worth fixing first — with `Publish` ignored, most callers have no published versions for a qualifier to select.

avison9:
## Summary

Closes #2821.

`GetFunction` and `GetFunctionConfiguration` ignored `Qualifier` and answered with `$LATEST` for
everything, including versions that were never published. Two consequences, both in the issue's
measurements against the live service: a caller pinning version 1 silently got whatever `$LATEST`
held at the time, which defeats the immutability guarantee that is the point of publishing; and a
read against a nonexistent version returned 200 instead of 404, so a typo or a stale alias read back
as a live function.

## Scope, which is narrower than the issue as filed

The issue also covers `Invoke`, and reports it against `main` at `26ae4ef7`. **That half is already
fixed.** `resolveInvokeTarget` on current `main` handles `$LATEST`, numeric versions with a proper
404, and aliases. Only the read paths were still ignoring the qualifier, so that is what this
changes.

The remaining invoke-side gap is #1988, the warm pool keyed by function name only. I have
deliberately left that alone: PR #2028 fixes it, @pgermosen reviewed it, verified the design end to
end and called the fix solid, and closed it only for merge conflicts with a request to rebase. That
is @aniketshukla1's work awaiting a rebase, not free ground, and reimplementing it here would
duplicate it.

## What changed

The read paths delegate to the same resolver the invoke path already uses, so a qualifier means the
same thing whether a caller reads it or runs it, rather than growing a second resolution rule that
could drift:

```java
public LambdaFunction getFunction(String region, String functionName, String qualifier) {
    LambdaArnUtils.ResolvedFunctionRef ref = LambdaArnUtils.resolve(functionName);
    enforceRegion(region, ref);
    String effective = resolveQualifier(ref.qualifier(), qualifier);
    if (effective == null) {
        return getFunction(region, functionName);
    }
    ...
    return resolveInvokeTarget(region, ref.name(), effective);
}
```

The existing no-qualifier `getFunction` is untouched, so the many internal callers are unaffected.

One case the issue does not raise but AWS supports: a qualifier can ride on the name itself, as
`fn:1` or a qualified ARN. An explicit `Qualifier` and one carried on the name must agree, and
disagreeing is `InvalidParameterValueException` rather than a silent pick of one.

## AWS Compatibility

Matches the table in the issue: `--qualifier 1` reports `Version: 1`, `--qualifier $LATEST` and no
qualifier both report `$LATEST`, and `--qualifier 99` or a name that is not an alias is a 404
`ResourceNotFoundException`.

## How it was tested

New `LambdaQualifiedReadIntegrationTest`, built from the reporter's measured table:

| Test | Covers |
|---|---|
| `aQualifierSelectsThePublishedVersionRatherThanLatest` | Version 1 stays version 1 after `$LATEST` moves on, on both endpoints, and unqualified plus explicit `$LATEST` still mean `$LATEST` |
| `aQualifierThatDoesNotResolveIsNotFound` | `99` and a non-alias name are 404 on both endpoints |
| `aQualifierMayBeCarriedOnTheNameAndMustAgreeWithAnExplicitOne` | `fn:1` resolves, agreeing is fine, disagreeing is a 400 |

Reverting only the controller change makes all three fail, so they are real regression tests.

Full suite: `Tests run: 15055, Failures: 0, Errors: 1, Skipped: 9`.

`java.sql.SQLException: No suitable driver found for jdbc:h2:mem:...`. It is **not from this
change**: I ran clean `main` with my work stashed and it fails there identically, and it passes when
that class runs on its own, so it looks like a driver registration race under the full parallel
suite. Happy to file it separately if it is not already tracked.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -88,9 +88,10 @@ public class LambdaController {
     @Path("/functions/{functionName}")
     public Response getFunction(@Context HttpHeaders headers,
                                 @Context UriInfo uriInfo,
-                                @PathParam("functionName") String functionName) {
+                                @PathParam("functionName") String functionName,
+                                @QueryParam("Qualifier") String qualifier) {
         String region = regionResolver.resolveRegion(headers);
-        LambdaFunction fn = lambdaService.getFunction(region, functionName);
+        LambdaFunction fn = lambdaService.getFunction(region, functionName, qualifier);
 
         ObjectNode root = objectMapper.createObjectNode();
         root.set("Configuration", objectMapper.valueToTree(buildFunctionConfiguration(fn)));
@@ -145,9 +146,10 @@ public class LambdaController {
     @GET
     @Path("/functions/{functionName}/configuration")
     public Response getFunctionConfiguration(@Context HttpHeaders headers,
-                                              @PathParam("functionName") String functionName) {
+                                              @PathParam("functionName") String functionName,
+                                              @QueryParam("Qualifier") String qualifier) {
         String region = regionResolver.resolveRegion(headers);
-        LambdaFunction fn = lambdaService.getFunction(region, functionName);
+        LambdaFunction fn = lambdaService.getFunction(region, functionName, qualifier);
         return Response.ok(buildFunctionConfiguration(fn)).build();
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -476,6 +476,50 @@ public class LambdaService implements ResourceProvider {
     }
 
     /**
+     * Reads a function, honouring a {@code Qualifier} that selects a published version or an alias.
+     *
+     * <p>The read paths used to ignore the qualifier entirely and answer with {@code $LATEST} for
+     * everything, including versions that were never published (issue #2821). That defeats the
+     * point of publishing: a caller pinning version 1 silently got whatever {@code $LATEST} held at
+     * the time of the call, and a typo or a stale alias read back as a live function instead of
+     * failing. Resolution is delegated to the same target resolver the invoke path uses, so a
+     * qualifier means the same thing whether you read it or run it.
+     *
+     * <p>A qualifier may also be carried on the name itself, as {@code fn:1} or a qualified ARN. An
+     * explicit {@code Qualifier} and one embedded in the name must agree; AWS rejects the
+     * combination when they do not.
+     */
+    public LambdaFunction getFunction(String region, String functionName, String qualifier) {
+        LambdaArnUtils.ResolvedFunctionRef ref = LambdaArnUtils.resolve(functionName);
+        enforceRegion(region, ref);
+        String effective = resolveQualifier(ref.qualifier(), qualifier);
+        if (effective == null) {
+            return getFunction(region, functionName);
+        }
+        if (functionName.startsWith("arn:")) {
+            AwsArnUtils.Arn arn = AwsArnUtils.parse(functionName);
+            return resolveInvokeTargetForAccount(arn.accountId(), region, ref.name(), effective);
+        }
+        return resolveInvokeTarget(region, ref.name(), effective);
+    }
+
+    /**
+     * Reconciles a qualifier carried on the function name with an explicit {@code Qualifier}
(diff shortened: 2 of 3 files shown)


pgermosen (approved):
Nice fix, and the alias fix in particular checks out cleanly against the docs, GetAlias's FunctionVersion is a single fixed field separate from RoutingConfig, so reads reporting the alias's primary version deterministically while Invoke keeps the weighted random pick is exactly right. The 20-iteration test on a 50/50 split is the right way to pin that, a regression there would show up almost immediately.

Also traced the name-vs-explicit qualifier conflict handling and confirmed the existing no-qualifier overload is untouched, so this doesn't put the many internal callers at risk.

PR #2933, opened 2026-09-02
fix(lambda): honour CodeSha256 and skip unchanged publishes on PublishVersion

### LINKED ISSUE #2822: [BUG] Lambda PublishVersion always creates a version and ignores the CodeSha256 precondition

avison9:
## Summary

Closes #2822.

Two documented `PublishVersion` behaviours were missing. `CodeSha256` was ignored, so a publish that
should have been rejected succeeded and produced a version. And publishing was unconditional, so
repeated publishes accumulated identical versions.

## What changed

**`CodeSha256` as a precondition.** The controller parsed only `Description` out of the body, so the
field never reached the service at all. A mismatch is now `InvalidParameterValueException` carrying
AWS's message shape, and nothing is published.

**Unchanged publishes return the existing version.** Rather than comparing the thirty-odd fields a
snapshot copies, this keys on a signal the codebase already maintains: `$LATEST`'s `revisionId` is
regenerated on every code and configuration update. A version records the revision it was cut from,
and a later publish that finds it unmoved returns that version.

```java
LambdaFunction unchanged = latestPublishedFrom(region, fn, description);
if (unchanged != null) {
    return unchanged;
}
```

Keying on the revision rather than a field by field comparison means a field added to versions later
is covered without anyone remembering to update a comparison. Versions published before this change
carry no source revision and never match, so existing state simply keeps its previous behaviour
rather than deduplicating against a revision nobody recorded.

`LambdaFunction` gains one field. That is safe for persisted state because the model is
`@JsonIgnoreProperties(ignoreUnknown = true)`, so older records load with it null.

## Description, and why it is part of the check

`Description` is supplied at publish time rather than carried on `$LATEST`, so a publish naming a new
one is a real publish even when nothing else moved.

I had this as an open question, since AWS's wording is "configuration and code", and I have no live
account to measure it against. **This repository already answers it**:
`LambdaVersionIntegrationTest.publishSecondVersion` publishes with a new `Description` and nothing
else changed, and asserts it gets version 2. So the effective description is part of the check, and
that existing test is the reason.

## One existing test changed, which I want visible rather than buried

`LambdaServicePersistenceTest.publishVersionContinuesNumberingAfterRestart` published three times
with nothing changed at all in order to reach version 3. No correct de-duplication can leave that
passing, because it encodes precisely the behaviour this issue asks to change.

Its subject is the version counter surviving a restart, not whether unchanged publishes mint
versions, so it now varies the description across the three calls. That still exercises the counter
across a reload, including the reload boundary, which is what the test is named for.

If maintainers read that test's intent differently, say so and I will adjust rather than defend it.

## AWS Compatibility

Matches the measurements in the issue: a repeat publish with nothing changed returns the existing
version instead of minting `2`, `3`, `4`; and `--code-sha256` with a value that does not match
`$LATEST` returns `InvalidParameterValueException` reading
`CodeSHA256 (...) is different from current CodeSHA256 in $LATEST`.

Out of scope, as the issue itself notes it is "not the substance of this report": the unordered
`list-versions-by-function` result. Happy to take that separately.

## How it was tested

New `LambdaPublishVersionSemanticsIntegrationTest`:

| Test | Covers | Fails without the fix |
|---|---|---|
| `publishingTwiceWithNothingChangedReturnsTheExistingVersion` | Three unchanged publishes return one version, and the version list does not grow | Yes, `expected: <1> but was: <2>` |
| `aMismatchedCodeSha256IsRejectedAndPublishesNothing` | 400, and no version created by the rejected call | Yes |
| `aRealChangeStillProducesANewVersion` | A code change and a configuration-only change each still mint a version | No, a do-no-harm guard on the de-duplication |
| `aMatchingCodeSha256Publishes` | The precondition does not block a correct hash | No, a do-no-harm guard |

Full suite: `Tests run: 15056, Failures: 0, Errors: 1, Skipped: 9`.

`java.sql.SQLException: No suitable driver found for jdbc:h2:mem:...`. It is not from this change: I
ran clean `main` with my work stashed and it fails there identically, and it passes when that class
runs alone, so it looks like a driver registration race under the full parallel suite.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -393,14 +393,16 @@ public class LambdaController {
                                    String body) {
         String region = regionResolver.resolveRegion(headers);
         String description = null;
+        String codeSha256 = null;
         if (body != null && !body.isBlank()) {
             try {
                 @SuppressWarnings("unchecked")
                 Map<String, Object> req = objectMapper.readValue(body, Map.class);
                 description = (String) req.get("Description");
+                codeSha256 = (String) req.get("CodeSha256");
             } catch (Exception ignored) {}
         }
-        LambdaFunction version = lambdaService.publishVersion(region, functionName, description);
+        LambdaFunction version = lambdaService.publishVersion(region, functionName, description, codeSha256);
         return Response.status(201).entity(buildFunctionConfiguration(version)).build();
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -44,6 +44,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -1278,6 +1279,29 @@ public class LambdaService implements ResourceProvider {
      * {@code concurrencyOpLocks}) so publishes of unrelated functions do not
      * serialize on the instance monitor for the storage round-trip.
      */
+    /**
+     * The already-published version cut from {@code $LATEST}'s current revision, if there is one.
+     *
+     * <p>Versions published before {@code sourceRevisionId} was recorded carry null and never
+     * match, so existing state simply keeps the previous behaviour rather than deduplicating
+     * against a revision nobody wrote down.
+     */
+    private LambdaFunction latestPublishedFrom(String region, LambdaFunction fn, String description) {
+        String current = fn.getRevisionId();
+        if (current == null) {
+            return null;
+        }
+        // Description is supplied at publish time rather than carried on $LATEST, so a publish that
+        // names a new one is a real publish even when nothing else moved. LambdaVersionIntegrationTest
+        // asserts exactly that: two publishes differing only in Description produce versions 1 and 2.
+        String effective = description != null ? description : fn.getDescription();
+        return functionStore.listVersions(region, fn.getFunctionName()).stream()
+                .filter(v -> current.equals(v.getSourceRevisionId()))
+                .filter(v -> Objects.equals(effective, v.getDescription()))
+                .findFirst()
+                .orElse(null);
+    }
+
     private int nextVersionNumber(String counterKey, String legacyCounterKey) {
         synchronized (versionCounterLocks.computeIfAbsent(counterKey, k -> new Object())) {
             Integer current = versionCounters.get(counterKey);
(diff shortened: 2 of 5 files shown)


avison9:
Closing in favour of a split, after the compatibility suite showed this needed one.

`CodeDeployTest` publishes twice with nothing changed and asserts the versions differ, so the de-duplication half turns that job red. The two behaviours in this PR are independent, and bundling them meant a maintainer had to take both or neither, with the safe one held up by the arguable one.

- #2939 carries the `CodeSha256` precondition. Additive, opt in, no existing test changes.
- The de-duplication half, with the compatibility suite change it forces, is opened separately so the decision is visible on its own.

Greptile's three findings here are all addressed in the split branches: the hash check moved inside the lock `UpdateFunctionCode` takes, the comparison now considers only the most recent version so an A then B then A description publishes again, and hot-reload functions are excluded because their code changes without touching `revisionId`.

GitHub will not let a PR change its head branch, hence a new one rather than an edit here.

PR #2935, opened 2026-09-02
fix(dynamodb): reject QueryFilter with KeyConditionExpression

### LINKED ISSUE #2894: [BUG] DynamoDB: Query accepts legacy QueryFilter alongside KeyConditionExpression
## Service

DynamoDB

## AWS API Action

`Query` (legacy `QueryFilter` alongside `KeyConditionExpression`)

## Expected behavior

Real AWS refuses to mix a legacy non-expression parameter with any expression parameter in the same request, not just with its direct expression counterpart. `QueryFilter` paired with `KeyConditionExpression` is rejected. Measured against the live service in `ap-southeast-1` with a read-only `Query`:

```
ValidationException: Can not use both expression and non-expression parameters in the same request:
Non-expression parameters: {QueryFilter} Expression parameters: {KeyConditionExpression}
```

## Actual behavior

Floci accepts the request and runs the query.

```
$ curl -s -X POST http://localhost:4566/ \
    -H 'X-Amz-Target: DynamoDB_20120810.Query' \
    -H 'Content-Type: application/x-amz-json-1.0' \
    -d '{"TableName":"ProbeTbl","KeyConditionExpression":"PK = :p",
         "ExpressionAttributeValues":{":p":{"S":"zz"}},
         "QueryFilter":{"n":{"ComparisonOperator":"EQ","AttributeValueList":[{"S":"x"}]}}}'
{"Items":[],"Count":0,"ScannedCount":0}
```

`handleQuery` in `DynamoDbJsonHandler` guards three pairs (lines 824 to 838):

- `KeyConditions` with `KeyConditionExpression`
- `QueryFilter` with `FilterExpression`
- `AttributesToGet` with `ProjectionExpression`

`QueryFilter` with `KeyConditionExpression` is not among them, so a request carrying `QueryFilter` and `KeyConditionExpression` but no `FilterExpression` passes every guard. AWS treats the two parameter families as mutually exclusive as families, not pairwise.

The error string Floci already emits for the `ScanFilter` case is byte-identical to the live service, so the fix is an additional guard reusing the existing wording rather than new message text.

## Reproduction

```bash
docker run -d --name floci -p 4566:4566 floci/floci:main
aws dynamodb create-table --endpoint-url http://localhost:4566 --table-name ProbeTbl \
  --attribute-definitions AttributeName=PK,AttributeType=S AttributeName=SK,AttributeType=S \

# Floci returns results, AWS returns ValidationException
aws dynamodb query --endpoint-url http://localhost:4566 --table-name ProbeTbl \
  --key-condition-expression 'PK = :p' \
  --expression-attribute-values '{":p":{"S":"zz"}}' \
  --query-filter '{"n":{"ComparisonOperator":"EQ","AttributeValueList":[{"S":"x"}]}}'
```

## A second observation I could not explain

While characterising this I also measured `Scan` with both `ProjectionExpression` and `AttributesToGet`:

```
live AWS   ValidationException: Can not use both expression and non-expression parameters in the same request:
           Non-expression parameters: {AttributesToGet} Expression parameters: {ProjectionExpression}

floci      HTTP 200  {"Items":[{"PK":{"S":"rw-ok"}}],"Count":1,"ScannedCount":1}
```

`handleScan` appears to guard exactly this at line 1004, unconditionally, and the neighbouring `ScanFilter` with `FilterExpression` guard at line 970 does fire on the same request path (HTTP 400). I could not work out why the later guard does not trigger, and I would rather report the measurement than guess at a cause. Both calls were made against the same running image with the raw JSON protocol.

## Context

Found while running the SDK compatibility suite against the container-backed backend in #2871. `amazon/dynamodb-local` rejects the `QueryFilter` case, Floci accepts it, and the live service agrees with the container. The conformance case is `QueryFilter (legacy) filters query results` in `compatibility-tests/sdk-test-node/tests/dynamodb-conformance.test.ts`, which asserts a successful result and would need updating alongside any fix.

Related but distinct from the unused `ExpressionAttributeNames` gap filed separately. I have deliberately not changed this in #2871, which adds a backend without touching native engine semantics.

## Environment

- Floci version / image tag: built from `main` at `2bf9b57`, reports `2.0.1`
- How you're running Floci: Docker, `docker build -f docker/Dockerfile`
- Live AWS comparison: `ap-southeast-1`, AWS CLI v2

aniketshukla1:
## Summary

- reject Query requests that combine the legacy `QueryFilter` parameter with `KeyConditionExpression`
- preserve valid legacy `QueryFilter` coverage by pairing it with legacy `KeyConditions`
- add Java protocol and Node SDK regression coverage for the exact AWS validation response

Closes #2894

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, Floci accepted `QueryFilter` alongside `KeyConditionExpression` and executed the query. AWS rejects this mixed legacy/expression request with HTTP 400 `ValidationException`.

This change rejects the request before table lookup with the AWS-compatible message:

`Can not use both expression and non-expression parameters in the same request: Non-expression parameters: {QueryFilter} Expression parameters: {KeyConditionExpression}`

Verified locally:

- `./mvnw test -Dtest=DynamoDbIntegrationTest,DynamoDbJsonHandlerTest` (73 tests passed)
- `npm test -- tests/dynamodb-conformance.test.ts` (42 tests passed)
- focused TypeScript compilation of `dynamodb-conformance.test.ts`

## Checklist

- [ ] `./mvnw test` passes locally (scoped DynamoDB suites pass as listed above)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbConformanceChangesTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbConformanceChangesTest.java
@@ -547,8 +547,10 @@ class DynamoDbConformanceChangesTest {
     void queryWithQueryFilter() {
         QueryResponse resp = ddb.query(QueryRequest.builder()
                 .tableName(TABLE)
-                .keyConditionExpression("pk = :pk")
-                .expressionAttributeValues(Map.of(":pk", av("p1")))
+                .keyConditions(Map.of("pk", Condition.builder()
+                        .comparisonOperator(ComparisonOperator.EQ)
+                        .attributeValueList(av("p1"))
+                        .build()))
                 .queryFilter(Map.of("name", Condition.builder()
                         .comparisonOperator(ComparisonOperator.EQ)
                         .attributeValueList(av("Item-0"))
--- a/compatibility-tests/sdk-test-node/tests/dynamodb-conformance.test.ts
+++ b/compatibility-tests/sdk-test-node/tests/dynamodb-conformance.test.ts
@@ -707,14 +707,32 @@ describe('DynamoDB — Phase 10: Legacy API', () => {
   it('QueryFilter (legacy) filters query results', async () => {
     const resp = await ddb.send(new QueryCommand({
       TableName: table,
-      KeyConditionExpression: 'pk = :pk',
-      ExpressionAttributeValues: { ':pk': s('p1') },
+      KeyConditions: {
+        pk: { ComparisonOperator: 'EQ', AttributeValueList: [s('p1')] },
+      },
       QueryFilter: {
         name: { ComparisonOperator: 'EQ', AttributeValueList: [s('Item-0')] },
       },
     }));
     expect(resp.Count).toBe(1);
     expect(resp.Items![0].name.S).toBe('Item-0');
+  });
+
+  it('QueryFilter and KeyConditionExpression are mutually exclusive', async () => {
+    const err = await ddb.send(new QueryCommand({
+      TableName: table,
+      KeyConditionExpression: 'pk = :pk',
+      ExpressionAttributeValues: { ':pk': s('p1') },
+      QueryFilter: {
+        name: { ComparisonOperator: 'EQ', AttributeValueList: [s('Item-0')] },
+      },
+    })).catch(e => e);
+
+    expect(err.name).toBe('ValidationException');
+    expect(err.message).toBe(
+      'Can not use both expression and non-expression parameters in the same request: '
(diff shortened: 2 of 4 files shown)


pgermosen (approved):
Clean fix, and it fits right into the existing pattern, the message template matches the two sibling checks already in handleQuery for the analogous KeyConditions/KeyConditionExpression and FilterExpression/QueryFilter conflicts, so this isn't a new format, just closing the one gap. Good catch updating the existing queryWithQueryFilter test in both compat suites too, since it used to combine KeyConditionExpression with QueryFilter, which is exactly what's now rejected, so it needed to move to the legacy KeyConditions form to keep testing what it was meant to test.

PR #2936, opened 2026-09-02
fix(cognito): name missing user pools in errors

### LINKED ISSUE #2853: [BUG] cognito: user pool not-found message is "User pool not found", AWS names the pool

aniketshukla1:
## Summary

- include the requested user pool ID in Cognito `ResourceNotFoundException` messages
- centralize missing-pool errors so shared and direct lookup paths stay consistent
- cover the exact JSON response plus direct `AdminGetUser` and `ConfirmSignUp` service paths

Closes #2853

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Missing pools previously returned the generic message `User pool not found`. Cognito names the requested pool: `User pool <poolId> does not exist.` The error remains `ResourceNotFoundException` with HTTP 400.

The integration test asserts the exact JSON `__type` and `message` returned by `DescribeUserPool`. Service tests also pin the direct-store paths used by `AdminGetUser` and `ConfirmSignUp`.

## Testing

- `CognitoServiceTest,CognitoIntegrationTest`: 243 tests, 0 failures, 0 errors
- `git diff --check`

## Checklist

### DIFF
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -17,6 +17,8 @@ Floci strips reserved `floci:*` tags from stored and returned `UserPoolTags` on 
 
 Standalone `TagResource` rejects reserved `floci:*` keys. `ListTagsForResource` and `UntagResource` operate on the persisted user-pool tag map.
 
+An action given a user pool ID that does not resolve returns `ResourceNotFoundException` with the live service's wording, `User pool <poolId> does not exist.`, so tooling that matches Cognito error text behaves the same way locally.
+
 ## Supported Actions
 
 ### User Pools
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -350,7 +350,7 @@ public class CognitoService implements ResourceProvider {
 
     public UserPool describeUserPool(String id) {
         UserPool pool = poolStore.get(id)
-                .orElseThrow(() -> new AwsException("ResourceNotFoundException", "User pool not found", 400));
+                .orElseThrow(() -> userPoolNotFound(id));
         boolean generatedKeys = ensureJwtSigningKeys(pool);
         boolean generatedSecret = ensureRefreshTokenSecret(pool);
         if (generatedKeys || generatedSecret) {
@@ -1144,7 +1144,7 @@ public class CognitoService implements ResourceProvider {
 
     public CognitoUser adminGetUser(String userPoolId, String username) {
         UserPool pool = poolStore.get(userPoolId).orElseThrow(
-                () -> new AwsException("ResourceNotFoundException", "User pool not found", 400));
+                () -> userPoolNotFound(userPoolId));
         LinkedHashMap<String, CognitoUser> matches = new LinkedHashMap<>();
         userStore.get(userKey(userPoolId, username))
                 .ifPresent(u -> matches.put(u.getUsername(), u));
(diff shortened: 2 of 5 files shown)


bsaurusrex:
I had a branch for #2853 in flight and opened #2951 before spotting this one. Yours is first, so I have closed mine. Two things from it that this PR may want, plus the measurements behind them.

## A stale assertion this PR does not update

`S3WellKnownKeyIntegrationTest` uses the old message as a fingerprint for a request being misrouted to Cognito instead of S3:

```java
if (body.contains("ResourceNotFoundException") || body.contains("User pool not found")) {
    throw new AssertionError("... was routed to Cognito instead of S3. Response: " + body);
}
```

After this change that second string can never appear, so half the guard quietly stops guarding. It still passes, which is why CI will not flag it. Matching on `"User pool "` keeps it fingerprinting the same thing and survives the reword.

## Docs

`docs/services/cognito.md` describes the service's error behaviour but has nothing about this. A sentence near the top, after the `floci:*` tag section, is where I put it:

> An action given a user pool id that does not resolve answers `ResourceNotFoundException` with the live service's own wording, `User pool <poolId> does not exist.`, rather than a generic string, so tooling that matches on Cognito error text behaves the same way locally.

## Measurements, if useful for the description

Live service, `ap-southeast-1`, AWS CLI v2, on a pool id that does not exist. Five actions, all identical, which confirms centralising the string is the right shape:

```
describe-user-pool      -> ResourceNotFoundException: User pool ap-southeast-1_ZZZZZZZZZ does not exist.
```

The trailing period is part of the live string.

A malformed id is a different path and is rejected before anything is resolved, so it is worth being explicit that this change does not cover it:

```
describe-user-pool --user-pool-id not-a-pool-id
  -> InvalidParameterException
     1 validation error detected: Value 'not-a-pool-id' at 'userPoolId' failed to satisfy
     constraint: Member must satisfy regular expression pattern: [\w-]+_[0-9a-zA-Z]+
```

Floci does not validate that pattern, so a malformed id falls through to the not-found path and will now report `User pool not-a-pool-id does not exist.` where AWS reports a validation error. Pre-existing, and a separate piece of work.

## One thing that surprised me

`ListUsers` and `ListUserPoolClients` never reach this message at all: they answer an absent pool with `200` and an empty list rather than resolving it, unlike `ListGroups` and `ListResourceServers`. Filed as #2952, out of scope here.

Happy to send any of the above as a PR against your branch if that is easier than folding it in.

pgermosen (commented):
Good fix, and it matches bsaurusrex's live measurement exactly, User pool <poolId> does not exist. with the trailing period, wired consistently into all three paths. confirmSignUp resolving through the client's pool id and naming that one is the right call too.

Two small things from bsaurusrex's comment that are still open. S3WellKnownKeyIntegrationTest fingerprints a Cognito misroute partly on the literal string "User pool not found", which can't appear again after this change, so that half of the check is now dead code, still passing today only because the other half (the exception type name) still catches it. Worth the one-line fix to match on "User pool " instead so it survives the reword.

And docs/services/cognito.md isn't updated for this, worth adding the note bsaurusrex already drafted in their comment.

pgermosen (approved):
Approving. Both open items are genuinely fixed - the S3 fingerprint now matches "User pool " so it survives the reword, and the docs sentence is in. Re-checked the three call sites too, they all route through the one userPoolNotFound helper now and it matches bsaurusrex's live measurement exactly, trailing period included.

PR #2938, opened 2026-09-02
feat(ecs): honor UpdateService forceNewDeployment and emit EventBridge lifecycle events

### LINKED ISSUE #2557: [BUG] ECS UpdateService forceNewDeployment is ignored; no ECS EventBridge events
## Service

ECS / EventBridge

## AWS API Action

- `UpdateService` with `forceNewDeployment=true` (same task definition)
- Implicit: publish `ECS Deployment State Change` and `ECS Task State Change` to the default EventBridge bus when deployments complete / tasks stop

## Expected behavior

On AWS:

1. `aws ecs update-service --force-new-deployment` starts a new rollout even when the task definition ARN is unchanged, and mints a new deployment id.
2. When a service reaches steady state, ECS emits `detail-type: ECS Deployment State Change` with `detail.eventName: SERVICE_DEPLOYMENT_COMPLETED` (resources = service ARN).
3. When a task stops, ECS emits `detail-type: ECS Task State Change` with `lastStatus`, `taskDefinitionArn`, container `exitCode`, etc.

## Actual behavior

1. `UpdateService` handler never reads `forceNewDeployment`. Same task definition + force flag → no redeploy, no new deployment id. Documented understatement in `docs/services/ecs.md`: “forceNewDeployment does not mint a new deployment id.”
2. ECS never calls EventBridge `PutEvents`. Rules matching `source: aws.ecs` never fire from Floci ECS lifecycle. `SubmitTaskStateChange` / `SubmitContainerStateChange` are agent ACK stubs only.

Note: the 5s service reconciler **does** replace exited containers when `runningCount < desiredCount`. The remaining gaps are force-new-deployment and EventBridge emission. DescribeServices already synthesizes a PRIMARY deployment for waiters (1.7.0).

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=test

# After a service exists on a cluster:
aws ecs update-service \
  --cluster my-cluster \
  --service my-service \
  --force-new-deployment \
  --endpoint-url "$AWS_ENDPOINT_URL"

# Expected: new deployment id / replacement tasks.
# Actual: no new deployment; running tasks unchanged.

# Separately: put an EventBridge rule for ECS Deployment State Change,
# force a real rollout (change task definition), then:
aws events list-rules --endpoint-url "$AWS_ENDPOINT_URL"
# No matching events are ever delivered to targets.
```

## Environment

- Floci version / image tag: 1.7.0
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker compose, AWS CLI against `http://localhost:4566`

vuthehuyht:
## Summary                                                                                                                                                                                                     
                                                                                                                                                                                                                   
- Adds support for `UpdateService` with `--force-new-deployment`, minting a new deployment ID and rolling existing service tasks even when the task definition remains unchanged.                              
- Implements `EcsEventPublisher` to emit AWS-shaped lifecycle events to the default EventBridge bus (`source: "aws.ecs"`):                                                                                     
- `ECS Task State Change` events synthesizing AWS phase ladders (`PROVISIONING -> PENDING -> ACTIVATING -> RUNNING` on start, `DEACTIVATING -> STOPPING -> DEPROVISIONING -> STOPPED` on stop).              
 - `ECS Deployment State Change` events (`SERVICE_DEPLOYMENT_STARTED`, `SERVICE_DEPLOYMENT_IN_PROGRESS`, `SERVICE_DEPLOYMENT_COMPLETED`).                                                                     
- Adds end-to-end integration tests verifying EventBridge rule routing for ECS events and updates service documentation.                                                                                       
                                                                                                                                                                                                                   Closes #2557                                                                                                                                                                                                   
                                                                                                                                                                                                                
## Type of change                                                                                                                                                                                              
                                                                                                                                                                                                               
- [x] New feature (`feat:`)                                                                                                                                                                                    
- [ ] Bug fix (`fix:`)
- [ ] Docs / chore  
    
## AWS Compatibility
  
- **`UpdateService --force-new-deployment`**: Aligned with AWS CLI / SDK behavior where requesting a forced deployment creates a fresh deployment identifier (`ecs-svc/<UUID>`) and triggers a rolling         
replacement of running tasks. Backward compatibility is preserved for legacy state via deterministic hash fallback.
- **EventBridge ECS Events (`aws.ecs`)**: Emits AWS-compatible event envelopes containing `clusterArn`, `taskArn`, `detail.lastStatus`, `desiredStatus`, `containers[].exitCode`, and deployment status        
transitions. Verified against AWS ECS EventBridge schema specifications.

## Checklist

### DIFF
--- a/docs/services/ecs.md
+++ b/docs/services/ecs.md
@@ -89,7 +89,10 @@ Known differences from AWS:
   `desiredCount` from that count; it is rejected for the Fargate launch type and for the
   `CODE_DEPLOY` / `EXTERNAL` controllers, as on AWS. Placement constraints are not evaluated.
 - `pendingCount` is always `0`, matching the top-level service field.
-- `forceNewDeployment` does not mint a new deployment `id`.
+- `forceNewDeployment` (with an unchanged task definition) mints a new deployment `id`
+  and rolls the running tasks: a replacement on the new deployment starts first, then
+  the task from the previous deployment is drained one reconciler tick later. The
+  `deployments` list still reports a single `PRIMARY` throughout.
 - `updatedAt` equals `createdAt`. AWS advances it as a rollout progresses; Floci has no
   intermediate rollout state to report.
 
--- a/docs/services/ecs.md
+++ b/docs/services/ecs.md
@@ -96,6 +96,31 @@ Known differences from AWS:
 - `updatedAt` equals `createdAt`. AWS advances it as a rollout progresses; Floci has no
   intermediate rollout state to report.
 
+#### ECS EventBridge events
+
+Floci publishes AWS-shaped lifecycle events to the **default** EventBridge bus
+(`source: aws.ecs`). Rules matching `aws.ecs` fire from ECS activity, in both docker
+and mock mode.
+
+| `detail-type` | When | Key `detail` fields |
+|---|---|---|
+| `ECS Task State Change` | a task starts or stops | `lastStatus`, `desiredStatus`, `taskDefinitionArn`, `group`, `startedBy`, `stoppedReason`, `containers[].exitCode` |
+| `ECS Deployment State Change` | a service deployment starts, is in progress, or reaches steady state | `eventType` (always `INFO`), `eventName`, `deploymentId` |
+
+`eventName` is one of `SERVICE_DEPLOYMENT_STARTED`, `SERVICE_DEPLOYMENT_IN_PROGRESS`,
+`SERVICE_DEPLOYMENT_COMPLETED`.
+
+Known differences from AWS:
+
+- The task phase ladder is **synthesized**. Floci's task model only occupies
+  `PENDING`, `RUNNING` and `STOPPED`, but a start emits
+  `PROVISIONING -> PENDING -> ACTIVATING -> RUNNING` and a stop emits
+  `DEACTIVATING -> STOPPING -> DEPROVISIONING -> STOPPED`, one `ECS Task State Change`
+  per phase, so rules that filter on `detail.lastStatus` behave as on AWS.
+- `SERVICE_DEPLOYMENT_FAILED` and the deployment circuit breaker are not emitted.
+- `SubmitTaskStateChange` / `SubmitContainerStateChange` remain ACK-only; Floci drives
+  the task lifecycle itself rather than via agent submissions.
+
 #### Unknown services
 
 A service reference that does not resolve is returned in `failures` with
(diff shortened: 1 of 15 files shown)


pgermosen (commented):
One more thing before anything else: the PR title needs to follow Conventional Commits (something like feat(ecs): ...), since it's what becomes the squashed commit message, and the description should go through the actual PR template rather than the freeform summary it has now. Both are required before this can land, not just the two items below.

The feature and event-shape work here is solid, but there's a real correctness gap worth fixing before merge: emitTaskLadder mutates the shared task object in place to synthesize the phase sequence, then restores it in a finally. I traced this against EcsService.java, tasks is a plain ConcurrentHashMap holding live object references, and it gets put into that map before emitTaskLadder is ever called, so the task is already visible to a concurrent DescribeTasks or ListTasks call while the ladder runs. Four synchronous phases, each doing a serialize plus an EventBridge put, is a real window, not an instant one. A caller polling at the wrong moment would see a lastStatus your own class Javadoc says this task model never actually occupies. Building a shallow copy for the synthetic events, instead of mutating the shared object, would close it.

On the ARN-parse finding, I don't think it's quite what greptile described, it's not an empty catch, it logs the exception and the task ARN. The real (minor) issue is that debugv is usually off by default, so the fallback is effectively invisible in practice even though the code does log it. Not blocking, just worth a bump to warn if you're touching that area anyway.

Everything else, the deploymentId legacy-fallback pattern, the in-progress dedup, and the integration test actually verifying EventBridge rule routing rather than just asserting the publisher was called, all look good.

pgermosen (approved):
Thanks for addressing everything from the last round. Title now reads clean Conventional Commits, and the description follows the template.

The shared-object mutation fix in `emitTaskLadder` is exactly right: building the one-time `EcsTask` snapshot and only ever varying its `lastStatus` through the ladder means `tasks.get(...)` can no longer observe a synthetic phase mid-launch or mid-stop. I checked the new copy constructor field-by-field against everything `emitTaskStateChange` actually reads, and it covers all of it, so there's no gap where the snapshot silently drops something the event serializer needs.

Also confirmed the ARN-parse-fallback logging is bumped to a level that'll actually show up by default now, so that fallback won't be silent anymore.

One small thing worth a follow-up at some point, not blocking here: there's no test that pins the actual invariant this fix establishes, something that would fail under the old implementation and pass under the new one (e.g. a concurrent reader observing the task mid-ladder). The existing assertion only checks that the status is restored once the method returns, which held true even before this fix. Not asking for it now, just flagging it as a nice one to add later.

Nothing else outstanding. This looks good to merge.

PR #2939, opened 2026-09-02
fix(lambda): honour the CodeSha256 precondition on PublishVersion

avison9:
## Summary

Part of #2822. This is the safe half, split out so it can land on its own; the unchanged-publish
half follows separately because it changes behaviour for existing callers.

`CodeSha256` is documented as a precondition on `PublishVersion`: "only publish a version if the hash
value matches the value that's specified". The controller parsed only `Description` out of the
request body, so the field never reached the service at all. A publish that should have been
rejected succeeded and produced a version, which is how a caller racing someone else's deploy ends
up publishing code it never authorised, with nothing to say so.

## What changed

`PublishVersion` gains a `CodeSha256` parameter, threaded from the controller. A mismatch is
`InvalidParameterValueException` carrying the message shape AWS uses, and no version is created.

The check is atomic with `UpdateFunctionCode`, but that required fixing the writer, not just the
reader. An earlier revision of this PR claimed the check was safe because it sat inside the lock
`UpdateFunctionCode` takes. **That was wrong**, and @pgermosen caught it in review: the code mutation
path, `extractZipCodeBytes`, sets `codeLocalPath`, `codeSizeBytes` and `codeSha256` with no lock
held, and the only `synchronized` block in there wraps an unrelated legacy-directory cleanup further
down. `PublishVersion` holding its own lock serialised nothing.

The writer now takes the same per-function lock:

```java
zipExtractor.extractTo(zipBytes, codePath);
String newSha256 = null;
try {
    byte[] digest = MessageDigest.getInstance("SHA-256").digest(zipBytes);
    newSha256 = Base64.getEncoder().encodeToString(digest);
} catch (NoSuchAlgorithmException ignored) {}
synchronized (lockForConcurrencyOp(fn.getFunctionArn())) {
    fn.setCodeLocalPath(codePath.toAbsolutePath().normalize().toString());
    fn.setCodeSizeBytes(zipBytes.length);
    if (newSha256 != null) {
        fn.setCodeSha256(newSha256);
    }
}
```

Hashing and extraction to disk stay outside the lock, so a large package does not serialise unrelated
work on the same function. Only the fields that decide what a snapshot copies are inside it, which is
the window between the hash being read and the snapshot being taken.

**Malformed preconditions are rejected rather than skipped.** Also from review: `{"CodeSha256": 12345}`
threw a `ClassCastException` into the pre-existing body-parse catch, leaving the field null and
indistinguishable from omitted, so the precondition was skipped and the publish succeeded with 201. An
unparseable body behaved the same way. Both are now 400, via a helper that rejects a wrong JSON type
instead of letting the cast fail silently.

## What this precondition does not cover

Raised by @pgermosen in review, and recorded here so it is not silently assumed to be handled.

There is no per-version code isolation in the codebase. `CodeStore.getCodePath` derives a path from
account and function name only, `extractZipCodeBytes` overwrites that same directory in place on
every `UpdateFunctionCode`, and `publishVersion`'s snapshot stores a reference to that path rather
than copying the code. So a published version does not actually freeze its code: a later
`UpdateFunctionCode`, racing or not, changes what invoking an already-published version runs.

In their words, this precondition "can't actually deliver what it implies once you look past the
moment of publish". What it does deliver is real and worth having: at the moment of publish, a
caller can require that `$LATEST` still holds the code it expects, and that check is now atomic
against a concurrent deploy. What it cannot do is keep the resulting version pinned to those bytes
afterwards.

Filed separately as **#2958** with a measured reproduction, rather than taken on here: it is
pre-existing, orthogonal to the precondition, and fixing it means giving each version its own code
directory, which reaches into `CodeStore`, `ContainerLauncher` and the delete paths.

## Why this is safe to merge on its own

It is additive and opt in. A request without `CodeSha256` behaves exactly as before, which
`omittingCodeSha256StillPublishes` pins explicitly. **No existing test changes**, which is the
property that separates this from the other half.

## AWS Compatibility

Matches the measurement in the issue:

```
publish-version --code-sha256 <wrong hash>
# CodeSHA256 (...) is different from current CodeSHA256 in $LATEST
```

## How it was tested

New `LambdaPublishVersionCodeSha256IntegrationTest`:

| Test | Covers |
|---|---|
| `aMismatchedCodeSha256IsRejectedAndPublishesNothing` | 400 with the AWS message, and the rejected call created no version |
| `aMatchingCodeSha256Publishes` | A correct hash is not blocked |
| `omittingCodeSha256StillPublishes` | Callers that never send the field are unaffected |
| `aNonStringCodeSha256IsRejectedRatherThanIgnored` | `{"CodeSha256": 12345}` is a 400 and publishes nothing |
| `aMalformedBodyIsRejectedRatherThanPublished` | An unparseable body is a 400 and publishes nothing |
| `aNullJsonBodyIsRejectedRatherThanPublished` | A literal `null` body is a 400, not the 500 it used to be |
| `aBlankCodeSha256IsRejectedRatherThanTreatedAsOmitted` | An empty string is compared and fails as a mismatch, rather than skipping the check |

Full suite: `Tests run: 15059, Failures: 0, Errors: 1, Skipped: 9`. Docs gate: `make docs-test` 50
passed, `make docs-check` exit 0.

`java.sql.SQLException: No suitable driver found for jdbc:h2:mem:...`. It is not from this change: it
fails identically on clean `main` with my work stashed, and passes when that class runs alone, so it
looks like a driver registration race under the full parallel suite. Happy to file it separately if
it is not already tracked.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -393,14 +393,16 @@ public class LambdaController {
                                    String body) {
         String region = regionResolver.resolveRegion(headers);
         String description = null;
+        String codeSha256 = null;
         if (body != null && !body.isBlank()) {
             try {
                 @SuppressWarnings("unchecked")
                 Map<String, Object> req = objectMapper.readValue(body, Map.class);
                 description = (String) req.get("Description");
+                codeSha256 = (String) req.get("CodeSha256");
             } catch (Exception ignored) {}
         }
-        LambdaFunction version = lambdaService.publishVersion(region, functionName, description);
+        LambdaFunction version = lambdaService.publishVersion(region, functionName, description, codeSha256);
         return Response.status(201).entity(buildFunctionConfiguration(version)).build();
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -395,16 +395,41 @@ public class LambdaController {
         String description = null;
         String codeSha256 = null;
         if (body != null && !body.isBlank()) {
+            Map<String, Object> req;
             try {
                 @SuppressWarnings("unchecked")
-                Map<String, Object> req = objectMapper.readValue(body, Map.class);
-                description = (String) req.get("Description");
-                codeSha256 = (String) req.get("CodeSha256");
-            } catch (Exception ignored) {}
+                Map<String, Object> parsed = objectMapper.readValue(body, Map.class);
+                req = parsed;
+            } catch (Exception e) {
+                // Swallowing this would treat a malformed body as an empty one, so a request whose
+                // CodeSha256 could not be read would publish instead of failing its precondition.
+                throw new AwsException("InvalidParameterValueException",
+                        "Could not parse request body: " + e.getMessage(), 400);
+            }
+            description = stringField(req, "Description");
+            codeSha256 = stringField(req, "CodeSha256");
         }
         LambdaFunction version = lambdaService.publishVersion(region, functionName, description, codeSha256);
         return Response.status(201).entity(buildFunctionConfiguration(version)).build();
     }
+
+    /**
+     * Reads a string field, rejecting a value of the wrong JSON type rather than letting a cast
+     * failure be swallowed. A {@code CodeSha256} that arrives as a number or an object is a
+     * malformed precondition, and treating it as absent would publish without checking anything.
+     */
+    private static String stringField(Map<String, Object> req, String name) {
+        Object value = req.get(name);
+        if (value == null) {
+            return null;
+        }
+        if (!(value instanceof String text)) {
+            throw new AwsException("InvalidParameterValueException",
+                    name + " must be a string", 400);
+        }
+        return text;
+    }
+
 
     @GET
     @Path("/functions/{functionName}/versions")
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaController.java
@@ -406,6 +406,13 @@ public class LambdaController {
                 throw new AwsException("InvalidParameterValueException",
                         "Could not parse request body: " + e.getMessage(), 400);
             }
+            if (req == null) {
+                // A literal "null" body parses cleanly to a null map, so it never reaches the catch
+                // above. Checked here rather than inside the try, where this exception would be
+                // caught by that same catch and rewrapped as a parse failure it is not.
+                throw new AwsException("InvalidParameterValueException",
+                        "Request body must be a JSON object", 400);
+            }
             description = stringField(req, "Description");
             codeSha256 = stringField(req, "CodeSha256");
         }
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -1334,8 +1334,12 @@ public class LambdaService implements ResourceProvider {
             // Inside the lock UpdateFunctionCode takes, so the hash cannot be checked against one
             // version of $LATEST and the snapshot then taken from another. Checking it outside
             // would let an overlapping deploy publish code the caller never authorised.
-            if (expectedCodeSha256 != null && !expectedCodeSha256.isBlank()
-                    && !expectedCodeSha256.equals(fn.getCodeSha256())) {
+            // No isBlank() exclusion here. A present but empty value was previously treated as
+            // absent, so it skipped the comparison entirely and published without checking
+            // anything, which is the failure this precondition exists to prevent. It is simply
+            // compared like any other value and fails as a mismatch, which avoids inventing an
+            // error shape for the empty case that has not been measured against the live service.
+            if (expectedCodeSha256 != null && !expectedCodeSha256.equals(fn.getCodeSha256())) {
                 throw new AwsException("InvalidParameterValueException",
                         "CodeSHA256 (" + expectedCodeSha256 + ") is different from current CodeSHA256 in $LATEST",
                         400);
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -1311,6 +1311,18 @@ public class LambdaService implements ResourceProvider {
     }
 
     public LambdaFunction publishVersion(String region, String functionName, String description) {
+        return publishVersion(region, functionName, description, null);
+    }
+
+    /**
+     * Publishes a version of {@code $LATEST}.
+     *
+     * <p>{@code CodeSha256} is a precondition: "only publish a version if the hash value matches the
+     * value that's specified". It was never parsed out of the request, so a caller that raced
+     * someone else's deploy published code it had not authorised, silently (issue #2822).
+     */
+    public LambdaFunction publishVersion(String region, String functionName, String description,
+                                         String expectedCodeSha256) {
         LambdaFunction fn = getFunction(region, functionName);
         functionName = fn.getFunctionName();
         // Shares the per-function lock deleteFunction and extractZipCodeBytes take around their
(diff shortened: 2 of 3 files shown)


pgermosen (commented):
Good intent and the additive framing is right, but I don't think this closes the race it's meant to close yet. The description says the check is safe because it sits inside the lock UpdateFunctionCode takes, I traced UpdateFunctionCode on main and that's not the case, its code mutation path (extractZipCodeBytes, where CodeSha256 actually gets set) runs entirely outside any synchronized(lockForConcurrencyOp(...)) block. The only lock acquisition in there wraps an unrelated legacy-directory cleanup further down. So PublishVersion's own lock isn't actually serializing against a concurrent UpdateFunctionCode, which means the exact race #2822 describes, a correctly-matched hash still publishing code from a deploy that landed in between, is still open.

Separately, the malformed-input case is worth a look too: the try/catch around the request body parse is pre-existing for Description, but now CodeSha256 shares it, and a non-string value like {"CodeSha256": 12345} throws a ClassCastException that gets swallowed there, leaving the field null, indistinguishable from omitted. So a malformed-but-present value bypasses the precondition the same way an absent one would, just less obviously, and succeeds with 201 instead of a 400.

Everything else here looks right, the message text matches your measured transcript and the opt-in behavior is genuinely tested.

avison9:
You are right on both counts, and thank you for tracing `UpdateFunctionCode` rather than taking the
description at its word. The claim that the check sat inside the lock `UpdateFunctionCode` takes was
simply wrong, and I should not have written it without checking. I have corrected it in the PR
description as well, since you reviewed against it.

## The race

Confirmed exactly as you describe. `extractZipCodeBytes` sets `codeLocalPath`, `codeSizeBytes` and
`codeSha256` at the top of the method with no lock held, and the only `synchronized` block sits
further down around `reclaimLegacyCodeDirectoryIfUnused`. `PublishVersion` holding its own lock
serialises nothing, because the writer never contends for it.

The fix makes the writer take it. Hashing stays outside the lock, since it is pure computation on
bytes already in hand; only the fields that decide what a snapshot copies are published inside it:

```java
zipExtractor.extractTo(zipBytes, codePath);
String newSha256 = null;
try {
    byte[] digest = MessageDigest.getInstance("SHA-256").digest(zipBytes);
    newSha256 = Base64.getEncoder().encodeToString(digest);
} catch (NoSuchAlgorithmException ignored) {}
synchronized (lockForConcurrencyOp(fn.getFunctionArn())) {
    fn.setCodeLocalPath(codePath.toAbsolutePath().normalize().toString());
    fn.setCodeSizeBytes(zipBytes.length);
    if (newSha256 != null) {
        fn.setCodeSha256(newSha256);
    }
}
```

Extraction to disk deliberately stays outside the lock too, so a large package does not serialise
unrelated work on the same function; the window that matters is the one between the hash being read
and the snapshot being taken, and that is now covered on both sides.

## The malformed input

Also confirmed. `{"CodeSha256": 12345}` throws a `ClassCastException` into the pre-existing catch,
leaves the field null, and is then indistinguishable from omitted, so the precondition is skipped and
the publish succeeds with 201.

Both halves of that are now rejected. A body that cannot be parsed is a 400 rather than being treated
as empty, and field extraction goes through a helper that rejects a wrong JSON type instead of
letting the cast fail silently:

```java
private static String stringField(Map<String, Object> req, String name) {
    Object value = req.get(name);
    if (value == null) {
        return null;
    }
    if (!(value instanceof String text)) {
        throw new AwsException("InvalidParameterValueException", name + " must be a string", 400);
    }
    return text;
}
```

Two tests added for it: `{"CodeSha256": 12345}` and an unparseable body each return 400 and leave the
function with only `$LATEST` listed, so a rejected precondition provably publishes nothing.

## One thing worth flagging from doing this

My first attempt at that helper placed it between the `@POST @Path` annotations and
`publishVersion`, so JAX-RS bound them to the helper and the endpoint started returning 405. It
compiled cleanly and the two new tests still failed for an unrelated-looking reason; what caught it
was the three pre-existing tests in the same class going red at the same time. Worth mentioning
because the failure mode of a silently unrouted endpoint is not obvious from a compile.

Full suite figures are in the checks. The `RdsDataServiceTest` H2 driver error there is not from this
change: it fails identically on clean `main` with my work stashed, and passes when that class runs
alone.

pgermosen (commented):
Good progress on the last round, both things I flagged are genuinely fixed, the lock now actually covers UpdateFunctionCode's mutation of the tracked hash, and the malformed-body handling is solid with real tests behind each shape.

Two more things came up on a closer pass though. First, small one: an explicit CodeSha256 of empty or whitespace-only string gets treated as if it were omitted, since the isBlank() check short-circuits the whole condition. That's the same class of thing the rest of this round's fix was about, a present-but-malformed value silently behaving like an absent one. I don't have a live AWS transcript for this specific case though, so worth a quick probe to confirm exactly what AWS does with an empty string there before locking in the fix.

Second, bigger one: I traced where codeLocalPath actually comes from, and there's no per-version code isolation anywhere in this codebase. CodeStore.getCodePath derives a path from account and function name only, extractZipCodeBytes overwrites that same directory in place on every UpdateFunctionCode, and publishVersion's snapshot just stores a reference to that same path string rather than copying the code. So a published version doesn't actually freeze its code, a later UpdateFunctionCode, racing or not, silently changes what invoking an already-published version runs. AWS's own PublishVersion docs describe a version as a snapshot of code and configuration that doesn't change, so this is a real gap against that guarantee, not just an edge case. It's bigger than this PR's stated scope, but the CodeSha256 precondition this PR adds can't actually deliver what it implies once you look past the moment of publish. Would want either this PR taking on a real per-version code copy, or an explicit note in the description plus a follow-up issue so it's not silently assumed to be handled.

avison9:
Both handled, though the first one differently from how I had it when you commented.

## The blank hash

You asked for a live probe before locking the fix in, and I cannot run one: I have no AWS account,
so every AWS behaviour I have cited on these PRs has come from the transcripts in the issues rather
than from my own measurement. I should be straight about that rather than quietly picking something.

I had already written a dedicated `CodeSha256 must not be blank` rejection. That was inventing an
error shape for the empty case with nothing to back it, which is exactly what your caution was
about, so I have taken it out.

What is there instead needs no assumption: the `isBlank()` exclusion is simply removed, so an empty
string is compared like any other value and fails as an ordinary mismatch.

```java
if (expectedCodeSha256 != null && !expectedCodeSha256.equals(fn.getCodeSha256())) {
    throw new AwsException("InvalidParameterValueException",
            "CodeSHA256 (" + expectedCodeSha256 + ") is different from current CodeSHA256 in $LATEST", 400);
}
```

The hole closes, and the error is one AWS demonstrably produces for a hash that does not match, per
the transcript in #2822. If a probe later shows AWS rejects an empty string with something more
specific, that is a one-line change on top.

## Per-version code isolation

Filed as **#2958**, and I have added the explicit note to the description as you asked, quoting your
framing that the precondition "can't actually deliver what it implies once you look past the moment
of publish", so nobody reads it as more than it is.

I reached the same conclusion from the Greptile pass shortly before your comment, and the issue
carries a measured reproduction. Worth flagging what I actually measured, because my first attempt at
it was wrong: I probed with `GetFunction?Qualifier=1` on this branch, which is cut from `main` and so
does not carry #2932's qualifier fix, so the read silently returned `$LATEST` and the numbers looked
like the code path was broken when I was really just re-measuring the other bug. Redone through
`ListVersionsByFunction`:

```
```

So it is slightly narrower than "the version's code changes": the version's *metadata* is immutable
and keeps its own hash, while the directory that hash describes has been rewritten underneath it. A
version therefore advertises one checksum over a different build. The issue says which half I
measured and which half is inferred from `CodeStore.getCodePath`, `extractZipCodeBytes` and
`ContainerLauncher`, since confirming the invoke needs a container runtime.

Happy to take #2958 next if you want it soon; it needs a real decision on per-version directories
versus content-addressed storage, which is why it is not folded in here.

Full suite: `Tests run: 15059, Failures: 0, Errors: 1, Skipped: 9`, the error being the
`RdsDataServiceTest` H2 driver race that fails identically on clean `main`.

pgermosen (approved):
Both closed, and closed well. Removing the isBlank() exclusion rather than inventing a dedicated error for the empty case is exactly right, it reuses only what's already been measured against the live service instead of guessing at the unmeasured part. Appreciate the honesty about not having an AWS account to probe with too, that's the right way to handle a gap you can't fully verify.

And #2958 is the right home for the code-isolation issue, that's a real architectural decision (per-version directories vs content-addressed storage) that doesn't belong in a precondition-check PR. Good catch re-measuring it properly after noticing the first probe was confounded by this branch not carrying #2932's fix.

This closes out everything from the last few rounds.

PR #2940, opened 2026-09-02
perf(build): keep only the snappy JNI libraries the image will use

exoego:
## Summary

This PR downsizes Docker image size from 274.4 to 267.5 MiB (uncompressed).
By preserving only four snappy-java's JNI libraries for linux (aarch64, amd64) and mac (aarch64, amd64) .

Adds linux/amd64 to compatibility matrix.

Floci uses snappy-java in

- FirehoseCompression. It writes the Snappy and HADOOP_SNAPPY formats for Firehose S3 delivery through SnappyOutputStream and SnappyHadoopCompatibleOutputStream. 
- Kafka-clients declares it at runtime scope. EventBridge Pipes reads MSK sources with a KafkaConsumer, so a topic
holding snappy-compressed records is decompressed through the same library.

In CI, only the Firehose path is exercised, by FirehoseCompressionDeliveryTest seemingly. The Kafka path has no test, but both need the same file, so a wrong exclusion still fails the Firehose test.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No change.

## Checklist

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -86,9 +86,20 @@ jobs:
             || { echo "::error::post-subcommand --endpoint-url flag was not honored over AWS_ENDPOINT_URL"; exit 1; }
 
   build-native:
-    name: Build native floci image
-    runs-on: ubuntu-24.04-arm
+    name: Build native floci image (${{ matrix.arch }})
+    runs-on: ${{ matrix.runner }}
     timeout-minutes: 30
+    # Both published architectures. The native image is the only place an
+    # arch-specific resource can go missing (snappy-java's JNI library is
+    # registered per platform), and one arch cannot prove the other.
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: arm64
+            runner: ubuntu-24.04-arm
+          - arch: amd64
+            runner: ubuntu-latest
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thank you, this is a clean cut and the amd64 compat leg is the part I am happiest about: the JVM test suite reads the jar, so before this PR nothing ever proved the x86_64 library survived the native build.

I checked the list against snappy-java 1.1.10.8 (the version quarkus-bom 3.32.3 pins): 28 library files, 23 excluded, 5 kept. The kept set covers linux/amd64 and linux/arm64 from release.yml plus a local Mac native build, and the runtime base is ubi9-minimal (glibc), so dropping x86_64-musl is right. The library registers itself through its own `META-INF/native-image/.../resource-config.json`, so an exclude list is the only lever short of shading the jar.

**(follow-up, separate PR)** `Mac/x86/**` (the 32-bit jnilib, 44 KB) also survives; harmless, but it can go with the rest.

For context: our audit put the native layer at +19.2 MB compressed between 1.7.0 and 2.0.0, and this recovers about 2.1 MB of that, so the rest is elsewhere.

No blockers, merging.

PR #2941, opened 2026-09-02
fix(lambda): do not publish a version when nothing has changed

### LINKED ISSUE #2822: [BUG] Lambda PublishVersion always creates a version and ignores the CodeSha256 precondition

avison9:
## ⚠️ This changes behaviour for existing callers, and changes the compatibility suite

Closes #2822. **Stacked on #2939**, which is the safe half of the same issue and should land first;
until it does, the diff here shows its commit too.

This is the half worth arguing about. Please read the compatibility suite section before the code.

## Summary

"AWS Lambda doesn't publish a version if the function's configuration and code haven't changed since
the last version." Publishing was unconditional, so repeated publishes accumulated identical
versions. The issue reporter measured the live service:

```
versions before:                              $LATEST  1
publish-version   (nothing changed)        -> 2        (AWS: returns the existing version)
publish-version   (nothing changed again)  -> 3        (AWS: returns the existing version)
versions after:                               $LATEST  1  3  2
```

## What changed

"Unchanged" is decided by `$LATEST`'s `revisionId`, which is regenerated on every code and
configuration update. A version records the revision it was cut from, and a later publish that finds
it unmoved returns that version.

Keying on the revision rather than comparing the thirty-odd fields a snapshot copies means a field
added to versions later is covered without anyone remembering to update a comparison. `LambdaFunction`
gains one field, which is safe for persisted state because the model is
`@JsonIgnoreProperties(ignoreUnknown = true)`; versions published before this carry null and never
match, so existing state keeps its previous behaviour.

Three carve-outs, each with a test:

**Description** is supplied at publish time rather than carried on `$LATEST`, so a publish naming a
new one is a real publish. I had this as an open question, and this repository answered it:
`LambdaVersionIntegrationTest.publishSecondVersion` publishes with a new description and asserts it
gets version 2.

**Only the most recent version counts**, because AWS compares against the last version. A description
going A then B then A must publish again rather than matching the older A, which would hand back
version 1 and leave the caller holding a version whose place in the history is wrong.

**Hot-reload functions always publish.** Their code lives in a bind-mounted directory that changes
with no API call, so `revisionId` cannot witness it, and de-duplicating would return a version whose
identity no longer describes what will run.

## The compatibility suite changes, and that deserves scrutiny

`CodeDeployTest.setupLambdaFunctionForDeployment` publishes twice with nothing changed and asserts
the two versions differ. Against this change it gets `"1"` both times, and the five downstream
blue/green assertions fail with it.

That suite exists to encode real AWS behaviour, so if it disagrees with a change the presumption
should be against the change. Here I think the test is encoding Floci's current behaviour rather than
AWS's: against the live service that assertion would fail, per the measurement above. A real
blue/green deployment ships a new build and then publishes; it does not publish identical bytes
twice. So it now deploys a change between the two publishes.

`LambdaServicePersistenceTest.publishVersionContinuesNumberingAfterRestart` reached version 3 the
same way. Its subject is the version counter surviving a restart, so it now varies the description
across the three publishes, which exercises the same thing.

**If maintainers would rather keep Floci's current behaviour, this is the PR to reject.** #2939
stands on its own and needs nothing from here. I would rather put that choice in front of you than
land it quietly inside a bug fix.

## How it was tested

New `LambdaPublishVersionDedupIntegrationTest`:

| Test | Covers | Fails without the fix |
|---|---|---|
| `publishingRepeatedlyWithNothingChangedReturnsTheExistingVersion` | Three unchanged publishes return one version, and the list does not grow | Yes, `expected: <1> but was: <2>` |
| `aRealChangeStillProducesANewVersion` | A code change and a configuration-only change each still mint a version | No, a do-no-harm guard |
| `aDescriptionReturningToAnEarlierValueStillPublishes` | A then B then A publishes three times | Yes, returns version 1 |

Full suite: `Tests run: 15058, Failures: 0, Errors: 1, Skipped: 9`. Docs gate: `make docs-test` 50
passed, `make docs-check` exit 0.

The single error is `RdsDataServiceTest.batchExecuteStatementRunsEveryParameterSet` with
`java.sql.SQLException: No suitable driver found for jdbc:h2:mem:...`. Not from this change: it fails
identically on clean `main` with my work stashed, and passes when that class runs alone.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

Not marked breaking because no API signature changes, but it is a deliberate behaviour change and I
would not object to it being treated as one.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CodeDeployTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CodeDeployTest.java
@@ -300,6 +300,14 @@ class CodeDeployTest {
         v1 = pv1.version();
         assertThat(v1).isNotBlank();
 
+        // A second version needs a real change behind it. AWS does not publish a version when the
+        // code and configuration have not moved since the last one, so two publishes in a row
+        // return the same version. A blue/green deployment shifts between two different builds
+        // anyway, so deploy one here rather than publishing the same bytes twice.
+        lambda.updateFunctionCode(r -> r
+                .functionName(DEPLOY_FUNCTION)
+                .zipFile(SdkBytes.fromByteArray(LambdaUtils.handlerZip())));
+
         PublishVersionResponse pv2 = lambda.publishVersion(r -> r.functionName(DEPLOY_FUNCTION));
         v2 = pv2.version();
         assertThat(v2).isNotBlank().isNotEqualTo(v1);
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -44,6 +44,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -1350,6 +1351,46 @@ public class LambdaService implements ResourceProvider {
      * {@code concurrencyOpLocks}) so publishes of unrelated functions do not
      * serialize on the instance monitor for the storage round-trip.
      */
+    /**
+     * The already-published version cut from {@code $LATEST}'s current revision, if there is one.
+     *
+     * <p>Versions published before {@code sourceRevisionId} was recorded carry null and never match,
+     * so existing state keeps its previous behaviour rather than deduplicating against a revision
+     * nobody wrote down.
+     */
+    private LambdaFunction latestPublishedFrom(String region, LambdaFunction fn, String description) {
+        String current = fn.getRevisionId();
+        if (current == null) {
+            return null;
+        }
+        // A hot-reload function's code lives in a bind-mounted directory that changes without any
+        // API call, so revisionId cannot witness it. Deduplicating would hand back a version whose
+        // identity no longer describes what will actually run, so these always publish.
+        if (fn.getHotReloadHostPath() != null) {
+            return null;
+        }
+        // Description is supplied at publish time rather than carried on $LATEST, so a publish that
+        // names a new one is a real publish even when nothing else moved. LambdaVersionIntegrationTest
+        // asserts exactly that: two publishes differing only in Description produce versions 1 and 2.
+        String effective = description != null ? description : fn.getDescription();
+        // Only the most recent version counts. AWS compares against the last version, so a
+        // description going A then B then A must publish again rather than matching the older A:
+        // scanning every version would return version 1 and leave the caller with a version whose
+        // place in the history is wrong.
+        return newestVersion(region, fn.getFunctionName())
+                .filter(v -> current.equals(v.getSourceRevisionId()))
+                .filter(v -> Objects.equals(effective, v.getDescription()))
+                .orElse(null);
+    }
+
+    /** The highest-numbered published version of a function, if any. */
+    private java.util.Optional<LambdaFunction> newestVersion(String region, String functionName) {
+        return functionStore.listVersions(region, functionName).stream()
+                .filter(v -> v.getVersion() != null && !"$LATEST".equals(v.getVersion()))
+                .filter(v -> v.getVersion().chars().allMatch(Character::isDigit))
+                .max(java.util.Comparator.comparingLong(v -> Long.parseLong(v.getVersion())));
+    }
+
     private int nextVersionNumber(String counterKey, String legacyCounterKey) {
         synchronized (versionCounterLocks.computeIfAbsent(counterKey, k -> new Object())) {
             Integer current = versionCounters.get(counterKey);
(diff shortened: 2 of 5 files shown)


avison9:
Rebased onto `main` now that #2939 has merged, so the conflict is gone. It was the squash merge: this branch still carried #2939's four original commits while `main` had them collapsed into `3d2b8828`, so the same changes arrived by two routes.

Dropped those rather than replaying them, so this is now a single commit and the diff is only the de-duplication work: five files, 183 insertions.

Verified against the new `main`, which has also picked up the kubernetes client rewrite and the cloudtrail work since my last run. Full suite `Tests run: 15225, Failures: 0, Errors: 1, Skipped: 9`, the error being the `RdsDataServiceTest` H2 driver race that fails identically on clean `main`. The four affected classes are 22/22, which now includes #2939's seven CodeSha256 tests arriving via `main` rather than via this branch, so the two halves compose.

Nothing in the change itself moved. The decision this PR asks for is unchanged and still the substance of it: whether Floci should match AWS in not publishing a version when nothing has changed, given that doing so requires `CodeDeployTest` to deploy a real change between its two publishes rather than publishing identical bytes twice.

pgermosen (approved):
Approving. The whole point of floci is to emulate real AWS, not invent behavior, and the measurement here is clear: AWS doesn't publish a duplicate version when nothing's changed, so this is fixing a real emulation gap rather than introducing new behavior. The CodeDeployTest change is a consequence of that, not scope creep - it was relying on the bug to get two distinct versions. Dedup logic checks out: revisionId matched against the newest version's sourceRevisionId, the description carve-out and hot-reload exemption both implemented as described, and the three new tests cover exactly what the table claims.

One thing worth a follow-up, not blocking this: publishVersion's comment claims it runs "inside the lock UpdateFunctionCode takes," but updateFunctionCode and updateFunctionConfiguration don't take lockForConcurrencyOp at all, so an overlapping update can mutate the shared LambdaFunction object mid-copy and produce a snapshot that's part old code, part new config, with a sourceRevisionId that doesn't cleanly identify either state. That's inherited from #2939's locking model, not something this PR introduces or worsens, but it's worth fixing at the source by having those two methods take the same lock.

I still leave the merge of this one to @hectorvent and his approval

PR #2943, opened 2026-09-02
fix(ec2): include network metadata in instance type responses

### LINKED ISSUE #2942: [FEAT] ec2: add NetworkInfo.EncryptionInTransitSupported to DescribeInstanceTypes responses
## Service

EC2

## API Action / Feature

DescribeInstanceTypes response compatibility for NetworkInfo.EncryptionInTransitSupported.

## AWS Documentation

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTypes.html

## Why is this needed?

Floci 2.0.1 returns HTTP 200 for DescribeInstanceTypes, but the response omits the NetworkInfo object and its EncryptionInTransitSupported boolean.

This is a valid AWS response-shape requirement: the field is part of the EC2 DescribeInstanceTypes network information model. Karpenter AWS provider 1.8.8 reads this field while computing instance-type requirements. Against Floci 2.0.1, the missing nested object causes a nil-pointer panic during EC2NodeClass reconciliation instead of allowing the class to reach Ready.

## Reproduction

Environment:

- Floci 2.0.1, pinned by OCI digest sha256:4e451c39c7bb88e3cd4f87e8fc0c25d5b47695a51185d521e2241fa00486e8eb.
- Karpenter AWS provider 1.8.8.
- Fake non-empty AWS credentials against the local Floci endpoint.
- A fresh EC2NodeClass using an AMI-by-ID selector, subnet tags, and security-group tags. The selector fixtures reconcile successfully.

Directly call DescribeInstanceTypes for representative catalog entries such as m5.large and t4g.medium through the Floci endpoint. The API returns HTTP 200 and valid instance-type, architecture, vCPU, and memory data, but no NetworkInfo.EncryptionInTransitSupported field.

The Karpenter discovery gate then shows:

- The DescribeInstanceTypes response-shape contract fails because NetworkInfo.EncryptionInTransitSupported is absent.
- EC2NodeClass reconciliation panics in Karpenter's instance-type requirement calculation at pkg/providers/instancetype/types.go:212.
- The EC2NodeClass does not publish Ready.
- The controller pod remains running with zero restarts, but the reconciliation log contains Observed a panic.
- No NodeClaims or EC2 instances are created; provisioning is disabled.

This should be fixed at the Floci EC2 wire-protocol layer, not by adding a Karpenter-specific workaround or masking the missing field with an unrelated fake success.

## Scope

- Extend the EC2 instance-type catalog/model with network information containing EncryptionInTransitSupported.
- Emit a complete NetworkInfo object with a boolean EncryptionInTransitSupported value for every returned instance type, using type-accurate catalog data.
- Preserve the existing DescribeInstanceTypes filters and pagination behavior.
- Add a focused regression/integration test that asserts the nested response shape through the AWS-compatible API.
- Add or update the catalog fixtures with values verified against the AWS API for representative types such as m5.large and t4g.medium.

The implementation should remain general EC2 compatibility rather than special-casing Karpenter.

## Docs

No new API action is introduced, so make docs-sync should not need generated action-table changes. Please document any new catalog configuration or fixture field if one is exposed.

- [x] Yes
- [ ] No

luiseterc:
## Summary

Closes #2942

Add the missing EC2 `networkInfo.encryptionInTransitSupported` field to `DescribeInstanceTypes` responses.

Floci's instance-type catalog now carries explicit network compatibility metadata, and the Query response emits the nested AWS wire shape that Karpenter expects. The catalog is validated at load time so future entries cannot silently regress to a null response field.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The previous response omitted `NetworkInfo.EncryptionInTransitSupported`, which AWS SDK clients deserialize as a nullable field. Karpenter can dereference that field during EC2NodeClass reconciliation and panic when it is absent.

The local catalog currently models this field as `false` for all included instance types. The value is explicit fixture data, not a fake success response, and can be refined as the catalog gains more AWS-accurate per-type metadata.

Validation:
- `mvn -Dtest=Ec2InstanceTypeCatalogTest,Ec2IntegrationTest test`: 186 tests passed.
- `make docs-check`: passed.
- AWS SDK compatibility tests: test sources compile successfully.
- The SDK suite's live run requires a Floci endpoint at `localhost:4566`; no endpoint was running in this checkout, so that suite was not reported as passed.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2Tests.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2Tests.java
@@ -195,6 +195,23 @@ class Ec2Tests {
 
     @Test
     @Order(7)
+    @DisplayName("DescribeInstanceTypes - network compatibility metadata")
+    void describeInstanceTypeNetworkCompatibilityMetadata() {
+        DescribeInstanceTypesResponse resp = ec2.describeInstanceTypes(DescribeInstanceTypesRequest.builder()
+                .instanceTypes(InstanceType.M5_LARGE, InstanceType.fromValue("t4g.medium"))
+                .build());
+
+        assertThat(resp.instanceTypes()).hasSize(2);
+        assertThat(resp.instanceTypes()).allSatisfy(instanceType -> {
+            assertThat(instanceType.networkInfo()).isNotNull();
+            assertThat(instanceType.networkInfo().encryptionInTransitSupported()).isNotNull();
+        });
+        assertThat(resp.instanceTypes()).allMatch(instanceType ->
+                !instanceType.networkInfo().encryptionInTransitSupported());
+    }
+
+    @Test
+    @Order(7)
     @DisplayName("DescribeInstanceTypes - processor architectures")
     void describeInstanceTypeProcessorArchitectures() {
         DescribeInstanceTypesResponse resp = ec2.describeInstanceTypes(DescribeInstanceTypesRequest.builder()
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2InstanceTypeCatalog.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2InstanceTypeCatalog.java
@@ -93,6 +93,10 @@ public class Ec2InstanceTypeCatalog {
             if (instanceType.supportedArchitectures == null || instanceType.supportedArchitectures.isEmpty()) {
                 throw new IllegalStateException("EC2 instance type catalog entry is missing supportedArchitectures: " + name);
             }
+            if (instanceType.encryptionInTransitSupported == null) {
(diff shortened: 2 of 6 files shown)


pgermosen (approved):
Solid fix, traced the full path and it holds up: the catalog validation at load time now throws if this field is missing, so a future entry can't silently reintroduce the gap, and the XML nesting after the split closes correctly. Good layered test coverage too, especially the SDK compat test, that's the one that actually reproduces the kind of nullability panic Karpenter would hit.

Appreciate being upfront that the false default is fixture data rather than verified per-type fact, the docs don't give per-type guidance so there's no way to confirm that from here, and the actual bug being fixed is the null field, not the specific value.

PR #2945, opened 2026-09-02
fix(cloudformation): preserve S3 physical IDs on updates

### LINKED ISSUE #2592: [BUG] CloudFormation: resources are recreated on every `update-stack`, so physical IDs and `Ref` values change
## Service

CloudFormation

## AWS API Action

`UpdateStack` - physical resource ID stability, and `Ref` resolution to it.

## Context

An unchanged resource is given a new physical ID on every `update-stack`. Real
CloudFormation only replaces a resource when a property requiring replacement
changes; anything else keeps its physical ID, and `Ref` keeps resolving to it.

## Expected behavior

The same name both times. The `Bkt` resource is untouched by the
update.

## Actual behavior

```
bucket after CREATE : ref-stability-bkt-97bbdb13758f
bucket after UPDATE : ref-stability-bkt-7e73ff84cc9a
```

Expected: the same name both times. The `Bkt` resource is untouched by the
update.

## Reproduction

```sh
export AWS_DEFAULT_REGION=us-east-1 AWS_ENDPOINT_URL=http://localhost:4566

cat > r1.json <<'JSON'
{ "AWSTemplateFormatVersion": "2010-09-09",
  "Resources": { "Bkt": { "Type": "AWS::S3::Bucket" } },
  "Outputs": { "B": { "Value": { "Ref": "Bkt" } } } }
JSON

# same bucket, plus an unrelated queue
cat > r2.json <<'JSON'
{ "AWSTemplateFormatVersion": "2010-09-09",
  "Resources": { "Bkt": { "Type": "AWS::S3::Bucket" },
                 "Q":   { "Type": "AWS::SQS::Queue" } },
  "Outputs": { "B": { "Value": { "Ref": "Bkt" } } } }
JSON

aws cloudformation create-stack --stack-name ref-stability --template-body file://r1.json
aws cloudformation describe-stacks --stack-name ref-stability \
  --query 'Stacks[0].Outputs[0].OutputValue' --output text

aws cloudformation update-stack --stack-name ref-stability --template-body file://r2.json
aws cloudformation describe-stacks --stack-name ref-stability \
  --query 'Stacks[0].Outputs[0].OutputValue' --output text
```

## Why this matters?

This makes the Serverless Framework unusable against Floci, and the failure is
very hard to read. Serverless deploys in two steps: create a stack containing
only `ServerlessDeploymentBucket`, upload the code artifact into that bucket,
then update the stack with the real template, where every Lambda has
`Code: { S3Bucket: { Ref: ServerlessDeploymentBucket }, S3Key: ... }`.

Because the update recreates the bucket, `Ref` resolves to a fresh, empty one
and the Lambda fails with:

```
Handler file '<handler>' not found in deployment package
```

which points at handler paths rather than at a missing artifact. Each attempt
also leaves the previous bucket orphaned.

Also, there's a element of data loss being a wider concern: any stateful resource in a stack. A bucket, a queue, a table, etc, is silently replaced whenever anything else in that stack changes.

## Environment

- Floci 1.7.0
(`floci/floci@sha256:bec9b9f749322444a57fb0f00d3e007ffd49015162361aab84ccecd9b4f5f8ed`),
- macOS arm64
- Docker 29.5.3
- aws-cli 2.34.10.
- Serverless Framework 3.39.0

- [ ] Yes
- [x] No

AbarnaaSree:
## Summary

Fixes #2592.

* Preserve the existing physical ID for `AWS::S3::Bucket` resources during stack updates.
* Avoid recreating an unchanged S3 bucket when unrelated resources are added or updated.
* Add a regression test verifying that the bucket physical ID remains stable after `UpdateStack`.

## Testing

* `mvn -Dtest=CloudFormationIntegrationTest#updateStack_s3BucketWithUnrelatedResourceKeepsPhysicalId test`
* Result: **BUILD SUCCESS**
* Tests run: **1**
* Failures: **0**
* Errors: **0**

Note: The broader `CloudFormationIntegrationTest` run had one unrelated ECR test timeout (`createStack_ecrAutoName_isLowercase`); the new regression test passes independently.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -665,17 +665,27 @@ public class CloudFormationResourceProvisioner {
 
     private void provisionS3Bucket(StackResource r, JsonNode props, CloudFormationTemplateEngine engine,
                                    String region, String accountId, String stackName) {
-        String bucketName = resolveOptional(props, "BucketName", engine);
+        String bucketName = r.getPhysicalId();
+
         if (bucketName == null || bucketName.isBlank()) {
-            bucketName = generatePhysicalName(stackName, r.getLogicalId(), 63, true);
+            bucketName = resolveOptional(props, "BucketName", engine);
+
+            if (bucketName == null || bucketName.isBlank()) {
+                bucketName = generatePhysicalName(stackName, r.getLogicalId(), 63, true);
+            }
+
+            s3Service.createBucket(bucketName, region);
         }
-        s3Service.createBucket(bucketName, region);
+
         applyBucketCorsConfiguration(bucketName, props, engine);
+
         r.setPhysicalId(bucketName);
         r.getAttributes().put("Arn", AwsArnUtils.Arn.of("s3", "", "", bucketName).toString());
         r.getAttributes().put("DomainName", bucketName + ".s3.amazonaws.com");
-        r.getAttributes().put("RegionalDomainName", bucketName + ".s3." + region + ".amazonaws.com");
-        r.getAttributes().put("WebsiteURL", "http://" + bucketName + ".s3-website." + region + ".amazonaws.com");
+        r.getAttributes().put("RegionalDomainName",
+                bucketName + ".s3." + region + ".amazonaws.com");
+        r.getAttributes().put("WebsiteURL",
+                "http://" + bucketName + ".s3-website." + region + ".amazonaws.com");
         r.getAttributes().put("BucketName", bucketName);
     }
 
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationIntegrationTest.java
@@ -8,8 +8,8 @@ import jakarta.inject.Inject;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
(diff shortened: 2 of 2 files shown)


hectorvent:
Thank you for picking up #2592; the end-to-end test pinning the bucket name across UpdateStack is exactly the coverage this fix needs.

**(blocking)** The branch conflicts with main: the S3 arm moved to `services/cloudformation/provisioners/S3CfnProvisioner.java` in #2794, so `provisionS3Bucket` no longer exists where this hunk lands. A rebase would target `provisionBucket` there, which already receives `ctx.isUpdate()` and `ctx.priorPhysicalId()`.

**(blocking)** Reading the prior physical id before `BucketName` means an update that sets or renames `BucketName` silently keeps the old bucket. The order needs to be explicit name, else prior id, else generated. #2788 does exactly that across seven provisioners including S3, so please have a look at it before rebasing; your integration test on top of it may be the most valuable shape for this PR.

**(blocking)** In the CORS test the two new extractions read `StackResourceSummaries`, but DescribeStackResources emits `StackResources` (`CloudFormationQueryHandler.java:642`), and a GET is not served by the Query controller (`AwsQueryController.java:253`). Both sides come back null, so the assertion passes without checking anything.

A rebase onto main, or reshaping this as the test over #2788, unblocks it.

hectorvent:
Closing this as superseded. #2788 landed on 2026-09-03 and closed #2592 with the same fix applied across seven provisioners, including S3, with the precedence explicit name, then prior physical id, then generated. On main the S3 arm now lives in `S3CfnProvisioner` and reads its name through `ctx.stablePhysicalName(...)`, and `CfnStableNameUpdatePathTest.s3UpdateDoesNotRecreateTheBucketItAlreadyOwns` pins the behaviour.

What main still does not have is your end-to-end version: an `UpdateStack` through the Query API that adds an unrelated resource and asserts the bucket's physical id is unchanged. That is the issue's own reproduction and it is worth having. If you would like to contribute it, a fresh PR against main with just that test (on the POST form, reading `StackResources` from `DescribeStackResources`) would be very welcome, and your first-PR CI gate will be approved promptly.

Thank you for picking up #2592.

PR #2950, opened 2026-09-02
fix(kinesis): handle minimum partition key hash

nkootstra:
## Summary

  Prevent Kinesis `PutRecord` from returning a `500` response when a partition key has the hash value `Integer.MIN_VALUE`.

  Use `Math.floorMod` to keep the selected shard index valid. Add an integration test that covers the failing partition key.

  ## Type of change

  - [x] Bug fix (`fix:`)
  - [ ] New feature (`feat:`)
  - [ ] Docs / chore

  ## AWS Compatibility

  The valid partition key `polygenelubricants` previously produced a negative shard index on a multi-shard stream. This caused `PutRecord` to return a `500` response.

  Verified with the AWS CLI against a three-shard stream. `PutRecord` completed successfully and returned a valid shard ID.

  ## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisService.java
@@ -766,7 +766,7 @@ public class KinesisService implements ResourceProvider {
 
     private KinesisShard selectShard(KinesisStream stream, String partitionKey) {
         // Simple hash-based shard selection among ALL shards, then resolve to open one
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Nice, minimal fix, and glad you went with widening to long rather than floorMod or a bitmask, both of those would've silently reshuffled shard assignment for every other negative-hash key on a stream that's already running, this one only touches the one value that used to crash. The negative-hash regression test is exactly the right thing to have caught that if it had gone the other way.

PR #2951, opened 2026-09-02
fix(cognito): name the user pool in the not-found message

### LINKED ISSUE #2853: [BUG] cognito: user pool not-found message is "User pool not found", AWS names the pool

bsaurusrex:
## Summary

Every Cognito action that resolves a user pool reported a missing one as `User pool not found`.
The live service names the pool. The error code (`ResourceNotFoundException`) and the status were
already correct; only the message diverged, which is enough to stop tooling and tests that match
on Cognito error text from behaving the same way locally, and Floci's own suites do assert on
messages elsewhere.

Three throw sites carried the literal. They now share one private helper, so the string has a
single home and cannot drift apart again.

Closes #2853

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Reference:
[DescribeUserPool](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DescribeUserPool.html).
The reference lists `ResourceNotFoundException` but does not publish the message, so it was
measured rather than inferred.

Live service, `ap-southeast-1`, AWS CLI v2, on a pool id that does not exist. Four actions on the
same resolution path, all identical:

```
describe-user-pool      -> ResourceNotFoundException: User pool ap-southeast-1_ZZZZZZZZZ does not exist.
```

Floci on `main`, same input:

```
describe-user-pool      -> ResourceNotFoundException: User pool not found
```

The trailing period is part of the live string and is reproduced.

### A malformed id is a different path, left alone

Worth stating so the change is not read as covering more than it does. AWS rejects an id that does
not match the pool-id pattern before resolving anything:

```
describe-user-pool --user-pool-id not-a-pool-id
  -> InvalidParameterException
     1 validation error detected: Value 'not-a-pool-id' at 'userPoolId' failed to satisfy
     constraint: Member must satisfy regular expression pattern: [\w-]+_[0-9a-zA-Z]+
```

Floci does not validate that pattern, so a malformed id falls through to the not-found path. This
PR does not change that: it is a pre-existing divergence in a different check, and adding
parameter validation across the service is a separate piece of work rather than a message fix.

## Checklist

Cognito package run, on the branch tip:

```
mvn -B test -Dtest='io.github.hectorvent.floci.services.cognito.**'
Tests run: 465, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

`make docs-check` exits 0. The full suite is left to CI rather than run locally.

New class `CognitoUserPoolNotFoundMessageIntegrationTest`, 4 tests, covering `DescribeUserPool`,
`ListGroups`, `ListResourceServers` and `AdminGetUser`.

One existing assertion moved. `S3WellKnownKeyIntegrationTest` used the old message as a
fingerprint for a request being misrouted to Cognito instead of S3. It now matches on
`"User pool "`, which fingerprints the same thing and survives the reword. Re-run with the change:

```
mvn -B test -Dtest='S3WellKnownKeyIntegrationTest,CognitoUserPoolNotFoundMessageIntegrationTest'
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0
```

## Found while measuring, filed separately

`ListUsers` and `ListUserPoolClients` do not resolve the pool at all: given an id that does not
exist they answer `200` with an empty list, where the live service returns
`ResourceNotFoundException`. That is why those two are not in the new test, despite being in the
measurement above.

```
FLOCI  list-users             --user-pool-id us-east-1_ZZZZZZZZZ  -> {"Users": []}
FLOCI  list-user-pool-clients --user-pool-id us-east-1_ZZZZZZZZZ  -> {"UserPoolClients": []}
FLOCI  list-groups            --user-pool-id us-east-1_ZZZZZZZZZ  -> ResourceNotFoundException
FLOCI  list-resource-servers  --user-pool-id us-east-1_ZZZZZZZZZ  -> ResourceNotFoundException
```

That is a missing existence check rather than a message, so it is out of scope here and filed on
its own.

### DIFF
--- a/docs/services/cognito.md
+++ b/docs/services/cognito.md
@@ -15,6 +15,8 @@ Floci serves pool-specific discovery and JWKS endpoints, plus a relaxed OAuth to
 
 Floci strips reserved `floci:*` tags from stored and returned `UserPoolTags` on both create and update paths, so the tag namespace acts as an input-only control channel and is never persisted as user-visible metadata.
 
+An action given a user pool id that does not resolve answers `ResourceNotFoundException` with the live service's own wording, `User pool <poolId> does not exist.`, rather than a generic string, so tooling that matches on Cognito error text behaves the same way locally.
+
 Standalone `TagResource` rejects reserved `floci:*` keys. `ListTagsForResource` and `UntagResource` operate on the persisted user-pool tag map.
 
 ## Supported Actions
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -358,7 +358,7 @@ public class CognitoService implements ResourceProvider {
 
     public UserPool describeUserPool(String id) {
         UserPool pool = poolStore.get(id)
-                .orElseThrow(() -> new AwsException("ResourceNotFoundException", "User pool not found", 400));
+                .orElseThrow(() -> userPoolNotFound(id));
         boolean generatedKeys = ensureJwtSigningKeys(pool);
         boolean generatedSecret = ensureRefreshTokenSecret(pool);
         if (generatedKeys || generatedSecret) {
@@ -1415,7 +1415,7 @@ public class CognitoService implements ResourceProvider {
 
     public CognitoUser adminGetUser(String userPoolId, String username) {
         UserPool pool = poolStore.get(userPoolId).orElseThrow(
-                () -> new AwsException("ResourceNotFoundException", "User pool not found", 400));
+                () -> userPoolNotFound(userPoolId));
         LinkedHashMap<String, CognitoUser> matches = new LinkedHashMap<>();
         userStore.get(userKey(userPoolId, username))
                 .ifPresent(u -> matches.put(u.getUsername(), u));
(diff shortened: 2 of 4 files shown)


bsaurusrex:
Closing as a duplicate of #2936, which was opened first and fixes #2853 the same way, with the same centralised helper.

I have posted the two things this branch had that #2936 does not (the stale `S3WellKnownKeyIntegrationTest` fingerprint, and the docs paragraph) plus the live-API measurements over there, so nothing is lost.

PR #2956, opened 2026-09-02
feat(ec2): implement instant CreateFleet

### LINKED ISSUE #2954: [FEAT] ec2: implement CreateFleet for Karpenter-compatible provisioning
## Service

EC2

## API Action / Feature

CreateFleet, including DryRun validation and the response/error behavior required by Karpenter provisioning.

## AWS Documentation

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html

## Why is this needed?

Floci currently reports CreateFleet as unsupported. Karpenter AWS provider 1.8.8 uses EC2 CreateFleet when it provisions a NodeClaim, including for an on-demand-only NodePool. A CreateFleet DryRun request is therefore enough to expose the current gap and prevents local provisioning from progressing beyond discovery and reconciliation.

The goal is a deterministic local EC2 implementation that follows the AWS Query wire protocol and never contacts or creates resources in real AWS.

Reproduction:

- Run Floci locally with fake AWS credentials and an endpoint such as `http://localhost:4566`.
- Configure Karpenter with a valid EC2NodeClass and a NodePool restricted to `karpenter.sh/capacity-type: on-demand`.
- Trigger a pending workload or invoke the same CreateFleet request through an AWS SDK client.
- The request reaches Floci but returns an unsupported-action error, so no NodeClaim can become launched and no local instance is created.

The implementation should support the request shapes Karpenter emits, including:

- `DryRun` validation with the AWS-compatible `DryRunOperation` response when the request is otherwise valid.
- Target capacity specification for a single instance.
- Launch template configuration and overrides.
- On-demand capacity options and the capacity-type requirements used by Karpenter.
- Subnet, security-group, tagging, and user-data inputs passed through the launch template.
- AWS-compatible success responses and structured fleet errors for invalid or unavailable capacity.

Please add focused Query-protocol tests and an AWS SDK compatibility test. The implementation must remain local and deterministic, preserve the existing no-real-provisioning safety boundary until explicitly enabled, and avoid special-casing Karpenter.

- [x] Yes
- [ ] No

luiseterc:
## Summary

Closes #2954

- Add EC2 Query routing and handling for instant CreateFleet requests.
- Resolve launch templates and Karpenter instance-type, AMI, subnet, security-group, user-data, IAM profile, and instance-tag overrides.
- Reuse Floci's existing local instance lifecycle so DescribeInstances and termination remain consistent.
- Return AWS-compatible DryRunOperation errors without launching instances.
- Group identical launch template and override results into one fleet instance entry with multiple instance IDs.

## Type of change

- [x] New feature (feat:)
- [ ] Bug fix (fix:)
- [ ] Breaking change
- [ ] Docs / chore

## AWS Compatibility

Verified with AWS SDK for Java v2 2.52.0 and the Karpenter 1.8.8 CreateFleet request shape:

- instant fleet type
- LaunchTemplateConfig.N / LaunchTemplateConfigs.N wire forms
- on-demand target capacity
- DryRun=true authorization validation
- fleetInstanceSet and instanceIds response parsing, including grouped IDs for repeated launches

The implementation is deterministic and local-only; it does not contact AWS.

## Checklist

- [ ] `./mvnw test` passes locally (the full suite was not run locally; CI executes it in four sharded test jobs. Focused CreateFleet coverage passes locally.)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
@@ -790,13 +790,31 @@ public class Ec2QueryHandler {
             if (!launchedInstanceIds.isEmpty()) {
                 LOG.warnv("CreateFleet failed after launching instances {0}; rolling them back: {1}",
                         launchedInstanceIds, e.getMessage());
+                List<String> rollbackFailureInstanceIds = new ArrayList<>();
+                List<RuntimeException> rollbackFailures = new ArrayList<>();
                 for (String instanceId : launchedInstanceIds) {
                     try {
                         service.terminateInstances(region, List.of(instanceId));
                     } catch (RuntimeException cleanupFailure) {
-                        LOG.warnv("CreateFleet rollback failed for instance {0}: {1}",
+                        rollbackFailureInstanceIds.add(instanceId);
+                        rollbackFailures.add(cleanupFailure);
+                        LOG.errorv(cleanupFailure, "CreateFleet rollback failed for instance {0}: {1}",
                                 instanceId, cleanupFailure.getMessage());
                     }
+                }
+                if (!rollbackFailures.isEmpty()) {
+                    String rollbackMessage = "CreateFleet rollback incomplete for instance(s): "
+                            + String.join(", ", rollbackFailureInstanceIds);
+                    if (e instanceof AwsException awsFailure) {
+                        AwsException enrichedFailure = new AwsException(
+                                awsFailure.getErrorCode(),
+                                awsFailure.getMessage() + " " + rollbackMessage,
+                                awsFailure.getHttpStatus());
+                        enrichedFailure.addSuppressed(e);
+                        rollbackFailures.forEach(enrichedFailure::addSuppressed);
+                        throw enrichedFailure;
+                    }
+                    rollbackFailures.forEach(e::addSuppressed);
                 }
             }
             throw e;
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
@@ -892,18 +892,18 @@ public class Ec2QueryHandler {
                     break;
                 }
                 hasOverride = true;
-                launches.add(fleetLaunch(template, launchTemplateId, launchTemplateName, version,
+                launches.add(fleetLaunch(region, template, launchTemplateId, launchTemplateName, version,
                         instanceType, imageId, subnetId, availabilityZone, fleetInstanceTags));
             }
             if (!hasOverride) {
-                launches.add(fleetLaunch(template, launchTemplateId, launchTemplateName, version,
+                launches.add(fleetLaunch(region, template, launchTemplateId, launchTemplateName, version,
                         null, null, null, null, fleetInstanceTags));
             }
         }
         return launches;
     }
 
-    private FleetLaunch fleetLaunch(LaunchTemplateData template, String launchTemplateId,
+    private FleetLaunch fleetLaunch(String region, LaunchTemplateData template, String launchTemplateId,
                                     String launchTemplateName, String version,
                                     String overrideInstanceType, String overrideImageId,
                                     String overrideSubnetId, String overrideAvailabilityZone,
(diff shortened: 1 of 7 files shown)


pgermosen (commented):
Nice work on the rollback path in particular - the test that mocks the first termination call itself failing and asserts the second is still attempted is exactly the right way to prove that behavior rather than just asserting it in prose. DryRun validating before the DryRunOperation check is also correctly ordered.

Two things worth addressing before this merges, left inline plus one general note.

On the checklist: "repository-wide Checkstyle is currently blocked by a pre-existing parser error in S3SelectEvaluator.java" doesn't match this repo - there's no Checkstyle plugin or config anywhere in it (checked pom.xml and the tree). Could you clarify what's actually blocking `./mvnw test` locally? As stated it doesn't point to anything real here.

src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
```diff
@@ -710,6 +711,225 @@ private LaunchTemplateData resolveRunInstancesLaunchTemplateData(MultivaluedMap<
...
+            }
+            throw e;
+        }
+
+        XmlBuilder xml = new XmlBuilder()
+                .start("CreateFleetResponse", AwsNamespaces.EC2)
+                .elem("requestId", UUID.randomUUID().toString())
+                .elem("fleetId", "fleet-" + UUID.randomUUID().toString().replace("-", ""))
+                .start("fleetInstanceSet");
+        for (FleetLaunchResult result : results) {
```
pgermosen, line 831: Checked AWS's actual CreateFleet response shape for this: instances sharing the same launch template and overrides get grouped into one CreateFleetInstance entry with multiple ids in instanceIds, not one entry per instance. This builds one <item> per launched instance instead, each with a single-element instanceIds list.

The only compatibility test uses totalTargetCapacity(1), so this can't surface there, but it's exactly the case Karpenter hits routinely - N identical on-demand instances from one launch template. Worth grouping results by (launch template, overrides) before building the XML, and adding a test with capacity > 1 against a single config to pin the grouped shape.

luiseterc: Addressed in c9fc2cc7. CreateFleet now groups results by launch template and override fields, preserving the first-seen order and placing all matching instance IDs in one fleetInstanceSet item. The integration test now launches capacity 2 from one config and asserts one item containing two IDs.

luiseterc:
Addressed the checklist note in the PR description. This repository has no Checkstyle plugin or configuration, so the previous S3SelectEvaluator explanation was inaccurate and has been removed. The focused CreateFleet integration and rollback tests pass locally; the full suite is executed by CI across four shards.

src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
```diff
@@ -710,6 +711,244 @@ private LaunchTemplateData resolveRunInstancesLaunchTemplateData(MultivaluedMap<
...
+            if (!launchedInstanceIds.isEmpty()) {
+                LOG.warnv("CreateFleet failed after launching instances {0}; rolling them back: {1}",
+                        launchedInstanceIds, e.getMessage());
+                for (String instanceId : launchedInstanceIds) {
+                    try {
+                        service.terminateInstances(region, List.of(instanceId));
+                    } catch (RuntimeException cleanupFailure) {
+                        LOG.warnv("CreateFleet rollback failed for instance {0}: {1}",
+                                instanceId, cleanupFailure.getMessage());
+                    }
```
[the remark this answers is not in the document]
luiseterc: Addressed in b06c4b19. Rollback remains best-effort and continues attempting every instance, but failed cleanup is now logged with its exception, attached to the failure diagnostics, and reported to the EC2 caller while preserving the original launch error code and status. The rollback test verifies both behaviors.

pgermosen (approved):
Traced both of the newest fixes rather than just taking the "addressed in..." replies at face value. The rollback-reporting one now enriches the original error with which instances failed cleanup and attaches everything via addSuppressed, with a test verifying both that every instance is still attempted and that the failure surfaces.

The subnet/AZ one was worth double-checking since it's easy for a fix like that to just move the bug: resolveFleetAvailabilityZone now derives the zone from the override's own subnet when no explicit override AZ is given, and the actual conflict check still lives in runInstances, so a genuinely conflicting override still gets rejected. Both cases have their own tests and they coexist correctly.

Nothing else outstanding, good work chasing this all the way through.

Issue #2375, opened 2026-08-18
[FEAT] elasticache: cache parameter groups (aws_elasticache_parameter_group)


ck-michals:
## Service

ElastiCache

## API Action / Feature

`CreateCacheParameterGroup` and the parameter-group surface around it

## AWS Documentation

- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateCacheParameterGroup.html

## Why is this needed?

`aws_elasticache_parameter_group` cannot be created at all, which blocks any Redis module that sets
parameters — most do, if only to set `maxmemory-policy`. The replication group itself works, so
this is the one thing standing between a Terraform Redis module and a clean apply.

## Minimal reproduction

Floci 1.7.0, `floci/floci:1.7.0`:

```
$ aws elasticache create-cache-parameter-group \
    --cache-parameter-group-name repro-pg \
    --cache-parameter-group-family redis7 --description t

An error occurred (UnsupportedOperation) when calling the CreateCacheParameterGroup operation:
Operation CreateCacheParameterGroup is not supported.
```

## Scope

`CreateCacheParameterGroup`, `DescribeCacheParameterGroups`, `ModifyCacheParameterGroup`,
`DescribeCacheParameters` and `DeleteCacheParameterGroup`. Metadata only: the parameters need
storing and returning, not applying to a running Redis.

Worth noting the neighbouring gap while the area is open — `ListTagsForResource` on a
parameter-group ARN returns 404, so tagging a group would fail even once creation works.

## Are you willing to contribute a PR?

- [x] Yes

PR #2957, opened 2026-09-02
feat(elasticache): add validation for cache parameter group existence in replication group operations

### LINKED ISSUE #2303: [FEAT] elasticache: cache parameter group CRUD (DescribeCacheParameterGroups is a hardcoded empty stub)
## Service

ElastiCache

## API Action / Feature

`CreateCacheParameterGroup`, `ModifyCacheParameterGroup`, `DeleteCacheParameterGroup`,
`DescribeCacheParameters` (and a real `DescribeCacheParameterGroups`)

## AWS Documentation

- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateCacheParameterGroup.html
- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_ModifyCacheParameterGroup.html
- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_DeleteCacheParameterGroup.html
- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_DescribeCacheParameters.html

## Why is this needed?

`ElastiCacheQueryHandler` dispatches `DescribeCacheParameterGroups`, but it is a deliberate stub
that always returns an empty list:

```java
private Response handleDescribeCacheParameterGroups(MultivaluedMap<String, String> params) {
    // Cache parameter groups are not modeled by the emulator; return the wire-accurate
    // empty result so SDK clients get a valid 200 instead of an unsupported-action 400.
    var xml = new XmlBuilder().start("CacheParameterGroups").end("CacheParameterGroups");
    ...
}
```

There is no `Create`, `Modify`, `Delete`, or `DescribeCacheParameters`, so
`aws_elasticache_parameter_group` fails on apply.

Parameter groups are how engine settings (`maxmemory-policy`, `notify-keyspace-events`,
`cluster-enabled`, …) are set, so a cache cluster or replication group that references a custom
parameter group cannot be created either.

The empty-list stub is also a silent read: it returns `200` rather than
`CacheParameterGroupNotFoundFault`, so a caller cannot tell a missing group from an unmodelled
feature.

## Scope

Four actions plus replacing the stub with a real describe, backed by a `StorageBackend` created
through `StorageFactory`.

Wire shapes:

- `CreateCacheParameterGroup` — `CacheParameterGroupName`, `CacheParameterGroupFamily`,
  `Description` → returns a `CacheParameterGroup` with `ARN`
- `ModifyCacheParameterGroup` — `ParameterNameValues.member.N.ParameterName` /
  `.ParameterValue` → returns `CacheParameterGroupName`. This is the action the provider calls
  for every `parameter {}` block, and it is called repeatedly rather than once
- `DeleteCacheParameterGroup` — empty response; a group still in use returns
  `InvalidCacheParameterGroupState`
- `DescribeCacheParameters` — returns the stored `Parameters` plus
  `CacheNodeTypeSpecificParameters`; the provider reads this back to detect drift on individual
  parameters
- `DescribeCacheParameterGroups` — an unknown name returns `CacheParameterGroupNotFoundFault`

Filed separately from the cache **subnet** group equivalent, which is the same shape but an
independent resource.

## Docs

Adds new actions, so `make docs-sync` is required.

- [ ] Yes
- [ ] No

matheusandre1:
## Sumary
Closes #2303

Most of the issue was already delivered by #2392 (which closed the sibling #2375): the five
parameter group actions exist, with the ARN, `CacheNodeTypeSpecificParameters`, and
`CacheParameterGroupNotFound` for an unknown name. Two scope items were still open, and this PR
closes them:

- `DeleteCacheParameterGroup` deleted a group that a replication group still referenced. It now
  refuses with `InvalidCacheParameterGroupState` until that replication group is gone.
- `CreateReplicationGroup` accepted a `CacheParameterGroupName` that did not exist. It now refuses
  with `CacheParameterGroupNotFound`, checked with the other validations before a port is taken or
  a container is started, so nothing is left behind.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Both errors are the ones the API Reference documents for these cases:

| Action | Case | Documented error | floci |
| --- | --- | --- | --- |
| `CreateReplicationGroup` | `CacheParameterGroupName` names a group that does not exist | `CacheParameterGroupNotFound` (HTTP 404) | `CacheParameterGroupNotFound`, 404, *CacheParameterGroup &lt;name&gt; not found.* |
| `DeleteCacheParameterGroup` | the group is associated with one or more cache clusters | `InvalidCacheParameterGroupState` (HTTP 400) | `InvalidCacheParameterGroupState`, 400, *One or more cache clusters are still members of this parameter group &lt;name&gt;, so the group cannot be deleted.* |

References:
- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateReplicationGroup.html#API_CreateReplicationGroup_Errors
- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_DeleteCacheParameterGroup.html#API_DeleteCacheParameterGroup_Errors

Deliberate limits:
- "In use" means a stored replication group references the group. Memcached cache clusters do not
  carry a parameter group in floci's model, so they cannot hold one.
- The create-time check runs before provisioning; a parameter group deleted between that check and
  the replication group being stored is not re-checked.
- The message text of `InvalidCacheParameterGroupState` follows the wording AWS uses for
  parameter groups that are still associated with clusters; it was matched against the docs, not
  against a live account.

## Tests

- `ElastiCacheServiceTest`: create with an unknown group is refused before any container starts;
  delete of a referenced group is refused and the group survives, then succeeds once the
  replication group is deleted.
- `ElastiCacheParameterGroupIntegrationTest` (Query protocol, asserting the XML error body):
  `CreateReplicationGroup` with an unknown group returns the 404 and stores nothing;
  `DeleteCacheParameterGroup` of a group held by a replication group returns the 400, the group
  still describes, and the delete succeeds after the replication group is removed. The second case
  needs Docker for the replication group and skips via `Assumptions` without it.
- `docs/services/elasticache.md` documents both behaviours; the generated action table is
  unchanged since no action was added.

## Checklist

### DIFF
--- a/docs/services/elasticache.md
+++ b/docs/services/elasticache.md
@@ -113,6 +113,11 @@ than rejecting names a partial catalogue happens to be missing, which would refu
 AWS accepts. `DescribeCacheParameters` returns those parameters; a request for `system` or
 `engine-default` parameters returns none, and listings are unpaged.
 
+A replication group that names a parameter group is refused with `CacheParameterGroupNotFound` when
+no such group exists, and a parameter group still referenced by a replication group cannot be
+deleted: `DeleteCacheParameterGroup` answers `InvalidCacheParameterGroupState` until that
+replication group is gone.
+
 ## Configuration
 
 | Variable | Default | Description |
--- a/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheService.java
@@ -144,6 +144,9 @@ public class ElastiCacheService implements ResourceProvider {
         // resolved with the other validations, before a port is taken or a container started
         ReplicationGroupSettings resolvedSettings =
                 settings.withKmsKeyId(resolveKmsKeyArn(settings.kmsKeyId(), request.region()));
+        if (request.cacheParameterGroupName() != null && !request.cacheParameterGroupName().isBlank()) {
+            requireParameterGroup(request.cacheParameterGroupName());
+        }
         if (groups.get(groupId).isPresent()) {
             throw new AwsException("ReplicationGroupAlreadyExistsFault",
                     "Replication group " + groupId + " already exists.", 400);
@@ -1143,9 +1146,20 @@ public class ElastiCacheService implements ResourceProvider {
                 throw new AwsException("CacheParameterGroupNotFound",
                         "CacheParameterGroupnot found: " + name, 404);
             }
+            if (isParameterGroupInUse(name)) {
+                throw new AwsException("InvalidCacheParameterGroupState",
+                        "One or more cache clusters are still members of this parameter group "
+                                + name + ", so the group cannot be deleted.", 400);
+            }
             parameterGroups.delete(name);
         }
         LOG.infov("Deleted cache parameter group {0}", name);
+    }
+
+    /** Whether a stored replication group still references the parameter group by that name. */
+    private boolean isParameterGroupInUse(String name) {
+        return groups.scan(key -> true).stream()
+                .anyMatch(group -> name.equals(group.getCacheParameterGroupName()));
     }
 
     @Override
--- a/src/test/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheIntegrationTest.java
@@ -336,7 +336,7 @@ class ElastiCacheIntegrationTest {
                     equalTo(GROUP_ID + "-reused"));
     }
 
-    private static boolean isDockerAvailable() {
+    static boolean isDockerAvailable() {
         try {
             Process process = new ProcessBuilder("docker", "version", "--format", "{{.Server.Version}}")
                     .redirectErrorStream(true)
--- a/src/test/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheParameterGroupIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheParameterGroupIntegrationTest.java
@@ -1,6 +1,7 @@
 package io.github.hectorvent.floci.services.elasticache;
 
 import io.quarkus.test.junit.QuarkusTest;
+import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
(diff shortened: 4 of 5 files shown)


pgermosen (commented):
Both fixes are correct, checked them against the real docs: CreateReplicationGroup documents CacheParameterGroupNotFound for exactly this case, and DeleteCacheParameterGroup documents InvalidCacheParameterGroupState with the same "associated with clusters" wording your check enforces.

One thing needed before merge: the AWS Compatibility section is missing from the PR body entirely, not just thin. This repo squash-merges, so the body becomes the permanent commit message, could you fill that in? What you verified above (the two documented errors) is exactly what belongs there.

Smaller thing: the delete-blocked-while-in-use case only has a service-level test, the create-side check got a full HTTP integration test too (asserting the actual XML error body) but the delete side didn't. Worth a matching case in ElastiCacheParameterGroupIntegrationTest if you have a moment, not blocking otherwise.

matheusandre1:
> Both fixes are correct, checked them against the real docs: CreateReplicationGroup documents CacheParameterGroupNotFound for exactly this case, and DeleteCacheParameterGroup documents InvalidCacheParameterGroupState with the same "associated with clusters" wording your check enforces.
> 
> One thing needed before merge: the AWS Compatibility section is missing from the PR body entirely, not just thin. This repo squash-merges, so the body becomes the permanent commit message, could you fill that in? What you verified above (the two documented errors) is exactly what belongs there.
> 
> Smaller thing: the delete-blocked-while-in-use case only has a service-level test, the create-side check got a full HTTP integration test too (asserting the actual XML error body) but the delete side didn't. Worth a matching case in ElastiCacheParameterGroupIntegrationTest if you have a moment, not blocking otherwise.


Thanks for checking both against the docs. I've rewritten the PR body with a proper AWS
Compatibility section: a table with the documented error for each case, the HTTP status and
message floci returns, links to the two API Reference error sections, and the deliberate limits
(what "in use" covers, the check-then-provision window, and that the message wording was matched
to the docs rather than a live account).

Also added the matching HTTP case for the delete side in ElastiCacheParameterGroupIntegrationTest:
it creates a replication group referencing the group, asserts the
<Code>InvalidCacheParameterGroupState</Code> body and message, checks the group still describes,
and tears the replication group down in a finally so the later delete test is not affected. Since
it needs a real container it skips via Assumptions when Docker is absent, reusing
ElastiCacheIntegrationTest.isDockerAvailable (made package-private) rather than copying it a third
time. Full class passes locally,

pgermosen (approved):
Both round-1 items look good now, the AWS Compatibility table with the doc links is exactly the shape I asked for, and the delete-side integration test properly mirrors the create-side one.

On Greptile's race-condition finding: verified it directly, it's real. createReplicationGroup validates the parameter group before provisioning but doesn't persist the replication group until after container/proxy startup, so a concurrent DeleteCacheParameterGroup landing in that window could delete a parameter group a create depends on. That said, this is the same thing your own "Deliberate limits" note already disclosed, it's not something you missed. Given it needs genuinely concurrent requests against the same parameter-group name to trigger, and the worst case is a dangling metadata reference rather than data loss or a crash, I'm fine merging this as-is with the race tracked separately.

Approving. Opening a follow-up issue for the race so it doesn't get lost.

PR #2959, opened 2026-09-02
feat(route53): update supported operations and add VPC association actions

### LINKED ISSUE #2298: [FEAT] route53: AssociateVPCWithHostedZone, DisassociateVPCFromHostedZone and ListHostedZonesByVPC

matheusandre1:
## Summary

Closes: #2298 

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [X] Docs / chore

## AWS Compatibility

No wire-protocol changes in this PR. The three actions (AssociateVPCWithHostedZone,
DisassociateVPCFromHostedZone, ListHostedZonesByVPC) landed in #2638; this PR closes the
documentation gap that kept #2298 open and registers the Route53 controller with the docs
generator so the action table is maintained by `make docs-sync`.

The endpoint paths and error codes cited in the docs (PublicZoneVPCAssociation,
LastVPCAssociation, ConflictingDomainExists) were checked against the Route 53
2013-04-01 API reference. Behaviour is covered by
Route53VpcAssociationIntegrationTest (raw REST XML, 20 tests). The compatibility suite
in compatibility-tests/sdk-test-java pins AWS SDK for Java 2.52.0.

The three endpoints had already been implemented in the code via PR #2638, but the issue remained open because the documentation was never updated: the actions table in `docs/services/route53.md` still listed 17 operations with no VPC actions, and Route 53 wasn't even registered in the generator used by `make docs-sync`. I only modified the documentation and the generator registration.

### DIFF
--- a/docs/services/index.md
+++ b/docs/services/index.md
@@ -98,7 +98,7 @@ Operation counts are exact. For dispatch-table services (Query and JSON 1.1) eac
 | [AWS FIS](fis.md) | `/experimentTemplates/*`, `/experiments/*`, `/actions/*`, `/targetResourceTypes/*`, `/safetyLevers/*`, `/tags/*` | REST JSON | 26 |
 | [CodeGuru Reviewer](codegurureviewer.md) | `/associations`, `/associations/{associationArn}`, `/tags/*` | REST JSON | 7 |
 | [CloudFront](cloudfront.md) | `/2020-05-31/distribution/*`, `/2020-05-31/cache-policy/*`, `/2020-05-31/function/*` | REST XML | 50 |
-| [Route53](route53.md) | `/2013-04-01/hostedzone/*`, `/2013-04-01/healthcheck/*`, `/2013-04-01/change/*` | REST XML | 17 |
+| [Route53](route53.md) | `/2013-04-01/hostedzone/*`, `/2013-04-01/healthcheck/*`, `/2013-04-01/change/*` | REST XML | 25 |
 | [Route 53 Resolver](route53resolver.md) | `POST /` + `X-Amz-Target: Route53Resolver.*` | JSON 1.1 | 18 |
 | [Cloud Map](cloudmap.md) | `POST /` + `X-Amz-Target: Route53AutoNaming_v20170314.*` | JSON 1.1 | 22 |
 | [AWS Config](config.md) | `POST /` + `X-Amz-Target: StarlingDoveService.*` | JSON 1.1 | 33 |
--- a/docs/services/route53.md
+++ b/docs/services/route53.md
@@ -4,25 +4,35 @@ Route53 management-plane emulation. Supports hosted zones, resource record sets,
 
 ## Supported Operations
 
-| Operation | Method | Path |
-|---|---|---|
-| CreateHostedZone | POST | `/2013-04-01/hostedzone` |
-| GetHostedZone | GET | `/2013-04-01/hostedzone/{Id}` |
-| DeleteHostedZone | DELETE | `/2013-04-01/hostedzone/{Id}` |
-| ListHostedZones | GET | `/2013-04-01/hostedzone` |
-| ListHostedZonesByName | GET | `/2013-04-01/hostedzonesbyname` |
-| GetHostedZoneCount | GET | `/2013-04-01/hostedzonecount` |
-| ChangeResourceRecordSets | POST | `/2013-04-01/hostedzone/{Id}/rrset` |
-| ListResourceRecordSets | GET | `/2013-04-01/hostedzone/{Id}/rrset` |
-| GetChange | GET | `/2013-04-01/change/{Id}` |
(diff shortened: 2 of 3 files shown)


pgermosen (commented):
Went through this carefully rather than trusting the table at face value: read Route53Controller.java directly and cross-checked all 25 rows against the actual @GET/@POST/@Path annotations, everything matches exactly, in the same order as the source, which is a good sign this came from the generator rather than being hand-typed to look right. Also checked the three error codes the new docs cite (PublicZoneVPCAssociation, LastVPCAssociation, ConflictingDomainExists) against the real Route 53 API reference, all three match word for word.

One thing I can't confirm myself since it needs a build tool: whether make docs-check was actually run given those floci:actions markers mean that table's meant to be generated. Could you run that and confirm, or paste the output if you already have it?

Also, the PR body is missing both the AWS Compatibility and Checklist sections entirely. Since this repo squash-merges, that body becomes the permanent commit message, worth filling in before merge.

matheusandre1:
> Went through this carefully rather than trusting the table at face value: read Route53Controller.java directly and cross-checked all 25 rows against the actual @GET/@POST/@path annotations, everything matches exactly, in the same order as the source, which is a good sign this came from the generator rather than being hand-typed to look right. Also checked the three error codes the new docs cite (PublicZoneVPCAssociation, LastVPCAssociation, ConflictingDomainExists) against the real Route 53 API reference, all three match word for word.
> 
> One thing I can't confirm myself since it needs a build tool: whether make docs-check was actually run given those floci:actions markers mean that table's meant to be generated. Could you run that and confirm, or paste the output if you already have it?
> 
> Also, the PR body is missing both the AWS Compatibility and Checklist sections entirely. Since this repo squash-merges, that body becomes the permanent commit message, worth filling in before merge.

The output is silent because the Makefile targets are prefixed with @ and only print when something fails.

pgermosen (approved):
Found the CI job that settles the docs-check question: "Docs in sync" runs make docs-check (and a strict mkdocs build) on any change under src/main, docs, or tools/docs, and it's green on this PR. That's the confirmation I needed, no further action there.

One thing for next time: the PR template's Checklist section is still missing (AWS Compatibility is filled in now, thanks). Since this repo squash-merges, the PR body becomes the permanent commit message on main, so it's worth going through the whole template rather than partially, even for a docs-only change like this one.

Approving.

PR #2962, opened 2026-09-02
fix(cognito): fail listUsers/listUserPoolClients on an absent user pool

### LINKED ISSUE #2952: [BUG] cognito: ListUsers and ListUserPoolClients answer an absent user pool with an empty list
## Bug Description

`ListUsers` and `ListUserPoolClients` do not resolve the user pool at all. Given a pool id that does not exist they return `200` with an empty list, where the live service returns `ResourceNotFoundException`. Sibling list actions on the same service do resolve the pool, so the behaviour is inconsistent within Cognito itself.

The practical cost is that a typo in a pool id, or a pool that was never created, reads as "this pool is empty" rather than as an error. Test code asserting that a user was created will pass against a pool that does not exist.

## Steps to Reproduce

```bash
EP=http://localhost:4599
P=us-east-1_ZZZZZZZZZ

aws --endpoint-url $EP cognito-idp list-users              --user-pool-id $P
aws --endpoint-url $EP cognito-idp list-user-pool-clients  --user-pool-id $P
aws --endpoint-url $EP cognito-idp list-groups             --user-pool-id $P
aws --endpoint-url $EP cognito-idp list-resource-servers   --user-pool-id $P --max-results 10
```

## Actual Behaviour

Measured on `main` at `54654760`, built from `docker/Dockerfile`:

```
list-users              -> {"Users": []}
list-user-pool-clients  -> {"UserPoolClients": []}
list-groups             -> ResourceNotFoundException: User pool not found
list-resource-servers   -> ResourceNotFoundException: User pool not found
```

The first two are the divergence. The last two already resolve the pool and are correct apart from the message, which is #2853.

## Expected Behaviour

Measured against the live service in `ap-southeast-1`, AWS CLI v2, on a pool id that does not exist:

```
```

All four are identical on AWS.

## Notes

Found while measuring #2853, which fixes the message on the paths that already resolve the pool and deliberately leaves this alone: adding an existence check changes a success into an error, which is a different kind of change from rewording one, and it deserves its own review.

`listUsers` and `listUserPoolClients` scan their stores by a `{poolId}::` prefix and by a `userPoolId` field respectively, neither of which touches `poolStore`, which is why an absent pool is indistinguishable from an empty one. The fix is presumably a `describeUserPool(userPoolId)` call at the top of each, matching what the group and resource server paths already do, but I have not written it and have not checked whether any existing test depends on the empty-list behaviour.

## Environment

- Floci version / image tag: built from `main` at `54654760`
- Java SDK version (if applicable): n/a, AWS CLI v2
- How you are running Floci: Docker, `docker/Dockerfile`

Mide69:
Fixes #2952

listUsers scanned its store by a "{poolId}::" prefix and listUserPoolClients by a userPoolId field, neither of which touched poolStore, so an absent pool was indistinguishable from an empty one: both returned 200 with an empty list instead of ResourceNotFoundException. list-groups and list-resource-servers already resolve the pool via describeUserPool and are correct; this brings the other two list actions in line with them and with real Cognito, where all four behave identically.

Verified by temporarily disabling each fix and confirming the corresponding new test fails with the exact expected error before restoring it. Full CognitoServiceTest suite (177 tests) passes clean.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -594,6 +594,7 @@ public class CognitoService implements ResourceProvider {
     }
 
     public List<UserPoolClient> listUserPoolClients(String userPoolId) {
+        describeUserPool(userPoolId);
         return clientStore.scan(k -> clientStore.get(k).map(c -> c.getUserPoolId().equals(userPoolId)).orElse(false));
     }
 
@@ -1604,6 +1605,7 @@ public class CognitoService implements ResourceProvider {
     }
 
     public List<CognitoUser> listUsers(String userPoolId, String filter) {
+        describeUserPool(userPoolId);
         String prefix = userPoolId + "::";
         List<CognitoUser> all = userStore.scan(k -> k.startsWith(prefix));
         if (filter == null || filter.isBlank()) {
--- a/src/test/java/io/github/hectorvent/floci/services/cognito/CognitoServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cognito/CognitoServiceTest.java
@@ -1950,6 +1950,43 @@ class CognitoServiceTest {
         assertTrue(result.isEmpty());
     }
 
+    // =========================================================================
+    // Issue #2952 - listUsers/listUserPoolClients against an absent user pool
+    // =========================================================================
+
+    @Test
+    void listUsersAgainstAnAbsentUserPoolFails() {
+        // listUsers scanned by a "{poolId}::" prefix without ever touching poolStore, so an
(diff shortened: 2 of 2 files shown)


pgermosen (approved):
Clean fix. Confirmed `describeUserPool` here is the exact same pattern `listGroups`/`listResourceServers` already use, and checked it against the real API docs, `ResourceNotFoundException` on a missing pool is documented for both [ListUsers](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListUsers.html) and [ListUserPoolClients](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListUserPoolClients.html).

Nice touch adding the empty-pool-still-returns-empty tests alongside the absent-pool ones, that's exactly the case that would've caught a sloppier version of this fix.

PR #2963, opened 2026-09-02
fix(stepfunctions): resolve the States.Format template argument like every other argument

### LINKED ISSUE #2927: [BUG] Step Functions: States.Format template argument is taken literally
## Service

Step Functions

## AWS API Action

Intrinsic functions, the format template argument of `States.Format`

## Expected behavior

AWS and Step Functions Local resolve a reference path written in the template position. A template
path that matches nothing fails the execution with `States.Runtime`.

## Actual behavior

Floci takes the template literally and succeeds with the path itself as the value.

```json
{"status":"SUCCEEDED","output":"{\"v\":\"$.nope\"}"}
```

`AslExecutor.applyIntrinsic` reads the first argument with `unquoteString(parts.get(0))`. Every
other argument goes through `resolveIntrinsicArg`, which resolves `$.` paths and raises the missing
argument error added in #2872. The template position skips both.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 stepfunctions create-state-machine \
  --name fmt-template \
  --role-arn arn:aws:iam::000000000000:role/service-role/x \
  --definition '{"StartAt":"P","States":{"P":{"Type":"Pass","Parameters":{"v.$":"States.Format($.nope, 1)"},"End":true}}}'

aws --endpoint-url http://localhost:4566 stepfunctions start-sync-execution \
  --state-machine-arn arn:aws:states:us-east-1:000000000000:stateMachine:fmt-template \
  --input '{"other":1}'
```

## Environment

- Floci version / image tag: 2.0.0
- Java SDK version (if applicable): n/a
- How you're running Floci (Docker / native / `mvn quarkus:dev`): `mvn quarkus:dev`

## Notes

Current behavior is pinned by
`AslExecutorIntrinsicMissingArgumentTest#aReferencePathAsTheFormatTemplateIsStillTakenLiterally`.
Update that test when this is fixed.

Split out of #2872 on review.

Mide69:
Fixes #2927

AslExecutor.applyIntrinsic read the format template with unquoteString(parts.get(0)), which only stripped surrounding quotes and otherwise took the argument literally. Every other argument goes through resolveIntrinsicArg, which resolves $. paths and raises the missing-argument error added in #2872. A reference path written in the template position skipped both, so States.Format($.nope, 1) succeeded with the literal string "$.nope" instead of resolving it.

The template argument now goes through resolveIntrinsicArg like the rest, so a template path that matches nothing fails with States.Runtime using the same cause format every other missing argument already produces, and a template path that does resolve is used as the template.

Removed the now-unused unquoteString helper.

Verified by temporarily reverting the fix and confirming both new tests fail with the exact expected error before restoring it. Full stepfunctions package regression (709 tests) passes clean.


pgermosen (commented):
Checked this against AWS docs and against the rest of the codebase rather than just the description. The fix itself is clean: the template argument now goes through the same resolveIntrinsicArg every other argument uses, and for the common case (a quoted literal template) that returns the same unquoted TextNode unquoteString used to produce, so there's no regression for the normal 'States.Format(\'...\', $.x)' usage. Confirmed unquoteString has no other callers left, so removing it is safe.

On the error message: "The JsonPath argument for the field '...' could not be found in the input '...'" isn't new here, it's the exact string already used and tested elsewhere in this same test file for other intrinsic arguments (from #2872), so this fix is genuinely reusing an already-verified AWS message rather than introducing a new one. AWS's own intrinsic-functions docs confirm the general shape (JsonPath arguments resolved against the state's input), consistent with what's implemented.

One hygiene note: this PR skipped the repo's PR template entirely, no Type of change checkbox, no AWS Compatibility section, no Checklist. The prose covers most of that ground informally (the revert-and-confirm-fail verification, the 709-test regression run), so I'm not blocking on it, just flagging it so the template gets used going forward.

Mide69:
Thanks for the thorough independent check, appreciated. Good call on the template, missed it since I passed --body directly via gh pr create rather than going through the editor. Will use it properly on the next one.

pgermosen (commented):
The one failing check, native arm64 / sdk-test-node, is tests/cloudwatch.test.ts > CloudWatch Metrics > should put metric data with statistic values (AssertionError: expected 0 to be greater than 0). This PR only touches AslExecutor.java and its own test, nothing CloudWatch-related, so this looks like unrelated flakiness rather than something this change caused.

Can you rebase onto main and push again just to get a clean run and rule that out for sure? Everything else here still looks good from the earlier review.

hectorvent (approved):
Seconding pgermosen's review as the maintainer: the template position going through `resolveIntrinsicArg` like every other argument is the right fix, and removing `unquoteString` with its last caller keeps the intrinsic path to one resolver.

On the red `sdk-test-node` check: your head already carries main, so there was nothing to rebase. It was the CloudWatch `put metric data with statistic values` assertion, unrelated to this change, and the rerun came back green.

No blockers from my side.

PR #2964, opened 2026-09-02
fix(cloudformation): persist FunctionResponseTypes on Lambda::EventSourceMapping

### LINKED ISSUE #2848: [BUG] CloudFormation AWS::Lambda::EventSourceMapping does not persist FunctionResponseTypes (CFN path not covered by #208)
## Service
  
  Lambda (Event Source Mappings), CloudFormation, SQS
  
##  AWS API Action
  
  CloudFormation::CreateStack provisioning AWS::Lambda::EventSourceMapping with FunctionResponseTypes: ["ReportBatchItemFailures"]
   (verified via Lambda::ListEventSourceMappings / GetEventSourceMapping)
  
## Expected behavior
  
When an AWS::Lambda::EventSourceMapping is deployed via CloudFormation with FunctionResponseTypes: ["ReportBatchItemFailures"], the created mapping should report FunctionResponseTypes: ["ReportBatchItemFailures"] — matching real AWS, and matching what floci's own API-level CreateEventSourceMapping already returns after #208 (which closed #202).
  
## Actual behavior
  
The CloudFormation-provisioned mapping comes back with FunctionResponseTypes: [], even though the template correctly declares ["ReportBatchItemFailures"]. As a result, partial batch responses are ignored for CFN-deployed SQS→Lambda mappings: failed messages returned via batchItemFailures are deleted on the 200 response instead of being retried.
  
Note: this is a separate, narrower gap than #202 / #208, which only fixed the API CreateEventSourceMapping path (and the poller).

The CloudFormation resource path was never covered and still drops the property. Observed side by side on 1.7.0:
  
template : ["ReportBatchItemFailures"]   # what the CFN template declares
CFN      : []                            # mapping floci created from that template  (BUG)
API      : ["ReportBatchItemFailures"]   # control: created via CreateEventSourceMapping (works)
  
## Reproduction
  
This python script deploys a hand-written CloudFormation template, then compares the CFN-created mapping against an API-created control.

```
pip install boto3
python repro.py

repro.py:

import os, sys, json, time
import boto3
from botocore.config import Config
from botocore.exceptions import ClientError

ENDPOINT = os.environ.get("AWS_ENDPOINT_URL", "http://localhost:4566")
REGION   = os.environ.get("AWS_DEFAULT_REGION", "us-east-1")
ACCOUNT  = "000000000000"

STACK_NAME    = "floci-esm-repro-py"
FN_NAME       = "floci-esm-repro-py-fn"
QUEUE_NAME    = "floci-esm-repro-py-queue"
CONTROL_QUEUE = "floci-esm-repro-py-control-q"
ROLE_ARN      = f"arn:aws:iam::{ACCOUNT}:role/lambda-role"
REQUIRED      = "ReportBatchItemFailures"

cfg = Config(retries={"max_attempts": 5, "mode": "standard"})
def client(svc):
  return boto3.client(svc, endpoint_url=ENDPOINT, region_name=REGION, config=cfg)

CODE_SRC = (
  "exports.handler = async (event) => {\n"
  "  const failures = (event.Records || []).map(r => ({ itemIdentifier: r.messageId }));\n"
  "  return { batchItemFailures: failures };\n"
  "};\n"
)

TEMPLATE = {
  "Resources": {
      "ReproQueue": {
          "Type": "AWS::SQS::Queue",
          "Properties": {"QueueName": QUEUE_NAME, "VisibilityTimeout": 30},
      },
      "ReproFn": {
          "Type": "AWS::Lambda::Function",
          "Properties": {
              "FunctionName": FN_NAME, "Runtime": "nodejs22.x",
              "Handler": "index.handler", "Role": ROLE_ARN, "Timeout": 30,
              "Code": {"ZipFile": CODE_SRC},
          },
      },
      "ReproEsm": {
          "Type": "AWS::Lambda::EventSourceMapping",
          "Properties": {
              "FunctionName": {"Ref": "ReproFn"},
              "EventSourceArn": {"Fn::GetAtt": ["ReproQueue", "Arn"]},
              "BatchSize": 1,
              "FunctionResponseTypes": [REQUIRED],   # <-- property under test
          },
      },
  },
}

def wait_stack(cfn, name, targets, timeout=120):
  deadline = time.time() + timeout
  while time.time() < deadline:
      try:
          st = cfn.describe_stacks(StackName=name)["Stacks"][0]["StackStatus"]
      except ClientError:
          return None
      if st in targets or st.endswith("FAILED") or st.endswith("ROLLBACK_COMPLETE"):
          return st
      time.sleep(2)
  return "TIMEOUT"

def cleanup(cfn, lam, sqs):
  try:
      cfn.delete_stack(StackName=STACK_NAME)
      wait_stack(cfn, STACK_NAME, {"DELETE_COMPLETE"})
  except ClientError:
      pass
  try:
      for m in lam.list_event_source_mappings(FunctionName=FN_NAME).get("EventSourceMappings", []):
          try: lam.delete_event_source_mapping(UUID=m["UUID"])
          except ClientError: pass
  except ClientError:
      pass
  try:
      url = sqs.get_queue_url(QueueName=CONTROL_QUEUE)["QueueUrl"]
      sqs.delete_queue(QueueUrl=url)
  except ClientError:
      pass

def main():
  cfn, lam, sqs = client("cloudformation"), client("lambda"), client("sqs")
  cleanup(cfn, lam, sqs)

  print(f"==> Deploying CloudFormation stack '{STACK_NAME}' to {ENDPOINT}")
  cfn.create_stack(StackName=STACK_NAME, TemplateBody=json.dumps(TEMPLATE),
                   Capabilities=["CAPABILITY_NAMED_IAM"])
  status = wait_stack(cfn, STACK_NAME, {"CREATE_COMPLETE"})
  print(f"    stack status: {status}")
  if status != "CREATE_COMPLETE":
      return 2

  template_rt = TEMPLATE["Resources"]["ReproEsm"]["Properties"]["FunctionResponseTypes"]
  mappings = lam.list_event_source_mappings(FunctionName=FN_NAME).get("EventSourceMappings", [])
  cfn_rt = mappings[0].get("FunctionResponseTypes", []) if mappings else None

  sqs.create_queue(QueueName=CONTROL_QUEUE)
  ctrl_arn = sqs.get_queue_attributes(
      QueueUrl=sqs.get_queue_url(QueueName=CONTROL_QUEUE)["QueueUrl"],
      AttributeNames=["QueueArn"],
  )["Attributes"]["QueueArn"]
  api_rt = lam.create_event_source_mapping(
      FunctionName=FN_NAME, EventSourceArn=ctrl_arn, BatchSize=1,
      FunctionResponseTypes=[REQUIRED],
  ).get("FunctionResponseTypes", [])

  print("\n======================= RESULT =======================")
  print(f"  template : {json.dumps(template_rt)}")
  print(f"  CFN      : {json.dumps(cfn_rt)}")
  print(f"  API      : {json.dumps(api_rt)}")
  print("======================================================")

  reproduced = cfn_rt == [] and api_rt == [REQUIRED]
  cleanup(cfn, lam, sqs)

  if reproduced:
      print("BUG REPRODUCED: CFN dropped FunctionResponseTypes; API create honored it.")
      return 0
  print(f"Bug NOT reproduced (CFN={cfn_rt!r}, API={api_rt!r}).", file=sys.stderr)
  return 1

if __name__ == "__main__":
  sys.exit(main())
```

Observed output on floci 1.7.0:

```
==> Deploying CloudFormation stack 'floci-esm-repro-py' to http://localhost:4566
  stack status: CREATE_COMPLETE

======================= RESULT =======================
template : ["ReportBatchItemFailures"]
CFN      : []
API      : ["ReportBatchItemFailures"]
======================================================
```

## Environment

- Floci version / image tag: floci/floci:latest — 1.7.0
- Java SDK version (if applicable): n/a (reproduced with Python boto3; also observed via AWS CDK aws-cdk-lib 2.262.0)
- How you're running Floci (Docker / native / mvn quarkus:dev): Docker (Docker Desktop on macOS), docker compose up

Mide69:
Fixes #2848

provisionLambdaEventSourceMapping never read the FunctionResponseTypes property, so a CFN-provisioned mapping always came back with an empty list even when the template declared it, unlike the direct CreateEventSourceMapping API path (fixed by #208). Partial batch responses were ignored for CFN-deployed SQS to Lambda mappings: failed messages returned via batchItemFailures were deleted on the 200 response instead of being retried.

Resolved the same way every other array property in this method already is, via the existing resolveStringList helper.

Verified by temporarily disabling the fix and confirming the extended integration test fails with the exact expected empty list before restoring it.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -4117,6 +4117,11 @@ public class CloudFormationResourceProvisioner {
             }
         }
 
+        List<String> functionResponseTypes = resolveStringList(props, "FunctionResponseTypes", engine);
+        if (!functionResponseTypes.isEmpty()) {
+            req.put("FunctionResponseTypes", functionResponseTypes);
+        }
+
         var esm = lambdaService.createEventSourceMapping(region, req);
         r.setPhysicalId(esm.getUuid());
         r.getAttributes().put("Id", esm.getUuid());
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationIntegrationTest.java
@@ -4403,7 +4403,8 @@ class CloudFormationIntegrationTest {
                     "FunctionName": { "Ref": "MyFunction" },
                     "EventSourceArn": { "Fn::GetAtt": ["MyQueue", "Arn"] },
                     "Enabled": true,
-                    "BatchSize": 5
+                    "BatchSize": 5,
+                    "FunctionResponseTypes": ["ReportBatchItemFailures"]
                   }
                 }
               }
@@ -4456,6 +4457,15 @@ class CloudFormationIntegrationTest {
 
         JsonNode esmList = OBJECT_MAPPER.readTree(esmJson);
         String esmUuid = esmList.path("EventSourceMappings").get(0).path("UUID").asText();
+
+        // github.com/floci-io/floci/issues/2848: the CloudFormation path never read
+        // FunctionResponseTypes, so a CFN-provisioned mapping always came back with an empty
+        // list even though the template declared it and the direct CreateEventSourceMapping API
+        // path already honored it.
+        JsonNode responseTypes = esmList.path("EventSourceMappings").get(0).path("FunctionResponseTypes");
+        assertTrue(responseTypes.isArray() && responseTypes.size() == 1
+                        && "ReportBatchItemFailures".equals(responseTypes.get(0).asText()),
+                "expected FunctionResponseTypes to carry through from the template but was: " + responseTypes);
 
         // 5. Delete stack and verify ESM is gone
         given()


hectorvent:
Thank you, this is the right shape: the property read the same way its sibling arrays in that arm already are, and the integration test pins exactly the field that was being dropped.

I checked the registry schema and the Lambda model: `FunctionResponseTypes` is a list of at most one `ReportBatchItemFailures` and it is not create-only, so passing it straight to `createEventSourceMapping`, which already honours it, is the whole fix.

**(follow-up, separate PR)** The PR body skips the repo template (type of change, checklist). Not needed for a change this small, just worth using next time so the squash message carries the type.

No blockers from my side.

hectorvent (approved):
Approving; notes above.

PR #2969, opened 2026-09-02
fix(stepfunctions): abort executions abandoned by a restart

### LINKED ISSUE #2965: [BUG] Executions left RUNNING forever after a restart
## Service

Step Functions

## AWS API Action

`DescribeExecution`, `GetExecutionHistory`

## Expected behavior

AWS has no equivalent of restarting the service, so it cannot arbitrate this case directly. What it does show is that every path out of `RUNNING` writes a terminal status, and which shape that status takes.

`StopExecution` with no `--error` and no `--cause` leaves `status: ABORTED`, and `DescribeExecution` **omits** the `error` and `cause` keys rather than returning them as null. The terminal event carries its details object present but empty:

```json
{"type": "ExecutionAborted", "id": 3, "previousEventId": 0, "executionAbortedEventDetails": {}}
```

A machine-level `TimeoutSeconds` that expires leaves `status: TIMED_OUT`, `stopDate` exactly `startDate` plus the timeout, and an event carrying `error` but no `cause` key:

```json
{"type": "ExecutionTimedOut", "id": 3, "previousEventId": 0, "executionTimedOutEventDetails": {"error": "States.Timeout"}}
```

Both measured against real AWS.

## Actual behavior

`EmulatorLifecycle.onStart` restores SQS, Kinesis, DynamoDB Streams, Pipes, RDS, ElastiCache and ELB from storage, but not Step Functions. An execution persisted as `RUNNING` stays `RUNNING` forever with no worker driving it. `DescribeExecution` keeps reporting `RUNNING`, and the history stops at the last event written before the restart, so the symptom does not say what happened.

A test suite that waits on such an execution hangs until its own timeout.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 stepfunctions create-state-machine \
  --name waiter --role-arn arn:aws:iam::000000000000:role/r \
  --definition '{"StartAt":"W","States":{"W":{"Type":"Wait","Seconds":600,"End":true}}}'

aws --endpoint-url http://localhost:4566 stepfunctions start-execution \
  --state-machine-arn <arn> --name e1

docker restart floci

aws --endpoint-url http://localhost:4566 stepfunctions describe-execution \
  --execution-arn <arn>
# still RUNNING, and it stays RUNNING forever
```

## Environment

facuparedes:
Closes #2965

An execution persisted as `RUNNING` survives a restart with no worker driving it, so `DescribeExecution` reports `RUNNING` forever and the history stops at the last event written before the restart.

`EmulatorLifecycle.onStart` now sweeps Step Functions executions still marked `RUNNING` and closes each one exactly the way AWS closes an abort with no reason, which is the shape the issue records from a measured `StopExecution`:

- `status: ABORTED`
- `DescribeExecution` omits the `error` and `cause` keys rather than sending them as null
- the terminal event carries `executionAbortedEventDetails` present but empty

A restart has no AWS counterpart, so nothing here invents a status or an error code AWS cannot emit. The reason the execution ended lives in the WARN the sweep logs, not in the API response.

A suite waiting on such an execution now fails on a terminal status instead of hanging until its own timeout.

5 commits, 5 files, test-first.

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -417,12 +417,16 @@ from is empty and the state is never reached.
 An execution runs in the Floci process. When Floci restarts while an execution is `RUNNING`,
 no worker survives it, so at startup every execution still stored as `RUNNING` is aborted:
 `DescribeExecution` reports `status` `ABORTED`, a `stopDate`, `error` `ExecutionAbandoned` and
-`cause` `The emulator restarted while this execution was RUNNING; no worker survived it.`, and
-`GetExecutionHistory` reports a single `ExecutionAborted` event carrying the same error and cause.
-Executions of every account are swept, each written back under its own account.
+`cause` `The emulator restarted while this execution was RUNNING; no worker survived it.`
+Executions of every account are swept, each written back under its own account. Executions that
+already reached a terminal status are left untouched, and so is the status, error and cause of one
+this sweep aborted on an earlier boot.
 
-Execution histories are held in memory, so the events recorded before the restart are gone and the
-execution cannot be resumed. Executions that already reached a terminal status are left untouched.
+Execution histories are held in memory, not in storage. The events recorded before the restart are
+gone, so the execution cannot be resumed, and `GetExecutionHistory` reports a single
+`ExecutionAborted` event carrying the same error and cause only for the boot that aborted it: after
+a further restart the execution is already terminal, no event is written, and the history is empty
+while `DescribeExecution` still reports the status, error and cause.
 
 ## Configuration
 
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -412,6 +412,18 @@ Local, the execution starts and fails with `States.Runtime` only if the state th
 it is entered. This keeps a generated mock file usable when the collection it was built
 from is empty and the state is never reached.
 
+## Executions abandoned by a restart
+
+An execution runs in the Floci process. When Floci restarts while an execution is `RUNNING`,
+no worker survives it, so at startup every execution still stored as `RUNNING` is aborted:
+`DescribeExecution` reports `status` `ABORTED`, a `stopDate`, `error` `ExecutionAbandoned` and
+`cause` `The emulator restarted while this execution was RUNNING; no worker survived it.`, and
+`GetExecutionHistory` reports a single `ExecutionAborted` event carrying the same error and cause.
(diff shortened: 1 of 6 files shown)


facuparedes:
This one was real and is already closed on this branch.

`ExecutionAbandoned` existed at `8533b0ed`, where the startup sweep passed a constant error and cause. The head commit `81cd6ff5` removed it, and the string is gone from the branch:

```
$ grep -rn --include='*.java' "ExecutionAbandoned" src/
(no output)
```

The sweep now passes both as null, `StepFunctionsService:713`:

```java
if (!markAborted(entry.key(), entry.value(), null, null)) {
```

and `StepFunctionsJsonHandler:289-290` omits the keys rather than emitting them as null, which is what AWS returns for `StopExecution` called without a reason:

```java
if (exec.getError() != null) response.put("error", exec.getError());
if (exec.getCause() != null) response.put("cause", exec.getCause());
```

`GetExecutionHistory` seals the terminal event with an empty details map, so `executionAbortedEventDetails` is present and empty rather than carrying an abort reason.

Both shapes are asserted:

```
StepFunctionsServicePersistenceTest.abandonedRunningExecutionIsAbortedOnRestart
  assertNull(reloaded.getError());
  assertNull(reloaded.getCause());
  assertEquals(Map.of(), aborted.getDetails());

StepFunctionsStopExecutionIntegrationTest.abortWithNoReasonOmitsErrorAndCause
  assertFalse(described.containsKey("error"));
  assertFalse(described.containsKey("cause"));
```

```
./mvnw test -Dtest=StepFunctionsServicePersistenceTest → 3 tests, 0 failures
```

facuparedes:
Clean against `main` and all checks pass. The one review finding here describes an intermediate commit and is answered above, so nothing is outstanding from my side.

This shares no file with #2970, so it does not need to wait on it. It overlaps #2972 in `StepFunctionsService.java` and `docs/services/step-functions.md`, but the two do not conflict, so whichever lands second is still a clean merge.

hectorvent (approved):
Thank you, this is exactly the shape #2965 measured: ABORTED with `error` and `cause` omitted rather than null, and an empty `executionAbortedEventDetails` on the terminal event, filed under the account that owns the execution since startup has no request context. Sharing `markAborted` with `StopExecution` keeps the two paths from drifting.

No blockers from my side.

PR #2970, opened 2026-09-02
fix(cloudformation): reconcile AWS::Pipes::Pipe on UpdateStack

### LINKED ISSUE #2966: [BUG] UpdateStack over an AWS::Pipes::Pipe fails instead of reconciling it
## Service

CloudFormation, EventBridge Pipes

## AWS API Action

`UpdateStack` over `AWS::Pipes::Pipe`

## Expected behavior

CloudFormation reconciles the pipe in place. The registry schema is the authoritative source and says so:

```bash
aws cloudformation describe-type --type RESOURCE --type-name AWS::Pipes::Pipe --query Schema
```

`createOnlyProperties` is `["/properties/Name", "/properties/Source"]`, `primaryIdentifier` is `/properties/Name`, and `handlers` declares `update`. So changing any property other than `Name` or `Source` updates the existing pipe and keeps its physical id.

## Actual behavior

`PipesCfnProvisioner.provision` calls `createPipe` unconditionally, ignoring `ProvisionContext.priorPhysicalId()`. On the second pass `PipesService.createPipe` raises `ConflictException`, so the whole stack update fails.

Any template change to a stack that contains a pipe therefore forces a full teardown and redeploy.

## Reproduction

Deploy a stack containing an `AWS::Pipes::Pipe` with an SQS source and an SQS target, then deploy the same stack again with any property changed. The second `UpdateStack` fails with `ConflictException`.

## Environment

facuparedes:
Closes #2966

`PipesCfnProvisioner.provision` called `createPipe` unconditionally and ignored `ProvisionContext.priorPhysicalId()`, so the second pass hit `ConflictException` and the whole stack update failed.

It now reconciles the existing pipe when a prior physical id is present, following the registry schema: `createOnlyProperties` is `["/properties/Name", "/properties/Source"]` and `primaryIdentifier` is `/properties/Name`, so any other property change updates in place and keeps the physical id. Tags are reconciled too, and a failed lookup reports the real fault instead of falling through to a create.

Verified with `aws cloudformation describe-type --type RESOURCE --type-name AWS::Pipes::Pipe`; the issue quotes the schema fields.

Opening the update path opened two more, both of which CloudFormation documents and both of which are closed here.

**A failed update rolls the pipe back.** "If a resource update fails, CloudFormation rolls back any resources that it has updated during the upgrade to their configurations before the update." The provisioner snapshots the pipe before `updatePipe` and restores it from a `rollbackUpdate` hook, so an extracted provisioner owns its own restore instead of adding an arm to `CloudFormationResourceProvisioner`. The restore goes through a new `PipesService.restorePipe`, not `updatePipe`: `UpdatePipe` reads an omitted property as "leave it alone", which is the AWS API contract, while a rollback has to put the pipe back to exactly its pre-update configuration, clearing what the failed update added. Both meanings now live in one writer with two named entry points.

**A rename deletes the displaced pipe in the committed cleanup phase, not inside `provision`.** "CloudFormation makes three attempts at deleting the old resource. If CloudFormation can't delete the old resource, it removes the old resource from the stack and continues updating the stack. When the stack update is complete, CloudFormation issues an `UPDATE_COMPLETE` stack event, but includes a `StatusReason` that states that one or more resources couldn't be deleted."

`CloudFormationService` already implements that paragraph in full, for Step Functions only, because its four entry points read one hardcoded attribute. `CfnResourceProvisioner` gains `default` hooks for `updateCleanupPhysicalId`, `completeUpdate`, `hasReplacementUpdate` and `clearUpdate`, each defaulting to "this type has no replacement cleanup", and the four methods on `CloudFormationResourceProvisioner` consult the registry before falling through to the Step Functions arm. Pipes implement them, so a rename gets the three attempts, the `DELETE_IN_PROGRESS` / `DELETE_COMPLETE` / `DELETE_FAILED` events and the stack status reason naming the orphan. The provisioner's own cleanup path is gone rather than duplicated.

**A failed rename rolls back too.** That reordering is what makes it possible: a rename rolled back because a **later** resource failed now finds the original pipe still on file, where before it had already been deleted. `rollbackUpdate` points the stack resource's physical id and `Arn` back at the prior pipe and deletes the replacement, so the pipe no longer lands in `UPDATE_FAILED` with "Rollback is not implemented". The delete is guarded: the rename record carries the replacement's creation time, and on any mismatch or absence nothing is deleted and nothing moves, the resource failing with a message naming both pipes instead of guessing which one this update created. The restore writes the physical id, the `Arn` and the cleared rename marker before it deletes the replacement, so a delete that fails leaves the resource naming the pipe the rollback kept rather than the one it was discarding.

**A failure in the pipe's own update sequence restores it before the exception leaves the provisioner.** The snapshot could not reach `rollbackUpdate` in that case: `provision` writes it onto a copy of the resource's attributes, the service puts the pre-update resource back in its place, and the restored marker it stamps short-circuits the rollback hook. So a failed tag reconciliation left the pipe carrying the failed update's configuration while the stack reported `UPDATE_ROLLBACK_COMPLETE`. `updateExistingPipe` now restores from its own snapshot inside `provision` and rethrows the original failure, which is the contract `CloudFormationService` states and two provisioners already honour. If that restore itself fails, the resource is marked so the stack lands `UPDATE_ROLLBACK_FAILED` naming the pipe rather than reporting a clean rollback.

**A cleanup phase that never finished belongs to the update that started it, so a new update is refused and `DeleteStack` is the way out.** Deferring the delete opened this question: `CloudFormationService` persists the stack with the rename marker pending and the displaced pipe alive, which is what makes the update durable, and nothing picks the phase back up afterwards. That matches AWS, which resumes nothing: `CancelUpdateStack` is documented only for `UPDATE_IN_PROGRESS`, `ContinueUpdateRollback` moves a stack only from `UPDATE_ROLLBACK_FAILED` to `UPDATE_ROLLBACK_COMPLETE`, and the troubleshooting guide's answer for a stack stuck in `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS` is to contact Support. A status ending in `_IN_PROGRESS` means an operation owns the stack. So `UpdateStack` and a non-CREATE `CreateChangeSet` are refused in that state, in CloudFormation's own wording, `Stack:<stack id> is in <status> state and can not be updated.`, and `DeleteStack` stays allowed in every status and now deletes what a resource's pending cleanup still names as well as what its physical id names. That last part goes through the same four hooks, so Step Functions gains the same escape. `REVIEW_IN_PROGRESS` is excluded, since a new stack sits in it between creation and execution and an unscoped guard would refuse every `CreateStack`.

That closes the resource's lifecycle on this path: create, update in place, roll back an update that failed elsewhere, roll back an update that failed here, replace by rename, roll back a rename, clean up the displaced pipe, and delete the one a cleanup phase abandoned.

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.pipes.**'          →  147 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**' → 1041 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'  →  708 tests, 0 failures
```

The third run is there because the cleanup seam is shared: Step Functions travels the same four methods and its behaviour is unchanged.

25 commits, 15 files, test-first. Rebased onto `main` with #2788's `stablePhysicalName`, `reusesPriorEntity` and `staleTagKeys` adopted rather than duplicated.

One variant is deliberately not here, because it belongs to CloudFormation rather than to this resource type: `executeChangeSet` does not check the stack status, so a second `UpdateStack` can enter `provision` while the first is still inside its cleanup phase. Real CloudFormation rejects an update while a stack is `*_IN_PROGRESS`.


facuparedes:
Thanks — the two findings landed on opposite sides once I checked them against what CloudFormation documents.

**The failed cleanup of the renamed pipe is what AWS does, so this one stays.** From [Monitor the progress of a stack update](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-monitor-stack.html):

> CloudFormation makes three attempts at deleting the old resource. If CloudFormation can't delete the old resource, it removes the old resource from the stack and continues updating the stack. When the stack update is complete, CloudFormation issues an `UPDATE_COMPLETE` stack event, but includes a `StatusReason` that states that one or more resources couldn't be deleted. [...] The old resource still exists and will continue to incur charges, but is no longer accessible through CloudFormation.

An orphaned old pipe on a stack that still reaches `UPDATE_COMPLETE` is the documented outcome, not a bug. Failing the update here would be less faithful, not more. The failure is not swallowed either: it is logged at `warn` naming both the orphan and its replacement.

**The partially applied update was real, and it is fixed in `9d2c8f66`.** The same page draws the opposite conclusion for that phase:

> If a resource update fails, CloudFormation rolls back any resources that it has updated during the upgrade to their configurations before the update.

That path did not exist before this PR — `UpdateStack` used to fail with `ConflictException` before mutating anything — so the PR opened it and now closes it. `PipesCfnProvisioner` snapshots the pipe before `updatePipe` and restores it from `rollbackUpdate`, following the pattern Step Functions already uses. The hook is a `default` on `CfnResourceProvisioner`, so an extracted provisioner owns its own restore instead of adding another arm to `CloudFormationResourceProvisioner`.

Two limits are documented in the code rather than hidden: a successful rename writes no snapshot, since it goes through the create arm, and `updatePipe` reads a null argument as "leave this one alone", so the restore is exactly as faithful as the update it undoes.

`./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**'` → 1006 tests, 0 failures.

facuparedes:
One of the two was right, and it is fixed in `1a6a644c`, `f2d67464` and `db3af58d`.

**The null-semantics gap in the restore was real.** The rollback was putting the pipe back through `UpdatePipe`, which reads a null property as "leave this one alone". So a property the pipe did not carry before, that the failed update then set, survived the rollback. CloudFormation documents the opposite for that phase:

> If a resource update fails, CloudFormation rolls back any resources that it has updated during the upgrade to their configurations before the update.

`PipesService` now has two entry points over one writer, `writePipeConfiguration(..., boolean clearUnsetProperties)`:

- `updatePipe` keeps its exact contract, a null property is left alone. That is the AWS `UpdatePipe` API and nothing about it changes.
- `restorePipe` writes the configuration in full, a null property is cleared. That is what a rollback needs, and it is not an API operation, it is CloudFormation putting the resource back.

The snapshot already held the whole configuration, so the restore now lands on exactly the pre-update pipe. `desiredState` is the one property left alone on both paths, because `CreatePipe` always sets it and `currentState` is derived from it, so clearing it would leave `currentState` asserting a state `desiredState` no longer names. That is commented at the line.

Test-first: `PipesCfnRollbackRestoresExactConfigurationTest` failed with

```
AssertionFailedError: the pipe carried no description before the update
  ==> expected: <null> but was: <added by the failed update>
```

before the fix, and its symmetric case (a property that existed and was changed comes back to its former value) was green in the same run, which is what isolates the fault to the unset arm. Both are driven through `provision` + `rollbackUpdate` over a real `PipesService`, spied only to make the update's tag call fail after the pipe was written.

**The failed cleanup of the renamed pipe is what AWS does, so that one stays.** From [Monitor the progress of a stack update](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-monitor-stack.html):

> CloudFormation makes three attempts at deleting the old resource. If CloudFormation can't delete the old resource, it removes the old resource from the stack and continues updating the stack. When the stack update is complete, CloudFormation issues an `UPDATE_COMPLETE` stack event, but includes a `StatusReason` that states that one or more resources couldn't be deleted. [...] The old resource still exists and will continue to incur charges, but is no longer accessible through CloudFormation.

An orphaned prior pipe on a stack that still reaches `UPDATE_COMPLETE` is the documented outcome. Failing the update there would be less faithful, not more. The failure is not swallowed either: it is logged at `warn` naming both the orphan and its replacement.

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.pipes.**'          → 147 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**' → 1008 tests, 0 failures
```

facuparedes:
You were right that the prior pipe could end up unmanaged, and wrong about the fix being to fail the update. Both halves are now settled in `0c1aa511`, `a895628f` and `737a5189`, by deleting this PR's own cleanup path rather than defending it.

CloudFormation's documented behaviour has two parts, and the code only had the first:

> CloudFormation makes three attempts at deleting the old resource. If CloudFormation can't delete the old resource, it removes the old resource from the stack and continues updating the stack. When the stack update is complete, CloudFormation issues an `UPDATE_COMPLETE` stack event, but includes a `StatusReason` that states that one or more resources couldn't be deleted.

Continuing was right. Reporting only into a log line was not: a client calling `DescribeStackEvents` had no way to learn a pipe had been left behind.

The repo already implements that whole paragraph, for Step Functions only. `CloudFormationService.finishCommittedResourceCleanup` runs after the update commits, emits `DELETE_IN_PROGRESS`, retries `completeUpdate` up to three times, emits `DELETE_FAILED` with the reason, and `finishCommittedStackUpdate` closes the stack as `UPDATE_COMPLETE` with `The following resource(s) could not be deleted during update cleanup: [...]`. Its four entry points read one hardcoded Step Functions attribute, so an extracted provisioner could not reach them.

So pipes now go through it instead of carrying a second copy of the concept:

- `CfnResourceProvisioner` gains `default` hooks for `updateCleanupPhysicalId`, `completeUpdate`, `hasReplacementUpdate` and `clearUpdate`, beside the `rollbackUpdate` hook this PR already added. Every default means "this type has no replacement cleanup", so no other provisioner changes behaviour.
- The four methods on `CloudFormationResourceProvisioner` consult the registry first and fall through to the Step Functions arm, exactly as `rollbackUpdate` does. `UpdateCleanupResult` moves out of that class into the `provisioners` package so the extracted package does not import the monolith it is being extracted from.
- `PipesCfnProvisioner.provision` no longer deletes anything. A rename records the displaced pipe's name, and the committed cleanup deletes it afterwards, one attempt per call, honouring `UpdateReplacePolicy: Retain` and treating an already-absent pipe as complete. `createPipeAndDeleteRenamedPrior`, its `try`/`catch` and its `warn` are gone.

That also closes a hole neither of us had named: with the prior pipe alive until the update commits, a rename that is rolled back because a **later** resource failed now still has the original on file. It was being deleted before the stack update was committed.

Proof, both directions, integration level:

```
permanentRenameCleanupFailureKeepsCommittedUpdateAndNamesTheOrphanedPipe
  → UPDATE_COMPLETE + "The following resource(s) could not be deleted during
    update cleanup: [MyPipe (pipe-cleanup-old-...)]." + a DELETE_FAILED event
transientRenameCleanupFailureSucceedsOnThirdAttempt
  → DELETE_IN_PROGRESS, three attempts, DELETE_COMPLETE, no status reason
```

Both failed before the change, the first with no `StackStatusReason` element at all and the second with no delete events at all.

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.pipes.**'          →  147 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**' → 1012 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'  →  708 tests, 0 failures
```

The third run is there because the seam is shared: Step Functions cleanup travels the same four methods and its behaviour is unchanged.

facuparedes:
The first finding was real and is fixed in `3858c234` and `be349835`. The second is what AWS does, and the code now reports it the way AWS reports it.

**A failed rename now rolls back.** `PipesCfnProvisioner.rollbackUpdate` had no rename arm, so a stack update that renamed a pipe and then failed at a later resource landed the pipe in `UPDATE_FAILED` with "Rollback is not implemented". It could not have been fixed before the previous commits on this branch: `provision` deleted the displaced pipe inline, so by rollback time it was gone. Now it survives until the update commits, and a rollback means the update never commits, so the restore has something to restore to.

The rollback points the stack resource's physical id and `Arn` back at the prior pipe, deletes the replacement this update created, and spends the rename record so the committed cleanup cannot fire on a stale entry. The prior pipe is not touched: it was displaced, never modified.

Deleting the replacement is guarded rather than assumed. The rename record carries the created pipe's creation time; at rollback the pipe standing under the resource's physical id is described and its creation time compared. On any mismatch, absence, or a record without that field, nothing is deleted and nothing moves:

```
IllegalStateException: Refusing to delete pipe X while rolling back resource Y:
it was not created by this stack update, so both pipes are left as they are.
```

`CloudFormationService.rollbackUpdatedResources` catches it and lands the resource in `UPDATE_FAILED` with that message. A rollback that cannot prove what it is about to delete leaves both pipes alive and says so, rather than guessing.

Red before the fix, at the integration level:

```
aRenameRolledBackPointsTheStackResourceAtThePriorPipe:172
Expected: a string containing "<StackStatus>UPDATE_ROLLBACK_COMPLETE</StackStatus>"
  Actual: <StackStatus>UPDATE_ROLLBACK_FAILED</StackStatus>
          <StackStatusReason>The following resource(s) failed to roll back: [MyPipe].</StackStatusReason>
```

**On the displaced pipe after a failed cleanup: that is the documented outcome, not a defect.** The same page you are reading the update contract from:

> The old resource still exists and will continue to incur charges, but is no longer accessible through CloudFormation.

Losing CloudFormation's tracking of it is the behaviour, stated by AWS in that sentence. What a client is owed is being told, and that is what changed: after three attempts the cleanup emits a `DELETE_FAILED` event naming the pipe, and the stack closes `UPDATE_COMPLETE` with `The following resource(s) could not be deleted during update cleanup: [MyPipe (...)]`, which is the `StatusReason` AWS documents for exactly this case. Keeping the pipe in the stack, or failing the update over it, would both diverge from AWS.

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.pipes.**'          →  147 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**' → 1016 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'  →  708 tests, 0 failures
```

One thing found on the way and left alone, since it belongs to every resource type rather than to pipes: after a provisioner's `rollbackUpdate` returns true, `CloudFormationService` copies the previous status but never restores the previous physical id or attributes, so a replacement-shaped rollback has to repoint the stack resource itself, as this one does.

hectorvent:
Thank you, this is careful work: the reconcile keeps the physical id per the schema's createOnly set, the `restorePipe` split gives a rollback that clears what the failed update added, and routing rename cleanup through the existing three-attempt path is the right call. The rename rollback in be34983 closes the case I was about to raise, and the creation-time guard on the replacement delete is a nice touch.

**(blocking)** #2788 landed on main today and rewrites the same `provision` for the update path, so this branch now conflicts there. It adds `ProvisionContext.reusesPriorEntity` and `staleTagKeys` plus the same Source refusal and tag reconcile; on the rebase I would keep this PR's lookup, snapshot, rollback and rename cleanup on top of those helpers rather than a second copy of the name check and tag diff.

The rebase is what unblocks it from my side.

facuparedes:
Rebased onto `main`, 16 commits, with #2788's helpers adopted rather than duplicated.

Only `PipesCfnProvisioner` conflicted, three times, once per commit here that rewrote `provision`. What is now main's and no longer has a second copy in this branch:

- `ctx.stablePhysicalName(...)` replaces this branch's `priorPhysicalId == null ? generate : keep`.
- `ctx.reusesPriorEntity(name)` replaces its own name-versus-prior-id check.
- Main's `Source` refusal wording and its `source != null && !source.equals(...)` shape replace the `Objects.equals` copy.
- `reconcileTags`, on `ProvisionContext.staleTagKeys`, replaces the inline `listTags`/filter/untag/tag block. There is exactly one tag diff in the file now, and both restore paths call it.

Three things from this branch stayed inside that structure, because they are behaviours main's shape does not have rather than a second implementation of them:

- the tolerant lookup, `pipeOnFile`, inside the reuse guard instead of a bare `describePipe`, so a pipe deleted out of band since the prior deploy is created again instead of failing the `UpdateStack`;
- `ValidationException` / `Source is required` for an absent or blank `Source`, which main's shape reads as unchanged;
- the rename arm, and with it the snapshot, `rollbackUpdate`, the committed-cleanup hooks and the eager restore.

One non-conflict edit: `CfnStableNameUpdatePathTest` lines 131, 178 and 346 go from `new PipesCfnProvisioner(pipes)` to `new PipesCfnProvisioner(pipes, mapper)`, reusing the file's existing `mapper` field. The snapshot needs an `ObjectMapper`, so main's test would not compile otherwise. No assertion touched, and it is folded into the commit that changes the constructor so every commit in the series still builds.

Since your comment, one more finding landed, from Greptile and confirmed by tracing it: when the pipe's **own** update sequence threw, the snapshot never reached the rollback. `CloudFormationResourceProvisioner:401` writes it onto a copy of the attributes, `CloudFormationService:1130` puts `previousResource` back in its place, and `:1516-1518` short-circuits `rollbackUpdate` because `:1118` already stamped `UPDATE_ROLLBACK_RESTORED_ATTR`. So a failed tag reconciliation left the pipe carrying the failed update's configuration while the stack reported `UPDATE_ROLLBACK_COMPLETE`. `updateExistingPipe` now restores from its own snapshot inside `provision` before rethrowing, which is the contract `CloudFormationService:1114-1120` states and two provisioners already honour. When that restore itself fails it writes `UPDATE_ROLLBACK_FAILURE_ATTR`, so the stack lands `UPDATE_ROLLBACK_FAILED` naming the pipe instead of reporting a clean rollback.

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.pipes.**'          →  147 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**' → 1038 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'  →  708 tests, 0 failures
```

The cloudformation run includes `CfnStableNameUpdatePathTest` 9/9 and `ProvisionContextTest` 9/9 from #2788. Nothing from main was adapted.

Two things spotted in #2788's own code while resolving, both outside this PR's resource type and untouched, in case they are worth their own issue:

- `FirehoseCfnProvisioner:77` calls `describeDeliveryStream(name)` on the reuse path with no not-found tolerance, so a stream deleted out of band since the prior deploy fails the `UpdateStack` rather than being created again.
- `EcrCfnProvisioner:58` has the same shape in `reconcileExisting`.

That is the defect `pipeOnFile` covers here.

facuparedes:
Real, and fixed in `542e86a3`, `54f76040` and `2691e22b`.

The rename cleanup is deferred to the committed phase on purpose, so the displaced pipe survives a rollback that a later resource triggers. What was missing is what happens when that phase never finishes.

`CloudFormationService:1200` persists the stack with the rename marker pending and the displaced pipe still alive, which is the point at which the update becomes durable. Nothing resumes it afterwards: `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS` is written at 1193, 1197, 1223 and 1228 and read nowhere, and `loadPersistedState` only loads stacks. On the next `UpdateStack`, `CloudFormationResourceProvisioner:400` carries the marker into the fresh `StackResource` and `PipesCfnProvisioner:50` removes it as the first statement of `provision`. From there `hasReplacementUpdate` is false, `updateCleanupPhysicalId` returns null, and nothing names the pipe again. `delete` only deletes the current physical id, so `DeleteStack` leaves it running too.

The fix resumes the phase rather than holding the marker longer. In `executeTemplate`, before the template is parsed:

```java
List<UpdateCleanupFailure> cleanupFailures = new ArrayList<>();
if (!isCreate && hasReplacementUpdates(stack)) {
    cleanupFailures.addAll(finishCommittedResourceCleanup(stack));
}
```

and the committed-cleanup block at the end adds into that list instead of building a second one. By the time `provision` reaches its `remove`, the marker is either spent by a successful delete or given up on after the third attempt with its `DELETE_FAILED` event, and the failure reaches this update's `UPDATE_COMPLETE` status reason.

The trigger is the marker, not the stack status: `executeChangeSet:590` has already set `UPDATE_IN_PROGRESS` before `executeTemplate` runs. A completed cleanup clears the marker in `clearUpdate` and a rollback spends it in `restoreRenamedPipe`, so a marker still present at entry means the phase never finished, and nothing else. Step Functions travels the same seam and gains the same resume.

Two cases, both red first:

```
anInterruptedRenameCleanupDeletesTheDisplacedPipeOnTheNextUpdate:189
  assertPipeMissing:469
  Expected status code <404> but was <200>

anInterruptedCleanupThatFailsOnResumeNamesTheOrphanInTheStatusReason:234
  Expected: a string containing "The following resource(s) could not be deleted
  during update cleanup: [MyPipe (pipe-cleanup-resume-failure-old-35u1dgxnlp)]."
```

The interruption comes from the code's own path rather than from editing the stack. The spy throws an `Error`; `completeUpdate` catches only `RuntimeException` and `executeTemplate` only `Exception`, so the throwable leaves the method and the stack is left exactly as `persistStack` wrote it, marker pending with zero attempts spent and the displaced pipe alive. The second case then makes the resumed delete fail persistently, so it burns all three attempts and lands in the status reason.

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.pipes.**'          →  147 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**' → 1040 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'  →  708 tests, 0 failures
```

One variant left out on purpose, because it belongs to CloudFormation rather than to this resource type: `executeChangeSet` does not check the stack status, so a second `UpdateStack` can enter `provision` while the first is still inside the cleanup phase. Real CloudFormation rejects an update while a stack is `*_IN_PROGRESS`. Happy to open a separate issue for that guard.

facuparedes:
Real, and worse than it reads: the previous commit on this branch is what made it reachable. Fixed in `3ccee456` and `cda26125`.

`restoreRenamedPipe` did the destructive call first and its bookkeeping after:

```java
deleteReplacementThisUpdateCreated(resource, cleanup, region);          // can throw
resource.setPhysicalId(priorPipeName);                                 // not reached
resource.getAttributes().put("Arn", priorPipe.getArn());               // not reached
resource.getAttributes().remove(CfnRollback.PIPE_RENAME_CLEANUP_ATTR); // not reached
```

So a rollback whose replacement delete fails leaves the physical id naming the replacement with the rename marker still set, and `CloudFormationService` persists that pair with `UPDATE_ROLLBACK_FAILED`. The next `UpdateStack` reads the marker, the guard `priorPipeName.equals(resource.getPhysicalId())` misses because the id is the replacement, and the prior pipe is deleted. That is the one pipe the rollback existed to preserve.

Measured on both sides of `54f76040`, same worktree, only `CloudFormationService.java` swapped:

```
at HEAD                          prior 404   replacement 200
with CloudFormationService.java  prior 200   replacement 200
from the commit before the resume
```

Before the resume the stale marker was inert: `provision` dropped it and the prior pipe merely leaked. Resuming the phase turned that leak into a deletion, and the stack closed `UPDATE_COMPLETE` with no `StackStatusReason`, so nothing recorded it.

The fix makes the rollback's bookkeeping durable before the destructive call, and passes the replacement name in rather than re-reading it off the resource:

```java
resource.setPhysicalId(priorPipeName);
resource.getAttributes().put("Arn", priorPipe.getArn());
resource.getAttributes().remove(CfnRollback.PIPE_RENAME_CLEANUP_ATTR);
deleteReplacementThisUpdateCreated(resource, replacementPipeName, cleanup, region);
```

The throw still propagates, so the stack still lands `UPDATE_ROLLBACK_FAILED` naming the resource and the failure stays visible. What changes is which pipe survives. The resource names the prior pipe, and what is left orphaned is the replacement the failed update created, which is the direction CloudFormation errs in: a rollback restores the original and leaks what it could not clean up, it never deletes the thing it was restoring.

Red first, at the integration level:

```
aRollbackThatCannotDeleteTheReplacementStillKeepsThePriorPipe
Expected: a string containing "<PhysicalResourceId>pipe-rename-rollback-keep-prior-...</PhysicalResourceId>"
  Actual: <PhysicalResourceId>pipe-rename-rollback-keep-replacement-...</PhysicalResourceId>
          <ResourceStatus>UPDATE_FAILED</ResourceStatus>
```

and, with that assertion relaxed so the run reaches the second `UpdateStack`, the destruction itself: `Expected status code <200> but was <404>` on the prior pipe.

Restoring the original statement order turns exactly two cases red and no others, which is what pins the ordering rather than the wording.

One committed assertion changed with it. `PipesCfnProvisionerTest#aRenameRollbackRefusesToDeleteAPipeThisUpdateDidNotCreate` asserted the resource still pointed at the replacement after the creation-time guard refused. That pinned the destructive state: the same id-plus-marker pair the next update reads. Its own docstring says the refusal leaves both pipes where they are, which that assertion contradicted one update later. It now asserts the prior name and that no replacement update is left pending.

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.pipes.**'          →  147 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**' → 1041 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'  →  708 tests, 0 failures
```

One cost this leaves standing, and I would rather name it than let it be found: after such a rollback the replacement pipe is still on file under the new name, so retrying the rename hits `ConflictException` on `createPipe` until an operator deletes it by hand, and nothing in the stack names that orphan. `UPDATE_ROLLBACK_FAILED` names the resource, not the pipe left behind. The committed-cleanup path already has the vocabulary for this in `UpdateCleanupFailure` and its status reason. Happy to extend it to the rollback path here, or in its own change if you would rather keep this one to the defect.

facuparedes:
Real, and it is the third finding against the same four lines, so I removed the mechanism instead of patching it. `1ef70a67`, `a9b11931`, `68e650f8`, `dee43ea7`.

First, confirmed. A resumed cleanup that exhausts its three attempts under a second update that then fails:

```
afterSecondUpdate status=UPDATE_ROLLBACK_COMPLETE
afterSecondUpdate reason=<absent>
orphanAliveBeforeDeleteStack=200
orphanAliveAfterDeleteStack=200
replacementAfterDeleteStack=404
```

A clean rollback, no `StackStatusReason` at all, and a pipe the emulator gave up on still alive. `deleteStackResources` walks `stack.getResources()` by physical id and never consults `updateCleanupPhysicalId`, so `DeleteStack` did not reach it either.

**Why the resume is gone rather than fixed.** Threading the failure list to the rollback exits is about fifteen lines and I had it working. It buys correct reporting on the two exits that exist and nothing against the next exit anyone adds, because the shape is wrong: `54f76040` made `UpdateStack` the vehicle for finishing a phase that belongs to a previous update, so that phase runs inside the new update's transaction, before the new update has decided whether it commits. Every non-happy exit then has to be taught about a failure it did not cause. The three findings on this branch are three exits that were not.

AWS does not resume. A status ending in `_IN_PROGRESS` means an operation owns the stack:

- `CancelUpdateStack`: "You can cancel only stacks that are in the UPDATE_IN_PROGRESS state." Status is a precondition for an operation, not something an operation repairs.
- `ContinueUpdateRollback`: "Continues rolling back a stack from UPDATE_ROLLBACK_FAILED to UPDATE_ROLLBACK_COMPLETE state." One status transition, scoped to resources in `UPDATE_FAILED`. Not a way to finish a cleanup phase, and this project does not implement it.
- Troubleshooting, on nested stacks stuck in `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS`: "To fix the stack, contact AWS Support." AWS does expose stacks whose cleanup never completes, and offers the client no operation to finish one.

So `UpdateStack` and a non-CREATE `CreateChangeSet` are now refused while the stack's status ends in `_IN_PROGRESS`, and nothing finishes an abandoned phase, deliberately. The way out is `DeleteStack`, which stays allowed in every status and now deletes what the resource's cleanup still names as well as what its physical id names. That goes through `updateCleanupPhysicalId`, `completeUpdate` and `clearUpdate`, which default to "no replacement cleanup", so it is generic: Step Functions gains the same escape with no change of its own.

`REVIEW_IN_PROGRESS` also ends in `_IN_PROGRESS` and a new stack sits in it between creation and execution, so an unscoped guard would refuse every `CreateStack` and break the CREATE change set path `aws cloudformation deploy` depends on. A CREATE change set on an existing stack is already refused two lines above.

**On the refusal message.** It emits what CloudFormation emits:

```
Stack:<stack id arn> is in <STATUS> state and can not be updated.
```

with AWS's two-word `can not`, the stack id rather than the name, and no space after the colon. That deviates from this file's `Stack [name]` bracket idiom on purpose: a client matching on this message is exactly who the refusal is for. Honest caveat on the sourcing: the API reference documents the `ValidationError` code but not the message text, so the wording is corroborated from four unrelated user reports quoting real AWS responses across three statuses, three regions and three accounts, 2018 to 2023, not from an AWS document. If you would rather this file keep its own idiom, say so and it is one string.

Two cases replace the two that encoded the resume, and both go on to assert the escape:

```
anUpdateIsRefusedWhileAnInterruptedCleanupOwnsTheStack:186     Expected status code <400> but was <200>
aChangeSetIsRefusedWhileAnInterruptedCleanupOwnsTheStack:243   Expected status code <400> but was <200>
```

The guard broke nothing elsewhere: every Query handler awaits its own execution before it answers, so the only externally reachable `_IN_PROGRESS` status is the one an interrupted cleanup leaves.

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.pipes.**'          →  147 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.cloudformation.**' → 1041 tests, 0 failures
./mvnw test -Dtest='io.github.hectorvent.floci.services.stepfunctions.**'  →  708 tests, 0 failures
```

Two things I will not claim are pinned. The removal of the resume block is not held out by any test and cannot be: with the guard in front of `executeTemplate` it is unreachable either way, so it is gone because it should not exist, not because a test says so. And `endsWith("_IN_PROGRESS")` would survive being narrowed to the one status these tests exercise, because every other such status is unobservable through the Query seam.

One thing left out, visible in the same sources: AWS emits that same sentence for `ROLLBACK_COMPLETE`, a terminal status this guard does not cover, so an update over a stack in `ROLLBACK_COMPLETE` is still allowed here. That is a different rule about a different status and I would rather it be its own change than widen this one further.

This is more than the rebase you unblocked, and it removes something you saw in an earlier round rather than adding to it. If you would rather have the fifteen-line fix that keeps the resume, or want the refusal split into its own pull request, I will do either.

facuparedes:
Caught up with `main` via merge commit `e858e8af`.

One conflict, in `CfnRollback.java`: #3008 landed `UPDATE_ROLLBACK_FAILURE_ATTR` under the same name and the same value this branch introduces. The resolution keeps `main`'s declaration and its javadoc verbatim and puts this branch's two pipe-scoped markers beside it, so that file contributes 18 added lines and no deletions.

```
pipes.**          →  147 tests, 0 failures
cloudformation.** → 1121 tests, 0 failures
stepfunctions.**  →  813 tests, 0 failures
```

The last two counts are higher than the run reported above because of the 50 upstream commits, not because of anything on this branch: `origin/main...HEAD` touches no Step Functions file at all and that suite still gained 105 passing tests. That is the check on the shared cleanup seam, since the four `updateCleanupPhysicalId` / `completeUpdate` / `hasReplacementUpdate` / `clearUpdate` hooks are travelled by both resource types.

hectorvent (approved):
Thank you, the rebase came out exactly as hoped: `stablePhysicalName`, `reusesPriorEntity` and `staleTagKeys` are main's, the Source refusal is main's wording, and the lookup, snapshot, rollback and rename cleanup sit on top of them. I also followed the three later rounds; the eager restore and the bookkeeping-before-delete fix are both right.

On the two questions. Keep the refusal in this PR: it replaces the resume this branch had introduced, and the message is the one LocalStack and moto emit for the same case, so a client matching on it is served. Keep AWS's wording, "can not" and `Stack:<id>` included, rather than the file's bracket idiom. Agreed that `ROLLBACK_COMPLETE` is a separate change; I will open an issue for it so it is not lost.

Merging, thank you for the persistence on this one.

PR #2971, opened 2026-09-02
feat(cloudformation): report a stubbed unsupported resource type instead of hiding it

### LINKED ISSUE #2967: [BUG] An unrecognised resource type is stubbed silently and the stack reports success
## Service

CloudFormation

## AWS API Action

`CreateStack`, `ValidateTemplate`

## Expected behavior

For a resource type that does not exist in the CloudFormation registry, AWS rejects the template at validation and never creates a stack:

```
$ aws cloudformation validate-template \
    --template-body '{"Resources":{"P":{"Type":"AWS::NotAReal::Thing"}}}'

ValidationError: Template format error: Unrecognized resource types: [AWS::NotAReal::Thing]
```

`create-change-set --change-set-type CREATE` returns the same error synchronously, and a template mixing one real type with one unrecognised type is rejected as a whole. `describe-type` on that type answers `TypeNotFoundException`.

For a type that **does** exist in the registry but that Floci has no provisioner for, AWS cannot arbitrate: that situation does not exist in AWS. The two cases need different answers, and today they take the same path.

## Actual behavior

`CloudFormationResourceProvisioner` assigns `arn:aws:stub:::<logicalId>` as the physical id, sets the resource to `CREATE_COMPLETE`, and logs at `debug`. The stack reports success, so a green stack does not prove its resources exist.

Floci implements 113 of roughly 1000 registry types, so this is reachable with ordinary templates rather than only with a typo.

## Reproduction

```bash
  --stack-name s --template-body '{"Resources":{"P":{"Type":"AWS::NotAReal::Thing"}}}'

aws --endpoint-url http://localhost:4566 cloudformation describe-stacks --stack-name s
# CREATE_COMPLETE

aws --endpoint-url http://localhost:4566 cloudformation describe-stack-resources --stack-name s
# PhysicalResourceId: arn:aws:stub:::P
```

## Environment

facuparedes:
Closes #2967

An unsupported resource type was stubbed at `debug`, so the stack reached `CREATE_COMPLETE` and a green stack did not prove its resources existed.

Three changes:

- The stub is reported at `warn`, naming the type and the reason it was stubbed.
- A new `allow-stub-unsupported-resource-types` setting turns the stub into a stack failure, for suites that need a green stack to mean every resource exists.
- An unexpanded SAM resource type warns on its own, and the delete path names the resource it leaves in place when no delete is implemented.

The default behaviour is unchanged, so existing templates keep deploying. AWS rejects a registry-absent type at `validate-template` and never creates the stack; a registry-present type with no provisioner has no AWS counterpart, which is why this is a setting rather than a hardcoded failure. Both measurements are in the issue.

9 commits, 10 files, test-first.

### DIFF
--- a/docs/configuration/environment-variables.md
+++ b/docs/configuration/environment-variables.md
@@ -297,6 +297,8 @@ See [Initialization Hooks](./initialization-hooks.md) for lifecycle phases and s
 | Variable | Default | Description |
 |---|---|---|
 | `FLOCI_SERVICES_CLOUDFORMATION_ENABLED` | `true` | Enable the CloudFormation service |
+| `FLOCI_SERVICES_CLOUDFORMATION_ALLOW_STUB_LAMBDA_CODE` | `false` | Fall back to the built-in stub handler for an `AWS::Lambda::Function` whose S3 code cannot be read. Off matches real CloudFormation, which fails the resource and rolls the stack back |
+| `FLOCI_SERVICES_CLOUDFORMATION_ALLOW_STUB_UNSUPPORTED_RESOURCE_TYPES` | `true` | Stub a resource whose type has no provisioner (synthetic physical ID, `arn:aws:stub:::` ARN attribute, `CREATE_COMPLETE`), logged at `WARN` with a resource status reason. Set `false` to fail the resource instead, which rolls the stack back |
 
 ### ACM (Certificate Manager)
 
--- a/docs/services/cloudformation.md
+++ b/docs/services/cloudformation.md
@@ -96,7 +96,18 @@ cross-resource references.
 
 All other resource types are accepted without error and assigned a synthetic physical ID (with an
 `arn:aws:stub:::<logicalId>` ARN attribute), so templates with unsupported types still reach
-`CREATE_COMPLETE` rather than failing.
+`CREATE_COMPLETE` rather than failing. Nothing is created for such a resource, so each one is
+logged at `WARN` and carries a resource status reason saying so, which `DescribeStackEvents`
+returns:
+
+```
+Resource type AWS::Fake::Thing is not supported by Floci. It was stubbed and nothing was created for it.
+```
+
+Set `floci.services.cloudformation.allow-stub-unsupported-resource-types` to `false`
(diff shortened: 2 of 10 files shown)


facuparedes:
The red `Build and Test (shard 2)` is unrelated to this PR.

```
[ERROR] io.github.hectorvent.floci.services.swf.SwfServiceTest.closingAChildWhileTheParentIsMutated_keepsBothHistoriesIntact
java.util.concurrent.ExecutionException: java.lang.AssertionError: no decision task for wf-cc-child-3
	at SwfServiceTest.pollFor(SwfServiceTest.java:1222)
```

`pollFor` is a bounded wait for a decision task the concurrent arm of that test races against, so it fails when the shard is loaded rather than when the code is wrong. The shard ran 3633 tests with that single error and no failures.

This PR touches CloudFormation and SAM only, no file under `services/swf` and nothing the SWF path reads. Locally on this branch the class is green 5 times out of 5:

```
./mvnw test -Dtest='io.github.hectorvent.floci.services.swf.SwfServiceTest'
run1..run5: Tests run: 46, Failures: 0, Errors: 0, Skipped: 0
```

A re-run of the shard should clear it. I have no rerun rights on this repo, so I cannot trigger it myself.

hectorvent:
Thank you, this is a careful change: default behaviour untouched, the new key wired through `EmulatorConfig`, both `application.yml` files and the docs the way `allow-stub-lambda-code` is, and Greptile's point about the delete warning claiming a survivor was already fixed in ba5c0c2 before I got here. The shard 2 red is the SWF timing test you diagnosed; I have rerun it.

**(follow-up, separate PR)** #2967 draws the line AWS draws: a registry-absent type fails `ValidateTemplate`/`CreateStack` synchronously with `Unrecognized resource types`, no stack created, while a registry-present type with no provisioner has no AWS answer. Both still take the same path here. Shipping the registry's type-name list would let Floci reject the first case like AWS regardless of the setting, and leave the setting for the second. Out of scope for this PR, I will track it.

No blockers from my side.

hectorvent (approved):
Approving; notes above.

PR #2972, opened 2026-09-02
fix(stepfunctions): reject the QueryLanguage definitions AWS rejects

### LINKED ISSUE #2968: [BUG] Definitions AWS rejects are accepted by CreateStateMachine
## Service

Step Functions

## AWS API Action

`CreateStateMachine`, `ValidateStateMachineDefinition`

## Expected behavior

Measured with `aws stepfunctions validate-state-machine-definition`, which needs no role. Four rules, each with the diagnostic AWS emits verbatim, all `code: SCHEMA_VALIDATION_FAILED`:

**1. A JSONata-only field on a state whose effective language is JSONPath is rejected.**

```
The QueryLanguage is set to 'JSONPath', but field 'Output' is only supported for the 'JSONata' QueryLanguage
```

A `Map` declaring `"QueryLanguage": "JSONata"` does not change this for the states inside its `ItemProcessor`: those inherit the machine's top-level language, and the diagnostic points at `/States/M/ItemProcessor/States/P`. Declaring it on the `Map` and omitting it entirely produce identical results. The same holds for `Parallel` branches.

**2. A state cannot declare JSONPath under a JSONata machine.**

```
'QueryLanguage' can not be 'JSONPath' if set to 'JSONata' for whole state machine
```

at the state's own path.

**3. `ItemProcessor`, and its deprecated alias `Iterator`, cannot declare `QueryLanguage` at all.**

```
Field 'QueryLanguage' is not supported
```

at `/States/M/ItemProcessor` and `/States/M/Iterator` respectively.

**4. A `QueryLanguage` value resolves to JSONPath only when it is exactly the string `JSONPath`.**

Every other present value resolves to JSONata: an unrecognised one, a lowercase `jsonpath`, a non-string, and `null`. The enum, type and null diagnostics are emitted independently of that resolution, at the field's own path. `null` additionally produces `Value cannot be null` at `/QueryLanguage`.

## Actual behavior

Floci accepts all of the above. A definition that is invalid in AWS passes locally, which is the failure mode that matters most for an emulator used to gain confidence before deploying.

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 stepfunctions validate-state-machine-definition \
  --definition '{"QueryLanguage":"JSONPath","StartAt":"M","States":{"M":{"Type":"Map","QueryLanguage":"JSONata","ItemProcessor":{"ProcessorConfig":{"Mode":"INLINE"},"StartAt":"P","States":{"P":{"Type":"Pass","Output":"{% $states.input %}","End":true}}},"End":true}}}'

# Floci: {"result": "OK", "diagnostics": []}
# AWS:   FAIL, SCHEMA_VALIDATION_FAILED at /States/M/ItemProcessor/States/P
```

## Environment

facuparedes:
Closes #2968

Definitions AWS rejects were accepted, which is the failure mode that matters most for an emulator used to gain confidence before deploying.

`ValidateStateMachineDefinition` and `CreateStateMachine` now apply the four rules the issue measures against AWS, each emitting AWS's own diagnostic text and JSON pointer:

- A JSONata-only field on a state whose effective language is JSONPath is rejected, located at the state, not at the machine.
- A state cannot declare `JSONPath` under a JSONata machine.
- `ItemProcessor`, and its deprecated alias `Iterator`, cannot declare `QueryLanguage` at all.
- A `QueryLanguage` resolves to JSONPath only on the exact string `JSONPath`; every other present value resolves to JSONata, while the enum and type diagnostics are emitted independently at the field's own path.

Diagnostics are emitted in a fixed order so their sequence is assertable.

One thing the measurement refuted rather than confirmed: propagating the machine's top-level query language into `ItemProcessor` and `Parallel` branches is what AWS does, and that behaviour is unchanged here. A comment records why.

All rules measured with `aws stepfunctions validate-state-machine-definition`, which needs no IAM role.

14 commits, 5 files, test-first.

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -357,8 +357,9 @@ only on `Task`, `Parallel` and `Map`.
 
 ## A state's QueryLanguage, and the fields it may carry
 
-A state's query language is its own `QueryLanguage` field, or the state machine's when it declares
-none. That is the whole resolution: a state inside a `Map`'s `ItemProcessor` or `Iterator`, or
+A state's query language is JSONPath when its `QueryLanguage` field is exactly the string
+`"JSONPath"`, the state machine's when the field is absent, and JSONata for any other value: the
+wrong case, an unknown string and a non-string alike. A state inside a `Map`'s `ItemProcessor` or `Iterator`, or
 inside one of a `Parallel`'s `Branches`, falls back to the **state machine's** language and not to
 the enclosing `Map`'s or `Parallel`'s, which the Amazon States Language calls independent of it. The
 enclosing state's own fields, `Items` and `MaxConcurrency` among them, do use its own language.
@@ -380,8 +381,9 @@ Three more refusals, each measured against `ValidateStateMachineDefinition` on r
 
 A `QueryLanguage` value outside the enum is reported at the field, `/States/X/QueryLanguage` or
 `/QueryLanguage`, with `Value should be one of the following: [JSONPath, JSONata]`, and a
-non-string value with `Expected value of type [STRING]`. The effective language is still resolved
-case-insensitively, so `"jsonata"` is a JSONata state that also carries that diagnostic.
+non-string value with `Expected value of type [STRING]`. That diagnostic is independent of the
+resolution above, so `"jsonpath"` is a JSONata state that also carries it, and it is not a
+downgrade: only the exact string `"JSONPath"` under a JSONata machine is.
 
 Locations follow AWS: the two messages above point at the state, `/States/X`, while every other
 schema error points at the offending field, `/States/X/MaxConcurrency`.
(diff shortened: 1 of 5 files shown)


facuparedes:
Clean against `main`, all checks pass, no review findings outstanding.

This shares no file with #2970, so it does not need to wait on it. It overlaps #2969 in `StepFunctionsService.java` and `docs/services/step-functions.md`, but the two do not conflict, so whichever lands second is still a clean merge.

hectorvent (approved):
Thank you, and thank you for measuring each rule against AWS and carrying its exact text and JSON pointer through. Folding the four location-explicit diagnostics into one marker instead of a special case per rule, and making the JSONPath-only field list deterministic so paging with `maxResults=1` is stable, are both the right calls.

No blockers from my side.

PR #2976, opened 2026-09-02
feat(dynamodb): resource policies and GSI-level OnDemandThroughput

lex00:
DynamoDB rejected PutResourcePolicy/GetResourcePolicy/DeleteResourcePolicy
with UnknownOperationException, and a GSI created with OnDemandThroughput
never stored the value, so DescribeTable could not report it back. Both
break the Terraform provider: aws_dynamodb_resource_policy fails at apply,
and an on-demand GSI shows phantom drift on every plan.

Policies are stored on the table with an opaque RevisionId and honor
ExpectedRevisionId on Put/Delete, answering PolicyNotFoundException on a
mismatch or a missing policy, per the API reference. The GSI parse and
emit mirror the existing table-level OnDemandThroughput handling on
CreateTable, UpdateTable and DescribeTable.

Four new handler tests cover both round-trips plus the stale
ExpectedRevisionId and invalid-ARN rejections; the full local DynamoDb*
test family passes. Happy to adjust to conventions.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbJsonHandler.java
@@ -66,6 +66,9 @@ public class DynamoDbJsonHandler {
             case "TagResource" -> handleTagResource(request, region);
             case "UntagResource" -> handleUntagResource(request, region);
             case "ListTagsOfResource" -> handleListTagsOfResource(request, region);
+            case "PutResourcePolicy" -> handlePutResourcePolicy(request, region);
+            case "GetResourcePolicy" -> handleGetResourcePolicy(request, region);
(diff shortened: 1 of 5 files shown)


pgermosen (approved):
Approving. Checked the resource-policy error semantics against AWS's own PutResourcePolicy docs - "if the revision ID doesn't match or if there's no policy attached, PolicyNotFoundException" - which is exactly what the single expectedRevisionId check here covers, correctly folding both cases into one condition. ARN validation reuses the same isValidDynamoDbTableArn TagResource already uses, so the table-only scope is consistent with the rest of the handler, not a new gap.

On greptile's blocking concern: I checked GlobalSecondaryIndexUpdates directly and it only ever handled Create and Delete, no Update branch at all for any GSI field, provisioned throughput included. That's pre-existing and out of scope for what this PR claims (a GSI created with OnDemandThroughput never storing the value), not something this PR regresses. Filing that as its own issue rather than holding this one on it.

PR #2979, opened 2026-09-02
ci: pin every GitHub Action to a commit SHA

hectorvent:
## Summary

Pins the 38 floating `uses:` tag refs across `ci`, `compatibility`, `doc-style`, `docs`, `docs-actions`, `conventional-commits`, `nightly`, `release` and `welcome` to full commit SHAs, using the `sha # vX.Y.Z` style that `release-cut` and `build-images` already follow.

Floating majors were pinned to their current release: `upload-artifact` v7.0.1, `download-artifact` v8.0.1, `setup-python` v7.0.0, `amazon-ecr-login` v2.0.1. Every SHA was checked against the commit its tag points at on GitHub.

No workflow logic changes: the diff is 38 one-line ref substitutions.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Not applicable, CI configuration only.

## Checklist

- [ ] `./mvnw test` passes locally (not run: no Java changes)
- [ ] New or updated integration test added (not applicable: workflow refs only)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,9 +39,9 @@ jobs:
       matrix:
         shard: [1, 2, 3, 4]
     steps:
-      - uses: actions/checkout@v7.0.1
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-      - uses: actions/setup-java@v6.0.0
+      - uses: actions/setup-java@dd06d9cba3e5552c54d9f8ea23572deb30010f7c # v6.0.0
         with:
           java-version: '25'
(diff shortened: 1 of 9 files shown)


pgermosen (approved):
Good hardening. Checked every distinct SHA against the actual tag on GitHub rather than just trusting the claim, all 9 action+version pairs behind the 38 substitutions match, including catching that amazon-ecr-login's tag is annotated so it needed an extra dereference to get to the real commit. Confirmed no workflow logic changed anywhere, every hunk is a pure ref substitution.

PR #2980, opened 2026-09-02
docs: fix stale image tags, release model, service nav and contributor checklists

hectorvent:
## Summary

The getting-started and contributor docs had drifted from what the workflows and the code actually do. This PR brings them back in line and adds gates so the same drift fails CI next time.

- quick-start and CONTRIBUTING pointed users at a `latest-jvm` image that the release pipeline never publishes. Both now name the `-compat` image, and `make docs-check` greps the docs for any `-jvm` tag.
- The CONTRIBUTING branching model now matches the workflows: `main` is published as the nightly channel, a maintainer runs the Release Cut workflow to tag a release, and the tag triggers the publish. PR builds produce compat images but never publish them.
- README and the docs landing page carried a hand-maintained service count and a second copy of the service table, both stale. They now link to the Services Overview.
- The Service Matrix gains the ELB Classic row it was missing, with an `elb` alias so the check no longer resolves the Classic key to the v2 page. `check_service_matrix.py` now also fails on a `docs/services` page that has no row, with a `facet_pages:` escape hatch.
- The README category table lists 16 documented services it omitted.
- mkdocs nav gains 11 service pages that were unreachable from the site, drops a duplicate CloudTrail entry, and the docs job now runs `mkdocs build --strict` with nav and link validation so a page without a nav entry fails CI.
- AGENTS.md, CONTRIBUTING.md and docs/contributing.md told contributors to register a service in `ServiceRegistry`, which has no registration API. All three now point at `ResolvedServiceCatalog` and list the config, storage, native-image, docs, tooling and compat-suite touchpoints a new service needs.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Not applicable: documentation and docs tooling only, no emulator behavior changes.

## Checklist

- [ ] `./mvnw test` passes locally (not run: no Java changes)
- [x] New or updated integration test added (`tools/docs/test_check_service_matrix.py` covers the new unlinked-page check; `make docs-test`, `make docs-check` and `mkdocs build --strict` pass locally)

### DIFF
--- a/.github/workflows/docs-actions.yml
+++ b/.github/workflows/docs-actions.yml
@@ -11,7 +11,8 @@ on:
     paths:
       - 'src/main/**'
       - 'src/test/resources/cloudformation/**'
-      - 'docs/services/**'
+      - 'docs/**'
+      - 'mkdocs.yml'
       - 'tools/docs/**'
       - 'Makefile'
       - '.github/workflows/docs-actions.yml'
@@ -21,7 +22,8 @@ on:
     paths:
       - 'src/main/**'
       - 'src/test/resources/cloudformation/**'
-      - 'docs/services/**'
+      - 'docs/**'
+      - 'mkdocs.yml'
       - 'tools/docs/**'
       - 'Makefile'
       - '.github/workflows/docs-actions.yml'
@@ -35,7 +37,7 @@ concurrency:
 
 jobs:
   action-tables:
-    name: Action tables in sync
+    name: Docs in sync
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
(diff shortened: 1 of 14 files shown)


pgermosen (approved):
Checked the central claim against the actual workflows rather than taking it on faith, release.yml really does only publish latest/-compat variants, no -jvm anywhere, and the old quick-start.md genuinely referenced it. Good fix too, it's not just a tag swap, the tab now correctly explains what -compat actually gives you.

The new reverse-direction service-matrix check is sound, and I like that it's tested against the actual elb/elb-classic bug it's fixing rather than a synthetic case. The CloudFormation relative-link fix is a real consequence of turning on strict mkdocs link validation, confirms the new gates are actually catching things, not just described. CloudTrail dedup left exactly one entry, verified that directly.

No dependency changes, no em-dashes.

PR #2981, opened 2026-09-02
feat(lambda): mint EKS bearer tokens natively for the kubernetes executor

pgermosen:
## Summary

Follow-up to #2919: dropping the Fabric8 Kubernetes client for a plain REST client meant kubeconfigs using the `aws eks get-token --cluster-name <name>` exec-credential plugin (what `aws eks update-kubeconfig` generates for a real EKS cluster) got rejected outright instead of authenticating. Adds `EksTokenMinter`, which recognizes that exec shape and mints the bearer token itself: a presigned `sts:GetCallerIdentity` URL with `x-k8s-aws-id` bound into the signature, base64url-encoded and prefixed `k8s-aws-v1.`, matching the wire format `aws-iam-authenticator` and the AWS CLI's own `aws eks get-token` use. No subprocess is spawned and no SDK dependency is added, so `KubernetesApiClient` stays as dependency-free as #2919 left it.

Credentials are resolved from `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN`, falling back to the shared credentials file (`AWS_PROFILE`, default `~/.aws/credentials`). `--role-arn` and any other exec command or auth-provider plugin still fail with a named error rather than being silently mishandled; SSO/IMDS/container-credentials/IRSA aren't implemented.

This is unrelated to Floci's own local EKS emulation (`docs/services/eks.md`, the k3s-backed `aws eks create-cluster` quickstart), worth spelling out since both involve `aws eks get-token`. In that flow:

1. `aws eks get-token --cluster-name <name>` runs.
2. Floci is not called at this step. The AWS CLI builds the presigned `sts:GetCallerIdentity` URL entirely locally, using local credentials and the local clock, no network call happens here. This is true of any exec-credential plugin, not specific to Floci.
3. The token is printed directly by the AWS CLI itself, Floci does not return it.
4. When `kubectl` is invoked, it talks directly to the k3s node Floci started, not through Floci as a proxy.
5. k3s doesn't recognize the token, so its configured authentication webhook POSTs a `TokenReview` to Floci. This is the only point Floci is involved in that flow, and it only checks for the `k8s-aws-v1.` prefix, not the actual signature or the URL inside it.

This PR is about the opposite direction: Floci itself acting as a client, connecting out to some real, externally-owned EKS cluster to run Lambda pods, where the token actually needs to be a real, verifiable credential.

## Usage

1. Point `kubectl` at the real EKS cluster the usual way:
   ```
   aws eks update-kubeconfig --name <cluster> --region <region>
   ```
   This writes a kubeconfig whose exec block looks like:
   ```yaml
   exec:
     command: aws
     args: [--region, <region>, eks, get-token, --cluster-name, <cluster>, --output, json]
   ```
2. Make sure Floci itself can see AWS credentials for that cluster: either `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` (and `AWS_SESSION_TOKEN` if temporary) in its own environment, or a profile in `~/.aws/credentials` selected via `AWS_PROFILE`. If `update-kubeconfig` was run with `--profile <p>`, that's already threaded into the kubeconfig's exec block automatically, no extra step needed.
3. Set `FLOCI_SERVICES_LAMBDA_EXECUTOR=kubernetes`, and point `KUBECONFIG` at that file (or leave it at the default `~/.kube/config`).

From there it's transparent: Floci reads the same kubeconfig `kubectl` would, recognizes the `aws eks get-token` exec block, and mints the bearer token itself instead of shelling out to the AWS CLI. Nothing to opt into, no changes needed to the kubeconfig.

One limit worth knowing: `update-kubeconfig --role-arn <arn>` isn't supported. Floci fails with a named error rather than silently authenticating as the wrong identity, so if cluster access requires assuming a role, resolve that to session credentials first and point Floci at those instead.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Matches the `aws-iam-authenticator`/`aws eks get-token` presigned-URL token format exactly (verified against its public source, since I don't have an AWS account to test live signature acceptance against real STS, noted as an open gap). Deliberately narrower than the AWS CLI's default credential chain: no `--role-arn` (assume-role), no SSO, no IMDS/container-credentials/IRSA. Each is rejected with a specific error rather than silently mishandled.

## Checklist


hectorvent (approved):
Thanks for closing the gap #2919 opened without adding a dependency, and for chasing every one of the bot rounds to the source instead of patching the symptom. I checked the result against the CLI and the SDK metadata: `update_kubeconfig.py` writes `--region` ahead of `eks get-token`, abbreviates `--role-arn` to `--role`, and threads `AWS_PROFILE` through `exec.env` ([update_kubeconfig.py](https://github.com/aws/aws-cli/blob/v2/awscli/customizations/eks/update_kubeconfig.py)), and the recognizer, the role guard and the credential precedence now match all three. The eight STS suffixes match `partitions.json` exactly ([partitions.json](https://github.com/boto/botocore/blob/develop/botocore/data/partitions.json)). The fixed-clock signature test is the right kind of proof.

**(follow-up, separate PR)** The description's item 5 still says Floci only checks the `k8s-aws-v1.` prefix; #2912 validates the signature, expiry and cluster binding now, so worth a one-line edit before the squash.

No blockers from my side.

PR #2982, opened 2026-09-02
fix(ecr): clean up registry storage after deletion

simonepri:
## Problem

`DeleteRepository --force` leaves repository manifest data and the shared registry volume behind after the final repository or a non-retained runtime is deleted.

## Solution

Remove the deleted repository's registry storage, including untagged manifests. Remove the namespaced registry volume when the final repository is gone or a runtime configured not to persist is stopped, but only for `memory` storage or when `prune-volumes-on-delete: true` is set. Persistent storage remains retained by default.


pgermosen (approved):
Approving. Checked the path-traversal angle since this does rm -rf on a path built from a repo name - the new segment check is solid, and it's actually redundant with validateRepoName's regex, which is AWS's real ECR naming pattern and structurally can't produce a "." or ".." segment anyway. Exec args go through Docker's API as an array, not a shell, so no injection surface there either.

Traced the failure-atomicity and cross-account claims against both the code and the tests: registry cleanup runs before metadata deletion so a failure leaves the repo intact, and hasRepositories() scans all accounts before pruning the shared volume, both backed by tests that actually exercise the failure/other-account paths rather than just the happy one. Good work.

PR #2985, opened 2026-09-03
fix(cloudformation): keep generated names across updates for Scheduler, SQS and MicrovmImage

hectorvent:
## Summary

Follow-up to #2788, which left `AWS::SQS::Queue` and `AWS::Lambda::MicrovmImage` with the regenerate-on-update defect and whose review flagged `SchedulerScheduleGroupCfnProvisioner` reading the physical id off the resource, the anti-pattern AGENTS.md step 4 now documents. All three names are `createOnlyProperties` in their registry schemas, so an unchanged template must keep its physical id.

- **Scheduler ScheduleGroup**: `stablePhysicalName` and `reusesPriorEntity` replace the `ConflictException` catch; the tag reconcile uses `ProvisionContext.staleTagKeys`. One behaviour change, which the first version of this description got wrong: the old path called `createScheduleGroup` and adopted only on `ConflictException`, so a group deleted outside the stack was silently recreated; the new path reads the group first and surfaces `ResourceNotFoundException`. That is what CloudFormation does, and neither LocalStack nor moto recreates a resource deleted out of band either.
- **SQS Queue**: the physical id is the queue URL, so the prior name is read from the `QueueName` attribute (the `SnsCfnProvisioner` pattern), only while its `.fifo` suffix still matches `FifoQueue`, which is createOnly too. On reuse the update goes through `SetQueueAttributes`, the schema's update handler, because `createQueue` on an existing name answers `QueueAlreadyExists` once any attribute differs. That also fixes an explicitly named queue whose `VisibilityTimeout` changes. An explicitly named queue whose `FifoQueue` changes now fails with `ValidationError` ("Updating FifoQueue requires resource replacement, which is not supported."), the same refusal the other provisioners use for a createOnly change under a kept name, instead of reconciling every other attribute and reporting success with the queue still in its old mode: the schema's own `QueueName` text says a named queue cannot take an update that requires replacement. `AWS::SQS::QueuePolicy` keeps its id across updates, the same fix `S3::BucketPolicy` got.
- **Lambda MicrovmImage**: `stablePhysicalName`, and on reuse `updateImage` (the schema's `UpdateMicrovmImage` handler) instead of `createImage`, which mints a new version on every call.

Follow-ups, deliberately out of scope: SQS `Tags` (the provisioner ignores them today), the `NetworkConnector` name (its physical id is the connector id, so it needs its own reuse branch), resetting an attribute the template dropped (`SetQueueAttributes` only removes on an explicit empty value), recreating a resource deleted out of band the way `EcsCapacityCfnProvisioner` does if that leniency is wanted (today all three fail the update, like LocalStack and moto), and MicrovmImage's `Ref` resolving to the name where the schema's `primaryIdentifier` is `ImageArn` (pre-existing).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

No wire shape changes. Behaviour checked against the registry schemas in `CloudformationSchema.zip`: `aws-scheduler-schedulegroup.json` (Name createOnly, update handler is a tag diff), `aws-sqs-queue.json` (QueueName and FifoQueue createOnly, update handler `sqs:SetQueueAttributes`), `aws-lambda-microvmimage.json` (Name createOnly, update handler `lambda:UpdateMicrovmImage`). Before this change an unnamed SQS queue got a new URL on every `UpdateStack`, and a named queue with a changed attribute failed the update with `QueueAlreadyExists`.

## Checklist

- [x] `./mvnw test` passes locally (focused: the seven provisioner unit classes, 62/62, and the CloudFormation integration classes touched)
- [x] New or updated integration test added (`updateStack_unnamedQueueKeepsItsUrlAndReconcilesAttributes`; three new `CfnStableNameUpdatePathTest` cases; new `LambdaMicrovmsCfnProvisionerTest`; `flippingFifoQueueOnANamedQueueIsRefused` and `aGroupDeletedOutOfBandFailsTheUpdate` from the review round)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/SqsCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/SqsCfnProvisioner.java
@@ -2,6 +2,7 @@ package io.github.hectorvent.floci.services.cloudformation.provisioners;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import io.github.hectorvent.floci.core.common.AwsArnUtils;
+import io.github.hectorvent.floci.core.common.AwsException;
 import io.github.hectorvent.floci.services.cloudformation.model.StackResource;
 import io.github.hectorvent.floci.services.sqs.SqsService;
 import io.github.hectorvent.floci.services.sqs.model.Queue;
@@ -36,7 +37,7 @@ public class SqsCfnProvisioner implements CfnResourceProvisioner {
     public void provision(StackResource r, JsonNode props, ProvisionContext ctx) {
         switch (r.getResourceType()) {
             case "AWS::SQS::Queue" -> provisionQueue(r, props, ctx);
-            case "AWS::SQS::QueuePolicy" -> provisionQueuePolicy(r);
+            case "AWS::SQS::QueuePolicy" -> provisionQueuePolicy(r, ctx);
             default -> throw new IllegalStateException("SqsCfnProvisioner cannot handle " + r.getResourceType());
         }
     }
@@ -55,12 +56,23 @@ public class SqsCfnProvisioner implements CfnResourceProvisioner {
                 : null;
         boolean fifo = "true".equalsIgnoreCase(fifoFlag);
         String queueName = ctx.resolveOptional(props, "QueueName");
+        // The physical id is the queue URL, so ctx.stablePhysicalName does not fit: the prior name
+        // comes from the QueueName attribute recorded at create time, as SnsCfnProvisioner reads
+        // the topic name beside its ARN. Keeping it means an unnamed queue survives an update
+        // instead of being orphaned with its messages. FifoQueue is createOnly like QueueName, so a
+        // prior name whose .fifo suffix no longer matches the flag is a replacing update and gets a
+        // fresh name.
+        String priorName = r.getAttributes() != null ? r.getAttributes().get("QueueName") : null;
         if (queueName == null || queueName.isBlank()) {
-            // Like real CloudFormation, generated names of FIFO queues must end in .fifo
-            // (SqsService rejects FifoQueue=true otherwise). Keep within the 80-char limit.
-            queueName = fifo
-                    ? ctx.generatePhysicalName(r.getLogicalId(), 75, false) + ".fifo"
-                    : ctx.generatePhysicalName(r.getLogicalId(), 80, false);
+            if (priorName != null && !priorName.isBlank() && priorName.endsWith(".fifo") == fifo) {
+                queueName = priorName;
+            } else {
+                // Like real CloudFormation, generated names of FIFO queues must end in .fifo
+                // (SqsService rejects FifoQueue=true otherwise). Keep within the 80-char limit.
+                queueName = fifo
+                        ? ctx.generatePhysicalName(r.getLogicalId(), 75, false) + ".fifo"
+                        : ctx.generatePhysicalName(r.getLogicalId(), 80, false);
+            }
         }
         Map<String, String> attrs = new HashMap<>();
         if (props != null) {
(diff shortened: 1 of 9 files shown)


src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/SqsCfnProvisioner.java
```diff
@@ -88,18 +99,35 @@ private void provisionQueue(StackResource r, JsonNode props, ProvisionContext ct
...
-        Queue queue = sqsService.createQueue(queueName, attrs, ctx.region());
+        // provision is also the update path. createQueue on a name that exists hands back the
+        // queue only when every attribute matches and answers QueueAlreadyExists otherwise, so the
+        // second UpdateStack that changes VisibilityTimeout must go through SetQueueAttributes, the
+        // registry schema's update handler, rather than a second create. The prior physical id is
+        // the queue URL SetQueueAttributes addresses. A replacing update derives a different name
+        // and still creates.
+        String queueUrl;
+        if (ctx.isUpdate() && queueName.equals(priorName)) {
+            attrs.remove("FifoQueue");
```
pgermosen, line 120: For an explicitly-named queue this branch is reachable with FifoQueue changed and priorName unchanged (the .fifo-suffix replacement check above only runs in the unnamed branch), so this silently drops the FIFO-mode change instead of doing anything AWS actually does here.

Checked the docs rather than guess: AWS::SQS::Queue's own reference says explicitly that if you provide a QueueName, you can't perform an update that requires replacement (FifoQueue is one), you'd need a new name instead. That's a documented CloudFormation-wide rule (S3, IAM, Elastic Beanstalk all say the same), not SQS-specific: a named resource can't be replaced because CloudFormation can't mint a new name for the replacement while the old one is still in use, so it fails the whole update rather than doing anything partial.

So the real options here are "reject the update" or "replace under a name the caller must supply," not "reconcile everything except FifoQueue." Current behavior reports UPDATE_COMPLETE while the queue silently keeps its old mode, which is the kind of divergence that surfaces somewhere else entirely, whatever depends on the queue actually being FIFO. Worth failing this case explicitly rather than reconciling through it. flippingFifoQueueIsAReplacingUpdate only covers the unnamed case, so this one has no coverage either.

pgermosen (commented):
Traced both of greptile's findings before weighing in.

The missing-backing-entity one doesn't hold up: all three reuse branches (setQueueAttributes, getScheduleGroup, updateImage->getImage) throw a proper AWS-shaped not-found exception when the entity is gone, which is what real CloudFormation actually does too, it doesn't self-heal an out-of-band-deleted resource either. Worth noting the Scheduler description's "behaviour unchanged" isn't quite right though: the old code tried createScheduleGroup first and only reconciled on a ConflictException, so a deleted group got silently recreated; the new reusesPriorEntity check skips straight to getScheduleGroup and would now surface a real error instead. That's arguably more correct, just not actually unchanged.

The FIFO one is real, and worse than it looks. Left inline.

hectorvent:
Thanks for tracing both, pgermosen. You are right on each.

The FIFO flip on an explicitly named queue was reachable exactly as you describe: the `.fifo` suffix check only guarded the generated-name branch, so a named queue whose `FifoQueue` changed fell into the `SetQueueAttributes` path and reported `UPDATE_COMPLETE` with the old mode. It now fails with `ValidationError` ("Updating FifoQueue requires resource replacement, which is not supported."), the same refusal the other provisioners use for a createOnly change under a kept name, and `flippingFifoQueueOnANamedQueueIsRefused` pins it. The rule I am leaning on is the SQS schema's own `QueueName` text, a named queue cannot take an update that requires replacement ([aws-sqs-queue.json](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-sqs/blob/master/aws-sqs-queue/aws-sqs-queue.json)); for what it is worth neither LocalStack nor moto refuses this case, LocalStack replaces and moto ignores the flip, so the refusal is our convention aligned with the schema rather than something either reference implements.

On Scheduler, "behaviour unchanged" was wrong and the description and commit now say so: the old path recreated a group deleted out of band, the new one surfaces `ResourceNotFoundException`, which is what CloudFormation does and what both reference implementations do too. `aGroupDeletedOutOfBandFailsTheUpdate` pins that. `EcsCapacityCfnProvisioner` is the one sibling that recreates instead; if we want that leniency across the board it is a separate change, listed in the follow-ups.

pgermosen (approved):
Both of the round-1 findings are fixed and I checked the fixes against the actual CloudFormation docs rather than just re-reading the description. QueueName and FifoQueue are both genuinely Update-requires-Replacement on AWS::SQS::Queue, and AWS's own docs are explicit that a named queue can't take a replacement update, it has to fail, which is exactly what the new ValidationError path does instead of silently reconciling everything but the FIFO mode. Scheduler's Name vs Tags split matches too.

Traced the logic across all three provisioners: Scheduler and MicrovmImage reuse the existing stablePhysicalName/reusesPriorEntity helpers correctly, and SQS's hand-rolled version makes sense given the physical id is the queue URL rather than the name, with a second createOnly property to guard. Stripping FifoQueue before SetQueueAttributes is correct too, that call doesn't accept it. Test coverage lines up with the logic exactly, including the replacing-update vs refused-update split and the full CreateStack/UpdateStack integration path.

Nice catch on the description accuracy too. Nothing else outstanding.

PR #2986, opened 2026-09-03
fix(cloudformation): preserve list intrinsics and conditional tags

### LINKED ISSUE #2983: [BUG] CloudFormation: a list-valued intrinsic (Fn::If, Fn::Split) collapses to a string, so Fn::Join over it is empty and Fn::If-wrapped Tags are dropped silently
## Service

CloudFormation

## AWS API Action

`CreateStack` / `UpdateStack`: template intrinsic resolution for list-valued intrinsics (`Fn::If` choosing between lists, `Fn::Split`, `Fn::Join` over either, and a `Tags` property wrapped in `Fn::If`).

## Expected behavior

`Fn::If` can select between two lists, `Fn::Split` returns a list, and `Fn::Join` accepts a list-valued intrinsic as its second argument. On AWS the three templates below produce:

- `/fnjoin/if` = `a,b` and `/fnjoin/split` = `x|y|z`
- a Kinesis stream and a DynamoDB table each carrying the tag `env=prod`

## Actual behavior

All three resolve to nothing, and the stack still reports `CREATE_COMPLETE` with no resource status reason, so nothing tells the user a value was lost.

```
$ aws ssm get-parameter --name /fnjoin/if --query Parameter.Value --output text
(empty string)
$ aws ssm get-parameter --name /fnjoin/split --query Parameter.Value --output text
(empty string)
$ aws kinesis list-tags-for-stream --stream-name fnif-stream --query Tags
[]
$ aws dynamodb list-tags-of-resource --resource-arn arn:aws:dynamodb:us-east-1:000000000000:table/fnif-table --query Tags
[]
```

A control stream in the same stack with a plain `Tags` list gets its tag, so this is specific to the intrinsic wrapping.

## Cause

`CloudFormationTemplateEngine.resolveNode` (`CloudFormationTemplateEngine.java:115-149`) sends any object carrying an intrinsic through `TextNode.valueOf(resolve(node))`, and `resolve` is `String`-typed, so a list-valued result is collapsed to a string before the array branch below it can run. Downstream:

- `resolveJoin` (`:234-248`) indexes its second argument as an array; an intrinsic object has size 0, so the join is `""`.
- `resolveIf` (`:377-384`) returns a `String`, so the chosen list is flattened.
- A list-aware `resolveList` exists (`:267`) but is only reached from `Fn::Select`.
- Every tag reader checks `isArray()` on the raw or resolved node and bails silently: `ProvisionContext.resolveTags` (`provisioners/ProvisionContext.java:61`), the monolith's `parseCfnTags` (`CloudFormationResourceProvisioner.java:4792`), `KinesisCfnProvisioner` (`:79`), and `OrganizationsCfnProvisioner` (`:288`), which returns `null`, the value it uses for "the template said nothing", so an update with `Fn::If` tags neither tags nor untags.

#2946 adds a list-aware `resolveStringList` for the monolith's string-list properties and teaches `resolveList` about `Fn::If`, which fixes `SubnetIds`-style properties, but it does not touch `resolveNode`, `resolveJoin` or the tag readers, so the cases above stay open after it. `SchedulerScheduleGroupCfnProvisioner` carries a `tagsAreResolvable` guard as a local workaround; a fix in the engine would let that go.

## Reproduction

```bash

cat > fnjoin.json <<'JSON'
{ "Conditions": { "IsProd": { "Fn::Equals": ["prod", "prod"] } },
  "Resources": {
    "P": { "Type": "AWS::SSM::Parameter", "Properties": { "Name": "/fnjoin/if", "Type": "String",
           "Value": { "Fn::Join": [",", { "Fn::If": ["IsProd", ["a", "b"], ["c"]] }] } } },
    "Q": { "Type": "AWS::SSM::Parameter", "Properties": { "Name": "/fnjoin/split", "Type": "String",
           "Value": { "Fn::Join": ["|", { "Fn::Split": [",", "x,y,z"] }] } } } } }
JSON
aws cloudformation create-stack --stack-name fnjoin --template-body file://fnjoin.json
aws cloudformation wait stack-create-complete --stack-name fnjoin
aws ssm get-parameter --name /fnjoin/if --query Parameter.Value --output text      # expected a,b
aws ssm get-parameter --name /fnjoin/split --query Parameter.Value --output text   # expected x|y|z

cat > fnif-tags.json <<'JSON'
{ "Conditions": { "IsProd": { "Fn::Equals": ["prod", "prod"] } },
  "Resources": {
    "Stream": { "Type": "AWS::Kinesis::Stream", "Properties": { "Name": "fnif-stream", "ShardCount": 1,
      "Tags": { "Fn::If": ["IsProd", [{"Key": "env", "Value": "prod"}], [{"Key": "env", "Value": "dev"}]] } } },
    "Table": { "Type": "AWS::DynamoDB::Table", "Properties": { "TableName": "fnif-table", "BillingMode": "PAY_PER_REQUEST",
      "AttributeDefinitions": [{"AttributeName": "pk", "AttributeType": "S"}], "KeySchema": [{"AttributeName": "pk", "KeyType": "HASH"}],
      "Tags": { "Fn::If": ["IsProd", [{"Key": "env", "Value": "prod"}], [{"Key": "env", "Value": "dev"}]] } } },
    "Control": { "Type": "AWS::Kinesis::Stream", "Properties": { "Name": "plain-stream", "ShardCount": 1,
      "Tags": [{"Key": "env", "Value": "prod"}] } } } }
JSON
aws cloudformation create-stack --stack-name fnif-tags --template-body file://fnif-tags.json
aws cloudformation wait stack-create-complete --stack-name fnif-tags
aws kinesis list-tags-for-stream --stream-name fnif-stream --query Tags     # expected env=prod, got []
aws kinesis list-tags-for-stream --stream-name plain-stream --query Tags    # env=prod (control)
aws dynamodb list-tags-of-resource --resource-arn arn:aws:dynamodb:us-east-1:000000000000:table/fnif-table --query Tags   # got []
```

## Environment

- Floci version / image tag: built from `main` at 309043baf (engine unchanged through 4ce53b51d), reports `floci 2.0.1 on JVM`
- Java SDK version (if applicable): AWS CLI 2.33.24
- How you're running Floci (Docker / native / `mvn quarkus:dev`): `java -jar target/quarkus-app/quarkus-run.jar` on macOS

7487:
## Summary

Fixes #2983. CloudFormation now preserves list-valued intrinsic results, so joining an `Fn::If` or `Fn::Split` produces the requested value and conditional tag lists reach resources.

- Keep the selected `Fn::If` branch's JSON type and preserve arrays from `Fn::Split`, `Fn::GetAZs`, and `Fn::Cidr`. Reuse the existing list resolver for `Fn::Join`.
- Resolve tag properties before checking for arrays in the affected provisioners. Keep the existing blank-key, absent-value, and malformed-tag handling.
- Wire DynamoDB table tags through the existing tag/untag operations. Its CloudFormation path previously ignored even literal tags; updates now reconcile removed keys.

The subnet-property work in #2946 is separate; these changes cover the Join and Tags reproductions that remain outside that PR.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

AWS permits `Fn::If` and `Fn::Split` in the [Fn::Join value list](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/intrinsic-function-reference-join.html), and documents [conditional array values](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/intrinsic-function-reference-conditions.html#intrinsic-function-reference-conditions-conditional-array-values).

The regression test creates and updates a stack, checks joined outputs, and reads tags through the Kinesis, DynamoDB, and Scheduler APIs. It covers both condition branches, an empty tag list, and empty split elements. The Scheduler malformed-tag guard remains covered.

## Validation

- TDD: reproduced empty Join output and missing tags before the fix; the focused CloudFormation suite passes all 417 tests.
- Full suite: all 1,138 test classes assigned once using the repository's four-way CI partitioner. Maven reports 15,289 tests: 15,280 passed, 9 skipped, 0 failures, 0 errors.
- The full suite uses `./mvnw test -B -Dsurefire.includesFile=...`. Two network-sensitive classes run separately: all 8 `ContainerNetworkReachabilityTest` cases in an offline Linux container, and all 6 `ElbV2LambdaTargetDataPlaneIntegrationTest` cases with `-Dfloci.hostname=localhost`. Other tests retain the default hostname.
- `git diff --check`, `make docs-check`, Java 25 clean main/test compilation, and `./mvnw -B package -DskipTests` pass. The repository has no separate Java lint or typecheck target.
- A boto3 1.43.87 smoke test against the packaged application verifies CreateStack and UpdateStack outputs plus Kinesis, DynamoDB, and Scheduler tags for both condition branches. The test stack is deleted afterward.

## Checklist

- [x] Complete Maven test suite passes locally
- [x] New integration regression test added


hectorvent:
Thank you, this is the engine-level fix #2983 was after: `resolveNode` keeping the selected `Fn::If` branch's type and `Fn::Join` going through `resolveList` covers the join and tag cases in one place, and reading the tags back through the Kinesis, DynamoDB and Scheduler APIs pins it properly.

**(blocking)** The branch now conflicts with main in the two Scheduler files: #2985 rewrote the provisioner's name and adopt path (`ctx.stablePhysicalName`, `ctx.reusesPriorEntity`) and the matching test. Would you rebase onto main? Your `resolveNode` line and the `"not-a-list"` test case carry over unchanged, and the `resolveNode` passthrough belongs in the test's `ctx(String)` helper now.

**(follow-up, separate PR)** #2946 added its own inline `Fn::If` arm to `resolveList` and `isListValuedIntrinsic`. Both could call your `selectIfBranch`, otherwise the second block in `resolveList` is dead code. Fine to fold in here if trivial, or leave it.

A rebase is all that unblocks this from my side.

7487:
Merged main (d914b52). The two Scheduler files now sit on #2985's `ctx.stablePhysicalName` / `ctx.reusesPriorEntity` path with only the `resolveNode` line and the `"not-a-list"` case carried over; the `resolveNode` passthrough lives in the test's `ctx(String)` helper. Dropped the now-stale "engine collapses a list-valued intrinsic to a string (#2983)" comment since that is what this PR fixes.

Also folded in the follow-up: `resolveList` and `isListValuedIntrinsic` now go through `selectIfBranch`, and the unreachable second `Fn::If` block in `resolveList` is gone (net -20 lines in the engine). Behaviour is unchanged for a malformed `Fn::If`: `selectIfBranch` returns an empty text node, so `isListValuedIntrinsic` still answers false.

`SchedulerScheduleGroupCfnProvisionerTest`, `CloudFormationTemplateEngineTest` and `CloudFormationListIntrinsicsIntegrationTest` pass (41/41); the whole `services.cloudformation` package is 1337/1338, the one failure being `SamTransformIntegrationTest.samFunction_withAutoPublishAlias_createsVersionAndAlias`, which fails identically on untouched main here because the Lambda invoke needs Docker.

7487:
Merged main again (7f67ca0). The only conflict was in `CloudFormationTemplateEngine`: #2996 landed its own `selectIfBranch` plus the `Fn::If` arms in `resolveNode` and `resolveList`, so I kept main's versions (null on a malformed `Fn::If`, with its comments) and dropped mine. What remains from this PR in the engine is the `Fn::Split`/`Fn::GetAZs`/`Fn::Cidr` array preservation in `resolveNode`, `Fn::Join` going through `resolveList`, and `isListValuedIntrinsic` routed through `selectIfBranch` (with the null check). Provisioner and test changes merged cleanly.

`CloudFormationTemplateEngineTest`, `SchedulerScheduleGroupCfnProvisionerTest`, `CloudFormationListIntrinsicsIntegrationTest` and `CloudFormationIntegrationTest` (which carries #2996's new cases) pass locally, 203/203.

hectorvent (approved):
Thanks for the two merges: the engine now carries only what #2996 did not (`Fn::Split`/`Fn::GetAZs`/`Fn::Cidr` preserved as arrays by `resolveNode`, `Fn::Join` over `resolveList`, and the `Fn::If` arms all through `selectIfBranch`), the tag call sites resolve the whole property before the `isArray` check, and the DynamoDB reconcile closes the other half of #2983. Ran the CloudFormation and DynamoDB packages on the merge result with main: 2,079 tests green. Approving.

PR #2992, opened 2026-09-03
perf(build): drop KotlinPoet and ICU4J from the protobuf schema path

exoego:
## Summary

Glue emulations uses Wire.
Wire's schema parser drags in KotlinPoet and ICU4J, which is not needed since Floci just parses .proto and never generates code.

Reducing image size 274.4 to 260.2 MiB, 88.7 to 85.0 MB gzipped.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No change.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/GlueSchemaRegistryTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/GlueSchemaRegistryTest.java
@@ -148,6 +148,24 @@ class GlueSchemaRegistryTest {
                     + "  string email = 2;\n"
                     + "}\n";
 
+    private static final String PROTOBUF_SCHEMA_RESPACED =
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Approving. Traced the actual imports rather than just trusting the claim - floci only touches Wire's parser layer (FieldElement, Field, Location) through Apicurio's validator/canonicalizer wrappers, never the code-gen packages that pull in KotlinPoet and ICU4J. The new SDK test is a real end-to-end guard too: it rejects an invalid protobuf schema through both CheckSchemaVersionValidity and CreateSchema, then re-registers a whitespace-only variant and asserts it canonicalizes to the same version rather than minting a new one, which would catch a class-loading or canonicalization regression if the exclusion broke anything. Clean, well-scoped change.

PR #2993, opened 2026-09-03
fix(athena): register every Glue database as a DuckDB schema so <db>.<table> resolves

### LINKED ISSUE #2859: [BUG] athena: floci-duck does not register Glue catalog tables — "schema does not exist"
### Description

Athena queries fail with a DuckDB catalog error even though the Glue database and tables exist and the data is present in S3. The `floci-duck` sidecar never registers the Glue catalog tables into DuckDB, so any `SELECT ... FROM <db>.<table>` fails with `schema does not exist`.

This also reproduces with the **official Athena + S3 example from floci.io/aws**, so it does not appear to be specific to a particular table definition.

### Environment

- floci image: `floci/floci:latest` (health reports version **1.7.0**, edition `floci-always-free`)
- floci-duck image: `floci/floci-duck:latest`
- Host: macOS (Docker Desktop, arm64)
- Endpoint: `http://localhost:4566`

### Steps to reproduce (official docs example)

```bash

aws s3 mb s3://analytics-bucket
printf '{"customer":"ana","amount":10}\n{"customer":"leo","amount":7}\n' > orders.json
aws s3 cp orders.json s3://analytics-bucket/orders/orders.json
aws glue create-database --database-input '{"Name":"shop"}'

QID=$(aws athena start-query-execution \
  --query-string "SELECT customer, SUM(amount) FROM shop.orders GROUP BY customer" \
  --result-configuration OutputLocation=s3://analytics-bucket/results/ \
  --query 'QueryExecutionId' --output text)

aws athena get-query-execution --query-execution-id "$QID" \
  --query 'QueryExecution.Status.StateChangeReason' --output text
```

### Expected

Query succeeds and returns the aggregated rows (as shown in the docs example).

### Actual

Query fails. `floci-duck` logs:

```
Athena mode detected. Output path: s3://analytics-bucket/results/<id>.csv
Executing final SQL: COPY (SELECT customer, SUM(amount) FROM shop.orders GROUP BY customer) TO 's3://...' (FORMAT CSV, HEADER);
ERROR: Catalog Error: Table with name "shop.orders" does not exist because schema "shop" does not exist.
```

### Additional notes

- Confirmed the Glue database and tables exist via `aws glue get-databases` / `aws glue get-tables`.
- Confirmed the data exists in S3.
- Also tried: creating the table explicitly with `aws glue create-table` (with `classification=csv` and `OpenCSVSerde`), setting `LocationUri` on the database, and restarting the `floci` container — same error.
- DDL workaround is not possible: floci wraps every statement in `COPY (<query>) TO 's3://...'`, so `CREATE SCHEMA` / `CREATE TABLE` fail with `Parser Error: syntax error at or near "CREATE"`.
- The `httpfs` extension loads fine in floci-duck, so S3 file reads work; only the Glue → DuckDB catalog registration appears to be missing.

vuthehuyht:
## Summary

Closes #2859

Athena query execution failed with `schema "<db>" does not exist` when referencing tables with qualified names (e.g. `FROM shop.orders`) or when omitting the database in `QueryExecutionContext`. Previously, `buildGlueDdl` only read tables for the context database (or `default`) and created unqualified views in DuckDB's default `main` schema.

This change:
1. Extracts DDL generation into an `@ApplicationScoped` `GlueViewDdlBuilder`.
2. Registers each Glue database as a DuckDB schema (`CREATE SCHEMA IF NOT EXISTS "<db>"`) and registers views qualified as `"<db>"."<table>"`.
3. In addition, creates unqualified view copies `"<table>"` in `main` for the context database so both `FROM db.table` and `FROM table` resolve seamlessly.
4. Optimizes table retrieval so the context database tables are fetched only once during DDL construction.
5. Isolates table errors individually so one invalid table does not break the entire catalog registration.
6. Propagates `getDatabases()` unrecoverable storage errors so query execution reports the real root cause rather than an empty catalog.
7. Updates documentation to explain schema qualification and removes em-dashes.
8. Adds comprehensive unit tests covering schema creation, alias creation, single-fetch verification, error propagation, identifier escaping, and S3 location literal escaping.
9. Adds an end-to-end compatibility test in `AthenaTest` executing qualified queries against DuckDB via the Athena SDK.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- In AWS Athena, queries can reference tables using qualified names (`FROM <db>.<table>`) regardless of the default database specified in `QueryExecutionContext`.
- Floci previously threw `schema "<db>" does not exist` because only the context database was loaded as unqualified views in the DuckDB default schema.
- Verified using AWS SDK for Java v2 (AthenaClient, GlueClient, S3Client).

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AthenaTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AthenaTest.java
@@ -152,22 +152,30 @@ class AthenaTest {
                 )
         );
 
-        GetTableMetadataResponse response = athena.getTableMetadata(
-                GetTableMetadataRequest.builder()
-                        .catalogName("AwsDataCatalog")
-                        .databaseName(dbName)
-                        .tableName(tableName)
-                        .build()
-        );
+        try {
+            GetTableMetadataResponse response = athena.getTableMetadata(
+                    GetTableMetadataRequest.builder()
+                            .catalogName("AwsDataCatalog")
+                            .databaseName(dbName)
+                            .tableName(tableName)
+                            .build()
+            );
 
-        assertThat(response.tableMetadata().createTime())
-                .as("createTime must be parseable by the AWS SDK")
-                .isNotNull();
-        assertThat(response.tableMetadata().lastAccessTime())
-                .as("lastAccessTime must be parseable by the AWS SDK")
-                .isNotNull();
-
-        glue.close();
+            assertThat(response.tableMetadata().createTime())
+                    .as("createTime must be parseable by the AWS SDK")
+                    .isNotNull();
+            assertThat(response.tableMetadata().lastAccessTime())
+                    .as("lastAccessTime must be parseable by the AWS SDK")
+                    .isNotNull();
+        } finally {
+            try {
+                glue.deleteTable(r -> r.databaseName(dbName).name(tableName));
+                glue.deleteDatabase(r -> r.name(dbName));
+            } catch (Exception ignored) {
+                // Best-effort test cleanup
+            }
+            glue.close();
+        }
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


pgermosen (commented):
This PR skips the repo's PR template entirely (no Type of change, no AWS Compatibility section, no Checklist) - that's required here, so please add it.

On the substance, I traced this in detail rather than just reading the description. The core fix is right: CREATE SCHEMA per Glue database, qualified views for every table, and a separate unqualified alias pass for the context database, which is exactly what makes both FROM db.table and FROM table resolve. The error-handling split is deliberate and correct too - getDatabases() failures propagate (verified by testGetDatabasesExceptionPropagates), while getTables() failures are isolated per-database so one broken database or table doesn't blank out the whole catalog. Identifier quoting and the S3 location's single-quote escaping are both handled correctly for the injection angle. Docs are updated accurately and follow the no-em-dash house style.

One real gap: there's no end-to-end test that actually runs a qualified query (SELECT ... FROM shop.orders, the exact shape from the linked issue) through a live floci-duck and asserts SUCCEEDED. GlueViewDdlBuilderTest is thorough for the DDL-string-generation logic in isolation, but the original bug was a DuckDB runtime error, not a Java-level string-building bug, and nothing here proves DuckDB actually accepts and resolves the generated schema+view DDL against a real qualified query. The AthenaTest.java changes in this PR are cleanup (proper try/finally teardown), not a new reproduction test. Worth adding one compatibility test that mirrors the issue's repro before this merges.

Smaller, non-blocking note: a table in the context database gets registered twice - once qualified via the main per-database loop, once unqualified via the context-database block - so its Glue metadata gets fetched and its view created twice per query. Not wrong, just a bit redundant.

pgermosen (commented):
The Athena work itself is in good shape. Checked the redundant-fetch fix directly, the context database's tables are now reused from the main loop instead of fetched twice, and the new end-to-end test is exactly what was needed, real S3 bucket, real Glue table, the exact FROM shop.orders shape from the issue with no QueryExecutionContext, run through the real SDK against a live instance and asserted SUCCEEDED. That closes the gap from last round properly.

One thing worth holding this on though: the last commit on this branch, fix(autoscaling): prevent reconciler from resurrecting deleted ASGs, has nothing to do with Athena. It touches AutoScalingService and adds an S3Service.bucketExists() helper, neither related to Glue schema registration. Looks like it may have ended up here by accident, maybe a rebase or cherry-pick slip. Since this repo squash-merges, leaving it in would carry an unrelated AutoScaling fix into main's history under an Athena-titled commit with no separate review, worth dropping it from this branch before merging.

Also, since last round, GlueViewDdlBuilder now silently skips creating a view entirely when a table's S3 bucket doesn't exist. Worth a second look against real Athena behavior, a table with a missing backing location should probably still be nameable in a query and fail differently at read time, rather than reporting as if the table itself doesn't exist. Not confident enough to call it wrong without checking against a live account, just flagging it since it's a real behavior change nobody's looked at yet.

pgermosen (approved):
All three things from the last two rounds are addressed: template's filled in, the stray AutoScaling commit is gone, and the S3-bucket-existence skip got cleanly reverted rather than patched further, a missing location now just fails at query time through DuckDB's lazy view evaluation instead of hiding the table, which matches real Athena much better. Combined with everything else already verified (schema/view qualification, error handling split, quoting, the real end-to-end qualified-query test), this is good to go. Approving.

PR #2994, opened 2026-09-03
fix(lambda): honor configured function architecture

### LINKED ISSUE #2840: [BUG] Lambda container launcher ignores Architectures — arm64 functions run on the host's architecture
## Service

Lambda

## AWS API Action

`Invoke` (the container launcher, for functions created with `Architectures: ["arm64"]`)

## Expected behavior

A function declared `Architectures: ["arm64"]` runs on arm64. Floci stores and echoes the value correctly, so the API surface already agrees with AWS — but the container launcher never consults it, and always runs the host-native image variant.

The base images Floci uses are multi-arch. Both variants are published:

```
$ docker manifest inspect public.ecr.aws/lambda/nodejs:22
 available: linux / arm64
 available: linux / amd64
```

So the correct variant exists and is simply not selected. `docker-java` supports `withPlatform` on both the pull and the create call.

## Actual behavior

`getArchitectures()` is read in exactly two places — `LambdaController` (to echo it in the response) and `CloudFormationResourceProvisioner` (drift detection). Nothing in `services/lambda/launcher/` or `core/common/docker/` passes a platform, so the daemon resolves the host's architecture every time.

**Interpreted runtimes fail silently.** The function is declared arm64, runs on amd64, and nothing reports the mismatch:

```
$ aws $E lambda get-function-configuration --function-name arch-probe --query 'Architectures'
[ "arm64" ]

$ aws $E lambda invoke --function-name arch-probe --payload '{}' out.json --query 'StatusCode'
200

$ docker image inspect public.ecr.aws/lambda/nodejs:22 --format '{{.Os}}/{{.Architecture}}'
linux/amd64
```

**Compiled runtimes fail confusingly.** With `provided.al2023` and an aarch64 `bootstrap`, the exec fails inside the container and the runtime API never registers, so the invoke dies on the function timeout rather than on the real cause:

```
[lambda:arch-compiled-probe] /lambda-entrypoint.sh: line 14: /var/runtime/bootstrap: cannot execute binary file: Exec format error
[lambda:arch-compiled-probe] /lambda-entrypoint.sh: line 14: /var/runtime/bootstrap: Success
WARN [LambdaExecutorService] Function arch-compiled-probe timed out after 5s
```

The caller sees:

```json
{"errorMessage":"Task timed out after 5 seconds","errorType":"Function.TimedOut"}
```

`Exec format error` appears in the Floci log but nowhere in the API response, and the reported error is a timeout — which points at the handler, not at the architecture. Raising the function's timeout makes it slower without making it clearer.

## Why it matters

arm64 is the cost-optimal default on Lambda and a very common choice in real Terraform and CDK configurations. Any such stack deploys cleanly against Floci and then fails at invoke time — silently on Python/Node, and with a misleading timeout on Go/Rust/`provided.*`.

There is also no way to opt out per function: `UpdateFunctionCode --architectures x86_64` is silently discarded (filed separately as #2839), so a user who diagnoses the problem cannot correct it through the API.

Two directions, and I did not want to guess between them:

1. **Pass the platform through.** Resolve `linux/arm64` when the function says arm64, on both the image pull and the container create. Correct where the host can run it (Docker Desktop, an arm64 host, or a Linux host with `binfmt_misc`/qemu installed) and slow-but-working elsewhere.
2. **Fail loudly.** Keep running the native variant, but detect the mismatch and raise a clear error instead of letting it surface as `Function.TimedOut`.

They are not exclusive — 1 with 2 as the fallback when the platform is unavailable is probably the honest combination — but the choice affects hosts with no emulation installed, where option 1 turns a silent wrong-arch success into a hard failure. That is a behaviour change worth a maintainer's call, so I have raised it rather than sending a patch.

## Reproduction

Interpreted case — shows the mismatch with no failure at all:

```bash

aws $E lambda create-function --function-name arch-probe --runtime nodejs22.x \
    --zip-file fileb://fn.zip --architectures arm64 --query 'Architectures' --output json
# [ "arm64" ]

aws $E lambda invoke --function-name arch-probe --payload '{}' out.json --query 'StatusCode' --output text
# 200

docker image inspect public.ecr.aws/lambda/nodejs:22 --format '{{.Os}}/{{.Architecture}}'
# linux/amd64   <- on an x86_64 host, for a function declared arm64

aws $E lambda delete-function --function-name arch-probe
```

Compiled case — needs any statically linked aarch64 ELF named `bootstrap`:

```bash
zip -q arch-compiled.zip bootstrap        # file bootstrap -> "ELF 64-bit LSB executable, ARM aarch64"

aws $E lambda create-function --function-name arch-compiled-probe --runtime provided.al2023 \
    --role arn:aws:iam::000000000000:role/lambda-role --handler bootstrap \
    --zip-file fileb://arch-compiled.zip --architectures arm64 --timeout 5

aws $E lambda invoke --function-name arch-compiled-probe --payload '{}' out2.json
# {"StatusCode": 200, "FunctionError": "Unhandled", ...}
cat out2.json
# {"errorMessage":"Task timed out after 5 seconds","errorType":"Function.TimedOut"}

docker logs <floci-container> | grep arch-compiled-probe
# ... /var/runtime/bootstrap: cannot execute binary file: Exec format error

aws $E lambda delete-function --function-name arch-compiled-probe
```

Both blocks were executed on an x86_64 Linux host with no `binfmt_misc` arm64 handler registered; the output above is what they printed.

## Environment

- Floci version / image tag: build of upstream `main` at `62dbd9d` ("fix(apigateway): report RestApi as available (#2816)")
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker, with `/var/run/docker.sock` mounted, on x86_64 Linux

nkootstra:
## Summary

  Allow Lambda containers to use the function's declared `arm64` or `x86_64` architecture when `FLOCI_SERVICES_LAMBDA_HONOUR_ARCHITECTURES=true`.

  Apply the selected platform to Docker image pulls, Lambda container creation, and helper containers used for large function packages.

  This behavior is opt-in because foreign-architecture containers require Docker Desktop or host emulation. Enabling it by default could break existing installations. The default therefore remains host-native
  execution.

  Fixes #2840.

  ## Type of change

  - [x] Bug fix (`fix:`)
  - [ ] New feature (`feat:`)
  - [ ] Docs / chore

  ## AWS Compatibility

  Floci previously stored and returned a Lambda function's declared architecture but did not use it when starting Docker containers. An `arm64` function could therefore run with an `amd64` image or fail with a
  misleading timeout.

  When architecture handling is enabled, Floci now selects `linux/arm64` or `linux/amd64` during image pulls and container creation.

  Verified with focused unit tests and a Docker integration test that creates a container with a foreign-platform image. All 66 focused tests passed.

  The full suite completed all 15,322 tests twice. Both runs reported an unrelated RDS Data isolation error. One run also reported an unrelated EC2 test flake. Each affected test passed when run alone. The Lambda
  architecture tests passed in both full runs.

  ## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -1506,8 +1506,14 @@ public class LambdaService implements ResourceProvider {
     }
 
     private static List<String> validateArchitectures(Object value) {
-        if (!(value instanceof List<?> architectures)) {
+        if (value == null) {
             return null;
+        }
+        if (!(value instanceof List<?> architectures)) {
+            throw new AwsException("InvalidParameterValueException",
+                    "1 validation error detected: Value '" + value + "' at 'architectures' "
+                            + "failed to satisfy constraint: Member must be a list",
+                    400);
         }
         if (architectures.isEmpty()) {
             throw new AwsException("InvalidParameterValueException",
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -75,6 +75,7 @@ public class LambdaService implements ResourceProvider {
             "arn:(aws[a-zA-Z-]*)?:iam::\\d{12}:role/?[a-zA-Z_0-9+=,.@\\-_/]+");
     private static final Pattern HANDLER_PATTERN = Pattern.compile("\\S+");
     private static final int MAX_HANDLER_LENGTH = 128;
+    private static final List<String> FUNCTION_ARCHITECTURES = List.of("x86_64", "arm64");
 
     private final LambdaFunctionStore functionStore;
     private final LambdaExecutorService executorService;
@@ -338,6 +339,7 @@ public class LambdaService implements ResourceProvider {
             throw new AwsException("ResourceConflictException",
                     "Function already exist: " + functionName, 409);
         }
+        List<String> architectures = validateArchitectures(request.get("Architectures"));
 
         LambdaFunction fn = new LambdaFunction();
         fn.setAccountId(regionResolver.getAccountId());
@@ -368,10 +370,6 @@ public class LambdaService implements ResourceProvider {
         Map<String, String> tags = (Map<String, String>) request.get("Tags");
         if (tags != null) fn.setTags(tags);
 
-        // Architectures
-        @SuppressWarnings("unchecked")
-        List<String> architectures = request.get("Architectures") instanceof List
-                ? (List<String>) request.get("Architectures") : null;
         if (architectures != null && !architectures.isEmpty()) {
             fn.setArchitectures(new ArrayList<>(architectures));
         }
@@ -555,6 +553,7 @@ public class LambdaService implements ResourceProvider {
     public LambdaFunction updateFunctionCode(String region, String functionName, Map<String, Object> request) {
         LambdaFunction fn = getFunction(region, functionName);
         functionName = fn.getFunctionName();
+        List<String> architectures = validateArchitectures(request.get("Architectures"));
 
         String zipFileBase64 = (String) request.get("ZipFile");
         String imageUri = (String) request.get("ImageUri");
@@ -577,13 +576,8 @@ public class LambdaService implements ResourceProvider {
             }
         }
 
-        if (request.containsKey("Architectures")) {
-            @SuppressWarnings("unchecked")
-            List<String> archs = request.get("Architectures") instanceof List
-                    ? (List<String>) request.get("Architectures") : null;
-            if (archs != null && !archs.isEmpty()) {
-                fn.setArchitectures(new ArrayList<>(archs));
-            }
+        if (architectures != null) {
+            fn.setArchitectures(new ArrayList<>(architectures));
         }
 
         fn.setLastModified(System.currentTimeMillis());
@@ -602,6 +596,7 @@ public class LambdaService implements ResourceProvider {
 
     public LambdaFunction updateFunctionConfiguration(String region, String functionName, Map<String, Object> request) {
         LambdaFunction fn = getFunction(region, functionName);
+        List<String> architectures = validateArchitectures(request.get("Architectures"));
 
         // Validated before any field mutation below, not inline where Layers is applied further
         // down - fn is the live object backing this store entry (InMemoryStorage#get returns the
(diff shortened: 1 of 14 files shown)


pgermosen (commented):
Really careful work on the platform-selection side - the (imageUri, platform) cache key, pinning container creation to the inspected image ID rather than the mutable tag, and the graceful fallback for malformed legacy persisted architectures are all the right calls, and defaulting to linux/amd64 when Architectures is unset correctly matches AWS's real default rather than just following the host.

One gap worth closing before this merges, left inline: this new validation function doesn't actually validate a malformed shape, and now that architecture drives real container behavior that has more teeth than before.

src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
```diff
@@ -1515,6 +1505,35 @@ private static List<LambdaFileSystemConfig> parseFileSystemConfigs(Object value)
         return parsed;
     }
 
+    private static List<String> validateArchitectures(Object value) {
+        if (!(value instanceof List<?> architectures)) {
```
pgermosen: A raw scalar or object value here (e.g. "Architectures": "arm64" with the brackets forgotten, an easy client mistake) falls through this instanceof check straight to `return null`, the same value used for an omitted field entirely. So the request succeeds and the value is silently ignored rather than rejected, and none of the new tests cover this shape (they all use List.of(...)).

This is the exact function this PR wrote to validate Architectures, so it seems like the right place to close it: reject when value is non-null and not a List, the same way the empty/multi-element/unknown-value cases below already do.

nkootstra: > A raw scalar or object value here falls through this instanceof check straight to return null.

Addressed in signed commit 21fe2e71. A non-null, non-list Architectures value now returns InvalidParameterValueException before any function state changes. Regression tests cover create, code update, and configuration update.

nkootstra:
Greptile’s 4/5 reflects a design disagreement, not an unresolved implementation defect. It prefers migrating invalid legacy Architectures records. This PR instead fails clearly when architecture handling is enabled and a stored value cannot be honored.

That tradeoff is intentional. Falling back to the Docker host would recreate #2840 by running the function on a platform that differs from its reported configuration. Silently rewriting the value to x86_64 would modify persisted user data without a defined migration rule. The setting remains opt-in and disabled by default, so existing installations are unaffected until they enable it. Invalid records can then be repaired through UpdateFunctionConfiguration.

pgermosen (approved):
Both open items are genuinely closed now. The malformed-shape gap in validateArchitectures is fixed and tested across create, code update, and configuration update. And the persisted-invalid-architecture case that used to fall back to the daemon's default platform now fails loudly instead with a dedicated test covering the empty, unknown, and multi-value cases, that's the right call given it can now only happen with corrupted or legacy state.

Spent some time on the platform-selection design too, not just the two fixes: the cache re-validates a resolved image is still present locally before reuse, and it cross-checks the actually-pulled image's OS/arch against what was requested rather than trusting the pull. Solid work carrying this through several rounds.

PR #2996, opened 2026-09-03
fix(cloudformation): resolve FunctionResponseTypes through whole-property intrinsics

Mide69:
## Summary

Follow-up to #2964. Greptile flagged there that `FunctionResponseTypes` gets silently dropped when set via a whole-property intrinsic (`Fn::If`, `Fn::Split`, or a `Ref` to a `CommaDelimitedList`). hectorvent merged #2964 anyway, treating this as an acceptable known limitation given the property's narrow value domain (a list of at most one `ReportBatchItemFailures`).

The real gap was in the engine, not the call site: `resolveNode` treated `Fn::If` as a scalar-collapsing intrinsic like `Ref`/`Fn::Sub`, so it stringified a chosen array or object branch instead of preserving it. This is the same engine gap Greptile flagged on #2796 (ScheduleGroup `Tags`), which @hectorvent asked to be raised separately rather than fixed there. `resolveNode` now special-cases `Fn::If` via a new `selectIfBranch` helper: it picks the branch the same way `resolve()` does, then resolves that branch recursively instead of collapsing it, so the result keeps whatever shape the chosen branch actually has. `resolve()` itself is unchanged.

For the `FunctionResponseTypes` call site itself: since #2946 landed on `main` while this was in review, the provisioner's `resolveStringList(props, field, engine)` helper already delegates to `engine.resolveStringList`, which resolves `Fn::If`/`Fn::Split`/a `CommaDelimitedList` `Ref` and drops blank entries. An earlier revision of this PR had that call site go through the engine's own `resolveList` directly instead - redundant once #2946 landed, and worse: `resolveList` keeps blank entries, so a conditional element resolving to `AWS::NoValue` was forwarded as an invalid empty-string response type (flagged independently by both hectorvent and Greptile). Reverted to the plain `resolveStringList(props, "FunctionResponseTypes", engine)` call, kept `resolveList` private (its only external caller was that now-removed hunk), and pointed the engine-level tests at `resolveStringList` instead. Added a test pinning the `AWS::NoValue`-drops-the-entry behavior.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Bug fix: `FunctionResponseTypes` set via `Fn::If`/`Fn::Split`/a `CommaDelimitedList` `Ref` reached `CREATE_COMPLETE` with an empty list instead of the selected value, same class of gap as the Tags issue on #2796.

## Checklist


hectorvent:
Thanks for coming back with a real engine fix: `selectIfBranch` plus `resolveNode` preserving the chosen `Fn::If` branch's shape is the right cut, and the engine tests pin it well.

**(blocking)** The provisioner hunk is no longer needed and it regresses main. Since #2946 merged, `resolveStringList(props, "FunctionResponseTypes", engine)` already goes through `engine.resolveStringList`, which handles `Fn::If`, `Fn::Split` and a `CommaDelimitedList` `Ref` and drops blank entries. `resolveList` keeps blanks, so a list element that resolves to `AWS::NoValue` is forwarded as an empty string (Greptile is right about that one). I checked locally: with main's provisioner file and your engine change, your new integration test passes as is. Would you drop that hunk, keep `resolveList` private, and point the four new tests at `resolveStringList`?

**(follow-up)** #2986 adds an identical `selectIfBranch`; whichever lands second will rebase over the other, so nothing to do here.

Dropping the hunk and refreshing the description to match is what unblocks this from my side.

Mide69:
Fixed in 68b9cbfd - dropped the resolveList hunk, kept it private, switched the provisioner call back to resolveStringList(props, "FunctionResponseTypes", engine), and pointed the four engine tests at resolveStringList. Also added a test pinning the AWS::NoValue-drops-the-entry case you and Greptile both flagged. Updated the PR description to match. Thanks for catching the #2946 overlap - I hadn't rebased far enough forward to see it land.

Noted on #2986 and the selectIfBranch duplication - agreed nothing to do here, whichever merges second rebases over it.

hectorvent (approved):
Thanks, that is exactly the shape I hoped for: the engine fix stays, `resolveList` stays private, and the `AWS::NoValue` case is pinned. Re-read the diff at 82a2acb and ran the whole CloudFormation package on the merge result with main: green. Approving.

hectorvent:
For the record, the red shard 4 was a stale main, not this change. The run merged the head onto main as of 19:57Z on the 6th, before #3158 landed the teardown for the service-managed StackSets test, and a rerun reuses that same merge commit. The leaked organization is what made `OrganizationsFeatureSetIntegrationTest` answer 400 and the Organizations stack time out. Merged onto current main with that shard's class order plus both CloudFormation suites: 202 tests, 0 failures.

PR #2998, opened 2026-09-03
fix(dynamodb): flush persistence once per affected table in batch and transact writes

### LINKED ISSUE #2995: [BUG] dynamodb: BatchWriteItem persists the entire item store per item, not per batch — same quadratic write amplification as #2500
## What happens

`DynamoDBService`'s persistence writes the **entire** item store to `data/dynamodb-items.json` (write to `.tmp`, rename) on every individual item write — not once per `BatchWriteItem` call, but once **per item inside the batch**. The write cost of persisting one item is therefore proportional to *all items ever stored across every table*, and total disk traffic grows quadratically with item count, exactly the same architecture and exact same failure mode as #2500 (`cwlogs`) — just triggered by `PutItem`/`BatchWriteItem` instead of `PutLogEvents`.

This is a real, non-synthetic scenario, not a stress test: a single logical create operation in a downstream service (see Reproduce) legitimately expands into dozens of individual DynamoDB items via `BatchWriteItem` — well within normal, documented DynamoDB usage (25 items/batch is the real API limit) — and that alone is enough to trigger multi-second-per-item persistence costs once the store has accumulated tens of megabytes from ordinary local dev use.

## Why it matters

Observed on Floci 1.7.0, community edition, `FLOCI_STORAGE_MODE=persistent`, with `data/dynamodb-items.json` at ~41-43MB (accumulated from weeks of ordinary local dev, not an adversarial workload):

- A single `BatchWriteItem` request writing 81 items (chunked into 4 batches of ≤25, sent in parallel via `Promise.all` — standard client-side batching) took **31.9s and 39.8s** across two separate runs (the store had grown between runs, consistent with the cost scaling with accumulated size).
- The same call pattern with 3 items took **1.4s**.
- Directly confirmed via inode tracking (`stat` polling the target file's inode at high frequency during a write): the 81-item write produced **84 distinct file-identity (inode) transitions** — i.e., 84 separate full rewrite-and-rename cycles — each one growing the file by ~2,125 bytes, the serialized size of exactly *one* item. Not 4 (one per batch call), not 1 (one for the whole request): one per item, confirmed directly rather than inferred from timing.
- At the point of measurement (~43MB file), that's roughly **84 full-file rewrites (~3.6GB of actual disk I/O) to persist ~179KB of genuinely new data** — an amplification factor around 20,000x, the same order of magnitude as #2500's own measured ~35,000x for cwlogs.
- Real-world consequence: this single write, sitting inside a synchronous Lambda invocation with a 30s local timeout, causes the invocation to time out and the caller to receive an empty-bodied 502 — a completely ordinary application operation (creating one record that happens to normalize into ~80 child rows) fails locally with no indication that store size is the cause.

## Reproduce

```bash
  --table-name write-amp-test --attribute-definitions AttributeName=id,AttributeType=S \
  --key-schema AttributeName=id,KeyType=HASH --billing-mode PAY_PER_REQUEST

# Once dynamodb-items.json has accumulated real size from normal use (tens of MB+),
# batch-write ~80 small items (any AWS SDK, chunked by 25, BatchWriteItem) and time it.
# Compare against the same pattern with ~3 items on a fresh/small store — the gap
# should already be dramatic in the tens-of-MB range, worsening as the store grows.
```
(Happy to attach a runnable Node script that does the write + times it + polls the target file's inode to show the per-item rewrite pattern directly, on request.)

## Expected

Persisting a batch write should cost `O(items in this write)`, not `O(all items ever stored)`. Ideally `BatchWriteItem` should also result in **one** flush for the whole batch, not N flushes for N items within it — a more targeted, minimal fix than a full persistence-architecture change, and worth calling out as its own smaller win independent of whatever gets done for #2500.

## Related

Same `PersistentStorage` write-to-`.tmp`-then-rename pattern as **#2500** (`cwlogs`), same quadratic scaling, same order-of-magnitude amplification — filed separately since the affected code path and the applicable fix (per-stream/per-table files, incremental persistence, debounced flush, or a real flush-per-batch rather than flush-per-item) differ enough that a shared thread would make either fix harder to scope and close independently.

## Workaround (meanwhile)

Avoid single logical requests that fan out into many individual `PutItem`/`BatchWriteItem` calls until this is addressed, or periodically stop Floci and reset `data/dynamodb-items.json` to keep the accumulated store small — same workaround #2500 documents for `cwlogs`.

vuthehuyht:
## Summary

Closes #2995

In `DynamoDbService`, `BatchWriteItem` and `TransactWriteItems` looped over items and directly called `putItem` / `deleteItem` / `updateItem`. In persistent storage mode (`PersistentStorage`), each item mutation immediately serialized the entire item store to a `.tmp` file and atomically renamed it to `dynamodb-items.json`. For a batch of N items, this resulted in N full store rewrites, triggering massive write amplification (O(N) full store rewrites per batch request) and causing timeouts.

This PR:
- Introduces internal overloads in `DynamoDbService` for `putItem`, `deleteItem`, and `updateItem` with a `shouldPersist` boolean flag, keeping public API signatures unchanged.
- Refactors `batchWriteItem` and `transactWriteItems` to defer disk flushes and execute `persistItems(storageKey)` exactly once per affected table at the end of the batch (inside the lock block for transactions).
- Documents the batched persistence behavior and recommends `wal` or `hybrid` storage modes for write-heavy workloads in `docs/services/dynamodb.md`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- Bug fix: `BatchWriteItem` and `TransactWriteItems` triggered multiple full-store disk flushes per request (one per item in the batch), causing high write amplification and latency.
- Behavior now matches expected batch persistence semantics: exactly one disk flush per affected table per request, while preserving per-item event emissions (DynamoDB Streams, Kinesis forwarding) and concurrency locking order.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -748,24 +748,7 @@ public class DynamoDbService implements ResourceProvider {
             }
 
             // Reject any attempt to modify a key attribute
-            String pkName = table.getPartitionKeyName();
-            JsonNode origPk = key.get(pkName);
-            JsonNode newPk = item.get(pkName);
-            if (origPk != null && newPk != null && !origPk.equals(newPk)) {
-                throw new AwsException("ValidationException",
-                        "One or more parameter values were invalid: Cannot update attribute " + pkName
-                        + ". This attribute is part of the key", 400);
-            }
-            String skName = table.getSortKeyName();
-            if (skName != null) {
-                JsonNode origSk = key.get(skName);
-                JsonNode newSk = item.get(skName);
-                if (origSk != null && newSk != null && !origSk.equals(newSk)) {
-                    throw new AwsException("ValidationException",
-                            "One or more parameter values were invalid: Cannot update attribute " + skName
-                            + ". This attribute is part of the key", 400);
-                }
-            }
+            validateKeyNotModified(table, key, item);
 
             // AWS validates index key values against the item the update produces.
             validateIndexKeyTypes(table, item, true);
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
```diff
@@ -1166,31 +1293,84 @@ public void transactWriteItems(List<JsonNode> transactItems, String region,
...
+                        JsonNode exprAttrValues = upd.has("ExpressionAttributeValues") ? upd.get("ExpressionAttributeValues") : null;
+                        UpdateResult res = updateItemInternal(tableName, key, null, updateExpression, exprAttrNames, exprAttrValues,
+                                   "NONE", null, region, "NONE", false, pendingStreamEvents::add);
+                        rollbackList.add(new RollbackEntry(storageKey, itemKey, res.oldItem()));
+                        affectedStorageKeys.add(storageKey);
+                    }
+                    // ConditionCheck-only items are handled in the first pass only
                 }
-                // ConditionCheck-only items are handled in the first pass only
+                for (String storageKey : affectedStorageKeys) {
```
pgermosen, line 1352: In a multi-table transaction, if persistItems succeeds for one table and then throws for a second, the finally block below still rolls back every participant's in-memory state, including the table that already got flushed. Its on-disk file is briefly ahead of memory until the next write to that table overwrites it, so a crash landing in exactly that window could resurrect a value from a transaction that was supposed to be rolled back.

Not asking for a fix here, this is a pre-existing limitation of the per-table-file persistence design (multi-file atomicity across tables was never guaranteed before this PR either), and it's a narrow compound scenario (disk I/O failure immediately followed by a crash). Worth its own follow-up issue if it seems worth closing, rather than folding into this one.

pgermosen (commented):
This grew well beyond the original flush-batching scope, and the growth is real correctness work, not scope creep. Traced the three-pass design in transactWriteItems (conditions, then prevalidation, then apply-with-rollback) and it holds up: the mocked-store test proves a later validation failure results in zero store.put() calls and zero visible mutation from the earlier item, not just a post-hoc rollback. Deferring stream events until transactionSucceeded is set is a genuine bug fix on its own, previously a mid-transaction failure after some items were already applied and streamed would have leaked events for writes that then got rolled back.

One edge case worth a comment rather than a blocker, left inline.

src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
vuthehuyht: Good catch, thank you! That makes complete sense. Given that multi-file cross-table storage atomicity is a broader architectural aspect across the storage layer, tracking it in a dedicated follow-up issue is definitely the right approach.

vuthehuyht: Good catch, thank you! That makes complete sense. Given that multi-file cross-table storage atomicity is a broader architectural aspect across the storage layer, tracking it in a dedicated follow-up issue is definitely the right approach.

pgermosen (approved):
Read batchWriteItem and transactWriteItems directly rather than just trusting the round-1 conclusion. All three original findings genuinely hold up as fixed. The one worth calling out specifically: the fix for repeated same-item updates in a transaction isn't a patch, it rejects that case outright with the same message real AWS uses for its own documented constraint against multiple operations on one item in a transaction, so the scenario is impossible by construction now rather than worked around. The rollback and deferred stream events are correct too, confirmed any mid-apply or persist failure reverts the whole transaction in memory and no stream event escapes for a write that gets rolled back.

The one deferred edge case, cross-table disk/memory skew if persistItems succeeds for one table then fails for a second, is real but a pre-existing limitation of the per-table-file design, not something this PR makes worse, and it's already agreed as a follow-up issue.

Nothing outstanding.

PR #2999, opened 2026-09-03
fix(codebuild): resolve project ARNs in batch lookup

### LINKED ISSUE #2799: codebuild: BatchGetProjects doesn't resolve project ARNs, only bare names
`CodeBuildService.batchGetProjects` looks projects up by exact key match only:

```java
public List<Project> batchGetProjects(String region, List<String> names) {
    Map<String, Project> store = projectsFor(region);
    return names.stream()
            .map(store::get)
            .filter(p -> p != null)
            .collect(Collectors.toList());
}
```

`store` is keyed by the bare project **name** (`store.put(name, project)` in `createProject`). Real AWS's `BatchGetProjects` accepts either project **names or ARNs** interchangeably in the `names` parameter. `terraform-provider-aws`'s read/refresh for `aws_codebuild_project` calls it with the ARN (Terraform stores the ARN as the resource ID), so the lookup misses and the read comes back empty even though the project exists:

```
Error: reading CodeBuild Project (arn:aws:codebuild:us-east-1:000000000000:project/floci-compat-codebuild): empty result
```

Confirmed still present on `main`. Fix direction: resolve each entry in `names` against both the bare name and the ARN (`AwsArnUtils.Arn.of("codebuild", region, account, "project/" + name)`) before the `store::get` lookup.

Blocks the Terraform/OpenTofu compatibility suites from including `aws_codebuild_project` (removed from both in #1834 to unblock the suite; should be re-added once this lands).

Origin: found while getting #1834 green — the buildspec-missing issue abanna originally flagged (`Terraform's own validation... never reaches Floci`) was fixed first, which is what exposed this deeper, previously-unreached bug.

aniketshukla1:
## Summary

Resolve BatchGetProjects identifiers against both stored project names and exact project ARNs. Keep successfully resolved ARNs out of projectsNotFound.

Closes #2799

## Type of change

- [x] Bug fix (fix:)
- [ ] New feature (feat:)
- [ ] Docs / chore

## AWS Compatibility

AWS CodeBuild accepts either names or ARNs in BatchGetProjects. Floci previously keyed its store only by bare name, so Terraform refreshes using an ARN returned an empty result. The JSON 1.1 integration test now looks up the created project by its full ARN and verifies only the genuinely missing identifier is returned in projectsNotFound.

## Checklist

- [x] ./mvnw test -Dtest=CodeBuildIntegrationTest passes locally: 13 tests
- [x] Updated integration test added

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/codebuild/CodeBuildJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/codebuild/CodeBuildJsonHandler.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 @ApplicationScoped
 public class CodeBuildJsonHandler {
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Approving. The fallback lookup and the foundIdentifiers change checking both name and ARN both check out, and the test genuinely exercises the reported scenario - requesting by ARN alongside a real miss, asserting both projects and projectsNotFound come back correct. Clean, minimal fix.

PR #3000, opened 2026-09-03
fix(kinesis): stop dropping shard records under concurrent producers and resharding

guiscaranse:
## Summary

Concurrent `PutItem` → Kinesis CDC forwarding raced on a plain `ArrayList` in `KinesisShard`, silently dropping roughly 1% to 3% of change records under load. The row persisted in DynamoDB, no stream event was emitted, and nothing errored, so the loss was invisible. A second race let lock-free shard readers observe a half-published topology while `SplitShard` was adding children, and a single-shard split gave both children the same shard id.

Changes:

- `KinesisShard` keeps its record list behind a monitor. `getRecords()` returns a snapshot; `addRecord()` and `recordCount()` replace direct list access.
- `KinesisService` holds a per-stream lock across the whole append: sequence allocation, shard selection, append, and `store.put`. That closes the lost-update, sequence-vs-append ordering, and stale-snapshot WAL replay races. `splitShard`, `mergeShards` and `deleteStream` take the same lock and re-resolve the stream inside it, closing a delete-vs-append resurrection race.
- `KinesisStream.shards` becomes a `volatile CopyOnWriteArrayList`, so lock-free readers (`getRecords`, `peekRecords`, `getRecordsForAccount`, iterators, inspection) get snapshots that cannot throw `ConcurrentModificationException` while a split appends under the lock. The list is only ever added to and shard counts are small, so the copy-on-write cost is negligible.
- `KinesisJsonHandler.paginateShards` snapshots with `List.copyOf` before `subList`: a `CopyOnWriteArrayList` sublist is a view rather than a snapshot, and could still throw on `ListShards` during a reshard.
- `splitShard` publishes both children in one atomic `addAll` and derives their ids from the pre-split size, so the ids are distinct and a reader can never observe `{parent, child1}` with a gap in the hash-range tiling. Previously both children were handed the same id.
- Stream metadata writes take the same per-stream monitor and re-resolve the stream inside it: `UpdateStreamMode`, tag add/remove, start/stop encryption, increase/decrease retention, enable/disable enhanced monitoring, and max record size. Previously each did a read-modify-write outside the lock, so an update overlapping `DeleteStream` could re-persist the deleted stream. On the WAL backend it also appended a `PUT` after the `DELETE`, so the resurrection survived replay. Added in review, addressing the P1 raised on this PR.
- `KinesisStreamingForwarder` (DynamoDB CDC) no longer swallows a forwarding failure as a `WARN`-and-drop; it logs at `ERROR` with the stack trace and counts the drop. It deliberately does not rethrow: forwarding runs after the write commits, and the TTL sweep forwards mid-loop before persisting, so rethrowing would turn a committed write into a spurious 5xx or abort the sweep.

Tests: concurrent CDC end-to-end (`DynamoDbConcurrencyIntegrationTest`), service-layer append stress with sequence-order and `AFTER_SEQUENCE_NUMBER` continuation plus a deterministic mutual-exclusion case (`KinesisServiceTest`), WAL persistence and reload (`KinesisWalPersistenceTest`), TTL happy path and failure isolation (`DynamoDbTtlForwardingTest`), a parameterized regression over all ten metadata actions proving an update cannot resurrect a concurrently deleted stream (deterministic, latch-based, asserting the deleter actually blocks), and reader-versus-reshard coverage (`KinesisListShardsConcurrencyTest`): a reader mid-iteration survives a split, a `ListShards` page survives a concurrent split, no half-published topology is observable (`RepeatedTest`), and child ids are unique and tile the parent's hash range.

Known residual, noted in the code: closing the parent shard is not fully atomic with respect to the children `addAll` for lock-free readers. Producers are unaffected.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

This was silent data loss rather than a wire-protocol mismatch: records accepted by CDC forwarding were acknowledged but never returned by `GetRecords`, and AWS does not lose them. `ListShards` could also throw during a concurrent `SplitShard`, and a single-shard split produced two children sharing one shard id, where AWS returns distinct ids that tile the parent's hash range. No request or response shapes change.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/KinesisStreamingForwarder.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/KinesisStreamingForwarder.java
@@ -14,6 +14,7 @@ import org.jboss.logging.Logger;
 import java.time.Instant;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
 
 @ApplicationScoped
 public class KinesisStreamingForwarder {
@@ -22,6 +23,7 @@ public class KinesisStreamingForwarder {
 
     private final KinesisService kinesisService;
     private final ObjectMapper objectMapper;
+    private final AtomicLong forwardFailures = new AtomicLong();
 
     @Inject
     public KinesisStreamingForwarder(KinesisService kinesisService, ObjectMapper objectMapper) {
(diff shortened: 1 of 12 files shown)


pgermosen (approved):
Traced the locking scheme directly rather than just reading the description, since this is exactly the kind of change where it's easy to miss a path. Lock keys are derived identically everywhere, monitors are deliberately never removed and the reasoning for that holds up, and the split-shard id fix correctly avoids both the collision and the half-published-topology cases by computing both child ids up front.

The CopyOnWriteArrayList subList claim checks out against how COW actually implements it, and the regression test proves it deterministically instead of relying on a timing race, same with the mutual-exclusion test, which actually spins on Thread.State.BLOCKED to prove the second producer is parked on the monitor rather than just measuring elapsed time. That's the right way to test this without flakiness.

The one thing left as a known gap, parent-close not being atomic with the children publish for lock-free readers, is honestly disclosed and I confirmed producers never see it. Nothing else outstanding, solid work.

Issue #2521, opened 2026-08-24
[BUG] Step Functions (SFN): unresolvable JSONPath in a payload template yields null instead of failing with States.Runtime


anikolaiev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

A JSONPath payload template field (`"name.$": "$.some.path"`) whose path does not match the input resolves to `null` and the state succeeds. Real AWS and Step Functions Local fail the execution with `States.Runtime`.

State machine:

```json
{
  "StartAt": "Pick",
  "States": {
    "Pick": {
      "Type": "Pass",
      "Parameters": { "missing.$": "$.nope.deep" },
      "End": true
    }
  }
}
```

Started with input `{"other": 1}`.

Floci:

```json
{ "status": "SUCCEEDED", "output": "{\"missing\":null}" }
```

Step Functions Local 2.0.0:

```json
{ "status": "FAILED" }
```

with `executionFailedEventDetails.cause`:

> The JSONPath '$.nope.deep' specified for the field 'missing.$' could not be found in the input '{"other":1}'

### Expected Behavior

The state fails with `States.Runtime` and a cause naming the unresolvable path and field, so `Catch` and `Retry` can act on it.

### How are you starting Floci?

Docker image built from `main` (`257e004`), `docker/Dockerfile`.

### Steps To Reproduce

```bash
aws stepfunctions create-state-machine --endpoint-url http://localhost:4566 \
  --name JsonPathTest --role-arn arn:aws:iam::123456789012:role/DummyRole \
  --definition file://sm.json
# -> arn:aws:states:eu-west-1:000000000000:stateMachine:JsonPathTest

aws stepfunctions start-execution --endpoint-url http://localhost:4566 \
  --state-machine-arn arn:aws:states:eu-west-1:000000000000:stateMachine:JsonPathTest \
  --name jp1 --input '{"other":1}'

aws stepfunctions describe-execution --endpoint-url http://localhost:4566 \
  --execution-arn arn:aws:states:eu-west-1:000000000000:execution:JsonPathTest:jp1
# -> "status": "SUCCEEDED", "output": "{\"missing\":null}"
```

### Anything else?

This affects JSONPath state machines only; it is unrelated to JSONata's documented "unresolved reference yields nothing" semantics.

[...]

PR #3002, opened 2026-09-03
feat(stepfunctions): close ASL gaps in intrinsics, Choice, StartExecution and ResultPath

guiscaranse:
## Summary

Five independent Amazon States Language conformance gaps, one per commit. Each is a case where Floci accepted a definition or an execution that AWS handles differently, and four of the five failed silently rather than erroring.

- **Intrinsics** (`feat`): `States.Base64Encode`, `Base64Decode`, `StringSplit`, `ArrayGetItem` and `Hash` aborted execution with `States.Runtime "Unsupported intrinsic function"`. All five are implemented in JSONPath mode following the existing `States.JsonMerge` pattern: RFC 4648 Basic codec over UTF-8 with a 10,000 code-point cap; `StringSplit` treating the delimiter as a set of code points and omitting empty segments; zero-based `ArrayGetItem` rejecting non-integer, negative and out-of-range indexes; `Hash` over an explicit MD5/SHA-1/SHA-256/SHA-384/SHA-512 allow-list. Argument and data failures raise the catchable `States.IntrinsicFailure`, leaving `States.Runtime` for a malformed or unrecognised call.

- **Choice comparators**: only 18 of the 39 documented data-test comparators existed. Every unrecognised one, including `NumericGreaterThanEqualsPath`, the `String*Path` ordering variants, the whole `Timestamp*` family and `IsTimestamp`, fell off the end of the comparison chain and evaluated silently false, so the Choice took its `Default` branch: a workflow routing on `>=` took the wrong path with no error, log line or history annotation. New `ChoiceOperators` is the single source of truth for all 39 operators, their evaluation and their rule validation, shared by `AslExecutor.evaluateCondition` and `StepFunctionsService.validateState` so runtime and create/update cannot disagree. Comparisons are now type-strict: a mismatched JSON type compares false rather than coercing, where previously a missing path compared equal to numeric `0` and numeric `5` matched string `"5"`. Numerics compare via `BigDecimal`, timestamps via an RFC 3339 lexical gate then instant comparison, and `StringMatches` honours `\*` and `\\` while rejecting a dangling escape. Invalid rules are rejected at `CreateStateMachine`/`UpdateStateMachine`/`ValidateStateMachineDefinition`, with a `States.Runtime` backstop at evaluation time for already-persisted definitions.

- **StartExecution idempotency**: any name collision was rejected with `ExecutionAlreadyExists`, and the check-and-create ran unsynchronized, so two concurrent same-name starts could both create and launch an execution. AWS treats a STANDARD start with the same name *and* the same input, while the original is still `RUNNING`, as an idempotent success that returns the original. The check-and-create is now serialized, and a matching input returns the existing execution while a different input or a closed execution still conflicts. EXPRESS keeps the conflict behaviour, since AWS gives it no idempotency guarantee.

- **Nested startExecution**: the optimized `states:startExecution` integration ignored the Task's `Name` and re-serialized the resolved `Input` unconditionally, so an `Input` built with `States.JsonToString(<object>)` was double-encoded and the child received a JSON string literal instead of the object. New `NestedExecutionInput` encodes the child input by provenance, sharing the intrinsic function-name parse with `AslExecutor.evaluateIntrinsic` so the two cannot drift. `Name`/`Name.$` is honoured, and a supplied `Name` that does not resolve to a non-empty string fails with `States.Runtime` rather than silently generating one.

- **ResultPath**: when a state's `ResultPath` addressed an object member (`$.field`) but the state input was not a JSON object, Floci silently discarded the input and returned the bare result, where AWS fails with `States.ResultPathMatchFailure`. The merge moves into a unit-testable `ResultPathMerge`, and a failure raised while applying a matched Catcher's own `ResultPath` now propagates under its real error name instead of being relabelled `States.Runtime`.

Tests: `AslExecutorIntrinsicFunctionsTest` (AWS doc examples, edge cases, execution-level `Catch`, Parameters-path wiring, and a nine-existing-intrinsic regression), `ChoiceOperatorsTest` and `AslExecutorChoiceRuntimeTest`, `StepFunctionsServiceStartExecutionIdempotencyTest` and `...RaceTest`, `NestedExecutionInputTest` and `AslExecutorNestedStartExecutionTest`, `ResultPathMergeTest` and `AslExecutorResultPathTest`.

Behaviour changes, all toward AWS: definitions relying on loose Choice coercion or a silently-false unknown comparator now fail loudly; a `$.field` `ResultPath` over a scalar or array input now fails instead of dropping the input; a `States.JsonToString(<object>)` nested `Input` now reaches the child as an object.

Known limitations: leap-second and >9-fractional-digit timestamps are treated as invalid; closed execution names are not reusable after AWS's 90-day uniqueness window (Floci conflicts for the process lifetime); the StartExecution lock is one global monitor rather than per-name; `$[...]`-style ResultPaths are still not applied, and a `$.a.b` path whose `$.a` is a scalar is overwritten rather than failing; JSONata `Arguments` and colliding `Input`/`Input.$` templates are out of scope.

Refs #2522: this implements 5 of the 9 intrinsics listed there (`ArrayGetItem`, `Base64Encode`, `Base64Decode`, `Hash`, `StringSplit`). `ArrayPartition`, `ArrayRange`, `ArrayUnique` and `MathRandom` remain, so the issue stays open.

Related but not addressed here: #2521 (an unresolvable JSONPath in a *payload template* yields null instead of failing) is the same fail-loud theme on a different code path, and #2927 (`States.Format` taking its template argument literally) is untouched, though every argument of the five new intrinsics does resolve through `resolveIntrinsicArg`.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the ASL specification rather than a live account: the Choice-state spec for all `*Path` and `Timestamp*` comparators and for rejecting unknown comparators at create time (AWS returns `InvalidDefinition` / `SCHEMA_VALIDATION_FAILED`); the intrinsic-function reference for the five added functions, including the AWS doc examples; the `StartExecution` idempotency contract; the payload-template and `States.JsonToString` semantics for a nested execution's input and name; and the `ResultPath` rule that a non-applicable path is `States.ResultPathMatchFailure` rather than a silent discard. No request or response shapes change.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -2816,19 +2816,13 @@ public class AslExecutor {
             return result;
         }
         String resultPath = stateDef.get("ResultPath").asText();
-        if (resultPath == null || resultPath.equals("null")) {
-            return input;
+        try {
+            return ResultPathMerge.merge(input, resultPath, result, objectMapper);
+        } catch (ResultPathMerge.ResultPathMatchException e) {
+            // AWS fails a non-applicable ResultPath with States.ResultPathMatchFailure rather than
+            // silently discarding the state input.
+            throw new FailStateException("States.ResultPathMatchFailure", e.getMessage());
         }
-        if ("$".equals(resultPath)) {
-            return result;
-        }
-        // Merge result into input at the given path
-        if (!input.isObject()) {
-            return result;
-        }
-        ObjectNode merged = input.deepCopy();
-        setPath(merged, resultPath, result);
-        return merged;
     }
 
     private JsonNode applyOutputPath(JsonNode stateDef, JsonNode input, JsonNode output) {
@@ -3770,28 +3764,6 @@ public class AslExecutor {
             return s.substring(1, s.length() - 1);
         }
         return s;
-    }
-
-    private void setPath(ObjectNode root, String path, JsonNode value) {
-        if (!path.startsWith("$.") && !"$".equals(path)) {
-            return;
-        }
-        if ("$".equals(path)) {
-            return;
-        }
-        String[] parts = path.substring(2).split("\\.");
-        ObjectNode current = root;
-        for (int i = 0; i < parts.length - 1; i++) {
-            JsonNode next = current.path(parts[i]);
-            if (!next.isObject()) {
-                ObjectNode newNode = objectMapper.createObjectNode();
-                current.set(parts[i], newNode);
-                current = newNode;
-            } else {
-                current = (ObjectNode) next;
-            }
-        }
-        current.set(parts[parts.length - 1], value);
     }
 
     // ──────────────────────────── History helpers ────────────────────────────
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -67,12 +67,16 @@ import org.jboss.logging.Logger;
 
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.HexFormat;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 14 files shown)


pgermosen (approved):
Went deep on the new comparator engine, the five intrinsics, and the StartExecution locking rather than skimming given the size. The type-strictness rules and the StringMatches regex compilation check out against the ASL spec, and I checked all five intrinsics against AWS's live docs directly. Small bonus along the way: verified the States.Hash SHA-1 worked example independently and AWS's own doc example is actually missing a character, your implementation produces the correct 40-char hash regardless, so nothing to fix there, just noting it.

The StartExecution idempotency contract matches AWS's documented behavior exactly, STANDARD plus matching input plus still-RUNNING returns the original, EXPRESS correctly gets no such guarantee. And the ResultPath/NestedExecutionInput extractions are scoped tightly to the actual bugs, the other known limitations are left alone rather than quietly expanded.

Nothing outstanding.

PR #3004, opened 2026-09-03
fix(kms): validate KeyUsage against KeySpec on CreateKey (#2787)

### LINKED ISSUE #2787: [BUG] KMS: CreateKey with ECC_NIST_EDWARDS25519 returns a NIST P-521 key, so Sign fails with InvalidKeyException
## Summary

`CreateKey` with `KeySpec=ECC_NIST_EDWARDS25519` returns HTTP 200 and a **NIST P-521 (secp521r1) keypair**, not an Edwards25519 key. `GetPublicKey` then returns a 158-byte P-521 SubjectPublicKeyInfo, and any subsequent `Sign` fails with `InvalidKeyException: Unsupported key type`.

The consequence is not "Ed25519 is unsupported". It is that key creation **succeeds with the wrong algorithm**, so a caller that checks the status code believes it holds an Ed25519 key and only discovers otherwise one layer later, at signing, where it presents as a signing bug rather than a key-generation bug.

## Root cause

`KmsKeySpec.java:20-21` (at `b76eadc98eb1`): the Edwards25519 row carries the curve name of the line above it.

```java
ECC_NIST_P521(KeyType.ECC, Algorithm.ECDSA_SHA_512, "secp521r1"),
ECC_NIST_EDWARDS25519(KeyType.ECC, List.of(Algorithm.ED25519_SHA_512, Algorithm.ED25519_PH_SHA_512),"secp521r1"),
```

Ed25519 is not a secp curve, and `KmsService.generateKeyMaterial` passes that string straight into the generic EC generator (`KmsService.java:198-212`):

```java
case ECC -> {
    String curveName = spec.curveName();
    ...
    generator.initialize(new ECGenParameterSpec(curveName));
```

So the `ECC` arm produces P-521 and the `KeyType.ECC` classification hides it.

## Reproduction

I could not run the container on this host (no container runtime available), so I re-executed the code path above in isolation on **JDK 21.0.12**, using the same JCA calls with the same arguments the service passes. Output:

```
=== 1. floci's ECC arm with curveName from ECC_NIST_EDWARDS25519 ===
  private key algorithm : EC
  public  key algorithm : EC
  public  key class     : sun.security.ec.ECPublicKeyImpl
  field size (bits)     : 521
  SPKI length (bytes)   : 158
  SPKI b64 (first 40)   : MIGbMBAGByqGSM49AgEGBSuBBAAjA4GGAAQB/SNt...
  contains Ed25519 OID (2b 65 70)? false

=== 2. floci's sign path over that key: Signature.getInstance("Ed25519") ===
  java.security.InvalidKeyException: Unsupported key type

=== 3. Does the EC generator accept "Ed25519" as an ECGenParameterSpec? ===
  java.security.InvalidAlgorithmParameterException: Unknown curve name: Ed25519
  => the curve string alone is NOT the fix; the arm must dispatch differently

=== 4. The correct generator for Edwards25519 ===
  public key algorithm  : EdDSA
  SPKI length (bytes)   : 44
  contains Ed25519 OID (2b 65 70)? true
  signature length      : 64 bytes
```

The 158-byte SPKI carries `1.2.840.10045.2.1` + `1.3.132.0.35` (id-ecPublicKey + secp521r1). A real Ed25519 SPKI is 44 bytes and carries `1.3.101.112`.

**Step 3 is the part worth reading before fixing:** correcting the curve string is not sufficient, because the JDK's EC generator has no `Ed25519` curve. The `ECC` arm has to dispatch Edwards keys to a different generator.

## Why the existing test does not catch it

`KmsIntegrationTest.java:1079-1082` asserts only the response status:

```
"ECC_NIST_EDWARDS25519, ENCRYPT_DECRYPT, 200",
"ECC_NIST_EDWARDS25519, SIGN_VERIFY, 200",
"ECC_NIST_EDWARDS25519, GENERATE_VERIFY_MAC, 400",
"ECC_NIST_EDWARDS25519, KEY_AGREEMENT, 200",
```

`SIGN_VERIFY -> 200` passes today, on a P-521 key. A single assertion on the generated key's algorithm (or on `GetPublicKey` returning a 44-byte SPKI) would have failed.

## Second defect visible in that same table

Two of those four rows also diverge from AWS. The `CreateKey` API reference lists `ECC_NIST_EDWARDS25519 (ed25519)` under **"signing and verification only"**, and for KeyUsage it directs NIST-standard ECC pairs to `SIGN_VERIFY` or `KEY_AGREEMENT` while Edwards25519 is signing/verification only. floci returns 200 for `ENCRYPT_DECRYPT` and for `KEY_AGREEMENT` on this spec. That follows from the same root cause: because the key is classified `KeyType.ECC`, it inherits generic NIST-ECC usage validation.

## Suggested fix

Three changes, because a fix limited to generation still breaks on reload:

1. Give Edwards25519 its own key type (or a null `curveName` plus an explicit branch) and generate with `KeyPairGenerator.getInstance("Ed25519")`.
2. `buildKeyFactory` (`KmsService.java:1228-1230`) maps every non-RSA spec to `KeyFactory.getInstance("EC")`, so `loadPrivateKey`/`loadPublicKey` need the same dispatch, otherwise a correctly generated Ed25519 key fails to load back.
3. Restrict KeyUsage for this spec to `SIGN_VERIFY`.

Worth noting for whoever picks this up: the `MessageType=DIGEST` branch at `KmsService.java:933-936` resolves to `"NONEwith" + (RSA ? "RSA" : "ECDSA")`, which would yield `NONEwithECDSA` for an Edwards key. AWS documents `ED25519_SHA_512` as requiring `MessageType:RAW` and `ED25519_PH_SHA_512` as requiring `MessageType:DIGEST`, and says the two are not interchangeable. That looks like the same root shape as the open #2570 (`Sign` ignoring the requested signing algorithm when `MessageType=DIGEST`), so the two may be worth fixing together.

## Context

Both `ECC_NIST_EDWARDS25519` and the `ML_DSA_*` specs entered in #1430 (`refactor(kms): use enums for type-safety`), which did not previously exist in the service: at its parent commit the only curve mapping was `case "ECC_NIST_P521" -> "secp521r1"`. The enum was widened to AWS's real spec list and the crypto arms did not follow. That refactor is a clear improvement and it also added the parameterised coverage above, which is what made this easy to pin down.

Not reporting the `ML_DSA_*` specs as a bug: the same test table marks them `// Not implemented`, and they fail closed with `InvalidCustomerMasterKeySpecException`. A declared-but-rejected spec is a documented gap, which reads to me as the same class as #497 (`HMAC_512`) and is presumably already on a roadmap.

## Environment

- floci `main` @ `b76eadc98eb1`, source-read; `KmsKeySpec.java` unchanged since `c2cd31216be5`
- JDK 21.0.12 for the isolated reproduction
- floci itself was **not** executed; every claim above is either source-read at that commit or observed in the isolated JCA run

---

Cowork:Alice RED (Reproducible Evidence & Disclosure) · Aunova

Method: manual source review at a pinned commit plus a local reproduction of the affected code path. No scanner was used for this report, and no runtime claim is made about the floci server itself, which was not run. Severity framing is ours. Not a certification and not an assurance engagement.

vuthehuyht:
## Summary

Validates `KeyUsage` against `KeySpec` on KMS `CreateKey`, rejecting incompatible combinations with HTTP 400 (`ValidationException`).

Fixes the second defect in #2787 (the primary bug around Ed25519 key generation was previously resolved in #2930).

Closes #2787

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

### Incorrect Behavior
Previously, `KmsService.validateKeyUsageForSpec` only enforced the HMAC <-> `GENERATE_VERIFY_MAC` pairing; all other `KeySpec` and `KeyUsage` combinations returned HTTP 200. Real AWS KMS rejects incompatible combinations at `CreateKey`.

### Target AWS Matrix
Derived from the [AWS KMS CreateKey API reference](https://docs.aws.amazon.com/kms/latest/APIReference/API_CreateKey.html):

| KeySpec | Valid KeyUsage |
| --- | --- |
| `SYMMETRIC_DEFAULT` | `ENCRYPT_DECRYPT` only |
| `HMAC_224` / `HMAC_256` / `HMAC_384` / `HMAC_512` | `GENERATE_VERIFY_MAC` only |
| `RSA_2048` / `RSA_3072` / `RSA_4096` | `ENCRYPT_DECRYPT` or `SIGN_VERIFY` |
| `ECC_NIST_P256` / `ECC_NIST_P384` / `ECC_NIST_P521` | `SIGN_VERIFY` or `KEY_AGREEMENT` |
| `ECC_NIST_EDWARDS25519` | `SIGN_VERIFY` only |
| `ECC_SECG_P256K1` | `SIGN_VERIFY` only |
| `ML_DSA_44` / `ML_DSA_65` / `ML_DSA_87` | `SIGN_VERIFY` only |
| `SM2` | `ENCRYPT_DECRYPT` / `SIGN_VERIFY` / `KEY_AGREEMENT` |

### Key Points and Nuances
- **Validation logic:** Added `allowedKeyUsages()` to `KmsKeySpec` enum returning an `EnumSet` per key type (with `ECC_SECG_P256K1` restricted to `SIGN_VERIFY`). `KmsService.validateKeyUsageForSpec` delegates to this method.
- **NIST ECC with KEY_AGREEMENT:** `ECC_NIST_P256/P384/P521` + `KEY_AGREEMENT` deliberately keeps returning HTTP 200 to match real AWS `CreateKey` behavior (even though `DeriveSharedSecret` is not yet implemented in Floci).
- **Provisional error type:** Error `__type` is `ValidationException`, chosen for consistency with existing in-method validation. The exact exception type/message that live AWS KMS returns for this `CreateKey` path is not verified against a live AWS account. Happy to adjust if maintainers confirm via live testing.
- **SM2 / ML_DSA:** Unimplemented key specs remain rejected upfront before reaching `validateKeyUsageForSpec`.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-python/tests/test_kms.py
+++ b/compatibility-tests/sdk-test-python/tests/test_kms.py
@@ -40,6 +40,15 @@ class TestKMSKey:
         response = kms_client.describe_key(KeyId=key_id)
         assert response["KeyMetadata"]["KeyState"] == "PendingDeletion"
 
+    def test_create_key_rejects_incompatible_key_usage(self, kms_client):
+        """CreateKey rejects a KeyUsage that AWS KMS does not allow for the KeySpec."""
+        with pytest.raises(ClientError) as excinfo:
+            kms_client.create_key(
+                KeySpec="ECC_NIST_EDWARDS25519",
+                KeyUsage="ENCRYPT_DECRYPT",
+            )
+        assert excinfo.value.response["Error"]["Code"] == "ValidationException"
+
 
 class TestKMSGrants:
     """Test KMS grant operations."""
--- a/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/KmsService.java
@@ -248,16 +248,9 @@ public class KmsService implements ResourceProvider {
     }
 
     private static void validateKeyUsageForSpec(KmsKeyUsage keyUsage, KmsKeySpec spec) {
-        if (isHmac(spec) && KmsKeyUsage.GENERATE_VERIFY_MAC != keyUsage) {
+        if (!spec.allowedKeyUsages().contains(keyUsage)) {
             throw new AwsException("ValidationException",
-                    "KeyUsage " + keyUsage + " is not compatible with KeySpec " + spec
-                            + ". HMAC key specs require KeyUsage GENERATE_VERIFY_MAC.",
-                    400);
-        }
-        if (KmsKeyUsage.GENERATE_VERIFY_MAC == keyUsage && !isHmac(spec)) {
-            throw new AwsException("ValidationException",
-                    "KeyUsage GENERATE_VERIFY_MAC requires an HMAC KeySpec (HMAC_224, HMAC_256, HMAC_384, or HMAC_512).",
-                    400);
+                    "KeyUsage " + keyUsage + " is not compatible with KeySpec " + spec + ".", 400);
         }
     }
 
--- a/src/main/java/io/github/hectorvent/floci/services/kms/model/KmsKeySpec.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kms/model/KmsKeySpec.java
@@ -6,6 +6,7 @@ import io.quarkus.runtime.annotations.RegisterForReflection;
 import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
 
 @RegisterForReflection
 public enum KmsKeySpec {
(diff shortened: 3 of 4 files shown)


pgermosen (approved):
Approving. Pulled the live CreateKey API reference and checked every row of the matrix against it word for word - all of it holds, including the two easy-to-get-wrong ones: NIST ECC genuinely allows KEY_AGREEMENT alongside SIGN_VERIFY, and Edwards25519 is documented as its own signing-only bullet separate from the NIST-standard group. allowedKeyUsages() implements that exactly.

On the error type you flagged as unverified: ValidationException/400 isn't a new guess, it's what the existing HMAC-vs-GENERATE_VERIFY_MAC check in this same method already used before this PR, and the refactor correctly generalizes that into one check without changing its behavior. Good, well-sourced fix.

PR #3009, opened 2026-09-03
fix(ecr): enforce immutable image tags

simonepri:
## Summary

Route ECR Distribution traffic through Floci's port `4566` listener and translate requests to isolated backing storage. Reject replacement manifest writes for immutable tags, bind the backing registry to loopback, direct EKS mirrors through the guarded listener, and preserve repositories created through the former direct endpoint.

Equivalent support for GCP Artifact Registry and Azure ACR will require further work first, as their respective data-plane surfaces are not yet implemented in Floci.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

- AWS ECR tag mutability: https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-tag-mutability.html
- Incorrect behavior: `PutImageTagMutability` recorded `IMMUTABLE`, but Docker manifest writes went directly to the backing distribution registry, allowing tags to be overwritten. Proxied manifest writes now reject replacements with `ImageTagAlreadyExistsException`.

## Checklist


KenkoGeek (commented):
I reviewed the current data-plane path, the shared immutability helper, account/Region storage isolation, the concurrency guard, and the current AWS ECR image-tag-mutability contract. The core behavior is directionally correct: an existing tag in an `IMMUTABLE` repository must reject a replacement with `ImageTagAlreadyExistsException`, and centralizing that check across PutImage and OCI manifest writes closes the bypass this PR targets.

I do not have a separate AWS-semantic blocker in the `IMMUTABLE` behavior from this pass, but the current head is not ready for approval for two concrete reasons.

**HIGH - the new ECR data-plane regression itself fails in CI.**

`EcrRegistryDataPlaneTest.immutableTagRejectsReplacementBeforeItReachesTheRegistry` errors with `TimeoutException` at the first `putManifest(...)` call. The job log also shows Vert.x reporting `Connection was closed` immediately before the timeout. This is the test that exercises the new proxy path end to end, so this cannot be treated as unrelated CI noise. The implementation needs to make the first proxied manifest PUT complete reliably on the failing runner/architecture, and the same compatibility job needs to pass before this path is considered validated.

I would debug the request/response lifecycle around `proxy()`, especially ownership of the incoming request stream and when the upstream response handler is attached, then keep this exact regression as the proof that both the initial write and the immutable rejection terminate correctly.

**Repository state**

- GitHub currently reports this PR as `CONFLICTING` / `DIRTY` against `main`, so the final diff must be rebased and reviewed again after conflicts are resolved.
- The current CI summary is red because of the ECR data-plane test above.
- The other policy/documentation/native checks shown for this head are green.

One scope note: current ECR also has `IMMUTABLE_WITH_EXCLUSION` and `MUTABLE_WITH_EXCLUSION`, but I am not treating those as a blocker for this PR because the stated bug and implementation are specifically about the repository's existing `IMMUTABLE` mode. That broader mutability-mode gap should be handled separately unless this PR expands its claimed support.

AWS references:
- https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-tag-mutability.html
- https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_PutImageTagMutability.html

simonepri:
Thanks for the thorough review and spot-on diagnosis, @KenkoGeek!

Rebased onto main and resolved the conflicts. The CI failure was indeed Vert.x consuming the body before the proxy piped it (Request has already been read). I've paused the incoming request before the async immutability check and switched to attaching the response handler before streaming via clientReq.send(req) + req.resume().

All the ECR tests (including immutableTagRejectsReplacementBeforeItReachesTheRegistry) are passing locally now.

KenkoGeek (commented):
Thanks for addressing the proxy lifecycle issue. The branch is rebased and mergeable now, and the new head no longer reproduces the ECR data-plane timeout from my previous review: all four JVM shards, native builds, and compatibility jobs shown for this head are green. I do not see a new ECR semantic blocker in the IMMUTABLE path from this pass.

I am still not approving because the PR remains Draft and the required `PR body keeps the required template sections` check is red. Once the description is brought back into the repository template and the PR is marked ready, I will recheck the final head/matrix. The broader `IMMUTABLE_WITH_EXCLUSION` / `MUTABLE_WITH_EXCLUSION` modes remain separate scope as noted before.

simonepri:
@KenkoGeek done!

KenkoGeek (approved):
Thanks for closing out the remaining contribution-state items. The PR is ready for review now, the current PR-template check is green, the branch is clean and mergeable, and the full CI/native/compatibility matrix is green. The ECR data-plane proxy and IMMUTABLE enforcement were already revalidated on this head after the stream-lifecycle fix, and I do not have a remaining blocker in this scope. Approved.

pgermosen:
The immutable-tag rejection doesn't match what real ECR actually sends back over the Docker Registry API. imageTagAlreadyExists() responds with code DENIED and message ImageTagAlreadyExistsException, but a real push against an actual immutable ECR repo returns code TAG_INVALID with a descriptive message like "The image tag X already exists in the Y repository and cannot be overwritten because the tag is immutable." ImageTagAlreadyExistsException is the name AWS uses on the separate management API (the ecr:* JSON calls), not something that shows up in the Distribution wire response.

This matters because tooling that branches on the registry's error code to detect this exact case exists in the wild, and it's specifically looking for TAG_INVALID. The current test only checks the 400 status, never the body, so nothing catches the mismatch.

Could you switch the code/message to match what ECR actually returns, and add a body assertion to the existing test so it doesn't regress?

simonepri:
@pgermosen Good catch! Switched the error response to `TAG_INVALID` with ECR's wire message and added body assertions to the test.

pgermosen (approved):
Confirmed the TAG_INVALID fix directly in the code rather than taking it on faith, error code and message both match what real ECR sends, and there's now a body assertion pinning it. Traced the in-flight tag guard through every exit path too, it releases correctly everywhere, and the thing that actually makes it safe is that the backing registry is bound to loopback only, so there's no way to reach it except through this proxy. Repository names flowing through the new parsing are either regex-constrained or just get forwarded as an HTTP path to a registry that validates them itself, nothing reaches a shell command from request input here. Approving.

simonepri:
@pgermosen Thank you! It's quite refreshing to see an OSS project review contributions so quickly!

PR #3011, opened 2026-09-03
feat(iam): support DynamoDB fine-grained access control (#2926)

### LINKED ISSUE #2926: [FEAT] DynamoDB fine-grained access control: dynamodb:LeadingKeys/Attributes condition keys and ForAllValues set operators
## Service

IAM (policy enforcement), affecting DynamoDB

## API Action / Feature

Support for DynamoDB's fine-grained access control condition keys, `dynamodb:LeadingKeys`, `dynamodb:Attributes` and `dynamodb:Select`, together with the `ForAllValues:` and `ForAnyValue:` set operators those policies are written with.

## AWS Documentation

- Using IAM policy conditions for fine-grained access control: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/specifying-conditions.html
- Condition keys for Amazon DynamoDB: https://docs.aws.amazon.com/service-authorization/latest/reference/list_dynamodb.html
- Item-level access based on a Cognito identity: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_dynamodb_items.html

## Why is this needed?

Item-level scoping is the standard way to make one DynamoDB table multi-tenant: a session role grants access only to items whose partition key begins with the caller's own identity, usually through a principal tag.

```json
{
  "Effect": "Allow",
  "Action": ["dynamodb:GetItem", "dynamodb:Query", "dynamodb:PutItem"],
  "Resource": "arn:aws:dynamodb:...:table/AppTable",
  "Condition": {
    "ForAllValues:StringLike": {
      "dynamodb:LeadingKeys": ["USER_${aws:PrincipalTag/userId}*"]
    }
  }
}
```

Floci can run the application that uses such a role, but not the boundary itself, so the isolation is only ever exercised against real AWS.

## Current behavior

Three pieces are missing, and they compound.

**Condition keys are never populated.** `IamConditionContextResolver` supplies `s3:prefix`, `s3:delimiter` and `s3:ListBucket` only, with no DynamoDB handling.

**The set operators are not implemented.** `ForAllValues:` and `ForAnyValue:` appear nowhere in the IAM package. `IamPolicyEvaluator.evaluateConditionBlock` strips only an `IfExists` suffix, so `ForAllValues:StringLike` reaches the operator switch as a literal unknown operator.

**The condition context is single-valued.** It is a `Map<String, String>`, while `dynamodb:LeadingKeys` and `dynamodb:Attributes` are multi-valued by definition, so representing them needs a shape change rather than another case.

The combined effect is fail-closed rather than fail-open, which is the safer direction but still wrong: a missing key hits

```java
return false; // key missing, no IfExists -> fail entire block
```

so the statement never matches and every request the policy was meant to allow is denied.

## Measurements

Against `floci/floci:2.0.1` with `FLOCI_SERVICES_IAM_ENFORCEMENT_ENABLED=true`, a role granting `dynamodb:GetItem` with `ForAllValues:StringLike` on `dynamodb:LeadingKeys` of `USER_alice*`:

```
GetItem USER_alice  (policy intends: allow)  -> AccessDeniedException
GetItem USER_bob    (policy intends: deny)   -> AccessDeniedException
```

Both are denied, so the policy cannot distinguish in-scope from out-of-scope items.

That result on its own does not isolate the condition, because #2925 means a policy naming a table ARN is denied regardless of any condition. Controls on the same session:

```
Action * / Resource *                  -> ALLOWED
Action * / real table ARN              -> AccessDeniedException   (this is #2925)
```

So the condition keys cannot even be reached today. #2925 has to land before this becomes observable, which is why they are filed separately.

## Proposed shape

- Populate `dynamodb:LeadingKeys` from the partition key values a request names, and `dynamodb:Attributes` from the attribute names it touches, in the same place the resource ARN is derived. Both need the parsed JSON body, which is the same prerequisite #2925 has.
- Widen the condition context to multi-valued so a set-typed key can be represented.
- Implement `ForAllValues:` and `ForAnyValue:` over that, with the empty-set semantics AWS documents. `ForAllValues:` evaluating true on an empty set is the behaviour policies guard against with a `Null` check, so it needs to match rather than be simplified away.

## What I could not measure

I did not verify AWS's own behaviour for these policies against a live account: doing so needs an IAM role and a scoped session in an account I was not going to mutate for a probe. The expected behaviour above is from the documentation linked at the top, not from a call I made. The Floci-side measurements are all executed.

- [ ] Yes
- [ ] No

vuthehuyht:
## Summary

Implements DynamoDB item-level fine-grained access control (FGAC) in the emulator, enabling IAM policies scoped with condition keys such as `dynamodb:LeadingKeys`, `dynamodb:Attributes`, and `dynamodb:Select` (via set operators like `ForAllValues:StringLike`) to accurately allow or deny requests on specific partition keys and attributes.

Closes #2926

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

### Gaps Closed

1. **Multi-valued condition context**: `IamPolicyEvaluator` evaluated condition contexts as `Map<String, String>`, which prevented multi-valued/set-typed context keys from being represented. The evaluation pipeline across `evaluate`, `simulateCustomPolicy`, `simulatePrincipalPolicy`, and internal helpers now carries `Map<String, List<String>>`.
2. **Set operators (`ForAllValues:` / `ForAnyValue:`)**: Added case-sensitive prefix parsing for `ForAllValues:` and `ForAnyValue:` set quantifiers ahead of `IfExists` stripping. Implemented AWS empty-set evaluation semantics: `ForAllValues:` over an empty set matches vacuously (`true`), whereas `ForAnyValue:` over an empty set does not match (`false`).
3. **SimulatePrincipalPolicy multi-value extraction**: `IamQueryHandler.extractContextEntries` previously read only `ContextKeyValues.member.1` and dropped subsequent members. It now reads all indexed members.
4. **DynamoDB condition keys extractor**: Added `DynamoDbConditionKeys` and `DynamoDbKeyConditionParser` to extract partition keys, referenced attributes, and `Select` parameter across `GetItem`, `PutItem`, `UpdateItem`, `DeleteItem`, `Query`, `BatchGetItem`, and `BatchWriteItem`.
5. **IAM enforcement filter wiring**: Wired `IamConditionContextResolver` with a `dynamodb` branch using the buffered JSON body and lazily injected `DynamoDbService` to resolve table schemas without hard cyclic dependencies.

### Security & Fail-Closed Behavior
- Whenever a leading key or table schema cannot be determined (e.g., unknown table, a `Key` that omits the partition key attribute, or an unparseable expression), the condition key is omitted rather than guessed.
- Under an IAM policy scoped purely through `dynamodb:LeadingKeys`, a malformed request (such as a `GetItem` omitting the partition key) results in `AccessDeniedException` (HTTP 403) instead of DynamoDB `ValidationException` (HTTP 400). Failing closed is the intended behavior for security policy enforcement.

### Deferred Scope
Follow-up areas (`Scan` leading keys, transactions across multiple tables, and PartiQL predicate extraction) are tracked separately.

## Checklist

### DIFF
--- a/docs/services/iam.md
+++ b/docs/services/iam.md
@@ -395,7 +395,7 @@ account key carries no identity policies of its own.
   own spelling (the prefix match is case-sensitive). They compose with `IfExists`
   (`ForAnyValue:StringEqualsIfExists`). `ForAllValues:` over an empty set matches vacuously
   and `ForAnyValue:` over an empty set does not match, so pair `ForAllValues:` with
-  `"Null":{"<key>":"false"}` as you would on AWS — `Null` treats a present-but-empty set as
+  `"Null":{"<key>":"false"}` as you would on AWS: `Null` treats a present-but-empty set as
   absent, so the guard fires either way.
 - When a condition lists several values, a positive operator matches if the request value
   equals **any** of them; a negated operator (`StringNotEquals`, `ArnNotLike`, `NotIpAddress`,
@@ -407,20 +407,20 @@ account key carries no identity policies of its own.
 A `Condition` operator can only match a key floci actually places in the request context.
 floci populates:
 
-- `s3:prefix`, `s3:delimiter`, `s3:max-keys` — from the S3 request parameters.
-- `aws:PrincipalArn` — the caller's ARN, resolved from the signing access key. It is the
+- `s3:prefix`, `s3:delimiter`, `s3:max-keys`: from the S3 request parameters.
+- `aws:PrincipalArn`: the caller's ARN, resolved from the signing access key. It is the
   IAM-user ARN for a user access key, the assumed-role ARN for an STS session, and
   `arn:aws:iam::<account>:root` for the bare account-id key (floci's account-root principal),
   matching the ARN shape AWS itself reports for the account root. It is **absent** only for
   unknown keys, where nothing about the caller can be resolved.
-- `dynamodb:LeadingKeys`, `dynamodb:Attributes`, `dynamodb:Select` — from the DynamoDB request
+- `dynamodb:LeadingKeys`, `dynamodb:Attributes`, `dynamodb:Select`: from the DynamoDB request
   body, for `GetItem`, `PutItem`, `UpdateItem`, `DeleteItem`, `Query`, `BatchGetItem` and
   `BatchWriteItem` (`dynamodb:Select` for `Query` and `Scan`). `LeadingKeys` holds the
-  partition-key values the request names — from `Key`, `Item`, the `KeyConditionExpression`
+  partition-key values the request names: from `Key`, `Item`, the `KeyConditionExpression`
   equality, the legacy `KeyConditions` `EQ` entry, or each `RequestItems` entry. `Attributes`
-  holds the attribute names the request *names* — item and key fields, `AttributesToGet`,
+  holds the attribute names the request *names* (item and key fields, `AttributesToGet`,
   projection / update / filter / condition / key-condition expressions, and
-  `ExpressionAttributeNames`; a request with no projection returns every attribute while
+  `ExpressionAttributeNames`); a request with no projection returns every attribute while
   reporting only the names it mentions, exactly as on AWS, which is why AWS pairs `Attributes`
   with `dynamodb:Select`. Each key is **omitted** when it cannot be determined (unknown table,
   a `Key` that omits the partition attribute, an unparseable `KeyConditionExpression`, a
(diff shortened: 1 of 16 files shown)


pgermosen (commented):
Really solid work here. Checked both of the earlier findings directly rather than trusting the commit message, the GSI partition-key resolution and the UpdateExpression right-hand-side attribute capture are both genuinely fixed. Also went and verified the ForAllValues/ForAnyValue empty-set semantics against AWS's own IAM documentation since that's an easy detail to get backwards, and it's exactly right. The enforcement wiring holds up too: the lazy DI genuinely avoids a cyclic dependency, table resolution skips the expensive item-count scan on the hot path, and a multi-table batch correctly fails closed instead of guessing a key schema.

One nit worth a quick fix: the class-level Javadoc on DynamoDbConditionKeys says an omitted condition key always "fails closed," but that's only true for LeadingKeys. For Attributes under ForAllValues it's actually the opposite, under-reporting is the unsafe direction, and your own inline comment on addExpressionAttributes says so correctly. The code itself is written right for that risk, it's just the top-level doc comment overgeneralizing in a way that could mislead a future reader into treating "fails closed" as a blanket guarantee across all three condition keys. Worth tightening that comment to call out Attributes as the exception.

Once that's updated, this is good to go.

vuthehuyht:
Updated the class-level Javadoc on `DynamoDbConditionKeys` to explicitly clarify that the "fail closed" behavior applies to `dynamodb:LeadingKeys`, whereas for `dynamodb:Attributes` evaluated under `ForAllValues`, under-reporting is the unsafe direction because omitting a referenced attribute could allow an unauthorized operation to pass.

pgermosen (approved):
The doc fix addresses the nit exactly, it now correctly scopes "fails closed" to LeadingKeys and calls out Attributes/ForAllValues as the case where under-reporting is the unsafe direction. Everything else from the earlier review still holds. Approving.

PR #3015, opened 2026-09-03
fix(cloudfront): echo TrustedSigners in cache behaviors so the AWS provider does not segfault

matanbaruch:
## What

Creating an `aws_cloudfront_distribution` against floci crashes the Terraform/pulumi AWS provider on read-back. `DefaultCacheBehavior` (and `CacheBehavior`) omit the `TrustedSigners` object, but the provider does `len(dcb.TrustedSigners.Items)` with **no nil guard** (`flattenDefaultCacheBehavior`), so an omitted object is a nil pointer dereference:

```
[signal SIGSEGV ...]
terraform-provider-aws/internal/service/cloudfront.flattenDefaultCacheBehavior(...)
  distribution_configuration_structure.go
...resourceDistributionRead -> resourceDistributionCreate
```

The create succeeds, then the read-back panics and kills the whole apply, orphaning the distribution.

## Fix

Floci models only the modern `TrustedKeyGroups`, but real AWS **always** echoes a `TrustedSigners` object even when unused (`{Enabled:false, Quantity:0}`). Emit that disabled form in both cache-behavior writers (`xmlDefaultCacheBehavior`, `xmlCacheBehavior`), right next to `TrustedKeyGroups`. One small helper, two call sites.

## Test

`defaultCacheBehaviorEchoesTrustedSignersSoTheAwsProviderDoesNotSegfault` creates a distribution and asserts `getDistribution`'s `DefaultCacheBehavior` echoes a disabled `TrustedSigners`. It fails (object absent) before the fix and passes after. All 174 CloudFront tests stay green.

## Scope

Refs #2767. That issue lists three read-back gaps in the same object; this PR fixes the one that **crashes** the provider (TrustedSigners). The other two are fidelity/perpetual-diff issues, not crashes, and are left as follow-ups:
- `ForwardedValues`, `MinTTL`, `DefaultTTL`, `MaxTTL` dropped on read-back
- `AllowedMethods.Items` returns cached methods merged in as duplicates, with `CachedMethods` missing

(A sibling gap from #2767, LambdaFunctionAssociations/FunctionAssociations, was already fixed by #2915.)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
@@ -1968,6 +1968,7 @@ public class CloudFrontController {
                 .elem("ResponseHeadersPolicyId", dcb.getResponseHeadersPolicyId())
                 .elem("Compress", dcb.isCompress());
 
+        xml.raw(xmlTrustedSigners());
         xml.raw(xmlTrustedKeyGroups(
                 dcb.isTrustedKeyGroupsEnabled(), dcb.getTrustedKeyGroups()));
 
@@ -1997,6 +1998,7 @@ public class CloudFrontController {
                 .elem("ResponseHeadersPolicyId", cb.getResponseHeadersPolicyId())
                 .elem("Compress", cb.isCompress());
 
+        xml.raw(xmlTrustedSigners());
         xml.raw(xmlTrustedKeyGroups(
                 cb.isTrustedKeyGroupsEnabled(), cb.getTrustedKeyGroups()));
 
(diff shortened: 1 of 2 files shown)


matanbaruch:
@hectorvent gentle ping on this one. CI is green (32/32) and it's a small self-contained fix for the TrustedSigners nil-deref (Refs #2767). Anything needed from me to get it reviewed?

hectorvent (approved):
Sorry for the wait, and thank you for the stack trace; that made the case in one read. Emitting the disabled `TrustedSigners` block is right: the model requires `Enabled` and `Quantity` inside it, AWS returns it on every read, and the provider dereferences it without a guard.

**(follow-up, separate PR)** The new test reads only the default behavior; one assertion on an ordered `CacheBehavior` would pin the second writer too.

No blockers from my side.

PR #3019, opened 2026-09-03
fix(ssm): set the Arn attribute on AWS::SSM::Parameter

mariotoffia:
## Summary

`AWS::SSM::Parameter` now returns its `Arn` through `Fn::GetAtt`.

The current AWS registry schema for this type lists `Arn` as its only read-only property. `SsmCfnProvisioner` did not set it, so `Fn::GetAtt [Param, Arn]` resolved to the text `Param.Arn`, and `CfnSchemaCoverageTest` fails on `main` as soon as the schema corpus is present in `local/aws/cfn-resource-schemas/us-east-1/`.

The ARN follows the AWS form `arn:aws:ssm:<region>:<account>:parameter/<name>`. A name without a leading slash gets one inserted, a name with one is not doubled.

| Item | Status | Note |
| --- | --- | --- |
| `Fn::GetAtt Arn` | ✅ | New |
| `Ref`, `Fn::GetAtt Name`, `Type`, `Value` | ✅ | Unchanged |

### Type of change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Docs / chore

### Checklist

* [ ] `./mvnw test` passes locally. Ran per class: `LeafCfnProvisionerTest`, `CfnSchemaCoverageTest` (with the schema corpus), `CfnResourceInventoryTest`, `CloudFormationSsmParameterIntegrationTest`. `make docs-check` passes.
* [x] New or updated integration test added. Unit tests in `LeafCfnProvisionerTest` assert the exact attribute set and the leading-slash rule. The existing integration test already covers the resource; no protocol change.
* [x] Commit messages follow Conventional Commits

### DIFF
--- a/docs/services/cloudformation.md
+++ b/docs/services/cloudformation.md
@@ -231,7 +231,7 @@ before provisioning:
 - `AWS::SSM::Parameter::Value<List<String>>` and `AWS::SSM::Parameter::Name` typed parameters are
   **not yet** resolved — they are passed through as their literal input.
 
-The `AWS::SSM::Parameter` **resource** type exposes `Value`, `Type`, and `Name` attributes through
+The `AWS::SSM::Parameter` **resource** type exposes `Value`, `Type`, `Name`, and `Arn` attributes through
 `Ref` / `Fn::GetAtt` so downstream resources can consume a parameter the same stack creates.
 
 ## AWS::Include (`Fn::Transform`)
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Thank you, this is the right fix: the registry schema lists `Arn` as the type's only read-only property, the value is the `parameter/<name>` form AWS uses for slashed and unslashed names alike (moto's [`parameter_arn`](https://github.com/getmoto/moto/blob/master/moto/ssm/utils.py) applies the same rule), and the exact attribute set is pinned in `LeafCfnProvisionerTest`.

**(follow-up, separate PR)** `services/ssm/SsmService.java:132` builds the stored ARN as `"parameter" + name`, so for a name without a leading slash `GetParameter` reports `parameterdb-host` while this attribute says `parameter/db-host`. Worth aligning the service with the rule you wrote here (`services/iam/ResourceArnBuilder.java:316` already has it as well), so a stack's `Fn::GetAtt` and the API agree on one ARN.

No blockers from my side.

PR #3023, opened 2026-09-03
fix(elasticache): implement parameter group reservation to prevent deletion during provisioning

### LINKED ISSUE #3020: [BUG] elasticache: concurrent create/delete can leave a replication group referencing a deleted parameter group
## Service

ElastiCache

## AWS API Action

`CreateReplicationGroup`, `DeleteCacheParameterGroup`

## Expected behavior

`DeleteCacheParameterGroup` refuses to delete a cache parameter group that any replication group
depends on, at any point where that dependency actually exists, so a replication group can never
end up persisted with a `CacheParameterGroupName` that names a parameter group that no longer
exists.

## Actual behavior

`CreateReplicationGroup` validates that `CacheParameterGroupName` names an existing parameter
group early, before provisioning starts, but the replication group itself is not persisted to
storage until provisioning (container start, proxy start) finishes. `DeleteCacheParameterGroup`'s
in-use check only scans already-persisted replication groups. So a `DeleteCacheParameterGroup`
call for the same parameter group landing inside that provisioning window succeeds, and once the
in-flight create finishes, it persists a replication group whose `CacheParameterGroupName` points
at a parameter group that no longer exists. Nothing revalidates this afterward, so the dangling
reference is silent: `DescribeReplicationGroups` reports the group with the now-nonexistent name as
if it were still valid.

This is a narrow, low-probability window (needs two genuinely concurrent requests against the same
parameter-group name, not anything a single `terraform apply` or sequential CLI usage would
trigger), and the consequence is a metadata inconsistency rather than data loss or a crash, so it
was disclosed and merged as a known limitation in #2957 rather than blocking that PR on it.

## Reproduction

Not reliably reproducible from the CLI alone (it's a timing race), but the mechanism is exact:

1. Thread A: `CreateReplicationGroup` with `CacheParameterGroupName=my-pg` — passes the existence
   check, begins container/proxy provisioning (real wall-clock time).
2. Thread B, before Thread A's replication group is persisted: `DeleteCacheParameterGroup` for
   `my-pg` — succeeds, since no replication group referencing it exists in storage yet.
3. Thread A finishes provisioning and persists the replication group, now referencing a parameter
   group that no longer exists.

## Suggested fix

`ElastiCacheService.java` already has `provisioningGroupIds`, a reservation set used to prevent a
sibling race (two concurrent creates of the same replication group ID). The same shape would close
this one: reserve the requested `CacheParameterGroupName` as in-use for the duration of a create's
provisioning attempt (not just check the persisted-groups scan), and have
`DeleteCacheParameterGroup`'s in-use check also consult that reservation set.

## Context

Surfaced during review of #2957 (which added the sequential-case validation this race sits behind:
`CreateReplicationGroup` rejecting an unknown parameter group, and `DeleteCacheParameterGroup`
rejecting deletion of one still referenced by a persisted replication group). The PR's own
"Deliberate limits" section disclosed this exact gap.

matheusandre1:
## Summary

Fixes: #3020 

`CreateReplicationGroup` checked that `CacheParameterGroupName` existed before provisioning, but
the replication group was only stored once the container had started. `DeleteCacheParameterGroup`'s
in-use check scanned stored groups only, so a delete landing inside that window succeeded and the
create then stored a group naming a parameter group that no longer existed. This was the gap
disclosed under "Deliberate limits" in #2957.

The fix reuses the shape `provisioningGroupIds` already gives the sibling same-id race:

- `ElastiCacheService` gains `reservedParameterGroups`, a name-to-count map of parameter groups
  claimed by in-flight creates. A count rather than a set, since several concurrent creates may
  name the same parameter group.
- `createReplicationGroup` now claims the name through `reserveParameterGroup`, which runs the
  existence check and the claim under the same `lockFor(name)` monitor
  `deleteCacheParameterGroup` takes, so the two cannot straddle a delete. The claim is released in
  a `finally` once the group is stored or the attempt fails, including failures before
  provisioning starts (duplicate id).
- `isParameterGroupInUse` consults the reservation map before the stored-groups scan, so a delete
  either sees the reservation or sees the stored group; there is no longer a window between them.

Error ordering is unchanged: `CacheParameterGroupNotFound` still precedes
`ReplicationGroupAlreadyExistsFault`.

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

No new action or error. `DeleteCacheParameterGroup` answers the same `InvalidCacheParameterGroupState`
(HTTP 400) that #2957 introduced for a referenced parameter group; it is now also returned while a
`CreateReplicationGroup` naming that parameter group is still provisioning. Incorrect behaviour
before: that delete succeeded and `DescribeReplicationGroups` then reported the new group with a
`CacheParameterGroupName` that no longer described.

## Checklist

### DIFF
--- a/docs/services/elasticache.md
+++ b/docs/services/elasticache.md
@@ -116,7 +116,9 @@ AWS accepts. `DescribeCacheParameters` returns those parameters; a request for `
 A replication group that names a parameter group is refused with `CacheParameterGroupNotFound` when
 no such group exists, and a parameter group still referenced by a replication group cannot be
 deleted: `DeleteCacheParameterGroup` answers `InvalidCacheParameterGroupState` until that
-replication group is gone.
+replication group is gone. The reference counts from the moment `CreateReplicationGroup` accepts
+the name, not from when the group is stored, so a delete that lands while that create is still
+provisioning its container is refused too.
 
 ## Configuration
 
--- a/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/elasticache/ElastiCacheService.java
@@ -8,6 +8,7 @@ import io.github.hectorvent.floci.core.common.RegionResolver;
 import io.github.hectorvent.floci.core.resource.ExplorerResource;
 import io.github.hectorvent.floci.core.resource.ResourceProvider;
 import io.github.hectorvent.floci.core.resource.SupportedResourceType;
+import io.github.hectorvent.floci.core.storage.AccountAwareStorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageBackend;
 import io.github.hectorvent.floci.core.storage.StorageFactory;
 import io.github.hectorvent.floci.services.elasticache.container.ElastiCacheContainerHandle;
@@ -57,7 +58,7 @@ public class ElastiCacheService implements ResourceProvider {
 
     private final StorageBackend<String, ReplicationGroup> groups;
     private final StorageBackend<String, ElastiCacheUser> users;
-    private final StorageBackend<String, CacheParameterGroup> parameterGroups;
+    private final AccountAwareStorageBackend<CacheParameterGroup> parameterGroups;
     private final StorageBackend<String, CacheSubnetGroup> subnetGroups;
     private final ElastiCacheContainerManager containerManager;
     private final ElastiCacheProxyManager proxyManager;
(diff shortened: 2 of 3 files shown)


pgermosen (approved):
Verified the account-isolation fix directly rather than trusting the description: accountId() is a real pre-existing method that resolves the current request's account, and it's wired into the reservation key correctly. Also traced the locking, reserveParameterGroup takes the exact same monitor deleteCacheParameterGroup already held, so the two genuinely can't interleave anymore.

Test coverage is strong too, four deterministic latch-based tests covering the core race, the account-isolation case specifically with a mocked per-thread RequestContext, and the reservation release on both failure paths. The unchecked integration-test box on the checklist reads like a naming technicality, not missing coverage.

Nothing outstanding.

PR #3029, opened 2026-09-04
feat(redshift): wire the Simple Query DDL interceptor into the proxy

vuthehuyht:
## Summary

Part 3 of the #2836 split. Parts 1–2 merged (#2913, #2947); parts 4–5 (S3 COPY/UNLOAD) pending.

Wires `RedshiftSqlInterceptor` into the Redshift auth proxy's data path so Redshift-only table DDL (`DISTSTYLE`, `DISTKEY`, `SORTKEY`, `ENCODE`) sent over the PostgreSQL Simple Query (`'Q'`) wire protocol executes against the stock PostgreSQL backend.

- **`PostgresWireDecoder`**: Frames frontend client byte stream into `FrontendMessage` records. Rejects frames with length `< 4` or `> 16 MiB` before body allocation. Provides `encodeQuery(String)` for re-encoding rewritten queries.
- **`RedshiftInterceptingBridge`**: Replaces transparent relay in `RedshiftAuthProxy`. Backend-to-client is a verbatim byte pump on a virtual thread; client-to-backend frames messages and rewrites Simple Query (`'Q'`) DDL using `RedshiftSqlInterceptor.rewrite()` (fail-open: unhandled statements or exceptions forward original bytes).
- **`RedshiftInterceptorIntegrationTest`**: End-to-end integration test asserting Redshift `CREATE TABLE` and `ALTER TABLE` execution over JDBC with `preferQueryMode=simple`.
- **Docs**: Updated `docs/services/redshift.md` describing SQL interceptor wire semantics, limitations, and scope.

Scope limit: DDL path only, Simple Query protocol only, no S3Service dependency.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Validated using real PostgreSQL wire protocol clients:
- pgjdbc (`preferQueryMode=simple`) executing `CREATE TABLE sales (id int ENCODE az64, d date) DISTSTYLE KEY DISTKEY (id) COMPOUND SORTKEY (d);` and `ALTER TABLE t ADD COLUMN note varchar(20) ENCODE lzo;`.
- Transparent passthrough verified for non-DDL queries, extended protocol messages (`Parse`/`Bind`/`Execute`), and backend response packets.

## Checklist

### DIFF
--- a/docs/services/redshift.md
+++ b/docs/services/redshift.md
@@ -135,9 +135,27 @@ cluster = redshift.create_cluster(
 print(cluster["Cluster"]["Endpoint"])
 ```
 
+## SQL Interceptor
+
+Floci's Redshift auth proxy inspects frontend queries on the PostgreSQL wire protocol (Simple Query `'Q'` protocol) and rewrites common Redshift-specific table DDL so it runs on the plain PostgreSQL backend.
+
+### DDL compatibility
+
+- Redshift-only table DDL keywords are stripped before the statement is forwarded: `DISTSTYLE ALL|EVEN|KEY|AUTO`, `DISTKEY (<col>)` and column-level `DISTKEY`, `[COMPOUND|INTERLEAVED] SORTKEY (<cols>)` and column-level `SORTKEY`, and `ENCODE <codec>` for the real Redshift column encodings (`raw`, `az64`, `bytedict`, `delta`, `delta32k`, `lzo`, `mostly8`, `mostly16`, `mostly32`, `runlength`, `text255`, `text32k`, `zstd`) or `auto`.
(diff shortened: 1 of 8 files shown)


pgermosen (approved):
Read through the decoder and bridge in full given how easy this class of protocol code is to get subtly wrong. readFully handles partial TCP reads correctly, the streaming-passthrough fix for non-Q and oversized-Q frames is the right design (no memory blowup, no wrongly rejecting a large CopyData or long statement), and the socket cleanup via compareAndSet correctly handles either side hitting EOF first. Good that real pgjdbc with preferQueryMode=simple against a live backend was used for verification rather than just internal assertions. Approving.

PR #3031, opened 2026-09-04
fix(ec2): reclaim instance containers orphaned by a hard restart

allensanborn:
## Summary

`TerminateInstances` flips the record to `shutting-down` synchronously and removes the container on the async executor; launch persists the container id only once Docker has created it. A process killed across either edge (SIGKILL, OOM, `docker kill`) leaves a container on the daemon that no surviving record refers to, so nothing on the next run ever collects it.

This sweeps orphaned EC2 containers on startup, right after the restore loop so anything legitimately revived is already accounted for.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

This is emulator-internal container lifecycle hygiene, not an AWS API behavior change — no wire-protocol implications.

Incorrect behavior: neither existing path reaches these orphans. `Ec2Service.stopManagedContainers` only runs on the graceful `ShutdownEvent`, and the startup restore loop's `needsMetadataRegistration` deliberately skips records in `terminated` or `shutting-down`. Once `pruneTerminatedInstances` drops the record an hour later, the container is unreferenced for good.

Containers are matched on the existing `io.floci.service=ec2` identity label and removed only when no live record claims them. Stopped instances are explicitly not orphans — their containers are what `StartInstances` revives — so a record in any non-terminal state keeps its container. The sweep is scoped by a new `floci_owner_port` label to this Floci's own API port, since several emulators can share one Docker daemon and an unscoped sweep would reap a sibling's live instances. Containers created before this label existed carry no owner and are never swept.

Configurable via `FLOCI_SERVICES_EC2_RECONCILE_CONTAINERS_ON_STARTUP` (default `true`).

Note: upstream's #2029 independently landed the launch-cleanup half of this fix (`createAndStartContainer`/`failLaunch`/`isLaunchCancelled`); this PR carries only the startup reconciler, which upstream still lacks.

## Checklist

### DIFF
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -654,6 +654,7 @@ State is reported settled rather than transitional, as elsewhere in this service
 | `FLOCI_SERVICES_EC2_MOCK` | `false` | Skip Docker; instances jump directly to final state (useful for tests) |
 | `FLOCI_SERVICES_EC2_AWS_FAITHFUL_PRIVATE_IP` | `false` | Report the CFN/subnet-allocated private IP instead of the container bridge IP; routing and IMDS are unaffected |
 | `FLOCI_SERVICES_EC2_CONTAINER_IPS_ROUTABLE` | auto-detect | Whether an instance's container IP is reachable from the machines consuming Floci's API (Terraform, Terratest, your shell). When it is, DescribeInstances and DescribeAddresses report the container IP, so port 22 really is port 22; when it is not, they report `127.0.0.1` and reachability goes through the published high host ports. Detected by a throwaway TCP connect; set explicitly when Floci itself runs as a container |
+| `FLOCI_SERVICES_EC2_RECONCILE_CONTAINERS_ON_STARTUP` | `true` | On startup, remove instance containers this Floci left on the daemon whose record did not survive the restart or came back terminated; stopped instances are never swept |
 
 ## Requirements
 
--- a/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
+++ b/src/main/java/io/github/hectorvent/floci/config/EmulatorConfig.java
@@ -2041,6 +2041,18 @@ public interface EmulatorConfig {
          */
         Optional<Boolean> containerIpsRoutable();
 
+        /**
+         * When true, Floci removes on startup any EC2 instance container left on the Docker
+         * daemon by a previous run of <em>this same</em> Floci (matched by the
+         * {@code floci_owner_port} label) whose instance record did not survive the restart, or
+         * came back already terminated. Stopped instances are never swept — their containers are
+         * exactly what StartInstances revives.
+         *
+         * Env var: FLOCI_SERVICES_EC2_RECONCILE_CONTAINERS_ON_STARTUP
+         */
+        @WithDefault("true")
+        boolean reconcileContainersOnStartup();
+
         /** Port on the Floci host for the IMDS HTTP server (169.254.169.254 equivalent). */
         @WithDefault("9169")
         int imdsPort();
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
@@ -103,6 +103,19 @@ public class Ec2ContainerManager {
      * Containers created before this label existed carry no owner and are therefore never swept.
      */
     static final String LABEL_OWNER_PORT = "floci_owner_port";
+
+    /**
+     * Identity of the Floci deployment that owns a container, for scoping the startup sweep.
+     * The API port alone collides when two independently namespaced Flocis share a Docker daemon
+     * on the same internal port, and each would then reap the other's live containers. Composing
+     * the documented resource namespace in front of it separates exactly those deployments; an
+     * unnamespaced single Floci keeps the bare port it already stamped.
+     */
+    private String ownerIdentity() {
+        String ns = config.docker() == null || config.docker().resourceNamespace() == null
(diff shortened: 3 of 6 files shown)


allensanborn:
Thanks - all three are real, and I've verified each against the code rather than just taking the summary at face value. Fixed in 4422e688.

**Account identity dropped (P1).** Confirmed. `instanceContainerStillWanted` called `instances.get(key(region, instanceId))`, and `key()` is only `region + "::" + id`. Since `reconcileOrphanedContainers` runs from `@PostConstruct` there is no request context, so `AccountAwareStorageBackend` resolves to the default account and a non-default-account record reads as absent - the sweep would then remove a container still backing a live instance. Now goes through a `findAnyInstance()` helper that consults every account, following the same `instanceof`-and-fall-back shape `findAnyVpcEntry()` already uses in this class. The lookup is read-only, so per `findAnyAccount`'s contract a cross-account hit stays in its owner's partition rather than forking a copy.

**Owner ports can collide (P1).** Confirmed, and agreed it's lower-probability than the first but worth closing. Ownership is now `docker.resource-namespace` composed in front of the port, since that config already exists for precisely this case ("Useful when multiple Floci processes share one Docker daemon"). An unnamespaced single Floci still stamps the bare port, so nothing changes for the common deployment.

**Missing `application.yml` entry (P2).** Confirmed - `reconcile-containers-on-startup` is now in the main `application.yml`, per the AGENTS.md rule that repository YAML defines effective runtime behavior.

Added `reconcileOrphanedContainersSparesAnotherNamespacesContainerOnTheSamePort`. I checked it's a real guard and not a tautology: with the old port-only ownership restored it fails (`expected: <1> but was: <0>`), and passes with the fix. Full `Ec2ContainerManagerTest` is green.

allensanborn:
Good catch, and the impact was worse than the comment says. Fixed in 9fb4dcd8.

`Ec2ServiceConfig` maps to `floci.services.ec2` - its own env var name says so (`FLOCI_SERVICES_EC2_RECONCILE_CONTAINERS_ON_STARTUP`) - and `application.yml` has an `ec2:` block under *both* `storage.services` and `services`. I added the key to the storage one.

The consequence is not a silently ineffective toggle. SmallRye rejects a key that maps to no root rather than ignoring it, so this was `SRCFG00050 ... does not map to any root` at startup: every `@QuarkusTest` failed to boot (55 occurrences, all Errors and zero Failures across the four shards), and the native images built fine but could not start, which is why every `compat-*` and `sdk-test-*` job died at exit 124.

My mistake was verifying the previous round with `mvnw compile` only. A compile cannot catch a config-mapping error, so I have re-verified this one by actually booting the app: `Ec2ImageCatalogTest` (a `@QuarkusTest`) passes and no `SRCFG00050` remains.

Also noting for the record, since it is cosmetic and I have left it alone: `LABEL_OWNER_PORT` still reads `floci_owner_port` although `ownerIdentity()` now composes namespace and port. The label key string is unchanged so behaviour and the `EmulatorConfig` javadoc are both still accurate; only the constant name under-describes the value now. Happy to rename it if you would prefer.

allensanborn:
CI is green on 9fb4dcd8 apart from one job, `native arm64 / compat-opentofu`, which I believe is a flake rather than this change. Evidence, since I cannot rerun it myself:

```
Error: creating IAM Role (floci-compat-lambda-exec): operation error IAM: CreateRole,
https response error StatusCode: 409, EntityAlreadyExists:
Role with name floci-compat-lambda-exec already exists.
```

The container in that job is fresh (`docker run -d --name floci`, no state volume), and the server log has exactly one `Created IAM role: floci-compat-lambda-exec` at 12:43:31,821, inside the same apply window. tofu logs `Creating...` once and then gets the 409. That shape is an SDK retry of a slow `CreateRole` whose first attempt had already succeeded server side; the emulator was busy in parallel at the time (elasticache 34s, kinesis 20s, sqs 25s in the same apply).

Nothing in this PR reaches IAM: the diff is EC2 container reconciliation plus one config key (`git diff upstream/main...HEAD` touches only `Ec2ContainerManager`, `Ec2Service`, `EmulatorConfig`, `application.yml`, `docs/services/ec2.md` and the EC2 test), the job runs with `FLOCI_SERVICES_EC2_MOCK=true`, and the reconciler runs at `@PostConstruct` before any apply, removing only containers labelled `io.floci.service=ec2`. The sibling `native arm64 / compat-terraform` passed 67/67 on the same image, and compat-opentofu passed on several other runs within the hour.

I want to be precise about the strength of that claim: I could not find the same failure on main's history, so I am calling it a flake by mechanism rather than by prior occurrence. A rerun would settle it, and I do not have permission to trigger one.

Worth noting separately: `CreateRole` is not retry-idempotent here. A retried create that already succeeded returns `EntityAlreadyExists` where real IAM would too, so this may be faithful behaviour rather than a bug, but it does make the compat suite sensitive to emulator latency. Happy to file that as its own issue if it is useful.

For the record on the earlier rounds, both remaining open threads are addressed at this head: the account-identity P1 (`Ec2Service.findAnyInstance` now goes through `AccountAwareStorageBackend.findAnyAccount`) and the missing YAML entry (now at `application.yml` under `floci.services.ec2`).

hectorvent (approved):
Thank you, and thank you for verifying each Greptile point against the code rather than the summary; the account-scoped lookup and the owner identity that composes the namespace with the port are exactly the two things that make a shared daemon safe. Keeping stopped instances out of the sweep is the right line.

**(follow-up, separate PR)** The new `FLOCI_SERVICES_EC2_RECONCILE_CONTAINERS_ON_STARTUP` variable could join `docs/configuration/environment-variables.md`.

No blockers from my side.

PR #3037, opened 2026-09-04
perf(docker): drop the gosu layer and two no-op native-image inputs

hectorvent:
## Summary

The native image's tools layer installed shadow-utils, which ubi9-minimal 9.7 already
ships, and downloaded a gosu binary: 19 MB on disk, 3.66 MB compressed, byte-identical on
every release since 1.5.34. The entrypoint still starts as root and still grants the
bind-mounted Docker socket's group to the floci user, but through coreutils, which the
base image has: `chroot --userspec=1001:0 --groups=0,<socket gid> --skip-chdir /` drops to
uid 1001 with the socket's gid as a supplementary group. The layer is now 24.6 kB (one
`useradd`). The JVM Dockerfile drops its gosu download too: the entrypoint is shared, so once it execs `chroot` nothing there calls gosu, and the binary plus the wget-and-checksum block on every build were dead weight. Floci's release and nightly pipelines never publish that image; `build-images.yml` is a reusable `workflow_call` a fork invokes to publish its own copy to its own registry, so it deserves the same trim. Its base (`eclipse-temurin:25-jre-noble`, coreutils 9.4) supports `--userspec`, `--groups` and `--skip-chdir`, verified on the built image.

Two native-build inputs go with it, neither changing a byte of the binary:
`--allow-incomplete-classpath` and `--report-unsupported-elements-at-runtime` have been
no-ops since GraalVM 22.1 (Quarkus passes `--link-at-build-time` itself), and the brotli4j
platform JNI jars are excluded from `quarkus-rest-jackson` because Floci never enables
Brotli compression and the jar was never embedded. A native build without them produces
the same 232.25 MB image and the same reachable set.

Part of the image-size work that started with #3016; this PR's job summary is the first
amd64 measurement of that baseline.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No emulated surface changes. Verified on the arm64 native image: `id` inside the container
is uid=1001 gid=0; a Docker socket owned by gid 999 yields groups=0,999 end to end; `/app/data`
is writable; the `sdk-test-java` compat suite passes 1,667/1,667 against the image with the
socket mounted. `--skip-chdir` is required: without it the first run failed 58 Lambda and
Docker-backed cases with "Failed to extract deployment package: /data".

## Checklist

- [ ] `./mvnw test` passes locally (no Java changed; the three native config guards
      `ApplicationDefaultsTest`, `NativeImageRuntimeInitializationTest` and
      `ServiceConfigYamlCoverageTest` pass on this tree)
- [ ] New or updated integration test added (Dockerfile and build-config change; the
      compat CLI job and the native suite legs in `compatibility.yml` exercise it)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,9 +22,8 @@ ENV AWS_ACCESS_KEY_ID=test
 ENV AWS_SECRET_ACCESS_KEY=test
 ENV AWS_CONFIG_FILE=/etc/floci/aws/config
 
-ENV GOSU_VERSION=1.17
-ENV GOSU_AMD64_SHA256=bbc4136d03ab138b1ad66fa4fc051bafc6cc7ffae632b069a53657279a450de3
-ENV GOSU_ARM64_SHA256=c3805a85d17f4454c23d7059bcb97e1ec1af272b90126e79ed002342de08389b
+# Privilege drop happens in the entrypoint through coreutils chroot, so no gosu
+# download; wget stays for the HEALTHCHECK below.
 RUN set -eux; \
     apt-get update; \
     apt-get install -y --no-install-recommends ca-certificates passwd python3 python3-pip wget; \
@@ -34,16 +33,7 @@ RUN set -eux; \
     mv "$aws_bin" /usr/bin/aws.real; \
     printf '#!/bin/sh\nexec /usr/bin/aws.real --endpoint-url=http://localhost:4566 "$@"\n' > /usr/bin/aws; \
     chmod +x /usr/bin/aws; \
-    useradd -r -u 1001 -g root -d /app -s /sbin/nologin floci; \
-    arch="$(dpkg --print-architecture)"; \
-    case "$arch" in \
-        amd64) gosuArch='amd64'; gosuSha256="$GOSU_AMD64_SHA256" ;; \
-        arm64) gosuArch='arm64'; gosuSha256="$GOSU_ARM64_SHA256" ;; \
-        *) echo >&2 "unsupported arch: $arch"; exit 1 ;; \
-    esac; \
-    wget -q -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/${GOSU_VERSION}/gosu-${gosuArch}"; \
-    echo "${gosuSha256}  /usr/local/bin/gosu" | sha256sum -c -; \
-    chmod +x /usr/local/bin/gosu
+    useradd -r -u 1001 -g root -d /app -s /sbin/nologin floci
 
 WORKDIR /app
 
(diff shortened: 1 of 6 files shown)


pgermosen (commented):
This is a clean change overall. The chroot --userspec/--groups swap for gosu is well explained and correct: chroot --groups takes raw gids directly so it also removes the old per-boot groupadd/getent dance, and --skip-chdir is exactly right since chroot / would otherwise reset the working directory away from /app. The no-op GraalVM flags and the brotli4j platform-jar exclusions check out too.

One thing worth fixing before merge: the summary says the JVM Dockerfile keeps gosu because it isn't a published image, but it is, build-images.yml pushes it as the JVM image alongside the native and compat ones. Since entrypoint.sh is shared between the two Dockerfiles and now execs chroot unconditionally, there's no code path left in the JVM image that ever calls gosu. The GOSU_VERSION/GOSU_*_SHA256 envs and the wget-download-checksum-chmod block in docker/Dockerfile are downloaded and verified on every build but never invoked, they're dead weight in a published image, which is the opposite of what this PR is going for elsewhere. Worth dropping that block the same way it came out of Dockerfile.native-package, and fixing the summary's claim about it since this repo squash-merges and that text becomes the commit message.

Small one along the same lines: the comments in test-entrypoint.sh still describe "the root-only gosu block" and say the root path re-execs via gosu, that's stale now that it's chroot. And separately, docker/Dockerfile.native (the docker-compose option for testing the native image locally, not part of any published build) still does the old shadow-utils-install-plus-gosu-download version, that one's out of scope for this PR but is now the last place doing it the old way if you ever want to circle back.

hectorvent:
Thanks, pgermosen. One correction to the premise first, then the change, which I made anyway.

Floci does not publish an image from `docker/Dockerfile`. `release.yml` and `nightly.yml` build only the native and compat images, and `build-images.yml` is `on: workflow_call`, a reusable workflow a fork invokes to publish its own copy to its own registry. So the JVM image is community-built rather than ours, and the summary's "not a published image" meant that, though it read as if nothing could build it.

Keeping the two Dockerfiles aligned is still the right call: the entrypoint is shared, so with it execing `chroot` nothing in the JVM image called gosu any more, and a fork building from it would have been downloading and checksumming a binary that never runs. At 3b207db the JVM Dockerfile drops the gosu block too (wget stays for the HEALTHCHECK), and I checked the base before relying on it: `eclipse-temurin:25-jre-noble` carries coreutils 9.4 with `--userspec`, `--groups` and `--skip-chdir`, and the built image drops to uid 1001, gid 0 with `/app` as the working directory. The stale `test-entrypoint.sh` comments now say chroot, and the description says the above instead of "not a published image".

`Dockerfile.native` is the compose-only holdout; I will take it in a follow-up as you suggested.

pgermosen (approved):
Looks good.

PR #3041, opened 2026-09-04
fix(lambda): give each published version its own copy of the code

### LINKED ISSUE #2958: [BUG] lambda: published versions share one code directory with $LATEST, so a redeploy changes what a version runs

avison9:
## Summary

Closes #2958.

A published Lambda version's metadata was immutable, but the directory that metadata named was not.
`CodeStore.getCodePath` has no version component, so every version shared one directory with
`$LATEST`, and `publishVersion` copied only the `codeLocalPath` pointer. A later
`UpdateFunctionCode` replaces that directory wholesale, so it silently rewrote what an
already-published version ran: the version kept advertising the `CodeSha256` it was published with
while the bytes behind it were a different build, and `ContainerLauncher` tar-copies that directory
at launch, so a qualified invoke executed `$LATEST`'s code. That defeats the point of pinning to a
version at all: an alias held at version 1 for a canary, a rollback to a known-good build, or a
stack referencing `:1` all quietly started running the newest deploy.

`publishVersion` now copies the code into a directory belonging to that version, so the version owns
the bytes it was published from. Image-backed and hot-reload functions are carved out and keep the
reference they had: an image is already immutable by digest, and a hot-reload function's whole point
is that its bind-mounted directory tracks the developer's working tree. A copy failure falls back to
the old shared path rather than failing the publish, which leaves the version exactly as good as
every version published before this existed instead of turning a working call into an error.

```bash
# handler.py returns the string it was built with.
$ zip -j v1.zip handler.py            # returns "v1"
$ aws --endpoint-url=http://localhost:4566 lambda create-function \
      --function-name greeter --runtime python3.12 --handler handler.handler \
      --role arn:aws:iam::000000000000:role/r --zip-file fileb://v1.zip

$ aws --endpoint-url=http://localhost:4566 lambda publish-version --function-name greeter
{ "Version": "1", "CodeSha256": "8Xw...=" }

$ zip -j v2.zip handler.py            # rebuilt, now returns "v2"
$ aws --endpoint-url=http://localhost:4566 lambda update-function-code \
      --function-name greeter --zip-file fileb://v2.zip

# Invoke the PINNED version, which nothing should have changed.
$ aws --endpoint-url=http://localhost:4566 lambda invoke \
      --function-name greeter:1 out.json && cat out.json
```

```
BEFORE:  "v2"     wrong. Version 1 runs the build that replaced it, while still
                  reporting CodeSha256 8Xw...= for the build it no longer has.

AFTER:   "v1"     correct. Version 1 runs the bytes it was published from, and
                  those bytes still hash to 8Xw...=.
```

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Real AWS treats a published version as an immutable snapshot of code plus configuration.
`UpdateFunctionCode` only ever moves `$LATEST`; it cannot alter a version that has already been
published, and `GetFunction` on a qualified name returns the code and the `CodeSha256` that version
was created with for the life of that version.

Floci diverged on the code, not on the metadata. The version record was correct and frozen, and
every existing version test asserts those fields and still passes. What was wrong was that the
`codeLocalPath` on that frozen record pointed at a directory the next deploy would overwrite, so the
`CodeSha256` a qualified `GetFunction` returned stopped describing the code a qualified `Invoke`
would run. There is no new wire protocol here and no new action: this restores an existing action's
documented behaviour, so the correction is verified against `LambdaVersionIntegrationTest` and the
unit coverage listed below rather than against a new SDK call.

Two secondary divergences are also fixed, both raised in review of the first cut of this PR:

- **`DeleteFunction` could destroy an unrelated function.** Version directories were named
  `<function>.v<n>` as siblings, and whole-function delete swept them with a bare
  `startsWith(name + ".v")`. Real AWS restricts `FunctionName` to `[a-zA-Z0-9-_]` so `foo.v1` cannot
  exist there, but Floci's `CreateFunction` validates only that the name is non-blank, making the
  collision reachable through Floci itself. Deleting `foo` wiped the code of a function genuinely
  named `foo.v1`, whose next invoke then launched an empty container and hung to the function
  timeout (the #1987 failure mode) with nothing recorded against the delete that caused it.
- **`DeleteFunction --qualifier` leaked storage.** Deleting a single published version removed the
  record but never its code, so only whole-function delete reclaimed anything and a repeated
  publish/delete cycle grew on disk without bound.

## What changed

- **`CodeStore`**: new `getVersionsPath` returning `<codePath>/<account>/<function>@versions`, a
  container holding one subdirectory per version. `deleteVersions` now deletes that one directory
  and the prefix scan is gone. New `deleteVersion` removes a single version's code.
- **`LambdaService`**: `deleteFunctionVersion` reclaims the deleted version's directory, inside the
  per-function lock it already held, and only when that version actually owned one. A version that
  fell back to `$LATEST`'s path, and every image-backed or hot-reload version, is left alone;
  deleting the path those record would take the live function's code with it.
- `versionCodePath` is called from inside `publishVersion`'s `lockForConcurrencyOp` block, so the
  copy cannot interleave with a concurrent `UpdateFunctionCode`.
- Rebased onto current `main`. `main` added `latestPublishedFrom` and `newestVersion` at the same
  point this branch inserted `versionCodePath`; the conflict is resolved by keeping all three, and
  `main`'s newer locking around the `PublishVersion` `CodeSha256` precondition is preserved intact.

### Why this shape rather than a delimiter-safe prefix match

Review asked for a delimiter-safe match instead of the bare `startsWith`. Tightening it to
`<name>.v<digits>` fixes `foo.v` but not `foo.v1`, which is the same collision one character over:
the real problem is that version directories and function directories shared one namespace, so no
matching rule over that layout can be correct. `@versions` is chosen deliberately, not
decoratively: `sanitizeName` maps every name into `[a-zA-Z0-9_.-]`, so no function, whatever a user
calls it, can produce a directory name containing `@`. The namespaces are disjoint by construction,
the prefix scan and its failure handling are deleted rather than patched, and the fix does not
depend on version qualifiers happening to be numeric.

## How it was tested

Baseline before this revision, on the rebased branch: **337 run, 0 failures, 0 errors, 0 skipped**.
After: **341 run, 0 failures, 0 errors, 0 skipped** (`Lambda*Test`, `CodeStore*Test`,
`ZipExtractor*Test`, integration excluded). Net +4 is two tests relocated and six added.

| Test | Covers |
|---|---|
| `LambdaServiceTest.aPublishedVersionKeepsItsOwnCodeWhenLatestIsRedeployed` | The reported bug: publish v1, redeploy `$LATEST`, assert the version still holds v1's bytes, that `$LATEST` extracted elsewhere, and that the version's `CodeSha256` still describes what is on disk |
| `LambdaServiceTest.deletingAFunctionRemovesItsPublishedVersionsCode` | A version's code does not outlive its function |
| `LambdaServiceTest.deletingOnePublishedVersionReclaimsOnlyThatVersionsCode` | Deleting version 1 of a two-version function reclaims only version 1, leaving version 2 and `$LATEST` intact and readable |
| `CodeStoreTest.aVersionDirectoryCannotCollideWithAnotherFunctionsOwnDirectory` | A function named `foo.v1` survives `delete("foo")` with both its `$LATEST` and its version code, while `foo`'s version code is still reclaimed |
| `CodeStoreTest.deleteVersionRemovesOneVersionAndLeavesTheRestInPlace` | `deleteVersion` is scoped to one version |
| `CodeStoreTest.versionCodeLivesOutsideTheDirectoryExtractionReplaces` | The layout invariant: a version directory is never under the directory extraction replaces (#2672) |

Both new regression tests were confirmed to fail against the pre-fix behaviour, by restoring the
flat `.v<n>` naming and removing the reclaim call:

```
CodeStoreTest.aVersionDirectoryCannotCollideWithAnotherFunctionsOwnDirectory:122
  deleting foo must not remove a function literally named foo.v1 ==> expected: <true> but was: <false>
LambdaServiceTest.deletingOnePublishedVersionReclaimsOnlyThatVersionsCode:1547
  the deleted version's code must be reclaimed ==> expected: <false> but was: <true>
```

The three service-level tests now live in `LambdaServiceTest` and the separate
`LambdaVersionCodeIsolationTest` class is deleted, resolving the naming point from review:
`AGENTS.md` documents `*ServiceTest.java` and `*IntegrationTest.java`, and that class was neither.

## Risk and rollback

Low. No public API signature change, no wire-protocol change, no configuration change. Behaviour
changes are confined to the on-disk layout under `floci.services.lambda.code-path` and to what
`DeleteFunction --qualifier` cleans up. No migration is needed: the `.v<n>` layout only ever existed
on this unmerged branch, so there is no released on-disk format to move off. Functions created
before this change keep working, including versions published while `$LATEST` was still on the
pre-account-scoped legacy path, which
`LambdaAccountScopedCodeTest.aPublishedVersionsCodeSurvivesLatestMigratingOffTheLegacyPath` covers.
Rollback is reverting the commit; the only residue would be orphaned `<function>@versions`
directories, which are inert and go when the function is deleted.

Disk use is now one extracted package per live published version, released when the version is
deleted. `copyForVersion` runs once per publish inside the per-function lock, so a very large
deployment package briefly serializes other operations on that one function. That is the price of
doing the copy where it cannot race a concurrent deploy.

## Note on the overlap with #3136

#3136 fixes the same issue with content addressing. It is the more elegant design and its storage
dedup is real, and I would be glad to see it land eventually. The case for this PR going first is
blast radius: it changes nothing on the deploy path, whereas content addressing changes the
extraction path for every function and needs a reclaim pass over historical builds whose filter has
to avoid deleting in-flight extractions. Review of #3136 found a null `codeLocalPath` crash and a
race from extracting outside the per-function lock in exactly that machinery, and greptile's own
citation there points at a prior rollback of a Lambda code-volume cleanup attempt. Happy to defer if
maintainers would rather take that approach; flagging it so two PRs do not sit on one issue.

## Checklist

## Out of scope, worth its own issue

`CreateFunction` does not validate the `FunctionName` charset (only that it is non-blank), which is
why the `foo.v1` collision above was reachable at all and is itself a divergence from AWS. That is a
separate behaviour change with its own compatibility question, so it is deliberately not bundled
here. This PR is correct with or without it.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -850,7 +850,8 @@ public class LambdaService implements ResourceProvider {
         synchronized (lockForConcurrencyOp(fn.getFunctionArn())) {
             warmPool.drainEnvironment(version.get());
             functionStore.deleteVersion(region, name, qualifier);
-            // The snapshot shares $LATEST's code directory, so this only reclaims once no
+            reclaimVersionCodeDirectory(fn, qualifier, version.get());
+            // The snapshot may still share $LATEST's code directory, so this only reclaims once no
             // remaining version references it.
             reclaimLegacyCodeDirectoryIfUnused(name);
         }
@@ -894,6 +895,28 @@ public class LambdaService implements ResourceProvider {
             }
         }
         LOG.infov("Deleted Lambda function: {0}", functionName);
+    }
+
+    /**
+     * Drops the code directory a deleted version owned. Without this the only thing that ever
+     * reclaimed version code was deleting the whole function, so a repeated publish/delete cycle
+     * left one package on disk per version ever published.
+     *
+     * <p>Guarded on the version actually owning that directory rather than deleting it outright.
+     * A version whose copy could not be made fell back to {@code $LATEST}'s path, and image-backed
+     * and hot-reload versions never had a copy at all: for those the recorded path is the live
+     * function's own directory, and removing it would delete the code {@code $LATEST} still runs.
+     */
+    private void reclaimVersionCodeDirectory(LambdaFunction fn, String version, LambdaFunction snapshot) {
+        String recorded = snapshot.getCodeLocalPath();
+        if (recorded == null) {
+            return;
+        }
+        String owned = codeStore.getVersionCodePath(ownerAccount(fn), fn.getFunctionName(), version)
+                .toAbsolutePath().normalize().toString();
+        if (owned.equals(Path.of(recorded).toAbsolutePath().normalize().toString())) {
+            codeStore.deleteVersion(ownerAccount(fn), fn.getFunctionName(), version);
+        }
     }
 
     /**
@@ -1792,6 +1815,31 @@ public class LambdaService implements ResourceProvider {
                 .max(java.util.Comparator.comparingLong(v -> Long.parseLong(v.getVersion())));
     }
 
+    /**
+     * Copies the function's current code into a directory belonging to this version, falling back
+     * to {@code $LATEST}'s path if there is nothing to copy or the copy fails.
+     *
+     * <p>A copy failure must not fail the publish: the version is still a correct snapshot of the
+     * configuration, and falling back leaves it exactly as good as every version published before
+     * this existed, rather than turning a working call into an error.
+     */
+    private String versionCodePath(LambdaFunction fn, String version) {
+        String current = fn.getCodeLocalPath();
+        if (current == null || fn.getHotReloadHostPath() != null) {
+            return current;
+        }
+        try {
+            Path copied = codeStore.copyForVersion(
+                    ownerAccount(fn), fn.getFunctionName(), version, Path.of(current));
+            return copied == null ? current : copied.toAbsolutePath().normalize().toString();
+        } catch (IOException e) {
+            LOG.warnv("Could not give version {0} of {1} its own code directory, "
+                            + "falling back to the shared one: {2}",
+                    version, fn.getFunctionName(), e.getMessage());
+            return current;
+        }
+    }
+
     private int nextVersionNumber(String counterKey, String legacyCounterKey) {
         synchronized (versionCounterLocks.computeIfAbsent(counterKey, k -> new Object())) {
             Integer current = versionCounters.get(counterKey);
@@ -1908,7 +1956,13 @@ public class LambdaService implements ResourceProvider {
             // with nothing in it, and hangs to the function timeout instead of failing (#1987). A
             // published version is an immutable snapshot of code plus configuration, so it carries the
             // code location for every package type, not only Zip.
-            snapshot.setCodeLocalPath(fn.getCodeLocalPath());
+            // A version's own copy of the code, not a reference to $LATEST's directory. Sharing that
+            // directory meant a later UpdateFunctionCode rewrote what an already-published version
+            // ran, so the version advertised one CodeSha256 over a different build (issue #2958).
+            // Nothing to copy for image-backed or hot-reload functions, which keep the reference
+            // they had: an image is already immutable by digest, and a hot-reload function's whole
+            // point is that its bind-mounted directory tracks the developer's working tree.
+            snapshot.setCodeLocalPath(versionCodePath(fn, String.valueOf(version)));
             snapshot.setCodeSha256(fn.getCodeSha256());
             snapshot.setS3Bucket(fn.getS3Bucket());
             snapshot.setS3Key(fn.getS3Key());
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/zip/CodeStore.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/zip/CodeStore.java
@@ -6,6 +6,7 @@ import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
 import java.io.IOException;
+import java.nio.file.StandardCopyOption;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Comparator;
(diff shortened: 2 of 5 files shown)


pgermosen (commented):
This needs a rebase first, showing as conflicting against main.

Beyond that, there are three findings on this from greptile, none addressed yet. The real one: deleteVersions matches version directories with a bare startsWith(functionName + ".v"), so deleting function foo would also wipe out a different function literally named foo.v, corrupting its live published versions with no error. Checked whether this is actually reachable, real AWS doesn't allow dots in function names, but floci's own CreateFunction has no charset validation on FunctionName today, only a non-blank check, so this collision is genuinely reachable through floci itself, not just a theoretical concern against real AWS. Worth a delimiter-safe match instead of a bare startsWith.

Second one is filed P2 but worth taking seriously given the project's history here: deleting an individual published version never cleans up its code directory, only whole-function delete does, so repeated publish/delete leaves every package on disk. Worth noting that greptile's own citation on this points at a prior rollback of an earlier Lambda code-volume cleanup attempt, this area has bitten the project before.

Third is minor, the new test class name doesn't match either of the two conventions AGENTS.md documents (*ServiceTest.java or *IntegrationTest.java).

Everything else holds up, the hot-reload and image-backed carve-outs are implemented correctly, the copy-failure fallback does what the description says, and making the version directory a sibling rather than a child of $LATEST's correctly respects the wholesale-replace constraint from #2672.

One more thing: this is missing the AWS Compatibility section the template requires, same gap as on #3010. Since this repo squash-merges that becomes the commit message.

hectorvent (approved):
pgermosen's review above is the one to read; as the maintainer I checked the current head against it: the version directories live under a per-function suffix so the cleanup cannot cross into another function, deleting one version reclaims just its package, and the branch merges cleanly now.

Nothing further from me. Merging.

PR #3043, opened 2026-09-04
feat(bedrock-runtime): stream ConverseStream incrementally

scottsteen:
## Summary

ConverseStream previously buffered the whole proxied response, translated it in one shot, and returned it as a single byte[]. Both backends now write frames through a JAX-RS StreamingOutput as they're known, and the proxy backend reads the upstream SSE response via HttpResponse.BodyHandlers.ofLines() instead of ofString() - text deltas reach SDK clients as the model generates them instead of landing all at once when the response finishes.

Once the StreamingOutput starts writing, JAX-RS has already committed the 200 response, so a failure discovered only after streaming begins - a stream that never reaches a finish_reason or "[DONE]" (upstream connection dropped mid-generation) - can no longer become an HTTP error status. It's now reported as a modelStreamErrorException event, the same wire shape (:message-type: exception, :exception-type: <name>) and modeled exception real Bedrock uses for a mid-stream failure, verified to decode correctly through BedrockRuntimeAsyncClient's generated ConverseStreamResponseHandler. A stream that completes normally (finish_reason or "[DONE]" seen) with no content is now treated as a legitimate empty completion rather than an error, matching real Bedrock.

Added a compatibility-tests/sdk-test-java case exercising BedrockRuntimeAsyncClient.converseStream end to end, proving the emitted event-stream frames decode correctly through the real AWS SDK rather than only Floci's own byte-level assertions.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ✅ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

`sdk-test-java` test added - streaming not supported by the CLI.

## Checklist

- [ ✅ ] `./mvnw test` passes locally
- [ ✅ ] New or updated integration test added
- [ ✅ ] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

### DIFF
--- a/compatibility-tests/sdk-test-java/pom.xml
+++ b/compatibility-tests/sdk-test-java/pom.xml
@@ -333,6 +333,11 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>bedrockagentcore</artifactId>
         </dependency>
+        <!-- Bedrock Runtime client (Converse/ConverseStream) -->
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>bedrockruntime</artifactId>
+        </dependency>
         <!-- AWS IoT client -->
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
--- a/compatibility-tests/sdk-test-java/src/main/java/com/floci/test/TestFixtures.java
+++ b/compatibility-tests/sdk-test-java/src/main/java/com/floci/test/TestFixtures.java
@@ -67,6 +67,7 @@ import software.amazon.awssdk.services.ec2.Ec2Client;
 import software.amazon.awssdk.services.ecr.EcrClient;
 import software.amazon.awssdk.services.bedrockagentcore.BedrockAgentCoreClient;
 import software.amazon.awssdk.services.bedrockagentcorecontrol.BedrockAgentCoreControlClient;
+import software.amazon.awssdk.services.bedrockruntime.BedrockRuntimeAsyncClient;
 import software.amazon.awssdk.services.pipes.PipesClient;
 import software.amazon.awssdk.services.codebuild.CodeBuildClient;
 import software.amazon.awssdk.services.codedeploy.CodeDeployClient;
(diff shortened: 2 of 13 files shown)


pgermosen (commented):
This is a nice piece of work, the modelStreamErrorException handling for a truncated upstream stream is exactly what real Bedrock does once the 200 is already committed, and the sdk-test-java case actually decoding through BedrockRuntimeAsyncClient's real unmarshaller is a good way to prove the wire format rather than just trusting Floci's own byte assertions.

One thing worth fixing: in ProxyBackend.converseStream, the non-2xx branch reads the upstream stream without closing it:

```java
if (response.statusCode() >= 300) {
    String errorBody = response.body().limit(50).collect(Collectors.joining("\n"));
    ...
}
```

response.body() here is the Stream<String> from HttpResponse.BodyHandlers.ofLines(), and per the JDK docs it needs to be closed to release the underlying connection, a terminal op like collect() doesn't do that for you. The success path handles this correctly with try (sseLines), but this branch doesn't, so every proxy-backend error response on this endpoint leaks the stream/connection instead of returning it to the pool. Worth wrapping it the same way, e.g. try (Stream<String> lines = response.body()) { errorBody = lines.limit(50).collect(...); }.

scottsteen:
> One thing worth fixing: in ProxyBackend.converseStream, the non-2xx branch reads the upstream stream without closing it

@pgermosen Nice catch! Have corrected that now. 

I also cleaned up a bunch of duplication I introduced in `ProxyBackend`.

pgermosen (approved):
Traced through the streaming changes in ProxyBackend, BedrockOpenAiTranslator, and BedrockStreamEncoder, and checked the modelStreamErrorException behavior against AWS's own ConverseStream API reference — it's a real documented in-band event for exactly this mid-stream-failure case, and the design here matches it. Also confirmed all three greptile findings (tool block indices, malformed tool identity validation, the em dash in the docs) are actually fixed in the current code, not just marked resolved, and that the stream-leak fix from earlier in the thread is in too. Nice work, approving.

PR #3048, opened 2026-09-04
fix(s3): verify virtual-hosted presigned URLs

### LINKED ISSUE #3045: [BUG] GetObject presigned URL verification fails (SignatureDoesNotMatch) for virtual-hosted-style requests when FLOCI_SERVICES_S3_ENFORCE_AUTH=true
## Service

S3 (presigned URL signature verification)

## AWS API Action

`GetObject` (presigned URL)

## Expected behavior

The presigned URL verifies regardless of addressing style.

Virtual-hosted-style (`bucket.host/key`) is the default addressing style for S3 on real AWS, and AWS accepts and serves presigned URLs generated in that style. Path-style (`host/bucket/key`) is legacy but still supported. Both must verify.

## Actual behavior

With `FLOCI_SERVICES_S3_ENFORCE_AUTH=true`, a presigned GET URL generated in virtual-hosted style (`bucket.localhost:4566/key`) is always rejected with `SignatureDoesNotMatch`. The same object, presigned in path-style (`localhost:4566/bucket/key`), verifies fine.

I recomputed SigV4 over the failing URL independently (secret `test`, region `eu-central-1`, host including port, `X-Amz-Content-Sha256` present in the query string) and reproduced the client's signature exactly, confirming the signature is valid for what was actually sent:

```
canonical URI:     /uploads/hello.txt
canonical headers: host:test-s3-bucket.localhost:4566
```

Floci appears to normalize virtual-hosted requests to path-style before verification, recomputing the canonical request over `/test-s3-bucket/uploads/hello.txt` with host `localhost:4566` instead. Both the URI and the host header are part of the signed canonical request, so the recomputed hash can never match a virtual-hosted signature. Verification needs to reconstruct the canonical request in the addressing style the request actually arrived in, not normalize it first.

## Reproduction

```bash
EP_HOST=localhost:4566
BUCKET=test-s3-bucket
KEY=uploads/hello.txt

aws --endpoint-url http://$EP_HOST s3 mb s3://$BUCKET
echo "hello" | aws --endpoint-url http://$EP_HOST s3 cp - s3://$BUCKET/$KEY

# force virtual-hosted addressing style (the AWS SDK default)
aws configure set default.s3.addressing_style virtual

URL=$(aws --endpoint-url http://$EP_HOST s3 presign s3://$BUCKET/$KEY --expires-in 300)
echo "$URL"
# https://test-s3-bucket.localhost:4566/uploads/hello.txt?X-Amz-...

curl -s "$URL"
# <Error><Code>SignatureDoesNotMatch</Code>...</Error>

# same object, path-style, for comparison
aws configure set default.s3.addressing_style path

URL2=$(aws --endpoint-url http://$EP_HOST s3 presign s3://$BUCKET/$KEY --expires-in 300)
echo "$URL2"
# http://localhost:4566/test-s3-bucket/uploads/hello.txt?X-Amz-...

curl -s "$URL2"
# 200, returns "hello"
```

Originally hit via AWS SDK for PHP 3.x (`use_path_style_endpoint` unset/false, i.e. virtual-hosted) through Laravel's `Storage::disk('s3')->temporaryUrl()`, then confirmed to be addressing-style-specific with the `aws-cli` script above.

## Environment

- Floci version / image tag: image `floci/floci:latest` (digest `sha256:4e451c39c7bb…`)
- Java SDK version (if applicable): n/a, reproduced with AWS SDK for PHP 3.x and aws-cli
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker, `FLOCI_SERVICES_S3_ENFORCE_AUTH=true`, region `eu-central-1`, credentials `test`/`test`

aniketshukla1:
## Summary

- preserve the original request URI before virtual-hosted S3 routing rewrites the path
- canonicalize SigV4 with the Host header and path that the client actually signed
- retain the existing behavior for path-style and HTTP/2 requests

Closes #3045

## Type of change

- [x] Bug fix

## AWS compatibility

Presigned GetObject URLs now verify in both virtual-hosted and path addressing styles when S3 auth enforcement is enabled.

## Testing

- S3AuthEnforcementIntegrationTest: 43 tests passed
- PreSignedUrlFilterTest: 9 tests passed
- S3VirtualHostFilterTest: 167 tests passed
- git diff --check passed

## Checklist

- [x] Focused tests pass locally with JDK 25
- [x] Virtual-hosted regression test added
- [x] Commit follows Conventional Commits

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/PreSignedUrlFilter.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/PreSignedUrlFilter.java
@@ -13,6 +13,7 @@ import org.jboss.logging.Logger;
 
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
+import java.net.URI;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
@@ -163,9 +164,10 @@ public class PreSignedUrlFilter implements ContainerRequestFilter {
             String credentialScope = date + "/" + region + "/" + service + "/aws4_request";
 
             // Build canonical headers from signed headers
-            String host = requestContext.getUriInfo().getRequestUri().getHost();
-            int port = requestContext.getUriInfo().getRequestUri().getPort();
-            String authority = (port > 0 && port != 80 && port != 443) ? host + ":" + port : host;
+            URI requestUri = requestContext.getProperty(S3VirtualHostFilter.ORIGINAL_REQUEST_URI_PROPERTY) instanceof URI uri
+                    ? uri
+                    : requestContext.getUriInfo().getRequestUri();
+            String authority = S3VirtualHostFilter.resolveHost(requestContext.getHeaderString("Host"), requestUri);
 
             StringBuilder canonicalHeaders = new StringBuilder();
             for (String header : signedHeaders.split(";")) {
@@ -178,7 +180,7 @@ public class PreSignedUrlFilter implements ContainerRequestFilter {
             }
 
             // Canonical request
-            String path = requestContext.getUriInfo().getRequestUri().getRawPath();
+            String path = requestUri.getRawPath();
             String canonicalQueryString = buildCanonicalQueryString(queryParams);
             String payloadHash = requestContext.getHeaderString("x-amz-content-sha256");
             if (payloadHash == null) {
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3VirtualHostFilter.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3VirtualHostFilter.java
@@ -25,6 +25,8 @@ import java.util.Set;
 @ApplicationScoped
 public class S3VirtualHostFilter implements ContainerRequestFilter {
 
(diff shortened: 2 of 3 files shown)


pgermosen (commented):
This looks right to me. Good catch on the root cause too, verifying against the pre-rewrite path/Host instead of the internally rewritten one is exactly what real SigV4 needs here, and reusing `resolveHost` keeps the HTTP/2 fallback consistent instead of duplicating it.

One small thing: the new test `virtualHostedPresignedRequestUsesOriginalHostAndPath` is `@Order(23)`, same as `presignedRequestWithMismatchedContentSha256IsRejected` right after it. Doesn't break anything since JUnit just leaves ties unspecified, but looks like it should've bumped the rest of the file by one. Not blocking, just worth tidying up. Will approve once that's sorted.

aniketshukla1:
Fixed in 0193aae3. The new virtual-hosted presigned test keeps order 23, and all following tests are renumbered through 43 so there are no duplicate order values. All 43 S3AuthEnforcementIntegrationTest cases pass, and git diff --check is clean.

pgermosen (approved):
Order fix landed clean, 24 through 43 with no gaps or dupes. Nothing else here — good to merge. The SWF failure on the last CI run is unrelated flakiness, filed as #3188.

PR #3058, opened 2026-09-04
fix(eventbridge): scope replay cancellation by ARN

nkootstra:
## Summary

Fix EventBridge replay cancellation when replay names repeat across accounts or regions. Cancellation state is now keyed by the replay ARN instead of only the replay name. Add regression coverage for dispatcher isolation and ARN forwarding through EventBridgeService.

  ## Type of change

  - [x] Bug fix (fix:)
  - [ ] New feature (feat:)
  - [ ] Docs / chore

  ## AWS Compatibility

Floci could cancel the wrong replay when two accounts or regions used the same replay name. AWS replay identity includes the account and region through the replay ARN. Cancellation now targets the exact replay ARN. Existing replay state and API responses are unchanged.

  ## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeService.java
@@ -1453,7 +1453,7 @@ public class EventBridgeService implements ResourceProvider {
             throw new AwsException("IllegalStatusException",
                     "Replay is not in a cancellable state: " + replay.getState(), 400);
         }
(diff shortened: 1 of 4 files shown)


pgermosen (approved):
Good catch, replay names aren't unique across accounts/regions any more than other resources here are, so keying cancelFlags by name alone was a real cross-tenant bug. ARN is the right key, and the value-checked remove is a nice bit of defensive concurrency hygiene. The regression test isolates exactly the scenario that was broken. Approving.

PR #3059, opened 2026-09-04
fix(stepfunctions): let Express executions reuse names and type the nested StartExecution refusal

guiscaranse:
## Summary

`StartExecution` enforced execution-name uniqueness on Express state machines. AWS does not: an Express name is immediately reusable and every start is its own concurrent execution. Uniqueness, and the 90-day reuse window, apply to Standard workflows only.

That also left a data-loss hazard. Same-name Express starts shared a single execution-store key, so a completing execution could overwrite another that was still running.

Changes:

- `StepFunctionsService.startExecution` gives an Express execution a per-start ARN, `express:<stateMachine>:<name>:<uuid>`, under the `express:` namespace `startSyncExecution` already uses, and skips the execution-store collision check. A reused name now starts a new independent execution instead of failing with `ExecutionAlreadyExists`. Standard behaviour is unchanged: idempotent same-input replay while `RUNNING`, `ExecutionAlreadyExists` otherwise.
- `AslExecutor.invokeNestedStateMachine` wraps the child `StartExecution` so a refusal becomes a typed, catchable error, `StepFunctions.ExecutionAlreadyExistsException`, rather than a `States.Runtime` that no `Catch` or `Retry` can match. This mirrors the `aws-sdk:sfn:startExecution` path, which keeps its `Sfn.` prefix and is unchanged.

Nothing else reads the execution ARN's shape, so the new Express form is safe: `listExecutions` filters on the `stateMachineArn` field rather than the ARN string, `parseVersionArn` only accepts an all-digits suffix (a UUID never is), and `taskEventProfile` parses task resource ARNs.

Tests: Express name reuse and a same-name no-clobber regression at the service level; nested-refusal typed error, `Catch`, `Retry`, and `Sfn.` versus `StepFunctions.` prefix cases for bare, `.sync` and `.sync:2`; a wire-level `StartExecution` integration test; and a `.sync`/`.sync:2` success guard.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Express `StartExecution` is non-idempotent and enforces no name uniqueness, and the `express:<stateMachine>:<name>:<id>` ARN shape matches the Step Functions resource-ARN definition. The optimized `states:startExecution` integration reports `StepFunctions.<Code>Exception` for a Standard-child collision, where `aws-sdk:sfn:startExecution` uses `Sfn.`. Left as documented deviations rather than changed here: the 90-day Standard name-reuse window is not modelled, and finished Express executions are retained rather than evicted.

## Checklist

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -261,6 +261,27 @@ with `States.QueryEvaluationError` where AWS answers a string, and `0/0` fails t
 it, because NaN is dropped as not-a-number without even a refusal to read it from. A state that
 fails is one a `Catch` fires on, which is the half of the divergence worth keeping.
 
+## Execution names
+
+A Standard execution name is unique per account, Region and state machine. Starting a Standard
+execution with a name and input that match one still running returns that original execution, so a
+retried call is idempotent; any other reuse of the name, whether a different input or a name whose
+execution has already closed, fails with `ExecutionAlreadyExists`.
+
+One deviation. AWS frees a Standard name 90 days after the execution closes; Floci keeps it taken for
+the life of the emulator, so a closed name never becomes reusable on its own.
+
+An Express execution name is not unique. Every `StartExecution` on an Express state machine is its
+own execution that runs alongside any others of the same name, and its ARN carries a per-start id
+after the name (`express:<stateMachine>:<name>:<id>`). Reusing an Express name never returns an
+earlier execution and never fails with `ExecutionAlreadyExists`.
+
+One more. AWS keeps no record of an Express execution once it ends, only its logs, and does not serve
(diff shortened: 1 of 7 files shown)


pgermosen (approved):
Solid fix, verified the core claim against the actual StartExecution docs: Express names really are immediately reusable and non-idempotent on AWS, and this also closes a real data-loss hazard where a completing execution could clobber another running one under the same name. Checked the ARN-shape safety claims against the current listExecutions/parseVersionArn code myself rather than just trusting the description, they hold up. The nested-refusal fix reuses the existing Sfn./StepFunctions. naming convention rather than inventing something new, and the concurrent same-name test is a nice touch. Approving.

PR #3060, opened 2026-09-04
fix(lambda): persist, validate and enforce event source mapping FilterCriteria

guiscaranse:
## Summary

`CreateEventSourceMapping` and `UpdateEventSourceMapping` accepted `FilterCriteria`, returned 200, then silently dropped it: `get-event-source-mapping` reported `null`, and every event was delivered as if no filter existed. `EventSourceMapping` had no such field, and the hand-maintained parse and response allowlists never mentioned it. This lands the persistence and validation first, then the poller enforcement, as two reviewable commits.

Changes:

- `FilterCriteria` persists as a typed nested `FilterCriteria`/`Filter` on `EventSourceMapping`, which Jackson round-trips via `EsmStore`, and is echoed back omit-when-unset.
- It is parsed and validated at create and update: well-formed JSON object patterns whose values are non-empty match arrays or nested objects, at most 5 filters, each pattern at most 4096 characters. An empty object, an empty `Filters` array, or `null` clears it. Validation happens at create and update time because the pollers drop and then checkpoint past (Kinesis, DynamoDB) or delete (SQS) any record a pattern fails to match, so an unvalidated pattern the matcher can never satisfy would silently discard every record.
- New `EsmFilterCriteriaUtils` wraps a stored `FilterCriteria` as the `{"FilterCriteria":{"Filters":[...]}}` node the existing `PipesFilterMatcher` consumes, and correlates matched records back to their sources by identity. `PipesFilterMatcher` itself is unchanged.
- Kinesis filters against the AWS dialect, the top-level base64-decoded `data` plus `partitionKey`, while the delivered event keeps `kinesis.data` base64. The shard checkpoint advances to the newest fetched sequence whenever a batch is disposed of, whether by a successful invoke or because a filter matched nothing, so filtered-out records are consumed rather than re-read forever. An invoke that errors or is throttled still leaves the checkpoint unmoved.
- DynamoDB Streams follows the same checkpoint discipline, where the per-record node (top-level `eventName` plus `dynamodb.*`) is both the delivery and the filter shape.
- SQS deletes non-matching messages immediately, since AWS consumes filtered SQS messages; a fully filtered batch returns without invoking, and all post-filter delete and return logic operates only on the matched subset.

Unfiltered mappings are unaffected: no filter nodes are built and the same list reference is passed through.

Tests: matching-only delivery, fully-filtered checkpoint advance and delete, invoke-error and throttle leaving the checkpoint unmoved with a same-window retry, a dialect pin asserting the `{"kinesis":{"data":...}}` envelope does not match, and non-JSON drop. Adds `KinesisEventSourcePollerTest` and extends the DynamoDB and SQS poller tests, plus validation and store round-trip coverage.

The final commit is a style-only pass removing em dashes per `AGENTS.md`; no code or string literals change.

Also documents `FilterCriteria` in `docs/services/lambda.md`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the documented `FilterCriteria` contract rather than a live account: the pattern grammar accepted at create and update time (`exists`, `anything-but`, prefix and numeric operators), the 5-filter and 4096-character limits, the clearing semantics of an empty object or empty `Filters`, and the dialects each source filters on. Kinesis filters top-level decoded `data`, not the invocation envelope; SQS consumes filtered messages rather than redelivering them. No request or response shapes change beyond `FilterCriteria` now being persisted and returned instead of dropped.

## Checklist

### DIFF
--- a/docs/services/lambda.md
+++ b/docs/services/lambda.md
@@ -802,14 +802,28 @@ of `{}` or with an empty `Filters` array clears any existing filters.
 
 !!! note "Supported operators"
     Filtering reuses Floci's shared EventBridge matcher (the same one EventBridge
-    Pipes uses). It supports exact match, `prefix`, `suffix`, `equals-ignore-case`,
-    `exists`, `anything-but` (string, array, including numeric arrays, or
-    `prefix`), and numeric operators. Known deviations from AWS, shared with Pipes:
-    boolean literals; event-value array intersection (matching when the record's
-    own field is itself an array and any element satisfies the pattern); and a
-    bare non-array numeric `anything-but` (wrap the number in an array,
-    `{"anything-but":[400]}`, to filter on it). Patterns using only the supported
-    operators behave as on AWS.
+    Pipes uses). It supports exact match on a string, number or `null`, plus
+    `prefix`, `suffix`, `equals-ignore-case`, `exists`, `anything-but` (a string,
+    a non-empty array of strings and numbers, or a nested `prefix`), and `numeric`
+    comparison/value pairs using `=`, `>`, `>=`, `<`, `<=`. Patterns using only
+    these behave as on AWS.
+
+    A pattern is validated at `CreateEventSourceMapping` and
+    `UpdateEventSourceMapping` and rejected with `InvalidParameterValueException`
+    when the matcher could not satisfy it: an unknown operator, an operator AWS
+    documents that Floci does not implement (`cidr`, `wildcard`), more than one
+    operator in a single match element, a boolean literal, a wrong operand type,
+    or a malformed `numeric` sequence such as an odd-length array or a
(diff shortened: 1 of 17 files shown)


pgermosen (approved):
Excellent work. Checked the checkpoint/delete behavior against AWS's actual docs on handling filtered records, the always-advance-on-disposal logic for Kinesis/DynamoDB and the immediate-delete for SQS both match exactly, including the filtering keys per source. The correlation logic in EsmFilterCriteriaUtils.selectMatched using reference identity rather than content equality is the right call, and the SQS poller's IdentityHashMap-backed filtered-message tracking is careful and correct. Nice honest disclosure of the known Pipes-matcher deviations in the docs too. Approving.

PR #3063, opened 2026-09-04
fix(dynamodb): route Kinesis CDC to the table owner's account and retry failed forwards

guiscaranse:
## Summary

Two ways DynamoDB change data capture lost records on its way to a Kinesis streaming destination, one per commit.

**Wrong account.** Forwarding resolved the target stream against the ambient account. Out of request scope, notably the TTL sweep which forwards a `REMOVE` per expired item after the request completes, the ambient account falls back to the default. CDC records could therefore be written into the default account's same-named stream, a cross-tenant leak, or dropped as `ResourceNotFound`. This was live on the TTL path.

**Dropped on transient failure.** Forwarding discarded the change event whenever `putRecord` threw, so a momentary Kinesis error silently lost `INSERT`, `MODIFY`, `REMOVE` and TTL events.

Changes:

- `KinesisService.putRecordForAccount` resolves, appends and persists against a specified account; a null account preserves the ambient behaviour. It keeps the per-stream append lock keyed on region and stream name rather than account, so explicit-account appends stay serialized with request-path producers and the append fix from #3000 still holds. It also migrates a legacy unprefixed stream into the account partition on write, with ownership validated by the stream's ARN account, so installs upgraded from pre-multi-account storage keep forwarding.
- `KinesisStreamingForwarder.forward` takes the table owner account, and `DynamoDbService` passes it at every forward site: request paths use the request account, and the TTL sweep uses the account it derives from the storage key.
- Forwarding becomes a bounded best-effort in-process retry queue. The write path only enqueues, never blocking and never throwing, so a slow or dead destination cannot stall a `PutItem` or the TTL sweep. A per-destination single-flight background drain performs the send outside the lock, and: transient or unknown failures retry with capped exponential backoff (250 ms doubling to an 8 s cap, up to 10 attempts) instead of dropping on the first exception; a deterministic terminal failure (`ValidationException`, `InvalidArgumentException`) drops only the poison record rather than wedging the queue behind it; per-destination FIFO is preserved across retries; the buffer is bounded at 1000 records per destination with the oldest evicted on overflow; and once the retry budget is exhausted the episode gives up, with a later event starting a fresh episode that self-heals when the stream recovers.
- Disabling a destination or deleting the table discards its buffered records and stops all future sends for it, though a send already in flight may still complete.
- Per-destination delivery health (forwarded, retried, dropped, overflowed, queue depth, last error type) is tracked and counted. The snapshot accessor is package-private: it is the observation seam the tests assert through, not a public API or an HTTP endpoint.

Tests: end-to-end concurrent CDC, per-destination FIFO across retries, poison-record drop without wedging, overflow eviction accounting, give-up and self-heal transitions, discard on disable and on table delete, and the TTL sweep forwarding under the owner account. The retry suite drives a manual scheduler step by step, so there are no real timers and no sleeps.

The final commit is a style-only pass removing em dashes per `AGENTS.md`; no code or string literals change.

Known limitations, stated rather than fixed here: buffered records are in memory only and are lost on restart, since this is an emulator rather than an at-least-once pipeline; and `DynamoDbStreamsService` has the same ambient-account exposure, which this does not address.

Added during review of this PR:

- `KinesisService.putRecordInternal` resolved the stream before taking the per-stream monitor. That resolve is not a read: for a legacy unprefixed stream it migrates the entry, so resolving outside the monitor let an ambient producer and the owner's producer migrate the same stream concurrently and fork it across account partitions, which is the leak this PR set out to close. The resolve and the record-size validation now happen inside the monitor, leaving one resolve rather than two.

- **This PR also changes a shared storage primitive, so it is worth a separate look.** `AccountAwareStorageBackend.get` migrates a pre-multi-account entry on read, and that migration is a write (a put then a delete) that did not hold the monitor the ownership-gated `getForAccountMigratingLegacy` holds. Two readers resolving the same legacy key from different account contexts could each copy it into their own partition, after which the two copies diverge independently. The migration branch now runs under that same monitor and re-checks the prefixed key inside it, so a reader that loses the race returns the winner's entry. The hit path stays outside the monitor, so only a miss pays for it. Fixing this at the primitive rather than call site by call site matters because the Kinesis batch `PutRecords` handler and roughly twenty read paths (`DescribeStream`, `GetRecords`, `GetShardIterator`, tag reads) resolve without any per-resource lock and were still exposed. A deterministic test pins it and fails without the change.

  Not addressed here: `get` still adopts an unprefixed entry into whatever account the request context resolves to, with no ownership check. It is generic over the stored type and has no notion of an owner, which is exactly why `getForAccountMigratingLegacy` takes that predicate from its caller. Gating it generically needs a per-resource-type predicate, so it is filed separately.

- Review also flagged that a write which resolved a table before `DeleteTable` can still forward one record afterwards. That predates this branch: `main` gates on the caller's `TableDefinition` snapshot in the same way and has no teardown hooks at all. A teardown marker consulted before enqueueing is not a sound fix, because the check has to be atomic with the enqueue under the destination monitor, so this is filed separately rather than grown into this PR.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

CDC records now land in the table owner's stream, matching AWS per-account isolation. On delivery, AWS retains change records durably and may reorder or duplicate them; this is deliberately weaker (in-process, non-durable) but stronger on ordering, since per-destination FIFO is preserved across retries. No request or response shapes change: the only externally visible difference is that a change event survives a transient Kinesis failure instead of vanishing, and lands in the correct account.

## Checklist

### DIFF
--- a/docs/services/dynamodb.md
+++ b/docs/services/dynamodb.md
@@ -165,3 +165,37 @@ aws dynamodb list-exports \
 ```
 
 The export writes to `s3://<bucket>/<prefix>/AWSDynamoDB/<exportId>/data/` as one or more `.json.gz` files, along with `manifest-summary.json` and `manifest-files.json`, the same layout as real AWS DynamoDB exports.
+
+## Kinesis change data capture (CDC)
+
+When a table has an **ACTIVE** Kinesis streaming destination (see
+`EnableKinesisStreamingDestination`), every item change — `INSERT`, `MODIFY`, and `REMOVE`,
+including TTL expirations — is forwarded to the destination stream as a Kinesis record in the
+AWS CDC envelope (`eventName`, `dynamodb.Keys`, `NewImage`/`OldImage`, `ApproximateCreationDateTime`).
+
+### Delivery contract
+
+Forwarding is **bounded best-effort with in-process retry**. A write is never blocked or failed by
+the destination stream: the change event is enqueued and delivered on a background drain, so a slow or
+unavailable Kinesis stream cannot stall a `PutItem`/`UpdateItem`/`DeleteItem` or the TTL sweep.
+
+The drain gives these guarantees per destination:
+
+- **Retries** transient/unknown send failures with capped exponential backoff (250 ms doubling to an
+  8 s cap, up to 10 attempts) rather than dropping the record on the first exception.
+- **Preserves FIFO** across retries — the head record is not skipped past (stronger than AWS, which may
+  reorder or duplicate).
+- **Drops a poison record** immediately on a deterministic terminal failure
+  (`ValidationException`/`InvalidArgumentException`) so it cannot wedge the queue behind it.
+- **Gives up an episode** after the retry budget is exhausted, dropping the buffered records and marking
+  the destination `GAVE_UP`; a later change event starts a fresh episode and self-heals once the stream
+  recovers.
+- **Bounds memory** at 1000 buffered records per destination, evicting the oldest on overflow.
+
+Disabling a destination or deleting the table discards that destination's buffered records and stops all
+future sends for it; a send already in flight at that instant may still complete (teardown cannot recall
+an in-flight request — an inherent, harmless race). Buffered records are held **in memory only** — they
+are not durable and are lost on restart (this is an emulator, not an at-least-once pipeline). Records that are permanently dropped (terminal, give-up,
+or overflow) are counted and logged, and per-destination delivery health (forwarded/retried/dropped
+counts, queue depth, last error, and current health) is tracked for inspection.
+```
--- a/docs/services/dynamodb.md
+++ b/docs/services/dynamodb.md
@@ -169,8 +169,8 @@ The export writes to `s3://<bucket>/<prefix>/AWSDynamoDB/<exportId>/data/` as on
 ## Kinesis change data capture (CDC)
 
 When a table has an **ACTIVE** Kinesis streaming destination (see
-`EnableKinesisStreamingDestination`), every item change — `INSERT`, `MODIFY`, and `REMOVE`,
-including TTL expirations — is forwarded to the destination stream as a Kinesis record in the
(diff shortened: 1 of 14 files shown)


pgermosen (commented):
Really solid work, especially the self-review notes catching two more races beyond the original scope, the putRecordInternal lock ordering and the AccountAwareStorageBackend.get migration race affecting the other ~20 read paths. Fixing that one at the shared primitive instead of call by call was the right call.

Heads up on the conflict though: it's not just textual noise. This PR and #3069 both touched the same lock-and-resolve section of putRecordInternal (now merged into main). #3069 added explicitHashKey validation using a stream resolved before the lock; this PR moved resolution inside the lock because resolving a legacy unprefixed stream is actually a write (it migrates the entry), so doing it outside the lock let two producers migrate the same stream into different account partitions concurrently.

Whoever rebases this needs to make sure the explicit-hash-key validation and the account-aware resolution both end up running against the same lock-protected resolve, rather than mechanically layering one diff on top of the other. If the rebase keeps #3069's resolve-before-lock shape and just bolts the account parameter on, it'll quietly reintroduce the exact race this PR fixed. Worth a careful look rather than letting git's auto-merge (or a quick rebase) paper over it.

guiscaranse:
@pgermosen Thanks for flagging this one. You were right that it was not textual noise: git offered both shapes, main's side looked perfectly clean, and taking it would have restored the resolve-before-lock and undone the fix. Rebased manually onto f677edad.

pgermosen (approved):
Confirmed the merge-conflict fix survived the rebase: putRecordInternal now resolves under the stream lock for both the ambient and explicit-account paths since they're the same method now, which closes the legacy-migration race greptile flagged. The teardown-race thread turned out to be a genuinely disclosed limitation rather than a silent gap, it's spelled out in the docs as an inherent, harmless race in the best-effort delivery model. Approving.

PR #3066, opened 2026-09-04
fix(wafv2): validate resource names on operations

nkootstra:
## Summary

Validate the required WAFv2 resource `Name` on get, update, and delete operations. Requests with a missing name now return `WAFInvalidParameterException`. Requests with a mismatched name return `WAFNonexistentItemException` without changing or deleting the resource.

## Type of change

  - [x] Bug fix (`fix:`)
  - [ ] New feature (`feat:`)
  - [ ] Docs / chore

  ## AWS Compatibility

- Floci previously used only `Scope` and `Id` to identify WAFv2 resources. A valid ID with an incorrect name could therefore read, update, or delete the resource.
- Floci now validates `Name` for Web ACLs, IP sets, regex pattern sets, and rule groups. CloudFormation callers provide the stored resource name during lifecycle operations.

Valid AWS SDK requests remain unchanged. Invalid requests now match the expected AWS error behavior.

## Checklist


pgermosen (approved):
Good catch, confirmed against the actual DeleteWebACL reference that Id, Name and Scope are all required together to identify a WAFv2 resource, so ignoring Name for lookups was a real gap, not just cosmetic. The fix is applied consistently across all four resource types through one shared helper, and the tests verifying a rejected update/delete actually leaves the resource untouched is exactly the right check. Approving.

PR #3068, opened 2026-09-04
feat(kms): support ImportKeyMaterial for Origin=EXTERNAL keys

### LINKED ISSUE #1916: [FEAT] KMS ImportKeyMaterial support
## Service

AWS KMS (Key Management Service) service.

## API Action / Feature

KMS `ImportKeyMaterial` API Action and dependant `GetParametersForImport` API Action should be implemented. `CreateKey` action should also be updated to support `EXTERNAL` type keys.

## AWS Documentation

[https://docs.aws.amazon.com/es_es/kms/latest/APIReference/API_ImportKeyMaterial.html](https://docs.aws.amazon.com/es_es/kms/latest/APIReference/API_ImportKeyMaterial.html)

## Why is this needed?

I think that adding support for these actions its interesting if we are in scenarios where our keys are externally managed and we only want to import the key material into KMS.

- [x] Yes
- [ ] No

omatheusmesmo:
## Summary

Closes #1916.

Adds the import-key-material (BYOK) flow to KMS. `CreateKey` accepts `Origin=EXTERNAL`, which
creates a key with no key material in state `PendingImport`. `GetParametersForImport` returns a
real RSA wrapping public key and an import token; `ImportKeyMaterial` unwraps material that a
standard client wrapped with that public key, validates it, and moves the key to `Enabled`.
`DeleteImportedKeyMaterial` and `ValidTo` expiry take it back to `PendingImport`.

The unwrap is real, not structural: material wrapped against a different key, or with a different
algorithm than the one requested at `GetParametersForImport`, fails with
`InvalidCiphertextException`. An HMAC key imported this way generates MACs with the material that
was actually imported.

New actions:

| Action | Behavior |
|---|---|
| `GetParametersForImport` | Generates an RSA wrapping key pair (`RSA_2048`/`RSA_3072`/`RSA_4096`), stores the private half on the key, returns `PublicKey`, `ImportToken` and `ParametersValidTo` (+24h) |
| `ImportKeyMaterial` | Validates the token, unwraps, validates material length, installs it, returns `KeyId` and `KeyMaterialId` |
| `DeleteImportedKeyMaterial` | Drops the material, returns `KeyId` and `KeyMaterialId` |

Changed actions:

- `CreateKey` reads `Origin`. `AWS_CLOUDHSM` and `EXTERNAL_KEY_STORE` are rejected with
  `UnsupportedOperationException`; an unmodelled value gets `ValidationException`.
- `DescribeKey`/`CreateKey` metadata: `Origin` was hardcoded to `AWS_KMS`
  (`KmsJsonHandler.java:608` on `main`) and is now per-key, joined by `ExpirationModel` and
  `ValidTo` when the key has imported material.
- `EnableKey`, `DisableKey`, the crypto operations and the MAC operations reject a key in
  `PendingImport` with `KMSInvalidStateException`. `validateMacOperationKey` did not check key
  state at all, so without this guard a `PendingImport` HMAC key reached
  `Base64.decode(null)` and surfaced as a 500.
- `EnableKeyRotation`/`DisableKeyRotation` reject `Origin=EXTERNAL`: KMS does not own imported
  material and cannot rotate it.

Structure: `KmsKeyImport` holds the wrapping cryptography as stateless functions, `KmsService`
keeps the state machine and the store. `KmsKeySpec.materialByteLength()` becomes the single source
for the raw-material length of `SYMMETRIC_DEFAULT` and the HMAC specs, which `hmacKeyByteLength`
now delegates to.

Expiry is evaluated when the key is next read rather than on a timer. Nothing can observe a key
without going through `resolveKey`, so the difference is not visible through the API and no
scheduler is introduced.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The wire protocol is exercised with **AWS SDK for Java v2 2.52.0** in
`compatibility-tests/sdk-test-java` (`KmsFeaturesTest#importedKeyMaterialMakesAnExternalKeyUsable`),
which drives the whole round trip through the SDK: `createKey(origin=EXTERNAL)` →
`getParametersForImport` → RSA-OAEP-SHA-256 wrapping performed by the test with the returned public
key → `importKeyMaterial` → `encrypt`/`decrypt` → `deleteImportedKeyMaterial`, asserting
`KeyState`, `Enabled` and `ExpirationModel` off `DescribeKey` at each step. `KmsIntegrationTest`
covers the same path over raw JSON 1.1.

Behavior and error codes come from the AWS KMS API Reference, **not** from measurement against a
live account. The statements it rests on, so a reviewer can check them directly:

- `ImportKeyMaterial`: `ExpirationModel` defaults to `KEY_MATERIAL_EXPIRES`; `ValidTo` is required
  for it and must be omitted for `KEY_MATERIAL_DOES_NOT_EXPIRE`; `ValidTo` "must be a future date
  and time. The maximum value is 365 days from the request date"; `InvalidCiphertextException`
  means "AWS KMS could not decrypt the encrypted (wrapped) key material";
  `InvalidImportTokenException` means the token "is invalid or is associated with a different KMS
  key"; the response carries `KeyMaterialId`.
- `GetParametersForImport`: "Each public key and import token set is valid for 24 hours"; the two
  "are permanently linked and must be used together"; `RSAES_PKCS1_V1_5` is marked deprecated,
  "As of October 10, 2023, AWS KMS does not support the RSAES_PKCS1_V1_5 wrapping algorithm", so
  it is rejected here rather than honoured.
- `DeleteImportedKeyMaterial`: "When the specified KMS key is in the `PendingDeletion` state, this
  operation does not change the KMS key's state. Otherwise, it changes the KMS key's state to
  `PendingImport`."
- `KeyMaterialId` is documented as derived "based on the KMS key ID and the imported key material"
  with a 64-character `^[a-f0-9]+$` shape, so it is computed here as
  `SHA-256(keyId || material)`. Keeping it after the material is deleted is what lets a later
  re-import reject different material with `IncorrectKeyMaterialException`, without holding a
  second copy of the secret.

Wrapping uses `RSA/ECB/OAEPPadding` with an explicit `OAEPParameterSpec`, not
`OAEPWithSHA-256AndMGF1Padding`, because several providers keep SHA-1 in the MGF1 of the latter and
would not interoperate with a standard client. This matches the approach #3038 took for RSAES-OAEP
in `Encrypt`/`Decrypt`.

Two deviations, both documented in `docs/services/kms.md`:

1. `Origin=EXTERNAL` is supported for `SYMMETRIC_DEFAULT` and the `HMAC_*` specs, whose material is
   a raw byte string. Real KMS also imports asymmetric material as a DER key pair; an asymmetric
   spec with `Origin=EXTERNAL` is rejected at `CreateKey` rather than creating a key that could
   never sign or decrypt. Worth noting the API Reference is inconsistent here: the `Origin`
   parameter says "The `EXTERNAL` origin value is valid only for symmetric KMS keys" while the
   narrative section of the same page lists asymmetric keys as importable.
2. Holding several imported materials on one symmetric key, which real KMS uses for on-demand
   rotation of imported material, is not emulated. `ImportType=NEW_KEY_MATERIAL` on a key that
   already has material is rejected with `UnsupportedOperationException` instead of being silently
   treated as a re-import, and `ListKeyRotations` is not implemented.

## Checklist

On the first box: 682 tests pass across the areas this touches (`KmsServiceTest`,
`KmsIntegrationTest`, `LeafCfnProvisionerTest`, `RdsServiceTest`, `ElastiCacheServiceTest`,
`DocDbServiceTest`), 29 of them new in `KmsServiceTest$ImportedKeyMaterial` plus 5 new cases in
`KmsIntegrationTest`. The full suite does not finish on my machine: `ContainerPlatformDockerIntegrationTest`
fails on a `linux/arm64` image pull and the Docker-backed `ApiGateway*IntegrationTest` classes time
out with `SocketTimeoutException`. Those are environmental and touch nothing in this change, but I
would rather flag it than tick a box I cannot stand behind. `make docs-check` is green.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/KmsFeaturesTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/KmsFeaturesTest.java
@@ -3,6 +3,7 @@ package com.floci.test;
 import org.junit.jupiter.api.*;
 import software.amazon.awssdk.core.SdkBytes;
 import software.amazon.awssdk.services.kms.KmsClient;
+import software.amazon.awssdk.services.kms.model.AlgorithmSpec;
 import software.amazon.awssdk.services.kms.model.CreateKeyResponse;
 import software.amazon.awssdk.services.kms.model.DisabledException;
 import software.amazon.awssdk.services.kms.model.EncryptionAlgorithmSpec;
@@ -12,9 +13,21 @@ import software.amazon.awssdk.services.kms.model.InvalidKeyUsageException;
 import software.amazon.awssdk.services.kms.model.KeySpec;
 import software.amazon.awssdk.services.kms.model.KeyUsageType;
 import software.amazon.awssdk.services.kms.model.KmsInvalidStateException;
+import software.amazon.awssdk.services.kms.model.ExpirationModelType;
+import software.amazon.awssdk.services.kms.model.KeyState;
 import software.amazon.awssdk.services.kms.model.ListResourceTagsResponse;
+import software.amazon.awssdk.services.kms.model.OriginType;
+import software.amazon.awssdk.services.kms.model.WrappingKeySpec;
 
+import javax.crypto.Cipher;
+import javax.crypto.spec.OAEPParameterSpec;
+import javax.crypto.spec.PSource;
 import java.nio.charset.StandardCharsets;
+import java.security.KeyFactory;
+import java.security.PublicKey;
+import java.security.spec.MGF1ParameterSpec;
+import java.security.spec.X509EncodedKeySpec;
+import java.util.Arrays;
 import java.util.Map;
 
 import static org.assertj.core.api.Assertions.*;
@@ -28,8 +41,9 @@ import static org.assertj.core.api.Assertions.*;
  *   #1844 — Decrypt enforces KeyId against the wrapping CMK (IncorrectKeyException)
  *   #1844 — ReEncrypt enforces SourceKeyId against the wrapping CMK (IncorrectKeyException)
  *   #3024: Encrypt/Decrypt apply real RSAES-OAEP for RSA keys
+ *   #1916: ImportKeyMaterial round trip on an Origin=EXTERNAL key
  */
-@DisplayName("KMS features (#258 #259 #269 #1844 #3024)")
+@DisplayName("KMS features (#258 #259 #269 #1844 #1916 #3024)")
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 class KmsFeaturesTest {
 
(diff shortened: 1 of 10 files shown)


omatheusmesmo:
Notes on the two automated findings, for reviewers.

**"Cancellation Enables Materialless Key" is real and is fixed.** `cancelKeyDeletion` set
`KeyState=Enabled` unconditionally, so an `Origin=EXTERNAL` key with no key material came back
usable after `CancelKeyDeletion`: symmetric `Encrypt` would accept data under a key that cannot
serve it, and an HMAC key reached `Base64.decode(null)`. This is the same class of problem as the
`validateMacOperationKey` guard already in the PR, a caller that predates the `PendingImport`
state and assumed every key holds material. `cancelKeyDeletion` now returns such a key to
`PendingImport` instead. Covered by `cancellingDeletionLeavesAKeyWithNoMaterialInPendingImport`,
`cancellingDeletionAfterTheMaterialWasDeletedKeepsTheKeyUnusable` and
`cancellingDeletionRestoresAKeyThatStillHasItsMaterial`.

**"Missing Material Check" I did not apply, because it would diverge from AWS.** The suggestion is
to reject `DeleteImportedKeyMaterial` on a key in `PendingImport` with `KMSInvalidStateException`.
The [key state table](https://docs.aws.amazon.com/kms/latest/developerguide/key-state.html) gives
that cell a plain success marker with no footnote:

| Operation | `Pending import` column |
|---|---|
| `EnableKey` | fails, footnote [5]: `KMSInvalidStateException: <key ARN> is pending import because no key material has ever been imported or one of the imported key materials is deleted or expired.` |
| `DisableKey` | fails, footnote [5] |
| `DeleteImportedKeyMaterial` | succeeds, no footnote |

So AWS models exactly the exception the finding asks for, applies it to `EnableKey` and
`DisableKey` in that column, and does not apply it to this operation. The footnote [9] on the
other three columns of the `DeleteImportedKeyMaterial` row is `UnsupportedOperationException` for
a key that cannot have imported key material, which is the `Origin != EXTERNAL` case that
`requireExternalOrigin` already covers.

`deletingMaterialFromAKeyThatHasNoneSucceeds` pins the documented behavior so the guard does not
get added later by mistake.

pgermosen (approved):
Exceptionally well-grounded PR. Went through essentially every cited AWS behavior claim against the actual API reference (ImportKeyMaterial, GetParametersForImport, CreateKey) and they all check out precisely, several word for word. The documented inconsistency in AWS's own CreateKey page about EXTERNAL origin and asymmetric keys is real, confirmed it myself, and treating it conservatively rather than guessing which contradictory AWS statement to trust was the right call.

On the one still-open greptile thread (DeleteImportedKeyMaterial missing a material check): checked AWS's key state compatibility table directly, and it shows DeleteImportedKeyMaterial succeeding unconditionally on a PendingImport key, so the current behavior here is actually correct and greptile's suggested guard would make it less compatible, not more. Not something to act on.

Approving.

PR #3069, opened 2026-09-04
fix: honor Kinesis ExplicitHashKey routing

### LINKED ISSUE #3065: [FEAT] kinesis: PutRecord/PutRecords should route records using ExplicitHashKey
## Service

Kinesis

## API Action / Feature

`PutRecord` and `PutRecords`: support `ExplicitHashKey` when selecting the destination shard.

When `ExplicitHashKey` is present, it should override the hash derived from `PartitionKey`.

## AWS Documentation

- [`PutRecords`](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecords.html)
- [`PutRecordsRequestEntry.ExplicitHashKey`](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecordsRequestEntry.html)
- [`PutRecord`](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html)

## Why is this needed?

AWS specifies that `ExplicitHashKey` overrides the hash derived from `PartitionKey` and determines which shard receives a record.

Applications use this behavior to route records deterministically to particular shards—for example, to align application partitions with Kinesis shards or control record distribution during testing.

FloCI currently ignores `ExplicitHashKey` and selects a shard using only `PartitionKey`. This differs from AWS behavior and prevents clients from testing explicit shard routing accurately.

Expected behavior:

1. If `ExplicitHashKey` is present, validate it as a decimal value in the Kinesis 128-bit hash-key space.
2. Select the open shard whose `HashKeyRange` contains that value.
3. Use the partition-key hash only when `ExplicitHashKey` is absent.
4. Return the selected shard ID in the `PutRecord` or `PutRecords` result.

- [ ] Yes
- [X] No

ousamabenyounes:
## Summary

Closes #3065.

- Route `PutRecord` and `PutRecords` entries with `ExplicitHashKey` to the open shard whose `HashKeyRange` contains the explicit value.
- Validate explicit hash keys as decimal Kinesis hash-space values before writing records.
- Split newly created streams across the 128-bit Kinesis hash-key space so shard ranges are usable for explicit routing.
- Document the record-routing behavior in the Kinesis service page.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Floci previously ignored `ExplicitHashKey` and selected the destination shard from `PartitionKey` only. AWS documents `ExplicitHashKey` on `PutRecord` and `PutRecordsRequestEntry` as the value used to determine the target shard, so this patch makes that value override partition-key routing when present.

Invalid explicit keys now fail as `InvalidArgumentException` before records are written.

## Test verification (RED -> GREEN)

RED, before production changes:

- `./mvnw test -B -Dtest=KinesisIntegrationTest,KinesisJsonHandlerTest`
- Result: expected failure, `115` tests run, `5` failures.
- Failures proved the issue: multi-shard ranges were full-span, explicit max hash routed to shard `000000000000`, and invalid explicit hash keys were accepted.

GREEN, after production changes:

- `./mvnw test -B -Dtest=KinesisIntegrationTest,KinesisJsonHandlerTest`
- Result: `115` tests run, `0` failures, `0` errors, `0` skipped.

Revert proof:

- Temporarily reverted only the production changes and reran `./mvnw test -B -Dtest=KinesisIntegrationTest,KinesisJsonHandlerTest`.
- Result: expected failure returned, `115` tests run, `5` failures.
- Reapplied the production changes and reran the focused suite successfully.

Full local suite proof:

- Attempted `./mvnw test -B` on upstream `main`; the local JVM was killed with exit code `137` before running tests.
- Attempted a local sharded replay matching the CI shard structure via ignored `run-ci.sh`; upstream `main` reached Docker-backed API Gateway WebSocket/Lambda tests and failed because Lambda runtime containers timed out calling back into the host.
- Those failures were present before this branch and are unrelated to the Kinesis files changed here. The final focused Kinesis suite and docs gate pass locally.

Additional checks:

- `make docs-check`
- `git diff --check HEAD~1 HEAD`

## Checklist

- [ ] `./mvnw test` passes locally, blocked by pre-existing local resource and Docker callback failures described above

---

## Review follow-up (commit `5ddcdf9e`)

Addressed the two Greptile review findings:

- **Persisted ranges remained overlapping (P1).** Multi-shard streams created before disjoint ranges existed carried the full 128-bit span on every open shard, so an explicit hash key matched all shards and `findFirst()` collapsed routing onto `shardId-...000`. `KinesisService.normalizeOpenShardRanges` now repartitions such legacy open shards into disjoint contiguous ranges (ordered by shard id) before explicit-hash selection, inside the per-stream lock; already-disjoint (new/split/merged) streams and single-open-shard streams are left untouched.
- **Numeric hash keys were accepted (P2).** `ExplicitHashKey` was read with `asText(null)`, coercing a JSON number into an accepted decimal string. `KinesisJsonHandler.readExplicitHashKey` now rejects a non-string node with `SerializationException` for both `PutRecord` and `PutRecords`, while an absent/null field still selects by partition key.

### Test verification (RED -> GREEN)

RED, on the prior branch HEAD (three new tests fail):

```
[ERROR] KinesisJsonHandlerTest.putRecordMigratesLegacyOverlappingShardRangesForExplicitHashKey
        expected: <shardId-000000000001> but was: <shardId-000000000000>
[ERROR] KinesisJsonHandlerTest.putRecordRejectsNumericExplicitHashKey
        Expected AwsException to be thrown, but nothing was thrown.
[ERROR] KinesisJsonHandlerTest.putRecordsRejectNumericExplicitHashKey
        Expected AwsException to be thrown, but nothing was thrown.
[ERROR] Tests run: 118, Failures: 3
```

GREEN, after the fix:

```
[INFO] Tests run: 118, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Run on Temurin 25 via `mvnw -Dtest=KinesisJsonHandlerTest,KinesisIntegrationTest test`, plus a whole-project `test-compile`. All GitHub CI checks (4 test shards, native amd64/arm64 images, the SDK/IaC compatibility matrix, Greptile) are green on `5ddcdf9e`.

### DIFF
--- a/docs/services/kinesis.md
+++ b/docs/services/kinesis.md
@@ -66,7 +66,11 @@ aws kinesis describe-stream --stream-arn arn:aws:kinesis:us-east-1:000000000000:
 
 `GetShardIterator` supports all five iterator types: `TRIM_HORIZON`, `LATEST`, `AT_SEQUENCE_NUMBER`, `AFTER_SEQUENCE_NUMBER`, `AT_TIMESTAMP`.
 
-A `LATEST` iterator is positioned at the shard tip at the moment the iterator is created, matching AWS: records written after the iterator was obtained are returned, records written before are not. This supports the standard tailing pattern — obtain a `LATEST` iterator, trigger the action that produces the record, then poll `GetRecords` following `NextShardIterator`.
+A `LATEST` iterator is positioned at the shard tip at the moment the iterator is created, matching AWS: records written after the iterator was obtained are returned, records written before are not. This supports the standard tailing pattern: obtain a `LATEST` iterator, trigger the action that produces the record, then poll `GetRecords` following `NextShardIterator`.
+
+## Record Routing
+
+`PutRecord` and `PutRecords` honor `ExplicitHashKey` when it is provided. The value must be a decimal integer in the Kinesis hash-key space, and records are written to the open shard whose `HashKeyRange` contains that value. Without `ExplicitHashKey`, Floci keeps using the partition key to choose a shard.
 
 ## Configuration
 
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -478,8 +478,10 @@ public class KinesisJsonHandler {
             throw new AwsException("SerializationException", "Data is not valid base64.", 400);
         }
         String partitionKey = request.path("PartitionKey").asText();
+        String explicitHashKey = request.path("ExplicitHashKey").asText(null);
 
-        KinesisService.PutRecordResult result = service.putRecordWithShardId(streamName, data, partitionKey, region);
+        KinesisService.PutRecordResult result = service.putRecordWithShardId(
+                streamName, data, partitionKey, explicitHashKey, region);
 
         ObjectNode response = objectMapper.createObjectNode();
         response.put("SequenceNumber", result.sequenceNumber());
@@ -501,7 +503,7 @@ public class KinesisJsonHandler {
         // check. The count cap is checked upfront and the byte cap as the
         // loop goes, so neither an over-long batch nor an oversized one is
         // fully decoded before it is rejected.
-        record Entry(JsonNode node, byte[] data) {}
+        record Entry(JsonNode node, byte[] data, String explicitHashKey) {}
         List<Entry> entries = new ArrayList<>();
         long totalBytes = 0;
         for (JsonNode node : recordsNode) {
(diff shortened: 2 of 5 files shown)


ousamabenyounes:
Addressed both review findings in `5ddcdf9e`:

- **Persisted ranges remained overlapping (P1):** legacy multi-shard streams whose open shards still carry the old full-span range are now repartitioned into disjoint contiguous ranges before explicit-hash routing (`normalizeOpenShardRanges`), so an explicit key no longer collapses onto the first shard after an upgrade. Already-disjoint (new/split/merged) and single-open-shard streams are untouched.
- **Numeric hash keys were accepted (P2):** `ExplicitHashKey` is now rejected with `SerializationException` when it is a non-string JSON node, for both `PutRecord` and `PutRecords`; an omitted field still routes by partition key.

Validation: reproduced both as failing tests first (explicit MAX key routed to `shardId-...000`; numeric key accepted) then green after the fix — Kinesis suite 118 run, 0 failures on Temurin 25. All PR checks are green on `5ddcdf9e`. Both review conversations resolved.

pgermosen (approved):
Solid fix, verified ExplicitHashKey's type, override semantics, and pattern against the actual PutRecord API reference and the implementation matches exactly, including the leading-zero and range checks. Both greptile findings got real fixes rather than just acknowledgment, the legacy-range migration correctly leaves already-disjoint streams alone, and the numeric-key rejection closes a real wire-compatibility gap. Good RED/GREEN evidence throughout. Approving.

PR #3070, opened 2026-09-04
fix(lambda): deliver extension INVOKE when the runtime receives the invocation

### LINKED ISSUE #2573: [BUG] Internal Lambda extension registers successfully but never receives INVOKE event
## Service

Lambda, specifically the Lambda Extensions API.

## AWS API Action

- `POST /2020-01-01/extension/register`
- `GET /2020-01-01/extension/event/next`
- Lambda invocation through API Gateway v1

## Expected behavior

An internal Lambda extension registers during Lambda initialization with:
```
POST /2020-01-01/extension/register
Lambda-Extension-Name: ScopedFlushTelemetry
Content-Type: application/json
 
{"events":["INVOKE"]}
```
Floci returns a successful response containing a valid Lambda-Extension-Identifier.

The extension then immediately starts the documented long poll:
```
GET /2020-01-01/extension/event/next
Lambda-Extension-Identifier: <identifier>
```
When the Lambda is invoked, the pending /extension/event/next request should return an INVOKE event payload for that invocation.

The extension should then be able to process the event and issue its next /extension/event/next request.

AWS Lambda supports internal extensions running as a separate thread in the runtime process and allows extensions to register for invocation lifecycle events:

https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html

## Actual behavior

Floci accepts the registration request and returns a valid Lambda-Extension-Identifier.

The extension immediately calls `GET /2020-01-01/extension/event/next`, but the request remains blocked when the Lambda function is invoked.

The Lambda handler executes and completes, but the registered internal extension does not receive the corresponding INVOKE event.
Observed sequence:
```
14:39:43.299 Extension registration succeeds
14:39:43.307 GET /2020-01-01/extension/event/next begins
14:39:45.719 Lambda invocation completes
14:39:53.314 Event request is still blocked
```
The event request was stopped after 10 seconds by a diagnostic client-side timeout. Without that diagnostic timeout, it remains blocked.

## Reproduction
Run the following code during initialization of a managed-runtime Lambda, such as a .NET Lambda.
```
using System.Text;

var runtimeApi =
    Environment.GetEnvironmentVariable(
        "AWS_LAMBDA_RUNTIME_API")
    ?? throw new InvalidOperationException(
        "AWS_LAMBDA_RUNTIME_API is not set.");

var runtimeApiUri = new Uri($"http://{runtimeApi}");

using var registrationClient = new HttpClient
{
    BaseAddress = runtimeApiUri
};

registrationClient.DefaultRequestHeaders.Add(
    "Lambda-Extension-Name",
    "TestInternalExtension");

using var registrationBody = new StringContent(
    """
    {"events":["INVOKE"]}
    """,
    Encoding.UTF8,
    "application/json");

using var registrationResponse =
    await registrationClient.PostAsync(
        "/2020-01-01/extension/register",
        registrationBody);

registrationResponse.EnsureSuccessStatusCode();

var extensionId =
    registrationResponse.Headers
        .GetValues("Lambda-Extension-Identifier")
        .Single();

Console.WriteLine(
    $"Extension registered: {extensionId}");

_ = Task.Run(
    async () =>
    {
        using var eventClient = new HttpClient
        {
            BaseAddress = runtimeApiUri,
            Timeout = Timeout.InfiniteTimeSpan
        };

        eventClient.DefaultRequestHeaders.Add(
            "Lambda-Extension-Identifier",
            extensionId);

        Console.WriteLine(
            "Waiting for next extension event");

        using var eventResponse =
            await eventClient.GetAsync(
                "/2020-01-01/extension/event/next");

        var eventPayload =
            await eventResponse.Content.ReadAsStringAsync();

        Console.WriteLine(
            $"Extension event received: {eventPayload}");
    });
```
After the extension begins waiting on /extension/event/next, invoke the Lambda function:

```bash
aws --endpoint-url http://localhost:4566 \
  --region eu-west-1 \
  lambda invoke \
  --function-name test-function \
  response.json
```
Expected output from the extension:
```
Extension event received: {"eventType":"INVOKE", ...}
```
For a fully reliable reproduction, store the returned background Task in a static field so it remains rooted for the lifetime of the execution environment:
```
private static Task? _extensionTask;
```
Then:
```
_extensionTask = Task.Run(...);
```

Actual output:
```
Extension registered: <identifier>
Waiting for next extension event
```

## Environment

Floci version / image tag: floci/floci:1.7.0
Lambda runtime: .NET 10
Lambda base image: public.ecr.aws/lambda/dotnet:10
Lambda deployment type: ZIP archive
Java SDK version: N/A
How Floci is running: Docker through Testcontainers for .NET
Host operating system: Windows with Docker Desktop
Docker Server version: 29.7.2
Docker API version: 1.55
Docker kernel: 6.12.76-linuxkit
Docker network: default bridge
Lambda container mode:
Reproduced with the default warm pool
Reproduced with FLOCI_SERVICES_LAMBDA_EPHEMERAL=true

## Additional context
This appears specific to an internal extension that registers from within the managed runtime process.

External extension support was introduced in:

https://github.com/floci-io/floci/issues/1771
https://github.com/floci-io/floci/pull/1773

A possible cause is that invocation-event dispatch is coordinated using extensions discovered under /opt/extensions, while an internal extension registering from the runtime process is not included in that invocation's event dispatch.

This is only a hypothesis. The externally observable problem is that registration succeeds, but the pending /extension/event/next request does not receive the Lambda invocation's INVOKE event.

ammaraltahan:
## Summary

Delivers the Lambda Extensions `INVOKE` event when the runtime actually receives the invocation (`GET /runtime/invocation/next`), instead of at `enqueue()` time. Closes #2573.

Previously, `enqueue()` built and dispatched the `INVOKE` event inline, to whatever was in `extensions` at that exact moment, with nothing reconsidering it afterward. An internal extension (no `/opt/extensions` filesystem footprint, so nothing waits for its registration before `enqueue()` runs) commonly registers concurrently with or after that call and misses the event entirely, with no way to recover it. This moves the fan-out into `sendInvocation()`, gated on the runtime's response write succeeding (so a requeue-on-failure never emits a duplicate `INVOKE` for the same request), matching the Init→Invoke transition point real AWS uses for delivery.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Bug fix, not a new action. Incorrect behavior: an extension that registers from inside the runtime process (e.g. aws-lambda-web-adapter) after the caller has already called `enqueue()` for the first invocation never received that invocation's `INVOKE` event: the pre-fix code checked `extensions` exactly once, synchronously, inside `enqueue()`. Real AWS delivers `INVOKE` once the runtime signals readiness for the next invocation via `/runtime/invocation/next`, relying on internal extensions having registered before that signal since they share the process; this fix matches that.

## Checklist

**Test Results Note:** `RuntimeApiServerTest` passes 56/56 on two machines, including verifying the regression test fails against unpatched code and passes after the fix. After merging `upstream/main` into this branch, a full local run is down to 8 failures/8 errors out of 15907 tests. Two are new to this branch (`AutoScalingServiceTest`, `ContainerPlatformDockerIntegrationTest`), both confirmed to have zero code path to `RuntimeApiServer` (no reference to it or `notifyExtensionsOfInvoke` anywhere in either package). The rest reproduced identically across three independent runs on two separate machines (`ZipExtractorTest`, `RamServiceTest`, `S3ConcurrentWriteReadTest`, `S3IntegrationTest`, `AmiImageToolTest`, `TlsIntegrationTest`/`TlsUserCertIntegrationTest`, `ContainerLauncherCodeVolumeReuseTest`, `RdsContainerManagerTest`, `RdsDataServiceTest`), none touch Lambda Extensions or this change.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/runtime/RuntimeApiServer.java
@@ -725,9 +725,53 @@ public class RuntimeApiServer {
         return close();
     }
 
+    /**
+     * Fan the INVOKE event out to every subscribed extension. Called from
+     * {@link #sendInvocation(RoutingContext, PendingInvocation)}, i.e. at the moment the runtime
+     * actually receives the invocation, not at enqueue() time. Nothing in the launch path waits for
+     * an internal extension to register, so an extension registering from inside the runtime process
+     * may not be in {@code extensions} yet when enqueue() runs; dispatching there dropped the event
+     * entirely (see #2573).
+     *
+     * <p>{@code stopped} is read under the lock for the same reason sendInvocation()'s failure path
+     * reads it: quiesce() may already have swept the extensions and queued SHUTDOWN, and an INVOKE
+     * offered afterwards would be drained ahead of that SHUTDOWN by /extension/event/next.
+     */
+    private void notifyExtensionsOfInvoke(PendingInvocation invocation) {
+        List<Runnable> dispatches = new ArrayList<>();
+        synchronized (lock) {
+            if (stopped || extensions.isEmpty()) {
+                return;
+            }
+            ExtensionEvent event = ExtensionEvent.invoke(
+                    invocation.getRequestId(), invocation.getDeadlineMs(), invocation.getFunctionArn());
+            for (RegisteredExtension ext : extensions.values()) {
+                if (!ext.isSubscribedTo(ExtensionEvent.Type.INVOKE)) {
+                    continue;
+                }
+                RoutingContext waitingCtx = ext.takeWaitingContext();
+                if (waitingCtx != null) {
+                    dispatches.add(() -> {
+                        if (!waitingCtx.response().ended()) {
+                            sendExtensionEvent(waitingCtx, event);
+                        } else {
+                            synchronized (lock) {
+                                ext.getPendingEvents().offer(event);
+                            }
+                        }
+                    });
+                } else {
+                    ext.getPendingEvents().offer(event);
+                }
+            }
+        }
+        for (Runnable dispatch : dispatches) {
+            vertx.runOnContext(v -> dispatch.run());
+        }
+    }
+
     public CompletableFuture<InvokeResult> enqueue(PendingInvocation invocation) {
         boolean rejected;
-        List<Runnable> dispatches = new ArrayList<>();
         RoutingContext waitingCtxForInvocation = null;
 
         synchronized (lock) {
@@ -738,30 +782,6 @@ public class RuntimeApiServer {
             // stays a single atomic step.
             rejected = stopped || faulted;
             if (!rejected) {
-                if (!extensions.isEmpty()) {
-                    ExtensionEvent event = ExtensionEvent.invoke(
-                            invocation.getRequestId(), invocation.getDeadlineMs(), invocation.getFunctionArn());
-                    for (RegisteredExtension ext : extensions.values()) {
-                        if (!ext.isSubscribedTo(ExtensionEvent.Type.INVOKE)) {
-                            continue;
-                        }
-                        RoutingContext waitingCtx = ext.takeWaitingContext();
-                        if (waitingCtx != null) {
-                            dispatches.add(() -> {
-                                if (!waitingCtx.response().ended()) {
-                                    sendExtensionEvent(waitingCtx, event);
-                                } else {
-                                    synchronized (lock) {
-                                        ext.getPendingEvents().offer(event);
-                                    }
-                                }
-                            });
-                        } else {
-                            ext.getPendingEvents().offer(event);
-                        }
-                    }
-                }
-
                 waitingCtxForInvocation = waitingContexts.poll();
                 if (waitingCtxForInvocation == null) {
                     pendingQueue.offer(invocation);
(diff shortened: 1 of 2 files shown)


pgermosen (commented):
Retitled this to match Conventional Commits, the branch name had ended up as the PR title.

There's also a merge commit in the history ("Merge remote-tracking branch 'origin/main' into fix/lambda-extension-..."), which is what's failing the Conventional Commits check on commits along with the title. Can you rebase onto main instead of merging so that commit drops out? Should just leave the three fix/test commits.

On the substance this looks good, verified the AWS Extensions API timing claim (INVOKE fans out to extensions at the runtime's own /next call, not earlier) against the actual docs and it matches, and gating the fan-out on the response write succeeding is the right way to avoid a duplicate INVOKE on requeue.

ammaraltahan:
Thank you @pgermosen, I was a bit hesitant about the merge commit but thought it might be ok, no worries I'll rebase and fix the merge titles blocker.

ammaraltahan:
rebased and pushed. It should be all good now. Thanks.

pgermosen (approved):
Checked the core fix against AWS's own Lambda Extensions API docs: the Invoke event is documented as firing in response to the runtime's own Next API request, which is exactly what moving the fan-out into sendInvocation()'s successful write achieves, versus the old enqueue()-time dispatch that could race an internal extension's registration and silently drop the event. Approving.

ammaraltahan:
can't wait for the next release.

PR #3081, opened 2026-09-05
fix(s3): verify SigV4 signature on presigned POST uploads

### LINKED ISSUE #2334: [BUG] S3: presigned POST upload accepts fabricated/garbage credentials, or none at all
## Service

S3

## AWS API Action

PostObject (browser-form `multipart/form-data` upload to a bucket)

## Expected behavior

A presigned POST upload should only succeed if the `policy` field is present, base64-decodes to a policy document, and `x-amz-signature` is a valid SigV4 signature of that policy computed with a known secret key — real S3 returns `403 AccessDenied` otherwise.

## Actual behavior

Floci's presigned-POST handler only validates the *conditions* inside the `policy` field if one happens to be present (`S3Controller.validatePolicyConditions`). It never checks `x-amz-credential` or `x-amz-signature` against a real key, and if the `policy` field is omitted entirely, no validation happens at all — the upload is accepted unconditionally.

## Reproduction

```bash
docker run -d -p 4566:4566 -e FLOCI_SERVICES_S3_ENFORCE_AUTH=true floci/floci:1.6.0

aws --endpoint-url http://localhost:4566 s3api create-bucket --bucket b

curl -s -o /dev/null -w '%{http_code}\n' \
  -F 'key=anon-upload.txt' \
  -F 'x-amz-credential=totally-fake/20260101/us-east-1/s3/aws4_request' \
  -F 'x-amz-signature=0000000000000000000000000000000000000000000000000000000000dead' \
  -F 'file=@-;filename=anon-upload.txt;type=text/plain' <<< 'uploaded with no real credentials' \
  http://localhost:4566/b
# 204 — accepted, no policy field at all, garbage credential/signature

aws --endpoint-url http://localhost:4566 s3 cp s3://b/anon-upload.txt -
# uploaded with no real credentials
```

## Notes

This is independent of `FLOCI_SERVICES_S3_ENFORCE_AUTH` — reproduces identically with the flag on or off, since nothing in this code path ever consults it.

Not found from a real user report — this turned up while investigating #2318 (S3 enforce-auth write-path gaps, see #2326) using AI-assisted analysis (Claude), while mapping every S3 write endpoint to see which ones actually check credentials. Filing separately since presigned POST is its own auth mechanism (policy-document + SigV4), distinct from both #2318's object-write paths and the header-signature gap.

## Environment

- Floci version / image tag: `1.6.0`
- How you're running Floci (Docker / native / `mvn quarkus:dev`): `mvn quarkus:dev`, reproduced against a clean `main` checkout (commit `ff0201a1`)

syedsadiquh:
## Summary

Presigned POST uploads (browser-form `multipart/form-data` upload to a bucket) never verified the SigV4 signature — only the `policy` field's conditions, and only if `policy` was present at all. Omitting `policy` entirely skipped all validation, so fabricated or absent credentials were accepted unconditionally, regardless of `FLOCI_SERVICES_S3_ENFORCE_AUTH`.

When `enforce-auth` is on, this PR now requires `policy` + `x-amz-credential` + `x-amz-signature`, verifies the signature against the resolved secret key (same resolution `PreSignedUrlFilter` uses for presigned URLs, including the `test`/`test` dev credential), and enforces the policy's `expiration`. Behavior is unchanged when `enforce-auth` is off.

Closes #2334

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Real S3 rejects a presigned POST unless `policy` decodes to a policy document and `x-amz-signature` is a valid SigV4 signature of that policy under a real secret key — returning `403 AccessDenied` otherwise. Floci accepted the upload unconditionally in that case. This PR brings the emulator's behavior in line: same required fields, same `403 AccessDenied` / `403 SignatureDoesNotMatch` failure modes.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -2763,7 +2763,7 @@ public class S3Controller {
         }
 
         if (s3Service.isAuthEnforced()) {
-            validatePresignedPostAuth(lcFields, bucket, fileData.length);
+            validatePresignedPostAuth(lcFields, bucket, key, fileData.length);
         } else {
             // Validate policy conditions if present
             String policy = lcFields.get("policy");
@@ -2819,7 +2819,7 @@ public class S3Controller {
      * SigV4-signed by a known secret key referenced through {@code x-amz-credential}, matching
      * real S3's behavior of rejecting fabricated or absent credentials with 403 AccessDenied.
      */
-    private void validatePresignedPostAuth(Map<String, String> fields, String bucket, int contentLength) {
+    private void validatePresignedPostAuth(Map<String, String> fields, String bucket, String key, int contentLength) {
         String policy = fields.get("policy");
         String credential = fields.get("x-amz-credential");
         String signature = fields.get("x-amz-signature");
@@ -2850,6 +2850,14 @@ public class S3Controller {
                     "The request signature we calculated does not match the signature you provided. "
                             + "Check your key and signing method.", 403);
         }
+
+        // Route through the same authorization check every other S3 write path uses, so this
+        // path automatically inherits any future strengthening (e.g. per-principal IAM policy
+        // evaluation) instead of silently staying behind. Today it only confirms accessKeyId is
+        // known - strictly weaker than the signature check just performed above - but that will
+        // change if authorizeObjectWrite grows real policy evaluation.
+        s3Service.authorizeObjectWrite(bucket, key, "s3:PutObject",
+                new S3Service.RequestAuthorization(true, accessKeyId));
 
         validatePolicyExpiration(policy);
         validatePolicyConditions(policy, bucket, fields, contentLength);
--- a/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/services/s3/S3Controller.java
@@ -11,6 +11,7 @@ import io.github.hectorvent.floci.core.common.XmlParser;
 import io.github.hectorvent.floci.core.common.RegionResolver;
 import io.github.hectorvent.floci.core.common.RequestContext;
 import io.github.hectorvent.floci.services.cloudtrail.CloudTrailService;
+import io.github.hectorvent.floci.services.iam.IamService;
 import io.github.hectorvent.floci.services.sns.SnsQueryHandler;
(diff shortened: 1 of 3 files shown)


pgermosen:
Went through this closely since it's auth-critical. The five SigV4 findings (required form fields, credential scope, date consistency, impossible timestamps, required expiration) are all genuinely fixed, confirmed in the code directly.

One thread is still open though: `handlePresignedPost` never calls `authorizeObjectWrite`, unlike every other S3 write path in this controller. It's marked resolved but I don't see a commit addressing it. Digging into what `authorizeObjectWrite` actually does for a signed request today, it just checks the access key is known, it doesn't evaluate per-principal IAM policy, so this path's own signature verification (which proves possession of the real secret key) is at least as strong a check for that specific case. But it's still worth routing through the shared function for consistency, so this path automatically picks up any future strengthening of that authorization logic instead of silently staying behind.

syedsadiquh:
Good point, done in syedsadiquh/floci@3ff9de0. `validatePresignedPostAuth` now calls `s3Service.authorizeObjectWrite(bucket, key, "s3:PutObject", new RequestAuthorization(true, accessKeyId))` right after the SigV4 signature check succeeds - same call every other write path makes. Agreed it's a no-op today given `authorizeObjectWrite` only checks the access key is known for signed requests (weaker than the signature check we just did), but this way the path picks up any future strengthening of that shared logic automatically instead of drifting.

syedsadiquh:
Follow-up on this: I looked closer at whether S3 checks IAM policy for signed requests at all, and it's not as bad as I first thought.

There's already a filter (`IamEnforcementFilter`/`IamPolicyEvaluator`) that correctly checks a user's IAM policy (Allow/Deny, etc.) for header-signed requests, across every service including S3 - not just "is this access key known."

The actual gap: that filter only looks at the `Authorization` header. Presigned URL and presigned POST requests sign via query params / form fields instead of a header, so they skip this check entirely, even with `floci.iam.enforcement-enabled` on. That's a different bug from what this PR fixes (signature verification) and from the `authorizeObjectWrite` fix in 3ff9de0 above, so I filed it separately: #3195.

pgermosen (approved):
authorizeObjectWrite wiring looks right, and I re-checked the credential-scope and strict-timestamp fixes are still solid. Also went and confirmed the IamEnforcementFilter gap you found - it really is header-only and pre-existing, so #3195 is the right place for that, not this PR. Good to merge.

PR #3082, opened 2026-09-05
feat(lambda): support self-managed Apache Kafka event source mappings (#3062)

### LINKED ISSUE #3062: [BUG] CreateEventSourceMapping rejects self-managed Kafka source with 'EventSourceArn is required'
## Service

Lambda (Event Source Mappings — self-managed Apache Kafka)

## AWS API Action

CreateEventSourceMapping (with `SelfManagedEventSource`, no `EventSourceArn`)

## Expected behavior

On real AWS, `CreateEventSourceMapping` for a self-managed Kafka source (i.e. any Kafka cluster that isn't MSK — Confluent Cloud, a self-hosted cluster, etc.) does **not** require `EventSourceArn`. Per the AWS API, the source is identified by `SelfManagedEventSource.Endpoints.KAFKA_BOOTSTRAP_SERVERS` instead — `EventSourceArn` only applies to ARN-addressable AWS-native sources (SQS, DynamoDB Streams, Kinesis, MSK). This is also how Terraform's `aws_lambda_event_source_mapping` resource is documented and used for `self_managed_event_source` — no `event_source_arn` argument is set.

## Actual behavior

The call fails, requiring an ARN that has no meaning for a self-managed source:

```
Error: creating Lambda Event Source Mapping (Self-Managed Apache Kafka): operation error Lambda:
CreateEventSourceMapping, https response error StatusCode: 400, RequestID: dfe5cf2b-d773-4f09-93ef-958a266e74ca,
InvalidParameterValueException: EventSourceArn is required
```

## Reproduction

```hcl
terraform {
  required_providers {
    aws = { source = "hashicorp/aws", version = "~> 5.0" }
  }
}

provider "aws" {
  region                      = "us-east-1"
  access_key                  = "test"
  secret_key                  = "test"
  skip_credentials_validation = true
  skip_metadata_api_check     = true
  skip_requesting_account_id  = true

  endpoints {
    lambda = "http://localhost:4566"
  }
}

# ... aws_lambda_function.example already exists / created earlier in the stack ...

resource "aws_lambda_event_source_mapping" "example" {
  function_name     = aws_lambda_function.example.arn
  topics            = ["orders.v1"]
  starting_position = "LATEST"

  self_managed_event_source {
    endpoints = {
      KAFKA_BOOTSTRAP_SERVERS = "pkc-xxxxx.region.provider.confluent.cloud:9092"
    }
  }

  source_access_configuration {
    type = "BASIC_AUTH"
    uri  = aws_secretsmanager_secret.kafka_credentials.arn
  }
}
```

```bash
terraform apply -auto-approve
# Error: InvalidParameterValueException: EventSourceArn is required
```

Equivalent raw CLI call (no ARN passed, matching the self-managed-source contract):

```bash
aws --endpoint-url http://localhost:4566 lambda create-event-source-mapping \
  --function-name example \
  --topics orders.v1 \
  --starting-position LATEST \
  --self-managed-event-source '{"Endpoints":{"KAFKA_BOOTSTRAP_SERVERS":"pkc-xxxxx.region.provider.confluent.cloud:9092"}}' \
  --source-access-configurations '[{"Type":"BASIC_AUTH","URI":"arn:aws:secretsmanager:us-east-1:000000000000:secret:example"}]'
```

## Environment

- Floci version / image tag: `floci/floci:latest` @ `2.0.1` (image built 2026-09-01, digest `sha256:4e451c39c7bb88e3cd4f87e8fc0c25d5b47695a51185d521e2241fa00486e8eb`)
- Java SDK version (if applicable): N/A (Terraform `hashicorp/aws` ~> 5.0 via `terraform-plugin-sdk`)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker (`docker compose`), with `/var/run/docker.sock` mounted

## Scope

`CreateEventSourceMapping` validation should treat `EventSourceArn` as optional when `SelfManagedEventSource` is supplied instead (mirroring how `EventSourceArn` is already presumably optional-vs-required correctly for the AWS-native source types). Didn't find an existing issue for this specifically — the closest related ones I found (#2848, #2369, #1537) are about MSK/managed-source ESM behavior (`FunctionResponseTypes`, `StartingPosition`, `DestinationConfig`), not about self-managed sources being rejected outright at creation time.

- [ ] Not right now — happy to test a fix against our repro if one lands.

vuthehuyht:
## Summary

Enable AWS Lambda Event Source Mappings (ESM) to accept, validate, store, and return self-managed Apache Kafka event source configurations without requiring EventSourceArn, resolving #3062.

Key changes:
- **Model**: Added SelfManagedEventSource, Topics, and SourceAccessConfigurations to EventSourceMapping.
- **REST Wire Protocol**: Updated LambdaController.buildEsmResponse to serialize Kafka fields and omit EventSourceArn when null (preventing Terraform refresh from detecting drift and triggering unintended recreation).
- **Service & Poller Safety**: Added Kafka source-type discrimination in LambdaService, validated mandatory Kafka fields (KAFKA_BOOTSTRAP_SERVERS and non-empty Topics), guarded pollers against null EventSourceArn, supported AT_TIMESTAMP starting position for Kafka, and supported updating Topics and SourceAccessConfigurations.
- **CloudFormation**: Dedicated LambdaEventSourceMappingCfnProvisioner with intrinsic resolution, rejection of create-only property mutations with ValidationError, and populating Id and EventSourceMappingArn attributes.
- **Docs**: Documented self-managed Kafka event source mapping support in docs/services/lambda.md.
- **Tests**: Added comprehensive unit tests for serialization, service validation, provisioner lifecycle, live HTTP REST integration tests (EsmIntegrationTest), and CloudFormation stack lifecycle tests (CloudFormationIntegrationTest).

Closes #3062

## Type of change

- [x] New feature (eat:)

## AWS Compatibility

- Verified against AWS Lambda REST JSON wire protocol specifications.
- Validated that EventSourceArn is omitted from the JSON payload when null, matching real AWS Lambda behaviour for Self-Managed Kafka mappings.
- Verified StartingPosition behavior: AT_TIMESTAMP is supported for Amazon Kinesis and self-managed Apache Kafka event sources (with StartingPositionTimestamp), missing KAFKA_BOOTSTRAP_SERVERS or empty Topics returns 400 InvalidParameterValueException.

## Checklist

### DIFF
--- a/docs/services/lambda.md
+++ b/docs/services/lambda.md
@@ -720,7 +720,7 @@ aws lambda update-function-code \
 
 ## Event Source Mappings
 
-Connect Lambda to SQS, Kinesis, or DynamoDB Streams:
+Connect Lambda to SQS, Kinesis, or DynamoDB Streams. Self-managed Apache Kafka event source mappings are accepted, validated, persisted, and returned on the wire, but Floci does not run an active Kafka consumer poller:
 
 ```bash
 # SQS trigger
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -4165,6 +4165,24 @@ public class CloudFormationResourceProvisioner {
             req.put("FunctionResponseTypes", functionResponseTypes);
         }
 
+        // Support self-managed event source (Kafka/SelfManagedEventSource)
+        JsonNode selfManagedEventSource = props.get("SelfManagedEventSource");
+        if (selfManagedEventSource != null && !selfManagedEventSource.isNull()) {
+            req.put("SelfManagedEventSource", objectMapper.convertValue(selfManagedEventSource, Map.class));
+        }
+
+        // Kafka topic list
+        List<String> topics = resolveStringList(props, "Topics", engine);
+        if (!topics.isEmpty()) {
+            req.put("Topics", topics);
+        }
+
+        // Event source access configurations (SourceAccessConfigurations)
+        JsonNode sourceAccessConfigurations = props.get("SourceAccessConfigurations");
+        if (sourceAccessConfigurations != null && !sourceAccessConfigurations.isNull()) {
+            req.put("SourceAccessConfigurations", objectMapper.convertValue(sourceAccessConfigurations, List.class));
+        }
+
         var esm = lambdaService.createEventSourceMapping(region, req);
         r.setPhysicalId(esm.getUuid());
         r.getAttributes().put("Id", esm.getUuid());
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -255,7 +255,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::IAM::ManagedPolicy",
             "AWS::IAM::Policy",
             "AWS::IAM::User",
-            "AWS::Lambda::EventSourceMapping",
             "AWS::Lambda::Function",
             "AWS::Lambda::LayerVersion",
             "AWS::RDS::DBCluster",
@@ -451,8 +450,6 @@ public class CloudFormationResourceProvisioner {
                                 region,
                                 accountId,
                                 stackName);
-                case "AWS::Lambda::EventSourceMapping" ->
-                        provisionLambdaEventSourceMapping(resource, properties, engine, region);
                 case "AWS::Cognito::UserPool" ->
                         provisionCognitoUserPool(resource, properties, engine, region, accountId, stackName);
                 case "AWS::Cognito::UserPoolClient" ->
@@ -743,7 +740,6 @@ public class CloudFormationResourceProvisioner {
             case "AWS::ApiGateway::RestApi" -> apiGatewayService.deleteRestApi(region, physicalId);
             case "AWS::ApiGatewayV2::Api" -> apiGatewayV2Service.deleteApi(region, physicalId);
             case "AWS::StepFunctions::StateMachine" -> stepFunctionsService.deleteStateMachine(physicalId);
-            case "AWS::Lambda::EventSourceMapping" -> lambdaService.deleteEventSourceMapping(physicalId);
             case "AWS::Lambda::LayerVersion" -> deleteLambdaLayerVersion(physicalId, region);
             case "AWS::Cognito::UserPool" -> cognitoService.deleteUserPool(physicalId);
             case "AWS::Cognito::UserPoolClient" -> cognitoService.deleteUserPoolClient(physicalId);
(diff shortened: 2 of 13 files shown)


hectorvent:
Thank you for taking the extraction route the bot suggested rather than growing the switch: the arm is gone, the provisioner owns create, update and delete, and I checked the new wire shape against the Lambda model (`EventSourceArn` optional, `KAFKA_BOOTSTRAP_SERVERS` as the only endpoint key, `Topics` length 1, the response omitting `EventSourceArn`), which is right. Seven suites pass locally on 405ec64.

**(blocking)** Create-only properties are silently ignored on update. The registry schema marks `EventSourceArn`, `StartingPosition`, `StartingPositionTimestamp` and `SelfManagedEventSource` as createOnly, but `provision` sends every update to `updateEventSourceMapping`, which only reads `BatchSize`, `Enabled`, `FunctionName`, `Topics` and a few others. A template that changes the bootstrap servers or the source ARN therefore reports `UPDATE_COMPLETE` while the mapping still points at the old source. Compare the prior mapping (`priorPhysicalId`) with the resolved values and either replace (create the new mapping and delete the old, the way `Ec2LaunchTemplateCfnProvisioner` and `PipesCfnProvisioner` do) or refuse with `ValidationError` "requires resource replacement, which is not supported" as `SqsCfnProvisioner` and `IamRoleCfnProvisioner` do. A unit case for a changed `SelfManagedEventSource` would pin it.

**(blocking)** `AT_TIMESTAMP` is refused for Kafka. `parseStartingPosition` only allows it when the ARN contains `:kinesis:`, so the new path always throws "AT_TIMESTAMP is only supported for Amazon Kinesis event sources", and the description lists that as verified AWS behaviour. The model says the opposite: `StartingPosition` documents AT_TIMESTAMP for "Amazon Kinesis streams, Amazon DocumentDB, Amazon MSK, and self-managed Apache Kafka". Let the Kafka branch through (there is nothing to honour since nothing polls, so storing the timestamp is enough) and fix the description line.

**(blocking)** `Fn::GetAtt` gap unrecorded. The schema declares `Id` and `EventSourceMappingArn` as read-only; the provisioner sets `Id` only, and there is no row in `src/test/resources/cloudformation/getatt-attribute-gaps.tsv`. `CfnSchemaCoverageTest` needs the schema corpus, so it passes in CI and fails locally with `local/aws/cfn-resource-schemas` present. Either set the attribute from the mapping's ARN if the service can produce one, or add the row with a reason.

**(docs)** Nothing in `docs/services/lambda.md` says a self-managed Kafka mapping is accepted and returned but never polls (there is no Kafka consumer in `services/lambda`). One sentence in the Event Source Mappings section, so a Terraform user knows the resource creates and what does not happen after.

Optional: `UpdateEventSourceMapping` has no `Topics` member in the model, so accepting it on the REST update is an extension only raw callers reach; and an SDK round trip for the Kafka shape in sdk-test-java would prove the new response fields deserialize, since EsmIntegrationTest is raw HTTP only.

Note for sequencing: #2996 edits the arm this PR deletes, so whichever lands second needs a rebase.

javierrojas10:
Tested this against the PR's CI-built `floci-native-image-amd64` artifact locally (Docker, `docker.sock` mounted per floci's usual RDS/etc. setup).

Ran a real `terraform apply` creating an `aws_lambda_event_source_mapping` with `self_managed_event_source` (Confluent Cloud, no MSK) + `BASIC_AUTH` — exactly the shape that failed with `EventSourceArn is required` before this PR.

Confirmed working two ways:

**Terraform state:**
```
state                      = "Enabled"
topics                     = ["staging.orders.v1"]
event_source_arn           = null
self_managed_event_source {
  endpoints = { "KAFKA_BOOTSTRAP_SERVERS" = "https://pkc-xxxxx....confluent.cloud:443" }
}
source_access_configuration {
  type = "BASIC_AUTH"
  uri  = "arn:aws:secretsmanager:...:secret:..."
}
```

**Raw AWS CLI against floci** (independent check, not just trusting Terraform's view):
```
aws --endpoint-url=http://localhost:4566 lambda list-event-source-mappings --function-name <fn>
```
returned the same mapping back, `EventSourceArn` correctly omitted, matching real AWS's wire protocol for self-managed sources.


This resolves #3062 for me— thanks for the fast turnaround!

pgermosen (approved):
Went through all twelve review findings directly in the code since none had a reply comment despite the fix commits, and every one is genuinely addressed: the EventSourceArn/Kafka mutual exclusivity check closes both the cross-region and mixed-source poller issues, Topics elements are validated as non-blank strings on both create and update, function target updates now correctly sync functionName alongside functionArn with a region check, and both SelfManagedEventSource and SourceAccessConfigurations go through the recursive intrinsic resolver now. Thorough work getting through that list. Approving.

PR #3088, opened 2026-09-05
fix(dynamodb): append list values for out-of-range indexes

### LINKED ISSUE #3073: [BUG] DynamoDB: SET on a list index past the end pads with NULLs instead of appending
## Service

DynamoDB

## AWS API Action

UpdateItem (UpdateExpression `SET list[n] = :v` with an index past the end of the list)

## Expected behavior

Measured on real DynamoDB (us-east-1, 2026-09-05). The item starts with `l = [a, b]`.

- `SET l[10] = :v` appends the element to the end. Result: `[a, b, c]`. No NULL padding. The developer guide says the same: "If the specified index does not exist, the new element is appended to the end of the list."
- `SET l[2000000000] = :v` also just appends. The response is instant.
- `SET l[4294967295] = :v` is rejected: `ValidationException: 1 validation error detected: Invalid UpdateExpression: List index is not within the allowable range; index: [4294967295]`. The maximum allowed index is 4294967294.

## Actual behavior

- `SET l[10] = :v` returns `[a, b, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, c]`. Floci fills the gap with NULL elements (`padAndSet` in `DynamoDbService`).
- Because of that padding loop, a large in-range index like `l[2000000000]` would allocate two billion list elements. Out of memory risk. (Not executed, read from the code.)
- `SET l[4294967295] = :v` returns HTTP 500. The index is parsed with `Integer.parseInt`, and this value does not fit in an int.

## Reproduction

```bash
  --table-name ListSetProbe \

  --table-name ListSetProbe \
  --item '{"pk":{"S":"x"},"l":{"L":[{"S":"a"},{"S":"b"}]}}'

aws --endpoint-url http://localhost:4566 dynamodb update-item \
  --table-name ListSetProbe --key '{"pk":{"S":"x"}}' \
  --update-expression 'SET l[10] = :v' \
  --expression-attribute-values '{":v":{"S":"c"}}' \
  --return-values ALL_NEW
# Floci: l has 11 elements, 8 of them NULL. Real AWS: [a, b, c].

aws --endpoint-url http://localhost:4566 dynamodb update-item \
  --table-name ListSetProbe --key '{"pk":{"S":"x"}}' \
  --update-expression 'SET l[4294967295] = :v' \
  --expression-attribute-values '{":v":{"S":"e"}}'
# Floci: HTTP 500. Real AWS: ValidationException (allowable range).
```

## Environment

- Floci version / image tag: 2.0.1, main
- How you're running Floci: `java -jar target/quarkus-app/quarkus-run.jar`

AbarnaaSree:
## Summary

Fixes #3073.

- Make DynamoDB list indexes use `long` instead of `int`
- Append to the list when `SET` uses an index past the current list size
- Prevent large list indexes from allocating unnecessary NULL padding
- Reject indexes above DynamoDB's maximum allowed index with a ValidationException
- Add regression tests for out-of-range, huge, maximum-valid, and invalid indexes

## Testing

- `mvn -Dtest=DynamoDbServiceTest test`
- 112 tests passed

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -2552,8 +2552,10 @@ public class DynamoDbService implements ResourceProvider {
         }
         return tokens;
     }
-    // Sets an existing list element or appends the value when the index is past the end.
-    // Pads arr with NULL elements up to idx-1, then sets/appends value at idx.
+    /**
+     * Replaces the element at idx when the index is within the list bounds;
+     * otherwise appends the value to the end of the list.
+    */
 
     private void setOrAppend(ArrayNode arr, long idx, JsonNode value) {
         if (idx < arr.size()) {
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -106,6 +106,7 @@ public class DynamoDbService implements ResourceProvider {
     private DynamoDbStreamService streamService;
     private KinesisStreamingForwarder kinesisForwarder;
     private S3Service s3Service;
+    private static final long MAX_DYNAMODB_LIST_INDEX = 4_294_967_294L;
 
     @Inject
     public DynamoDbService(StorageFactory storageFactory, RegionResolver regionResolver,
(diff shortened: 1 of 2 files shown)


pgermosen (commented):
Checked the conflict directly rather than just noting it exists, it's worth a heads up because it's not a simple textual clash. main already has a padAndSet method at the exact same two call sites this PR touches, and it does the thing this PR's own description calls out as the bug: padding the list with NULL elements up to the target index before setting the value. That method's been on main since an old, unrelated PR (#826) and has no test coverage anywhere pinning that behavior, so it looks like a longstanding, untested bug rather than a competing fix in flight.

Checked AWS's own UpdateExpressions docs for SET on a list index: "If the element doesn't already exist, SET appends the new element at the end of the list", no mention of padding at all. So padAndSet's current behavior on main doesn't match real AWS, and this PR's setOrAppend (replace if in range, otherwise append once, no padding) is the one that's actually correct.

Worth flagging for the rebase: this isn't a "keep both" merge, padAndSet and its two call sites need to go away entirely and be replaced by setOrAppend, or the fix this PR is making gets silently lost in the merge.

pgermosen:
Rebase looks good, the old padAndSet path is fully gone and setOrAppend handles both call sites now. One small thing: the comment right above setOrAppend still describes the old behavior ("Pads arr with NULL elements up to idx-1, then sets/appends value at idx"), but that's not what this method does anymore, it just replaces in-range or appends a single element, no padding. Looks like a leftover from the merge that didn't get updated. Worth a quick fix so it doesn't confuse the next person reading it.

AbarnaaSree:
Fixed, thanks! Updated the `setOrAppend` comment to accurately describe the current replace-or-append behavior.

hectorvent (approved):
Seconding pgermosen's review as the maintainer: with `padAndSet` gone and `setOrAppend` serving both call sites, the append-instead-of-pad semantics match what DynamoDB does for an out-of-range index, and the comment now says what the method does.

Nothing further from me. Merging.

PR #3090, opened 2026-09-05
refactor(ses): move the tag dispatch helpers into their domain services

okinaka:
## Summary

Follow-up to #2356, continuing the post-split cleanup (#2881, #2984, #3030): moves the last nine
tag-dispatch helpers out of the `SesService` facade into their domain services, matching the
pattern the other four taggable types already follow. The tenant, configuration-set, and template
`listTags` / `tag` / `untag` operations now live on `SesTenantService`,
`SesConfigurationSetService`, and `SesTemplateService` respectively, and the three dispatch
switches call `<domain>Service.listTags/tag/untag` uniformly for all seven types.

Each helper depended only on its own domain service plus the shared dependency-free `SesTags`, so
this is mesh-free code motion. As a bonus, `SesTenantService.tenantForTagArn` and `mutateTags`
lose their last external callers and drop to package-private (kept non-private for the existing
unit tests that pin the stale-id and defensive-copy regressions).

No behavior change: wordings, ordering, and log lines are verbatim.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Pure code motion behind the facade; no wire-visible change. The full SES suite (1037 tests,
including the per-type tag integration tests and the tenant tag lifecycle) passes unchanged.

## Checklist

- [x] New or updated integration test added (behavior-preserving refactor: the existing tag integration tests pin the wire behavior, and each service gained a unit test pinning the new public listTags/tag/untag surface)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesConfigurationSetService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesConfigurationSetService.java
@@ -9,6 +9,7 @@ import io.github.hectorvent.floci.services.ses.model.CloudWatchDimensionConfigur
 import io.github.hectorvent.floci.services.ses.model.ConfigurationSet;
 import io.github.hectorvent.floci.services.ses.model.EventDestination;
 import io.github.hectorvent.floci.services.ses.model.SuppressionOptions;
+import io.github.hectorvent.floci.services.ses.model.Tag;
 import io.github.hectorvent.floci.services.ses.model.VdmOptions;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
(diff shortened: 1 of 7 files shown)


pgermosen (approved):
Traced the moved code line by line rather than trusting the no-behavior-change claim: the comparator ordering for tenant tags, the copy-on-write handling, and the locking in SesTenantService.mutateTags are all verbatim, just relocated with call sites updated to go through the domain services directly. Also checked the NotFoundException/404 shape against SES's ListTagsForResource/TagResource docs, matches exactly.

Clean move, good bonus of dropping tenantForTagArn/mutateTags to package-private now that nothing outside the domain needs them public. Approving.

PR #3092, opened 2026-09-05
fix(appconfigdata): honor requested poll interval

nkootstra:
## Summary

Fix AppConfigData poll interval handling.

Floci now returns the requested RequiredMinimumPollIntervalInSeconds value instead of always returning 15. It validates the AWS range of 15 to 86400 seconds and normalizes missing or invalid legacy session values to the 15-second default.

## Type of change

- [x] Bug fix (fix:)
- [ ] New feature (feat:)
- [ ] Docs / chore

## AWS Compatibility

Floci previously ignored the requested poll interval and always returned 15 seconds. Clients requesting another valid interval received incorrect polling guidance.

The implementation now:

- Returns the requested interval on every poll.
- Rejects fractional, null, non-numeric, and out-of-range values with BadRequestException.
- Normalizes missing or invalid values in legacy sessions to 15.
- Preserves the requested interval across successive polls.

Verified with AppConfig integration tests, boundary tests, malformed-input tests, legacy-session tests, and updated Java SDK compatibility assertions. All 38 focused tests passed.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppConfigTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppConfigTest.java
@@ -204,8 +204,8 @@ class AppConfigTest {
 
     @Test
     @Order(12)
-    @DisplayName("Poll interval: requested 60s but emulator returns 15s (known deviation from AWS)")
-    void requiredMinimumPollIntervalIsAcceptedButNotEnforced() {
+    @DisplayName("Poll interval: requested minimum is returned to the client")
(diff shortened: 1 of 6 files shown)


pgermosen (approved):
Checked the range/validation against AppConfigData's StartConfigurationSession reference, the 15-86400 integer range and BadRequestException match exactly, including rejecting fractional values since AWS types it strictly as Integer. Also traced normalizePollInterval to confirm the legacy-session finding is actually handled, out-of-range persisted values get clamped back to 15 and the session self-heals. Approving.

PR #3096, opened 2026-09-05
fix(ec2): honor AMI architecture for containers

nkootstra:
## Summary

Honor the declared architecture of EC2 AMIs when starting Docker-backed instances.

Map arm64 to linux/arm64 and x86_64 to linux/amd64 for image pulls and container creation. Reject Windows AMIs before creating instance state. Keep unknown AMIs on the existing host-native path.

## Type of change

- [x] Bug fix (fix:)
- [ ] New feature (feat:)
- [ ] Docs / chore

## AWS Compatibility

Floci previously stored an AMI architecture but did not use it when starting the Docker container. This could run an image on the wrong architecture or cause a misleading launch timeout.

The selected Linux platform now applies to both Docker image pulls and container creation. Windows AMIs return an AWS-shaped UnsupportedOperation error because Floci runs EC2 guests as Linux containers.

Focused EC2 tests and the Docker platform integration test pass.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/AmiImageResolver.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/AmiImageResolver.java
@@ -1,5 +1,6 @@
 package io.github.hectorvent.floci.services.ec2;
 
+import io.github.hectorvent.floci.core.common.AwsException;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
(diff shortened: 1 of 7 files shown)


pgermosen (approved):
Traced the CreateImage-derived AMI finding directly: confirmed createImage captures the source's architecture into the registered AMI at creation time, and both the instance-type validation and the architecture-for-launch resolution check the registered image before falling back to the catalog, so a captured AMI's real architecture is now used consistently everywhere instead of falling back to instance-type guessing. Platform mapping and the Windows rejection both check out too. Approving.

PR #3099, opened 2026-09-05
fix(kinesis): reject non-string tag members instead of coercing them

soreavis:
## Summary

`CreateStream`, `AddTagsToStream` and `RemoveTagsFromStream` read tag values and `TagKeys` entries with `asText()`, so a number, boolean or null member was stored as its text — `"Foo": 5` landed as the tag value `"5"` — and an object or array member as the empty string, instead of being rejected. Follow-up to #2494, whose commit message split this out as a fix to land on its own; it takes the same line #2512 took for the SES tag parser.

Both parse sites on those three operations now reject a present-but-non-string tag value or `TagKeys` entry, a `Tags` member that isn't an object, and a `TagKeys` member that isn't a list, with `SerializationException` 400. A missing or null `Tags`/`TagKeys` member still yields no tags, as before. Two ordering effects ride along: `CreateStream` created the stream before parsing `Tags`, so a rejected `Tags` member would have left an untagged stream behind — it now parses first and the whole request is rejected with nothing created; and because the tags are parsed before the stream is created or looked up, a malformed member on an unknown or already-existing stream answers `SerializationException` where it previously answered `ResourceNotFoundException` or `ResourceInUseException` — all 400s, in the deserialize-before-validate order Smithy's server SDK documents.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

AWS models `Tags` as a string-to-string map on `CreateStream` and `AddTagsToStream`, and `TagKeys` as an array of strings on `RemoveTagsFromStream`, so none of the rejected shapes is a valid request under that model. The route reuses the `SerializationException` 400 the emulator already answers for a non-string SES tag member (#2512, where it was probe-confirmed against AWS) and for undecodable Kinesis `Data`. A null map value is rejected as well: the model's `TagMap` is a dense string map with no `sparse` trait, and unlike SES's `Key`/`Value` members nothing downstream validates it — that shape was not probed against real AWS, since no credentials are available here.

## Checklist

- [x] `./mvnw test` run locally: 15,960 tests, with the failures confined to 46 classes that fail identically on the pristine `c3d977a3` tip in a clean checkout — 24 Docker-backed classes (no Docker socket on this machine) and all 22 nested classes of the new `ResourceExplorer2IntegrationTest`, which trip on a duplicate DynamoDB `CreateTable` between nested classes — and none of them is a Kinesis class.
  - The Kinesis-adjacent suites are fully green: 365/365 across the Kinesis, KinesisAnalytics, DynamoDB-streaming, Pipes and AwsJsonCbor suites, with `KinesisJsonHandlerTest` at 76/76 including seven new tests — every non-string value shape rejected on `CreateStream` (stream proven not created) and on `AddTagsToStream` (existing tags untouched), a non-object `Tags` rejected, `RemoveTagsFromStream` rejecting non-string, non-list and object-shaped `TagKeys` with tags untouched, the preserved null and missing arms, and string `TagKeys` still removed — plus a wire-level test in `KinesisIntegrationTest` pinning the `__type: SerializationException` body and that no stream was created.
  - Red-proof: five of the eight new tests go red against the unfixed handler; the null-`Tags`, null-or-missing-`TagKeys` and string-`TagKeys` tests pass there by design.
  - Mutation check: eleven mutants — the four gates removed and the `TagKeys` list gate also loosened, the `CreateStream` ordering, the null arms and the message and status values, each killed by a named test.
- [x] New or updated integration test added — `createStreamRejectsANonStringTagValueWithSerializationException` in `KinesisIntegrationTest`.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -114,13 +114,15 @@ public class KinesisJsonHandler {
                 streamMode = mode;
             }
         }
-        service.createStream(streamName, shardCount, streamMode,
-                optionalMaxRecordSize(request), region);
         // CreateStream's optional Tags member was being dropped. That is invisible to a
         // hand-written script but not to Terraform: aws_kinesis_stream sets tags at create
         // time and then reads them back with ListTagsForStream, so an empty read produces a
         // permanent "tags will be updated in-place" diff on an unchanged configuration.
+        // Parsed before the stream exists so a malformed Tags member rejects the whole
+        // request instead of leaving an untagged stream behind.
         Map<String, String> tags = parseTags(request);
+        service.createStream(streamName, shardCount, streamMode,
+                optionalMaxRecordSize(request), region);
         if (!tags.isEmpty()) {
             service.addTagsToStream(streamName, tags, region);
         }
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Checked the shapes against Kinesis's real service model: TagMap is a dense string-to-string map and TagKeyList is a list of strings, so every rejected shape here (non-string values, non-object Tags, non-array TagKeys) is genuinely invalid under AWS's own model. The CreateStream reordering so a malformed Tags member doesn't leave an untagged stream behind is a nice catch too. Appreciate the honesty about what was and wasn't probed against real AWS. Approving.

PR #3102, opened 2026-09-05
fix(rds): bound PostgreSQL handshake packet lengths

nkootstra:
## Summary

Reject malformed or oversized PostgreSQL handshake packets before they can trigger large memory allocations. Apply a fixed 1 MiB limit during startup and authentication while keeping post-authentication traffic unchanged.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

PostgreSQL packet lengths were previously used for reads and allocations without an upper bound. A malformed client or backend packet could therefore request excessive memory during the handshake.

The 1 MiB limit includes PostgreSQL's 4-byte length field. It is large enough for normal startup and authentication messages while bounding untrusted handshake input. AWS does not define a smaller native PostgreSQL handshake limit, so this is an implementation safety bound rather than a wire-protocol change.

Post-authentication traffic remains unrestricted and continues to use the existing raw proxy bridge.

Added regression coverage for exact-limit acceptance, negative and undersized lengths, and oversized startup, password, backend authentication, SCRAM, and buffered backend messages.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/proxy/PostgresProtocolHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/proxy/PostgresProtocolHandler.java
@@ -43,6 +43,7 @@ public class PostgresProtocolHandler {
 
     private static final int SSL_REQUEST_CODE = 80877103;
     private static final int STARTUP_PROTOCOL_VERSION = 196608; // v3.0
+    private static final int MAX_HANDSHAKE_PACKET_LENGTH = 1_048_576;
 
     public static Socket authenticate(Socket client, Socket backend,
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Checked every minimum-length constant against the actual PostgreSQL wire protocol spec: 8 bytes for messages with a guaranteed follow-on Int32 code, 5 for PasswordMessage's length-plus-null-terminator floor, 4 for the generic backend-message case, all correct. Negative lengths fall out naturally from the same below-minimum check. Good test coverage on the exact-limit and rejection cases too. Approving.

Issue #2207, opened 2026-08-10
[BUG] cloudformation: deleting a rolled-back stack masks every failure as CDK NoStack


sam-fakhreddine:
## What's wrong

When a stack fails during CREATE, floci rolls it back and then **deletes it**:

```
INFO [CloudFormationService] Stack AWSAccelerator-LoggingStack-841986307243-us-east-1
  rolled back to a clean slate (ROLLBACK_COMPLETE)
```

Real CloudFormation does not do this. A stack whose first CREATE fails ends in `ROLLBACK_COMPLETE` and **remains describable** — that is precisely the state in which an operator reads `DescribeStackEvents` to find which resource failed and why. (It cannot be updated, only deleted and recreated, which is likely where the "clean slate" behaviour came from — but "cannot be updated" is not "does not exist".)

## Why it matters

The deletion destroys the diagnostic. The CDK toolkit's very next call is `DescribeStacks`, which now returns nothing, and the toolkit reports:

```
NoStack: CloudFormationStack object does not hold a stack
```

That message describes the toolkit's own confusion and says nothing about the failure. **Every primary error is masked identically**, so `NoStack` is what an operator sees whether the real cause was an IAM denial, a missing resource, or — in the case that produced this report — Lambda runtime-API port exhaustion:

```
Custom resource handler reported FAILED: Error: Failed to start Lambda container:
  No free ports in range 9200-9299
```

Recovering that line required grepping floci's container logs by hand and correlating timestamps: the real error was logged at 17:09:00.029 and the stack was deleted at 17:09:04.410, four seconds later. Anyone driving floci through a CDK/LZA pipeline — where the toolkit output is the only thing normally read — gets a dead end.

## Suggested fix

Keep the stack after rollback, in `ROLLBACK_COMPLETE`, with its stack events intact, and let `DeleteStack` remove it. That matches AWS, restores `DescribeStackEvents` as the diagnostic path, and lets the toolkit print the resource-level failure it already knows how to print.

If some existing behaviour depends on the stack being gone (a redeploy that would otherwise hit `AlreadyExistsException`), the AWS-faithful equivalent is to allow a create over a stack in `ROLLBACK_COMPLETE` by deleting it *at that point* — on the next create attempt, not eagerly at rollback time. That preserves the window in which the operator can actually read what happened.

## Severity

Moderate-to-high. No incorrect resource state results, but it removes the primary means of diagnosing any stack failure, and it converts every distinct root cause into the same uninformative message.

PR #3104, opened 2026-09-05
fix(cloudformation): refuse UpdateStack on a ROLLBACK_COMPLETE stack

### LINKED ISSUE #3097: [BUG] CloudFormation: UpdateStack on a ROLLBACK_COMPLETE stack is accepted instead of refused
## Service

CloudFormation

## AWS API Action

`UpdateStack` (and `CreateChangeSet` with `ChangeSetType=UPDATE`) on a stack whose status is `ROLLBACK_COMPLETE`

## Expected behavior

A stack whose creation failed and rolled back sits in `ROLLBACK_COMPLETE` and cannot be updated on AWS; it has to be deleted and created again. AWS refuses the update synchronously:

```
An error occurred (ValidationError) when calling the UpdateStack operation:
Stack:arn:aws:cloudformation:us-east-1:000000000000:stack/demo/<id> is in ROLLBACK_COMPLETE state and can not be updated.
```

moto emits the same message from `_validate_status` (`moto/cloudformation/responses.py:411`). The CDK and Terraform both key on this state: the CDK CLI tells the user to delete the stack, and a client that gets a 200 here instead proceeds against a stack AWS would have rejected.

## Actual behavior

Since #2970 (`a31a9599f`) Floci refuses an update while a stack is in any `*_IN_PROGRESS` state, with AWS's wording (`CloudFormationService.createChangeSet`). `ROLLBACK_COMPLETE` is not part of that guard, so the update is accepted and Floci re-provisions the resources of a stack AWS treats as terminal.

## Reproduction

```bash
# a template whose only resource fails, so the stack lands in ROLLBACK_COMPLETE
awslocal cloudformation create-stack --stack-name demo --template-body '{"Resources":{"B":{"Type":"AWS::S3::Bucket","Properties":{"BucketName":"INVALID_NAME"}}}}'
awslocal cloudformation describe-stacks --stack-name demo --query 'Stacks[0].StackStatus'   # ROLLBACK_COMPLETE
awslocal cloudformation update-stack --stack-name demo --template-body '{"Resources":{"B":{"Type":"AWS::S3::Bucket"}}}'
# expected: ValidationError "... is in ROLLBACK_COMPLETE state and can not be updated."
# actual: the update is accepted
```

## Notes

Follow-up promised in the #2970 review. The fix is one more condition on the guard #2970 added, plus a test beside `CloudFormationPipesCleanupIntegrationTest` pinning the message. `DeleteStack` on a `ROLLBACK_COMPLETE` stack must keep working (#2207 covers that path).

omatheusmesmo:
## Summary

Closes #3097. Follow-up promised in the #2970 review.

A stack whose creation failed sits in `ROLLBACK_COMPLETE` holding nothing but its name, and AWS refuses to update it — only `DeleteStack` frees the name. #2970 added a guard that refuses an update while a status ends in `_IN_PROGRESS`; `ROLLBACK_COMPLETE` was not part of it, so Floci accepted the update and re-provisioned a stack AWS treats as terminal. The CDK CLI keys on this state and tells the user to delete the stack; a client that gets a 200 here instead proceeds against something real AWS would have rejected.

The guard now covers both grounds, and `UpdateStack` and `CreateChangeSet` with `ChangeSetType=UPDATE` share it, so a client cannot route around the refusal by building the change set itself.

**`ROLLBACK_COMPLETE` is matched exactly, not by suffix.** `UPDATE_ROLLBACK_COMPLETE` ends in the same eleven characters and means the opposite: an update failed over a stack that is still live and intact, and retrying the update is how it is repaired. Refusing it would strand every stack whose update ever failed. `updateStaysAllowedAfterAnUpdateRolledBack` pins that.

## StackSets

Adding the condition alone regressed `UpdateStackSet`: an instance whose stack rolled back is `INOPERABLE`, and deploying into it would raise the refusal out of the loop and fail the **whole call with a 400**, updating no instance at all. That is not what AWS does — it leaves such an instance alone, updates the rest, and reports the operation `FAILED`.

`updateStackSet` now pre-checks the instance's stack status through the same predicate rather than deploying and catching, so nothing is swallowed: only a stack that genuinely refuses an update is skipped, and a real validation error still surfaces as a 400. Verified by reverting only `StackSetService` with the guard in place — the new StackSets test then fails with `Expected status code <200> but was <400>`.

## AWS Compatibility

**Incorrect behavior:** `UpdateStack` on a `ROLLBACK_COMPLETE` stack returned 200 and re-provisioned the stack's resources. Real AWS refuses it synchronously:

```
An error occurred (ValidationError) when calling the UpdateStack operation:
Stack:arn:aws:cloudformation:us-east-1:000000000000:stack/demo/<id> is in
ROLLBACK_COMPLETE state and can not be updated.
```

moto emits the same string from `_validate_status` (`moto/cloudformation/responses.py`). The message is reproduced exactly, down to AWS's own "can not" spelling and the `Stack:<arn>` with no space, since clients match on it.

`DeleteStack` on a `ROLLBACK_COMPLETE` stack keeps working — the #2207 path — and `deleteStackStillClearsARollbackCompleteStack` covers it: the stack goes away and its name is free for a fresh `CreateStack`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Tests

New `CloudFormationRollbackCompleteUpdateIntegrationTest`, four cases: `UpdateStack` refused with the pinned message, `CreateChangeSet` refused on the same terms, `DeleteStack` still clearing the stack, and `UPDATE_ROLLBACK_COMPLETE` still updatable. Plus `updateStackSetReportsFailedRatherThanRefusingOverARolledBackInstance` in `CloudFormationStackSetsIntegrationTest`.

Run against the unfixed service, the three that should fail do; the two control cases pass either way, which is the point of including them.

Full CloudFormation package: 1176 tests, 1 failure — `SamTransformIntegrationTest#samFunction_withAutoPublishAlias_createsVersionAndAlias`, a Docker-backed Lambda invocation that fails identically on a clean `main` in this environment.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
@@ -167,6 +167,38 @@ public class CloudFormationService implements ResourceProvider {
         return stack != null ? stack.getParameters() : Map.of();
     }
 
+    /**
+     * The status of a stack, or {@code null} when no stack of that name exists in the region.
+     * Unlike {@link #describeStacks}, asking about a stack that is not there is not an error:
+     * StackSet deployment uses this to decide what to do with an instance before it acts on it.
+     */
+    String stackStatus(String stackName, String region) {
+        Stack stack = resolveStack(stackName, region);
+        return stack != null ? stack.getStatus() : null;
+    }
+
+    /**
+     * Whether a stack in this status refuses an update, on the two grounds AWS refuses one.
+     *
+     * <p>A status ending in {@code _IN_PROGRESS} says an operation owns the stack: a create, an
+     * update, a rollback or the cleanup phase of a committed update. AWS refuses to start a second
+     * one over it, and offers nothing that finishes a phase whose process is gone - DeleteStack is
+     * the only way out of one. Refusing keeps every abandoned phase out of the next update's
+     * transaction.
+     *
+     * <p>{@code ROLLBACK_COMPLETE} is terminal for a different reason: a create that failed rolled
+     * its resources back, so the stack holds the name and nothing else, and only DeleteStack frees
+     * it. The CDK CLI keys on this state and tells the user to delete the stack; a client that is
+     * accepted here instead proceeds against a stack AWS would have rejected. It is matched exactly
+     * rather than by suffix, because {@code UPDATE_ROLLBACK_COMPLETE} ends the same way and is a
+     * perfectly updatable stack: it is where an update that failed and rolled back settles, and
+     * retrying the update is how it is repaired.
+     */
+    static boolean refusesUpdate(String status) {
+        return status != null
+                && (status.endsWith("_IN_PROGRESS") || "ROLLBACK_COMPLETE".equals(status));
+    }
+
     // ── CreateChangeSet ───────────────────────────────────────────────────────
 
     public ChangeSet createChangeSet(String stackName, String changeSetName, String changeSetType,
@@ -293,17 +325,10 @@ public class CloudFormationService implements ResourceProvider {
                     throw new AwsException("AlreadyExistsException",
                             "Stack [" + stackName + "] already exists", 400);
                 }
-                // A status ending in _IN_PROGRESS says an operation owns the stack: a create, an
-                // update, a rollback or the cleanup phase of a committed update. AWS refuses to
-                // start a second one over it, and offers nothing that finishes a phase whose
-                // process is gone - DeleteStack is the only way out of one. Refusing here keeps
-                // every abandoned phase out of the next update's transaction.
-                //
                 // The message is the one real CloudFormation emits, down to its own "can not"
                 // spelling and the stack id carried as "Stack:<arn>" with no space: clients match
                 // on this string.
-                if (!isCreateType && existing.getStatus() != null
-                        && existing.getStatus().endsWith("_IN_PROGRESS")) {
+                if (!isCreateType && refusesUpdate(existing.getStatus())) {
                     throw new AwsException("ValidationError",
                             "Stack:" + existing.getStackId() + " is in " + existing.getStatus()
                                     + " state and can not be updated.", 400);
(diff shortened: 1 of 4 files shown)


hectorvent (approved):
Thank you for picking #3097 up so quickly, and for going one step further than the issue asked: I had the same guard change on a local branch and had not seen that `UpdateStackSet` would raise the refusal out of its loop and fail the whole call over one rolled-back instance. The pre-check through the shared `refusesUpdate` predicate is the right shape, the exact match against `ROLLBACK_COMPLETE` with `UPDATE_ROLLBACK_COMPLETE` pinned as updatable is exactly AWS's behaviour, and the message matches moto's and LocalStack's byte for byte. Built the merge with main locally: the five affected suites pass.

One note for later, not for this PR: Floci marks an instance whose backing stack failed as `INOPERABLE`, which the model reserves for a failed `DeleteStackInstances` or import; AWS leaves a failed create or update `OUTDATED` with `DetailedStatus FAILED`. Your change follows the existing convention, so it belongs in a separate issue.

Merging, thank you.

omatheusmesmo:
> Thank you for picking #3097 up so quickly, and for going one step further than the issue asked: I had the same guard change on a local branch and had not seen that `UpdateStackSet` would raise the refusal out of its loop and fail the whole call over one rolled-back instance. The pre-check through the shared `refusesUpdate` predicate is the right shape, the exact match against `ROLLBACK_COMPLETE` with `UPDATE_ROLLBACK_COMPLETE` pinned as updatable is exactly AWS's behaviour, and the message matches moto's and LocalStack's byte for byte. Built the merge with main locally: the five affected suites pass.
> 
> One note for later, not for this PR: Floci marks an instance whose backing stack failed as `INOPERABLE`, which the model reserves for a failed `DeleteStackInstances` or import; AWS leaves a failed create or update `OUTDATED` with `DetailedStatus FAILED`. Your change follows the existing convention, so it belongs in a separate issue.
> 
> Merging, thank you.

Thank you!! :1st_place_medal:

Issue #1634, opened 2026-06-29
[BUG] CloudFormation delete-stack fails after persistent restore when ECS task definition is missing


robertjosephwayne:
## Service

CloudFormation, ECS

## AWS API Action

`cloudformation delete-stack` / `cloudformation wait stack-delete-complete` for a stack containing an `AWS::ECS::TaskDefinition`

Related ECS checks:

- `ecs describe-task-definition`
- `ecs list-task-definitions`
- `ecs describe-services`

## Expected behavior

After restoring Floci persistent runtime state, CloudFormation and ECS state should remain consistent. If a CloudFormation stack still tracks an `AWS::ECS::TaskDefinition`, either:

- ECS should still be able to describe that task definition, and CloudFormation should delete it successfully, or
- CloudFormation delete should treat an already-missing task definition as delete-complete.

The stack should reach `DELETE_COMPLETE`.

## Actual behavior

After restoring persistent Floci state, CloudFormation still had a stack with an `AWS::ECS::TaskDefinition` physical resource id, but ECS no longer had that task definition or the related ECS service. Deleting the stack failed with `DELETE_FAILED`.

Stack status:

```json
{
  "StackStatus": "DELETE_FAILED",
  "StackStatusReason": "The following resource(s) failed to delete: [ServiceTaskDefinition]"
}
```

CloudFormation event:

```json
{
  "LogicalResourceId": "ServiceTaskDefinition",
  "PhysicalResourceId": "arn:aws:ecs:us-east-2:000000000000:task-definition/example-service-family-local:2",
  "ResourceType": "AWS::ECS::TaskDefinition",
  "ResourceStatus": "DELETE_FAILED",
  "ResourceStatusReason": "Unable to describe task definition: arn:aws:ecs:us-east-2:000000000000:task-definition/example-service-family-local:2"
}
```

ECS direct checks showed the task definition and service were missing:

```json
{
  "taskDefinitionArns": []
}
```

```text
An error occurred (ClientException) when calling the DescribeTaskDefinition operation: Unable to describe task definition: arn:aws:ecs:us-east-2:000000000000:task-definition/example-service-family-local:2
```

```json
{
  "services": [],
  "failures": []
}
```

## Reproduction

```bash
# Start Floci with persistent storage.
docker run --rm \
  -p 4566:4566 \
  -e FLOCI_STORAGE_MODE=persistent \
  -e FLOCI_STORAGE_PERSISTENT_PATH=/app/data \
[...]

PR #3114, opened 2026-09-05
refactor(cloudformation): move AWS::ECS::Cluster, TaskDefinition and Service into EcsCfnProvisioner

hectorvent:
## Summary

Moves `AWS::ECS::Cluster`, `AWS::ECS::TaskDefinition` and `AWS::ECS::Service` out of the legacy
`CloudFormationResourceProvisioner` switch into a new `EcsCfnProvisioner`, next to the existing
capacity-provider one. The ECS-only parsers and the three idempotent delete helpers move with the
arms, and the monolith no longer depends on `EcsService` at all (its constructor loses that
parameter; `CfnProvisionerFixture` passes one argument less).

Behaviour is unchanged except where the move fixes it: an unnamed cluster keeps the name its first
execution generated instead of creating a second cluster on every `UpdateStack`; a service's
create-vs-update branches on the provision context and the stored `Name` attribute rather than the
resource's mutable id, and a renamed service is created as a replacement; a non-integer
`DesiredCount` is a `ValidationError` instead of an uncaught `NumberFormatException`. Deletes go
through `CfnDeletes.safeDelete` tolerating only the service's own not-found codes, so a
`ClusterContainsTasksException` still fails the stack delete (#1634).

Part of the per-service provisioner migration (AGENTS.md, "Adding a CloudFormation Resource Type").

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No emulated wire surface changes. `Ref` and `Fn::GetAtt` are unchanged and match the registry
schemas' read-only properties: `Arn` on the cluster, `TaskDefinitionArn` on the task definition,
`Name` and `ServiceArn` on the service. `CloudFormationIntegrationTest` asserts them end to end.

## Checklist

- [x] `./mvnw test` passes locally (the whole CloudFormation package: 1,169 tests, the only failure being the pre-existing `CfnSchemaCoverageTest` ACM `Id` row that fails on main too with the schema corpus present; `make docs-check` and the sdk-test-java compat suite green)
- [x] New or updated integration test added (`EcsCfnProvisionerTest`; the existing ECS stack tests
      keep asserting the exact `Fn::GetAtt` keys)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -46,7 +46,6 @@ import io.github.hectorvent.floci.services.ec2.Ec2Service;
 import io.github.hectorvent.floci.services.kinesis.KinesisService;
 import io.github.hectorvent.floci.services.kinesis.model.KinesisStream;
 import io.github.hectorvent.floci.services.ec2.model.Tag;
-import io.github.hectorvent.floci.services.ecs.EcsService;
 import io.github.hectorvent.floci.services.firehose.FirehoseService;
 import io.github.hectorvent.floci.services.firehose.model.DeliveryStreamDescription;
 import io.github.hectorvent.floci.services.rds.RdsService;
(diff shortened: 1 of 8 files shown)


pgermosen (approved):
Confirmed the cluster-change-update fix directly: provisionService now requires both the name and the cluster to match before taking the in-place path, a cluster change correctly falls through to create+replace. The rollback finding is real but pre-existing across the whole provisioner migration, not something this PR regresses, the interface itself documents the no-op default gracefully. And the test naming finding doesn't hold up, EcsCfnProvisionerTest matches the actual convention used by all 30+ other provisioner tests, not the service-test one greptile applied. Approving.

PR #3115, opened 2026-09-05
refactor(cloudformation): move the ElasticLoadBalancingV2 types into ElbV2CfnProvisioner

hectorvent:
## Summary

Moves `AWS::ElasticLoadBalancingV2::LoadBalancer`, `TargetGroup`, `Listener` and `ListenerRule`
out of the legacy `CloudFormationResourceProvisioner` switch into a new `ElbV2CfnProvisioner`,
together with the ELBv2-only helpers (ELB-safe name generation, full-name derivation, matcher,
certificates, action and rule-condition parsing). The monolith no longer depends on `ElbV2Service`
(its constructor loses that parameter; `CfnProvisionerFixture` passes one argument less).

Behaviour is unchanged except where the move fixes it:

- an unnamed load balancer or target group keeps the name its first execution generated, read back
  from the `LoadBalancerName` / `TargetGroupName` attribute since the physical id is the ARN; the
  switch generated a fresh name on every `UpdateStack` and left the previous resource behind;
- `Listener` and `ListenerRule` branch create-vs-update on the provision context, and a listener
  moved to another load balancer or a rule moved to another listener (both create-only) is created
  as a replacement instead of being modified under the old parent; one removed out of band is
  created again;
- a non-integer `Port` or `Priority` is a `ValidationError` instead of an uncaught
  `NumberFormatException`.

Deletes call the service directly on purpose: `ElbV2Service` already returns silently for an
unknown ARN, and its refusals (`ResourceInUse`, `OperationNotPermitted`) must keep failing the
stack delete.

Part of the per-service provisioner migration (AGENTS.md, "Adding a CloudFormation Resource Type"),
independent of #3114 (ECS).

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No emulated wire surface changes. `Fn::GetAtt` follows the registry schemas' read-only
properties: the five load balancer attributes, `ListenerArn`, `RuleArn` and `IsDefault` are as
before, and `TargetGroup` gains `LoadBalancerArns` (comma-joined, empty until a listener forwards
to the group), which the switch never set. `CloudFormationIntegrationTest` keeps asserting the
stack outputs end to end.

## Checklist

- [x] `./mvnw test` passes locally (the whole CloudFormation package: 1,169 tests, the only failure
      being the pre-existing `CfnSchemaCoverageTest` ACM `Id` row that fails on main too with the
      schema corpus present; `make docs-check` and the sdk-test-java compat suite green)
- [x] New or updated integration test added (`ElbV2CfnProvisionerTest`; the existing ELBv2 stack,
      listener-update and no-orphans cases keep asserting the outputs)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -58,13 +58,6 @@ import io.github.hectorvent.floci.services.rds.model.DbSubnetGroup;
 import io.github.hectorvent.floci.services.eks.EksService;
 import io.github.hectorvent.floci.services.eks.model.CreateClusterRequest;
 import io.github.hectorvent.floci.services.eks.model.Nodegroup;
-import io.github.hectorvent.floci.services.elbv2.ElbV2Service;
-import io.github.hectorvent.floci.services.elbv2.model.Action;
-import io.github.hectorvent.floci.services.elbv2.model.Listener;
-import io.github.hectorvent.floci.services.elbv2.model.LoadBalancer;
-import io.github.hectorvent.floci.services.elbv2.model.Rule;
-import io.github.hectorvent.floci.services.elbv2.model.RuleCondition;
-import io.github.hectorvent.floci.services.elbv2.model.TargetGroup;
 import io.github.hectorvent.floci.services.iam.IamService;
 import io.github.hectorvent.floci.services.iam.model.IamRole;
 import io.github.hectorvent.floci.services.kms.KmsService;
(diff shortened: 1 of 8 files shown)


hectorvent:
Thanks for the rounds. Everything raised is in: rollback of a replacement, cleanup in the template's dependency order, an orphaned replacement kept owed after a failed rollback delete, and entities with attempts left surviving a give-up. On the deletes, I have routed the four through `CfnDeletes.safeDelete` with the not-found code `ElbV2Service` defines for each type; the service's delete paths return silently on an unknown ARN today, so behaviour is unchanged, and `ResourceInUse` and `OperationNotPermitted` still fail the stack delete. The cleanup now names one entity per resource for both its progress and its failure event, and the unused parameter is gone.

pgermosen (approved):
Went through the rollback/cleanup-ordering fixes in ReplacementCleanup carefully since that utility is shared with Cognito and ECS too, and they hold up generally, not just for this case. The replacement integration test doing a real create/update/rollback cycle through the whole engine is exactly the kind of coverage this needed. Good migration.

Issue #2313, opened 2026-08-14
[FEAT] ram: Resource Access Manager (resource shares and associations)


pgermosen:
## Service

RAM (Resource Access Manager) — new service

## API Action / Feature

Resource shares, associations and invitations

## AWS Documentation

- https://docs.aws.amazon.com/ram/latest/APIReference/Welcome.html
- https://docs.aws.amazon.com/ram/latest/APIReference/API_CreateResourceShare.html
- https://docs.aws.amazon.com/ram/latest/APIReference/API_AssociateResourceShare.html

## Why is this needed?

There is no `services/ram` package, so every RAM action fails. `aws_ram_resource_share`,
`aws_ram_resource_association`, `aws_ram_principal_association` and
`aws_ram_resource_share_accepter` are all unsupported.

RAM is how a resource is shared across accounts — most commonly a Transit Gateway, a Route 53
Resolver rule, a subnet, or a License Manager configuration. Without it, the cross-account half
of those topologies cannot be expressed.

## Priority

**Lowest of the new-service tracking issues, deliberately.** RAM only matters for
*cross-account* sharing. A single-account local run never exercises it, and Floci's typical
usage is single-account. It is filed for completeness rather than urgency, and it is a
reasonable one to leave open indefinitely.

The one place it becomes load-bearing is cross-account Transit Gateway attachment acceptance
(`AcceptTransitGatewayVpcAttachment` on a shared gateway) — so it is worth doing *after*
Transit Gateway, if at all, and not before.

## Scope

This is a tracking issue for a new service package. Suggested split:

**1. Resource shares**
`CreateResourceShare`, `GetResourceShares`, `UpdateResourceShare`, `DeleteResourceShare`.
Backs `aws_ram_resource_share`.

**2. Associations**
`AssociateResourceShare`, `DisassociateResourceShare`, `GetResourceShareAssociations`.
Backs `aws_ram_resource_association` and `aws_ram_principal_association`. Note that resource and
principal associations are the *same* action distinguished by which parameter is set
(`resourceArns` vs `principals`), while `GetResourceShareAssociations` separates them by
`associationType` — a shape worth getting right rather than modelling as two actions.

**3. Invitations** (only for genuine multi-account emulation)
`GetResourceShareInvitations`, `AcceptResourceShareInvitation`,
`RejectResourceShareInvitation`. Backs `aws_ram_resource_share_accepter`.

[...]

PR #3120, opened 2026-09-06
feat(ram): support resource-share invitation accept/reject

pgermosen:
## Summary

Closes the last open piece of #2313. `services/ram` already covers resource shares and associations (PR #2579); the one thing left was invitation accept/reject: `GetResourceShareInvitations` was hardcoded to return an empty list and there was no `Accept`/`RejectResourceShareInvitation` at all, so `aws_ram_resource_share_accepter` had nothing to work against.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the RAM API reference and botocore's `ram/2018-01-04/service-2.json` (not a live SDK/CLI call, no AWS credentials in this environment): request/response shapes for `GetResourceShareInvitations`/`Accept`/`RejectResourceShareInvitation`, the `ResourceShareInvitationStatus` enum, and the modeled error list (`OperationNotPermittedException`, `ResourceShareInvitationAlready{Accepted,Rejected}Exception`, `ResourceShareInvitationArnNotFoundException`, `MalformedArnException`). `GetResourceShareInvitations` is receiver-only per the reference text ("invitations that you have received"), not sender+receiver as I originally assumed.

- Real `ResourceShareInvitation` records: created for a bare AWS account-id principal added while organization sharing is off. An organization/OU principal never gets one (matching real AWS's own auto-accept for org-based sharing), and neither does an account principal once `EnableSharingWithAwsOrganization` has been called.
- Deliberately not a visibility gate: `GetResourceShares`/`ListResources` under `OTHER-ACCOUNTS` keep showing every non-owned share regardless of invitation status, same as before this change. floci's launched-container credentials all resolve to one default account, so gating on acceptance would hide shares from the very consumers real AWS invited, the existing tests for that flow depend on it.
- Invitation creation and accept/reject are each serialized behind a `synchronized(this)` scoped to just this new code, closing two races (duplicate invitations from concurrent associates on the same principal; a double-accept both succeeding), plus a lost-update race where two concurrent associates on the same share could overwrite each other's principal.

Not covered: time-based `EXPIRED` transitions, and whether accepting/rejecting should fail once the underlying share is deleted (AWS's own error list for these operations has nothing modeled for that case, so there's nothing verified to build against).

## Checklist

### DIFF
--- a/docs/services/ram.md
+++ b/docs/services/ram.md
@@ -7,8 +7,9 @@
 Floci implements the AWS Resource Access Manager calls AWS Landing Zone Accelerator makes
 while bootstrapping shared Transit Gateway attachments and other cross-account resources:
 the organization-sharing opt-in, resource-share CRUD, association management, principal
-listing, and tagging. Sharing is modeled as a flat in-memory store, not full RAM semantics —
-see Current Scope below for what's simplified.
+listing, tagging, and invitation accept/reject (backing `aws_ram_resource_share_accepter`).
+Sharing is modeled as a flat in-memory store, not full RAM semantics: see Current Scope
+below for what's simplified.
 
 ## Supported Actions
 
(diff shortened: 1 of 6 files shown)


hectorvent (approved):
Looks right against the model: the three paths and every error code are the declared ones, and limiting invitations to bare account principals while organization sharing is off is the honest reading of how AWS auto-accepts the rest. The lock around accept/reject closes the double-accept window properly.

No blockers from my side.

PR #3127, opened 2026-09-06
fix(lambda): bound ZIP extraction resources

nkootstra:
## Summary

Bound Lambda ZIP extraction so compressed packages, expanded content, and entry counts cannot exhaust local disk, inodes, or CPU during function and layer deployment.

The existing package is preserved when extraction fails. The entry-count guard defaults to 100,000 entries and can be changed with `FLOCI_SERVICES_LAMBDA_ZIP_MAX_ENTRIES`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Lambda ZIP extraction previously had no expanded-size or entry-count bound. A small compressed archive could expand into a very large tree or contain enough entries to exhaust local resources.

This change applies AWS-aligned package limits:

- Direct inline ZIP uploads over 50 MiB return `RequestEntityTooLargeException` with HTTP 413.
- ZIP extraction is limited to 250 MiB of cumulative expanded content.
- A single entry is limited to 250 MiB.
- S3-backed archives are bounded at 250 MiB compressed before extraction.
- Failed extraction is staged and discarded, so the previous deployment remains intact.

AWS documents deployment-package size limits but does not define a ZIP entry-count quota. The default of 100,000 entries is therefore a Floci safety bound, not an AWS protocol value. It is high enough for normal Lambda packages while bounding inode use and traversal work for archives containing many tiny files. The `FLOCI_SERVICES_LAMBDA_ZIP_MAX_ENTRIES` setting allows environments with unusually large, trusted package layouts to raise the limit.

Direct oversized uploads use AWS's `RequestEntityTooLargeException`. Expanded-size and entry-count failures use the existing `InvalidParameterValueException` response path.

Focused validation passed:

- `./mvnw test -Dtest=ZipExtractorTest,LambdaServiceTest,LambdaLayerIntegrationTest`
- 113 tests passed

## Checklist

### DIFF
--- a/docs/configuration/advanced/application-yml.md
+++ b/docs/configuration/advanced/application-yml.md
@@ -138,6 +138,7 @@ floci:
       runtime-api-base-port: 12000            # Port range for Lambda Runtime API
       runtime-api-max-port: 12499             # One port per running container = concurrency ceiling
       code-path: ./data/lambda-code           # Where ZIP archives are stored
+      zip-max-entries: 100000                  # Maximum ZIP entries extracted per deployment package
       poll-interval-ms: 1000
       container-idle-timeout-seconds: 300     # Remove idle containers after this
       region-concurrency-limit: 1000          # Concurrent executions ceiling per region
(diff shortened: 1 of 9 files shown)


pgermosen (approved):
Checked the package size limits against the real Lambda quotas page, 50 MB compressed and 250 MB unzipped both match exactly. More importantly, traced the actual defense: the per-entry and cumulative caps aren't just checked against the ZIP's declared size, the streaming copy loop bounds the real decompressed output byte by byte, which is what actually stops a zip bomb since declared sizes can lie. Path traversal is guarded properly too, and the staged-extraction-then-swap means a failed extract never touches the live deployment. Nice, careful work. Approving.

PR #3136, opened 2026-09-06
fix(lambda): give each published version its own code directory

### LINKED ISSUE #2958: [BUG] lambda: published versions share one code directory with $LATEST, so a redeploy changes what a version runs

trh-ds:
> @avison9 — you're assigned to #2958 and said you'd take it, so apologies if this duplicates work already in flight. Happy to close this in favour of yours, or to hand over any part of it; the diagnosis in your issue is what this is built on. The reclaim-pass details below are the only part that goes beyond what you'd already worked out.

## Summary

Fixes #2958 — a published Lambda version's metadata was immutable, but the code directory it named was not.

`CodeStore.getCodePath` had no version component, so every version shared one directory with `$LATEST`, and `publishVersion` copied only the `codeLocalPath` *pointer*. A later `UpdateFunctionCode` rewrote the very directory each published version pointed at, leaving a version advertising one `CodeSha256` while the bytes behind it were a different build — and `ContainerLauncher` tar-copies that directory at launch, so a qualified invoke ran `$LATEST`'s code.

**The fix — content-addressing.** The deployment package is hashed up front and extracted to `<account>/<function>/<hex sha256>`. A published version's `codeLocalPath` therefore names the directory for the exact bytes it was published with, and a redeploy of `$LATEST` lands in a *different* directory rather than overwriting it. Identical packages resolve to the same directory, so redeploying unchanged code reuses storage instead of duplicating it. `publishVersion` needs no change: the pointer it copies is now stable by construction.

**Reclaiming superseded builds.** Since a changed deploy no longer overwrites in place, old builds would otherwise accumulate one directory per historical deploy, forever. A reclaim pass (run under the same per-function lock as `publishVersion`, on code update and version delete) drops directories no version still references — including `$LATEST`.

That pass deliberately matches **hash-shaped names only**, which is load-bearing in two ways I hit while building it:

- `ZipExtractor` stages and retires trees as `<hash>.floci-staging-<uuid>` / `<hash>.floci-previous-<uuid>` *siblings* of its target, so they live in the same directory — and extraction runs outside the lock, so a concurrent deploy of the same function can have one live mid-flight. An unfiltered sweep would delete it out from under that deploy.
- A function deployed before this change has its code sitting directly in the function directory, possibly still named by a version published before the upgrade. An unfiltered sweep would strip that live version's subdirectories (its `node_modules`, say) while leaving its loose files — corrupting it. There is a regression test for exactly this, and I confirmed it fails without the filter.

## Test plan

- [x] `LambdaServiceTest.publishedVersionCodeDirectorySurvivesLatestRedeploy` — publish v1, redeploy `$LATEST`, assert the version's directory still holds v1's bytes and `$LATEST` extracted elsewhere. **Verified it fails on the pre-fix code** with `expected: <v1> but was: <v2>`, i.e. it reproduces the reported bug.
- [x] `LambdaServiceTest.redeployingLatestWithoutPublishingReclaimsTheSupersededContentDirectory` — no version references the old build, so it must not leak. Also verified failing pre-fix.
- [x] `LambdaServiceTest.reclaimLeavesThePreContentAddressingLayoutAlone` — migration safety. Verified failing without the hash-shape filter.
- [x] `CodeStoreTest` — distinct hashes are distinct sibling directories, equal hashes coincide, whole-function delete still removes every build under it, and `deleteContentDirectory` removes one build only and refuses paths outside the store.
- [x] Full lambda unit suite green: `CodeStoreTest`, `LambdaServiceTest`, `LambdaAccountScopedCodeTest`, `LambdaServicePersistenceTest` — 83 tests, 0 failures.
- [ ] `LambdaVersionIntegrationTest.publishedVersionKeepsRunningItsOwnCodeAfterLatestIsRedeployed` — added, but **not run locally**: it needs a container runtime and Docker was unavailable on my machine. This is the same measurement gap the issue author flagged, now covered at the disk level by the unit test above. Worth a look from CI.

Note: it asserts only the version-qualified invoke, not a follow-up `$LATEST` invoke on the same function — the warm pool is keyed by function name alone (#1988), so a second invoke could reuse the first's container and make this test pass or fail on that unrelated bug.

Two pre-existing `ZipExtractorTest` failures show up on Windows (`\` is a real path separator there; `setPosixFilePermissions` is unsupported on NTFS). Both are in a file this PR does not touch.

https://claude.ai/code/session_01XvQwz2o9EStW3wFwJJxGyg

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -75,6 +75,8 @@ public class LambdaService implements ResourceProvider {
             "arn:(aws[a-zA-Z-]*)?:iam::\\d{12}:role/?[a-zA-Z_0-9+=,.@\\-_/]+");
     private static final Pattern HANDLER_PATTERN = Pattern.compile("\\S+");
     private static final int MAX_HANDLER_LENGTH = 128;
+    /** Shape of the hex SHA-256 directory names {@link CodeStore#getCodePath} creates. */
+    private static final Pattern CONTENT_DIRECTORY = Pattern.compile("[0-9a-f]{64}");
 
     private final LambdaFunctionStore functionStore;
     private final LambdaExecutorService executorService;
@@ -774,9 +776,10 @@ public class LambdaService implements ResourceProvider {
         synchronized (lockForConcurrencyOp(fn.getFunctionArn())) {
             warmPool.drainEnvironment(version.get());
             functionStore.deleteVersion(region, name, qualifier);
-            // The snapshot shares $LATEST's code directory, so this only reclaims once no
-            // remaining version references it.
+            // Reclaims a directory (legacy or content-addressed) only once no remaining
+            // version, including $LATEST, still references it.
             reclaimLegacyCodeDirectoryIfUnused(name);
+            reclaimUnusedCodeVersionDirectories(region, fn);
         }
         LOG.infov("Deleted version {0} of Lambda function: {1}", qualifier, name);
     }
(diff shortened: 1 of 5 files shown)


pgermosen:
Went through this closely, there's a bit to work through here.

**Blocking, before anything else:** the PR description has an AI-attribution line at the bottom ("Generated with Claude Code" plus a session link). Floci's contribution rules ban that in the PR body, not just commit trailers, so that needs to come out.

**Two real correctness bugs, both confirmed directly in the code:**

- The null-path crash is real: `reclaimUnusedCodeVersionDirectories` does `Path.of(fn.getCodeLocalPath())` for the $LATEST function with no null check, while the loop right below it correctly skips a null `codeLocalPath` for every other version. An Image or hot-reload function has no local ZIP code, so deleting a published version of one hits this and throws, even though the deletion itself already completed. That's a misleading failure response and makes retries unsafe.
- The concurrent-update race is real too: `extractZipCodeBytes` calls `zipExtractor.extractTo(...)` and sets `fn.setCodeLocalPath(...)` *before* acquiring the per-function lock, only the reclaim pass itself is locked. Two concurrent `UpdateFunctionCode` calls on the same function can each extract to their own hash directory unlocked, and one's reclaim pass can delete the other's directory before it's registered anywhere the reclaim logic can see it.

**Conflicts with main, and this one needs real care, not a pick-a-side resolution:** `LambdaService.java` has a hunk where main (already merged since this branch was cut) wraps the same `codeLocalPath`/`codeSizeBytes`/`codeSha256` field mutations in the `lockForConcurrencyOp` lock, closing a check-then-act race on `PublishVersion`'s `CodeSha256` precondition, and also calls `extractTo` with a third `maxEntries` argument that doesn't exist on this branch (from the zip-extraction-limits PR that landed after this one branched). Taking this branch's side as-is would silently drop both of those fixes and won't even compile against the current `extractTo` signature. The other conflict, in `LambdaServiceTest.java`, is just two large but independent blocks of new tests landing at the same spot, safe to keep both.

**Smaller stuff:** there are em dashes in the new Javadoc in `LambdaService.java`, `CodeStore.java`, and `LambdaVersionIntegrationTest.java` that need to go.

One more thing worth knowing: the description mentions this might overlap with someone else already assigned to the same issue (#2958), so you two may want to sync up on who's driving this fix.

trh-ds:
﻿Closing this. #3041 landed on main and closes #2958, so the bug this PR targets is already fixed.

@pgermosen, thanks for the close read. Worth flagging for anyone finding this later: #3041 merged at 19:07 UTC on 6 Sep, about four hours after your review, so the conflict analysis above describes the pre-#3041 main. The conflict is no longer just the `lockForConcurrencyOp` hunk and the `extractTo` signature, it is a full alternative implementation of the same fix.

The two approaches are not mergeable into one. This PR content-addresses `$LATEST` extraction into `<function>/<sha256>` and reclaims unreferenced directories with a sweep. #3041 copies the code into `<function>@versions/<version>` at publish time, and on a qualified delete removes exactly that one version's directory inside the lock it already holds. Both P1s found here, the null `codeLocalPath` crash and the reclaim racing concurrent updates, are properties of that sweep. #3041 has no sweep, so it has neither bug, and there is nothing left to port across.

#3041 also handles a case I did not think through: `FunctionName` has no charset validation in floci today, so the `@versions` suffix is what stops a version directory from colliding with a real function's own directory. A flat `<name>.v<n>` sibling, which is roughly where my layout would have ended up, does not hold that property.

One note a maintainer may want to pick up separately: #3041's tests assert on disk, and there is no invoke-level test on main covering a version-qualified invoke after a `$LATEST` redeploy. This PR had one in `LambdaVersionIntegrationTest`, written against the REST API only, so it is not tied to either implementation. I never got to run it because Docker was unavailable on my machine. Happy to send it as a standalone PR if that gap is worth filling.

Apologies for the duplicated effort, @avison9. I should have checked in on #2958 before building rather than after.

PR #3137, opened 2026-09-06
refactor(ses): move three single-owner helpers off the SesService facade

okinaka:
## Summary

Follow-up to #2356, continuing the post-split cleanup (#2881, #2984, #3030, #3090): moves three
facade helpers whose ownership became unambiguous after the store split, each to its single
natural home:

- `validateConfigurationSet` becomes `SesConfigurationSetService.validateForSending`: its only
  dependency was the configuration-set domain itself (existence plus the sending-paused gate), so
  unlike the cross-domain tracking/delivery validators it needs no callback and moves whole, with
  its v1/v2 exception-remap documentation.
- `templateNameFromArn` moves onto `SesTemplateService` as the same public static; the four static
  call sites in `SesQueryHandler` and `SesController` follow, and the Javadoc sentence that kept it
  on the facade is dropped.
- `stripXml10InvalidChars` (with its private `isXml10Char`) moves into `SesQueryHandler`, its sole
  production caller: sanitizing the v1 `TestRenderTemplate` XML response is a protocol concern,
  not domain logic. It stays package-private for the existing unit test.

No behavior change: wordings, gate order, and the ARN parsing are verbatim.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Pure code motion; no wire-visible change. The full SES suite (1046 tests) plus the native-image
runtime-initialization guard passes unchanged.

## Checklist

- [x] New or updated integration test added (not applicable: behavior-preserving refactor, existing tests pin the behavior and only their static references moved)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ses/SesConfigurationSetService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ses/SesConfigurationSetService.java
@@ -33,8 +33,9 @@ import java.util.regex.Pattern;
  * <p>The boundary follows the cross-domain seams: option validation that reads OTHER domains stays
  * in the facade (tracking options check a verified domain identity, delivery options check a
  * dedicated IP pool), which validates first (or resolves existence through {@link #get}) and then
- * mutates through {@link #save}. The ARN-dispatched configuration-set tagging lives here; the
- * facade keeps the send-path reads (pause check, event publishing, effective suppression reasons)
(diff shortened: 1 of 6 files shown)


pgermosen (approved):
Checked the three moved helpers directly against their old locations, all byte-for-byte identical logic and error handling, just relocated to their natural owner. Clean refactor. Approving.

PR #3139, opened 2026-09-06
fix(cloudwatch): fail Logs Insights queries with unsupported syntax instead of silently dropping it

### LINKED ISSUE #2042: [BUG] Logs Insights silently changes the result set for unsupported query syntax

omatheusmesmo:
## Summary

Closes #2042.

`LogsInsightsQuery` ignored any filter expression or command it could not parse (e.g. `like`, `=~`, `stats`, `parse`), logging a server-side warning and continuing as if that stage had never been in the query. The client had no way to tell "the filter matched nothing" from "the filter was silently dropped": `StartQuery`/`GetQueryResults` reported a normal `Complete` status with a result set that did not correspond to the query it asked for.

The mis-parsed-operator case (`>=`, `<=`, `=~` losing their comparison character to the equality scan and matching a mangled field/value instead of warning) was already fixed on main separately; this PR addresses the remaining case: syntax the engine recognizes as unsupported but still silently drops.

Unsupported syntax is now detected while parsing the query (`LogsInsightsQuery.isUnsupported()` / `getUnsupportedReason()`). `StartQuery` still succeeds (a query with syntax errors is a valid request shape), but the query reports `GetQueryResults` status `Failed` once it would otherwise have completed, matching a real AWS `QueryStatus` value, rather than a `Complete` status with an incorrect result set.

## Checklist

- [x] `CloudWatchLogsInsightsQueryTest`, `CloudWatchLogsServiceTest`, `CloudWatchLogsHandlerTest`, `CloudWatchLogsInputConstraintTest` pass (146 tests, 0 failures)
- [x] Added/updated regression tests: unsupported commands and unsupported filter operators now fail the query with a `Failed` status and a reason instead of being ignored; existing supported-syntax tests (`fields`, `filter`/`where` with `=`/`!=`/`==`, `sort`/`order`, `dedup`, `limit`) still pass unchanged

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudwatch/logs/CloudWatchLogsService.java
@@ -731,15 +731,21 @@ public class CloudWatchLogsService implements ResourceProvider {
 
     // ──────────────────────────── Logs Insights Queries ────────────────────────────
 
-    /** A query's status and (once Complete) its projected rows — the AWS GetQueryResults shape. */
+    /**
+     * A query's status and (once Complete) its projected rows: the AWS GetQueryResults shape.
+     * {@code failureReason} is {@code null} unless {@code status} is {@code Failed}; it is not
+     * part of the AWS wire response (GetQueryResults carries no such field) but is exposed here
+     * for callers and tests that want to know why an unsupported query failed.
+     */
     public record QueryState(String status, List<LinkedHashMap<String, String>> rows,
-                             long recordsScanned, long recordsMatched) {}
+                             long recordsScanned, long recordsMatched, String failureReason) {}
 
     /** Lifecycle state of a stored Insights query; {@link #label()} is the AWS wire form. */
     private enum InsightsQueryStatus {
         RUNNING("Running"),
         COMPLETE("Complete"),
-        CANCELLED("Cancelled");
+        CANCELLED("Cancelled"),
+        FAILED("Failed");
 
         private final String label;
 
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Thank you. Failing the query is the honest answer: a `Complete` result that silently dropped a `stats` or a `like` filter cannot be told apart from an empty match. `Failed` is in the model's `QueryStatus` enum, and keeping the reason off the wire since `GetQueryResults` has no such field is right.

No blockers from my side.

PR #3140, opened 2026-09-06
fix(cognito): invoke the CustomMessage Lambda trigger before delivering verification emails

### LINKED ISSUE #3018: [BUG] Cognito: user pool LambdaConfig.CustomMessage is stored and returned by DescribeUserPool but never invoked
## Service

Cognito

## AWS API Action

`CreateUserPool` / `UpdateUserPool` (`--lambda-config CustomMessage=<arn>`), `DescribeUserPool`, `SignUp`

## Expected behavior

When a user pool is configured with a `CustomMessage` Lambda trigger, Cognito should invoke that function before delivering a verification, reset or attribute-change message, and send the `response.emailSubject` / `response.emailMessage` the function returns.

Failing that, Floci should reject or visibly ignore the unsupported configuration, so that a user can tell the trigger is not in effect.

This is the same principle already accepted in #1483 for the `CUSTOM_AUTH` triggers: a trigger configuration that is silently not honoured "can hide integration or security issues during local testing".

## Actual behavior

Floci **accepts and persists** `LambdaConfig.CustomMessage` — `DescribeUserPool` faithfully returns the ARN — and then **never invokes the function**. The signup email is Cognito's built-in default message. There is no error, no warning, and no log entry indicating the trigger was skipped.

Because the configuration round-trips correctly, a test that asserts *"the user pool has a CustomMessage trigger configured"* passes while the feature is entirely inert. The failure is only visible by inspecting the delivered message.

Note that `CUSTOM_AUTH` triggers (`DefineAuthChallenge`, `CreateAuthChallenge`, `VerifyAuthChallengeResponse`) **are** dispatched (#1483), so the Cognito→Lambda dispatch path already exists — `CustomMessage` appears not to be wired to it. The Cognito service documentation does not list `LambdaConfig` or any user-pool trigger, so this may be intended-but-undocumented rather than a regression; if so, having `CreateUserPool` reject the field, or documenting it as stored-and-ignored, would be enough to remove the trap.

## Reproduction

Version: `floci/floci:latest`, `_floci/info` reports `{"version":"2.0.1","edition":"community"}`.
Container started with the Lambda prerequisites from the README:

```bash
docker run -d --name floci -p 4566:4566 \
  -u root \
  floci/floci:latest
```

**1. A minimal CustomMessage function.** `handler.py`:

```python
def handler(event, context):
    event.setdefault("response", {})["emailSubject"] = "TRIGGER-FIRED"
    event["response"]["emailMessage"] = "TRIGGER-FIRED code={####}"
    return event
```

```bash
zip fn.zip handler.py

  --function-name custom-message \
  --runtime python3.12 --handler handler.handler \
  --role arn:aws:iam::000000000000:role/lambda-role \
  --zip-file fileb://fn.zip
```

**2. Confirm the function itself works** (it does — this step passes):

```bash
echo '{"triggerSource":"CustomMessage_SignUp","response":{}}' > ev.json
  --function-name custom-message --payload fileb://ev.json out.json
cat out.json
# {"triggerSource":"CustomMessage_SignUp","response":{"emailSubject":"TRIGGER-FIRED", ...}}
```

**3. Create a pool wired to it, and confirm the config round-trips:**

```bash
ARN=$(aws --endpoint-url http://localhost:4566 lambda get-function \
  --function-name custom-message --query 'Configuration.FunctionArn' --output text)

POOL=$(aws --endpoint-url http://localhost:4566 cognito-idp create-user-pool \
  --pool-name triggerpool \
  --username-attributes email --auto-verified-attributes email \
  --lambda-config CustomMessage=$ARN \
  --query 'UserPool.Id' --output text)

aws --endpoint-url http://localhost:4566 cognito-idp describe-user-pool \
  --user-pool-id $POOL --query 'UserPool.LambdaConfig'
# { "CustomMessage": "arn:aws:lambda:us-east-1:000000000000:function:custom-message" }   <-- stored
```

**4. Sign a user up:**

```bash
CLIENT=$(aws --endpoint-url http://localhost:4566 cognito-idp create-user-pool-client \
  --user-pool-id $POOL --client-name c1 --no-generate-secret \
  --query 'UserPoolClient.ClientId' --output text)

  --client-id $CLIENT --username spike@example.com --password 'Spike@Test1' \
  --user-attributes Name=email,Value=spike@example.com
```

**5. Read the delivered message:**

```bash
curl -s http://localhost:4566/_aws/ses
```

Observed:

```json
{"Subject": "Your verification code",
 "Body": {"text_part": "Your verification code is 246962.", "html_part": null}}
```

Expected `TRIGGER-FIRED`. The function was not called — `describe-log-streams` for `/aws/lambda/custom-message` shows only the one stream created by the direct `invoke` in step 2, and none from the signup.

omatheusmesmo:
## Summary

Closes #3018

`LambdaConfig.CustomMessage` was accepted, persisted, and echoed back by `DescribeUserPool`, but the function was never invoked. Every signup/resend/forgot-password message used Cognito's built-in default text, with no error or log entry indicating the trigger was skipped.

## What changed

Real Cognito invokes `CustomMessage` before delivering a message and uses the trigger's `response.emailSubject` / `response.emailMessage` (or `response.smsMessage`) instead of the default. This follows the same dispatch pattern already used for the `CUSTOM_AUTH` triggers (#1483):

- `CognitoAuthFlowHandler.fireCustomMessage(...)` invokes the `CustomMessage` Lambda with `codeParameter` (`{####}`) and `usernameParameter`, mirroring the real event shape. It is non-blocking: an unconfigured or failing trigger falls back to the pool's default templated message, so a broken function cannot brick signup (matching how `PostAuthentication`/`PostConfirmation` already tolerate trigger errors).
- `CognitoMessageDispatcher.dispatch(...)` gained an overload that takes the trigger's response and prefers its subject/body/SMS text over the pool's `VerificationMessageTemplate`, still substituting the real code for `{####}`.
- Wired into `SignUp` (`CustomMessage_SignUp`), `ResendConfirmationCode` (`CustomMessage_ResendCode`) and `ForgotPassword` (`CustomMessage_ForgotPassword`), all of which already had the app client in scope needed to build the trigger event.

**Scoped out:** `GetUserAttributeVerificationCode` (`CustomMessage_UpdateUserAttribute`/`CustomMessage_VerifyUserAttribute`). That path only has an access token, not the app client id needed for the trigger's `callerContext`, and the issue's own reproduction is the signup flow, so I kept the fix to the three flows that share the same call shape and are directly testable end to end.

## Checklist

- [x] `CognitoLambdaTriggersTest` (new `CustomMessage` tests reproducing the issue's scenario, plus non-blocking-on-error and not-configured cases): 35/35 pass
- [x] `CognitoMessageDispatcherTest`: 7/7 pass
- [x] `CognitoServiceTest` (updated one existing mock stub for the new dispatch overload): 206/206 pass
- [x] `CognitoJsonHandlerTest`: 17/17 pass

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoAuthFlowHandler.java
@@ -41,6 +41,7 @@ final class CognitoAuthFlowHandler {
 
     private static final Logger LOG = Logger.getLogger(CognitoAuthFlowHandler.class);
     private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final String CUSTOM_MESSAGE_CODE_PARAMETER = "{####}";
 
     private final CognitoService service;
     private final LambdaService lambdaService;
@@ -852,6 +853,28 @@ final class CognitoAuthFlowHandler {
                 Boolean.TRUE.equals(resp.get("autoConfirmUser")),
                 Boolean.TRUE.equals(resp.get("autoVerifyEmail")),
                 Boolean.TRUE.equals(resp.get("autoVerifyPhone")));
+    }
+
+    /**
+     * Fires the CustomMessage trigger, if configured, so the function can override the
+     * subject/body Cognito would otherwise deliver. Non-blocking: an unconfigured or
+     * failing trigger falls back to the pool's default templated message, mirroring how
+     * {@link #firePostAuthentication} and {@link #firePostConfirmation} tolerate errors.
+     */
+    Map<String, Object> fireCustomMessage(UserPool pool, UserPoolClient client, CognitoUser user,
+                                           String triggerSource) {
+        Map<String, Object> req = new HashMap<>();
+        req.put("codeParameter", CUSTOM_MESSAGE_CODE_PARAMETER);
+        req.put("usernameParameter", user.getUsername());
+        req.put("clientMetadata", Map.of());
+        TriggerResult result = invokeTrigger(pool, client, user, "CustomMessage", triggerSource, req);
+        if (!result.configured()) return null;
+        if (result.errored()) {
+            LOG.warnv("CustomMessage trigger failed for pool {0} (source {1}): {2}",
+                    pool.getId(), triggerSource, result.errorMessage());
+            return null;
+        }
+        return result.response();
     }
 
     private CognitoService.ClaimsOverride firePreTokenGeneration(UserPool pool, UserPoolClient client, CognitoUser user,
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -2258,8 +2258,10 @@ public class CognitoService implements ResourceProvider {
             try {
                 String code = verificationCodeService.issue(pool.getId(), user.getUsername(),
                         VerificationCode.Purpose.SIGNUP_CONFIRMATION, Duration.ofHours(24));
+                Map<String, Object> customMessage = authFlowHandler.fireCustomMessage(
+                        pool, client, user, "CustomMessage_SignUp");
                 messageDispatcher.dispatch(pool, user, VerificationCode.Purpose.SIGNUP_CONFIRMATION,
-                        code, List.of(deliveryTarget.deliveryMedium()));
+                        code, List.of(deliveryTarget.deliveryMedium()), customMessage);
             } catch (VerificationCodeException e) {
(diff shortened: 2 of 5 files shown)


hectorvent (approved):
Thank you, this closes a silent gap: the trigger was accepted and echoed but never ran. Following the CUSTOM_AUTH dispatch pattern and letting the trigger's subject and body win over the pool template while still rendering `{####}` is the right shape.

**(follow-up, separate PR)** Two small things: `clientMetadata` is always empty where AWS forwards the request's `ClientMetadata` on these three calls, and the new one-line `if` statements want braces per AGENTS.md.

No blockers from my side.

PR #3141, opened 2026-09-06
fix(cloudcontrol): reject ListResources for unsupported resource types instead of returning empty

### LINKED ISSUE #2043: [BUG] Cloud Control ListResources returns an empty list for unsupported resource types

omatheusmesmo:
## Summary

Closes #2043

`CloudControlService.listResources` returned an empty list for any `TypeName`
outside the nine types it enumerates (S3 buckets, EC2 VPCs/subnets/security
groups/instances/launch templates, IAM roles/users/instance profiles). That
empty result was indistinguishable from "supported type, zero resources," so
a real but unbacked type (e.g. `AWS::SQS::Queue`, `AWS::Logs::LogGroup`) and a
type that does not exist in AWS at all both looked like a successful,
resource-free call. Tools that sweep a list of type names to discover
inventory silently missed real resources.

## Approach

Went with a combination of the issue's options 1 and 2:

- `ListResources` now throws `UnsupportedActionException` (HTTP 400) for any
  `TypeName` it does not enumerate, matching real Cloud Control API semantics
  for a resource type that does not support the requested action.
  `TypeNotFoundException` was considered but doesn't fit: it means the type
  isn't in the CloudFormation registry, and Floci has no such registry to
  check against, so it can't distinguish a real-but-unbacked type from an
  invented one. Both get the same error.
- `docs/services/cloudcontrol.md` already documents the supported type list;
  added a note describing the new error behavior.
- `GetResource` and the internal `CreateResource` read-back still work for
  types outside the enumerated list (falling back to create-time state), so
  this only changes `ListResources`'s own contract, not the read side as a
  whole.

Skipped option 3 (extending `ListResources` to cover SQS, CloudWatch Logs,
etc.) as a larger, separately-scoped change.

## Checklist

- [x] `CloudControlServiceTest` and `CloudControlIntegrationTest` pass
      (11/11), including new regression tests for a supported type, a real
      but unbacked type, and a nonexistent type

### DIFF
--- a/docs/services/cloudcontrol.md
+++ b/docs/services/cloudcontrol.md
@@ -34,6 +34,11 @@ Cloud Control.
   which is broader than the read side. A resource created through Cloud Control but
   outside the read-side type list is still readable via `GetResource`, from
   create-time state, but will not appear in `ListResources`.
+- **`ListResources` on an unlisted type**: `TypeName` values outside the list above
+  fail with `UnsupportedActionException` (HTTP 400) instead of returning an empty
+  `ResourceDescriptions`. This applies to any type Floci does not enumerate on the
+  read side, whether or not the type is real in AWS, since Floci has no CloudFormation
+  type registry to tell the two apart.
 - **Delete needs create-time state**: types whose delete depends on attributes
   captured at create time (currently `AWS::EKS::Nodegroup` and `AWS::IAM::Policy`, plus
   any `Custom::*` / `AWS::CloudFormation::CustomResource`) fail `DeleteResource` with a
--- a/src/main/java/io/github/hectorvent/floci/services/cloudcontrol/CloudControlService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudcontrol/CloudControlService.java
@@ -143,9 +143,12 @@ public class CloudControlService {
      */
     private String resourceModel(String region, String typeName, String physicalId, JsonNode desiredState) {
         try {
-            for (ResourceDescription d : listResources(region, typeName)) {
-                if (physicalId.equals(d.identifier())) {
-                    return d.properties();
+            List<ResourceDescription> listed = resourcesForType(region, typeName);
+            if (listed != null) {
+                for (ResourceDescription d : listed) {
+                    if (physicalId.equals(d.identifier())) {
+                        return d.properties();
+                    }
                 }
             }
         } catch (Exception ignored) {
@@ -206,9 +209,12 @@ public class CloudControlService {
 
     /** Cloud Control {@code GetResource}: a single resource from the read side, by identifier. */
     public ResourceDescription getResource(String region, String typeName, String identifier) {
-        for (ResourceDescription d : listResources(region, typeName)) {
-            if (d.identifier().equals(identifier)) {
-                return d;
+        List<ResourceDescription> listed = resourcesForType(region, typeName);
+        if (listed != null) {
+            for (ResourceDescription d : listed) {
+                if (d.identifier().equals(identifier)) {
+                    return d;
+                }
             }
         }
         // CreateResource provisions the whole CFN type set while the read side lists six types, so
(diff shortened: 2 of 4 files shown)


hectorvent (approved):
Thank you. Throwing `UnsupportedActionException`, which the model declares on `ListResources`, is the right way to make "not enumerated" distinguishable from "empty", and keeping `GetResource` on the nullable lookup means nothing else changed. The docs note makes the read-side scope clear.

No blockers from my side.

PR #3143, opened 2026-09-06
fix(apigateway): include apiKeyId in the REQUEST authorizer identity context

### LINKED ISSUE #3113: [BUG] apigateway: REST REQUEST authorizer event carries identity.apiKey but no identity.apiKeyId (follow-up to #807)
## Service

API Gateway (REST v1) — Lambda **REQUEST** authorizer event (follow-up to #807 / PR #1002)

## AWS API Action

Data-plane invocation of a REST method with `AuthorizationType: CUSTOM`, a `REQUEST` authorizer (`IdentitySource: method.request.header.x-api-key`) and `apiKeyRequired: true`.

## Expected behavior

Per the AWS docs for the REQUEST authorizer input ([Input to an API Gateway Lambda authorizer](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-input.html)) and the `$context.identity.apiKeyId` variable, `requestContext.identity` carries **both** the key value and the key id:

```json
"identity": {
  "apiKey": "<key value>",
  "apiKeyId": "<key id>",
  "sourceIp": "127.0.0.1",
  "userAgent": "curl/8.7.1",
  "accountId": null, "caller": null, "user": null, "userArn": null, "accessKey": null,
  "cognitoIdentityId": null, "cognitoIdentityPoolId": null, "cognitoAuthenticationType": null,
  "cognitoAuthenticationProvider": null, "principalOrgId": null, "clientCert": null
}
```

`apiKeyId` is the field authorizers actually use: the common pattern is `GetApiKey(apiKey: event.requestContext.identity.apiKeyId)` to read the key's tags / usage plan and build the principal. On AWS `GetApiKey` takes the **id**, not the value.

## Actual behavior

Floci 2.0.0 sends `apiKey` only. The authorizer receives:

```json
{"sourceIp":"127.0.0.1","userAgent":"curl/8.7.1","apiKey":"2895257d80a64f218f3cca32a6b8125d","clientCert":null}
```

`apiKeyId` is absent (as are the other null-valued AWS identity fields, which is harmless). A production authorizer that reads `identity.apiKeyId` gets `undefined` and denies every request → `403 {"message":"User is not authorized to access this resource"}` even with a valid, usage-plan-bound key.

Verified with our real authorizer bundle on 2.0.0: patching it to `identity.apiKeyId ?? identity.apiKey` is enough for the whole `x-api-key → REQUEST authorizer → AWS_PROXY Lambda → SQS FIFO` flow to pass, so this is the only remaining gap on that path.

Side note: Floci's `CreateApiKey` returns `id == value` (see the repro output), so the fix can simply set `apiKeyId` to the stored key id, and `null` for both when the header is missing or the key is unknown, as AWS does. The same `identity` object is reused for the AWS_PROXY integration event (`buildProxyEvent`), where AWS also sends `apiKeyId`.

## Reproduction

```bash
#!/usr/bin/env bash
set -euo pipefail
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_PAGER=""
EP="--endpoint-url http://localhost:4566 --region eu-west-3"
W=$(mktemp -d); cd "$W"
cat > index.js <<'EOF'
  console.log('IDENTITY ' + JSON.stringify(event.requestContext.identity));
  return { principalId: 'p', policyDocument: { Version: '2012-10-17', Statement: [{ Action: 'execute-api:Invoke', Effect: 'Allow', Resource: event.methodArn }] } };
};
EOF
zip -q fn.zip index.js
aws $EP lambda create-function --function-name echo-auth --runtime nodejs22.x --handler index.handler \
  --role arn:aws:iam::000000000000:role/lambda-role --zip-file fileb://fn.zip --query FunctionArn --output text
API=$(aws $EP apigateway create-rest-api --name repro --query id --output text)
ROOT=$(aws $EP apigateway get-resources --rest-api-id "$API" --query 'items[0].id' --output text)
RES=$(aws $EP apigateway create-resource --rest-api-id "$API" --parent-id "$ROOT" --path-part ping --query id --output text)
AUTH=$(aws $EP apigateway create-authorizer --rest-api-id "$API" --name a --type REQUEST \
  --identity-source method.request.header.x-api-key --authorizer-result-ttl-in-seconds 0 \
  --authorizer-uri "arn:aws:apigateway:eu-west-3:lambda:path/2015-03-31/functions/arn:aws:lambda:eu-west-3:000000000000:function:echo-auth/invocations" \
  --query id --output text)
aws $EP apigateway put-method --rest-api-id "$API" --resource-id "$RES" --http-method GET --authorization-type CUSTOM --authorizer-id "$AUTH" --api-key-required >/dev/null
aws $EP apigateway put-integration --rest-api-id "$API" --resource-id "$RES" --http-method GET --type MOCK --request-templates '{"application/json":"{\"statusCode\":200}"}' >/dev/null
aws $EP apigateway put-method-response --rest-api-id "$API" --resource-id "$RES" --http-method GET --status-code 200 >/dev/null
aws $EP apigateway put-integration-response --rest-api-id "$API" --resource-id "$RES" --http-method GET --status-code 200 >/dev/null
aws $EP apigateway create-deployment --rest-api-id "$API" --stage-name dev >/dev/null
KEY_ID=$(aws $EP apigateway create-api-key --name k --enabled --query id --output text)
KEY_VALUE=$(aws $EP apigateway get-api-key --api-key "$KEY_ID" --include-value --query value --output text)
PLAN=$(aws $EP apigateway create-usage-plan --name p --api-stages "apiId=$API,stage=dev" --query id --output text)
aws $EP apigateway create-usage-plan-key --usage-plan-id "$PLAN" --key-id "$KEY_ID" --key-type API_KEY >/dev/null
echo "key id=$KEY_ID value=$KEY_VALUE"
curl -s -w ' -> HTTP %{http_code}\n' "http://localhost:4566/restapis/$API/dev/_user_request_/ping" -H "x-api-key: $KEY_VALUE"
sleep 3
aws $EP logs filter-log-events --log-group-name /aws/lambda/echo-auth --filter-pattern IDENTITY --query 'events[-1].message' --output text
```

Output on `floci/floci:2.0.0`:

```
key id=2895257d80a64f218f3cca32a6b8125d value=2895257d80a64f218f3cca32a6b8125d
 -> HTTP 200
2026-09-05T22:05:21.914Z  b6557baf-...  INFO  IDENTITY {"sourceIp":"127.0.0.1","userAgent":"curl/8.7.1","apiKey":"2895257d80a64f218f3cca32a6b8125d","clientCert":null}
```

## Environment

- Floci version / image tag: `floci/floci:2.0.0`
- How you're running Floci: Docker (Colima, macOS arm64), `/var/run/docker.sock` mounted, `FLOCI_DEFAULT_REGION=eu-west-3`
- AWS CLI 2.27.22

omatheusmesmo:
## Summary

Closes #3113

For a REST API with a REQUEST authorizer and `apiKeyRequired: true`, real API Gateway populates both `requestContext.identity.apiKey` (the key value) and `requestContext.identity.apiKeyId` (the key's id) on the authorizer event. A REQUEST authorizer that needs to look up the key calls `GetApiKey` with the id, not the value, so `apiKeyId` is the field authorizers actually use.

Floci only set `identity.apiKey` and left `apiKeyId` entirely absent, so any authorizer reading `identity.apiKeyId` saw `undefined` and denied every request, even one carrying a valid, usage-plan-bound key.

## Fix

- `resolveApiKeyForRequest` now returns both the matched key's id and value (previously only the value), via a small `ResolvedApiKey` record.
- Both call sites that build the identity object (`toAuthorizerEvent` for the REQUEST authorizer invocation, and `buildProxyEvent` for the AWS_PROXY integration event) now set `apiKeyId` alongside `apiKey`, and set both to null when the header is missing or the key does not resolve.
- The id is taken from the usage plan key's own id rather than assumed equal to the value, so it stays correct even when a key was created with `generateDistinctId: true` (id different from value).

## Checklist

- [x] Regression tests added: `ApiGatewayRequestAuthorizerIntegrationTest` (REQUEST authorizer path, including a case where `generateDistinctId` makes id and value differ) and `ApiGatewayApiKeyRevocationIntegrationTest` (AWS_PROXY path, plus enable/disable/delete revocation behavior for `apiKeyId`).
- [x] `make docs-check` passes.
- [ ] Full local suite: local runs hit intermittent `SocketTimeoutException` / Lambda warm-pool timeouts under this sandbox's Docker load. This reproduces on tests unrelated to this change (confirmed on `ApiGatewayAuthorizerContextIntegrationTest`, untouched by this diff), so it is environment flakiness, not a regression from this fix. The unit-level tests exercising the same code path (`ProxyEventIamIdentityTest`, `ApiGatewayExecuteControllerTest`) passed cleanly, along with the non-Lambda-invoking portions of the integration suites.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -358,7 +358,7 @@ public class ApiGatewayExecuteController {
         }
 
         // 1. Authorizer
-        String resolvedApiKey = resolveApiKeyForRequest(region, apiId, stageName, headers);
+        ResolvedApiKey resolvedApiKey = resolveApiKeyForRequest(region, apiId, stageName, headers);
 
         // AWS_IAM is verified before the CUSTOM authorizer path because it gates the request on the
         // caller's signature rather than on a Lambda's verdict, and a method carries one
@@ -415,7 +415,7 @@ public class ApiGatewayExecuteController {
                                  Stage stage,
                                  Integration integration, HttpHeaders headers,
                                  UriInfo uriInfo, byte[] body,
-                                 AuthorizerResult authorizerResult, String resolvedApiKey,
+                                 AuthorizerResult authorizerResult, ResolvedApiKey resolvedApiKey,
                                  ExecuteApiSigV4Authorizer.CallerIdentity iamIdentity) {
         String functionName = functionNameFromUri(integration.getUri());
         if (functionName == null) {
@@ -448,7 +448,7 @@ public class ApiGatewayExecuteController {
                                               String resourceId,
                                               Stage stage,
                                               MethodConfig method,
-                                              HttpHeaders headers, UriInfo uriInfo, String resolvedApiKey) {
+                                              HttpHeaders headers, UriInfo uriInfo, ResolvedApiKey resolvedApiKey) {
         if ("CUSTOM".equals(method.getAuthorizationType())) {
             String authorizerId = method.getAuthorizerId();
             if (authorizerId == null) {
@@ -589,7 +589,7 @@ public class ApiGatewayExecuteController {
                                      HttpHeaders headers, String region, String apiId, String stageName,
                                      String httpMethod, String requestPath,
                                      String resourcePath, String resourceId, Stage stage, UriInfo uriInfo,
-                                     String resolvedApiKey) {
+                                     ResolvedApiKey resolvedApiKey) {
         // Recover the trailing slash the JAX-RS {proxy} binding strips, so the authorizer sees
         // the same raw path the Lambda later receives from buildProxyEvent (AWS parity). Path
         // matching and path-parameter extraction keep using the normalized requestPath.
@@ -639,15 +639,17 @@ public class ApiGatewayExecuteController {
             ctx.put("requestId", UUID.randomUUID().toString());
             ctx.put("requestTimeEpoch", System.currentTimeMillis());
 
-            // identity.apiKey: resolve from usage plans linked to this (apiId, stage)
+            // identity.apiKey / identity.apiKeyId: resolve from usage plans linked to this (apiId, stage)
             ObjectNode identity = ctx.putObject("identity");
             identity.put("sourceIp", "127.0.0.1");
             String userAgent = headers.getHeaderString("User-Agent");
             identity.put("userAgent", userAgent != null ? userAgent : "");
             if (resolvedApiKey != null) {
-                identity.put("apiKey", resolvedApiKey);
+                identity.put("apiKey", resolvedApiKey.value());
+                identity.put("apiKeyId", resolvedApiKey.id());
             } else {
                 identity.putNull("apiKey");
+                identity.putNull("apiKeyId");
             }
             identity.putNull("clientCert"); // null when mTLS is not configured (Floci does not support mTLS)
         }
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Thank you, this is the field REQUEST authorizers actually use: they look the key up by id, so `identity.apiKey` alone denied everything. Carrying id and value together through one record and setting both on the authorizer and proxy events, null when nothing matched, is the right shape.

No blockers from my side.

PR #3144, opened 2026-09-06
fix(ec2): bound user data output capture

nkootstra:
## Summary

Bound EC2 Docker command output capture to a 2,048-byte UTF-8 tail. Noisy UserData and setup commands no longer grow an in-memory buffer without limit, while CloudWatch streaming remains unchanged.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Before this change, Floci retained all Docker exec output and truncated it only after the command completed. A noisy command could exhaust the heap before the 2,048-byte diagnostic tail was selected.

Floci now keeps only the final 2,048 bytes during capture, reports when truncation occurred, and avoids splitting UTF-8 characters at the start of the retained tail. This applies to UserData execution and shared container command diagnostics. CloudWatch streaming continues to receive every Docker output frame.

Validated with the related EC2 test suite: 359 tests passed.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2ContainerManager.java
@@ -30,6 +30,7 @@ import org.jboss.logging.Logger;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thank you. A bounded tail with the truncation notice is the right fix: the old buffer held everything until the command finished, so a noisy script could exhaust the heap before the 2048-byte cut. Aligning the tail to a UTF-8 boundary is a nice touch.

No blockers from my side.

PR #3145, opened 2026-09-06
fix(rds): support tagging DB proxy target groups

pgermosen:
## Summary

RDS's tag resolver (`RdsService.resolveTagHandle`) already covers DB instances, clusters, subnet groups, parameter groups, cluster parameter groups, option groups, and DB proxies, but fell through to "not yet implemented by Floci" for `target-group` ARNs, even though every DB proxy Floci creates gets its own default target group with a real ARN. This adds a `tags` field to `DbProxyTargetGroup` and wires `target-group` ARNs into `AddTagsToResource`/`RemoveTagsFromResource`/`ListTagsForResource`, the same way `db-proxy` ARNs already work.

Snapshot, reserved-instance, and event-subscription ARNs still fall through to the same "not implemented" message, since Floci doesn't model those resources at all yet; tagging them is a separate, larger effort.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

On a live account, `AddTagsToResource` operates on DB proxy target-group ARNs directly (it reports `DBProxyTargetGroupNotFoundFault` when one doesn't exist). Floci instead rejected every target-group ARN with `InvalidParameterValue: Tagging for resource type 'target-group' is not yet implemented by Floci`, even for a target group it had just created. `resolveTagHandle` now resolves the target group the same way it already resolves `db-proxy` ARNs (scan by ARN, confirm ownership through `findProxyTargetGroup`, persist the updated tag map back through the account-scoped store).

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/rds/RdsService.java
@@ -840,8 +840,30 @@ public class RdsService implements Resettable, ResourceProvider {
                     putOptionGroupForRegion(resourceId, effectiveRegion, group);
                 });
             }
-            // Valid RDS resource types Floci does not model yet (snapshot, ri, ...) — taggable
-            // on real AWS, so the message states the Floci limitation rather than AWS semantics.
+            case "target-group" -> {
+                DbProxyTargetGroup targetGroup = proxyTargetGroups.scan(k -> true).stream()
+                        .filter(candidate -> effectiveRegion.equals(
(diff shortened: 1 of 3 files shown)


hectorvent (approved):
Looks right: the `target-group` arm follows the `db-proxy` one, and `DBProxyTargetGroupNotFoundFault` is the declared code on all three tag operations. Carrying tags through `copyProxyTargetGroup` is the part that would have been easy to miss.

No blockers from my side.

PR #3147, opened 2026-09-06
fix(eventbridge,scheduler): read cron day-of-week as AWS 1-7 SUN-SAT

MaxFreedomPollard:
## Summary

Both schedule expression parsers read a numeric cron day-of-week with Unix numbering, 0-6 for SUN-SAT, but AWS numbers that field 1-7 for SUN-SAT. Every numeric day fired one day late, and Saturday could not be expressed at all.

`cron(0 12 ? * 2 *)` is Monday to AWS and fired on Tuesday. AWS's own documented example `cron(15 10 ? * 6L 2022-2023)`, described in the user guide as the last Friday of the month, fired on the last Saturday. `cron(0 12 ? * 7 *)` did not parse at all: cron-utils rejected it with `Value 7 not in range [0, 6]`. Day names were always right, because cron-utils resolves `MON` against its own numbering, so only expressions written with numbers were affected.

The cause is the cron definition in `ScheduleExpressionParser.java:32` (eventbridge) and `SchedulerExpressionParser.java:55` (scheduler). Both called `.withDayOfWeek()` with no range and no Monday value, which leaves cron-utils on its Unix default. Both now use `.withValidRange(1, 7).withMondayDoWValue(2)`, the same shape cron-utils ships for Quartz, whose day-of-week dialect AWS follows. `7` parses as Saturday, and `0` is rejected in line with AWS's documented `1-7` range.

One existing test had the old behavior baked into it. `getNextFireTimeFirstMondayOfMonth` parsed `cron(30 2 ? * 2#1 *)`, named itself for the first Monday, then asserted the fire day was a Tuesday. It now asserts `2026-03-02`, the first Monday of that month. `getNextFireTimeWeekdays` used `1-5`, which is SUN-THU in AWS numbering, so it now uses `2-6` and asserts the fire day really is a weekday.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behavior was the day-of-week mapping described above: numeric days shifted one day later, and `7` rejected as unparseable.

The field is documented as `1-7 or SUN-SAT` in the EventBridge Scheduler user guide, https://docs.aws.amazon.com/scheduler/latest/UserGuide/schedule-types.html, which also spells the mapping out for `#`: "3#2 would be the second Tuesday of the month: the 3 refers to Tuesday because it is the third day of each week". EventBridge Rules use the same six-field dialect. I checked this against the published field table and AWS's two worked examples, not against a live account.

## Checklist

Both parsers are pure functions with plain JVM unit tests, so the new coverage went there rather than into an integration test. I ran the affected suites rather than the whole build, because the full `./mvnw test` wants Docker for the Lambda, RDS and ElastiCache containers and there is no daemon on this machine.

`./mvnw test -Dtest='SchedulerExpressionParserTest,ScheduleExpressionParserTest,ScheduleDispatcherTest,SchedulerServiceTest,ScheduleInvokerTest,EventBridgeSchedulerIntegrationTest,EventBridgeServiceTest,EventBridgeInvokerTest,RulePersistenceTest,ReplayDispatcherTest,InputTransformerTest'` gives 253 tests, 0 failures. On the unmodified parsers the 13 new day-of-week tests and the one corrected test fail: 12 assertion failures plus 2 errors, both `Value 7 not in range [0, 6]`. `make docs-check` passes.

### DIFF
--- a/docs/services/scheduler.md
+++ b/docs/services/scheduler.md
@@ -29,7 +29,8 @@ background dispatcher fires schedule targets on time. Supported expressions:
   (default UTC) and `ActionAfterCompletion=DELETE`.
 - `rate(N unit)` — repeating fire (`minutes`, `hours`, `days`, `weeks`).
 - `cron(minute hour day-of-month month day-of-week year)` — AWS 6-field cron;
-  honors `ScheduleExpressionTimezone`.
+  honors `ScheduleExpressionTimezone`. Day-of-week is `1-7` or `SUN-SAT`, so
+  `2` is Monday and `cron(30 2 ? * 2#1 *)` is the first Monday of the month.
 
 `State=DISABLED` schedules and schedules outside their `StartDate`/`EndDate`
 window are skipped. The dispatcher ticks every
--- a/src/main/java/io/github/hectorvent/floci/services/eventbridge/ScheduleExpressionParser.java
+++ b/src/main/java/io/github/hectorvent/floci/services/eventbridge/ScheduleExpressionParser.java
@@ -29,7 +29,11 @@ public final class ScheduleExpressionParser {
                 .withHours().and()
                 .withDayOfMonth().supportsHash().supportsL().supportsW().supportsQuestionMark().and()
                 .withMonth().and()
-                .withDayOfWeek().supportsHash().supportsL().supportsW().supportsQuestionMark().and()
+                // AWS numbers day-of-week 1-7 as SUN-SAT, not the Unix 0-6 SUN-SAT that cron-utils
+                // defaults to, so Monday is 2. Without this every numeric day fires one day late and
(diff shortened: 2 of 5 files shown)


pgermosen (approved):
Checked the day-of-week mapping against AWS's own cron field table and the `3#2` example in the user guide — 1-7 SUN-SAT with Monday=2 is exactly right, and `.withValidRange(1, 7).withMondayDoWValue(2)` gets there. Reverted the fix locally and got the exact failures described (12 assertions + the two `Value 7 not in range [0, 6]` errors), so the tests are actually pinned to this behavior. Calendar dates in the new tests all check out too. Nice catch, especially the Saturday-can't-be-expressed-at-all part.

PR #3158, opened 2026-09-06
fix(test): tear down the organizations the service-managed StackSets test creates

hectorvent:
## Summary

Since #3055 merged, `main` fails one shard on `OrganizationsCfnIntegrationTest` with
`AlreadyInOrganizationException` (last green run: 9df197f; red at 41022d0, 7b5c6e2, d45c645).
`CloudFormationServiceManagedStackSetsIntegrationTest` created an organization under management
account `666666666666`, the account the Organizations CloudFormation test also uses, and never
deleted it. Organizations state is shared by every test in the JVM and keyed by the management
account, so the next class to create one under that account fails; which shard shows it depends on
class distribution (shards 2, 3 and 4 so far).

The StackSets test now records every management account it creates an organization under and
dismantles each after every test: member accounts out, organizational units bottom-up, then
`DeleteOrganization`. The one test that used `666666666666` uses an account of its own, like its
six siblings.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Test-only change; no emulated surface touched.

## Checklist

- [x] `./mvnw test` passes locally (both classes in one JVM in the order that failed, then the whole
      `cloudformation` package: 1,307 tests, the only failure being the pre-existing
      `CfnSchemaCoverageTest` ACM `Id` row that needs the local schema corpus)
- [ ] New or updated integration test added (the change is the teardown of an existing test class)

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationServiceManagedStackSetsIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationServiceManagedStackSetsIntegrationTest.java
@@ -4,9 +4,13 @@ import io.github.hectorvent.floci.testing.RestAssuredJsonUtils;
 import io.quarkus.test.junit.QuarkusTest;
 import io.restassured.path.json.JsonPath;
 import io.restassured.specification.RequestSpecification;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
+import java.util.LinkedHashSet;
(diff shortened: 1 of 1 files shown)


pgermosen (approved):
Nice catch — the leaked `666666666666` org from `activateOrganizationsAccessRequiresAllFeatures` lining up exactly with `OrganizationsCfnIntegrationTest`'s management account explains the shard-dependent flakiness perfectly. Teardown logic looks right: strips member accounts, deletes OUs bottom-up, then the org, and skips cleanly via the DescribeOrganization check for tests that never created one. Good fix.

Issue #2914, opened 2026-09-02
Native image size: quarkus-kubernetes-client is baked in even when the k8s Lambda executor is unused


pgermosen:
## Service

Lambda (kubernetes executor) / native image build

## Problem

The published native image jumped from ~120MB to ~135MB in one day (Aug 20-21), and it traces to `bc138011` ("feat(lambda): support kubernetes as the lambda runner", #1941), which added `quarkus-kubernetes-client` as a plain, unconditional dependency in `pom.xml`.

`KubernetesPodLauncher` is a normal `@ApplicationScoped` bean with no build-time guard, so GraalVM's native-image reachability analysis compiles the fabric8 client and its generated Kubernetes model classes into every build — even though `floci.services.lambda.executor` defaults to `docker` and (presumably) most users never set it to `kubernetes`.

`LambdaRuntimeLauncherProducer` already does the right thing at runtime — it looks up the launcher via `Instance<>` so the client object is never *instantiated* unless `executor: kubernetes` is configured. But that's a runtime optimization; it doesn't help native-image size, since reachability is decided statically at build time regardless of what actually runs.

Net effect: everyone pays the fabric8-client size cost in the shipped binary, whether or not they ever use the kubernetes executor.

## Suggested fix

Gate the kubernetes executor behind a build-time toggle instead of an always-on dependency — e.g. `@IfBuildProperty` on `KubernetesPodLauncher` (and whatever else pulls in `quarkus-kubernetes-client`), defaulting to off, so a normal build doesn't carry the fabric8 client's reachability closure at all. Anyone who wants the kubernetes executor opts in with a build property.

## Why is this needed?

This was the single largest contributor to a ~20% native image size increase over the past month (measured off the published Docker image sizes across that window). Shrinking it back is the highest-leverage single change available right now.

## Are you willing to contribute a PR?

- [ ] Yes
- [ ] No

PR #3160, opened 2026-09-06
fix(pipes): replace kafka-clients with a Kafka REST Proxy client

### LINKED ISSUE #2916: Native image size: kafka-clients is baked in even when Pipes never uses a Kafka source
## Service

Pipes (Kafka source) / native image build

## Problem

`kafka-clients` is a plain, unconditional dependency in `pom.xml`, added for exactly one thing: polling a self-managed Kafka source configured on an EventBridge Pipe (`PipesKafkaConsumerManager`, `PipesKafkaNativeSupport`, `PipesPoller`). `PipesKafkaConsumerManager` is a normal `@ApplicationScoped` bean with no build-time guard.

`kafka-clients` is a ~10MB jar on its own — the single largest dependency added for a feature this narrow anywhere in the tree, bigger even than the fabric8 kubernetes-client from #2914. `services.pipes.enabled` is a runtime toggle (`application.yml`), so it doesn't help here either: GraalVM decides native-image reachability at build time, before that config is ever read, so every build pays for `kafka-clients` whether or not anyone ever points a Pipe at Kafka.

## Suggested fix

Same shape as #2914: gate the Kafka pipe-source path behind a build-time toggle (`@IfBuildProperty`, or split into a separate Quarkus extension), defaulting to off, so a normal build doesn't carry `kafka-clients` and its transitive closure. Opt in with a build property if you actually need Kafka-sourced Pipes.

## Why is this needed?

Same underlying issue as #2914, different dependency: a feature only a subset of users touch is baked into every shipped binary at full weight. This is the heaviest single offender by jar size found so far.

- [ ] Yes
- [ ] No

pgermosen:
## Summary

Closes #2916

`kafka-clients` (plus the `zstd-jni` and `lz4-java` it pulls in transitively) was an always-on dependency for Kafka-sourced Pipes, reachable in every build regardless of whether a Pipe ever used a Kafka source, MSK-backed or self-managed. Rather than gating the existing kafka-clients-based code behind a build flag, this drops the dependency entirely: `kafka-clients-4.3.1.jar` (10,204,032 bytes) + `zstd-jni-1.5.6-10.jar` (7,305,999 bytes) + `lz4-java-1.10.1.jar` (910,232 bytes) = **18,420,263 bytes (~18.4 MB) removed from the dependency closure**, confirmed via `mvn dependency:tree` as exclusively brought in by `kafka-clients` (nothing else in the tree needs them). This applies to both native-image and JVM-mode builds, unconditionally, no opt-in flag required.

**Measured native-image impact:** built `docker/Dockerfile.native` for this branch and for unmodified `main`, same machine, same day, pushed both to a local registry and compared the real compressed layer sizes (`config.size` + sum of `layers[].size`):

| | Compressed size |
|---|---:|
| `main` (baseline) | 80,168,902 bytes |
| this branch | 74,035,435 bytes |
| **Reduction** | **6,133,467 bytes (~5.85 MiB / 6.13 MB), ~7.65% smaller** |

That's smaller than the 18.4 MB dependency-jar figure above, for two reasons: gzip compresses the delta further than a raw jar-size comparison would suggest, and GraalVM's native-image only ever compiled the portion of those jars actually reachable from `PipesKafkaConsumerManager`/`PipesKafkaNativeSupport` in the first place, so the full jar sizes were always an upper bound rather than the real contribution.

`PipesKafkaConsumerManager` and `PipesPoller`'s Kafka poll/deliver path now talk to a Karapace REST Proxy sidecar instead of embedding a native Kafka consumer. `KarapaceManager` starts one Karapace container on demand per distinct `bootstrap.servers` target (shared by every pipe that reads the same cluster), the same on-demand container-per-target pattern already used for MSK's Redpanda backing. `PipesKafkaRestClient` speaks Karapace's REST API with the JDK `HttpClient`, the same approach already used elsewhere in this codebase for other REST-shaped backends (e.g. the Kubernetes executor's REST client from #2914).

Kafka record fidelity (topic, partition, offset, timestamp, key, value, headers) is preserved end to end; `timestampType` is always reported as `CreateTime` since the REST Proxy reads the real value internally but does not expose it in its records response. `CreateTime` is Kafka's default for any topic that does not explicitly configure `message.timestamp.type`, so this only diverges from real behavior for a topic explicitly set to `LogAppendTime`.

Validated locally against a real Redpanda + Karapace pair in Docker (not just the fake-server unit tests), which caught two real issues before they shipped: Karapace returns a consumer's `base_uri` as a path relative to the proxy root rather than an absolute URL, and it rejects a bodyless offset-commit request instead of committing everything consumed, unlike the generic REST Proxy reference spec. Both are handled explicitly now.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Not applicable to the AWS API surface: `CreatePipe`/`UpdatePipe`/`DescribePipe` request and response shapes are unchanged, and the Kafka event record shape delivered to a Pipe's target (Lambda/SQS/etc.) is unchanged field-for-field except the documented `timestampType` deviation noted above. This replaces the transport underneath Kafka-sourced Pipes; the AWS-observable behavior is otherwise the same.

## Checklist

- [x] `./mvnw test` passes locally (full suite: 14,961 tests, 0 failures; the only 3 errors are pre-existing, unrelated flakiness in Neptune/RDS Data/SWF tests, confirmed against `git log` showing Neptune is already tracked as flaky in this repo)
- [x] New or updated integration test added (`PipesKafkaRestClientTest`, `KarapaceManagerTest`, updated `PipesKafkaConsumerManagerTest` and `PipesPollerTest`)


pgermosen:
@greptile review

pgermosen:
@greptile review

hectorvent (approved):
Thanks, this is the right cut: the 18.4 MB closure goes away without a build flag, and the sidecar path reuses `ContainerBuilder`, `ContainerStorageHelper` and `PortAllocator` the way the other managed containers do. I checked Greptile's five inline points against the current head and each is already handled: the localhost rewrite plus `withHostDockerInternalOnLinux()`, cleanup when `waitUntilReady` fails, closing the REST consumer when `subscribe` throws, and `blockedPartitions` keeping a failed partition from committing past the failure.

**(optional)** `docs/services/pipes.md` still lists only `FLOCI_SERVICES_PIPES_ENABLED`. The new `kafka-rest-bridge` block (image, host port range) is user-facing, and a line saying an `smk://` source now starts a sidecar container, so Docker is needed where it was not before, would save the next reader a surprise.

Approving as is.

pgermosen:
Added the docs update you suggested, docs/services/pipes.md now covers the kafka-rest-bridge config (image, host port range) and calls out that a Kafka-sourced pipe needs Docker for the sidecar. Thanks for the review.

PR #3163, opened 2026-09-07
feat(ses): implement v1 receipt filters (CreateReceiptFilter, ListReceiptFilters, DeleteReceiptFilter)

okinaka:
## Summary

Implements the three SES v1 receipt-filter actions as a stored-but-inert IP filter store alongside the receipt-rule domain: `CreateReceiptFilter`, `ListReceiptFilters`, and `DeleteReceiptFilter`. Filters round-trip through the management API but never filter an inbound connection, since Floci accepts none.

Validation follows real AWS as probed in us-west-2 (2026-09):

- A request omitting the whole `Filter` structure gets the single Smithy violation at `filter`; once any member is present, the filter name is a single service-level layer (no Smithy pattern, unlike rule and rule-set names): every malformed non-null shape gets `Not a valid filterName`, and an empty name its own message.
- `IpFilter` members are Smithy-enforced (not-null paths and the `[Allow, Block]` policy enum). `Cidr` accepts a single IPv4 or IPv6 address or an in-range CIDR block through a strict ASCII grammar: bracketed, zone-scoped, and IPv4-mapped IPv6 spellings, fullwidth Unicode digits, and out-of-range masks are all rejected with `Invalid CIDR block`, matching probes.
- Duplicates are `AlreadyExists` with the original kept, deletes are idempotent with only an absent `FilterName` rejected, and the 101st filter in a region is rejected with `LimitExceeded` "Too many filters". List responses are name-sorted for determinism since the probe could not pin the wire ordering.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Wire behavior probed against real AWS SES in us-west-2 with boto3 1.43.69, client-side validation disabled where needed (name and CIDR validation including exotic IPv6 and non-ASCII digit spellings, Smithy member paths, the bare-request `filter` violation, duplicate and idempotency semantics, the 100-filter limit). Verified with AWS SDK for Java v2 via the receipt-filter tests merged into `SesReceiptRuleSetTest` in `compatibility-tests/sdk-test-java`, and Terraform/OpenTofu compat coverage adds `aws_ses_receipt_filter`.

## Checklist


hectorvent (approved):
Thank you for the careful probe notes; they made this quick to check. Against the SES model: `AlreadyExists` and `LimitExceeded` are exactly what `CreateReceiptFilter` declares, list and delete declare nothing and the silent delete matches, the `ValidationError` violations mirror the required members of `ReceiptFilter` and `ReceiptIpFilter`, and the list XML follows the output shape. `InvalidParameterValue` for the name and CIDR checks is consistent with the rest of the SES v1 handler.

Approving as is.

PR #3164, opened 2026-09-07
refactor(cloudformation): move Cognito UserPool and UserPoolClient to CognitoCfnProvisioner

hectorvent:
## Summary

Batch 3 of the `CloudFormationResourceProvisioner` split: `AWS::Cognito::UserPool` and `AWS::Cognito::UserPoolClient` move into `CognitoCfnProvisioner`, which already owned `UserPoolDomain`. The monolith loses the two arms and its `CognitoService` dependency.

The move is the first commit and preserves behaviour. The following commits fix what reading the moved code against the registry schema and the service turned up, one defect each:

- a client whose `UserPoolId` or `GenerateSecret` (createOnly) changed was sent to `UpdateUserPoolClient`, which fails for a moved client and ignores the secret flag; it is now replaced, with the displaced client deleted once the update completes or restored on rollback
- `UserPoolTags` is resolved as the map the schema declares (the Key/Value list shape stays accepted)
- `Fn::GetAtt ProviderName` returns `cognito-idp.<region>.amazonaws.com/<pool id>` instead of the pool name; `ProviderURL` stays the local issuer, now noted in the docs table
- `UpdateUserPool` applies `PoolName` (a member of `UpdateUserPoolRequest`), so a stack rename takes effect; an unnamed pool keeps its generated name
- the non-schema `ClientName` attribute is dropped in favour of the schema's `Name`
- non-integer token validity values and a missing `UserPoolId` fail with `ValidationError` instead of being dropped

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Wire surface unchanged except `UpdateUserPool` honouring `PoolName`, which botocore declares on `UpdateUserPoolRequest`. Attribute shapes checked against `aws-cognito-userpool.json` and `aws-cognito-userpoolclient.json` (readOnly, createOnly, required) and CDK's `UserPool.fromUserPoolArn` for the `ProviderName` string. Templates that read `Fn::GetAtt Client.ClientName` (not an AWS attribute) now resolve the literal, as they would fail validation on AWS.

## Checklist

- [x] `./mvnw test` passes locally (CloudFormation and Cognito packages; the pre-existing ACM `Id` schema-gap row is the only failure, as on main)
- [x] New or updated integration test added (`CognitoCfnIntegrationTest` client replacement case; `CognitoCfnProvisionerTest` +24 cases; `CognitoServiceTest` rename case)

### DIFF
--- a/docs/services/cloudformation.md
+++ b/docs/services/cloudformation.md
@@ -90,7 +90,7 @@ cross-resource references.
 | CodePipeline | `Pipeline`, `CustomActionType`, `Webhook` |
 | CodeBuild | `Project` |
 | Batch | `ComputeEnvironment`, `JobQueue`, `JobDefinition` |
-| Cognito | `UserPool`, `UserPoolClient`, `UserPoolDomain` |
+| Cognito | `UserPool` (`ProviderURL` is the local issuer of the tokens Floci mints, `<base-url>/<pool id>`), `UserPoolClient`, `UserPoolDomain` |
 | ACM | `Certificate` |
 | EventBridge | `Rule`, `EventBus`, `EventBusPolicy` |
 | EventBridge Scheduler | `ScheduleGroup` |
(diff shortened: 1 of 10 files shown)


pgermosen (approved):
Nice split, and I like how each fix commit is scoped to one defect with the schema/service reasoning spelled out in the message. The deterministic-client-id guard ended up in a good place, going from the equal-ids case through the concurrent-create race to the orphaned-record exception without overcomplicating the happy path. Good to merge.

PR #3167, opened 2026-09-07
feat(sfn):  publish history events for Parallel branches and Map iterations

### LINKED ISSUE #2868: [FEAT] Step Functions: publish history events for Parallel branches and Map iterations
## Service

Step Functions

## API Action / Feature

Action: `GetExecutionHistory`
Feat: history events for the states inside a Parallel branch or a Map iteration

## AWS Documentation

https://docs.aws.amazon.com/step-functions/latest/apireference/API_GetExecutionHistory.html

## Why is this needed?

Floci publishes no history events for the states inside a Parallel branch or a Map iteration, so when one of them fails it is not easy to debug from error message.

Floci:

```
1 ExecutionStarted
2 ParallelStateEntered
3 ExecutionFailed        cause: The JSONPath '$.nope' specified for the field 'm.$' could not be found in the input '{"x":"a"}'
```

Real AWS:

```
1 ExecutionStarted
2 ParallelStateEntered   P
3 ParallelStateStarted
4 PassStateEntered       Fast
5 ExecutionFailed        cause: An error occurred while executing the state 'Fast' (entered at the event id #4). The JSONPath '$.nope' ...
```

The name `Fast` appears nowhere in Floci's output. With several branches of the same shape there is no way to work out which one failed.

Two deliverables, in order:

1. Publish the branch and iteration events. On AWS they interleave into the single execution history, verified in us-east-1 with two concurrent branches.
2. Add the cause prefix `An error occurred while executing the state '<name>' (entered at the event id #<n>).`. After step 1 the state is usually recoverable by reading back to the last `StateEntered`, so the prefix is what makes the attribution exact instead of a guess, and it is what makes the cause match AWS byte for byte.

The prefix is not on every failure, checked against real AWS:

| Failure | Prefix |
|---|---|
| Payload template path matches nothing | yes |
| JSONata expression returns nothing (`States.QueryEvaluationError`) | yes |
| State has no next state | yes |
| Fail state `Cause` | no, passes through |
| Task failure (`S3.NoSuchBucketException`) | no, passes through |

Step 2 also rewrites the cause assertions in `src/test/java/io/github/hectorvent/floci/services/stepfunctions/`, where 55 lines assert on a cause today. The JSONata side is missing the same prefix, already recorded as a deviation in `docs/services/step-functions.md`.

- [x] Yes
- [ ] No

exoego:
## Summary

Closes #2868

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

API surfaces are verified with real AWS in us-east-1.

- Events inside Parallel branches and inline Map iterations are now recorded in the parent execution's history.
- More event types are supported
- previousEventId chaining matches AWS. Each branch has its own chain. *StateSucceeded, MapRunSucceeded, MapIterationFailed and TaskStateAborted do not advance it. Interleaving order varies per run, the chain shape does not.
- Distributed Map items are child executions and publish nothing into the parent. Only MapRun* events appear.
- Branch events count toward the 25,000 event history limit.
- Interpreter failures carry the AWS cause prefix An error occurred while executing the state '<name>' (entered at the event id #<n>), added once at the innermost state.
- A Choice with no matching rule and no Default fails with States.Runtime and the AWS message, instead of States.NoChoiceMatched.
- A missing .$ payload template path fails with States.Runtime and the AWS wording. An array index past the end resolves to null
- A failed JSONata expression records EvaluationFailed with error, cause, location and state, once per retry attempt.

Remaining gaps are documented.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/StepFunctionsExecutionHistoryTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/StepFunctionsExecutionHistoryTest.java
@@ -126,6 +126,149 @@ class StepFunctionsExecutionHistoryTest {
         }
     }
 
+    /**
+     * Issue #2868: the states inside a Parallel branch and an inline Map iteration publish into the
+     * execution's history, chained through previousEventId to the Started event that opened them.
+     * The SDK reads each event's details under the field named for its type, which is what this
+     * test pins. Shapes verified against real AWS in us-east-1.
+     */
+    @Test
+    void parallelBranchesAndMapIterationsPublishTheirEvents() throws Exception {
+        var smDef = """
+                {
+                  "StartAt": "P",
+                  "States": {
+                    "P": {
+                      "Type": "Parallel",
+                      "Branches": [
+                        {"StartAt": "A1", "States": {"A1": {"Type": "Pass", "End": true}}}
+                      ],
+                      "Next": "M"
+                    },
+                    "M": {
+                      "Type": "Map",
+                      "ItemsPath": "$[0].items",
+                      "MaxConcurrency": 1,
+                      "ItemProcessor": {
+                        "ProcessorConfig": {"Mode": "INLINE"},
+                        "StartAt": "I1",
+                        "States": {"I1": {"Type": "Pass", "End": true}}
+                      },
+                      "End": true
+                    }
+                  }
+                }
+                """;
+
+        var smArn = sfn.createStateMachine(b -> b
+                .name(TestFixtures.uniqueName("sfn-history-branches-sm"))
+                .definition(smDef)
+                .roleArn(ROLE_ARN)).stateMachineArn();
+
+        try {
+            var execArn = sfn.startExecution(b -> b
+                    .stateMachineArn(smArn)
+                    .input("{\"items\": [1, 2]}")).executionArn();
+
+            var execution = pollUntilDone(execArn);
+            assertThat(execution.status()).isEqualTo(ExecutionStatus.SUCCEEDED);
+
+            var events = sfn.getExecutionHistory(b -> b
+                    .executionArn(execArn)
+                    .includeExecutionData(true)).events();
+
+            var types = events.stream().map(HistoryEvent::type).toList();
+            assertThat(types).containsExactly(
+                    HistoryEventType.EXECUTION_STARTED,
+                    HistoryEventType.PARALLEL_STATE_ENTERED,
+                    HistoryEventType.PARALLEL_STATE_STARTED,
+                    HistoryEventType.PASS_STATE_ENTERED,
+                    HistoryEventType.PASS_STATE_EXITED,
+                    HistoryEventType.PARALLEL_STATE_SUCCEEDED,
+                    HistoryEventType.PARALLEL_STATE_EXITED,
+                    HistoryEventType.MAP_STATE_ENTERED,
+                    HistoryEventType.MAP_STATE_STARTED,
+                    HistoryEventType.MAP_ITERATION_STARTED,
+                    HistoryEventType.PASS_STATE_ENTERED,
+                    HistoryEventType.PASS_STATE_EXITED,
+                    HistoryEventType.MAP_ITERATION_SUCCEEDED,
+                    HistoryEventType.MAP_ITERATION_STARTED,
+                    HistoryEventType.PASS_STATE_ENTERED,
+                    HistoryEventType.PASS_STATE_EXITED,
+                    HistoryEventType.MAP_ITERATION_SUCCEEDED,
+                    HistoryEventType.MAP_STATE_SUCCEEDED,
+                    HistoryEventType.MAP_STATE_EXITED,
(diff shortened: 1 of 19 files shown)


src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
```diff
@@ -2151,7 +2194,19 @@ private StateResult executeMapState(String name, JsonNode stateDef, JsonNode inp
         int effectiveConcurrency = effectiveMapConcurrency(
                 itemCount, requestedConcurrency, distributed);
 
+        chain.publish("MapStateStarted", Map.of("length", itemCount));
+        MapRunIdentity mapRun = null;
+        if (distributed) {
+            mapRun = newMapRunIdentity(stateDef, sm, context);
```
pgermosen, line 2201: This publishes MapRunStarted with a mapRunArn for every distributed Map, but recordMapRun (the only thing that actually persists a run for DescribeMapRun to find) only gets called from applyResultWriter's hasResultWriter branch. So a Distributed Map with no ResultWriter puts a mapRunArn in the history that DescribeMapRun will then reject with ResourceNotFound. Same finding Greptile raised below - confirmed it by tracing recordMapRun's only call site.

src/main/java/io/github/hectorvent/floci/services/stepfunctions/HistoryChain.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+        return false;
+    }
+
+    /** Returns the event id, or 0 once the execution has ended. */
+    long publish(String type, Map<String, Object> details) {
+        if (isAbandoned()) {
+            return 0L;
+        }
+        AslExecutor.countTowardsHistoryEventLimit(producedEventCount);
```
pgermosen: Agreeing with Copilot's comment on this one after tracing it through: countTowardsHistoryEventLimit runs before append() checks under the history lock whether the history is already ended, so a branch publishing right after another branch/failure ends the execution still inflates producedEventCount for an event that append() then discards. In a long-running execution with a lot of branch churn near the 25,000 limit that can throw States.Runtime for an execution that never actually recorded that many events.

src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
exoego: Addressed in https://github.com/floci-io/floci/pull/3167/commits/88d3224c75ef14c80a4090700f6ac6bdb0fd3a10

src/main/java/io/github/hectorvent/floci/services/stepfunctions/HistoryChain.java
exoego: Addressed in https://github.com/floci-io/floci/pull/3167/commits/15d26f082cffacfac7d541dfbbc6b78b1d1c91dd

compatibility-tests/sdk-test-java/src/test/java/com/floci/test/StepFunctionsExecutionHistoryTest.java
```diff
@@ -126,6 +126,149 @@ void taskEventsAreEmittedInOrderWithChainedPreviousEventId() throws Exception {
...
+            }
+
+            assertThat(events.get(8).mapStateStartedEventDetails().length()).isEqualTo(2);
+            assertThat(events.get(9).mapIterationStartedEventDetails().name()).isEqualTo("M");
+            assertThat(events.get(9).mapIterationStartedEventDetails().index()).isEqualTo(0);
+            assertThat(events.get(16).mapIterationSucceededEventDetails().name()).isEqualTo("M");
+            assertThat(events.get(16).mapIterationSucceededEventDetails().index()).isEqualTo(1);
+            assertThat(events.get(18).stateExitedEventDetails().output()).isEqualTo("[1,2]");
+        } finally {
+            try { sfn.deleteStateMachine(b -> b.stateMachineArn(smArn)); } catch (Exception ignored) {}
```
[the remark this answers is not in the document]
exoego: Ignoring uninterested exceptions are common pattern in tests.
Logging every exceptions messes up terminal outputs.

pgermosen:
Checked both fixes, they're right - recordMapRun now runs on every terminal path regardless of whether ResultWriter is configured, and the history-event counter only advances once append() has confirmed under the lock that the event isn't being dropped. Good catches on your end tracing through the concurrency on that second one.

This has conflicts with main now though, in AslExecutor.java - looks like it's overlapping with the JSONPath fix that landed there recently. Can you rebase and push?

exoego:
@pgermosen rebased 🙇

pgermosen (approved):
Good rebase, and nice handling of the overlap with the JSONPath fix that landed in the meantime - routing your context-reference case through the same resolvePayloadTemplateReference instead of keeping a separate implementation was the right call. Verified the two earlier fixes came through untouched. Approving.

PR #3168, opened 2026-09-07
fix(ec2): describe launchable unknown AMI ids in DescribeImages

matanbaruch:
## Summary

DescribeImages now returns an image for any launchable AMI id, so creating an aws.ec2.Instance against Floci with a real-world AMI id works.

Root cause: RunInstances already launches an unrecognised AMI id by falling back to the catalog default image (AmiImageResolver, plus the runInstancesKeepsX8664FallbackForUnknownImageAndType test), but DescribeImages returned an empty set for that same id. IaC providers read the AMI root device via DescribeImages before RunInstances whenever a root_block_device is set, so the empty result aborted the create with "collecting instance settings: empty result". The instance still launches, so a CLI describe-instances afterwards returns it, which is what made this look like a read race on DescribeInstances. It is not. The failing call is DescribeImages, before RunInstances runs.

Fix: DescribeImages builds a describable fallback image (state available, x86_64, /dev/xvda root device) for an explicitly requested, otherwise-unknown ami- id, mirroring the launch-time fallback. List queries with no image ids are unchanged, and a requested id that matches the catalog or a registered image is still served from there.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: DescribeImages returned an empty imagesSet for an explicitly requested `ami-` id that RunInstances would launch via the catalog fallback. The provider reads the AMI root device via DescribeImages before RunInstances when a root_block_device is set, so the empty result aborted instance create with "collecting instance settings: empty result".

Verified against the real pulumi-aws v5 provider (bridged terraform-provider-aws) driving `aws.ec2.Instance` with `ami-0c02fb55956c7d316` (not in the catalog): the create fails with the error above before the fix, and creates then destroys clean after. Floci-side, a new unit test `describeImagesResolvesUnknownLaunchableAmiId` fails on main and passes with the fix; full `Ec2ServiceTest` is green (126 tests).

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -3956,7 +3956,45 @@ public class Ec2Service implements ContainerTeardown, ResourceProvider {
                 .collect(Collectors.toList());
         List<Image> images = new ArrayList<>(catalogImages);
         images.addAll(createdImages);
+        addFallbackLaunchableImages(region, images, imageIds, owners, filters);
         return images;
+    }
+
+    // RunInstances launches an unrecognised AMI id by falling back to the catalog default image (see
+    // AmiImageResolver), so an explicitly requested but unknown id is still launchable. IaC providers read
+    // the AMI's root device via DescribeImages before RunInstances, so an empty result there aborts the
+    // create (the aws provider reports "collecting instance settings: empty result"). Mirror the launch-time
+    // fallback so a describe of a launchable id returns a matching image instead of nothing.
+    private void addFallbackLaunchableImages(String region, List<Image> images, List<String> imageIds,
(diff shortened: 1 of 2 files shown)


pgermosen:
@matanbaruch nice catch tracking this down to DescribeImages instead of the read race it looked like at first — the reasoning and the pulumi verification are solid. One thing to fix up: the description doesn't follow the PR template (missing the Type of change / AWS Compatibility / Checklist sections). Mind filling that in? Didn't turn up anything else.

matanbaruch:
@pgermosen done, filled in the Type of change, AWS Compatibility, and Checklist sections. Thanks for the review.

pgermosen (approved):
Template's filled in, thanks. Good to merge.

PR #3172, opened 2026-09-07
fix(ec2): carry DeliverLogsPermissionArn through CreateFlowLogs and DescribeFlowLogs

lex00:
## Summary

Ports lex00/floci#96.

A flow log with a CloudWatch Logs destination delivers through the IAM role given as `DeliverLogsPermissionArn`. Floci dropped the value on `CreateFlowLogs` and never returned it from `DescribeFlowLogs`. Terraform reads `aws_flow_log.iam_role_arn` back from describe, so it always came back empty and every plan proposed replacing the flow log, even right after an apply that passed a real ARN.

The Query handler and the `AWS::EC2::FlowLog` provisioner now pass the value into `FlowLogService.createFlowLog`, and `DescribeFlowLogs` returns it as `deliverLogsPermissionArn`. The provisioner treats a changed ARN on update the way it treats the other createOnly properties, since AWS replaces the flow log for it.

Tests cover the Query round trip through a real create and describe. The provisioner test checks the pass-through and the update rejection.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior. `DescribeFlowLogs` omitted `deliverLogsPermissionArn` for a flow log created with one. AWS returns it whenever the flow log was created with the parameter.

## Checklist

- [x] `./mvnw test` passes locally for the flow log service and provisioner tests plus the EC2 flow log integration test

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/Ec2FlowLogCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/Ec2FlowLogCfnProvisioner.java
@@ -63,6 +63,7 @@ public class Ec2FlowLogCfnProvisioner implements CfnResourceProvisioner {
                 ctx.resolveOptional(props, "TrafficType"),
                 ctx.resolveOptional(props, "LogDestinationType"),
                 ctx.resolveOptional(props, "LogDestination"),
+                ctx.resolveOptional(props, "DeliverLogsPermissionArn"),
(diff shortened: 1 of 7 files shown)


pgermosen (approved):
Good port, and nice touch treating the ARN change the same as the other replacement-only properties — checked that against the CFN reference and it's exactly right. The end-to-end create/describe test is a good way to pin this down.

PR #3175, opened 2026-09-07
chore(deps): update to Quarkus 3.39.2 and refresh pinned libraries

hectorvent:
## Summary

Supersedes #2906 (dependabot group at Quarkus 3.38.3, native build red).

Quarkus platform 3.37.4 -> 3.39.2 (BOM and plugin), Jackson 2.22.2, wire 6.4.7, bcprov 1.85.2, graphql-java 26.1, swagger-parser 2.1.48, assertj 3.27.7, neo4j-java-driver 6.2.1, maven-compiler-plugin 3.16.0, maven-surefire-plugin 3.6.0. Vert.x 4.5.33, Netty 4.1.137, REST Assured 6.0.1 and the Kubernetes client move with the BOM.

Two pins are kept deliberately, each with a comment in the pom:

- **httpclient5 stays at 5.5.1.** 5.6 adds transparent content decoding through commons-compress (`CommonsCompressCodecFactory`), and its XZ codec references `org.tukaani.xz.XZInputStream`, which is not on the classpath. GraalVM reports it at link time through docker-java's `FrameReader`, so the native image does not build. The Quarkus 3.39 BOM manages 5.6.3, so the explicit pin is what keeps it out. Later options: add `org.tukaani:xz`, or exclude the decoder.
- **protobuf-java 3.25.9 and proto-google-common-protos 2.49.0 stay** for apicurio 2.6.13. Newer common-protos are built against protobuf 4.33+ (`com.google.protobuf.GeneratedFile`), the mixed-classpath native failure that blocks #2906 and blocked #2083 before it. apicurio 3.x is the real fix and is its own change.

Also left for their own changes: h2 2.5 (MVStore is the persistence format), json-schema-validator 3.x (API change), pre-releases (wire 7 RC, assertj 4 M1, httpclient5 5.7 alpha), and kafka-clients, which #3160 removes.

Migration notes for 3.38 and 3.39 checked: Host header validation now defaults on when `quarkus.http.host` is localhost (Floci binds 0.0.0.0); 3.39 only changes Quarkus Data, unused here. One new warning to look at separately: Quarkus reports `quarkus.native.resources.excludes` as unsupported with GraalVM 25 reachability metadata, so the snappy exclusion list in `application.yml` may already be a no-op on the Mandrel 25 builder.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No emulated surface changes. The native binary was booted after the build and answered STS `GetCallerIdentity` on the Query protocol.

## Checklist

- [x] `./mvnw test` passes locally: full suite 14190 tests, 0 failures, 1 environmental error (`ContainerPlatformDockerIntegrationTest` wants a linux/amd64 busybox on an arm64 host). The unsharded fork hit the 6g heap after 1160 of 1294 classes, so the remaining 188 classes ran separately: 3479 tests, 2 failures in `RuntimeApiServerTest` that reproduce on unmodified main (`extensionShutdown_racedAgainstStop_isNeverOrphaned` 2/2, `quiesceAtomicallyClearsInFlightWithStopped` flaky). Native image: `-Dnative -Ob` on GraalVM 25.0.3 aarch64 builds in under two minutes.
- [ ] New or updated integration test added: dependency-only change, no behaviour to pin.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -181,7 +181,12 @@ jobs:
           github-token: ${{ github.token }}
 
       - name: Build native executable
-        run: ./mvnw clean package -Dnative -DskipTests -B -Dquarkus.native.additional-build-args-append="-march=x86-64-v2"
+        # Quarkus 3.39 switches on -H:+AOTSingleCallsiteInline for Mandrel 25.0.4+. The generated
+        # REST invokers for AwsQueryController.dispatch and AwsJson11Controller.handle are each the
+        # single call site of a very large method; inlined, the method outgrows the aarch64
+        # conditional-branch range and the build dies with BranchTargetOutOfBoundsException.
+        # The flag only exists on Mandrel, so it is passed here rather than in application.yml.
+        run: ./mvnw clean package -Dnative -DskipTests -B -Dquarkus.native.additional-build-args-append="-march=x86-64-v2,-H:-AOTSingleCallsiteInline"
 
       - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
@@ -216,7 +221,12 @@ jobs:
           github-token: ${{ github.token }}
 
       - name: Build native executable
-        run: ./mvnw clean package -Dnative -DskipTests -B
+        # Quarkus 3.39 switches on -H:+AOTSingleCallsiteInline for Mandrel 25.0.4+. The generated
+        # REST invokers for AwsQueryController.dispatch and AwsJson11Controller.handle are each the
+        # single call site of a very large method; inlined, the method outgrows the aarch64
+        # conditional-branch range and the build dies with BranchTargetOutOfBoundsException.
+        # The flag only exists on Mandrel, so it is passed here rather than in application.yml.
+        run: ./mvnw clean package -Dnative -DskipTests -B -Dquarkus.native.additional-build-args-append="-H:-AOTSingleCallsiteInline"
 
       - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -162,7 +162,12 @@ jobs:
           restore-keys: maven-${{ runner.os }}-
 
       - name: Build native executable (quick build)
-        run: mvn clean package -Dnative -DskipTests -B -Dquarkus.native.additional-build-args-append="-Ob"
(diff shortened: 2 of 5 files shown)


hectorvent:
CI on 5b77ee3: everything green except "Build native floci image (arm64)", which died in the compile step with `BranchTargetOutOfBoundsException: Branch target 1851048 out of bounds` on the generated REST invokers for `AwsQueryController.dispatch` and `AwsJson11Controller.handle`.

Cause: Quarkus 3.39 turns on `-H:+AOTSingleCallsiteInline` by default for Mandrel 25.0.4+ (quarkusio/quarkus#55482). Each invoker is the single call site of a very large dispatch method, so it gets inlined and the result outgrows the AArch64 conditional-branch range of plus or minus 1 MB. amd64 passes only because x86 branches reach further, and its compile step still went from about 60 s to 250 s.

Fix in 0a5fe1f: `-H:-AOTSingleCallsiteInline` appended on every Mandrel native build command (build-images, compatibility, nightly, release; both arches). It cannot live in application.yml because the option is a Mandrel-only backport, and Oracle GraalVM 25.0.3 rejects it as unrecognized, which would break local builds.

hectorvent:
Second CI round on 0a5fe1f: both native image builds green, but the native compat suites failed on both architectures with one root cause. The binary could not talk to the Docker daemon: every request to the unix socket failed with "Failure setting extended socket option", so ECR, ECS, RDS, Redshift and the CDK bootstrap all went red while the JVM shards stayed green.

Cause: the Quarkus 3.38+ BOM manages httpcore5 at 5.4.3 (the 3.37 BOM did not, so main resolves the 5.3.6 that httpclient5 5.5.1 declares). httpcore5 5.4 changed the SocketConfig defaults to keep-alive on with a 5 s keep-idle. docker-java copies those defaults, so httpclient5 sets TCP_KEEPIDLE through reflection (`SocketSupport.setOption`), which is unregistered in the native image. Bisected locally: Quarkus 3.37.4 clean, 3.38.3 and 3.39.2 fail, and the failing frame is `DefaultHttpClientConnectionOperator.connect:203`.

Fix in ae3e9426d: httpcore5 and httpcore5-h2 pinned to 5.3.6 next to the httpclient5 pin; docker-java is the only consumer. Verified on a local native build: Quarkus 3.39.2 boots and ECR GetAuthorizationToken starts its backing registry container over the Docker socket.

pgermosen (approved):
Diffed every version bump and both native-build fixes against the writeup and it all lines up exactly. Good that the compat suites on both arches confirm the httpcore5 pin actually restores Docker socket connectivity, not just that the build compiles.

PR #3177, opened 2026-09-07
fix(lambda): enforce handler deadline at dispatch

### LINKED ISSUE #3154: [BUG] Lambda Invoke accepts successful handler responses after the configured timeout
## Service

Lambda

## AWS API Action

`CreateFunction`, `Invoke`

## Expected behavior

A handler that sleeps for four seconds with a three-second timeout should return a function timeout error. The Runtime API's `Lambda-Runtime-Deadline-Ms` header identifies the actual timeout instant.

References checked on 2026-09-06:

- [AWS function timeout](https://docs.aws.amazon.com/lambda/latest/dg/configuration-timeout.html)
- [AWS Runtime API](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html)
- [AWS execution environment lifecycle](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtime-environment.html)

## Actual behavior

Floci 2.0.1 returned success after 4.007 seconds for a four-second handler and a three-second timeout, with no `FunctionError`. The runtime reported 3000 milliseconds remaining at handler entry.

The same grace-period acceptance path remains in current main `9df197f3af3c28b72dba793553f6a44b5c622866` (source inspection): `LambdaExecutorService` advertises the configured deadline but waits for the configured timeout plus two seconds and accepts a successful result during that extra interval.

Removing the extra wait alone also needs care: container acquisition currently does not wait for the runtime's first `/next` request, so runtime/static initialization can consume the handler's advertised allowance.

## Reproduction

This is an emulator-only probe. Invoke it a second time if necessary to distinguish cold-start overhead from the warm handler deadline; the decisive result is a successful four-second handler with a three-second timeout. An HTTP status of 200 for the Invoke API itself is allowed; the missing `FunctionError` is the bug.

Run against a local Floci instance with Docker enabled and the AWS CLI installed:

```bash
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-east-1 AWS_PAGER=
endpoint=http://localhost:4566
function_name="lambda-deadline-probe-$RANDOM"
python3 - <<'PY'
import zipfile
with zipfile.ZipFile('/tmp/lambda-deadline-probe.zip', 'w') as archive:
    archive.writestr('handler.py', '''import time
def handler(event, context):
    time.sleep(4)
    return {"ok": True}
''')
PY
aws --endpoint-url "$endpoint" lambda create-function \
  --function-name "$function_name" --runtime python3.12 \
  --handler handler.handler --timeout 3 \
  --role arn:aws:iam::000000000000:role/deadline-probe \
  --zip-file fileb:///tmp/lambda-deadline-probe.zip
aws --endpoint-url "$endpoint" lambda invoke \
  --function-name "$function_name" /tmp/lambda-deadline-response.json
cat /tmp/lambda-deadline-response.json
aws --endpoint-url "$endpoint" lambda delete-function --function-name "$function_name"
```

Expected: invocation metadata includes `FunctionError` and the payload describes a timeout. Actual: metadata omits `FunctionError` and payload is `{"ok": true}`.

## Environment

- Released baseline: `floci/floci:2.0.1`, digest `sha256:4e451c39c7bb88e3cd4f87e8fc0c25d5b47695a51185d521e2241fa00486e8eb`
- Main regression: `dec348d38259eb79783863adc21e42ad226b08f8`
- Java: OpenJDK 25.0.2; Maven wrapper 3.9.16
- Real Lambda runtime: Python 3.12 Docker image
- AWS CLI used for the proposed fix's independent container check: 2.33.28

No AWS account execution is claimed. Expected behavior is grounded in the official documentation; emulator behavior was reproduced locally.

## Verification limits

These are local emulator reproductions, not live AWS comparisons. Current main `9df197f3af3c28b72dba793553f6a44b5c622866` was inspected again before filing and still contains the relevant implementation. Prior container results are identified by their tested version below; current main was not rebuilt for this report.

Source: [`LambdaExecutorService.java`](https://github.com/floci-io/floci/blob/9df197f3af3c28b72dba793553f6a44b5c622866/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaExecutorService.java#L102). The issue is acceptance of late success, not a requirement that Docker teardown or the client response finish at exactly three seconds.

7487:
## Summary

- Start the Lambda handler deadline when the Runtime API successfully dispatches the invocation.
- Complete the invocation result with a timeout at the advertised deadline, while retaining the existing two-second allowance only for initial runtime dispatch.
- Refresh the deadline on redelivery after a failed Runtime API write and keep extension events on the same deadline.

Closes #3154

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, a successful handler response could be accepted during the executor's two-second grace period after the configured timeout. The grace period also made a direct removal unsafe because a cold runtime may not have polled `/next` yet.

The handler deadline now starts on successful Runtime API dispatch and the result future expires at that advertised deadline. A late response becomes `Function.TimedOut`, while cold-start initialization does not consume the handler's execution budget.

## Verification

- `./mvnw -q -Dtest='LambdaExecutorServiceTest,RuntimeApiServerTest' test` — 64 passed.
- `./mvnw -q -Dtest=LambdaVersionIntegrationTest test` — 9 passed with real Lambda containers.
- `./mvnw -q -DskipTests package` — passed.
- `./mvnw test` reached 14,188 tests locally. It ended with an out-of-memory error plus three host-specific failures: two TEST-NET reachability assertions and one ALB listener response. The unchanged upstream worktree reproduces those same three failures under this Colima environment.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaExecutorService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaExecutorService.java
@@ -13,6 +13,7 @@ import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
@@ -27,8 +28,8 @@ import java.util.concurrent.TimeoutException;
 public class LambdaExecutorService {
 
     private static final Logger LOG = Logger.getLogger(LambdaExecutorService.class);
-    /** Grace period beyond the configured function timeout to allow the runtime to report back. */
-    private static final int TIMEOUT_GRACE_SECONDS = 2;
+    /** Extra time for a newly started runtime to request its first invocation. */
+    private static final int RUNTIME_DISPATCH_GRACE_SECONDS = 2;
 
     private final WarmPool warmPool;
     private final ObjectMapper objectMapper;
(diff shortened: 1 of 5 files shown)


pgermosen (approved):
Traced the dispatch/markDispatched sequencing carefully since it's a concurrency-sensitive fix, and checked the core premise against AWS's own execution-lifecycle docs too - Init and Invoke really are separate phases there, so starting the handler's timeout clock at dispatch instead of at queue time is correct, not just internally consistent. The regression test that completes the result at 1200ms against a 1s timeout is exactly the right way to pin down the original bug. Good fix.

PR #3182, opened 2026-09-07
fix(cloudfront): echo OriginSslProtocols and round-trip legacy cache behavior fields

### LINKED ISSUE #3208: CloudFront: legacy ForwardedValues/MinTTL/DefaultTTL/MaxTTL dropped on cache-behavior read-back
Follow-up from #2767, split out per #3015's scope note.

`DefaultCacheBehavior`/`CacheBehavior` accept `ForwardedValues`, `MinTTL`, `DefaultTTL`, and `MaxTTL` on create (the legacy, pre-cache-policy way of configuring a cache behavior: real AWS requires either `CachePolicyId` or `ForwardedValues`), but none of the four are stored or echoed back on read. A Terraform config still using the legacy `forwarded_values` block (instead of `cache_policy_id`) will see these fields as permanently absent on every `plan` after `apply`, producing an unresolvable diff.

This doesn't affect the modern `cache_policy_id` path, which floci already fully supports and echoes correctly.

Scope: model `ForwardedValues` (`QueryString`, `QueryStringCacheKeys`, `Cookies.Forward`/`WhitelistedNames`, `Headers`), parse it plus `MinTTL`/`DefaultTTL`/`MaxTTL` in `CloudFrontController`'s cache-behavior parsers, and echo them in `xmlDefaultCacheBehavior`/`xmlCacheBehavior` only when present (i.e. legacy mode was actually used).

Not a crash, a diff/fidelity issue, lower urgency than #2767's TrustedSigners nil-deref (already fixed).

matanbaruch:
## Summary

After #3015 echoed TrustedSigners, creating a CloudFront distribution against Floci still crashed the pulumi-aws v5 provider on the read that runs at the end of create. A custom origin's `OriginSslProtocols` was dropped, and the provider flattens `CustomOriginConfig.OriginSslProtocols.Items` with no nil guard, so the missing object segfaults it.

The legacy `forwarded_values` cache-behavior path lost more than that, so this also closes the read-back fidelity gap tracked separately.

Fixes #3208.

### The crash

```
[signal SIGSEGV: segmentation violation]

cloudfront.FlattenCustomOriginConfigSSL(...)  distribution_configuration_structure.go:1053
cloudfront.FlattenCustomOriginConfig(...)     distribution_configuration_structure.go:1019
cloudfront.FlattenOrigin(...)                 distribution_configuration_structure.go:749
cloudfront.FlattenOrigins(...)                distribution_configuration_structure.go:680
cloudfront.flattenDistributionConfig(...)     distribution_configuration_structure.go:153
cloudfront.resourceDistributionRead(...)      distribution.go
cloudfront.resourceDistributionCreate(...)    distribution.go
```

Line 1053 is `return flex.FlattenStringSet(osp.Items)`, where `osp` is the nil `OriginSslProtocols`. Line numbers are from the terraform-provider-aws revision bridged by pulumi-aws v5.43.0. Newer providers panic in the same read for the same reason, they just moved the code.

### The fix

1. `CustomOriginConfig` now echoes `OriginSslProtocols`, round-tripping the submitted protocols and defaulting to `TLSv1.2` when the client sends none.
2. `DefaultTTL` and `MaxTTL` were emitted only when greater than 0, so an explicit `default_ttl = 0` (the usual "do not cache" with `forwarded_values`) read back as absent and drifted on every plan. Presence is now tracked separately from the value: a submitted 0 round-trips, and a TTL the client never sent stays absent.
3. `ForwardedValues` echoed `Headers` and `QueryStringCacheKeys` as a bare `Quantity 0`, and the parser kept only `QueryString` and `Cookies.Forward`, so whitelisted headers, query-string cache keys and cookie names came back empty. All three now round-trip as Quantity/Items pairs, scoped so the shared `Name` element lands in the right block.
4. Both cache behaviors also round-trip `ForwardedValues`, `SmoothStreaming` and `FieldLevelEncryptionId`.

The `CachedMethods` part of this landed separately in #3217, so it is rebased out and main's version is used.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: the `GetDistribution` and `CreateDistribution` responses omitted `CustomOriginConfig.OriginSslProtocols` entirely, emitted `DefaultTTL` / `MaxTTL` only when non-zero, and flattened `ForwardedValues` down to `QueryString` plus `Cookies.Forward` with empty `Headers` and `QueryStringCacheKeys`. Real CloudFront echoes all of these, per the model in [botocore's cloudfront service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/cloudfront/2020-05-31/service-2.json).

Verified against the real pulumi-aws v5 provider (bridged terraform-provider-aws): a distribution with a custom origin, `defaultTtl = 0` and whitelisted headers and cookies creates, reads back with no segfault, and destroys clean. The same run panics in `flattenDistributionConfig` before the change.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
@@ -1934,6 +1934,7 @@ public class CloudFrontController {
                     .elem("HTTPSPort", coc.getOrDefault("HTTPSPort", "443").toString())
                     .elem("OriginProtocolPolicy",
                             coc.getOrDefault("OriginProtocolPolicy", "https-only").toString())
+                    .raw(xmlOriginSslProtocols(coc.get("OriginSslProtocols")))
                     .end("CustomOriginConfig");
         }
 
@@ -1966,13 +1967,19 @@ public class CloudFrontController {
                 .elem("CachePolicyId", dcb.getCachePolicyId())
                 .elem("OriginRequestPolicyId", dcb.getOriginRequestPolicyId())
                 .elem("ResponseHeadersPolicyId", dcb.getResponseHeadersPolicyId())
-                .elem("Compress", dcb.isCompress());
+                .elem("FieldLevelEncryptionId",
+                        dcb.getFieldLevelEncryptionId() != null ? dcb.getFieldLevelEncryptionId() : "")
+                .elem("Compress", dcb.isCompress())
+                .elem("SmoothStreaming", dcb.isSmoothStreaming());
+
+        xml.raw(xmlCacheBehaviorTtls(dcb.getMinTTL(), dcb.getDefaultTTL(), dcb.getMaxTTL()));
 
         xml.raw(xmlTrustedSigners());
         xml.raw(xmlTrustedKeyGroups(
                 dcb.isTrustedKeyGroupsEnabled(), dcb.getTrustedKeyGroups()));
 
         xml.raw(xmlAllowedMethods(dcb.getAllowedMethods(), dcb.getCachedMethods()));
+        xml.raw(xmlCacheBehaviorForwardedValues(dcb.getCachePolicyId(), dcb.getForwardedValues()));
 
         xml.raw(xmlFunctionAssociations(dcb.getFunctionAssociations()));
         xml.raw(xmlLambdaFunctionAssociations(dcb.getLambdaFunctionAssociations()));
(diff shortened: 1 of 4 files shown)


hectorvent:
Thanks, the `OriginSslProtocols` echo is the right fix for the provider crash. Two of Greptile's points hold against the CloudFront model (https://github.com/boto/botocore/blob/develop/botocore/data/cloudfront/2020-05-31/service-2.json) and the Terraform use case:

**(blocking)** `DefaultTTL` and `MaxTTL` are emitted only when greater than 0, so an explicit `default_ttl = 0` (the usual "do not cache" with `forwarded_values`) reads back as absent and drifts. Tracking presence rather than reusing 0 for "unset" would fix it.

**(blocking)** `ForwardedValues` always echoes `Headers` and `QueryStringCacheKeys` as `Quantity 0`; the parser keeps only `QueryString` and `Cookies.Forward`, so whitelisted headers and cookie names come back empty.

**(blocking)** The `CachedMethods` part has since landed on main through #3217, so this branch now conflicts in `CloudFrontController`. A rebase that drops this PR's copy of that hunk clears the conflict and keeps the diff focused.

Those three unblock the merge from my side.

matanbaruch:
@hectorvent all three are addressed.

1. TTL presence is now tracked separately from the value, so an explicit `default_ttl = 0` round-trips and a TTL the client never sent stays absent. The model fields moved from `long` to `Long` for that; all their usages were in this controller.
2. `ForwardedValues` now round-trips `Headers`, `QueryStringCacheKeys` and `Cookies.WhitelistedNames` as Quantity/Items pairs. The parser scopes the shared `Name` element per block, same trap `<Method>` had in #3217.
3. Rebased onto main and dropped my `CachedMethods` hunk, so #3217's version is the one in place. Diff is down to OriginSslProtocols plus the forwarded_values round-trip.

Also marked it `Fixes #3208` since point 2 is that issue.

New tests cover each: explicit `DefaultTTL` of 0 survives, an unsubmitted TTL stays absent, and whitelisted headers, query-string cache keys and cookie names come back non-empty. They fail without the change. Full CloudFront suite is green at 181, and I re-ran it end to end against the real pulumi-aws v5 provider with a custom origin, `defaultTtl = 0` and whitelisted headers and cookies: creates, reads back with no segfault, destroys clean.

hectorvent (approved):
Thank you, all three are closed cleanly: presence tracked apart from the value so an explicit zero TTL survives the round trip, the three forwarded lists echoed as Quantity/Items, and the CachedMethods half dropped now that #3217 is in. Approving.

PR #3184, opened 2026-09-07
fix(cognito): verify self-service access tokens

nkootstra:
## Summary

Cognito self-service operations previously trusted decoded JWT claims without verifying the token signature. This allowed forged claims to authorize user lookup, attribute mutations, verification-code delivery, global sign-out, MFA preference updates, and UserInfo access.

This PR adds one shared access-token verifier and routes all affected entry points through it. It validates the RS256 signature against the persisted user-pool key, issuer, key ID, client, token use, required claims, expiry, and token revocation state. Valid signed access-token flows remain unchanged.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Self-service Cognito operations now reject unsigned, modified, incorrectly signed, expired, wrongly issued, wrong-pool, wrong-client, wrong-token-use, and revoked access tokens instead of trusting their decoded claims. UserInfo uses the same verifier and continues to return the existing Cognito-compatible response for valid signed access tokens.

Open Cognito PRs were checked for overlap; none cover access-token signature and claim verification.

## Checklist

- [ ] `./mvnw test` passes locally — the full suite exceeded the 15-minute local timeout before completion

Focused validation: `CognitoServiceTest` and `CognitoUserInfoIntegrationTest` passed with 223 tests. The Java Cognito compatibility suite was attempted but the emulator was not running at `http://localhost:4566`, so it failed with connection-refused errors. Project diagnostics and `git diff --check` are clean.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cognito/CognitoService.java
@@ -1655,20 +1655,9 @@ public class CognitoService implements ResourceProvider {
                     "1 validation error detected: Value at 'accessToken' failed to satisfy constraint: Member must not be null", 400);
         }
 
-        String username = extractUsernameFromToken(accessToken);
-        String poolId = extractPoolIdFromToken(accessToken);
-        String jti = extractJtiFromToken(accessToken);
-
-        if (username == null || poolId == null || jti == null) {
-            throw new AwsException("NotAuthorizedException", "Invalid access token", 400);
-        }
-
-        // A token that was already revoked (or issued before an earlier sign-out) cannot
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Traced verifyAccessToken end to end since this closes a real forge-any-token gap across seven self-service operations plus UserInfo. The signing key comes from server-side pool state, not the token, alg is locked to RS256, and token_use/client/pool binding are all checked. The cross-pool key-confusion test (a validly-signed token from one pool relabeled with another pool's iss, still carrying the original signature) is exactly the right thing to pin down here. Great work.

PR #3185, opened 2026-09-07
fix(cloudformation): scope stack state by account

nkootstra:
## Summary

CloudFormation stack and export state is now isolated by caller account and region. This prevents identical stack or export names in different accounts from colliding, and prevents cross-account reads and mutations from reaching another account’s state.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

CloudFormation stack identity, export lookup, stack/change-set ARNs, persistence keys, nested stacks, and StackSet target-account lookups now preserve the owning account. Legacy persisted records without an owner are attributed to the configured default account during load.

## Checklist

- [ ] `./mvnw test` passes locally — full suite not run

Focused validation passed:

- `./mvnw -Dtest=CloudFormationAccountOwnershipIntegrationTest test`
- `./mvnw -Dtest=CloudFormationPersistenceIntegrationTest test`
- `./mvnw -Dtest=CloudFormationStackSetsIntegrationTest test`
- `./mvnw -DskipTests compile`
- `git diff --check`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationService.java
@@ -133,10 +133,10 @@ public class CloudFormationService implements ResourceProvider {
     }
 
     private String ownerAccount(Stack stack) {
-        if (stack.getAccountId() == null) {
-            throw new IllegalStateException("CloudFormation stack has no owner: " + stack.getStackName());
-        }
-        return stack.getAccountId();
+        // Persisted stacks are assigned their storage partition's account during startup. The
+        // default is only a compatibility attribution for transient ownerless fixtures created
+        // directly by callers of the service.
+        return stack.getAccountId() != null ? stack.getAccountId() : config.defaultAccountId();
     }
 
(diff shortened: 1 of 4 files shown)


nkootstra:
Addressed Greptile's foreign-stack existence oracle in d0d698d6. Update change sets now reject only when the caller's account/region/name partition is absent; this produces the same response for an actually absent stack and a stack owned by another account. Delete-by-name now retains the existing absent-stack no-op without scanning foreign state, while ARN deletes continue to validate the explicit account and region. The ownership integration test and related CloudFormation tests pass.

hectorvent (approved):
Thanks, this is a clean cut: the in-memory key carries the account, the storage key does not, so existing persisted stacks load into the default account's partition unchanged, and the executor path already ran under the owner's account through `runUnderAccount`, which is what makes `currentAccount()` inside `executeTemplate` line up. Greptile's oracle point was right and the fix in d0d698d6 (no foreign-partition scan; by-name delete stays a no-op, by-ARN delete checks the ARN's account and region) is what I would have asked for. Ran the whole CloudFormation package on the merge result with main: 1,428 tests green. Approving.

PR #3189, opened 2026-09-07
fix(dynamodb): make TransactWriteItems atomic

nkootstra:
## Summary

Fixes partial mutations from failed `TransactWriteItems` requests. Mutations are now staged per participant, validated before live storage changes, committed only after the complete transaction succeeds, and side effects are published afterward.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

A failed transaction previously could leave earlier writes committed and could publish stream or Kinesis effects. The implementation now preserves all-or-nothing item behavior while retaining deterministic lock ordering and idempotency handling.

The fix does not include stream-iterator, ownership, or unrelated DynamoDB refactors.

## Checklist

- [ ] `./mvnw test` passes locally — the full suite exceeded the available timeout

### Validation run

- `./mvnw -q -Dtest=DynamoDbServiceTest,DynamoDbConcurrencyIntegrationTest test`
- `./mvnw -q -Dtest=DynamoDbIntegrationTest#transactWriteFailureIsAtomicThroughAwsProtocol test`
- Diagnostics and `git diff --check` passed

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbService.java
@@ -86,6 +86,7 @@ public class DynamoDbService implements ResourceProvider {
     // Items stored per table: storageKey -> Map<itemKey, item>
     // itemKey is "pk" or "pk#sk" depending on table schema
     private final ConcurrentHashMap<String, ConcurrentSkipListMap<String, JsonNode>> itemsByTable = new ConcurrentHashMap<>();
+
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Traced the staging path through itemsFor and the commit loop since this is a data-integrity fix - reads and writes consistently route to the staged overlay during a transaction, the commit only touches the actual participant keys, and stream/Kinesis events are correctly deferred until after a successful commit. The verifyNoInteractions test for failed transactions is exactly the right check. Good fix.

PR #3196, opened 2026-09-07
fix(lambda): correct Function URL body encoding, cookies, and error status

### LINKED ISSUE #3153: [BUG] Lambda Function URLs corrupt binary requests, drop cookies, and return 200 on function errors
## Service

Lambda Function URLs

## AWS API Action

`CreateFunctionUrlConfig` followed by HTTP invocation (payload format 2.0).

## Expected behavior

Binary request bytes must survive transport to the handler, with a base64 body and `isBase64Encoded=true`. Response `cookies` entries must become separate `Set-Cookie` headers. An unhandled function exception must be represented as a 5xx HTTP response, rather than a successful 200.

References: [Function URL request/response contract](https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html), [Function URL 5xx metrics](https://docs.aws.amazon.com/lambda/latest/dg/urls-monitoring.html).

## Actual behavior

On released 2.0.1:

- Sending bytes `00 ff 80 0a` with `Content-Type: application/octet-stream` delivers a text body with Unicode replacement characters and `isBase64Encoded=false`, irreversibly changing the payload.
- Returning `cookies: ["probe=one", "other=two"]` produces no `Set-Cookie` headers.
- Raising a handler exception returns HTTP 200.

UTF-8 JSON requests, binary responses, and plain JSON response inference passed as controls.

## Reproduction

Save the four files below together. They create a disposable function using a public Python Lambda image, dummy credentials and synthetic payloads. Docker must be available to Floci through the mounted socket.

```sh
python3 -m venv .venv
. .venv/bin/activate
pip install boto3==1.42.62 botocore==1.42.62
export COMPOSE_PROJECT_NAME=floci-function-url-repro
docker build -t floci-function-url-repro-handler .
docker compose up -d
# Wait for Floci to finish starting.
endpoint="http://$(docker compose port floci 4566)"
python probe.py --endpoint "$endpoint" --image floci-function-url-repro-handler
docker compose down
```

Expected baseline output: `binary request`, `response cookies`, and `function error` are false; the other three checks are true. The function is deleted in `finally`.

The error check below is limited to the documented **200 versus 5xx** distinction. Exact 502/body formatting is not claimed as independently verified Function URL parity. This reporting copy narrows the previously executed probe's stricter 502/body assertion; the captured baseline returned 200, so it also fails this narrower check.

<details>
<summary>Dockerfile</summary>

```dockerfile
FROM public.ecr.aws/lambda/python:3.12
COPY handler.py ${LAMBDA_TASK_ROOT}/handler.py
CMD ["handler.handler"]
```

</details>

<details>
<summary>compose.yaml</summary>

```yaml
services:
  floci:
    image: ${FLOCI_IMAGE:-floci/floci:2.0.1}
    environment:
      FLOCI_BASE_URL: http://floci:4566
      FLOCI_HOSTNAME: floci
      FLOCI_DEFAULT_REGION: eu-central-1
      FLOCI_STORAGE_MODE: memory
      FLOCI_AUTH_VALIDATE_SIGNATURES: "false"
      FLOCI_SERVICES_LAMBDA_DOCKER_NETWORK: ${COMPOSE_PROJECT_NAME:?Set COMPOSE_PROJECT_NAME}_default
      FLOCI_SERVICES_LAMBDA_DOCKER_HOST_OVERRIDE: floci
    volumes:
      - ${DOCKER_SOCKET:-/var/run/docker.sock}:/var/run/docker.sock
    ports:
      - "127.0.0.1::4566"
```

</details>

<details>
<summary>handler.py</summary>

```python
import base64
import json


def handler(event, context):
    path = event.get("rawPath")
    if path == "/error":
        raise RuntimeError("sensitive-error-sentinel")
    if path == "/binary-response":
        return {
            "statusCode": 200,
            "headers": {"content-type": "application/octet-stream"},
            "body": base64.b64encode(bytes([0, 255, 128, 10])).decode(),
            "isBase64Encoded": True,
        }
    if path == "/plain-json":
        return {"ok": True}
    return {
        "statusCode": 201,
        "headers": {"content-type": "application/json"},
        "cookies": ["probe=one", "other=two"],
        "body": json.dumps(event),
    }
```

</details>

<details>
<summary>probe.py</summary>

```python
import argparse
import base64
import json
import uuid
from urllib.error import HTTPError
from urllib.parse import urlsplit
from urllib.request import Request, urlopen

import boto3
from botocore.config import Config


def call(url, path, body=None, content_type=None):
    headers = {"content-type": content_type} if content_type else {}
    request = Request(url + path, data=body, headers=headers)
    try:
        response = urlopen(request, timeout=30)
    except HTTPError as error:
        response = error
    with response:
        return response.status, response.headers, response.read()


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--endpoint", required=True)
    parser.add_argument("--image", required=True)
    args = parser.parse_args()
    session = boto3.Session(
        aws_access_key_id="test",
        aws_secret_access_key="test",
        region_name="eu-central-1",
    )
    client = session.client(
        "lambda",
        endpoint_url=args.endpoint,
        config=Config(connect_timeout=3, read_timeout=30, retries={"total_max_attempts": 1}),
    )
    function_name = "function-url-repro-" + uuid.uuid4().hex[:8]
    client.create_function(
        FunctionName=function_name,
        Role="arn:aws:iam::000000000000:role/repro",
        PackageType="Image",
        Code={"ImageUri": args.image},
        Timeout=10,
    )
    try:
        config = client.create_function_url_config(FunctionName=function_name, AuthType="NONE")
        url_id = urlsplit(config["FunctionUrl"]).hostname.split(".")[0]
        url = f"{args.endpoint}/lambda-url/{url_id}"
        binary = bytes([0, 255, 128, 10])

        text_status, text_headers, text_body = call(
            url, "/echo", b'{"text":"hello"}', "application/json"
        )
        text_event = json.loads(text_body)
        binary_status, _, binary_body = call(
            url, "/echo", binary, "application/octet-stream"
        )
        binary_event = json.loads(binary_body)
        output_status, _, output_body = call(url, "/binary-response")
        error_status, _, error_body = call(url, "/error")
        plain_status, plain_headers, plain_body = call(url, "/plain-json")

        checks = {
            "text request": text_status == 201
            and text_event["body"] == '{"text":"hello"}'
            and text_event["isBase64Encoded"] is False,
            "binary request": binary_status == 201
            and binary_event["body"] == base64.b64encode(binary).decode()
            and binary_event["isBase64Encoded"] is True,
            "binary response": output_status == 200 and output_body == binary,
            "response cookies": text_headers.get_all("set-cookie") == ["probe=one", "other=two"],
            "function error": 500 <= error_status < 600,
            "plain JSON": plain_status == 200
            and plain_headers.get_content_type() == "application/json"
            and json.loads(plain_body) == {"ok": True},
        }
        print(json.dumps(checks, indent=2))
        raise SystemExit(0 if all(checks.values()) else 1)
    finally:
        client.delete_function(FunctionName=function_name)


if __name__ == "__main__":
    main()
```

</details>

## Environment

- Floci image: `floci/floci:2.0.1`, ARM64 image ID `sha256:4e451c39c7bb88e3cd4f87e8fc0c25d5b47695a51185d521e2241fa00486e8eb`.
- Python SDK: boto3/botocore 1.42.62; Python 3.12 Lambda image.
- Docker on macOS/ARM64; Java SDK not applicable.

## Verification limits

These are local emulator reproductions, not live AWS comparisons. Current main `9df197f3af3c28b72dba793553f6a44b5c622866` was inspected again before filing and still contains the relevant implementation. Prior container results are identified by their tested version below; current main was not rebuilt for this report.

Source: [`LambdaUrlInvocationController.java`](https://github.com/floci-io/floci/blob/9df197f3af3c28b72dba793553f6a44b5c622866/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaUrlInvocationController.java#L171): unconditional text decoding, no response-cookie mapping, and no `FunctionError` handling in `buildResponse`. A local focused correction and regression tests are available.

omatheusmesmo:
## Summary

Fixes three Lambda Function URL bugs in `LambdaUrlInvocationController`:

1. Binary request bodies (e.g. `Content-Type: application/octet-stream` with raw bytes) were lossily text-decoded (`new String(body)`) and always sent with `isBase64Encoded=false`, corrupting the payload before it reached the handler.
2. A `cookies` array in the handler's response payload (payload format 2.0) was ignored, so cookies set by the handler never became `Set-Cookie` response headers.
3. When the invoked function raised an unhandled exception (`FunctionError` set on the invoke result), the Function URL still returned the underlying invoke result's status (200) instead of a 5xx error.

Closes #3153

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

Per the Lambda Function URLs documentation (payload format 2.0):
- https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html: "If the content type of the request is binary, the body is base64-encoded" and `isBase64Encoded` is `TRUE` for binary payloads. AWS does not publish an exhaustive content-type list for HTTP API integrations (confirmed via API Gateway HTTP API docs and AWS blog posts on binary handling), so the request body is now classified by attempting a strict UTF-8 decode: if it decodes cleanly it is sent as text with `isBase64Encoded=false`, otherwise it is base64-encoded with `isBase64Encoded=true`. This guarantees the handler never receives corrupted bytes regardless of the declared content type.
- Same page, "Cookies" section: response `cookies` entries must become individual `set-cookie` response headers, not a single combined header. `buildResponse` now emits one `Set-Cookie` header per array entry.
- Unhandled Lambda errors surfaced through a Function URL return `502 Bad Gateway` (matches AWS's documented behavior for proxy-style Lambda integrations and the existing `ApiGatewayController` in this codebase, which already maps `FunctionError` to 502). `buildResponse` now checks `InvokeResult.getFunctionError()` first and short-circuits to 502.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaUrlInvocationController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaUrlInvocationController.java
@@ -28,10 +28,12 @@ import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.UriInfo;
 import org.jboss.logging.Logger;
 
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.Base64;
+import java.util.Locale;
 import java.util.UUID;
 
 /**
@@ -46,6 +48,7 @@ import java.util.UUID;
 public class LambdaUrlInvocationController {
 
     private static final Logger LOG = Logger.getLogger(LambdaUrlInvocationController.class);
+    private static final int FUNCTION_ERROR_STATUS = 502;
 
     private final LambdaService lambdaService;
     private final RegionResolver regionResolver;
@@ -169,8 +172,13 @@ public class LambdaUrlInvocationController {
         httpNode.put("userAgent", headers.getHeaderString("user-agent"));
 
         if (body != null && body.length > 0) {
-            root.put("body", new String(body));
-            root.put("isBase64Encoded", false);
+            if (isTextMediaType(headers.getHeaderString(HttpHeaders.CONTENT_TYPE))) {
+                root.put("body", new String(body, StandardCharsets.UTF_8));
+                root.put("isBase64Encoded", false);
+            } else {
+                root.put("body", Base64.getEncoder().encodeToString(body));
+                root.put("isBase64Encoded", true);
+            }
         } else {
             root.putNull("body");
             root.put("isBase64Encoded", false);
@@ -179,7 +187,23 @@ public class LambdaUrlInvocationController {
         return root.toString();
     }
 
+    /**
+     * Mirrors how AWS decides isBase64Encoded for Function URL / API Gateway proxy
+     * integration requests: it is driven by the Content-Type header, not by whether
+     * the raw bytes happen to be valid UTF-8. Same content-type allowlist already used
+     * for the equivalent ALB-Lambda integration in ElbV2DataPlane, normalized to
+     * lowercase first since media types are case-insensitive.
+     */
+    private boolean isTextMediaType(String contentType) {
+        String normalized = contentType == null ? null : contentType.toLowerCase(Locale.ROOT);
+        return normalized == null || normalized.startsWith("text/") || normalized.contains("json")
+                || normalized.contains("xml") || normalized.contains("form");
+    }
+
     private Response buildResponse(InvokeResult result) {
+        if (result.getFunctionError() != null) {
+            return buildFunctionErrorResponse(result);
+        }
         if (result.getPayload() == null || result.getPayload().length == 0) {
             return Response.status(result.getStatusCode()).build();
         }
@@ -191,10 +215,13 @@ public class LambdaUrlInvocationController {
                 if (node.has("headers")) {
                     node.get("headers").fields().forEachRemaining(e -> builder.header(e.getKey(), e.getValue().asText()));
                 }
+                if (node.has("cookies") && node.get("cookies").isArray()) {
+                    node.get("cookies").forEach(cookie -> builder.header("Set-Cookie", cookie.asText()));
+                }
                 if (node.has("body")) {
                     String body = node.get("body").asText();
                     boolean isBase64 = node.path("isBase64Encoded").asBoolean(false);
-                    byte[] bytes = isBase64 ? Base64.getDecoder().decode(body) : body.getBytes();
+                    byte[] bytes = isBase64 ? Base64.getDecoder().decode(body) : body.getBytes(StandardCharsets.UTF_8);
                     builder.entity(bytes);
                 }
                 return builder.build();
@@ -204,6 +231,16 @@ public class LambdaUrlInvocationController {
         } catch (Exception e) {
             return Response.ok(result.getPayload()).build();
         }
+    }
+
+    /**
+     * Real Function URLs never surface the raw invocation error payload (stack traces,
+     * error types, internal fields) to the caller. A raw error payload has no "body"
+     * field, so, matching ApiGatewayController's Lambda proxy integration, the response
+     * carries the 502 status with no entity at all.
+     */
+    private Response buildFunctionErrorResponse(InvokeResult result) {
+        return Response.status(FUNCTION_ERROR_STATUS).build();
     }
 
     private String jsonMessage(String message) {
--- a/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaUrlInvocationControllerTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/lambda/LambdaUrlInvocationControllerTest.java
@@ -1,11 +1,13 @@
 package io.github.hectorvent.floci.services.lambda;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.github.hectorvent.floci.core.common.RegionResolver;
 import io.github.hectorvent.floci.core.common.RequestContext;
 import io.github.hectorvent.floci.services.lambda.model.InvocationType;
 import io.github.hectorvent.floci.services.lambda.model.InvokeResult;
 import io.github.hectorvent.floci.services.lambda.model.LambdaAlias;
+import io.github.hectorvent.floci.services.lambda.model.LambdaFunction;
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.MultivaluedHashMap;
 import jakarta.ws.rs.core.Response;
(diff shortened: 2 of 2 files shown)


src/main/java/io/github/hectorvent/floci/services/lambda/LambdaUrlInvocationController.java
```diff
@@ -179,7 +190,22 @@ private String buildEvent(String method, String urlId, String proxy, HttpHeaders
         return root.toString();
     }
```
pgermosen, line 188: Confirming this one holds up. isUtf8Text() decides text-vs-binary by attempting a UTF-8 decode, but that's not how AWS classifies it - it's Content-Type driven. A declared application/octet-stream body whose bytes happen to be valid UTF-8 would still be sent to the handler as text here, when it should be base64-encoded. There's already a correct implementation of this in the codebase for the equivalent ALB-Lambda integration - ElbV2DataPlane.java lines 560-568 use a content-type allowlist (text/*, json, xml, form -> text, everything else -> base64), which I'd verified earlier against AWS's actual ALB-Lambda docs. Worth reusing that same approach here instead of the UTF-8-validity heuristic.

src/main/java/io/github/hectorvent/floci/services/lambda/LambdaUrlInvocationController.java
```diff
@@ -206,6 +235,14 @@ private Response buildResponse(InvokeResult result) {
         }
     }
 
+    private Response buildFunctionErrorResponse(InvokeResult result) {
+        Response.ResponseBuilder builder = Response.status(FUNCTION_ERROR_STATUS);
+        if (result.getPayload() != null && result.getPayload().length > 0) {
```
pgermosen: This one holds up too, and it's actually not consistent with the ApiGatewayController precedent the PR description cites. ApiGatewayController's 502 path looks for a "body" field inside the parsed Lambda payload - a raw error object ({errorMessage, errorType, trace}) never has one, so its response ends up with no entity at all. This method instead attaches the raw payload directly as the response body, so errorMessage/errorType/stackTrace reach the client verbatim. Worth aligning with what ApiGatewayController actually does rather than echoing the raw invocation result.

omatheusmesmo:
@pgermosen Both points addressed: media-type classification now reuses ElbV2DataPlane's content-type allowlist, lowercased first so mixed-case types like `Application/JSON` still classify as text; the function-error path now mirrors ApiGatewayController's actual behavior (502 with no entity, since a raw error payload has no `body` field to surface). Automated review is clean now, ready whenever you have time to take another look.

pgermosen:
Checked both, they're right. isTextMediaType now lowercases the content type before matching against the same allowlist ElbV2DataPlane uses, and the function-error path just returns 502 with no entity, matching ApiGatewayController exactly. Looks good.

pgermosen (approved):
Both fixes check out in the code. Thanks for the quick turnaround. Approving.

PR #3197, opened 2026-09-07
feat(floci-core): resolve active Docker context when DOCKER_HOST is unset

### LINKED ISSUE #1990: [FEAT] Resolve the active Docker context when DOCKER_HOST is unset (Colima/OrbStack/Podman)
## Service

Core / Lambda container launching (developer experience)

## API Action / Feature

Resolve the active Docker **context** when no `DOCKER_HOST` is set, or document the
requirement.

## AWS Documentation

n/a — this is about reaching the Docker daemon, not an AWS API.

## Why is this needed?

On a machine where Docker is provided by Colima (also OrbStack, Rancher Desktop,
Podman, or any non-default socket), `docker` CLI commands work because the CLI reads
the active docker **context**, but Floci does not: docker-java falls back to
`unix://localhost:2375`, which does not exist. Every test or feature that starts a
container then fails, with a message that points nowhere near the cause:

```
recoverable I/O exception (java.net.SocketException) caught when sending request to
unix://localhost:2375
Failed to start ECR backing registry container: java.net.SocketException: No such file or directory
```

Running `./mvnw test` on such a machine produced **73 failures + 53 errors across 21
test classes** — WebSocket, Lambda-backed authorizers, DocDB, ElastiCache, MemoryDB,
Neptune, ECR, and Lambda data-plane tests. Setting

```bash
export DOCKER_HOST=unix://$HOME/.colima/default/docker.sock
```

took that to **1 pre-existing failure**. Nothing was wrong with the code; the
daemon was simply not where the client looked.

The cost is misdirected debugging: the failures look like real product bugs spread
across a dozen services, and there is nothing in `CONTRIBUTING.md` (which lists
Docker as a prerequisite) to suggest the socket path matters.

Two options, either of which would have saved the detour:

1. **Resolve the active context** when `DOCKER_HOST` is unset — read
   `~/.docker/config.json` for `currentContext` and that context's endpoint, the way
   Testcontainers' `DockerClientProviderStrategy` does. `DockerClientProducer`
   already has a normalization + env-override seam
   (`resolveEffectiveDockerHost(configuredHost, dockerHostEnv)`) that this would
   slot into.
2. **Document it** — a line in `CONTRIBUTING.md` under Prerequisites noting that
   non-Docker-Desktop setups need `DOCKER_HOST` set, with the
   `docker context inspect --format '{{.Endpoints.docker.Host}}'` one-liner.

(1) is the better experience; (2) is nearly free. Happy to send either.

- [x] Yes
- [ ] No

omatheusmesmo:
## Summary

Resolves the active Docker context endpoint when neither `floci.docker.docker-host` nor `DOCKER_HOST` is explicitly set. On machines where Docker is provided by Colima, OrbStack, Rancher Desktop, or Podman, `docker-java` previously fell back to the hardcoded `unix:///var/run/docker.sock` (or the Windows named pipe), which does not exist there, causing every container-backed test to fail with an unrelated-looking `SocketException`.

`DockerClientProducer.resolveEffectiveDockerHost` now uses this priority order:

1. Explicit `floci.docker.docker-host` config (non-default value)
2. `DOCKER_HOST` env var (normalized)
3. The active Docker context's endpoint, resolved from `~/.docker/config.json`'s `currentContext` and that context's metadata file under `~/.docker/contexts/meta/<sha256(name)>/meta.json` (same storage format the Docker CLI and Testcontainers use)
4. The existing hardcoded platform default (unix socket on Linux/macOS, named pipe on Windows)

The built-in `"default"` context has no metadata file on disk, so it is treated the same as no context being configured, falling through to step 4. A missing or malformed `config.json`, or a context with no metadata file, is handled gracefully by falling through as well; nothing throws.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

Not applicable; this is a local developer-experience fix for Docker client bootstrapping, not an AWS API.

## Verified

- Added unit tests covering: resolution from a synthetic `config.json` + context metadata fixture, explicit config/`DOCKER_HOST` precedence over context resolution, the built-in `default` context falling through, a missing/malformed `config.json`, a missing context metadata file, and context resolution taking priority over the Windows named-pipe default.
- `./mvnw -Dtest=DockerClientProducerTest test`: 27/27 passing.
- `./mvnw compile`: succeeds.

Closes #1990

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/docker/DockerClientProducer.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/docker/DockerClientProducer.java
@@ -1,16 +1,35 @@
 package io.github.hectorvent.floci.core.common.docker;
 
 import io.github.hectorvent.floci.config.EmulatorConfig;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.api.exception.DockerClientException;
 import com.github.dockerjava.core.DefaultDockerClientConfig;
 import com.github.dockerjava.core.DockerClientImpl;
+import com.github.dockerjava.core.SSLConfig;
+import com.github.dockerjava.core.util.CertificateUtils;
 import com.github.dockerjava.httpclient5.ApacheDockerHttpClient;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Produces;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.X509Certificate;
 import java.time.Duration;
+import java.util.Optional;
 
 /**
  * CDI producer for the DockerClient singleton bean.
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thanks, this removes a real first-run trap on Colima, OrbStack and Podman. I checked the four Greptile points against the head and each is handled: the empty `config.json` guard, `DOCKER_CONFIG` and the config-path setting relocating the directory, `DOCKER_CONTEXT` winning over `currentContext`, and the context's TLS material and `SkipTLSVerify` flowing into the client. Approving.

PR #3198, opened 2026-09-07
fix(networkfirewall): model CreateFirewall* collisions and enforce UpdateToken

### LINKED ISSUE #2600: networkfirewall: unmodeled ResourceAlreadyExistsException and unenforced UpdateToken on mutating ops
Two disclosed, deliberately deferred items from #2580's review by pgermosen, filed here so they're visible post-merge rather than only living in the PR thread.

## 1. Unmodeled `ResourceAlreadyExistsException`

An earlier commit on #2580 (`164b951d2`) polished the message on a `ResourceAlreadyExistsException` thrown when creating a firewall/rule group/policy with a name already in use. Per the current botocore Network Firewall model (`2020-11-12/service-2.json`), no `*AlreadyExist*` shape exists at all — `CreateFirewall`'s modeled errors are `InvalidRequestException`, `LimitExceededException`, `InternalServerError`, `ThrottlingException`, `InsufficientCapacityException`, `InvalidOperationException`.

An unmodeled error code means an SDK client can't deserialize the response into a typed exception it recognizes, so any caller branching on error type will not match this case correctly.

**Fix**: replace `ResourceAlreadyExistsException` with the modeled `InvalidRequestException` (or whichever of the modeled errors best fits AWS's actual behavior on a name collision — worth checking real AWS behavior here, not just picking the nearest modeled shape).

## 2. `UpdateToken` accepted but never enforced

Every `UpdateFirewall*`/`Associate*`/`Disassociate*` mutation on a firewall accepts an `UpdateToken` parameter (per the model, an optimistic-concurrency token that should be validated against the resource's current token and rejected with a conflict error on mismatch). Floci currently accepts any `UpdateToken` value without checking it against the firewall's actual current token.

**Fix**: thread the current token through to each mutating op, compare against the caller-supplied token, and reject with the modeled conflict error (likely `InvalidTokenException`, needs checking against the actual model errors) on mismatch. This is a small matrix across every mutating path (6 `UpdateFirewall*` ops + 4 `Associate*`/`Disassociate*` ops), not a single seam — worth scoping as its own PR rather than folding into an unrelated fix.

Both items were disclosed in #2580's review thread with pgermosen rather than discovered independently.

omatheusmesmo:
## Summary

Fixes two gaps in the Network Firewall emulation, both deferred from #2580's review with pgermosen and filed as #2600.

1. `CreateFirewall`, `CreateRuleGroup` and `CreateFirewallPolicy` raised an unmodeled `ResourceAlreadyExistsException` on a name collision. Checked against botocore's `network-firewall/2020-11-12/service-2.json`, none of these three operations model any `*AlreadyExist*` error shape; their only modeled client-fault error is `InvalidRequestException`. All three now raise `InvalidRequestException` on a name collision.
2. Every `UpdateFirewall*` (6 ops) and `Associate*`/`Disassociate*` (4 ops, plus `AssociateFirewallPolicy`, which shares the same code path) mutation on a firewall accepted any `UpdateToken` value without validating it. Each of these operations models `UpdateToken` as optional: per the AWS docs (confirmed via `API_UpdateFirewallDescription`), omitting it applies the change unconditionally, while supplying it must be checked against the firewall's current token and rejected with `InvalidTokenException` on a mismatch. The firewall's current token is now tracked and rotated on every successful mutation, and every mutating op validates a supplied token against it.

While implementing (2), also found and fixed two adjacent gaps confirmed against the model:
- `DescribeFirewallResponse` is documented to carry `UpdateToken` as a top-level field (sibling of `Firewall`/`FirewallStatus`); floci's response omitted it entirely, leaving no way to look up the current token to submit with the next update.
- `AssociateFirewallPolicyResponse` is documented to carry `UpdateToken` like every other mutating response; floci's response omitted it.

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

Verified against the current botocore Network Firewall model (`network-firewall/2020-11-12/service-2.json`) and the public AWS API reference (`docs.aws.amazon.com/network-firewall/latest/APIReference/`) for `CreateFirewall`, `CreateRuleGroup`, `CreateFirewallPolicy`, `UpdateFirewallDescription`, and the `Associate*`/`Disassociate*` operations:

- No `Create*` operation in this service models an "already exists" error; `InvalidRequestException` is the correct modeled fit for a name collision.
- `InvalidTokenException` (`{Message}`, HTTP 400) is the modeled conflict error for a stale/mismatched `UpdateToken` across all `UpdateFirewall*`, `Associate*` and `Disassociate*` operations.
- `UpdateToken` is optional everywhere it appears; the documented contract is "omit it to apply unconditionally, supply it to get optimistic-locking behavior."

`UpdateRuleGroup` and `UpdateFirewallPolicy` also require `UpdateToken` per the model and have the same unenforced-token gap, but that is out of scope here per the issue's explicit scoping (it calls out only the firewall's 6 Update + 4 Associate/Disassociate ops) and is left for a follow-up.

## How verified

- Extended `NetworkFirewallIntegrationTest` with new cases: name-collision on `CreateFirewall` returns `InvalidRequestException`; `DescribeFirewall` exposes the current `UpdateToken`; `UpdateFirewallDescription` and `AssociateSubnets` succeed with the current token and rotate it; both reject a stale token with `InvalidTokenException` and leave stored state untouched; `UpdateFirewallDescription` without a token still applies unconditionally; `AssociateFirewallPolicy` rejects a stale token and now returns `UpdateToken` in its response.
- Updated the existing `CreateRuleGroup` duplicate-name test to expect `InvalidRequestException`.
- `NetworkFirewallConcurrencyTest` (calls with no `UpdateToken`) still passes unchanged, confirming the optional-token contract.
- `./mvnw test -Dtest=NetworkFirewallIntegrationTest,NetworkFirewallConcurrencyTest`: 45 tests, 0 failures.

## Checklist

Closes #2600

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/networkfirewall/NetworkFirewallService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/networkfirewall/NetworkFirewallService.java
@@ -126,23 +126,27 @@ public class NetworkFirewallService {
         firewall.put("DeleteProtection", request.path("DeleteProtection").asBoolean(true));
         firewall.put("AvailabilityZoneChangeProtection",
                 request.path("AvailabilityZoneChangeProtection").asBoolean(false));
+        rotateToken(firewall);
         firewalls.put(firewallArn, firewall);
         return firewallResponse(firewall, region);
     }
 
-    public ObjectNode describeFirewall(String firewallArn, String firewallName, String region, String accountId) {
+    public synchronized ObjectNode describeFirewall(String firewallArn, String firewallName, String region,
+                                                    String accountId) {
         requireIdentifier(firewallArn, firewallName);
         ObjectNode firewall = find(firewalls, firewallArn, firewallName, "FirewallArn", "FirewallName");
         if (firewall == null) {
             throw notFound("Firewall", firewallArn == null ? firewallName : firewallArn);
         }
-        return firewallResponse(firewall, region);
+        ObjectNode response = firewallResponse(firewall, region);
+        response.put("UpdateToken", ensureToken(firewall));
+        return response;
     }
 
     /**
      * Each UpdateFirewall* operation models exactly one mutable field (botocore
-     * 2020-11-12). Anything else in the raw request — including another operation's
-     * field, or unmodeled members like SubnetMappings — must not be persisted, matching
+     * 2020-11-12). Anything else in the raw request, including another operation's
+     * field, or unmodeled members like SubnetMappings, must not be persisted, matching
      * AWS ignoring unmodeled request members and scoping each op to its own field.
      */
     private static final Map<String, String> UPDATE_ACTION_FIELDS = Map.of(
@@ -161,10 +165,19 @@ public class NetworkFirewallService {
      */
     private static final Set<String> CLEARED_WHEN_OMITTED = Set.of("UpdateFirewallDescription");
 
-    public ObjectNode updateFirewall(String action, JsonNode request, String region, String accountId) {
+    /**
+     * Synchronized so that validating the caller's token against the firewall's current
+     * one and rotating to a new token happen as a single atomic step. Without a common
+     * lock, two overlapping calls can each read and match the same current token before
+     * either commits its rotation, letting both changes apply when the second one should
+     * have been rejected as stale. See {@link #associateSubnets} for the sibling case
+     * this reuses the same lock for.
+     */
+    public synchronized ObjectNode updateFirewall(String action, JsonNode request, String region, String accountId) {
         String arn = textOrNull(request, "FirewallArn");
         String name = textOrNull(request, "FirewallName");
         ObjectNode existing = require(firewalls, arn, name, "Firewall", "FirewallArn", "FirewallName");
+        requireCurrentToken(existing, request);
         String field = UPDATE_ACTION_FIELDS.get(action);
         JsonNode value = request.get(field);
         if (value != null) {
@@ -172,6 +185,7 @@ public class NetworkFirewallService {
         } else if (CLEARED_WHEN_OMITTED.contains(action)) {
             existing.remove(field);
         }
+        String newToken = rotateToken(existing);
         firewalls.put(existing.path("FirewallArn").asText(), existing);
 
         // Each UpdateFirewall* response is a flat {FirewallArn, FirewallName,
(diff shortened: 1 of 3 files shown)


pgermosen:
Thanks for closing out #2600, and the AWS-model verification on the `InvalidRequestException` vs `ResourceAlreadyExistsException` swap is spot on.

Found two issues in the `UpdateToken` enforcement itself, both of which undercut what this PR is trying to guarantee:

1. The check-and-rotate isn't atomic on most of the ops. `updateFirewall` (and the other five `UpdateFirewall*` variants) and `associateFirewallPolicy` aren't `synchronized`, unlike the four subnet/AZ association methods. Two concurrent calls carrying the same current token can both pass `requireCurrentToken` before either one rotates it, so both succeed and the second silently overwrites the first — the exact race `UpdateToken` exists to prevent, on 7 of the 11 ops this PR covers.

2. `DescribeFirewall` reports `UpdateToken` as `""` for a firewall that's never been updated (line 141's `.asText()` default), but `requireCurrentToken` compares against `.asText(null)`. A caller who describes one of these firewalls and submits back the `""` they were shown gets rejected with `InvalidTokenException` — breaking the describe-then-submit flow for exactly the firewalls that predate this change.

Given AWS models `InvalidTokenException` as the client-fault error on all of `UpdateFirewallDescription`/`AssociateSubnets`/`AssociateFirewallPolicy`, the intent here is right — just needs the synchronization applied consistently across all mutating ops and the token default aligned between `DescribeFirewall` and `requireCurrentToken`.

omatheusmesmo:
@pgermosen Both points addressed: `updateFirewall` and `associateFirewallPolicy` are now `synchronized` alongside the four existing association ops, so token validation and rotation are atomic on every mutating call; `DescribeFirewall` now backfills and persists (via `firewalls.put`) a real token for firewalls that predate UpdateToken support, matching `requireCurrentToken`'s null-based default instead of exposing an empty string. Automated review is clean now, ready whenever you have time to take another look.

pgermosen (approved):
Checked both fixes directly, they're solid. updateFirewall and associateFirewallPolicy are synchronized now alongside the mapping ops, and the DescribeFirewall backfill persists the token through firewalls.put so it survives a restart, not just an in-memory patch. Good catch on that last part too, that would've bitten someone eventually. Approving.

PR #3199, opened 2026-09-07
fix(sfn): fail States.Runtime on an unresolvable JSONPath in a payload template

### LINKED ISSUE #2521: [BUG] Step Functions (SFN): unresolvable JSONPath in a payload template yields null instead of failing with States.Runtime

omatheusmesmo:
## Summary

A payload template field (`"name.$": "$.some.path"`) whose JSONPath does not resolve against the effective input currently returns `null` and lets the state succeed. Real AWS and Step Functions Local fail the execution with `States.Runtime`, naming the unresolvable path and field in the cause.

`resolveParameters` (shared by `Parameters`, `ResultSelector`, `ItemSelector`/`ItemTransform`, and `ItemReader` `Parameters`) now routes each `.$` field through a path walk that distinguishes an out-of-range array index (kept as a documented AWS leniency, still resolves to null) from every other unresolved reference (missing object key at any depth, indexing into a non-container), which now raises `States.Runtime` with cause:

`The JSONPath '<path>' specified for the field '<key>' could not be found in the input '<input>'`

Wildcard paths (`$[*]...`) keep their existing null-collapsing behavior, since projections already filter their own misses.

This is a separate code path from PR #3002's `MissingIntrinsicArgumentException` mechanism for `States.*` intrinsic function arguments (which already fails on any miss with no leniency); the two were kept independent rather than merged.

Closes #2521

## Type of change

- [x] Bug fix (`fix:`)

## AWS Compatibility

Per the Amazon States Language input/output processing documentation, a `.$` field whose JSONPath does not resolve against the effective input must fail the state with error `States.Runtime`. Verified against the issue's exact repro (compared to Step Functions Local 2.0.0's `executionFailedEventDetails.cause` wording) and matched that cause format.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -2871,14 +2871,17 @@ public class AslExecutor {
                     String path = val.asText();
                     if (path.startsWith("$$.")) {
                         // Context reference: $$. → resolve against context as $.
-                        resolved.set(realKey, resolvePath("$." + path.substring(3), context));
+                        String contextPath = "$." + path.substring(3);
+                        resolved.set(realKey, context == null
+                                ? NullNode.getInstance()
+                                : resolvePayloadTemplateReference(contextPath, context, null, key, context));
                     } else if ("$$".equals(path)) {
                         resolved.set(realKey, context);
                     } else {
                         // Pass the Context Object through so a $$. reference nested inside an
                         // intrinsic (e.g. States.Format(..., $$.Map.Item.Value.x)) can resolve it;
                         // for a plain $. or States.* input reference, context is simply ignored.
-                        resolved.set(realKey, resolvePath(path, input, context));
+                        resolved.set(realKey, resolvePayloadTemplateReference(path, input, context, key, input));
                     }
                 } else if (val.isObject() || val.isArray()) {
                     resolved.set(key, resolveParameters(val, input, context));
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Nice catch on the original issue, and good handling of the follow-up too. Traced through walkPathTracked and the overflow fix on parseArrayIndex, both check out — a missing object key or a step through a non-container now fails with States.Runtime naming the path and field, while an out-of-range index (including one too large to parse) still collapses to null the way AWS actually behaves. Good test coverage on both. Approving.

PR #3200, opened 2026-09-07
feat(kinesis): implement UpdateShardCount API

### LINKED ISSUE #3064: [FEAT] kinesis: UpdateShardCount API support
## Service

Kinesis

## API Action / Feature

UpdateShardCount

## AWS Documentation

https://docs.aws.amazon.com/kinesis/latest/APIReference/API_UpdateShardCount.html

## Why is this needed?

I have a service that I want to make sure can adapt it's data processing behavior on across an `UpdateShardCount` action. Right now I have to rely on other paid emulators to run these automated test cases.

- [ ] Yes
- [X] No

omatheusmesmo:
## Summary

Implements the Kinesis `UpdateShardCount` action, which was previously unimplemented (fell through to `UnsupportedOperation`). Closes #3064.

`ScalingType` accepts only `UNIFORM_SCALING`, matching AWS. Reshard is expressed as a sequence of the existing `splitShard`/`mergeShards` operations rather than a new resharding algorithm:

- Scale up: the widest open shards are split (via `splitShard`) until the target count is reached, so shard lineage (`ParentShardId`) is produced by the same code path `SplitShard` uses.
- Scale down: open shards are paired up by a fixed (0,1), (2,3), ... adjacency (by starting hash key) so any subset of pairs is always mergeable without conflicts, and the narrowest-combined-width pairs are merged first (via `mergeShards`) until the target count is reached, giving correct `ParentShardId`/`AdjacentParentShardId` lineage.

Validation enforces AWS's documented default limits for this API:
- `TargetShardCount` must not exceed double, or fall below half (rounded up), the current open shard count -> `LimitExceededException`. The round-up direction for an odd current count is floci's own conservative choice: AWS's docs say only "below half" with no stated rounding, and this is now called out explicitly in docs/services/kinesis.md and the method's Javadoc rather than implied as observed AWS behavior.
- `TargetShardCount` must not exceed 10000 shards per stream, and a stream already above that limit can only be scaled further down -> `LimitExceededException`.
- No more than 10 `UpdateShardCount` calls per stream in a rolling 24-hour window -> `LimitExceededException`. AWS separately documents a 10 TPS call-rate limit on this action; that one is intentionally unsimulated, matching floci's lack of real-time request throttling elsewhere, and this is now documented explicitly too.
- The stream must be `PROVISIONED` (not `ON_DEMAND`) -> `ValidationException`.
- The stream must be `ACTIVE` -> `ResourceInUseException`.
- `ScalingType` other than `UNIFORM_SCALING`, or a malformed `TargetShardCount` -> `InvalidArgumentException`.

Response returns `StreamName`, `StreamARN`, `CurrentShardCount` (the open shard count before scaling, matching the documented example), and `TargetShardCount`.

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

Verified against the `UpdateShardCount` API reference (https://docs.aws.amazon.com/kinesis/latest/APIReference/API_UpdateShardCount.html) for request/response shape, the documented default limits, and the modeled error list. The resharding dev guide (https://docs.aws.amazon.com/streams/latest/dev/kinesis-using-sdk-java-resharding.html) confirms resharding is always pairwise (one parent splits into two children, or two parents merge into one child), which is exactly what the existing `SplitShard`/`MergeShards` machinery already implements and this change reuses.

## Checklist

### DIFF
--- a/docs/services/kinesis.md
+++ b/docs/services/kinesis.md
@@ -36,6 +36,7 @@
 | `DisableEnhancedMonitoring` | - |
 | `UpdateStreamMode` | - |
 | `UpdateMaxRecordSize` | - |
+| `UpdateShardCount` | - |
 <!-- floci:actions:end -->
 
 ## Local Inspection Endpoints
@@ -71,6 +72,24 @@ A `LATEST` iterator is positioned at the shard tip at the moment the iterator is
 ## Record Routing
 
 `PutRecord` and `PutRecords` honor `ExplicitHashKey` when it is provided. The value must be a decimal integer in the Kinesis hash-key space, and records are written to the open shard whose `HashKeyRange` contains that value. Without `ExplicitHashKey`, Floci keeps using the partition key to choose a shard.
+
+## Shard Scaling (UpdateShardCount)
+
+`UpdateShardCount` enforces the documented default limits: at most double the current
+open shard count per call, at most ten calls per rolling 24-hour period per stream, a
+10000-shard ceiling, and the asymmetric rule that a stream already above 10000 shards can
+only move to a target strictly below it.
+
+The minimum bound (`TargetShardCount` cannot go below half the current open shard count)
+rounds up for an odd current count, so `current=5` accepts `target=3` but rejects
+`target=2`. AWS's documentation says only "below half" with no stated rounding, so which
+direction an odd count rounds is Floci's own call rather than observed AWS behavior; this
+is the more conservative reading (fewer shards allowed, not more).
+
+AWS also documents a separate 10 TPS call-rate limit on this action ("Make over 10 TPS.
+TPS over 10 will trigger the LimitExceededException"), independent of the ten-calls-per-24h
+limit above. Floci does not simulate real-time request throttling for this or any other
+action, so that limit is intentionally unenforced here.
 
 ## Configuration
 
--- a/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/kinesis/KinesisJsonHandler.java
@@ -72,6 +72,7 @@ public class KinesisJsonHandler {
             case "DisableEnhancedMonitoring" -> handleDisableEnhancedMonitoring(request, region);
             case "UpdateStreamMode" -> handleUpdateStreamMode(request, region);
             case "UpdateMaxRecordSize" -> handleUpdateMaxRecordSize(request, region);
+            case "UpdateShardCount" -> handleUpdateShardCount(request, region);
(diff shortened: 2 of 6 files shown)


pgermosen:
Nice implementation, the scaling logic and validation limits all look right. One thing blocking CI though: docs/services/kinesis.md's action table is out of sync, it's missing the UpdateShardCount row. Can you run make docs-sync and push the result?

omatheusmesmo:
Done, ran make docs-sync and pushed - kinesis.md now lists UpdateShardCount.

pgermosen (commented):
Nice work, thanks for the docs-sync fix earlier. I went through the implementation against the actual UpdateShardCount API reference rather than just the PR description, and it holds up well: the response shape matches AWS's documented example exactly, all four error codes (InvalidArgumentException, LimitExceededException, ValidationException, ResourceInUseException) match their documented meaning, and the double/half shard-count limits are right, including the trickier asymmetric rule where a stream already above 10000 shards can only move to a target strictly below 10000. The locking between updateShardCount and the splitShard/mergeShards calls it makes internally is also correctly reentrant, and computing the split/merge candidates up front from the pre-scaling snapshot is the right call since it avoids any risk of the operations interfering with each other mid-sequence.

Two things worth a reply before this is done. The AWS docs list a separate limit alongside the "ten scales per rolling 24 hours" one you implemented: a 10 TPS call-rate limit that also raises LimitExceededException. It's not addressed here. I'd guess it's fine to skip since this codebase doesn't simulate real-time request throttling anywhere else either, but it'd be good to say that explicitly rather than leave it silently uncovered.

Second, the "half rounded up" minimum you're enforcing (current=5 rejects target=2 but allows target=3) matches what I'd expect, but AWS's own text just says "below half" without spelling out the rounding for odd counts, so I don't have a doc citation to confirm the direction. Can you say where that came from, real-account testing or another source? Would help close the loop on that one.

omatheusmesmo:
@pgermosen Good catches. Pushed: the 10 TPS limit is intentionally unsimulated (floci doesn't do real-time throttling anywhere), now called out explicitly in docs/services/kinesis.md and the method's Javadoc. On the rounding: no doc citation, AWS only says "below half" with no stated direction for odd counts - the round-up is floci's own conservative choice, now documented as such rather than implied as observed AWS behavior.

pgermosen (approved):
Both open questions from last round are answered properly now, not just replied to: the docs and the Javadoc both spell out that the 10 TPS limit is intentionally unenforced (matching the rest of the codebase) and that the half-rounding direction is Floci's own conservative call rather than confirmed AWS behavior. Nothing else changed and the core logic already checked out clean against the API reference. Approving.

PR #3204, opened 2026-09-07
fix(sts): validate SAML assertions before assuming roles

nkootstra:
## Summary

Validate SAML assertions before issuing credentials for `AssumeRoleWithSAML`, including signature, issuer, audience, expiry, recipient, role/principal pair, and registered provider checks. Add minimal SAML provider persistence and AWS Query API support for create, list, and get operations.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, malformed, unsigned, expired, tampered, or otherwise invalid SAML assertions could reach role assumption. The implementation now returns AWS-compatible `InvalidIdentityToken` or `AccessDenied` errors while preserving the STS Query protocol, XML response format, and valid SAML credential response fields.

## Checklist

- [ ] `./mvnw test` passes locally — full suite timed out after 10 minutes in unrelated API Gateway/Lambda container integration tests
- [x] Focused tests pass: `./mvnw -q -Dtest=AssumeRoleWithSamlValidationIntegrationTest,AssumeRoleTrustPolicyIntegrationTest,IamIntegrationTest test`

Additional validation passed:

- `git diff --check`

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/StsTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/StsTest.java
@@ -1,10 +1,46 @@
 package com.floci.test;
 
 import org.junit.jupiter.api.*;
+import org.bouncycastle.asn1.x500.X500Name;
+import org.bouncycastle.cert.X509CertificateHolder;
+import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
+import org.bouncycastle.cert.jcajce.JcaX509v3CertificateBuilder;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
+import software.amazon.awssdk.services.iam.IamClient;
+import software.amazon.awssdk.services.iam.model.CreateRoleRequest;
+import software.amazon.awssdk.services.iam.model.CreateSamlProviderRequest;
+import software.amazon.awssdk.services.iam.model.CreateSamlProviderResponse;
 import software.amazon.awssdk.services.sts.StsClient;
 import software.amazon.awssdk.services.sts.model.*;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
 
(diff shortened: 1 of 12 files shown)


pgermosen:
Thanks for taking this on, SAML support is a meaningful gap to close. Went through the new evaluator and provider service, and there are a few things that need fixing before this can merge, the first one especially.

`SAMLTrustPolicyEvaluator.conditionsSatisfied` only handles `StringLike` and `StringEquals`. Any other operator, including `StringNotEquals`/`StringNotLike`, always evaluates to a non-match, which makes the whole condition fail, which makes the statement itself not apply. So a trust policy that pairs a broad Allow with a conditional Deny using one of those operators has its Deny silently dropped, and the Allow goes through anyway. That's a real hole in the exact thing this PR is supposed to be locking down.

`GetSAMLProvider` also crosses accounts: `IamQueryHandler.handleGetSAMLProvider` hands the raw request ARN straight to `SAMLProviderService.get()`, which picks the storage partition from whatever account is embedded in that ARN rather than the authenticated caller. Account A can ask for account B's provider ARN and get back B's entity ID and certificate.

And `SAMLProviderService.create()` doesn't check for an existing provider before writing, so creating one with a name that's already taken silently replaces its certificate instead of returning `EntityAlreadyExists` the way the OIDC provider path does. That's also going to make any assertion already signed against the old cert start failing without warning.

Greptile's other notes (the `SubjectConfirmationData` method check, the shared `XmlParser`, brace style) are worth picking up too, but the three above are the ones I'd want fixed first.

pgermosen (approved):
Went back through all three security findings directly in the code rather than trusting the resolved threads. All genuinely fixed: the trust policy evaluator now correctly implements StringNotEquals/StringNotLike as the real negation instead of silently dropping the condition, GetSAMLProvider now scopes the lookup to the authenticated caller's account, and provider creation rejects a duplicate with EntityAlreadyExists under a lock that also closes the check-then-write race. Nice turnaround. Approving.

PR #3206, opened 2026-09-07
fix(cloudtrail): implement CreateTrail with tagging support and add integration tests

### LINKED ISSUE #2800: cloudtrail: ListTags is not supported
`terraform-provider-aws`'s read/refresh for `aws_cloudtrail` calls the CloudTrail `ListTags` action (to populate tags, as part of its default refresh cycle regardless of whether `tags` is set on the resource). Floci's CloudTrail handler doesn't support this action at all:

```
Error: listing tags for CloudTrail Trail ...: operation error CloudTrail: ListTags,
       https response error StatusCode: 400, api error InvalidAction: Could not find operation ListTags
```

Confirmed still missing on `main` — no `ListTags` handling under `services/cloudtrail`.

Note: this is distinct from the `CreateTrail`/`UnknownOperationException` gap already fixed by #1460 — that fix only covers trail creation, not the tag-refresh path.

Blocks the Terraform/OpenTofu compatibility suites from including `aws_cloudtrail` (removed from both in #1834 to unblock the suite; should be re-added once this lands).

Origin: found while getting #1834 green, after the CloudTrail CreateTrail conflict (resolved by rebasing onto post-#1460 main) cleared and the suite reached the trail's post-create refresh for the first time.

matheusandre1:
## Summary
Closes : #2800 

Here is what I found: The `ListTags` action reported as missing in the issue is already in `main`, having been added via PR #2934; the specific error described in the issue no longer occurs. However, the Terraform workflow was still failing for a related reason: the provider sends tags within the `CreateTrail` call (in the `TagsList` field) but never calls `AddTags` during creation. Floci was discarding this field, causing the `ListTags` call during the post-creation update to return an empty result; consequently, any `aws_cloudtrail` resource with tags resulted in a perpetual diff.
## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/compatibility-tests/compat-opentofu/main.tf
+++ b/compatibility-tests/compat-opentofu/main.tf
@@ -675,3 +675,24 @@ resource "aws_transfer_server" "compat" {
 output "transfer_server_id" {
   value = aws_transfer_server.compat.id
 }
+
+# -- CloudTrail Trail (issue #2800) --------------------------------------------
+# The provider sends tags inside CreateTrail and reads them back with ListTags on
+# every refresh, so this covers the post-create tag refresh path end to end.
+resource "aws_s3_bucket" "trail_logs" {
+  bucket        = "floci-compat-trail-logs"
+  force_destroy = true
+}
+
+resource "aws_cloudtrail" "compat" {
+  name           = "floci-compat-trail"
+  s3_bucket_name = aws_s3_bucket.trail_logs.id
+
+  tags = {
+    Environment = "compat-test"
+  }
+}
+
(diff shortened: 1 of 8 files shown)


pgermosen (approved):
Good root-cause work tracing this past the original ListTags symptom to the real gap in CreateTrail. The 50-tag limit and TagsLimitExceededException match AWS exactly, and validating before persisting the trail is the right call. Nice to see Terraform/OpenTofu compat coverage added too. Approving.

PR #3207, opened 2026-09-07
fix(swf): improve polling mechanism with timeout and backoff strategy

### LINKED ISSUE #3188: [BUG] SwfServiceTest.closingAChildWhileTheParentIsMutated_keepsBothHistoriesIntact flakes under CI load
## Service

SWF

## AWS API Action

StartChildWorkflowExecution / PollForDecisionTask (child-workflow decision task scheduling)

## Expected behavior

`SwfServiceTest.closingAChildWhileTheParentIsMutated_keepsBothHistoriesIntact` passes
deterministically: after the parent's decision task schedules 6 child workflows via
`StartChildWorkflowExecution`, each child's initial decision task should become pollable
promptly so 6 concurrent threads can each claim one and close it.

## Actual behavior

Intermittent failure in CI, unrelated to whatever PR happens to trigger the run:

```
[ERROR] io.github.hectorvent.floci.services.swf.SwfServiceTest.closingAChildWhileTheParentIsMutated_keepsBothHistoriesIntact -- Time elapsed: 0.032 s <<< ERROR!
java.util.concurrent.ExecutionException: java.lang.AssertionError: no decision task for wf-cc-child-5
	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:124)
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:207)
	at io.github.hectorvent.floci.services.swf.SwfServiceTest.closingAChildWhileTheParentIsMutated_keepsBothHistoriesIntact(SwfServiceTest.java:449)
Caused by: java.lang.AssertionError: no decision task for wf-cc-child-5
	at io.github.hectorvent.floci.services.swf.SwfServiceTest.pollFor(SwfServiceTest.java:1222)
```

Observed on PR #3048's `Build and Test (shard 2)` run, on a branch that only touches
`src/main/java/.../services/s3/PreSignedUrlFilter.java`, `S3VirtualHostFilter.java`, and its
own test — no SWF code in the diff:
https://github.com/floci-io/floci/actions/runs/34125462200/job/101752988811

Possible root cause worth checking: `pollFor` (SwfServiceTest.java:1201-1222) retries up to
200 times, sleeping 5ms only on an *empty* poll — when it repeatedly claims a sibling's task
instead (a likely outcome with 6 threads polling the same `tl` task list at once), it loops
back immediately with no sleep, so the 200-attempt budget can burn through in well under a
second under CI contention. If a child's `StartChildWorkflowExecution` decision task isn't
schedulable that fast, the poller for that child exhausts its attempts and fails, even though
the task would show up shortly after.

## Reproduction

Not reliably reproducible locally; run the full `SwfServiceTest` under load (e.g. in parallel
with the rest of the suite, as CI does) enough times, or add loop iterations, to hit the race.

```bash
./mvnw test -Dtest=io.github.hectorvent.floci.services.swf.SwfServiceTest#closingAChildWhileTheParentIsMutated_keepsBothHistoriesIntact
```

## Environment

- Floci version / image tag: main branch, observed 2026-09-07
- Java SDK version (if applicable): N/A (JUnit test failure, not an SDK client issue)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): GitHub Actions CI (`./mvnw test`)

matheusandre1:
## Summary

Fixes: #3188 
In SwfService, the six child decision tasks are scheduled synchronously within `respondDecisionTaskCompleted`, and releasing a sibling task with zero decisions causes it to be immediately rescheduled; thus, the service is functioning correctly. The root cause is the hypothesis raised in the issue: `pollFor` had a fixed limit of 200 attempts and only slept when a poll returned empty. With six threads stealing and returning tasks to one another within the same task list, each attempt takes mere microseconds, and the limit is exhausted in ~30ms—exactly the duration recorded in the CI failure.

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

- [X ] `./mvnw test` passes locally

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/swf/SwfServiceTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/swf/SwfServiceTest.java
@@ -58,6 +58,8 @@ class SwfServiceTest {
 
     private static final String REGION = "us-east-1";
     private static final String DOMAIN = "unit-domain";
+    /** Upper bound for {@link #pollFor}; generous so a loaded CI runner cannot starve it. */
+    private static final Duration POLL_TIMEOUT = Duration.ofSeconds(10);
 
     private MutableClock clock;
     private SwfService service;
@@ -1199,9 +1201,15 @@ class SwfServiceTest {
      *
      * <p>Tolerates a transient empty poll so the concurrency tests can call this from
      * several threads: another thread may hold the task this caller wants at that instant.
+     *
+     * <p>Bounded by wall-clock time rather than an attempt count: with several pollers on
+     * one task list, a caller can repeatedly claim and release siblings' tasks, and those
+     * round trips take microseconds. A fixed attempt budget then burns out well before the
+     * wanted task becomes pollable under CI contention.
      */
     private SwfDecisionTask pollFor(String workflowId) {
-        for (int attempt = 0; attempt < 200; attempt++) {
+        long deadline = System.nanoTime() + POLL_TIMEOUT.toNanos();
+        while (System.nanoTime() < deadline) {
             Optional<SwfDecisionTask> claimed = service.pollForDecisionTask(REGION, DOMAIN, "tl", "d");
             if (claimed.isPresent()) {
                 SwfDecisionTask task = claimed.get();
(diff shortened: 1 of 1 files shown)


pgermosen (approved):
Good diagnosis, matches the issue's hypothesis exactly. Checked that the new 10s wall-clock bound in pollFor still comfortably fits inside the caller's 20s future timeout, so this trades a fast flaky failure for a reliable pass without introducing a new hang risk. Approving.

PR #3211, opened 2026-09-07
fix(cloudwatch): wrap TagResource and UntagResource in their declared Result element

lex00:
## Summary

Ports lex00/floci#89. CloudWatch's `TagResource` and `UntagResource` have an empty output structure in the API model, and the model still declares a `resultWrapper` for it. Floci answered both with a bare `TagResourceResponse` and no Result child. The AWS Go SDK v2 Query unmarshaler behind the Terraform AWS provider rejects that with "TagResourceResult node not found", so tagging an alarm from Terraform failed client side even though the tag had landed.

Both responses now go through `envelopeEmptyResult`, the helper `DeleteDashboards` already uses for the same shape. Nothing changes for the operations whose model has no output shape at all, such as `PutMetricAlarm`, which keep the bare response.

Unit tests assert the wrapper on both responses and that the tag write still lands. An integration test drives the Query protocol against a real alarm and checks the tag through a full tag and untag cycle.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior. AWS returns `<TagResourceResponse><TagResourceResult/><ResponseMetadata>...` for a successful `TagResource`. Floci omitted the Result element, which the Go SDK treats as a deserialization failure.

## Checklist

- [x] `./mvnw test` passes locally for the CloudWatch tags and dashboards tests

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudwatch/metrics/CloudWatchMetricsQueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudwatch/metrics/CloudWatchMetricsQueryHandler.java
@@ -285,7 +285,10 @@ public class CloudWatchMetricsQueryHandler {
         } else {
             metricsService.tagResource(arn, tags, region);
         }
-        return Response.ok(AwsQueryResponse.envelopeNoResult("TagResource", null)).build();
+        // TagResourceOutput is an empty structure with a declared resultWrapper, so the
+        // wrapper element must be present. The AWS Go SDK v2 unmarshaler, behind the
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Confirmed against the actual CloudWatch model - TagResource and UntagResource both declare a resultWrapper even though the output shape is empty, same as DeleteDashboards, while PutMetricAlarm correctly stays untouched since it has no output shape at all. Clean, minimal fix reusing the existing helper. Approving.

PR #3213, opened 2026-09-07
feat(s3): support bucket analytics and inventory configurations

lex00:
## Summary

Ports the S3 half of lex00/floci#42. #2382 landed the metrics and intelligent-tiering bucket configurations, and this adds the remaining two sub-resources in the same shape. `PutBucketAnalyticsConfiguration` and `PutBucketInventoryConfiguration` store the configuration on the bucket record under its id. The single and list reads return it, and the deletes remove it with `NoSuchConfiguration` for an unknown id. Terraform's `aws_s3_bucket_analytics_configuration` and `aws_s3_bucket_inventory` resources read back what they wrote.

Each body is validated against the published schema and re-serialized in AWS element order, the way the two sibling records do it. Analytics accepts the same filters as metrics plus a `StorageClassAnalysis` with an optional `DataExport`. Inventory checks every required element and the destination with its optional `SSE-S3` or `SSE-KMS` encryption. A body whose id disagrees with the query string is `MalformedXML`, as on AWS. The PUT and DELETE forms are routed ahead of the unqualified CreateBucket and DeleteBucket so neither can fall through and create or destroy the bucket.

Unit tests cover both parsers. Integration tests drive the REST XML protocol for each sub-resource and include the fall-through regressions. `S3ServiceTest` checks that a record persisted before these fields existed still loads.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Shapes follow the S3 API reference for the eight actions. The list responses use `ListBucketAnalyticsConfigurationResult` and `ListInventoryConfigurationsResult` with `IsTruncated` false, since Floci returns every configuration in one page. Before this change the four PUT forms fell through to CreateBucket and answered `BucketAlreadyOwnedByYou`.

## Checklist

- [x] `./mvnw test` passes locally for the S3 configuration suites and `S3ServiceTest` with 164 tests

### DIFF
--- a/docs/services/s3.md
+++ b/docs/services/s3.md
@@ -26,6 +26,8 @@
 | **Public Access Block** | PutPublicAccessBlock, GetPublicAccessBlock, DeletePublicAccessBlock |
 | **Metrics** | PutBucketMetricsConfiguration, GetBucketMetricsConfiguration, ListBucketMetricsConfigurations, DeleteBucketMetricsConfiguration |
 | **Intelligent-Tiering** | PutBucketIntelligentTieringConfiguration, GetBucketIntelligentTieringConfiguration, ListBucketIntelligentTieringConfigurations, DeleteBucketIntelligentTieringConfiguration |
+| **Analytics** | PutBucketAnalyticsConfiguration, GetBucketAnalyticsConfiguration, ListBucketAnalyticsConfigurations, DeleteBucketAnalyticsConfiguration |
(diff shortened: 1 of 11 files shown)


hectorvent (approved):
Checked the eight operations against the S3 model: routes, the `id` query parameter separating single reads from lists, the required members of both configuration shapes, and `NoSuchConfiguration` on an unknown id all line up. Approving.

PR #3215, opened 2026-09-07
fix(ec2): store instance metadata options and add ModifyInstanceMetadataOptions

lex00:
## Summary

Ports the EC2 half of lex00/floci#116. The SSM half merged as #3173.

`RunInstances` accepted `MetadataOptions.*` and dropped it, and `DescribeInstances` answered a hardcoded block with `httpTokens` always `optional`. terraform-aws-modules' ec2-instance module defaults to `http_tokens = required`, so every apply produced a permanent second plan.

The launch now stores the options it is given, with AWS's launch defaults filling whatever the request left out. A launch template's metadata options seed the launch and explicit request fields override them. `DescribeInstances` reads the stored values back, and a record persisted before this change reads as the defaults. `ModifyInstanceMetadataOptions` is new. It changes only the fields it names and returns the `instanceMetadataOptions` shape the SDK unmarshals. A value outside the AWS enumerations or a hop limit outside 1 to 64 is `InvalidParameterValue`.

One difference from the fork. It also modelled ModifyInstanceMetadataDefaults, which upstream does not have, so the region-level default layer is left out here.

Unit tests cover the launch defaults and an explicit launch. Two more cover a partial modify and the validation errors. The integration test drives the same through the Query protocol, including a launch template supplying the options.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior. AWS returns the metadata options a launch requested and lets `ModifyInstanceMetadataOptions` change them afterwards. Floci returned the same defaults for every instance and had no modify action.

## Checklist

- [x] `./mvnw test` passes locally for the EC2 service and handler tests plus the EC2 integration suites

### DIFF
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -169,6 +169,7 @@ Floci seeds the following resources on first use in each region so Terraform, th
 | DescribeInstanceStatus | Returns status records for stored instances. |
 | DescribeInstanceAttribute | Returns a supported attribute for an instance. |
 | ModifyInstanceAttribute | Updates supported mutable attributes for an instance. |
+| ModifyInstanceMetadataOptions | Updates an instance's IMDS options, changing only the fields the request names. |
 
 ### VPCs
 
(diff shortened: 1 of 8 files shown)


hectorvent (approved):
Checked against the EC2 model: every `MetadataOptions` member, the three enums, the 1 to 64 hop-limit range and the `state applied` on the modify response line up, and the launch defaults are AWS's. Approving. The shard 3 red is the Organizations leak (`AlreadyInOrganizationException` across 29 cases), not this change.

PR #3216, opened 2026-09-07
fix(compat): retry RedshiftTest's SDK calls past a brief connection-refused hiccup

### LINKED ISSUE #2789: RedshiftTest intermittently fails CI with connection-refused, unrelated to the PR under test
## Summary

`RedshiftTest.testCreateCluster` and `RedshiftTest.testDeleteCluster` (in
`compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java`)
intermittently fail the `native / sdk-test-java` CI job with a TCP-level connection failure,
on PRs that never touch Redshift or anything network-related.

## Observed failures

Both occurrences show the identical signature — both tests in the class fail within ~1.4s of
each other, right at the start of their run:

```
[ERROR] io.github.hectorvent.floci.compat.RedshiftTest.testCreateCluster -- Time elapsed: 0.991 s <<< ERROR!
software.amazon.awssdk.core.exception.SdkClientException: Unable to execute HTTP request: Connect to http://localhost:4566 failed: Connection refused (SDK Attempt Count: 4)

[ERROR] io.github.hectorvent.floci.compat.RedshiftTest.testDeleteCluster -- Time elapsed: 0.428 s <<< ERROR!
software.amazon.awssdk.core.exception.SdkClientException: Unable to execute HTTP request: Connect to http://localhost:4566 failed: Connection refused (SDK Attempt Count: 4)

[ERROR] Tests run: 1656, Failures: 0, Errors: 2, Skipped: 30
```

- **PR #2472** (`feat(redshift): implement Redshift emulation...`) — same two tests, same error.
- **PR #2729** (`test(stepfunctions): restore AslExecutorTaskHistoryEventsTest compilation`) —
  a StepFunctions-only test-compile fix with zero relation to Redshift or networking; same two
  tests failed the same way.

## Why this looks like a CI-side flake, not a product bug

Investigated the PR #2472 run in detail:

- Only these 2 of ~1650+ tests failed in that run.
- The floci server's own internal log shows a single continuous run (`AWS Local Emulator Ready`
  logged once, no restart/crash events) for the whole job.
- A *different*, larger Redshift compat test in the same module
  (`com.floci.test.RedshiftOperationsTest`) ran later in the same job and passed — its container
  creation logs show normal Postgres init proceeding, confirming the server was reachable and
  Redshift provisioning worked correctly at that point.
- `Connection refused` is a TCP-level rejection (nothing listening on the port from the caller's
  perspective at that instant), not an AWS-shaped error response — consistent with a brief
  network/port-publish hiccup between the test runner container and the floci container that
  self-resolves almost immediately (no long backoff was needed for later tests to succeed).

## Suggested next steps

- Check whether `RedshiftTest` runs early enough in the compat suite's execution order that it
  could race a still-warming-up port publish/network alias, even after the "wait for floci to be
  ready" step reports success.
- Consider a short client-side retry/backoff specifically for this test class's first request
  (or reordering it later in the suite) if the root cause can't be pinned down quickly, so this
  stops intermittently failing unrelated PRs' CI.

Mide69:
## Summary

Fixes #2789. `RedshiftTest.testCreateCluster` and `testDeleteCluster` intermittently fail the `native / sdk-test-java` CI job with `Connection refused`, on PRs that never touch Redshift or anything network-related (confirmed on #2472 and #2729).

The AWS SDK's own built-in retries (4 attempts) already ran and still failed - their total backoff is under a second, and the issue's own investigation found the floci server's log showed a single continuous run (no restart/crash) for the whole job, with a *different*, larger Redshift test in the same suite running later and passing. That points to a brief port-publish/network-alias hiccup between the test runner and floci containers that self-resolves almost immediately, just not within the SDK's own sub-second retry window.

Wraps each SDK call in this class in a 30-second retry loop (`withRetry`) that only catches `SdkClientException` - a genuine service-level error (e.g. `ClusterAlreadyExists`) still surfaces immediately rather than being retried into a timeout. This mirrors the retry pattern `RedshiftOperationsTest.awaitPostgresConnection` already uses successfully for its JDBC connection in the same module.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

N/A - this only changes a compatibility test's resilience to a CI-infrastructure race, not product behavior.

## Checklist

- [x] Verified against a local floci instance: `testCreateCluster` and `testDeleteCluster` pass
- [x] Could not reproduce the actual CI-only race locally (same as the issue reporter, who diagnosed it from CI logs only)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
@@ -5,6 +5,7 @@ import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestMethodOrder;
+import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.services.redshift.RedshiftClient;
 import software.amazon.awssdk.services.redshift.model.Cluster;
 import software.amazon.awssdk.services.redshift.model.CreateClusterRequest;
@@ -24,6 +25,10 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.Statement;
 
+import java.time.Duration;
+import java.time.Instant;
+import java.util.function.Supplier;
+
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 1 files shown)


pgermosen:
Good diagnosis on the root cause, but the retry is a bit broader than what you actually found. withRetry catches SdkClientException, which also covers failures that can happen after floci already processed the request, like a response getting reset mid-stream, not just the pre-handshake Connection refused this is meant to paper over. If that happens on createCluster or deleteCluster, the retry resubmits and gets a real ClusterAlreadyExists or ClusterNotFound back, which isn't an SdkClientException, so it isn't caught, and the test fails with a different error instead of the flake actually getting fixed.

Might be worth narrowing the catch to the specific failure you diagnosed, checking the cause chain for ConnectException/Connection refused rather than the whole SdkClientException hierarchy, so a retry only ever happens for the case you know is safe to retry.

Mide69:
Good catch, fixed in 174e86c7. Narrowed the catch to walk the cause chain for a ConnectException (the pre-handshake failure this was diagnosed from); any other SdkClientException now propagates immediately rather than resubmitting.

Verified the discrimination directly rather than just by inspection: an isolated throwaway test against a guaranteed-refused port (localhost:1) showed withRetry looping 6 times over its window with root cause HttpHostConnectException (which is a ConnectException), while a live run against floci with Docker unavailable in my sandbox produced a genuine "Read timed out" that withRetry correctly did NOT retry - it rethrew after a single call, confirmed by the elapsed time matching the SDK's own internal 4-attempt retry duration exactly rather than the 30s window.

pgermosen (approved):
The narrowing is right, walking the cause chain for ConnectException rather than checking the top-level type is what makes it actually catch wrapped forms like HttpHostConnectException while still letting a response-reset-after-processing case through uncaught. What really makes this credible for a flake that can't be reproduced on demand is the verification: an isolated test against a guaranteed-refused port showing the retry loop actually engage, and a separate run showing a genuine read timeout does not get retried, with the elapsed time as the tell. That's real evidence in both directions, not just inspection.

Good to merge.

PR #3217, opened 2026-09-07
fix(cloudfront): stop duplicating CachedMethods into AllowedMethods.Items

pgermosen:
## Summary

Follow-up to #2767 (the TrustedSigners nil-deref crash from that issue was already fixed by #3015). That fix explicitly left two smaller fidelity gaps from the same issue as follow-ups. This PR fixes one of them: `AllowedMethods`/`CachedMethods` handling on `DefaultCacheBehavior` and `CacheBehavior`.

`CachedMethods` is a nested `Quantity`/`Items` object inside `AllowedMethods`, not a sibling element. The request parser tracked a single "inside AllowedMethods" flag across the whole element, so `Method` entries nested inside `CachedMethods` were folded into the parsed `AllowedMethods` list too, and `CachedMethods` itself was silently dropped.

Since `terraform-provider-aws`'s `aws_cloudfront_distribution` requires both `allowed_methods` and `cached_methods` on every cache behavior, this produced a permanent per-plan diff on every floci-managed distribution: `AllowedMethods.Items` came back with duplicates (e.g. `["GET","HEAD","GET","HEAD","OPTIONS"]`), and `CachedMethods` was missing entirely.

The other gap from #2767 (`ForwardedValues`/`MinTTL`/`DefaultTTL`/`MaxTTL` dropped on read-back for the legacy `forwarded_values` cache-behavior path) is a separate, larger change and is filed as #3208. It doesn't affect the modern `cache_policy_id` path, which floci already supports fully.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `GetDistribution`/`GetDistributionConfig` echoed `AllowedMethods.Items` with `CachedMethods`' entries duplicated in, and never echoed `CachedMethods` itself, even though it was accepted on create/update. Verified against the AWS `AllowedMethods` API reference: `CachedMethods` is documented as a nested `Quantity`/`Items` sub-object, optional but with its own item list distinct from the parent's.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
@@ -1972,12 +1972,7 @@ public class CloudFrontController {
         xml.raw(xmlTrustedKeyGroups(
                 dcb.isTrustedKeyGroupsEnabled(), dcb.getTrustedKeyGroups()));
 
-        List<String> allowed = dcb.getAllowedMethods();
-        if (allowed == null || allowed.isEmpty()) {
-            allowed = List.of("GET", "HEAD");
-        }
-        xml.raw(xmlQuantityItems("AllowedMethods", "Method", allowed.size(),
-                allowed.stream().map(m -> "<Method>" + XmlBuilder.escape(m) + "</Method>").toList()));
+        xml.raw(xmlAllowedMethods(dcb.getAllowedMethods(), dcb.getCachedMethods()));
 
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Matches the model (`CachedMethods` nested under `AllowedMethods`, each with its own Quantity/Items). Approving; an ordered `CacheBehavior` case in the test would be a nice addition. Heads-up: #3182 carries the same CachedMethods change and conflicts here, so landing this first and asking that PR to drop its copy keeps both clean.

PR #3219, opened 2026-09-08
fix(autoscaling): validate monitoring booleans

KenkoGeek:
## Summary

- validate `InstanceMonitoring.Enabled` as a strict boolean for Auto Scaling launch configurations
- reject malformed values instead of silently coercing them to `false`
- add integration coverage for invalid monitoring booleans

## AWS Compatibility

Validated against the official Amazon EC2 Auto Scaling `CreateLaunchConfiguration` API/CLI contract, where `InstanceMonitoring.Enabled` is a Boolean (`true` or `false`). Invalid strings must not be accepted as a valid false value.

## Validation

- `AutoScalingIntegrationTest` passes
- `git diff --check origin/main...HEAD` passes

## Context

This change was split out of #3128 per review so the Verified Permissions PR remains service-scoped.


hectorvent (approved):
Thanks for splitting this out. The two call sites now share `parseOptionalBoolean` with the rest of the handler from #3174, so a malformed `InstanceMonitoring.Enabled` is refused instead of silently reading as false. Approving.

PR #3220, opened 2026-09-08
test(stepfunctions): add test for result selector unwrapping payload with bare function name

### LINKED ISSUE #2544: [BUG] `states:::lambda:invoke` returns the bare payload, not the Invoke API response
## Service

Step Functions (Lambda service integration)

## AWS API Action

`StartExecution` — specifically a `Task` state whose `Resource` is the optimized
Lambda integration `arn:aws:states:::lambda:invoke`.

Not affected: a `Task` whose `Resource` is a direct function ARN
(`arn:aws:lambda:<region>:<account>:function:<name>`). That form already matches
AWS.

## Expected behavior

AWS Step Functions sets the Task result to the **Lambda Invoke API response**,
with the function's return value nested under `Payload`:

```json
{
  "ExecutedVersion": "$LATEST",
  "Payload": { "greeting": "hello", "echo": { "n": 1 } },
  "StatusCode": 200
}
```

This is why `"OutputPath": "$.Payload"` and
`"ResultSelector": {"x.$": "$.Payload"}` are the documented way to unwrap an
optimized Lambda integration.

## Actual behavior

Floci sets the Task result to the function's **raw return value**, with no
envelope:

```json
{ "greeting": "hello", "echo": { "n": 1 } }
```

There is no `Payload` key, so `$.Payload` resolves to `null`:

```json
{ "unwrapped": null }
```

**No error is raised.** The execution reports `SUCCEEDED`, and the `null`
propagates into downstream `Choice` / `Pass` states — so a workflow completes
having taken the wrong branch, rather than failing.

## Reproduction

```bash
mkdir -p flocibug && cd flocibug
E=http://localhost:4566          # Floci endpoint
export AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-west-2

cat > handler.py <<'EOF'
def handler(event, context):
    return {"greeting": "hello", "echo": event}
EOF
zip -q fn.zip handler.py

aws lambda create-function --endpoint-url $E --function-name echofn \
  --runtime python3.9 --handler handler.handler \
  --role arn:aws:iam::000000000000:role/r --zip-file fileb://fn.zip
aws lambda wait function-active-v2 --endpoint-url $E --function-name echofn

# (a) direct function ARN — matches AWS today
cat > a.json <<'EOF'
{"StartAt":"T","States":{"T":{"Type":"Task",
 "Resource":"arn:aws:lambda:us-west-2:000000000000:function:echofn","End":true}}}
EOF

# (b) optimized integration, result untouched
cat > b.json <<'EOF'
{"StartAt":"T","States":{"T":{"Type":"Task",
 "Resource":"arn:aws:states:::lambda:invoke",
 "Parameters":{"FunctionName":"echofn","Payload.$":"$"},"End":true}}}
EOF

# (c) optimized integration, unwrapping $.Payload as AWS documents
cat > c.json <<'EOF'
{"StartAt":"T","States":{"T":{"Type":"Task",
 "Resource":"arn:aws:states:::lambda:invoke",
 "Parameters":{"FunctionName":"echofn","Payload.$":"$"},
 "ResultSelector":{"unwrapped.$":"$.Payload"},"End":true}}}
EOF

for m in a b c; do
  aws stepfunctions create-state-machine --endpoint-url $E --name "sm-$m" \
    --role-arn arn:aws:iam::000000000000:role/sfn --definition "file://$m.json" >/dev/null
  EX=$(aws stepfunctions start-execution --endpoint-url $E \
    --state-machine-arn "arn:aws:states:us-west-2:000000000000:stateMachine:sm-$m" \
    --input '{"n":1}' --query executionArn --output text)
  sleep 5
  echo "$m -> $(aws stepfunctions describe-execution --endpoint-url $E \
    --execution-arn "$EX" --query output --output text)"
done
```

Output observed:

```
a -> {"greeting":"hello","echo":{"n":1}}     # matches AWS
b -> {"greeting":"hello","echo":{"n":1}}     # AWS would return the envelope
c -> {"unwrapped":null}                      # AWS: {"unwrapped":{"greeting":"hello",...}}
```

## Environment

- **Floci version / image tag:** `floci/floci:latest`, `FLOCI_VERSION=1.6.0`
- **Java SDK version (if applicable):** n/a — reproduced with `aws-cli/2.34.37`
  (Python 3.14.6). Also observed from `aws-sdk-go-v2`.
- **How you're running Floci:** Docker, `/var/run/docker.sock` mounted, on macOS
  arm64 with Docker Desktop 29.5.3. Services in use: `lambda`, `stepfunctions`.

matheusandre1:
## Summary

 Fixes: #2544 
The behavior described in issue #2544 has already been fixed in the current `main` branch. The fix was introduced via PR #2165; I simply ran a regression test.

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutorLambdaInvokeResultTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutorLambdaInvokeResultTest.java
@@ -171,6 +171,33 @@ class AslExecutorLambdaInvokeResultTest {
     }
 
     @Test
+    void resultSelectorUnwrapsPayloadWhenFunctionNameIsABareName() throws Exception {
+        // Scenario (c) of issue #2544: the function is referenced by its bare name and the
+        // documented "$.Payload" selector must reach the function output, not resolve to null.
+        Execution execution = run("""
+                {
+                  "StartAt": "T",
+                  "States": {
+                    "T": {
+                      "Type": "Task",
+                      "Resource": "arn:aws:states:::lambda:invoke",
+                      "Parameters": {"FunctionName": "%s", "Payload.$": "$"},
+                      "ResultSelector": {"unwrapped.$": "$.Payload"},
+                      "End": true
+                    }
+                  }
+                }
+                """.formatted(FUNCTION_NAME));
+
+        assertEquals("SUCCEEDED", execution.getStatus());
+        JsonNode output = objectMapper.readTree(execution.getOutput());
+        assertTrue(output.path("unwrapped").isObject(), "$.Payload must resolve to the function output");
+        assertEquals("RET", output.path("unwrapped").path("marker").asText());
+        assertEquals(1, output.path("unwrapped").path("echo").path("in").asInt());
+        assertEquals(1, output.size());
+    }
+
+    @Test
     void outputPathPayloadUnwrapsTheOptimizedInvokeResult() throws Exception {
         Execution execution = run("""
                 {


pgermosen (approved):
Confirmed #2165 already fixed this back in August, and the new test is a real assertion (checks the actual unwrapped payload contents, not just that the execution succeeds), so it'll catch a real regression if this ever breaks again. Good to close the coverage gap. Approving.

PR #3223, opened 2026-09-08
refactor(cloudformation): move the EC2 networking types into Ec2NetworkCfnProvisioner

hectorvent:
## Summary

Batch 4 of the `CloudFormationResourceProvisioner` split: `AWS::EC2::Subnet`, `InternetGateway`, `RouteTable`, `Route`, `NatGateway`, `EIP` and `SubnetRouteTableAssociation` move into a new `Ec2NetworkCfnProvisioner`. The move is the first commit and preserves behaviour; the following commits fix what reading the moved code against the registry schemas and `Ec2Service` turned up, one defect each:

- none of the seven was deleted on `DeleteStack` (the switch had no delete arm for them); each is now deleted through its own API, tolerating only its not-found code, and a Route's physical id becomes `<RouteTableId>|<destination>` (the registry primary identifier, AWS's Ref) so it can be
- every `UpdateStack` re-created all seven and re-pointed the stack at the copies; the prior entity is now kept when its create-only properties are unchanged, mutable ones applied in place, and a changed create-only property replaces it through the cleanup record
- `Tags` were dropped on the five taggable types; they are written on create and reconciled on update
- a NAT gateway was always public; `ConnectivityType` is honoured and its pairing with `AllocationId` validated
- missing required properties reached the service as nulls; they fail with `ValidationError` naming the property

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No wire surface changes. Attribute, createOnly and required sets checked against the seven registry schemas; the attributes Floci cannot produce are recorded in `getatt-attribute-gaps.tsv`. A Route's `Ref` changes from a generated `LogicalId-<random>` to `<RouteTableId>|<destination>`, which is what AWS returns.

## Checklist

- [x] `./mvnw test` passes locally (CloudFormation, provisioner and EC2 packages; the pre-existing ACM `Id` schema-gap row is the only failure, as on main)
- [x] New or updated integration test added (`CloudFormationEc2NetworkingIntegrationTest`; `Ec2NetworkCfnProvisionerTest`)


pgermosen (approved):
Good, thorough work on the rollback cases here, that's a hard area to get right. Checked the CIDR canonicalization fix and the replacement-cleanup identity fix directly, both hold up. Now that the full suite including all four shards and the native/compat matrix is green, approving.

PR #3224, opened 2026-09-08
feat(asl): add support for States.ArrayPartition, States.ArrayRange, …States.ArrayUnique, and States.MathRandom intrinsics with comprehensive tests

### LINKED ISSUE #2522: [FEAT] Step Functions (SFN): implement the remaining ASL intrinsic functions

matheusandre1:
## Summary

Closes: #2522 

What I found: of the nine special functions in the issue, five had already been implemented in the current main branch (Base64Encode, Base64Decode, Hash, StringSplit, ArrayGetItem), with tests in `AslExecutorIntrinsicFunctionsTest`. Only four remained.

What I implemented in `src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java`:

- `States.ArrayPartition(array, chunkSize)`: splits the array into chunks, with the last chunk containing the remainder. Non-integer sizes are rounded, as in AWS; non-positive sizes or a first argument that is not an array result in a `States.IntrinsicFailure`.
- `States.ArrayRange(start, end, step)`: inclusive when the step lands exactly on the end; accepts negative steps; rounds non-integer arguments. A step of zero causes failure, as does a result with more than 1,000 elements, matching AWS behavior. Counting is performed using `BigInteger` to avoid infinite loops with steps close to `Long.MAX_VALUE`.
- `States.ArrayUnique(array)`: preserves the first occurrence and the original order. Comparison is structural and compares numbers by value, because Jackson's equality check distinguishes between `IntNode` and `LongNode` (e.g., a literal in the expression becomes a `LongNode` while the input becomes an `IntNode`).
- `States.MathRandom(start, end[, seed])`: returns an integer in the closed interval [start, end], with rounding of non-integer boundaries. When a seed is provided, it uses `java.util.Random`—the generator used by AWS—so the same seed reproduces the same value.

All argument errors throw `States.IntrinsicFailure` (catchable via a `Catch` block), following the pattern of the five existing functions. A non-existent `$.` path continues to fail with `States.Runtime`.

Tests: I created `AslExecutorArrayAndMathIntrinsicsTest.java` with 44 test cases, including examples from the AWS documentation, exact reproduction of the end-to-end issue scenario, composition of intrinsic functions, and catchability within the full execution loop. 

One caveat: regarding MathRandom, the AWS documentation does not state whether the upper limit is inclusive. I adopted a closed interval, which matches the behavior of LocalStack as validated against AWS.

## Type of change

- [ ] Bug fix (`fix:`)
- [X] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -65,6 +65,7 @@ import jakarta.ws.rs.core.MultivaluedHashMap;
 import jakarta.ws.rs.core.MultivaluedMap;
 import org.jboss.logging.Logger;
 
+import java.math.BigInteger;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
@@ -82,6 +83,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Random;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Callable;
@@ -135,6 +137,8 @@ public class AslExecutor {
     // AWS caps the string input of States.Base64Encode/Base64Decode/Hash at 10,000 characters
     // (measured here in Unicode code points).
     private static final int INTRINSIC_MAX_INPUT_LENGTH = 10_000;
+    // AWS refuses a States.ArrayRange result of more than 1,000 elements.
+    private static final int ARRAY_RANGE_MAX_ELEMENTS = 1_000;
     // Must mirror the identical private set in JsonataEvaluator ($hash): both query languages
     // expose exactly these five algorithms, case-sensitively.
     private static final Set<String> HASH_ALGORITHMS =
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Good work overall, and glad the MathRandom bound got fixed - checked AWS's docs directly and it's explicit that the end is exclusive, so nextLong(start, end) with the strict start < end check is the right call. The BigInteger safeguard in ArrayRange and the structural equality handling in ArrayUnique are nice touches too. Approving.

PR #3228, opened 2026-09-08
fix(cloudformation): provision CloudFront policies and origin access controls

### LINKED ISSUE #2441: [BUG] CloudFormation Ref to AWS::CloudFront::ResponseHeadersPolicy resolves to a missing policy
## Service

CloudFormation / CloudFront

## AWS API Action

`CreateStack`, which creates an `AWS::CloudFront::ResponseHeadersPolicy` and an `AWS::CloudFront::Distribution`

## Expected behavior

CloudFormation resolves the distribution's `ResponseHeadersPolicyId` reference to the ID of the response headers policy created in the same stack. The policy and distribution reach `CREATE_COMPLETE`.

## Actual behavior

The response headers policy reaches `CREATE_COMPLETE`. The distribution then fails immediately, and the stack rolls back:

```text
AWS::CloudFront::ResponseHeadersPolicy CREATE_COMPLETE
AWS::CloudFront::Distribution CREATE_IN_PROGRESS
AWS::CloudFront::Distribution CREATE_FAILED The specified response headers policy does not exist.
AWS::CloudFormation::Stack ROLLBACK_COMPLETE
```

Deleting and recreating the stack produces the same result.

## Reproduction

```bash
cat > /tmp/floci-cloudfront-response-policy.yaml <<'YAML'
AWSTemplateFormatVersion: "2010-09-09"
Resources:
  ResponseHeadersPolicy:
    Type: AWS::CloudFront::ResponseHeadersPolicy
    Properties:
      ResponseHeadersPolicyConfig:
        Name: floci-response-policy-ref-repro
        Comment: Minimal response headers policy reference reproduction
        CustomHeadersConfig:
          Items:
            - Header: X-Floci-Repro
              Value: enabled
              Override: true

  Distribution:
    Type: AWS::CloudFront::Distribution
    Properties:
      DistributionConfig:
        Enabled: true
        Origins:
          - Id: origin
            DomainName: example-bucket.s3.amazonaws.com
            S3OriginConfig: {}
        DefaultCacheBehavior:
          TargetOriginId: origin
          ViewerProtocolPolicy: redirect-to-https
          ForwardedValues:
            QueryString: false
          ResponseHeadersPolicyId: !Ref ResponseHeadersPolicy
YAML

  --stack-name floci-response-policy-ref-repro \
  --template-body file:///tmp/floci-cloudfront-response-policy.yaml

aws --endpoint-url http://localhost:4566 cloudformation wait stack-create-complete \
  --stack-name floci-response-policy-ref-repro

aws --endpoint-url http://localhost:4566 cloudformation describe-stack-events \
  --stack-name floci-response-policy-ref-repro
```

## Environment

- Floci version / image tag: `floci/floci:nightly@sha256:033d872eb7b84de86cbdab3b75cd221bf9d80749bf1827ab27938476b200bbc6`
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker on Linux ARM64

hectorvent:
## Summary

`AWS::CloudFront::ResponseHeadersPolicy`, `CachePolicy`, `OriginRequestPolicy` and `OriginAccessControl` had no provisioner, so the dispatcher's stub arm reported them `CREATE_COMPLETE` with a synthetic id, and a distribution in the same stack that referenced one failed with "The specified response headers policy does not exist." (an origin's `OriginAccessControlId` failed the same way).

A new `CloudFrontCfnProvisioner` creates the four types through `CloudFrontService`:

- the physical id is the service id; `Fn::GetAtt` exposes `Id` and `LastModifiedTime` (`Id` only for the origin access control), the registry schemas' read-only properties
- updates are in place through the service's etag-guarded update, reading the current etag first as a client would; a prior id the service no longer knows is recreated
- deletes read the etag and tolerate only the type's own not-found code, so a policy still attached to a distribution fails the stack delete with `<Type>InUse`, as on AWS
- the response headers configuration is reshaped into the flattened form the service's codec stores through a new `ResponseHeadersPolicyConfigCodec.fromItemsTree` (list blocks unwrapped from `{Items: [...]}`, remove-header items reduced to names, scalars as text)

Known limit, stated in the class javadoc: a committed in-place update is not reverted when a later resource fails the stack update.

Closes #2441

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behaviour: `Ref` to any of the four types resolved to a stub id no CloudFront call accepted, so the stack rolled back at the distribution. Attribute names come from the CloudFormation registry schemas (`readOnlyProperties`: `Id`, `LastModifiedTime`; the origin access control declares only `Id`). No wire format changes: the CloudFront API surface is untouched apart from the new codec entry point used by the provisioner.

## Checklist

- [x] `./mvnw test` passes locally: CloudFormation package 156 classes, 1,472 tests, the only failure the pre-existing `AWS::CertificateManager::Certificate` `Id` schema-gap row on main; CloudFront package 197 tests green; `make docs-check` green once the regenerated table is committed
- [x] New or updated integration test added: `CloudFormationCloudFrontPoliciesIntegrationTest` (the issue's template verbatim in YAML, update keeps the id and bumps the etag, delete removes distribution and policy; a second stack with the other three types asserting the exact `Fn::GetAtt` keys), `CloudFrontCfnProvisionerTest` (15), two `fromItemsTree` cases in `ResponseHeadersPolicyConfigCodecTest`

Compat Docker suite not run: no SDK-facing wire change, CloudFormation provisioning only.

### DIFF
--- a/docs/services/cloudformation.md
+++ b/docs/services/cloudformation.md
@@ -99,7 +99,7 @@ cross-resource references.
 | Kinesis | `Stream` |
 | Kinesis Data Firehose | `DeliveryStream` |
 | IoT Core | `DomainConfiguration` (`ServerCertificates` resolves to a JSON string), `Policy` (deleted after detaching it from its principals; on AWS the delete fails with `DeleteConflictException` while the policy is attached), `Thing`, `TopicRule` |
-| CloudFront | `Distribution` |
+| CloudFront | `CachePolicy`, `Distribution`, `OriginAccessControl`, `OriginRequestPolicy`, `ResponseHeadersPolicy` |
 | CloudWatch | `Alarm` |
 | CloudWatch Logs | `LogGroup` |
 | WAFv2 | `WebACL` |
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/CloudFrontCfnProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/provisioners/CloudFrontCfnProvisioner.java
@@ -1,0 +1,259 @@
+package io.github.hectorvent.floci.services.cloudformation.provisioners;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.github.hectorvent.floci.core.common.AwsException;
+import io.github.hectorvent.floci.services.cloudformation.model.StackResource;
+import io.github.hectorvent.floci.services.cloudfront.CloudFrontService;
+import io.github.hectorvent.floci.services.cloudfront.ResponseHeadersPolicyConfigCodec;
+import io.github.hectorvent.floci.services.cloudfront.model.CachePolicy;
+import io.github.hectorvent.floci.services.cloudfront.model.OriginAccessControl;
+import io.github.hectorvent.floci.services.cloudfront.model.OriginRequestPolicy;
+import io.github.hectorvent.floci.services.cloudfront.model.ResponseHeadersPolicy;
+import jakarta.enterprise.context.ApplicationScoped;
+import org.jboss.logging.Logger;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+/**
+ * Provisions the CloudFront configuration types a distribution references by id: response headers
+ * policies, cache policies, origin request policies and origin access controls.
+ *
+ * <p>Each type has a single required {@code <Type>Config} property and no create-only property, so
+ * the physical id is the id the service assigns and every update is applied in place through the
+ * service's etag-guarded update, reading the current etag first as a client would. {@code Fn::GetAtt}
+ * exposes {@code Id} and {@code LastModifiedTime} (origin access controls only {@code Id}), the
+ * registry schema's read-only properties. Before this provisioner these types fell through to the
+ * dispatcher's stub arm, whose synthetic id a distribution then refused (issue #2441).
+ *
+ * <p>A policy configuration is handed to the service as nested maps and lists with every scalar as
+ * text, since the service's validator and codec read the blocks as string maps; the response headers
+ * config is additionally reshaped by {@link ResponseHeadersPolicyConfigCodec#fromItemsTree} into the
+ * flattened form that codec stores. {@code Name} and {@code Comment} are model fields, not part of
+ * the map.
+ *
+ * <p>A committed in-place update is not reverted when a later resource fails the update; the stack
+ * keeps the new policy configuration, the same limit the Cognito user pool has.
+ */
+@ApplicationScoped
+public class CloudFrontCfnProvisioner implements CfnResourceProvisioner {
+
+    private static final Logger LOG = Logger.getLogger(CloudFrontCfnProvisioner.class);
+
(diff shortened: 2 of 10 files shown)


hectorvent:
Thanks, two of the three land and are fixed; the first does not survive a look at the rollback path.

**Recreated objects leak (P1).** I do not think this one holds. `rollbackCreatedResources` deletes any resource with a non-null physical id that reached `CREATE_COMPLETE`, and `ROLLBACK_OWNED_ATTR` is only consulted for `CREATE_FAILED` resources; `rollbackUpdatedResources` likewise deletes a newly added resource when the physical id is set. This provisioner assigns the physical id immediately after each create with nothing that can throw in between, so the window the marker exists to cover never opens, and on the stale-prior path the previous object is already gone by definition. There is a real limitation nearby, though it is a different one: `rollbackUpdate` is not implemented, so a committed in-place update is not reverted when a later resource fails and the stack reports UPDATE_ROLLBACK_FAILED. That is pre-existing and shared with `AWS::Cognito::UserPool`, and it is noted in the PR description.

**Attached policies can disappear (P1).** Correct, and fixed in b229a855c. botocore declares `CachePolicyInUse` and `OriginRequestPolicyInUse` on those two deletes, both 409 sender faults, while Floci only checked response headers policies and origin access controls. Both now raise the modeled error, and the three near-identical scans are unified. Worth noting the provisioner was already right: it tolerates only `NoSuch<Type>`, so it fails the stack properly now the service raises the error.

**XML parsed with regex (P2).** Correct, and fixed in bf647b822. It now uses `XmlParser.extractPairs`, which seventeen sibling CloudFormation tests already wrap in the same one-line helper.

hectorvent:
Correction to my earlier reply: on the recreation finding you were right and I was not, so it is now fixed in eb595048d.

Where I went wrong. I argued from `rollbackCreatedResources`, which does delete any resource holding a physical id, and generalised it to the update path. It does not carry there. `rollbackUpdatedResources` only calls `deleteResourcePhysically` inside its `previous == null` branch, so a resource that already existed and merely acquires a new physical id mid-update never reaches a delete; it falls through to the "Rollback is not implemented" arm. That is exactly the stale-prior recreation you described, and the replacement did stay behind.

What stands from that reply: `ROLLBACK_OWNED_ATTR` really is consulted only for `CREATE_FAILED` resources, and a genuinely new resource really is cleaned up by its physical id. Both are just beside the point for this path. One nuance on the finding itself: "orphaned" overstates it slightly, since the stack still records the new id, so a later DeleteStack removes it. The rollback gap was real regardless.

The fix wires `ReplacementCleanup`, which already models this and which `CognitoCfnProvisioner` uses for the same situation, so a rollback now restores the prior id and attributes and deletes the object the failed update created. One limit stays and is deliberate: an in-place update keeps its physical id, records nothing to undo, and still reports as not rolled back, because reverting it needs a configuration snapshot this provisioner does not keep. A test pins that so it stays a decision rather than drifting.

pgermosen (approved):
Really solid one. I pulled the actual CloudFront docs to check the GetAtt attributes, the in-use error codes, and the response-headers config reshaping, and all of it matches AWS verbatim, down to the error message text.

The part I spent the most time on was the recreate-rollback fix, since that's the one you and Greptile went back and forth on. Your correction holds up: rollbackUpdatedResources really does only clean up through its previous == null branch, so a resource that keeps its identity but gets a new physical id mid-update was falling through untouched. Wiring it to ReplacementCleanup the same way Cognito does is the right call.

Good to merge.

PR #3229, opened 2026-09-08
fix(test): tear down the organization DetectiveIntegrationTest creates

### LINKED ISSUE #3226: [BUG] OrganizationsIntegrationTest fails when sharded with DetectiveIntegrationTest
## Service

Organizations (test isolation with Detective)

## AWS API Action

N/A — integration test isolation bug, not a wire-protocol issue. Surfaces as `CreateOrganization` failing with `AlreadyInOrganizationException` when it shouldn't.

## Expected behavior

`OrganizationsIntegrationTest` should pass regardless of which other test classes land in the same Maven Surefire shard/JVM fork. Its `@Order(1) createOrganization` test expects account `000000000000` to have no existing organization when it runs.

## Actual behavior

`DetectiveIntegrationTest` (`src/test/java/io/github/hectorvent/floci/services/detective/DetectiveIntegrationTest.java`) calls `organizationsService.createOrganization("000000000000", "ALL")` directly (bypassing HTTP, via `@Inject OrganizationsService`) in its first `@Test`, and the class has no `@AfterEach`/`@AfterAll` to delete it afterward. `OrganizationsService` treats an organization as a single global resource per account, and `@QuarkusTest` classes in the same shard share one JVM/application instance, so once `DetectiveIntegrationTest` runs, account `000000000000` permanently carries an organization for the rest of that JVM's life.

When `OrganizationsIntegrationTest` later runs in the same shard, its first test (`createOrganization`, `@Order(1)`) gets HTTP 400 `AlreadyInOrganizationException` instead of a fresh `Organization`, and every subsequent ordered test in that class cascades to failure because `organizationId`/`rootId`/etc. were never set. Reproduced on CI shard 3 of a 4-shard split: `Tests run: 50, Failures: 29` in `OrganizationsIntegrationTest`, on two separate reruns with the identical failure count, confirming this is deterministic given that shard's class grouping rather than random timing flakiness.

Confirmed from the CI log (`Build and Test (shard 3)`):
```
[INFO] Running io.github.hectorvent.floci.services.controltower.ControlTowerControllerIntegrationTest
...
2026-09-08 02:20:31,922 INFO  [...OrganizationsService] Created organization o-vn8fhzmext with root r-1yzy
2026-09-08 02:20:32,103 INFO  [...OrganizationsService] Created organization o-cjbktiyuli with root r-x7qe
2026-09-08 02:20:32,201 INFO  [...OrganizationsService] Created organization o-az8ujt43ia with root r-07b1
...
[INFO] Running io.github.hectorvent.floci.services.organizations.OrganizationsIntegrationTest
2026-09-08 02:20:57,860 ... organizations action: CreateOrganization
HTTP/1.1 400 Bad Request
x-amzn-query-error: AlreadyInOrganizationException;Sender
```
This is the same class of bug as #3156 (a test that creates an organization on a shared account without tearing it down, breaking any other Organizations suite that later shares its JVM fork), just a different leaking test and a different account (`000000000000`, the default, rather than a dedicated one).

## Reproduction

Run `OrganizationsIntegrationTest` and `DetectiveIntegrationTest` together in the same JVM fork, `DetectiveIntegrationTest` first:

```bash
./mvnw test -Dtest=DetectiveIntegrationTest,OrganizationsIntegrationTest
```

`OrganizationsIntegrationTest.createOrganization` fails with `AlreadyInOrganizationException` because `DetectiveIntegrationTest` already created an organization on account `000000000000` and never deleted it.

## Environment

- Floci version / image tag: `main` (reproduced at `65db87d6`)
- How you're running Floci (Docker / native / `mvn quarkus:dev`): `./mvnw test`, CI's default sharded Surefire run

## Related work

Same root-cause pattern as #3156 (closed): give `DetectiveIntegrationTest` its own dedicated account id instead of the default `000000000000`, following the per-test-account convention already used elsewhere (e.g. `ControlTowerControllerIntegrationTest`'s `000000000201`, `000000000204`, `000000000207`), or add cleanup (`deleteOrganization`) after the test. Surfaced while investigating a CI failure on #3215, which does not touch Organizations or Detective code and is not the cause.

7487:
## Summary

`DetectiveIntegrationTest.organizationGraphAndMemberLifecycleMatchesAwsContract()` calls `organizationsService.createOrganization("000000000000", "ALL")` through the injected service and the class has no teardown, so the organization outlives the test. Organizations state is shared by every `@QuarkusTest` class in the same Surefire fork, so once Detective runs, `OrganizationsIntegrationTest`'s `@Order(1) createOrganization` on the same default account gets `AlreadyInOrganizationException` and the 28 ordered tests after it cascade (`Tests run: 50, Failures: 29` on CI shard 3).

Fix: delete the organization in `@AfterEach`, mirroring the teardown #3158 added for the StackSets leak. The teardown is guarded by a flag set right after `createOrganization`, so it only runs for the test that created one — `-Dtest=DetectiveIntegrationTest#invalidAdminAccountIdReturnsValidationError` keeps working instead of failing with `AWSOrganizationsNotInUseException`. The organization has no member accounts or OUs at that point (Detective members are not Organizations accounts), so `deleteOrganization` succeeds directly.

Fixes #3226

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

No production behavior change; test isolation only.

## Checklist

- [x] `./mvnw test` passes locally — verified with the reproduction from the issue:
  - before: `./mvnw test -Dtest=DetectiveIntegrationTest,OrganizationsIntegrationTest` → `Tests run: 52, Failures: 29`, `OrganizationsIntegrationTest.createOrganization` failing with `AlreadyInOrganizationException`
  - after: same command passes (Detective's `Created organization` is followed by `Deleted organization` before Organizations creates its own); `-Dtest=DetectiveIntegrationTest#invalidAdminAccountIdReturnsValidationError` also passes
- [x] New or updated integration test added — the existing `DetectiveIntegrationTest` is updated; no new test since the fix is the teardown itself

### DIFF
--- a/src/test/java/io/github/hectorvent/floci/services/detective/DetectiveIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/detective/DetectiveIntegrationTest.java
@@ -4,6 +4,7 @@ import io.github.hectorvent.floci.services.organizations.OrganizationsService;
 import io.github.hectorvent.floci.testing.RestAssuredJsonUtils;
 import io.quarkus.test.junit.QuarkusTest;
 import jakarta.inject.Inject;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
@@ -19,14 +20,30 @@ class DetectiveIntegrationTest {
     @Inject
     OrganizationsService organizationsService;
 
+    private boolean createdOrganization;
+
     @BeforeAll
     static void configureRestAssured() {
         RestAssuredJsonUtils.configureAwsContentTypes();
     }
 
+    /**
+     * Organizations state is shared by every test in the JVM and keyed by the management account,
+     * so an organization left behind here makes the next class that creates one under
+     * {@code 000000000000} (e.g. {@code OrganizationsIntegrationTest}) fail with
+     * AlreadyInOrganizationException.
+     */
+    @AfterEach
+    void deleteTheOrganizationThisTestCreated() {
+        if (createdOrganization) {
+            organizationsService.deleteOrganization("000000000000");
+        }
+    }
+
     @Test
     void organizationGraphAndMemberLifecycleMatchesAwsContract() {
         organizationsService.createOrganization("000000000000", "ALL");
+        createdOrganization = true;
         String enableBody = post("/orgs/enableAdminAccount", "{\"AccountId\":\"000000000000\"}")
                 .statusCode(200).extract().asString();
         assertTrue(enableBody.isEmpty());


7487:
The shard 1 failure is unrelated to this change.

This PR touches one file, `src/test/java/io/github/hectorvent/floci/services/detective/DetectiveIntegrationTest.java`, and that class is not in shard 1's include list: `DetectiveIntegrationTest` does not appear anywhere in the shard 1 log, and `partition.sh` puts each class in exactly one shard (the CI summary job verifies the split is disjoint and complete). Shards 2, 3 and 4, one of which does run it, are green, including the `OrganizationsIntegrationTest` cascade this PR is about.

What failed is a startup race in the RDS backend container, at `RdsAwsIntegrationTest.java:66` (the `CreateDBSnapshot` step):

```
05:23:24,687 Initialized PostgreSQL IAM role in RDS container floci-rds-db-...
05:23:24,775 [rds:...:db:test-db-7ofgxosk] waiting for server to shut down...received fast shutdown request
05:23:24,828 RDS action: CreateDBSnapshot
05:23:24,862 [rds:...:db:test-db-7ofgxosk] FATAL:  the database system is shutting down
             -> 400 InvalidDBInstanceState, "Failed to create snapshot: Failed to create postgres snapshot"
05:23:24,991 [rds:...:db:test-db-7ofgxosk] LOG:  listening on IPv4 address "0.0.0.0", port 5432
```

The readiness gate is `initializePostgresIamRole` -> `execUntilSuccess`, which runs `psql -U <master> -d postgres` over the container's Unix socket. The postgres entrypoint's temporary init server listens on the socket only (the log shows `listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"` with no TCP line at that point), so the probe can succeed during initdb, before the entrypoint stops that server and execs the real one. `createPostgresSnapshot` then runs `pg_dumpall` once with no retry, unlike `restorePostgresSnapshot`, which retries while psql exits 2, so it fails hard inside that window.

The `Unable to negotiate PostgreSQL SSL` trace in the same log is separate noise, not a failure: it is thrown on a leftover virtual thread in `PostgresProtocolHandlerTest` after the client socket closes (`Caused by: java.net.SocketException: Socket closed`), and that class reports `Tests run: 24, Failures: 0`.

I do not have re-run rights on this repo. Could a maintainer re-run shard 1? I am happy to open a separate PR for the readiness race (gate on TCP reachability, for example `pg_isready -h 127.0.0.1`, since the temporary init server never listens on TCP) if that would be useful.

pgermosen (approved):
Correct fix for #3226, and the flag guard is exactly right, checked that the other test in this class never creates an org so an unconditional teardown would have broken it. Confirmed the org has nothing but the management account attached when this runs, since the test only touches Detective's own member registry, never real Organizations accounts. Approving.

PR #3237, opened 2026-09-08
fix(iam): use AWS-managed policy documents

nkootstra:
## Summary

Replace the unrestricted fallback used for AWS-managed IAM policies with the real policy documents for the policies already present in Floci's catalog.

Previously, every bundled managed policy could evaluate as `Allow` for all actions and resources. This could grant excessive permissions and make permissions boundaries ineffective in local tests. Managed policies now use their AWS-compatible statements, while unavailable policy data fails closed through the existing `NoSuchEntity` behavior.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The managed-policy catalog continues to preserve existing policy names, paths, ARNs, descriptions, default version IDs, and API response shapes.

Policy documents are loaded from the bundled AWS managed-policy dataset. The existing catalog contains 1,566 policies, all exposed with the existing `v1` version. Policies without an available document are omitted from the resolvable catalog instead of receiving an unrestricted fallback.

Added coverage verifies:

- `AmazonS3ReadOnlyAccess` allows reads but denies writes.
- Service-scoped policies deny actions outside their scope.
- `AdministratorAccess` remains broad where AWS defines it as broad.
- A managed policy used as a permissions boundary constrains a broader identity policy.
- Policy retrieval preserves expected fields.
- Unavailable policy versions return `NoSuchEntity`.

## Checklist

Focused IAM tests and `git diff --check` passed before commit. The full Maven suite was not run; a later focused rerun was stopped during Quarkus startup without a reported test failure.


nkootstra:
Addressed the Greptile finding in commit `8687641a`: added `iam/*.json` to `quarkus.native.resources.includes` and updated the resource-loading comment, so `managed-policy-documents.json` is embedded in native images.

pgermosen (approved):
Went and actually checked the bundled documents against AWS's live managed-policy pages for AmazonS3ReadOnlyAccess and AdministratorAccess, both match (the Action-as-array vs bare-string difference is a harmless normalization from the source dataset, semantically identical). Fails-closed behavior, native image registration, and the test coverage all hold up too. Approving.

PR #3242, opened 2026-09-08
feat(athena): support ListTagsForResource for workgroups

### LINKED ISSUE #2791: athena: WorkGroup ListTagsForResource is not supported
Real AWS's `terraform-provider-aws` calls `ListTagsForResource` when reading back an `aws_athena_workgroup` (as part of its default refresh cycle, even when no `tags` are set on the resource). Floci's Athena handler doesn't support this action:

```
Error: listing tags for Athena WorkGroup ...: InvalidAction: Action ListTagsForResource is not supported
```

Confirmed still missing on `main` — no `ListTagsForResource` handling under `services/athena`.

Blocks the Terraform/OpenTofu compatibility suites from including `aws_athena_workgroup` (removed from both in #1834 to unblock the suite; should be re-added once this lands).

Origin: https://github.com/floci-io/floci/pull/1834#issuecomment (abanna/hectorvent review, 2026-08-05 / 2026-08-15).

Mide69:
## Summary

Fixes #2791. `terraform-provider-aws`'s read/refresh for `aws_athena_workgroup` calls `ListTagsForResource` on every refresh cycle, even when no `tags` are set on the resource. Floci's Athena handler didn't support the action:

```
Error: listing tags for Athena WorkGroup ...: InvalidAction: Action ListTagsForResource is not supported
```

Parses the workgroup name out of the `ResourceARN` (using the existing `AwsArnUtils` helper already used elsewhere in the codebase) and returns the tags already captured on `CreateWorkGroup` - `WorkGroup`/`WorkGroupTag` already had everything needed, this was purely a missing action.

The primary workgroup always answers with an empty list, matching a live account: `CreateWorkGroup` already rejects `"primary"`, so it can never carry real tags. An `ARN` naming an unsupported resource type, or a workgroup that doesn't exist, faults with `InvalidRequestException`, matching the rest of this handler's not-found cases (e.g. `GetWorkGroup`).

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Adds `ListTagsForResource` support for Athena workgroup ARNs (`arn:aws:athena:<region>:<account>:workgroup/<name>`). Unblocks `aws_athena_workgroup` in the Terraform/OpenTofu compatibility suites, which needed this to complete its read/refresh cycle.

## Checklist

- [x] `./mvnw test` passes locally (full `athena` package: 51 tests, 0 failures - the 1 error is the pre-existing, unrelated `SocketTimeoutException` flake documented on other PRs in this repo)
- [x] New integration test added (`AthenaListTagsForResourceIntegrationTest`, 6 cases)

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AthenaTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AthenaTest.java
@@ -326,4 +326,30 @@ class AthenaTest {
             }
         }
     }
+
+    /**
+     * github.com/floci-io/floci/issues/2791: terraform-provider-aws's read/refresh for
+     * aws_athena_workgroup calls ListTagsForResource unconditionally. Round-trips through the
+     * real AthenaClient (rather than raw JSON) to validate the request/response shapes.
+     */
+    @Test
+    @Order(10)
+    @DisplayName("listTagsForResource returns the tags set on workgroup creation")
+    void listTagsForResourceReturnsTagsSetOnCreation() {
+        String groupName = "sdk-tags-" + UUID.randomUUID().toString().replace("-", "").substring(0, 8);
+        athena.createWorkGroup(CreateWorkGroupRequest.builder()
+                .name(groupName)
+                .tags(software.amazon.awssdk.services.athena.model.Tag.builder()
+                        .key("env").value("prod").build())
+                .build());
+
+        ListTagsForResourceResponse response = athena.listTagsForResource(
+                ListTagsForResourceRequest.builder()
+                        .resourceARN("arn:aws:athena:us-east-1:000000000000:workgroup/" + groupName)
+                        .build());
+
+        assertThat(response.tags()).hasSize(1);
+        assertThat(response.tags().get(0).key()).isEqualTo("env");
+        assertThat(response.tags().get(0).value()).isEqualTo("prod");
+    }
 }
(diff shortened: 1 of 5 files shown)


pgermosen:
Good fix for a real Terraform gap. One thing worth tightening: listTagsForResource only checks that the ARN's resource path starts with workgroup/, it never checks the service. Since Athena workgroups aren't account-partitioned in floci, this isn't a cross-tenant leak, but it does mean any ARN with a coincidentally matching region and workgroup name, even one for a completely different service, resolves successfully instead of getting rejected the way real AWS would reject a ResourceARN naming the wrong service. Worth adding an explicit check on arn.service().

Also, since this is meant to unblock the Terraform read/refresh cycle specifically, it'd be good to add an SDK-client test alongside the RestAssured one, matching how the other management API PRs in this repo validate request/response shapes.

Mide69:
Both addressed in 64e2193d.

1. Added the explicit arn.service() == "athena" check. Verified it's a real gap, not just theoretical: disabled the check, confirmed the new test (an ARN naming a different service but the same region/workgroup name, against a workgroup that genuinely exists) failed with the exact expected 400-vs-200 mismatch, then restored it.
2. Added an SDK-client test in the compat suite (AthenaTest.listTagsForResourceReturnsTagsSetOnCreation) round-tripping through the real AthenaClient - create a tagged workgroup, list its tags back, assert the SDK unmarshals them correctly. Ran it against a live local instance.

pgermosen (approved):
Fixed correctly, and the test proves it rather than just asserting it: creating a real workgroup first, then hitting it with an ARN naming a different service but the same name and region, is exactly the case an unguarded resource-path check would have let through. Good that you verified it by disabling the check and watching the test actually fail before restoring it, that's the right level of confidence for a test like this. Appreciate the SDK round-trip test too.

Good to merge.

PR #3243, opened 2026-09-08
fix(ec2): nest cidrBlockState in AssociateVpcCidrBlock response

### LINKED ISSUE #3241: [BUG] AssociateVpcCidrBlock response has wrong cidrBlockState shape, on secondary_cidr_blocks
## Service

EC2 (VPC)

## AWS API Action

AssociateVpcCidrBlock

## Expected behavior

When calling `AssociateVpcCidrBlock`, the real AWS API returns the new CIDR association with a **nested** `CidrBlockState` structure, e.g.:

```xml
<AssociateVpcCidrBlockResponse xmlns="http://ec2.amazonaws.com/doc/2016-11-15/">
  <requestId>...</requestId>
  <vpcId>vpc-xxxxxxxx</vpcId>
  <cidrBlockAssociation>
    <associationId>vpc-cidr-assoc-xxxxxxxx</associationId>
    <cidrBlock>10.250.44.0/22</cidrBlock>
    <cidrBlockState>
      <state>associating</state>
    </cidrBlockState>
  </cidrBlockAssociation>
</AssociateVpcCidrBlockResponse>
```

This is the same nested shape AWS uses everywhere else it reports a CIDR block association state (e.g. inside `DescribeVpcs`'s `cidrBlockAssociationSet`). The AWS Go SDK v2 (used by `terraform-provider-aws`) deserializes `CidrBlockState` as a `types.VpcCidrBlockState{ State: ..., StatusMessage: ... }` struct, and the provider's create-waiter for `aws_vpc_ipv4_cidr_block_association` polls `DescribeVpcs` (filtering on `cidr-block-association.association-id`) until this state reaches `associated`.

## Actual behavior

Floci's `AssociateVpcCidrBlock` response serializes `cidrBlockState` as a **flat string** instead of the nested struct:

```xml
<AssociateVpcCidrBlockResponse xmlns="http://ec2.amazonaws.com/doc/2016-11-15/">
  <requestId>818cc694-eaf7-4700-90db-3fc380afad6c</requestId>
  <vpcId>vpc-4e8cf7ca</vpcId>
  <cidrBlockAssociation>
    <associationId>vpc-cidr-assoc-b0689d12</associationId>
    <cidrBlock>10.250.44.0/22</cidrBlock>
    <cidrBlockState>associated</cidrBlockState>
  </cidrBlockAssociation>
</AssociateVpcCidrBlockResponse>
```

Note this is inconsistent with Floci's own `DescribeVpcs` response for the *same* VPC/association, which correctly nests the state:

```xml
<cidrBlockAssociationSet>
  <item>
    <associationId>vpc-cidr-assoc-b0689d12</associationId>
    <cidrBlock>10.250.44.0/22</cidrBlock>
    <cidrBlockState>
      <state>associated</state>
    </cidrBlockState>
  </item>
</cidrBlockAssociationSet>
```

Because of this malformed initial response, `terraform-provider-aws`'s creation waiter for `aws_vpc_ipv4_cidr_block_association` never observes a valid state transition and polls `DescribeVpcs` in a loop roughly every 5-10s, retrying up to 25 attempts (`amz_sdk_request="attempt=1; max=25"`), effectively hanging `terraform apply` for minutes even though:

- the association is created correctly in Floci's internal state on the first call (confirmed via subsequent `DescribeVpcs` calls, which show `cidrBlockState><state>associated</state>` immediately),
- and a raw unsigned `curl` call to `AssociateVpcCidrBlock` also returns instantly (200 OK) with the malformed body.

This reproduces consistently from a clean state (no pre-existing VPC/CIDR associations), not just on retries.

Eventually `terraform apply` does not converge on its own within the default operation and has to be interrupted; Terraform then reports:

```
Error: execution halted
Error: Plugin did not respond

  with module.vpc.aws_vpc_ipv4_cidr_block_association.this[0],
  on .terraform/modules/vpc/main.tf line 55, in resource "aws_vpc_ipv4_cidr_block_association" "this":
  55: resource "aws_vpc_ipv4_cidr_block_association" "this" {

The plugin encountered an error, and failed to respond to the
plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain
more details.
```

## Reproduction

```bash
# Minimal Terraform snippet that triggers the issue
cat > vpc.tf <<'HCL'
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "~> 6.0"

  name                  = "test-vpc"
  cidr                  = "10.248.1.128/25"
  secondary_cidr_blocks = ["10.250.44.0/22"]

  manage_default_security_group = false
  manage_default_network_acl    = false
  manage_default_route_table    = false
  enable_nat_gateway             = false
}
HCL

eval $(floci env)
terraform init
terraform apply -auto-approve
# Hangs on: module.vpc.aws_vpc_ipv4_cidr_block_association.this[0]: Still creating...
```

Or, to see the malformed response directly without Terraform:

```bash
eval $(floci env)

VPC_ID=$(aws --endpoint-url "$AWS_ENDPOINT_URL" ec2 create-vpc \
  --cidr-block 10.248.1.128/25 --query 'Vpc.VpcId' --output text)

curl -s -X POST "$AWS_ENDPOINT_URL/" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  --data "Action=AssociateVpcCidrBlock&CidrBlock=10.250.44.0%2F22&Version=2016-11-15&VpcId=$VPC_ID"

# Observe: <cidrBlockState>associated</cidrBlockState>
# Expected: <cidrBlockState><state>associated</state></cidrBlockState>
```

## Environment

- Floci version / image tag: 2.0.1
- Docker container (`floci start`), accessed via `http://localhost.floci.io:4566`

angelabad:
## Summary

`AssociateVpcCidrBlock` serialized `cidrBlockState` as a bare string
(`<cidrBlockState>associated</cidrBlockState>`) instead of the nested `<state>`
element the real AWS API uses — the same shape Floci's own `DescribeVpcs`
already returns correctly for the same association
(`<cidrBlockState><state>associated</state></cidrBlockState>`).

`terraform-provider-aws`'s AWS SDK Go v2 client deserializes `cidrBlockState`
as a `VpcCidrBlockState{ State, StatusMessage }` struct. With the flat string,
the create-waiter for `aws_vpc_ipv4_cidr_block_association` never observes a
valid state on the initial response and polls `DescribeVpcs` up to 25 times,
hanging `terraform apply` for minutes on any module/config using
`secondary_cidr_blocks` — e.g. `terraform-aws-modules/vpc/aws`.

`Ec2QueryHandler.handleAssociateVpcCidrBlock` built the response with
`.elem("cidrBlockState", ...)` (flat), while the sibling code path for
`DescribeVpcs` (`vpcXml`) and the IPv6 equivalent already used
`.start("cidrBlockState").elem("state", ...).end("cidrBlockState")` (nested).
This PR aligns `AssociateVpcCidrBlock` with the existing, correct pattern.

Closes #3241

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behaviour: the wire response for `AssociateVpcCidrBlock` nested
`cidrBlockState` as a bare string instead of AWS's `<cidrBlockState><state>...`
element, which `aws-sdk-go-v2` (and therefore `terraform-provider-aws`) can't
deserialize into a valid state — the create-waiter for
`aws_vpc_ipv4_cidr_block_association` polls up to 25 times before giving up.
No other wire changes: `DescribeVpcs` and the IPv6 equivalent already used the
correct nested shape.

## Checklist

- [x] `./mvnw test` passes locally: reverting just the handler fix makes the
  new `Ec2VpcCidrBlockAssociationIntegrationTest` fail (`BUILD FAILURE`, 1
  failure); with the fix restored, the new test and the existing
  `Ec2VpcIpv6CidrBlockIntegrationTest` suite both pass (4 tests, 0 failures)
- [x] New or updated integration test added:
  `Ec2VpcCidrBlockAssociationIntegrationTest` — creates a VPC, associates a
  secondary CIDR block, and asserts
  `cidrBlockAssociation.cidrBlockState.state == "associated"` on the
  `AssociateVpcCidrBlock` response itself (the regression check — it fails
  against the old flat-string shape), then confirms the same association is
  readable via `DescribeVpcs` with a matching state

Also reproduced end-to-end with `terraform-aws-modules/vpc/aws` +
`secondary_cidr_blocks`, confirming `terraform apply` no longer hangs on
`aws_vpc_ipv4_cidr_block_association`.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2QueryHandler.java
@@ -2632,7 +2632,7 @@ public class Ec2QueryHandler {
                 .start("cidrBlockAssociation")
                 .elem("associationId", assoc.getAssociationId())
                 .elem("cidrBlock", assoc.getCidrBlock())
-                .elem("cidrBlockState", assoc.getCidrBlockState())
+                .start("cidrBlockState").elem("state", assoc.getCidrBlockState()).end("cidrBlockState")
                 .end("cidrBlockAssociation")
                 .end("AssociateVpcCidrBlockResponse");
         return xmlResponse(xml.build());
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2VpcCidrBlockAssociationIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2VpcCidrBlockAssociationIntegrationTest.java
@@ -1,0 +1,65 @@
+package io.github.hectorvent.floci.services.ec2;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.equalTo;
+
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.junit.QuarkusTest;
+
+/**
+ * AssociateVpcCidrBlock returned cidrBlockState as a bare string
+ * ({@code <cidrBlockState>associated</cidrBlockState>}) instead of the nested shape the real API
+ * uses ({@code <cidrBlockState><state>associated</state></cidrBlockState>}), the same shape
+ * DescribeVpcs already returned for the identical association. The AWS SDK waiter that
+ * terraform-provider-aws uses for aws_vpc_ipv4_cidr_block_association parses the nested form, so
+ * it never saw a valid state and polled DescribeVpcs until it gave up, hanging `terraform apply`
+ * on any secondary_cidr_blocks entry.
+ */
+@QuarkusTest
+class Ec2VpcCidrBlockAssociationIntegrationTest {
+
+    private static final String AUTH_HEADER =
+            "AWS4-HMAC-SHA256 Credential=test/20260205/us-east-1/ec2/aws4_request";
+
+    @Test
+    void associateVpcCidrBlockReturnsNestedCidrBlockState() {
+        String vpcId = given()
+            .formParam("Action", "CreateVpc")
+            .formParam("CidrBlock", "10.94.0.0/24")
+            .header("Authorization", AUTH_HEADER)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .extract().path("CreateVpcResponse.vpc.vpcId");
+
+        String associationId = given()
+            .formParam("Action", "AssociateVpcCidrBlock")
+            .formParam("VpcId", vpcId)
+            .formParam("CidrBlock", "10.95.0.0/24")
+            .header("Authorization", AUTH_HEADER)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .body("AssociateVpcCidrBlockResponse.cidrBlockAssociation.cidrBlock", equalTo("10.95.0.0/24"))
+            // The regression: this must be the nested element, not a bare string value.
+            .body("AssociateVpcCidrBlockResponse.cidrBlockAssociation.cidrBlockState.state",
+                    equalTo("associated"))
+            .extract().path("AssociateVpcCidrBlockResponse.cidrBlockAssociation.associationId");
+
+        // Same association, read back through DescribeVpcs, must agree on both shape and state.
+        given()
+            .formParam("Action", "DescribeVpcs")
+            .formParam("VpcId.1", vpcId)
+            .header("Authorization", AUTH_HEADER)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .body("DescribeVpcsResponse.vpcSet.item.cidrBlockAssociationSet.item.find { "
+                    + "it.associationId == '" + associationId + "' }.cidrBlockState.state",
+                    equalTo("associated"));
+    }
+}


pgermosen (commented):
Checked this against the real AssociateVpcCidrBlock docs rather than taking the description's word for it, and the nested shape is exactly right. It's actually a tighter fix than described too: this same handler already nests the IPv6 sibling field correctly two lines below the line you changed, and DescribeVpcs's XML builder already nests this exact field, so this was a genuine one-off bug rather than a style choice between two existing patterns. Good test too, it asserts the shape on the AssociateVpcCidrBlock response itself rather than only through a follow-up DescribeVpcs read, which was already correct and wouldn't have caught this.

One thing needs fixing before this can merge: the body doesn't follow this repo's PR template (Summary/Type of change/AWS Compatibility/Checklist), it uses different headers and drops the checkboxes. Since this repo squash-merges, whatever's in the body becomes the permanent commit message on main, so this isn't just a formatting nitpick, could you update it to the template shape?

Unrelated aside while I was in this code: DisassociateVpcCidrBlock has a similar wire-shape gap (it returns a bare boolean instead of the association details AWS actually sends back), but that's a separate issue, not something to fold in here.

angelabad:
Hi @pgermosen, thanks for your review. I edited the pr body, I hope it's correct now.

pgermosen (approved):
Body reads clean now, thanks for updating it. Everything else checked out in the first pass, good to merge.

hectorvent (approved):
Confirmed against the EC2 model: `CidrBlockState` is a `VpcCidrBlockState` structure with `State` and `StatusMessage`, so the nested element is the right shape, and this now agrees with what `DescribeVpcs` already returned. Approving.

PR #3244, opened 2026-09-08
fix(cloudcontrol): scope resources by caller account

nkootstra:
## Summary

Fix Cloud Control account isolation for resource provisioning, asynchronous status polling, lookup, deletion, and restart recovery.

Cloud Control previously created backing resources under the default account (`000000000000`) regardless of the caller. Request tokens and create-time resource metadata were also not reliably account-scoped.

This change:

- Propagates the caller account from both JSON request boundaries.
- Runs asynchronous provisioning and deletion under the caller’s account context.
- Includes account ownership in Cloud Control resource keys and progress metadata.
- Rejects cross-account status polling, lookup, and deletion.
- Persists request and resource metadata through the existing account-aware storage system.
- Restores persisted Cloud Control state after restart and resumes persisted in-progress creates.
- Preserves default-account behavior for existing direct callers.
- Adds two-account unit and integration coverage.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Cloud Control request and response shapes are unchanged.

The previous implementation could provision a resource for a non-default caller under account `000000000000`. It could also allow request tokens and create-time metadata to cross account boundaries.

The implementation now preserves the caller account through provisioning, asynchronous status polling, resource lookup, deletion, persistence, and restart recovery. Cross-account operations fail without modifying the owning account’s resource.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/core/common/AwsJson11Controller.java
+++ b/src/main/java/io/github/hectorvent/floci/core/common/AwsJson11Controller.java
@@ -315,7 +315,8 @@ public class AwsJson11Controller {
                         action, request, region, regionResolver.getAccountId());
                 case "servicecatalog" -> serviceCatalogJsonHandler.handle(
                         action, request, region, regionResolver.getAccountId());
-                case "cloudcontrol" -> cloudControlJsonHandler.handle(action, request, region);
+                case "cloudcontrol" -> cloudControlJsonHandler.handle(
+                                        action, request, region, regionResolver.getAccountId());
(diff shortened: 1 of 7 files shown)


pgermosen (approved):
Solid work on the recovery correctness, especially the request-token correlation to avoid double-provisioning on restart, that required a real model change (adding requestToken to CreatedResource), not a quick patch. Checked the core account isolation too - deleteResource, requestStatus and getResource all key strictly off the caller's account and fail closed without confirming another account's resource exists. Approving.

PR #3257, opened 2026-09-08
fix(ec2): match cidr-block-association filters on DescribeVpcs

### LINKED ISSUE #3256: [BUG] DescribeVpcs ignores cidr-block-association.* / ipv6-cidr-block-association.* filters
## Service

EC2 (VPC)

## AWS API Action

DescribeVpcs

## Expected behavior

`DescribeVpcs` should apply the `cidr-block-association.*` and `ipv6-cidr-block-association.*`
filters documented for the action, matching a VPC only when one of its CIDR block associations
satisfies the filter:

- `cidr-block-association.association-id`
- `cidr-block-association.cidr-block`
- `cidr-block-association.state`
- `ipv6-cidr-block-association.association-id`
- `ipv6-cidr-block-association.ipv6-cidr-block`
- `ipv6-cidr-block-association.state`

`terraform-provider-aws`'s waiter for `aws_vpc_ipv4_cidr_block_association`
(`waitVPCCIDRBlockAssociationCreated` / `findVPCCIDRBlockAssociationByID`) polls exactly this way:
it calls `DescribeVpcs` with `Filter.1.Name=cidr-block-association.association-id` set to the new
association's ID, and asserts the response contains **exactly one** VPC before it looks at that
VPC's `CidrBlockState`.

## Actual behavior

Floci's `matchesFilter` has a `switch` over VPC filter names that only recognizes `vpc-id`,
`state`, `is-default`, and `cidr`/`cidr-block`. Every other filter name — including all six
`cidr-block-association.*` / `ipv6-cidr-block-association.*` filters above — falls through to
`default -> true`, so the filter is silently ignored and every VPC in the account is returned,
regardless of whether it owns the requested association.

With just the default VPC plus one VPC created by the test (the minimal realistic case — no
extra VPCs needed to reproduce this), filtering `DescribeVpcs` by the new VPC's own
`cidr-block-association.association-id` still returns **both** VPCs:

```
$ curl -s -X POST http://localhost:4566/ \
    --aws-sigv4 "aws:amz:us-east-1:ec2" --user "test:test" \
    -H "Content-Type: application/x-www-form-urlencoded" \
    --data-urlencode "Action=DescribeVpcs" \
    --data-urlencode "Version=2016-11-15" \
    --data-urlencode "Filter.1.Name=cidr-block-association.association-id" \
    --data-urlencode "Filter.1.Value.1=vpc-cidr-assoc-13df28ea"

<DescribeVpcsResponse xmlns="http://ec2.amazonaws.com/doc/2016-11-15/"><requestId>20cc9f53-d811-45f6-93d2-b757ed290c36</requestId><vpcSet><item><vpcId>vpc-default-us-east-1</vpcId><state>available</state><cidrBlock>172.31.0.0/16</cidrBlock>...<cidrBlockAssociationSet><item><associationId>vpc-cidr-assoc-default</associationId>...</item></cidrBlockAssociationSet><tagSet/></item><item><vpcId>vpc-dbe908fb</vpcId><state>available</state><cidrBlock>10.248.1.128/25</cidrBlock>...<cidrBlockAssociationSet><item><associationId>vpc-cidr-assoc-a47c57c8</associationId><cidrBlock>10.248.1.128/25</cidrBlock>...</item><item><associationId>vpc-cidr-assoc-13df28ea</associationId><cidrBlock>10.250.44.0/22</cidrBlock>...</item></cidrBlockAssociationSet><tagSet/></item></vpcSet></DescribeVpcsResponse>
```

Only `vpc-dbe908fb` owns `vpc-cidr-assoc-13df28ea`; `vpc-default-us-east-1` should not be in the
result at all.

Because `terraform-provider-aws`'s `findVPCCIDRBlockAssociationByID` asserts a single VPC comes
back (`tfresource.AssertSingleValueResult`), two VPCs in the response raises a
`TooManyResultsError`. That error type implements `As(**retry.NotFoundError)`, so the SDK
retry/waiter machinery treats "too many results" the same as "not found yet" instead of failing
fast — the `StateRefreshFunc` just reports "still pending" forever. The waiter then polls
`DescribeVpcs` every ~10s until its 10-minute create timeout expires, hanging `terraform apply`
on any `secondary_cidr_blocks` entry even though the association itself was created correctly on
the first call (confirmed via `AssociateVpcCidrBlock`'s own response and immediate follow-up
`DescribeVpcs` calls, both showing `cidrBlockState.state = associated` right away).

## Reproduction

```bash
export AWS_ACCESS_KEY_ID=test

VPC_ID=$(aws ec2 create-vpc --cidr-block 10.248.1.128/25 --query 'Vpc.VpcId' --output text)
ASSOC_ID=$(aws ec2 associate-vpc-cidr-block --vpc-id "$VPC_ID" \
  --cidr-block 10.250.44.0/22 --query 'CidrBlockAssociation.AssociationId' --output text)

aws ec2 describe-vpcs \
  --filters "Name=cidr-block-association.association-id,Values=$ASSOC_ID" \
  --query 'Vpcs[].VpcId' --output text
# Expected: only $VPC_ID
# Actual:   every VPC in the account, including the default VPC
```

Terraform reproduction (same module as #3241, since it's the consumer that surfaces this):

```hcl
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "~> 6.0"

  name                  = "test-vpc"
  cidr                  = "10.248.1.128/25"
  secondary_cidr_blocks = ["10.250.44.0/22"]

  manage_default_security_group = false
  manage_default_network_acl    = false
  manage_default_route_table    = false
  enable_nat_gateway             = false
}
```

```bash
eval $(floci env)
terraform init
terraform apply -auto-approve
# Hangs on module.vpc.aws_vpc_ipv4_cidr_block_association.this[0] until the 10-minute
# create timeout, even with #3243 applied.
```

## Related

This is a second, independent bug in the same `secondary_cidr_blocks` / `aws_vpc_ipv4_cidr_block_association`
path as #3241 / #3243. #3243 fixed the *shape* of `AssociateVpcCidrBlock`'s `cidrBlockState`
response; this issue is about `DescribeVpcs` ignoring the `cidr-block-association.*` filters that
the provider's waiter uses on every subsequent poll. Applying #3243 alone is not sufficient to
unblock `terraform apply` with `secondary_cidr_blocks` — this filter bug reproduces even with
#3243's fix in place.

## Environment

- Floci version / image tag: 2.0.1
- How you're running Floci: `mvn quarkus:dev` (also reproduces via `floci start` / Docker)

angelabad:
## Summary

`DescribeVpcs` ignores the `cidr-block-association.*` and `ipv6-cidr-block-association.*`
filters. `matchesFilter`'s VPC `switch` only recognized `vpc-id`, `state`, `is-default`, and
`cidr`/`cidr-block`; every other filter name fell through to `default -> true`, so the filter
was silently ignored and every VPC in the account matched.

This adds the six documented filters:

- `cidr-block-association.association-id`
- `cidr-block-association.cidr-block`
- `cidr-block-association.state`
- `ipv6-cidr-block-association.association-id`
- `ipv6-cidr-block-association.ipv6-cidr-block`
- `ipv6-cidr-block-association.state`

`terraform-provider-aws`'s waiter for `aws_vpc_ipv4_cidr_block_association`
(`waitVPCCIDRBlockAssociationCreated` / `findVPCCIDRBlockAssociationByID`) polls `DescribeVpcs`
with `Filter.1.Name=cidr-block-association.association-id` and asserts the response contains
exactly one VPC. With the filter ignored, any account with more than one VPC (e.g. the default
VPC plus the one being created) gets every VPC back, and the provider's
`tfresource.AssertSingleValueResult` raises a `TooManyResultsError`. That error type implements
`As(**retry.NotFoundError)`, so the SDK's retry machinery treats "too many results" the same as
"not found yet" instead of failing fast — the waiter just polls forever until its 10-minute
create timeout, hanging `terraform apply` on any `secondary_cidr_blocks` entry.

Fixes #3256, which has the full repro and evidence (raw XML showing the filter being ignored).

Related to #3241 / #3243: same `secondary_cidr_blocks` / `aws_vpc_ipv4_cidr_block_association`
path, but an independent bug. #3243 fixes the *shape* of `AssociateVpcCidrBlock`'s
`cidrBlockState`; this fixes `DescribeVpcs` ignoring the filter the waiter uses on every
subsequent poll. Applying #3243 alone does not unblock `terraform apply` with
`secondary_cidr_blocks` — this reproduces independently of it.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behaviour: `DescribeVpcs` accepted the six `cidr-block-association.*` /
`ipv6-cidr-block-association.*` filters syntactically but never evaluated them, matching every
VPC in the account regardless of the filter value. Filter names and semantics follow the
documented `DescribeVpcs` filter list (`association-id`, `cidr-block` / `ipv6-cidr-block`, and
`state` per association). No other wire changes.

## Checklist

- [x] `./mvnw test` passes locally: ran the full EC2 VPC-related suite (`Ec2IntegrationTest`,
  `Ec2ServiceTest`, `Ec2ServicePersistenceTest`, `CloudControlServiceTest`, plus the new and
  existing CIDR association tests) — all green, no regressions
- [x] New or updated integration test added: `Ec2VpcCidrBlockAssociationFilterIntegrationTest`,
  covering `cidr-block-association.association-id` and `cidr-block-association.cidr-block`
  against a multi-VPC account (asserts the filter narrows to the one VPC owning the
  association, not every VPC)

Verified locally against a real `terraform apply` with `terraform-aws-modules/vpc/aws ~> 6.0`
and `secondary_cidr_blocks`: previously hung until timeout, now completes in ~14s.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java
@@ -6144,6 +6144,18 @@ public class Ec2Service implements ContainerTeardown, ResourceProvider {
                 // 2026-08-25: it matches only the primary block, not a secondary
                 // cidr-block-association entry).
                 case "cidr", "cidr-block" -> matchesValue(values, vpc.getCidrBlock());
+                case "cidr-block-association.association-id" -> vpc.getCidrBlockAssociationSet().stream()
+                        .anyMatch(a -> matchesValue(values, a.getAssociationId()));
+                case "cidr-block-association.cidr-block" -> vpc.getCidrBlockAssociationSet().stream()
+                        .anyMatch(a -> matchesValue(values, a.getCidrBlock()));
+                case "cidr-block-association.state" -> vpc.getCidrBlockAssociationSet().stream()
+                        .anyMatch(a -> matchesValue(values, a.getCidrBlockState()));
+                case "ipv6-cidr-block-association.association-id" -> vpc.getIpv6CidrBlockAssociationSet().stream()
+                        .anyMatch(a -> matchesValue(values, a.getAssociationId()));
+                case "ipv6-cidr-block-association.ipv6-cidr-block" -> vpc.getIpv6CidrBlockAssociationSet().stream()
+                        .anyMatch(a -> matchesValue(values, a.getIpv6CidrBlock()));
+                case "ipv6-cidr-block-association.state" -> vpc.getIpv6CidrBlockAssociationSet().stream()
+                        .anyMatch(a -> matchesValue(values, a.getIpv6CidrBlockState()));
                 default -> true;
             };
         }
--- a/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2VpcCidrBlockAssociationFilterIntegrationTest.java
+++ b/src/test/java/io/github/hectorvent/floci/services/ec2/Ec2VpcCidrBlockAssociationFilterIntegrationTest.java
@@ -1,0 +1,285 @@
+package io.github.hectorvent.floci.services.ec2;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.equalTo;
+
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.junit.QuarkusTest;
+
+/**
+ * DescribeVpcs ignored the {@code cidr-block-association.*} and {@code ipv6-cidr-block-association.*}
+ * filters (unrecognized filter names fell through to the catch-all {@code default -> true} in
+ * {@code matchesFilter}, so every VPC in the account matched regardless of the requested
+ * association). terraform-provider-aws's waiter for aws_vpc_ipv4_cidr_block_association polls
+ * DescribeVpcs filtered by {@code cidr-block-association.association-id} and asserts exactly one
+ * VPC comes back; with more than one VPC in the account the assertion failed with a "too many
+ * results" error that the SDK treats as not-found-yet, so the waiter retried until it timed out,
+ * hanging `terraform apply` on any secondary_cidr_blocks entry.
+ */
+@QuarkusTest
+class Ec2VpcCidrBlockAssociationFilterIntegrationTest {
+
+    private static final String AUTH_HEADER =
+            "AWS4-HMAC-SHA256 Credential=test/20260908/us-east-1/ec2/aws4_request";
+
+    @Test
+    void describeVpcsFiltersByCidrBlockAssociationId() {
+        String vpcId = given()
+            .formParam("Action", "CreateVpc")
+            .formParam("CidrBlock", "10.96.0.0/24")
+            .header("Authorization", AUTH_HEADER)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200)
+            .extract().path("CreateVpcResponse.vpc.vpcId");
+
+        // A second, unrelated VPC must not match the first VPC's association filter.
+        given()
+            .formParam("Action", "CreateVpc")
+            .formParam("CidrBlock", "10.97.0.0/24")
+            .header("Authorization", AUTH_HEADER)
+        .when()
+            .post("/")
+        .then()
+            .statusCode(200);
+
+        String associationId = given()
+            .formParam("Action", "AssociateVpcCidrBlock")
+            .formParam("VpcId", vpcId)
+            .formParam("CidrBlock", "10.98.0.0/24")
+            .header("Authorization", AUTH_HEADER)
+        .when()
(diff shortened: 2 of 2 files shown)


hectorvent (approved):
Confirmed the six filter names against the `DescribeVpcs` documentation in the model, including `ipv6-cidr-block-association.ipv6-cidr-block`. Approving.

One for a separate change, since it is the shape of the original bug rather than yours: the VPC switch still ends in `default -> true`, so any other unmodeled filter name silently matches every VPC.

PR #3261, opened 2026-09-08
fix(memorydb): scope resources by account and region

nkootstra:
## Summary

- Scope MemoryDB clusters, users, and ACLs by region in addition to the existing account-aware storage partition.
- Persist account and region ownership and propagate it through lookups, mutations, ARNs, authentication, containers, proxies, and ACL relationships.
- Preserve legacy unscoped records for the configured default account and region by migrating them on access.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

MemoryDB resources were keyed primarily by name, allowing same-name clusters, users, and ACLs from different regions to collide or be returned through the wrong client. This fix scopes lookups and mutations by caller account and region, preserves account- and region-qualified ARNs, and keeps the AWS request and response shapes unchanged.

The MemoryDB AWS SDK integration tests and focused service, handler, container, and persistence tests pass. No separate SDK or CLI wire-protocol shape changes were introduced.

## Checklist

The full `./mvnw -q test` suite was started but exceeded the five-minute execution timeout during Quarkus startup-heavy tests without reporting a failure. The existing MemoryDB integration test passed as part of the focused MemoryDB test run; the new regression coverage is currently in `MemoryDbServiceTest`.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/memorydb/MemoryDbHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/memorydb/MemoryDbHandler.java
@@ -43,15 +43,15 @@ public class MemoryDbHandler {
         try {
             return switch (action) {
                 case "CreateCluster" -> handleCreateCluster(request, region);
-                case "DescribeClusters" -> handleDescribeClusters(request);
-                case "UpdateCluster" -> handleUpdateCluster(request);
-                case "DeleteCluster" -> handleDeleteCluster(request);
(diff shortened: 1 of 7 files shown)


hectorvent (approved):
Thank you, the second commit is the part that matters: migrating legacy records on scan as well as on get means an upgraded deployment still sees its existing clusters, and qualifying the container identity keeps delete pointed at the right container. Approving.

PR #3265, opened 2026-09-08
fix(lambda): scope extracted function code by region

nkootstra:
## Summary

Function ZIP extraction paths now include both the owning account and AWS region. This prevents same-name functions in different regions from overwriting or deleting each other’s extracted code, including published-version snapshots.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The previous canonical extraction layout omitted the region, so functions with the same name and account in different regions shared a directory. Extraction, update, publish-version copying, and deletion now use the request’s region while preserving existing legacy-path cleanup behavior.

Layer extraction and layer-ARN behavior are intentionally unchanged because PR #2948 is still active and covers that overlapping area.

## Checklist

Focused validation passed:

`./mvnw -B -Dtest=io.github.hectorvent.floci.services.lambda.zip.CodeStoreTest,io.github.hectorvent.floci.services.lambda.LambdaAccountScopedCodeTest test`

22 tests passed, including account/region isolation, update/delete isolation, restart persistence, and concurrent extraction.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -866,7 +866,7 @@ public class LambdaService implements ResourceProvider {
         synchronized (lockForConcurrencyOp(fn.getFunctionArn())) {
             warmPool.drainEnvironment(version.get());
             functionStore.deleteVersion(region, name, qualifier);
-            reclaimVersionCodeDirectory(fn, qualifier, version.get());
+            reclaimVersionCodeDirectory(region, fn, qualifier, version.get());
             // The snapshot may still share $LATEST's code directory, so this only reclaims once no
             // remaining version references it.
             reclaimLegacyCodeDirectoryIfUnused(name);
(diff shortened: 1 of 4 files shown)


nkootstra:
Addressed the Greptile finding in commit `1d7683df`. The concern is valid: the preceding account-scoped layout could make a legacy function named like a region the parent of the new region path. New canonical paths now use a reserved namespace: `<account>/@regions/<region>/<function>`. `@` cannot be emitted by `CodeStore.sanitizeName`, so no function path can overlap this namespace. Existing persisted account-scoped paths remain readable and are not deleted or migrated, avoiding unsafe guesses about ownership; they cannot contain or expose any newly extracted package. Added a regression test covering a legacy `us-east-1` function directory and the new same-region function path. Focused validation now passes 23 tests.

nkootstra:
Investigated the failed CI checks. The CI shard failure was unrelated to this PR: `RdsAwsIntegrationTest.createDescribeRestoreRoundTrip` returned HTTP 400, with an additional unrelated EOF in `PostgresProtocolHandlerTest`; no Lambda tests or changed files were involved. The initially reported arm64 compatibility check subsequently completed successfully. Repository admin permissions were required to rerun the failed run, so I triggered a fresh PR-synchronization run with the empty Conventional Commit `99eec0a` (`chore(ci): retry transient pull request checks`). The new CI and compatibility checks are now pending.

hectorvent (approved):
The `@regions` reserved segment is a better answer than plain nesting: since `sanitizeName` can never emit `@`, a function named after a region cannot become the parent of another function's path. Approving.

PR #3266, opened 2026-09-08
fix(ram): enforce resource share principal visibility

nkootstra:
## Summary

Prevent RAM read operations from exposing resource shares to unrelated accounts. Owners, invited account principals, and eligible organization or OU members retain visibility.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, `OTHER-ACCOUNTS` returned every non-owned share regardless of its principals or invitation state. RAM now applies the same principal visibility rule to `GetResourceShares`, `ListResources`, and `ListPrincipals`. Account principals require a pending or accepted invitation; rejected invitations and unrelated caller accounts are hidden. Existing Organizations membership and OU ancestry semantics are used for organization principals.

## Checklist

Focused validation passed: `RamServiceTest`, `RamIntegrationTest`, and `RamServicePersistenceTest` (47 tests total).

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ram/RamController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ram/RamController.java
@@ -63,7 +63,7 @@ public class RamController {
     @Consumes(MediaType.WILDCARD)
     public Response enableSharingWithAwsOrganization() {
         ObjectNode response = objectMapper.createObjectNode();
-        response.put("returnValue", service.enableSharingWithAwsOrganization());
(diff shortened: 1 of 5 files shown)


hectorvent (approved):
Both closed cleanly: the fallback now checks the caller is in the same organization as the owner rather than any organization, and matching on a live invitation rather than the first record means a reassociated principal stays visible. Approving.

PR #3267, opened 2026-09-08
fix(firehose): isolate streams by account and region

nkootstra:
## Summary

Isolate Firehose delivery-stream state by caller account, AWS region, and delivery-stream name. Identical stream names can now coexist without sharing persisted state, buffers, buffering timestamps, or Kinesis iterator checkpoints.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, Firehose state was keyed primarily by stream name, allowing records, flush state, and Kinesis checkpoints to cross account or region boundaries. The AWS request and response shapes and ARN format are unchanged; generated ARNs now consistently identify the owning account and region.

Direct EventBridge and SES Firehose delivery callers now route records using the account and region from the destination ARN.

## Checklist

Validation run locally:

- `./mvnw -q -Dtest=FirehoseServiceTest test`
- Focused Firehose integration suites: `FirehoseIntegrationTest`, `FirehoseFlushIntegrationTest`, `FirehoseFlushRecordCountIntegrationTest`, `FirehoseExtendedS3IntegrationTest`, and `FirehoseCompressionIntegrationTest`

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
+++ b/src/main/java/io/github/hectorvent/floci/services/eventbridge/EventBridgeInvoker.java
@@ -122,11 +122,12 @@ public class EventBridgeInvoker {
                     LOG.warnv("EventBridge Firehose target missing Firehose service: {0}", arn);
                     return;
                 }
-                String streamName = arn.substring(
-                        arn.indexOf(":deliverystream/") + ":deliverystream/".length());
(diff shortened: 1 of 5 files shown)


hectorvent (approved):
Thank you, and the last commit is the right call: an opaque iterator token with no region cannot be attributed to one of two same-named streams, so not migrating it and starting from the configured timestamp is safer than guessing. Approving.

PR #3268, opened 2026-09-08
fix(mq): match CreateBroker EngineType and DeploymentMode without regard to case

allensanborn:
## Summary

`CreateBroker` compared `EngineType` and `DeploymentMode` with `String.equals`, so the spelling the AWS docs, the console and the Terraform provider all use was rejected:

```
An error occurred (BadRequestException) when calling the CreateBroker operation:
Only RABBITMQ EngineType is supported
```

That is Floci refusing the exact engine it is being asked for, because the request said `RabbitMQ` rather than `RABBITMQ`. `engine_type = "RabbitMQ"` is the spelling in every published example, so `aws_mq_broker` could not create a broker at all.

Both comparisons are now case-insensitive, and `deploymentMode` is stored canonicalised so `DescribeBroker` reads back the enum value the SDKs expect however the request happened to be spelled.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: a documented, SDK-emitted casing was rejected with `BadRequestException`.

Real MQ accepts any casing for these members and normalises to the upper-case wire enum. Verified against the botocore model shipped with awscli 2.36.30 (`mq/2017-11-27/service-2.json`): `EngineType` and `DeploymentMode` are string members whose documented values are the upper-case forms, and the service normalises rather than rejecting. This change only widens what is accepted; nothing that worked before now fails.

## Checklist

Coverage: `AmazonMqEngineTypeCaseTest` (new) plus cases in `AmazonMqServiceTest`; 27 tests green across `AmazonMq*`. Reverting `AmazonMqService.java` to main leaves 1 failure and 4 errors, so the tests are load-bearing rather than shape-matching. Two of the seven pass either way by design: they pin the already-working upper-case path so a future change cannot regress it silently.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/amazonmq/AmazonMqService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/amazonmq/AmazonMqService.java
@@ -88,10 +88,17 @@ public class AmazonMqService implements ResourceProvider {
         }
         String deploymentMode = params.deploymentMode() == null
                 ? DEPLOYMENT_SINGLE_INSTANCE : params.deploymentMode();
-        if (!DEPLOYMENT_SINGLE_INSTANCE.equals(deploymentMode)) {
+        // DeploymentMode has the same problem as EngineType above: the wire enum is upper case,
+        // but callers spell it however their tooling does, and the real API matches without
+        // regard to case. An exact compare rejected "single_instance" -- the one mode this
+        // emulator does support -- as unsupported.
+        if (!DEPLOYMENT_SINGLE_INSTANCE.equalsIgnoreCase(deploymentMode)) {
             throw new AwsException("BadRequestException",
                     "Only SINGLE_INSTANCE DeploymentMode is supported", 400);
         }
+        // Store the canonical wire casing, not the caller's, so DescribeBroker reads back the
+        // enum value the SDKs expect however the request happened to be spelled.
+        deploymentMode = DEPLOYMENT_SINGLE_INSTANCE;
         // RabbitMQ brokers require exactly one user at creation; that user becomes the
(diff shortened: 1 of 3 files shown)


allensanborn:
Pushing back on this one, with numbers, since it is flagged as "must be satisfied before merging".

`AGENTS.md:205` does say "Unit tests: `*ServiceTest.java`", but that is not what the repository does. Of the 1247 test files under `src/test/java/io/github/hectorvent/floci/services/`, **100** match `*ServiceTest.java` and **1147** do not. `AcmEdgeCaseTest`, `AcmIdempotencyTest`, `AccessAnalyzerIntegrationTest` and `ApiGatewayOpenApiImportTest` are all long-standing examples of the same shape as `AmazonMqEngineTypeCaseTest`: a focused class named for the behaviour it pins rather than for the service.

The convention as practised looks like `*ServiceTest` for broad per-service unit suites, and a descriptive name when a class exists to pin one specific behaviour. That is what this file is.

Happy to fold the cases into `AmazonMqServiceTest` if a maintainer would rather, but I do not think a 100-versus-1147 split should read as a merge blocker.

Worth noting the same reviewer did not flag `ApiGatewayRestApiResponseFieldsTest`, a new file on #3269 with the identical naming shape, so the rule is not being applied consistently either.

pgermosen (approved):
Checked the model side of this - botocore's EngineType/DeploymentMode enums are upper-case only, and validate.py doesn't enforce enum membership client-side, so the mixed-case request genuinely reaches the wire as-is, consistent with the bug you're describing. The DescribeBroker read-back canonicalization is a nice touch too. Since this only widens acceptance, nothing that worked before can regress. Approving.

PR #3269, opened 2026-09-08
fix(apigateway): report rootResourceId and the API key defaults

allensanborn:
## Summary

`GetRestApi` and `GetRestApis` never returned `rootResourceId`, so a caller had to list resources and find the `/` entry itself. Terraform's `aws_api_gateway_rest_api` reads `root_resource_id` directly off the API, and anything referencing `aws_api_gateway_rest_api.x.root_resource_id` could not resolve.

This reports `rootResourceId` on both read paths, plus the two API key defaults (`apiKeySource`, `disableExecuteApiEndpoint`).

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `rootResourceId` was absent from a response where real API Gateway always includes it. The value is the id of the auto-created `/` resource, which Floci already creates; it was simply never surfaced.

On the two defaults, being explicit about what is measured and what is chosen: `apiKeySource=HEADER` and `disableExecuteApiEndpoint=false` are emitted as constants. That is honest here rather than a guess, because neither member is stored on the v1 `RestApi` model nor settable through this API surface: `createRestApi` ignores both, and `updateRestApi` patches only `/name` and `/description`. If either becomes settable later, the constant is the thing to replace. The v2 `disableExecuteApiEndpoint` path is unaffected; it goes through a different builder.

An earlier revision of this branch also set `apiStatus`. That has since landed on main independently, so the hunk was dropped rather than duplicated.

## Checklist

Coverage: `ApiGatewayRestApiResponseFieldsTest`, 5 tests. Reverting `ApiGatewayController.java` to main fails all 5. Full `-Dtest=ApiGateway*` sweep is 836 tests green, including the v2 endpoint tests, so nothing regressed.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
@@ -8,7 +8,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 
 import com.fasterxml.jackson.core.type.TypeReference;
@@ -1934,7 +1933,7 @@ public class ApiGatewayController {
         // as aws_api_gateway_rest_api.root_resource_id, which is how the first resource
         // under "/" gets its parent, so leaving it out breaks the conventional way of
         // building a REST API.
-        rootResourceId(region, api.getId()).ifPresent(id -> node.put("rootResourceId", id));
+        service.findRootResourceId(region, api.getId()).ifPresent(id -> node.put("rootResourceId", id));
 
         // Neither member is settable on a REST API here: createRestApi ignores both and
         // updateRestApi patches only /name and /description, so the emulated value is always
@@ -1945,13 +1944,6 @@ public class ApiGatewayController {
         node.put("disableExecuteApiEndpoint", false);
 
         return node;
-    }
-
-    private Optional<String> rootResourceId(String region, String apiId) {
-        return service.getResources(region, apiId).stream()
-                .filter(r -> "/".equals(r.getPath()))
-                .map(ApiGatewayResource::getId)
-                .findFirst();
     }
 
     private ObjectNode toResourceNode(ApiGatewayResource r) {
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayController.java
@@ -8,6 +8,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import com.fasterxml.jackson.core.type.TypeReference;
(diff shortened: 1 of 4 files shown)


allensanborn:
All three land. Verified each against the code before agreeing, and two are being fixed now.

**P1 "Root ID can disappear"** - confirmed. `ApiGatewayService.deleteResource()` has no guard on the `/` resource, so it can be deleted and `rootResourceId` then drops out of every subsequent read. Real AWS rejects deleting a REST API's root resource, so the fix is the guard rather than recreating the resource after the fact.

**P1 "Concurrent delete breaks listing"** - confirmed, and worth being precise about the mechanism: `getResources` opens with `getRestApi(region, apiId)`, which throws `NotFoundException`. Because the root lookup happens per item during serialization, an API deleted between the snapshot and rendering turns the entire `ListRestApis` response into a 404 rather than dropping one entry. Making the per-item lookup tolerate a vanished API, catching only the not-found case so unrelated failures still surface.

**P2 "Listing repeats full scans"** - agreed on the analysis: `resourceStore.scan(k -> k.startsWith(prefix))` once per API on an unpaginated endpoint is O(N x R). I would rather not widen this PR to add an index or a single-scan root lookup, since that is a storage-layer change with its own review surface and this PR is meant to be the read-field fix. If the fix turns out to be small and local I will include it; otherwise I will file it separately and link it here rather than leaving it unrecorded.

allensanborn:
Both P1s fixed in 71903521 and 93ebf979. The P2 I am deferring, with a reason.

**Root resource deletion** - `deleteResource` now refuses the resource whose path is `/`, raising `BadRequestException`/400, the type and shape this file already uses for AWS rejections. Only two callers reach it: the HTTP route, and `putRestApi`, which already skips `/` explicitly. `deleteRestApi` cascades through the store directly, so deleting a whole API still works.

Being precise about the evidence, since this project rightly separates measured from inferred. The botocore model (`apigateway/2015-07-09/service-2.json`, awscli 2.36.30) lists `BadRequestException` among `DeleteResource`'s errors, so the error *type* is model-grounded. The model does **not** state that the root is undeletable, and I have not probed a live account, so I am not claiming verified AWS parity for the refusal itself. The reasoning for refusing anyway: the root is created with the API, has no `pathPart` to address it by, and this PR now reports its id as `rootResourceId`, so deleting it leaves an API that can neither report a root nor parent a new resource, with no way back short of recreating the API. Happy to relax it if someone confirms real AWS allows the delete.

**Concurrent delete** - the root lookup moved into `ApiGatewayService.findRootResourceId(region, apiId)`, which catches only `NotFoundException` from `getResources` and returns `Optional.empty()`. Every other `AwsException` still propagates, and there is a test asserting that narrowness rather than only the happy path: `getResources` on a vanished id still throws, so the catch is not swallowing the read wholesale.

Worth recording a wrong turn, because it would have been an unpleasant surprise in CI. The first version of that test used `@InjectSpy ApiGatewayService` to stage the race at the HTTP layer. It passed alone but broke six unrelated tests in the full `ApiGateway*` sweep, all 502/403 on the data plane, because the spy bean leaks across test classes. It is replaced with a mock-free service-level test.

**Repeated full scans (P2)** - deferring deliberately, not dropping. The analysis is right: `resourceStore.scan` walks every key, once per API, on an unpaginated endpoint. The clean fix is to stop scanning at all - carry `rootResourceId` on the `RestApi` model, set it where the root is created, read it directly - but that means a model field, a persistence change, a fallback for APIs already stored without it, and the `importRestApi` path. That is a larger and riskier diff than the two correctness fixes it would ride along with, and it belongs in its own PR. I will file it as an issue and link it here rather than leave it unrecorded.

Verification: `ApiGateway*` sweep is 841 tests, up from 836, with the five new ones green. Both fixes checked load-bearing: removing the root guard fails `theRootResourceCannotBeDeleted` alone, and collapsing `findRootResourceId` back to the bare lookup fails `theRootLookupOfAVanishedApiIsEmptyRatherThanNotFound` alone.

One caveat on my own numbers: this machine was running another full build concurrently, and under that load the three data-plane classes (`ApiGatewayRequestAuthorizerIntegrationTest`, `ApiGatewayApiKeyRevocationIntegrationTest`, `ApiGatewayAuthorizerContextIntegrationTest`) fail on unmodified `upstream/main` too, 16 failures there against 20 here across 10 more tests. Same classes, same 502s on both arms, so it is load, not this change. The five tests this PR adds pass cleanly in isolation.

hectorvent (approved):
`rootResourceId` is a real `RestApi` member in the model, so returning it is the fix. Refusing to delete the root resource rather than recreating it matches AWS and is the better of the two options, and the concurrent-delete guard keeps a listing from failing wholesale. Approving; the per-item scan is fine to leave as a follow-up.

allensanborn:
Filed the deferred P2 as #3305 so it is recorded rather than left in this thread: `ListRestApis` scans every resource once per API, O(N x R), on an unpaginated endpoint.

The issue includes the two call sites, why the prefix filter does not avoid the walk, and the two candidate fixes (carry `rootResourceId` on the `RestApi` model, or a single grouped scan), along with the migration and `importRestApi` considerations that made it too big to ride along with the correctness fixes here.

PR #3270, opened 2026-09-08
docs(lambda,s3): mark two leniencies as Floci's call, not observed AWS

allensanborn:
## Summary

Two documentation lines claimed AWS parity for behaviour that is Floci's own choice. Both were raised in earlier reviews and left as follow-ups; this is that follow-up.

- `docs/services/lambda.md`: the `ApplicationLogLevel` / `SystemLogLevel` sentence said the no-op-under-`Text` behaviour matched "the fact that the response never surfaces them for Text". That points at Floci's own response shape, not at anything observed on AWS, so the request behaviour was being justified by the response behaviour and both are our call. Now marked as such.
- `docs/services/s3.md`: `log-delivery-write` was described as accepted on `PutBucketAcl`/`PutObjectAcl` "matching AWS". AWS documents that canned ACL as bucket-only and the SDK's `ObjectCannedACL` enum omits it. The bucket path matches AWS; the object path is a Floci leniency, and it now reads that way.

Docs only, no behaviour change.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No API behaviour is changed. The point is to stop the docs asserting parity that was never probed, so nobody later cites those lines as a measured result.

The S3 leniency is real and shared: `putBucketAcl` and `putObjectAcl` both resolve through the same `cannedObjectAclXml` helper, so the object path also accepts `log-delivery-write`, and the bucket path accepts `aws-exec-read` which the SDK models as object-only. That mirror-image case is named in the docs rather than silently left. Enum membership confirmed against the botocore `s3` model shipped with awscli 2.36.30.

## Checklist

No test: the change is prose in two docs files, with no production change to exercise. Flagging that explicitly rather than leaving the unticked box unexplained.


pgermosen:
Good instinct on this cleanup, but the aws-exec-read half of the new S3 sentence has the same problem it's fixing. Checked AWS's actual canned ACL reference table (the "Applies to" column), and aws-exec-read is listed as "Bucket and object", not object-only. What's actually object-only is botocore's BucketCannedACL enum, which omits it for client-side validation on PutBucketAcl, while ObjectCannedACL includes it. AWS's own documented behavior isn't restricting it the way the SDK enum does.

So "which AWS lists only for objects" isn't quite right, it's the SDK modeling that's narrower there, not AWS itself. Something like "which botocore's BucketCannedACL enum omits, though AWS's canned-ACL reference documents it as valid for both bucket and object" would hold up. The log-delivery-write half of the sentence is accurate as written.

pgermosen (commented):
The lambda.md half is good as is. The s3.md one still has the issue from the earlier comment though, looks like the merge commit synced with main but didn't actually fold in the fix. Pulled up AWS's own canned ACL table to check rather than just going by the earlier comment: aws-exec-read is listed there as bucket and object, not object-only, so the current sentence has it backwards. What's actually narrower is botocore's BucketCannedACL enum, not AWS's documented behavior. log-delivery-write being bucket-only is correct as written, that half just needs the aws-exec-read framing flipped to point at the SDK enum rather than AWS itself.

allensanborn:
You are right on both counts, and I owe you an apology for the second one.

**On the substance:** I checked the shipped botocore model rather than just taking the correction, and it confirms exactly what you said. `BucketCannedACL` is `private, public-read, public-read-write, authenticated-read`; `ObjectCannedACL` adds `aws-exec-read`, `bucket-owner-read`, `bucket-owner-full-control`. So the narrowing is client-side SDK validation on `PutBucketAcl`, not AWS declining the ACL on buckets. My sentence attributed it to AWS, which is the precise error this PR exists to correct, made inside the correction.

Reworded in d9b7e0b6 to: "which botocore's `BucketCannedACL` enum omits, though AWS's canned-ACL reference documents it as valid for both bucket and object; there the narrower party is the SDK model, not AWS." The `log-delivery-write` half is unchanged, since you confirmed it was accurate.

**On the process:** you were right that the merge commit synced with main without folding in your fix, and the reason is worse than an oversight. I had been checking for review feedback with the inline-comments endpoint only (`pulls/N/comments`), which does not return top-level PR comments (`issues/N/comments`). Your comment was top-level, so my check reported "no findings" and I pushed a routine upstream merge on top of it. From your side that reads as having seen your review and ignored it, which is a fair reading of what it looked like. I have fixed how I check.

pgermosen (approved):
That's exactly right now, matches AWS's canned ACL table on both halves. Appreciate you checking the shipped botocore model directly rather than just taking the correction at face value, that's the right instinct for a PR whose whole point is not asserting things that weren't actually checked.

Good to merge.

PR #3271, opened 2026-09-08
docs(lambda): say blank LogGroup is deliberately read as not supplied

allensanborn:
## Summary

`validateLogGroup` returns early when `LogGroup` is blank, so `applyLoggingConfig` then treats it as absent and falls back to `/aws/lambda/<name>`. The behaviour is deliberate, but nothing said so: the javadoc above it only restated the documented 1-512 length constraint, which a blank value violates. A future reader "fixing" the apparent gap would change behaviour without realising it was a decision.

This documents the decision and pins it with a test.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Deliberately NOT adding a rejection, and the reasoning is worth recording since the opposite is the tempting change.

`LogGroup` is modelled `{"min": 1, "max": 512}`, so `""` is a violation on paper. But botocore enforces that minimum **client-side**: `validate.py:_validate_string` calls `range_check` on `len(param)` before anything is sent. An empty `LogGroup` therefore never reaches the Lambda service from any SDK caller, and nobody has observed what the service itself answers to one. Returning a `ValidationException` here would be inferred behaviour presented as measured.

Given that, the cheaper error is to keep accepting it: an emulator that accepts slightly more than AWS costs nothing, while one that rejects a valid request blocks real work. If someone does probe a live account and finds AWS rejects it, the javadoc says exactly where to change.

Model checked against botocore `lambda/2015-03-31/service-2.json` shipped with awscli 2.36.30.

## Checklist

`blankLogGroupIsReadAsNotSuppliedRatherThanRejected` in `LambdaVpcSnapStartLoggingIntegrationTest`; 22 tests green. The production side is javadoc, so reverting it cannot fail anything. The meaningful check is the other direction: making `validateLogGroup` throw on a blank group fails exactly that one test and nothing else, so the decision is now guarded rather than merely described.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaService.java
@@ -2266,6 +2266,16 @@ public class LambdaService implements ResourceProvider {
     /**
      * LogGroup is the one LoggingConfig member with a documented length and character
      * constraint rather than an enum: 1-512 characters, {@code [.\-_/#A-Za-z0-9]+}.
+     *
+     * <p>A blank LogGroup (empty or whitespace-only) is deliberately read as "not supplied"
+     * rather than as a violation of that 1-character minimum, so {@link #applyLoggingConfig}
+     * falls back to the {@code /aws/lambda/} default exactly as it does for an absent member.
+     * The minimum is real in the service model, but botocore enforces it client side, so an
+     * empty LogGroup never reaches the wire from an SDK caller and nobody has observed what
+     * the service itself answers to one. Rejecting it here would be a 400 we inferred rather
+     * than measured; accepting it costs a caller nothing. That leniency is pinned by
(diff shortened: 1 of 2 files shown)


allensanborn:
Fair point on the whitespace case, and it is a gap between what the javadoc claims and what the test pins: the doc says "empty or whitespace-only", the guard is `isBlank()`, but the test only exercises `""`. I will extend the test to cover a whitespace-only value so the assertion matches the documented contract rather than a subset of it.

pgermosen (approved):
Checked the botocore claim directly - validate.py's _validate_string does call range_check before anything is sent, and the current LogGroup shape is exactly {min:1, max:512} with the pattern quoted in the javadoc, so the reasoning here is solid, not assumed. Good call pinning it with a test instead of just a comment, and nice catch closing the whitespace-only gap too. Approving.

PR #3272, opened 2026-09-08
feat(ec2): back each VPC with a real Docker network in RFC 1918 space

allensanborn:
## Summary

Every EC2 instance container is attached to the default Docker bridge, so a VPC's declared CIDR is decorative: instances get bridge addresses unrelated to their subnet, two VPCs cannot use overlapping ranges meaningfully, and a private address is not stable across a restart.

This backs each VPC with a real Docker network in RFC 1918 space and allocates instance addresses from the declared subnet.

- New `services/ec2/net/` package: `Cidr4` (parsing, containment, sub-block allocation) and `VpcNetworkManager` (per-VPC network lifecycle, per-subnet address pools, leases).
- Addresses come from the declared subnet CIDR, skipping the network address and gateway.
- Three substitution triggers when the declared CIDR cannot be used verbatim: it is not RFC 1918, it is absent, or it collides with an existing Docker network. Two VPCs may declare the same CIDR; the first keeps it and the second is substituted.
- Leases are returned when a launch fails or an instance never gets an address, and re-reserved on startup for instances that survived a restart.
- Orphan reconciliation removes only this emulator's own networks, matched on labels and owner port.
- Off by default in tests, on in `application.yml`, switchable via `floci.services.ec2.vpc-networks.enabled`.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Addressing only, and the docs say so rather than overclaiming. An instance now gets an address inside its declared subnet and keeps it across a restart, which is what the AWS provider and most Terraform modules actually read.

This is explicitly NOT network isolation. Docker network separation is not enforced by every runtime: under OrbStack in particular, containers on separate networks remain mutually reachable, so the guarantee here is addressing and planning, not traffic containment. That limit is stated in `docs/services/ec2.md` rather than left for someone to discover.

Every test mocks the Docker client and asserts addressing, allocation and lifecycle. None asserts that traffic between two networks is blocked, deliberately, because that would not hold on all runtimes.

## Checklist

Coverage: `Cidr4Test` 6, `VpcNetworkManagerTest` 21, plus wiring cases in `Ec2ContainerManagerTest` and `Ec2ServiceTest`. 265 tests green across `Cidr4Test`, `VpcNetworkManagerTest`, `Ec2ContainerManagerTest`, `Ec2ServiceTest`, `EksServiceTest`. The two new test classes cannot compile without the new production classes; the four wiring tests were checked by neutering the lease-release and re-reservation call sites, which fails exactly those four and nothing else.

One design note for review: on the containerless-degrade path (`isDockerAvailable()` false) the lease is deliberately kept rather than released, because the instance still reports its declared private address and the lease is released at terminate.


allensanborn:
Thanks - four findings, and they do not all land the same way. Taking them in order, with what I verified for each.

### P1 "VPC leases become permanent" - I do not think this holds

The claim is that start/restore replace `privateIpAddress` with the container's **bridge** address, so `terminate()` releases the wrong one. Tracing it:

- `start()` calls `waitForContainerBridgeIp(...)` -> `getContainerBridgeIp(containerId)` (`Ec2ContainerManager.java:1665`).
- That delegates to `preferredMetadataSourceIp(networks)` (`:1671`), which filters **out** the `"bridge"` entry and returns the first non-bridge address (`:1707-1716`); the bridge is only the fallback when no other network has an address.
- A VPC-attached container has a non-bridge network, so despite the method name the value returned is the **VPC** address.
- `exposeReachablePrivateAddress` therefore sets `privateIpAddress` to the VPC address, which is exactly what `terminate()` releases (`:995`).

So in the normal case the correct address is released. The residual risk is narrower than described: if the VPC attachment is genuinely gone, `preferredMetadataSourceIp` falls back to the bridge address and the lease would then leak. I would rather fix that precise case than the one described. Correct me if I have misread the ordering.

### P1 "Stale network IPAM is reused" - confirmed, fixing

`materialise()` adopts a surviving network on a successful `inspectNetworkCmd` and sets `created = true` without comparing the network's IPAM subnet to `vpc.effective`. The collision-disappears-before-restart scenario is real. Fixing by comparing the inspected subnet against the plan and not adopting silently on mismatch.

### P2 "Small fallback blocks cannot allocate" - confirmed, and slightly worse than stated

`FIRST_HOST_OFFSET` is 10 (`:106`) and `fallbackPrefixLength()` is `@WithDefault("16")` with no upper bound. A /29 has 8 addresses and a /30 has 4, so allocation can never return one. The declared-prefix path also caps at `Math.min(declaredPrefix, 30)` (`:300`), so /29 and /30 are reachable from a declared CIDR too, not only from misconfiguration. Fixing so an unusable prefix fails loudly instead of silently degrading to synthetic addressing.

### P2 "Documentation uses prohibited punctuation" - not a blocker here

`AGENTS.md:338` does say no em-dashes anywhere. But the CI check that enforces it is named "New files omit em-dashes" and scopes itself with `--diff-filter=A`; its own inline comment says including modified files "would make this check noise rather than signal". `docs/services/ec2.md` is pre-existing and already contains em-dashes on lines 3, 108, 125 and 280, none of which this PR introduced. The check passes.

I have left those four lines matching the surrounding file rather than making this PR the one that half-cleans an existing document. Happy to strip them if a maintainer prefers, but it is a style call rather than a requirement being violated.

allensanborn:
Fixed in 670224f2 and 6414f871, plus one finding that was not raised here.

**Stale network IPAM (P1)** - `materialise()` now compares the surviving network's IPAM subnet against the VPC's plan. On a mismatch it removes and recreates rather than re-planning, because by that point `attach()` has already validated live callers against `vpc.effective`, and after a restart some of those addresses are leases held by running containers - the plan cannot move out from under them. Recreating is safe here specifically because the network name embeds this emulator's port and region (`floci-vpc-<port>-<region>-<vpcId>`), so a leftover under that name is by construction Floci's own and never a user's network. A leftover with unreadable IPAM is treated as a mismatch rather than adopted. If removal fails, the instance stays on the shared bridge, which is the existing documented degradation, instead of holding an address the network does not route.

**Small fallback blocks (P2)** - covered at all three entry points, since they fail differently. The first-host offset is now per-subnet: 10 when the block has at least 12 addresses, otherwise 2, so a /29 yields .2-.6 while a /24 still yields 245 unchanged. A block with no usable host at all (/31, /32) is refused at plan time with a WARN and falls back, rather than claiming a range and then returning nothing from every allocation, which reads as exhaustion. And `fallback-prefix-length` is clamped into `[pool prefix, 28]` with a WARN; that same constant replaces the hardcoded `30` in the declared-prefix path, which is what made /29 and /30 reachable from a declared CIDR rather than only from misconfiguration.

**IMDS source address** - not flagged in review, found while tracing the lease question above, and I think it is the more serious of the two. `imdsSourceIp` had exactly one write site (in `launch()`) and one read site (in `terminate()`). Neither `start()` nor `restoreMetadataRegistration()` refreshed it, and both registered only the address `preferredMetadataSourceIp` returns, which for a VPC-attached container is the VPC address. But the container's default route is the bridge, so its IMDS requests arrive from the new bridge address Docker hands out on restart - which was never registered. IMDS would break for a VPC-attached instance after a stop/start or an emulator restart, and the old address would stay registered indefinitely. Both paths now call `refreshImdsSourceRegistration(...)`, which inspects the `"bridge"` network specifically via a new `bridgeNetworkIp(containerId)` - deliberately separate from the misleadingly named `getContainerBridgeIp`, which does not return the bridge address. A bridge-only instance is unaffected and keeps a null `imdsSourceIp`.

Verification: 274 tests green across `Cidr4Test`, `VpcNetworkManagerTest`, `Ec2ContainerManagerTest`, `Ec2ServiceTest`, `EksServiceTest`, up from 265, and the boot test is clean with no `SRCFG00050`. Six new tests, each checked load-bearing by reverting the corresponding production change and confirming only the expected ones fail. Three further guard tests pass either way by design and are labelled as such.

One test I did not add, for the record: a check on the pool-prefix floor. `allocateSubBlock` already does `Math.max(desiredPrefix, parent.prefix())`, so that floor is behaviourally a no-op and no test of it could be load-bearing. It stays in the code only so the WARN reports the real range.

The remaining open item from your review is the em-dash one, which I addressed in my previous comment: the enforcing check scopes itself to newly added files and `docs/services/ec2.md` is pre-existing.

allensanborn:
The one red check on this branch is `Build and Test (shard 4)`, and I do not think it is this change. Evidence rather than assertion:

The failure is a single test, `RdsAwsIntegrationTest.createDescribeRestoreRoundTrip`, on `CreateDBSnapshot`:

```
<Message>Failed to create snapshot: Failed to create postgres snapshot</Message>
```

Reading the surrounding log, the RDS container logged `database system is ready to accept connections` at 17:30:26.952 and the snapshot request went out at 17:30:26.970. Eighteen milliseconds. That is a Postgres startup race, not a behaviour change: `ready to accept connections` is logged before the server is reliably able to serve a dump.

This PR touches EC2 VPC networking and cannot reach RDS. `git diff upstream/main...HEAD --name-only` lists no RDS file, and the only case-insensitive "rds" matches in the diff body are inside the words `Records` and `records` in new javadoc.

Being honest about the limits of the claim: I could not find this exact test failing on another branch, so I am calling it a flake by mechanism rather than by prior occurrence. What I can say is that `main` itself has a red CI run from the same window, failing a different container-dependent integration test, which is consistent with a flaky period rather than something specific to this branch. Shard 4 is also where the container-heavy suites live.

Locally the areas this PR does touch are green: 274 tests across `Cidr4Test`, `VpcNetworkManagerTest`, `Ec2ContainerManagerTest`, `Ec2ServiceTest` and `EksServiceTest`, plus a clean boot with no `SRCFG00050`.

A rerun of that job would settle it. I do not have permission to trigger one on this repository, so I am leaving it rather than pushing an empty commit to force a new run.

hectorvent:
This is a real capability upgrade, and the review round is exactly right: you traced the lease claim rather than accepting it, and `clampAllocatablePrefix` plus the `firstHostOffset` scaling already close the small-fallback-block case, with a test pinning it.

One thing before merge. The added text carries six em dashes: four in `docs/services/ec2.md` and two in a javadoc comment in `Instance.java`. AGENTS.md forbids them in all content. The "New files omit em-dashes" check runs with `--diff-filter=A`, so it only scans files a PR adds, and both of these are existing files. CI is green on a real rule violation.

Unrelated: shard 4 is red on `RdsAwsIntegrationTest.createDescribeRestoreRoundTrip`, which is failing the same way on other PRs today. Nothing to chase there.

allensanborn:
Merged current `main` into the branch, which brings it up to date and re-triggers CI. The previous red check was a stale run of `RdsAwsIntegrationTest.createDescribeRestoreRoundTrip` from before the fix commits; no CI had run since, so the failure shown was not against the current head.

Re-verified after the merge, against today's `main`:

- `./mvnw test-compile` clean. Worth a note: the merge was textually clean, but that is exactly the case where a semantic break hides, so I did not treat a clean merge as verification. Compiling test sources is what would catch a constructor-arity drift, and it is clean.
- Targeted suites: **278 tests, 0 failures** across `Cidr4Test`, `VpcNetworkManagerTest`, `Ec2ContainerManagerTest`, `Ec2ServiceTest`, `EksServiceTest`. That is up from 274, since `main` added tests to those classes in the interim and they pass alongside this change.
- Boot test `Ec2ImageCatalogTest`: green, zero `SRCFG00050`.

No behaviour changed in this push; it is the merge plus re-verification.

allensanborn:
Fixed in c92e579a. This one was mine to clean up: it was introduced by the `refreshImdsSourceRegistration` change earlier in this PR, not by the original feature.

`bridgeNetworkIp` returned `null` for three different situations the caller could not tell apart: the container genuinely has no bridge address, the inspect response had no network settings, or `inspectContainerCmd(...).exec()` threw and was swallowed by a `catch (Exception)`. Only the first two justify unregistering. In the third the code cleared a healthy instance's IMDS source registration and put nothing back, so its metadata requests stopped resolving until some later start or restore happened to succeed.

Fixed by separating the indeterminate case from the definite one rather than by retrying: `bridgeNetworkIp` no longer swallows the exception, and `refreshImdsSourceRegistration` catches it and returns early, leaving the existing registration untouched. An inspect that failed says nothing about where the container is attached, and the unregister is only correct for a container that definitely has no separate bridge address.

Coverage: `restoreMetadataRegistrationKeepsTheImdsSourceWhenTheContainerCannotBeInspected`. Load-bearing checked by reverting only the production hunks, which fails it on `never wanted ec2MetadataServer.unregisterContainer`. 279 tests green across `Cidr4Test`, `VpcNetworkManagerTest`, `Ec2ContainerManagerTest`, `Ec2ServiceTest` and `EksServiceTest`, and the boot test is clean with no `SRCFG00050`.

On the four other comments in this round, all of which GitHub re-anchored to new line numbers rather than raising afresh: "Stale network IPAM is reused" and "Small fallback blocks cannot allocate" were fixed in 670224f2 and are described in an earlier comment; "VPC leases become permanent" I still believe does not hold, for the `preferredMetadataSourceIp` reason set out above; and the em-dash one concerns a pre-existing file that the enforcing check deliberately does not cover. Happy to revisit any of those if you disagree with the reasoning.

allensanborn:
You are right, and my earlier reasoning on this was wrong. Fixed in cc1b8eb7.

I had argued the em dashes were fine because the enforcing check scopes itself with `--diff-filter=A` and these are existing files, so CI passes. That conflates "the check cannot see it" with "it is not a violation". AGENTS.md forbids them in all content; the check's narrowness is a limitation of the check, not a licence. And the six in question are in lines this branch *adds*, which is the part I should have looked at rather than at which files it touches.

Six replaced with colons and commas, wording otherwise unchanged: four in `docs/services/ec2.md`, one in an `Ec2ContainerManager` javadoc, one in an `Instance` javadoc. `git diff upstream/main...HEAD | grep '^+' | grep -c '—'` is now 0.

I left the em dashes already present in those files alone. They predate this branch, and rewriting them here would bury a networking change in unrelated churn. Happy to do a separate sweep if you want the existing ones cleaned up, since the check will never catch those either.

279 tests still green across `Cidr4Test`, `VpcNetworkManagerTest`, `Ec2ContainerManagerTest`, `Ec2ServiceTest`, `EksServiceTest`.

Thanks also for the note on shard 4. That matches what I found: `RdsAwsIntegrationTest.createDescribeRestoreRoundTrip` fails on a `CreateDBSnapshot` eighteen milliseconds after the container logs "database system is ready to accept connections", and this branch touches no RDS file. Good to know it is failing the same way elsewhere; I will stop treating it as something to chase.

hectorvent:
That is exactly right, and thank you for saying so plainly. "The check cannot see it" and "it is not a violation" being different things is the whole of it, and I checked: no em dashes in any added line on this branch now.

Leaving the pre-existing ones alone was the right call too. Burying a networking change in unrelated churn would have made this harder to review, not easier. I will take you up on the separate sweep, since `--diff-filter=A` means nothing will ever catch those.

The rest of the round stands as it was: the lease refutation holds up when I follow the same path, and the fallback clamp was already handled by `clampAllocatablePrefix` and the host-offset scaling.

Your diagnosis of the RDS flake is more precise than mine was, and useful beyond this PR: a `CreateDBSnapshot` eighteen milliseconds after the container reports ready is a readiness race in that test, which is worth fixing at source rather than rerunning. That one is on me.

Merging this.

hectorvent (approved):
Approving: all review rounds closed.

PR #3273, opened 2026-09-08
feat(redshift): emulate GetClusterCredentials and GetClusterCredentialsWithIAM

vuthehuyht:
## Summary

Emulates `redshift:GetClusterCredentials` and `redshift:GetClusterCredentialsWithIAM`. Both mint a short-lived `DbUser` / `DbPassword` pair that the PostgreSQL wire auth proxy and the Redshift Data API accept for a non-master user, without creating a real PostgreSQL role.

How it works: a Floci-side credential shim. The action stores a `TempCredential` (with expiry) in an in-memory `RedshiftCredentialBroker`. When the proxy or the Data API sees a live minted credential, it opens the real backend connection as the cluster master (the `DbUser` is nominal, so `current_user`, `GRANT`, and object ownership are the master's; this is documented as a limitation).

To carry that decision, the shared `PasswordValidator` contract used by the RDS/Redshift PostgreSQL proxy handlers widens from `boolean` to a three-value `AuthResult` (`REJECT` / `PASSTHROUGH` / `MASTER_EQUIVALENT`). RDS behaviour is unchanged: its own master-password callback (renamed `RdsAuthProxy.MasterPasswordCheck` to free the `PasswordValidator` name) is adapted at the handler boundary, and a non-master RDS user still passes straight through to the backend.

Main changes:
- `RedshiftCredentialBroker` + `TempCredential` (in-memory, `Resettable`, no `StorageFactory`).
- `PasswordValidator` -> `AuthResult`; `PostgresProtocolHandler.resolveBackendLogin(...)` selects the backend login.
- `RedshiftService.passwordValidatorFor` consults the broker; the master username is always decided on its own password (match -> `MASTER_EQUIVALENT`, mismatch -> `REJECT`), and an absent cluster row -> `REJECT`.
- `RedshiftQueryHandler`: two new Query-protocol actions; `AwsQueryController` registers them and forwards the `Authorization` header.
- `RedshiftIamDbUserResolver`: maps the caller's IAM identity to `IAM:<user>` / `IAMR:<role>`, `IAMR:floci` fallback.
- `RedshiftDataResourceResolver`: a non-master `DbUser` with a live minted credential resolves to the master connection instead of `ValidationException`.
- New config `floci.services.redshift.default-credential-duration-seconds` (default `900`; `DurationSeconds` range 900-3600 inclusive).
- Docs: `docs/services/redshift.md` (regenerated action table + limitations), `docs/services/redshift-data.md`, Service Matrix.

Note: the `GetClusterCredentials` response also echoes a `<DbGroups>` element, which the real AWS shape omits. It is an intentional Floci extension (SDK v2 ignores unknown elements on unmarshal).

## Type of change

- [x] New feature (`feat:`)

## AWS Compatibility

- New actions verified against AWS SDK for Java v2 (`software.amazon.awssdk:redshift`, the version pinned in `compatibility-tests/sdk-test-java`): `RedshiftClient.getClusterCredentials` / `getClusterCredentialsWithIAM`, asserting `dbUser` / `dbPassword` / `expiration` are populated. Wire format is AWS Query protocol XML built with `XmlBuilder`, routed through `AwsQueryController.REDSHIFT_ACTIONS`.
- `DurationSeconds` outside 900-3600, a non-integer value, and a missing `DbUser` return `InvalidParameterValue`; an unknown `ClusterIdentifier` propagates `ClusterNotFound` (404).

## Checklist

- [x] `./mvnw test` passes locally: every affected suite passes (`RedshiftServiceTest`, `RedshiftQueryHandlerTest`, `RedshiftCredentialBrokerTest`, `RedshiftIamDbUserResolverTest`, `PostgresProtocolHandlerTest`, `PostgresProtocolHandlerAuthResultTest`, `MySqlProtocolHandlerTest`, `RdsServiceTest`, `RedshiftAuthProxyTest`, `RedshiftProxyManagerTest`, `RedshiftDataResourceResolverTest`, `RedshiftDataServiceTest`, `RedshiftOperationsTest`). A full-suite local run could not complete on this machine (the compile step was repeatedly killed by an OS resource limit, unrelated to these changes); CI runs the full suite.
- [x] New or updated integration test added: `RedshiftGetClusterCredentialsIntegrationTest` (Docker-gated, end-to-end: mint a credential for a non-master `DbUser`, then run `SELECT 1` through the Data API as that user), plus SDK compat coverage in `compatibility-tests/sdk-test-java/.../RedshiftTest.java`.
- [x] Commit messages follow Conventional Commits.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
@@ -153,14 +153,14 @@ public class RedshiftTest {
     }
 
     @Test
-    @Order(2)
-    public void testGetClusterCredentialsReturnsTemporaryCredentials() {
+    @Order(3)
+    public void testGetClusterCredentialsReturnsTemporaryCredentials() throws Exception {
         RedshiftClient client = getClient();
-        GetClusterCredentialsResponse res = client.getClusterCredentials(b -> b
+        GetClusterCredentialsResponse res = withRetry(() -> client.getClusterCredentials(b -> b
                 .clusterIdentifier("test-cluster")
                 .dbUser("analyst")
                 .dbName("dev")
-                .durationSeconds(900));
+                .durationSeconds(900)));
 
         assertNotNull(res.dbUser());
         assertTrue(res.dbUser().contains("analyst"));
@@ -170,13 +170,13 @@ public class RedshiftTest {
     }
 
     @Test
-    @Order(2)
-    public void testGetClusterCredentialsWithIamReturnsIamPrefixedUser() {
+    @Order(4)
+    public void testGetClusterCredentialsWithIamReturnsIamPrefixedUser() throws Exception {
         RedshiftClient client = getClient();
-        GetClusterCredentialsWithIamResponse res = client.getClusterCredentialsWithIAM(b -> b
+        GetClusterCredentialsWithIamResponse res = withRetry(() -> client.getClusterCredentialsWithIAM(b -> b
                 .clusterIdentifier("test-cluster")
                 .dbName("dev")
-                .durationSeconds(900));
+                .durationSeconds(900)));
 
         assertNotNull(res.dbUser());
         assertTrue(res.dbUser().startsWith("IAM"));
@@ -186,7 +186,7 @@ public class RedshiftTest {
     }
 
     @Test
-    @Order(3)
+    @Order(5)
     public void testDeleteCluster() throws Exception {
         RedshiftClient client = getClient();
         DeleteClusterResponse res = withRetry(() -> client.deleteCluster(DeleteClusterRequest.builder()
--- a/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/github/hectorvent/floci/compat/RedshiftTest.java
@@ -14,6 +14,8 @@ import software.amazon.awssdk.services.redshift.model.DescribeClustersRequest;
 import software.amazon.awssdk.services.redshift.model.DescribeClustersResponse;
 import software.amazon.awssdk.services.redshift.model.DeleteClusterRequest;
 import software.amazon.awssdk.services.redshift.model.DeleteClusterResponse;
+import software.amazon.awssdk.services.redshift.model.GetClusterCredentialsResponse;
+import software.amazon.awssdk.services.redshift.model.GetClusterCredentialsWithIamResponse;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.CreateBucketRequest;
 import software.amazon.awssdk.services.s3.model.GetObjectRequest;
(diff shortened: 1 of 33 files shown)


hectorvent:
Nice work, and the wire surface holds up: `ClusterNotFound` and `ClusterAlreadyExists` are the model's codes for those faults, `DbUser` is required only on the non-IAM call, and the 900 to 3600 bounds match the documented range.

One blocking item. On `GetClusterCredentials` the `AutoCreate` prefix is inverted. [Botocore](https://github.com/boto/botocore/blob/develop/botocore/data/redshift/2012-12-01/service-2.json) and the Java, Go and Ruby SDKs all say the returned name is prefixed `IAM:` when `AutoCreate` is False and `IAMA:` when it is True. The handler returns `IAM:` when True and no prefix when False.

`RedshiftQueryHandlerTest.getClusterCredentialsAutoCreatePrefixesIamOnDbUser` pins the current mapping and will need updating; the compat test asserts only `contains("analyst")`, so it passes either way. The broker keys on whatever string it issues, so the fix stays local.

Minor: three methods now share `@Order(2)` in `RedshiftTest`.

vuthehuyht:
Both addressed in 0fc844a and 057f8cd.

**AutoCreate prefix** (0fc844a): confirmed against the API reference ("prefixed with `IAM:` if `AutoCreate` is `False` or `IAMA:` if `AutoCreate` is `True`"). `GetClusterCredentials` now returns `IAMA:` when true and `IAM:` when false, so it also lines up with `GetClusterCredentialsWithIAM`, which already issued prefixed identities. The broker keys on the returned name, so the proxy and Data API now expect the prefixed `DbUser`; `RedshiftQueryHandlerTest`, the integration test, and the redshift / redshift-data docs are updated. `getClusterCredentialsAutoCreatePrefixesIamOnDbUser` is split into the true (`IAMA:`) and false (`IAM:`) cases. The compat test still asserts only `contains("analyst")`, so it is unchanged.

**`@Order(2)` collision** (057f8cd): the two credential tests are now `@Order(3)` and `@Order(4)`, and `testDeleteCluster` moves to `@Order(5)`.

hectorvent:
Both fixes confirmed, thank you. The prefix now reads `IAMA:` when `AutoCreate` is true and `IAM:` when false, matching the operation's own documentation, and splitting the test into both directions closes the gap that let the inverted mapping pass. Orders are 1 to 5 with delete last.

**(blocking)** It needs a rebase, and the one conflict is worth a warning. `RedshiftTest.testDeleteCluster` collides: main now has it at `@Order(3)` with `throws Exception` and its body wrapped in `withRetry(...)` from #3216, which merged this morning; your side has `@Order(5)` with the older unwrapped body. Neither side is right alone. Taking yours silently reverts #3216; taking main's reinstates the order collision you just fixed. Use main's body and `throws Exception` with your `@Order(5)`. I tried it, and nothing else conflicts once that is done.

**(follow-up)** While you are there: your two new tests call the SDK directly, where everything else in that class now goes through `withRetry`. #3216 added it because this suite flaked on connection-refused.

pgermosen (approved):
Went through the auth logic end to end rather than trusting the fix commits alone. passwordValidatorFor correctly rejects on an absent cluster row and on a wrong master password without ever falling through to PASSTHROUGH, PostgresProtocolHandler terminates on REJECT before backend-login resolution is even reached, and the Data API path checks the same live (non-expired) credential the proxy does. RedshiftCredentialBroker uses SecureRandom, keeps concurrently issued credentials independent, and revocation is wired into both delete and both createCluster rollback branches.

Checked the wire surface against botocore and the live API reference: DurationSeconds 900-3600 default 900 matches both operations exactly, ClusterNotFound and required DbUser match, and the IAM:/IAMA: AutoCreate prefix matches word for word including the sample response.

Also confirmed hector's two blocking items and his non-blocking follow-up are all actually done: the prefix fix, the exact merge resolution he specified for the RedshiftTest.testDeleteCluster conflict with #3216, and both new credential tests now wrapped in withRetry like the rest of the class.

Approving.

PR #3275, opened 2026-09-08
fix(iam): report the real default version of AWS-managed policies

### LINKED ISSUE #3250: [BUG] AWS-managed policies always report DefaultVersionId v1, ignoring AWS's real revision history
## Service

IAM

## AWS API Action

GetPolicyVersion / GetPolicy / ListPolicyVersions (AWS-managed policies)

## Expected behavior

Real AWS reports each managed policy's actual default version number, which reflects how many times AWS has revised that policy's document. For example, `AmazonS3ReadOnlyAccess`'s current default version is `v3` (last revised August 2023), and its earlier `v1`/`v2` documents remain individually retrievable via `GetPolicyVersion`.

## Actual behavior

Floci reports every AWS-managed policy's default version as `v1`, regardless of how many times AWS has actually revised it, and only ever stores the single current document. `AmazonS3ReadOnlyAccess` is a concrete example: floci's `GetPolicy` reports `DefaultVersionId: v1`, matching real AWS's version *numbering scheme* only by coincidence for policies AWS has never revised (e.g. `AdministratorAccess`, genuinely `v1` on real AWS). For a policy AWS has revised, like `AmazonS3ReadOnlyAccess`, floci's reported version number simply does not match what a real AWS account would show, and there is no way to retrieve an older version's document since only the current one is bundled.

This is a pre-existing simplification, not a regression: `IamPolicy.defaultVersionId` defaults to `"v1"` for every policy (customer-managed and AWS-managed alike) in the general `IamPolicy` model, and was left out of scope by #3237 (which fixed the far more serious problem of AWS-managed policies evaluating as unrestricted `Allow: *` on `Resource: *`, and correctly did not try to also solve version-number accuracy in the same change).

## Reproduction

```bash
aws --endpoint-url http://localhost:4566 iam get-policy \
  --policy-arn arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess
# -> DefaultVersionId: "v1"

# Real AWS returns DefaultVersionId: "v3" for the same ARN.
```

## Environment

- Floci version / image tag: main branch, following #3237
- How you're running Floci (Docker / native / `mvn quarkus:dev`): N/A, applies regardless of deployment mode

## Related work

Surfaced while independently verifying #3237's AWS-managed-policy-document fix against the live AWS managed-policy reference pages. #3237 fixed the document *content* (no more permissive fallback); this issue tracks the separate, smaller problem of the reported version *number* not reflecting AWS's real revision history. Fixing this well would mean bundling each managed policy's actual version history (not just its current document) from a source like https://github.com/iann0036/iam-dataset, which already backs the current catalog, and wiring `GetPolicyVersion`/`ListPolicyVersions` to serve the right one per policy.

matheusandre1:
## Summary

Closes #3250

AWS-managed policies were always reported as `DefaultVersionId: v1`, regardless of how many times AWS has revised them. `AmazonS3ReadOnlyAccess` is on `v3` on real AWS (last revised August 2023); Floci only matched real AWS by coincidence for policies that were never revised, such as `AdministratorAccess`.

The catalog now bundles each policy's real default version id and that version's creation date, taken from the same public dataset (iann0036/iam-dataset) that already backs the policy documents. `GetPolicy`, `GetPolicyVersion` and `ListPolicyVersions` serve the bundled document under the version id AWS actually reports, with `CreateDate`/`UpdateDate` matching that version instead of the process start time.

- New resource `iam/managed-policy-versions.json` keyed by policy name (`defaultVersionId`, `createDate`), generated from the dataset's `version` and `updatedate` fields for all 1,566 catalog entries.
- `AwsManagedPolicies` loads it and carries the version on `ManagedPolicyDef`; a missing or malformed entry falls back to `v1` with a log line rather than dropping the policy.
- `IamPolicy` gains a constructor that accepts an arbitrary single default version; `PolicyVersion` gains a constructor with an explicit create date. Customer-managed policies still start at `v1`.
- Docs: new "Version numbers" section in `docs/services/iam.md`, and the stale sentence saying seeded policies use a wildcard document (obsolete since #3237) is replaced.

The dataset only publishes each policy's current document, so superseded versions (`v1`/`v2` of `AmazonS3ReadOnlyAccess`) return `NoSuchEntity`, which is what AWS returns once it prunes a managed policy's history. `ListPolicyVersions` lists only the default version. Bundling the full revision history would require mining the dataset's git history for every policy and is left out of scope.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `aws iam get-policy --policy-arn arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess` returned `DefaultVersionId: v1`, while real AWS returns `v3`. It now returns `v3` with `UpdateDate: 2023-08-10T21:31:39Z`; `AdministratorAccess` still returns `v1` with its real 2015 create date. Response shapes, policy names, paths, ARNs and descriptions are unchanged. AWS-managed policies remain read-only (`CreatePolicyVersion`, `SetDefaultPolicyVersion`, `DeletePolicyVersion` still return `AccessDenied`).

Added coverage in `IamManagedPolicyCatalogTest` and `IamServiceTest`:

- Default version reflects AWS's revision history (`v3` vs `v1`) with the published dates.
- `ListPolicyVersions` returns the single default version with the bundled document, and `GetPolicyVersion` serves it.
- A superseded version id returns `NoSuchEntity` instead of the current document under the wrong id.
- Every catalog entry has a well-formed version id and date.

## Checklist

Focused runs of `IamManagedPolicyCatalogTest`, `IamServiceTest`, `IamManagedPolicyAccountScopeTest`, `IamPolicyEvaluatorTest`, `SsoAdminServiceTest`, `CloudFormationIamAttachmentProvisionerTest` and `CloudFormationManagedPolicyUpdateIntegrationTest` pass. The full Maven suite was not run on this machine.

### DIFF
--- a/docs/services/iam.md
+++ b/docs/services/iam.md
@@ -274,7 +274,23 @@ Floci seeds a catalog of commonly-used AWS managed policies at startup. These ar
 **Other execution roles**
 `AmazonS3ObjectLambdaExecutionRolePolicy` · `CloudWatchLambdaInsightsExecutionRolePolicy` · `CloudWatchLambdaApplicationSignalsExecutionRolePolicy` · `AWSConfigRulesExecutionRole` · `AWSMSKReplicatorExecutionRole` · `AWS-SSM-DiagnosisAutomation-ExecutionRolePolicy` · `AWS-SSM-RemediationAutomation-ExecutionRolePolicy` · `AmazonSageMakerGeospatialExecutionRole` · `AmazonSageMakerCanvasEMRServerlessExecutionRolePolicy` · `SageMakerStudioBedrockFunctionExecutionRolePolicy` · `SageMakerStudioDomainExecutionRolePolicy` · `SageMakerStudioQueryExecutionRolePolicy` · `AmazonDataZoneDomainExecutionRolePolicy` · `AmazonBedrockAgentCoreMemoryBedrockModelInferenceExecutionRolePolicy` · `AWSPartnerCentralSellingResourceSnapshotJobExecutionRolePolicy`
 
-All seeded policies use a permissive wildcard document since Floci does not enforce IAM policy evaluation by default.
+Every catalog entry carries the real policy document of its current default version, generated from the public [iam-dataset](https://github.com/iann0036/iam-dataset), so `GetPolicyVersion` returns the same statements a real account would and enforcement mode evaluates them faithfully.
+
+### Version numbers
+
+AWS revises its managed policies in place, so their default version is rarely `v1`: `AmazonS3ReadOnlyAccess` is on `v3`, `ReadOnlyAccess` far beyond that, while `AdministratorAccess` has never been revised. Floci reports the version id AWS publishes for each policy, the date the policy was first created as `CreateDate`, and the date of its current default version as `UpdateDate`:
+
+```bash
+aws --endpoint-url http://localhost:4566 iam get-policy \
+  --policy-arn arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess
+# -> DefaultVersionId: "v3", CreateDate: 2015-02-06T18:40:00Z, UpdateDate: 2023-08-10T21:31:39Z
+
+aws --endpoint-url http://localhost:4566 iam list-policy-versions \
+  --policy-arn arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess
+# -> a single entry, v3, IsDefaultVersion: true
+```
+
+Only the default version's document is bundled. Requesting a superseded version (`v1` or `v2` of `AmazonS3ReadOnlyAccess`) returns `NoSuchEntity`, the same answer AWS gives once it has pruned a managed policy's history, and `ListPolicyVersions` lists only the default. AWS managed policies remain read-only: `CreatePolicyVersion`, `SetDefaultPolicyVersion` and `DeletePolicyVersion` are rejected with `AccessDenied`.
 
 ## Optional Local Deployer Principal
 
--- a/src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
+++ b/src/main/java/io/github/hectorvent/floci/services/iam/AwsManagedPolicies.java
@@ -10,15 +10,18 @@ import org.jboss.logging.Logger;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
+import java.time.Instant;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /**
  * Catalog of AWS managed policies, loaded from {@code iam/managed-policies.yaml}.
  *
  * <p>Floci resolves {@code arn:aws:iam::aws:policy/*} ARNs against this catalog, so an ARN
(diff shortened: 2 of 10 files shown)


hectorvent:
Thank you, this is a real fidelity gap and taking the version ids from the same dataset that already backs the documents is the right source. Two small things:

**(blocking)** The new javadoc in `AwsManagedPolicies` carries an em dash. The repo forbids them in all content, and the em-dash check only scans new files, so an existing file slips through green.

**(blocking)** The revised version's creation date is also passed as the policy's `CreateDate`, so a policy last revised in 2023 now reports 2023 as its creation date rather than when it was first published. The two dates are separate fields on AWS.

CI is fully green on the rebase, so the shard that was red earlier was the flake it looked like.

matheusandre1:
> Thank you, this is a real fidelity gap and taking the version ids from the same dataset that already backs the documents is the right source. Two small things:
> 
> **(blocking)** The new javadoc in `AwsManagedPolicies` carries an em dash. The repo forbids them in all content, and the em-dash check only scans new files, so an existing file slips through green.
> 
> **(blocking)** The revised version's creation date is also passed as the policy's `CreateDate`, so a policy last revised in 2023 now reports 2023 as its creation date rather than when it was first published. The two dates are separate fields on AWS.
> 
> CI is fully green on the rebase, so the shard that was red earlier was the flake it looked like.

the one I had introduced in the `AwsManagedPolicies` javadoc was already gone in 7586453e. The remaining one on line 24 predates this PR (it came in with #3237), which is why the check stayed green. I removed it anyway since I'm touching the file, along with the only other one in `IamManagedPolicyCatalogTest`. Both files are now em-dash free.

**CreateDate**: agreed, they are separate fields on AWS. This is fixed in 7586453e. The dataset only carries the current version's date, so `createDate` in `managed-policy-versions.json` now comes from the "Creation time" published on each policy's page of the AWS Managed Policy Reference (minute precision; the dataset's exact timestamp is used for policies still on v1, where the two dates coincide). `updateDate` remains the current version's date from the dataset. `IamPolicy` takes the two dates separately, so `AmazonS3ReadOnlyAccess` reports `CreateDate: 2015-02-06T18:40:00Z` and `UpdateDate: 2023-08-10T21:31:39Z`, while `AdministratorAccess` reports the same date for both. One retired policy (`AWSConfigRole`) has no reference page, so it falls back to its version date.

hectorvent:
Both closed, thank you, and you were right to push back on how I framed the second one.

I checked the `CreateDate` split against the data rather than the code path I had guessed at: `managed-policy-versions.json` carries `defaultVersionId`, `createDate` and `updateDate` separately across all 1566 entries, and `AmazonS3ReadOnlyAccess` now reports `CreateDate: 2015-02-06T18:40:00Z` against `UpdateDate: 2023-08-10T21:31:39Z`. `AdministratorAccess` carries the same value for both, and `AWSConfigRole` falls back cleanly. That matches what the model expects, since `Policy.CreateDate` and `Policy.UpdateDate` are separate members.

Taking the creation time from the Managed Policy Reference is a sensible answer to the dataset not carrying it, and I liked that the fallback is written down rather than left implicit.

Good catch on the line 24 em dash predating this PR too. Merging.

hectorvent (approved):
Approving: both asks closed.

PR #3276, opened 2026-09-08
fix(elasticache): honor the requested Port on CreateReplicationGroup

hectorvent:
## Summary

`CreateReplicationGroup` models `Port` as an optional input, but Floci ignored it:
`allocateProxyPort` took no argument and always handed out the next free port from the
base of its range. A caller that pinned a port read back a different one, which
Terraform treats as replacement-forcing and therefore reports as permanent drift.

`allocateProxyPort` now honors a requested port that is free and inside the proxy
range, mirroring `NeptuneService.allocateProxyPort`. Floci multiplexes every group's
proxy onto one host, so two groups genuinely cannot share a port; when the requested
one is taken or out of range it logs why and falls back to the next free port rather
than failing the create. In cluster mode only the first node can take the requested
port, because the group reports its port from that node.

`CreateCacheCluster` also models `Port`, but the memcached path takes its port from
the container handle rather than this allocator, so it is a separate mechanism and is
left for a follow-up.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The incorrect behavior: a pinned `Port` was silently replaced by whatever Floci
allocated, so `aws_elasticache_replication_group` re-planned dirty forever.

botocore models `Port` on `CreateReplicationGroup` as "the port number on which each
member of the replication group accepts connections"
([service-2.json](https://github.com/boto/botocore/blob/develop/botocore/data/elasticache/2015-02-02/service-2.json)).
Verified end to end with the Terraform AWS provider through the compatibility suite.

Both compatibility fixtures previously pinned `6379`, which is the bottom of the proxy
range and the port Floci would have allocated anyway, so the assertion passed by
coincidence rather than by behavior. They now pin `6395`, which fails without this fix.

## Checklist

- [x] `./mvnw test` passes locally (129 ElastiCache tests)

### DIFF
--- a/compatibility-tests/compat-opentofu/main.tf
+++ b/compatibility-tests/compat-opentofu/main.tf
@@ -287,7 +287,11 @@ resource "aws_elasticache_replication_group" "compat" {
   engine               = "redis"
   node_type            = "cache.t3.micro"
   num_cache_clusters   = 1
-  port                 = 6379
+  # Deliberately not 6379: that is the bottom of floci's ElastiCache proxy range and the
+  # port it would hand out anyway, so pinning it asserts nothing. A non-default port is
+  # what actually exercises CreateReplicationGroup's Port input, and a mismatch shows up
+  # as permanent drift because Terraform treats the port as replacement-forcing.
+  port                 = 6395
 }
 
 # ── Firehose Delivery Stream ────────────────────────────────────────────────
--- a/compatibility-tests/compat-terraform/main.tf
+++ b/compatibility-tests/compat-terraform/main.tf
@@ -287,7 +287,11 @@ resource "aws_elasticache_replication_group" "compat" {
   engine               = "redis"
   node_type            = "cache.t3.micro"
   num_cache_clusters   = 1
-  port                 = 6379
+  # Deliberately not 6379: that is the bottom of floci's ElastiCache proxy range and the
+  # port it would hand out anyway, so pinning it asserts nothing. A non-default port is
+  # what actually exercises CreateReplicationGroup's Port input, and a mismatch shows up
+  # as permanent drift because Terraform treats the port as replacement-forcing.
+  port                 = 6395
 }
 
 # -- Firehose Delivery Stream --------------------------------------------------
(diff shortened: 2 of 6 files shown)


hectorvent:
Good catch, and fixed in 8a3134e19.

The fallback was worse than it looked. It could only ever fire for a caller who did pin a port, because one who does not care sends no `Port` at all and never reaches that branch, so the case it was written to protect ("callers that do not care which port they get", per my own comment) could not reach it. All it did was hand a pinned caller a different port and reproduce the replacement-forcing drift this PR exists to remove, in a narrower case.

An explicit port is now either honored or refused with `InvalidParameterValue`, the sender fault `CreateReplicationGroup` declares, with the port and the reason in the message so the caller can pick another. Unpinned creates still scan the range as before, and the restore path for persisted groups is untouched since it asks for no particular port. The two tests that pinned the old fallback are inverted rather than dropped, and one more holds the unpinned path.

Two neighbours have the same shape and are left for their own changes: `NeptuneService.allocateProxyPort` substitutes identically, and `RdsService` ignores a requested port altogether even though `Port` is modeled on `CreateDBInstance`.

pgermosen:
@hectorvent now that this is in, `NeptuneService.allocateProxyPort` has the exact same issue this PR just fixed for ElastiCache: when a requested port is taken or out of range it logs and silently falls back to the next free one instead of refusing with `InvalidParameterValue`, so a caller pinning `Port` on `CreateDBCluster` still hits the same replacement-forcing drift you just closed off here. Since the fix pattern is fresh, that one seems like the natural next pickup.

`RdsService` is further behind — `allocateProxyPort()` there doesn't take a port argument at all, so a requested `Port` on `CreateDBInstance`/`CreateDBCluster` is dropped entirely rather than honored or refused.

And as you noted, the memcached `CreateCacheCluster` path ignores `Port` too, though that one's a different mechanism (container-assigned rather than this allocator) so it probably deserves its own design pass rather than a quick port of this fix.

Worth splitting into follow-up issues/PRs whenever you get to it.

PR #3278, opened 2026-09-09
fix(guardduty): scope organization state across accounts

KenkoGeek:
## Summary

Fix GuardDuty account scoping so detectors remain account-local while organization delegated-administrator state is visible across account partitions, matching delegated administration flows.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

GuardDuty detector state must be scoped to the caller account and Region. Organization delegated-administrator state must remain visible when the delegated administrator calls member-management APIs from its own account context. Previously Floci could report a delegated admin as enabled from the management account while `CreateMembers` from the delegated-admin account could not see that organization state and left members in the wrong relationship state.

Verified with AWS SDK for Java v2 `GuardDutyClient` using separate management and delegated-administrator account credentials against the branch-local Floci endpoint. The SDK regression proves each account sees only its own detector, while delegated-administrator state remains visible across account partitions and `CreateMembers` yields an associated member with relationship status `Enabled`.

## Checklist

Focused validation: `./mvnw test -Dtest=GuardDutyServiceTest,GuardDutyControllerIntegrationTest` — 29 tests, 0 failures, 0 errors.

AWS SDK compatibility validation: `FLOCI_ENDPOINT=http://127.0.0.1:4696 ./mvnw -f compatibility-tests/sdk-test-java/pom.xml -Dtest=GuardDutyTest test` — 4 tests, 0 failures, 0 errors.

`make docs-check` and `git diff --check` pass.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/main/java/com/floci/test/TestFixtures.java
+++ b/compatibility-tests/sdk-test-java/src/main/java/com/floci/test/TestFixtures.java
@@ -887,10 +887,14 @@ public final class TestFixtures {
     }
 
     public static GuardDutyClient guardDutyClient() {
+        return guardDutyClient("test");
+    }
+
+    public static GuardDutyClient guardDutyClient(String accountId) {
         return GuardDutyClient.builder()
                 .endpointOverride(ENDPOINT)
                 .region(REGION)
-                .credentialsProvider(CREDENTIALS)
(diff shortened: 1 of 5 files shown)


hectorvent (approved):
The split is right: detectors account-local, organization delegated-administrator state readable across partitions, which is what the delegated admin needs when it calls member management from its own account. Good to see the SDK-driven compat case alongside the service test. Approving.

PR #3279, opened 2026-09-09
fix(ram): emulate organization sharing side effects

KenkoGeek:
## Summary

Fix AWS RAM `EnableSharingWithAwsOrganization` so Floci also enables Organizations trusted access for `ram.amazonaws.com` and creates the expected IAM service-linked role `AWSServiceRoleForResourceAccessManager`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

AWS RAM organization sharing has side effects beyond returning success: trusted access is enabled in AWS Organizations and the RAM service-linked role is created. Previously Floci returned success without those effects, which forced consumers to pre-seed Organizations/IAM state.

Verified with AWS SDK for Java v2 `RamClient`, `OrganizationsClient`, and `IamClient` against the branch-local Floci endpoint. The SDK regression asserts that `EnableSharingWithAwsOrganization` returns success, `ram.amazonaws.com` appears in `ListAWSServiceAccessForOrganization`, and `AWSServiceRoleForResourceAccessManager` is retrievable through IAM.

## Checklist

Focused validation: `./mvnw test -Dtest=RamIntegrationTest,IamServiceTest` — 124 tests, 0 failures, 0 errors.

AWS SDK compatibility validation: `FLOCI_ENDPOINT=http://127.0.0.1:4695 ./mvnw -f compatibility-tests/sdk-test-java/pom.xml -Dtest=RamOrganizationSharingTest test` — 1 test, 0 failures, 0 errors.

`make docs-check` and `git diff --check` pass.

### DIFF
--- a/compatibility-tests/sdk-test-java/pom.xml
+++ b/compatibility-tests/sdk-test-java/pom.xml
@@ -341,6 +341,11 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>guardduty</artifactId>
         </dependency>
+        <!-- AWS Resource Access Manager client -->
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>ram</artifactId>
+        </dependency>
         <!-- Secrets Manager client -->
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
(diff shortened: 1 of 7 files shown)


hectorvent (approved):
Emulating the side effects through the real Organizations and IAM services rather than a local flag is what makes this worth having: a caller can enable sharing and then see `ram.amazonaws.com` in the enabled service principals and read the service-linked role back, which is the sequence real tooling follows. The `findRole` guard keeps a repeat call idempotent, and the new SDK client sits in the compatibility pom where its siblings are. Approving.

PR #3282, opened 2026-09-09
ci(compat): run the Terraform and OpenTofu bats files concurrently

hectorvent:
## Summary

`compat-terraform` is the slowest leg of the compatibility workflow and the job that ends it.
Across recent runs its "Run tests" step is flat at 13m18s to 13m32s, while every other leg is
finished several minutes earlier, so it alone sets the wall clock of the workflow on every PR
that touches `src/**`.

The six `test/*.bats` files ran as a single `bats test/` process, so they executed one after
another even though each is an independent root module with its own state and identifiers.
Their cost is dominated by the AWS provider's own waiters rather than by Floci: Neptune and
ElastiCache sleep 30s before the first poll on create and delete, and RDS sleeps 60s with
`ContinuousTargetOccurence: 3`. Nothing can shorten those, only overlap them.

Each file now runs in its own process, reaped with `wait`, writing one `junit-<file>.xml`.
bats' own `--jobs` needs GNU parallel or rush, neither of which is in the image, so the files
are forked directly. Output is collected per file and replayed in a fixed order, because six
interleaved streams are unreadable. `BATS_PARALLEL_FILES=0` restores sequential execution for
isolating a failure, and both `run.sh` wrappers and the justfile recipes delegate to the same
runner so a local run and a CI run execute identical logic.

The provider plugin cache is warmed once before forking. Every fixture deletes its lock file
in `setup_file` and the AWS provider unpacks to 770 MB, so six concurrent inits would
otherwise pull roughly 4.6 GB. The warm-up is sequential because that cache is safe for
concurrent readers but not writers, and a failure there is logged and tolerated rather than
fatal.

Also adds the `.dockerignore` both suites were missing. Each directory can hold a local
`.terraform/` of several hundred megabytes plus stray state, none of which exists in a CI
checkout, so image builds were copying about 1.4 GB of host state and behaving differently
from CI.

Measured locally on the same image, all 84 tests passing either way: **sequential 1130s,
concurrent 260s**.

No workflow YAML changes are needed: `compatibility.yml` and `.github/ci/compat-timing-report.py`
already glob `test-results/*.xml`, so per-file reports are picked up as they are.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

N/A. This changes only how the compatibility suites are executed, not any emulated surface.
The suites themselves are unchanged apart from one comment on the ElastiCache port pin,
noting it must stay above the block the cluster-mode fixture allocates now that the files run
concurrently.

## Checklist

- [x] `./mvnw test` passes locally (no Java changed, so the suite is unaffected)
- [x] New or updated integration test added (the suites are the test: 84 tests green in both
      modes, and a run with `BATS_PARALLEL_FILES=0` confirms the sequential fallback)

### DIFF
--- a/compatibility-tests/README.md
+++ b/compatibility-tests/README.md
@@ -114,8 +114,13 @@ Bats-based suites keep their normal console output and also write JUnit XML repo
 
 - `sdk-test-awscli/test-results/junit.xml`
 - `compat-cdk/test-results/junit.xml`
-- `compat-terraform/test-results/junit.xml`
-- `compat-opentofu/test-results/junit.xml`
+- `compat-terraform/test-results/junit-<bats file>.xml`
+- `compat-opentofu/test-results/junit-<bats file>.xml`
+
+The Terraform and OpenTofu suites run each `test/*.bats` file in its own process
+concurrently, so they write one report per file rather than a single `junit.xml`.
+Every consumer globs the directory, so nothing downstream changes. Set
+`BATS_PARALLEL_FILES=0` to run the files one at a time when isolating a failure.
 
 ## Configuration
 
--- a/compatibility-tests/compat-opentofu/.dockerignore
+++ b/compatibility-tests/compat-opentofu/.dockerignore
@@ -1,0 +1,11 @@
+# Keep host-side Terraform working state out of the image. CI builds from a fresh
+# checkout where none of this exists (.terraform/ and .terraform.lock.hcl are
+# gitignored), so copying a local run's leftovers in would make the image behave
+# differently from CI and quietly skew any timing comparison.
+.terraform
+.terraform.lock.hcl
+*.tfstate
+*.tfstate.*
+errored.tfstate
+test-results
+large-object.bin
--- a/compatibility-tests/compat-opentofu/main.tf
+++ b/compatibility-tests/compat-opentofu/main.tf
@@ -291,6 +291,9 @@ resource "aws_elasticache_replication_group" "compat" {
   # port it would hand out anyway, so pinning it asserts nothing. A non-default port is
   # what actually exercises CreateReplicationGroup's Port input, and a mismatch shows up
   # as permanent drift because Terraform treats the port as replacement-forcing.
+  #
+  # It must also stay above the block the cluster-mode fixture allocates: the bats files
+  # run concurrently and that fixture takes one port per node from the base of the range.
   port                 = 6395
 }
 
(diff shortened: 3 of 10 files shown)


hectorvent:
Thanks, fixed in bb9a62a5f.

You were right that it only bites a reused workspace, and it was already happening: both suites still held the single `junit.xml` the previous sequential runner wrote, and the new `junit-<file>.xml` names never overwrite it. So a local run was globbing months-old results alongside the current ones, including in the bench script the timing numbers in this PR came from. CI was never affected, since it starts from a fresh checkout with an empty `test-results`.

Took your suggested edit. The `junit*.xml` glob is the right shape because it catches both the old single-file name and a report left behind by a bats file that was renamed or removed.

For the record, the CI effect of the PR itself, measured against four same-day runs on other branches so runner variance is controlled: the `compat-terraform` test step went from 13m13s-13m34s to 5m38s, the workflow span from about 21m20s to 14m13s, and the leg stopped being the job that ends the workflow. 84 tests, 0 failures, unchanged.

hectorvent:
Right again, and the first fix was worse than you flagged. Fixed in 9511c1a34.

`docker/run-docker-tests.sh` mounts a single `test-results` for all eight suites, so the `junit*.xml` glob did not just wipe `compat-terraform`'s reports when `compat-opentofu` started. It also deleted the `junit.xml` that `sdk-test-awscli` and `compat-cdk` had written earlier in the same loop. Two unrelated suites, not one.

The deletion now names the reports the run is about to write, taken from the bats files already enumerated above it, so it cannot reach a filename this suite does not own. I verified it against a seeded shared directory: another suite's `junit.xml` and `junit-terraform.xml` both survive while this suite's own stale report is refreshed. That check fails on the previous commit.

One case is deliberately given up: a report whose bats file was renamed or removed is no longer cleared, since nothing in the current run names it. A lingering stale report is cosmetic, and deleting another suite's results is not, so that seems the right way round.

Worth separating out, since it predates this PR: `sdk-test-awscli` and `compat-cdk` already overwrite each other's `junit.xml` in that shared directory, cdk running second. The real repair is per-suite result directories in `run-docker-tests.sh`, the way the compat-docker skill already does it. I have left that for its own change rather than widening this one.

pgermosen (approved):
Nice speedup, and I like that you caught your own cleanup bug before I even got to it. Traced the report-globbing issue against docker/run-docker-tests.sh myself to be sure: it really does loop all eight suites over one shared mounted test-results directory, so the original junit*.xml cleanup would have wiped compat-awscli's and compat-cdk's reports the moment compat-terraform ran. Naming the exact files instead is the right fix.

Went through the concurrency mechanics in run-bats-in-container.sh by hand since I can't run it here: the per-file work directories avoid the report.xml collision, and the exit-status handling through the .status files is sound for backgrounded jobs. Only thing I couldn't verify myself is whether concurrent terraform init reads against the warmed plugin cache are as safe as assumed, but that's exactly what the 84/84-passing measurement is there to cover.

Good to merge.

PR #3283, opened 2026-09-09
fix(ec2,ssm): explain missing IMDS registration for SSM-only managed instances

### LINKED ISSUE #3280: [see-something] docs/ec2: SSM container-backed managed instances have no path to a working IMDS endpoint
## What happens

\`Ec2MetadataServer\` (port 9169, proxied to 169.254.169.254 inside a container) only learns about an instance via \`Ec2ContainerManager.registerContainer(containerIp, instanceId, instance)\`, called exactly once, right after \`Ec2ContainerManager\` launches a container through EC2 \`RunInstances\`. That same call site is also the only place that installs the IMDS proxy itself (\`metadataProxyInstallCommand\`/\`metadataProxyStartCommand\`: adds \`169.254.169.254/32\` to \`lo\`, starts a \`socat\` forwarder to the Floci host's IMDS port).

SSM's container-backed managed instance feature (used so \`SendCommand\` actually executes against a container, per \`SsmSendCommandIntegrationTest#sendCommandDirectlyExecutesForContainerBackedInstance\`) registers a container as a managed instance through a completely separate path that never touches \`Ec2ContainerManager\` or \`Ec2MetadataServer\`.

Net effect: a container registered as an SSM managed instance that was **not** also launched via EC2 \`RunInstances\` has no IMDS proxy installed and no \`Instance\` record in \`containerIpToInstance\`/\`tokenToInstance\`. Any curl to \`169.254.169.254\` or directly to the IMDS port from such a container fails outright (proxy not installed), and even a manually-replicated proxy pointed at Floci's IMDS port gets a real, well-formed \`404 Instance not found\` response, since the lookup keys on the source container IP and nothing ever registered it.

## Why it matters

A realistic simulation (EC2 host + SSM Automation/State Manager association driving work on it + IMDS-sourced role credentials, e.g. for a nightly patch/inventory job writing to S3) reasonably assumes "the instance I registered with SSM" and "the instance IMDS answers for" are the same thing, the way they are on real AWS. Discovered while standing up a proof-of-concept against a vanilla Floci build (fork commit 90399ae38, upstream 9d4e71f8a) for exactly this scenario.

## Suggested fix (documentation, and ideally the underlying gap)

- Document, in \`docs/services/ec2.md\`'s IMDS section and wherever SSM managed-instance registration is documented, that the two are independent: to get a container that has both real IMDS access and can run \`SendCommand\`, launch it via EC2 \`RunInstances\` first (which wires the proxy and instance record), then register that same container's SSM agent as a managed instance, rather than registering an arbitrary/pre-existing container directly.
- Longer-term, it'd be more AWS-faithful if registering a container as an SSM managed instance also registered it with \`Ec2MetadataServer\`, or at least logged/returned a clear signal (rather than a bare \`Instance not found\`) explaining that this container has no EC2 instance record backing it.

## Environment

- Fork: sam-fakhreddine/floci, main at 90399ae38 (merges upstream/main 9d4e71f8a)
- Reproduced with \`docker/Dockerfile\`, JVM mode, \`FLOCI_STORAGE_MODE=hybrid\`, default port 4566 + published 9169

matheusandre1:
## Summary

Fixes: #3280 

Closes #3280

`Ec2MetadataServer` only learns about a container when `Ec2ContainerManager` launches it through EC2 `RunInstances`; that is also the only place the link-local `169.254.169.254` proxy is installed. Registering a container's SSM agent as a managed instance (`UpdateInstanceInformation`) goes through a separate path that never touches EC2 or IMDS. A container that was only registered with SSM therefore had no working IMDS endpoint and got a bare `404 Instance not found` with no hint of why.

This PR documents that the two registrations are independent and makes the gap visible at runtime:

- `docs/services/ec2.md`: new "IMDS and SSM managed instances" subsection under the IMDS section explaining the dependency and the correct order (launch with `RunInstances`, then register that container's SSM agent).
- `docs/services/ssm.md`: note in "Run Command Execution" and in the `UpdateInstanceInformation` row saying it does not create an EC2 instance nor register the container with IMDS. Both pages cross-link.
- `Ec2MetadataServer`: the 404 for an unregistered source IP now says which IP was seen, that IMDS only serves containers launched via `RunInstances`, and that SSM registration does not register with IMDS. Same text goes to the warn log. An IMDSv2 token request from an unregistered IP logs at debug.
- `SsmDirectCommandExecutor`: extracted `isContainerBacked(instanceId)`, reused by `supports`.
- `SsmCommandService.updateInstanceInformation`: on the first registration of an instance that is not backed by a running Floci EC2 container, logs a warning that `SendCommand` will use the agent polling flow and IMDS will not answer for it. Only on first registration so periodic agent pings do not spam the log.

Registering an SSM-only container with IMDS automatically was not done: there is no EC2 `Instance` record to serve and no Docker container id to install the proxy into.

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [X] Docs / chore

## AWS Compatibility

On AWS, "the instance I registered with SSM" and "the instance IMDS answers for" are the same thing. In Floci they are only the same when the container was launched through EC2 `RunInstances`. Previously a container registered only through SSM got a well-formed `404 Instance not found` from IMDS with no explanation. The status code is unchanged; only the body and logs gained an explanation, so SDK credential-chain behavior is unaffected.

## Checklist

### DIFF
--- a/docs/services/ec2.md
+++ b/docs/services/ec2.md
@@ -138,6 +138,14 @@ curl -s -H "x-aws-ec2-metadata-token: $TOKEN" \
 
 IAM credentials are served when the instance has an `IamInstanceProfile.Arn` set at launch. The container can then call other Floci services with full SigV4 validation using the standard AWS SDK credential chain.
 
+### IMDS and SSM managed instances
+
+IMDS only knows about containers that Floci itself launched through `RunInstances`. That launch is what installs the link-local `169.254.169.254` proxy inside the container and maps the container's IP to its instance record.
+
+Registering a container with SSM is independent of this: an SSM agent that calls `UpdateInstanceInformation` becomes a managed instance, but that does not create an EC2 instance record, does not install the IMDS proxy, and does not register the container with IMDS. From such a container, `curl http://169.254.169.254/...` fails outright (no proxy), and a request sent straight to the host IMDS port returns `404` with a message explaining that no EC2 instance is registered for the source IP. Floci also logs a warning when an SSM agent registers from a container that is not backed by a Floci EC2 instance.
+
+To get a container that both answers IMDS (including instance profile credentials) and executes `SendCommand` directly, launch it with `RunInstances` first, then register that same container's SSM agent as a managed instance. Do not register an arbitrary or pre-existing container directly with SSM and expect IMDS to work. See [SSM](ssm.md#run-command-execution) for the SendCommand side of this.
+
 ## Default Resources
 
 Floci seeds the following resources on first use in each region so Terraform, the AWS CLI, and SDK clients work out of the box without any setup:
--- a/docs/services/ssm.md
+++ b/docs/services/ssm.md
@@ -28,7 +28,7 @@
 
 | Action | Description |
 |---|---|
-| `UpdateInstanceInformation` | Register or update an SSM agent record for an instance |
+| `UpdateInstanceInformation` | Register or update an SSM agent record for an instance. Does not create an EC2 instance or register the container with IMDS (see [Run Command Execution](#run-command-execution)) |
 | `DescribeInstanceInformation` | List registered SSM managed instances |
 | `SendCommand` | Create command invocations for target instances |
 | `GetCommandInvocation` | Return a command invocation result |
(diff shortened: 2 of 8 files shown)


hectorvent (approved):
This is the right shape for #3280, and the model backs the call. SSM's `InstanceInformation` frames managed instances as servers or VMs registered through an activation, `ResourceType` splits EC2 instances from managed instances, and the id pattern admits both `i-` and `mi-`. A hybrid activation node has no EC2 IMDS on real AWS either, so explaining the gap is more accurate than installing the proxy on SSM registration.

The `isContainerBacked` extraction is faithful, and gating the warning on first registration keeps it off the heartbeat path, which the `times(1)` test pins directly. Both doc cross-links resolve.

One nit: the checklist ticks an integration test, but all three new test files are unit tests. Unit level is right for this change, so just untick it.

Approving.

PR #3285, opened 2026-09-09
refactor(cloudformation): move the Batch types into BatchCfnProvisioner

hectorvent:
## Summary

Continues dismantling the legacy `CloudFormationResourceProvisioner` switch: `AWS::Batch::ComputeEnvironment`, `JobQueue` and `JobDefinition` move into a new `BatchCfnProvisioner`. Monolith 6,457 to 6,292 lines, legacy types 47 to 44.

The move comes first and is behaviour preserving, then one commit per defect, following the shape of #3115, #3164 and #3223. Reading these arms against `BatchService` and the registry schemas turned up three real defects:

1. **Nothing was ever deleted.** There was no `AWS::Batch` arm in any of the monolith's four delete paths, so a stack delete left the compute environment, job queue and job definition behind, and the stack still reported DELETE_COMPLETE. Each delete now disables first, matching AWS: DeleteJobQueue refuses an `ENABLED` queue and DeleteComputeEnvironment refuses one that is not `DISABLED` or is still attached to a queue. Deliberately **not** `CfnDeletes.safeDelete`: every `BatchService` failure carries the same `ClientException` code, so tolerating it would be the catch-all AGENTS.md rules out and would swallow a genuine "still associated with a job queue". Already-gone is handled by looking first, which also makes a repeat delete idempotent.

2. **Updates were broken both ways.** `provision` re-runs on every UpdateStack. A named compute environment or job queue called create again and both services reject a duplicate name, so the stack update failed. An unnamed one generated a fresh random name each time, creating a second entity and orphaning the first. In-place updates now go through `UpdateComputeEnvironment` / `UpdateJobQueue` and keep the prior physical id. Note `ctx.stablePhysicalName` and `ctx.reusesPriorEntity` do **not** fit these types: both fall back to the physical id, which for Batch is an ARN rather than the name, as their javadoc warns. The prior name is read from the recorded attribute the way `SqsCfnProvisioner` does, with an ARN-derived fallback for stacks created before that attribute existed.

3. **Required properties were optional.** `JobQueue.Priority` was defaulted to 1 and `Type` was optional on the other two, with `JobDefinition` defaulting to `container`, so a template missing one applied cleanly here and failed on AWS. All three now raise `ValidationError`.

Also drops the `Revision` attribute from `AWS::Batch::JobDefinition`: it is not a property of that type in the registry schema and the legacy spec gives the type no attributes at all, so `Fn::GetAtt` on it fails on AWS. The physical id stays the ARN despite the schema's `primaryIdentifier` naming `JobDefinitionName`, because the legacy spec and the documented `Ref` both point at the ARN and the evidence does not agree with itself.

`CreateCacheCluster`-style follow-up left out on purpose: `CreateCacheCluster` also models `Port`, and the memcached path takes its port from the container handle rather than this allocator, so it is a different mechanism.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Incorrect behaviour fixed: Batch resources survived stack deletion, stack updates either failed or silently duplicated resources, and three schema-required properties were accepted as absent.

Sourced from the CloudFormation registry schemas in `local/aws/cfn-resource-schemas` (`required`, `readOnlyProperties`, `primaryIdentifier` for each of the three types) and the AWS Batch API semantics for disable-before-delete and for `RegisterJobDefinition` recording a new revision.

## Checklist

- [x] `./mvnw test` passes locally (1,598 CloudFormation and Batch tests, only the pre-existing `AWS::CertificateManager::Certificate Id` schema-coverage row)
- [x] New or updated integration test added (`BatchCfnProvisionerTest` 19 cases; `CloudFormationBatchIntegrationTest` extended so an update keeps both physical ids and a delete leaves neither entity, both of which fail without these commits)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -3,7 +3,6 @@ package io.github.hectorvent.floci.services.cloudformation;
 import io.github.hectorvent.floci.core.common.AwsArnUtils;
 import io.github.hectorvent.floci.core.common.AwsNamespaces;
 import io.github.hectorvent.floci.core.common.XmlBuilder;
-import io.github.hectorvent.floci.services.batch.BatchService;
 import io.github.hectorvent.floci.services.cloudfront.CloudFrontService;
 import io.github.hectorvent.floci.services.cloudfront.model.CacheBehavior;
 import io.github.hectorvent.floci.services.cloudfront.model.DefaultCacheBehavior;
@@ -194,9 +193,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::ApiGatewayV2::Stage",
             "AWS::AutoScaling::AutoScalingGroup",
             "AWS::AutoScaling::LaunchConfiguration",
-            "AWS::Batch::ComputeEnvironment",
-            "AWS::Batch::JobDefinition",
-            "AWS::Batch::JobQueue",
             "AWS::CloudFormation::CustomResource",
             "AWS::CloudFront::Distribution",
             "AWS::DynamoDB::GlobalTable",
@@ -246,7 +242,6 @@ public class CloudFormationResourceProvisioner {
     private final CustomResourceResponseStore customResourceResponseStore;
     private final ContainerReachableEndpoint reachableEndpoint;
     private final StepFunctionsService stepFunctionsService;
-    private final BatchService batchService;
     private final Ec2Service ec2Service;
     private final RdsService rdsService;
     private final EksService eksService;
@@ -275,7 +270,6 @@ public class CloudFormationResourceProvisioner {
                                              CustomResourceResponseStore customResourceResponseStore,
                                              ContainerReachableEndpoint reachableEndpoint,
                                              StepFunctionsService stepFunctionsService,
-                                             BatchService batchService,
                                              Ec2Service ec2Service,
                                              RdsService rdsService,
                                              EksService eksService,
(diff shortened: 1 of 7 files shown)


pgermosen (commented):
The three fixes themselves are well reasoned and each traced against real Batch API behavior, but Greptile's two findings are real and worth addressing before this merges. Renaming a ComputeEnvironment or JobQueue creates the replacement but never records the old one for cleanup, no ReplacementCleanup usage and no hasReplacementUpdate override, so the engine never learns a replacement happened and the old entity leaks in Batch forever. That's the exact pattern already solved correctly in CloudFront's provisioner a few PRs back in this same series, worth reusing rather than re-deriving.

The missing rollbackUpdate is a more acceptable gap, but only when it's disclosed the way CloudFront's provisioner does it, a line in the class javadoc and the PR body comparing it to Cognito's known limitation. Worth adding that here too rather than leaving it silent.

Smaller one: Integer.parseInt(priority) is unguarded in both the create and update paths, pre-existing rather than something this PR introduced, but since this PR's whole point in the third commit is turning silently-wrong inputs into ValidationError, a non-numeric Priority throwing an uncaught NumberFormatException is right next to work already being done here.

Also worth a rebase when you get to it, conflicts with the Events PR that just merged, but I checked and it's trivial, just two competing import insertions at the same unsorted spot.

pgermosen (approved):
This addresses everything from the last round well. The replacement cleanup is wired through `ReplacementCleanup` the same way CloudFront does it, and `rollbackUpdate` actually restores the prior state now instead of just disclosing the gap, which is better than what I asked for. The one thing it genuinely can't fix, a rollback not being able to strip a `serviceRole` or `computeResources` key a failed update added, is disclosed the same way Cognito's known limitation is, with a test pinning it so it can't regress silently.

Checked the Priority fix against the Batch schema too (integer, 0-1000), and it's applied consistently on both the create and update paths now. Confirmed `DeleteComputeEnvironment`/`DeleteJobQueue`/`DeregisterJobDefinition` really do only define `ClientException`/`ServerException` with no not-found code, so the look-first-then-delete approach instead of `safeDelete` is the right call, not a workaround.

Rebase concern is moot now, it merges clean against current main. Good to go.

PR #3286, opened 2026-09-09
refactor(cloudformation): move the Events types into EventsCfnProvisioner

hectorvent:
## Summary

Continues dismantling the legacy `CloudFormationResourceProvisioner` switch: `AWS::Events::Rule`, `EventBus` and `EventBusPolicy` move into a new `EventsCfnProvisioner`. Monolith 6,457 to 5,954 lines, legacy types 47 to 44.

The block was already contiguous and self-contained, 18 methods and five constants with nothing referenced from outside, and `EventBridgeService` was the monolith's only user, so the field, its constructor parameter and both imports go too. The event bus arm in particular already carried real logic that moves unchanged: rollback ownership, the created-time check that refuses to delete a bus recreated out of band, managed tag and policy tracking, and the rejection of an unsupported mutable update.

**Structural difference from the other batches.** `Rule` and `EventBus` were listed in `DELETE_NEEDS_STACK_RESOURCE`, because deleting them needs stored attributes rather than the physical id alone: a rule remembers the custom bus it lives on, and a bus remembers the creation time proving the stack still owns it. That logic moves into the provisioner's `delete(StackResource, String)` override and both entries leave the set, which is what `CfnDeletePrecedenceTest` gates. `EventBusPolicy` keeps the id-only form, its composite `busName|statementId` being enough.

Then one commit per defect, following #3115, #3164 and #3223:

1. **A rule lost its identity and its targets across an update.** An unnamed rule regenerated its physical name on every UpdateStack, creating a second rule and leaving the first behind, still matching events and still delivering. `Name` is createOnly and a rule's physical id is its name, so `ctx.stablePhysicalName` applies directly here, unlike the Batch types whose id is an ARN. Separately, `PutTargets` upserts by id and removes nothing, so a target dropped from the template stayed on the rule and kept receiving events indefinitely; the desired set is now reconciled after the put, which also covers a template that declares no `Targets` at all.

2. **A failed policy removal was hidden.** Deleting an `AWS::Events::EventBusPolicy` wrapped the whole removal in a bare `catch (Exception)` that logged at debug and returned, so the statement stayed on the bus while the stack reported DELETE_COMPLETE. `RemovePermission` distinguishes its failures, unlike the Batch services, raising `ResourceNotFoundException` both when the bus is gone and when the statement is not on it, so the removal now goes through the shared `CfnDeletes.safeDelete` tolerating only that code and anything else propagates.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Incorrect behaviour fixed: an unnamed `AWS::Events::Rule` was duplicated and orphaned on every stack update, targets removed from a template kept receiving events, and a failed event bus policy removal was reported as a successful stack delete.

## Checklist

- [x] `./mvnw test` passes locally (1,854 CloudFormation and EventBridge tests, only the pre-existing `AWS::CertificateManager::Certificate Id` schema-coverage row)
- [x] New or updated integration test added (new `EventsCfnProvisionerTest` with 7 cases; the four existing Events test classes needed no changes and their 40 tests now exercise the new provisioner through `CfnProvisionerFixture`)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -20,9 +20,7 @@ import io.github.hectorvent.floci.services.cloudformation.provisioners.CfnResour
 import io.github.hectorvent.floci.services.cloudformation.provisioners.Ec2SecurityGroupRuleCfnProvisioner;
 import io.github.hectorvent.floci.services.cloudformation.provisioners.UpdateCleanupResult;
 import io.github.hectorvent.floci.services.dynamodb.DynamoDbService;
-import io.github.hectorvent.floci.services.eventbridge.EventBridgeService;
 import io.github.hectorvent.floci.services.eventbridge.model.BatchParameters;
-import io.github.hectorvent.floci.services.eventbridge.model.EventBus;
 import io.github.hectorvent.floci.services.eventbridge.model.InputTransformer;
 import io.github.hectorvent.floci.services.eventbridge.model.RuleState;
 import io.github.hectorvent.floci.services.eventbridge.model.SqsParameters;
@@ -126,13 +124,6 @@ public class CloudFormationResourceProvisioner {
     private static final String INLINE_CLEANUP_GROUP_TARGETS_ATTR = "__FlociInlineCleanupGroupTargets";
     private static final String SFN_NAME_MODE_ATTR = "FlociStepFunctionsNameMode";
     static final String SFN_UPDATE_SNAPSHOT_ATTR = "__FlociStepFunctionsUpdateSnapshot";
-    private static final String EVENT_BUS_CREATED_TIME_ATTR = "FlociEventBusCreatedTime";
-    private static final String EVENT_BUS_MANAGED_TAG_KEYS_ATTR = "FlociEventBusManagedTagKeys";
-    private static final String EVENT_BUS_MANAGED_POLICY_ATTR = "FlociEventBusManagedPolicy";
-    private static final Set<String> EVENT_BUS_SUPPORTED_PROPERTIES =
-            Set.of("Name", "Description", "Tags", "Policy");
-    private static final Pattern EVENT_BUS_TAG_PATTERN =
-            Pattern.compile("[\\p{L}\\p{N}\\p{Z}_.:/=+\\-@]*");
     private static final String NAME_MODE_EXPLICIT = "explicit";
     private static final String NAME_MODE_GENERATED = "generated";
     private static final int GENERATED_NAME_SUFFIX_LENGTH = 12;
@@ -177,8 +168,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::ApiGatewayV2::Authorizer",
             "AWS::CloudFormation::CustomResource",
             "AWS::EKS::Nodegroup",
-            "AWS::Events::EventBus",
-            "AWS::Events::Rule",
             "AWS::IAM::ManagedPolicy",
             "AWS::IAM::Policy",
             "AWS::SecretsManager::SecretTargetAttachment",
@@ -210,9 +199,6 @@ public class CloudFormationResourceProvisioner {
             "AWS::EC2::SecurityGroup",
             "AWS::EKS::Cluster",
             "AWS::EKS::Nodegroup",
-            "AWS::Events::EventBus",
-            "AWS::Events::EventBusPolicy",
-            "AWS::Events::Rule",
             "AWS::IAM::AccessKey",
             "AWS::IAM::InstanceProfile",
             "AWS::IAM::ManagedPolicy",
@@ -247,7 +233,6 @@ public class CloudFormationResourceProvisioner {
     private final LambdaService lambdaService;
     private final IamService iamService;
     private final SecretsManagerService secretsManagerService;
-    private final EventBridgeService eventBridgeService;
     private final ApiGatewayService apiGatewayService;
     private final ApiGatewayV2Service apiGatewayV2Service;
     private final LambdaLayerService lambdaLayerService;
@@ -275,7 +260,6 @@ public class CloudFormationResourceProvisioner {
                                              LambdaService lambdaService, IamService iamService,
                                              SsmService ssmService, KmsService kmsService,
                                              SecretsManagerService secretsManagerService,
-                                             EventBridgeService eventBridgeService,
                                              ApiGatewayService apiGatewayService,
                                              ApiGatewayV2Service apiGatewayV2Service,
                                              EcrService ecrService,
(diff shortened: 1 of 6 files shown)


hectorvent:
Thanks, this was a useful pass. Four of the five are addressed in `cc3d4c9`, `5752d10`,
`aa6d4f8` and `22800eb`. Two of them need a correction to the reasoning even where the ask
itself was right, and on the fifth I owe you a partial concession.

**Rule targets are not restored on rollback (P1).** Real, and fixed in `22800eb`. One
correction to the reporting though: the stack does not report a completed rollback. A
provisioner without the hook returns the interface default of false, so
`rollbackUpdatedResources` (`CloudFormationService.java:1613`) falls to the next arm and the
resource gets `UPDATE_FAILED` with "Rollback is not implemented for AWS::Events::Rule", which
puts the stack in `UPDATE_ROLLBACK_FAILED` naming the rule. The targets were still lost, which
is the part that matters, and it is newly reachable on this branch: before the reconciliation
sweep landed here, targets were only ever upserted, so there was nothing to restore.

The fix follows `PipesCfnProvisioner`, which is the established shape in this package. The
rule's targets are snapshotted before the first target call of an update and put back on
failure, by `provision` when this rule's own reconciliation is what failed and by
`rollbackUpdate` when a later resource is. The restore is exact rather than additive, so a
target the failed update added does not survive it, and a restore that fails itself is
recorded on the resource rather than thrown over the failure already on its way out.

Scope note: this restores targets, not the rule's own properties, which `putRule` overwrites
on the same path. That gap predates this branch and is tracked separately.

**Composite policy id splits ambiguously (P1).** The specific failure does not hold. The
registry schema for `AWS::Events::EventBusPolicy` models `StatementId` as `[a-zA-Z0-9-_]+`
(1 to 64 characters) and `EventBusName` as `[\.\-_A-Za-z0-9]+`, and botocore agrees. Neither
admits a pipe, so `sid|1` cannot exist on AWS and the collision you describe is unreachable
there.

It did point at something real, so thank you: Floci was not validating `StatementId` at all,
which is a parity gap on its own and is what would let an AWS-invalid value reach the split in
the first place. `aa6d4f8` holds it to the modeled pattern and length. That closes the
ambiguity at source rather than by guessing at the separator.

**Update identity read off the resource (P2).** Correct, fixed in `5752d10`. It happened to
be right because the value was captured before anything assigned the new id, but the
convention exists precisely so the next edit cannot quietly break it. It now reads
`ctx.isUpdate()` and `ctx.priorPhysicalId()`, with a test pinning it.

**Extra dependency injected (P2).** You are right that this is a written rule, and my first
read of this finding was too dismissive. AGENTS.md line 278 does say a provisioner is
"`@ApplicationScoped`, injecting only the service it wraps", so precedent alone would not
settle it.

Having checked all 50 provisioners, though, I think the rule as written does not cover this
case rather than being broken by it. Four inject more than one collaborator, and the three
that predate this branch are exactly the ones whose resource properties are JSON documents:
`PipesCfnProvisioner` takes `PipesService` and `ObjectMapper`, and `CodeBuildCfnProvisioner`
and `CodePipelineCfnProvisioner` each add `RegionResolver` as well. `AWS::Events::EventBusPolicy`
carries a `Statement` that has to be merged into the bus policy by `Sid`, which is JSON work
the wrapped service does not do, and that code came across from the monolith, which held an
`ObjectMapper` for the same reason.

The alternative is constructing a mapper in the class rather than reusing the configured one,
which trades a documented rule for a worse one. Pipes is also the design this branch's new
rollback follows, so the two now match deliberately. I would rather amend the rule to record
the JSON exception than special-case this provisioner, and that is worth its own change rather
than widening a refactor PR. Happy to be argued out of it.

**Em dash (P2).** Correct, and it was the failing check. Fixed in `cc3d4c9` with your
suggested wording. Worth noting for context: the javadoc came across from
`CloudFormationResourceProvisioner` verbatim, which holds 30 of them. The gate scans only the
files a PR adds, so moving the prose into a new file is what surfaced a violation that was
always there.

Tests: 1865 in the CloudFormation and EventBridge packages, one pre-existing failure
(`CfnSchemaCoverageTest`, the `AWS::CertificateManager::Certificate` `Id` row) that reproduces
unchanged at this branch's previous head.

hectorvent:
Correcting myself on the injection thread. Greptile was right and my reply was wrong, so I want
that on the record rather than quietly fixed in `d3e8b49`.

I wrote that the alternative was "constructing a mapper in the class rather than reusing the
configured one, which trades a documented rule for a worse one". That does not survive a look at
the package. Four classes in `services/cloudformation/provisioners/` already hold a
`private static final ObjectMapper MAPPER` and inject only their service:
`LambdaEventInvokeConfigCfnProvisioner`, `LambdaEventSourceMappingCfnProvisioner`,
`Ec2NetworkCfnProvisioner`, and `ReplacementCleanup`, the shared rollback helper.

The first of those is the closest analogue there is. It is the other snapshot-based rollback
provisioner, it owns `CfnRollback.EVENT_INVOKE_CONFIG_SNAPSHOT_ATTR`, and it does the same JSON
snapshot work with a static mapper and a one-argument constructor. So the package already had a
convention that satisfies the rule, and I did not check for it: I built this rollback on
`PipesCfnProvisioner` and carried its injection across with the design.

My "the rule does not anticipate JSON-valued properties" argument was therefore answering a
question nobody needed to ask, and the proposal to amend AGENTS.md is withdrawn.

`d3e8b49` drops the injected mapper for a private static one and leaves
`EventsCfnProvisioner(EventBridgeService)`. What the configured mapper adds is a 100 MB string
limit and lenient empty-bean serialization, from `core/common/JacksonConfig.java` and
`quarkus.jackson`. Neither reaches this code: EventBridge caps a bus policy near 10 KB and a
target's `Input` at 8192 characters, both far under Jackson's default, and every type serialized
here has real properties. No behaviour changed, so the existing tests are the check that the
swap is inert. The CloudFormation and EventBridge packages run 1865 tests with only the
pre-existing `CfnSchemaCoverageTest` ACM row failing, which reproduces on this branch's earlier
heads.

Separately, and not for this PR: `PipesCfnProvisioner`, `CodeBuildCfnProvisioner` and
`CodePipelineCfnProvisioner` still inject a mapper and now stand alone against both the rule and
the rest of the package. That is worth its own change rather than widening a refactor PR.

pgermosen (approved):
Went through this carefully, comparing the deleted monolith code side by side with the new provisioner rather than just trusting "behaviour preserving," and it holds up: the extraction is faithful, nothing was dropped or duplicated, and both bug fixes check out exactly as described against the old code. The rule-identity fix is real (an unnamed rule really did regenerate its name on every update and orphan the previous one), and the policy-removal fix closes a genuine bare catch-all that was hiding failures behind a debug log while the stack reported success.

The target-rollback snapshot mechanism is the most complex new piece here, and it's the one I spent the most time tracing: snapshot taken once per update before the first mutating call, spent either during the same provision call on its own failure or later via the rollback hook when a different resource in the update fails, and a rule this run never touched correctly reports nothing to roll back rather than clearing live targets. The test suite exercises every one of those branches, including the case where the restore itself fails. Good use of the existing single-service-injection and stablePhysicalName/priorPhysicalId conventions too.

Only thing I noticed is a cosmetic import ordering slip in CfnProvisionerFixture.java, nothing worth holding this up for. Good to merge.

PR #3292, opened 2026-09-09
fix(sqs): bound message move task history

nkootstra:
## Summary

SQS message move task terminal state is now bounded, so repeated redrive tasks cannot grow in-memory state without limit. The service keeps the latest 10 terminal summaries per source queue, expires them after one hour, and removes cancellation flags when workers reach a terminal state.

The CloudTrail advanced selectors integration test now uses an isolated profile with a long scheduled flush interval. This removes the race where the 60-second background flush could consume records before the test assertions.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Previously, every completed or cancelled task remained in `moveTasksByHandle`, so a long-running emulator could accumulate terminal state. AWS documents that `ListMessageMoveTasks` returns the most recent tasks, up to 10, for a specific source queue. AWS also supports only one active message movement task per queue. This change preserves those limits while keeping recent task history available.

The one-hour terminal-summary TTL is an internal memory policy. AWS does not specify a terminal-summary TTL.

## CI Stability

CI previously failed in `CloudTrailAdvancedSelectorsIntegrationTest` when the scheduled writer flush started at the same 60-second boundary as the test. The test now disables that scheduled race and calls its explicit flush path.

## Checklist

- [ ] `./mvnw test` passes locally (the full suite run did not finish within 16 minutes)
- [x] `./mvnw -Dtest=CloudTrailAdvancedSelectorsIntegrationTest test` passes

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/sqs/SqsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/sqs/SqsService.java
@@ -25,6 +25,8 @@ import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
 
+import java.time.Clock;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Good fix, and the eviction logic is more careful than it needed to be in the right way, scoping the cap of 10 per source ARN rather than globally, and using a sequence number instead of wall clock so two tasks finishing in the same millisecond still evict in the right order. I also went looking for the parallel leak in the cancellation-flag map you mention in the description, and confirmed it's cleaned up in a finally block so it's covered no matter how the worker thread exits.

Tests are thorough, especially using a swappable clock for the TTL assertions instead of real sleeps, and checking that an expired task handle now 404s on cancel rather than silently succeeding.

One thing for next time rather than this PR: the CloudTrail flakiness fix is unrelated to the SQS change and would have been cleaner as its own PR, but it's a small, correctly scoped fix so not worth splitting out now.

Good to merge.

PR #3294, opened 2026-09-09
test(ec2): cover launch template not found behavior

### LINKED ISSUE #3234: [BUG] DescribeLaunchTemplates returns empty success for nonexistent explicit names
## Service

EC2

## AWS API Action

`DescribeLaunchTemplates` with an explicit nonexistent `LaunchTemplateName`.

## Expected behavior

An explicitly requested launch-template name that does not exist should return the AWS error `InvalidLaunchTemplateName.NotFoundException`, rather than a successful empty list.

Karpenter 1.8.8 relies on this error to create a missing launch template. Its [lookup implementation](https://github.com/aws/karpenter-provider-aws/blob/v1.8.8/pkg/providers/launchtemplate/launchtemplate.go#L229-L248) calls `DescribeLaunchTemplates` with a name and creates the template only when the error is recognized as NotFound; its [error definitions](https://github.com/aws/karpenter-provider-aws/blob/v1.8.8/pkg/errors/errors.go#L27-L45) include this exact code. This report was reproduced against Floci; no real AWS account was used for comparison.

Normal unqualified listing and filter queries with no matches should continue to return an empty successful list. The fix should distinguish those cases from missing explicitly named resources.

## Actual behavior

The AWS CLI exits successfully and returns:

```json
{
    "LaunchTemplates": []
}
```

Karpenter consequently fails EC2NodeClass validation with:

```text
validating ec2:CreateLaunchTemplate authorization, expected to find one launch template (launch-template-count=0)
```

`ValidationSucceeded` and `Ready` remain `Unknown`, preventing progression to provisioning. AMI, subnet, security-group, and instance-profile readiness are True. After the ENI metadata fix in #3179, the instance-type response checks pass and the controller is panic-free; this is the next discovery blocker. No fleet execution or NodeClaim provisioning was attempted after the failed gate.

The [pinned Floci implementation](https://github.com/floci-io/floci/blob/4944764fa381b34a4118d89640f84eb7ea5e5f66/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2Service.java#L4417-L4426) filters the launch-template collection and returns an empty list without checking whether explicitly requested names exist.

## Reproduction

With Floci listening on localhost:4566, run this read-only probe. The generated name must not already exist:

```bash
(
  export AWS_ACCESS_KEY_ID=000000000000
  export AWS_EC2_METADATA_DISABLED=true
  export AWS_PAGER=""
  unset AWS_SESSION_TOKEN AWS_SECURITY_TOKEN

  missing_name="floci-missing-template-$(date +%s)-$$"
  aws --endpoint-url http://localhost:4566 --region us-east-1 \
    ec2 describe-launch-templates \
    --launch-template-names "$missing_name" \
    --output json
  printf 'CLI exit status: %s\n' "$?"
)
```

Observed: `LaunchTemplates: []` and exit status 0. Expected: a nonzero CLI exit status containing `InvalidLaunchTemplateName.NotFoundException`.

Suggested regression coverage: an unknown explicit name, an existing name, mixed existing/missing names, and empty list/filter results. Explicit IDs should also be checked against their distinct AWS error contract.

## Environment

- Floci version / image tag: `floci/floci:nightly-09072026@sha256:0575ab250927ca8665cf326667f99ac8c8212ef1480eb1c52c0d0d2e8460a95d`
- OCI source revision: `4944764fa381b34a4118d89640f84eb7ea5e5f66`, containing merged #3179 (`f9686d5eb2468dbd49ffa0362aa6e3d35ab102cd`).
- Build provenance: successful [nightly run 34185659332](https://github.com/floci-io/floci/actions/runs/34185659332); arm64/amd64 build and publishing checkout revisions match the OCI label.
- AWS CLI: `aws-cli/2.36.36 Python/3.14.7 Darwin/25.5.0 source/arm64`.
- Java SDK version (if applicable): not applicable; reproduced with AWS CLI and observed with Karpenter's Go AWS SDK client.
- How you're running Floci (Docker / native / `mvn quarkus:dev`): Docker Compose, native Linux arm64 image, on macOS/OrbStack; local fake credentials only.
- Karpenter: 1.8.8; local Floci EKS backed by k3s `v1.34.1+k3s1`.
- Observed: 2026-09-08.

luiseterc:
## Summary

Closes #3234. Add the two regression test files requested after #3245 was superseded by #3246.

This PR is based on the current upstream `main` and contains no production changes. The merged implementation already returns AWS-compatible errors for missing explicit launch-template IDs and names while preserving filter-only empty results.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

Adds coverage for the EC2 Query protocol and AWS SDK for Java 2.52.0:

- missing explicit names return `InvalidLaunchTemplateName.NotFoundException`;
- missing explicit IDs return `InvalidLaunchTemplateId.NotFound`;
- existing and mixed requests behave correctly;
- cross-region lookups return the appropriate not-found error;
- unmatched filter queries continue to return an empty successful list;
- an existing template can be created and then described through the SDK.

## Checklist

- [x] `./mvnw test` passes locally for the new AWS Query integration test

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2DescribeLaunchTemplatesTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/Ec2DescribeLaunchTemplatesTest.java
@@ -1,0 +1,67 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.ec2.Ec2Client;
+import software.amazon.awssdk.services.ec2.model.Ec2Exception;
+import software.amazon.awssdk.services.ec2.model.RequestLaunchTemplateData;
+
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class Ec2DescribeLaunchTemplatesTest {
+
+    @Test
+    void missingNameCanTriggerCreateThenDescribe() {
+        String name = "sdk-describe-" + UUID.randomUUID();
+        try (Ec2Client ec2 = TestFixtures.ec2Client()) {
+            Ec2Exception error = assertThrows(Ec2Exception.class,
+                    () -> ec2.describeLaunchTemplates(r -> r.launchTemplateNames(name)));
+            assertThat(error.statusCode()).isEqualTo(400);
+            assertThat(error.awsErrorDetails().errorCode())
+                    .isEqualTo("InvalidLaunchTemplateName.NotFoundException");
+
+            String id = ec2.createLaunchTemplate(r -> r.launchTemplateName(name)
+                    .launchTemplateData(RequestLaunchTemplateData.builder()
+                            .imageId("ami-0abcdef1234567890").build()))
+                    .launchTemplate().launchTemplateId();
+            try {
+                assertThat(ec2.describeLaunchTemplates(r -> r.launchTemplateNames(name)).launchTemplates())
+                        .singleElement().satisfies(t -> assertThat(t.launchTemplateId()).isEqualTo(id));
+                assertThat(ec2.describeLaunchTemplates(r -> r.launchTemplateIds(id)).launchTemplates())
+                        .singleElement().satisfies(t -> assertThat(t.launchTemplateName()).isEqualTo(name));
+                Ec2Exception mixed = assertThrows(Ec2Exception.class,
+                        () -> ec2.describeLaunchTemplates(r -> r.launchTemplateNames(name, name + "-missing")));
+                assertThat(mixed.awsErrorDetails().errorCode())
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Thanks for following through on this, exactly the two files hectorvent asked for when #3245 got closed in favor of #3246. I checked the new assertions against the current DescribeLaunchTemplates implementation rather than just trusting the description, and they match: a missing name gets InvalidLaunchTemplateName.NotFoundException, a missing id gets InvalidLaunchTemplateId.NotFound, mixed present/absent lists fail on the first absent one, cross-region lookups correctly report not-found, and filter-only queries still come back as an empty successful list rather than an error.

The round-trip create-then-describe case, the mixed-id case, the cross-region case, and the SDK-driven compat test are all genuinely new coverage that didn't exist anywhere before this, so this fills a real gap rather than just duplicating what #3246 already added inline. Good to merge.

PR #3295, opened 2026-09-09
fix(appconfigdata): return empty payload on repeat polls

nkootstra:
## Summary

AppConfigData now tracks the last served hosted configuration version per configuration session. Repeat polls with no new deployment return an empty payload and a fresh poll token. A later deployment is returned correctly, and sessions remain independent.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

AppConfigData clients must use each returned next poll token. Before this fix, repeat polls could return the same configuration payload again when the active version had not changed.

The behavior is covered by RestAssured integration tests and AWS SDK for Java v2 compatibility tests.

## Checklist

Full-suite note: the run reported one unrelated failure in `MacieOrganizationIntegrationTest.sharedAdminRouteKeepsGuardDutyBehavior` and was stopped after later S3 tests made no progress. Focused AppConfig tests, compatibility tests, compilation, packaging, and diff checks passed.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppConfigTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/AppConfigTest.java
@@ -28,6 +28,7 @@ class AppConfigTest {
     private static String deploymentStrategyId;
     private static String configurationToken;
     private static String secondConfigurationToken;
+    private static String unchangedConfigurationToken;
     private static String intervalSessionToken;
     private static String emptyAppId;
     private static String emptyEnvId;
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Clean fix, and the one thing that actually needed proving here, that sessions track their last-served version independently rather than sharing any state, is directly tested rather than assumed. Starting two sessions against the same profile and confirming both still get served the current version, then that only the one which already polled it gets the empty response next time, is exactly the right test for this.

Also like that the SDK compat test goes through the real AWS SDK's response deserialization for the empty-payload case instead of just checking raw JSON, that's a stronger check on the content type and version label being right.

Good to merge.

PR #3298, opened 2026-09-09
fix: bound CloudTrail delivery retries

nkootstra:
## Summary

Bound CloudTrail delivery retries by trail with shared record and byte limits. Retry buffers now keep separate event-region queues under one trail-wide budget, preserve retry order, and serialize scheduled and manual flushes.

Delivery status now reports only completed delivery time and the latest delivery error. Failed status persistence does not cause duplicate S3 writes, and post-write self-delivery failures do not re-queue delivered records.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Before this change, `GetTrailStatus` reported `LatestDeliveryTime` at `StartLogging` time and emitted the retired `LatestDeliveryAttemptTime` field. Delivery retries could also grow without a trail-wide bound or duplicate a batch after a successful S3 write.

Verified with the AWS SDK for Java v2.52.0 `CloudTrailTest`, JSON 1.1 integration tests, and S3 log retrieval after destination recovery.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudTrailTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/CloudTrailTest.java
@@ -87,7 +87,7 @@ class CloudTrailTest {
 
         GetTrailStatusResponse started = cloudTrail.getTrailStatus(r -> r.name(trailName));
         assertThat(started.isLogging()).isTrue();
-        assertThat(started.latestDeliveryTime()).isNotNull();
+        assertThat(started.latestDeliveryTime()).isNull();
 
         cloudTrail.stopLogging(r -> r.name(trailName));
 
--- a/src/main/java/io/github/hectorvent/floci/services/cloudtrail/CloudTrailEntry.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudtrail/CloudTrailEntry.java
@@ -22,7 +22,17 @@ public record CloudTrailEntry(
         boolean logging,
         Long startLoggingTime,
         Long stopLoggingTime,
-        Map<String, String> tags) {
+        Map<String, String> tags,
+        Long latestDeliveryTime,
+        String latestDeliveryError) {
+
+    public CloudTrailEntry(Trail trail, List<EventSelector> selectors,
+                           List<AdvancedEventSelector> advancedSelectors, boolean logging,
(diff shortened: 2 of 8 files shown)


pgermosen (commented):
The retry-budget idea here is good and both things Greptile flagged are genuinely addressed on this branch, I traced the in-flight tracking myself rather than just taking the commit messages at face value, and it correctly closes the unbounded-buffer gap. But the bigger issue is that this now conflicts with #3293, which merged a little before your last commit here, and it's not a simple rebase. You and that PR both redesigned the same pending-records structure independently: that one bounds flush-batch memory with a per-trail lock and bounded batch looping, this one bounds retry backlog with a trail-wide budget and in-flight tracking. Resolving conflict markers alone would drop one fix or the other, this branch still wraps the whole flushAll in one global lock instead of the per-trail one that just landed, and still does a single unbounded drain per cycle. Worth rebasing onto main and deliberately layering the retry budget on top of what's there now rather than just fixing markers.

One smaller thing I couldn't tell either way: trimTailToLimit still trims the just-failed region's own queue first before touching others, which is what Greptile's second finding was about. That could be the bug, or it could be intentional so one flaky region eats its own backlog instead of starving healthy ones. Can you say which it's meant to be?

nkootstra:
@pgermosen It’s intentional. We drop newer records from the failing region first, so older records and healthy regions keep their place.

nkootstra:
Blocked by PR #3320.

hectorvent:
pgermosen's review was right to insist on this, and I want to record that it landed, since the thread does not say so.

Their concern was that this PR and #3293 had independently redesigned the same pending-records structure, so fixing conflict markers alone would drop one of the two. It did not: #3293 merged at 16:23 and you rebased onto it at 16:39, then followed with a commit specifically preserving per-trail flush isolation. Both designs are on the branch now. #3293's per-trail `FlushLock` and bounded batch loop are intact, and your retry budget, in-flight tracking and `trimTailToLimit` are all still there.

Your answer on `trimTailToLimit` settles the open question too: dropping newer records from the failing region first, so a flaky region eats its own backlog instead of starving healthy ones, is a sensible policy and it is what the code does.

Merging this.

hectorvent (approved):
Approving: pgermosen's finding verified closed.

PR #3299, opened 2026-09-09
fix(cloudfront): emit origin timeout fields

### LINKED ISSUE #3221: CloudFront: Origin omits ResponseCompletionTimeout, and CustomOriginConfig omits OriginKeepaliveTimeout/OriginReadTimeout/OriginSslProtocols
Found while adding a Terraform compatibility test for #3217 (the AllowedMethods/CachedMethods fix). Even with an origin that has no custom fields set, a `terraform plan` run immediately after `apply` is never clean for `aws_cloudfront_distribution`, for two separate reasons in `CloudFrontController.xmlOrigin`:

1. `ResponseCompletionTimeout` is never emitted on `Origin`, for either an S3 or a custom origin. The Terraform AWS provider treats it as an Optional+Computed attribute, so an absent value reads back as unknown and the `origin` block is replaced (remove+add) on every plan, forcing `Plan: 0 to add, 1 to change, 0 to destroy` forever.
2. For a custom origin specifically, `CustomOriginConfig` only echoes `HTTPPort`, `HTTPSPort`, and `OriginProtocolPolicy`. `OriginKeepaliveTimeout`, `OriginReadTimeout`, and `OriginSslProtocols` are accepted on create/update (there's parsing code for them) but never stored or echoed back, so they always read back as their zero value regardless of what was requested.

Repro: `terraform apply` any `aws_cloudfront_distribution` (S3 or custom origin), then `terraform plan` again immediately. Expect "No changes"; get an origin block replacement instead.

Scope: in `xmlOrigin`, always emit `ResponseCompletionTimeout` (defaulting appropriately when unset), and store/echo `OriginKeepaliveTimeout`/`OriginReadTimeout`/`OriginSslProtocols` on `CustomOriginConfig` the same way `HTTPPort`/`HTTPSPort`/`OriginProtocolPolicy` already are.

omatheusmesmo:
## Summary

The CloudFront Origin response serializer dropped optional timeout fields that AWS accepts on create/update but never echoes back, causing perpetual terraform plan diffs on aws_cloudfront_distribution origins.

Closes #3221

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior before this fix: CustomOriginConfig responses omitted OriginKeepaliveTimeout and OriginReadTimeout entirely, and Origin never emitted ResponseCompletionTimeout, even though AWS always echoes these per the CloudFront API reference.

- Origin now carries a responseCompletionTimeout and xmlOrigin always emits ResponseCompletionTimeout at the Origin level (0 when unset, matching CloudFront's "no maximum enforced" behavior).
- CustomOriginConfig responses now include OriginKeepaliveTimeout (default 5s) and OriginReadTimeout (default 30s) per the CloudFront API reference defaults.
- Added parsing for OriginKeepaliveTimeout and OriginReadTimeout in CreateDistribution/UpdateDistribution request bodies, since neither was previously captured despite being accepted in the XML schema.

OriginSslProtocols round-tripping is already handled on main (#3182), so this PR only covers the remaining timeout fields from #3221.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
@@ -40,6 +40,21 @@ public class CloudFrontController {
     private static final String DEFAULT_GEO_RESTRICTION_TYPE = "none";
     private static final int EMPTY_QUANTITY = 0;
 
+    /**
+     * Origin timeout constraints from the CloudFront {@code CustomOriginConfig} and
+     * {@code Origin} API references. {@code OriginReadTimeout} accepts 1-120 seconds, default 30.
+     * {@code OriginKeepaliveTimeout} accepts 1-300 seconds, default 5. {@code ResponseCompletionTimeout}
+     * has no enforced maximum when unset (represented here as 0), but when set must be at least
+     * {@code OriginReadTimeout}.
+     */
+    private static final int DEFAULT_ORIGIN_READ_TIMEOUT_SECONDS = 30;
+    private static final int MIN_ORIGIN_READ_TIMEOUT_SECONDS = 1;
+    private static final int MAX_ORIGIN_READ_TIMEOUT_SECONDS = 120;
+    private static final int DEFAULT_ORIGIN_KEEPALIVE_TIMEOUT_SECONDS = 5;
+    private static final int MIN_ORIGIN_KEEPALIVE_TIMEOUT_SECONDS = 1;
+    private static final int MAX_ORIGIN_KEEPALIVE_TIMEOUT_SECONDS = 300;
+    private static final int DEFAULT_RESPONSE_COMPLETION_TIMEOUT_SECONDS = 0;
+
     private final CloudFrontService service;
 
     @Inject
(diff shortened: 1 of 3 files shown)


hectorvent:
Thanks for this, and for splitting it cleanly from the `OriginSslProtocols` work that already landed. The three members are the right ones: `ResponseCompletionTimeout` on `Origin`, `OriginReadTimeout` and `OriginKeepaliveTimeout` on `CustomOriginConfig`, none of which were parsed or emitted before.

**(blocking)** The bound checks throw `InvalidArgument`, but CloudFront declares a dedicated code for each of these two: `InvalidOriginReadTimeout` and `InvalidOriginKeepaliveTimeout`, both 400 and sender-fault, both declared on `CreateDistribution` and `UpdateDistribution` ([cloudfront/2020-05-31](https://github.com/boto/botocore/blob/develop/botocore/data/cloudfront/2020-05-31/service-2.json)). Would you use those instead? SDK clients map them to typed exceptions, which `InvalidArgument` does not give them.

**(follow-up, separate PR)** `ResponseCompletionTimeout` is emitted as `0` even when the caller never set it. AWS omits an optional member it has no value for, so a placeholder is a small deviation in the other direction; worth leaving the element out when the field is null.

The bounds themselves are fine, and citing the API reference for them is right since the model carries no min/max on those integers.

omatheusmesmo:
@hectorvent Fixed, now using InvalidOriginReadTimeout/InvalidOriginKeepaliveTimeout. Ready for another look.

hectorvent (approved):
That is exactly it, thank you. Both bound checks now raise the dedicated codes, and the ordering check keeps `InvalidArgument`, which is the right split since CloudFront declares no specific code for that one.

I re-read it against the model at your new head and it holds: `InvalidOriginReadTimeout` and `InvalidOriginKeepaliveTimeout` are declared on `CreateDistribution`, `UpdateDistribution` and `CreateDistributionWithTags` alike, so every path a caller can reach this through maps to a typed exception now.

The rebase in the same push cleared the conflict too. The `ResponseCompletionTimeout` placeholder is still the one open item and it stays a follow-up, no need to touch it here.

Merging. Thanks for turning both rounds around so quickly.

PR #3300, opened 2026-09-09
fix(lambda): reject dotted and over-length function names

### LINKED ISSUE #3238: [BUG] lambda: CreateFunction accepts function names AWS rejects (dots, and up to 256 characters)
## Service

Lambda (function names)

## AWS API Action

`CreateFunction`, and every action that resolves a `FunctionName`

## Expected behavior

Real AWS constrains the name-only form of `FunctionName` to `[a-zA-Z0-9-_]+`, at most 64 characters.
A name containing a dot is rejected:

```
An error occurred (ValidationException) when calling the CreateFunction operation:
1 validation error detected: Value 'foo.v1' at 'functionName' failed to satisfy constraint:
Member must satisfy regular expression pattern: (arn:(aws[a-zA-Z-]*)?:lambda:)?([a-z]{2}(-gov)?-[a-z]+-\d{1}:)?(\d{12}:)?(function:)?([a-zA-Z0-9-_]+)(:(\$LATEST|[a-zA-Z0-9-_]+))?
```

Floci's own rule is that the AWS SDK must work against it without modification. The failure mode
here is the inverse of the usual one: a name that Floci accepts is not a name AWS will accept, so
code that passes locally fails on deploy, and the emulator is the thing that taught it the wrong
contract.

## Actual behavior

`LambdaArnUtils` validates the name, but more permissively than AWS on two axes:

```java
private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z0-9-_.]+");
private static final int MAX_NAME_LENGTH = 256;
```

The character class carries a `.` that AWS's does not, and the cap is 256 rather than 64.

Measured against `main` at `cc8b8e9f`, posting `CreateFunction` to `/2015-03-31/functions` with an
otherwise valid request and varying only `FunctionName`:

| `FunctionName` | Floci | Real AWS |
|---|---|---|
| `control-valid-fn` | `201 Created` | accepted |
| `foo.v1` | **`201 Created`** | `ValidationException` |
| `..` | **`201 Created`** | `ValidationException` |
| `aaa...` (200 chars) | **`201 Created`** | `ValidationException`, over the 64 limit |
| `foo@versions` | `400 InvalidParameterValueException` | rejected |
| `foo bar` | `400 InvalidParameterValueException` | rejected |
| `foo/bar` | `400 InvalidParameterValueException` | rejected |
| `naive-fné` | `400 InvalidParameterValueException` | rejected |
| `tab\tname` | `400 InvalidParameterValueException` | rejected |

The rejections all report `FunctionName contains invalid characters: <name>`. So the validation is
present and mostly right; it is the dot and the length ceiling that diverge.

## Why this is worth fixing rather than tolerating

It has already cost the project a design decision. In #3041 a published version's code directory was
originally a `<function>.v<n>` sibling of the function's own directory. Because a function may
genuinely be named `foo.v1`, that name collided with the directory version 1 of `foo` would claim,
and deleting either function silently corrupted the other. The fix was to move version directories
under a `<function>@versions` container, chosen specifically because `@` is outside the character
set `CodeStore.sanitizeName` can emit. That is a sound defence and it should stay, but it exists to
work around this divergence rather than to express anything about versions.

`..` being an accepted function name is the sharper edge. It is inert today only because
`CodeStore.sanitizeName` collapses an all-dots segment to `_` before it reaches `Path.resolve`,
which is exactly what that code was written for. Every future component that derives a filesystem
path, an object key, a container name or a log group from a function name inherits the same
obligation, and gets no help from the type system in remembering it.

## Correction to an existing comment

The Javadoc merged with #3041 on `CodeStore.getVersionsPath` says Floci "does not restrict the
character set of `FunctionName` today, only that it is non-blank". That is wrong, and it is mine.
The validation above has existed since #450. The conclusion the comment draws is still correct, since
`foo.v1` really is creatable, but the stated reason is not. Worth correcting alongside whatever
happens here.

## Reproduction

```bash

printf "exports.handler = async () => ({});\n" > index.js && zip -q fn.zip index.js

# Accepted by Floci, rejected by AWS
awsl lambda create-function --function-name foo.v1 --runtime nodejs20.x \
  --role arn:aws:iam::000000000000:role/r --handler index.handler --zip-file fileb://fn.zip

# Also accepted
awsl lambda create-function --function-name .. --runtime nodejs20.x \
  --role arn:aws:iam::000000000000:role/r --handler index.handler --zip-file fileb://fn.zip

awsl lambda list-functions --query 'Functions[].FunctionName' --output text
```

## Suggested direction

Bring `NAME_PATTERN` to `[a-zA-Z0-9-_]+` and the name-only ceiling to 64, matching the live service.

Two things worth a maintainer's call before anyone writes that:

1. **It is a behaviour change, not purely a fix.** Any function already created with a dotted or
   over-long name becomes unaddressable, and in `persistent` mode those records survive a restart.
   Rejecting only on create, while continuing to resolve names that already exist, is the gentler
   option; failing both is the more honest one. I do not think this should be decided in a PR
   description.
2. **The 256 ceiling may be deliberate.** `MAX_NAME_LENGTH` is applied to the name segment, but AWS
   allows 140 characters for the full ARN form and 64 for the name-only form, so the correct fix is
   probably two limits rather than one lowered number. Worth confirming against
   `LambdaArnUtils.parseNameWithOptionalQualifier`'s ARN path rather than changing the constant
   alone.

## Environment

- Floci version / image tag: `main` @ `cc8b8e9f`
- Java SDK version (if applicable): n/a, reproduced against the REST endpoint directly

- [x] Yes
- [ ] No

omatheusmesmo:
## Summary

LambdaArnUtils validated FunctionName more permissively than the live service: NAME_PATTERN allowed a dot and MAX_NAME_LENGTH capped a bare name at 256 characters.

Closes #3238

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Real AWS's CreateFunction FunctionName pattern is `[a-zA-Z0-9-_]+` (no dot), and a name-only FunctionName is limited to 64 characters (the 140-character ceiling documented in the API reference applies to the full ARN string, not the bare name).

Changes in `LambdaArnUtils`:
- `NAME_PATTERN` no longer allows `.`
- `MAX_NAME_LENGTH` corrected from 256 to 64, applied only to a bare (non-ARN) name
- Added a 140-character ceiling on the full input string, matching AWS's full-ARN constraint
- A name that fails the character-set or length constraint now surfaces as `ValidationException` with a message resembling AWS's own wording, instead of a generic `InvalidParameterValueException`

`CreateFunction`, `GetFunction`, `UpdateFunctionConfiguration` and the other Lambda name-resolving APIs all route through this same `LambdaArnUtils.resolve`/`resolveWithQualifier`, so the fix applies uniformly; no other resource type (layers, aliases, versions) shares this pattern/constant.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaArnUtils.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/LambdaArnUtils.java
@@ -13,8 +13,17 @@ import java.util.regex.Pattern;
  */
 public final class LambdaArnUtils {
 
-    private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z0-9-_.]+");
-    private static final int MAX_NAME_LENGTH = 256;
+    /**
+     * FunctionName constraints from the Lambda API reference (CreateFunction, FunctionName):
+     * {@code [a-zA-Z0-9-_]+}, with no dot. A bare name (no ARN prefix) is capped at 64
+     * characters; the ARN and partial-ARN forms embed the same character class but are
+     * bounded instead by {@link #MAX_TOTAL_LENGTH} on the full string.
+     */
+    private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z0-9-_]+");
+    private static final int MAX_NAME_LENGTH = 64;
+    private static final int MAX_TOTAL_LENGTH = 140;
+    private static final String NAME_REGEX = "(arn:(aws[a-zA-Z-]*)?:lambda:)?([a-z]{2}(-gov)?-[a-z]+-\\d{1}:)?"
+            + "(\\d{12}:)?(function:)?([a-zA-Z0-9-_]+)(:(\\$LATEST|[a-zA-Z0-9-_]+))?";
     private static final Pattern ACCOUNT_PATTERN = Pattern.compile("\\d{12}");
     private static final Pattern QUALIFIER_PATTERN = Pattern.compile("\\$LATEST|[a-zA-Z0-9-_]+");
 
@@ -31,13 +40,18 @@ public final class LambdaArnUtils {
     public record ResolvedFunctionRef(String name, String qualifier, String region) {}
 
     /**
-     * Parses a {@code FunctionName} path parameter. Throws
-     * {@link AwsException} ({@code InvalidParameterValueException}, HTTP 400)
-     * on any malformed input.
+     * Parses a {@code FunctionName} path parameter. Throws {@link AwsException}, HTTP 400,
+     * on any malformed input: {@code ValidationException} when the name fails the
+     * {@code FunctionName} pattern or length constraint, {@code InvalidParameterValueException}
+     * for other malformed forms (bad ARN structure, mismatched qualifiers, and the like).
      */
     public static ResolvedFunctionRef resolve(String input) {
         if (input == null || input.isBlank()) {
             throw invalid("FunctionName must not be blank");
+        }
+        if (input.length() > MAX_TOTAL_LENGTH) {
+            throw validationFailure(input,
+                    "Member must have length less than or equal to " + MAX_TOTAL_LENGTH);
         }
 
         if (input.startsWith("arn:")) {
@@ -96,7 +110,7 @@ public final class LambdaArnUtils {
             throw invalid("ARN resource type must be 'function': " + input);
         }
         String name = resParts[1];
-        validateName(name);
+        validateNamePattern(name);
         String qualifier = resParts.length == 3 ? resParts[2] : null;
         if (qualifier != null) {
             validateQualifier(qualifier);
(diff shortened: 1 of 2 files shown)


pgermosen:
Checked this against the live Lambda FunctionName shape directly, the 140/64 split and the dropped dot match exactly, including the shape's own doc wording about the bare-name limit.

One gap worth closing before merge: the new tests pin the 140/141 boundary for a partial ARN, but not for a full ARN. The check happens once in resolve() before branching so the protection is already there, it's just not independently pinned for that form the way the repo's test conventions ask for.

pgermosen (approved):
Took another look at this, the 64/140 split and dropped dot match AWS's CreateFunction and UpdateFunctionConfiguration docs exactly, including the ValidationException vs InvalidParameterValueException split for shape-constraint failures versus malformed ARNs. Correcting my earlier note: the full-ARN 140/141 boundary is tested too, right alongside the partial-ARN one, resolveAccepts140CharFullArn and resolveRejectsFullArnLongerThan140Chars cover it. That wasn't a real gap, sorry for the noise. Approving.

PR #3301, opened 2026-09-09
fix(cloudformation): resolve dynamic references for all template values

### LINKED ISSUE #2213: [BUG] CloudFormation dynamic references (`{{resolve:ssm:…}}`) pass through as literal strings — env vars land set-but-garbage instead of resolved or failing
### What happens

CloudFormation **dynamic references** are not resolved. A template value such as

```yaml
Environment:
  Variables:
    PAYMENTS_GATEWAY_BASE_URL: '{{resolve:ssm:/my-service/gateway-url:1}}'
```

reaches the deployed Lambda as the **literal string** `{{resolve:ssm:/my-service/gateway-url:1}}`. No resolution, no error, no warning.

### Why it matters

The failure is worse than "unsupported", because the variable ends up **set but meaningless** rather than absent. Any code that branches on presence takes the configured path:

```js
const client = process.env.PAYMENTS_GATEWAY_BASE_URL
  ? new HttpClient(process.env.PAYMENTS_GATEWAY_BASE_URL)   // taken, with a garbage URL
  : new UnconfiguredClient();                                // the intended fail-closed path
```

So a fallback that is correct on real AWS is never exercised locally, and a service that would fail closed appears to be configured. `{{resolve:ssm-secure:…}}` used as a shared secret is the same shape with a sharper edge: both sides compare against the same literal template text and "match".

### Where

Repo-wide grep finds no resolver:

```
$ grep -rn 'resolve:ssm\|ssm-secure\|{{resolve\|DynamicReference' src/main/java/
(no matches)
```

`CloudFormationTemplateEngine` implements the intrinsics (`Ref`, `Fn::Sub`, `FindInMap`, `GetAtt`, …) but has no dynamic-reference stage. #1809 tracks resolving dynamic references for RDS credentials specifically; this is the general case, and the Lambda-environment-variable one in particular.

### Reproduce

```bash
aws ssm put-parameter --name /demo/url --value https://real.example.com --type String \
  --endpoint-url http://localhost:4566

cat > t.yaml <<'YAML'
Resources:
  Fn:
    Properties:
      Runtime: nodejs24.x
      Handler: index.handler
      InlineCode: "exports.handler = async () => ({statusCode:200});"
      Environment:
        Variables:
          URL: '{{resolve:ssm:/demo/url:1}}'
YAML

aws cloudformation deploy --template-file t.yaml --stack-name dynref \
  --endpoint-url http://localhost:4566 --capabilities CAPABILITY_IAM

aws lambda get-function-configuration --function-name <name> \
  --endpoint-url http://localhost:4566 --query 'Environment.Variables.URL'
# Expected: "https://real.example.com"
# Observed: "{{resolve:ssm:/demo/url:1}}"
```

### Suggested behaviour

Resolving `{{resolve:ssm:…}}` and `{{resolve:ssm-secure:…}}` against the emulator's own SSM would be ideal — Parameter Store with version pins and SecureString is already supported, so the backing data is there.

If that is not near-term, **failing the deploy on an unresolved dynamic reference** would still be a large improvement over literal passthrough, for the same reason as #1759: a value that is silently wrong is harder to notice than a deploy that stops.

omatheusmesmo:
## Summary

CloudFormation dynamic references (`{{resolve:ssm:...}}`, `{{resolve:secretsmanager:...}}`) were only resolved for the RDS MasterUsername/MasterUserPassword properties, which call CfnDynamicReferences directly. Every other property value goes through CloudFormationTemplateEngine#resolveNode, which had no dynamic-reference stage at all, so a value such as a Lambda Environment.Variables entry reached the deployed resource as the literal `{{resolve:ssm:...}}` text instead of the resolved value.

Closes #2213

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

`resolveNode` now runs the same CfnDynamicReferences resolution over every textual result it produces: a plain string literal carrying the syntax outright, and the text an intrinsic function (e.g. Fn::Sub) produces. `ssm-secure` stays restricted to the RDS master-credential path, matching the existing behavior there. The RDS code path is unchanged, since it resolves MasterUsername/MasterUserPassword directly rather than through resolveNode.

- `CloudFormationTemplateEngine`: added an optional dynamicReferenceResolver collaborator (a UnaryOperator<String>, following the same nullable-functional-collaborator pattern already used for importValueResolver) and applied it in resolveNode for both plain string values and the output of resolved intrinsics.
- `CloudFormationService` / `CloudFormationResourceProvisioner`: wired the new resolver into every place a CloudFormationTemplateEngine is constructed, backed by the existing CfnDynamicReferences bean.
- Tests: unit tests in CloudFormationTemplateEngineTest covering the resolver wiring (dynamic reference in a plain string, in an Fn::Sub result, a plain literal left untouched, and no-resolver backward compatibility), plus a new CloudFormationDynamicReferenceIntegrationTest that deploys a real stack with a Lambda Environment.Variables entry using `{{resolve:ssm:...}}` and `{{resolve:secretsmanager:...}}` and asserts the deployed function has the resolved value, alongside a regression case for plain literal strings.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudformation/CloudFormationResourceProvisioner.java
@@ -477,7 +477,8 @@ public class CloudFormationResourceProvisioner {
     public StackResource provisionStandalone(String resourceType, JsonNode properties, String region, String accountId) {
         CloudFormationTemplateEngine engine = new CloudFormationTemplateEngine(
                 accountId, region, "cloudcontrol", "cloudcontrol",
-                Map.of(), new HashMap<>(), new HashMap<>(), Map.of(), Map.of(), objectMapper, name -> null);
+                Map.of(), new HashMap<>(), new HashMap<>(), Map.of(), Map.of(), objectMapper, name -> null,
+                value -> dynamicReferences.resolveDynamicReferences(value, region, false));
         return provision("resource", resourceType, properties, engine, region, accountId, "cloudcontrol");
     }
 
@@ -1391,7 +1392,7 @@ public class CloudFormationResourceProvisioner {
         if (instance != null) {
             instance = rdsService.modifyDbInstance(
                     id,
-                    resolveDynamicReferences(resolveOptional(props, "MasterUserPassword", engine), region, true),
+                    resolveDynamicReferences(resolveOptionalWithoutDynamicReferences(props, "MasterUserPassword", engine), region, true),
                     parseBoolProp(props, "EnableIAMDatabaseAuthentication", engine),
                     resolveOptional(props, "DBSubnetGroupName", engine));
         } else {
@@ -1399,8 +1400,8 @@ public class CloudFormationResourceProvisioner {
                     id,
                     resolveOptional(props, "Engine", engine),
                     resolveOptional(props, "EngineVersion", engine),
-                    resolveDynamicReferences(resolveOptional(props, "MasterUsername", engine), region, false),
-                    resolveDynamicReferences(resolveOptional(props, "MasterUserPassword", engine), region, true),
+                    resolveDynamicReferences(resolveOptionalWithoutDynamicReferences(props, "MasterUsername", engine), region, false),
+                    resolveDynamicReferences(resolveOptionalWithoutDynamicReferences(props, "MasterUserPassword", engine), region, true),
                     resolveOptional(props, "DBName", engine),
                     firstNonBlank(resolveOptional(props, "DBInstanceClass", engine), "db.t3.micro"),
                     parseIntProp(props, "AllocatedStorage", engine, 20),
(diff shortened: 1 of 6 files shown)


hectorvent:
Thanks, this is a real gap and the fix is the right shape: dynamic references only worked where a provisioner called the resolver by hand, so `{{resolve:ssm:...}}` anywhere else reached the resource as literal text.

**(blocking)** The new general stage rejects the one case `ssm-secure` is legitimately allowed in. `executeTemplate` builds both engines with `allowSsmSecure` false, so every value passes that stage first, including an RDS `MasterUserPassword`, and `CfnDynamicReferences` throws `ValidationError` for `ssm-secure` whenever the flag is false. The reference is rejected one stage before the RDS path that would have resolved it, leaving the `allowSsmSecure=true` call site unreachable.

That is what CI shows: `CloudFormationRdsDynamicReferenceTest.resolvesSecureSsmMasterPassword` expects `CREATE_COMPLETE` and gets `ROLLBACK_COMPLETE`. It is the only failure; the other 1745 pass.

Your new javadoc is where the assumption shows: it says every other property reaches the resolver through the general path, but a master password is both. One direction if it helps: leave `ssm-secure` untouched in the general stage and let the property-specific path keep that call, since it alone knows the resource type and property.

omatheusmesmo:
@hectorvent Fixed, ssm-secure now goes through the dedicated RDS path only, rejected everywhere else. Ready for another look.

hectorvent (approved):
That is the right fix, and in the right place. Splitting the engine so the RDS master-credential path takes intrinsics only and then resolves the reference itself keeps the decision with the one caller that knows the resource type and the property name, which is what the general stage can never know.

Thanks especially for the negative test. `createStack_ssmSecureDynamicReferenceOnANonRdsResourceFailsResourceCreation` pins the half that would otherwise rot silently: it is easy to make the RDS case work again and quietly stop rejecting `ssm-secure` everywhere else, and now that cannot happen unnoticed.

Full suite is green at your head and the rebase was clean. Merging, and thanks for the quick turnaround on both rounds.

PR #3302, opened 2026-09-09
feat(stepfunctions): resolve Fail state ErrorPath and CausePath

### LINKED ISSUE #3255: [BUG] Step Functions (SFN): Fail state ignores ErrorPath and CausePath

### Current Behavior

A `Fail` state accepts `ErrorPath` and `CausePath` at `CreateStateMachine` but ignores them at runtime. The error falls back to `States.Runtime` and the cause to the empty string.

**Both paths:**

```json
{
  "StartAt": "Boom",
  "States": {
    "Boom": { "Type": "Fail", "ErrorPath": "$.e", "CausePath": "$.c" }
  }
}
```

Input `{ "e": "StudentDeletionFailed", "c": "boom" }`:

| | error | cause |
| --- | --- | --- |
| Floci | `States.Runtime` | `""` |
| Step Functions Local 2.0.0 | `StudentDeletionFailed` | `boom` |

**A literal `Error` with a `CausePath`:**

```json
{
  "StartAt": "Boom",
  "States": {
    "Boom": { "Type": "Fail", "Error": "422", "CausePath": "$.Cause" }
  }
}
```

Input `{ "Cause": "Unable to create an account" }`:

| | error | cause |
| --- | --- | --- |
| Floci | `422` | `""` |
| Step Functions Local 2.0.0 | `422` | `Unable to create an account` |

So a literal `Error` is honoured while the `CausePath` beside it is dropped.

### Expected Behavior

`ErrorPath` and `CausePath` resolve against the state input and populate `executionFailedEventDetails.error` and `.cause`, as they do on Step Functions Local and real AWS.

Docker image built from `main` (`af42564`), `docker/Dockerfile`.

### Steps To Reproduce

Create either definition above, start an execution with the matching input, then:

```bash
aws stepfunctions get-execution-history --endpoint-url http://localhost:4566 \
  --execution-arn <arn> \
  --query 'events[?executionFailedEventDetails].executionFailedEventDetails'
```

### Cause

`AslExecutor.executeFail` (`AslExecutor.java:2021`) reads only `Error` and `Cause`, plus their JSONata forms. In JSONPath mode it never looks at `ErrorPath` or `CausePath`, so an `ErrorPath` leaves the error unset — which surfaces as `States.Runtime` — and a `CausePath` leaves the cause `""`.

### Anything else?

This is easy to miss from a test suite, because the execution does fail either way; only the error code and message are wrong. In our 735-example suite it accounts for 17 failures, all of them assertions on the error a state machine raises on its failure route. They were the hardest bucket to attribute, since the assertion helper reports a mismatch without printing the actual code.

Filed alongside #3253 and #3254 from the same run.

omatheusmesmo:
## Summary

A Fail state's `ErrorPath` and `CausePath` were parsed from the state machine definition but never evaluated at runtime: the executor only read the static `Error` and `Cause` fields, so a Fail state that used `ErrorPath`/`CausePath` always reported `States.Runtime` and an empty cause instead of the dynamic value AWS resolves from the state input.

Closes #3255

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the AWS Step Functions ASL specification and documentation for the `Fail` state (`ErrorPath`/`CausePath` fields), and against the behavior already established in this codebase for issue #2521 (unresolvable JSONPath in a payload template fails with `States.Runtime`).

Incorrect behavior before this fix: a `Fail` state that specified `ErrorPath`/`CausePath` instead of the static `Error`/`Cause` always reported `error = "States.Runtime"` and `cause = ""`, regardless of what the input actually contained, and a state machine definition that declared both `Error` and `ErrorPath` (or `Cause` and `CausePath`) on the same `Fail` state was accepted instead of rejected.

- `AslExecutor.executeFail` now resolves `ErrorPath` and `CausePath` against the state's input (JSONPath mode only, matching the ASL spec), reusing the same reference-path resolver a `.$` payload template field already uses (`resolvePayloadTemplateReference`, added for issue #2521). `ErrorPath`/`CausePath` can also be a `States.*` intrinsic per the ASL spec, which this resolver already supports.
- An unresolvable `ErrorPath`/`CausePath`, or one that resolves to a non-string, fails the state with `States.Runtime`, consistent with how an unresolvable JSONPath fails elsewhere in this executor.
- `CreateStateMachine`/`ValidateStateMachineDefinition` now reject a Fail state that declares both `Error` and `ErrorPath`, or both `Cause` and `CausePath`, matching real AWS.
- `ErrorPath`/`CausePath` are added to the JSONPath-only field list, since AWS only supports them under the JSONPath query language.

No model/DTO changes were needed: this codebase parses ASL state definitions as generic `JsonNode`, not into typed per-state-type classes, so `Fail` state fields are read directly off the definition JSON the same way `Error`/`Cause` already are.

## Checklist

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/AslExecutor.java
@@ -2039,9 +2039,31 @@ public class AslExecutor {
             if (cause != null && JsonataEvaluator.isExpression(cause)) {
                 cause = jsonataEvaluator.evaluateField(cause, "Cause", statesVar, variables).asText();
             }
+        } else {
+            if (stateDef.has("ErrorPath")) {
+                error = resolveFailDynamicField(stateDef.get("ErrorPath").asText(), "ErrorPath", input, context);
+            }
+            if (stateDef.has("CausePath")) {
+                cause = resolveFailDynamicField(stateDef.get("CausePath").asText(), "CausePath", input, context);
+            }
         }
         // AWS does not prefix a Fail state's Cause.
         throw new FailStateException(error, cause, true);
+    }
+
+    /**
+     * Resolves a Fail state's {@code ErrorPath} or {@code CausePath}: a reference path or a
+     * {@code States.*} intrinsic, evaluated against the state's input through the same resolver a
+     * {@code ".$"} payload template field uses. AWS requires the resolved value to be a string;
+     * an unresolvable path or a non-string result both fail the state with {@code States.Runtime},
+     * since a Fail state has no {@code Catch} of its own to route around either failure.
+     */
+    private String resolveFailDynamicField(String path, String field, JsonNode input, JsonNode context) {
+        JsonNode resolved = resolvePayloadTemplateReference(path, input, context, field, input);
+        if (!resolved.isTextual()) {
+            throw new FailStateException("States.Runtime", field + " must resolve to a string");
+        }
+        return resolved.asText();
     }
 
     private StateResult executeParallelState(String name, JsonNode stateDef, JsonNode input,
--- a/src/main/java/io/github/hectorvent/floci/services/stepfunctions/StepFunctionsService.java
+++ b/src/main/java/io/github/hectorvent/floci/services/stepfunctions/StepFunctionsService.java
@@ -72,7 +72,7 @@ public class StepFunctionsService implements Resettable, ResourceProvider {
     // AWS-observed, it is simply the one this code commits to.
     private static final List<String> JSONPATH_ONLY_FIELDS = List.of(
             "InputPath", "OutputPath", "ResultPath", "ResultSelector", "Parameters", "Result", "ItemsPath",
-            "MaxConcurrencyPath");
+            "MaxConcurrencyPath", "ErrorPath", "CausePath");
     // Fields that are valid only in JSONata mode. Validated against real AWS: a JSONPath state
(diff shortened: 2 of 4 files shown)


hectorvent:
Thanks, this is carefully done and the part I most expected to be wrong is right.

`States.Runtime` for a path that does not resolve matches a real-AWS capture: the recorded `executionFailedEventDetails` is `error: "States.Runtime"` with cause "The JSONPath '$.ErrorX' specified for the field 'ErrorPath' could not be found in the input" ([localstack snapshot](https://github.com/localstack/localstack/blob/main/tests/aws/services/stepfunctions/v2/scenarios/test_base_scenarios.snapshot.json)), and the same for `CausePath`. Your tests go further than the code needed to: they pin AWS's whole cause string, down to the state name, the entered-at event id and the offending input, not just the error code. Treating a non-string result the same way is the consistent reading.

Two more that check out: the mutual exclusion is structural in the ASL grammar, where `error_decl` and `cause_decl` are single alternatives rather than a pair of optional fields, and `ErrorPath` really is JSONPath-only, since `Error` accepts a JSONata expression and `ErrorPath` does not. So both additions to `JSONPATH_ONLY_FIELDS` are correct.

Mirroring `validateMapConcurrency` for the new pair, rather than inventing a second shape, is the right instinct too.

No blockers from my side.

hectorvent (approved):
Approving: no blockers.

PR #3306, opened 2026-09-09
feat(ec2): capture the instance file system in CreateImage, and implement Deregister/CopyImage

allensanborn:
## Summary

`CreateImage` records metadata only. It registers a new AMI id and copies the block device mappings, but nothing captures the source instance's file system, so an AMI created from a provisioned instance launches the base image with everything provisioned on it discarded.

This makes the capture real, and adds the two operations that manage the result.

- `CreateImage` commits the running container to a Docker image and stores it on the AMI, so `RunInstances` from that AMI starts from the captured file system.
- `DeregisterImage` and `CopyImage` are implemented; `RunInstances` from a deregistered AMI is rejected.
- Deregistering releases the captured Docker image rather than leaking it on the daemon.

## Type of change

- [x] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: an AMI created by `CreateImage` was not a snapshot of anything. Real EC2 captures the instance's volumes, so a launch from that AMI reproduces the source instance. Floci returned a plausible AMI id whose launches silently reverted to the base image, which is the failure mode that looks like success.

Upstream has since landed the metadata half of this area independently (the reboot, `findImageForCapture`, `captureBlockDeviceMappings`, and the `resolveLaunchableImageId` ancestry walk). This PR is the complementary half and builds on it: the file-system capture itself, plus Deregister and Copy. Verified before submitting that `commitInstance`, `removeCommittedImage`, `Image.dockerImage`, `case "DeregisterImage"` and `case "CopyImage"` are all still absent from `main`.

`CopyImage` copies within the emulator; it does not model cross-region semantics beyond the id and metadata, and the docs say so rather than implying parity.

## Checklist

Coverage: `Ec2CreateImageCaptureTest` 8, `Ec2DeregisterAndCopyImageTest` 19, `Ec2CapturedImageReclaimTest` 4. Load-bearing check: reverting `src/main` to `main` makes all three classes fail to compile, since the symbols they exercise do not exist there.

### One cross-PR note for whoever merges second

`Ec2CreateImageCaptureTest` constructs `Ec2ContainerManager` with the current 12-argument constructor. My other open PR #3272 replaces that constructor with a 13-argument form (adding `VpcNetworkManager`) without retaining a 12-arg overload. Each branch is correct against `main` today, and git merges them textually clean, so only `testCompile` catches the interaction.

Whichever of the two lands second will need a one-line adjustment to that test, passing the extra argument. I have deliberately not pre-adapted it here, because that would not compile against `main` as it stands. Happy to push the fix to whichever PR is second as soon as the first merges.


allensanborn:
All six addressed in 07401c73. Four were correctness bugs, and one of them made the feature's headline promise unreliable, so thank you for the catch.

**Copy drops the captured filesystem (P1)** - `CopyImage` carried the catalog ancestor but not the `dockerImage` reference, so a copied AMI launched the base image and silently lost everything provisioned. That is the exact failure this PR exists to remove, reintroduced on the copy path. The capture now survives the copy, and deregistering the source keeps a capture its copy still carries.

**Capture failure reported success (P1)** - a failed commit still produced an immediately-available AMI, so an image build could report success and quietly hand back the ancestor. It now fails loudly rather than presenting a ready AMI with nothing captured. Accepted-then-silently-wrong is the one option that cannot be right here.

**Two lock-ordering races (P1)** - both real. `RunInstances` now shares the registry lock for the resolve-and-publish step only, so a launch cannot resolve a captured tag and then have deregistration remove it before the instance is stored; and registration and deregistration are on the same monitor, so a snapshot cannot be reused between the shared-reference scan and the delete. The container launches themselves stay outside the lock.

One design note worth surfacing: that monitor is single rather than striped per region, deliberately. Docker image references are global to the daemon, and a `CopyImage` shares a layer with its source across regions, so a per-region lock would be unsound for the capture-release check specifically. The other two invariants would be fine striped; the layer one is what forces a single monitor.

**Failed cleanup forgot the capture (P2)** - `dockerImage` was cleared regardless of whether removal succeeded, so a transient daemon error leaked the layer permanently, with no path to rediscover the tag. The reference is now kept when removal did not succeed, and removal reports whether it actually happened.

**Retained captures were never reclaimed (P2)** - taken rather than deferred, since it turned out small: terminate now releases a capture once the last dependent instance goes away, instead of relying on a second deregistration that is rejected anyway.

**Collateral worth flagging, because it changed test fixtures rather than production code.** The wider `-Dtest=Ec2*` sweep surfaced 5 errors the three named test classes did not: several `Ec2ServiceTest` cases build a container-backed service with a bare `mock(Ec2ContainerManager)`, so their instances have no container and the new "nothing to capture" rejection fired. That is a fixture artifact - a running instance with no container cannot occur in container-backed mode - so the fix is a helper that stubs the commit, plus one resolver stub in `runInstancesOnACreatedImageResolvesTheSourceGuest` that would otherwise NPE now that a created AMI really does carry a capture. Calling it out explicitly so it does not read as loosening a test to make a change fit.

875 tests green across the full `Ec2*` sweep. Every one of the seven production changes was reverted in isolation and the matching test confirmed red before restoring; the mapping is in the commit body.

allensanborn:
Housekeeping so the thread list reflects reality: I have resolved the two threads at `Ec2Service.java:2335` ("Deregistration races with launch") and `:4581` ("Snapshot deletion races registration"). Both were fixed in 07401c73 and described in the comment above; the threads stayed open because a fix does not auto-resolve them, and four of the six from that round were marked outdated while these two were not, so they still read as live. Reopen either if you disagree with the fix rather than the bookkeeping.

The two genuinely new items from the 15:21 re-review are being worked and I have not resolved those:

- **P1 `Ec2Service.java:2517`, launch race orphans resources.** This one is distinct from the two above rather than a restatement: they were about the lock, this is about what happens when the second `requireNotDeregistered` check throws *after* root volumes, tags, the subnet IP allocation and possibly a caller-supplied ENI attachment have already been persisted. Nothing rolls those back.
- **P2 `Ec2CapturedImageReclaimTest.java:52`, integration test naming.** Fair, and the convention is real here rather than merely documented: 607 of the 736 `@QuarkusTest` classes on main are named `*IntegrationTest`. Renaming.

allensanborn:
Both new items done in c5aecfdb and 2d9b8670.

**Launch race (P1).** Confirmed before changing anything, and the enumeration matters because it shows why a rollback would have been the wrong shape. Between the early `requireNotDeregistered` and the second one, `runInstances` persists four things per instance: the root volume, the instance tags, a caller-supplied ENI written back with its attachment and status `in-use`, and a bump of the subnet IP counter, which is monotonic with no free list. So a late tombstone left an orphan volume, orphan tags, an ENI stuck in-use, and a permanently consumed address.

Rather than unwinding those on the failure path, the build loop now sits inside `synchronized (imageRegistryLock)` after the check, so nothing is persisted before the AMI is known good. Nothing to roll back is a better invariant than rolling back correctly, particularly for the IP counter, which has no release path to roll back into.

The critical section did not grow to cover the slow part: `containerManager.launch` stays outside the lock, with a comment saying why. The lock protects the registry and a reference into it, not the containers.

Coverage: `aTombstoneLandingMidLaunchLeavesNoOrphanedResources`. Load-bearing checked by reverting the production change, which fails it on `a rejected launch must not leave its root volume behind ==> expected: <0> but was: <1>`.

**Test naming (P2).** Renamed `Ec2CapturedImageReclaimTest` and `Ec2DeregisterAndCopyImageTest` to `*IntegrationTest`. I left `Ec2CreateImageCaptureTest` alone deliberately: it is a plain Mockito unit test with no `@QuarkusTest`, so the integration-test convention does not apply to it and renaming it would misdescribe what it is.

876 tests green across the full `Ec2*` sweep, up one from the added case.

Also worth flagging since it came up in the earlier round: I resolved the two threads at `Ec2Service.java:2335` and `:4581` after confirming both fixes are in the head. They had stayed open because a fix does not auto-resolve a thread, not because anything was outstanding.

hectorvent:
Thank you, and sorry this sat: you closed eight findings across two rounds without anyone replying, and the two `CopyImage` ones were the sort that would have quietly undone the point of the PR.

I found nothing to fix in the code. The shapes check out member by member against the model, including the nested `deleteSnapshotResultSet` wrapper, and every error code is a real EC2 one used in its documented sense.

**(blocking)** It needs a rebase, and there is a trap in it that is our fault. The only textual conflict is a one-line `ArgumentMatchers.any` import in `Ec2ServiceTest`; keep your side. But #3272 merged an hour ago and added a 13th constructor parameter to `Ec2ContainerManager`, and git auto-merges `Ec2CreateImageCaptureTest` cleanly because neither side touched those lines. So the break only appears at `test-compile`, and `./mvnw compile` still passes. Adding `mock(VpcNetworkManager.class)` at line 45 fixes it; I tried it and 193 tests pass afterwards, including #3272's and #3319's own.

allensanborn:
Rebased in d6ff91de, and your diagnosis was exact.

Merged current `main`. The only textual conflict was the `ArgumentMatchers.any` import in `Ec2ServiceTest` and I kept our side, since this branch's tests use it and `main` no longer does.

The constructor trap reproduces precisely as you described, and I checked it in both directions before touching anything rather than just applying the fix:

```
./mvnw compile       exit 0
./mvnw test-compile   exit 1   no suitable constructor found for Ec2ContainerManager(...)
```

`mock(VpcNetworkManager.class)` as the 13th argument fixes it. The full `Ec2*` sweep is 911 green, up from 876 as `main`'s own tests came across in the merge.

Worth saying that this one was foreseeable and I flagged it in the PR description when I opened this: #3272 replaced the 12-argument constructor without keeping an overload, git merges the test cleanly because neither side touches those lines, and only `test-compile` sees it. I offered then to push the fix to whichever PR landed second. That should have happened when #3272 merged rather than waiting for you to hit it, and I am sorry you spent time reproducing something already predicted in the thread.

Thank you for reviewing it shape by shape against the model. On the two `CopyImage` findings you singled out: the first was that `CopyImage` carried the catalog ancestor but not the `dockerImage` reference, so a copied AMI silently launched the base image, which is exactly the failure this PR exists to remove, reintroduced on the copy path.

hectorvent (approved):
That is it, and thank you for checking it in both directions rather than just applying the fix. The `compile` clean / `test-compile` broken pair is the whole reason that trap is worth naming: git had no conflict to show you, so nothing before test-compile could have caught it.

You are right that this should not have been yours to hit. You called the 12-argument constructor out in the description when you opened this and offered to push the fix to whichever PR landed second, and that offer sat in the thread unanswered while #3272 merged. The gap was on our side, not yours, and the same goes for the eight findings you closed across two rounds before anyone replied.

On the red shard, so it is on the record rather than waved away: there are two failures on it, not one, and both are our flakes rather than anything in your branch. `IotMqttWebSocketIntegrationTest.manyConcurrentWebSocketClientsEachReceiveThePublish` is fixed on main by #3431, and `CloudTrailLogWriterBoundedRetryIntegrationTest` by #3363. Your branch predates both, so a rerun would just re-roll the same two tests.

I am not going to ask you to rebase a third time to work around bugs of ours. Both fixes are on main, so the squashed result lands green there.

Merging.

PR #3307, opened 2026-09-09
fix(ec2): generate real key material in CreateKeyPair and fingerprint imported keys

allensanborn:
## Summary

`CreateKeyPair` returned a hardcoded placeholder as `KeyMaterial`:

```
-----BEGIN RSA PRIVATE KEY-----\nMIIEpA...FAKE\n-----END RSA PRIVATE KEY-----
```

It is a well-formed PEM envelope containing nothing usable, and every call returned the same one. `ImportKeyPair` was worse in a quieter way: it reported a constant all-zero fingerprint regardless of the key supplied, so two different imported keys were indistinguishable.

This generates a real 2048-bit RSA key pair per call, stores the public half in `ssh-rsa` framing, and fingerprints the actual key material on import.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `KeyMaterial` was not a key. Anything that parsed it failed, and anything that merely checked its shape passed, which is what let this survive. The same placeholder for every call also meant two key pairs were interchangeable, and a fingerprint that never varied meant `ImportKeyPair` could not distinguish the keys it was given.

Real EC2 returns a unique private key on create and a fingerprint derived from the key on import. Both are now true here. The material is a genuine PEM-encoded RSA private key, and two calls differ in material, public key and fingerprint.

## Checklist

Coverage: `Ec2KeyMaterialTest` (9) covers the helper; two cases in `Ec2ServiceTest` cover the wiring. 150 tests green across both classes.

Worth explaining why the second commit exists, since it is a test-only commit on a bug-fix PR. The first version of this change had 9 passing tests and was still not load-bearing: reverting the production code left the whole EC2 suite green, because the tests exercised the key-material helper in isolation and nothing covered the `Ec2Service` call sites that use it. The added tests go through `Ec2Service` and **parse** the returned PEM to a `RSAPrivateCrtKey` rather than pattern-matching its shape, precisely because the placeholder was a valid-looking envelope that any regex would accept. With the production change reverted they now fail with `malformed PEM data` and a constant all-zero fingerprint respectively.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2KeyMaterial.java
+++ b/src/main/java/io/github/hectorvent/floci/services/ec2/Ec2KeyMaterial.java
@@ -1,0 +1,150 @@
+package io.github.hectorvent.floci.services.ec2;
+
+import org.bouncycastle.openssl.jcajce.JcaPEMWriter;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.security.KeyFactory;
+import java.security.KeyPairGenerator;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.interfaces.RSAPublicKey;
+import java.security.spec.RSAPublicKeySpec;
+import java.util.Base64;
+import java.util.HexFormat;
+
+/**
+ * Real key material for EC2 key pairs.
+ *
+ * <p>CreateKeyPair is the only time AWS ever discloses a private key, and callers are
+ * expected to write the response straight to a file and use it: the Packer amazon-ebs
+ * builder, for one, creates a temporary key pair, saves the returned material, and SSHes
+ * in with it. A placeholder string cannot serve that, so the key is generated here.
+ *
+ * @see <a href="https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateKeyPair.html">CreateKeyPair</a>
+ * @see <a href="https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ImportKeyPair.html">ImportKeyPair</a>
+ */
+public final class Ec2KeyMaterial {
+
+    /** AWS generates 2048-bit RSA keys for KeyType=rsa, which is the default. */
+    private static final int RSA_KEY_SIZE = 2048;
+
+    private Ec2KeyMaterial() {}
+
+    /**
+     * @param privateKeyPem   PKCS#1 PEM, the "BEGIN RSA PRIVATE KEY" form AWS returns
+     * @param openSshPublicKey the matching "ssh-rsa AAAA..." line, for authorized_keys
+     * @param fingerprint     SHA-1 of the DER private key, colon-separated hex
+     */
+    public record Generated(String privateKeyPem, String openSshPublicKey, String fingerprint) {}
+
+    public static Generated generateRsa() {
+        try {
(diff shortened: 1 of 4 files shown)


allensanborn:
Fixed in abb58eba. One of the two holds, the other I think is half right, and the RSA half turns out to rest on two AWS sources that contradict each other, so I want to lay that out rather than just assert a verdict.

**Unbounded allocation (P1) - correct, fixed.** `SshReader.readBytes` did `new byte[buffer.getInt()]` with a length taken straight from the client-supplied blob, before `buffer.get` checked those bytes existed, and `OutOfMemoryError` is not caught by the surrounding `catch (Exception)`. The length is now validated against the remaining buffer before allocating, and an over-long field is rejected as invalid key material. Load-bearing: removing the bounds check aborts `Ec2KeyMaterialTest` with `OutOfMemoryError: Requested array size exceeds VM limit`, `Tests run: 0`.

**ED25519 fingerprint - correct, fixed.** AWS returns a base64-encoded SHA-256 digest for ED25519; the code MD5-hashed every key type. Now type-specific.

**RSA fingerprint - I do not think this half holds, and here is the awkward part.** The claim is that AWS fingerprints the SSH wire blob for imported RSA keys. AWS says both things:

- The API model (botocore `ImportKeyPairResult.KeyFingerprint`) says "the MD5 public key fingerprint as specified in section 4 of RFC 4716", which reads as the SSH blob.
- The User Guide's verify-keys page gives the reproduction command for an imported RSA key as `openssl rsa -in path_to_private_key -pubout -outform DER | openssl md5 -c`, which is MD5 over the DER SubjectPublicKeyInfo, not the SSH blob.

Those are different values. On a throwaway 2048-bit key here: DER route `7c:64:af:a7:51:a3:...`, SSH-blob route (`ssh-keygen -l -E md5`) `92:f0:47:07:52:bc:...`.

I went with the DER reading, because it is the one AWS publishes an exact reproduction command for, and because Java's `RSAPublicKey.getEncoded()` is already X.509 SubjectPublicKeyInfo DER, so the current code matches it. But I would rather flag the conflict than claim it is settled: if anyone has probed a real account and has the actual value, that beats both documents and I will change it.

Note this is also why `CreateKeyPair` and `ImportKeyPair` do not share a scheme: for AWS-generated RSA keys the model says SHA-1 over the DER private key, which is a third thing again.

**On the test criticism, which was fair regardless.** The old tests asserted digest shape or compared the service against the same helper, so they could not detect a scheme mismatch. The new ones pin known-good values computed independently, so they now fail if the scheme changes: disabling the DER conversion makes them fail with the MD5-of-blob value, which is precisely the mistake the review alleged was already present.

156 tests green across `Ec2KeyMaterialTest` (14, was 9) and `Ec2ServiceTest` (142), up from 150. Each fix reverted individually and the matching test confirmed red before restoring.

pgermosen (approved):
Pulled the actual AWS fingerprint docs to check both bot findings myself rather than taking either side at face value. The allocation fix is real and correctly bounded three ways. The RSA fingerprint one turned out to be a false positive on Greptile's part, AWS's own documented command confirms MD5 over the DER SubjectPublicKeyInfo is exactly right, so good call pushing back on that instead of just applying the suggestion. The ed25519 half of that same finding was real though, and SHA-256 with padding kept matches AWS's documented behavior exactly.

Also liked that the create-fingerprint hashes the native PKCS8 DER encoding even though the PEM handed back is deliberately PKCS1 to match what AWS returns, that's an easy pair of encodings to mix up and it doesn't happen here.

Good to merge.

PR #3308, opened 2026-09-09
test(ec2): cover security group rule sources at the Terraform layer

allensanborn:
## Summary

Adds compat-terraform coverage for security-group rules whose source is another security group, the case fixed by #2266.

`main` has Java-level coverage in `Ec2SecurityGroupRuleSourcesIntegrationTest` and an awscli case in `ec2.bats`, but nothing exercises this through the Terraform provider. That matters here because the failure this guards against is provider-visible: a rule losing its source group shows up as permanent plan drift rather than as an API error, so the layer that reads it back is the layer worth pinning.

Test-only. No production code.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No behaviour change. The fixture applies a security group with three rule shapes (source group by id, self-reference, and an all-protocols rule) plus a CIDR rule as a control, then asserts through both `DescribeSecurityGroups` and `DescribeSecurityGroupRules` that the source group survives the round trip and that the CIDR rule reports no `ReferencedGroupInfo`. It finishes with a re-plan asserting `No changes`, which is the assertion that would have caught the original drift.

## Checklist

The bats suite runs green: 6/6 against a running Floci with Terraform 1.15.8, after the usual `setup-bats` step to fetch `bats-core`, `bats-support` and `bats-assert` into `compatibility-tests/lib/` (gitignored, not included here). No Java is touched, so `./mvnw test` is unaffected by this change; I am ticking that box on the basis that nothing in it can move the Java suite rather than claiming a fresh full run.

No production change exists to revert, so there is no load-bearing proof to offer for this one. Saying that explicitly rather than leaving the impression it was checked the way a fix would be.

### DIFF
--- a/compatibility-tests/compat-terraform/test/security-group-rule-sources-tf/main.tf
+++ b/compatibility-tests/compat-terraform/test/security-group-rule-sources-tf/main.tf
@@ -1,0 +1,93 @@
+# Regression coverage for #2266 at the Terraform layer.
+#
+# "Allow traffic from security group X" is how one tier is wired to another -- an ALB
+# to its backend tasks, an application to its database, a cluster to itself. When
+# AuthorizeSecurityGroupIngress dropped the source, the provider's create-wait could
+# not find the rule it had just made and the apply failed with
+# "couldn't find resource"; a response that reported the rule but lost the reference
+# would instead apply cleanly and then show as drift on every subsequent plan.
+#
+# Both of the standard spellings are exercised, because they read the source back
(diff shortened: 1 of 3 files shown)


pgermosen (approved):
Good coverage of exactly the failure modes an API-level test can't reach - the apply itself failing, and the quieter drift-on-replan case. The provider.tf explanation for why this can't reuse the shared module is a nice touch too. Approving.

PR #3309, opened 2026-09-09
fix(dynamodb): read every attribute through an ALL-projection index in PartiQL

nvgoldin:
## Summary

A PartiQL read qualified with a secondary index returns the wrong item when the index projects `ALL`: `SELECT *` drops every attribute that is not part of a key, and naming one of those attributes explicitly is rejected as unprojected. Both are wrong answers rather than errors, and `Query` against the same index already returns the whole item, so the two read paths disagree.

### What it looks like

Table `queue`, partition key `id`, GSI `idx-queue-run_id` on `run_id` with `"ProjectionType": "ALL"`, holding one item.

```
SELECT * FROM "queue"."idx-queue-run_id" WHERE "run_id" = 'abc123'
```

Before, on `main` at 520568ef. `rows_inserted` and `status` are gone, with no error:

```json
{"Items":[{"id":{"S":"r1"},"run_id":{"S":"abc123"}}]}
```

After:

```json
{"Items":[{"id":{"S":"r1"},"run_id":{"S":"abc123"},"rows_inserted":{"N":"7"},"status":{"S":"done"}}]}
```

`Query` on that same index returns the full item both before and after, which is what makes the disagreement visible:

```json
[{"id":{"S":"r1"},"run_id":{"S":"abc123"},"rows_inserted":{"N":"7"},"status":{"S":"done"}}]
```

Naming a non-key attribute fails outright before the change:

```
SELECT rows_inserted FROM "queue"."idx-queue-run_id" WHERE "run_id" = 'abc123'
```

Before: `ValidationException: One or more parameter values were invalid: Global secondary index idx-queue-run_id does not project [rows_inserted]`

After: `{"Items":[{"rows_inserted":{"N":"7"}}]}`

### Why it happens

`DynamoDbAccessPath.projectedAttributeNames` builds its set from the table key schema, the index key schema, and, for an `INCLUDE` projection, `NonKeyAttributes`. It has no `ALL` arm, because the full attribute set of a schemaless table cannot be derived from its schema. For an `ALL` projection it therefore returns the key attributes and nothing else.

Every other caller already accounts for that by short-circuiting before it reads the set:

- `DynamoDbService.readItemSize`: `if (!accessPath.isIndex() || "ALL".equals(accessPath.projectionType()))`
- `DynamoDbJsonHandler.applyIndexProjection`: `if ("ALL_ATTRIBUTES".equals(select) || "ALL".equals(accessPath.projectionType()))`
- `DynamoDbAccessPathValidator.validateSelection`: `if (!accessPath.isGlobalSecondaryIndex() || "ALL".equals(accessPath.projectionType()))`

`DynamoDbPartiQLHandler.executeSelect` is the only caller without it, at both points where it uses the set: the unprojected-column rejection and the `SELECT *` trim. That is why `Query` and `Scan` are correct and `ExecuteStatement` is not.

### The change

Adds the same guard at those two points. `INCLUDE` and `KEYS_ONLY` indexes trim and reject exactly as before, and base-table reads never reached this code.

The existing coverage used only `INCLUDE` and `KEYS_ONLY` indexes, which is why neither case surfaced. `DynamoDbAccessPathIntegrationTest` gains an `ALL`-projection index and asserts `SELECT *`, an explicit non-key column, and `Query` over the same index agreeing; `DynamoDbPartiQLTest` in the Java SDK compatibility suite asserts the same two reads through the AWS SDK against the existing `ALL`-projection `status-index`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

`Projection.ProjectionType` documents `ALL` as "All of the table attributes are projected into the index", and notes that `ALL` projects "all attributes from the source table, even if the table has more than 100 attributes" ([API_Projection](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Projection.html), read 2026-09-09). The `Query` API reference is where the reachable-attribute rule is stated: a global secondary index query "can only request attributes that are projected into the index", which for an `ALL` projection is every attribute.

This change is derived from those documented projection semantics and from Floci's own `Query`, `Scan` and selection-validator behaviour against an `ALL` index. It is not backed by a fresh probe against a live account, so I have not quoted a characterisation run for it; the before and after above are against a packaged build of `main` and of this branch.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbPartiQLTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/DynamoDbPartiQLTest.java
@@ -349,4 +349,20 @@ class DynamoDbPartiQLTest {
                             .isEqualTo("NextToken does not match request");
                 });
     }
+
+    @Test
+    @Order(20)
+    void executeStatementReadsEveryAttributeThroughAllProjectionIndex() {
+        // status-index projects ALL, so a qualified read returns attributes that
+        // belong to no key, and naming one explicitly stays legal.
+        ExecuteStatementResponse all = ddb.executeStatement(r -> r
+                .statement("SELECT * FROM \"" + INDEX_TABLE + "\".\"status-index\" WHERE status = 'active'"));
+        assertThat(all.items()).hasSize(2);
+        assertThat(all.items().get(0)).containsKey("alternate");
+
+        ExecuteStatementResponse explicit = ddb.executeStatement(r -> r
+                .statement("SELECT alternate FROM \"" + INDEX_TABLE + "\".\"status-index\" WHERE status = 'active'"));
+        assertThat(explicit.items()).hasSize(2);
+        assertThat(explicit.items().get(0).get("alternate").s()).startsWith("alt-");
+    }
 }
--- a/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbPartiQLHandler.java
+++ b/src/main/java/io/github/hectorvent/floci/services/dynamodb/DynamoDbPartiQLHandler.java
@@ -56,7 +56,7 @@ class DynamoDbPartiQLHandler {
         // projects; the message preserves statement order (characterised on
         // real AWS, eu-west-1, 2026-09-02). An LSI read reaches the co-located
         // base item, so non-projected columns stay legal there.
-        if (accessPath.isGlobalSecondaryIndex()) {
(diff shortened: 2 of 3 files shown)


pgermosen (approved):
Checked the root cause directly rather than taking the description on faith, projectedAttributeNames really has no ALL arm, and the three call sites you point to as already handling this correctly do have exactly the guard you describe. The fix is a narrow, consistent application of an already-established pattern rather than something new.

Good test too, asserting SELECT *, an explicit non-key column, and Query all agree together means it can't pass by only fixing half the bug.

Good to merge.

PR #450, opened 2026-04-15
fix(lambda): accept name, partial ARN, and full ARN for {FunctionName}

hampsterx:
## Summary

Real AWS accepts bare name, partial ARN (`ACCT:function:NAME`), and full ARN (`arn:aws:lambda:REGION:ACCT:function:NAME`) for every endpoint that takes `{FunctionName}`, optionally suffixed with `:qualifier`. Floci only matched the stored short name, so every Lambda endpoint 404'd on ARN forms.

This change introduces `LambdaArnUtils.resolve()` as the single parse point and normalizes inputs in `LambdaService` at every user-facing method, so downstream store, warm-pool, code-store, and alias-store operations all use the canonical short name. `createFunction` now deduplicates across name/ARN forms.

## Behavior changes

- All existing Lambda endpoints now accept bare name, partial ARN, and full ARN (with or without `:qualifier`).
- Region mismatch between a full function ARN and the request region (or an event-source ARN region) returns `400 InvalidParameterValueException`.
- Tag endpoints (`ListTags`, `TagResource`, `UntagResource`) require an unqualified full function ARN; qualified ARNs and bare names are rejected with 400.
- URL config endpoints reconcile an ARN-embedded qualifier with the `?Qualifier=` query param and return 400 on conflict.
- `ListEventSourceMappings` normalizes its `FunctionName` filter so partial ARNs no longer miss stored mappings.
- Malformed inputs (blank, wrong service/partition, bad account segment, trailing junk, empty qualifier) now return 400 rather than best-effort parsing.

## Scope

Closes #434, split from #339. Credit to @okinaka for enumerating the affected endpoints.

Out of scope (follow-ups if wanted):
- Adding `?Qualifier=` support to endpoints that don't currently accept it (Invoke, GetFunction, AddPermission, etc.). This PR only fixes the 404; routing to a specific version/alias via qualifier is a separate feature.
- Region-aware `ListEventSourceMappings` filtering. Pre-existing behavior; unchanged here.
- Numeric-version qualifiers on Function URL config. Pre-existing limitation.

## Test plan

- [x] New `LambdaArnUtilsTest` with 31 parameterized cases covering every ARN form, qualifier placement, and malformed input class
- [x] New `LambdaServiceTest` cases covering `getFunction`/`deleteFunction`/`putFunctionConcurrency`/`createFunction` with bare name, partial ARN, full ARN, region mismatch, and ARN/name dedup
- [x] Full `./mvnw test` suite green: 2091 tests, 0 failures
[...]

PR #3310, opened 2026-09-09
docs(lambda): correct the FunctionName charset claim in CodeStore

avison9:
## Summary

Refs #3238. Corrects two comments I wrote in #3041 that misdescribe the codebase.

The Javadoc on `CodeStore.getVersionsPath`, and the matching comment in `CodeStoreTest`, both say
Floci "does not restrict the character set of `FunctionName` today, only that it is non-blank".
That is wrong. `LambdaArnUtils.validateName` has validated the name since #450:

```java
private static final Pattern NAME_PATTERN = Pattern.compile("[a-zA-Z0-9-_.]+");
private static final int MAX_NAME_LENGTH = 256;
```

The conclusion both comments draw is unaffected and still correct: a dot **is** an accepted
character, so `foo.v1` really is a function a user can create, and it really would have collided with
a `<name>.v<n>` sibling directory. That is why the `@versions` suffix exists and it should stay.
Only the stated reason was wrong.

It is worth correcting rather than leaving because of how it misleads. A reader who takes the comment
at face value goes looking for a validation that the comment says is not there, finds it, and now has
to work out which of the two to believe. Both comments now name the actual pattern and point at
\#3238, which tracks the divergence from the live service.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## AWS Compatibility

No behaviour change, so nothing moves toward or away from the live service here. The divergence the
corrected comments now describe accurately is real and is filed separately as #3238: Floci accepts
`FunctionName` values AWS rejects, specifically a dot in the name and a ceiling of 256 characters
against the live service's 64 for the name-only form. Measured against `main`, `foo.v1`, `..`, and a
200 character name are all accepted with `201 Created`. Changing that is a behaviour change with a
migration question attached, which is why it is an issue for a maintainer to direct rather than
something folded in here.

## How it was tested

Comment text only. No production statement, signature, or control flow is touched, and the diff is
entirely inside a Javadoc block and a `//` comment.

- `CodeStoreTest`: 11 run, 0 failures, 0 errors, on this branch
- A full `./mvnw test` is running locally as I open this; I will confirm the result in a comment
  rather than assert it here before it lands. CI covers it either way.

I re-checked `NAME_PATTERN` against this branch rather than relying on the earlier measurement, since
`CodeStore` has changed on `main` since #3041 and now carries a `region` parameter.

No new test. Per the project's Testing Policy, a change with no observable behaviour does not require
one, and there is nothing here to assert against.

## Checklist

- [ ] `./mvnw test` passes locally (affected class green, full run in progress, see above)

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/lambda/zip/CodeStore.java
+++ b/src/main/java/io/github/hectorvent/floci/services/lambda/zip/CodeStore.java
@@ -70,11 +70,12 @@ public class CodeStore {
      *
      * <p>The {@link #VERSIONS_DIR_SUFFIX} is what keeps that sibling from colliding with a real
      * function. {@link #sanitizeName} maps every name into {@code [a-zA-Z0-9_.-]}, so no function
-     * can produce a directory name containing {@code @}, whatever it is called. Floci does not
-     * restrict the character set of {@code FunctionName} today, only that it is non-blank, so a
-     * plainer {@code <name>.v<n>} sibling carried no such guarantee: a function genuinely named
-     * {@code foo.v1} owns the exact directory {@code foo}'s version 1 would otherwise claim, and
-     * deleting either one silently corrupts the other.
+     * can produce a directory name containing {@code @}, whatever it is called. A plainer
+     * {@code <name>.v<n>} sibling carried no such guarantee, because a dot is an accepted character
+     * in a function name: {@code LambdaArnUtils} validates against {@code [a-zA-Z0-9-_.]+}, which is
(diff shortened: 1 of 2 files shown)


pgermosen (approved):
Checked the claim against the actual source rather than just taking it on faith, LambdaArnUtils.validateName really does validate against that pattern with the 256 character ceiling, so the correction is accurate. Good instinct pointing at #3238 for the actual divergence instead of trying to fold a real behavior change into what should stay a comment fix.

Good to merge.

PR #3312, opened 2026-09-09
fix(s3): validate multipart completion parts

nkootstra:
## Summary

Validate the ETags and order in S3 CompleteMultipartUpload requests before assembling the object. Incorrect manifests no longer succeed with corrupted or unexpected content.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Floci previously reduced each completion request to part numbers. It ignored the submitted ETags and accepted duplicate or decreasing part order.

The controller now preserves each ordered `(PartNumber, ETag)` pair. The service requires strictly increasing part numbers and compares each submitted ETag with the uploaded part. It returns `InvalidPartOrder` for duplicate or decreasing order and `InvalidPart` for missing parts or ETag mismatches. Non-consecutive ascending part numbers remain valid.

Focused service and REST integration tests pass, including retry-after-failure and affected multipart fixture tests. The full `./mvnw test` suite was not run.

## Checklist


pgermosen (commented):
Real fix and the validation logic checks out precisely against AWS's actual documented error codes and messages for InvalidPart and InvalidPartOrder, word for word in places. Traced the ETag-quoting leniency in the second commit too, that's a reasonable, honestly-disclosed accommodation for a known friction point rather than an overclaimed parity fix.

Went through the ripple effect across the other test files to check for stragglers rather than assuming the sweep was complete, and found one call site in S3ConditionalWriteIntegrationTest still using a fake ETag. Traced it though: that test expects an If-Match precondition failure, which short-circuits before the new validation ever runs, so it's fine as is, not a gap.

The conflict itself is trivial once actually merged rather than just trusting the flag, S3Controller and S3Service both merge clean since this touches a different part of completeMultipartUpload than what landed in #2145. The only real collision is two unrelated PRs' new tests claiming the same @Order slot in S3MultipartIntegrationTest, which is just a renumber. Worth a rebase before this can land, but nothing to reconcile semantically.

nkootstra:
Updated the branch with the latest `main` and resolved the merge conflict in `S3MultipartIntegrationTest` by assigning unique `@Order` values to the multipart validation and SSE-KMS tests. The `S3ConditionalWriteIntegrationTest` fake ETag remains intentional because its `If-Match` precondition fails before multipart validation runs.

Focused verification passes:
- `S3MultipartServiceTest`: 32 tests
- `S3MultipartIntegrationTest`: 37 tests

pgermosen (approved):
Conflict's resolved and the renumbering looks clean, no dropped tests, no duplicate order values. Thanks for confirming the If-Match test's fake ETag is intentional, that matches what I traced through as well. Nothing else outstanding here.

PR #3316, opened 2026-09-09
feat(stepfunctions): add SNS Publish service integration

### LINKED ISSUE #2093: [FEAT] Step Functions (SFN): SNS Support
## Service

Step Functions (SFN)

## API Action / Feature

SNS publish support

## AWS Documentation

https://docs.aws.amazon.com/step-functions/latest/dg/connect-sns.html

## Why is this needed?

Common use-case

- [X] Yes
- [ ] No

matheusandre1:
## Summary

Adds the Step Functions SNS service integration described in
https://docs.aws.amazon.com/step-functions/latest/dg/connect-sns.html.

- `arn:aws:states:::sns:publish` (optimized) and `arn:aws:states:::aws-sdk:sns:publish`
  call the SNS Publish API through `SnsJsonHandler` and return the Publish response (`{MessageId}`).
- `.waitForTaskToken` works through the existing generic suffix handling; a `Message` given as an
  object (e.g. carrying `$$.Task.Token`) is serialized to its JSON text before publishing, as AWS does.
- Failures use the SDK exception class name: `SNS.NotFoundException` / `SNS.InvalidParameterException`
  on the optimized integration and `Sns.NotFoundException` on the `aws-sdk` one, so they are
  catchable via `Catch`/`Retry`.
- Both `Parameters` (JSONPath) and `Arguments` (JSONata) are supported.
- `SnsJsonHandler.Publish` now forwards `MessageGroupId` and `MessageDeduplicationId`; it was
  dropping them, which would have broken publishing to FIFO topics from a state machine
  (and from any JSON-protocol client).
- Docs: new "Publishing to SNS" section and an `aws-sdk:sns:publish` row in
  `docs/services/step-functions.md`.

Closes #2093

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Resource ARNs, parameter names (PascalCase `TopicArn`, `Message`, `Subject`, `MessageAttributes`,
`MessageGroupId`, `MessageDeduplicationId`, …), the `{MessageId}` result and the object-`Message`
serialization follow the AWS developer guide linked above. The optimized integration has no special
behavior beyond request/response and `.waitForTaskToken`, matching the guide. Error names follow
the `<Service>.<SdkExceptionClass>` convention already used by the `aws-sdk:*` integrations in this
repo.

Verified with the Step Functions JSON 1.0 protocol (`AWSStepFunctions.*` targets) and SQS/SNS
JSON protocols in the integration test; no new wire endpoints were added.

## Checklist

### DIFF
--- a/docs/services/step-functions.md
+++ b/docs/services/step-functions.md
@@ -357,10 +357,40 @@ the wire and the task fails with `Sfn.StateMachineDoesNotExistException`.
 | `arn:aws:states:::aws-sdk:sfn:sendTaskFailure` | `{}` | `Sfn.InvalidTokenException` |
 | `arn:aws:states:::aws-sdk:scheduler:createSchedule` | `{ScheduleArn}` | `Scheduler.ConflictException` when the name is taken |
 | `arn:aws:states:::aws-sdk:scheduler:updateSchedule` | `{ScheduleArn}` | `Scheduler.ResourceNotFoundException` |
+| `arn:aws:states:::aws-sdk:sns:publish` | `{MessageId}` | `Sns.NotFoundException` when the topic does not exist |
 
 `sendTaskSuccess` and `sendTaskFailure` resolve a token a `.waitForTaskToken` task is parked on. A
 token nobody is waiting for fails the calling task rather than reporting a delivery that never
 happened.
+
+## Publishing to SNS
+
+`arn:aws:states:::sns:publish` calls the SNS Publish API with the task's parameters and returns the
+Publish response, `{MessageId}`. `TopicArn`, `TargetArn`, `PhoneNumber`, `Message`, `Subject`,
+`MessageStructure`, `MessageAttributes`, `MessageGroupId` and `MessageDeduplicationId` are the API's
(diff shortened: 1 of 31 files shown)


hectorvent:
Thanks, and the detail I want to call out is the error prefix, because it is the thing that would have quietly made this wrong.

`sns:publish` and `aws-sdk:sns:publish` do not report failures the same way, and you have `"SNS."` for the first and `"Sns."` for the second. A real-AWS capture shows the optimized form producing `SNS.NotFoundException` and `SNS.InvalidParameterException` ([localstack snapshot](https://github.com/localstack/localstack/blob/main/tests/aws/services/stepfunctions/v2/services/test_sns_task_service.snapshot.json)), and it matches how the SQS integration in the same file already splits the two.

Routing through `SnsJsonHandler` rather than reimplementing Publish is the other good call: it means there is one implementation to keep correct rather than two that drift.

The non-string `Message` serialization with the `$$.Task.Token` case named in the javadoc is a nice touch too.

No blockers from my side.

hectorvent (approved):
Approving: no blockers.

PR #3317, opened 2026-09-09
feat(agentcore): expand Bedrock AgentCore control-plane coverage

KenkoGeek:
## Summary

Expands Amazon Bedrock AgentCore emulation with 30 additional control-plane operations across Browser, Browser Profile, Code Interpreter, Gateway Rule, API key credential provider, OAuth2 credential provider, and Resource Policy families.

The implementation follows the current AWS Bedrock AgentCore Control API Reference and AWS SDK for Java v2 wire contracts, preserves Floci's Controller -> Service -> Storage architecture, uses account-aware StorageFactory persistence, and adds AWS SDK compatibility coverage for the new behavior.

The final senior review also tightened documented validation and idempotency behavior for the new operations, including nested Gateway Rule unions, VPC network configuration, tag types, and delete clientToken replays.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against the current Amazon Bedrock AgentCore Control API Reference and AWS SDK for Java v2 `2.52.0` request marshallers.

The added actions use the documented REST-JSON methods, paths, success status codes, required members, identifier constraints, ARN shapes, pagination inputs, and error semantics for their respective operation families.

AWS CLI was not used for wire-protocol verification. Compatibility coverage uses the AWS SDK for Java v2 client against Floci.

## Checklist

Validation performed locally:

- `./mvnw -q -DskipTests test-compile`
- `../../mvnw -q -DskipTests test-compile` from `compatibility-tests/sdk-test-java`
- `make docs-check`
- `git diff --check origin/main...HEAD`

The full local Maven test suite was not run because it is too heavy for the available machine. CI is expected to run the complete upstream test matrix.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/BedrockAgentCoreCredentialProviderTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/BedrockAgentCoreCredentialProviderTest.java
@@ -52,4 +52,16 @@ class BedrockAgentCoreCredentialProviderTest {
         assertThat(response.apiKeySecretArn().secretArn()).contains(":secretsmanager:");
         assertThat(response.apiKeySecretJsonKey()).isEqualTo("apiKey");
     }
+
+    @Test
+    @Order(2)
+    void getApiKeyCredentialProvider() {
+        var response = client.getApiKeyCredentialProvider(builder -> builder.name(apiKeyProviderName));
+
+        assertThat(response.name()).isEqualTo(apiKeyProviderName);
+        assertThat(response.apiKeySecretSourceAsString()).isEqualTo("MANAGED");
+        assertThat(response.credentialProviderArn()).contains(":apikeycredentialprovider/");
+        assertThat(response.createdTime()).isNotNull();
+        assertThat(response.lastUpdatedTime()).isNotNull();
+    }
 }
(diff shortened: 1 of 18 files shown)


hectorvent:
Thank you for this, and sorry it sat unreviewed: nobody had looked at it at all, which is not the reception a 34-commit design-documented PR deserves.

I checked the routes against the model rather than skimming them, and the three that are easy to get backwards are all right. `Create` being `PUT` while `List` is `POST` on the same path, across browsers, browser profiles and code interpreters, is the one most additions get wrong by reaching for a POST create. `UpdateGatewayRule` really is `PATCH`. And `{resourceArn:.+}` is necessary rather than decorative, since a bare `{resourceArn}` stops at the first slash in the ARN.

One note rather than a request: `GetBrowser` and `GetCodeInterpreter` are the only operations here whose declared errors omit `ValidationException`, which you throw on a bad id. I nearly raised it, then checked the neighbours: `DeleteBrowser` and `DeleteCodeInterpreter` take the same parameter with the same pattern and do declare it. That reads as an inconsistency in AWS's model, not a distinction.

No blockers from my side.

hectorvent (approved):
Approving: no blockers.

PR #3318, opened 2026-09-09
fix(cloudfront): report Logging on distribution reads

allensanborn:
## Summary

`xmlDistributionConfig` never emits the `Logging` member, so `aws_cloudfront_distribution` sees permanent drift on `logging_config`: Terraform sends it, Floci stores it, and the read back omits it, so every plan shows a change.

This emits it on distribution reads.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `DistributionConfig` was missing a member the provider reads on every plan. Real CloudFront returns `Logging` with its `Enabled`, `IncludeCookies`, `Bucket` and `Prefix` fields; Floci returned the surrounding config without it.

Scope note, since an earlier revision of this branch was wider: it also added `Restrictions` and `OriginGroups`. Both have since landed on `main` independently, so those hunks were dropped rather than duplicated. Only `Logging` remains, which `main` still does not emit.

## Checklist

57 tests green across `CloudFrontDistributionConfigMembersTest`, `CloudFrontControllerTest` and `CloudFrontServiceTest`. Load-bearing check: removing the single `xml.raw(xmlLogging(cfg.getLogging()))` line fails the member test and nothing else. The test also asserts `Restrictions` and `OriginGroups`, which `main` already emits, so it guards those alongside the new member rather than only the added one.

### DIFF
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
@@ -1843,8 +1843,27 @@ public class CloudFrontController {
 
         xml.raw(xmlViewerCertificate(cfg.getViewerCertificate()));
         xml.raw(xmlRestrictions(cfg.getGeoRestriction()));
+        xml.raw(xmlLogging(cfg.getLogging()));
 
         return xml.build();
+    }
+
+    /**
+     * Logging is always present on a real DistributionConfig response, disabled by
+     * default. Callers read it unconditionally, so omitting it when the caller did not
+     * supply one leaves the member missing from every read.
+     */
+    private String xmlLogging(Map<String, Object> logging) {
+        return new XmlBuilder()
+                .start("Logging")
+                .elem("Enabled", logging != null
+                        && Boolean.parseBoolean(str(logging.get("Enabled"))))
+                .elem("IncludeCookies", logging != null
+                        && Boolean.parseBoolean(str(logging.get("IncludeCookies"))))
+                .elem("Bucket", logging != null ? str(logging.get("Bucket")) : "")
+                .elem("Prefix", logging != null ? str(logging.get("Prefix")) : "")
+                .end("Logging")
+                .build();
     }
 
     private String xmlEmptyOriginGroups() {
--- a/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/cloudfront/CloudFrontController.java
@@ -2508,6 +2508,7 @@ public class CloudFrontController {
         cfg.setViewerCertificate(parseViewerCertificate(body));
         cfg.setCustomErrorResponses(parseCustomErrorResponses(body));
         cfg.setGeoRestriction(parseGeoRestriction(body));
+        cfg.setLogging(parseLogging(body));
 
         return cfg;
     }
@@ -2639,6 +2640,20 @@ public class CloudFrontController {
             LOG.debugv("Ignoring malformed CustomErrorResponses during parse: {0}", e.getMessage());
         }
         return result;
+    }
+
+    /**
+     * Parses the optional {@code Logging} block into {@code Enabled}, {@code IncludeCookies},
+     * {@code Bucket} and {@code Prefix} (values kept as strings). A request that omits the block
+     * yields an empty map rather than {@code null}, which {@link #xmlLogging} renders as the
+     * disabled defaults CloudFront reports for a distribution that never asked for access logs.
+     */
+    private Map<String, Object> parseLogging(String body) {
+        List<Map<String, String>> groups = XmlParser.extractGroups(body, "Logging");
+        if (groups.isEmpty()) {
+            return Map.of();
+        }
+        return new LinkedHashMap<>(groups.getFirst());
     }
 
     private Map<String, Object> parseGeoRestriction(String body) {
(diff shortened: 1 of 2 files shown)


allensanborn:
Good catch, and the more important half of it is the part about the test. Fixed in f1ab7941.

**The parser.** You were right that `Logging` was write-never: `parseDistributionConfig` set sixteen members and that was not one of them, and `setLogging` appeared nowhere in the controller. So the PR emitted a member the write path never populated, which does not actually fix the drift it claims to fix: Terraform sends `logging_config`, Floci drops it, and the read returns the disabled defaults. `parseDistributionConfig` now calls a `parseLogging(body)` written alongside the existing `parseGeoRestriction` and `parseViewerCertificate`, using the same `XmlParser.extractGroups` shape. An absent `Logging` block yields an empty map, so the defaults are unchanged.

**The test observation was the sharper point, and it reproduces exactly as you described.** With the parser line reverted and the test file untouched:

- the three original tests all still PASS, because they build a `DistributionConfig`, call `setLogging` in Java, and insert it through `CloudFrontService`, never touching the parser
- the two new tests, which drive `CreateDistribution` and `UpdateDistribution` over the request path, both FAIL

So the original tests could never have caught this. That is the same shape as a hole found on another PR of mine this week, where nine passing tests exercised a helper in isolation while nothing covered the wiring that called it. Worth me watching for as a pattern rather than treating each instance as a one-off.

209 tests green across the `CloudFront*` sweep; the members test is now 6, three old plus three new.

**On the field-injection P2, I would rather not change it, and the reason is in AGENTS.md itself.** Line 326 does say "Use constructor injection". Four lines later, 330 says "Follow existing project patterns". Those point in opposite directions here: there are 164 `@Inject` field occurrences across `src/test/java` and no test class in the repository uses constructor injection, including all ten CloudFront ones. Making this the single exception would satisfy the first rule by breaking the second, and would leave the next person reading the CloudFront tests with one that looks unlike its neighbours. Happy to switch it if you would rather the new tests lead a change in convention, but I did not want to start that unilaterally in a bug-fix PR.

allensanborn:
On the field-injection P2, I would rather leave it, and I want to give the reasoning rather than just decline.

`AGENTS.md:326` does say "Use constructor injection", and it sits under `## Code Style` alongside guidance clearly aimed at production classes ("Prefer self-explanatory code over comments", "Always use braces in conditionals"). Four lines below it, `:330` says "Follow existing project patterns". For a test class those two point in opposite directions, and the repository has already chosen: of the 736 `@QuarkusTest` classes on `main`, 119 use field `@Inject` and **none** uses constructor injection. All ten CloudFront test classes are in that first group.

There is also a mechanical reason the pattern is what it is. JUnit 5 constructs test classes itself and will only pass constructor arguments that a registered `ParameterResolver` supplies. Quarkus does not register one for arbitrary CDI beans in a `@QuarkusTest`, so a constructor taking `CloudFrontService` is not a style choice that happens to be unpopular here; it is not the supported wiring. That is why the count is 0 of 736 rather than merely low.

So making this file the exception would satisfy `:326` by breaking `:330`, and would leave one CloudFront test looking unlike its nine neighbours for no behavioural gain. If the project does want to move tests toward constructor injection I am happy to help, but that is a repo-wide change with a mechanical question to answer first, and not something to start unilaterally inside a bug-fix PR.

Happy to be overruled if a maintainer would rather these new tests lead that change.

pgermosen:
Verified the Logging shape and the always-present-on-read behavior against the real CloudFront model, matches exactly, and good catch fixing the write side too since the read fix alone wouldn't have been enough.

One thing left before merging: CloudFrontDistributionConfigMembersTest injects CloudFrontService into a field instead of through the constructor, which AGENTS.md asks for. Small fix, but worth doing before this lands.

allensanborn:
Done in dc9fcacc, and I owe you a correction rather than just a fix.

`CloudFrontDistributionConfigMembersTest` now takes `CloudFrontService` through the constructor. 209 CloudFront tests still pass.

I argued against this earlier, and my main reason was factually wrong. I said a `@QuarkusTest` cannot take constructor injection, because JUnit 5 only passes constructor arguments that a registered `ParameterResolver` supplies and I believed Quarkus does not register one for arbitrary CDI beans. Before making the change I tested that claim with a throwaway `@QuarkusTest` whose constructor takes `CloudFrontService`. It compiles and passes. So the count I cited, no constructor injection in any of the repository's 736 `@QuarkusTest` classes, is habit rather than a constraint, and I presented a habit as a mechanical impossibility.

That is the kind of claim that is tedious to disprove, and you asked for the change anyway rather than taking my word for it, which was the right call.

The rest of the count stands as context if it is ever useful: 164 `@Inject` field occurrences across `src/test/java`, all ten CloudFront test classes among them. If constructor injection is the direction you want tests to move, this one can be the first and I am happy to convert the other CloudFront classes in a separate PR rather than widening this one.

Thanks also for checking the Logging shape against the model. The write-side gap was Greptile's catch, and the sharper half of it was noticing that the original tests could not have caught it, since they inserted the model through `CloudFrontService` and never touched the parser.

pgermosen (approved):
Both fixed and confirmed. Checked LoggingConfig's actual shape fresh, the four members and disabled-default behavior match exactly, and confirmed ListDistributions' DistributionSummary correctly doesn't need the same treatment since it has no Logging member in the real API, so nothing else is missing here. Also, nice catch testing your own claim on the constructor-injection question rather than just taking the pushback at face value. Approving.

PR #3319, opened 2026-09-09
feat(ec2): answer MonitorInstances and UnmonitorInstances

allensanborn:
## Summary

`MonitorInstances` and `UnmonitorInstances` were not dispatched at all, so both returned an unsupported-operation error. `aws_instance` with `monitoring = true` calls `MonitorInstances` directly, and one unsupported action fails a whole apply regardless of how much of the rest succeeded.

This answers both actions, resolves the instances they name, and stores the state so a later read agrees.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Detailed monitoring is a CloudWatch billing switch with no emulated behaviour behind it, so what matters is that the calls are answered honestly rather than that anything is measured.

Two things the first draft of this got wrong, both caught in review on the fork before this PR was opened:

- It echoed every `InstanceId` back with a 200 without resolving it, and did not take a region to resolve one with, so naming an instance that does not exist returned success saying its monitoring was enabled. Every other instance operation raises `InvalidInstanceID.NotFound`; this now does too, resolving through the same `getRequiredInstance` path `stopInstances` and friends use.
- It never stored the state. `Instance` already carries a `monitoring` member and `DescribeInstances` already emits it, so `MonitorInstances` answered "enabled" while every later read still said "disabled". The state now persists, so the round trip agrees with itself.

## Checklist

`Ec2MonitorInstancesTest` is 5 tests: enable, disable, a `DescribeInstances` round trip, and both not-found paths. The earlier version of this class asserted the old behaviour, passing a made-up instance id and expecting 200, so it was rewritten to launch a real instance rather than extended.

Both fixes checked load-bearing independently: dropping the validation fails the two not-found tests, and keeping validation while not storing fails exactly the `DescribeInstances` round-trip test. 837 tests green across the full `Ec2*` sweep.

### DIFF
--- a/compatibility-tests/sdk-test-awscli/test/ec2.bats
+++ b/compatibility-tests/sdk-test-awscli/test/ec2.bats
@@ -12,9 +12,13 @@ setup() {
     TGW_ATTACHMENT_ID=""
     TGW_VPC_ID=""
     TGW_ROUTE_TABLE_ID=""
+    MONITOR_INSTANCE_ID=""
 }
 
 teardown() {
+    if [ -n "$MONITOR_INSTANCE_ID" ]; then
+        aws_cmd ec2 terminate-instances --instance-ids "$MONITOR_INSTANCE_ID" >/dev/null 2>&1 || true
+    fi
     if [ -n "$PREFIX_LIST_ID" ]; then
         aws_cmd ec2 delete-managed-prefix-list --prefix-list-id "$PREFIX_LIST_ID" >/dev/null 2>&1 || true
     fi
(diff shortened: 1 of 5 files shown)


allensanborn:
Both addressed, one by complying and one by pushing back, and the difference is the evidence rather than the priority.

**Integration test naming: you are right, renamed in ba513da2.** I checked what the repository actually does rather than just the directive, and here they agree: 607 of the 736 `@QuarkusTest` classes on `main` are named `*IntegrationTest`. This class drives the HTTP endpoint, so it belongs in that 607. Now `Ec2MonitorInstancesIntegrationTest`, 5/5 still green.

Worth saying explicitly, because I disputed a naming finding on another PR of mine this week: there the numbers ran the other way, 100 of 1247 files matching the claimed rule against 1147 not, so the "requirement" was not the practice. Here the practice is real and I am following it.

**SDK coverage: I do not think this one applies to this test.** The directive you are citing lives under `## Compatibility Project` in AGENTS.md, whose first line is "Compatibility test suite: `./compatibility-tests/`", and its two bullets are "Prefer AWS SDK clients over raw HTTP for management-plane validation" and "Use this suite when changes may affect real SDK behavior". It is guidance about that suite, not a rule for `src/test`.

The practice matches that reading: of the EC2 tests under `src/test/java/.../services/ec2/`, 34 use RestAssured and 0 use the AWS SDK. Rewriting this one in SDK form would make it the only one, which is not the convention it is being held to.

There is a fair version of the point, from AGENTS.md:110, "Management APIs should be validated with AWS SDK clients, not only handcrafted HTTP requests" - the "not only" being the operative part. If you would like SDK-level coverage for `MonitorInstances` and `UnmonitorInstances`, the right home is `compatibility-tests/sdk-test-*` rather than this class, and I am happy to add it there as a follow-up. I would rather not widen this PR to do it, since it is a different suite with its own runner.

allensanborn:
I take the SDK point back: you were right and my earlier scoping argument was too narrow. Added in 2de5e1c3.

I had argued the SDK preference lives under `## Compatibility Project` and so applies to `./compatibility-tests/` rather than to this class. That part is true, but AGENTS.md:110 says separately that "Management APIs should be validated with AWS SDK clients, not only handcrafted HTTP requests", and the operative words are "not only". Coverage had to exist somewhere, and pointing at the right suite while not writing anything in it is not an answer.

`compatibility-tests/sdk-test-awscli/test/ec2.bats` is where EC2 SDK coverage already lives, so three cases go there: enable and read back through `DescribeInstances`, disable and read back, and reject an instance that does not exist. The read-backs are the point rather than padding, because the bug this guards against was the handler answering "enabled" while every later read still said "disabled", which asserting on the mutating call's own response cannot detect.

The Java test stays as it is. It pins the exact Query XML, which the SDK cases do not, so the two cover different things.

**On verification, being precise about what I did and did not confirm.** `run-instances`, the teardown and the harness plumbing all work against a local emulator, and fixing them found a real error in my first draft: I had written `--min-count`/`--max-count`, which the CLI rejects, and only running it showed that. The three assertions themselves cannot pass on this machine, because the released `floci/floci:latest` 2.0.1 image predates `MonitorInstances` and answers `UnsupportedOperation`, which is the gap this PR closes. They will run against the branch build in CI, and that is where their result should be read.

pgermosen (approved):
Checked the fix directly, both corrections hold up: setInstanceMonitoring resolves through the same getRequiredInstance path stopInstances uses, so a nonexistent instance correctly gets InvalidInstanceID.NotFound instead of a false 200, and the state actually persists so DescribeInstances agrees with it afterward. The CLI-vs-SDK-client coverage point is a fair methodology call rather than a real gap, botocore backs the CLI either way. Approving.

PR #3320, opened 2026-09-09
fix(compat): avoid CloudWatch statistics time race

nkootstra:
## Summary

Make the CloudWatch statistics compatibility test use an end time captured after `PutMetricData` completes. This prevents a one-second boundary race from producing an empty datapoint response and blocking unrelated pull requests.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The test previously captured `EndTime` before sending `PutMetricData`. If the request crossed a second boundary, Floci assigned the metric a timestamp later than that end time, so `GetMetricStatistics` returned no datapoints. The test now captures the end time after the write completes. This changes test timing only and does not change Floci's AWS wire behavior.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-node/tests/cloudwatch.test.ts
+++ b/compatibility-tests/sdk-test-node/tests/cloudwatch.test.ts
@@ -77,8 +77,7 @@ describe('CloudWatch Metrics', () => {
   });
 
   it('should put metric data with statistic values', async () => {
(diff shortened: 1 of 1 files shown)


pgermosen (approved):
Simple, correct timing fix for a real race. Approving.

PR #3321, opened 2026-09-09
fix(compat): require Kinesis stream details

nkootstra:
## Summary

Make the Kinesis Go compatibility test fail when `DescribeStream` omits the stream description, shards, shard ID, or stream ARN. Previously, missing fields could skip dependent checks and allow an invalid response to pass.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

The test now validates the required `DescribeStream` response fields before using them. This improves compatibility coverage only and does not change Floci production behavior or AWS wire responses.

## Checklist


pgermosen (approved):
Straightforward tightening of the test, closes a real gap where a missing shard or stream ARN could silently skip dependent checks instead of failing. Approving.

PR #3324, opened 2026-09-09
feat(s3): add object annotations support

### LINKED ISSUE #3239: [FEAT] Support S3 object annotations
## Service

S3

## API Action / Feature

S3 Object Annotations: named payloads (1 byte to 1 MiB) attached to object
versions, managed through dedicated API operations without re-uploading the
object. Up to 1,000 annotations per object version.

None of the annotation surface is implemented today. `S3Controller` handles the
`tagging`, `retention`, `legal-hold`, `acl`, `attributes`, `uploadId`, and
`versionId` subresources, but there is no `?annotation` subresource anywhere in
the S3 service, model, tests, or docs.

### Core operations (missing)

**PutObjectAnnotation**
PUT /{Key+}?annotation&annotationName={name}&versionId={vid}
- Body: annotation payload (raw UTF-8 text, 1 byte to 1 MiB)
- Headers: `x-amz-object-if-match` (validates the parent object ETag),
  `x-amz-sdk-checksum-algorithm` (CRC32, CRC32C, CRC64NVME, SHA1, SHA256,
  SHA512, MD5, XXHASH64, XXHASH3, XXHASH128), per-algorithm `x-amz-checksum-*`
  headers, `Content-MD5`, `x-amz-expected-bucket-owner`, `x-amz-request-payer`
- Response: 200 with `ETag` (the annotation ETag, distinct from the object
  ETag), `x-amz-object-version-id`, checksum headers, `x-amz-checksum-type`,
  `x-amz-server-side-encryption`; XML body
  `PutObjectAnnotationOutput { Key, AnnotationName }`
- Errors: `AnnotationLimitExceeded` (400), `AnnotationNameTooLong` (400),
  `InvalidAnnotationName` (400), `InvalidRequest` (400),
  `NoSuchBucket`/`NoSuchKey` (404), `UnsupportedMediaType` (415, payload not
  valid UTF-8)

**GetObjectAnnotation**
GET /{Key+}?annotation&annotationName={name}&versionId={vid}
- Header: `x-amz-checksum-mode: ENABLED` to receive checksum headers
- Response: 200 with the payload as body, `Content-Length`, `ETag`,
  `Last-Modified`, checksum headers, `x-amz-object-version-id`,
  `x-amz-replication-status`, `x-amz-server-side-encryption`
- Errors: `NoSuchAnnotation` (404), `NoSuchBucket`/`NoSuchKey` (404)

**ListObjectAnnotations**
GET /{Key+}?annotation&annotation-prefix={prefix}&continuation-token={tok}&max-versionId={vid}
- Response XML: `ListObjectAnnotationsOutput` containing
  `Annotations/AnnotationEntry { AnnotationName, ChecksumAlgorithm, ETag,
  LastModified, ReplicationStatus, Size }` plus `Bucket`, `Key`,
  `AnnotationPrefix`, `MaxAnnotationResults`, `AnnotationCount`,
  `ContinuationToken`, `NextContinuationToken`
- Errors: `InvalidPrefix` (400), `NoSuchBucket`/`NoSuchKey` (404)

**DeleteObjectAnnotation**
DELETE /{Key+}?annotation&annotationName={name}&versionId={vid}
- Headers: `x-amz-object-if-match`, `x-amz-bypass-governance-retention`
  (Object Lock governance mode)
- Response: 204 with `x-amz-object-version-id`, `x-amz-request-charged`
- Errors: `NoSuchBucket`/`NoSuchKey` (404). Deleting a nonexistent annotation
  is not an error.

### Behaviors to emulate

- Limits: 1,000 annotations per object version; annotation name up to 512
  bytes, restricted to letters, digits, `_`, `.`, `-`, and must not start with
  `aws` or `s3` (case-insensitive); payload 1 byte to 1 MiB; UTF-8 text only
- Versioning: annotations attach to a specific object version; a new version
  does not inherit them; overwriting the object replaces them; a simple DELETE
  preserves annotations on the underlying version while deleting a specific
  version ID deletes them; deleting an annotation is permanent, there are no
  annotation delete markers
- Adding, updating, or removing an annotation does not change the object ETag
- Annotations on SSE-C encrypted objects are rejected

### Adjacent integration points

- `CopyObject`: annotations copy by default; support
  `x-amz-annotation-directive: COPY | EXCLUDE` (currently only
  `x-amz-metadata-directive` exists)
- Event notifications: `s3:ObjectAnnotation:Put` and `s3:ObjectAnnotation:Delet
  (only `ObjectCreated:*`/`ObjectDeleted` are fired today)
- IAM actions: `s3:PutObjectAnnotation`, `s3:GetObjectAnnotation`,
  `s3:ListObjectAnnotations`, `s3:DeleteObjectAnnotation`

### Out of scope for this issue (candidates for follow-ups)

- S3 Metadata annotation tables (`CreateBucketMetadataConfiguration`,
  `UpdateBucketMetadataAnnotationTableConfiguration`,
  `GetBucketMetadataConfiguration`): Floci has no S3 Metadata configuration
  support at all, so this is a separate, larger feature
- Annotation replication (`s3:ReplicateObjectAnnotation`,
  `s3:GetObjectVersionAnnotationForReplication`)

## AWS Documentation

- Overview: https://docs.aws.amazon.com/AmazonS3/latest/userguide/annotations-o
- PutObjectAnnotation: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectAnnotation.html
- GetObjectAnnotation: https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetO
- ListObjectAnnotations: https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectAnnotations.html
- DeleteObjectAnnotation: https://docs.aws.amazon.com/AmazonS3/latest/API/API_D

## Why is this needed?

I want to use floci as a CI provider for S3, but my service relies on object annotation to store metadata (the `x-amz-meta-...` are not enough to store the JSON I have that needs to be stored together with the object).

Without S3 annotation support, I cannot use floci without accepting some test blind spots.

- [x] Yes
- [ ] No

rafaelsierra:
## Summary

Adds S3 Object Annotations: named UTF-8 payloads (1 B to 1 MiB, up to 1,000 per object version) attached to object versions without re-uploading the object, matching the AWS annotations API. Closes #3239.

- PutObjectAnnotation, GetObjectAnnotation, ListObjectAnnotations, DeleteObjectAnnotation via the `?annotation` subresource (REST XML), including `x-amz-object-if-match`,`x-amz-checksum-mode`, checksum validation, and pagination with continuation tokens
- Annotations are per-object-version: new versions do not inherit them, overwrites and non-versioned deletes drop them, delete markers preserve them
- CopyObject copies annotations by default; `x-amz-annotation-directive: COPY | EXCLUDE`
- `s3:ObjectAnnotation:Put` / `s3:ObjectAnnotation:Delete` notification events and CloudTrail events; IAM actions enforced when auth is enabled
- Storage: annotation metadata in a new `s3-annotations.json` store; payload bytes as `.s3ann` blobs (object-body pattern), keeping payloads out of the single-document serialization of the metadata stores

Known follow-ups (not in this PR): S3 Metadata annotation tables, annotation replication, XXH*/SHA512/MD5 checksum algorithms, O(total-annotations) scan on annotation-heavy stores, and the residual non-atomicity window between annotation cleanup and the replacement metadata publish on overwrite. Note: DeleteObject now takes the bucket monitor, so notification delivery serializes with other writes on the same bucket.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Verified against AWS SDK for Java v2 2.52.0 (compat suite, `S3AnnotationsTest`, 11/11 passing
against a running Floci container): SDK round-trips for put/get/list/delete-object-annotation, checksum headers with `checksumMode(ENABLED)`, aws-chunked/streaming-trailer PUT framing, the modeled `NoSuchAnnotationException`, versioned-bucket per-version attachment, and the `x-amz-object-annotation-directive` COPY/EXCLUDE directive on CopyObject (the real header name, which this PR also accepts alongside the `x-amz-annotation-directive` spelling).

Verified with AWS CLI 2.36.41: put/get/list/delete-object-annotation round-trip, including
the CRC64NVME default checksum. The existing `S3Test` compat suite still passes.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ For operation-level compatibility, see the [Services Overview](https://floci.io/
 | SSM | In-process + EC2 containers | Parameter Store (version history, labels, SecureString, tagging); Run Command (SendCommand, GetCommandInvocation, direct EC2 container execution, agent polling) |
 | SQS | In-process | Standard and FIFO queues, DLQ, visibility timeout, batch operations, tagging |
 | SNS | In-process | Topics, subscriptions, SQS, Lambda and HTTP delivery, tagging |
-| S3 | In-process | Versioning, multipart upload, pre-signed URLs, Object Lock, event notifications |
+| S3 | In-process | Versioning, multipart upload, pre-signed URLs, Object Lock, object annotations, event notifications |
 | S3 Vectors | In-process | Vector buckets, indexes, put / get / list / delete vectors, cosine similarity queries |
 | DynamoDB | In-process | GSI, LSI, Query, Scan, TTL, transactions, batch operations; Streams with shard iterators and Lambda event source mapping |
 | Lambda | Real Docker | Runtime environment, execution model, warm container pool, aliases, Function URLs |
--- a/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/S3AnnotationsTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/com/floci/test/S3AnnotationsTest.java
@@ -1,0 +1,224 @@
+package com.floci.test;
+
+import org.junit.jupiter.api.*;
+
+import software.amazon.awssdk.core.ResponseBytes;
+import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.AnnotationDirective;
+import software.amazon.awssdk.services.s3.model.AnnotationEntry;
+import software.amazon.awssdk.services.s3.model.BucketVersioningStatus;
+import software.amazon.awssdk.services.s3.model.ChecksumMode;
+import software.amazon.awssdk.services.s3.model.CreateBucketRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectAnnotationResponse;
+import software.amazon.awssdk.services.s3.model.ListObjectAnnotationsRequest;
+import software.amazon.awssdk.services.s3.model.ListObjectAnnotationsResponse;
+import software.amazon.awssdk.services.s3.model.NoSuchAnnotationException;
+import software.amazon.awssdk.services.s3.model.PutBucketVersioningRequest;
+import software.amazon.awssdk.services.s3.model.PutObjectAnnotationRequest;
+import software.amazon.awssdk.services.s3.model.VersioningConfiguration;
+
+import static org.assertj.core.api.Assertions.*;
+
+@DisplayName("S3 Object Annotations")
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+class S3AnnotationsTest {
+
+    private static S3Client s3;
+    private static final String BUCKET = "sdk-annotations-bucket";
+    private static final String KEY = "docs/annotated.txt";
+
+    @BeforeAll
+    static void setup() {
+        s3 = TestFixtures.s3Client();
+        s3.createBucket(CreateBucketRequest.builder().bucket(BUCKET).build());
+        s3.putObject(r -> r.bucket(BUCKET).key(KEY), RequestBody.fromString("annotated object body"));
+    }
+
+    @AfterAll
+    static void teardown() {
+        for (String key : new String[]{"docs/annotated.txt", "docs/annotated-copy.txt", "docs/annotated-copied.txt"}) {
+            try {
+                s3.deleteObject(r -> r.bucket(BUCKET).key(key));
+            } catch (Exception ignored) {
+            }
+        }
+        try {
+            s3.deleteBucket(r -> r.bucket(BUCKET));
+        } catch (Exception ignored) {
+        }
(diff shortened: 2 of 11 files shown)


pgermosen (approved):
Traced both follow-up fixes myself rather than taking them at face value, and they hold up, especially the lock-monitor one: synchronizing on requireBucket's result instead of resolveBucket's would have silently locked on a different object than what the actual write path locks under globalBucketNamespace, the kind of bug that produces no error until a race actually hits it. Also checked the x-amz-object-annotation-directive claim against the authoritative API reference myself since the user guide prose actually disagrees with it, and the PR trusted the right source.

Everything else, the shapes, limits, error codes, versioning semantics, checked out clean across two independent passes. Good to merge.

PR #3327, opened 2026-09-09
fix(apigateway): implement MOCK integration response selection based on status code

### LINKED ISSUE #2847: [BUG] restapis call from browser does not use gateway CORs settings
## Service

API Gateway

## AWS API Action

Make a POST request to a defined /restapis endpoint route (POST with a body value).

## Expected behavior

Necessary CORs headers attached to the response.

## Actual behavior

Response in browser indicate missing headers;

No 'Access-Control-Allow-Origin' header is present on the requested resource.

## Reproduction

Set up an API Gateway with necessary CORs handling attached and then call the endpoint

```bash
curl -I -X OPTIONS \
  -H "Origin: http://localhost:3000" \
  -H 'Access-Control-Request-Method: GET' \
  "${MY_URL}/test" 2>&1
```
You should see a missing Access-Control-Allow-Origin header.

## Environment

- Floci latest image tag (v1.7.0):
- Running using docker compose;
```
  floci:
    image: floci/floci:latest
    container_name: floci
    environment:
      # FLOCI_STORAGE_MODE: hybrid
      EXTRA_CORS_ALLOWED_ORIGINS: http://localhost:3000
      DISABLE_CORS_CHECKS: 1
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock # Required for Lambda container execution
      # - ./data:/app/data
    ports:
      - "4566:4566"  
```

### LINKED ISSUE #2878: [BUG] API Gateway OPTIONS (MOCK) responses missing CORS headers in 2.0 (regression from 1.7)
## Service

API Gateway (REST/v1)

## AWS API Action

N/A — this is a data-plane execute-api call (CORS preflight via a `cors: true` Serverless Framework MOCK integration), not a control-plane API. The OPTIONS method itself is registered correctly (confirmed via CreateRestApi/PutMethod), the issue is in the response the MOCK integration returns.

## Expected behavior

The OPTIONS preflight response should include `Access-Control-Allow-Origin`, `Access-Control-Allow-Methods`, and `Access-Control-Allow-Headers`, as it did in 1.7.0.

## Actual behavior

`200 OK` with `content-length: 0` and no CORS headers at all:

```
HTTP/1.1 200 OK
Date: Tue, 01 Sep 2026 09:22:39 GMT
x-amz-id-2: cbfb726b-97e5-4156-acdd-1766729f208a
x-amz-request-id: cbfb726b-97e5-4156-acdd-1766729f208a
x-amzn-RequestId: cbfb726b-97e5-4156-acdd-1766729f208a
content-length: 0
```

This causes browsers to block the actual request with "Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource."

Debug log shows the request is routed but presumably not producing CORS headers:

```
DEBUG [io.git.hec.flo.ser.api.ApiGatewayExecuteController] execute-api: OPTIONS aquassay/local/infrastructures/config → MOCK
```

Setting `FLOCI_SECURITY_EXTRA_CORS_ALLOWED_ORIGINS=*` and `FLOCI_SECURITY_EXTRA_CORS_ALLOWED_HEADERS=*` (which fully fixed this on 1.7.0) has no effect on 2.0.

## Reproduction

```bash
# API deployed via Serverless Framework with `cors: true`, generating a standard
# MOCK integration for OPTIONS on the resource.

curl -i -X OPTIONS 'http://localhost:4566/restapis/<rest_api_id>/local/_user_request_/infrastructures/config' \
  -H 'Origin: https://example.local' \
  -H 'Access-Control-Request-Method: GET' \
  -H 'Access-Control-Request-Headers: authorization,x-api-key'

# Expected: Access-Control-Allow-Origin / -Methods / -Headers in the response
# Actual: 200, content-length: 0, no CORS headers
```

## Environment

- Floci version / image tag: `floci/floci:latest` (2.0), also reproduced not present on `floci/floci:latest-compat` 1.7.0 with the workaround env vars above — regression is specific to 2.0
- Java SDK version (if applicable): N/A (browser XHR via Axios / API Gateway execute-api, not a service-side SDK call)

matheusandre1:
## Summary

 Closes: #2847

Likely also fixes #2878 (same symptom on a Serverless Framework `cors: true` MOCK) — not reproduced with that exact setup, so left as "likely".

A `MOCK` integration on the REST execute plane only ever looked up the integration response keyed `"200"`. Any CORS preflight declared with a different status came back as a bare `200` with `content-length: 0` and no `Access-Control-*` headers, so browsers failed with "No 'Access-Control-Allow-Origin' header is present". This is exactly what CDK's `addCorsPreflight` produces (a `{ statusCode: 200 }` request template plus a single `204` integration response), and it also affected any setup relying on `selectionPattern`.

`invokeMock` now does what API Gateway does:

- renders the request template through VTL (picked by the request `Content-Type`, then `application/json`, then the only template configured) and reads the `statusCode` it produces — the unquoted `{ statusCode: 200 }` shorthand CDK emits is accepted;
- selects the integration response for that status: the first `selectionPattern` that matches wins, otherwise the response without a pattern (the default) answers;
- responds with the selected response's `statusCode`, `responseTemplates` and `responseParameters` (static `'value'` headers are what carry the CORS headers).

Leniencies kept/documented: with **no** integration responses at all Floci still answers an empty `200` so a bare MOCK stays usable as a stub (AWS returns a 500 configuration error). With responses configured but none selectable, it now returns `500 Internal server error` like AWS.

The other two pieces of the browser flow (routing OPTIONS to the integration instead of the global CORS filter, #1955, and the `EXTRA_CORS_ALLOWED_ORIGINS` alias) already landed in 2.0.x; this was the remaining gap.

## Type of change

- [X] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## AWS Compatibility

Incorrect behavior: `OPTIONS /execute-api/{id}/{stage}/path` (and the `_user_request_` form) against a MOCK whose only integration response is `204` returned `200` with no headers. AWS renders the MOCK request template, matches the resulting `statusCode` against each integration response's `selectionPattern`, and falls back to the default (pattern-less) response — documented in "Set up mock integrations in API Gateway" / "Integration response selection". No SDK/CLI wire change; this is data-plane behavior only.

## Checklist

### DIFF
--- a/docs/services/api-gateway.md
+++ b/docs/services/api-gateway.md
@@ -186,7 +186,7 @@ These management-plane operations have no handler in v1. Calls will return `404`
 - Client Certificates (5 operations)
 - `GetExport` / `ImportDocumentationParts`
 
-The execute plane (actual proxied HTTP traffic via `/restapis/{id}/{stage}/_user_request_/…`) is implemented separately and is not counted as management-plane operations. It supports `AWS_PROXY` (Lambda proxy), `AWS` (Lambda with VTL request/response templates), and `MOCK` integrations; other integration types return an error.
+The execute plane (actual proxied HTTP traffic via `/restapis/{id}/{stage}/_user_request_/…`) is implemented separately and is not counted as management-plane operations. It supports `AWS_PROXY` (Lambda proxy), `AWS` (Lambda with VTL request/response templates), and `MOCK` integrations; other integration types return an error. A `MOCK` integration renders its request template and uses the `statusCode` it produces to pick the integration response, exactly as AWS does: the first response whose `selectionPattern` matches wins, otherwise the response without a pattern (the default) answers. This is what makes CORS preflights declared with a `204` response (CDK's `addCorsPreflight`) carry their `Access-Control-*` headers.
 
 ### Examples
 
--- a/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
+++ b/src/main/java/io/github/hectorvent/floci/services/apigateway/ApiGatewayExecuteController.java
@@ -57,6 +57,7 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 /**
  * Executes API Gateway stage requests, routing them through the configured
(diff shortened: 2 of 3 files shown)


pgermosen:
This looks solid, checked the MOCK status-code selection against the exact AWS doc page you cited and it matches precisely, including the selectionPattern-vs-default fallback and the 500-on-no-match behavior. Test coverage hits every case well, including the exact CDK addCorsPreflight shape from the original issue.

CI is red but not because of this PR. Shard 3's SsoAdminIntegrationTest failures are a pre-existing identity-center bug that's already fixed on main (#3388), your branch is just behind it. Shard 1's RdsAwsIntegrationTest failure is an unrelated known flake. Neither touches anything in this PR.

Can you rebase onto current main? That should clear shard 3 at least. Once CI is green this looks ready to merge.

pgermosen (approved):
Re-checked the AWS behavior myself rather than carrying the earlier pass forward, since it's been rebased since. The subtlest part of this is right: the default response is picked because it has no selectionPattern, not because of whatever status it happens to be keyed under, which is exactly what AWS's own docs describe and a naive implementation would likely get backwards. The render-before-inspect ordering holds up too.

Good to merge.

